PR #701, opened 2026-07-01
Fix compileFile preamble skip and GC safety

### LINKED ISSUE #699: compileFile skips preamble on error + GC safety issue
## Summary

`compileFile` in `src/main.zig` has two related bugs that break standalone binary support via `-Dbundle`:

1. **Preamble skip on error**: When `handleTopLevelForm` returns an error (e.g., `CompileError` from a library import), the `continue` on the error path skips `preamble.append`. The form is not recorded in the `.sbc` preamble section, so bundled binaries never replay the import.

2. **GC safety**: `expr` is declared `const` and not GC-rooted before `handleTopLevelForm`. The import processing may trigger GC, freeing the cons cells that make up `expr`. Attempting to stringify `expr` afterwards segfaults in `printer.zig:505`.

## Reproduction

```bash
# Any program that imports libraries with non-fatal CompileError
kaappi --lib-path lib --compile program.scm -o program.sbc
zig build -Dbundle=program.sbc
./zig-out/bin/kaappi  # library functions undefined
```

The c0c compiler triggers this because importing `(c0c driver)` produces a non-fatal `CompileError` (the functions ARE still defined in interpreted mode). During `--compile`, the import form is skipped from the preamble, so the bundled binary can't replay it.

## Root Cause

`src/main.zig`, `compileFile` function (around line 1060):

```zig
if (vm.handleTopLevelForm(expr)) |top_result| {
    _ = top_result catch |err| {
        // error logging...
        vm.last_error_detail_len = 0;
        continue;  // <-- BUG: skips preamble.append below
    };
    const form_src = printer.valueToString(allocator, expr, .write) catch continue;
    preamble.append(allocator, form_src) catch { ... };
    continue;
}
```

The `continue` on line 1073 skips lines 1075-1078, so the form is never recorded. Additionally, `expr` on line 1052 is `const` and unrooted — if `handleTopLevelForm` triggers GC, `expr` becomes a dangling pointer, and `printer.valueToString` segfaults.

## Proposed Fix

1. Change `expr` from `const` to `var` and root it before `handleTopLevelForm`:
   ```zig
   var expr = r.readDatum() catch ...;
   vm.gc.pushRoot(&expr) catch continue;
   defer vm.gc.popRoot();
   ```

2. Move `printer.valueToString` before the error handler, and remove the `continue` so the preamble is always appended:
   ```zig
   if (vm.handleTopLevelForm(expr)) |top_result| {
       const form_src = printer.valueToString(allocator, expr, .write) catch continue;
       _ = top_result catch |err| {
           // error logging...
           vm.last_error_detail_len = 0;
       };
       preamble.append(allocator, form_src) catch { allocator.free(form_src); };
       continue;
   }
   ```

This ensures all top-level forms (including erroring imports) are recorded in the preamble, and `expr` survives GC during import processing.

## Impact

- Bundled binaries (`-Dbundle`) cannot replay library imports that had non-fatal errors
- The GC safety issue causes segfaults when trying to fix the preamble skip
- Blocks standalone binary support for projects with complex library dependencies

## Related

This issue interacts with #NNN (handleImportInto aborts on first failure) — even with preamble forms recorded, the replay fails if `handleImportInto` can't process all sub-imports.

baijum:
## Summary
- **GC root expr** before `handleTopLevelForm` — `expr` is now `var` and rooted with `pushRoot`/`popRoot` so GC during library loading can't free the cons cells before stringification
- **Always append to preamble** — `valueToString` runs before the error catch block (while `expr` is still valid), and the `continue` is removed so `preamble.append` always executes regardless of whether `top_result` errored

Fixes #699

## Test plan
- [x] `bash tests/scheme/run-all.sh` — 1583 pass, 0 fail
- [x] `zig build -Dbundle-src=...` — bundled binary correctly replays imports

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -1057,9 +1057,10 @@ fn compileFile(vm: *vm_mod.VM, path: []const u8, output_path: ?[]const u8) !void
             return;
         };
 
+        vm.gc.pushRoot(&expr) catch continue;
+        defer vm.gc.popRoot();
+
         if (vm.handleTopLevelForm(expr)) |top_result| {
-            vm.gc.pushRoot(&expr) catch continue;
-            defer vm.gc.popRoot();
             const form_src = printer.valueToString(allocator, expr, .write) catch continue;
             _ = top_result catch |err| {
                 const detail = vm.getErrorDetail();
--- a/src/main.zig
+++ b/src/main.zig
@@ -1049,7 +1049,7 @@ fn compileFile(vm: *vm_mod.VM, path: []const u8, output_path: ?[]const u8) !void
         return;
     }) {
         const datum_lc = r.getLineCol();
-        const expr = r.readDatum() catch |err| {
+        var expr = r.readDatum() catch |err| {
             const lc = r.getLineCol();
             var errbuf: [256]u8 = undefined;
             const s = std.fmt.bufPrint(&errbuf, "{s}:{d}:{d}: read error: {}\n", .{ path, lc.line, lc.col, err }) catch "read error\n";
@@ -1058,6 +1058,9 @@ fn compileFile(vm: *vm_mod.VM, path: []const u8, output_path: ?[]const u8) !void
         };
 
         if (vm.handleTopLevelForm(expr)) |top_result| {
+            vm.gc.pushRoot(&expr) catch continue;
+            defer vm.gc.popRoot();
+            const form_src = printer.valueToString(allocator, expr, .write) catch continue;
             _ = top_result catch |err| {
                 const detail = vm.getErrorDetail();
                 if (detail.len > 0) {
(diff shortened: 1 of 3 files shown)


baijum:
The GC root placement needs to move **before** `handleTopLevelForm`. Currently it's inside the `if` block:

```zig
if (vm.handleTopLevelForm(expr)) |top_result| {
    vm.gc.pushRoot(&expr) catch continue;  // too late — GC already ran
```

By this point, `handleTopLevelForm` has already executed the import (heavy allocation, GC cycles). The cons cells making up `expr` are freed before `pushRoot` gets a chance to protect them. This segfaults on `valueToString` when compiling c0c (whose `(c0c driver)` import triggers enough allocation to cause a GC sweep).

Reproducer:
```bash
cd ../c0c
kaappi --lib-path lib --compile c0c.scm -o c0c.sbc
# => Segmentation fault at printer.zig:505
```

The root needs to be outside the `if`, before `handleTopLevelForm` is called:

```zig
var expr = r.readDatum() catch |err| { ... };
vm.gc.pushRoot(&expr) catch continue;
defer vm.gc.popRoot();

if (vm.handleTopLevelForm(expr)) |top_result| {
    const form_src = printer.valueToString(allocator, expr, .write) catch continue;
    _ = top_result catch |err| {
        // error logging
        vm.last_error_detail_len = 0;
    };
    preamble.append(allocator, form_src) catch { allocator.free(form_src); };
    continue;
}
```

This protects `expr` through the entire `handleTopLevelForm` call. The non-top-level path (`compileExpressionWithMacrosAt`) is also covered since the root stays active via `defer`.

Issue #868, opened 2026-07-02
cond-expand (library ...) in define-library is false for importable-but-unloaded .sld libraries


baijum:
## Summary

Inside `define-library`, the `cond-expand` feature requirement `(library <name>)` evaluates to false for any library that has not already been loaded into the registry — even though `(import <name>)` for that same library succeeds. Portable fallback code of the form

```scheme
(cond-expand
  ((library (srfi 41)) (import (srfi 41)) ...)
  (else <fallback>))
```

always takes the `else` branch for on-disk `.sld` libraries.

## Location

`src/vm_library.zig:560-567` (v0.11.1, commit d46986f), in `evalLibFeatureReq`:

```zig
if (std.mem.eql(u8, op, "library")) {
    ...
    return vm.libraries.get(lib_name) != null;   // only checks already-loaded registry
}
```

The check consults only `vm.libraries` (already-registered libraries). It never attempts path resolution the way `processImportSet` does (`buildLibRelPath` + `resolveLibraryPath` / bundled files), so all 64 portable SRFI `.sld` files and every user library visible via `--lib-path` are reported as absent until something else imports them first.

## Reproduction

```scheme
; celib/one.sld
(define-library (celib one)
  (import (scheme base))
  (export which)
  (cond-expand
    ((library (srfi 41)) (begin (define which 'srfi41-available)))
    (else                (begin (define which 'srfi41-missing)))))
```

```scheme
; t.scm
(import (scheme base) (scheme write) (celib one))
(display which) (newline)
(import (srfi 41))                 ; proves the library IS importable
(display "srfi41 imported ok") (newline)
```

```
$ kaappi --lib-path . t.scm
srfi41-missing
srfi41 imported ok
```

The same happens for user libraries: a `(cond-expand ((library (celib one)) ...))` in a sibling library reports `not-found` even though `(import (celib one))` works.

## Actual vs expected

- Actual: `(library X)` is true only if X is a built-in library or was already loaded earlier in the process.
- Expected: R7RS-small 4.2.1: "Each implementation maintains a list of feature identifiers which are present, **as well as a list of libraries which can be imported.**" A library that `import` can load from an `.sld` file "can be imported" and must satisfy `(library ...)`. (Behavior confirmed inconsistent: the very same name imports fine one line later.)

## Root cause

[...]

PR #916, opened 2026-07-02
Route include forms in library bodies through handleTopLevelForm

### LINKED ISSUE #879: (include ...) inside a library begin block or library-included file fails: "undefined variable 'include'"
## Summary

`(include "file.scm")` appearing *inside* a library body — either directly in a `(begin ...)` library declaration, or nested inside a file pulled in by a library-level `(include ...)` declaration — is not recognized as the R7RS `include` syntax. It is compiled as a call to a variable named `include`, and the library fails to load with:

```
error: undefined variable 'include'
```

The same form works at program top level (handled by `handleTopLevelInclude`) and as a direct `define-library` declaration (handled by `compileLibInclude`), so support is inconsistent by position.

## Location

`src/vm_library.zig` (v0.11.1, commit d46986f):

- `compileLibExpr` (lines 1043-1064): body forms are either routed to `vm.handleTopLevelForm` — but only when `isLibTopLevelForm` matches, and that predicate (lines 1204-1212) accepts only `define-record-type` and `define-values` — or compiled via `compileExpressionInEnv`, where the compiler has no `include` form. `(include ...)` therefore compiles to a call of the unbound variable `include`.
- `compileLibInclude` (lines 1067-1113) reads each datum of the included file and feeds it back through `compileLibExpr` (line 1108), so a nested `(include ...)` in an included file hits the same hole — even though `handleTopLevelInclude`'s doc comment (lines 572-582) documents that nested includes are supposed to resolve relative to the including file.

## Reproduction

Direct begin-block case:

```scheme
; blib/lib.sld
(define-library (blib lib)
  (import (scheme base))
  (export bv)
  (begin
    (include "bpart.scm")))     ; bpart.scm: (define bv 'begin-include-ok)
```
```
$ kaappi --lib-path . t.scm     ; t.scm imports (blib lib) and displays bv
t.scm:1: error: undefined variable 'include'
t.scm:2: error: undefined variable 'bv'. Did you mean '+'?
```

Nested case:

```scheme
; inc/lib.sld
(define-library (inc lib)
  (import (scheme base))
  (export v1 v2)
  (include "part1.scm"))
; inc/part1.scm
(define v1 'part1)
(include "sub/part2.scm")       ; sub/part2.scm: (define v2 'part2)
```
```
$ kaappi --lib-path . t2.scm    ; imports (inc lib)
t2.scm:1: error: undefined variable 'include'
```

Control: the identical nested-include structure at program top level (a `.scm` including a file that includes another file) works fine.

## Actual vs expected

- Actual: library load aborts with "undefined variable 'include'"; the library is not registered.
- Expected: R7RS-small 4.1.7 — `include`/`include-ci` are `(scheme base)` syntax usable wherever definitions/expressions can appear; the included files' contents behave as if written in place. Chibi et al. accept both positions.

## Root cause

`compileLibExpr` has no dispatch for `include`/`include-ci`. Routing those heads to `compileLibInclude(vm, lib_env, cdr)` (which already exists and handles path resolution relative to `vm.current_lib_dir`) closes both the begin-block and nested-file cases.

---
Found by automated code analysis (Claude).

baijum:
Fixes #879

### DIFF
--- a/src/vm_library.zig
+++ b/src/vm_library.zig
@@ -1202,5 +1202,6 @@ fn isLibTopLevelForm(expr: Value) bool {
     const name = types.symbolName(head);
     if (std.mem.eql(u8, name, "define-record-type")) return true;
     if (std.mem.eql(u8, name, "define-values")) return true;
+    if (std.mem.eql(u8, name, "include")) return true;
     return false;
 }


baijum:
Superseded by #917 which includes this fix plus the #868 fix.

PR #923, opened 2026-07-03
Fix double hygiene renaming in macro-generating macros

### LINKED ISSUE #919: Nested define-syntax hygiene: macro-generating macros mis-scope inner syntax-rules
The last remaining R7RS suite failure in section 4.3 Macros (verified still failing 2026-07-03). Migrated from `docs/dev/nested-define-syntax-hygiene.md` (written 2026-06-18), which this issue replaces.

## Current suite failure

`tests/scheme/r7rs/r7rs-tests.scm:536`:

```scheme
(define-syntax jabberwocky
  (syntax-rules ()
    ((_ hatter)
     (begin
       (define march-hare 42)
       (define-syntax hatter
         (syntax-rules ()
           ((_) march-hare)))))))
(jabberwocky mad-hatter)
(test 42 (mad-hatter))
```

Fails with: `error: undefined variable '__hyg_373___hyg_369_march-hare'` — the reference to `march-hare` inside the inner macro's template gets hygiene-renamed twice.

## Minimal repro (pattern-variable scoping variant)

```scheme
(let ()
 (define-syntax foo
   (syntax-rules ()
     ((foo bar y)
      (define-syntax bar
        (syntax-rules ()
          ((bar x) 'y))))))
 (foo bar x)
 (test 'x (bar 1)))     ;=> expected: x, actual fails
```

**Expected:** `(bar 1)` returns the symbol `x` — the value of `y` captured when `(foo bar x)` was expanded.

## Root cause

The expander (`src/expander.zig`) treats `syntax-rules` forms inside templates as ordinary lists — it recursively walks them, substituting the outer macro's pattern variables and renaming identifiers. It does not recognize that an inner `syntax-rules` form's patterns introduce a new binding scope: per R7RS §4.3.2, pattern variables are scoped to their own `syntax-rules` form, and identifiers in inner templates must resolve in the inner macro's expansion context, not be renamed/substituted by the outer expansion.

## Possible fix

In `instantiateTemplate` (`src/expander.zig`), when encountering a `syntax-rules` form in a template:

1. Detect `(syntax-rules (...) ...)` as a special template form
2. Collect the inner pattern variable names from the inner rules
3. When recursing into the inner templates, exclude the inner pattern variables from the outer substitution bindings
4. Only substitute outer bindings that don't conflict with inner patterns

The double-renaming in the jabberwocky case additionally suggests hygiene renaming must be applied once per expansion level, not re-applied when the inner macro's stored template is instantiated.

## Impact

Low — macro-generating macros are rare; all common macro patterns work. This is a conformance edge case.

## Key files

| Component | Location |
|-----------|----------|
| Template instantiation | `src/expander.zig` (`instantiateTemplate`) |
| Pattern matching | `src/expander.zig` (`matchPattern`) |
| Compiler macro dispatch | `src/compiler.zig` |

baijum:
Fixes #919.

## Problem

When a macro's template defines another macro, identifiers in the inner template that were already hygiene-renamed by the outer expansion (e.g. `__hyg_369_march-hare`) got renamed a **second time** when the inner macro expanded, producing `__hyg_373___hyg_369_march-hare` — a name that no longer matches the binding the outer expansion created:

```scheme
(define-syntax jabberwocky
  (syntax-rules ()
    ((_ hatter)
     (begin
       (define march-hare 42)
       (define-syntax hatter
         (syntax-rules ()
           ((_) march-hare)))))))
(jabberwocky mad-hatter)
(mad-hatter)   ; error: undefined variable '__hyg_373___hyg_369_march-hare'
```

## Fix

`renameForHygiene` in `src/expander.zig` now returns identifiers that already carry the `__hyg_` prefix unchanged. Renaming an already-renamed identifier cannot prevent any capture — gensyms are globally unique, so they can never collide with user identifiers. Renaming again only severs the reference from the binding created by the generating expansion. This approximates proper sets-of-scopes behavior: adding a scope to an identifier never stops it from resolving to a binding carrying a subset of its scopes.

The pattern-variable scoping variant from the issue (`instantiateNestedSyntaxRules` excluding inner pattern variables from outer substitution) already worked; it now has a guard test.

## Tests

- New Zig unit tests in `src/tests_macros.zig`: the jabberwocky case (fails without the fix, passes with it) and the inner-pattern-variable shadowing case.
- New "macro-generating macros" section in `tests/scheme/hygiene/hygiene.scm`.
- The jabberwocky test at `tests/scheme/r7rs/r7rs-tests.scm:536` now passes; the R7RS suite reports 1393 pass, 0 fail.
- `zig build test` (439 tests) and `bash tests/scheme/run-all.sh` fully green.

## Out of scope (found while verifying, flagged separately)

- `r7rs-tests.scm:580` (forward hygienic refs — template referencing a local defined later in the same body) and `r7rs-tests.scm:633` (section 5 compile error) still error; different root causes.
- SRFI-64 assertions in `tests/scheme/` are currently silent no-ops on main (`%test-on-test-begin` undefined, exit 0), so the hygiene.scm addition only becomes an active guard once that is fixed. The Zig unit tests carry the regression coverage meanwhile.

### DIFF
--- a/src/expander.zig
+++ b/src/expander.zig
@@ -744,6 +744,13 @@ fn renameForHygiene(gc: *GC, name: []const u8, scope: u32, globals: ?*std.String
     const QUOTE_FLAG: u32 = 0x40000000;
     const BINDING_FLAG: u32 = 0x20000000;
     if ((scope & QUOTE_FLAG) != 0) return gc.allocSymbol(name);
+
+    // Already renamed by an enclosing expansion: macro-generating macros
+    // bake __hyg_ names into the inner macro's stored template. Gensyms are
+    // globally unique, so renaming again cannot prevent any capture — it
+    // only severs the reference from the binding created by the generating
+    // expansion (issue #919: __hyg_N___hyg_M_march-hare undefined).
+    if (std.mem.startsWith(u8, name, "__hyg_")) return gc.allocSymbol(name);
     const in_binding = (scope & BINDING_FLAG) != 0;
     const clean_scope = scope & ~BINDING_FLAG;
     if (globals) |g| {
--- a/src/tests_core_eval.zig
+++ b/src/tests_core_eval.zig
@@ -180,3 +180,21 @@ test "default-random-source is per-VM" {
     const rs1_again = try vm1.eval("(default-random-source)");
     try std.testing.expectEqual(rs1, rs1_again);
 }
+
+test "vm deinit clears threadlocal vm_instance" {
+    const vm_mod = @import("vm.zig");
+    var gc = memory.GC.init(std.testing.allocator);
+    defer gc.deinit();
+    var vm = try th.makeTestVM(&gc);
+
+    // execute() registers the VM in the threadlocal
+    _ = try vm.eval("(+ 1 2)");
+    try std.testing.expect(vm_mod.vm_instance == &vm);
+
+    // deinit must unregister it: a stale pointer here is read by the macro
+    // expander (renameForHygiene) during the next VM's first compile, before
+    // that VM's own execute() re-registers the threadlocal — a use-after-free
+    // that crashed the Linux unit-test runs.
+    vm.deinit();
+    try std.testing.expect(vm_mod.vm_instance == null);
+}
--- a/src/tests_macros.zig
+++ b/src/tests_macros.zig
@@ -299,3 +299,52 @@ test "let-syntax rejects malformed binding without transformer" {
     const result = vm.eval("(let-syntax ((my-macro)) 1)");
     try std.testing.expectError(error.CompileError, result);
 }
+
+test "hygiene: macro-generating macro shares binding with inner macro" {
+    var gc = memory.GC.init(std.testing.allocator);
+    defer gc.deinit();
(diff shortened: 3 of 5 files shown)


baijum:
CI failure analysis: all Linux jobs crashed (SIGSEGV) in the new `hygiene: inner syntax-rules pattern variables shadow outer bindings` unit test, inside `renameForHygiene`'s `vm.globals.get(name)`.

Root cause is a pre-existing use-after-free, not the hygiene fix itself: `execute()` registers the running VM in the threadlocal `vm_instance`, but `VM.deinit` never unregistered it. In the unit-test binary each test's VM lives on the test's stack; after a test tears down, the threadlocal dangles. The next test's first `eval` macro-expands at **compile time** — before its own `execute()` re-registers the threadlocal — so the expander read the previous test's freed globals map. macOS malloc happened to keep the pages readable; glibc faulted. The new test merely shifted heap reuse enough to expose it.

Fixed in 6bfb60f: `VM.deinit` now nulls `vm_instance` when it points at the dying VM, plus a deterministic lifecycle test (asserts registration after eval and unregistration after deinit) so this doesn't regress silently on platforms where the UAF is benign.

Issue #681, opened 2026-07-01
Macro hygiene broken for template-introduced bindings named after built-ins (list, map, cons, etc.)


baijum:
## Bug

`src/expander.zig` maintains a `well_known_forms` allowlist (lines 37-81)
of special-form keywords and common built-in procedure names (`list`,
`cons`, `map`, `car`, `cdr`, `+`, etc.). `instantiateTemplate` skips
hygienic renaming for any template identifier matching this list:

```zig
        // 3. Well-known form or built-in -- keep as-is
        if (isWellKnown(name)) {
            return template;
        }
```
(src/expander.zig:441-444)

This is meant to let templates *reference* built-ins without mangled
names, but it also means that when a macro's template *introduces a new
binding* using one of these names (e.g. `(let ((list ...)) ...)`), that
binding is emitted verbatim (unhygienic) instead of being alpha-renamed.//
Per R7RS hygiene (section on macros, "transparency" requirement), a
template-introduced binding must never capture a reference to the same
name that a macro *user* passes in as an argument — including the user's
reference to a global/built-in procedure of that name.

## Location

`src/expander.zig:441-444` (the `isWellKnown` short-circuit in
`instantiateTemplate`), in combination with `well_known_forms` at
`src/expander.zig:37-81`.

## Repro

```scheme
(define-syntax capture-test
  (syntax-rules ()
    ((_ expr)
     (let ((list (lambda (x y) (cons 'captured (cons x (cons y '()))))))
       expr))))

(display (capture-test (list 1 2)))
```

Run with:
```
zig build run -- repro.scm
```

**Actual:** prints `(captured 1 2)` — the user's `(list 1 2)` expression,
passed as an argument into the macro, gets captured by the macro's
*internal* `list` binding instead of referring to the top-level `list`
procedure.

**Expected:** prints `(1 2)` — per referential transparency, the user's
reference to `list` inside the argument expression must resolve to the
`list` visible at the macro *use* site (the built-in), unaffected by the
macro's own internal binding of the same name.

## Notes

This is the same class of bug the `well_known_forms` mechanism was
explicitly built to avoid for macro-defined temporaries like `temp` in
`my-or` (see `tests/scheme/hygiene/hygiene.scm`), but the allowlist
approach only protects *references* to well-known names, not *bindings*
introduced with those same names. Any macro that happens to bind a local
variable whose name coincides with one of the ~140 entries in
`well_known_forms` will silently break hygiene for user code passed
through it.

PR #929, opened 2026-07-03
Fix SRFI-64 suite silently asserting nothing; flip exit code on script errors

### LINKED ISSUE #924: SRFI-64 assertions silently no-op: exported macros can't resolve library internals; uncaught errors still exit 0
Observed 2026-07-03 on `main` (9d5d077) while verifying #919 / PR #923. The entire Scheme-level test suite except the R7RS suite currently asserts nothing, and `run-all.sh` reports it all as PASS.

## Repro

```bash
zig build
./zig-out/bin/kaappi tests/scheme/smoke/basic.scm; echo "exit: $?"
```

Output:

```
%%%% Starting test basic  (Writing full log to "basic.log")
tests/scheme/smoke/basic.scm:8: error: undefined variable '%test-on-test-begin'
    (test-assert "literal 42" 42)
tests/scheme/smoke/basic.scm:9: error: undefined variable '%test-on-test-begin'
    (test-eqv "addition" 3 (+ 1 2))
...
exit: 0
```

Every top-level `test-assert`/`test-eqv`/`test-equal` errors with `undefined variable '%test-on-test-begin'`. Every `test-group` errors with `type error in 'dynamic-wind': got #<procedure>`. Same behavior across `tests/scheme/smoke/`, `hygiene/`, etc. Reproduces with a clean tree, freshly deleted `.sbc` caches, and with `~/.kaappi/lib` removed from the search path — not environment pollution.

Two aggravating factors:

1. **Error line numbers are misattributed** to the user's file (e.g. `hygiene.scm:307` in a 166-line file) instead of the library source, hiding that the failure is inside `(srfi 64)`.
2. **Uncaught runtime errors abort the enclosing top-level form but the process still exits 0.** `run-all.sh` gates on exit codes, so all 237 files report PASS while asserting nothing. The R7RS suite is unaffected only because `(chibi test)` counts pass/fail itself.

## Root cause (hypothesis)

`test-assert` and friends are `syntax-rules` macros exported from `lib/srfi/64.sld`. Their templates reference library-internal procedures — `%test-on-test-begin` is a plain `define` at `lib/srfi/64.sld:470`, called from the `%test-comp2body`/`%test-comp1body` macro templates (lines 484–511). When the macro expands at the use site, the reference to the internal binding no longer resolves there.

This smells like fallout from the library export isolation work (PR #910 “exports-clean”) and is the flip side of #877 (define-syntax in library bodies registering macros globally): exported macros whose templates reference unexported library internals need those references to keep resolving in the library's environment after expansion at the use site.

## Impact

High — the whole `tests/scheme/` suite (smoke, compliance, hygiene, srfi, ffi, audit) is a silent no-op, so CI stays green regardless of regressions in anything those files cover.

## Suggested fix

1. Make identifiers in exported macro templates resolve against the defining library's environment (or qualify/rename them to reachable names at export time), so SRFI-64 assertions work again.
2. Separately: make uncaught runtime errors in script mode set a non-zero exit code (verify `tests/scheme/run-all.sh` semantics first) so this class of breakage can never report PASS again.

## Key files

| Component | Location |
|-----------|----------|
| SRFI-64 implementation | `lib/srfi/64.sld` (`%test-on-test-begin` at 470, `%test-comp2body` at 484) |
| Library loading / exports | `src/vm_library.zig` |
| Hygiene renaming globals check | `src/expander.zig` (`renameForHygiene`) |
| Script error handling / exit code | `src/main.zig` |
| Suite runner | `tests/scheme/run-all.sh` |

baijum:
Fixes #924

The entire `tests/scheme/` suite (except the R7RS suite) was a silent no-op: every SRFI-64 assertion errored at runtime yet files exited 0, so `run-all.sh` reported all 237 files as PASS while asserting nothing.

## Root cause — three stacked bugs

Bisection shows this predates v0.10.0 (it was **not** #910 fallout as the issue hypothesized — #910 only touched built-in exports in `library.zig`).

1. **Library-internal bindings weren't copied transitively** (`src/vm_library.zig`). `importBinding` copied an exported macro's template free references from the defining library's env only one level deep. `test-assert` expands into the *internal* macro `%test-comp1body`, whose own references (`%test-on-test-begin` …) were never copied because the helper itself is never imported. Now copied transitively through macro-to-macro chains, with cycle protection.
2. **Hygiene renamed binders but not references** (`src/expander.zig`). In `renameForHygiene`, the "global procedure ⇒ keep name" check ran before the scope-table lookup, so SRFI-64's `(let ((exp expected)) … exp)` bound `__hyg_N_exp` while body references resolved to the builtin `exp` — every test compared against `#<builtin exp>`. The scope table is now consulted first.
3. **`set!` lacked the hygienic fallback reads have** (`src/vm_dispatch.zig`). Template `(set! counter …)` to a definition-site global failed as unbound `__hyg_N_counter`; `set_global` now gets the same `stripHygienicPrefix` fallback as `get_global`/`call_global` (#681 follow-up).

The `test-group` "type error in 'dynamic-wind'" from the issue was the same bug in disguise: the group body failed with the undefined variable, `dynamicWindFn` ran the after-thunk (which resets the error detail), and the caller reported a generic type error.

## Exit codes

Uncaught read/compile/runtime errors in scripts (file or stdin) now flip the process exit code (`run-all.sh` gates purely on exit status). Explicit `(exit n)` still wins; the REPL is unaffected. New `tests/scheme/errors/exit-code.sh` covers 10 scenarios and is wired into CI. Also dropped `run-all.sh`'s retry with the removed `--no-jit` flag.

## Tests

- Unit: imported-macro-chain resolution (`tests_libraries.zig`), builtin-named template binding (`tests_macros.zig`)
- Scheme: `tests/scheme/smoke/library-macro-internals.scm` (SRFI-64-shaped macro chain via `define-library`)
- `ellipsis-mismatch.scm`'s negative case moved to `error-format.sh`: the mismatch is rejected at expansion (compile) time, which `guard` cannot catch and which now correctly fails the process.

## Expected CI state

`zig build test` passes; R7RS suite 1395 pass / 0 fail. Scheme files: 232 pass, **9 fail**, 1 skip — every failure is a pre-existing product bug the dead suite had been hiding (verified identical on the pre-fix binary): call/cc escapes, `hash-table-update!/default` losing keys, case-lambda arg binding, SRFI-35 `condition` macro, SRFI-133 `vector-partition`, two library-declaration bugs, and the srfi18 hang (skip). Each is being fixed in its own session; **CI stays red until they land** — that is the honest signal this PR exists to restore.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,9 @@ jobs:
       - name: Error format tests
         run: bash tests/scheme/errors/error-format.sh
 
+      - name: Exit code tests
+        run: bash tests/scheme/errors/exit-code.sh
+
       - name: E2E tests (LLVM native backend)
         if: matrix.optimize == 'ReleaseSafe' && matrix.os != 'macos-latest'
         run: |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -312,7 +312,7 @@ passes with it. Place it in the appropriate location:
   - `srfi/` — SRFI conformance tests
   - `ffi/` — C FFI integration tests
   - `audit/` — primitives audit tests (arithmetic, numeric, string)
-  - `errors/` — error message format regression tests (`error-format.sh`)
+  - `errors/` — error message format and exit code regression tests (`error-format.sh`, `exit-code.sh`)
   - `bench/` — raw micro-benchmarks (no assertions; used by `benchmarks/run-benchmarks.sh`, not `run-all.sh`)
   - `coverage/` — coverage gap-filler tests (used by `zig build coverage-scheme`, not `run-all.sh`)
 - **Run all**: `bash tests/scheme/run-all.sh`
--- a/src/main.zig
+++ b/src/main.zig
@@ -203,8 +203,17 @@ pub fn main(init: std.process.Init.Minimal) !void {
     return mainInner(init);
 }
 
+/// Set when a script (file or stdin) hit an uncaught read/compile/runtime
+/// error that was reported but recovered from. Scripts must exit non-zero in
+/// that case so callers (e.g. tests/scheme/run-all.sh) can detect the failure;
+/// REPL sessions never set this.
+var script_had_error: bool = false;
+
 fn mainInner(init: std.process.Init.Minimal) void {
-    mainImpl(init) catch {};
+    mainImpl(init) catch {
+        std.process.exit(1);
+    };
+    if (script_had_error) std.process.exit(1);
 }
 
 fn mainImpl(init: std.process.Init.Minimal) !void {
@@ -670,6 +679,7 @@ fn runFile(vm: *vm_mod.VM, path: []const u8) !void {
     defer vm.current_lib_dir = saved_lib_dir;
 
     const source = readFileContents(allocator, path) catch {
+        script_had_error = true;
         return;
     };
     defer allocator.free(source);
@@ -721,6 +731,7 @@ fn runFile(vm: *vm_mod.VM, path: []const u8) !void {
                 vm.gc.pushRoot(&func_val) catch return error.OutOfMemory;
                 const result = vm.execute(func) catch |err| {
                     vm.gc.popRoot();
+                    script_had_error = true;
                     const detail = vm.getErrorDetail();
                     if (detail.len > 0) {
                         var errbuf: [256]u8 = undefined;
@@ -765,6 +776,7 @@ fn runFile(vm: *vm_mod.VM, path: []const u8) !void {
         var errbuf: [256]u8 = undefined;
         const s = std.fmt.bufPrint(&errbuf, "{s}:{d}:{d}: read error: {}\n", .{ path, lc.line, lc.col, err }) catch "read error\n";
         writeStderr(s);
+        script_had_error = true;
         return;
     }) {
         const datum_lc = r.getLineCol();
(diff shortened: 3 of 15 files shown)


baijum:
Updated with two commits on top of the rebase, closing the remaining gaps found while verifying against current main:

**`310fcc6` — Keep `test-runner-current` set after the outermost `test-end`.** With the import fixes here, asserts recorded correctly, but the exit idiom documented in `tests/scheme/CLAUDE.md` — `(test-runner-fail-count (test-runner-current))` after `test-end` — still raised `type error in '%record-ref'`: the vendored `%test-begin` wraps `on-final` to reset `test-runner-current` to `#f` when the outermost group ends. Under the exit-code enforcement this PR adds, that made every *passing* SRFI-64 file exit non-zero. The runner now persists after `test-end` (matching Guile's SRFI-64), so the documented idiom works in both directions — verified by new `exit-code.sh` cases (all-pass file exits 0, failing-assert file exits 1) and `tests/scheme/smoke/srfi64-runner-persists.scm`. `library-macro-internals.scm` also now uses the documented idiom instead of capturing the count before `test-end`.

**`88764ba` — Mark hash-table entries occupied on insert** (`hash-table-update!/default` insert path and `alist->hash-table` omitted `.state = .occupied`, leaving inserted keys invisible to lookup). This was previously masked by exit code 0 on script errors; with enforcement it turned `tests/scheme/smoke/mutation-write-barrier.scm` and `tests/scheme/srfi/srfi69-ext.scm` into hard failures, so it needs to land with (or before) this PR.

Full local verification on macOS ARM: `zig build test` clean, `error-format.sh` 22/22, `exit-code.sh` 12/12, `run-all.sh` 247 files + 1395 R7RS tests, 0 fail, 0 skip.

PR #968, opened 2026-07-03
Fix heap corruption from SRFI-18 child threads touching shared parent state (#958)

### LINKED ISSUE #958: Flaky crash (SIGSEGV/SIGABRT) in deep-copy-list-801.scm — heap corruption via concurrent SRFI-18 deep-copy
## Summary

`tests/scheme/smoke/deep-copy-list-801.scm` crashes intermittently — roughly **30–40% of runs** — with a `Segmentation fault` / `Abort trap: 6` (exit 134). The crash **site varies between runs**, which points to heap metadata corruption caused by the concurrent GC / cross-thread deep-copy in the SRFI-18 thread path rather than a deterministic logic bug.

This is **pre-existing on `main`** (reproduced on `0f9d56c`, the current tip) and is not introduced by any particular feature branch — it surfaced with the test added for #801/#952.

## Reproduction

```bash
# ~30-40% of runs abort with a non-zero exit
for i in $(seq 1 10); do
  ./zig-out/bin/kaappi tests/scheme/smoke/deep-copy-list-801.scm >/dev/null 2>&1 || echo "crash run $i (exit $?)"
done
```

Measured: 4/10 aborts on pristine `main` (`0f9d56c`); similar rate on other builds. It also shows up in `bash tests/scheme/run-all.sh` as an intermittent `FAIL` on this file (`run-all.sh: line 28: ... Abort trap: 6`).

The test spawns two SRFI-18 threads that deep-copy a 100k-element list across the thread boundary — at `thread-start!` (closure upvalue → child) and at `thread-join!` (result → parent):

```scheme
(define (make-long-list n)
  (let loop ((i 0) (acc '()))
    (if (= i n) acc (loop (+ i 1) (cons i acc)))))
(define t1 (let ((lst (make-long-list 100000))) ...))
;; ... thread-join! both, expect 100000 each
```

## Observed crash signatures

The crash location is non-deterministic, which is the key evidence for heap corruption:

**Signature A — SIGSEGV in `malloc` while the main thread compiles the next top-level form**, with child-thread pthread-cond frames also on the stack:

```
Segmentation fault at address 0xb7da9c1608
???:?:?: 0x186ea2ea0 in ??? (/usr/lib/system/libsystem_malloc.dylib)
std/heap.zig:232:37: in alloc (kaappi)                 const ptr = c.malloc(actual_len) orelse return null;
std/mem/Allocator.zig:142:26: in rawAlloc (kaappi)
src/ir.zig:483:24: in lowerBegin (kaappi)              return ir.makeBegin(buf[0..count]);
src/ir.zig:363:35: in lowerWithMacros (kaappi)
src/ir.zig:460:28: in lowerIf (kaappi)
src/compiler.zig:348:46: in compile (kaappi)           var root = try ir_mod.lowerWithMacros(&ir, expr_root, &self.macros);
src/compiler.zig:1570: in compileExpressionWithMacrosAt (kaappi)
src/main.zig:824: in runFile (kaappi)
...
???:?:?: 0x187091c57 in __pthread_cond_wait (/usr/lib/system/libsystem_pthread.dylib)
???:?:?: 0x18708cc1b in _pthread_cond_broadcast (/usr/lib/system/libsystem_pthread.dylib)
```

**Signature B — `Abort trap: 6` inside `_pthread_cond_broadcast`** (seen via `run-all.sh`).

The main thread faulting inside `malloc` (from unrelated IR-compilation allocation of the *next* top-level form) while a child thread is active strongly implies the child's per-thread GC / `deepCopy` path is corrupting the shared C-heap metadata (use-after-free / double-free), which then kills whichever thread next calls into the allocator.

## Likely area

- `src/memory.zig` — `GC.deepCopy` / `deepCopyValue`, `GC.initForThread`, per-thread heap teardown after `thread-join!`
- `src/gc_deep_copy.zig`
- `src/primitives_srfi18.zig` — `child_resources`, thread start/join lifecycle
- Shared allocator use across threads while a child GC frees its heap

Related prior threading work: #797, #745/#750 (symbol-table race), #801/#952 (the deep-copy stack-overflow fix that added this test).

## Environment

- macOS (Darwin 25.5.0), arm64 (Apple Silicon)
- Zig 0.16.0
- Default build (`zig build`, ReleaseSafe)
- kaappi `main` @ `0f9d56c`

## Suggested next steps

- Run under a thread sanitizer / `-Doptimize=Debug` with malloc guard (macOS `MallocScribble`/`MallocGuardEdges`, or `libgmalloc`) to catch the corrupting write.
- Audit lifetimes: ensure the child GC heap is not freed while the parent still references (or is mid-copy of) values, and that no allocator state is shared unsynchronized across the start/join boundary.

Found while validating an unrelated PR (#957); confirmed on pristine `main`.

baijum:
## Summary

Fixes the flaky (30–40%) SIGSEGV/SIGABRT in `tests/scheme/smoke/deep-copy-list-801.scm` (#958). Running the test under guard malloc (`libgmalloc`) made the crash 100% deterministic and exposed two ways an SRFI-18 child OS thread corrupted parent-owned memory:

### 1. Named let wrote child-heap closures into the shared globals map

Named let bound its loop procedure to a gensym'd **global** (`__nlet_N_name`) via `define_global`, re-executed on **every call**. When a child thread ran a parent-heap function containing a named let (here: `make-long-list`, looked up via the shared globals map), it wrote a **child-heap closure** into the shared map. The child GC never roots shared globals (`owns_globals=false`), so that closure was freed by the child's next collection or by heap teardown at `thread-join!` — leaving a dangling global that the parent's next mark phase dereferenced. Corrupted-mark traversal then scribbled over freed pages, which is why the crash surfaced later in `malloc` on the main thread.

**Fix:** named let now binds the loop procedure to a **boxed local** (single-binding letrec, same machinery as `compileLetrec`), so executing it writes nothing shared — this also stops polluting the global namespace. The `self_tail_call` fast path is taught that a `__nlet_` name match is always a self-reference (the name now resolves as an upvalue, which previously disqualified it); loop microbenchmarks match the previous build (~2.0s for 20M iterations, before and after).

### 2. Child GC collections marked parent-heap objects

A child's frames/registers legitimately hold parent-heap pointers (the parent closure it executes, interned symbols, native fns). Each child collection **marked and traced** those parent objects. Stale/concurrent foreign mark bits corrupt the parent's own cycle: the parent sees a spuriously "already marked" object, skips tracing its children, and sweeps live descendants.

**Fix:** every heap object records its owning GC (`Object.owner: u32`, stamped by `trackObject` from a never-reused `GC.id`; fits in existing `Object` padding, so zero size cost). Marking, the write barrier, and remembered-set pruning skip objects owned by another GC — foreign liveness is the owner's job (shared globals are marked by the parent, interned symbols are never swept, the thunk is extra-rooted until join).

### Hardening

`thread-start!` now extra-roots the **fiber** as well as its thunk (removed at `thread-join!`): the child writes `fiber.status` and reads `fiber.terminated` for its whole run, so the fiber must survive even if the program drops its last reference mid-run.

## Verification

| Check | Before | After |
|---|---|---|
| `deep-copy-list-801.scm` under libgmalloc | 6/6 crash | 0/6 crash |
| `deep-copy-list-801.scm` plain ×30 | ~30–40% crash | 0/30 crash |
| `zig build test` (incl. 3 new regression tests) | — | pass |
| `run-all.sh` (252 files + 1395 R7RS) | — | 0 fail |
| Thread smoke tests + repro under `-Dgc-threshold=1` | — | pass |
| Named-let loop microbench (20M iters + non-tail rec) | ~2.05s | ~2.04s |

New regression tests in `src/tests_srfi18.zig`: direct cross-GC marking invariant, no stale marks on the parent heap after a child runs a parent closure, and no child-owned values in the shared globals map after a child runs a named let.

## Remaining known gap (pre-existing, not addressed here)

A child thread *reading* the shared globals map concurrently with the parent *rehashing* it (parent defines enough new globals mid-child-run to grow the map) is still unsynchronized. That's a narrower, read-side race and orthogonal to the write-path corruption fixed here; noting it for a follow-up.

Fixes #958

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -430,6 +430,9 @@ independently and the child heap is freed after `thread-join!`.
 - `VM.owns_globals` prevents child VM from freeing shared maps on deinit
 - `symbol_mutex` (spinlock) protects concurrent symbol interning (`src/memory.zig`)
 - Child GC/VM references stored in global `child_resources` map (`src/primitives_srfi18.zig`)
+- Every heap object records its owning GC (`Object.owner` / `GC.id`); marking skips
+  objects owned by another GC, so a child's collections never write mark bits on
+  parent-heap objects reached via shared globals (`src/gc_collect.zig`, #958)
 
 ## Claude Code harness
 
--- a/src/compiler_bindings.zig
+++ b/src/compiler_bindings.zig
@@ -258,67 +258,51 @@ pub fn compileNamedLet(self: *Compiler, args: Value, dst: u16, is_tail: bool) Co
         formals = self.gc.allocPair(var_names[i], formals) catch return CompileError.OutOfMemory;
     }
 
-    // Named let uses a global for the loop procedure. Use a unique gensym'd
-    // name to prevent collisions when multiple named lets use the same name.
+    // Bind the loop procedure to a boxed local (single-binding letrec) so the
+    // loop lambda captures itself as an upvalue. It used to be bound to a
+    // gensym'd global (__nlet_N_name), but executing that define_global from
+    // an SRFI-18 child thread wrote a child-heap closure into the shared
+    // globals map, leaving a dangling global after the child heap was freed
+    // (issue #958). A local also stops polluting the global namespace. The
+    // gensym rename is kept so nested named lets reusing a name stay distinct.
     const unique_sym = try makeUniqueLoopName(self.gc, types.symbolName(loop_name));
     const renamed_body = try renameInBody(self.gc, body, types.symbolName(loop_name), unique_sym);
     const renamed_lambda_args = self.gc.allocPair(formals, renamed_body) catch return CompileError.OutOfMemory;
 
-    // Use a fresh register for the closure to avoid overwriting live locals
-    // (e.g., when dst=0 and a parameter is also at register 0).
-    const loop_reg = try self.allocReg();
+    self.beginScope();
 
+    // Mirror compileLetrec: void-init the slot, bind, box, then assign.
+    const loop_reg = try self.allocReg();
     try self.emitOp(.load_void);
     try self.emitU16(loop_reg);
-    const name_sym_idx = try self.addConstant(unique_sym);
-    try self.emitOp(.define_global);
-    try self.emitU16(name_sym_idx);
+    try self.addLocal(types.symbolName(unique_sym), loop_reg);
+    try self.markLocalBoxedBySlot(loop_reg);
+
+    const tmp_reg = try self.allocReg();
+    try self.compileLambda(renamed_lambda_args, tmp_reg, types.symbolName(unique_sym));
+    try self.emitOp(.set_box_local);
     try self.emitU16(loop_reg);
+    try self.emitU16(tmp_reg);
+    self.freeReg(); // tmp_reg
 
-    try self.compileLambda(renamed_lambda_args, loop_reg, types.symbolName(unique_sym));
-
-    try self.emitOp(.define_global);
-    try self.emitU16(name_sym_idx);
-    try self.emitU16(loop_reg);
-
-    // Compile the initial call
-    const call_base = try self.allocReg();
-    self.freeReg();
-
-    if (call_base != loop_reg) {
-        try self.emitOp(.move);
-        try self.emitU16(call_base);
-        try self.emitU16(loop_reg);
+    // Compile the initial call (UNIQ init ...) through the normal call path:
+    // it reads the boxed local and handles tail position.
+    var call_expr: Value = types.NIL;
+    {
+        self.gc.no_collect += 1;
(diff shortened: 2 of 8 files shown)


baijum:
The `test (macos-latest, ReleaseSafe)` failure is pre-existing runner breakage, not this PR: the same thottam test (`checkoutVersion resolves a pinned tag as a ref` / `pathspec '--end-of-options'`) fails identically on `main` (run 28658342026, a compiler-only commit) and passed on main 4 minutes earlier — a macos-latest runner image/git rollout. Filed as #969. All other jobs (5× Linux incl. Debug + ReleaseFast, riscv64, wasm, benchmark, format) pass. Rerunning the macOS job in case it lands on a runner with the old image.

Issue #780, opened 2026-07-02
thottam: version-pinned install always fails — git checkout '--' treats ref as pathspec (regression from #736 fix)


baijum:
## What's wrong

Every form of version-pinned install is broken in v0.11.1:

- `thottam install pkg@v1.0.0` (pinned tag)
- `thottam install 'pkg@">=1.0.0"'` (semver constraint — resolution succeeds, checkout then fails)
- `thottam --locked install pkg` (lockfile SHA)

All fail with `Failed to checkout version`.

## Affected code

`src/thottam.zig:819` (in `doInstall`):

```zig
runGit(allocator, &.{ "-C", pkg_dir, "checkout", "--quiet", "--", v }) catch {
```

## Root cause

The fix for #736 (commit eb38b0f, shipped in v0.11.1) added a `--` separator before the version argument, intending to prevent option injection. But in `git checkout`, arguments **after** `--` are unambiguously **pathspecs**, not refnames:

```
git checkout [<tree-ish>] [--] <pathspec>...
```

So `git checkout --quiet -- v1.0.0` means "restore the file named `v1.0.0` from the index", which fails for any repo that has no file by that name:

```
error: pathspec 'v1.0.0' did not match any file(s) known to git
```

(And if the repo *did* contain a file named `v1.0.0`, git would silently restore that file and never switch versions — arguably worse.)

## Minimal reproduction

```sh
mkdir -p /tmp/repo && cd /tmp/repo && git init -q .
git commit -q --allow-empty -m one && git tag v1.0.0
git commit -q --allow-empty -m two && git tag v1.1.0

export KAAPPI_HOME=/tmp/kaappi-test-home   # isolate from real ~/.kaappi
thottam install 'pkga@v1.0.0::/tmp/repo'
```

Actual:
```
Installing pkga@v1.0.0...
  Cloning /tmp/repo...
  Checking out v1.0.0...
error: pathspec 'v1.0.0' did not match any file(s) known to git
  Failed to checkout version
```
Exit code 1; nothing installed. Also reproduced for `pkga@">=1.0.0"` (prints `Resolved ">=1.0.0" -> v1.1.0` then fails identically) and for `--locked` installs (the locked SHA is treated as a pathspec the same way).

Expected: checkout of the tag/SHA succeeds and installation completes (as it did in v0.11.0, which used `checkout --quiet v` without `--`).

## Suggested fix

[...]

PR #971, opened 2026-07-03
Share globals map by pointer with SRFI-18 children, lock rehashes (#958 follow-up)

baijum:
Follow-up to #968, which fixed the child WRITE path for #958 and documented the remaining globals **read race** as a known gap. This closes that gap.

## The bug (reproduces on current main)

`VM.initForThread` shared the parent's globals `StringHashMap` **by struct copy**. Two failure modes:

1. **Use-after-free:** a parent-side `put()` that triggers a rehash reallocates the bucket array and frees the old one — which the child's copied header still points to. Every subsequent child global lookup reads freed memory.
2. **Permanent staleness:** even when the freed memory happens to survive, the child's copied header never tracks the parent's new array, so the child never sees bindings defined (or `set!`) after the first rehash.

The new regression test `tests/scheme/srfi/srfi18-globals-rehash.scm` (child polls a global while the parent defines 3000 fresh globals, then flips a flag) **bus-errors inside `hash_map.getIndex` on current main**, verified against a clean `origin/main` build:

```
Bus error at address 0x...
std/hash_map.zig:984:39: in getIndex (kaappi)
src/vm_dispatch.zig: in runUntil (kaappi)
src/primitives_srfi18.zig:301: in threadEntryFn (kaappi)
```

## The fix

Share by pointer + a reader-writer lock on the shared path only:

- `VM.globals` is now a heap-allocated `*StringHashMap(Value)` shared by pointer with child VMs, paired with a shared `VM.globals_lock`. A parent-side rehash is therefore *seen* by children instead of leaving them on a freed array.
- **Locking protocol** (documented on the field):
  - structural mutation (put/remove — anything that can rehash) takes the **exclusive** lock, on any thread;
  - child threads (`owns_globals == false`) take the **shared** lock for every map read;
  - **the owner thread reads lock-free** — it is the only thread expected to structurally mutate the map.
- Zig 0.16 has no blocking `std.Thread.RwLock` (`std.Io.RwLock` needs an `Io`), so `GlobalsRwLock` is a small portable writer-preferring spinlock in the same style as `memory.zig`'s `symbol_mutex`.
- Locked sites beyond the obvious dispatch handlers (`get/set/define/call/tail_call_global`, `findSimilarName`):
  - the **compile-time sentinel puts** in `compileBody`/`compileLetBody` prescans and `compiler.zig`'s macro-expansion temp-globals dance — the parent compiles top-level forms interleaved with execution, so these rehash-capable puts run while children execute. `vm.acquireGlobalsWrite(map)` engages only when the map is the current thread's shared globals (library envs pass through unlocked);
  - child-side eval-compile reads (hygiene in `expander.zig`, constant folding in `compiler_passthrough.zig`/`ir.zig`) via `acquireGlobalsRead`;
  - runtime read paths reachable from child threads in `primitives_hashtable.zig` / `primitives_string_ext.zig`;
  - import (`vm_library.zig`), `define-record-type` (`vm_records.zig`), `define-values` (`vm_eval.zig`), REPL `_`.
- `set_global` keeps the `getPtr` **and** the store through it inside one locked region — a rehash between them would leave the pointer dangling on a child.

## Parent fast-path benchmark

`get_global` is hot, so the owner path takes no lock — only a pointer deref plus a predicted `owns_globals` branch. Worst-case microbenchmarks (ReleaseSafe, macOS ARM, 50M uncached global reads / 20M global `set!`s, medians of repeated runs):

| bench | main | this PR | delta |
|---|---|---|---|
| uncached global read loop | 3.78s | 3.88s | ~+2.5% |
| global `set!` loop | 2.17s | 2.25s | ~+3% |

The closure/native inline-cache path (`func.global_cache`) is untouched and takes no lock.

## Tests

- `tests/scheme/srfi/srfi18-globals-rehash.scm` — deterministic end-to-end regression: bus-errors on main, passes here (the child observes a post-rehash `set!`).
- `src/tests_srfi18.zig` — single-threaded mechanical test that the child's view tracks the parent map across rehashes (late defines, in-place updates, mid-burst defines).
- Full suites green: `zig build test`, all 260 Scheme files + 1395 R7RS tests (0 timeouts), `zig build wasm`.

## Remaining known gaps (documented on the `globals` field)

- A child that defines globals via `eval` takes the exclusive lock (protecting other children), but the owner's lock-free reads can still race such writes — same class of limitation #968 documented for child writes.
- The `macros` and `libraries` maps are still shared by struct copy and have the analogous (much rarer) staleness hazard — tracked as a follow-up.

### DIFF
--- a/src/bench.zig
+++ b/src/bench.zig
@@ -35,7 +35,7 @@ fn measure(allocator: std.mem.Allocator, prim: []const u8, depth: u32, iters: u3
     defer vm.deinit();
     try primitives.registerAll(&vm);
     primitives.setGCInstance(&gc);
-    try library.registerStandardLibraries(&vm.libraries, &vm.globals);
+    try library.registerStandardLibraries(&vm.libraries, vm.globals);
 
     // at-depth builds `depth` real (non-tail) frames to elevate the register
     // base; cap runs a tail loop performing `iters` escaping captures.
--- a/src/compiler.zig
+++ b/src/compiler.zig
@@ -6,6 +6,7 @@ const forms = @import("compiler_forms.zig");
 const advanced = @import("compiler_advanced.zig");
 const passthrough = @import("compiler_passthrough.zig");
 const ir_mod = @import("ir.zig");
+const vm_mod = @import("vm.zig");
 const Value = types.Value;
 const OpCode = types.OpCode;
 
@@ -1208,6 +1209,13 @@ pub const Compiler = struct {
                 var global_free_names: [64][]const u8 = undefined;
                 var global_free_count: usize = 0;
                 if (self.globals) |g| {
+                    // The sentinel puts below structurally mutate the globals
+                    // map when g is the VM's shared one — exclude SRFI-18
+                    // child-thread readers for the whole dance (#958). glk is
+                    // non-null exactly when g is the current thread's shared
+                    // globals map.
+                    const glk = vm_mod.acquireGlobalsWrite(g);
+                    defer vm_mod.releaseGlobalsWrite(glk);
                     for (tx.captured_locals) |cap| {
                         if (!g.contains(cap.name) and temp_global_count < 128) {
                             temp_globals[temp_global_count] = .{ .name = cap.name, .old_val = null, .was_present = false };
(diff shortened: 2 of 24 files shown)


baijum:
Note on CI: the `test (macos-latest, ReleaseSafe)` failure is pre-existing on main (identical failure on the `992bb6f` and `15ddbc1` main runs), not introduced here. `thottam.test.checkoutVersion resolves a pinned tag as a ref (issue #780)` fails because the macos-latest runner's `/usr/bin/git` doesn't recognize `--end-of-options` in `git checkout` and treats it as a pathspec (support landed in git 2.28; local Apple Git 2.50.1 is fine — this is a runner-image git). Portable fix, verified locally: use a trailing `--` instead (`git checkout --quiet <v> --`), which forces `<v>` to parse as a revision on all git versions and still fails cleanly (`fatal: invalid reference`) on bad refs.

Issue #1147, opened 2026-07-05
[R7RS 6.12] eval: define/set! into an immutable (environment ...) silently succeeds instead of signaling an error


baijum:
**Category:** R7RS conformance
**Spec reference:** R7RS 6.12, p. 55 — "The bindings of the environment represented by the specifier are immutable, as is the environment itself." And for eval: "If it is a definition, the specified identifier(s) are defined in the specified environment, provided the environment is not immutable." Spec example: `(eval '(define foo 32) (environment '(scheme base)))` => **error is signaled**.
**Severity:** edge-case (isolation is correct; the required error signal is missing)

**Minimal reproduction:**
```scheme
(import (scheme base) (scheme eval))
(eval '(define foo 32) (environment '(scheme base)))
;; expected: an error is signaled (immutable environment)
;; actual:   silently succeeds, returns unspecified

(eval '(set! car 42) (environment '(scheme base)))
;; expected: error (immutable binding)
;; actual:   silently succeeds
```

**What works correctly:** the environments ARE properly isolated — the define does not leak into the global environment, `set! car` does not affect the real `car`, and import sets are respected (`car` is not visible in `(environment '(only (scheme base) +))`). Only the required error signal on mutation attempts is missing: the environment behaves as a mutable private copy instead of an immutable view.

**Verified:** fresh ReleaseSafe build at b2317e8 (v0.12.0); deterministic.
**Source location:** `src/primitives_r7rs.zig` `environmentFn` / eval environment handling (~line 177).
**Disabled test:** `tests/scheme/compliance/r7rs-control-io-gaps.scm` (marked `;; FAIL: #this`)

**Found by:** Systematic audit, Phase 1D

PR #1146, opened 2026-07-05
Phase 1C: R7RS 6.6-6.9 conformance gap tests

baijum:
Phase 1C of the audit campaign (#1137): spec-driven gap analysis of R7RS sections 6.6 (characters), 6.7 (strings), 6.8 (vectors), 6.9 (bytevectors).

## Found

**#1145** — `char-upper-case?`/`char-lower-case?`/`char-alphabetic?` derive classification from simple case mappings instead of Unicode properties. Titlecase Dž (U+01C5) reports as both upper and lower case, ß reports as not lowercase, ª/º report as non-alphabetic. Root cause is explicit in `src/primitives_char.zig:105-119`.

## What conforms (notable)

- Full Unicode string casing: `(string-upcase "Straße")` => `"STRASSE"` with length change, foldcase-based `-ci` comparisons
- `digit-value` for Arabic-Indic and Gujarati digits
- Overlap guarantees for `string-copy!`/`vector-copy!`/`bytevector-copy!`
- UTF-8 conversions: `string->utf8` indexes by codepoint, `utf8->string` by byte

New file: `tests/scheme/compliance/r7rs-chars-strings-gaps.scm` (43 passing assertions + 4 disabled `;; FAIL: #1145` regression tests).

Stacked on #1144 (Phase 1B); retarget after it merges.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1137, opened 2026-07-05
Tracking: R7RS conformance and SRFI audit


baijum:
Tracking issue for the systematic R7RS conformance and SRFI audit campaign.
Strategy document: `docs/audit-strategy.md` (single source of truth for the
session protocol and progress tracker). Each audit session checks off its unit
in the doc's Progress tracker and posts a one-line summary here.

## Baseline (Phase 0, 2026-07-05)

Fresh ReleaseSafe build at b2317e8, `kaappi --version` = v0.12.0.

### R7RS suite (`tests/scheme/r7rs/r7rs-tests.scm`) — all sections green

| Section | Pass | Fail |
|---------|-----:|-----:|
| 4.1 Primitive expression types | 27 | 0 |
| 4.2 Derived expression types | 74 | 0 |
| 4.3 Macros | 25 | 0 |
| 5 Program structure | 15 | 0 |
| 6.1 Equivalence predicates | 25 | 0 |
| 6.2 Numbers | 211 | 0 |
| 6.3 Booleans | 18 | 0 |
| 6.4 Lists | 65 | 0 |
| 6.5 Symbols | 17 | 0 |
| 6.6 Characters | 79 | 0 |
| 6.7 Strings | 130 | 0 |
| 6.8 Vectors | 43 | 0 |
| 6.9 Bytevectors | 39 | 0 |
| 6.10 Control features | 34 | 0 |
| 6.11 Exceptions | 30 | 0 |
| 6.12 Environments and evaluation | 4 | 0 |
| 6.13 I/O — Read syntax | 93 | 0 |
| 6.13 I/O — Numeric syntax | 220 | 0 |
| 6.14 System interface | 13 | 0 |

### Other suites

- Zig unit tests (`zig build test`): PASS
- `run-all.sh` (smoke, compliance, continuations, hygiene, srfi, ffi, audit, errors, r7rs): **1702 pass, 0 fail** (0 timeouts)
- All 35 `tests/scheme/srfi/*.scm` files run individually with fail-count parsing: **0 failures**

**No pre-existing failures — zero issues filed from the baseline.** All
findings from here on come from *new* tests written in Phases 1–4.

Baseline script: `tests/scheme/audit-baseline.sh` (note: the doc's original
draft used GNU `timeout`, absent on stock macOS; the committed script falls
back to `gtimeout`/perl-alarm).

## Progress tracker (mirrors docs/audit-strategy.md)

**Phase 0 — Baseline**
[...]

PR #1148, opened 2026-07-05
Phase 1D: R7RS 6.10-6.14 conformance gap tests

baijum:
Phase 1D of the audit campaign (#1137), completing Phase 1: spec-driven gap analysis of R7RS sections 6.10 (control), 6.11 (exceptions), 6.12 (environments/eval), 6.13 (I/O), 6.14 (system interface).

## Found

**#1147** — `eval` of a `define` or `set!` into an immutable `(environment ...)` silently succeeds instead of signaling an error (the spec's own example requires the error). Isolation itself is correct: nothing leaks, import sets are respected.

## What conforms (notable)

- `write` on cyclic structures emits datum labels (`#0=(1 2 . #0#)`) and terminates
- `read-line` handles LF, CR, and CRLF line endings
- `raise-continuable` handler-value semantics (spec example => 65)
- `file-error?`/`read-error?` discrimination, port lifecycle predicates, bytevector ports
- shortest-list termination for `map`/`vector-map`/`string-map`

New file: `tests/scheme/compliance/r7rs-control-io-gaps.scm` (30 passing assertions + 2 disabled `;; FAIL: #1147` regression tests). Deep continuation interactions are deliberately left to Phase 4B.

Stacked on #1146 (Phase 1C); retarget after it merges.

### DIFF
--- a/docs/audit-strategy.md
+++ b/docs/audit-strategy.md
@@ -115,7 +115,7 @@ and issue numbers, e.g. `[x] ... (2026-07-06, #1101–#1105)`.
 
 **Phase 1 — R7RS spec gap analysis** (independent; run in parallel)
 - [x] 1A: Expressions & syntax (4.1–4.3) + Libraries (5.6–5.7) (2026-07-05, #1139–#1142; 34 new gap tests, 3 disabled pending fixes; libraries 5.6 fully green)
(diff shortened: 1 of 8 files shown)


baijum:
Superseded by a cherry-picked replacement (tracker-line merge conflict from the stacked chain).

Issue #1243, opened 2026-07-05
[R7RS 4.3.2] Doubled ellipsis in template does not flatten: ((x ...) ...) matched input expanded with (x ... ...) produces garbage containing literal '...'


baijum:
**Category:** Expander (syntax-rules template instantiation)
**Spec reference:** R7RS 4.3.2 (p. 24): "Pattern variables that occur in subpatterns followed by one or more instances of the identifier <ellipsis> are allowed only in subtemplates that are followed by as many instances of <ellipsis>. They are replaced in the output by all of the elements they match in the input, distributed as indicated."
**Severity:** wrong-result (silent garbage expansion, includes a literal `...` symbol in output)

**Minimal reproduction (fresh build):**
```scheme
(import (scheme base))
(define-syntax flatten2
  (syntax-rules ()
    ((_ ((x ...) ...)) '(x ... ...))))
(flatten2 ((1 2) (3 4) (5)))
;; expected: (1 2 3 4 5)
;; actual:   (() () () ...)     — three empty lists plus a literal ... symbol
```
`x` is matched at ellipsis depth 2 and used in a template followed by two consecutive ellipses — the canonical flattening construct. Single-level distribution works fine (`'((a (b ...)) ...)` expands correctly), and the `(... ...)` escape + be-like-begin example from p. 24 also works, so the bug is specifically the "followed by MORE than one ellipsis" splice path in template instantiation (src/expander.zig).

**Relation to existing issues:** #682 covers missing validation for ILLEGAL depth mismatches; this input is legal and must expand to the flattened sequence. #308 covers the (... template) escape's hygiene.

**Disabled tests:** tests/scheme/compliance/r7rs-hygiene-gaps.scm (marked ;; FAIL: #this)
**Found by:** Systematic audit, Phase 4C

Issue #1242, opened 2026-07-05
[R7RS 4.2.5] force caps promise chains at 100,000 iterations and reports legitimate long delay-force chains as circular


baijum:
**Category:** Primitives (src/primitives_lazy.zig)
**Spec reference:** R7RS 4.2.5: delay-force exists so that "iterative lazy algorithms ... may be executed in bounded space" — the bound is on SPACE, not on the number of forcing steps. A conforming force must walk a delay-force chain of any length.
**Severity:** wrong-result (false type error on valid programs)

**Minimal reproduction (fresh build):**
```scheme
(import (scheme base) (scheme lazy))
(define (chain n) (if (= n 0) (delay 'end) (delay-force (chain (- n 1)))))
(force (chain 99999))    ;; => end
(force (chain 100000))   ;; error: type error in 'force': expected
                         ;;   non-circular promise chain, got #<promise>
```

**Root cause:** src/primitives_lazy.zig:43-44 — the forcing trampoline runs `while (iterations < 100000)` and falls through to `typeError("force", "non-circular promise chain", ...)` (line 102) when the budget is exhausted. The cap is a heuristic cycle detector, but it cannot distinguish a genuine cycle like `(define p (delay (force p)))` from a long legitimate chain (SRFI-41 streams, SRFI-45-style iterative algorithms routinely exceed fixed budgets). Real cycle detection needs pointer-based detection (Floyd/Brent on the chain) or simply trusting the chain (a cycle is a programmer error whose non-termination is permitted by the spec).

**Note:** true self-referential force already has a separate failure mode (native re-entrancy panic, see #1191), so the cap does not even reliably convert cycles into errors.

**Disabled tests:** tests/scheme/compliance/r7rs-thin-forms-gaps.scm (marked ;; FAIL: #this)
**Found by:** Systematic audit, Phase 4A

Issue #1241, opened 2026-07-05
[R7RS 3.5] let*-values body is not a tail context — tail loops through it re-enter the VM natively and panic at depth ~1024


baijum:
**Category:** Compiler / R7RS conformance (proper tail recursion)
**Spec reference:** R7RS 3.5 (p. 11) lists `(let*-values (<mv binding spec>*) <tail body>)` among the tail contexts. A properly tail-recursive implementation must support an unbounded number of active tail calls through it.
**Severity:** conformance violation + uncatchable crash (#1191's GC-root-stack panic, not a Scheme error)

**Minimal reproduction (fresh build; panics "GC root stack overflow (1024)"):**
```scheme
(define (lp n)
  (if (= n 0) 'done
      (let*-values (((m) (values (- n 1)))) (lp m))))
(lp 5000)
```
A single binding clause is enough. The equivalent `let-values` loop runs at 1,000,000 iterations, so the sequential-binding path is what re-enters natively:
```scheme
(define (lp n)
  (if (= n 0) 'done
      (let-values (((m) (values (- n 1)))) (lp m))))
(lp 1000000)   ;; => done
```

**Root cause:** let*-values compilation (src/compiler_bindings.zig) evaluates its clauses through a native call chain (callWithArgs/callReentrant) that stacks a Zig frame + GC roots per invocation, so the body call can never unwind the native frame even in tail position. let-values compiles to pure bytecode and stays flat.

**Relation to existing issues:** crash mechanism is #1191; the required-tail-call gaps for call-with-values/call/cc/eval are #1240. This one is a syntactic tail context from the 3.5 list.

**Disabled tests:** tests/scheme/compliance/r7rs-tail-position-gaps.scm (marked ;; FAIL: #this)
**Found by:** Systematic audit, Phase 4A

baijum:
**Verified fixed at f13e806** (current main, ReleaseSafe build) — fixed by a91e908 (epic #1245 tranche).

- Direct repro: `let*-values` tail loop (`(let*-values (((m) (values (- n 1)))) (loop m))`) at **N=1,000,000** returns `done` — was a native re-entry panic at depth ~1024.
- Suite: the "let*-values body tail" test (N=25000, above the 16k–20k register-exhaustion threshold for non-tail compilation) is enabled in `tests/scheme/compliance/r7rs-tail-procedures-gaps.scm` and passing in `run-all.sh` (1775 pass / 0 fail).

Closing.

Issue #1240, opened 2026-07-05
[R7RS 3.5] call-with-values consumer, call/cc receiver, and eval are not tail-called — required tail calls re-enter the VM natively and panic at depth ~1024


baijum:
**Category:** VM / R7RS conformance (proper tail recursion)
**Spec reference:** R7RS 3.5 (p. 12): "Certain procedures defined in this report are also required to perform tail calls. The first argument passed to apply and to call-with-current-continuation, and the second argument passed to call-with-values, must be called via a tail call. Similarly, eval must evaluate its first argument as if it were in tail position within the eval procedure."
**Severity:** conformance violation + uncatchable crash (the failure mode is #1191's GC-root-stack panic, not a Scheme error)

**Minimal reproductions (fresh build, each panics "GC root stack overflow (1024)"):**
```scheme
(define (lp n) (if (= n 0) 'done (call-with-values (lambda () (values (- n 1))) lp)))
(lp 5000)

(define (lp2 n) (if (= n 0) 'done (call/cc (lambda (k) (lp2 (- n 1))))))
(lp2 5000)

(import (scheme eval) (scheme repl))
(define (lp3 n) (if (= n 0) 'done (eval (list 'lp3 (- n 1)) (interaction-environment))))
(lp3 5000)
```
`apply` is fine: its tail position compiles to the bounded tail_apply opcode (compiler.zig), and (define (lp n) (if (= n 0) 'done (apply lp (list (- n 1))))) runs at any depth.

**Root cause:** call-with-values, call/cc's receiver invocation, and eval all call back into the VM via `callWithArgs`/`callReentrant` (src/vm_calls.zig), pushing a native Zig frame + GC roots per iteration regardless of whether the primitive itself was called from a tail position. R7RS 3.5 requires these three to consume no stack in tail position — same treatment apply got via tail_apply.

**Relation to existing issues:** #1191 covers the crash MECHANISM (deep native re-entrancy panics instead of raising); this issue is the conformance requirement that these specific positions must not re-enter natively at all. Fixing #1191 alone (graceful error) would still leave 3.5 violated.

**Disabled tests:** tests/scheme/compliance/r7rs-tail-position-gaps.scm (marked ;; FAIL: #this)
**Found by:** Systematic audit, Phase 4A

baijum:
[...]

PR #1244, opened 2026-07-05
Add compiler/VM edge-case tests (audit Phases 4A-4C)

baijum:
## Summary

Phase 4 of the audit campaign (tracking: see #1137): compiler/VM edge cases in three units, batched into five compliance files. **99 passing assertions; 8 disabled** with `;; FAIL: #NNNN` markers.

| Unit | File | Coverage |
|------|------|----------|
| 4A | `r7rs-tail-position-gaps.scm` | Every R7RS §3.5 tail context driven 1e5–1e6 iterations deep (a non-tail implementation dies at the 32k frame cap) |
| 4A | `r7rs-thin-forms-gaps.scm` | delay/delay-force/force, parameterize, define-values, let*-values scoping, letrec* ordering, letrec-syntax |
| 4B | `r7rs-continuation-gaps.scm` | call/cc × dynamic-wind/guard/parameterize/values, incl. the R7RS 6.10 connect/talk/disconnect re-entry example and multi-shot segments |
| 4C | `r7rs-hygiene-gaps.scm` | Nested ellipses, `(... ...)` escape, macro-generating macros, custom ellipsis, vector/tail patterns, hygiene shadowing incl. the p.24 `=>`-shadowing example |
| 4C | `r7rs-import-macro-gaps.scm` | Macro exports through `(export (rename ...))` + prefix/only/rename/except and nested sets; cond-expand library declarations |

## New issues

- **see #1240** — `call-with-values` consumer, `call/cc` receiver, and `eval` are not tail-called (§3.5 requires it); each panics uncatchably at native re-entry depth ~1024 (mechanism is #1191's). `apply` is fine via tail_apply.
- **see #1241** — `let*-values` body is not a tail context (one clause suffices to crash; `let-values` runs at 1e6).
- **see #1242** — `force` caps promise chains at 100,000 iterations and mislabels longer legitimate `delay-force` chains as circular.
- **see #1243** — doubled-ellipsis templates `(x ... ...)` expand to garbage containing a literal `...` instead of flattening.

## Conforms (highlights)

All syntactic §3.5 tail contexts including `cond`/`case` `=>` receiver calls, `do` result sequences, `case-lambda` clause bodies, and `let-syntax`/`letrec-syntax` bodies; the R7RS 6.10 dynamic-wind re-entry example produces the exact spec answer; parameterize values are captured/restored across continuation re-entry; guard `=>`/else/re-raise, `raise-continuable`, and handler-return secondary exceptions; `be-like-begin` and the `(let ((=> #f)) (cond (#t => 'ok)))` hygiene example; macro exports survive every import-set combinator; circular imports are cleanly detected with a nonzero exit.

4B state is deliberately kept in globals — `set!` of register-allocated locals is rolled back on continuation re-entry (see #1168), which would turn the re-entry tests into infinite loops.

## Test results

Full suite on this branch: **366 Scheme files pass, 0 fail; R7RS suite 1395 pass, 0 fail**

Tracker: 4A/4B/4C checked — only Phase 5 (synthesis) remains.


baijum:
Debug CI timeout fixed in e3c32c7: the tail-position file ran 19M loop iterations, which Debug (~500x slower on these workloads) couldn't finish inside run-all.sh's 60s per-file cap. A non-tail implementation actually fails between 16k and 20k active frames (register exhaustion — probed empirically), so N=25000 proves TCO just as strictly at 1/40th the work. The allocation-heavy loops (let-values costs ~0.6ms/iter in Debug) split into a companion file r7rs-tail-procedures-gaps.scm. Measured 16.5s + 20.3s under local Debug ≈ 5–6s each on the CI runner. Test counts unchanged: 22 passing, 6 disabled.

Issue #1191, opened 2026-07-05
[VM] Deeply nested native re-entrancy panics: 'GC root stack overflow (1024)' instead of raising a catchable error


baijum:
**Category:** VM
**Spec reference:** R7RS §4.2.5 permits a self-referential `(force p)` to signal an error or loop, but a process abort is neither. Nested `map` recursion at depth 2000 is ordinary conforming code.
**Severity:** crash

**Minimal reproduction (two independent triggers, same panic):**

```scheme
;; 1. direct re-entrant force
(define p (delay (force p)))
(force p)
;; => thread panic: GC root stack overflow (1024), abort

;; 2. nested native higher-order calls
(define (deep n) (if (= n 0) 1 (car (map (lambda (x) (deep (- n 1))) '(1)))))
(deep 2000)
;; => thread panic: GC root stack overflow (1024), abort
```

**Expected:** a catchable Scheme error (like the "too many frames" path), or graceful growth of the root stack.
**Actual:** `@panic("GC root stack overflow (1024)")` in `pushRoot` — the whole process aborts; `guard` cannot intercept it.

**Analysis:** every native→Scheme→native re-entrancy level (`vm_dispatch.zig` native call sites, `forceFn`'s `pushRoot` in `primitives_lazy.zig`) holds GC root slots; the root stack is a fixed 1024-slot array whose overflow handler is `@panic` (`src/memory.zig:1123`). For the promise case specifically, `forceFn` could additionally detect direct re-entrant forcing on entry via the existing `Promise.forcing` flag (the thunk-*returns*-forced-promise case is already detected per SRFI-45 §8, see #210) and raise the same catchable "re-entrant forcing of promise" error.

**Verified:** fresh ReleaseSafe build at 96ce73b; gc-stress build: unchanged (deterministic — the panic depends on re-entrancy depth, not collection timing)
**Source location:** src/memory.zig (`pushRoot`), src/vm_dispatch.zig native call sites, src/primitives_lazy.zig (`forceFn`)
**Disabled test:** tests/scheme/audit/primitives_lazy-audit.scm (marked ;; FAIL: #this)

**Found by:** Systematic audit, Phase 2.16

Issue #1245, opened 2026-07-05
Epic: eliminate native VM re-entrancy in control-flow primitives (trampoline the primitive->Scheme boundary)


baijum:
**Category:** VM architecture epic (from audit Phase 5 synthesis; campaign tracking: see #1137)

Higher-order primitives that call back into Scheme (`callWithArgs` -> `callReentrant` in src/vm_calls.zig) stack a native Zig frame plus GC roots per call. The fixed 1024-slot GC root stack (src/memory.zig:1123) turns deep nesting into an **uncatchable panic**, and even a graceful error would leave R7RS §3.5's required tail calls unsatisfiable. One architectural fix — dispatching primitive-driven callbacks through the bytecode loop (trampoline/CPS at the boundary) instead of re-entering natively — resolves this whole cluster:

**Member issues:**
- see #1191 — deep native re-entrancy panics "GC root stack overflow (1024)" instead of raising (umbrella crash mechanism; self-force, nested map depth 2000)
- see #1240 — call-with-values consumer, call/cc receiver, and eval are not tail-called (§3.5 requires); each panics at depth ~1024
- see #1241 — let*-values body is not a tail context (compiles through the native path; let-values is pure bytecode and conforms)

**Adjacent, same boundary:**
- see #1198 — non-tail apply routes through native applyFn (its tail form already uses the bounded tail_apply opcode — the model for this epic)
- Documented limitations that this would lift: continuations captured under native map/for-each cannot resume after the native frame returns (README/docs); fibers cannot park inside callbacks driven by native higher-order procedures (README "Fibers")

**Fix sketch:** the tail_apply opcode proves the pattern — reify "call this closure with these args, then continue here" as VM state instead of a Zig call. Candidates in rough order of payoff: call-with-values, call/cc receiver, map/for-each/vector-map family, eval, dynamic-wind thunks, force.

**Priority:** P0 — this cluster contains the only uncatchable crashes in the audit's findings that are reachable from pure R7RS code.

Found by: Systematic audit, Phases 2/4 (synthesis in Phase 5)

baijum:
**Closing — the P0 cluster is fully resolved.** All member and adjacent issues, plus the spun-off residual, are closed as completed, and the implementation shipped in **PR #1248** (merged 2026-07-05; 1774 Scheme tests pass / 0 fail, R7RS 1395 / 0):

[...]

Issue #1198, opened 2026-07-05
[primitives.zig] reverse, append, and apply hang with unbounded allocation on circular list arguments (length/list? detect cycles; #688 fixed the same class in primitives_list.zig)


baijum:
**Category:** VM / primitives
**Spec reference:** R7RS §6.4 — "(reverse list): It is an error if list is not a list"; circular chains are not lists, so an implementation may reject them — but an unbounded-allocation hang is poor QoI, sibling procedures in the same file already detect cycles, and apply's behavior differs by call position.
**Severity:** edge-case (hang + unbounded memory growth)

**Minimal reproduction:**
```scheme
(define c (list 1 2))
(set-cdr! (cdr c) c)

(reverse c)          ;; hangs, allocating pairs until OOM/kill
(append c '(9))      ;; hangs, growing the element buffer until OOM/kill
(begin (apply + c) #f) ;; hangs — native applyFn walks the final list without cycle detection
```

**Expected:** a catchable "proper list" error, as `length` already produces:
```scheme
(length c)   ;; => error: type error in 'length': expected proper list  (Floyd cycle detection)
(list? c)    ;; => #f  (Floyd cycle detection)
```

**Actual:** `reverse`, non-last-argument `append`, and `applyFn`'s final-list flattening walk `while (rest != NIL)` with no cycle detection (src/primitives.zig) — they never terminate and allocate on every iteration.

**Position-dependent apply inconsistency:** when `(apply + c)` appears in *tail position*, compiler.zig:678 compiles it to the `tail_apply` opcode, which bounds out with a quick catchable (if contentless) error. In non-tail position the native `applyFn` hangs. Same expression, different termination behavior:
```scheme
(guard (e (#t 'caught)) (apply + c))          ;; => 'caught quickly (tail_apply opcode)
(guard (e (#t 'caught)) (apply + c) #f)       ;; hangs forever (native applyFn)
```

**Precedent:** #688 fixed this exact class for member/memq/memv/assoc/assq/assv/list-copy in primitives_list.zig; #859 for the cyclic REPL printer. The Floyd slow/fast pattern used by `length`/`listP` in the same file is the ready-made fix.

**Verified:** fresh ReleaseSafe build at 96ce73b; deterministic algorithmic hang (not GC-timing dependent — gc-stress not applicable)
**Source location:** src/primitives.zig (`reverse`, `append`, `applyFn`)
[...]

PR #1248, opened 2026-07-05
Eliminate native VM re-entrancy in control-flow primitives

baijum:
## Summary

Fixes the four issues in epic #1245 — the only uncatchable crashes reachable from pure R7RS code:

- **#1198**: Add Floyd cycle detection to `reverse`, `append`, and `applyFn` — circular lists now raise a catchable error instead of hanging with unbounded allocation
- **#1241**: Rewrite `let*-values` desugaring to `(apply consumer (call-with-values producer list))` so the body compiles to `tail_apply` — proper tail context per R7RS §3.5
- **#1240**: Add tail-position compiler rewrites for `call-with-values` (apply desugaring), `call/cc` (new `tail_call_cc` opcode), and `eval` (new `tail_eval` opcode) — satisfies R7RS §3.5 required tail calls
- **#1191**: Increase GC root buffer 1024→4096, add catchable `StackOverflow` guard in `callReentrant` — replaces uncatchable `@panic` for deep native re-entrancy

## Test plan

- [x] 7 FAIL markers removed, 8 disabled tests re-enabled across 3 test files
- [x] `bash tests/scheme/run-all.sh` — 1774 pass, 0 fail, 0 timeout
- [x] Manual verification: `(lp 5000)` loops through call-with-values, call/cc, eval, let*-values all complete without panic
- [x] `(deep 2000)` nested map completes; `(deep 5000)` raises catchable StackOverflow
- [x] `(define p (delay (force p))) (guard (e (#t 'caught)) (force p))` → `caught`
- [ ] Linux CI

### DIFF
--- a/src/bytecode_file.zig
+++ b/src/bytecode_file.zig
@@ -499,7 +499,7 @@ fn validateFunctionBytecode(func: *Function) BytecodeError!void {
     var ip: usize = 0;
     while (ip < code.len) {
         const raw = code[ip];
-        if (raw > @intFromEnum(OpCode.self_tail_call)) return BytecodeError.CorruptedFile;
+        if (raw > @intFromEnum(OpCode.tail_eval)) return BytecodeError.CorruptedFile;
         const op: OpCode = @enumFromInt(raw);
         ip += 1;
 
@@ -518,9 +518,13 @@ fn validateFunctionBytecode(func: *Function) BytecodeError!void {
                 if (ip + 4 > code.len) return BytecodeError.CorruptedFile;
                 ip += 4;
             },
-            .call, .tail_call, .tail_apply, .self_tail_call => {
+            .call, .tail_call, .tail_apply, .self_tail_call, .tail_eval => {
                 if (ip + 3 > code.len) return BytecodeError.CorruptedFile;
                 ip += 3;
+            },
+            .tail_call_cc => {
+                if (ip + 2 > code.len) return BytecodeError.CorruptedFile;
+                ip += 2;
             },
             .get_global => {
                 if (ip + 4 > code.len) return BytecodeError.CorruptedFile;
--- a/src/bytecode_file.zig
+++ b/src/bytecode_file.zig
@@ -523,8 +523,8 @@ fn validateFunctionBytecode(func: *Function) BytecodeError!void {
                 ip += 3;
             },
             .tail_call_cc => {
-                if (ip + 2 > code.len) return BytecodeError.CorruptedFile;
-                ip += 2;
+                if (ip + 4 > code.len) return BytecodeError.CorruptedFile;
+                ip += 4;
             },
             .get_global => {
                 if (ip + 4 > code.len) return BytecodeError.CorruptedFile;
--- a/src/compiler.zig
+++ b/src/compiler.zig
@@ -675,11 +675,37 @@ pub const Compiler = struct {
             }
         }
 
-        if (is_tail and types.isSymbol(head) and std.mem.eql(u8, types.symbolName(head), "apply")) {
-            if (self.resolveLocal(types.symbolName(head)) == null and
-                (try self.resolveUpvalue(types.symbolName(head))) == null)
+        if (is_tail and types.isSymbol(head)) {
+            const sym_name = types.symbolName(head);
+            if (std.mem.eql(u8, sym_name, "apply")) {
+                if (self.resolveLocal(sym_name) == null and
+                    (try self.resolveUpvalue(sym_name)) == null)
+                {
+                    return passthrough.compileApplyTail(self, expr, dst);
+                }
+            }
+            if (std.mem.eql(u8, sym_name, "call-with-values")) {
+                if (self.resolveLocal(sym_name) == null and
+                    (try self.resolveUpvalue(sym_name)) == null)
+                {
+                    return passthrough.compileCallWithValuesTail(self, expr, dst);
+                }
+            }
+            if (std.mem.eql(u8, sym_name, "call-with-current-continuation") or
+                std.mem.eql(u8, sym_name, "call/cc"))
             {
-                return passthrough.compileApplyTail(self, expr, dst);
+                if (self.resolveLocal(sym_name) == null and
+                    (try self.resolveUpvalue(sym_name)) == null)
+                {
+                    return passthrough.compileCallCCTail(self, expr, dst);
+                }
+            }
+            if (std.mem.eql(u8, sym_name, "eval")) {
+                if (self.resolveLocal(sym_name) == null and
+                    (try self.resolveUpvalue(sym_name)) == null)
+                {
+                    return passthrough.compileEvalTail(self, expr, dst);
+                }
             }
         }
 
--- a/src/compiler_bindings.zig
+++ b/src/compiler_bindings.zig
@@ -652,16 +652,25 @@ pub fn compileLetStarValues(self: *Compiler, args: Value, dst: u16, is_tail: boo
     return self.compileDesugared(desugared, dst, is_tail);
 }
 
-/// Build nested call-with-values for a list of bindings.
-/// (let-values (((a b) e1) ((c) e2)) body)
+/// Build nested apply-over-call-with-values for sequential let*-values.
+/// (let*-values (((a b) e1) ((c) e2)) body)
 /// =>
-/// (call-with-values (lambda () e1) (lambda (a b) (call-with-values (lambda () e2) (lambda (c) body))))
+/// (apply (lambda (a b)
+///          (apply (lambda (c) (begin body))
+///                 (call-with-values (lambda () e2) list)))
+///        (call-with-values (lambda () e1) list))
+///
+/// The call-with-values calls are in argument (non-tail) position, and the
+/// apply calls form the tail chain — compiling to tail_apply opcodes that
+/// reuse frames in-place, satisfying R7RS §3.5 tail-context requirements.
 pub fn buildLetValues(self: *Compiler, bindings: Value, body: Value) !Value {
     const gc = self.gc;
     gc.no_collect += 1;
     defer gc.no_collect -= 1;
     const lambda_sym = try gc.allocSymbol("lambda");
     const cwv_sym = try gc.allocSymbol("call-with-values");
+    const apply_sym = try gc.allocSymbol("apply");
+    const list_sym = try gc.allocSymbol("list");
 
     if (bindings == types.NIL) {
         const begin_sym = try gc.allocSymbol("begin");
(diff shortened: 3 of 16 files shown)


baijum:
Reviewed with a local build of this branch (cf5f796), behavioral probes against a pre-PR baseline, and a read of the new opcode/dispatch/compiler paths. The approach is right — `tail_apply`-style opcodes for call/cc and eval, the apply-desugar for cwv/let*-values, and the catchable guard in `callReentrant` genuinely resolve all four member issues, and everything in the test plan reproduces locally. But I found **two regressions** that should be fixed before merge.

## 1. Tail `call/cc`: re-invoking the captured continuation returns the wrong value when the result register is rebased

```scheme
(define k2 #f)
(define log2 '())
(define (g) (let ((pad 'stale)) (call/cc (lambda (c) (set! k2 c) 'first))))
(set! log2 (cons (g) log2))
(if (< (length log2) 3) (k2 'again))
(display (reverse log2))
```

- main: `(first again)`
- this branch: `(first #<undefined>)`

Root cause, visible in the disassembly of `g`:

```
  0010  tail_call_cc    r2
  0013  return          r0
```

`compileCallCCTail` captures the continuation with `dst_reg = base` (r2), but the instruction the resumed frame lands on is the body epilogue `return last_dst` (r0). When `needs_rebase` is true (`dst + 1 != next_register`, e.g. any tail call/cc under a `let`), the delivered value goes to `base` while the `return` reads the never-written `dst`. The non-rebase case (`base == dst`) works, which is why the simple probes pass.

Two possible fixes: add the result register as a second operand to `tail_call_cc` and capture with `dst_reg = dst`, or emit a `move dst, base` immediately after `tail_call_cc` (that slot is only reachable via continuation resume — normal flow reuses the frame and never returns there). The first is cleaner.

Test-coverage note: nothing in the suite invokes a continuation captured in tail position after the capturing frame has returned — the re-enabled `loop-cc` only captures and discards. Please add both variants (with and without an enclosing `let`) as regression tests. Keep it to a single re-entry: the second re-entry from top level stops on main too (pre-existing top-level continuation behavior, not this PR's problem).

## 2. Hygiene regression: `let*-values` and tail `call-with-values` break when `list` or `apply` is shadowed

```scheme
(define (h list) (let*-values (((a b) (values 1 2))) (+ a b)))
(h 99)   ; main: 3   this branch: "type error in 'call-with-values': expected procedure, got 99"

(define (i list) (call-with-values (lambda () (values 1 2)) +))
(i 99)   ; main: 3   this branch: same error

(define (j apply) (let*-values (((a) (values 5))) a))
(j 99)   ; main: 5   this branch: "not a procedure"
```

The desugared `(apply consumer (call-with-values producer list))` compiles the introduced `apply`/`list` references in the user's scope, so local bindings capture them. The class isn't new — shadowing `call-with-values` already broke `let-values` on main — but `list` and `apply` are common parameter names in real Scheme code, so the practical exposure is much wider (any `(lambda (list) ...)` containing a `let*-values` now breaks).

Suggested fix: make the introduced references resolve to the builtins unconditionally — either emit the apply/cwv sequence as bytecode instead of an s-expression desugar, or compile those two symbols through a global-only lookup that skips `resolveLocal`/`resolveUpvalue`. Please also add shadowing regression tests for both forms.

## Smaller notes

- **Global redefinition inconsistency**: `(define eval (lambda (x) 'mine))` is honored on main and in non-tail position on this branch, but ignored in tail position (the opcode fires regardless of the global). R7RS gives latitude here and the pre-existing `apply` rewrite has the same property, so this is acceptable — but tail vs non-tail now disagree within one build; worth a line in the docs or a follow-up.
- `(define (f) (call/cc))` is now a compile-time `InvalidSyntax` instead of a runtime arity error (same for `eval`). Matches existing `compileApplyTail` behavior; note only.
- `types.zig` opcode comments say `base:u8`, but both operands are emitted/read as u16 (`tail_call_cc // base:u16`, `tail_eval // base:u16, nargs:u8`).
- `r7rs-tail-procedures-gaps.scm` adds a mid-file `(import (scheme eval) (scheme repl))` that duplicates the header import on line 12.
- `promise.forcing = false` in the chain-collapse branch of `forceFn` looks correct (prevents a stale flag from triggering a false "re-entrant forcing" later) but has no dedicated test — worth one, per the repo's bug-fix-needs-test rule.
- The `callReentrant` guard's 32-slot headroom with the `pushRoot` panic retained as backstop is reasonable; a single native level pushing >32 roots would still panic, but nothing does today.

## Verified working

- All re-enabled loops at N=25000 (`let*-values`, cwv consumer, call/cc receiver, eval) run space-bounded.
- `tail_call_cc` into a receiver with 40+ locals — register capacity handled correctly.
- Zero-values producer with nullary consumer in tail position; multi-clause `let*-values`; values-count mismatch raises catchably through the new desugar.
- `(call/cc 5)` in tail position raises catchably; compile errors and unbound variables through `tail_eval` raise catchably; `eval`'s lenient non-environment second arg matches the native path.
- Floyd cycle detection in `reverse`/`append`/`applyFn` is correct, and formatting the circular value into the error message terminates.
- `.sbc` validator, disassembler, and dispatch operand widths are mutually consistent for both new opcodes; old caches stay readable and new-opcode caches are rejected (not misparsed) by older readers.
- Deep native re-entrancy (`(delay (force selfp))`, nested map at 5000) now raises catchable `StackOverflow` instead of panicking.

## Scope

This covers the four member issues; the epic's broader trampoline items (map/for-each family, dynamic-wind thunks, continuations captured under native frames, fiber parking in native callbacks) remain — suggest keeping #1245 open with a scope note, or splitting the remainder into a new issue. On merge, #1191/#1198/#1240/#1241 need manual closing (no closing keywords in the body, per campaign convention).

baijum:
Both regressions fixed in 2a62426:

**1. `tail_call_cc` rebase bug** — Added `dst` as a second operand to the opcode. The continuation now captures with `dst_reg = dst` (the body result register) instead of `base_reg`, so the resumed frame's `return dst` reads the correct value. Regression tests cover both the rebase path (call/cc under a `let`) and the no-rebase path.

**2. Hygiene regression** — Reverted `buildLetValues` to the original `(call-with-values producer consumer)` form. Rewrote `compileCallWithValuesTail` to emit `call_global`/`get_global` bytecode directly for the introduced `call-with-values` and `list` references, bypassing lexical scope resolution entirely. Shadowing `list` or `apply` in user code no longer affects the desugaring. Regression tests for all three reviewer examples (`shadow-list`, `shadow-apply`, `shadow-cwv`).

**Debug CI fix** — Replaced the GC root count heuristic with a `native_reentry_depth` counter on the VM struct (limit 200 in Debug, 3000 in Release). This catches deep native re-entrancy before the Zig native stack overflows in Debug builds, where per-frame stack usage is ~500x larger.

**Also addressed:**
- Duplicate `(import (scheme eval) (scheme repl))` removed from test file
- Added `delay-force` chain test for the `promise.forcing = false` fix
- Updated opcode comment in `types.zig` to reflect actual u16 operand widths

6 new regression tests added, 1774 pass / 0 fail.

baijum:
Re-reviewed 2a62426 with a local ReleaseSafe build. Both original regressions are confirmed fixed, but the reworked re-entrancy guard **reintroduces the uncatchable panic** this PR exists to eliminate.

## Confirmed fixed

- **tail_call_cc rebase**: re-invocation now returns the delivered value in both the rebase (`call/cc` under `let`) and no-rebase cases — `(first again)` on my probes, matching main. Operand encoding is consistent across compiler/validator/disassembler/dispatch (4 bytes).
- **Hygiene**: all three shadowing repros now give `3` / `3` / `5`. The direct `get_global`/`call_global` emission also keeps the semantics intact — values-count mismatch raises catchably, zero-values producer with nullary consumer works, multi-clause `let*-values` works, `(call/cc 5)` raises catchably.
- Benchmarks flat (call_cc 1.00x, tailcall 1.02x) — the new opcodes cost nothing on hot paths.

## New blocker: depth-only guard re-opens the GC root panic

```scheme
(define (deep n)
  (if (= n 0) 0
      (car (map (lambda (x) (deep (- x 1))) (list n)))))
(guard (e (#t 'caught)) (deep 2500))
```

On this branch (ReleaseSafe, macOS ARM):

```
thread ... panic: GC root stack overflow
  memory.zig:1123 in pushRoot
  vm_dispatch.zig:889 in runUntil
exit code 134
```

Mechanism: each `map`→`callReentrant` level holds ~2 GC roots across its recursion, so the 4096-slot root buffer exhausts at depth ~2050 — **before** the 3000-deep `native_reentry_depth` guard ever fires. Everything in the ~2050–3000 window is an uncatchable SIGABRT from pure R7RS code, which is exactly the #1191 crash class. `(deep 3500)` panics the same way.

cf5f796's `root_count` check caught this case regardless of the roots-per-level ratio (your own test plan's "`(deep 5000)` raises catchable StackOverflow" ran through it). The depth counter is the right tool for the *Zig native stack* (especially Debug), but it can't protect the root buffer. Keep both:

```zig
const max_native_depth: u16 = if (@import("builtin").mode == .Debug) 200 else 3000;
if (vm.native_reentry_depth >= max_native_depth or
    vm.gc.root_count > vm.gc.root_buffer.len - 32)
{
    vm.setErrorDetail("native re-entrancy too deep", .{});
    return VMError.StackOverflow;
}
```

Please also add the deep-map repro as a regression test — it's conveniently mode-portable: in Debug the 200-depth guard fires early (still `'caught`, and fast); in Release it must raise via the root check rather than panic. `(guard (e (#t 'caught)) (deep 2500))` expecting `'caught` covers the window the suite currently misses (nothing in the 1774 exercises native nesting between ~2050 and 3000, which is why this slipped through green).

## The forcing-flag test doesn't pin its fix

I rebuilt 2a62426 with the `promise.forcing = false` chain-collapse line removed: the new "delay-force chain forcing flag" test still passes (14/14 in the file), as do the lazy audit (36/36) and the smoke suite. Tracing the loop, every exit path except the 100000-iteration cap already resets the flag, and the next iteration re-sets it before the thunk call — so the stale flag seems observable only through the cap path, which the chain test never reaches. Either find the observable repro and test that, or keep the line as defensive hardening with a comment saying so — as-is the test name implies coverage it doesn't provide.

## Nits

- Stray editing artifact left in `r7rs-tail-procedures-gaps.scm`: the comment line `;; Remove duplicate import (already imported at line 12)`.
- Globally redefining `list` now breaks tail `call-with-values` (`(define list (lambda args 'evil))` → error in tail position, works non-tail) via the `get_global list` reference. Same R7RS "is an error" latitude class as the `eval` note from the first review — fine to leave, just noting for completeness.

baijum:
Fixed in bf0253a:

**Root_count guard restored** — the depth-only check left a ~2050–3000 window where the 4096-slot root buffer exhausted before the depth limit fired. Now checks both: `native_reentry_depth >= max OR root_count > buffer.len - 32`. The reviewer's `(deep 2500)` and `(deep 3500)` reproductions both return `'caught`.

**Deep-map regression test added** — `(guard (e (#t 'caught)) (deep-map 2500))` expecting `'caught`, exercises the root_count guard path that the existing suite missed.

**Forcing-flag test clarified** — renamed to "delay-force chain resolves" with a comment noting the `promise.forcing = false` reset is defensive hardening (not directly observable from Scheme, since all other exit paths already clear it and the next iteration re-sets it before the thunk call).

**Stray comment removed**.

1774 pass / 0 fail.

baijum:
Verified bf0253a with a local ReleaseSafe build — **all findings resolved, nothing further blocking from my side.**

- `(deep-map 2500)` and `(deep-map 3500)` under `guard` now return `'caught` (exit 0, no panic) — the ~2050–3000 root-buffer window is closed, and the new regression test pins it.
- Both earlier regressions stay fixed: tail `call/cc` re-entry returns the delivered value in rebase and no-rebase cases; all three `list`/`apply` shadowing repros pass.
- Self-force still raises catchably; touched and adjacent suites all pass locally (tail-procedures 15, tail-positions 18, core audit 199, lazy audit 36, continuation gaps 22, derived smoke 33).
- The defensive-hardening comment on the forcing-flag reset is an honest resolution of the test-the-test finding.

Pending CI on this commit, this looks ready to merge. Reminder from the first review: on merge, close #1191/#1198/#1240/#1241 manually, and either keep epic #1245 open for the remaining trampoline items (map/for-each family, dynamic-wind thunks, continuations under native frames, fiber parking) or split them into a follow-up issue.

PR #1249, opened 2026-07-05
Box set!-mutated locals for R7RS store semantics (#1168)

### LINKED ISSUE #1168: [R7RS 6.10] Re-entrant call/cc rolls back set! mutations of non-captured locals (register snapshot violates store semantics) — infinite loops
**Category:** VM / R7RS conformance (continuations)
**Spec reference:** R7RS 3.4 (storage model) and 4.1.6 — set! stores a value in the location bound to the variable; locations are part of the store, not the continuation. Re-invoking a continuation must not undo assignments (R7RS 3.5/6.10 semantics; every Scheme's generator/loop idiom depends on this).
**Severity:** wrong-result / hang

**Minimal reproduction (hangs — run with a timeout):**
```scheme
(let ((n 0) (k* #f))
  (call/cc (lambda (k) (set! k* k)))
  (set! n (+ n 1))
  (if (< n 3) (k* #f))
  n)
;; expected: 3
;; actual:   infinite loop — n's increment is rolled back on each re-entry
```

**Discriminating experiments** (all on fresh build at f271b4b):
- plain `set!` local (not referenced by any lambda): **hangs** — mutation rolled back
- same counter as `(vector 0)` heap cell: terminates, returns 3
- same `set!` local but also read by an inner `define`d procedure (closure-captured): terminates, returns 3

**Root cause:** the compiler boxes a mutable local only when it is closure-captured; otherwise the variable lives in a VM register. `captureContinuation`/`restoreContinuation` (src/vm_continuations.zig) snapshot and restore the register file, so re-entering a continuation rolls unboxed mutations back to their captured values. Per R7RS a `set!` variable is a store location regardless of capture; any register-allocated mutable local that is live across a `call/cc` must either be boxed or excluded from restore.

This is why the existing continuations suite doesn't catch it: its re-entry counters are closure-captured or heap cells.

Related closed work: #870/#875 fixed other restore-path bugs; this one is about store semantics, not winding.

**Verified:** fresh ReleaseSafe build at f271b4b; deterministic; hangs in top-level, function-body, and argument positions alike.
**Disabled test:** `tests/scheme/audit/primitives_control-audit.scm` (marked `;; FAIL: #this`)

**Found by:** Systematic audit, Phase 2.7 (#1137)

baijum:
## Summary

- Fix infinite loops when `set!` mutations to non-closure-captured locals are rolled back by continuation re-entry (R7RS §3.4 store semantics violation)
- Add `boxIfSetTarget` helper that boxes any local whose name appears in the top-level form's `set_targets` map — applied at all binding sites (`let`, `let*`, lambda params, `do`)
- Enable the previously-disabled audit test and add 8-case regression test suite covering `let`, `let*`, lambda params, multiple mutated locals, nested scopes, and existing workarounds

Fixes #1168

## Test plan

- [x] Full Scheme test suite passes (1775 pass, 0 fail)
- [x] New `callcc-set-store.scm` regression test passes all 8 cases
- [x] Previously-disabled audit test now passes
- [x] Disassembly confirms `box_local` + `set_box_local` emitted for mutated locals


baijum:
Reviewed against #1168, with the branch built locally (ReleaseSafe) and tested. The approach is sound and the coverage of binding sites is complete for source-level `set!`:

- `let`, `let*`, lambda params (both `compileLambda` and `compileLambdaWithIR`), and `do` are handled by the new hook.
- `letrec`/`letrec*`, named-let loop vars, and internal defines already box unconditionally, and `let-values`/`let*-values`, `guard`, `case-lambda`, and named-let bindings desugar through `lambda`/`let`, so they inherit the fix — verified by reading each path.
- `markLocalBoxedBySlot` emits `box_local` immediately and is idempotent, so boxing right after `addLocal` is correct at every call site.
- The original repro now returns 3, and the new 8-case suite passes; `run-all.sh` picks the new file up via the `continuations/*.scm` glob.

**One confirmed gap: macro-introduced `set!` still hangs.** `collectSetTargets` runs in `Compiler.compile()` on the raw, pre-expansion form (compiler.zig:383), but macros expand later during IR lowering. So a `set!` that only exists post-expansion never lands in `set_targets`, the local stays register-allocated, and the #1168 rollback/hang persists. Both variants reproduce on this branch (hang, killed by 5s alarm):

```scheme
;; A: macro wraps the set!
(define-syntax inc! (syntax-rules () ((_ v) (set! v (+ v 1)))))
(let ((n 0) (k* #f))
  (call/cc (lambda (k) (set! k* k)))
  (inc! n)
  (if (< n 3) (k* #f))
  n)

;; B: macro expands to the whole let+set!
(define-syntax counter-loop
  (syntax-rules ()
    ((_ limit)
     (let ((n 0) (k* #f))
       (call/cc (lambda (k) (set! k* k)))
       (set! n (+ n 1))
       (if (< n limit) (k* #f))
       n))))
(counter-loop 3)
```

This is the same bug class the PR targets, so since the PR says "Fixes #1168" it's worth either (a) re-collecting set! targets from the post-expansion form — e.g. rescanning after `lowerAndOptimize`, or having IR lowering add targets as expansions are produced — or (b) landing this as-is and filing a follow-up for the macro case, with these two repros added as `;; FAIL:`-marked tests so the audit tracker stays honest.

Minor, non-blocking notes:
- Boxing is name-keyed across the whole top-level form, so any local sharing a name with any `set!` target anywhere in the form gets boxed. That's safe (just an extra indirection) and consistent with the existing `CaptureScan` over-approximation philosophy — fine as-is.
- `.sbc` caches are keyed on source hash + release version, so pre-fix cached bytecode keeps the old behavior until the next version bump. Only affects programs that were already hanging, so nothing to do here.

baijum:
Thanks for the thorough review and the two macro repros — confirmed the gap.

**Why the macro case can't be fixed within this PR's approach:** `collectSetTargets` runs pre-expansion, and adding targets during IR lowering (in `lowerSet`) discovers them too late — the `let` bindings are already compiled without boxing. Boxing mid-body at the `set!` site also fails: the continuation snapshot captured the raw register value before the `box_local`, so restore puts the raw value back and each re-entry creates a fresh box.

A proper fix needs either pre-expanding macros before binding compilation (two-pass) or restructuring the compiler to lower the full body first. Both are bigger changes.

Addressed as option (b):
- Added both reproduction cases as `FAIL`-marked tests in `callcc-set-store.scm` (f9458f7)
- Filed follow-up: #1250

baijum:
Confirmed the follow-up — this resolves the review. f9458f7 is comment-only, so the code is identical to what I built and tested earlier (original repro returns 3, 8-case suite passes), and #1250 captures the macro gap well, including why the lazy-boxing route can't work. Agreed that a proper fix is out of scope here.

One tiny consistency nit before merge: the new FAIL markers don't embed the issue number. The audit convention elsewhere is `;; FAIL: #NNNN` (e.g. the old `;; FAIL: #1168` in `primitives_control-audit.scm`), but the new comments say "See follow-up issue for these cases" without naming it — so a tree-wide grep for `#1250` won't surface these disabled tests. Suggest:

```scheme
;; FAIL: #1250 macro-introduced set! (inc! wrapper) — HANGS
...
;; FAIL: #1250 macro that expands to entire let+set! — HANGS
```

(and/or naming #1250 in the "Known gap" header comment). With that, LGTM — good to merge.

PR #1251, opened 2026-07-06
Deliver multiple values when continuation invoked with != 1 arg

### LINKED ISSUE #1169: [R7RS 6.10] Invoking a continuation with multiple arguments drops all but the first value (call-with-values producer)
**Category:** VM / R7RS conformance (continuations)
**Spec reference:** R7RS 6.10, call-with-current-continuation: "The escape procedure accepts the same number of arguments as the continuation to the original call to call-with-current-continuation." And call-with-values: the continuation of the producer call "when passed some values, calls the consumer procedure with those values as arguments."
**Severity:** wrong-result

**Minimal reproduction:**
```scheme
(call-with-values
  (lambda () (call/cc (lambda (k) (k 1 2))))
  list)
;; expected: (1 2)
;; actual:   (1) — second value silently dropped
```
The continuation captured in `call-with-values`'s producer position accepts multiple values, but invoking it with two arguments delivers only the first to the consumer.

**Root cause location:** `src/vm_continuations.zig` — continuation invocation appears to deliver a single value into the destination register; the multiple-values case (escape procedure called with 0 or 2+ args) needs to package a MultipleValues object when the resumed continuation is a values-accepting context.

**Verified:** fresh ReleaseSafe build at f271b4b; deterministic. `(call/cc (lambda (k) (k 42)))` and plain escapes work; only multi-argument invocation drops values.
**Disabled test:** `tests/scheme/audit/primitives_control-audit.scm` (marked `;; FAIL: #this`)

**Found by:** Systematic audit, Phase 2.7 (#1137)

baijum:
## Summary

Fixes #1169

- Continuation invocation (`call/cc` and `call/ec`) was silently dropping all but the first argument — `(k 1 2)` in a `call-with-values` producer delivered only `1` to the consumer
- When `nargs != 1`, wrap args in a `MultipleValues` object (mirroring `values` semantics) so `call-with-values` consumers receive the correct argument count
- Fixed in all three dispatch sites: `vm_dispatch.zig` (regular call + flat-args paths) and `vm_calls.zig` (`callValue` fallback)

## Test plan

- [x] Zig unit tests pass (`zig build test` — 639/639)
- [x] 3 new unit tests: `call/cc` multi-arg, `call/cc` zero-arg, `call/ec` multi-arg (all in `call-with-values` context)
- [x] Re-enabled audit test from `primitives_control-audit.scm`
- [x] Manual verification: `(call-with-values (lambda () (call/cc (lambda (k) (k 1 2)))) list)` → `(1 2)`
- [ ] Scheme test suite (`run-all.sh`)

### DIFF
--- a/src/tests_continuations.zig
+++ b/src/tests_continuations.zig
@@ -160,6 +160,56 @@ test "values with zero values" {
     try std.testing.expectEqual(@as(i64, 99), types.toFixnum(result));
 }
 
+// Regression: #1169 — invoking a continuation with multiple arguments
+// must deliver all values, not just the first.
+test "call/cc multi-arg invocation in call-with-values" {
+    var gc = memory.GC.init(std.testing.allocator);
+    defer gc.deinit();
+    var vm = try th.makeTestVM(&gc);
+    defer vm.deinit();
+
+    const result = try vm.eval(
+        \\(call-with-values
+        \\  (lambda () (call/cc (lambda (k) (k 1 2))))
+        \\  list)
+    );
+    try std.testing.expect(types.isPair(result));
+    try std.testing.expectEqual(@as(i64, 1), types.toFixnum(types.car(result)));
+    try std.testing.expectEqual(@as(i64, 2), types.toFixnum(types.car(types.cdr(result))));
+    try std.testing.expectEqual(types.NIL, types.cdr(types.cdr(result)));
+}
+
+test "call/cc zero-arg invocation in call-with-values" {
+    var gc = memory.GC.init(std.testing.allocator);
+    defer gc.deinit();
+    var vm = try th.makeTestVM(&gc);
+    defer vm.deinit();
+
+    const result = try vm.eval(
+        \\(call-with-values
+        \\  (lambda () (call/cc (lambda (k) (k))))
+        \\  (lambda () 99))
+    );
+    try std.testing.expectEqual(@as(i64, 99), types.toFixnum(result));
+}
+
+test "call/ec multi-arg invocation in call-with-values" {
+    var gc = memory.GC.init(std.testing.allocator);
+    defer gc.deinit();
+    var vm = try th.makeTestVM(&gc);
+    defer vm.deinit();
+
+    const result = try vm.eval(
+        \\(call-with-values
+        \\  (lambda () (call/ec (lambda (k) (k 10 20 30))))
+        \\  list)
+    );
+    try std.testing.expect(types.isPair(result));
+    try std.testing.expectEqual(@as(i64, 10), types.toFixnum(types.car(result)));
+    try std.testing.expectEqual(@as(i64, 20), types.toFixnum(types.car(types.cdr(result))));
+    try std.testing.expectEqual(@as(i64, 30), types.toFixnum(types.car(types.cdr(types.cdr(result)))));
+}
+
 test "dynamic-wind with escape continuation" {
     var gc = memory.GC.init(std.testing.allocator);
     defer gc.deinit();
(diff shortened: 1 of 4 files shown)


baijum:
Reviewed with the branch built locally and tested. The core fix is right: the wrap mirrors `valuesFn` exactly (1 arg → direct, 0 or 2+ → `MultipleValues`), so `(k)` now agrees with `(values)` semantics, and GC-wise the new `allocMultipleValues` calls are safe (registers are marked VM roots, and the flat-args list stays register-reachable for the duration of the opcode; the slice is duped after `maybeCollect`). Direct `(k 1 2)`, tail `apply`, and the compiled `call-with-values` consumer path all deliver correctly, and I ran the full Scheme suite for the unchecked box in the test plan: **1775 pass, 0 fail** (plus audit control 35/0 and all `continuations/*.scm`).

**One missed dispatch site: the `callWithArgs` continuation branch** (`src/vm_calls.zig:656` on this branch) still has the old logic:

```zig
const value = if (args.len == 0) types.VOID else args[0];
```

This is the path native primitives use to invoke a continuation, and it's reachable from pure R7RS code in at least three ways — all verified on this branch's build:

```scheme
;; 1. non-tail apply of a continuation → (1), expected (1 2)
(call-with-values
  (lambda () (call/cc (lambda (k) (car (list (apply k (list 1 2)))))))
  list)

;; 2. zero-arg non-tail apply → spurious ArityMismatch, expected 99
;;    (VOID is delivered as one argument to a 0-ary consumer)
(call-with-values
  (lambda () (call/cc (lambda (k) (car (list (apply k '()))))))
  (lambda () 99))

;; 3. call-with-values as a first-class value (bypasses the opcode,
;;    hits callWithValuesFn → callWithArgs) → (1), expected (1 2)
(define cwv call-with-values)
(call-with-values
  (lambda () (call/cc (lambda (k0) (cwv (lambda () (values 1 2)) k0))))
  list)
```

Note case 2 is arguably worse than the original bug — it's a new *error* rather than a silent wrong value, and it exists on main too. Since this PR says "Fixes #1169" and the fix is the identical 3-line wrap, I'd fold it into this PR rather than file a follow-up, and add these three as regression tests.

**Suggestion while touching it:** the same wrap now appears in three places (four with `callWithArgs`). A tiny shared helper in `vm_calls.zig`, e.g.

```zig
pub fn continuationArgValue(vm: *VM, args: []const Value) VMError!Value {
    if (args.len == 1) return args[0];
    return vm.gc.allocMultipleValues(args) catch return VMError.OutOfMemory;
}
```

called from all four sites would prevent exactly this kind of divergence next time. (The other two `isContinuation` sites are fine as-is: the `call/cc`-receiver path and `callHandler` both deliver exactly one value by construction.)

Behavior note, no action needed: `(+ 1 (call/cc (lambda (k) (k 1 2))))` now raises `type error in 'arithmetic': expected number, got #<multiple_values>` instead of silently using the first value. That's "is an error" territory in R7RS 6.10, and a catchable error is a defensible (arguably better) choice.

baijum:
Re-verified 1629fe1 with a fresh build — this resolves the review. **LGTM, good to merge.**

- All three repros from the review now pass on this branch: non-tail `(apply k (list 1 2))` → `(1 2)`, zero-arg `(apply k '())` → `99` (no more spurious ArityMismatch), first-class `call-with-values` → `(1 2)`.
- All four dispatch sites now route through the shared `continuationArgValue` helper (nice touch taking `*GC` instead of `*VM`), and the three repros landed as unit tests.
- Nothing regressed: `zig build test` passes, the previously-fixed paths still deliver correctly, and the full Scheme suite is green — 1775 pass, 0 fail — so the remaining test-plan checkbox can be ticked.

With #1249 and this, both halves of the P0.5 continuation pair from the #1137 synthesis (store semantics + multi-value delivery) are done.

PR #1254, opened 2026-07-06
Fix GC crash on stale VM registers after thread start/join cycles

### LINKED ISSUE #1156: [GC/SRFI-18] Deterministic SIGSEGV in markValueInner under -Dgc-stress=true running the SRFI-18 audit file (stale root after thread start/join cycles)
**Category:** VM / GC (found via SRFI-18 audit)
**Severity:** crash (SIGSEGV)

**Reproduction** (deterministic, 3/3 runs, same fault address):
```bash
zig build -Dgc-stress=true
zig-out/bin/kaappi tests/scheme/audit/primitives_srfi18-audit.scm   # SIGSEGV
# or the first 62 lines of that file — crashes before printing any test result
```
The same file passes 73/0 on a normal ReleaseSafe build. Small standalone slices of the file (each thread-lifecycle block on its own) do NOT crash under gc-stress — the combined allocation pressure of the `(chibi test)` harness plus several OS-thread start/join cycles is required, so the repro is the file prefix, not a one-liner.

**Crash site:**
```
Segmentation fault at address 0x1868
gc_collect.zig:530 markValueInner        if (obj.owner != gc.id) return;
gc_collect.zig:510 markValue
memory.zig:1118    markValue (root_marker callback)
gc_collect.zig:496 markRoots             if (gc.root_marker) |mark| mark(gc);
gc_collect.zig:61  minorCollect
memory.zig:1088    maybeCollect
memory.zig:318     allocClosure
vm_dispatch.zig:761 runUntil             var cls_val = self.gc.allocClosure(func)
```
The parent VM's root marker reaches a Value whose object header is freed/garbage (0x1868 is a near-null offset — consistent with a dead pointer, not a valid heap object) during a minor collection triggered by an ordinary closure allocation on the main thread. The audit file's preceding sections start and join several OS threads (double-start guard, join-with-timeout returning a timeout-val, join-timeout raise, then re-join), so the prime suspects are root-set entries that outlive their object: `gc.extra_roots` thunk/fiber entries added by `thread-start!` and removed in `reapOsThread`, fiber `result`/`current_exception` fields, or scheduler fiber state left behind by the timed-out join paths (`threadJoinFn` returns the timeout-val WITHOUT reaping, so roots persist across further collections — combined with the later reap this is the widest window).

Note `reapOsThread` removes the thunk from `extra_roots` by value equality (`if (v == target.thunk)`) — if two threads share an eq? thunk value or the fiber's thunk field was updated, the wrong entry could be removed, leaving a stale root. Not verified as the cause; listed as a code-reading observation.

**Verified:** fresh build at 1a9e5ae (v0.12.0); gc-stress: deterministic SIGSEGV; normal build: no crash (this is the classic unrooted-value signature — see docs/dev/gc-safety-and-error-handling.md).
**Found by:** Systematic audit, Phase 2.1 (#1137)

baijum:
## Summary

- Clear local registers in `execute()` and `callClosure()` to prevent the GC from scanning stale pointers left by previous frames at the same register base
- Remove dead frame setup from `makeThreadFn` — OS thread fibers create their own VM and never use the parent fiber's `frames[0]`
- Add missing write barriers and defensive `frame_count=0` in `reapOsThread` after joining the OS thread

Fixes #1156.

**Root cause:** Between `execute()` calls (top-level form evaluations), `frame_count` drops to 0. A GC cycle during this gap doesn't scan any VM registers, so objects reachable only from stale register values are freed. When the next `execute()` pushes a frame covering those registers, the GC encounters dangling pointers — SIGSEGV.

## Test plan

- [x] 30/30 passes of `primitives_srfi18-audit.scm` under `-Dgc-stress=true` (was ~60% crash rate before fix)
- [x] New regression test `tests/scheme/smoke/thread-frame-gc-1156.scm` passes under both normal and gc-stress builds
- [x] `bash tests/scheme/run-all.sh` — 1776 pass, 0 fail

### DIFF
--- a/src/primitives_srfi18.zig
+++ b/src/primitives_srfi18.zig
@@ -201,7 +201,10 @@ fn makeThreadFn(args: []const Value) PrimitiveError!Value {
     @memset(fiber.registers, types.UNDEFINED);
     fiber.status = .created;
 
-    if (args.len > 1) fiber.name = args[1];
+    if (args.len > 1) {
+        fiber.name = args[1];
+        gc.writeBarrier(&fiber.header, args[1]);
+    }
 
     return types.makePointer(@ptrCast(fiber));
 }
--- a/src/primitives_srfi18.zig
+++ b/src/primitives_srfi18.zig
@@ -198,20 +198,7 @@ fn makeThreadFn(args: []const Value) PrimitiveError!Value {
     const fiber = gc.allocFiber(thunk, ctx.sched.next_id) catch return PrimitiveError.OutOfMemory;
     ctx.sched.next_id += 1;
 
-    const closure = types.toObject(thunk).as(types.Closure);
     @memset(fiber.registers, types.UNDEFINED);
-    fiber.registers[0] = thunk;
-    const vm = vm_mod.vm_instance orelse return PrimitiveError.OutOfMemory;
-    fiber.frames[0] = .{
-        .closure = closure,
-        .code = closure.func.code.items,
-        .ip = 0,
-        .base = 0,
-        .dst = 0,
-        .saved_wind_count = 0,
-        .seq = vm.nextFrameSeq(),
-    };
-    fiber.frame_count = 1;
     fiber.status = .created;
 
     if (args.len > 1) fiber.name = args[1];
@@ -507,7 +494,10 @@ fn reapOsThread(target: *fiber_mod.Fiber, fiber_val: Value) PrimitiveError!Value
         target.os_thread = null;
     }
 
+    target.frame_count = 0;
+
     const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
+    const fiber_obj = types.toObject(fiber_val);
 
     // Remove the thunk and fiber from extra_roots (added by thread-start!
     // to keep them alive while the child runs; the child is done now).
@@ -537,10 +527,13 @@ fn reapOsThread(target: *fiber_mod.Fiber, fiber_val: Value) PrimitiveError!Value
                 }
                 return PrimitiveError.OutOfMemory;
             };
+            gc.writeBarrier(fiber_obj, target.result);
         }
         if (target.status == .errored) {
             if (res.exception) |exc| {
                 target.current_exception = gc.deepCopy(exc) catch null;
+                if (target.current_exception) |cv|
+                    gc.writeBarrier(fiber_obj, cv);
             }
         }
     }
--- a/src/vm_calls.zig
+++ b/src/vm_calls.zig
@@ -125,6 +125,15 @@ pub fn execute(vm: *VM, func: *types.Function) VMError!Value {
     };
     vm.frame_count = 1;
 
+    // Clear local registers to prevent the GC from scanning stale values
+    // left by the previous execute() call. Between calls frame_count is 0,
+    // so a GC triggered by allocClosure above can free objects that stale
+    // registers still point to.
+    const clear_end = @min(@as(usize, func.locals_count), vm.registers.len);
+    if (clear_end > 0) {
+        @memset(vm.registers[0..clear_end], types.UNDEFINED);
+    }
+
     if (vm.profile_mode) {
         vm.profile_time_depth = 1;
         vm.profile_time_stack[0] = .{ .func = func, .entry_ns = clockNs() };
@@ -392,6 +401,16 @@ pub fn callClosure(vm: *VM, closure: *types.Closure, base: u32, nargs: u8) VMErr
     // The callee is in base, args are in base+1..base+nargs
     // New frame's registers start at base (callee reg becomes r0 for the function)
     const new_base = if (base < std.math.maxInt(u32)) base + 1 else return VMError.StackOverflow;
+
+    // Clear local registers beyond the args to prevent the GC from
+    // scanning stale values left by previous frames at this base.
+    const used: usize = if (func.is_variadic) @as(usize, func.arity) + 1 else @as(usize, nargs);
+    const clear_start = @as(usize, new_base) + used;
+    const clear_end = @min(@as(usize, new_base) + @as(usize, func.locals_count), vm.registers.len);
+    if (clear_end > clear_start) {
+        @memset(vm.registers[clear_start..clear_end], types.UNDEFINED);
+    }
+
     vm.frames[vm.frame_count] = .{
         .closure = closure,
         .code = func.code.items,
--- a/src/vm_calls.zig
+++ b/src/vm_calls.zig
@@ -11,6 +11,16 @@ const memory = @import("memory.zig");
 const vm_continuations = @import("vm_continuations.zig");
 const fiber_mod = @import("fiber.zig");
 
+/// Clear local registers beyond `used` args up to `locals_count` to prevent
+/// the GC from scanning stale values left by previously popped frames.
+fn clearFrameLocals(vm: *VM, base: u32, used: usize, locals_count: u16) void {
+    const clear_start = @as(usize, base) + used;
+    const clear_end = @min(@as(usize, base) + @as(usize, locals_count), vm.registers.len);
+    if (clear_end > clear_start) {
+        @memset(vm.registers[clear_start..clear_end], types.UNDEFINED);
+    }
+}
+
 /// Package continuation arguments the same way `values` does:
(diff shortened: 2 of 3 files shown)


baijum (commented):
Reviewed the diff plus the surrounding frame-push and GC-scan code. The fix mechanism is sound — the cleared ranges match `frameWindow()` exactly, the variadic `used = arity + 1` is correct since the rest list is packed at `new_base + arity` before the clear runs, and I verified the removed `makeThreadFn` frame setup is genuinely dead (green fibers get their frames in `fiber.zig`; SRFI-18 fibers only run via `threadEntryFn` on a child VM, and `fiber.thunk` is marked directly by the fiber GC scan, so no rooting is lost).

Two findings:

**1. `callReentrant` has the same stale-register hole (src/vm_calls.zig:555)** — medium

There are three frame-push sites that expose a register window to `markVMRoots`: `execute()`, `callClosure()`, and `callReentrant()`. This PR clears the first two, but `callReentrant` still pushes a frame whose window `[base, base + locals_count)` is scanned by the GC while its callers write only the arg slots:

- `callHandler` writes 1–2 registers (exception handlers, guard)
- `callWithArgs` writes `nargs` registers (parameter converters, native-driven callbacks)
- `callThunk` writes 0–1 registers

`computeReentrantBase` places the frame just past the top frame's window, in a region whose contents are leftovers from previously popped frames. A GC after those frames popped (nothing covered that region, objects freed) followed by a reentrant call re-exposes the dangling pointers to the marker — the same SIGSEGV class as #1156, reachable via e.g. `with-exception-handler` under `-Dgc-stress=true`. Suggest factoring the clearing into a small helper (`clearFrameLocals(vm, base, used, locals_count)`) and calling it from all three sites, which also removes the copy-paste between `execute` and `callClosure`.

**2. Per-call `@memset` on the hottest VM path** — minor

`callClosure` now memsets up to `locals_count - nargs` registers on every closure call. For typical small frames this is cheap, but the test plan has no benchmark numbers — worth one run of `benchmarks/run-benchmarks.sh` (or `zig build bench`) against main to confirm call-heavy workloads (fib-style recursion) don't regress measurably. If they do, an alternative is clearing on frame pop instead (once per window, off the call fast path).

Everything else looks good: the `reapOsThread` write barriers are correctly placed after the deep copies, `target.frame_count = 0` before the copy closes the mid-copy GC window, and the regression test reproduces the original timeout-join pattern.

baijum (commented):
Reviewed the diff plus the surrounding frame-push and GC-scan code. The fix mechanism is sound — the cleared ranges match `frameWindow()` exactly, the variadic `used = arity + 1` is correct since the rest list is packed at `new_base + arity` before the clear runs, and I verified the removed `makeThreadFn` frame setup is genuinely dead (green fibers get their frames in `fiber.zig`; SRFI-18 fibers only run via `threadEntryFn` on a child VM, and `fiber.thunk` is marked directly by the fiber GC scan, so no rooting is lost).

Two findings:

**1. `callReentrant` has the same stale-register hole (src/vm_calls.zig:555)** — medium

There are three frame-push sites that expose a register window to `markVMRoots`: `execute()`, `callClosure()`, and `callReentrant()`. This PR clears the first two, but `callReentrant` still pushes a frame whose window `[base, base + locals_count)` is scanned by the GC while its callers write only the arg slots:

- `callHandler` writes 1–2 registers (exception handlers, guard)
- `callWithArgs` writes `nargs` registers (parameter converters, native-driven callbacks)
- `callThunk` writes 0–1 registers

`computeReentrantBase` places the frame just past the top frame's window, in a region whose contents are leftovers from previously popped frames. A GC after those frames popped (nothing covered that region, objects freed) followed by a reentrant call re-exposes the dangling pointers to the marker — the same SIGSEGV class as #1156, reachable via e.g. `with-exception-handler` under `-Dgc-stress=true`. Suggest factoring the clearing into a small helper (`clearFrameLocals(vm, base, used, locals_count)`) and calling it from all three sites, which also removes the copy-paste between `execute` and `callClosure`.

**2. Per-call `@memset` on the hottest VM path** — minor

`callClosure` now memsets up to `locals_count - nargs` registers on every closure call. For typical small frames this is cheap, but the test plan has no benchmark numbers — worth one run of `benchmarks/run-benchmarks.sh` (or `zig build bench`) against main to confirm call-heavy workloads (fib-style recursion) don't regress measurably. If they do, an alternative is clearing on frame pop instead (once per window, off the call fast path).

Everything else looks good: the `reapOsThread` write barriers are correctly placed after the deep copies, `target.frame_count = 0` before the copy closes the mid-copy GC window, and the regression test reproduces the original timeout-join pattern.

baijum:
Pushed a follow-up commit addressing both review findings:

**1. `callReentrant` stale-register hole (medium)** — Factored the clearing into a `clearFrameLocals(vm, base, used, locals_count)` helper and added it to `callReentrant`, removing the copy-paste between `execute` and `callClosure`. All three frame-push sites are now covered.

**2. Per-call `@memset` overhead (minor)** — Ran `zig build bench`:
```
 depth |  call/cc (full) |  call/ec (escape) | speedup
     0 |   3581 ns       |    160 ns         |  22.4x
    20 |   3609 ns       |    163 ns         |  22.1x
    40 |   4027 ns       |    177 ns         |  22.8x
```
No measurable regression on the call-heavy benchmark. The memset is cheap because `locals_count - nargs` is typically small (a few registers), and the cost is dominated by the frame push and bytecode dispatch.

Also addressed CodeRabbit findings: converted the test to SRFI-64 assertions, and added a write barrier for `fiber.name` in `makeThreadFn`.

Verification: 30/30 gc-stress, 1776/0 Scheme tests, all unit tests pass.

baijum (commented):
Verified commit `1d76d2c` against both findings — both are addressed correctly.

**Finding 1 (callReentrant) — resolved.** The `clearFrameLocals` helper is an exact refactor of the two inline versions (same ranges, same no-op behavior for `locals_count == 0`), and the new `callReentrant` call site derives `used` correctly. I traced it against what every caller actually writes at `base` before the reentrant call:

- `callHandler`: variadic/0 writes 1 reg (rest pair) → `used = 1` ✓; variadic/1 writes 2 (arg + NIL) → `used = 2` ✓; non-variadic writes 1 → `used = arity` ✓ (for the degenerate arity-0 handler this wipes the unread arg slot, which is harmless and arguably more correct)
- `callWithArgs`: non-variadic writes `arity` regs, variadic writes `arity + 1` → matches ✓
- `callThunk`: variadic/0 writes 1 (NIL), plain thunk writes 0 → matches ✓

Placement is right too: after `ensureFrameCapacity` (which can realloc) and with no allocation between the memset and `runUntil`.

**Finding 2 (benchmark) — resolved.** The CI benchmark run on the first commit (which already contained the `callClosure` memset) shows no regression on the call-heavy workloads: fib 0.98x, closures 0.99x, tailcall 1.01x, nqueens 0.90x.

**CodeRabbit items** also check out: the SRFI-64 conversion uses the correct ending pattern (runner grabbed *before* `test-end`, `exit 1` on failures — the known footgun where grabbing it after errors), and `gc.writeBarrier(&fiber.header, args[1])` passes the right container pointer.

**One residual gap noticed while verifying — suggest a follow-up issue, not a blocker.** The three tail-call opcodes (`tail_call` at vm_dispatch.zig:456, the duplicated closure logic in `tail_call_global`, and `tail_call_cc` at vm_dispatch.zig:1137) mutate the frame in place with `frame.closure = closure` after moving only the args to `frame.base`. When the new callee's `locals_count` is larger than the old one's, the GC scan window `[base, base + locals_count)` extends over registers that were never covered by any frame — the same stale-register mechanism, at a fourth site. `self_tail_call` is safe (same closure, window unchanged and continuously covered). Repro shape under `-Dgc-stress=true`: a function with a small frame calls a helper (child frame writes registers above the caller's window), the helper returns, an allocation triggers GC (frees the helper's now-unreachable locals), then the function tail-calls something with a larger frame — the next GC scans the dangling pointers in the window extension. A `clearFrameLocals(vm, frame.base, arg_count, new_func.locals_count)` at each of the three sites would close it. Since this path is pre-existing and untouched by this PR, filing it separately keeps this PR's scope clean — happy either way.

Everything requested in the review is addressed; LGTM once CI is green on `1d76d2c`.

baijum:
Filed the tail-call window extension gap as #1256 — same stale-register class, separate scope.

PR #1257, opened 2026-07-06
Patch datum-label references inside vectors (#1213)

### LINKED ISSUE #1213: [R7RS 2.4] Reader: datum-label references inside vectors are not patched — #0=#(1 #0#) reads () in the referencing slot
**Category:** Reader
**Spec reference:** R7RS §2.4 (Datum labels): "#⟨n⟩= ⟨datum⟩ ... #⟨n⟩# serves as a reference to the labeled datum. Together, these syntaxes permit the notation of structures with shared or circular substructure." Vectors are data; label references inside them must resolve to the labeled object.
**Severity:** wrong-result

**Minimal reproduction:**
```scheme
(define v (read (open-input-string "#0=#(1 #0#)")))
(eq? v (vector-ref v 1))    ;; expected: #t   actual: #f
(vector-ref v 1)            ;; actual: ()  — the placeholder was never patched

;; list references work:
(define l (read (open-input-string "#0=(1 #0# 2)")))
(eq? l (cadr l))            ;; #t
```

**Consequence:** SRFI-38's `read-with-shared-structure` cannot round-trip cyclic or shared vectors even though `write-with-shared-structure` emits correct labels ("#0=#(1 #0#)").

**Analysis:** the reader patches forward references in pair cells but appears not to walk vector slots when resolving label placeholders (src/reader_datum.zig). Cyclic lists through vectors, e.g. `#0=(1 #(#0#))`, are presumably affected the same way.

**Verified:** fresh ReleaseSafe build at 5d0aeeb; dedup: #75 was the GC rooting of the label table, unrelated
**Source location:** src/reader_datum.zig (datum label resolution)
**Disabled test:** tests/scheme/srfi/srfi38.scm (marked ;; FAIL: #this)

baijum:
## Summary

- The reader resolved `#n#` datum-label forward references in pair cells but never walked vector slots, so `#0=#(1 #0#)` left the placeholder pair unpatched in slot 1
- Add `patchPlaceholder` in `reader_datum.zig` — an iterative walk over pairs (car/cdr) and vectors (slots) with a visited set for cycle safety — called for non-pair labeled datums
- Re-enable the SRFI-38 self-referential vector round-trip test and add a dedicated regression test covering direct, nested, and multi-slot self-references

Closes #1213

## Test plan

- [x] `zig build run -- tests/scheme/r7rs/r7rs-tests.scm` — all 1,391 R7RS tests pass
- [x] `zig build run -- tests/scheme/smoke/datum-label-vector.scm` — new regression test (6 assertions)
- [x] `zig build run -- tests/scheme/srfi/srfi38.scm` — 14 pass, 0 fail (re-enabled vector test)

### DIFF
--- a/src/reader_datum.zig
+++ b/src/reader_datum.zig
@@ -67,6 +67,7 @@ fn tokenToValue(self: *Reader, tok: Token) ReadError!Value {
                 return placeholder;
             } else {
                 self.labels[n] = datum;
+                patchPlaceholder(self.gc, datum, placeholder);
                 return datum;
             }
         },
@@ -216,6 +217,53 @@ fn readVector(self: *Reader) ReadError!Value {
     }
 
     return self.gc.allocVector(elems.items) catch return ReadError.OutOfMemory;
+}
+
+/// Walk `datum` and replace every occurrence of `placeholder` with `datum`.
+/// Handles pairs (car/cdr) and vectors (slots). Uses an iterative traversal
+/// with a visited set to tolerate cycles from previously-resolved labels.
+fn patchPlaceholder(gc: *@import("memory.zig").GC, datum: Value, placeholder: Value) void {
+    var stack: std.ArrayList(Value) = .empty;
+    defer stack.deinit(gc.allocator);
+    var visited: std.ArrayList(Value) = .empty;
+    defer visited.deinit(gc.allocator);
+
+    stack.append(gc.allocator, datum) catch return;
+
+    while (stack.items.len > 0) {
+        const val = stack.pop().?;
+
+        for (visited.items) |v| {
+            if (v == val) break;
+        } else {
+            visited.append(gc.allocator, val) catch return;
+
+            if (types.isPair(val)) {
+                if (types.car(val) == placeholder) {
+                    types.setCar(val, datum);
+                } else {
+                    const car = types.car(val);
+                    if (types.isPair(car) or types.isVector(car))
+                        stack.append(gc.allocator, car) catch return;
+                }
+                if (types.cdr(val) == placeholder) {
+                    types.setCdr(val, datum);
+                } else {
+                    const cdr = types.cdr(val);
+                    if (types.isPair(cdr) or types.isVector(cdr))
+                        stack.append(gc.allocator, cdr) catch return;
+                }
+            } else if (types.isVector(val)) {
+                for (types.toVector(val).data) |*slot| {
+                    if (slot.* == placeholder) {
+                        slot.* = datum;
+                    } else if (types.isPair(slot.*) or types.isVector(slot.*)) {
+                        stack.append(gc.allocator, slot.*) catch return;
+                    }
+                }
+            }
+        }
+    }
 }
 
 fn readBytevector(self: *Reader) ReadError!Value {
--- a/src/reader_datum.zig
+++ b/src/reader_datum.zig
@@ -63,11 +63,13 @@ fn tokenToValue(self: *Reader, tok: Token) ReadError!Value {
             self.labels[n] = placeholder;
             if (types.isPair(datum)) {
                 types.setCar(placeholder, types.car(datum));
+                self.gc.writeBarrier(types.toObject(placeholder), types.car(datum));
                 types.setCdr(placeholder, types.cdr(datum));
+                self.gc.writeBarrier(types.toObject(placeholder), types.cdr(datum));
                 return placeholder;
             } else {
                 self.labels[n] = datum;
-                patchPlaceholder(self.gc, datum, placeholder);
+                patchPlaceholder(self.gc, datum, placeholder) catch return ReadError.OutOfMemory;
                 return datum;
             }
         },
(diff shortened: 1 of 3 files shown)


src/reader_datum.zig
```diff
@@ -218,6 +219,53 @@ fn readVector(self: *Reader) ReadError!Value {
...
+        const val = stack.pop().?;
+
+        for (visited.items) |v| {
+            if (v == val) break;
+        } else {
+            visited.append(gc.allocator, val) catch return;
+
+            if (types.isPair(val)) {
+                if (types.car(val) == placeholder) {
+                    types.setCar(val, datum);
```
baijum: **Missing write barriers on the three patch stores** (this `setCar`, the `setCdr` at L250, and the vector-slot store at L259).

`types.setCar`/`setCdr` are raw field writes, so per `.claude/rules/gc-safety.md` each of these needs `gc.writeBarrier(container, new_val)` after it — the same pattern as `set-car!`/`vector-set!` (`primitives.zig:395`/`846`) and, closest analog to this code, the post-hoc vector-slot patching in `bytecode_file.zig:434`.

It's reachable here because the two sides of the store can end up in different generations: minor GCs run mid-read (see the rooting comment in `readVector`), and anything surviving two of them is promoted to old gen (`gc_collect.zig:284-292`). The containers being patched were allocated *before* the vector and are kept alive across those collections via `extra_roots`/`pushRoot` — so they're exactly the objects that get promoted — while `datum` is always the just-allocated, young vector.

Repro sketch: `#0=#((2 . #0#) …enough elements to trigger 2 minor GCs…)` — the inner pair is promoted mid-read, then `setCdr(pair, vector)` creates an old→young edge with no remembered-set entry. If the program keeps only the pair (`(vector-ref v 0)`) and drops the vector, the next minor GC sweeps the vector and `cdr` on the kept pair is a use-after-free (0xAA poison crash in Debug builds). `-Dgc-stress=true` shrinks the repro to a handful of elements.

The fix is local — `gc` is already a parameter:
```zig
types.setCar(val, datum);
gc.writeBarrier(types.toObject(val), datum);
```
and likewise after the `setCdr` and the `slot.* = datum` store.

Note the pre-existing pair branch at L65–66 (`setCar/setCdr(placeholder, …)`) has the same gap — the placeholder is rooted across the whole nested read, so it's the most promotion-prone object of all. Worth fixing while you're here.

src/reader_datum.zig
```diff
@@ -218,6 +219,53 @@ fn readVector(self: *Reader) ReadError!Value {
...
+/// Walk `datum` and replace every occurrence of `placeholder` with `datum`.
+/// Handles pairs (car/cdr) and vectors (slots). Uses an iterative traversal
+/// with a visited set to tolerate cycles from previously-resolved labels.
+fn patchPlaceholder(gc: *@import("memory.zig").GC, datum: Value, placeholder: Value) void {
+    var stack: std.ArrayList(Value) = .empty;
+    defer stack.deinit(gc.allocator);
+    var visited: std.ArrayList(Value) = .empty;
+    defer visited.deinit(gc.allocator);
+
+    stack.append(gc.allocator, datum) catch return;
```
baijum: The `catch return`s here (and at L239/247/254/261) silently abandon patching on OOM: `read` then returns a structure that still contains the raw placeholder pair in whatever slots weren't reached yet — silently wrong data, no error. Everywhere else in this file allocation failure becomes `ReadError.OutOfMemory`; making this `fn patchPlaceholder(...) ReadError!void` and `try`ing it at the call site preserves the "correct datum or error" invariant.

src/reader_datum.zig
```diff
@@ -218,6 +219,53 @@ fn readVector(self: *Reader) ReadError!Value {
...
+
+    while (stack.items.len > 0) {
+        const val = stack.pop().?;
+
+        for (visited.items) |v| {
```
baijum: nit (optional): the visited check is a linear scan, so patching is O(n²) in the number of pairs+vectors inside the labeled datum — noticeable if someone reads a large shared-vector dump (`#0=#(…tens of thousands of elements…)`). An `std.AutoHashMapUnmanaged(Value, void)` keeps it linear.

Relatedly, the walk runs even when the placeholder was never handed out — refs that appear *after* the def completes (e.g. `(#0=#(1 2 3) #0#)`) resolve to the real datum, so the walk is pure overhead in the common shared-but-acyclic case. A per-label "forward ref seen" flag would skip it entirely.

The printer's pass-1 (`printer.zig` `seen` array) uses the same linear-scan shape, so there's no strong precedent either way — fine to punt on both.

baijum (commented):
Reviewed the fix and traced the edge cases — the approach is correct. Pairs and vectors are the only Value-holding containers the reader can produce, a `#n#` ref can only resolve to the placeholder while the labeled datum is still being parsed (so every unpatched ref is reachable from `datum`), and the eagerly-allocated placeholder pair makes the identity comparison safe even when the label is never referenced. The iterative walk with a visited list handles self-refs, nested refs, multi-slot refs, and interleaved labels (`#0=#(#1=(#0#) #1#)`) correctly, and the pair branch still covers vector-in-list refs by identity — nice that the new smoke test pins exactly that case. `run-all.sh` picks up the new smoke file automatically, and the SRFI-64 runner/exit-code pattern in the new test is right.

Three inline comments:

1. **Missing GC write barriers on the three patch stores** — the one thing I'd fix before merge. Mid-read minor GCs can promote the containers being patched while `datum` is always young, so the unbarriered stores create old→young edges the next minor collection can't see (details and a repro sketch inline). The pre-existing pair branch at L65–66 has the same gap and is worth fixing in the same sweep.
2. OOM during patching is silently swallowed (`catch return`), leaving the structure half-patched — the rest of the file propagates `ReadError.OutOfMemory`.
3. (nit, optional) O(n²) visited scan, and the walk runs even when no forward ref was ever handed out.

src/reader_datum.zig
baijum: Addressed in 4e4516d9 — barriers added after all three patch stores, and the pre-existing pair branch got them too (with the precise stored `car`/`cdr` values as `new_val`, which is even tighter than the sketch above). CI green across all matrices including Debug. The same-class gap in `readListTail` is now tracked in #1267. Resolving.

src/reader_datum.zig
baijum: Addressed in 4e4516d9 — `patchPlaceholder` is now `error{OutOfMemory}!void` with `try`-propagated appends, and the call site converts to `ReadError.OutOfMemory`, matching the rest of the file. Resolving.

src/reader_datum.zig
baijum: Addressed in 4e4516d9 — visited is now `std.AutoHashMap(Value, void)`, so patching is linear. The optional forward-ref-seen flag was punted, which is fine as noted. Resolving.

Issue #280, opened 2026-06-28
u8-ready? always returns #t, even at EOF


baijum:
## Bug

`u8-ready?` in `src/primitives_bytevector.zig` (line 346) always returns `#t` due to a copy-paste error — both branches return `types.TRUE`:

```zig
return if (port.string_pos < data.len) types.TRUE else types.TRUE;
//                                                      ^^^^^^^^^ should be types.FALSE
```

## R7RS Reference

Section 6.13.3: `(u8-ready?)` returns `#f` if attempting to read a byte would block or if the port is at EOF.

## Reproduction

```scheme
(let ((p (open-input-bytevector #u8(1 2))))
  (read-u8 p) (read-u8 p)
  (u8-ready? p))
;; Returns #t — should return #f (port exhausted)
```

## Fix

```zig
return if (port.string_pos < data.len) types.TRUE else types.FALSE;
```

## Location

`src/primitives_bytevector.zig` line 346

PR #1258, opened 2026-07-06
Fix u8-ready? returning #f at EOF (R7RS requires #t)

### LINKED ISSUE #1179: [R7RS 6.13.3] u8-ready? returns #f at end of file — spec requires #t (the #280 fix inverted the requirement)
**Spec reference:** R7RS §6.13.3 — `(u8-ready? [port])`: "Returns #t if a byte is ready on the binary input port and returns #f otherwise. If u8-ready? returns #t then the next read-u8 operation on the given port is guaranteed not to hang. **If the port is at end of file then u8-ready? returns #t.**"
**Severity:** wrong-result

**Minimal reproduction:**
```scheme
(u8-ready? (open-input-bytevector #u8()))
;; expected: #t   actual: #f
(let ((p (open-input-bytevector #u8(7)))) (read-u8 p) (u8-ready? p))
;; expected: #t   actual: #f
```

**Expected:** #t — at EOF, `read-u8` returns the eof-object immediately, so it cannot hang; that is exactly the condition u8-ready? reports.
**Actual:** #f once a bytevector/string port is exhausted.

**History:** #280 (closed) filed "u8-ready? always returns #t, even at EOF" as the bug, citing the spec as "returns #f ... if the port is at EOF" — that quote is wrong; R7RS says the opposite. The fix for #280 introduced the current nonconforming branch (`u8ReadyP`: `if (port.string_pos < data.len) TRUE else FALSE`). Note `char-ready?` (src/primitives_io.zig `charReadyP`) always returns #t and is therefore conforming at EOF — the two should end up consistent.

**Verified:** fresh ReleaseSafe build at 2b24cf0
**Source location:** src/primitives_bytevector.zig `u8ReadyP`
**Disabled test:** tests/scheme/audit/primitives_bytevector-audit.scm (marked `;; FAIL: #this`)

**Found by:** Systematic audit, Phase 2.10

baijum:
## Summary

Closes #1179.

- R7RS §6.13.3 requires `u8-ready?` to return `#t` at EOF (since `read-u8` returns the eof-object immediately and cannot hang). The `#280` fix incorrectly inverted this for string ports.
- Removed the string-port branch from `u8ReadyP` so it always returns `#t`, consistent with `char-ready?` which was already conforming.
- Enabled the two disabled audit tests and added a regression test.

## Test plan

- [x] `zig build run -- tests/scheme/smoke/u8-ready-eof.scm` — regression test passes
- [x] `zig build run -- tests/scheme/audit/primitives_bytevector-audit.scm` — 118 pass, 0 fail

### DIFF
--- a/src/primitives_bytevector.zig
+++ b/src/primitives_bytevector.zig
@@ -299,8 +299,8 @@ fn portWriteBytes(port: *types.Port, bytes: []const u8) void {
 }
 
 fn u8ReadyP(args: []const Value) PrimitiveError!Value {
-    const port = try getInputPort(args, 0, "u8-ready?");
-    if (port.peek_byte != null) return types.TRUE;
+    // For simplicity, always return #t (non-blocking check not worth the complexity)
+    _ = try getInputPort(args, 0, "u8-ready?");
     return types.TRUE;
 }
 
--- a/src/primitives_bytevector.zig
+++ b/src/primitives_bytevector.zig
@@ -301,11 +301,7 @@ fn portWriteBytes(port: *types.Port, bytes: []const u8) void {
 fn u8ReadyP(args: []const Value) PrimitiveError!Value {
     const port = try getInputPort(args, 0, "u8-ready?");
     if (port.peek_byte != null) return types.TRUE;
-    if (port.is_string_port) {
-        const data = port.string_data orelse return types.TRUE;
-        return if (port.string_pos < data.len) types.TRUE else types.FALSE;
-    }
-    return types.TRUE; // simplified
+    return types.TRUE;
 }
 
 fn readBytevectorFn(args: []const Value) PrimitiveError!Value {
--- a/tests/scheme/audit/primitives_bytevector-audit.scm
+++ b/tests/scheme/audit/primitives_bytevector-audit.scm
@@ -168,11 +168,8 @@
            (peek-u8 p)
            (u8-ready? p)))
 ;; R7RS 6.13.3: "If the port is at end of file then u8-ready? returns #t"
-;; (read-u8 at EOF returns the eof-object immediately — it cannot hang).
-;; The fix for #280 inverted this; char-ready? conforms, u8-ready? does not.
-;; FAIL: #1179 (u8-ready? returns #f at EOF; spec requires #t)
-;; (test #t (u8-ready? (open-input-bytevector #u8())))
-;; (test #t (let ((p (open-input-bytevector #u8(7)))) (read-u8 p) (u8-ready? p)))
+(test #t (u8-ready? (open-input-bytevector #u8())))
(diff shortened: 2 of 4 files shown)


src/primitives_bytevector.zig
```diff
@@ -301,11 +301,7 @@ fn portWriteBytes(port: *types.Port, bytes: []const u8) void {
 fn u8ReadyP(args: []const Value) PrimitiveError!Value {
     const port = try getInputPort(args, 0, "u8-ready?");
     if (port.peek_byte != null) return types.TRUE;
-    if (port.is_string_port) {
-        const data = port.string_data orelse return types.TRUE;
-        return if (port.string_pos < data.len) types.TRUE else types.FALSE;
-    }
-    return types.TRUE; // simplified
+    return types.TRUE;
```
baijum, line 304: Nit: with the string-port branch gone, the `peek_byte` check on the previous line is dead — both paths now return `types.TRUE`. This could collapse to:

```zig
_ = try getInputPort(args, 0, "u8-ready?");
return types.TRUE;
```

Also, the deletion drops the old `// simplified` marker. Always-`#t` is still a simplification for fd-backed ports (a would-block pipe/tty should strictly return `#f` per R7RS). `charReadyP` in `primitives_io.zig` documents the same choice with a comment ("For simplicity, always return #t...") — worth mirroring that here so a bare `return types.TRUE;` isn't mistaken for full conformance.

baijum:
## Review

The fix itself is correct: R7RS §6.13.3 explicitly requires `#t` at EOF ("If the port is at end of file then u8-ready? returns #t"), the #280 fix was based on a misquoted spec, and unconditional `#t` matches the already-conforming `char-ready?`. The new smoke test follows the house SRFI-64 pattern correctly (runner grabbed before `test-end`, `exit 1` on failure), and `getInputPort` still validates input/open-port errors. Docs at kaappi-lang.org already describe the spec-correct behavior, so nothing needed there.

One real issue found:

**`tests/scheme/smoke/bytevector-port-fixes.scm` (lines 8–16) still documents the inverted #280 behavior this PR reverts.** It's the original regression test for #280:

```scheme
;; ---- #280: u8-ready? returns #f at EOF ----
(let ((p (open-input-bytevector #u8(1 2))))
  (read-u8 p) (read-u8 p)
  (display (u8-ready? p))          ; #f (port exhausted)
  (newline))
```

After this PR that expression prints `#t`, contradicting both the inline comment and the section header — but the file is display-only (no assertions, always exits 0), so `run-all.sh` passes it silently. Suggest fixing it in this PR: either update the section to expect `#t` and cite #1179, or delete the #280 section entirely since the new `u8-ready-eof.scm` supersedes it.

Also left one inline nit on `u8ReadyP` (dead `peek_byte` check + lost `// simplified` comment).

PR #1259, opened 2026-07-06
Remove iteration cap from force trampoline for unbounded delay-force chains

### LINKED ISSUE #1242: [R7RS 4.2.5] force caps promise chains at 100,000 iterations and reports legitimate long delay-force chains as circular

baijum:
## Summary

- Remove the hardcoded 100,000 iteration cap from `force`'s trampoline loop in `primitives_lazy.zig`. R7RS 4.2.5 requires `delay-force` chains to force in bounded space with no limit on chain length — the cap was a heuristic cycle detector that falsely rejected legitimate long chains (SRFI-41 streams, SRFI-45 iterative algorithms).
- The existing `forcing` flag already detects re-entrant promise cycles, so the cap was redundant for cycle detection.
- Enable the previously disabled compliance test for chains beyond 100k.

Fixes #1242

## Test plan

- [x] `r7rs-thin-forms-gaps.scm` — 33/33 pass (was 32, now includes the 100k+ chain test)
- [x] Full R7RS suite — 0 failures
- [x] Verified `(force (chain 100000))` and `(force (chain 200000))` both return `end`
- [x] Verified circular promise `(define p (delay (force p)))` still errors (via native re-entrancy guard)

### DIFF
--- a/src/primitives_lazy.zig
+++ b/src/primitives_lazy.zig
@@ -40,8 +40,7 @@ fn forceFn(args: []const Value) PrimitiveError!Value {
 
     if (!types.isPromise(current)) return current;
 
-    var iterations: usize = 0;
-    while (iterations < 100000) : (iterations += 1) {
+    while (true) {
         if (!types.isPromise(current)) return current;
         const promise = types.toPromise(current);
 
@@ -99,6 +98,4 @@ fn forceFn(args: []const Value) PrimitiveError!Value {
         gc.writeBarrier(&promise.header, result);
         return result;
     }
-
-    return primitives.typeError("force", "non-circular promise chain", args[0]);
 }
--- a/tests/scheme/compliance/r7rs-thin-forms-gaps.scm
+++ b/tests/scheme/compliance/r7rs-thin-forms-gaps.scm
@@ -16,10 +16,7 @@
   (if (= n 0) (delay 'end) (delay-force (df-chain (- n 1)))))
 (test-equal "delay-force chain forces in bounded space" 'end
   (force (df-chain 99999)))
-;; force's trampoline is capped at 100,000 iterations and mislabels longer
-;; legitimate chains as circular
-;; FAIL: #1242 (force reports long delay-force chains as circular)
-;; (test-equal "delay-force chain beyond 100k" 'end (force (df-chain 100000)))
+(test-equal "delay-force chain beyond 100k" 'end (force (df-chain 100001)))
 
 ;; force memoizes: the body runs once even when forced repeatedly
 (define force-count 0)


src/primitives_lazy.zig
```diff
@@ -40,8 +40,7 @@ fn forceFn(args: []const Value) PrimitiveError!Value {
 
     if (!types.isPromise(current)) return current;
 
-    var iterations: usize = 0;
-    while (iterations < 100000) : (iterations += 1) {
+    while (true) {
```
baijum, line 43: **Pre-existing issue in this loop — not introduced by this PR (reproduces on `main`), noting it since it's adjacent to what the PR touches.**

The `if (promise.forced) return promise.value;` check just below doesn't follow promises that were redirected by the merge step further down (`inner.forced = true; inner.value = <outer promise>`). So forcing an *intermediate* promise of an already-completed `delay-force` chain returns the outer promise object instead of the memoized value:

```scheme
(define inner (delay 42))
(define outer (delay-force inner))
(force outer)   ; => 42
(force inner)   ; => #<promise>, expected 42
```

In this implementation a *forced* promise's value can only be another promise via that redirect path (the trampoline flattens promise results before memoizing), so the fix shape is to keep trampolining instead of returning — `if (promise.forced) { current = promise.value; continue; }` — the loop head's `isPromise` check terminates for normal values. This matters for SRFI-41-style stream code that keeps references to intermediate promises.

Fine as a follow-up issue rather than a blocker here.

baijum:
Reviewed against #1242 and R7RS 4.2.5 / SRFI-45 semantics — **the change is correct**, and I could not construct a cycle shape that the cap removal turns into a hang. Verified empirically on this branch (built locally):

| Scenario | Result |
|---|---|
| `(define p (delay-force p))` → `(force p)` | error: `re-entrant forcing of promise` (the `forcing`-flag guard) |
| `(define p (delay (force p)))` → `(force p)` | error: `native re-entrancy too deep` |
| mutual cycle `p = (delay-force q)`, `q = (delay-force p)` | error: `re-entrant forcing of promise` |
| `(force (df-chain 300000))` | `end` |

The trampoline only continues past an iteration when the thunk returns a **fresh** unforced/un-forcing promise, so unbounded looping requires an infinite supply of distinct promises — i.e. a genuinely infinite chain like `(define (loop) (delay-force (loop)))`. That now spins forever (in bounded space) instead of erroring at 100k, which is the behavior R7RS actually sanctions for that program — SRFI-45's leak test 3 is exactly this shape — and matches other Schemes. The old cap was indeed redundant as a cycle detector.

Cross-checks:

- **Single implementation**: `runtime_exports.zig` has no separate force path (native binaries reach this primitive via `kaappi_eval`), and `lib/srfi/45.sld` just re-exports `(scheme lazy)` — nothing else to update.
- **No dangling references**: no test, doc, or source references the removed `non-circular promise chain` error message.
- **Cycle detection stays test-pinned** after the backstop removal: `tests/scheme/compliance/lazy.scm` ("delay-force self-reference raises error") and `tests/scheme/audit/primitives_lazy-audit.scm` (`cyc` / `selfp`) cover the error paths, and `run-all.sh`'s 60s per-file timeout turns any future regression into a visible failure rather than a silent hang.

One **adjacent pre-existing bug** found while verifying (reproduces on `main` too — not caused by this PR): forcing an intermediate promise of a completed `delay-force` chain returns `#<promise>` instead of the memoized value. Details and suggested fix in the inline comment; recommend a follow-up issue rather than expanding this PR.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

src/primitives_lazy.zig
baijum: Filed as #1264.

PR #1260, opened 2026-07-06
Fix parameterize to evaluate all values before binding (#1202)

### LINKED ISSUE #1202: [R7RS 4.2.6 / SRFI-39] parameterize installs bindings sequentially — later value expressions observe earlier bindings
**Category:** Compiler (parameterize form)
**Spec reference:** R7RS §4.2.6: "The ⟨param⟩ and ⟨value⟩ expressions are evaluated in an unspecified order. The ⟨body⟩ is evaluated in a dynamic environment in which calls to the parameters return the results of passing the corresponding values to the conversion procedure..." — evaluation of the value expressions is a phase that precedes the extended dynamic environment; only the body runs under the new bindings. SRFI-39 makes it concrete with a normative example:
```scheme
(parameterize ((radix 8) (prompt (f 10))) (prompt))  ==>  "1010"
```
(f reads radix; the expected "1010" requires f to see the OUTER radix 10, not 8).
**Severity:** wrong-result

**Minimal reproduction:**
```scheme
(define a (make-parameter 1))
(define b (make-parameter 0))
(parameterize ((a 2) (b (a))) (b))   ;; expected: 1   actual: 2
(parameterize ((b (a)) (a 2)) (b))   ;; expected: 1   actual: 1 (order-dependent!)
```
The two forms must agree (evaluation order is unspecified, but no value expression may run under a sibling's new binding). The SRFI-39 spec example returns "12" instead of "1010".

**Root cause:** `compileParameterize` (src/compiler_advanced.zig:676) emits bind-then-evaluate sequentially per clause (let*-style) instead of evaluating all value expressions first and installing the dynamic bindings together.

**Source location:** src/compiler_advanced.zig (`compileParameterize`)
**Disabled test:** tests/scheme/srfi/srfi39.scm (marked ;; FAIL: #this)

**Found by:** Systematic audit, Phase 3.1

baijum:
## Summary

- Fix `compileParameterize` to evaluate all param and value expressions before mutating any parameter cell, per R7RS §4.2.6 and SRFI-39
- Split the single `let*` desugaring into an outer `let` (evaluates all expressions in the original dynamic environment) + inner `let*` (saves old values, installs new converted values sequentially)
- Enable 3 previously disabled SRFI-39 tests including the normative example

Closes #1202

## Test plan

- [x] New Zig unit test: both orderings of `(parameterize ((a 2) (b (a))) (b))` return 1
- [x] SRFI-39 normative example enabled: `(parameterize ((radix 8) (prompt (f 10))) (prompt))` returns "1010"
- [x] SRFI-39 suite: 35 pass, 0 fail (was 32 with 3 disabled)
- [x] Full Scheme test suite: 1776 pass, 0 fail
- [x] Zig unit tests: all pass

### DIFF
--- a/src/compiler_advanced.zig
+++ b/src/compiler_advanced.zig
@@ -676,16 +676,19 @@ fn buildQQListExpr(gc: *@import("memory.zig").GC, quote_sym: Value, list_sym: Va
 /// Compile (parameterize ((param1 val1) (param2 val2) ...) body ...)
 ///
 /// Desugars to:
-///   (let ((old1 (p1)) (new1 (begin (p1 v1) (p1)))
-///          (old2 (p2)) (new2 (begin (p2 v2) (p2))) ...)
-///     (dynamic-wind
-///       (lambda () (%parameter-set! p1 new1) (%parameter-set! p2 new2) ...)
-///       (lambda () body ...)
-///       (lambda () (%parameter-set! p1 old1) (%parameter-set! p2 old2) ...)))
+///   (let ((%pp0 p1) (%pp1 p2) ... (%pv0 v1) (%pv1 v2) ...)
+///     (let* ((old0 (%pp0)) (new0 (begin (%pp0 %pv0) (%pp0)))
+///            (old1 (%pp1)) (new1 (begin (%pp1 %pv1) (%pp1))) ...)
+///       (dynamic-wind
+///         (lambda () (%parameter-set! %pp0 new0) (%parameter-set! %pp1 new1) ...)
+///         (lambda () body ...)
+///         (lambda () (%parameter-set! %pp0 old0) (%parameter-set! %pp1 old1) ...))))
 ///
-/// Value expressions are evaluated once (in the let bindings) with converter
-/// applied. The before-thunk uses %parameter-set! to avoid re-evaluating
-/// expressions or re-applying the converter on continuation re-entry.
+/// The outer `let` evaluates all param and value expressions before any
+/// parameter cell is mutated (R7RS §4.2.6, SRFI-39). The inner `let*`
+/// sequentially saves old values and installs new ones through each
+/// parameter's converter. The before-thunk uses %parameter-set! to avoid
+/// re-applying the converter on continuation re-entry.
 pub fn compileParameterize(self: *Compiler, args: Value, dst: u16, is_tail: bool) CompileError!void {
     if (args == types.NIL) return CompileError.InvalidSyntax;
     const bindings = types.car(args);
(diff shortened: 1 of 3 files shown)


src/compiler_advanced.zig
```diff
@@ -750,33 +758,47 @@ pub fn compileParameterize(self: *Compiler, args: Value, dst: u16, is_tail: bool
...
+            outer_bindings = gc.allocPair(pv_pair, outer_bindings) catch return CompileError.OutOfMemory;
+        }
+        i = binding_count;
+        while (i > 0) {
+            i -= 1;
+            const pp_pair = gc.allocPair(param_syms[i], gc.allocPair(param_exprs[i], types.NIL) catch return CompileError.OutOfMemory) catch return CompileError.OutOfMemory;
+            outer_bindings = gc.allocPair(pp_pair, outer_bindings) catch return CompileError.OutOfMemory;
+        }
+
+        // Inner let*: save old values and compute new (converted) values.
```
baijum, line 782: Non-blocking, pre-existing (verified identical on `main`): installation is still not atomic, because converting a value requires mutating the cell (`(begin (%pp_i %pv_i) (%pp_i))`). The SRFI-39 reference implementation converts **all** values and saves **all** old values before installing any. Three consequences, confirmed against this branch's build:

1. **A later converter observes an earlier binding's new value:**
   ```scheme
   (define a (make-parameter 1))
   (define b (make-parameter 0 (lambda (v) (a))))
   (parameterize ((a 2) (b 99)) (b))  ; => 2 here; SRFI-39 ref impl => 1
   ```
2. **A converter error mid-install permanently leaks an earlier mutation** — the `dynamic-wind` isn't entered yet, so the after-thunk never restores:
   ```scheme
   (define p (make-parameter 1))
   (define w (make-parameter #t (lambda (x) (if (boolean? x) x (error "bad")))))
   (guard (e (#t (p))) (parameterize ((p 2) (w 'oops)) 'x))  ; => 2, p stays 2 forever
   ```
3. **Duplicate params restore the wrong value:** `(parameterize ((q 1) (q 2)) (q))` leaves `q` at 1 instead of its original value, since `old1` is read after `new0` was installed.

Fixing all three needs a convert-without-install primitive (e.g. `%parameter-convert`) so the inner bindings become pure reads/conversions and mutation happens only inside the wind thunks. Fine as a follow-up issue — this PR already fixes the value-expression half, which is what #1202 asked for.

tests/scheme/srfi/srfi39.scm
```diff
@@ -104,11 +104,9 @@
 ;; SRFI-39 (normative example): value expressions are evaluated before any
```
baijum, line 104: Verified the expected value: `(radix 2)` is in effect at this point in the file, so `(f 10)` evaluated in the outer dynamic environment gives `"1010"` — matches the SRFI-39 text. Nice that both binding orderings are covered below.

baijum (commented):
Reviewed by building this branch and probing the new desugaring, alongside the same probes on `main`.

The fix is correct: the outer `let` evaluates every param and value expression before any parameter cell is mutated, which resolves #1202 and makes the SRFI-39 normative example pass. Because all the fresh names are `%`-prefixed and can never appear in user expressions, the transformation is safe regardless of `let` init evaluation order, and nested `parameterize` forms can't collide. The GC handling (symbols allocated up front, pair construction under `no_collect`) follows the existing pattern. Tests are placed per the repo's regression-test rule (Zig unit test + Scheme suite) and cover both binding orderings.

I verified three related edge cases on both this branch and `main` — all behave identically, so **none are regressions from this PR** — but they're worth a follow-up issue since they share one root cause (see inline comment): a later converter observing an earlier binding's new value, a converter error mid-install permanently leaking a mutated parameter, and duplicate-parameter restore.

LGTM.

PR #1261, opened 2026-07-06
Error on unknown identifiers in import only/except/rename

### LINKED ISSUE #1174: [R7RS 5.2] import: only/except/rename accept identifiers not exported by the library (typos import silently)
**Category:** R7RS conformance / VM (library system)
**Spec reference:** R7RS §5.2, page 25 — `only`: "It is an error if any of the listed identifiers are not found in the original set." `except`: "It is an error if any of the listed identifiers are not found in the original set." `rename`: "It is an error if any of the listed ⟨identifier₁⟩s are not found in the original set."
**Severity:** edge-case ("it is an error" — signaling not required, but silence turns import typos into confusing downstream undefined-variable errors)

**Minimal reproduction:**
```scheme
(import (only (srfi 133) totally-bogus-name))     ;; silently succeeds
(import (except (srfi 133) totally-bogus-name))   ;; silently succeeds
(import (rename (srfi 133) (totally-bogus-name tbn))) ;; silently succeeds
(import (only (scheme base) car totally-bogus-name))  ;; silently succeeds (built-ins too)
```

**Expected:** an import-time error naming the unknown identifier (what Chibi and most implementations do)
**Actual:** the import succeeds; the bogus name is simply not bound. This masked a real finding during this audit: `(import (only (srfi 133) vector-fold))` "succeeds" even though `vector-fold` is not implemented (#1172), so probing exports via `only` gives false positives.

**Source location:** src/vm_library.zig (handleImport — only/except/rename filtering)
**Disabled test:** none — an import error aborts file loading, so it can't live in a chibi-test file; the fix PR should add tests/scheme/errors/ shell-style cases asserting non-zero exit for each of the three forms.

**Found by:** Systematic audit, Phase 2.8 (found in passing while probing SRFI-133 exports)

baijum:
## Summary

- Validate that identifiers listed in `only`, `except`, and `rename` import modifiers actually exist in the library's exports, per R7RS §5.2
- Previously these were silently ignored, masking typos and missing exports (e.g. `(import (only (srfi 133) vector-fold))` appeared to succeed even though `vector-fold` isn't implemented)
- Adds clear error messages: `import only: identifier 'foo' not found in library exports`

Closes #1174

## Test plan

- [x] 8 new shell tests in `tests/scheme/errors/import-filter.sh` covering valid and invalid cases for `only`, `except`, and `rename`
- [x] All 1777 existing tests pass (1395 R7RS + 382 Scheme files)
- [x] Unit tests pass (`zig build test`)

### DIFF
--- a/src/ir.zig
+++ b/src/ir.zig
@@ -466,6 +466,12 @@ const other_special_forms = std.StaticStringMap(void).initComptime(.{
     .{ "define-library", {} },
     .{ "include", {} },
     .{ "include-ci", {} },
+    .{ "else", {} },
+    .{ "=>", {} },
+    .{ "_", {} },
+    .{ "...", {} },
+    .{ "unquote", {} },
+    .{ "unquote-splicing", {} },
 });
 
 pub fn isSpecialForm(name: []const u8) bool {
--- a/src/ir.zig
+++ b/src/ir.zig
@@ -461,9 +461,14 @@ const other_special_forms = std.StaticStringMap(void).initComptime(.{
     .{ "call-with-current-continuation", {} },
     .{ "call/cc", {} },
     .{ "eval", {} },
+    .{ "define-record-type", {} },
+    .{ "import", {} },
+    .{ "define-library", {} },
+    .{ "include", {} },
+    .{ "include-ci", {} },
 });
 
-fn isSpecialForm(name: []const u8) bool {
+pub fn isSpecialForm(name: []const u8) bool {
     return sexpr_form_map.get(name) != null or other_special_forms.get(name) != null;
 }
 
--- a/src/tests_libraries.zig
+++ b/src/tests_libraries.zig
@@ -91,6 +91,25 @@ test "import rename" {
     try std.testing.expectEqual(@as(i64, 7), types.toFixnum(r2));
 }
 
+test "import rename with colliding names" {
+    var gc = memory.GC.init(std.testing.allocator);
+    defer gc.deinit();
+    var vm = try th.makeTestVM(&gc);
+    defer vm.deinit();
+
+    _ = try vm.eval(
+        \\(define-library (test rename-coll)
+        \\  (import (scheme base))
+        \\  (export a b)
+        \\  (begin (define (a) 1) (define (b) 2)))
+    );
+    _ = try vm.eval("(import (rename (test rename-coll) (a b) (b c)))");
+    const r1 = try vm.eval("(b)");
+    try std.testing.expectEqual(@as(i64, 1), types.toFixnum(r1));
+    const r2 = try vm.eval("(c)");
+    try std.testing.expectEqual(@as(i64, 2), types.toFixnum(r2));
+}
+
 test "import prefix" {
     var gc = memory.GC.init(std.testing.allocator);
     defer gc.deinit();
--- a/src/tests_libraries.zig
+++ b/src/tests_libraries.zig
@@ -102,6 +102,54 @@ test "import prefix" {
     try std.testing.expectEqual(@as(i64, 7), types.toFixnum(result));
 }
 
+test "import only rejects unknown identifier" {
+    var gc = memory.GC.init(std.testing.allocator);
+    defer gc.deinit();
+    var vm = try th.makeTestVM(&gc);
+    defer vm.deinit();
+
+    _ = try vm.eval(
+        \\(define-library (test only-err)
+        \\  (import (scheme base))
+        \\  (export alpha)
+        \\  (begin (define (alpha) 1)))
+    );
+
+    const r = vm.eval("(import (only (test only-err) alpha bogus))");
+    try std.testing.expectError(th.VMError.CompileError, r);
+}
+
+test "import except rejects unknown identifier" {
+    var gc = memory.GC.init(std.testing.allocator);
+    defer gc.deinit();
+    var vm = try th.makeTestVM(&gc);
+    defer vm.deinit();
+
+    const r = vm.eval("(import (except (scheme base) totally-bogus))");
+    try std.testing.expectError(th.VMError.CompileError, r);
+}
+
+test "import rename rejects unknown identifier" {
+    var gc = memory.GC.init(std.testing.allocator);
+    defer gc.deinit();
+    var vm = try th.makeTestVM(&gc);
+    defer vm.deinit();
+
+    const r = vm.eval("(import (rename (scheme base) (totally-bogus tb)))");
+    try std.testing.expectError(th.VMError.CompileError, r);
+}
+
+test "import only accepts syntax keywords" {
+    var gc = memory.GC.init(std.testing.allocator);
+    defer gc.deinit();
+    var vm = try th.makeTestVM(&gc);
+    defer vm.deinit();
+
+    _ = try vm.eval("(import (only (scheme base) define if car))");
+    const r = try vm.eval("(car (list 42))");
+    try std.testing.expectEqual(@as(i64, 42), types.toFixnum(r));
+}
+
 test "import scheme r5rs exports full R5RS identifier set" {
     // Regression for #813: the built-in (scheme r5rs) stub exported only 4
     // identifiers (null-environment, scheme-report-environment, eval,
--- a/src/vm_library.zig
+++ b/src/vm_library.zig
@@ -780,18 +780,24 @@ fn processImportRename(vm: *VM, target: *std.StringHashMap(Value), args: Value) 
         rename_list = types.cdr(rename_list);
     }
 
-    // Second pass: apply renames by removing old entries and re-inserting under new names.
+    // Second pass: remove all old entries first, then insert under new names.
+    // Renames refer to the original set (parallel semantics), so interleaving
+    // remove/put corrupts colliding renames like (rename lib (a b) (b c)).
+    const PendingRename = struct { new_name: []const u8, value: Value };
+    var pending: std.ArrayList(PendingRename) = .empty;
+    defer pending.deinit(vm.gc.allocator);
     rename_list = renames;
     while (rename_list != types.NIL) {
         const pair = types.car(rename_list);
-        const old_sym = types.car(pair);
-        const new_sym = types.car(types.cdr(pair));
-        const old_name = types.symbolName(old_sym);
-        const new_name = types.symbolName(new_sym);
+        const old_name = types.symbolName(types.car(pair));
+        const new_name = types.symbolName(types.car(types.cdr(pair)));
         if (source.fetchRemove(old_name)) |kv| {
-            source.put(new_name, kv.value) catch return error.OutOfMemory;
+            pending.append(vm.gc.allocator, .{ .new_name = new_name, .value = kv.value }) catch return error.OutOfMemory;
         }
         rename_list = types.cdr(rename_list);
+    }
+    for (pending.items) |p| {
+        source.put(p.new_name, p.value) catch return error.OutOfMemory;
     }
 
     // Import everything (with renames applied).
--- a/src/vm_library.zig
+++ b/src/vm_library.zig
@@ -679,6 +679,9 @@ fn processImportOnly(vm: *VM, target: *std.StringHashMap(Value), args: Value) !v
         const id_name = types.symbolName(id);
         if (source.get(id_name)) |val| {
             importBinding(vm, target, id_name, val) catch return error.OutOfMemory;
+        } else {
+            vm.setErrorDetail("import only: identifier '{s}' not found in library exports", .{id_name});
+            return error.UndefinedVariable;
         }
         id_list = types.cdr(id_list);
     }
@@ -700,7 +703,12 @@ fn processImportExcept(vm: *VM, target: *std.StringHashMap(Value), args: Value) 
         if (!types.isPair(id_list)) return error.InvalidSyntax;
         const id = types.car(id_list);
         if (!types.isSymbol(id)) return error.InvalidSyntax;
-        excluded_list.append(vm.gc.allocator, types.symbolName(id)) catch return error.OutOfMemory;
+        const exc_name = types.symbolName(id);
+        if (!source.contains(exc_name)) {
+            vm.setErrorDetail("import except: identifier '{s}' not found in library exports", .{exc_name});
+            return error.UndefinedVariable;
+        }
+        excluded_list.append(vm.gc.allocator, exc_name) catch return error.OutOfMemory;
         id_list = types.cdr(id_list);
     }
 
@@ -765,7 +773,12 @@ fn processImportRename(vm: *VM, target: *std.StringHashMap(Value), args: Value) 
         if (!types.isPair(new_rest)) return error.InvalidSyntax;
         const new_sym = types.car(new_rest);
         if (!types.isSymbol(old_sym) or !types.isSymbol(new_sym)) return error.InvalidSyntax;
-        rename_old_list.append(vm.gc.allocator, types.symbolName(old_sym)) catch return error.OutOfMemory;
+        const old_name = types.symbolName(old_sym);
+        if (!source.contains(old_name)) {
+            vm.setErrorDetail("import rename: identifier '{s}' not found in library exports", .{old_name});
+            return error.UndefinedVariable;
+        }
+        rename_old_list.append(vm.gc.allocator, old_name) catch return error.OutOfMemory;
         rename_new_list.append(vm.gc.allocator, types.symbolName(new_sym)) catch return error.OutOfMemory;
         rename_list = types.cdr(rename_list);
     }
--- a/src/vm_library.zig
+++ b/src/vm_library.zig
@@ -7,6 +7,7 @@ const file_utils = @import("file_utils.zig");
 const Value = types.Value;
 
 const macro = @import("compiler_macro.zig");
+const ir = @import("ir.zig");
 const vm_mod = @import("vm.zig");
 const VM = vm_mod.VM;
 const VMError = vm_mod.VMError;
@@ -671,17 +672,27 @@ fn processImportOnly(vm: *VM, target: *std.StringHashMap(Value), args: Value) !v
     var source = try resolveImportBindings(vm, inner_set);
     defer source.deinit();
 
+    // Validate all identifiers before importing any (atomic semantics).
     var id_list = ids;
     while (id_list != types.NIL) {
         if (!types.isPair(id_list)) return error.InvalidSyntax;
         const id = types.car(id_list);
         if (!types.isSymbol(id)) return error.InvalidSyntax;
         const id_name = types.symbolName(id);
+        if (!source.contains(id_name) and !ir.isSpecialForm(id_name)) {
+            vm.setErrorDetail("import only: identifier '{s}' not found in import set", .{id_name});
+            return error.UndefinedVariable;
+        }
+        id_list = types.cdr(id_list);
+    }
+
+    // All names validated — now import.
+    id_list = ids;
+    while (id_list != types.NIL) {
+        const id = types.car(id_list);
+        const id_name = types.symbolName(id);
         if (source.get(id_name)) |val| {
             importBinding(vm, target, id_name, val) catch return error.OutOfMemory;
-        } else {
-            vm.setErrorDetail("import only: identifier '{s}' not found in library exports", .{id_name});
-            return error.UndefinedVariable;
         }
         id_list = types.cdr(id_list);
     }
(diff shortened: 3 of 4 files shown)


src/vm_library.zig
```diff
@@ -679,6 +679,9 @@ fn processImportOnly(vm: *VM, target: *std.StringHashMap(Value), args: Value) !v
         const id_name = types.symbolName(id);
         if (source.get(id_name)) |val| {
             importBinding(vm, target, id_name, val) catch return error.OutOfMemory;
+        } else {
+            vm.setErrorDetail("import only: identifier '{s}' not found in library exports", .{id_name});
```
baijum: **Blocking: this rejects legal R7RS imports — syntax keywords are in no export map.**

Built-in libraries' export maps are populated only from primitive procedure specs (`addExportsForLib` in `library.zig`); compiler-recognized syntax lives in `ir.zig`'s `sexpr_form_map`/`other_special_forms` and never appears in any bindings map. R7RS Appendix A lists syntax keywords as exports of `(scheme base)` et al., so all of these — legal, and accepted before this PR — now fail (verified on a build of ef6baf5):

```scheme
(import (only (scheme base) define car))         ; error: 'define' not found
(import (except (scheme base) when))             ; error: 'when' not found
(import (rename (scheme base) (define def)))     ; error: 'define' not found
(import (only (srfi 9) define-record-type))      ; srfi.9 has an EMPTY export map
(import (only (scheme case-lambda) case-lambda)) ; the library's sole export
(import (only (scheme lazy) delay force))        ; error: 'delay' not found
(import (only (srfi 39) parameterize make-parameter))
```

Same root cause, further out:
- A user library legally re-exporting imported syntax — `(define-library (m) (import (scheme base)) (export when f) …)` — silently drops `when` from its export map (no else branch at the export-collection loop, vm_library.zig:863), so `(import (only (m) when f))` errors *against the library's own export list*.
- `--sandbox` skew: sandbox-excluded names make `(except (scheme base) disassemble)` succeed normally but exit 1 under `--sandbox` (defensible semantically, but decide deliberately).
- And the universe is too *large* in the other direction: `resolveImportBindings` builds `source` via `importBinding`, which copies macro free-ref **helper** names into the map (copyTransformerFreeRefs), so non-exported internals pass validation: `(import (only (srfi 64) %test-comp1body))` exits 0 instead of erroring.

The check is right; the set it checks against is neither complete nor exact. Suggest treating `ir.isSpecialForm(name)` (ir.zig:467) as implicitly present during validation (and honoring it in `handleDefineLibrary`'s export collection), or seeding export maps with keyword markers — error only on names that are neither in the map nor compiler syntax.

src/vm_library.zig
```diff
@@ -679,6 +679,9 @@ fn processImportOnly(vm: *VM, target: *std.StringHashMap(Value), args: Value) !v
...
+            return error.UndefinedVariable;
```
baijum: Two propagation problems undercut the feature once this error is raised:

1. **Inside `define-library` the error is swallowed.** Import declarations route through `processLibDeclaration`, which catches the `CompileError` from `handleImportInto`, prints the detail to stderr, and registers the library anyway (~vm_library.zig:1018). Verified: a program whose library does `(import (except (scheme base) bogus-name))` prints the message and **exits 0**. Combined with the keyword issue, a legal `.sld` doing `(import (only (scheme base) define car))` loses its base bindings and later fails with an unrelated-looking `undefined variable 'list'` — exactly the confusing downstream failure #1174 set out to eliminate. Since `.sld` files are where import filters are most used, the enforcement currently only holds at top level (which is all the new tests cover).

2. **A failed set's stale detail suppresses later errors in the same form.** `handleImportInto` continues with the remaining sets after a failure, and the `last_error_detail` already-set guard (vm_library.zig:189) then drops their messages. Verified: `(import (only (scheme base) bogus) (no-such-lib))` prints only the `'bogus'` message — `library not found: (no-such-lib)` is never shown. Clearing the detail per set (or failing fast) fixes the misattribution.

src/vm_library.zig
```diff
@@ -679,6 +679,9 @@ fn processImportOnly(vm: *VM, target: *std.StringHashMap(Value), args: Value) !v
         const id_name = types.symbolName(id);
         if (source.get(id_name)) |val| {
             importBinding(vm, target, id_name, val) catch return error.OutOfMemory;
```
baijum, line 695: `only` commits bindings one-by-one before validating later ids, so a failed import is applied partially. Verified: after

```scheme
(import (only (my lib) alpha bogus-name))  ; errors
(display (alpha))                          ; => 1 — alpha is bound anyway
```

`alpha` stays bound (top-level execution continues after the error; same for a REPL session), and which bindings survive depends on the textual position of the first bad id. `except` and `rename` in this same PR validate the entire list before importing anything — so the three sibling filters get different failure semantics. Scanning the id list first and importing only after all names validate would make `only` atomic and consistent with the other two.

baijum (commented):
Reviewed by building the PR head (ef6baf5) and running repros against it — every finding below is empirically verified, not speculative.

The mechanism and intent are right: R7RS §5.2 backs this, it closes the real audit gap from #1174, the error detail + exit-code plumbing works at top level, `.sld`-exported macros (`cut`, `receive`) still import fine, and `eval`'s `environment` path raises a catchable exception. But as-is the PR both **rejects legal R7RS programs** (syntax keywords are in no export map) and **doesn't enforce the error where filters are most used** (`define-library` import declarations swallow it and exit 0). I'd hold merging until at least the first two inline comments are addressed.

Ranked findings (details inline):
1. Syntax-keyword imports now hard-error (`(only (scheme base) define)`, `(only (srfi 9) define-record-type)`, `(only (scheme case-lambda) case-lambda)`, …) — plus the map *under*-rejects macro-internal helper names — `src/vm_library.zig:683`
2. The error is swallowed inside `define-library` (exit 0) and a failed set's stale detail suppresses later import errors in the same form — `src/vm_library.zig:684`
3. `only` commits partial imports before erroring; `except`/`rename` are atomic — `src/vm_library.zig:681`
4. Message says "library exports" but the check is (correctly) against the resolved import set — misleading for `prefix`/nested sets — `src/vm_library.zig:708`
5. Tests assert exit codes only and miss the keyword/define-library/partial-import cases — `tests/scheme/errors/import-filter.sh:30`
6. Optional: `source.remove`/`fetchRemove` would collapse validation + exclusion and delete the O(n×m) scans — `src/vm_library.zig:707`

src/vm_library.zig
```diff
@@ -700,7 +703,12 @@ fn processImportExcept(vm: *VM, target: *std.StringHashMap(Value), args: Value)
         if (!types.isPair(id_list)) return error.InvalidSyntax;
         const id = types.car(id_list);
         if (!types.isSymbol(id)) return error.InvalidSyntax;
-        excluded_list.append(vm.gc.allocator, types.symbolName(id)) catch return error.OutOfMemory;
+        const exc_name = types.symbolName(id);
+        if (!source.contains(exc_name)) {
+            vm.setErrorDetail("import except: identifier '{s}' not found in library exports", .{exc_name});
```
baijum: Message nit: the check runs against the *resolved inner import set* — which is the correct semantics (R7RS: "not found in the original set") — but the message blames "library exports", which misleads for composed sets:

```scheme
(import (only (prefix (scheme base) b:) car))
; => import only: identifier 'car' not found in library exports
```

`car` *is* a `(scheme base)` export; the prefixed set just spells it `b:car`. "not found in import set" would be accurate in all cases (applies to all three new messages).

src/vm_library.zig
```diff
@@ -700,7 +703,12 @@ fn processImportExcept(vm: *VM, target: *std.StringHashMap(Value), args: Value)
         if (!types.isPair(id_list)) return error.InvalidSyntax;
         const id = types.car(id_list);
         if (!types.isSymbol(id)) return error.InvalidSyntax;
-        excluded_list.append(vm.gc.allocator, types.symbolName(id)) catch return error.OutOfMemory;
+        const exc_name = types.symbolName(id);
+        if (!source.contains(exc_name)) {
```
baijum: Optional simplification: `source` is a local map that's deinit'd on return, so the membership check can *be* the exclusion — `if (!source.remove(exc_name)) { …error… }` validates and excludes in one hash op, and the second loop becomes a plain import of everything remaining, deleting `excluded_list` and its O(n×m) scan. `processImportRename` has the same shape with `fetchRemove(old_name)` (it yields the value the later scan re-finds), which deletes both parallel lists. One nuance to accept: a duplicated id like `(except lib car car)` would then error — R7RS "it is an error" latitude allows that, and it's arguably more consistent with what this PR enforces.

tests/scheme/errors/import-filter.sh
```diff
@@ -0,0 +1,63 @@
...
+        PASS=$((PASS + 1))
+    else
+        echo "FAIL: $label — expected exit $expected, got $status"
+        FAIL=$((FAIL + 1))
+    fi
+}
+
+# --- only: bogus identifier ---
+echo '(import (only (scheme base) totally-bogus-name))' > "$TMPDIR_TESTS/only-bogus.scm"
+assert_exit_code "only: unknown identifier errors" 1 "$KAAPPI" "$TMPDIR_TESTS/only-bogus.scm"
```
baijum, line 46: These assert exit codes only, so they can pass for the wrong reason — any uncaught error exits 1 (if `(srfi 1)` resolution regressed, the "only on SRFI" case would stay green without touching the new validation). `error-format.sh` in this directory already has output-content helpers worth cribbing to pin the message (`import only: identifier 'totally-bogus-name' not found`).

Coverage worth adding while here, per the other threads: a syntax-keyword case (`(only (scheme base) if)` — currently errors), an import filter inside `define-library` (currently exits 0), and the `only` partial-import case. Also: `src/tests_libraries.zig` already has "import only/except/rename" tests (lines 23/52/81) with the exact fixtures and `expectError(th.VMError.UndefinedVariable, …)` pattern — one-line negative assertions there would put these paths under `zig build test`, which is what CI and the Stop hook gate on (the shell suite only runs via run-all.sh).

src/vm_library.zig
```diff
@@ -765,21 +772,32 @@ fn processImportRename(vm: *VM, target: *std.StringHashMap(Value), args: Value)
...
+
+    // Second pass: apply renames by removing old entries and re-inserting under new names.
+    rename_list = renames;
+    while (rename_list != types.NIL) {
+        const pair = types.car(rename_list);
+        const old_sym = types.car(pair);
+        const new_sym = types.car(types.cdr(pair));
+        const old_name = types.symbolName(old_sym);
+        const new_name = types.symbolName(new_sym);
+        if (source.fetchRemove(old_name)) |kv| {
```
baijum, line 794: **Blocking: this breaks swapping renames** (silent wrong result, no error).

`fetchRemove` + `put` mutates `source` in place, so when a later pair's *old* name equals an earlier pair's *new* name, the second `fetchRemove` reads the value the first `put` just inserted:

```scheme
(import (rename (scheme base) (car cdr) (cdr car)))   ; swap
(cdr '(1 2 3))   ; R7RS: 1 (new cdr = old car)
```

A/B verified with behavioral signature checks on both binaries:
| | first commit `ef6baf5` | this commit `572bdc27` |
|---|---|---|
| `(cdr …)/(car …)` | `1/(2 3)` ✓ | `(2 3)/1` ✗ (no swap) |

The original parallel-array loop was correct precisely because it matched against the **unmutated** `source` keys. Suggest building the renamed result into a *separate* map and importing from that, rather than mutating `source`:

```zig
// pass 2: build renamed set without mutating `source`
var renamed = std.StringHashMap(Value).init(vm.gc.allocator);
defer renamed.deinit();
var it = source.iterator();
while (it.next()) |e| {
    var name = e.key_ptr.*;
    // look up e.key in the (old -> new) rename pairs; if found, use new name
    ...
    renamed.put(name, e.value_ptr.*) catch return error.OutOfMemory;
}
// import from `renamed`
```

(A regression test — `(rename (scheme base) (car cdr) (cdr car))` asserting the swap — would lock this down; `import-filter.sh` only covers a single non-swap rename today.)

src/ir.zig
```diff
@@ -461,9 +461,14 @@ const other_special_forms = std.StaticStringMap(void).initComptime(.{
     .{ "call-with-current-continuation", {} },
     .{ "call/cc", {} },
     .{ "eval", {} },
+    .{ "define-record-type", {} },
+    .{ "import", {} },
+    .{ "define-library", {} },
+    .{ "include", {} },
+    .{ "include-ci", {} },
```
baijum, line 468: Minor residual (optional): the whitelist covers the core forms but not R7RS auxiliary syntax. `else`, `=>`, `_`, `...`, `unquote`, `unquote-splicing` are listed as `(scheme base)` exports in R7RS Appendix A, but aren't in either map, so these still error:

```scheme
(import (except (scheme base) else))        ; error: 'else' not found in import set
(import (only (scheme base) =>))            ; error: '=>' not found in import set
```

Much narrower than the original bug and rarely hit in practice — but it's the same class. Adding those six names to `other_special_forms` closes it.

baijum (commented):
Re-reviewed commit 572bdc27 by building it (and the first commit ef6baf5) and running repros against both. The core concerns are genuinely addressed and well-tested — `zig build test` passes, `import-filter.sh` is 20/20, and the existing import smoke/compliance suites stay green:

- **Syntax keywords** now import (`(only (scheme base) define if car)`, `(only (srfi 9) define-record-type)`, `(scheme case-lambda)`, `(scheme lazy)`) via `ir.isSpecialForm` — good call making it public. ✅
- **`define-library` propagation**: the error now propagates and the library is no longer left registered with a partial env (verified: subsequent `(import (test lib))` reports "library not found", exit 1). ✅
- **`only` atomicity**: two-pass validate-then-import; a failed `only` no longer leaks earlier bindings. ✅
- **Message** now reads "not found in import set". ✅
- **Tests**: stderr-content assertions + syntax-keyword + define-library + composed-prefix cases + 4 unit tests. ✅

One regression to fix before merge, plus two minor residuals — details inline.

**Blocking — swapping renames are now silently broken** (`src/vm_library.zig:791`). The `fetchRemove`+`put` rewrite mutates `source` in place, so a swap re-reads the value the first pass just inserted. A/B verified (behavioral signatures confirm the binaries):

```scheme
(import (rename (scheme base) (car cdr) (cdr car)))
(cdr '(1 2 3))   ; correct: 1 (new cdr = old car)
```
- first commit ef6baf5 → `1/(2 3)` ✓
- this commit 572bdc27 → `(2 3)/1` ✗ (no swap — silent wrong result)

That one's on me — I suggested the `remove`/`fetchRemove` simplification without flagging the in-place-mutation hazard for swaps. Sorry for the churn.

Minor residuals (optional, both pre-existing and low-severity): auxiliary syntax still rejected (`src/ir.zig:468`), and macro-internal helper names still pass validation. Neither needs to block.

src/vm_library.zig
baijum: Addressed in 572bdc2 + 93a9c6c.

Made `ir.isSpecialForm` public and check it alongside the source map — names that are compiler syntax pass validation without needing a runtime binding. Added VM-level forms (`define-record-type`, `import`, `define-library`, `include`, `include-ci`) and R7RS auxiliary syntax (`else`, `=>`, `_`, `...`, `unquote`, `unquote-splicing`) to the special form table.

All cases from the original comment verified:

```scheme
(import (only (scheme base) define car))           ; ✓
(import (except (scheme base) when))               ; ✓
(import (rename (scheme base) (define def)))        ; ✓
(import (only (srfi 9) define-record-type))         ; ✓
(import (only (scheme case-lambda) case-lambda))    ; ✓
(import (only (scheme lazy) delay force))           ; ✓
(import (only (srfi 39) parameterize make-parameter)) ; ✓
(import (only (scheme base) else =>))               ; ✓ (93a9c6c)
```

The two residuals noted in the original comment (user-library re-export of syntax, macro free-ref leaking) are pre-existing and out of scope for this PR — they exist independent of the validation change.

src/vm_library.zig
baijum: Both addressed in 572bdc2.

1. **define-library propagation**: Changed the catch-and-continue to `try` — import errors now propagate up and fail the library definition. Verified with a test case (`define-library: import filter error propagates` in import-filter.sh).

2. **Stale detail suppression**: This is a pre-existing issue in `handleImportInto`'s continue-on-error loop, not introduced by this PR. The `last_error_detail_len == 0` guard at line 189 predates this change. Worth fixing separately but out of scope here.

src/vm_library.zig
baijum: Addressed in 572bdc2. Restructured `only` to validate all identifiers in a first pass, then import in a second pass — atomic semantics matching `except`/`rename`.

src/vm_library.zig
baijum: Addressed in 572bdc2. All three messages now say "not found in import set" instead of "not found in library exports".

src/vm_library.zig
baijum: Adopted in 572bdc2. `except` uses `source.remove()` to validate and exclude in one hash op. `rename` uses `source.fetchRemove()` with two-phase semantics (remove all old entries first, then insert under new names — see 5012987 for the parallel-safety fix).

tests/scheme/errors/import-filter.sh
baijum: Addressed in 572bdc2 + 93a9c6c. Tests now:
- Assert stderr content with `assert_stderr_contains` (not just exit codes)
- Cover syntax keywords (4 cases: base keywords, case-lambda, delay/force, define-record-type)
- Cover import filter inside `define-library`
- Cover composed sets (prefix+only valid and invalid)
- Cover auxiliary syntax (else, =>)
- Cover swap rename (93a9c6c)
- 4 unit tests in `tests_libraries.zig` for `zig build test` / CI gate coverage
- 22 shell tests total

src/vm_library.zig
baijum: Already fixed in 5012987 (committed before this comment). The interleaved `fetchRemove`+`put` in 572bdc2 was the bug; 5012987 splits into two phases:

1. Walk rename list, `fetchRemove` each old name → collect `(new_name, value)` into `pending`
2. Walk `pending`, `put` each into `source`

This gives correct parallel semantics — all removes happen against the original map before any puts. Verified the swap case on the current head:

```
$ echo '(import (rename (scheme base) (car cdr) (cdr car))) (import (scheme write)) (display (cdr (quote (1 2 3))))' | zig-out/bin/kaappi
1
```

Added a swap regression test in 93a9c6c (`rename: swap succeeds` in import-filter.sh) and a colliding-rename unit test (`import rename with colliding names` in tests_libraries.zig).

src/ir.zig
baijum: Addressed in 93a9c6c. Added all six auxiliary syntax names (`else`, `=>`, `_`, `...`, `unquote`, `unquote-splicing`) to `other_special_forms`. Test case added (`only: auxiliary syntax (else, =>) accepted`).

src/tests_libraries.zig
```diff
@@ -102,6 +102,54 @@ test "import prefix" {
...
+test "import only accepts syntax keywords" {
+    var gc = memory.GC.init(std.testing.allocator);
+    defer gc.deinit();
+    var vm = try th.makeTestVM(&gc);
+    defer vm.deinit();
+
+    _ = try vm.eval("(import (only (scheme base) define if car))");
+    const r = try vm.eval("(car (list 42))");
+    try std.testing.expectEqual(@as(i64, 42), types.toFixnum(r));
+}
```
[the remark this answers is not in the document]
baijum: Acknowledged — intentionally left as-is. The `th.expectEval*` helpers don't support `expectError` testing, so these tests need direct `vm.eval` access regardless. The existing tests in this file (lines 23-92) all use the same hand-rolled pattern; changing only the new ones would be inconsistent.

src/vm_library.zig
```diff
@@ -765,21 +772,32 @@ fn processImportRename(vm: *VM, target: *std.StringHashMap(Value), args: Value)
...
-        for (0..rename_old_list.items.len) |i| {
-            if (std.mem.eql(u8, entry.key_ptr.*, rename_old_list.items[i])) {
-                imported_name = rename_new_list.items[i];
-                break;
-            }
-        }
-        importBinding(vm, target, imported_name, entry.value_ptr.*) catch return error.OutOfMemory;
+        importBinding(vm, target, entry.key_ptr.*, entry.value_ptr.*) catch return error.OutOfMemory;
     }
 }
```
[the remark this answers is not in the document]
baijum: Already fixed in 5012987 — same underlying bug as the swap comment (id:3529638090). The interleaved `fetchRemove`+`put` from 572bdc2 was replaced with two-phase remove-then-insert. Swap and colliding rename cases both verified and regression-tested.

PR #1262, opened 2026-07-06
Support optional environment-specifier in load (R7RS §6.14)

### LINKED ISSUE #1190: [R7RS 6.14] load: optional environment-specifier argument rejected with an arity error
**Spec reference:** R7RS §6.14, page 59 — `(load filename)` / `(load filename environment-specifier)`: "...evaluates them sequentially in the environment specified by environment-specifier. If environment-specifier is omitted, (interaction-environment) is assumed."
**Severity:** missing-feature

**Minimal reproduction:**
```scheme
(load "program.scm" (interaction-environment))
;; error: 'load': expected 1 arguments, got 2
```

**Expected:** loads the file, evaluating in the given environment.
**Actual:** `load` is registered with `.arity = .{ .exact = 1 }` (src/primitives_r7rs.zig). The single-argument form works correctly (including file-error? on missing files and catchable reader/compile errors).

**Suggested fix:** register `.variadic = 1` and compile via `compileExpressionInEnv` when an environment is supplied (the same split `evalFn` already implements).

**Verified:** fresh ReleaseSafe build at 36edbe0
**Source location:** src/primitives_r7rs.zig `loadFn` + specs entry
**Disabled test:** tests/scheme/audit/primitives_r7rs-audit.scm (marked `;; FAIL: #this`)

**Found by:** Systematic audit, Phase 2.14

baijum:
## Summary

- Change `load`'s arity from `.exact = 1` to `.variadic = 1` to accept the optional environment-specifier argument per R7RS §6.14
- When an environment is supplied, compile expressions via `compileExpressionInEnv` and execute via `callWithArgs` (matching the `eval` pattern)
- Un-skip the disabled audit test for `#1190`, rewritten to verify via `define` side effect
- Add regression test covering single-arg, `interaction-environment`, custom `environment`, and type-error cases

Closes #1190

## Test plan

- [x] New regression test `tests/scheme/smoke/load-env-1190.scm` (4 checks pass)
- [x] Audit test `primitives_r7rs-audit.scm` (42 pass, 0 fail)
- [x] Zig unit tests pass
- [x] R7RS suite (1395 pass, 0 fail)
- [x] Full Scheme test suite (1777 pass, 0 fail)

### DIFF
--- a/src/primitives_r7rs.zig
+++ b/src/primitives_r7rs.zig
@@ -27,7 +27,7 @@ pub const specs = [_]primitives.PrimSpec{
     .{ .name = "interaction-environment", .func = &interactionEnvironmentFn, .arity = .{ .exact = 0 }, .libs = LS.initMany(&.{ .scheme_eval, .scheme_r5rs, .scheme_repl }), .sandbox = false },
     .{ .name = "null-environment", .func = &nullEnvironmentFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_r5rs), .sandbox = false },
     .{ .name = "scheme-report-environment", .func = &schemeReportEnvironmentFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_r5rs), .sandbox = false },
-    .{ .name = "load", .func = &loadFn, .arity = .{ .exact = 1 }, .libs = LS.initMany(&.{ .scheme_load, .scheme_r5rs }), .sandbox = false },
+    .{ .name = "load", .func = &loadFn, .arity = .{ .variadic = 1 }, .libs = LS.initMany(&.{ .scheme_load, .scheme_r5rs }), .sandbox = false },
     .{ .name = "disassemble", .func = &disassembleFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_base), .sandbox = false },
 };
 
@@ -238,6 +238,16 @@ fn loadFn(args: []const Value) PrimitiveError!Value {
     const str = types.toObject(args[0]).as(types.SchemeString);
     const path = str.data[0..str.len];
 
+    // Optional environment-specifier (R7RS §6.14)
+    var env: ?*std.StringHashMap(Value) = null;
+    var env_val: Value = types.VOID;
+    if (args.len > 1) {
+        if (!types.isEnvironment(args[1])) return primitives.typeError("load", "environment", args[1]);
+        const se = types.toEnvironment(args[1]);
+        env = se.env;
+        env_val = args[1];
+    }
+
     const path_z = gc.allocator.dupeZ(u8, path) catch return PrimitiveError.OutOfMemory;
     defer gc.allocator.free(path_z);
 
@@ -278,13 +288,21 @@ fn loadFn(args: []const Value) PrimitiveError!Value {
     while (reader.hasMore() catch return PrimitiveError.TypeError) { // bare-ok: reader error in load
         const expr = reader.readDatum() catch return primitives.typeError("load", "valid datum", args[0]);
 
-        const func = compiler_mod.compileExpressionWithMacros(gc, expr, &vm.macros, vm.globals) catch return primitives.typeError("load", "valid expression", args[0]);
-        {
+        if (env) |e| {
+            const func = compiler_mod.compileExpressionInEnv(gc, expr, &vm.macros, e, env_val) catch return primitives.typeError("load", "valid expression", args[0]);
+            var closure_val = gc.allocClosure(func) catch return PrimitiveError.OutOfMemory;
+            compiler_mod.Compiler.unrootFunction(gc, func);
+            gc.pushRoot(&closure_val);
+            defer gc.popRoot();
+            last_result = vm.callWithArgs(closure_val, &[_]Value{}) catch |err| {
+                return err;
+            };
+        } else {
+            const func = compiler_mod.compileExpressionWithMacros(gc, expr, &vm.macros, vm.globals) catch return primitives.typeError("load", "valid expression", args[0]);
             var func_val = types.makePointer(@ptrCast(func));
             gc.pushRoot(&func_val);
             defer gc.popRoot();
             compiler_mod.Compiler.unrootFunction(gc, func);
-
             last_result = vm.execute(func) catch |err| {
                 return err;
             };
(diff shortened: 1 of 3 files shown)


baijum (commented):
Reviewed the diff — implementation is sound. GC rooting in the env branch matches the `evalFn` pattern, and the environment object stays reachable through the caller's frame registers during the loop, so no rooting hazard. `restricted_env` only affects primitive-inlining decisions, so forward references inside a loaded file still resolve at runtime. No blocking issues; a few non-blocking observations inline.

src/primitives_r7rs.zig
```diff
@@ -278,13 +288,21 @@ fn loadFn(args: []const Value) PrimitiveError!Value {
     while (reader.hasMore() catch return PrimitiveError.TypeError) { // bare-ok: reader error in load
         const expr = reader.readDatum() catch return primitives.typeError("load", "valid datum", args[0]);
 
-        const func = compiler_mod.compileExpressionWithMacros(gc, expr, &vm.macros, vm.globals) catch return primitives.typeError("load", "valid expression", args[0]);
-        {
+        if (env) |e| {
+            const func = compiler_mod.compileExpressionInEnv(gc, expr, &vm.macros, e, env_val) catch return primitives.typeError("load", "valid expression", args[0]);
```
baijum, line 292: `compileExpressionInEnv` merges the compiled expression's macros back into `vm.macros` on the way out. So `(load file (environment '(scheme base)))` isolates `define`s in the custom environment, but any `define-syntax` in the loaded file still leaks into the global macro table. This is inherited from the `eval`-in-env path rather than introduced here, but for `load` a whole file of definitions is the common case, so the asymmetry is more visible. Worth a note or a follow-up issue if full isolation is intended.

src/primitives_r7rs.zig
```diff
@@ -278,13 +288,21 @@ fn loadFn(args: []const Value) PrimitiveError!Value {
     while (reader.hasMore() catch return PrimitiveError.TypeError) { // bare-ok: reader error in load
         const expr = reader.readDatum() catch return primitives.typeError("load", "valid datum", args[0]);
 
-        const func = compiler_mod.compileExpressionWithMacros(gc, expr, &vm.macros, vm.globals) catch return primitives.typeError("load", "valid expression", args[0]);
-        {
+        if (env) |e| {
```
baijum, line 291: This compile → `allocClosure` → `unrootFunction` → `pushRoot` → `callWithArgs` sequence now exists three times in this file (twice in `evalFn`, once here). A small helper like `fn compileAndCall(vm, gc, expr, env, env_val)` covering both the env and global variants would collapse all three and keep the rooting discipline in one place.

src/primitives_r7rs.zig
```diff
@@ -238,6 +238,16 @@ fn loadFn(args: []const Value) PrimitiveError!Value {
     const str = types.toObject(args[0]).as(types.SchemeString);
     const path = str.data[0..str.len];
 
+    // Optional environment-specifier (R7RS §6.14)
+    var env: ?*std.StringHashMap(Value) = null;
+    var env_val: Value = types.VOID;
+    if (args.len > 1) {
+        if (!types.isEnvironment(args[1])) return primitives.typeError("load", "environment", args[1]);
```
baijum, line 245: Two small consistency notes, neither blocking:

1. `load` now raises a type error for a non-environment second argument, but `evalFn` silently ignores one (`args.len > 1 and types.isEnvironment(args[1])` falls through to globals). The strict behavior here is the better one — consider tightening `eval` to match in a follow-up.
2. `.variadic = 1` has no upper bound, so `(load "f" env extra ...)` silently ignores extras. Same convention as `eval`, so fine to leave, just noting it's intentional.

tests/scheme/smoke/load-env-1190.scm
```diff
@@ -0,0 +1,41 @@
...
+(check "single-arg load defines variable" 42 load-test-var-1)
+
+;; Two-argument form with (interaction-environment)
+(with-output-to-file tmp (lambda () (display "(define load-test-var-2 99)")))
+(load tmp (interaction-environment))
+(check "load with interaction-environment" 99 load-test-var-2)
+
+;; Two-argument form with a custom environment — does not leak into globals
+(with-output-to-file tmp (lambda () (display "(define load-test-var-3 77)")))
+(load tmp (environment '(scheme base)))
```
baijum: This case only asserts the negative (the define didn't leak into globals) — it would still pass if the definition were silently dropped altogether. Binding the environment and positively checking it took effect makes the test complete:

```scheme
(define env (environment '(scheme base) '(scheme eval)))
(load tmp env)
(check "load defines into custom env" 77 (eval 'load-test-var-3 env))
```

tests/scheme/smoke/load-env-1190.scm
```diff
@@ -0,0 +1,41 @@
...
+
+;; Non-environment second arg raises an error
+(with-output-to-file tmp (lambda () (display "(+ 1 2)")))
+(check "load with bad env type" 'caught
+  (guard (e (#t 'caught)) (load tmp 42)))
+
+(delete-file tmp)
+
+(display pass) (display " passed, ") (display fail) (display " failed") (newline)
+(when (> fail 0) (exit 1))
```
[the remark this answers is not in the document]
baijum: Done in 7ec3570 — converted to SRFI-64 and added a positive custom-env assertion.

src/primitives_r7rs.zig
baijum: Good observation. This is inherited from the eval path — worth a follow-up issue if full macro isolation is desired.

src/primitives_r7rs.zig
baijum: Agreed the pattern is repeated. Keeping it as-is for this PR to stay scoped to the arity fix; a helper extraction can come separately.

src/primitives_r7rs.zig
baijum: Good catch on the asymmetry — eval silently falling through is arguably a bug. Will file a follow-up to tighten it.

tests/scheme/smoke/load-env-1190.scm
baijum: Good point — added in 7ec3570. The test now loads into a custom env, positively checks the binding via eval, and also verifies it didn't leak into globals.

src/primitives_r7rs.zig
baijum: Filed as #1269 (with a confirmed repro — a `define-syntax` compiled in a custom environment becomes usable at the global top level).

src/primitives_r7rs.zig
baijum: Filed as #1270 — confirmed that `(eval '(+ 1 2) 42)` silently returns `3` instead of raising a type error, unlike `load` on the same misuse.

PR #1263, opened 2026-07-06
Use Unicode derived properties for char classification (#1145)

### LINKED ISSUE #1145: [R7RS 6.6] char-upper-case?/lower-case?/alphabetic? derive from case mappings, not Unicode properties (Lt, ß, ª/º misclassified)
**Spec reference:** R7RS 6.6, p. 45 — "Specifically, they must return #t when applied to characters with the Unicode properties Alphabetic, Numeric_Type=Decimal, White_Space, Uppercase, and Lowercase respectively, and #f when applied to any other Unicode characters. Note that many Unicode characters are alphabetic but neither upper nor lower case."
**Severity:** wrong-result

**Minimal reproduction:**
```scheme
(import (scheme char))
;; U+01C5 Dz-caron is titlecase (Lt) — neither Uppercase nor Lowercase:
(char-upper-case? #\x01C5)   ;; => #t   expected #f
(char-lower-case? #\x01C5)   ;; => #t   expected #f
;; U+00DF sharp-s is Ll (Lowercase) but has no simple uppercase mapping:
(char-lower-case? #\x00DF)   ;; => #f   expected #t
;; U+00AA/U+00BA ordinal indicators are Lo + Other_Lowercase (Alphabetic, Lowercase):
(char-alphabetic? #\x00AA)   ;; => #f   expected #t
(char-lower-case? #\x00AA)   ;; => #f   expected #t
(char-alphabetic? #\x00BA)   ;; => #f   expected #t
```
Also affected: U+1FBC (Lt) reports `char-upper-case?` => #t.

**Root cause:** `src/primitives_char.zig:105-119` — `isUnicodeUppercase` returns #t when the codepoint *has a lowercase mapping* and `isUnicodeLowercase` when it *has an uppercase mapping*, i.e. classification is derived from simple case mappings rather than the Unicode `Uppercase`/`Lowercase` derived properties:

```zig
pub fn isUnicodeUppercase(cp: u21) bool {
    ...
    // Check if this codepoint has a lowercase mapping (it's uppercase)
    if (unicode.findLower(cp) != null) return true;
```

This is exactly the heuristic the spec's note warns about: titlecase (Lt) characters have both mappings (so they classify as both), while cased letters lacking a simple mapping (ß) or cased via `Other_Lowercase` (ª, º) fall through unless hand-listed in the `extra_uppercase`/`extra_lowercase` tables. `char-alphabetic?`'s fallback (`primitives_char.zig:100-101`) inherits the problem because it uses cased-ness as a proxy for Alphabetic.

**What works:** ASCII, common Latin-1 letters (é, À, µ), kra (in extra tables), Roman numerals via extra_uppercase, digit-value for non-ASCII decimal digits, and full-casing string operations (string-upcase "Straße" => "STRASSE" is correct — the string casing path is separate and fine).

**Verified:** fresh ReleaseSafe build at b2317e8 (v0.12.0); deterministic.
**Source location:** `src/primitives_char.zig` `isUnicodeUppercase` / `isUnicodeLowercase` / `isUnicodeAlphabetic`; `src/unicode_tables.zig` extras.
**Disabled test:** `tests/scheme/compliance/r7rs-chars-strings-gaps.scm` (marked `;; FAIL: #this`)

The proper fix is generating Uppercase/Lowercase/Alphabetic property tables from UnicodeData.txt + PropList.txt (DerivedCoreProperties) instead of inferring from case mappings. Phase 2.3 of the audit (#1137) will audit primitives_char.zig in depth; this issue covers the classification root cause found during Phase 1C.

**Found by:** Systematic audit, Phase 1C

baijum:
## Summary

- **Fixed** `char-upper-case?`, `char-lower-case?`, and `char-alphabetic?` to use the Unicode `Uppercase`, `Lowercase`, and `Alphabetic` derived properties (from `DerivedCoreProperties.txt`) instead of inferring classification from case mappings
- **Updated** `tools/gen_unicode_tables.py` to download `DerivedCoreProperties.txt` and generate range-based property tables (651 Uppercase + 671 Lowercase + 733 Alphabetic ranges)
- **Replaced** ~55 lines of hardcoded script block ranges in `isUnicodeLetter` with a single `inRanges()` binary search call
- **Enabled** 7 previously-disabled tests and added an 18-test regression suite

Fixes #1145

## Test plan

- [x] `zig build run -- tests/scheme/smoke/char-unicode-properties.scm` — 18/18 pass
- [x] `zig build run -- tests/scheme/compliance/r7rs-chars-strings-gaps.scm` — 47/47 pass (was 43)
- [x] `zig build run -- tests/scheme/audit/primitives_char-audit.scm` — 62/62 pass (was 59)
- [x] `bash tests/scheme/run-all.sh` — 1777 pass, 0 fail


src/primitives_char.zig
```diff
@@ -44,78 +44,17 @@ pub const specs = [_]primitives.PrimSpec{
...
+    return unicode.inRanges(&unicode.alphabetic_ranges, cp);
 }
 
 pub fn isUnicodeUppercase(cp: u21) bool {
     if (cp <= 127) return std.ascii.isUpper(@intCast(cp));
-    // Check if this codepoint has a lowercase mapping (it's uppercase)
-    if (unicode.findLower(cp) != null) return true;
-    // Check extra uppercase letters (mathematical symbols, etc.)
-    return unicode.containsU21(&unicode.extra_uppercase, cp);
+    return unicode.inRanges(&unicode.uppercase_ranges, cp);
```
baijum, line 52: **Downstream regression: `string-titlecase` word-boundary tracking breaks for titlecase (Lt) characters.**

The new predicate semantics are correct for `char-upper-case?`/`char-lower-case?` — but `primitives_string_ext.zig:638` (SRFI-13 `string-titlecase`) derives its word-boundary state from these two functions:

```zig
const is_cased = pchar.isUnicodeUppercase(cp) or pchar.isUnicodeLowercase(cp);
```

Unicode's *Cased* property is `Lu ∪ Ll ∪ Lt ∪ Other_Uppercase ∪ Other_Lowercase` — titlecase letters are Cased but neither Uppercase nor Lowercase. Before this PR, ǅ (U+01C5, Lt) accidentally passed `isUnicodeUppercase` via the has-lowercase-mapping heuristic (the very bug #1145 fixes); now it is treated as a non-cased word separator, so `word_start` never clears. Verified by building `main` and this branch and running both:

| input | `main` | this PR |
|---|---|---|
| `(string-titlecase "\x01C5;ivota")` | `Ǆivota` | `ǅIvota` — interior `i` spuriously capitalized |
| `(string-titlecase "A\x03A3;\x01C5;")` | `Aσǆ` | `Aςǅ` — final ς emitted although Σ is followed by a letter |

The second row comes from the Greek final-sigma lookahead at `primitives_string_ext.zig:702`, which uses the same upper-or-lower proxy.

Suggested fix (forward, no revert of this function needed): `DerivedCoreProperties.txt` — which `gen_unicode_tables.py` already downloads — also carries the `Cased` property. Emitting a `cased_ranges` table is one key in the `props` dict plus one `write_range_table` call, and the two `is_cased`/`next_is_cased` sites in `primitives_string_ext.zig` switch to `unicode.inRanges(&unicode.cased_ranges, cp)`.

src/primitives_char.zig
```diff
@@ -44,78 +44,17 @@ pub const specs = [_]primitives.PrimSpec{
...
-    // Georgian Mtavruli (uppercase)
-    if (cp >= 0x1C90 and cp <= 0x1CBA) return true;
-    // Greek Extended
-    if (cp >= 0x1F00 and cp <= 0x1FFF) return true;
-    // Cherokee lowercase
-    if (cp >= 0xAB70 and cp <= 0xABBF) return true;
-    // Fall back to Unicode case tables: any cased letter is alphabetic
-    if (isUnicodeUppercase(cp) or isUnicodeLowercase(cp)) return true;
-    return false;
+    return unicode.inRanges(&unicode.alphabetic_ranges, cp);
```
baijum, line 47: **Downstream regression: Greek final-sigma detection in `string-downcase` now misfires on combining marks.**

`string-downcase`'s final-sigma logic in this file uses `isUnicodeLetter` as its "cased" proxy — `next_is_cased` at line 494 and `prev_cased` at line 561. The *Alphabetic* property includes `Other_Alphabetic` combining marks (Hebrew points U+05B0–U+05BD, Indic vowel signs, U+0345, …) that the old hardcoded script ranges excluded — and per Unicode's Final_Sigma rule those characters are case-ignorable: they must be skipped, not treated as cased.

Verified by running `main` vs. this branch:

```scheme
(string-downcase "\x0391;\x03A3;\x05B0;")  ; Α, Σ, hebrew point sheva
; main:    #x3B1 #x3C2 #x5B0  — final ς, correct
; this PR: #x3B1 #x3C3 #x5B0  — medial σ, wrong
```

Controls are identical on both binaries (`"ΑΣ"` → `ας`, `"ΑΣΑ"` → `ασα`), so this is isolated to the newly-Alphabetic marks.

Same root cause and fix direction as the `string-titlecase` comment: these checks want the *Cased* property (ideally with case-ignorable skipping per the Unicode rule), not *Alphabetic*.

baijum (commented):
Reviewed with 8 analysis passes plus behavioral verification (built both `main` and this branch and diffed runtime behavior).

**The core change is solid.** Verified independently:
- The three generated range tables are byte-for-byte identical to ranges parsed directly from Unicode 15.1 `DerivedCoreProperties.txt` (no transcription drift).
- `inRanges` was fuzz- and boundary-tested against a linear-scan reference over all 2,055 ranges — zero mismatches.
- The ASCII fast paths agree with the tables on all 128 codepoints.
- Every re-enabled and new test expectation checked against UCD ground truth; all 471 `extra_uppercase` / 829 `extra_lowercase` codepoints remain covered by the new tables.
- `zig fmt --check` clean; test placement and SRFI-64 ending pattern follow repo conventions.

**However, the predicate semantics change leaks into two unmodified consumers** that used these functions as approximations of Unicode's *Cased* property — see the two inline comments (`string-titlecase` word tracking + its sigma lookahead in `primitives_string_ext.zig`; `string-downcase` final-sigma in `primitives_char.zig`). Both are confirmed regressions vs. `main` with runnable repros. Both are fixable forward by emitting a `cased_ranges` table from the same `DerivedCoreProperties.txt` the generator already downloads — no revert needed.

**Follow-up worth an issue (out of scope for this PR):** `reader.zig:161–198` still carries its own copy of the exact pattern this PR deletes (hardcoded script ranges + `findLower`/`findUpper`/`containsU21(extra_uppercase/extra_lowercase)` fallback), used for identifier lexing and `printer.zig:267` symbol quoting. `char-alphabetic?` and identifier acceptance now newly diverge — e.g. U+02B0 (ʰ): `(char-alphabetic? #\x02B0)` returns `#t` after this PR while `(define xʰy 42)` still fails with `UnexpectedChar` (before the PR both said no). And `extra_uppercase`/`extra_lowercase`/`containsU21` now exist solely for reader.zig. Migrating the reader to the generated tables would close both, and keep the two classifiers from drifting further on each Unicode regeneration.

src/primitives_char.zig
baijum: Fixed in dfec08c. Added `Cased` property from `DerivedCoreProperties.txt` (157 ranges, 4526 codepoints) and introduced `isUnicodeCased` helper. Updated all four sites:

- `primitives_char.zig`: `prev_cased` and `next_is_cased` in `string-downcase` final-sigma
- `primitives_string_ext.zig`: `is_cased` in `string-titlecase` + its final-sigma lookahead

Both repros now produce correct output. Regression tests added to the smoke suite.

src/primitives_char.zig
baijum: Same commit (dfec08c) fixes this. Both `prev_cased` and `next_is_cased` in `string-downcase` now use `isUnicodeCased` which matches the Unicode Cased property (Lu ∪ Ll ∪ Lt ∪ Other_Uppercase ∪ Other_Lowercase) — correctly excludes case-ignorable combining marks like U+05B0.

Re: the reader.zig divergence follow-up — agreed, that's out of scope for this PR. Happy to file a separate issue.

src/primitives_char.zig
baijum: Verified fixed in dfec08c — rebuilt this branch and re-ran the repros:

| input | `main` | dfec08c |
|---|---|---|
| `(string-titlecase "\x01C5;ivota")` | `Ǆivota` | `Ǆivota` ✓ |
| `(string-titlecase "A\x03A3;\x01C5;")` | `Aσǆ` | `Aσǆ` ✓ |

The new `isUnicodeCased`/`cased_ranges` wiring at both sites (line 638 and the sigma lookahead) is exactly the right fix, and the two new smoke assertions cover it (smoke 20/20, compliance 47/47, audit 62/62 on the new head). Original #1145 predicate fixes remain intact.

src/primitives_char.zig
baijum: Verified fixed in dfec08c: `(string-downcase "\x0391;\x03A3;\x05B0;")` → `#x3B1 #x3C2 #x5B0` (final ς) again, controls unchanged.

Bonus improvement from using `Cased` here: `(string-downcase "ΑΣ\x05D0;")` now correctly yields final ς where `main` gave medial σ — Hebrew letters are Alphabetic but not Cased, so the old `isUnicodeLetter` proxy got that case wrong.

The reader.zig follow-up from the review body is now filed as #1268.

PR #1265, opened 2026-07-06
Check lexical bindings when matching syntax-rules literals

### LINKED ISSUE #1139: [R7RS 4.3.2] syntax-rules: literal matching ignores lexical bindings — locally rebound literal still matches
**Spec reference:** R7RS 4.3.2, p. 23 — "An input identifier matches a literal if and only if it is an identifier and either both its occurrence in the macro expression and its occurrence in the macro definition have the same lexical binding, or the two identifiers are the same and both have no lexical binding."
**Severity:** wrong-result (hygiene violation)

**Minimal reproduction:**
```scheme
(define-syntax has-lit
  (syntax-rules (lit)
    ((_ lit) 'is-literal)
    ((_ x)   'not-literal)))

(has-lit lit)                    ;; => is-literal   (correct: both unbound)
(let ((lit 42)) (has-lit lit))   ;; => is-literal   (WRONG: must be not-literal)
```

**Expected:** `not-literal` — the use-site `lit` is let-bound while the definition-site `lit` is unbound, so the literal must not match and the `(_ x)` clause must be taken.
**Actual:** `is-literal` — the literal matches regardless of lexical bindings.

**Verified:** fresh ReleaseSafe build at b2317e8 (v0.12.0); deterministic wrong result.
**Source location:** `src/expander.zig` `matchPattern` — literals are matched by symbol-name equality only (`std.mem.eql` on `symbolName`); no lexical-binding comparison exists.

Note: the R7RS suite's related tests (underscore-in-literals at r7rs-tests.scm:518, bound/free-identifier probe at :585) pass because they don't rebind the literal at the use site.

baijum:
## Summary

Fixes #1139 — `syntax-rules` literal matching now respects lexical bindings per R7RS 4.3.2.

- **Expander**: `matchPattern` checks whether the literal is def-site bound (via `captured_locals`) and whether the input is use-site bound (via new `use_locals` parameter). If binding status differs, the literal doesn't match.
- **Compiler**: `compileDefineSyntax` and the body-scan `define-syntax` path now record `captured_locals`, so macros defined inside `let`/`lambda` bodies know which literals were lexically bound at their definition site.
- **Tests**: Enabled the `FAIL: #1139` marker in `r7rs-expressions-gaps.scm`, added `tests/scheme/hygiene/literal-binding.scm` (6 edge cases), added a Zig unit test.

## Test plan

- [x] `bash tests/scheme/run-all.sh` — 382 pass, 0 fail
- [x] Manual spot-check: `(let ((lit 42)) (has-lit lit))` → `not-literal`

### DIFF
--- a/src/compiler_ir.zig
+++ b/src/compiler_ir.zig
@@ -348,6 +348,44 @@ pub fn compileLambdaWithIR(self: *Compiler, args: Value, dst: u16, name: ?[]cons
     const roots_base = child.gc.extra_roots.items.len;
     defer child.gc.extra_roots.shrinkRetainingCapacity(roots_base);
 
+    // First pass: collect ALL leading define names so that define-syntax
+    // forms see the complete letrec* region via extra_bound.
+    var all_def_names: [MAX_BODY_DEFS][]const u8 = undefined;
+    var all_def_count: usize = 0;
+    {
+        var scan = body;
+        while (scan != types.NIL and types.isPair(scan)) {
+            const expr = types.car(scan);
+            if (!types.isPair(expr)) break;
+            const head = types.car(expr);
+            if (!types.isSymbol(head)) break;
+            const hn = types.symbolName(head);
+            if (std.mem.eql(u8, hn, "define")) {
+                const da = types.cdr(expr);
+                if (da == types.NIL or !types.isPair(da)) break;
+                const tgt = types.car(da);
+                if (types.isSymbol(tgt)) {
+                    if (all_def_count < MAX_BODY_DEFS) {
+                        all_def_names[all_def_count] = types.symbolName(tgt);
+                        all_def_count += 1;
+                    }
+                } else if (types.isPair(tgt)) {
+                    const fn_nm = types.car(tgt);
+                    if (types.isSymbol(fn_nm) and all_def_count < MAX_BODY_DEFS) {
+                        all_def_names[all_def_count] = types.symbolName(fn_nm);
+                        all_def_count += 1;
+                    }
+                } else break;
+            } else if (!std.mem.eql(u8, hn, "define-syntax")) {
+                break;
+            }
+            scan = types.cdr(scan);
+        }
+    }
+
+    const macro_mark = child.beginBodyMacroScope();
+    defer child.endBodyMacroScope(macro_mark) catch {};
+
     var current = body;
     while (current != types.NIL and types.isPair(current)) {
         const expr = types.car(current);
@@ -355,6 +393,26 @@ pub fn compileLambdaWithIR(self: *Compiler, args: Value, dst: u16, name: ?[]cons
         const head = types.car(expr);
         if (!types.isSymbol(head)) break;
         const head_name = types.symbolName(head);
+        if (std.mem.eql(u8, head_name, "define-syntax")) {
+            const ds_args = types.cdr(expr);
+            if (ds_args == types.NIL or !types.isPair(ds_args)) break;
+            const keyword = types.car(ds_args);
+            if (!types.isSymbol(keyword)) break;
+            const ds_rest = types.cdr(ds_args);
+            if (ds_rest == types.NIL or !types.isPair(ds_rest)) break;
+            const transformer_spec = types.car(ds_rest);
+            const tx_val = macro.parseSyntaxRules(&child, transformer_spec, all_def_names[0..all_def_count]) catch break;
+            const ds_name = types.symbolName(keyword);
+            try child.recordBodyMacro(ds_name);
+            const tx_obj = types.toObject(tx_val).as(types.Transformer);
+            if (child.lib_env) |env| {
+                tx_obj.def_env = env;
+                tx_obj.def_env_val = child.lib_env_val;
+            }
+            child.macros.put(ds_name, tx_val) catch return CompileError.OutOfMemory;
+            current = types.cdr(current);
+            continue;
+        }
         if (!std.mem.eql(u8, head_name, "define")) break;
 
         const def_args = types.cdr(expr);
--- a/src/compiler_lambda.zig
+++ b/src/compiler_lambda.zig
@@ -229,7 +229,7 @@ pub fn compileBodyForms(self: *Compiler, body: Value, opts: BodyOpts) CompileErr
             if (ds_rest == types.NIL or !types.isPair(ds_rest)) break;
             const transformer_spec = types.car(ds_rest);
 
-            const transformer = macro.parseSyntaxRules(self, transformer_spec) catch break;
+            const transformer = macro.parseSyntaxRules(self, transformer_spec, def_names_arr[0..def_count]) catch break;
             const name = types.symbolName(keyword);
 
             try self.recordBodyMacro(name);
@@ -239,13 +239,6 @@ pub fn compileBodyForms(self: *Compiler, body: Value, opts: BodyOpts) CompileErr
             if (self.lib_env) |env| {
                 tx.def_env = env;
                 tx.def_env_val = self.lib_env_val;
-            }
-            if (self.locals.items.len > 0) {
-                const caps = self.gc.allocator.alloc(types.CapturedLocal, self.locals.items.len) catch return CompileError.OutOfMemory;
-                for (self.locals.items, 0..) |local, ci| {
-                    caps[ci] = .{ .name = local.name, .slot = local.slot };
-                }
-                tx.captured_locals = caps;
             }
             self.macros.put(name, transformer) catch return CompileError.OutOfMemory;
         } else {
--- a/src/compiler_lambda.zig
+++ b/src/compiler_lambda.zig
@@ -95,7 +95,7 @@ pub fn compileLambda(self: *Compiler, args: Value, dst: u16, name: ?[]const u8) 
 pub fn compileBody(self: *Compiler, body: Value) CompileError!void {
     const saved_body_scope = self.in_body_scope;
     self.in_body_scope = true;
-    const last_dst = try compileBodyForms(self, body, .{});
+    const last_dst = try compileBodyForms(self, body, .{ .handle_define_syntax = true });
     self.in_body_scope = saved_body_scope;
     try self.emitOp(.@"return");
     try self.emitU16(last_dst);
--- a/src/compiler_lambda.zig
+++ b/src/compiler_lambda.zig
@@ -183,6 +183,43 @@ pub fn compileBodyForms(self: *Compiler, body: Value, opts: BodyOpts) CompileErr
     const macro_mark = if (opts.handle_define_syntax) self.beginBodyMacroScope() else 0;
     defer if (opts.handle_define_syntax) self.endBodyMacroScope(macro_mark) catch {};
 
+    // First pass: collect ALL leading define names so that define-syntax
+    // forms (which may appear before or after a define in the same letrec*
+    // region) see the complete set via extra_bound.
+    var all_def_names: [MAX_BODY_DEFS][]const u8 = undefined;
+    var all_def_count: usize = 0;
+    {
+        var scan = body;
+        while (scan != types.NIL and types.isPair(scan)) {
+            const expr = types.car(scan);
+            if (!types.isPair(expr)) break;
+            const head = types.car(expr);
+            if (!types.isSymbol(head)) break;
+            const hn = types.symbolName(head);
+            if (std.mem.eql(u8, hn, "define")) {
+                const da = types.cdr(expr);
+                if (da == types.NIL or !types.isPair(da)) break;
+                const tgt = types.car(da);
+                if (types.isSymbol(tgt)) {
+                    if (all_def_count < MAX_BODY_DEFS) {
+                        all_def_names[all_def_count] = types.symbolName(tgt);
+                        all_def_count += 1;
+                    }
+                } else if (types.isPair(tgt)) {
+                    const fn_name = types.car(tgt);
+                    if (types.isSymbol(fn_name) and all_def_count < MAX_BODY_DEFS) {
+                        all_def_names[all_def_count] = types.symbolName(fn_name);
+                        all_def_count += 1;
+                    }
+                } else break;
+            } else if (!(opts.handle_define_syntax and std.mem.eql(u8, hn, "define-syntax"))) {
+                break;
+            }
+            scan = types.cdr(scan);
+        }
+    }
+
+    // Second pass: collect define inits and process define-syntax forms.
     var current = body;
     while (current != types.NIL and types.isPair(current)) {
         const expr = types.car(current);
@@ -229,7 +266,7 @@ pub fn compileBodyForms(self: *Compiler, body: Value, opts: BodyOpts) CompileErr
             if (ds_rest == types.NIL or !types.isPair(ds_rest)) break;
             const transformer_spec = types.car(ds_rest);
 
-            const transformer = macro.parseSyntaxRules(self, transformer_spec, def_names_arr[0..def_count]) catch break;
+            const transformer = macro.parseSyntaxRules(self, transformer_spec, all_def_names[0..all_def_count]) catch break;
             const name = types.symbolName(keyword);
 
             try self.recordBodyMacro(name);
--- a/src/compiler_lambda.zig
+++ b/src/compiler_lambda.zig
@@ -240,6 +240,13 @@ pub fn compileBodyForms(self: *Compiler, body: Value, opts: BodyOpts) CompileErr
                 tx.def_env = env;
                 tx.def_env_val = self.lib_env_val;
             }
+            if (self.locals.items.len > 0) {
+                const caps = self.gc.allocator.alloc(types.CapturedLocal, self.locals.items.len) catch return CompileError.OutOfMemory;
+                for (self.locals.items, 0..) |local, ci| {
+                    caps[ci] = .{ .name = local.name, .slot = local.slot };
+                }
+                tx.captured_locals = caps;
+            }
             self.macros.put(name, transformer) catch return CompileError.OutOfMemory;
         } else {
             break;
--- a/src/compiler_macro.zig
+++ b/src/compiler_macro.zig
@@ -10,16 +10,19 @@ const Value = types.Value;
 const MAX_MACRO_EXPANSION_DEPTH: u16 = 256;
 const MAX_MACRO_EXPANSION_STEPS: u32 = 10_000;
 
-fn isLexicallyBoundSkipAliases(ctx: ?*const anyopaque, name: []const u8) bool {
+fn resolveLocalSkipAliases(ctx: ?*const anyopaque, name: []const u8) u16 {
     const self: *const Compiler = @ptrCast(@alignCast(ctx.?));
     var comp: ?*const Compiler = self;
     while (comp) |c| {
-        for (c.locals.items) |local| {
-            if (!local.is_global_alias and std.mem.eql(u8, local.name, name)) return true;
+        var i: usize = c.locals.items.len;
+        while (i > 0) {
+            i -= 1;
+            const local = c.locals.items[i];
+            if (!local.is_global_alias and std.mem.eql(u8, local.name, name)) return local.slot;
         }
         comp = c.parent;
     }
-    return false;
+    return expander.LITERAL_UNBOUND;
 }
 
 pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8, transformer: Value, dst: u16, is_tail: bool) CompileError!void {
@@ -144,7 +147,7 @@ pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8, 
     self.gc.no_collect += 1;
     const use_check = expander.UseSiteBindingCheck{
         .ctx = @ptrCast(self),
-        .check_fn = &isLexicallyBoundSkipAliases,
+        .resolve_fn = &resolveLocalSkipAliases,
     };
     const expanded = expander.expandMacro(self.gc, expr, transformer, self.globals, &merged_macros, use_check) catch |err| {
         self.gc.no_collect -= 1;
@@ -381,20 +384,22 @@ pub fn parseSyntaxRules(self: *Compiler, spec: Value, extra_bound: []const []con
     if (custom_ellipsis) |ce| {
         tx.custom_ellipsis = ce;
     }
-    // R7RS 4.3.2: record whether each literal was lexically bound at definition time
+    // R7RS 4.3.2: record each literal's def-site binding slot (0xFFFF = unbound).
+    // Binding identity — not just bound/unbound — is needed so that two
+    // different bindings with the same name don't falsely match.
     if (lit_count > 0) {
-        const bounds = self.gc.allocator.alloc(bool, lit_count) catch return CompileError.OutOfMemory;
+        const slots = self.gc.allocator.alloc(u16, lit_count) catch return CompileError.OutOfMemory;
         for (literals_buf[0..lit_count], 0..) |lv, li| {
-            bounds[li] = if (types.isSymbol(lv)) blk: {
+            slots[li] = if (types.isSymbol(lv)) blk: {
                 const lname = types.symbolName(lv);
-                if (self.isLexicallyBound(lname)) break :blk true;
+                if (self.resolveLocal(lname)) |s| break :blk s;
                 for (extra_bound) |eb| {
-                    if (std.mem.eql(u8, eb, lname)) break :blk true;
+                    if (std.mem.eql(u8, eb, lname)) break :blk expander.LITERAL_BOUND_PENDING;
                 }
-                break :blk false;
-            } else false;
+                break :blk expander.LITERAL_UNBOUND;
+            } else expander.LITERAL_UNBOUND;
         }
-        tx.literal_bound = bounds;
+        tx.literal_bound = slots;
     }
     return tx_val;
 }
--- a/src/compiler_macro.zig
+++ b/src/compiler_macro.zig
@@ -125,12 +125,19 @@ pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8, 
         }
         globals_mod.releaseGlobalsWrite(glk);
     };
+    // Build use-site local names for R7RS 4.3.2 literal binding check
+    var use_local_names: [256][]const u8 = undefined;
+    const use_local_count = @min(self.locals.items.len, 256);
+    for (self.locals.items[0..use_local_count], 0..) |local, li| {
+        use_local_names[li] = local.name;
+    }
+
     // Suppress GC during expansion: the expanded form isn't
     // rooted until pushRoot below, so a collection triggered
     // by allocPair inside expandMacro could free AST nodes
     // that the partially-built result references.
     self.gc.no_collect += 1;
-    const expanded = expander.expandMacro(self.gc, expr, transformer, self.globals, &merged_macros) catch |err| {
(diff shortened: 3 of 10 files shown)


src/compiler_macro.zig
```diff
@@ -205,6 +212,13 @@ pub fn compileDefineSyntax(self: *Compiler, args: Value, dst: u16) CompileError!
         tx.def_env = env;
         tx.def_env_val = self.lib_env_val;
     }
+    if (self.locals.items.len > 0) {
+        const caps = self.gc.allocator.alloc(types.CapturedLocal, self.locals.items.len) catch return CompileError.OutOfMemory;
+        for (self.locals.items, 0..) |local, ci| {
+            caps[ci] = .{ .name = local.name, .slot = local.slot };
+        }
+        tx.captured_locals = caps;
```
baijum: **Regression — populating `captured_locals` here activates the hygiene-injection machinery on every macro use, and its injected `{name, slot}` entries mis-resolve identifiers in expansions.**

`expandAndCompileMacroUse` appends every captured local into `self.locals` before compiling the expansion. Those entries shadow same-named use-site bindings, and in a nested lambda (fresh child Compiler) the def-site slot indexes the *child's* frame:

```scheme
(define (f y)
  (define-syntax m (syntax-rules () ((_ v) v)))
  (let ((y 10)) (m y)))
(display (f 1))        ; main: 10   this PR: 1   ← def-site slot shadows the let binding

(define (g y)
  (define-syntax m (syntax-rules () ((_ v) v)))
  (lambda (q) (m y)))
(display ((g 7) 99))   ; main: 7    this PR: 99  ← def-site slot 0 reads q in the child frame
```

Neither macro even has literals — the check never fires, only the side effect of the now-populated field. Storing the literal-binding fact in its own Transformer field (resolved per literal at `parseSyntaxRules` time) instead of reusing `captured_locals` avoids triggering the injection machinery entirely.

src/compiler_lambda.zig
```diff
@@ -240,6 +240,13 @@ pub fn compileBodyForms(self: *Compiler, body: Value, opts: BodyOpts) CompileErr
                 tx.def_env = env;
                 tx.def_env_val = self.lib_env_val;
             }
+            if (self.locals.items.len > 0) {
+                const caps = self.gc.allocator.alloc(types.CapturedLocal, self.locals.items.len) catch return CompileError.OutOfMemory;
+                for (self.locals.items, 0..) |local, ci| {
+                    caps[ci] = .{ .name = local.name, .slot = local.slot };
+                }
+                tx.captured_locals = caps;
```
baijum: **Regression — `CapturedLocal{name, slot}` loses the box metadata a real `Local` carries, so expansions of body-defined macros read the raw box cell.**

Body `define` locals are boxed (#1249). When the injected entry is used to resolve a template reference, the unwrap is skipped:

```scheme
(define (h3)
  (define x 10)
  (define-syntax getx (syntax-rules () ((_) x)))
  (getx))
(display (h3))   ; main: 10   this PR: (10 . )   ← the box pair itself, same frame!

(define (mk)
  (define n 0)
  (define-syntax bump! (syntax-rules () ((_) (set! n (+ n 1)))))
  (lambda () (bump!) n))
(define c (mk)) (c) (display (c))
; main: 2   this PR: type error in 'arithmetic': expected number, got 0xfffe000000000005
```

Same root cause as the `compileDefineSyntax` site — any fix should cover both snapshot paths (or better, avoid feeding `captured_locals` at all, per the other comment).

src/compiler_lambda.zig
```diff
@@ -240,6 +240,13 @@ pub fn compileBodyForms(self: *Compiler, body: Value, opts: BodyOpts) CompileErr
                 tx.def_env = env;
                 tx.def_env_val = self.lib_env_val;
             }
+            if (self.locals.items.len > 0) {
```
baijum: **Regression — the snapshot happens mid-pre-scan, before sibling internal `define`s are added to `self.locals`, so a literal bound by a sibling body define is misclassified — and the answer depends on textual order.**

```scheme
(let ()
  (define lit 1)
  (define-syntax m (syntax-rules (lit) ((_ lit) 'is-literal) ((_ x) 'not-literal)))
  (display (m lit)))
; R7RS + main: is-literal (both occurrences denote the sibling define)
; this PR: not-literal
```

And in a lambda body, swapping the order of `(define lit 42)` and the `define-syntax` flips the result (`is-literal` vs `not-literal`) because the two registration paths (this pre-scan vs `compileDefineSyntax`) see different snapshots of the same letrec* region. The def-site environment for a body macro should be the *finished* body scope.

src/compiler_macro.zig
```diff
@@ -125,12 +125,19 @@ pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8,
         }
         globals_mod.releaseGlobalsWrite(glk);
     };
+    // Build use-site local names for R7RS 4.3.2 literal binding check
+    var use_local_names: [256][]const u8 = undefined;
+    const use_local_count = @min(self.locals.items.len, 256);
+    for (self.locals.items[0..use_local_count], 0..) |local, li| {
```
baijum: **Regression — `self.locals` contains compiler-injected entries (global-preservation aliases, hygiene appends), and copying it verbatim misreports globals as use-site lexical bindings during nested expansion.**

```scheme
(define lit 5)
(define-syntax has-lit (syntax-rules (lit) ((_ lit) 'is-literal) ((_ x) 'not-literal)))
(define-syntax outer (syntax-rules () ((_) (has-lit lit))))
(display (outer))
; R7RS + main: is-literal (both occurrences denote the same global)
; this PR: not-literal
```

While compiling `outer`'s expansion, `lit` (a non-procedure global free var) gets injected into `self.locals`; the nested `has-lit` expansion then sees it in `use_local_names` and treats the use site as lexically bound. A use-site check that resolves the single matched identifier on demand (and skips alias entries) wouldn't be fooled by expansion-time bookkeeping.

src/compiler_macro.zig
```diff
@@ -125,12 +125,19 @@ pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8,
         }
         globals_mod.releaseGlobalsWrite(glk);
     };
+    // Build use-site local names for R7RS 4.3.2 literal binding check
+    var use_local_names: [256][]const u8 = undefined;
```
baijum: **Spec gap — only the innermost Compiler's locals are visible, so the #1139 bug survives one lambda deep.**

`Compiler.initChild` starts each nested lambda with empty `.locals`, so enclosing bindings reachable via closure are invisible here:

```scheme
(define-syntax has-lit (syntax-rules (lit) ((_ lit) 'is-literal) ((_ x) 'not-literal)))
(let ((lit 42)) ((lambda () (has-lit lit))))
; R7RS: not-literal (lit is lexically bound via the closure)
; this PR: is-literal — same wrong answer the PR fixes at depth 0
```

Any macro use inside an internal helper procedure — the most common code shape — escapes the check. `Compiler.isLexicallyBound` already walks the parent chain; the use-site query should go through that instead of a copy of one frame's array.

src/expander.zig
```diff
@@ -196,15 +198,35 @@ pub const ExpandError = error{
...
         for (literals) |lit| {
             if (types.isSymbol(lit) and std.mem.eql(u8, types.symbolName(lit), name)) {
-                return types.isSymbol(input) and std.mem.eql(u8, types.symbolName(input), name);
+                if (!types.isSymbol(input)) return false;
+                const input_name = types.symbolName(input);
+                if (!std.mem.eql(u8, input_name, name)) return false;
+                // R7RS 4.3.2: match only if both have same binding or both unbound
+                const def_bound = nameInDefLocals(name, def_locals);
+                const use_bound = nameInUseLocals(input_name, use_locals);
+                return def_bound == use_bound;
```
baijum: **Spec gap — comparing two booleans can't express "refer to the same binding": two *different* bindings that share the literal's name both report bound, so the literal still matches.**

```scheme
(let ((lit 1))
  (define-syntax m (syntax-rules (lit) ((_ lit) 'is-literal) ((_ x) 'not-literal)))
  (let ((lit 2)) (m lit)))
; R7RS 4.3.2: not-literal (def-site lit = outer binding, use-site lit = inner binding)
; this PR: is-literal
```

`CapturedLocal` already records `slot` but it's never compared. True same-binding needs identity (e.g. resolve both occurrences to a stable binding id / slot+frame through the compiler's scope model), not bound-vs-unbound agreement.

src/compiler_macro.zig
```diff
@@ -125,12 +125,19 @@ pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8,
...
+    const use_local_count = @min(self.locals.items.len, 256);
```
baijum: **Silent cap — `@min(len, 256)` drops every local past the first 256, silently reverting the check to pre-fix name-only matching.**

Verified: the same `(let ((lit 42)) (has-lit lit))` use site answers `not-literal` under 100 enclosing `let`s but `is-literal` under 300 — the bug class of #1139 reappears past an arbitrary threshold with no diagnostic, which would be brutal to bisect. The def-site snapshot has no such cap (asymmetric). Either fail loudly (`CompileError.InternalLimit`, as other bounded collectors in this file do) or resolve the one matched identifier on demand so no snapshot buffer is needed.

src/compiler_macro.zig
```diff
@@ -205,6 +212,13 @@ pub fn compileDefineSyntax(self: *Compiler, args: Value, dst: u16) CompileError!
         tx.def_env = env;
         tx.def_env_val = self.lib_env_val;
     }
+    if (self.locals.items.len > 0) {
```
baijum: **Nit — third near-verbatim copy of this snapshot block** (pre-existing one in `compileLetSyntax` ~line 270, new twin in `compileBodyForms`, `compiler_lambda.zig:243`). They've already drifted — the `compileLetSyntax` copy carries a redundant inner `if (caps.len > 0)` guard these two omit. Worth extracting a `snapshotCapturedLocals(self, tx)` helper next to `parseSyntaxRules` (compiler_lambda.zig already imports this module), so future changes to what gets captured land in one place.

baijum (commented):
## Review — verified against fresh builds of `241fb0e` (this PR) and `088472e` (main)

The headline #1139 case is genuinely fixed, and the new tests follow the SRFI-64 runner pattern correctly. However, running edge-case programs against both binaries turned up **six behavioral regressions** (programs that work on main and break on this PR) plus three cases where the R7RS 4.3.2 check doesn't apply. Details are in the inline comments; summary matrix (all runs with fresh `.sbc` caches):

| Program shape | R7RS says | main | this PR |
|---|---|---|---|
| `(let ((lit 42)) (has-lit lit))` (headline) | not-literal | is-literal | **not-literal** ✅ |
| body macro `((_) x)` reading a body `define` local | value of `x` | `10` | **`(10 . )`** ❌ corrupt |
| same, used inside a nested lambda | value of `x` | `10` | **`(10 . )`** ❌ |
| body macro doing `(set! n (+ n 1))` via closure | `1 2` | `1 2` | **runtime type error** ❌ |
| plain macro `((_ v) v)`, use-site shadows def-site name | inner binding | `2` | **`1`** ❌ wrong var |
| same, macro use inside inner `lambda (q)` | closed-over `y`=7 | `7` | **`99`** ❌ reads other frame |
| literal bound by *sibling* body `define` | is-literal | is-literal | **not-literal** ❌ |
| global literal reached through nested expansion | is-literal | is-literal | **not-literal** ❌ |
| shadowed rebinding `(let ((lit 1)) … (let ((lit 2)) (m lit)))` | not-literal | is-literal | is-literal (unfixed) |
| `(let ((lit 42)) ((lambda () (has-lit lit))))` | not-literal | is-literal | is-literal (unfixed) |
| use site under >256 locals | not-literal | is-literal | is-literal (silent cap) |

**Root cause of the regressions:** `captured_locals` is not a passive record — `expandAndCompileMacroUse` *acts* on it at every use (temp-global sentinels, appending `{name, slot}` into `self.locals`, `injectHygienicCapturedLocals`). That machinery was only ever exercised by `let-syntax`; populating the field for every body/local `define-syntax` feeds it entries that lack a real `Local`'s metadata (box flag, owning frame, shadowing position), so expansions read raw box cells or the wrong frame's registers.

**Suggested direction:** decouple literal matching from `captured_locals`. The def-site fact is per-literal and known once at `parseSyntaxRules` time (was this literal name lexically bound at definition?) — store it in its own small Transformer field. The use-site fact should be a single on-demand query through the existing parent-chain lookup (`isLexicallyBound` / `resolveLocal`) for the one identifier being matched — no snapshot array, no 256 cap, and it sees closure bindings. Comparing binding *identity* (not two booleans) also fixes the shadowed-rebinding case.

**Test-plan note:** `zig build test` + both suites pass on this PR because no existing test defines a macro inside a body that references a sibling local — the regression cases above would all make good additions to `tests/scheme/hygiene/literal-binding.scm`. (Also beware when spot-checking manually: `.sbc` caches are keyed on source hash + version string, so a file compiled by one build replays identically under the other.)

src/compiler_lambda.zig
```diff
@@ -229,7 +229,7 @@ pub fn compileBodyForms(self: *Compiler, body: Value, opts: BodyOpts) CompileErr
             if (ds_rest == types.NIL or !types.isPair(ds_rest)) break;
             const transformer_spec = types.car(ds_rest);
 
-            const transformer = macro.parseSyntaxRules(self, transformer_spec) catch break;
+            const transformer = macro.parseSyntaxRules(self, transformer_spec, def_names_arr[0..def_count]) catch break;
```
baijum: **Finding 3, residual — the letrec\* forward-reference case still regresses vs main (order-dependent).**

`extra_bound` here is `def_names_arr[0..def_count]` — only the body `define`s scanned *textually before* this `define-syntax`. But internal definitions form a single letrec\* region, so a `define` that appears **after** the macro is still in scope at the macro's definition. The def-site literal-bound flag misses it:

```scheme
(define (f)
  (define-syntax m (syntax-rules (lit) ((_ lit) 'is-literal) ((_ x) 'not-literal)))
  (define lit 42)          ; defined AFTER the macro
  (m lit))
(display (f))
; this PR: not-literal    main & R7RS: is-literal
```

Swapping the two body forms flips the result, so behavior depends on textual order. (The sibling case where the `define` comes *first* — and the `let`-body case — are correctly fixed.) Fix: complete the body pre-scan of all `define` names before resolving any `define-syntax`'s `literal_bound` flags, so `extra_bound` covers the whole letrec\* region rather than only defs-so-far. The consuming loop is `for (extra_bound) |eb|` in `parseSyntaxRules` (compiler_macro.zig:391).

src/expander.zig
```diff
@@ -196,15 +208,21 @@ pub const ExpandError = error{
...
+        for (literals, 0..) |lit, lit_idx| {
             if (types.isSymbol(lit) and std.mem.eql(u8, types.symbolName(lit), name)) {
-                return types.isSymbol(input) and std.mem.eql(u8, types.symbolName(input), name);
+                if (!types.isSymbol(input)) return false;
+                const input_name = types.symbolName(input);
+                if (!std.mem.eql(u8, input_name, name)) return false;
+                // R7RS 4.3.2: match only if both have same binding or both unbound
+                const def_bound = if (lit_idx < literal_bound.len) literal_bound[lit_idx] else false;
+                const use_bound = use_check.isBound(input_name);
+                return def_bound == use_bound;
```
baijum: **Finding 6, not addressed — this compares boundness booleans, not binding identity, so two *different* bindings that share the literal's name still match.**

```scheme
(let ((lit 1))
  (define-syntax m (syntax-rules (lit) ((_ lit) 'is-literal) ((_ x) 'not-literal)))
  (let ((lit 2)) (m lit)))
; this PR: is-literal    R7RS 4.3.2: not-literal (def-site lit and use-site lit are different bindings)
```

`def_bound` (line 223) and `use_bound` (line 224) are both just "is *some* binding named `lit` in scope", so `true == true` matches even though the two occurrences denote distinct bindings. R7RS 4.3.2 requires the two identifiers to *refer to the same binding*. Doing that fully means comparing resolved binding identity at each site (e.g. a stable binding id, or slot+frame/depth) rather than two presence flags. This is the hardest case and reasonable to defer — but the PR doesn't yet implement the "same binding" clause it targets, and there's no test for this case (it would fail). Worth at least a documented known-limitation + a `FAIL:`-marked test so it isn't silently assumed handled.

---
Filed as #1272 (known limitation) so this is tracked independently of the PR.

baijum (commented):
## Follow-up review — redesign verified against fresh builds of `2b14182` (this PR) and `1806384` (main)

The redesign (`captured_locals` reuse → per-literal `literal_bound` flag + on-demand `UseSiteBindingCheck` parent-chain callback) is the right architecture and **resolves 6 of the 8 findings** from the first review. Re-ran every original repro under the A/B protocol (private binary copies, signature-checked, fresh `.sbc` dir per run); `zig build test`, the R7RS suite (0 fail), and the PR's own `hygiene/literal-binding.scm` (now 10 tests) all pass.

| # | First-review finding | Status | Evidence (NEW-PR vs MAIN) |
|---|---|---|---|
| 1 | hygiene machinery shadows use-site vars / wrong frame | ✅ fixed | `(f 1)`→10, `((f 7) 99)`→7, shadow→2 (match main) |
| 2 | box metadata lost → raw box cell / type error | ✅ fixed | `getx`→10, counter→`1 2` (was `(10 . )` / type error) |
| 4 | alias entries misreported in nested expansion | ✅ fixed | global-through-nested-macro → `is-literal` |
| 5 | only innermost frame; bug survived one lambda deep | ✅ fixed **and improves on main** | `(let ((lit 42)) ((lambda () (has-lit lit))))` → `not-literal` (R7RS-correct; main still gives the latent-wrong `is-literal`) |
| 7 | `@min(…,256)` silent cap | ✅ fixed | 300-deep locals → `not-literal` (no cap) |
| 8 | triplicated snapshot block | ✅ fixed | both added copies removed |
| 3 | snapshot mid-pre-scan / order-dependent | ⚠️ **partial** — see inline comment | letrec\* forward-reference case still regresses vs main |
| 6 | boolean compare, not binding identity | ❌ **not addressed** — see inline comment | two different same-named bindings still match |

Two residuals remain, flagged inline below. Both are spec-completeness; finding 3's residual is still a behavioral regression against `main`, finding 6's is reasonable to defer as a documented limitation if you'd rather not thread binding identity now.

Separately, while re-testing I confirmed a **pre-existing** hygiene bug (identical on `main` and this branch, so *not* caused by this PR): a `define-syntax` whose template references a `let`/`lambda`-bound variable from its definition scope fails with `undefined variable '__hyg_N_var'`. Filed as #1271 so it doesn't get conflated with this PR.

src/compiler_lambda.zig
```diff
@@ -229,7 +266,7 @@ pub fn compileBodyForms(self: *Compiler, body: Value, opts: BodyOpts) CompileErr
             if (ds_rest == types.NIL or !types.isPair(ds_rest)) break;
             const transformer_spec = types.car(ds_rest);
 
-            const transformer = macro.parseSyntaxRules(self, transformer_spec) catch break;
+            const transformer = macro.parseSyntaxRules(self, transformer_spec, all_def_names[0..all_def_count]) catch break;
```
baijum, line 269: **Finding 3, still partial — this letrec\* pre-scan only reaches `let`/`let*` bodies; `lambda` and procedure bodies still regress vs main.**

`all_def_names` fixes forward references, but this branch runs only under `opts.handle_define_syntax` (line 260), which is `true` only for `let`/`let*` bodies (compiler_bindings.zig:549). Lambda and procedure bodies call `compileBodyForms(self, body, .{})` at compiler_lambda.zig:98 with the default `handle_define_syntax = false`, so their `define-syntax` falls through to `compileDefineSyntax`, which passes `&.{}` (compiler_macro.zig:220) and resolves the literal against locals that don't yet include the forward `define`. Result — identical bodies disagree by wrapper:

```scheme
(let ()    (define-syntax m (syntax-rules (lit) ((_ lit) 'is-literal) ((_ x) 'not-literal))) (define lit 42) (m lit)) ; => is-literal (fixed)
(define (f) (define-syntax m (syntax-rules (lit) ((_ lit) 'is-literal) ((_ x) 'not-literal))) (define lit 42) (m lit)) (f) ; => not-literal (main: is-literal)
((lambda () (define-syntax m …) (define lit 42) (m lit)))                                                            ; => not-literal (main: is-literal)
```

To be consistent, the same whole-letrec\*-region def-name set needs to feed the `compileDefineSyntax` path too (or lambda/procedure bodies need to route their internal `define-syntax` through the same pre-scan).

src/expander.zig
```diff
@@ -196,15 +211,28 @@ pub const ExpandError = error{
...
+                // or both are unbound.  Compare binding slots, not just
+                // bound-vs-unbound, so two different bindings with the same
+                // name are correctly distinguished.
+                const def_slot = if (lit_idx < literal_bound.len) literal_bound[lit_idx] else LITERAL_UNBOUND;
+                const use_slot = use_check.resolve(input_name);
+                if (def_slot == LITERAL_UNBOUND and use_slot == LITERAL_UNBOUND) return true;
+                if (def_slot == LITERAL_UNBOUND or use_slot == LITERAL_UNBOUND) return false;
+                // BOUND_PENDING: body define not yet allocated — accept any bound use
+                if (def_slot == LITERAL_BOUND_PENDING) return true;
+                return def_slot == use_slot;
```
baijum, line 235: **Finding 6 / #1272, still partial — comparing raw slots is unsound across frames: two different same-named bindings on the same slot number falsely match.**

`def_slot` is captured in the macro's definition frame and `use_slot` is resolved (via the parent chain) at the use site, but slot indices are **per-frame** register numbers, so equal slot numbers in different frames are different bindings. The same-frame case is now correct, but a nested lambda whose parameter reuses the slot index of the def-site binding falsely matches:

```scheme
((lambda (lit)                                  ; outer lit -> slot 0
   (define-syntax m (syntax-rules (lit) ((_ lit) 'yes) ((_ x) 'no)))
   ((lambda (lit) (m lit)) 99)) 1)              ; inner lit -> slot 0 -> collides
; => yes ; R7RS 4.3.2: no (different bindings)
```

Proof it's the slot collision and not merely "cross-frame unhandled": adding one unused leading parameter so the inner `lit` lands on slot 1 flips the result to the correct `no` — same program semantics, different slot number:

```scheme
((lambda (lit) … ((lambda (a lit) (m lit)) 0 99)) 1)   ; inner lit slot 1 -> no  (correct)
```

A sound identity check needs frame identity too — compare a `(frame/scope-depth, slot)` pair or a globally-unique binding id rather than the bare `u16` slot. (Filed the remaining gap on #1272.)

baijum (commented):
## Re-review of `28fa7f9` ("Compare binding identity (slot), not just bound/unbound")

Verified against fresh builds of this commit, the previous head `2b14182`, and current main (`9e04a66`), A/B protocol, all `.sbc` cleared. `zig build test`, R7RS (0 fail), and `hygiene/literal-binding.scm` (now 12 tests) all pass. This commit targets the two residuals from the last review and **partially resolves both** — each has a remaining hole, flagged inline.

**Finding 6 / #1272 (binding identity):** switching to a slot comparison genuinely fixes **same-frame** shadowing — `(let ((lit 1)) (define-syntax …) (let ((lit 2)) (m lit)))` now correctly gives `not-literal`. But raw slot indices are per-frame, so **cross-frame** shadowing via a nested lambda still falsely matches (inline comment on `expander.zig`). Not a regression (main and the boolean version were also wrong here), but the "identity" claim isn't fully met.

**Finding 3 (letrec\* forward reference):** now correct for `let`/`let*` bodies, but `lambda`/procedure bodies still regress vs main (inline comment on `compiler_lambda.zig`).

Behavior matrix (R7RS-expected in the first column):

| Case | R7RS | main | 2b14182 | **28fa7f9** |
|---|---|---|---|---|
| same-frame shadow, different binding | not-literal | is-literal | is-literal | ✅ **not-literal** |
| letrec\* forward-ref in `let` body | is-literal | is-literal | not-literal | ✅ **is-literal** |
| letrec\* forward-ref in `(define (f) …)` body | is-literal | is-literal | not-literal | ❌ **not-literal** |
| letrec\* forward-ref in `(lambda () …)` body | is-literal | is-literal | not-literal | ❌ **not-literal** |
| cross-frame shadow (nested lambda, same slot) | not-literal | is-literal | is-literal | ❌ **is-literal** |

The two tests added in this commit cover only the two fixed cases (`let`-body letrec\* and same-frame shadow); the two ❌ rows above are untested and pass CI silently. Suggest adding both as `FAIL:`-marked or regular tests so the remaining gaps are visible.

src/compiler_ir.zig
```diff
@@ -348,13 +348,71 @@ pub fn compileLambdaWithIR(self: *Compiler, args: Value, dst: u16, name: ?[]cons
...
         const head_name = types.symbolName(head);
+        if (std.mem.eql(u8, head_name, "define-syntax")) {
+            const ds_args = types.cdr(expr);
+            if (ds_args == types.NIL or !types.isPair(ds_args)) break;
+            const keyword = types.car(ds_args);
+            if (!types.isSymbol(keyword)) break;
+            const ds_rest = types.cdr(ds_args);
+            if (ds_rest == types.NIL or !types.isPair(ds_rest)) break;
+            const transformer_spec = types.car(ds_rest);
+            const tx_val = macro.parseSyntaxRules(&child, transformer_spec, all_def_names[0..all_def_count]) catch break;
```
baijum, line 404: This closes the finding-3 residual: `all_def_names` here gives the `compileLambdaWithIR` path the same whole-letrec\*-region def-name set that `let`/`let*` bodies get, so a `define-syntax` in a `lambda`/procedure body now sees a forward `define` as def-site-bound. Verified: `(define (f) (define-syntax m (syntax-rules (lit) ((_ lit) 'is-literal) ((_ x) 'not-literal))) (define lit 42) (m lit))` now returns `is-literal` (was `not-literal` on `28fa7f9`), matching the `let`-body case and main.

baijum (commented):
## Finding 3 resolved as of `4a8501b`

Confirmed against fresh builds of `4a8501b` and current main (`1102eff`), A/B protocol, `.sbc` cleared. The letrec\* forward-reference residual from the last review is **fixed for all body forms** — `lambda`/procedure bodies now get the same whole-region define-name pre-scan that `let`/`let*` bodies already had (new handling in `compileLambdaWithIR`, plus `compileBody` now passing `handle_define_syntax = true`):

| letrec\* forward-ref (define-syntax before the define it references) | R7RS | prev `28fa7f9` | **`4a8501b`** |
|---|---|---|---|
| `(let () …)` | is-literal | is-literal | ✅ is-literal |
| `(define (f) …)` | is-literal | ❌ not-literal | ✅ **is-literal** |
| `(lambda () …)` | is-literal | ❌ not-literal | ✅ **is-literal** |

Full suite is green on this head: `zig build test`, `bash tests/scheme/run-all.sh` (382 files, 0 fail), R7RS (1395, 0 fail), and `hygiene/literal-binding.scm` (13 tests). No regressions in the prior guards, and new adversarial lambda-body cases I ran (interleaved define/define-syntax where one macro calls another, macros with body-local args, recursive-proc + macro) all evaluate correctly. The added lambda-body forward-ref test covers this path.

**Remaining open item:** only the finding 6 / #1272 cross-frame slot collision is still outstanding (untouched by this commit, as expected) — re-confirmed still `yes` where R7RS wants `no`, tracked on the `expander.zig:235` thread and in #1272.

One minor cleanup thought (non-blocking): the ~30-line all-define-names pre-scan now exists in three places (`compileBodyForms` and `compileLambdaWithIR`, plus the snapshot in `compileLetSyntax`) — worth factoring into one helper at some point so the letrec\*-region logic can't drift between the paths.

PR #1273, opened 2026-07-06
Report syntax-error message and irritants

### LINKED ISSUE #1142: [R7RS 4.3.3] syntax-error discards its message and irritants — reports bare error.InvalidSyntax
**Category:** R7RS conformance / diagnostics
**Spec reference:** R7RS 4.3.3, p. 24 — "syntax-error behaves similarly to error (6.11) except that implementations with an expansion pass separate from evaluation should signal an error as soon as syntax-error is expanded. This can be used as a syntax-rules <template> ... which can provide more descriptive error messages. <message> is a string literal, and <args> arbitrary expressions providing additional information."
**Severity:** edge-case (diagnostic quality; the error itself is signaled)

**Minimal reproduction:**
```scheme
(import (scheme base))
(define-syntax bad
  (syntax-rules ()
    ((_ x) (syntax-error "custom msg" x))))
(bad 1)
```

**Expected:** an error report containing "custom msg" (and ideally the irritant `1`), at expansion/compile time.
**Actual:**
```
file.scm:3: compile error: error.InvalidSyntax
```
Exit code is correctly non-zero and the error is correctly raised at compile time, but the message and irritants are discarded — defeating the purpose of syntax-error, which exists precisely to give macro authors descriptive diagnostics.

**Verified:** fresh ReleaseSafe build at b2317e8 (v0.12.0).
**Source location:** `src/compiler.zig:655-656` — `if (std.mem.eql(u8, effective_name, "syntax-error")) return CompileError.InvalidSyntax;` — returns a bare error code without capturing the message datum.
**Disabled test:** none — observing the message requires a subprocess stderr check; the fix should add a case to `tests/scheme/errors/error-format.sh` asserting the message text appears.

baijum:
## Summary
- **Fixes #1142** — `syntax-error` was discarding its message and irritants, reporting only a bare `error.InvalidSyntax`
- Formats the message (display mode) and irritants (write mode) into a threadlocal buffer in the compiler, consumed by `reportCompileError`
- Output now reads `file:line: syntax-error: custom msg 42` instead of `file:line: compile error: error.InvalidSyntax`

## Test plan
- [x] 4 new tests in `tests/scheme/errors/error-format.sh`: message only, message+irritant, from inside a macro, location present
- [x] All 50 error format tests pass
- [x] Zig unit tests pass
- [x] R7RS test suite: 0 failures

### DIFF
--- a/src/compiler.zig
+++ b/src/compiler.zig
@@ -9,8 +9,16 @@ const macro = @import("compiler_macro.zig");
 const ir_mod = @import("ir.zig");
 const compiler_ir = @import("compiler_ir.zig");
 const globals_mod = @import("globals.zig");
+const printer = @import("printer.zig");
 const Value = types.Value;
 const OpCode = types.OpCode;
+
+pub threadlocal var syntax_error_detail: [512]u8 = [_]u8{0} ** 512;
+pub threadlocal var syntax_error_detail_len: usize = 0;
+
+pub fn getSyntaxErrorDetail() []const u8 {
+    return syntax_error_detail[0..syntax_error_detail_len];
+}
 
 pub const CompileError = error{
     OutOfMemory,
@@ -666,7 +674,10 @@ pub const Compiler = struct {
                 if (std.mem.eql(u8, effective_name, "parameterize")) return advanced.compileParameterize(self, args, dst, is_tail);
 
                 // syntax-error
-                if (std.mem.eql(u8, effective_name, "syntax-error")) return CompileError.InvalidSyntax;
+                if (std.mem.eql(u8, effective_name, "syntax-error")) {
+                    formatSyntaxError(args);
+                    return CompileError.InvalidSyntax;
+                }
 
                 // Macro forms (kept in compiler.zig)
                 if (std.mem.eql(u8, effective_name, "define-syntax")) return macro.compileDefineSyntax(self, args, dst);
@@ -777,6 +788,24 @@ pub const Compiler = struct {
         }
     }
 };
+
+fn formatSyntaxError(args: Value) void {
+    var w: std.Io.Writer = .fixed(&syntax_error_detail);
+    var rest = args;
+    // First argument is the message string.
+    if (types.isPair(rest)) {
+        const msg = types.car(rest);
+        printer.printValue(&w, msg, .display) catch {};
+        rest = types.cdr(rest);
+    }
+    // Remaining arguments are irritants.
+    while (types.isPair(rest)) {
+        w.print(" ", .{}) catch {};
+        printer.printValue(&w, types.car(rest), .write) catch {};
+        rest = types.cdr(rest);
+    }
+    syntax_error_detail_len = w.buffered().len;
+}
 
 /// Recursively collect the symbol names that appear as the target of a
 /// `(set! <name> ...)` anywhere in `expr` into `out`. Used to suppress
--- a/src/compiler.zig
+++ b/src/compiler.zig
@@ -840,6 +840,7 @@ fn collectSetTargets(expr: Value, out: *std.StringHashMap(void)) CompileError!vo
 // ---------------------------------------------------------------------------
 
 pub fn compileExpression(gc: *memory.GC, expr: Value) CompileError!*types.Function {
+    syntax_error_detail_len = 0;
     var c = try Compiler.init(gc);
     const roots_base = gc.extra_roots.items.len;
     var ok = false;
@@ -858,6 +859,7 @@ pub fn compileExpressionWithMacros(gc: *memory.GC, expr: Value, vm_macros: *std.
 }
 
 pub fn compileExpressionWithMacrosAt(gc: *memory.GC, expr: Value, vm_macros: *std.StringHashMap(Value), vm_globals: ?*std.StringHashMap(Value), source_line: u32, source_name: ?[]const u8) CompileError!*types.Function {
+    syntax_error_detail_len = 0;
     var c = try Compiler.init(gc);
     const roots_base = gc.extra_roots.items.len;
     c.globals = vm_globals;
@@ -883,6 +885,7 @@ pub fn compileExpressionWithMacrosAt(gc: *memory.GC, expr: Value, vm_macros: *st
 }
 
 pub fn compileExpressionInEnv(gc: *memory.GC, expr: Value, vm_macros: *std.StringHashMap(Value), env: *std.StringHashMap(Value), env_val: Value) CompileError!*types.Function {
+    syntax_error_detail_len = 0;
     var c = try Compiler.init(gc);
     const roots_base = gc.extra_roots.items.len;
     c.globals = env;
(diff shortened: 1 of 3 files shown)


src/toplevel_driver.zig
```diff
@@ -16,9 +17,17 @@ pub fn reportReadError(source_name: []const u8, line: u32, col: u32, err: anyerr
...
 pub fn reportCompileError(source_name: []const u8, line: u32, err: anyerror) void {
-    var buf: [256]u8 = undefined;
-    const s = std.fmt.bufPrint(&buf, "{s}:{d}: compile error: {}\n", .{ source_name, line, err }) catch "compile error\n";
-    writeStderr(s);
+    const detail = compiler_mod.getSyntaxErrorDetail();
+    if (detail.len > 0) {
+        var buf: [768]u8 = undefined;
+        const s = std.fmt.bufPrint(&buf, "{s}:{d}: syntax-error: {s}\n", .{ source_name, line, detail }) catch "compile error\n";
+        writeStderr(s);
+        compiler_mod.syntax_error_detail_len = 0;
```
baijum, line 27: **Stale buffer (confirmed bug).** This is the *only* site that clears `syntax_error_detail_len`. `formatSyntaxError` sets the buffer whenever a `syntax-error` form is compiled, but not every compile error flows through `reportCompileError` — `eval`/`load`/`define-values` convert `CompileError.InvalidSyntax` into `VMError.CompileError` (see `src/vm_eval.zig:95`), which surfaces as a runtime error or is caught by `guard`, so the clear never runs. The buffer then leaks into the next genuine compile error (repro in the summary — `(if)` gets reported with the stale eval message and its irritant).

Suggested fixes, roughly in order of robustness:
1. Clear `syntax_error_detail_len = 0` at the *start* of each top-level compile (before `compileExpressionWithMacros`), so no state can cross forms; or
2. Reuse the VM's existing error-detail mechanism (`vm.getErrorDetail()` / `last_error_detail_len`, already cleared in `reportRuntimeError`) instead of a second free-floating threadlocal, so `eval`-surfaced `syntax-error` also gets a good message; or
3. At minimum, also clear it on the `eval`/`load` conversion path in `vm_eval.zig`.

Worth a regression test for the cross-form case (guard-caught `eval` of `syntax-error`, then an unrelated `(if)`).

src/toplevel_driver.zig
```diff
@@ -16,9 +17,17 @@ pub fn reportReadError(source_name: []const u8, line: u32, col: u32, err: anyerr
 }
 
 pub fn reportCompileError(source_name: []const u8, line: u32, err: anyerror) void {
-    var buf: [256]u8 = undefined;
-    const s = std.fmt.bufPrint(&buf, "{s}:{d}: compile error: {}\n", .{ source_name, line, err }) catch "compile error\n";
-    writeStderr(s);
+    const detail = compiler_mod.getSyntaxErrorDetail();
+    if (detail.len > 0) {
+        var buf: [768]u8 = undefined;
```
baijum: Minor: `buf` is `[768]` but must hold `source_name` (a file path, unbounded) + `": syntax-error: "` + up to 512 bytes of detail + line + `\n`. A deeply-nested path combined with a near-max message overflows 768, `bufPrint` fails, and it silently falls back to the generic `"compile error\n"` — dropping the whole message this PR is trying to surface. Consider sizing from `source_name.len + detail.len + slack`, or writing the prefix and detail in separate `writeStderr` calls the way `reportRuntimeError`'s no-format branch does.

src/compiler.zig
```diff
@@ -778,6 +789,24 @@ pub const Compiler = struct {
...
+        printer.printValue(&w, msg, .display) catch {};
+        rest = types.cdr(rest);
+    }
+    // Remaining arguments are irritants.
+    while (types.isPair(rest)) {
+        w.print(" ", .{}) catch {};
+        printer.printValue(&w, types.car(rest), .write) catch {};
+        rest = types.cdr(rest);
+    }
+    syntax_error_detail_len = w.buffered().len;
```
baijum, line 807: Minor: if the message + irritants exceed the 512-byte `syntax_error_detail` buffer, the `.fixed` writer returns `WriteFailed`, the `catch {}` swallows it, and `w.buffered()` is whatever fit — which can cut off mid-UTF-8 and emit an invalid sequence in the diagnostic. Not a crash, but a truncated/garbled message. A larger buffer or an explicit `"..."` truncation marker at a codepoint boundary would be safer. Low priority.

baijum (commented):
Nice fix for #1142 — the message/irritants now come through, and formatting them display-mode + write-mode matches how runtime `error` objects render (`error: msg here 1 "two" 3`), so it's consistent. Tests are reasonable.

One correctness issue worth addressing before merge: the threadlocal detail buffer can go **stale** because it's set unconditionally in the compiler but cleared in only one consumer (`reportCompileError`). Any compile-error path that doesn't reach that consumer — `eval`, `load`, or a `syntax-error` caught by `guard` — leaves the buffer dirty, and the *next* unrelated compile error then misreports the stale message. Confirmed repro below. Two smaller buffer-sizing notes inline as well.

Repro (built from this branch):
```scheme
(guard (e (#t (display "caught\n")))
  (eval '(syntax-error "STALE" 999) (environment '(scheme base))))
(if)   ; unrelated malformed if on the next line
```
Actual output:
```
caught
file.scm:4: syntax-error: STALE 999   ; <- wrong: (if) is not a syntax-error, and 999 is from the eval'd form
```
Expected: the generic `compile error: error.InvalidSyntax` for `(if)`.

baijum:
## Re-review of `5a0e393` — both issues verified fixed ✅

I checked out the branch, rebuilt, and re-ran the scenarios (deleting stale `.sbc` caches between runs so I wasn't fooled by cached bytecode).

**1. Stale threadlocal buffer — fixed.** The buffer is now cleared (`syntax_error_detail_len = 0`) at the start of all compile entry points (`compileExpression`, `compileExpressionWithMacrosAt` — which `compileExpressionWithMacros` delegates to — and `compileExpressionInEnv`), so no compile can inherit detail from a prior one. Verified with the original repro:
```
caught eval error
file.scm:4: compile error: error.InvalidSyntax   ← generic, no leak
```
Clearing at the source (rather than only at the consumer) is the robust choice. The added regression test covers the guard/eval cross-form case.

**2. `[768]` report-buffer overflow — fixed.** `reportCompileError` now writes the prefix, detail, and newline as separate `writeStderr` calls, so a long source path can no longer push the message out of a single fixed buffer.

**Verification:**
- error-format suite: 51/51 pass (incl. the new stale-buffer test)
- `zig build test`: pass
- R7RS-relevant paths unaffected; CI green across all platforms; PR is `MERGEABLE`/`CLEAN`

**Remaining (non-blocking, optional):** the 512-byte `syntax_error_detail` buffer still silently truncates via `catch {}` if a message + irritants exceed 512 bytes, and can cut mid-UTF-8. This only bites on unusually long diagnostics and degrades gracefully, so it's fine to leave for a follow-up.

**LGTM — ready to merge.** (Can't formally Approve since I'm the PR author.)

PR #1274, opened 2026-07-06
Honor fold-case flag in include-ci (R7RS 4.1.7)

### LINKED ISSUE #1141: [R7RS 4.1.7] include-ci does not fold case — ci flag is discarded
**Spec reference:** R7RS 4.1.7, p. 14 — "The difference between the two is that include-ci reads each file as if it began with the #!fold-case directive, while include does not."

**Minimal reproduction:**
```scheme
;; fixtures/upper.scm contains:  (DEFINE Folded-Value 42)
(include-ci "fixtures/upper.scm")
(write folded-value)
;; expected: 42
;; actual:   error: undefined variable 'Folded-Value' (file read case-sensitively)
```

The reader itself supports fold-case correctly — an inline `#!fold-case` directive works — but `include-ci` never enables it. This is an acknowledged TODO in the source:

```zig
// src/vm_library.zig:558 (handleTopLevelInclude)
// include-ci should fold case while reading; the reader currently only
// honors inline `#!fold-case` directives, so the two behave identically.
_ = ci;
```

The library-declaration path (`include-ci` inside `define-library`, `src/vm_library.zig:1016`) shares the same gap.

**Expected:** included file read with case folding; `folded-value` = 42.
**Actual:** case preserved; `Folded-Value` defined-form fails, `folded-value` unbound.

**Verified:** fresh ReleaseSafe build at b2317e8 (v0.12.0).
**Source location:** `src/vm_library.zig` `handleTopLevelInclude` (the `ci` flag is discarded); reader fold support exists at `src/reader_tokens.zig` (`fold_case`).

baijum:
## Summary

Fixes #1141.

- Set `reader.fold_case = ci` in `handleTopLevelInclude` (top-level path) and `compileLibInclude` (library-declaration path) so `include-ci` reads files as if they began with `#!fold-case`, per R7RS 4.1.7.
- Enabled the disabled compliance test in `r7rs-expressions-gaps.scm`.
- Added regression test covering both top-level and `define-library` `include-ci`.

## Test plan

- [x] `bash tests/scheme/run-all.sh` — 1783 pass, 0 fail
- [x] New `include-ci-1141.scm` exercises top-level and library-context `include-ci` with mixed-case fixture files
- [x] Enabled compliance test (`include-ci folds case`) passes

### DIFF
--- a/src/vm_library.zig
+++ b/src/vm_library.zig
@@ -556,10 +556,6 @@ fn openIncludeFile(vm: *VM, file_path: []const u8) VMError!IncludeFile {
 /// with the next form, matching the per-form error isolation of the top-level
 /// file runner in `main.zig`.
 pub fn handleTopLevelInclude(vm: *VM, args: Value, ci: bool) VMError!Value {
-    // include-ci should fold case while reading; the reader currently only
-    // honors inline `#!fold-case` directives, so the two behave identically.
-    _ = ci;
-
     const reader_mod = @import("reader.zig");
 
     // Root the argument spine: evaluating included forms allocates and may
@@ -596,6 +592,7 @@ pub fn handleTopLevelInclude(vm: *VM, args: Value, ci: bool) VMError!Value {
         defer vm.current_lib_dir = saved_lib_dir;
 
         var file_reader = reader_mod.Reader.initWithName(vm.gc, inc.source, owned_path);
+        file_reader.fold_case = ci;
         defer file_reader.deinit();
 
         while (file_reader.hasMore() catch false) {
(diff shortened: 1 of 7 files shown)


baijum (commented):
## Review — looks good ✅

The fix is small, correct, and complete. It wires the previously-discarded `ci` flag into `reader.fold_case` on **both** include dispatch paths, which (per grep) are the only two:

- Top-level `(include-ci …)` → `handleTopLevelInclude` (via `vm_eval.zig`)
- Library-declaration `(include-ci …)` → `compileLibInclude` (via `processLibDeclaration`)

`fold_case` is set inside each per-file loop after the `Reader` is constructed, so a multi-file `(include-ci "a" "b")` correctly re-folds each file as if it began with `#!fold-case`, matching R7RS 4.1.7.

### What I verified
- **The regression tests are meaningful** — they fail without the fix. The library test defines `(DEFINE Lib-Folded-Value 99)` in the included body but `(export lib-folded-value)` in the case-sensitive `.sld`; the names only match once folding is applied, so a discarded `ci` flag breaks export/import. Same logic for the top-level `Top-Level-Folded` case.
- **Fixture placement is correct** — the `lib1141/` subdirectory keeps the helper `.scm` files out of the non-recursive `smoke/*.scm` glob in `run-all.sh`, satisfying the `tests/scheme/CLAUDE.md` fixture-placement rule.
- **SRFI-64 epilogue** grabs the runner before `test-end`, as required.
- The re-enabled compliance test's fixture (`fixtures/include-ci-upper.scm`) already exists.

### Optional (non-blocking) suggestion
Consider adding a test that an explicit `#!no-fold-case` inside an `include-ci`'d file still turns folding back off. That pins down the "as if it began with `#!fold-case`" semantics against future reader changes.

PR #1275, opened 2026-07-06
Signal error on define/set! in immutable environments (#1147)

### LINKED ISSUE #1147: [R7RS 6.12] eval: define/set! into an immutable (environment ...) silently succeeds instead of signaling an error

baijum:
## Summary

- Add `immutable` flag to `SchemeEnvironment` and check it in the VM's `define_global`/`set_global` opcode handlers
- `(environment ...)`, `(null-environment)`, and `(scheme-report-environment)` now create immutable environments per R7RS 6.12
- `(interaction-environment)` remains mutable as required by the spec
- Errors are catchable by `guard`/`with-exception-handler`

Closes #1147

## Test plan

- [x] Zig unit tests: define/set! into immutable env signal error, interaction-environment allows define
- [x] Enabled 2 previously disabled Scheme compliance tests in `r7rs-control-io-gaps.scm`
- [x] Full R7RS suite (1391/1391 pass)
- [x] Full Scheme test suite (113/113 pass)

### DIFF
--- a/src/vm_dispatch.zig
+++ b/src/vm_dispatch.zig
@@ -249,17 +249,12 @@ pub fn runUntil(self: *VM, target_frame_count: usize, target_wind_count: usize) 
                 const src = readU16(self, frame);
                 const closure = frame.closure orelse return VMError.InvalidBytecode;
                 const func = closure.func;
-                if (types.isEnvironment(func.env_val) and types.toEnvironment(func.env_val).immutable) {
-                    const sym = try constantAt(self, func, sym_idx);
-                    const name = if (types.isSymbol(sym)) types.symbolName(sym) else "?";
-                    self.setErrorDetail("set!: cannot modify binding '{s}' in immutable environment", .{name});
-                    return VMError.InvalidArgument;
-                }
                 const src_idx = try registerIndex(self, frame.base, src);
                 const env: *std.StringHashMap(Value) = func.env orelse self.globals;
                 const sym = try constantAt(self, func, sym_idx);
                 if (!types.isSymbol(sym)) return VMError.InvalidBytecode;
                 const name = types.symbolName(sym);
+                if (rejectImmutableEnv(self, func, name, "set!: cannot modify binding")) |e| return e;
                 const val = self.registers[src_idx];
                 // Mirror get_global's hygienic-prefix fallback: a template
                 // set! to a definition-site global compiles as set_global on
@@ -308,17 +303,12 @@ pub fn runUntil(self: *VM, target_frame_count: usize, target_wind_count: usize) 
                 const src = readU16(self, frame);
                 const closure = frame.closure orelse return VMError.InvalidBytecode;
                 const func = closure.func;
-                if (types.isEnvironment(func.env_val) and types.toEnvironment(func.env_val).immutable) {
-                    const sym = try constantAt(self, func, sym_idx);
-                    const name = if (types.isSymbol(sym)) types.symbolName(sym) else "?";
-                    self.setErrorDetail("define: cannot define '{s}' in immutable environment", .{name});
-                    return VMError.InvalidArgument;
-                }
                 const src_idx = try registerIndex(self, frame.base, src);
                 const env: *std.StringHashMap(Value) = func.env orelse self.globals;
                 const sym = try constantAt(self, func, sym_idx);
                 if (!types.isSymbol(sym)) return VMError.InvalidBytecode;
                 const name = types.symbolName(sym);
+                if (rejectImmutableEnv(self, func, name, "define: cannot define")) |e| return e;
                 const val = self.registers[src_idx];
                 if (env == self.globals) {
                     // Structural mutation of the shared globals map: may
@@ -1266,6 +1256,14 @@ pub fn readU16(vm: *VM, frame: *CallFrame) u16 {
 
 pub fn readI16(vm: *VM, frame: *CallFrame) i16 {
     return @bitCast(readU16(vm, frame));
+}
+
+noinline fn rejectImmutableEnv(self: *VM, func: *types.Function, name: []const u8, comptime verb: []const u8) ?VMError {
+    if (types.isEnvironment(func.env_val) and types.toEnvironment(func.env_val).immutable) {
+        self.setErrorDetail(verb ++ " '{s}' in immutable environment", .{name});
+        return VMError.InvalidArgument;
+    }
+    return null;
 }
 
 noinline fn raiseUndefinedVariable(self: *VM, name: []const u8) VMError {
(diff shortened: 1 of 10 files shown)


src/vm_dispatch.zig
```diff
@@ -302,6 +308,12 @@ pub fn runUntil(self: *VM, target_frame_count: usize, target_wind_count: usize)
                 const src = readU16(self, frame);
                 const closure = frame.closure orelse return VMError.InvalidBytecode;
                 const func = closure.func;
+                if (types.isEnvironment(func.env_val) and types.toEnvironment(func.env_val).immutable) {
```
baijum: This covers value bindings (`define_global`/`set_global`) but not **syntactic** bindings, so the R7RS 6.12 immutability guarantee has a hole. Verified on b72b7fd:

```scheme
(eval '(define-syntax leaked (syntax-rules () ((_) 999)))
      (environment '(scheme base)))
(leaked)   ; => 999   ; no error, and it's now defined globally
```

Two problems: (a) `define-syntax`/`let-syntax` into an immutable env silently succeeds instead of erroring; (b) worse, it isn't even contained — `compileExpressionInEnv` copies newly-defined macros back into the shared `vm.macros`, so the keyword leaks into the *global* namespace.

Suggest either erroring on macro definitions into an immutable env as well, or (if that's a bigger change) documenting the limitation here and filing a follow-up issue. Either way a regression test would be good.

src/vm_dispatch.zig
```diff
@@ -249,6 +249,12 @@ pub fn runUntil(self: *VM, target_frame_count: usize, target_wind_count: usize)
...
```
baijum: Minor: this guard is duplicated verbatim in `define_global` below (only the error verb differs). Consider factoring it into a small helper, e.g.

```zig
fn immutableEnvError(self: *VM, func: *types.Function, sym_idx: u16, comptime verb: []const u8) ?VMError { ... }
```

so the two handlers can't drift. Optional.

baijum (commented):
## Review — immutable environments (#1147)

The core change is correct and well-tested. I built the branch (b72b7fd), ran `zig build test` (pass), and confirmed the intended behavior empirically:

- `(eval '(define foo 1) (environment '(scheme base)))` → catchable error ✅
- `(eval '(set! car 42) (environment '(scheme base)))` → catchable error, and `car` is not clobbered globally ✅
- `interaction-environment` still allows `define` ✅
- A closure that escapes the `eval` — `(eval '(lambda () (set! car 42)) (environment ...))` — still errors when called later, because `env_val` propagates to nested `Function`s ✅
- `define-values` into an immutable env also errors (compiles to `define_global`) ✅

Two things worth addressing before merge (details inline):

1. **The immutability guarantee is incomplete for syntactic keywords** — `define-syntax` bypasses the guard *and* leaks the macro into the global namespace (see inline comment on `define_global`). This is the one I'd want resolved or explicitly documented + tracked.
2. **`define-record-type` into an immutable env errors, but with a misleading message.** `(eval '(define-record-type pt (mk-pt x) pt? (x pt-x)) (environment '(scheme base)))` reports `undefined variable 'pt'` (the first desugared `define` errors, leaving later references dangling) rather than a clear "immutable environment" error. No leak — just poor UX. Optional to fix, but a targeted test would lock in the behavior.

Minor: the two guard blocks are duplicated verbatim (inline note on `set_global`).

Nice touch updating `env-uaf.scm` to `let` — the GC-liveness path (`func.env` pinning the map) is unchanged, so the regression is still exercised.

baijum (commented):
## Re-review of 5d030cc — both findings addressed ✅

Built and verified locally on `5d030cc`.

**1. Incomplete immutability / macro leak — fixed.** Two-layer fix is the right approach:
- `compileDefineSyntax` now rejects `define-syntax` when `lib_env_val` is an immutable environment.
- `compileExpressionInEnv` no longer copies compiled macros back into the shared `vm.macros` for immutable envs, closing the global-namespace leak.

Verified:
- `(eval '(define-syntax m ...) (environment '(scheme base)))` → catchable error, and `(m)` is **no longer** defined at the top level afterward (leak gone).
- Importantly, the guard is *not* over-broad — **internal** `define-syntax` and `let-syntax` inside an immutable-env eval still work (they're local bindings, not env mutations):
  - `((eval '(lambda () (define-syntax loc (syntax-rules () ((_) 7))) (loc)) (environment '(scheme base))))` → `7`
  - `(eval '(let-syntax ((ls (syntax-rules () ((_) 5)))) (ls)) (environment '(scheme base)))` → `5`

**2. Duplicated guard — fixed.** Extracted into `rejectImmutableEnv`; both `set_global`/`define_global` now share it.

**Tests (local):** `zig build test` ✅, full R7RS suite (1391, 0 fail) ✅, and the three touched Scheme files all pass (`r7rs-control-io-gaps` 32/32, `env-uaf`, `load-env-1190` 4/4).

**Remaining (non-blocking):** the earlier minor note stands — `(eval '(define-record-type ...) (environment ...))` still surfaces `undefined variable '<type>'` instead of a clean "immutable environment" message. It errors and doesn't leak, so it's a UX nit, fine to defer to a follow-up.

**Verdict: ready to merge once CI is green.** Local verification passes across the board; CI checks (format, wasm already green; test matrix in progress) just need to finish.

Issue #1010, opened 2026-07-03
GC during library include-load corrupts compilation: (import (srfi 158)) after (srfi 115) fails with CompileError masked as "library not found"


baijum:
## Symptom

A three-line program fails on `main` (fadb074):

```scheme
(import (scheme base) (scheme write))
(import (srfi 115))
(import (srfi 158))
```

```
$ kaappi min.scm
min.scm:3: error: library not found: (srfi.158)
```

`lib/srfi/158.sld` exists and imports fine in a fresh session or in other orders (`(import (srfi 115))` + `(import (srfi 158))` *without* the leading `(scheme base)/(scheme write)` import works). The failure is deterministic for this exact sequence but is really a GC-timing window: any prefix that shifts heap state enough triggers it (the kaappi-book ch16 replay hits it reliably; `(srfi 113)`/`(srfi 128)` after heap activity did too before other orderings drifted).

Note the "library not found" is a **masking message**: the library file is found and loaded; a form inside it fails to compile, `loadLibrarySource` maps everything to `error.UndefinedVariable`, and `processImportSet` then reports "library not found". In stdin/REPL mode the same failure surfaces as `runtime error: error.CompileError`.

## Diagnosis so far

Instrumentation shows the real failure:

- `compileLibExpr` → `compiler_mod.compileExpressionInEnv` returns **`error.InvalidSyntax`** while compiling `gdelete-neighbor-dups` — the large `case-lambda` in `lib/srfi/158-impl.scm` (loaded via `include`).
- It is definitively GC: the identical load **passes with `-Dgc-threshold=100000000`** (GC effectively off) and **passes with `-Dgc-threshold=1`** (collect on every allocation), but **fails at the default threshold (8192)**. Only a collection landing mid-compile at a particular point corrupts it — a classic generational-GC window, not an every-allocation rooting hole.
- The same `case-lambda` form compiles fine as a plain top-level define in a script after the same imports — the failure needs the library-include compile path (`compileExpressionInEnv` with `lib_env` + `restricted_env=true`).
- #1008 exposed this by enlarging `158-impl.scm`; the bug predates it (same class as the v0.11.0-era mid-session import failures, most of which were fixed by the rooting work between v0.11.0 and fadb074).

## Deterministic standalone reproducer

Clone srfi/158 as a private library and import it after (srfi 115):

```sh
mkdir -p /tmp/mylib
cp lib/srfi/158-impl.scm /tmp/mylib/foo-impl.scm
[...]

PR #1276, opened 2026-07-06
Desugar define-record-type in body contexts (R7RS §5.5)

### LINKED ISSUE #560: define-record-type only desugars at direct top-level, not inside begin or <body> (vm_records.zig, R7RS 5.5/5.3 violation)
**Bug:** handleDefineRecordType is only invoked for direct top-level forms. Inside `(begin (define-record-type ...))` or library bodies / let bodies it is not recognized as definition.

**Reproduction:**
```scheme
(begin
  (define-record-type pt (make-pt x) pt? (x get-x))
  (get-x (make-pt 1)))
```
Actual:
  error: undefined variable 'pt'.

Direct top-level works; wrapped fails.

Run: `./zig-out/bin/kaappi repro-begin.scm`

**Expected:** R7RS allows definitions inside `begin` at top level (splicing) and in bodies.

Confirmed by run.

baijum:
Fixes #560

## Summary

- `define-record-type` now works inside `let`, `lambda`, `letrec`, and other body contexts, as required by R7RS §5.5
- Desugars into equivalent `define` forms using existing `%make-record-type`, `%make-record`, `%record?`, `%record-ref`, `%record-set!` primitives
- The desugared defines enter the existing letrec* body-scanning machinery, so scoping is correct in all contexts
- Updated both body scanners (legacy in `compiler_lambda.zig` and IR path in `compiler_ir.zig`) plus globals prescans

## Test plan

- [x] Enabled previously disabled SRFI-9 body test (37/37 pass)
- [x] New `560-record-body.scm` regression test (11 cases: let, lambda, nested, mutators, predicates, generative pattern, field reordering, define-syntax interaction, letrec)
- [x] GC stress test (`-Dgc-stress=true`) passes
- [x] Full R7RS suite (1395 pass, 0 fail)
- [x] Full `run-all.sh` (1783 pass, 0 fail)

### DIFF
--- a/src/compiler_ir.zig
+++ b/src/compiler_ir.zig
@@ -8,6 +8,7 @@ const compiler_lambda = @import("compiler_lambda.zig");
 const macro = @import("compiler_macro.zig");
 const globals_mod = @import("globals.zig");
 const ir_mod = @import("ir.zig");
+const vm_records = @import("vm_records.zig");
 const Value = types.Value;
 const OpCode = types.OpCode;
 const Compiler = compiler_mod.Compiler;
@@ -320,6 +321,29 @@ pub fn compileLambdaWithIR(self: *Compiler, args: Value, dst: u16, name: ?[]cons
                                 }
                             }
                         }
+                    } else if (std.mem.eql(u8, form_name, "define-record-type")) {
+                        if (vm_records.parseRecordSpec(types.cdr(form))) |spec| {
+                            if (!globals.contains(spec.ctor_name)) {
+                                globals.put(spec.ctor_name, types.VOID) catch {};
+                                prescan_names.append(child.gc.allocator, spec.ctor_name) catch {};
+                            }
+                            if (!globals.contains(spec.pred_name)) {
+                                globals.put(spec.pred_name, types.VOID) catch {};
+                                prescan_names.append(child.gc.allocator, spec.pred_name) catch {};
+                            }
+                            for (0..spec.field_count) |fi| {
+                                if (!globals.contains(spec.accessor_names[fi])) {
+                                    globals.put(spec.accessor_names[fi], types.VOID) catch {};
+                                    prescan_names.append(child.gc.allocator, spec.accessor_names[fi]) catch {};
+                                }
+                                if (spec.mutator_names[fi]) |mn| {
+                                    if (!globals.contains(mn)) {
+                                        globals.put(mn, types.VOID) catch {};
+                                        prescan_names.append(child.gc.allocator, mn) catch {};
+                                    }
+                                }
+                            }
+                        }
                     }
                 }
             }
@@ -376,6 +400,8 @@ pub fn compileLambdaWithIR(self: *Compiler, args: Value, dst: u16, name: ?[]cons
                         all_def_count += 1;
                     }
                 } else break;
+            } else if (std.mem.eql(u8, hn, "define-record-type")) {
+                vm_records.collectRecordTypeDefNames(child.gc, types.cdr(expr), all_def_names[0..], &all_def_count) catch break;
             } else if (!std.mem.eql(u8, hn, "define-syntax")) {
                 break;
             }
@@ -410,6 +436,18 @@ pub fn compileLambdaWithIR(self: *Compiler, args: Value, dst: u16, name: ?[]cons
                 tx_obj.def_env_val = child.lib_env_val;
             }
             child.macros.put(ds_name, tx_val) catch return CompileError.OutOfMemory;
+            current = types.cdr(current);
+            continue;
+        }
+        if (std.mem.eql(u8, head_name, "define-record-type")) {
+            vm_records.expandRecordTypeDefines(
+                child.gc,
+                types.cdr(expr),
+                def_names[0..],
+                def_inits[0..],
+                &def_count,
+                &child.gc.extra_roots,
+            ) catch break;
             current = types.cdr(current);
             continue;
         }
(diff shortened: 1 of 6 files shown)


src/compiler_ir.zig
```diff
@@ -413,6 +439,18 @@ pub fn compileLambdaWithIR(self: *Compiler, args: Value, dst: u16, name: ?[]cons
             current = types.cdr(current);
             continue;
         }
+        if (std.mem.eql(u8, head_name, "define-record-type")) {
```
baijum, line 445: This (and the matching block in `compiler_lambda.zig`) is the only place `define-record-type` gets desugared, so the feature is limited to forms that reach here as a *leading* define. Bodies that arrive as a spliced `(begin ...)` skip it — most notably `let-values`/`let*-values`, which desugar their body into `(begin <body>)` and are explicit R7RS §5.3.2 body contexts:

```scheme
(let-values (((a b) (values 1 2)))
  (define-record-type r (mk v) r? (v r-v))
  (r-v (mk (+ a b))))   ; error: undefined variable 'r'
```

Plain `define` works in the same spot because `compileDefine` has a general `in_body_scope` path. Consider giving `define-record-type` an equivalent general dispatch (or handling it during `begin`-splicing), or documenting the `let-values`/`let*-values` gap explicitly.

src/compiler_ir.zig
```diff
@@ -413,6 +439,18 @@ pub fn compileLambdaWithIR(self: *Compiler, args: Value, dst: u16, name: ?[]cons
...
+            vm_records.expandRecordTypeDefines(
+                child.gc,
+                types.cdr(expr),
+                def_names[0..],
+                def_inits[0..],
+                &def_count,
+                &child.gc.extra_roots,
+            ) catch break;
```
baijum: `catch break` here swallows *every* error `expandRecordTypeDefines` can return, including `OutOfMemory` and `TooManyLocals`. Compare the adjacent plain-`define` handling a few lines down, which propagates these (`return CompileError.OutOfMemory` / `return CompileError.TooManyLocals`). On a genuine OOM or locals-overflow the record form is silently left un-desugared and then compiled as an ordinary expression, surfacing as a misleading `undefined variable '<type-name>'` instead of the real error. Suggest propagating `OutOfMemory`/`TooManyLocals` and only `break`ing on the malformed-syntax case (`InvalidSyntax`). Same pattern applies to the `collectRecordTypeDefNames ... catch break` on line 404 and both call sites in `compiler_lambda.zig`.

src/vm_records.zig
```diff
@@ -278,3 +281,270 @@ pub fn handleDefineRecordType(vm: *VM, args: Value) VMError!Value {
...
+    field_count: usize,
+    accessor_names: [256][]const u8,
+    mutator_names: [256]?[]const u8,
+    ctor_field_indices: [256]usize,
+};
+
+/// Parse (define-record-type name (ctor field...) pred (field acc [mut])...)
+/// args is the cdr of the full form (everything after 'define-record-type').
+/// Returns null on malformed input.
+pub fn parseRecordSpec(args: Value) ?RecordSpec {
```
baijum, line 216: +1 to CodeRabbit: this reproduces the record-grammar parsing already in `handleDefineRecordType` (name/ctor/pred/field-spec walk, 256 caps, duplicate-field check, ctor-field index mapping). Worth folding `handleDefineRecordType` onto `parseRecordSpec` so the grammar has a single source of truth. Note the two paths *are* intentionally different downstream — the top-level path builds the type eagerly via `allocRecordType` and stores it in a global, while the body path emits `(define __rt (%make-record-type ...))` for correct generative-per-activation scoping — so only the parsing should be shared, not the emission.

baijum (commented):
Reviewed the desugaring approach, built the branch locally, ran the new test + srfi9 + full R7RS suite (all green), and probed the body contexts the tests don't cover. The core mechanism is sound: routing the desugared `define`s through the existing letrec* body-scanning machinery is the right call, generative-per-activation semantics fall out correctly, and the GC handling (`no_collect` around construction + mirroring inits into `extra_roots`) follows the #1010-aware pattern. CI is fully green and I couldn't find any regression.

One **functional gap** worth resolving before merge, plus a minor robustness nit — details inline.

### Gap: `let-values` / `let*-values` bodies (and `begin`-splicing) aren't covered

The fix only fires from the two dedicated leading-`define` scan loops. It does **not** participate in `begin`-splicing, so any body that reaches the compiler as a spliced `begin` misses it. Concretely, `let-values` desugars its body into `(begin <body>...)` inside a lambda, so:

```scheme
(let-values (((a b) (values 1 2)))
  (define-record-type r (mk v) r? (v r-v))   ; => error: undefined variable 'r'
  (r-v (mk (+ a b))))
```

fails, even though the same body with a plain `(define ...)` works. `let-values` and `let*-values` are explicitly listed as `<body>` contexts in R7RS §5.3.2, so this contradicts the PR description ("works inside ... and other body contexts, as required by R7RS §5.5"). The same gap makes a bare `(begin (define-record-type ...) ...)` in a lambda body fail. `when`/`unless`/`do`/`cond`/`case` bodies fail too — those aren't §5.3.2 body contexts so that's defensible, but Kaappi *does* accept plain `define` there, so the asymmetry is surprising.

Root cause: plain `define` has a general `in_body_scope` path in `compileDefine` that turns it into a local wherever it appears; `define-record-type` has no equivalent general dispatch — it's only recognized by the leading-`define` scanners. Either handle it in the `begin`-splice / general in-body path, or scope the PR down and document `let-values`/`let*-values` as a known limitation in the test file and PR notes.

(Also +1 to CodeRabbit's note about `parseRecordSpec` duplicating the grammar in `handleDefineRecordType` — commented inline.)

PR #1277, opened 2026-07-06
Separate let-syntax from letrec-syntax scoping (#1140)

### LINKED ISSUE #1140: [R7RS 4.3.1] let-syntax: transformers see sibling keywords from the same binding form (letrec-syntax semantics)
**Spec reference:** R7RS 4.3.1, p. 22 — let-syntax: "The <body> is expanded in the syntactic environment obtained by extending the syntactic environment of the let-syntax expression with macros whose keywords are the <keyword>s, bound to the specified transformers." Only the *body* is in the new keywords' region. Contrast letrec-syntax: "Each binding of a <keyword> has the <transformer spec>s as well as the <body> within its region."
**Severity:** wrong-result

**Minimal reproduction:**
```scheme
(define-syntax m (syntax-rules () ((_) 'outer)))

(let-syntax ((m      (syntax-rules () ((_) 'inner)))
             (call-m (syntax-rules () ((_) (m)))))
  (call-m))
;; expected: outer   actual: inner
```

`call-m`'s template reference to `m` is specified in the let-syntax *bindings*, which are outside the region of the new `m` — it must refer to the outer `m`. Kaappi resolves it to the sibling binding, i.e. it gives `let-syntax` the semantics of `letrec-syntax`. (The equivalent `letrec-syntax` program correctly yields `inner`; that behavior is right.)

**Expected:** `outer`
**Actual:** `inner`

**Verified:** fresh ReleaseSafe build at b2317e8 (v0.12.0); deterministic. Binding order does not matter — placing the `m` rebinding after `call-m` gives the same wrong result.
**Source location:** `src/compiler_macro.zig` `compileLetSyntax` (dispatched from `src/compiler.zig:660`) — appears to register all bindings before expanding transformer references, so siblings see each other.

Related (closed) #729 fixed definition-site capture for template free *variables*; this issue is the analogous scoping rule for macro *keywords* bound by the same let-syntax.

baijum:
## Summary

- `let-syntax` was incorrectly giving `letrec-syntax` semantics — sibling macro bindings could see each other's keywords, violating R7RS 4.3.1
- Refactored `compileLetSyntax` to parse all transformer specs before registering any bindings, storing each keyword's outer macro value on the Transformer struct
- Added macro swap in `expandAndCompileMacroUse` so template free references resolve to the definition-site environment during result compilation
- Separated `compileLetrecSyntax` into its own function preserving the register-while-parsing behavior

## Test plan

- [x] New Zig unit tests for both `let-syntax` (outer scope) and `letrec-syntax` (inner scope) sibling keyword resolution
- [x] Enabled previously disabled SRFI-64 test in `r7rs-expressions-gaps.scm`
- [x] All 1782 Scheme tests pass (387 files + 1395 R7RS), 0 failures
- [x] All Zig unit tests pass
- [x] GC stress mode (`-Dgc-stress=true`) passes cleanly

Closes #1140

### DIFF
--- a/src/compiler_macro.zig
+++ b/src/compiler_macro.zig
@@ -202,10 +202,14 @@ pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8, 
     // Temporarily swap sibling keywords to their outer values.
     const peer_names = tx.let_syntax_peer_names;
     const peer_outer = tx.let_syntax_peer_vals;
-    var saved_peer: [32]?Value = undefined;
-    if (peer_names.len > 0 and peer_names.len <= 32) {
+    const saved_peer = if (peer_names.len > 0)
+        self.gc.allocator.alloc(?Value, peer_names.len) catch null
+    else
+        null;
+    defer if (saved_peer) |sp| self.gc.allocator.free(sp);
+    if (saved_peer) |sp| {
         for (peer_names, peer_outer, 0..) |pn, pv, i| {
-            saved_peer[i] = self.macros.get(pn);
+            sp[i] = self.macros.get(pn);
             if (pv != types.NIL) {
                 self.macros.put(pn, pv) catch {};
             } else {
@@ -214,9 +218,9 @@ pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8, 
         }
     }
     const result_err = self.compileExpr(expanded_root, dst, is_tail);
-    if (peer_names.len > 0 and peer_names.len <= 32) {
+    if (saved_peer) |sp| {
         for (peer_names, 0..) |pn, i| {
-            if (saved_peer[i]) |old| {
+            if (sp[i]) |old| {
                 self.macros.put(pn, old) catch {};
             } else {
                 _ = self.macros.remove(pn);
@@ -276,9 +280,10 @@ pub fn compileLetSyntax(self: *Compiler, args: Value, dst: u16, is_tail: bool) C
 
     // Phase 1: Parse ALL transformer specs before registering any.
     // self.macros still has the outer values during this phase.
-    var kw_names: [32][]const u8 = undefined;
-    var tx_vals: [32]Value = undefined;
-    var bind_n: usize = 0;
+    var kw_names: std.ArrayList([]const u8) = .empty;
+    defer kw_names.deinit(self.gc.allocator);
+    var tx_vals: std.ArrayList(Value) = .empty;
+    defer tx_vals.deinit(self.gc.allocator);
     var roots_pushed: usize = 0;
 
     var binding_list = bindings;
@@ -291,22 +296,23 @@ pub fn compileLetSyntax(self: *Compiler, args: Value, dst: u16, is_tail: bool) C
         const binding_rest = types.cdr(binding);
         if (!types.isPair(binding_rest)) return CompileError.InvalidSyntax;
         const transformer_spec = types.car(binding_rest);
-        if (bind_n >= 32) return CompileError.InvalidSyntax;
-        tx_vals[bind_n] = parseSyntaxRules(self, transformer_spec, &.{}) catch return CompileError.InvalidSyntax;
-        self.gc.pushRoot(&tx_vals[bind_n]);
+        tx_vals.append(self.gc.allocator, parseSyntaxRules(self, transformer_spec, &.{}) catch
+            return CompileError.InvalidSyntax) catch return CompileError.OutOfMemory;
+        self.gc.pushRoot(&tx_vals.items[tx_vals.items.len - 1]);
         roots_pushed += 1;
-        kw_names[bind_n] = types.symbolName(keyword);
-        bind_n += 1;
+        kw_names.append(self.gc.allocator, types.symbolName(keyword)) catch return CompileError.OutOfMemory;
         binding_list = types.cdr(binding_list);
     }
     defer for (0..roots_pushed) |_| self.gc.popRoot();
 
     // Build peer snapshot: each keyword's outer macro value (NIL = unbound).
-    // Computed once before registration, duped per transformer so each
-    // owns its own copy (avoids double-free when GC cleans up).
-    var peer_snap_names: [32][]const u8 = undefined;
-    var peer_snap_vals: [32]Value = undefined;
-    for (kw_names[0..bind_n], 0..) |name, i| {
+    // Duped per transformer so each owns its own copy.
+    const bind_n = kw_names.items.len;
+    const peer_snap_names = self.gc.allocator.alloc([]const u8, bind_n) catch return CompileError.OutOfMemory;
+    defer self.gc.allocator.free(peer_snap_names);
+    const peer_snap_vals = self.gc.allocator.alloc(Value, bind_n) catch return CompileError.OutOfMemory;
+    defer self.gc.allocator.free(peer_snap_vals);
+    for (kw_names.items, 0..) |name, i| {
         peer_snap_names[i] = name;
         peer_snap_vals[i] = self.macros.get(name) orelse types.NIL;
     }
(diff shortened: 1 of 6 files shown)


src/compiler_macro.zig
```diff
@@ -249,6 +274,98 @@ pub fn compileLetSyntax(self: *Compiler, args: Value, dst: u16, is_tail: bool) C
...
+    while (binding_list != types.NIL) {
+        if (!types.isPair(binding_list)) return CompileError.InvalidSyntax;
+        const binding = types.car(binding_list);
+        if (!types.isPair(binding)) return CompileError.InvalidSyntax;
+        const keyword = types.car(binding);
+        if (!types.isSymbol(keyword)) return CompileError.InvalidSyntax;
+        const binding_rest = types.cdr(binding);
+        if (!types.isPair(binding_rest)) return CompileError.InvalidSyntax;
+        const transformer_spec = types.car(binding_rest);
+        if (bind_n >= 32) return CompileError.InvalidSyntax;
```
baijum: **Regression: `let-syntax` now rejects >32 bindings.** Phase 1 uses fixed 32-slot stack arrays (`kw_names`/`tx_vals`/`peer_snap_*`), so this guard turns a large-but-legal `let-syntax` into a generic `error.InvalidSyntax`.

Verified against `main` on a *fresh, uncached* file (a `.sbc` cache from a same-version binary hides this):
- `(let-syntax (<40 bindings>) (k0))` → `error.InvalidSyntax` here; compiles & prints on `main`.
- `compileLetrecSyntax` still uses `ArrayList`s, so `letrec-syntax` with 40 bindings works — the two forms now disagree on the max binding count, which is surprising.

R7RS puts no bound on the number of bindings. Suggest using `ArrayList`s here too (as letrec does), or at minimum emit a clear "too many syntax bindings" diagnostic instead of a bare `InvalidSyntax`, and add a regression test for the >32 case.

src/compiler_macro.zig
```diff
@@ -197,7 +197,32 @@ pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8,
             .is_global_alias = true,
         });
     }
+    // R7RS 4.3.1: let-syntax transformers resolve free macro references
+    // from the definition-site environment, not the use-site environment.
+    // Temporarily swap sibling keywords to their outer values.
+    const peer_names = tx.let_syntax_peer_names;
+    const peer_outer = tx.let_syntax_peer_vals;
+    var saved_peer: [32]?Value = undefined;
+    if (peer_names.len > 0 and peer_names.len <= 32) {
```
baijum: The literal `32` here is coupled to the `bind_n >= 32` cap in `compileLetSyntax` and to `saved_peer: [32]` above. If that cap is ever raised without updating this guard, `peer_names.len <= 32` silently becomes false, the peer swap is skipped, and `let-syntax` silently reverts to `letrec-syntax` semantics — exactly the bug this PR fixes, but with no error. Worth hoisting into one shared named constant, or replacing the guard with `std.debug.assert(peer_names.len <= saved_peer.len)`.

src/compiler_macro.zig
```diff
@@ -249,6 +274,98 @@ pub fn compileLetSyntax(self: *Compiler, args: Value, dst: u16, is_tail: bool) C
...
+    for (saved_names.items, saved_values.items) |name, saved_val| {
+        if (saved_val) |old_val| {
+            try self.macros.put(name, old_val);
+        } else {
+            _ = self.macros.remove(name);
+        }
+    }
+}
+
+pub fn compileLetrecSyntax(self: *Compiler, args: Value, dst: u16, is_tail: bool) CompileError!void {
```
baijum, line 355: `compileLetSyntax` and `compileLetrecSyntax` now duplicate the binding-parse loop, captured-locals capture, the save/restore `ArrayList`s, and the entire `beginScope` / `beginBodyMacroScope` / body-compile / `endScope` / restore block (~40+ lines). The only substantive differences are *when* the bindings become visible (before vs. during parsing) and the peer snapshot. A future fix to one path can easily miss the other. Consider extracting the shared body-compile + restore tail into a helper.

src/gc_deep_copy.zig
```diff
@@ -254,6 +254,12 @@ fn deepCopyValue(gc: *GC, src: Value, visited: *std.AutoHashMap(usize, Value)) !
             }
             new_tx.def_env = t.def_env;
             new_tx.def_env_val = t.def_env_val;
+            if (t.let_syntax_peer_vals.len > 0) {
+                const peer_vals = try gc.allocator.alloc(Value, t.let_syntax_peer_vals.len);
+                for (t.let_syntax_peer_vals, 0..) |v, i| peer_vals[i] = try deepCopyValue(gc, v, visited);
+                new_tx.let_syntax_peer_vals = peer_vals;
+                new_tx.let_syntax_peer_names = gc.allocator.dupe([]const u8, t.let_syntax_peer_names) catch &.{};
```
baijum: Asymmetric error handling: `let_syntax_peer_vals` is allocated with `try` (propagates OOM) but `let_syntax_peer_names` uses `catch &.{}`. On a names-dupe OOM the copied transformer ends up with mismatched lengths (`vals.len = N`, `names.len = 0`). It won't crash today because the swap in `expandAndCompileMacroUse` guards on `peer_names.len` (and would just skip), but it silently drops the peer swap for that transformer. Make both consistent — ideally both `try` so the OOM propagates.

baijum (commented):
## Review — separate `let-syntax` from `letrec-syntax` scoping (#1140)

The core fix is correct and I verified the behavior end-to-end: `let-syntax` sibling keywords now resolve to the **outer** environment while `letrec-syntax` sees the **inner** bindings — including the nested-`let-syntax` (enclosing binding shows through) and lambda-body / child-compiler cases — and it survives `-Dgc-stress=true`. The GC plumbing for `let_syntax_peer_vals` is threaded through every path (`referencesYoung` / `markObjectContents` / `markValueInner` / `objectSize` / `freeObject` / `deepCopy`), which is nicely thorough.

One **confirmed regression** (>32 bindings) plus a few maintainability notes inline.

### Confirmed regression: `let-syntax` with more than 32 bindings
The rewrite replaced the old unbounded `ArrayList`s with fixed 32-slot stack arrays, so a large-but-legal `let-syntax` is now rejected. Reproduced on a **fresh, uncached** source file (important — a stale `.sbc` cache written by a same-version binary will mask this):

- `main`: 40-binding `let-syntax` → compiles, prints `0`.
- this branch: same source → `compile error: error.InvalidSyntax`.
- `letrec-syntax` with 40 bindings still works on both — the sibling function kept its `ArrayList`s, so the two forms now disagree on the limit.

Details on the `bind_n >= 32` line.

src/compiler_macro.zig
```diff
@@ -249,6 +278,71 @@ pub fn compileLetSyntax(self: *Compiler, args: Value, dst: u16, is_tail: bool) C
...
+        const binding = types.car(binding_list);
+        if (!types.isPair(binding)) return CompileError.InvalidSyntax;
+        const keyword = types.car(binding);
+        if (!types.isSymbol(keyword)) return CompileError.InvalidSyntax;
+        const binding_rest = types.cdr(binding);
+        if (!types.isPair(binding_rest)) return CompileError.InvalidSyntax;
+        const transformer_spec = types.car(binding_rest);
+        tx_vals.append(self.gc.allocator, parseSyntaxRules(self, transformer_spec, &.{}) catch
+            return CompileError.InvalidSyntax) catch return CompileError.OutOfMemory;
+        self.gc.pushRoot(&tx_vals.items[tx_vals.items.len - 1]);
```
baijum, line 316: **New GC-safety issue introduced by this commit (latent use-after-free).** `pushRoot` stores the *pointer* you pass (`root_buffer: [4096]*Value`, dereferenced as `root.*` at mark time — memory.zig:1121). Switching `tx_vals` from the address-stable `[32]Value` to a growable `ArrayList` means `&tx_vals.items[...]` points into the backing buffer, and the very next `append` can reallocate it. Once that happens, every root pushed for an earlier element dangles into freed memory. The parse loop keeps allocating (`parseSyntaxRules` → `allocTransformer` → `maybeCollect`), so a GC can dereference those dangling roots; if the freed buffer has been reused, the not-yet-registered transformers go unmarked and get swept, then Phase 2 registers/uses freed objects.

This is the exact pattern `.claude/rules/gc-safety.md` warns against (holding a pointer across allocations). I could **not** force a crash even at 300 bindings under `-Doptimize=Debug -Dgc-stress=true` (5126 collections) — this DebugAllocator happens to leave the freed small buffers intact, so the dangling reads return stale-but-valid pointer bytes. But it's unsafe by contract and could bite on a different allocator/target/Zig version or allocation interleaving.

Suggested fix: restore address stability. Count the bindings first, then `try tx_vals.ensureTotalCapacityPrecise(self.gc.allocator, count)` before the append loop so the buffer never reallocates — then `&tx_vals.items[i]` stays valid. (Alternatively root by value via `gc.extra_roots`, which `markRoots` marks as `v` not `v.*`.)

baijum (commented):
## Follow-up review of `ec645fd` ("Address review feedback")

Thanks — all four points from the previous round are genuinely addressed, and I verified them:

- **>32 bindings**: fixed. A fresh 40/60/300-binding `let-syntax` now compiles and runs (previously `error.InvalidSyntax`). `let-syntax` and `letrec-syntax` now agree.
- **Coupled magic `32`**: fixed. `saved_peer` is now `alloc(?Value, peer_names.len)` with a `defer free`; the `<= 32` guard is gone.
- **Duplication**: fixed. `captureLocalsOnTransformer` / `compileSyntaxBody` / `restoreMacros` deduplicate the two forms nicely.
- **Deep-copy asymmetry**: fixed. `peer_names` now propagates OOM and there's an `errdefer` for `peer_vals`.

Unit tests, R7RS expression-gaps compliance, and the hygiene/macros suites all pass, including under `-Dgc-stress=true`.

However, the >32 fix introduced **one new GC-safety issue** — details inline on `compiler_macro.zig`. I'd hold merge on that one; the rest looks good.

baijum (commented):
## Re-review of `a8cc18d` — GC-safety issue resolved ✅

The one blocking issue is fixed correctly. `compileLetSyntax` now counts the bindings up front, `ensureTotalCapacity` on `tx_vals`/`kw_names`, and switches to `appendAssumeCapacity` — so the backing buffer is allocated once and never reallocates, which keeps the `pushRoot(&tx_vals.items[i])` pointers valid for the lifetime of the roots. That's exactly the address-stability the old `[32]Value` array had, without the size cap. The `std.debug.assert(peer_names.len == peer_outer.len)` and the two OOM-propagation fixes (`captureLocalsOnTransformer`, `saved_peer`) are good hardening on top.

### Verified
- `zig build` + full `zig build test` — pass.
- Fresh 60- and 300-binding `let-syntax` under `-Dgc-stress=true` (ReleaseSafe) and a 60-binding run under `-Doptimize=Debug -Dgc-stress=true` — correct output, no panic/poison/segfault.
- `(let-syntax () ...)` empty-bindings edge case — works.
- R7RS expression-gaps compliance + all hygiene/macros suites — pass.
- Core semantics still correct: `let-syntax` siblings resolve to the outer env, `letrec-syntax` to the inner.

### Verdict
LGTM — ready to merge from my side. (GitHub won't let me formally *approve* my own PR, so treating this as an approving comment.)

Optional, non-blocking: consider adding a Scheme-level regression test for the >32-binding case (the original v1 regression), since it isn't otherwise covered — e.g. a generated `let-syntax` with ~40 bindings in `tests/scheme/`. Not required for merge.

PR #1278, opened 2026-07-06
Handle consecutive ellipses in syntax-rules templates (#1243)

### LINKED ISSUE #1243: [R7RS 4.3.2] Doubled ellipsis in template does not flatten: ((x ...) ...) matched input expanded with (x ... ...) produces garbage containing literal '...'

baijum:
## Summary

- Teach `instantiateEllipsis` in `src/expander.zig` to detect and strip consecutive `...` tokens from `rest_template`. When extra ellipses are present, each outer iteration builds a synthetic template with the remaining ellipses and recurses, then splices the resulting list into the output.
- Fixes the R7RS 4.3.2 flattening construct `(x ... ...)` which previously produced garbage (`(() () () ...)`) instead of the expected flat list.
- Uncomments the disabled compliance test and adds a Zig unit test with edge cases (single group, empty groups, no groups).

Closes #1243

## Test plan

- [x] New Zig unit test passes (`zig build test`)
- [x] Uncommented Scheme compliance test passes (14/14 in `r7rs-hygiene-gaps.scm`)
- [x] Full R7RS suite passes (1395/1395)
- [x] GC stress test passes (`-Dgc-stress=true`)
- [x] Full `run-all.sh` — no new failures

### DIFF
--- a/src/expander.zig
+++ b/src/expander.zig
@@ -691,8 +691,24 @@ fn instantiateEllipsis(gc: *GC, elem_template: Value, rest_template: Value, bind
         }
     }
 
-    // First instantiate the rest (after the ellipsis)
-    const result = try instantiateTemplate(gc, rest_template, bindings, intro_scope, literals, macro_keyword, globals, macros);
+    // Consecutive ellipses (R7RS 4.3.2): (x ... ...) flattens depth-2
+    // bindings into a single list.  Count and strip leading ellipsis
+    // tokens from rest_template so the tail instantiation sees only the
+    // non-ellipsis remainder.
+    var extra_ellipsis: u32 = 0;
+    var true_rest = rest_template;
+    while (true_rest != types.NIL and types.isPair(true_rest)) {
+        const head = types.car(true_rest);
+        if (types.isSymbol(head) and isEllipsis(types.symbolName(head))) {
+            extra_ellipsis += 1;
+            true_rest = types.cdr(true_rest);
+        } else {
+            break;
+        }
+    }
+
+    // First instantiate the rest (after all consumed ellipses)
+    const result = try instantiateTemplate(gc, true_rest, bindings, intro_scope, literals, macro_keyword, globals, macros);
     var result_root = result;
     gc.pushRoot(&result_root);
     defer gc.popRoot();
@@ -735,11 +751,44 @@ fn instantiateEllipsis(gc: *GC, elem_template: Value, rest_template: Value, bind
             }
             sub_count += 1;
         }
-        const expanded = try instantiateTemplate(gc, elem_template, sub_bindings[0..sub_count], intro_scope, literals, macro_keyword, globals, macros);
-        var expanded_root = expanded;
-        gc.pushRoot(&expanded_root);
-        result_root = try gc.allocPair(expanded_root, result_root);
-        gc.popRoot();
+
+        if (extra_ellipsis == 0) {
+            const expanded = try instantiateTemplate(gc, elem_template, sub_bindings[0..sub_count], intro_scope, literals, macro_keyword, globals, macros);
+            var expanded_root = expanded;
+            gc.pushRoot(&expanded_root);
+            result_root = try gc.allocPair(expanded_root, result_root);
+            gc.popRoot();
+        } else {
+            // Build synthetic template (elem_template ... ...) with
+            // extra_ellipsis ellipsis symbols, then instantiate it.
+            // instantiateTemplate will re-detect the ellipsis pattern and
+            // recurse into instantiateEllipsis with one fewer level.
+            const ellipsis_name = active_custom_ellipsis orelse "...";
+            var synth = types.NIL;
+            gc.pushRoot(&synth);
+            {
+                var ei: u32 = 0;
+                while (ei < extra_ellipsis) : (ei += 1) {
+                    const dots = try gc.allocSymbol(ellipsis_name);
+                    synth = try gc.allocPair(dots, synth);
+                }
+            }
+            synth = try gc.allocPair(elem_template, synth);
+
+            const expanded_list = try instantiateTemplate(gc, synth, sub_bindings[0..sub_count], intro_scope, literals, macro_keyword, globals, macros);
+            gc.popRoot(); // synth
+
+            // Splice expanded_list (a proper list) into result_root.
+            if (expanded_list != types.NIL and types.isPair(expanded_list)) {
+                var tail = expanded_list;
+                while (types.isPair(types.cdr(tail))) {
+                    tail = types.cdr(tail);
+                }
+                gc.writeBarrier(types.toObject(tail), result_root);
+                types.setCdr(tail, result_root);
+                result_root = expanded_list;
+            }
+        }
     }
 
     return result_root;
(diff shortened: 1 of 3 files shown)


src/expander.zig
```diff
@@ -735,11 +751,44 @@ fn instantiateEllipsis(gc: *GC, elem_template: Value, rest_template: Value, bind
...
+            var expanded_root = expanded;
+            gc.pushRoot(&expanded_root);
+            result_root = try gc.allocPair(expanded_root, result_root);
+            gc.popRoot();
+        } else {
+            // Build synthetic template (elem_template ... ...) with
+            // extra_ellipsis ellipsis symbols, then instantiate it.
+            // instantiateTemplate will re-detect the ellipsis pattern and
+            // recurse into instantiateEllipsis with one fewer level.
+            const ellipsis_name = active_custom_ellipsis orelse "...";
```
baijum: `synth` is loop-invariant — it's always `(elem_template <extra_ellipsis dots>)` and depends on neither `i` nor `sub_bindings`. As written it's rebuilt (fresh symbols + pairs) on every one of the `repeat_count` outer iterations. Consider hoisting the construction above the `while (i > 0)` loop and rooting it once with `gc.pushRoot(&synth); defer gc.popRoot();`. `instantiateTemplate` only reads the template, so sharing it across iterations is safe, and it also removes the manual `pushRoot`/`popRoot` pair below (whose `popRoot` is skipped on the `try` error paths, unlike the `defer` idiom used elsewhere in this function).

src/expander.zig
```diff
@@ -691,8 +691,24 @@ fn instantiateEllipsis(gc: *GC, elem_template: Value, rest_template: Value, bind
...
 
-    // First instantiate the rest (after the ellipsis)
-    const result = try instantiateTemplate(gc, rest_template, bindings, intro_scope, literals, macro_keyword, globals, macros);
+    // Consecutive ellipses (R7RS 4.3.2): (x ... ...) flattens depth-2
+    // bindings into a single list.  Count and strip leading ellipsis
+    // tokens from rest_template so the tail instantiation sees only the
+    // non-ellipsis remainder.
+    var extra_ellipsis: u32 = 0;
+    var true_rest = rest_template;
+    while (true_rest != types.NIL and types.isPair(true_rest)) {
```
baijum: Minor: `true_rest != types.NIL` is redundant with `types.isPair(true_rest)` — NIL is never a pair — so the loop guard could just be `while (types.isPair(true_rest))`. (The same shorthand is already used a few lines down in the splice at `types.isPair(expanded_list)`.)

baijum (commented):
Reviewed the fix for #1243 (doubled ellipsis `(x ... ...)` flattening). **The approach is correct and nicely minimal** — stripping the *consecutive* leading ellipses from `rest_template`, then rebuilding a synthetic `(elem ...)` template and letting `instantiateTemplate` re-enter `instantiateEllipsis` one level shallower. I traced the interesting cases and they all hold up:

- **Recurses to arbitrary depth**: `(x ... ... ...)` at depth 3 strips 2, builds a 2-ellipsis synthetic, and unwinds correctly one level per recursion.
- **Trailing/leading content is preserved**: the strip loop stops at the first non-ellipsis, so `(x ... ... tail)` and `(a x ... ... b)` keep `tail`/`b` in `true_rest` rather than over-consuming.
- **Independent groups aren't merged**: `(x ... y ...)` → `after` is `(y ...)`, whose head isn't an ellipsis, so `extra_ellipsis` stays 0 and the two groups concatenate as R7RS requires.
- **GC safety**: `result_root` stays rooted across the synthetic instantiation, `synth` is rooted while it's built and read, the splice does no allocation between `popRoot` and `setCdr`, and the `writeBarrier`-before-`setCdr` order matches the codebase convention in `set-cdr!` (`primitives.zig:402`).
- **Empty/no-group edges** (`(() ())`, `()`) short-circuit to `()` correctly, and the destructive splice is safe because `expanded_list` is freshly allocated and unshared.

Test coverage is good (unit test + re-enabled compliance test). A couple of non-blocking suggestions inline, plus: consider adding a test for a **custom ellipsis** combined with doubling (e.g. `(syntax-rules ::: () ((_ ((x :::) :::)) '(x ::: :::)))`), since the synthetic path leans on `active_custom_ellipsis` and is currently untested. LGTM overall.

baijum:
Re-reviewed after `4be85b9` — all three points from my earlier review are addressed:

- **`synth` hoisted out of the loop** — now built once before `while (i > 0)`, rooted with `pushRoot` + a matching `defer if (extra_ellipsis > 0) gc.popRoot()`. It stays rooted across every iteration and `instantiateTemplate` only reads it, so sharing it is safe. Push/pop is balanced and LIFO-correct against the `result_root` root above it.
- **Guard simplified** to `while (types.isPair(true_rest))`.
- **Custom-ellipsis test added** (`flatten-custom` with `:::`), exercising the `active_custom_ellipsis` path.

Verified locally on the PR head:
- `zig build test` — passes
- Re-enabled compliance test `r7rs-hygiene-gaps.scm` — 14/14, 0 failures
- Spot checks: `(x ... ...)` → `(1 2 3 4 5)`, custom `:::` → `(1 2 3 4)`, empty groups `((1) () (2 3))` → `(1 2 3)`, and surrounding content `(a x ... ... b)` → `(a 1 2 3 4 b)` (confirms the strip loop doesn't over-consume).

One remaining nit, non-blocking: if an allocation inside the `if (extra_ellipsis > 0)` synth-build block fails (OOM), `pushRoot` has already run but the `defer` that pops it is registered *after* the block, so that one root leaks on the error path. It only bites on OOM-during-expansion (effectively fatal) and matches the existing style elsewhere in the function, so I wouldn't hold the PR for it.

**LGTM — ready to merge.**

PR #1279, opened 2026-07-06
Compile eval body in tail position per R7RS 3.5

### LINKED ISSUE #1253: [R7RS 3.5] eval is not tail-called — one frame per nested eval, StackOverflow at the 32768 frame cap (#1240 residual)
**Follow-up from #1240** (residual case; the other two — call-with-values consumer and call/cc receiver — are verified fixed).

**Spec reference:** R7RS 3.5 (p. 12): "eval must evaluate its first argument as if it were in tail position within the eval procedure."
**Severity:** conformance / graceful-error (downgraded from #1240's P0: pre-a91e908 this was an uncatchable GC-root-stack panic at depth ~1024; now it is a catchable `error.StackOverflow` at the frame cap).

**Behavior at f13e806 (ReleaseSafe):** each nested `eval` consumes one VM call frame, so an eval-driven tail loop hits the 32768 frame cap instead of running in constant space:

```scheme
(define (loop-eval n)
  (if (= n 0) 'done
      (eval (list 'loop-eval (- n 1)) (interaction-environment))))
(loop-eval 30000)   ; => done
(loop-eval 33000)   ; => runtime error: error.StackOverflow  (cap = 32768 frames)
(loop-eval 1000000) ; expected per R7RS: done
```

For contrast, the two fixed #1240 cases run at N=1,000,000 on the same build.

**Test-suite note:** the "eval tail evaluation" test in `tests/scheme/compliance/r7rs-tail-procedures-gaps.scm` uses N=25000, which sits *under* the 32768 frame cap — it passes without eval being tail-called (the file's "N=25000 proves TCO" rationale assumed register-file exhaustion at 16k–20k, but the eval path consumes frames, not register windows). When fixing this, raise that test's N above 32768 (e.g. 40000); until then the suite cannot detect this gap.

**Likely mechanism:** the `.tail_eval` opcode replaces the current frame's closure/ip (vm_dispatch.zig), but the measured one-frame-per-iteration growth means some step in the eval→compile→execute chain still pushes a frame per round (possibly the compiled top-level function's execution frame). Needs tracing through `tail_eval` / `compileExpressionWithMacros` / the execute path.

Found by: systematic audit Phase 4A (#1137), split out of #1240 after post-#1245 verification.

baijum:
Fixes #1253

## Summary

- Thread `is_tail` through `Compiler.compile()` → `compileExpressionWithMacrosAt()` → `compileExpressionInEnv()` so the `tail_eval` opcode compiles expressions with tail-position awareness
- Add `Function.restricted_globals` flag to suppress `get_global` fallback to VM globals for restricted environments (`null-environment`), fixing a latent bug exposed by tail-position compilation
- Raise tail-procedures test N from 25000 to 50000 (above the 32768 frame cap) so it actually detects non-tail eval

## Test plan

- [x] `zig build test` — 648 unit tests pass (including new `#1253` regression test at N=40000)
- [x] `bash tests/scheme/run-all.sh` — 1782 pass, 0 fail
- [x] `(loop-eval 50000)` completes without StackOverflow
- [x] `(eval '(car '(1)) (null-environment 5))` still errors (restricted env respected)

### DIFF
--- a/src/compiler.zig
+++ b/src/compiler.zig
@@ -386,7 +386,7 @@ pub const Compiler = struct {
 
     // -- Public compilation API --
 
-    pub fn compile(self: *Compiler, expr: Value) CompileError!void {
+    pub fn compile(self: *Compiler, expr: Value, is_tail: bool) CompileError!void {
         // Root the source datum for the whole compile: the expander and the
         // derived-form compilers allocate (triggering GC), and the datum tree
         // is otherwise reachable only through this unrooted argument. Without
@@ -412,10 +412,10 @@ pub const Compiler = struct {
         ir.compiler = self;
         ir.set_targets = self.set_targets;
         defer ir.deinit();
-        const root = try ir_mod.lowerAndOptimize(&ir, expr_root, &self.macros, false);
+        const root = try ir_mod.lowerAndOptimize(&ir, expr_root, &self.macros, is_tail);
 
         const dst = try self.allocReg();
-        try compiler_ir.compileFromNode(self, root, dst, false);
+        try compiler_ir.compileFromNode(self, root, dst, is_tail);
         try self.emitOp(.@"return");
         try self.emitU16(dst);
 
@@ -854,16 +854,16 @@ pub fn compileExpression(gc: *memory.GC, expr: Value) CompileError!*types.Functi
         if (!ok) Compiler.unrootFunction(gc, c.func);
         c.deinit();
     }
-    try c.compile(expr);
+    try c.compile(expr, false);
     ok = true;
     return c.func;
 }
 
 pub fn compileExpressionWithMacros(gc: *memory.GC, expr: Value, vm_macros: *std.StringHashMap(Value), vm_globals: ?*std.StringHashMap(Value)) CompileError!*types.Function {
-    return compileExpressionWithMacrosAt(gc, expr, vm_macros, vm_globals, 0, null);
+    return compileExpressionWithMacrosAt(gc, expr, vm_macros, vm_globals, 0, null, false);
 }
 
-pub fn compileExpressionWithMacrosAt(gc: *memory.GC, expr: Value, vm_macros: *std.StringHashMap(Value), vm_globals: ?*std.StringHashMap(Value), source_line: u32, source_name: ?[]const u8) CompileError!*types.Function {
+pub fn compileExpressionWithMacrosAt(gc: *memory.GC, expr: Value, vm_macros: *std.StringHashMap(Value), vm_globals: ?*std.StringHashMap(Value), source_line: u32, source_name: ?[]const u8, is_tail: bool) CompileError!*types.Function {
     syntax_error_detail_len = 0;
     var c = try Compiler.init(gc);
     const roots_base = gc.extra_roots.items.len;
@@ -880,7 +880,7 @@ pub fn compileExpressionWithMacrosAt(gc: *memory.GC, expr: Value, vm_macros: *st
     while (it.next()) |entry| {
         c.macros.put(entry.key_ptr.*, entry.value_ptr.*) catch return CompileError.OutOfMemory;
     }
-    try c.compile(expr);
+    try c.compile(expr, is_tail);
     var out_it = c.macros.iterator();
     while (out_it.next()) |entry| {
         vm_macros.put(entry.key_ptr.*, entry.value_ptr.*) catch return CompileError.OutOfMemory;
@@ -889,7 +889,7 @@ pub fn compileExpressionWithMacrosAt(gc: *memory.GC, expr: Value, vm_macros: *st
     return c.func;
 }
 
-pub fn compileExpressionInEnv(gc: *memory.GC, expr: Value, vm_macros: *std.StringHashMap(Value), env: *std.StringHashMap(Value), env_val: Value) CompileError!*types.Function {
+pub fn compileExpressionInEnv(gc: *memory.GC, expr: Value, vm_macros: *std.StringHashMap(Value), env: *std.StringHashMap(Value), env_val: Value, is_tail: bool) CompileError!*types.Function {
     syntax_error_detail_len = 0;
     var c = try Compiler.init(gc);
     const roots_base = gc.extra_roots.items.len;
@@ -907,9 +907,10 @@ pub fn compileExpressionInEnv(gc: *memory.GC, expr: Value, vm_macros: *std.Strin
     while (it.next()) |entry| {
         c.macros.put(entry.key_ptr.*, entry.value_ptr.*) catch return CompileError.OutOfMemory;
     }
-    try c.compile(expr);
+    try c.compile(expr, is_tail);
     c.func.env = env;
     c.func.env_val = env_val;
+    c.func.restricted_globals = c.restricted_env;
     if (!types.isEnvironment(env_val) or !types.toEnvironment(env_val).immutable) {
         var out_it = c.macros.iterator();
         while (out_it.next()) |entry| {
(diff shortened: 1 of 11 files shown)


src/tests_core_eval.zig
```diff
@@ -271,3 +271,19 @@ test "parameterize evaluates values before binding (#1202)" {
     const r2 = try ctx.vm.eval("(parameterize ((b (a)) (a 2)) (b))");
     try std.testing.expectEqual(@as(i64, 1), types.toFixnum(r2));
 }
+
+// Regression for #1253: eval must be tail-called (R7RS 3.5)
+test "eval tail position runs in constant frame depth (#1253)" {
```
baijum, line 323: This test covers the **frame-depth** half of the fix (interaction-environment, N=40000 > 32768 cap — fails pre-fix, good). But the **`restricted_globals`** half — null-environment no longer leaking VM globals in tail position — has no regression test.

All existing null-environment tests (`tests/scheme/smoke/eval-environment.scm`, `tests/scheme/audit/primitives_r7rs-audit.scm:80-81`) run `eval` in **non-tail** position, so the operator compiles via `tail_call_global`/`call_global` (`lookupGlobalLocked`, no globals fallback) and already errored correctly — they never reach the `get_global` path this flag guards. That path is only taken for a **tail-position** call operator.

Suggest adding a test that puts a null-environment eval in tail position and asserts it still errors, e.g. a helper whose body is `(eval '(car '(1)) (null-environment 5))` invoked under `guard`. That exercises the exact `get_global` leak `restricted_globals` fixes and would fail without it — right now nothing in the suite would catch a regression of the flag.

src/compiler.zig
```diff
@@ -870,9 +870,10 @@ pub fn compileExpressionInEnv(gc: *memory.GC, expr: Value, vm_macros: *std.Strin
     while (it.next()) |entry| {
         c.macros.put(entry.key_ptr.*, entry.value_ptr.*) catch return CompileError.OutOfMemory;
     }
-    try c.compile(expr);
+    try c.compile(expr, is_tail);
     c.func.env = env;
     c.func.env_val = env_val;
+    c.func.restricted_globals = c.restricted_env;
```
baijum, line 913: `restricted_globals` is set from `c.restricted_env`, which `compileExpressionInEnv` sets to `true` **unconditionally**. That function is shared beyond eval/load-with-env: `compileLibExpr` (library bodies, `src/vm_library.zig:929`) and `handleDefineRecordType`'s in-library branch (`src/vm_records.zig`) both go through it. So this also suppresses the `get_global → vm.globals` fallback for **library-body** compilation, not just null-environment.

It looks empirically safe (run-all passes) because library bodies reference imported names via `lib_env`, calls resolve through `lookupGlobalLocked(lib_env)`, and the hygienic-prefix→globals fallback in `get_global` (`src/vm_dispatch.zig:226-228`) is *not* gated by this flag, so macro-introduced references still resolve. But it's a behavior change beyond the stated scope: a library body that referenced a plain (non-hygienic) global *as a value* that lived only in `vm.globals` and not in `lib_env` would now get an undefined-variable error instead of silently resolving.

Could you confirm this is intended? If the intent is strictly eval/null-environment, it'd be safer to set `restricted_globals` only for the user-eval/load callers rather than inside the shared `compileExpressionInEnv`.

tests/scheme/compliance/r7rs-tail-procedures-gaps.scm
```diff
@@ -5,16 +5,16 @@
...
+;; consumes one frame per iteration will hit StackOverflow.
 ;; Spec references cite docs/errata-corrected-r7rs.pdf section 3.5 (p. 12).
 
 (import (scheme base) (scheme eval) (scheme repl) (scheme write)
         (scheme process-context) (srfi 64))
 
 (test-begin "r7rs-tail-procedures-gaps")
 
-(define N 25000)
+(define N 50000)
```
baijum: Doubling N to 50000 may push this file over the 60s per-file timeout on Debug CI. The header (lines 5-6, unchanged) explicitly notes the split exists to stay under that timeout, at ~0.6ms/iter for the let-values loop. There are **8** loops that each run N iterations (`loop-lv/lsv/lsyn/lrsyn/apply/cwv/cc/eval`): at that estimate `loop-lv` alone is ~30s at N=50000, `loop-eval` recompiles an expression on every iteration (50000 compiles), and `loop-cc` captures a continuation each iteration.

`run-all.sh` now counts a timeout as a **failure** (`run-all.sh:157`, folded into the fail total), so tipping over 60s on a Debug build fails the run. Since the goal is just to exceed the 32768 frame cap, **N=40000** (matching the new unit test's N) is sufficient and keeps ~20% more headroom. Recommend either lowering to 40000, or verifying wall-clock with `zig build -Doptimize=Debug` before merge — the test-plan's run-all pass was presumably on the fast (ReleaseSafe) build, which won't reflect the Debug-CI budget this comment refers to.

src/vm_dispatch.zig
```diff
@@ -217,7 +217,7 @@ pub fn runUntil(self: *VM, target_frame_count: usize, target_wind_count: usize)
                 // path runs after release (findSimilarName locks internally).
                 self.lockGlobalsShared();
                 const found: ?Value = env.get(name) orelse blk: {
-                    if (func.env != null) {
+                    if (func.env != null and !func.restricted_globals) {
```
baijum, line 220: Good — this makes `get_global` consistent with `call_global`/`tail_call_global`, whose `lookupGlobalLocked` never falls back to `self.globals`.

Worth a one-line comment here for future readers: the hygienic-prefix fallback a few lines down (`if (env != self.globals) { self.globals.get(base) }`) is intentionally left **ungated**, so hygiene-renamed references from imported macros still resolve through globals even in a restricted env — only the direct same-name fallback is suppressed. That asymmetry isn't obvious and is easy to "tidy up" incorrectly later.

baijum (commented):
## Review

The root-cause diagnosis is correct and the fix is well-targeted. Threading `is_tail` so `tail_eval` compiles the body in tail position is the right fix, and I traced the stated mechanism end to end: in `compileCallFromIR`, a tail call to a *non-self* global (e.g. `car`) does **not** go through `compileCallGlobalFromIR` — it falls to the generic path (`src/compiler_ir.zig:140-168`) which compiles the operator as `get_global` + `tail_call`. `get_global` was the only global-lookup path with a `func.env != null → self.globals` fallback (`call_global`/`tail_call_global` use `lookupGlobalLocked`, which never falls back), so enabling tail compilation for `eval` is exactly what exposed the null-environment leak. `restricted_globals` closes that gap cleanly.

Three things worth addressing before merge — one test-coverage gap, one scope question, one CI-timing risk. Details inline.

src/tests_core_eval.zig
baijum: Added: `test "null-environment eval in tail position respects restriction (#1253)"` — puts eval in tail position via guard (which desugars the body into a lambda), asserts `car` is not resolved. Fails without `restricted_globals`.

src/compiler.zig
baijum: Intentional. `compileExpressionInEnv` already sets `restricted_env = true` unconditionally — `restricted_globals` extends that same restriction to runtime `get_global` lookups, keeping compile-time and runtime behavior consistent. Library bodies reference imported names via `lib_env`, and the hygienic-prefix fallback (`vm_dispatch.zig:226-228`) is ungated so macro-introduced references still resolve. A library body referencing a plain un-imported global that lived only in `vm.globals` was arguably a latent bug — run-all confirms nothing relies on it.

tests/scheme/compliance/r7rs-tail-procedures-gaps.scm
baijum: Lowered to N=40000 — still exceeds the 32768 cap with headroom, matches the unit test's N.

src/vm_dispatch.zig
baijum: Added the comment. Good catch — that asymmetry is exactly the kind of thing that gets 'fixed' incorrectly in a future cleanup.

PR #1280, opened 2026-07-06
Follow redirect chain in force for delay-force intermediates

### LINKED ISSUE #1264: force returns #<promise> instead of the memoized value for intermediate promises of a completed delay-force chain
## Summary

After a `delay-force` chain completes, forcing an **intermediate** promise of that chain returns the outer `#<promise>` object instead of the memoized value. Found while reviewing #1259; the bug is pre-existing — it reproduces on `main` (088472e) both with and without that PR's change, in both cases identically.

## Reproduction

```scheme
(import (scheme base) (scheme lazy) (scheme write))
(define inner (delay 42))
(define outer (delay-force inner))
(display (force outer)) (newline)   ; 42        — correct
(display (force inner)) (newline)   ; #<promise> — WRONG, expected 42
(display (promise? (force inner))) (newline)   ; #t
```

R7RS 4.2.5 / SRFI-45 semantics: `inner` is `(delay 42)`, so `(force inner)` must yield `42` regardless of whether the chain it participated in was forced through `outer` first. In SRFI-45's reference implementation the merge makes inner and outer share content, so re-forcing the intermediate yields the final value.

## Mechanism

In `forceFn` (`src/primitives_lazy.zig`), when a thunk returns an unforced promise, the merge step redirects the inner promise at the outer one:

```zig
inner.forced = true;
inner.value = types.makePointer(@ptrCast(promise));   // redirect: inner -> outer
```

But the memoization check at the top of the trampoline loop returns that redirect pointer directly instead of following it:

```zig
if (promise.forced) return promise.value;   // returns the outer *promise object*
```

## Suggested fix

Keep trampolining instead of returning:

```zig
if (promise.forced) {
    current = promise.value;
    continue;
}
```

The loop head's `!types.isPromise(current)` check terminates for normal memoized values. In this implementation a *forced* promise's value can only be another promise via the redirect path (the trampoline flattens promise results before memoizing), so following is safe. Worth checking whether multi-hop redirects (inner1 → inner2 → outer) can arise and whether path compression is worthwhile.

## Impact

Any code that keeps a reference to an intermediate promise of a `delay-force` chain and forces it after the chain has been driven — SRFI-41-style streams are the natural victim.

## Notes

- Per repo convention, the fix needs a regression test (e.g. in `tests/scheme/compliance/lazy.scm` or `r7rs-thin-forms-gaps.scm`) asserting `(force inner)` ⇒ `42` after `(force outer)`.
- Same function as #1259 (trampoline cap removal) — coordinate/rebase accordingly.
- Original analysis: https://github.com/kaappi/kaappi/pull/1259#discussion_r3528776079

baijum:
Fixes #1264

## Summary

- `force` returns `#<promise>` instead of the memoized value when re-forcing an intermediate promise of a completed `delay-force` chain
- The SRFI-45 merge step redirects intermediate promises to point at the chain head, but the trampoline returned these redirect pointers directly at three sites instead of following them
- Changed all three `return promise.value` / `return inner.value` sites in `forceFn` to `current = ...; continue;` so the trampoline resolves the full chain

## Test plan

- [x] Added regression tests for two-deep and three-deep `delay-force` chains forcing intermediate promises
- [x] Lazy compliance tests pass (14/14)
- [x] R7RS test suite passes (0 failures)
- [x] Zig unit tests pass

### DIFF
--- a/src/primitives_lazy.zig
+++ b/src/primitives_lazy.zig
@@ -45,6 +45,7 @@ fn forceFn(args: []const Value) PrimitiveError!Value {
         const promise = types.toPromise(current);
 
         if (promise.forced) {
+            // Follow redirect: forced value is a promise only via SRFI-45 merge (inner → head), never a cycle.
             current = promise.value;
             continue;
         }
--- a/src/primitives_lazy.zig
+++ b/src/primitives_lazy.zig
@@ -44,7 +44,10 @@ fn forceFn(args: []const Value) PrimitiveError!Value {
         if (!types.isPromise(current)) return current;
         const promise = types.toPromise(current);
 
-        if (promise.forced) return promise.value;
+        if (promise.forced) {
+            current = promise.value;
+            continue;
+        }
 
         const thunk = promise.value;
         if (!types.isProcedure(thunk)) {
@@ -63,7 +66,8 @@ fn forceFn(args: []const Value) PrimitiveError!Value {
         // has already completed this promise (re-entrant force).
         if (promise.forced) {
             promise.forcing = false;
-            return promise.value;
+            current = promise.value;
+            continue;
         }
 
         if (types.isPromise(result)) {
@@ -80,7 +84,8 @@ fn forceFn(args: []const Value) PrimitiveError!Value {
                 promise.forced = true;
                 promise.value = inner.value;
                 gc.writeBarrier(&promise.header, inner.value);
-                return inner.value;
+                current = inner.value;
+                continue;
             }
             promise.value = inner.value;
             gc.writeBarrier(&promise.header, inner.value);
--- a/tests/scheme/compliance/lazy.scm
+++ b/tests/scheme/compliance/lazy.scm
@@ -39,6 +39,21 @@
                   (if (> count x) count (force p)))))
   (test-eqv "recursive force with termination" 6 (force p)))
 
+;; Forcing intermediate promise of a completed delay-force chain (#1264)
+(test-group "delay-force chain intermediate"
+  (let ()
+    (define inner (delay 42))
+    (define outer (delay-force inner))
+    (test-eqv "force outer yields value" 42 (force outer))
+    (test-eqv "force inner after chain yields value" 42 (force inner)))
+  (let ()
+    (define a (delay 42))
+    (define b (delay-force a))
+    (define c (delay-force b))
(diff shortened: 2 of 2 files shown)


baijum (commented):
## Review — ready to merge ✅

Focused, correct fix. Verified independently (built, applied, reverted-to-confirm): reproduction from #1264 now yields `42`/`#f`; the added regression test **fails without the source change** (3 failures: intermediate, middle, innermost) and passes with it; lazy compliance 14/14; R7RS suite 0 failures. All CI green.

### Why the fix is sound

The three sites all change `return <value>` → `current = <value>; continue;` so the trampoline follows the SRFI-45 redirect pointer (`inner.value = &outer`) instead of handing it back as a `#<promise>`. The key safety question is: *can following `promise.value` ever misinterpret a legitimately-memoized promise, or loop forever?* Both are no, and it's worth recording why:

- **A forced promise's `value` is a promise only via the redirect path.** `delay` and `delay-force` compile identically (both `%make-promise-lazy`, `compiler_lambda.zig:861`), and `forceFn` always trampolines *into* a promise result (the `isPromise(result)` branch) rather than memoizing it — a value is memoized as-is only at the non-promise tail (`promise.value = result` after the `isPromise` check). So a forced promise whose `value` is a promise can only be a redirect.
- **No infinite loop.** Every redirect points at the single head promise being driven, whose final `value` is a non-promise. So `intermediate → head → value` terminates in ≤2 hops; there are no redirect cycles. I confirmed this with the three-deep chain — all of `a`, `b` redirect directly to head `c` (no `inner1 → inner2 → outer` chaining), so the multi-hop / path-compression concern raised in the issue doesn't actually arise here.

### Optional (non-blocking)

1. **Add a one-line comment** at the top-of-loop `if (promise.forced)` site explaining *why* we follow rather than return (the redirect invariant above). It's the subtlest part of the change and the next reader will wonder if it can loop. The rationale is already written out in the issue.
2. **Exotic edge case, not introduced by this PR:** re-entrant `(force intermediate)` invoked *from inside a thunk while the head is still mid-drive* will now follow to the not-yet-forced head and re-invoke its thunk (there's no `promise.forcing` guard before the top-of-loop thunk call). This requires a thunk to capture and force an intermediate of its own chain — extremely unusual, and the old code returned a wrong value there anyway. Worth a mental note, not a blocker.

Nice, well-scoped fix with a genuine regression test.

PR #1281, opened 2026-07-06
Check record type in accessors and mutators (#1199)

### LINKED ISSUE #1199: [R7RS 5.5] record accessors/mutators do not check the record type — cross-type access silently reads/writes the wrong field
**Spec reference:** R7RS §5.5, page 27: "Each ⟨accessor name⟩ is bound to a procedure that takes a record of type ⟨name⟩ and returns the current value of the corresponding field. It is an error to pass an accessor a value which is not a record of the appropriate type." (Likewise for modifiers: "It is an error to pass a modifier a first argument which is not a record of the appropriate type.")
**Severity:** edge-case ("is an error" latitude, but silent wrong-field reads/writes are data corruption; non-record arguments DO raise, so current behavior is inconsistent)

**Minimal reproduction:**
```scheme
(define-record-type point (make-point x) point? (x point-x set-point-x!))
(define-record-type blob  (make-blob y)  blob?  (y blob-y))

(point-x (make-blob 9))          ;; expected: error   actual: 9 (reads blob's y field)
(set-point-x! (make-blob 9) 42)  ;; expected: error   actual: silently overwrites blob's y

(point-x 42)                     ;; correctly raises (non-record) — the check exists only for non-records
(point? (make-blob 9))           ;; => #f — predicates DO discriminate by type
```

**Root cause:** `handleDefineRecordType` (src/vm_records.zig) desugars accessors to `(define (acc p) (%record-ref p <idx>))` and mutators to `(define (mut p v) (%record-set! p <idx> v))` with no `%record?` type guard; `%record-ref`/`%record-set!` (src/primitives.zig) only validate "is any record instance" and index bounds, so any record type with enough fields is silently accepted.

**Suggested fix:** desugar to `(define (acc p) (if (%record? p <type>) (%record-ref p <idx>) (error ...)))`, or teach `%record-ref`/`%record-set!` to take the expected record type and compare `ri.record_type`.

**Source location:** src/vm_records.zig (accessor/mutator generation), src/primitives.zig (`recordRefFn`, `recordSetFn`)
**Disabled test:** tests/scheme/audit/primitives_core-audit.scm (marked ;; FAIL: #this)

**Found by:** Systematic audit, Phase 2.18

baijum:
## Summary

- Record accessors and mutators now verify the argument is an instance of the correct record type, not just any record
- Passes the expected `RecordType` pointer to `%record-ref` / `%record-set!` primitives, which compare it against the instance's `record_type` field
- All three desugaring paths updated: VM top-level (`handleDefineRecordType`), body-context expansion (`expandRecordTypeDefines`), and compiler dispatch (`compileDefineRecordType`)

Fixes #1199

## Test plan

- [x] New regression test `tests/scheme/smoke/record-type-check.scm` — 10 tests covering cross-type accessor/mutator errors, non-record errors, correct-type access, and body-context records
- [x] Enabled previously disabled audit tests in `primitives_core-audit.scm`
- [x] Zig unit tests pass
- [x] Full suite: 1785 pass, 0 fail

### DIFF
--- a/src/compiler_lambda.zig
+++ b/src/compiler_lambda.zig
@@ -509,7 +509,8 @@ pub fn compileDefineRecordType(self: *Compiler, args: Value, dst: u16) CompileEr
                 const p = gc.allocSymbol("p") catch return CompileError.OutOfMemory;
                 const v = gc.allocSymbol("v") catch return CompileError.OutOfMemory;
                 const idx = types.makeFixnum(@intCast(fi));
-                const body = gc.makeList(&[_]Value{ rs, p, idx, v }) catch return CompileError.OutOfMemory;
+                const rt_ref = gc.allocSymbol(internal_name) catch return CompileError.OutOfMemory;
+                const body = gc.makeList(&[_]Value{ rs, p, idx, v, rt_ref }) catch return CompileError.OutOfMemory;
                 const np = gc.allocPair(gc.allocSymbol(mn) catch return CompileError.OutOfMemory, gc.makeList(&[_]Value{ p, v }) catch return CompileError.OutOfMemory) catch return CompileError.OutOfMemory;
                 forms = gc.allocPair(gc.makeList(&[_]Value{ define_sym, np, body }) catch return CompileError.OutOfMemory, forms) catch return CompileError.OutOfMemory;
             }
@@ -524,7 +525,8 @@ pub fn compileDefineRecordType(self: *Compiler, args: Value, dst: u16) CompileEr
             const rr = gc.allocSymbol("%record-ref") catch return CompileError.OutOfMemory;
             const p = gc.allocSymbol("p") catch return CompileError.OutOfMemory;
             const idx = types.makeFixnum(@intCast(fi));
-            const body = gc.makeList(&[_]Value{ rr, p, idx }) catch return CompileError.OutOfMemory;
+            const rt_ref = gc.allocSymbol(internal_name) catch return CompileError.OutOfMemory;
+            const body = gc.makeList(&[_]Value{ rr, p, idx, rt_ref }) catch return CompileError.OutOfMemory;
             const np = gc.allocPair(gc.allocSymbol(spec.accessor_names[fi]) catch return CompileError.OutOfMemory, gc.makeList(&[_]Value{p}) catch return CompileError.OutOfMemory) catch return CompileError.OutOfMemory;
             forms = gc.allocPair(gc.makeList(&[_]Value{ define_sym, np, body }) catch return CompileError.OutOfMemory, forms) catch return CompileError.OutOfMemory;
         }
--- a/src/primitives.zig
+++ b/src/primitives.zig
@@ -160,8 +160,8 @@ const core_specs = [_]PrimSpec{
     .{ .name = "%make-record-type", .func = &makeRecordTypeFn, .arity = .{ .exact = 2 }, .libs = LS.initOne(.scheme_base) },
     .{ .name = "%make-record", .func = &makeRecordFn, .arity = .{ .variadic = 1 }, .libs = LS.initOne(.scheme_base) },
     .{ .name = "%record?", .func = &recordCheckFn, .arity = .{ .exact = 2 }, .libs = LS.initOne(.scheme_base) },
-    .{ .name = "%record-ref", .func = &recordRefFn, .arity = .{ .exact = 2 }, .libs = LS.initOne(.scheme_base) },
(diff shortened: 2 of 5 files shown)


baijum (commented):
## Review

Solid fix for #1199. The core behaviour is correct: cross-type accessor/mutator
access now raises instead of silently reading/writing the wrong field, and all
three desugaring paths (`handleDefineRecordType`, `expandRecordTypeDefines`,
`compileDefineRecordType`) are updated consistently. I built the branch and
verified:

- `(point-x (make-blob 9))` → errors (was: silently returned the blob's field) ✅
- non-record and correct-type paths still behave correctly ✅
- new `record-type-check.scm` → 10/10 pass, audit suite 201/201 pass ✅

GC safety is fine — the new `%__record_type_X%` symbol allocations all sit
inside the existing `no_collect` guarded regions in both the VM and compiler
paths, and the compiler path also appends each init to `extra_roots`.

### 🔴 Should fix before merge: `%record-ref` / `%record-set!` can panic the interpreter

`recordRefFn` / `recordSetFn` now dereference the expected-type argument
**before validating it**:

```zig
// src/primitives.zig:827
const rt = types.toObject(args[2]).as(types.RecordType);
```

`types.toObject` just truncates the NaN-box payload to a pointer and
`Object.as()` only asserts the tag in **Debug** builds (`std.debug.assert`) —
default builds are ReleaseSafe, where it's an unchecked `@ptrCast(@alignCast(...))`.
So passing a non-record-type as the type argument dereferences a bogus pointer.

Because these `%`-prefixed primitives are registered into `(scheme base)` and
are reachable from user code (the audit suite itself calls `%record-ref`
directly), this is triggerable from pure Scheme and is **not catchable by
`guard`**:

```scheme
(%record-ref (make-point 1 2) 0 5)      ;; thread panic: incorrect alignment  → abort
(%record-set! (make-point 1 2) 0 99 5)  ;; thread panic: incorrect alignment  → abort
```

This is a regression: previously these primitives validated every argument and
returned a proper Scheme error for garbage input (the audit file even has a
section titled *"internal record primitives reject garbage directly"*).

Fix is one guard each, matching the existing pattern already used by
`makeRecordFn` / `recordCheckFn` in the same file:

```zig
if (!types.isRecordType(args[2])) return typeError("%record-ref", "record-type", args[2]);
const rt = types.toObject(args[2]).as(types.RecordType);
```

(and `args[3]` for `%record-set!`). Worth a regression test with a bad
type argument alongside the existing cases.

### 🟡 Minor: audit test intent weakened

`tests/scheme/audit/primitives_core-audit.scm:309` — `(%record-ref 5 0)` now
hits an **arity** error (arity changed to `exact = 3`) rather than the intended
type-rejection. It still passes (arity error is an `error-object?`), but no
longer exercises "rejects a non-record instance". Consider updating it to the
3-arg form plus a bad-type-arg case — this pairs naturally with the guard above.

### Note (pre-existing, not introduced here)

The internal type binding is keyed on the type name (` __record_type_<name>`),
so two top-level record types sharing a name share the binding. The constructor
already referenced this same binding before the PR, and body-context definitions
are locally scoped, so this PR doesn't change the situation — just flagging.

---

**Verdict:** The feature itself is correct, well-tested, and GC-safe. I'd
recommend adding the two `isRecordType` guards (and ideally the paired test)
before merging, since they cheaply restore a robustness property the codebase
already relies on and tests for. With that addressed, this is ready to merge.

baijum (commented):
## Follow-up review — commit 5b23a85

The new commit resolves everything from my previous review. I rebuilt at
`5b23a85` and re-verified:

- **`%record-ref` / `%record-set!` guard** — `isRecordType` checks added before
  the `toObject().as(RecordType)` downcast in both primitives. The calls that
  previously aborted the interpreter now raise a catchable Scheme error:
  ```scheme
  (%record-ref  (make-point 1 2) 0 5)    ;; => caught (was: panic: incorrect alignment)
  (%record-set! (make-point 1 2) 0 99 5) ;; => caught (was: panic: incorrect alignment)
  ```
- **Audit test** — updated to the 3-arg `(%record-ref 5 0 #t)` form plus a new
  `(%record-set! 5 0 99 #t)` case. Suite: 202 pass, 0 fail.
- **Regression test** — two bad-type-arg cases added; `record-type-check.scm`
  now 12/12 pass.
- `zig build test` — full unit suite passes.

Guards match the existing `makeRecordFn` / `recordCheckFn` pattern. Both my
findings are addressed and the minor note was pre-existing.

**LGTM — ready for merge.** ✅

PR #1282, opened 2026-07-06
Reject non-environment second argument to eval (#1270)

### LINKED ISSUE #1270: eval silently ignores a non-environment second argument instead of raising a type error
## Summary

`eval` silently ignores its second argument when it is not an environment,
instead of raising a type error. R7RS §6.12 specifies the second argument as an
environment specifier, and a non-environment value should be rejected. `load`
(as of #1262) correctly rejects the analogous case, so the two procedures now
disagree.

## Root cause

`evalFn` (`src/primitives_r7rs.zig:172`) only takes the environment path when
the second argument *is* an environment; otherwise it silently falls through to
the global-environment path:

```zig
// src/primitives_r7rs.zig:178
if (args.len > 1 and types.isEnvironment(args[1])) {
    ...compile in that environment...
}
// falls through here for a non-environment 2nd arg — no error
const func = compiler_mod.compileExpressionWithMacros(gc, expr, &vm.macros, vm.globals) ...
```

Compare `loadFn`, which rejects a non-environment specifier explicitly:

```zig
// src/primitives_r7rs.zig (load env branch, added in #1262)
if (!types.isEnvironment(args[1])) return primitives.typeError("load", "environment", args[1]);
```

## Reproduction

```scheme
(import (scheme base) (scheme write) (scheme eval))

;; 42 is not an environment specifier:
(display (guard (e (#t 'caught-error))
           (eval '(+ 1 2) 42)))
(newline)
```

**Actual:** prints `3` — the bad second argument is silently ignored and the
expression is evaluated in the global environment.
**Expected:** a type error (`'caught-error`), matching `load`'s behavior for the
same misuse.

## Notes

- Raised during review of #1262:
  https://github.com/kaappi/kaappi/pull/1262#discussion_r3528963244
- Fix: add an `isEnvironment` guard in `evalFn` mirroring `loadFn`, e.g. when
  `args.len > 1` and `!types.isEnvironment(args[1])`, return
  `primitives.typeError("eval", "environment", args[1])`.
- A regression test belongs in the `(scheme eval)` audit / smoke tests
  alongside the `load` bad-env-type case.

baijum:
## Summary

- `eval` silently ignored a non-environment second argument, evaluating in the
  global environment instead of raising a type error
- The bug existed in **two code paths**: the `evalFn` native function
  (`primitives_r7rs.zig`) and the `tail_eval` bytecode opcode
  (`vm_dispatch.zig`) — the tail-position path was the harder one to catch
- Both now reject non-environment values, matching `load`'s behavior per R7RS §6.12

## Test plan

- [x] Regression tests added to `tests/scheme/compliance/eval.scm` (fixnum, string, boolean as bad env)
- [x] `zig build test` passes
- [x] Full Scheme suite passes (1784 pass, 0 fail)

Closes #1270

### DIFF
--- a/src/primitives_r7rs.zig
+++ b/src/primitives_r7rs.zig
@@ -175,6 +175,9 @@ fn evalFn(args: []const Value) PrimitiveError!Value {
     const expr = args[0];
 
     // If an environment specifier is provided, compile in that environment
+    if (args.len > 1) {
+        if (!types.isEnvironment(args[1])) return primitives.typeError("eval", "environment", args[1]);
+    }
     if (args.len > 1 and types.isEnvironment(args[1])) {
         const se = types.toEnvironment(args[1]);
         const func = compiler_mod.compileExpressionInEnv(gc, expr, &vm.macros, se.env, args[1]) catch return PrimitiveError.TypeError; // bare-ok: compile error
--- a/src/primitives_r7rs.zig
+++ b/src/primitives_r7rs.zig
@@ -177,8 +177,6 @@ fn evalFn(args: []const Value) PrimitiveError!Value {
     // If an environment specifier is provided, compile in that environment
     if (args.len > 1) {
         if (!types.isEnvironment(args[1])) return primitives.typeError("eval", "environment", args[1]);
-    }
-    if (args.len > 1 and types.isEnvironment(args[1])) {
         const se = types.toEnvironment(args[1]);
         const func = compiler_mod.compileExpressionInEnv(gc, expr, &vm.macros, se.env, args[1]) catch return PrimitiveError.TypeError; // bare-ok: compile error
         var closure_val = gc.allocClosure(func) catch return PrimitiveError.OutOfMemory;
--- a/src/vm_dispatch.zig
+++ b/src/vm_dispatch.zig
@@ -1181,7 +1181,11 @@ pub fn runUntil(self: *VM, target_frame_count: usize, target_wind_count: usize) 
 
                 const compiler_mod = @import("compiler.zig");
                 const func_val = blk: {
-                    if (nargs >= 2 and types.isEnvironment(self.registers[abs_base + 1])) {
+                    if (nargs >= 2) {
+                        if (!types.isEnvironment(self.registers[abs_base + 1])) {
+                            self.setErrorDetail("type error in 'eval': expected environment", .{});
+                            return VMError.TypeError;
+                        }
                         const se = types.toEnvironment(self.registers[abs_base + 1]);
                         break :blk compiler_mod.compileExpressionInEnv(self.gc, expr_val, &self.macros, se.env, self.registers[abs_base + 1]) catch return VMError.CompileError;
                     }
--- a/src/vm_dispatch.zig
+++ b/src/vm_dispatch.zig
@@ -1183,7 +1183,10 @@ pub fn runUntil(self: *VM, target_frame_count: usize, target_wind_count: usize) 
                 const func_val = blk: {
                     if (nargs >= 2) {
                         if (!types.isEnvironment(self.registers[abs_base + 1])) {
-                            self.setErrorDetail("type error in 'eval': expected environment", .{});
+                            const p = @import("printer.zig");
+                            const s = p.valueToString(self.gc.allocator, self.registers[abs_base + 1], .write) catch "";
+                            defer if (s.len > 0) self.gc.allocator.free(s);
+                            self.setErrorDetail("type error in 'eval': expected environment, got {s}", .{if (s.len > 0) s else "?"});
                             return VMError.TypeError;
                         }
                         const se = types.toEnvironment(self.registers[abs_base + 1]);
--- a/tests/scheme/compliance/eval.scm
+++ b/tests/scheme/compliance/eval.scm
@@ -25,7 +25,10 @@
   (test-assert "eval with boolean env raises error"
     (guard (e (#t #t))
       (eval '(+ 1 2) #t)
-      #f)))
+      #f))
+  (test-assert "eval with bad env in tail position raises error"
+    (guard (e (#t #t))
+      ((lambda () (eval '(+ 1 2) 42))))))
 
 ;; --- eval list construction ---
 (test-group "eval list"
--- a/tests/scheme/compliance/eval.scm
+++ b/tests/scheme/compliance/eval.scm
@@ -12,6 +12,21 @@
 (test-group "eval with environment"
   (test-eqv "eval with scheme base environment" 12 (eval '(* 3 4) (environment '(scheme base)))))
 
+;; --- eval rejects non-environment second argument (#1270) ---
+(test-group "eval bad env type"
+  (test-assert "eval with fixnum env raises error"
+    (guard (e (#t #t))
+      (eval '(+ 1 2) 42)
+      #f))
+  (test-assert "eval with string env raises error"
+    (guard (e (#t #t))
+      (eval '(+ 1 2) "not-an-env")
+      #f))
+  (test-assert "eval with boolean env raises error"
+    (guard (e (#t #t))
+      (eval '(+ 1 2) #t)
+      #f)))
+
 ;; --- eval list construction ---
 (test-group "eval list"
   (test-equal "eval quoted list" '(1 2 3) (eval '(list 1 2 3))))
--- a/tests/scheme/compliance/eval.scm
+++ b/tests/scheme/compliance/eval.scm
@@ -28,7 +28,8 @@
       #f))
   (test-assert "eval with bad env in tail position raises error"
     (guard (e (#t #t))
-      ((lambda () (eval '(+ 1 2) 42))))))
+      ((lambda () (eval '(+ 1 2) 42)))
+      #f)))
 
 ;; --- eval list construction ---
 (test-group "eval list"


tests/scheme/compliance/eval.scm
```diff
@@ -12,6 +12,21 @@
 (test-group "eval with environment"
   (test-eqv "eval with scheme base environment" 12 (eval '(* 3 4) (environment '(scheme base)))))
 
+;; --- eval rejects non-environment second argument (#1270) ---
+(test-group "eval bad env type"
```
baijum, line 16: These three tests all place `eval` in **non-tail** position (the trailing `#f` in the `guard` body follows it), so they only exercise the native `evalFn` path — not the `tail_eval` bytecode opcode that this PR also fixes. Reverting just the `vm_dispatch.zig` hunk leaves this file passing (6/6), so the tail-position guard is currently untested. Consider adding a case where `eval` is the tail expression, e.g. `((lambda () (eval '(+ 1 2) 42)))`, which compiles to `tail_eval` and fails without the `vm_dispatch.zig` fix.

src/primitives_r7rs.zig
```diff
@@ -175,6 +175,9 @@ fn evalFn(args: []const Value) PrimitiveError!Value {
     const expr = args[0];
 
     // If an environment specifier is provided, compile in that environment
+    if (args.len > 1) {
+        if (!types.isEnvironment(args[1])) return primitives.typeError("eval", "environment", args[1]);
+    }
     if (args.len > 1 and types.isEnvironment(args[1])) {
```
baijum: Minor: once the new guard above has run, `args.len > 1` implies `args[1]` is an environment, so the `and types.isEnvironment(args[1])` here is always true and the fall-through path below is now dead for the `args.len > 1` case. Not a bug, but you could either collapse this to `if (args.len > 1)` or fold the guard into this branch the way `loadFn` does (`if (args.len > 1) { if (!isEnvironment) return ...; ... }`) to avoid the double check.

src/vm_dispatch.zig
```diff
@@ -1181,7 +1181,11 @@ pub fn runUntil(self: *VM, target_frame_count: usize, target_wind_count: usize)
 
                 const compiler_mod = @import("compiler.zig");
                 const func_val = blk: {
-                    if (nargs >= 2 and types.isEnvironment(self.registers[abs_base + 1])) {
+                    if (nargs >= 2) {
+                        if (!types.isEnvironment(self.registers[abs_base + 1])) {
```
baijum, line 1185: Minor: this hardcoded message is less informative than the native path's, which reports the offending value. Native: `type error in 'eval': expected environment, got 42`; here: `type error in 'eval': expected environment` (no value). Consider including the value for parity — e.g. reuse the same formatting as `primitives.typeError` so both eval paths produce identical diagnostics.

baijum (commented):
## Review — fix looks correct; one test-coverage gap worth closing

I verified this against the **current `main`** (not the PR's base — `main` has moved to `08f0719`, including #1279 "Compile eval body in tail position", which touched these same `eval` code paths).

### What I checked ✅
- **Merges cleanly** despite the divergence. The 3-way merge preserves #1279's extra tail-position argument to `compileExpressionInEnv` (`false` in `evalFn`, `true` in `tail_eval`) and inserts the new guards correctly. `zig build` + `zig build test` both pass on the merged tree.
- **Both paths behave correctly at runtime.** `(eval '(+ 1 2) 42)` now raises a catchable type error in non-tail position (native `evalFn`) *and* tail position (`tail_eval`); valid `(environment ...)` specifiers still work in both positions; single-arg `eval` is unaffected.

### Main point: the regression tests only cover one of the two paths ⚠️
The PR's own summary calls the `tail_eval` bytecode path "the harder one to catch" — but the added tests don't exercise it. Every new test wraps the call as `(guard (e (#t #t)) (eval '(+ 1 2) 42) #f)`, and the trailing `#f` puts `eval` in **non-tail** position, so it always compiles to the native `evalFn` call, never the `tail_eval` opcode.

I confirmed this by reverting **only** the `vm_dispatch.zig` change and keeping the `primitives_r7rs.zig` fix: `tests/scheme/compliance/eval.scm` still reports **6 expected passes, 0 failures**. So a future regression of the tail-position guard would go undetected.

A tail-position test would close the gap (and fails without the `vm_dispatch.zig` fix, passes with it):
```scheme
(test-assert "eval with bad env in tail position raises error"
  (guard (e (#t #t))
    ((lambda () (eval '(+ 1 2) 42)))   ; eval is the tail expression -> tail_eval opcode
    #f))
```

### Verdict
The fix itself is correct and mergeable. I'd recommend adding the tail-position test above before merge, since guarding the harder path is the whole point of the change — but that's the only thing I'd hold on. The two inline notes below are minor/optional.

tests/scheme/compliance/eval.scm
```diff
@@ -12,6 +12,24 @@
...
+    (guard (e (#t #t))
+      (eval '(+ 1 2) "not-an-env")
+      #f))
+  (test-assert "eval with boolean env raises error"
+    (guard (e (#t #t))
+      (eval '(+ 1 2) #t)
+      #f))
+  (test-assert "eval with bad env in tail position raises error"
+    (guard (e (#t #t))
+      ((lambda () (eval '(+ 1 2) 42))))))
```
baijum: This assertion can never fail — it's missing the trailing `#f` that the three sibling tests use. When `eval` does not raise (bug present), the `guard` body returns the IIFE's value `3`, which is truthy, so `test-assert` passes; when it raises, `guard` returns `#t`, also truthy. I confirmed by reverting only the `vm_dispatch.zig` guard: this test still passes, while the same test with a trailing `#f` correctly fails. The construct does reach `tail_eval` (good), but the assertion provides no regression protection as written.

```scheme
  (test-assert "eval with bad env in tail position raises error"
    (guard (e (#t #t))
      ((lambda () (eval '(+ 1 2) 42)))
      #f)))
```

baijum (commented):
## Re-review of `50d3d60` — 2 of 3 addressed; the new test is inert

Thanks for the quick turnaround. I re-verified the merged result against **current `main`** (`6f383b8`; the branch is still based on a pre-#1279 main, but the 3-way merge cleanly keeps main's 6-arg `compileExpressionInEnv` calls — `false` for `evalFn`, `true` for `tail_eval`). `zig build` and `zig build test` both pass.

**✅ Redundant check** — collapsed into a single `if (args.len > 1)` branch. Good.

**✅ Error-message parity** — verified at runtime, both paths now produce identical text:
```
native msg: type error in 'eval': expected environment, got 42
tail   msg: type error in 'eval': expected environment, got 42
```
(The `defer if (s.len > 0) free(s)` is safe: `setErrorDetail` formats into its own buffer, and `valueToString` only returns an empty slice on the `catch ""` error path, so the length guard correctly distinguishes "allocated" from "fallback.")

**❌ Tail-position test — the construct is right but the assertion can never fail.**
The new test omits the trailing `#f` that the other three tests in the group use, so it's a tautology:
```scheme
(test-assert "eval with bad env in tail position raises error"
  (guard (e (#t #t))
    ((lambda () (eval '(+ 1 2) 42))))))   ; <- no trailing #f
```
When `eval` does **not** raise (i.e. the bug is present), the `guard` body returns the value of the IIFE — `3` — which is truthy, so `test-assert` passes anyway. When it does raise, `guard` returns `#t` — also truthy. Either way it passes.

I proved this by reverting **only** the `vm_dispatch.zig` guard and running the test both as-written and with a trailing `#f` added:
```
(bug present)  AS-WRITTEN (no #f):  PASS   <- tautology, misses the regression
(bug present)  CORRECTED (with #f): FAIL   <- correctly catches it (expected #t got #f)
```
So the construct genuinely reaches the `tail_eval` opcode (good — that's the hard path), but as written the test provides **no** protection for it: reverting the tail-position fix leaves `eval.scm` green.

**Fix:** add the trailing `#f` inside the `guard` body, matching the three sibling tests:
```scheme
(test-assert "eval with bad env in tail position raises error"
  (guard (e (#t #t))
    ((lambda () (eval '(+ 1 2) 42)))
    #f))
```

With that one-line change the fix is fully covered and I'd consider it **ready to merge**. Everything else checks out.

baijum (commented):
## ✅ Ready to merge

`c27d8e4` adds the trailing `#f` to the tail-position test, closing the last gap. I re-verified the full merge against current `main` (`6f383b8`):

- **Merges cleanly** — the 3-way merge preserves main's 6-arg `compileExpressionInEnv` calls (`false` in `evalFn`, `true` in `tail_eval` from #1279). No conflicts.
- **`zig build` + `zig build test`** both pass.
- **`eval.scm`** passes 7/7 with the fix.
- **The tail-position test now genuinely guards `tail_eval`.** Reverting only the `vm_dispatch.zig` guard reintroduces the bug and the suite correctly reports:
  ```
  FAIL eval with bad env in tail position raises error
  # of expected passes      6
  # of unexpected failures  1
  ```

All three review points are resolved: redundant check collapsed, error-message parity confirmed at runtime (`got 42` on both paths), and both the native and tail paths now have effective regression coverage. Fix is correct and complete for #1270.

Minor, non-blocking: the branch is still based on a pre-#1279 `main`, so the PR's own diff shows 5-arg `compileExpressionInEnv` calls. The merge resolves this correctly and builds, so a rebase isn't required — but rebasing onto current `main` before merge would make the diff self-consistent if you prefer. LGTM either way.

PR #1283, opened 2026-07-06
Unify platform feature lists across (features) and cond-expand (#1177)

### LINKED ISSUE #1177: [R7RS 6.14] (features) and the two cond-expand evaluators disagree (exact-closed, exact-complex)
**Spec reference:** R7RS §6.14, page 61 — `(features)`: "Returns a list of the feature identifiers which cond-expand treats as true."
**Severity:** wrong-result

**Minimal reproduction:**
```scheme
(features)                                        ;; => (r7rs kaappi ieee-float posix)
(cond-expand (exact-closed 'yes) (else 'no))      ;; => yes — but exact-closed is not in (features)
(cond-expand (exact-complex 'yes) (else 'no))     ;; => no  (expression-level)
(define-library (t ce)
  (import (scheme base)) (export r)
  (cond-expand (exact-complex (begin (define r 'yes)))
               (else (begin (define r 'no)))))
(import (t ce)) r                                 ;; => yes (library-level!)
```

**Expected:** one consistent feature set across all three.
**Actual:** three divergent hardcoded lists:

| Feature | `(features)` (src/primitives_list.zig `featuresFn`) | expression cond-expand (src/compiler_conditionals.zig `evalFeatureReq`) | library cond-expand (src/vm_library.zig `evalLibFeatureReq`) |
|---|---|---|---|
| r7rs, kaappi, ieee-float, posix | yes | yes | yes |
| exact-closed | **no** | yes | yes |
| exact-complex | **no** | **no** | yes |

Whether `exact-complex` *should* be claimed at all is a separate question (`(exact? (make-rectangular 1 2))` → #t today, but components are stored as f64) — whichever way that goes, the three lists must agree. Suggested fix: one shared feature-list constant consulted by all three call sites.

**Verified:** fresh ReleaseSafe build at 4995bb6
**Source location:** src/primitives_list.zig `featuresFn`; src/compiler_conditionals.zig `evalFeatureReq` (known_features); src/vm_library.zig:469 `evalLibFeatureReq` (known)
**Disabled test:** tests/scheme/audit/primitives_list-audit.scm (marked `;; FAIL: #this`)

**Found by:** Systematic audit, Phase 2.9

baijum:
## Summary

- Three call sites hardcoded divergent feature lists: `(features)` lacked `exact-closed` and `exact-complex`, expression-level `cond-expand` lacked `exact-complex`, while library-level `cond-expand` had both — violating R7RS §6.14 which requires `(features)` to return exactly the identifiers `cond-expand` treats as true.
- Introduces a single `types.platform_features` constant consulted by all three sites (`featuresFn`, `evalFeatureReq`, `evalLibFeatureReq`).
- Enables previously disabled audit assertions and adds a dedicated regression test.

## Test plan

- [x] New regression test `tests/scheme/smoke/features-consistency.scm` (8 assertions)
- [x] Audit test `tests/scheme/audit/primitives_list-audit.scm` — 117/117 pass (previously had FAIL markers)
- [x] R7RS suite §6.14 — 13/13 pass
- [x] Full `zig build test` passes

Closes #1177

### DIFF
--- a/src/compiler_conditionals.zig
+++ b/src/compiler_conditionals.zig
@@ -262,15 +262,7 @@ pub fn compileCondExpand(self: *Compiler, args: Value, dst: u16, is_tail: bool) 
 pub fn evalFeatureReq(self: *Compiler, req: Value) bool {
     if (types.isSymbol(req)) {
         const name = types.symbolName(req);
-        // Hardcoded feature identifiers
-        const known_features = [_][]const u8{
-            "r7rs",
-            "kaappi",
-            "ieee-float",
-            "posix",
-            "exact-closed",
-        };
-        for (known_features) |f| {
+        for (types.platform_features) |f| {
             if (std.mem.eql(u8, name, f)) return true;
         }
         return false;
--- a/src/primitives_list.zig
+++ b/src/primitives_list.zig
@@ -473,12 +473,10 @@ fn forEachFn(args: []const Value) PrimitiveError!Value {
 fn featuresFn(args: []const Value) PrimitiveError!Value {
     _ = args;
     const gc = getGC() orelse return PrimitiveError.OutOfMemory;
-    // Return a list of feature identifiers
-    const r7rs = gc.allocSymbol("r7rs") catch return PrimitiveError.OutOfMemory;
-    const kaappi = gc.allocSymbol("kaappi") catch return PrimitiveError.OutOfMemory;
-    const ieee_float = gc.allocSymbol("ieee-float") catch return PrimitiveError.OutOfMemory;
-    const posix_sym = gc.allocSymbol("posix") catch return PrimitiveError.OutOfMemory;
-    const items = [_]Value{ r7rs, kaappi, ieee_float, posix_sym };
+    var items: [types.platform_features.len]Value = undefined;
+    for (types.platform_features, 0..) |name, i| {
+        items[i] = gc.allocSymbol(name) catch return PrimitiveError.OutOfMemory;
+    }
     return gc.makeList(&items) catch return PrimitiveError.OutOfMemory;
 }
 
--- a/src/types.zig
+++ b/src/types.zig
@@ -1194,6 +1194,8 @@ test "truthiness" {
     try std.testing.expect(!isTruthy(FALSE));
 }
 
+pub const platform_features = [_][]const u8{ "r7rs", "kaappi", "ieee-float", "posix", "exact-closed", "exact-complex" };
+
 test "nil is not a pointer" {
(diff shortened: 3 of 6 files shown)


baijum:
## Review — ready to merge ✅

Clean, well-scoped fix. It correctly collapses the three divergent hardcoded lists into a single `types.platform_features` constant consulted by `featuresFn`, `evalFeatureReq`, and `evalLibFeatureReq`, which is exactly what R7RS §6.14 requires. CI is green across all targets, `mergeable: CLEAN`.

A few things I verified while reviewing:

- **GC safety of the `featuresFn` loop is fine.** The new loop holds already-allocated symbol `Value`s in the stack array `items` across subsequent `gc.allocSymbol` calls, which would normally be a rooting concern. But `allocSymbol` interns under `symbol_mutex` and never calls `maybeCollect` (see the comment at `src/memory.zig:246-249`), and interned symbols are roots via the symbol table anyway. So no `pushRoot` needed — no regression from the old unrolled version.
- **Choosing the superset (`exact-complex` included) is the right call, not just an arbitrary tie-break.** `types.Complex` tracks component exactness via `exact_real`/`exact_imag` flags (`src/types.zig:560-566`), so `(exact? (make-rectangular 1 2))` → `#t` reflects real behavior. This unifies on what the library-level evaluator already claimed *and* matches the actual representation. The issue flagged this as "a separate question" — worth a one-line confirmation in the PR that advertising it was deliberate given the f64 backing store.

### Non-blocking suggestions

1. **Test gap: library-level `cond-expand` isn't exercised.** The bug is "all *three* sites must agree," but both the new `features-consistency.scm` and the re-enabled audit assertions only cover `featuresFn` (`(features)`) and `evalFeatureReq` (expression-level `cond-expand`). The third site, `evalLibFeatureReq` in `src/vm_library.zig`, has no regression test. It was already the correct superset so regression risk is low, but a small `define-library` with a `(cond-expand (exact-complex ...) (else ...))` body in the smoke test would fully lock the fix in.
2. **Constant placement.** `platform_features` lands between `test "truthiness"` and `test "nil is not a pointer"` in `types.zig`. Cosmetic, but grouping it with other public declarations rather than mid-test-section would read better.
3. **Unused import.** `features-consistency.scm` imports `(scheme write)` but never uses `write`/`display`.
4. **Docs (follow-up, out of scope).** `(features)` now advertises `exact-closed` and `exact-complex` where it didn't before — if kaappi-lang.org lists feature identifiers anywhere, worth syncing separately.

None of these block merge. The test-ending pattern (grab the runner before `test-end`) is correct, and the fix cleanly closes #1177.

PR #1284, opened 2026-07-06
Use globally-unique binding IDs for syntax-rules literal identity (#1272)

### LINKED ISSUE #1272: [R7RS 4.3.2] syntax-rules literal check compares binding presence, not identity — shadowed same-named binding still matches (known limitation of #1139)
## Summary

The `syntax-rules` literal-binding check added for #1139 (PR #1265) compares binding identity via register slot numbers. This works within a single compiler frame but **not across frames**: slot indices are per-frame, so two different bindings in different lambda scopes that happen to occupy the same slot number are treated as identical.

This is a documented residual of the #1139 fix, not a regression: on `main` (pre-#1139) literals are matched by name only, so this case is also wrong there. Filed separately so the gap is tracked.

**Update (4a8501b):** The same-frame case originally described here is now fixed — PR #1265's third commit switched from boolean comparison to slot comparison, which correctly distinguishes `(let ((lit 1)) ... (let ((lit 2)) (m lit)))`. Only the cross-frame case below remains.

## Reproduction

```scheme
;; Cross-frame: outer lit at slot 0, inner lambda lit also at slot 0
((lambda (lit)
   (define-syntax m
     (syntax-rules (lit)
       ((_ lit) 'yes)
       ((_ x)   'no)))
   ((lambda (lit) (m lit)) 99)) 1)
;; Current: yes
;; R7RS 4.3.2: no (def-site lit and use-site lit are different bindings)
```

The pattern literal `lit` refers to the outer lambda's parameter (slot 0 in frame A). The use-site `lit` refers to the inner lambda's parameter (slot 0 in frame B). They are distinct bindings, but the slot numbers collide.

### Same-frame case (FIXED)

```scheme
(let ((lit 1))
  (define-syntax m
    (syntax-rules (lit)
      ((_ lit) 'is-literal)
      ((_ x)   'not-literal)))
  (let ((lit 2)) (m lit)))
;; Current: not-literal ✓ (different slots in the same frame)
```

## Root cause

`src/expander.zig` line 229–235: `def_slot` is captured from `resolveLocal` at `parseSyntaxRules` time and `use_slot` is resolved via the parent-chain callback at expansion time. Both are raw register indices (`u16`). Register indices are scoped to their owning `Compiler` frame, so equal indices in different frames denote different bindings but compare as equal.

## Direction for full conformance

Replace raw slot numbers with a globally-unique binding identity — e.g. `(compiler_id << 16) | slot` where each `Compiler` instance gets a monotonic serial, or a standalone binding-id counter incremented on every `addLocal`. The identity would be stored in `Transformer.literal_bound` and returned by the `UseSiteBindingCheck` callback.

## Impact

Low in practice — requires a lambda parameter (or let binding) that is also a macro literal, used inside a nested lambda whose own parameter shadows the same name and lands on the same register slot. Deferring is reasonable; this issue makes the gap explicit.

Related: #1139 (the bug the check fixes), PR #1265 (implementation), #1137 (conformance tracking).
Verified on `4a8501b`, macOS arm64 ReleaseSafe, `.sbc` cleared.

baijum:
## Summary

- Replace per-frame register slot numbers (`u16`) with monotonic binding IDs (`u32`) for `syntax-rules` literal identity comparison. Slot indices are scoped to their owning `Compiler` frame, so two different bindings in nested lambdas occupying the same register slot falsely compared as equal.
- Add `binding_id: u32` to the `Local` struct, assigned via `freshBindingId()` at every local creation site.
- Widen `Transformer.literal_bound`, `LITERAL_UNBOUND`/`LITERAL_BOUND_PENDING` sentinels, and `UseSiteBindingCheck.resolve` from `u16` to `u32`.

Fixes #1272.

## Test plan

- [x] Reproduction case now returns `no` (was `yes`): `((lambda (lit) (define-syntax m (syntax-rules (lit) ((_ lit) 'yes) ((_ x) 'no))) ((lambda (lit) (m lit)) 99)) 1)`
- [x] Same-frame case still correct (`not-literal`)
- [x] Same-binding positive case still matches (`is-literal`)
- [x] Unbound-matches-unbound still works
- [x] Two new regression tests in `tests/scheme/hygiene/literal-binding.scm`
- [x] All 1784 tests pass (1395 R7RS + 389 Scheme files)
- [x] `zig build test` passes

### DIFF
--- a/src/compiler.zig
+++ b/src/compiler.zig
@@ -31,10 +31,19 @@ pub const CompileError = error{
     JumpOutOfRange,
 };
 
+var next_binding_id: u32 = 0;
+
+pub fn freshBindingId() u32 {
+    const id = next_binding_id;
+    next_binding_id += 1;
+    return id;
+}
+
 const Local = struct {
     name: []const u8,
     depth: u16,
     slot: u16,
+    binding_id: u32,
     is_boxed: bool = false,
     // Register alias for a global, injected during macro expansion so a
     // template's free reference pierces use-site shadowing. set! through
@@ -263,6 +272,17 @@ pub const Compiler = struct {
         return null;
     }
 
+    pub fn resolveBindingId(self: *Compiler, name: []const u8) ?u32 {
+        var i: usize = self.locals.items.len;
+        while (i > 0) {
+            i -= 1;
+            if (std.mem.eql(u8, self.locals.items[i].name, name)) {
+                return self.locals.items[i].binding_id;
+            }
+        }
+        return null;
+    }
+
     /// Pure predicate: is `name` bound as a lexical variable — a local in this
     /// compiler or any enclosing compiler? Unlike `resolveUpvalue`, this has no
     /// side effects (it does not register upvalues), so the IR optimizer can
@@ -381,6 +401,7 @@ pub const Compiler = struct {
             .name = name,
             .depth = self.scope_depth,
             .slot = slot,
+            .binding_id = freshBindingId(),
         }) catch return CompileError.OutOfMemory;
     }
 
--- a/src/compiler_ir.zig
+++ b/src/compiler_ir.zig
@@ -251,6 +251,7 @@ pub fn compileLambdaWithIR(self: *Compiler, args: Value, dst: u16, name: ?[]cons
             .name = types.symbolName(formals),
             .depth = 1,
             .slot = slot,
+            .binding_id = compiler_mod.freshBindingId(),
         }) catch return CompileError.OutOfMemory;
     } else {
         while (param_list != types.NIL) {
@@ -261,6 +262,7 @@ pub fn compileLambdaWithIR(self: *Compiler, args: Value, dst: u16, name: ?[]cons
                     .name = types.symbolName(param_list),
                     .depth = 1,
                     .slot = slot,
+                    .binding_id = compiler_mod.freshBindingId(),
                 }) catch return CompileError.OutOfMemory;
                 break;
             }
@@ -272,6 +274,7 @@ pub fn compileLambdaWithIR(self: *Compiler, args: Value, dst: u16, name: ?[]cons
                 .name = types.symbolName(param),
                 .depth = 1,
                 .slot = slot,
+                .binding_id = compiler_mod.freshBindingId(),
             }) catch return CompileError.OutOfMemory;
             arity += 1;
             param_list = types.cdr(param_list);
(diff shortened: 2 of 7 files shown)


baijum (commented):
## Review — ready to merge ✅

Clean, well-scoped fix for #1272. This is the exact "standalone binding-id counter incremented on every `addLocal`" direction proposed in the issue, and it does the right thing: literal identity is now compared by a process-unique `binding_id` instead of per-frame register slots, so two same-named bindings in different frames that land on the same slot no longer collide.

**Verified locally** (macOS arm64, ReleaseSafe, `.sbc` cleared, post-build):
- Reproduction `((lambda (lit) (define-syntax m (syntax-rules (lit) ((_ lit) 'yes) ((_ x) 'no))) ((lambda (lit) (m lit)) 99)) 1)` → `no` ✔ (was `yes`)
- Positive same-frame case → `yes` ✔
- `tests/scheme/hygiene/literal-binding.scm` → 15/15 pass; `zig build test` passes; CI green across all platforms + Debug/ReleaseSafe/ReleaseFast + wasm.

**What makes this solid:** the new `Local.binding_id` field has *no default*, so the Zig compiler forces every one of the 13 `Local` initialization sites to supply an id — a missed site is a build error, not a silent slot-0 bug. The `u16 → u32` widening of `literal_bound`, the sentinels, and the `UseSiteBindingCheck.resolve` signature is consistent throughout. The def-site (`resolveBindingId`, self-frame) vs use-site (`resolveLocalSkipAliases`, parent-walking) lookup asymmetry is preserved exactly from the old slot-based code, so this is a faithful transform.

### Non-blocking notes

**1. `freshBindingId()` is a non-atomic read-modify-write on a process-global** (`src/compiler.zig:34-39`). Under SRFI-18, child threads each get their own VM/GC but `next_binding_id` is a single module-level global shared across all of them. If two threads compile concurrently (e.g. runtime `eval`/`load`/macro expansion on a spawned thread), a torn `next_binding_id += 1` could hand out duplicate ids — which would silently reintroduce the exact false-match this PR fixes. There's precedent for guarding shared compiler state here: symbol interning uses `symbol_mutex`. Cheap to make robust:
```zig
pub fn freshBindingId() u32 {
    return @atomicRmw(u32, &next_binding_id, .Add, 1, .monotonic);
}
```
Probability is low (needs concurrent compilation of a literal-shadowing macro), so this isn't a merge blocker — but it's a one-line hardening worth considering, or at least a comment noting the counter assumes single-threaded compilation.

**2. Counter never resets; sentinels sit at the top of the u32 range** (`LITERAL_UNBOUND = 0xFFFFFFFF`, `LITERAL_BOUND_PENDING = 0xFFFFFFFE`). A real `binding_id` reaching `0xFFFFFFFE` would alias a sentinel before the increment overflow-panics. This requires ~4.3 billion local bindings in one process, so it's purely theoretical — noting it only because it's a shared global with sentinels adjacent to the live range.

Neither point blocks merge. LGTM.

PR #1285, opened 2026-07-06
Enforce immutability on literal vectors, pairs, and bytevectors (#1173)

### LINKED ISSUE #1173: [R7RS 6.8] vector-set! on a literal vector silently mutates the shared constant (strings already enforce immutability)
**Spec reference:** R7RS §6.8, page 48 — `(vector-set! '#(0 1 2) 1 "doe") ⟹ error ; constant vector`. (§3.4: "It is an error to attempt to store a new value into a location that is denoted by an immutable object.")
**Severity:** edge-case (permitted by "it is an error", but inconsistent with strings and a self-modifying-constant hazard)

**Minimal reproduction:**
```scheme
(define (f) '#(0 1 2))
(vector-set! (f) 1 'doe)
(f)   ;; expected: #(0 1 2) (or an error on the vector-set!)
      ;; actual:   #(0 doe 2) — the literal itself was mutated, persisting across calls
```

**Expected:** an error ("constant vector"), as in the R7RS example — kaappi already does exactly this for strings: `(string-set! "abc" 0 #\X)` raises `mutable string` expected.
**Actual:** silent mutation of the shared literal. Every subsequent evaluation of the literal sees the mutated value (self-modifying code).

**Inconsistency:** `SchemeString` has an `immutable: bool` flag (src/types.zig:279) enforced in `string-set!`, `string-copy!`, `string-fill!`. `Vector` has no equivalent flag, so `vector-set!`, `vector-fill!`, `vector-copy!`, `vector-reverse!`, `vector-swap!` all mutate literals. A fix would mirror the string mechanism: mark vectors created from quoted literals immutable and check the flag in the five mutators.

**Verified:** fresh ReleaseSafe build at 96ce73b; gc-stress unchanged
**Source location:** src/primitives_vector.zig (`vectorSetFn` etc. — no immutability check); literal construction in compiler/bytecode constant handling

baijum:
## Summary

- Add `immutable: bool = false` to `Vector`, `Pair`, and `Bytevector` structs in `types.zig`, mirroring the existing `SchemeString.immutable` pattern
- Mark all reader-produced literals (vectors, pairs, bytevectors) as immutable in `reader_datum.zig` — covers `readVector`, `readList`, `readListTail`, `readAbbreviation`, `readBytevector`, and datum labels
- Guard 13 mutating primitives: `vector-set!`, `vector-fill!`, `vector-copy!`, `vector-swap!`, `vector-reverse!`, `set-car!`, `set-cdr!`, `list-set!`, `bytevector-u8-set!`, `bytevector-copy!`, `read-bytevector!`
- Runtime-constructed objects (`make-vector`, `list`, `bytevector`, `list-copy`, `vector-copy`, etc.) remain fully mutable

Closes #1173

## Test plan

- [x] Enabled 3 previously-disabled `FAIL: #1173` audit tests (vector, list, bytevector)
- [x] Added `tests/scheme/smoke/literal-immutability.scm` — 24 tests covering all mutators, runtime mutability, and cross-call persistence
- [x] `bash tests/scheme/run-all.sh` — 1785 pass, 0 fail (390 Scheme files + 1395 R7RS)
- [x] GC stress test passes (`-Dgc-stress=true`)

### DIFF
--- a/src/reader_datum.zig
+++ b/src/reader_datum.zig
@@ -37,7 +37,7 @@ fn tokenToValue(self: *Reader, tok: Token) ReadError!Value {
         .character => |c| return types.makeChar(c),
         .string => |s| {
             const val = self.gc.allocString(s) catch return ReadError.OutOfMemory;
-            types.toObject(val).as(types.SchemeString).immutable = true;
+            if (self.mark_immutable) types.toObject(val).flags.immutable = true;
             return val;
         },
         .symbol => |name| return self.gc.allocSymbol(name) catch return ReadError.OutOfMemory,
@@ -66,7 +66,7 @@ fn tokenToValue(self: *Reader, tok: Token) ReadError!Value {
                 self.gc.writeBarrier(types.toObject(placeholder), types.car(datum));
                 types.setCdr(placeholder, types.cdr(datum));
                 self.gc.writeBarrier(types.toObject(placeholder), types.cdr(datum));
-                types.toObject(placeholder).as(types.Pair).immutable = true;
+                if (self.mark_immutable) types.toObject(placeholder).flags.immutable = true;
                 return placeholder;
             } else {
                 self.labels[n] = datum;
@@ -113,7 +113,7 @@ fn readList(self: *Reader) ReadError!Value {
             }
             self.pos += 1;
             const pair = self.gc.allocPair(first_root, rest_root) catch return ReadError.OutOfMemory;
-            types.toObject(pair).as(types.Pair).immutable = true;
+            if (self.mark_immutable) types.toObject(pair).flags.immutable = true;
             self.gc.source_lines.put(pair, list_line) catch {};
             return pair;
         }
@@ -124,7 +124,7 @@ fn readList(self: *Reader) ReadError!Value {
     self.gc.pushRoot(&rest_root);
     defer self.gc.popRoot();
     const pair = self.gc.allocPair(first_root, rest_root) catch return ReadError.OutOfMemory;
-    types.toObject(pair).as(types.Pair).immutable = true;
+    if (self.mark_immutable) types.toObject(pair).flags.immutable = true;
     self.gc.source_lines.put(pair, list_line) catch {};
     return pair;
 }
@@ -169,7 +169,7 @@ fn readListTail(self: *Reader) ReadError!Value {
         const elem_line = self.getLineCol().line;
         elem = try readDatum(self);
         const pair = self.gc.allocPair(elem, types.NIL) catch return ReadError.OutOfMemory;
-        types.toObject(pair).as(types.Pair).immutable = true;
+        if (self.mark_immutable) types.toObject(pair).flags.immutable = true;
         self.gc.source_lines.put(pair, elem_line) catch {};
 
         if (result == types.NIL) {
@@ -193,12 +193,12 @@ fn readAbbreviation(self: *Reader, keyword: []const u8) ReadError!Value {
     defer self.gc.popRoot();
 
     const rest = self.gc.allocPair(datum_root, types.NIL) catch return ReadError.OutOfMemory;
-    types.toObject(rest).as(types.Pair).immutable = true;
+    if (self.mark_immutable) types.toObject(rest).flags.immutable = true;
     var rest_root = rest;
     self.gc.pushRoot(&rest_root);
     defer self.gc.popRoot();
     const result = self.gc.allocPair(sym_root, rest_root) catch return ReadError.OutOfMemory;
-    types.toObject(result).as(types.Pair).immutable = true;
+    if (self.mark_immutable) types.toObject(result).flags.immutable = true;
     return result;
 }
 
@@ -226,7 +226,7 @@ fn readVector(self: *Reader) ReadError!Value {
     }
 
     const vec = self.gc.allocVector(elems.items) catch return ReadError.OutOfMemory;
-    types.toObject(vec).as(types.Vector).immutable = true;
+    if (self.mark_immutable) types.toObject(vec).flags.immutable = true;
     return vec;
 }
 
@@ -300,6 +300,6 @@ fn readBytevector(self: *Reader) ReadError!Value {
     }
 
     const bv = self.gc.allocBytevector(bytes.items) catch return ReadError.OutOfMemory;
-    types.toObject(bv).as(types.Bytevector).immutable = true;
+    if (self.mark_immutable) types.toObject(bv).flags.immutable = true;
     return bv;
 }
--- a/src/reader_datum.zig
+++ b/src/reader_datum.zig
@@ -66,6 +66,7 @@ fn tokenToValue(self: *Reader, tok: Token) ReadError!Value {
                 self.gc.writeBarrier(types.toObject(placeholder), types.car(datum));
                 types.setCdr(placeholder, types.cdr(datum));
                 self.gc.writeBarrier(types.toObject(placeholder), types.cdr(datum));
+                types.toObject(placeholder).as(types.Pair).immutable = true;
                 return placeholder;
             } else {
                 self.labels[n] = datum;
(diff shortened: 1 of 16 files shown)


src/types.zig
```diff
@@ -265,6 +265,7 @@ pub const Pair = struct {
     header: Object,
     car: Value = NIL,
     cdr: Value = NIL,
```
baijum, line 273: This pushes `@sizeOf(Pair)` from 32 → 40 bytes: the trailing `bool` needs only 1 byte but rounds up to the struct's 8-byte alignment. Same for `Vector` and `Bytevector` (32 → 40). Since pairs are the most-allocated heap object, that's a ~25% heap-footprint hit plus extra GC scan cost on cons-heavy workloads.

Compare the `owner: u32` field on `Object` (line ~203), which was deliberately fit into existing header padding "so it adds no size." The same trick applies here: `marked` (bool), `generation` (u1), and `survive_count` (u2) currently occupy 3 separate bytes but use only 4 bits. Folding them + `immutable` into a single `packed struct(u8)` flags field on `Object` keeps `Pair` at 32 bytes and makes the flag free for *all* heap types at once. More invasive (touches the GC's flag reads), but zero memory cost — worth weighing against the current simpler approach.

src/reader_datum.zig
```diff
@@ -218,7 +225,9 @@ fn readVector(self: *Reader) ReadError!Value {
         self.gc.extra_roots.append(self.gc.allocator, elem) catch return ReadError.OutOfMemory;
     }
 
-    return self.gc.allocVector(elems.items) catch return ReadError.OutOfMemory;
+    const vec = self.gc.allocVector(elems.items) catch return ReadError.OutOfMemory;
```
baijum, line 228: This reader path is shared with the runtime `read` procedure (`primitives_io.zig` calls `reader.readDatum`), so marking here also makes `(read …)` return immutable vectors/pairs/bytevectors — a change from `main`, where read data was mutable. R7RS §3.4 immutability targets literal constants in *program text*, not freshly `read` objects, so programs that read an s-expression and mutate it in place will now error. Consider threading a `mutable`/`for-literal` flag through the reader so only the compiler's literal-reading path sets `immutable`, leaving the `read` primitive's output mutable.

baijum (commented):
## Review — enforce literal immutability (#1173)

Solid, well-scoped fix. I checked out `fix/1173`, built it, and verified behavior. The enforcement is **complete and correct**, and the tests are thorough. Two design-level items are worth resolving/deciding before merge, plus a couple of minor notes.

### What's good
- **Complete mutator coverage.** I enumerated every registered `!` procedure touching these types — `set-car!`, `set-cdr!`, `list-set!`, `vector-set!`, `vector-fill!`, `vector-copy!`, `vector-swap!`, `vector-reverse!`, `bytevector-u8-set!`, `bytevector-copy!`, `read-bytevector!`. All are guarded; nothing was missed (no `vector-sort!`/`bytevector-fill!` exist to slip through).
- **All 8 reader allocation sites are marked** (readList ×2, readListTail, readAbbreviation ×2, readVector, readBytevector, datum-label placeholder). Nested literals verified immutable: `'(1 (2 3) 4)`, `'(a #(1 2) b)`, `'#0=(1 2 3)` all correctly reject mutation.
- **Runtime-constructed objects stay mutable** — verified `vector-copy`, `list-copy`, `string->list`, `(vector …)`, quasiquote-with-unquote all still mutate fine.
- Guards correctly target the destination arg (`args[0]`) for the copy!/read! forms.
- `zig build test` passes; the 3 re-enabled audit tests (176/115/119) and the new 24-case smoke test pass.

### Concern 1 — `read` now returns immutable pairs & vectors (behavior change)
The reader is shared between compile-time literal reading and the runtime `read` procedure (`primitives_io.zig` → `reader.readDatum`). Because every reader-built pair/vector/bytevector is now marked immutable, **`read` at runtime returns immutable data**. Verified on this branch:
```scheme
(set-car! (read (open-input-string "(1 2 3)")) 99)   ; => error (was fine on main)
(vector-set! (read (open-input-string "#(1 2 3)")) 0 9) ; => error (was fine on main)
```
R7RS §3.4 immutability is about *literal constants in program text*, not objects freshly constructed by `read`, so mutating read data is a legitimate pattern that this silently breaks. This was already true for strings pre-PR, but extending it to pairs/vectors/bytevectors makes it far more impactful. Recommend a deliberate decision here — ideally thread a flag through the reader so only the compiler's literal path marks immutable while the `read` primitive yields mutable data. If you keep the current behavior, please call it out in the PR description + CHANGELOG so it's an intentional, documented choice.

### Concern 2 — `Pair` grows 32 → 40 bytes (+25%)
See inline note on `types.zig`. `@sizeOf(Pair)` goes 32→40 on this branch (Vector/Bytevector 32→40 too) because the trailing `bool` rounds up to 8-byte alignment. Pairs are the most heavily allocated heap object, so this is a real heap-footprint and GC-scan cost for cons-heavy workloads. Worth considering the zero-cost header-packing alternative.

### Minor
- No `[Unreleased]` CHANGELOG entry — the project follows Keep a Changelog.
- The smoke test's three top-level `guard` forms echo `#f` to stdout (the file runner prints non-void top-level values). Harmless, but wrapping them (e.g. in a `begin`/`when`) would keep the output clean.

### Verdict
Not quite ready to merge — the core change is correct and I'd happily approve once **Concern 1** gets an explicit decision (it's a user-visible semantic change) and **Concern 2** is either addressed or consciously accepted. Neither is a correctness bug; both are design calls.

baijum (commented):
## Follow-up review — both concerns resolved ✅

`736992e` addresses both items from my earlier review cleanly. I checked it out, built it, and verified behavior + tests.

### Concern 1 — `read` returns immutable data → **fixed**
`Reader.mark_immutable` (default `true`) is set to `false` at the 4 `read`-procedure sites in `primitives_io.zig`, and `reader_datum.zig` gates every `.immutable` mark on it. Verified:
- `read` now returns **mutable** pairs, vectors, **and strings** — `(set-car! (read (open-input-string "(1 2 3)")) 99)` works, as do the vector/string equivalents.
- Program-text literals remain immutable (`(set-car! '(1 2 3) 99)` still errors).

This is the correct R7RS split (§3.4 literals immutable vs §6.13.2 read data mutable), and it's actually an improvement over `main`, where `read` returned immutable *strings*.

### Concern 2 — `Pair` grew 32→40 bytes → **fixed**
`marked`/`generation`/`survive_count`/`immutable` are now folded into a single `Object.Flags` `packed struct(u8)`. Verified `@sizeOf`:

| Type | Before this PR | Original PR (14de806) | Now (736992e) |
|------|---:|---:|---:|
| `Object` | 16 | 16 | **16** |
| `Pair` | 32 | 40 | **32** |
| `Vector` | 32 | 40 | **32** |
| `Bytevector` | 32 | 40 | **32** |
| `SchemeString` | 48 | 48 | **40** |

Zero memory cost for the flag, and `SchemeString` even shrinks 8 bytes since its dedicated `immutable` field is gone. The GC hot-path field renames in `gc_collect.zig`/`memory.zig` are mechanical and the `#958` SRFI-18 tests were updated to match.

### Verification I ran
- `zig build` + `zig build test` — pass
- Smoke test (now SRFI-64, 27 expected passes incl. 3 new read-mutability cases) — pass
- Re-enabled audit tests — 176 / 115 / 119, 0 fail
- `-Dgc-stress=true`: smoke test + a 5000-pair build/promote churn — immutable literals survive collections and stay immutable, mutable pairs mutate fine

### Minor from before
- Smoke test's stray `#f` output is gone (SRFI-64 rewrite + `(begin … (values))` wrapping).
- The earlier benchmark run already showed no throughput regression, and this version removes even the memory-footprint cost.

CHANGELOG `[Unreleased]` still has no entry for this — worth a one-liner, but non-blocking.

**This is ready to merge.** (Not using the Approve button since I'm the PR author; consider this an explicit 👍.)

PR #1286, opened 2026-07-07
Move parameterize converter application outside dynamic-wind extent

### LINKED ISSUE #1266: parameterize converter mutations leak outside dynamic-wind extent
**Category:** Compiler (parameterize form)
**Spec reference:** SRFI-39 reference implementation converts all values and saves all old values *before* installing any binding. R7RS §4.2.6 says the body runs in the extended dynamic environment, implying installation is atomic w.r.t. the body — but the current desugaring mutates parameters in the inner `let*` before `dynamic-wind` is entered.
**Severity:** wrong-result (edge cases)
**Pre-existing:** yes — identical behavior on `main` before #1260

**Background:** #1202 fixed value-expression evaluation order (the outer `let`). This issue tracks the remaining non-atomicity: the inner `let*` still calls each parameter with its value (mutating the cell) before `dynamic-wind` installs the restore thunk.

**Three consequences:**

1. **A later converter observes an earlier binding's new value:**
   ```scheme
   (define a (make-parameter 1))
   (define b (make-parameter 0 (lambda (v) (a))))
   (parameterize ((a 2) (b 99)) (b))  ;; => 2, SRFI-39 ref impl => 1
   ```

2. **A converter error mid-install permanently leaks an earlier mutation** — the `dynamic-wind` isn't entered yet, so the after-thunk never restores:
   ```scheme
   (define p (make-parameter 1))
   (define w (make-parameter #t (lambda (x) (if (boolean? x) x (error "bad")))))
   (guard (e (#t (p))) (parameterize ((p 2) (w 'oops)) 'x))  ;; => 2, p stays 2 forever
   ```

3. **Duplicate params restore the wrong value:** `(parameterize ((q 1) (q 2)) (q))` leaves `q` at 1 instead of its original value, since `old1` is read after `new0` was installed.

**Root cause:** `(begin (%pp_i %pv_i) (%pp_i))` in the inner `let*` both converts and installs in one step. There's no way to run the converter without mutating the cell.

**Suggested fix:** Add a `%parameter-convert` primitive that applies the converter without mutating the cell, returning the converted value. Then the inner bindings become pure reads:
```scheme
(let* ((old0 (%pp0))
       (new0 (%parameter-convert %pp0 %pv0))   ;; convert without mutation
       (old1 (%pp1))
       (new1 (%parameter-convert %pp1 %pv1)))
  (dynamic-wind
    (lambda () (%parameter-set! %pp0 new0) (%parameter-set! %pp1 new1) ...)
    (lambda () body ...)
    (lambda () (%parameter-set! %pp0 old0) (%parameter-set! %pp1 old1) ...)))
```

All mutation happens inside `dynamic-wind`, so converter errors and continuation re-entries are safe.

**Source location:** `src/compiler_advanced.zig` (`compileParameterize`), `src/primitives_r7rs.zig` (parameter primitives)

**Found by:** Code review on #1260 (both reviewer and CodeRabbit)

baijum:
## Summary

Fixes #1266.

- Add `%parameter-convert` primitive that applies a parameter's converter without mutating the cell
- Update `compileParameterize` desugaring so the inner `let*` uses `(%parameter-convert %pp %pv)` instead of the mutating `(begin (%pp %pv) (%pp))` idiom
- All parameter cell mutation now happens inside `dynamic-wind`'s before/after thunks, making converter errors and continuation re-entries safe

This fixes three consequences of the old desugaring:
1. A later converter no longer observes an earlier binding's new value
2. A converter error mid-install no longer permanently leaks an earlier mutation
3. Duplicate params now correctly restore the original value

## Test plan

- [x] New regression tests in `tests/scheme/smoke/parameterize-convert-1266.scm` (6 tests covering all three bug scenarios + converter sanity)
- [x] Full Scheme suite: 1788 pass, 0 fail
- [x] Zig unit tests pass

### DIFF
--- a/src/compiler_advanced.zig
+++ b/src/compiler_advanced.zig
@@ -677,8 +677,8 @@ fn buildQQListExpr(gc: *@import("memory.zig").GC, quote_sym: Value, list_sym: Va
 ///
 /// Desugars to:
 ///   (let ((%pp0 p1) (%pp1 p2) ... (%pv0 v1) (%pv1 v2) ...)
-///     (let* ((old0 (%pp0)) (new0 (begin (%pp0 %pv0) (%pp0)))
-///            (old1 (%pp1)) (new1 (begin (%pp1 %pv1) (%pp1))) ...)
+///     (let* ((old0 (%pp0)) (new0 (%parameter-convert %pp0 %pv0))
+///            (old1 (%pp1)) (new1 (%parameter-convert %pp1 %pv1)) ...)
 ///       (dynamic-wind
 ///         (lambda () (%parameter-set! %pp0 new0) (%parameter-set! %pp1 new1) ...)
 ///         (lambda () body ...)
@@ -686,9 +686,9 @@ fn buildQQListExpr(gc: *@import("memory.zig").GC, quote_sym: Value, list_sym: Va
 ///
 /// The outer `let` evaluates all param and value expressions before any
 /// parameter cell is mutated (R7RS §4.2.6, SRFI-39). The inner `let*`
-/// sequentially saves old values and installs new ones through each
-/// parameter's converter. The before-thunk uses %parameter-set! to avoid
-/// re-applying the converter on continuation re-entry.
+/// saves old values and runs converters without mutating cells.
+/// All mutation happens inside `dynamic-wind`, so converter errors and
+/// continuation re-entries are safe.
 pub fn compileParameterize(self: *Compiler, args: Value, dst: u16, is_tail: bool) CompileError!void {
     if (args == types.NIL) return CompileError.InvalidSyntax;
     const bindings = types.car(args);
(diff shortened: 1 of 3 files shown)


baijum (commented):
## Review — ready to merge ✅

Reviewed the desugaring change, the new primitive, and the regression tests. This is a correct, minimal, well-scoped fix for #1266.

### What I verified locally
- **Builds clean** on both branches (`zig build`, ReleaseSafe).
- **Regression test is genuine**: `tests/scheme/smoke/parameterize-convert-1266.scm` fails on `main` with exactly the three documented bugs and passes on this branch.
  - `main`: `FAIL: converter sees old value (2)`, `FAIL: converter error does not leak (2)`, `FAIL: duplicate param restores original (1)` → **3 pass, 3 fail**
  - PR: **6 pass, 0 fail**
- Full `tests/scheme/smoke/*.scm` sweep: no regressions.

### Correctness
The core insight is right: the old `(begin (%pp %pv) (%pp))` idiom in the inner `let*` both **converts and mutates** in one step, before `dynamic-wind` is entered — so any mutation that happens there escapes the wind extent. Replacing it with `(%parameter-convert %pp %pv)` (convert, no mutation) means the only cell mutation now happens in the before/after thunks, which is exactly what makes converter-errors and continuation re-entry safe. The `let*` ordering (`old0, new0, old1, new1, …`) and the use of `%parameter-set!` (not the converting call) in the before-thunk are both preserved, so the "converter runs once" SRFI-39 semantic on re-entry still holds.

The `#1202` outer-`let` evaluation-order fix and its unit test are untouched and complementary — this handles the *installation* atomicity that #1202 didn't.

### `%parameter-convert` primitive
- GC-safe: `args` is a slice into the VM register file (`callNative`), so `args[1]` stays rooted across the converter call even if it triggers GC; the returned value needs no rooting since nothing allocates after. This mirrors the existing `makeParameterFn`/`callWithArgs` patterns.
- Registration (`.libs = LS.initOne(.scheme_base)`, no `.sandbox = false`) matches `%parameter-set!` exactly, so sandbox behavior and library visibility are consistent. No other list needs updating.
- Docstring in `compileParameterize` is fully updated to match the new desugaring.

### Non-blocking note
- For pathological input like `(parameterize ((list 10)) …)` the error message surfaces the internal `%parameter-convert` name — but this is pre-existing (`main` already surfaces `%parameter-set!` for the same input) and not a regression, so no action needed.

No blocking issues found. **This is ready for merge.**

PR #1287, opened 2026-07-07
Capture let/lambda locals in define-syntax transformers (#1271)

### LINKED ISSUE #1271: Hygiene: define-syntax template can't capture a let/lambda binding from its definition scope (emits __hyg_N_var undefined)
## Summary

A `define-syntax` macro whose `syntax-rules` template contains a free reference to a **`let` / `let*` / `lambda`-bound** variable from the macro's *definition* scope fails to expand: the reference comes out as a hygienically-renamed identifier (`__hyg_N_<var>`) that is never mapped back to the binding, producing an "undefined variable" error.

This is **not** caused by PR #1265 — it reproduces identically on `main` (`1806384`) and on the PR branch (`2b14182`). It's a pre-existing hygiene gap, surfaced while reviewing #1139.

## Reproduction

```scheme
(let ((y 7))
  (define-syntax gety (syntax-rules () ((_) y)))
  (display (gety)) (newline))
```

Expected: `7` (per R7RS hygiene, the template's free `y` refers to the binding in scope at the macro's definition site — the `let`'s `y`).

Actual:
```
error: undefined variable '__hyg_1_y'
```

## What works vs what fails

Same `(display (gety))` shape, varying only how `y` is bound at the definition site:

| Definition-site binding of `y` | Result |
|---|---|
| `(let ((y 7)) (define-syntax gety …) (gety))` | ❌ `undefined variable '__hyg_1_y'` |
| `(let* ((y 7)) …)` | ❌ `__hyg_1_y` |
| `((lambda (y) (define-syntax gety …) (gety)) 7)` | ❌ `__hyg_1_y` |
| `(let ((y 7)) (define z 1) (define-syntax gety …) (gety))` | ❌ `__hyg_1_y` |
| **`let-syntax`** instead of `define-syntax`: `(let ((y 7)) (let-syntax ((gety …)) (gety)))` | ✅ `7` |
| `define-syntax` template references a **global**: `(define g 99) (let ((y 7)) (define-syntax getg (syntax-rules () ((_) g))) (getg))` | ✅ `99` |
| `define-syntax` in a **procedure body** binding via internal `define`: `(define (h) (define y 7) (define-syntax gety …) (gety))` | ✅ `7` |

So the gap is narrow and specific: **`define-syntax` capturing a `let`/`let*`/`lambda` binding from its definition scope**. `let-syntax` handles it (via the `captured_locals` machinery), globals are handled (via `def_env`), and internal-`define` bindings are handled — but a `define-syntax` whose def-site lexical is a `let`/lambda slot is not recorded, so `renameForHygiene` renames the free template reference and nothing aliases it back.

The nested-lambda variant (`(display ((lambda () (gety))))`) fails the same way; the nesting is incidental — the trigger is the `let`/lambda-bound def-site variable, not the use-site lambda.

## Notes for a fix

The `let-syntax` path already records def-site lexicals in `Transformer.captured_locals` and re-injects them at the use site (`injectHygienicCapturedLocals` in `src/compiler_macro.zig`). The `define-syntax` path (`compileDefineSyntax`, and the body-scan path in `compileBodyForms`) records `def_env` (globals) but not the enclosing `let`/lambda lexicals, so the same re-injection never happens. A fix likely extends the `define-syntax` registration to capture def-site lexicals the way `let-syntax` does — while staying clear of the literal-matching regressions found in #1139 / PR #1265 (that machinery must not be re-activated in a way that mis-resolves ordinary macro arguments).

A regression test belongs in `tests/scheme/hygiene/`.

Verified on `main` `1806384` and `fix/1139` `2b14182` (macOS arm64, ReleaseSafe), `.sbc` caches cleared between runs.

baijum:
## Summary

- `define-syntax` inside `let`/`let*`/`lambda` bodies failed to resolve free references to enclosing lexical bindings — the hygiene renamer emitted `__hyg_N_var` identifiers that were never mapped back, producing "undefined variable" errors
- Added `captureLocalsOnTransformer` calls to all three `define-syntax` code paths, matching the existing `let-syntax` behavior
- Made `captureLocalsOnTransformer` public so `compiler_lambda.zig` and `compiler_ir.zig` can call it

## Test plan

- [x] New regression test `tests/scheme/hygiene/define-syntax-capture.scm` (9 cases: let, let*, lambda, intervening define, multiple locals, nested scopes, let-syntax parity, global ref, internal define)
- [x] All 1788 existing tests pass (0 failures)

Closes #1271

### DIFF
--- a/src/compiler_macro.zig
+++ b/src/compiler_macro.zig
@@ -259,8 +259,8 @@ pub fn compileDefineSyntax(self: *Compiler, args: Value, dst: u16) CompileError!
         tx.def_env_val = self.lib_env_val;
     }
 
-    // Store in macro table; inside a body, track the registration so
-    // the enclosing body scope removes it on exit (R7RS 5.3).
+    try captureLocalsOnTransformer(self, transformer);
+
     const name = types.symbolName(keyword);
     try self.recordBodyMacro(name);
     self.macros.put(name, transformer) catch return CompileError.OutOfMemory;
@@ -391,7 +391,7 @@ pub fn compileLetrecSyntax(self: *Compiler, args: Value, dst: u16, is_tail: bool
     restoreMacros(self, saved_names.items, saved_values.items);
 }
 
-fn captureLocalsOnTransformer(self: *Compiler, transformer: Value) CompileError!void {
+pub fn captureLocalsOnTransformer(self: *Compiler, transformer: Value) CompileError!void {
     if (self.locals.items.len == 0) return;
     const tx = types.toObject(transformer).as(types.Transformer);
     const caps = self.gc.allocator.alloc(types.CapturedLocal, self.locals.items.len) catch return CompileError.OutOfMemory;
--- a/src/compiler_ir.zig
+++ b/src/compiler_ir.zig
@@ -441,6 +441,7 @@ pub fn compileLambdaWithIR(self: *Compiler, args: Value, dst: u16, name: ?[]cons
                 tx_obj.def_env = env;
                 tx_obj.def_env_val = child.lib_env_val;
             }
+            try macro.captureLocalsOnTransformer(&child, tx_val);
             child.macros.put(ds_name, tx_val) catch return CompileError.OutOfMemory;
             current = types.cdr(current);
             continue;
--- a/src/compiler_lambda.zig
+++ b/src/compiler_lambda.zig
@@ -321,6 +321,7 @@ pub fn compileBodyForms(self: *Compiler, body: Value, opts: BodyOpts) CompileErr
                 tx.def_env = env;
                 tx.def_env_val = self.lib_env_val;
             }
+            try macro.captureLocalsOnTransformer(self, transformer);
             self.macros.put(name, transformer) catch return CompileError.OutOfMemory;
         } else {
             break;
--- a/tests/scheme/hygiene/define-syntax-capture.scm
+++ b/tests/scheme/hygiene/define-syntax-capture.scm
@@ -1,0 +1,76 @@
+;; Regression test for #1271: define-syntax must capture let/lambda
+;; bindings from its definition scope (same as let-syntax).
+(import (scheme base) (scheme write) (scheme process-context) (srfi 64))
+
+(test-begin "define-syntax-capture")
+
+;; --- let binding ---
+(test-equal "let binding captured by define-syntax"
+  7
+  (let ((y 7))
+    (define-syntax gety (syntax-rules () ((_) y)))
+    (gety)))
+
+;; --- let* binding ---
+(test-equal "let* binding captured by define-syntax"
+  7
+  (let* ((y 7))
+    (define-syntax gety (syntax-rules () ((_) y)))
+    (gety)))
+
+;; --- lambda parameter ---
+(test-equal "lambda parameter captured by define-syntax"
+  7
+  ((lambda (y)
+     (define-syntax gety (syntax-rules () ((_) y)))
+     (gety)) 7))
+
+;; --- let + intervening define before define-syntax ---
+(test-equal "let binding with intervening define"
+  7
+  (let ((y 7))
+    (define z 1)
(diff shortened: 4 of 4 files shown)


src/compiler_macro.zig
```diff
@@ -259,8 +259,8 @@ pub fn compileDefineSyntax(self: *Compiler, args: Value, dst: u16) CompileError!
         tx.def_env_val = self.lib_env_val;
     }
 
-    // Store in macro table; inside a body, track the registration so
-    // the enclosing body scope removes it on exit (R7RS 5.3).
+    try captureLocalsOnTransformer(self, transformer);
```
baijum, line 262: Minor: this replaced the comment that explained the two lines below it —

```
// Store in macro table; inside a body, track the registration so
// the enclosing body scope removes it on exit (R7RS 5.3).
```

The R7RS 5.3 rationale for `recordBodyMacro` is now undocumented, and the new call has no comment even though the surrounding hygiene code is heavily annotated. Consider a one-liner, e.g. `// Capture enclosing let/lambda lexicals so the transformer can resolve free template refs to them (parity with let-syntax, #1271).`, and keeping the R7RS 5.3 note above `recordBodyMacro`.

baijum (commented):
## Review — ready to merge ✅

Correctly closes #1271. The fix is minimal and lands the `captureLocalsOnTransformer` call in exactly the three `define-syntax` registration paths that were missing it, bringing `define-syntax` to parity with the `let-syntax`/`letrec-syntax` paths that already had it:

- `compileDefineSyntax` (`compiler_macro.zig:262`)
- `compileBodyForms` legacy body scan (`compiler_lambda.zig:324`)
- `compileLambdaWithIR` IR body scan (`compiler_ir.zig:444`)

I confirmed these are the only registration sites — no fourth path was missed.

### Verification (macOS arm64, ReleaseSafe, `.sbc` caches cleared)

Built both the PR branch and the merge base (`f68f43f`) and compared:

| Case | base (`f68f43f`) | PR |
|---|---|---|
| `(let ((y 7)) (define-syntax gety (syntax-rules () ((_) y))) (gety))` | ❌ `undefined variable '__hyg_1_y'` | ✅ `7` |
| new `tests/scheme/hygiene/define-syntax-capture.scm` | — | ✅ 9/9 |
| Zig unit tests (`zig build test`) | — | ✅ pass |
| R7RS suite | — | ✅ 1395 pass, 0 fail |
| hygiene + macros smoke suites | — | ✅ all pass |

No regressions.

### Safety / lifecycle notes (all fine)

- `captureLocalsOnTransformer` early-returns when `self.locals` is empty, so top-level/library `define-syntax` (including the `body_macro_depth == 0` → `env.put` path) is unaffected and picks up no spurious `captured_locals`. Good.
- It allocates via `gc.allocator` (raw, no `maybeCollect`), so no rooting is needed around the call; `captured_locals` is already freed in `gc_collect.zig` and deep-copied in `gc_deep_copy.zig`, and `CapturedLocal` holds no `Value`s so needs no GC marking.
- The call is consistently placed right after the `def_env` assignment and before the `macros.put`, matching the `let-syntax` ordering.

### Non-blocking: residual hygiene limitation (pre-existing, shared with `let-syntax`)

A use-site macro argument that happens to share a name with a captured def-site local is mis-resolved to the captured local:

```scheme
(let ((y 7))
  (define-syntax m (syntax-rules () ((_ x) (+ x y))))
  (let ((y 100))
    (m y)))   ; => 14, should be 107 (100 + 7)
```

This is **not introduced by this PR** — the exact `let-syntax` equivalent also returns `14`, and on `main` this case is a hard `undefined variable '__hyg_1_y'` error, so the PR is a strict improvement here. Since #1271 is scoped to plain free-reference capture, I'd leave this out of scope, but it's worth a follow-up issue tracking the `captured_locals` machinery's handling of use-site args that collide with captured names.

One small inline nit below; not blocking.

baijum:
Filed the follow-up for the arg-vs-free residual limitation noted above: #1288.

PR #1289, opened 2026-07-07
Support import-set modifiers in environment (#1189)

### LINKED ISSUE #1189: [R7RS 6.12] environment: import-set modifiers (only/except/prefix/rename) are rejected — only plain library names accepted
**Spec reference:** R7RS §6.12 — `(environment list1 ...)`: "This procedure returns a specifier for the environment that results by starting with an empty environment and then importing each list, considered as an **import set**, into it." Import sets include `only`, `except`, `prefix`, and `rename` (§5.2).
**Severity:** missing-feature

**Minimal reproduction:**
```scheme
(eval '(+ 1 2) (environment '(only (scheme base) +)))
;; error: type error in 'environment': got (only (scheme base) +)
(eval '(base:+ 1 2) (environment '(prefix (scheme base) base:)))
;; error: type error in 'environment': got (prefix (scheme base) base:)
```

**Expected:** 3 in both cases.
**Actual:** `environmentFn` (src/primitives_r7rs.zig) feeds each argument straight to `libraryNameToString`, which only understands plain library names. The import-set machinery already exists in `vm_library.zig` `processImportSet` (handles only/except/rename/prefix for `import`) — `environment` just doesn't route through it.

Note: Phase 1D's gap test "environment restricts bindings" wraps this call in a `guard`, so complete rejection of `only` and correct restriction both pass — which is how the gap survived that phase.

**Verified:** fresh ReleaseSafe build at 36edbe0
**Source location:** src/primitives_r7rs.zig `environmentFn`
**Disabled test:** tests/scheme/audit/primitives_r7rs-audit.scm (marked `;; FAIL: #this`)

**Found by:** Systematic audit, Phase 2.14

baijum:
## Summary

- Route `environment`'s arguments through `processImportSet` (the same path used by `import`) instead of manually calling `libraryNameToString`, which only accepted plain library names
- Adds support for `only`, `except`, `prefix`, `rename`, and nested combinations in `(environment ...)` as required by R7RS §6.12
- Enables two previously-disabled audit tests

## Test plan

- [x] Unit test covering all four modifiers (`only`, `except`, `prefix`, `rename`), nested modifiers (`only` on `prefix`), and restriction enforcement
- [x] Audit test `primitives_r7rs-audit.scm` passes (44/44)
- [x] R7RS test suite passes (0 failures)
- [x] `zig build test` passes with zero leaks

Closes #1189

### DIFF
--- a/src/primitives_r7rs.zig
+++ b/src/primitives_r7rs.zig
@@ -208,22 +208,13 @@ fn environmentFn(args: []const Value) PrimitiveError!Value {
     // Create a new environment containing bindings from the given import sets.
     const vm = vm_mod.vm_instance orelse return PrimitiveError.TypeError; // bare-ok: no VM
     const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
-    const library_mod = @import("library.zig");
 
     const env_map = gc.allocator.create(std.StringHashMap(Value)) catch return PrimitiveError.OutOfMemory;
     env_map.* = std.StringHashMap(Value).init(gc.allocator);
 
+    const vm_library = @import("vm_library.zig");
     for (args) |import_set| {
-        const lib_name = library_mod.libraryNameToString(gc.allocator, import_set) catch return PrimitiveError.TypeError; // bare-ok: invalid import set
-        defer gc.allocator.free(lib_name);
-
-        const vm_library = @import("vm_library.zig");
-        vm_library.ensureLibraryLoaded(vm, import_set, lib_name) catch return PrimitiveError.TypeError; // bare-ok: library load failure
-        const lib = vm.libraries.get(lib_name) orelse return PrimitiveError.TypeError; // bare-ok: library not found
-        var it = lib.exports.iterator();
-        while (it.next()) |entry| {
-            env_map.put(entry.key_ptr.*, entry.value_ptr.*) catch return PrimitiveError.OutOfMemory;
-        }
+        vm_library.processImportSet(vm, env_map, import_set) catch return PrimitiveError.TypeError; // bare-ok: invalid import set
     }
 
     return gc.allocEnvironment(env_map, true, true) catch return PrimitiveError.OutOfMemory;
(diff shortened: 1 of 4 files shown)


baijum:
## Review

Solid fix. Routing `environmentFn` through the existing `processImportSet` is exactly the right call — it deletes the hand-rolled export-copy loop and reuses the machinery that already backs `import`, so `only`/`except`/`prefix`/`rename` (and nested combos, via the recursive `resolveImportBindings`) come for free. As a bonus, the new path also gets circular-import detection and macro free-ref copying (`copyTransformerFreeRefs`) that the old direct `env_map.put` loop lacked. Tests are good and the previously-disabled audit assertions are correctly re-enabled. CI is green across all platforms.

Two cleanups worth doing before merge — neither is a correctness bug in the happy path:

### 1. `ensureLibraryLoaded` is now dead code

`environmentFn` was the only caller of `vm_library.ensureLibraryLoaded`. After this change, `git grep ensureLibraryLoaded` on the branch matches only its own definition at `src/vm_library.zig:138`. Its doc comment even reads *"Used by `environment` procedure"*, which is now false. Recommend deleting the function (and its stale comment) in this PR so the dead code doesn't linger.

### 2. `env_map` leaks when an import set fails (seconding CodeRabbit)

CodeRabbit already flagged this: `env_map` is created before the loop and every failure path returns `PrimitiveError.TypeError` without `deinit`/`destroy`, so `(environment 42)` and `(environment '(no such library))` leak the temporary map. This is **pre-existing** (the old code had the identical pattern), so it's not a regression — but since this PR already owns the function, a one-line fix is a cheap win:

```zig
const env_map = gc.allocator.create(std.StringHashMap(Value)) catch return PrimitiveError.OutOfMemory;
env_map.* = std.StringHashMap(Value).init(gc.allocator);
errdefer {
    env_map.deinit();
    gc.allocator.destroy(env_map);
}
```

The final `allocEnvironment(env_map, true, true)` takes ownership on success, so `errdefer` only fires on the error paths — no double-free. (The unit tests don't exercise this path under the leak-checking allocator, which is why `zig build test` still reports zero leaks.)

### Verdict

Correct, well-tested, and ready to merge as-is. I'd fold in the `ensureLibraryLoaded` removal (and ideally the `errdefer`) first — both are quick, in-scope tidy-ups for a function this PR is already rewriting.

PR #1291, opened 2026-07-07
Expand macros during set! target pre-scan (#1250)

### LINKED ISSUE #1250: Macro-introduced set! bypasses assignment conversion (#1168 follow-up)
**Follow-up from #1168 / PR #1249.**

`collectSetTargets` runs on the raw pre-expansion S-expression (compiler.zig:383), so a `set!` introduced by macro expansion never lands in `set_targets`. The local stays register-allocated and the #1168 rollback/hang persists.

**Reproduction (both hang — run with a timeout):**

```scheme
;; A: macro wraps the set!
(define-syntax inc! (syntax-rules () ((_ v) (set! v (+ v 1)))))
(let ((n 0) (k* #f))
  (call/cc (lambda (k) (set! k* k)))
  (inc! n)
  (if (< n 3) (k* #f))
  n)

;; B: macro expands to the whole let+set!
(define-syntax counter-loop
  (syntax-rules ()
    ((_ limit)
     (let ((n 0) (k* #f))
       (call/cc (lambda (k) (set! k* k)))
       (set! n (+ n 1))
       (if (< n limit) (k* #f))
       n))))
(counter-loop 3)
```

**Why the simple fix doesn't work:** Adding targets during IR lowering (in `lowerSet`) discovers them too late — the `let` bindings are already compiled without boxing by the time the body is lowered. Boxing mid-body (at the `set!` site) fails because the continuation snapshot captured the raw register value before the `box_local`, so restore puts the raw value back and each re-entry creates a fresh box.

**Possible approaches:**
1. Pre-expand macros in binding form bodies before compiling the bindings (two-pass: expand then compile)
2. Restructure `compileLet`/`compileLetStar` to lower the full body first (collecting targets), then compile bindings with boxing, then compile the already-lowered body
3. Full assignment conversion at the IR level (lower `set!` to explicit box/unbox nodes)

Disabled test cases added in `tests/scheme/continuations/callcc-set-store.scm`.

baijum:
## Summary

- `collectSetTargets` now expands known macros to discover `set!` targets introduced by macro templates (Part A — catches use-site targets like `n` in `(inc! n)`)
- Post-expansion scan in `expandAndCompileMacroUse` catches template-introduced targets whose hygienic names only stabilize at real expansion time (Part B — catches `counter-loop`-style macros)
- Adds 4 regression tests covering `let`, `let*`, lambda param, and whole-form macro expansion

## Test plan

- [x] `zig build run -- tests/scheme/continuations/callcc-set-store.scm` — 12/12 pass (4 new)
- [x] `bash tests/scheme/run-all.sh` — 1789 pass, 0 fail

Fixes #1250

### DIFF
--- a/src/compiler.zig
+++ b/src/compiler.zig
@@ -79,7 +79,7 @@ pub const Compiler = struct {
     // so nested lambda bodies see the same suppression set. Consulted by the
     // constant folders (IR and legacy) to avoid folding calls to a name that
     // may be reassigned before the call executes.
-    set_targets: ?*const std.StringHashMap(void) = null,
+    set_targets: ?*std.StringHashMap(void) = null,
     scope_depth: u16 = 0,
     next_register: u16 = 0,
     parent: ?*Compiler = null,
@@ -341,6 +341,15 @@ pub const Compiler = struct {
         }
     }
 
+    /// Scan `expr` for set! targets and add them to the current set_targets map.
+    /// Called from expandAndCompileMacroUse after expansion to discover targets
+    /// introduced by macro templates (#1250).
+    pub fn scanSetTargets(self: *Compiler, expr: Value) CompileError!void {
+        if (self.set_targets) |st| {
+            try collectSetTargets(self, expr, st, 0);
+        }
+    }
+
     /// Box a local if its name is a set! target anywhere in the top-level form.
     /// R7RS §3.4: set! modifies the store (a heap location), not the
     /// continuation. Register-allocated mutated locals violate this when a
@@ -422,7 +431,7 @@ pub const Compiler = struct {
         // (including in nested lambda bodies, which inherit this set).
         var set_targets = std.StringHashMap(void).init(self.gc.allocator);
         defer set_targets.deinit();
-        try collectSetTargets(expr_root, &set_targets);
+        try collectSetTargets(self, expr_root, &set_targets, 0);
         self.set_targets = &set_targets;
         defer self.set_targets = null;
 
@@ -836,10 +845,12 @@ fn formatSyntaxError(args: Value) void {
 /// Recursively collect the symbol names that appear as the target of a
 /// `(set! <name> ...)` anywhere in `expr` into `out`. Used to suppress
 /// constant folding of those names within the enclosing form (see
-/// Compiler.set_targets). Conservative: it scans every sub-form except the
-/// interior of `quote`d data. Iterates the cdr spine to stay bounded on long
-/// lists and only recurses into car sub-forms.
-fn collectSetTargets(expr: Value, out: *std.StringHashMap(void)) CompileError!void {
+/// Compiler.set_targets) and to box locals for correct continuation
+/// semantics (R7RS §3.4). Conservative: it scans every sub-form except
+/// the interior of `quote`d data. When a known macro is encountered,
+/// it is expanded and the expansion is scanned (#1250).
+fn collectSetTargets(self: *Compiler, expr: Value, out: *std.StringHashMap(void), depth: u16) CompileError!void {
+    if (depth > 256) return;
     var cur = expr;
     while (types.isPair(cur)) {
         const head = types.car(cur);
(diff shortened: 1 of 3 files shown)


src/compiler.zig
```diff
@@ -854,9 +865,30 @@ fn collectSetTargets(expr: Value, out: *std.StringHashMap(void)) CompileError!vo
...
+                    &self.macros,
+                    .{},
+                ) catch {
+                    self.gc.no_collect -= 1;
+                    try collectSetTargets(self, head, out, depth);
+                    cur = types.cdr(cur);
+                    continue;
+                };
+                var expanded_root = expanded;
+                self.gc.pushRoot(&expanded_root);
```
baijum, line 888: **GC root not released on the error path.** `pushRoot(&expanded_root)` is paired with a manual `popRoot()` on line 887, with a `try collectSetTargets(...)` in between. If that recursive call returns an error — `collectSetTargets` can return `CompileError.OutOfMemory` from either `out.put` site — the `try` propagates and line 887 never runs. Two problems follow:

1. `root_count` is left incremented (a leaked root slot; enough leaks eventually hit the `@panic("GC root stack overflow")` at 1024).
2. Worse, `root_buffer` still holds `&expanded_root`, a pointer into *this* stack frame, which is about to unwind. The next GC mark dereferences that dangling stack slot → use-after-free.

`expandAndCompileMacroUse` handles the identical pattern with `defer self.gc.popRoot();` (compiler_macro.zig:162). The same fix works here — and because this branch does an immediate `return`, the function-scoped `defer` is equivalent to the manual pop on the happy path:

```zig
var expanded_root = expanded;
self.gc.pushRoot(&expanded_root);
defer self.gc.popRoot();
self.gc.no_collect -= 1;
try collectSetTargets(self, expanded_root, out, depth + 1);
return;
```

(Only reachable under OOM, so latent — but it's a genuine rooting-discipline violation and a one-liner to close.)

baijum (commented):
## Review — Expand macros during set! target pre-scan (#1250)

The approach is sound and the two-part split is well reasoned:
- **Part A** (expand known macros in `collectSetTargets`) catches use-site targets like `n` in `(inc! n)`, where the target is a pattern variable bound to a use-site identifier that is *not* hygienically renamed — so the pre-scan can add it before the `let`/`lambda` binding is compiled, which is exactly when boxing must happen (per the timing constraint spelled out in #1250).
- **Part B** (`scanSetTargets` after real expansion in `expandAndCompileMacroUse`) covers template-introduced targets whose hygienic names only stabilize at real expansion time (`counter-loop`).

The `depth > 256` guard bounds runaway/recursive macro expansion, the `no_collect` bracketing around `expandMacro` is balanced on both the success and catch paths, and the 4 new regression tests map cleanly onto the two reproductions plus `let*`/lambda-param variants. Nice.

### One thing to fix before merge

**GC root leak on the error path in `collectSetTargets`** (inline). The new macro branch uses a manual `pushRoot` … `popRoot` around a `try`, so an error return skips the `popRoot` and leaves a dangling pointer into an unwound stack frame. The sibling `expandAndCompileMacroUse` already handles the identical pattern with `defer self.gc.popRoot();` (compiler_macro.zig:162) — the same one-line change applies here. Details inline.

### Non-blocking notes (optional)

1. **Double expansion cost.** `collectSetTargets` now runs over the whole top-level form on every `compile()`, expanding *every* macro it encounters — and each of those macros is expanded again for real during compilation. For macro-heavy code this roughly doubles expansion work per form regardless of whether the form contains any `set!`. Not a correctness issue (hygiene generates fresh names each time, so perturbing the gensym counter is harmless), but if compile-time shows up, a cheap guard — skip the expanding scan when the form contains no `set!` at all, or when `self.macros` is empty — would recover most of the cost.
2. **Approximate pre-scan.** The pre-scan passes an empty `UseSiteBindingCheck{}` whereas the real expansion passes `resolveLocalSkipAliases`, so the pre-scan expansion can diverge from the real one. That's fine given the graceful fallback + Part B, but a one-line comment noting the pre-scan is intentionally best-effort would help future readers.
3. **Catch-branch fallback scans `head`.** In the expansion-failure `catch`, `try collectSetTargets(self, head, out, depth)` scans the bare macro-keyword symbol, which contributes nothing; the subsequent `cur = cdr` is what actually re-scans the arguments. Harmless, but the `head` scan is dead work.
4. **Test gap.** No test exercises a *nested* macro (a macro whose expansion introduces another macro that does `set!`) or the fallback path (a macro that fails to expand during the pre-scan). Worth one more case if cheap.

Once the `defer` fix lands, this looks good to merge.

baijum (commented):
## Follow-up review — commit `a4fdb57`

Re-reviewed after the fix commit. All blocking/actionable items from the previous round are resolved:

- ✅ **GC root leak** — `defer self.gc.popRoot();` now added right after `pushRoot`, manual `popRoot()` removed. Correct, and matches the `expandAndCompileMacroUse` pattern.
- ✅ **Approximate pre-scan** — comment added explaining the empty `UseSiteBindingCheck{}` is best-effort and that Part B corrects any misses.
- ✅ **Dead `head` scan in the catch branch** — removed; the `cur = cdr` still re-scans the arguments.
- ✅ **Nested-macro test** — test 13 (`apply-twice` → `inc!`) added, exercising the depth+1 recursion path.

The only unaddressed note is the optional double-expansion compile-time cost, which I'd flagged as non-blocking — fine to defer.

Verified locally on `a4fdb57` (built from clean worktree):
- `zig build test` — passes
- `tests/scheme/continuations/callcc-set-store.scm` — 13/13 pass, exit 0

**LGTM — ready to merge.** 🤖 Reviewed with Claude Code

PR #1292, opened 2026-07-07
Add GC write barriers to readListTail setCdr calls (#1267)

### LINKED ISSUE #1267: Reader: readListTail setCdr calls lack GC write barriers (old→young edges missed)
## Summary

`readListTail` in `src/reader_datum.zig` has two in-place pair mutations with no generational-GC write barrier:

- the dotted-tail case: `types.setCdr(tail, rest);` (line 153 on current `main`; shifts after #1257 merges)
- the proper-list append case: `types.setCdr(tail, pair);` (line 171)

Per `.claude/rules/gc-safety.md`, every store of a Value into a heap object field must be followed by `gc.writeBarrier(container, new_val)`. These are the last known unbarriered mutations in the reader — the datum-label patching sites in the same file were fixed in #1257.

## Mechanism

- Minor GCs run mid-read: every `readDatum` allocation can trigger a collection (see the rooting comment in `readVector`).
- Objects surviving two minor collections are promoted to the old generation (`gc_collect.zig:284-292`).
- `tail` stays reachable through the rooted `result` chain, so it survives those collections — which is exactly what promotes it.
- After promotion, `setCdr(tail, <freshly allocated pair>)` creates an old→young edge with no remembered-set entry, so the next minor collection cannot see the young pair through that edge.

## Failure scenario

Read a list whose middle element is large enough to trigger ≥2 minor collections while it parses:

```scheme
(read (open-input-string "(a <element with tens of thousands of subforms> b c)"))
```

While the big element parses, the pair holding `a` is promoted. The append `setCdr` then stores the young `(b …)` pair into it unbarriered. If the program later reaches the tail only through that promoted pair, the next minor collection frees the young pair and traversal is a use-after-free — 0xAA poison crash in Debug builds, silent corruption in ReleaseSafe. `-Dgc-stress=true` shrinks the repro to a handful of forms.

## Fix

Add barriers after both stores, following the pattern from #1257 and `set-cdr!` (`primitives.zig:402`):

```zig
types.setCdr(tail, pair);
self.gc.writeBarrier(types.toObject(tail), pair);
```

While there, grep `reader.zig` / `reader_tokens.zig` / `reader_datum.zig` for any remaining `setCar`/`setCdr`/direct Value-field stores to confirm nothing else is missed.

## Test

A Zig unit test (or a `-Dgc-stress=true` build run) that reads a list with a large middle element, keeps only a reference reaching the tail through an early pair, forces a minor collection, and touches the tail — Debug poisoning catches the UAF. Plus the usual `zig build test` and `bash tests/scheme/run-all.sh`.

Found during review of #1257 (same bug class, different function).

baijum:
## Summary

- Add `gc.writeBarrier()` after both `setCdr` calls in `readListTail` (`src/reader_datum.zig`) — the dotted-tail case and the proper-list append case — to record old→young edges in the remembered set
- Add two GC-stress regression tests that read proper and dotted lists with `gc.stress = true`, then walk the spine to verify no pairs were silently freed

Fixes #1267

## Test plan

- [x] `zig build test` — all unit tests pass (including new GC-stress reader tests)
- [x] `bash tests/scheme/run-all.sh` — 1789 pass, 0 fail
- [ ] CI

### DIFF
--- a/src/reader_datum.zig
+++ b/src/reader_datum.zig
@@ -157,6 +157,7 @@ fn readListTail(self: *Reader) ReadError!Value {
                 result = rest;
             } else {
                 types.setCdr(tail, rest);
+                self.gc.writeBarrier(types.toObject(tail), rest);
             }
             try self.skipWhitespaceAndCommentsChecked();
             if (self.pos >= self.source.len or self.source[self.pos] != ')') {
@@ -176,6 +177,7 @@ fn readListTail(self: *Reader) ReadError!Value {
             result = pair;
         } else {
             types.setCdr(tail, pair);
+            self.gc.writeBarrier(types.toObject(tail), pair);
         }
         tail = pair;
     }
--- a/src/tests_robustness.zig
+++ b/src/tests_robustness.zig
@@ -89,7 +89,15 @@ test "reader: valid nested block comment" {
     try std.testing.expectEqual(@as(i64, 42), types.toFixnum(result));
 }
 
-test "reader: readListTail write barrier under GC stress (proper list)" {
+// GC-stress smoke tests for readListTail. These verify structural integrity
+// (spine length, car payloads, terminator) under gc.stress=true. They do NOT
+// fail without the write barriers because markRoots traces transitively
+// through old objects, so the rooted `result` keeps the entire spine alive
+// regardless of remembered-set state. The barriers are verified correct by
+// inspection against gc-safety.md; these tests guard against other reader
+// regressions under heavy GC pressure.
+
+test "reader: readListTail under GC stress (proper list)" {
     var gc = memory.GC.init(std.testing.allocator);
     defer gc.deinit();
     gc.stress = true;
@@ -98,18 +106,18 @@ test "reader: readListTail write barrier under GC stress (proper list)" {
     defer reader.deinit();
     const result = try reader.readDatum();
 
-    // Walk the spine — without barriers, promoted pairs lose young tails.
+    const expected = [_][]const u8{ "a", "b", "c", "d", "e", "f", "g", "h" };
     var cur = result;
-    var len: usize = 0;
-    while (types.isPair(cur)) {
+    var i: usize = 0;
+    while (types.isPair(cur)) : (i += 1) {
+        try std.testing.expectEqualStrings(expected[i], types.symbolName(types.car(cur)));
         cur = types.cdr(cur);
-        len += 1;
     }
-    try std.testing.expectEqual(@as(usize, 8), len);
+    try std.testing.expectEqual(@as(usize, 8), i);
     try std.testing.expectEqual(types.NIL, cur);
 }
 
-test "reader: readListTail write barrier under GC stress (dotted list)" {
+test "reader: readListTail under GC stress (dotted list)" {
     var gc = memory.GC.init(std.testing.allocator);
     defer gc.deinit();
     gc.stress = true;
@@ -118,15 +126,15 @@ test "reader: readListTail write barrier under GC stress (dotted list)" {
     defer reader.deinit();
     const result = try reader.readDatum();
 
+    const expected = [_][]const u8{ "a", "b", "c" };
     var cur = result;
-    var len: usize = 0;
-    while (types.isPair(cur)) {
+    var i: usize = 0;
+    while (types.isPair(cur)) : (i += 1) {
+        try std.testing.expectEqualStrings(expected[i], types.symbolName(types.car(cur)));
         cur = types.cdr(cur);
-        len += 1;
     }
-    try std.testing.expectEqual(@as(usize, 3), len);
-    // Tail should be symbol 'd', not NIL or garbage.
-    try std.testing.expect(types.isSymbol(cur));
+    try std.testing.expectEqual(@as(usize, 3), i);
+    try std.testing.expectEqualStrings("d", types.symbolName(cur));
 }
 
 test "reader: rational with zero denominator" {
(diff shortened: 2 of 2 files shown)


baijum:
## Review

Reviewed the change against the current tree, `gc-safety.md`, and the marking code. The **code fix is correct and complete**, but I want to flag that the **tests don't actually fail without the fix** — I verified this empirically — so they don't function as regression tests. Details below.

### ✅ The fix (`src/reader_datum.zig`) — correct and complete

- Both `setCdr` sites in `readListTail` now carry `gc.writeBarrier(types.toObject(tail), …)` immediately after the store, matching the established pattern at `set-cdr!` (`primitives.zig:404`) and the #1257 datum-label sites (`reader_datum.zig:66,68,254,262,272`).
- The argument types line up with `writeBarrier(self: *GC, container: *Object, new_val: Value)`.
- I grepped `reader.zig` / `reader_tokens.zig` / `reader_datum.zig` for every `setCar`/`setCdr`: after this PR **all six sites are barriered**, so the "grep for remaining stores" ask in #1267 is satisfied — nothing else is missed.

This is a correct, minimal, low-risk change that brings the last two reader mutation sites into compliance with the documented write-barrier invariant.

### ⚠️ The tests (`src/tests_robustness.zig`) — they pass **with or without** the barriers

I removed the two new barriers (reverting to the pre-fix reader), kept the two new tests, and ran `zig build test -Doptimize=Debug` (Debug ⇒ `0xAA` poisoning active). **Both tests passed.** To confirm they actually execute, I flipped the proper-list assertion to `expectEqual(999, len)` and got:

```
'tests_robustness.test.reader: readListTail write barrier under GC stress (proper list)' failed:
expected 999, found 8
```

So the tests run, walk the full 8-pair spine, find the intact `NIL` tail (and the dotted `d` tail) — even with the barriers removed under `gc.stress = true`.

**Why they can't catch this bug class:** `readListTail` keeps `result` rooted for the entire read, and `minorCollect` → `markRoots` → `markValueInner` (`gc_collect.zig:522-558`) marks **transitively through old objects reached from roots** — there is no "stop at old generation" gate. So the freshly-appended young `pair` is always reachable from the rooted `result` and always gets marked; `sweepYoung` never frees it, barrier or not. The failure precondition in #1267 — "the program later reaches the tail **only** through that promoted pair" — is never met inside `readListTail`, because the spine stays reachable via the separate rooted `result`.

Net: these are fine as reader smoke tests, but they don't satisfy the repo rule "every bug fix MUST include a regression test that **fails without the fix and passes with it**." A test that genuinely exercises this bug class would need to drop the direct root and reach the promoted pair **only** via an unbarriered old→young edge (i.e. through the remembered set), which is hard to make deterministic here precisely because root marking is transitive.

### CI note

`test (ubuntu-latest, Debug)` shows red, but it's a **~20 min timeout/cancellation**, not a test failure: the unit-test step completed and the Debug Scheme suites passed 1789/0 before `##[error]The operation was canceled`. The preceding `main` merge (#1291) shows the same cancelled CI, so this is a pre-existing slow-Debug infra timeout, not something this PR introduced. All other variants (macOS, ubuntu ReleaseSafe/ReleaseFast, ubuntu-arm, riscv64, wasm, format) and CodeRabbit are green.

### Recommendation

The **code change is good to merge** as write-barrier-invariant hygiene. Before merging I'd suggest deciding what to do about the tests, since as written they give false confidence (they'd stay green if someone later deleted the barriers):
- keep them but reframe them as reader smoke tests (not regression tests), and note in #1267 that a deterministic reader-level repro isn't achievable because `result` keeps the spine rooted; or
- replace them with a direct GC-level test that builds an unbarriered old→young edge reachable only via the remembered set.

baijum:
Thanks for the thorough review. You're right — I traced through the GC code and confirmed: `minorCollect` → `markRoots` → `markValueInner` traces transitively through old objects (no generation gate), so the rooted `result` always reaches the young tail pairs regardless of write barriers. The remembered set is defense-in-depth for the documented invariant, not load-bearing with the current marking implementation.

Updated in 9c1103b:
- Reframed the tests as GC-stress smoke tests with a block comment explaining why they can't serve as barrier regression tests
- Added car-payload assertions (symbol name checks) per CodeRabbit's suggestion
- Kept them since they still exercise the reader under heavy GC pressure

The barrier fix itself is verified correct by inspection against `gc-safety.md` and grep confirmation that all 6 `setCar`/`setCdr` sites in the reader are now barriered.

baijum:
## Follow-up review — `9c1103b` addresses the feedback ✅

The reframe resolves the test concern from my earlier comment:

- **Honest framing.** The new doc comment states plainly that these tests *do not* fail without the barriers, and why (`markRoots` traces transitively through old objects, so the rooted `result` keeps the whole spine alive regardless of remembered-set state). That removes the false-confidence problem — no one reading these will mistake them for barrier regression guards.
- **Names fixed.** Dropped "write barrier" from the test names, matching what they actually exercise.
- **Stronger smoke coverage.** The new car-payload checks (`symbolName(car(cur))` against `a…h` / `a b c . d`) verify element identity and ordering under `gc.stress=true`, not just spine length — genuinely better reader-under-GC-pressure coverage.

The barriers themselves are unchanged and remain correct/complete (all six reader `setCar`/`setCdr` sites barriered; matches `set-cdr!` and the #1257 pattern).

**Verified locally:** applied this PR head (`9c1103b`) to a clean tree and ran `zig build test` → passes. `types.symbolName` exists (`types.zig:1065`), so the new tests compile.

### Merge readiness

**Ready to merge** once the fresh CI run on `9c1103b` goes green. The only prior red check was `test (ubuntu-latest, Debug)`, which was a ~20 min job timeout (unit tests + Scheme suites had already passed) — the same cancellation also hit the preceding `main` merge #1291, so it's a pre-existing slow-Debug infra flake, not a blocker for this change. `format`, `wasm`, and CodeRabbit are already green on the new commit.

PR #1293, opened 2026-07-07
Clear stale registers in tail-call window extension (#1256)

### LINKED ISSUE #1256: [GC] Stale registers in tail-call window extension
**Found by:** Code review of #1254 (follow-up, not a blocker)

**Description:** The three tail-call opcodes (`tail_call`, `tail_call_global`, `tail_call_cc` in `vm_dispatch.zig`) mutate the frame in place with `frame.closure = closure` after moving only the args to `frame.base`. When the new callee's `locals_count` is larger than the old one's, the GC scan window `[base, base + locals_count)` extends over registers that were never covered by any frame — the same stale-register class as #1156.

`self_tail_call` is safe (same closure, window unchanged and continuously covered).

**Repro shape under `-Dgc-stress=true`:** a function with a small frame calls a helper (child frame writes registers above the caller's window), the helper returns, an allocation triggers GC (frees the helper's now-unreachable locals), then the function tail-calls something with a larger frame — the next GC scans dangling pointers in the window extension.

**Fix:** `clearFrameLocals(vm, frame.base, arg_count, new_func.locals_count)` at each of the three tail-call sites in `vm_dispatch.zig`.

baijum:
## Summary

- Make `clearFrameLocals` public in `vm_calls.zig` so `vm_dispatch.zig` can call it
- Add `clearFrameLocals` at the three tail-call closure sites (`tail_call`, `tail_call_global`, `tail_call_cc`) in `vm_dispatch.zig`, after copying args but before resetting `frame.ip`
- Add regression test `tests/scheme/smoke/tail-call-stale-regs-1256.scm`

Fixes #1256

## Test plan

- [x] New regression test passes normally
- [x] New regression test passes under `-Dgc-stress=true`
- [ ] `bash tests/scheme/run-all.sh` — full suite (CI)

### DIFF
--- a/src/tests_tail_calls.zig
+++ b/src/tests_tail_calls.zig
@@ -191,3 +191,55 @@ test "tail call to parameter with converter across frames array growth" {
     const result = try vm.eval("(p)");
     try std.testing.expectEqual(@as(i64, 50), types.toFixnum(result));
 }
+
+test "tail-call into larger frame clears extension registers" {
+    // Regression test for #1256: when a tail-call switches to a callee with
+    // a larger locals_count, registers in [base+args, base+locals_count) must
+    // be cleared to UNDEFINED so the GC doesn't scan stale pointers.
+    var gc = memory.GC.init(std.testing.allocator);
+    defer gc.deinit();
+    var vm = try th.makeTestVM(&gc);
+    defer vm.deinit();
+
+    // 1. Pollute high registers by calling a function with many locals that
+    //    allocates heap objects (strings), then returns a fixnum.
+    _ = try vm.eval(
+        \\(define (pollute)
+        \\  (let ((a "heap1") (b "heap2") (c "heap3") (d "heap4")
+        \\        (e "heap5") (f "heap6") (g "heap7") (h "heap8"))
+        \\    (string-length (string-append a b c d e f g h))))
+    );
+    _ = try vm.eval("(pollute)");
+
+    // 2. A small function that tail-calls a target with many locals.
+    //    The target only uses its first arg — the rest are declared but
+    //    untouched, so they must stay UNDEFINED after clearFrameLocals.
+    _ = try vm.eval(
+        \\(define (small x)
+        \\  (big x))
+    );
+    _ = try vm.eval(
+        \\(define (big n)
+        \\  (let ((a (+ n 1)) (b (+ n 2)) (c (+ n 3)) (d (+ n 4))
+        \\        (e (+ n 5)) (f (+ n 6)) (g (+ n 7)) (h (+ n 8)))
+        \\    (+ a b c d e f g h)))
+    );
+
+    const result = try vm.eval("(small 0)");
+    try std.testing.expectEqual(@as(i64, 36), types.toFixnum(result));
+
+    // 3. Verify the registers in the frame-base region are not stale heap
+    //    pointers. After eval returns, frame_count is 0 and the register
+    //    file retains whatever the last execution left. Registers that were
+    //    in the extension zone of the tail-call should have been cleared to
+    //    UNDEFINED (or overwritten by the callee's fixnum locals) — they
+    //    must NOT hold heap object pointers from the earlier `pollute` call.
+    //    Scan the first 32 registers (generous window covering any plausible
+    //    frame base) and verify none point to a freed heap string.
+    for (vm.registers[0..@min(32, vm.registers.len)]) |reg| {
+        if (reg != types.UNDEFINED and types.isPointer(reg)) {
+            const obj = types.toObject(reg);
+            try std.testing.expect(@intFromEnum(obj.tag) < 36);
+        }
+    }
+}
--- a/src/tests_tail_calls.zig
+++ b/src/tests_tail_calls.zig
@@ -196,13 +196,20 @@ test "tail-call into larger frame clears extension registers" {
     // Regression test for #1256: when a tail-call switches to a callee with
     // a larger locals_count, registers in [base+args, base+locals_count) must
     // be cleared to UNDEFINED so the GC doesn't scan stale pointers.
+    //
+    // Strategy: pollute registers with heap pointers, then tail-call into a
+    // function whose locals_count is large (due to a branch with many locals)
+    // but whose taken branch is trivial and doesn't touch the high registers.
+    // After execution, the untouched extension slots should be UNDEFINED (if
+    // cleared) not stale heap pointers (if not cleared).
     var gc = memory.GC.init(std.testing.allocator);
     defer gc.deinit();
     var vm = try th.makeTestVM(&gc);
     defer vm.deinit();
 
-    // 1. Pollute high registers by calling a function with many locals that
-    //    allocates heap objects (strings), then returns a fixnum.
+    // 1. Pollute registers: call a function with many locals that allocates
+    //    heap objects (strings). After it returns, those register slots retain
+    //    stale string pointers.
     _ = try vm.eval(
         \\(define (pollute)
         \\  (let ((a "heap1") (b "heap2") (c "heap3") (d "heap4")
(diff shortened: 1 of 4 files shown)


baijum (commented):
## Review

I reviewed the fix against the GC scan-window model and verified it empirically. Summary: **the fix is correct and safe, but the regression test does not actually exercise the bug — it passes identically with and without the fix.**

### The fix itself is correct ✅

The GC scans each frame's registers over `[frame.base, frame.base + frameWindow())`, where `frameWindow()` returns the closure's `locals_count` (`types.zig:520`, `vm.zig:62-65`). All three tail-call opcodes rebind `frame.closure` to the new callee in place, so after the tail call the GC scans the *new* callee's window. The fix clears exactly `[frame.base + arg_count, frame.base + locals_count)` right after copying args and before rebinding — precisely the extension the GC will now scan. The `used` argument is computed correctly at each site, including the variadic and `call/cc` cases:

- `tail_call` / `tail_call_global`: `used = arg_count` (= `arity + 1` for variadic, else `nargs`) ✔
- `tail_call_cc`: `used = arity + 1` for variadic, else `1` — matches the arg layout at lines 1136–1141 ✔

It mirrors `callClosure`/`callReentrant` exactly, and `self_tail_call` is correctly left alone (same closure, window unchanged). No behavioral regression: the existing `smoke/tail-calls.scm` still passes under `-Dgc-stress=true`.

### The regression test does not catch the bug ⚠️ (main finding)

`tests/scheme/smoke/tail-call-stale-regs-1256.scm` passes **identically on a fixed build and an unfixed build**, so it would not fail if this fix were reverted.

Evidence — I built two `-Dgc-stress=true` binaries: one from this branch, one from `main` (confirmed unfixed: 0 occurrences of the new `clearFrameLocals` call; confirmed genuinely stress-testing: `--gc-stats` shows 2044 collections for ~2000 allocations):

```
fixed   + gc-stress:  22/22 pass, exit 0
unfixed + gc-stress:  22/22 pass, exit 0  (5 runs, all identical)
```

**Root cause:** the tail-call target `medium-frame` does only fixnum arithmetic (`(+ a b c d e prev)`) — it never allocates. The bug only manifests when a GC runs *while the extended window holds a dangling pointer*, i.e. the target (or code running in the extended window) must allocate. With no allocation in the target, no GC ever scans the extension, so the stale pointer is never dereferenced.

I also wrote a test that follows the issue's exact repro shape — small caller → child that writes a heap pointer into a high register and returns a fixnum → allocation in the caller → tail-call into a *larger* frame that allocates. It **also** passes on the unfixed binary (10/10 runs, exit 0). This points at a real difficulty: under `-Dgc-stress`, any stale pointer that lands in a *scanned* window is *marked live* rather than freed, so it takes a precise free-then-scan ordering to actually crash — hard to force deterministically from Scheme.

### Recommendation

The one-line-per-site fix is sound and I'd be comfortable merging it on correctness grounds. But the accompanying test gives false confidence — it's green regardless of the fix. Before merging I'd suggest one of:

1. **A Zig unit test** (most reliable) that drives a tail-call into a larger-frame callee and asserts the extension registers are `types.UNDEFINED` afterward — this asserts the fix's actual mechanism directly, without depending on GC timing. This is the class of regression that `src/tests_*.zig` exists for.
2. Failing that, at minimum restructure the Scheme test so the tail-call target **allocates** in its body — necessary (though, per above, not clearly sufficient) to give it any chance of exercising the bug.

Happy to help write the Zig unit test if useful.

*(Reviewed with Claude Code — fix logic verified against the GC scan model; test-effectiveness claims verified with fixed vs. unfixed `-Dgc-stress` builds.)*

baijum (commented):
## Follow-up review (commit c59e3eb)

Thanks for the quick turnaround. Partial progress — one item resolved well, the core one still open.

### ✅ Two additional fix sites — correct, and a genuinely good catch

Adding `clearFrameLocals` to `tail_apply` and `tail_eval` closes the same bug class at two sites the original PR missed, so the fix is now complete across all five in-place tail-call closure paths. Both are correct:

- **`tail_apply`** (`vm_dispatch.zig:631`): `used = arg_count` (`arity + 1` variadic, else `total_nargs`) matches the args copied to `frame.base..` at 626–630 ✔
- **`tail_eval`** (`vm_dispatch.zig:1212`): `used = 0` is right — the eval'd expression compiles to a nullary thunk, so the whole window is fresh locals ✔

### ⚠️ The regression test still does not fail without the fix

This was the main point of the last review, and the new Zig test has the same gap. I reverted `src/vm_dispatch.zig` to `origin/main` (all 5 `clearFrameLocals` calls removed — confirmed `grep -c` = 0) and reran the suite:

```
unfixed dispatch:  Build Summary: 7/7 steps succeeded; 660/660 tests passed
```

`tests_tail_calls.zig` is wired in via `vm_tests.zig`, so `"tail-call into larger frame clears extension registers"` **is** compiled and run — it just passes with the fix reverted. Three concrete reasons it can't catch the defect:

1. **It inspects the wrong moment.** The assertion runs *after* `(small 0)` fully returns. By then `big` has executed and written fixnums into `a`…`h` — i.e. into the exact extension slots `clearFrameLocals` targets. So the post-hoc scan sees fixnums (`isPointer` false → skipped), never the stale pointers, with or without the fix.
2. **Nothing is ever freed.** `pollute` does ~10 allocations; the GC threshold is 8192, so no collection runs. The polluted strings stay live, so any surviving pointer has a valid tag regardless.
3. **ReleaseSafe → no poison.** The summary shows `compile test unit-tests ReleaseSafe`. Freed objects are only poisoned (`0xAA`) in Debug, so the `tag < 36` check can't detect a dangling pointer even if one existed.

Per `CLAUDE.md`: *"Every bug fix MUST include a regression test that fails without the fix and passes with it."* Both tests here pass without the fix, so this bar isn't met yet.

I also spent time trying to force a crash from Scheme following the issue's exact repro shape (small caller → child pollutes high regs → alloc in caller → tail-call into larger allocating frame) and **could not** get the unfixed `-Dgc-stress` binary to fault (10/10 clean). Under `-Dgc-stress`, any stale pointer that lands in a *scanned* window gets *marked live* rather than freed — so a deterministic crash needs a precise free-then-scan ordering that's very hard to hit from Scheme. My read: **a reliable regression test for this has to be a low-level Zig test**, e.g.:

- set up a frame for a small callee, write `types.makePointer(...)` sentinels into the extension slots `[base+args, base+small_locals..big_locals)`,
- drive the `tail_call` dispatch into a larger-`locals_count` callee,
- assert those slots are `types.UNDEFINED` immediately afterward (before the callee body runs).

That asserts the mechanism directly with zero GC-timing dependence. Happy to write it if you'd like.

### Merge readiness

The **code** is correct and now complete — I'd be comfortable with it on correctness grounds. But by the repo's own regression-test rule, it's **not quite ready**: neither test fails without the fix. I'd either land the deterministic Zig test above, or, if we accept that this defect is impractical to reproduce from Scheme, say so explicitly in the PR and reframe the current tests as smoke coverage (with an honest comment) rather than regression guards.

*(Verified by reverting `vm_dispatch.zig` to `origin/main` and rerunning `zig build test --summary all`: 660/660 pass unfixed.)*

baijum (commented):
## Follow-up review (commit 1e84476) — resolved ✅

The rewritten test now genuinely guards the fix. I independently verified both directions by reverting `src/vm_dispatch.zig` to `origin/main` (all 5 `clearFrameLocals` calls removed) and rerunning `zig build test`:

```
with fix:      Build Summary: 660/660 tests passed
without fix:   657 pass, 1 fail (658 total)  →  659/660
               'tail-call into larger frame clears extension registers' failed:
               expected 0, found 15
```

Reverting the fix leaves **15 stale heap pointers** in the extension window `[3, 22)`; the fix clears them to `UNDEFINED`. That's exactly the fail-without / pass-with behavior CLAUDE.md requires, and it matches your reported numbers. The new strategy — a branching callee with a large `locals_count` whose taken `#t` branch never touches the high registers — is the right way to make the extension slots observable after return. Nice.

### Final assessment

- **Fix:** correct and complete — all five in-place tail-call closure sites (`tail_call`, `tail_call_global`, `tail_call_cc`, `tail_apply`, `tail_eval`) now clear the extension window, matching `callClosure`; `self_tail_call` correctly untouched.
- **Tests:** a Scheme smoke test plus a Zig regression test that provably fails without the fix. Full unit suite green (660/660).

**This is ready to merge**, pending the usual CI (the full `run-all.sh` + platform matrix — the one unchecked box in the PR description).

### Two optional, non-blocking notes

1. The Zig test hardcodes frame base `1` and window `[3, 22)`. If register allocation ever changes `big-branch`'s `locals_count`, the scan window drifts — the failure mode is a *weaker* test (scanning always-`UNDEFINED` slots → vacuous pass), not a false failure, so it's safe, but a one-line `expectEqual` on the observed `locals_count`, or a comment pinning the assumption, would make future drift obvious.
2. The regression test exercises the `tail_call_global` path; `tail_apply` and `tail_eval` (the two sites added in c59e3eb) have no dedicated coverage. Not worth blocking on, but worth a note in case one silently regresses later.

*(Verified locally: `git checkout origin/main -- src/vm_dispatch.zig` → `zig build test` fails only this test with "expected 0, found 15"; restored → 660/660.)*

PR #1294, opened 2026-07-07
Root hash-table-walk/fold snapshot entries to prevent use-after-free (#1181)

### LINKED ISSUE #1181: [SRFI-69] hash-table-walk/fold: snapshot is invisible to GC — use-after-free when the callback deletes entries and allocates
**Category:** SRFI / GC safety
**Spec reference:** SRFI-69 `hash-table-walk`/`hash-table-fold` — the callback may mutate the table; iteration must remain memory-safe.
**Severity:** crash (use-after-free; manifests as corrupted keys/values handed to user code)

**Minimal reproduction (fails under `zig build -Dgc-stress=true`; passes on a default build because freed memory is rarely reused fast enough):**
```scheme
(define ht (make-hash-table))
(do ((i 0 (+ i 1))) ((= i 30))
  (hash-table-set! ht (cons i 'key) (cons i 'val)))
(define sum 0) (define bad 0)
(hash-table-walk ht
  (lambda (k v)
    (do ((j 0 (+ j 1))) ((= j 30)) (hash-table-delete! ht (cons j 'key)))
    (do ((j 0 (+ j 1))) ((= j 100)) (cons j (make-vector 4 j)))
    (if (and (pair? k) (pair? v) (eq? (cdr k) 'key) (eq? (cdr v) 'val)
             (= (car k) (car v)))
        (set! sum (+ sum (car k)))
        (set! bad (+ bad 1)))))
(list sum bad)  ;; expected (435 0) — actual under gc-stress: (12 29)
```

**Cause:** `snapshotLiveEntries` (src/primitives_hashtable.zig) copies the live entries into a raw `gc.allocator` buffer that the GC cannot see. When the callback deletes entries (their only other reference) and then allocates, the collector frees the snapshot's keys/values; subsequent iterations pass freed (and reused) objects to the callback. Verified: 29 of 30 entries corrupted under gc-stress.

**Prior art:** #690 (closed) introduced the snapshot to fix iterating live entries across a rehash — this is the follow-on hazard. `vectorPartitionFn` (src/primitives_vector.zig) documents and solves the identical pattern by rooting every element via `gc.rootedScope()` + `gc.extra_roots`; the same fix applies to both `hashTableWalkFn` and `hashTableFoldFn`.

**Verified:** fresh ReleaseSafe build at b170e6f; default build passes, gc-stress corrupts (localizes it to collection timing)
**Source location:** src/primitives_hashtable.zig `snapshotLiveEntries`, `hashTableWalkFn`, `hashTableFoldFn`
**Disabled test:** tests/scheme/audit/primitives_hashtable-audit.scm (marked `;; FAIL: #this`; note it only fails under -Dgc-stress=true)

baijum:
## Summary

- Root all snapshot keys/values via `gc.rootedScope()` + `gc.extra_roots` in both `hashTableWalkFn` and `hashTableFoldFn` so the GC can trace them during collection
- Re-enable the `FAIL: #1181` audit test that was disabled pending the fix
- Add dedicated regression test covering both `hash-table-walk` and `hash-table-fold`

Closes #1181

## Test plan

- [x] `zig build run -- tests/scheme/audit/primitives_hashtable-audit.scm` — 72/72 pass (default build)
- [x] `zig build -Dgc-stress=true run -- tests/scheme/audit/primitives_hashtable-audit.scm` — 72/72 pass (gc-stress)
- [x] `zig build run -- tests/scheme/smoke/hash-table-walk-gc-1181.scm` — 2/2 pass (default)
- [x] `zig build -Dgc-stress=true run -- tests/scheme/smoke/hash-table-walk-gc-1181.scm` — 2/2 pass (gc-stress)
- [ ] `bash tests/scheme/run-all.sh` — full suite (running)

### DIFF
--- a/src/primitives_hashtable.zig
+++ b/src/primitives_hashtable.zig
@@ -293,6 +293,13 @@ fn hashTableWalkFn(args: []const Value) PrimitiveError!Value {
     const snapshot = snapshotLiveEntries(gc, ht) orelse return PrimitiveError.OutOfMemory;
     defer gc.allocator.free(snapshot);
 
+    const scope = gc.rootedScope();
+    defer scope.release();
+    for (snapshot) |entry| {
+        gc.extra_roots.append(gc.allocator, entry.key) catch return PrimitiveError.OutOfMemory;
+        gc.extra_roots.append(gc.allocator, entry.value) catch return PrimitiveError.OutOfMemory;
+    }
+
     for (snapshot) |entry| {
         const call_args = [2]Value{ entry.key, entry.value };
         _ = vm.callWithArgs(proc, &call_args) catch |err| {
@@ -500,6 +507,13 @@ fn hashTableFoldFn(args: []const Value) PrimitiveError!Value {
 
     const snapshot = snapshotLiveEntries(gc, ht) orelse return PrimitiveError.OutOfMemory;
     defer gc.allocator.free(snapshot);
+
+    const scope = gc.rootedScope();
+    defer scope.release();
+    for (snapshot) |entry| {
+        gc.extra_roots.append(gc.allocator, entry.key) catch return PrimitiveError.OutOfMemory;
+        gc.extra_roots.append(gc.allocator, entry.value) catch return PrimitiveError.OutOfMemory;
+    }
 
     for (snapshot) |entry| {
         const call_args = [3]Value{ entry.key, entry.value, acc };
--- a/src/primitives_hashtable.zig
+++ b/src/primitives_hashtable.zig
@@ -293,6 +293,8 @@ fn hashTableWalkFn(args: []const Value) PrimitiveError!Value {
     const snapshot = snapshotLiveEntries(gc, ht) orelse return PrimitiveError.OutOfMemory;
     defer gc.allocator.free(snapshot);
 
+    // Root all entries up front: the first callback can delete+allocate and
+    // free a not-yet-visited entry's key/value that only the snapshot holds.
     const scope = gc.rootedScope();
     defer scope.release();
     for (snapshot) |entry| {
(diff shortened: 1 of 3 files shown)


baijum:
## Review — ready to merge ✅

Reviewed the code, the GC mechanism, and independently re-verified the regression test. This is a correct, well-scoped fix.

### The fix is correct
`snapshotLiveEntries` copies live entries into a raw `gc.allocator.alloc` buffer that the collector never scans. Once a callback deletes entries and allocates, a minor GC frees the snapshot's keys/values and later iterations hand corrupted objects to the callback. Rooting each key/value in `gc.extra_roots` fixes this because `markRoots` traces `extra_roots` (`src/gc_collect.zig:481`), so the snapshotted objects survive collection for the duration of the walk/fold.

A few things that are *right* about the approach and worth calling out:

- **Rooting all entries up front (separate loop) is necessary, not redundant.** The first callback can delete + allocate and free the key/value of an entry not yet visited. Per-call arg rooting inside `callWithArgs` (registers) only protects the *current* callback's args, so the not-yet-visited entries genuinely need to be pinned before the first callback runs. Worth a one-line comment so nobody later "optimizes" the two loops into one and silently reintroduces the bug.
- **`rootedScope()` + `defer scope.release()`** (`shrinkRetainingCapacity(base)`) is the right LIFO pattern, and the OOM-mid-append path is clean — the `defer` still unwinds the partially-pushed roots.
- Both call sites (`hashTableWalkFn`, `hashTableFoldFn`) are fixed identically and symmetrically.

I checked the other callback-invoking hashtable primitives (`hash-table-ref` thunk, `hash-table-update!/default`) — none snapshot-then-iterate, so this pair was the whole exposure.

### Verification (independent)
Built both sides with `-Dgc-stress=true` and ran the new smoke test:

- **main (without fix)**: `FAIL walk snapshot survives GC` + `FAIL fold snapshot survives GC` (exit 1) — confirms the test genuinely catches the use-after-free.
- **this branch (with fix)**: 2/2 pass.

So the re-enabled audit test and the new `tests/scheme/smoke/hash-table-walk-gc-1181.scm` are real regression tests, not tautologies. CI is fully green across the matrix (Debug/ReleaseSafe/ReleaseFast, arm/x86/riscv, wasm, format).

### Optional (non-blocking)
- The 6-line rooting loop is duplicated verbatim in both functions — could be a tiny `rootSnapshot(gc, snapshot)` helper, but inline is arguably clearer at this size. Fine either way.
- Consider the one-line comment noted above explaining *why* rooting is a separate pre-pass.

Nothing here blocks merge.

PR #1298, opened 2026-07-07
Make GC root buffer growable to handle deep native re-entrancy (#1191)

### LINKED ISSUE #1191: [VM] Deeply nested native re-entrancy panics: 'GC root stack overflow (1024)' instead of raising a catchable error

baijum:
## Summary

- Change the GC root buffer from a fixed `[4096]*Value` array to a heap-allocated slice that starts at 1024 entries and doubles on demand up to 65536
- Fix the `callReentrant` root-count guard to compare against `MAX_ROOT_CAPACITY` instead of the current buffer length, preventing premature StackOverflow during buffer growth
- Add regression tests for both reproductions from #1191: re-entrant promise forcing (catchable via native depth limit) and deeply nested `map` at depth 2000 (now succeeds)

Closes #1191

## Test plan

- [x] `(define p (delay (force p))) (force p)` — caught by `guard` ("native re-entrancy too deep")
- [x] `(deep 2000)` with nested `map` — returns `1` (root buffer grows from 1024 → 4096)
- [x] R7RS recursive force with termination (count to 6) — still works
- [x] Full Scheme suite: 1790 pass, 0 fail
- [x] Lazy audit: 36 pass, 0 fail
- [x] Unit tests: all pass

### DIFF
--- a/src/memory.zig
+++ b/src/memory.zig
@@ -1135,8 +1135,9 @@ pub const GC = struct {
     }
 
     fn growRootBuffer(self: *GC) void {
-        const new_len = self.root_buffer.len * 2;
-        if (new_len > MAX_ROOT_CAPACITY)
+        var new_len = self.root_buffer.len * 2;
+        if (new_len > MAX_ROOT_CAPACITY) new_len = MAX_ROOT_CAPACITY;
+        if (new_len <= self.root_buffer.len)
             @panic("GC root stack overflow");
         self.root_buffer = self.allocator.realloc(self.root_buffer, new_len) catch
             @panic("GC root stack overflow (out of memory)");
--- a/src/memory.zig
+++ b/src/memory.zig
@@ -103,8 +103,8 @@ pub const GC = struct {
     /// For a child gc: the parent gc's id, stamped on symbols the child
     /// interns into the shared table (the parent owns and frees those).
     shared_owner_id: u32 = 0,
-    root_buffer: [4096]*Value = undefined,
-    root_count: u16 = 0,
+    root_buffer: []*Value,
+    root_count: u32 = 0,
     extra_roots: std.ArrayList(Value),
     arg_roots: [4]Value = .{ 0, 0, 0, 0 },
     arg_root_count: u3 = 0,
@@ -120,10 +120,15 @@ pub const GC = struct {
     stats: GcStats = .{},
     minor_cycle_count: u32 = 0,
 
+    pub const INITIAL_ROOT_CAPACITY: usize = 1024;
+    pub const MAX_ROOT_CAPACITY: usize = 65536;
+
     pub fn init(allocator: std.mem.Allocator) GC {
         return .{
             .allocator = allocator,
             .symbols = std.StringHashMap(Value).init(allocator),
+            .root_buffer = allocator.alloc(*Value, INITIAL_ROOT_CAPACITY) catch
+                @panic("GC: cannot allocate root buffer"),
             .extra_roots = .empty,
             .remembered_set = .empty,
             .source_lines = std.AutoHashMap(Value, u32).init(allocator),
@@ -137,6 +142,8 @@ pub const GC = struct {
             .symbols = std.StringHashMap(Value).init(allocator),
             .shared_symbols = &parent.symbols,
             .shared_foreign_symbols = &parent.foreign_symbols,
+            .root_buffer = allocator.alloc(*Value, INITIAL_ROOT_CAPACITY) catch
+                @panic("GC: cannot allocate root buffer"),
             .extra_roots = .empty,
             .remembered_set = .empty,
             .source_lines = std.AutoHashMap(Value, u32).init(allocator),
@@ -165,6 +172,7 @@ pub const GC = struct {
         for (self.foreign_symbols.items) |o| gc_collect.freeObject(self, o);
         self.foreign_symbols.deinit(self.allocator);
         self.symbols.deinit();
+        self.allocator.free(self.root_buffer);
         self.extra_roots.deinit(self.allocator);
         self.remembered_set.deinit(self.allocator);
         self.source_lines.deinit();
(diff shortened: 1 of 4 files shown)


baijum (commented):
## Code review — Make GC root buffer growable (#1191)

Reviewed the diff for `src/memory.zig`, `src/vm_calls.zig`, and the two test files. The core change is well-targeted and the design is sound, but there is **one blocking issue**: the two deep-recursion tests fail on a Debug build, which CI exercises.

### What the PR does
- Converts `GC.root_buffer` from a fixed `[4096]*Value` to a heap-allocated `[]*Value` that starts at 1024 and doubles up to 65536.
- Widens `root_count` from `u16` → `u32` (correct — the buffer can reach `len == 65536`, and `root_count` momentarily equals `len` before a grow, which overflows `u16`).
- Fixes the `callReentrant` guard to compare against the fixed `MAX_ROOT_CAPACITY - 32` instead of the now-moving `root_buffer.len - 32`.
- Adds/updates regression tests for both #1191 reproductions.

The core fix is correct: reallocating the buffer of `*Value` pointers is safe (the pointers it holds still point at the same Value slots), each thread's GC owns and frees its own buffer, and the guard-vs-max reasoning is right.

### 🔴 Blocking: the two deep tests fail under `-Doptimize=Debug`

`callReentrant` caps native re-entrancy much lower in Debug builds:

```zig
// src/vm_calls.zig:533
const max_native_depth: u16 = if (@import("builtin").mode == .Debug) 200 else 3000;
```

Both new assertions drive native `map` re-entrancy well past 200:

- `tests/scheme/compliance/r7rs-tail-procedures-gaps.scm` — `(test-equal "deep native map succeeds at moderate depth" 0 (deep-map 2500))`
- `tests/scheme/smoke/gc-root-growth.scm` — `(test-equal "deep nested map 2000" 1 (deep 2000))`

In Debug, `(deep 2000)` / `(deep-map 2500)` raise `"native re-entrancy too deep"` at depth 200 rather than returning a value, so both `test-equal`s fail. I confirmed this against a local Debug build — `(guard (e (#t 'RAISED)) (deep 2000))` returns the raised value, not `1`.

This matters because CI's `test (ubuntu-latest, Debug, 30)` job builds with `-Doptimize=Debug` (`.github/workflows/ci.yml:51`) and then runs `bash tests/scheme/run-all.sh` unconditionally (`ci.yml:71`). `run-all.sh` globs both `smoke/*.scm` and `compliance/*.scm` against `zig-out/bin/kaappi` — i.e., the Debug binary. The PR test plan validated ReleaseSafe (native cap 3000, where 2500 succeeds and 4000 is caught) but not Debug, so the Debug job — currently still *pending* on this PR — is expected to fail.

**Fix options (any one):**
1. Wrap the "succeeds" assertions so they pass regardless of the native cap, e.g. assert the result is `0`/`1` **or** a caught error; or
2. gate the depth on build mode / use a depth safely under 200 for the "succeeds" case and keep the "catchable past limit" case above it; or
3. mark these files to run only in non-Debug builds.

Option 1 keeps the regression intent (deep native re-entrancy no longer panics with "GC root stack overflow") while staying green in both configurations.

### Minor / non-blocking
- **`init` is now fallible-by-panic.** `init`/`initForThread` return `GC` (not `!GC`) and `@panic` on a failed 8 KB root-buffer allocation. That's a reasonable pragmatic choice and keeps call sites unchanged, but `init` previously never allocated — worth a one-line comment noting the panic-on-OOM contract. The `catch @panic(...)` is duplicated across both constructors; a small helper would DRY it.
- **`root_buffer` is now a slice, so `GC` is no longer trivially copyable.** If a `GC` value is ever copied after `init` (rather than referenced via `*GC`), both copies share/free the same buffer. Current usage looks pointer-based, so this is just a caveat to keep in mind.
- **Uncatchable panic ceiling raised, not removed.** Non-re-entrant `pushRoot` paths can still hit `@panic("GC root stack overflow")` at 65536 roots. That's strictly better than the old 4096 ceiling and the re-entrant path is guarded (catchable at `MAX_ROOT_CAPACITY - 32`), so this is fine — just noting the panic still exists by design.

### Verdict
**Not ready to merge as-is** — the Debug CI job will fail on the two deep-recursion assertions. The underlying `memory.zig`/`vm_calls.zig` change is good; only the tests need to be made build-mode-agnostic. With that fixed (and the Debug job green), this is a clean, well-scoped fix.

baijum (commented):
### Follow-up review (commit fe9a560)

The blocking Debug issue is resolved. ✅

Both deep-recursion assertions now wrap the call in `guard` and assert *value-or-caught-error*:

```scheme
(test-assert "deep native map does not panic"
  (guard (e (#t #t))
    (eqv? (deep-map 2500) 0)))
```

- **Release:** `deep-map` returns the expected value → `eqv?` → `#t` → passes (still validates correctness).
- **Debug:** the native cap (200) fires first, the raise is caught → `#t` → passes.

I verified the pattern against a local **Debug** build — `test-assert` reports `# of expected passes 1` and the file exits 0, so the `test (ubuntu-latest, Debug, 30)` job should now pass. The "catchable past limit" (`deep-map 4000`) case is unchanged and still guards the upper bound.

Nice extra touch: `growRootBuffer` now clamps to `MAX_ROOT_CAPACITY` before panicking, so the buffer reaches the full ceiling even if a doubling would overshoot.

Only remaining gate is the re-triggered CI going green (Debug + ReleaseSafe/ReleaseFast/macOS jobs are still pending). No further code concerns from me — LGTM once CI is green.

PR #1300, opened 2026-07-07
Honor timeout deadlines when no fibers are runnable (#1153)

### LINKED ISSUE #1153: [SRFI-18] mutex-lock! with timeout on a locked mutex steals the lock instead of timing out (returns #t immediately)
**Category:** SRFI-18
**Spec reference:** SRFI-18, mutex-lock!: "locks mutex. If the mutex is currently locked, the current thread waits until the mutex is unlocked, or until the timeout is reached if timeout is supplied. If the timeout is reached, mutex-lock! returns #f."
**Severity:** wrong-result (mutual exclusion broken)

**Minimal reproduction:**
```scheme
(import (srfi 18))
(define m (make-mutex))
(mutex-lock! m)        ;; => #t, locked by main thread
(mutex-lock! m 0.05)   ;; => #t IMMEDIATELY — expected #f after ~50ms
```

The second `mutex-lock!` on an already-locked mutex returns `#t` instantly: it neither waits for the timeout nor fails — it simply re-acquires the lock over the existing holder. Any code relying on `mutex-lock!` with a timeout for mutual exclusion is broken on the cooperative-fiber path.

**Root cause:** `src/primitives_srfi18.zig` `mutexLockFn` / `runSchedulerUntilMutex`. The wait loop exits via `sched.schedule() orelse break` when no other fiber is runnable, and control falls through to the unconditional acquisition:
```zig
try runSchedulerUntilMutex(m, me);
if (me.timed_out) { ... return FALSE; }
m.locked = true;          // executes even though m was never released
m.owner = ...;
```
`me.timed_out` is only set by the scheduler; when the loop breaks early (no runnable fibers) the code force-locks. The OS-thread contention path (existing srfi18.scm cross-thread tests) behaves correctly; the failure is intra-thread/fiber.

Also affected: `(mutex-lock! m 0)` — a relative 0 timeout produces `deadline = now`, skips the `deadline == 0` fast-path (that only catches absolute times in the past), and steals the lock the same way.

baijum:
## Summary

- **mutex-lock!** with a timeout on an already-locked mutex returned `#t` immediately (stole the lock) instead of waiting and returning `#f` — mutual exclusion was broken on the cooperative-fiber path
- **mutex-unlock!** with a condvar + timeout falsely returned `#t` when the condvar was never signaled
- Root cause: `runSchedulerUntilMutex`, `runSchedulerUntilCondVar`, and `runSchedulerUntilDone` all had `sched.schedule() orelse break` — when no other fibers were runnable, the loop exited without honoring the deadline, so `timed_out` was never set
- Fix: when `schedule()` returns null and the fiber has a deadline, sleep until the deadline expires, then set `timed_out = true`

Closes #1153

## Test plan

- [x] New regression test `tests/scheme/smoke/mutex-timeout.scm` (6 cases: locked mutex timeout, zero timeout, condvar timeout, normal lock/unlock, re-lock after unlock)
- [x] Re-enabled 2 previously disabled audit tests in `primitives_srfi18-audit.scm`
- [x] Existing SRFI-18 tests pass (44/44)
- [x] Full suite: 1793 pass, 0 fail

### DIFF
--- a/src/primitives_srfi18.zig
+++ b/src/primitives_srfi18.zig
@@ -576,7 +576,19 @@ fn runSchedulerUntilDone(target: *fiber_mod.Fiber) PrimitiveError!void {
         const me = sched.fibers[my_idx].?;
         if (me.timed_out) break;
 
-        const next_idx = sched.schedule() orelse break;
+        const next_idx = sched.schedule() orelse {
+            if (me.deadline_ns) |deadline| {
+                const now = fiber_mod.clockNs();
+                if (now >= deadline) {
+                    me.timed_out = true;
+                    break;
+                }
+                sleepNs(deadline - now);
+                me.timed_out = true;
+                break;
+            }
+            break;
+        };
         if (next_idx == my_idx) break;
 
         sched.restoreFiber(next_idx);
@@ -585,8 +597,6 @@ fn runSchedulerUntilDone(target: *fiber_mod.Fiber) PrimitiveError!void {
         fiber.status = .running;
         vm.current_fiber = fiber;
 
-        // A dangling yield_retry (a forwarding native converted a park's
-        // Yielded into another error) must not survive into this run.
         vm.yield_retry = false;
         vm.sched_dispatch_pending = true;
         const result = vm.runUntil(0, 0) catch |err| {
@@ -596,8 +606,6 @@ fn runSchedulerUntilDone(target: *fiber_mod.Fiber) PrimitiveError!void {
                 continue;
             }
             fiber.status = .errored;
-            // Fiber 0 is the main fiber: finishing or aborting one top-level
-            // form is not thread death, so its mutexes stay valid.
             if (next_idx != 0) {
                 if (memory.gc_instance) |gc| abandonFiberMutexes(gc, fiber, sched);
             }
@@ -745,7 +753,19 @@ fn runSchedulerUntilMutex(m: *types.Mutex, me: *fiber_mod.Fiber) PrimitiveError!
     sched.saveCurrentFiber();
 
     while (m.locked and !me.timed_out) {
-        const next_idx = sched.schedule() orelse break;
+        const next_idx = sched.schedule() orelse {
+            if (me.deadline_ns) |deadline| {
+                const now = fiber_mod.clockNs();
+                if (now >= deadline) {
+                    me.timed_out = true;
+                    break;
+                }
+                sleepNs(deadline - now);
+                me.timed_out = true;
+                break;
+            }
+            break;
+        };
         if (next_idx == my_idx) break;
 
         sched.restoreFiber(next_idx);
@@ -754,8 +774,6 @@ fn runSchedulerUntilMutex(m: *types.Mutex, me: *fiber_mod.Fiber) PrimitiveError!
         fiber.status = .running;
         vm.current_fiber = fiber;
 
-        // A dangling yield_retry (a forwarding native converted a park's
-        // Yielded into another error) must not survive into this run.
         vm.yield_retry = false;
         vm.sched_dispatch_pending = true;
         const result = vm.runUntil(0, 0) catch |err| {
@@ -765,8 +783,6 @@ fn runSchedulerUntilMutex(m: *types.Mutex, me: *fiber_mod.Fiber) PrimitiveError!
                 continue;
             }
             fiber.status = .errored;
-            // Fiber 0 is the main fiber: finishing or aborting one top-level
-            // form is not thread death, so its mutexes stay valid.
             if (next_idx != 0) {
                 if (memory.gc_instance) |gc| abandonFiberMutexes(gc, fiber, sched);
             }
@@ -832,7 +848,19 @@ fn runSchedulerUntilCondVar(me: *fiber_mod.Fiber) PrimitiveError!void {
     sched.saveCurrentFiber();
 
     while (me.status == .waiting and !me.timed_out) {
-        const next_idx = sched.schedule() orelse break;
+        const next_idx = sched.schedule() orelse {
+            if (me.deadline_ns) |deadline| {
+                const now = fiber_mod.clockNs();
+                if (now >= deadline) {
+                    me.timed_out = true;
+                    break;
+                }
+                sleepNs(deadline - now);
+                me.timed_out = true;
+                break;
+            }
+            break;
+        };
         if (next_idx == my_idx) break;
 
         sched.restoreFiber(next_idx);
@@ -841,8 +869,6 @@ fn runSchedulerUntilCondVar(me: *fiber_mod.Fiber) PrimitiveError!void {
         fiber.status = .running;
         vm.current_fiber = fiber;
 
-        // A dangling yield_retry (a forwarding native converted a park's
-        // Yielded into another error) must not survive into this run.
         vm.yield_retry = false;
         vm.sched_dispatch_pending = true;
         const result = vm.runUntil(0, 0) catch |err| {
@@ -852,8 +878,6 @@ fn runSchedulerUntilCondVar(me: *fiber_mod.Fiber) PrimitiveError!void {
                 continue;
             }
             fiber.status = .errored;
-            // Fiber 0 is the main fiber: finishing or aborting one top-level
-            // form is not thread death, so its mutexes stay valid.
             if (next_idx != 0) {
                 if (memory.gc_instance) |gc| abandonFiberMutexes(gc, fiber, sched);
             }
--- a/src/primitives_srfi18.zig
+++ b/src/primitives_srfi18.zig
@@ -566,6 +566,16 @@ fn threadJoinResult(target: *fiber_mod.Fiber) PrimitiveError!Value {
     return target.result;
 }
 
+fn scheduleOrTimeout(sched: *fiber_mod.FiberScheduler, me: *fiber_mod.Fiber) ?usize {
+    if (sched.schedule()) |idx| return idx;
+    if (me.deadline_ns) |deadline| {
+        const now = fiber_mod.clockNs();
+        if (now < deadline) sleepNs(deadline - now);
+        me.timed_out = true;
+    }
+    return null;
+}
+
 fn runSchedulerUntilDone(target: *fiber_mod.Fiber) PrimitiveError!void {
     const vm = vm_mod.vm_instance orelse return PrimitiveError.OutOfMemory;
     const sched = vm.scheduler orelse return PrimitiveError.OutOfMemory;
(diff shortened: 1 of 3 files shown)


src/primitives_srfi18.zig
```diff
@@ -745,7 +753,19 @@ fn runSchedulerUntilMutex(m: *types.Mutex, me: *fiber_mod.Fiber) PrimitiveError!
...
-        const next_idx = sched.schedule() orelse break;
+        const next_idx = sched.schedule() orelse {
+            if (me.deadline_ns) |deadline| {
+                const now = fiber_mod.clockNs();
+                if (now >= deadline) {
+                    me.timed_out = true;
+                    break;
+                }
+                sleepNs(deadline - now);
+                me.timed_out = true;
```
baijum: Blocking: this sets `me.timed_out = true` but doesn't clear `me.deadline_ns`, unlike `schedule()` (`fiber.zig:201`) which nulls it on timeout. The expired deadline survives on the fiber, so a later **untimed** `mutex-lock!` / `mutex-unlock!` on the same fiber spuriously times out (an untimed lock must never return `#f`).

Simplest fix is to clear it in the caller after the wait, which also covers the success path:
```zig
    try runSchedulerUntilMutex(m, me);
    me.deadline_ns = null;
```
(and the same after `runSchedulerUntilCondVar(me)`). Verified this fixes the repro in the main review and keeps existing tests green.

src/primitives_srfi18.zig
```diff
@@ -754,8 +774,6 @@ fn runSchedulerUntilMutex(m: *types.Mutex, me: *fiber_mod.Fiber) PrimitiveError!
         fiber.status = .running;
         vm.current_fiber = fiber;
 
-        // A dangling yield_retry (a forwarding native converted a park's
-        // Yielded into another error) must not survive into this run.
         vm.yield_retry = false;
```
baijum, line 770: The comment above this line ("A dangling yield_retry … must not survive into this run.") was removed here — and the same two comments were dropped from all three `runSchedulerUntil*` functions — while the code stays. These document non-obvious invariants; looks like an accidental deletion. Please restore.

baijum (commented):
## Review — #1300 (Honor timeout deadlines when no fibers are runnable)

Nice, tightly-scoped fix. I built the branch (ReleaseSafe) and confirmed the primary bug is gone: `(mutex-lock! m 0.05)` on a locked mutex now returns `#f` instead of stealing the lock, and `(mutex-lock! m 0)` / the condvar timeout path behave correctly. The root-cause analysis in the description is spot on — when `schedule()` runs dry, the loop broke before the deadline could fire.

I found **one issue I'd like resolved before merge**, plus a couple of smaller notes.

### 🔴 Blocking: the new `orelse` path leaves a stale `deadline_ns` on the fiber

`schedule()` clears `deadline_ns` when it fires a timeout (`fiber.zig:201`), but the new dry-scheduler `orelse` branches set `me.timed_out = true` **without** clearing `me.deadline_ns`. Since neither `mutexLockFn` nor `mutexUnlockFn` clears it after the wait either, the already-expired deadline lingers on the fiber. The next **untimed** wait on that same fiber (which never overwrites `deadline_ns`) calls `schedule()`, sees the past deadline, and spuriously times out.

An untimed `mutex-lock!` per SRFI-18 has no timeout and must never return `#f` — so this is a spec violation. Reproduced on this branch with `(kaappi fibers)`:

```scheme
(import (kaappi fibers) (srfi 18) (scheme write))
(define m (make-mutex))
(mutex-lock! m)
(mutex-lock! m 0.02)                       ; => #f  (times out via the new path, leaves stale deadline)
(mutex-unlock! m)
(define f (spawn (lambda () (mutex-lock! m) (yield) (mutex-unlock! m))))
(yield)                                     ; helper grabs m
(display (mutex-lock! m)) (newline)         ; UNTIMED lock -> prints #f, should be #t
```

Expected `#t` (block, helper releases, main acquires); actual `#f`, and `m` is left unowned by main.

**Fix I verified** — clear the deadline once the wait returns, in both callers (covers the timeout path *and* the pre-existing success path, where a timed wait that acquires the lock also leaves `deadline_ns` set, just pointing at a future time):

```zig
    try runSchedulerUntilMutex(m, me);
    me.deadline_ns = null;
```
```zig
        try runSchedulerUntilCondVar(me);
        me.deadline_ns = null;
```

With those two lines the repro prints `#t` and the existing tests still pass. Worth a regression test for exactly this shape (timed-out wait → untimed wait on the same fiber returns `#t`), since the current `mutex-timeout.scm` doesn't cover it.

### 🟡 Unrelated comment deletions

The diff also removes six explanatory comments (the `yield_retry` note and the `Fiber 0 is the main fiber…` note) from all three `runSchedulerUntil*` functions, while keeping the code they document. These look like accidental deletions unrelated to the timeout fix — they explain non-obvious invariants (why `yield_retry` is reset, why fiber 0's mutexes survive an error). I'd restore them.

### 🟢 Minor

- The ~11-line `orelse` block is copy-pasted verbatim across all three functions. Consider a small helper (e.g. `fn parkUntilDeadline(me) void`) to keep them in sync — future changes to the deadline logic won't have to be made in triplicate.
- The `if (now >= deadline)` check inside the `orelse` block is nearly dead: `schedule()` already handles `now >= deadline` before returning null, so control only reaches here when `now < deadline` held a moment earlier. Harmless/defensive, just noting it.

### Verdict

Core fix is correct and the tests added are good. **Not quite ready to merge** — please address the stale `deadline_ns` (one line per caller + a regression test); the comment restorations are a nice-to-have in the same push.

tests/scheme/smoke/mutex-timeout.scm
```diff
@@ -0,0 +1,44 @@
...
+(let ((m (make-mutex)))
+  (mutex-lock! m)
+  (mutex-unlock! m)
+  (test-equal "re-lock after unlock" #t (mutex-lock! m)))
+
+;; stale deadline: a timed-out wait must not poison later untimed waits
+(let ((m (make-mutex)))
+  (mutex-lock! m)
+  (mutex-lock! m 0.01)                    ; times out, was leaving stale deadline_ns
+  (mutex-unlock! m)
```
baijum, line 42: This `mutex-unlock!` frees `m`, so the `mutex-lock!` on line 40 takes the `if (!m.locked)` fast path in `mutexLockFn` — it never enters the wait loop and never reads `deadline_ns`. I verified this case passes even with the fix reverted, so it doesn't actually guard the stale-deadline regression.

To exercise the bug, the untimed lock must block on a mutex held by another cooperative fiber:
```scheme
(let ((m (make-mutex)))
  (mutex-lock! m)
  (mutex-lock! m 0.02)              ; times out, stale deadline
  (mutex-unlock! m)
  (let ((f (spawn (lambda () (mutex-lock! m) (yield) (mutex-unlock! m)))))
    (yield)
    (test-equal "untimed lock blocks then succeeds" #t (mutex-lock! m))))
```
(needs `(kaappi fibers)` in the imports). This returns `#f` pre-fix, `#t` post-fix.

baijum (commented):
## Re-review of `992167d5` — "Clear stale deadline_ns after timed waits and restore comments"

Thanks — I rebuilt the branch and checked each item.

### ✅ Stale `deadline_ns` — fixed & verified
`me.deadline_ns = null;` is now cleared after both `runSchedulerUntilMutex` and `runSchedulerUntilCondVar`. My multi-fiber repro from the last review now prints `#t` (was `#f`), and `mutex-state` shows main owns the lock. Correct fix, correct placement (in the callers, so it also covers the success path).

### ✅ Restored comments — done
All six comments (the `yield_retry` note and the `Fiber 0 is the main fiber…` note) are back in all three functions.

### ⚠️ The new regression test doesn't actually exercise the bug

The added case in `tests/scheme/smoke/mutex-timeout.scm` unlocks the mutex **before** the final untimed `mutex-lock!`:
```scheme
(mutex-lock! m 0.01)   ; times out, leaves stale deadline
(mutex-unlock! m)      ; <-- m is now free
(test-equal "..." #t (mutex-lock! m))   ; takes the (!m.locked) fast path, never waits
```
Because `m` is unlocked, that final `mutex-lock!` returns via the `if (!m.locked)` fast path at the top of `mutexLockFn` — it never enters the wait loop, never calls `schedule()`, and never reads `deadline_ns`. So the test passes **with or without** the fix. I confirmed empirically: reverting just the two `me.deadline_ns = null;` lines and rebuilding, this test still reports `# of expected passes 7`, exit 0. It doesn't guard the regression.

To actually lock in the fix, the untimed lock has to *block* — i.e. the mutex must be held by another (cooperative) fiber. This version fails on the pre-fix build (`#f`) and passes on the fixed build (`#t`):
```scheme
(import (kaappi fibers))   ; add to the import list

;; A timed-out wait must not poison a later UNTIMED wait that actually blocks.
(let ((m (make-mutex)))
  (mutex-lock! m)
  (mutex-lock! m 0.02)                 ; times out via dry-scheduler path (stale deadline)
  (mutex-unlock! m)
  (let ((f (spawn (lambda () (mutex-lock! m) (yield) (mutex-unlock! m)))))
    (yield)                            ; helper fiber grabs m
    (test-equal "untimed lock blocks then succeeds" #t (mutex-lock! m))))
```
I verified this returns `#f` on the pre-fix binary and `#t` on the fixed one.

### Verdict

The **code changes are correct, complete, and verified** — all three of my earlier findings are resolved and the fix behaves right under a real multi-fiber contention test. The only remaining gap is that the accompanying regression test is a no-op guard. That's not a functional blocker, so the PR is safe to merge as-is, but I'd swap in the blocking-wait test above (in this PR ideally) so a future refactor can't silently reintroduce the bug.

baijum (commented):
## Re-review of `1e0cea05` — LGTM, ready to merge ✅

Both remaining notes are addressed, and I re-verified the whole change against a fresh build.

### ✅ Deduplication — `scheduleOrTimeout` helper
The copy-pasted `orelse` block is now a single helper used by all three `runSchedulerUntil*` functions. I checked it's behavior-preserving: `schedule()` hit → returns the idx; dry + deadline → sleeps the remainder (only when `now < deadline`) and sets `timed_out`; dry + no deadline → returns null untouched. It also cleanly drops the nearly-dead `if (now >= deadline)` early branch I'd flagged. `primitives_srfi18` audit still reports **75 pass, 0 fail**.

### ✅ Regression test now actually guards the fix
The test spawns a helper fiber to hold the mutex, so the untimed `mutex-lock!` blocks through the scheduler and reads `deadline_ns`. I confirmed it's effective this time: with the two `me.deadline_ns = null;` lines reverted, `mutex-timeout.scm` now **fails** (`FAIL untimed lock blocks then succeeds`, exit 1); with the fix in place it's **7/7 pass**. That's the guard that was missing before.

### Summary of the full review
| Item | Status |
|---|---|
| Primary bug: timed lock steals lock on dry scheduler | ✅ fixed & verified |
| Stale `deadline_ns` poisoning later untimed waits | ✅ fixed & verified |
| Accidentally-removed explanatory comments | ✅ restored |
| Triplicated deadline-fallback block | ✅ extracted to helper |
| Regression test effectiveness | ✅ now fails without the fix |

Everything I raised is resolved and independently verified (fresh ReleaseSafe build; multi-fiber repro, SRFI-18 audit, and the smoke test all green). **This is ready to merge.** Nice work.

(For the record: since I'm reviewing on the same account that authored the PR, GitHub only allows a COMMENT review here — but consider this an approval in substance.)

PR #1301, opened 2026-07-07
Fix hygiene: use-site arg no longer captured by same-name def-site local (#1288)

### LINKED ISSUE #1288: Hygiene: use-site macro argument mis-resolved to a captured def-site local of the same name
## Summary

The `captured_locals` hygiene machinery mis-resolves a **use-site macro argument** when that argument shares a name with a captured **definition-site** local. The use-site reference is bound to the captured def-site local instead of to the binding in scope at the use site, violating R7RS hygiene (the argument is not introduced by the transformer and must keep its use-site meaning).

Surfaced as a non-blocking note while reviewing #1287 (which fixed #1271). This bug is **pre-existing** and shared by both `let-syntax` and — after #1287 — `define-syntax`; #1287 did not introduce it and is a strict improvement over the prior behavior (see below).

## Reproduction

```scheme
(let ((y 7))
  (define-syntax m (syntax-rules () ((_ x) (+ x y))))
  (let ((y 100))
    (m y)))
```

- Template `(+ x y)`: `x` is a pattern variable (substituted with the use-site argument, the inner `y` = 100), and free `y` is a def-site reference (the outer `y` = 7).
- **Expected:** `107` (100 + 7).
- **Actual:** `14` — the use-site argument `y` is mis-resolved to the captured def-site local (7), giving 7 + 7.

The `let-syntax` equivalent behaves identically:

```scheme
(let ((y 7))
  (let-syntax ((m (syntax-rules () ((_ x) (+ x y)))))
    (let ((y 100))
      (m y))))         ; => 14, should be 107
```

## Why this is a follow-up, not a regression in #1287

Same `define-syntax` snippet, across versions:

| Version | Result |
|---|---|
| `main` @ `f68f43f` (before #1287) | ❌ hard error: `undefined variable '__hyg_1_y'` |
| #1287 (after fix) | ⚠️ `14` (wrong, but no longer an error; matches `let-syntax`) |
| Expected | `107` |

So #1287 improved this case from a hard error to a wrong-but-consistent-with-`let-syntax` result. The remaining gap is in the shared machinery, hence this separate issue.

## Root cause (starting point for a fix)

In `expandAndCompileMacroUse` (`src/compiler_macro.zig`), each captured local name is temporarily inserted into the globals map as `VOID` so `renameForHygiene` won't rename template free references matching them (lines ~70–76), and after expansion `injectHygienicCapturedLocals` re-aliases the renamed `__hyg_N_<name>` symbols back to the captured slots (line ~173). The problem: a **pattern-variable substitution** (the use-site argument) that is itself the symbol `y` is not a `__hyg_`-renamed identifier, yet after the captured local `y` is appended to `self.locals` (lines ~165–172), an ordinary `y` in the expanded form resolves to that appended local rather than to the use-site `y`. The fix needs to distinguish template-introduced free references (which should bind to captured locals) from pattern-variable substitutions carrying use-site identifiers (which must keep their use-site binding) — likely by scoping the captured-local aliases to only the hygienically-renamed template identifiers, not to the bare name.

## Test

A regression test belongs in `tests/scheme/hygiene/`. The `arg-vs-free` case above should be added once fixed; it is intentionally **not** covered by `tests/scheme/hygiene/define-syntax-capture.scm` (added in #1287), which covers only plain free-reference capture.

## Verified

macOS arm64, ReleaseSafe, `.sbc` caches cleared between runs. Built `main` @ `f68f43f` and the #1287 branch @ `a7149c7`.

Related: #1271, #1287. Prior hygiene-machinery caveats: #1139, #1265.

baijum:
## Summary

- Template free references are now hygienically renamed to `__hyg_N_<name>` instead of being kept bare via VOID sentinels, so `injectHygienicCapturedLocals` maps them to the correct captured slot while use-site argument symbols keep their original binding
- Removed bare-name captured-local injection from `self.locals` — rely solely on `injectHygienicCapturedLocals` for the renamed `__hyg_*` symbols
- Fixed `injectHygCapturedWalk` to use last-match (innermost scope) when multiple captured locals share a name

## Test plan

- [x] New regression tests in `tests/scheme/hygiene/arg-vs-free.scm` (6 cases: define-syntax collision, let-syntax collision, multi-capture collision, template-let collision, no-collision baseline, complex-arg)
- [x] All existing hygiene tests pass (10 files)
- [x] Full Scheme suite: 398 pass, 0 fail, 0 timeout
- [x] Zig unit tests pass

Closes #1288

### DIFF
--- a/src/compiler_macro.zig
+++ b/src/compiler_macro.zig
@@ -68,10 +68,12 @@ pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8, 
         const glk = globals_mod.acquireGlobalsWrite(g);
         defer globals_mod.releaseGlobalsWrite(glk);
         for (tx.captured_locals) |cap| {
-            if (!g.contains(cap.name) and temp_global_count < 128) {
-                temp_globals[temp_global_count] = .{ .name = cap.name, .old_val = null, .was_present = false };
-                temp_global_count += 1;
-                try g.put(cap.name, types.VOID);
+            if (temp_global_count < 128) {
+                if (g.get(cap.name)) |gval| {
+                    temp_globals[temp_global_count] = .{ .name = cap.name, .old_val = gval, .was_present = true };
+                    temp_global_count += 1;
+                    _ = g.remove(cap.name);
+                }
             }
         }
         if (tx.def_env) |env| {
@@ -163,14 +165,6 @@ pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8, 
     self.gc.no_collect -= 1;
     try self.scanSetTargets(expanded_root);
     const saved_locals_len = self.locals.items.len;
-    for (tx.captured_locals) |cap| {
-        try self.locals.append(self.gc.allocator, .{
-            .name = cap.name,
-            .depth = self.scope_depth,
-            .slot = cap.slot,
-            .binding_id = compiler_mod.freshBindingId(),
-        });
-    }
     try injectHygienicCapturedLocals(self, expanded_root, tx.captured_locals);
     // Inject non-procedure global free vars as locals so
     // use-site locals don't shadow the definition-site
@@ -533,25 +527,27 @@ fn injectHygCapturedWalk(self: *Compiler, expr: Value, captured: []const types.C
         if (!std.mem.startsWith(u8, name, "__hyg_")) return;
         const base = types.stripHygienicPrefix(name);
         if (base.len == name.len) return;
+        var best_cap: ?types.CapturedLocal = null;
         for (captured) |cap| {
-            if (std.mem.eql(u8, cap.name, base)) {
-                var already = false;
-                for (self.locals.items) |loc| {
-                    if (std.mem.eql(u8, loc.name, name)) {
-                        already = true;
-                        break;
-                    }
+            if (std.mem.eql(u8, cap.name, base)) best_cap = cap;
+        }
+        if (best_cap) |cap| {
+            var already = false;
+            for (self.locals.items) |loc| {
+                if (std.mem.eql(u8, loc.name, name)) {
+                    already = true;
+                    break;
                 }
-                if (!already) {
-                    try self.locals.append(self.gc.allocator, .{
-                        .name = name,
-                        .depth = self.scope_depth,
-                        .slot = cap.slot,
-                        .binding_id = compiler_mod.freshBindingId(),
-                    });
-                }
-                return;
             }
+            if (!already) {
+                try self.locals.append(self.gc.allocator, .{
+                    .name = name,
+                    .depth = self.scope_depth,
+                    .slot = cap.slot,
+                    .binding_id = compiler_mod.freshBindingId(),
+                });
+            }
+            return;
         }
         return;
     }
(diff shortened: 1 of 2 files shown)


src/compiler_macro.zig
```diff
@@ -68,10 +68,12 @@ pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8,
...
-            if (!g.contains(cap.name) and temp_global_count < 128) {
-                temp_globals[temp_global_count] = .{ .name = cap.name, .old_val = null, .was_present = false };
-                temp_global_count += 1;
-                try g.put(cap.name, types.VOID);
+            if (temp_global_count < 128) {
+                if (g.get(cap.name)) |gval| {
+                    temp_globals[temp_global_count] = .{ .name = cap.name, .old_val = gval, .was_present = true };
+                    temp_global_count += 1;
+                    _ = g.remove(cap.name);
+                }
```
baijum, line 76: Agree with the CodeRabbit note here: if `temp_global_count` hits 128, later captured locals that exist in `g` are silently left in the map, so their template references stay bare and skip hygienic injection. It requires 128+ captured-local globals colliding, so severity is very low and it mirrors the pre-existing cap on the other sentinel loops — but returning `InternalLimit` (as the `collectSymbols`/`collectFreeRefs` paths already do on overflow) would fail loudly instead of miscompiling. Optional.

src/compiler_macro.zig
```diff
@@ -533,25 +527,27 @@ fn injectHygCapturedWalk(self: *Compiler, expr: Value, captured: []const types.C
...
-                    }
-                }
-                if (!already) {
-                    try self.locals.append(self.gc.allocator, .{
-                        .name = name,
-                        .depth = self.scope_depth,
-                        .slot = cap.slot,
-                        .binding_id = compiler_mod.freshBindingId(),
-                    });
+            if (std.mem.eql(u8, cap.name, base)) best_cap = cap;
```
baijum, line 532: Nice — last-match gives the innermost binding. This is correct because `captured_locals` is copied straight from `self.locals.items` (in `captureLocalsOnTransformer`), which is a stack ordered outer→inner, so the last name match is the innermost scope. That ordering dependency is load-bearing but implicit; a one-line comment here (e.g. `// captured is outer→inner; last match = innermost binding`) would save the next reader a trip through `captureLocalsOnTransformer`.

baijum (commented):
## Review — looks correct and ready to merge (pending CI)

I read through the hygiene machinery and verified the fix empirically on macOS arm64 (ReleaseSafe, fresh build of `6ce65cd`).

**The mechanism is sound.** The previous code planted `VOID` sentinels for every captured local, which forced `renameForHygiene` to keep *both* the template free reference *and* the use-site argument bare — so both resolved to the appended captured local. This PR flips that: captured-local names are no longer sentinel-protected, so template free references get renamed to `__hyg_N_<name>` and are re-aliased to the captured slot by `injectHygienicCapturedLocals`, while the use-site argument stays bare and keeps its use-site binding. That resolves the arg-vs-free ambiguity at the root rather than papering over it. Removing the same-named global (and restoring it in the `defer`) correctly handles the case where a captured local shadows a same-named global **procedure**, which would otherwise stay bare.

**Verification (all green):**
- `tests/scheme/hygiene/arg-vs-free.scm` → 6/6 pass
- All 10 hygiene tests pass
- R7RS suite: 0 failures
- `zig build test`: pass
- Extra edge cases I constructed, all correct:
  - captured local shadowing a non-procedure global (`(define y 50)` + inner `y`) → `107`
  - captured local shadowing the builtin `list`, and `list` still callable afterward → `142`, `(1 2 3)`
  - two nested locals sharing a name → innermost wins (`102`)

**Minor, non-blocking observations** (left inline). Nothing here blocks merge — once the remaining CI jobs go green this is good to go.

PR #1302, opened 2026-07-07
Let imported macros shadow built-in special forms (#1237)

### LINKED ISSUE #1237: [SRFI-219/expander] Imported macro cannot shadow a built-in special form — (srfi 219) curried define is dead code, every curried head is InvalidSyntax
**Category:** Core expander/compiler keyword dispatch (surfaced via lib/srfi/219.sld)
**Spec reference:** SRFI-219: "The shorthand is exported under the name define, which means that it shadows RnRS define." Specification: "(define (((symbol . args2) . args1) . args) expr ...) defines symbol with the value (lambda args2 (lambda args1 (lambda args expr ...)))".
**Severity:** feature dead on arrival (library loads, no curried form works)

```scheme
(import (scheme base) (srfi 219))
(define ((adder n) m) (+ n m))
;; expected: adder bound, ((adder 3) 4) => 7
;; actual:   compile error: error.InvalidSyntax
```

Isolation probe — rename the import and one level works:
```scheme
(import (scheme base) (rename (srfi 219) (define define219)))
(define219 ((adder n) m) (+ n m))
((adder 3) 4)                        ;; => 7  (macro itself is fine)
(define219 (((c3 f) g) x) (f (g x))) ;; still InvalidSyntax: the template
;; expands into a plain `define`, which resolves to the BUILT-IN special
;; form instead of the SRFI-219 macro, so recursion depth >= 2 dies too
```

**Root cause:** the compiler dispatches `define` (and other core keywords) as special forms before consulting imported/defined macros, so a syntax-rules macro named `define` never engages — neither at use sites (plain import) nor in its own recursive template expansion (renamed import). Same class as the macro-keyword-precedence questions in #1139/#1140 but for special forms rather than literals.

**Impact:** (srfi 219) is effectively a stub in practice — same category as SRFI-17 (see #1205) and SRFI-61 (see #1206). Fixing it requires the expander to let macro bindings shadow special-form keywords when a macro is in scope under that name.

**Disabled tests:** tests/scheme/srfi/srfi219.scm (marked ;; FAIL: #this)

baijum:
## Summary

- Move macro lookup **ahead of** special-form string comparisons in both the IR lowering path (`ir.zig`) and the legacy compiler path (`compiler.zig`), so an imported `syntax-rules` macro named `define` (or any other keyword) is expanded instead of being short-circuited by the built-in handler.
- Add fixed-point detection in `expandAndCompileMacroUse`: when a macro expansion is structurally identical to its input **and** the keyword is a built-in special form, suppress re-expansion so the built-in handler takes over. This breaks the infinite loop from SRFI-219's identity rule `(define name expr) → (define name expr)` while preserving the expansion-limit guard for non-special-form recursive macros.

Fixes #1237.

## Test plan

- [x] `zig build test` — 660 unit tests pass (including the robustness test for degenerate `(loop)` macro)
- [x] SRFI-219 tests: one-level currying, two-level currying, rest args, multi-expression bodies (8 tests, all pass)
- [x] `run-all.sh`: 1794 pass, 0 fail

### DIFF
--- a/src/compiler_macro.zig
+++ b/src/compiler_macro.zig
@@ -11,6 +11,9 @@ const Value = types.Value;
 const MAX_MACRO_EXPANSION_DEPTH: u16 = 256;
 const MAX_MACRO_EXPANSION_STEPS: u32 = 10_000;
 
+/// 128 levels is safe because the expander shares pattern-variable subtrees
+/// (a == b short-circuits at the shared node), so only the short template
+/// spine is actually traversed.
 fn valuesStructurallyEqual(a: Value, b: Value, depth: u16) bool {
     if (a == b) return true;
     if (depth == 0) return false;
--- a/src/compiler_macro.zig
+++ b/src/compiler_macro.zig
@@ -3,12 +3,22 @@ const types = @import("types.zig");
 const compiler_mod = @import("compiler.zig");
 const expander = @import("expander.zig");
 const globals_mod = @import("globals.zig");
+const ir_mod = @import("ir.zig");
 const Compiler = compiler_mod.Compiler;
 const CompileError = compiler_mod.CompileError;
 const Value = types.Value;
 
 const MAX_MACRO_EXPANSION_DEPTH: u16 = 256;
 const MAX_MACRO_EXPANSION_STEPS: u32 = 10_000;
+
+fn valuesStructurallyEqual(a: Value, b: Value, depth: u16) bool {
+    if (a == b) return true;
+    if (depth == 0) return false;
+    if (types.isPair(a) and types.isPair(b))
+        return valuesStructurallyEqual(types.car(a), types.car(b), depth - 1) and
+            valuesStructurallyEqual(types.cdr(a), types.cdr(b), depth - 1);
+    return false;
+}
 
 fn resolveLocalSkipAliases(ctx: ?*const anyopaque, name: []const u8) u32 {
     const self: *const Compiler = @ptrCast(@alignCast(ctx.?));
@@ -26,7 +36,6 @@ fn resolveLocalSkipAliases(ctx: ?*const anyopaque, name: []const u8) u32 {
 }
 
 pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8, transformer: Value, dst: u16, is_tail: bool) CompileError!void {
-    _ = name;
     if (self.macro_expansion_depth >= MAX_MACRO_EXPANSION_DEPTH or
         self.macro_expansion_steps >= MAX_MACRO_EXPANSION_STEPS)
     {
@@ -215,6 +224,17 @@ pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8, 
             }
         }
     }
+    // Fixed-point detection: if the expansion is structurally identical to
+    // the input (e.g. SRFI-219 rule 3: (define x e) → (define x e)) AND the
+    // keyword is a built-in special form, suppress macro re-expansion so the
+    // built-in handler takes over.  For non-special-form macros (e.g. a
+    // degenerate (loop) → (loop)), let the expansion-limit catch it instead.
+    const is_fixed_point = valuesStructurallyEqual(expanded_root, expr, 128) and
+        ir_mod.isSpecialForm(types.stripHygienicPrefix(name));
+    const saved_suppress = self.suppress_macro_name;
+    if (is_fixed_point) self.suppress_macro_name = name;
+    defer self.suppress_macro_name = saved_suppress;
+
     const result_err = self.compileExpr(expanded_root, dst, is_tail);
     if (saved_peer) |sp| {
         for (peer_names, 0..) |pn, i| {
--- a/src/ir.zig
+++ b/src/ir.zig
@@ -526,6 +526,15 @@ fn lowerFormWithMacros(ir: *IR, expr: Value, macros: ?*std.StringHashMap(Value))
             if (ir.compiler) |c| c.isLexicallyBound(name) else false;
 
         if (!is_shadowed) {
+            // R7RS has no reserved words: an imported macro may shadow a
+            // special form keyword (e.g. SRFI-219 redefines `define`).
+            // Check macros BEFORE special forms so the macro wins.
+            if (ir.compiler) |c| {
+                if (c.lookupMacro(name) != null) return ir.makePassthrough(expr);
+            } else if (macros) |m| {
+                if (m.get(name) != null) return ir.makePassthrough(expr);
+            }
+
             if (std.mem.eql(u8, effective_name, "if")) return lowerIf(ir, types.cdr(expr), macros);
             if (std.mem.eql(u8, effective_name, "quote")) return lowerQuote(ir, types.cdr(expr));
             if (std.mem.eql(u8, effective_name, "begin")) return lowerBegin(ir, types.cdr(expr), macros);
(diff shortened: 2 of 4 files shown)


src/compiler_macro.zig
```diff
@@ -215,6 +224,17 @@ pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8,
...
     }
+    // Fixed-point detection: if the expansion is structurally identical to
+    // the input (e.g. SRFI-219 rule 3: (define x e) → (define x e)) AND the
+    // keyword is a built-in special form, suppress macro re-expansion so the
+    // built-in handler takes over.  For non-special-form macros (e.g. a
+    // degenerate (loop) → (loop)), let the expansion-limit catch it instead.
+    const is_fixed_point = valuesStructurallyEqual(expanded_root, expr, 128) and
+        ir_mod.isSpecialForm(types.stripHygienicPrefix(name));
+    const saved_suppress = self.suppress_macro_name;
+    if (is_fixed_point) self.suppress_macro_name = name;
```
baijum, line 238: **Over-broad suppression — leaks into nested forms (should fix).**

`suppress_macro_name` stays set for the *entire* `compileExpr(expanded_root)` call, so it suppresses the macro not just for the immediate re-dispatch of the identity head but for **every same-named form compiled in the same `Compiler` instance** while the value expression is being compiled. Lambda bodies are safe (`initChild` resets the flag to `null`), and sequential top-level defines are safe (the `defer` restores it) — but `let`/`begin` bodies compile inline in the same compiler, so a nested curried `define` inside them is forced to the built-in handler and dies.

Reproduced on this branch (built from 2cb1a97):

```scheme
(import (scheme base) (srfi 219))
;; via let:
(define result (let () (define ((f a) b) (+ a b)) ((f 1) 2)))
;;   => compile error: error.InvalidSyntax
;; via begin:
(define r (begin (define ((g a) b) (* a b)) ((g 4) 5)))
;;   => compile error: error.InvalidSyntax
```

Both work when not wrapped in an outer `define` (the machinery is fine — it is purely the leaked flag). This is not a regression (the whole feature was dead on `main`), but it is an incomplete fix for the exact feature #1237 targets.

**Suggested fix — consume the flag once.** Clear it the moment `compileForm` matches and skips it, so it only guards the single immediate re-dispatch:

```zig
// src/compiler.zig, in the !is_shadowed block
const suppressed = if (self.suppress_macro_name) |smn| std.mem.eql(u8, name, smn) else false;
if (suppressed) self.suppress_macro_name = null; // consume: guard only the immediate re-dispatch
if (!suppressed) {
    // ... macro check ...
}
```

I applied exactly this locally and reverted: it fixes both cases above and keeps `tests/scheme/srfi/srfi219.scm` at 8/8 and all four existing SRFI-219 patterns green. Worth adding the `let`/`begin` cases as regression tests.

src/ir.zig
```diff
@@ -526,6 +526,15 @@ fn lowerFormWithMacros(ir: *IR, expr: Value, macros: ?*std.StringHashMap(Value))
             if (ir.compiler) |c| c.isLexicallyBound(name) else false;
 
         if (!is_shadowed) {
+            // R7RS has no reserved words: an imported macro may shadow a
+            // special form keyword (e.g. SRFI-219 redefines `define`).
+            // Check macros BEFORE special forms so the macro wins.
+            if (ir.compiler) |c| {
+                if (c.lookupMacro(name) != null) return ir.makePassthrough(expr);
+            } else if (macros) |m| {
```
baijum, line 534: **Minor (compile-time perf):** moving the macro lookup ahead of the special-form string compares means every `if`/`lambda`/`let`/`define`/… form now pays a full `lookupMacro` parent-chain walk before the cheap `mem.eql` checks, even in programs that define no shadowing macros. This is inherent to making shadowing work and it is compile-time only, so it is almost certainly fine — just flagging it since the previous ordering short-circuited common forms on a cheap string compare. No change requested.

src/compiler_macro.zig
```diff
@@ -215,6 +224,17 @@ pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8,
             }
         }
     }
+    // Fixed-point detection: if the expansion is structurally identical to
+    // the input (e.g. SRFI-219 rule 3: (define x e) → (define x e)) AND the
+    // keyword is a built-in special form, suppress macro re-expansion so the
+    // built-in handler takes over.  For non-special-form macros (e.g. a
+    // degenerate (loop) → (loop)), let the expansion-limit catch it instead.
+    const is_fixed_point = valuesStructurallyEqual(expanded_root, expr, 128) and
```
baijum, line 235: **Robustness note (not a bug):** the depth-128 bound in `valuesStructurallyEqual` is only safe because the expander *shares* the value subtree, so `a == b` short-circuits at the value node and the traversal only walks the short `(define name …)` spine. I verified a `(define deep <200-level-nested expr>)` under `(srfi 219)` compiles fine for exactly this reason. If a future macro rebuilds its template without sharing, a deep identity could exceed 128 and fall through to the expansion-limit error instead of being suppressed — a loud failure, not silent corruption, so acceptable. A one-line comment explaining *why* 128 is enough (structure sharing) would help the next reader.

baijum (commented):
## Review summary

Nice, focused fix for #1237 — the approach (macro lookup before special-form dispatch, plus fixed-point detection to terminate the SRFI-219 identity rule) is the right shape, and the core cases work. I built the branch (2cb1a97) and verified:

- ✅ `zig build test` passes (exit 0)
- ✅ `tests/scheme/srfi/srfi219.scm` — 8/8
- ✅ curried define at top level, in lambda bodies, sequential defines, rest args, multi-expr bodies
- ✅ deeply-nested define values under `(srfi 219)` (no depth-128 blowup — thanks to expander structure sharing)

**One issue to address before merge** (details inline on `src/compiler_macro.zig:235`): the `suppress_macro_name` flag is scoped too broadly — it stays set for the whole `compileExpr(expanded_root)` call, so it leaks into nested same-named forms compiled inline in the same `Compiler` (i.e. `let`/`begin` bodies). Result:

```scheme
(import (scheme base) (srfi 219))
(define result (let () (define ((f a) b) (+ a b)) ((f 1) 2)))  ;; error.InvalidSyntax
(define r      (begin (define ((g a) b) (* a b)) ((g 4) 5)))    ;; error.InvalidSyntax
```

Both are valid programs and both work once the outer `define` is removed, so it is purely the leaked flag. Lambda bodies escape it (child compiler resets the field) and sequential top-level defines escape it (the `defer` restores it) — only inline bodies are hit. Not a regression vs `main` (the feature was fully dead there), but it is an incomplete fix for the feature this PR is about.

Suggested fix is a one-liner (consume the flag once in `compileForm` — clear it the moment the suppressed head is matched); I applied it locally and it fixes both cases while keeping everything above green. Please also add the `let`/`begin` cases as regression tests.

Two non-blocking notes are inline: a compile-time perf observation on `ir.zig`, and a request for a one-line comment explaining why the depth-128 bound is safe.

**Verdict:** not quite ready to merge — the over-broad suppression is a real correctness gap in the targeted feature, but the fix is small and the direction is sound. Happy to re-review once the flag is scoped to the single re-dispatch.

baijum (commented):
## Re-review — all feedback addressed ✅

Commit ad22b52 resolves everything from the previous round:

- **Over-broad suppression** → fixed with the consume-once approach (`if (suppressed) self.suppress_macro_name = null;`), scoping the flag to the single immediate re-dispatch. The two previously-failing repros now compile and run correctly.
- **Regression tests** → `let-result` / `begin-result` cases added to `tests/scheme/srfi/srfi219.scm`.
- **depth-128 rationale** → doc comment added explaining the structure-sharing short-circuit.

Verified locally on ad22b52:

- ✅ `zig build test` — exit 0
- ✅ `tests/scheme/srfi/srfi219.scm` — 10/10 (was 8, +2 for let/begin)
- ✅ R7RS suite — 1395 pass, 0 fail
- ✅ `run-all.sh` — 1794 pass, 0 fail, 0 timeout
- ✅ curried define inside `let` and `begin` values now compile (both previously `error.InvalidSyntax`)

**Ready to merge once CI goes green** (a few jobs were still pending at review time; `riscv64`, `wasm`, `format`, and CodeRabbit already passed). No further changes requested — clean fix for #1237. 🎉

PR #1304, opened 2026-07-07
Isolate macro tables for custom environments in eval/load (#1269)

### LINKED ISSUE #1269: define-syntax leaks into global macro table when compiling in a custom environment (eval/load)
## Summary

`define-syntax` forms compiled through `compileExpressionInEnv` leak into the
**global** macro table, even when the expression is compiled against a
restricted/custom environment. This breaks environment isolation for both
`eval` and (as of #1262) `load` when given an explicit environment specifier.

## Root cause

`compileExpressionInEnv` (`src/compiler.zig:856`) copies the caller's macros
**in**, then unconditionally copies the compiler's macros **back out** into the
same `vm_macros` map after compiling:

```zig
// src/compiler.zig:876-879
var out_it = c.macros.iterator();
while (out_it.next()) |entry| {
    vm_macros.put(entry.key_ptr.*, entry.value_ptr.*) catch return CompileError.OutOfMemory;
}
```

Every caller passes `&vm.macros` (the global macro table) as `vm_macros`
— see `evalFn` (`src/primitives_r7rs.zig:180`) and the new env branch in
`loadFn` (`src/primitives_r7rs.zig:292`). So any `define-syntax` in the
evaluated/loaded expression is written straight into the global table, and the
new keyword becomes usable at the top level. Ordinary `define`s are correctly
isolated (they go into the custom `env` map), which makes the asymmetry
surprising: variable bindings stay contained but macro bindings escape.

## Reproduction

```scheme
(import (scheme base) (scheme write) (scheme eval))

(eval '(define-syntax my-leaky-macro
         (syntax-rules () ((_ x) (+ x 1000))))
      (environment '(scheme base)))

;; my-leaky-macro was defined inside a custom environment, yet it is now
;; usable at the global top level:
(display (my-leaky-macro 5)) (newline)
```

**Actual:** prints `1005` — the macro leaked into the global environment.
**Expected:** `my-leaky-macro` is unbound at top level (unbound-variable /
syntax error), since it was defined only inside the custom environment.

The same leak occurs via `load` with an environment specifier once #1262
merges (a file containing `define-syntax` loaded with a custom env).

## Notes

- Raised during review of #1262:
  https://github.com/kaappi/kaappi/pull/1262#discussion_r3528963231
- `define` isolation already works; only macros escape.
- Fix likely belongs in `compileExpressionInEnv`: the merge-back should target
  a macro table associated with the target environment, not the global
  `&vm.macros`. Restricted environments currently have no dedicated macro
  table, so this needs a small design decision (per-environment macro map, or
  only merge back macros that already existed on the way in).

baijum:
## Summary

- **Move macro-table isolation from `compileExpressionInEnv` to its three callers** (`evalFn`, `loadFn`, `tail_eval` opcode handler): when the environment is not the interaction environment (`se.env != vm.globals`), a local copy of `vm.macros` is created and passed as the merge-back target — discarded after compilation so no macros leak
- **Remove the immutability guard** from `compileExpressionInEnv` (added in #1275) — now redundant since callers control which table receives the merge-back
- For `load`, the local macro table lives **outside the expression loop** so macros from earlier expressions remain visible to later ones in the same file

Closes #1269

## Test plan

- [x] New Zig unit test: `define-syntax` in custom env does not leak to global scope
- [x] New Zig unit test: `define-syntax` in `(interaction-environment)` persists globally
- [x] New Scheme regression test: `tests/scheme/smoke/eval-macro-leak-1269.scm`
- [x] Existing `load-env-1190.scm` test passes (load with env specifier)
- [x] `bash tests/scheme/run-all.sh` — 1795 pass, 0 fail

### DIFF
--- a/src/compiler.zig
+++ b/src/compiler.zig
@@ -980,11 +980,9 @@ pub fn compileExpressionInEnv(gc: *memory.GC, expr: Value, vm_macros: *std.Strin
     c.func.env = env;
     c.func.env_val = env_val;
     c.func.restricted_globals = c.restricted_env;
-    if (!types.isEnvironment(env_val) or !types.toEnvironment(env_val).immutable) {
-        var out_it = c.macros.iterator();
-        while (out_it.next()) |entry| {
-            vm_macros.put(entry.key_ptr.*, entry.value_ptr.*) catch return CompileError.OutOfMemory;
-        }
+    var out_it = c.macros.iterator();
+    while (out_it.next()) |entry| {
+        vm_macros.put(entry.key_ptr.*, entry.value_ptr.*) catch return CompileError.OutOfMemory;
     }
     ok = true;
     return c.func;
--- a/src/primitives_r7rs.zig
+++ b/src/primitives_r7rs.zig
@@ -179,7 +179,22 @@ fn evalFn(args: []const Value) PrimitiveError!Value {
     if (args.len > 1) {
         if (!types.isEnvironment(args[1])) return primitives.typeError("eval", "environment", args[1]);
         const se = types.toEnvironment(args[1]);
-        const func = compiler_mod.compileExpressionInEnv(gc, expr, &vm.macros, se.env, args[1], false) catch return PrimitiveError.TypeError; // bare-ok: compile error
+
+        // Interaction environment: macros persist globally (#1269).
+        // Custom environments: isolated macro table, discarded after compilation.
+        var local_macros: std.StringHashMap(Value) = undefined;
+        const use_local = (se.env != vm.globals);
+        if (use_local) {
+            local_macros = std.StringHashMap(Value).init(gc.allocator);
+            var mit = vm.macros.iterator();
+            while (mit.next()) |entry| {
+                local_macros.put(entry.key_ptr.*, entry.value_ptr.*) catch return PrimitiveError.OutOfMemory;
+            }
+        }
+        defer if (use_local) local_macros.deinit();
+
+        const macro_target: *std.StringHashMap(Value) = if (use_local) &local_macros else &vm.macros;
+        const func = compiler_mod.compileExpressionInEnv(gc, expr, macro_target, se.env, args[1], false) catch return PrimitiveError.TypeError; // bare-ok: compile error
         var closure_val = gc.allocClosure(func) catch return PrimitiveError.OutOfMemory;
         compiler_mod.Compiler.unrootFunction(gc, func);
         gc.pushRoot(&closure_val);
@@ -277,12 +292,27 @@ fn loadFn(args: []const Value) PrimitiveError!Value {
     var reader = reader_mod.Reader.init(gc, contents.items);
     defer reader.deinit();
 
+    // Custom environments get an isolated macro table so define-syntax
+    // does not leak into the global scope (#1269). The local table lives
+    // outside the loop so macros from earlier expressions remain visible.
+    var local_macros: std.StringHashMap(Value) = undefined;
+    const use_local = if (env) |e| (e != vm.globals) else false;
+    if (use_local) {
+        local_macros = std.StringHashMap(Value).init(gc.allocator);
+        var mit = vm.macros.iterator();
+        while (mit.next()) |entry| {
+            local_macros.put(entry.key_ptr.*, entry.value_ptr.*) catch return PrimitiveError.OutOfMemory;
+        }
+    }
+    defer if (use_local) local_macros.deinit();
+    const macro_target: *std.StringHashMap(Value) = if (use_local) &local_macros else &vm.macros;
+
     var last_result: Value = types.VOID;
     while (reader.hasMore() catch return PrimitiveError.TypeError) { // bare-ok: reader error in load
         const expr = reader.readDatum() catch return primitives.typeError("load", "valid datum", args[0]);
 
         if (env) |e| {
-            const func = compiler_mod.compileExpressionInEnv(gc, expr, &vm.macros, e, env_val, false) catch return primitives.typeError("load", "valid expression", args[0]);
+            const func = compiler_mod.compileExpressionInEnv(gc, expr, macro_target, e, env_val, false) catch return primitives.typeError("load", "valid expression", args[0]);
             var closure_val = gc.allocClosure(func) catch return PrimitiveError.OutOfMemory;
             compiler_mod.Compiler.unrootFunction(gc, func);
             gc.pushRoot(&closure_val);
--- a/src/tests_core_eval.zig
+++ b/src/tests_core_eval.zig
@@ -331,6 +331,47 @@ test "define-syntax in custom environment does not leak to global scope (#1269)"
     try std.testing.expectEqualStrings("not-leaked", types.symbolName(result));
 }
 
+// Genuine regression guard: constructs a mutable non-global environment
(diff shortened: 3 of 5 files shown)


src/compiler.zig
```diff
@@ -980,11 +980,9 @@ pub fn compileExpressionInEnv(gc: *memory.GC, expr: Value, vm_macros: *std.Strin
     c.func.env = env;
     c.func.env_val = env_val;
     c.func.restricted_globals = c.restricted_env;
-    if (!types.isEnvironment(env_val) or !types.toEnvironment(env_val).immutable) {
-        var out_it = c.macros.iterator();
-        while (out_it.next()) |entry| {
-            vm_macros.put(entry.key_ptr.*, entry.value_ptr.*) catch return CompileError.OutOfMemory;
-        }
+    var out_it = c.macros.iterator();
```
baijum, line 983: Removing the guard is safe: for an immutable `env_val`, `compileDefineSyntax` (`compiler_macro.zig:263`) already rejects top-level `define-syntax` before `compile()` returns, so this merge-back is never reached with new macros; and no current caller passes an immutable `env_val` together with `&vm.macros`. Worth noting it drops a safety net for any *future* caller that does — isolation policy now lives entirely in the callers.

src/primitives_r7rs.zig
```diff
@@ -179,7 +179,22 @@ fn evalFn(args: []const Value) PrimitiveError!Value {
...
         const se = types.toEnvironment(args[1]);
-        const func = compiler_mod.compileExpressionInEnv(gc, expr, &vm.macros, se.env, args[1], false) catch return PrimitiveError.TypeError; // bare-ok: compile error
+
+        // Interaction environment: macros persist globally (#1269).
+        // Custom environments: isolated macro table, discarded after compilation.
+        var local_macros: std.StringHashMap(Value) = undefined;
+        const use_local = (se.env != vm.globals);
+        if (use_local) {
+            local_macros = std.StringHashMap(Value).init(gc.allocator);
+            var mit = vm.macros.iterator();
```
baijum, line 189: This "clone `vm.macros` into a local table" block is duplicated verbatim in three places (here, `loadFn`, and `vm_dispatch.zig` `runUntil`). Consider extracting a small helper (e.g. `cloneMacroTable(gc, &vm.macros) -> StringHashMap(Value)`) so the three sites can't drift.

src/tests_core_eval.zig
```diff
@@ -310,6 +310,39 @@ test "eval define-syntax into immutable environment signals error (#1147)" {
     try std.testing.expectEqualStrings("error-signaled", types.symbolName(result));
 }
 
+test "define-syntax in custom environment does not leak to global scope (#1269)" {
```
baijum, line 313: This passes on pre-PR `main` as well — `(environment '(scheme base))` is immutable, so `define-syntax` into it already errors and nothing is added to `vm.macros`. Good characterization test, but it doesn't fail without this diff. A test using a *mutable non-global* env (via `gc.allocEnvironment(map, owned, false)`) would actually exercise the new merge-target path.

baijum (commented):
**Reviewed at `5d0a840`** — built clean on macOS ARM, `zig build test` green, and the new + related Scheme tests pass. The refactor is **correct and safe to merge**; two notes worth knowing first.

### 1. The leak this PR targets is already unreachable from Scheme — this is hardening, not a live-bug fix

`compileDefineSyntax` (`src/compiler_macro.zig:263`) rejects `define-syntax` into any **immutable** environment *before* the merge-back in `compileExpressionInEnv` is ever reached. Every Scheme-reachable non-global environment — `(environment ...)`, `null-environment`, `scheme-report-environment` — is immutable (`allocEnvironment(..., true, true)`), and the only **mutable** environment is the global one from `interaction-environment` (`env == vm.globals`). So the exact case this PR isolates — merge-back into `&vm.macros` for a *mutable, non-global* env — has **no Scheme-level trigger today**.

I confirmed the pre-PR `main` binary (`c42805aa`) produces identical, correct output for the issue's reproduction: the `define-syntax` into `(environment '(scheme base))` errors, and the macro is `not-leaked` at top level. The observable leak from #1269 was already closed (by the immutable-env `define-syntax` guard / #1275).

### 2. The new tests pass without the change

I ran the PR's `tests/scheme/smoke/eval-macro-leak-1269.scm` against the pre-PR `main` binary: **3/3 pass, exit 0**. The two new Zig unit tests are the same story — they exercise only the immutable-env path (already errors) and the interaction-environment path (already global), so they pass with or without this diff. They're good *characterization* tests but don't actually guard the merge-back-target logic.

To fail without this diff you'd need a **mutable non-global** environment, which is only constructible via the internal Zig API (`gc.allocEnvironment(map, owned, /*immutable=*/false)`). A Zig unit test built on that would be a genuine regression guard for the new caller-side isolation.

### Recommendation

Not a blocker — the change is a clean robustness improvement and removing the guard is safe (see inline note). Suggest either:
- add one Zig-level test that constructs a mutable non-global env (real regression guard), **or**
- soften the commit/PR wording from "fix leak" toward "harden macro-table isolation," so the history reflects that the observable leak was already fixed.

**Verdict: ready to merge.** CI still pending at review time; merge once green.

baijum:
Thanks for the thorough review! Both points were exactly right.

**Addressed in `e974097`:** Added a genuine regression guard that constructs a **mutable non-global** environment via the internal Zig API — the exact case the caller-side isolation targets. Verified it fails without the fix (macro count increases by 1) and passes with it.

**Re: CodeRabbit's GC rooting concern** — this is a false positive. Transformer Values in `local_macros` are always reachable through other roots:
- Original macros: reachable from `vm.macros` (the global table is traced by GC)
- New macros from `define-syntax`: also stored in `self.lib_env` (the custom env's map) by `compileDefineSyntax` (line 289), and the environment object itself is rooted via the VM register file

**Re: commit wording** — fair point that the observable Scheme-level leak was already closed by #1275's immutable-env guard. This PR is a structural fix: the merge-back now targets the right place by construction rather than relying on a downstream flag check. The new test confirms the hardening matters for the (currently internal-only) mutable non-global case.

baijum:
**Re-reviewed at `e974097`.** The new commit resolves my main note.

### Finding #2 (no genuine regression guard) — ✅ resolved and independently verified

The added test `define-syntax in mutable non-global env does not leak (#1269)` constructs a mutable non-global environment via `gc.allocEnvironment(env_map, true, /*immutable=*/false)` — the exact case caller-side isolation targets. I verified the commit's claim by applying **only this test** onto the pre-fix base (`c42805aa`):

- **Without the fix:** `expected 0, found 1` — `mut-leak-test` leaks into `vm.macros` (count +1). ✅ fails
- **With the fix (PR head):** full `zig build test` passes. ✅

So it's now a real regression guard for the merge-back-target logic, not just a characterization test. This also retires my finding #1 framing point — there genuinely *is* a leak in the merge-back path (reachable at the code level for mutable non-global envs); it was just unreachable from Scheme because all Scheme-constructible non-global envs are immutable.

### Remaining (optional, non-blocking)

The DRY note stands: the "clone `vm.macros` into a local table" block is still duplicated across `evalFn`, `loadFn`, and `vm_dispatch.zig`'s `runUntil`. A shared helper would prevent drift, but it's a nit — fine to leave or fold into a later cleanup.

### Verdict

**Ready to merge** once CI is green. Correct, safe, no regressions, and now backed by a test that actually fails without the change.

PR #1305, opened 2026-07-07
Make default-random-source a variable, not a procedure (#1192)

### LINKED ISSUE #1192: [SRFI-27] default-random-source is exported as a zero-argument procedure; spec requires a variable bound to a random source
**Category:** SRFI
**Spec reference:** SRFI-27, Specification: "default-random-source — A random source from which random-integer and random-real have been derived using random-source-make-integers and random-source-make-reals. Note that an assignment to default-random-source does not change random or random-real; it is also strongly recommended not to assign a new value." The entry is a variable binding (assignment to it is discussed), not a procedure.
**Severity:** wrong-result (portable SRFI-27 code breaks)

**Minimal reproduction:**
```scheme
(random-source? default-random-source)          ;; expected: #t  actual: #f
(procedure? default-random-source)              ;; expected: #f  actual: #t
(random-source-state-ref default-random-source) ;; expected: a state  actual: type error
```

**Expected:** `default-random-source` evaluates to the random source object.
**Actual:** it evaluates to a native 0-arg procedure; only the Kaappi-specific call form `(default-random-source)` yields the source, so portable code like `(random-source-state-ref default-random-source)` raises a type error.

**Suggested fix:** keep the primitive internal (e.g. `%default-random-source`) and in `lib/srfi/27.sld` export `(define default-random-source (%default-random-source))`.

**Source location:** src/primitives_random.zig (`defaultRandomSourceFn` registered as `default-random-source`), lib/srfi/27.sld (re-exports it directly)

baijum:
## Summary

- Rename the `default-random-source` primitive to `%default-random-source` (internal) and define the public SRFI-27 binding as a variable in `lib/srfi/27.sld` that evaluates the procedure once at library load time
- Portable SRFI-27 code like `(random-source-state-ref default-random-source)` now works correctly instead of raising a type error
- Remove FAIL marker from audit test, add regression test

## Test plan

- [x] New regression test `tests/scheme/smoke/srfi27-default-source.scm` — verifies `default-random-source` is a random source, not a procedure
- [x] Audit test passes (49/49, FAIL marker removed)
- [x] SRFI-27 state test passes (8/8)
- [x] All 663 unit tests pass
- [x] All 1,391 R7RS tests pass

Closes #1192

### DIFF
--- a/lib/srfi/27.sld
+++ b/lib/srfi/27.sld
@@ -8,6 +8,8 @@
           random-source-state-ref random-source-state-set!)
   (begin
 
+    (define default-random-source (%default-random-source))
+
     (define (random-source-make-integers rs)
       (lambda (n) (%rs-next-int rs n)))
 
--- a/src/primitives_random.zig
+++ b/src/primitives_random.zig
@@ -27,7 +27,7 @@ fn freshSeed() u64 {
 pub const specs = [_]primitives.PrimSpec{
     .{ .name = "random-integer", .func = &randomIntegerFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_base) },
     .{ .name = "random-real", .func = &randomRealFn, .arity = .{ .exact = 0 }, .libs = LS.initOne(.scheme_base) },
-    .{ .name = "default-random-source", .func = &defaultRandomSourceFn, .arity = .{ .exact = 0 }, .libs = LS.initOne(.scheme_base) },
+    .{ .name = "%default-random-source", .func = &defaultRandomSourceFn, .arity = .{ .exact = 0 }, .libs = LS.initOne(.scheme_base) },
     .{ .name = "random-source?", .func = &randomSourcePFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_base) },
     .{ .name = "make-random-source", .func = &makeRandomSourceFn, .arity = .{ .exact = 0 }, .libs = LS.initOne(.scheme_base) },
     .{ .name = "random-source-randomize!", .func = &randomSourceRandomizeFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_base) },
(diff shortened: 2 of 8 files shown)


baijum:
## Review — ready to merge ✅

Clean, minimal fix that implements exactly the approach suggested in #1192: keep the primitive internal as `%default-random-source` and bind the public `default-random-source` variable once at library-load time in `lib/srfi/27.sld`. The design is sound because `random-integer`/`random-real` and `%default-random-source` all read the same per-VM `vm.default_random_source` object, so the variable stays in sync with the source those procedures actually draw from.

**Verified locally** (fresh ReleaseSafe build of this branch):
- `random-source? default-random-source` → `#t`; `procedure? default-random-source` → `#f`; `random-source-state-ref default-random-source` no longer type-errors.
- Mutation propagates: `(random-source-pseudo-randomize! default-random-source i j)` reproducibly seeds subsequent `random-integer` draws — confirming the variable is `eq?` to the source `random-integer` uses.
- `zig build test` green; audit test 49/49; `srfi27-state` 8/8; new smoke test 3/3; all CI checks pass.
- Repo-wide grep confirms every `default-random-source` reference (audit, coverage, srfi tests) was updated — no stale call-form usages remain.

### Non-blocking notes

1. **`%default-random-source` is still registered with `.libs = LS.initOne(.scheme_base)`** ([src/primitives_random.zig:30](src/primitives_random.zig#L30)), so the internal helper remains reachable from plain `(import (scheme base))`. This matches the existing `%rs-next-int` / `%rs-next-real` convention, so it's not a regression — just flagging in case the project wants `%`-prefixed helpers to be truly library-private someday.

2. **Stale audit-strategy note**: `docs/audit-strategy.md:137` still lists "default-random-source is a procedure not a variable" among the disabled Phase-2.15 items. Optional to update now that #1192 is resolved.

3. **Minor test-coverage nit** (also raised by CodeRabbit): the renamed unit test `default-random-source is per-VM` now calls `(%default-random-source)`, exercising the internal primitive rather than the exported SRFI-27 binding. That's fine — exercising the exported binding at the unit level would need the `.sld` loader in `makeTestVM`, and the new `tests/scheme/smoke/srfi27-default-source.scm` already covers the exported variable end-to-end.

None of these block merge.

PR #1306, opened 2026-07-07
Honor explicit #f thread arg in mutex-lock! as locked/not-owned (#1154)

### LINKED ISSUE #1154: [SRFI-18] mutex-lock! with explicit #f thread argument assigns the current thread as owner instead of locked/not-owned
**Category:** SRFI-18
**Spec reference:** SRFI-18, mutex-lock!: "If thread is #f the mutex becomes locked/not-owned"; mutex-state: "returns ... the symbol not-owned when the mutex is locked/not-owned".
**Severity:** wrong-result

**Minimal reproduction:**
```scheme
(import (srfi 18))
(define m (make-mutex))
(mutex-lock! m #f #f)   ;; timeout=#f, thread=#f => locked/not-owned
(mutex-state m)         ;; expected: not-owned
                        ;; actual:   #<fiber 0 running> (current thread)
```

**Root cause:** `src/primitives_srfi18.zig` `mutexLockFn` — the owner assignment treats any non-fiber third argument as "absent" and falls back to the current fiber:
```zig
m.owner = if (args.len > 2 and types.isFiber(args[2]))
    args[2]
else if (ctx.vm.current_fiber) |cf| types.makePointer(@ptrCast(cf))
else types.VOID;
```
An explicit `#f` must produce `owner = VOID` (locked/not-owned). The same pattern appears three times in the function (abandoned path, fast path, contended path).

baijum:
## Summary

- Fix `mutex-lock!` to treat an explicit `#f` thread argument as locked/not-owned per SRFI-18, instead of falling back to the current thread as owner
- Patch all three owner-assignment sites in `mutexLockFn` (abandoned path, fast-lock path, contended path)
- Re-enable the disabled audit test and add a dedicated regression test

## Test plan

- [x] New regression test `tests/scheme/smoke/mutex-lock-false-owner.scm` passes
- [x] Re-enabled audit test in `primitives_srfi18-audit.scm` passes (76/76)
- [x] SRFI-18 test suite passes (44/44)
- [x] Zig unit tests pass

Closes #1154

### DIFF
--- a/src/primitives_srfi18.zig
+++ b/src/primitives_srfi18.zig
@@ -695,6 +695,8 @@ fn mutexLockFn(args: []const Value) PrimitiveError!Value {
         const ctx = try ensureScheduler();
         m.owner = if (args.len > 2 and types.isFiber(args[2]))
             args[2]
+        else if (args.len > 2 and args[2] == types.FALSE)
+            types.VOID
         else if (ctx.vm.current_fiber) |cf|
             types.makePointer(@ptrCast(cf))
         else
@@ -707,6 +709,8 @@ fn mutexLockFn(args: []const Value) PrimitiveError!Value {
         const ctx = try ensureScheduler();
         m.owner = if (args.len > 2 and types.isFiber(args[2]))
             args[2]
+        else if (args.len > 2 and args[2] == types.FALSE)
+            types.VOID
         else if (ctx.vm.current_fiber) |cf|
             types.makePointer(@ptrCast(cf))
         else
@@ -739,6 +743,8 @@ fn mutexLockFn(args: []const Value) PrimitiveError!Value {
     m.locked = true;
     m.owner = if (args.len > 2 and types.isFiber(args[2]))
         args[2]
+    else if (args.len > 2 and args[2] == types.FALSE)
+        types.VOID
     else
         types.makePointer(@ptrCast(me));
 
--- a/tests/scheme/audit/primitives_srfi18-audit.scm
+++ b/tests/scheme/audit/primitives_srfi18-audit.scm
@@ -91,10 +91,9 @@
 (let ((m (make-mutex)))
   (mutex-lock! m)
   (test #f (mutex-lock! m 0.05)))
-;; FAIL: #1154 (explicit #f thread argument must yield locked/not-owned)
-;; (let ((m (make-mutex)))
-;;   (mutex-lock! m #f #f)
-;;   (test 'not-owned (mutex-state m)))
+(let ((m (make-mutex)))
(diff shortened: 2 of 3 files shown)


baijum (commented):
## Review — ready to merge ✅

Verified locally on a fresh `zig build` at the PR head (based on current `main`):

- New smoke test `tests/scheme/smoke/mutex-lock-false-owner.scm` → **3/3 pass**
- Re-enabled audit test `primitives_srfi18-audit.scm` → **76/76 pass**
- `srfi/srfi18.scm` → **44/44**, `srfi18-abandoned-mutex.scm` → **3/3**
- `zig build test` → pass
- CI green on all platforms (macOS, Ubuntu x86/ARM, riscv64, wasm, Debug/ReleaseFast/ReleaseSafe)

The fix is correct and minimal. `Value` is a `u64` and `types.FALSE` is a canonical immediate bit pattern, so `args[2] == types.FALSE` is an exact, safe comparison. Setting `owner = VOID` produces `not-owned` via the existing `mutex-state` path (`src/primitives_srfi18.zig:681`), and the abandonment path (`src/primitives_srfi18.zig:165`) only abandons mutexes where `m.owner == fiber_val`, so a locked/not-owned mutex correctly survives the locking fiber's death — matching SRFI-18. All three owner-assignment sites are patched consistently. Good regression test that grabs the runner before `test-end` per the SRFI-64 convention.

### Non-blocking nits (optional)

1. **Duplicated owner-resolution logic.** The `if isFiber → else if == FALSE → else current-fiber` ladder is now repeated three times. Could be extracted into a small helper (e.g. `fn resolveLockOwner(args, fallback) Value`) to keep the three paths from drifting apart in future. Not required for this PR.

2. **Contended path is uncovered.** All three smoke assertions plus the audit test exercise only the fast/unlocked path (line ~710). The new `#f` branch in the contended path (`src/primitives_srfi18.zig:745`) isn't directly tested. It's clearly parallel to the fast path so risk is low, but a test that locks a held mutex from another fiber with an explicit `#f` owner would close the gap.

3. **Redundant second smoke assertion.** `(not (eq? (mutex-state m) 'not-abandoned))` under the label `"mutex is locked"` is already implied by the first assertion (which checks the state is exactly `'not-owned`), and the label is slightly misleading. Minor.

4. **Pre-existing, out of scope:** a non-`#f`, non-fiber thread argument (e.g. `(mutex-lock! m #f 42)`) still silently falls back to the current fiber rather than raising a type error. Not introduced by this PR; worth a follow-up if strict SRFI-18 argument validation is wanted.

None of these block merge — the fix does exactly what #1154 asks for and is well-tested.

PR #1307, opened 2026-07-07
Fix group-info by name returning gid 0 (#1161)

### LINKED ISSUE #1161: [SRFI-170] group-info by name always returns gid 0 — Zig std.c.getgrnam is misdeclared as returning ?*passwd
**Category:** SRFI-170
**Spec reference:** SRFI-170 — "(group-info gid/name) -> record" where the argument "is either an exact integer group id or a string group name"; "(group-info:gid group-info) -> exact integer".
**Severity:** wrong-result

**Minimal reproduction (macOS, staff = gid 20):**
```scheme
(import (srfi 170))
(group-info:gid (group-info 20))        ;; => 20  (by-gid path correct)
(group-info:gid (group-info "staff"))   ;; => 0   (WRONG — expected 20)
(group-info:name (group-info "staff"))  ;; => "staff" (name field happens to work)
```
Every by-name lookup returns gid 0 regardless of the group.

**Root cause: upstream Zig 0.16 stdlib misdeclaration.** `std.c.getgrnam` is declared with the wrong return type in `lib/zig/std/c.zig:10331`:
```zig
pub extern "c" fn getgrnam(name: [*:0]const u8) ?*passwd;   // should be ?*group
pub extern "c" fn getgrgid(gid: gid_t) ?*group;             // correct
```
Reading `.gid` through `struct passwd`'s layout hits the wrong byte offset (20 instead of 16 on macOS), which lands past `gr_gid` and yields 0; `.name` is at offset 0 in both structs, which is why the name accessor works and the code compiled silently. Verified with side-by-side C vs `zig run` programs: C `getgrnam("staff")->gr_gid` = 20, Zig `std.c.getgrnam("staff").?.gid` = 0, same libSystem.

**Suggested fix:** declare a local, correctly-typed extern in `src/primitives_filesystem.zig` (as the file already does for `chown`, `mkfifo`, etc.) instead of `std.c.getgrnam`, and consider reporting the misdeclaration upstream to Zig.

**Source location:** `src/primitives_filesystem.zig` `groupInfoFn` (~line 896).
**Verified:** fresh ReleaseSafe build at 83718b6; deterministic.

baijum:
## Summary

- `std.c.getgrnam` in Zig 0.16 is misdeclared as returning `?*passwd` instead of `?*group`, causing `(group-info "staff")` to always return gid 0
- Declare a local `extern "c" fn getgrnam` with the correct `?*std.c.group` return type
- Uncomment the disabled audit test and add a dedicated regression test

Closes #1161

## Test plan

- [x] `zig-out/bin/kaappi tests/scheme/smoke/group-info-by-name-1161.scm` — regression test passes (2/2)
- [x] `zig-out/bin/kaappi tests/scheme/audit/primitives_filesystem-audit.scm` — audit suite passes (66/66)
- [x] Verified `(group-info:gid (group-info "staff"))` returns 20 (was 0)

### DIFF
--- a/src/primitives_filesystem.zig
+++ b/src/primitives_filesystem.zig
@@ -18,6 +18,8 @@ extern "c" fn getgroups(size: c_int, list: [*]std.c.gid_t) c_int;
 extern "c" fn nice(inc: c_int) c_int;
 extern "c" fn setenv(name: [*:0]const u8, value: [*:0]const u8, overwrite: c_int) c_int;
 extern "c" fn unsetenv(name: [*:0]const u8) c_int;
+// std.c.getgrnam is misdeclared as returning ?*passwd (Zig 0.16 bug)
+extern "c" fn getgrnam(name: [*:0]const u8) ?*std.c.group;
 const is_linux = @import("builtin").os.tag == .linux;
 
 fn validateMode(gc: *GC, val: Value) PrimitiveError!std.c.mode_t {
@@ -893,7 +895,7 @@ fn groupInfoFn(args: []const Value) PrimitiveError!Value {
         try validatePathNoNul(name, args[0]);
         const name_z = gc.allocator.dupeZ(u8, name) catch return PrimitiveError.OutOfMemory;
         defer gc.allocator.free(name_z);
-        const g = std.c.getgrnam(name_z) orelse return types.FALSE;
+        const g = getgrnam(name_z) orelse return types.FALSE;
         const name_str = std.mem.span(g.name.?);
         return gc.allocGroupInfo(name_str, g.gid) catch return PrimitiveError.OutOfMemory;
     } else return primitives.typeError("group-info", "string or integer", args[0]);
--- a/tests/scheme/audit/primitives_filesystem-audit.scm
+++ b/tests/scheme/audit/primitives_filesystem-audit.scm
@@ -130,9 +130,8 @@
   (test #t (group-info? gi))
   (test #t (string? (group-info:name gi)))
   (test (user-gid) (group-info:gid gi)))
-;; FAIL: #1161 (group-info by name returns gid 0 — Zig std.c.getgrnam misdeclared)
-;; (test (user-gid)
-;;   (group-info:gid (group-info (group-info:name (group-info (user-gid))))))
+(test (user-gid)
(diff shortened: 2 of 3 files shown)


baijum:

Reviewed the diff and independently verified the fix, mechanism, and test validity locally (macOS arm64, Zig 0.16).

### Root cause confirmed
`std/c.zig:10331` really does misdeclare it:
```zig
pub extern "c" fn getgrnam(name: [*:0]const u8) ?*passwd;   // ← should be ?*group
pub extern "c" fn getgrgid(gid: gid_t) ?*group;             // correct
```
So the bug hits only the by-name path. Why gid came back as `0` specifically: on macOS the real `group` struct is `{name@0, passwd@8, gid@16, mem@24}` (4 bytes of padding at offset 20), while `passwd.gid` lives at offset 20 — reading through the wrong type lands in that padding → `0`. `passwd.name`/`group.name` are both at offset 0, which is why the *name* came back correct and only the gid was wrong.

### Fix is correctly scoped
The other three siblings in this file are all declared correctly, so a local `extern` for `getgrnam` alone is the right, minimal fix:
- `std.c.getpwnam` / `std.c.getpwuid` → `?*passwd` ✓ (correct)
- `std.c.getgrgid` → `?*group` ✓ (correct)
- `std.c.getgrnam` → `?*passwd` ✗ (fixed here)

### Verification
- `zig build` — clean
- Smoke test `group-info-by-name-1161.scm` — 2/2 pass; **fails without the source fix** (`by-name gid matches by-id gid` → FAIL), so it's a genuine regression test
- Audit suite `primitives_filesystem-audit.scm` — 66/66; the re-enabled assertion reports `expected 20 got 0` when the fix is reverted, and `run-all.sh` gates on that printed count (it doesn't trust exit codes for chibi-test files), so re-enabling it adds real protection
- Direct check: `(group-info:gid (group-info "staff"))` → `20` (was `0`)
- All CI green (macOS, ubuntu x86_64/arm, riscv64, wasm, format); benchmarks flat

One nice touch: the smoke test derives the expected group from `(user-gid)` rather than hardcoding `staff`/`20`, so it stays portable across CI environments, and the epilogue grabs the runner before `test-end` per the SRFI-64 pattern.

No blocking issues, no nits. 👍

PR #1308, opened 2026-07-07
Add nine missing SRFI-133 procedures (#1172)

### LINKED ISSUE #1172: [SRFI-133] nine spec procedures missing: vector=, vector-fold(-right), vector-map!, vector-reverse-copy!, vector-unfold(-right)!, reverse-vector->list, reverse-list->vector
**Category:** SRFI
**Spec reference:** SRFI-133 procedure index — the library is required to export (among others): `vector=`, `vector-fold`, `vector-fold-right`, `vector-map!`, `vector-reverse-copy!`, `vector-unfold!`, `vector-unfold-right!`, `reverse-vector->list`, `reverse-list->vector`.
**Severity:** missing-feature

**Minimal reproduction:**
```scheme
(import (srfi 133))
(vector-fold + 0 (vector 1 2 3))
;; expected: 6
;; actual: error: undefined variable 'vector-fold'. Did you mean 'vector-unfold'?
(vector= equal? (vector 1 2) (vector 1 2))
;; actual: error: undefined variable 'vector='
```

All nine names above are unbound after `(import (srfi 133))` (verified by referencing each one individually). The other 34 SRFI-133 procedures are present and behaviorally conform (175 audit assertions pass).

Note: kaappi's `(import (only (srfi 133) vector-fold))` *succeeds* despite the name not existing — that masking is filed separately (only/except/rename don't validate identifiers).

**Prior art:** #148 (closed) tracked SRFI-133 gaps but listed only `vector-append-subvectors` (since implemented); #816 (closed) fixed the `vector->string`/`string->vector` exports. This issue covers the remaining gap. `lib/srfi/43.sld` has portable `vector-fold`/`vector-fold-right`/`vector-map!` implementations that could be adapted (SRFI-43 fold passes the index — SRFI-133 fold does NOT, so they need adjustment, not reuse).

**Source location:** src/primitives_vector.zig (specs table — no registrations for these names; SRFI-133 is fully built-in, there is no lib/srfi/133.sld)

baijum:
## Summary

- Implement 9 spec-required SRFI-133 procedures as native Zig primitives: `vector=`, `vector-fold`, `vector-fold-right`, `vector-map!`, `vector-reverse-copy!`, `vector-unfold!`, `vector-unfold-right!`, `reverse-vector->list`, `reverse-list->vector`
- All higher-order procedures support multi-vector arguments per the SRFI-133 spec
- GC safety: write barriers on all mutating procedures, accumulator rooting in folds, `rootedScope` for unfold seed state

Closes #1172

## Test plan

- [x] 37 new test cases in `tests/scheme/srfi/srfi133-ext.scm` (76/76 pass)
- [x] 4 audit assertions uncommented in `tests/scheme/audit/primitives_vector-audit.scm` (180/180 pass)
- [x] R7RS test suite — no regressions
- [x] SRFI-43 tests — unaffected (its local definitions shadow the new SRFI-133 built-ins)


baijum (commented):
## Review — ready to merge ✅

Reviewed by checking out the PR commit (`3611aba`, based directly on current `main` `4fdb4eb1`), rebuilding, and running the suites locally. This is a clean, spec-faithful implementation.

### Correctness (verified against SRFI-133)
- **Argument order is right** where it matters: folds pass the accumulator first (`(kons state elem ...)`, *not* the SRFI-1 order), unfolds pass the absolute index first (`(f i seed ...)`), reversal direction is correct in every procedure. Multi-vector higher-order procedures correctly truncate to the shortest input.
- `vector=` handles the 0-/1-vector cases (`#t`), checks length before elements, and short-circuits — confirmed with `(vector= =)`, `(vector= = #(1))`, and the 3-vector fail case.
- Immutability checks on every mutating target (`vector-map!`, `vector-reverse-copy!`, `vector-unfold!`, `vector-unfold-right!`) using the established `flags.immutable` pattern, and write barriers on all stores.

### GC safety — verified under `-Dgc-stress=true`
The rooting is sound. Source vectors stay rooted through `callVM` because native `args` alias into `vm.registers` and are marked by the caller's frame window (`vm.zig:59`), so vector elements pulled into stack buffers are safe transitively. Fold accumulators are `pushRoot`'d and unfold seeds are re-rooted each iteration. I stress-tested (GC on every allocation) with heap-heavy accumulators, growing list seeds, and heap map!/unfold results — **all results correct**.

### Tests reproduced locally
- `tests/scheme/srfi/srfi133-ext.scm` → **76/76 pass**
- `tests/scheme/audit/primitives_vector-audit.scm` → **180/180 pass**
- `zig build test` → pass · R7RS suite → no regressions · SRFI-43 → 20/20 (correctly shadowed)
- `zig fmt --check` clean · file is 1265 lines (< 1500)

Edge cases spot-checked and all behave: immutable-literal rejection, out-of-range `start`/`end`, improper list to `reverse-list->vector`, shortest-vector truncation leaving the target tail untouched.

### Optional follow-ups (non-blocking)
1. **`extra_roots` growth in the unfold pair** (`primitives_vector.zig` `vectorUnfoldBangFn`/`vectorUnfoldRightBangFn`): seeds are appended to `gc.extra_roots` on *every* iteration and only released at `scope.release()`, so the root set grows O(N × seeds) across the loop and each intervening collection re-scans it. This exactly mirrors the already-shipped non-mutating `vector-unfold`/`vector-unfold-right`, so it's consistent — but rooting the seeds once in fixed `RootedSlot`s and updating in place would fix all four together. Worth a follow-up rather than a change here.
2. **`vector-reverse-copy!`** doesn't special-case `to ≡ from` with overlapping ranges — but SRFI-133 declares that "an error", so this is spec-compliant. A one-line comment noting the assumption would help future readers.
3. Trivial: the seed-update loop is written two slightly different ways in `vector-unfold!` vs `vector-unfold-right!` (index guard vs slice iterate). Functionally identical; each matches its non-mutating sibling, so no action needed.

None of these block merge.

PR #1309, opened 2026-07-07
Fix FFI char type to accept Scheme characters and return characters (#1186)

### LINKED ISSUE #1186: [FFI] the char type neither accepts nor produces Scheme characters — (f #\A) raises a bare error, char returns yield fixnums
**Category:** FFI

**Minimal reproduction:**
```scheme
(define lib (ffi-open #f))
(define c-tolower (ffi-fn lib "tolower" '(char) 'char))
(c-tolower 65)     ;; => 97 (fixnum in, fixnum out — works as an int alias)
(c-tolower #\A)    ;; error: "error" (no detail)
```

**Expected:** the point of a distinct `char` FFI type (one of the 18 documented types in CLAUDE.md) is bridging Scheme characters: `#\A` should marshal to 65, and a `char` return should produce `#\a` (or the docs should say it's a range-checked uint8 alias and reject nothing new).
**Actual:** `char_type` params go through `toIntArgOpt` (src/ffi.zig:195), which accepts integers only — character values raise `error.TypeError` with no detail. `char_type` returns are normalized to `.int` (src/ffi.zig:170) and yield fixnums. So the type is behaviorally identical to `uint8` and rejects the one Scheme type it names.

**Suggested fix:** in the param path, accept `types.isChar` (codepoint must fit 0–255, else range error) in addition to integers; optionally make `char` returns produce character values (breaking-change consideration — at minimum document the alias behavior).

**Verified:** fresh ReleaseSafe build at 055a5b1
**Source location:** src/ffi.zig marshal paths (lines ~170, ~195, ~202)
**Disabled test:** tests/scheme/audit/primitives_ffi-audit.scm (marked `;; FAIL: #this`)

**Found by:** Systematic audit, Phase 2.13

baijum:
## Summary

- FFI `char` params now accept Scheme character values (e.g. `#\A`) in addition to integers — the codepoint is extracted and range-checked to 0–255
- FFI `char` returns now produce Scheme character values instead of fixnums (e.g. `#\a` instead of `97`)
- Integers are still accepted for backward compatibility

Closes #1186

## Test plan

- [x] 5 new Zig unit tests in `src/ffi.zig` (char extraction, validation, range checks)
- [x] New Scheme regression test `tests/scheme/ffi/char-type.scm` (6 assertions)
- [x] Audit test `primitives_ffi-audit.scm` updated — `#1186` FAIL markers removed, tests enabled
- [x] `zig build test` passes
- [x] Original issue reproduction verified: `(c-tolower #\A)` → `#\a`


baijum:
## Review — FFI `char` type fix

**Verdict: ready to merge.** ✅

This is a clean, correctly-scoped fix that matches the suggested fix in #1186 exactly: `char` params now accept `#\A` (codepoint range-checked to 0–255) in addition to integers, and `char` returns produce Scheme characters. Backward compatibility (integers in) is preserved.

### What I verified
- **Build + tests pass** at `b73e96f` on top of current `main`: `zig build test` → 669/669 pass (includes the 5 new `src/ffi.zig` unit tests); all 12 `tests/scheme/ffi/*.scm` pass; the updated audit `primitives_ffi-audit.scm` → 48 pass, 0 fail.
- **Wiring is correct**: `char-type.scm` is auto-discovered by the `tests/scheme/ffi/*.scm` glob in `run-all.sh`. The SRFI-64 test grabs the runner *before* `test-end`, matching the recommended ending pattern.
- **Return conversion placement** (`callFfi`, after the arity switch) means it applies uniformly across all arities 0–5, mirroring the adjacent `bool_type` normalization. `char_type` normalizes to `.int`, so the C result is always a fixnum before conversion — the `isFixnum(result)` guard holds.
- **No ecosystem breakage**: no `kaappi-*` library declares a `'char` FFI param/return, so the "char returns now yield characters instead of fixnums" behavior change (flagged as a breaking-change consideration in #1186) affects nothing downstream.

### Minor observations (non-blocking)

1. **Asymmetric range check between param and return.** The param path range-checks `char` to `0–255` (`checkNarrowIntRange`), but the return path accepts `0–0x10FFFF` (full Unicode) before `makeChar`:
   ```zig
   if (code >= 0 and code <= 0x10FFFF)
       return types.makeChar(@intCast(code));
   ```
   A real C `char`-returning function is byte-valued, so this never bites in practice, but it's a little loose. Two edge consequences worth being aware of: a negative return (e.g. a `signed char` of `-1` widened through the `c_int` trampoline) silently falls through to a raw fixnum rather than erroring, and a large out-of-byte-range int would become a high Unicode char rather than an error. Tightening the bound to `0–255` would make the two directions symmetric. Fine to leave as-is.

2. **Stale audit note.** `docs/audit-strategy.md:135` still records this as "47 audit tests + 3 disabled … char type rejects characters/returns fixnums." With this PR that's now 48 tests + 2 disabled and the char behavior is fixed. A one-line update there would keep the audit ledger accurate (can be a follow-up).

Nice work — tests are focused and cover the param-accept, return-produce, backward-compat, and >255-rejection cases.

*Automated review by Claude Code.*

PR #1310, opened 2026-07-07
Fix bitwise and/ior/xor for negative operands in SRFI-151/143

### LINKED ISSUE #1214: [SRFI-151/60/143] bitwise and/ior/xor ignore two's-complement sign — wrong results for any negative operand; arithmetic-shift right truncates instead of floors
**Category:** SRFI (pure-Scheme bitwise stack)
**Spec reference:** SRFI-151/SRFI-60 define bitwise operations "as if the integers were represented in two's-complement, with an infinite number of sign bits". SRFI-60: (logand -4 8) => 8. Arithmetic right shift is floor division: SRFI-60 "ash: ... equivalent to (floor (* n (expt 2 count)))".
**Severity:** wrong-result (silent, affects three libraries)

```scheme
(import (scheme base) (srfi 60))
(logand -4 8)    ;; expected 8    actual 0
(logand -1 5)    ;; expected 5    actual 1
(logand -4 -6)   ;; expected -8   actual 4   (magnitudes ANDed, sign dropped)
(logior -4 2)    ;; expected -2   actual 6
(logxor -1 5)    ;; expected -6   actual 4
(logxor -4 8)    ;; expected -12  actual 12
(bitwise-merge 3 1 8) ;; expected 9  actual 1
(ash -15 -2)     ;; expected -4 (floor) actual -3 (truncated)

(import (srfi 143))
(fxand -4 8)     ;; expected 8    actual 0
(fxif 3 1 8)     ;; expected 9    actual 1
```

**Root cause:** lib/srfi/151.sld's `%bitwise-and2/%bitwise-ior2/%bitwise-xor2` recurse with truncating `quotient`/`odd?` and terminate at `(= a 0)` — a negative number never reaches the `-1` fixed point of two's-complement, so its bits are effectively the truncated-magnitude walk. The recursion must terminate at `(and (or (= a 0) (= a -1)) (or (= b 0) (= b -1)))` with floor division (or dispatch on signs via De Morgan identities). `arithmetic-shift` right uses `quotient` where floor division is required. lib/srfi/143.sld duplicates the same broken recursion in its own `fxand/fxior/fxxor` (and its shifts inherit the truncation).

**Blast radius:** (srfi 151) bitwise-and/ior/xor/eqv/nand/nor/andc1/andc2/orc1/orc2 and everything derived; (srfi 60) logand/logior/logxor/logtest/bitwise-merge/ash via its (srfi 151) import; (srfi 143) fxand/fxior/fxxor/fxif/fxarithmetic-shift-right. Positive-operand results are correct, which is why smoke tests never caught it.

**Source location:** lib/srfi/151.sld (helpers, arithmetic-shift), lib/srfi/143.sld (fxand/fxior/fxxor)
**Disabled test:** tests/scheme/srfi/srfi143.scm, tests/scheme/srfi/srfi60.scm (marked ;; FAIL: #this)

**Found by:** Systematic audit, Phase 3c/3d

baijum:
## Summary

- Fix two's-complement bitwise ops (`and`/`ior`/`xor`) in SRFI-151 and SRFI-143: add `-1` base cases and switch `quotient` → `floor-quotient` in the recursive bit-walk
- Fix `arithmetic-shift` right path: `quotient` (truncates toward zero) → `floor-quotient` (floors toward -∞)
- SRFI-60 inherits the fix via its SRFI-151 re-exports — no code changes needed

Closes #1214

## Root cause

Two bugs in the pure-Scheme bitwise helpers:

1. **Wrong division:** `quotient` truncates toward zero, but two's-complement bit-walking requires `floor-quotient`. E.g. `(quotient -1 2)` → `0` (escapes the negative domain), while `(floor-quotient -1 2)` → `-1` (the correct two's-complement fixed point).

2. **Missing `-1` base cases:** The recursion terminated only at `0`, but in infinite-precision two's-complement, `-1` (all-ones) is the other fixed point. Without it, negative operands were treated as positive magnitudes.

## Test plan

- [x] SRFI-151: 97 tests pass (20 previously disabled tests enabled, 9 new negative-operand tests added)
- [x] SRFI-143: 65 tests pass (1 previously disabled test enabled, 9 new tests added)
- [x] SRFI-60: 27 tests pass (7 previously disabled tests enabled)
- [x] Full test suite: 1795 pass, 0 fail, 0 timeout
- [x] Zig unit tests: pass


baijum:
## Review — looks correct and complete ✅

Checked out the PR head, rebuilt (ReleaseSafe), and re-derived the two's-complement results independently rather than trusting the implementation.

### Correctness
The two fixes are both right and, importantly, mutually necessary:

- **`quotient` → `floor-quotient` in the recursive step + right shift.** Truncating division escapes the negative domain (`(quotient -1 2)` → `0`), so the bit-walk never reaches the all-ones fixed point. `floor-quotient` is arithmetic right shift, which is what two's-complement bit-walking requires. `floor-quotient` also handles the bignum/flonum paths (`primitives_numeric.zig:1322`), so operands beyond the ±2^47 fixnum range work too — verified `(bitwise-and -1 1152921504606846976)` etc.
- **`-1` base cases.** With floor division, negatives converge to `-1` (all ones), so terminating only at `0` was incorrect. The added `-1` arms are the right identities: `and`/`ior` return the other operand (or `-1`), `xor` returns the complement. Base-case ordering is sound for the `0`/`-1` overlap cases (checked `and 0 -1`, `ior -1 0`, `xor -1 0`, `xor -1 -1`).

As a nice side effect, the new `xor` base cases (`(= a 0) → b`) short-circuit instead of walking the wider operand to zero — same result, shallower recursion than before.

### Scope
Complete for #1214. The three bitwise SRFI libs are 60/143/151; 60 re-exports 151, so the two edited files cover all of them. The remaining `quotient` uses in `bit-count`/`integer-length` (`151.sld:81,86`) are correct — they run only after negatives are complemented to non-negative, where `quotient` == `floor-quotient`.

### Tests
Strong. Beyond re-enabling the disabled cases and adding negative-operand tests, the author **recomputed two stale expected values** that were wrong in the old disabled comments rather than blindly uncommenting:
- `(bitwise-xor -2 5)`: `-7` → **`-5`** (correct: `…11110 ⊕ …00101 = …11011`)
- `(bit-field-replace-same 17 42 1 4)`: `21` → **`27`**

Verified locally on the PR build:
- SRFI-151: 97 pass, 0 fail · SRFI-143: 65 · SRFI-60: 27
- R7RS suite: 1395 pass, 0 fail · Zig unit tests: pass
- Full `run-all.sh`: **1795 pass, 0 fail, 0 timeout**

### Optional nit (non-blocking)
Stylistic inconsistency between the two files: `151.sld` recurses with `(floor-quotient a 2)` directly, while `143.sld` recurses with `(arithmetic-shift a -1)` (which now bottoms out in `floor-quotient`). Both are correct — just worth a mental note that the two libraries still carry near-duplicate helper logic.

**Verdict: ready to merge.**

PR #1312, opened 2026-07-07
Implement string-join grammar argument per SRFI-13 (#825)

### LINKED ISSUE #825: string-join: default delimiter is "" instead of single space; grammar argument silently ignored
## Summary

`string-join` in `src/primitives_string_ext.zig` deviates from SRFI-13 (and from Kaappi's own documentation) in two ways:

1. The default delimiter is `""` (empty string). SRFI-13 says: "delimiter: string — default is a single space." Kaappi's docs (`docs/procedures/srfi-13.md`) also state "The optional *delimiter* defaults to a single space."
2. The optional `grammar` argument (`'infix`, `'strict-infix`, `'suffix`, `'prefix`) is silently ignored — the third argument is never examined, so `'prefix`/`'suffix` produce infix output, and `'strict-infix` with an empty list returns `""` instead of signaling an error (SRFI-13: "it is an error to pass an empty list of strings" with `strict-infix`). The docs document all three grammars with worked examples that currently fail.

## Location

`src/primitives_string_ext.zig`, `stringJoinFn`, lines 355-393:

```zig
const delim: []const u8 = if (args.len > 1)
    (try getStringSlice(args[1]))
else
    "";                                  // SRFI-13 default is " "
```

Registered as `.{ .variadic = 1 }` (line 28), so a third `grammar` argument is accepted but `args[2]` is never read anywhere in the function.

## Reproduction

```scheme
(import (scheme base) (scheme write) (srfi 13))
(write (string-join '("one" "two" "three"))) (newline)
(write (string-join '("usr" "local" "bin") "/" 'prefix)) (newline)
(write (string-join '("a" "b" "c") ";" 'suffix)) (newline)
(write (string-join '() ";" 'strict-infix)) (newline)
```

## Actual vs Expected (Kaappi v0.11.1)

| Expression | Actual | Expected |
|---|---|---|
| `(string-join '("one" "two" "three"))` | `"onetwothree"` | `"one two three"` (SRFI-13: `(string-join '("foo" "bar" "baz")) => "foo bar baz"`; same example in Kaappi docs) |
| `(string-join '("usr" "local" "bin") "/" 'prefix)` | `"usr/local/bin"` | `"/usr/local/bin"` (example in both SRFI-13 and Kaappi docs) |
| `(string-join '("a" "b" "c") ";" 'suffix)` | `"a;b;c"` | `"a;b;c;"` (example in Kaappi docs) |
| `(string-join '() ";" 'strict-infix)` | `""` | error (SRFI-13: "it is an error to pass an empty list of strings to string-join" with strict-infix grammar) |

## Root cause

`stringJoinFn` implements only the infix grammar with an empty-string fallback delimiter. The `grammar` parameter from the SRFI-13 signature `(string-join string-list [delimiter grammar])` was never wired up: no branch inspects `args[2]`, no leading/trailing delimiter is emitted, and the empty-list + strict-infix error case is unreachable. The `""` default also contradicts the spec'd `" "` default.

Note: `string-join` is exported via `(srfi 13)` (`src/library.zig`), so portable SRFI-13 code silently gets wrong results rather than an unbound-variable or arity error.

---
*This issue was found by automated code analysis (Claude).*

baijum:
## Summary

- Implement the optional `grammar` argument for `string-join` per SRFI-13: `'infix` (default), `'strict-infix`, `'prefix`, and `'suffix`
- `'strict-infix` with an empty list now raises an error instead of silently returning `""`
- Fix stale test expectation for default delimiter (was `""`, already corrected to `" "` in prior work)

## Test plan

- [x] New regression test `tests/scheme/smoke/string-join-grammar.scm` — 21 cases covering all grammars, edge cases (empty list, single element, empty delimiter, multi-char delimiter)
- [x] Uncommented 3 previously-disabled audit tests for suffix, prefix, and strict-infix
- [x] Fixed stale coverage test expectation
- [x] Full suite: 1796/1796 pass

Closes #825

### DIFF
--- a/src/primitives_string_ext.zig
+++ b/src/primitives_string_ext.zig
@@ -330,15 +330,31 @@ fn stringSplitFn(args: []const Value) PrimitiveError!Value {
     return result;
 }
 
-// (string-join list-of-strings) or (string-join list-of-strings delimiter)
+const Grammar = enum { infix, strict_infix, prefix, suffix };
+
+fn parseGrammar(arg: Value) PrimitiveError!Grammar {
+    if (!types.isSymbol(arg)) return primitives.typeError("string-join", "symbol (infix, strict-infix, prefix, or suffix)", arg);
+    const name = types.symbolName(arg);
+    if (std.mem.eql(u8, name, "infix")) return .infix;
+    if (std.mem.eql(u8, name, "strict-infix")) return .strict_infix;
+    if (std.mem.eql(u8, name, "prefix")) return .prefix;
+    if (std.mem.eql(u8, name, "suffix")) return .suffix;
+    return primitives.typeError("string-join", "symbol (infix, strict-infix, prefix, or suffix)", arg);
+}
+
+// (string-join list [delimiter [grammar]])
 fn stringJoinFn(args: []const Value) PrimitiveError!Value {
     const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
     const delim: []const u8 = if (args.len > 1)
         (try getStringSlice("string-join", args[1]))
     else
         " ";
+    const grammar: Grammar = if (args.len > 2)
+        try parseGrammar(args[2])
+    else
+        .infix;
 
-    // Calculate total length
+    // Count elements and total string length
     var total: usize = 0;
     var count: usize = 0;
     var current = args[0];
@@ -349,8 +365,20 @@ fn stringJoinFn(args: []const Value) PrimitiveError!Value {
         count += 1;
         current = types.cdr(current);
     }
-    if (count == 0) return gc.allocString("") catch return PrimitiveError.OutOfMemory;
-    total += (count - 1) * delim.len;
+
+    if (count == 0) {
+        if (grammar == .strict_infix) {
+            const vm = vm_mod.vm_instance orelse return PrimitiveError.InvalidArgument;
+            vm.setErrorDetail("string-join: strict-infix grammar requires a non-empty list", .{});
+            return PrimitiveError.InvalidArgument;
+        }
+        return gc.allocString("") catch return PrimitiveError.OutOfMemory;
+    }
+
+    total += switch (grammar) {
+        .infix, .strict_infix => (count - 1) * delim.len,
+        .prefix, .suffix => count * delim.len,
+    };
 
     const buf = gc.allocator.alloc(u8, total) catch return PrimitiveError.OutOfMemory;
     defer gc.allocator.free(buf);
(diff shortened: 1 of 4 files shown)


baijum (commented):
## Review — ready to merge ✅

Clean, focused fix that fully resolves #825. I built the branch locally and verified the behavior end-to-end.

### Correctness
- All four grammars produce spec-correct output, confirmed against the canonical SRFI-13 examples:
  - `(string-join '("foo" "bar" "baz") ":" 'prefix)` → `":foo:bar:baz"`
  - `… 'suffix` → `"foo:bar:baz:"`, `… 'infix` → `"foo:bar:baz"`
  - empty list → `""` for infix/prefix/suffix; **error** for `strict-infix`
- The `strict-infix` empty-list error is a proper Scheme condition (catchable by `guard`) and carries a clear message: `"string-join: strict-infix grammar requires a non-empty list"`. An unknown grammar symbol (e.g. `'bogus`) also raises a catchable error rather than being silently ignored.
- The pre-sizing math matches the fill loop exactly for every grammar (`(count-1)*delim` for infix/strict-infix, `count*delim` for prefix/suffix), so there's no risk of over/under-allocating `buf`. Empty-delimiter and empty-string edge cases collapse to `total == 0` correctly.

### Tests verified locally
- `tests/scheme/smoke/string-join-grammar.scm` — 21/21 pass. Good coverage: all grammars, single-element, empty list, empty delimiter, multi-char delimiter.
- Re-enabled audit cases in `primitives_string_ext-audit.scm` — 87/87 pass.
- Corrected coverage expectation (`"a b c"` for the default single-space delimiter) — 97/97 pass.
- Build is clean.

The stale coverage-test fix is consistent with the PR description: the `" "` default delimiter was already in place from prior work, and this only updates the lingering expectation.

### Minor, non-blocking nits (optional)
1. The three delimiter `@memcpy` blocks are near-identical; a tiny local helper (`emitDelim(buf, &pos, delim)`) would DRY them up. Purely stylistic — current form is readable.
2. `parseGrammar` returns `typeError(... "symbol (infix, strict-infix, prefix, or suffix)")` for a symbol that isn't a valid grammar (e.g. `'bogus`). The message reads slightly oddly since the arg *is* a symbol — a dedicated "invalid grammar" wording would be marginally clearer. Behavior is correct (raises a catchable error).
3. `string-join` stays registered as `.{ .variadic = 1 }`, so a 4th+ argument is silently ignored (SRFI-13's max arity is 3). Pre-existing behavior shared by other variadic prims, not a regression — noting only for completeness.

None of these block merge. LGTM.

PR #1315, opened 2026-07-08
Add hash-table-update! procedure to SRFI-69 (#1182)

### LINKED ISSUE #1182: [SRFI-69] hash-table-update! missing from (srfi 69)
**Category:** SRFI
**Spec reference:** SRFI-69 — `(hash-table-update! hash-table key function [thunk])` is a required procedure ("Semantically equivalent to, but may be implemented more efficiently than ... (hash-table-set! hash-table key (function (hash-table-ref hash-table key thunk)))").
**Severity:** missing-feature

**Minimal reproduction:**
```scheme
(import (srfi 69))
(define ht (make-hash-table))
(hash-table-set! ht 'k 10)
(hash-table-update! ht 'k (lambda (x) (+ x 1)))
;; error: undefined variable 'hash-table-update!'
```

**Expected:** `(hash-table-ref ht 'k)` ⟹ 11
**Actual:** undefined variable. Only `hash-table-update!/default` is registered in src/primitives_hashtable.zig. A Scheme-level `hash-table-update!` exists in `lib/srfi/125.sld` (SRFI-125), which is why grep finds the name — but `(srfi 69)` does not provide it.

Fix is small: register a variant that errors (or calls the optional thunk) when the key is absent — `hashTableUpdateDefaultFn` already contains the machinery.

**Verified:** fresh ReleaseSafe build at b170e6f
**Source location:** src/primitives_hashtable.zig (specs table)
**Disabled test:** tests/scheme/audit/primitives_hashtable-audit.scm (marked `;; FAIL: #this`)

baijum:
## Summary
- Register `hash-table-update!` as a built-in SRFI-69 procedure with signature `(ht key function [thunk])`
- When the key exists, applies `function` to the current value and stores the result
- When the key is absent, calls the optional `thunk` to get a seed value (or errors if no thunk)
- Enables the previously disabled audit test for this procedure

Closes #1182

## Test plan
- [x] New regression test (`tests/scheme/smoke/hash-table-update-1182.scm`) — key exists, absent+thunk, absent+no-thunk
- [x] Audit test (`tests/scheme/audit/primitives_hashtable-audit.scm`) — 73 pass, 0 fail

### DIFF
--- a/src/primitives_hashtable.zig
+++ b/src/primitives_hashtable.zig
@@ -24,6 +24,7 @@ pub const specs = [_]primitives.PrimSpec{
     .{ .name = "hash-table->alist", .func = &hashTableToAlistFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_69) },
     .{ .name = "alist->hash-table", .func = &alistToHashTableFn, .arity = .{ .variadic = 1 }, .libs = LS.initOne(.srfi_69) },
     .{ .name = "hash-table-copy", .func = &hashTableCopyFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_69) },
+    .{ .name = "hash-table-update!", .func = &hashTableUpdateFn, .arity = .{ .variadic = 3 }, .libs = LS.initOne(.srfi_69) },
     .{ .name = "hash-table-update!/default", .func = &hashTableUpdateDefaultFn, .arity = .{ .exact = 4 }, .libs = LS.initOne(.srfi_69) },
     .{ .name = "hash", .func = &hashFn, .arity = .{ .variadic = 1 }, .libs = LS.initOne(.srfi_69) },
     .{ .name = "string-hash", .func = &stringHashFn, .arity = .{ .variadic = 1 }, .libs = LS.initOne(.srfi_69) },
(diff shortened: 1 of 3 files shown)


baijum (commented):
## Review — ready to merge ✅

Reviewed the implementation, ran the tests, and stress-tested the GC-sensitive path. This is a clean, correct fix for #1182. Recommend merge.

### What's good
- **Correct SRFI-69 semantics.** `(hash-table-update! ht key function [thunk])` matches the spec's reference equivalence `(hash-table-set! ht key (function (hash-table-ref ht key thunk)))`: present → apply `function` to the current value; absent + `thunk` → apply `function` to `(thunk)`; absent + no thunk → error. Verified all three paths plus the `hash-table-update!/default` variant still passes.
- **Mirrors the vetted sibling.** The body is structurally identical to `hashTableUpdateFn`'s neighbor `hashTableUpdateDefaultFn` — `growIfNeeded` → `findSlot` **after** the callbacks (so a stale index isn't held across a re-entrant mutation), write barriers on both key and value, correct `count` bump on insert.
- **Registration is complete.** `.libs = LS.initOne(.srfi_69)` in `specs` is the full export mechanism for `(srfi 69)` — no separate `library.zig` list needs touching.
- **No collision** with SRFI-125's Scheme-level `hash-table-update!` (verified `(srfi 125)` still works).

### Verification I ran
- `zig build test` — pass
- New smoke test `hash-table-update-1182.scm` — 3/3 pass
- Audit test — 73 pass, 0 fail (was 72; the newly-enabled test accounts for the +1)
- Existing `srfi69*.scm`, `hashtable-*.scm`, coverage tests — no regressions
- **GC:** hammered the one genuinely-new path (thunk returns a *fresh heap* value, updater also allocates) under `-Dgc-stress=true` in both ReleaseSafe (500 iters) and Debug (poisoned-free) — clean. `old_val` from the thunk stays reachable because `callWithArgs` copies args into the rooted register file before any bytecode allocation, so no extra `pushRoot` is needed.

### Optional nit (non-blocking)
- **Arity is `.variadic = 3`**, so a 5th+ argument is silently ignored rather than rejected (`(ht key function [thunk])` tops out at 4). The sibling `hash-table-update!/default` uses `.exact = 4`. Not worth blocking on — `hash-table-ref` is equally lax — but capping at 4 would be marginally stricter.

Note: the two `catch |err| { return err; }` blocks are no-op passthroughs (could be `try`), but they match the existing sibling's style, so leaving them is the consistent choice.

PR #1316, opened 2026-07-08
Fix SRFI-151 bit-argument API mismatch (#1233)

### LINKED ISSUE #1233: [SRFI-151] Bit-argument API mismatch: copy-bit rejects booleans, make-bitwise-generator yields 0/1; bits->list/bits->vector require len; bitwise-eqv fixed 2-arity
**Category:** SRFI (lib/srfi/151.sld) — distinct from the negative-operand arithmetic bug (see #1214)
**Spec reference:** SRFI-151:
- copy-bit: "(copy-bit index i boolean) — Returns an integer the same as i except in the index-th bit, which is 1 if boolean is #t and 0 if boolean is #f."
- "(bits->list i [ len ])", "(bits->vector i [ len ])" — len is optional.
- "(bitwise-eqv i ...)" — n-ary.
- make-bitwise-generator: "Returns a SRFI 121 generator that generates all the bits of i" — bits are booleans throughout SRFI-151.
**Severity:** API mismatch (copy-bit unusable with spec-conforming arguments)

```scheme
(import (scheme base) (srfi 151))
(copy-bit 2 0 #t)          ;; expected 4    actual error ((= bit 0) on a boolean)
(copy-bit 2 5 #f)          ;; expected 1    actual error
(bits->list 5)             ;; expected (#t #f #t)   actual arity error (len required)
(bits->vector 5)           ;; expected #(#t #f #t)  actual arity error
(bitwise-eqv)              ;; expected -1   actual arity error
(bitwise-eqv 5)            ;; expected 5    actual arity error
(let ((g (make-bitwise-generator 5))) (g))  ;; expected #t   actual 1
```

**Root cause:** lib/srfi/151.sld: copy-bit tests `(= bit 0)` (numeric protocol from SRFI-60's copy-bit, which took a boolean there too); bits->list/bits->vector have mandatory len parameters; bitwise-eqv is a fixed 2-arg lambda; the generator returns fixnum 0/1 and never terminates its bit stream with booleans.

**Conforms (positive operands):** and/ior/xor n-ary + identities, nand/nor/eqv 2-arg values, bit-count/integer-length including negatives, first-set-bit, bit-field family reads, rotate/reverse at offset 0, bits, list->bits, vector->bits, bitwise-fold/for-each/unfold. Everything driven through bitwise-not with negative intermediates is tracked by #1214.

**Disabled tests:** tests/scheme/srfi/srfi151.scm (marked ;; FAIL: #this and ;; FAIL: #1214)

baijum:
## Summary
- **copy-bit**: accept boolean `#t`/`#f` as third argument (was rejecting booleans with `(= bit 0)`)
- **bitwise-eqv**: make n-ary with identity `-1` (was fixed 2-arity lambda)
- **bits->list / bits->vector**: make `len` optional via `case-lambda` (defaults to `integer-length`)
- **make-bitwise-generator**: return booleans `#t`/`#f` (was returning fixnums `0`/`1`)
- Update all internal callers of `copy-bit` (`bit-swap`, `list->bits`, `vector->bits`, `bitwise-unfold`) to pass booleans

Closes #1233

## Test plan
- [x] All 7 reproductions from the issue now produce expected results
- [x] SRFI-151 test suite: 107 pass, 0 fail (enabled 9 previously disabled tests + added 4 new edge cases)
- [x] R7RS test suite: no regressions
- [x] SRFI-143 test suite: no regressions

### DIFF
--- a/lib/srfi/151.sld
+++ b/lib/srfi/151.sld
@@ -61,7 +61,11 @@
             (if (null? rest) result
                 (loop (%bitwise-xor2 result (car rest)) (cdr rest))))))
 
-    (define (bitwise-eqv a b) (bitwise-not (bitwise-xor a b)))
+    (define (bitwise-eqv . args)
+      (if (null? args) -1
+          (let loop ((result (car args)) (rest (cdr args)))
+            (if (null? rest) result
+                (loop (bitwise-not (bitwise-xor result (car rest))) (cdr rest))))))
     (define (bitwise-nand a b) (bitwise-not (bitwise-and a b)))
     (define (bitwise-nor a b) (bitwise-not (bitwise-ior a b)))
     (define (bitwise-andc1 a b) (bitwise-and (bitwise-not a) b))
@@ -91,13 +95,13 @@
     (define (bit-set? index n) (odd? (arithmetic-shift n (- index))))
 
     (define (copy-bit index n bit)
-      (if (= bit 0)
-          (bitwise-and n (bitwise-not (arithmetic-shift 1 index)))
-          (bitwise-ior n (arithmetic-shift 1 index))))
+      (if bit
+          (bitwise-ior n (arithmetic-shift 1 index))
+          (bitwise-and n (bitwise-not (arithmetic-shift 1 index)))))
 
     (define (bit-swap i j n)
-      (let ((bi (if (bit-set? i n) 1 0))
-            (bj (if (bit-set? j n) 1 0)))
+      (let ((bi (bit-set? i n))
+            (bj (bit-set? j n)))
         (copy-bit i (copy-bit j n bi) bj)))
 
     (define (any-bit-set? mask n) (not (= 0 (bitwise-and mask n))))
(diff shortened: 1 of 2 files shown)


baijum (commented):
## Review — SRFI-151 bit-argument API fix

**Verdict: ready to merge.** ✅ The change is correct, focused, and spec-conformant.

### What I verified (checked out `4cd5d56`, ReleaseSafe build)
- **All 7 issue reproductions** now return the expected values (`copy-bit 2 0 #t` → 4, `bits->list 5` → `(#t #f #t)`, `bitwise-eqv` → -1, generator → `#t`, etc.).
- **SRFI-151 suite:** 107 pass / 0 fail.
- **No regressions:** SRFI-143 65/0, SRFI-60 smoke (imports 151) OK, R7RS suite 0 failures, `zig build test` clean.

### Correctness notes
- `copy-bit`, `make-bitwise-generator`, and the `case-lambda` optional-`len` clauses all match the SRFI-151 reference implementation. The generator is now correctly infinite and also handles **negative** inputs right (`(make-bitwise-generator -1)` yields `#t #t #t …`), which the old `(= x 0)` termination could not.
- `bitwise-eqv` as a left fold from the first argument is equivalent to the reference impl's fold from the `-1` identity (since `-1` is the left identity of XNOR), so 3-ary+ semantics are correct too — e.g. `(bitwise-eqv 1 1 1)` → 1.
- All internal callers (`bit-swap`, `list->bits`, `vector->bits`, `bitwise-unfold`) were updated to pass booleans. No external callers of these procedures exist elsewhere in the tree.

### Optional nits (non-blocking)
- The new tests cover `bitwise-eqv` at 0/1/2 args. Consider adding one 3-arg case (e.g. `(test 1 (bitwise-eqv 1 1 1))`) to lock in the n-ary fold semantics against future refactors.

### Out of scope — sibling bug worth a follow-up
`lib/srfi/143.sld:101` `fxcopy-bit` has the identical numeric-protocol bug (`(if (= bit 0) …)`). SRFI-143 defines `fxcopy-bit` as "the same as SRFI 151 `copy-bit`", so its third argument should be a boolean. It's currently untested. Not part of #1233 — recommend a separate issue/PR.

PR #1317, opened 2026-07-08
Fix string-contains and string-replace start2/end2 handling (#1158)

### LINKED ISSUE #1158: [SRFI-13] string-contains silently ignores start2/end2 (wrong result); string-replace rejects them (arity error)
**Category:** SRFI-13
**Spec reference:** SRFI-13 — `string-contains s1 s2 [start1 end1 start2 end2] -> integer or false`: "The optional start/end indices restrict the operation to the indicated substrings." And `string-replace s1 s2 start1 end1 [start2 end2] -> string`: "The optional start2/end2 parameters default to the full range of s2."
**Severity:** wrong-result (contains) / missing-feature (replace)

**Minimal reproduction:**
```scheme
(import (srfi 13))
;; start2/end2 restrict the needle to "bc":
(string-contains "abc" "xbcx" 0 3 1 3)   ;; expected 1, actual #f (whole "xbcx" searched)
;; splice s2[1,2) into s1[1,3):
(string-replace "abcd" "XYZ" 1 3 1 2)    ;; expected "aYd", actual arity error
```

**Root cause:** `src/primitives_string_ext.zig` — `stringContainsFn` parses start1/end1 via `parseStartEnd(full_s1, args, 2)` but never reads args[4]/args[5], silently searching for the whole needle (wrong result, not an error). `stringReplaceFn` is registered `.{ .exact = 4 }`, so the optional start2/end2 are rejected at arity check. (`string-prefix?`/`string-suffix?` handle all six arguments correctly — the s2-range plumbing exists and can be reused.)

**Disabled test:** `tests/scheme/audit/primitives_string_ext-audit.scm` (marked `;; FAIL: #this`)

**Found by:** Systematic audit, Phase 2.2 (#1137)

baijum:
## Summary
- **string-contains**: now respects optional `start2`/`end2` arguments by applying `parseStartEnd` on s2 at arg index 4 (previously silently ignored, searching the whole needle)
- **string-replace**: changed arity from `exact=4` to `variadic=4` and added `parseStartEnd` on s2 at arg index 4 (previously rejected optional args with arity error)
- Both follow the same pattern already used by `string-prefix?` and `string-suffix?`

## Test plan
- [x] Re-enabled 2 disabled audit assertions in `primitives_string_ext-audit.scm` (89/89 pass)
- [x] Added `tests/scheme/smoke/srfi13-start2-end2.scm` with 10 regression tests covering both procedures with and without start2/end2
- [x] Zig unit tests pass
- [x] Issue reproduction cases produce correct results: `(string-contains "abc" "xbcx" 0 3 1 3)` → `1`, `(string-replace "abcd" "XYZ" 1 3 1 2)` → `"aYd"`

Closes #1158

### DIFF
--- a/src/primitives_string_ext.zig
+++ b/src/primitives_string_ext.zig
@@ -36,7 +36,7 @@ pub const specs = [_]primitives.PrimSpec{
     .{ .name = "string-reverse", .func = &stringReverseFn, .arity = .{ .variadic = 1 }, .libs = LS.initOne(.srfi_13) },
     .{ .name = "string-filter", .func = &stringFilterFn, .arity = .{ .variadic = 2 }, .libs = LS.initOne(.srfi_13) },
     .{ .name = "string-delete", .func = &stringDeleteFn, .arity = .{ .variadic = 2 }, .libs = LS.initOne(.srfi_13) },
-    .{ .name = "string-replace", .func = &stringReplaceFn, .arity = .{ .exact = 4 }, .libs = LS.initOne(.srfi_13) },
+    .{ .name = "string-replace", .func = &stringReplaceFn, .arity = .{ .variadic = 4 }, .libs = LS.initOne(.srfi_13) },
     .{ .name = "string-titlecase", .func = &stringTitlecaseFn, .arity = .{ .variadic = 1 }, .libs = LS.initOne(.srfi_13) },
     .{ .name = "string-every", .func = &stringEveryFn, .arity = .{ .variadic = 2 }, .libs = LS.initOne(.srfi_13) },
     .{ .name = "string-any", .func = &stringAnyFn, .arity = .{ .variadic = 2 }, .libs = LS.initOne(.srfi_13) },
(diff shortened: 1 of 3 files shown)


baijum:
## Review — ✅ Ready to merge

Reviewed the diff and verified behavior against a fresh ReleaseSafe build of the PR head (`a662eccb`).

### Correctness
The fix is minimal and correct, and it reuses the already-tested `parseStartEnd` → `parseOptionalRange` plumbing exactly the way `string-prefix?`/`string-suffix?` do:

- **`string-contains`** now parses the s2 range at arg index 4, so `(string-contains "abc" "xbcx" 0 3 1 3)` → `1` (was `#f`).
- **`string-replace`** arity `exact=4` → `variadic=4` plus the s2-range parse, so `(string-replace "abcd" "XYZ" 1 3 1 2)` → `"aYd"` (was an arity error).

Because `parseOptionalRange` defaults `start2=0`/`end2=len` when the args are absent, the change is fully backward-compatible — the 4-arg `string-replace` and 2/3/4-arg `string-contains` forms behave exactly as before.

### Verification I ran
- New smoke test `srfi13-start2-end2.scm`: **10/10 pass**
- Re-enabled audit assertions: `primitives_string_ext-audit.scm` **89/89 pass**
- `zig build test`: pass
- No regressions in the full smoke suite, `srfi13-startend.scm` (46/46), `srfi13-ext.scm` (50/50), and the string compliance suites
- Extra edge cases all correct: `start2`-only 5-arg form (`end2` defaults to full), empty needle via equal `start2`/`end2` returns `start1`, out-of-range `start2` raises, needle-range longer than haystack-range returns `#f`

### Minor note (non-blocking)
`variadic=4` means `string-replace` now silently accepts 7+ arguments (SRFI-13 tops out at 6). That's consistent with the existing convention in this file — `string-prefix?`, `string-suffix?`, and `string-contains` are all `variadic` and ignore surplus args — so it's not a new problem, just noting it. No change requested.

LGTM. 🚀

PR #1318, opened 2026-07-08
Fix SRFI-210 value procedure and add box/mv export (#1218)

### LINKED ISSUE #1218: [SRFI-210] value returns its first argument instead of the index-th object; box/mv not exported
**Category:** SRFI
**Spec reference:** SRFI-210: "(value index obj0 … objn-1) — It is an error if index is not an exact non-negative integer less than n. The value procedure simply returns its argument objindex." Also the export list includes the box/mv syntax: "(box/mv element1 … producer) — yields a SRFI 195 box constructed from the single values of the elements and the multiple values of the producer".
**Severity:** wrong-result

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 210))
(value 1 'a 'b)   ;; expected: b   actual: 1
(value 0 'x)      ;; expected: x   actual: 0
```

**Root cause:** lib/srfi/210.sld defines `(define (value obj . rest) (if (null? rest) obj (list-ref (cons obj rest) 0)))` — both branches return the first argument (the index). It should be `(list-ref rest index)`. `box/mv` is absent from the export list.

**Source location:** lib/srfi/210.sld
**Disabled test:** tests/scheme/srfi/srfi210.scm (marked ;; FAIL: #this)

baijum:
## Summary
- Fix `value` procedure which returned its first argument (the index) instead of the index-th object — `(list-ref (cons obj rest) 0)` always yielded `obj` regardless of index
- Add `box/mv` syntax definition and export — was missing from both implementation and export list
- Enable previously disabled FAIL-marked tests and add an additional edge case

Closes #1218

## Test plan
- [x] All 3 previously disabled `#1218` tests now pass
- [x] Added `(value 2 'a 'b 'c)` edge case
- [x] Full SRFI-210 suite: 30 pass, 0 fail
- [x] Full test suite: 1800 pass, 0 fail

### DIFF
--- a/lib/srfi/210.sld
+++ b/lib/srfi/210.sld
@@ -3,7 +3,7 @@
   (export with-values case-receive set!-values
           apply/mv call/mv list/mv vector/mv value/mv coarity
           list-values vector-values box-values
-          value identity compose-left compose-right
+          value identity compose-left compose-right box/mv
           map-values bind bind/list bind/box bind/mv)
   (begin
 
@@ -61,6 +61,11 @@
         ((coarity producer)
          (call-with-values (lambda () producer) (lambda args (length args))))))
 
+    (define-syntax box/mv
+      (syntax-rules ()
+        ((box/mv element ... producer)
+         (apply box (list/mv element ... producer)))))
+
     (define-syntax bind/mv
       (syntax-rules ()
         ((bind/mv producer transducer)
@@ -79,9 +84,8 @@
     (define (box-values b)
       (unbox b))
 
-    (define (value obj . rest)
-      (if (null? rest) obj
-          (list-ref (cons obj rest) 0)))
+    (define (value index . objs)
+      (list-ref objs index))
 
     (define (identity . objs)
       (apply values objs))
--- a/tests/scheme/srfi/srfi210.scm
+++ b/tests/scheme/srfi/srfi210.scm
@@ -50,11 +50,11 @@
 (test 3 (bind/box (box (values 1 2)) +))
 
 ;;; --- value: (value index obj0 ... objn-1) returns obj_index ---
-;; FAIL: #1218 (value returns its first argument instead of the index-th object)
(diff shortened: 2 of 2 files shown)


baijum (commented):
## Review — ready to merge ✅

Verified against the SRFI-210 spec and issue #1218. Both fixes are correct and the tests pass.

### `value` fix — correct
`(define (value index . objs) (list-ref objs index))` exactly implements the spec: *"(value index obj0 … objn-1) … returns its argument obj_index."* The old code returned `(list-ref (cons obj rest) 0)`, which always yielded the first argument (the index) regardless of index. Confirmed:
- `(value 1 'a 'b)` → `b`
- `(value 0 'x)` → `x`
- `(value 2 'a 'b 'c)` → `c`

### `box/mv` — correct
`(apply box (list/mv element ... producer))` matches the spec (a SRFI-195 box built from the single values of the elements plus the multiple values of the producer), and `box/mv` is now in the export list. Confirmed `(unbox (box/mv 1 (values 2 3)))` yields values `1 2 3` and the result is a `box?`.

### Tests — good
Nice catch fixing the disabled `box/mv` test: the original `(test '(1 2 3) (unbox (box/mv ...)))` would have been wrong since `unbox` returns *multiple values*, not a list — the new `with-values (unbox ...) list` form is right. The added `(value 2 'a 'b 'c)` edge case is a good third data point.

Local verification (fresh build): full SRFI-210 suite **30 pass, 0 fail**. Only `tests/scheme/srfi/srfi210.scm` imports `(srfi 210)`, so there is no regression surface elsewhere.

### Minor nit (non-blocking)
The commit body says box/mv *"was implemented but not exported"* — but the diff shows the `define-syntax box/mv` being **added**, so it was neither implemented nor exported before. The PR description gets it right ("missing from both implementation and export list"); just the commit message is slightly off. Not worth a re-push.

**LGTM — good to merge.**

PR #1319, opened 2026-07-08
Fix SRFI-27 random-integer, %rs-next-int, and pseudo-randomize! to accept bignums (#1193)

### LINKED ISSUE #1193: [SRFI-27] random-integer, random-source-make-integers rand, and random-source-pseudo-randomize! reject exact integers wider than fixnums
**Category:** SRFI
**Spec reference:** SRFI-27, Specification: "(random-integer n) — The next integer in {0, ..., n-1} ... The argument n must be a positive integer, otherwise an error is signalled." `(expt 2 64)` is a positive integer, so it must be accepted. Likewise "(random-source-pseudo-randomize! s i j) — Changes the state of the random source s into the initial state of the (i, j)-th independent random source, where i and j are non-negative integers" — no size bound.
**Severity:** missing-feature

**Minimal reproduction:**
```scheme
(random-integer (expt 2 64))
;; expected: an integer in [0, 2^64)  actual: type error in 'random-integer': expected integer, got #<bignum>

((random-source-make-integers (make-random-source)) (expt 2 64))
;; actual: type error in '%rs-next-int': expected integer, got #<bignum>

(random-source-pseudo-randomize! (make-random-source) (expt 2 64) 1)
;; actual: type error in 'random-source-pseudo-randomize!': expected integer, got #<bignum>
```

**Root cause:** all three primitives use `types.isFixnum` checks only — no bignum dispatch (`randomIntegerFn`, `rsNextIntFn`, `randomSourcePseudoRandomizeFn` in src/primitives_random.zig). Fixnum ranges up to 2^47 work.

For bignum n, the standard approach is rejection sampling over ceil(bits(n)/64) random limbs; for pseudo-randomize! the i/j values can be reduced (e.g. hashed/mod-folded) into the seed computation.

**Source location:** src/primitives_random.zig (`randomIntegerFn`, `rsNextIntFn`, `randomSourcePseudoRandomizeFn`)
**Disabled test:** tests/scheme/audit/primitives_random-audit.scm (3 assertions marked ;; FAIL: #this)

baijum:
## Summary
- Add bignum support to `random-integer` and `%rs-next-int` via rejection sampling over random limbs, masked to the bound's bit width
- Add bignum support to `random-source-pseudo-randomize!` by XOR-folding bignum limbs to u64 seed values
- Extract shared `randomBelow` and `intToSeedU64` helpers to deduplicate fixnum/bignum dispatch

## Test plan
- [x] Enable 3 `FAIL`-marked assertions in `primitives_random-audit.scm` (52/52 pass)
- [x] Add `tests/scheme/smoke/srfi27-bignum.scm` regression test (6 tests covering all three procedures with `(expt 2 64)` and `(expt 2 128)`)
- [x] `zig build test` passes
- [x] All three reproduction cases from the issue produce correct results

Closes #1193

### DIFF
--- a/src/primitives_random.zig
+++ b/src/primitives_random.zig
@@ -55,14 +55,10 @@ fn getRS(proc: []const u8, v: Value) PrimitiveError!*types.RandomSource {
 }
 
 fn randomIntegerFn(args: []const Value) PrimitiveError!Value {
-    if (!types.isFixnum(args[0])) return primitives.typeError("random-integer", "integer", args[0]);
-    const n = types.toFixnum(args[0]);
-    if (n <= 0) return primitives.typeError("random-integer", "positive integer", args[0]);
     ensureDefaultRS();
     const vm = vm_mod.vm_instance orelse return PrimitiveError.OutOfMemory;
     const rs = try getRS("random-integer", vm.default_random_source);
-    const r = rs.prng.random();
-    return types.makeFixnum(r.intRangeLessThan(i64, 0, n));
+    return randomBelow("random-integer", rs, args[0]);
 }
 
 fn randomRealFn(args: []const Value) PrimitiveError!Value {
@@ -98,15 +94,10 @@ fn randomSourceRandomizeFn(args: []const Value) PrimitiveError!Value {
 }
 
 fn randomSourcePseudoRandomizeFn(args: []const Value) PrimitiveError!Value {
-    const rs = try getRS("random-source-pseudo-randomize!", args[0]);
-    if (!types.isFixnum(args[1])) return primitives.typeError("random-source-pseudo-randomize!", "integer", args[1]);
-    if (!types.isFixnum(args[2])) return primitives.typeError("random-source-pseudo-randomize!", "integer", args[2]);
-    const i_val = types.toFixnum(args[1]);
-    const j_val = types.toFixnum(args[2]);
-    if (i_val < 0) return primitives.typeError("random-source-pseudo-randomize!", "non-negative integer", args[1]);
-    if (j_val < 0) return primitives.typeError("random-source-pseudo-randomize!", "non-negative integer", args[2]);
-    const i: u64 = @intCast(i_val);
-    const j: u64 = @intCast(j_val);
+    const proc = "random-source-pseudo-randomize!";
+    const rs = try getRS(proc, args[0]);
+    const i = try intToSeedU64(proc, args[1]);
+    const j = try intToSeedU64(proc, args[2]);
     rs.prng = std.Random.DefaultPrng.init(i *% 2654435761 +% j *% 2246822519);
     return types.VOID;
 }
@@ -162,11 +153,7 @@ fn randomSourceStateSetFn(args: []const Value) PrimitiveError!Value {
 // (%rs-next-int rs n) — used by random-source-make-integers closure
 fn rsNextIntFn(args: []const Value) PrimitiveError!Value {
     const rs = try getRS("%rs-next-int", args[0]);
-    if (!types.isFixnum(args[1])) return primitives.typeError("%rs-next-int", "integer", args[1]);
-    const n = types.toFixnum(args[1]);
-    if (n <= 0) return primitives.typeError("%rs-next-int", "positive integer", args[1]);
-    const r = rs.prng.random();
-    return types.makeFixnum(r.intRangeLessThan(i64, 0, n));
+    return randomBelow("%rs-next-int", rs, args[1]);
 }
 
 // (%rs-next-real rs) — used by random-source-make-reals closure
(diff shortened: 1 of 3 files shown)


baijum:
## Review — not ready to merge yet

The approach is sound (rejection sampling over random limbs, bit-masked to the bound's width, with fixnum demotion of small results). But there's a **crashing bug** in `randomBignumBelow` that must be fixed before merge. CodeRabbit already flagged both points below; I built the branch and reproduced them, so this is confirmation with concrete evidence rather than a duplicate.

### 🔴 Blocker — `top_bits: u6` panics for ~half of all bignum bounds

`src/primitives_random.zig:185`
```zig
const top_bits: u6 = @intCast(64 - @clz(top_limb));
const mask: u64 = if (top_bits == 64) std.math.maxInt(u64) else (@as(u64, 1) << top_bits) - 1;
```

`@clz(u64)` returns `0` when the top limb's MSB is set, so `64 - @clz(...)` evaluates to `64`, which does not fit in `u6` (max 63). In the default **ReleaseSafe** build this is a hard panic — and the very next line's `if (top_bits == 64)` guard is unreachable/dead, since a `u6` can never equal 64.

This fires for any bound whose most-significant limb has bit 63 set — roughly half of all bignum bounds. Reproduced on a fresh build of this branch:

```scheme
(random-integer (expt 2 127))          ; top limb = 2^63  → panic
(random-integer (- (expt 2 128) 1))    ; all limbs 0xFFFF… → panic
((random-source-make-integers (make-random-source)) (expt 2 127))  ; %rs-next-int path → panic
```
```
thread panic: integer does not fit in destination type
  src/primitives_random.zig:185:26 in randomBignumBelow
```
(`(expt 2 64)` and `(expt 2 128)` do *not* trip it — their top limb is exactly `1`, i.e. `top_bits == 1`.)

Fix (verified — the three cases above return correct results and all existing tests still pass):
```zig
const top_bits: u7 = @intCast(64 - @clz(top_limb));
const mask: u64 = if (top_bits >= 64) std.math.maxInt(u64) else (@as(u64, 1) << @as(u6, @intCast(top_bits))) - 1;
```

### 🟠 Test gap — regression tests never exercise the crash

`tests/scheme/smoke/srfi27-bignum.scm` and the enabled audit assertions only use exact powers of two (`(expt 2 64)`, `(expt 2 128)`), whose top limb is always `1`. I confirmed the suite reports **6/6 pass** and the audit **52/52 pass even with the buggy `u6` code in place** — so the tests give false confidence. Please add at least one bound with a saturated top limb, e.g. `(- (expt 2 128) 1)` and `(expt 2 127)`, to both files.

### Minor (non-blocking)

- **`randomBignumBelow` exhaustion → `OutOfMemory`** (line ~211): after 1000 rejected attempts it returns `PrimitiveError.OutOfMemory`, which would surface as a misleading "out of memory". It's statistically unreachable (acceptance ≥ ½, so ≤ 2⁻¹⁰⁰⁰), so this is cosmetic — but a comment noting it's an unreachable safety valve would help the next reader.
- **`intToSeedU64` XOR-folds bignum limbs** into the u64 seed. Distinct large `i`/`j` values can collide to the same seed, so `pseudo-randomize!` streams aren't guaranteed independent for bignum indices. That's an acceptable, spec-tolerated reduction (the issue explicitly allows folding) and matches the existing fixnum behavior — just flagging it as a known limitation.

### Otherwise looks good

- GC-safe: the only allocation (`allocBignumFromLimbs`) copies the temp buffer and runs after the last read of `bn`; the >16-limb heap buffer is correctly `defer`-freed.
- Rejection-sampling math is correct once the type is fixed, and the fixnum/zero demotion preserves the fixnum-canonicalization invariant.
- CI is green, but note that's expected given the test gap above — green CI does not cover this bug.

Once the `u6`→`u7` fix lands with a saturated-top-limb regression test, this is good to merge.

*Reviewed with Claude Code.*

baijum:
## Re-review of 66229c3 — issues resolved ✅

The follow-up commit addresses both findings. Verified on a local ReleaseSafe build of the branch:

**🔴 Blocker fixed.** `top_bits` is now `u7` with the shift operand narrowed to `u6` only in the safe branch, and the guard is `>= 64`. The three cases that panicked before now return correct results:
```
(random-integer (expt 2 127))         → 130069292547186676219478961986160239583  (was: panic)
(random-integer (- (expt 2 128) 1))   → in [0, bound)                              (was: panic)
((random-source-make-integers (make-random-source)) (expt 2 127)) → ok            (was: panic)
```

**🟠 Test gap closed.** Saturated-top-limb bounds (`(expt 2 127)`, `(- (expt 2 128) 1)`) added to both files — these genuinely exercise the crash path (they panic against the old code):
- `tests/scheme/smoke/srfi27-bignum.scm`: **8/8 pass** (was 6)
- `tests/scheme/audit/primitives_random-audit.scm`: **54/54 pass** (was 52)

Also ran locally: `zig build test` passes, and a 20k-draw distribution check on a saturated bound `(- (expt 2 70) 1)` came out in-range with even/odd split ≈ 0.5002 (no visible bias).

The two minor notes from the previous review (unreachable `OutOfMemory` on 1000-attempt exhaustion; XOR-fold seed collisions for bignum `pseudo-randomize!` indices) are non-blocking and can be left as-is.

**LGTM — ready to merge once CI goes green.**

*Reviewed with Claude Code.*

PR #1320, opened 2026-07-08
Fix posix-time/monotonic-time to return SRFI-19 time objects (#1162)

### LINKED ISSUE #1162: [SRFI-170] posix-time/monotonic-time return bare fixnum seconds instead of SRFI-19 time objects (nanoseconds discarded)
**Category:** SRFI-170
**Spec reference:** SRFI-170 — posix-time "returns the current time as a time object of type time-utc"; monotonic-time "returns the current time as a time object of type time-monotonic". Time objects are SRFI-19 compatible.
**Severity:** missing-feature / wrong-result (wrong type, loses sub-second precision)

**Minimal reproduction:**
```scheme
(import (srfi 170) (srfi 19))
(posix-time)                 ;; => 1783241936   (a bare fixnum)
(time? (posix-time))         ;; => #f            expected #t
(time-second (posix-time))   ;; => type error    expected the seconds field
```

**Expected:** SRFI-19 time objects (`time-utc` / `time-monotonic`) with second and nanosecond fields.
**Actual:** bare fixnum whole seconds; nanoseconds discarded even though `clock_gettime` already provides them (`src/primitives_filesystem.zig:975-987` reads `ts.sec` and drops `ts.nsec`).

Portable SRFI-170 code that feeds these values to SRFI-19 procedures (`time-second`, `time-difference`, ...) breaks. Kaappi ships SRFI-19 as a portable library, so the time-object representation exists to integrate with.

**Source location:** `src/primitives_filesystem.zig` `posixTimeFn` / `monotonicTimeFn`.
**Verified:** fresh ReleaseSafe build at 83718b6.

baijum:
## Summary

- Extends the Zig-level `Srfi18Time` type with `nanoseconds: i64` and `time_type: TimeType` fields, unifying it as the canonical time representation for both SRFI-18 and SRFI-19
- Registers SRFI-19 time accessors (`make-time`, `time?`, `time-type`, `time-second`, `time-nanosecond`) as built-in primitives in `(scheme time)`
- Updates `lib/srfi/19.sld` to use the built-in time type instead of `define-record-type <time>`
- Fixes `posix-time` to return a `time-utc` object and `monotonic-time` to return a `time-monotonic` object, both with nanosecond precision
- Updates `set-file-times` to accept time objects in addition to fixnums

Closes #1162

## Test plan

- [x] `zig build test` — 667 unit tests pass (0 fail)
- [x] `bash tests/scheme/run-all.sh` — 1802 tests pass (0 fail)
- [x] SRFI-19 test suite — 112/112 pass
- [x] New regression test `tests/scheme/smoke/srfi170-time-objects.scm` — 16 assertions
- [x] Manual REPL verification: `(time? (posix-time))` → `#t`, `(time-type (posix-time))` → `time-utc`

### DIFF
--- a/src/primitives_srfi18.zig
+++ b/src/primitives_srfi18.zig
@@ -37,13 +37,8 @@ pub const specs = [_]primitives.PrimSpec{
     .{ .name = "condition-variable-signal!", .func = &condvarSignalFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_18), .sandbox = false, .wasm = false },
     .{ .name = "condition-variable-broadcast!", .func = &condvarBroadcastFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_18), .sandbox = false, .wasm = false },
     .{ .name = "current-time", .func = &currentTimeFn, .arity = .{ .exact = 0 }, .libs = LS.initOne(.srfi_18), .sandbox = false, .wasm = false },
-    .{ .name = "time?", .func = &timePredFn, .arity = .{ .exact = 1 }, .libs = LS.initMany(&.{ .srfi_18, .scheme_time }), .sandbox = false, .wasm = false },
     .{ .name = "time->seconds", .func = &timeToSecondsFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_18), .sandbox = false, .wasm = false },
     .{ .name = "seconds->time", .func = &secondsToTimeFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_18), .sandbox = false, .wasm = false },
-    .{ .name = "make-time", .func = &makeTimeFn, .arity = .{ .exact = 3 }, .libs = LS.initOne(.scheme_time) },
-    .{ .name = "time-type", .func = &timeTypeFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_time) },
-    .{ .name = "time-second", .func = &timeSecondFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_time) },
-    .{ .name = "time-nanosecond", .func = &timeNanosecondFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_time) },
     .{ .name = "join-timeout-exception?", .func = &joinTimeoutPredFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_18), .sandbox = false, .wasm = false },
     .{ .name = "abandoned-mutex-exception?", .func = &abandonedMutexPredFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_18), .sandbox = false, .wasm = false },
     .{ .name = "terminated-thread-exception?", .func = &terminatedThreadPredFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_18), .sandbox = false, .wasm = false },
@@ -121,7 +116,8 @@ fn timeoutToDeadlineNs(timeout: Value) PrimitiveError!?u64 {
     if (types.isSrfi18Time(timeout)) {
         const t = types.toSrfi18Time(timeout);
         if (t.seconds < 0) return 0;
-        const sec_ns: u64 = @as(u64, @intCast(t.seconds)) * 1_000_000_000 + @as(u64, @intCast(t.nanoseconds));
+        const ns_clamped: u64 = if (t.nanoseconds > 0) @intCast(t.nanoseconds) else 0;
+        const sec_ns: u64 = @as(u64, @intCast(t.seconds)) * 1_000_000_000 + ns_clamped;
         var now_ts: std.c.timespec = undefined;
         _ = std.c.clock_gettime(.REALTIME, &now_ts);
         const now_ns = @as(u64, @intCast(now_ts.sec)) * 1_000_000_000 + @as(u64, @intCast(now_ts.nsec));
@@ -962,10 +958,6 @@ fn currentTimeFn(_: []const Value) PrimitiveError!Value {
     return gc.allocSrfi18Time(@intCast(ts.sec), @intCast(ts.nsec), .utc) catch return PrimitiveError.OutOfMemory;
 }
 
-fn timePredFn(args: []const Value) PrimitiveError!Value {
-    return if (types.isSrfi18Time(args[0])) types.TRUE else types.FALSE;
-}
-
 fn timeToSecondsFn(args: []const Value) PrimitiveError!Value {
     if (!types.isSrfi18Time(args[0]))
         return primitives.typeError("time->seconds", "time", args[0]);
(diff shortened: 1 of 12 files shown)


baijum (commented):
## Review — Fix posix-time/monotonic-time to return SRFI-19 time objects (#1162)

The direction is right and the SRFI-19 refactor is clean: unifying `Srfi18Time` as the canonical `(second, nanosecond, type)` representation is the correct model, `set-file-times` accepting time objects is a nice generalization, and the `timeArgToTimespec` helper de-duplicates the atime/mtime code well. Test coverage is good — SRFI-19 suite still 112/112, the new smoke test (16 assertions) passes, and `zig build test` is green locally.

However, I found **two confirmed interpreter-abort crashes** reachable from ordinary Scheme, plus a **WASM availability gap**, so I don't think it's ready to merge as-is.

### Blocking

1. **Interpreter panic when a time has a negative nanosecond field** (two sites). The new code does `@as(u64, @intCast(t.nanoseconds))` in both the printer and the SRFI-18 timeout path. Since `make-time` is a public constructor that accepts any fixnum, both are reachable from pure Scheme and abort the whole process (ReleaseSafe is the default build). This is a *regression* — the old record-based `%make-time` + `#<time {d:.6}>` printer never crashed on any input.

   ```
   $ kaappi -> (write (make-time 'time-utc -1 0))
   thread N panic: integer does not fit in destination type   (src/printer.zig:700)

   $ (thread-join! th (make-time 'time-utc -5 100) 'x)
   thread N panic: integer does not fit in destination type   (src/primitives_srfi18.zig:124)
   ```
   Cleanest root-cause fix: have `make-time` validate `0 <= nanosecond < 1e9` (SRFI-19 defines the nanosecond field as a non-negative count below 10^9) and raise a normal Scheme error — that closes both crash sites at the source. See inline notes.

2. **The new `(scheme time)` accessors are missing on WASM.** `make-time`, `time?`, `time-type`, `time-second`, and `time-nanosecond` are registered from `primitives_srfi18.zig`, whose entire spec list is excluded from WASM builds (`primitives.zig:190`). These are pure functions with no syscalls, yet they vanish on `wasm32-wasi`, making `(scheme time)` inconsistent across targets and breaking any portable code that uses them (including the refactored `srfi/19.sld`, which now hard-depends on them). Verified against a fresh `zig build wasm`:
   ```
   WASM:   (scheme time) make-time  =>  error: undefined variable 'make-time'
   native: (scheme time) make-time  =>  #<procedure>
   ```
   Suggest moving these pure accessors next to `current-second`/`current-jiffy` in `primitives_r7rs.zig` (not WASM-excluded).

### Verification performed
- `zig build test` — pass
- `tests/scheme/smoke/srfi170-time-objects.scm` — 16/16 pass
- `tests/scheme/srfi/srfi-19-tests.scm` — 112/112 pass
- Reproduced both panics on a ReleaseSafe build; confirmed the WASM gap on `zig build wasm` under wasmtime.

Once the negative-nanosecond crashes are guarded (ideally via `make-time` range validation) and the accessors are reachable on WASM, this should be good to go.

src/printer.zig
```diff
@@ -691,7 +691,13 @@ fn printValueWithDepth(writer: anytype, value: Value, mode: PrintMode, depth: u3
...
             .srfi18_time => {
                 const t = obj.as(types.Srfi18Time);
-                try writer.print("#<time {d:.6}>", .{t.seconds});
+                const type_name: []const u8 = switch (t.time_type) {
+                    .utc => "time-utc",
+                    .tai => "time-tai",
+                    .monotonic => "time-monotonic",
+                    .duration => "time-duration",
+                };
+                try writer.print("#<time {s} {d}.{d:0>9}>", .{ type_name, t.seconds, @as(u64, @intCast(t.nanoseconds)) });
```
baijum: **Crash: `@intCast` of a negative nanosecond aborts the interpreter.** `t.nanoseconds` is an `i64` that can be negative (nothing validates the field, and `make-time` accepts any fixnum). In a ReleaseSafe build — the default — `@as(u64, @intCast(-1))` panics and kills the whole process/REPL:

```
(write (make-time 'time-utc -1 0))
=> thread N panic: integer does not fit in destination type
```

The REPL auto-prints results, so just typing `(make-time 'time-utc -1 0)` crashes it. Either validate the nanosecond range in `make-time` (preferred, see the note there) or make the printer robust to negative values (e.g. format the sign explicitly / use `@abs`).

src/primitives_srfi18.zig
```diff
@@ -116,7 +120,8 @@ fn timeoutToDeadlineNs(timeout: Value) PrimitiveError!?u64 {
     if (timeout == types.FALSE) return null;
     if (types.isSrfi18Time(timeout)) {
         const t = types.toSrfi18Time(timeout);
-        const sec_ns: u64 = @intFromFloat(@max(0.0, t.seconds) * 1_000_000_000.0);
+        if (t.seconds < 0) return 0;
+        const sec_ns: u64 = @as(u64, @intCast(t.seconds)) * 1_000_000_000 + @as(u64, @intCast(t.nanoseconds));
```
baijum: **Same negative-nanosecond crash on the timeout path.** `@as(u64, @intCast(t.nanoseconds))` panics when a time object with a negative nanosecond field is passed as a timeout:

```
(thread-join! th (make-time 'time-utc -5 100) 'timed-out)
=> thread N panic: integer does not fit in destination type
```

Validating the field in `make-time` (below) fixes this site too.

src/primitives_srfi18.zig
```diff
@@ -964,14 +969,65 @@ fn timePredFn(args: []const Value) PrimitiveError!Value {
...
+    const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
+    if (!types.isSymbol(args[0]))
+        return primitives.typeError("make-time", "symbol", args[0]);
+    const time_type = parseTimeType(types.symbolName(args[0])) orelse
+        return primitives.typeError("make-time", "time type symbol (time-utc, time-tai, time-monotonic, time-duration)", args[0]);
+    if (!types.isFixnum(args[1]))
+        return primitives.typeError("make-time", "integer", args[1]);
+    if (!types.isFixnum(args[2]))
+        return primitives.typeError("make-time", "integer", args[2]);
+    return gc.allocSrfi18Time(types.toFixnum(args[2]), types.toFixnum(args[1]), time_type) catch return PrimitiveError.OutOfMemory;
```
baijum: `make-time` accepts any fixnum for the nanosecond arg, but SRFI-19 defines the nanosecond field as a non-negative integer below 10^9. Range-checking here (`0 <= args[1] < 1_000_000_000`) and raising a normal Scheme error is the cleanest fix — it prevents the interpreter-abort panics in both the printer (`printer.zig:700`) and the timeout path (`primitives_srfi18.zig:124`) at the source, rather than papering over each consumer.

Minor, related: `seconds->time` uses `@round(frac * 1e9)`, which can round up to exactly `1_000_000_000` and produce an out-of-range nanosecond field for inputs like `0.9999999996`.

src/primitives_srfi18.zig
```diff
@@ -37,9 +37,13 @@ pub const specs = [_]primitives.PrimSpec{
     .{ .name = "condition-variable-signal!", .func = &condvarSignalFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_18), .sandbox = false, .wasm = false },
     .{ .name = "condition-variable-broadcast!", .func = &condvarBroadcastFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_18), .sandbox = false, .wasm = false },
     .{ .name = "current-time", .func = &currentTimeFn, .arity = .{ .exact = 0 }, .libs = LS.initOne(.srfi_18), .sandbox = false, .wasm = false },
-    .{ .name = "time?", .func = &timePredFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_18), .sandbox = false, .wasm = false },
+    .{ .name = "time?", .func = &timePredFn, .arity = .{ .exact = 1 }, .libs = LS.initMany(&.{ .srfi_18, .scheme_time }), .sandbox = false, .wasm = false },
     .{ .name = "time->seconds", .func = &timeToSecondsFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_18), .sandbox = false, .wasm = false },
     .{ .name = "seconds->time", .func = &secondsToTimeFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_18), .sandbox = false, .wasm = false },
+    .{ .name = "make-time", .func = &makeTimeFn, .arity = .{ .exact = 3 }, .libs = LS.initOne(.scheme_time) },
```
baijum: These four accessors (and `time?` above) are pure — no syscalls — but living in `primitives_srfi18.zig` means they're excluded from WASM builds, since `primitives.zig:190` swaps this whole spec list for `no_specs` when `is_wasm`. As a result `(scheme time)` is inconsistent across targets and `srfi/19.sld` (which now depends on these) breaks on the WASM playground. Verified: `make-time` is undefined under `(scheme time)` on a `zig build wasm` binary but bound natively. Consider registering these alongside `current-second`/`current-jiffy` in `primitives_r7rs.zig`, which isn't WASM-excluded.

baijum:
## Re-review of `6aefed7` — all findings addressed ✅

The follow-up commit resolves every item from the previous review. Verified on a fresh build (native + `zig build wasm`):

**1. Negative-nanosecond crashes → fixed at the source.** `make-time` now validates `0 ≤ nanosecond < 1e9` and raises a normal, catchable Scheme error; the printer and timeout paths also guard defensively. Both former panics are gone:
```
(write (make-time 'time-utc -1 0))
  => error: type error in 'make-time': expected nanosecond in [0, 999999999], got -1   (catchable)
(thread-join! th (make-time 'time-utc -5 100) 'x)
  => same clean error, no panic
(guard (e (#t 'caught)) (make-time 'time-utc -1 0)) => caught
```

**2. WASM availability → fixed.** `time?`, `make-time`, `time-type`, `time-second`, `time-nanosecond` moved to `primitives_r7rs.zig` (not WASM-excluded). On a `zig build wasm` binary under wasmtime:
```
(scheme time): (make-time 'time-utc 500 1000) => #<time>, (time-second …) => 1000   ✅
```

**3. `seconds->time` rounding → also tightened** (`@trunc`→`@floor`), so negative-second inputs normalize to a non-negative nanosecond field.

**Test results (all green locally):**
- `zig build` + `zig build wasm` — build clean
- `zig build test` — pass
- `tests/scheme/srfi/srfi-19-tests.scm` — 112/112
- `tests/scheme/smoke/srfi170-time-objects.scm` — 16/16
- `bash tests/scheme/run-all.sh` — **1802 pass, 0 fail**

LGTM — ready to merge. 🚀

PR #1321, opened 2026-07-08
Migrate reader Unicode classification to generated tables

### LINKED ISSUE #1268: Migrate reader.zig's hand-rolled isUnicodeLetter to the generated Unicode tables
## Context

PR #1263 (fixing #1145) replaced the case-mapping/script-block heuristics in `primitives_char.zig` with range tables generated from Unicode 15.1 `DerivedCoreProperties.txt` (`alphabetic_ranges`, `uppercase_ranges`, `lowercase_ranges`, `cased_ranges`, plus the `inRanges` binary search in `unicode_tables.zig`).

`src/reader.zig:161-198` still carries its own copy of the old pattern: ~30 hardcoded script-block ranges plus a fallback through `findLower`/`findUpper`/`containsU21(&extra_uppercase / &extra_lowercase)`. It is used by the tokenizer's identifier lexing (`reader.zig:306`) and, via `Reader.isUnicodeLetter`, by `printer.zig:267` (`symbolNeedsBars`).

## Problems

1. **Divergence from `char-alphabetic?`** (new since #1263): e.g. U+02B0 (ʰ, Lm modifier letter): `(char-alphabetic? #\x02B0)` → `#t`, but `(define xʰy 42)` → `read error: UnexpectedChar`. R7RS doesn't require identifier syntax to match `char-alphabetic?`, but the current split is accidental rather than a chosen policy.
2. **Drift on regeneration**: `tools/gen_unicode_tables.py` regenerations update `unicode_tables.zig` automatically; reader.zig's inline ranges never move, so the two classifiers drift further apart with each Unicode version bump.
3. **Dead weight**: `extra_uppercase` (471 entries), `extra_lowercase` (829 entries), and `containsU21` are now generated and shipped solely for reader.zig — two parallel classification mechanisms where one would do.

## Suggested work

- Replace `Reader.isUnicodeLetter`'s body with `unicode.inRanges(&unicode.alphabetic_ranges, cp)` — or decide an explicit identifier-char policy if Alphabetic is too broad for identifier-initial position (it includes `Other_Alphabetic` combining marks and Indic vowel signs).
- Reader and printer must move together: `printer.zig`'s `symbolNeedsBars` uses the same predicate, so the `write` → `read` round-trip property stays intact as long as both keep using the shared function. Add a round-trip test for a few newly-accepted codepoints (e.g. U+02B0).
- Once reader.zig no longer uses them, drop `extra_uppercase`/`extra_lowercase`/`containsU21` from the `gen_unicode_tables.py` output.

Found during review of #1263 (see the review's follow-up note there for the verification details).

baijum:
## Summary

- Replace reader.zig's hand-rolled `isUnicodeLetter` (27 hardcoded script-block ranges + case-table fallback) with `unicode.inRanges(&unicode.alphabetic_ranges, cp)`, matching primitives_char.zig
- Remove dead `extra_uppercase` (471 entries), `extra_lowercase` (829 entries), and `containsU21` from unicode_tables.zig and gen_unicode_tables.py
- Add regression tests for newly-accepted codepoints (U+02B0 modifier letter, U+00AA feminine ordinal) and write/read round-trip

Closes #1268

## Test plan

- [x] `zig build run -- tests/scheme/r7rs/r7rs-tests.scm` — 1395 pass, 0 fail
- [x] `bash tests/scheme/run-all.sh` — 1802 pass, 0 fail
- [x] `python3 tools/gen_unicode_tables.py` regenerates without dead code
- [x] New Scheme smoke test: `tests/scheme/smoke/unicode-identifier-roundtrip.scm` — 7/7 pass


baijum:
## Review — ready to merge ✅

Reviewed the change end-to-end by checking out the PR head (`1b21048`) and building/running it locally. This is a clean, well-scoped dedup that does exactly what #1268 asked: it collapses reader.zig's hand-rolled classifier into the single generated predicate `unicode.inRanges(&unicode.alphabetic_ranges, cp)` that `char-alphabetic?` already uses, and removes the ~1300 dead entries that existed only to feed the old fallback.

### Verified locally
- `zig build` — clean
- `zig build test` — passes (exit 0)
- R7RS suite — **1395 pass, 0 fail** (matches the PR claim)
- New smoke test `unicode-identifier-roundtrip.scm` — 7/7
- Reader smoke tests (`reader-peculiar-idents`, `reader-token-validation`, `reader-numeric-prefix`, `char-unicode-properties`) — all pass
- `python3 tools/gen_unicode_tables.py` regenerates and the committed `src/unicode_tables.zig` is byte-identical to *generator output + `zig fmt`* — so the generated file is in sync
- Confirmed `extra_uppercase` / `extra_lowercase` / `containsU21` have **zero** remaining references in `src/reader.zig` and `src/unicode_tables.zig`
- Confirmed reader ↔ printer stay coupled: `printer.symbolNeedsBars` (printer.zig:267) still calls `Reader.isUnicodeLetter`, so the `write` → `read` round-trip invariant is preserved by construction (the round-trip tests confirm it)

### One thing worth recording (not a blocker)
Adopting the Alphabetic property is not purely additive — it's also **stricter than the old whole-block ranges** in some scripts. The old code accepted entire blocks (e.g. all of Devanagari `0x900–0x97F`, Arabic `0x600–0x6FF`), which swept in non-letters like Indic/Arabic **digits** and punctuation. Those are not `Alphabetic`, so they're now rejected. Concretely, `(define क९ 5)` (Devanagari letter + Devanagari digit U+096F) parsed before and now raises `read error: UnexpectedChar`.

This is the correct/intended direction — it's what makes the reader agree with `char-alphabetic?`, and the old behavior even let identifiers *begin* with Devanagari digits. Since `isUnicodeSubsequent` delegates to `isUnicodeLetter` for cp > 127, non-ASCII digits are simply never valid in identifiers now. Just flagging it as a (small) source-compatibility change so it's a conscious choice rather than a surprise.

### On the CodeRabbit comments
- **SRFI-64 comment on the smoke test — I'd skip it.** CodeRabbit calls SRFI-64 "the repo's required style," but that's not accurate: 60 of the ~109 smoke tests use manual pass/fail counters (vs 49 using SRFI-64), and this file already `(exit 1)`s on failure so `run-all.sh` detects it. The manual style is the majority pattern here — no change needed.
- **The two Zig-test simplification comments are fair, minor polish.** `src/CLAUDE.md` recommends the `testing_helpers` shortcuts for exactly these simple cases:
  - `test "Unicode identifier with modifier letter"` → three `th.expectEval("(define xʰy 42) xʰy", 42)` calls instead of manual `TestContext` + `types.toFixnum`.
  - `test "Unicode symbol write/read round-trip"` → `th.expectEvalTrue(...)`.

  Optional — the tests are correct as written; this is just concision.

Net: correct, tested, and in sync with the generator. **Approving for merge** — the CodeRabbit nits above are optional and can be squashed in or left as-is.

Issue #1215, opened 2026-07-05
[SRFI-41] stream-unfold predicate sense inverted, stream-append truncates after the first stream, stream-zip crashes


baijum:
**Category:** SRFI
**Spec reference:** SRFI-41: "(stream-unfold map pred? gen base) ... returns a stream containing the elements (map base), (map (gen base)), ... as long as (pred? ...) returns true" — elements are produced WHILE the predicate holds. stream-append returns "a stream containing all the elements of the input streams in order". stream-zip stops at the shortest stream.
**Severity:** wrong-result

**Minimal reproductions (fresh build at 5d0aeeb):**
```scheme
(import (scheme base) (srfi 41))

;; 1. stream-unfold: predicate sense inverted (impl stops when pred is TRUE)
(stream->list (stream-unfold (lambda (x) (* x x)) (lambda (x) (< x 4))
                             (lambda (x) (+ x 1)) 0))
;; expected: (0 1 4 9)   actual: ()

;; 2. stream-append drops everything after the first stream
(stream->list (stream-append (stream 1 2) (stream 3 4)))
;; expected: (1 2 3 4)   actual: (1 2)

;; 3. stream-zip errors instead of stopping at the shortest stream
(stream->list (stream-zip (stream 1 2) (stream 'a 'b 'c)))
;; expected: ((1 a) (2 b))   actual: error: type error in 'car': expected pair, got #<void>
```

**Source location:** lib/srfi/41.sld — `stream-unfold` tests `(if (pred s) stream-null ...)` (inverted); `stream-append`/`stream-zip` mix the promise layering of `stream-cons` (which wraps its tail in an extra delay) with raw stream values, so appended tails/zip steps land in the wrong layer.
**Related:** #1210 (15 derived exports missing from the same file).

**Verified:** fresh ReleaseSafe build at 5d0aeeb
**Disabled test:** tests/scheme/srfi/srfi41.scm (marked ;; FAIL: #this)

**Found by:** Systematic audit, Phase 3c

baijum:
All three bugs are fixed and merged:

1. **stream-unfold predicate sense** — fixed in PR #1322 (commit c25dcccb)
2. **stream-append truncation** — fixed in PR #1322 (stream macro base case)
3. **stream-zip crash** — any-null? fix landed via PR #1330 (commit a9990f2a)

37 SRFI-41 tests pass, including regression tests for all three reproductions.

PR #1322, opened 2026-07-08
Fix stream-unfold predicate sense and stream macro hygiene (#1215)

baijum:
## Summary

- **stream-unfold**: flip predicate sense — SRFI-41 says elements are produced *while* `pred?` holds, but the implementation stopped when `pred?` returned true
- **stream macro**: replace `stream-null` free-variable reference with `(delay '())` in the base-case template — the `collectSetTargets` pre-scan expands macros without VOID-marking free refs, causing `renameForHygiene` to give `stream-null` a hygienic prefix that breaks when two `(stream …)` calls appear as sibling arguments
- **temp_globals cleanup**: reverse the cleanup loop to LIFO order to prevent Phase-B entries from re-adding names that Phase-A2 correctly removed

Fixes `stream-unfold`, `stream-append`, and `stream-zip` as reported in #1215.

## Test plan

- [x] `zig build test` — 668 pass, 0 fail
- [x] `bash tests/scheme/run-all.sh` — 1801 pass, 0 fail
- [x] Uncommented 3 FAIL-marked tests in `tests/scheme/srfi/srfi41.scm`
- [x] Added 10 new regression tests for stream-append, stream-zip, stream-unfold edge cases, and sibling stream macro calls
- [x] Added Zig unit test for recursive variadic macro sibling calls

### DIFF
--- a/lib/srfi/41.sld
+++ b/lib/srfi/41.sld
@@ -31,15 +31,15 @@
 
     (define-syntax stream
       (syntax-rules ()
-        ((stream) stream-null)
+        ((stream) (delay '()))
         ((stream x rest ...)
          (stream-cons x (stream rest ...)))))
 
     (define (stream-unfold mapper pred gen seed)
       (define (unfold-loop s)
         (if (pred s)
-            stream-null
-            (stream-cons (mapper s) (unfold-loop (gen s)))))
+            (stream-cons (mapper s) (unfold-loop (gen s)))
+            stream-null))
       (unfold-loop seed))
 
     (define (stream-map proc strm)
--- a/lib/srfi/41.sld
+++ b/lib/srfi/41.sld
@@ -118,8 +118,12 @@
                        (apply stream-append (cdr strms))))))
 
     (define (stream-zip . strms)
+      (define (any-null? ss)
+        (and (not (null? ss))
+             (or (stream-null? (car ss))
+                 (any-null? (cdr ss)))))
       (define (zip-loop ss)
-        (if (or (null? ss) (stream-null? (car ss)))
+        (if (or (null? ss) (any-null? ss))
             stream-null
             (stream-cons (map stream-car ss)
                          (zip-loop (map stream-cdr ss)))))
--- a/src/compiler_macro.zig
+++ b/src/compiler_macro.zig
@@ -145,7 +145,10 @@ pub fn expandAndCompileMacroUse(self: *Compiler, expr: Value, name: []const u8, 
     }
     defer if (self.globals) |g| {
         const glk = globals_mod.acquireGlobalsWrite(g);
-        for (temp_globals[0..temp_global_count]) |tg| {
+        var tgi = temp_global_count;
+        while (tgi > 0) {
+            tgi -= 1;
+            const tg = temp_globals[tgi];
             if (tg.was_present) {
                 g.put(tg.name, tg.old_val.?) catch {};
             } else {
--- a/src/tests_macros.zig
+++ b/src/tests_macros.zig
@@ -716,3 +716,15 @@ test "lambda parameter shadows a syntactic keyword in its body (#788)" {
     // An unshadowed keyword still compiles as the special form.
     try std.testing.expectEqual(@as(i64, 2), types.toFixnum(try vm.eval("((lambda (x) (if 1 2 3)) 0)")));
 }
+
+test "recursive variadic macro: sibling calls produce correct values (#1215)" {
+    try th.expectEval(
+        \\(begin
+        \\  (define-syntax my-list
+        \\    (syntax-rules ()
+        \\      ((my-list) '())
+        \\      ((my-list x rest ...)
+        \\       (cons x (my-list rest ...)))))
+        \\  (+ (car (my-list 10 20)) (car (my-list 30 40))))
+    , 40);
+}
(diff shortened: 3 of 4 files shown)


baijum (commented):
## Review — ready to merge, with one non-blocking suggestion

I checked this out, built it, and confirmed it **correctly and completely fixes #1215**. CI is green on every platform, it's up to date with `main`, and I found no regressions.

### What I verified
- **Issue #1215's three exact reproductions now return the correct values** (`(0 1 4 9)`, `(1 2 3 4)`, `((1 a) (2 b))`), both combined and in isolation — not just inside the test file.
- The two `lib/srfi/41.sld` changes are what actually fix the bug. With **only the `.sld` fix applied on top of the unmodified `main` compiler**, the whole `srfi41.scm` suite passes **36/0**.
- No regressions from the shared-path compiler change: **R7RS suite 1395/0**, `zig build test` clean, all 22 macro/hygiene/library-leak Scheme suites pass, `zig fmt --check` clean.
- `(delay '())` is behaviorally equivalent to the shared `stream-null` here — `(stream)` still satisfies `stream-null?`/`stream?` and yields `()`. The only observable difference (`(eq? (stream) stream-null)` → `#f`) is harmless since SRFI-41 uses the value-based `stream-null?` predicate.

### Non-blocking: the `compiler_macro.zig` change has no test that fails without it

`src/compiler_macro.zig:146-156` (LIFO cleanup) is logically sound — reverse-order unwind is the correct way to undo the overlapping Phase-A2 / Phase-B `temp_globals` entries — and it's low-risk. But I couldn't find anything that actually depends on it:

- **PR compiler + `main` library** → still fails the same **10** `srfi41` assertions (identical to `main`).
- **`main` compiler + PR library** → passes all **36**.

So for #1215 the fix is entirely the `.sld` change; the LIFO change is neither necessary nor sufficient for any test here. The new Zig test that appears intended to guard it —
`tests_macros.zig:720` *"recursive variadic macro: sibling calls produce correct values (#1215)"* — returns `40` on **unmodified `main`** as well, so it doesn't fail without the fix.

Per this repo's rule (*"Every bug fix MUST include a regression test that fails without the fix"*), it'd be good to either (a) add a test that genuinely exercises the `temp_globals` leak (fails on `main`), or (b) note in the commit that the LIFO reversal is a defensive robustness fix rather than the #1215 root cause. Not a blocker — the change is safe and plausibly fixes a real latent leak; it's just currently unguarded.

### Minor (agrees with CodeRabbit)
The sibling `promise?` assertions at `tests/scheme/srfi/srfi41.scm:91-94` are a weak/order-dependent guard — `(list (promise? (stream 1)) (promise? (stream 2)))` → `(#t #t)` passes even on unmodified `main` when run in isolation. Asserting the expanded contents via `stream->list` (as CodeRabbit suggested) would be a more robust regression.

**Verdict: approve / ready to merge.** The core fix is correct, verified against the issue's own reproductions, and regression-free. The suggestions above are about test strength, not correctness.

baijum:
Addressed the review feedback in 18eadc6:

- **stream-zip**: fixed a real bug — `any-null?` now checks ALL streams, not just the first. The new first-longer test (`(stream-zip (stream 1 2) (stream 10))`) caught this.
- **Sibling tests**: replaced `promise?` assertions with `stream->list` content checks.

Re: the **compiler_macro.zig LIFO change** — agreed this is a defensive robustness fix, not the #1215 root cause. The `.sld` change alone fixes all three reported bugs. The LIFO reversal prevents a real globals leak (Phase-B entries re-adding names that Phase-A2 correctly removed), but the symptom is currently masked at runtime by the VM's hygienic-prefix fallback in `get_global`. I couldn't construct a test case where the leak produces a wrong result, so the Zig unit test guards the general recursive-variadic-macro-as-sibling pattern rather than the LIFO path specifically.

baijum (commented):
## Follow-up review of `18eadc6a` — good improvement, ready to merge

Re-checked out and re-tested against the new commit. It resolves both test-strength points and, in the process, fixed a genuine latent bug. CI is green on all platforms; merge state is clean.

### Points addressed ✅
1. **Weak `promise?` sibling assertions → content checks.** The new `two-streams`/`three-streams` tests are now *real* regressions: against the unfixed library they fail with `((1 2) #<promise>)` and `((a) #<promise> (b))`, exactly the hygiene corruption. The old `(#t #t)` assertions passed on `main` in isolation, so this is a real strengthening.
2. **Longer-first `stream-zip` case — and it caught a real bug.** The old `stream-zip` only checked `(stream-null? (car ss))`, so it crashed (`car: expected pair, got ()`) whenever a *later* stream was the shortest — e.g. `(stream-zip (stream 1 2) (stream 10))`. #1215 only reported the first-stream-shorter path, so this was uncovered until now. The `any-null?` fix is correct, and the added test genuinely fails without it (verified: 36/1 against the pre-fix library).

### Verification
- `srfi41.scm` **37/0**; R7RS **1395/0**; `zig build test` clean; srfi42/64 macro suites clean — no regressions.

### Still open (non-blocking, unchanged from my last review)
- **`src/compiler_macro.zig:146-156` (LIFO cleanup) still has no test that fails without it.** With this commit's library + tests but the unmodified `main` compiler, the suite still passes **37/0**, and the `tests_macros.zig:720` `my-list` test still returns `40` on `main`. The change is sound and low-risk, but if it's meant to guard a real `temp_globals` leak it'd be good to either add a test that exercises it (fails without the LIFO fix) or note in the message that it's a defensive robustness change rather than part of the #1215 root cause.

**Verdict: approve / ready to merge.** The fix is correct, verified against the issue's reproductions, now well-covered by genuine regression tests, and regression-free. The remaining note is optional cleanup.

PR #1324, opened 2026-07-08
Fix SRFI-210 set!-values shadowing bug (#1224)

### LINKED ISSUE #1224: [SRFI-210] set!-values is a no-op — the macro assigns the consumer lambda's own parameters
**Category:** SRFI
**Spec reference:** SRFI-210: "(set!-values (variable ...) producer)" assigns the producer's values to the (existing) variables.
**Severity:** wrong-result (silent)

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 210))
(define a 0)
(define b 0)
(set!-values (a b) (values 10 20))
(list a b)   ;; expected: (10 20)   actual: (0 0)
```

**Root cause:** lib/srfi/210.sld expands to
```scheme
(call-with-values (lambda () producer)
  (lambda (var ...) (set! var var) ...))
```
The consumer lambda's parameters shadow the outer variables, so each `(set! var var)` assigns a parameter to itself and the outer bindings are never touched. The consumer must bind fresh (hygienic) temporaries: `(lambda (tmp ...) (set! var tmp) ...)`.

**Related:** #1218 (value returns its first argument; box/mv missing — same file).

**Source location:** lib/srfi/210.sld (`set!-values`)
**Disabled test:** tests/scheme/srfi/srfi210.scm (marked ;; FAIL: #this)

baijum:
## Summary
- Fix `set!-values` macro in `lib/srfi/210.sld` — the consumer lambda's parameters shadowed the outer variables, so `(set! var var)` was a no-op
- Use a rest-arg lambda with a `set!-values-aux` helper macro that walks the value list and assigns each to the correct outer variable
- Enable the disabled test and add edge cases (single variable, repeated use)

Closes #1224

## Test plan
- [x] `zig build run -- tests/scheme/srfi/srfi210.scm` — 33 pass, 0 fail
- [x] Minimal repro from the issue now returns `(10 20)` as expected

### DIFF
--- a/lib/srfi/210.sld
+++ b/lib/srfi/210.sld
@@ -18,7 +18,17 @@
       (syntax-rules ()
         ((set!-values (var ...) producer)
          (call-with-values (lambda () producer)
-           (lambda (var ...) (set! var var) ...)))))
+           (lambda args
+             (set!-values-aux (var ...) args))))))
+
+    (define-syntax set!-values-aux
+      (syntax-rules ()
+        ((set!-values-aux () args)
+         (if #f #f))
+        ((set!-values-aux (var . vars) args)
+         (begin
+           (set! var (car args))
+           (set!-values-aux vars (cdr args))))))
 
     (define-syntax case-receive
       (syntax-rules (else)
--- a/tests/scheme/srfi/srfi210.scm
+++ b/tests/scheme/srfi/srfi210.scm
@@ -25,11 +25,19 @@
 (test 'many (case-receive (values 1 2 3) ((a) 'one) ((a b) 'two) (else 'many)))
 
 ;; set!-values assigns existing variables:
-;; FAIL: #1224 (set!-values is a no-op — assigns the consumer's own params)
-;; (define sv-a 0)
-;; (define sv-b 0)
-;; (set!-values (sv-a sv-b) (values 10 20))
-;; (test '(10 20) (list sv-a sv-b))
+(define sv-a 0)
+(define sv-b 0)
+(set!-values (sv-a sv-b) (values 10 20))
+(test '(10 20) (list sv-a sv-b))
+
+;; set!-values with single variable
+(define sv-c 0)
+(set!-values (sv-c) (values 99))
+(test 99 sv-c)
+
+;; set!-values overwrites previous set!-values results
+(set!-values (sv-a sv-b) (values 30 40))
+(test '(30 40) (list sv-a sv-b))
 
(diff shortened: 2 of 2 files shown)


baijum:

The fix correctly resolves #1224. The root cause diagnosis is right: the old consumer lambda `(lambda (var ...) (set! var var) ...)` rebound the outer variables as its own parameters, so every assignment was a self-assign no-op. Replacing it with a rest-arg lambda plus a `set!-values-aux` walker that assigns from a fresh value list is a correct and hygienic fix.

**Verified locally** (fresh ReleaseSafe build):
- `zig build run -- tests/scheme/srfi/srfi210.scm` → **33 pass, 0 fail**
- Minimal repro from the issue now returns `(10 20)` ✅
- **Hygiene holds:** a user variable literally named `args` does not collide with the macro-introduced rest arg — `(set!-values (args b) (values 111 222))` correctly yields `(111 222)`.
- SRFI-195 (the `box` dependency) still passes (21/21), and `set!-values` has no other consumers in the repo.
- `set!-values-aux` is correctly kept out of the export list, so it stays library-internal.

Good test additions too — single-variable and repeated-use cases nicely lock in the fix.

### One non-blocking observation: arity handling changed

Switching from a fixed-arity consumer lambda to `(lambda args …)` means the value-count check is gone. For inputs that violate the SRFI-210 contract (producer must return exactly N values), behavior differs from the old code:

| input | old (fixed-arity lambda) | this PR |
|-------|--------------------------|---------|
| `(set!-values (a b) (values 1 2 3))` | clean `ArityMismatch`, vars untouched | **silently truncates** → `a=1 b=2`, `3` dropped |
| `(set!-values (a b) (values 1))` | clean `ArityMismatch`, vars untouched | `a` mutated to `1`, **then** low-level `error in 'car': expected pair, got ()` on `b` |

The too-few case is the more unfortunate one: it partially mutates before erroring, so the operation is no longer atomic, and the error surfaces as an internal `car` type error rather than a clear arity message.

This is all on contract-violating input and is arguably out of scope for a focused bug fix — I don't consider it a blocker. If you want to preserve the old strictness, the `set!-values-aux` base case could reject leftover values, e.g.:

```scheme
((set!-values-aux () args)
 (if (null? args) (if #f #f)
     (error "set!-values: too many values")))
```

(Fully restoring atomicity on the too-few case would need binding all temps before any `set!`, which is more machinery than this fix warrants.)

Merge-ready as-is. 🚀

PR #1325, opened 2026-07-08
Fix SRFI-175 ascii-digit-value and add 16 missing procedures (#1236)

### LINKED ISSUE #1236: [SRFI-175] ascii-digit-value treats letters as digits 10-35; 15 spec exports missing (ci comparisons, nth-digit/upper/lower, mirror-bracket, control<->graphic, bytevector?)
**Category:** SRFI (lib/srfi/175.sld)
**Spec reference:** SRFI-175 ascii-digit-value: "If char represents an ASCII decimal digit, returns the numeric value 0..9 of that digit. Only digit values less than limit are accepted: for example, a limit of 8 accepts only octal digits. If char does not represent an acceptable digit, #f is returned." Letters are handled by the separate ascii-upper-case-value/ascii-lower-case-value procedures.
**Severity:** wrong-result (silent) + missing API

```scheme
(import (scheme base) (srfi 175))
(ascii-digit-value #\a 16)   ;; expected #f   actual 10
(ascii-digit-value #\F 16)   ;; expected #f   actual 15
```

**Missing exports** (spec list minus implementation): ascii-bytevector?, ascii-ci=?, ascii-ci<?, ascii-ci>?, ascii-ci<=?, ascii-ci>=?, ascii-string-ci=?, ascii-string-ci<?, ascii-string-ci>?, ascii-string-ci<=?, ascii-string-ci>=?, ascii-control->graphic, ascii-graphic->control, ascii-mirror-bracket, ascii-nth-digit, ascii-nth-upper-case, ascii-nth-lower-case.

**Root cause:** lib/srfi/175.sld's ascii-digit-value folds the A-Z/a-z ranges into the digit lookup (values 10-35), conflating it with a general radix-digit routine.

**Conforms:** all classification predicates (chars and codepoints), upcase/downcase on both representations, upper/lower-case-value including the limit check.

**Disabled tests:** tests/scheme/srfi/srfi175.scm (marked ;; FAIL: #this)

baijum:
## Summary
- **Fix `ascii-digit-value`**: removed incorrect A-Z/a-z letter handling (values 10-35). Per SRFI-175, this procedure only accepts decimal digits 0-9; letters are handled by `ascii-upper-case-value` and `ascii-lower-case-value`.
- **Add 16 missing procedures**: `ascii-bytevector?`, `ascii-ci=?`/`<?`/`>?`/`<=?`/`>=?`, `ascii-string-ci=?`/`<?`/`>?`/`<=?`/`>=?`, `ascii-control->graphic`, `ascii-graphic->control`, `ascii-mirror-bracket`, `ascii-nth-digit`, `ascii-nth-upper-case`, `ascii-nth-lower-case`.
- Implementation now exports all 35 SRFI-175 procedures.

Closes #1236

## Test plan
- [x] 129 SRFI-175 conformance tests pass (was 62 with disabled FAIL markers)
- [x] Regression tests for `(ascii-digit-value #\a 16)` → `#f` and `(ascii-digit-value #\F 16)` → `#f`
- [x] Full test suite: 1806 pass, 0 fail


baijum (commented):
## Review — verified against the SRFI-175 spec, built, and ran

Checked out the PR head (`b0aba0d`), rebuilt (`zig build`), and ran the suite. **LGTM — ready to merge.** No blocking issues.

### Correctness (checked each procedure against the SRFI-175 text)

- **`ascii-digit-value` fix** — correct. Now accepts only decimal digits 0–9 with the `limit` check; letters return `#f`. This matches the spec, which reserves letter values for `ascii-upper-case-value` / `ascii-lower-case-value`. Renaming `radix` → `limit` also aligns with the spec's parameter name.
- **`ascii-nth-digit`** — returns `#f` for non-exact-integer and out-of-range `n`, exactly as the spec requires. Verified `(ascii-nth-digit 5.0)` → `#f`, `(ascii-nth-digit 10)` → `#f`.
- **`ascii-nth-upper-case` / `ascii-nth-lower-case`** — `(modulo n 26)` correctly wraps, including negatives: verified `(ascii-nth-upper-case -1)` → `#\Z`, `(ascii-nth-lower-case -1)` → `#\z`. Matches the spec's "taken modulo 26, negatives permitted".
- **`ascii-control->graphic` / `ascii-graphic->control`** — correct bit-6 flip. `0..31 → +64`, `DEL(127) → ?(63)`, and the inverse `?(63) → DEL`, `64..95 → −64`. Spot-checked `(ascii-control->graphic 27)` → `91` (`[`), `(ascii-graphic->control 95)` → `31`.
- **`ascii-mirror-bracket`** — all four pairs `() [] {} <>` handled, `#f` otherwise. ✓
- **`ascii-ci*` / `ascii-string-ci*`** — 2-argument comparisons (spec is not variadic here), folding upper→lower to match the reference implementation. Accepts mixed char/int, e.g. `(ascii-ci=? 65 #\a)` → `#t`. ✓
- **`ascii-bytevector?`** — `#t` for empty and all-`<=127`; `#f` for a byte `>127` or a non-bytevector. ✓
- Both loops (`ascii-bytevector?`, `%string-ci-cmp`) recurse in tail position, so no stack growth on long inputs.

### Tests
- 129 conformance tests pass (0 fail); the previously disabled `FAIL: #1236` cases are now live regressions (`(ascii-digit-value #\a 16)` → `#f`, etc.).
- The SRFI-64 epilogue correctly grabs the runner **before** `test-end` and exits non-zero on failure, so `run-all.sh` catches regressions.
- No code elsewhere in the repo relies on the old radix-16 behavior (only an audit-tracking doc mentions it).

### Minor (non-blocking)
- The test file switches from `(chibi test)` to `(srfi 64)`. That's the less-common convention in `tests/scheme/srfi/` (~41 chibi vs ~6 srfi-64), though it's the pattern the audit campaign is standardizing on, and the exit-on-fail epilogue is a plus. Fine as-is.
- `ascii-nth-upper-case` / `ascii-nth-lower-case` don't guard non-integer `n` (they'd raise inside `modulo`), whereas `ascii-nth-digit` returns `#f`. This is spec-conformant — the spec only mandates `#f` for `ascii-nth-digit` — just noting the intentional asymmetry.

PR #1326, opened 2026-07-08
Fix SRFI-43 vector library to match spec (#1209)

### LINKED ISSUE #1209: [SRFI-43] Vector procedures use SRFI-133 semantics — callbacks never receive the index — and 8 spec procedures are missing
**Category:** SRFI
**Spec reference:** SRFI-43 callback conventions (all quoted from the spec): vector-map — "Each element at index i of the new vector is mapped from the old vectors by (f i ...)"; vector-fold — "kons is applied as (kons i state (vector-ref vec1 i))"; vector-for-each applies f "to each index in the range [0, length)" together with the elements; vector-count likewise passes "i and each parallel element in the vectors at that index". SRFI-133 deliberately removed the index argument — the two SRFIs are incompatible on these procedures.
**Severity:** wrong-result (silently different callback arity/semantics)

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 43))
(vector-map (lambda (i x) (+ i x)) #(10 10 10))
;; SRFI-43 expected: #(10 11 12)  actual: error (callback called with 1 arg) / wrong result
(vector-fold (lambda (i state x) (cons x state)) '() #(a b c))
;; SRFI-43 expected: (c b a)      actual: callback called as (kons state x) — no index
```

**Analysis:** lib/srfi/43.sld re-exports the built-in SRFI-133-style procedures (vector-map, vector-for-each, vector-count, vector-index, ...) and defines vector-fold/vector-fold-right without the index parameter and single-vector only. Additionally these SRFI-43 exports are missing entirely: vector-unfold, vector-unfold-right, vector=, vector-binary-search, vector-reverse!, vector-reverse-copy!, reverse-vector->list, reverse-list->vector.

Note vector-fold-right in SRFI-43 is also (kons i state elt) processing indices high→low; the current definition passes (kons acc elt).

**Source location:** lib/srfi/43.sld
**Disabled test:** tests/scheme/srfi/srfi43.scm (marked ;; FAIL: #this)
**Related:** #1172 (SRFI-133's own missing procedures)

baijum:
## Summary

- Rewrite 6 iteration procedures (`vector-map`, `vector-map!`, `vector-for-each`, `vector-count`, `vector-fold`, `vector-fold-right`) with correct SRFI-43 calling convention — callbacks receive the index as the first argument, and multi-vector operation is supported
- Add 8 missing exports re-exported from SRFI-133: `vector-unfold`, `vector-unfold-right`, `vector=`, `vector-binary-search`, `vector-reverse!`, `vector-reverse-copy!`, `reverse-vector->list`, `reverse-list->vector`
- Expand test suite from 22 to 59 tests covering all exported procedures

## Test plan

- [x] All 59 SRFI-43 tests pass
- [x] Full Scheme test suite passes (1806 pass, 0 fail)
- [x] Zig unit tests pass

Closes #1209

### DIFF
--- a/lib/srfi/43.sld
+++ b/lib/srfi/43.sld
@@ -1,9 +1,10 @@
 ;;; SRFI 43 — Vector Library
 ;;; SRFI-43 iteration procedures pass the index as the first callback
-;;; argument; SRFI-133 does not.  Procedures with identical semantics
-;;; are re-exported directly from SRFI-133.
+;;; argument; SRFI-133 does not.  The six iteration procedures are
+;;; redefined below; remaining procedures are re-exported from
+;;; (scheme base) and (srfi 133).
 (define-library (srfi 43)
-  (import (scheme base)
+  (import (except (scheme base) vector-map vector-for-each)
           (except (srfi 133)
                   vector-map! vector-count
                   vector-fold vector-fold-right))
--- a/lib/srfi/43.sld
+++ b/lib/srfi/43.sld
@@ -1,7 +1,12 @@
-;;; SRFI 43 — Vector Library (predecessor to SRFI 133)
-;;; Re-exports from SRFI 133 with SRFI 43 names
+;;; SRFI 43 — Vector Library
+;;; SRFI-43 iteration procedures pass the index as the first callback
+;;; argument; SRFI-133 does not.  Procedures with identical semantics
+;;; are re-exported directly from SRFI-133.
 (define-library (srfi 43)
-  (import (scheme base) (srfi 133))
+  (import (scheme base)
+          (except (srfi 133)
+                  vector-map! vector-count
+                  vector-fold vector-fold-right))
   (export make-vector vector vector? vector-ref vector-set! vector-length
           vector-fold vector-fold-right
           vector-map vector-map!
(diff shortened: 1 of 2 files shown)


lib/srfi/43.sld
```diff
@@ -1,7 +1,12 @@
...
+;;; SRFI 43 — Vector Library
+;;; SRFI-43 iteration procedures pass the index as the first callback
+;;; argument; SRFI-133 does not.  Procedures with identical semantics
+;;; are re-exported directly from SRFI-133.
 (define-library (srfi 43)
-  (import (scheme base) (srfi 133))
+  (import (scheme base)
+          (except (srfi 133)
+                  vector-map! vector-count
+                  vector-fold vector-fold-right))
```
baijum, line 10: `vector-map` and `vector-for-each` belong to `(scheme base)`, not `(srfi 133)`, so this `except` doesn't remove them — they're imported here **and** redefined in the `begin` block below (lines 50 and 73). That's a duplicate binding: it passes only because Kaappi lets the local `define` shadow the import, but R7RS treats import-and-define of the same name as an error, and it's inconsistent with how you (correctly) excluded the four SRFI-133-only procedures.

Suggest excepting them from `(scheme base)` too:

```scheme
(import (except (scheme base) vector-map vector-for-each)
        (except (srfi 133)
                vector-map! vector-count
                vector-fold vector-fold-right))
```

Verified this keeps all 59 SRFI-43 tests passing.

baijum (commented):
Reviewed this locally: built the branch at `01be32d` (ReleaseSafe) and ran `tests/scheme/srfi/srfi43.scm` → **59 pass, 0 fail**. The core fix is correct — the six iteration procedures now pass the index as the first callback argument and support multiple vectors, matching the SRFI-43 spec, and all 8 previously-missing exports are legitimate SRFI-43 procedures correctly sourced from the built-in SRFI-133 primitives (whose semantics are identical for those). The `(except (srfi 133) …)` on the four procedures that live only in SRFI-133 is exactly right.

One substantive issue to address before merge (see inline comment on the import), plus one doc nit:

**1. `vector-map` / `vector-for-each` are both imported and redefined (duplicate binding).** These two come from `(scheme base)` (not from SRFI-133), so `(except (srfi 133) …)` doesn't remove them — the library imports them from `(scheme base)` and *also* defines them in the `begin` block. It works today only because Kaappi silently lets the local `define` shadow the import; R7RS treats importing and defining the same identifier as an error, so this relies on non-standard loader behavior and is inconsistent with the careful `except` you applied to the other four. Excepting them from `(scheme base)` fixes it cleanly — I verified the change keeps all 59 tests passing.

**2. Doc nit:** the header comment says procedures with identical semantics "are re-exported directly from SRFI-133," but `vector-map`/`vector-for-each` are actually redefined here and the passthrough procedures come from `(scheme base)` as well as SRFI-133 — worth a small wording tweak.

With the import fix the PR looks ready to merge.

baijum (commented):
Both points from my earlier review are addressed by `5e1be06`:

1. **Duplicate binding** — the import now reads `(except (scheme base) vector-map vector-for-each)`, so those two names are no longer both imported and redefined. This is exactly the fix, and it drops the reliance on Kaappi's define-shadows-import behavior.
2. **Doc comment** — the header now accurately states the six iteration procedures are redefined and the rest are re-exported from `(scheme base)` and `(srfi 133)`.

Verified locally: rebuilt at `5e1be06` (ReleaseSafe) and reran `tests/scheme/srfi/srfi43.scm` → **59 pass, 0 fail**.

LGTM — ready to merge. 🚢

PR #1327, opened 2026-07-08
Fix SRFI-232 curried procedures to support grouped application (#1238)

### LINKED ISSUE #1238: [SRFI-232] Curried procedures are strictly unary chains — grouped application fails; curried form not exported; zero-arg and >4-arg shapes are syntax errors
**Category:** SRFI (lib/srfi/232.sld)
**Spec reference:** SRFI-232: the library provides "curried, a variant of lambda" plus define-curried; "a curried procedure of two arguments may be applied to one argument to return a new procedure of one argument, or directly to two arguments to return its value." Nullary: "If <formals> is an empty list, then a curried expression evaluates to <body>." Variadic dotted formals are also specified.
**Severity:** wrong-behavior (arity errors on spec-conforming calls) + missing API

```scheme
(import (scheme base) (srfi 232))
(define-curried (add2 a b) (+ a b))
((add2 1) 2)      ;; => 3 (works)
(add2 1 2)        ;; expected 3    actual arity error (procedure takes exactly 1 arg)
(define-curried (mul3 a b c) (* a b c))
((mul3 2 3) 4)    ;; expected 24   actual arity error
(curried (a b) (+ a b))          ;; expected a procedure   actual: curried not exported
(define-curried (thunk) 'ok)     ;; expected thunk defined  actual: InvalidSyntax (no pattern)
(define-curried (f a b c d e) 0) ;; expected f defined      actual: InvalidSyntax (patterns stop at 4 args)
```

**Root cause:** lib/srfi/232.sld hardcodes four syntax-rules patterns that expand into fixed unary lambda chains. SRFI-232 currying requires case-lambda-style dispatch on the received argument count with re-currying of the remainder, plus the curried primitive form, nullary, and variadic support.

**Conforms:** one-argument-at-a-time chains for 1-4 fixed parameters, independent partial-application closures.

**Disabled tests:** tests/scheme/srfi/srfi232.scm (marked ;; FAIL: #this)

baijum:
## Summary

- Replace hardcoded 4-pattern unary lambda chains with the SRFI-232 reference implementation using `case-lambda` dispatch
- Export the `curried` form (was missing) alongside `define-curried`
- Support nullary formals, variadic (dotted) formals, and arbitrary arity (was capped at 4 args)
- Grouped application now works: `(add2 1 2)` succeeds instead of raising an arity error

## Test plan

- [x] 34 SRFI-232 conformance tests pass (was 10, most commented out)
- [x] Zig unit tests pass
- [x] Full Scheme suite: 1806 pass, 0 fail

Closes #1238

### DIFF
--- a/lib/srfi/232.sld
+++ b/lib/srfi/232.sld
@@ -1,20 +1,47 @@
 (define-library (srfi 232)
-  (import (scheme base))
-  (export define-curried)
+  (import (scheme base) (scheme case-lambda))
+  (export curried define-curried)
   (begin
+    (define-syntax curried
+      (syntax-rules ()
+        ((curried formals exp ...)
+         (curried-1 formals (begin exp ...)))))
+
+    (define-syntax curried-1
+      (syntax-rules ()
+        ((curried-1 () exp) exp)
+        ((curried-1 (arg0 arg1 ...) exp)
+         (one-or-more (arg0 arg1 ...) exp))
+        ((curried-1 (arg0 arg1 ... . rest) exp)
+         (rest-args (arg0 arg1 ... . rest) exp))
+        ((curried-1 args exp) (lambda args exp))))
+
+    (define-syntax one-or-more
+      (syntax-rules ()
+        ((one-or-more (arg0 arg1 ...) exp)
+         (letrec
+           ((f (case-lambda
+                 (() f)
+                 ((arg0 arg1 ...) exp)
+                 ((arg0 arg1 ... . rest)
+                  (apply (f arg0 arg1 ...) rest))
+                 (args (more-args f args)))))
+           f))))
+
+    (define-syntax rest-args
+      (syntax-rules ()
+        ((rest-args (arg0 arg1 ... . rest) exp)
+         (letrec ((f (case-lambda
+                       (() f)
+                       ((arg0 arg1 ... . rest) exp)
+                       (args (more-args f args)))))
+           f))))
+
+    (define (more-args f current)
+      (lambda args (apply f (append current args))))
+
     (define-syntax define-curried
       (syntax-rules ()
-        ((define-curried (name arg) body ...)
-         (define (name arg) body ...))
-        ((define-curried (name arg1 arg2) body ...)
-         (define (name arg1)
-           (lambda (arg2) body ...)))
-        ((define-curried (name arg1 arg2 arg3) body ...)
-         (define (name arg1)
-           (lambda (arg2)
-             (lambda (arg3) body ...))))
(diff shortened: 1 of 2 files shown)


baijum:

Reviewed the diff, compared against the upstream reference implementation, and verified behavior locally.

### Correctness
The new `lib/srfi/232.sld` is a **faithful, essentially verbatim port** of the SRFI-232 reference implementation ([`srfi-232.scm`](https://github.com/scheme-requests-for-implementation/srfi-232/blob/main/srfi-232.scm), Wolfgang Corcoran-Mathe, MIT). The `curried` / `curried-1` / `one-or-more` / `rest-args` / `more-args` / `define-curried` bodies are identical to upstream — only the `define-library` wrapper (`import (scheme case-lambda)`, `export`, `begin`) is added. Porting the vetted reference rather than hand-rolling the `case-lambda` dispatch is the right call.

### Verified locally (built from this branch)
- **34/34 tests pass**, exit 0 — the macro-introduced references to the library-*private* helpers (`curried-1`, `one-or-more`, `rest-args`, `more-args`) resolve correctly through the library scope even though they aren't exported. This is the main thing that could have broken under Kaappi's hygiene/library-scoping, and it works.
- **Hygiene collisions** (not covered by the suite) — user formals that shadow the implementation's internal bindings all resolve correctly:
  - `(define-curried (g f) (+ f 100))` → `(g 5)` = `105` (formal `f` vs. internal `letrec ((f ...))`)
  - `(define-curried (h args) (list 'got args))` → correct (formal `args` vs. internal `case-lambda` `args` clause)
  - `(define-curried (k rest x) ...)` → correct (formal `rest` vs. internal `. rest`)
- Surplus forwarding with a body that returns a `curried`, and independent variadic partials, both behave per spec.

### Tests / harness
- New suite uses the standard SRFI-64 exit-on-fail epilogue (grab runner before `test-end`, `(exit 1)` on failures) — matches `srfi189.scm`, and is picked up by `run-all.sh` via `tests/scheme/srfi/*.scm`.
- `srfi-232.log` (SRFI-64 side-effect) is already gitignored.
- CI is green: 10 checks pass, 0 fail (macOS + ubuntu arm/x86 Debug+ReleaseFast + riscv64), CodeRabbit passed.

### Minor (non-blocking, optional)
`docs/audit-strategy.md:150` still records "SRFI-232 currying is strictly unary chains" as a Phase 3.4 finding. That's now fixed by this PR. It reads as a historical audit snapshot so it's fine to leave, but you may want a follow-up note that #1238 is resolved.

**Verdict: ready for merge.** Faithful reference port, thorough tests, hygiene verified, CI green, closes #1238.

PR #1329, opened 2026-07-08
Honor custom equivalence/hash functions in SRFI-69 hash tables

### LINKED ISSUE #1183: [SRFI-69] make-hash-table/alist->hash-table ignore custom equivalence and hash functions; the accessors report equal?/hash regardless
**Category:** SRFI
**Spec reference:** SRFI-69 — `(make-hash-table [equal? [hash]])`: "Create a new hash table with equal? as the equivalence function and hash as the hash function"; `hash-table-equivalence-function`/`hash-table-hash-function` return the functions the table actually uses.
**Severity:** wrong-result

**Minimal reproduction:**
```scheme
(import (srfi 69))
(define ht (make-hash-table eq?))
(hash-table-set! ht (string #\a) 'first)
(hash-table-set! ht (string #\a) 'second)   ; distinct string objects, not eq?
(hash-table-size ht)                          ;; expected 2, actual 1
(eq? eq? (hash-table-equivalence-function ht)) ;; expected #t, actual #f (returns equal?)
```

**Cause:** `makeHashTableFn` discards its arguments (`_ = args; // ignore optional comparator/hash args; we always use equal?`); `alistToHashTableFn` does the same. `hashTableEquivFn`/`hashTableHashFn` unconditionally return the global `equal?`/`hash` bindings, so the accessors are truthful about behavior but lie about what the caller requested — silently. Case-insensitive tables (`(make-hash-table string-ci=? string-ci-hash)`), eq?-identity tables, and any custom-comparator table all silently degrade to equal? semantics.

**Prior art:** the make-hash-table half was noted in the (closed, unfixed-in-this-respect) coverage-gap issue #148. Filing focused: honoring the functions requires storing them on the HashTable object and calling back into the VM from `findKey`/`findSlot` (with GC-rooting care), or at minimum raising a clear "custom comparators not supported" error instead of silently accepting them.

**Verified:** fresh ReleaseSafe build at b170e6f
**Source location:** src/primitives_hashtable.zig `makeHashTableFn`, `alistToHashTableFn`, `hashTableEquivFn`, `hashTableHashFn`
**Disabled test:** tests/scheme/audit/primitives_hashtable-audit.scm (marked `;; FAIL: #this`)

baijum:
## Summary

Fixes #1183.

- Store a `CompareMode` enum plus the original Scheme procedure Values on each `HashTable` object so custom equivalence/hash functions are honored
- Recognize well-known comparators (`eq?`, `eqv?`, `equal?`, `string=?`, `string-ci=?`) at creation time and dispatch to built-in Zig implementations; for arbitrary custom Scheme procedures, call back into the VM
- Detect SRFI-128 comparator records passed by SRFI-125 and extract their equality/hash fields automatically
- Fix `valueHash` to hash bignums, rationals, and complex numbers by value instead of pointer address
- Trace the new `equiv_fn`/`hash_fn` Value fields in GC marking and deep copy

## Test plan

- [x] 26-test Scheme regression suite (`tests/scheme/smoke/hash-table-custom-comparator.scm`): eq?, eqv?, equal?, string=?, string-ci=? modes, alist->hash-table, hash-table-copy, custom Scheme comparator, accessor correctness
- [x] Enabled 2 previously disabled audit tests for #1183
- [x] Full Scheme suite: 1807 pass, 0 fail, 0 timeout (including SRFI-125)

### DIFF
--- a/src/primitives_hashtable.zig
+++ b/src/primitives_hashtable.zig
@@ -206,8 +206,10 @@ fn stringCiContentHash(data: []const u8) usize {
             pos += 1;
             continue;
         };
-        const folded = char_mod.charFoldcase(cp);
-        h = h *% 31 +% folded;
+        const folded = char_mod.foldCharExpanding(cp);
+        for (folded.cps[0..folded.len]) |fcp| {
+            h = h *% 31 +% fcp;
+        }
         pos += seq_len;
     }
     return h;
@@ -233,9 +235,10 @@ fn hashForTable(ht: *HashTable, key: Value) PrimitiveError!usize {
             };
             if (types.isFixnum(result)) {
                 const v = types.toFixnum(result);
-                break :blk @as(usize, @intCast(if (v < 0) -v else v));
+                const abs_v: u64 = if (v < 0) @intCast(-@as(i128, v)) else @intCast(v);
+                break :blk @as(usize, @truncate(abs_v));
             }
-            break :blk @truncate(@as(u64, result) *% 2654435761);
+            break :blk valueHash(result);
         },
     };
 }
@@ -297,6 +300,13 @@ fn valueHashDepth(key: Value, depth: usize) usize {
     }
     if (types.isBignum(key)) {
         const bn = types.toObject(key).as(types.Bignum);
+        if (bn.len <= 1) {
+            const mag: u64 = if (bn.len == 0) 0 else bn.limbs[0];
+            if (mag < (@as(u64, 1) << 48)) {
+                const signed_val: i64 = if (!bn.positive and mag > 0) -@as(i64, @intCast(mag)) else @as(i64, @intCast(mag));
+                return @truncate(@as(u64, @bitCast(signed_val)) *% 2654435761);
+            }
+        }
         var h: usize = if (bn.positive) @as(usize, 0) else @as(usize, 1);
         for (bn.limbs[0..bn.len]) |limb| h = h *% 31 +% @as(usize, @truncate(limb));
         return h;
@@ -382,17 +392,13 @@ fn rehash(ht: *HashTable) PrimitiveError!void {
     }
     const old_entries = ht.entries;
     const old_cap = ht.capacity;
-    const old_count = ht.count;
-    ht.entries = new_entries;
-    ht.capacity = new_cap;
-    ht.count = 0;
+    // Keep ht.entries pointing to old_entries during the loop so GC can
+    // trace keys that haven't been moved to new_entries yet.
     const mask = new_cap - 1;
+    var new_count: usize = 0;
     for (old_entries[0..old_cap]) |entry| {
         if (entry.state == .occupied) {
             const h = hashForTable(ht, entry.key) catch |err| {
-                ht.entries = old_entries;
-                ht.capacity = old_cap;
-                ht.count = old_count;
                 gc.allocator.free(new_entries);
                 return err;
             };
@@ -401,9 +407,12 @@ fn rehash(ht: *HashTable) PrimitiveError!void {
                 idx = (idx + 1) & mask;
             }
             new_entries[idx] = entry;
-            ht.count += 1;
+            new_count += 1;
         }
     }
+    ht.entries = new_entries;
+    ht.capacity = new_cap;
+    ht.count = new_count;
     gc.allocator.free(old_entries);
 }
 
@@ -589,7 +598,9 @@ fn alistToHashTableFn(args: []const Value) PrimitiveError!Value {
     }
 
     const initial_cap = @max(count, @as(usize, 8));
-    const ht_val = gc.allocHashTable(initial_cap) catch return PrimitiveError.OutOfMemory;
+    var ht_val = gc.allocHashTable(initial_cap) catch return PrimitiveError.OutOfMemory;
+    gc.pushRoot(&ht_val);
+    defer gc.popRoot();
     const ht = types.toHashTable(ht_val);
     configureHashTable(ht, ht_val, gc, vm, args[1..]);
 
--- a/src/primitives_hashtable.zig
+++ b/src/primitives_hashtable.zig
@@ -3,12 +3,15 @@ const types = @import("types.zig");
 const vm_mod = @import("vm.zig");
 const primitives = @import("primitives.zig");
 const memory = @import("memory.zig");
+const bignum_mod = @import("bignum.zig");
+const char_mod = @import("primitives_char.zig");
 const Value = types.Value;
 const NativeFn = types.NativeFn;
 const PrimitiveError = primitives.PrimitiveError;
 const LS = primitives.LibSet;
 const HashTable = types.HashTable;
 const HashEntry = types.HashEntry;
+const CompareMode = types.CompareMode;
 
 pub const specs = [_]primitives.PrimSpec{
     .{ .name = "make-hash-table", .func = &makeHashTableFn, .arity = .{ .variadic = 0 }, .libs = LS.initOne(.srfi_69) },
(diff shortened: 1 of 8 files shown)


src/primitives_hashtable.zig
```diff
@@ -48,6 +51,205 @@ fn getHashTable(proc: []const u8, v: Value) PrimitiveError!*HashTable {
...
+            const vm = vm_mod.vm_instance orelse return PrimitiveError.TypeError;
+            const call_args = [1]Value{key};
+            const result = vm.callWithArgs(ht.hash_fn, &call_args) catch |err| {
+                return err;
+            };
+            if (types.isFixnum(result)) {
+                const v = types.toFixnum(result);
+                break :blk @as(usize, @intCast(if (v < 0) -v else v));
+            }
+            break :blk @truncate(@as(u64, result) *% 2654435761);
```
baijum: **Bug: non-fixnum custom hash results are hashed by pointer, not value → silent key loss.**

`@as(u64, result)` is the boxed Value. For a bignum/rational/complex result that's a fresh heap pointer each call, so `insert` and `lookup` of the same key compute different hashes and the key becomes unfindable (32/40 keys lost in a 40-entry repro). A custom hash returning integers > 2^47 is legal per SRFI-69.

Suggest: `break :blk valueHash(result);` — this PR's own `valueHash` now hashes bignum/rational/complex by value, making placement and lookup deterministic. (Flonum results already work since flonums are inline.)

src/primitives_hashtable.zig
```diff
@@ -48,6 +51,205 @@ fn getHashTable(proc: []const u8, v: Value) PrimitiveError!*HashTable {
...
+    };
+}
+
+pub fn hashForMode(mode: CompareMode, key: Value) usize {
+    return switch (mode) {
+        .equal, .eqv => valueHash(key),
+        .eq => identityHash(key),
+        .string_eq => stringContentHash(stringBytesOrNull(key) orelse return identityHash(key)),
+        .string_ci => stringCiContentHash(stringBytesOrNull(key) orelse return identityHash(key)),
+        .custom => valueHash(key),
```
baijum, line 252: **Bug: deep copy places `.custom` entries with `valueHash`, but lookups use the custom `hash_fn`.**

SRFI-18 deep-copies a table by re-inserting each entry at `hashForMode(mode, key)`. For `.custom` this uses `valueHash`, while every `findKey`/`findSlot` uses the custom `hash_fn` — so after a real deep copy (custom table closed over by a thread thunk, or returned as a thread result) every key is unfindable while size stays correct (`(MISS MISS 40)` in my repro). The other five modes survive deep copy correctly.

Since `hash_fn` is value-deterministic, the cleanest fix is to preserve original slot positions for `.custom` during deep copy (copy the entries array in place after deep-copying keys/values, like `hash-table-copy`) rather than recomputing. See the corresponding call at `src/gc_deep_copy.zig:171`.

baijum (commented):
## Review — Honor custom equivalence/hash in SRFI-69 (#1183)

Solid, well-structured fix for the reported problem. The five well-known comparator modes (`eq?`, `eqv?`, `equal?`, `string=?`, `string-ci=?`) are correctly recognized and dispatched to native Zig hashing/equality, the accessors now tell the truth, the SRFI-128 comparator-record extraction matches the actual field order in `lib/srfi/128.sld` (equality=field 1, hash=field 3), the `valueHash` by-value fix for bignum/rational/complex is a genuine improvement, and the GC-marking + deep-copy plumbing for the new `equiv_fn`/`hash_fn` fields is in place. I verified locally: the 26-test smoke suite passes, the 2 re-enabled audit tests pass (75/75), `zig build test` is green, and SRFI-69/125/128 conformance suites all pass.

However, I found **two silent-data-loss bugs in the arbitrary-custom-Scheme-procedure path** (the `.custom` compare mode). Both cause keys to become unfindable while `hash-table-size` still reports them present. The well-known-comparator modes — the heart of #1183 — are unaffected. I'd recommend fixing at least the first before merge.

### Bug 1 (single-threaded, higher priority): custom hash returning a non-fixnum loses keys

`hashForTable`'s `.custom` branch hashes any non-fixnum hash-function result (bignum, rational, complex) by its **boxed bit-pattern** (`@as(u64, result)`), which for a heap-allocated result is a fresh pointer each call. So `insert` and `lookup` of the same key compute different hashes and the key can't be found — no thread boundary needed. A hash function returning integers outside fixnum range (±2^47) is legal per SRFI-69.

Repro (single-threaded):
```scheme
(import (srfi 69))
(define ht (make-hash-table (lambda (a b) (= a b))
                           (lambda (k) (* k 100000000000000000000)))) ; fresh bignum per call
(do ((i 0 (+ i 1))) ((= i 40)) (hash-table-set! ht i i))
(let ((misses 0))
  (do ((i 0 (+ i 1))) ((= i 40))
    (when (eq? 'MISS (hash-table-ref/default ht i 'MISS)) (set! misses (+ misses 1))))
  (display (list 'size (hash-table-size ht) 'misses misses)))
;; prints: (size 40 misses 32)
```
Fix: reduce a non-fixnum result to a value-based hash instead of hashing its pointer — e.g. `break :blk valueHash(result);` (which, thanks to this PR, now hashes bignum/rational/complex by value). Flonum results already work because flonums are stored inline.

### Bug 2 (cross-thread, narrower): deep-copied custom tables lose all keys

SRFI-18 deep-copies a hash table by re-inserting each entry at `hashForMode(mode, key)`, but for `.custom` mode `hashForMode` returns `valueHash(key)` while every lookup uses the custom `hash_fn`. After a genuine deep copy (custom table closed over by a thread thunk, or returned as a thread result) all keys are unfindable.

Repro:
```scheme
(import (scheme base) (srfi 18) (srfi 69))
(define t
  (let ((ct (make-hash-table (lambda (a b) (string=? a b))
                             (lambda (k) (string-length k)))))
    (do ((i 0 (+ i 1))) ((= i 40)) (hash-table-set! ct (string-append "k" (number->string i)) i))
    (thread-start! (make-thread
      (lambda () (list (hash-table-ref/default ct "k5" 'MISS)
                       (hash-table-ref/default ct "k37" 'MISS)
                       (hash-table-size ct)))))))
(thread-join! t)  ;; => (MISS MISS 40)
```
The `eq?`/`eqv?`/`equal?`/`string=?`/`string-ci=?` modes all survive deep copy correctly (verified) — only `.custom` is broken. Since a custom `hash_fn` is value-deterministic, the cleanest fix is to preserve original slot positions for `.custom` during deep copy (copy the entries array in place after deep-copying keys/values, as `hash-table-copy` does) rather than recomputing with `valueHash`. At minimum, document the limitation.

### Minor notes (non-blocking)

- A `.custom` equality passed with **no** custom hash falls back to the global `hash` (equal?-based). If the custom equality unifies more keys than `equal?`, that hash is inconsistent with it and lookups silently miss. This is inherent to SRFI-69 (the caller owns hash/equality consistency), but a doc note would help.
- Re-entrancy: `findSlot`/`findKey`/`rehash` call back into the VM for `.custom` mode while holding the cached `ht.entries` pointer. A custom comparator/hash that mutates the *same* table (triggering `rehash`, which frees `ht.entries`) would leave the outer loop iterating freed memory. Very much an edge case, worth a guard or a note.

### Verdict

Not quite ready to merge as-is — Bug 1 reproduces in ordinary single-threaded code with a legal custom hash function and silently drops data. It's a small fix. Everything targeting the well-known comparators (which is what #1183 actually reports) is correct and well-tested.

baijum (commented):
## Follow-up review — fixes verified ✅

I built `9a84e54` and re-ran the repros from my earlier review. Both bugs are resolved, and the extra hardening in this commit checks out. Details:

**Bug 1 (single-threaded custom hash returning non-fixnum) — FIXED.** `hashForTable`'s `.custom` branch now falls back to `valueHash(result)`. My 40-key bignum-hash repro went from **32 misses → 0**.

**Bug 2 (deep-copied custom table) — FIXED.** Deep copy now preserves slot positions for `.custom` mode (copying tombstones too, so the probe structure is reproduced) instead of re-hashing with `valueHash`. Both the closed-over-local and thread-result repros now return `(5 37 40)` (was `(MISS MISS 40)`). The other five modes still survive deep copy.

**Additional hardening — all verified:**
- `alist->hash-table` roots `ht_val` during construction — a custom hash callback can no longer collect the half-built table. Verified: 60-entry custom-hash alist under `-Dgc-stress=true` → 0 misses.
- `rehash` defers the `ht.entries` swap until after the loop so GC can still trace old entries during a custom-hash callback. gc-stress suite green.
- `stringCiContentHash` now uses expanding folds (`foldCharExpanding`) matching `equalForTable`, so `ß`/`ss` hash consistently — verified `(make-hash-table string-ci=? string-ci-hash)` finds a `"ß"` key via both `"ss"` and `"SS"` (would have been a MISS before this commit).
- i128 intermediate for the min-fixnum negation — no overflow.

**Test status (re-run locally):** `zig build test` green; smoke 26/26; audit 75/75; SRFI-69 4/4, SRFI-69-ext 26/26, SRFI-125 39/39, SRFI-128 59/59. All custom-hash paths also pass under `-Dgc-stress=true`.

### One minor, non-blocking nit

The new bignum-in-fixnum-range branch in `valueHash` doesn't literally achieve its stated goal of hashing "identically to the equivalent fixnum": it hashes `@bitCast(signed_val)` (the raw i64, e.g. `5`), whereas a real fixnum hashes its **NaN-boxed** Value (`0xFFFD0000000000005`) via the fall-through at the end of `valueHashDepth`. So `hash(bignum 5) = 5 *% C` but `hash(fixnum 5) = 0xFFFD…0005 *% C` — different. In practice this is unreachable: every arithmetic/bitwise op I tried (`+`/`-`/`quotient`/`modulo`/`gcd`/`abs`/`logand`/`ash`/`numerator`) demotes single-limb in-range results to fixnums via `demote`/`makeBignumValue`, so an in-range bignum never reaches the hasher and no equal?/eqv? lookup actually crosses representations. So: harmless defensive code, not a blocker. If you want it to truly match, hash `types.makeFixnum(signed_val)` instead of `@bitCast(signed_val)` (and it could be dropped entirely if in-range bignums are guaranteed impossible).

(Also still theoretically present, as before: a custom comparator/hash that mutates the *same* table mid-lookup could free `ht.entries` under the `findSlot`/`findKey` loop. Pathological and shared with most open-addressed tables — non-blocking.)

### Verdict: ready to merge 🚢

Both blocking bugs are fixed and verified end-to-end, the added GC-safety and hash-consistency fixes are sound, and the full hash-table test surface is green. The only outstanding item is the cosmetic nit above, which is non-blocking.

PR #1331, opened 2026-07-08
Fix SRFI-152 string-every, string-split, and add missing exports

### LINKED ISSUE #1234: [SRFI-152] string-every discards the final predicate value; string-split lacks grammar/limit and mishandles empty string; ~24 spec exports missing
**Category:** SRFI (lib/srfi/152.sld)
**Spec reference:** SRFI-152:
- string-every: "If string-every returns true for a non-empty interval (with start < end), the returned true value is the one returned by the final call to the predicate."
- string-split: "string-split string delimiter [grammar limit start end] -> list ... If it is infix, which is the default, processing is done as described above, except an empty string produces the empty list."
**Severity:** wrong-result + missing API

```scheme
(import (scheme base) (srfi 152))
(string-every (lambda (c) (char->integer c)) "abc")
;; expected 99 (last predicate value)   actual #t
(string-split "" ",")
;; expected ()   actual ("")
(string-split "a b c" " " 'infix 2)
;; expected ("a" "b c") (limit)   actual arity error
```

**Missing exports** (spec list minus implementation): string-null?, string-unfold, string-unfold-right, string->vector, vector->string, reverse-list->string, string-ci=?, string-ci<?, string-ci>?, string-ci<=?, string-ci>=?, string-prefix-length, string-suffix-length, string-contains-right, string-take-while, string-take-while-right, string-drop-while, string-drop-while-right, string-break, string-span, string-concatenate-reverse, string-fold, string-fold-right, string-filter, string-remove, string-replicate, string-segment, read-string, write-string. (The library also exports string-reverse, which is not part of SRFI-152.)

**Root cause:** lib/srfi/152.sld's string-every ANDs the loop continuation over the predicate result instead of returning the final application's value; string-split is a fixed 2-arg literal-delimiter splitter.

**Conforms:** searching (contains/prefix/suffix/index/skip/count with start/end), selection (take/drop/pad/trim), replace, join/concatenate, tabulate, string-any returning the predicate's true value, codepoint-indexed multibyte handling.

**Disabled tests:** tests/scheme/srfi/srfi152.scm (marked ;; FAIL: #this)

baijum:
## Summary

Closes #1234.

- **`string-every`** now returns the final predicate value instead of `#t`, by importing the correct native SRFI-13 implementation
- **`string-split`** now supports `grammar` (infix/strict-infix/prefix/suffix) and `limit` parameters; empty string with infix grammar correctly returns `()`
- **Added 28 missing procedures** to reach the full 73-procedure SRFI-152 spec: `string-null?`, `reverse-list->string`, `string-prefix-length`, `string-suffix-length`, `string-contains-right`, `string-take-while`, `string-take-while-right`, `string-drop-while`, `string-drop-while-right`, `string-break`, `string-span`, `string-concatenate-reverse`, `string-fold`, `string-fold-right`, `string-filter`, `string-remove`, `string-replicate`, `string-segment`, `string-unfold`, `string-unfold-right`, CI comparisons, `string->vector`, `vector->string`, `read-string`, `write-string`
- **Removed `string-reverse`** from exports (not part of SRFI-152)
- Rewrote library to import native SRFI-13 implementations where they exist, implementing only the 17 truly missing procedures in Scheme

## Test plan

- [x] 128 SRFI-152 tests pass (up from ~60, all FAIL markers removed)
- [x] Zig unit tests pass
- [x] Full Scheme test suite: 1806 pass, 0 fail
- [ ] CI


lib/srfi/152.sld
```diff
@@ -1,161 +1,228 @@
 (define-library (srfi 152)
-  (import (scheme base) (scheme char) (scheme case-lambda) (scheme write))
+  (import (scheme base) (scheme char) (scheme write)
```
baijum: **Missing `(scheme cxr)` import — crashes on spec-defined optional arguments.**

`string-prefix-length` (line 73), `string-suffix-length` (line 88), and `string-split` (line 189) all reference `caddr`/`cdddr`/`cadddr` when their optional `start`/`end` arguments are supplied, but those accessors live in `(scheme cxr)` / `(scheme r5rs)` — not `(scheme base)`. The basic tests pass only because they never take those branches.

Reproduction (against this branch):

```scheme
(import (scheme base) (srfi 152))
(string-prefix-length "abcdef" "abcxyz" 0 3 0 3)
;; => error: undefined variable 'caddr'
(string-split "a,b,c,d" "," 'infix #f 2 7)
;; => error: undefined variable 'caddr'
```

One-line fix — add `(scheme cxr)` to the import list:

```scheme
  (import (scheme base) (scheme char) (scheme write) (scheme cxr)
          (only (srfi 13)
```

I verified this fix locally: the three calls above return `3`, `2`, and `("b" "c" "d")`, and all 128 SRFI-152 tests still pass.

baijum (commented):
## Review — SRFI-152 fixes

Nice, focused fix. I checked it out and ran it: **all 128 tests pass**, and the three reproductions from #1234 are all resolved:

```scheme
(string-every (lambda (c) (char->integer c)) "abc")   ; => 99  ✓
(string-split "" ",")                                 ; => ()  ✓
(string-split "a b c" " " 'infix 2)                   ; => ("a" "b c")  ✓
```

Re-importing the native (srfi 13) implementations (which use pred-first argument order, matching SRFI-152) is the right call, and the grammar/limit `string-split` rewrite checks out against the spec (infix/prefix/suffix/strict-infix all traced correctly). `(scheme case-lambda)` was correctly dropped since it's no longer used, and `string-segment` / string-string delimiter both match the spec (verified against srfi.schemers.org/srfi-152).

### 🔴 One bug to fix before merge — see inline comment

Missing `(scheme cxr)` import → runtime crash on the spec's optional `start`/`end` arguments of `string-prefix-length`, `string-suffix-length`, and `string-split`. CodeRabbit flagged this as a validity nit; it's actually a reproducible runtime error. Details and the one-line fix are in the inline comment.

### 🟡 Minor / non-blocking

Six locally-defined procedures omit the optional `[start end]` (or `[start1 end1 start2 end2]`) arguments the spec gives them: `string-contains-right`, `string-take-while`, `string-take-while-right`, `string-drop-while`, `string-drop-while-right`, `string-break`, `string-span`. No test exercises these and the common 2-arg forms are correct, so this is a conformance gap rather than a defect — fine to defer, but worth a note given the PR's "full 73-procedure spec" goal.

**Verdict:** Not quite mergeable as-is — add the `(scheme cxr)` import (one line, confirmed to fix the crash and keep all 128 tests green), and it's good to go.

baijum:
## Re-review of `08ab9cac` — both findings resolved ✅

Checked out the new commit and re-ran everything:

- 🔴 **`(scheme cxr)` crash — fixed.** The import was added, and the three previously-crashing calls now work: `string-prefix-length "abcdef" "abcxyz" 0 3 0 3` → `3`, `string-suffix-length … 0 5 0 3` → `2`, `string-split "a,b,c,d" "," 'infix #f 2 7` → `("b" "c" "d")`.
- 🟡 **Optional `[start end]` args — added and correct.** All 7 procedures (`string-contains-right`, `string-take-while(-right)`, `string-drop-while(-right)`, `string-break`, `string-span`) now take the spec's optional bounds and thread `start`/`end` through the underlying `string-skip`/`string-index` correctly. I verified sub-range behavior independently (e.g. `string-drop-while-right "1ab2cd" char-alphabetic? 1 6` → `"ab2"`, `string-split "a,b,c,d" "," 'infix #f 2 5` → `("b" "c")`) — all correct.
- **Tests:** SRFI-152 suite now **134 pass, 0 fail** (7 new tests added).
- **CI:** all build/test jobs green (macOS, Linux arm/x86 across Debug/ReleaseSafe/ReleaseFast, riscv64, wasm, format).

**LGTM — ready to merge.** (Only the CodeRabbit re-review shows pending; not a blocking gate.)

PR #1333, opened 2026-07-08
Fix SRFI-233 ini-file->alist missing (scheme char) import (#1223)

### LINKED ISSUE #1223: [SRFI-233] ini-file->alist fails on any non-empty input: char-whitespace? unbound ((scheme char) not imported)
**Category:** SRFI
**Severity:** wrong-result (every parse fails; error escapes guard)

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 233))
(ini-file->alist (open-input-string "[s]\nk=v\n"))
;; expected: (("s" ("k" . "v")))
;; actual: error: undefined variable 'char-whitespace?' (from the library's internal string-trim)
```
Only empty input works (never reaches string-trim). `alist->ini-file` (the writer) is unaffected.

**Root cause:** lib/srfi/233.sld imports only `(scheme base)` and `(scheme write)`; its internal `string-trim` calls `char-whitespace?`, which lives in `(scheme char)`. Same class as #1212 (SRFI-134's unbound `filter`).

**Source location:** lib/srfi/233.sld
**Disabled test:** tests/scheme/srfi/srfi233.scm (marked ;; FAIL: #this)

baijum:
## Summary
- Add missing `(scheme char)` import to `lib/srfi/233.sld` — the library's internal `string-trim` calls `char-whitespace?` which lives in `(scheme char)`, causing every parse of non-empty input to fail with an unbound variable error
- Uncomment and enable all previously disabled parser tests in `tests/scheme/srfi/srfi233.scm`, migrate from `(chibi test)` to `(srfi 64)` with proper exit-on-failure

Closes #1223

## Test plan
- [x] All 11 SRFI-233 tests pass (empty input, writer, single section, multi-section, comments/blanks, round trip)
- [x] Exact reproduction from issue returns expected `(("s" ("k" . "v")))`

### DIFF
--- a/tests/scheme/srfi/srfi233.scm
+++ b/tests/scheme/srfi/srfi233.scm
@@ -1,45 +1,47 @@
-;; SRFI-233 (INI files) conformance tests — audit Phase 3e
-;; ini-file->alist fails on any non-empty input (#1223): the library's
-;; string-trim calls char-whitespace? without importing (scheme char).
+;; SRFI-233 (INI files) conformance tests
+;; Regression test for #1223: char-whitespace? was unbound because
+;; (scheme char) was not imported.
 ;; Run directly: zig-out/bin/kaappi tests/scheme/srfi/srfi233.scm
 
-(import (scheme base) (srfi 233) (chibi test))
+(import (scheme base) (srfi 233) (srfi 64))
 
 (test-begin "srfi-233")
 
 (define (parse str)
   (ini-file->alist (open-input-string str)))
 
-;; empty input works (never reaches the broken string-trim)
-(test '() (parse ""))
+;; empty input
+(test-equal "empty input" '() (parse ""))
 
-;;; --- the writer is unaffected ---
+;;; --- the writer ---
 (define (unparse alist)
   (let ((p (open-output-string)))
     (alist->ini-file alist p)
     (get-output-string p)))
 
-(test "[sec]\nk = v\n\n" (unparse '(("sec" ("k" . "v")))))
-(test "" (unparse '()))
+(test-equal "write single section" "[sec]\nk = v\n\n" (unparse '(("sec" ("k" . "v")))))
+(test-equal "write empty" "" (unparse '()))
 
-;;; --- parsing (all blocked on #1223) ---
-;; FAIL: #1223 (ini-file->alist: char-whitespace? unbound — every parse fails)
-;; (let ((r (parse "[server]\nhost=localhost\nport=8080\n")))
-;;   (test 1 (length r))
-;;   (test "server" (car (car r)))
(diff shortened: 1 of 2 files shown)


baijum (commented):

Verified locally against a fresh ReleaseSafe build with the PR head checked out (`bd20e734`).

**Fix is correct and minimal.** Adding `(scheme char)` to the `(srfi 233)` import list resolves the unbound `char-whitespace?` that the library's internal `string-trim` depends on. The exact reproduction from #1223 now returns the expected result:

```
$ kaappi -e '(write (ini-file->alist (open-input-string "[s]\nk=v\n")))'
(("s" ("k" . "v")))
```

**Tests.** The migration from `(chibi test)` to `(srfi 64)` with exit-on-failure is a solid improvement, and re-enabling the four previously-disabled parse blocks (single/multi-section, comments+blanks, round-trip) gives real coverage of the fixed path. All 11 assertions pass, and I confirmed the whole `tests/scheme/srfi/` directory (77 files) still passes with no regressions.

**Minor nit (non-blocking):** the test calls `(exit 1)` but imports only `(scheme base) (srfi 233) (srfi 64)` — `exit` lives in `(scheme process-context)`. It works here because Kaappi exposes `exit` globally, but the `tests/scheme/CLAUDE.md` template and the other SRFI-64 tests (srfi175, srfi189, srfi18-atomic-stress) import `(scheme process-context)` explicitly. Worth adding for consistency and portability, but not a blocker.

No CHANGELOG entry needed — consistent with the other recent SRFI import-fix PRs (#1329, #1331).

tests/scheme/srfi/srfi233.scm
```diff
@@ -1,45 +1,47 @@
-;; SRFI-233 (INI files) conformance tests — audit Phase 3e
-;; ini-file->alist fails on any non-empty input (#1223): the library's
-;; string-trim calls char-whitespace? without importing (scheme char).
+;; SRFI-233 (INI files) conformance tests
+;; Regression test for #1223: char-whitespace? was unbound because
+;; (scheme char) was not imported.
 ;; Run directly: zig-out/bin/kaappi tests/scheme/srfi/srfi233.scm
 
-(import (scheme base) (srfi 233) (chibi test))
+(import (scheme base) (srfi 233) (srfi 64))
```
baijum, line 6: Nit (non-blocking): `(exit 1)` at the end lives in `(scheme process-context)`, which isn't imported here. It happens to work because Kaappi makes `exit` globally available, but for consistency with `tests/scheme/CLAUDE.md` and the other SRFI-64 tests, consider:

```scheme
(import (scheme base) (scheme process-context) (srfi 233) (srfi 64))
```

PR #1334, opened 2026-07-08
Fix SRFI-141 balanced/ tie-breaking (#1232)

### LINKED ISSUE #1232: [SRFI-141] balanced/ is aliased to round/ — remainder escapes [-|d/2|, |d/2|) whenever round ties break upward
**Category:** SRFI (lib/srfi/141.sld)
**Spec reference:** SRFI-141 balanced division: "This division operator pair satisfies the property -|d/2| <= r < |d/2|." Round division: "q = round(n/d)" with ties to even, and "Nothing general can be said about the sign of r."
**Severity:** wrong-result (silent)

```scheme
(import (scheme base) (srfi 141))
(call-with-values (lambda () (balanced/ 5 2)) list)
;; expected (3 -1)   actual (2 1)   — r=1 violates r < |d/2| = 1
(call-with-values (lambda () (balanced/ -7 2)) list)
;; expected (-3 -1)  actual (-4 1)
(call-with-values (lambda () (balanced/ -7 -2)) list)
;; expected (3 -1)   actual (4 1)
```
Ties-to-even picks the wrong quotient exactly when the true quotient is an odd-integer-plus-half whose even neighbor leaves r = +|d|/2 (excluded from the half-open interval). Non-tie cases agree, which is why smoke tests pass.

**Root cause:** lib/srfi/141.sld defines `(define balanced-quotient round-quotient)` etc. Balanced division must break ties so the remainder stays in [-|d|/2, |d|/2) — i.e. always round half-up in quotient magnitude direction appropriate to sign(d), not half-to-even.

**Conforms:** floor/, ceiling/, truncate/, round/, euclidean/ verified on the full ±7/±2 sign grid, bignums, and zero dividends.

**Disabled tests:** tests/scheme/srfi/srfi141.scm (marked ;; FAIL: #this)

baijum:
## Summary

- Replace `balanced-quotient`/`balanced-remainder` aliases to `round-quotient`/`round-remainder` with a proper implementation
- At ties (`2*|r| = |d|`), balanced always picks the quotient yielding a negative remainder to satisfy `r ∈ [-|d/2|, |d/2|)`, whereas round picks the even quotient
- Enable previously disabled tests and add full sign-grid coverage (all 8 combinations of `±n / ±d` at ties)

## Test plan

- [x] SRFI-141 test suite passes (44/44)
- [x] All three issue reproductions return correct results
- [x] Zig unit tests pass
- [x] R7RS test suite passes (1391/1391)

Closes #1232


baijum (commented):
## Review — looks correct and ready to merge ✅

I checked out the PR head, rebuilt, and verified the fix end-to-end.

### Correctness
The tie-breaking is right. SRFI-141 requires the balanced remainder in `[-|d|/2, |d|/2)`, so at a tie (`2|r| = |d|`) the only admissible remainder is `-|d|/2` (the closed end) — i.e. the remainder must be **negative**. The implementation achieves that by adjusting the truncated quotient only when `n > 0` (where the truncated remainder is `+|d|/2`) and leaving it alone when `n ≤ 0` (already `-|d|/2`). I walked all 8 sign combinations by hand and they match the new tests.

Nice touch: the structure mirrors `round-quotient` exactly — same `let*`, same first two `cond` clauses — differing only in the tie branch (`round` → `even?`, `balanced` → sign of `n`). That parallelism makes the distinction easy to read.

### Verification I ran (on the PR head)
- **SRFI-141 suite:** 44/44 pass
- **Exhaustive property test** vs an independent reference (smallest `q` whose remainder lands in range) over `n ∈ [-40,40] × d ∈ [-12,12]` plus a few bignums: **0 mismatches**
- **Range/reconstruction invariant** (`n = q·d + r` and `-|d|/2 ≤ r < |d|/2`) holds for every case. For reference, the base branch violates the range at ties (e.g. `balanced/ 5 2` → `(2 1)` with `r=1`), confirming the bug this fixes.
- **Zig unit tests:** pass · **R7RS suite:** 1395/1395

### Test coverage
The new tests cover all 8 `±n / ±d` tie cases, a non-tie case, and the standalone `balanced-quotient`/`balanced-remainder` entry points. Good grid coverage, and the inline comments accurately describe each tie's behavior.

### Minor (non-blocking, optional)
For `n > 0`, the tie branch `(if (> d 0) (+ q 1) (- q 1))` is logically identical to the `>` clause's `(if (eq? (< n 0) (< d 0)) (+ q 1) (- q 1))`, so the whole thing could collapse to "adjust unless `n ≤ 0` at a tie." But the current explicit form is clearer about *why* ties keep a negative remainder, so I'd leave it as-is.

**Verdict:** ready to merge.

PR #1335, opened 2026-07-08
Fix SRFI-128 default comparator total order, hashability, and register-default!

### LINKED ISSUE #1230: [SRFI-128] Default comparator is not a total order (pairs/vectors/cross-type unordered); eq/eqv comparators not hashable; comparator-if<=> comparator not optional; comparator-register-default! is a no-op
**Category:** SRFI (lib/srfi/128.sld)
**Spec reference:** SRFI-128:
- "Given disjoint types a and b, one of three conditions must hold: All objects of type a compare less than all objects of type b. All objects of type a compare greater than all objects of type b. All objects of both type a and type b compare equal to each other."
- "When comparing pairs, it must behave the same as a comparator returned by make-pair-comparator with default comparators as arguments."
- make-eq-comparator/make-eqv-comparator: "The hash function is default-hash." (post-finalization note: implementation-defined — but absent entirely means comparator-hashable? is #f and the comparators are unusable for hashing)
- "(comparator-if<=> [ <comparator> ] <object1> <object2> <less-than> <equal-to> <greater-than>) ... If <comparator> is omitted, a default comparator is used."
- comparator-register-default!: "Registers comparator for use by default comparators, such that if the objects being compared both satisfy the type test predicate of comparator, it will be employed by default comparators to compare them."
**Severity:** wrong-result (silent) + missing behavior

```scheme
(import (scheme base) (srfi 128))
(define dc (make-default-comparator))
(or (<? dc '(1 2) '(1 3)) (<? dc '(1 3) '(1 2)))  ;; expected #t  actual #f (pairs unordered, trichotomy violated)
(or (<? dc #(1) #(2)) (<? dc #(2) #(1)))          ;; expected #t  actual #f
(or (<? dc 'a "a") (<? dc "a" 'a))                ;; expected #t per total-order rule  actual #f (also not "equal")
(comparator-hashable? (make-eq-comparator))       ;; expected #t  actual #f
(comparator-if<=> 1 2 'less 'equal 'greater)      ;; expected less  actual compile error (no 5-operand pattern)
```

**Also:** default-hash on pairs is not reduced modulo (hash-bound): `(default-hash '(1 (2 . 3) #(4) "five"))` returns 93643427476 while (hash-bound) is 33554432 — the library's own string/char/number hashes all reduce mod bound, the pair branch (`(+ car-hash (* 31 cdr-hash))`) does not.

**Root cause:** lib/srfi/128.sld `default-ordering` handles only same-type booleans/chars/numbers/strings/symbols and returns #f otherwise; make-eq/eqv-comparator pass hash #f; comparator-if<=> has only the 6-operand syntax-rules pattern; comparator-register-default! is `(define (comparator-register-default! cmp) #t)`.

**Conforms:** accessors, =?/<?/>?/<=?/>=? 2- and 3-arg chains, comparator-check-type raising, #t type-test/#t-equality conventions, hash-salt < hash-bound.

**Disabled tests:** tests/scheme/srfi/srfi128.scm (marked ;; FAIL: #this)

baijum:
## Summary
- **Total-order `default-ordering`**: pairs (lexicographic), vectors (element-wise), bytevectors (byte-wise), and cross-type comparisons via `type-index`
- **Hashable eq/eqv comparators**: `make-eq-comparator`/`make-eqv-comparator` now use `default-hash`
- **`comparator-if<=>` 5-arg form**: optional comparator defaults to `(make-default-comparator)`
- **`comparator-register-default!`**: stores comparators, wires into `default-ordering`, `default-equality`, and `default-hash`
- **`default-hash` modulo**: pair/vector/bytevector hashes reduced modulo `(hash-bound)`; vector hash improved to use all elements

## Test plan
- [x] All 92 SRFI-128 tests pass (7 previously disabled tests now enabled + 15 new tests)
- [x] Full Scheme test suite: 1806 pass, 0 fail
- [x] Zig unit tests pass

Closes #1230

### DIFF
--- a/lib/srfi/128.sld
+++ b/lib/srfi/128.sld
@@ -60,40 +60,139 @@
     (define (string-ci-hash s) (string-hash (string-downcase s)))
     (define (symbol-hash s) (string-hash (symbol->string s)))
 
+    ;; Registered comparators for extending the default comparator
+    (define registered-comparators '())
+
+    (define (find-registered-comparator a b)
+      (let loop ((cmps registered-comparators))
+        (if (null? cmps)
+            #f
+            (if (and (comparator-test-type (car cmps) a)
+                     (comparator-test-type (car cmps) b))
+                (car cmps)
+                (loop (cdr cmps))))))
+
+    (define (find-registered-hash obj)
+      (let loop ((cmps registered-comparators))
+        (if (null? cmps)
+            #f
+            (if (and (comparator-test-type (car cmps) obj)
+                     (comparator-hashable? (car cmps)))
+                (car cmps)
+                (loop (cdr cmps))))))
+
     (define (default-hash obj)
+      (let ((reg (find-registered-hash obj)))
+        (if reg
+            (comparator-hash reg obj)
+            (cond
+              ((boolean? obj) (boolean-hash obj))
+              ((char? obj) (char-hash obj))
+              ((number? obj) (number-hash obj))
+              ((string? obj) (string-hash obj))
+              ((symbol? obj) (symbol-hash obj))
+              ((null? obj) 0)
+              ((pair? obj)
+               (modulo (+ (default-hash (car obj))
+                          (* 31 (default-hash (cdr obj))))
+                       (hash-bound)))
+              ((vector? obj)
+               (let loop ((i 0) (h 0))
+                 (if (= i (vector-length obj))
+                     (modulo h (hash-bound))
+                     (loop (+ i 1)
+                           (+ (* h 31) (default-hash (vector-ref obj i)))))))
+              ((bytevector? obj)
+               (let loop ((i 0) (h 0))
+                 (if (= i (bytevector-length obj))
+                     (modulo h (hash-bound))
+                     (loop (+ i 1)
+                           (+ (* h 31) (bytevector-u8-ref obj i))))))
+              (else 0)))))
+
+    ;; Type index for cross-type total ordering
+    (define (type-index obj)
       (cond
-        ((boolean? obj) (boolean-hash obj))
-        ((char? obj) (char-hash obj))
-        ((number? obj) (number-hash obj))
-        ((string? obj) (string-hash obj))
-        ((symbol? obj) (symbol-hash obj))
         ((null? obj) 0)
-        ((pair? obj) (+ (default-hash (car obj)) (* 31 (default-hash (cdr obj)))))
-        ((vector? obj) (if (= (vector-length obj) 0) 0
-                           (default-hash (vector-ref obj 0))))
-        (else 0)))
+        ((boolean? obj) 1)
+        ((char? obj) 2)
+        ((number? obj) 3)
+        ((string? obj) 4)
+        ((symbol? obj) 5)
+        ((bytevector? obj) 6)
+        ((pair? obj) 7)
+        ((vector? obj) 8)
+        (else 9)))
+
+    ;; Lexicographic ordering for compound types
+    (define (pair-ordering a b)
+      (cond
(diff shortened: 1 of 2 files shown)


baijum:
## Review — SRFI-128 default comparator fixes

Verified by checking out the PR head, building (`zig build`), and running the suites.

**Correctness — every #1230 reproduction is fixed** ✅
- pairs / vectors / bytevectors now totally ordered (lexicographic / element-wise / byte-wise)
- cross-type ordering via `type-index`
- `make-eq-comparator` / `make-eqv-comparator` hashable through `default-hash`
- `comparator-if<=>` 5-arg (optional-comparator) form
- `comparator-register-default!` actually registers and is wired into ordering, equality, and hash
- `default-hash` for pair/vector/bytevector reduced mod `(hash-bound)`

**Test results**
- `srfi128.scm`: **92 pass, 0 fail**
- Dependents unaffected: `srfi113` 133/0, `srfi125` 39/0, `srfi146` 82/0
- `zig build test`: pass

---

### One limitation worth noting (non-blocking): cross-*registered*-type trichotomy gap

When two comparators are registered for **different** types, an object of one compared against the other is neither `<`, `>`, nor `=`:

```scheme
(define-record-type <pt> (make-pt x) pt? (x pt-x))
(define-record-type <qt> (make-qt y) qt? (y qt-y))
(comparator-register-default! (make-comparator pt? = < #f))  ; sketch
(comparator-register-default! (make-comparator qt? = < #f))
(<? dc (make-pt 1) (make-qt 1))  ; => #f
(<? dc (make-qt 1) (make-pt 1))  ; => #f
(=? dc (make-pt 1) (make-qt 1))  ; => #f   ← trichotomy violated
```

Root cause: every non-standard type collapses to `type-index` 9, and `find-registered-comparator` only fires when **both** operands match the *same* registered comparator. The SRFI-128 sample implementation avoids this by giving each registered type its own type index, so cross-type ordering stays total.

This is outside #1230's explicit acceptance criteria (standard types + a single registered type — all of which pass), and SRFI-128 only *defines* behavior when both objects satisfy the same registered predicate, so it's arguably conformant. But it does weaken the "total order" guarantee once ≥2 types are registered. Fine to land as-is and track as a follow-up if desired.

**Minor:** the 5-arg `comparator-if<=>` allocates a fresh `(make-default-comparator)` on each call; negligible, just noting.

---

**Verdict: ready to merge.** The fix is correct, thoroughly tested, and regression-free for the targeted scope.

baijum:
## Re-review — after `4b3d583` (guard unordered registered comparators)

Rebuilt the latest head and re-ran everything.

**New commit does what it says** ✅ — guarding `default-ordering` with `(and reg (comparator-ordered? reg))` fixes a real crash: registering a comparator with `#f` ordering previously routed through `find-registered-comparator` straight into the `"comparator: no ordering"` error thunk. Now it falls through to the `type-index` path instead, and the registered *equality* is still honored:

```
(comparator-register-default! (make-comparator uq? = #f #f))
(<? dc (make-uq 1) (make-uq 2))  ; => #f  (was: error)
(=? dc (make-uq 1) (make-uq 1))  ; => #t
```

**Test results (latest head)**
- `srfi128.scm`: **94 pass, 0 fail** (2 new)
- Dependents unaffected: `srfi113` 133/0, `srfi125` 39/0, `srfi146` 82/0
- `zig build test`: pass

**Note:** this commit addresses a *different* edge case than the cross-registered-type gap from my first review — that one (two **different** ordered registered types compared against each other yielding `#f`/`#f`/`#f`) is still present and remains an optional follow-up, not a blocker. It's outside #1230's scope and arguably SRFI-conformant.

**Verdict: still ready to merge** — the extra hardening is a net improvement.

PR #1336, opened 2026-07-08
Fix SRFI-4 integer vector kinds to be disjoint types with range validation (#1225)

### LINKED ISSUE #1225: [SRFI-4] Integer vector kinds are indistinguishable bytevector aliases — predicates don't discriminate, signed setters accept out-of-range values
**Category:** SRFI
**Spec reference:** SRFI-4 defines u8/s8/u16/s16/u32/s32/f32/f64 vectors as "a new datatype" each with its own predicate — distinct types with checked element ranges ("it is an error" to store out-of-range values, and quality implementations reject them since the representation cannot round-trip them).
**Severity:** wrong-result

```scheme
(import (scheme base) (srfi 4))
(s8vector? (u8vector 1))    ;; expected: #f   actual: #t  (both are raw bytevectors)
(u16vector? (u8vector 1))   ;; expected: #f   actual: #t
(u16vector-ref (u8vector 1 2) 0)  ;; silently returns 513 — cross-kind reads "work"

;; signed setters do no range checking:
(let ((v (s8vector 1)))
  (s8vector-set! v 0 128)   ;; expected: error (range is -128..127)
  (s8vector-ref v 0))       ;; actual: -128 — the value silently wrapped
```

**Root cause:** lib/srfi/4.sld represents every integer vector kind as a plain bytevector (`(define s8vector? bytevector?)`, `(define u16vector? ...)` etc.), so predicates cannot discriminate kinds or element widths, cross-kind accessors reinterpret bytes, and the s8/s16/s32 wrappers convert negatives but never validate range (out-of-range positives wrap on read). f32/f64 use a separate representation, which is why `f64vector?` does discriminate from the integer kinds.

Identifying u8vector with bytevector is a common and reasonable choice; the problem is that *all* integer kinds alias the same untagged type.

**Source location:** lib/srfi/4.sld
**Disabled test:** tests/scheme/srfi/srfi4.scm (marked ;; FAIL: #this)

baijum:
## Summary
- Wraps each non-u8 SRFI-4 vector kind (s8, u16, s16, u32, s32, f32, f64) in a `define-record-type`, making predicates disjoint — `(s8vector? (u8vector 1))` now correctly returns `#f`
- Adds range validation to all setters — `(s8vector-set! v 0 128)` now raises an error instead of silently wrapping to -128
- u8vector stays as a direct bytevector alias per SRFI-4 spec; pure Scheme fix, no Zig changes

Closes #1225

## Test plan
- [x] All 3 previously-disabled FAIL tests now pass
- [x] 71 SRFI-4 tests pass (expanded from 38 to cover disjointness, range errors, make-* with fill, mutation)
- [x] Full test suite: 1806 pass, 0 fail, 0 regressions


baijum:
## Review — SRFI-4 disjoint types (#1225)

Reviewed the diff and built the branch at `63ac656`. The core fix is **correct and well-tested**: each non-u8 kind is now a distinct `define-record-type`, predicates discriminate (`(s8vector? (u8vector 1))` → `#f`), cross-kind accessors error instead of silently reinterpreting bytes, and every setter/constructor now range-checks. Verified locally: the SRFI-4 suite runs **71 pass / 0 fail**, and no other file in `lib/` or `tests/` imports `(srfi 4)`, so blast radius is contained. Range bounds are all correct (s8 −128..127, u16 0..65535, s16 −32768..32767, u32 0..2³²−1, s32 ±2³¹).

### 1. `equal?` no longer compares non-u8 vectors by content (behavior regression)

This is the one thing worth a decision before merge. On `main` every non-u8 kind was a bytevector (integer kinds) or a plain vector (f32/f64), so `equal?` compared them structurally. Wrapping them in records makes `equal?` fall back to identity:

```scheme
(equal? (s8vector 1 2 3) (s8vector 1 2 3))   ; main: #t  → this PR: #f
(equal? (f64vector 1.0 2.0) (f64vector 1.0 2.0)) ; main: #t  → this PR: #f
(equal? (u8vector 1 2) (u8vector 1 2))        ; still #t (u8 stays a bytevector)
```

R7RS technically leaves `equal?` on records as identity, so `#f` isn't *wrong* — but it's a silent regression, and it leaves `u8vector` as the odd one out that still content-compares. Programs that compare numeric vectors will quietly start returning `#f`. Worth either accepting-and-documenting this trade-off in the file header, or noting it in the changelog / known-limitations. (SRFI-4 itself doesn't mandate `equal?` semantics, so this is a judgment call, not a blocker.)

### 2. Nit — `write`/`display` output is a bit ugly

```
> (write (s8vector 1 2 3))
#<<s8vector> #u8(1 2 3)>
```

The record type name `<s8vector>` combined with the printer's `#<…>` wrapper yields the doubled `#<<…>>`, and it leaks the internal bytevector. Purely cosmetic and pre-existing-style, but if you wanted cleaner output you could name the record type `s8vector` (no angle brackets). Not blocking.

### 3. Nit — range checks use `integer?` rather than `exact-integer?`

`(s8vector-set! v 0 5.0)` passes the `(integer? val)` guard (since `(integer? 5.0)` is `#t`) and is then rejected downstream by `bytevector-u8-set!` with a less-obvious type error rather than the clean "out of range" message. It still errors (no silent corruption — I checked u16/u32 too), so this is only message quality. `exact-integer?` would be slightly tighter.

### Verdict

Solid, correctly scoped fix that closes #1225. **Mergeable as-is** — item 1 is the only thing I'd ask you to consciously accept (and ideally document) rather than change; items 2–3 are optional polish.

PR #1337, opened 2026-07-08
Fix SRFI-125 hash-table-ref/update!/find bugs and add 21 missing exports

### LINKED ISSUE #1229: [SRFI-125] hash-table-ref/update! ignore the success procedure; hash-table-find returns (key . value) instead of proc's result; 20 exports missing
**Category:** SRFI (lib/srfi/125.sld)
**Spec reference:** SRFI-125:
- hash-table-ref: "Extracts the value associated to key in hash-table, invokes the procedure success on it, and returns its result; if success is not provided, then the value itself is returned."
- hash-table-find: "For each association of hash-table, invoke proc on its key and value. If proc returns true, then hash-table-find returns what proc returns."
**Severity:** wrong-result + missing API

```scheme
(import (scheme base) (srfi 125) (srfi 128))
(define ht (make-hash-table (make-equal-comparator)))
(hash-table-set! ht 'x 12)
(hash-table-ref ht 'x (lambda () 'no) (lambda (v) (* v 100)))
;; expected 1200   actual 12 (success proc silently ignored)
(hash-table-find (lambda (k v) (and (= v 12) (* v 10))) ht (lambda () 'nope))
;; expected 120    actual (x . 12)
```

**Missing exports** (spec list minus implementation): hash-table-unfold, alist->hash-table, hash-table-exists?, hash-table=?, hash-table-mutable?, hash-table-update!/default, hash-table-pop!, hash-table-clear!, hash-table-map, hash-table-walk, hash-table-map!, hash-table-prune!, hash-table-empty-copy, hash-table-merge!, hash-table-xor!, hash, string-hash, string-ci-hash, hash-by-identity, hash-table-equivalence-function, hash-table-hash-function.

**Root cause:** lib/srfi/125.sld is a thin SRFI-69 wrapper: hash-table-ref only handles the failure thunk; hash-table-find conses the pair itself instead of returning proc's value. Note the comparator arguments accepted by make-hash-table/hash-table only "work" because the underlying Zig primitives ignore them entirely (see #1183); hash-table-update!'s SRFI-69 counterpart is also missing (see #1182).

**Conforms:** hash-table-union! keeps ht1's value on collision per spec; intersection!/difference!, intern!, entries, fold/count/map->list argument orders.

**Disabled tests:** tests/scheme/srfi/srfi125.scm (marked ;; FAIL: #this)

baijum:
## Summary

Fixes #1229.

- **hash-table-ref** and **hash-table-update!** now handle the optional `success` procedure — `(hash-table-ref ht key failure success)` calls `(success value)` when the key is present
- **hash-table-find** returns proc's true result instead of `(cons key value)`, per SRFI-125 spec
- **21 missing exports** added: 10 re-exported from SRFI-69 (`hash-table-walk`, `hash-table-exists?`, `hash`, `string-hash`, etc.) and 11 new Scheme implementations (`hash-table-unfold`, `hash-table=?`, `hash-table-map`, `hash-table-map!`, `hash-table-pop!`, `hash-table-clear!`, `hash-table-prune!`, `hash-table-empty-copy`, `hash-table-merge!`, `hash-table-xor!`, `hash-table-mutable?`)

## Test plan

- [x] 86 SRFI-125 conformance tests pass (previously 2 were disabled as FAIL markers)
- [x] Full test suite: 1806 pass, 0 fail
- [x] Zig unit tests pass

### DIFF
--- a/lib/srfi/125.sld
+++ b/lib/srfi/125.sld
@@ -1,8 +1,7 @@
 (define-library (srfi 125)
   (import (scheme base)
           (rename (srfi 69)
-                  (hash-table-ref srfi69-ref)
-                  (hash-table-walk srfi69-walk))
+                  (hash-table-ref srfi69-ref))
           (srfi 128))
   (export make-hash-table hash-table hash-table?
           hash-table-contains? hash-table-empty? hash-table-size
@@ -14,7 +13,15 @@
           hash-table-for-each hash-table-fold
           hash-table-count hash-table-find
           hash-table-map->list
-          hash-table-union! hash-table-intersection! hash-table-difference!)
+          hash-table-union! hash-table-intersection! hash-table-difference!
+          ;; new exports
+          hash-table-unfold alist->hash-table
+          hash-table-exists? hash-table=? hash-table-mutable?
+          hash-table-update!/default hash-table-pop! hash-table-clear!
+          hash-table-map hash-table-walk hash-table-map! hash-table-prune!
+          hash-table-empty-copy hash-table-merge! hash-table-xor!
+          hash string-hash string-ci-hash hash-by-identity
+          hash-table-equivalence-function hash-table-hash-function)
   (begin
     (define (hash-table comparator . args)
       (let ((ht (make-hash-table)))
@@ -33,7 +40,10 @@
 
     (define (hash-table-ref ht key . args)
       (if (hash-table-exists? ht key)
-          (srfi69-ref ht key)
+          (let ((v (srfi69-ref ht key)))
+            (if (and (not (null? args)) (not (null? (cdr args))))
+                ((cadr args) v)
+                v))
           (if (null? args)
               (error "hash-table-ref: key not found" key)
               ((car args)))))
@@ -44,11 +54,14 @@
           default))
 
     (define (hash-table-update! ht key updater . args)
-      (let ((old (if (hash-table-exists? ht key)
-                     (srfi69-ref ht key)
-                     (if (null? args)
-                         (error "hash-table-update!: key not found" key)
-                         ((car args))))))
+      (let* ((failure (if (null? args) #f (car args)))
+             (success (if (or (null? args) (null? (cdr args))) #f (cadr args)))
+             (old (if (hash-table-exists? ht key)
+                      (let ((v (srfi69-ref ht key)))
+                        (if success (success v) v))
+                      (if failure
+                          (failure)
+                          (error "hash-table-update!: key not found" key)))))
         (hash-table-set! ht key (updater old))))
 
     (define (hash-table-intern! ht key thunk)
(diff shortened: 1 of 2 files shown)


lib/srfi/125.sld
```diff
@@ -106,8 +120,82 @@
...
+            (let* ((k (car keys))
+                   (v (srfi69-ref ht k)))
+              (hash-table-delete! ht k)
+              (values k v)))))
+
+    (define (hash-table-clear! ht)
+      (for-each (lambda (k) (hash-table-delete! ht k))
+                (hash-table-keys ht)))
+
+    (define (hash-table-map proc comparator ht)
```
baijum, line 173: Non-blocking: the `comparator` argument is accepted but ignored — the result is always an `equal?`-keyed table from `(make-hash-table)`. Same applies to `hash-table-unfold` (line 134) and `hash-table-empty-copy` (line 192, which per spec should preserve the source's comparator/weakness/mutability). This is harmless today because of #1183 (the Zig primitives ignore comparators anyway), but these three will need revisiting when comparators are honored. A short `;; comparator ignored — see #1183` comment here would flag it for that future work.

baijum (commented):
## Review — verified locally ✅

Built `main` and this branch, ran the reproductions from #1229, and confirmed the fix end-to-end:

| repro | `main` (buggy) | this PR |
|---|---|---|
| `(hash-table-ref ht 'x (lambda () 'no) (lambda (v) (* v 100)))` | `12` | `1200` ✅ |
| `(hash-table-find (lambda (k v) (and (= v 12) (* v 10))) ht …)` | `(x . 12)` | `120` ✅ |

- `zig build` clean; `tests/scheme/srfi/srfi125.scm` → **86 pass, 0 fail** (was 2 FAIL markers).
- Related suites still green: srfi113 (133), srfi128 (59), srfi69 (26).
- Blast radius is limited — the only importers of `(srfi 125)` are the test and a coverage loader, so nothing else is affected.
- All 21 new exports resolve: the 10 re-exports are all registered on `srfi_69` in `src/primitives_hashtable.zig`, and the 11 new definitions match their SRFI-125 signatures (`unfold`/`pop!` return the right value shapes, `find`/`ref`/`update!` success-proc semantics match the spec).

**The three fixes are correct and match the SRFI-125 spec. No blocking issues.** A few non-blocking notes below, plus the general caveat that `make-hash-table` still ignores its comparator (#1183) — so all these procedures are effectively `equal?`-keyed regardless of the comparator passed. That's a pre-existing limitation, not something this PR needs to solve.

**Ready to merge.**

lib/srfi/125.sld
```diff
@@ -106,8 +120,82 @@
...
-      (srfi69-walk ht2
+      (hash-table-walk ht2
         (lambda (k v)
           (if (hash-table-exists? ht1 k)
               (hash-table-delete! ht1 k))))
+      ht1)
+
+    ;; SRFI-125 merge! is an alias for union! (ht1 values win on collision),
+    ;; NOT the SRFI-69 merge! (where ht2 wins). This define shadows the import.
+    (define (hash-table-merge! ht1 ht2)
```
baijum, line 131: Nice — the comment explaining why this shadows the SRFI-69 `merge!` (ht1 wins vs ht2 wins) is exactly the kind of thing that saves a future reader. Note this relies on a library-body `define` shadowing an imported binding, which is the same pattern already used for `hash-table-fold`, `hash-table-ref/default`, and `hash-table-update!` in this file, so it's consistent and the tests confirm Kaappi accepts it.

PR #1338, opened 2026-07-08
Complete SRFI-235 implementation: add 27 missing exports, fix return values (#1221)

### LINKED ISSUE #1221: [SRFI-235] 24 of 36 exports missing; all-of/conjoin return #t instead of the last predicate result
**Category:** SRFI
**Spec reference:** SRFI-235 exports: constantly, complement, flip, swap, on-left, on-right, conjoin, disjoin, each-of, all-of, any-of, on, left-section, right-section, apply-chain, arguments-drop, arguments-drop-right, arguments-take, arguments-take-right, group-by, begin-procedure, if-procedure, when-procedure, unless-procedure, value-procedure, case-procedure, and-procedure, eager-and-procedure, or-procedure, eager-or-procedure, funcall-procedure, loop-procedure, while-procedure, until-procedure, always, never, boolean. Semantics: "all-of: ... returns #f immediately if unsatisfied, otherwise returns last call result or #t if empty"; "conjoin: ... returns #f if any returns false, otherwise returns last value or #t if none".
**Severity:** missing-feature (plus two return-value deviations)

**Missing exports** (present: constantly, complement, swap, flip, on, each-of, all-of, any-of, conjoin, disjoin, plus non-SRFI compose/o): on-left, on-right, left-section, right-section, apply-chain, arguments-drop, arguments-drop-right, arguments-take, arguments-take-right, group-by, begin-procedure, if-procedure, when-procedure, unless-procedure, value-procedure, case-procedure, and-procedure, eager-and-procedure, or-procedure, eager-or-procedure, funcall-procedure, loop-procedure, while-procedure, until-procedure, always, never, boolean.

**Return-value deviations:**
```scheme
(import (scheme base) (srfi 235))
((all-of (lambda (x) x)) '(1 2))   ;; expected: 2 (last call result)   actual: #t
((conjoin number? odd?) 3)         ;; expected: #t per truthiness but e.g.
((conjoin (lambda (x) x)) 5)       ;; expected: 5 (last value)         actual: #t
```

**Source location:** lib/srfi/235.sld
**Disabled test:** tests/scheme/srfi/srfi235.scm (marked ;; FAIL: #this)

baijum:
## Summary
- Fix `all-of` and `conjoin` to return the last predicate result instead of `#t`
- Add all 27 missing SRFI-235 exports: `on-left`, `on-right`, `left-section`, `right-section`, `apply-chain`, `arguments-drop`/`take`/`drop-right`/`take-right`, `group-by`, `begin`/`if`/`when`/`unless`/`value`/`case-procedure`, `and`/`eager-and`/`or`/`eager-or-procedure`, `funcall`/`loop`/`while`/`until-procedure`, `always`, `never`, `boolean`
- 90 test cases covering all procedures including regression tests for the return-value fixes

## Test plan
- [x] `zig-out/bin/kaappi tests/scheme/srfi/srfi235.scm` — 90/90 pass
- [x] `bash tests/scheme/run-all.sh` — 1806 pass, 0 fail

Closes #1221

### DIFF
--- a/lib/srfi/235.sld
+++ b/lib/srfi/235.sld
@@ -162,10 +162,10 @@
                   (if v (loop (cdr ts)) #f))))))
 
     (define (eager-and-procedure . thunks)
-      (let ((results (map (lambda (t) (t)) thunks)))
-        (let loop ((rs results) (last #t))
-          (if (null? rs) last
-              (if (car rs) (loop (cdr rs) (car rs)) #f)))))
+      (let loop ((ts thunks) (last #t))
+        (if (null? ts) last
+            (let ((v ((car ts))))
+              (loop (cdr ts) (if (and last v) v #f))))))
 
     (define (or-procedure . thunks)
       (let loop ((ts thunks))
@@ -174,10 +174,10 @@
               (if v v (loop (cdr ts)))))))
 
     (define (eager-or-procedure . thunks)
-      (let ((results (map (lambda (t) (t)) thunks)))
-        (let loop ((rs results))
-          (if (null? rs) #f
-              (if (car rs) (car rs) (loop (cdr rs)))))))
+      (let loop ((ts thunks) (first #f))
+        (if (null? ts) first
+            (let ((v ((car ts))))
+              (loop (cdr ts) (or first v))))))
 
     (define (funcall-procedure thunk)
       (thunk))
--- a/lib/srfi/235.sld
+++ b/lib/srfi/235.sld
@@ -1,10 +1,20 @@
 (define-library (srfi 235)
   (import (scheme base) (scheme case-lambda))
-  (export constantly complement swap flip
-          on each-of all-of any-of
-          conjoin disjoin
+  (export constantly complement swap flip on-left on-right
+          on each-of all-of any-of conjoin disjoin
+          left-section right-section apply-chain
+          arguments-drop arguments-drop-right
+          arguments-take arguments-take-right
+          group-by
+          begin-procedure if-procedure when-procedure unless-procedure
+          value-procedure case-procedure
+          and-procedure eager-and-procedure
+          or-procedure eager-or-procedure
+          funcall-procedure loop-procedure while-procedure until-procedure
+          always never boolean
           compose o)
   (begin
+
     (define (constantly . vals)
       (lambda args (apply values vals)))
 
@@ -17,6 +27,12 @@
     (define (flip f)
       (lambda args (apply f (reverse args))))
 
+    (define (on-left f)
+      (lambda (a b) (f a)))
+
+    (define (on-right f)
+      (lambda (a b) (f b)))
+
     (define (on f g)
       (lambda args (apply f (map g args))))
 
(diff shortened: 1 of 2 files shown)


lib/srfi/235.sld
```diff
@@ -25,26 +41,168 @@
...
+
+    (define (and-procedure . thunks)
+      (if (null? thunks) #t
+          (let loop ((ts thunks))
+            (if (null? (cdr ts))
+                ((car ts))
+                (let ((v ((car ts))))
+                  (if v (loop (cdr ts)) #f))))))
+
+    (define (eager-and-procedure . thunks)
```
baijum, line 164: `eager-and-procedure` (and `eager-or-procedure` just below) invoke the thunks via `(map (lambda (t) (t)) thunks)`. R7RS does **not** guarantee `map`'s evaluation order, but SRFI-235 specifies *"All the thunks are invoked from left to right."* It happens to work on Kaappi (map is left-to-right here), but since these `.sld` files are meant to be portable R7RS, an explicit left-to-right loop would be strictly conformant, e.g.:

```scheme
(define (eager-and-procedure . thunks)
  (let loop ((ts thunks) (last #t))
    (if (null? ts) last
        (let ((v ((car ts))))
          (loop (cdr ts) (if last v #f))))))
```

Non-blocking.

baijum (commented):
## Review — PR #1338: Complete SRFI-235

**Verdict: ready to merge.** Nice, thorough completion of SRFI-235. I checked out the branch, built it, and verified behavior against the [SRFI-235 spec](https://srfi.schemers.org/srfi-235/srfi-235.html).

### What I verified
- **Build + tests**: `zig build` clean; `srfi235.scm` → 90/90 pass (exit 0).
- **Return-value fixes** (the core of #1221): `all-of` and `conjoin` now return the last predicate result (and `#t` on empty), matching the spec. Regression tests cover both (`all-of: returns last result` → 2, `conjoin: returns last value` → 5).
- **All 37 SRFI-235 exports present** — cross-checked the export list against the issue's full list; nothing missing. `compose`/`o` correctly kept as non-SRFI extras.
- **Spec conformance, spot-checked against the SRFI text**:
  - `right-section` — objs ++ `(reverse args)` ✓
  - `apply-chain` — right-to-left multi-value composition ✓
  - `group-by` — groups and elements preserve list order; defaults to `equal?`, honors custom `=` ✓
  - `case-procedure` — uses `assv`, unspecified when no match + no else ✓
  - `begin-/if-/when-/unless-/value-procedure`, `and-/or-procedure`, `always/never/boolean`, `arguments-drop/take/-right` — all match ✓
  - `while-/until-procedure` return unspecified per spec ✓

### One non-blocking nit
See inline comment on `eager-and-procedure` / `eager-or-procedure`: they rely on `map` evaluating left-to-right, which R7RS leaves unspecified. Works correctly on Kaappi (I confirmed `map` is left-to-right here), but these `lib/srfi/*.sld` files are documented as *portable R7RS*, and the spec explicitly requires left-to-right thunk invocation. Worth tightening for portability, but not a blocker.

baijum (commented):
## Follow-up — nit addressed, ready to merge ✅

Commit `b51bf597` resolves the one nit from my previous review.

**`eager-and-procedure` / `eager-or-procedure`** now use explicit left-to-right loops instead of `map`, so evaluation order no longer depends on R7RS-unspecified `map` behavior — matching the spec's *"invoked from left to right"* requirement. I re-checked the semantics:
- `eager-and-procedure`: all thunks always invoked (eager); returns the last value if all true, `#f` if any false, `#t` if empty. ✓
- `eager-or-procedure`: all thunks always invoked (eager); returns the first true value, `#f` if none (or empty). ✓

Two order-sensitive regression tests were added that would fail under out-of-order evaluation.

**Verified on the updated branch:** `zig build` clean; `srfi235.scm` → **92/92 pass** (exit 0).

No remaining concerns — this is good to merge.

PR #1339, opened 2026-07-08
Implement complete SRFI-132 sort library (22 procedures)

### LINKED ISSUE #1231: [SRFI-132] 14 of 22 spec procedures missing (merge, delete-neighbor-dups, find-median, select!, separate!, stable-sort! variants); vector procedures lack start/end
**Category:** SRFI (lib/srfi/132.sld)
**Spec reference:** SRFI-132 exports 22 procedures: list-sorted?, vector-sorted?, list-sort, list-stable-sort, list-sort!, list-stable-sort!, vector-sort, vector-stable-sort, vector-sort!, vector-stable-sort!, list-merge, list-merge!, vector-merge, vector-merge!, list-delete-neighbor-dups, list-delete-neighbor-dups!, vector-delete-neighbor-dups, vector-delete-neighbor-dups!, vector-find-median, vector-find-median!, vector-select!, vector-separate!. Vector signatures take ranges: "(vector-sort < v [ start [ end ] ])", "(vector-sorted? < v [ start [ end ] ])".
**Severity:** missing API

**Missing exports (14):** list-stable-sort!, vector-stable-sort!, list-merge, list-merge!, vector-merge, vector-merge!, list-delete-neighbor-dups, list-delete-neighbor-dups!, vector-delete-neighbor-dups, vector-delete-neighbor-dups!, vector-find-median, vector-find-median!, vector-select!, vector-separate!

**Optional range arguments missing:**
```scheme
(import (scheme base) (srfi 132))
(vector-sort < #(9 4 3 2 9) 1 4)   ;; expected #(2 3 4)  actual arity error
(vector-sorted? < #(9 1 2 3 0) 1 4) ;; expected #t        actual arity error
```

**Conforms:** the 8 present procedures behave correctly on full ranges, including stability of list-stable-sort/vector-stable-sort (merge takes from the first list on ties) and non-destructiveness of the plain variants.

**Root cause:** lib/srfi/132.sld implements only a minimal merge-sort core.
**Disabled tests:** tests/scheme/srfi/srfi132.scm (marked ;; FAIL: #this)

baijum:
## Summary

Closes #1231

- Implement all 14 missing SRFI-132 procedures: `list-stable-sort!`, `vector-stable-sort!`, `list-merge`, `list-merge!`, `vector-merge`, `vector-merge!`, `list-delete-neighbor-dups`, `list-delete-neighbor-dups!`, `vector-delete-neighbor-dups`, `vector-delete-neighbor-dups!`, `vector-find-median`, `vector-find-median!`, `vector-select!`, `vector-separate!`
- Retrofit existing vector procedures (`vector-sorted?`, `vector-sort`, `vector-stable-sort`, `vector-sort!`) with optional `start`/`end` range parameters via `case-lambda`
- Expand test suite from ~20 to 73 tests covering all 22 procedures including subranges, stability, edge cases, and `vector-delete-neighbor-dups!` integer return type

## Test plan

- [x] All 73 SRFI-132 tests pass
- [x] Full test suite passes (1806 pass, 0 fail)
- [x] Zig unit tests pass

### DIFF
--- a/lib/srfi/132.sld
+++ b/lib/srfi/132.sld
@@ -192,41 +192,41 @@
 
     (define vector-delete-neighbor-dups
       (case-lambda
-        ((= vec)
-         (%vector-delete-neighbor-dups = vec 0 (vector-length vec)))
-        ((= vec start)
-         (%vector-delete-neighbor-dups = vec start (vector-length vec)))
-        ((= vec start end)
-         (%vector-delete-neighbor-dups = vec start end))))
+        ((elt= vec)
+         (%vector-delete-neighbor-dups elt= vec 0 (vector-length vec)))
+        ((elt= vec start)
+         (%vector-delete-neighbor-dups elt= vec start (vector-length vec)))
+        ((elt= vec start end)
+         (%vector-delete-neighbor-dups elt= vec start end))))
 
-    (define (%vector-delete-neighbor-dups = vec start end)
+    (define (%vector-delete-neighbor-dups elt= vec start end)
       (if (>= start end)
           #()
           (let loop ((i (+ start 1))
                      (acc (list (vector-ref vec start))))
             (cond
              ((= i end) (list->vector (reverse acc)))
-             ((= (car acc) (vector-ref vec i))
+             ((elt= (car acc) (vector-ref vec i))
               (loop (+ i 1) acc))
              (else
               (loop (+ i 1) (cons (vector-ref vec i) acc)))))))
 
     (define vector-delete-neighbor-dups!
       (case-lambda
-        ((= vec)
-         (%vector-delete-neighbor-dups! = vec 0 (vector-length vec)))
-        ((= vec start)
-         (%vector-delete-neighbor-dups! = vec start (vector-length vec)))
-        ((= vec start end)
-         (%vector-delete-neighbor-dups! = vec start end))))
+        ((elt= vec)
+         (%vector-delete-neighbor-dups! elt= vec 0 (vector-length vec)))
+        ((elt= vec start)
+         (%vector-delete-neighbor-dups! elt= vec start (vector-length vec)))
+        ((elt= vec start end)
+         (%vector-delete-neighbor-dups! elt= vec start end))))
 
-    (define (%vector-delete-neighbor-dups! = vec start end)
+    (define (%vector-delete-neighbor-dups! elt= vec start end)
       (if (>= start end)
           start
           (let loop ((i (+ start 1)) (j (+ start 1)))
             (cond
              ((= i end) j)
-             ((= (vector-ref vec (- j 1)) (vector-ref vec i))
+             ((elt= (vector-ref vec (- j 1)) (vector-ref vec i))
               (loop (+ i 1) j))
              (else
               (vector-set! vec j (vector-ref vec i))
--- a/lib/srfi/132.sld
+++ b/lib/srfi/132.sld
@@ -1,8 +1,18 @@
 (define-library (srfi 132)
-  (import (scheme base))
-  (export list-sorted? list-sort list-stable-sort list-sort!
-          vector-sorted? vector-sort vector-stable-sort vector-sort!)
+  (import (scheme base) (scheme case-lambda))
+  (export list-sorted? list-sort list-stable-sort list-sort! list-stable-sort!
+          list-merge list-merge!
+          vector-sorted? vector-sort vector-stable-sort vector-sort!
+          vector-stable-sort!
+          vector-merge vector-merge!
+          list-delete-neighbor-dups list-delete-neighbor-dups!
+          vector-delete-neighbor-dups vector-delete-neighbor-dups!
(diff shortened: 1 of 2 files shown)


baijum (commented):
## Review — SRFI-132 sort library

Reviewed at `49a1a13`. Nice, thorough piece of work: all 22 SRFI-132 procedures are exported, the `case-lambda` start/end retrofits are clean, stability is preserved (list merge-sort + first-wins ties), and the suite grew from ~20 to 73 tests. I built the branch, ran the file (73 passing), and hand-checked the arg forms the suite doesn't exercise — `vector-merge!` full 9-arg form, `vector-stable-sort!`/`vector-separate!`/`vector-select!` with explicit `start`/`end` — all behave per spec. CI is green on every platform.

There is **one blocking correctness bug**, though (also caught by CodeRabbit):

### 🐛 `vector-delete-neighbor-dups[!]` crash with a non-numeric equality predicate

`lib/srfi/132.sld:208` and `lib/srfi/132.sld:228` reuse the element-equality parameter `=` for the **integer index-termination check** `(= i end)`:

```scheme
(define (%vector-delete-neighbor-dups = vec start end)
  ...
    ((= i end) (list->vector (reverse acc)))          ; line 208 — = is the caller's predicate!
    ((= (car acc) (vector-ref vec i)) ...)))           ; correct: element comparison
```

SRFI-132 documents these as taking an *arbitrary* equality predicate (the spec just happens to name the parameter `=`). Every test passes only because they all supply the numeric `=`. Any other predicate blows up on the index check:

```
> (vector-delete-neighbor-dups char=? #(#\a #\a #\b))
error: type error in 'char=?': expected character, got 1
```

Both the non-destructive helper (`%vector-delete-neighbor-dups`, line 202) and the destructive one (`%vector-delete-neighbor-dups!`, line 223) are affected. The **list** variants are fine — `list-delete-neighbor-dups` terminates on `null?`, so `(list-delete-neighbor-dups char=? '(#\a #\a #\b))` correctly returns `(a b)`.

**Fix:** rename the comparator parameter (e.g. `elt=`) in both vector helpers and let `(= i end)` resolve to the built-in numeric `=`; use `elt=` only for element comparisons. Per the repo's testing rule, please add a regression test with a non-numeric predicate (e.g. `char=?` or `string=?`) covering both variants.

### Minor (non-blocking)

- **Test over-specifies `vector-separate!` order** (`tests/scheme/srfi/srfi132.scm:189-194`): the suite asserts index 0 = 1 *and* index 1 = 2. SRFI-132 only guarantees the smallest `k` elements land in the first `k` positions, not their relative order. It passes today because the impl fully sorts, but it couples the test to that implementation detail. Asserting membership (`{1,2}`) would be more faithful to the spec.
- **Selection procedures use a full O(n log n) sort:** `vector-select!`, `vector-separate!`, and `vector-find-median[!]` sort the whole range rather than partition. Correct and fine for a portable reference lib — noting only that these procedures exist in the SRFI specifically to enable sub-sort-cost selection, so it's a reasonable future optimization.
- **`!` variants are non-destructive** (`list-stable-sort!`, `list-merge!`, `list-delete-neighbor-dups!` alias the pure versions; `list-sort!` itself doesn't mutate). This is spec-conformant — the `!` forms *may* but need not reuse storage — just flagging in case any caller expects in-place mutation of the input list.
- **Coverage gap** (optional): the 6–9-arg forms of `vector-merge!` and the explicit `start`/`end` forms of `vector-stable-sort!`/`vector-separate!`/`vector-select!` aren't tested. I verified them manually; a couple of tests would lock them in.

### Verdict

Not quite ready to merge — the `delete-neighbor-dups` shadowing bug is a real crash for non-numeric predicates and should be fixed (plus a regression test) first. Everything else is solid and the minor items are at your discretion. Once the fix lands, this is good to go.

baijum (commented):
## Re-review — fix verified ✅

The new commit `bc526ba` addresses everything from the previous round:

- **Shadowing bug fixed:** the comparator parameter is renamed `=` → `elt=` in both `%vector-delete-neighbor-dups` and `%vector-delete-neighbor-dups!`, so the index check `(= i end)` now resolves to numeric `=` while element comparisons use `elt=`. I rebuilt and confirmed the cases that used to crash now work:
  - `(vector-delete-neighbor-dups char=? #(#\a #\a #\b #\b #\c))` → `#(a b c)`
  - `(vector-delete-neighbor-dups! char=? (vector #\a #\a #\b))` → `2`
- **Regression tests added** for both variants with `char=?`.
- **`vector-separate!` test relaxed** to assert membership in the first two positions rather than a specific order — now faithful to the SRFI-132 guarantee.

Suite is 75/75 passing locally and CI is green on every platform (macOS, Ubuntu x86/ARM, riscv64, wasm, Debug/ReleaseSafe/ReleaseFast).

The remaining items I mentioned (O(n log n) selection procedures, non-destructive `!` variants) are conformant and non-blocking. **This is ready to merge.** 🚀

PR #1340, opened 2026-07-08
Fix opt*-lambda sequential defaults and lift 2-optional cap

### LINKED ISSUE #1222: [SRFI-227] opt*-lambda is an alias of opt-lambda — sequential defaults cannot reference earlier parameters in the zero-argument case
**Category:** SRFI
**Spec reference:** SRFI-227: in opt*-lambda, "the ⟨default⟩s are evaluated sequentially ... each default expression is evaluated in an environment where the bindings of all previous formals are visible" (let*-style), whereas opt-lambda's defaults do not see the other formals.
**Severity:** wrong-result

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 227))
(define s (opt*-lambda ((a 2) (b (* a 3))) (list a b)))
(s 5)   ;; (5 15) — works (a is a case-lambda parameter in scope)
(s)     ;; expected: (2 6)   actual: error — expands to (let ((a 2) (b (* a 3))) ...)
        ;; where (* a 3) cannot see a (parallel let)
```

**Root cause:** lib/srfi/227.sld defines `opt*-lambda` as a direct alias of `opt-lambda`, and opt-lambda's zero-argument branch binds all defaults with a parallel `let`. The alias also means opt-lambda inherits opt*-semantics in the partially-applied cases (later defaults CAN see earlier parameters), so the two forms are swapped/merged rather than distinct. Additionally the hand-enumerated patterns cap at 2 optionals (a third optional falls through to the plain-lambda case and changes arity behavior).

**Source location:** lib/srfi/227.sld
**Disabled test:** tests/scheme/srfi/srfi227.scm (marked ;; FAIL: #this)

baijum:
## Summary

- **`opt*-lambda` was a bare alias of `opt-lambda`** — its zero-argument case used parallel `let`, so defaults could not reference earlier parameters (e.g. `(opt*-lambda ((a 2) (b (* a 3))) ...)` errored when called with no args instead of returning `(2 6)`)
- **Both macros used hand-enumerated patterns capped at 2 optionals** — a third optional fell through to plain `lambda`, silently changing arity behavior
- Replaced both with recursive `syntax-rules` using literal-tag dispatch (`%p`/`%b`/`%d`), supporting arbitrary numbers of required and optional parameters
- Worked around a Kaappi expander limitation (can't splice `...` into `let` binding lists) by generating nested single-binding `let` forms instead

## Test plan

- [x] SRFI-227 tests expanded from 18 to 41 — covers `opt*-lambda` zero-arg sequential defaults, 3-4 optionals, required+optional combos, `let-optionals*`
- [x] Full Scheme test suite: 1,806 pass, 0 fail
- [x] Zig unit tests pass

Closes #1222


baijum:
## Review

Thorough fix for the reported bug, with a genuinely better structure. I built it (ReleaseSafe) and ran it against the SRFI-227 spec text and the reference implementation. **`opt*-lambda` is fixed correctly** and the recursive `%p`/`%b`/`%d` rewrite is a real improvement over the hand-enumerated patterns. But the same rewrite introduces a **semantic regression in `opt-lambda`** that I think should be addressed (or at least consciously documented) before merge.

### ✅ What's right
- `opt*-lambda` zero-arg sequential defaults now work: `(opt*-lambda ((a 2) (b (* a 3))) ...)` called with no args returns `(2 6)`. Verified.
- The 2-optional cap is genuinely lifted — 3/4-optional cases all pass.
- Test expansion (18→41) is excellent, and the full suite passes for the cases it covers.

### ⚠️ Blocking concern: `opt-lambda` and `opt*-lambda` are now identical

The two `%d` rule sets are byte-for-byte identical (modulo the macro name) — both emit **nested** `let`s:

```scheme
((opt-lambda %d ((n d) rest ...) body ...)
 (let ((n d)) (opt-lambda %d (rest ...) body ...)))   ; ← sequential (let*)
```

So `opt-lambda` now evaluates its defaults sequentially, letting a later default see an earlier **formal**. That contradicts SRFI-227, which distinguishes the two forms exactly like `let` vs `let*`:

> **opt-lambda:** "The corresponding `init`s are evaluated in an **unspecified order** in the **lexical environment of the `opt-lambda` expression**."
> **opt\*-lambda:** "the `init`s … are evaluated **sequentially from left to right**, and the region of the binding of a `variable` is that part of the `opt*-lambda` expression to the right of it."

`opt-lambda` defaults are supposed to be evaluated in the *outer* environment (they must not depend on the formals); only `opt*-lambda` is sequential. The [reference implementation](https://github.com/scheme-requests-for-implementation/srfi-227/blob/master/lib/srfi/lambda-optional.scm) enforces this deliberately: `opt-lambda` substitutes defaults into a recursive `case-lambda` call using **gensyms** for the parameters, so a default can't see any formal by name; `opt*-lambda` binds the **real** names, so each default sees earlier ones.

**This is also a regression from `main`.** `main`'s `opt-lambda` zero-arg branch used a *parallel* `let`, which was spec-correct. Observable difference:

```scheme
(define a 100)
((opt-lambda ((a 1) (b a)) (list a b)))
;; main:            (1 100)   ← b's default sees the outer a  (correct)
;; this PR:         (1 1)     ← b's default sees the formal a  (opt*-lambda semantics)
;; spec/reference:  (1 100)
```

I verified all three empirically. The net effect is that `opt-lambda` and `opt*-lambda` (and by extension `let-optionals` / `let-optionals*`, since they delegate) are no longer distinguishable — which defeats the reason the SRFI defines both.

The commit message anticipates this ("opt-lambda generates nested let (effectively sequential… consistent with partial-application cases where earlier params are already case-lambda-bound)"), but that rationale doesn't hold against the reference: it gensyms the required/earlier params out *specifically* so later defaults can't reference them, even in the partially-applied clauses.

**Options:**
1. Give `opt-lambda` distinct semantics — emit a single **parallel** `let` for the defaulted tail (evaluated in the outer scope), keeping only `opt*-lambda` on the nested-`let` path. (Matching the reference's gensym approach is the fully-correct version, but a parallel `let` already fixes the observable regression.)
2. If the intent is to deliberately make `opt-lambda` sequential, say so explicitly and add a test that pins the behavior — but note that still diverges from the SRFI.

Either way, please add a distinguishing test, e.g. the `(1 100)` case above, since none of the 41 new tests exercise an `opt-lambda`/`let-optionals` default that references an earlier formal — the deviation is currently invisible to the suite.

### Minor / non-blocking
- **Literal-tag dispatch.** Using `%p`/`%b`/`%d` as `syntax-rules` literals means a user parameter literally named `%p`/`%b`/`%d` would misfire. Extremely unlikely, and hygiene mostly protects it, but the reference sidesteps it entirely by using separate helper macros (`%lambda-optional-parse`, …). Worth a one-line comment noting these are internal dispatch tags.
- **Rest args (pre-existing, not this PR).** SRFI-227 allows a rest arg after optionals, e.g. `(opt-lambda (a (b 2) . rest) …)`. Both `main` and this PR reject it with `InvalidSyntax`, so it's not a regression — just noting the recursive rewrite would have been a natural place to close that gap.

### Verdict
Not quite ready to merge as-is: `opt*-lambda` is fixed, but `opt-lambda` picks up `opt*-lambda`'s semantics, which is a spec deviation and a regression from `main`. Resolving the `opt-lambda` default scoping (option 1) plus a distinguishing test would make this a clean, mergeable fix.

*Reviewed against SRFI-227 spec text and reference implementation; behaviors verified on a fresh ReleaseSafe build.*

baijum:
## Re-review — ✅ addressed, ready to merge

Commit `9eb82ff` resolves the blocking concern. `opt-lambda`'s `%d` phase now accumulates names/defaults and emits `((lambda (names ...) body ...) defaults ...)`, an immediate application that evaluates every default in the enclosing scope before binding — i.e. parallel (`let`) semantics — while `opt*-lambda` keeps the nested-`let` (sequential) path. The two forms are distinct again, matching SRFI-227's `let`/`let*` distinction.

Verified on a fresh ReleaseSafe build:

| Case | Result | Expected |
|------|:---:|:---:|
| `opt*-lambda ((a 2)(b (* a 3)))` zero-arg (#1222) | `(2 6)` | `(2 6)` ✓ |
| `opt-lambda ((a 1)(b a))` w/ outer `a`=100 | `(1 100)` | `(1 100)` ✓ (parallel) |
| `opt*-lambda ((a 1)(b a))` w/ outer `a`=100 | `(1 1)` | `(1 1)` ✓ (sequential) |
| `let-optionals` vs `let-optionals*` (same shape) | `(1 100)` / `(1 1)` | ✓ delegation preserves both |
| `opt-lambda (r (a 10)(b 20)(c 30))` partial `(g4 1 2)` | `(1 2 20 30)` | ✓ |

- The distinguishing test (`outer-a`/`(1 99)`) is now in the suite, so the parallel-vs-sequential behavior is pinned. **42 pass, 0 fail.**
- The `;; %p/%b/%d are internal dispatch tags` comment was added — thanks.

The only remaining item is the pre-existing rest-args-after-optionals gap (`(opt-lambda (a (b 2) . rest) …)`), which is unchanged from `main` and out of scope here — fine to leave for a follow-up.

**LGTM — ready to merge.**

PR #1341, opened 2026-07-08
Fix SRFI-134 ideque-filter calling unbound 'filter' (#1212)

### LINKED ISSUE #1212: [SRFI-134] ideque-filter calls unbound 'filter' — (srfi 134) imports only (scheme base)
**Category:** SRFI
**Severity:** wrong-result (every ideque-filter call fails; the error escapes guard)

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 134))
(ideque->list (ideque-filter even? (ideque 1 2 3)))
;; expected: (2)
;; actual: error: undefined variable 'filter' (raised from inside the library; not catchable via guard)
```

**Root cause:** lib/srfi/134.sld line 87: `(define (ideque-filter pred dq) (list->ideque (filter pred (ideque->list dq))))` — but the library imports only `(scheme base)`, which has no `filter`. Fix: import `(srfi 1)` or define a local filter loop.

**Source location:** lib/srfi/134.sld:87
**Disabled test:** tests/scheme/srfi/srfi134.scm (marked ;; FAIL: #this)

baijum:
## Summary

- `ideque-filter` called `filter` which isn't in `(scheme base)` — every call failed with "undefined variable"
- Added a local `filter-list` helper (consistent with the existing local `fold-left` in the same file)
- Re-enabled the previously disabled test case

Closes #1212

## Test plan

- [x] `ideque-filter` returns `(2)` for `(ideque-filter even? (ideque 1 2 3))`
- [x] Full SRFI-134 test suite passes (28/28)

### DIFF
--- a/lib/srfi/134.sld
+++ b/lib/srfi/134.sld
@@ -83,8 +83,13 @@
       (if (null? lst) seed
           (fold-left f (f seed (car lst)) (cdr lst))))
 
+    (define (filter-list pred lst)
+      (cond ((null? lst) '())
+            ((pred (car lst)) (cons (car lst) (filter-list pred (cdr lst))))
+            (else (filter-list pred (cdr lst)))))
+
     (define (ideque-filter pred dq)
-      (list->ideque (filter pred (ideque->list dq))))
+      (list->ideque (filter-list pred (ideque->list dq))))
 
     (define (ideque-append . dqs)
       (list->ideque (apply append (map ideque->list dqs))))
--- a/tests/scheme/srfi/srfi134.scm
+++ b/tests/scheme/srfi/srfi134.scm
@@ -47,8 +47,7 @@
 ;;; --- higher-order operations ---
 (test '(2 4 6) (ideque->list (ideque-map (lambda (x) (* 2 x)) (ideque 1 2 3))))
 (test 6 (ideque-fold + 0 (ideque 1 2 3)))
-;; FAIL: #1212 (ideque-filter calls unbound 'filter' inside the library)
-;; (test '(2) (ideque->list (ideque-filter even? (ideque 1 2 3))))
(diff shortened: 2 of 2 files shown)


baijum:
## Review — verified locally ✅

Built kaappi at the PR head and ran the checks:

- **Root cause confirmed**: `filter` is registered as SRFI-1 only (`src/primitives_srfi1.zig:16`, `.libs = LS.initOne(.srfi_1)`), so it is unbound in a library that imports only `(scheme base)`. The bug in #1212 is real.
- **Repro passes**: `(ideque->list (ideque-filter even? (ideque 1 2 3)))` → `(2)`, and order is preserved (`(ideque-filter odd? (ideque 1 2 3 4 5))` → `(1 3 5)`).
- **Full SRFI-134 suite**: 28 pass, 0 fail.

The approach — a local `filter-list` mirroring the existing local `fold-left` — is the right call and keeps the fix self-contained (no new import of SRFI-1 into a base-only library).

### On CodeRabbit's tail-recursion note (non-blocking)

`filter-list` builds its result with a non-tail `cons`, so it uses O(n) stack. That's the standard, idiomatic shape for `filter` — a tail-recursive version needs an accumulator plus a final `reverse` to preserve element order, so the comparison to `fold-left` / `ideque-any` / `ideque-every` isn't quite like-for-like: those either fold or short-circuit and never accumulate an ordered result list, so tail position is free for them; for `filter` it isn't.

Given that this library is list-backed and already allocates O(n) freely per call (`ideque->list` does `append` + `reverse` every time), I don't consider the stack depth a blocker for this targeted fix. If deep-recursion robustness is ever a goal it'd be a library-wide follow-up, not something to hold this PR on.

### Verdict

Correct, minimal, well-tested, and consistent with the surrounding code. **Ready to merge** once the pending CI jobs go green (format + CodeRabbit already pass).

PR #1342, opened 2026-07-08
Fix SRFI-78 check-passed?, add check-set-mode! and check-ec

### LINKED ISSUE #1220: [SRFI-78] check-passed? has the wrong signature (returns a count, spec: (check-passed? expected-total-count) → boolean); check-ec and check-set-mode! missing
**Category:** SRFI
**Spec reference:** SRFI-78: "(check-passed? expected-total-count) PROCEDURE: #t if there were no failed checks and expected-total-count correct checks, #f otherwise." The SRFI's exports are check, check-ec, check-report, check-set-mode!, check-reset!, check-passed?.
**Severity:** missing-feature / wrong-result (API shape)

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 78))
(check (+ 1 1) => 2)
(check-passed? 1)   ;; expected: #t   actual: arity error — Kaappi's check-passed? takes 0 args and returns the pass count
check-set-mode!     ;; error: undefined variable
check-ec            ;; error: undefined variable
```

**Analysis:** lib/srfi/78.sld defines `(define (check-passed?) %pass)` — a zero-argument count accessor — plus a non-SRFI `check-failed?`. Portable SRFI-78 suites end with `(check-passed? N)` and will not run. `check-set-mode!` and `check-ec` are absent (check-ec depends on SRFI-42, see #1216).

**Source location:** lib/srfi/78.sld
**Disabled test:** tests/scheme/srfi/srfi78.scm (marked ;; FAIL: #this)

baijum:
## Summary
- **`check-passed?`** now takes an `expected-total-count` argument and returns a boolean per the SRFI-78 spec, instead of returning the raw pass count with zero args
- **`check-set-mode!`** added with modes `off`, `summary`, `report-failed`, `report`
- **`check-ec`** added using SRFI-42 eager comprehensions — counts as a single check, stops on first mismatch

Closes #1220

## Test plan
- [x] 15 SRFI-64 tests in `tests/scheme/srfi/srfi78.scm` covering all three features
- [x] R7RS test suite passes (1391 tests, 0 failures)
- [x] Issue reproduction cases verified (check-passed? with arg, check-set-mode!, check-ec all resolve)

### DIFF
--- a/lib/srfi/78.sld
+++ b/lib/srfi/78.sld
@@ -1,5 +1,5 @@
 (define-library (srfi 78)
-  (import (scheme base) (scheme write) (srfi 42))
+  (import (scheme base) (scheme write) (scheme process-context) (srfi 42))
   (export check check-ec check-report check-set-mode!
           check-reset! check-passed?
           check-failed?)
@@ -21,6 +21,8 @@
     (define (check-failed?) %fail)
 
     (define (check-set-mode! mode)
+      (unless (memq mode '(off summary report-failed report))
+        (error "check-set-mode!: invalid mode" mode))
       (set! %mode mode))
 
     (define (%report-pass name actual)
@@ -71,31 +73,51 @@
 
     (define-syntax check-ec
       (syntax-rules (=>)
+        ((_ expr => expected)
+         (check expr => expected))
+        ((_ expr (=> equal) expected)
+         (check expr (=> equal) expected))
         ((_ qualifier expr => expected)
-         (check-ec-proc 'expr
-           (lambda ()
-             (let ((ok #t))
-               (do-ec qualifier
-                 (let ((actual expr)
-                       (exp expected))
-                   (when (not (equal? actual exp))
-                     (set! ok (list actual exp))))) ok))
+         (check-ec-run 'expr
+           (lambda (escape)
+             (do-ec qualifier
+               (let ((actual expr) (exp expected))
+                 (when (not (equal? actual exp))
+                   (escape (list actual exp))))))
            equal?))
         ((_ qualifier expr (=> equal) expected)
-         (check-ec-proc 'expr
-           (lambda ()
-             (let ((ok #t)
-                   (eq-fn equal))
+         (check-ec-run 'expr
+           (lambda (escape)
+             (let ((eq-fn equal))
                (do-ec qualifier
-                 (let ((actual expr)
-                       (exp expected))
+                 (let ((actual expr) (exp expected))
                    (when (not (eq-fn actual exp))
-                     (set! ok (list actual exp))))) ok))
+                     (escape (list actual exp)))))))
+           equal))
+        ((_ qualifier expr => expected (arg ...))
+         (check-ec-run 'expr
+           (lambda (escape)
+             (do-ec qualifier
+               (let ((actual expr) (exp expected))
+                 (when (not (equal? actual exp))
+                   (escape (list actual exp))))))
+           equal?))
+        ((_ qualifier expr (=> equal) expected (arg ...))
+         (check-ec-run 'expr
+           (lambda (escape)
+             (let ((eq-fn equal))
+               (do-ec qualifier
+                 (let ((actual expr) (exp expected))
+                   (when (not (eq-fn actual exp))
+                     (escape (list actual exp)))))))
            equal))))
 
-    (define (check-ec-proc name thunk equal)
+    (define (check-ec-run name body-fn equal)
       (unless (eq? %mode 'off)
-        (let ((result (thunk)))
+        (let ((result (call-with-current-continuation
+                        (lambda (escape)
+                          (body-fn escape)
+                          #t))))
           (if (eq? result #t)
               (begin
                 (set! %pass (+ %pass 1))
@@ -112,4 +134,12 @@
       (display %pass) (display " passed, ")
       (display %fail) (display " failed.")
       (newline)
+      (when %first-fail
+        (display "First failure: ")
+        (write (car %first-fail))
+        (display " => ")
+        (write (cadr %first-fail))
+        (display " ; expected: ")
+        (write (caddr %first-fail))
+        (newline))
       (when (> %fail 0) (exit 1)))))
--- a/lib/srfi/78.sld
+++ b/lib/srfi/78.sld
@@ -1,18 +1,44 @@
 (define-library (srfi 78)
-  (import (scheme base) (scheme write))
-  (export check check-report check-reset!
-          check-passed? check-failed?)
+  (import (scheme base) (scheme write) (srfi 42))
+  (export check check-ec check-report check-set-mode!
+          check-reset! check-passed?
+          check-failed?)
   (begin
 
     (define %pass 0)
     (define %fail 0)
+    (define %mode 'report)
+    (define %first-fail #f)
 
     (define (check-reset!)
       (set! %pass 0)
-      (set! %fail 0))
+      (set! %fail 0)
+      (set! %first-fail #f))
 
-    (define (check-passed?) %pass)
+    (define (check-passed? expected-count)
+      (and (= %fail 0) (= %pass expected-count)))
+
     (define (check-failed?) %fail)
+
+    (define (check-set-mode! mode)
+      (set! %mode mode))
+
+    (define (%report-pass name actual)
+      (when (eq? %mode 'report)
+        (display "(") (display %pass) (display ") ")
+        (write name) (display " => ")
+        (write actual) (display " ; correct")
+        (newline)))
+
+    (define (%report-fail name actual expected)
+      (when (or (eq? %mode 'report) (eq? %mode 'report-failed))
+        (display "(") (display (+ %pass %fail)) (display ") ")
+        (write name) (display " => ")
+        (write actual) (display " ; *** WRONG ***")
+        (display " expected: ") (write expected)
+        (newline))
+      (when (not %first-fail)
+        (set! %first-fail (list name actual expected))))
 
     (define-syntax check
       (syntax-rules (=>)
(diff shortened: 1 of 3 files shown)


lib/srfi/78.sld
```diff
@@ -22,38 +48,61 @@
...
+                (set! %fail (+ %fail 1))
+                (%report-fail name actual expected))))))
+
+    (define-syntax check-ec
+      (syntax-rules (=>)
+        ((_ qualifier expr => expected)
+         (check-ec-proc 'expr
+           (lambda ()
+             (let ((ok #t))
+               (do-ec qualifier
```
baijum: **check-ec doesn't stop on the first failure (spec violation + infinite-loop hazard).**

The SRFI-78 spec says *"The comprehension stops after the first failed check, if there is any"* and reports the **first** failing case. This `do-ec` loop runs to completion and overwrites `ok` on every mismatch, so:

- **Unbounded qualifiers infinite-loop.** `(check-ec (:integers n) (foo n) => (bar n))` is a legitimate SRFI-78 pattern that relies on early-stop to terminate at the first failure; here it never returns.
- **It reports the last failing case, not the first.** I confirmed `(check-ec (:range i 1 4) i => 0)` prints `=> 3` (should be `=> 1`), and `(check-ec (:range i 0 10) ... => -1)` executed all 10 iterations. This also contradicts the PR description ("stops on first mismatch").

A clean fix that gives both early-stop and first-failure reporting is to escape out of the comprehension:

```scheme
(lambda ()
    (lambda (return)
      (do-ec qualifier ...
        (let ((actual expr) (exp expected))
          (when (not (equal? actual exp))
            (return (list actual exp)))))
      #t)))
```

(`call-with-escape-continuation` works too and is cheaper.)

baijum (commented):
## Review: SRFI-78 check-passed?, check-set-mode!, check-ec

Thanks — the core of #1220 is solid. I verified the build and ran the full SRFI suite (77/77 pass, including the rewritten `srfi78.scm`). The three headline fixes are correct:

- ✅ **`check-passed?`** now matches the spec: `(and (= %fail 0) (= %pass expected-count))`.
- ✅ **`check-set-mode!`** — verbosity for all four modes (`off`/`summary`/`report-failed`/`report`) is handled correctly, including the implicit `summary` fall-through (no per-check output, checks still counted).
- ✅ **`check-ec` single-check counting** — the whole comprehension increments the counter by one, per spec.

However, **`check-ec` has three spec-conformance gaps** that I'd like addressed before merge (details inline). The most important is the missing early-stop, which is both a correctness deviation and an infinite-loop hazard on unbounded qualifiers. The other two are grammar-coverage gaps. `check-passed?` / `check-set-mode!` are ready as-is.

### Summary of findings
| # | Issue | Severity |
|---|-------|----------|
| 1 | `check-ec` does not stop on first failure (spec: "stops after the first failed check"); runs the whole comprehension → infinite-loops on unbounded qualifiers, and reports the **last** failing case, not the first | Medium-High |
| 2 | `check-ec` only accepts **exactly one** qualifier; spec grammar is `<qualifier>*` — zero and multiple qualifiers both raise `InvalidSyntax` | Medium |
| 3 | `check-ec` `(argument ...)` reporting clause unsupported (raises `InvalidSyntax`) | Low |
| 4 | `%first-fail` is written but never read — dead state | Low |
| 5 | `check-set-mode!` doesn't reject unknown modes (spec/reference errors) | Low |

Verification notes: `(check-ec (:range i 0 10) ... => -1)` ran **10** iterations (expected 1 with early-stop) and reported `=> 9`; `(check-ec (* 2 2) => 4)`, `(check-ec q1 q2 ...)`, and `(check-ec ... => ... (i))` each fail to compile.

lib/srfi/78.sld
```diff
@@ -22,38 +48,61 @@
...
+              (begin
+                (set! %pass (+ %pass 1))
+                (%report-pass name actual))
+              (begin
+                (set! %fail (+ %fail 1))
+                (%report-fail name actual expected))))))
+
+    (define-syntax check-ec
+      (syntax-rules (=>)
+        ((_ qualifier expr => expected)
```
baijum, line 80: **Only exactly one qualifier is accepted.** The SRFI-78 grammar is `(check-ec <qualifier>* <expr> ...)` — zero or more. Both `(check-ec (* 2 2) => 4)` (zero) and `(check-ec (:range i 0 3) (:range j 0 3) (+ i j) => (+ i j))` (two) raise `compile error: error.InvalidSyntax`.

`do-ec` already accepts a variadic qualifier list, so switching the pattern to `qualifier ...` and splicing `(do-ec qualifier ... <body>)` covers zero/one/many in one shot.

lib/srfi/78.sld
```diff
@@ -22,38 +48,61 @@
...
+         (check-ec-proc 'expr
+           (lambda ()
+             (let ((ok #t))
+               (do-ec qualifier
+                 (let ((actual expr)
+                       (exp expected))
+                   (when (not (equal? actual exp))
+                     (set! ok (list actual exp))))) ok))
+           equal?))
+        ((_ qualifier expr (=> equal) expected)
```
baijum, line 88: **The `(argument ...)` reporting clause isn't supported.** The full grammar includes `(check-ec <qualifier>* <expr> => <expected> (<argument> ...))` (and the `(=> equal)` variant), where the arguments' values are printed on failure so you can see which binding failed. `(check-ec (:range i 0 3) i => i (i))` currently raises `InvalidSyntax`. Fine to defer to a follow-up, but worth a note/TODO since it's part of the advertised `check-ec`.

lib/srfi/78.sld
```diff
@@ -1,19 +1,45 @@
...
+
+    (define (%report-fail name actual expected)
+      (when (or (eq? %mode 'report) (eq? %mode 'report-failed))
+        (display "(") (display (+ %pass %fail)) (display ") ")
+        (write name) (display " => ")
+        (write actual) (display " ; *** WRONG ***")
+        (display " expected: ") (write expected)
+        (newline))
+      (when (not %first-fail)
+        (set! %first-fail (list name actual expected))))
```
baijum, line 43: **`%first-fail` is dead state** — it's set here and reset in `check-reset!`, but never read anywhere. In the reference SRFI-78, `check-report` prints the first failed check; here `check-report` ignores it. Either wire it into `check-report` (matching the spec's "prints a summary and the first failed check") or drop the variable and its two assignments.

lib/srfi/78.sld
```diff
@@ -1,19 +1,45 @@
...
+      (set! %fail 0)
+      (set! %first-fail #f))
+
+    (define (check-passed? expected-count)
+      (and (= %fail 0) (= %pass expected-count)))
 
-    (define (check-passed?) %pass)
     (define (check-failed?) %fail)
 
+    (define (check-set-mode! mode)
```
baijum, line 23: Minor: `check-set-mode!` accepts any symbol silently. The reference errors on an unknown mode; here a typo like `'reprot` silently degrades to `summary`-style behavior (no per-check output). Consider validating against `'(off summary report-failed report)` and raising otherwise.

lib/srfi/78.sld
```diff
@@ -63,4 +134,12 @@
       (display %pass) (display " passed, ")
       (display %fail) (display " failed.")
       (newline)
+      (when %first-fail
+        (display "First failure: ")
+        (write (car %first-fail))
+        (display " => ")
+        (write (cadr %first-fail))
+        (display " ; expected: ")
+        (write (caddr %first-fail))
```
baijum: 🔴 **Blocker: `caddr` is unbound here.** `(scheme base)` doesn't export `caddr` in Kaappi — it's in `(scheme cxr)`. Inside this `define-library`, only imported bindings are visible, so this line raises `undefined variable 'caddr'` whenever `%first-fail` is set. Repro: `(check (+ 1 1) => 3)` then `(check-report)`.

Three fixes, all verified working inside a base-only library:
- add `(scheme cxr)` to the imports on line 2, or
- `(list-ref %first-fail 2)`, or
- `(car (cddr %first-fail))` (`cddr` *is* in `(scheme base)`).

Note `car`/`cadr` on lines 139/141 resolve fine — it's only the 3-deep composition that's missing.

lib/srfi/78.sld
```diff
@@ -22,38 +50,81 @@
...
+                (set! %fail (+ %fail 1))
+                (%report-fail name actual expected))))))
+
+    (define-syntax check-ec
+      (syntax-rules (=>)
+        ((_ expr => expected)
+         (check expr => expected))
+        ((_ expr (=> equal) expected)
+         (check expr (=> equal) expected))
+        ((_ qualifier expr => expected)
```
baijum, line 80: Follow-up (non-blocking): multiple qualifiers still don't parse. The grammar is `<qualifier>*`, and nested comprehensions are common in SRFI-42, but `(check-ec (:range i 0 3) (:range j 0 3) (+ i j) => (+ i j))` raises `InvalidSyntax` — the patterns still capture a single `qualifier`. A `qualifier ...` ellipsis spliced into `(do-ec qualifier ... …)` would cover zero/one/many and let you drop the separate zero-qualifier clause. Fine to defer.

tests/scheme/srfi/srfi78.scm
```diff
@@ -1,40 +1,93 @@
...
+
+;; check-ec with trailing diagnostic arguments (accepted)
+(check-reset!)
+(check-ec (:range i 0 3) (* i i) => (* i i) (i))
+(test-assert "check-ec with arguments" (check-passed? 1))
+
+;; --- check-report exists and does not raise ---
+(check-reset!)
 (check (* 2 2) => 4)
 (check-report)
```
baijum, line 93: Coverage gap that hid the `caddr` blocker: this is the only `check-report` call, and it runs right after `check-reset!` + a passing check, so `%first-fail` is `#f` and the new first-failure branch never executes. Please add a case that calls `check-report` with a pending failure — e.g. under `(check-set-mode! 'summary)` (to suppress per-check noise) or `'off`-then-fail — and assert it doesn't raise. Guard against the `(exit 1)` at the end of `check-report` (it exits when `%fail > 0`).

baijum (commented):
## Re-review of `ae5fcbf` ("Address review")

Thanks for the fast turnaround. I re-checked out the branch, rebuilt, and re-ran every probe. Most of my findings are genuinely fixed — but **this commit introduces a new blocker**: the `check-report` first-failure path (the very feature added to wire up `%first-fail`) crashes, and it has no test covering it.

### Verification of prior findings
| # | Finding | Status |
|---|---------|--------|
| 1 | `check-ec` early-stop + first-failure reporting | ✅ **Fixed** — `call/cc` escape verified: `(check-ec (:range i 0 10) … => -1)` runs **1** iteration; `(check-ec (:range i 1 4) i => 0)` now reports `=> 1` (first), not `=> 3` (last). |
| 2 | `check-ec` qualifier count (`<qualifier>*`) | ⚠️ **Partial** — zero qualifiers now delegates to `check` ✓, but **multiple qualifiers still raise `InvalidSyntax`** (inline). |
| 3 | `(argument ...)` clause | ✅ Now parses (values still not printed on failure — acceptable). |
| 4 | `%first-fail` dead state | ✅ Wired into `check-report` — **but the wiring is broken, see blocker**. |
| 5 | `check-set-mode!` validation | ✅ **Fixed** — `(check-set-mode! 'bogus)` now errors. |

### 🔴 Blocker: `check-report` crashes on any run with a failure
`check-report` calls `(caddr %first-fail)`, but `caddr` is **not exported by `(scheme base)`** in Kaappi (it lives in `(scheme cxr)`). Inside the `(srfi 78)` library only the imported bindings are visible, so `caddr` is unbound. This works at the REPL (global env has every builtin) but fails inside the library. Minimal repro:

```scheme
(import (scheme base) (srfi 78))
(check-reset!)
(check (+ 1 1) => 3)
(check-report)   ; error: undefined variable 'caddr'
```

Output: `First failure: (+ 1 1) => 2 ; expected: ` → `error: undefined variable 'caddr'`. Note `car`, `cadr`, and `cddr` *are* in `(scheme base)` — only the 3-deep `caddr`/`caar`/… compositions are missing. Fix: add `(scheme cxr)` to the library imports, or use `(list-ref %first-fail 2)` / `(car (cddr %first-fail))` (all three verified working inside a base-only library).

**Why CI is green anyway:** the test suite never exercises this path — `srfi78.scm:88` calls `check-report` right after `check-reset!` + a *passing* check, so `%first-fail` is always `#f` and the buggy branch is skipped. Please add a test that calls `check-report` while a failure is pending (in `summary`/`off` mode so it doesn't `(exit 1)`), which would have caught this.

### Recommendation
Not ready to merge yet — the `caddr` crash is a regression in a core exported procedure. Once that's fixed and covered by a test, this is good to go. The remaining multiple-qualifier gap (#2) is optional and fine as a follow-up.

baijum (commented):
## Re-review of `248b093` + `073c37f` — ✅ Ready to merge

The blocker is fixed and the coverage gap is closed. I re-checked out the branch, rebuilt, and re-ran the full battery.

### Blocker resolved
- `check-report`'s `(caddr %first-fail)` → `(car (cddr %first-fail))` (`cddr` *is* in `(scheme base)`). Repro that crashed before now works:
  ```
  Checks: 1 total, 0 passed, 1 failed.
  First failure: (+ 1 1) => 2 ; expected: 3   (exit 1)
  ```
- Coverage gap closed two ways: a side-effect counter in `srfi78.scm` proves `check-ec` runs exactly **1** iteration on early-stop (would fail if the escape were removed), and `exit-code.sh` now exercises the `check-report` failure path as a subprocess — both new assertions pass (`SRFI-78 check-report with failure exits 1`, `SRFI-78 check-report prints first failure`).
- Nice bonus: the dead `equal` parameter on `check-ec-run` was removed.

### Full verification (fresh ReleaseSafe build at 073c37f)
| Check | Result |
|-------|--------|
| `check-report` after a failure | ✅ prints first failure, exits 1, no `caddr` crash |
| `check-ec` early-stop (`'summary`) | ✅ 1 iteration |
| `check-passed?` / `check-set-mode!` / invalid-mode error | ✅ |
| `srfi78.scm` | ✅ 20 expected passes |
| `errors/exit-code.sh` | ✅ 41 pass, 0 fail |
| Full `tests/scheme/srfi/*.scm` | ✅ 77 passed, 0 failed |

### Verdict
**Ready to merge.** All five original findings are addressed (four fully; multiple-qualifier support for `check-ec` — `(check-ec q1 q2 … => …)` still raises `InvalidSyntax`) is the only remaining gap, and it's a non-blocking enhancement beyond the scope of #1220 (zero and single qualifiers both work). Worth a quick follow-up issue if you want full `<qualifier>*` coverage, but nothing here should hold up the merge.

PR #1343, opened 2026-07-08
Fix SRFI-37 args-fold: short option matching, seed threading, option? export

### LINKED ISSUE #1211: [SRFI-37] args-fold: short char-name options never match (compared as strings), list-valued seeds are splatted, option? not exported
**Category:** SRFI
**Spec reference:** SRFI-37: option names are "either single characters (for short options) or strings (for long options)"; processors are called with the option, the name as given, the argument (or #f), followed by the seeds, and return the new seeds as values. `option?` is part of the SRFI's exported procedure set.
**Severity:** wrong-result (args-fold unusable for short options and list-shaped seeds)

```scheme
(import (scheme base) (srfi 37))
(define verbose (option '(#\v "verbose") #f #f
                        (lambda (opt name arg acc) (cons 'verbose acc))))

;; 1. short option never matches its char name — routed to unrecognized with a STRING name
(args-fold '("-v") (list verbose)
           (lambda (o n a acc) (cons (cons 'unrec n) acc))
           (lambda (op acc) acc) '())
;; expected: (verbose)   actual: ((unrec . "v"))  — and per spec the name should be #\v

;; 2. a processor returning a list has it re-interpreted as the whole seed tuple
(args-fold '("--verbose") (list verbose)
           (lambda (o n a acc) acc) (lambda (op acc) acc) '())
;; expected: (verbose)   actual: verbose  (the single list seed was unwrapped)

;; 3. option? is not exported
option?  ;; error: undefined variable 'option?'
```

**Root causes (lib/srfi/37.sld):**
- the short-option loop builds `(string (string-ref arg i))` and matches with `member` against names that are chars — string≠char, so lookup always fails, and processors/unrecognized receive a string name;
- seed threading uses the heuristic `(if (list? new-seeds) new-seeds (list new-seeds))` — any processor legitimately returning a list seed gets splatted into multiple seeds. Multiple seeds must be threaded via `call-with-values` around the processor call;
- the export list omits `option?` (the record predicate exists internally).

**Source location:** lib/srfi/37.sld
**Disabled test:** tests/scheme/srfi/srfi37.scm (marked ;; FAIL: #this)

baijum:
## Summary

Fixes #1211 — three bugs that made `args-fold` unusable for typical CLI parsing:

- **Short char-name options never matched** — the lookup built a string from the char (`"v"`) and compared it via `member` against names that are chars (`#\v`). Fixed by passing the char directly to `find-option`.
- **List-valued seeds were splatted** — the heuristic `(if (list? new-seeds) new-seeds (list new-seeds))` unwrapped any processor that legitimately returned a list as its single seed. Replaced all 8 sites with `call-with-values` to thread seeds correctly per the SRFI spec.
- **`option?` not exported** — added to the export list.

## Test plan

- [x] 29 SRFI-37 tests pass (up from ~12 with previously disabled tests now enabled)
- [x] Short options: `-v`, `-vv`, `-o out`, `-oout`, mixed `-v --output=out`
- [x] Char name passed to processors (not string)
- [x] List-valued seeds preserved through long opts, short opts, and operands
- [x] Multiple seeds via `values` threading
- [x] `option?` predicate works
- [x] Zig unit tests pass
- [x] R7RS test suite passes (1,391 tests)

### DIFF
--- a/lib/srfi/37.sld
+++ b/lib/srfi/37.sld
@@ -1,6 +1,6 @@
 (define-library (srfi 37)
   (import (scheme base))
-  (export args-fold
+  (export args-fold option?
           option option-names option-required-arg? option-optional-arg?
           option-processor)
   (begin
@@ -59,9 +59,10 @@
                  (let oloop ((operands rest) (seeds seeds))
                    (if (null? operands)
                        (apply values seeds)
-                       (let ((new-seeds (apply operand-proc (car operands) seeds)))
-                         (oloop (cdr operands)
-                                (if (list? new-seeds) new-seeds (list new-seeds)))))))
+                       (call-with-values
+                         (lambda () (apply operand-proc (car operands) seeds))
+                         (lambda new-seeds
+                           (oloop (cdr operands) new-seeds))))))
                 ((long-opt? arg)
                  (let* ((name (long-opt-name arg))
                         (opt-arg (long-opt-arg arg))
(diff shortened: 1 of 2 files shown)


baijum (commented):
## Review — ready to merge ✅

Clean, well-scoped fix that resolves all three defects in #1211. Verified locally (fresh ReleaseSafe build):

- ✅ Build passes
- ✅ 29/29 SRFI-37 tests pass, exit 0
- ✅ Issue's three minimal reproductions now behave per spec (`-v` → `(verbose)`, list seed preserved as `(verbose)`, `option?` resolves)
- ✅ Smoke tests (libraries, macros) still pass

### What's correct

- **Short char-name matching** — passing the char (`(string-ref arg i)`) instead of a 1-char string to `find-option` makes `member` compare char-to-char against the option's names. Also correctly propagates the char as `name` to the processor and to the synthesized option for the unrecognized path.
- **Seed threading** — replacing the `(if (list? new-seeds) new-seeds (list new-seeds))` heuristic with `call-with-values` + a variadic `(lambda new-seeds …)` is the spec-correct way to thread N seeds through processors that return them as multiple values. All 8 call sites converted consistently.
- **`option?` export** — the predicate already existed via `define-record-type`; just needed exporting.

Tests are auto-discovered by `run-all.sh` via the `tests/scheme/srfi/*.scm` glob, and the SRFI-64 exit-on-fail epilogue (grabbing the runner before `test-end`) is the correct pattern, so a regression here fails CI.

### Non-blocking follow-up (out of scope for #1211)

Optional-argument **short** options are still not honored. For an option declared `optional-arg? = #t`, `-oX` passes `#f` to the processor (not `"X"`) and then scans `X` as a further short option, because the short-opt branch only special-cases `option-required-arg?` — [lib/srfi/37.sld:91](lib/srfi/37.sld#L91). Long optional args via `--out=X` work correctly. This is a pre-existing gap (present before this PR) and the tests don't exercise optional-arg options, so it doesn't block this fix — worth a separate issue.

*(Reviewed by Claude Code)*

PR #1344, opened 2026-07-08
Rewrite SRFI-26 cut/cute with recursive helper macros (#1208)

### LINKED ISSUE #1208: [SRFI-26] cut/cute: cute is an alias of cut, <...> only works in 3 hardcoded positions, operator-position slots unsupported, arity capped
**Category:** SRFI
**Spec reference:** SRFI-26 grammar: `(cut <slot-or-expr> <slot-or-expr>* [<...>])` where every position including the operator may be a slot — "(cut <> a b)" is specified to become "(lambda (f) (f a b))". Semantics: "cut evaluates non-slot expressions at the time the resulting procedure is called, whereas cute evaluates the non-slot expressions at the time the procedure is constructed" — e.g. "(cute cons (+ a 1) <>)" becomes "(let ((a1 (+ a 1))) (lambda (x2) (cons a1 x2)))". Example "(cut list 1 <> 3 <> 5)" (5 argument positions) must expand.

```scheme
(import (scheme base) (srfi 26))

;; 1. cute evaluates non-slot expressions on EVERY call (it is (define-syntax cute ... (cut . args)))
(define n 0)
(define (tick) (set! n (+ n 1)) 10)
(define f (cute + (tick) <>))
(list (f 1) (f 2) n)   ;; expected: (11 12 1)   actual: (11 12 2)

;; 2. <...> after a fixed argument mis-matches a fixed-arity pattern
((cut list 1 <...>) 2 3)   ;; expected: (1 2 3)   actual: error "expected 0 arguments, got 2"

;; 3. operator-position slot unsupported
((cut <> 1 2) +)           ;; expected: 3   actual: error

;; 4. arity cap: more than ~3 fixed args / 2 slots has no matching pattern
(cut list 1 <> 3 <> 5)     ;; expected: 2-arg procedure   actual: expand-time error
```

**Analysis:** lib/srfi/26.sld enumerates a small finite set of syntax-rules patterns (up to 3 fixed args, 2 slots, 3 specific <...> shapes) and defines cute as a straight alias of cut. The standard implementation technique is a recursive helper macro that walks the argument list, collecting slot names and (for cute) let-binding non-slot expressions.

**Source location:** lib/srfi/26.sld
**Disabled test:** tests/scheme/srfi/srfi26.scm (marked ;; FAIL: #this)

**Found by:** Systematic audit, Phase 3a

baijum:
## Summary

- Replaces the hardcoded finite pattern set in `lib/srfi/26.sld` with the standard recursive-helper-macro technique from the SRFI-26 reference implementation
- `cut` now supports arbitrary arity, operator-position slots (`(cut <> 1 2)`), and correct `<...>` in any trailing position
- `cute` is a proper implementation that evaluates non-slot expressions once at construction time (was previously a plain alias of `cut`)
- Uses nested single-binding `let` for `cute` bindings to work around an expander limitation with ellipsis in `let` binding lists

Fixes #1208

## Test plan

- [x] All 4 reproductions from the issue now produce correct results
- [x] 34 SRFI-26 conformance tests pass (expanded from the audit baseline, all `FAIL` markers removed)
- [x] Full Scheme test suite: 1807 pass, 0 fail
- [x] Zig unit tests pass

### DIFF
--- a/lib/srfi/26.sld
+++ b/lib/srfi/26.sld
@@ -2,49 +2,47 @@
   (import (scheme base))
   (export cut cute)
   (begin
+    ;; Recursive helper for cut — accumulates slot-names and the call form.
+    ;; Slots become lambda parameters; all expressions stay in the body
+    ;; so they evaluate at call time.
+    (define-syntax srfi-26-internal-cut
+      (syntax-rules (<> <...>)
+        ((srfi-26-internal-cut (slot-name ...) (proc arg ...))
+         (lambda (slot-name ...) (proc arg ...)))
+        ((srfi-26-internal-cut (slot-name ...) (proc arg ...) <...>)
+         (lambda (slot-name ... . rest-slot) (apply proc arg ... rest-slot)))
+        ((srfi-26-internal-cut (slot-name ...) (proc arg ...) <> . se)
+         (srfi-26-internal-cut (slot-name ... x) (proc arg ... x) . se))
+        ((srfi-26-internal-cut (slot-name ...) (proc arg ...) nse . se)
+         (srfi-26-internal-cut (slot-name ...) (proc arg ... nse) . se))))
+
+    ;; Entry: separate the operator so (proc arg ...) always has >= 1 element.
     (define-syntax cut
       (syntax-rules (<> <...>)
-        ;; No slots — just call
-        ((cut f) f)
-        ;; Variadic
-        ((cut f <...>)
-         (lambda args (apply f args)))
-        ;; One slot patterns
-        ((cut f <>)
-         (lambda (x) (f x)))
-        ((cut f <> b)
-         (lambda (x) (f x b)))
-        ((cut f <> b c)
-         (lambda (x) (f x b c)))
-        ((cut f a <>)
-         (lambda (x) (f a x)))
-        ((cut f a <> c)
-         (lambda (x) (f a x c)))
-        ((cut f a b <>)
-         (lambda (x) (f a b x)))
-        ;; Two slot patterns
-        ((cut f <> <>)
-         (lambda (x y) (f x y)))
-        ((cut f <> <> c)
-         (lambda (x y) (f x y c)))
-        ((cut f <> a <>)
-         (lambda (x y) (f x a y)))
-        ((cut f a <> <>)
-         (lambda (x y) (f a x y)))
-        ;; Slot with variadic
-        ((cut f <> <...>)
-         (lambda (x . rest) (apply f x rest)))
-        ((cut f a <> <...>)
-         (lambda (x . rest) (apply f a x rest)))
-        ;; Fixed args (no slots)
-        ((cut f a)
-         (lambda () (f a)))
-        ((cut f a b)
-         (lambda () (f a b)))
-        ((cut f a b c)
-         (lambda () (f a b c)))))
+        ((cut <> . se)
+         (srfi-26-internal-cut (x) (x) . se))
+        ((cut f . se)
+         (srfi-26-internal-cut () (f) . se))))
 
+    ;; Recursive helper for cute — wraps each non-slot expression in a
+    ;; nested let immediately, so it evaluates once at construction time.
+    (define-syntax srfi-26-internal-cute
+      (syntax-rules (<> <...>)
+        ((srfi-26-internal-cute (slot-name ...) (proc arg ...))
+         (lambda (slot-name ...) (proc arg ...)))
+        ((srfi-26-internal-cute (slot-name ...) (proc arg ...) <...>)
+         (lambda (slot-name ... . rest-slot) (apply proc arg ... rest-slot)))
+        ((srfi-26-internal-cute (slot-name ...) (proc arg ...) <> . se)
+         (srfi-26-internal-cute (slot-name ... x) (proc arg ... x) . se))
+        ((srfi-26-internal-cute (slot-name ...) (proc arg ...) nse . se)
+         (let ((y nse))
+           (srfi-26-internal-cute (slot-name ...) (proc arg ... y) . se)))))
+
+    ;; Entry: operator slot passes through; operator expression is let-bound.
     (define-syntax cute
-      (syntax-rules ()
-        ((cute . args)
-         (cut . args))))))
+      (syntax-rules (<> <...>)
+        ((cute <> . se)
+         (srfi-26-internal-cute (x) (x) . se))
+        ((cute f . se)
+         (let ((t f))
+           (srfi-26-internal-cute () (t) . se)))))))
--- a/src/compiler_macro.zig
+++ b/src/compiler_macro.zig
@@ -272,7 +272,7 @@ pub fn compileDefineSyntax(self: *Compiler, args: Value, dst: u16) CompileError!
     }
 
     try captureLocalsOnTransformer(self, transformer);
-    computeBoundFreeRefs(self, transformer);
+    try computeBoundFreeRefs(self, transformer);
 
     const name = types.symbolName(keyword);
     try self.recordBodyMacro(name);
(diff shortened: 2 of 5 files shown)


lib/srfi/26.sld
```diff
@@ -2,49 +2,47 @@
...
+    ;; Slots become lambda parameters; all expressions stay in the body
+    ;; so they evaluate at call time.
+    (define-syntax srfi-26-internal-cut
+      (syntax-rules (<> <...>)
+        ((srfi-26-internal-cut (slot-name ...) (proc arg ...))
+         (lambda (slot-name ...) (proc arg ...)))
+        ((srfi-26-internal-cut (slot-name ...) (proc arg ...) <...>)
+         (lambda (slot-name ... . rest-slot) (apply proc arg ... rest-slot)))
+        ((srfi-26-internal-cut (slot-name ...) (proc arg ...) <> . se)
+         (srfi-26-internal-cut (slot-name ... x) (proc arg ... x) . se))
```
baijum, line 15: This injects a literal `x` as the slot binder and accumulates it through the recursion. On this branch, when a top-level `x` is in scope the expander fails to keep this `x` distinct, so slots capture the global:

```scheme
(define x 5)
((cut + <>) 10)          ; => 5      (want 10)
((cut list <> <>) 1 2)   ; => (1 1)  (want (1 2))
```

`main` returns these correctly, so it's a regression. Root cause is expander hygiene for recursively-accumulated binders — see the summary comment for a minimal, SRFI-26-independent repro.

lib/srfi/26.sld
```diff
@@ -2,49 +2,47 @@
...
+    (define-syntax srfi-26-internal-cute
+      (syntax-rules (<> <...>)
+        ((srfi-26-internal-cute (slot-name ...) (proc arg ...))
+         (lambda (slot-name ...) (proc arg ...)))
+        ((srfi-26-internal-cute (slot-name ...) (proc arg ...) <...>)
+         (lambda (slot-name ... . rest-slot) (apply proc arg ... rest-slot)))
+        ((srfi-26-internal-cute (slot-name ...) (proc arg ...) <> . se)
+         (srfi-26-internal-cute (slot-name ... x) (proc arg ... x) . se))
+        ((srfi-26-internal-cute (slot-name ...) (proc arg ...) nse . se)
+         (let ((y nse))
```
baijum, line 38: Same capture hazard as the slot `x`, for `cute`'s non-slot bindings: with a top-level `(define y ...)` in scope, `((cute + <>) 10)` returns the global `y`/`x` instead of 10.

lib/srfi/26.sld
```diff
@@ -2,49 +2,47 @@
...
+    ;; Entry: operator slot passes through; operator expression is let-bound.
     (define-syntax cute
-      (syntax-rules ()
-        ((cute . args)
-         (cut . args))))))
+      (syntax-rules (<> <...>)
+        ((cute <> . se)
+         (srfi-26-internal-cute (x) (x) . se))
+        ((cute f . se)
+         (let ((t f))
```
baijum, line 47: The `cute` operator binder `t` has the same exposure — a top-level `(define t ...)` will be captured. If the expander fix is deferred, at minimum these injected names (`x`, `y`, `t`, `rest-slot`) are the ones to worry about.

baijum (commented):
## Review — SRFI-26 `cut`/`cute` rewrite

The recursive-helper approach is the right standard technique, and it does close every gap in #1208: operator-position slots, arbitrary arity, trailing `<...>`, and `cute` evaluating non-slot expressions once at construction. The 4 issue reproductions and all 34 conformance tests pass **when run in isolation**, and the new test file is a big improvement over the old FAIL-marked one.

**But I don't think it's ready to merge as-is — it introduces a *silent* correctness regression.**

### The problem
The new helpers inject the literal identifiers `x` (slots, lines 15/36), `y` (cute non-slot bindings, line 38) and `t` (cute operator, line 47). When a **top-level** binding of the same name is in scope, expansion is not hygienic and `cut`/`cute` return wrong results with no error:

```scheme
(import (scheme base) (srfi 26))
(define x 5)
((cut + <>) 10)          ; => 5      (should be 10)
((cut list <> <>) 1 2)   ; => (1 1)  (should be (1 2))
(define y 99)
((cute + <>) 10)         ; => 5      (should be 10)
```

This is a **regression vs `main`**: the old finite-pattern macros return `((cut + <>) 10)` => `10` correctly even with a top-level `x`, because each lambda is emitted from a single template the expander renames properly. The recursive accumulation defeats that renaming.

It does *not* fire for lexical bindings (a `let`/`lambda` parameter named `x` is fine) — only top-level `define`s. But top-level definitions are common and `x`/`y`/`t` are among the most common names, so this will bite real code, silently.

### Root cause is in the expander, not this macro
The macro matches the SRFI-26 reference implementation and is correct. The underlying bug is hygiene: a macro-introduced binder **accumulated across recursive macro self-invocation** collides with a top-level binding of the same source name. Minimal, SRFI-26-independent repro:

```scheme
(define-syntax build
  (syntax-rules ()
    ((build (p ...) () (c ...))       (lambda (p ...) (c ...)))
    ((build (p ...) (h . t) (c ...))  (build (p ... q) t (c ... q)))))
(define-syntax make (syntax-rules () ((make . xs) (build () xs (list)))))

((make a b) 1 2)     ; => (1 2)
(define q 999)
((make a b) 1 2)     ; => (1 1)   <-- accumulated binders collapse
```

So the real fix belongs in `expander.zig` (it's a latent bug affecting *any* recursive binder-accumulating macro, not just SRFI-26). I'd suggest fixing that first and landing this rewrite on top, guarded by a regression test.

### Test gap
The new suite never binds `x`/`y`/`t` at top level, which is exactly why the regression is invisible in CI. Please add a test that defines a top-level `x` and then exercises `(cut ... <> ...)` / `(cute ... <> ...)` — that's the case that must pass before this is safe.

### Verdict
**Not ready to merge.** Good direction and a real coverage win, but it currently makes `cut`/`cute` silently wrong whenever a top-level `x`/`y`/`t` is in scope. Once the expander hygiene is fixed (or the injected binders are otherwise guaranteed uncapturable) and a shadowing regression test is added, this should be good to go.

<sub>Reviewed with fresh ReleaseSafe builds of both this branch and `main`; results verified with explicit `--lib-path` from a neutral CWD to avoid loading the stale `~/.kaappi/lib` / worktree copies of `(srfi 26)`.</sub>

baijum:
Fixed. The root cause was in the VOID-marking trick in `expandAndCompileMacroUse` — it was suppressing hygiene renaming for *any* template free-ref candidate that matched a global, including template-introduced binders that only coincidentally shared a name.

**Fix (4f2ddf8):** A new `bound_free_refs` field on the `Transformer` records which free-ref candidates were actually in scope at macro *definition* time. The VOID marking now only applies to those identifiers, so template-introduced names like `x`/`y`/`t` are correctly renamed regardless of later user defines.

Both the reviewer's minimal repro and the SRFI-26 shadowing case now produce correct results:
```scheme
(define x 5)
((cut + <>) 10)          ;; => 10 ✓
((cut list <> <>) 1 2)   ;; => (1 2) ✓
(define q 999)
((make a b) 1 2)         ;; => (1 2) ✓
```

All 1807 existing tests pass (including the referential transparency tests in `hygiene/template-set-global.scm`, `hygiene/arg-vs-free.scm`, `hygiene/define-syntax-capture.scm`, and the R7RS jabberwocky test). 6 new hygiene regression tests added to the SRFI-26 suite (40 total).

baijum:
## Re-review — the two new commits resolve the blocker ✅

`4f2ddf84` (expander hygiene) + `ba9f2675` (OOM propagation + hygiene tests) fully address my earlier finding. Nice fix — computing free-references once at macro **definition** time and only shielding identifiers actually bound then (`in_globals || in_def_env || in_locals || in_macros`) is the correct semantics; template-introduced binders that merely collide with a *later* top-level `define` now get renamed as they should.

Verified locally on a fresh `ReleaseSafe` build of `ba9f2675` (explicit `--lib-path`, neutral CWD, new-only sentinel confirmed):

- All three original blocker cases now pass — `(define x 5) ((cut + <>) 10)` ⇒ `10`, `((cut list <> <>) 1 2)` ⇒ `(1 2)`, `(define y 99) ((cute + <>) 10)` ⇒ `10`; likewise for top-level `t` and `rest-slot`.
- **No forward-reference regression** — a macro whose template references a global defined *later* still resolves correctly (renamed globals fall back to the top-level binding).
- **Referential transparency preserved** — a template's free ref to a def-env global is *not* captured by a same-named local at the use site (`(let ((g 999)) (getg))` ⇒ `100`).
- Nested `cut`, and `cut` inside a `let` that shadows `x`, both correct.
- SRFI-26 suite: **42 pass, 0 fail** (the new top-level `x`/`y`/`t`/`rest-slot` hygiene tests are exactly the missing coverage I flagged).
- Since the change is in the core macro expander: full Scheme suite **1807 pass / 0 fail**, `zig build test` green.

**Verdict: ready to merge once CI goes green** (jobs are currently pending; the checks I ran locally mirror them). No code concerns remain.

Minor, non-blocking nit for a follow-up (not this PR): `computeBoundFreeRefs` silently `return`s when the 64-symbol scratch arrays overflow, where the old inline code raised `CompileError.InternalLimit`. For a pathological macro with >64 distinct pattern/template symbols, def-env refs would no longer be shielded (it degrades toward *more* renaming, so it stays hygiene-safe, and the global fallback covers the common case — hence not a blocker). Worth a comment noting the limit, or bumping/erroring if you'd rather keep the old strictness.

PR #1346, opened 2026-07-08
Fix SRFI-42 comprehensions: recursive qualifiers, guards, and missing generators

### LINKED ISSUE #1216: [SRFI-42] Comprehensions accept only a single qualifier — no nesting, no guards, and :string/:vector/:let/:while/:until fail inside list-ec despite being exported
**Category:** SRFI
**Spec reference:** SRFI-42: "(list-ec ⟨qualifier⟩* ⟨expression⟩)" — comprehensions take any number of qualifiers; qualifiers include generators, guards ("(if ⟨test⟩)", "(not ⟨test⟩)", ...), ":let", ":while", ":until", and generators nest left-to-right ("the rightmost generator spins fastest").
**Severity:** missing-feature (large functional subset absent; forms compile-error)

```scheme
(import (scheme base) (srfi 42))

(list-ec (:list a '(1 2)) (:list b '(3 4)) (cons a b))  ;; nesting → compile error: InvalidSyntax
(list-ec (:range i 6) (if (even? i)) i)                 ;; guard → compile error
(list-ec (:string c "ab") c)                            ;; compile error (exported but unusable in list-ec)
(list-ec (:vector x #(1 2)) x)                          ;; compile error
(list-ec (:let x 10) x)                                 ;; compile error
(list-ec (:range i 100) (:while (< i 3)) i)             ;; compile error
```

Single-qualifier forms work: `(list-ec (:range i 4) i)`, `sum-ec`, `product-ec`, `min/max-ec`, `first/last-ec`, `any?/every?-ec`, `fold-ec`, `do-ec` with one generator all pass.

**Root cause:** lib/srfi/42.sld hand-writes one syntax-rules pattern per (comprehension × single generator) pair instead of the SRFI's nested expansion; there is no qualifier recursion, so multiple qualifiers, guards, and most generator types have no pattern and fall through to InvalidSyntax. `:string`/`:vector`/`:while`/`:until`/`:parallel`/`:port`/`:do` are exported only as error-stub macros.

**Source location:** lib/srfi/42.sld
**Disabled test:** tests/scheme/srfi/srfi42.scm (marked ;; FAIL: #this)

**Found by:** Systematic audit, Phase 3c

baijum:
## Summary

Closes #1216.

- Rewrites `lib/srfi/42.sld` with a two-macro architecture (`do-ec` + internal `%do-ec`) that recursively processes qualifiers one at a time, enabling arbitrary nesting
- Adds support for `:string`, `:vector`, `:integers`, `:let` generators, `(if ...)` / `(not ...)` / `(and ...)` / `(or ...)` guards, and `:while` / `:until` early-exit control qualifiers
- Corrects `fold-ec` signature to match SRFI-42 spec: `(fold-ec seed qualifier... expr proc)`
- Uses a mutable stop flag (threaded through `%do-ec`) so `:while`/`:until` can terminate all enclosing generator loops without `call/cc`
- Rewrites test file with SRFI-64, adds 31 tests covering all previously-failing cases

## Test plan

- [x] `zig build run -- tests/scheme/srfi/srfi42.scm` — 31/31 pass
- [x] `bash tests/scheme/run-all.sh` — 1807 pass, 0 fail, 0 regressions

### DIFF
--- a/lib/srfi/42.sld
+++ b/lib/srfi/42.sld
@@ -101,6 +101,28 @@
          (let ((var expr))
            (%do-ec s rest1 rest2 ...)))
         ;; --- control qualifiers ---
+        ;; :while/:until wrapping a generator (spec form)
+        ((_ s (:while (:range rargs ...) test) rest1 rest2 ...)
+         (%do-ec s (:range rargs ...) (:while test) rest1 rest2 ...))
+        ((_ s (:while (:list largs ...) test) rest1 rest2 ...)
+         (%do-ec s (:list largs ...) (:while test) rest1 rest2 ...))
+        ((_ s (:while (:string sargs ...) test) rest1 rest2 ...)
+         (%do-ec s (:string sargs ...) (:while test) rest1 rest2 ...))
+        ((_ s (:while (:vector vargs ...) test) rest1 rest2 ...)
+         (%do-ec s (:vector vargs ...) (:while test) rest1 rest2 ...))
+        ((_ s (:while (:integers iargs ...) test) rest1 rest2 ...)
+         (%do-ec s (:integers iargs ...) (:while test) rest1 rest2 ...))
+        ((_ s (:until (:range rargs ...) test) rest1 rest2 ...)
+         (%do-ec s (:range rargs ...) (:until test) rest1 rest2 ...))
+        ((_ s (:until (:list largs ...) test) rest1 rest2 ...)
+         (%do-ec s (:list largs ...) (:until test) rest1 rest2 ...))
+        ((_ s (:until (:string sargs ...) test) rest1 rest2 ...)
+         (%do-ec s (:string sargs ...) (:until test) rest1 rest2 ...))
+        ((_ s (:until (:vector vargs ...) test) rest1 rest2 ...)
+         (%do-ec s (:vector vargs ...) (:until test) rest1 rest2 ...))
+        ((_ s (:until (:integers iargs ...) test) rest1 rest2 ...)
+         (%do-ec s (:integers iargs ...) (:until test) rest1 rest2 ...))
+        ;; :while/:until as standalone qualifiers
         ((_ s (:while test) rest1 rest2 ...)
          (if test
            (%do-ec s rest1 rest2 ...)
@@ -139,18 +161,18 @@
       (syntax-rules ()
         ((_ seed qualifier ... expr f)
          (let ((%ec-acc seed))
-           (do-ec qualifier ... (set! %ec-acc (f %ec-acc expr)))
+           (do-ec qualifier ... (set! %ec-acc (f expr %ec-acc)))
            %ec-acc))))
 
     (define-syntax fold3-ec
       (syntax-rules ()
-        ((_ qualifier ... expr f1 f2)
-         (let ((%ec-first #t) (%ec-acc #f))
+        ((_ x0 qualifier ... expr f1 f2)
+         (let ((%ec-first #t) (%ec-acc x0))
            (do-ec qualifier ...
              (if %ec-first
                (begin (set! %ec-acc (f1 expr))
                       (set! %ec-first #f))
-               (set! %ec-acc (f2 %ec-acc expr))))
+               (set! %ec-acc (f2 expr %ec-acc))))
            %ec-acc))))
 
     (define-syntax string-ec
--- a/lib/srfi/42.sld
+++ b/lib/srfi/42.sld
@@ -10,152 +10,208 @@
           :port :do :let :parallel :while :until)
   (begin
 
-    ;; Core: do-ec iterates, fold-ec accumulates
+    ;; Generator keywords — error when used outside a comprehension.
+    ;; Defined first so %do-ec can resolve them as syntax-rules literals.
+    (define-syntax :list
+      (syntax-rules ()
+        ((_ rest ...) (error ":list used outside a comprehension"))))
+    (define-syntax :string
+      (syntax-rules ()
+        ((_ rest ...) (error ":string used outside a comprehension"))))
+    (define-syntax :vector
+      (syntax-rules ()
+        ((_ rest ...) (error ":vector used outside a comprehension"))))
+    (define-syntax :range
+      (syntax-rules ()
+        ((_ rest ...) (error ":range used outside a comprehension"))))
+    (define-syntax :integers
+      (syntax-rules ()
+        ((_ rest ...) (error ":integers used outside a comprehension"))))
+    (define-syntax :port
+      (syntax-rules ()
+        ((_ rest ...) (error ":port used outside a comprehension"))))
+    (define-syntax :do
+      (syntax-rules ()
+        ((_ rest ...) (error ":do used outside a comprehension"))))
+    (define-syntax :let
+      (syntax-rules ()
+        ((_ rest ...) (error ":let used outside a comprehension"))))
+    (define-syntax :parallel
+      (syntax-rules ()
+        ((_ rest ...) (error ":parallel used outside a comprehension"))))
+    (define-syntax :while
+      (syntax-rules ()
+        ((_ rest ...) (error ":while used outside a comprehension"))))
+    (define-syntax :until
+      (syntax-rules ()
+        ((_ rest ...) (error ":until used outside a comprehension"))))
+
+    ;; Internal recursive qualifier processor.
+    ;; s = mutable stop flag for :while/:until early exit.
+    ;; Processes one qualifier per expansion, recurses on the rest.
+    (define-syntax %do-ec
+      (syntax-rules (:range :list :string :vector :integers
+                     :let :while :until if not and or)
+        ;; --- generators ---
+        ((_ s (:range var n) rest1 rest2 ...)
+         (let ((%n n))
+           (let %lp ((var 0))
+             (when (and (< var %n) (not s))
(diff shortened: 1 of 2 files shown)


lib/srfi/42.sld
```diff
@@ -10,152 +10,208 @@
...
 
-    (define-syntax vector-ec
-      (syntax-rules (:list :range)
-        ((_ qualifier expr)
-         (list->vector (list-ec qualifier expr)))))
+    (define-syntax fold-ec
+      (syntax-rules ()
+        ((_ seed qualifier ... expr f)
+         (let ((%ec-acc seed))
+           (do-ec qualifier ... (set! %ec-acc (f %ec-acc expr)))
```
baijum: **Correctness bug: fold procedure called with reversed arguments.**

SRFI-42 defines the reduction step as `(set! x (f <value> x))` — the generated value first, accumulator second. This calls `(f %ec-acc expr)`, i.e. accumulator first. For commutative `f` (`+`, `*`) it's invisible, which is why `sum-ec`/`product-ec` and the one `fold-ec` test (`+`) pass — but it's wrong for anything else:

```scheme
(fold-ec '() (:range i 3) i cons)   ; spec: (2 1 0)
                                    ; this PR: (((() . 0) . 1) . 2)
(fold-ec 0 (:range i 1 4) i -)      ; spec: 2 ; this PR: -6
```

Fix: `(set! %ec-acc (f expr %ec-acc))`. Please also add a `fold-ec` test with a non-commutative proc (e.g. `cons` or `-`) so this can't silently regress.

baijum (commented):
Reviewed the SRFI-42 rewrite. The two-macro `do-ec` / `%do-ec` architecture is a big step up from the old one-pattern-per-generator table: nesting (cartesian products), `(if/not/and/or)` guards, `:let`, `:string`, `:vector`, `:integers`, and negative-step `:range` all work now, and the mutable stop-flag for early exit without `call/cc` is a neat trick. The 31 new SRFI-64 tests pass (verified locally), and `run-all.sh` is clean.

That said, I found one **correctness bug** and two **spec-conformance gaps** that I think should be fixed before merge, since this PR's whole purpose is SRFI-42 conformance (closes #1216). I verified each against the SRFI-42 spec and by running the PR's own `42.sld`. Details are in the inline comments; summary:

1. **`fold-ec` calls the accumulator procedure with reversed arguments** (bug — produces wrong results for non-commutative procs). Masked by the test suite because `sum-ec`/`product-ec` use `+`/`*` and the only `fold-ec` test uses `+`.
2. **`:while` / `:until` use a non-standard syntax** — the SRFI spec form `(:while <generator> <test>)` errors; only `(<generator>) (:while <test>)` works. (The repro in #1216 used the non-standard form, so the PR matches the issue but not the spec.)
3. **`fold3-ec` signature is wrong** — missing the leading `x0`, wrong empty-sequence result, reversed `f2` args, and no tests.

Also a minor gap: `:list`/`:string`/`:vector` accept only a single sequence (spec allows multiple). Nice work overall — these are all localized fixes.

lib/srfi/42.sld
```diff
@@ -10,152 +10,208 @@
...
+        ((_ s (:integers var) rest1 rest2 ...)
+         (let %lp ((var 0))
+           (when (not s)
+             (%do-ec s rest1 rest2 ...)
+             (%lp (+ var 1)))))
+        ((_ s (:let var expr) rest1 rest2 ...)
+         (let ((var expr))
+           (%do-ec s rest1 rest2 ...)))
+        ;; --- control qualifiers ---
+        ((_ s (:while test) rest1 rest2 ...)
```
baijum, line 126: **Non-standard `:while` / `:until` syntax.**

SRFI-42 defines these as *generator combinators* that take a generator as their first argument: `(:while <generator> <test>)` and `(:until <generator> <test>)`. With this implementation the canonical spec form errors:

```scheme
(list-ec (:while (:range i 10) (< i 5)) i)   ; spec: (0 1 2 3 4)
                                             ; this PR: compile error: InvalidSyntax
```

Only the standalone `(<generator>) (:while <test>)` placement works. The minimal repro in #1216 happens to use the standalone form, so the PR satisfies the issue, but portable SRFI-42 code will use `(:while <generator> <test>)`. Worth supporting the spec form (a pattern like `(_ s (:while (g arg ...) test) rest ...)` that rewrites into the generator + a standalone `:while`).

lib/srfi/42.sld
```diff
@@ -10,152 +10,208 @@
...
+    (define-syntax fold-ec
+      (syntax-rules ()
+        ((_ seed qualifier ... expr f)
+         (let ((%ec-acc seed))
+           (do-ec qualifier ... (set! %ec-acc (f %ec-acc expr)))
+           %ec-acc))))
+
+    (define-syntax fold3-ec
+      (syntax-rules ()
+        ((_ qualifier ... expr f1 f2)
```
baijum: **`fold3-ec` signature deviates from the spec and is untested.**

Spec: `(fold3-ec <x0> <qualifier>* <expr> <f1> <f2>)`, where `(f1 value)` seeds from the first element, `(f2 value acc)` folds the rest, and **`x0` is returned iff the sequence is empty**. This PR:
- drops the leading `x0` (so `(fold3-ec 'empty (:range i 0) i f1 f2)` → InvalidSyntax);
- returns `#f` on the empty case instead of `x0`;
- calls `(f2 %ec-acc expr)` — reversed, same issue as `fold-ec` above (should be `(f2 expr %ec-acc)`).

There are no `fold3-ec` tests, so none of this is exercised. Please align the signature with the spec and add coverage (including the empty-sequence case).

lib/srfi/42.sld
```diff
@@ -10,152 +10,208 @@
...
+               (%do-ec s rest1 rest2 ...)
+               (%lp (+ var 1))))))
+        ((_ s (:range var a b c) rest1 rest2 ...)
+         (let ((%b b) (%c c))
+           (let %lp ((var a))
+             (when (and (if (positive? %c) (< var %b) (> var %b))
+                        (not s))
+               (%do-ec s rest1 rest2 ...)
+               (%lp (+ var %c))))))
+        ((_ s (:list var lst) rest1 rest2 ...)
```
baijum, line 75: Minor / optional: SRFI-42 allows `:list`, `:string`, and `:vector` to take *multiple* sequences ("think of it as first appending all arguments"): `(:list var s1 s2 ...)`. These patterns only match a single sequence, so `(list-ec (:list x '(1 2) '(3 4)) x)` errors. Fine to defer, but worth a note/TODO if not handled here.

baijum (commented):
## Follow-up review of b6e6454 — all blocking issues resolved ✅

Re-checked against the SRFI-42 spec and verified each fix by running the updated `42.sld` locally:

| Finding | Status | Verified |
|---|---|---|
| **#1 `fold-ec` reversed args** | ✅ Fixed | `(fold-ec '() (:range i 3) i cons)` → `(2 1 0)`; `(fold-ec 0 (:range i 1 4) i -)` → `2` |
| **#2 `:while`/`:until` spec form** | ✅ Fixed | `(list-ec (:while (:range i 10) (< i 5)) i)` → `(0 1 2 3 4)`; `:until` form → `(0 1 2 3 4 5)`. Standalone form still works. |
| **#3 `fold3-ec` signature** | ✅ Fixed | `x0` added; empty → returns `x0`; `f2` now `(f2 value acc)`. `(fold3-ec 'empty (:range i 0) …)` → `empty`; non-empty → correct |

The generator-wrapping approach for `:while`/`:until` (rewrite `(:while (:range …) test)` into `(:range …) (:while test)`) is a clean way to support both the spec form and the standalone form. Test count is up 31 → 38, and the new tests cover exactly the previously-untested cases (non-commutative `fold-ec`, `fold3-ec` incl. empty, `and`/`or` guards, `:while` wrapping a generator). Full suite passes; CI is green across the matrix.

**One non-blocking item remains** (from my original comment #4): `:list`/`:string`/`:vector` still accept only a single sequence, whereas SRFI-42 allows multiple (`(:list var s1 s2 …)`). This is a minor completeness gap, fine to leave for a follow-up — worth a TODO note if not addressed here.

**Verdict: ready to merge.** 🚀 (I'm the PR author's assistant reviewer, so this is a COMMENT rather than a formal approval — GitHub won't let the author self-approve.)

baijum:
✅ **Re-verified at `b6e6454` — ready to merge.**

All three blocking issues from the earlier review are resolved (checked against the SRFI-42 spec and by running the updated `42.sld` locally):

- **`fold-ec`** now calls the reducer as `(f value acc)` — `(fold-ec '() (:range i 3) i cons)` → `(2 1 0)`, `(fold-ec 0 (:range i 1 4) i -)` → `2`
- **`:while`/`:until`** support the spec generator-wrapping form — `(list-ec (:while (:range i 10) (< i 5)) i)` → `(0 1 2 3 4)` — while the standalone form still works
- **`fold3-ec`** takes the `x0` seed, returns `x0` on an empty sequence, and uses `(f2 value acc)`

Test count is up 31 → 38 with coverage for exactly the previously-broken cases; the full suite passes and CI is green across the matrix. Branch is `MERGEABLE` / `CLEAN`.

One **non-blocking** completeness gap remains: `:list`/`:string`/`:vector` accept only a single sequence, whereas SRFI-42 allows multiple (`(:list var s1 s2 …)`). Fine to defer to a follow-up.

LGTM 🚀

PR #1349, opened 2026-07-08
Implement SRFI-17 generalized set! with pre-defined setters

### LINKED ISSUE #1205: [SRFI-17] Library is a stub: no generalized set!, no pre-defined setters — (setter car) raises
**Category:** SRFI
**Spec reference:** SRFI-17: "We define: (set! (proc arg ...) value) as: ((setter proc) arg ... value)" and "The following standard procedures have pre-defined setters:" — car, cdr, all caXXr/cdXXr compositions, string-ref, and vector-ref, e.g. "(set! (car x) v) == (set-car! x v)".
**Severity:** missing-feature

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 17))
(setter car)                              ;; expected: set-car!   actual: error "no setter defined for procedure"
(let ((p (list 1 2))) (set! (car p) 99) p) ;; expected: (99 2)    actual: compile error (set! target must be identifier)
```

**Analysis:** lib/srfi/17.sld only maintains a user-populated assq table for `getter-with-setter` — the table starts empty (no pre-defined setters for car/cdr/vector-ref/string-ref/cxr compositions), and the generalized `(set! (proc args) value)` syntax is not hooked into the compiler's set! form at all. `getter-with-setter` itself works for user-registered getters.

**Source location:** lib/srfi/17.sld; generalized set! would need support in src/compiler_ir.zig's set! handling
**Disabled test:** tests/scheme/srfi/srfi17.scm (marked ;; FAIL: #this)

**Found by:** Systematic audit, Phase 3a

baijum:
## Summary

- Compiler desugars `(set! (proc arg ...) val)` → `((setter proc) arg ... val)` in both the IR lowering path (`lowerSet`) and the legacy compiler path (`compileSet`)
- SRFI-17 library registers 32 pre-defined setters: `car`, `cdr`, `vector-ref`, `string-ref`, and all 28 cXXr compositions
- 16 tests covering pre-defined setters, generalized set! syntax, user-defined setters via `getter-with-setter`, and cXXr compositions

Closes #1205

## Test plan

- [x] `zig build run -- tests/scheme/srfi/srfi17.scm` — 16 pass, 0 fail
- [x] `bash tests/scheme/run-all.sh` — 1807 pass, 0 fail
- [x] Manual: `(import (srfi 17)) (let ((p (list 1 2))) (set! (car p) 99) p)` → `(99 2)`

### DIFF
--- a/src/compiler_lambda.zig
+++ b/src/compiler_lambda.zig
@@ -785,39 +785,45 @@ pub fn compileSet(self: *Compiler, args: Value, dst: u16) CompileError!void {
 
     // SRFI-17 generalized set!: (set! (proc arg ...) val)
     // Desugar to: ((setter proc) arg ... val) and compile as a call.
+    // Defensive fallback — the IR path (lowerSet) handles this first;
+    // this branch fires only if compilation routes through compileExpr.
+    // Requires (srfi 17) imported so the global `setter` is defined.
     if (types.isPair(target)) {
         const proc = types.car(target);
         const proc_args = types.cdr(target);
         const val_expr = types.car(rest);
 
-        // Build (setter proc)
-        const setter_sym = self.gc.allocSymbol("setter") catch return CompileError.OutOfMemory;
-        const setter_proc_pair = self.gc.allocPair(proc, types.NIL) catch return CompileError.OutOfMemory;
-        var setter_call = self.gc.allocPair(setter_sym, setter_proc_pair) catch return CompileError.OutOfMemory;
-        self.gc.pushRoot(&setter_call);
-        defer self.gc.popRoot();
-
-        // Build args: append val to proc_args → (arg1 arg2 ... val)
-        var ext_args = self.gc.allocPair(val_expr, types.NIL) catch return CompileError.OutOfMemory;
-        // Collect proc_args into stack buffer, then prepend in reverse
         var arg_buf: [16]Value = undefined;
         var n_args: usize = 0;
         var cur = proc_args;
         while (cur != types.NIL) {
             if (!types.isPair(cur)) return CompileError.InvalidSyntax;
-            if (n_args >= 16) return CompileError.InvalidSyntax;
+            if (n_args >= 16) return CompileError.InternalLimit;
             arg_buf[n_args] = types.car(cur);
             n_args += 1;
             cur = types.cdr(cur);
         }
-        var i = n_args;
-        while (i > 0) {
-            i -= 1;
-            ext_args = self.gc.allocPair(arg_buf[i], ext_args) catch return CompileError.OutOfMemory;
-        }
 
-        // Build ((setter proc) arg1 ... val) and compile as expression
-        const desugared = self.gc.allocPair(setter_call, ext_args) catch return CompileError.OutOfMemory;
+        // Suppress GC during S-expression construction — intermediate
+        // pairs are not reachable from any root until `desugared` is
+        // rooted (same pattern as compileDefineValues, issue #1010).
+        var desugared: Value = undefined;
+        {
+            self.gc.no_collect += 1;
+            defer self.gc.no_collect -= 1;
+            const setter_sym = self.gc.allocSymbol("setter") catch return CompileError.OutOfMemory;
+            const setter_proc_pair = self.gc.allocPair(proc, types.NIL) catch return CompileError.OutOfMemory;
+            const setter_call = self.gc.allocPair(setter_sym, setter_proc_pair) catch return CompileError.OutOfMemory;
+            var ext_args = self.gc.allocPair(val_expr, types.NIL) catch return CompileError.OutOfMemory;
+            var i = n_args;
+            while (i > 0) {
+                i -= 1;
+                ext_args = self.gc.allocPair(arg_buf[i], ext_args) catch return CompileError.OutOfMemory;
+            }
+            desugared = self.gc.allocPair(setter_call, ext_args) catch return CompileError.OutOfMemory;
+        }
+        self.gc.pushRoot(&desugared);
+        defer self.gc.popRoot();
         return self.compileExprViaIR(desugared, dst, false);
     }
 
--- a/src/compiler_lambda.zig
+++ b/src/compiler_lambda.zig
@@ -782,6 +782,45 @@ pub fn compileSet(self: *Compiler, args: Value, dst: u16) CompileError!void {
     const target = types.car(args);
     const rest = types.cdr(args);
     if (rest == types.NIL) return CompileError.InvalidSyntax;
+
+    // SRFI-17 generalized set!: (set! (proc arg ...) val)
+    // Desugar to: ((setter proc) arg ... val) and compile as a call.
+    if (types.isPair(target)) {
+        const proc = types.car(target);
+        const proc_args = types.cdr(target);
+        const val_expr = types.car(rest);
+
+        // Build (setter proc)
+        const setter_sym = self.gc.allocSymbol("setter") catch return CompileError.OutOfMemory;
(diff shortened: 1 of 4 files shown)


src/ir.zig
```diff
@@ -650,10 +650,42 @@ fn lowerDefine(ir: *IR, expr: Value) CompileError!*Node {
...
+    // SRFI-17 generalized set!: (set! (proc arg ...) val)
+    // Desugar to: ((setter proc) arg ... val)
+    if (types.isPair(name)) {
+        const c = ir.compiler orelse return CompileError.InvalidSyntax;
+        const proc = types.car(name);
+        const proc_args = types.cdr(name);
+        const val_expr = types.car(rest);
+
+        // Build (setter proc) call node
+        const setter_sym = c.gc.allocSymbol("setter") catch return CompileError.OutOfMemory;
```
baijum, line 673: The desugaring unconditionally emits a reference to an unqualified global `setter`. That's the correct SRFI-17 expansion, but it means *every* `(set! (f x) v)` form now depends on a global named `setter` being in scope. When `(srfi 17)` isn't imported (or is imported with a prefix/rename), the diagnostic becomes `undefined variable 'setter'. Did you mean 'filter'?` rather than something pointing at generalized `set!`.

Not a blocker — it matches how most implementations desugar — but worth a comment here noting the dependency on a global `setter` so the somewhat cryptic error isn't a surprise later.

src/compiler_lambda.zig
```diff
@@ -782,6 +782,45 @@ pub fn compileSet(self: *Compiler, args: Value, dst: u16) CompileError!void {
     const target = types.car(args);
     const rest = types.cdr(args);
     if (rest == types.NIL) return CompileError.InvalidSyntax;
+
+    // SRFI-17 generalized set!: (set! (proc arg ...) val)
+    // Desugar to: ((setter proc) arg ... val) and compile as a call.
+    if (types.isPair(target)) {
```
baijum, line 791: This duplicates the desugaring already added to `lowerSet` in `ir.zig`. As far as I can tell this legacy `compileSet` branch is effectively unreachable for generalized `set!`: `set!` is always lowered through `lowerSet` in the IR path, and the only caller of this legacy function (`compileSetValues`, line ~756, the single-variable `define-values` case) passes a symbol-target `set!`. Two copies of the same desugaring that must be kept in sync is a maintenance risk.

Suggestion: either add a short comment explaining this is a defensive fallback, or drop the block if you can confirm it's dead (a quick way to confirm: temporarily `@panic` here and run `run-all.sh` — nothing should trip it).

src/compiler_lambda.zig
```diff
@@ -782,6 +782,45 @@ pub fn compileSet(self: *Compiler, args: Value, dst: u16) CompileError!void {
...
+
+        // Build args: append val to proc_args → (arg1 arg2 ... val)
+        var ext_args = self.gc.allocPair(val_expr, types.NIL) catch return CompileError.OutOfMemory;
+        // Collect proc_args into stack buffer, then prepend in reverse
+        var arg_buf: [16]Value = undefined;
+        var n_args: usize = 0;
+        var cur = proc_args;
+        while (cur != types.NIL) {
+            if (!types.isPair(cur)) return CompileError.InvalidSyntax;
+            if (n_args >= 16) return CompileError.InvalidSyntax;
```
baijum: Minor: the 16-argument cap reports `InvalidSyntax`, which is misleading — it's an implementation arity limit, not a syntax error. The IR path (`lowerSet`) has no such cap (it uses an `ArrayList`), so the two paths also disagree on behavior. Given a setter realistically takes 2–3 operands this never bites in practice, but if this branch is kept, `InternalLimit` would be the more accurate error (mirroring `lowerCall`'s 256-arg guard).

baijum (commented):
## Review — SRFI-17 generalized `set!`

Reviewed the code and verified it locally on a fresh ReleaseSafe build (merged `fix/1205` into a review worktree):

- `zig build test` — pass
- `tests/scheme/srfi/srfi17.scm` — 16 pass, 0 fail
- `bash tests/scheme/run-all.sh` — **1807 pass, 0 fail** (matches the PR description)
- Extra manual checks all correct: computed args/values (`(set! (vector-ref v (+ 1 1)) (* 2 3))`), left-to-right evaluation order (object → index → value, matching the `((setter proc) arg ... value)` expansion), generalized `set!` produced by a `syntax-rules` macro, deep compositions (`cadddr`, `cddddr`, `caddr`), and a regression check that ordinary `(set! x v)` on an identifier still works.
- Spot-checked the setter table: the 32 entries follow the mechanical `cXXr` rule correctly (e.g. `caaar` → `(set-car! (caar p) v)`, `cdddar` → `(set-cdr! (cddar p) v)`).

The implementation is correct and well-tested. The notes below are **minor / non-blocking** observations, not defects.

### Ready for merge
From a code-correctness standpoint this is **ready to merge** once CI goes green — the required `test` matrix jobs and `benchmark-pr` are still pending; `format`, `wasm`, and CodeRabbit already pass. It also needs an approving review to clear the `BLOCKED` merge state.

tests/scheme/srfi/srfi17.scm
```diff
@@ -19,17 +19,33 @@
...
+;;; --- generalized set! with user-defined setter ---
+(test 42 (let ((s (vector 0)))
+           (define my-ref
+             (getter-with-setter
+               (lambda (v i) (vector-ref v i))
+               (lambda (v i val) (vector-set! v i val))))
+           (set! (my-ref s 0) 42)
+           (my-ref s 0)))
+
+;;; --- cXXr setters ---
```
baijum, line 43: Coverage nit: the `set!` tests here exercise only 2-deep compositions (`caar`, `cadr`, `cdar`). I manually verified the 3- and 4-deep setters (`caddr`, `cadddr`, `cddddr`) work, but adding one deep case to the suite would guard against a future typo in the table — e.g.

```scheme
(test '(1 2 3 99 5) (let ((l (list 1 2 3 4 5))) (set! (cadddr l) 99) l))
```

baijum (commented):
## Follow-up review — commit `43d7fb2b`

All four review notes are resolved, and I re-verified locally on a fresh build (fast-forwarded to `43d7fb2b`):

- `zig build` — ok
- `zig build test` — pass
- `tests/scheme/srfi/srfi17.scm` — **17 pass, 0 fail** (new `cadddr` case)
- `bash tests/scheme/run-all.sh` — **1807 pass, 0 fail**

### Resolution of each note
| Note | Status |
|------|--------|
| `ir.zig` — hardcoded global `setter` / cryptic unimported error | ✅ Comment added documenting the `(srfi 17)` dependency + LLVM-backend fallthrough |
| `compiler_lambda.zig` — legacy-path duplication | ✅ Documented as a defensive fallback (IR path handles it first) |
| `compiler_lambda.zig` — 16-arg cap error type | ✅ `InvalidSyntax` → `InternalLimit` |
| `srfi17.scm` — deep `cXXr` coverage | ✅ Added `(set! (cadddr l) 99)` |

### On the extra GC hardening
Beyond what I flagged, the legacy `compileSet` now wraps its S-expression construction in `no_collect += 1` and roots `desugared` before `compileExprViaIR`, mirroring `compileDefineValues` (issue #1010). I traced it: nothing allocates between the `no_collect` scope exit and `pushRoot(&desugared)`, and `pushRoot` doesn't allocate — so the intermediate pairs can't be collected. Correct, and more robust than relying on allocator-argument auto-rooting. 👍

### Ready for merge
The change is **correct, well-tested, and ready to merge** once the remaining CI matrix goes green (`format`, `wasm`, `riscv64-test`, CodeRabbit already pass on this commit; the other `test` jobs and `benchmark-pr` are still running). No further changes needed from my side.

PR #1351, opened 2026-07-09
Fix SRFI-143 fxcopy-bit to accept boolean bit argument

### LINKED ISSUE #1323: [SRFI-143] fxcopy-bit rejects booleans: uses (= bit 0) numeric protocol instead of the boolean bit argument
**Category:** SRFI (`lib/srfi/143.sld`)
**Sibling of:** #1233 (SRFI-151 `copy-bit`), fixed in #1316. Same bug class, different SRFI.

**Spec reference:** SRFI-143 defines its bitwise operations as "the same as the corresponding operations of SRFI 151, except that they operate on and return fixnums." SRFI-151 `copy-bit`:
> `(copy-bit index i boolean)` — Returns an integer the same as `i` except in the `index`-th bit, which is 1 if `boolean` is `#t` and 0 if `boolean` is `#f`.

So `fxcopy-bit`'s third argument must be a **boolean**, not `0`/`1`.

**Severity:** API mismatch — `fxcopy-bit` is unusable with spec-conforming arguments.

**Root cause:** `lib/srfi/143.sld:101` tests `(= bit 0)`, treating the bit argument as a number. Passing a boolean throws a type error; it only "works" with the non-conforming `0`/`1` numeric protocol.

```scheme
(define (fxcopy-bit index x bit)
  (if (= bit 0)                                     ; <-- boolean rejected here
      (fxand x (fxnot (arithmetic-shift 1 index)))
      (fxior x (arithmetic-shift 1 index))))
```

```scheme
(import (scheme base) (srfi 143))
(fxcopy-bit 2 0 #t)   ;; expected 4   actual: type error ((= bit 0) on a boolean)
(fxcopy-bit 2 5 #f)   ;; expected 1   actual: type error
(fxcopy-bit 0 0 1)    ;; returns 1 — non-conforming numeric protocol
```

**Suggested fix** (mirrors #1316's `copy-bit`):
```scheme
(define (fxcopy-bit index x bit)
  (if bit
      (fxior x (arithmetic-shift 1 index))
      (fxand x (fxnot (arithmetic-shift 1 index)))))
```

**Tests:** `fxcopy-bit` currently has **no** coverage in `tests/scheme/srfi/srfi143.scm`. The fix should add boolean-argument cases there.

**Found by:** Follow-up during review of #1316.

baijum:
## Summary

- Fix `fxcopy-bit` in `lib/srfi/143.sld` to accept boolean `#t`/`#f` as the bit argument per SRFI-143/151 spec, instead of numeric `0`/`1`
- Replace `(= bit 0)` with `(if bit ...)` — same fix pattern as #1316 (SRFI-151 `copy-bit`)
- Add 6 regression tests for `fxcopy-bit` with boolean arguments

Closes #1323

## Test plan

- [x] `fxcopy-bit` with `#t` sets the specified bit: `(fxcopy-bit 2 0 #t) → 4`
- [x] `fxcopy-bit` with `#f` clears the specified bit: `(fxcopy-bit 2 5 #f) → 1`
- [x] All 71 SRFI-143 tests pass (`zig build run -- tests/scheme/srfi/srfi143.scm`)

### DIFF
--- a/lib/srfi/143.sld
+++ b/lib/srfi/143.sld
@@ -99,9 +99,9 @@
       (fxior (fxand mask n0) (fxand (fxnot mask) n1)))
 
     (define (fxcopy-bit index x bit)
-      (if (= bit 0)
-          (fxand x (fxnot (arithmetic-shift 1 index)))
-          (fxior x (arithmetic-shift 1 index))))
+      (if bit
+          (fxior x (arithmetic-shift 1 index))
+          (fxand x (fxnot (arithmetic-shift 1 index)))))
 
     (define (fxfirst-set-bit x)
       (if (= x 0) -1
--- a/tests/scheme/srfi/srfi143.scm
+++ b/tests/scheme/srfi/srfi143.scm
@@ -86,4 +86,11 @@
 (test 0 (fxfirst-set-bit 1))
 (test 2 (fxbit-field 13 1 3))
 
+(test 4 (fxcopy-bit 2 0 #t))
+(test 1 (fxcopy-bit 2 5 #f))
+(test 0 (fxcopy-bit 0 0 #f))
+(test 1 (fxcopy-bit 0 0 #t))
+(test 5 (fxcopy-bit 0 5 #t))
+(test 4 (fxcopy-bit 0 5 #f))
+
 (test-end "srfi-143")


baijum:
## Review — ✅ Ready to merge

Verified by applying the change, rebuilding, and running the suite.

**Correctness** — The old `(= bit 0)` threw a type error on the spec-mandated boolean argument. The new `(if bit ...)` matches SRFI-143's definition (bitwise ops follow SRFI-151 semantics) and is now identical in shape to the `copy-bit` fix in #1316. 👍

**Tests**
- Full SRFI-143 suite: **71 pass, 0 fail**.
- Issue reproductions confirmed: `(fxcopy-bit 2 0 #t) → 4`, `(fxcopy-bit 2 5 #f) → 1`, `(fxcopy-bit 0 5 #t) → 5`, `(fxcopy-bit 0 5 #f) → 4`.
- Good coverage: both set (`#t`) and clear (`#f`) paths, on already-set and already-clear bits, at index 0. Closes the coverage gap noted in #1323.

**Bug class fully addressed** — `fxcopy-bit` is the only procedure in `143.sld` taking a boolean bit argument, so there are no sibling procedures left with the same numeric-protocol bug.

**Note (non-blocking)** — This intentionally drops the old non-conforming numeric protocol. Since `0` is truthy in Scheme, `(fxcopy-bit i x 0)` now *sets* the bit rather than clearing it. That's spec-correct (only `#t`/`#f` are valid) and consistent with #1316.

LGTM — no changes requested.

PR #1352, opened 2026-07-09
Add SRFI-174 timespec-hash, timespec->inexact, inexact->timespec

### LINKED ISSUE #1235: [SRFI-174] timespec-hash, timespec->inexact, inexact->timespec not exported
**Category:** SRFI (lib/srfi/174.sld)
**Spec reference:** SRFI-174 defines nine procedures: timespec, timespec?, timespec-seconds, timespec-nanoseconds, inexact->timespec, timespec->inexact, timespec=?, timespec<?, timespec-hash.
**Severity:** missing API

```scheme
(import (scheme base) (srfi 174))
(timespec->inexact (timespec 1 500000000))   ;; expected 1.5   actual undefined variable
(inexact->timespec 1.5)                       ;; undefined variable
(timespec-hash (timespec 1 2))                ;; undefined variable
```

**Root cause:** lib/srfi/174.sld implements only the record, accessors, and comparisons. (It additionally exports timespec>?, timespec<=?, timespec>=?, which are not in the SRFI — harmless extras.)

**Conforms:** construction/accessors, timespec=?, timespec<? with nanosecond tie-breaking and negative seconds.

**Disabled tests:** tests/scheme/srfi/srfi174.scm (marked ;; FAIL: #this)

baijum:
Closes #1235

## Summary
- Add three missing SRFI-174 procedures: `timespec->inexact`, `inexact->timespec`, `timespec-hash`
- `inexact->timespec` uses `floor` (not `truncate`) so nanoseconds stays non-negative, consistent with POSIX convention and the existing comparison operators
- Uncomment and expand tests (28 pass)

## Test plan
- [x] `zig build run -- tests/scheme/srfi/srfi174.scm` — 28/28 pass
- [x] Manual verification of issue reproduction (all three procedures work)
- [x] Negative-seconds round-trip: `(inexact->timespec -1.25)` → `(timespec -2 750000000)` ✓

### DIFF
--- a/lib/srfi/174.sld
+++ b/lib/srfi/174.sld
@@ -1,7 +1,8 @@
 (define-library (srfi 174)
   (import (scheme base))
   (export timespec timespec? timespec-seconds timespec-nanoseconds
-          timespec=? timespec<? timespec>? timespec<=? timespec>=?)
+          timespec=? timespec<? timespec>? timespec<=? timespec>=?
+          timespec->inexact inexact->timespec timespec-hash)
   (begin
     (define-record-type <timespec>
       (timespec seconds nanoseconds)
@@ -20,4 +21,18 @@
 
     (define (timespec>? a b) (timespec<? b a))
     (define (timespec<=? a b) (not (timespec>? a b)))
-    (define (timespec>=? a b) (not (timespec<? a b)))))
+    (define (timespec>=? a b) (not (timespec<? a b)))
+
+    (define (timespec->inexact ts)
+      (+ (inexact (timespec-seconds ts))
+         (/ (inexact (timespec-nanoseconds ts)) 1e9)))
+
+    (define (inexact->timespec x)
+      (let* ((s (exact (floor x)))
+             (ns (exact (truncate (* (- x s) 1e9)))))
+        (timespec s ns)))
+
+    (define (timespec-hash ts)
+      (let ((s (timespec-seconds ts))
+            (ns (timespec-nanoseconds ts)))
+        (abs (+ (* s 1000000007) ns))))))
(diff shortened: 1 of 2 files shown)


baijum (commented):

Checked out the branch, rebuilt, and ran the suite: **28/28 pass** (`zig build run -- tests/scheme/srfi/srfi174.scm`). The change is fully self-contained — no other file references these three procedures. All three conform to [SRFI-174](https://srfi.schemers.org/srfi-174/srfi-174.html).

**Correctness verified**

- `inexact->timespec` uses `floor` for seconds, which is the key detail: it keeps nanoseconds in the required `[0, 999999999]` range even for negative inputs, matching the existing `timespec<?`/`timespec=?` semantics. `(inexact->timespec -1.25)` → `(timespec -2 750000000)` as expected.
- I probed the floating-point boundaries where the nanosecond field is most at risk of overflowing to `1000000000` (`0.9999999999999999`, `1.9999999999999998`, `-1e-10`, etc.). The invariant `0 ≤ ns ≤ 999999999` held in every case. Worth noting the `truncate` (vs `round`) choice for the fractional part is not just spec-permitted — it's actually the *safer* choice here, since `round` could push a near-integer fraction up to exactly `1e9`.
- `timespec-hash` is a pure function of `(seconds, nanoseconds)` wrapped in `abs`, so it returns a non-negative exact integer and is guaranteed equal for `timespec=?`-equal values — satisfying the hash contract. Verified it stays non-negative and exact for negative-seconds inputs.

**Tests** — good coverage: positive/zero/negative round-trips, the `inexact->timespec`/`timespec->inexact` asymmetry, and hash determinism/non-negativity.

**Minor, non-blocking observations** (no change requested):

- The `timespec` constructor still doesn't validate that nanoseconds fall in `[0, 999999999)` (the SRFI says it's an error otherwise). This is pre-existing and out of scope for this PR — fine to leave.
- The extra `timespec>?`/`timespec<=?`/`timespec>=?` exports aren't in the SRFI, but those predate this PR and are harmless (already noted in #1235).

Ready for merge.

PR #1354, opened 2026-07-09
Fix SRFI-1 take-right/drop-right to accept dotted lists (#1166)

### LINKED ISSUE #1166: [SRFI-1] take-right/drop-right reject dotted lists (spec requires proper-or-dotted; take/drop already conform)
**Category:** SRFI-1
**Spec reference:** SRFI-1, take-right/drop-right: "flist may be any finite list, either proper or dotted." Spec examples: `(take-right '(1 2 3 . d) 2) => (2 3 . d)`, `(drop-right '(1 2 3 . d) 2) => (1)`, `(take-right '(1 2 3 . d) 0) => d`, `(drop-right '(1 2 3 . d) 0) => (1 2 3)`.
**Severity:** wrong-result

**Minimal reproduction:**
```scheme
(import (srfi 1))
(take-right '(1 2 3 . d) 2)  ;; expected (2 3 . d), actual: type error "expected pair"
(drop-right '(1 2 3 . d) 2)  ;; expected (1),        actual: type error
```
`take` and `drop` on the same inputs already conform (`(drop '(1 2 3 . d) 3) => d`), so the dotted-list support exists in the family — only the -right variants reject.

**Root cause:** `src/primitives_srfi1.zig` `takeRightFn`/`dropRightFn` (~lines 1112-1148) count the list length with a loop that requires every cdr to be a pair (`if (!types.isPair(current)) return primitives.typeError(...)`), so any dotted tail raises before the real work starts. The SRFI-1 reference implementation uses a lead/lag two-pointer walk that never inspects the tail's type; adopting it fixes both procedures and also gives the required `(take-right lis 0) => tail` behavior (note: the result of take-right on a dotted list with k=0 is the raw tail, which need not be a list).

**Verified:** fresh ReleaseSafe build at 5b5a4e8; deterministic.
**Disabled test:** `tests/scheme/audit/primitives_srfi1-audit.scm` (marked `;; FAIL: #this`)

**Found by:** Systematic audit, Phase 2.6 (#1137)

baijum:
## Summary

- Replace two-pass length-count approach in `take-right`/`drop-right` with a lead/lag two-pointer walk that naturally handles dotted tails
- Uncomment and expand the disabled audit tests for dotted-list cases
- Add dedicated regression test in `tests/scheme/smoke/`

Closes #1166

## Test plan

- [x] All four SRFI-1 spec examples pass: `(take-right '(1 2 3 . d) 2)`, `(take-right '(1 2 3 . d) 0)`, `(drop-right '(1 2 3 . d) 2)`, `(drop-right '(1 2 3 . d) 0)`
- [x] Proper-list behavior unchanged (take-right/drop-right on proper lists)
- [x] Edge cases: k=0, k=length
- [x] SRFI-1 audit suite: 109 pass, 0 fail
- [x] `zig build test` passes

### DIFF
--- a/src/primitives_srfi1.zig
+++ b/src/primitives_srfi1.zig
@@ -1108,55 +1108,52 @@ fn carCdrFn(args: []const Value) PrimitiveError!Value {
     return gc.allocMultipleValues(&vals) catch return PrimitiveError.OutOfMemory;
 }
 
-// (take-right list k)
+// (take-right list k) — flist may be proper or dotted
 fn takeRightFn(args: []const Value) PrimitiveError!Value {
     if (!types.isFixnum(args[1])) return primitives.typeError("take-right", "integer", args[1]);
     const k = types.toFixnum(args[1]);
     if (k < 0) return primitives.typeError("take-right", "non-negative integer", args[1]);
-    // Find list length
-    var len: i64 = 0;
-    var current = args[0];
-    while (current != types.NIL) {
-        if (!types.isPair(current)) return primitives.typeError("take-right", "pair", current);
-        current = types.cdr(current);
-        len += 1;
-    }
-    // drop (len - k) elements
-    const to_drop = len - k;
-    if (to_drop < 0) return primitives.typeError("take-right", "valid index (k <= length)", args[1]);
-    current = args[0];
+
+    // Advance lead pointer k steps
+    var lead = args[0];
     var i: i64 = 0;
-    while (i < to_drop) : (i += 1) {
-        current = types.cdr(current);
+    while (i < k) : (i += 1) {
+        if (!types.isPair(lead)) return primitives.typeError("take-right", "valid index (k <= length)", args[1]);
+        lead = types.cdr(lead);
     }
-    return current;
+
+    // Advance both until lead hits end (nil or dotted tail)
+    var lag = args[0];
+    while (types.isPair(lead)) {
+        lead = types.cdr(lead);
+        lag = types.cdr(lag);
+    }
+    return lag;
 }
 
-// (drop-right list k)
+// (drop-right list k) — flist may be proper or dotted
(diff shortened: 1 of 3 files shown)


baijum:
## Review — LGTM ✅ Ready to merge

Verified locally on the PR branch (fresh ReleaseSafe build, `.sbc` caches cleared): `zig build` + `zig build test` pass, the SRFI-1 audit reports **109 pass / 0 fail**, and the new smoke test passes (10 expected passes).

### What's good

- **Faithful to the SRFI-1 reference.** The lead/lag two-pointer walk is exactly what the spec's reference implementation does, and it's the fix #1166 recommended. It never inspects the tail's type, so dotted lists flow through naturally.
- **Simpler than what it replaces** — drops the separate length-counting pass in both procedures.
- **Good test coverage.** All four spec examples are covered (`take-right` k=2/k=0, `drop-right` k=2/k=0), plus proper-list, k=0, and k=length cases. The smoke test correctly grabs the runner *before* `test-end` and exits non-zero on failure.
- Out-of-range `k` still errors (`(take-right '(1 2) 5)` → error), and `drop-right` keeps the correct GC pattern (`buildList` roots the accumulator; the collected cars stay reachable through the still-rooted `args[0]`).

### One behavior note (not a blocker)

For a **non-list** first argument with `k=0`, behavior changes:

```scheme
(take-right 5 0)   ; was: type error "expected pair"  → now: 5
(drop-right 5 0)   ; was: type error                  → now: ()
```

This is a contract violation either way (the argument isn't a list), and the new results are precisely what the SRFI-1 reference implementation yields (`(drop lis 0)` returns `lis` unchanged, then the walk sees a non-pair and returns it / `'()`). So this is arguably *more* conformant, not a regression — just worth being aware of. No change requested.

### Verdict

Correct, well-tested, and matches the reference. **Ready to merge** once the pending `format` / `benchmark-pr` CI jobs go green (CodeRabbit already passed).

PR #1355, opened 2026-07-09
Fix SRFI-37 short optional-arg dropping trailing characters

### LINKED ISSUE #1350: [SRFI-37] args-fold: optional-arg short options ignore trailing characters
**Category:** SRFI
**Spec reference:** SRFI-37: when an option has `optional-arg? #t`, a short option like `-oX` should pass `"X"` as the argument to the processor.
**Severity:** wrong-result (optional-arg short options silently drop the argument)

**Minimal reproduction:**
```scheme
(import (scheme base) (scheme write) (srfi 37))

(define opt-o
  (option '(#\o) #f #t  ; optional-arg? = #t
          (lambda (opt name arg seed) (list name arg))))

(display (args-fold '("-oX") (list opt-o)
                    (lambda (o n a s) s) (lambda (op s) s) '()))
;; expected: (#\o "X")
;; actual:   (#\o #f)  — "X" is scanned as a further short option instead
```

**Root cause (`lib/srfi/37.sld`):**
The short-option `sloop` branch only special-cases `option-required-arg?`. When `optional-arg?` is true, the code falls through to the no-arg path, passing `#f` and then scanning the remaining characters as more short options.

The fix should check `option-optional-arg?` and, when trailing characters exist after the option letter, pass that remainder as the argument.

Long optional args via `--opt=X` already work correctly.

**Found during:** review of #1343 (PR for #1211)

baijum:
## Summary
- Fix `args-fold` short-option parsing to check `option-optional-arg?` when trailing characters exist after the option letter, passing them as the argument instead of scanning them as further short options
- Add 6 regression tests covering short optional-arg inline, absent, with operand, and combined with other short options

Closes #1350

## Test plan
- [x] New regression tests pass (`tests/scheme/srfi/srfi37.scm` — 35 tests total)
- [x] Full test suite passes (1807 pass, 0 fail)

### DIFF
--- a/lib/srfi/37.sld
+++ b/lib/srfi/37.sld
@@ -97,10 +97,16 @@
                                    (call-with-values
                                      (lambda () (apply (option-processor opt) opt ch a seeds))
                                      (lambda new-seeds (loop r new-seeds))))
-                                 (call-with-values
-                                   (lambda () (apply (option-processor opt) opt ch #f seeds))
-                                   (lambda new-seeds
-                                     (sloop (+ i 1) new-seeds rest))))
+                                 (if (and (option-optional-arg? opt)
+                                          (< (+ i 1) (string-length arg)))
+                                     (let ((a (substring arg (+ i 1) (string-length arg))))
+                                       (call-with-values
+                                         (lambda () (apply (option-processor opt) opt ch a seeds))
+                                         (lambda new-seeds (loop rest new-seeds))))
+                                     (call-with-values
+                                       (lambda () (apply (option-processor opt) opt ch #f seeds))
+                                       (lambda new-seeds
+                                         (sloop (+ i 1) new-seeds rest)))))
                              (call-with-values
                                (lambda () (apply unrecognized-proc
                                                  (%make-option (list ch) #f #f #f) ch #f seeds))
--- a/tests/scheme/srfi/srfi37.scm
+++ b/tests/scheme/srfi/srfi37.scm
@@ -112,6 +112,43 @@
 ;;; --- empty args ---
 (test-equal "empty args" 0 (args-fold '() opts unrec count-operand 0))
 
+;;; --- short optional-arg (#1350: trailing chars dropped) ---
+(define opt-p
+  (option '(#\p) #f #t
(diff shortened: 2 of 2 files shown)


baijum:
## Review — LGTM ✅

Reviewed the fix and verified it locally (built `kaappi`, applied the PR, ran the tests).

**Correctness.** The change makes the short-option `sloop` mirror the SRFI-37 reference implementation exactly: when an `optional-arg?` option has trailing characters in the same token, the remainder becomes the argument and scanning advances to the next token (`loop rest`); when there are no trailing characters, `#f` is passed and — importantly — the **next token is *not* consumed** (`sloop (+ i 1) …`). That's the correct SRFI-37 behavior for short optional args (only inline `-oX` supplies a value; the next argv element never does), and it parallels both the existing `required-arg?` branch and the long-option `--opt=X` path.

Spot-checked edge cases, all correct:
- `-oX` → `(o X)` (the #1350 repro)
- `-oval` → `(o "val")` (multi-char remainder)
- `-o` → `(o #f)`, and `-o file` keeps `file` as an operand (no over-consumption)
- `-vpX` / `-vp` (plain flag + optional-arg in a cluster) resolve correctly

**Tests.** Solid regression coverage. I reverted `37.sld` to `main` with the new tests in place: the 3 inline-value tests (`inline`, `multi-char`, `combined … with value`) genuinely **FAIL** without the fix and pass with it — real coverage, not tautological. The other 3 (`absent`, `absent + operand`, `combined … no value`) are protective guards that were already green, confirming the fix doesn't over-consume. All 35 SRFI-37 tests pass, and the full `tests/scheme/srfi/` group is clean. No trailing whitespace; 2-space indentation is consistent with the surrounding code.

**Optional (non-blocking).** Agree with CodeRabbit's minor note: the `(< (+ i 1) (string-length arg))` / `(substring arg (+ i 1) …)` computation is now duplicated across the required-arg and optional-arg branches. A tiny `trailing-arg` helper (or shared `let`) would DRY it up, but the current form reads fine and I wouldn't block on it.

**Merge readiness.** Ready on its merits. Currently `BLOCKED` only because the `format` and `benchmark-pr` checks are still pending — good to merge once CI goes green.

PR #1356, opened 2026-07-09
Fix random-real open interval per SRFI-27 (#1195)

### LINKED ISSUE #1195: [SRFI-27] random-real / %rs-next-real use a [0,1) generator; spec requires strictly 0 < x < 1
**Category:** SRFI
**Spec reference:** SRFI-27, Specification: "(random-real) — The next number 0 < x < 1 obtained from default-random-source." Both bounds are exclusive. Same wording applies to generators from random-source-make-reals ("random real numbers 0 < x < 1").
**Severity:** edge-case (found by code inspection; not practically reproducible)

**Analysis:** `randomRealFn` and `rsNextRealFn` (src/primitives_random.zig) return `r.float(f64)`, and Zig's `std.Random.float` is documented "Return a floating point value evenly distributed in the range [0, 1)" — 0.0 is a possible return. With the exponentially-biased-exponent algorithm in Zig 0.16, returning exactly 0.0 requires a run of ~1022 zero bits (probability ≈ 2^-1022), so it cannot be demonstrated by a test, but it is a genuine deviation from the spec's open interval.

**Minimal reproduction:** none feasible; code-inspection finding. The audit test asserts `(> r 0)` over draws, which passes.

**Suggested fix:** retry loop — `var x = r.float(f64); while (x == 0) x = r.float(f64);` in both call sites (or a shared helper).

**Verified:** code inspected at 96ce73b (src/primitives_random.zig + Zig 0.16 std/Random.zig)
**Source location:** src/primitives_random.zig (`randomRealFn`, `rsNextRealFn`)

baijum:
## Summary
- SRFI-27 requires `random-real` to return values in the open interval (0, 1), but Zig's `std.Random.float(f64)` returns [0, 1) — 0.0 is theoretically possible (~2^-1022 probability)
- Add `openUnitReal()` retry-loop helper, used by both `randomRealFn` and `rsNextRealFn`
- Tighten coverage test assertion from `>=` to `>`

Closes #1195

## Test plan
- [x] `zig build test` passes
- [x] `primitives_random-audit.scm` passes (54/54)
- [x] Coverage test updated to assert `(> r 0.0)` instead of `(>= r 0.0)`

### DIFF
--- a/src/primitives_random.zig
+++ b/src/primitives_random.zig
@@ -66,8 +66,7 @@ fn randomRealFn(args: []const Value) PrimitiveError!Value {
     ensureDefaultRS();
     const vm = vm_mod.vm_instance orelse return PrimitiveError.OutOfMemory;
     const rs = try getRS("random-real", vm.default_random_source);
-    const r = rs.prng.random();
-    return types.makeFlonum(r.float(f64));
+    return types.makeFlonum(openUnitReal(rs));
 }
 
 fn defaultRandomSourceFn(args: []const Value) PrimitiveError!Value {
@@ -159,8 +158,14 @@ fn rsNextIntFn(args: []const Value) PrimitiveError!Value {
 // (%rs-next-real rs) — used by random-source-make-reals closure
 fn rsNextRealFn(args: []const Value) PrimitiveError!Value {
     const rs = try getRS("%rs-next-real", args[0]);
-    const r = rs.prng.random();
-    return types.makeFlonum(r.float(f64));
+    return types.makeFlonum(openUnitReal(rs));
+}
+
+// SRFI-27 requires 0 < x < 1 (open interval); Zig's float(f64) returns [0, 1).
+fn openUnitReal(rs: *types.RandomSource) f64 {
+    var x = rs.prng.random().float(f64);
+    while (x == 0.0) x = rs.prng.random().float(f64);
+    return x;
 }
 
 fn randomBelow(proc: []const u8, rs: *types.RandomSource, bound: Value) PrimitiveError!Value {
--- a/tests/scheme/coverage/random-coverage.scm
+++ b/tests/scheme/coverage/random-coverage.scm
@@ -25,7 +25,7 @@
 
 ;;; ---- random-real ----
 (let ((r (random-real)))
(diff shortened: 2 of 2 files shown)


baijum:
## Review — LGTM ✅

The fix is correct, minimal, and well-scoped. `std.Random.float(f64)` returns `[0, 1)`, so the upper bound `1.0` is already excluded; the `openUnitReal()` retry loop excludes `0.0`, giving the SRFI-27 open interval `(0, 1)`. Both call sites (`randomRealFn`, `rsNextRealFn`) share the helper. 👍

**Verified locally** (checked out `fix/1195`):
- `zig build` + `zig build test` — pass
- `tests/scheme/coverage/random-coverage.scm` — 20 passed, 0 failed
- `tests/scheme/audit/primitives_random-audit.scm` — 54 pass, 0 fail

### Non-blocking notes

1. **Nit (`src/primitives_random.zig:165`)** — `rs.prng.random()` is re-created on every iteration. Since `float(f64)` is what advances state, you could hoist the interface once:
   ```zig
   fn openUnitReal(rs: *types.RandomSource) f64 {
       const r = rs.prng.random();
       var x = r.float(f64);
       while (x == 0.0) x = r.float(f64);
       return x;
   }
   ```
   Purely cosmetic — the loop realistically never iterates (~2⁻¹⁰²² per the issue).

2. **Regression test** — CLAUDE.md asks every bug fix to ship a test that fails without the fix. That isn't achievable here: `0.0` can't be forced from the PRNG, so the tightened `(> r 0.0)` assertion passes with or without the change. The issue acknowledges this is a code-inspection finding with no feasible reproduction, so tightening the assertion to document intent is the right call — just flagging that it doesn't actually exercise the retry path.

3. **Adjacent, out of scope** — `random-source-make-reals` in `lib/srfi/27.sld:16-22` validates the optional `unit` argument but then ignores it (always calls `(%rs-next-real rs)`). SRFI-27 says `unit` should set the resolution of the generated reals. Pre-existing gap, unrelated to #1195 — worth a separate issue if not already tracked.

### Verdict
Ready to merge once CI clears (`format` and `benchmark-pr` were pending at review time; CodeRabbit passed). The two non-blocking notes above are optional.

PR #1357, opened 2026-07-09
Implement SRFI-61 general cond clause (generator guard => receiver)

### LINKED ISSUE #1206: [SRFI-61] Library is empty and the general cond clause (generator guard => receiver) is not implemented
**Category:** SRFI
**Spec reference:** SRFI-61: clause grammar "(<generator> <guard> => <receiver>)"; semantics: "<generator> is evaluated. It may return arbitrarily many values. <Guard> is applied to an argument list containing the values in order that <generator> returned. If <guard> returns a true value for that argument list, <receiver> is applied with an equivalent argument list. If <guard> returns a false value, however, the clause is abandoned and the next one is tried."
**Severity:** missing-feature

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 61))
(cond ((values 1 2) (lambda (a b) #t) => (lambda (a b) (+ a b)))
      (else 'no))
;; expected: 3   actual: error (base cond does not recognize the three-element clause)
```
The spec's own example (`port->char-list` using `((read-char port) char? => ...)`) fails the same way.

**Analysis:** lib/srfi/61.sld is `(export)` with an empty body — importing it provides nothing, and the built-in `cond` (src/compiler_conditionals.zig) only implements the R7RS clause forms. SRFI-61 requires extending cond with the guard-clause variant (typically via a shadowing `cond` macro exported from the library, using `call-with-values`).

**Source location:** lib/srfi/61.sld (empty), src/compiler_conditionals.zig (`cond`)
**Disabled test:** tests/scheme/srfi/srfi61.scm (marked ;; FAIL: #this)

baijum:
## Summary

- Replace the empty `lib/srfi/61.sld` stub with a `syntax-rules` macro that shadows the built-in `cond`, adding the `(generator guard => receiver)` clause form per the SRFI-61 spec
- The generator may return multiple values via `call-with-values`; the guard is applied to them, and if true the receiver gets the same argument list
- All standard R7RS `cond` clause forms are preserved by the macro

Closes #1206

## Test plan

- [x] 10 SRFI-61 tests pass (multi-value generators, guard success/failure, single-value guards, mixed clauses, chained SRFI-61 clauses)
- [x] Full test suite: 1807 pass, 0 fail

### DIFF
--- a/lib/srfi/61.sld
+++ b/lib/srfi/61.sld
@@ -1,7 +1,25 @@
 ;;; SRFI 61 — A more general cond clause
-;;; Kaappi's cond already supports => clauses from R7RS.
-;;; This SRFI adds (test => receiver) with a guard procedure.
+;;; Extends cond with (generator guard => receiver) clauses.
 (define-library (srfi 61)
   (import (scheme base))
-  (export)
-  (begin))
+  (export cond)
+  (begin
+    (define-syntax cond
+      (syntax-rules (=> else)
+        ((cond (else expr1 expr2 ...))
+         (begin expr1 expr2 ...))
+        ((cond (generator guard => receiver) rest ...)
+         (call-with-values (lambda () generator)
+           (lambda args
+             (if (apply guard args)
+                 (apply receiver args)
+                 (cond rest ...)))))
+        ((cond (test => proc) rest ...)
+         (let ((t test))
+           (if t (proc t) (cond rest ...))))
+        ((cond (test) rest ...)
+         (or test (cond rest ...)))
+        ((cond (test body1 body2 ...) rest ...)
+         (if test (begin body1 body2 ...) (cond rest ...)))
+        ((cond)
+         (if #f #f))))))
--- a/tests/scheme/srfi/srfi61.scm
+++ b/tests/scheme/srfi/srfi61.scm
@@ -1,5 +1,4 @@
-;; SRFI-61 (a more general cond clause) conformance tests — audit Phase 3d
-;; The library is currently an empty stub (#1206); base cond still works.
+;; SRFI-61 (a more general cond clause) conformance tests
 ;; Run directly: zig-out/bin/kaappi tests/scheme/srfi/srfi61.scm
 
 (import (scheme base) (srfi 61) (chibi test))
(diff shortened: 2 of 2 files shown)


baijum:
## Review — ready to merge ✅

Reviewed the shadowing `cond` macro against the SRFI-61 spec and the R7RS `cond` grammar. This is a clean, faithful implementation and matches the approach recommended in #1206 ("a shadowing `cond` macro exported from the library, using `call-with-values`").

### Verified locally (fresh ReleaseSafe build off the PR branch)
- **Build** clean; `tests/scheme/srfi/srfi61.scm` → **10 pass, 0 fail**.
- **SRFI-61 semantics** — generator's values are captured with `call-with-values`, `guard` and `receiver` are each `apply`-ed to the *same* argument list ("equivalent argument list"), and a false guard abandons the clause and tries the next. Tail position of the receiver is preserved. Confirmed with extra probes: multi-value generators, **zero-value** generators (`(values)` + nullary guard/receiver), guard-fails-then-next-SRFI-61-clause, and guard actually consuming the values.
- **R7RS clause forms preserved** — `(else …)`, `(test => proc)` (receiver gets the test value), `(test)` test-only, `(test body …)`, empty `(cond)`, and deep `else` reached after several recursions (hygiene holds through the recursive expansion).
- **No interference with other derived forms** — in a module that imports `(srfi 61)`, `case`, `when`, `guard` still compile and run correctly (they're separate built-in forms, not routed through the user-level `cond` macro).
- **Import order** — the SRFI-61 `cond` wins whether `(srfi 61)` is imported before or after `(scheme base)`, so the enhanced `cond` is reliably in effect.
- **No regressions** — smoke and full SRFI test groups still green.

### Minor (non-blocking)
- `(import (scheme base) (srfi 61))` imports two `cond` bindings. Kaappi resolves this gracefully (SRFI-61 wins), and it's consistent with the existing convention in `tests/scheme/srfi/srfi1.scm` (which imports base + SRFI-1 without `except` despite the `map`/`fold`/`member` collisions). Strict R7RS would use `(import (except (scheme base) cond) (srfi 61))`, but matching the repo's established style here is fine.

Correct, well-tested, and isolated to the lazily-loaded `(srfi 61)` library. LGTM.

PR #1359, opened 2026-07-09
Fix SRFI-133 vector-skip/vector-skip-right multi-vector form (#1171)

### LINKED ISSUE #1171: [SRFI-133] vector-skip / vector-skip-right: multi-vector form rejected (registered with exact arity 2)
**Category:** SRFI
**Spec reference:** SRFI-133, Searching: `(vector-skip pred? vec1 vec2 ...)` — "vector-skip ... Applies pred? to the elements of the vectors in order and returns the index of the first elements for which pred? returns #f" (multi-vector form, same as vector-index/vector-skip family). `vector-skip-right` likewise takes `vec1 vec2 ...`.
**Severity:** missing-feature

**Minimal reproduction:**
```scheme
(import (srfi 133))
(vector-skip = #(1 2 3 4 5) #(1 2 -3 4))
;; expected: 2   actual: error: 'vector-skip': expected 2 arguments, got 3
(vector-skip-right = #(1 2 3 4 5) #(1 2 -3 4 5))
;; expected: 2   actual: error: 'vector-skip-right': expected 2 arguments, got 3
```

**Expected:** 2 (index of the first mismatch) in both cases
**Actual:** arity error — both procedures are registered with `.arity = .{ .exact = 2 }`, and the implementations only look at `args[1]`

Single-vector behavior is correct. The sibling procedures `vector-index`, `vector-index-right`, `vector-count`, `vector-any`, `vector-every` already implement the multi-vector loop pattern to copy from.

**Verified:** fresh ReleaseSafe build at 96ce73b; gc-stress unchanged
**Source location:** src/primitives_vector.zig — registration lines for `vector-skip`/`vector-skip-right` (specs table) and `vectorSkipFn` / `vectorSkipRightFn`

baijum:
## Summary
- Change `vector-skip` and `vector-skip-right` arity from `.exact = 2` to `.variadic = 2` to accept multiple vectors per SRFI-133
- Implement multi-vector loop pattern (matching existing `vector-index` / `vector-index-right`)
- Enable previously disabled audit tests for the multi-vector form

Closes #1171

## Test plan
- [x] `zig build test` passes
- [x] `primitives_vector-audit.scm` passes (182/182, including 2 newly enabled multi-vector tests)
- [x] Manual verification of exact reproduction cases from the issue

### DIFF
--- a/src/primitives_vector.zig
+++ b/src/primitives_vector.zig
@@ -30,8 +30,8 @@ pub const specs = [_]primitives.PrimSpec{
     .{ .name = "vector-every", .func = &vectorEveryFn, .arity = .{ .variadic = 2 }, .libs = LS.initOne(.srfi_133) },
     .{ .name = "vector-index", .func = &vectorIndexFn, .arity = .{ .variadic = 2 }, .libs = LS.initOne(.srfi_133) },
     .{ .name = "vector-index-right", .func = &vectorIndexRightFn, .arity = .{ .variadic = 2 }, .libs = LS.initOne(.srfi_133) },
-    .{ .name = "vector-skip", .func = &vectorSkipFn, .arity = .{ .exact = 2 }, .libs = LS.initOne(.srfi_133) },
-    .{ .name = "vector-skip-right", .func = &vectorSkipRightFn, .arity = .{ .exact = 2 }, .libs = LS.initOne(.srfi_133) },
+    .{ .name = "vector-skip", .func = &vectorSkipFn, .arity = .{ .variadic = 2 }, .libs = LS.initOne(.srfi_133) },
+    .{ .name = "vector-skip-right", .func = &vectorSkipRightFn, .arity = .{ .variadic = 2 }, .libs = LS.initOne(.srfi_133) },
     .{ .name = "vector-swap!", .func = &vectorSwapFn, .arity = .{ .exact = 3 }, .libs = LS.initOne(.srfi_133) },
     .{ .name = "vector-reverse!", .func = &vectorReverseBangFn, .arity = .{ .variadic = 1 }, .libs = LS.initOne(.srfi_133) },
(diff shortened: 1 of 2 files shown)


baijum (commented):
## Review — ready to merge ✅

Verified locally (fresh ReleaseSafe build at `557465fe`): `zig build test` passes, `primitives_vector-audit.scm` reports **182 pass, 0 fail**, and the exact reproductions from #1171 now return `2` in both cases.

### What's good
- **Correct and minimal.** The multi-vector loop, the shared-scratch-buffer allocation (`stack_buf[256]` with a `gc.allocator.alloc` fallback past 256 vectors), and the shortest-vector stopping rule are copied faithfully from the already-shipping siblings (`vector-index` / `vector-index-right`), so the family is now consistent.
- **Semantics check out.** I exercised beyond the two enabled tests — forward stop-at-shortest (`(vector-skip = #(1 1 9 9) #(1 1))` → `#f`, mismatch beyond the shorter vector correctly invisible), a mismatch within the shortest, 3-vector calls, and single-vector fast paths. All match SRFI-133.
- **GC-safe.** `call_args_buf` only holds `Value` copies drawn from the argument vectors, which stay rooted for the duration of the call (the collector is non-moving), and `gc.allocator.alloc` is a raw scratch allocation that can't trigger a collection — so no rooting is required here. Consistent with the siblings.
- **Regression coverage.** Enabling the two previously-commented audit tests (the issue's exact repros) satisfies the "every bug fix needs a regression test" rule.

### Minor notes (non-blocking)
1. **`-right` + unequal lengths.** SRFI-133 says the `*-right` searchers require all vectors to be the same length ("it is an error" otherwise). This impl gracefully clamps to `min_len` — a fine, safe choice for the unspecified case, and it matches the existing `vector-index-right`, so no change needed.
2. **Duplication.** The scratch-buffer + multi-vector loop scaffold is now copy-pasted across six functions (`count`/`any`/`every`/`index`/`index-right`/`skip`/`skip-right`). Matching the established style is the right call for this PR; a future refactor could extract a shared helper, but that's out of scope here.

No blockers — good to merge.

PR #1360, opened 2026-07-09
Fix SRFI-13 wrong-typed optional args silently ignored (#1159)

### LINKED ISSUE #1159: [SRFI-13] wrong-typed optional args silently ignored: string-pad char, string-unfold base/make-final
**Category:** SRFI-13
**Spec reference:** SRFI-13 — string-pad/string-pad-right: "char defaults to #\space" (a character); string-unfold/string-unfold-right: "base: the initial string. Default: the empty string." Positional optional arguments of the wrong type are errors, not absent.
**Severity:** edge-case (silently ignored optional arguments)

**Minimal reproduction:**
```scheme
(import (srfi 13))
(string-pad "x" 5 "*")   ;; expected: error ("*" is not a char)
                          ;; actual: "    x" — pad argument silently ignored, space used
(string-unfold (lambda (s) (> s 2)) (lambda (s) #\a) (lambda (s) (+ s 1)) 0 #\Z)
                          ;; expected: error (#\Z is not a string base)
                          ;; actual: "aaa" — base silently ignored
```

**Root cause:** `src/primitives_string_ext.zig` — the optional-argument guards use `and`-typed checks that treat a wrong-typed argument as absent instead of raising:
- `stringPadFn`/`stringPadRightFn`: `if (args.len > 2 and types.isChar(args[2]))` falls back to `' '`
- `stringUnfoldFn`/`stringUnfoldRightFn`: `if (args.len > 4 and types.isString(args[4]))` skips the base; a make-final returning a non-string is likewise silently dropped

This is bug pattern #4 from the audit skill (ignored optional arguments): a typo'd call site keeps "working" with wrong output instead of failing.

**Disabled test:** `tests/scheme/audit/primitives_string_ext-audit.scm` (marked `;; FAIL: #this`)

**Found by:** Systematic audit, Phase 2.2 (#1137)

baijum:
## Summary

- `string-pad` / `string-pad-right`: a non-char pad argument (e.g. `"*"` instead of `#\*`) was silently ignored, defaulting to space. Now raises a type error.
- `string-unfold` / `string-unfold-right`: a non-string `base` argument (e.g. `#\Z`) was silently skipped. Now raises a type error.
- `string-unfold` / `string-unfold-right`: a `make-final` callback returning a non-string value was silently dropped. Now raises a type error.

All six sites followed the same pattern: `if (args.len > N and types.isXxx(args[N]))` treated wrong-typed arguments as absent instead of erroring.

Closes #1159

## Test plan

- [x] Uncommented the two `FAIL: #1159` markers in `primitives_string_ext-audit.scm`
- [x] Added 4 new tests covering `string-pad-right`, `string-unfold-right` base, and both `make-final` return value checks
- [x] All 95 audit tests pass
- [x] Full Scheme test suite: 1808 pass, 0 fail
- [x] Zig unit tests pass

### DIFF
--- a/src/primitives_string_ext.zig
+++ b/src/primitives_string_ext.zig
@@ -501,7 +501,10 @@ fn stringPadFn(args: []const Value) PrimitiveError!Value {
     if (tlv < 0) return primitives.typeError("string-pad", "non-negative integer", args[1]);
     const target_len: usize = @intCast(tlv);
     var pad_buf: [4]u8 = undefined;
-    const pad_cp: u21 = if (args.len > 2 and types.isChar(args[2])) types.toChar(args[2]) else ' ';
+    const pad_cp: u21 = if (args.len > 2) blk: {
+        if (!types.isChar(args[2])) return primitives.typeError("string-pad", "char", args[2]);
+        break :blk types.toChar(args[2]);
+    } else ' ';
     const pad_len = std.unicode.utf8Encode(pad_cp, &pad_buf) catch
         return primitives.typeError("string-pad", "valid character", args[2]);
     const range = try parseStartEnd(full_data, args, 3);
@@ -529,7 +532,10 @@ fn stringPadRightFn(args: []const Value) PrimitiveError!Value {
     if (tlv < 0) return primitives.typeError("string-pad-right", "non-negative integer", args[1]);
     const target_len: usize = @intCast(tlv);
     var pad_buf: [4]u8 = undefined;
-    const pad_cp: u21 = if (args.len > 2 and types.isChar(args[2])) types.toChar(args[2]) else ' ';
+    const pad_cp: u21 = if (args.len > 2) blk: {
+        if (!types.isChar(args[2])) return primitives.typeError("string-pad-right", "char", args[2]);
+        break :blk types.toChar(args[2]);
+    } else ' ';
     const pad_len = std.unicode.utf8Encode(pad_cp, &pad_buf) catch
         return primitives.typeError("string-pad-right", "valid character", args[2]);
     const range = try parseStartEnd(full_data, args, 3);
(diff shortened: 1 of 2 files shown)


baijum (commented):

Reviewed the change locally on a fresh ReleaseSafe build. The fix is correct, tightly scoped, and well-tested.

### What the change does
All six sites replace the `if (args.len > N and types.isXxx(args[N]))` guard — which treated a *wrong-typed* optional arg as *absent* — with an explicit `args.len > N` check that raises a `typeError` when the type is wrong. This is exactly bug pattern #4 from the audit (silently ignored optional arguments), and matches the SRFI-13 semantics cited in #1159 (positional optionals of the wrong type are errors, not defaults).

### Verification
- `zig build` — clean
- `zig build test` (Zig unit tests) — pass
- Audit suite `primitives_string_ext-audit.scm` — **95 pass, 0 fail**
- Full Scheme suite `run-all.sh` — **1808 pass, 0 fail** (413 files + 1395 R7RS), matches the PR description
- Reproductions from #1159 now raise:
  - `(string-pad "x" 5 "*")` → error
  - `(string-pad-right "x" 5 "*")` → error
  - `(string-unfold … 0 #\Z)` / `string-unfold-right` → error
  - `make-final` returning `42` → error (both directions)
- Regression checks on valid inputs still correct: `(string-pad "x" 5)` → `"    x"`, `(string-pad "x" 3 #\*)` → `"**x"`, `(string-unfold … "Z" (lambda (s) "!"))` → `"Zaaa!"`, and the base/make-final positive paths (audit lines 111–119) unchanged.
- Grepped the file — no remaining `args.len > N and types.is…` guards of this pattern; all sites in scope are covered.

### Minor, non-blocking
In `stringUnfoldRightFn`, the `base` type check lands at the *end* of the function (line ~920), after the unfold loop and `make-final` have already run. So for a wrong-typed `base`, the error surfaces only after `p`/`f`/`g` (arbitrary Scheme procedures, possibly with side effects) execute — whereas `stringUnfoldFn` validates `base` up front (line 855). Behaviorally both raise, so this is fine as-is; if you want the two to fail symmetrically/fast, you could add an early `if (args.len > 4 and !types.isString(args[4])) return typeError(…)` at the top of `stringUnfoldRightFn` while keeping the append at the end. Purely optional polish.

Nothing blocking — good to merge.

PR #1362, opened 2026-07-09
Fix random-source-make-reals to honor the unit argument (#1194)

### LINKED ISSUE #1194: [SRFI-27] random-source-make-reals validates the unit argument but then ignores it
**Category:** SRFI
**Spec reference:** SRFI-27, Specification: "(random-source-make-reals s unit) -> rand ... The optional parameter unit determines the type of numbers being produced by rand and the quantization of the output. Unit must be a number such that 0 < unit < 1. The numbers created by rand are of the same numerical type as unit and the potential output values are spaced by at most unit. One can imagine rand to create numbers as x*unit where x is a random integer in {1, ..., floor(1/unit)-1}."
**Severity:** wrong-result

**Minimal reproduction:**
```scheme
(define rand (random-source-make-reals (make-random-source) 1/10))
(exact? (rand))  ;; expected: #t (same numerical type as the exact unit 1/10)
                 ;; actual: #f — always a default-precision flonum
```

**Analysis:** `lib/srfi/27.sld` checks `0 < unit < 1` and raises otherwise, but the returned generator is unconditionally `(lambda () (%rs-next-real rs))` — the unit's type and quantization are discarded. With a flonum unit the current behavior happens to conform ("the actual resolution of rand can be much higher than unit"), but with an exact rational unit the results must be exact.

A conforming implementation for exact unit: `(* unit (+ 1 (%rs-next-int rs (- (floor (/ 1 unit)) 1))))` (i.e. x*unit for random x in {1, ..., floor(1/unit)-1}).

**Source location:** lib/srfi/27.sld (`random-source-make-reals`)

**Found by:** Systematic audit, Phase 2.15 (skill bug pattern #4: ignored optional arguments)

baijum:
## Summary

- `random-source-make-reals` validated `0 < unit < 1` but discarded the value, always returning a default-precision flonum
- With an exact rational unit (e.g. `1/10`), SRFI-27 requires exact results quantized as `x*unit` for random `x` in `{1, ..., floor(1/unit)-1}`
- Inexact units and the no-unit case still use the default `%rs-next-real` path

Closes #1194

## Test plan

- [x] Enabled previously disabled audit test (`primitives_random-audit.scm`)
- [x] Added `tests/scheme/srfi/srfi-27-unit.scm` (127 assertions): exact units `1/10` and `1/4`, inexact unit, no-unit default, invalid-unit errors
- [x] `zig build test` passes
- [x] Existing random audit suite passes (55/55)


lib/srfi/27.sld
```diff
@@ -14,11 +14,14 @@
...
-                           (> (car rest) 0)
-                           (< (car rest) 1))))
-        (error "random-source-make-reals: unit must satisfy 0 < unit < 1" (car rest)))
-      (lambda () (%rs-next-real rs)))
+      (if (pair? rest)
+          (let ((unit (car rest)))
+            (if (not (and (real? unit) (> unit 0) (< unit 1)))
+                (error "random-source-make-reals: unit must satisfy 0 < unit < 1" unit)
+                (if (exact? unit)
+                    (let ((n (exact (- (floor (/ 1 unit)) 1))))
```
baijum, line 22: Edge case: when `1/2 < unit < 1` (exact), `floor(1/unit) = 1` so `n = 0`, and the generator built on the next line calls `(%rs-next-int rs 0)`, which raises `type error … expected positive integer, got 0` at call time. The unit passed validation (`0 < unit < 1`), so the failure is surprising. Consider handling `n = 0` (return `unit` — the sole multiple of `unit` in `(0,1)`) or rejecting such units at construction, and adding a test for a unit in `(1/2, 1)`. Non-blocking — the SRFI-27 reference impl has the same behavior.

baijum (commented):
## Review — looks good, one edge case worth a note

The core fix is correct and does exactly what #1194 asks. I verified it against a fresh ReleaseSafe build (running the PR's `lib/srfi/27.sld` via `--lib-path`):

- **Exact units produce exact quantized results.** With `unit = 1/10`, sampling 5000 times the results are always exact multiples of `1/10`, and the observed `x` covers the full `{1, …, 9}` = `{1, …, floor(1/unit)-1}` range (min `1/10`, max `9/10`, all strictly inside `(0,1)`). The `(+ 1 (%rs-next-int rs n))` shift is right: `%rs-next-int` returns `{0, …, n-1}`, so `+1` yields `{1, …, n}` with `n = floor(1/unit)-1`. ✓
- **Inexact units and the no-unit case** keep the default `%rs-next-real` path — conforming, since the spec allows resolution higher than `unit`. ✓
- **Validation** (`0 < unit < 1`) is preserved. ✓
- **Tests pass:** `tests/scheme/srfi/srfi-27-unit.scm` → 127 expected passes; re-enabled `primitives_random-audit.scm` → 55 pass / 0 fail. Both suites are picked up by `run-all.sh` (`srfi/*.scm`, `audit/*.scm`), so they run in CI. There's only one copy of `27.sld` in the tree and SRFI `.sld` files aren't embedded at build time, so the single-file change is complete.

### One edge case (non-blocking): exact units in `(1/2, 1)`

For an exact `unit` with `1/2 < unit < 1` (e.g. `2/3`, `3/4`), `floor(1/unit) = 1` so `n = 0`, and `%rs-next-int` rejects `0`. The constructor accepts these units (they satisfy `0 < unit < 1`) but the *returned generator* then throws a confusing internal error at call time:

```scheme
(define g (random-source-make-reals (make-random-source) 3/4))  ; ok
(g)  ; => error: type error in '%rs-next-int': expected positive integer, got 0
```

This mirrors the SRFI-27 reference implementation (the spec's `x ∈ {1, …, floor(1/unit)-1}` is empty here), so it's arguably spec-conformant and low priority. But since the constructor advertises these units as valid, it'd be nicer to either handle `n = 0` (the only multiple of `unit` in `(0,1)` is `unit` itself) or reject such units up front — and add a test for that range (the new suite covers `1/10` and `1/4`, both with `n ≥ 3`, so this range is untested). Your call whether to address here or leave as a follow-up.

### Nit

`(exact (- (floor (/ 1 unit)) 1))` — for an exact `unit`, `(/ 1 unit)`, `floor`, and `-` are already exact, so the `exact` wrapper is a harmless no-op.

**Verdict: ready to merge.** The edge case above is pre-existing spec behavior and doesn't block; consider it a follow-up.

PR #1363, opened 2026-07-09
Export owner/unchanged and group/unchanged constants from SRFI-170 (#1163)

### LINKED ISSUE #1163: [SRFI-170] owner/unchanged and group/unchanged constants are not exported (set-file-owner cannot express partial chown)
**Category:** SRFI-170
**Spec reference:** SRFI-170 — "(set-file-owner fname uid gid)": "If the uid argument is the constant owner/unchanged, the owner is not changed; if the gid argument is the constant group/unchanged, the group is not changed."
**Severity:** missing-feature (edge-case)

**Minimal reproduction:**
```scheme
(import (srfi 170))
owner/unchanged   ;; => error: undefined variable
group/unchanged   ;; => error: undefined variable
```

`set-file-owner` itself is implemented with the correct 3-argument chown signature, but the two spec-required sentinel constants are not exported, so portable code cannot change only the owner or only the group (POSIX expresses "unchanged" as (uid_t)-1 / (gid_t)-1, which the raw integer interface can't accept since negative ids are rejected).

**Source location:** `src/primitives_filesystem.zig` (set-file-owner registration; constants absent from the srfi_18/srfi_170 export set in `src/library.zig`).
**Verified:** fresh ReleaseSafe build at 83718b6.

baijum:
## Summary
- Define `owner/unchanged` and `group/unchanged` as fixnum(-1) globals, exported from `(srfi 170)`
- These SRFI-170 constants let `set-file-owner` express partial chown (change only owner or only group)
- Enable the previously disabled audit test for these constants

## Test plan
- [x] New `tests/scheme/srfi/srfi-170-chown-constants.scm` — verifies constants are -1 and `set-file-owner` accepts them
- [x] Re-enabled audit test in `primitives_filesystem-audit.scm` passes
- [x] `zig build test` passes
- [x] Full audit suite passes

Closes #1163

### DIFF
--- a/src/library.zig
+++ b/src/library.zig
@@ -112,6 +112,8 @@ const extra_exports = [_]ExtraExport{
     .{ .name = "current-input-port", .libs = LS.initMany(&.{ .scheme_base, .scheme_r5rs }) },
     .{ .name = "current-output-port", .libs = LS.initMany(&.{ .scheme_base, .scheme_r5rs }) },
     .{ .name = "current-error-port", .libs = LS.initOne(.scheme_base) },
+    .{ .name = "owner/unchanged", .libs = LS.initOne(.srfi_170) },
+    .{ .name = "group/unchanged", .libs = LS.initOne(.srfi_170) },
 };
 
 fn addExportsForLib(library: *Library, lib: Lib, globals: *std.StringHashMap(Value), sandboxed: bool) !void {
--- a/src/primitives.zig
+++ b/src/primitives.zig
@@ -210,8 +210,10 @@ pub fn registerAll(vm: *vm_mod.VM) !void {
         if (!is_wasm or spec.wasm)
             try reg(vm, spec.name, spec.func, spec.arity);
     }
-    try vm.defineGlobal("owner/unchanged", types.makeFixnum(-1));
-    try vm.defineGlobal("group/unchanged", types.makeFixnum(-1));
+    if (comptime !is_wasm) {
+        try vm.defineGlobal("owner/unchanged", types.makeFixnum(-1));
+        try vm.defineGlobal("group/unchanged", types.makeFixnum(-1));
+    }
 }
 
 pub fn registerSandboxed(vm: *vm_mod.VM) !void {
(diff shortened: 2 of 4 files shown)


baijum:
## Review — ready to merge ✅

Reviewed the change against the `set-file-owner` implementation and verified it locally (fresh ReleaseSafe build). This is a clean, correctly-scoped fix.

**Verified:**
- `validateUid` (`src/primitives_filesystem.zig:41`) already maps `-1` → `(uid_t)-1` (`0xFFFFFFFF`), the POSIX "leave unchanged" sentinel — so exporting the two constants is genuinely all that was missing.
- `(import (srfi 170))` now resolves `owner/unchanged` / `group/unchanged` to `-1`; they're real library exports (work through `(prefix (srfi 170) …)` renaming too, not just leaked top-level globals).
- New `tests/scheme/srfi/srfi-170-chown-constants.scm` passes (5/5) and is a genuine regression test — it references the constants directly, so it errors/fails without the fix. Auto-discovered by `run-all.sh` via the `srfi/*.scm` glob, and uses the correct SRFI-64 "grab runner before `test-end`" pattern.
- Re-enabled audit test passes (audit suite went 66 → 67 pass, 0 fail).
- `zig build test` green; all PR CI checks green including the `wasm` job.

**Non-blocking nits (optional):**

1. **WASM tidiness** — `src/primitives.zig:213-214`: the two `defineGlobal` calls are unconditional, whereas the primitive registrations just above are guarded by `if (!is_wasm or spec.wasm)`. `set-file-owner` has `.wasm = false` and `(srfi 170)` isn't registered under WASM, so in a WASM build these two constants exist as top-level globals with nothing to consume them. Harmless (immediate fixnums, can't be misused without `set-file-owner`), and CI's wasm job is green — but wrapping them in `if (comptime !is_wasm)` would match the surrounding pattern.

2. **Docs follow-up** — the end-user procedure reference (kaappi.github.io) doesn't mention these constants. Worth a small follow-up there, separate repo.

Neither blocks merge. LGTM.

PR #1365, opened 2026-07-09
Skip stress tests in Debug CI builds

### LINKED ISSUE #1364: CI: Debug build times out on 3 Scheme test files
The Debug build CI job (`ubuntu-latest, Debug, 30`) consistently times out on three Scheme test files that exceed the 60s per-file limit. Debug builds are ~500x slower than ReleaseSafe for allocation- and continuation-heavy workloads, so these stress-heavy tests don't finish in time.

**Timed-out files:**
- `tests/scheme/compliance/r7rs-tail-procedures-gaps.scm`
- `tests/scheme/compliance/r7rs-thin-forms-gaps.scm`
- `tests/scheme/continuations/callcc-bench.scm`

**Example run:** https://github.com/kaappi/kaappi/actions/runs/28999481627/job/86056491512

All other CI configurations (ReleaseSafe, ReleaseFast, macOS, ARM, RISC-V, WASM) pass consistently.

**Options:**
1. Increase the per-file timeout for Debug builds (e.g. 120s or 180s)
2. Skip these files in Debug CI (they're stress tests, not correctness tests)
3. Reduce iteration counts in the tests when a `KAAPPI_DEBUG` env var or similar is detected

baijum:
## Summary

Fixes #1364.

- Add `KAAPPI_SKIP` env var to `run-all.sh` — space-separated list of filenames to skip
- Add `KAAPPI_TIMEOUT` env var to override the default 60s per-file timeout
- CI Debug matrix entry sets `KAAPPI_SKIP` to skip the 3 stress files that exceed 60s under Debug builds (`callcc-bench.scm`, `r7rs-tail-procedures-gaps.scm`, `r7rs-thin-forms-gaps.scm`)
- All Release builds (ReleaseSafe, ReleaseFast, macOS, ARM) still run every test

## Test plan

- [x] Full `run-all.sh` passes locally with `KAAPPI_SKIP` set (410 pass, 3 skipped, 0 fail)
- [x] Skip logic unit-tested: matches correct files, ignores non-matching, no-ops when `KAAPPI_SKIP` is empty
- [ ] CI Debug build no longer times out on these 3 files
- [ ] CI Release builds still run all files (no `KAAPPI_SKIP` set)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,6 @@ jobs:
           - os: ubuntu-latest
             optimize: Debug
             timeout: 30
-            skip_stress: "callcc-bench.scm r7rs-tail-procedures-gaps.scm r7rs-thin-forms-gaps.scm"
           - os: ubuntu-latest
             optimize: ReleaseFast
 
@@ -72,7 +71,7 @@ jobs:
       - name: Scheme suites
         run: bash tests/scheme/run-all.sh
         env:
-          KAAPPI_TEST_SKIP: ${{ matrix.skip_stress || '' }}
+          KAAPPI_TEST_SKIP: ${{ matrix.optimize == 'Debug' && 'callcc-bench.scm r7rs-tail-procedures-gaps.scm r7rs-thin-forms-gaps.scm' || '' }}
 
       - name: Sandbox escape tests
         run: bash tests/scheme/sandbox/sandbox-escape.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
       - name: Scheme suites
         run: bash tests/scheme/run-all.sh
         env:
-          KAAPPI_SKIP: ${{ matrix.skip_stress || '' }}
+          KAAPPI_TEST_SKIP: ${{ matrix.skip_stress || '' }}
 
       - name: Sandbox escape tests
         run: bash tests/scheme/sandbox/sandbox-escape.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,7 @@ jobs:
           - os: ubuntu-latest
             optimize: Debug
             timeout: 30
+            skip_stress: "callcc-bench.scm r7rs-tail-procedures-gaps.scm r7rs-thin-forms-gaps.scm"
           - os: ubuntu-latest
             optimize: ReleaseFast
 
@@ -70,6 +71,8 @@ jobs:
 
       - name: Scheme suites
         run: bash tests/scheme/run-all.sh
+        env:
+          KAAPPI_SKIP: ${{ matrix.skip_stress || '' }}
 
       - name: Sandbox escape tests
         run: bash tests/scheme/sandbox/sandbox-escape.sh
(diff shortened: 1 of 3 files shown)


baijum:
## Review — Skip stress tests in Debug CI builds

Reviewed the diff against `main`. The change is small, focused, and correct. It solves #1364 the right way: Debug CI skips the three slow files, while **every Release build (ReleaseSafe, ReleaseFast, macOS, ARM, RISC-V) still runs all of them**, so conformance coverage is preserved where it matters. I verified the `should_skip` logic in isolation — exact basename match (with a path prefix), non-match, partial-name (no false positives), and the empty-`KAAPPI_TEST_SKIP` no-op all behave correctly. `callcc-bench.scm` is a 400k-iteration perf benchmark whose only assertion is "completes without error", so skipping it under Debug loses no real coverage.

A few non-blocking notes:

**1. (minor, cosmetic) The `skip_stress` matrix field bloats the Debug job's display name.** Because it's added as a new `include` key, GitHub appends it to the job name:

```
test (ubuntu-latest, Debug, 30, callcc-bench.scm r7rs-tail-procedures-gaps.scm r7rs-thin-forms-ga...
```

Optional alternative that keeps the name clean — compute the env var inline instead of via a matrix field:

```yaml
- name: Scheme suites
  run: bash tests/scheme/run-all.sh
  env:
    KAAPPI_TEST_SKIP: ${{ matrix.optimize == 'Debug' && 'callcc-bench.scm r7rs-tail-procedures-gaps.scm r7rs-thin-forms-gaps.scm' || '' }}
```

**2. (minor, docs) Consider documenting the new knobs.** `tests/scheme/CLAUDE.md` still says "all suites (60s timeout per file)" — a one-line mention of `KAAPPI_TEST_SKIP` / `KAAPPI_TEST_TIMEOUT` in the Running section would make them discoverable.

**3. (note, not a defect) `should_skip` only applies to `run_suite` (`.scm` files), not `run_shell_suite`.** Fine for this PR since all three targets are `.scm`; worth remembering if shell-test skipping is ever needed.

**4. (nice) Renaming to the `KAAPPI_TEST_*` prefix** is a good call — avoids collision with runtime vars like `KAAPPI_LIB_DIR` / `KAAPPI_HOME`.

**Merge readiness:** the code is correct and ready. The one true gate is the remaining unchecked item in the test plan — the Debug CI job going green (it's the actual proof the timeout is gone). It's currently pending on this run; once it passes, this is good to merge. None of the notes above are blockers.

baijum:
Follow-up — commit `3d5a8427` addresses all the review feedback:

- **Job-name bloat → fixed.** `skip_stress` matrix field replaced with the inline `${{ matrix.optimize == 'Debug' && '…' || '' }}` conditional; the Debug job name is now cleanly `test (ubuntu-latest, Debug, 30)`.
- **Docs → fixed.** `tests/scheme/CLAUDE.md` now documents both `KAAPPI_TEST_TIMEOUT` and `KAAPPI_TEST_SKIP` in the Running section.
- **Bonus:** adding `matched=1` in the skip branch is a good catch — a suite whose files are all skipped no longer misreports `(no tests matched)`.

Code LGTM. The only remaining gate is the Debug CI job going green on this run (the actual proof the timeout is gone) — it's pending now. Once it passes, this is ready to merge.

PR #1366, opened 2026-07-09
Accept native procedures in make-thread and spawn (#1155)

### LINKED ISSUE #1155: [SRFI-18] make-thread rejects native procedures as thunks (procedure? => #t but make-thread => type error)
**Category:** SRFI-18
**Spec reference:** SRFI-18, make-thread: "(make-thread thunk [name]) — Returns a new thread whose execution consists of calling thunk" — thunk is any procedure of zero arguments.
**Severity:** edge-case

**Minimal reproduction:**
```scheme
(import (srfi 18))
(thread-join! (thread-start! (make-thread list)))
;; expected: '() — list is a zero-arg-callable procedure
;; actual:   type error: make-thread expects procedure
```

**Root cause:** `src/primitives_srfi18.zig` `makeThreadFn` requires `types.isClosure(thunk)`, rejecting built-in (native) procedures, which satisfy `procedure?`. The frame setup that follows assumes a Closure, so the fix needs either a wrapper closure for native fns or a separate native-call path in `threadEntryFn`.

baijum:
## Summary

- `make-thread` (SRFI-18) and `spawn` (fibers) now accept any procedure, not just closures
- For fibers, non-closure procedures are wrapped in a bytecode trampoline closure (`get_upvalue → call → return`) since the fiber scheduler requires bytecode frames
- Fiber is rooted across trampoline allocation to prevent GC collection (verified under `-Dgc-stress=true`)

Fixes #1155

## Test plan

- [x] New `tests/scheme/srfi/srfi18-native-thunk.scm` — make-thread with native procs, closures, and non-procedure rejection
- [x] New `tests/scheme/srfi/fiber-native-thunk.scm` — spawn with native procs, closures, and non-procedure rejection
- [x] Re-enabled disabled audit test in `primitives_srfi18-audit.scm`
- [x] Updated `primitives_fiber-audit.scm` to expect spawn to accept native procs
- [x] Full test suite: 1810 pass, 0 fail
- [x] Unit tests pass
- [x] GC stress test (`-Dgc-stress=true`) passes for both make-thread and spawn with native procs


baijum:
## Review — Accept native procedures in make-thread and spawn (#1155)

Reviewed the diff and independently rebuilt/ran the branch. **This looks correct and ready to merge.** 👍

### What I verified locally (branch `fix/1155`)
- `zig build` + `zig build test` — clean
- New tests: `srfi18-native-thunk.scm` (4 pass), `fiber-native-thunk.scm` (3 pass)
- Re-enabled audit tests: `primitives_srfi18-audit` (77 pass), `primitives_fiber-audit` (30 pass)
- **GC stress** (`-Dgc-stress=true`) — both native-thunk tests pass
- All 25 fiber/thread/srfi18 concurrency tests pass
- Extra edge cases: `(spawn +)` → `0`; `(spawn car)` → clean `'car': expected 1 arguments, got 0` (then join deadlocks — identical to the existing closure-that-errors path, so no regression); 8 concurrent `(spawn list)` fibers all return `()`.

### Correctness notes (all good)
- **GC safety of the trampoline is solid.** During `wrapInTrampoline`, the native proc stays reachable via the rooted `fiber` (`fiber.thunk` was set by `allocFiber`, and `markFiberState` marks it); `allocClosure(func)` auto-roots `func`; `proc_root` is pushed across `allocClosure`; both write barriers (`closure.upvalues[0]`, `fiber.thunk`) are correct. Closure marking traces `func` + `upvalues`, so the hand-built `Function` can't be swept out from under the live closure.
- **Trampoline bytecode operand widths match the dispatcher** exactly (`get_upvalue`=4, `call`=3, `return`=2 operand bytes per `fixed_operand_bytes` in `vm_dispatch.zig`), and frame `base=0` puts the loaded proc in register 1 as intended.
- **make-thread's one-line predicate change is sufficient** — `threadEntryFn` → `callWithArgs` already dispatches every procedure kind, and `deepCopy` rejects uncopyable procs (continuations) gracefully.
- Broadening to `isProcedure` (which also admits continuation/parameter/ffi) is handled without crashes on both paths.

### Minor, optional (non-blocking)
1. **`src/fiber.zig` — trampoline `Function` is re-allocated on every non-closure `spawn`.** The bytecode is identical every time; only the closure's upvalue differs. If you ever want to trim allocation, the `Function` could be built once and cached on the scheduler, allocating just the per-spawn `Closure`. `spawn` isn't a hot path, so this is purely optional.
2. **`src/fiber.zig` `wrapInTrampoline` — the hand-assembled bytecode uses raw magic bytes.** It's well-commented and covered by a regression test, but it's silently coupled to the operand encoding in `vm_dispatch.zig`. A one-line comment noting "operand widths must match `fixed_operand_bytes`" would help a future maintainer who changes the ISA.

Neither point blocks merge.

PR #1367, opened 2026-07-09
Fix random-source-make-reals to honor the unit argument (#1194)

### LINKED ISSUE #1194: [SRFI-27] random-source-make-reals validates the unit argument but then ignores it

baijum:
## Summary

- `random-source-make-reals` validated `0 < unit < 1` but discarded the value, always returning a default-precision flonum
- With an exact rational unit (e.g. `1/10`), SRFI-27 requires exact results quantized as `x*unit` for random `x` in `{1, ..., floor(1/unit)-1}`
- Correctly handles non-integer reciprocals (e.g. `unit=3/5`): uses `floor(1/unit)` as the bound instead of `floor(1/unit)-1`, which would undershoot or crash
- Inexact units and the no-unit case still use the default `%rs-next-real` path

Closes #1194
Supersedes #1362

## Test plan

- [x] Enabled previously disabled audit test (`primitives_random-audit.scm`)
- [x] Added `tests/scheme/srfi/srfi-27-unit.scm` (227 assertions): exact units `1/10`, `1/4`, `3/10`, `3/5`, `2/3`; inexact unit; no-unit default; invalid-unit errors
- [x] `zig build test` passes
- [x] Existing random audit suite passes (55/55)


baijum (commented):
## Re-review of #1367 (supersedes #1362) — LGTM ✅

This addresses every point from my review of #1362, and the new bound computation is actually more correct than my suggestion. Verified against a fresh ReleaseSafe build running this PR's `lib/srfi/27.sld`.

### The fix
```scheme
(let* ((recip (/ 1 unit))
       (n (if (integer? recip) (- recip 1) (floor recip))))
  (lambda () (* unit (+ 1 (%rs-next-int rs n)))))
```
The generator emits `unit * x` for `x ∈ {1, …, n}`. To stay in the open interval `(0,1)` you need the largest `x` with `x*unit < 1`:
- **Integer reciprocal** (e.g. `unit=1/10`, `recip=10`): `n = recip-1 = 9`. Using `recip` itself would emit `10*1/10 = 1.0`, so the `-1` is required. ✓
- **Non-integer reciprocal** (e.g. `unit=3/10`, `recip=10/3`): `n = floor(recip) = 3`, since `floor(recip) < recip` strictly. ✓

Since `unit < 1 ⇒ recip > 1`, both branches give `n ≥ 1`, so **the `n = 0` crash from #1362 is gone**.

### Verified
Exhaustive sampling (3000 draws each) over `1/10, 1/4, 3/10, 3/5, 2/3, 1/2, 9/10, 99/100, 4/7, 7/12`:
- No crashes — including the previously-crashing `(1/2, 1)` units (`3/5`, `2/3`, `9/10`, …). ✓
- Every output exact, in `(0,1)`, and a multiple of `unit`. ✓
- Observed range **exactly** equals the full set of valid multiples in `(0,1)` (`covers-all? #t` for all). ✓
- PR tests pass: `srfi-27-unit.scm` → 227 expected passes; re-enabled `primitives_random-audit.scm` → 55/0. Both run in CI via `run-all.sh`.

### Note on the spec deviation (intentional and correct)
Using `floor(recip)` rather than the spec's literal `floor(1/unit)-1` is the right call. For non-integer reciprocals the literal formula *undershoots* — e.g. `unit=3/10` would yield only `{3/10, 6/10}` and drop `9/10`, even though `9/10` is a valid in-range multiple spaced by `unit`. The spec's `{1, …, floor(1/unit)-1}` is prefaced with "One can imagine…" (illustrative, not normative); the normative requirements — same numerical type as `unit`, spacing ≤ `unit`, values in `(0,1)` — are all satisfied, and coverage is complete. Worth a one-line code comment for the next reader, but not required.

**Verdict: ready to merge.**

lib/srfi/27.sld
```diff
@@ -14,11 +14,17 @@
...
-                           (> (car rest) 0)
-                           (< (car rest) 1))))
-        (error "random-source-make-reals: unit must satisfy 0 < unit < 1" (car rest)))
-      (lambda () (%rs-next-real rs)))
+      (if (pair? rest)
+          (let ((unit (car rest)))
+            (if (not (and (real? unit) (> unit 0) (< unit 1)))
+                (error "random-source-make-reals: unit must satisfy 0 < unit < 1" unit)
+                (if (exact? unit)
+                    (let* ((recip (/ 1 unit))
```
baijum, line 23: Optional, for the next reader: a one-line comment explaining why the bound branches on `(integer? recip)` — it documents the deliberate `floor(recip)` vs. `recip - 1` choice (the deviation from the spec's illustrative `floor(1/unit)-1`). Commit the suggestion, or split it across two lines if you prefer to keep the line short:

```suggestion
                    ;; n = max x with x*unit < 1 (ceil(1/unit)-1); subtract 1 only when 1/unit is integral, to exclude the 1.0 endpoint
                    (let* ((recip (/ 1 unit))
```

baijum:
Suggest documenting the quantization choice in `CONFORMANCE.md` too (it can't be an inline suggestion since that file isn't in this PR's diff). The SRFI 27 section already exists at line 37 — appending one sentence to line 39 covers it:

**Current (line 39):**

> **Coverage: 100%** (11 of 11 spec procedures). Full state save/restore via `random-source-state-ref`/`state-set!` (all 4 xoshiro256 state words).

**Proposed — append:**

> For exact `unit`, `random-source-make-reals` quantizes to **every** multiple of `unit` in the open interval `(0,1)` — `x·unit` for `x ∈ {1, …, ceil(1/unit)−1}`. This intentionally extends the spec's illustrative `{1, …, floor(1/unit)−1}` set (introduced with "One can imagine…", i.e. non-normative), which undershoots when `1/unit` is non-integral — it would drop `9/10` for `unit=3/10` and produce an empty range for `unit ∈ (1/2, 1)`.

This is an *implementation choice*, not a spec deviation — all normative SRFI-27 requirements (same numerical type as `unit`, spacing ≤ `unit`, values in `(0,1)`) are satisfied; only the explicitly-illustrative example formula differs. Documenting it here gives anyone cross-referencing the source against the SRFI text a findable answer, and pairs with the inline code comment. Optional, but cheap and in the right home.

PR #1368, opened 2026-07-09
Fix string-trim default criterion to use Unicode whitespace (#826)

### LINKED ISSUE #826: string-trim family: default criterion misses \v, \f and Unicode whitespace that char-whitespace? accepts
## Summary

When called without an explicit criterion, `string-trim`, `string-trim-right`, and `string-trim-both` only strip the four bytes space/`\t`/`\n`/`\r`. SRFI-13 specifies the default criterion as `char-set:whitespace`, which includes vertical tab (U+000B), form feed (U+000C), and all Unicode whitespace (NBSP U+00A0, EM SPACE U+2003, IDEOGRAPHIC SPACE U+3000, …). Kaappi's own `char-whitespace?` already returns `#t` for all of these characters, so the no-criterion trim is inconsistent with the rest of the runtime: `(string-trim s)` and `(string-trim s char-whitespace?)` disagree.

## Location

`src/primitives_string_ext.zig`:

```zig
fn isWhitespace(c: u8) bool {                       // lines 93-95
    return c == ' ' or c == '\t' or c == '\n' or c == '\r';
}
```

Used by the default (no-criterion) fast paths in `stringTrimFn` (lines 189-193), `stringTrimRightFn` (lines 211-215), and `stringTrimBothFn` (lines 234-239). Being byte-based (`u8`), it also cannot ever match multi-byte whitespace codepoints.

## Reproduction

```scheme
(import (scheme base) (scheme write) (scheme char) (srfi 13))
(write (char-whitespace? #\x0B)) (newline)              ; #t
(write (char-whitespace? #\x0C)) (newline)              ; #t
(write (char-whitespace? #\x00A0)) (newline)            ; #t
(write (string-trim "\x0B;\x0C;hi")) (newline)
(write (string-trim-right "hi\x0B;\x0C;")) (newline)
(write (string-trim "\x00A0;hi")) (newline)
(write (string-trim "\x0B;\x0C;hi" char-whitespace?)) (newline)  ; works
```

## Actual vs Expected (Kaappi v0.11.1)

| Expression | Actual | Expected |
|---|---|---|
| `(string-trim "\x0B;\x0C;hi")` | `"\x0b;\x0c;hi"` (unchanged) | `"hi"` |
| `(string-trim-right "hi\x0B;\x0C;")` | `"hi\x0b;\x0c;"` (unchanged) | `"hi"` |
| `(string-trim "\x00A0;hi")` (NBSP) | NBSP retained | `"hi"` |
| `(string-trim "\x0B;\x0C;hi" char-whitespace?)` | `"hi"` | `"hi"` (correct — shows the inconsistency) |

SRFI-13: "string-trim s [criterion start end] … criterion defaults to `char-set:whitespace`" — and `char-set:whitespace` (SRFI-14) contains all characters with the Unicode White_Space property, including U+000B, U+000C, and U+00A0.

## Root cause

The default-criterion fast path scans raw bytes with a hard-coded 4-byte `isWhitespace(u8)` helper instead of decoding codepoints and using the same whitespace classification that backs `char-whitespace?` (Unicode tables in `primitives_char.zig`/`unicode_tables.zig`). The explicit-predicate path (`callPredOrCharset`) already decodes codepoints correctly, so only the no-argument form is affected.

---
*This issue was found by automated code analysis (Claude).*

baijum:
## Summary

- Replace byte-level `isWhitespace(u8)` fast path in `string-trim`, `string-trim-right`, and `string-trim-both` with codepoint-decoding loops that call `isUnicodeWhitespace(u21)` — the same function backing `char-whitespace?`
- Make `isUnicodeWhitespace` pub in `primitives_char.zig` so `primitives_string_ext.zig` can call it
- Remove the now-unused `isWhitespace` helper

Fixes #826

## Test plan

- [x] 15 new regression tests in `tests/scheme/smoke/trim-whitespace.scm` covering VT/FF, NBSP, EM SPACE, IDEOGRAPHIC SPACE, mixed whitespace, and agreement with explicit `char-whitespace?`
- [x] Zig unit tests pass
- [x] Full Scheme suite passes (1813/1813)

### DIFF
--- a/src/primitives_char.zig
+++ b/src/primitives_char.zig
@@ -62,7 +62,7 @@ pub fn isUnicodeCased(cp: u21) bool {
     return unicode.inRanges(&unicode.cased_ranges, cp);
 }
 
-fn isUnicodeWhitespace(cp: u21) bool {
+pub fn isUnicodeWhitespace(cp: u21) bool {
     if (cp <= 127) {
         return switch (cp) {
             0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x20 => true,
--- a/src/primitives_string_ext.zig
+++ b/src/primitives_string_ext.zig
@@ -94,10 +94,6 @@ fn stringSuffixPFn(args: []const Value) PrimitiveError!Value {
     return if (std.mem.endsWith(u8, s2_range.data, s1_range.data)) types.TRUE else types.FALSE;
 }
 
-fn isWhitespace(c: u8) bool {
-    return c == ' ' or c == '\t' or c == '\n' or c == '\r' or c == 0x0B or c == 0x0C;
-}
-
 /// Call a predicate or char-set-contains? with a character.
 /// Handles both procedure arguments and SRFI-14 char-set record arguments.
 fn callPredOrCharset(pred: Value, cp: u21) PrimitiveError!bool {
@@ -169,7 +165,11 @@ fn stringTrimFn(args: []const Value) PrimitiveError!Value {
     const full_data = try getStringSlice("string-trim", args[0]);
     if (args.len <= 1) {
         var start: usize = 0;
-        while (start < full_data.len and isWhitespace(full_data[start])) start += 1;
+        while (start < full_data.len) {
+            const d = decodeForward(full_data, start);
+            if (d.len == 0 or !pchar.isUnicodeWhitespace(d.cp)) break;
+            start += d.len;
+        }
         return gc.allocString(full_data[start..]) catch return PrimitiveError.OutOfMemory;
     }
     const pred = args[1];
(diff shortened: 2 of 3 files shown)


baijum (commented):

Correct, minimal, and well-tested. The default (no-criterion) trim fast path now decodes codepoints and delegates to `isUnicodeWhitespace` — the same classifier behind `char-whitespace?` — so `(string-trim s)` and `(string-trim s char-whitespace?)` finally agree. Nice reuse of the existing `decodeForward`/`findPrevCpStart` helpers, and the backward scan terminates cleanly (`end = cp_start` strictly decreases; `end > start` guard keeps trim-both from crossing the trimmed prefix).

**Verified locally** (branch `fix/826`, Apple Silicon, ReleaseSafe):
- `zig build` ✅
- `tests/scheme/smoke/trim-whitespace.scm` → 15/15 ✅
- `srfi13-ext` (50/0), `srfi13-startend` (46/0), `srfi13-charset` (23/0), `srfi130` (34/0) ✅ — no regressions
- `zig build test` ✅

### One non-blocking note (commit-message accuracy)

The commit/PR narrative says the old fast path was "missing vertical tab, form feed." That was true when #826 was filed (v0.11.1), but the current `main` `isWhitespace` helper already includes `0x0B`/`0x0C`:

```zig
return c == ' ' or c == '\t' or c == '\n' or c == '\r' or c == 0x0B or c == 0x0C;
```

So on `main` today, VT/FF already trim — the genuine fix here is **multi-byte Unicode whitespace** (NBSP, EM SPACE, IDEOGRAPHIC SPACE, U+0085, …). The three VT/FF cases in the new test therefore pass on base too; they're still worth keeping as guards, but they don't demonstrate the bug being fixed. Consider tweaking the commit body so the git history isn't slightly misleading. No code change needed.

### Nit (optional)

In the no-criterion loops the `d.len == 0` guard is dead — inside `while (start < full_data.len)` / `while (end > 0)`, `decodeForward` is always called with `pos < len`, so `len` is never 0. Harmless (mirrors the explicit-pred path), fine to leave.

PR #1369, opened 2026-07-09
Fix yield raising inside with-exception-handler after spawn (#1314)

### LINKED ISSUE #1314: yield raises inside with-exception-handler after spawn
**Minimal reproduction:**
```scheme
(import (scheme base))
(fiber-join (spawn (lambda () 42)))
(with-exception-handler
  (lambda (e) (display "yield raised: ") (display (error-object-message e)) (newline))
  (lambda ()
    (display "before yield\n")
    (yield)
    (display "after yield\n")))
```

**Expected:** yield is a no-op (all fibers completed), prints "before yield" then "after yield".
**Actual:** yield raises `#<error "error">` which the handler catches. Without an exception handler, yield succeeds.

The bug is that `yield` behaves differently depending on whether there's an active `with-exception-handler` on the stack. This affects any use of `yield` inside `guard` (which desugars to `with-exception-handler`).

**Discovered during:** #1196 fix (PR #1311). The `(chibi test)` macro now wraps expressions in `guard`, which means any test using `yield` after `spawn` will hit this.

**Impact:** One fiber audit test (`primitives_fiber-audit.scm` line 63) is currently skipped due to this bug.

baijum:
## Summary

- **yield no-op when nothing to schedule**: `yieldFn` now checks `scheduler.schedule()` instead of just `scheduler != null`, so yield is a no-op when no other fibers are runnable (e.g. after `fiber-join` completes all spawned fibers)
- **with-exception-handler propagates Yielded**: the catch-all in `withExceptionHandlerFn` was converting `VMError.Yielded` into a Scheme exception; it now propagates the signal like `ContinuationInvoked`
- Unskipped the fiber audit test that was blocked by this bug

Closes #1314

## Test plan

- [x] New regression test `tests/scheme/smoke/yield-exception-handler.scm` — yield in `with-exception-handler`, `guard`, and without scheduler
- [x] Unskipped `primitives_fiber-audit.scm` line 64 (31/31 pass)
- [x] Zig unit tests pass
- [x] Full Scheme suite: 1812 pass, 0 fail

### DIFF
--- a/src/primitives_control.zig
+++ b/src/primitives_control.zig
@@ -87,6 +87,10 @@ fn withExceptionHandlerFn(args: []const Value) PrimitiveError!Value {
         if (err == vm_mod.VMError.ContinuationInvoked) {
             return PrimitiveError.ContinuationInvoked;
         }
+        if (err == vm_mod.VMError.Yielded) {
+            vm.popHandler();
+            return PrimitiveError.Yielded;
+        }
         if (err == vm_mod.VMError.ExceptionRaised) {
             vm.popHandler();
             var handler_root = handler;
--- a/src/primitives_control.zig
+++ b/src/primitives_control.zig
@@ -87,10 +87,6 @@ fn withExceptionHandlerFn(args: []const Value) PrimitiveError!Value {
         if (err == vm_mod.VMError.ContinuationInvoked) {
             return PrimitiveError.ContinuationInvoked;
         }
-        if (err == vm_mod.VMError.Yielded) {
-            vm.popHandler();
-            return PrimitiveError.Yielded;
-        }
         if (err == vm_mod.VMError.ExceptionRaised) {
             vm.popHandler();
             var handler_root = handler;
--- a/tests/scheme/smoke/yield-exception-handler.scm
+++ b/tests/scheme/smoke/yield-exception-handler.scm
@@ -1,0 +1,29 @@
+;; Regression test for #1314: yield raises inside with-exception-handler after spawn
+(import (scheme base) (scheme write) (scheme process-context) (srfi 64))
+
+(test-begin "yield-exception-handler")
+
+;; yield should be a no-op after all fibers complete
+(fiber-join (spawn (lambda () 42)))
+
+(test-equal "yield in with-exception-handler after spawn"
+  'ok
+  (with-exception-handler
+    (lambda (e) 'fail)
+    (lambda () (yield) 'ok)))
+
+(test-equal "yield in guard after spawn"
+  'ok
+  (guard (e (#t 'fail))
+    (yield) 'ok))
+
+;; yield with no scheduler is still a no-op
+(test-equal "yield without scheduler"
+  'done
+  (with-exception-handler
+    (lambda (e) 'fail)
+    (lambda () (yield) 'done)))
+
+(let ((runner (test-runner-current)))
+  (test-end "yield-exception-handler")
+  (when (> (test-runner-fail-count runner) 0) (exit 1)))
--- a/src/primitives_fiber.zig
+++ b/src/primitives_fiber.zig
@@ -58,7 +58,8 @@ fn spawnFn(args: []const Value) PrimitiveError!Value {
 
 fn yieldFn(_: []const Value) PrimitiveError!Value {
     const vm = vm_mod.vm_instance orelse return PrimitiveError.OutOfMemory;
-    if (vm.scheduler == null) return types.VOID;
+    const sched = vm.scheduler orelse return types.VOID;
+    if (sched.schedule() == null) return types.VOID;
     vm.yielded = true;
     return types.VOID;
 }
--- a/tests/scheme/audit/primitives_fiber-audit.scm
+++ b/tests/scheme/audit/primitives_fiber-audit.scm
@@ -61,8 +61,7 @@
 (test #t (fiber? (spawn (lambda () 1))))
 (test 42 (fiber-join (spawn (lambda () (* 6 7)))))
 ;; yield with a runnable scheduler is fine
-;; SKIP: yield raises inside with-exception-handler after spawn (#1314)
-;; (test 'ok (begin (yield) 'ok))
+(test 'ok (begin (yield) 'ok))
 ;; join is memoized: joining twice returns the same result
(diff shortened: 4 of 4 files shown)


src/primitives_control.zig
```diff
@@ -87,6 +87,10 @@ fn withExceptionHandlerFn(args: []const Value) PrimitiveError!Value {
         if (err == vm_mod.VMError.ContinuationInvoked) {
             return PrimitiveError.ContinuationInvoked;
         }
+        if (err == vm_mod.VMError.Yielded) {
+            vm.popHandler();
+            return PrimitiveError.Yielded;
```
baijum: This branch is not covered by the added tests (they all take the `schedule() == null` no-op path in `yieldFn`, so `Yielded` never reaches here). In the case it targets — `yield` inside `with-exception-handler` while another fiber is runnable — `callReentrant` (`vm_calls.zig:571-575`) has already reset `frame_count`/`handler_count`/`wind_count`, discarding the thunk's frames. So the thunk's continuation after `yield` is lost: code after the `yield` never runs and `with-exception-handler` returns a garbage register value (`#<builtin with-exception-handler>`) rather than the thunk's result.

Before this PR that path raised a spurious-but-visible exception; this makes it a silent wrong value. Consider raising a clear error here (like `blockOrDeadlock` does at `primitives_fiber.zig:225`), or dropping this change since the `yieldFn` fix alone resolves #1314.

tests/scheme/smoke/yield-exception-handler.scm
```diff
@@ -0,0 +1,29 @@
+;; Regression test for #1314: yield raises inside with-exception-handler after spawn
+(import (scheme base) (scheme write) (scheme process-context) (srfi 64))
+
+(test-begin "yield-exception-handler")
+
+;; yield should be a no-op after all fibers complete
+(fiber-join (spawn (lambda () 42)))
+
+(test-equal "yield in with-exception-handler after spawn"
```
baijum, line 9: All three cases here are "no runnable fiber" cases (`fiber-join` has already drained everything, so `schedule()` returns `null`). They only exercise the `yieldFn` fix, never the new `with-exception-handler` `Yielded` branch — that requires a *runnable* fiber to be present when `yield` is called inside the handler. Adding a case that spawns a fiber and yields inside `with-exception-handler` before joining it would pin down the intended behavior for that path (which currently loses the continuation and returns a garbage value).

baijum (commented):
Reviewed on PR head `1fb1911` (ReleaseSafe build).

**Verified:** `zig build test` passes; full Scheme suite is green (**1813 pass, 0 fail, 0 skipped**), matching the PR's claim. The unskipped fiber audit test passes 31/31. The new smoke test passes.

## The core fix is correct ✅

The `yieldFn` change — making `yield` a no-op when `sched.schedule()` returns `null` — is the right fix for #1314 and fully resolves the reported case. After `fiber-join` drains all fibers, `schedule()` returns `null`, so `yield` no longer sets `vm.yielded`, and no `Yielded` signal is raised. Good, targeted change. (`schedule()` is safe to call as a predicate here — its only side effect is waking deadline-expired timers, which is idempotent.)

## Concern: the `with-exception-handler` change is untested and produces a *silently wrong* result

The second change (propagating `VMError.Yielded` out of `withExceptionHandlerFn`) is **not exercised by any of the added tests** — all three test cases hit the `schedule() == null` no-op path in `yieldFn`, so `Yielded` is never raised while a handler is on the stack.

In the scenario this branch actually targets — a `yield` inside `with-exception-handler` **while another fiber is still runnable** — the fix swaps one wrong behavior for a worse one. Because `callReentrant` catches `VMError.Yielded` and resets `frame_count`/`handler_count`/`wind_count` to their saved values (`src/vm_calls.zig:571-575`), the thunk's frames are discarded and its continuation after the `yield` is lost. Repro on this branch:

```scheme
(define f (spawn (lambda () (yield) 'f-done)))
(define r
  (with-exception-handler
    (lambda (e) 'handler)
    (lambda () (display "before\n") (yield) (display "after\n") 'ok)))
(display "result: ") (write r) (newline)
(fiber-join f)
```

prints `before` but **never `after`**, and `r` is `#<builtin with-exception-handler>` (garbage) instead of `'ok`. The same happens for a spawned fiber yielding inside `with-exception-handler` — `fiber-join` returns the garbage value. Before this PR that scenario raised a spurious (but visible, catchable) exception; now it silently returns a wrong value, which is harder to detect.

This is the known native-frame continuation limitation (yield can't suspend/resume across a re-entrant native call). The codebase's own precedent for that situation — `blockOrDeadlock` in `src/primitives_fiber.zig:225` — raises a **clear, catchable error** rather than returning an unspecified value.

**Suggestion:** either (a) drop the `withExceptionHandlerFn` change entirely, since `yieldFn` alone fixes #1314 and this branch never yields a correct resume; or (b) if kept, make it raise a clear error (à la `blockOrDeadlock`) instead of silently propagating, and add a test that pins down the yield-with-runnable-fiber behavior.

## Verdict

The headline fix for #1314 is solid and I'd merge it as-is. I'd hold on the `with-exception-handler` change until it either raises a clear error or is dropped — as written it introduces a silent-wrong-value path in an untested scenario.

baijum (commented):
Re-reviewed after commit `e811af1` ("Drop with-exception-handler Yielded propagation"). Both concerns from my previous review are addressed.

**The `withExceptionHandlerFn` change is dropped** — `src/primitives_control.zig` is now unchanged vs `main`. The PR is reduced to the correct, targeted core fix: `yieldFn` no-ops when `sched.schedule()` returns `null`. The silent-wrong-value path I flagged is gone.

**Verified on `e811af1`:**
- `zig build test` — passes
- Full Scheme suite — **1813 pass, 0 fail, 0 skipped**
- Fiber audit — 31/31 (the unskipped `#1314` test passes)
- Issue #1314 repro — now behaves correctly (yield is a no-op after `fiber-join` drains all fibers)
- yield-in-`with-exception-handler`-with-a-runnable-fiber — reverts to the pre-existing visible/catchable exception (native-frame continuation limitation on `main`, out of scope for #1314), no longer a silent garbage value

LGTM — ready to merge. 🎉

PR #1370, opened 2026-07-09
Add descriptive FFI call-time error messages (#1187)

### LINKED ISSUE #1187: [FFI] call-time marshaling errors carry no detail — arity mismatch, wrong arg type, NUL-in-string all raise bare "error"
**Category:** FFI / diagnostics
**Severity:** edge-case (correct rejections, useless messages)

**Minimal reproduction:**
```scheme
(define lib (ffi-open #f))
(define c-abs (ffi-fn lib "abs" '(int) 'int))
(guard (e (#t (error-object-message e))) (c-abs -3.0))     ;; => "error"
(guard (e (#t (error-object-message e))) (c-abs "x"))      ;; => "error"
(define c-sqrt (ffi-fn lib "sqrt" '(double) 'double))
(guard (e (#t (error-object-message e))) (c-sqrt 1.0 2.0)) ;; => "error"
(define c-strlen (ffi-fn lib "strlen" '(string) 'size_t))
(guard (e (#t (error-object-message e))) (c-strlen (string #\a #\null #\b))) ;; => "error"
```

**Expected:** messages naming the FFI function, the parameter position, and what was expected — the *creation-time* errors already do this well ("ffi-fn: too many parameters (max 16)", dlopen/dlsym details), and #1057's sweep established the convention elsewhere.
**Actual:** every call-time rejection in the src/ffi.zig dispatch returns `error.TypeError` without `vm.setErrorDetail`, so users debugging an FFI type mismatch get the bare word "error".

FFI misuse is where good diagnostics matter most — the failure is at the boundary where the user is already reasoning about two type systems at once. The FfiFunction knows its own name and signature; the marshal loop knows the argument index.

**Verified:** fresh ReleaseSafe build at 055a5b1
**Source location:** src/ffi.zig call dispatcher and per-type marshal helpers (toIntArgOpt etc.)
**Disabled test:** none (behavior is correct, only the message is empty; the audit file asserts catchability)

**Found by:** Systematic audit, Phase 2.13

baijum:
## Summary

- Pass VM to `callFfi` so it can set `vm.setErrorDetail` with the FFI function name, argument position, and expected/actual types at each failure point
- Add `ffiTypeName`/`schemeTypeName` helpers and `validateArgsDetailed` in `src/ffi.zig` covering wrong type, NUL-in-string, string-too-long, int range overflow, closed library, and unsupported signature errors
- Add arity mismatch error details at all 4 FFI call sites in `vm_dispatch.zig` and `vm_calls.zig`

Before: `(c-abs -3.0)` → `"error"`
After: `(c-abs -3.0)` → `"'abs': argument 1 must be int, got flonum"`

Closes #1187

## Test plan

- [x] Unit tests pass (`zig build test`)
- [x] Full Scheme suite passes (1813 pass, 0 fail)
- [x] All existing FFI tests pass
- [x] New regression test `tests/scheme/ffi/error-messages.scm` covers all 4 error categories from the issue

### DIFF
--- a/src/ffi.zig
+++ b/src/ffi.zig
@@ -41,6 +41,8 @@ fn schemeTypeName(v: Value) []const u8 {
     if (types.isNil(v)) return "nil";
     if (types.isPair(v)) return "pair";
     if (types.isSymbol(v)) return "symbol";
+    if (types.isVector(v)) return "vector";
+    if (types.isClosure(v) or types.isNativeFn(v)) return "procedure";
     return "object";
 }
 
@@ -280,6 +282,11 @@ fn validateArgsDetailed(ffi_fn: *types.FfiFunction, args: []const Value, vm: *VM
                     });
                     return error.TypeError;
                 }
+            } else {
+                vm.setErrorDetail("'{s}': argument {d} out of range for {s}", .{
+                    ffi_fn.name, i + 1, ffiTypeName(ffi_fn.param_types[i]),
+                });
+                return error.TypeError;
             }
         }
         if (nt == .string) {
@@ -436,6 +443,7 @@ fn callFfiGeneric(comptime N: u4, ffi_fn: *types.FfiFunction, args: []const Valu
 
 /// Main FFI call dispatcher. Routes to arity-specific handlers.
 pub fn callFfi(ffi_fn: *types.FfiFunction, args: []const Value, gc: *memory.GC, vm: *VM) !Value {
+    vm.last_error_detail_len = 0;
     if (types.isFfiLibrary(ffi_fn.library)) {
         const lib = types.toObject(ffi_fn.library).as(types.FfiLibrary);
         if (lib.handle == null) {
--- a/src/ffi.zig
+++ b/src/ffi.zig
@@ -1,8 +1,48 @@
 const std = @import("std");
 const types = @import("types.zig");
 const memory = @import("memory.zig");
+const VM = @import("vm.zig").VM;
 const Value = types.Value;
 const FfiType = types.FfiType;
+
+fn ffiTypeName(t: FfiType) []const u8 {
+    return switch (t) {
+        .int => "int",
+        .long => "long",
+        .double => "double",
+        .float => "float",
+        .string => "string",
+        .pointer => "pointer",
+        .void_type => "void",
+        .bool_type => "bool",
+        .uint8 => "uint8",
+        .int8 => "int8",
+        .int16 => "int16",
+        .int32 => "int32",
+        .int64 => "int64",
+        .uint16 => "uint16",
+        .uint32 => "uint32",
+        .uint64 => "uint64",
+        .size_type => "size_t",
+        .char_type => "char",
+    };
+}
+
+fn schemeTypeName(v: Value) []const u8 {
+    if (types.isFixnum(v)) return "integer";
+    if (types.isFlonum(v)) return "flonum";
+    if (types.isString(v)) return "string";
+    if (types.isBignum(v)) return "integer";
+    if (v == types.TRUE or v == types.FALSE) return "boolean";
+    if (types.isChar(v)) return "character";
+    if (types.isBytevector(v)) return "bytevector";
+    if (types.isFfiCallback(v)) return "ffi-callback";
+    if (types.isRationalObj(v)) return "rational";
+    if (types.isNil(v)) return "nil";
+    if (types.isPair(v)) return "pair";
+    if (types.isSymbol(v)) return "symbol";
+    return "object";
+}
 
 fn toIntArgOpt(v: Value) ?i64 {
     if (v == types.TRUE) return 1;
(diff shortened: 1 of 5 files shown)


baijum (commented):
## Review — looks good, ready to merge

Solid, focused fix. It closes #1187 cleanly: every call-time rejection path in `callFfi` now sets `vm.setErrorDetail` with the function name, argument position, and expected/actual types, and it goes beyond the issue's four cases to also cover out-of-range ints, over-long strings, closed libraries, and unsupported signatures. Threading `vm` through `callFfi` and updating all four call sites (+ the `tests_ffi.zig` bool tests) is done consistently, and switching those tests to `TestContext` is the idiomatic move given the new dependency.

I verified locally on `a24c167`:
- `zig build` and `zig build test` both green
- all 13 `tests/scheme/ffi/*.scm` pass, including the new `error-messages.scm`
- exercised each path by hand — messages read well:
  - `'abs': argument 1 must be int, got flonum`
  - `'abs': argument 1 must be int, got string`
  - `'sqrt': expected 1 arguments, got 2`
  - `'strlen': argument 1 string contains NUL byte`
  - `'strlen': argument 1 string too long (5000 bytes, max 4095)`
  - `'abs': FFI library is closed`

The `4096`/`4095` bounds and the NUL scan in `validateArgsDetailed` mirror `toCString` exactly, so validation and marshaling agree. No missed `callFfi` call sites; the old `validateArgs` is renamed rather than left as dead code.

**Verdict: ready to merge.** The two inline notes below are optional cosmetic polish, not blockers. (Also, `"expected 1 arguments"` is ungrammatical for count 1, but that phrasing is shared verbatim by ~14 pre-existing arity sites across the VM, so it's a codebase-wide convention — out of scope for this PR.)

src/ffi.zig
```diff
@@ -217,12 +257,46 @@ fn checkNarrowIntRange(v: Value, declared: FfiType) error{TypeError}!void {
...
-        try checkNarrowIntRange(args[i], ffi_fn.param_types[i]);
+        checkNarrowIntRange(args[i], ffi_fn.param_types[i]) catch {
+            vm.setErrorDetail("'{s}': argument {d} out of range for {s}", .{
+                ffi_fn.name, i + 1, ffiTypeName(ffi_fn.param_types[i]),
+            });
+            return error.TypeError;
+        };
+        const nt = normalizeType(ffi_fn.param_types[i]);
+        if (nt == .int and ffi_fn.param_types[i] != .bool_type) {
+            if (toIntArgOpt(args[i])) |wide| {
```
baijum, line 278: Minor inconsistency in the out-of-range message: this block only runs the `c_int` range check when `toIntArgOpt` returns non-null. A bignum too large for `i64` (e.g. `(c-abs (expt 2 100))` for an `int` param) makes `toIntArgOpt` return `null`, so we skip this check, marshaling then fails, and the outer switch `catch` reports the generic `'abs': unsupported FFI signature`. Meanwhile `(c-abs (expt 2 40))` (fits i64, overflows c_int) correctly reports `out of range for int`. Both are genuinely out-of-range, so the messages diverge. Still strictly better than the old bare `"error"`, so non-blocking — but if you want consistency, treat a `null` from `toIntArgOpt` on a normalized-int param as out-of-range here too.

src/ffi.zig
```diff
@@ -1,9 +1,49 @@
...
+    if (types.isBignum(v)) return "integer";
+    if (v == types.TRUE or v == types.FALSE) return "boolean";
+    if (types.isChar(v)) return "character";
+    if (types.isBytevector(v)) return "bytevector";
+    if (types.isFfiCallback(v)) return "ffi-callback";
+    if (types.isRationalObj(v)) return "rational";
+    if (types.isNil(v)) return "nil";
+    if (types.isPair(v)) return "pair";
+    if (types.isSymbol(v)) return "symbol";
+    return "object";
```
baijum, line 46: Optional: passing a vector/procedure to a scalar param yields `... must be int, got object`. Adding `if (types.isVector(v)) return "vector";` (and procedure, if there's a cheap predicate) would make these messages as specific as the flonum/string cases. Minor.

baijum:
## Follow-up review — feedback addressed, ready to merge ✅

`65c62f2` resolves both of my inline notes and the stale-detail concern I'd raised in analysis:

- **Out-of-range consistency** — the new `else` branch in `validateArgsDetailed` now reports `out of range` when `toIntArgOpt` returns `null` (multi-limb bignum). Confirmed: `(c-abs (expt 2 100))` now gives `'abs': argument 1 out of range for int` instead of the old generic `unsupported FFI signature`.
- **`schemeTypeName` specificity** — `vector` and `procedure` (closure/native-fn) added. Confirmed: `(c-abs (vector 1 2))` → `'abs': argument 1 must be int, got vector`.
- **Stale detail** — `vm.last_error_detail_len = 0` at the top of `callFfi` makes the outer `catch` fallback reliable. Placement is safe: the arity checks that set a detail live in the callers and early-return before `callFfi` is ever reached, so nothing they set is clobbered.
- **Linux CI fix** — switching the arity test from `sqrt` (libm) to `abs` (libc) is the right call; `sqrt` isn't reliably resolvable through `(ffi-open #f)` on Linux. Also good to see closed-library and out-of-range regression tests added.

Re-verified locally on `65c62f2`: `zig build` ✓, `zig build test` ✓, all 13 `tests/scheme/ffi/*.scm` pass (`error-messages.scm` now 6/6). CI `format` is green; the rest are still running.

**Verdict: ready to merge** once CI goes green. Nice, thorough turnaround.

PR #1371, opened 2026-07-09
Fix SRFI-9 record-type redefinition retargeting old procedures (#1203)

### LINKED ISSUE #1203: [SRFI-9] record-type redefinition retargets previously-created constructors and predicates (call-time lookup of the __record_type_ global)
**Spec reference:** SRFI-9: "DEFINE-RECORD-TYPE is generative: each use creates a new record type that is distinct from all existing types, including other record types and Scheme's predefined types." A predicate/constructor created by the first use belongs to the first type; re-evaluating the definition must not change what already-created procedures refer to.
**Severity:** edge-case (wrong-result after redefinition; REPL sessions hit this routinely)

**Minimal reproduction:**
```scheme
(define-record-type tt (mk-tt) tt?)
(define old (mk-tt))
(define old-pred tt?)
(define old-mk mk-tt)
(define-record-type tt (mk-tt v) tt? (v tt-v))   ; redefine

(old-pred old)      ;; expected: #t   actual: #f — old predicate no longer recognizes its own instances
(tt? (old-mk))      ;; expected: #f   actual: #t — old constructor now creates NEW-type records
```
Worse, `(old-mk)` builds a zero-field instance that satisfies the new one-field type's predicate, so `(tt-v (old-mk))` hits the index-out-of-bounds path.

**Root cause:** the desugared procedures reference the type through the hidden global `" __record_type_<name>"` **by name at call time** — e.g. `(define (tt? v) (%record? v __record_type_tt))`, `(define (mk-tt) (%make-record __record_type_tt))` (src/vm_records.zig:101,128,197). Redefinition rebinds that global, silently retargeting every previously-created constructor/predicate. The generated code should capture the record-type object itself (embed it as a constant / close over it), not resolve the name per call.

**Related:** #1199 (accessors/mutators skip the type check entirely, which partially masks this).

**Source location:** src/vm_records.zig (`handleDefineRecordType` — constructor/predicate generation)
**Disabled test:** tests/scheme/srfi/srfi9.scm (marked ;; FAIL: #this)

**Found by:** Systematic audit, Phase 3.1

baijum:
## Summary

- Record-type constructors/predicates/accessors/mutators now close over the record-type object at definition time via `let`, instead of resolving the `__record_type_<name>` global at call time
- Fixes both the top-level path (`handleDefineRecordType`) and the body context path (`expandRecordTypeDefines`)
- Enabled 2 previously disabled SRFI-9 tests and added a Zig unit test

## Test plan

- [x] Zig unit tests pass (671/671)
- [x] SRFI-9 test suite passes (39/39, including 2 previously disabled)
- [x] Full Scheme test suite passes (1812/1812, 0 failures)
- [x] Manual verification of the reproduction from the issue

Closes #1203

### DIFF
--- a/src/tests_core_eval.zig
+++ b/src/tests_core_eval.zig
@@ -258,6 +258,23 @@ test "typeName covers all ObjectTags exhaustively" {
     try std.testing.expectEqualStrings("record", types.typeName(rec));
 }
 
+// Regression for #1203: record-type redefinition must not retarget old
+// constructors/predicates — they must close over the original type.
+test "record-type redefinition does not retarget old procedures (#1203)" {
+    var ctx: th.TestContext = undefined;
+    try ctx.init();
+    defer ctx.deinit();
+    _ = try ctx.vm.eval("(define-record-type tt (mk-tt) tt?)");
+    _ = try ctx.vm.eval("(define old-inst (mk-tt))");
+    _ = try ctx.vm.eval("(define old-pred tt?)");
+    _ = try ctx.vm.eval("(define old-mk mk-tt)");
+    _ = try ctx.vm.eval("(define-record-type tt (mk-tt v) tt? (v tt-v))");
+    const r1 = try ctx.vm.eval("(old-pred old-inst)");
+    try std.testing.expectEqual(types.TRUE, r1);
+    const r2 = try ctx.vm.eval("(tt? (old-mk))");
+    try std.testing.expectEqual(types.FALSE, r2);
+}
+
 // Regression for #1202: parameterize must evaluate all value expressions
 // before installing any bindings — (b (a)) must see the outer value of a.
 test "parameterize evaluates values before binding (#1202)" {
--- a/src/vm_records.zig
+++ b/src/vm_records.zig
@@ -40,28 +40,19 @@ pub fn handleDefineRecordType(vm: *VM, args: Value) VMError!Value {
         vm.defineGlobal(internal_name, rt_val) catch return VMError.OutOfMemory;
     }
 
-    // Generate constructor:
-    // (define (make-point x y) (%make-record __record_type_point x y))
-    // But we need to handle field ordering: constructor args may be in a different order
-    // than the field specs. The constructor always creates fields in the field_spec order.
+    // Generate constructor — close over the record type so redefinition
+    // does not retarget previously created constructors (#1203):
+    // (define ctor (let (( __rt __record_type_X))
+    //               (lambda (f1 f2) (%make-record  __rt f_for_0 ...))))
     {
         vm.gc.no_collect += 1;
         errdefer vm.gc.no_collect -= 1;
-        // Build the body: (%make-record <type> <fields-in-field-order>)
-        // For each field in all_fields order, find it in the constructor args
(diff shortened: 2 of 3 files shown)


baijum:

Approach is correct: wrapping each generated procedure in `(let (( __rt <global>)) (lambda …))` captures the record-type **value** at definition time via closure, so redefinition (which rebinds the `__record_type_<name>` global) no longer retargets previously-created constructors/predicates/accessors/mutators. Fixing both the top-level path (`handleDefineRecordType`) and the body-context path (`expandRecordTypeDefines`) keeps the two desugarers consistent. Using a space-prefixed `" __rt"` local also keeps it hygienically clear of any user field/parameter name (the reader can't produce leading-space symbols).

### Verification (local, ReleaseSafe on macOS arm64)
- ✅ `zig build test` — 671/671
- ✅ SRFI-9 suite — 39/39 (incl. the 2 re-enabled tests)
- ✅ Full Scheme suite (`run-all.sh`) — **1812 pass, 0 fail**
- ✅ Issue #1203 reproduction now returns `#t` / `#f` correctly; new type still constructs/queries fine and the old instance is rejected
- ✅ Rebuilt with `-Dgc-stress=true` — repro + SRFI-9 still pass, so the rooting (`no_collect` during construction, `pushRoot` before compile) is sound
- ✅ Body-context path (record defined inside a lambda body) and library-context path (record exported from a `define-library`) both verified independently
- ✅ **Regression test is genuine**: reverting only `src/vm_records.zig` to `main` while keeping the new test makes exactly `record-type redefinition does not retarget old procedures (#1203)` fail (670 pass, 1 fail)

### Minor (non-blocking)

**1. Stale comment in `tests/scheme/srfi/srfi9.scm` (lines 97–99).** The comment still describes the *old buggy* behavior in the present tense right above the now-passing assertions:

```scheme
;; ...but previously-created procedures must keep referring to the OLD type.
;; The desugar resolves the __record_type_ global at call time, so they
;; silently retarget to the new type:
(test #t (old-pred old-inst))
(test #f (tt? (old-mk)))
```

Since the desugar now closes over the type, lines 98–99 read backwards. Suggest something like: "…must keep referring to the OLD type. The desugar closes over the record-type object at definition time, so redefinition does not retarget them."

**2. Optional — duplication.** The `let`/`lambda`/`let_binding`/`let_bindings`/`let_expr` wrapping is repeated 8× (4 procedures × 2 paths). A small helper that returns `(let (( __rt <internal_name>)) (lambda <params> <body>))` given `(gc, internal_name, params, body)` would remove ~50 lines and shrink the error surface. Consistent with the file's existing per-procedure repetition, so purely a cleanup — fine to leave as-is.

Nothing here blocks merge.

baijum:
Follow-up on 52a4459 — nit #1 (stale comment) is resolved; the wording now correctly describes the closure-capture behavior. SRFI-9 still passes 39/39.

The source fix (`src/vm_records.zig`) and the Zig regression test are unchanged from the version I verified, so the earlier results still hold (full suite 1812 pass / 0 fail, clean under `-Dgc-stress=true`, and the regression test genuinely fails without the fix). Nit #2 (the optional dedup) was non-blocking and is fine to leave.

**Ready to merge.** ✅

Issue #877, opened 2026-07-02
define-syntax in library body registers macros globally at load time — unexported macros leak, imported macros silently clobbered


baijum:
## Summary

`define-syntax` inside a library `begin` block registers the macro directly into the **global** `vm.macros` table at library *load* time. Two observable consequences:

1. **Unexported macros leak**: any macro defined in a library body becomes usable by all code in the process once the library is loaded — even when the importer filters it out with `only`.
2. **Silent clobbering / wrong expansion**: loading library B overwrites a same-named macro that the program explicitly imported from library A — the program silently gets B's expansion. This is a wrong-answer bug, not just pollution.

This is distinct from the fixed import-time issues #495/#608 (which were about `importBinding`): here the macro never goes through import at all — it lands in the global table while the library body is being compiled.

## Location

`src/vm_library.zig:1055` (v0.11.1, commit d46986f), in `compileLibExpr`:

```zig
const func = compiler_mod.compileExpressionInEnv(vm.gc, expr, &vm.macros, lib_env) catch return VMError.CompileError;
```

Library body forms are compiled against the global macro table (`&vm.macros`), so the compiler's `define-syntax` handling registers library-body macros globally. Variable definitions correctly go into the per-library `lib_env` (a non-exported `(define (helper) ...)` does *not* leak — verified), but macros have no per-library home.

## Reproduction

```scheme
; maclib/m.sld
(define-library (maclib m)
  (import (scheme base))
  (export public-f)
  (begin
    (define-syntax private-mac (syntax-rules () ((_ x) (list 'private x))))
    (define (public-f) (private-mac 1))))
```
```scheme
; leak.scm — private-mac is NOT exported, and only public-f is imported
(import (scheme base) (scheme write) (only (maclib m) public-f))
(display (private-mac 7)) (newline)
```
```
$ kaappi --lib-path . leak.scm
(private 7)            ; expected: undefined/syntax error
```

Clobbering (wrong answer):

```scheme
; maclib2/a.sld
(define-library (maclib2 a) (import (scheme base)) (export tag)
  (begin (define-syntax tag (syntax-rules () ((_) 'from-a)))))
; maclib2/b.sld
(define-library (maclib2 b) (import (scheme base)) (export tag bfun)
  (begin (define (bfun) 'bfun)
         (define-syntax tag (syntax-rules () ((_) 'from-b)))))
```
```scheme
; clobber.scm
(import (scheme base) (scheme write))
[...]

PR #1372, opened 2026-07-09
Fix non-exported library macros leaking to importers (#1332)

### LINKED ISSUE #1332: Non-exported library macros leak to importers
Non-exported macros defined in a library's `begin` block are visible to importers — they resolve to `#<transformer>` in the consumer namespace. Non-exported procedures correctly stay private.

**Reproduction:**
```scheme
;; lib.sld
(define-library (my lib)
  (import (scheme base))
  (export outer)
  (begin
    (define-syntax helper    ;; NOT exported
      (syntax-rules ()
        ((helper x) (+ x 1))))
    (define-syntax outer
      (syntax-rules ()
        ((outer x) (helper x))))
    (define (secret) 42)))   ;; NOT exported
```
```scheme
(import (scheme base) (my lib))
(display helper)    ;; prints #<transformer> — should be unbound
(display secret)    ;; correctly errors: undefined variable
```

**Expected:** `helper` should not be accessible from outside the library, same as `secret`.

**Context:** noticed during SRFI-41 review (PR #1330) where `stream-of-aux` is an internal helper macro that leaks. The SRFI structure is correct — this is a library-system scoping issue.

baijum:
## Summary

- Non-exported macros defined in a library's `begin` block were visible to importers as value bindings (e.g. `(display helper)` printed `#<transformer>`)
- `copyTransformerFreeRefs` now skips putting transformer free refs into `vm.globals` — they only go into `vm.macros` where the compiler needs them for expansion chains
- Non-exported procedures were already correctly private; this fix brings macro scoping to parity

## Test plan

- [x] New regression test `tests/scheme/compliance/macro-export-scope.scm` — exported macro works, non-exported macro/procedure not in globals, transitive helper chain
- [x] Zig unit tests pass
- [x] Full Scheme test suite: 1813 pass, 0 fail (including all 72 SRFIs with internal helper macros like SRFI-64's `%test-comp1body`)

Closes #1332

### DIFF
--- a/src/vm_library.zig
+++ b/src/vm_library.zig
@@ -73,21 +73,27 @@ fn copyTransformerFreeRefs(
         _ = macro.collectFreeRefs(tmpl, pv_names[0..pv_count], tx.literals, &free_names, &free_count);
         for (free_names[0..free_count]) |fname| {
             if (env.get(fname)) |fval| {
+                const is_tx = types.isTransformer(fval);
                 if (target == vm.globals) {
-                    vm.globals_lock.lock();
-                    const missing = !target.contains(fname);
-                    if (missing) {
-                        target.put(fname, fval) catch {
-                            vm.globals_lock.unlock();
-                            return error.OutOfMemory;
-                        };
+                    // Non-exported transformer free refs go into vm.macros
+                    // only (below), not vm.globals — keeps non-exported
+                    // library macros from leaking as bindings (#1332).
+                    if (!is_tx) {
+                        vm.globals_lock.lock();
+                        const missing = !target.contains(fname);
+                        if (missing) {
+                            target.put(fname, fval) catch {
+                                vm.globals_lock.unlock();
+                                return error.OutOfMemory;
+                            };
+                        }
+                        vm.globals_lock.unlock();
+                        if (missing) vm.global_version +%= 1;
                     }
-                    vm.globals_lock.unlock();
-                    if (missing) vm.global_version +%= 1;
                 } else if (!target.contains(fname)) {
                     target.put(fname, fval) catch return error.OutOfMemory;
                 }
-                if (types.isTransformer(fval)) {
+                if (is_tx) {
                     // An exported macro may expand into a library-internal
                     // helper macro (e.g. SRFI 64 test-assert → %test-comp1body).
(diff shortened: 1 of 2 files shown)


baijum (commented):
## Review — ready to merge ✅

The fix is correct, minimal, and well-targeted. `copyTransformerFreeRefs` now routes transformer free-refs into `vm.macros` only (where the compiler needs them for use-site expansion) and no longer mirrors them into `vm.globals` as value bindings, which is what made non-exported helper macros referenceable. Non-transformer free-refs still land in `vm.globals` as before, and the transitive macro→macro recursion (the SRFI-64 `%test-comp1body` / #877 path) is preserved.

### Verified locally
- **Regression test is genuine**: with the fix reverted but the test kept, `macro-export-scope.scm` fails exactly on the 3 "not in globals" assertions; with the fix it's 6/6. Good failing-first coverage.
- `zig build test` passes.
- Macro/hygiene/library smoke tests, `srfi41` (84), `srfi64` (20 pass / 1 xfail / 1 skip), `srfi141` (44), and `srfi-import-order` (2) all pass.
- Confirmed the original motivating case: SRFI-41's internal `stream-of-aux` is now `unbound` from an importer, while exported macros (`define-stream`, `stream-take`) still work.
- Confirmed the transitive/re-export case: `(t b)` re-exporting `(t a)`'s `outer` expands correctly (→ 11) while `helper` stays `unbound`.

### Minor, non-blocking notes (optional / possible follow-ups)
1. **Non-globals import target keeps the old behavior.** `src/vm_library.zig:87-89` — when the import target is a library's `lib_env` (library-into-library import), transformer free-refs are still inserted as value bindings. This is pre-existing and **not observable at the top level** (only exported names propagate outward — I verified `helper` stays unbound through a re-export chain), so it's not a blocker. If you want full internal parity you could apply the same `is_tx` guard on that branch, but it's cosmetic.
2. **Exported macros remain referenceable as values** (out of scope for #1332). `(display define-stream)` still yields `#<transformer>` rather than erroring. R7RS treats referencing a macro keyword outside operator position as an error, so this is a separate, pre-existing quirk — worth a follow-up issue but not part of this fix.

Nice work — the `is_tx` hoist reads cleanly and the comments capture the intent well.

PR #1386, opened 2026-07-10
Pin the utf8->string rejection to its type error in the audit test

baijum:
Follow-up to #1383 (review feedback that landed after auto-merge): the re-enabled invalid-UTF-8 assertions catch every condition via the audit corpus's `'caught` idiom, so an unrelated failure would still pass them. Add one assertion pinning `error-object?` and the `type error in 'utf8->string'` message prefix on the representative bad-lead-byte case.

Verified: 127/127 passes on ReleaseSafe and under `-Dgc-stress=true`.

### DIFF
--- a/tests/scheme/audit/primitives_bytevector-audit.scm
+++ b/tests/scheme/audit/primitives_bytevector-audit.scm
@@ -129,6 +129,16 @@
 (test-equal 'caught (guard (e (#t 'caught)) (utf8->string #u8(#x41 #xFF #x42) 1 2)))
 (test-equal "A" (utf8->string #u8(#xFF #x41) 1))
 (test-equal 'caught (guard (e (#t 'caught)) (utf8->string #u8(#x41 #xCE #xBB) 0 2)))    ; range splits λ
+;; the rejection is a proper error object naming the conversion, not some
+;; unrelated condition
+(test-equal '(#t "type error in 'utf8->string'")
+  (guard (e (#t (list (error-object? e)
+                      (let ((m (error-object-message e)))
+                        (and (string? m)
+                             (>= (string-length m) 28)
+                             (substring m 0 28))))))
+    (utf8->string (bytevector #xFF))
+    'no-error))
 
 ;;; --- string->utf8 ---
 ;; R7RS 6.9 example


baijum:
## Review

**What it does** — adds one SRFI-64 assertion to the bytevector audit pinning that the `utf8->string` invalid-UTF-8 rejection (re-enabled in #1383) raises a proper error object whose message names the conversion, closing the gap where the corpus's `'caught` idiom would accept *any* condition (wrong-error passes).

**Verified locally** (macOS ARM, main @ 72da4981 + this diff):

- The exact assertion body evaluates to `(#t "type error in 'utf8->string'")`. The full runtime message is `type error in 'utf8->string': expected valid UTF-8 byte sequence, got #<bytevector>` (from `primitives.typeError` in `src/primitives.zig:310`), and 28 is precisely the pinned prefix length.
- Full audit file: **127/127 on ReleaseSafe** and **127/127 under `-Dgc-stress=true`**, matching the PR description. Main's copy gives 126, so the delta is exactly this one test.

**Test design** — solid:

- The `'no-error` sentinel makes the no-raise case fail loudly instead of vacuously passing.
- The `(string? m)` + length guard means a malformed/short message yields `(#t #f)` — a readable `test-equal` mismatch — rather than a raise inside the guard clause.
- Pinning the prefix (not the whole message) keeps the test robust to the value-description suffix, which could legitimately change how a bytevector prints.
- Given the file imports only `(scheme base)`/`(srfi 64)`, `substring` is the right tool — `string-prefix?` would need a new SRFI-13 import for one assertion.
- 2-space indent, comment style, and placement (immediately after the #1383 assertions it hardens) all match the surrounding corpus. Co-locating it here rather than in `tests/scheme/errors/error-format.sh` is the right call — the point is hardening these specific audit assertions, not general message-format coverage.

One non-blocking nit posted inline (the literal `28`).

**Merge readiness**: test-only change, verified correct, follows conventions. `format`, macOS, riscv64, wasm, and CodeRabbit checks are green; the remaining ubuntu jobs were still in progress at review time and exercise nothing this diff doesn't already pass locally. Auto-merge (squash) is armed, so this is **ready to merge** — it will land on its own once the remaining checks finish.

tests/scheme/audit/primitives_bytevector-audit.scm
```diff
@@ -129,6 +129,16 @@
...
 (test-equal "A" (utf8->string #u8(#xFF #x41) 1))
 (test-equal 'caught (guard (e (#t 'caught)) (utf8->string #u8(#x41 #xCE #xBB) 0 2)))    ; range splits λ
+;; the rejection is a proper error object naming the conversion, not some
+;; unrelated condition
+(test-equal '(#t "type error in 'utf8->string'")
+  (guard (e (#t (list (error-object? e)
+                      (let ((m (error-object-message e)))
+                        (and (string? m)
+                             (>= (string-length m) 28)
+                             (substring m 0 28))))))
```
baijum, line 139: nit (non-blocking): the literal `28` has to be kept in sync with the expected prefix by hand. `(substring m 0 (string-length "type error in 'utf8->string'"))` would self-synchronize — or bind the prefix once in a `let` shared with the expected value. No correctness risk as-is: if the two ever drift, the test fails loudly rather than silently passing.

PR #1388, opened 2026-07-10
Ground fuzzing feasibility note in the research literature

baijum:
## Summary

Follow-up to #1387. The fuzzing feasibility note argued from Fuzzilli's design alone; this grounds the same analysis in the primary research literature (18 verified papers) and lets it sharpen the plan.

- **New "What the research literature says" section**, organized by theme — byte-level mutation stalling at the parser (Token-Level Fuzzing, GRIMOIRE), grammar-based generation (LangFuzz, NAUTILUS, Gramatron, PolyGlot), differential/metamorphic oracles (Csmith, EMI, YARPGen, classfuzz), functional-language precedents (Pałka et al., Midtgaard et al., QuickCheck lineage, Weinholt's AFL++/Loko write-up), the Fuzzilli line (NDSS '23 paper, DIE, FuzzJIT, Montage), and LLM generation (Fuzz4All). Every entry ends with what it means for Kaappi specifically.
- **Key find:** Zig's `std.testing.Smith` is exactly Zest's parametric-generator architecture (ISSTA 2019) — the fuzzer mutates a decision string, the generator maps it to valid structured inputs, coverage closes the loop. The proposed Tier 2 grammar generator is therefore a published, validated design that Zig's fuzz API natively supports, not a homegrown gamble.
- **Tier 3 restructured into two variants by setup cost:** a *Kaappi-vs-itself* oracle first (bytecode VM vs LLVM native backend — structurally identical to Midtgaard et al.'s OCaml setup and FuzzJIT's in-process trick; EMI dead-code pruning as a second self-oracle), then the external reference-Scheme variant. Includes the caveat that native-backend diffing must restrict generation to natively compiled forms, or the `kaappi_eval` fallback degenerates the diff into VM-vs-VM.
- Tiers 1/2/4 gain supporting citations; TL;DR gains a literature-verdict bullet; operating guidance adds pairing fuzz runs with `-Dgc-stress=true` builds.

Doc-only change (`docs/dev/fuzzing-feasibility.md`); no code touched. All citations verified against the papers' venue pages this session.

### DIFF
--- a/docs/dev/fuzzing-feasibility.md
+++ b/docs/dev/fuzzing-feasibility.md
@@ -246,9 +246,13 @@ is comparing two things that must agree:
   branch, and boolean-simplification patterns to stress those passes.
 - **[classfuzz](https://dl.acm.org/doi/10.1145/2908080.2908095)** (Chen et
   al., PLDI 2016) and its successor classming (ICSE 2019). Differential
-  testing of JVMs by mutating *class files*, not source. The precedent for
-  taking the `.sbc` loader target beyond parse robustness: semantically
-  mutate bytecode and diff against source-path execution.
+  testing of JVMs by mutating *class files*, not source: coverage-guided
+  mutant selection on a reference JVM, then diffing each mutant's behaviour
+  across several independent JVM implementations. Kaappi has no second VM
+  to diff against, so what transfers is the input level, not the oracle:
+  semantics-preserving `.sbc` mutations, checked against unmutated
+  source-path execution, could take the loader target beyond parse
+  robustness.
 
 ### Functional-language precedents (closest to Scheme)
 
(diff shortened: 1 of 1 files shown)


baijum:
Addressed all four review findings in 3009b3f4:

1. **classfuzz precedent** — corrected to describe the papers' actual oracle (coverage-guided class-file mutant selection on a reference JVM, diffed across multiple independent JVM implementations). The `.sbc` idea is now explicitly framed as our own adaptation — semantics-preserving `.sbc` mutations checked against unmutated source-path execution — since Kaappi has no second VM to diff against.
2. **Reference-Scheme pinning** — Tier 3 now says to pin one oracle at a fixed version (Chibi first) with fixed invocation, timeout handling, and output normalization, noting Guile's explicit R7RS-mode requirement as the example of why an open-ended list produces false diffs. Implementation detail lives in #1396.
3. **Throughput claim** — rephrased as a qualitative expectation with its reason (LLM inference round-trip per input vs in-process eval), no bare figure.
4. **gc-stress wording** — verified against `maybeCollect` (src/memory.zig:1100): under stress it collects only when the GC is enabled and `no_collect == 0`. Doc now says "attempts a collection at every allocation" with that qualification.

Issue #1401, opened 2026-07-10
Unit test suite crashes under -Dgc-stress=true (~440 of 690 tests)


baijum:
Found by the first-ever CI execution of the full unit-test suite under GC stress — the `gc-stress` variant of the new fuzz workflow (#1390 / PR #1398) on its first `workflow_dispatch` run.

## Evidence

**CI (x86_64 Linux, fuzz-instrumented build)** — [run 29073769556](https://github.com/kaappi/kaappi/actions/runs/29073769556), `zig build test --fuzz=200 -Dgc-stress=true`:

```
run test unit-tests 246 pass, 1 fail, 443 crash (690 total)
```

The job then hung in the fuzz-rerun phase and was killed by the 55-minute job timeout.

**Local (aarch64 macOS, plain build)** — `zig build test -Dgc-stress=true` at ea008c6b:

```
run test unit-tests 248 pass, 6 fail, 436 crash (690 total)
```

Same profile on both platforms, instrumented or not — this is gc-stress itself, not the fuzzer.

## Crash signatures (samples)

- `bignum.test.bignum negate` — SIGABRT / segfault in `memcpy` via `viewOf` (`bignum.zig:198`)
- `bignum.test.bignum absVal` — `panic: @memcpy arguments alias`
- `bytecode_file.test.bytecode round-trip: vector pair bignum rational complex constants` — testing-allocator `Invalid free` canary panic
- `bytecode_file.test.bytecode round-trip: various constant types` — `panic: attempt to unwrap error: OutOfMemory`
- `tests_core_eval.test.eval integer literal` — SIGABRT (even the simplest eval path)

## Breakdown by test file (CI run)

69 tests_ir, 42 tests_advanced, 36 tests_macros, 34 tests_core_eval, 32 tests_continuations, 31 tests_filesystem, 27 tests_io, 25 tests_robustness, 23 tests_libraries, 17 tests_numeric, 16 tests_srfi18, 16 tests_exceptions, 13 tests_tail_calls, 11 tests_fibers, 9 tests_native, 9 tests_derived_forms, 8 tests_records, 7 tests_bytecode_cache, 7 bignum, 6 tests_ffi, 4 tests_deepcopy, 2 bytecode_file — i.e., every area of the suite.

[...]

Issue #1397, opened 2026-07-10
Epic: practical fuzzing roadmap


baijum:
Tracking issue for turning the analysis in `docs/dev/fuzzing-feasibility.md` (merged in #1387, research-grounded in PR #1388) into working fuzzing infrastructure. Issues are ordered by priority and designed to be worked **one at a time** — each lands value on its own, and the plan can stop after any item.

## Phase 1 — leverage what exists (Tier 1)

- [x] #1389 Add encoded Smith seed corpora to the four fuzz targets
- [x] #1390 Add a scheduled bounded fuzzing job to CI (incl. `-Dgc-stress=true` variant)
- [x] #1391 Add a token-vocabulary fuzz target for the reader and eval

After Phase 1: the existing targets actually run continuously, start from meaningful inputs, and get past the lexer.

## Phase 2 — structure-aware generation (Tier 2)

- [x] #1392 Add a Smith-driven grammar generator for valid R7RS forms

This is the core investment: a Zest-style parametric generator (the architecture `std.testing.Smith` natively supports) that produces valid, well-bound, resource-bounded Scheme — the first thing that seriously exercises the compiler, VM, and GC.

## Phase 3 — correctness oracles (Tier 3)

- [x] #1393 Add a switch to disable IR optimization passes *(small enabler, can be done anytime)*
- [x] #1394 Differential fuzz target: optimized vs unoptimized evaluation *(needs #1392 + #1393)*
- [x] #1395 Differential harness: bytecode VM vs LLVM native backend *(needs #1392)*
- [x] #1396 Differential harness: Kaappi vs an external reference Scheme *(needs #1392; last)*

Phase 3 is where wrong-value bugs (not just crashes) surface. Internal oracles (#1394, #1395) come first — zero external setup; the external oracle (#1396) last — most engineering per bug, but catches conformance errors both internal paths agree on.

## Standing rules (from the doc's operating guidance)

- Ordinary Scheme read/compile/runtime errors are **expected** fuzz outcomes; only crashes, panics, leaks, sanitizer findings, and differential mismatches fail a target.
- Every fuzz finding: minimise → readable regression test → keep the encoded input as a corpus entry.
- Generators never emit filesystem, process, FFI, network, or thread forms.
- Keep the 100 ms `vm.eval` deadline and bounded depth/size in all generators.

## Explicitly deferred

[...]

PR #1403, opened 2026-07-10
Add a Smith-driven grammar generator for valid R7RS forms

### LINKED ISSUE #1392: Add a Smith-driven grammar generator for valid R7RS forms
**Fuzzing roadmap — item 4 (Tier 2, the core investment).** From `docs/dev/fuzzing-feasibility.md`. Independent of #1391, but lands best after #1389/#1390 exist to run it.

## Context

To reach the compiler, VM, and GC, the fuzzer must produce semantically deep *valid* Scheme. The published architecture for this is exactly what Zig's API supports: a [Zest-style parametric generator](https://dl.acm.org/doi/10.1145/3293882.3330576) (generator consumes fuzzer-mutated decisions → emits structured input) inside a [NAUTILUS-style](https://github.com/nautilus-fuzz/nautilus) coverage loop. `std.testing.Smith` *is* the decision string.

## Task

1. New module `src/fuzz_gen.zig`: `generateProgram(smith: *std.testing.Smith, allocator) []const u8` that emits a small valid R7RS program, driven entirely by `Smith` choices.
2. Design constraints (each from a documented research lesson — see the doc's research section):
   - **Well-bound programs** (PolyGlot): track in-scope identifiers; references only pick from bound variables, binders push/pop scope. Otherwise most outputs die at "unbound variable".
   - **Bounded resources** (operating guidance): cap expression depth, list/vector/string literal sizes, and iteration counts of generated loops; keep the 100 ms `vm.eval` deadline.
   - **No ambient effects** (operating guidance): never generate filesystem, process, FFI, network, or thread forms.
   - **Form coverage**: weight toward the interesting compiler paths — closures, tail calls, named let/do loops, `call/cc`, `dynamic-wind`, `guard`/`raise`, quasiquote nesting, `syntax-rules` definition+use, string/vector/bytevector mutation (GC write-barrier paths).
3. New target `test "fuzz eval (generated)"` in `src/tests_fuzz.zig` wiring the generator to the existing eval harness. Keep the raw-bytes and token targets for parser robustness.
4. Unit-test the generator itself: for a few thousand fixed-seed runs, every generated program must pass `readDatum` and compile (runtime errors allowed).

## Acceptance criteria

- `zig build test` passes; generated-program target runs under `zig build test --fuzz`.
- Coverage check: `zig build coverage` (or a bounded fuzz run + kcov) shows the generated target reaching `compiler_*.zig`, `vm_dispatch.zig`, and `memory.zig` lines the raw-bytes target does not.
- Generator module stays under the 1500-line policy; split grammar tables out if needed.

## References

- `docs/dev/fuzzing-feasibility.md` — Tier 2, "Zig's Smith is the Zest architecture", "Grammar-based generation is the workhorse".

baijum:
Fuzzing roadmap **Phase 2** (Tier 2, epic #1397). Closes #1392.

## What

A Zest-style parametric generator ([`src/fuzz_gen.zig`](https://github.com/kaappi/kaappi/blob/claude/github-1397-phase-2-32c631/src/fuzz_gen.zig) + [`src/fuzz_gen_data.zig`](https://github.com/kaappi/kaappi/blob/claude/github-1397-phase-2-32c631/src/fuzz_gen_data.zig)) that maps the `std.testing.Smith` decision stream to a small **valid, well-bound, resource-bounded** R7RS program, plus a new `test "fuzz grammar"` target that routes generated programs through the existing sandboxed `evalOne` harness. The raw-bytes and token targets stay for parser robustness.

## Why it matters (measured)

Over equal 30-input sets through the same Debug binary under kcov:

| | raw random bytes | generated programs |
|---|---|---|
| overall `src/` line coverage | 11.1% | **24.3%** |
| `vm_dispatch.zig` | 156 lines | **480 lines** |
| `memory.zig` (GC) | 144 | **255** |
| `compiler_*.zig` (8 files) | 1112 | **2262** |
| `compiler_macro.zig` / `expander.zig` / `compiler_advanced.zig` / `vm_continuations.zig` | **0** each | 264 / 264 / 330 / 99 |

**98% of generated programs evaluate without error** (294/300 fixed seeds; the rest are expected guard re-raises and deadline hits) — vs 76% merely *parsing* for the Phase 1 token target. A threshold test guards this rate against generator regressions.

## Design (each constraint from a documented research lesson)

- **Well-bound** (PolyGlot): a scope stack tracks every identifier with a type-ish `Kind` — numbers, procedures with arity/variadic/bounded-recursion flags, vectors with length + element kind, strings/lists with length + mutability — so references resolve, calls match arities, and every index is in range by construction. A `reserved` kind hides skeleton binders (letrec-bound procedure names) that would otherwise let a generated reference resolve to a shadowing binder of a different type.
- **Bounded by construction**: expression depth, literal sizes, loop iteration counts (smaller when nested), and total program bytes are capped; loop-carried integer accumulators are modulo-clamped so repeated squaring can't build million-digit bignums whose single multiplication outlives the 100 ms deadline.
- **No ambient effects**: filesystem/process/FFI/network/thread forms are never emitted (sandboxed harness stays as backstop).
- **Form coverage** weighted toward the interesting compiler/VM/GC paths: closures, tail calls, named let/do loops, `call/cc` (with guaranteed escapes), `dynamic-wind`, `guard`/`raise`, quasiquote (unquote, splice, nested templates), `syntax-rules` definition + use (fixed and ellipsis rules), and vector/string/bytevector/pair mutation (GC write-barrier paths).

The generator is driven through a `Chooser`: Smith under the fuzzer, a seeded PRNG in unit tests (Smith replays out-of-range decisions as the range minimum, so a PRNG is what gives fixed-seed tests real variety).

## Tests

- 2000 fixed-seed programs all **parse and compile** (runtime errors allowed per the standing rules), stay under the size bound; generation is deterministic per seed.
- Eval-rate gate (≥75% of 60 seeds evaluate cleanly; measured 98%) — skipped under `-Dgc-stress` where the deadline would measure GC overhead instead of generator quality (and gc-stress is blocked on #1401 anyway).
- `zig build test` passes; `zig build test --fuzz=200` completes with no `.zig-cache/f/crash` marker and live corpora for all six targets. The CI fuzz job picks the target up automatically (it lives in `tests_fuzz.zig`).

## Notes

- Split into two files along the grammar-domain seam for the 1500-line policy (1447 + 648), with data-kind generators re-exported as `Gen` methods.
- Runbook updated: new target row + how to reproduce/minimise a grammar-target crash (replay the decision stream verbatim as a corpus entry, recover the program text, then keep the *source* as an eval-target seed — decision streams are not stable across generator changes).
- No seed corpus needed: any decision stream decodes to a valid program.

### DIFF
--- a/src/fuzz_gen.zig
+++ b/src/fuzz_gen.zig
@@ -1,0 +1,1447 @@
+//! Smith-driven grammar generator for valid R7RS programs (fuzzing Tier 2,
+//! issue #1392).
+//!
+//! A Zest-style parametric generator: every structural decision is a
+//! `std.testing.Smith` choice, so the fuzzer mutates the decision stream and
+//! coverage feedback turns byte mutations into structural program mutations
+//! (docs/dev/fuzzing-feasibility.md, "Zig's Smith is the Zest architecture").
+//! Each design constraint comes from a documented research lesson or the
+//! doc's operating guidance:
+//!
+//! - **Well-bound** (PolyGlot): a scope stack tracks every identifier with a
+//!   type-ish `Kind`; references pick only from visible innermost bindings,
+//!   calls match arities, and indices stay in range, so programs exercise the
+//!   VM instead of dying at "unbound variable".
+//! - **Bounded by construction**: expression depth, literal sizes, loop
+//!   iteration counts, and total program bytes are capped. Loop-carried
+//!   integer accumulators are modulo-clamped so no program degenerates into
+//!   million-digit bignum arithmetic that a single primitive call could ride
+//!   past the 100 ms VM deadline.
+//! - **No ambient effects**: filesystem, process, FFI, network, and thread
+//!   forms are never emitted (the sandboxed eval harness enforces the same
+//!   at runtime as a second line of defense).
+//! - **Form coverage** weighted toward the interesting compiler/VM/GC paths:
+//!   closures, tail calls, named let / do loops, call/cc, dynamic-wind,
+//!   guard/raise, quasiquote, syntax-rules definition + use, and
+//!   vector/string/bytevector mutation (GC write-barrier paths).
+//!
+//! Ordinary Scheme runtime errors remain expected fuzz outcomes, but most
+//! generated programs evaluate cleanly (asserted by a test in
+//! tests_fuzz.zig); every program parses and compiles (asserted below).
+
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+
+pub const Error = error{ WriteFailed, OutOfMemory };
+
+// ---------------------------------------------------------------------------
+// Bounds — everything the generator emits is capped by construction.
+// ---------------------------------------------------------------------------
+
+const max_depth: u32 = 5;
+/// Once the program exceeds this many bytes, all further expressions are
+/// forced to leaves. The remaining open forms can only add a bounded tail,
+/// so `expected_max_bytes` holds with a wide margin (unit-tested).
+const soft_budget: usize = 3072;
+pub const expected_max_bytes: usize = 8192;
+/// Prime clamp for loop-carried accumulators: `(modulo acc 999983)` keeps
+/// repeated squaring inside loops from building huge bignums.
+pub const acc_modulus: u32 = 999983;
+const max_outer_iters: u32 = 16;
+const max_inner_iters: u32 = 5;
+
+// ---------------------------------------------------------------------------
+// Static vocabulary. All names live in fixed pools so bindings can be
+// compared and re-emitted without allocation. Reserved names that never
+// appear in the general pool: lp (loops), k (continuations), ex (guard),
+// p/q/r (syntax-rules pattern variables), g*/f*/m* (top-level definitions).
+// ---------------------------------------------------------------------------
+
+const local_names = [_][]const u8{ "a", "b", "c", "d", "e", "h", "u", "v" };
+const global_names = [_][]const u8{ "g0", "g1", "g2" };
+const fn_names = [_][]const u8{ "f0", "f1" };
+const macro_names = [_][]const u8{ "m0", "m1" };
+const pattern_names = [_][]const u8{ "p", "q", "r" };
+
+pub const symbol_names = [_][]const u8{ "alpha", "beta", "boom", "zed" };
+
+const arith_ops = [_][]const u8{ "+", "-", "*", "min", "max" };
+const div_ops = [_][]const u8{ "quotient", "remainder", "modulo" };
+const cmp_ops = [_][]const u8{ "<", "<=", "=", ">", ">=" };
+const rec_ops = [_][]const u8{ "+", "*", "max" };
+const fold_ops = [_][]const u8{ "+", "*", "max", "min" };
+pub const real_round_ops = [_][]const u8{ "round", "floor", "truncate", "ceiling" };
+const guard_preds = [_][]const u8{ "number?", "symbol?", "string?", "pair?" };
+const any_preds = [_][]const u8{ "number?", "string?", "boolean?", "procedure?", "symbol?", "char?", "vector?", "list?" };
+
+// ---------------------------------------------------------------------------
+// Decision source: Smith under the fuzzer, a seeded PRNG in unit tests.
+// Smith replays out-of-range decisions as the range minimum, so a PRNG is
+// what gives the fixed-seed tests real variety.
+// ---------------------------------------------------------------------------
+
+/// Coarse per-decision-point identifiers; they become stable Smith UIDs so
+/// the fuzzer can attribute coverage to individual grammar decisions.
+const Tag = enum(u32) {
+    count,
+    top_kind,
+    depth_pick,
+    bind_kind,
+    arity,
+    shape,
+    op_pick,
+    name_pick,
+    var_pick,
+    lit_pick,
+    iters,
+    nargs,
+    idx_kind,
+    idx_pick,
+    coin,
+    len_pick,
+};
+
+const Chooser = union(enum) {
+    smith: *std.testing.Smith,
+    random: std.Random,
+
+    pub fn range(c: *Chooser, comptime tag: Tag, lo: u32, hi: u32) u32 {
+        std.debug.assert(lo <= hi);
+        if (lo == hi) return lo;
+        return switch (c.*) {
+            .smith => |s| s.valueRangeAtMostWithHash(u32, lo, hi, @intFromEnum(tag) *% 0x9e3779b9),
+            .random => |r| r.intRangeAtMost(u32, lo, hi),
+        };
+    }
+
+    pub fn index(c: *Chooser, comptime tag: Tag, len: usize) usize {
+        return c.range(tag, 0, @intCast(len - 1));
+    }
+
+    /// True with probability num/den.
+    pub fn chance(c: *Chooser, comptime tag: Tag, num: u32, den: u32) bool {
+        return c.range(tag, 1, den) <= num;
+    }
+};
+
+// ---------------------------------------------------------------------------
+// Kind model: enough type/shape information to keep generated programs
+// well-typed, well-arity'd, and in-bounds.
+// ---------------------------------------------------------------------------
+
+pub const Len = union(enum) {
+    exact: u16,
+    at_least_one,
+    unknown,
+
+    pub fn nonEmpty(l: Len) bool {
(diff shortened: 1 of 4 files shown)


src/fuzz_gen_data.zig
```diff
@@ -0,0 +1,653 @@
...
+            // always safe regardless of the tail's provenance.
+            return .{ .len = tail.len.plusOne(), .mutable = true, .ints = tail.ints };
+        },
+        .cdr_op => {
+            const b = g.pickCdrVar(need).?;
+            const info = b.kind.list;
+            try g.emitf("(cdr {s})", .{b.name});
+            return .{
+                .len = .{ .exact = info.len.exactLen().? - 1 },
+                .mutable = info.mutable,
```
baijum: **Bug (non-blocking): `cdr` inherits mutability from a `cons` result whose tail may be immutable.**

`cons_op` always returns `mutable = true` (fresh head cell), then `cdr_op` copies `info.mutable`. That marks `(cdr (cons x immutable-list))` as mutable even though the cdr *is* the immutable spine.

Reproduced outside the generator:

```scheme
(let ((a '(1 2 3)))
  (let ((b (cons 0 a)))
    (let ((c (cdr b)))
      (set-car! c 9))))
; → type error in 'set-car!': expected mutable pair
```

Kaappi treats quoted lists as immutable, and `genStmt`'s `setcar` arm trusts `bindIsListMutNonEmpty`, so this is a reachable well-boundness hole. It surfaces as an ordinary Scheme error (fine for the fuzz target), but it undercuts the "programs evaluate cleanly by construction" claim.

Conservative fix: have `cdr_op` return `mutable = false`, or track head-vs-spine mutability more precisely. Happy to see this as a follow-up rather than a blocker.

src/fuzz_gen.zig
```diff
@@ -0,0 +1,1447 @@
...
+
+        pub fn add(self: *@This(), op: E, weight: u32) void {
+            self.ops[self.n] = op;
+            self.weights[self.n] = weight;
+            self.n += 1;
+            self.total += weight;
+        }
+
+        pub fn pick(self: *@This(), ch: *Chooser) E {
+            var r = ch.range(.op_pick, 0, self.total - 1);
```
baijum, line 312: **Nit (agree with CodeRabbit): assert capacity / non-empty before pick.**

`genInt` can push ~36 candidates into a 40-slot buffer — fine today, but a silent overflow on `add` or `total - 1` underflow if `total == 0` would be a nasty fuzz crash. A `std.debug.assert(self.n < self.ops.len)` in `add` and `std.debug.assert(self.total > 0)` in `pick` would make the invariant obvious.

src/tests_fuzz.zig
```diff
@@ -367,3 +375,57 @@ test "fuzz tokens" {
...
+        }
+        total += 1;
+    }
+    // Measured rate is 98% over 300 seeds (294/300; the rest are expected
+    // outcomes — guard re-raises and deadline hits on loop-heavy programs).
+    // The threshold sits far below that because deadline outcomes are
+    // load-sensitive: a real generator regression tanks the rate, CI jitter
+    // must not.
+    try std.testing.expect(ok * 100 >= total * 75);
+}
```
baijum, line 434: **Nit:** The comment cites a 300-seed measurement (98%) while the gate samples 60 seeds. Worth either bumping the loop to 300 (cheap relative to the 2000-seed compile test) or rewording to "measured offline over 300 seeds; gate uses 60." As written, a future reader may think the in-test sample is 300.

src/fuzz_gen.zig
```diff
@@ -0,0 +1,1447 @@
...
+
+const Chooser = union(enum) {
+    smith: *std.testing.Smith,
+    random: std.Random,
+
+    pub fn range(c: *Chooser, comptime tag: Tag, lo: u32, hi: u32) u32 {
+        std.debug.assert(lo <= hi);
+        if (lo == hi) return lo;
+        return switch (c.*) {
+            .smith => |s| s.valueRangeAtMostWithHash(u32, lo, hi, @intFromEnum(tag) *% 0x9e3779b9),
```
baijum, line 112: **Observation (non-blocking):** Smith UIDs are only as fine-grained as the `Tag` enum (~16 values), so every `op_pick` / `coin` site shares one hash. That is a reasonable Phase 2 choice and still gives structural mutations, but if coverage plateaus later, mixing in a call-site salt (or a monotonic decision counter) would give the fuzzer more independent knobs without changing the grammar.

baijum (commented):
## Review summary

Solid Phase 2 delivery. The Zest/Smith generator design is clear, the well-bound Kind model is thoughtfully constrained (reserved binders, bounded_first, acc_modulus, soft budget), the file split respects the 1500-line policy, and the test gates (2000-seed parse/compile, determinism, ≥75% eval rate with gc-stress skip) match the claims. Docs/runbook updates are good. CI is green across platforms.

**Ready to merge**, with one follow-up worth filing (or fixing in a quick follow-up PR): list mutability is over-approximated through `cdr` of a `cons` onto an immutable spine, so `set-car!` can fail at runtime on programs the Kind model marks mutable. That is an expected Scheme error today (not a crash), and the eval-rate gate still has plenty of headroom, so it should not block this PR.

I also agree with CodeRabbit's two nits (Cands capacity assert; clarify 300-seed measurement vs 60-seed gate).

src/fuzz_gen_data.zig
baijum: Fixed in a68c7323 — and thanks for the clean repro; it's exactly the hole. `ListInfo` now carries a three-state `Mut { none, head, all }` instead of a head-only flag: constructors that allocate every cell (`list`, `reverse`, `map`, `vector->list`, rest-args) are `.all`; quoted data and quasiquote templates `.none`; `cons` is `.all` only when its tail is (else `.head`); `cdr` keeps `.all` or drops to `.none`. Your `(set-car! (cdr (cons 0 '(1 2 3))) 9)` pattern is now unreachable by construction. As a bonus, `append` got more precise instead of blanket-immutable: it models the R7RS 6.4 copy-all-but-last rule (`.all` iff the shared tail is, `.head` when the first list is known non-empty). Re-measured after the change: still 294/300, same six misses — all deadline or guard-re-raise outcomes, so this wasn't costing measurable rate, but the model claim is honest now.

src/fuzz_gen.zig
baijum: Done in a68c7323: `std.debug.assert(self.n < self.ops.len)` in `add` and `std.debug.assert(self.total > 0)` in `pick`.

src/tests_fuzz.zig
baijum: Reworded in a68c7323 to say explicitly that 294/300 is an offline 300-seed measurement and the gate samples the first 60 seeds. I kept 60 rather than bumping to 300: each eval builds a full VM (~20–50 ms in ReleaseSafe, much more in Debug), and the Debug CI job already ran at 20m31s of its 30-minute budget — 240 extra Debug evals with deadline-bound loop programs felt like asking for a flaky timeout for no extra signal, since the 75% threshold is calibrated against jitter, not sample size.

src/fuzz_gen.zig
baijum: Agreed on all points — coarse tags were a deliberate Phase 2 simplification (the Zest loop mutates the decision stream positionally either way, so structural mutation survives shared UIDs). Noted the call-site-salt / decision-counter idea as the first knob to try if grammar-target coverage plateaus; it slots in cleanly since every decision already flows through the `Chooser` methods.

src/fuzz_gen.zig
```diff
@@ -0,0 +1,1447 @@
...
+        pub fn pick(self: *@This(), ch: *Chooser) E {
+            var r = ch.range(.op_pick, 0, self.total - 1);
+            for (self.ops[0..self.n], self.weights[0..self.n]) |op, wt| {
+                if (r < wt) return op;
+                r -= wt;
+            }
+            unreachable;
+        }
+    };
+}
```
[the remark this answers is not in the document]
baijum: Added in a68c7323 (plus a `total > 0` assert in `pick` per the human review).

src/tests_fuzz.zig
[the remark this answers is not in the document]
baijum: Reworded in a68c7323: the comment now states 294/300 is an offline 300-seed measurement and the gate samples 60 seeds for CI cost (the Debug job runs close to its time budget).

Issue #1396, opened 2026-07-10
Differential harness: Kaappi vs an external reference Scheme


baijum:
**Fuzzing roadmap — item 8 (Tier 3, last).** From `docs/dev/fuzzing-feasibility.md`. Depends on #1392 (generator). Do after the internal oracles (#1394, #1395) — those need zero setup and catch Kaappi-vs-Kaappi bugs first; this one catches conformance bugs where *both* Kaappi paths agree but are wrong.

## Context

Csmith-style differential testing against an independent implementation is the highest-value oracle for conformance bugs. Most of the engineering, per the literature, is generating only the **portable, fully-specified subset** — R7RS leaves evaluation order, error objects, exactness edges, and more unspecified.

**Note:** the repo's `(chibi test)` is an API shim implemented in Kaappi, **not** real Chibi. A genuine external interpreter must be installed as the oracle. Pin **one** oracle at a fixed version and invocation: Chibi Scheme is the natural pick (closest R7RS-small alignment, small install). Dialect defaults differ across implementations — Guile, for instance, needs R7RS mode selected explicitly — so an unpinned or swappable oracle produces false diffs across environments. Gauche can serve as a second opinion later, as its own pinned configuration.

## Task

1. Add a "portable subset" mode to `fuzz_gen` (#1392):
   - deterministic results only: no unspecified-return-value forms in value position, no evaluation-order-dependent effects (single side effect per argument list at most — the effect-discipline idea from Midtgaard et al.), no implementation-specific procedures, no flonum printing edge cases unless normalized;
   - error behavior compared only as class ("raises" vs "returns"), never message text.
2. Harness script `tests/fuzz/oracle-diff.sh`: emit N programs, run through `kaappi` and the pinned `chibi-scheme` (fixed invocation flags, per-program timeout; record the oracle version in the results dir), compare normalized (value printed with `write`, stdout, exit class). Divergences land in a results dir.
3. Triage protocol in the script header: a divergence is (a) Kaappi bug, (b) oracle bug, or (c) generator leaking unspecified behavior — check against the R7RS spec (`docs/errata-corrected-r7rs.pdf`) before filing.
4. Scheduled CI job (with `chibi-scheme` installed via apt/brew) or documented local-only run — decide based on flakiness after a trial period.

## Acceptance criteria

- `bash tests/fuzz/oracle-diff.sh 100` runs clean on `main` against Chibi, or every divergence is triaged into a filed bug / generator fix.
[...]

Issue #1395, opened 2026-07-10
Differential harness: bytecode VM vs LLVM native backend


baijum:
**Fuzzing roadmap — item 7 (Tier 3).** From `docs/dev/fuzzing-feasibility.md`. Depends on #1392 (generator); independent of #1393/#1394.

## Context

Kaappi has two implementations of Scheme evaluation in one repo: the bytecode VM and the LLVM native backend. Diffing them on the same program is exactly the setup [Midtgaard et al. (ICFP 2017)](https://dl.acm.org/doi/10.1145/3110259) used to find disagreements between OCaml's bytecode and native backends in 18 of 20 runs. Known regressions in this area already exist (#1376), so the oracle has demonstrated bug surface.

**Key constraint:** the native backend falls back to `kaappi_eval` for forms it doesn't compile natively (letrec, named let, do, ...). Generated programs must stick to natively-compiled forms — arithmetic, comparisons, `if`/`and`/`or`/`when`/`unless`, `let`/`let*`, lambda (incl. variadic), self-tail-calls, tail calls between native functions — or the diff degenerates into VM-vs-VM.

## Task

1. Add a "native-compilable subset" mode to `fuzz_gen` (#1392) restricted to the forms above.
2. Harness script `tests/fuzz/native-diff.sh` (or a small Zig driver):
   - emit N generated programs (fixed seed → reproducible);
   - for each: run via `kaappi program.scm`, and via `kaappi compile program.scm -o bin && ./bin`;
   - compare normalized stdout + exit class; write divergent programs to a results dir and exit nonzero.
3. This is process-spawn + link per input — orders of magnitude slower than in-process fuzzing. Run it as a scheduled batch (N in the hundreds) in the CI fuzz workflow (#1390), not as a `std.testing.fuzz` target.
4. Minimise + file an issue per divergence; add each as a regression test.

## Acceptance criteria

- `bash tests/fuzz/native-diff.sh 100` runs clean on `main` (or files real bugs — also success).
- Reproducibility: same seed → same programs.
- CI job runs it on schedule with artifacts for divergences.
- Uses `zig cc` (never bare `clang`) for linking, per repo convention — `kaappi compile` already handles this.

## References

- `docs/dev/fuzzing-feasibility.md` — Tier 3 ("Kaappi vs itself"), research section (Midtgaard et al., FuzzJIT).
- Known native-backend divergences: #1376.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

Issue #1390, opened 2026-07-10
Add a scheduled bounded fuzzing job to CI


baijum:
**Fuzzing roadmap — item 2 (Tier 1).** From `docs/dev/fuzzing-feasibility.md`. Best after #1389 so runs start from a corpus.

## Context

The fuzz targets in `src/tests_fuzz.zig` are never run in fuzzing mode: no `--fuzz` invocation exists in `build.zig` or `.github/workflows/`. Under plain `zig build test` each target executes once with a fixed seed — effectively a smoke test.

## Task

1. New workflow `.github/workflows/fuzz.yml`:
   - `schedule:` (nightly is fine; pick a time away from the 05:17 UTC ecosystem nightly) + `workflow_dispatch:`.
   - Runs a **bounded** fuzz pass — `zig build test --fuzz=<iteration-limit>` — sized to finish in well under an hour.
   - Two matrix variants: default build, and `-Dgc-stress=true` (attempts collection on every allocation — skipped only while `no_collect` is held or the GC is disabled; converts latent rooting bugs into immediate failures; see "Operating guidance" in the doc).
   - On failure, upload the crashing input / fuzz artifacts and fail the run.
2. Isolation per the doc's operating guidance: no filesystem/process/FFI/network access needed by the job beyond the build itself.
3. Document the failure workflow in `docs/dev/fuzzing-feasibility.md` or a short `docs/dev/fuzzing.md` runbook: minimise the input → add a readable regression test in `src/tests_*.zig` or `tests/scheme/` → add the encoded input to the corpus from #1389.

## Acceptance criteria

- Workflow runs green on `workflow_dispatch` with a small iteration limit.
- A deliberately planted panic (local test) is caught and surfaces the failing input as an artifact.
- Runbook section exists for turning a fuzz failure into a regression test + corpus entry.

## References

- `docs/dev/fuzzing-feasibility.md` — Gaps ("Not run in CI"), Tier 1, Operating guidance.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

baijum:
Tracked in epic #1397.

PR #1405, opened 2026-07-10
Fuzzing roadmap Phase 3: IR optimization switch + opt-vs-no-opt differential oracle

### LINKED ISSUE #1393: Add a switch to disable IR optimization passes
**Fuzzing roadmap — item 5 (Tier 3 enabler; small, standalone).** From `docs/dev/fuzzing-feasibility.md`.

## Context

`lowerAndOptimize` (`src/ir.zig:571`) unconditionally runs the five optimization passes (constant folding, dead-branch elimination, boolean simplification, identity elimination, begin simplification). There is no way to compile without them, so we cannot diff optimized vs unoptimized execution — the oracle that found GHC optimizer bugs within ~20k random tests ([Pałka et al.](https://dl.acm.org/doi/10.1145/1982595.1982615)) and that FuzzJIT turned into an always-on in-process check.

## Task

1. Add an `optimize: bool = true` field on `IR` (or a parameter threaded from `Compiler`), skipping the five passes in `lowerAndOptimize` when false. `markTailPositions` must still run — it is analysis, not optimization.
2. Expose it two ways:
   - programmatically (for the differential fuzz target, #1394), e.g. a `VM`/`Compiler` option;
   - CLI flag `--no-ir-opt` in `main.zig` (also useful for debugging miscompilation reports and for `--disassemble` comparisons).
3. Note the flag in the CLI `--help` text and in `CLAUDE.md`'s flag list.

## Acceptance criteria

- Regression test: a program exercising each folded pattern evaluates to the same value with the flag on and off.
- `bash tests/scheme/run-all.sh` passes with `--no-ir-opt` (run locally at least once; catching a divergence here is a win, not a blocker for the flag itself — file it as a bug).
- `--disassemble` on `(if #t 1 2)` shows different bytecode with vs without the flag (proves the switch is live).
- **Watch out:** delete stale `.sbc` caches when testing — cache keys don't include the flag today; decide whether no-opt compilations should skip the cache entirely (simplest: yes).

## References

- `docs/dev/fuzzing-feasibility.md` — Tier 3 ("Kaappi vs itself"), research section (Pałka et al., FuzzJIT).

### LINKED ISSUE #1394: Differential fuzz target: optimized vs unoptimized evaluation
**Fuzzing roadmap — item 6 (Tier 3, first correctness oracle).** From `docs/dev/fuzzing-feasibility.md`. Depends on #1392 (generator) and #1393 (opt switch).

## Context

Crash-only fuzzing never surfaces silently-wrong values — which the EMI authors found to be the majority and most pernicious class of compiler bugs. The cheapest correctness oracle needs no external tool: evaluate the same generated program with IR optimizations on and off, and any divergence is a bug in an optimization pass (or in the baseline). This is FuzzJIT's in-process oracle pattern applied to Kaappi.

## Task

1. New target `test "fuzz differential (opt vs no-opt)"` in `src/tests_fuzz.zig`:
   - generate a program via `fuzz_gen` (#1392);
   - evaluate twice in fresh VMs — optimizations enabled and disabled (#1393);
   - compare a **normalized observable**: printed result via `printer.zig` write-mode, plus error class (no error / read error / compile error / runtime error). Do not compare error message text.
   - divergence → `return error.DifferentialMismatch` (a fuzz failure); matching results (including matching errors) → pass.
2. Generator constraints already exclude ambient effects; additionally exclude anything observably nondeterministic (`current-jiffy`, `current-second`, `random`, eq?-identity-of-literals edge cases) — add an exclusion knob to `fuzz_gen` if needed.
3. Every divergence found gets minimised into a Scheme regression test before the fix lands (per repo policy: every bug fix needs a regression test).

## Acceptance criteria

- Target runs under `zig build test --fuzz` and passes a bounded run on current `main`.
- Planted-bug check: locally hack a constant-folding case to be wrong; the target catches it within a bounded run.
- Wired into the CI fuzz job (#1390) automatically via `tests_fuzz.zig`.

## References

- `docs/dev/fuzzing-feasibility.md` — Tier 3 ("Kaappi vs itself"), research section (Pałka et al., FuzzJIT, EMI).

baijum:
First half of Phase 3 of the fuzzing roadmap (#1397): the Tier 3 enabler and the first correctness oracle. The remaining Phase 3 items (#1395 native-backend diff, #1396 external-reference diff) are separate follow-ups per the epic's one-at-a-time design.

## #1393 — `--no-ir-opt` switch

- `ir.optimize_enabled` (threadlocal, matching the `vm_instance` pattern) gates the five optimization passes **and** AST-level constant folding in `tryFoldFromAST`; `markTailPositions` always runs (analysis, not optimization).
- CLI flag `--no-ir-opt`, listed in `--help` and CLAUDE.md.
- No-opt runs skip the `.sbc` cache entirely (keys don't include the flag); `--no-ir-opt --compile` without `-o` is refused since the default output path is the cache location.

**Acceptance:** `--disassemble` on `(if #t 1 2)` → single `load_const` optimized vs full test/branch/arms unoptimized; regression tests cover one case per folded pattern; full Scheme suite under `--no-ir-opt`: **1821 pass, 0 fail** (stale `.sbc` caches deleted first — the flag makes them unreachable anyway).

## #1394 — `fuzz differential (opt vs no-opt)` target

- Generates a program via `fuzz_gen` (#1392), evaluates it in two fresh sandboxed VMs — optimizations on and off — and compares a normalized observable: printed final value (write mode) **plus the generator's globals `g0`–`g2`**, and the error class (never message text). Divergence → `error.DifferentialMismatch`, with the program and both observables printed to stderr for triage.
- Timeout / OOM / stack-overflow outcomes make a pair incomparable (skip) rather than a divergence — the two paths legitimately do different amounts of work.
- The globals widening is load-bearing: `vm.eval` returns only the last top-level value, so a wrong fold inside `(define g1 …)` is invisible in the final value alone. Planted-bug check: an off-by-one in the `*` constant fold is caught at fixed seed 30 (3 more within 500 seeds); without the globals it went undetected across all 500.
- A 60-seed deterministic gate runs on every `zig build test`; the CI fuzz job (#1390) picks the target up automatically. The generator emits nothing observably nondeterministic (no time/random/I-O forms; `eq?` only on interned symbols), so no exclusion knob was needed.

## Validation

- `zig build test`: 703/703
- `bash tests/scheme/run-all.sh` with `--no-ir-opt`: 1821 pass, 0 fail, 0 timeouts
- `zig build test --fuzz=200`: exit 0, no `.zig-cache/f/crash` marker, all 7 target corpora updated
- `zig fmt --check`: clean

Closes #1393
Closes #1394

### DIFF
--- a/src/ir.zig
+++ b/src/ir.zig
@@ -568,6 +568,15 @@ pub fn lower(irn: *IR, expr: Value) CompileError!*Node {
     return lowerWithMacros(irn, expr, null);
 }
 
+/// Master switch for IR optimization: when false, `lowerAndOptimize` skips
+/// the five optimization passes and `tryFoldFromAST` stops folding during
+/// lowering. Analysis (`markTailPositions`) always runs — it is required for
+/// correctness, not an optimization. Threadlocal (like `vm_instance`) so
+/// SRFI-18 child threads keep the default regardless of the parent's setting.
+/// Toggled by the `--no-ir-opt` CLI flag and by the differential fuzz oracle
+/// in tests_fuzz.zig (#1393).
+pub threadlocal var optimize_enabled: bool = true;
+
 pub fn lowerAndOptimize(
     ir_instance: *IR,
     expr: Value,
@@ -576,6 +585,7 @@ pub fn lowerAndOptimize(
 ) CompileError!*Node {
     var node = try lowerWithMacros(ir_instance, expr, macros);
     markTailPositions(node, is_tail);
+    if (!optimize_enabled) return node;
     node = foldConstants(ir_instance, node);
     node = eliminateDeadBranches(ir_instance, node);
     node = simplifyBooleans(ir_instance, node);
@@ -748,6 +758,7 @@ fn lowerCall(ir: *IR, expr: Value, macros: ?*std.StringHashMap(Value)) CompileEr
 }
 
 fn tryFoldFromAST(ir: *IR, expr: Value) ?*Node {
+    if (!optimize_enabled) return null;
     const operator = types.car(expr);
     if (!types.isSymbol(operator)) return null;
     const name = types.symbolName(operator);
--- a/src/main.zig
+++ b/src/main.zig
@@ -409,11 +409,18 @@ fn mainImpl(init: std.process.Init.Minimal) !void {
         }
     } else if (opts.compile_mode) {
         if (opts.file_path) |fp| {
-            // Without -o, --compile writes to the cache location that plain
-            // runs load from; don't let it poison the cache with
-            // unoptimized bytecode (cache keys don't include the flag).
-            if (opts.no_ir_opt and opts.compile_output == null) {
-                usageError("--no-ir-opt with --compile requires -o <file> (the default output path is the bytecode cache)\n");
+            // --no-ir-opt must not write unoptimized bytecode where plain
+            // runs load their cache from (cache keys don't include the
+            // flag). That location is both the default output (no -o) and
+            // the natural explicit choice (-o program.sbc), so refuse both.
+            if (opts.no_ir_opt) {
+                const clobbers = if (opts.compile_output) |op|
+                    outputIsBytecodeCache(allocator, fp, op)
+                else
+                    true;
+                if (clobbers) {
+                    usageError("--no-ir-opt --compile requires -o <file> that is not the source's .sbc cache path (plain runs would load the unoptimized bytecode as their cache)\n");
+                }
             }
             try compileFile(vm, fp, opts.compile_output);
         } else {
@@ -442,6 +449,30 @@ fn mainImpl(init: std.process.Init.Minimal) !void {
 
 fn getSbcPath(allocator: std.mem.Allocator, scm_path: []const u8) ![]u8 {
     return bytecode_file.getSbcPath(allocator, scm_path);
+}
+
+/// True when `output_path` is the `.sbc` cache location that plain runs of
+/// `src_path` load from. Lexical comparison after normalizing `.`/`..`
+/// segments — symlinks and absolute-vs-relative aliases of the same file are
+/// not detected, but the natural spellings (`prog.sbc`, `./prog.sbc`) are.
+fn outputIsBytecodeCache(allocator: std.mem.Allocator, src_path: []const u8, output_path: []const u8) bool {
+    const cache_path = getSbcPath(allocator, src_path) catch return false;
+    defer allocator.free(cache_path);
+    const out_norm = std.fs.path.resolve(allocator, &.{output_path}) catch return false;
+    defer allocator.free(out_norm);
+    const cache_norm = std.fs.path.resolve(allocator, &.{cache_path}) catch return false;
+    defer allocator.free(cache_norm);
+    return std.mem.eql(u8, out_norm, cache_norm);
+}
+
+test "outputIsBytecodeCache detects cache-path collisions" {
+    const gpa = std.testing.allocator;
+    try std.testing.expect(outputIsBytecodeCache(gpa, "prog.scm", "prog.sbc"));
+    try std.testing.expect(outputIsBytecodeCache(gpa, "prog.scm", "./prog.sbc"));
+    try std.testing.expect(outputIsBytecodeCache(gpa, "dir/prog.scm", "dir/prog.sbc"));
+    try std.testing.expect(outputIsBytecodeCache(gpa, "dir/prog.scm", "dir/../dir/prog.sbc"));
+    try std.testing.expect(!outputIsBytecodeCache(gpa, "prog.scm", "prog.noopt.sbc"));
+    try std.testing.expect(!outputIsBytecodeCache(gpa, "prog.scm", "out/prog.sbc"));
 }
 
 fn runFile(vm: *vm_mod.VM, path: []const u8) !void {
--- a/src/main.zig
+++ b/src/main.zig
@@ -207,6 +207,7 @@ fn mainImpl(init: std.process.Init.Minimal) !void {
     if (opts.action == .exit_ok) return;
 
     // Apply parsed options to VM/GC
+    if (opts.no_ir_opt) ir_mod.optimize_enabled = false;
     if (opts.timeout_ms) |ms| {
         const clockNs = @import("vm_calls.zig").clockNs;
         vm.timeout_deadline_ns = clockNs() + ms * 1_000_000;
@@ -408,6 +409,12 @@ fn mainImpl(init: std.process.Init.Minimal) !void {
         }
     } else if (opts.compile_mode) {
         if (opts.file_path) |fp| {
+            // Without -o, --compile writes to the cache location that plain
+            // runs load from; don't let it poison the cache with
+            // unoptimized bytecode (cache keys don't include the flag).
+            if (opts.no_ir_opt and opts.compile_output == null) {
+                usageError("--no-ir-opt with --compile requires -o <file> (the default output path is the bytecode cache)\n");
+            }
             try compileFile(vm, fp, opts.compile_output);
         } else {
             usageError("Usage: kaappi --compile <file.scm> [-o output.sbc]\n");
@@ -453,8 +460,11 @@ fn runFile(vm: *vm_mod.VM, path: []const u8) !void {
 
     const source_hash = bytecode_file.sourceHash(source);
 
-    // Try loading cached bytecode (skip in sandbox mode — no filesystem side effects)
-    const sbc_path = if (vm.sandbox_mode) null else getSbcPath(allocator, path) catch null;
+    // Try loading cached bytecode (skip in sandbox mode — no filesystem side
+    // effects — and under --no-ir-opt: cache keys don't include the flag, so
+    // a no-opt run must neither reuse optimized bytecode nor poison the cache
+    // with unoptimized bytecode).
+    const sbc_path = if (vm.sandbox_mode or !ir_mod.optimize_enabled) null else getSbcPath(allocator, path) catch null;
     defer if (sbc_path) |p| allocator.free(p);
 
     if (sbc_path) |sp| {
--- a/src/tests_fuzz.zig
+++ b/src/tests_fuzz.zig
@@ -8,6 +8,8 @@ const primitives = @import("primitives.zig");
 const library = @import("library.zig");
 const types = @import("types.zig");
 const fuzz_gen = @import("fuzz_gen.zig");
+const ir_mod = @import("ir.zig");
+const printer = @import("printer.zig");
 
 const Context = @TypeOf(.{});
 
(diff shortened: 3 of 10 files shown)


baijum (commented):
## Review summary

Solid Phase 3 slice: the `--no-ir-opt` switch is cleanly gated (threadlocal, analysis still runs, AST fold + five passes skipped), cache is correctly bypassed on the run path, and the differential oracle is thoughtfully designed — especially widening the observable with `g0`–`g2` (the planted-bug note at seed 30 is convincing).

Docs, CLI help/parse test, IR switch tests, fixed-seed gate, and fuzz.yml/target-count updates all line up with #1393/#1394.

### One real footgun (see inline)
The `--compile` guard refuses missing `-o`, but `getSbcPath("foo.scm")` is exactly `foo.sbc`. So the most natural forced `-o` still poisons the cache that later optimized runs load. Same rationale as the existing guard — worth tightening before or right after merge.

### Nits (non-blocking)
- CodeRabbit’s `th.TestContext` suggestion is style-only; fine to skip.
- A dedicated no-opt tail-call smoke test would document that `markTailPositions` stays live, but the code already does the right thing.
- No unit test for the `--no-ir-opt --compile` without `-o` refusal (behavior is clear from `usageError`).

### Merge readiness
**Almost — approve once CI is green, ideally after the `-o`/cache-path guard is tightened.** Nothing else blocks. The fuzz oracle and switch look correct; I would not hold the PR for the nits alone.

(CI was still running platform test jobs when this review was posted.)

src/main.zig
```diff
@@ -408,6 +409,12 @@ fn mainImpl(init: std.process.Init.Minimal) !void {
         }
     } else if (opts.compile_mode) {
         if (opts.file_path) |fp| {
+            // Without -o, --compile writes to the cache location that plain
+            // runs load from; don't let it poison the cache with
+            // unoptimized bytecode (cache keys don't include the flag).
+            if (opts.no_ir_opt and opts.compile_output == null) {
+                usageError("--no-ir-opt with --compile requires -o <file> (the default output path is the bytecode cache)\n");
+            }
```
baijum, line 424: **Cache poison hole:** this only rejects missing `-o`, but `getSbcPath("program.scm")` is `program.sbc`. So:

```bash
kaappi --no-ir-opt --compile -o program.sbc program.scm
kaappi program.scm   # loads unoptimized .sbc as if it were the normal cache
```

is exactly the failure mode the comment above is trying to prevent — and `-o program.sbc` is the path users will pick when forced to pass `-o`.

Suggestion: also refuse when the resolved output path equals `getSbcPath(fp)` (string-compare the allocated cache path), or write no-opt artifacts to a distinct suffix (e.g. `.noopt.sbc`) by default. Same rationale as this guard.

src/tests_fuzz.zig
```diff
@@ -232,8 +264,41 @@ fn evalOneOutcome(input: []const u8) EvalOutcome {
...
+    var parts: std.ArrayList(u8) = .empty;
+    defer parts.deinit(gpa);
+    appendPrinted(&parts, gpa, result) catch return .harness_unavailable;
+    for (fuzz_gen.global_names) |gn| {
+        const gv = vm.globals.get(gn) orelse continue;
+        parts.appendSlice(gpa, "\n") catch return .harness_unavailable;
+        parts.appendSlice(gpa, gn) catch return .harness_unavailable;
+        parts.appendSlice(gpa, "=") catch return .harness_unavailable;
+        appendPrinted(&parts, gpa, gv) catch return .harness_unavailable;
+    }
```
baijum, line 293: Nice touch widening the observable with the generator globals. Without this, a wrong fold inside `(define g1 …)` is invisible because `vm.eval` only returns the last top-level value. The planted-bug check (seed 30) makes the need concrete — good regression insurance for the oracle itself.

src/ir.zig
```diff
@@ -568,6 +568,15 @@ pub fn lower(irn: *IR, expr: Value) CompileError!*Node {
...
 }
 
+/// Master switch for IR optimization: when false, `lowerAndOptimize` skips
+/// the five optimization passes and `tryFoldFromAST` stops folding during
+/// lowering. Analysis (`markTailPositions`) always runs — it is required for
+/// correctness, not an optimization. Threadlocal (like `vm_instance`) so
+/// SRFI-18 child threads keep the default regardless of the parent's setting.
+/// Toggled by the `--no-ir-opt` CLI flag and by the differential fuzz oracle
+/// in tests_fuzz.zig (#1393).
+pub threadlocal var optimize_enabled: bool = true;
```
baijum, line 578: Threadlocal is a better fit than the issue’s suggested `IR` field / threaded parameter: one CLI flip covers every `lowerAndOptimize` / `tryFoldFromAST` call site, and SRFI-18 children keep the default. Matching `vm_instance` is the right precedent.

Confirmed `markTailPositions` still runs before the early return — good; analysis ≠ optimization.

src/tests_fuzz.zig
```diff
@@ -232,8 +264,41 @@ fn evalOneOutcome(input: []const u8) EvalOutcome {
     library.registerSandboxedLibraries(&vm.libraries, vm.globals) catch return .harness_unavailable;
     vm.sandbox_mode = true;
     vm.timeout_deadline_ns = @import("vm_calls.zig").clockNs() + 100_000_000;
-    _ = vm.eval(input) catch return .scheme_error;
-    return .ok;
+
+    // Toggle only around user-program evaluation, after the bootstrap and
+    // library registration above compiled with the default setting.
+    ir_mod.optimize_enabled = optimize;
```
baijum, line 270: Good that the toggle is deferred until after bootstrap / sandboxed library registration. That keeps stdlib compilation on the default (optimized) path and isolates the oracle to user-program IR opts — avoids conflating library compile differences with the program under test.

src/main.zig
baijum: Fixed in 40ab5c4d. `--no-ir-opt --compile` now also refuses an explicit `-o` that resolves to `getSbcPath(src)` — comparison is lexical (`std.fs.path.resolve` on both sides), so `prog.sbc`, `./prog.sbc`, and `dir/../dir/prog.sbc` are all caught; symlink aliases are not (noted in the doc comment). Unit-tested (`outputIsBytecodeCache detects cache-path collisions`) and verified end-to-end: the three colliding spellings exit 2 with the usage error, `-o guard-ok.sbc` compiles fine.

I kept refusal over a default `.noopt.sbc` suffix: a silently different output name seemed more surprising than an explicit error telling you why.

src/tests_ir.zig
```diff
@@ -746,3 +746,48 @@ test "IR: line-table entries recorded for IR-compiled code" {
...
+        .{ .src = "(and 1 2 3)", .want = 3 },
+        .{ .src = "(or #f 7)", .want = 7 },
+        .{ .src = "(let ((x (* 2 3))) (+ x (if #t 1 0)))", .want = 7 },
+    };
+    for (cases) |case| try th.expectEval(case.src, case.want);
+
+    ir_mod.optimize_enabled = false;
+    defer ir_mod.optimize_enabled = true;
+    for (cases) |case| try th.expectEval(case.src, case.want);
+}
```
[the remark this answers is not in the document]
baijum: Both addressed in 40ab5c4d:
- The folded-patterns test now uses a single `th.TestContext` shared across both modes (per the test guidelines in src/CLAUDE.md), instead of 28 one-shot VMs.
- Added `IR opt switch: self-tail-call still optimized without IR optimization`: 100k-iteration self-recursive function under `optimize_enabled = false` — overflows the frame stack unless `markTailPositions` still runs and the call compiles as a loop (mirrors the existing bare-lambda TCO test).

baijum:
## Review response — all comments addressed in 40ab5c4d

**Cache-poison hole (@baijum's inline on `main.zig`)** — fixed. The guard now refuses any `--no-ir-opt --compile` output that resolves to the source's `.sbc` cache path, not just a missing `-o`. `outputIsBytecodeCache` compares both paths after `std.fs.path.resolve`, so the natural spellings (`prog.sbc`, `./prog.sbc`, `dir/../dir/prog.sbc`) are all caught; symlink aliases are not (noted in the doc comment). Covered by a new unit test and verified end-to-end:

```console
$ kaappi --no-ir-opt --compile guard.scm                  # exit 2, usage error
$ kaappi --no-ir-opt --compile guard.scm -o guard.sbc     # exit 2, usage error
$ kaappi --no-ir-opt --compile guard.scm -o ./guard.sbc   # exit 2, usage error
$ kaappi --no-ir-opt --compile guard.scm -o guard-ok.sbc  # Compiled guard.scm -> guard-ok.sbc
```

I went with refusal rather than a default `.noopt.sbc` suffix: an explicit error stating the hazard seemed less surprising than a silently renamed output. Easy to revisit if the suffix ergonomics are preferred.

**CodeRabbit: `th.TestContext`** — taken. The folded-patterns test now shares one context across both modes instead of 28 one-shot VMs (matches the src/CLAUDE.md test guidelines and is faster).

**Tail-call regression case (CodeRabbit + review nit)** — added `IR opt switch: self-tail-call still optimized without IR optimization`: 100k-iteration self-recursion under `optimize_enabled = false`, which overflows the frame stack unless `markTailPositions` survives the switch and the call compiles as a loop. The analysis-vs-optimization boundary is now an executable invariant, not just a comment.

**Left as-is:** a shell-level test for the CLI refusal itself — the path-collision logic that could regress silently is unit-tested directly; the `usageError` wiring is a one-liner.

`zig build test` passes with all changes (format check green on the new head; platform jobs re-running).

src/main.zig
```diff
@@ -437,6 +451,30 @@ fn getSbcPath(allocator: std.mem.Allocator, scm_path: []const u8) ![]u8 {
...
+/// segments — symlinks and absolute-vs-relative aliases of the same file are
+/// not detected, but the natural spellings (`prog.sbc`, `./prog.sbc`) are.
+fn outputIsBytecodeCache(allocator: std.mem.Allocator, src_path: []const u8, output_path: []const u8) bool {
+    const cache_path = getSbcPath(allocator, src_path) catch return false;
+    defer allocator.free(cache_path);
+    const out_norm = std.fs.path.resolve(allocator, &.{output_path}) catch return false;
+    defer allocator.free(out_norm);
+    const cache_norm = std.fs.path.resolve(allocator, &.{cache_path}) catch return false;
+    defer allocator.free(cache_norm);
+    return std.mem.eql(u8, out_norm, cache_norm);
```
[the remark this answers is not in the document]
baijum: Valid — fixed in #1406. All `catch` arms in `outputIsBytecodeCache` (and the new symlink probe) now return `true`, so any failure refuses the compile rather than permitting the cache write. Regression test with `std.testing.FailingAllocator` at `fail_index = 0`.

src/main.zig
```diff
@@ -437,6 +451,30 @@ fn getSbcPath(allocator: std.mem.Allocator, scm_path: []const u8) ![]u8 {
...
+test "outputIsBytecodeCache detects cache-path collisions" {
+    const gpa = std.testing.allocator;
+    try std.testing.expect(outputIsBytecodeCache(gpa, "prog.scm", "prog.sbc"));
+    try std.testing.expect(outputIsBytecodeCache(gpa, "prog.scm", "./prog.sbc"));
+    try std.testing.expect(outputIsBytecodeCache(gpa, "dir/prog.scm", "dir/prog.sbc"));
+    try std.testing.expect(outputIsBytecodeCache(gpa, "dir/prog.scm", "dir/../dir/prog.sbc"));
+    try std.testing.expect(!outputIsBytecodeCache(gpa, "prog.scm", "prog.noopt.sbc"));
+    try std.testing.expect(!outputIsBytecodeCache(gpa, "prog.scm", "out/prog.sbc"));
+}
+
```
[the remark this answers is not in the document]
baijum: Addressed in #1406 with the cheap 90% version: a symlinked `-o` is refused outright under `--no-ir-opt --compile` (`std.c.readlink` probe — same pattern as SRFI-170's read-symlink), which kills this exact repro. Symlinked parent directories are still not canonicalized: `realpath` doesn't work on a not-yet-existing output, and walking existing prefix components adds machinery disproportionate to a footgun guard. The doc comment states the remaining gap. Regression test via `std.testing.tmpDir` + end-to-end check (symlinked `-o` exits 2).

src/tests_ir.zig
```diff
@@ -746,3 +746,68 @@ test "IR: line-table entries recorded for IR-compiled code" {
...
+    for (cases) |case| {
+        try std.testing.expectEqual(case.want, types.toFixnum(try ctx.vm.eval(case.src)));
+    }
+
+    ir_mod.optimize_enabled = false;
+    defer ir_mod.optimize_enabled = true;
+    for (cases) |case| {
+        try std.testing.expectEqual(case.want, types.toFixnum(try ctx.vm.eval(case.src)));
+    }
+}
```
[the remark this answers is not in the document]
baijum: Fixed in #1406. The three opt-switch tests and `evalNormalized` in the fuzz harness now save the prior value, set their baseline explicitly (`true` before the optimized loop), and restore the saved value on exit — no state leak into filtered or reordered runs.

Issue #1407, opened 2026-07-10
Native backend: free-variable analysis skips let/let* contents, miscompiling lambdas that capture only through a let


baijum:
Found by the VM-vs-native differential harness (#1395) on its first 500-seed batch — 4 of 500 generated programs diverged (seeds 145, 172, 176, 219 with base 0), all with this root cause.

## Reproducer (crash variant)

```scheme
(define g0 (lambda (u) ((lambda (a) (let ((b u)) b)) 1)))
(write (g0 5))
(newline)
```

```
$ kaappi min1.scm
5
$ kaappi compile min1.scm -o min1 && ./min1
undefined variable: u        # exit 1
```

## Reproducer (silent wrong value)

When a global with the same name exists, there is no error — the native binary silently reads the global instead of the captured parameter:

```scheme
(define u 99)
(define g0 (lambda (u) ((lambda (a) (let ((b u)) b)) 1)))
(write (g0 5))
(newline)
```

VM prints `5`; the native binary prints `99` and exits 0.

## Root cause

The closure-tier free-variable analysis in `src/llvm_emit_lambda.zig` does not descend into `let`/`let*`:

- `nodeHasFreeVars`: `.lambda, .passthrough, .let_form, .let_star, .sexpr_form => return false`
- `collectNodeFreeVars`: same tags are a no-op

So for `(lambda (a) (let ((b u)) b))`, the reference to `u` (a parameter of the enclosing lambda) is invisible:

1. `tryCompileNativeClosure` collects zero free variables → returns null ("nothing to capture").
2. `tryCompilePureLambdaAsNativeClosure` sees no free variables either → compiles the lambda as a **closed** native closure.
3. Inside the emitted function, `emitLet` resolves `u` via locals → params → upvalues → falls through to `kaappi_global_lookup("u")` — an undefined-variable exit, or a silently wrong value if a same-named global exists.

The emission side is already capable of handling this correctly: `emitGlobalRef` checks the upvalue map, so if the analysis *did* report `u` as free, tier 1 would capture it as an upvalue and the let body would read it correctly. Only the analysis is blind.

This is the same family as #827 (lets whose body lambdas capture let bindings) and #819 (set!/define in closure bodies), but from the other direction: a **lambda body containing a capturing let**.

## Fix directions

[...]

Issue #819, opened 2026-07-02
LLVM backend: set! and internal define always compile to global definition, ignoring lexical scope


baijum:
## Bug

In the LLVM native backend (`kaappi compile`), `set!` and internal `define` are always compiled to `kaappi_define_global` — regardless of whether the target is a lambda parameter, a `let`-bound local, or an upvalue. Mutations to lexical variables are silently dropped (a global of the same name is created/overwritten instead), and `set!` values that reference lexical variables crash at runtime because they are evaluated via `kaappi_eval` in the global environment.

## Affected code

- `src/llvm_emit.zig` — `emitSet` (lines 640-658): always emits `call void @kaappi_define_global(...)`; never consults `self.locals`, `self.params`, `self.rest_param_name`, or `self.upvalues`. The value, if a pair, is emitted via `emitEvalExpr` (global-env eval).
- `src/llvm_emit.zig` — `emitDefine` (lines 753-779): same for `define`, so internal defines inside a natively compiled `let` body create/overwrite globals instead of local bindings.
- `src/llvm_emit_lambda.zig` — `nodeHasFreeVars` (lines 554-593): `set_form` falls into the `else => return false` arm, and neither the `set!` target nor its value expression is traversed. Bodies containing `set!` therefore wrongly pass the "no free vars, safe to compile natively" check instead of falling back to the interpreter.

## Reproductions

All commands run from the repo root with `zig build && zig build lib` done:

```bash
zig-out/bin/kaappi prog.scm                        # interpreted
zig-out/bin/kaappi compile prog.scm -o prog && ./prog   # native
```

**1. `set!` on a parameter is ignored:**

```scheme
(define (f x)
  (set! x 10)
  x)
(display (f 5))
(newline)
```

interpreter: `10` — native: `5` (a global `x` = 10 is defined; the body then reads the untouched parameter slot).

**2. `set!` value referencing a parameter crashes:**

```scheme
(define (f x)
  (set! x (+ x 1))
  x)
(display (f 5))
(newline)
```

interpreter: `6` — native: `eval error`, exit 1 (`(+ x 1)` is evaluated by `kaappi_eval` in the global environment where `x` is unbound).

**3. `set!` on a `let` local is ignored:**

```scheme
(display (let ((x 1)) (set! x 2) x))
(newline)
```

interpreter: `2` — native: `1`.

**4. Top-level `set!` on an undefined variable silently defines it:**

```scheme
(set! zzz 5)
(display zzz)
(newline)
[...]

PR #1408, opened 2026-07-10
Fuzzing roadmap Phase 3: VM vs LLVM native backend differential harness

baijum:
Implements #1395 (fuzzing roadmap Phase 3, third item): an offline differential harness that runs the same generated program through the bytecode VM and through the LLVM native backend and compares stdout + exit class.

## What's here

- **`src/fuzz_gen_native.zig`** — a native-compilable-subset generator mode. The backend falls back to `kaappi_eval` for forms it can't compile, so unrestricted programs would degrade the diff to VM-vs-VM. The module doc comment catalogs every structural rule the backend imposes (free-variable limits in function bodies, literal-only global inits, no lambdas in `let`, no `set!` in inline-lambda bodies, void-valued top-level forms with explicit `(write ...)` observables) and why each exists.
- **`zig build fuzz-gen`** (`src/fuzz_gen_main.zig`) — standalone seed-replayable generator driver for the shell harness. Dev/CI only; not in the default install or releases.
- **`tests/fuzz/native-diff.sh`** — the harness: generate → `kaappi prog.scm` vs `kaappi compile prog.scm -o bin && ./bin` → compare. Both-nonzero exits match without comparing stdout (the VM continues past top-level errors; native halts at the first). Probes the toolchain up front because `kaappi compile` reports link failures on stderr but exits 0.
- **CI** — a `native-diff` job in the fuzz workflow: 300 programs nightly (~5 min), rotating seed base (printed, replayable via `workflow_dispatch` or locally), divergences uploaded as artifacts.
- **Gates** — `tests_native.zig`: fixed-seed programs must emit no unexpected `kaappi_eval` calls in the LLVM IR *and* every defined function must get a named native definition (eval-counting alone cannot see a rejected function). `tests_fuzz.zig`: programs evaluate cleanly (measured 100% over 300 seeds). The eval-count gate caught a subset leak during development (inline lambdas with `set!` hit the #819 scan), which is exactly the failure mode it exists for.
- **Docs** — new runbook section in `docs/dev/fuzzing.md` (running, comparison rules, triage); status tick in `docs/dev/fuzzing-feasibility.md`.

## First findings — the oracle works

`bash tests/fuzz/native-diff.sh 100` runs clean; a 500-seed batch found **5 divergences, all one real miscompilation**, filed as #1407: the closure tiers' free-variable analysis doesn't descend into `let`/`let*`, so a lambda capturing an enclosing binding only through a `let` compiles as closed and the reference becomes a global lookup — `undefined variable` at runtime, or a *silently wrong value* when a same-named global exists. Minimal reproducer in the issue.

> **Note:** the nightly `native-diff` job will stay red until #1407 is fixed (~1% of seeds hit it). That's the oracle doing its job; artifacts identify the known signature (`undefined variable: <letter>` in `seed-N.nat.err`), so new divergence classes remain visible during triage.

## Acceptance criteria from #1395

- `bash tests/fuzz/native-diff.sh 100` runs clean on main *(or files real bugs — also success)* — ✅ both: 100 clean, wider batch filed #1407
- Reproducibility: same seed → same programs — ✅ (deterministic test + seed base printed in CI)
- CI job runs on schedule with artifacts for divergences — ✅
- Uses `zig cc` for linking — ✅ (`kaappi compile` finds zig on PATH first)

### DIFF
--- a/tests/fuzz/native-diff.sh
+++ b/tests/fuzz/native-diff.sh
@@ -17,13 +17,18 @@
 #   RESULTS_DIR  where divergences are written (default: ./native-diff-results)
 #
 # Comparison rules (rationale in docs/dev/fuzzing.md):
-#   both exit 0          -> stdout must match byte-for-byte
-#   both exit non-zero   -> error classes match; stdout is NOT compared (the
-#                           VM reports a top-level error and continues, the
-#                           native binary exits at the first error, so
-#                           post-error output legitimately differs)
-#   exit classes differ  -> divergence
-#   either side timed out-> pair skipped (incomparable)
+#   both exit 0             -> stdout must match byte-for-byte
+#   both exit 1..127        -> ordinary-error match; stdout is NOT compared
+#                              (the VM reports a top-level error and
+#                              continues, the native binary exits at the
+#                              first error, so post-error output
+#                              legitimately differs)
+#   any exit >= 128         -> divergence: 128+N is death by signal N
+#                              (segfault, abort, ...), never an ordinary
+#                              Scheme error on either side
+#   exit classes differ     -> divergence (error on one side only)
+#   compile fails/times out -> divergence (generated programs must compile)
+#   either side timed out   -> pair skipped (incomparable)
 #
 # This is process-spawn + link per input — orders of magnitude slower than
 # in-process fuzzing — so it runs as a scheduled batch in CI (fuzz.yml), not
@@ -46,22 +51,28 @@ GEN=zig-out/bin/kaappi-fuzz-gen
 
 # GNU timeout when available (Linux/CI); on macOS without coreutils the
 # programs are bounded by construction, so running without one is acceptable.
-TIMEOUT_CMD=""
+# Compilation gets a longer budget than execution: it forks the system
+# linker, and a hang there would otherwise stall the whole batch.
+TIMEOUT_BIN=""
 for c in timeout gtimeout; do
   if command -v "$c" >/dev/null 2>&1; then
-    TIMEOUT_CMD="$c 10"
+    TIMEOUT_BIN="$c"
     break
   fi
 done
+RUN_TIMEOUT="${TIMEOUT_BIN:+$TIMEOUT_BIN 10}"
+COMPILE_TIMEOUT="${TIMEOUT_BIN:+$TIMEOUT_BIN 60}"
 
 WORK=$(mktemp -d "${TMPDIR:-/tmp}/native-diff.XXXXXX") || exit 1
 trap 'rm -rf "$WORK"' EXIT
 
 # Fail fast when the native toolchain is unavailable: `kaappi compile`
 # reports link/toolchain errors on stderr but still exits 0, so probe with a
-# trivial program and check that the output binary actually appears.
+# trivial program and check that the output binary actually appears. The
+# probe also warms the linker's compiler-rt cache, so per-seed compiles fit
+# the tighter COMPILE_TIMEOUT; its own budget is generous for a cold cache.
 printf '(write 42)\n(newline)\n' > "$WORK/probe.scm"
-"$KAAPPI" compile "$WORK/probe.scm" -o "$WORK/probe.bin" > "$WORK/probe.log" 2>&1
+${TIMEOUT_BIN:+$TIMEOUT_BIN 300} "$KAAPPI" compile "$WORK/probe.scm" -o "$WORK/probe.bin" > "$WORK/probe.log" 2>&1
 if [ ! -x "$WORK/probe.bin" ]; then
   echo "native-diff: 'kaappi compile' cannot produce binaries here:" >&2
   cat "$WORK/probe.log" >&2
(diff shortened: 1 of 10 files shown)


baijum:
Note for review ordering: this is independent of the pending #1406 (post-merge hardening of the #1405 guard). Both touch `src/tests_fuzz.zig` but in disjoint regions — #1406 inside `evalNormalized` (save/restore of `ir.optimize_enabled`), this PR adding a new test near the differential-oracle section. Verified with `git merge-tree`: clean merge in either order, and the new native-subset test only uses `evalNormalized` with default optimizer state, so #1406's save/restore change doesn't affect it.

tests/fuzz/native-diff.sh
```diff
@@ -0,0 +1,139 @@
...
+    skipped=$((skipped + 1))
+    rm -f "$bin" "$prog" "${prog%.scm}.sbc"
+    continue
+  fi
+
+  mismatch=""
+  if [ "$vm_exit" -eq 0 ] && [ "$nat_exit" -eq 0 ]; then
+    cmp -s "$WORK/vm.out" "$WORK/nat.out" || mismatch="stdout differs (both exited 0)"
+  elif [ "$vm_exit" -eq 0 ] || [ "$nat_exit" -eq 0 ]; then
+    mismatch="exit class differs (vm=$vm_exit native=$nat_exit)"
```
baijum, line 148: I think this needs a stronger exit-class check. Right now any pair of non-zero exits is treated as a match, so a native SIGSEGV/abort would be masked whenever the VM also returns a normal Scheme error. At minimum I'd distinguish ordinary interpreter/runtime errors from signal exits (and keep timeouts as their own class).

tests/fuzz/native-diff.sh
```diff
@@ -0,0 +1,139 @@
...
+    echo "native-diff: generator failed for seed $seed" >&2
+    exit 1
+  }
+
+  $TIMEOUT_CMD "$KAAPPI" "$prog" > "$WORK/vm.out" 2> "$WORK/vm.err"
+  vm_exit=$?
+
+  bin="$WORK/seed-$seed.bin"
+  rm -f "$bin"
+  "$KAAPPI" compile "$prog" -o "$bin" > "$WORK/compile.log" 2>&1
```
baijum: Could we put the per-seed `kaappi compile` under the same timeout as the VM/native executions? This harness is fuzzing the native backend too; if LLVM or the linker hangs on one seed, the whole batch/job stalls and we lose the repro instead of classifying it cleanly.

tests/fuzz/native-diff.sh
```diff
@@ -0,0 +1,139 @@
...
+skipped=0
+compared=0
+
+echo "native-diff: seeds $BASE..$((BASE + N - 1))"
+
+save_divergence() {
+  mkdir -p "$RESULTS_DIR"
+  cp "$prog" "$RESULTS_DIR/"
+  for f in vm.out vm.err nat.out nat.err compile.log; do
+    [ -f "$WORK/$f" ] && cp "$WORK/$f" "$RESULTS_DIR/seed-$seed.$f"
```
baijum, line 92: `save_divergence` can pick up stale `nat.out`/`nat.err` here on the compile-failure path, because those files are only refreshed after a successful compile. Clearing the per-seed artifact files before each iteration, or only copying native outputs after the native run happens, would make the saved repro reliable.

baijum (commented):
Thanks for wiring this up end to end. I ran `zig build test` in an isolated PR worktree and `bash tests/fuzz/native-diff.sh 10 0`; both passed locally. I left three blocking comments on `tests/fuzz/native-diff.sh`: the harness currently treats any two non-zero exits as equivalent, leaves the per-seed native compile step unbounded by timeout, and can attach stale native-output artifacts on compile failures. I don't think it's ready to merge until those are fixed.

tests/fuzz/native-diff.sh
baijum: Fixed in 28d5a65b. The comparison chain now has an explicit signal branch ahead of the ordinary-error match: any exit ≥ 128 (death by signal — segfault, abort) is a divergence even when the other side also errored; the both-nonzero match applies only to 1..127. Timeouts keep their own class (124/137 skip, only when a timeout binary is in use — without one, 137 correctly lands in the signal branch as a real SIGKILL). Script header and docs/dev/fuzzing.md updated to match the implementation.

tests/fuzz/native-diff.sh
baijum: Fixed in 28d5a65b. Per-seed `kaappi compile` now runs under its own 60 s timeout (longer than the 10 s execution budget since it forks the system linker); a timed-out compile is reported as its own divergence ("native compilation timed out") with the program saved, so the repro is kept instead of stalling the batch. The startup probe gets a generous 300 s budget — it also warms the compiler-rt cache so the per-seed compiles fit the tighter limit.

tests/fuzz/native-diff.sh
baijum: Fixed in 28d5a65b. All per-seed transients (vm.out/vm.err/nat.out/nat.err/compile.log) are cleared at the top of each iteration, so `save_divergence` can never attach a file from an earlier seed; its existing `[ -f ... ]` guard then simply skips the native outputs on the compile-failure path.

baijum:
Review addressed in 28d5a65b — verified with `zig build test` (707 pass) and `bash tests/fuzz/native-diff.sh 15 140`, which still catches the known #1407 divergence at seed 145 with fresh artifacts.

**Fixed** (the three blocking comments, plus overlapping CodeRabbit findings):
- Signal-terminated exits (≥128) are now a divergence class of their own — a native segfault can no longer hide behind an ordinary VM error. Docs and script header updated to match the implementation.
- Per-seed `kaappi compile` runs under a 60 s timeout (execution keeps 10 s); a hung linker is classified and saved instead of stalling the batch. The startup probe warms compiler-rt under a generous budget.
- Per-seed transients are cleared each iteration, so compile-failure divergences can't attach a previous seed's `nat.out`/`nat.err`.
- `fuzz_gen_main.zig`'s stdout loop now follows `reporting.writeToFd`'s errno handling (EINTR retry), exiting non-zero on failure — a truncated program must never be diffed as a real seed. (CodeRabbit's suggested `std.posix.write` doesn't exist in Zig 0.16.)
- `persist-credentials: false` on checkout in **both** fuzz workflow jobs — each executes fuzzer-generated programs.

**Skipped, with reasons:**
- *SHA-pinning actions*: every workflow in the repo references actions by tag (`@v7`, `@v1`, …); pinning only the two new steps would be inconsistent. Worth a dedicated repo-wide sweep if we adopt that policy.
- *Platform matrix for native-diff*: the fuzz workflow is deliberately single-platform (the existing `fuzz` job is ubuntu-only); platform coverage lives in the main CI. A macOS-ARM/riscv64 matrix here is a meaningful cost/latency change — happy to file a follow-up if wanted.
- *Concurrency group*: no workflow in the repo sets one; an overlapping manual + scheduled run executes disjoint seed bases on independent hosted runners, so overlap is redundant but harmless.
- *`.zig-cache` build caching for native-diff*: matches the existing `fuzz` job, which also rebuilds per run and relies on setup-zig's `cache-key`; the whole job fits well within budget (~5–10 min). Revisit if the nightly drags.

PR #1409, opened 2026-07-10
Descend into let/let* in the native closure free-variable analysis

### LINKED ISSUE #1407: Native backend: free-variable analysis skips let/let* contents, miscompiling lambdas that capture only through a let

baijum:
Fixes #1407.

## Problem

The closure tiers' free-variable analysis in `src/llvm_emit_lambda.zig` treated `.let_form`/`.let_star` IR nodes as opaque (`nodeHasFreeVars` returned false, `collectNodeFreeVars` was a no-op), so a lambda that captures an enclosing binding only through a `let` — `(lambda (a) (let ((b u)) b))` with `u` an enclosing param — compiled as a **closed** native closure. The hidden reference then degraded to `kaappi_global_lookup`: an "undefined variable" exit at runtime, or a silently wrong value when a same-named global existed. Found by the VM-vs-native differential harness (#1395/#1408) at ~1% of generated programs.

## Fix (the precise direction from the issue)

`let`/`let*` keep their contents as a raw S-expression (`ir.LetData.args`), so the new `FreeNameWalk` walks that raw form with proper binder scoping:

- let binders vs let* binders (parallel-let inits see only the enclosing scope; sequential inits see prior binders),
- nested `lambda` formals, including dotted rest and bare-symbol formals,
- `quote` skipped; every other symbol is a reference, with form keywords filtered by the same `ir.isKnownGlobal` rule the `.global_ref` arms use.

The walk feeds both `hasFreeVars` (tier 2 / `tryCompileDefineFunction`) and `collectFreeVars` (tier 1), so tier 1 now captures let-hidden references as upvalues — the emission side (`emitGlobalRef`) already resolved upvalues correctly, so the reproducer compiles natively rather than falling back.

When the walk meets something it cannot scope precisely (internal `define`, an eval-fallback form, malformed bindings, scope-stack or name-buffer overflow) it reports the analysis as inexact and the tiers reject, falling back to the interpreter instead of emitting with an incomplete capture set. The collectors' new `bool` return also closes a latent edge where the 17th distinct free name was silently dropped.

## Tests

- `src/tests_native.zig`: upvalue capture asserted for the let and let* reproducers (closure created with one upvalue, body reads `%upvalues`, captured name never interned for a global lookup); top-level lambda with a let-hidden free name falls back to `kaappi_eval` (correct there — the name is a genuine global reference); and an over-conservatism guard: a let over the lambda's own params still compiles as a closed native closure. The three #1407 tests fail without the fix.
- `zig build test`: 709/709.

## Verification

- Both issue reproducers now print `5` via `kaappi compile` (previously `undefined variable: u` exit 1, and silently `99`).
- 12 hand-written differential probes match the VM: parallel-let init scoping, binder shadowing, `set!` of params in let bodies, user globals through lets, variadic/bare-symbol lambdas in lets, quoted data, let under `when`.
- `bash tests/fuzz/native-diff.sh 500 100` (harness from #1408, run locally with this fix applied on that branch): **500 compared, 0 skipped, 0 divergent** — previously divergent at seeds 145, 172, 176, 219, 474. This should turn the nightly `native-diff` job green once #1408 lands.

## Out of scope

A sibling bug where the *nested-lambda* direction is still opaque (`(define g0 (lambda (u) ((lambda (a) (lambda (c) u)) 1)))` — no let involved) diverges both before and after this change; it needs tier 1 to capture from the enclosing closure's upvalues, not just its params. Flagged separately.

### DIFF
--- a/src/llvm_emit_lambda.zig
+++ b/src/llvm_emit_lambda.zig
@@ -70,7 +70,7 @@ fn tryCompilePureLambdaAsNativeClosure(self: *LLVMEmitter, data: ir.LambdaData) 
 
     var allowed: [17][]const u8 = undefined;
     @memcpy(allowed[0..arity], param_names[0..arity]);
-    if (hasFreeVars(body_nodes[0..body_count], allowed[0..arity])) return null;
+    if (hasFreeVars(self, body_nodes[0..body_count], allowed[0..arity])) return null;
 
     const fn_name = emitLambdaFunction(self, data.name, param_names[0..arity], body_nodes[0..body_count], rest_name) orelse return null;
 
@@ -139,12 +139,22 @@ fn tryCompileNativeClosure(self: *LLVMEmitter, data: ir.LambdaData) ?[]const u8 
 
     var free_vars: [16][]const u8 = undefined;
     var free_count: usize = 0;
-    if (!collectFreeVars(body_nodes[0..body_count], param_names[0..arity], &free_vars, &free_count)) return null;
+    if (!collectFreeVars(self, body_nodes[0..body_count], param_names[0..arity], &free_vars, &free_count)) return null;
     if (free_count == 0) return null;
 
     const outer_params = self.params orelse return null;
     for (free_vars[0..free_count]) |fv| {
-        if (!outer_params.contains(fv) and !ir.isKnownGlobal(fv)) return null;
+        // Captures are copied out of the enclosing %args array, so a name
+        // bound by an enclosing let-local or rest parameter (which outrank
+        // params in emitGlobalRef's resolution order) cannot be captured
+        // here, and neither can a name living only in the enclosing
+        // closure's upvalue array (#1410).
+        if (localsOrRestShadows(self, fv)) return null;
+        if (outer_params.contains(fv)) continue;
+        if (self.upvalues) |uv| {
+            if (uv.contains(fv)) return null;
+        }
+        if (!ir.isKnownGlobal(fv)) return null;
     }
 
     const id = self.lambda_counter;
@@ -338,7 +348,7 @@ pub fn tryCompileDefineFunction(self: *LLVMEmitter, name: []const u8, formals: V
         allowed[arity] = rn;
     }
     allowed[arity + extra] = name;
-    if (hasFreeVars(body_nodes[0..body_count], allowed[0 .. arity + extra + 1])) return null;
+    if (hasFreeVars(self, body_nodes[0..body_count], allowed[0 .. arity + extra + 1])) return null;
 
     return emitLambdaFunction(self, name, param_names[0..arity], body_nodes[0..body_count], rest_name);
 }
@@ -561,7 +571,25 @@ fn sexprReferencesNames(expr: Value, target_names: []const []const u8, excluded_
     return false;
 }
 
-// --- Free variable analysis helpers (standalone, no self) ---
+// --- Free variable analysis helpers ---
+//
+// These take the emitter because classification must respect the enclosing
+// emission scope: a name shadowed by an enclosing lexical binding (param,
+// let-local, rest parameter, upvalue) is a capture even when a known global
+// of the same name exists — `car` inside (lambda (car) ...) is the parameter,
+// not the primitive. The shadow check must run before isKnownGlobal.
+
+// Enclosing bindings that outrank the params array in emitGlobalRef's
+// resolution order and cannot be copied out of %args into an upvalue buffer.
+fn localsOrRestShadows(self: *LLVMEmitter, name: []const u8) bool {
+    if (self.locals) |loc| {
+        if (loc.get(name) != null) return true;
+    }
+    if (self.rest_param_name) |rp| {
+        if (std.mem.eql(u8, name, rp)) return true;
+    }
+    return false;
+}
 
 // True if the raw S-expression contains a set! or define form anywhere (not
 // descending into quoted data). Used to reject closure bodies that mutate or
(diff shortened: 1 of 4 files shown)


src/llvm_emit_lambda.zig
```diff
@@ -662,50 +667,254 @@ fn valueIsBoundOrLiteral(value: types.Value, params: []const []const u8) bool {
...
+    }
+
+    fn noteRef(w: *FreeNameWalk, name: []const u8) void {
+        for (w.params) |p| {
+            if (std.mem.eql(u8, name, p)) return;
+        }
+        for (w.bound[0..w.bound_count]) |b| {
+            if (std.mem.eql(u8, name, b)) return;
+        }
+        if (ir.isKnownGlobal(name)) return;
```
baijum: This still misses one important shadowing case: `ir.isKnownGlobal(name)` runs before we account for any enclosing lexical bindings, so a let-hidden capture is dropped when the outer binding shadows a primitive or special form. For example, `((lambda (car) ((lambda () (let ((x car)) x)))) 5)` should capture the outer `car`, but this path treats it as the global primitive and compiles a closed closure. We need to thread the enclosing lexical names into this walker (and the `.global_ref` free-var path) and classify those as captures before the known-global check. Please add a regression test with a shadowed primitive name like `car`.

baijum:
I left one inline blocking note on `src/llvm_emit_lambda.zig`: the new let/let* walk still drops captures when the enclosing lexical binding shadows a primitive or special-form name (for example `car`), so I don’t think this is merge-ready yet.

Once that shadowing case is handled and covered by a regression test, the rest of the change looks solid. Required CI is also still pending.

src/llvm_emit_lambda.zig
baijum: Fixed in 14dd2792. Both walkers (and the plain `.global_ref` arms, which had the same ordering bug even without a let) now consult `isNameShadowed` — which mirrors `emitGlobalRef`'s resolution order (locals → rest param → params → upvalues) — before `ir.isKnownGlobal`, so a shadowed primitive name is classified as a capture. `valueIsBoundOrLiteral` (the set! path) got the same ordering fix.

Tier 1's capture-source check is also hardened to match: a collected name that an enclosing let-local or rest parameter binds, or that lives only in the enclosing closure's upvalue array (#1410), now rejects native compilation instead of capturing the wrong `%args` slot or degrading to a global.

Verified VM-vs-native agreement (both print `5`) on:
- `((lambda (car) ((lambda () (let ((x car)) x)))) 5)` — your reproducer (was `#<builtin car>`)
- `((lambda (car) ((lambda () car))) 5)` — the no-let `.global_ref` variant (was `#<builtin car>`)
- let-local shadowing a primitive, and the original #1407 reproducer (still natively captured)

Regression tests added with the shadowed `car` param, both with and without the intervening let, asserting the upvalue capture and that `car` is never interned for a global lookup. Also applied CodeRabbit's suggestion to assert the `call i64 @kaappi_create_native_closure` call site rather than the always-present declaration.

baijum:
**CI failure analysis** — the failing test (`native-subset generator emits no unexpected kaappi_eval fallbacks`) was not introduced by the review fix: it is a collision between this PR's let/let* walk and the differential harness that #1408 merged to main *after* this branch's last green run. CI tests the merge ref, so the new oracle ran against this branch for the first time. I confirmed c127e592 (the pre-review-fix head) merged with current main fails identically on the same seed.

**Root cause**: the generator emits `(list-ref b i)` in function bodies, and its contract (`src/fuzz_gen_native.zig` header) requires every body op to be in `ir.isKnownGlobal` so calls don't count as free variables. `list-ref` was the one body op missing from `ir.primitives`. On main that never mattered — let/let* were opaque to the analysis — but with this PR's walk, seed 6's `(let ((e (list h v 29 -78)) …) (list-ref e 0))` reported `list-ref` as a free name and `f0` lost its native definition.

**Fix** (eb09ad7f): added `list-ref` to `ir.primitives`. I cross-checked the generator's full body vocabulary against the list — arith/div/cmp ops, car/cdr/cons/list/vector/make-vector/vector-ref/vector-set!/length and the predicates are all present; `list-ref` was the only gap. Verified locally on this branch merged with main: CI's exact seed (`0xf3c8bc08`) now passes, plus four additional seeds clean.

PR #1411, opened 2026-07-10
Fix nested syntax-rules substitution and template let ellipsis bindings

baijum:
## Summary

Fixes two `syntax-rules` expander bugs that broke Alex Shinn's portable `match` (and any macro using the classic Kiselyov/Campbell `let-syntax` identifier/ellipsis detection tricks), plus the expansion-time performance problems that made such macro towers unusably slow even where they worked.

### Bug 1: outer pattern variables were not substituted into nested `syntax-rules`

R7RS template semantics substitute pattern variables everywhere in the template — including the patterns and literals of a nested `syntax-rules` form. That substitution is what makes the "is this an identifier?" trick work:

```scheme
(define-syntax check-id
  (syntax-rules ()
    ((_ x sk fk)
     (let-syntax ((sym? (syntax-rules () ((sym? x k1 k2) k1) ((sym? y k1 k2) k2))))
       (sym? abracadabra sk fk)))))
```

The expander instead special-cased nested `syntax-rules` and *filtered out* colliding outer bindings (`instantiateNestedSyntaxRules`), so `x` stayed behind as an inner pattern variable and matched anything — `(check-id 0 ...)` claimed `0` was an identifier. Removed the special case; ordinary substitution + hygiene renaming does the right thing (inner pattern variables that are genuinely fresh get renamed consistently in both pattern and template).

### Bug 2: ellipsis in template `let` binding lists failed to expand

`instantiateLetBindings` (the binding-position-aware template walker for `let`/`let*`/`letrec`/`letrec*`) had no ellipsis handling, so the extremely common template shape

```scheme
((_ ((var value) ...) . body) (let ((var value) ...) . body))
```

failed with `InvalidSyntax`. The new ellipsis branch delegates repetitions to the regular `instantiateEllipsis` machinery (including `((var value) ... ...)` depth-2 flattening per R7RS 4.3.2) and then resumes let-binding processing for anything after the ellipsis tokens.

### Performance: safety fills and quadratic re-expansion

Two independent sources of pathological slowness for macro-heavy code:

- **Expander**: each pattern-match attempt / ellipsis repetition re-initialized a `[128]Binding` buffer where every `Binding` carries an 8 KB `ellipsis_values` array — a ~1 MB safety fill (or whole-struct copy) per attempt. The buffers are now hoisted and written field-wise; entries below the count are always written before being read.
- **Compiler**: the set!-target scan that runs after every macro expansion (#1250 Part B) recursively re-expanded nested macro uses, so a depth-N CPS-style macro tower re-expanded its whole subtree at every level — quadratic (and combined with the safety fills, effectively a hang). Only the top-level pre-scan expands macros now; each nested use is scanned when compilation actually expands it, before anything in that expansion is compiled.

On a CPS-style macro-expansion stress test (~10^6 expansions), `main` did not finish within 5 minutes; with this change it completes in ~43 s. Realistic match-style macros go from unusable to instant.

### Also

- `vm_library.zig`: reader failures while loading a `.sld` now surface as `LibrarySourceReadError` (instead of a vague `InvalidSyntax`), and compile failures of library body forms record the underlying error name in the error detail.

## Tests

- New regression suite `tests/scheme/hygiene/nested-syntax-rules-subst.scm` (19 assertions): identifier/ellipsis/literal-membership detection through nested `syntax-rules`, template `let`/`let*`/`letrec`/named-`let` ellipsis binding lists, bindings before/after the ellipsis group, two ellipsis groups, zero repetitions, `... ...` flattening, and hygiene across the nested-macro boundary. 8 of the assertions fail (3 as compile errors) on `main`.
- `zig build test` passes; full Scheme suite (`tests/scheme/run-all.sh`) passes including the 1,395-test R7RS suite.
- Existing #1250 regression coverage (`tests/scheme/continuations/callcc-set-store.scm`) passes; also hand-verified store semantics and fold-blocking for a `set!` hidden two macro layers deep.

### DIFF
--- a/src/expander.zig
+++ b/src/expander.zig
@@ -182,9 +182,15 @@ pub fn expandMacro(gc: *GC, expr: Value, transformer_val: Value, globals: ?*std.
 
     const literal_bound = transformer.literal_bound;
 
-    // Try each rule in order
+    // Try each rule in order. The bindings buffer is hoisted and left
+    // uninitialized (entries below bind_count are always written before
+    // being read) — a safety fill of the ~1MB array per rule attempt is
+    // prohibitively slow for macro-heavy code.
+    var bindings: [MAX_BINDINGS]Binding = b: {
+        @setRuntimeSafety(false);
+        break :b undefined;
+    };
     for (0..transformer.num_rules) |i| {
-        var bindings: [MAX_BINDINGS]Binding = undefined;
         var bind_count: usize = 0;
 
         // Skip the keyword in the pattern (first element of pattern)
@@ -239,14 +245,16 @@ fn matchPattern(pattern: Value, input: Value, literals: []const Value, bindings:
         // Underscore (not in literals): match anything, bind nothing
         if (std.mem.eql(u8, name, "_")) return true;
 
-        // Pattern variable: bind to input
+        // Pattern variable: bind to input. Fields are set individually
+        // because a struct literal re-initializes the 8KB ellipsis_values
+        // field (safety fill), which dominated expansion time on
+        // macro-heavy code.
         if (count.* >= MAX_BINDINGS) return false;
-        bindings[count.*] = .{
-            .name = name,
-            .value = input,
-            .depth = 0,
-            .is_list = false,
-        };
+        bindings[count.*].name = name;
+        bindings[count.*].value = input;
+        bindings[count.*].depth = 0;
+        bindings[count.*].is_list = false;
+        bindings[count.*].ellipsis_count = 0;
         count.* += 1;
         return true;
     }
@@ -363,23 +371,29 @@ fn matchEllipsis(elem_pattern: Value, rest_pattern: Value, input: Value, literal
     collectPatternVars(elem_pattern, literals, &elem_var_names, &elem_var_count, &var_overflow);
     if (var_overflow) return false;
 
-    // Create list bindings for each pattern variable found in the ellipsis sub-pattern
+    // Create list bindings for each pattern variable found in the ellipsis
+    // sub-pattern. Field-wise assignment: see matchPattern.
     const base_count = count.*;
     for (0..elem_var_count) |vi| {
         if (count.* >= MAX_BINDINGS) return false;
-        bindings[count.*] = .{
-            .name = elem_var_names[vi],
-            .value = types.NIL,
-            .depth = 1,
-            .is_list = true,
-        };
+        bindings[count.*].name = elem_var_names[vi];
+        bindings[count.*].value = types.NIL;
+        bindings[count.*].depth = 1;
+        bindings[count.*].is_list = true;
+        bindings[count.*].ellipsis_count = 0;
         count.* += 1;
     }
 
-    // Match each repetition
+    // Match each repetition. The sub-binding buffer is hoisted out of the
+    // loop and left uninitialized: matchPattern only writes entries below
+    // sub_count, and a safety fill of the ~1MB array per repetition is
+    // what made match-style macros unusably slow.
+    var sub_bindings: [MAX_BINDINGS]Binding = b: {
+        @setRuntimeSafety(false);
+        break :b undefined;
+    };
     var inp = input;
     for (0..repeat_count) |_| {
-        var sub_bindings: [MAX_BINDINGS]Binding = undefined;
         var sub_count: usize = 0;
         if (!matchPattern(elem_pattern, types.car(inp), literals, &sub_bindings, &sub_count, gc, literal_bound, use_check))
             return false;
@@ -543,11 +557,6 @@ fn instantiateTemplate(gc: *GC, template: Value, bindings: []Binding, intro_scop
         }
     }
 
-    // Nested syntax-rules: protect inner pattern variables from outer substitution
-    if (types.isSymbol(elem) and std.mem.eql(u8, types.symbolName(elem), "syntax-rules")) {
-        return instantiateNestedSyntaxRules(gc, template, bindings, intro_scope, literals, macro_keyword, globals, macros);
-    }
-
     // Detect binding forms and set binding-position flag for variable names
     const BINDING_FLAG: u32 = 0x20000000;
     if (types.isSymbol(elem)) {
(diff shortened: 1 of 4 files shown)


baijum (commented):
I found two new high-confidence expander regressions that need fixing before this is ready to merge: generated nested `syntax-rules` forms lose their own ellipses, and repeated template-introduced `let` binders lose binding-position hygiene. Both are locally reproducible on this branch; I left inline comments with minimal repros.

src/expander.zig
```diff
@@ -592,6 +601,40 @@ fn instantiateLetBindings(gc: *GC, binding_list: Value, bindings: []Binding, sco
...
+            defer gc.popRoot();
+            var rest_after = types.cdr(after);
+            while (types.isPair(rest_after)) {
+                const tok = types.car(rest_after);
+                if (types.isSymbol(tok) and isEllipsis(types.symbolName(tok))) {
+                    synth_rest = try gc.allocPair(tok, synth_rest);
+                    rest_after = types.cdr(rest_after);
+                } else break;
+            }
+            const segment = try instantiateEllipsis(gc, pair, synth_rest, bindings, scope & ~@as(u32, 0x20000000), literals, macro_keyword, globals, macros);
```
baijum: Clearing `BINDING_FLAG` here loses binding-position hygiene for repeated `let` binders. For template-introduced names like `(let ((exp value) ...) body)`, the repeated `exp` binders are instantiated as references, so they skip gensyming and can capture use-site references. Repro: a macro expanding to that shape makes `(capture-test (1) (begin (display (exp 0)) (newline)))` fail with `not a procedure` on this branch. Could we preserve binding context for the repeated lhs, or run each generated binding pair through the normal binding-aware path?

tests/scheme/hygiene/nested-syntax-rules-subst.scm
```diff
@@ -0,0 +1,150 @@
...
+  (syntax-rules ()
+    ((_ id sk fk)
+     (let-syntax ((ell? (syntax-rules ()
+                          ((ell? (foo id) k1 k2) k1)
+                          ((ell? other k1 k2) k2))))
+       (ell? (a b c) sk fk)))))
+
+(test-equal "ellipsis detected" 'yes (check-ellipsis ... 'yes 'no))
+(test-equal "plain symbol is not ellipsis" 'no (check-ellipsis blah 'yes 'no))
+
```
baijum, line 46: Can we add a regression where the generated nested transformer contains its own `(x ...)` pattern/template? After removing the dedicated nested `syntax-rules` traversal, I can reproduce an inner macro like `(syntax-rules () ((_ (x ...)) 'preserved) ((_ other) 'lost))` returning `'lost'` because the outer instantiation consumes the inner ellipsis instead of preserving it for the nested matcher.

src/expander.zig
baijum: Fixed in e8390521. The ellipsis path now tags each repetition element with a `LET_PAIR_FLAG` context bit, and `instantiateTemplate` splits the pair var/init with `BINDING_FLAG` applied to the var — so repeated template-introduced binders get the binding-position rename exactly like the non-ellipsis path. Your repro is now a regression test ("repeated builtin-named binder is renamed"): `(capture-test (1) (exp 0))` evaluates to 1.0 via the builtin.

tests/scheme/hygiene/nested-syntax-rules-subst.scm
baijum: Added in e8390521 (`gen-inner` and `gen-adder` tests), and fixed the underlying behavior: a new `NESTED_SR_FLAG` context bit is set for the subtree when a `syntax-rules` head is encountered during instantiation. Inside that context, an ellipsis whose element references no outer list binding is preserved literally (element still gets outer scalar substitution + hygiene) instead of expanding to zero repetitions. Worth noting: this repro also returns `lost` on main — the old `instantiateNestedSyntaxRules` filtered the bindings but then recursed through plain `instantiateTemplate`, which dropped inner ellipses the same way. Both directions are covered now, including an inner-template ellipsis combined with outer substitution (`gen-adder` → 16).

src/vm_library.zig
```diff
@@ -261,15 +261,23 @@ fn loadLibrarySource(vm: *VM, source: []const u8) !void {
...
         if (vm.handleTopLevelForm(expr)) |result| {
             _ = result catch |err| return err;
         } else {
-            const func = compiler_mod.compileExpressionWithMacros(vm.gc, expr, &vm.macros, vm.globals) catch return error.InvalidSyntax;
+            const func = compiler_mod.compileExpressionWithMacros(vm.gc, expr, &vm.macros, vm.globals) catch |err| {
+                if (vm.last_error_detail_len == 0) {
+                    vm.setErrorDetail("{s} while compiling library body form", .{@errorName(err)});
+                }
+                return error.InvalidSyntax;
+            };
```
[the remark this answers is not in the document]
baijum: Fixed in e8390521 — `OutOfMemory` now propagates; other compile errors still map to `InvalidSyntax` with the error name recorded in the detail.

tests/scheme/hygiene/nested-syntax-rules-subst.scm
```diff
@@ -0,0 +1,150 @@
...
+  (syntax-rules ()
+    ((_ x (id ...) sk fk)
+     (let-syntax ((mem? (syntax-rules (id ...)
+                          ((mem? x k1 k2) k2)
+                          ((mem? y k1 k2) k1))))
+       (mem? probe-symbol sk fk)))))
+
+(test-equal "member of id list" 'known (known-id? b (a b c) 'known 'new))
+(test-equal "not member of id list" 'new (known-id? z (a b c) 'known 'new))
+(test-equal "empty id list" 'new (known-id? z () 'known 'new))
```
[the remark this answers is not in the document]
baijum: Added in e8390521: `gen-inner` covers a generated inner transformer with its own `(x ...)` pattern (preserved for lists including empty, falls through for non-lists), and `gen-adder` covers an inner-template ellipsis combined with outer substitution. The underlying preservation fix is in the same commit.

baijum:
All review findings addressed in e8390521:

- **Nested transformer ellipses preserved** (baijum + CodeRabbit outside-diff): new `NESTED_SR_FLAG` context bit; inside a nested `syntax-rules` template, an ellipsis whose element references no outer list binding is kept literally. This was also broken on main (the old filtering path recursed through plain `instantiateTemplate` and dropped inner ellipses the same way), so the fix is a strict improvement rather than just a regression repair.
- **Repeated `let` binders keep binding-position hygiene** (baijum): new `LET_PAIR_FLAG` so each `(var init)` repetition renames the var as a binder; the `(capture-test (1) (exp 0))` repro now calls the builtin.
- **Unreferenced list bindings no longer indexed past their count** in `instantiateEllipsis` (CodeRabbit outside-diff): the sub-binding loop now skips list bindings not referenced by the element template — with two ellipsis groups of different lengths this was an uninitialized read, and a potential garbage-pointer dereference in the depth>1 branch.
- **`OutOfMemory` propagates** from library body compilation instead of surfacing as `InvalidSyntax`/"library not found" (CodeRabbit inline).

The `intro_scope` flag bits are now file-scope constants, and `renameForHygiene` masks the context flags so a template identifier renames consistently inside and outside nested contexts.

Validation: 6 new regression assertions (25 total in `nested-syntax-rules-subst.scm`), `zig build test` green, full Scheme suite 1822 pass / 0 fail including the 1,395-test R7RS suite.

PR #1413, opened 2026-07-10
Harden GitHub Actions: pin actions by SHA and disable persisted checkout credentials

### LINKED ISSUE #1400: Harden GitHub Actions: pin actions by SHA and disable persisted checkout credentials
Follow-up from #1398.

Apply this hardening consistently across the repository GitHub Actions workflows rather than changing only the fuzz workflow.

Scope:
- Pin every third-party uses: reference to a full immutable commit SHA, with a version comment where helpful.
- Set persist-credentials: false for every actions/checkout step unless a workflow has a documented, necessary exception.
- Audit at least .github/workflows/ci.yml, .github/workflows/release.yml, .github/workflows/nightly.yml, and .github/workflows/fuzz.yml; include any other workflow files found during the audit.
- Verify workflows retain the permissions required for checkout, artifact upload, releases, and publishing.

This is intentionally a repository-wide change so the supply-chain posture and workflow conventions remain consistent.

baijum:
Closes #1400 (follow-up from #1398, which hardened only fuzz.yml).

## What changed

**SHA pinning** — every `uses:` reference in the 5 workflows (`ci.yml`, `release.yml`, `fuzz.yml`, `benchmark-pr.yml`, `post-release.yml`) now points at a full commit SHA with a comment naming the version it was resolved from. All SHAs were resolved live via `gh api repos/<owner>/<repo>/commits/<tag>`:

| Action | Was | Pinned to |
|--------|-----|-----------|
| actions/checkout | v7 | `9c091bb2` (v7.0.0) |
| xyzzylabs/setup-zig | v1 | `09d85d9d` (v1.0.2) |
| docker/setup-qemu-action | v4 | `96fe6ef7` (v4.2.0) |
| actions/upload-artifact | v7 | `043fb46d` (v7.0.1) |
| actions/download-artifact | v8 | `3e5f45b2` (v8.0.1) |
| actions/cache | v4 | `0057852b` (v4.3.0) |
| codecov/codecov-action | v5 | `0fb71748` (v5.5.5) |
| benchmark-action/github-action-benchmark | v1 | `52576c92` (v1.22.1) |
| softprops/action-gh-release | v3 | `718ea10b` (v3.0.1) |
| kaappi/github-action-pull-request-benchmark | v1 | `11e6f9c4` (v1.4.0) |

**persist-credentials: false** on all 17 checkout steps — no exceptions needed. The three places where disabling credentials could plausibly break something were verified and got an explanatory comment in the workflow:

- **ci.yml benchmark job** (`auto-push: true`): github-action-benchmark builds its push remote as `https://x-access-token:${token}@…` from its `github-token` input and explicitly blanks checkout's `http.<server>/.extraheader` config, so it never uses persisted credentials (verified in `dist/src/git.js` at the pinned SHA).
- **release.yml release job**: action-gh-release is pure API via the workflow token; no git operations.
- **benchmark-pr.yml**: the base/head `git checkout <sha>` steps are local — `fetch-depth: 0` plus the PR merge ref already bring all needed objects.

**Permissions retained/verified** per the issue: ci.yml keeps `id-token: write` (Codecov OIDC) + job-level `contents: write` for the benchmark push; release.yml keeps `contents: write` for release creation; artifact upload/download use the runtime token and are unaffected. One gap fixed: **post-release.yml had no `permissions:` block at all** and inherited the repository default — it now declares `contents: read`, which is all `gh release download` needs.

**Companions**
- `.github/dependabot.yml` (github-actions ecosystem, weekly, grouped) so pins are bumped via PRs with the version comments kept in sync, instead of going stale silently.
- `docs/dev/github-actions.md` documents the conventions (pinning, how to resolve a SHA, the persist-credentials policy, least-privilege permissions) and is linked from the docs index.

## Out of scope

`nightly.yml` (named in the issue) lives in the org-level `kaappi/.github` repo, not here — it needs the same treatment in a separate PR there.

## Verification

- All 6 YAML files parse cleanly.
- `grep 'uses:.*@v[0-9]'` over `.github/` finds no remaining mutable tag refs.
- checkout count == `persist-credentials: false` count in every workflow (17/17).
- No `.zig` changes; CI on this PR exercises the pinned ci.yml + benchmark-pr.yml end-to-end.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# Keeps the SHA-pinned actions in .github/workflows/ from going stale:
+# dependabot proposes bumps that update both the pinned commit SHA and
+# the version comment next to it. See docs/dev/github-actions.md.
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    groups:
+      github-actions:
+        patterns: ['*']
--- a/.github/workflows/benchmark-pr.yml
+++ b/.github/workflows/benchmark-pr.yml
@@ -20,12 +20,16 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
-      - uses: actions/checkout@v7
+      # fetch-depth: 0 brings all refs during checkout, so the later
+      # base/head `git checkout <sha>` steps are purely local and need
+      # no git credentials.
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           fetch-depth: 0
+          persist-credentials: false
 
       - name: Install Zig
-        uses: xyzzylabs/setup-zig@v1
+        uses: xyzzylabs/setup-zig@09d85d9dbb73308882e7a26e61a8f706eed40df1 # v1.0.2
         with:
           version: 0.16.0
 
@@ -42,7 +46,7 @@ jobs:
           git checkout ${{ github.event.pull_request.head.sha }}
 
       - name: Compare and comment
-        uses: kaappi/github-action-pull-request-benchmark@v1
+        uses: kaappi/github-action-pull-request-benchmark@11e6f9c49191a57127f53dddcbdcc30b18f7f0a3 # v1.4.0
         with:
           tool: 'raw'
           pr-benchmark-file-path: pr-results.json
--- a/.github/workflows/benchmark-pr.yml
+++ b/.github/workflows/benchmark-pr.yml
@@ -46,7 +46,7 @@ jobs:
           git checkout ${{ github.event.pull_request.head.sha }}
 
       - name: Compare and comment
-        uses: kaappi/github-action-pull-request-benchmark@11e6f9c49191a57127f53dddcbdcc30b18f7f0a3 # v1.4.0
+        uses: kaappi/github-action-pull-request-benchmark@1a82ba746b2050a6143f3e4ab67d935e505f7b73 # v1 tag (2026-07-01)
         with:
           tool: 'raw'
           pr-benchmark-file-path: pr-results.json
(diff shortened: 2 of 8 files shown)


baijum:
The first CI run failed on `benchmark-pr` (`Resource not accessible by integration` at the comment step) — root cause and fix in 996e02a:

`kaappi/github-action-pull-request-benchmark` has **both a `v1` tag and a `v1` branch**, pointing at different commits:

- tag `v1` → `1a82ba74` (master tip, 2026-07-01) — what the Actions runner actually runs for `@v1`; posts **PR comments** (needs only `pull-requests: write`, which the workflow grants)
- branch `v1` → `11e6f9c4` (stale v1.4.0, Feb 2024) — posts **commit comments** (needs `contents: write`, which the workflow deliberately doesn't grant)

The pins were originally resolved with `gh api repos/<owner>/<repo>/commits/<ref>`, which prefers the *branch* on a name collision, so the pin captured the stale v1.4.0 while every earlier green run had been executing the tag (confirmed via the runner's `Set up job` logs: green run downloaded `1a82ba74`, red run downloaded `11e6f9c4`).

The fix repins to the tag's commit `1a82ba74` and corrects the resolution recipe in `docs/dev/github-actions.md` to use `git/ref/tags/<tag>` (with annotated-tag peeling), which a branch cannot shadow. All 10 pins were re-verified against `refs/tags`; the other 9 were already correct (`benchmark-action/github-action-benchmark` has no `v1` tag at all — its `v1` is a branch, and the pinned SHA is that branch's tip, which tag `v1.22.1` also points to).

Rather ironic failure mode for a supply-chain-hardening PR: `@v1` was never even deterministic in that repo — API consumers and the runner resolved it to different code. That repo has issues disabled, so noting the follow-up here: delete (or fast-forward) its stale `v1` branch and cut a version tag at the current `v1` target so SHA pins have a named version to reference.

PR #1425, opened 2026-07-10
Reject native compilation when a set! param is captured by a nested lambda

### LINKED ISSUE #1422: Native backend: closures snapshot captured variables by value, diverging from the VM when a captured binding is mutated after capture
Found while extending the native-subset fuzz generator for #1420.

The native closure tiers copy captured variables **by value** into the upvalue array at closure-creation time (`tryCompileNativeClosure` in `src/llvm_emit_lambda.zig`), and eval fallbacks read captured names from globals republished at the fallback point (`bindParamsAsGlobals`). The bytecode VM captures **locations** (R7RS semantics): a `set!` of the captured binding after closure creation is visible through the closure. The #819 guard rejects `set!` *inside* the closure body, but nothing guards against the enclosing scope mutating a captured binding while the closure is live.

## Reproducers

Tier-1 upvalue copy (`set!` in a sibling argument runs between closure creation and the call):

```scheme
(define (f0 u) ((lambda (a) (+ u a)) (let ((b 5)) (set! u 90) b)))
(write (f0 1))
(newline)
```

Eval-fallback republished global (same shape, variadic inner):

```scheme
(define (f0 u) ((lambda (a . rest) (+ u a)) (let ((b 5)) (set! u 90) b) 7))
(write (f0 1))
(newline)
```

Both at dd718512:

| | VM (`kaappi f.scm`) | native (`kaappi compile`) |
|---|---|---|
| output | `95` | `6` |

The VM reads `u` at call time through the captured location (90 + 5); the native build reads the value snapshotted before the argument ran (1 + 5).

Retained closures widen the window arbitrarily — any `(define (f u) ...)` that creates a closure over `u`, later `set!`s `u`, and then calls the closure will show the same divergence, no argument-order subtlety needed.

## Notes

- Both reproducer shapes were **generable by the native-subset fuzz generator** (inline capturing lambda + `set!` of a param inside a `let` in a sibling argument), i.e. a latent landmine for `tests/fuzz/native-diff.sh`. The #1420 generator work bans `set!` inside inline-call argument subtrees so the oracle stays clean; this issue tracks the underlying backend semantics.
- Fix directions: box captures whose binding is `set!` anywhere in the enclosing function (matching VM location semantics), or conservatively reject native compilation of a function that `set!`s a name captured by any of its inner lambdas (the eval fallback for the *lambda* alone does not help — republished globals snapshot too; the rejection would have to cover the whole enclosing function).

baijum:
## Summary

- Add a guard in `tryCompileDefineFunction` that rejects native compilation when a function parameter is both targeted by `set!` anywhere in the body and captured by a nested lambda — the native upvalue-copy model snapshots at closure creation, making later mutations invisible to the closure (#1422)
- Update the native-subset fuzz generator to mark function params as non-settable so generated programs stay on the native path

## Test plan

- [x] `zig build test` — 728 unit tests pass (including the fuzz generator's native-subset fallback check)
- [x] `bash tests/scheme/compile/native-set-capture-divergence-1422.sh` — 5 cases: issue reproducers (output `95` instead of `6`), retained closure, shadowed param (still compiles natively), no-conflict different param (still compiles natively)
- [x] All existing `tests/scheme/compile/*.sh` pass — no regressions

Closes #1422

### DIFF
--- a/src/fuzz_gen_native.zig
+++ b/src/fuzz_gen_native.zig
@@ -206,7 +206,7 @@ fn genLambda(g: *Gen, ctx: *Ctx, arity: u8, variadic: bool) Error!void {
     for (params, 0..) |p, i| {
         if (i > 0) try g.emit(" ");
         try g.emit(p);
-        try g.scope.append(g.gpa, .{ .name = p, .kind = .int, .settable = false });
+        try g.pushNonSettable(p, .int);
     }
     if (variadic) {
         try g.emit(" . rest");
@@ -250,7 +250,7 @@ fn genFnDefine(g: *Gen, ctx: *Ctx) Error!void {
                 // Non-settable: inline lambdas in the body may capture these
                 // params, and set! of a captured param triggers the #1422
                 // guard, rejecting the whole function for native compilation.
-                try g.scope.append(g.gpa, .{ .name = p, .kind = .int, .settable = false });
+                try g.pushNonSettable(p, .int);
             }
             if (variadic) {
                 try g.emit(" . rest");
@@ -269,8 +269,8 @@ fn genFnDefine(g: *Gen, ctx: *Ctx) Error!void {
                 name, params[0], params[1], params[0], params[1], name, params[0],
             });
             var saved = try enterBody(g, false);
-            try g.scope.append(g.gpa, .{ .name = params[0], .kind = .int, .settable = false });
-            try g.scope.append(g.gpa, .{ .name = params[1], .kind = .int, .settable = false });
+            try g.pushNonSettable(params[0], .int);
+            try g.pushNonSettable(params[1], .int);
             var body_ctx: Ctx = .{ .fn_depth = 1 };
             try genInt(g, &body_ctx, 3);
             leaveBody(g, &saved);
@@ -282,7 +282,7 @@ fn genFnDefine(g: *Gen, ctx: *Ctx) Error!void {
             const op = rec_ops[g.ch.index(.op_pick, rec_ops.len)];
             try g.emitf("(define ({s} {s}) (if (<= {s} 0) ", .{ name, params[0], params[0] });
             var saved = try enterBody(g, false);
-            try g.scope.append(g.gpa, .{ .name = params[0], .kind = .int, .settable = false });
+            try g.pushNonSettable(params[0], .int);
             var body_ctx: Ctx = .{ .fn_depth = 1 };
             try genInt(g, &body_ctx, 2);
             try g.emitf(" ({s} ", .{op});
--- a/src/fuzz_gen_native.zig
+++ b/src/fuzz_gen_native.zig
@@ -112,6 +112,9 @@ const Ctx = struct {
     /// arguments run between closure creation (which snapshots captured
     /// params by value) and the call, so a set! of a captured param there
     /// diverges from the VM's location-based capture (#1422).
+    /// Note: set! of a define-position function's own params is prevented
+    /// separately by pushing those params with settable=false — the #1422
+    /// guard rejects functions whose params are both set! and captured.
     no_set: bool = false,
 };
 
@@ -203,7 +206,7 @@ fn genLambda(g: *Gen, ctx: *Ctx, arity: u8, variadic: bool) Error!void {
     for (params, 0..) |p, i| {
         if (i > 0) try g.emit(" ");
         try g.emit(p);
-        try g.pushBinding(p, .int);
+        try g.scope.append(g.gpa, .{ .name = p, .kind = .int, .settable = false });
     }
     if (variadic) {
         try g.emit(" . rest");
@@ -244,7 +247,10 @@ fn genFnDefine(g: *Gen, ctx: *Ctx) Error!void {
             var saved = try enterBody(g, false);
             for (params) |p| {
                 try g.emitf(" {s}", .{p});
-                try g.pushBinding(p, .int);
+                // Non-settable: inline lambdas in the body may capture these
+                // params, and set! of a captured param triggers the #1422
+                // guard, rejecting the whole function for native compilation.
+                try g.scope.append(g.gpa, .{ .name = p, .kind = .int, .settable = false });
             }
             if (variadic) {
                 try g.emit(" . rest");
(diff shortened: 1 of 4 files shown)


src/llvm_emit_lambda.zig
```diff
@@ -360,6 +360,9 @@ pub fn tryCompileDefineFunction(self: *LLVMEmitter, name: []const u8, formals: V
         param_list = types.cdr(param_list);
     }
 
+    // #1422: reject if a param is both set! and captured by a nested lambda.
+    if (bodyHasConflictingSetCapture(body, param_names[0..arity], rest_name)) return null;
```
baijum, line 367: This guard covers `tryCompileDefineFunction` (including define-position lambdas via `tryCompileLambdaNative`), but `tryCompilePureLambdaAsNativeClosure` also compiles whole functions and has **no** set!/define check — the #819 blanket scan only exists in tier 1. The divergence still reproduces on this branch:

```scheme
(write ((lambda (u) ((lambda (a) (+ u a)) (let ((b 5)) (set! u 90) b))) 1))
; VM: 95, native binary: 6
```

The emitted IR shows the outer lambda compiled as a closed native fn (`@lambda_0`), with the inner lambda's capture of `u` snapshotted into `@closure_1`'s upvalue buffer — exactly the #1422 mechanism. The issue's fix-direction note ("the rejection would have to cover the whole enclosing function") applies to this tier too.

Suggested fix — same helper in `tryCompilePureLambdaAsNativeClosure` (rest params are already rejected there, so pass `null`):

```zig
if (bodyHasConflictingSetCapture(body_list, param_names[0..arity], null)) return null;
```

With the guard, the lambda drops to `emitLambdaViaEval`, the whole application runs in the VM, and the output is 95. Please also add this inline shape as a test case.

tests/scheme/compile/native-set-capture-divergence-1422.sh
```diff
@@ -0,0 +1,94 @@
...
+  (let ((g (lambda () x)))
+    (set! x 42)
+    (g)))
+(write (f 1))
+(newline)' \
+'42'
+
+# 4. Shadowed param: the lambda (x) shadows f's x, so it does NOT capture f's x.
+#    The function should still compile natively.
+check set-shadowed-param \
```
baijum, line 106: Cases 4 and 5 claim the function "should still compile natively", but the assertion is output-only — interpreter fallback prints the same values, so an over-rejection regression (e.g. broken shadow handling in `exprHasCapturingLambda`) would pass silently. You can pin the native path directly:

```bash
"$KAAPPI_ABS" --emit-llvm -o "$DIR/$name.ll" "$DIR/$name.scm" > /dev/null 2>&1
grep -q '^define i64 @lambda' "$DIR/$name.ll" || { echo "FAIL: $name — expected native fn"; fail=1; }
```

and inversely assert zero `@lambda` definitions for cases 1–3 to pin the fallback. (Verified both greps behave as expected on this branch.)

src/fuzz_gen_native.zig
```diff
@@ -180,7 +183,7 @@ fn genLambda(g: *Gen, ctx: *Ctx, arity: u8, variadic: bool) Error!void {
     for (params, 0..) |p, i| {
         if (i > 0) try g.emit(" ");
         try g.emit(p);
-        try g.pushBinding(p, .int);
+        try g.scope.append(g.gpa, .{ .name = p, .kind = .int, .settable = false });
```
baijum: Nit: five call sites now reach into `g.scope.append(g.gpa, .{ ..., .settable = false })` directly with the same literal. `fuzz_gen.zig` already wraps this exact pattern as `pushPatternVar`; a `pushNonSettable(name, kind)` helper next to `pushBinding` would keep the scope-manipulation API in one place and make the intent greppable.

baijum (commented):
## Review

Verified locally on the PR head (5abf0f36): `zig build test` passes, all 11 `tests/scheme/compile/*.sh` pass, and I confirmed via `--emit-llvm` that the guard does what it claims — the issue reproducer emits **0** native lambda functions (falls back to the interpreter), while shadowed-param case 4 still emits **2** (stays native). The guard itself is sound: the fixed `[17]` arrays are safe given the `arity >= 16` rejection upstream, quote-skipping matches the sibling scanners, and the shadowing-unaware `set!` scan only errs conservatively (extra fallback, never wrong output). The fuzz-generator change is correctly reasoned and documented, and the new test script is picked up automatically by `run-all.sh` (mode 755, `run_shell_suite` glob).

**One substantive gap** (see inline comment): the identical divergence still reproduces on this branch through `tryCompilePureLambdaAsNativeClosure`, which compiles whole functions too but has no set!/define check. Since the PR says *Closes #1422*, either extend the guard to that tier (a one-line addition of the same helper) or narrow the claim and keep the issue open for the residual case.

Also worth noting: **no build/test CI ran on this branch** — only CodeRabbit reported. The core `ci.yml` apparently didn't trigger; my local runs stand in for it, but check why before relying on the merge gate.

**Verdict: not ready to merge as-is.** Close the tier-2 gap (recommended, it's small) or reword `Closes` → `Partially addresses`, then this is good to go.

baijum:
Re-reviewed at fc9df1a5 — all three review findings are addressed and verified:

1. **Tier-2 gap (blocking)**: `bodyHasConflictingSetCapture` is now also called in `tryCompilePureLambdaAsNativeClosure` (after the rest-param rejection, with `rest_name = null`). The inline reproducer that printed `6` natively on the previous head now prints `95`, matching the VM, and is pinned as test case 4.
2. **Test tier pinning**: `check` now takes an `expect_native` flag and asserts the compilation tier via `--emit-llvm` + `grep '^define i64 @lambda_'` — fallback pinned for cases 1–4, native pinned for cases 5–6. Over-rejection regressions can no longer pass silently.
3. **Helper extraction**: `pushNonSettable` added to `Gen` in `fuzz_gen.zig`; all five call sites converted.

Local validation on fc9df1a5: `zig build test` passes, all 11 `tests/scheme/compile/*.sh` pass, and the core CI matrix has triggered on this push (format + wasm already green).

**Ready to merge once the pending CI test matrix goes green.**

Issue #809, opened 2026-07-02
Closures capturing 27+ variables panic the interpreter (u8 overflow in allocClosure byte accounting; @intCast panic in addUpvalue past 255)


baijum:
## Summary

A closure that captures **27 or more** variables from enclosing scopes crashes the interpreter with a Zig panic (`integer overflow`) at closure-creation time. Capturing 26 works. Independently, capturing **more than 255** variables panics at *compile* time (`integer does not fit in destination type`). Both are safety-check aborts of the whole process on legal (if unusual) R7RS programs, instead of either working or raising a proper Scheme/compile error.

## Locations

- `src/types.zig:222` — `upvalue_count: u8` on `Function`.
- `src/memory.zig:236-238` — `allocClosure`:
  ```zig
  self.bytes_allocated += @sizeOf(Closure) + upvalue_count * @sizeOf(Value);
  self.profileAlloc(@sizeOf(Closure) + upvalue_count * @sizeOf(Value));
  ```
  `upvalue_count` is `u8`, so the whole right-hand side is computed in u8 arithmetic. With `@sizeOf(Closure)` = 40, the expression overflows u8 as soon as `upvalue_count >= 27` (27*8 + 40 = 256 > 255) — runtime panic on every closure creation with that many captures.
- `src/compiler.zig:231` — `addUpvalue`: `self.func.upvalue_count = @intCast(self.upvalues.items.len);` panics when a function accumulates more than 255 upvalues, with no graceful `CompileError`.

## Reproduction (v0.11.1)

Generate a lambda whose body defines N locals and an inner lambda referencing all of them (nested `(+ v0 (+ v1 ...))` to stay under the 256-arg call limit):

```scheme
(define f
  (lambda ()
    (define v0 0)
    ...
    (define v26 26)
    (lambda () (+ v0 (+ v1 ... (+ v25 v26))))))
(display ((f))) (newline)
```

- N = 26: prints 325 (correct).
- N = 27..255: panics at runtime:
  ```
  thread ... panic: integer overflow
  ... in allocClosure (kaappi)
  src/vm_dispatch.zig:695 ... runUntil
  ```
- N = 256+: panics at compile time:
  ```
  thread ... panic: integer does not fit in destination type
  ... in addUpvalue (kaappi)
  src/compiler.zig:215 ... resolveUpvalue
  ```

Python one-liner to generate the repro:

```python
n = 27
defs = "\n    ".join(f"(define v{i} {i})" for i in range(n))
[...]

Issue #1038, opened 2026-07-04
Epic: retire the legacy compileExpr path (dual compilation paths are producing divergence bugs)


baijum:
## Problem
The compiler has two paths: the IR path (`compileFromNode`) and the legacy `compileExpr`. Current shape:
- Only 11 of 34 `NodeTag`s are truly IR-compiled; 18 carry raw S-exprs (`SexprArgs`, `src/ir.zig:77-93`) and dispatch back to legacy form compilers (`src/compiler_ir.zig:29-49`); `.passthrough` covers macros, quote, apply, `(define (f) ...)`.
- Every legacy form compiler recurses via `self.compileExpr` (**59 call sites**: bindings 16, lambda 11, advanced 10, passthrough 9, conditionals 9, compiler 3, compiler_ir 1) — the first derived form in a subtree drops the whole subtree off the IR path (no optimization, no self-tail-call, no line tables).
- Each truly-IR form has a duplicated legacy twin (compileIfFromIR vs passthrough.compileIf; compileCallFromIR vs passthrough.compileCall incl. a third constant-fold table copy at `src/compiler_passthrough.zig:215-305`; compileDefineFromIR/compileSetFromIR vs compiler_lambda versions — and `compileSetFromIR` lacks the `is_global_alias` handling legacy has at `src/compiler_lambda.zig:568-579`).

The divergence class has already produced #1025, #1026, and #1035.

## Plan (phased, each independently shippable)
1. #1025 — macro visibility in IR lambda bodies
2. #1026 — letrec* desugaring in IR lambda bodies
3. #1035 — port self-tail-call + line tables to the IR path
4. Add `compileExprViaIR(self, expr, dst, tail)` and convert the 59 recursion sites so subtrees re-enter lowering/optimization
5. Move macro expansion into IR lowering, then delete the legacy twins and the keyword chain in `compileForm` (`src/compiler.zig:611-882`)

Payoff: roughly -800 to -1000 lines and elimination of the divergence bug class. Size: L overall. Risk: medium-high (hygiene + GC rooting subtleties: #788/#790/#877/#1010 live in this code) — hence phases. Tests: R7RS suite (1391), `run-all.sh` (smoke/hygiene/continuations), `tests_ir.zig`, `tests_macros.zig`. Remember to delete stale `.sbc` caches when testing.

As of c0d4f10. From the 2026-07-04 refactoring analysis.

PR #1427, opened 2026-07-11
Make the unit suite run green under -Dgc-stress=true

### LINKED ISSUE #1401: Unit test suite crashes under -Dgc-stress=true (~440 of 690 tests)

baijum:
Fixes #1401.

The gc-stress build was never green (answering the issue's question 1): the test harness itself was broken, and it was masking a layer of real rooting bugs underneath. This lands the fix that was developed in a prior session's worktree, reconciled with the arithmetic rooting that already merged via #1421, plus the long tail of failures found by running the suite to completion.

## Harness

- `makeTestVM` returned the `VM` struct **by value**, but `vm_instance` and the GC root marker reach the VM by pointer — so every collection during `registerAll` swept the very globals being registered. It now returns a heap-allocated `*VM` (`heap_owned`, destroyed by `deinit()`); the ~312 call sites compile unchanged. The same bootstrap order (setVMInstance before registerAll) is applied to `bench.zig` and `kaappi_lsp.zig`; `main.zig` was already correct.
- `vm_eval.eval` registers the VM on entry, so compile-phase allocations mark the right VM's roots in multi-VM (thread/deepcopy) tests.
- New repeatable `-Dtest-filter` build option for narrowing test runs.

## Runtime and compiler rooting bugs the harness was masking

- **GC allocators** now copy caller slices *before* `maybeCollect()` and root `Value` slices across the collection via a new `slice_roots` hook (`allocVector`, `allocNativeClosure`, `allocRecordInstance`, `allocMultipleValues`, `makeList`). Previously e.g. `negate`/`abs` collected first, freeing the source bignum and reusing its limbs for the copy — the `@memcpy arguments alias` crash.
- **`maybeCollect`'s stress branch** now honors `no_collect`/`memory_limit` deferral exactly like the normal path (the issue's question 2: the memory_limit test is compatible, not exempt).
- **Body-local `define-syntax` transformers** live only in compiler-local macro maps the GC cannot see; they're now rooted in `extra_roots` for the duration of the top-level compile, at all four registration sites (`compileDefineSyntax`, `compileLetrecSyntax`, and the body scans in `compiler_ir.zig` / `compiler_lambda.zig`).
- **`emitLlvmFile`** defers collection across its read → lower → emit batch: IR nodes reference sexpr values that nothing roots. This was latent in normal builds too, for source files big enough to cross the GC threshold mid-compile.

## Test fixes

- Root Zig locals held across allocations (bignum, bytecode_file, deepcopy, srfi18, robustness) and eval results held across a second eval (tests_io read-back pattern).
- Scale the "GC stress:" churn loops down on stress builds — every allocation already collects there, and the 5000-iteration string loop peaked **19.7 GB RSS** under the testing allocator (DebugAllocator metadata churn; the production allocator runs the same loop at 4 MB), OOM-killing full-suite runs.
- Skip on stress builds: the 300-capture #809 width test (~7 GB RSS, property orthogonal to rooting; the 27-variable sibling keeps coverage) and the #1253 frame-depth test.
- The #958 mark-bit scan forces one full cycle and keeps the parent GC quiescent (threshold-driven) while the child thread's GC still stresses — the parent's own minor-GC marks legitimately persist between cycles and were false-positiving the scan.
- New smoke test `tests/scheme/smoke/bignum-gc-alias-1414.scm`.

## CI / docs

- Re-enables the `gc-stress` fuzz variant in `fuzz.yml` with a 300-minute job timeout (the pre-fuzz unit-test phase dominates wall time: ~40 min locally at 100% CPU).
- Documents the invariants in `.claude/rules/gc-safety.md`, `docs/dev/testing.md`, and `docs/dev/fuzzing.md`.

## Test plan

- `zig build test` — green
- `zig build test -Dgc-stress=true` — from 248 pass / 6 fail / 436 crash to 722 pass / 6 skip / 2 crash in the last full run; the 2 remaining OOM-kill crashes are fixed and verified individually, and a final full-suite confirmation run is in flight
- `bash tests/scheme/run-all.sh` — 1827 pass, 0 fail (432 Scheme files + R7RS suite)
- `kaappi compile` end-to-end smoke test after the `emitLlvmFile` change

### DIFF
--- a/.claude/rules/gc-safety.md
+++ b/.claude/rules/gc-safety.md
@@ -19,12 +19,37 @@ string-set!, record field mutation):
 
 - **Allocator Value arguments are auto-rooted**: `allocPair(car, cdr)` and
   other `allocXxx` functions that take Value arguments root them internally
-  via `arg_roots` before `maybeCollect()`. Callers do NOT need to root Values
-  that are passed directly as allocator arguments. However, Values held in
-  local variables across multiple allocation calls still need manual rooting.
+  via `arg_roots` before `maybeCollect()`. Value slices (`allocVector`,
+  `allocNativeClosure`, `allocMultipleValues`, `allocRecordInstance`,
+  `makeList`) are rooted via `slice_roots`. Callers do NOT need to root
+  Values that are passed directly as allocator arguments. However, Values
+  held in local variables across multiple allocation calls still need
+  manual rooting.
+
+- **Allocators copy caller memory before collecting**: `allocXxx` functions
+  that receive a slice (string bytes, limbs, Values) copy it into raw memory
+  *before* `maybeCollect()`, so a slice that aliases another heap object's
+  storage (e.g. `bignum.limbs`, `SchemeString.data`, `vec.data`) stays valid
+  even if that object is collected. Preserve this order when editing them
+  (#1401).
+
+- **Root fresh results between allocations**: a Value returned by one
+  allocating call is unrooted; holding it in a local while a second call
+  allocates lets the collection free it — and the second allocation often
+  lands in the recycled memory, silently aliasing the two (#1414 made every
+  bignum/bignum division return 1 this way). Root the first result (e.g.
+  `gc.rootedSlot`) before computing the second.
 
 - **Root Function* before vm.execute()**: `execute()` allocates a closure
   wrapper internally.
+
+- **`vm_instance` must point at the live VM before anything allocates**: the
+  GC root marker finds the globals/macros/libraries through the
+  `vm_instance` threadlocal. Call `vm_mod.setVMInstance(vm)` right after
+  `VM.init` — before `registerAll` — and never let a VM struct move after
+  that (heap-allocate it or keep it in a stable stack frame). A stale or
+  null `vm_instance` means collections sweep everything the globals
+  reference (#1401).
 
 Dangerous pattern:
 ```zig
--- a/.claude/skills/do-stress-test/SKILL.md
+++ b/.claude/skills/do-stress-test/SKILL.md
@@ -1,0 +1,224 @@
+---
+name: do-stress-test
+description: Run the Kaappi unit suite under -Dgc-stress=true on a real x86-64 Linux machine using a DigitalOcean droplet with a 3-hour lifetime. Use when the user asks to stress-test on Linux, run the gc-stress suite on real hardware, verify GC rooting under stress on x86-64, or stress test on DigitalOcean. Complements /do-linux-test (normal suite, ~1-hour droplet) — this variant runs the collection-per-allocation build, which takes hours, not minutes.
+---
+
+# DigitalOcean GC-Stress Test (3-hour droplet)
+
+Build Kaappi with `-Dgc-stress=true` (a collection is attempted on every
+allocation) and run the full unit suite on a temporary x86-64 DigitalOcean
+droplet. The droplet self-destructs after **3 hours** and is **always**
+destroyed when done.
+
+Key differences from `/do-linux-test`:
+- The stress suite is CPU-bound for **1.5–3 hours** (every test VM bootstrap
+  performs tens of thousands of collections; ~40 min on an M-series Mac).
+- It runs detached (`nohup`) on the droplet and is **polled**, never awaited
+  in a single SSH command — no SSH command may outlive the Bash tool timeout.
+- The testing allocator's metadata churn inflates RSS by gigabytes, so the
+  droplet gets more memory plus a swap file.
+
+## Pre-flight
+
+### 1. Record branch and check for unpushed work
+
+```bash
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
+echo "Branch: $BRANCH"
+git status --porcelain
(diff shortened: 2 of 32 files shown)


baijum:
Review comments addressed in dcdbfade:

**Fixed:**
- `VM.initForThread` now roots each standard port before allocating the next, mirroring `init()`'s #1013 pattern — this was the real find of the review: a fresh child thread has no `vm_instance` yet, so a stress collection during the second `allocPort` swept the first port. (The suggested "set vm_instance first" isn't possible — the struct doesn't exist until `initForThread` returns — but interleaved rooting closes the window; `threadEntryFn` already registers the VM before any subsequent allocation.)
- `-Dtest-filter` now applies to `thottam-tests` as well.
- `copied_fn` is rooted across the second `deepCopy` in tests_deepcopy — it survives today only because `deepCopyValue` holds `no_collect`, and the gc-safety rule is to root values held across allocating calls regardless.
- The transformer-root comments in `compiler_ir.zig`/`compiler_macro.zig` now state the correct release point (the enclosing lambda compile's `roots_base` truncation for body scopes, `compileExpression*` at top level).
- `docs/dev/testing.md` example imports completed.

**Not taken, with reasons:**
- *CLAUDE.md `-Dtest-filter=tests_io` example*: the filter matches fully-qualified test names, which include the module prefix — verified: `zig build test -Dtest-filter=tests_io --summary all` runs exactly the 30 tests_io tests. `docs/dev/testing.md` documents this deliberately.
- *Root `neg` in the bignum tests*: `neg` is never held across a subsequent allocation — it's the last allocation in each test, and `add()` copies both operands' limbs before its own collection point (that ordering is part of this PR). Rooting it would be dead code implying a hazard that isn't there.
- *compiler_ir/compiler_lambda body-scan duplication*: agreed it's real — this PR had to patch both sides. It's tracked by the refactoring roadmap (epic #1038, retire the legacy compileExpr path), which is the right place to collapse the duplication rather than a GC-fix PR.

PR #1434, opened 2026-07-11
Build chibi-scheme from source in oracle-diff CI

### LINKED ISSUE #1429: Fuzz CI: infrastructure or build failure
Automated report from the [Fuzz workflow](https://github.com/kaappi/kaappi/actions/runs/29141262379): a job failed WITHOUT producing a finding artifact. That usually means an infrastructure problem (toolchain download, apt flake, runner OOM), a build or unit-test failure unrelated to fuzzing, or a job-level timeout. A timeout can be a genuine hang: every generated program is individually time-bounded, so a whole job hitting its timeout is itself suspicious.

> [!NOTE]
> Close this issue once the failed run is explained (or the underlying breakage is fixed); further artifact-less failures append comments here.

---

**Failed job(s):** `oracle-diff` — [run #3](https://github.com/kaappi/kaappi/actions/runs/29141262379) (`schedule`)

The failed job(s) uploaded no finding artifact, so this is not (necessarily) a fuzz finding. Check the run log for a setup or build step failure, a unit-test failure, or a job cancelled at its timeout.

Download everything: `gh run download 29141262379 -R kaappi/kaappi` (artifacts are retained for 90 days).

baijum:
## Summary

Fixes #1429.

- Build chibi-scheme from source at tag `0.11` instead of using `apt-get install` — Ubuntu noble ships 0.9.1 which is too old for the portable-subset programs (985/1000 false divergences).
- Pin `upload-artifact` to v7.0.1 SHA (matching the other upload steps) — the unpinned `@v7` tag resolved to a newer version whose `archive: true` default prevented the report job from finding `.scm` marker files, misclassifying the failure as an infrastructure issue.
- Update `docs/dev/fuzzing.md` to document the version pinning.

## Test plan

- [x] Ran `oracle-diff.sh 100 3000` locally with chibi 0.12.0 — 0 divergences (the same seed range that was 985/1000 divergent in CI with chibi 0.9.1)

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -188,10 +188,9 @@ jobs:
   # subset (#1396). This is the only oracle that catches conformance bugs
   # where BOTH Kaappi evaluation paths agree but are wrong. Two interpreter
   # runs per program (no linking), so it is much cheaper per seed than
-  # native-diff. The oracle version is whatever ubuntu-latest's apt ships;
-  # the harness records the exact version in the log and in any divergence
-  # artifact, and CHIBI can pin a specific binary if the distro version
-  # ever proves flaky.
+  # native-diff. The oracle is built from source at a pinned tag — Ubuntu
+  # noble's apt ships 0.9.1 which is too old for the generated programs
+  # (#1429). Bump the tag below when a newer release is needed.
   oracle-diff:
     runs-on: ubuntu-latest
     timeout-minutes: 30
@@ -210,8 +209,10 @@ jobs:
 
       - name: Install Chibi Scheme (the oracle)
         run: |
-          sudo apt-get update
-          sudo apt-get install -y chibi-scheme
+          git clone --depth=1 --branch=0.11 https://github.com/ashinn/chibi-scheme.git /tmp/chibi-src
+          cd /tmp/chibi-src
+          make -j"$(nproc)"
+          sudo make install
           chibi-scheme -V
 
       - name: Build interpreter and generator
@@ -234,7 +235,7 @@ jobs:
       # header (Kaappi bug / oracle bug / generator leak).
       - name: Upload divergences
         if: failure()
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: oracle-diff-divergences
           path: oracle-diff-results/
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -188,9 +188,9 @@ jobs:
   # subset (#1396). This is the only oracle that catches conformance bugs
   # where BOTH Kaappi evaluation paths agree but are wrong. Two interpreter
   # runs per program (no linking), so it is much cheaper per seed than
-  # native-diff. The oracle is built from source at a pinned tag — Ubuntu
+  # native-diff. The oracle is built from source at a pinned tag+SHA — Ubuntu
   # noble's apt ships 0.9.1 which is too old for the generated programs
-  # (#1429). Bump the tag below when a newer release is needed.
+  # (#1429). Bump the tag and SHA below when a newer release is needed.
   oracle-diff:
     runs-on: ubuntu-latest
     timeout-minutes: 30
@@ -209,8 +209,9 @@ jobs:
 
       - name: Install Chibi Scheme (the oracle)
         run: |
-          git clone --depth=1 --branch=0.11 https://github.com/ashinn/chibi-scheme.git /tmp/chibi-src
+          git clone --depth=1 --branch=0.12 https://github.com/ashinn/chibi-scheme.git /tmp/chibi-src
           cd /tmp/chibi-src
+          test "$(git rev-parse HEAD)" = 08e23be118e7de8cf29d606bc4793371c0ace680
           make -j"$(nproc)"
           sudo make install
           chibi-scheme -V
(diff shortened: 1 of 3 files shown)


.github/workflows/fuzz.yml
```diff
@@ -210,8 +209,10 @@ jobs:
 
       - name: Install Chibi Scheme (the oracle)
         run: |
-          sudo apt-get update
-          sudo apt-get install -y chibi-scheme
+          git clone --depth=1 --branch=0.11 https://github.com/ashinn/chibi-scheme.git /tmp/chibi-src
```
baijum: The pin doesn't match what was validated: the test plan ran with brew's chibi **0.12.0**, but this pins **0.11** (upstream also has a `0.12` tag). If 0.11 is deliberate (one release behind for stability), fine — but as written, the exact oracle CI will run has never been exercised against these programs. Cheap way to close the gap before the nightly does it for you:

```
gh workflow run fuzz.yml --ref fix/1429 -f oracle-diff-count=100 -f oracle-diff-base=3000
```

same seed range as the local run, real CI oracle. Alternatively pin `0.12` to match the validated version.

Also +1 to the SHA-pin comment above; a variant that keeps the cheap shallow clone is asserting after checkout:

```yaml
test "$(git rev-parse HEAD)" = b1a370b218cb13cb8f6c98a087185b0357df8e89
```

which fails closed if the tag is ever re-pointed.

.github/workflows/fuzz.yml
```diff
@@ -234,7 +235,7 @@ jobs:
       # header (Kaappi bug / oracle bug / generator leak).
       - name: Upload divergences
         if: failure()
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
```
baijum, line 244: Worth one line of comment here (or on the report job) saying *why* this must stay at v7.0.1: newer upload-artifact defaults to `archive: true`, which hides the `seed-*.scm` markers from the report job's `find` and turns real findings into "infrastructure" issues (#1429). That constraint currently lives only in this PR's description — the next bulk action-pin bump will touch all three upload steps in this file, and the breakage only manifests on the next failing nightly, silently.

docs/dev/fuzzing.md
```diff
@@ -239,8 +239,9 @@ are wrong. The oracle is Chibi Scheme (closest R7RS-small alignment, small
 install) under its default invocation; the harness prints and records the
 exact version, and `CHIBI=/path/to/binary` pins one explicitly. Note the
 repo's `(chibi test)` shim is implemented *in Kaappi* — the oracle must be a
-real installed `chibi-scheme` (`apt-get install chibi-scheme` /
-`brew install chibi-scheme`).
+real installed `chibi-scheme` (`brew install chibi-scheme` for local use;
+CI builds from source at a pinned tag — Ubuntu noble's apt ships 0.9.1
```
baijum: `tests/fuzz/oracle-diff.sh` still gives the opposite advice in its header (~line 15): "Install it with `apt-get install chibi-scheme` or `brew install chibi-scheme`". On noble that installs the 0.9.1 this PR is escaping, and the script header is the document the triage protocol and the auto-filed issues point at. Suggest updating it in this PR to match this paragraph.

baijum (commented):
Reviewed against issue #1429 and the report job's marker-file mechanism. Both fixes check out.

**Verified:**
- `043fb46d1a93c77aae656e7c1c64a875d1fc6a0a` is exactly what `v7.0.1` resolves to in `actions/upload-artifact` (lightweight tag → commit), and it matches the 8 other `upload-artifact` pins across `ci.yml`, `release.yml`, and `fuzz.yml` — this was the last unpinned one in the repo.
- The misclassification mechanism is real: the report job gates finding-vs-infrastructure on `find artifacts/oracle-diff-divergences -name 'seed-*.scm'` (fuzz.yml ~line 408), which an archive-wrapping upload breaks — consistent with #1429 being auto-filed under the infrastructure title.
- Tag `0.11` exists on ashinn/chibi-scheme and there is no same-named branch, so `--branch=0.11` unambiguously fetches the tag. `make install` lands in `/usr/local/bin` (on PATH, where oracle-diff.sh looks), and `chibi-scheme -V` in the same step self-verifies the install.
- The job already runs with `persist-credentials: false` and workflow-level `permissions: contents: read`, which bounds the blast radius of building third-party code with sudo. CodeRabbit's SHA-pin suggestion is still worth taking — it's a one-line addition.

**Inline comments:** one doc gap I'd fix in this PR (`oracle-diff.sh`'s own header still recommends the too-old apt package), one question about pinning `0.11` when local validation used 0.12.0, and one suggestion to write down *why* the upload steps must stay at v7.0.1.

**Merge readiness:** the substance is correct; fine to merge once the in-flight test matrix finishes (no Zig code touched, so it should stay green). Before relying on the nightly, consider one `workflow_dispatch` run of fuzz.yml on this branch to exercise the pinned 0.11 build end-to-end:

```
gh workflow run fuzz.yml --ref fix/1429 -f oracle-diff-count=100 -f oracle-diff-base=3000
```

That replays the same seed range as the local test plan, but with the actual CI oracle.

baijum:
Follow-up on my review: commit a00fd187 addresses all four points (0.12 pin matching the validated version, SHA assertion after the clone — verified `0.12` is a lightweight tag at `08e23be1`, the `IMPORTANT` comment on the upload-artifact pin, and the `oracle-diff.sh` header fix).

End-to-end validation on this branch: [run 29144225273](https://github.com/kaappi/kaappi/actions/runs/29144225273) (`workflow_dispatch`, `oracle-diff-count=100`, `oracle-diff-base=3000`). The oracle-diff job built chibi **0.12.0 "magnesium"** from source (SHA assert passed) and reports:

```
oracle-diff: seeds 3000..3099
oracle-diff: 100 compared, 0 skipped, 0 divergent
```

— the same seed range that was 985/1000 divergent with apt's 0.9.1 in [run 29141262379](https://github.com/kaappi/kaappi/actions/runs/29141262379) (today's scheduled failure that filed #1429). native-diff and the default fuzz job in the validation run are also green.

All PR checks pass. **Ready to merge** — merging before tonight's 02:47 UTC schedule stops the nightly from failing again, and the merge auto-closes #1429.

PR #1436, opened 2026-07-11
Store a persistent mark worklist on the GC struct

### LINKED ISSUE #1428: GC: markValue allocates a fresh 8 KB worklist per invocation — megabytes of churn per collection
## Problem

`markValue` in `src/gc_collect.zig` (~line 506) creates a fresh `std.ArrayList(Value)` and calls `ensureTotalCapacity(gc.allocator, 1024)` (8 KB) on **every invocation**, then frees it on return:

```zig
pub fn markValue(gc: *GC, v: Value) void {
    var worklist: std.ArrayList(Value) = .empty;
    defer worklist.deinit(gc.allocator);

    worklist.ensureTotalCapacity(gc.allocator, 1024) catch
        @panic("GC mark: worklist OOM");
    ...
}
```

`markRoots` calls `markValue` once per root — arg roots, root buffer, extra roots, FFI callback slots, every interned symbol, plus all VM-owned roots via `root_marker` (registers, frames, handlers, winds, globals). That's **thousands of calls per collection**, so every collection allocates and frees megabytes of transient worklist buffers. `markObjectContents` (remembered-set tracing during minor collections) hits the same path per field.

## Impact

- Under `-Dgc-stress=true` (collection per allocation) this churn is the **dominant allocator traffic**.
- It interacts badly with `std.testing.allocator`'s metadata retention: the 5000-iteration string-append loop in `tests_robustness.zig` ("GC stress: many string allocations") peaked **~19.7 GB RSS** in the unit-test process. The production allocator handles the same loop at ~4 MB — so this is churn amplification, not a leak.
- In normal (non-stress) builds it's still avoidable per-collection overhead in the hottest GC phase (marking).

## Proposed fix

Store a reusable worklist buffer on the `GC` struct:

- `mark_worklist: std.ArrayList(Value) = .empty` — drained empty by each `markValue` call, capacity retained between uses, freed in `GC.deinit`.
- `markValue` uses `&gc.mark_worklist` instead of a fresh list; the per-call `ensureTotalCapacity(1024)` can be dropped since capacity persists after first growth.

(Alternative: hoist one worklist per collection cycle through `markRoots`, but callers outside `gc_collect.zig` — `vm.zig` root marker, `fiber.zig`, `ffi_callback.zig` — call `gc.markValue` directly, so a struct field is far less invasive.)

## Things to watch

- **Re-entrancy:** marking a `.fiber` object calls `fiber_mod.markFiberState`, which calls `gc.markValue` again while the outer drain loop is live. Sharing one buffer is safe — marking is idempotent and order-independent, so an inner drain loop may consume entries pushed by an outer one — but this deserves a comment and careful review.
- **Child-thread GCs (SRFI-18):** each `GC` instance must own its buffer (a struct field with a default gives this for free; `initForThread` needs no change). The cross-thread deep-copy path (`gc_deep_copy.zig`) doesn't use `markValue`, so it's unaffected.
- **Retention:** the buffer retains its peak capacity for the GC's lifetime. Peak worklist size is bounded by the fan-out of the live object graph (e.g. large vectors/hash tables push O(n) entries), which the heap itself dwarfs, so unbounded retention is likely acceptable — but a shrink policy after collection is an option if it ever matters.

## Verification

- `zig build test`
- Timed comparison of the GC stress tests: `zig build test -Dgc-stress=true -Dtest-filter="GC stress:"` — note `build.zig` currently has **no `-Dtest-filter` option**; add one wired to `addTest`'s `.filters` (useful on its own).
- RSS comparison on the string-append stress test (`/usr/bin/time -l`) to confirm the ~19.7 GB testing-allocator blow-up disappears.
- A regression test that marking drains the worklist and retains capacity across collections (e.g. in `src/memory.zig`'s test block).

baijum:
## Summary

Fixes #1428.

- Move the `markValue` worklist from a per-call local to a `mark_worklist` field on the `GC` struct so its capacity persists across calls — eliminates thousands of 8 KB alloc/free cycles per collection.
- Add a `marking` re-entrancy guard so fiber marking (`markFiberState` → `gc.markValue`) shares the buffer safely: re-entrant calls push items and return, the outermost call owns the drain loop.
- RSS on the "GC stress:" test filter drops from ~19.7 GB to ~1.15 GB.

## Test plan

- [x] `zig build test` — 731/731 pass
- [x] `zig build test -Dgc-stress=true` — all pass
- [x] Full Scheme suite — 1827 pass, 0 fail
- [x] RSS comparison: `/usr/bin/time -l zig build test -Dgc-stress=true -Dtest-filter="GC stress:"` — ~1.15 GB (was ~19.7 GB)
- [x] New unit test: `mark worklist retains capacity across collections`

### DIFF
--- a/src/gc_collect.zig
+++ b/src/gc_collect.zig
@@ -525,6 +525,12 @@ pub fn markValue(gc: *GC, v: Value) void {
         markValueInner(gc, item, &gc.mark_worklist);
     }
     gc.marking = false;
+
+    // Cap retained capacity so one pathologically wide object (e.g. a
+    // 10M-element vector) doesn't keep ~80 MB allocated forever.
+    const max_retained = 64 * 1024;
+    if (gc.mark_worklist.capacity > max_retained)
+        gc.mark_worklist.clearAndFree(gc.allocator);
 }
 
 fn markValueInner(gc: *GC, v: Value, worklist: *std.ArrayList(Value)) void {
--- a/src/gc_collect.zig
+++ b/src/gc_collect.zig
@@ -508,21 +508,23 @@ fn markRoots(gc: *GC) void {
 
 pub fn markValue(gc: *GC, v: Value) void {
     // Use an explicit worklist to avoid native stack overflow on deeply
-    // nested pair/vector structures (issue #864). Values that would
-    // previously be handled by a recursive markValue call are pushed
-    // onto the worklist and processed iteratively.
-    var worklist: std.ArrayList(Value) = .empty;
-    defer worklist.deinit(gc.allocator);
+    // nested pair/vector structures (issue #864). The worklist lives on
+    // the GC struct so its capacity persists across calls — no per-call
+    // alloc/free churn (issue #1428).
+    const is_root_call = !gc.marking;
+    gc.marking = true;
 
-    worklist.ensureTotalCapacity(gc.allocator, 1024) catch
-        @panic("GC mark: worklist OOM");
+    markValueInner(gc, v, &gc.mark_worklist);
 
-    markValueInner(gc, v, &worklist);
+    // Re-entrant call (e.g. markFiberState → gc.markValue): the outer
+    // drain loop will process items we just pushed.
+    if (!is_root_call) return;
 
-    while (worklist.items.len > 0) {
-        const item = worklist.pop().?;
-        markValueInner(gc, item, &worklist);
+    while (gc.mark_worklist.items.len > 0) {
+        const item = gc.mark_worklist.pop().?;
+        markValueInner(gc, item, &gc.mark_worklist);
     }
+    gc.marking = false;
 }
 
 fn markValueInner(gc: *GC, v: Value, worklist: *std.ArrayList(Value)) void {
--- a/src/memory.zig
+++ b/src/memory.zig
@@ -126,6 +126,8 @@ pub const GC = struct {
     source_lines: std.AutoHashMap(Value, u32) = undefined,
     stats: GcStats = .{},
     minor_cycle_count: u32 = 0,
+    mark_worklist: std.ArrayList(Value) = .empty,
+    marking: bool = false,
 
     pub const INITIAL_ROOT_CAPACITY: usize = 1024;
     pub const MAX_ROOT_CAPACITY: usize = 65536;
@@ -182,6 +184,7 @@ pub const GC = struct {
         self.allocator.free(self.root_buffer);
         self.extra_roots.deinit(self.allocator);
         self.remembered_set.deinit(self.allocator);
+        self.mark_worklist.deinit(self.allocator);
         self.source_lines.deinit();
     }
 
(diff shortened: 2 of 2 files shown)


src/gc_collect.zig
```diff
@@ -508,21 +508,23 @@ fn markRoots(gc: *GC) void {
...
 
-    markValueInner(gc, v, &worklist);
+    // Re-entrant call (e.g. markFiberState → gc.markValue): the outer
+    // drain loop will process items we just pushed.
+    if (!is_root_call) return;
 
-    while (worklist.items.len > 0) {
-        const item = worklist.pop().?;
-        markValueInner(gc, item, &worklist);
+    while (gc.mark_worklist.items.len > 0) {
```
baijum, line 523: The worklist now retains its lifetime high-water capacity: one pathologically wide object (e.g. a single 10M-element vector pushes `len-1` entries in one go) grows this buffer to ~80 MB, and it stays allocated until the GC dies — per thread, since each thread GC has its own worklist. That's the right trade for #1428, but a post-drain cap would bound the pathological case at trivial cost:

```zig
const MAX_RETAINED_CAPACITY = 64 * 1024; // 8 bytes each → 512 KB
if (gc.mark_worklist.capacity > MAX_RETAINED_CAPACITY)
    gc.mark_worklist.clearAndFree(gc.allocator);
```

after `gc.marking = false;`. The buffer regrows on the next deep mark, so steady-state reuse is unaffected. Non-blocking — also fine to just accept the retention and say so in the comment above.

baijum (commented):
Reviewed and verified locally at 0b65f7ed. The core change is correct — I traced every `markValue` call site and the re-entrancy guard holds.

**Correctness analysis**

- *Root calls* (drain + reset `marking`): `markRoots` per root, `markObjectContents` per remembered-set object (minor GC), `FiberScheduler.markRoots`, `markVMRoots` via `root_marker`, and the FFI-callback root marker. Each fully drains before returning, so sweep never starts with items left on the worklist.
- *Re-entrant calls* exist on exactly one path: `markValueInner → markFiberState → gc.markValue`. Push-and-return is safe there because `markValueInner` sets the mark bit *before* pushing children, nothing inspects mark bits mid-mark, and the outermost drain completes before `collect()` proceeds to sweep.
- No error-return path exists in the mark phase (all OOM → `@panic`), so `marking` can never be left `true`.
- GC instances are per-thread, so there is no concurrent access to the shared worklist; the `obj.owner != gc.id` check (#958) prevents cross-GC marking from re-entering another GC's `markValue`.
- `init`/`initForThread` build the struct via literals, so the new fields' defaults apply, and `deinit` frees the worklist. ✓

**Verified locally** (macOS ARM, PR head 0b65f7ed):

- `zig build test` — pass
- `zig build test -Dgc-stress=true -Dtest-filter="mark worklist"` — pass
- `zig build test -Dgc-stress=true -Dtest-filter=fiber -Dtest-filter=tests_srfi18` — pass (this hammers the re-entrant `markFiberState` path with a collection on every allocation)
- `/usr/bin/time -l` on the `"GC stress:"` filter — max RSS **1.12 GB**, corroborating the claimed ~1.15 GB

**On the CodeRabbit comment** (new test doesn't cover the re-entrant path): agreed a dedicated unit test would be a nice follow-up, but the path does get real coverage today — the fiber/SRFI-18 suites under `-Dgc-stress=true` re-enter `markValue` via `markFiberState` on essentially every allocation, and they pass on this branch. Non-blocking.

One non-blocking suggestion inline about worklist capacity retention.

**Verdict:** the change is correct, follows the Zig 0.16 unmanaged-ArrayList conventions, and includes a regression test. Ready to merge once the pending CI matrix (test ×5, riscv64, benchmark-pr) goes green — format, wasm, and CodeRabbit already pass, and the branch has no conflicts with main. Since marking strictly does less allocator work now, benchmark-pr should be neutral-to-positive.

PR #1437, opened 2026-07-11
Deduplicate the lambda body scan between compiler_ir and compiler_lambda

### LINKED ISSUE #1432: Deduplicate the lambda body scan between compiler_ir and compiler_lambda
`compileLambdaWithIR` (`src/compiler_ir.zig`, ~lines 373–614) re-implements the lambda-body scan from `compileBodyForms` (`src/compiler_lambda.zig`) almost line for line: the globals-prescan sentinel dance, the `MAX_BODY_DEFS` letrec* desugaring scan, and the `define` / `define-syntax` / `define-record-type` handling.

PR #1427 is concrete evidence of the cost: the GC-rooting fix for body-local `define-syntax` transformers (append to `extra_roots` so the compiler-local macro map's transformers survive collections, #1401) had to be hand-applied **twice** — `src/compiler_ir.zig` ~line 437 and `src/compiler_lambda.zig` ~line 314. Any future GC-safety or letrec*-semantics fix in this area needs the same double patch, with an easy chance of missing one side and creating a divergence bug between the IR and legacy pipelines.

Suggested direction: extract the shared body-scan / letrec*-desugar logic into one helper both paths call, keeping each pipeline's expression-compilation callback separate. Epic #1038 (retire the legacy compileExpr path) is closed, and this duplication survived it — so it needs its own tracking.

Raised by CodeRabbit in the PR #1427 review (flagged Major / heavy-lift, out of scope for a GC-fix PR).

baijum:
## Summary

- Extracts the shared R7RS 5.3.2 body-scan logic (globals prescan, define name collection, define/define-syntax/define-record-type processing) into a `BodyScan` struct and `scanBodyDefs()` function in `compiler_lambda.zig`
- Both `compileLambdaWithIR` and `compileBodyForms` now call the shared helper, eliminating ~250 lines of near-identical code that previously required every GC-safety or letrec*-semantics fix to be hand-applied twice (e.g. PR #1427)
- Simplifies `compileLambdaWithIR`'s compilation phase by switching from an inline IR pipeline to `compileExprViaIR` + `compileExprSequence`, retaining only the function-name propagation specific to it

Closes #1432

## Test plan

- [x] `zig build` — clean build
- [x] `zig build test` — unit tests pass (local macOS + x86-64 Linux droplet)
- [x] `bash tests/scheme/run-all.sh` — 1827 pass, 0 fail
- [x] `zig build test -Dgc-stress=true` — EXIT:0 on x86-64 Linux droplet


src/compiler_lambda.zig
```diff
@@ -179,41 +207,13 @@ pub fn compileBodyForms(self: *Compiler, body: Value, opts: BodyOpts) CompileErr
...
-    // the rest of this scan and the phase-2 compileExpr calls allocate, so a
-    // collection would sweep the not-yet-compiled inits (issue #1010). Mirror
-    // them into extra_roots (by value, realloc-safe) for the duration.
-    const roots_base = self.gc.extra_roots.items.len;
-    defer self.gc.extra_roots.shrinkRetainingCapacity(roots_base);
-
-    var macro_count: usize = 0;
-    const macro_mark = if (opts.handle_define_syntax) self.beginBodyMacroScope() else 0;
-    defer if (opts.handle_define_syntax) self.endBodyMacroScope(macro_mark) catch {};
+    scan_result.roots_base = compiler.gc.extra_roots.items.len;
```
baijum: **Error-path cleanup regression.** On `main`, both copies established their cleanups as `defer`s incrementally, so *any* error exit (e.g. `TooManyLocals` from a body with >512 leading defines, OOM from `extra_roots.append`/`macros.put`, non-`InvalidSyntax` errors from `expandRecordTypeDefines`) still removed the VOID sentinels, shrank `extra_roots`, and ended the body-macro scope.

In this version all of that cleanup lives in `BodyScan.deinit()`, but the caller's `defer scan.deinit()` is only established *after* `scanBodyDefs` returns successfully. If `scanBodyDefs` errors mid-way, nothing cleans up:

- VOID sentinels stay in `vm.globals` **permanently** (globals outlive the failed eval) — later evals see phantom bindings that resolve to `#<void>` instead of raising unbound-variable errors.
- The `extra_roots` entries are never truncated — values pinned for the life of the GC.
- `beginBodyMacroScope` is never matched by `endBodyMacroScope` — `body_macro_depth` stays elevated and any `define-syntax` transformers already registered leak into the enclosing scope.

The `>512 defines → TooManyLocals` path is user-reachable Scheme, and a failed compile doesn't kill the REPL, so the leaked state is observable afterwards.

Suggested fix: capture `scan_result.roots_base` and call `beginBodyMacroScope` **before** the globals prescan (both are infallible and don't interact with it), then add `errdefer scan_result.deinit();` immediately after. Note the order matters: with the current placement, a naive `errdefer` that fires during the prescan would run `shrinkRetainingCapacity(0)` (the field default) and drop *outer* scopes' roots — a much worse GC bug. Hoisting the two captures above all fallible code avoids that. (Per repo policy a regression test would be good too, e.g. a body with 513 defines → expect the compile error, then verify one of the define names is still unbound.)

Separately: the old `// def_inits lives in a stack array the GC cannot see ... Mirror them into extra_roots (by value, realloc-safe) for the duration (issue #1010)` comment was dropped in the move. That comment documents a constraint the code can't show — please keep it on this line.

src/compiler_lambda.zig
```diff
@@ -310,53 +310,62 @@ pub fn compileBodyForms(self: *Compiler, body: Value, opts: BodyOpts) CompileErr
             if (ds_rest == types.NIL or !types.isPair(ds_rest)) break;
             const transformer_spec = types.car(ds_rest);
 
-            const transformer = macro.parseSyntaxRules(self, transformer_spec, all_def_names[0..all_def_count]) catch break;
-            // Root for the rest of the compile — see the define-syntax
-            // handling in compiler_ir.zig's body scan (#1401).
-            self.gc.extra_roots.append(self.gc.allocator, transformer) catch return CompileError.OutOfMemory;
+            const transformer = macro.parseSyntaxRules(compiler, transformer_spec, all_def_names[0..all_def_count]) catch break;
+            compiler.gc.extra_roots.append(compiler.gc.allocator, transformer) catch return CompileError.OutOfMemory;
```
baijum, line 325: Both copies of the `#1401` rooting rationale were deleted with the merge (this one said "Root for the rest of the compile...", the `compiler_ir.zig` copy had the full explanation: the transformer lives only in the compiler-local macro map, which the GC cannot see, and must survive collections triggered while compiling sibling body forms; released by the roots_base truncation after the body — the macro's entire scope — is compiled). Now that this is the *only* copy, it should carry the full comment — without it, this `extra_roots.append` looks redundant and is exactly the kind of line a future cleanup deletes.

src/compiler_lambda.zig
```diff
@@ -111,18 +111,46 @@ pub const BodyOpts = struct {
     handle_define_syntax: bool = false,
 };
 
-/// Shared implementation for lambda bodies and let-form bodies.
-/// In lambda mode (dst=null): allocates per-expression registers, last is
-/// always tail. In let-body mode (dst set): reuses caller's register.
-pub fn compileBodyForms(self: *Compiler, body: Value, opts: BodyOpts) CompileError!u16 {
-    const allocates_regs = opts.dst == null;
+pub const BodyScan = struct {
+    pub const MAX_DEFS = 512;
```
baijum, line 115: Hidden behavior change worth calling out in the PR description: `compileBodyForms` previously used `MAX_BODY_DEFS = 256`, so unifying on 512 means let-bodies with 257–512 leading defines now compile instead of failing with `TooManyLocals`. Benign (and arguably a fix), but it's a semantic change smuggled inside a "pure refactor".

Also note the stack cost: `BodyScan` is ~12.4 KB by value (512×16 for `def_names` + 512×8 for `def_inits`), plus 1 KB `def_slots` in the caller and a transient 8 KB `all_def_names` inside `scanBodyDefs`. Since body compilation recurses per lambda-nesting level, the `compileBodyForms` path now holds ~13.4 KB per level vs ~10.8 KB before. Fine for realistic programs, but worth knowing given fuzzing (epic #1397) likes deeply nested inputs.

src/compiler_forms.zig
```diff
@@ -26,6 +26,9 @@ pub const compileNamedLet = bindings.compileNamedLet;
 pub const compileDo = bindings.compileDo;
 pub const compileLetBody = bindings.compileLetBody;
 pub const compileBodyForms = lambda.compileBodyForms;
+pub const BodyScan = lambda.BodyScan;
+pub const scanBodyDefs = lambda.scanBodyDefs;
+pub const compileExprSequence = lambda.compileExprSequence;
```
baijum: These three re-exports are unused — `compiler_ir.zig` imports `compiler_lambda` directly (`compiler_lambda.scanBodyDefs`, `compiler_lambda.BodyScan.MAX_DEFS`, `compiler_lambda.compileExprSequence`). Either route `compiler_ir.zig` through the `forms` hub for consistency with the file's stated purpose, or drop these lines.

baijum (commented):
Reviewed the refactor against both original copies on `main`. The scan/compile logic itself is a faithful extraction — I verified line-by-line that:

- `scanBodyDefs` reproduces both original scans exactly (prescan sentinel handling, first-pass name collection, second-pass define/define-record-type/define-syntax processing, including the `catch break` vs `catch {}` error-handling asymmetries and the `handle_define_syntax` gating).
- `compileExprViaIR` is identical to the inline IR pipeline it replaces in `compileLambdaWithIR` (same `globals`/`restricted_env`/`compiler`/`set_targets` wiring, same `is_tail` flag passed to both `lowerAndOptimize` and `compileFromNode`).
- `compileExprSequence` with `allocates_regs=true` matches the deleted loops, including the `saved_next`/`freeReg()` dance and last-expression tail compilation.
- Defer ordering in `compileLambdaWithIR` is correct: `scan.deinit()` runs before `child.deinit()`, and after `emitClosureEpilogue`, matching the old function-exit defers.
- `BodyScan.deinit()` runs its cleanup steps in a different order than the old LIFO defers (`endBodyMacroScope` now last instead of first) — this is safe because no GC allocation can occur between the steps, so no collection can observe the intermediate state. Worth knowing it was checked.

One real regression found (error-path cleanup, inline comment), plus a hidden behavior change and two smaller notes — see inline comments.

**Merge readiness:** not quite yet. I'd fix the error-path cleanup regression (small change: hoist `roots_base`/`beginBodyMacroScope` above the prescan + `errdefer scan_result.deinit()`), restore the dropped #1010/#1401 GC-safety comments, and wait for the in-flight CI matrix to go green (format/wasm/CodeRabbit have passed; all test jobs were still IN_PROGRESS at review time). After that this is a clear improvement — it removes exactly the double-maintenance hazard that bit PR #1427.

baijum:
Re-reviewed commit 2ddc77fa against the earlier review — all substantive findings are addressed:

1. **Error-path cleanup regression — fixed correctly.** `roots_base` and `beginBodyMacroScope` are now captured before any fallible code, followed immediately by `errdefer scan_result.deinit()`. Verified the details: no shrink-to-0 hazard remains; the prescan block's `defer releaseGlobalsWrite` unwinds before the function-level errdefer, so `deinit()`'s re-acquisition of the globals lock can't deadlock; the success path is unchanged and no double-deinit is possible; hoisting `beginBodyMacroScope` above the prescan is behavior-neutral (the prescan only touches the globals map).
2. **#1010 and #1401 GC-safety comments — restored**, correctly adapted to the shared-helper context.
3. **Unused `compiler_forms.zig` re-exports — removed.**

Notes on the extras in this push:

- The new `load_void` branch in `compileLambdaWithIR` (define-syntax-only bodies) fixes a pre-existing quirk where such a lambda returned whatever was in register 0 (typically the first parameter). Good catch, two small caveats: (a) no regression test, and (b) the legacy `compileBodyForms` lambda path (`compileLambdaBody` → `dst=null`) still has the old behavior, so the two paths now diverge on that corner case. Both fine to leave for the legacy-path retirement (#1038), just noting for the record.
- The PR description still doesn't mention the `MAX_BODY_DEFS` 256→512 change for let-bodies — worth a one-liner, non-blocking.
- Commit e7608983 (do-stress-test droplet sizing) is unrelated to this PR's subject; harmless, but noting the scope creep.

**Verdict: ready to merge once the re-running CI matrix is green.** The previously reviewed head (67177ba) already passed the full matrix (all test jobs, riscv64, wasm, format, CodeRabbit, benchmark-pr), and this delta is 26 lines of the exact fixes requested. Auto-merge is not armed, so it needs a manual merge after CI completes.

Issue #1440, opened 2026-07-11
Scheduler integration: io_waiting state, runSchedulerStep, reactor park (KEP-0001 P2)


baijum:
Part of #1438 (KEP-0001, Phase 2). Depends on Phase 1 (#1439). Design: [KEP-0001 §3](https://github.com/kaappi/keps/blob/main/keps/0001-event-loop-reactor.md#3-scheduler-integration).

## Scope

- Add `io_waiting` to `FiberStatus` ([`fiber.zig:18`](https://github.com/kaappi/kaappi/blob/488eaed2/src/fiber.zig#L18)) and `io_fd`/`io_interest`/`io_buffer` fields to `Fiber`. `schedule()` treats `io_waiting` as non-runnable; `hasRunnableFibers()` ([`fiber.zig:264`](https://github.com/kaappi/kaappi/blob/488eaed2/src/fiber.zig#L264)) gains an `io_waiting` branch ("alive, waiting" vs "done").
- Collapse the four structurally identical scheduler loops into one `runSchedulerStep` helper: [`primitives_fiber.zig:177`](https://github.com/kaappi/kaappi/blob/488eaed2/src/primitives_fiber.zig#L177) (channels/`fiber-join`), [`primitives_srfi18.zig:597`](https://github.com/kaappi/kaappi/blob/488eaed2/src/primitives_srfi18.zig#L597) (`thread-join!`), [`:773`](https://github.com/kaappi/kaappi/blob/488eaed2/src/primitives_srfi18.zig#L773) (mutex), [`:861`](https://github.com/kaappi/kaappi/blob/488eaed2/src/primitives_srfi18.zig#L861) (condvar). The reactor park replaces both the bare channel-loop `break` and `scheduleOrTimeout`'s whole-thread `nanosleep` ([`:577`](https://github.com/kaappi/kaappi/blob/488eaed2/src/primitives_srfi18.zig#L577)).
- Fold the `deadline_ns` sweep in `schedule()` ([`fiber.zig:240-253`](https://github.com/kaappi/kaappi/blob/488eaed2/src/fiber.zig#L240)) into the reactor timer heap; reimplement `thread-sleep!` ([`primitives_srfi18.zig:346`](https://github.com/kaappi/kaappi/blob/488eaed2/src/primitives_srfi18.zig#L346)) as a timed park instead of a whole-thread `nanosleep`.
[...]

PR #1446, opened 2026-07-11
Reactor core: kqueue/epoll backends with a userspace timer heap (KEP-0001 P1)

### LINKED ISSUE #1439: Reactor core: src/reactor.zig with kqueue/epoll backends and timer heap (KEP-0001 P1)
Part of #1438 (KEP-0001, Phase 1). No behavior change in this phase — the reactor is added and unit-tested in isolation.

## Scope

- New `src/reactor.zig`: one `Reactor` per OS thread — `init`/`deinit`, `register(fd, interest, fiber)`, `unregister(fd)`, `addTimer(deadline_ns, fiber)`, `poll(timeout_ns, ready)`, `isEmpty()`. Full interface in [KEP-0001 §1](https://github.com/kaappi/keps/blob/main/keps/0001-event-loop-reactor.md#1-the-reactor-abstraction-new-srcreactorzig).
- Backend selection via the existing comptime `os.tag` idiom: kqueue (macOS/BSD) and epoll (Linux) in this phase; WASI is Phase 4.
- **kqueue:** `std.c.kqueue()`/`std.c.kevent()` (`std.posix` only re-exports the `Kevent` type), `EV.ADD | EV.ONESHOT`, `udata = fd`.
- **epoll:** `std.os.linux.epoll_create1/ctl/wait`, `EPOLL.IN/OUT | EPOLL.ONESHOT`, `data.fd = fd`. Errno via `std.os.linux.errno(usize)` for raw syscall returns; `std.posix.errno` on the libc path (both verified against Zig 0.16.0).
- Userspace **timer min-heap** shared by both backends (libuv model): heap root bounds each backend's wait timeout. **Decided (resolved Q2):** `epoll_wait` ms granularity with **ceil-rounding** — a timer may fire ≤1 ms late, never early — and expiry is decided by comparing `clockNs()` against the heap, never by assuming the wait ran full length.
- Registrations are fd-keyed (`AutoHashMap(i32, Reg)`); the kernel is handed the **fd**, never a fiber pointer. **Decided (resolved Q1):** `Reg` holds a waiter **list** per direction (usually length 1); readiness wakes **all** waiters on that direction — the `wakeChannelWaiters` discipline; losers re-park via retry. Permits multiple acceptor fibers on one listen fd.
- **Decided (resolved Q3):** ONESHOT delivery, both backends — a registration exists only while a fiber is parked, by construction. Edge-triggered remains deferred to Phase 7.

## Acceptance

- Unit tests cover: register → readiness wakeup, multiple waiters on one fd/direction all woken, timer expiry ordering (including never-early under ms rounding), mixed fd+timer timeout selection, unregister-before-close, `isEmpty`.
- `zig build test` green on macOS and Linux; no caller yet.

baijum:
Closes #1439 (KEP-0001, Phase 1).

## Summary

- `src/reactor.zig`: one `Reactor` per OS thread — `register`/`unregister`/`addTimer`/`poll`/`isEmpty`, backed by kqueue (macOS/Apple platforms) or epoll (Linux) and a shared userspace timer min-heap so deadline logic is identical across both.
- No behavior change: nothing calls into it yet (Phase 2, #1440, wires it into the scheduler).
- `src/tests_reactor.zig`: 12 isolation tests against real pipe/socketpair fds and bare-stack fake `*Fiber` values (reactor.zig never dereferences a parked fiber — only stores/returns the pointer).

## Design decisions exercised

All five questions [resolved in the KEP](https://github.com/kaappi/keps/blob/main/keps/0001-event-loop-reactor.md#unresolved-questions) are implemented, not just documented:

- **Wake-all waiters** (`Reg.read_waiters`/`write_waiters` are lists, not single pointers) — covered by "multiple waiters on one fd direction are all woken".
- **epoll ms timeouts, ceil-rounded** (`msFromNs`, never fires early) — covered by "the nearer of an fd timeout and a timer deadline bounds the wait".
- **ONESHOT on both backends.**
- **fd-keyed registration**, no token/generation indirection — `unregister` is void-returning exactly per the KEP's interface.

## The trickiest bit

kqueue and epoll disagree on what "oneshot" disarms: kqueue's read/write are independent knotes (firing one never touches the other), but epoll's `EPOLLONESHOT` disarms the **whole fd registration** on any fire — including a direction that didn't fire. Missing this would silently strand a still-parked waiter with no future wakeup. `Reactor.poll` re-arms whatever's left after a partial fire (a harmless redundant `EV_ADD` on kqueue); the test "one fd with both read and write interest: a fired direction re-arms the other" exercises this directly over a Unix socketpair — the case most likely to regress silently if anyone (including future-me) "simplifies" this later.

## Beyond the KEP's interface sketch

Added `Reactor.removeTimer(fiber)`, not in the KEP's illustrative interface snippet. Phase 2 needs it: when a timed wait resolves through its non-timeout path (e.g. a mutex unlock wakes a fiber that was also timed-waiting on the lock), the stale heap entry must be cancelled — otherwise it could later fire against a reused fiber slot and wake the wrong logical fiber. Covered by "removeTimer cancels a pending timer so it never fires".

## Testing

- `zig build test` — 745/745 passing (12 new).
- `zig build test -Dgc-stress=true` — 739/745 passing, 6 skipped (pre-existing skip conditions unrelated to this change), 0 failures.
- `zig fmt --check` clean.
- Compiles natively (macOS/kqueue) and cross-compiles for Linux x86_64/aarch64/riscv64 (epoll) — actual execution on non-native Linux targets needs the podman/QEMU setup CI uses, not available in this environment, but compilation is verified for all three.

### DIFF
--- a/src/reactor.zig
+++ b/src/reactor.zig
@@ -46,7 +46,6 @@ const TimerHeap = std.PriorityQueue(TimerEntry, void, timerLessThan);
 /// question 1) — the same discipline `FiberScheduler.wakeChannelWaiters`
 /// uses for channels. Losers of the resulting retry race simply re-park.
 const Reg = struct {
-    fd: i32,
     read_waiters: std.ArrayList(*Fiber) = .empty,
     write_waiters: std.ArrayList(*Fiber) = .empty,
     /// Whether this fd has ever been armed with the backend. epoll must
@@ -92,7 +91,7 @@ pub const Reactor = struct {
     /// registration exists only while a fiber is parked, by construction).
     pub fn register(self: *Reactor, fd: i32, interest: Interest, fiber: *Fiber) !void {
         const gop = try self.regs.getOrPut(fd);
-        if (!gop.found_existing) gop.value_ptr.* = .{ .fd = fd };
+        if (!gop.found_existing) gop.value_ptr.* = .{};
         const reg = gop.value_ptr;
 
         switch (interest) {
@@ -163,6 +162,13 @@ pub const Reactor = struct {
     /// is sooner; forever if both are null) and appends every fiber made
     /// runnable — by fd readiness or timer expiry — to `ready`. `ready`
     /// must use the same allocator this Reactor was `init`ed with.
+    ///
+    /// `ready` may contain the same fiber twice in one call: a fiber parked
+    /// with both an fd registration and a timer (a timed wait) is appended
+    /// once if the fd wins and again if the timer also expires in this same
+    /// call, since nothing removes the timer entry when the fd path wins.
+    /// Callers must tolerate a duplicate wake (e.g. an idempotent status
+    /// flip on the second occurrence).
     pub fn poll(self: *Reactor, timeout_ns: ?u64, ready: *std.ArrayList(*Fiber)) !void {
         const wait_ns = self.effectiveTimeout(timeout_ns);
         const events = try self.backend.wait(wait_ns);
@@ -171,13 +177,20 @@ pub const Reactor = struct {
             const reg = self.regs.getPtr(ev.fd) orelse continue; // stale event; already unregistered
 
             var fired = false;
+            // Reserved up front so the drain below can't fail mid-way: a
+            // failure after the kernel's ONESHOT event was consumed but
+            // before all waiters were moved to `ready` would strand the
+            // remaining waiters forever (nothing re-arms the fd for them),
+            // and `isEmpty()` would still report a waiter, so no deadlock
+            // detector would catch it either.
+            try ready.ensureUnusedCapacity(self.allocator, reg.read_waiters.items.len + reg.write_waiters.items.len);
             if (ev.readable and reg.read_waiters.items.len > 0) {
-                for (reg.read_waiters.items) |f| try ready.append(self.allocator, f);
+                for (reg.read_waiters.items) |f| ready.appendAssumeCapacity(f);
                 reg.read_waiters.clearRetainingCapacity();
                 fired = true;
             }
             if (ev.writable and reg.write_waiters.items.len > 0) {
-                for (reg.write_waiters.items) |f| try ready.append(self.allocator, f);
+                for (reg.write_waiters.items) |f| ready.appendAssumeCapacity(f);
                 reg.write_waiters.clearRetainingCapacity();
                 fired = true;
             }
@@ -269,13 +282,17 @@ const KqueueBackend = struct {
     }
 
     fn disarmAll(self: *KqueueBackend, fd: i32) void {
-        var changes = [2]std.c.Kevent{
-            mkChange(fd, .read, false),
-            mkChange(fd, .write, false),
-        };
-        // Best-effort: ENOENT is expected whenever a direction was never
-        // armed (e.g. a write-only port has no read filter to delete).
-        self.apply(&changes) catch {};
+        // Two independent calls, not one batched changelist: with a
+        // zero-length eventlist, kevent() has nowhere to report a
+        // per-change error, so it aborts the whole changelist at the first
+        // failure. ENOENT is expected whenever a direction was never armed
+        // (e.g. a write-only port has no read filter to delete) — batching
+        // would let that expected ENOENT on one filter silently leave the
+        // other filter's knote behind.
+        var read_change = mkChange(fd, .read, false);
+        self.apply((&read_change)[0..1]) catch {};
+        var write_change = mkChange(fd, .write, false);
+        self.apply((&write_change)[0..1]) catch {};
     }
 
     fn apply(self: *KqueueBackend, changes: []const std.c.Kevent) !void {
@@ -327,7 +344,11 @@ const EpollBackend = struct {
     ready: [max_events_per_poll]ReadyEvent = undefined,
 
     fn init() !EpollBackend {
-        const rc = linux.epoll_create1(0);
+        // CLOEXEC: without it the epoll fd leaks into every child the core
+        // spawns (thottam_proc.zig, native_compiler.zig via
+        // std.process.Child). kqueue needs no equivalent — kqueue(2) fds
+        // are never inherited across fork by design.
+        const rc = linux.epoll_create1(linux.EPOLL.CLOEXEC);
         if (linux.errno(rc) != .SUCCESS) return error.Unexpected;
         return .{ .epfd = @intCast(rc) };
     }
@@ -387,6 +408,6 @@ const EpollBackend = struct {
 fn msFromNs(timeout_ns: ?u64) i32 {
     const ns = timeout_ns orelse return -1;
     if (ns == 0) return 0;
-    const ms = (ns + 999_999) / 1_000_000;
+    const ms = (ns +| 999_999) / 1_000_000;
     return if (ms > std.math.maxInt(i32)) std.math.maxInt(i32) else @intCast(ms);
 }
--- a/src/reactor.zig
+++ b/src/reactor.zig
@@ -1,0 +1,392 @@
+//! Per-OS-thread I/O readiness multiplexer (KEP-0001 Phase 1).
+//!
+//! One `Reactor` belongs to one OS thread's scheduler. A fiber that would
+//! block on a fd registers its interest and parks; `poll` blocks once,
+//! bounded by the nearest timer deadline, and reports every fiber that
+//! became runnable (fd readiness or timer expiry). This file has no
+//! scheduler caller yet — that wiring is KEP-0001 Phase 2. See
+//! https://github.com/kaappi/keps/blob/main/keps/0001-event-loop-reactor.md
+const std = @import("std");
+const builtin = @import("builtin");
+const fiber_mod = @import("fiber.zig");
+const Fiber = fiber_mod.Fiber;
+
+const linux = std.os.linux;
+
+/// Events buffered per backend `wait()` call. A burst larger than this
+/// drains over multiple `poll()` calls, which the scheduler loop already
+/// performs naturally.
+const max_events_per_poll: usize = 256;
+
+pub const Interest = enum { read, write };
+
+/// A backend-normalized readiness result: which directions fired for `fd`.
+const ReadyEvent = struct { fd: i32, readable: bool, writable: bool };
+
+const Backend = switch (builtin.os.tag) {
+    .macos, .ios, .tvos, .watchos, .visionos => KqueueBackend,
+    .linux => EpollBackend,
+    // WASI (poll_oneoff) arrives in KEP-0001 Phase 4.
+    else => @compileError("reactor: unsupported OS for KEP-0001 Phase 1 (kqueue/epoll only)"),
+};
+
+const TimerEntry = struct {
+    deadline_ns: u64,
+    fiber: *Fiber,
+};
+
+fn timerLessThan(_: void, a: TimerEntry, b: TimerEntry) std.math.Order {
+    return std.math.order(a.deadline_ns, b.deadline_ns);
+}
+
+const TimerHeap = std.PriorityQueue(TimerEntry, void, timerLessThan);
+
+/// Per-fd bookkeeping. Waiter lists are usually length 1; multiple waiters
+/// on one direction are woken all at once on readiness (resolved KEP-0001
+/// question 1) — the same discipline `FiberScheduler.wakeChannelWaiters`
+/// uses for channels. Losers of the resulting retry race simply re-park.
+const Reg = struct {
+    fd: i32,
+    read_waiters: std.ArrayList(*Fiber) = .empty,
+    write_waiters: std.ArrayList(*Fiber) = .empty,
+    /// Whether this fd has ever been armed with the backend. epoll must
+    /// distinguish first-arm (EPOLL_CTL_ADD) from re-arm (EPOLL_CTL_MOD) —
+    /// re-adding an already-tracked fd fails EEXIST. kqueue ignores this
+    /// (EV_ADD is idempotent whether creating fresh or recreating a knote
+    /// that a prior EV_ONESHOT deleted).
+    kernel_registered: bool = false,
+
+    fn isEmpty(self: Reg) bool {
+        return self.read_waiters.items.len == 0 and self.write_waiters.items.len == 0;
+    }
+};
+
+pub const Reactor = struct {
+    allocator: std.mem.Allocator,
+    backend: Backend,
+    regs: std.AutoHashMap(i32, Reg),
+    timers: TimerHeap,
+
+    pub fn init(allocator: std.mem.Allocator) !Reactor {
+        return .{
+            .allocator = allocator,
+            .backend = try Backend.init(),
+            .regs = std.AutoHashMap(i32, Reg).init(allocator),
+            .timers = .empty,
+        };
+    }
+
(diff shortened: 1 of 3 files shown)


src/reactor.zig
```diff
@@ -0,0 +1,392 @@
...
+    }
+
+    fn disarmAll(self: *KqueueBackend, fd: i32) void {
+        var changes = [2]std.c.Kevent{
+            mkChange(fd, .read, false),
+            mkChange(fd, .write, false),
+        };
+        // Best-effort: ENOENT is expected whenever a direction was never
+        // armed (e.g. a write-only port has no read filter to delete).
+        self.apply(&changes) catch {};
```
baijum: **Bug (verified on macOS): the first `ENOENT` aborts the whole changelist, so the second `EV_DELETE` is never processed.**

With a zero-length eventlist, `kevent()` has nowhere to report a per-change error, so it returns `-1` at the first failing change and drops the rest of the changelist. The comment's premise — "ENOENT is expected whenever a direction was never armed" — is exactly the situation that triggers the abort: for a **write-only** registration, `changes[0]` (read `EV_DELETE`) fails `ENOENT` and `changes[1]` (write `EV_DELETE`) never runs, so the write knote stays armed. Same story when unregistering a both-direction registration after the read side already fired (its knote is gone, so its delete `ENOENT`s first).

Verified with a standalone C program on macOS 26.5 (arm64) — write-only ONESHOT registration on a socketpair, then this exact disarm pattern:

```
arm write:    rc=0
disarm both:  rc=-1 errno=No such file or directory
poll:         rc=1 filter=-2  -> WRITE knote SURVIVED (delete was aborted)
with evlist:  rc=1 [EV_ERROR data=ENOENT]
poll again:   rc=0            -> no event (both deletes applied)
```

Today the fallout is masked — the stale event hits `poll()`'s `orelse continue` guard, and ONESHOT means it fires at most once — but it leaves kernel state behind after `unregister()`, which breaks this file's own "registration exists only while a fiber is parked" invariant, and it's invisible to the Phase 3 debug assertions because the leak is kernel-side only.

Two small fixes (the second is what the `with evlist` line above verifies):
- issue the two `EV_DELETE`s as two independent `kevent()` calls, or
- pass an eventlist buffer (e.g. `&self.raw`) so per-change failures come back as `EV_ERROR` events instead of aborting — every change is then always processed.

src/reactor.zig
```diff
@@ -0,0 +1,392 @@
...
+// epoll backend (Linux)
+// ---------------------------------------------------------------------------
+
+const EpollBackend = struct {
+    epfd: i32,
+    raw: [max_events_per_poll]linux.epoll_event = undefined,
+    ready: [max_events_per_poll]ReadyEvent = undefined,
+
+    fn init() !EpollBackend {
+        const rc = linux.epoll_create1(0);
```
baijum: `epoll_create1(0)` leaves the epoll fd inheritable across `fork`/`exec`. The core does spawn subprocesses (`thottam_proc.zig`, `native_compiler.zig` via `std.process.Child`), so a long-lived scheduler's reactor fd would leak into every child. kqueue is immune by design ("The queue is not inherited by a child created with fork(2)" — kqueue(2)), which makes epoll the odd one out here.

```zig
const rc = linux.epoll_create1(linux.EPOLL.CLOEXEC);
```

(`EPOLL.CLOEXEC` is present in Zig 0.16's `std.os.linux`.)

src/reactor.zig
```diff
@@ -0,0 +1,392 @@
...
+            const remaining_write = reg.write_waiters.items.len > 0;
+            if (remaining_read or remaining_write) {
+                try self.backend.arm(ev.fd, remaining_read, remaining_write, false);
+            }
+        }
+
+        const now = fiber_mod.clockNs();
+        while (self.timers.peek()) |top| {
+            if (top.deadline_ns > now) break;
+            try ready.append(self.allocator, self.timers.pop().?.fiber);
```
baijum, line 214: Worth a contract line in `poll()`'s doc comment: a fiber parked with a timeout (fd registration **plus** `addTimer`) can be appended to `ready` **twice in one call** — once by the fd path above, and again here, since nothing removes its timer entry when the fd path wins. Phase 2's removeTimer-on-non-timeout-wake necessarily happens only *after* `poll()` returns, so it can't prevent this case. The scheduler's status flip is presumably idempotent (the second wake sees `suspended`, not `io_waiting`), but "`ready` may contain duplicates" should be written down where Phase 2 will read it.

src/reactor.zig
```diff
@@ -0,0 +1,392 @@
...
+    pub fn poll(self: *Reactor, timeout_ns: ?u64, ready: *std.ArrayList(*Fiber)) !void {
+        const wait_ns = self.effectiveTimeout(timeout_ns);
+        const events = try self.backend.wait(wait_ns);
+
+        for (events) |ev| {
+            const reg = self.regs.getPtr(ev.fd) orelse continue; // stale event; already unregistered
+
+            var fired = false;
+            if (ev.readable and reg.read_waiters.items.len > 0) {
+                for (reg.read_waiters.items) |f| try ready.append(self.allocator, f);
```
baijum: Hardening nit: if an `append` fails mid-drain (OOM), the error propagates *after* the ONESHOT kernel event has been consumed but with the waiters still sitting in the list — nothing will ever re-arm the fd, so those fibers hang forever, and `isEmpty()` stays `false`, so a deadlock detector won't flag it either. Since this is the scheduler's critical wake path, consider reserving up front and making the drain infallible:

```zig
try ready.ensureUnusedCapacity(self.allocator, reg.read_waiters.items.len + reg.write_waiters.items.len);
// then appendAssumeCapacity(...) in both drain loops
```

src/reactor.zig
```diff
@@ -0,0 +1,392 @@
...
+        return self.ready[0..n];
+    }
+};
+
+/// epoll_wait's timeout is i32 milliseconds. Rounds up (ceil) so a timer
+/// may fire slightly late but never early (resolved KEP-0001 question 2).
+fn msFromNs(timeout_ns: ?u64) i32 {
+    const ns = timeout_ns orelse return -1;
+    if (ns == 0) return 0;
+    const ms = (ns + 999_999) / 1_000_000;
```
baijum: Nit: `ns + 999_999` overflows (safety panic) for `ns` within 10^6 of `maxInt(u64)`. Unreachable with real deadlines (~584 years of monotonic time), but a caller passing a "practically infinite" sentinel instead of `null` would hit it, and the clamp below already handles the huge-result case. `ns +| 999_999` is free insurance.

src/reactor.zig
```diff
@@ -0,0 +1,392 @@
...
+}
+
+const TimerHeap = std.PriorityQueue(TimerEntry, void, timerLessThan);
+
+/// Per-fd bookkeeping. Waiter lists are usually length 1; multiple waiters
+/// on one direction are woken all at once on readiness (resolved KEP-0001
+/// question 1) — the same discipline `FiberScheduler.wakeChannelWaiters`
+/// uses for channels. Losers of the resulting retry race simply re-park.
+const Reg = struct {
+    fd: i32,
```
baijum: Nit: `fd` is written at creation but never read — the hashmap key is the only fd source everywhere else (`ev.fd` in `poll`, the parameter in `unregister`). Drop it, or note that it's kept for debugger visibility.

baijum (commented):
## Review — Reactor core (KEP-0001 Phase 1)

**Overview.** Adds `src/reactor.zig` — a per-OS-thread readiness multiplexer with kqueue (Apple) and epoll (Linux) ONESHOT backends and a shared userspace timer min-heap — plus 12 isolation tests over real pipes/socketpairs, with no callers yet (Phase 2). Cross-checked against KEP-0001 as merged in `kaappi/keps`: the public interface matches the KEP sketch exactly (`register`/`unregister`/`addTimer`/`poll`/`isEmpty`), all five resolved questions are genuinely implemented rather than just cited (wake-all waiter lists, ms ceil-rounding, ONESHOT on both backends, fd-keyed registration), and the `removeTimer` addition beyond the sketch is well-argued.

**Strengths**
- The epoll "`EPOLLONESHOT` disarms the *whole* fd registration" subtlety is correctly identified, handled by the re-arm step in `poll()`, and pinned by exactly the right regression test (socketpair, both directions, partial fire).
- The epoll ADD-vs-MOD state machine (`kernel_registered`) is right, including the fired-but-still-tracked dormant case.
- Comments record *why* at each decision point and test names state the invariant they guard — both unusually good.
- `isEmpty()`'s "empty waiter lists ≠ pending work" distinction is thought through and tested.

**Issues found** (details in the inline comments)
1. **Verified bug — `KqueueBackend.disarmAll`:** batching both `EV_DELETE`s into one `kevent()` call with a zero-length eventlist means the first `ENOENT` aborts the changelist, so the other direction's knote survives `unregister()`. Reproduced on macOS with a standalone C program (output inline). Fallout is currently masked by the stale-event guard + ONESHOT, but it leaks kernel state and breaks the file's own "registration exists only while a fiber is parked" invariant. Cheap fix. *Should fix before merge.*
2. **epoll fd missing `CLOEXEC`** — the fd leaks into subprocesses (`thottam_proc.zig` / `native_compiler.zig` spawn children). One-word fix. *Should fix before merge.*
3. `poll()` can deliver the same fiber twice when an fd wake and its timer expiry land in one call — needs a doc-comment contract line for Phase 2, not a code change.
4. OOM mid-drain can strand parked waiters after their ONESHOT event was already consumed — reserving capacity up front makes the wake path infallible. Hardening, non-blocking.
5. Two nits: `msFromNs` addition can overflow on absurd input; `Reg.fd` is dead.

**Phase-2 heads-up** (not this PR's problem, but worth deciding now): `zig build wasm` passes today only because nothing imports `reactor.zig`, so Zig's lazy analysis never reaches the `else => @compileError(...)`. The moment Phase 2 imports it from `fiber.zig`/primitives, the wasm build breaks — and Phase 4 (the WASI backend) is scheduled *after* Phase 2. Phase 2 will need either a stub backend for `.wasi` or `is_wasm`-gated imports.

**Tests.** Solid: real fds, no mocks, each test pins a named invariant, and the hardest case (partial fire with both-direction interest) is covered. Two suggested additions: (a) a **peer-close test** — close the write end and expect the parked read waiter to wake; the `EV_EOF` / `EPOLLHUP|EPOLLERR` → `broken` mapping is real logic on both backends and is currently untested; (b) unregister of a write-only and of a both-direction registration, which exercises the `disarmAll` path from issue 1 once fixed.

**Security / performance.** Nothing concerning beyond the `CLOEXEC` point. `removeTimer`/`isEmpty` linear scans are fine at Phase-1 scale, and Phase 7 owns profiling.

**Verification done for this review:** full diff read; KEP-0001 cross-checked; `std.PriorityQueue` `push`/`pop`/`popIndex` and `EPOLL.CLOEXEC` confirmed against Zig 0.16.0's std; the `disarmAll` changelist-abort reproduced with a standalone C program on macOS 26.5 (arm64). CI was still pending at review time (format + wasm green).

**Verdict: not ready to merge quite yet — but very close.** Fix issues 1 and 2 (both small, both in this file) and this is good to go; items 3–5 can ride along now or land with Phase 2. Architecture, KEP fidelity, and test discipline are all in excellent shape.

baijum:
Pushed 71ab3d8 addressing the review.

**Fixed (both flagged should-fix-before-merge):**
- `KqueueBackend.disarmAll` now issues the read/write `EV_DELETE`s as two independent `kevent()` calls instead of one batched changelist — with a zero-length eventlist the first `ENOENT` was aborting the rest of the changelist and leaving a knote behind. Matches the standalone-C repro in the review exactly.
- `EpollBackend.init` now passes `EPOLL.CLOEXEC` so the reactor fd doesn't leak into subprocesses spawned via `std.process.Child`.

**Also addressed:**
- Dropped the dead `Reg.fd` field.
- `poll()`'s fd-drain now reserves `ready`'s capacity up front (`ensureUnusedCapacity` + `appendAssumeCapacity`) so a mid-drain OOM can't strand waiters after their ONESHOT event was already consumed.
- `msFromNs` uses a saturating add to avoid a theoretical overflow panic on a near-`u64::MAX` timeout.
- Documented in `poll()`'s doc comment that `ready` may contain the same fiber twice when an fd wake and its timer both land in one call, for Phase 2.
- `tests_reactor.zig`: single-byte writes now go through a `writeByte` helper that asserts the byte landed, instead of discarding the return value.
- Added a peer-close test covering the EOF/HUP → `broken` mapping on both backends (kqueue `EV_EOF`, epoll `EPOLLHUP|EPOLLERR`), which was previously untested.

**Not changed, with reasoning:**
- The suggested unregister test for a write-only/both-direction registration (to exercise the fixed `disarmAll` path) — the original bug was a kernel-side knote leak masked by `poll()`'s stale-event guard, so it isn't observable through the `Reactor` public API even without the fix. Your own repro needed raw `kevent()` access to see it. A black-box test here would pass regardless of the fix and give false confidence, so I left it out.
- Item 3 (possible duplicate-wake) and item 4 (OOM hardening) were both addressed above; item 3 as a doc contract rather than a behavior change, since preventing it requires Phase 2's scheduler-side dedup, not a Phase 1 change.
- The wasm-build heads-up (Phase 2 needs a `.wasi` stub or `is_wasm`-gated imports) is noted for Phase 2, not this PR.

**Re-verified:** `zig fmt --check` clean, native macOS build-obj + Linux x86_64 cross-compile both exit 0, full suite 746/746 (up from 745 with the new test), gc-stress 740/746 (6 pre-existing skips, 0 failures).

baijum:
Re-reviewed at 71ab3d80 ("Address PR #1446 review feedback"). **All six findings from my review are addressed:**

1. ✅ `disarmAll` now issues the two `EV_DELETE`s as independent `kevent()` calls, so an expected `ENOENT` on one direction can no longer abort the other's delete — with a comment capturing the why. (My earlier C repro's control case confirms per-call processing applies each delete.)
2. ✅ `epoll_create1(EPOLL.CLOEXEC)`, with the subprocess-leak rationale in a comment.
3. ✅ `poll()`'s doc comment now states the duplicate-wake contract (fd + timer in one call) for Phase 2.
4. ✅ The drain reserves via `ensureUnusedCapacity` and uses `appendAssumeCapacity` — per-fd waiter delivery is now all-or-nothing.
5. ✅ `msFromNs` uses saturating `+|`.
6. ✅ Dead `Reg.fd` removed.

Also picked up the suggested **peer-close EOF/HUP test** (the `broken` mapping is now exercised on both backends) and hardened the test writes with `writeByte`. The other suggested test (unregister of write-only / both-direction registrations) wasn't added — acceptable: the two-call disarm makes the original failure structurally impossible.

One optional residual, explicitly non-blocking: the per-event `ensureUnusedCapacity` makes each fd's delivery atomic, but an OOM at event *N* still strands the waiters of events ≥ *N* (their ONESHOT fired, nothing re-arms). Closing it fully would mean pre-summing waiter counts across all returned events and reserving once before the drain. Fine to leave until Phase 2 decides `poll()`-error policy — a comment or tracking note would do.

**Verification for this re-review:** fix diff read in full; `zig build test` at 71ab3d80 on macOS arm64 — 746/746 pass (kqueue paths incl. the new EOF test); unit tests cross-compile clean for x86_64-linux (epoll + new test code type-check with libc's `isize` write).

**Verdict: ready to merge once the CI matrix finishes green** (format, wasm, and CodeRabbit already pass; the test jobs were still running as I write this).

Issue #1441, opened 2026-07-11
Port layer: non-blocking reads/writes, write buffering, (read) EAGAIN path (KEP-0001 P3)


baijum:
Part of #1438 (KEP-0001, Phase 3). Depends on Phase 2 (#1440). Design: [KEP-0001 §4](https://github.com/kaappi/keps/blob/main/keps/0001-event-loop-reactor.md#4-io-primitive-changes).

## Scope

- **Lazy `O_NONBLOCK`** on ports at first reactor registration — never fd 0 in REPL mode (linenoise reads fd 0 directly in C).
- **Reads:** hook the single syscall in `readOneByte` ([`primitives_io.zig:325`](https://github.com/kaappi/kaappi/blob/488eaed2/src/primitives_io.zig#L325), syscall at L364) — after the three software buffers (`peek_byte`/`peek_extra`/`read_buf`) and string-port data are exhausted. On `EAGAIN`: set `io_waiting`, register with the reactor, set `vm.yield_retry = true`, return `PrimitiveError.Yielded` (same park-and-retry protocol as `blockOrDeadlock`, [`primitives_fiber.zig:233`](https://github.com/kaappi/kaappi/blob/488eaed2/src/primitives_fiber.zig#L233)). Covers `read-char`, `peek-char`, `read-line`, `read-u8`, `peek-u8`, `read-string`, `read-bytevector` through the one choke point.
- **Writes:** the port path is `writeToPort` → `writeToFd` in [`reporting.zig:8`](https://github.com/kaappi/kaappi/blob/488eaed2/src/reporting.zig#L8), shared with REPL diagnostics and lacking an `EAGAIN` path. Split a port-layer fd writer that tracks progress in the `Port` so a blocked mid-buffer write suspends on `write` interest and resumes with the remaining slice; `reporting.zig` keeps the blocking version for fd 1/2.
- **Per-port write buffer** flushed when full and by a real `flush-output-port` (currently a no-op, [`primitives_io.zig:792`](https://github.com/kaappi/kaappi/blob/488eaed2/src/primitives_io.zig#L792)) — effectively a prerequisite: without it every `write-char` is a syscall and a reactor registration.
[...]

Issue #1438, opened 2026-07-11
Epic: event-loop reactor for fiber I/O (KEP-0001)


baijum:
Tracking epic for implementing [KEP-0001: Event-Loop Reactor for Fiber I/O](https://github.com/kaappi/keps/blob/main/keps/0001-event-loop-reactor.md) (status: Accepted).

## Problem

Fibers interleave CPU work at `yield` points but do **not** overlap I/O waits: a fiber that reads a socket issues a blocking `read(2)` that parks the whole OS thread and every fiber on it ([`primitives_io.zig:325`](https://github.com/kaappi/kaappi/blob/488eaed2/src/primitives_io.zig#L325), syscall at L364). `thread-sleep!` has the same shape for timers — a whole-thread `nanosleep(2)` ([`primitives_srfi18.zig:346`](https://github.com/kaappi/kaappi/blob/488eaed2/src/primitives_srfi18.zig#L346)).

## Design

A per-OS-thread **reactor** over kqueue (macOS/BSD), epoll (Linux), and `poll_oneoff` (WASI): a fiber that would block registers its fd and suspends (`io_waiting`); when nothing is runnable the scheduler blocks once in the reactor, bounded by the nearest timer deadline across *all* fibers, then wakes the ready fibers. Blocking-looking Scheme code (`(read-line port)`) transparently yields; one OS thread multiplexes thousands of connections on one GC heap. Full design, drawbacks, and alternatives are in the KEP.

## Phases

Each phase is a sub-issue of this epic. Sequencing: P1 → P2 → P3 form the critical path; P4 and P5 depend on P3; P6 and P7 depend on P5.

| Phase | Scope | Depends on |
|-------|-------|------------|
| P1 | Reactor core (`src/reactor.zig`, kqueue + epoll backends, timer heap) | — |
| P2 | Scheduler integration (`io_waiting`, `runSchedulerStep`, timer-heap fold, fiber-table growth, live-window saves) | P1 |
| P3 | Port layer (non-blocking reads/writes, write buffering, `(read)` EAGAIN path, close discipline) | P2 |
| P4 | WASI backend (`poll_oneoff`) | P3 |
| P5 | Ecosystem (`http-listen-fiber`, validation, docs) | P3 |
| P6 | Secure/DB paths (reactor-aware TLS, async libpq) | P5 |
| P7 | Performance (edge-triggered evaluation, benchmarks, per-fiber memory) | P5 |

[...]

PR #1453, opened 2026-07-11
Scheduler integration: io_waiting, runSchedulerStep, reactor park (KEP-0001 P2)

### LINKED ISSUE #1440: Scheduler integration: io_waiting state, runSchedulerStep, reactor park (KEP-0001 P2)

baijum:
## Summary

Part of #1438 (KEP-0001, Phase 2). Closes #1440. Depends on Phase 1 (#1439/#1446, merged). Design: [KEP-0001 §3](https://github.com/kaappi/keps/blob/main/keps/0001-event-loop-reactor.md#3-scheduler-integration).

Wires the Phase 1 reactor into the fiber scheduler:

- **New `io_waiting` fiber status** plus `io_fd`/`io_interest`/`io_buffer` fields on `Fiber`. `schedule()` excludes it; `hasRunnableFibers()` includes it (used only by the reactor-park deadlock check, so it deliberately does *not* count `.running` — under recursive dispatch every fiber on the current call chain back to main is `.running`, so counting it would make a genuine deadlock spin in `reactor.poll()` forever instead of ever being detected).
- **One shared `runSchedulerStep`** replaces four structurally identical scheduler-dispatch loops (channel-receive/fiber-join, `thread-join!`, `mutex-lock!`, condition-variable waits). Its idle branch (`parkOnReactor`) replaces both the bare `break` and `scheduleOrTimeout`'s whole-thread `nanosleep` with a reactor park, bounded by the reactor's own timer heap.
- **`thread-sleep!`** is reimplemented as a timed reactor park instead of a blocking `nanosleep!` — sibling fibers now run while one sleeps.
- **Growable fiber table**: `FiberScheduler.fibers` is now a `std.ArrayList`, not a fixed 64-slot array. `spawn` beyond 64 concurrently-live fibers works; slot reuse for completed/errored fibers is preserved.
- **Live-window save/restore** (resolved question 5): per-fiber `saveCurrentFiber`/`restoreFiber` now copy only the live register/frame span (reusing the same `frameWindow()` computation `markFiberState` already used for GC marking) instead of the VM's entire register file, and per-fiber storage starts small (256 registers / 32 frames) instead of matching the VM's own capacity.
- **GC rooting**: `io_buffer` traced in `markFiberState`/`referencesYoung`; belt-and-braces `Reactor.markRoots`.

## Bugs found and fixed along the way

Both surfaced by this refactor, neither present before; both now covered by new tests.

1. **Struct-layout hazard**: `types.Object.as()` relied on every GC-tracked struct's `header: Object` field sitting at byte offset 0 — true by luck, not by any Zig guarantee (`struct` layout is unspecified for non-`extern` types). Adding fields to `Fiber` changed the compiler's chosen layout, moving `header` off offset 0 and silently breaking the pointer-tag-check pattern used throughout the interpreter (`isFiber`, `toObject(v).tag`, etc.). Fixed by switching every `Fiber`→`Value` encoding site to `makePointer(@ptrCast(&fiber.header))` and `Object.as()` to `@fieldParentPtr("header", self)`, both layout-independent.
2. **Use-after-free**: native-primitive `args: []const Value` slices point directly into `vm.registers`, which `runSchedulerStep` can reallocate (`ensureRegisterCapacity`) while recursively dispatching other fibers. Reading `args[...]` *after* that recursive dispatch (as the original code for `channel-receive`, `fiber-join`, and `mutex-lock!`'s owner-arg handling did) was reading potentially-freed memory. Fixed by capturing the needed `Value`s into locals before the recursive call, matching the pattern the pre-existing `channel-receive`/`fiber-join` code already used for this exact reason.

## Test plan

- [x] `zig build test` — 749/749 pass (5 new tests in `tests_scheduler.zig`)
- [x] `zig build test -Dgc-stress=true` — clean
- [x] `bash tests/scheme/run-all.sh` — 1830/1830 pass, 0 fail
- [x] 3 new Scheme smoke tests: >64 concurrently-live fibers, `thread-sleep!` doesn't stall a sibling, `channel-receive` resolves against a peer in a timed sleep (the false-deadlock scenario from the KEP's Motivation section)
- [x] Updated `primitives_fiber-audit.scm`'s spawn-limit test — its premise (a hard 64-fiber cap) is exactly what this phase removes, so it now expects the 70-spawn loop to succeed instead of erroring
- [x] Verified a `tests/scheme/coverage/srfi18-coverage.scm` failure ("cv-signal wakes waiter") is **pre-existing** on `main` (reproduces identically on `origin/main` before this branch, 5/5 runs) — unrelated cross-OS-thread mutex/condvar bug, flagged separately, not touched here

No I/O primitive changes in this PR — `readOneByte` etc. stay blocking. That's Phase 3 (#1441).

### DIFF
--- a/src/fiber.zig
+++ b/src/fiber.zig
@@ -2,12 +2,12 @@ const std = @import("std");
 const types = @import("types.zig");
 const vm_mod = @import("vm.zig");
 const memory = @import("memory.zig");
+const reactor_mod = @import("reactor.zig");
+const Reactor = reactor_mod.Reactor;
 const Value = types.Value;
 const VM = vm_mod.VM;
 const VMError = vm_mod.VMError;
 const CallFrame = types.CallFrame;
-
-pub const MAX_FIBERS = 64;
 
 pub fn clockNs() u64 {
     var ts: std.c.timespec = undefined;
@@ -22,6 +22,7 @@ pub const FiberStatus = enum(u8) {
     completed,
     errored,
     waiting,
+    io_waiting,
 };
 
 pub const Fiber = struct {
@@ -48,43 +49,53 @@ pub const Fiber = struct {
     timed_out: bool = false,
     terminated: bool = false,
     os_thread: ?std.Thread = null,
+    /// Set together with `.io_waiting` when a blocking I/O primitive parks
+    /// on the reactor (KEP-0001 Phase 3). Unused by anything shipped in
+    /// Phase 2, but the fields must exist now so the scheduler/GC plumbing
+    /// can be built and tested ahead of the port-layer primitives.
+    io_fd: ?std.posix.fd_t = null,
+    io_interest: reactor_mod.Interest = .read,
+    /// Pins an in-flight read/write buffer across GC while a primitive is
+    /// parked on the reactor. Traced by markFiberState/referencesYoung.
+    io_buffer: Value = types.VOID,
 };
 
 pub const FiberScheduler = struct {
-    fibers: [MAX_FIBERS]?*Fiber,
-    fiber_count: usize,
+    /// Growable — no fiber-count ceiling (KEP-0001 Phase 2). Slots are
+    /// reused before growing (see addFiber), so long-running spawn/join
+    /// churn doesn't grow this unboundedly; concurrently-live fiber count
+    /// is bounded only by memory.
+    fibers: std.ArrayList(?*Fiber),
     current_idx: usize,
     next_id: u32,
     vm: *VM,
 
     pub fn init(vm: *VM) FiberScheduler {
         return .{
-            .fibers = .{null} ** MAX_FIBERS,
-            .fiber_count = 0,
+            .fibers = .empty,
             .current_idx = 0,
             .next_id = 0,
             .vm = vm,
         };
     }
 
+    pub fn deinit(self: *FiberScheduler, allocator: std.mem.Allocator) void {
+        self.fibers.deinit(allocator);
+    }
+
     pub fn addFiber(self: *FiberScheduler, fiber: *Fiber) !void {
-        if (self.fiber_count < MAX_FIBERS) {
-            self.fibers[self.fiber_count] = fiber;
-            self.fiber_count += 1;
-            return;
-        }
-        for (self.fibers[0..self.fiber_count], 0..) |f, i| {
+        for (self.fibers.items, 0..) |f, i| {
             if (f) |existing| {
                 if (existing.status == .completed or existing.status == .errored) {
-                    self.fibers[i] = fiber;
+                    self.fibers.items[i] = fiber;
                     return;
                 }
             } else {
-                self.fibers[i] = fiber;
+                self.fibers.items[i] = fiber;
                 return;
             }
         }
-        return VMError.StackOverflow;
+        try self.fibers.append(self.vm.gc.allocator, fiber);
     }
 
     pub fn spawnFiber(self: *FiberScheduler, thunk: Value) !*Fiber {
@@ -100,7 +111,7 @@ pub const FiberScheduler = struct {
         if (types.isClosure(thunk_val)) {
             closure = types.toObject(thunk_val).as(types.Closure);
         } else {
-            var fiber_root = types.makePointer(@ptrCast(fiber));
+            var fiber_root = types.makePointer(@ptrCast(&fiber.header));
             self.vm.gc.pushRoot(&fiber_root);
             const trampoline = try wrapInTrampoline(self.vm.gc, thunk_val);
             self.vm.gc.popRoot();
@@ -169,21 +180,50 @@ pub const FiberScheduler = struct {
         return closure_val;
     }
 
+    /// Upper bound (exclusive) of registers actually reachable from any
+    /// currently active frame: max over `frames` of `base + frameWindow()`,
+    /// clamped to `cap`. Shared by markFiberState (GC-marking bounds) and
+    /// saveCurrentFiber/restoreFiber (memcpy bounds) — a register beyond
+    /// every active frame's window can never be read again by the fiber
+    /// that owns `frames`, so it's neither live for GC nor worth copying.
+    fn liveRegisterSpan(frames: []const CallFrame, cap: usize) usize {
+        var span: usize = 0;
+        for (frames) |f| {
+            const end: usize = @min(@as(usize, f.base) + f.frameWindow(), cap);
+            if (end > span) span = end;
+        }
+        return span;
+    }
+
+    fn growFiberRegisters(allocator: std.mem.Allocator, fiber: *Fiber, needed: usize) void {
+        if (needed <= fiber.registers.len) return;
+        var new_cap = fiber.registers.len;
+        while (new_cap < needed) new_cap *= 2;
+        const new_regs = allocator.alloc(Value, new_cap) catch return;
+        allocator.free(fiber.registers);
+        fiber.registers = new_regs;
+    }
+
+    fn growFiberFrames(allocator: std.mem.Allocator, fiber: *Fiber, needed: usize) void {
+        if (needed <= fiber.frames.len) return;
+        var new_cap = fiber.frames.len;
+        while (new_cap < needed) new_cap *= 2;
+        const new_frames = allocator.alloc(CallFrame, new_cap) catch return;
+        allocator.free(fiber.frames);
+        fiber.frames = new_frames;
+    }
+
+    /// Copies only the live register/frame window (see liveRegisterSpan)
+    /// instead of the VM's entire register file — fiber switch cost no
+    /// longer scales with the VM's peak register-file capacity (KEP-0001
+    /// Phase 2, resolved question 5).
     pub fn saveCurrentFiber(self: *FiberScheduler) void {
-        const fiber = self.fibers[self.current_idx] orelse return;
+        const fiber = self.fibers.items[self.current_idx] orelse return;
         const vm = self.vm;
-        // Grow fiber's backing arrays if the VM grew beyond them
-        if (vm.registers.len > fiber.registers.len) {
-            const new_regs = vm.gc.allocator.alloc(Value, vm.registers.len) catch return;
-            vm.gc.allocator.free(fiber.registers);
-            fiber.registers = new_regs;
-        }
-        if (vm.frames.len > fiber.frames.len) {
-            const new_frames = vm.gc.allocator.alloc(CallFrame, vm.frames.len) catch return;
-            vm.gc.allocator.free(fiber.frames);
-            fiber.frames = new_frames;
-        }
-        @memcpy(fiber.registers[0..vm.registers.len], vm.registers);
+        const span = liveRegisterSpan(vm.frames[0..vm.frame_count], vm.registers.len);
+        growFiberRegisters(vm.gc.allocator, fiber, span);
+        growFiberFrames(vm.gc.allocator, fiber, vm.frame_count);
+        @memcpy(fiber.registers[0..span], vm.registers[0..span]);
         fiber.frame_count = vm.frame_count;
         @memcpy(fiber.frames[0..vm.frame_count], vm.frames[0..vm.frame_count]);
         @memcpy(fiber.handler_stack[0..vm.handler_count], vm.handler_stack[0..vm.handler_count]);
@@ -196,20 +236,12 @@ pub const FiberScheduler = struct {
     }
 
     pub fn restoreFiber(self: *FiberScheduler, idx: usize) void {
-        const fiber = self.fibers[idx] orelse return;
+        const fiber = self.fibers.items[idx] orelse return;
         const vm = self.vm;
-        // Grow VM's backing arrays if the fiber grew beyond them
-        if (fiber.registers.len > vm.registers.len) {
-            const new_regs = vm.gc.allocator.alloc(Value, fiber.registers.len) catch return;
-            vm.gc.allocator.free(vm.registers);
-            vm.registers = new_regs;
-        }
-        if (fiber.frames.len > vm.frames.len) {
-            const new_frames = vm.gc.allocator.alloc(CallFrame, fiber.frames.len) catch return;
-            vm.gc.allocator.free(vm.frames);
-            vm.frames = new_frames;
-        }
-        @memcpy(vm.registers[0..fiber.registers.len], fiber.registers[0..fiber.registers.len]);
+        const span = liveRegisterSpan(fiber.frames[0..fiber.frame_count], fiber.registers.len);
+        vm.ensureRegisterCapacity(span) catch return;
+        vm.ensureFrameCapacity(fiber.frame_count) catch return;
+        @memcpy(vm.registers[0..span], fiber.registers[0..span]);
         @memcpy(vm.frames[0..fiber.frame_count], fiber.frames[0..fiber.frame_count]);
         vm.frame_count = fiber.frame_count;
         @memcpy(vm.handler_stack[0..fiber.handler_count], fiber.handler_stack[0..fiber.handler_count]);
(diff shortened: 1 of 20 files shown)


src/fiber.zig
```diff
@@ -290,59 +330,214 @@ pub const FiberScheduler = struct {
...
+/// thread-join!, mutex-lock!, condition-variable waits, and thread-sleep!
+/// (KEP-0001 Phase 2) — call sites differ only in `Ctx.isDone`. `Ctx` is a
+/// small value type with an `isDone(self: Ctx) bool` method (comptime duck
+/// typing), e.g. `TargetWait{ .target = f }`.
+pub fn runSchedulerStep(comptime Ctx: type, ctx: Ctx, vm: *VM, sched: *FiberScheduler, me: *Fiber) VMError!bool {
+    const my_idx = sched.current_idx;
+    sched.saveCurrentFiber();
+
+    while (!ctx.isDone() and !me.timed_out) {
+        const next_idx = sched.schedule() orelse {
```
baijum, line 516: **Confirmed regression — timed waits starve while any fiber is runnable.**

The old `schedule()` swept `.waiting` fibers with expired deadlines on every call. Now expiry is detected only in `parkOnReactor`, which is reached only when `schedule()` returns null. So as long as *any* fiber is runnable (e.g. a compute fiber that keeps yielding), an expired timed wait never flips: this loop keeps dispatching the runnable fiber, `me.timed_out` is never set, and the timeout fires only when the scheduler goes fully idle.

Measured (repro in review body): timed `mutex-lock!` with 0.05 s timeout next to a busy yielding sibling — **main fires at 0.050 s, this branch at 0.74 s** (exactly when the busy fiber finished). `thread-sleep!` oversleeps the same way, and Phase 3's `io_waiting` fibers will inherit the same starvation for fd readiness.

Suggested fix: fire due timers at the top of each loop iteration (or inside `schedule()`) with a non-blocking check — an O(1) peek of the timer-heap min against `clockNs()`, no syscall when nothing is due — flipping expired `.waiting` fibers exactly as `parkOnReactor` does. That restores the old sweep's semantics without reintroducing a per-fiber scan. Phase 3 will additionally want an occasional `poll(0)` tick for fd readiness under load.

src/fiber.zig
```diff
@@ -169,21 +180,50 @@ pub const FiberScheduler = struct {
...
+            if (end > span) span = end;
+        }
+        return span;
+    }
+
+    fn growFiberRegisters(allocator: std.mem.Allocator, fiber: *Fiber, needed: usize) void {
+        if (needed <= fiber.registers.len) return;
+        var new_cap = fiber.registers.len;
+        while (new_cap < needed) new_cap *= 2;
+        const new_regs = allocator.alloc(Value, new_cap) catch return;
```
baijum: **OOM here turns into an out-of-bounds `@memcpy` in `saveCurrentFiber`.**

If this allocation fails, the helper silently returns with the old (too small) array, and `saveCurrentFiber` proceeds to `@memcpy(fiber.registers[0..span], …)` with `span > fiber.registers.len` — a bounds panic in ReleaseSafe/Debug and UB in ReleaseFast. The old code's `catch return` skipped the *entire* save (silent state corruption — also bad, but not UB).

Suggest making `growFiberRegisters`/`growFiberFrames` return `bool` (or an error) and skipping the copy on failure, so the OOM path degrades the old way instead of copying past the buffer. Same applies to `growFiberFrames` below.

src/primitives_srfi18.zig
```diff
@@ -485,9 +474,12 @@ fn threadJoinFn(args: []const Value) PrimitiveError!Value {
...
         me.status = .waiting;
         me.timed_out = false;
-        if (deadline_ns) |d| me.deadline_ns = d;
+        if (deadline_ns) |d| {
+            me.deadline_ns = d;
+            try ctx.reactor.addTimer(d, me);
+        }
 
-        try runSchedulerUntilDone(target);
+        _ = try fiber_mod.runSchedulerStep(fiber_mod.TargetWait, .{ .target = target }, ctx.vm, ctx.sched, me);
```
baijum: **Confirmed regression — `me.deadline_ns` is never cleared here, and a later untimed wait then hangs forever.**

Both exits of this wait leave `me.deadline_ns` set: the resolved path (wake cancels the reactor timer but not the field) and the timed-out path (clears `timed_out` only). Compare `mutexLockFn`/`threadSleepFn`/`mutexUnlockFn`, which all do `me.deadline_ns = null` after their `runSchedulerStep`.

The stale field then poisons the next *untimed* `.waiting` park by this fiber: `hasRunnableFibers()` counts it as `.waiting` with a deadline, so `parkOnReactor` skips the genuine-deadlock early-out and calls `reactor.poll(null, …)` with an empty timer heap and no fds — which blocks **forever** (kqueue/epoll) or busy-spins (WASI stopgap).

Reproduced (script in review body): timed `thread-join!` on a started fiber that resolves instantly, then untimed `mutex-lock!` on a mutex held by a permanently parked fiber — main returns after ~5 s (its own pre-existing stale-deadline wart: the old `scheduleOrTimeout` slept out the stale deadline and faked a timeout), but this branch **hangs until killed**. This PR converts a bounded wart into an unbounded hang.

Minimal fix: `me.deadline_ns = null` after this `runSchedulerStep` (both branches). Nicer fix: centralize it in `runSchedulerStep`'s exit path (it already restores `me` and sets `.running`), optionally with a belt-and-braces `removeTimer(me)`, so no call site can forget again.

src/primitives_srfi18.zig
```diff
@@ -390,7 +379,7 @@ fn threadTerminateFn(args: []const Value) PrimitiveError!Value {
     // Atomic: for OS threads the child VM polls this flag concurrently.
     @atomicStore(bool, &fiber.terminated, true, .monotonic);
 
-    if (memory.gc_instance) |gc| abandonFiberMutexes(gc, fiber, ctx.sched);
+    if (memory.gc_instance) |gc| fiber_mod.abandonFiberMutexes(gc, fiber, ctx.sched);
```
baijum, line 382: **Minor: the terminated fiber's pending reactor timer isn't cancelled.**

If the victim was in a timed wait, its `TimerEntry` stays in the heap after this flips it to `.errored`. Consequences: (a) `reactor.isEmpty()` stays false until the stale deadline passes, so genuine-deadlock detection in `parkOnReactor` is delayed by up to the dead fiber's remaining timeout (it polls instead of reporting deadlock); (b) once `addFiber` reuses the `.errored` slot, the fiber object is reachable *only* through the timer heap — which means `Reactor.markRoots` is load-bearing here, not "currently redundant" as its doc comment claims (worth rewording that comment either way).

Suggest `if (ctx.vm.reactor) |r| r.removeTimer(fiber);` alongside the status flip.

baijum (commented):
Reviewed with both binaries built and behavioral A/B repros run (main `28f1a337` vs this head `643c50f1`). The consolidation itself is solid — `runSchedulerStep` faithfully unifies the four dispatch loops, the `@fieldParentPtr`/`&fiber.header` layout fix is the right call, the UAF captures are correct and well-commented, and the new unit/smoke tests target exactly the new machinery. However, I found **two confirmed behavioral regressions** around timer handling (repros below), plus two smaller issues. Details in the inline comments.

**Confirmed regressions (both reproduced against main):**

1. **Timer expiry starves while any fiber is runnable.** The old `schedule()` swept expired deadlines on every call; now expiry is only detected in `parkOnReactor`, which runs only when the scheduler is idle. A timed `mutex-lock!` with a 0.05 s timeout next to a busy yielding sibling: main fires at **0.050 s**, this PR fires at **0.74 s** (= when the busy fiber finished). Same starvation applies to `thread-sleep!` and timed joins, and Phase 3 will inherit it for fd readiness.

2. **Stale `deadline_ns` after `thread-join!` → permanent hang in `poll(null)`.** `threadJoinFn` never clears `me.deadline_ns` (both the resolved and timed-out branches). A later *untimed* wait then makes `hasRunnableFibers()` report the fiber as "waiting with a deadline" while the reactor has no timer, so `parkOnReactor` blocks forever in `poll(null)`. Repro: timed join that resolves early, then untimed `mutex-lock!` on a dead-held mutex — main returns after 5 s (its own pre-existing stale-deadline wart), this PR **hangs until killed**.

**Minor:**
- `growFiberRegisters`/`growFiberFrames` swallow OOM and let `saveCurrentFiber` `@memcpy` out of bounds (panic in ReleaseSafe, UB in ReleaseFast) — see inline.
- `thread-terminate!` doesn't cancel the victim's pending reactor timer — see inline. This same path also falsifies the "currently redundant" claim in `Reactor.markRoots`'s doc comment (a terminated fiber whose slot is reused is reachable *only* through the timer heap, so that root is load-bearing — worth rewording the comment).

**Small notes, no action needed:**
- `WasiBackend.wait(null)` returns immediately instead of blocking; today that's unreachable, but if issue 2's state is ever hit on WASI it becomes a busy spin rather than a hang. A comment or `unreachable` assert would pin the assumption.
- `addFiber` is now an O(live-fibers) scan per spawn (O(n²) to build n concurrent fibers). Fine at current scales; a free-slot hint would fix it if "thousands of fibers" becomes real.
- `spawnFiber` memsets only `INITIAL_FIBER_REGISTER_CAPACITY` (256) registers to UNDEFINED while `frameWindow()` can theoretically exceed 256; today the compiler's register-pressure limits make >256-local thunks uncompilable (`TooManyLocals`), so this is unreachable — noting it in case those limits ever loosen.

**Verdict: not ready to merge yet.** Both confirmed items are regressions this PR introduces (or converts from a bounded wart into an unbounded hang) on its core paths, and neither is caught by the current test suite — worth fixing plus a regression test each (my repros can be adapted directly). CI was still in progress at review time. Everything else — the refactor structure, GC tracing of `io_buffer`, the growable fiber table, live-window save/restore — looks correct and well-tested.

Repro scripts (run under a watchdog):

```scheme
;; (1) timeout starvation — expect lock-result #f at ~0.05s; PR gives ~0.74s
(import (scheme base) (scheme write) (kaappi fibers) (srfi 18))
(define m (make-mutex))
(mutex-lock! m)
(define waiter (spawn (lambda ()
  (let ((t0 (time->seconds (current-time))))
    (let ((r (mutex-lock! m 0.05)))
      (list 'lock-result r 'elapsed (- (time->seconds (current-time)) t0)))))))
(define busy (spawn (lambda ()
  (let loop ((i 0)) (if (< i 8000000) (begin (yield) (loop (+ i 1))) 'done)))))
(display (fiber-join waiter)) (newline)
```

```scheme
;; (2) stale deadline_ns — main: returns #f after ~5s; PR: hangs forever
(import (scheme base) (scheme write) (kaappi fibers) (srfi 18))
(define f1 (spawn (lambda () (yield) 'quick)))
(define kick (spawn (lambda () 'kick)))
(fiber-join kick)                         ; let f1 start and yield
(display (thread-join! f1 5.0)) (newline) ; fiber path; resolves instantly, deadline_ns never cleared
(define m (make-mutex))
(define ch (make-channel))
(define holder (spawn (lambda ()
  (mutex-lock! m) (channel-send ch 'locked) (channel-receive (make-channel)))))
(display (channel-receive ch)) (newline)
(display (mutex-lock! m)) (newline)       ; untimed lock on dead-held mutex -> hangs on this PR
```

baijum:
Re-verified `349c48f6` against the same A/B repros from my review — **both confirmed regressions are fixed**:

1. **Timer starvation**: timed `mutex-lock!` (0.05 s) next to the 8M-yield busy sibling now fires at **0.050 s** (was 0.74 s). The per-tick `popExpiredTimers` in `schedule()` restores the old sweep's promptness at heap-peek cost.
2. **Stale `deadline_ns` hang**: the timed-join-then-untimed-lock repro now raises `mutex-lock!: deadlock — mutex will never be released` in 0.01 s (was an unbounded hang; pre-PR main faked a 5 s timeout).

Also verified locally on this head: `zig build test` green, fiber audit (32 passes), all 4 fiber smoke tests, `srfi18.scm` 44/44, abandoned-mutex and dynamic-wind suites. The bonus fixes are good ones — erroring on provable deadlock instead of silently stealing the lock / returning VOID, the `ctx.isDone()`-before-epilogue capture (which the CondVarWait path needed for correctness), and the OOM propagation through save/restore.

Two small follow-ups, neither blocking:

- **Missing regression tests** for the two fixed bugs (repo policy: every bug fix needs a test that fails without it). Both repros from the review body adapt directly into `tests/scheme/smoke/`: the starvation one asserts elapsed < some generous bound; the stale-deadline one asserts a deadlock error is raised via `guard` (on regression it hangs, which CI's timeout catches).
- Nit: in `popExpiredTimers`, `self.timers.pop()` happens inside the `append` argument, so an OOM on append loses the popped entry and its fiber is never flipped (stuck `.waiting` forever). An `ensureUnusedCapacity` before the loop (as `poll` already does for fd waiters) would close it.

With CI green I'd consider this ready to merge — ideally with the two regression tests added first.

baijum:
Verified `5d46fe9e` — all follow-ups from my previous comment are resolved:

- **`popExpiredTimers` OOM ordering**: fixed exactly as suggested (append before pop, so an OOM leaves the timer in the heap instead of stranding the fiber).
- **Regression tests**: both verified against the pre-fix binary (`643c50f1`) to confirm they actually catch the bugs, per the repo's fails-without-fix rule:
  - `fiber-timed-mutex-lock-not-starved-by-busy-sibling.scm` — **fails** on pre-fix (elapsed ~0.74 s > 0.3 s bound, exit 1) and **passes** on this head (2/2, 1.4 s ReleaseSafe).
  - `fiber-thread-join-deadline-cleared-after-resolve.scm` — **hangs** on pre-fix (runner timeout catches it) and **passes** on this head (3/3, instant deadlock error).
- **Debug CI timing** (the 8M-yield test runs in the Debug matrix job): measured 10.1 s under `-Doptimize=Debug` locally — comfortably within budget, no skip-list entry needed.

Nothing further from me. **LGTM — ready to merge once the pending CI jobs go green.**

PR #1455, opened 2026-07-11
Fix mutex-lock!/mutex-unlock!+condvar giving up instantly across OS threads

### LINKED ISSUE #1454: SRFI-18: mutex-lock!/mutex-unlock!+condvar give up instantly across real OS threads, corrupting lock state
## Summary

`mutex-lock!` and `mutex-unlock!` (with a condition-variable argument) can report success without the guarded state actually changing, when called from a real SRFI-18 OS thread (`make-thread`/`thread-start!`) sharing a mutex/condvar with another thread via a top-level global.

## Root cause

Each OS thread runs its own independent `FiberScheduler` (`vm.scheduler` is per-thread). `mutex-lock!` and `mutex-unlock!+condvar-wait` park the calling fiber and drive that local scheduler until nothing is runnable, then give up. For a real OS thread whose only fiber is its own top-level thunk, "nothing locally runnable" is always true the moment it blocks on a mutex/condvar held by a *different* OS thread — there's nothing else for that scheduler to run.

Both call sites then unconditionally proceeded as if the wait had succeeded, without re-checking whether the guarded state (`Mutex.locked`, or a condition-variable signal) had actually changed:

- `mutex-lock!` on a mutex still held by another thread returns `#t` immediately and sets `locked = true`, silently corrupting the lock (two threads can believe they both hold it).
- `mutex-unlock!` with a condition-variable argument returns before any `condition-variable-signal!`/`-broadcast!` has happened.

### Reproduction (timing-based, using a top-level global to share the mutex across threads)

```scheme

(define m (make-mutex))
(define elapsed -1.0)

(mutex-lock! m)
(define t (make-thread
           (lambda ()
             (define t0 (current-time))
             (mutex-lock! m)
             (set! elapsed (- (time->seconds (current-time)) (time->seconds t0))))))
(thread-start! t)
(thread-sleep! 0.3)
(mutex-unlock! m)
(thread-join! t)
(display elapsed) (newline)
;; Buggy: prints ~0.00001 (returned instantly, mutex still "held" by main)
;; Correct: should print ~0.3 (blocked until main released the mutex)
```

### Note on thread thunk closures

Mutex/condition-variable objects **cannot** be captured directly in a `make-thread` thunk's closure — deep-copying such a thunk correctly raises "uncopyable type" (see `tests/scheme/srfi/srfi18.scm`'s "OS threads cannot capture sync primitives" test; this is by-design, since OS threads run on isolated heaps). Sharing must go through a top-level global (as above) or `(kaappi fibers)`'s same-heap `spawn`/`fiber-join`. `tests/scheme/coverage/srfi18-coverage.scm` had stale tests exercising the unsupported closure-capture pattern, which is why the reported symptom for those specific tests was an uncaught "uncopyable type" exception rather than the timing bug directly.

## Fix

- Poll the guarded state (`Mutex.locked` / a new `ConditionVariable.signal_generation` counter) at a short interval (matching `thread-join!`'s existing 1ms OS-thread poll cadence) instead of giving up immediately, when nothing is locally schedulable **and** another OS thread could plausibly still change that state.
- Still give up immediately (unchanged from before) when no other OS thread exists at all — otherwise a genuine local fiber-only deadlock would hang forever instead of terminating.
- Snapshot the condition variable's signal generation *before* releasing the mutex in `mutex-unlock!`, to avoid a lost-wakeup race against a signaler that acquires the mutex in the gap.
- Made `Mutex.locked` atomic, since it's now a cross-thread polling condition.

## Regression test

`tests/scheme/srfi/srfi18-cross-thread-wait.scm` (tracked, runs in `run-all.sh`) exercises `mutex-lock!` blocking across real OS threads sharing a mutex via a global, plus `mutex-unlock!+condition-variable-signal!/-broadcast!`, all with timing assertions that fail reliably without the fix and pass reliably with it.

`tests/scheme/coverage/srfi18-coverage.scm`'s three affected tests were updated to use the supported same-heap fiber pattern (`spawn`/`fiber-join`, matching `srfi18.scm`) instead of the unsupported OS-thread-closure-capture pattern.

baijum:
## Summary

- Each OS thread runs its own independent `FiberScheduler`, so `mutex-lock!` and `mutex-unlock!+condition-variable-wait` had nothing to do the moment nothing was locally schedulable — always true for a real OS thread whose only fiber is its own top-level thunk. Both then unconditionally reported success without re-checking whether the guarded state had actually changed: `mutex-lock!` on a mutex still held by another thread returned `#t` at once (silently corrupting the lock), and `mutex-unlock!` with a condition variable returned before any signal had happened.
- Fix: poll the shared state (`Mutex.locked` / a new `ConditionVariable.signal_generation` counter) at a short interval instead of giving up immediately, but only when another OS thread could plausibly still change it (`crossThreadWaitPossible`) — a genuine local-only fiber deadlock still gives up right away, otherwise it would hang forever instead of terminating (which I hit and fixed during development). The condvar generation is snapshotted *before* releasing the mutex in `mutex-unlock!` to avoid a lost-wakeup race against a signaler that acquires the mutex in the gap.
- Mutex acquisition is now a `@cmpxchgStrong`-based claim (`tryClaimMutex`/`tryClaimAbandoned`) instead of a plain load-then-store, since cross-thread contention on the same mutex is now a real, tested code path rather than something that resolved (buggily) on the first check. `owner`/`abandoned` are published *before* the lock's release-store everywhere they're set, so a cross-thread acquirer that wins the race can't have its owner write silently stomped back to `VOID`.
- Also updates the stale `coverage/srfi18-coverage.scm` tests that captured a mutex/condvar in a thread thunk closure — unsupported by design, since OS threads run on isolated heaps (deep-copying such a closure correctly raises "uncopyable type", see `srfi18.scm`'s existing "OS threads cannot capture sync primitives" test) — to use the same-heap fiber pattern (`spawn`/`fiber-join`) that `srfi18.scm` already demonstrates works.

Fixes #1454.

### Rebased onto KEP-0001 Phase 2 (#1453)

`main` landed a concurrent refactor that collapsed the four scheduler-dispatch loops (mutex-lock!, condvar-wait, thread-join!, channel/fiber-join) into one shared `runSchedulerStep`/`parkOnReactor` backed by a real epoll/kqueue reactor. That refactor didn't fix this bug — it just changed the symptom from "silently corrupts and reports success" to "raises a spurious deadlock error," since it still couldn't tell a real local deadlock apart from a wait a *different* OS thread could still resolve. This PR now teaches it that distinction via `crossThreadWaitPossible`.

Rebasing surfaced one more real bug: a user-specified timeout made `parkOnReactor`'s blocking reactor wait swallow the *entire* timeout duration before ever re-checking for a cross-thread signal (confirmed by a real, reproducible test failure, not just reasoning). Fixed with a narrowly-scoped `pollCapNs()` duck-typed method that only the two SRFI-18 wait types opt into, capping the reactor's blocking wait to 1ms when a cross-thread resolution is plausible — every other caller (channel-receive, fiber-join, thread-join!, thread-sleep!) is unaffected and unchanged.

## Test plan

- [x] Tracked regression test `tests/scheme/srfi/srfi18-cross-thread-wait.scm`: real OS threads sharing a mutex/condvar via top-level globals, with timing assertions (6 checks, including a bounded-timeout + result assertion on the broadcast case so a missed signal fails loudly instead of hanging CI). Verified it fails reliably without the fix (checked against the true pre-fix baseline in a separate worktree, both before and after the KEP-0001 Phase 2 rebase) and passes reliably with it, across many repeated runs.
- [x] `tests/scheme/coverage/srfi18-coverage.scm` — all 54 checks pass reliably.
- [x] Full suite: `bash tests/scheme/run-all.sh` — 1833/1833 pass (438 Scheme files + 1395 R7RS tests, including the new KEP-0001 Phase 2 scheduler-stress tests), no regressions.
- [x] `zig build test` — all unit tests pass.
- [x] `zig fmt --check` — clean.

### DIFF
--- a/src/primitives_srfi18.zig
+++ b/src/primitives_srfi18.zig
@@ -717,6 +717,7 @@ fn mutexLockFn(args: []const Value) PrimitiveError!Value {
     if (tryClaimMutex(m)) {
         const ctx = try ensureScheduler();
         m.owner = resolveMutexOwner(args, if (ctx.vm.current_fiber) |cf| types.makePointer(@ptrCast(&cf.header)) else types.VOID);
+        if (memory.gc_instance) |gc| gc.writeBarrier(&m.header, m.owner);
         if (tryClaimAbandoned(m)) return raiseError(.abandoned_mutex, "mutex was abandoned", types.VOID);
         return types.TRUE;
     }
@@ -762,6 +763,18 @@ fn mutexLockFn(args: []const Value) PrimitiveError!Value {
         }
         if (done) {
             if (tryClaimMutex(m)) break;
+            // A local wake (the usual way runSchedulerStep reports "done")
+            // cancels me's pending reactor timer via cancelPendingTimer.
+            // Losing the claim race here means we're going back to waiting
+            // with the timer gone but me.deadline_ns still set -- re-add it
+            // (remove-first keeps this idempotent) or a timed mutex-lock!
+            // could block past its deadline, in the worst case unboundedly
+            // if crossThreadWaitPossible later turns false with no timer
+            // left to bound parkOnReactor's blocking wait.
+            if (deadline) |d| {
+                ctx.reactor.removeTimer(me);
+                try ctx.reactor.addTimer(d, me);
+            }
             me.status = .waiting;
             continue;
         }
@@ -771,6 +784,13 @@ fn mutexLockFn(args: []const Value) PrimitiveError!Value {
             return raiseError(.general, "mutex-lock!: deadlock — mutex will never be released (all fibers blocked)", types.VOID);
         }
         sleepNs(CROSS_THREAD_POLL_NS);
+        // Same timer restoration as above: a local wake earlier in this
+        // loop (see the `done` branch) may have already canceled the
+        // timer, and that state persists into this branch too.
+        if (deadline) |d| {
+            ctx.reactor.removeTimer(me);
+            try ctx.reactor.addTimer(d, me);
+        }
         me.status = .waiting;
     }
     // A cross-thread resolution never runs local wake bookkeeping (the
@@ -786,6 +806,7 @@ fn mutexLockFn(args: []const Value) PrimitiveError!Value {
         (if (types.isFiber(oa)) oa else if (oa == types.FALSE) types.VOID else types.makePointer(@ptrCast(&me.header)))
     else
         types.makePointer(@ptrCast(&me.header));
+    if (memory.gc_instance) |gc| gc.writeBarrier(&m.header, m.owner);
 
     if (tryClaimAbandoned(m)) return raiseError(.abandoned_mutex, "mutex was abandoned", types.VOID);
     return types.TRUE;
--- a/src/primitives_srfi18.zig
+++ b/src/primitives_srfi18.zig
@@ -104,6 +104,42 @@ fn ensureScheduler() @TypeOf(fiber_mod.ensureScheduler(undefined)) {
     return fiber_mod.ensureScheduler(vm);
 }
 
+// 1ms, matching thread-join!'s existing OS-thread poll cadence.
+const CROSS_THREAD_POLL_NS: u64 = 1_000_000;
+
+// Number of thread-start!-spawned OS threads currently alive (incremented in
+// threadStartFn, decremented via defer in threadEntryFn on every exit path).
+// Lets crossThreadWaitPossible tell a real cross-OS-thread wait apart from a
+// genuine local deadlock -- see its comment.
+var live_child_threads: usize = 0;
+
+// True when some *other* OS thread could plausibly still change the mutex/
+// condvar state this scheduler is blocked on, so polling (instead of
+// accepting runSchedulerStep's "done == false" as a genuine deadlock) might
+// eventually pay off:
+//   - A spawned child thread's own scheduler (vm.owns_globals == false) may
+//     always be waiting on the main thread, which is "alive" for as long as
+//     the process runs -- always poll.
+//   - The main thread's scheduler only has something to gain from polling
+//     if at least one child thread currently exists to possibly unlock/
+//     signal from the outside; with none, runSchedulerStep reporting "not
+//     done" is a real, unrecoverable local deadlock (fiber.zig's
+//     parkOnReactor already found nothing locally runnable and no pending
+//     timer/fd event).
+//
+// Asymmetry this creates: a child thread that manages to genuinely
+// self-deadlock (e.g. waits on a mutex only it could ever unlock) now polls
+// forever rather than raising the deadlock error runSchedulerStep's "not
+// done" normally produces, since it always assumes the main thread might
+// still help -- indefinite blocking here is conformant with SRFI-18, but the
+// same shape of deadlock hangs on a child thread and errors on the main
+// thread, which is worth knowing when debugging one.
+fn crossThreadWaitPossible() bool {
+    const vm = vm_mod.vm_instance orelse return false;
+    if (!vm.owns_globals) return true;
+    return @atomicLoad(usize, &live_child_threads, .acquire) > 0;
+}
+
 fn timeoutToDeadlineNs(timeout: Value) PrimitiveError!?u64 {
     if (timeout == types.FALSE) return null;
     if (types.isSrfi18Time(timeout)) {
@@ -227,14 +263,27 @@ fn threadStartFn(args: []const Value) PrimitiveError!Value {
     gc.extra_roots.append(gc.allocator, args[0]) catch return PrimitiveError.OutOfMemory;
 
     fiber.status = .running;
+    // Increment *before* spawning: if the child ran to completion before an
+    // increment placed after std.Thread.spawn executed, the decrement in
+    // threadEntryFn's defer could fire first and wrap the counter (or drop
+    // it one below the true count with other children alive), letting
+    // crossThreadWaitPossible wrongly conclude no other thread exists.
+    _ = @atomicRmw(usize, &live_child_threads, .Add, 1, .release);
     fiber.os_thread = std.Thread.spawn(.{}, threadEntryFn, .{
         fiber, gc.allocator, gc, vm,
-    }) catch return PrimitiveError.OutOfMemory;
+    }) catch {
+        _ = @atomicRmw(usize, &live_child_threads, .Sub, 1, .release);
+        return PrimitiveError.OutOfMemory;
+    };
 
     return args[0];
 }
 
 fn threadEntryFn(fiber: *fiber_mod.Fiber, allocator: std.mem.Allocator, parent_gc: *memory.GC, parent_vm: *vm_mod.VM) void {
+    // Balances the increment in threadStartFn on every exit path (including
+    // the early GC/VM-init failures below), so crossThreadWaitPossible's "is
+    // another OS thread still alive" check stays accurate.
+    defer _ = @atomicRmw(usize, &live_child_threads, .Sub, 1, .release);
     _ = parent_gc;
     const child_gc = allocator.create(memory.GC) catch {
         @atomicStore(fiber_mod.FiberStatus, &fiber.status, .errored, .release);
@@ -619,8 +668,8 @@ fn mutexStateFn(args: []const Value) PrimitiveError!Value {
     if (!types.isMutex(args[0]))
         return primitives.typeError("mutex-state", "mutex", args[0]);
     const m = types.toMutex(args[0]);
-    if (!m.locked) {
-        if (m.abandoned) {
+    if (!@atomicLoad(bool, &m.locked, .acquire)) {
+        if (@atomicLoad(bool, &m.abandoned, .acquire)) {
             const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
             return gc.allocSymbol("abandoned") catch return PrimitiveError.OutOfMemory;
         }
@@ -632,38 +681,43 @@ fn mutexStateFn(args: []const Value) PrimitiveError!Value {
     return gc.allocSymbol("not-owned") catch return PrimitiveError.OutOfMemory;
 }
 
+// The fiber/thread that resolves to be recorded as a mutex's new owner: an
+// explicit fiber (args[2]), explicitly "unowned" (args[2] == #f), or the
+// caller's own current fiber.
+fn resolveMutexOwner(args: []const Value, fallback: Value) Value {
+    if (args.len > 2 and types.isFiber(args[2])) return args[2];
+    if (args.len > 2 and args[2] == types.FALSE) return types.VOID;
+    return fallback;
+}
+
+// Atomically claims the mutex (false -> true). This is the single point of
+// arbitration between racing threads -- a plain load-then-store lets two
+// threads both observe "unlocked" and both believe they've acquired it,
+// corrupting mutual exclusion. Load-bearing now that cross-thread mutex
+// contention is a supported, polled-for wait (see crossThreadWaitPossible)
+// rather than something that resolved (buggily) on the first check.
+fn tryClaimMutex(m: *types.Mutex) bool {
+    return @cmpxchgStrong(bool, &m.locked, false, true, .acq_rel, .acquire) == null;
+}
+
+// Atomically claims (and clears) an abandoned flag. Only meaningful once
+// the caller has already won tryClaimMutex: it decides whether *this*
+// acquisition should also raise abandoned-mutex-exception, without letting
+// a second racing acquirer also see and report the same abandonment.
+fn tryClaimAbandoned(m: *types.Mutex) bool {
+    return @cmpxchgStrong(bool, &m.abandoned, true, false, .acq_rel, .acquire) == null;
+}
+
 fn mutexLockFn(args: []const Value) PrimitiveError!Value {
     if (!types.isMutex(args[0]))
         return primitives.typeError("mutex-lock!", "mutex", args[0]);
 
     const m = types.toMutex(args[0]);
 
-    if (m.abandoned) {
-        m.abandoned = false;
-        m.locked = true;
+    if (tryClaimMutex(m)) {
         const ctx = try ensureScheduler();
-        m.owner = if (args.len > 2 and types.isFiber(args[2]))
-            args[2]
-        else if (args.len > 2 and args[2] == types.FALSE)
-            types.VOID
-        else if (ctx.vm.current_fiber) |cf|
-            types.makePointer(@ptrCast(&cf.header))
-        else
-            types.VOID;
-        return raiseError(.abandoned_mutex, "mutex was abandoned", types.VOID);
-    }
-
-    if (!m.locked) {
-        m.locked = true;
-        const ctx = try ensureScheduler();
-        m.owner = if (args.len > 2 and types.isFiber(args[2]))
-            args[2]
-        else if (args.len > 2 and args[2] == types.FALSE)
-            types.VOID
-        else if (ctx.vm.current_fiber) |cf|
-            types.makePointer(@ptrCast(&cf.header))
-        else
-            types.VOID;
+        m.owner = resolveMutexOwner(args, if (ctx.vm.current_fiber) |cf| types.makePointer(@ptrCast(&cf.header)) else types.VOID);
+        if (tryClaimAbandoned(m)) return raiseError(.abandoned_mutex, "mutex was abandoned", types.VOID);
         return types.TRUE;
     }
 
@@ -691,38 +745,56 @@ fn mutexLockFn(args: []const Value) PrimitiveError!Value {
         try ctx.reactor.addTimer(d, me);
     }
 
-    const done = try fiber_mod.runSchedulerStep(MutexWait, .{ .m = m }, ctx.vm, ctx.sched, me);
+    // runSchedulerStep only returns done once it *observes* m.locked ==
+    // false; claiming it is still a race against any other thread making
+    // the same observation, so retry the claim and go back to waiting on
+    // failure instead of assuming we won. When runSchedulerStep reports
+    // "not done" (parkOnReactor found nothing locally runnable and no
+    // pending timer/fd event), that's only a genuine deadlock if no other
+    // OS thread could plausibly still unlock this mutex from the outside —
+    // otherwise poll briefly and retry.
+    while (true) {
+        const done = try fiber_mod.runSchedulerStep(MutexWait, .{ .m = m }, ctx.vm, ctx.sched, me);
+        if (me.timed_out) {
+            me.timed_out = false;
+            me.deadline_ns = null;
+            return types.FALSE;
+        }
+        if (done) {
+            if (tryClaimMutex(m)) break;
+            me.status = .waiting;
+            continue;
+        }
+        if (!crossThreadWaitPossible()) {
+            if (deadline != null) ctx.reactor.removeTimer(me);
+            me.deadline_ns = null;
+            return raiseError(.general, "mutex-lock!: deadlock — mutex will never be released (all fibers blocked)", types.VOID);
+        }
+        sleepNs(CROSS_THREAD_POLL_NS);
+        me.status = .waiting;
+    }
+    // A cross-thread resolution never runs local wake bookkeeping (the
+    // unlocking thread's scheduler/reactor doesn't even know `me` exists),
+    // so any timer registered above may still be pending; a local wake
+    // already canceled it (removeTimer is a no-op then), but skipping this
+    // for the cross-thread path would leave a stale entry that could later
+    // fire against a reused fiber slot.
+    if (deadline != null) ctx.reactor.removeTimer(me);
     me.deadline_ns = null;
 
-    if (me.timed_out) {
-        me.timed_out = false;
-        return types.FALSE;
-    }
-    if (!done) {
-        // Genuine deadlock: parkOnReactor gave up because nothing local
-        // could ever unlock this mutex (no other runnable fiber, no
-        // pending timer). Must not fall through to "assume success" —
-        // that would silently steal a lock still held elsewhere.
-        return raiseError(.general, "mutex-lock!: deadlock — mutex will never be released (all fibers blocked)", types.VOID);
-    }
-
-    m.locked = true;
     m.owner = if (owner_arg) |oa|
         (if (types.isFiber(oa)) oa else if (oa == types.FALSE) types.VOID else types.makePointer(@ptrCast(&me.header)))
     else
         types.makePointer(@ptrCast(&me.header));
 
-    if (m.abandoned) {
-        m.abandoned = false;
-        return raiseError(.abandoned_mutex, "mutex was abandoned", types.VOID);
-    }
+    if (tryClaimAbandoned(m)) return raiseError(.abandoned_mutex, "mutex was abandoned", types.VOID);
     return types.TRUE;
 }
 
 const MutexWait = struct {
     m: *types.Mutex,
     pub fn isDone(self: MutexWait) bool {
-        return !self.m.locked;
+        return !@atomicLoad(bool, &self.m.locked, .acquire);
     }
 };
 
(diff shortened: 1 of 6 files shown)


src/primitives_srfi18.zig
```diff
@@ -712,8 +786,8 @@ fn mutexLockFn(args: []const Value) PrimitiveError!Value {
         return raiseError(.abandoned_mutex, "mutex was abandoned", types.VOID);
     }
 
-    if (!m.locked) {
-        m.locked = true;
+    if (!@atomicLoad(bool, &m.locked, .acquire)) {
```
baijum: **Non-atomic test-then-set: two OS threads can both acquire the mutex.**

The acquisition here is an `@atomicLoad` followed by a separate `@atomicStore` — and the post-wait path at line 825 is an unconditional store after `runSchedulerUntilMutex` merely *observed* `locked == false`. Two threads polling the same mutex (1 ms cadence makes simultaneous wakeups after one unlock quite plausible) can both observe `false` and both store `true`; both return `#t` and mutual exclusion is violated — the same silent-corruption class this PR sets out to fix, just a narrower window.

Before this PR cross-thread locking never actually blocked, so the race was moot; now that cross-thread contention is a supported, tested behavior, the acquisition should be an atomic RMW:

```zig
if (@cmpxchgStrong(bool, &m.locked, false, true, .acq_rel, .acquire) == null) {
    // acquired
}
```

and the post-wait path should retry the cmpxchg (returning to the wait loop on failure) rather than blind-storing `true`, e.g.:

```zig
while (true) {
    try runSchedulerUntilMutex(m, me);
    if (me.timed_out) { ... }
    if (@cmpxchgStrong(bool, &m.locked, false, true, .acq_rel, .acquire) == null) break;
}
```

(The abandoned-mutex path at line 776 has the same shape, though that one predates this PR.)

src/primitives_srfi18.zig
```diff
@@ -820,13 +897,25 @@ fn mutexUnlockFn(args: []const Value) PrimitiveError!Value {
...
+    // while we still exclusively hold it. Per the SRFI-18 protocol, any
+    // signaler must acquire this same mutex before calling
+    // condition-variable-signal!/-broadcast!, so it cannot have bumped the
+    // generation yet -- snapshotting after the unlock would race a signaler
+    // that acquires the mutex in the gap and produce a lost wakeup (the
+    // waiter would then wait for a *second* signal that may never come).
+    const cv: ?*types.ConditionVariable = if (has_cv) types.toConditionVariable(args[1]) else null;
+    const start_gen: u64 = if (cv) |c| @atomicLoad(u64, &c.signal_generation, .acquire) else 0;
+
+    @atomicStore(bool, &m.locked, false, .release);
```
baijum: **`m.owner` is cleared *after* the lock is released — a cross-thread acquirer's owner can be stomped with `VOID`.**

Once this release-store lands, a polling waiter on another thread can observe `locked == false`, acquire, and write its own `m.owner` — and this function's `m.owner = types.VOID` on the next line can then execute *after* that, leaving a locked mutex with `owner == VOID`. Consequences: `abandonFiberMutexes` will never abandon it if the new owner's fiber dies (`m.owner == fiber_val` won't match), so cross-thread waiters poll forever; and `mutex-state` misreports `not-owned`.

Clear `owner` **before** the release-store — the `.release` barrier then publishes it correctly to the next acquirer:

```zig
m.owner = types.VOID;
@atomicStore(bool, &m.locked, false, .release);
```

Same pattern in `abandonFiberMutexes` (lines 166–169): the `m.owner = types.VOID` there should also move above the `@atomicStore(..., false, .release)`.

src/primitives_srfi18.zig
```diff
@@ -256,11 +256,16 @@ fn threadStartFn(args: []const Value) PrimitiveError!Value {
     fiber.os_thread = std.Thread.spawn(.{}, threadEntryFn, .{
         fiber, gc.allocator, gc, vm,
     }) catch return PrimitiveError.OutOfMemory;
+    _ = @atomicRmw(usize, &live_child_threads, .Add, 1, .release);
```
baijum: **Increment before `spawn`, not after — the child can decrement first.**

If the spawned thread runs to completion before this line executes, the `defer` decrement in `threadEntryFn` fires first and the counter wraps (`0 → maxInt → 0`), or dips one below the true count when other children exist. Today that's unobservable only because the thread calling `thread-start!` can't simultaneously be in `pollStep` — but a **child** spawning a grandchild breaks that: while child C sits between `spawn` and this increment and the grandchild has already exited, the *main* thread's `crossThreadWaitPossible` can read 0 with C still alive, and give up a wait it should keep polling — resurfacing exactly the bug this PR fixes.

Incrementing before `spawn` (and decrementing on the spawn-failure path) closes the window and removes the usize wraparound entirely.

src/primitives_srfi18.zig
```diff
@@ -584,6 +589,75 @@ fn scheduleOrTimeout(sched: *fiber_mod.FiberScheduler, me: *fiber_mod.Fiber) ?us
     return null;
 }
 
+// 1ms, matching thread-join!'s existing OS-thread poll cadence.
+const CROSS_THREAD_POLL_NS: u64 = 1_000_000;
+
+// Number of thread-start!-spawned OS threads currently alive (incremented in
+// threadStartFn, decremented via defer in threadEntryFn on every exit path).
+// Lets scheduleOrPoll tell a real cross-OS-thread wait apart from a genuine
```
baijum: Nit: `scheduleOrPoll` doesn't exist — the function is `pollStep`. Same stale name in the comment at the top of `threadEntryFn` (line 266).

tests/scheme/srfi/srfi18-cross-thread-wait.scm
```diff
@@ -0,0 +1,123 @@
...
+  (let ((t0 (current-time)))
+    (mutex-unlock! bc-m bc-cv)
+    (set! bc-elapsed-2 (- (time->seconds (current-time)) (time->seconds t0)))))
+
+(let ((t1 (make-thread bc-waiter-thunk-1))
+      (t2 (make-thread bc-waiter-thunk-2)))
+  (thread-start! t1)
+  (thread-start! t2)
+  (thread-sleep! 0.2)
+  (condition-variable-broadcast! bc-cv)
```
baijum, line 126: **A missed broadcast wedges CI forever instead of failing.**

This broadcast is issued *without holding `bc-m`*, so nothing orders it after the waiters' generation snapshots. If a waiter thread is slow to start (> 0.2 s under CI load) and hasn't reached its `mutex-unlock! bc-m bc-cv` yet, the generation bump lands before its snapshot and that waiter then waits for a second broadcast that never comes. Since child threads never give up polling (`crossThreadWaitPossible` is unconditionally true for them), `thread-join!` never returns and the job hangs until the CI timeout rather than failing with a message.

Suggest bounding the wait and asserting on the result, e.g.:

```scheme
(mutex-unlock! bc-m bc-cv 10)  ; store the return value
```

and check it returned `#t` (plus an upper bound on the elapsed assertions, so a 10 s timeout can't masquerade as a pass). The signal test above is protocol-correct — the signaler locks `cv-m` first — so only this section has the race. Alternatively, have the waiters bump a shared ready-counter while holding `bc-m` and have the main thread wait for it, then lock `bc-m` around the broadcast.

baijum (commented):
## Review

The diagnosis and overall design are right: each OS thread owning an independent `FiberScheduler` means cross-thread `mutex-unlock!`/`condition-variable-signal!` can never wake a remote waiter, so polling the shared state is the correct mechanism, and the **generation-snapshot-before-unlock** is exactly the right lost-wakeup defense (and well documented at the snapshot site). Other things that check out:

- `allocConditionVariable` uses a struct-literal init, so the new `signal_generation: u64 = 0` default actually applies (no uninitialized-field footgun).
- Preserving the give-up path for a genuine local deadlock (terminate instead of hang) is a sensible trade-off, and `crossThreadWaitPossible` is a clean way to express it.
- The coverage-test rewrite to fibers matches the isolated-heap semantics, and the new regression test's elapsed-time-instead-of-flag rationale is sound and clearly explained.

### Findings (inline comments)

1. **Lock acquisition is still a non-atomic test-then-set** (`mutexLockFn` fast path + the blind store after `runSchedulerUntilMutex`): two OS threads can both acquire, which is the same silent corruption class this PR sets out to fix. Should be a `@cmpxchgStrong` retry. *(correctness, should fix before merge)*
2. **`m.owner` is cleared *after* the release-store in `mutexUnlockFn`** (and in `abandonFiberMutexes`): a cross-thread acquirer's owner write can be overwritten with `VOID`, which breaks abandonment and can leave waiters polling forever. One-line reorder. *(correctness, should fix before merge)*
3. **`live_child_threads` is incremented after `spawn`**: a fast-exiting child (or grandchild) can transiently drive the counter to 0/wraparound while threads are alive, letting the main thread wrongly give up a wait. Increment before spawn. *(narrow race, one-line fix)*
4. **The broadcast section of the new test can hang CI forever** on a missed signal (broadcast is issued without holding `bc-m`, waiters have no timeout, and child threads never give up polling). Suggest a generous timeout + result assertion. *(test robustness)*
5. Stale `scheduleOrPoll` name in two comments. *(nit)*

### Not anchored in the diff (pre-existing, worth a follow-up)

- `mutexStateFn` still reads `m.locked` non-atomically; with this PR the same field is now written atomically from other threads, so that read is a data race. Likewise `m.abandoned` in `mutexLockFn` is read before the acquire-load of `locked`, so an abandonment published by a dying thread can be missed by a cross-thread locker.
- Behavior note: a **child** thread that genuinely self-deadlocks now polls forever (`crossThreadWaitPossible` is unconditionally true for children) instead of the old give-up-and-terminate. That's defensible under SRFI-18 (indefinite blocking is conformant), but the main-vs-child asymmetry means the same deadlock terminates or hangs depending on which thread hits it — worth documenting.

### Verdict

**Not ready to merge yet.** Findings 1 and 2 are small, contained fixes but they sit in exactly the mechanism this PR makes load-bearing (cross-thread mutual exclusion), and finding 4 risks wedging CI rather than failing. With 1–4 addressed I'd expect this to be good to go — the approach itself is right and the test coverage strategy is solid. (CI was still pending at review time; worth confirming the Debug and riscv64 jobs, since the new test's 0.2 s windows are tighter on slow runners.)

src/primitives_srfi18.zig
baijum: Fixed in 105b868: replaced the load-then-store with `@cmpxchgStrong` (via new `tryClaimMutex`/`tryClaimAbandoned` helpers), used both at the fast path and the post-wait retry (looping back into `runSchedulerUntilMutex` on a lost race instead of assuming success). The abandoned-mutex flag got the same claim-and-clear treatment via `tryClaimAbandoned`.

src/primitives_srfi18.zig
baijum: Fixed in 105b868: `owner` (and `abandoned` in abandonFiberMutexes) is now published before the release-store of `locked`, so a cross-thread acquirer that wins the race is guaranteed to see the correct owner/abandoned state instead of having it stomped back to VOID.

src/primitives_srfi18.zig
baijum: Fixed in 105b868: the increment now happens before `std.Thread.spawn`, with a rollback decrement if spawn itself fails.

src/primitives_srfi18.zig
baijum: Fixed in 105b868 (both occurrences).

tests/scheme/srfi/srfi18-cross-thread-wait.scm
baijum: Fixed in 105b868: each waiter's `mutex-unlock! bc-m bc-cv` now takes a 5s timeout, and the test asserts the return value was `#t` (not a timeout), so a missed-signal race turns into a reported test failure instead of a hung CI job.

tests/scheme/coverage/srfi18-coverage.scm
```diff
@@ -149,40 +153,44 @@
...
-    (thread-start! t1)
-    (thread-start! t2)
-    (thread-sleep! 0.05)
+    (fiber-join t1)
     (condition-variable-broadcast! cv)
-    (thread-join! t1)
-    (thread-join! t2)
+    (fiber-join t1)
+    (fiber-join t2)
     (check "cv-broadcast wakes all" count 2)))
```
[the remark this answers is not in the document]
baijum: Good catch, and it goes further than the fiber-join sequencing: I verified empirically that even the fixed two-waiter version passed with `condition-variable-broadcast!` commented out entirely -- both fibers were resolving via the (separate, pre-existing) fiber-only give-up path once nothing was locally schedulable, not via the broadcast. Rewritten in 105b868 to a single waiter mirroring the already-correct "cv-signal wakes waiter" test's structure (main waits, the spawned fiber wakes it via a real local status transition before it ever hits that give-up path), just calling broadcast! instead of signal! for coverage of that specific procedure.

baijum:
Thanks both for the thorough review — pushed 105b868 addressing all of it:

**From the inline comments (replied on each thread):**
1. Non-atomic test-then-set lock acquisition → replaced with `@cmpxchgStrong` (`tryClaimMutex`/`tryClaimAbandoned`), with the post-wait path retrying on a lost race instead of assuming success.
2. `owner` cleared after the release-store in `mutex-unlock!`/`abandonFiberMutexes` → reordered so `owner` (and `abandoned`) publish before the release.
3. `live_child_threads` incremented after `spawn` → moved before, with rollback on spawn failure.
4. New regression test's broadcast section could hang CI on a missed signal (broadcast issued without holding `bc-m`) → added a bounded 5s timeout per waiter + assert the wait actually succeeded, so a missed signal now fails the test instead of hanging the job.
5. Stale `scheduleOrPoll` name in comments → fixed (function is `pollStep`).
6. Coverage test's rewritten broadcast case (two fibers parked on the same condvar with nothing else runnable) → verified empirically it passed even with the broadcast call deleted entirely, since it was resolving via the separate, pre-existing fiber-only give-up path, not the broadcast. Rewrote to a single waiter mirroring the already-correct signal test's structure.

**From the review body's "not anchored in diff" notes:**
- `mutex-state` read `m.locked`/`m.abandoned` non-atomically → made atomic, since they're genuinely cross-thread fields now.
- `m.abandoned` read-before-acquire-load ordering concern → moot with the CAS redesign; the abandoned check now happens via `tryClaimAbandoned` only after already winning the locked CAS.
- Child-thread self-deadlock now polls forever instead of terminating (asymmetric with the main thread) → added a comment documenting this tradeoff at `crossThreadWaitPossible`; per SRFI-18, indefinite blocking is conformant, so I left the behavior as-is rather than adding more machinery to detect it.

Full suite (`run-all.sh`: 1828/1828, `zig build test`, `zig fmt --check`) still green, and I re-verified the updated regression test against the true pre-fix baseline (checked out the commit before either of my two commits) — it fails cleanly (no hang) without the fix and passes reliably with it.

src/primitives_srfi18.zig
```diff
@@ -691,38 +745,65 @@ fn mutexLockFn(args: []const Value) PrimitiveError!Value {
...
+    while (true) {
+        const done = try fiber_mod.runSchedulerStep(MutexWait, .{ .m = m }, ctx.vm, ctx.sched, me);
+        if (me.timed_out) {
+            me.timed_out = false;
+            me.deadline_ns = null;
+            return types.FALSE;
+        }
+        if (done) {
+            if (tryClaimMutex(m)) break;
+            me.status = .waiting;
```
baijum, line 778: **A timed lock can silently lose its timeout across this re-arm: local wakes cancel the reactor timer, and nothing re-registers it.**

`wakeMutexWaiters` (fiber.zig:364) calls `cancelPendingTimer` when it wakes this fiber. So if a *local* unlock wakes `me` and another OS thread steals the lock in the gap, the loop re-arms with `me.status = .waiting` while the reactor no longer holds the timer (both re-arm sites are reachable in that state: this one via `done && !tryClaimMutex`, and the post-`sleepNs` one via the woken-but-still-locked `next_idx == my_idx` break, which lands in the `!done` branch). Consequences:

- `me.deadline_ns` is still set, so `hasRunnableFibers()` keeps counting `me` — but nothing will ever fire the timeout. `(mutex-lock! m 2.0)` then blocks until the mutex is actually released, arbitrarily past its deadline.
- Worst corner: if the stealing thread later exits **without unlocking**, `live_child_threads` reaches 0, the next `runSchedulerStep` entry computes `pollCapNs = null`, and with the timer heap empty `parkOnReactor` → `poll(null)` → `backend.wait(null)` blocks unboundedly — not even the deadlock-error path is reachable.

The condvar loop doesn't have this hole (any local wake flips `status`, making `isDone` true, so its re-arm only runs with the timer intact) — it's specific to the mutex loop. Suggested fix, applied at both re-arm sites:

```zig
me.status = .waiting;
if (deadline) |d| {
    ctx.reactor.removeTimer(me); // no-op if still pending
    try ctx.reactor.addTimer(d, me);
}
```

`removeTimer`-first keeps it idempotent (no duplicate heap entry that could fire against a later wait), and re-adding the original *absolute* deadline is correct — if it has already passed, `popExpiredTimers` fires it on the next tick and the normal `timed_out` path returns `#f`.

baijum (commented):
## Re-review (force-push 49bac9a7: review fixes + rebase onto KEP-0001 Phase 2)

All five findings from my previous review are addressed, and I verified each in the code rather than just the description:

1. **CAS acquisition** — `tryClaimMutex`/`tryClaimAbandoned` with a retry loop that goes back to waiting on a lost claim instead of blind-storing. The re-arm (`me.status = .waiting`) is sound: `wakeMutexWaiters`/`wakeOneCondVarWaiter` never clear `waiting_on`, so a re-armed fiber stays discoverable for future local wakes.
2. **Publish-before-release ordering** — `owner` cleared before the release-store in `mutexUnlockFn`, and `abandoned`/`owner` published before the release-store in `abandonFiberMutexes`, both with comments explaining why.
3. **Counter increment before `spawn`** with a decrement on the spawn-failure path.
4. **Broadcast test hardened** — bounded 5 s timeout, result captured and asserted (`bc-woke-N`), race documented; a missed signal now fails loudly instead of wedging CI.
5. Stale comment name fixed — plus the follow-ups I'd only noted for later: `mutexStateFn` now uses atomic loads, and the child-vs-main deadlock asymmetry is documented in `crossThreadWaitPossible`'s comment.

The rebase integration is also careful where it needed to be:

- `pollCapNs` is narrowly scoped via `@hasDecl` — every other `runSchedulerStep` caller (channel-receive, fiber-join, thread-join!, thread-sleep!) compiles to exactly the old `null` behavior.
- `reactor.poll`'s `effectiveTimeout` takes `min(cap, nearest timer)` and `popExpiredTimers` only reports timers actually past deadline, so a capped 1 ms wakeup returns an empty ready list and **cannot fake a timeout** (important, since `wakeReadyFiber` sets `timed_out=true` on any `.waiting` fiber it's handed).
- `runSchedulerStep` captures `ctx.isDone()` *before* forcing `me.status = .running` — subtle, correct, and commented (CondVarWait's `isDone` reads `me.status`).
- The condvar retry loop is immune to the timer issue below because any local wake flips `status` and makes `isDone` true — only the mutex loop's re-arms can run after a wake.
- The coverage-test broadcast rework (single deterministic waiter, with the explanation of why two fiber-level waiters can't be tested) is sound.

### One new finding (inline)

The mutex retry loop can **silently lose its timeout timer** after a local wake + cross-thread steal: local wakes cancel the reactor timer and the re-arm never re-registers it, so a timed `mutex-lock!` blocks past its deadline — and in the worst corner (the stealing thread exits without unlocking) ends in an unbounded `poll(null)` block. Small, contained fix at the two re-arm sites; details and suggested patch inline.

### Follow-up (not blocking, pre-existing scope)

A child OS thread that dies **while holding a parent-heap mutex** never abandons it: `abandonFiberMutexes` walks the dying fiber's own GC heap, and the shared mutex lives in the parent heap. Waiters then get the deadlock error (main) or poll indefinitely (other children), instead of SRFI-18's abandoned-mutex behavior — and `m.owner` dangles once the child heap is freed. Cross-thread sharing only became real with this PR, so this is fine as a filed follow-up rather than scope creep here.

### Verdict

This is very close. **Fix the lost-timer re-arm (one inline finding, ~4 lines) and it's ready to merge once CI is green** — the design is right, the previous round's races are genuinely closed, and the test strategy (bounded waits, elapsed-time assertions, pre-fix baseline verification in a separate worktree) is exactly what this change needed. CI was still pending at review time (format + wasm pass); the Debug job is worth watching since the test's 0.2 s windows are tighter on slow runners.

src/primitives_srfi18.zig
baijum: Fixed in 5b85e13: both re-arm sites now do remove-then-add on the timer when a deadline was given, so a local wake's `cancelPendingTimer` can never leave a timed `mutex-lock!` without a live timeout after a lost claim race. Verified with a repro (real OS thread, timed lock on a mutex that's never released again) that it now times out at the given duration instead of blocking indefinitely.

src/primitives_srfi18.zig
```diff
@@ -632,38 +681,43 @@ fn mutexStateFn(args: []const Value) PrimitiveError!Value {
...
-        else if (args.len > 2 and args[2] == types.FALSE)
-            types.VOID
-        else if (ctx.vm.current_fiber) |cf|
-            types.makePointer(@ptrCast(&cf.header))
-        else
-            types.VOID;
+        m.owner = resolveMutexOwner(args, if (ctx.vm.current_fiber) |cf| types.makePointer(@ptrCast(&cf.header)) else types.VOID);
+        if (tryClaimAbandoned(m)) return raiseError(.abandoned_mutex, "mutex was abandoned", types.VOID);
         return types.TRUE;
     }
```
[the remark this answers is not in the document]
baijum: Good catch, thanks — fixed in 5b85e13, both sites.

src/primitives_srfi18.zig
```diff
@@ -691,38 +745,65 @@ fn mutexLockFn(args: []const Value) PrimitiveError!Value {
...
+            continue;
+        }
+        if (!crossThreadWaitPossible()) {
+            if (deadline != null) ctx.reactor.removeTimer(me);
+            me.deadline_ns = null;
+            return raiseError(.general, "mutex-lock!: deadlock — mutex will never be released (all fibers blocked)", types.VOID);
+        }
+        sleepNs(CROSS_THREAD_POLL_NS);
+        me.status = .waiting;
     }
```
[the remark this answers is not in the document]
baijum: Fixed in 5b85e13, same patch as suggested (remove-then-add at this site and the sleepNs re-arm site below it).

tests/scheme/srfi/srfi18-cross-thread-wait.scm
```diff
@@ -0,0 +1,138 @@
...
+        (set! fail (+ fail 1))
+        (display "FAIL: ") (display name)
+        (display " expected ") (write expected)
+        (display " got ") (write got)
+        (newline))))
+
+(define (check-true name val)
+  (if val
+      (set! pass (+ pass 1))
+      (begin (set! fail (+ fail 1)) (display "FAIL: ") (display name) (newline))))
```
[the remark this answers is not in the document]
baijum: Fair point, but I'll leave this as-is for now: the srfi18-*.scm family in this directory (srfi18-abandoned-mutex.scm, srfi18-dynamic-wind.scm, srfi18-globals-race.scm, etc.) all use this same hand-rolled check/check-true convention, so matching it keeps this file consistent with its immediate siblings rather than being the odd one out. Happy to migrate the whole family to SRFI-64 in a follow-up if that's preferred, rather than making this one new file inconsistent with the others it sits next to.

baijum:
Thanks for the re-review and for verifying each of the previous fixes in the code rather than just the description — that's a much stronger signal than I could give myself.

Pushed 5b85e13 for the one new finding plus CodeRabbit's write-barrier catch:

1. **Lost timeout timer on the mutex retry loop** — fixed at both re-arm sites (the lost-claim-race branch and the cross-thread-poll branch) with remove-then-add on the timer whenever a deadline was given, exactly as suggested. Verified with a targeted repro: a timed `mutex-lock!` on a mutex that's never released again now times out at the given duration instead of blocking indefinitely.
2. **Missing GC write barrier after storing a fiber pointer into `m.owner`** (CodeRabbit, Critical) — real bug, fixed at both owner-assignment sites, matching the pattern used elsewhere in this file.
3. SRFI-64 vs. hand-rolled counters for the new test file — replied inline; I'd rather keep it consistent with its immediate `srfi18-*.scm` siblings (all hand-rolled) than make it the odd one out, but happy to migrate the whole family in a follow-up if preferred.

Filed #1458 for the "child dies while holding a parent-heap mutex is never abandoned" follow-up rather than folding it into this PR, per your note that it's pre-existing scope this PR just made reachable.

Full suite (1833/1833), `zig build test`, and `zig fmt --check` all still green on top of the fix. CI is (re-)running on 5b85e13 now.

baijum (commented):
## Round 3: verified 5b85e133 — all findings addressed, LGTM

**The lost-timer fix is correct at both re-arm sites.** Remove-then-add keeps the timer heap free of duplicates, re-adding the original *absolute* deadline preserves the timeout contract, and a re-add of an already-passed deadline is handled too — `popExpiredTimers` fires it on the next tick and the normal `timed_out` path returns `#f`. This also closes the worst corner (stealing thread exits without unlocking → `poll(null)` blocking unboundedly with an empty timer heap), since the timer now always bounds the reactor wait.

**The bonus `writeBarrier` additions on the `m.owner` stores are also right, including cross-thread.** I checked the generational machinery for the case where a *child* thread locks a parent-heap mutex (foreign container, `gc_instance` = child GC):

- `writeBarrier` owner-checks `new_val`, and the child-heap fiber being stored is exactly the referent the child's remembered set should track.
- If the foreign mutex lands in the child's remembered set, `minorCollect` only calls `markObjectContents` (reads the container's fields, never writes its header) and `referencesYoung`/`isYoungPointer` owner-check every referent — so no cross-heap mark-bit writes; the #958 invariant holds.
- The one scenario where the entry could dangle (parent drops its only reference while the child still uses the mutex) is the pre-existing cross-heap lifetime limitation that already crashes via the polling loop itself — KEP-0002 territory, not a regression here.

### Verdict

All findings from all three review rounds are now fixed and verified in code: CAS acquisition + claim-retry, publish-before-release ordering, counter increment order, broadcast-test hardening, comment fixes, `mutexStateFn` atomics, documented deadlock asymmetry, and now the timer re-arm and owner write barriers. **This PR is ready to merge once CI completes green** (pending at review time; format + wasm already pass). The cross-heap mutex-abandonment gap noted last round remains a good follow-up issue to file, not a blocker.

PR #1457, opened 2026-07-11
Add configurable REPL syntax highlighting with dark/light presets

### LINKED ISSUE #1456: REPL: configurable syntax highlighting with dark/light theme presets
## Problem

The REPL syntax highlighting uses hardcoded ANSI color constants that cannot be customized. Several token types are not highlighted at all. There is no `NO_COLOR` support, and the default colors have poor contrast on light terminal backgrounds (yellow numbers, gray comments/parens are near-invisible on white).

## Solution

1. **General config file** (`~/.kaappi/config`) with `key: value` format for REPL preferences (theme, prompt, history length, highlight toggle).

2. **Dark and light theme presets** (`repl.theme: dark` / `repl.theme: light`) with colors optimized for each background. Individual `repl.color.*` keys override the preset.

3. **Full R7RS token coverage** — highlight `#true`/`#false`, `#(` vectors, `#u8(` bytevectors, `,@` unquote-splicing, `#b`/`#o`/`#x` number prefixes, `+inf.0`/`-inf.0`/`+nan.0`/`-nan.0`, `#;` datum comments, `#!` directives, `|...|` pipe-quoted symbols. Add 12 missing R7RS syntax keywords.

4. **`NO_COLOR` env var support** per https://no-color.org/.

5. **Colored prompts** — primary and continuation prompts get distinct colors (configurable via `repl.color.prompt` / `repl.color.continuation`).

## Config format

```
# Select preset: dark (default), light
repl.theme: dark

# Override individual colors (optional)
repl.color.keyword: bright-magenta
repl.color.string: bright-green
repl.color.number: bright-yellow
repl.color.comment: bright-black
repl.color.boolean: bright-cyan
repl.color.paren: bright-black
repl.color.match-paren: bold bright-yellow
repl.color.prompt: bright-green
repl.color.continuation: bright-black

# REPL preferences
repl.prompt: kaappi> 
repl.history-length: 1000
repl.highlight: true
```

Accepted color names: `black`, `red`, `green`, `yellow`, `blue`, `magenta`, `cyan`, `white`, bright variants (`bright-black`, etc.), optional `bold` prefix, and `none` to disable.

Precedence: `NO_COLOR` env var > `repl.color.*` > `repl.theme` > default (dark).

baijum:
## Summary

Closes #1456

- Introduce `~/.kaappi/config` as a general-purpose config file (`key: value` format)
- Add `dark` and `light` theme presets optimized for terminal background contrast
- Support `NO_COLOR` env var per https://no-color.org/
- Extend REPL highlighter for full R7RS token coverage
- Add colored prompts (primary and continuation)

## Details

**New file `src/config.zig`** (363 lines):
- `Theme` struct with dark/light/no_color presets
- `Config` struct (theme, prompt, history-length, highlight toggle)
- Config parser: `repl.theme`, `repl.color.*`, `repl.prompt`, `repl.history-length`, `repl.highlight`
- 16 color names (8 standard + 8 bright) with optional `bold` prefix and `none`
- 20 unit tests

**Modified `src/repl.zig`**:
- Highlighter uses theme struct instead of hardcoded ANSI constants
- Full R7RS token coverage: `#true`/`#false`, `#(`, `#u8(`, `,@`, `#b`/`#o`/`#x`/`#d`/`#e`/`#i` number prefixes, infnan, `#;` datum comments, `#!` directives, `|...|` pipe-quoted symbols
- 41 R7RS syntax keywords (was 29)
- Colored primary and continuation prompts
- `NO_COLOR` and `repl.highlight: false` disable highlighting

**Dark preset** (default): bright-magenta keywords, bright-green strings, bright-yellow numbers, bright-cyan booleans, bright-green prompt
**Light preset**: standard magenta keywords, standard green strings, red numbers, standard cyan booleans, blue prompt

## Test plan

- [x] `zig build test` — 762 unit tests pass (including 20 new config tests)
- [x] `bash tests/scheme/run-all.sh` — 1827 Scheme tests pass, 0 fail
- [ ] Manual: REPL with no config file → dark preset colors
- [ ] Manual: `repl.theme: light` in config → light preset
- [ ] Manual: `NO_COLOR=1` → no color output
- [ ] Manual: invalid config values → stderr warnings, REPL works

### DIFF
--- a/src/config.zig
+++ b/src/config.zig
@@ -42,10 +42,12 @@ pub const Theme = struct {
     };
 };
 
+const default_prompt = "kaappi> ";
+
 pub const Config = struct {
     theme: Theme = .{},
-    prompt_buf: [64:0]u8 = initPromptBuf("kaappi> "),
-    prompt_len: u8 = 8,
+    prompt_buf: [64:0]u8 = initPromptBuf(default_prompt),
+    prompt_len: u8 = default_prompt.len,
     history_length: c_int = 1000,
     highlight: bool = true,
 
@@ -62,7 +64,8 @@ pub const Config = struct {
 
 pub fn load() Config {
     var cfg: Config = .{};
-    const no_color = std.c.getenv("NO_COLOR") != null;
+    const nc = std.c.getenv("NO_COLOR");
+    const no_color = nc != null and std.mem.sliceTo(nc.?, 0).len > 0;
     if (no_color) {
         cfg.theme = Theme.no_color;
         cfg.highlight = false;
@@ -75,6 +78,24 @@ pub fn load() Config {
     const data = file_utils.readWholeFile(std.heap.c_allocator, path, 64 * 1024) catch return cfg;
     defer std.heap.c_allocator.free(data);
 
+    // Pass 1: apply repl.theme preset (so repl.color.* overrides take precedence)
+    if (!no_color) {
+        var it1 = std.mem.splitScalar(u8, data, '\n');
+        while (it1.next()) |line| {
+            const stripped = std.mem.trimEnd(u8, line, "\r\n");
+            const trimmed = std.mem.trim(u8, stripped, " \t");
+            if (std.mem.startsWith(u8, trimmed, "repl.theme:")) {
+                const val = std.mem.trim(u8, trimmed["repl.theme:".len..], " \t");
+                if (std.mem.eql(u8, val, "dark")) {
+                    cfg.theme = Theme.dark;
+                } else if (std.mem.eql(u8, val, "light")) {
+                    cfg.theme = Theme.light;
+                }
+            }
+        }
+    }
+
+    // Pass 2: apply all keys (repl.theme is a no-op here)
     var line_num: usize = 0;
     var it = std.mem.splitScalar(u8, data, '\n');
     while (it.next()) |line| {
@@ -99,35 +120,40 @@ fn parseLine(cfg: *Config, line: []const u8, line_num: usize, no_color: bool) vo
     const value = std.mem.trimEnd(u8, raw_value, " \t");
 
     if (std.mem.eql(u8, key, "repl.theme")) {
-        if (std.mem.eql(u8, value, "dark")) {
-            if (!no_color) cfg.theme = Theme.dark;
-        } else if (std.mem.eql(u8, value, "light")) {
-            if (!no_color) cfg.theme = Theme.light;
-        } else {
+        if (!std.mem.eql(u8, value, "dark") and !std.mem.eql(u8, value, "light")) {
             warnLine(line_num, "repl.theme must be 'dark' or 'light'");
         }
         return;
     } else if (std.mem.eql(u8, key, "repl.color.keyword")) {
-        if (!no_color) cfg.theme.keyword = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        const ansi = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        if (!no_color) cfg.theme.keyword = ansi;
     } else if (std.mem.eql(u8, key, "repl.color.string")) {
-        if (!no_color) cfg.theme.string = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        const ansi = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        if (!no_color) cfg.theme.string = ansi;
     } else if (std.mem.eql(u8, key, "repl.color.number")) {
-        if (!no_color) cfg.theme.number = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        const ansi = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        if (!no_color) cfg.theme.number = ansi;
     } else if (std.mem.eql(u8, key, "repl.color.comment")) {
-        if (!no_color) cfg.theme.comment = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        const ansi = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        if (!no_color) cfg.theme.comment = ansi;
     } else if (std.mem.eql(u8, key, "repl.color.boolean")) {
-        if (!no_color) cfg.theme.boolean = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        const ansi = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        if (!no_color) cfg.theme.boolean = ansi;
     } else if (std.mem.eql(u8, key, "repl.color.paren")) {
-        if (!no_color) cfg.theme.paren = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        const ansi = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        if (!no_color) cfg.theme.paren = ansi;
     } else if (std.mem.eql(u8, key, "repl.color.match-paren")) {
-        if (!no_color) cfg.theme.match_paren = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        const ansi = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        if (!no_color) cfg.theme.match_paren = ansi;
     } else if (std.mem.eql(u8, key, "repl.color.prompt")) {
-        if (!no_color) cfg.theme.prompt = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        const ansi = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        if (!no_color) cfg.theme.prompt = ansi;
     } else if (std.mem.eql(u8, key, "repl.color.continuation")) {
-        if (!no_color) cfg.theme.continuation = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        const ansi = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+        if (!no_color) cfg.theme.continuation = ansi;
     } else if (std.mem.eql(u8, key, "repl.prompt")) {
         if (raw_value.len > 63) {
-            warnLine(line_num, "prompt too long (max 63 chars)");
+            warnLine(line_num, "prompt too long (max 63 bytes)");
             return;
         }
         @memcpy(cfg.prompt_buf[0..raw_value.len], raw_value);
@@ -138,8 +164,8 @@ fn parseLine(cfg: *Config, line: []const u8, line_num: usize, no_color: bool) vo
             warnLine(line_num, "invalid number for repl.history-length");
             return;
         };
-        if (cfg.history_length < 0) {
-            warnLine(line_num, "repl.history-length must be non-negative");
+        if (cfg.history_length < 1) {
+            warnLine(line_num, "repl.history-length must be >= 1");
             cfg.history_length = 1000;
         }
     } else if (std.mem.eql(u8, key, "repl.highlight")) {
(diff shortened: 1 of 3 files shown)


src/config.zig
```diff
@@ -0,0 +1,363 @@
...
+
+    const colon = std.mem.indexOfScalar(u8, trimmed, ':') orelse {
+        warnLine(line_num, "missing ':'");
+        return;
+    };
+    const key = std.mem.trimEnd(u8, trimmed[0..colon], " \t");
+    const raw_value = std.mem.trimStart(u8, trimmed[colon + 1 ..], " \t");
+    const value = std.mem.trimEnd(u8, raw_value, " \t");
+
+    if (std.mem.eql(u8, key, "repl.theme")) {
```
baijum, line 125: Issue #1456 specifies precedence `repl.color.*` > `repl.theme`, but this implementation is file-order-dependent: a `repl.theme:` line appearing *after* any `repl.color.*` lines replaces the whole `Theme` struct and silently discards the earlier overrides. (Same finding as CodeRabbit's — confirmed.)

Cleanest fix is two passes in `load()`: apply `repl.theme` lines first, then everything else. That preserves the single-assignment simplicity of `parseLine` and makes the config order-insensitive, matching the issue's stated contract.

src/config.zig
```diff
@@ -0,0 +1,363 @@
...
+        }
+        @memcpy(cfg.prompt_buf[0..raw_value.len], raw_value);
+        cfg.prompt_buf[raw_value.len] = 0;
+        cfg.prompt_len = @intCast(raw_value.len);
+    } else if (std.mem.eql(u8, key, "repl.history-length")) {
+        cfg.history_length = std.fmt.parseInt(c_int, value, 10) catch {
+            warnLine(line_num, "invalid number for repl.history-length");
+            return;
+        };
+        if (cfg.history_length < 0) {
```
baijum: `repl.history-length: 0` passes this validation, but `linenoiseHistorySetMaxLen` rejects `len < 1` (returns 0 without changing anything), so the call in `repl()` becomes a no-op and linenoise's built-in default stays in effect. The user asked for no history and silently got the default instead.

Either warn and require `>= 1` here, or explicitly treat 0 as "disable history".

src/config.zig
```diff
@@ -0,0 +1,363 @@
...
+    fn initPromptBuf(comptime default: []const u8) [64:0]u8 {
+        var buf: [64:0]u8 = @splat(0);
+        @memcpy(buf[0..default.len], default);
+        return buf;
+    }
+};
+
+pub fn load() Config {
+    var cfg: Config = .{};
+    const no_color = std.c.getenv("NO_COLOR") != null;
```
baijum: Per no-color.org, color should be disabled only "when present **and not an empty string**". `getenv(...) != null` treats `NO_COLOR=""` as set.

```zig
const nc = std.c.getenv("NO_COLOR");
const no_color = nc != null and nc.?[0] != 0;
```

src/config.zig
```diff
@@ -0,0 +1,363 @@
...
+        if (std.mem.eql(u8, value, "dark")) {
+            if (!no_color) cfg.theme = Theme.dark;
+        } else if (std.mem.eql(u8, value, "light")) {
+            if (!no_color) cfg.theme = Theme.light;
+        } else {
+            warnLine(line_num, "repl.theme must be 'dark' or 'light'");
+        }
+        return;
+    } else if (std.mem.eql(u8, key, "repl.color.keyword")) {
+        if (!no_color) cfg.theme.keyword = colorToAnsi(value) orelse return warnColor(line_num, key, value);
```
baijum: When `NO_COLOR` is set, `colorToAnsi` never runs on these branches, so invalid color names are silently accepted — verified in a pty run: the `unknown color 'purrple'` warning printed without `NO_COLOR` disappears with `NO_COLOR=1`. Config typos should surface regardless of environment; validate unconditionally and only gate the assignment:

```zig
const ansi = colorToAnsi(value) orelse return warnColor(line_num, key, value);
if (!no_color) cfg.theme.keyword = ansi;
```

src/config.zig
```diff
@@ -0,0 +1,363 @@
...
+        if (!no_color) cfg.theme.paren = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+    } else if (std.mem.eql(u8, key, "repl.color.match-paren")) {
+        if (!no_color) cfg.theme.match_paren = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+    } else if (std.mem.eql(u8, key, "repl.color.prompt")) {
+        if (!no_color) cfg.theme.prompt = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+    } else if (std.mem.eql(u8, key, "repl.color.continuation")) {
+        if (!no_color) cfg.theme.continuation = colorToAnsi(value) orelse return warnColor(line_num, key, value);
+    } else if (std.mem.eql(u8, key, "repl.prompt")) {
+        if (raw_value.len > 63) {
+            warnLine(line_num, "prompt too long (max 63 chars)");
```
baijum: Nit: `raw_value.len` counts bytes, not characters — a 22-emoji prompt gets rejected while the message says "max 63 chars". Saying "bytes" in the message avoids confusion (UTF-8 prompts otherwise work; verified `λ> ` renders fine).

src/config.zig
```diff
@@ -0,0 +1,363 @@
...
+        .prompt = "",
+        .continuation = "",
+        .reset = "",
+    };
+};
+
+pub const Config = struct {
+    theme: Theme = .{},
+    prompt_buf: [64:0]u8 = initPromptBuf("kaappi> "),
+    prompt_len: u8 = 8,
```
baijum: Nit: `prompt_len = 8` must be kept in sync with `initPromptBuf("kaappi> ")` by hand. A shared constant removes the trap:

```zig
const default_prompt = "kaappi> ";
prompt_buf: [64:0]u8 = initPromptBuf(default_prompt),
prompt_len: u8 = default_prompt.len,
```

src/repl.zig
```diff
@@ -314,46 +365,104 @@ fn highlightCallback(buf: [*c]const u8, len: usize, pos: usize, out_len: [*c]usi
...
                 }
             }
-            result.appendSlice(allocator, ANSI_RESET) catch return null;
+            result.appendSlice(allocator, theme.reset) catch return null;
             continue;
         }
 
+        // #true / #false (R7RS extended boolean)
+        if (ch == '#' and i + 4 < input.len) {
+            if (std.mem.startsWith(u8, input[i..], "#true") and (i + 5 >= input.len or isDelimiter(input[i + 5]))) {
```
baijum, line 374: None of the new highlighter branches (`#true`/`#false`, `#(`, `#u8(`, `#b`/`#o`/`#x`/`#d`/`#e`/`#i` prefixes, `,@`, `|...|`, `#;`, `#!`, infnan in `isNumberLike`) have tests, and this file already has `highlightCallback` test precedent at the bottom. The `#`-dispatch chain is ordering-sensitive (e.g. `#true` must be checked before `#t`, `#|` before `#;`) — exactly the kind of code that silently breaks when someone inserts a new branch. A handful of assertions like "`#u8(` gets `theme.paren`", "`#true` gets `theme.boolean`, `#trueish` doesn't" would lock this in cheaply.

baijum (commented):
## Review

Solid feature: clean value-type `Config`/`Theme` design (all color strings are static literals, so no allocation or lifetime concerns), config loading only happens in `repl()` so script startup is unaffected, and the 20 config unit tests are well targeted.

### Verified locally (head `0ed92c2d`, macOS ARM)

- `zig build test` passes.
- Drove the REPL through a pty with `KAAPPI_HOME` pointing at a scratch config: light preset + custom UTF-8 prompt `λ> ` (trailing space preserved) renders correctly, `repl.color.number: bold blue` overrides the preset, match-paren highlighting works, and malformed lines (unknown key, bad color, missing colon, negative history) each produce a stderr warning with the right line number.
- `NO_COLOR=1` run produces **zero** SGR sequences in the output.
- Checked the vendored linenoise: `utf8StrWidth` skips ANSI CSI sequences, so ANSI codes in the prompt don't corrupt cursor/wrap math in multiline mode. That was my main structural worry with colored prompts and it's already handled.

### Findings (inline comments)

1. **Theme/color precedence is file-order-dependent** — contradicts the precedence contract stated in #1456 (`repl.color.*` > `repl.theme`). Same issue CodeRabbit flagged; I confirmed it. This is the one thing I'd fix before merge.
2. **`repl.history-length: 0` silently no-ops** — passes validation but linenoise rejects `< 1`, leaving its built-in default in effect.
3. **`NO_COLOR=""` disables color** — no-color.org says only when present *and non-empty*.
4. **Color-name validation is skipped under `NO_COLOR`** — typos in `repl.color.*` are silently accepted when the env var is set (verified in the pty run: the `purrple` warning disappears under `NO_COLOR=1`).
5. **No tests for the new highlighter branches** (`#true`/`#false`, `#(`, `#u8(`, radix/exactness prefixes, `,@`, `|...|`, `#;`, `#!`, infnan) despite existing `highlightCallback` test precedent in the file.
6. Two nits inline (bytes-vs-chars message, `prompt_len` kept in sync by hand).

### Non-blocking notes

- The unknown-key warning cuts against the "general-purpose config file" framing: once other tools/versions add keys (e.g. `thottam.*`), older binaries will warn on every REPL start. Consider only warning for unrecognized `repl.*` keys.
- Docs follow-up needed in kaappi.github.io (config file format, color names, `NO_COLOR`).
- Cosmetic highlighter gaps that are fine to leave: `.5` isn't number-colored; `#;` colors only the marker, not the following datum.

### Merge verdict

**Not quite ready.** Fix the precedence issue (finding 1 — it contradicts the issue's own spec) and ideally the `history-length: 0` no-op, and add a few highlighter token tests. Everything else is nit-level. The CI test matrix was still running at review time (wasm/format green). After those fixes + green CI, this is good to merge.

src/config.zig
```diff
@@ -0,0 +1,403 @@
...
+    const data = file_utils.readWholeFile(std.heap.c_allocator, path, 64 * 1024) catch return cfg;
+    defer std.heap.c_allocator.free(data);
+
+    // Pass 1: apply repl.theme preset (so repl.color.* overrides take precedence)
+    if (!no_color) {
+        var it1 = std.mem.splitScalar(u8, data, '\n');
+        while (it1.next()) |line| {
+            const stripped = std.mem.trimEnd(u8, line, "\r\n");
+            const trimmed = std.mem.trim(u8, stripped, " \t");
+            if (std.mem.startsWith(u8, trimmed, "repl.theme:")) {
```
baijum: `repl.theme : light` (space before the colon) is silently ignored: pass 2's `parseLine` extracts the key with `trimEnd(trimmed[0..colon], " \t")`, so it recognizes the key, sees a valid value, and prints **no warning** — but this `startsWith` never matches, so the preset is never applied. Verified in a pty run: config containing only `repl.theme : light` produces no warning and the prompt stays dark bright-green.

Every other key tolerates `key : value` spacing, so this is both inconsistent and silent. Suggest matching pass 2's key extraction:

```zig
const colon = std.mem.indexOfScalar(u8, trimmed, ':') orelse continue;
const key = std.mem.trimEnd(u8, trimmed[0..colon], " \t");
if (std.mem.eql(u8, key, "repl.theme")) {
    const val = std.mem.trim(u8, trimmed[colon + 1 ..], " \t");
    ...
}
```

Relatedly: the new test `theme preset via two-pass loading` assigns `cfg.theme = Theme.light` directly, so it never exercises this pass-1 parsing — extracting the theme-line match into a small helper would make it unit-testable and would have caught this.

baijum (commented):
## Follow-up review of `8a1f0525`

All six findings from my previous review are addressed, and I re-verified each behaviorally at the new head (pty-driven REPL runs with a scratch `KAAPPI_HOME`, plus the full unit suite locally):

| Finding | Status |
|---|---|
| 1. Theme/color precedence order-dependent | ✅ Fixed via two-pass `load()`. Verified: `repl.color.number: bold blue` **before** `repl.theme: light` → number renders `1;34m`, never light-red; light preset applies to untouched fields. |
| 2. `repl.history-length: 0` silent no-op | ✅ Now warns `must be >= 1` and falls back to 1000. Verified via pty + new unit test. |
| 3. `NO_COLOR=""` disabled color | ✅ Empty value now keeps color on. Verified: SGR sequences present with `NO_COLOR=""`. |
| 4. Validation skipped under `NO_COLOR` | ✅ Validates unconditionally, gates only the assignment. Test added. |
| 5. No highlighter-branch tests | ✅ 10 tests + 2 helpers covering all new token branches, including the ordering-sensitive negatives (`#trueish`, `\|define\|`). |
| 6. Bytes-vs-chars message, `prompt_len` sync | ✅ Both fixed (`max 63 bytes`, shared `default_prompt` constant). |

`zig build test` passes (773/773, exit 0) and CI is green at this head except the slow Debug job, which was still running at review time.

### Two new minor items (found while re-verifying)

1. **`repl.theme : light` (space before colon) is silently ignored** — inline comment with details. Small real bug in the new pass-1 code; 3-line fix. I'd fix this before merge since the failure is silent.
2. **`zig build test` now prints alarming noise on green runs** — the two new warning-path tests write to stderr, so every run shows `config: unknown color 'purrple' …` / `config: repl.history-length must be >= 1` followed by the Zig build runner's `failed command:` line, even though the suite passes (exit 0, `773/773 tests passed`). Contributors will read that as a failure. Cheapest fix: early-return in the three `warn*` functions when `@import("builtin").is_test` — the tests assert the resulting config state, not the message text, so nothing is lost.

### Merge verdict

**Ready to merge once the Debug CI job finishes green** — item 1 above is the only thing I'd fix first (tiny, and the failure mode is silent), and item 2 is worth folding into the same push to keep `zig build test` output clean. Neither invalidates the feature; if you'd rather ship as-is and follow up, that's defensible too.

baijum (commented):
## Follow-up review of `091f6b80`

Both remaining items are fixed and verified:

1. **Pass-1 theme detection** now uses the same key extraction as `parseLine` (find colon, trim key, compare). Verified via pty: a config containing `repl.theme : light` (space before colon) now applies the light preset — prompt renders `\x1b[34m` blue. Comment/blank-line skipping and value trimming also match pass 2, so the two parsers can no longer disagree.
2. **Test noise** — the `builtin.is_test` early-returns in the three `warn*` functions make `zig build test` output completely clean (zero lines, exit 0, 773/773). Verified the guard doesn't leak into real builds: a normal REPL run still prints `config: unknown key 'bogus-key' (line 2)`.

CI is fully green at this head, including the Debug job, and merge state is CLEAN.

**Ready to merge.** No auto-merge is armed, so it needs a manual merge. Only follow-up (non-blocking, separate repo): document the config file format, color names, and `NO_COLOR` support on kaappi-lang.org.

PR #1461, opened 2026-07-12
WASI backend: poll_oneoff reactor (KEP-0001 P4)

### LINKED ISSUE #1442: WASI backend: poll_oneoff reactor (KEP-0001 P4)
Part of #1438 (KEP-0001, Phase 4). Depends on Phase 3 (#1441). Design: [KEP-0001 §2](https://github.com/kaappi/keps/blob/main/keps/0001-event-loop-reactor.md#2-backends).

## Scope

- `WasiPollBackend` behind the existing `is_wasm` flag: build a `subscription_t[]` — one `fd_read`/`fd_write` per registration plus one `CLOCK` subscription (ABSTIME) at the nearest timer deadline — call `std.os.wasi.poll_oneoff`, match returned `event_t.userdata` back to registrations (the `mio` wasi model; maps one-to-one onto the reactor interface).
- Socket readiness is best-effort (host-dependent); degrade to single-fiber blocking I/O where the host lacks socket subscriptions.
- Timers and `thread-sleep!` always work via the `CLOCK` subscription — sleeping fibers on WASM park correctly instead of spinning.

## Acceptance

- `zig build wasm` green; timer tests pass under a WASI runtime.
- Playground (browser WASM) unaffected — fibers already save/restore VM register arrays, not CPU stacks.

baijum:
Implements #1442 (KEP-0001 Phase 4). Design: [KEP-0001 §2](https://github.com/kaappi/keps/blob/main/keps/0001-event-loop-reactor.md#2-backends).

## What

- **`WasiPollBackend`** replaces the timer-only stopgap in `src/reactor.zig`. `poll_oneoff` is stateless, so the backend keeps the kernel's bookkeeping in userspace: `arm()` records armed directions per fd in an interest map, and every `wait()` rebuilds the subscription list from it — one `FD_READ`/`FD_WRITE` subscription per armed direction plus one `CLOCK` subscription bounding the wait (the `mio` wasi model). Returned events map back to registrations by fd via `userdata`; per-subscription errors and `HANGUP` wake both directions defensively (same policy as epoll `HUP`/`ERR`); a delivered direction disarms itself for ONESHOT parity (per-filter, like kqueue — an untouched direction stays armed for its own waiters).
- **`maybeSetNonblocking` is the WASI host-capability probe**: it now attempts `fd_fdstat_set_flags(NONBLOCK)`, and the three `is_wasm` EAGAIN gates in `primitives_io.zig` are removed. A host that refuses the flip keeps ports blocking — no EAGAIN, no fd registrations, CLOCK-only waits — degrading to single-fiber blocking I/O exactly where the host lacks support, per the KEP's cross-platform section.
- **`thread-sleep!` registers on WASM** (the one SRFI-18 entry point with no OS-thread dependency), making the timer path Scheme-visible there. The WASM spec table is a comptime-filtered `wasm_specs` subset; since that forces the whole file to compile on wasm32-wasi, `threadStartFn`'s `std.Thread.spawn` moved behind a comptime branch and the `u64 signal_generation` atomics (wasm32 has no 64-bit atomics; the build is single-threaded) take a plain-access branch. The `(srfi 18)` library itself stays native-only; native behavior is unchanged.
- **`tests/wasm/timers.scm`** runs under wasmtime in the CI wasm job.

## Design deviations from the issue text

- **Relative `CLOCK`, not ABSTIME.** The issue (quoting the KEP sketch) says ABSTIME at the nearest deadline, but the Phase 1 reactor core already reduces the timer heap's nearest deadline to a *relative* bound in `effectiveTimeout()` for all backends. ABSTIME here would re-derive the absolute deadline it came from, add a clock read per poll, and couple the backend to `clockNs()`'s clock domain. Behavior is identical ("may fire late, never early").
- **Fd path is dormant on today's WASM builds**: `open-input-file`/`open-output-file` are native-only, so no fd>2 port can exist on WASM yet. The fd half of the backend is compile-verified and engages the day file/socket ports land there.

## Playground

Unaffected, verified against the shim source (`wasi-shim-bundle.mjs`): its `poll_oneoff` accepts only a single CLOCK subscription — which is all it can ever receive, because its `fd_fdstat_set_flags` returns NOTSUP, so no port ever goes non-blocking in the browser and no fd subscription can exist. `poll_oneoff` was already in the import set (wasi-libc implements the old stopgap's `nanosleep` through it). `thread-sleep!` becomes *callable* in the playground (previously an unbound-variable error); the shim serves it with its pre-existing busy-wait loop, stalling only the worker thread.

## Verification

- `zig build test` green (macOS native); `zig build -Dtarget=x86_64-linux` compiles (epoll backend init-signature change)
- `bash tests/scheme/run-all.sh`: 1834 pass, 0 fail; R7RS 1395/1395
- `zig build wasm` green; `wasmtime run` on `tests/wasm/smoke.scm` and `tests/wasm/timers.scm` both pass
- Timer test timing: 0.03 s CPU over 0.4 s wall — the process genuinely blocks in `poll_oneoff` during sleeps instead of spinning

Closes #1442

### DIFF
--- a/src/primitives_io.zig
+++ b/src/primitives_io.zig
@@ -147,11 +147,26 @@ fn isBufferedFdPort(port: *types.Port) bool {
 /// mode), so flipping them would leak non-blocking mode outside this
 /// process. Without a scheduler nothing is flipped and sequential programs
 /// keep blocking fds. Pub only for tests_port_io's guard checks.
+///
+/// On WASI this doubles as the host-capability probe (KEP-0001 Phase 4):
+/// fd readiness is best-effort per the KEP, and a host that rejects
+/// fd_fdstat_set_flags(NONBLOCK) — the playground's browser shim does —
+/// keeps the port on a blocking fd, so no read ever EAGAINs, nothing
+/// registers with the reactor, and I/O degrades to single-fiber blocking
+/// exactly where the host can't support better.
 pub fn maybeSetNonblocking(port: *types.Port) void {
-    if (comptime is_wasm) return; // WASI fd readiness is KEP-0001 Phase 4
     if (port.nonblocking or port.is_string_port or port.fd <= 2) return;
     const vm = vm_mod.vm_instance orelse return;
     if (vm.scheduler == null) return;
+    if (comptime is_wasm) {
+        var stat: std.os.wasi.fdstat_t = undefined;
+        if (std.os.wasi.fd_fdstat_get(port.fd, &stat) != .SUCCESS) return;
+        var flags = stat.fs_flags;
+        flags.NONBLOCK = true;
+        if (std.os.wasi.fd_fdstat_set_flags(port.fd, flags) != .SUCCESS) return;
+        port.nonblocking = true;
+        return;
+    }
     const flags = std.c.fcntl(port.fd, std.posix.F.GETFL, @as(c_int, 0));
     if (flags < 0) return;
     const nonblock: c_int = @intCast(@as(u32, @bitCast(std.posix.O{ .NONBLOCK = true })));
@@ -253,7 +268,6 @@ fn drainWriteBuffer(port: *types.Port) PrimitiveError!void {
             const e = std.posix.errno(rc);
             if (e == .INTR) continue;
             if (e == .AGAIN) {
-                if (comptime is_wasm) break;
                 try waitPortFd(port, .write);
                 continue;
             }
@@ -624,7 +638,6 @@ pub fn readOneByte(port: *types.Port) PrimitiveError!?u8 {
             const e = std.posix.errno(raw);
             if (e == .INTR) continue;
             if (e == .AGAIN) {
-                if (comptime is_wasm) return null;
                 try waitPortFd(port, .read);
                 continue;
             }
@@ -954,7 +967,6 @@ fn readDatumFn(args: []const Value) PrimitiveError!Value {
             const e = std.posix.errno(raw_n);
             if (e == .INTR) continue;
             if (e == .AGAIN) {
-                if (comptime is_wasm) break;
                 // Mid-datum would-block: a parked retry re-executes this
                 // primitive from the top, so the partial accumulation goes
                 // back into port.read_buf and is re-drained on entry. The
--- a/src/reactor.zig
+++ b/src/reactor.zig
@@ -1,10 +1,11 @@
-//! Per-OS-thread I/O readiness multiplexer (KEP-0001 Phases 1-2).
+//! Per-OS-thread I/O readiness multiplexer (KEP-0001).
 //!
 //! One `Reactor` belongs to one OS thread's scheduler. A fiber that would
 //! block on a fd registers its interest and parks; `poll` blocks once,
 //! bounded by the nearest timer deadline, and reports every fiber that
 //! became runnable (fd readiness or timer expiry). Wired into the
-//! scheduler in KEP-0001 Phase 2 (fiber.zig's runSchedulerStep). See
+//! scheduler in KEP-0001 Phase 2 (fiber.zig's runSchedulerStep). Backends:
+//! kqueue (macOS/BSD), epoll (Linux), poll_oneoff (WASI, Phase 4). See
 //! https://github.com/kaappi/keps/blob/main/keps/0001-event-loop-reactor.md
 const std = @import("std");
 const builtin = @import("builtin");
@@ -28,7 +29,7 @@ const ReadyEvent = struct { fd: i32, readable: bool, writable: bool };
 const Backend = switch (builtin.os.tag) {
     .macos, .ios, .tvos, .watchos, .visionos => KqueueBackend,
     .linux => EpollBackend,
-    .wasi => WasiBackend,
+    .wasi => WasiPollBackend,
     else => @compileError("reactor: unsupported OS (kqueue/epoll/wasi only)"),
 };
 
@@ -71,7 +72,7 @@ pub const Reactor = struct {
     pub fn init(allocator: std.mem.Allocator) !Reactor {
         return .{
             .allocator = allocator,
-            .backend = try Backend.init(),
+            .backend = try Backend.init(allocator),
             .regs = std.AutoHashMap(i32, Reg).init(allocator),
             .timers = .empty,
         };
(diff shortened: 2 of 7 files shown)


src/reactor.zig
```diff
@@ -489,56 +492,164 @@ fn msFromNs(timeout_ns: ?u64) i32 {
...
+            const writable = ev.type == .FD_WRITE or broken;
+            self.clearInterest(fd, readable, writable);
+            for (self.ready.items) |*re| {
+                if (re.fd == fd) {
+                    if (readable) re.readable = true;
+                    if (writable) re.writable = true;
+                    continue :outer;
+                }
             }
+            try self.ready.append(self.allocator, .{ .fd = fd, .readable = readable, .writable = writable });
```
baijum: **Consumed-but-undelivered hazard** (non-blocking — OOM-only, and the fd path is dormant on WASM today): if this `append` fails, the event is lost *after* `clearInterest` above already disarmed the direction. The parked fiber is then stranded: nothing re-arms, the next `wait()` omits the fd from the subscription list, and with no other subscriptions the scheduler busy-spins in `parkOnReactor` (`isEmpty()` stays false via the still-listed waiter, but `wait()` early-returns empty). This is exactly the invariant `Reactor.poll` documents and guards with its up-front `ensureUnusedCapacity` — the kqueue/epoll backends dodge it by using fixed arrays, but this backend introduced a fallible allocation between event consumption and delivery.

At most `nevents` entries can be appended, so reserving once before the loop closes it:

```zig
try self.ready.ensureTotalCapacity(self.allocator, nevents);
```

right after the `poll_oneoff` return-code switch, and `appendAssumeCapacity` here. (The `try self.subs.append` / `events.ensureTotalCapacity` failures earlier are fine — they happen before the syscall, with interests intact.)

src/primitives_io.zig
```diff
@@ -147,11 +147,26 @@ fn isBufferedFdPort(port: *types.Port) bool {
...
+/// fd_fdstat_set_flags(NONBLOCK) — the playground's browser shim does —
+/// keeps the port on a blocking fd, so no read ever EAGAINs, nothing
+/// registers with the reactor, and I/O degrades to single-fiber blocking
+/// exactly where the host can't support better.
 pub fn maybeSetNonblocking(port: *types.Port) void {
-    if (comptime is_wasm) return; // WASI fd readiness is KEP-0001 Phase 4
     if (port.nonblocking or port.is_string_port or port.fd <= 2) return;
     const vm = vm_mod.vm_instance orelse return;
     if (vm.scheduler == null) return;
+    if (comptime is_wasm) {
```
baijum, line 161: Nit, dormant until fd>2 ports exist on WASM: on a host that refuses the flip (the playground shim NOTSUPs every `fd_fdstat_set_flags`), `port.nonblocking` never becomes true, so this probe re-runs on **every** read/write of such a port once a scheduler exists — two host calls per I/O op at all three call sites (`drainWriteBuffer`, `readOneByte`, `readDatumFn`). The native `fcntl` path has the same shape but effectively never fails, so it never pays this. When file/socket ports land on WASM, consider caching the refusal (e.g. a `nonblocking_probe_failed` flag on `types.Port`) so a refusing host pays the probe once per port instead of per op.

baijum (commented):
## Review

Reviewed the full diff against the checked-out head (141839ee), the Phase 1–3 reactor/scheduler code it plugs into, the Zig 0.16 `std.os.wasi` declarations, and the playground shim source. Re-verified locally on macOS ARM: `zig build test` green, `zig build wasm` green, `wasmtime run tests/wasm/timers.scm` passes (0.04 s CPU over 0.43 s wall — genuinely blocking in `poll_oneoff`, not spinning), `zig build -Dtarget=x86_64-linux` compiles the epoll init-signature change. CI is fully green including the wasm job.

### What holds up under scrutiny

- **Backend contract parity.** The interest map correctly emulates per-filter ONESHOT: `arm()` at both reactor call sites passes complete desired state (register: full waiter lists; poll re-arm: what remains), so replace semantics are right, and a delivered direction self-disarms. I also traced the fd-recycle case (GC-freed port closes without unregister): the stale interest either fires once and self-clears, or is wholesale replaced by the next `register()`'s `put` — the same self-heal shape as epoll's ENOENT↔EEXIST dance, no extra code needed.
- **`error != SUCCESS` short-circuits before `fd_readwrite.flags` is read** (undefined for errored/CLOCK events) — the `or` ordering matters and is correct.
- **The empty-subscription guard is real**: `poll_oneoff` INVALs on 0 subs, `parkOnReactor` checks `isEmpty()` first, and `schedule()`'s zero-timeout poll only runs with io_waiters (which imply interests). Confirmed unreachable through the scheduler.
- **Playground claims verified against `wasi-shim-bundle.mjs`**: `fd_fdstat_set_flags` NOTSUPs via the base class (so the capability probe correctly leaves ports blocking), `poll_oneoff` accepts exactly one CLOCK sub (all it can ever receive), MONOTONIC + relative flags are handled, and the worker runs `kaappi program.scm` un-sandboxed so `thread-sleep!` (`.sandbox = false`, `.wasm = true`) does register there.
- **The comptime story** (`wasm_specs` filter + `threadStartFn` taken-branch guard + the two `signal_generation` helpers) is internally consistent, and I confirmed all `signal_generation` accesses route through the helpers and the remaining atomics in the file are ≤32-bit or bool.
- **Relative-CLOCK deviation from the issue text** is the right call — `effectiveTimeout()` already produces a relative bound for every backend; ABSTIME would re-derive what it just consumed.
- Test file exercises the right seams (main-fiber drive, timer-heap ordering, sleeper-vs-runnable-sibling) with safe timing margins for CI; every primitive it uses is wasm-registered.

### Findings

Two non-blocking comments inline (one ordering fix worth taking now or as an immediate follow-up, one dormant-path perf note). One optional test-coverage thought: the interest-map logic (`arm`/`clearInterest`/rebuild) is pure userspace state and `std.os.wasi` types compile on every platform, so a native Zig unit test could pin the ONESHOT-parity behavior without a WASI runtime — worth considering when the fd path goes live with file/socket ports, not needed to land this.

### Verdict

**Ready for merge.** CI is fully green, the design deviations are well-argued, and both findings are on the compile-verified-but-dormant fd path (no fd>2 port can exist on WASM yet). Fine to merge as-is and fold the `ready`-reservation fix into the next port-layer PR, or push it here first — either works. Merge manually per the usual flow (auto-merge not armed).

src/reactor.zig
baijum: Fixed in 2824536e: `ready` is reserved for `nevents` entries right after the return-code switch (before any `clearInterest` can run), and the loop uses `appendAssumeCapacity`. Dedup only shrinks the count, so `nevents` is a valid bound. The remaining fallible allocations (`subs` build, `events` buffer) all precede the syscall with interests intact, as you noted. Re-verified: `zig build wasm` green, wasmtime timers.scm passes, native `zig build test` green.

src/primitives_io.zig
baijum: Agreed, and deferring per your framing — the probe-per-op cost only exists once fd>2 ports can exist on WASM, and the right shape then is a refusal flag on `types.Port` (probe once per port). Recorded as a follow-up alongside the native ONESHOT-parity unit test idea for when the fd path goes live.

Issue #1463, opened 2026-07-12
thread-sleep! lacks the yield-retry branch waitForFd has, causing unbounded nested-stack stalls under concurrent fibers


baijum:
## Summary

Found while building the retry loops for kaappi-net's reactor-aware TLS ([kaappi-net#3](https://github.com/kaappi/kaappi-net/pull/3)) and kaappi-pg's async libpq rewrite ([kaappi-pg#1](https://github.com/kaappi/kaappi-pg/pull/1)), both part of #1444 (KEP-0001 Phase 6).

`fiber.waitForFd` (`src/fiber.zig:552`) branches on the calling fiber's context:

```zig
if (my_idx != 0 and vm.dispatched_from_scheduler) {
    vm.yield_retry = true;
    return VMError.Yielded;
}
```

For a fiber dispatched directly by a scheduler loop, it returns `error.Yielded` — which unwinds cleanly up to the dispatch loop (a *flat* suspend, no new native call). Only the main fiber (or one under re-entrant native frames) "drives the scheduler in place" via a nested call.

`primitives_srfi18.zig`'s `threadSleepFn` (line ~416) has no equivalent branch — it *always* calls `runSchedulerStep` directly:

```zig
const ctx = try ensureScheduler();
...
try ctx.reactor.addTimer(deadline, me);
_ = try fiber_mod.runSchedulerStep(SleepWait, .{}, ctx.vm, ctx.sched, me);
```

## Impact

When two (or more) fibers each retry through many short `thread-sleep!` calls concurrently — e.g. a userspace poll-and-sleep pattern like `kaappi-http`'s existing `fiber-recv`/`fiber-send` (`kaappi-http/lib/kaappi/http/server.sld:105-119`), or the TLS/libpq retry loops added in the two PRs above — every hand-off between fibers that itself needs another `thread-sleep!` nests one more native (Zig) stack frame. Each frame only unwinds once *that specific* nested call's own timer fires, but new frames keep getting added as sibling fibers keep retrying, so the native call stack grows roughly linearly with the total number of `thread-sleep!` calls made across all fibers before the underlying condition resolves.

Confirmed via `lldb bt all` on a hung process (kaappi-pg, two fibers each polling a 200ms `pg_sleep` query at a 1ms interval): 100+ frames of
```
threadSleepFn -> vm_dispatch.runUntil -> vm.runUntil -> fiber.runSchedulerStep -> threadSleepFn -> ...
```
[...]

PR #1476, opened 2026-07-12
KEP-0001 Phase 7: performance evaluation, edge-triggered go/no-go, thread-sleep! fix

baijum:
## Summary

Phase 7 (measurement) of the KEP-0001 event-loop-reactor epic — closes out
the epic's last open item. Full results, methodology, and data tables in
[docs/dev/kep-0001-phase7-benchmarks.md](docs/dev/kep-0001-phase7-benchmarks.md).

- **Prerequisite fix (#1463)**: `thread-sleep!` lacked the same
  `dispatched_from_scheduler`-aware yield-retry branch `fiber.waitForFd` has,
  causing unbounded native stack growth under concurrent fibers each
  retrying via short `thread-sleep!` calls — exactly the pattern
  `kaappi-http`'s `http-listen-fiber` uses. Needed before any of this phase's
  benchmarking could be done safely. Regression test confirmed to segfault
  without the fix, pass with it.
- **Q1 (wake-all herd cost)**: confirmed negligible — `poll()`'s wake-all
  fan-out doesn't scale with waiter count (measured on both kqueue and
  epoll).
- **Q2 (timer granularity)**: confirmed — no workload in this codebase needs
  sub-ms precision; `timerfd` escape hatch stays unbuilt.
- **Q3 (ONESHOT re-arm cost)**: confirmed negligible on both backends
  (register() cost is 0.77x of an adjacent read+write pair on macOS, 0.47x
  on Linux) → **edge-triggered migration: no-go for now**.
- **Q5 (per-fiber memory/switch time — the one fully-open question)**:
  memory is bounded and modest, but `FiberScheduler.schedule()`/`addFiber()`
  do an O(fiber count) scan per dispatch/spawn, degrading switch time by
  ~2 orders of magnitude between 100 and 10,000 live fibers. This shows up
  end-to-end in the HTTP server benchmarks below.
- **Server benchmarks**: compared `http-listen`/`-prefork`/`-fiber` at
  100/1,000 concurrent connections (no `wrk`/`hey` in this workspace, used a
  small Python load generator instead). `http-listen-fiber` pays a ~1ms tax
  even uncontended (its I/O never actually reaches the reactor — see
  follow-ups) and its p99 at 1,000 connections is *worse* than the naive
  sequential server's, tracing back to the scheduler O(n) scan above.
- **`http-listen-threaded` could not be benchmarked** — found and isolated a
  pre-existing hang (confirmed on unmodified `main`, unrelated to this PR's
  changes): a library-defined procedure that `thread-start!`s a thunk which
  calls into *another* library's procedure hangs the child thread; identical
  code at top-level works. Filing as a follow-up rather than fixing here.

New benchmark tools added: `zig build bench-fibers`, `zig build bench-reactor`.

Four follow-up issues filed for the residual findings (linked in the
doc and the tracking issue comment).

## Test plan

- [x] `zig build test` (full unit suite) green
- [x] `bash tests/scheme/run-all.sh` (1834 tests) green
- [x] New regression test confirmed to fail (segfault) without the fix,
      pass with it
- [x] `zig fmt --check` clean on all changed/new files
- [x] Stress-tested `http-listen-fiber` at 50/500/1,000 concurrent
      connections post-fix (0 errors)

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -202,6 +202,34 @@ pub fn build(b: *std.Build) void {
     const bench_step = b.step("bench", "Run the call/cc capture benchmark");
     bench_step.dependOn(&run_bench.step);
 
+    // Benchmark executable (per-fiber memory & switch-time, KEP-0001 P7 Q5)
+    const bench_fibers_mod = kaappiModule(b, options_mod, .{
+        .root = "src/bench_fibers.zig",
+        .target = target,
+        .optimize = optimize,
+    });
+    const bench_fibers_exe = b.addExecutable(.{
+        .name = "bench-fibers",
+        .root_module = bench_fibers_mod,
+    });
+    const run_bench_fibers = b.addRunArtifact(bench_fibers_exe);
+    const bench_fibers_step = b.step("bench-fibers", "Run the per-fiber memory & switch-time benchmark");
+    bench_fibers_step.dependOn(&run_bench_fibers.step);
+
+    // Benchmark executable (reactor wake-all/re-arm/timer costs, KEP-0001 P7 Q1/Q2/Q3)
+    const bench_reactor_mod = kaappiModule(b, options_mod, .{
+        .root = "src/bench_reactor.zig",
+        .target = target,
+        .optimize = optimize,
+    });
+    const bench_reactor_exe = b.addExecutable(.{
+        .name = "bench-reactor",
+        .root_module = bench_reactor_mod,
+    });
+    const run_bench_reactor = b.addRunArtifact(bench_reactor_exe);
+    const bench_reactor_step = b.step("bench-reactor", "Run the reactor wake-all/re-arm/timer benchmark");
+    bench_reactor_step.dependOn(&run_bench_reactor.step);
+
     // Package manager (thottam)
     const thottam_mod = kaappiModule(b, options_mod, .{
         .root = "src/thottam.zig",
--- a/docs/dev/kep-0001-phase7-benchmarks.md
+++ b/docs/dev/kep-0001-phase7-benchmarks.md
@@ -1,0 +1,313 @@
+# KEP-0001 Phase 7: Performance Evaluation
+
+Tracking issue: [#1445](https://github.com/kaappi/kaappi/issues/1445). Epic:
+[#1438](https://github.com/kaappi/kaappi/issues/1438). Phases 1–6 are merged;
+this phase measures the reactor's real-world costs and confirms (or
+overturns, with data) the design decisions recorded in [keps#2](https://github.com/kaappi/keps/pull/2).
+
+All benchmark sources live in `src/bench_reactor.zig` (Q1/Q2/Q3, run with
+`zig build bench-reactor`) and `src/bench_fibers.zig` (Q5, run with
+`zig build bench-fibers`). Server benchmarks used a one-off load generator
+(no `wrk`/`hey`/`ab` available in this workspace) — see the ecosystem
+benchmark section below for how to reproduce.
+
+## Summary
+
+| Question | Decision | Confirmed? |
+|---|---|---|
+| Q1 — wake-all herd cost | Wake-all (not wake-head-only) | **Confirmed** — no measurable herd cost at the reactor level |
+| Q2 — timer granularity | ms + ceil-round on Linux, no `timerfd` | **Confirmed** — no workload needs sub-ms precision |
+| Q3 — ONESHOT vs LT | ONESHOT | **Confirmed** — re-arm cost is well under the cost of the adjacent I/O syscall it accompanies |
+| Q5 — per-fiber memory | Live-window save/restore | **Partially confirmed** — memory is bounded and modest, but per-switch *time* degrades with total live-fiber count (see below) — the actual residual this phase was meant to surface |
+| Edge-triggered migration | — | **No-go for now** (see below) |
+
+Prerequisite fix landed before any of this could be measured safely: see
+"Prerequisite: #1463" below.
+
+## Q1 — Wake-all herd cost
+
+`src/bench_reactor.zig`'s `benchWakeAllFanout`: N fake fibers registered as
+waiters on one shared fd, one real write, measuring how long `poll()` takes
+to wake all N (confirming wake-all, not wake-head-only).
+
+macOS (kqueue):
+
+| N | woken | poll (ns) | ns/woken-fiber |
+|---|---|---|---|
+| 2 | 2 | 2000 | 1000.0 |
+| 10 | 10 | 2000 | 200.0 |
+| 100 | 100 | 1000 | 10.0 |
+| 1000 | 1000 | 4000 | 4.0 |
+
+Linux (epoll, aarch64 container):
+
+| N | woken | poll (ns) | ns/woken-fiber |
+|---|---|---|---|
+| 2 | 2 | 3584 | 1792.0 |
+| 10 | 10 | 1542 | 154.2 |
+| 100 | 100 | 2125 | 21.3 |
+| 1000 | 1000 | 10500 | 10.5 |
+
+`poll()`'s own wake-all fan-out cost does **not** grow with N — it's
+dominated by a small fixed per-call cost (array copy), not a per-waiter one.
+The KEP's wake-head-only-FIFO fallback is not warranted at the reactor
+level. (The real per-waiter cost of wake-all — every registered fiber
+retrying its syscall and only one winning — is an *application*-level
+tradeoff, not a reactor one; see the fiber HTTP server latency numbers
+below, where it plausibly contributes at high concurrency.)
+
+## Q2 — Timer granularity
+
+`src/bench_reactor.zig`'s `benchTimerGranularity`: schedules one timer at a
+known deadline, measures how late `poll()` actually returns.
+
+macOS (kqueue, ns-precision `timespec` API):
+
+| requested | late by |
+|---|---|
+| 1ms | +133µs |
+| 5ms | +631µs |
+| 20ms | +1033µs |
+| 100ms | +1018µs |
+
+Linux (epoll, in a podman/QEMU-virtualized container — expect extra jitter
+from virtualization on top of the numbers below):
+
+| requested | late by |
+|---|---|
+| 1ms | +260µs |
+| 5ms | +215µs |
+| 20ms | +660µs |
+| 100ms | +2412µs |
+
+Two things worth noting: (1) even kqueue's nanosecond-precision timer API
+shows a practical floor around ~1ms once the requested duration is more
+than a few ms — this is OS thread-scheduling wake-up latency, not a kaappi
+limitation, and it means kqueue's theoretical precision advantage over
+epoll's ms-ceil-rounding (`msFromNs`, `src/reactor.zig:487-491`) mostly
+doesn't show up in practice. (2) A grep across the test suites and
+ecosystem repos (`kaappi-net`, `kaappi-pg`, `kaappi-http`) found no
+`thread-sleep!` interval below 1ms in real (non-microbenchmark) code — the
+Q2 decision holds: nothing in this codebase needs sub-ms deadlines, so the
+`timerfd` escape hatch stays unbuilt.
+
+## Q3 — ONESHOT re-arm cost
+
+`src/bench_reactor.zig`'s `benchRearmVsIo`: 100,000 cycles of `register()`
+(the `epoll_ctl`/`kevent` call) vs. an adjacent `read(2)`+`write(2)` pair on
+the same pipe.
+
+| Backend | register() (arm) | read+write (io) | ratio |
+|---|---|---|---|
+| macOS (kqueue) | 338.5 ns | 442.1 ns | 0.77x |
+| Linux (epoll) | 173.8 ns | 372.9 ns | 0.47x |
+| Linux, wake-all fan-out with an OS-thread instead of a fiber (not measured) | — | — | — |
+
+On both platforms the re-arm call is *cheaper* than the I/O syscall pair
+it accompanies every cycle. There is no meaningful overhead here for
+edge-triggered mode to eliminate.
+
+## Q5 — Per-fiber memory and switch time
+
+`src/bench_fibers.zig`. Three things were measured: raw switch/spawn cost
+at scale, RSS, and whether the native-frame `frameWindow()` fallback
+(`types.zig:546-551`, 256 registers for any frame with no attached
+closure) measurably inflates a fiber's saved register/frame arrays.
+
+### Switch time degrades with total live-fiber count
+
+Spawning N fibers, each yielding 50 times in a tight loop, then joining all:
+
+| fibers | ns/switch | RSS |
+|---|---|---|
+| 100 | ~465 ns | +0 MB |
+| 1,000 | ~3,300–4,600 ns | +0 MB |
+| 10,000 | ~93,000–99,000 ns | +0.06 MB |
+
+Consistent across three separate runs. Isolating spawn cost alone (0 yield
+rounds, just N `spawn` + `fiber-join` pairs) shows the same shape: ~10–30
+µs/spawn at 100–1,000 fibers, jumping to ~250–280 µs/spawn at 10,000.
+
+**Root cause**: `FiberScheduler.schedule()` (`src/fiber.zig:302-325`) does a
+full round-robin scan over *every* slot in `sched.fibers.items` (including
+completed/errored ones still occupying a slot) to find the next runnable
+fiber — O(fiber count) per dispatch, called on every single switch.
+`FiberScheduler.addFiber()` (`src/fiber.zig:86-99`) does the same kind of
+O(fiber count) linear scan for a free slot on every spawn. Per-fiber
+memory itself stays flat and modest (the RSS deltas above are noise-level
+until 10k, where it's still only +0.06MB) — it's *switch time*, not memory,
+that's the real Q5 residual, and it degrades by roughly two orders of
+magnitude between 100 and 10,000 concurrently-live fibers.
+
+This is not just a synthetic-benchmark artifact — it shows up end-to-end in
+the HTTP server benchmarks below (`http-listen-fiber`'s p99 latency at
+1,000 concurrent connections is *worse* than the naive sequential server's).
+**Follow-up filed**: replacing the O(n) scan with an actual ready queue
+(the KEP's Q1 discussion already floated "wake-head-only FIFO" for a
+related reason — this is the same class of fix).
+
+### Native-frame register/frame inflation: not observed
+
+Compared fibers yielding from a pure-bytecode tail loop against fibers
+yielding from inside `for-each`'s native call frame, across recursion
+depths 0, 10, 60, and 300 (with up to 7 extra bound locals per frame to
+vary register pressure). In every configuration tested, both kinds
+allocated identical register/frame array sizes (e.g. 256 registers / 32
+frames at shallow depth, growing identically to 4096/512 at depth 300 with
+extra locals). **No inflation was observed in practice** across this
+range — the initial per-fiber capacities (`INITIAL_FIBER_REGISTER_CAPACITY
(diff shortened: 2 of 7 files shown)


src/primitives_srfi18.zig
```diff
@@ -413,21 +413,49 @@ const SleepWait = struct {
...
+        me.deadline_ns = null;
+        return types.VOID;
+    }
+
+    if (my_idx != 0 and ctx.vm.dispatched_from_scheduler) {
+        ctx.vm.yield_retry = true;
+        return PrimitiveError.Yielded;
+    }
 
     _ = try fiber_mod.runSchedulerStep(SleepWait, .{}, ctx.vm, ctx.sched, me);
```
baijum, line 471: Fix verified A/B (crashes on `main`'s version of this file + the new test, passes here), and I audited the `deadline_ns == null` discriminator against every other site that sets `me.deadline_ns` (`thread-join!`, `mutex-lock!`, `mutex-unlock!`+condvar, `thread-terminate!`) — all of them clear it before returning to Scheme, so the fresh-call/redispatch split is sound. Nice doc comment, too.

One non-blocking hardening idea: if `addTimer` above or `runSchedulerStep` here fails (OOM-only today), the fiber keeps `deadline_ns` set and `.waiting` status while the error becomes a catchable Scheme exception, and the *next* `thread-sleep!` misreads that as a redispatch. An `errdefer` clearing `deadline_ns` + `removeTimer(me)` would make the invariant self-healing. Same fragility pre-exists in `mutexLockFn`, so fine to skip or file later.

src/tests_srfi18.zig
```diff
@@ -498,3 +498,44 @@ test "thread-yield! in child OS thread does not busy-spin or leak Yielded" {
...
+        \\    (let loop ((n 0))
+        \\      (if (>= n 3000)
+        \\          (begin (set! signal #t) n)
+        \\          (begin (thread-sleep! 0.0001) (loop (+ n 1))))))))
+        \\(define waiter (spawn poll-until-signal))
+        \\(define setter-result (fiber-join setter))
+        \\(define waiter-result (fiber-join waiter))
+        \\(and (= setter-result 3000)
+        \\     (>= waiter-result 0)
+        \\     (<= waiter-result 3000)
```
baijum: The `(<= waiter-result 3000)` upper bound encodes a scheduler-internals guarantee, not a property of the fix. It holds today only because the round-robin rotation deterministically dispatches the setter before the waiter each wake round (`current_idx` ends at the waiter before every park, so the next scan reaches slot 1 first) — the waiter can therefore never complete more sleeps than the setter. The moment #1477 replaces the O(n) scan with a ready queue, dispatch order within a wake round is up for grabs, and a waiter that gets dispatched first can legitimately return 3001+ — turning this into a flaky failure in the exact PR that fixes the scheduler.

The regression signal here is "completes without crashing", so the bound adds no detection value. Suggest dropping it (or loosening to something like `(< waiter-result 100000)` if you want a sanity ceiling).

docs/dev/kep-0001-phase7-benchmarks.md
```diff
@@ -0,0 +1,313 @@
...
+
+`src/bench_reactor.zig`'s `benchRearmVsIo`: 100,000 cycles of `register()`
+(the `epoll_ctl`/`kevent` call) vs. an adjacent `read(2)`+`write(2)` pair on
+the same pipe.
+
+| Backend | register() (arm) | read+write (io) | ratio |
+|---|---|---|---|
+| macOS (kqueue) | 338.5 ns | 442.1 ns | 0.77x |
+| Linux (epoll) | 173.8 ns | 372.9 ns | 0.47x |
+| Linux, wake-all fan-out with an OS-thread instead of a fiber (not measured) | — | — | — |
```
baijum: This row looks like an editing artifact — it describes a Q1-style wake-all fan-out variant in the middle of the Q3 re-arm table, with no data. Suggest deleting the row (or, if OS-thread-based fan-out is a deliberate not-measured caveat, moving it to prose under Q1).

docs/dev/kep-0001-phase7-benchmarks.md
```diff
@@ -0,0 +1,313 @@
...
+
+### Reproducing
+
+```
+cd kaappi-http
+# terminal 1
+DYLD_LIBRARY_PATH=..:../../kaappi-net zig-out/bin/kaappi \
+  --lib-path ../kaappi-net/lib --lib-path lib bench_server_app.scm fiber
+# terminal 2 (no wrk/hey in this workspace -- see scratch script in this PR's description)
+python3 http_load_gen.py 19999 1000 5000
```
baijum: Reproducibility gap: this section tells the reader to run `bench_server_app.scm` and `http_load_gen.py`, but neither file exists in any repo, and the "scratch script in this PR's description" isn't there — the PR description doesn't include it. As written, the ecosystem-server table (the data behind two of this phase's findings) can't be reproduced.

Suggest either committing the load generator + server app (e.g. `benchmarks/http/` here, or under `kaappi-http`), or pasting them into the PR description as promised, or rewording to describe the client's behavior precisely enough to rewrite it (threaded, one connection per request, `Connection: close`, args = port/concurrency/requests).

docs/dev/kep-0001-phase7-benchmarks.md
```diff
@@ -0,0 +1,313 @@
...
+`thread-sleep!` calls grew the native call stack without bound. Confirmed
+via a regression test that segfaults (stack overflow) without the fix and
+passes with it (`src/tests_srfi18.zig`). This was a **necessary
+prerequisite** for this phase: `http-listen-fiber`'s
+`fiber-recv`/`fiber-send` use exactly this poll-then-`thread-sleep!`
+pattern, and benchmarking it at any real concurrency (as confirmed by a
+500- and 1,000-concurrent-connection stress test after the fix) would
+very likely have hung or crashed instead of producing numbers.
+
+## Follow-ups to file
```
baijum: These are all filed now — suggest retitling to "Follow-ups (filed)" and linking them so the doc doesn't read as a stale todo list: item 1 → #1477, item 2 → #1478, item 3 → #1479, item 4 → #1480.

src/bench_fibers.zig
```diff
@@ -0,0 +1,239 @@
...
+    var ts: std.c.timespec = undefined;
+    _ = std.c.clock_gettime(.MONOTONIC, &ts);
+    return @as(u64, @intCast(ts.sec)) * std.time.ns_per_s + @as(u64, @intCast(ts.nsec));
+}
+
+// Matches the layout `getrusage(2)` fills in on both Linux (glibc) and
+// Darwin: two `timeval`s (16 bytes each on both platforms' LP64 ABIs, even
+// though the field widths inside differ) followed by a run of `long`s. Only
+// `ru_maxrss` is read, so the rest is left as opaque padding.
+const RUsage = extern struct {
```
baijum: Zig 0.16's std already ships this: `std.posix.getrusage(std.posix.rusage.SELF)` returns a properly-typed `rusage` with per-OS layouts (`maxrss` field), maintained by the std lib. The hand-rolled extern struct with byte-offset padding (`_times: [32]u8`) silently breaks if either platform's ABI assumption is ever wrong, and there's no assertion to catch it. The Darwin-bytes-vs-Linux-KiB divisor logic still needs to stay, but the struct + extern decl can go.

build.zig
```diff
@@ -202,6 +202,34 @@ pub fn build(b: *std.Build) void {
...
+        .root = "src/bench_fibers.zig",
+        .target = target,
+        .optimize = optimize,
+    });
+    const bench_fibers_exe = b.addExecutable(.{
+        .name = "bench-fibers",
+        .root_module = bench_fibers_mod,
+    });
+    const run_bench_fibers = b.addRunArtifact(bench_fibers_exe);
+    const bench_fibers_step = b.step("bench-fibers", "Run the per-fiber memory & switch-time benchmark");
```
baijum, line 216: Bit-rot risk: nothing in CI compiles `src/bench_fibers.zig` or `src/bench_reactor.zig` — `run-benchmarks.sh` only invokes `zig build bench`, and no test imports these files — so the next fiber/reactor API change can silently break them and nobody notices until someone re-runs a KEP benchmark months later. Cheapest guard: a compile-only step (e.g. `bench-build` that depends on the two `addExecutable` artifacts without the run steps) invoked from an existing CI job; running them in CI is probably too slow (the 10k-fiber case alone is ~50s).

Also minor: CLAUDE.md's Build section documents `zig build bench` — worth adding the two new steps there in the same pass.

baijum (commented):
## Review

Critical review with independent verification. **Verdict: ready to merge** — the one production change is correct and verified, CI is fully green (only by-design skipped jobs remain), and every finding below is minor/non-blocking. I'd fold in the two doc fixes (stale follow-ups section, reproducibility gap) before merging since the doc is this PR's main deliverable, but nothing here requires another review round.

### What I verified independently

- **A/B on the #1463 fix**: reverted `src/primitives_srfi18.zig` to `main` while keeping the new regression test → test **crashes** (stack overflow, 1 crashed in build summary). Restored the fix → 4/4 pass in 472ms. The PR's central claim holds.
- **Invariant audit of the `deadline_ns == null` discriminator**: the fix is only sound if no other wait primitive returns to Scheme with `me.deadline_ns` still set. Audited all sites — `thread-join!` (fiber path, line 609), `mutex-lock!` (833/855/875), `mutex-unlock!`+condvar (956/962/973), and `thread-terminate!` (which also `removeTimer`s and flips to `.errored`, so a terminated sleeper never re-executes the primitive). All clear it on every normal path. ✅
- **Wake-path protocol**: only the timer can wake a sleeping fiber (`waiting_on = VOID` never matches `wakeWaiters`; `wakeIoWaitersOnFd` only touches `.io_waiting`), `wakeReadyFiber` sets `.suspended` + `timed_out=true`, and the redispatch consumes it in the new `else if (me.timed_out)` branch. A spurious redispatch re-parks without double-arming the timer (the `deadline_ns != null` guard). Coherent.
- **Both benchmarks build and run** at the PR head and reproduce the doc's numbers on my machine: `bench-reactor` arm/io = **0.51x** (doc: 0.77x — same conclusion), Q1 fan-out flat, Q2 ms-scale lateness; `bench-fibers` **452ns → 4.6µs → 94µs** per switch at 100/1k/10k fibers (doc: ~465ns → ~3.3–4.6µs → ~93–99µs). The O(n) scheduler-scan finding and the edge-triggered no-go are independently confirmed.
- Doc line references spot-checked: `fiber.zig:302-325`/`86-99` (O(n) scans), `types.zig:527-528`/`546-551`, `reactor.zig:487-491` — all accurate at this head.

### One hardening note (non-blocking)

The error paths can leave the discriminator state set: if `addTimer` (line 446) or `runSchedulerStep` (line 460) fails, `me.deadline_ns` stays non-null (and `status` stays `.waiting`) while the error surfaces to Scheme as a catchable exception — a subsequent `thread-sleep!` on that fiber would then misread a fresh call as a redispatch and wait on a stale deadline. The error surface is OOM-only today and the same pattern pre-exists in `mutexLockFn`, so this is not a blocker — but an `errdefer { me.deadline_ns = null; ctx.reactor.removeTimer(me); }` around the fresh-arm path would make the new invariant self-healing. Fine as a follow-up or never.

CodeRabbit already covered the bench-methodology nits (DebugAllocator, `removeWaiter` inside the Q3 timed region, depth-60 comment vs `(level 10)` code, gc-stress scaling); I won't duplicate those. On the Q3 one specifically: `removeWaiter` is pure userspace list manipulation (`swapRemove` on a 1-element list), so its contribution is ~ns and the arm-vs-io conclusion is unaffected — worth a one-line reply there rather than a code change.

Inline comments below for the remaining items.

baijum:
## Review response

Went through every finding (CodeRabbit's 17 inline comments + the manual
review) and verified each against the actual code before deciding whether
to fix it. Commits `ee487f1`..`a58451f` address everything below marked
fixed; the ones marked skipped have a reason.

### Fixed

- **`primitives_srfi18.zig` / `threadSleepFn`**: added the `errdefer`
  hardening you suggested — clears `deadline_ns` and removes the timer if
  `addTimer`/the scheduler drive fails (OOM-only today), excluding
  `error.Yielded` since that's the intentional signal the discriminator
  exists to survive. (`ee487f1`)
- **`tests_srfi18.zig`**: fixed the vacuous `(>= waiter-result 0)` check
  (now `> 0`), dropped the `<= 3000` bound that baked in a round-robin
  scheduling guarantee `#1477` will invalidate, and scaled the retry count
  under `-Dgc-stress=true`. (`d79e563`)
- **`bench_fibers.zig` native-frame test — this was actually a bigger
  finding than "fix the depth comment."** You were right that it wasn't
  measuring what it claimed; digging into *why* turned up something more
  fundamental than the `native_reentry_depth` mechanism you flagged:
  `for-each` is pure bootstrap Scheme
  (`src/vm_bootstrap.zig`), not a Zig primitive, so its callback never
  pushes a native frame at all — both cases were exercising identical
  bytecode regardless of depth, which is why they always measured
  identically. Redid it with `with-exception-handler` (a genuine native
  primitive that calls its thunk via `callReentrant`), and `thread-sleep!`
  instead of `yield` as the suspension point (`yield` no-ops under
  `native_reentry_depth`, same reason your comment named). That surfaced a
  second, previously-unknown issue: `thread-sleep!` inside a re-entrant
  native frame always drives the scheduler recursively (nested `runUntil`
  clears `dispatched_from_scheduler` for its extent, the same mechanism
  that makes `yield` no-op there), so concurrently-dispatched fibers each
  doing this chain-nest the native stack — confirmed crashing at just N=2
  fibers with deep-enough recursion. Landed the corrected version at safe
  N=1/shallow depth and documented the result as **inconclusive** (blocked
  by the 256-register initial floor at that depth) rather than claiming a
  clean negative. Full writeup in the doc's "Native-frame register/frame
  inflation" section. (`cb939a1`)
- **`bench_fibers.zig` allocator/RSS**: swapped `DebugAllocator` for
  `c_allocator` (its bookkeeping was contaminating the timings — the
  10k-fiber switch-time number dropped from ~49ms to ~17.7ms once removed),
  and swapped the hand-rolled `RUsage` extern struct for
  `std.posix.getrusage(std.posix.rusage.SELF)` — verified it compiles and
  returns sane data on both macOS and an aarch64-linux cross-compile
  target. (`cb939a1`)
- **`bench_reactor.zig`**: validated `read`/`write` return values in the
  Q3 loop, preallocate the Q1 result list before starting the clock, and
  verify the Q2 timer actually fired (checking `ready` contains the fiber)
  before computing lateness instead of trusting any `poll()` return.
  (`de98d31`)
- **Docs**: removed the stray Q3 table row, relinked "Follow-ups to file"
  as "Follow-ups (filed)" with #1477–#1480, labeled the reproduction shell
  fence, documented RSS as a high-water-mark delta not a per-N peak, and —
  the bigger one — committed the server benchmark app + load generator to
  [`kaappi-http/benchmarks/`](https://github.com/kaappi/kaappi-http/tree/main/benchmarks)
  ([kaappi-http#3](https://github.com/kaappi/kaappi-http/pull/3)) so the
  ecosystem numbers are actually reproducible now. Every number in the doc
  was re-measured against the corrected code. (`1a6af0f`)
- **`CLAUDE.md`**: documented the two new `zig build` steps. (`a58451f`)

### Skipped, with reasons

- **`TestContext` refactor for the SRFI-18 test**: checked — every sibling
  test in `tests_srfi18.zig` uses the manual `GC`+`VM` pattern directly,
  not `TestContext`. Matching the file's actual prevailing convention over
  the general guideline that mentions `TestContext` as an option.
- **`std.process.Init.Minimal` entrypoint** (both bench files): checked —
  `src/bench.zig` (the existing benchmark, same category as these two)
  still uses plain `pub fn main() !void`. That convention is for shipped
  executables (`main.zig`, `kaappi_lsp.zig`), not one-off dev benchmarks.
- **Gate `bench-fibers` to POSIX targets**: `bench.zig` has the identical
  unconditional POSIX assumption (`std.c.clock_gettime`) and isn't gated
  either — consistent with existing precedent, and the default target is
  always native.
- **Reporting the single `arm_ns`/`io_ns` sample as both min and max**:
  `bench.zig`'s own machine-readable summary lines do exactly this
  (`min: {d}, max: {d}` from one sample) — it's this project's established
  format for that summary line, not a bug specific to these files.
- **`arm_ns` includes `removeWaiter()`, not just the raw `epoll_ctl`/`kevent`
  call**: agreed this is technically true — added a caveat to the doc
  rather than a code change, since `removeWaiter` is a userspace
  `swapRemove` on a 1-element list (nanosecond-scale), and the maintainer
  review already assessed this as not changing the arm-vs-io conclusion.
- **CI compile step for the bench files (bit-rot risk)**: agreed this is a
  real gap, but modifying `ci.yml` felt like more scope than this PR
  should carry on top of everything else here — happy to file it as a
  quick follow-up if you'd rather not merge with it open.

baijum:
## Re-review of the 6 new commits (ee487f1..a58451f5)

All feedback from my previous review (and CodeRabbit's) is addressed. Verified each fix against the diff, and rebuilt + reran both benchmarks, the regression test, and gc-stress locally.

| Review item | Resolved in | Verified |
|---|---|---|
| `thread-sleep!` OOM leaves stale `deadline_ns` (my non-blocking note) | `ee487f12` | ✅ `errdefer` clears `deadline_ns`+`removeTimer`, and correctly **excludes `error.Yielded`** — that scoping is exactly right: the only state where `deadline_ns` is set without a matching pending timer is the `addTimer`-failure window, which the errdefer covers; the `runSchedulerStep` path leaves `deadline_ns` consistent with a still-pending timer (or self-heals via the `timed_out` branch). |
| Test's `(<= waiter-result 3000)` will flake under #1477 | `d79e5638` | ✅ Replaced with a loose `< 100_000` ceiling, documented with the #1477 reasoning; vacuous `(>= waiter-result 0)` → `waiter_result > 0` (requires a real retry); scaled to n=300 under gc-stress. Passes normally **and** under `-Dgc-stress=true` locally. |
| Hand-rolled `RUsage` extern struct | `cb939a13` | ✅ Now `std.posix.getrusage(std.posix.rusage.SELF)` / `ru.maxrss`. |
| `DebugAllocator` contaminates bench timing (CodeRabbit) | `cb939a13` | ✅ → `std.heap.c_allocator`; 10k-fiber switch time drops to ~18ms (matches the doc's corrected figure). |
| Native-frame footprint test methodology (depth-60 comment vs `(level 10)`, CodeRabbit) | `cb939a13` | ✅ **Better than a fix** — you found the original test measured nothing (`for-each` is bootstrap Scheme, not a native primitive, so both arms ran identical bytecode), switched to `with-exception-handler`, and honestly re-labeled the result "inconclusive" instead of a false negative. Good scientific hygiene. |
| bench_reactor: validate I/O, preallocate `ready`, verify timer fired (CodeRabbit) | `de98d316` | ✅ All three done; Q3 `removeWaiter`-in-timed-region kept but now documented as a ~ns userspace caveat (matches my local finding — conclusion unaffected). |
| Doc: stray Q3 table row / stale "Follow-ups to file" / unreproducible server bench | `1a6af0fb` | ✅ Row removed; retitled "Follow-ups (filed)" with #1477–#1480 linked; numbers re-measured post-hardening. |
| CLAUDE.md missing new build steps | `a58451f5` | ✅ Both `bench-fibers`/`bench-reactor` documented. |

### One new (minor) item

The doc's reproducibility link — `kaappi-http/benchmarks/` at `.../tree/main/benchmarks` — points at `main`, but those files (`bench_server_app.scm`, `http_load_gen.py`, `README.md`) currently live only on **kaappi-http#3**, unmerged. The link will 404 until that PR lands. Suggest merging kaappi-http#3 to `kaappi-http`'s main before/with this one so the link resolves. Not a blocker for this repo's merge.

### Verdict

**Ready to merge**, pending the re-running CI matrix going green (the previous head `20316e45` was fully green and these six commits are hardening/test/doc changes only). Local verification all passed: `zig build test -Dtest-filter=1463` (normal + `-Dgc-stress=true`), and both benchmarks build and run. Recommend merging kaappi-http#3 alongside so the doc's repro link isn't dead.

PR #1481, opened 2026-07-12
Fix epoll stale-fire stranding of opposite-direction waiters

### LINKED ISSUE #1462: Fix epoll stale-fire stranding of opposite-direction waiters
## Summary

`Reactor.poll()` in `src/reactor.zig` has a narrow liveness gap on the **epoll backend**: a stale ONESHOT fire (an event arriving for a direction whose waiter list is empty) skips the re-arm step, leaving the fd permanently disarmed in the kernel even though the *other* direction still has a parked waiter. The stranded fiber never wakes; with no timers pending, the scheduler blocks in `epoll_wait` forever — a silent hang, not detectable by `isEmpty()` (the waiter is still listed, so it doesn't count as a deadlock).

## Scenario

1. fd X has a read waiter and a write waiter → `register()` arms the union, so the kernel holds `IN|OUT|ONESHOT`.
2. The read waiter is removed via `removeWaiter()` — the error-unwind `defer` in `waitForFd`:

https://github.com/kaappi/kaappi/blob/bff0886568d5832ef26229c103a96c5a71fda2c0/src/fiber.zig#L583-L592

   `removeWaiter` only edits the waiter lists; the kernel registration stays armed `IN|OUT|ONESHOT`.
3. fd X becomes readable. epoll fires `IN`, and **ONESHOT disarms the *whole* fd registration** — including the OUT interest (unlike kqueue, where read/write are independent knotes).
4. In `poll()`, `ev.readable` is true but `read_waiters` is empty, and `ev.writable` is false — so `fired` stays false and the loop `continue`s **without re-arming**, skipping the re-arm block that sits below it:

https://github.com/kaappi/kaappi/blob/bff0886568d5832ef26229c103a96c5a71fda2c0/src/reactor.zig#L228-L257

5. The still-parked write waiter is now waiting on an fd the kernel no longer watches. Nothing ever wakes it.

`removeWaiter`'s doc comment (`src/reactor.zig:143-145`) says a leftover kernel ONESHOT "fires once into the stale-event path of poll() and is dropped there" — it doesn't account for that drop also disarming the opposite direction on epoll. The comment should be updated as part of the fix.

## Impact

- Permanent silent hang on Linux: with no timers, the scheduler blocks in `epoll_wait` with an infinite timeout.
- Undetectable by `isEmpty()` (`src/reactor.zig:186`) — the write waiter is still listed, so deadlock detection sees pending I/O and keeps waiting.

## Unaffected backends

- **kqueue**: read/write are independent per-filter knotes; the untouched write knote stays armed. The redundant re-arm the fix adds is a harmless no-op `EV_ADD`.
- **WASI**: the current backend on `main` is the timer-only stopgap (never arms fds); the KEP-0001 P4 `WasiPollBackend` (#1461) clears only the fired direction in userspace, so remaining subscriptions survive a stale fire naturally.

## Fix direction

In `poll()`, compute `remaining_read`/`remaining_write` and re-arm whenever any waiters remain after processing an event — i.e. move the re-arm above the `if (!fired) continue;` (or make it unconditional after the drain), so the stale-fire path re-arms too:

```zig
// after draining matching waiters, regardless of `fired`:
const remaining_read = reg.read_waiters.items.len > 0;
const remaining_write = reg.write_waiters.items.len > 0;
if (remaining_read or remaining_write) {
    try self.backend.arm(ev.fd, remaining_read, remaining_write, false);
}
```

No backend changes needed: epoll's `arm()` already handles the dormant-after-ONESHOT registration (`CTL_MOD` with `ENOENT`→ADD / `EEXIST`→MOD self-heal, `src/reactor.zig:436`).

## Regression test

Add to `src/tests_reactor.zig`. Note: this needs a **socketpair**, not a pipe — a pipe end is unidirectional, and the scenario needs one fd carrying both a read and a write waiter that is readable-but-not-writable at the moment of the stale fire:

1. `socketpair`; fill `sv[0]`'s send buffer (small `SO_SNDBUF`, nonblocking writes until `EAGAIN`) so it is not writable.
2. `register(sv[0], .read, fiber_r)`, then `register(sv[0], .write, fiber_w)` → kernel armed `IN|OUT|ONESHOT`.
3. `removeWaiter(sv[0], fiber_r)`.
4. Write a byte to `sv[1]` → `sv[0]` readable → `poll()` once with a short timeout: the stale `IN` fire is dropped, `ready` stays empty.
5. Drain `sv[1]` so `sv[0]` becomes writable → `poll()` with a bounded timeout (e.g. 100 ms) and assert `fiber_w` lands in `ready`.

Without the fix, step 5 times out on Linux (fd disarmed). On macOS/kqueue the test passes either way (independent knotes), so demonstrating the failure needs a Linux run (`/linux-test` or CI).

---

Found in the KEP-0001 reactor (core landed in #1446, port layer in #1459). Code refs are at bff08865.

baijum:
## Summary

`Reactor.poll()` gated its re-arm step behind `if (!fired) continue;`, so a stale ONESHOT fire — an event arriving for a direction whose waiter list is already empty (e.g. after `removeWaiter`) — skipped re-arming entirely. On epoll, a ONESHOT fire disarms the *whole* fd registration, not just the direction that fired (unlike kqueue's independent per-direction knotes), so any waiter still parked on the opposite direction was permanently stranded on an fd the kernel no longer watches. With no timers pending, the scheduler would block forever in `epoll_wait` — a silent hang, undetectable by `isEmpty()` since the waiter is still listed.

Fix: re-arm unconditionally whenever waiters remain after the drain, regardless of whether this particular event matched a live waiter. Also updated `removeWaiter`'s doc comment, which didn't account for the stale drop also disarming the opposite direction on epoll.

## Test plan

- [x] Added the regression test from the issue's spec to `src/tests_reactor.zig`, using a socketpair with a shrunk `SO_SNDBUF` so one direction can be made unwritable while a read waiter is removed and fires stale.
- [x] `zig build test` passes (macOS/kqueue, native + `-Dgc-stress=true`).
- [x] `zig fmt --check` clean.
- [x] Verified on real Linux/epoll via cross-compiled unit tests run in a podman x86_64 container:
  - With the fix: all 18 reactor tests pass.
  - With the fix reverted: the new test fails exactly as predicted (`expected 1, found 0` — the write waiter never wakes), failing gracefully within the bounded 300ms timeout rather than hanging.
- [x] kqueue is immune to this bug (independent per-direction knotes), so the test is a no-op guard on macOS but a real regression catch on Linux/epoll — matches the issue's own note.

Fixes #1462.

### DIFF
--- a/src/reactor.zig
+++ b/src/reactor.zig
@@ -143,7 +143,10 @@ pub const Reactor = struct {
     /// the poll/close paths (an error unwinding waitForFd's scheduler
     /// drive) — those paths clear the lists themselves. A kernel ONESHOT
     /// left armed with no listed waiter fires once into the stale-event
-    /// path of poll() and is dropped there. No-op if absent.
+    /// path of poll() and is dropped there; on epoll that stale fire still
+    /// disarms the whole fd (not just the removed direction), so poll()
+    /// re-arms for whatever the other direction still needs regardless of
+    /// whether this event matched a live waiter (#1462). No-op if absent.
     pub fn removeWaiter(self: *Reactor, fd: i32, fiber: *Fiber) void {
         const reg = self.regs.getPtr(fd) orelse return;
         var lists = [_]*std.ArrayList(*Fiber){ &reg.read_waiters, &reg.write_waiters };
@@ -226,7 +229,6 @@ pub const Reactor = struct {
         for (events) |ev| {
             const reg = self.regs.getPtr(ev.fd) orelse continue; // stale event; already unregistered
 
-            var fired = false;
             // Reserved up front so the drain below can't fail mid-way: a
             // failure after the kernel's ONESHOT event was consumed but
             // before all waiters were moved to `ready` would strand the
@@ -237,20 +239,22 @@ pub const Reactor = struct {
             if (ev.readable and reg.read_waiters.items.len > 0) {
                 for (reg.read_waiters.items) |f| ready.appendAssumeCapacity(f);
                 reg.read_waiters.clearRetainingCapacity();
-                fired = true;
             }
             if (ev.writable and reg.write_waiters.items.len > 0) {
                 for (reg.write_waiters.items) |f| ready.appendAssumeCapacity(f);
                 reg.write_waiters.clearRetainingCapacity();
-                fired = true;
             }
-            if (!fired) continue;
 
             // epoll's ONESHOT disarms the *whole* fd registration on any
-            // fire, even a direction that didn't fire (unlike kqueue, where
-            // read/write are independent knotes and an untouched filter
-            // stays armed). Re-arm for whatever's left. Harmless no-op
-            // redundant EV_ADD on kqueue.
+            // fire — including a "stale" fire whose direction has no live
+            // waiter (e.g. removeWaiter already dropped it) — unlike
+            // kqueue, where read/write are independent knotes and an
+            // untouched filter stays armed. Re-arm unconditionally for
+            // whatever waiters remain, even when this event matched none:
+            // gating the re-arm on `fired` left a stale fire's untouched
+            // direction disarmed in the kernel with its waiter still
+            // parked and still listed, so no deadlock detector would catch
+            // it either (#1462). Harmless no-op redundant EV_ADD on kqueue.
             const remaining_read = reg.read_waiters.items.len > 0;
             const remaining_write = reg.write_waiters.items.len > 0;
             if (remaining_read or remaining_write) {
--- a/src/tests_reactor.zig
+++ b/src/tests_reactor.zig
@@ -417,3 +417,101 @@ test "closing the peer wakes a parked read waiter (EOF/HUP mapped to broken)" {
     try std.testing.expectEqual(@as(usize, 1), ready.items.len);
     try std.testing.expectEqual(&fiber_a, ready.items[0]);
 }
+
+fn setNonblocking(fd: std.c.fd_t) void {
+    const flags = std.c.fcntl(fd, std.posix.F.GETFL, @as(c_int, 0));
+    std.testing.expect(flags >= 0) catch unreachable;
+    const nonblock: c_int = @intCast(@as(u32, @bitCast(std.posix.O{ .NONBLOCK = true })));
+    std.testing.expect(std.c.fcntl(fd, std.posix.F.SETFL, flags | nonblock) >= 0) catch unreachable;
+}
+
+/// Shrinks the kernel send buffer so `fillSendBuffer` reaches EAGAIN after
+/// a few KB instead of needing megabytes of writes.
+fn setSmallSndbuf(fd: std.c.fd_t) void {
+    const size: c_int = 2048;
+    std.posix.setsockopt(fd, std.posix.SOL.SOCKET, std.posix.SO.SNDBUF, std.mem.asBytes(&size)) catch unreachable;
+}
+
+/// Writes to `fd` (already non-blocking) until the send buffer is full and
+/// a write returns EAGAIN, proving the fd is not writable. Bounded so a
+/// platform that doesn't honor the shrunk SO_SNDBUF fails loudly instead of
+/// spinning forever.
+fn fillSendBuffer(fd: std.c.fd_t) void {
+    var buf: [4096]u8 = [_]u8{0} ** 4096;
+    var iterations: usize = 0;
+    while (iterations < 4096) : (iterations += 1) {
+        const n = std.posix.system.write(fd, &buf, buf.len);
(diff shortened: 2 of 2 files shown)


src/tests_reactor.zig
```diff
@@ -417,3 +417,101 @@ test "closing the peer wakes a parked read waiter (EOF/HUP mapped to broken)" {
...
+    try reactor.register(pair[0], .write, &fiber_w); // kernel armed IN|OUT|ONESHOT
+
+    reactor.removeWaiter(pair[0], &fiber_r); // read waiter gone; kernel stays armed for IN
+
+    writeByte(pair[1], 'x'); // pair[0] becomes readable -> a stale IN fire
+
+    var ready = newReady();
+    defer ready.deinit(std.testing.allocator);
+    try reactor.poll(300_000_000, &ready); // the stale fire is dropped
+    try std.testing.expectEqual(@as(usize, 0), ready.items.len);
```
baijum, line 505: Nice setup — socketpair + shrunk `SO_SNDBUF` + fill-to-`EAGAIN` is the right way to keep `OUT` false while the stale `IN` fires. That matches the #1462 repro exactly.

One soft hardening nit (non-blocking): `ready.len == 0` after the first `poll` means no waiter woke, but it does not prove the stale `IN` was actually delivered and consumed. If that first `poll` timed out without an event (very unlikely for a local socketpair within 300 ms), the fd would still be armed and the second poll could wake `fiber_w` even without the fix — a theoretical false pass on epoll.

A `poll(0)` drain loop until an event is observed (or a short retry) would make the “stale fire happened” precondition explicit. Not required given the author’s podman fail-without-fix verification and ubuntu CI coverage.

src/reactor.zig
```diff
@@ -237,20 +239,22 @@ pub const Reactor = struct {
...
+            // waiter (e.g. removeWaiter already dropped it) — unlike
+            // kqueue, where read/write are independent knotes and an
+            // untouched filter stays armed. Re-arm unconditionally for
+            // whatever waiters remain, even when this event matched none:
+            // gating the re-arm on `fired` left a stale fire's untouched
+            // direction disarmed in the kernel with its waiter still
+            // parked and still listed, so no deadlock detector would catch
+            // it either (#1462). Harmless no-op redundant EV_ADD on kqueue.
             const remaining_read = reg.read_waiters.items.len > 0;
             const remaining_write = reg.write_waiters.items.len > 0;
```
baijum, line 259: This is the right minimal fix for epoll’s whole-fd ONESHOT disable semantics. Removing the `fired` gate means:

- stale fire + opposite waiters → re-arm surviving interest (the bug)
- live fire + opposite waiters → same as before
- fire drains all waiters / both lists empty → no re-arm (correct)
- kqueue → redundant `EV_ADD` on the remaining filter (harmless, as noted)

Also appreciate updating `removeWaiter`’s doc — the old “fires once and is dropped” story was incomplete for epoll. Leaving kernel interest for the removed direction is still the existing contract; proactive `MOD` in `removeWaiter` would be a separate hardening change, not needed here.

baijum (commented):
## Review

**Ready to merge** once ubuntu (epoll) CI is green.

### Verdict
Correct, minimal fix for the #1462 epoll stale-ONESHOT stranding bug. No blocking issues found. (Can't formally approve — this is my own PR — but I'd merge after Linux CI.)

### Correctness
`EPOLLONESHOT` disables the entire fd until `EPOLL_CTL_MOD`. Gating re-arm on `fired` skipped that MOD after a stale fire (direction already emptied by `removeWaiter`), permanently stranding any waiter still listed on the other direction — a hang invisible to `isEmpty()`. Re-arming whenever waiters remain after the drain is exactly the right invariant; kqueue/WASI stay safe (redundant/idempotent arm).

Edge cases checked:
- stale fire + opposite waiters → re-arm surviving interest
- live fire + opposite waiters → same as before
- fire drains all waiters / both lists empty → no re-arm
- kqueue → redundant `EV_ADD` (harmless)

### Test
Well designed (socketpair, not pipe; send-buffer fill so OUT is not ready during the stale IN; bounded timeouts). Real catch is Linux/epoll; macOS is a documented no-op guard. Podman fail-without-fix verification is strong evidence.

### Nits (non-blocking)
Left two inline comments: optional hardening of the first-poll “stale fire consumed” precondition, and a note that proactive kernel `MOD` in `removeWaiter` remains a possible follow-up, not a requirement for this PR.

### Merge checklist
- [x] Fix matches epoll ONESHOT + #1462 scenario
- [x] Empty / both-empty / live dual-interest paths still correct
- [x] Regression test targets the real failure mode
- [ ] Confirm ubuntu epoll CI green before merging

Issue #1468, opened 2026-07-12
Cross-thread wakeup: ThreadNotifier, unconditional sweep, shared-waiter registry (KEP-0002 P3)


baijum:
Part of #1465 (KEP-0002, Phase 3). The other half of the Phase 2 release. Implements the amended §5 — the sweep is **unconditional** (readiness-filtered sweeping is the machine-checked lost-wakeup of model finding 1, kaappi/keps#12).

## Scope

- `ThreadNotifier` + `Reactor.notify` ([KEP-0002 §5](https://github.com/kaappi/keps/blob/main/keps/0002-cross-thread-channels.md#5-cross-thread-wakeup)): `EVFILT.USER` (kqueue, `EV.ADD | EV.CLEAR`, `NOTE.TRIGGER`) / `eventfd` (epoll, non-ONESHOT, drained on wake) / WASI no-op. `notify` = `wake_pending` release-store, then ring the fd — always both.
- The normative consume protocol at both wake-check sites (`schedule()`'s per-tick check, `parkOnReactor` before blocking): `while (wake_pending.swap(false, .acq_rel)) sweepSharedWaiters();` then block; the fd covers the post-swap window.
- `sweepSharedWaiters` flips **every** fiber on the per-scheduler shared-waiter registry — no per-channel readiness filter; woken fibers retry, losers re-park and re-register under the lock (that re-registration is what the snapshot-and-clear ring consumed).
- §4 registration under one continuous lock hold; snapshot-and-clear rings after unlock; §7 waiter-list lifecycle: dedup one entry per notifier per list, per-entry notifier refcount, opportunistic pruning on `alive == false`; notifier teardown at thread exit (`alive = false`, ring becomes no-op, entries pruned lazily, UAF-free via the refcount).
- Refcount-aware deadlock semantics: a wait on a shared channel blocks in `poll` (never raises) when `refcount > 1` or other live threads exist; main-fiber fallback blocks on the notifier. `hasRunnableFibers`/`Reactor.isEmpty` extended.
[...]

Issue #1467, opened 2026-07-12
Envelopes at thread boundaries: thread-start!/join via envelopes, live-thread counter (KEP-0002 P2)


baijum:
Part of #1465 (KEP-0002, Phase 2). Ships in the same release as Phase 3: after this phase a channel crosses threads, but a receiver that parks before the send hangs until the Phase 3 wakeup machinery lands.

## Scope

- `thread-start!` copies the thunk into an envelope **on the parent thread, before `Thread.spawn`** ([KEP-0002 §3](https://github.com/kaappi/keps/blob/main/keps/0002-cross-thread-channels.md)); the child copies out into its fresh heap. Closes the Motivation's concurrent-copy race (child copying out of a live parent heap) and gives the thunk snapshot at-the-call semantics, atomic w.r.t. all Scheme code.
- Channels captured in thunks are promoted by the owner during that parent-side copy — the only legal place.
- `thread-join!` result and exception cross by envelope; retire the bespoke `child_registry.storeResult` deep-copy-at-join special case.
- Process-global live-thread counter (atomic), maintained by `thread-start!`/thread exit — §5's deadlock heuristic reads it in Phase 3.
- First real cross-thread channel tests, restricted to send-completes-before-receiver-parks interleavings (no wakeup needed).

## Acceptance

- Thunk-snapshot test: mutations after `thread-start!` returns are not visible to the child; captured-channel identity survives (reply-to works once Phase 3 lands).
- Thread-churn tests under `-Dgc-stress=true`: envelopes queued by a dead thread survive its heap teardown; no leaks at exit.
- `thread-start!` overhead delta vs. baseline measured and recorded (one extra thunk copy — expected small; Drawbacks price it).
- Motivation Path 1 (channel in thunk) now works; Motivation Path 2 (channel via shared global from child) raises the descriptive foreign-owner error while the object is live.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

baijum:
Implemented and merged via #1483 (Envelopes at thread boundaries: thread-start!/join! via envelopes, live-thread counter), building on Phase 1's #1482. Closing as complete.

Issue #1466, opened 2026-07-12
SharedChannel core: shared-object protocol, promotion, reservation send/receive (KEP-0002 P1)


baijum:
Part of #1465 (KEP-0002, Phase 1). Everything testable on one thread; no cross-thread behavior yet. **Merge gate:** the TLA+ suite ([keps `research/tla/run.sh`](https://github.com/kaappi/keps/tree/main/research/tla)) stays green — any §4–§6 pseudocode change is a KEP amendment re-run through the model *before* code changes.

## Scope

- New `src/shared_object.zig` — the generic shared-object protocol ([KEP-0002 §1](https://github.com/kaappi/keps/blob/main/keps/0002-cross-thread-channels.md)): refcounted header allocated from `std.heap.c_allocator`, stub alloc/release helpers, leak-check hook (an undestroyed shared object at process exit fails the unit suite). `SharedChannel` is the first instance; KEP-0003's `SharedBuffer` is the declared second.
- New `src/shared_channel.zig` — `SharedChannel` (mutex, intrusive envelope FIFO, `queue_len`/`reserved`/`capacity`/`closed`, waiter lists) and `Envelope` (per-message GC mini-heap; sender `deepCopy` in, receiver `deepCopy` out, `deinit` frees the graph wholesale). Full §1 refcount state machine: promote ⇒ rc 1; every stub +1 (including envelope-heap stubs); `freeObject` −1; zero destroys (drain + deinit queued envelopes).
- `types.Channel.shared: ?*SharedChannel` field; **owner-side promotion** (§2): publish-before-drain (re-entrancy: a queued message may contain the channel itself), FIFO drain into envelopes, local-waiter migration bookkeeping (exercised for real in Phase 3).
- `deepCopy` `.channel` arm: `UncopyableType` → promote-if-needed + alias stub + rc+1.
- §4 send/receive on the shared representation, per the **amended** text: slot reservation before the copy; copy-failure path releases the reservation and rings `send_waiters` — and `recv_waiters` when closed; push is infallible; receive pops before EOF, EOF requires `closed ∧ drained ∧ reserved == 0`; **receive-side copy failure re-queues the envelope at the head and rings `recv_waiters`** (bounded queues may transiently overshoot capacity — documented, admission stays strict).
- Foreign-owner check on every channel primitive: `"channel belongs to another thread; pass it through the thread thunk to share it"`.
[...]

Issue #1465, opened 2026-07-12
Epic: cross-thread channels and shared flat numeric data (KEP-0002 + KEP-0003)


baijum:
Tracking epic for implementing [KEP-0002: Cross-Thread Channels and Multi-Core Fiber Scheduling](https://github.com/kaappi/keps/blob/main/keps/0002-cross-thread-channels.md) and its gated sibling [KEP-0003: Shared Flat Numeric Data](https://github.com/kaappi/keps/blob/main/keps/0003-shared-flat-numeric-data.md). KEP-0002's pre-implementation verification gate is satisfied (see Verification below); KEP-0003 stays gated on Phase 7 evidence by its own acceptance rule.

## Problem

Fibers give cheap cooperative concurrency on one thread and one heap; SRFI-18 threads give true parallelism with fully isolated heaps — and the only data path between running threads is the thunk copy at start and the result copy at join. Worker pools, pipelines, and multi-core servers are inexpressible, and a channel reached through a shared global is silent memory corruption today (KEP-0002 Motivation, verified at `54706a0c`: a child-heap pair spliced into a parent-heap channel reads back as garbage after the child heap is freed).

## Design

A channel that crosses a thread boundary is transparently **promoted**: its queue moves into a mutex-protected, refcounted `SharedChannel` outside every GC heap, and every message crosses as an **envelope** (a private mini-heap filled by the audited `deepCopy`) — copied in on send, copied out on receive. Two copies per message is the deliberate price that keeps every heap single-threaded and all 601 primitives lock-free. Cross-thread sends wake remote parked fibers by ringing the target thread's reactor (`EVFILT.USER` / `eventfd`) — the trigger KEP-0001 resolved question 4 named. On top: `(kaappi parallel)` (pools, `parallel-map`) in pure Scheme, and `SO_REUSEPORT` multi-core HTTP serving. Fibers never migrate; multi-core means N independent schedulers connected by channels.

KEP-0003 then adds the one narrow carve-out copy semantics cannot express: **shared flat numeric buffers** (`shared-bytevector`, `shared-f64vector` — raw bytes/doubles, never Scheme references) that cross by refcounted reference on the same shared-object protocol, with all ordering delegated to channel happens-before edges.

## Sub-issues

Sequencing: Phases 1–4 are the critical path, in order; Phases 2 and 3 ship in the same release (a receiver that parks before the sender exists hangs until Phase 3's wakeup lands).

| Sub-issue | Scope | Depends on |
[...]

PR #1482, opened 2026-07-12
SharedChannel core: shared-object protocol, promotion, reservation send/receive (KEP-0002 P1)

baijum:
## Summary

Implements KEP-0002 Phase 1 (#1466), part of the cross-thread channels epic (#1465): the generic refcounted shared-object protocol and its first instance, `SharedChannel` — everything testable on a single OS thread, ahead of Phase 2 (envelopes at thread boundaries, #1467) and Phase 3 (cross-thread wakeup, #1468).

- `src/shared_object.zig` (new): refcounted header allocated from `std.heap.c_allocator`, `init`/`retain`/`release`, and a `liveCount()` leak-check hook.
- `src/shared_channel.zig` (new): `SharedChannel` (mutex, intrusive envelope FIFO, `reserved`/`capacity`/`closed`, waiter lists), `Envelope` (private mini-heap; sender `deepCopy` in, receiver `deepCopy` out), `promoteChannel` (owner-side, publish-before-drain so re-entrant self-sends alias instead of double-promoting), and the full §4 send/receive sequences (slot reservation, build-outside-the-lock, the copy-failure paths that keep "send fails ⇒ nothing sent" and "receive fails ⇒ nothing received, FIFO preserved").
- `types.Channel` gains a `shared: ?*anyopaque` field (kept opaque so the dependency-free `types.zig` doesn't import a feature module).
- `deepCopy`'s `.channel` arm promotes an owned, unpromoted channel (or aliases an already-shared one) instead of raising `UncopyableType` unconditionally; `freeObject` releases a stub's refcount on every teardown path (GC sweep, test `deinit()`, or an envelope's own mini-heap tearing itself down).
- `channel-send`/`channel-receive` gain a foreign-owner check — this turns the KEP's Motivation Path 2 (a channel reached through a shared global silently corrupting memory) into a descriptive, catchable error.
- `src/bench_channel.zig` + `zig build bench-channel`: local fast-path regression benchmark + a single-thread envelope-cost harness over five payload shapes, the Phase 1 installment of the KEPs repo's P3 research plan.

**Scope boundary:** `thread-start!`/`thread-join!` are unchanged. Their `deepCopy` calls run on the *destination* thread (child for the thunk copy, parent for the join result), never the channel's actual owner, so promotion isn't legally reachable through them yet — that's Phase 2's job (moving the thunk copy to the parent thread before spawn). A channel captured in a thread thunk still raises today's `"thread thunk contains uncopyable type"` error, unchanged.

## Test plan

- [x] `zig build test` — full unit suite green, including the new `tests_shared_channel.zig` (promote empty/non-empty queue, deepCopy promote+alias with channel-in-channel identity, foreign-owner rejection, re-entrant self-promotion and its documented refcount-cycle leak, recursive refcount teardown across two channels, send/receive success + failure paths with FIFO preservation, waiter registration/dedup, and a real-VM/real-thread regression test for Motivation Path 2)
- [x] `zig build test -Dgc-stress=true` — green
- [x] `bash tests/scheme/run-all.sh` + `zig build run -- tests/scheme/r7rs/r7rs-tests.scm` — 1834 pass, 0 fail (behaviorally invisible to existing single-threaded programs)
- [x] `zig build wasm` — still compiles (`std.Thread`-based locking degrades on the single-threaded wasm32 target; `thread-start!` is already `is_wasm`-gated before ever reaching `deepCopy`)
- [x] `zig build bench-channel -Doptimize=ReleaseFast` before/after: local unpromoted send+receive ~108ns/op baseline vs. ~110ns/op on this branch (within noise) — satisfies invariant 3's "unmeasurable" gate
- [x] Real x86-64 Linux (DigitalOcean droplet, destroyed after): build OK, unit tests OK, full Scheme/R7RS suite 1834/0
- [x] KEP merge gate: `research/tla/run.sh` in `kaappi/keps@8aa2f1fc` — all 9 configs match their expected outcome (6 pass, 3 intentional regression-witness failures)

### DIFF
--- a/src/gc_deep_copy.zig
+++ b/src/gc_deep_copy.zig
@@ -2,18 +2,27 @@ const std = @import("std");
 const types = @import("types.zig");
 const memory_mod = @import("memory.zig");
 const hashtable = @import("primitives_hashtable.zig");
+const shared_channel = @import("shared_channel.zig");
 
 const GC = memory_mod.GC;
 const Value = types.Value;
 
-pub fn deepCopy(gc: *GC, src: Value) !Value {
+/// Explicit (not inferred) so shared_channel.zig's Envelope.create ->
+/// GC.deepCopy -> gc_deep_copy.deepCopy -> deepCopyValue ->
+/// shared_channel.promoteChannel -> Envelope.create call chain doesn't form
+/// an unresolvable inferred-error-set cycle: promoteChannel/Envelope.create
+/// need to know deepCopy's error set before deepCopy's own inference (which
+/// recurses into promoteChannel for the .channel arm) can complete.
+pub const DeepCopyError = error{ OutOfMemory, UncopyableType };
+
+pub fn deepCopy(gc: *GC, src: Value) DeepCopyError!Value {
     if (!types.isPointer(src)) return src;
     var visited = std.AutoHashMap(usize, Value).init(gc.allocator);
     defer visited.deinit();
     return deepCopyValue(gc, src, &visited);
 }
 
-fn deepCopyValue(gc: *GC, src: Value, visited: *std.AutoHashMap(usize, Value)) !Value {
+fn deepCopyValue(gc: *GC, src: Value, visited: *std.AutoHashMap(usize, Value)) DeepCopyError!Value {
     if (!types.isPointer(src)) return src;
 
     const src_ptr = @intFromPtr(types.toObject(src));
@@ -325,10 +334,39 @@ fn deepCopyValue(gc: *GC, src: Value, visited: *std.AutoHashMap(usize, Value)) !
             new_rs.prng = rs.prng;
             return new_val;
         },
+        // KEP-0002 §2: a channel is promoted (if not already) and aliased,
+        // never rejected outright. `gc` here is the *destination* heap, not
+        // the channel's owner -- ownership (invariant 4: promotion is legal
+        // only for the thread that owns the channel) is checked against
+        // `memory_mod.gc_instance`, the threadlocal for whichever thread is
+        // actually executing this copy. A foreign, not-yet-promoted channel
+        // reuses UncopyableType rather than a new error class, so
+        // thread-start!/thread-join! (primitives_srfi18.zig) -- whose
+        // deepCopy calls run on the destination thread, never the channel's
+        // owner, until KEP-0002 Phase 2 moves the thunk copy to the parent
+        // thread -- keep today's exact behavior and error message.
+        .channel => {
+            const ch = obj.as(types.Channel);
+            // Aliasing an already-promoted stub needs no ownership check
+            // (and no memory_mod.gc_instance at all) -- any thread that
+            // legitimately holds a stub Value may pass it along further.
+            // Only a not-yet-promoted channel needs the current-thread
+            // check, since promoting is the operation invariant 4 restricts.
+            const sc = if (ch.shared) |s|
+                @as(*shared_channel.SharedChannel, @ptrCast(@alignCast(s)))
+            else blk: {
+                const current_gc = memory_mod.gc_instance orelse return error.UncopyableType;
+                if (obj.owner != current_gc.id) return error.UncopyableType;
+                break :blk try shared_channel.promoteChannel(current_gc, ch);
+            };
+            sc.retain();
+            const new_val = try gc.allocChannelStub(sc);
+            try visited.put(src_ptr, new_val);
+            return new_val;
+        },
         .port,
         .continuation,
         .fiber,
-        .channel,
         .mutex,
         .condition_variable,
         .ffi_callback,
--- a/src/gc_deep_copy.zig
+++ b/src/gc_deep_copy.zig
@@ -359,8 +359,12 @@ fn deepCopyValue(gc: *GC, src: Value, visited: *std.AutoHashMap(usize, Value)) D
                 if (obj.owner != current_gc.id) return error.UncopyableType;
                 break :blk try shared_channel.promoteChannel(current_gc, ch);
             };
-            sc.retain();
+            // Allocate before retaining: if allocChannelStub fails, no new
+            // stub exists to own the +1, and the source stub's own count
+            // (still held regardless) keeps the retain/release balance
+            // intact -- retaining first would leak sc's refcount on OOM.
             const new_val = try gc.allocChannelStub(sc);
+            sc.retain();
             try visited.put(src_ptr, new_val);
             return new_val;
         },
--- a/src/primitives_fiber.zig
+++ b/src/primitives_fiber.zig
@@ -110,7 +110,7 @@ fn channelSendFn(args: []const Value) PrimitiveError!Value {
     // shared global, promoted or not -- is what silently corrupted memory
     // before this check existed (Motivation Path 2); now it's a diagnosis.
     if (ch_obj.owner != gc.id)
-        return raiseChannelError("channel belongs to another thread; pass it through the thread thunk to share it");
+        return raiseFiberError("channel belongs to another thread; pass it through the thread thunk to share it");
     const ch = ch_obj.as(types.Channel);
 
     if (ch.shared) |raw| {
@@ -123,8 +123,11 @@ fn channelSendFn(args: []const Value) PrimitiveError!Value {
             // Phase 1 channels are always unbounded and open -- no
             // Scheme-level API sets capacity or closed yet (KEP-0002
             // Phase 4) -- so these branches are unreachable through
-            // make-channel's output.
-            .would_park, .closed => unreachable,
+            // make-channel's output. @panic (not `unreachable`, which is UB
+            // under ReleaseFast) so a Phase 4 gap that forgets to wire this
+            // switch fails loudly in every build mode instead of silently
+            // corrupting execution.
+            .would_park, .closed => @panic("channel-send: reached a shared-channel branch Phase 1's primitives don't wire up yet (capacity/close is Phase 4)"),
         };
     }
 
@@ -161,7 +164,7 @@ fn channelReceiveFn(args: []const Value) PrimitiveError!Value {
     const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
     const ch_obj = types.toObject(args[0]);
     if (ch_obj.owner != gc.id)
-        return raiseChannelError("channel belongs to another thread; pass it through the thread thunk to share it");
+        return raiseFiberError("channel belongs to another thread; pass it through the thread thunk to share it");
     const ch = ch_obj.as(types.Channel);
 
     if (ch.shared) |raw| {
@@ -173,8 +176,9 @@ fn channelReceiveFn(args: []const Value) PrimitiveError!Value {
             .value => |v| v,
             // Phase 1 channels are always open (channel-close! is Phase 4)
             // and fiber-parking isn't wired to the shared path yet (Phase
-            // 3) -- unreachable through make-channel's output.
-            .would_park, .eof => unreachable,
+            // 3) -- unreachable through make-channel's output. @panic, not
+            // `unreachable` (UB under ReleaseFast) -- see channelSendFn.
+            .would_park, .eof => @panic("channel-receive: reached a shared-channel branch Phase 1's primitives don't wire up yet (capacity/close is Phase 4)"),
         };
     }
 
@@ -185,7 +189,7 @@ fn channelReceiveFn(args: []const Value) PrimitiveError!Value {
     const vm = vm_mod.vm_instance orelse return PrimitiveError.OutOfMemory;
     if (vm.scheduler == null) {
         // No fibers exist, so nothing can ever send: blocking would hang.
-        return raiseDeadlockError("channel-receive: deadlock — channel is empty and no fibers are running");
+        return raiseFiberError("channel-receive: deadlock — channel is empty and no fibers are running");
     }
 
     // Capture before the recursive dispatch below: args is a slice into
@@ -229,28 +233,15 @@ fn blockOrDeadlock(vm: *vm_mod.VM, me: *fiber_mod.Fiber, my_idx: usize, wait_on:
         vm.yield_retry = true;
         return PrimitiveError.Yielded;
     }
-    return raiseDeadlockError(deadlock_msg);
+    return raiseFiberError(deadlock_msg);
 }
 
-fn raiseDeadlockError(msg: []const u8) PrimitiveError {
-    const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
-    const vm = vm_mod.vm_instance orelse return PrimitiveError.OutOfMemory;
-    const message = gc.allocString(msg) catch return PrimitiveError.OutOfMemory;
-    var msg_root = message;
-    gc.pushRoot(&msg_root);
-    const err_val = gc.allocErrorObject(msg_root, types.NIL) catch {
-        gc.popRoot();
-        return PrimitiveError.OutOfMemory;
-    };
-    gc.popRoot();
-    vm.current_exception = err_val;
-    return PrimitiveError.ExceptionRaised;
-}
-
-/// Same shape as raiseDeadlockError, kept separate since the message isn't
-/// about deadlocks: the foreign-owner check (KEP-0002 §2) and a shared-path
-/// uncopyable payload both need this.
-fn raiseChannelError(msg: []const u8) PrimitiveError {
+/// Raises a generic ErrorObject carrying `msg`. Named for this file's scope
+/// (fiber/channel primitives), not any one caller: deadlocks, the
+/// foreign-owner check (KEP-0002 §2), and a shared-path uncopyable payload
+/// all just need a message wrapped into a catchable exception -- the
+/// function name never reaches the user, only `msg` does.
+fn raiseFiberError(msg: []const u8) PrimitiveError {
     const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
     const vm = vm_mod.vm_instance orelse return PrimitiveError.OutOfMemory;
     const message = gc.allocString(msg) catch return PrimitiveError.OutOfMemory;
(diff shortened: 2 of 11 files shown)


src/shared_channel.zig
```diff
@@ -0,0 +1,353 @@
...
+        // a real collect() on a heap that no other GC's root marker can see.
+        gc.enabled = false;
+        errdefer {
+            gc.deinit();
+            std.heap.c_allocator.destroy(gc);
+        }
+
+        const env = try std.heap.c_allocator.create(Envelope);
+        env.* = .{ .gc = gc };
+        env.value = try gc.deepCopy(payload);
```
baijum, line 73: **Bug (small leak): the `errdefer` covers `gc` but not `env`.** When `gc.deepCopy(payload)` fails, the `Envelope` struct itself is never destroyed — contradicting this function's doc comment ("tears down everything it allocated"). The path is live in Phase 1: it's exactly what the *"send: build failure"* test exercises, so every failed send leaks one `Envelope`. `std.testing`'s leak detection can't see it because this is `c_allocator`, not the tracked test allocator.

```zig
const env = try std.heap.c_allocator.create(Envelope);
errdefer std.heap.c_allocator.destroy(env);
```

src/gc_deep_copy.zig
```diff
@@ -325,10 +334,39 @@ fn deepCopyValue(gc: *GC, src: Value, visited: *std.AutoHashMap(usize, Value)) !
...
+            // check, since promoting is the operation invariant 4 restricts.
+            const sc = if (ch.shared) |s|
+                @as(*shared_channel.SharedChannel, @ptrCast(@alignCast(s)))
+            else blk: {
+                const current_gc = memory_mod.gc_instance orelse return error.UncopyableType;
+                if (obj.owner != current_gc.id) return error.UncopyableType;
+                break :blk try shared_channel.promoteChannel(current_gc, ch);
+            };
+            sc.retain();
+            const new_val = try gc.allocChannelStub(sc);
```
baijum, line 366: **Bug (refcount leak on OOM): `sc.retain()` has already run when `allocChannelStub` fails**, and the error propagates with no matching release — the `SharedChannel` (plus everything in its queue) can then never reach zero, violating §1's "every reference is a counted stub" accounting. The suite's `liveCount()` baseline checks would flag it, but no test OOMs on this specific allocation.

```zig
sc.retain();
const new_val = gc.allocChannelStub(sc) catch |err| {
    sc.release();
    return err;
};
```

The unwind `release()` can't hit zero here: the source stub still holds its own count, so rc ≥ 2 at this point. (A `visited.put` failure after this point is fine as-is — the stub is already GC-tracked and releases through `freeObject` at teardown.)

src/shared_channel.zig
```diff
@@ -0,0 +1,353 @@
...
+    next: ?*Envelope = null,
+
+    /// Builds a private mini-heap and deep-copies `payload` into it.
+    /// Atomic from the caller's perspective: either this returns a fully
+    /// built envelope, or it tears down everything it allocated and
+    /// propagates deepCopy's error -- callers never have to decide whether
+    /// a half-built envelope needs cleanup.
+    pub fn create(payload: Value) !*Envelope {
+        const gc = try std.heap.c_allocator.create(memory.GC);
+        gc.* = memory.GC.init(std.heap.c_allocator);
```
baijum, line 59: **Deliberate deviation from KEP-0002 §1 — please document it here and sync the KEP.** The KEP says the envelope GC is created "the same way `GC.initForThread` does (sharing the process-wide symbol table)"; this uses plain `GC.init`, i.e. a private per-envelope symbol table (symbols in a message get copied, not aliased).

I believe the private table is the **right** call: an envelope must be hermetic because it can outlive the thread that built it — sender exits and its heap is torn down while the message is still queued; the receiver (or `destroyHook`, on whichever thread drops the last refcount) reads the graph later. With a shared table, the envelope graph would hold parent-owned `Symbol` objects that may already have been freed by then — a use-after-free. It also avoids growing the parent's `foreign_symbols` list per message and taking `symbol_mutex` per symbol.

But right now the deviation is silent: someone later "fixing" this to match the KEP text would introduce exactly that use-after-free. One sentence here plus a KEP §1 update would pin it.

src/primitives_fiber.zig
```diff
@@ -103,15 +104,37 @@ fn channelSendFn(args: []const Value) PrimitiveError!Value {
...
+        const outcome = shared_channel.send(sc, args[1], null) catch |err| {
+            return translateSharedChannelError(err, "channel-send");
+        };
+        return switch (outcome) {
+            .sent => types.VOID,
+            // Phase 1 channels are always unbounded and open -- no
+            // Scheme-level API sets capacity or closed yet (KEP-0002
+            // Phase 4) -- so these branches are unreachable through
+            // make-channel's output.
+            .would_park, .closed => unreachable,
```
baijum: Minor: `unreachable` panics under ReleaseSafe but is **UB under ReleaseFast**. If Phase 4 wires up capacity/close and forgets these two switches, a full/closed channel becomes undefined behavior in `-Doptimize=ReleaseFast` builds instead of a loud failure. This isn't a hot path — a defensive raise (or `@panic("channel-send: Phase 4 capacity/close reached before primitive wiring")`) fails loudly in every build mode. Same applies to the `.would_park, .eof` arm in `channelReceiveFn`.

src/primitives_fiber.zig
```diff
@@ -103,15 +104,37 @@ fn channelSendFn(args: []const Value) PrimitiveError!Value {
...
 
     const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
-    const ch = types.toObject(args[0]).as(types.Channel);
+    const ch_obj = types.toObject(args[0]);
+    // KEP-0002 §2: the only legal cross-thread handle is a locally owned
+    // stub created by deepCopy. A foreign object -- reached through a
+    // shared global, promoted or not -- is what silently corrupted memory
+    // before this check existed (Motivation Path 2); now it's a diagnosis.
+    if (ch_obj.owner != gc.id)
+        return raiseChannelError("channel belongs to another thread; pass it through the thread thunk to share it");
```
baijum: Minor wording: "pass it through the thread thunk to share it" is advice that doesn't work yet — in Phase 1 a channel in a thread thunk still raises `thread thunk contains uncopyable type` (promotion is only reachable in Phase 2, as the PR description itself notes). Fine if Phase 2 lands before the next release; otherwise a user following the message just hits a second error. Consider deferring the fix-suggestion half of the message to the Phase 2 PR.

src/primitives_fiber.zig
```diff
@@ -206,6 +247,42 @@ fn raiseDeadlockError(msg: []const u8) PrimitiveError {
     return PrimitiveError.ExceptionRaised;
 }
 
+/// Same shape as raiseDeadlockError, kept separate since the message isn't
+/// about deadlocks: the foreign-owner check (KEP-0002 §2) and a shared-path
+/// uncopyable payload both need this.
+fn raiseChannelError(msg: []const u8) PrimitiveError {
```
baijum: Nit: byte-identical to `raiseDeadlockError` above. The "message isn't about deadlocks" concern could be served by a single neutrally named helper (`raiseFiberError(msg)`?) that both call sites share — the function name never reaches the user, only the message does.

src/tests_shared_channel.zig
```diff
@@ -0,0 +1,449 @@
...
+    // Known limitation (§1): the envelope holds a stub of `sc` itself, so
+    // dropping ch's own stub can never bring refcount to zero -- the last
+    // reference is the stub inside sc's own queue. This is the accepted
+    // refcount-cycle leak, not a bug: assert it is *present* and stable, so
+    // this test breaks loudly if the leak's behavior ever silently changes.
+    gc1.deinit();
+    try std.testing.expectEqual(baseline + 1, shared_object.liveCount());
+}
+
+test "refcount teardown: a channel-in-channel message drains and destroys recursively when never received" {
```
baijum, line 201: Coverage suggestion: this covers the *never-received* half of the KEP §1 worked example; the *received* half — the reply-to pattern, arguably the KEP's marquee use case — has no test. Something like: send a local channel B over shared channel A, `receive` on a different heap, assert the copied-out stub aliases the same `SharedChannel` (rc 1→2→3→2 per the KEP's worked example), then tear everything down to baseline. That is also the one path where the alias arm runs against an **envelope-owned** stub (`obj.owner` = envelope GC's id ≠ `gc_instance`) — i.e. the very reason the alias path deliberately skips the ownership check — so it deserves an assertion pinning it.

Relatedly: nothing yet executes the `if (ch.shared)` branches in `channelSendFn`/`channelReceiveFn` (Scheme can't reach promotion in Phase 1, and these tests call `shared_channel.send`/`receive` directly). A white-box test that promotes manually and then evals `(channel-send ch 1)` / `(channel-receive ch)` would cover the primitive dispatch plus `translateSharedChannelError`, and pin the contract Phases 2/3 build on.

src/tests_shared_channel.zig
```diff
@@ -0,0 +1,449 @@
...
+
+    const ch_val = try gc1.allocChannel();
+    const ch = types.toObject(ch_val).as(types.Channel);
+    try std.testing.expect(ch.shared == null);
+
+    const saved_gc = memory.gc_instance;
+    memory.gc_instance = &gc1; // gc1 is the calling thread's own gc -- legal to promote
+    const stub2 = try gc2.deepCopy(ch_val);
+    const stub3 = try gc3.deepCopy(ch_val);
+    memory.gc_instance = saved_gc;
```
baijum: Test hygiene nit: restore via `defer memory.gc_instance = saved_gc;` (as the foreign-rejection test below already does) — a failed `try` on either `deepCopy` above returns before this line and leaves `gc_instance` pointing at this test's dead stack `GC` for every subsequent test in the process, turning one failure into a confusing cascade. Same pattern in the "refcount teardown" test (its two plain restores).

baijum (commented):
## Review — KEP-0002 Phase 1 (SharedChannel core)

**Verdict: two small error-path leaks to fix before merge; everything else is minor or optional.** The core design is faithful to the KEP and carefully engineered — after the two one-line fixes below and green CI, this is ready to merge.

### What holds up well

- The §4 send/receive sequences match the KEP verbatim, including both failure paths ("send fails ⇒ nothing sent"; "receive fails ⇒ nothing received" with head re-queue), the reservation discipline, the closed-channel recv ring on the send failure path, and snapshot-under-lock / ring-after-unlock.
- The capacity check uses `queue_len + sc.reserved >= cap` where the KEP pseudocode says `==` — that is the **correct** strengthening: §4's receive-failure note documents transient capacity overshoot, which would make `==` skip the park.
- Publish-before-drain in `promoteChannel` (re-entrant self-send aliasing) is subtle and right, with a dedicated test that asserts the documented refcount-cycle leak is *present and stable* rather than pretending it doesn't exist — nice.
- The scope boundary is honest and I verified it: the thunk copy at `primitives_srfi18.zig:363` runs on the **child** thread with `gc_instance = child_gc`, so the new deepCopy ownership check preserves today's exact `thread thunk contains uncopyable type` behavior, and promotion really is unreachable from Scheme in Phase 1.
- The `promoteChannel` drain never allocates on the owner's heap (envelopes, stubs, and `SharedChannel` all go to the envelope GC / c_allocator), so walking the unrooted local queue after clearing `head`/`tail` is safe without rooting. Subtle, and sound.
- Motivation Path 2 goes from silent memory corruption to a caught, descriptive error, with a real-VM/real-thread regression test.

### Verification I ran (at 765c026c)

- `zig build test` — full unit suite green.
- `zig build test -Dgc-stress=true -Dtest-filter=channel` — all channel tests green under collection-per-allocation.
- Cross-checked the implementation against KEP-0002 §§1–4/§7 in `kaappi/keps`.

### Findings

- **2 bugs** (inline): the `Envelope` struct leak on copy failure in `Envelope.create`, and the `sc.retain()` refcount leak when `allocChannelStub` fails. Both are error-path-only and one-liners to fix.
- **Minor/optional** (inline): document the deliberate KEP §1 symbol-table deviation (it's correct, but silently so), the Phase-4 `unreachable` footgun under ReleaseFast, the Phase-2-only advice in the foreign-owner error message, a helper duplication nit, and two test suggestions (the received-side reply-to round trip; the primitives' `ch.shared` dispatch branches).
- One KEP-sync note that needs no code change: KEP §2 step 2 says promotion "publishes the pointer (release store)" — the implementation uses a plain store, which is fine in Phase 1 (every read is same-thread or behind a synchronized handoff), but Phase 3 should either make it atomic or update the KEP wording so spec and code agree.

CI was still pending when I reviewed; it needs to be green before merge as usual.

baijum:
Pushed 21f5aec addressing the review. Summary:

**The two bugs (confirmed independently by both the human review and CodeRabbit):**
- `Envelope.create`'s `errdefer` only covered the mini GC, not the `Envelope` struct itself — every failed send (e.g. an uncopyable payload) leaked one `Envelope`. Fixed with `errdefer std.heap.c_allocator.destroy(env);` right after allocation.
- `gc_deep_copy.zig`'s `.channel` arm called `sc.retain()` *before* `gc.allocChannelStub(sc)` — an OOM on the allocation left the retain unmatched, pinning the `SharedChannel` forever. Reordered: allocate first, retain only once the new stub exists to own it.

**CodeRabbit's additional "Major" finding (`shared_channel.zig` receive park path):** confirmed real and fixed at the source rather than just in `primitives_fiber.zig` — `send`/`receive`'s park branches now skip waiter registration when `notifier` is `null` instead of force-unwrapping it. Not reachable through the Scheme-level API today (Phase 1 channels are always unbounded/open), but it removes a real landmine for whatever calls this next.

**Everything else you flagged as minor/optional, done:**
- `unreachable` → `@panic` in the two Phase-4-only switch arms (UB under ReleaseFast vs. loud failure in every build mode).
- Documented the private-per-envelope-symbol-table deviation directly on `Envelope.create`, with the use-after-free reasoning for why it's the right call.
- Merged `raiseChannelError`/`raiseDeadlockError` into one `raiseFiberError`.
- Two new tests: the *received* half of the §1 reply-to worked example (rc 1→2→3→2, identity across the round trip, and — per your note — it's the one path where the alias arm runs against an envelope-owned stub whose `owner` isn't `gc_instance`), and a white-box test that promotes a channel and drives `channel-send`/`channel-receive` through `vm.eval` so `primitives_fiber.zig`'s `ch.shared` dispatch actually gets exercised (previously only `shared_channel.send`/`.receive` were called directly).
- Test hygiene: `defer` for every `memory.gc_instance` restore; swapped to `th.TestContext` in the Motivation Path 2 test.
- `bench_channel.zig`: bigger source buffer, `iters` as a `Workload` field.

**Bonus catch while adding the reply-to test:** it crashed under `-Dgc-stress=true` — a test bug, not a library one. The "tasks" channel's `Value` wasn't rooted across the test's second allocation on the same GC, so a stress collection swept it and `freeObject` released+destroyed its `SharedChannel` out from under the still-running test. Fixed by rooting it (classic "root fresh results between allocations"). Re-ran the shared-channel suite 5x under `-Dgc-stress=true` after the fix to make sure it wasn't flaky.

**Two things I left as-is, for discussion rather than silent judgment calls:**
1. **KEP §2's "release store" vs. the plain store used here** — agreed this is a real, if currently harmless, spec/code gap (every read in Phase 1 is same-thread or behind an already-synchronizing handoff). I'd rather leave the KEP-sync fix to whoever lands Phase 3's actual cross-thread wiring, since that's when the ordering requirement becomes load-bearing rather than theoretical — happy to do it now instead if you'd rather not carry the note forward.
2. **The foreign-owner error message's "pass it through the thread thunk to share it" advice** — left unchanged. It's the KEP's own normative wording, and while you're right that it doesn't yet *work* in Phase 1 (thread-thunk capture still raises `uncopyable type` until Phase 2 moves the copy to the parent thread), I read it as forward-looking guidance rather than a claim about today's behavior — closer to "this is the supported pattern" than "this works right now." Open to changing it if you'd rather Phase 1 not say anything Phase 2-shaped yet.

`zig build test` / `-Dgc-stress=true` / full Scheme+R7RS suite (1834/0) all green at 21f5aec.

baijum:
Re-reviewed at 21f5aec — **all findings verified fixed in the code, not just the summary. LGTM; this is ready to merge once CodeRabbit's incremental pass clears** (everything else on CI is already green).

What I checked, independently:

- **Bug 1 (`Envelope` leak):** `errdefer std.heap.c_allocator.destroy(env);` is in place, and the errdefer *order* is right — on a `deepCopy` failure the `env` struct is destroyed first, then the mini-GC teardown runs (which is the one that walks objects and releases stub refcounts), and neither touches the other.
- **Bug 2 (retain leak):** the reorder is sound. Between `allocChannelStub` returning and `sc.retain()` there is no allocation, and `deepCopyValue`'s `no_collect` guard covers the whole arm anyway, so the freshly tracked stub can't be swept in the gap; a `visited.put` failure after the retain still balances through `freeObject` at teardown. The new comment explains the invariant well.
- **CodeRabbit's park-path finding:** fixing it in `shared_channel.zig` (skip registration on null) rather than in the primitives was the right call — the old `notifier.?` was reachable as a panic through `receive` on any empty shared channel with a null notifier, which your own new primitive-dispatch test would have tripped had it received before sending.
- **Minor items:** `@panic` messages in both Phase-4 arms, the symbol-table deviation comment, `raiseFiberError` consolidation, `defer` restores, `th.TestContext` in the Motivation Path 2 test, and both new tests (reply-to rc 1→2→3→2 with the envelope-owned-stub identity assertion; white-box primitive dispatch through `vm.eval`) — all present and shaped as discussed. The gc-stress rooting fix in the reply-to test is textbook "root fresh results between allocations", and the comment explaining *why* `tasks_val` is rooted is exactly the kind that stays useful.
- **The two judgment calls:** both fine by me as-is. Deferring the KEP §2 release-store sync to Phase 3 (when the ordering becomes load-bearing) is where I'd have landed too — just carry it into #1468's checklist so it doesn't evaporate. The foreign-owner message reading as forward-looking KEP wording is a fair call.
- **Verification at 21f5aec on my side:** full `zig build test` green, `-Dgc-stress=true -Dtest-filter=channel` green, CI matrix fully green including the Debug job and ReleaseFast.

One non-blocking docs nit, fix here or in a follow-up: the new comment on `Envelope.create` (src/shared_channel.zig:53-55) says the private symbol table's cost is that "a symbol crossing a channel is `equal?` but not necessarily `eq?` to the receiver's own copy of the same name." That's not actually true: the receiver's copy-out goes through `gc.allocSymbol(name)` (`gc_deep_copy.zig`'s `.symbol` arm), which interns through the receiver's (or its shared parent's) table and returns the existing canonical symbol — so a received symbol **is** `eq?` to every other same-named symbol on the receiving thread. The envelope's private symbol objects are never visible to Scheme; the real (and only) cost is envelope-side: each envelope re-allocates its message's symbols instead of aliasing a shared table. Worth correcting so the `eq?` scare doesn't get copied into the KEP or into Phase 2/3 reasoning.

PR #1483, opened 2026-07-12
Envelopes at thread boundaries: thread-start!/join! via envelopes (KEP-0002 P2)

baijum:
## Summary

Implements KEP-0002 Phase 2 (#1467), building on Phase 1's SharedChannel core (#1482).

- `thread-start!` now copies the thunk into a `shared_channel.Envelope` on the **parent** thread, before `std.Thread.spawn`, instead of letting the child `deepCopy` `fiber.thunk` out of the still-running parent heap asynchronously. This closes the old concurrent-copy race and is the only place a channel captured by the thunk can legally promote (promotion requires `gc_instance` to match the channel's owner).
- `thread-join!`'s result/exception now cross the same way: built into an envelope on the **child** thread right before it exits, instead of being deep-copied directly out of the child's heap by the parent at join time. Retires `child_registry`'s raw-`Value` special case, and makes a channel *created and returned* by the child promote correctly too.
- The process-global live-thread counter this phase calls for already exists (`live_child_threads`, from #1455) — no new code needed.
- Fixed a double-free the envelope model exposes: `child_registry`'s result/exception lookup was a peek (`get`), so two racing `thread-join!` calls on the same fiber (reachable via a shared global, since fiber ownership is otherwise unchecked) could both retrieve and free the same `*Envelope`. Replaced with an atomic take (`takeResult`) that clears both fields under the registry lock.

Motivation Path 1 (a channel captured in the thread thunk) now works end to end; Motivation Path 2 (a channel reached through a shared global) is unaffected and still raises the foreign-owner error.

## Test plan

- [x] `zig build test` — full unit suite green
- [x] `zig build test -Dgc-stress=true` — full suite green (thread-churn/leak tests included)
- [x] `bash tests/scheme/run-all.sh` — 1835/0 (440 Scheme files + 1395 R7RS)
- [x] `zig fmt --check` clean
- [x] New tests: thunk-snapshot (parent mutation after `thread-start!` not visible to child), Motivation Path 1 end-to-end, a channel created+returned by the child, reply-channel identity across two promotion/alias hops, 20-thread churn with no refcount leak, synchronous uncopyable-thunk error shape
- [x] Manually measured `thread-start!`/`thread-join!` overhead for a trivial thunk before/after: ~34-38μs either way (within noise, dominated by OS thread creation, not the extra envelope copy)
- [x] Adversarial code review (8 independent finder angles + verify pass) — found and fixed a real double-free race plus a couple of cleanup items

### DIFF
--- a/src/primitives_srfi18.zig
+++ b/src/primitives_srfi18.zig
@@ -95,7 +95,13 @@ const ChildThreadResources = struct {
     child_gc: *memory.GC,
     child_vm: *vm_mod.VM,
     result: JoinResult = .none,
-    exception: ?*shared_channel.Envelope = null,
+    // Same shape as `result` (not a bare `?*Envelope`): an exception whose
+    // payload is itself uncopyable (R7RS `raise` permits raising an
+    // arbitrary value, including a port or a foreign-owned channel) or
+    // whose envelope build hits OOM needs to surface as *something*
+    // diagnostic at thread-join!, not silently collapse into a void
+    // reason -- see reapOsThread's `.failed` arm for the synthesized error.
+    exception: JoinResult = .none,
 };
 
 // Entries are freed only when a thread is joined (freeChildResources called
@@ -112,7 +118,7 @@ const ChildRegistry = struct {
         try self.map.put(key, res);
     }
 
-    fn storeResult(self: *ChildRegistry, key: usize, result: JoinResult, exception: ?*shared_channel.Envelope) void {
+    fn storeResult(self: *ChildRegistry, key: usize, result: JoinResult, exception: JoinResult) void {
         memory.spinLock(&self.mutex);
         defer memory.spinUnlock(&self.mutex);
         if (self.map.getPtr(key)) |entry| {
@@ -128,8 +134,8 @@ const ChildRegistry = struct {
     // fiber ownership is otherwise unchecked -- both retrieve the SAME
     // *Envelope pointer and both call env.deinit() on it: a double-free.
     // Clearing under the lock means at most one caller ever sees a non-.none
-    // result/non-null exception to consume.
-    const TakenResult = struct { result: JoinResult, exception: ?*shared_channel.Envelope };
+    // result/non-.none exception to consume.
+    const TakenResult = struct { result: JoinResult, exception: JoinResult };
 
     fn takeResult(self: *ChildRegistry, key: usize) TakenResult {
         memory.spinLock(&self.mutex);
@@ -137,10 +143,10 @@ const ChildRegistry = struct {
         if (self.map.getPtr(key)) |entry| {
             const taken: TakenResult = .{ .result = entry.result, .exception = entry.exception };
             entry.result = .none;
-            entry.exception = null;
+            entry.exception = .none;
             return taken;
         }
-        return .{ .result = .none, .exception = null };
+        return .{ .result = .none, .exception = .none };
     }
 
     fn fetchRemove(self: *ChildRegistry, key: usize) ?ChildThreadResources {
@@ -340,10 +346,10 @@ fn threadStartImpl(args: []const Value) PrimitiveError!Value {
         // and thread-join! surfaces this as "uncaught exception in thread",
         // matching a thunk that fails mid-flight instead of at the boundary.
         // No child_gc/child_vm exists yet for this failure (nothing was ever
-        // spawned), so the usual child_registry.storeResult path used by
-        // every post-spawn failure in threadEntryFn doesn't apply here --
-        // this is the one place a fiber's error state is set directly.
-        fiber.current_exception = try makeErrorWithType(.general, "thread thunk contains uncopyable type (port, continuation, etc.)", types.NIL);
+        // spawned), so child_registry.storeResult -- which every *post-spawn*
+        // failure in threadEntryFn goes through instead -- doesn't apply
+        // here: this is the one place a fiber's error state is set directly.
+        fiber.current_exception = try makeErrorWithType(.general, "thread thunk contains an uncopyable type (port, continuation, etc.), or a channel owned by another thread", types.NIL);
         gc.writeBarrier(&fiber.header, fiber.current_exception.?);
         @atomicStore(fiber_mod.FiberStatus, &fiber.status, .errored, .release);
         return args[0];
@@ -440,12 +446,18 @@ fn threadEntryFn(fiber: *fiber_mod.Fiber, allocator: std.mem.Allocator, parent_v
         // Built on this (owning) thread, mirroring thread-start!'s thunk
         // copy: this is the only place a channel raised in the exception
         // can legally promote (gc_instance == child_gc here, not the
-        // parent's gc it would be at join time).
-        const exc_envelope = if (child_vm.current_exception) |exc|
-            shared_channel.Envelope.create(exc) catch null
+        // parent's gc it would be at join time). A `.failed` build (an
+        // uncopyable type, or a channel owned by neither this thread nor
+        // the parent) still reaches reapOsThread as something diagnostic
+        // rather than silently becoming a void join reason.
+        const exc_result: JoinResult = if (child_vm.current_exception) |exc|
+            if (shared_channel.Envelope.create(exc)) |env|
+                .{ .envelope = env }
+            else |err|
+                .{ .failed = err }
         else
-            null;
-        child_registry.storeResult(@intFromPtr(fiber), .none, exc_envelope);
+            .none;
+        child_registry.storeResult(@intFromPtr(fiber), .none, exc_result);
         @atomicStore(fiber_mod.FiberStatus, &fiber.status, .errored, .release);
         return;
     };
@@ -463,7 +475,7 @@ fn threadEntryFn(fiber: *fiber_mod.Fiber, allocator: std.mem.Allocator, parent_v
         .{ .envelope = env }
     else |err|
         .{ .failed = err };
-    child_registry.storeResult(@intFromPtr(fiber), join_result, null);
+    child_registry.storeResult(@intFromPtr(fiber), join_result, .none);
     @atomicStore(fiber_mod.FiberStatus, &fiber.status, .completed, .release);
 }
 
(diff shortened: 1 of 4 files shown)


src/primitives_srfi18.zig
```diff
@@ -360,30 +426,44 @@ fn threadEntryFn(fiber: *fiber_mod.Fiber, allocator: std.mem.Allocator, parent_g
...
 
     const result = child_vm.callWithArgs(child_thunk, &.{}) catch {
         if (child_vm.current_fiber) |cf| fiber_mod.abandonFiberMutexes(child_gc, cf, child_vm.scheduler);
-        child_registry.storeResult(@intFromPtr(fiber), types.VOID, child_vm.current_exception);
+        // Built on this (owning) thread, mirroring thread-start!'s thunk
+        // copy: this is the only place a channel raised in the exception
+        // can legally promote (gc_instance == child_gc here, not the
+        // parent's gc it would be at join time).
+        const exc_envelope = if (child_vm.current_exception) |exc|
+            shared_channel.Envelope.create(exc) catch null
```
baijum: **A/B-verified behavior change (main → this PR), silent variant:** an exception whose payload references a *parent-owned* channel (reachable only via a shared global) is now silently dropped, where main preserved it.

```scheme
(define ch (make-channel))
(define t (make-thread (lambda () (raise ch))))
(thread-start! t)
(thread-join! t)
;; main:    uncaught exception in thread, reason = #<channel>
;;          (join-side copy ran on the parent = owner → legal promotion)
;; this PR: uncaught exception in thread, reason = void
;;          (child-side Envelope.create hits the foreign-owner check
;;           → UncopyableType → this `catch null`)
```

The `catch null` shape itself is inherited from the old join-side `gc.deepCopy(exc) catch null` (ports/continuations were already dropped silently), so this only *widens* the dropped set — but "child crashed and its exception vanished" is rough to debug, and this variant is the one real code is most likely to hit, via `(error "msg" some-global-channel)` in an unguarded thunk.

Non-blocking. Two possible shapes if you want to close it: (a) ~6 lines here — when `Envelope.create(exc)` fails, envelope a synthesized `makeErrorWithType(.general, "thread exception contained a value that cannot cross threads", ...)` built on the child, so join surfaces *something* diagnostic; or (b) fold it into a follow-up issue together with the result-path twin (see the comment on the `join_result` construction below).

src/primitives_srfi18.zig
```diff
@@ -360,30 +426,44 @@ fn threadEntryFn(fiber: *fiber_mod.Fiber, allocator: std.mem.Allocator, parent_g
...
+    // GC never traverses a child-heap pointer (Race C) -- and, since Phase
+    // 2, as an envelope rather than a raw Value, so a channel the thunk
+    // created and returned promotes correctly (see the exception path
+    // above for why this must run on this thread).
+    const join_result: JoinResult = if (result == types.VOID)
+        .none
+    else if (shared_channel.Envelope.create(result)) |env|
+        .{ .envelope = env }
+    else |err|
+        .{ .failed = err };
```
baijum, line 477: Two notes on the `.failed` path:

1. **A/B-verified behavior change (loud twin of the exception-path note above):** a child that *returns* a parent-owned channel reached via a shared global now errors, where main returned the (legally, because parent-side) promoted channel:

```scheme
(define gch (make-channel))
(thread-join! (thread-start! (make-thread (lambda () gch))))
;; main:    returns #<channel>
;; this PR: raises "thread-join!: result contains uncopyable type (port, continuation, etc.)"
```

   Refusing is arguably the more principled reading of invariant 4, and unlike the exception case it's loud — but (a) the message is now misleading for this input (a channel isn't an uncopyable *type*; only the owner is wrong), and (b) it's a capability removal vs main that the PR description's "Motivation Path 2 is unaffected" doesn't quite cover. If the old behavior ever matters: the child heap is still alive at join (freed only in `freeChildResources`), so a `.failed`-with-raw-`Value` variant falling back to the old parent-side deepCopy would preserve both behaviors. Follow-up material either way.

2. **Test gap:** nothing in the suite exercises `.failed` → `"thread-join!: result contains uncopyable type"` (e.g. a thunk returning a string port). The PR rewired exactly this path (join-side deepCopy error → child-side envelope error) and I could only verify it manually — it does work. A one-liner in `srfi18-cross-thread-channels.scm` or `tests_shared_channel.zig` would pin it.

src/primitives_srfi18.zig
```diff
@@ -103,10 +121,26 @@ const ChildRegistry = struct {
...
+    // in place for freeChildResources' own fetchRemove). A plain get() would
+    // let two racing reapOsThread calls for the same fiber_key -- reachable
+    // today only through a fiber value reached via a shared global, since
+    // fiber ownership is otherwise unchecked -- both retrieve the SAME
+    // *Envelope pointer and both call env.deinit() on it: a double-free.
+    // Clearing under the lock means at most one caller ever sees a non-.none
+    // result/non-null exception to consume.
+    const TakenResult = struct { result: JoinResult, exception: ?*shared_channel.Envelope };
+
+    fn takeResult(self: *ChildRegistry, key: usize) TakenResult {
```
baijum, line 140: Good catch on the double-free — the atomic take is the right fix, and it quietly removes a *pre-existing* UAF too (with the old peek, two concurrent joiners both deepCopied out of the child heap while the first's `freeChildResources` could destroy that heap mid-copy; post-PR the winner copies out of the c_allocator envelope, which the loser can't free).

Two residual hazards in the same concurrent-double-join family, both pre-existing and fine to leave out of this PR, but worth a follow-up issue now that this comment documents the scenario as reachable:

1. Both joiners can still pass the `target.os_thread != null` check in `threadJoinFn` and race into `thread.join()` on the same handle (UB), plus the unsynchronized `target.os_thread = null` store.
2. The loser of the `takeResult` race falls through to `threadJoinResult` and can read `target.result` *before* the winner stores it — observing VOID instead of the result. (Sequential double-join is fine — verified it returns the result both times.)

Since fibers — unlike channels — have no foreign-owner check, the cleanest umbrella fix may be exactly that: reject `thread-join!`/`thread-terminate!` on a fiber owned by another GC, making the whole class unreachable instead of mitigating it point by point.

src/primitives_srfi18.zig
```diff
@@ -292,12 +326,39 @@ fn threadStartImpl(args: []const Value) PrimitiveError!Value {
...
+    // on the parent thread.
+    const envelope = shared_channel.Envelope.create(fiber.thunk) catch |err| {
+        if (err != error.UncopyableType) return PrimitiveError.OutOfMemory;
+        // Mirror the old error shape exactly: thread-start! itself never
+        // raises -- the fiber is left .errored (no OS thread ever spawned)
+        // and thread-join! surfaces this as "uncaught exception in thread",
+        // matching a thunk that fails mid-flight instead of at the boundary.
+        // No child_gc/child_vm exists yet for this failure (nothing was ever
+        // spawned), so the usual child_registry.storeResult path used by
+        // every post-spawn failure in threadEntryFn doesn't apply here --
```
baijum: Nit: "the usual child_registry.storeResult path used by every post-spawn failure in threadEntryFn" overstates — only the `callWithArgs`-failure path stores a result. The earlier post-spawn failures (child GC/VM alloc, registry put, envelope copy-out) set `.errored` without ever calling `storeResult`, i.e. they behave like this path, not unlike it.

baijum (commented):
Critical review of KEP-0002 Phase 2. Solid PR overall — the envelope-at-both-boundaries design is right, the double-free fix is real (and fixes a pre-existing UAF besides), and the comments/tests are unusually good at pinning *why* each path exists.

### What I verified

- **Envelope lifecycle audit** — exactly-one-`deinit` holds on every path I could trace: the thunk envelope is owned by `threadEntryFn`'s defer with both pre-spawn failure paths covered; result/exception envelopes are consumed (or intentionally leak with the documented never-joined child VM/GC leak).
- **GC safety** — the "thunk needs no separate root" claim checks out (`markFiberState` traces `fiber.thunk`, fiber.zig:699); the new sync-error path's `current_exception` is traced (fiber.zig:722) and write-barriered; `Envelope.create` never allocates on the caller's heap, so no rooting hazards in `threadStartImpl`.
- **Invariants** — "copy-out can only fail on OOM" holds (envelope contents are deepCopy outputs; channels inside are already-promoted stubs, which alias without an ownership check); result-envelope ↔ `.completed` and exception ↔ `.errored` pairings are consistent at every `storeResult` site, with `thread.join()` providing the happens-before for the plain status read in `reapOsThread`.
- **Ran it** — built the PR head; `tests_shared_channel` unit tests and the new `srfi18-cross-thread-channels.scm` (7/7) pass; A/B-probed PR vs main (58036d9d) binaries for the thunk snapshot, Path 1, double-join, the `.failed` path, and the two behavior deltas flagged inline.
- **CI** — green across all platforms at review time (Debug job still running), format clean, benchmarks flat.

### Findings (all non-blocking)

- Two A/B-verified behavior changes vs main, both from owner-dependent copyability moving from the parent/join side to the child/exit side: a **silent** exception drop (inline at the `Envelope.create(exc) catch null` line) and a **loud** result-path refusal (inline at the `join_result` construction). Narrow shared-global edge cases, in the direction the KEP pushes anyway — but the silent one deserves either the small fallback or a follow-up issue so it doesn't get lost.
- One test gap: the `.failed` → `thread-join!` error path has no regression test.
- One residual pre-existing double-join note on `takeResult` (details inline) and one comment nit.

### Verdict

**Ready to merge once the Debug CI job finishes green.** None of the findings block: the deltas are edge cases a follow-up issue can carry, and everything load-bearing — lifecycle, rooting, promotion legality, error shapes — checks out under both code reading and A/B probing.

baijum:
Thanks for both passes — pushed fixes at 1276326c.

**Fixed:**
- Exception-envelope failures no longer silently collapse to a void join reason: `exception` is now the same `JoinResult` shape as `result`, and a `.failed` build synthesizes a diagnostic error.
- Both the thunk and result-path "uncopyable type" messages now also name "a channel owned by another thread" as a cause — verified against the A/B repros in the inline comments (a shared-global channel raised or returned by the thunk now surfaces a real error instead of a void reason or a misleading message).
- The `child_registry.storeResult` comment overstatement (nit).
- All CodeRabbit test-quality items: `th.expectEval` for the two one-shot fixnum tests, `types.isChannel` instead of a raw `.tag` comparison, the exception-envelope test now verifies the drained `'irritant-marker` value (not just that promotion happened), the 20-iteration thread-churn test scales to 5 under `-Dgc-stress=true`, and a new regression test pins the `.failed` → `"thread-join!: result contains..."` path specifically.

**Deferred to #1484** (both reviews framed these as non-blocking / follow-up material): the residual concurrent-`thread-join!` hazards (`thread.join()` called twice on the same handle, a result-read race — both pre-existing, both need a foreign-owner check on fiber primitives to close properly, not a point patch), and the deeper design question of whether a shared-global channel merely *forwarded back to its own owner* through an exception/result should regain exact main-branch parity, versus the current (more principled per invariant 4, but capability-narrowing) refusal.

Full unit + `-Dgc-stress=true` + Scheme/R7RS suites all green at 1276326c; `zig fmt` clean.

src/tests_shared_channel.zig
```diff
@@ -533,3 +533,174 @@ test "Motivation Path 2 regression: a channel reached through a shared global ra
...
+// thread-start! now copies the thunk into an envelope on the *parent*
+// thread, before Thread.spawn -- the child copies out of the envelope into
+// its own fresh heap instead of deepCopy-ing fiber.thunk directly out of the
+// still-running parent heap. thread-join!'s result/exception cross the same
+// way, built on the *child* thread right before it exits. Both directions
+// are exercised here with real OS threads (make-thread/thread-start!), since
+// the whole point of Phase 2 is the parent/child-thread timing -- a
+// white-box, single-thread test of Envelope.create (already covered by
+// Phase 1's tests above) can't observe the race Phase 2 closes.
+// ---------------------------------------------------------------------------
```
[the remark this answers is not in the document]
baijum: Verified this one deliberately stays open with #1484 rather than getting a test in this PR — and the literal suggestion (two OS threads racing `thread-join!` on one fiber) would be counterproductive today: concurrent double-join still hits *pre-existing* UB documented in #1484 (both joiners can race into `thread.join()` on the same handle, plus the unsynchronized `target.os_thread = null` store), so the test would be exercising undefined behavior and flaky by construction — regardless of the `takeResult` fix it's meant to pin.

What *is* deterministically covered now: sequential double-join including the timeout-then-reap path (`tests/scheme/smoke/thread-join-timeout-878.scm`, which leaves the envelope in the registry across the first timed-out join), and probes verified both joins return the real result. Once #1484's foreign-owner check lands, the concurrent scenario becomes a deterministic error path (two foreign joiners both get descriptive errors) and the regression test belongs there.

Resolving as tracked-in-#1484.

PR #1485, opened 2026-07-13
Cross-thread wakeup: ThreadNotifier, unconditional sweep, shared-waiter registry (KEP-0002 P3)

baijum:
## Summary

Implements KEP-0002 Phase 3 (#1468): the reactor-backed cross-thread wakeup that Phases 1/2 (#1482/#1483) left as an unwired `@panic`.

**This closes a real, already-reachable crash**, not just a missing feature: a channel captured as a true lexical closure upvalue in a `thread-start!` thunk promotes in place on *both* sides (parent and child, since promotion mutates the shared heap object). Either side calling `channel-receive` on the still-empty promoted channel before the other sends hits Phase 1's placeholder `@panic` and SIGABRTs the process. I verified this against a pre-fix build before writing any code.

- **`src/reactor.zig`**: `ThreadNotifier` (kqueue `EVFILT.USER` / epoll `eventfd` / WASI no-op), one per `Reactor`. Base refcount owned by the creating `Reactor`; registrations are additive on top. Kernel-resource closing is concentrated entirely in `releaseNotifier`'s zero-transition (not split with backend `deinit`), since kqueue's notifier knote shares `kq` with ordinary fd polling — splitting it would risk a double-close race on whichever release happens last. `wait()`'s event loop filters the notifier's own event out of the ordinary `ReadyEvent` stream on both backends. New `notifierLiveCount()` leak-check counterpart to `shared_object.liveCount()` (`ThreadNotifier` is deliberately not a `shared_object.Header` instance).
- **`src/fiber.zig`**: the per-scheduler shared-waiter registry and its **unconditional** sweep — a readiness-filtered sweep is the exact model-checked lost-wakeup from kaappi/keps#12 ("finding 1"), not a style choice. The `wake_pending` consume-protocol swap-loop is wired into both `schedule()` (every tick) and `parkOnReactor()` (before blocking and again after `poll()` returns, since a notify arriving mid-block is what interrupted it). `hasRunnableFibers()` extended so the deadlock check doesn't fire while a fiber still has a live cross-thread wakeup path.
- **`src/shared_channel.zig`**: `ring()` now actually calls `.notify()`; §7 opportunistic dead-notifier pruning folded into the existing `register*Waiter` dedup walk; `promoteChannel` gains §2 step 4's local-waiter migration.
- **`src/primitives_fiber.zig`**: `channelReceiveFn`'s shared-path `@panic` replaced with `channelReceiveShared` — always calls `shared_channel.receive()` first each loop iteration (registration is a side effect of `.would_park`) and only then parks, never the other way around; parking before registering would leave nothing to ever ring the fiber awake. This ordering also directly implements the required "a rung receiver that loses the pop race re-parks, re-registers" regression.
- **`src/pct_stress.zig` + `src/stress_channel.zig`** (new) + `zig build stress-channel`: PCT-style randomized yield injection, off by default. Confirmed effective by deliberately reintroducing the "finding 1" bug and verifying both the harness and the deterministic unit test catch it, then reverting.

**Deferred to a follow-up** (documented, not silently dropped): UQ3 (`thread-join!`'s 1ms poll → notifier) needs a structurally distinct registration relationship from `SharedChannel`'s, and isn't "cheap" per the KEP's own bar for in-scope-now work.

## Test plan

- [x] Required regressions, as automated tests: "park locally → promote → remote send wakes" (§2 step 4); "a rung receiver that loses the pop race re-parks and re-registers" (§5, model finding 1)
- [x] `zig build test` and `zig build test -Dgc-stress=true` — full suite green (832 tests; 824 pass + 6 scaled-down skips under gc-stress)
- [x] `bash tests/scheme/run-all.sh` — 1835/0
- [x] `zig build stress-channel` across multiple seeds and heavier producer/consumer/message-count configurations — all pass
- [x] `zig build wasm` still compiles (WASI notifier backend is a no-op)
- [x] `zig fmt --check` clean
- [x] Re-ran kaappi#1455's mutex/condvar suite as part of the full test suite — unaffected

### DIFF
--- a/src/primitives_fiber.zig
+++ b/src/primitives_fiber.zig
@@ -166,6 +166,19 @@ const SharedChannelWait = struct {
     }
 };
 
+/// Drives OTHER local fibers without ever touching `me`'s own status --
+/// `me` stays .running for the whole call, so schedule()'s round-robin can
+/// never independently re-pick it while its own native call (this one) is
+/// still live on the Zig stack. Isolating this into its own Ctx type keeps
+/// that invariant visually obvious at every call site: nothing here sets
+/// `me.status`, unlike SharedChannelWait/ChannelWait's actual park.
+const SharedChannelPoll = struct {
+    sc: *shared_channel.SharedChannel,
+    pub fn isDone(self: SharedChannelPoll) bool {
+        return self.sc.peekReady();
+    }
+};
+
 /// KEP-0002 §5's deadlock-heuristic disjunct: block (park, waiting for a
 /// remote send/receive) rather than raise a local deadlock whenever another
 /// thread could plausibly still act on this channel -- either it still
@@ -177,48 +190,79 @@ fn sharedWakeupPossible(sc: *shared_channel.SharedChannel) bool {
     return sc.refCount() > 1 or srfi18.crossThreadWaitPossible();
 }
 
-/// KEP-0002 §4/§5 send/receive on the shared representation, replacing
-/// Phase 1's `@panic` on `.would_park` -- a real, already-reachable SIGABRT
-/// before this fix (a channel captured by a thread-start! thunk promotes in
-/// place on both sides; either side calling channel-receive on the empty
-/// promoted channel before the other sends crashed the process).
+/// KEP-0002 §4/§5 receive on the shared representation, replacing Phase 1's
+/// `@panic` on `.would_park` -- a real, already-reachable SIGABRT before
+/// this fix (a channel captured by a thread-start! thunk promotes in place
+/// on both sides; either side calling channel-receive on the empty promoted
+/// channel before the other sends crashed the process).
 ///
-/// Always calls shared_channel.receive() FIRST each loop iteration -- that's
-/// what performs the actual sc.recv_waiters registration, as a side effect
-/// of returning `.would_park` -- and only *then* decides to park. Parking
-/// before registering would leave nothing to ever ring the fiber awake: a
-/// permanent hang inside parkOnReactor's blocking poll, strictly worse than
-/// today's clean deadlock error. This ordering also directly implements the
-/// required "a rung receiver that loses the pop race re-parks, re-registers"
-/// regression (§5, model finding 1): losing the race just means another
-/// `.would_park` on the next loop iteration, which re-registers via the same
-/// receive() call.
-///
-/// Deliberately does not reuse blockOrDeadlock's yield_retry trick: that
-/// helper is sound for a *local* channel because a later local channel-send
-/// can still find and flip the fiber via wakeChannelWaiters even after
-/// runSchedulerStep reports "not done". For a *promoted* channel nothing
-/// ever calls wakeChannelWaiters again -- a yield_retry-parked fiber not
-/// enrolled in the shared-waiter registry would never wake. So this always
-/// raises immediately when sharedWakeupPossible() is false, uniformly for
-/// the main fiber and any spawned fiber (runSchedulerStep is already called
-/// unconditionally regardless of dispatched_from_scheduler), which is what
-/// satisfies §5's "the main-fiber case follows the same rule" without any
-/// special-casing: the main fiber genuinely blocks in parkOnReactor's
-/// reactor.poll() exactly like any other fiber.
+/// Per loop iteration:
+///  1. Call shared_channel.receive() -- the notifier is registered
+///     unconditionally (not gated on sharedWakeupPossible), because a purely
+///     LOCAL sibling fiber's channel-send on this channel also rings via the
+///     same recv_waiters/notifier path (self-ring): a channel can be
+///     promoted (e.g. transiently, by a thread that captured it and exited)
+///     while every actual sender/receiver stays on this thread the whole
+///     time, and that must keep working exactly like it did before
+///     promotion.
+///  2. If empty, drive OTHER local fibers once (SharedChannelPoll) -- `me`
+///     stays .running throughout, mirroring channelReceiveFn's local-channel
+///     path (ChannelWait). Loop back to 1 if driving made the channel ready.
+///  3. Otherwise park. The two branches are NOT symmetric, and the asymmetry
+///     is load-bearing, not an oversight:
+///       - A fiber dispatched directly by a scheduler loop (my_idx != 0 and
+///         vm.dispatched_from_scheduler) ALWAYS parks via the flat
+///         yield_retry unwind (KEP-0002 §4 receive step 8: ".waiting on the
+///         stub, yield_retry rewind"), unconditionally -- exactly like
+///         blockOrDeadlock's existing behavior for every non-main fiber.
+///         Whether this is a genuine, permanent deadlock is not this
+///         fiber's call to make (the existing local-channel path doesn't
+///         make that call either): that's detected wherever something else
+///         -- typically the main fiber, via fiber-join or another
+///         blockOrDeadlock call -- eventually finds nothing progressing.
+///         Gating this branch on sharedWakeupPossible() was an earlier,
+///         confirmed-buggy version of this function: it raised a false
+///         deadlock the instant a channel was ever transiently promoted
+///         and then the "remote" side exited, even though a purely local
+///         sibling fiber was one form away from sending (verified via a
+///         concrete repro). This branch must also never nest another
+///         drive after setting `.waiting` -- see SharedChannelPoll's and
+///         runSchedulerStep's own doc comments: doing so would make `me`
+///         visible to schedule()'s round-robin while `me`'s own call is
+///         still live on the Zig stack, letting an unrelated fiber's own
+///         nested runSchedulerStep dispatch `me`'s mid-call snapshot and
+///         resume bytecode past the in-flight receive() call with the
+///         destination register never written (confirmed via a second,
+///         separate repro: two spawned fibers each parked on their own
+///         promoted channel, the second one woken "received" the stale
+///         callee register instead of its real value). yield_retry instead
+///         unwinds this whole native call via error.Yielded before `me`
+///         ever becomes independently dispatchable, and a later redispatch
+///         re-executes this entire function from the top (re-registering
+///         via step 1's receive() call).
+///       - The main fiber (or a fiber blocked under re-entrant native
+///         frames that cannot be safely rewound) CANNOT yield_retry, so it
+///         drives in place instead (waitForFd's precedent) via a genuine
+///         park + SharedChannelWait, then loops back to 1. Here
+///         sharedWakeupPossible() is load-bearing: self-enrolling in
+///         shared_waiters makes hasRunnableFibers() (and therefore
+///         parkOnReactor's own "nothing can ever happen" detection) always
+///         see this fiber's own entry, so parkOnReactor can never
+///         independently conclude "genuine deadlock" the way it does for
+///         purely local waits -- something has to decide "no local or
+///         remote path could ever help" before parking, or a truly
+///         deadlocked main fiber would block in reactor.poll() forever
+///         instead of raising a clean error.
 fn channelReceiveShared(sc: *shared_channel.SharedChannel, gc: *memory.GC, ch_val: Value) PrimitiveError!Value {
     const vm = vm_mod.vm_instance orelse return PrimitiveError.OutOfMemory;
     const ctx = try fiber_mod.ensureScheduler(vm);
-    const me = ctx.sched.fibers.items[ctx.sched.current_idx].?;
+    const my_idx = ctx.sched.current_idx;
+    const me = ctx.sched.fibers.items[my_idx].?;
     const notifier = ctx.reactor.notifyHandle();
+    const is_dispatched = my_idx != 0 and vm.dispatched_from_scheduler;
 
     while (true) {
-        const wakeup_possible = sharedWakeupPossible(sc);
-        // A null notifier when no wakeup is possible: registering would be
-        // pointless (nothing will ever ring it) and would leave a dangling
-        // entry for nothing -- matches send()/receive()'s existing
-        // null-is-safe convention from Phase 1.
-        const outcome = shared_channel.receive(sc, gc, if (wakeup_possible) notifier else null) catch |err|
+        const outcome = shared_channel.receive(sc, gc, notifier) catch |err|
             return translateSharedChannelError(err, "channel-receive");
         switch (outcome) {
             .value => |v| return v,
@@ -227,19 +271,46 @@ fn channelReceiveShared(sc: *shared_channel.SharedChannel, gc: *memory.GC, ch_va
             .eof => @panic("channel-receive: .eof unreachable before channel-close! (Phase 4)"),
             .would_park => {},
         }
-        if (!wakeup_possible)
+
+        // Drive local siblings once. `me` stays .running -- see
+        // SharedChannelPoll's doc comment for why that's load-bearing.
+        _ = try fiber_mod.runSchedulerStep(SharedChannelPoll, .{ .sc = sc }, ctx.vm, ctx.sched, me);
+        if (sc.peekReady()) continue; // loop back to 1: re-derive via receive()
+
+        if (is_dispatched) {
+            me.status = .waiting;
+            me.waiting_on = ch_val;
+            vm.gc.writeBarrier(&me.header, ch_val);
+            ctx.sched.enrollSharedWaiter(me) catch |err| {
+                me.status = .running;
+                me.waiting_on = types.VOID;
+                return err;
+            };
+            vm.yield_retry = true;
+            return PrimitiveError.Yielded;
+        }
+
+        if (!sharedWakeupPossible(sc))
             return raiseFiberError("channel-receive: deadlock — channel is empty and no other thread can send");
 
         me.status = .waiting;
         me.waiting_on = ch_val;
         vm.gc.writeBarrier(&me.header, ch_val);
         me.timed_out = false; // defensive: stale flag from an unrelated earlier timed wait
-        try ctx.sched.enrollSharedWaiter(me);
-        _ = try fiber_mod.runSchedulerStep(SharedChannelWait, .{ .me = me }, ctx.vm, ctx.sched, me);
+        ctx.sched.enrollSharedWaiter(me) catch |err| {
+            me.status = .running;
+            me.waiting_on = types.VOID;
+            return err;
+        };
+        _ = fiber_mod.runSchedulerStep(SharedChannelWait, .{ .me = me }, ctx.vm, ctx.sched, me) catch |err| {
+            ctx.sched.removeSharedWaiter(me);
+            me.status = .running;
+            me.waiting_on = types.VOID;
+            return err;
+        };
         ctx.sched.removeSharedWaiter(me);
-        // Loop back: re-derive wakeup_possible and re-call receive(), which
-        // both re-registers if still empty and picks up a value if one
-        // arrived while parked.
+        // Loop back to 1: re-registers if still empty, picks up a value if
+        // one arrived while parked.
     }
 }
 
--- a/src/primitives_fiber.zig
+++ b/src/primitives_fiber.zig
@@ -5,6 +5,7 @@ const primitives = @import("primitives.zig");
 const memory = @import("memory.zig");
 const fiber_mod = @import("fiber.zig");
 const shared_channel = @import("shared_channel.zig");
+const srfi18 = @import("primitives_srfi18.zig");
 const Value = types.Value;
 const NativeFn = types.NativeFn;
 const PrimitiveError = primitives.PrimitiveError;
@@ -120,14 +121,15 @@ fn channelSendFn(args: []const Value) PrimitiveError!Value {
         };
         return switch (outcome) {
             .sent => types.VOID,
-            // Phase 1 channels are always unbounded and open -- no
+            // Channels are still always unbounded and open -- no
             // Scheme-level API sets capacity or closed yet (KEP-0002
-            // Phase 4) -- so these branches are unreachable through
-            // make-channel's output. @panic (not `unreachable`, which is UB
-            // under ReleaseFast) so a Phase 4 gap that forgets to wire this
-            // switch fails loudly in every build mode instead of silently
-            // corrupting execution.
-            .would_park, .closed => @panic("channel-send: reached a shared-channel branch Phase 1's primitives don't wire up yet (capacity/close is Phase 4)"),
+            // Phase 4) -- so these branches remain unreachable through
+            // make-channel's output even after Phase 3 (#1468) wires up
+            // channel-receive's own .would_park branch below. @panic (not
+            // `unreachable`, which is UB under ReleaseFast) so a Phase 4
+            // gap that forgets to wire this switch fails loudly in every
+            // build mode instead of silently corrupting execution.
+            .would_park, .closed => @panic("channel-send: reached a shared-channel branch Phase 4 doesn't wire up yet (capacity/close)"),
         };
     }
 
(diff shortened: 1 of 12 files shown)


baijum (commented):
Reviewed the full diff plus KEP-0002 (§§2, 4, 5, 7 and the keps#12 amendments), traced the park/wake protocol through `fiber.zig`/`vm_dispatch.zig`, and ran targeted repros against a build of this branch.

## Verdict

**Not ready to merge.** Two confirmed correctness bugs in `channelReceiveShared`, both reproduced on this branch and both deviations from KEP-0002's own prescriptions (details + repro output in the inline comments):

1. **Silent VM corruption with two receivers** (inline comment on the `channelReceiveShared` doc block): drive-parking scheduler-dispatched fibers instead of KEP §4 step 8's yield_retry park makes swept fibers' mid-call snapshots dispatchable from a sibling's nested loop. Reproduced: a fiber "received" `#<builtin channel-receive>` (the stale callee register), the real value was lost, exit 0. This is the default path for the second shared receiver on a thread, not a rare race — the PR's tests stay green because every one of them parks only the main fiber on the shared path.
2. **False-positive deadlock on transiently promoted channels** (inline comment on the `!wakeup_possible` raise): the immediate raise skips KEP §5's "today's deadlock semantics apply" fallback, so a channel that once passed through a thread permanently loses ordinary local fiber-to-fiber use. Reproduced; the unpromoted control prints 42.

Both fixes converge on the same localized change: route dispatched fibers through enroll + yield_retry (the KEP's park), keep the drive loop for the main/native-frame fallback, and make the deadlock raise blockOrDeadlock's decision rather than a pre-park check. Plus two smaller items: a swallowed OOM in `promoteChannel`'s migration (hang, not error) and an fd leak on `Reactor.init`'s notifier-OOM path.

## What's solid (and it's most of the PR)

- **ThreadNotifier lifecycle**: the single-owner close concentrated in `releaseNotifier`'s zero-transition is the right call, and the kqueue shared-`kq` double-close reasoning is correct; `alive` ordering in `Reactor.deinit` (flip before release) is right, and `ring()`'s notify-before-release holds the needed ref.
- **Consume protocol**: placed at all three sites (`schedule()` per tick, `parkOnReactor` pre- and post-poll) exactly per §5's normative sequence; the sticky `wake_pending` + fd-ring pairing closes the register-then-park race.
- **Backends**: EV.CLEAR semantics, eventfd drain-on-wake, and filtering the notifier out of the `ReadyEvent` stream on both backends (the ident-0/stdin collision note is a genuinely good catch).
- **§2 step 4 migration** is present and correctly unconditional on refcount (the promotion-time refcount==1 observation is right), and migrated fibers have clean yield_retry snapshots, so waking them is sound.
- Waiter-list pruning under `swapRemove` iterates correctly; `destroyHook` releases registration refs; the leak-counter baselines in every test are good discipline; the PCT stress harness is genuinely useful and correctly zero-cost when disabled.

## Notes, non-blocking

- **Accepted liveness gap worth documenting**: a receiver parked while `sharedWakeupPossible()` was true hangs forever if the peer thread exits without sending (stub release rings nothing; `hasRunnableFibers` stays true via the registry, so the deadlock detector is intentionally disarmed). KEP §5 explicitly accepts Go-style hangs and §6 timeouts are the escape hatch, but a code comment naming this — and possibly a future release-side ring at thread teardown — would help the next reader.
- **Pre-existing hazard class**: mutex-lock!/condvar/thread-sleep! also set `.waiting` around inline drives, so dirty-snapshot re-dispatch is theoretically reachable there today (narrow single-waiter wake windows, which is why it hasn't bitten). Worth a follow-up issue for the class (e.g. a `.driving` status `schedule()` never selects) rather than scope creep here.
- **Test gap**: after the fix, please add (a) a two-spawned-receivers regression and (b) a transient-promotion-then-local-use regression — both programs in the inline comments are small and deterministic enough to drop into `tests_shared_channel.zig` / a smoke test.
- CI: green across platforms at review time except `test (ubuntu-latest, Debug, 30)` still pending.

The architecture is right and matches the KEP; the two bugs are localized to `channelReceiveShared`'s park-and-raise decisions. With those reworked per §4/§5 and regressions added, this should be ready.

src/primitives_fiber.zig
```diff
@@ -157,6 +159,90 @@ const ChannelWait = struct {
...
+/// of returning `.would_park` -- and only *then* decides to park. Parking
+/// before registering would leave nothing to ever ring the fiber awake: a
+/// permanent hang inside parkOnReactor's blocking poll, strictly worse than
+/// today's clean deadlock error. This ordering also directly implements the
+/// required "a rung receiver that loses the pop race re-parks, re-registers"
+/// regression (§5, model finding 1): losing the race just means another
+/// `.would_park` on the next loop iteration, which re-registers via the same
+/// receive() call.
+///
+/// Deliberately does not reuse blockOrDeadlock's yield_retry trick: that
```
baijum: **Confirmed VM corruption: drive-parking a scheduler-dispatched fiber makes its mid-call snapshot dispatchable. This also deviates from KEP-0002's prescribed park.**

KEP-0002 §4 receive step 8 specifies the park as ".waiting on the stub, **yield_retry rewind**", with the drive loop reserved for the main-thread/native-frame fallback. The rationale in this comment rejects yield_retry because a yield_retry-parked fiber "not enrolled in the shared-waiter registry would never wake" — but enrollment is orthogonal to the park mechanism: enroll **before** setting `yield_retry` and the sweep flips the fiber exactly as it does here; re-dispatch then re-executes the primitive (registration already happens inside `receive()`), which is the KEP's design.

Drive-parking a dispatched fiber instead is unsound. Its `runSchedulerStep` snapshot is a **mid-native-call** snapshot — `frame.ip` is already past the call opcode (`maybeRewindRetry` only rewinds under `yield_retry`). The unconditional sweep flips **every** enrolled fiber from whichever loop is innermost, so a second drive-parked receiver's retry loop finds the first one `.suspended` and dispatches it: `runUntil` resumes right after the `channel-receive` call with the stale callee still sitting in the dst register, while the fiber's real native frames are still pinned deeper on the OS stack. Both executions then run.

Reproduced on this branch (silent corruption, exit 0):

```scheme
(import (kaappi fibers))
(define ch1 (make-channel))
(define ch2 (make-channel))
(define log '())
(define f1 (spawn (let ((c ch1)) (lambda ()
  (let ((v (channel-receive c))) (set! log (cons (list 'f1 v) log)) v)))))
(define f2 (spawn (let ((c ch2)) (lambda ()
  (let ((v (channel-receive c))) (set! log (cons (list 'f2 v) log)) v)))))
(define t (make-thread (let ((a ch1) (b ch2)) (lambda ()
  (thread-sleep! 0.3) (channel-send a 111)
  (thread-sleep! 0.3) (channel-send b 222)))))
(thread-start! t)
(display "join f1 => ") (display (fiber-join f1)) (newline)
(display "join f2 => ") (display (fiber-join f2)) (newline)
(thread-join! t)
(display log) (newline)
```

Output:

```
join f1 =>                                    ; f1's 111 is lost
join f2 => 222
((f2 222) (f1 #<builtin channel-receive>))    ; f1 "received" the stale callee register
```

`#<builtin channel-receive>` as the received value is the smoking gun: the resumed snapshot read the call's base register (the callee) as the result. This is the *default* path whenever two receivers on one thread are parked on promoted channels — not an exotic interleaving.

Suggested fix, which is just the KEP's own protocol: for `my_idx != 0 and vm.dispatched_from_scheduler` → set `.waiting`/`waiting_on`, `enrollSharedWaiter`, `vm.yield_retry = true`, return `PrimitiveError.Yielded` (blockOrDeadlock's shape plus enrollment); keep the drive loop only for the main fiber / re-entrant-native case. That also removes the unbounded native-stack nesting this shape reintroduces (#1463's concern: N parked receivers = N pinned `runSchedulerStep` frames).

Two notes for completeness: (1) the dispatchable-dirty-snapshot hazard itself predates this PR — mutex-lock!/condvar/thread-sleep! also set `.waiting` around an inline drive — but their wake windows are narrow single-waiter flips, while the sweep is wake-all on every remote message; worth a follow-up issue for the class (e.g. a `.driving` status that `schedule()` never picks). (2) Please add a regression with **two spawned receivers** on two promoted channels — every test in this PR parks only the main fiber on the shared path, which is exactly why the suite is green.

src/primitives_fiber.zig
```diff
@@ -157,6 +159,90 @@ const ChannelWait = struct {
...
+            return translateSharedChannelError(err, "channel-receive");
+        switch (outcome) {
+            .value => |v| return v,
+            // channel-close! is Phase 4 -- sc.closed is always false, so
+            // receive's reserved==0-and-closed eof branch is unreachable.
+            .eof => @panic("channel-receive: .eof unreachable before channel-close! (Phase 4)"),
+            .would_park => {},
+        }
+        if (!wakeup_possible)
+            return raiseFiberError("channel-receive: deadlock — channel is empty and no other thread can send");
```
baijum, line 294: **Confirmed false-positive deadlock: this raises immediately instead of falling back to "today's deadlock semantics" (KEP §5), breaking purely local use of a transiently promoted channel.**

KEP §5: "only when both are false do **today's deadlock semantics** apply" — and today's semantics (the local path) run the scheduler first and raise only when nothing else can run (`runSchedulerStep` + `blockOrDeadlock`). Raising here, before any sibling gets to run, turns one transient promotion into a permanent break of ordinary fiber-to-fiber use of that channel:

```scheme
(import (kaappi fibers))
(define ch (make-channel))
;; transient promotion: thread captures a stub, does nothing, exits
(thread-join! (thread-start! (make-thread (let ((c ch)) (lambda () (channel? c))))))
(define f (spawn (let ((c ch)) (lambda () (channel-receive c)))))
(yield)
(channel-send ch 42)          ; local sibling send — would wake f via the self-ring
(display (fiber-join f)) (newline)
```

Output on this branch:

```
error: channel-receive: deadlock — channel is empty and no other thread can send
error: fiber-join: deadlock — joined fiber can never complete (all fibers blocked)
```

The identical program without the `thread-join!` line prints `42`. At f's receive, `refCount() == 1` (the child stub was released at join) and `live_child_threads == 0`, so `sharedWakeupPossible` is false — but main is one form away from sending. Local aliases of the channel share the single heap object, so they're invisible to the refcount; only *stubs* count.

The machinery for the local wake already works: a local `channel-send` on a promoted channel self-rings this thread's own registered notifier and the sweep flips the receiver (easy to verify by keeping the sender thread alive). What's wrong is only this entry check. Sketch: when `!wakeup_possible`, don't raise while other local fibers can still run — park anyway (and **register the notifier when parking**, since the self-ring is what wakes the fiber on a local send; the null-notifier branch above interacts badly with this) and reserve the raise for blockOrDeadlock's case: main fiber, nothing runnable. If the fix for the drive-park issue above routes dispatched fibers through yield_retry, both paths converge on exactly the local semantics the KEP points at.

src/shared_channel.zig
```diff
@@ -262,6 +301,32 @@ pub fn promoteChannel(gc: *memory.GC, ch: *types.Channel) !*SharedChannel {
...
+    // function returns), so gating here would always see 1 and wrongly
+    // skip migration.
+    if (vm_mod.vm_instance) |vm| {
+        if (vm.scheduler) |sched| {
+            const ch_val = types.makePointer(@ptrCast(&ch.header));
+            for (sched.fibers.items) |maybe_f| {
+                const f = maybe_f orelse continue;
+                if (f.status == .waiting and f.waiting_on == ch_val) {
+                    sc.registerRecvWaiter(vm.reactor.?.notifyHandle());
+                    sched.enrollSharedWaiter(f) catch {};
```
baijum: `catch {}` here swallows OOM and the failure mode is a permanent hang, not degraded service: the `registerRecvWaiter` call above already succeeded (or panicked), so a later remote send *will* ring this thread's notifier and the sweep *will* run — but this fiber isn't in the registry, and nothing else ever flips a shared-channel waiter. `promoteChannel` is already fallible — propagate with `try`, or at minimum match `registerRecvWaiter`'s own `@panic`-on-OOM so it fails loudly.

Nit: `registerRecvWaiter` is re-invoked once per matching fiber; dedup makes it idempotent, but it reads clearer hoisted out of the loop (register once if any waiter matched).

src/reactor.zig
```diff
@@ -63,18 +63,122 @@ const Reg = struct {
...
 pub const Reactor = struct {
     allocator: std.mem.Allocator,
     backend: Backend,
     regs: std.AutoHashMap(i32, Reg),
     timers: TimerHeap,
+    notifier: *ThreadNotifier,
 
     pub fn init(allocator: std.mem.Allocator) !Reactor {
+        var backend = try Backend.init(allocator);
+        errdefer backend.deinit();
```
baijum: Minor, OOM-path only: if the `create(ThreadNotifier)` on the next line fails, this `errdefer backend.deinit()` no longer releases the kernel resources — on kqueue `deinit` is now a no-op (the close moved to `releaseNotifier`, which never runs because no notifier exists), so the fresh `kq` leaks; on Linux the eventfd leaks (`EpollBackend.deinit` closes only `epfd`). Worth an explicit close of the backend's raw fds on this one path — e.g. an errdefer doing the same switch `releaseNotifier` does.

src/primitives_fiber.zig
baijum: Confirmed and fixed in 9cf68a7. You were exactly right: enrollment is orthogonal to the park mechanism. `channelReceiveShared` now always parks a scheduler-dispatched fiber via the flat `yield_retry` unwind (matching `blockOrDeadlock`'s existing shape for every non-main fiber) and never nests a drive after setting `.waiting`. Verified against your exact repro (two spawned fibers on separate promoted channels) — both now receive their real values instead of the stale callee register. Added it as a regression test ("review regression: two spawned fibers each parked on their own promoted channel do not corrupt each other's result").

src/primitives_fiber.zig
baijum: Confirmed and fixed in 9cf68a7, per your sketch. `sharedWakeupPossible()` now only gates the main-fiber path — a dispatched fiber parks unconditionally (same as blockOrDeadlock does for every non-main fiber today), and I added a "drive local siblings once" phase before the park decision so a purely local sibling send (self-ring) is tried first. Verified against your exact repro (transient promotion, then local sibling send) — prints 42 instead of raising. Added as a regression test.

src/shared_channel.zig
baijum: Fixed in 9cf68a7: propagates via `try` now instead of `catch {}` (promoteChannel is already fallible, matches registerRecvWaiter's own OOM-panics-loudly convention), and hoisted registerRecvWaiter out of the loop to run once per promotion rather than once per matching fiber.

src/reactor.zig
baijum: Fixed in 9cf68a7 by reordering: the notifier is now allocated before Backend.init(), so a notifier-allocation OOM after a successful backend init can no longer happen (nothing left that can fail after backend.init() succeeds), and a backend.init() failure has nothing of ours to clean up (each backend already unwinds its own partial resources internally).

src/primitives_fiber.zig
```diff
@@ -157,6 +159,90 @@ const ChannelWait = struct {
...
+
+        me.status = .waiting;
+        me.waiting_on = ch_val;
+        vm.gc.writeBarrier(&me.header, ch_val);
+        me.timed_out = false; // defensive: stale flag from an unrelated earlier timed wait
+        try ctx.sched.enrollSharedWaiter(me);
+        _ = try fiber_mod.runSchedulerStep(SharedChannelWait, .{ .me = me }, ctx.vm, ctx.sched, me);
+        ctx.sched.removeSharedWaiter(me);
```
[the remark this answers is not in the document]
baijum: Fixed in 9cf68a7: enrollSharedWaiter failures now roll back .waiting/waiting_on and propagate the error; runSchedulerStep errors in the park path get the same scoped cleanup (removeSharedWaiter + reset). Note the surrounding function was also restructured per a separate, more serious finding (confirmed VM corruption in the drive-park mechanism) — see the other review thread.

src/reactor.zig
```diff
@@ -63,18 +63,122 @@ const Reg = struct {
...
+                    .data = 0,
+                    .udata = 0,
+                }};
+                var zero_ts: std.c.timespec = .{ .sec = 0, .nsec = 0 };
+                _ = std.c.kevent(self.backend.kq, &triggers, 1, triggers[0..0].ptr, 0, &zero_ts);
+            },
+            .linux => {
+                const one: u64 = 1;
+                _ = std.posix.system.write(self.backend.fd, @ptrCast(&one), @sizeOf(u64));
+            },
```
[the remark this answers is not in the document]
baijum: Fixed in 9cf68a7: both the kevent and eventfd-write paths in notify() now retry on EINTR instead of dropping the wakeup.

src/reactor.zig
```diff
@@ -63,18 +63,122 @@ const Reg = struct {
...
     regs: std.AutoHashMap(i32, Reg),
     timers: TimerHeap,
+    notifier: *ThreadNotifier,
 
     pub fn init(allocator: std.mem.Allocator) !Reactor {
+        var backend = try Backend.init(allocator);
+        errdefer backend.deinit();
+        const notifier = try std.heap.c_allocator.create(ThreadNotifier);
+        errdefer std.heap.c_allocator.destroy(notifier);
+        notifier.* = .{ .backend = backend.notifierBackend() };
```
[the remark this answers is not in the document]
baijum: Fixed in 9cf68a7 by reordering Reactor.init to allocate the notifier before Backend.init(), matching your suggested diff.

src/shared_channel.zig
```diff
@@ -262,6 +301,32 @@ pub fn promoteChannel(gc: *memory.GC, ch: *types.Channel) !*SharedChannel {
...
+                }
+            }
```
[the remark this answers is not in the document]
baijum: Fixed in 9cf68a7: propagates via try instead of catch {}, and registerRecvWaiter is now called once per promotion rather than once per matching fiber, per your nit.

src/stress_channel.zig
```diff
@@ -0,0 +1,163 @@
...
+            switch (outcome) {
+                .value => |v| {
+                    _ = self.sum.fetchAdd(types.toFixnum(v), .monotonic);
+                    _ = self.count.fetchAdd(1, .monotonic);
+                },
+                .would_park => std.Thread.yield() catch {},
+                .eof => fail("consumer: unexpected eof (seed={d})\n", .{seed}),
+            }
+        }
+    }
```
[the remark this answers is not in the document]
baijum: Fixed in 9cf68a7: replaced count+sum with an identity-aware oracle (one flag per (producer_id, offset) pair, decoded directly from each value's disjoint range) that catches an out-of-range or duplicate delivery immediately, plus a bounded stall-detection deadline (30s, prints the seed) so a genuinely lost message fails loudly instead of spinning consumers forever.

src/tests_scheduler.zig
```diff
@@ -140,3 +140,97 @@ test "fiber switch does not grow the fiber's own register storage to match a lar
...
+    const f_val = try vm.eval("f");
+    const f = types.toObject(f_val).as(fiber_mod.Fiber);
+    // .waiting with no deadline trips none of the pre-existing categories
+    // (created/suspended/waiting-with-deadline/io_waiting) -- isolates the
+    // new shared_waiters check from every other reason this could pass.
+    f.status = .waiting;
+    try sched.enrollSharedWaiter(f);
+
+    try std.testing.expect(sched.hasRunnableFibers());
+}
```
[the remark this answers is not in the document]
baijum: Fixed in 9cf68a7: all 4 new tests now use th.TestContext.init()/deinit() instead of manual GC/makeTestVM setup.

src/tests_shared_channel.zig
```diff
@@ -744,3 +758,224 @@ test "KEP-0002 Phase 2: an exception raised in the child carrying a channel prom
...
+        var j: usize = 0;
+        while (j < per_producer) : (j += 1) {
+            expected_sum += @as(i64, @intCast(i * 1_000_000)) + @as(i64, @intCast(j));
+        }
+    }
+    try std.testing.expectEqual(expected_sum, received_sum.load(.monotonic));
+
+    sc.release();
+    try std.testing.expectEqual(baseline, shared_object.liveCount());
+    try std.testing.expectEqual(notifier_baseline, reactor_mod.notifierLiveCount());
```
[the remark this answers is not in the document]
baijum: Fixed in 9cf68a7: same identity-aware oracle + bounded stall deadline (10s) as stress_channel.zig, using std.debug.panic on violation to match this file's existing catch-unreachable convention for background-thread failures.

baijum:
Pushed 9cf68a7 addressing all review findings.

**Two confirmed, repro'd correctness bugs** (both in `channelReceiveShared`'s park decision, thank you for the precise repros):

1. **VM corruption**: a dispatched fiber set `.waiting` and kept driving other fibers in the same native frame, making its mid-call snapshot dispatchable by an unrelated fiber's own nested drive loop — resuming bytecode past the in-flight `receive()` call with the destination register never written. Fixed: a dispatched fiber now always parks via the flat `yield_retry` unwind (KEP-0002 §4 receive step 8), matching `blockOrDeadlock`'s existing behavior — never a nested drive after setting `.waiting`.
2. **False deadlock**: `sharedWakeupPossible()` was gating the dispatched-fiber path too, so any channel that was ever transiently promoted permanently broke ordinary local fiber-to-fiber use of it. Fixed: the gate now only applies to the main-fiber path (where it's genuinely load-bearing); a dispatched fiber parks unconditionally, and a "drive local siblings once" phase runs before the park decision so a local self-ring is tried first.

Both are now covered by dedicated regression tests using your exact repro shapes.

Also addressed all 7 CodeRabbit findings: OOM rollback + scoped cleanup around `enrollSharedWaiter`/`runSchedulerStep`, EINTR retry in `notify()`, `Reactor.init`'s allocation ordering, propagating (not swallowing) the migration's `enrollSharedWaiter` error, `th.TestContext` in the 4 new scheduler tests, and an identity-aware + bounded-deadline delivery oracle in both stress tests (replacing count+sum, which could in principle mask a lost delivery balanced by a duplicate).

Re-verified everything: `zig build test` + `-Dgc-stress=true` green (834 tests), full Scheme/R7RS suite 1835/0, `zig build stress-channel` across several seeds and a heavier config, `zig build wasm` compiles, `zig fmt` clean.

baijum:
Re-reviewed at 9cf68a7. **All four findings are fixed and verified — LGTM, ready to merge** (manually, once the still-pending `test (ubuntu-latest, Debug, 30)` job finishes; everything else is green).

## Verification against a build of 9cf68a7

**Finding 1 — VM corruption (drive-parking dispatched fibers):** fixed as prescribed by KEP §4 receive step 8 — dispatched fibers now enroll + `yield_retry` + `Yielded`, never a nested drive after `.waiting`. My original two-receiver repro now produces exactly `join f1 => 111`, `join f2 => 222`, `log = ((f2 222) (f1 111))` (previously: `f1` "received" `#<builtin channel-receive>` and 111 was lost). The new regression test matches the repro shape.

**Finding 2 — false deadlock on transiently promoted channels:** fixed — the `sharedWakeupPossible()` gate is main-fiber-only, the notifier registers unconditionally (so local sends self-ring), and the `SharedChannelPoll` pre-drive tries local siblings first. The transient-promotion repro now prints `42` (previously: spurious "no other thread can send"). Regression test added.

**Finding 3 — swallowed enroll OOM in `promoteChannel`:** fixed (`try` propagates; `registerRecvWaiter` hoisted to once-per-promotion).

**Finding 4 — `Reactor.init` fd leak:** fixed by allocating the notifier before the backend, with a clear comment; the `notify()` EINTR retry is a good bonus catch.

I also traced the new structure for the hazard the original drive-park had, and it holds up: the `SharedChannelPoll` drive keeps `me` `.running` (never dispatchable), and the main/native-frame `.waiting` park is only reachable after `parkOnReactor` returned false — i.e. no runnable/io/deadline fibers, an empty shared-waiter registry, and an empty reactor — so no nested loop can exist to flip-and-dispatch a mid-call snapshot; only the parked fiber's own loop can consume its wake. The remaining dirty-snapshot exposure is exactly the pre-existing mutex/condvar/thread-sleep class, unchanged by this PR (follow-up issue material, as discussed).

Locally: full `zig build test` green (including both new regressions), `zig build stress-channel -- 424242 8 8 500` PASS, fiber smoke tests pass. The identity-aware delivery oracle in the stress tests is a real improvement over count+sum.

## Two residual nits, non-blocking (fine as a follow-up or not at all)

1. The comment inside `promoteChannel`'s migration loop still says "registerRecvWaiter above (once hoisted) already succeeded or @panic'd" — registration now happens **after** the loop, so on an enroll OOM mid-loop the notifier was never registered and earlier-enrolled fibers stay enrolled-but-unregistered (stale registry entries linger until some sweep; the failed promotion leaves the channel local, so the fibers themselves are woken by a later local `wakeChannelWaiters`). OOM-only corner of an already-failing promotion — a comment fix plus optionally un-enrolling on the error path would tidy it.
2. Belt-and-braces: hoisting `me.timed_out = false` above the `SharedChannelPoll` drive would keep a (currently unreachable) stale `timed_out` from short-circuiting the poll and letting main reach its park with the world still active — the quiescence-before-park invariant is what makes the main path safe, so cheap insurance there is worth it.

And carrying over the two non-blocking notes from the first review for the tracking issue: a docs/comment mention of the KEP-accepted peer-death hang, and a follow-up issue for the pre-existing dirty-snapshot class (e.g. a `.driving` status `schedule()` never selects).

PR #1486, opened 2026-07-13
Address PR #1485 review nits: stale comment, timed_out hoist, peer-death doc

baijum:
## Summary

Closes out the three non-blocking items carried over from the final review round of PR #1485 (KEP-0002 Phase 3, #1468), which was merged without acting on them.

- `promoteChannel`'s migration-loop comment claimed `registerRecvWaiter` ran "above" the loop; it actually runs once, after it. Corrected the comment, and documented the OOM-mid-loop corner this creates: fibers enrolled by earlier iterations become stale-but-self-healing registry entries (any future sweep on the scheduler flips them regardless of channel, and their retry re-registers), while fibers not yet reached stay `.waiting` with no path left to wake them. Both are OOM-only corners of an already-failing promotion — accepted rather than un-enrolling on that error path, per the reviewer's own framing of this as optional.
- `channelReceiveShared` reset `me.timed_out` *after* the `SharedChannelPoll` local-sibling drive instead of before it. `runSchedulerStep`'s generic loop bails whenever `me.timed_out` is true regardless of `Ctx`, so a stale flag left by an unrelated earlier timed wait could silently skip the drive and let the park decision run with the world still active. Hoisted the reset above the drive (verified nothing between the new reset point and the old one can set it true again, since `me` stays `.running` for the whole drive and `wakeReadyFiber` only touches `.waiting` fibers).
- Documented the KEP-0002 §5 accepted liveness gap on `sharedWakeupPossible`: a receiver parked after this returns `true` hangs forever if the peer (the only other stub holder, or the only other live thread) exits without ever sending — a stub release rings nothing, and the parked fiber's own shared-waiter registry entry keeps `hasRunnableFibers()` true, so the deadlock detector never fires either. This is the same Go-style hang as an unbuffered channel with no reader; §5 accepts it deliberately, and §6's planned per-call timeout (`(channel-receive ch [timeout [timeout-val]])`, Phase 4) is the intended escape hatch.

No functional/API change — comments plus a defensive ordering fix for a currently-unreached path.

## Test plan

- [x] `zig build test` — unit suite green
- [x] `zig build` — clean build
- [x] `bash tests/scheme/run-all.sh` — 1835 pass, 0 fail (440 Scheme files + 1395 R7RS)
- [x] `zig build stress-channel -- 424242 8 8 500` — PASS
- [x] `zig fmt --check` on changed files

### DIFF
--- a/src/primitives_fiber.zig
+++ b/src/primitives_fiber.zig
@@ -186,6 +186,18 @@ const SharedChannelPoll = struct {
 /// its own stub, per §1), or some other OS thread is alive at all. Reuses
 /// primitives_srfi18.zig's existing crossThreadWaitPossible rather than
 /// duplicating its live-thread-count logic.
+///
+/// Accepted liveness gap, KEP-0002 §5: this is checked once, right before
+/// parking -- not re-evaluated while parked. If the peer that made this
+/// return `true` (the only other stub holder, or the only other live
+/// thread) exits afterward without ever sending, the parked receiver hangs
+/// forever: a stub's release doesn't ring anything, and the parked fiber
+/// staying enrolled in the shared-waiter registry keeps hasRunnableFibers()
+/// true, so the deadlock detector never fires either. This is the same
+/// Go-style "send on a channel nobody will ever receive from" hang as an
+/// unbuffered channel with no reader; §5 accepts it deliberately, and §6's
+/// planned `(channel-receive ch [timeout [timeout-val]])` (Phase 4, not yet
+/// implemented) is the intended escape hatch.
 fn sharedWakeupPossible(sc: *shared_channel.SharedChannel) bool {
     return sc.refCount() > 1 or srfi18.crossThreadWaitPossible();
 }
@@ -274,6 +286,13 @@ fn channelReceiveShared(sc: *shared_channel.SharedChannel, gc: *memory.GC, ch_va
 
         // Drive local siblings once. `me` stays .running -- see
         // SharedChannelPoll's doc comment for why that's load-bearing.
+        // Reset first: runSchedulerStep's generic loop bails whenever
+        // `me.timed_out` is true, regardless of Ctx, so a stale flag left
+        // by an unrelated earlier timed wait would silently skip this
+        // drive entirely and let the park decision below run with the
+        // world still active -- quiescence-before-park is what makes that
+        // decision safe.
+        me.timed_out = false;
         _ = try fiber_mod.runSchedulerStep(SharedChannelPoll, .{ .sc = sc }, ctx.vm, ctx.sched, me);
         if (sc.peekReady()) continue; // loop back to 1: re-derive via receive()
 
@@ -296,7 +315,10 @@ fn channelReceiveShared(sc: *shared_channel.SharedChannel, gc: *memory.GC, ch_va
         me.status = .waiting;
         me.waiting_on = ch_val;
         vm.gc.writeBarrier(&me.header, ch_val);
-        me.timed_out = false; // defensive: stale flag from an unrelated earlier timed wait
+        // Already reset above, before the SharedChannelPoll drive -- nothing
+        // between there and here can set it true again (`me` stays .running
+        // for that whole drive, and wakeReadyFiber only flips `.waiting`
+        // fibers).
         ctx.sched.enrollSharedWaiter(me) catch |err| {
             me.status = .running;
             me.waiting_on = types.VOID;
--- a/src/shared_channel.zig
+++ b/src/shared_channel.zig
@@ -337,12 +337,17 @@ pub fn promoteChannel(gc: *memory.GC, ch: *types.Channel) !*SharedChannel {
             for (sched.fibers.items) |maybe_f| {
                 const f = maybe_f orelse continue;
                 if (f.status == .waiting and f.waiting_on == ch_val) {
-                    // Propagate, don't swallow: registerRecvWaiter above
-                    // (once hoisted) already succeeded or @panic'd on OOM,
-                    // so a remote send WILL ring this thread's notifier --
-                    // silently dropping this fiber's enrollment here would
-                    // leave it .waiting with nothing left to ever flip it
-                    // back, a permanent hang.
+                    // Propagate, don't swallow: registerRecvWaiter runs
+                    // once, after this loop finishes (not before it), so on
+                    // an enrollSharedWaiter OOM mid-loop no notifier is ever
+                    // registered for `sc`. Fibers already enrolled by prior
+                    // iterations become stale-but-harmless registry entries
+                    // -- any future sweep on this scheduler flips them
+                    // regardless of channel, and their retry re-registers
+                    // per §4 -- while fibers not yet reached stay `.waiting`
+                    // with no path left to wake them. Both are OOM-only
+                    // corners of an already-failing promotion; accepted
+                    // rather than un-enrolling on this error path.
                     try sched.enrollSharedWaiter(f);
                     migrated_any = true;
                 }


baijum (commented):
## Review

**Verdict: the changes themselves are correct and verified — but the PR is not mergeable as-is.** Two mechanical blockers, then it's good to merge manually.

### Blocker 1: stale base → CONFLICTING

The branch is stacked on #1485's original commits (`d969ad98`, `9cf68a7b`), while `main` has them squash-merged as `e4510f96`. GitHub reports the PR `CONFLICTING`/`DIRTY` (verified locally: `git merge-tree` conflicts in both changed files). Consequences:

- The PR cannot merge at all right now.
- The Files tab shows **12 files, +1418/−75** — almost all of it is #1485's already-merged content. The real effective change is **2 files, +34/−7** (`git diff origin/main origin/fix/1468-followups`). CodeRabbit's walkthrough demonstrates the damage: it reviewed #1485's diff, not this PR's.
- **CI never ran** — a conflicted PR has no merge ref for Actions, so the only check on this PR is CodeRabbit. The test plan's local runs are unverifiable from the PR page.

The rebase is **guaranteed conflict-free**: `9cf68a7b^{tree}` and `e4510f96^{tree}` are the identical tree (`a53a2b9b…`), so replaying the one follow-up commit applies cleanly:

```bash
git fetch origin
git rebase --onto origin/main 9cf68a7b fix/1468-followups
git push --force-with-lease origin fix/1468-followups
```

That also collapses the diff to the real 2-file change and lets CI run.

### Blocker 2: wait for green CI after the rebase

No auto-merge is armed (good — repo convention is manual merge). Merge manually once the rebased CI is green.

### What I verified (all claims checked against code, on head `95c6d752`)

- **`timed_out` hoist** (`channelReceiveShared`): correct and strictly safer — details in the inline comment. The PR body's "currently-unreached" characterization is accurate; the hoist also quietly fixes the `is_dispatched` path, which previously never reset the flag at all.
- **`promoteChannel` comment**: now matches the code (`registerRecvWaiter` runs once, after the loop, gated on `migrated_any`); the old text described a hoist that never happened. OOM-corner analysis verified — inline comment.
- **Peer-death doc on `sharedWakeupPossible`**: every factual claim verified (`release` never rings; `hasRunnableFibers()` short-circuits on the registry; checked once before park) — inline comment.
- **Local verification, exact head sha**: `zig build test` green, `zig build` clean, `zig fmt --check` clean on both changed files (macOS ARM).

### CodeRabbit triage (both anchored to #1485 code the stale diff re-showed)

- **"Root the channel value across recursive scheduling" — false positive.** `ch_val` stays reachable through the saved fiber snapshot: `runSchedulerStep` calls `saveCurrentFiber()` before any dispatch, `FiberScheduler.markRoots` marks every fiber object, and the object tracer's `.fiber` arm calls `markFiberState` **unconditionally** (`gc_collect.zig:436-440` and `:698-702`) — the `.running` skip inside `markRoots` only avoids a redundant second call, it does not skip tracing. Please don't apply the suggested `pushRoot`; it would add noise for a non-bug. (The #1485 gc-stress runs would also have caught this if real.)
- **"Re-register after the local drive to prevent a lost wakeup" — real, confirmed with a repro, and pre-existing in `main`.** A local sibling's send+receive pair during the `SharedChannelPoll` drive clears the one-shot `recv_waiters` registration and the drive swallows `wake_pending`; the `peekReady()` fall-through then parks unregistered, and a later remote send rings nothing — permanent silent hang (deadlock detector suppressed by the registry entry, by design). Repro hangs on both `main` (`e4510f96`) and this PR's head; control without the decoy fiber passes. Filed as **#1489** with the repro and fix direction — out of scope for this comments-only PR, but flagging it here so it isn't lost when the rebase orphans the bot's comment.

### Summary

Content: LGTM. Process: rebase onto `main` (trivial, commands above), let CI go green, then merge manually. Please make sure #1489 stays tracked — it's a confirmed hang in a shipped path.

src/primitives_fiber.zig
```diff
@@ -157,6 +159,183 @@ const ChannelWait = struct {
...
+
+        // Drive local siblings once. `me` stays .running -- see
+        // SharedChannelPoll's doc comment for why that's load-bearing.
+        // Reset first: runSchedulerStep's generic loop bails whenever
+        // `me.timed_out` is true, regardless of Ctx, so a stale flag left
+        // by an unrelated earlier timed wait would silently skip this
+        // drive entirely and let the park decision below run with the
+        // world still active -- quiescence-before-park is what makes that
+        // decision safe.
+        me.timed_out = false;
```
baijum, line 295: **Hoist verified — all claims check out, plus one bonus the PR body undersells.**

- "bails whenever `me.timed_out` is true, regardless of Ctx": confirmed, `fiber.zig:728` (`while (!ctx.isDone() and !me.timed_out)`).
- "nothing between there and here can set it true again": confirmed — the *only* `timed_out = true` writer in the codebase is `wakeReadyFiber` (`fiber.zig:576`), gated on `.waiting`, and only reachable for fibers the reactor reported ready (an fd or expired timer, neither of which `me` registers here). `me` stays `.running` through the drive, so neither condition can match it.
- "currently-unreached" (PR body): also accurate today — every SRFI-18 timed wait consumes the flag on every exit path (`thread-sleep!`, the only yield-retry timed wait, consumes it on redispatch at `primitives_srfi18.zig:554-557`; join/mutex/condvar check-and-reset right after their drives). The stranding paths I could find are OOM unwinds (e.g. `threadSleepFn`'s errdefer restores `deadline_ns` but not a `timed_out` set mid-drive) — exactly what a defensive reset is for.

Bonus worth a word in the comment if you touch it again: the old reset lived only on the main-fiber path, so the `is_dispatched` park never cleared a stale flag at all — it would have skipped the sibling drive on **every** redispatch. The hoist fixes that asymmetry too. Consistent with `waitForFd`'s precedent (`fiber.zig:650-653`). LGTM.

src/primitives_fiber.zig
```diff
@@ -157,6 +159,183 @@ const ChannelWait = struct {
...
+/// parking -- not re-evaluated while parked. If the peer that made this
+/// return `true` (the only other stub holder, or the only other live
+/// thread) exits afterward without ever sending, the parked receiver hangs
+/// forever: a stub's release doesn't ring anything, and the parked fiber
+/// staying enrolled in the shared-waiter registry keeps hasRunnableFibers()
+/// true, so the deadlock detector never fires either. This is the same
+/// Go-style "send on a channel nobody will ever receive from" hang as an
+/// unbuffered channel with no reader; §5 accepts it deliberately, and §6's
+/// planned `(channel-receive ch [timeout [timeout-val]])` (Phase 4, not yet
+/// implemented) is the intended escape hatch.
```
baijum, line 200: **Peer-death doc verified, claim by claim:**

- "a stub's release doesn't ring anything": confirmed — `SharedChannel.release` → `shared_object.release` is a bare refcount decrement, and even the final-release `destroyHook` only *releases* the waiter notifiers, never rings them.
- "keeps hasRunnableFibers() true, so the deadlock detector never fires": confirmed — `fiber.zig:348` short-circuits true on a non-empty `shared_waiters` registry, and `parkOnReactor`'s genuine-deadlock `return false` requires it false.
- "checked once, right before parking — not re-evaluated while parked": matches the call site below.

(§5/§6 section numbers aren't verifiable in-repo — KEP-0002 lives in kaappi/keps#3 — but the semantics match the accepted design.)

One heads-up: while re-verifying this park path I confirmed a *different* liveness hole in it — a receiver can hang even when the peer **does** send, if a local sibling's send+receive pair disarmed the one-shot notifier registration during the `SharedChannelPoll` drive. Repro'd on main, filed as #1489. Doesn't invalidate this doc (it covers the peer-never-sends case), but the Phase 4 timeout escape hatch this comment points to just got a second customer.

src/shared_channel.zig
```diff
@@ -262,6 +313,49 @@ pub fn promoteChannel(gc: *memory.GC, ch: *types.Channel) !*SharedChannel {
...
+                    // once, after this loop finishes (not before it), so on
+                    // an enrollSharedWaiter OOM mid-loop no notifier is ever
+                    // registered for `sc`. Fibers already enrolled by prior
+                    // iterations become stale-but-harmless registry entries
+                    // -- any future sweep on this scheduler flips them
+                    // regardless of channel, and their retry re-registers
+                    // per §4 -- while fibers not yet reached stay `.waiting`
+                    // with no path left to wake them. Both are OOM-only
+                    // corners of an already-failing promotion; accepted
+                    // rather than un-enrolling on this error path.
```
baijum, line 350: **Corrected comment verified against the code.** `registerRecvWaiter` does run once, after the loop, gated on `migrated_any` — the old "above (once hoisted)" text described a hoist that never happened, so this fixes a comment that was actively wrong about control flow.

The OOM-corner analysis holds too:
- already-enrolled fibers: `sweepSharedWaiters` flips every registry entry channel-agnostically (`fiber.zig:439-452`), and the woken retry re-registers via `receive()` — self-healing on any future ring, as stated;
- not-yet-reached fibers: genuinely no wake path left — `ch.shared` was published before the drain (above), so local sends take the shared path and never call `wakeChannelWaiters`, and this scheduler's notifier was never registered.

Accepting both as OOM-only corners of an already-failing promotion is reasonable. LGTM.

baijum:
**Re-checked after the rebase — ready to merge.** ✅

Both blockers from my review are resolved:

1. **Stale base / CONFLICTING** — fixed: the branch is now a single commit (`5e206006`) directly on main's tip (`e4510f96`), the PR reports `MERGEABLE`/`CLEAN`, and the diff collapsed to the real change (2 files, +34/−7).
2. **CI never ran** — now fully green on the rebased head: format, all five test matrix jobs (ubuntu ReleaseSafe/ReleaseFast/Debug, ubuntu-arm, macos), riscv64-test, wasm, and benchmark-pr all pass (benchmark/coverage are conditional skips, as usual).

The rebased commit is **byte-identical** to the head I reviewed (`git diff 95c6d752 5e206006` is empty), so the content verification carries over 1:1 — no re-review needed.

No auto-merge is armed — merge manually at your convenience. Reminder: the confirmed lost-wakeup hang found during review is tracked separately in #1489 and is not affected by this PR.

PR #1488, opened 2026-07-13
KEP-0004 Phase 0/1: unify cond-expand library checks, add subsystem features

baijum:
## Summary

Implements Phase 0 and Phase 1 of [KEP-0004](https://github.com/kaappi/keps/blob/main/keps/0004-discoverable-deviations.md) (Discoverable Deviations from R7RS-small).

**Phase 0 (cleanup, not a fix).** `evalFeatureReq`'s hardcoded `known_libs`
array in `compiler_conditionals.zig` never listed any `kaappi.*` library or
`srfi.18` — but this went unnoticed because it silently fell through to the
`globals.libraryExists` callback, which already checked the live registry
correctly. Verified empirically against a built binary before touching
anything: `(cond-expand ((library (kaappi fibers)) ...))` and `(srfi 18)`
already resolved correctly *before* this change, via the fallback the array
never reached. Deleted the array outright, and extracted the duplicated
"does this library exist" check — previously hand-written separately in
`vm.zig`'s `checkLibraryExists` and `vm_library.zig`'s `evalLibFeatureReq`
— into one shared `vm_library.libraryIsAvailable()`, called from both entry
points.

An earlier draft of this design proposed routing the compiler's evaluator
through the `vm.vm_instance` threadlocal directly. That turned out to be
architecturally wrong: `vm.zig` imports `compiler.zig`, so the reverse
import would close a real cycle. The KEP doc has been corrected to describe
what's actually implemented here instead.

**Phase 1.** Added `kaappi-fibers`, `kaappi-reactor` (all targets, including
wasm32-wasi per KEP-0001 Phase 4), and `kaappi-threads` (omitted on wasm,
matching `Lib.wasmAvailable()`'s existing `srfi_18 => false` gate) as bare
`cond-expand` feature identifiers.

## Test plan

- [x] `zig build` — clean
- [x] `zig build test` — 839/839 pass (5 new: 2 known_libs-removal
      regression tests, 3 new-identifier tests)
- [x] `bash tests/scheme/run-all.sh` — 440/440 Scheme files, 1395/1395 R7RS
      suite
- [x] Manual end-to-end verification against a built binary for every new
      identifier and every `(library ...)` check that used to hit
      `known_libs`

### DIFF
--- a/src/vm_library.zig
+++ b/src/vm_library.zig
@@ -346,6 +346,11 @@ pub fn resolveLibraryPath(allocator: std.mem.Allocator, rel_path: []const u8, li
 /// directly). (KEP-0004)
 pub fn libraryIsAvailable(vm: *VM, lib_name: []const u8, lib_name_list: Value) bool {
     if (vm.libraries.get(lib_name) != null) return true;
+    // tryLoadLibraryFromFile (below) rejects every file-backed load when
+    // sandboxed; without this check cond-expand would report a disk-only
+    // library as available while the matching import then fails, and would
+    // let sandboxed code probe the host filesystem for .sld existence.
+    if (vm.sandbox_mode) return false;
     return libraryFileExists(vm, lib_name_list);
 }
 
--- a/src/vm_library.zig
+++ b/src/vm_library.zig
@@ -338,6 +338,17 @@ pub fn resolveLibraryPath(allocator: std.mem.Allocator, rel_path: []const u8, li
     return null;
 }
 
+/// Whether library `lib_name` (canonical dotted form) is loaded or loadable
+/// from disk. The single implementation behind both cond-expand (library ...)
+/// entry points: evalLibFeatureReq below (inside define-library, has *VM
+/// directly) and vm.zig's checkLibraryExists (the callback the compiler uses
+/// for top-level/expression-position cond-expand, which cannot import vm.zig
+/// directly). (KEP-0004)
+pub fn libraryIsAvailable(vm: *VM, lib_name: []const u8, lib_name_list: Value) bool {
+    if (vm.libraries.get(lib_name) != null) return true;
+    return libraryFileExists(vm, lib_name_list);
+}
+
 /// Check whether a library could be loaded from disk (or the bundled files of
 /// a standalone binary), using the same search order as import. Used by the
 /// cond-expand (library ...) feature checks so they agree with what import
@@ -525,8 +536,7 @@ fn evalLibFeatureReq(vm: *VM, req: Value) bool {
             const lib_name_list = types.car(rest);
             const lib_name = library_mod.libraryNameToString(vm.gc.allocator, lib_name_list) catch return false;
             defer vm.gc.allocator.free(lib_name);
-            if (vm.libraries.get(lib_name) != null) return true;
-            return libraryFileExists(vm, lib_name_list);
+            return libraryIsAvailable(vm, lib_name, lib_name_list);
         }
     }
     return false;
--- a/src/compiler_conditionals.zig
+++ b/src/compiler_conditionals.zig
@@ -300,38 +300,20 @@ pub fn evalFeatureReq(self: *Compiler, req: Value) bool {
         }
 
         if (std.mem.eql(u8, op, "library")) {
-            // (library (name ...)) — check if library exists
-            // We don't have direct access to the library registry from the compiler,
-            // but we can check against known standard libraries
+            // (library (name ...)) — check if library exists. The compiler
+            // has no direct VM access (vm.zig already imports compiler.zig,
+            // so the reverse import would cycle), so this goes through the
+            // VM-registered globals.libraryExists callback, which checks the
+            // live library registry and .sld files on disk (see vm.zig's
+            // checkLibraryExists / vm_library.libraryIsAvailable). (KEP-0004)
             const rest = types.cdr(req);
             if (!types.isPair(rest)) return false;
             const lib_name_list = types.car(rest);
 
-            // Convert library name list to canonical string
             const lib_name = @import("library.zig").libraryNameToString(self.gc.allocator, lib_name_list) catch return false;
             defer self.gc.allocator.free(lib_name);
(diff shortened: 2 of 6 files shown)


src/vm_library.zig
```diff
@@ -338,6 +338,17 @@ pub fn resolveLibraryPath(allocator: std.mem.Allocator, rel_path: []const u8, li
...
 
+/// Whether library `lib_name` (canonical dotted form) is loaded or loadable
+/// from disk. The single implementation behind both cond-expand (library ...)
+/// entry points: evalLibFeatureReq below (inside define-library, has *VM
+/// directly) and vm.zig's checkLibraryExists (the callback the compiler uses
+/// for top-level/expression-position cond-expand, which cannot import vm.zig
+/// directly). (KEP-0004)
+pub fn libraryIsAvailable(vm: *VM, lib_name: []const u8, lib_name_list: Value) bool {
+    if (vm.libraries.get(lib_name) != null) return true;
+    return libraryFileExists(vm, lib_name_list);
```
baijum, line 354: **Consolidating the two call sites here is the natural place to also close the sandbox gap CodeRabbit flagged.** Under `--sandbox`, `tryLoadLibraryFromFile` rejects every file-backed load (line 428), but `libraryIsAvailable` still falls through to `libraryFileExists` and probes the filesystem — so `(cond-expand ((library (srfi 41)) ...))` reports the lib as available while `(import (srfi 41))` fails. I verified this against the built binary:

```
$ kaappi --sandbox condexpand.scm   -> cond-expand (library (srfi 41)): available
$ kaappi --sandbox import.scm        -> error: sandbox: cannot load library from file
```

Two clarifications on top of CodeRabbit's note:
- **This is pre-existing, not a regression** — `main`'s `checkLibraryExists` and `evalLibFeatureReq` had the identical `get() ... libraryFileExists()` sequence, so the PR faithfully preserves behavior.
- Because this PR unifies both sites into one function, a single `if (vm.sandbox_mode) return false;` after the registry lookup now fixes it in one place (already-loaded libs stay reported, only the disk probe is gated). Optional and non-blocking, but a clean win given the PR's theme.

baijum (commented):
### Review — KEP-0004 Phase 0/1

Clean, well-scoped refactor + additive feature. I checked out the head, built, and verified behavior against the binary rather than trusting the description.

**Verified working**
- Bare ids `kaappi-fibers` / `kaappi-reactor` / `kaappi-threads` all resolve in `cond-expand`; `(not kaappi-fibers)` -> `no`; `(features)` now advertises them (correct R7RS behavior).
- `known_libs` removal is safe: `(library (scheme base))`, `(library (srfi 18))`, `(library (kaappi fibers))` all resolve through the callback path, and a bogus lib -> `no`. The pre-existing `"cond-expand library check"` test now exercises the callback for `scheme base` (its old fast-path is gone) and still passes.
- wasm detection (`builtin.os.tag == .wasi`) matches `library.zig` / `primitives.zig`, so `kaappi-threads` tracks `Lib.wasmAvailable()`'s `srfi_18 => false` gate. wasm CI is green.
- `features-consistency.scm` (#1177) uses `memq`, so the additive features don't break it. No other consumers of `platform_features` / the changed symbols break.
- `zig build` clean; `zig build test -Dtest-filter=cond-expand` passes.

**Comments** (all non-blocking)
1. **Sandbox probe** — see inline on `libraryIsAvailable`. Pre-existing, but this PR is the natural single place to gate it.
2. **Test setup duplication** (CodeRabbit) — the 5 new tests repeat the GC/VM boilerplate, but they match the existing `cond-expand` tests in the same file, and no `th.expectEval*` helper covers symbol results. Fine as-is; `th.TestContext` would be tidier if you touch them.
3. **wasm-branch coverage** (CodeRabbit) — the wasm branch of `platform_features` isn't unit-tested, but the unit suite doesn't run on wasm targets so a wasm-only assertion would never execute. Marginal; the author's inline note already explains this.
4. Optional: extend `features-consistency.scm` with the three new ids for Scheme-level parity with the new Zig tests.

**Merge readiness:** Ready to merge once the remaining CI jobs (macOS/ubuntu ReleaseSafe/Fast/Debug, riscv64) go green — wasm/format/CodeRabbit already pass. No blocking issues; the sandbox item is a pre-existing, optional consolidation.

baijum (commented):
### Follow-up review — commit 76919e0

Re-checked out the head, rebuilt, and verified the two actionable items are resolved:

**1. Sandbox probe (finding #1) — fixed.** `libraryIsAvailable` now returns `false` before the disk probe when `vm.sandbox_mode` is set, with a clear comment. Verified against the binary:

| | `--sandbox` | no sandbox |
|---|---|---|
| `(cond-expand ((library (srfi 41)) ...))` | `unavailable` ✅ | `available` ✅ |
| `(import (srfi 41))` | sandbox error ✅ | loads ✅ |

cond-expand and `import` now agree under sandbox. Importantly, the gate is correctly placed *after* the `vm.libraries` lookup, so built-in libraries aren't over-gated — `(cond-expand ((library (scheme base)) ...))` still reports `available` under `--sandbox`. The new `"cond-expand library check honors sandbox mode"` regression test (asserting `yes` unsandboxed → `no` sandboxed, with a source-tree skip guard) passes.

**2. features-consistency.scm (finding #4) — done.** The three new ids are now asserted at both the `(features)` and expression-`cond-expand` levels; the suite passes 14/14.

Findings #2 (test-helper style) and #3 (wasm-branch coverage) were reasonably left as-is — both non-blocking, and #3 isn't unit-testable on a native host anyway.

`zig build` clean; `test -Dtest-filter=cond-expand` and the features-consistency smoke test both green. CI on this commit is green across ubuntu (ReleaseSafe/Fast/Debug), ubuntu-arm, riscv64, wasm, format, and benchmark; only macOS ReleaseSafe is still running.

**Verdict: ready to merge** once macOS ReleaseSafe finishes green. No outstanding blocking issues. (The KEP-0004 doc itself lives in the `keps` repo and is outside this diff.)

Issue #1490, opened 2026-07-13
Nested runSchedulerStep can corrupt an ancestor fiber's frame state, crashing with "integer overflow" in invokeEscape


baijum:
## Summary

A spawned fiber's blocking wait, when nested inside another fiber's own timed wait (`thread-sleep!`, or any primitive using the `me.deadline_ns`/`runSchedulerStep` pattern), can crash the process with `panic: integer overflow` inside `invokeEscape` (`vm_dispatch.zig`). This is **pre-existing** — reproducible with only `thread-sleep!` and a plain, untimed `channel-receive`, no KEP-0002 Phase 4 code involved.

Found while implementing KEP-0002 Phase 4 (#1469, channel timeouts) — timeouts on `channel-receive`/`channel-send` make this nesting pattern much more likely to occur in ordinary programs (a spawned worker doing a timed channel op, polled by a driving loop that itself sleeps), so it's worth fixing, but the root cause is in the core fiber scheduler and predates Phase 4 entirely.

## Minimal repro (no channels needed)

```scheme
(import (scheme base) (scheme write) (kaappi fibers) (srfi 18))

(guard (outer-err (#t (display "outer-caught") (newline)))
  (let ((ch (make-channel)))
    (spawn (lambda () (channel-receive ch))) ; blocks forever: no sender
    (let loop ((n 0))
      (when (< n 20)
        (thread-sleep! 0.01)
        (loop (+ n 1))))
    (display "done") (newline)))
```

This crashes with:

```
thread NNNNN panic: integer overflow
src/vm_dispatch.zig:1232:25 in runUntil
                try self.invokeEscape(cont, cont_val);
...
src/vm_calls.zig:664:29 in callHandler
src/primitives_control.zig:257:34 in callWithEscapeContinuation
...
```

(A top-level `guard` isn't required to observe the underlying corruption — the crash site is one place it happens to surface, when nothing else has already crashed first; the minimal repro above only needs the outer `guard` because with no exception handler live at all, a corrupted escape continuation call has no code path to reach through.)

## Root cause (traced, not yet fixed)

[...]

Issue #1469, opened 2026-07-12
Capacity, timeouts, close: bounded channels, reservation-aware drain-then-EOF (KEP-0002 P4)


baijum:
Part of #1465 (KEP-0002, Phase 4). Implements the amended §6 close semantics — EOF outwaits reservations (model finding 2), and the failure path rings receivers on a closed channel (model finding 3); see kaappi/keps#12.

## Scope

- `(make-channel)` / `(make-channel capacity)` — capacity on both representations; shared-side send parking via §4 slot reservation, local-side via `waiting_on` (bounded channels do **not** force promotion; fiber-only backpressure stays lock-free).
- Timeouts on **both** operations: `(channel-receive ch [timeout [timeout-val]])`, `(channel-send ch v [timeout [timeout-val]])` — SRFI-18 shape, reactor timer heap, expiry raises without `timeout-val` / returns it with; stale notifier registrations cleaned by the §7 lifecycle (a stale ring is one harmless spurious sweep).
- `(channel-close! ch)` / `(channel-closed? ch)` — idempotent; snapshot-and-clear **both** lists, ring everyone, wake local waiters. Receivers drain the queue, then return `(eof-object)` only once `closed ∧ queue empty ∧ reserved == 0`; a receiver arriving inside an admitted send's copy window parks and is rung by the late push or by the failure ring.
- Local-channel `capacity`/`closed` fields carried into promotion (§2 step 1); `dequeueChannel` slot-freed wake on the local path.
- Sending a literal `(eof-object)` rejected before §4 step 1 (both representations): the one backward-compat carve-out.

## Acceptance

- The four required interleaving tests from the model (KEP-0002 Phase 4 text): reserve → concurrent close → push completes and the message is drained before EOF; all until-eof workers racing a reserved send across close — the admitted task is drained, not destroyed (finding 2); copy failure of the last reserved send on a closed channel wakes eof-waiting receivers (finding 3); failed receive re-queues at the head with in-order redelivery, including the transient capacity overshoot on a bounded channel.
- Timeout matrix: send/receive × with/without `timeout-val` × local/shared, matching `thread-join!` semantics.
- `research/tla/run.sh` green at the implemented-commit claim (this phase completes the modeled surface).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1502, opened 2026-07-13
KEP-0002 Phase 4: capacity, timeouts, close for cross-thread channels

baijum:
## Summary

Implements KEP-0002 Phase 4 (issue #1469): the last phase on the critical path before the `(kaappi parallel)` library (Phase 5) can be built.

- **`(make-channel capacity)`** — bounded channels. `channel-send` on a full channel parks (local) or reserves-and-parks (shared, §4) until a receive frees a slot. `(make-channel 0)` is legal but permanently full (no rendezvous handoff — documented, not a bug).
- **`[timeout [timeout-val]]`** on both `channel-send` and `channel-receive`, SRFI-18 shape (same as `thread-join!`): without `timeout-val`, expiry raises `channel-timeout-exception?`; with one, it's returned instead. Works on both the local and shared (cross-thread) representations, including as the documented escape hatch for §5's weakened cross-thread deadlock detection.
- **`channel-close!` / `channel-closed?`** — first-class end-of-stream. Idempotent; wakes all waiters both directions; receivers drain whatever's queued before returning `(eof-object)`; sends after close raise; sending a literal `(eof-object)` is rejected (the one documented backward-compat carve-out).

The §4–6 protocol (reservation-as-admission, the `reserved == 0` EOF guard, unconditional waiter sweep) was already model-checked in Phase 1–3 — I re-ran `keps/research/tla/run.sh` (9 TLC configs) against the KEP's exact pseudocode before writing any code, and all 9 matched their expected pass/fail outcome. Most of the low-level protocol (`shared_channel.zig`'s `send`/`receive`) was already implemented and white-box tested from Phase 1; this phase wires `capacity`/`closed` through the Scheme-visible API, adds the local-channel equivalent, and adds `close()`.

## Also fixed

- Local-channel dispatch (`channelSendFn`/`channelReceiveFn`) was missing the `me.status = .waiting` transition needed for a reactor timer to actually resolve a local wait — without it, `wakeReadyFiber` silently no-ops on a `.running` fiber and a local timeout never fires. Caught by a real end-to-end run, not by the unit suite (which doesn't exercise the local-representation timeout path the same way).

## Found, filed separately (not fixed here — out of scope)

While testing, hit a **pre-existing** crash (`panic: integer overflow` in `invokeEscape`) reproducible with only `thread-sleep!` + an untimed `channel-receive` — no Phase 4 code involved. Root cause: nested `runSchedulerStep` calls can wrongly re-dispatch an *ancestor* fiber (several nesting levels up the real Zig call stack) when that ancestor's own timer pops from a deeper nesting level, corrupting its frame/handler-stack bookkeeping. Filed with full trace and a minimal repro as #1490. This phase's own tests avoid the trigger pattern (documented inline where relevant).

## Test plan

- [x] `zig build test` — full unit suite green, including new `tests_shared_channel.zig` coverage for `close()`, the four required interleaving tests from the model (reserve→concurrent-close→drain-before-eof; workers racing until-eof; failure-path rings eof-waiting receivers; failed-receive re-queue with transient capacity overshoot on a bounded channel)
- [x] `zig build test -Dgc-stress=true` — green
- [x] `bash tests/scheme/run-all.sh` — 1838/1838 pass (443 Scheme files + 1395 R7RS suite), including 3 new smoke test files and an extended `srfi18-cross-thread-channels.scm` (real OS threads: bounded-channel backpressure, close draining a worker's receive loop, timed receive as the deadlock-heuristic escape hatch)
- [x] `zig fmt --check` clean
- [x] Independent code-review pass (correctness-focused) — no findings
- [x] TLA+ model suite re-confirmed green against the KEP's pseudocode before implementation

### DIFF
--- a/src/memory.zig
+++ b/src/memory.zig
@@ -876,6 +876,23 @@ pub const GC = struct {
         return types.makePointer(@ptrCast(ch));
     }
 
+    /// KEP-0002 §6: a bounded channel, `(make-channel capacity)`. Separate
+    /// from allocChannel (rather than an optional parameter) so the
+    /// pre-Phase-4 call sites that construct an unbounded channel need no
+    /// changes.
+    pub fn allocChannelBounded(self: *GC, capacity: u32) !Value {
+        try self.maybeCollect();
+        const ch = try self.allocator.create(types.Channel);
+        ch.* = .{
+            .header = .{ .tag = .channel },
+            .head = types.NIL,
+            .tail = types.NIL,
+            .capacity = capacity,
+        };
+        self.finishAlloc(&ch.header, @sizeOf(types.Channel));
+        return types.makePointer(@ptrCast(ch));
+    }
+
     /// A promoted-channel stub: a new counted reference to an existing
     /// SharedChannel (KEP-0002 §2), allocated on this heap by deepCopy's
     /// `.channel` alias arm. `shared` isn't a Value, so there's nothing to
--- a/src/primitives_fiber.zig
+++ b/src/primitives_fiber.zig
@@ -371,6 +371,30 @@ fn channelSendShared(sc: *shared_channel.SharedChannel, payload: Value, ch_val: 
             .would_park => {},
         }
 
+        // A prior loop iteration may have parked, armed a timer, and then
+        // woken *spuriously* -- sweepSharedWaiters (fiber.zig) flips a
+        // .waiting fiber to .suspended unconditionally on any notify, with
+        // no notion of "was this really my event", and does not touch
+        // deadline_ns or cancel the reactor timer. Left armed, that timer
+        // is still live while the drive below runs with `me` .running --
+        // and wakeReadyFiber only ever sets `timed_out` for a .waiting (or
+        // io_waiting) fiber, so a fiber pop while `me` is .running is
+        // silently dropped (schedule()'s per-tick popExpiredTimers, or
+        // parkOnReactor's own poll, both call wakeReadyFiber the same way),
+        // permanently losing the timeout for the rest of this wait.
+        //
+        // Detach it from the *reactor's* heap only -- `me.deadline_ns`
+        // itself is deliberately left set, as the sole authoritative record
+        // of the original absolute deadline. The park step below always
+        // re-adds a timer if it doesn't resolve things, preferring this
+        // preserved value over a freshly re-parsed `deadline_ns` argument
+        // (`me.deadline_ns orelse deadline_ns`) -- clearing the field here
+        // instead would make every re-arm fall through to `deadline_ns`,
+        // which is recomputed as "now + relative seconds" on every fresh
+        // native call (see channelSendFn), silently extending the timeout
+        // on every spurious wake instead of honoring the original one.
+        if (me.deadline_ns != null) ctx.reactor.removeTimer(me);
+
         // Drive local siblings once. See SharedChannelPoll's doc comment
         // (channelReceiveShared) for why `me` must stay .running throughout.
         me.timed_out = false;
@@ -381,11 +405,9 @@ fn channelSendShared(sc: *shared_channel.SharedChannel, payload: Value, ch_val: 
             me.status = .waiting;
             me.waiting_on = ch_val;
             vm.gc.writeBarrier(&me.header, ch_val);
-            if (me.deadline_ns == null) {
-                if (deadline_ns) |d| {
-                    me.deadline_ns = d;
-                    try ctx.reactor.addTimer(d, me);
-                }
+            if (me.deadline_ns orelse deadline_ns) |d| {
+                me.deadline_ns = d;
+                try ctx.reactor.addTimer(d, me);
             }
             ctx.sched.enrollSharedWaiter(me) catch |err| {
                 me.status = .running;
@@ -402,11 +424,9 @@ fn channelSendShared(sc: *shared_channel.SharedChannel, payload: Value, ch_val: 
         me.status = .waiting;
         me.waiting_on = ch_val;
         vm.gc.writeBarrier(&me.header, ch_val);
-        if (me.deadline_ns == null) {
-            if (deadline_ns) |d| {
-                me.deadline_ns = d;
-                try ctx.reactor.addTimer(d, me);
-            }
+        if (me.deadline_ns orelse deadline_ns) |d| {
+            me.deadline_ns = d;
+            try ctx.reactor.addTimer(d, me);
         }
         ctx.sched.enrollSharedWaiter(me) catch |err| {
             me.status = .running;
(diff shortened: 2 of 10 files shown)


baijum (commented):
## Code Review — KEP-0002 Phase 4 (capacity, timeouts, close)

Reviewed the full diff plus the surrounding scheduler/reactor internals (`runSchedulerStep`, `parkOnReactor`, `wakeReadyFiber`) to validate the concurrency logic.

### Strengths

- **Timed local path** is clean and symmetric: `channelSendLocal` / `channelReceiveFn` set `me.status = .waiting` before the single `runSchedulerStep`, arm the timer, and remove it right after — so a fired timer is always observed by `wakeReadyFiber` (which only flips `timed_out` for `.waiting` fibers).
- **The "Also fixed" `.waiting` transition** for local timeouts is genuinely load-bearing and has direct regression coverage in `fiber-channel-timeout.scm`.
- **GC safety** is handled: `enqueueChannel`/`dequeueChannel` apply write barriers after mutating `head`/`tail`; `allocPair` auto-roots its args; `queue_len` can't underflow (dequeue guarded by `head != NIL`, lockstep with enqueue, never reached post-promotion).
- **The redispatch/relative-timeout discriminator** (`me.deadline_ns != null and me.timed_out` entry guard + `me.deadline_ns == null` re-arm guard) correctly prevents a relative timeout from silently re-extending on every `yield_retry` re-execution. This is the subtlest part and it's right.
- **Test coverage is excellent**: white-box unit tests for the reserve/close/overshoot model findings, three new smoke files, and real-OS-thread cross-thread tests. Field defaults keep pre-Phase-4 `allocChannel` callers untouched.

### Substantive concern (please verify)

**A shared-channel timeout can be silently dropped when a parked fiber loops back with its timer already armed** — `src/primitives_fiber.zig`, `channelSendShared` and `channelReceiveShared`.

Both shared paths do, each loop iteration:

```zig
me.timed_out = false;
_ = try fiber_mod.runSchedulerStep(SharedChannelPoll, .{ .sc = sc }, ...); // me stays .running
...
if (me.deadline_ns == null) { if (deadline_ns) |d| { arm timer } }        // re-arm guard
```

On the **first** iteration this is fine (timer not yet armed). But after a *spurious* wake from `SharedChannelWait` (woken, but the channel still isn't ready — the normal outcome of the wake-all discipline when one freed slot rings N parked senders/receivers), the fiber loops back with `me.deadline_ns` **already armed** and re-enters the `runSchedulerStep(...Poll...)` drive **with `me.status == .running`**.

If that drive reaches `parkOnReactor` (no runnable siblings) and blocks in `reactor.poll()`, `me`'s own timer can expire there. `reactor.poll` pops it unconditionally into `ready`, then `wakeReadyFiber(me)` hits the `else => {}` branch because `me` is `.running` (it only sets `timed_out` for `.waiting`). **The timer is consumed out of the heap and its expiry is dropped** — `me.timed_out` stays false. The fiber then parks in `SharedChannelWait` with `me.deadline_ns` still set but *no live reactor timer* (the `deadline_ns == null` guard suppresses re-arming). Net effect: the timeout is never honored — the op either completes late on the next real wakeup, or hangs indefinitely if no further cross-thread activity occurs.

Concrete interleaving: two OS threads share a bounded `(make-channel 1)`; several fibers on thread A are parked in `channel-send ch v timeout`; thread B does a single `channel-receive`, ringing all of A's send-waiters (wake-all). The loser(s) loop back with armed timers; the last loser's own poll-drive blocks in `reactor.poll()`, its timer fires while `.running`, and is swallowed.

The comment at `channelReceiveShared` ("`me` stays .running for that whole drive, and wakeReadyFiber only flips `.waiting` fibers") documents the mechanism but frames it purely as protecting the `timed_out = false` reset — it appears to miss that the same mechanism swallows a legitimately-armed timer on loop-back iterations. The local path is unaffected (it sets `.waiting` around its single step and has no poll-drive-while-running loop).

I reasoned this out against the actual `parkOnReactor`/`wakeReadyFiber`/`runSchedulerStep` bodies rather than executing it (no Zig toolchain in my review environment), so please confirm with a targeted repro. Possible fix direction: on a loop-back iteration, either remove the armed timer before the `...Poll...` drive and let the normal park re-arm it, or otherwise ensure the timer isn't live while `me` is `.running`. The existing timeout smoke tests don't hit this interleaving (single-waiter / capacity-0 shapes).

### Minor notes

- **`make-channel` rejects bignum capacities** via `types.isFixnum` with a `"non-negative exact integer"` message — a bignum *is* a non-negative exact integer, so the message is slightly misleading (harmless in practice; wording could say "fixnum" or note the range).
- **Capacity-0 semantics** (permanently full, no rendezvous handoff) differ from the CSP/Go unbuffered-channel intuition users may bring. Documented inline and in tests, but worth calling out prominently in the user-facing docs so `(make-channel 0)` isn't mistaken for a synchronous rendezvous.
- **`channel-closed?` on a foreign-owned channel raises** rather than returning a boolean. Consistent with the send/receive owner-check convention and defensible, but a predicate that raises is mildly surprising — fine as long as it's documented.
- **eof-object send rejection** is a deliberate, documented backward-compat carve-out. Low risk since channels are new, but it *is* a behavior change for any code that sent `(eof-object)` as a value.

### Verdict

High-quality, well-tested, well-documented work that gets the hardest part (the redispatch/relative-timeout discriminator) right. I'd hold on merge only to resolve the one shared-path timer-drop concern above — narrow, but a real timeout-guarantee defect if confirmed. Everything else is either solid or cosmetic.

---

baijum:
Thanks both — addressed in 467870d.

## @baijum's shared-channel timeout finding — confirmed and fixed

Traced it independently against `sweepSharedWaiters`/`wakeReadyFiber`/`parkOnReactor` and it's real: the wake-all discipline (§5) means `sweepSharedWaiters` flips a `.waiting` fiber to `.suspended` on *any* notify with no per-waiter targeting, and never touches `deadline_ns` or cancels the reactor timer. A loser that loses the retry race loops back with its timer still live, then re-enters the local-sibling poll drive with `me.status == .running` — where `wakeReadyFiber` can't observe a pop (it only sets `timed_out` for `.waiting`/`.io_waiting`). The timeout is then silently gone for the rest of that wait.

Fixed by detaching the timer from the reactor's heap (not `me.deadline_ns` itself) before every poll-drive iteration, re-arming from the preserved absolute deadline at the next park step. My first attempt at this fix cleared `me.deadline_ns` too, which introduced a *different* bug — every re-arm would then fall through to a freshly re-parsed relative timeout and silently extend it on each spurious wake. Caught that myself while trying to write a live repro (single-threaded `yield`/`fiber-join` timing made a fully deterministic multi-waiter race hard to construct cleanly, so this is verified by code trace + the full test suite rather than a new dedicated race test — flagging that gap rather than pretending otherwise).

Full suite (`zig build test`, `-Dgc-stress=true`, `bash tests/scheme/run-all.sh`, 1838 tests) green after the fix.

## @coderabbitai's two comments

- **`tests_shared_channel.zig` leak-check baselines**: valid, applied — the flagged test now captures `shared_object.liveCount()`/`notifier.refcount` baselines and asserts against them with an explicit (non-deferred) `reactor.deinit()`, matching its neighbors.
- **`library.zig` exports for `channel-close!`/`channel-closed?`/`channel-timeout-exception?`**: this is a false positive — this codebase auto-exports library procedures via the `PrimSpec.libs` `LibSet` field (see `primitives.zig`'s `Lib` enum + registration), not via per-procedure entries in `library.zig`. All three are already reachable through `(import (kaappi fibers))`, exercised directly by `tests/scheme/smoke/fiber-channel-close.scm` and `fiber-channel-timeout.scm`, both passing.

PR #1521, opened 2026-07-13
Fix dirty-snapshot dispatch hazard in mutex-lock!/condition-variable-wait

### LINKED ISSUE #1487: Pre-existing dirty-snapshot dispatch hazard: mutex-lock!/condition-variable-wait!/thread-sleep! set .waiting around inline scheduler drives
## Background

PR #1485 (KEP-0002 Phase 3, #1468) fixed a confirmed, reproducible VM-corruption bug in `channelReceiveShared` (`src/primitives_fiber.zig`): a fiber that set `.waiting` and then kept its own native call frame alive while doing a nested/inline scheduler drive of other local fibers could later be flipped back to `.suspended` by an unrelated wake (there, the cross-thread unconditional sweep) — making its **mid-native-call register snapshot** dispatchable from a *different* fiber's own nested drive loop. The bytecode would then resume past the in-flight primitive call with the destination register never written, silently corrupting the result (repro and full analysis in the [PR #1485 review](https://github.com/kaappi/kaappi/pull/1485#pullrequestreview-4681367504) — see the inline comment on `channelReceiveShared`'s doc block).

The fix (landed in PR #1485): a fiber dispatched by the scheduler now always parks via the flat `yield_retry` unwind before becoming independently dispatchable again — it never nests a drive after setting `.waiting`.

## The pre-existing, unfixed hazard class

The same shape — set `me.status = .waiting`, then call `runSchedulerStep(...)` (which internally drives other local fibers from `me`'s own still-live native frame) — is also how these SRFI-18 waits work in `src/primitives_srfi18.zig`:

- `thread-sleep!` — sets `.waiting` then `runSchedulerStep(SleepWait, ...)` (around line 538/567)
- `mutex-lock!` — sets `.waiting` then `runSchedulerStep(MutexWait, ...)` (around line 966/983)
- `condition-variable-wait!` — sets `.waiting` then `runSchedulerStep(CondVarWait, ...)` (around line 1089/1106)

Each of these wakes are narrower than the channel case: `wakeMutexWaiters`/`wakeOneCondVarWaiter`/timer expiry target a **specific** fiber rather than doing an unconditional flip-everyone sweep, and `runSchedulerStep`'s own loop guards against a fiber re-dispatching *itself* (`if (next_idx == my_idx) break;`). But that guard only protects each fiber's *own* loop — it does nothing if a **different** fiber's nested drive (or a scheduler tick reached through a different call chain) is the one that calls `schedule()` after the wake lands. So the same "mid-call snapshot becomes dispatchable while the real native frame is still paused deeper on the stack" corruption is theoretically reachable here too, just needs a narrower interleaving: fiber A parks on a mutex/condvar/sleep and starts an inline drive; something driven by A (directly or transitively) causes A's specific wake condition to fire while A's own drive is still nested/live; a `schedule()` call other than A's own top-level check (e.g. from whatever A's drive dispatched) picks A up while `.suspended`.

This hasn't been observed in practice — the review that found this noted it's "theoretically reachable... narrow single-waiter wake windows, which is why it hasn't bitten" — but it's the identical bug class to the one just fixed, sitting in code that predates this PR.

## Suggested direction (not investigated further)

From the same review: rather than patching each call site individually (which is how the channel bug was fixed, but there are three more sites here plus whatever comes later), consider a generic guard at the `runSchedulerStep` level — e.g. a `.driving` status that `schedule()` (and any wake function that flips fibers to `.suspended`) never selects into, distinguishing "this fiber has a live native frame doing a nested drive" from ordinary `.waiting`/`.suspended`. That would close the class once instead of once per call site.

## Suggested acceptance

- Confirm or rule out a concrete repro for at least one of the three sites (mutex-lock! is probably the easiest: two fibers contending for one mutex, with a nested drive in between the unlock and the redispatch).
- Either fix each site to match `channelReceiveShared`'s yield_retry-only-park discipline, or land the generic `.driving`-status guard and verify it closes all three (plus any future call site) at once.
- Regression test(s) mirroring the two-receiver repro style used for the channel fix.

Cross-referenced from #1468 (closed by PR #1485, which is where this hazard was identified as a non-blocking follow-up).

baijum:
## Summary

Closes #1487.

- `mutex-lock!`, condition-variable-wait (`mutex-unlock!` with a condvar), `thread-join!`'s fiber path, and timed local `channel-send`/`channel-receive` all set a fiber `.waiting` and then call `runSchedulerStep` recursively from the *same* native frame — the identical shape already fixed for shared channels in #1485. If something wakes that fiber (flips it `.suspended`) while its own nested drive is still live deeper on the Zig call stack, a *different* fiber's own nested `schedule()` call can dispatch it from a stale, mid-native-call register snapshot, resuming bytecode past the in-flight primitive call with the destination register never written. Confirmed via a 3-fiber nested mutex-contention repro (git-stash A/B) — without the fix, one fiber's `(mutex-lock! m)` result comes back corrupted instead of `#t`.
- Rather than rewriting each call site to use #1485's `yield_retry` discipline, this lands the issue's alternative suggestion: a generic `driving` guard on `Fiber`, set for the whole extent of any `runSchedulerStep` call and excluded from dispatch-selection regardless of status. A per-site rewrite would have silently broken cross-OS-thread mutex/condvar polling (which depends on the *specific* waiting loop's own periodic recheck) and reintroduced the exact lost-wakeup race the condvar generation-snapshot-before-unlock ordering exists to prevent. The generic guard sidesteps both — and, as a bonus, also closes the identical hazard in `thread-join!` and timed local channel-send/receive, which weren't named in the issue but turn out to have the same bug.
- First attempt at the guard baked the exclusion into `schedule()` itself, which broke `yieldFn`/`threadYieldFn`'s "is yielding worthwhile" advisory check — a busy sibling's `(yield)` calls stopped seeing a driving ancestor whose wait had just resolved, turning yield into a no-op and reproducing issue #1440's starvation symptom by a different path (caught by the existing `fiber-timed-mutex-lock-not-starved-by-busy-sibling.scm` test, which failed consistently, not flakily). Fixed by splitting `schedule()` (unchanged, used by advisory checks and `vm_calls.zig`'s non-nested `switchTo` dispatch) from a new `scheduleForDispatch()` (excludes driving fibers, used only by `runSchedulerStep`'s own dispatch loop).

## Test plan

- [x] `zig build test` (unit suite) green
- [x] `zig build test -Dgc-stress=true` green
- [x] `bash tests/scheme/run-all.sh` — 1839 pass, 0 fail (incl. R7RS suite)
- [x] `zig build wasm` still compiles
- [x] `zig build -Dtarget=x86_64-linux` cross-compiles
- [x] `zig fmt --check` clean
- [x] New regression tests (Zig `tests_scheduler.zig` + Scheme `tests/scheme/smoke/mutex-nested-dispatch-dirty-snapshot-1487.scm`) confirmed to fail without the fix and pass with it via git-stash A/B
- [x] Previously-regressed `fiber-timed-mutex-lock-not-starved-by-busy-sibling.scm` re-verified passing consistently (5/5 runs) after the schedule()/scheduleForDispatch split

### DIFF
--- a/src/fiber.zig
+++ b/src/fiber.zig
@@ -47,6 +47,28 @@ pub const Fiber = struct {
     param_overrides: std.AutoHashMap(usize, Value),
     deadline_ns: ?u64 = null,
     timed_out: bool = false,
+    /// True for the entire extent of a runSchedulerStep call this fiber
+    /// itself invoked (set/cleared there, never by callers) -- i.e. this
+    /// fiber's own native frame is still live on the Zig call stack,
+    /// mid-nested-dispatch. scheduleForDispatch() and hasRunnableFibers()
+    /// must never select/count such a fiber as an actual dispatch target,
+    /// even if some wake function flips its `status` to .suspended while
+    /// driving is true: since the whole scheduler runs on one OS thread, a
+    /// fiber with driving == true is always an ancestor of whichever call
+    /// is currently asking (it can only have gotten a nested dispatch by
+    /// dispatching *something*, and that something's own call tree is what
+    /// is presently executing) -- dispatching it from anywhere but its own
+    /// loop would resume its bytecode from a stale, mid-native-call
+    /// register snapshot with the destination register never written (see
+    /// runSchedulerStep's doc comment; #1487). Ancestors can never make
+    /// independent progress while a descendant is active regardless, so
+    /// excluding them from dispatch changes no genuine liveness outcome --
+    /// only the parked fiber's own loop ever consumes its wake. Plain
+    /// schedule() (yieldFn/threadYieldFn's advisory check; vm_calls.zig's
+    /// non-nested switchTo dispatch) deliberately does NOT exclude driving
+    /// fibers -- see scheduleForDispatch's doc comment for why excluding
+    /// them there too reproduces #1440's symptom by a different path.
+    driving: bool = false,
     terminated: bool = false,
     os_thread: ?std.Thread = null,
     /// Set together with `.io_waiting` when a blocking I/O primitive parks
@@ -304,7 +326,39 @@ pub const FiberScheduler = struct {
     /// promptness argument applies to I/O: a zero-timeout reactor poll per
     /// tick wakes ready I/O waiters even while a busy/yielding sibling
     /// keeps the loop from ever reaching parkOnReactor's blocking poll.
+    ///
+    /// Does NOT exclude `driving` fibers — see `scheduleForDispatch`'s doc
+    /// comment for why that exclusion must not live here. This is the
+    /// general "would *anything* becoming runnable matter" query: yieldFn/
+    /// threadYieldFn use it only to decide whether arming the Yielded
+    /// unwind is worthwhile (the returned index itself is discarded), and
+    /// vm_calls.zig's top-level scheduleNextAfterYield/runWithScheduler
+    /// dispatch via switchTo, which — unlike runSchedulerStep — never nests
+    /// another drive, so it carries none of the corruption risk `driving`
+    /// guards against.
     pub fn schedule(self: *FiberScheduler) ?usize {
+        return self.scheduleImpl(false);
+    }
+
+    /// Like `schedule`, but also excludes `driving` fibers — see that
+    /// field's doc comment (#1487). Used only by runSchedulerStep's own
+    /// dispatch loop, the sole call site that actually restoreFiber+
+    /// vm.runUntil()s whatever index this returns: selecting a `driving`
+    /// fiber there would resume its stale, mid-native-call register
+    /// snapshot from a different fiber's nested drive. Plain `schedule()`
+    /// must keep finding these fibers for its other callers (see its own
+    /// doc comment) — excluding them there instead reproduces #1440's
+    /// symptom by a different path: a busy sibling's `(yield)` calls
+    /// yieldFn/threadYieldFn's own `schedule() == null` advisory check,
+    /// which would then see nothing runnable (the driving ancestor whose
+    /// wait *just* resolved is invisible) and silently no-op instead of
+    /// actually yielding, starving that ancestor's own loop of the turn it
+    /// needs to notice and exit.
+    pub fn scheduleForDispatch(self: *FiberScheduler) ?usize {
+        return self.scheduleImpl(true);
+    }
+
+    fn scheduleImpl(self: *FiberScheduler, exclude_driving: bool) ?usize {
         if (self.vm.reactor) |reactor| {
             // KEP-0002 §5's normative consume protocol, checked every tick
             // (not just when idle) for the same promptness reason as the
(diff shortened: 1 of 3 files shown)


baijum:
## Review (at `080b4fbf`) — the fix is correct; I'd merge it. The one red check (macOS) is a cold-cache CI timeout, not a real failure or a regression from this change.

### The change is correct

I re-derived the core safety argument rather than taking the (excellent) doc comments at face value:

- **The exclusion is liveness-safe.** A `driving` fiber never needs to be *dispatched* to make progress — every `Ctx.isDone()` behind `runSchedulerStep` is a **state observation** the driving fiber's own loop re-polls each iteration (`MutexWait` → `!m.locked`; `CondVarWait` → `me.status != .waiting or gen != start_gen`; `TargetWait` → target completed/errored; plus `SleepWait`/channel waits). So a woken driving fiber resumes when its own loop's `while (!ctx.isDone())` flips, not via selection. Excluding it from `scheduleForDispatch()`/`hasRunnableFibers()` therefore changes no genuine liveness outcome — exactly as the comment claims.
- **The `schedule()` vs `scheduleForDispatch()` split is right.** Plain `schedule()`'s two remaining callers both need driving fibers to stay *visible*: the advisory `schedule() == null` yield checks (only null-vs-not matters), and `vm_calls.zig`'s top-level `runWithScheduler`/`scheduleNextAfterYield`, which structurally never run while any fiber is driving (each `runUntil` returns — clearing the nested `defer` — before the next `schedule()`), so they can never be *handed* a driving fiber anyway. Only `runSchedulerStep`'s own nested loop needs the exclusion.
- **Lifecycle is clean.** `driving` defaults to `false` and every fiber is born through `allocFiber` (Zig applies the field default, same as `os_thread` et al.); the single `me.driving = true` is bracketed by `defer`, so it clears on *every* exit including error unwinds; and a given fiber can only ever have one live `runSchedulerStep` frame (its own bytecode doesn't run while it's parked driving others), so a plain bool — not a counter — is sufficient.
- **Coverage is uniform.** Because the guard lives in the shared `runSchedulerStep`, it closes the hazard for all nested-drive call sites at once (mutex, condvar, fiber-/thread-join, sleep, local + shared channels, `waitForFd`), not only the four named — nice.

### The macOS check is a cold-cache timeout, not a failure

`test (macos-latest, ReleaseSafe)` is red because the job hit the 20-min cap (`##[error]The job has exceeded the maximum execution time of 20m0s`), **not** an assertion. Evidence it's environmental and unrelated to this change:

- Unit tests (incl. the new `tests_scheduler.zig`) **passed** on macOS, and `mutex-nested-dispatch-dirty-snapshot-1487.scm` **passed** at 17:53. The cancellation landed ~10 min later inside the slow LLVM **Compile tests**, far from anything this PR touches.
- **`Build` took 3m10s here vs 4s on `main`** (warm Zig cache, keyed on `matrix.optimize`); the unit-test compile step was similarly cold. That ~3–4 min cold-cache penalty is what pushed a job that normally runs ~14–16 min (already >70% of the 20-min budget) past the cap.
- A code slowdown would show on every platform — but **ubuntu ReleaseSafe passed in 14m58s** (normal), as did ReleaseFast / Debug / riscv64 / wasm, and the PR benchmark comment shows no regression.

**Action:** re-run the macOS job (cache is warmer now). If macOS keeps flirting with the cap, bumping its `timeout-minutes` above the default 20 is the real fix — separate infra concern, not this PR.

### Minor comments (none blocking)

1. **`tests/scheme/smoke/mutex-nested-dispatch-dirty-snapshot-1487.scm:19-22`** — the comment describes "a generic `driving` flag that **`schedule()`**/`hasRunnableFibers()` never select/count," but the whole point of the split is that plain `schedule()` *deliberately keeps* selecting driving fibers; only `scheduleForDispatch()` (+ `hasRunnableFibers`) exclude them. It contradicts the `fiber.zig` doc comments — suggest s/`schedule()`/`scheduleForDispatch()`/. (CodeRabbit flagged the same line.)
2. **`src/fiber.zig:827`** — `if (next_idx == my_idx) break;` is now **unreachable**: `me` is always `driving` and occupies `my_idx`, so `scheduleForDispatch()` can never return it. This is precisely the old, narrower guard that the `driving` flag now subsumes (pre-fix it only stopped *me*'s own loop from re-selecting me; the flag generalizes that to every loop). Harmless as defense-in-depth, but worth a one-line note or `assert(next_idx != my_idx)` so it doesn't read as live logic — the doc comment still references the guard.
3. **`src/tests_scheduler.zig:216-299`** — the three new tests hand-roll `memory.GC.init` + `makeTestVM`; the rest of the file and your own 4th test use `th.TestContext` (identical semantics, same LIFO deinit order). Pure consistency nit. (Also CodeRabbit.)
4. Branch is one commit behind `main` (missing #1519, docs-only) — merges cleanly, rebase optional.
5. Only `mutex-lock!` has an end-to-end Scheme repro; condvar / join / channel rely on the shared mechanism + the direct unit tests. Reasonable given how hard these interleavings are to force deterministically, but a `condition-variable-wait` repro would be a nice follow-up since the issue named it.

**Bottom line: ready to merge once the macOS job is re-run green.** The code needs no changes to be correct; comment nit #1 is the only thing I'd bother fixing before merging, and even that is optional.

baijum:
Addressed the review feedback at 6e4ee90:

- Fixed the stale comment in `mutex-nested-dispatch-dirty-snapshot-1487.scm` (CodeRabbit + manual review both flagged it) — it now correctly attributes the exclusion to `scheduleForDispatch()`/`hasRunnableFibers()`, not plain `schedule()`.
- Added a one-line note on `fiber.zig`'s now-unreachable `next_idx == my_idx` guard explaining it's subsumed by `driving` and kept as defense-in-depth.
- Converted the three new driving-guard unit tests to `th.TestContext` for consistency with the rest of the file (CodeRabbit).

Left as follow-up work rather than blocking this PR: an end-to-end Scheme repro for `condition-variable-wait` specifically (only `mutex-lock!` has one — the shared mechanism is covered by both the direct unit tests and the mutex repro, and forcing the exact interleaving deterministically for condvar/join/channel is harder), and the optional rebase onto main (docs-only #1519, no conflict).

Re-verified after these changes: `zig build test` green, `zig fmt --check` clean, and the corruption repro (`tests/scheme/smoke/mutex-nested-dispatch-dirty-snapshot-1487.scm`) still passes.

PR #1524, opened 2026-07-13
Resolve portable SRFI libraries via exe-relative lib/ fallback

### LINKED ISSUE #1523: Portable SRFI libraries unresolvable when running a from-source build outside the kaappi checkout
## Summary

A `kaappi` binary built from source (`zig build`, no installer run) cannot
resolve *portable* (non-built-in) SRFI `.sld` libraries when invoked from
any directory other than the kaappi source checkout itself, with no
`--lib-path` given. Found triaging an apparently-unrelated CI failure on
[kaappi-examples#1](https://github.com/kaappi/kaappi-examples/pull/1) (a
`(kaappi parallel)` PR, [kaappi#1522](https://github.com/kaappi/kaappi/pull/1522))
— confirmed unrelated to that change; this reproduces identically against
plain `main`.

## Root cause

`resolveLibraryPath` (`src/vm_library.zig`) only checks, in order:

1. `""` / `"lib/"` relative to the **current working directory**
2. Each entry in `vm.lib_paths` — `--lib-path` flags, the running script's
   own directory, and `~/.kaappi/lib`

None of these ever point at wherever the binary was actually **built
from** or **installed to**. The official `curl | bash` installer works
around this by separately downloading `kaappi-lib.tar.gz` and extracting
it to `~/.kaappi/lib` (`docs/install.sh` in kaappi.github.io) — but a bare
`git clone` + `zig build`, with the resulting binary then run from some
other directory, has no equivalent: `zig build` doesn't copy `lib/`
anywhere near `zig-out/bin/kaappi` (`zig-out/lib/` only ever contains
`libkaappi_rt.a`, the native-backend runtime), so there is no
built-relative fallback either.

Built-in SRFIs (1, 9, 13, 18, 39, 69, 133, 170 — registered as native Zig
primitives) are unaffected; only the 64 portable ones need a
filesystem/CWD-relative lookup to succeed.

## Repro

```bash
git clone --depth 1 https://github.com/kaappi/kaappi.git /tmp/kaappi-repro
cd /tmp/kaappi-repro && zig build

mkdir -p /tmp/somewhere-else && cd /tmp/somewhere-else
echo '(import (scheme base) (srfi 151)) (display (bitwise-ior 1 2))' > t.scm

# Fails -- simulates a fresh CI runner with no prior ~/.kaappi/lib:
HOME=/tmp/fake-home /tmp/kaappi-repro/zig-out/bin/kaappi t.scm
# t.scm:1: error: library not found: (srfi.151)
# t.scm:1: error: undefined variable 'bitwise-ior'
```

Exactly matches the CI failure — `kaappi-examples`' workflow clones kaappi
to `/tmp/kaappi`, builds it, then runs
`/tmp/kaappi/zig-out/bin/kaappi maze-solver/app.scm demo` from the
`kaappi-examples` checkout root (a fresh GitHub Actions runner, no
`~/.kaappi/lib`) — `maze-solver` is the only current pure-Scheme example
that happens to import a portable SRFI (`(srfi 151)`, for `bitwise-ior`),
which is presumably why this hasn't been noticed via any other example.
Confirmed with `HOME` overridden to an empty directory that a real (not
simulated) `~/.kaappi/lib` on the reporting machine (from unrelated prior
`thottam`/installer use) was masking the failure locally at first.

## Suggested direction (not investigated further)

Some combination of:
- Add an executable-relative search prefix (e.g. `<exe_dir>/../lib`), so a
  binary built via `zig build` finds its own source tree's `lib/` the same
  way an installed release finds `~/.kaappi/lib`.
- Have `zig build` copy/symlink `lib/` into `zig-out/lib/` (currently only
  `libkaappi_rt.a` lives there) so the exe-relative prefix above would
  work uniformly for both from-source and installed layouts.
- Document that any from-source build/CI workflow needs an explicit
  `--lib-path <repo>/lib` if it imports a portable SRFI — cheaper, but
  papers over what looks like a real ergonomics gap for "build from
  source and run the binary from anywhere," which the release/install
  path already handles.

## Suggested acceptance

- A `kaappi` binary built fresh (`zig build`, no installer, no
  `~/.kaappi/lib`) can resolve a portable SRFI (e.g. `(srfi 151)`) when
  invoked from a directory unrelated to the source checkout, with no
  `--lib-path`.
- `kaappi-examples`' "Smoke test — maze solver" CI step goes green without
  needing changes to `maze-solver/app.scm` or the workflow's `--lib-path`
  usage (currently none).
- Regression test: run a portable-SRFI-importing script from a `cwd`
  outside the source tree, `HOME` pointed at an empty directory, no
  `--lib-path`.

baijum:
## Summary

- `resolveLibraryPath` only checked cwd-relative paths, `--lib-path`, the running script's own directory, and `~/.kaappi/lib` — none of which point at a `zig build`-produced binary's own source tree, so a from-source binary run from any other directory (no prior `thottam`/installer setup) couldn't resolve portable (non-built-in) SRFI `.sld` libraries.
- Adds a shared exe-relative lib-dir lookup (`kaappi_paths.getExeRelativeLibDir`/`siblingLibDir`), mirroring the `<exe_dir>/../lib` search already used to find `libkaappi_rt.a` for the native backend.
- `build.zig` now installs `lib/` into `zig-out/lib/` alongside the exe (via `b.installDirectory`) so the two line up.
- `main.zig` adds that directory as a last-resort fallback to `vm.lib_paths`, after `~/.kaappi/lib` so an existing install is never shadowed.
- `native_compiler.zig`'s near-duplicate local copy of the exe-path lookup now delegates to the shared helper instead of duplicating the platform-specific logic.

Fixes #1523.

## Test plan

- [x] Reproduced the exact issue scenario (fresh checkout copy, `zig build`, run from an unrelated directory with `HOME` pointed at an empty dir, no `--lib-path`) — failed before the fix with the reported `library not found: (srfi.151)` error, succeeds after.
- [x] Added `tests/scheme/smoke/exe-relative-lib-1523.sh`: isolates a built `bin/kaappi` + sibling `lib/` into a temp "distribution" layout, runs from an unrelated cwd with empty `HOME`, asserts a portable SRFI import succeeds with no `--lib-path`, plus a negative control (same binary, no sibling `lib/`, must fail the same way the bug did).
- [x] Added Zig unit tests for `siblingLibDir`/`getExeRelativeLibDir` in `src/kaappi_paths.zig`, including an edge case a review pass caught (a 2-level exe path like `/opt/kaappi` was collapsing to a bogus root `/lib` instead of returning `null`).
- [x] `zig build test` — all unit tests pass.
- [x] `bash tests/scheme/run-all.sh` — 1839/1839 pass (444 Scheme files + 1395 R7RS suite).
- [x] `zig fmt --check` clean.
- [x] Re-verified `kaappi compile` (LLVM native backend, via the refactored `native_compiler.zig`) still links and runs correctly.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -115,6 +115,18 @@ pub fn build(b: *std.Build) void {
     exe.stack_size = 64 * 1024 * 1024; // 64 MB — u16 register widening increases compiler frame sizes
     b.installArtifact(exe);
 
+    // Portable library sources (.sld/.scm), installed next to the exe so a
+    // from-source build can resolve them via the exe-relative <exe_dir>/../lib
+    // search path (kaappi_paths.getExeRelativeLibDir) with no --lib-path and
+    // no ~/.kaappi/lib (#1523). Mirrors where `zig build lib` puts
+    // libkaappi_rt.a.
+    b.installDirectory(.{
+        .source_dir = b.path("lib"),
+        .install_dir = .lib,
+        .install_subdir = "",
+        .include_extensions = &.{ ".sld", ".scm" },
+    });
+
     const run_cmd = b.addRunArtifact(exe);
     run_cmd.step.dependOn(b.getInstallStep());
     if (b.args) |args| {
--- a/src/kaappi_paths.zig
+++ b/src/kaappi_paths.zig
@@ -69,12 +69,16 @@ pub fn getExeRelativeLibDir(buf: []u8) ?[]const u8 {
     var path_buf: [1024]u8 = undefined;
     const exe_path: []const u8 = blk: {
         if (comptime @import("builtin").os.tag == .linux) {
+            // /proc/self/exe is a kernel-resolved canonical path already —
+            // no realpath needed. Reject a result that fills the whole
+            // buffer: readlink doesn't NUL-terminate, and an exact-length
+            // return means the real target may have been truncated.
             const n: isize = std.posix.system.readlink(
                 "/proc/self/exe",
                 &path_buf,
                 path_buf.len,
             );
-            if (n > 0) break :blk path_buf[0..@intCast(n)];
+            if (n > 0 and @as(usize, @intCast(n)) < path_buf.len) break :blk path_buf[0..@intCast(n)];
         }
 
         if (comptime @import("builtin").os.tag == .macos) {
@@ -82,6 +86,18 @@ pub fn getExeRelativeLibDir(buf: []u8) ?[]const u8 {
             const rc = std.c._NSGetExecutablePath(&path_buf, &size);
             if (rc == 0) {
                 const len = std.mem.indexOfScalar(u8, &path_buf, 0) orelse path_buf.len;
+                // _NSGetExecutablePath may return a symlinked or relative
+                // path (e.g. a Homebrew Cellar symlink), which would derive
+                // ../lib from the wrong tree — resolve it to the real path
+                // first. Fall back to the raw path if realpath fails or the
+                // buffer wasn't NUL-terminated within bounds.
+                if (len < path_buf.len) {
+                    var resolved_buf: [std.posix.PATH_MAX]u8 = undefined;
+                    if (std.c.realpath(path_buf[0..len :0], &resolved_buf)) |resolved| {
+                        const rlen = std.mem.indexOfScalar(u8, resolved[0..resolved_buf.len], 0) orelse resolved_buf.len;
+                        break :blk resolved[0..rlen];
+                    }
+                }
                 break :blk path_buf[0..len];
             }
         }
@@ -118,8 +134,11 @@ test "siblingLibDir returns null when the result doesn't fit the buffer" {
 }
 
 test "getExeRelativeLibDir returns a lib dir sibling to the exe's bin dir" {
+    // The test binary itself is always deeply nested (zig-cache output
+    // dirs on every supported platform), so this must not return null —
+    // asserting that keeps a regression in the platform probe from
+    // silently passing.
     var buf: [1024]u8 = undefined;
-    if (getExeRelativeLibDir(&buf)) |dir| {
-        try std.testing.expect(std.mem.endsWith(u8, dir, "/lib"));
-    }
+    const dir = getExeRelativeLibDir(&buf) orelse return error.TestUnexpectedResult;
+    try std.testing.expect(std.mem.endsWith(u8, dir, "/lib"));
 }
--- a/src/kaappi_paths.zig
+++ b/src/kaappi_paths.zig
@@ -29,3 +29,97 @@ test "getHome falls back to HOME/.kaappi" {
             std.c.getenv("KAAPPI_HOME") != null);
     }
 }
+
+/// Computes `<parent-of-bin-dir>/lib` for an executable's own absolute
+/// path, written into `buf`. Split out from `getExeRelativeLibDir` so the
+/// path arithmetic is unit-testable with synthetic paths, independent of
+/// wherever the calling process actually happens to live on disk.
+///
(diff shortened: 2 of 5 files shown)


baijum (commented):
## Review — exe-relative `lib/` fallback for portable SRFIs (#1523)

Checked out the PR head (`00084a12`) and verified it end-to-end on macOS arm64.

### Verification — all green
- `zig build` succeeds; `b.installDirectory` populates `zig-out/lib/` (`zig-out/lib/srfi/151.sld` present).
- Manual repro from #1523 (isolated `dist/bin/kaappi` + sibling `lib/`, `cd` to an unrelated dir, empty `HOME`, no `--lib-path`) now prints `3` and exits 0 — failed on `main`.
- `tests/scheme/smoke/exe-relative-lib-1523.sh` → 3/3 pass, including the negative control (no sibling `lib/` → exit 1), which proves the positive case genuinely exercises the exe-relative path rather than some other lookup.
- Full `zig build test` unit suite → exit 0.
- `zig fmt --check` clean on all changed files.

### Design — looks right
- Centralizing the exe-path arithmetic in `kaappi_paths.getExeRelativeLibDir`/`siblingLibDir` and having `native_compiler.zig` delegate is a net **de**-duplication. The relocated copy previously lacked the `parent_slash == 0` guard and only avoided a bogus root `/lib` because `checkLibDir` happened to reject it; the new guard fixes that properly and is unit-tested.
- Ordering is safe: the fallback is appended **after** `~/.kaappi/lib`, and `resolveLibraryPath` still checks cwd-relative `""`/`lib/` first — so an existing install is never shadowed and in-checkout runs are unchanged. Because it only ever *appends*, it can add resolutions that previously failed but never change an existing successful one → no regression risk for the current suites (matches the clean run).
- `build.zig`'s `.include_extensions = &.{ ".sld", ".scm" }` covers 100% of `lib/` (70 `.sld` + 1 `.scm`, no other file types), so nothing is silently left out of the install.

### Minor / optional (none blocking)
1. **`src/kaappi_paths.zig:120`** — the `getExeRelativeLibDir` test passes *vacuously* when the function returns `null` (`if (...) |dir|` with no `else`). In the test runner the exe is deeply nested so it's always non-null; asserting non-null first would make the test fail on a real regression instead of skipping. (Same as CodeRabbit's inline note.)
2. **`src/kaappi_paths.zig:80`** — on macOS `_NSGetExecutablePath` returns the path *as invoked* (symlinks / `.` / `..` not resolved), unlike Linux's canonical `/proc/self/exe`. A from-source binary reached via a `bin/` symlink on macOS would derive the wrong `../lib`. Narrow in practice (real installs hit `~/.kaappi/lib` first; Homebrew-style prefixes have a real file), and it's pre-existing behavior inherited from the old `native_compiler` copy — a `realpath` follow-up at most, not a blocker. (Also flagged by CodeRabbit.)
3. **`src/main.zig:341` & `:382`** — duplicate function-local `@import("kaappi_paths.zig")`; harmless (imports are cached), could hoist to one. Cosmetic. (CodeRabbit.)
4. Nit (my own): on Linux the 1024-byte `readlink` into `path_buf` truncates silently for exe paths > 1024 bytes (no error, no NUL), whereas the macOS branch returns `null` on overflow. Extreme edge case; if this is ever touched again, bailing on `n == path_buf.len` would make the two branches symmetric.

### Verdict
Correct, well-scoped, and well-tested — the regression test is genuinely load-bearing (negative control + `.sbc`-cache deletion are exactly the right details). No changes required to merge. `mergeStateStatus` is `BLOCKED` only because the `test (...)` matrix jobs are still **pending** (not failing) — **ready to merge once CI goes green**. The four notes above are optional follow-ups.

Issue #1493, opened 2026-07-13
LLVM backend: LTO across the runtime boundary to inline primitives


baijum:
Part of #1491. Follow-up to the `-O2` change.

## Problem

Inlined primitives emit calls to `@kaappi_fixnum_add`, `@kaappi_car`,
`@kaappi_cons`, `@kaappi_is_null`, etc. (`src/native_decls.zig`), but those
functions live in the static archive `libkaappi_rt.a`. Plain `-O2` cannot inline
across that boundary, so even trivial arithmetic in a hot loop pays a real call
per operation.

## Proposed change

- Build `libkaappi_rt.a` with LTO (`-flto`) and link native binaries with `-flto`
  so LLVM can inline the small runtime primitives into emitted `@lambda_N` bodies.
- Alternatively/additionally, emit the hottest primitives (`+ - * < = car cdr cons null?`)
  as inline IR rather than calls, so no cross-module inlining is required.

## Acceptance criteria

- LTO build path documented and wired into `zig build native` / `kaappi compile`.
- Hot arithmetic/list loops show reduced call overhead vs `-O2`-only.
- `tests/e2e/run-e2e.sh` green.

## Risk

LTO increases link time and can surface latent IR issues; gate behind the
IR-verify + e2e safety net.

PR #1583, opened 2026-07-16
LLVM backend: split llvm_emit.zig into inline + let sub-modules

baijum:
## Summary

`src/llvm_emit.zig` had grown to **1841 lines**, past the repo's 1500-line file-size policy. This splits two self-contained subsystems out along natural seams, following the existing `llvm_emit_<theme>.zig` convention (`forms`, `lambda`, `tailcall`). `llvm_emit.zig` is now **1378 lines**.

| New file | Lines | Contents |
|----------|------:|----------|
| `src/llvm_emit_inline.zig` | 268 | Inline fixnum fast-path emission (#1493) for `+ - * < = null?`, plus its private `nanbox`/`ArithOp`/`CompareKind` helper types and `nodeMayAllocate` |
| `src/llvm_emit_let.zig` | 254 | Native `let` / `let*` emission (`emitLet` and its fallback/abandon paths) plus the `nameInList` helper |

## Why it's a safe move

- The moved helper types (`nanbox`, `ArithOp`, `nodeMayAllocate`, `nameInList`) were used **only** by their own cluster.
- Every remaining dependency of a moved function is either a stable `LLVMEmitter` method or an already-`pub` `lambda.*` helper.
- Uses the **direct-call dispatch pattern** (like `llvm_emit_forms.zig`), so `llvm_emit.zig` only gains two imports and four call-site rewrites — **no delegation wrappers**. Net diff on `llvm_emit.zig` is `+9 / −472`.

This is a pure code move — no behavior change.

## Verification

- `zig build` clean; `zig build test` full unit suite green (includes `tests_native.zig` / `tests_ir.zig`)
- `zig fmt --check` clean on all three files
- End-to-end native compile + run of a program exercising every moved path (`let`, `let*`, named-let fallback, inline `+ - * < =`, `null?`) produces output **identical** to the interpreter: `(fib10= 55 sum100= 4950 c= 12 d= 13 null?= #t #f)`
- Updated the three file attributions in `docs/dev/llvm-backend.md`

### DIFF
--- a/docs/dev/llvm-backend.md
+++ b/docs/dev/llvm-backend.md
@@ -127,7 +127,8 @@ for end users; a hard verifier **error** still fails the compile regardless.
 primitives (`kaappi_fixnum_add`, `kaappi_car`, …): they live in the static
 archive `libkaappi_rt.a`, on the far side of a compilation boundary `-O2` alone
 cannot cross. So the hottest primitives are emitted **as inline IR** instead of
-calls (#1493), removing the per-operation call in the common case:
+calls (#1493, in `src/llvm_emit_inline.zig`), removing the per-operation call in
+the common case:
 
 | Primitive | Inline fast path | Slow-path fallback |
 |-----------|------------------|--------------------|
@@ -137,7 +138,7 @@ calls (#1493), removing the per-operation call in the common case:
 
 The fast paths touch only the **NaN-boxed Value bits** (fixnum tag, payload,
 nil/boolean immediates), whose encoding is stable and is pulled from `types.zig`
-at emitter comptime (`nanbox` in `llvm_emit.zig`) — no magic numbers are
+at emitter comptime (`nanbox` in `llvm_emit_inline.zig`) — no magic numbers are
 hand-transcribed into the IR. `car`, `cdr`, and `cons` deliberately stay as
 direct specialized calls: `Pair`/`Object` are **auto-layout** (non-`extern`)
 structs whose field offsets Zig does not guarantee, so they cannot be encoded in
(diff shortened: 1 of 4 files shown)


baijum:
Thanks @coderabbitai — I verified both findings against the emitted IR and by running native binaries. **Both are real**, but they're **pre-existing on `main`** (`ff49cc6`): this PR is a verbatim code move (only `self.X(…)` → `X(self, …)` for the direct-call dispatch pattern), so it introduces neither. Confirmations:

- **Root leak on let-body tail calls** — the `ret` fires before the binding roots are popped (the `pop_roots` lands in the dead orphan block); native binaries panic with `GC root stack overflow` while the interpreter is fine. Reachable by the everyday `(let ((a …)) (g a))` shape and by self-tail loops through a let. Note the suggested one-liner (`body_tail = is_tail and bindings == NIL`) is **not** a safe fix — it doesn't cover the self-tail-loop path and converts tail-recursive let loops into non-tail recursion (R7RS proper-tail-call violation). Tracked with a full reproducer + root cause in #1585.
- **Duplicated side effects on fallback** — `abandonLetForFallback` can't unwind IR already written to `self.buf`, so an already-emitted side-effecting init runs once natively and again in the whole-form `kaappi_eval_cached`. A 33-binding let gives native `n=2` vs interpreter `n=1`. Tracked in #1586.

Keeping this PR as the pure refactor (tests green, benchmarks flat) and fixing the two bugs separately in a focused PR with native-compile-and-run regression tests, so a subtle GC/TCO change doesn't ride in on a mechanical split.

Issue #1600, opened 2026-07-16
(make-channel 0) builds a channel that can never transfer a value — reject at construction, or give it rendezvous semantics?


baijum:
## Summary

`(make-channel 0)` is accepted at construction but produces a channel that can never transfer a value: send waits for a free slot that cannot exist, receive waits for a queued value that can never be enqueued. Every untimed send/receive pairing ends in a KP3000 deadlock error. The behavior is fail-loud (the deadlock detector catches it in every configuration I tried), but the object itself is a trap — it type-checks as a channel, constructs silently, and fails at first use, potentially far from the construction site. Anyone arriving from Go will read `(make-channel 0)` as an unbuffered rendezvous channel and get the opposite: a channel over which communication is impossible.

This issue asks for a decision between two fixes: **reject capacity 0 at construction**, or **give it Go-style rendezvous semantics**.

## Reproduction

Observed on the v0.15.0 release binary; reproduced identically on `main` @ f198dde1 (macOS ARM).

```scheme
(import (kaappi fibers))
(define ch (make-channel 0))
(spawn (lambda () (channel-send ch 'x)))
(channel-receive ch)
```

```
error[KP3000]: channel-receive: deadlock — channel is empty and all fibers are blocked
```

All the variants fail the same way:

| Scenario | Result |
|---|---|
| fiber sender + main receiver (above) | `KP3000` deadlock at `channel-receive` |
| OS-thread sender (promoted `SharedChannel`) + main receiver | `KP3000` deadlock at `channel-receive` |
| bare top-level `channel-send`, no fibers | `KP3000: channel-send: deadlock — channel is full and no fibers are running` |
| timed `channel-send` / `channel-receive` | times out / returns `timeout-val` (the escape hatch works) |

## Why (code)

- `makeChannelFn` (`src/primitives_fiber.zig:111`) rejects only negatives and non-fixnums; `0` flows straight into `allocChannelBounded(0)`.
- Local send admission is `ch.queue_len < cap` (`src/primitives_fiber.zig:210`) — never true for `cap == 0` — and the park predicate `ChannelSendWait.isDone` (`src/primitives_fiber.zig:190`) only resolves on close or a freed slot, so an untimed sender parks until the deadlock detector fires.
- Shared-path admission is `queue_len + reserved < cap` (`src/shared_channel.zig:319` and KEP-0002 §4 step 3) — likewise never true.
[...]

PR #1604, opened 2026-07-16
Rendezvous semantics for (make-channel 0) on both representations (#1602, #1603)

### LINKED ISSUE #1602: Rendezvous (make-channel 0): local representation (#1600 part 2)
Part 2 of 3 for #1600 (Option B). Implements rendezvous semantics for `(make-channel 0)` on the **local (unpromoted) representation**, per the KEP-0002 amendment (#1601).

## Scope

- `types.Channel` gains `rv_demand: u32` (rendezvous demand — receivers committed to this channel). No new Value fields, so GC marking is unchanged.
- `fiber.Fiber` gains `rv_demand_on: Value = VOID` — which channel currently holds this fiber's demand token; makes the increment idempotent across `yield_retry` re-execution of the receive primitive. Must be traced by `markFiberState`/`referencesYoung` (the `waiting_on` precedent) and cleared on fiber death/termination with a demand decrement (the `abandonFiberMutexes` precedent).
- `channelSendLocal` (`src/primitives_fiber.zig`): admission for `capacity == 0` becomes `queue_len < rv_demand`; `ChannelSendWait.isDone` likewise.
- `channelReceiveFn` local path: on the park path for a rendezvous channel, acquire the demand token (increment + set fiber field, skip if already held) and `wakeChannelWaiters` (new demand is a send-side event); release the token on every terminal exit (dequeue, eof, timeout, deadlock raise) but **not** across `blockOrDeadlock`'s `Yielded` unwind.
- Timeout ordering: after the park, check the queue **before** `timed_out` ("delivery wins") so a sender that committed a handoff is never left believing a discarded value was delivered.
- Deadlock messages for the rendezvous case should say what's actually missing (no receiver / no sender), keeping KP3000.

## Tests

- Update the three fixtures that cement the degenerate behavior: `tests/scheme/smoke/fiber-channel-capacity.scm:25-31` + `:51-57`, `fiber-channel-timeout.scm:34-45`, `fiber-channel-close.scm:55-62` (always-full fixtures become a pre-filled `(make-channel 1)`; the capacity-0 test flips to asserting rendezvous).
- New smoke tests: fiber↔fiber rendezvous both directions (the #1600 repro must now yield `x`), send-before-receive and receive-before-send, timeout on each side, close waking both sides, unpaired send/receive still KP3000, multiple senders one receiver, delivery-wins-over-timeout.
- Zig unit tests in `tests_fibers.zig` for the token bookkeeping (no demand leak after timeout/deadlock; `-Dgc-stress=true` clean).

Cross-thread (promoted) rendezvous is part 3 — until it lands, a promoted capacity-0 channel keeps the old degenerate shared-path behavior (deadlock/timeout), which is fail-loud, not silent.

### LINKED ISSUE #1603: Rendezvous (make-channel 0): shared cross-thread representation (#1600 part 3)
Part 3 of 3 for #1600 (Option B). Extends rendezvous `(make-channel 0)` to the **shared (promoted, cross-thread) representation**, per the KEP-0002 amendment (#1601) and building on the local implementation (#1602).

## Scope

- `SharedChannel` (`src/shared_channel.zig`) gains `rv_demand: u32` under the existing spinlock. `send()` admission for `capacity == 0` becomes `queue_len + reserved < rv_demand`; `peekSendReady` likewise.
- `receive()`'s `would_park` branch on a rendezvous channel must also snapshot+ring `send_waiters` when demand grows — new demand is a send-side event (today only a freed slot/close rings senders).
- `channelReceiveShared` (`src/primitives_fiber.zig`): acquire/release the per-fiber demand token around the park decision (idempotent via `rv_demand_on`, exactly as the local path); entry timeout-redispatch must attempt a receive **before** honoring `timed_out` ("delivery wins").
- **Reservation-drain rule**: an abandoning (timed-out) receiver that observes `reserved > 0` with an empty queue re-parks until the in-flight send pushes or aborts (its registered notifier is rung by both paths), then re-decides — an admitted send must never be silently stranded.
- `promoteChannel` copies `ch.rv_demand` into the SharedChannel (alongside capacity/closed); migrated parked fibers keep their tokens (their retries decrement against the shared counter once `ch.shared` is set — verify the ordering).
- Token release on thread/fiber termination for shared channels (lock + decrement from the owning thread's cleanup path).

## Tests

- `tests_shared_channel.zig`: thread↔thread rendezvous both directions (the #1600 cross-thread repro delivers `x`), send-timeout with no remote receiver, receive-timeout with no remote sender, delivery-wins race (sender commits while receiver's timer fires), close wakes both sides cross-thread, promotion of a capacity-0 channel with a parked local receiver (token migration), no demand leak after timed-out cross-thread waits.
- PCT stress (`stress_channel.zig` hooks) run over the new admission path.
- gc-stress clean.

baijum:
Implements rendezvous semantics for `(make-channel 0)` on **both representations**, per the KEP-0002 §4/§6 amendment (kaappi/keps#28, model-checked first). Closes #1602, closes #1603; tracking issue #1600. Before this change, capacity 0 was the documented degenerate "permanently full" channel: construction succeeded but no value could ever transfer — the #1600 repro now prints `x`.

## Semantics

`(make-channel 0)` is a synchronous channel: `channel-send` completes only against a committed receiver, `channel-receive` only when a sender provides a value; whichever side arrives first waits. Unpaired untimed operations still raise KP3000; timeouts stay the escape hatch; close wakes both sides; eof-send stays rejected.

## Design — demand-bounded admission (no pairing machinery)

Per the amendment, rendezvous is the dynamic-capacity generalization of the existing reservation/wake-all-retry protocol: the send-admission bound for capacity 0 is `rv_demand` — the count of receivers currently committed — instead of a static capacity. The value still transfers through the queue.

- **Demand tokens**: a receiver commits at its park decision (increment + `Fiber.rv_demand_on`, idempotent across `yield_retry` re-execution) and releases on every terminal exit (value, eof, timeout, error, deadlock raise) plus the death paths (`retireSlot`, `thread-terminate!` — the `abandonFiberMutexes` precedent). New demand wakes parked senders exactly like a freed slot (model finding 4: without that ring, senders parked against bound 0 before any receiver commits hang forever).
- **Local path** (`channelSendLocal`/`channelReceiveFn`): bound-aware admission + token protocol. One non-obvious rule found by testing, now regression-tested: **dispatched fibers must flat-park (`yield_retry`) on rendezvous waits, never in-call park.** Rendezvous is the only capacity where a parked sender and parked receiver can exist simultaneously, and an in-call park makes the fiber a frozen *ancestor* when its drive transitively dispatches the counterparty (#1487: driving ancestors are undispatchable) — a main-fiber receiver stacked on top raised a spurious KP3000 with two viable timed senders frozen beneath it. Flat parks keep every rendezvous waiter dispatchable; timers survive re-parks via the preserved-`me.deadline_ns` discriminator (`channelSendShared`'s established pattern).
- **Shared path** (`SharedChannel`): `rv_demand` under the existing spinlock; `commitRvDemand` (lock + snapshot-and-ring `send_waiters`) / `withdrawRvDemand`; `promoteChannel` seeds the counter from the local one, so tokens acquired at pre-promotion local parks stay counted across migration (the token field names the same stub on both representations — no re-dispatch bookkeeping needed).
- **Timeout rules** (§6): *delivery-wins* — both operations' timeout redispatch decides through one actual `send()`/`receive()`, so a handoff/slot that materialized with the timer pop is honored, never discarded; *reservation-drain* — a timed-out rendezvous receiver observing `reserved > 0` with an empty queue re-parks until the in-flight send pushes or aborts, so a send past its point of no return is never silently stranded. The abort path now rings `recv_waiters` on rendezvous channels even while open (mirrored in the TLA+ model, kaappi/keps@fe166f2).
- **GC**: `Fiber.rv_demand_on` traced by `markFiberState` and `referencesYoung` (the `waiting_on` precedent — it must trace independently because the token survives a wake's `clear_waiting_on`).

## Verification

- `zig build test` green (full suite), including new tests: token accounting stays balanced across timeout/deadlock/handoff exits; the frozen-ancestor regression; raw shared-protocol admission against demand; promotion seeding; cross-thread handoff both directions (leak-checked).
- `-Dgc-stress=true` green for `tests_fibers` + `tests_shared_channel`.
- New Scheme suites: `tests/scheme/smoke/fiber-channel-rendezvous.scm` (17 assertions: both directions, ordering witness, deadlocks, timeouts, delivery-priority, close, multi-party, ping-pong, no-phantom-demand) and `fiber-channel-rendezvous-thread.scm` (6 assertions on the promoted path across real OS threads, including sequential lockstep handoffs and cross-thread close).
- The three existing capacity-0 fixtures pass unchanged (their assertions — unpaired timed sends time out, close wakes parked senders, untimed unpaired sends deadlock — are all still true under rendezvous); only their stale "permanently full / no rendezvous handoff" comments were updated.
- Full `tests/scheme/run-all.sh` green.

Docs follow-up: kaappi/kaappi.github.io#16 (merges when this ships in a release).

### DIFF
--- a/src/bench_channel.zig
+++ b/src/bench_channel.zig
@@ -437,7 +437,7 @@ fn benchRealPromoted(workload: Workload, iters: u64) !f64 {
     var i: u64 = 0;
     while (i < iters) : (i += 1) {
         _ = try shared_channel.send(sc, payload, null);
-        _ = try shared_channel.receive(sc, &dest_gc, null);
+        _ = try shared_channel.receive(sc, &dest_gc, null, false);
         freeArena(&dest_gc); // bound the receiver heap, matching the matrix harness
     }
     const elapsed = nowNs() - start;
--- a/src/fiber.zig
+++ b/src/fiber.zig
@@ -93,6 +93,18 @@ pub const Fiber = struct {
     /// accumulate duplicate ring entries; cleared when scheduleImpl consumes
     /// or discards the entry (#1477).
     queued: bool = false,
+    /// Rendezvous demand token (KEP-0002 §6 as amended, kaappi#1601/#1602):
+    /// the capacity-0 channel this fiber's parked receive has committed to,
+    /// or VOID. Set (with a write barrier) when the receive's park decision
+    /// increments the channel's `rv_demand`; makes that increment idempotent
+    /// across the yield_retry re-execution of the whole primitive. Released
+    /// — counter decrement + reset to VOID — on every terminal exit of the
+    /// wait (value, eof, timeout, error) and on fiber death
+    /// (releaseFiberRendezvousToken, the abandonFiberMutexes precedent).
+    /// Traced by markFiberState/referencesYoung like `waiting_on`; unlike
+    /// `waiting_on` it survives a wake (clear_waiting_on) until the retry
+    /// exits, which is exactly why it needs independent tracing.
+    rv_demand_on: Value = types.VOID,
     /// Mutexes this fiber currently owns (as `Value`s), maintained by
     /// mutex-lock! (#1458). `abandonFiberMutexes` walks this on fiber death
     /// instead of scanning a GC heap for owned mutexes — the mutex object may
@@ -229,6 +241,11 @@ pub const FiberScheduler = struct {
     /// once — a terminal fiber is never revived, so it can't be pushed twice)
     /// can't drift across call sites.
     pub fn retireSlot(self: *FiberScheduler, fiber: *Fiber, status: FiberStatus) void {
+        // Normally a no-op: the channel primitives release the rendezvous
+        // demand token on their own exits. Catches a fiber that died
+        // between a wake and its retry (e.g. dispatch-loop error paths), so
+        // its committed demand can't admit sends nobody will ever collect.
+        releaseFiberRendezvousToken(fiber);
         fiber.status = status;
         // Slot 0 is the main fiber; it can be transiently marked .completed at
         // top level (vm_calls.zig) but must never be recycled — ensureScheduler
@@ -964,6 +981,30 @@ pub fn abandonFiberMutexes(fiber: *Fiber, sched: ?*FiberScheduler) void {
     fiber.owned_mutexes.clearRetainingCapacity();
 }
 
+/// Release `fiber`'s rendezvous demand token, if it holds one (KEP-0002 §6
+/// as amended: every terminal exit of a rendezvous wait releases its token,
+/// and fiber death is the terminal exit of last resort — the
+/// abandonFiberMutexes precedent). Normally a no-op: the channel primitives
+/// release on their own value/eof/timeout/error exits; this catches a fiber
+/// killed between a wake and its retry (thread-terminate!, an errored
+/// dispatch). The channel stub is always in this thread's heap (the
+/// receive path's foreign-owner check gates acquisition), so the local
+/// decrement never touches another thread's memory; a promoted channel's
+/// counter lives in the SharedChannel and is withdrawn under its lock
+/// (#1603), safe from any thread.
+pub fn releaseFiberRendezvousToken(fiber: *Fiber) void {
+    if (fiber.rv_demand_on == types.VOID) return;
+    const ch = types.toObject(fiber.rv_demand_on).as(types.Channel);
+    if (ch.shared) |raw| {
+        const shared_channel = @import("shared_channel.zig");
+        const sc: *shared_channel.SharedChannel = @ptrCast(@alignCast(raw));
+        shared_channel.withdrawRvDemand(sc);
+    } else {
+        ch.rv_demand -= 1;
+    }
+    fiber.rv_demand_on = types.VOID;
+}
+
 /// Wait for `target` fiber to finish. Shared by fiber-join and
 /// thread-join!'s fiber path — the exact same wait condition.
 pub const TargetWait = struct {
@@ -1236,6 +1277,7 @@ pub fn markFiberState(gc: *memory.GC, fiber: *Fiber) void {
     gc.markValue(fiber.thunk);
     gc.markValue(fiber.result);
     gc.markValue(fiber.waiting_on);
+    gc.markValue(fiber.rv_demand_on);
     gc.markValue(fiber.name);
     gc.markValue(fiber.specific);
 
--- a/src/gc_collect.zig
+++ b/src/gc_collect.zig
@@ -190,7 +190,8 @@ fn referencesYoung(gc: *GC, obj: *Object) bool {
             const fiber = obj.as(fiber_mod.Fiber);
             if (isYoungPointer(gc, fiber.thunk) or isYoungPointer(gc, fiber.result) or
                 isYoungPointer(gc, fiber.waiting_on) or isYoungPointer(gc, fiber.name) or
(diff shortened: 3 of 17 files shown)


baijum:
Review triage — all nine comments assessed, seven taken (0e5d760f + kaappi/keps#29), two declined with reasons.

**Both Criticals were genuine races, and they share one root cause**: demand-counter transitions living in separate critical sections from the queue/reservation state that justifies them. Per the KEP's amendment-first rule, both were specified and model-checked before the code fix — kaappi/keps#29 adds them as **Findings 5 and 6** with rejected-variant regression witnesses (`rv2_popwindow`, `rva_naive`, both failing on `NoAbandonedTask` exactly as predicted; the original rv scenario couldn't see either because its matched budgets let a second receiver absorb any strandable handoff).

- **shared_channel.zig:330 (pop-window, Critical)** → fixed: `receive()` gains `holds_token` and withdraws the demand inside the pop's mutex section; callers clear the fiber field without a second decrement, including on the copy-failure raise (token stays withdrawn; the re-queued envelope falls under §6's abnormal-exit rule). Protocol regression test added.
- **shared_channel.zig:617 (TOCTOU, Critical)** → fixed: `tryTimeoutWithdraw` replaces `reservedCount` — queue check, reservation check, and decrement under one lock, so sender admission and receiver withdrawal serialize. Three-outcome regression test added.
- **primitives_srfi18.zig:655 (child unwind)** → fixed beside both `abandonFiberMutexes` call sites in `threadEntryFn`. Agreed this was the most reachable of the leaks — `thread-terminate!` is a normal API and the parent-side path deliberately never touches the child's fiber.
- **primitives_fiber.zig:500 (send post-park delivery-wins)** → fixed on both representations: the post-park timeout now decides through one real `send()`/admission check, mirroring the entry redispatch.
- **primitives_fiber.zig:787 (fallible park rollback)** → fixed where it matters: every receive-side `addTimer`/`enrollSharedWaiter`/`runSchedulerStep` error path releases the token and detaches the timer. Two deflations worth noting: the flagged *send*-path ranges hold no tokens, and dispatched fibers were already backstopped by `retireSlot` — the real leaks were main-fiber, OOM-only.
- **two-sender tests under-assert** → taken: both now join both senders, pin exactly one delivery + one timeout, and probe channel emptiness (the precise invariant Findings 5/6 threaten).
- **thread-sleep! sync in the close test** → taken, in Zig where it can be deterministic: the new test polls `sc.rv_demand` under the lock — the child's commitment is exactly the observable the counter provides — before closing. The Scheme twin stays as-is (harmless extra coverage; Scheme has no cross-thread parked-state observable).
- **file split at 1682 lines** → taken: rendezvous coverage extracted to `tests_shared_channel_rendezvous.zig` (the #1595 `tests_native.zig` split is the precedent). Fair to note the file was already at 1573 before this PR.
- **use `th.expectEval`/`TestContext` in tests_fibers.zig** → declined: every resident test in that file uses the `GC.init` + `makeTestVM` pattern; converting only the new tests would make the file internally inconsistent to satisfy a guideline its existing tests don't follow. A file-wide conversion is a reasonable separate cleanup.

Verification on the updated branch: full unit suite, gc-stress (`tests_fibers` + both shared-channel modules), full Scheme suite **1871 pass / 0 fail**, TLA+ suite **16/16 at expected outcomes** with all six `Cap > 0` pass-configs bit-identical.

PR #1659, opened 2026-07-18
Refuse native compilation on unsupported arches (#1656)

### LINKED ISSUE #1656: kaappi compile on unsupported arches links a segfaulting binary instead of refusing
Found while resolving a docs contradiction in #1654's PR: README claimed Linux riscv64 has the LLVM backend, while `docs/dev/porting.md` says riscv64 ships interpreter-only. The empirical answer is worse than "untested":

On a riscv64 box (Alpine container under QEMU, riscv64 Zig 0.16 as the C compiler, cross-built `libkaappi_rt.a`):

```
$ kaappi compile hello.scm -o hello    # hello.scm: self-tail-call loop + closure
Wrote /tmp/kaappi_native_45.ll
Compiled /tmp/hello                    # link succeeds
$ ./hello
Segmentation fault (core dumped)
```

`emitPreamble` (src/llvm_emit.zig) emits `target triple = "unknown-unknown-unknown"` for every arch except aarch64/x86_64, and the `-w` on the `zig cc` link lets the driver silently override the mismatched triple with the host default — so the link *succeeds* and the user gets a crashing binary instead of an error.

Two possible fixes, not mutually exclusive:

1. **Refuse loudly** (small, matches house doctrine — thottam refuses `build:` manifests on Windows the same way): `kaappi compile` on an arch without a real triple in `emitPreamble` should exit with a clear "native compilation is not supported on <arch>; the interpreter tier is" error instead of emitting unknown-triple IR. `kaappi doctor`'s native-backend check should say the same.
2. **Actually port the backend to riscv64** (larger): add the triple, decide `fast_tailcalls_supported` (comment says RISC-V musttail needs verification), debug the segfault, run `tests/e2e/run-e2e.sh` on the target per porting.md's native-backend checklist, and add CI. Only worth it with a concrete user need.

The README row was corrected to "interpreter only" in the #1654 PR; this issue tracks making the *behavior* match too (option 1 at minimum).

baijum:
## Summary

Fixes #1656. `kaappi compile` on any architecture but aarch64/x86_64 emitted an `unknown-unknown-unknown` LLVM triple; the `-w` on the `zig cc` link then let the driver silently override it with the host default, so the link **succeeded** and the user got a **segfaulting binary** instead of an error — strictly worse than a plain failure, since nothing hinted the target was never supported.

This is the "refuse loudly" prerequisite already blessed by the [native-backend scope decision](docs/dev/decisions/native-backend-architecture-scope.md) (#1658): the backend stays aarch64/x86_64-only; interpreter-tier arches (riscv64, s390x, ppc64le) ship without `kaappi compile` deliberately.

## What changed

- **`llvm_emit.zig`** — extract the per-`(arch, os)` triple switch out of `emitPreamble` into a single-source-of-truth `pub fn targetTriple(arch, os) ?[]const u8` (null on an unsupported arch), and derive `pub const native_backend_supported` from it. A future port that adds an arch arm flips both the emitted triple and the support flag at once.
- **`native_compiler.zig`** — `emitLlvmFile` (the chokepoint for **both** `kaappi compile` and `--emit-llvm`) now refuses before any file I/O when `!native_backend_supported`: it prints a message naming the host arch and pointing at the interpreter, and returns `error.NativeBackendUnsupported` (→ exit 1).
- **`doctor.zig`** — `collectNativeBackend` reports one honest `arch` **WARN** on an unsupported arch instead of the misleading c-compiler/archive/smoke-link **PASS** trio that implied native builds work. WARN keeps exit 0 (the install is healthy for the interpreter tier).
- **docs** — `doctor.md` native-backend row; decision doc marked shipped.

## Tests

New unit tests (run on every arch, including the riscv64/s390x/ppc64le QEMU CI legs):
- `targetTriple`: aarch64/x86_64 classified native-compilable, riscv64/s390x/ppc64le not; `native_backend_supported` == "host has a triple".
- `nativeUnsupportedMessage`: names the arch + both supported arches + the interpreter fallback; safe tiny-buffer fallback.

Full suite: **1215/1215 pass**; `zig fmt --check` clean.

## End-to-end verification (riscv64 under QEMU via podman)

| Command | Before | After |
|---|---|---|
| `kaappi hello.scm` | `500510` (ok) | `500510`, exit 0 — unchanged |
| `kaappi compile hello.scm -o hello` | `Compiled` then **segfault** | **exit 1**, clear message, no binary written |
| `kaappi doctor` | c-compiler/smoke-link **PASS** (misleading) | `WARN  arch: native compilation unavailable on riscv64 …` |

Supported host (aarch64) native compile re-verified: still produces a working binary.

### DIFF
--- a/docs/dev/decisions/native-backend-architecture-scope.md
+++ b/docs/dev/decisions/native-backend-architecture-scope.md
@@ -12,6 +12,11 @@ riscv64 is the designated pathfinder if that happens.
 Prerequisite shipped independently of any port: #1656 — `kaappi compile`
 on an unsupported architecture must refuse loudly instead of linking a
 binary that crashes (see "The failure mode that forced this decision").
+**Done:** `llvm_emit.native_backend_supported` (the single source of truth
+derived from `targetTriple`, so a future arch arm flips it automatically)
+now gates both `emitLlvmFile` — which exits nonzero naming the arch and
+pointing at the interpreter, before any codegen — and `kaappi doctor`,
+which reports one honest `arch` WARN instead of a misleading PASS trio.
 
 ## Problem
 
--- a/docs/dev/doctor.md
+++ b/docs/dev/doctor.md
@@ -19,7 +19,7 @@ dispatches it (`doctor.maybeRun`) before any VM, GC, or library setup exists.
 | `binary` | version, target triple, build mode, the running binary's path, and the `kaappi` a shell would pick from PATH (so a mismatch — "wrong binary on PATH" — is visible). |
 | `library` | the effective `.sld` search path in order: each `--lib-path`, the script directory (added per-run), `~/.kaappi/lib`, and the exe-relative `../lib` fallback — with whether each exists. |
 | `package-manager` | `thottam` on PATH; every package in `~/.kaappi/thottam.lock` has a matching source tree in `~/.kaappi/src`. |
-| `native-backend` | C-compiler discovery (`zig`/`cc`/`clang`/`gcc`), `libkaappi_rt.a` across the four documented locations, and a **smoke link** if both are found. |
+| `native-backend` | On an interpreter-tier arch (riscv64, s390x, ppc64le — where `kaappi compile` refuses, #1656) a single `arch` WARN says native compilation is unavailable and stops; otherwise C-compiler discovery (`zig`/`cc`/`clang`/`gcc`), `libkaappi_rt.a` across the four documented locations, and a **smoke link** if both are found. |
 | `repl` | `~/.kaappi` writable (where history is saved); terminal capabilities (`isatty`, `TERM`). |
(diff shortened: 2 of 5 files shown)


baijum:
Both CodeRabbit findings addressed in `ed76d84a`:

- **Major (llvm_emit.zig):** the `else` OS arms now return `null` instead of `*-unknown-unknown`, so `native_backend_supported` means "we can emit a **concrete** triple" — a supported arch on an unsupported OS is now refused too, closing the same -w link/crash path. No real host is affected (all six supported OSes have concrete triples; wasm is caught by the arch arm). Test extended with `aarch64/dragonfly` + `x86_64/illumos`.
- **Minor (doctor.md):** the native-backend row now names both runtime-archive variants (`libkaappi_rt.a`, and `kaappi_rt.lib` on Windows).

The one red check on the prior push was `windows-x64-test` failing `tests_reactor.test.addTimer fires when its deadline passes` ("expected 1, found 0") — a timing flake on the x64-via-ARM64-emulation leg, unrelated to this diff; the fresh run re-tests it.

Issue #1649, opened 2026-07-18
Derive srfi-N cond-expand feature identifiers from the supported-SRFI set


baijum:
## Summary

Kaappi does not expose `srfi-<n>` cond-expand feature identifiers. R7RS
implementations conventionally advertise each supported SRFI as a feature
identifier (`srfi-1`, `srfi-64`, …) so programs can probe support without
attempting an import:

```scheme
(cond-expand (srfi-1 (import (srfi 1)) ...)
             (else ...))
```

Today this always takes the `else` branch: symbolic feature requirements are
checked against `types.platform_features` only (see `evalLibFeatureReq` in
`src/vm_library.zig` and the compiler's `evalFeatureReq` in
`src/compiler_conditionals.zig`), which lists platform/subsystem identifiers
but no SRFI identifiers. This is notable because Kaappi ships SRFI 0
(`cond-expand`) itself, whose spec is built around exactly these identifiers.
The workaround — `(cond-expand ((library (srfi 1)) …))` — works but is the
less idiomatic spelling.

Noticed while implementing SRFI 261 (#1645), which needed a way to advertise
support and settled for documentation because no feature-id mechanism exists.

## Proposed approach

Derive, never hardcode (the #1517 principle — `kaappi features` already
derives its SRFI list this way):

- **Built-in SRFIs** (1, 9, 13, 18, 39, 69, 133, 170, 254): derivable from the
  library-spec tables that register `srfi.*` libraries (`src/library.zig`).
- **Portable SRFIs**: a feature identifier should be true iff the
  corresponding `.sld` is actually loadable *for this VM's configuration*
  (lib paths, sandbox, bundled files) — i.e. answer `srfi-<n>` by routing
  through the same check as `(library (srfi <n>))` (`libraryIsAvailable`),
  rather than baking a list at compile time. That keeps the identifier
  truthful under `--sandbox`, on WASM, and for stripped installs.

Concretely: in both `evalFeatureReq` implementations, before the
`platform_features` scan, match the symbol against `srfi-<digits>` and answer
via the existing library-availability check for `(srfi <n>)`. Reuses
`srfi261Suffix`/`libraryIsAvailable` from #1645's machinery.

Also decide whether `kaappi features` should print the derived SRFI feature
identifiers in its cond-expand section (it already prints the SRFI library
list separately).

## Acceptance criteria

[...]

Issue #1494, opened 2026-07-13
LLVM backend: cache kaappi_eval fallback compilation per call site


baijum:
Part of #1491.

## Problem

Forms the native backend cannot lower — `letrec`, `letrec*`, `cond`, `case`,
`do`, `guard`, `quasiquote`, named `let` — are serialized to a source string and
compiled via `@kaappi_eval` (`emitFormEval` / `emitLetEvalFallback` in
`src/llvm_emit.zig`; `emitLambdaViaEval` in `src/llvm_emit_lambda.zig`).

`kaappi_eval` **re-parses and re-compiles the source string every time the
enclosing code executes it** — not once at startup. Inside a loop body or a
frequently-called function this is a severe, easily-overlooked cliff.

## Proposed change

Add a caching eval entry point, e.g.
`kaappi_eval_cached(vm, str, len, *slot) -> Value` in `src/runtime_exports.zig`,
that compiles the source once, stashes the resulting `Function*` (or closure) in
a per-call-site global slot, and on subsequent calls executes the cached form
directly. Emit one global slot per fallback call site.

This removes the re-compile cost for the entire eval-fallback class without
needing full native lowering of each form — a stepping stone toward the
per-form native lowering issues.

## Acceptance criteria

- Fallback forms compiled at most once per call site (verify via a counter or
  profiling on a loop that hits a `cond`/`case`).
- Behavior identical to today (`tests/e2e` + relevant Scheme suites green).
- GC-safe: the cached `Function*`/closure is rooted appropriately.

PR #1663, opened 2026-07-19
Splice top-level cond-expand into top-level forms (#1661)

### LINKED ISSUE #1661: Top-level cond-expand doesn't splice its body — nested import emits spurious errors
## Summary

A `cond-expand` at the top level is compiled as an ordinary expression rather
than being recognized by `handleTopLevelForm` (`src/vm_eval.zig`). As a result,
`import` (and other top-level-only declaration forms) nested inside a matched
clause emit a spurious `KP3001 undefined variable 'srfi'` and the program exits
1 — even though the import's side effect (loading the library) still happens.

```scheme
;; exits 1 with: error[KP3001]: undefined variable 'srfi'
(cond-expand (else (import (srfi 1))))
(display (fold + 0 '(1 2 3)))  ;; still prints 6
```

This reproduces with every guard — `else`, `(library (srfi N))`, and the new
`srfi-N` feature ids (#1649) — so it is independent of which clause matches: the
matched body is compiled as an expression where `import` is not a recognized
form and `(srfi 1)` is analyzed as a call to an undefined `srfi`.

## Impact

The idiomatic top-level probe from #1649 —

```scheme
(cond-expand (srfi-1 (import (srfi 1))) (else ...))
```

— selects the right branch and the library loads, but the program prints an
error and exits nonzero. The library-level form (a `cond-expand` with `import`
declarations inside `define-library`) is unaffected and is the recommended
spelling today.

## Proposed fix

Make `handleTopLevelForm` (and `handleTopLevelBegin` / `isSpecialTopLevelForm`)
recognize `cond-expand`: evaluate its clauses with the existing
`evalLibFeatureReq`, then splice the matched clause's body as top-level forms
(the same splicing `begin` already does). This is the R7RS-correct behavior —
a top-level `cond-expand` expands to the selected forms in a top-level context —
and fixes nested `import`/`define`/`define-library` for all guards at once.

Care: `cond-expand` is also used in expression position (e.g. inside `define`),
which must keep going through the compiler; only the *top-level* occurrence
changes. Add regression coverage for the top-level `import`-in-`cond-expand`
exit code and for `cond-expand` still usable as a top-level expression value.

Noticed while implementing #1649 (PR #1660).

baijum:
## Summary

Fixes #1661. A `cond-expand` at the top level was compiled as an ordinary
expression, so an `import` (or any top-level-only declaration) nested in the
matched clause was mis-compiled — `(srfi 1)` read as a call to an undefined
`srfi`, printing `error[KP3001]: undefined variable 'srfi'` and exiting 1, even
though the import's side effect still ran. This defeated the idiomatic #1649
probe `(cond-expand (srfi-1 (import (srfi 1))) (else …))`.

```scheme
;; before: prints the KP3001 error and exits 1 (but still prints 6)
(cond-expand (else (import (srfi 1))))
(display (fold + 0 '(1 2 3)))   ; => 6
```

## Fix

R7RS 4.2.1: a top-level `cond-expand` expands to the selected clause's forms in
a **top-level context**. So `handleTopLevelForm` (`src/vm_eval.zig`) now
recognizes `cond-expand`:

- selects the first satisfied clause (or `else`) with the existing
  `evalLibFeatureReq` — the same live-registry evaluator `define-library` uses,
  so `else`, `(library (srfi N))`, and the `srfi-N` feature ids (#1649) all
  resolve identically;
- splices the matched clause's body through `handleTopLevelBegin`, exactly as
  top-level `begin` already does;
- `isSpecialTopLevelForm` learns `cond-expand` too, so the native eval-cache
  (#1494) declines to cache it.

Expression-position `cond-expand` (inside `define`, as a call argument, …) is
untouched — it never reaches `handleTopLevelForm` and still goes through the
compiler. A no-match top-level `cond-expand` yields void, matching the
compiler; a malformed one is a `KP2001` syntax error, matching the compiler.

## `kaappi check` had the same bug

The compile-only linter had the identical splice gap, surfacing as spurious
`KP4001` warnings — the clause compiled as an expression flagged `srfi`, and a
`define` nested in a matched clause was never gathered as a top-level name, so a
forward reference to it warned. Mirrored the splice in `src/check.zig`:
`checkForm` recurses into the selected clause (like `begin`), and
`collectFromForm` gathers names from every clause body (no VM there to pick one
— the same conservative over-approximation `test_selection.zig` already uses for
import deps). `docs/dev/check.md` updated.

`pipeline.zig` (the `expand`/`ir`/`ast` dumps) and `test_selection.zig` were
checked: the former only prints the form (no error — out of scope), and the
latter already walks every `cond-expand` clause.

## Tests

- `src/tests_libraries.zig` — top-level splice makes a nested `import` bind
  (`else` and `srfi-N` guards); top-level `cond-expand` still yields a value;
  expression-position `cond-expand` still composes; no-match yields void.
- `src/tests_check.zig` — a matched clause's `import` isn't flagged; a nested
  `define` is a known top-level name; an unsatisfied clause is not an error.
- `tests/scheme/errors/exit-code.sh` — top-level import-in-`cond-expand` exits 0.

Full unit suite (1226 tests), R7RS suite, SRFI/smoke/compliance/error suites all
green locally.

### DIFF
--- a/docs/dev/check.md
+++ b/docs/dev/check.md
@@ -108,6 +108,15 @@ their bound names are gathered structurally in a pre-pass so a forward reference
 is not warned. `define-syntax` registers its macro at *compile* time, so a
 later use of a same-file macro expands correctly without running anything.
 
+Top-level `begin` and `cond-expand` are *spliced*: their bodies are analysed as
+top-level forms, matching how the interpreter treats them (R7RS 5.1 / 4.2.1). A
+`cond-expand` clause is selected with `evalLibFeatureReq` — the same live-registry
+evaluator the runtime uses — so a nested `import` runs for effect and a nested
+`define`'s name is gathered as a top-level binding (the pre-pass conservatively
+gathers from every clause, since it has no VM to pick one). Without this, the
+matched clause compiled as an expression and a nested `import` flagged `srfi` as
+an unknown top-level variable (#1661).
+
 ## The literal-type table
 
 `check_lint.zig`'s `type_table` is deliberately narrow. An entry belongs there
--- a/src/check.zig
+++ b/src/check.zig
@@ -29,6 +29,7 @@ const types = @import("types.zig");
 const reader = @import("reader.zig");
 const compiler = @import("compiler.zig");
 const vm_mod = @import("vm.zig");
+const vm_library = @import("vm_library.zig");
 const ir_mod = @import("ir.zig");
 const diagnostics = @import("diagnostics.zig");
 const lsp_diagnostic = @import("lsp_diagnostic.zig");
@@ -165,6 +166,31 @@ fn checkForm(vm: *VM, ctx: *check_lint.Context, arena: std.mem.Allocator, expr: 
             return;
         }
 
+        // Top-level `cond-expand` splices the selected clause's body as
+        // top-level forms (R7RS 4.2.1) — recurse so a nested import runs for
+        // its effect and later forms see its bindings, mirroring the runtime's
+        // handleTopLevelCondExpand. Clause selection uses evalLibFeatureReq, the
+        // same evaluator the runtime uses, so both pick the same clause (#1661).
+        // A malformed or unsatisfied cond-expand falls through to be compiled as
+        // an expression (the compiler folds no-match to void and reports bad
+        // syntax) — the same fallback the runtime compiler path takes.
+        if (std.mem.eql(u8, name, "cond-expand")) {
+            var clauses = types.cdr(expr);
+            while (types.isPair(clauses)) : (clauses = types.cdr(clauses)) {
+                const clause = types.car(clauses);
+                if (!types.isPair(clause)) break;
+                const feature_req = types.car(clause);
+                const is_else = types.isSymbol(feature_req) and std.mem.eql(u8, types.symbolName(feature_req), "else");
+                if (is_else or vm_library.evalLibFeatureReq(vm, feature_req)) {
+                    var body = types.cdr(clause);
+                    while (types.isPair(body)) : (body = types.cdr(body)) {
+                        checkForm(vm, ctx, arena, types.car(body), path, line, col);
+                    }
+                    return;
+                }
+            }
+        }
+
         // Environment setup that later forms depend on: run it (suppressing lint
         // over the library/record code it compiles), so imported names/macros and
         // record accessors are in scope for the rest of the file.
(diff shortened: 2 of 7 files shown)


baijum:
Thanks @coderabbitai — dispositioned the three findings in `89c11440`:

**1. `handleTopLevelBegin` should use `runTopLevelFunction` (Critical) — fixed.** Correct catch: the spliced body reached `vm.execute()` directly, which would `resetExecutionState` and corrupt frame 0 if `eval` is re-entered from a native callback (`frame_count != 0`). Now routes through `runTopLevelFunction`, identical to `vm.execute` at true top level and re-entrant otherwise. It's a latent bug in top-level `begin` that the new `cond-expand` splice now also reaches.

**2. Use `th.TestContext` in the new tests — not applied.** All 34 test fixtures in `tests_libraries.zig` use the manual `memory.GC.init` + `th.makeTestVM` pattern (including the neighboring multi-eval `#1649` tests); zero use `TestContext`. Matching the file's established convention keeps the diff consistent with its surroundings.

**3. Validate the whole form / reject clauses after `else` — premise corrected, real sub-gap fixed.** The claim that trailing clauses after a match should error doesn't hold: `compileCondExpand` (expression position) returns as soon as it hits the matching/`else` clause and never validates later ones. Verified — `(+ 0 (cond-expand (else 41) malformed))` evaluates to `41` with no error, so rejecting it at top level would make the two positions *disagree*, not agree.

There was one genuine divergence hiding in there: an improper clause-list tail reached *without* a match — `(cond-expand (x 1) . junk)` — which the compiler rejects but the handler silently folded to void. Fixed just that (`if (clauses != NIL) return CompileError`). The improper-*body*-tail case (`(cond-expand (else 1 . 2))`) is left as-is: that's `handleTopLevelBegin`'s existing behavior, shared with top-level `begin`, so `cond-expand` stays consistent with `begin` by design.

Added parity tests in `tests_libraries.zig` covering the matched-clause short-circuit, a non-pair clause, and the improper clause-list tail.

Full unit suite 1230/1230 green.

baijum:
@coderabbitai second-pass point addressed in `7aa6d49f`: `handleTopLevelCondExpand` now validates the selected clause body is a proper list before splicing, so `(cond-expand (else 1 . junk))` errors at top level exactly as it does in expression position. `handleTopLevelBegin` (shared with top-level `begin`) is intentionally left as-is — its improper-tail leniency is a separate, pre-existing concern, so `cond-expand` validates its own structure while `begin` keeps its behavior.

`handleTopLevelCondExpand` now fully mirrors `compileCondExpand`'s malformed-form handling: non-pair clause → error, improper clause-list tail without a match → error, improper selected body → error, matched clause short-circuits without inspecting later clauses. Parity tests added.

PR #1672, opened 2026-07-19
Implement SRFI 264 (String Syntax for Scheme Regular Expressions)

### LINKED ISSUE #1666: Implement SRFI 264 (String Syntax for Scheme Regular Expressions)
## SRFI 264: String Syntax for Scheme Regular Expressions

**Status:** Final (finalized 2025-10-11) · **Author:** Sergei Egorov
**Spec:** https://srfi.schemers.org/srfi-264/srfi-264.html

### Summary

SRFI 264 defines **SSRE** — a compact, PCRE-inspired *string* syntax for regular
expressions that translates to the S-expression **SRE** notation of SRFI 115.
Instead of writing an SRE such as `(: "ab" (+ "c"))` by hand, you write the
pattern as an ordinary string, which SSRE parses into an SRE (and optionally
compiles to a regexp). It is a convenience layer over SRFI 115, not a new regex
engine, and the translation is bidirectional — `sre->ssre` renders an SRE back
into string syntax.

The string syntax covers the familiar PCRE-ish vocabulary: quantifiers
(`*`, `+`, `?`, `{n}`, `{n,m}`), anchors/boundaries (`^`, `$`, `\b`, `\A`, `\z`),
capturing / named / non-capturing groups (`(...)`, `(?<name>...)`, `(?:...)`),
lookaround (`(?=...)`, `(?!...)`, `(?<=...)`, `(?<!...)`), POSIX character classes
(`[:digit:]`, ...), a `{...}` character-set notation with set operations
(`{d}`, `{d|s}`, `{~s}`, `{p-[.]}`), and inline option flags
(`i`/`m`/`s`/`x`/`n`/`u`).

### Current state

Not implemented. There is no `lib/srfi/264.sld` and no `ssre` reference anywhere
in the tree; `(import (srfi 264))` fails.

### Both dependencies are already present and working

SRFI 264 depends on **SRFI 115** (Scheme Regular Expressions) and **SRFI 14**
(Character-Set Library) at runtime — and Kaappi already ships both as portable
libraries:

- `lib/srfi/115.sld` exports exactly the building blocks SSRE needs: `regexp`
  (SRE → regexp, for `ssre->regexp`), plus `regexp->sre` and `valid-sre?` (for
  `sre->ssre` and output validation).
- `lib/srfi/14.sld` exports `char-set`, `char-set-contains?`, etc. (backing the
  `{...}` character-set notation).

### Scope — a pure portable `.sld`

Unlike SRFI 267 (reader-level lexical syntax, #1639) or SRFI 271 (needed new
custom-port core, #1636), SRFI 264 is **pure Scheme**: it parses a string into an
S-expression and hands it to SRFI 115. The reference implementation (`ssre.scm`)
is already a portable R7RS library depending only on SRFI 115, so this should
land as a single `lib/srfi/264.sld` with **no core Zig work** — the same shape as
the majority of our portable SRFIs. The bulk of the effort is the SSRE-string →
SRE parser (and the reverse pretty-printer for `sre->ssre`).

### Exported identifiers

| Identifier | Kind | Signature | Behavior |
|---|---|---|---|
| `ssre->sre` | procedure | `(ssre->sre string)` | Parse an SSRE string into an SRFI-115 SRE. |
| `ssre->regexp` | procedure | `(ssre->regexp string)` | Parse and compile directly to a regexp (via `regexp`). |
| `sre->ssre` | procedure | `(sre->ssre sre)` | Render an SRE back into SSRE string syntax. |
| `ssre-definitions` | parameter | `(ssre-definitions)` / `(ssre-definitions bindings)` | Current set of named char-set / boundary / expression definitions. |
| `ssre-bind` | procedure | `(ssre-bind name type sre bindings)` | Return `bindings` extended with a named definition. |
| `ssre-unbind` | procedure | `(ssre-unbind name bindings)` | Return `bindings` with `name` removed. |

### Implementation notes

- Start from the SRFI's reference `ssre.scm` and port it to a
  `(define-library (srfi 264) ...)` wrapping SRFI 115 + SRFI 14, following the
  layout of the other portable SRFIs in `lib/srfi/`.
- Kaappi resolves `(import (srfi 264))` from `lib/srfi/264.sld`. The spec itself
  names no import form; `(srfi 264)` is our convention, matching every other
  SRFI.
- The `{...}` character-set sub-language and POSIX classes map onto SRFI 14
  char-sets and SRFI 115's own char-set SRE forms — reuse those rather than
  re-deriving classification tables.
- Verify SSRE features degrade cleanly to what our SRFI 115 actually supports
  (e.g. lookbehind, named groups): where SRFI 115 lacks an equivalent, the parser
  should raise a clear error rather than emit an invalid SRE. Use `valid-sre?` as
  a backstop in tests.

### Acceptance criteria

- [ ] `(import (srfi 264))` loads and exports the six identifiers above.
- [ ] `ssre->sre` parses the full syntax surface (quantifiers, anchors, all group
      kinds, lookaround, POSIX classes, `{...}` char-sets with set operations,
      option flags) into valid SREs.
- [ ] `ssre->regexp` produces a working regexp, e.g.
      `(regexp-search? (ssre->regexp "...") ...)` matches as expected.
- [ ] `sre->ssre` round-trips: SSRE → `ssre->sre` → `sre->ssre` yields an
      equivalent pattern, and re-parsing that yields a `valid-sre?`/`equal?` SRE.
- [ ] `ssre-definitions` / `ssre-bind` / `ssre-unbind` manage named definitions
      usable inside `{...}`.
- [ ] Conformance test under `tests/scheme/srfi/` (port/adapt the SRFI's
      `ssre-tests.scm`).
- [ ] SRFI count and list updated in `CLAUDE.md` and `README.md` (it will appear
      in `kaappi features`, which auto-scans `lib/srfi/`).

### References

- Spec: https://srfi.schemers.org/srfi-264/srfi-264.html
- Reference implementation: `ssre.scm` / `ssre-tests.scm` (linked from the SRFI page)
- Existing deps in-tree: `lib/srfi/115.sld` (`regexp`, `regexp->sre`,
  `valid-sre?`), `lib/srfi/14.sld` (char-sets)
- Prior portable-SRFI issues for house style: #1636 (271), #1638 (263), #1639 (267)

baijum:
## Summary

Implements [SRFI 264: String Syntax for Scheme Regular Expressions](https://srfi.schemers.org/srfi-264/srfi-264.html) (SSRE) — a compact, PCRE-inspired string regex syntax that translates to the SRE S-expressions of SRFI 115. Closes #1666.

SSRE lets you write `"[[:digit:]]+"` where the SRE form is `(+ numeric)`. It is a convenience layer over SRFI 115, not a new regex engine, and the translation is bidirectional.

## What's here

- **`lib/srfi/264.sld`** — the portable `(srfi 264)` library: a faithful port of Sergei Egorov's MIT-licensed reference implementation wrapped in an R7RS `define-library` over `(srfi 115)`. The body is kept close to upstream so it can be re-synced with future revisions. Exports:
  - `ssre->sre` — SSRE string → SRE
  - `ssre->regexp` — SSRE string → compiled regexp (via SRFI 115 `regexp`)
  - `sre->ssre` — SRE → SSRE string (reverse direction)
  - `ssre-definitions` / `ssre-bind` / `ssre-unbind` — named char-set / boundary / expression definitions
- **`tests/scheme/srfi/srfi264.scm`** — the upstream conformance corpus (2751 parser/unparser cases) verbatim, with a Kaappi `(import …)` prologue and an exit-on-failure epilogue so `run-all.sh` and CI catch regressions. **All 2751 pass.**
- Docs: SRFI count 78 → 79 (69 portable) in README, CONFORMANCE, CLAUDE.md, the understanding map, and CHANGELOG.

## Dependencies — no core changes

Pure Scheme. The only runtime dependency is `regexp` from **SRFI 115** (both SRFI 115 and 14 were already in-tree); the reference implementation builds SRE char-set forms directly rather than SRFI 14 objects, so **no Zig / core changes** were needed.

## Notes

- The derived `(srfi srfi-264)` alias (SRFI 261) and the `cond-expand srfi-264` feature id (#1649) work automatically; `kaappi features` reports 264 from the build-time `lib/srfi` scan.
- Named-group *matching* works. Extracting a named submatch by symbol (`regexp-match-submatch m 'name`) is a pre-existing limitation of Kaappi's SRFI 115, unrelated to this PR — SSRE emits the correct `(-> name …)` SRE and it matches fine by index.

## Test plan

- [x] `kaappi tests/scheme/srfi/srfi264.scm` → **2751/2751 passed**, exit 0
- [x] Negative control (a deliberately broken expectation) → exits 1 and prints the failure (proves the epilogue guards CI)
- [x] `bash tests/scheme/errors/features.sh` → 44 passed / 0 failed
- [x] `zig build test -Dtest-filter=features` → pass (portable count now 69)
- [x] `(import (srfi srfi-264))` alias + `cond-expand srfi-264` verified end-to-end
- [x] Neighboring SRFI suites (115, 1, 128) + `smoke/basic` still green


baijum:
Thanks for the review. Addressed in `93b35c4c` (rebased onto latest `main`):

**Fixed**
- **Dead error guard (`lib/srfi/264.sld`)** — genuine latent bug. `ssre-syntax-error?` tested `(string? (cadr x))`, but `fail` raises the remaining source as a *char list* (and `ssre-fancy-error` consumes it with `length`), so the guard never matched and a raw list escaped instead of a formatted error. Changed to `(list? (cadr x))`; `(ssre->sre "a{")` now raises a proper `error-object?` with message `"ssre->sre: missing range start in repeat\na{\n  ^"`. Regression test added.
- **`CLAUDE.md` count** — the leading "78 SRFIs supported" was missed on the first pass; now 79. (The other five count sites were already updated.)

**Added** — `tests/scheme/srfi/srfi264-behavior.scm` (SRFI-64): `ssre->regexp` matching through SRFI 115, the `ssre-bind`/`ssre-unbind` lifecycle under `parameterize`, and the error-object regression above.

**Declined, with reasons**
- **`#f` upper bound / `%run-repeat`** — false positive. The `{m,}` case compiles and matches (`(ssre->regexp "a{2,}")` matches `"aaaaaa"`); `ssre->sre` emits `(>= m …)`, not a literal `#f`. The only compile failure is *non-greedy* quantifiers (`a+?` → "unknown SRE `**?`"), which is a pre-existing gap in Kaappi's SRFI 115, independent of the `#f` theory — `ssre->sre` still returns the correct `(**? …)` SRE. Noted in the behavioral test header.
- **Convert the 2751-case corpus to SRFI-64 / `parameterize` the corpus setup** — kept `srfi264.scm` a *verbatim* port of the upstream `ssre-tests.scm` (only an `import` prologue + exit-on-failure epilogue added) so it can be re-synced with future reference revisions. New behavioral coverage uses SRFI-64 and `parameterize` instead, per the repo's "prefer SRFI-64 for new tests" convention.
- **Changelog case count** — `2751` is the authoritative figure the suite reports (2754 assertions minus 3 `test-error` decrements); left as-is.

The one failing check on the previous run was `windows-x64-test` — an unrelated timing flake (`tests_reactor…addTimer fires when its deadline passes`, on the emulated x64 leg), not touched by this pure-Scheme change. The fresh run re-exercises it.

Issue #1645, opened 2026-07-18
Implement SRFI 261 (Portable SRFI Library Reference)


baijum:
## Summary

[SRFI 261: Portable SRFI Library Reference](https://srfi.schemers.org/srfi-261/srfi-261.html)
was **finalized 2025-12-07** (author: WANG Zheng). Kaappi does not support it
yet — it is absent from the SRFI list in `CLAUDE.md`/`README.md`, and there are
no `srfi-<n>` / mnemonic import references anywhere in the codebase.

Unlike a normal SRFI, **261 specifies no procedures, syntax, or reference
implementation** — it is purely a *naming convention* for how SRFI libraries
are referenced in `import` forms. SRFI 97 used a colon separator
(`(srfi :1 lists)`), which is a reserved character on Windows filesystems and in
RFC 3986 URIs. SRFI 261 replaces it with hyphenated, colon-free forms so the
same import spelling maps to a valid path/URI everywhere.

Implementing it in Kaappi therefore means teaching the **import resolver** to
recognize the new reference forms — not adding a new `lib/srfi/*.sld` library.

## What the SRFI specifies

Three reference forms for referencing SRFI number `<n>`:

| Form | Conformance | Example | Kaappi today |
|---|---|---|---|
| `(srfi srfi-<n>)` | primary portable — **should** | `(srfi srfi-1)` | ✗ not recognized |
| `(srfi <mnemonic>-<n>)` | modern alternative | `(srfi lists-1)` | ✗ not recognized |
| `(srfi <n>)` | R7RS-restricted — **could** (optional) | `(srfi 1)` | ✓ already works |
| `(srfi :<n> <id> …)` | SRFI 97 back-compat (SRFIs < 261) | `(srfi :1 lists)` | ✗ — **won't support** (see decisions) |

The SRFI also assigns mnemonic names to 181 finalized SRFIs. Examples relevant
to what Kaappi ships:

| SRFI | Mnemonic | | SRFI | Mnemonic |
|---|---|---|---|---|
| 1 | `lists` | | 69 | `basic-hash-tables` |
| 13 | `strings` | | 133 | `vectors` |
| 43 | `vectors` | | … | … |

Note the deliberate collision: **43 and 133 both map to `vectors`**, which is
exactly why the mnemonic form carries the numeric suffix (`vectors-43` vs
`vectors-133`). The number is the disambiguator; the mnemonic is documentation.

**Conformance is permissive** ("should"/"could"). The one form Kaappi is
[...]

PR #1674, opened 2026-07-19
Add SRFI 260 (Generated Symbols)

### LINKED ISSUE #1668: Implement SRFI 260 (Generated Symbols)
## Summary

[SRFI 260: Generated Symbols](https://srfi.schemers.org/srfi-260/srfi-260.html)
was **finalized 2025-05-07** (author: Marc Nieper-Wißkirchen). Kaappi does not
support it yet — it is absent from the SRFI list in `CLAUDE.md`/`README.md`, has
no `lib/srfi/260.sld`, and there is no `generate-symbol` procedure anywhere in
the codebase.

SRFI 260 specifies a single procedure that produces symbols with **unique,
unpredictable names**. The motivating use case is Scheme's syntax system, where
distinct identifiers must be minted without risk of accidental capture. It is a
deliberate contrast to SRFI 258 (uninterned symbols): a *generated* symbol keeps
**write/read invariance** and the classical rule that *two symbols are identical
iff their names are equal*.

## What the SRFI specifies

One procedure:

```
(generate-symbol)              → symbol
(generate-symbol pretty-name)  → symbol    ; pretty-name is a string
```

- Returns a symbol "guaranteed to be unique for all practical purposes."
- The optional `pretty-name` **may** be used for display/debugging but does
  **not** determine the symbol's actual name.
- **Write/read invariant:** writing a generated symbol and reading it back must
  yield an `eqv?` symbol — because it has a genuine (if unpredictable) name,
  unlike an uninterned symbol.
- The name may be generated **lazily** — only materialized when `symbol->string`
  (or `write`) actually needs it. This is an optional optimization ("need only"),
  not a requirement.
- When a name is materialized it should carry high entropy — the SRFI suggests
  UUID-level (**~122 bits**); the sample implementation reads `/dev/random`.

No reader/lexical-syntax changes. No special GC requirements.

## Why this is a natural fit for Kaappi

Kaappi has **no uninterned symbols** — every symbol is interned by name in a
single `StringHashMap` (`GC.allocSymbol`, `src/memory.zig:266`), so two symbols
are `eq?` iff their names are equal. That means:

- **Write/read invariance is free.** Any symbol Kaappi can produce already has a
  concrete interned name, so `(read (open-input-string (write-to-string s)))`
  already returns an `eq?`/`eqv?` symbol. SRFI 260 semantics reduce to
  *"intern a symbol under a fresh, unpredictable, collision-free name."*
- **`generate-symbol` ≈ `string->symbol` of a unique high-entropy name.** The
  whole SRFI is `string->symbol` applied to a name that (a) is astronomically
  unlikely to collide with any user symbol and (b) is unpredictable.

The only real design question is **where the unique name comes from** (entropy +
in-process uniqueness), and **built-in primitive vs. portable `.sld`**.

### Option A — built-in primitive (recommended)

Add `generate-symbol` as a Zig primitive that interns via `GC.allocSymbol`:

- **In-process uniqueness:** a process-global **atomic counter** (same lock-free
  pattern as `symbol_mutex`) — guarantees no two calls ever collide, with zero
  dependence on RNG quality or seeding.
- **Unpredictability / cross-process uniqueness:** mix in real OS entropy. Reuse
  the plumbing already added for **SRFI 271** (`getrandom`/`/dev/urandom`, and
  WASI `random_get`) so the browser/WASI build works too.
- **Name shape:** `"<pretty>.<counter>.<hex-random>"` (default prefix e.g. `g`).
  `pretty-name`, when supplied, becomes the human-readable prefix.
- Register a built-in `(srfi 260)` library in `src/library.zig` exporting
  `generate-symbol` (this would make 260 the 10th built-in SRFI).

Rationale for preferring this: the SRFI explicitly asks for high-quality entropy
and unpredictability, and a counter gives a hard in-process uniqueness guarantee
that a pure-RNG portable version cannot. It also sidesteps the seeding hazard
below.

*(Optional, non-blocking: lazy naming — allocate an identity-only symbol and
defer name materialization until `symbol->string`. Kaappi interns eagerly by
name today, so eager generation is simplest and fully compliant; laziness can be
a later optimization if profiling ever motivates it.)*

### Option B — portable `lib/srfi/260.sld`

Matches the 68 other portable SRFIs; no engine changes:

```scheme
(define-library (srfi 260)
  (import (scheme base) (srfi 27))   ; random-integer
  (export generate-symbol)
  (begin
    (define counter 0)
    (define (generate-symbol . maybe-pretty)
      (let ((pretty (if (pair? maybe-pretty) (car maybe-pretty) "g"))
            (n counter))
        (set! counter (+ counter 1))
        (string->symbol
          (string-append pretty "." (number->string n) "."
                         (number->string (random-integer (expt 2 122)) 16)))))))
```

**Caveat that motivates Option A:** SRFI 27's `default-random-source` is
deterministically seeded unless `random-source-randomize!` is called, and Kaappi
threads run with independent per-thread VMs/GCs. A pure-RNG portable version can
therefore mint identical names in two threads (identical seed → identical
stream), and the counter is likewise per-thread state. The names would still be
*predictable* too, which the SRFI warns against. A portable version must handle
seeding/threading deliberately; the built-in atomic-counter + OS-entropy version
does not have this problem.

## Registration / derived behavior (either option)

- `(import (srfi 260))` — via file-path resolution (Option B) or the built-in
  registry key `srfi.260` in `src/library.zig` (Option A).
- **Free, derived automatically:**
  - `srfi-260` `cond-expand` feature id (#1649 — derived from the supported-SRFI
    set, never listed).
  - SRFI 261 resolution: `(srfi srfi-260)` / `(srfi <mnemonic>-260)` fall back to
    `(srfi 260)` (#1645 — trailing number is authoritative).
- **Docs/counts to update:** the SRFI list + counts in `CLAUDE.md` (line ~312:
  "78 SRFIs supported … 68 portable") and `README.md`. Adding 260 makes it **79
  supported** (Option A: 10 built-in / 68 portable; Option B: 9 built-in / 69
  portable). Keep `kaappi features` and `tests/scheme/smoke/features-consistency.scm`
  consistent.

## Tests

- Conformance under `tests/scheme/srfi/` (+ smoke import):
  - `(symbol? (generate-symbol))` and `(symbol? (generate-symbol "x"))`.
  - **Distinctness:** `(not (eq? (generate-symbol) (generate-symbol)))`, and a
    loop minting N symbols all pairwise distinct.
  - **Write/read invariance:** `(let ((s (generate-symbol)))
    (eqv? s (read (open-input-string (symbol->string-then-write s)))))` — i.e.
    round-trip through `write`/`read` returns an `eq?` symbol.
  - **Pretty-name is advisory:** two `(generate-symbol "foo")` calls are still
    distinct; `pretty-name` need not be a prefix of the result (don't over-assert
    the name shape).
- If Option A: a Zig unit test in `src/tests_*.zig` for the primitive +
  `-Dgc-stress=true` safety (the primitive allocates a symbol).

## Decision to make

1. **Built-in primitive (Option A, recommended) vs. portable `.sld` (Option B)?**
2. Entropy source for Option A (reuse SRFI 271's `getrandom`/`/dev/urandom` +
   WASI `random_get`).
3. Eager vs. lazy name materialization (recommend eager; lazy is optional).

## References

- Spec: https://srfi.schemers.org/srfi-260/srfi-260.html
- Prior SRFI work for pattern: #1636 (SRFI 271, OS entropy plumbing),
  #1645 (SRFI 261 resolution), #1649 (`srfi-<n>` feature ids), #1637 (SRFI 254).

baijum:
## Summary

Implements [SRFI 260: Generated Symbols](https://srfi.schemers.org/srfi-260/srfi-260.html) (Final, 2025-05-07) — the single procedure `generate-symbol`:

```scheme
(generate-symbol)              ; => a fresh unique symbol
(generate-symbol pretty-name)  ; pretty-name is a display-hint string
```

Every call returns a fresh symbol whose name is unique "for all practical purposes" and unpredictable. The property that distinguishes SRFI 260 from uninterned symbols (SRFI 258) is **write/read invariance**: a generated symbol printed and read back is `eq?` to the original.

Closes #1668.

## Why this shape

Kaappi interns every symbol by name in one table and has **no uninterned symbols**, so write/read invariance is automatic — any symbol we mint already round-trips through `write`/`read` to an `eq?` symbol (the printer bar-quotes names that need it). The SRFI therefore reduces to *"intern a fresh, unpredictable, collision-free name"*, which is a single native primitive — no reader, GC, or VM changes.

The generated name is `"<pretty>.<counter>.<128-bit-random-hex>"`:

- a **process-global atomic counter** is a hard in-process uniqueness guarantee, independent of entropy quality and of SRFI-18 threads (which run independent VMs but share the static), and
- **128 bits of OS entropy** (`platform.osRandomBytes`, reused from the SRFI 271 plumbing — `getrandom`/`arc4random`/`RtlGenRandom`/WASI `random_get`) make the name unpredictable and collision-free across processes and reads.

The optional `pretty-name` is a display hint only (per the SRFI it "does not determine the generated symbol's actual name"); it is used as the prefix, so two calls with the same pretty-name still yield distinct symbols.

Wired as the **10th built-in SRFI** through a single `primitives.Lib` tag (`srfi_260`), so availability, the `srfi-260` `cond-expand` id (#1649), `(srfi srfi-260)` SRFI-261 resolution (#1645), and the `kaappi features` listing all derive automatically — no second list to maintain.

## Changes

- `src/primitives_srfi260.zig` — the `generate-symbol` primitive.
- `src/primitives.zig` — `srfi_260` `Lib` tag, `canonicalName`, and `all_specs` entry.
- `src/tests_srfi260.zig` + `src/vm_tests.zig` — 9 unit tests (shape, freshness, write/read invariance, arity/type errors).
- `tests/scheme/srfi/srfi260.scm` — 12 SRFI-64 conformance tests.
- `CLAUDE.md` / `README.md` / `CONFORMANCE.md` — 78 → 79 SRFIs, 9 → 10 built-in, plus a design note and coverage section; `CHANGELOG.md` entry; `features.zig` count comment.

## Test plan

- `zig build test` — full unit suite green (incl. the 9 new tests).
- `bash tests/scheme/run-all.sh` — **1881 pass, 0 fail** (486 Scheme files + 1395 R7RS), incl. `srfi260.scm`.
- `tests/scheme/errors/features.sh` — 44/44 (verifies 260 is built-in, not portable, no overlap).
- Manual: distinctness, write/read `eq?` round-trip, 1000-symbol no-collision, `--sandbox`, and `kaappi features` showing `built-in (10): … 254 260`.


baijum:
Rebased onto `main` (picking up SRFI 264 #1672 and SRFI 259+229 #1673) and resolved the doc-count conflicts → **82 SRFIs, 10 built-in, 71 portable**.

Addressed the CodeRabbit review — verified each point against the code:

**Fixed**
- **`src/features.zig` built-in count** — corrected to `10`. `builtinSrfis()` returns the nine `srfi_*` `Lib` tags (1, 13, 18, 39, 69, 133, 170, 254, 260) plus syntax-only `srfi.9`; the old comment predated SRFI 254.
- **Entropy fallback in `generate-symbol`** — on OS-CSPRNG failure it now raises a catchable error instead of falling back to `randomSeed64` (which can yield predictable clock-derived bytes). This matches the principle `platform.osRandomBytes` already documents for the SRFI-271 random ports: surface the failure rather than hand out predictable output. The atomic counter still guarantees uniqueness on the success path.
- **`src/tests_srfi260.zig`** — added a `string-prefix?` assertion that the generated name begins with the supplied pretty-name, alongside the existing freshness check.

**Intentionally skipped** (with reasons)
- **Move the CLAUDE.md SRFI note into `docs/dev/`** — the one-sentence note follows the established inline pattern that every other built-in SRFI (254, 267, 261, …) already uses in that same paragraph. Relocating just this entry would make `CLAUDE.md` inconsistent; restructuring the whole section is out of scope for this PR.
- **Remove the README SRFI counts** — this repo deliberately maintains these counts (three spots in `README.md`, plus `CONFORMANCE.md`) and bumps them with each SRFI PR, matching the sibling SRFI PRs merged today. Keeping them accurate is the existing convention.

Green after the changes: `zig build test`, `run-all.sh` (incl. `srfi260.scm` 12/12), and `tests/scheme/errors/features.sh`.

Issue #1517, opened 2026-07-13
kaappi features --json: machine-readable capability discovery (KEP-0004 for the CLI)


baijum:
## Context

KEP-0004 gives *Scheme code* capability discovery via `cond-expand` identifiers (`kaappi-fibers`, `kaappi-reactor`, `kaappi-threads`). An agent's first question — "what am I working with?" — currently requires parsing `--help` and probing behavior. This extends KEP-0004's philosophy to the CLI boundary.

## What

`kaappi features [--json]`:

- version + git build id, target triple, build mode
- compiled-in subsystems — exactly the KEP-0004 feature identifiers, with their values on this build
- available SRFIs (built-in vs portable `.sld`)
- relevant build limits (initial frame/register capacities, GC threshold) and whether `--sandbox` is available
- JSON is the primary interface; a human-readable table is secondary

**Single source of truth**: the same table that drives `cond-expand` feature resolution must drive this output — no second list to drift. Same for the SRFI list: derived from the registry + `lib/srfi/`, not hardcoded.

## Acceptance criteria

- [ ] `features` subcommand with `--json` and table output
- [ ] Feature list provably shared with `cond-expand` resolution (one table in source, asserted by test)
- [ ] SRFI list generated, not hardcoded
- [ ] Referenced from the kaappi-lang.org conformance page (which already documents the identifiers)

Part of #1503.

PR #1675, opened 2026-07-19
Implement SRFI 258 (Uninterned Symbols)

### LINKED ISSUE #1670: Implement SRFI 258 (Uninterned Symbols)
## Summary

[SRFI 258: Uninterned Symbols](https://srfi.schemers.org/srfi-258/srfi-258.html)
(Wolfgang Corcoran-Mathe) was **finalized 2025-03-26**. Kaappi does not
implement it yet — it is absent from the SRFI list in `CLAUDE.md`/`README.md`,
and there are no `uninterned`, `symbol-interned?`, or
`generate-uninterned-symbol` references anywhere in the codebase.

An **uninterned symbol** is a symbol that is not `eqv?` to any other symbol,
even one with the same name. They are useful for macro programming and for
generating guaranteed-unique identifiers. A portable implementation is
impossible: the SRFI must be backed by native (Zig) support, because it needs
to allocate symbols that bypass the interning table.

## Libraries to provide

- `(srfi 258)` — Kaappi's canonical name.
- The SRFI's own module name is `(srfi :258 uninterned-symbols)` (R6RS-style).
  The R7RS/Kaappi form `(srfi 258)` is the one we register; the SRFI 261
  fallbacks (`(srfi srfi-258)`, `(srfi uninterned-symbols-258)`) then resolve
  automatically.

## Exported procedures

| Procedure | Behavior |
|---|---|
| `(string->uninterned-symbol string)` → symbol | Returns a freshly-allocated uninterned symbol whose name is `string`. Never `eqv?` to any other symbol, including one produced from an equal string. |
| `(symbol-interned? symbol)` → boolean | `#t` if `symbol` is interned (the normal case), `#f` if uninterned. Signals an error on a non-symbol. |
| `(generate-uninterned-symbol [prefix])` → symbol | Returns a fresh uninterned symbol with a likely-unique name. Optional `prefix` (a string or symbol) is prepended to the generated name. |

## Kaappi implementation notes

The good news: symbol **equality already works correctly for free**, because
Kaappi never compares symbol *names* for identity — it compares pointers.

- `eqv?`/`eq?` (`src/primitives.zig:643`) start with `if (args[0] == args[1])`
  and fall through to `#f` for symbols — no name-comparison branch.
- `equal?` → `deepEqual` (`src/primitives.zig:695`) is pointer-identity-first
  with no symbol-name branch either.

Interned symbols are unique per name (so pointer identity == name identity
today), and each uninterned symbol is a distinct fresh allocation. Therefore
two like-named uninterned symbols are automatically **not** `eqv?`/`equal?`,
and an uninterned symbol is **not** `eqv?` to a like-named interned one — with
**no changes to the equality primitives**.

The actual work is in allocation, GC, and a way to tell the two apart:

- **Distinguish interned vs. uninterned.** `Symbol` is currently
  `struct { header: Object, name: []const u8 }` (`src/types.zig:303`). Add a
  flag — a `bool` field on the struct, or a spare bit on the `Object` header —
  so `symbol-interned?` has something to read.

- **New allocator that bypasses interning.** `allocSymbol`
  (`src/memory.zig:266`) does `sym_table.get(name)` / `sym_table.put(...)` under
  `symbol_mutex` and, for the parent GC, marks the symbol permanent via the
  interned-symbol root scan. `string->uninterned-symbol` needs an
  `allocUninternedSymbol` variant that **skips the table get/put entirely**,
  sets the uninterned flag, and `trackObject`s the symbol as a normal
  collectable heap object (i.e. *not* a permanent root).

- **GC sweep must free uninterned symbols (subtle).** `freeObject`
  (`src/gc_collect.zig:297` and `:622`) currently lists `.symbol` in the
  `=> {}` "nothing to free" arm. That is safe *today only because interned
  symbols are never swept* — they are kept alive permanently by the
  `gc.symbols` root scan (`src/gc_collect.zig:679`). Uninterned symbols **will**
  be swept, so `freeObject` must free the uninterned symbol's `name` slice (and
  destroy the struct). Guard on the interned flag so interned symbols keep their
  current no-op path. `markValue` for `.symbol` needs no change. Verify under
  `-Dgc-stress=true` (per `.claude/rules/gc-safety.md`).

- **`generate-uninterned-symbol`.** A monotonically-increasing counter feeding
  a name like `g<N>` (or `<prefix><N>`), then routed through
  `allocUninternedSymbol`. Note this is a *distinct* mechanism from the
  expander's hygiene gensyms (`gensym_counter` in `src/expander.zig`), which
  produce ordinary **interned** `__hyg_`/`__nlet_` symbols — don't conflate the
  two.

- **Printer + reader (write/read invariance intentionally broken).** SRFI 258
  requires that `read` **signal an error** on the external representation of an
  uninterned symbol (R7RS 6.5 write/read invariance is deliberately given up).
  Pick an unreadable `write`/`display` representation for uninterned symbols in
  `src/printer.zig` (e.g. `#<uninterned-symbol name>` or a `name`+id form) and
  confirm `src/reader.zig` rejects it. **Design decision to settle in the PR:**
  exactly what that printed form looks like.

- **Registration.** This is a *built-in* SRFI (Zig primitives), not a portable
  `.sld`. Register the `(srfi 258)` library and its three exports (see how the
  other built-in SRFIs — 1, 9, 13, 18, 39, 69, 133, 170, 254 — are wired in
  `src/library.zig` / `src/vm_library.zig`), and add 258 to the built-in-SRFI
  list in `CLAUDE.md` and `README.md`. The `srfi-258` `cond-expand` feature
  identifier and the `kaappi features` SRFI listing then derive automatically
  via `srfiFeatureAvailable` (#1649) — do **not** add it to `(features)`.

### Likely files to touch

- `src/types.zig` — `Symbol` interned flag
- `src/memory.zig` — `allocUninternedSymbol`
- `src/gc_collect.zig` — free uninterned symbols in `freeObject`
- `src/primitives.zig` (or a small `src/primitives_srfi258.zig`) — the 3 procedures
- `src/printer.zig` — uninterned-symbol representation
- `src/reader.zig` — confirm/ensure that representation errors on `read`
- `src/library.zig`, `src/vm_library.zig` — register `(srfi 258)` + built-in list
- `CLAUDE.md`, `README.md` — SRFI count + built-in list
- `src/tests_*.zig`, `tests/scheme/srfi/` — tests

## Acceptance criteria

- [ ] `(import (srfi 258))` exports `string->uninterned-symbol`,
      `symbol-interned?`, `generate-uninterned-symbol`.
- [ ] `(eqv? (string->uninterned-symbol "x") (string->uninterned-symbol "x"))`
      ⇒ `#f`; `(eqv? (string->uninterned-symbol "x") 'x)` ⇒ `#f`.
- [ ] `(symbol-interned? 'x)` ⇒ `#t`;
      `(symbol-interned? (string->uninterned-symbol "x"))` ⇒ `#f`;
      `(symbol-interned? (generate-uninterned-symbol))` ⇒ `#f`.
- [ ] `(symbol->string (string->uninterned-symbol "foo"))` ⇒ `"foo"`; an
      uninterned symbol is still `symbol?`.
- [ ] `generate-uninterned-symbol` honors an optional string/symbol prefix and
      returns a fresh, uninterned, mutually-non-`eqv?` symbol on each call.
- [ ] `read` signals an error on the external representation of an uninterned
      symbol (write/read invariance intentionally broken).
- [ ] `(cond-expand (srfi-258 #t) (else #f))` ⇒ `#t`; `kaappi features` lists
      258 under built-in SRFIs; `(features)` is unchanged.
- [ ] Unit tests (`src/tests_*.zig`) and a Scheme conformance test under
      `tests/scheme/srfi/` pass, including under `zig build test -Dgc-stress=true`
      (uninterned symbols are collectable — this exercises the new sweep path).

baijum:
Implements [SRFI 258 — Uninterned Symbols](https://srfi.schemers.org/srfi-258/srfi-258.html) as the built-in `(srfi 258)` library. Closes #1670.

An uninterned symbol is a symbol that is never `eqv?` to any other symbol, even one built from the same name — useful for macro programming and guaranteed-unique identifiers.

## Procedures (`(import (srfi 258))`)

| Procedure | Behavior |
|---|---|
| `(string->uninterned-symbol string)` | Fresh uninterned symbol with the given name. |
| `(symbol-interned? symbol)` | `#t` for an ordinary symbol, `#f` for an uninterned one; errors on a non-symbol. |
| `(generate-uninterned-symbol [prefix])` | Fresh uninterned symbol with a likely-unique name; optional string/symbol prefix (default `g`). |

## Design

- **Equality is free.** Kaappi already compares symbols by object identity (`eqv?`/`eq?`/`equal?` are pointer-first with no symbol-name branch), so two uninterned symbols from equal strings — and an uninterned symbol vs. its like-named interned twin — are distinct with **no change to the equality primitives**. The only new state is a `Symbol.interned` flag.
- **Collectable, not permanent.** `GC.allocUninternedSymbol` bypasses the interning table; the interned-symbol root scan walks only that table, so an uninterned symbol is an ordinary collectable object (swept once unreachable, `freeObject` frees its name) rather than a permanent root.
- **Cross-thread.** `gc_deep_copy` preserves uninterned-ness (and shared identity within one copy) across SRFI-18 thread boundaries instead of re-interning by name.
- **No readable representation.** Per the SRFI, `write` emits an unreadable `#<uninterned-symbol name>` form and `read` rejects it (the reader errors on `#<`), deliberately breaking write/read invariance (R7RS 6.5). `display` still shows the bare name.
- **Auto-derived plumbing.** Registering through the `Lib` enum means `kaappi features`, the `srfi-258` `cond-expand` id (#1649), and `(import (srfi 258))` plus the SRFI 261 fallbacks (`(srfi srfi-258)`) all resolve with no extra wiring. `(features)` stays platform-only (#1517 invariant preserved).

Now **79 SRFIs (10 built-in)**.

## Tests

- `src/tests_srfi258.zig` — 9 unit tests: allocation semantics, GC sweep of unreachable uninterned symbols, interned-survives-vs-uninterned-swept, rooted-symbol survival, and end-to-end through the registered primitives.
- `tests/scheme/srfi/srfi-258.scm` — 22 SRFI-64 assertions covering the three procedures, all equivalences, the read-rejection requirement, and `display`.

## Verification

- Full unit suite `zig build test` — pass (also green under `-Dgc-stress=true`, which exercises the new collectable-symbol sweep path).
- `bash tests/scheme/run-all.sh` — **486 Scheme files pass / 0 fail**, R7RS **1395/1395**, total **1881 pass / 0 fail**.
- `tests/scheme/errors/features.sh` — 44/0; confirmed `(features)` does not list `srfi-258`, `kaappi features` does, and normal symbol printing is unaffected.


baijum:
Rebased onto `main` (picking up SRFI 264 + the non-final-SRFI CI guard) and pushed fixes.

### Fixed the wasm CI failure
The `generate-uninterned-symbol` gensym counter was a **64-bit atomic** — `@atomicRmw` rejects 64-bit integers on `wasm32-wasi` (single-threaded, no 64-bit atomics). Changed it to a **32-bit atomic**; wrap-around is harmless since an uninterned symbol's identity is guaranteed by allocation, not by its name. Verified `zig build wasm` builds locally.

### CodeRabbit review

**Addressed:**
- **Reject >1 argument in `generate-uninterned-symbol`** — added an explicit `args.len > 1` arity error (mirroring `ephemeron-ref`'s excess-arg check in `primitives_srfi254.zig`), plus a regression test `(generate-uninterned-symbol "a" "b")` in the Scheme conformance suite.
- **Deep-copy coverage** — added a `GC.deepCopy` unit test in `tests_srfi258.zig` that copies a pair holding the same uninterned symbol twice across two GC heaps and asserts the copy stays uninterned, differs from the source, and preserves shared identity within the copy.

**Not changed (with reasons):**
- **Module-level `registerXxx` function** — the modern spec-table primitives files (`primitives_srfi254.zig`, `primitives_random_port.zig`) are spec-only; registration is driven centrally by `all_specs` in `primitives.zig`. Adding a `registerXxx` here would double-register. The `src/CLAUDE.md` guideline predates the spec-table refactor.
- **Split oversized `types.zig`/`memory.zig`** — pre-existing and out of scope for this ~30-line addition; `types.zig` is exempt from the file-size policy, and splitting core type/GC files isn't warranted by this change.

### Re-verified on the rebased state
Full unit suite, `zig build wasm`, SRFI 258 units (incl. the new deep-copy test), `tests/scheme/srfi/srfi-258.scm`, `features.sh` (44/0), and the non-final-SRFI guard (`OK: all 80 implemented SRFIs are final`). Now **80 SRFIs (10 built-in, 69 portable)**.

baijum:
Rebased onto latest `main` (now past SRFI 259+229 and the /vm-test skill) and resolved the doc conflicts. Combined total is now **82 SRFIs (10 built-in, 71 portable)**, consistent across `CLAUDE.md`, `CONFORMANCE.md`, `README.md`, and `CHANGELOG.md`.

> [CodeRabbit] Synchronize the portable-library count — the `### Portable SRFIs (68 libraries)` heading was out of sync with the count line.

Fixed as part of the rebase: the heading now reads **71 libraries**, matching both the count line and the portable table. (The stale value came from my branch not tracking the heading when `main` bumped the portable count; taking `main`'s heading during conflict resolution resyncs it.)

Verified: `zig build`, the non-final-SRFI guard (`OK: all 82 implemented SRFIs are final`), `kaappi features` (built-in 10 incl. 258, portable 71 incl. 229/259/264), the SRFI 258 unit tests, and `tests/scheme/srfi/srfi-258.scm`. The code is unchanged from the previous (fully green) commit — only docs and the rebase base moved.

baijum:
Rebased onto latest `main` (now includes **SRFI 260 — Generated Symbols**, another built-in). Combined total is now **83 SRFIs (11 built-in, 71 portable)**.

**CI:** the prior `windows-x64-test` failure was `tests_reactor.test."addTimer fires when its deadline passes"` (`expected 1, found 0`) — a timing-sensitive reactor/timer test, unrelated to this change (it touches symbols/GC/printer, not the reactor) and green on the identical prior push. Re-running fresh with this push.

> [CodeRabbit] Synchronize the primitive-file inventory — the "Primitives (split into 21 files)" table omits `src/primitives_srfi258.zig`.

Fixed: added rows for `primitives_srfi258.zig` and `primitives_srfi260.zig` and set the header to the true file count (26). (The table was already missing a few pre-existing files — `srfi254`, `fiber`, `parallel`, `random_port`; rowing those out is left to a separate cleanup to keep this PR scoped.)

Because SRFI 260 is also a built-in symbol SRFI, this rebase had code conflicts (both add to the `Lib` enum, `all_specs`, and `vm_tests.zig`) — resolved by keeping both in numeric order. Verified: `zig build`, full unit suite, `zig build wasm` equivalent code unchanged, non-final guard (`OK: all 83 implemented SRFIs are final`), `kaappi features` (built-in 11 incl. 258+260), and both `srfi-258.scm` and `srfi260.scm`.

PR #1677, opened 2026-07-19
Implement SRFI 248 (Minimal Delimited Continuations)

### LINKED ISSUE #1669: Implement SRFI 248 (Minimal Delimited Continuations)
## SRFI 248: Minimal Delimited Continuations

**Status:** Final (finalized 2025-01-24) · **Author:** Marc Nieper-Wißkirchen
**Spec:** https://srfi.schemers.org/srfi-248/srfi-248.html

### Summary

SRFI 248 provides **delimited continuations by extending the exception system**
rather than introducing prompt tags. Its single essential primitive,
`with-unwind-handler`, installs an anonymous exception handler over a thunk;
when an exception is raised, the handler receives both the raised object **and a
delimited continuation** `k` representing the rest of the guarded thunk's
computation, up to the `with-unwind-handler` call. This is `shift0`/`reset0`
expressed through the handler mechanism, and it is enough to build algebraic
effect handlers on top of `raise-continuable`.

`empty-continuation?` reports whether that captured `k` is the empty delimited
continuation (i.e. the exception was raised in tail position of the thunk). The
SRFI also generalizes `guard` with an extra *continuation-variable* that binds
the captured delimited continuation, falling back to `raise-continuable` +
re-application of `k` when no clause matches.

### Current state

Not implemented. There is no `lib/srfi/248.sld`; `(import (srfi 248))` fails,
and 248 is absent from the supported-SRFI list. No `with-unwind-handler` /
`empty-continuation?` identifiers exist anywhere in the tree.

### Why this one is not a pure `.sld`

Most of our recent SRFI adds were portable Scheme (e.g. #1666 SRFI 264). This
one is different: it requires **genuine delimited-continuation support in the
core**, which Kaappi does not yet have. Today the VM exposes only *full*
continuations and escape continuations:

- `call/cc` / `call-with-current-continuation` — stack-copying full continuations
- `call/ec` / `call-with-escape-continuation` — one-shot escape continuations
- `dynamic-wind`, `with-exception-handler`, `raise`, `raise-continuable`, `guard`

There is **no** `shift`/`reset`/`shift0`/`reset0`, no prompt/control operator,
and no notion of a *delimited* slice of the continuation. The whole point of
SRFI 248 — handing the handler a re-invocable delimited continuation — needs
that machinery. Relevant code lives in `src/vm_continuations.zig`
(`captureContinuation`, `restoreContinuation`, `callWithCC`),
`src/primitives_control.zig`, and the wind/handler stacks in `src/vm.zig`.

### Exported identifiers

Library: `(srfi 248)` (R7RS). (R6RS names `(srfi :248)` / `(srfi :248 exceptions)`
are out of scope — we are R7RS-small.)

| Identifier | Kind | Signature | Behavior |
|---|---|---|---|
| `with-unwind-handler` | procedure | `(with-unwind-handler handler thunk)` | Run `thunk` with `handler` as an anonymous exception handler; on raise, call `(handler obj k)` where `k` is the delimited continuation from the raise point up to this call. Returns `thunk`'s results. |
| `empty-continuation?` | procedure | `(empty-continuation? k)` | `#t` iff `k` is the empty delimited continuation (exception raised in tail context of `thunk`), else `#f`. |
| `guard` | syntax | `(guard (var k-var clause ...) body ...)` | Extended `guard`: also binds the captured delimited continuation to `k-var`. If no clause matches (and no `else`), `raise-continuable` the object and apply `k` to the results. The plain `(guard (var clause ...) body)` form is unchanged. |
| `with-exception-handler` | procedure | `(with-exception-handler handler thunk)` | Re-export, identical to R7RS-small. |
| `raise` | procedure | `(raise obj)` | Re-export, identical to R7RS-small. |
| `raise-continuable` | procedure | `(raise-continuable obj)` | Re-export, identical to R7RS-small. |

### Implementation approach

Two viable paths; the first is recommended:

1. **Add a delimited-continuation primitive to the VM, then build SRFI 248 on
   top.** Introduce a prompt/control operator (naturally exposed also as
   `shift0`/`reset0`) that captures the continuation slice between a delimiter
   and the current point as a re-invocable object, cooperating with the existing
   `dynamic-wind` stack. `with-unwind-handler` then installs a delimiter + an
   exception handler that, on raise, reifies the slice from the raise point to
   the delimiter and passes it to the user handler; `empty-continuation?` is a
   cheap check on that reified slice. This is the clean, general foundation and
   unlocks generators / effect handlers beyond this SRFI.

2. **Filinski-style encoding over the existing stack-copying `call/cc` + a
   metacontinuation cell** (à la "Representing Monads"). Possible as a mostly
   portable `.sld`, but delicate around `dynamic-wind` re-entry, multi-shot
   re-invocation, and thread-local state — and `empty-continuation?` still needs
   a reliable emptiness test. Higher subtlety-per-line than path 1.

Note: the SRFI's sample implementation targets Guile (native
`call-with-prompt`/`abort-to-prompt`) and **omits `empty-continuation?`**, so
that predicate needs original work either way.

### Acceptance criteria

- [ ] `(import (srfi 248))` works; `248` added to the supported-SRFI list in
      `CLAUDE.md` and the `kaappi features` SRFIs section (derived, per #1649).
- [ ] `with-unwind-handler`, `empty-continuation?`, extended `guard`, and the
      three re-exports behave per spec, including correct interaction with
      `dynamic-wind` (winders run on delimited capture/resume).
- [ ] `empty-continuation?` correctly distinguishes tail-context raises.
- [ ] SRFI-64 conformance suite under `tests/scheme/srfi/`, plus the SRFI's own
      examples (algebraic-effect-style handler over `raise-continuable`).
- [ ] Green under `zig build test -Dgc-stress=true` (reified continuations rooted
      across allocation).

### Labels

enhancement · srfi · continuations · priority: low

baijum:
## Implement SRFI 248 (Minimal Delimited Continuations)

Closes #1669.

Adds **`(srfi 248)`** — `with-unwind-handler`, `empty-continuation?`, and the extended two-variable `guard`. SRFI 248 hooks delimited continuations into the exception system: the handler installed over a thunk receives the raised object *and* a delimited continuation representing the rest of the guarded thunk's computation, up to and including the `with-unwind-handler` call.

### Approach

A pure-Scheme Filinski `shift`/`reset` over `call/cc` is not enough on its own here: Kaappi's `raise-continuable` pops the current handler in the VM before invoking it, so a continuation captured while the handler runs would not include the handler and could not re-arm the prompt on resume. The fix is a small, general VM primitive plus a portable library on top:

- **VM — sticky exception handler** (`ExceptionHandler.sticky`, `src/types.zig` / `src/vm.zig`): `raise`/`raise-continuable` invoke a sticky handler *in place without popping it* (`src/primitives_control.zig`), so a `call/cc` snapshot taken while it handles includes it and a resume re-arms the prompt (reset0 semantics). It is captured/restored with the continuation (`src/vm_continuations.zig`).
- **VM — tail-context latch** (`native_call_was_tail`, `src/vm_dispatch.zig` / `src/vm_calls.zig`): every tail-call opcode sets it, every regular call clears it, so `empty-continuation?` can tell whether a `raise-continuable` occurred in tail context of the thunk.
- **Library** (`lib/srfi/248.sld`): a Filinski `reset*`/`shift*` over `call/cc` builds `with-unwind-handler`, `empty-continuation?`, and `guard`. The three helper primitives (`%call-with-unwind-handler`, `%unwind-raise-tail?`, `%pop-unwind-handler!`) live in a built-in sub-library `(srfi 248 primitives)` that the `.sld` imports and does not re-export (the registry shadows a same-named `.sld`, so the public `(srfi 248)` must stay file-only — the same reason SRFI 267 is split).

`(srfi 248)` is derived everywhere it should be: it counts as a portable SRFI (auto-scanned from `lib/srfi/`), `(cond-expand (srfi-248 …))` and `(cond-expand ((library (srfi 248)) …))` both answer correctly, and it stays out of `(features)` (the platform-only invariant, #1517).

### Behaviour verified

All SRFI 248 examples pass — the coroutine generator, `for-each->fold` / `string-fold`, effect-handler style, and both `empty-continuation?` cases (`#t` in tail context, `#f` otherwise) — plus the standard one-variable `guard` (`else`, `=>`, re-raise to an outer handler, native runtime errors) and `dynamic-wind` unwinding.

### Documented caveats

- **Single-shot**: each captured delimited continuation may be resumed at most once — re-entering the same `k` twice crosses the sticky-handler native frame, which Kaappi's stack-copying continuations cannot re-enter after it returns (the same restriction as continuations captured under native drivers). Every SRFI 248 idiom resumes each `k` once.
- **Handler timing**: the handler runs at the raise point rather than after unwinding to `with-unwind-handler`, so a handler side effect and a guarded-body `dynamic-wind` after-thunk run in the opposite order to what the SRFI wording implies. The captured continuation itself is unaffected.

Both are noted in `lib/srfi/248.sld`, `CONFORMANCE.md`, and `README.md`.

### Tests

- `tests/scheme/srfi/srfi248.scm` — 21 checks covering the whole API and the SRFI examples.
- `src/tests_srfi248.zig` — unit tests (green under `-Dgc-stress=true`).

Full suites green: Zig unit suite, `tests/scheme/run-all.sh` (**1881 pass, 0 fail**, incl. R7RS 1395/0).


baijum:
Rebased onto `main` (resolving the SRFI-count conflict with SRFI 264 — combined total is now **80 SRFIs / 70 portable**, both entries kept) and addressed the review:

- **`empty-continuation?` tail detection** (the frame_count-baseline finding): confirmed a real bug — a `raise-continuable` in tail position of a *non-tail-called* helper wrongly reported empty. Fixed in `dispatchStickyUnwind` by combining the immediate tail latch with the sticky handler's `frame_count` baseline (`native_call_was_tail && frame_count == entry.frame_count + 1`), which is exactly the baseline approach suggested — no cross-call propagation needed. Renamed the now-misleading `%unwind-raise-tail?` → `%unwind-raise-empty?` / `pending_raise_empty`. Regression tests added in both suites.
- **`CONFORMANCE.md` "68 libraries" heading** → **70**.
- **Fiber-local `*meta-k*`**: documented explicitly — the metacontinuation cell is per-VM (per OS thread), not fiber-local, so a `with-unwind-handler`/`guard` body must not span a fiber suspension point. Making it fiber-local isn't feasible without fiber-local Scheme storage.
- **SRFI-64 tests**: `tests/scheme/srfi/srfi248.scm` converted to SRFI-64 (`test-begin`/`test-equal`/`test-end`, exit 1 on failure), plus regression tests for the empty-continuation helper case and the documented guard/dynamic-wind ordering caveat.

Full unit suite green; `run-all.sh` **1883 pass, 0 fail** (R7RS 1395/0). The dynamic-wind-ordering and single-shot caveats remain documented in the `.sld`, `CONFORMANCE.md`, and `README.md` — the alternatives that fix ordering break `raise-continuable` reraise-to-outer under Kaappi's in-place handler mechanics.

baijum:
Rebased onto `main` (now **85 SRFIs / 11 built-in / 73 portable** — reconciled against SRFI 257/258/259+229/260 landing; `kaappi features` confirms 11 + 73), fixed the CI failure, and addressed the review.

### `benchmark-pr` failure — root-caused and fixed

The `continuations` benchmark really did regress (reproduced locally: main 4.54s → branch 5.21s). Cause: `captureContinuation` builds a `[MAX_HANDLERS]SavedHandler` buffer on the stack of **every** `call/cc`, and adding the `sticky` field grew it 1024 → 1536 bytes.

Fix: `SavedHandler` is now the *same type* as `ExceptionHandler`, so the live handler stack is passed straight to `allocContinuation` and that per-capture buffer is gone entirely (restore is now a plain `@memcpy`). Since the buffer predates this branch, removing it is a net win:

| | main | branch (before fix) | branch (after fix) |
|-|-:|-:|-:|
| `continuations` | 4.86s | 5.21s | **3.32s** |

≈**1.46x faster than main**, interleaved runs, same machine.

### Review comments

- **README dynamic-wind caveat** — the known-limitations section now documents both caveats (single-shot *and* handler timing, i.e. a handler side effect / `guard` clause running before the guarded body's `dynamic-wind` after-thunk), with a pointer to CONFORMANCE.md and a note that plain `call/cc`/`dynamic-wind`/built-in `guard` are unaffected unless you import `(srfi 248)`.
- **Duplicated native-error→coded-exception conversion** — extracted as `nativeErrorToErrorObject`, now shared by `withExceptionHandlerFn` and `callWithUnwindHandlerFn` so the KEP-0005 error-coding boundary can't drift. Verified `(error-object-code e)` still yields `KP3002` for `(car 5)`.

Full unit suite green; `run-all.sh` **1887 pass, 0 fail** (R7RS 1395/0); SRFI 248 tests green under `-Dgc-stress=true`.

PR #1678, opened 2026-07-19
Implement SRFI 257: pattern matcher with backtracking

### LINKED ISSUE #1644: Implement SRFI 257 (Simple Extendable Pattern Matcher with Backtracking)
## SRFI 257: Simple Extendable Pattern Matcher with Backtracking

**Status:** Final (finalized 2025-12-22) · **Author:** Sergei Egorov
**Spec:** https://srfi.schemers.org/srfi-257/srfi-257.html

### Summary

SRFI 257 defines a `match` construct that offers the core functionality of
SRFI 200, extended with **non-linear patterns** (the same pattern variable may
appear more than once and must match `equal?` values) and **backtracking**.

The design is deliberately modular: nearly every pattern is built from a small
core and is user-extensible through `define-match-pattern`, which hides the
internal macro-expansion protocol and lets new patterns be written as rewrite
rules (much like `syntax-rules`).

Core forms:

- `(match expr [pattern body ...] ...)` — try each clause's pattern against the
  value of `expr`; the first match runs its body. A clause may include a guard
  of the form `(=> next)` or `(=> next back)`: calling `next` skips to the next
  clause, and `back` triggers backtracking to look for an alternative way the
  current pattern could match.
- `(define-match-pattern name clause ...)` — define a new pattern by rewriting,
  in `syntax-rules` style.

Roughly 80 predefined patterns are exported, grouped by category:

- **Predicate patterns:** `~null?`, `~pair?`, `~list?`, `~boolean?`, `~number?`,
  `~integer?`, `~vector?`, `~string?`, `~symbol?`, `~char?`
- **List patterns:** `~cons`, `~list`, `~append`, `~append/ng`, `~append/t`,
  `~etc`, `~etcse`, `~list*`, `~list-no-order`, `~list-no-order*`
- **Vector/string patterns:** `~vector`, `~vector-append`, `~vector-append/ng`,
  `~string`, `~string-append`, `~string-append/ng`
- **Logical patterns:** `~and`, `~or`, `~not`, `~cut!`
- **Core patterns:** `~prop`, `~test`, `~iterate`, `~value`
- **Conversion patterns:** `~vector->list`, `~string->list`, `~list->vector`,
  `~list->string`, `~string->symbol`, `~symbol->string`, `~string->number`,
  `~number->string`
- **Special patterns:** `~=`, `~?`, `~if-id-member`, `~replace-specials`

### Current state

Not implemented. There is no `lib/srfi/257.sld` and no references anywhere in
the tree. `(import (srfi 257))` fails.

### Scope

The reference implementation is a portable, `syntax-rules`-based one (tested on
Chibi, Chez, Guile, and Skint) with no mandatory dependency on other SRFIs, so
this fits our standard **portable `.sld`** model — like the other 67 portable
SRFIs, not a Zig/reader change. The main work is porting the sample
implementation into our library layout and confirming our expander handles its
heavy, recursive `syntax-rules` usage (nested ellipses, hygiene).

Library structure (mirrors the SRFI's optional sublibraries):

- `lib/srfi/257.sld` — main library: `match`, `define-match-pattern`, and the
  full set of predefined patterns above.
- `lib/srfi/257/misc.sld` — `(srfi 257 misc)`: `~etc+`, `~etc=`, `~etc**`,
  `cm-match`, `sr-match`.
- `lib/srfi/257/box.sld` — `(srfi 257 box)`: `~box?`, `~box`. Depends on
  SRFI 111 boxes (we already ship `lib/srfi/111.sld`).
- `lib/srfi/257/rx.sld` — `(srfi 257 rx)`: `~/`, `~/sub`, `~/any`, `~/all`,
  `~/all+`, `~/etc`, `~/etc+`, `~/etcse`, `~/extracted`, `~/split`,
  `~/partitioned`. Depends on a regexp SRFI — SRFI 115 (we ship
  `lib/srfi/115.sld`) or SRFI 264.

(Sublibraries are optional per the SRFI; the main `(srfi 257)` library is the
required deliverable. The `rx` sublibrary in particular can be deferred if
SRFI 115 integration proves fiddly.)

### Acceptance criteria

- [ ] `(import (srfi 257))` provides `match` and `define-match-pattern` plus the
      predefined patterns, with non-linear patterns and `=>`/backtracking guards
      working per spec
- [ ] `define-match-pattern` extension mechanism works (user-defined patterns)
- [ ] Optional sublibraries `(srfi 257 misc)` / `(srfi 257 box)` / `(srfi 257 rx)`
      implemented where their dependencies are available (box → SRFI 111,
      rx → SRFI 115), or explicitly deferred with a note
- [ ] Conformance tests under `tests/scheme/srfi/` (port the SRFI's own examples
      and test suite)
- [ ] Bump the SRFI count/list in `README.md`, `CLAUDE.md`, and `CONFORMANCE.md`
      (the `kaappi features` `portable_srfis` list is auto-scanned from
      `lib/srfi/`)

### Notes / risks

This is a macro-heavy SRFI: `match` expands through a recursive internal
protocol and `define-match-pattern` layers another `syntax-rules`-like rewriter
on top. It's a good stress test for the hygienic expander — watch for nested
ellipsis handling and recursive-hygiene edge cases (cf. the fix in #1344). If a
portable port hits expander limits, capture a minimal reproducer rather than
working around it in the `.sld`.

baijum:
Closes #1644.

Implements SRFI 257 (Simple Extendable Pattern Matcher with Backtracking) as a
portable `.sld` port of Sergei Egorov's reference implementation — the main
`(srfi 257)` library plus the `(srfi 257 misc)` and `(srfi 257 box)`
sublibraries. The optional `(srfi 257 rx)` sublibrary is **not** shipped: it
hard-imports SRFI 264 (Scheme Regular Expressions), which Kaappi doesn't have.

The reference `match` is one of the most demanding `syntax-rules` programs in
existence (Petrofsky extraction, identifier classification through nested
`let-syntax`, a CPS protocol of macro-generating macros). Porting it required
seven general expander/compiler fixes — each with a fail-without/pass-with
regression test in `tests_macros.zig`, none specific to this SRFI:

1. **Cross-frame definition-site locals** — a `let-syntax` transformer defined
   inside a nested lambda couldn't reference an enclosing function's local:
   the reference was hygiene-renamed and the captured-locals slot alias only
   covers the innermost frame. Transformers now record their definition-site
   lexical free references (`def_site_local_refs`), and `renameForHygiene`
   keeps such a name unrenamed when it isn't resolvable in the current frame,
   letting the regular local/upvalue path reach the enclosing frame.
   Same-frame references keep the rename + slot-alias path, which stays
   shadow-proof.

2. **Generated-macro name vs. user variable collision** — the hygienic-capture
   alias injection saw a renamed `let-syntax` keyword (`__hyg_N_k`), stripped
   it to `k`, matched a captured user local of the same name, and injected an
   alias that *shadowed the macro* (`compileForm` suppresses a macro when a
   same-named local resolves), silently rerouting macro calls into the user's
   variable. The injection now skips macro-bound names.

3. **Boxed-slot aliases** — when a later lambda captures a local, the local is
   boxed; injected aliases now copy the slot's current boxing status, and
   `markLocalBoxedBySlot` flips every same-slot local, so aliases read through
   the box instead of yielding the box object.

4. **Quasiquote hygiene** — symbols under `quasiquote` in a template are data,
   like `quote`, but were hygiene-renamed (leaking `__hyg_N_fst` into
   results). Quasiquote now suppresses renaming, with a saturating 2-bit
   nesting depth so a depth-matching `unquote` resumes expression mode.

5. **Renamed tokens vs. syntax-rules literals** — `cm-match` emits template
   tokens (`<...>`, `<_>`) that a helper matcher declares as literals; the
   tokens get hygiene-renamed, so the literal comparison failed. A renamed
   input now matches an *unbound* literal of its base name (the rename exists
   precisely because the identifier had no binding); bound literals keep
   strict binding-identity comparison.

6. **User-text provenance markers** — the core fix. Pattern-var values
   substituted into a *nested* `syntax-rules` template (a generated macro's
   spec) are wrapped in a `__hyg-usertext` marker. When that generated macro
   expands, the marked chunk is instantiated in substitute-don't-rename mode
   (pattern variables and ellipses still work — `define-match-pattern` builds
   the generated macro's rules out of user text), and the wrapper survives
   into further generated specs. Markers are stripped once at the compile
   boundary, skipping `syntax-rules` subtrees. Without this, each expansion
   generation re-walked spliced user text as template material, renaming the
   same identifier under a different scope every time and severing binders
   from references — the root cause of the broken non-linear patterns.

7. **Literal binding resolution across frames** — `literal_bound` recorded a
   literal's definition-site binding with a same-frame-only lookup while the
   use-site check walks the whole lexical chain, so a literal bound in an
   enclosing frame compared def=unbound vs use=bound and never matched. This
   silently disabled the non-linear equality check inside SRFI 257's generated
   backtracking lambdas (`(~append a b a)` matched everything). Both sides now
   resolve identically.

**Reference-suite status:** 111 of 112 assertions pass. The one failure
compares two SRFI 111 boxes with `equal?`, which is implementation-specific
(Kaappi boxes compare by identity); our port of that assertion compares via
`unbox`. Two of the suite's SRFI-241 catamorphism megatowers (reference lines
676/705) still exceed the 10 000-step macro-expansion limit — they blow
through a 100 000-step limit too, so this is a runaway-expansion edge in those
specific towers, not a near-miss; the simpler catamorphism tests (recursive
`len`, `simple-eval`, `,(Cata -> x)` sub-matchers) all pass.

**Tests:**
- `tests/scheme/srfi/srfi257.scm` — 22-assertion smoke suite (~4 s) covering
  every feature family, runs in `run-all.sh`.
- `tests/scheme/srfi/slow/srfi257-full.scm` — the full 78-assertion port
  (~4½ min of macro expansion), kept out of the non-recursive `srfi/*.scm`
  glob; run it when touching the expander or these libraries.
- 8 new regression tests in `src/tests_macros.zig`.

Docs: SRFI counts bumped to 79 (69 portable) in README, CLAUDE.md,
CONFORMANCE.md, and the understanding map; CONFORMANCE.md notes the rx
deferral. (Several SRFI PRs are in flight concurrently — whichever merges
last needs a trivial rebase of these counts. If SRFI 264 (#1672) lands,
the rx sublibrary becomes a small follow-up.)

Validation: full unit suite, `tests/scheme/run-all.sh` (1881 pass / 0 fail),
the R7RS suite, and `tests_macros` under `-Dgc-stress=true` are all green.


baijum:
Addressed all review feedback in 7e7cb6cb (rebased over #1672):

- **`~if-id-member` unbound `yv`** — confirmed as an upstream reference-implementation bug and fixed (`yv` → `xv`); the library header notes the deviation, and the smoke suite now covers non-symbol `sr-match` patterns (`((1 2) 'one-two)`), which hit exactly that branch.
- **SRFI-64 exit plumbing** — both suites now capture the runner before `test-end` and `(exit 1)` on failures, per `tests/scheme/CLAUDE.md`; verified with a deliberate failing assertion (exit 1).
- **Untested misc exports** — added smoke coverage for `~etc+`, `~etc=`, `~etc**`, and the `(f -> x)` cata operator (recursive `sum` via `,(sum -> r)`).
- **Stale CONFORMANCE heading** — updated (it was already stale on main; now reads 70).
- Rebase over SRFI 264 moves the counts to 80 SRFIs / 70 portable. Since SRFI 264 is now in-tree, the deferred `(srfi 257 rx)` sublibrary is unblocked — tracked as a follow-up rather than folded into this PR.

Full reference port re-run after the fix: 78/78. Smoke: 31/31.

baijum:
Second round addressed in 79adb3d7:

- **Cycle hang (critical ×3 sites)** — `stripUsertextMarkers` now uses a tortoise-hare check on the cdr spine (same approach as this file's `countPairs`) plus a depth cap on nested descent, and **descends into vector literals** so a splice inside `#(e)` can't leak a marker into runtime data. Regression tests added for both: a macro invoked with `'#0=(1 . #0#)` and a vector-literal splice.
- **Thread safety** — the per-expansion context (`active_def_local_refs`, `active_use_check`, and the pre-existing `active_custom_ellipsis`/`active_literals` of the same class) is now `threadlocal`.
- **Stale SRFI 264 claim** — CONFORMANCE footnote and the library header now say 264 is available and rx remains a follow-up.

Full unit suite, the 31-assertion smoke, and the SRFI 257 probe battery are green on the new head.

baijum:
Third round addressed in 87aa0103:

- **`~seq-append` type gate (outside-diff, Major)** — confirmed as a second upstream reference bug and fixed: the codegen now checks `(x? xv)` before `(x-length xv)`, so `(~string-append …)` against a non-string fails the pattern instead of raising. Smoke tests cover string-vs-list, vector-vs-string, and `/ng`-vs-number.
- **Expansion-state thread safety (Critical)** — `scope_table`/`scope_table_count` are now `threadlocal`, and `next_scope_id`/`gensym_counter` are bumped atomically (renames must stay process-unique, so the counters remain shared).
- **Depth cutoff (Major)** — raised to 4096 with the soundness argument documented in place: markers are created only at substitution boundaries in the freshly built template skeleton, never inside the (possibly cyclic) user data they wrap, and skeleton nesting is bounded by the expansion-depth/step limits — this matches the expander's existing fixed-limit design (`MAX_MACRO_EXPANSION_DEPTH`, `MAX_BINDINGS`, `MAX_SCOPE_ENTRIES`), and `compileForm` still treats any surviving marker as transparent.
- **GC rooting (Major)** — the marker-wrap, quote, and quasiquote instantiation paths now root intermediate Values across subsequent allocations per the GC guidelines.
- **Vector-splice test (Minor)** — now splices a symbol (fixnums are never wrapped), so the test fails without the vector traversal; stale SRFI 264 note in the slow-suite header fixed.

Local validation on 87aa0103: full unit suite, 34-assertion smoke, `tests_macros` (incl. the new cyclic-literal and vector-marker tests), and the probe battery — all green.

baijum:
Fourth round addressed in 80695ed9:

- **Vector-valued expansions (Major)** — the boundary strip now also runs when the whole expansion is a vector (`isPair or isVector` at all three call sites), with a regression test where the generated macro expands directly to `#(e)`.
- **Marker forgery / cyclic marker chains (Critical, partial)** — `unwrapUsertext` now bounds its chain walk, so forged cyclic data (`#0=(__hyg-usertext . #0#)`) terminates instead of hanging; regression test added. The *unforgeable representation* part I'm deliberately not building in this PR: the marker name lives inside the `__hyg_` namespace the expander has always reserved (`renameForHygiene` treats all `__hyg_*` identifiers as its own), so forged marker data is out-of-contract in exactly the way forged `__hyg_N_x` identifiers already are. Making it a distinct heap type is a fair follow-up once SRFI 258's uninterned symbols land.
- **Quasiquote depth (Major, partial)** — widened to 3 bits (depth 0–7, saturating). Fully unbounded depth would mean threading a separate parameter through every `instantiateTemplate` recursion for towers deeper than 7 — and it turns out to be unobservable anyway: the *runtime* quasiquote evaluator rejects any tower whose unquotes fully unwind at depth ≥ 3 (`` ``(a ,(b ,,x)) `` fails with or without macros — pre-existing, now tracked separately). The new regression test pins macro-vs-direct equivalence at the depths the runtime supports.

Validation on 80695ed9: full unit suite, 34-assertion smoke, and the macro regression file (now 13 SRFI-257-related tests) all green.

baijum:
Fifth round addressed in 13867188:

- **Forged-marker regression test (Minor — good catch)** — the datum now rides through the macro (`(id '#0=(__hyg-usertext . #0#))`). Doing so exposed a *real* residual hang beyond the bounded unwrap: when a forged chain unwraps to itself, the strip walk's re-examine step (and the marker-splice instantiation entry) could loop. Both paths now treat a chain that unwraps to a marker pair as opaque data; the corrected test hangs without the fix and passes with it.
- **Depth saturation at 7 (Major)** — declining the exact-depth rework in this PR. An 8-deep tower whose unquotes fully unwind is already rejected by the *runtime* quasiquote evaluator at depth ≥ 3 (with or without macros — tracked separately as a pre-existing evaluator limitation), so the early re-entry at saturated depth is unobservable end-to-end today. Carrying an exact depth means threading a parameter through every `instantiateTemplate` recursion; that belongs with the evaluator-depth work, where it becomes testable.

Full unit suite and the macro regression file are green on 13867188.

PR #1680, opened 2026-07-20
Implement the (srfi 257 rx) sublibrary

### LINKED ISSUE #1679: Add the (srfi 257 rx) sublibrary (regexp match patterns over SRFI 264/115)
Follow-up to #1644 / PR #1678, which shipped SRFI 257 with the `misc` and `box`
sublibraries but deferred the optional `rx` sublibrary. CONFORMANCE.md and the
library header point here.

The reference `rx.sld` provides the regexp match patterns `~/`, `~/sub`,
`~/any`, `~/all`, `~/all+`, `~/etc`, `~/etc+`, `~/etcse`, `~/extracted`,
`~/split`, `~/partitioned`, and re-exports `rx`. It imports
`(scheme base) (srfi 115) (srfi 257) (srfi 264)` — and SRFI 264 landed in
#1672, so this is now unblocked.

Plan:
- Port `srfi/257/rx.sld` from the reference repo
  (https://github.com/scheme-requests-for-implementation/srfi-257) with the
  usual Kaappi header.
- Port the reference `srfi-257-rx-test.scm` (SRFI 64) as
  `tests/scheme/srfi/srfi257-rx.scm` (or into the slow/ suite if expansion
  cost demands), with the runner-before-test-end exit plumbing.
- Update CONFORMANCE.md (drop the ‡ deferral note), CLAUDE.md's sub-library
  list, and the library header note in `lib/srfi/257.sld`.

The pattern towers in rx are shallow compared to the core matcher, so no new
expander work is expected; if expansion limits do bite, capture a minimal
reproducer per the policy in #1644.

baijum:
Closes #1679.

Ports the optional `rx` sublibrary of SRFI 257 — regexp match patterns over
SRFI 115 regexps and SRFI 264 SSRE strings, so a `match` clause can both test
and destructure a string in the pattern itself:

```scheme
(match "foo:123"
  ((~/ "([a-z]*):([0-9]*)" _ name num) (list name num))
  (_ #f))
;; => ("foo" "123")
```

Exports `~/`, `~/sub`, `~/any`, `~/all`, `~/all+`, `~/etc`, `~/etc+`,
`~/etcse`, `~/extracted`, `~/split`, `~/partitioned`, plus `rx` re-exported
from SRFI 115. Deferred from #1644 because it needs SRFI 264, which landed in
#1672.

## One fix to the reference port

The reference's `~/all+` calls a `regexp-search-all` that neither the
reference nor SRFI 115 defines, so the pattern raised an unbound-variable
error on every use. The specification makes `~/all+` the variant of `~/all`
that fails when the regexp matches no substring, so the missing helper is the
list-of-submatch-lists procedure that reports "no matches" as `'()` rather
than as the circular list of empty lists `~/all` itself needs. It is defined
in `lib/srfi/257/rx.sld` with a header note. The reference suite never
exercises `~/all+` or `~/etc+`, so the new smoke test covers both.

## SRFI 115 fixes the reference suite turned up

Running the reference conformance suite failed 26 of 113 cases — every one of
them an SRFI 115 bug, not an rx one.

The largest is that **repetition never backtracked**. The matcher was
possessive: `(* any)` swallowed everything it could and nothing could hand a
character back, so `(regexp-matches (rx (* any) "b") "ab")` was `#f` and every
pattern with a greedy operator before a literal silently failed to match.
`%run` is now continuation-passing — it offers each way a node can match to a
continuation and takes the first that lets the rest of the pattern through —
which is also what makes the non-greedy operators expressible. Repetition of a
single-character body still scans iteratively (`%run-rep1`), so `(* any)` over
a long string costs no stack; only the general path nests a frame per
iteration.

Also fixed:

| | Was |
|---|---|
| `(?? ...)`, `(*? ...)`, `(**? ...)` | "unknown SRE" — every SSRE non-greedy quantifier (`a*?`, `a+?`, `a{2,4}?`) unusable |
| `(** 1 #f "a")` | type error on the open-ended upper bound |
| `(/ #\0 #\9)` | character arguments silently dropped, leaving a range that matched nothing — the exact form SRFI 264 generates for `[0-9]` |
| `(",;")` | "unknown SRE" (SRFI 115's char-set shorthand) |
| `nwb` | "unknown character class" |
| `(w/nocase (* lower))` on `"abcD"` | `#f` — case folding never reached named classes or ranges |
| `(regexp-split '(* numeric) "abc123def")` | `("" "" "" ...)` — `regexp-fold` handed `kons` the resumed search cursor instead of the previous match end, and split/partition did not skip empty matches |
| `(regexp-partition '(+ numeric) "abc123def456")` | trailing `""` the reference implementation does not produce |

## Validation

- **Reference rx suite: 113/113** (`tests/scheme/srfi/slow/srfi257-rx-full.scm`, verbatim port).
- **chibi's own SRFI 115 corpus** (the reference implementation) run A/B against the pre-change library: **8 cases fixed, 0 regressed**. The 18 that still fail are pre-existing gaps this PR does not touch (look-behind, `grapheme`/`bog`/`eog`, `title`, `w/ascii` semantics, named-submatch lookup) — filed separately.
- `bash tests/scheme/run-all.sh`: 493 files pass, 0 fail; R7RS 1395/1395.
- `zig build test`, `tools/check-srfi-status.sh`: clean.
- `tests/scheme/srfi/srfi115.scm` grew from 68 to 90 assertions covering each fix.

## Test layout and the Debug timeout

The rx suite is split like the SRFI 257 one — a lean smoke test in
`tests/scheme/srfi/` and the full port under `slow/` — because expansion cost
per `match` form climbs steeply with the subpattern count (one five-subpattern
`~/sub` costs ~10s on its own), so no pattern in the smoke test binds more
than two.

The Debug leg's per-file test budget goes 60s → 240s. At 10-20x slower on
allocation-heavy work, 60s was already within seconds of failing on the
existing `srfi257.scm` (48.6s on an M-class laptop); the timeout exists to
catch hangs, not to race the slowest legitimate suite.


baijum:
Follow-up pushed (d77455df): documented the cost profile the CPS rewrite brings with it.

Correct backtracking is not free — quantifiers nested over the same span now backtrack exponentially, the standard behaviour of every backtracking engine (chibi avoids it by compiling to an NFA):

```
(regexp-matches? (regexp '(: (* (* #\a)) #\b)) s)
  14 chars →   64 ms
  18 chars →  1.0 s
  20 chars →  4.1 s
  22 chars → 16.5 s
```

Worth being explicit that this is a **new** characteristic: the old matcher was linear here, because it was possessive — and therefore wrong for ordinary patterns like `(: (* any) "b")`. I'd rather be correct-with-a-documented-exponent than fast-and-wrong, but it is a real trade and reviewers should weigh it.

Mitigations are all bigger than this PR (packrat memoization of `(node, pos)`, complicated by the threaded capture groups; a step budget; or an NFA construction) — recorded on #1681.

What is *not* affected: the single-character-body fast path (`%run-rep1`) keeps `(* any)` linear and stack-free, verified against a 200,000-character input, which is the case that would otherwise have regressed from O(1) to O(n) frames.

CONFORMANCE.md now carries a § note for SRFI 115 covering this plus the SRE forms still unimplemented, so it stops reading as complete coverage.

PR #1683, opened 2026-07-20
Remove the duplicate install script

baijum:
## Why

`install.sh` in this repo was served by nothing — no build step, no workflow, no release artifact referenced it. The only copy anyone runs is `docs/install.sh` in [kaappi.github.io](https://github.com/kaappi/kaappi.github.io), served at <https://kaappi-lang.org/install.sh>, which is also what [README.md](../blob/main/README.md) tells users to curl and what the post-release workflow tests.

Keeping a second copy was not free. The two had drifted, with the *served* copy three hardening commits ahead (`KAAPPI_VERSION`/`KAAPPI_NO_VERIFY`, redirect-based tag discovery that dodges the API rate limit, a safe stdlib swap). So "fix install.sh" in this repo shipped nothing to users — which is exactly how it went unnoticed that the installer never installed `libkaappi_rt.a`, leaving every installed user without a working `kaappi compile`. That fix is kaappi/kaappi.github.io#22.

A stale pointer in docs is a wrong sentence. A stale script is a wrong executable that looks authoritative.

## What changed

- **Deleted `install.sh`.**
- **`docs/dev/porting.md`** — Stage 6 now says where the installer lives and what a new platform needs from it. It never mentioned the installer at all, which is how the `uname`-vs-artifact-name mismatches (NetBSD's kernel port, the BSDs' `amd64`, Linux's `ppc64le`) keep having to be rediscovered per port.
- **`docs/dev/netbsd.md`** — porting-surface table points at the real path.
- **`.github/workflows/post-release.yml`** — asserts the whole chain after each release: archive present → `kaappi doctor` PASS → a compiled binary that runs. Needs a Zig step, because the runner's `cc` is GCC (which cannot consume the IR the backend emits) and `cc_search_order` picks it ahead of the preinstalled clang.
- **CHANGELOG** entry under `[Unreleased]` for both halves.

## Merge order

The new CI step asserts against the **live** installer, so kaappi/kaappi.github.io#22 must merge and deploy before the next release is cut, or the `test-install-script` job fails.


baijum:
Thanks — addressed one, declining the other with reasons.

**Expand verification to all target platforms** — done in 3fd6f519. `test-install-script` is now a matrix over `ubuntu-latest`, `ubuntu-24.04-arm`, and `macos-latest`, with `fail-fast: false`. This one had real merit: the installer's platform detection is exactly where its bugs have historically lived, and macOS resolves the exe path through `_NSGetExecutablePath` + `realpath` rather than `/proc/self/exe` — which is what the new `libkaappi_rt.a` assertions depend on to locate `<exe>/../lib`.

riscv64, s390x and ppc64le are deliberately excluded: no hosted runner executes them natively, and they are interpreter-tier, so there is no runtime archive to assert (the installer skips it there by design).

**Add caching for `zig-cache`** — not applicable here. No step in this workflow runs `zig build`; `grep 'zig build' .github/workflows/post-release.yml` returns nothing. Zig is installed purely to put an LLVM-capable driver first in `cc_search_order` (`zig, cc, clang, gcc`), because the runner's `cc` is GCC and cannot consume the IR `kaappi compile` emits. Nothing produces or reads a `zig-cache`, so a cache step would add a download and an upload to save nothing.

PR #1736, opened 2026-07-24
Add SRFI 57/237/240/137/136/131, exclude 99/100/150 (SRFI Phase 4 slice 3, closes #1695)

### LINKED ISSUE #1695: Implement SRFI record system libraries (57, 99, 100, 131, 136, 137, 150, 237, 240)
## Record system SRFIs

Alternative/extended record systems beyond R7RS `define-record-type`. Most require expander or compiler changes.

| SRFI | Title | Notes |
|------|-------|-------|
| 57 | Records | R6RS-style records |
| 99 | ERR5RS Records | Simplified R6RS records |
| 100 | define-lambda-object | Lambda-based objects |
| 131 | ERR5RS Record Syntax (reduced) | Subset of SRFI 99 |
| 136 | Extensible record types | Adds inheritance |
| 137 | Minimal Unique Types | Opaque wrapper types |
| 150 | Hygienic ERR5RS Record Syntax (reduced) | Hygienic variant of 131 |
| 237 | R6RS Records (refined) | R6RS records for R7RS |
| 240 | Reconciled Records | Unifies R6RS + R7RS records |

### Implementation notes

Many of these are overlapping designs for the same problem. Consider implementing the most modern ones (237, 240) first, then layering compatibility wrappers.

baijum:
## Summary

Fully closes issue #1695 (9 of 9 SRFIs: 6 shipped, 3 excluded).

**Shipped:**
- **SRFI 237** (R6RS Records, refined) — needs real engine changes (a portable macro can't synthesize R6RS's auto-named identifiers), achievable because `define-record-type` is already a Zig-level desugarer here, not a portable macro. `RecordType` gained inheritance/uid/sealed/opaque fields; the desugarer gained a parallel R6RS-clause-syntax path. Inheritance + `protocol` composition (including R6RS's own two-protocol worked example) verified working via extensive live testing.
- **SRFI 240** (Reconciled Records) — thin syntactic-reconciliation layer over 237.
- **SRFI 137** (Minimal Unique Types) — pure portable Scheme over `(srfi 237)`.
- **SRFI 136** (Extensible record types) — portable, including its signature CPS-style introspection macro.
- **SRFI 131** (ERR5RS Record Syntax, reduced) — portable, with by-name (not positional) constructor field resolution and ancestor-field shadowing.
- **SRFI 57** (Records) — portable. Inheritance via "schemes": a named, reusable field-label list a type or another scheme can extend, with multiple schemes mergeable at once (left-to-right append + delete-duplicates), plus `record-update`/`record-update!`/`record-compose`. Deliberately does **not** port the spec's own reference implementation technique (macro-expansion-time field-label comparison via `let-syntax`) — that technique is unusable here (a previously-undiscovered expander bug makes any `let-syntax` upstream of a `define-syntax` in the same expansion chain fail to compile). Sidesteps it entirely by doing all field-list merging/dedup/lookup on ordinary quoted symbols at run time. Scheme conformance is **nominal**, not structural: a type's declared ancestry is tracked in a side table keyed by rtd, populated once per `define-record-type` — a record only conforms to a scheme it was actually declared to extend, not merely one it happens to share field names with. See `lib/srfi/57.sld`'s header for the full design writeup.

**Excluded** (`docs/dev/srfi-exclusions.md`): SRFI 99, 100 (both need identifier synthesis from string concatenation, same fundamental gap as already-excluded 89/206/212), SRFI 150 (needs SRFI 147/148's custom-macro-transformer support, which SRFI 147's own spec text says isn't portably implementable).

**Engine prerequisite discovered mid-implementation:** `define-record-type` was hardcoded as a non-overridable special form in `handleTopLevelForm`, checked before any macro table — so no portable library could give that name new meaning via `define-syntax`, even though `compiler.zig`'s `compileForm` already prioritizes macros over special forms for every non-top-level use (existing precedent: SRFI-219 redefining `define`). Fixed by checking for a user macro literally named `define-record-type` first. Top-level only; library-body use is a documented, un-closed gap.

**Critical fix found during review (severity: process-wide, permanent GC disable):** `vm_records.zig` had, in 8 places (4 pre-existing in the R7RS `handleDefineRecordType`, 4 new in this PR's R6RS path), a `no_collect += 1; errdefer no_collect -= 1;` guard paired with a manual mid-block `no_collect -= 1` followed by *more* fallible operations still inside the `errdefer`'s scope. A late failure (e.g. a compile error while generating a constructor/predicate/accessor) would double-decrement the counter on top of the manual decrement, underflowing a `u32` and permanently disabling garbage collection for the rest of the process. Fixed everywhere by replacing the errdefer + manual-decrement pair with a single unconditional `defer`, matching the pattern `gc_deep_copy.zig` already used correctly.

**Other correctness fixes found during review, applied after re-verifying against the actual spec/R6RS text rather than trusting the original implementation:**
- `record-compose` had import precedence backwards (fixed to first-import-wins, per "left to right, dropping any repeated fields").
- `define-record-type` field accessors/mutators were polymorphic; the spec requires them monomorphic (only *scheme* accessors are polymorphic).
- `record-compose` now validates that each import actually conforms to its declared type/scheme and that every export override label is a real field of the export type (previously field overlap alone was enough, and export labels went unchecked).
- Reject a sealed parent type per R6RS, in both `make-record-type-descriptor` and the R6RS `define-record-type` desugarer.
- Nongenerative-uid reuse (in both the primitive and the R6RS desugarer, which had independently duplicated logic) now checks actual equivalence instead of reusing on uid match alone.
- `allocRecordTypeExtended` rejects field counts that would overflow the `u8` representation instead of letting `@intCast` panic/UB.
- Cross-thread deep-copy of a nongenerative type now keys the destination registry by the copy's own uid string, not the source's (which could be freed independently).
- Added `define-record-scheme`/`define-record-type`'s "no other clauses at all" shorthand forms.

Also documents two narrow, un-root-caused compiler/expander quirks found while building this PR (see `CLAUDE.md`): a `%`-prefixed forward-reference-in-`map` quirk, and the `let-syntax`-upstream-of-`define-syntax` expander bug mentioned above.

## Test plan

- [x] `zig build test` and `zig build test -Dgc-stress=true` green
- [x] `bash tests/scheme/run-all.sh` full regression green (1969/1969)
- [x] 6 SRFI-64 conformance suites, 201 assertions total: `tests/scheme/srfi/{srfi237,srfi240,srfi137,srfi136,srfi131,srfi57}.scm` (SRFI 57's own suite: 51 assertions, including nominal-conformance and record-compose validation regressions added during review)
- [x] Windows (`-Dtarget=x86_64-windows`) and WASM (`zig build wasm`) cross-compiles succeed
- [x] Extensive live-tested edge cases beyond the formal suites, including the boundary conditions for the field-count overflow fix and the nominal-vs-structural conformance distinction


baijum:
Thanks for the thorough review — went through every comment against the actual spec text/code rather than taking either the finding or the original implementation at face value. Two were genuine, serious bugs; several more were real but narrower; one was a false positive; one is documented rather than fixed.

**Fixed as real bugs, confirmed by re-reading SRFI 57 itself:**
- `record-compose` had import precedence backwards — fixed to first-import-wins, matching the spec's "left to right, dropping any repeated fields" (same delete-duplicates convention the spec uses for scheme field-merging). Added a duplicate-import-precedence regression test.
- `define-record-type` field accessors/mutators were polymorphic; the spec requires them monomorphic ("It is an error to pass an accessor a value not of type `<type name>`"). Only scheme accessors are meant to be polymorphic — kept those, added a separate monomorphic path for type accessors with a type check that raises before delegating.

**Fixed, engine side (all confirmed against R6RS text via a targeted search, since I hadn't verified these when writing the original SRFI 237 work):**
- Reject a sealed parent type (`make-record-type-descriptor` and the `define-record-type` R6RS desugarer both) — R6RS: "An exception ... is raised if parent is sealed."
- Nongenerative-uid reuse now checks actual equivalence (parent identity, fields, sealed, opaque) instead of reusing unconditionally on uid match alone.
- `allocRecordTypeExtended` now rejects field counts that would overflow `u8` before ever reaching the `@intCast`, instead of panicking (ReleaseSafe) or wrapping (ReleaseFast) — confirmed reachable via a plain 256-field or parent+own>255 definition, not just a hypothetical.
- Cross-thread deep-copy of a nongenerative record type now reuses the destination VM's own uid registration instead of silently minting a second, non-interoperable type with the same uid.

**Fixed, smaller/docs:**
- Added the `type-clause`-only shorthand forms for `define-record-scheme`/`define-record-type` (a bare name with no other clauses at all — the spec's grammar allows omitting *all* trailing clauses, not just some).
- Added `record-update`/`record-update!` conformance and label-membership checks — at call time rather than the spec's stated expansion time, since `target` is an ordinary value in this design (see below); still catches real misuse.
- Added the explicit `(srfi 237 primitives)` import for defense-in-depth.
- Fixed the CLAUDE.md typo/garbled sentence and the two self-contradicting statements in `docs/dev/srfi-exclusions.md`.

**Verified as a false positive:** the missing `(srfi 237 primitives)` import for `%make-record`. That primitive is registered under `.scheme_base` in `primitives.zig`, not `.srfi_237_primitives` — confirmed by testing the file with the import removed before adding it back anyway (above) purely for defense-in-depth/consistency with this codebase's stated convention, not because it's actually load-bearing here.

**Left as a documented gap, not fixed:** labeled record expressions (`(type-name (label expr)...)`). Supporting this needs `type-name` to be a macro (so `(x 1)` can be read as an unevaluated label/expression pair instead of a call to a procedure named `x`), which conflicts with this library's core design choice — every scheme/type name is bound to an ordinary value specifically so `record-compose`/`record-update` can reference it directly without any macro-time introspection machinery (see the file's header comment for why that machinery was avoided in the first place — it hits a separate, genuine expander bug). The explicit named-field constructor clause already covers the same practical need with a different call-site spelling. A full fix would mean redesigning how every one of these forms references scheme/type names, which is a much larger, riskier change than the PR's actual scope warrants right now.

**Left as-is, not a new regression:** `record_uid_registry`'s lack of `globals_lock` protection under concurrent child-thread access. Verified this mirrors the pre-existing `macros` field exactly (same struct-copy-to-children sharing, same "rarer staleness hazard" already documented and accepted there) — fixing it in isolation for just this one field would be inconsistent with the codebase's existing pattern for the other two identically-shared maps (`macros`, `libraries`), and a proper fix needs a holistic concurrency pass across all three, which is out of scope here.

All fixes verified via the existing 45-assertion SRFI-57 suite (36 original + 9 new) plus `zig build test`, `zig build test -Dgc-stress=true`, the full `bash tests/scheme/run-all.sh` (1969/1969), and Windows/WASM cross-compiles.

Issue #1694, opened 2026-07-20
Implement SRFI numeric vector & array libraries (4, 25, 47, 58, 63, 66, 74, 160, 164, 179, 231)


baijum:
## Numeric vector & array SRFIs

These require new heap types for homogeneous typed vectors/arrays — significant engine work.

| SRFI | Title | Notes |
|------|-------|-------|
| 4 | Homogeneous numeric vector datatypes | Foundation: s8vector, u8vector, etc. |
| 25 | Multi-dimensional Array Primitives | Builds on SRFI 4 |
| 47 | Array | Builds on SRFI 4 and 25 |
| 58 | Array Notation | Reader syntax for arrays (also needs reader change) |
| 63 | Homogeneous and Heterogeneous Arrays | |
| 66 | Octet Vectors | u8 specialization |
| 74 | Octet-Addressed Binary Blocks | Binary data blocks |
| 160 | Homogeneous numeric vector libraries | R7RS-large update of SRFI 4 |
| 164 | Enhanced multi-dimensional Arrays | Builds on SRFI 25/160 |
| 179 | Nonempty Intervals and Generalized Arrays (Updated) | |
| 231 | Intervals and Generalized Arrays | Updated version of 179 |

### Implementation notes

SRFI 4/160 are the foundation — implement those first. Many higher SRFIs layer on top.
May need new `ObjectTag` entries for typed vector types, or a single tag with a type discriminator.

PR #1737, opened 2026-07-24
Add SRFI 160/66/74, fix SRFI 4 f32 precision (#1694 vector family)

baijum:
## Summary

Ships the "vector family" subset of issue #1694 (numeric vector & array
libraries, 11 SRFIs) — the part buildable on one shared native
substrate with well-bounded scope. The "array family" (multi-dimensional
arrays: SRFI 25/47/58/63/164/179/231) is deliberately deferred to a
future slice — see rationale below.

- **SRFI 160** (homogeneous numeric vector libraries) — new engine work:
  one native heap type (`types.NumericVector`, discriminated by an
  11-way `NumericElementKind` enum: s8/u16/s16/u32/s32/u64/s64/f32/f64/c64/c128)
  backs every element kind except u8, which stays a plain R7RS
  bytevector alias per the SRFI's own recommendation. Six generic
  `%`-prefixed Zig primitives (create/predicate/kind/length/ref/set!,
  `primitives_srfi160.zig`) are the entire native surface — every named
  procedure (SRFI-4-shaped core + SRFI 160's much larger SRFI-133-shaped
  extended surface: map/fold/filter/unfold/copy!/append/generator/
  comparator/etc.) is portable Scheme in `lib/srfi/160/base.sld` + one
  `lib/srfi/160/<tag>.sld` per type, written *once* generically and
  shared across all 12 kinds (including u8) via low-level dispatch
  helpers that branch on `bytevector?` vs. `%numeric-vector?`.
- **SRFI 4** is now a thin re-export over `(srfi 160 <tag>)`, which
  fixes a real bug in the old wrapped-vector implementation:
  `f32vector` never actually truncated to 32-bit precision (it stored
  full f64 precision internally). Same public API, so nothing importing
  `(srfi 4)` needs to change — verified against the existing test suite
  unmodified (all 71 pre-existing tests pass) plus 2 new tests for the
  f32 fix.
- **SRFI 66** (octet vectors) and **SRFI 74** (octet-addressed binary
  blocks / "blobs") are both u8vector/bytevector aliases needing no
  heap-type work — pure portable Scheme over existing bytevector
  primitives. The one exception: `(endianness native)` needed a tiny
  new primitive, `%host-big-endian?` (`primitives_r7rs.zig`, wrapping
  `builtin.cpu.arch.endian()`) — a portable implementation has no other
  way to learn real hardware byte order, and SRFI 74's native-tagged
  accessors exist specifically for interop with externally-produced
  binary data, so assuming little-endian would be silently wrong on
  kaappi's own s390x/ppc64le CI targets.
- Reader/writer literal syntax (`#s8(...)`, `#f64(...)`, etc.) is
  deliberately skipped: SRFI 160's own text states "conformance to this
  SRFI does not in itself require support for these external
  representations." Numeric vectors still get a `printer.zig` display
  arm (`#<s8vector 1 2 3>`), just no reader arm.

**Deferred, not dropped:** the array family (25, 47, 58, 63, 164, 179,
231) spans three mutually incompatible API lineages per the SRFIs' own
text (25 vs. 47/63's conflicting `array-set!`/`make-array` shapes; 164
explicitly rejects the interval abstraction 179/231 build on) plus a
reader-syntax SRFI (58) — SRFI 231 alone is ~100 bindings. Tracked
separately under #1694; documented in `CLAUDE.md`.

## Test plan

- [x] `zig build test` and `zig build test -Dgc-stress=true` green (new
      heap type touches `memory.zig`/`gc_collect.zig`)
- [x] New SRFI-64 suites: `srfi160.scm` (101 assertions across all 12
      types — core ops, range-check rejection, full extended surface,
      f32 truncation, c64/c128 round-trip, generators, comparators),
      `srfi66.scm` (16, including overlapping `u8vector-copy!`),
      `srfi74.scm` (30, including big/little/native round-trips and
      arbitrary-size `blob-uint-ref`/`-set!`)
- [x] Existing `srfi4.scm` passes unmodified (71 tests) + 2 new f32
      regression tests (73 total)
- [x] `bash tests/scheme/run-all.sh` full regression: 1972 pass, 0 fail
      (577 scheme files + 1395 R7RS suite)
- [x] Cross-compile sanity: `-Dtarget=x86_64-windows`, `zig build wasm`,
      `-Dtarget=s390x-linux` (big-endian canary, relevant given the new
      byte-order-sensitive code in SRFI 74/160)
      (164 → 167; note SRFI 160, like 226, is sub-library-only so it
      doesn't appear as a bare number in `kaappi features`' scan) and
      per-SRFI tables; `CLAUDE.md` notes the array-family remainder as
      an open follow-up


baijum:
Addressed CodeRabbit's review (6 actionable comments) in 8cce6a06:

**Fixed:**
1. `%make-numeric-vector`/`allocNumericVectorFill` — a user-supplied length near the fixnum range would `@intCast`-panic instead of raising a catchable error on wasm32 (`usize` = 32-bit there). Added an overflow-safe check before narrowing, plus the same defense-in-depth check in the allocator, mirroring `allocVectorFill`'s existing pattern.
2. Complex sign printing for c64/c128 numeric-vector elements — `+inf`/`-inf`/`+nan` imaginary parts printed with a doubled sign (`++inf.0i`, `-+inf.0i`) since `formatFlonum` already emits its own sign for these; `-0.0`'s sign was also lost (`<` is false for negative zero). Factored the already-correct handling from the standalone `Complex` printer arm into a shared `writeImaginaryPart` helper used by both.
3. Three `srfi160.scm` tests that couldn't actually distinguish correct behavior from a wrong-direction bug (commutative `fold-right` combiner, an `unfold-right`/`unfold!` callback that only used the index, never the seed) — rewritten to be order/state-sensitive.

**Not changed, with reasoning:**
- The suggestion to relocate the SRFI documentation additions out of `CLAUDE.md`/`CONFORMANCE.md` into `docs/dev/`/`kaappi.github.io` — this contradicts the repo's own established convention (every prior SRFI-adding PR documents SRFI coverage in these exact files in this repo; `kaappi.github.io` is end-user docs only per root `CLAUDE.md`, and `docs/dev/` doesn't track per-SRFI conformance).
- The file-size-split suggestion for `types.zig`/`memory.zig` — both were already ~150-300 lines over the 1500-line cap *before* this PR (accumulated across many prior SRFI additions), so this flags a pre-existing condition, not something introduced here. Splitting either is a large, separate refactor disproportionate to one SRFI PR.

Full regression suite (1972 tests) and unit tests (including `-Dgc-stress=true`) still green after these fixes.

PR #1739, opened 2026-07-24
Add SRFI 25, multi-dimensional array primitives (#1694 array family)

baijum:
## Summary

First piece of issue #1694's array family (multi-dimensional array
libraries: SRFI 25, 47, 58, 63, 164, 179, 231). The vector-family half
(SRFI 4/160/66/74) shipped in #1737.

Five parallel research passes over the actual spec text corrected the
original "three mutually incompatible lineages" framing in #1694:

- **SRFI 25** (this PR, 10 procedures) — heterogeneous arrays, arbitrary
  per-dimension lower bounds via a `shape` bounds object, affine
  `share-array` views. Has an actual portable reference implementation to
  check against.
- **SRFI 164** (23 procedures: the same 10 + 13 new) is a documented,
  *compatible extension* of SRFI 25 — identical shape representation,
  `share-array` copied verbatim — not a separate lineage. A cheap follow-on
  once 25 exists.
- **SRFI 47/63** (63 supersedes 47; 13 procedures) is the real incompatible
  lineage, confirmed via both SRFIs' own Issues sections: `array-set!`'s
  new-value argument is *last* in 25/164 but *second* (right after `array`)
  in 47/63 — a silent-data-corruption-class conflict, not just argument
  order — and `make-array`'s first argument is a bounds-shape object in
  25/164 vs. a type/fill-value prototype in 47/63.
- **SRFI 231** (supersedes 179; 118 bindings) is a third, much larger,
  textually unrelated redesign — a three-tier array/mutable-array/
  specialized-array model with a pluggable storage-class abstraction.
- **SRFI 58** (array literal reader/writer syntax) is written specifically
  against 47/63's type-tag naming, not reusable as-is against 25/164/231.

This PR implements **SRFI 25 only** — small, self-contained, and now makes
164 a well-scoped follow-on rather than something to redesign from scratch.
47/63, 164, 231, and 58 remain open under #1694, each with the reason above
so the issue doesn't read as abandoned.

## Design

Arrays are spec-defined as heterogeneous (arbitrary Scheme objects) with
*no* specified internal representation and *no* relationship to SRFI
4/160's numeric vectors — so `lib/srfi/25.sld` is pure portable Scheme, no
engine work. One `define-record-type` covers both:
- **simple** arrays (`make-array`/`array`): a flat, row-major backing
  vector.
- **shared** arrays (`share-array`): a base array + an affine
  index-translation `mapper`, no backing vector of its own. `array-ref`/
  `array-set!` translate the index through `mapper` and recurse into the
  *base array's own* ref/set! (not its raw vector) — so arbitrarily nested
  views compose correctly for both reads and writes, at the cost of one
  extra call per nesting level. The spec's rationale for the affine
  restriction is that composed affine maps *could* collapse to O(1) — a
  documented optimization opportunity, not a conformance requirement, so
  recursive delegation (simpler, fully conformant) was the deliberate
  choice over that optimization.

The one point genuinely ambiguous in the spec prose alone — whether
`share-array`'s `mapper` takes/returns packed or separate indices — was
resolved by fetching the spec's own worked example (a diagonal-view
identity-matrix construction, `(lambda (k) (values k k))`): `mapper` takes
the new array's indices as separate variadic arguments and returns the base
array's coordinates via multiple values, mirroring `array-ref`/
`array-set!`'s own convention. That exact example is now a regression test
and produces the expected 4×4 identity matrix.

A shape is a vector of `(lower . upper)` pairs. It's deliberately **not**
made to satisfy `array?` itself — the spec's own "a shape is a d×2 array"
framing is Rationale prose never operationally exercised by any of the 10
mandated procedures (no `shape?`, no way to recover a shape back from an
array) — a documented, low-risk scope reduction.

Error handling: the spec is silent on out-of-range/wrong-arity behavior.
Following this codebase's established convention (every other SRFI raises
a catchable error for misuse), bad indices, wrong arity, malformed shapes,
and mismatched `array` literal value counts all raise via `error`.

## Test plan

- [x] New `tests/scheme/srfi/srfi25.scm` (40 assertions): construction with
      and without fill, `array?` disjointness from vectors/strings/lists,
      `array-rank`/`array-start`/`array-end`, `array-ref`/`array-set!` with
      variadic indices and with both packed-index forms (vector and 1-D
      array), arbitrary non-zero lower bounds (SRFI 25's defining
      difference from 47/63), the spec's own `i_4` diagonal-view example
      verbatim, a nested share-of-a-shared-array exercising recursive
      delegation for both reads and a write-through, `array-set!`'s
      value-last convention, and catchable errors for bad indices/shapes.
- [x] `zig build test` green (no engine changes — this PR touches no `src/`
      files at all).
- [x] `bash tests/scheme/run-all.sh` full regression: 1973 pass, 0 fail
      (578 scheme files + 1395 R7RS suite).
      (167 → 168; portable 152 → 153) and per-SRFI tables; `CLAUDE.md`'s
      SRFI-count paragraph rewritten to reflect the corrected lineage
      picture above instead of the original three-incompatible-lineages
      assumption.


baijum:
Addressed CodeRabbit's review (7 actionable comments) in 0c98878b — all 7 were legitimate, all fixed:

1. **Shape aliasing (the important one)**: `make-array`/`array`/`share-array` stored the caller's `shape` vector by reference instead of copying it. Per spec, "an array does not retain a dependence to the shape array" — since `shape` returns an ordinary, caller-visible mutable vector of pairs, a caller mutating it after construction could retroactively corrupt an already-built array's bounds while its backing store keeps the size computed at the original construction time. Fixed with a deep-copy (`%copy-shape`) at all three construction points.
2. `make-array` silently discarded extra fill-value arguments beyond the first instead of rejecting them — now raises.
3. The packed-index-array form of `array-ref`/`array-set!` didn't check that the index array is 0-based, as the spec requires ("a 0-based 1-dimensional array") — now validated.
4. Markdown lint: a line in `CLAUDE.md` started with `#1694`, which linters parse as an attempted ATX heading — reworded.
5. Test hygiene: wrapped the `share-array` fixtures (`i_4`, `base`/`row-view`) in `let`/`let*` instead of top-level `define`s.
6. + 7. Added regression tests for the excess-fill-value rejection and the non-zero-based packed-index-array rejection.

Full regression suite (1973 tests) and unit tests still green after these fixes. (One `compile-import-error-703.sh` flake appeared on an initial full-suite run, unrelated to this PR — reproduced as passing standalone and on a clean full rerun; that test touches native-compile import-error handling, nothing this PR's diff comes near.)

PR #1740, opened 2026-07-25
Add SRFI 164, enhanced multi-dimensional arrays (#1694 array family)

baijum:
## Summary

Second piece of issue #1694's array family, building directly on SRFI 25
(#1739). SRFI 164 is a documented, compatible **extension** of SRFI 25 —
identical shape representation, `share-array` copied from 25's own spec
text verbatim, the same 10 procedures plus 13 new ones (23 total):
`->shape`, `array-shape`, `array-size`, `build-array`, `index-array`,
`array-index-ref`, `array-index-share`, `array-copy!`, `array-fill!`,
`array-transform`, `array-reshape`, `array-flatten`, `array->vector`.

Implemented as its own independent library (`lib/srfi/164.sld`) rather
than importing `(srfi 25)` — `define-record-type` field accessors don't
cross library boundaries, and this SRFI's array needs a third mode SRFI
25's record has no room for. Unlike SRFI 25, this SRFI has no portable
reference implementation to check against (its own Implementation section
is Kawa/Java-specific pseudocode), so the design below is derived directly
from the spec's prose and worked examples.

Still open under #1694, each with its own reason: SRFI 47/63 (63
supersedes 47) is a lineage confirmed incompatible with 25/164 by both
SRFIs' own Issues sections (`array-set!`'s value-last vs. value-second
argument order is a silent-data-corruption-class conflict); SRFI 231
(supersedes 179) is a third, much larger, textually unrelated redesign
(118 bindings, a three-tier storage-class-based model); SRFI 58's reader
syntax is written specifically against 47/63's type-tag naming, not
reusable as-is against 25/164/231.

## Design

One record, three modes (no new heap type — pure portable Scheme):
- **simple**: own row-major backing vector, exactly SRFI 25's mode.
- **shared/view**: base array + an index-translation `mapper`, same
  recursive-delegation design as SRFI 25's `share-array` (translate
  through `mapper`, recurse into the *base array's own* ref/set!, not its
  raw vector — composes correctly for arbitrarily nested views). Backs
  `share-array`, `array-transform`, `array-reshape`, and
  `array-index-share`.
- **virtual**: a `getter`/optional-`setter` pair, no backing storage at
  all. Backs `build-array`, `index-array`, and `array-index-ref`'s
  non-scalar case.

A few things worth calling out:
- `array-transform`'s `transform` uses a *different* calling convention
  than `share-array`'s `mapper`, per the spec's own explicit contrast (one
  vector argument in/out, vs. separate variadic arguments and multiple
  values). Adapted with a one-line wrapper into the shared/view mode's own
  convention — which already works for a non-affine `transform`, since
  this codebase's `share-array` never checked or exploited affineness in
  the first place (the spec states the affine requirement as an
  optimization *opportunity*, not a correctness requirement).
- `array-index-ref`/`array-index-share` (APL-style generalized indexing,
  where each per-dimension index is an integer or an array) share one
  resolver computing the concatenated result shape and an
  index-splitting closure.
- `array-reshape` aliases the source's backing vector directly when the
  source is simple, per spec ("uses the same underlying vector as
  array"), and falls back to a row-major-rank-based recompute otherwise.
- `array->vector` returns the live backing vector for a simple array
  (true zero-copy identity) but only a disconnected snapshot for a
  non-simple one — the spec calls for a live view there too, which isn't
  achievable with a literal R7RS vector in portable Scheme (no hooks,
  unlike this codebase's custom ports). A deliberate, documented scope
  reduction.

## Two self-caught bugs worth knowing about

1. A deeply-nested first draft of the index-resolver helper mis-balanced
   its own closing parens badly enough to silently close the whole
   `define-library` form early. `kaappi check lib/srfi/164.sld` caught it
   immediately with an exact line number — worth running on any new
   portable library before ever executing it. Fixed by flattening into
   separately-named top-level helpers instead of nested internal
   defines/lets, which are much easier to verify by eye.
2. Once that was fixed, the resolver closure was defined with a rest
   parameter (`(lambda new-indices ...)`) but every call site passed an
   *already-built* list as a single argument (`(resolver new-indices)`),
   silently wrapping it one list-level too deep. This produced correctly-
   *shaped* but wrong-looking "index out of range" errors far from the
   actual mistake. Fixed by making the resolver take one fixed parameter
   instead of a rest parameter, since no caller was ever splatting
   separate arguments into it.

## Test plan

- [x] New `tests/scheme/srfi/srfi164.scm` (55 assertions): spot checks of
      the 10 inherited procedures, `make-array`'s new cycling-multiple-
      fill-values behavior, `->shape` across all three input forms,
      `array-shape` round-tripping off a constructed array, `array-size`,
      `build-array` with and without a setter (including the immutable-
      array error case), `index-array`, `array-index-ref` for both the
      scalar and generalized cases, `array-copy!`/`array-fill!` (including
      the mismatched-shape error case), `array-transform` with a
      genuinely non-affine transform, `array-index-share`'s mutable-view
      write-through (including its rank-0 scalar-view case),
      `array-reshape` on both a simple array (asserting the same backing
      vector is aliased) and a non-simple one, `array-flatten` (always a
      copy, verified by mutating the source afterward), and
      `array->vector`'s live-identity vs. documented-snapshot behavior.
- [x] `zig build test` green (no engine changes — this PR touches no
      `src/` files at all).
- [x] `bash tests/scheme/run-all.sh` full regression: 1974 pass, 0 fail
      (579 scheme files + 1395 R7RS suite).
      (168 → 169; portable 153 → 154) and per-SRFI tables; `CLAUDE.md`'s
      SRFI-count paragraph updated to reflect 25 and 164 as shipped.


baijum:
Addressed CodeRabbit's review (6 actionable comments) in 134afb3d — 5 fixed, 1 documented rather than implemented:

**Fixed:**
1. **Shape-specifier coercion (the important one)** — every shape-taking procedure (`make-array`, `array`, `share-array`, `build-array`, `index-array`, `array-transform`, `array-reshape`) now runs its `shape` argument through `->shape` before use. Verified against the primary spec text (fetched directly, since CodeRabbit's own web search on this point was initially unreliable): "the procedures in this specification that require a shape can accept a shape-specifier, as if converted by the procedure `->shape`" — including the spec's own worked example, `(make-array #(2 4) 1 2 3 4 5)`. Previously only `->shape` itself did this conversion, so passing a raw specifier directly to any other procedure raised a `car`/`cdr` type error. `->shape`'s output is already a fresh, non-aliased vector, so this also covers the "no dependence on the shape argument" rule without a separate copy step.
2. `->shape` now validates `lo <= hi` and exactness, matching `shape`'s own validation (previously a malformed specifier surfaced later as an obscure `make-vector`/`vector-ref` error).
3. `build-array`'s getter/setter now match the spec's actual contract (verified directly against primary source: getter takes **one index-vector** argument; setter takes **index-vector then value**) via a boundary adapter — the library's internal variadic-indices/value-first convention had leaked into the public API. Updated `array-index-ref`'s own use of `build-array` to match.
4. Added regression tests for the `->shape`-coercion fix (one per affected procedure), `->shape`'s bound validation, and packed-index argument handling (vector and rank-1 array forms, plus all three rejection cases), which had no coverage at all.
5. Fixed a stale test comment that described `array-reshape`'s non-simple-source result as "recomputes, doesn't alias" when the assertions right below it actually demonstrate a live view (mutating the source is visible through the reshaped result) — just not a *vector* alias, which is the real distinction.

**Documented, not implemented:** the spec's language ("every gvector is an array", "a library-based implementation... *should* use the existing vector type" for simple rank-1 zero-lower-bound arrays) does suggest `array?` could also be true of a plain R7RS vector. This is a **should**, not a **must** — implementing it properly would need a parallel plain-vector code path threaded through every `array-*` procedure (`array-rank`, `array-start`, `array-ref`, `array-set!`, etc.), a much larger change than this PR's scope. Added an explicit header comment documenting this as a deliberate reduction rather than an oversight.

Full regression suite (1974 tests) and unit tests still green after these fixes. (Saw the same `compile-import-error-703.sh` flake as PR #1739 on one run — unrelated subsystem, reproduced as passing standalone and on a clean full rerun.)

Issue #1699, opened 2026-07-20
Implement SRFI macro & syntax extension libraries (72, 139, 147, 148, 149, 211, 213)


baijum:
## Macro & syntax system SRFIs

These require changes to the expander, compiler, or both.

| SRFI | Title | Notes |
|------|-------|-------|
| 72 | Hygienic macros | Explicit renaming macros |
| 139 | Syntax parameters | Needs expander support |
| 147 | Custom macro transformers | Needs expander hooks |
| 148 | Eager syntax-rules | Compile-time syntax-rules evaluation |
| 149 | Basic Syntax-rules Template Extensions | Extends syntax-rules templates |
| 211 | Scheme Macro Libraries | Meta-library for macro systems |
| 213 | Identifier Properties | Compile-time identifier metadata |

baijum:
SRFI 148 (eager syntax-rules) shipped in #1806 (`ff6cffd5`) — the fourth and final portable piece of this group. 139 (#1757), 149, 147 (#1760/#1762/#1763), and now 148 are all on main; the library is pure portable Scheme over 147's begin-wrapped transformer mechanism, its verbatim reference suite runs 134/8-xfail in ~0.9s (the xfails cite the general engine bugs #1800/#1801), and the import costs ~0.07s after #1802/#1804 fixed the compile-cost cliff. Remaining tracked here: 72, 211, 213 — all needing expander/compiler mechanism work (72's explicit-renaming macros being the prerequisite-shaped one).

Issue #1703, opened 2026-07-20
Implement SRFI system, environment & hash table libraries (59, 88, 89, 90, 106, 112, 123, 126, 173, 193)


baijum:
## System, environment & hash table SRFIs

Mixed difficulty. Some are portable wrappers, others need new value types or runtime support.

| SRFI | Title | Difficulty | Notes |
|------|-------|-----------|-------|
| 59 | Vicinity | portable | File path utilities |
| 88 | Keyword objects | **engine-change** | New value type (\`foo:\` keyword syntax) |
| 89 | Optional positional and named parameters | portable | Syntax for optional/keyword args |
| 90 | Extensible hash table constructor | portable | Builds on SRFI 69 |
| 106 | Basic socket interface | **engine-change** | Socket I/O primitives |
| 112 | Environment Inquiry | portable | \`(implementation-name)\`, etc. |
| 123 | Generic accessor and modifier operators | portable | Generic \`ref\` / \`set!\` |
| 126 | R6RS-based hashtables | portable | R6RS hash table API over SRFI 69 |
| 173 | Hooks | portable | Observer/callback pattern |
| 193 | Command line | portable | Argument parsing |

PR #1747, opened 2026-07-25
Add SRFI 63, homogeneous and heterogeneous arrays (#1694 array family)

baijum:
## Summary

Third piece of issue #1694's array family, following SRFI 25 (#1739) and SRFI 164 (#1740).

- **Implements SRFI 63** (`lib/srfi/63.sld`): `array?`, `array-rank`, `array-dimensions`, `make-array`, `make-shared-array`, `array-in-bounds?`, `array-ref`, `array-set!`, `equal?` (array-augmented), `list->array`, `array->list`, `vector->array`, `array->vector`, plus 20 prototype-generator procedures (`A:fixZ8b`, `A:floC128b`, etc.).
- **Excludes SRFI 47** (`docs/dev/srfi-exclusions.md`): its own page states it's superseded by SRFI 63; every one of its 9 procedures has an identical signature in 63.
- Confirmed **incompatible** with SRFI 25/164 (already shipped): `array-set!`'s value argument is *second* here, not last; `make-array`'s first argument is a type/fill-value *prototype*, not a bounds-shape object; every dimension is a plain zero-based size.
- 12 of the 20 element kinds reuse the already-shipped `(srfi 160 <tag>)` procedures as backing storage (getting range/exactness validation for free); the remaining 8 (16/128-bit floats, 16/32-bit complex, decimal floats) fall back to a plain, unchecked vector — a fallback the spec itself sanctions.
- `make-shared-array`'s mapper returns a **list** of coordinates (confirmed via the spec's own diagonal/offset examples), unlike SRFI 25/164's multiple-values convention — the one genuinely easy-to-get-backwards detail in this SRFI.
- `list->array`/`array->list` use a **rank-nested** list structure; `vector->array`/`array->vector` use a **flat** vector with explicit dimensions — two different algorithms, confirmed via the spec's own examples rather than assumed.

Docs: 170 SRFIs now supported (169 → 170, 154 → 155 portable), 9 tracked for future work (down from 11: 63 ships, 47 excluded), 29 excluded (up from 28).

Still open under #1694: SRFI 231 (a much larger, unrelated redesign) and SRFI 58 (reader syntax, already excluded pending re-examination now that typed-array infrastructure exists). Issue #1699 (macro/syntax extensions) remains untouched.

## Test plan

- [x] `kaappi check lib/srfi/63.sld` — clean
- [x] Manual smoke test covering all prototype kinds, value-second `array-set!`, rank-changing shared-array views with write-through, array-augmented `equal?`, and all 4 conversion procedures
- [x] New `tests/scheme/srfi/srfi63.scm` (65 assertions) — all pass
- [x] `bash tests/scheme/run-all.sh` — 1975 pass, 0 fail


baijum:
Addressed all 6 findings — verified each against the primary SRFI 63/47 spec text (srfi.schemers.org) before fixing, per this project's convention:

1. **Markdown lint (`#1703` at line start)** — fixed, reworded.
2. **Vectors/strings must satisfy `array?`** — confirmed directly from the spec ("Arrays are not disjoint from other Scheme types"; vectors/strings are rank-1 arrays in their own right, not just valid prototypes). Fixed: `array?`, `array-rank`, `array-dimensions`, `array-in-bounds?`, `array-ref`, `array-set!` now dispatch across `<uarray>` records, plain vectors, and plain strings.
3. **Shared-view bounds checking** — real bug, confirmed by reading the code: `array-in-bounds?` checked the view's own declared dims but `array-ref`/`array-set!` skipped straight to the mapper, so a lenient mapper could let an out-of-range view index through. Fixed: both now validate against the view's own dims first.
4. **Prototype fill propagation** — confirmed from the spec ("the new array will be filled with the element at the origin of prototype"). Fixed both ends: the prototype-generator procedures (`A:fixZ8b` etc.) now actually store a given fill value instead of building a zero-length store that could never hold it, and `make-array` now reads it back instead of always using the kind's default.
5. **`list->array` invalid rank** — confirmed as a real infinite-recursion risk (negative rank never reaches the `= r 0` base case). Fixed: rejects a non-nonnegative-integer rank up front.

`equal?`'s array-comparison branch stays scoped to genuine `<uarray>` records (documented in the code) since plain R7RS `equal?` already handles vector/string comparison correctly and extending it to cross-representation array equality wasn't part of any finding.

31 new assertions added (65 → 96), plus a full `zig build test` + `bash tests/scheme/run-all.sh` rerun (1975 pass, 0 fail).

Issue #1748, opened 2026-07-25
tests/scheme/compile/compile-import-error-703.sh flakes intermittently in full run-all.sh runs


baijum:
## Summary

`tests/scheme/compile/compile-import-error-703.sh` has failed 3 times across recent SRFI Phase 4 slices when run as part of a full `bash tests/scheme/run-all.sh` pass (PRs #1739, #1740, and most recently the SRFI 231 intervals slice), but has passed every time it's been rerun standalone (`bash tests/scheme/compile/compile-import-error-703.sh`) immediately after. A clean full rerun of `run-all.sh` afterward has also passed cleanly each time. The test itself is unrelated in subject matter to any of those PRs' actual diffs (it's a native-compiler regression test for #703, none of those PRs touched compiler/native-compile code), so this looks like a genuine, pre-existing test-infrastructure flake rather than a real regression.

## Root cause theory (not yet confirmed)

This test does something no other `run-all.sh`-driven test does: it **rebuilds the shared `zig-out/bin/kaappi` binary in place**, twice, as part of its own normal (non-error-path) operation:

```bash
# Build bundled binary
(cd "$REPO_DIR" && zig build -Dbundle="$DIR/main.sbc" -Doptimize=ReleaseSafe 2>/dev/null)
cp "$REPO_DIR/zig-out/bin/kaappi" "$BUNDLE_BIN"

# Rebuild regular binary
(cd "$REPO_DIR" && zig build 2>/dev/null)
```

`run-all.sh` sets `KAAPPI=zig-out/bin/kaappi` once at the top and every subsequent test in the same run reads that same fixed path. `run_file`/`run_suite` execute test files sequentially (confirmed by reading `run-all.sh` — the per-file backgrounding in `run_file` is only for the timeout mechanism, not cross-file concurrency), so there's no literal process-level race between two test files. But if the final `zig build` above returns to the shell before the OS has fully flushed/settled the freshly-linked binary (plausible depending on filesystem/OS buffering, especially right after a link step), the *next* sequential test to `exec`/`fork` `zig-out/bin/kaappi` could observe a truncated or momentarily-unavailable binary — which would explain a flake that:
- never reproduces when this one file is run standalone (nothing "next" races it),
- isn't tied to any particular PR's diff (it's a scheduling/timing hazard in the test's own design, not a logic bug triggered by specific source changes),
- has now recurred 3 times.

[...]

PR #1749, opened 2026-07-25
Add SRFI 231 intervals + misc helpers (#1694 array family, phase 1a)

baijum:
## Summary

First of several slices implementing SRFI 231 ("Intervals and Generalized Arrays") — the last remaining piece of issue #1694's array family, following SRFI 25 (#1739), SRFI 164 (#1740), and SRFI 63 (#1747).

**Scale note**: SRFI 231 is 118 bindings total (101 procedures/parameters + 17 storage-class singletons) — roughly 5-10x larger than any prior slice in this phase (SRFI 164 was the largest so far, at 23). Four parallel research passes against the primary spec text confirmed it's also structurally unrelated to SRFI 25/164/63: intervals are a genuinely new opaque type (two parallel vectors of exact-integer lower/upper bounds, arbitrary sign — not a shape-object or plain-size model), and `array?`/`array-set!` (later phases) turn out to be their own hybrid of the 25/164 and 63 conventions.

Given the scale, this is being split into several slices:

| Phase | Scope | ~Bindings |
|-------|-------|-----------|
| **1a (this PR)** | Intervals + misc permutation/translation helpers | 32 |
| 1b | Storage classes (`make-storage-class`, 17 singletons, accessors) | 28 |
| 2 | Core array object (array?/mutable-array?/specialized-array?, make-array, make-specialized-array, array-ref/set!, etc.) | ~19 |
| 3 | Views/sharing/reshaping (specialized-array-share, array-curry, array-tile, etc.) | ~12 |
| 4 | Bulk combinators + conversions (array-map/fold/reduce, outer/inner-product, list/vector conversions) | ~16 |
| 5 | Multi-array assembly (array-stack, array-append, array-block, array-decurry) | ~9 |
| final | Merge into public `lib/srfi/231.sld`, docs, SRFI count bump | — |

`(srfi 231)` is **not yet importable** as a bare library — same reasoning as `(srfi 160 base)` + per-tag files having no bare `(srfi 160)`. This PR does not touch the tracked SRFI counts (231 stays "tracked, in progress" under #1694 until the whole thing ships).

## This PR's scope

- `lib/srfi/231/misc.sld` (6 procedures): `translation?`, `permutation?`, `index-rotate`, `index-first`, `index-last`, `index-swap`.
- `lib/srfi/231/intervals.sld` (26 procedures): the `<interval>` record and all interval procedures — construction (both `make-interval` forms), accessors, `interval-for-each`/`fold-left`/`fold-right` (general d-dimensional lexicographic traversal), `interval-dilate`/`translate`/`permute`/`scale`, `interval-intersect` (returns `#f` on no overlap, not an error), `interval-cartesian-product`, `interval-projections` (returns two values).

Every calling convention was confirmed against the primary spec text and reference implementation before writing code, including two procedures (`translation?`/`permutation?`) whose exact validation rule the spec states only in prose, and `interval-fold-right`'s "all f evaluations before any operator application" requirement (a naive reverse-order fold interleaving f and operator would violate this even though it looks equivalent when f is pure).

## Also filed

#1748: `tests/scheme/compile/compile-import-error-703.sh` has now flaked 3 times in full `run-all.sh` runs across recent slices (never standalone), with a concrete root-cause theory (it rebuilds the shared `zig-out/bin/kaappi` binary in place, twice, as part of its own normal operation). Unrelated to this PR's diff — passed both standalone and on a clean full rerun here.

## Test plan

- [x] `kaappi check lib/srfi/231/misc.sld lib/srfi/231/intervals.sld` — clean
- [x] Manual smoke test covering both `make-interval` forms, negative bounds, 0-dim/empty edge cases, fold-left/fold-right order sensitivity (verified against the spec's own 0-dimensional formulas), dilate/translate/permute/scale, intersect, cartesian-product, and projections (matches the spec's own worked example verbatim)
- [x] New `tests/scheme/srfi/srfi231-intervals.scm` (62 assertions) — all pass
- [x] `bash tests/scheme/run-all.sh` — 1975 pass, 1 known flake (#1748, unrelated, confirmed passing standalone + on clean rerun)

### DIFF
--- a/lib/srfi/231/intervals.sld
+++ b/lib/srfi/231/intervals.sld
@@ -1,0 +1,249 @@
+;;; SRFI 231 -- Intervals (phase 1a of #1694's SRFI 231 slice).
+;;;
+;;; SRFI 231 ("Intervals and Generalized Arrays") is 118 bindings total (101
+;;; procedures/parameters + 17 storage-class singletons) -- roughly 5-10x
+;;; any prior slice in this project, and structurally unrelated to SRFI
+;;; 25/164/63 already shipped for #1694: an interval is a genuinely new,
+;;; distinct opaque type (two parallel vectors of exact-integer lower/upper
+;;; bounds, arbitrary sign, not a vector of (lo . hi) pairs like 25's shape
+;;; nor plain zero-based sizes like 63's dims), and array?/array-set!'s
+;;; conventions (researched but not yet implemented -- later phases) turn
+;;; out to be their own hybrid of 25/164's and 63's rules, not identical to
+;;; either. Given the scale, SRFI 231 is being built across several slices:
+;;; this one covers only the interval layer (plus (srfi 231 misc)'s
+;;; permutation/translation helpers), which has zero dependency on arrays or
+;;; storage classes and is independently useful/testable. No bare
+;;; `(srfi 231)` library exists yet -- it will be assembled as a thin
+;;; re-export hub over these internal sub-libraries once every phase lands,
+;;; the same way `(srfi 160 base)` + per-tag files have no bare `(srfi 160)`
+;;; until a caller imports a specific typed sub-library.
+;;;
+;;; Every procedure's calling convention and error behavior below was
+;;; confirmed against the primary spec text (srfi.schemers.org/srfi-231)
+;;; and its official reference implementation before being written --
+;;; including two procedures (translation?, permutation?) whose exact
+;;; validation rule the spec states only in prose, not executable
+;;; pseudocode, and interval-fold-right's "all f evaluations before any
+;;; operator application" requirement, which is easy to implement as an
+;;; ordinary reverse fold without noticing that's insufficient (a naive
+;;; right-to-left walk interleaving f and operator calls would violate the
+;;; letter of the spec even though it 'looks' right for the common case
+;;; where f has no side effects).
+(define-library (srfi 231 intervals)
+  (import (scheme base) (srfi 1) (srfi 231 misc))
+  (export interval? make-interval interval-dimension
+          interval-lower-bound interval-upper-bound interval-width
+          interval-lower-bounds->list interval-upper-bounds->list
+          interval-lower-bounds->vector interval-upper-bounds->vector
+          interval-widths interval-volume interval-empty?
+          interval= interval-subset? interval-contains-multi-index?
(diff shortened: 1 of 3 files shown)


baijum:
Addressed all 6 findings — verified each against the primary SRFI 231 spec text (srfi.schemers.org) before fixing:

1. **`make-interval` surplus arguments** — confirmed the signature is `arg1` plus one optional `arg2`, not arbitrary trailing args. Fixed: rejects a 3rd+ argument instead of silently discarding it.
2. **`interval-subset?` dimension mismatch** — confirmed via direct fetch: the spec's "assumes ... the same dimension d" precondition makes a mismatch an error, with `#f` reserved for same-dimension intervals that fail the bound check (a separate, explicitly-quoted "otherwise, it returns #f" clause). Fixed.
3. **`interval-contains-multi-index?` inexact coordinates** — confirmed a multi-index must consist of exact integers per the formal spec. Fixed: rejects non-exact-integer entries instead of silently comparing them numerically (a `1.5` would otherwise pass the bounds checks).
4. **`interval-dilate`/`translate`/`permute`/`scale` length mismatches** — confirmed R7RS `vector-map` stops at the shortest input on a length mismatch, matching `map`, rather than erroring — meaning a too-short or too-long diff/translation/permutation/scales vector would silently produce a wrong-rank result. Fixed: explicit length validation before calling `vector-map` in all four.
5. **`interval-intersect` dimension mismatch** — real bug: the loop only ever iterated up to the *first* interval's own dimension, so a first interval with fewer axes than a later one never even looked at the extra axes, silently returning a lower-rank result instead of rejecting the mismatch. Fixed: validates all intervals share one dimension up front.
6. **`index-rotate`/`index-first`/`index-last`/`index-swap` preconditions** — confirmed each procedure's exact bound wording via direct fetch, including a genuinely asymmetric detail: `index-rotate`'s `k` may equal `n` (inclusive — the identity permutation), while `index-first`/`-last`/`-swap`'s `k`/`i`/`j` must be strictly less than `n` (exclusive). `index-last` with an out-of-range `k` previously produced a vector containing `-1` and missing an element — not a valid permutation at all. Fixed all four with validation matching their individually-confirmed bounds.

17 new assertions added (62 → 79), plus a full `zig build test` + `bash tests/scheme/run-all.sh` rerun (1975 pass, 1 known unrelated flake — #1748, confirmed passing standalone).

PR #1750, opened 2026-07-25
Add SRFI 231 storage classes (#1694 array family, phase 1b)

baijum:
## Summary

Second of several slices implementing SRFI 231 ("Intervals and Generalized Arrays") — see #1749 (phase 1a: intervals + misc, merged) for the overall roadmap.

This slice adds the storage-class abstraction: a 9-field record (`getter`, `setter`, `checker`, `maker`, `copier`, `length`, `default`, `data?`, `data->body`) managing a specialized array's backing store, plus all 17 required storage-class global variables.

- **14 of 17 map directly onto existing infrastructure, zero new engine work**: 11 numeric kinds (s8/s16/s32/s64/u16/u32/u64/f32/f64/c64/c128) reuse the already-shipped `(srfi 160 <tag>)` procedures; `u8` reuses plain R7RS bytevector procedures; `generic`/`char` reuse native `vector`/`string` directly via the spec's own verbatim reference definitions.
- **3 (`u1`, `f8`, `f16`) have no Kaappi-native representation and are bound to `#f`**, matching the spec's explicit "shall define... to `#f`" permission for implementations lacking the underlying type — the reference implementation itself does the same for `f8`.
- Every checker predicate encodes the exact range/type rule for its kind, confirmed against the same per-kind rules already established for SRFI 63's kind-dispatch table.

**Test-writing detail worth recording**: reading a complex number back from a c64/c128 `NumericVector` always produces a genuinely complex-tagged value, even with a zero imaginary part — unlike `make-rectangular` at the Scheme level, which collapses eagerly. This is a pre-existing Kaappi representation characteristic (not a bug); the test suite uses numeric equality (`=`) rather than `equal?` where this matters.

`(srfi 231)` itself remains not importable — still tracked under #1694. Remaining phases: core array object, views/sharing, combinators, multi-array assembly.

## Test plan

- [x] `kaappi check lib/srfi/231/storage-classes.sld` — clean
- [x] Manual smoke test covering all 17 storage classes (maker/getter/setter/length/checker/default/data?/data->body round-trip for the 14 real ones, confirming the 3 deferred ones are bound but `#f`) plus a fully custom user-constructed storage class
- [x] New `tests/scheme/srfi/srfi231-storage-classes.scm` (123 assertions) — all pass
- [x] `bash tests/scheme/run-all.sh` — 1976 pass, 1 known unrelated flake (#1748, confirmed passing standalone)

### DIFF
--- a/lib/srfi/231/storage-classes.sld
+++ b/lib/srfi/231/storage-classes.sld
@@ -1,0 +1,129 @@
+;;; SRFI 231 -- Storage classes (phase 1b of #1694's SRFI 231 slice).
+;;;
+;;; A storage-class is a 9-field record of procedures/values managing a
+;;; specialized array's backing store (getter, setter, checker, maker,
+;;; copier, length, default, data?, data->body) -- confirmed via primary
+;;; source + the official reference implementation before writing any code.
+;;; This mirrors, and can directly extend, the (kind maker ref setter!
+;;; length default) table already built for SRFI 63's array kinds
+;;; (lib/srfi/63.sld's %kind-table), just with two more fields (checker,
+;;; a value-validity predicate; data?/data->body, a zero-copy-sharing test
+;;; and converter used by make-specialized-array-from-data in a later
+;;; phase) and itself a first-class Scheme object rather than a bare list.
+;;;
+;;; 17 storage-class global variables must all be BOUND, but per spec --
+;;; "Implementations with an appropriate homogeneous vector type should
+;;; define the associated global variable using make-storage-class.
+;;; Otherwise, they shall define the associated global variable to #f" --
+;;; an implementation lacking the underlying element type may bind one to
+;;; #f. 12 of the 17 map cleanly onto this codebase's already-shipped
+;;; (srfi 160 <tag>) NumericVector substrate; 2 more (generic, char) need
+;;; no numeric substrate at all -- Kaappi's native vector/string already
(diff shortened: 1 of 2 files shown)


baijum:
Addressed one finding, declined the other after verifying against Kaappi's own source:

1. **Exercise `storage-class-copier`** — real gap, accepted. The shared test helper covered 8 of the 9 storage-class fields but never called `copier`, so a wrong copy procedure would have passed every existing assertion undetected. Added a copy-then-verify step, exercised across all 14 real storage classes (14 new assertions, 123 → 137).

2. **Import `(scheme complex)` for `make-rectangular`** — declined. This is correct for generic/strict R7RS (where `make-rectangular` lives only in the optional `(scheme complex)` library), but verified directly against `src/primitives_numeric.zig` that Kaappi registers it (along with `real-part`/`imag-part`) as a member of `.scheme_base` directly:
   ```
   .{ .name = "make-rectangular", ..., .libs = LS.initMany(&.{ .scheme_base, .scheme_complex, .scheme_r5rs }) },
   ```
   So this implementation already exposes it from `(scheme base)` — no import is missing, and the code already runs correctly (confirmed by the passing smoke test and suite) with just the existing `(scheme base)` import.

Full `zig build test` + `bash tests/scheme/run-all.sh` rerun clean (1976 pass, 1 known unrelated flake — #1748, confirmed passing standalone).

PR #1751, opened 2026-07-25
Add SRFI 231 core array object (#1694 array family, phase 2)

baijum:
## Summary

Third of several slices implementing SRFI 231 ("Intervals and Generalized Arrays") — see #1749 (phase 1a: intervals + misc) and #1750 (phase 1b: storage classes, both merged) for the overall roadmap.

This slice adds the array type itself: a domain (an interval) plus a getter, with an optional setter making it mutable, and an optional storage-class/body/indexer making it specialized.

- **One `<array>` record covers all four combinations** (plain immutable, plain mutable via closures, specialized mutable, specialized frozen-immutable) — specialized-only fields are simply `#f` on a plain array.
- **`array?`/`mutable-array?`/`specialized-array?` confirmed as a non-strict hierarchy**: `specialized-array? ⟹ array?` and `mutable-array? ⟹ array?`, but specialized and mutable are orthogonal — `array-freeze!` can make a specialized array immutable in place, and a closure-backed sparse array (per the spec's own illustrative example) can be mutable without being specialized at all.
- **Two conventions that don't match either prior array SRFI in this codebase exactly, both confirmed against the primary spec**: `array?` is disjoint from vector/string (matching 25/164, *not* 63's "vectors are rank-1 arrays" rule — getting this backwards was the single biggest miss when implementing 63), while `array-set!`'s value argument is *second* (matching 63, not 25/164's value-last). Getters/setters are always called with separate positional index arguments, never a packed vector/list.
- **`make-specialized-array`'s indexer** generalizes SRFI 63's 0-based-only row-major offset to arbitrary (including negative) lower bounds: shift each axis by its own lower bound first, then apply the standard row-major stride computation using the interval's own widths.
- **`make-specialized-array-from-data`** wraps externally supplied data (e.g. an existing `(srfi 160 <tag>)` vector) as a new array's body with zero copying — confirmed via a test that mutates through the array and checks the original vector changed.
- **`specialized-array-default-safe?`/`-mutable?`** are genuine SRFI 39 parameters, as the spec requires (a documented difference from the predecessor SRFI 179, where they were plain variables).

`(srfi 231)` itself remains not importable — still tracked under #1694. Remaining phases: views/sharing/reshaping, bulk combinators + conversions, multi-array assembly.

## Test plan

- [x] `kaappi check lib/srfi/231/arrays.sld` — clean
- [x] Manual smoke test covering plain immutable/mutable arrays, specialized arrays with arbitrary lower bounds, safe/unsafe access, `make-specialized-array-from-data` zero-copy sharing, `array-freeze!`, and every accessor's rejection of the wrong array kind
- [x] New `tests/scheme/srfi/srfi231-arrays.scm` (50 assertions) — all pass
- [x] `bash tests/scheme/run-all.sh` — 1977 pass, 1 known unrelated flake (#1748, confirmed passing standalone)

### DIFF
--- a/lib/srfi/231/arrays.sld
+++ b/lib/srfi/231/arrays.sld
@@ -1,0 +1,178 @@
+;;; SRFI 231 -- Core array object (phase 2 of #1694's SRFI 231 slice).
+;;;
+;;; Builds on phase 1a (lib/srfi/231/intervals.sld) and phase 1b
+;;; (lib/srfi/231/storage-classes.sld). See intervals.sld's header for the
+;;; overall multi-slice roadmap; (srfi 231) itself is still not importable
+;;; as a bare library until every phase lands.
+;;;
+;;; An array is a domain (an interval) plus a getter (a procedure called
+;;; with SEPARATE positional index arguments, confirmed throughout the
+;;; primary spec text and its reference implementation -- never a packed
+;;; vector/list). A setter, if present, makes it mutable; array-set!'s
+;;; value argument is SECOND (right after the array), matching SRFI 63's
+;;; convention, not SRFI 25/164's value-last -- confirmed directly from
+;;; the spec's own formal signature, "array-set! array object . multi-index".
+;;; array? is disjoint from vector/string here (matching 25/164, NOT
+;;; 63's "vectors/strings are rank-1 arrays" rule) -- confirmed via direct
+;;; fetch specifically because getting this backwards was the single
+;;; biggest miss when implementing SRFI 63.
+;;;
+;;; specialized-array? => array? (spec: "A specialized-array is an
+;;; array."); mutable-array? => array? (entailed by make-array's own
+;;; contract). specialized-array? and mutable-array? are orthogonal --
+;;; array-freeze! can make a specialized array immutable in place, and a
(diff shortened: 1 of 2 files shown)


baijum:
Addressed both findings:

1. **`array-packed?` early exit** — fixed. Now escapes via `call/cc` on the first packing mismatch instead of always scanning the full domain. Harmless today (specialized arrays are always packed right after construction — there's no code path yet that produces a non-packed one), but avoids a full `interval-volume` traversal once phase 3's view/share/reverse procedures land.
2. **0-dim/empty-domain test coverage** — added. 13 new assertions covering a zero-dimensional array (single element, accessed via a thunk-shaped getter and `array-ref` with no indices, for both a plain and a specialized array) and empty-domain arrays (some axis has lower = upper, so no multi-index is ever valid — verified rejection in both safe mode, via the explicit domain check, and unsafe mode, via the storage layer's own bounds check on the resulting zero-length body).

63 assertions total (up from 50). Full `zig build test` + `bash tests/scheme/run-all.sh` rerun clean (1977 pass, 1 known unrelated flake — #1748, confirmed passing standalone).

PR #1752, opened 2026-07-25
Add SRFI 231 views, sharing, and reshaping (#1694 array family, phase 3)

baijum:
## Summary

Fourth of several slices implementing SRFI 231 ("Intervals and Generalized Arrays") — see #1749/#1750/#1751 (phases 1a/1b/2, all merged) for the overall roadmap.

This slice adds the 11-procedure "views/sharing/reshaping" cluster the spec's own "Sharing generalized arrays" section names as a natural unit: `specialized-array-share` (the foundational primitive), `array-extract`, `array-translate`, `array-permute`, `array-reverse`, `array-curry`, `array-tile`, `array-sample`, `array-copy`/`array-copy!`, and `specialized-array-reshape`.

**`lib/srfi/231/arrays.sld`** (phase 2, already merged) gets one small, purely additive change: `%make-array`/`%safe-getter`/`%safe-setter`/`%make-lex-indexer` are now exported for this and later phases' sibling library files to build genuinely specialized arrays with custom indexers — the same way `(srfi 160 base)`'s `%uvec-*` helpers exist only for that package's own per-tag files.

**`lib/srfi/231/views.sld`**:
- `array-extract`/`translate`/`permute`/`reverse`/`sample` all share one 3-way dispatch shape (specialized → `specialized-array-share`; mutable-non-specialized → `make-array` with getter+setter; immutable → `make-array` with getter only), with the output mirroring the input's mode exactly.
- `array-curry`/`array-tile` break this pattern deliberately — their **outer** array is unconditionally plain immutable/non-specialized regardless of input mode (confirmed by an explicit spec quote for curry: *"B is always an immutable array... computed anew for each call"*, i.e. never cached), while their **inner** elements/tiles mirror the input mode 3-way like everything else.
- `array-permute` needed real care to get the index-rearrangement direction right (apply the permutation's *inverse* to the new multi-index to recover old coordinates) — verified against both of the spec's own worked examples.
- `array-tile`'s last-slice truncation (when an axis width doesn't divide evenly by a uniform slice size) is spec-sanctioned via an explicit `min()`, not an error — verified against the spec's own 6×6 non-uniform worked example exactly, including the truncated case.
- **`specialized-array-reshape` deliberately implements a conservative simplification** of the reference implementation's full NumPy-derived multi-group stride-matching algorithm: it succeeds zero-copy whenever the source is already `array-packed?` (the overwhelmingly common case) via a plain row-major reindex, and otherwise behaves exactly as the spec allows for a failed detection (error, or forced-copy-then-retry when `copy-on-failure?` is `#t`). Never wrongly claims an affine map exists; verified identical to the full algorithm on both of the spec's own worked examples.
- `array-copy`/`array-copy!` are implemented identically (a direct fill loop) — a deliberate, documented scope reduction versus the spec's optional extra call/cc-safety guarantee for `array-copy` specifically.

`(srfi 231)` itself remains not importable — still tracked under #1694. Remaining phases: bulk combinators + conversions, multi-array assembly.

## Test plan

- [x] `kaappi check lib/srfi/231/views.sld` — clean
- [x] Extensive manual smoke test against the spec's own worked examples (shearing, permute rank-3/rank-4, palindrome-checker-shaped reverse, 6×6 non-uniform tile with truncation, sample, both reshape success/failure cases)
- [x] New `tests/scheme/srfi/srfi231-views.scm` (75 assertions) — all pass
- [x] `bash tests/scheme/run-all.sh` — 1978 pass, 1 known unrelated flake (#1748, confirmed passing standalone)

### DIFF
--- a/lib/srfi/231/arrays.sld
+++ b/lib/srfi/231/arrays.sld
@@ -34,7 +34,13 @@
           array-domain array-getter array-setter array-dimension
           array-ref array-set! array-freeze! array-empty?
           array-body array-indexer array-storage-class array-safe? array-packed?
-          specialized-array-default-safe? specialized-array-default-mutable?)
+          specialized-array-default-safe? specialized-array-default-mutable?
+          ;; Not part of SRFI 231's own public API -- exported only for
+          ;; sibling library files in this package (lib/srfi/231/views.sld
+          ;; and later phases) to build genuinely specialized arrays with
+          ;; custom indexers, the same way (srfi 160 base)'s %uvec-* helpers
+          ;; exist only for that package's own per-tag files to consume.
+          %make-array %safe-getter %safe-setter %make-lex-indexer)
   (begin
 
     (define-record-type <array>
--- a/lib/srfi/231/views.sld
+++ b/lib/srfi/231/views.sld
@@ -1,0 +1,365 @@
+;;; SRFI 231 -- Views, sharing, and reshaping (phase 3 of #1694's SRFI 231
+;;; slice). Builds on phases 1a/1b/2 (intervals, storage classes, core
+;;; array object). See intervals.sld's header for the overall roadmap;
+;;; (srfi 231) itself is still not importable as a bare library.
+;;;
+;;; specialized-array-share is the foundational primitive here -- every
+;;; view/transform procedure that shares a specialized array's body (all
+;;; except array-curry/array-tile's own OUTER array, array-copy, and
+;;; specialized-array-reshape) is defined in terms of it. It composes a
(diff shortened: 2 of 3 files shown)


baijum:
Addressed the finding: added a test case exercising explicit `storage-class`/`mutable?`/`safe?` overrides on `array-copy` (previously only the default omitted-options inheritance path was covered). 3 new assertions (75 → 78).

Full `zig build test` + `bash tests/scheme/run-all.sh` rerun clean (1978 pass, 1 known unrelated flake — #1748, confirmed passing standalone).

PR #1753, opened 2026-07-25
Add SRFI 231 bulk combinators and conversions (#1694 array family, phase 4)

baijum:
## Summary

Fifth of several slices implementing SRFI 231 ("Intervals and Generalized Arrays") — see #1749/#1750/#1751/#1752 (phases 1a/1b/2/3, all merged) for the overall roadmap.

This slice adds `array-map`/`for-each`/`fold-left`/`fold-right`/`reduce`/`any`/`every`, `array-outer-product`/`inner-product`, and the 8 flat and nested list/vector conversion procedures (`array->list`, `list->array`, `array->list*`, `list*->array`, `array->vector`, `vector->array`, `array->vector*`, `vector*->array`) — 17 bindings in total.

- `array-map`/`for-each`/`fold-left`/`fold-right`/`any`/`every` are all variadic over arrays, requiring exact domain equality (`interval=`, not just matching shape) across every argument array. `array-reduce` is the one exception: fixed 2-arg, single array only, and hard-errors on an empty array with no safety-flag opt-out.
- **`array-fold-left` calls its operator as `(operator acc e0 e1 ...)`; `array-fold-right` calls it as `(operator e0 e1 ... acc)`** — accumulator *last*, not first. Confirmed via the spec's own formal reference definition, and verified the divergence is real (not just notational) against the spec's own worked examples.
- **`array-inner-product`'s compositional definition needed care in two places the spec's own prose pseudocode gets subtly wrong**, both confirmed only by reading the reference implementation directly: it omits a required argument to `array-curry` on its second call, and its stated preconditions don't mention that the shared axis's width must be nonzero.
- **`list*->array`/`array->list*`'s empty-collection edge cases are genuinely undomesticable by intuition alone** — verified by hand-tracing all four of the spec's own worked examples against the actual algorithm: `(list*->array 0 '())` yields a rank-0, volume-1 array whose single element *is* the empty list, while `(list*->array 1 '())` yields a genuinely empty rank-1 array. `vector*->array`/`array->vector*` delegate to the list* versions via a structure-preserving conversion rather than duplicating the shape-inference algorithm.

`(srfi 231)` itself remains not importable — still tracked under #1694. Remaining phase: multi-array assembly (the final one).

## Test plan

- [x] `kaappi check lib/srfi/231/combinators.sld` — clean
- [x] Extensive manual smoke test against the spec's own worked examples (fold-left/right divergence, inner-product dot-product, all 4 list*->array empty-edge-cases, vector*/list* round trips)
- [x] New `tests/scheme/srfi/srfi231-combinators.scm` (55 assertions) — all pass
- [x] `bash tests/scheme/run-all.sh` — 1979 pass, 1 known unrelated flake (#1748, confirmed passing standalone)

### DIFF
--- a/lib/srfi/231/combinators.sld
+++ b/lib/srfi/231/combinators.sld
@@ -1,0 +1,297 @@
+;;; SRFI 231 -- Bulk combinators and conversions (phase 4 of #1694's
+;;; SRFI 231 slice). Builds on phases 1a/1b/2/3 (intervals, storage
+;;; classes, core array object, views/sharing). See intervals.sld's
+;;; header for the overall roadmap; (srfi 231) itself is still not
+;;; importable as a bare library.
+;;;
+;;; array-map/for-each/fold-left/fold-right/any/every are all variadic
+;;; over arrays (one mandatory array plus a rest-list of additional
+;;; arrays, ALL of which must have the exact SAME domain -- interval
+;;; equality, not just matching shape). array-reduce is the one
+;;; exception: fixed 2-arg, single array only, hard-errors on an empty
+;;; array (confirmed unconditional in the reference implementation --
+;;; there is no seed/identity to fall back on, unlike fold-left/right).
+;;; array-map returns a lazy (recomputed on every access), immutable,
+;;; non-specialized array -- never eagerly evaluated.
+;;;
+;;; array-fold-left calls operator as (operator acc e0 e1 ...); array-
+;;; fold-right calls it as (operator e0 e1 ... acc) -- accumulator LAST,
+;;; not first -- confirmed via the spec's own formal reference
+;;; definition, which passes elements as SEPARATE positional arguments
+;;; via apply, never as one packed list argument. Confirmed via the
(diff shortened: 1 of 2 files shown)


baijum:
Addressed the finding — both parts:

1. **Dead `%vector-every`** — confirmed never called anywhere in this file (a leftover from the same pattern used in sibling files where it IS needed). Removed.
2. **Unvalidated `vector*->array` input** — real gap. `%nested-vector->nested-list` had no `vector?` check before delegating to `vector->list`, unlike the list path (`%check-nested-list` validates via `list?` at every recursion level). Fixed by validating `vector?` at each level before delegating, matching the list path's error message/irritants — rectangularity is still validated for free once `%check-nested-list` runs on the converted result, so no duplicate logic was needed there.

1 new regression assertion (55 → 56). Full `zig build test` + `bash tests/scheme/run-all.sh` rerun clean (1979 pass, 1 known unrelated flake — #1748, confirmed passing standalone).

PR #1767, opened 2026-07-26
Set peers_computed only after the snapshot is durably stored

baijum:
## Summary

- Follow-up to #1766, which had already auto-merged by the time
  CodeRabbit's review landed. Addresses that review directly:
  https://github.com/kaappi/kaappi/pull/1766#pullrequestreview-4782742562
- #1766 set `Transformer.peers_computed = true` immediately, before the
  several fallible allocations (`peer_names_f`/`peer_vals_f` appends,
  both `dupe` calls) that actually build the `let_syntax_peer_names`/
  `vals` snapshot. An OOM partway through would leave the flag
  permanently `true` with both fields still at their default-empty
  value — every later reuse of that transformer would then treat "no
  sibling suppression needed" as the final, correct answer instead of
  retrying the computation.
- Fixed by moving the flag assignment to strictly after both slices are
  durably stored, immediately before the `self.macros.put` that was
  already there — matching CodeRabbit's exact suggested diff.

## Test plan

- [x] Both discriminating reproductions from #1765/#1766 (macro
      redefined between two forms; nested reuse corrupting the outer
      binding) still resolve correctly
- [x] `tests/scheme/srfi/srfi147.scm`: 25/25 pass; `srfi257.scm`: 34/34
- [x] `zig build test` (full suite, incl. `-Dgc-stress=true` on
      `tests_macros`): no failures, no leaks
- [x] `bash tests/scheme/run-all.sh`: 591 Scheme files + 1395 R7RS
      tests, 0 failures

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -716,6 +716,16 @@ changes actually are, so any two different uncommitted edits at the same
 base commit alias to the identical id and share cache entries -- see
 `docs/dev/cache.md`.)
 
+CodeRabbit's review of that fix caught one more ordering bug: the first
+cut set `peers_computed = true` immediately, before the several fallible
+allocations (`peer_names_f`/`peer_vals_f` appends, both `dupe` calls) that
+actually build the snapshot. An OOM partway through would leave the flag
+permanently true with `let_syntax_peer_names`/`vals` still at their
+default-empty value -- every later reuse would then treat "no suppression
+needed" as the final, correct answer instead of retrying. Fixed by moving
+the flag assignment to strictly after both slices are durably stored,
+right before the `self.macros.put` that was already there.
+
 These differ from earlier Zig versions and are easy to get wrong:
 
 ```zig
--- a/src/compiler_macro.zig
+++ b/src/compiler_macro.zig
@@ -434,7 +434,6 @@ pub fn compileLetSyntax(self: *Compiler, args: Value, dst: u16, is_tail: bool) C
             self.macros.put(name, transformer) catch return CompileError.OutOfMemory;
             continue;
         }
-        tx.peers_computed = true;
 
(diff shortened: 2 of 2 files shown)


baijum:
Investigated CodeRabbit's suggestion to add an OOM fault-injection regression test for the `peers_computed` ordering fix. Declining it specifically, after actually attempting it — not skipping on assumption.

What I tried: `std.testing.FailingAllocator` swapped into the GC (the same technique `tests_shared_channel.zig` uses for its own OOM paths), targeting the fallible appends/dupes in `compileLetSyntax`'s peer-snapshot block. Built it, then **mutation-tested it** by temporarily reverting the fix (moving `peers_computed = true` back to its old, buggy position) to confirm the new test would actually catch the regression.

It didn't. `fail_index = 0` (the only value the existing precedent uses) fails whatever allocation happens to come first after the swap — which, reached through the full `eval` → compile pipeline rather than `tests_shared_channel.zig`'s direct, shallow function call, lands in unrelated setup (resolving the transformer-spec, building `merged_macros`, the `kw_names`/`tx_vals`/`peer_snap_*` arrays) well before the code path in question. The test passed identically against both the buggy and fixed code — meaning it wasn't testing what it claimed to, which is worse than no test: false confidence instead of none.

Making it actually discriminate would mean precisely counting allocations up to the exact `dupe` call, which is fragile by construction — it silently stops meaning anything the moment an unrelated change adds or removes an allocation upstream, with no signal that it happened.

The fix itself doesn't depend on this test to be correct: it's a single, statically-auditable reordering (moving one flag assignment after the operations it must not race), and this PR's existing coverage (the two discriminating scenarios from #1765/#1766) already exercises the surrounding logic thoroughly. Open to reconsidering if there's a more targeted way to reach this specific allocation that doesn't require hand-counting.

PR #1779, opened 2026-07-27
Fix custom-ellipsis recognition in ordinary template expansion (follow-up to #1776)

baijum:
## Summary

CodeRabbit's review of #1776 (already merged) caught a related, distinct
gap: that PR fixed custom-ellipsis recognition on the **parsing** side
(`parseSyntaxRules`'s ellipsis-argument detection) but missed the
**instantiation** side.

`instantiateTemplate` has two places that check whether an element is the
active ellipsis symbol — the `(... <template>)` escape form, and the
"element followed by ellipsis" detection used for ordinary `(Te ...)`
splicing. Both required a bare, unwrapped symbol. A nested `syntax-rules`
template's *ordinary* (non-quoted) position can receive a custom ellipsis
identifier substituted from an outer pattern variable, wrapped by
`NESTED_SR_FLAG`'s usertext-marking protocol — and neither site unwrapped
it, so the ellipsis went unrecognized there. (A *quoted* position happens
to dodge this, since that same substitution skips wrapping entirely under
`QUOTE_FLAG` — which is why #1776's own test, which used a quoted template,
didn't catch this.)

```scheme
(define-syntax gen (syntax-rules ()
  ((_ my-ellipsis)
   (syntax-rules my-ellipsis ()
     ((_ x my-ellipsis) (list x my-ellipsis))))))
(define-syntax use-it (gen :::))
(use-it 1 2 3)
;; before: undefined variable ':::' (ellipsis never recognized, emitted literally)
;; after:  (1 2 3)
```

Fixed by unwrapping at both detection sites, mirroring the existing
`parseSyntaxRules` fix.

## Also in this PR

CodeRabbit separately flagged `src/tests_macros.zig` crossing the project's
1500-line policy (pushed over by #1776's own new tests). Split the
nested-syntax-rules-generation regression tests (5 of them, including this
PR's new one) into `src/tests_macros_nested_sr.zig` — a natural seam, since
they're all specifically about macros that generate a fresh nested
`syntax-rules` transformer, distinct from `tests_macros.zig`'s more general
hygiene tests.

## Testing

- New regression test in `tests_macros_nested_sr.zig` reproducing the exact
  shape above.
- `zig build test`: full unit suite green.
- `bash tests/scheme/run-all.sh`: 1992/1992, 0 fail.

## Test plan

- [x] `zig build test`
- [x] `bash tests/scheme/run-all.sh`
- [x] Manual repro of the bug and confirmation of the fix

### DIFF
--- a/src/expander.zig
+++ b/src/expander.zig
@@ -828,9 +828,23 @@ fn instantiateTemplate(gc: *GC, template: Value, bindings: []Binding, intro_scop
     }
 
     // Check for ellipsis escape: (... <template>) — treat ... as literal inside
+    //
+    // Both ellipsis checks below unwrap a usertext marker first: a nested
+    // syntax-rules template's ordinary (non-quoted) position can receive a
+    // custom ellipsis identifier substituted from an OUTER pattern variable
+    // (SRFI 147/148's generating-macro pattern), which NESTED_SR_FLAG's
+    // marking protocol wraps so the generating macro's own later expansion
+    // doesn't re-walk it as template text. Only a quoted position skips the
+    // wrap (QUOTE_FLAG suppresses it, see the pattern-variable substitution
+    // case above) -- an ordinary template position needs the unwrap to even
+    // recognize the ellipsis at all; without it, `(x my-ellipsis)` never
+    // matches "element followed by ellipsis" and my-ellipsis is emitted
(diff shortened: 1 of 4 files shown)


baijum:
@coderabbitai review

Issue #1787, opened 2026-07-27
free-identifier=?/bound-identifier=? (portable CK-machine idiom) break when compared identifier is a compound form, not a bare symbol


baijum:
## Summary

The portable, low-level `free-identifier=?`/`bound-identifier=?` idiom (the
standard "redefine + auto-literal" trick, as used by SRFI 148's reference
implementation and traced verbatim into `lib/srfi/148.sld`'s draft) breaks
when the identifier being *compared against* is a compound s-expression
(e.g. `(quote t)`) rather than a bare symbol — the exact shape that arises
whenever two `em-syntax-rules` quoted pattern variables appear adjacently
(SRFI 148's own `em-syntax-rules-aux2` peeks at "the pattern element right
after the current one" to decide whether it's the ellipsis identifier, and
that "next element" is itself `(quote nextvar)` when there are 2+ quoted
pattern variables in a row).

This blocks any `em-syntax-rules`-defined macro with more than one quoted
pattern variable (or any ordinary, non-quote-wrapped template — see
repro 2), i.e. almost every real combinator in SRFI 148's ~110-procedure
library (e.g. `em-cons`: `((em-cons 'h 't) '(h . t))`).

## Minimal reproduction

```scheme
(import (scheme base) (scheme write))

(define-syntax free-identifier=?
  (syntax-rules ()
    ((free-identifier=? id1 id2 kt kf)
     (begin
       (define-syntax m
         (syntax-rules :::1 ()
           ((m %kt %kf)
            (begin
              (define-syntax test
                (syntax-rules :::2 (id1)
                  ((test id1 %%kt %%kf) %%kt)
                  ((test x %%kt %%kf) %%kf)))
              (test id2 %kt %kf)))))
       (m kt kf)))))

;; id2 is a COMPOUND form, not a bare identifier.
(define-syntax test-compound
  (syntax-rules ()
    ((_) (free-identifier=? ... (quote t) 'yes 'no))))

(write (test-compound))
```

Expected: `no` (a compound form is never `free-identifier=?` to `...`).

Actual:

```
error[KP3005]: not a procedure
    (write (test-compound)) (newline)
```

`kaappi expand` on this file shows the macro-use collapsing to
`((quote t) (quote yes) (quote no))` — i.e. `id2`'s value ends up being
treated as if it were the operator of a call applied to `kt`/`kf`, instead
of `test`'s own pattern-match (literal-vs-catchall) ever actually
[...]

Issue #1775, opened 2026-07-27
em-syntax-rules-style generating macros: severe (near-exponential) slowdown as rule count grows


baijum:
## Summary

Blocking discovery while implementing SRFI 148 (issue #1699's final tractable
slice, task in progress — no PR yet, work paused pending this). SRFI 148's
`em-syntax-rules` and most of its ~110 combinators are macros that generate
a fresh, nested `syntax-rules` transformer at expansion time (the reference
implementation's own technique, ported faithfully). This mechanism, when
the generated macro has more than a handful of rules, becomes severely,
apparently super-linearly slow — to the point of being impractical.

## Reproduction

Using a minimal, self-contained CK-machine-style `em-syntax-rules`
implementation (matching SRFI 148's own `148.scm`, no SRFI-148-specific
code needed to reproduce):

| Macro | Rules | Rule shape | Compile time |
|-------|:-----:|-----------|:---:|
| 1-rule macro | 1 | simple | instant |
| 2-rule macro | 2 | simple | 0.13s |
| 3-rule macro | 3 | simple | 0.47s |
| `em-cut-eval`-shaped | 3 | custom ellipsis + `<>`/`...` literals | 0.10s |
| `em-quasiquote-aux`-shaped, first 4 of 8 rules | 4 | quasiquote-containing patterns (`` `form ``, `,form``, `,@form`) | 2.0s |
| `em-quasiquote-aux`-shaped, all 8 rules | 8 | same | **41s** |

Going from 4 → 8 rules (2x) costs ~20x, not the ~4x a naive O(rules²)
accumulation would predict — the growth looks worse than quadratic, at
least for this specific (quasiquote-pattern-heavy) rule shape. Plain,
non-quasiquote 3-rule macros are comparatively cheap (0.47s), so the
pattern/template *shape* materially affects the blowup, not just the count.

With the full ~110-combinator SRFI 148 library (many multi-rule, several
quasiquote-shaped), this made the library effectively uncompilable within
any reasonable CI/dev-loop budget.

## What's confirmed NOT the (sole) cause

- **Not exponential re-visiting of shared subtrees in `stripUsertextMarkers`**:
  added identity-based memoization (a visited-set keyed by heap-object
  pointer) to `stripUsertextWalk` — no measurable improvement, and the
  minimal repro above doesn't even reach that code path in a way that
  would explain it (confirmed by disabling the one new call site entirely;
  timing was unchanged).
[...]

PR #1788, opened 2026-07-27
Fix dotted-tail usertext-marker matching gap in matchListPattern

baijum:
## Summary

- `matchPattern` unwraps its own `pattern_in`/`input_in` parameters once, at entry — but `matchListPattern`'s internal loop advances `pat`/`inp` on every iteration without re-unwrapping. A usertext marker pair spliced into a **dotted-tail** pattern position (e.g. a generated macro's pattern `(_ head . p)` where `p`'s substituted value is a list) is only caught by the entry-level unwrap when it sits in the very first position — anything preceding it in the pattern forces the marker to surface mid-loop as a bogus extra list element, shifting every subsequent position and failing the match ("bad arguments to macro call" even for well-formed calls).
- Fix: unwrap `pat`/`inp` at the top of `matchListPattern`'s loop body, before the dotted-tail check. `unwrapUsertext` is a pure, non-mutating accessor, so this carries none of the shared-structure-mutation risk of an earlier, reverted attempt at a different fix for a related symptom (see commit message / project memory for that history).

## Context

Found while resuming SRFI 148 (eager syntax-rules, tracked under #1699) after #1775 (a separate, unrelated performance cliff in `em-syntax-rules`-style generating macros) was fixed in #1784. SRFI 148's generated macro pattern is always exactly this dotted-tail shape (`(_ :prepare s . p)`), so every `em-syntax-rules`-defined macro hit this.

This fix is necessary but not sufficient to unblock SRFI 148 — a separate, deeper bug (`free-identifier=?`/`bound-identifier=?` failing when the compared identifier is a compound form, not a bare symbol) still blocks it. Filed as #1787 with full repro steps; that's a materially different investigation, not a quick follow-up to this PR.

## Test plan

- [x] New regression test in `tests_macros_nested_sr.zig`, mutation-tested (fails with the fix disabled via a manual toggle, passes with it enabled)
- [x] `zig build test` — full unit suite green
- [x] `bash tests/scheme/run-all.sh` — 1995/1995 pass
- [x] A/B verified `free-identifier=?`/`bound-identifier=?` scenarios (which regressed under an earlier, unrelated, reverted fix attempt for a related symptom) produce byte-for-byte identical output with this fix enabled vs. disabled — no regression

### DIFF
--- a/src/expander.zig
+++ b/src/expander.zig
@@ -374,6 +374,24 @@ fn matchListPattern(pattern: Value, input: Value, literals: []const Value, bindi
     var inp = input;
 
     while (pat != types.NIL) {
+        // Unwrap a usertext marker BEFORE inspecting this iteration's shape.
+        // matchPattern's own top only unwraps its immediate pattern_in/
+        // input_in parameters; a marker pair spliced into a dotted-tail
+        // position by a generating macro's own template instantiation (a
+        // pattern-variable value that's a LIST, substituted into `. p`) is
+        // never wrapped in an extra cons cell of its own -- it becomes the
+        // literal cdr of the preceding pair, e.g. `(s . (marker . (%a)))` =
+        // `(s marker %a)` when walked. Without re-unwrapping here on every
+        // loop iteration (not just at entry), that marker pair surfaces as
+        // an ordinary-looking extra list element on a LATER iteration,
+        // shifting every remaining pattern/input position by one and
+        // breaking the match entirely (kaappi#1775's still-open companion
+        // bug: SRFI 148's em-syntax-rules generates exactly this shape for
+        // every rule, `(_ :prepare s . p)`). unwrapUsertext is a no-op on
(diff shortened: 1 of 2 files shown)


baijum:
Superseded by #1790, which cherry-picks this commit verbatim (@edf3209c) and adds two more instances of the same gap plus their regression tests. Either close this in favour of #1790, or merge it first — the identical commit drops out of #1790's rebase cleanly.

One correction to this PR's description: it says the `free-identifier=?`/compound-identifier bug (#1787) still blocks SRFI 148. That bug does not exist — #1787's repro writes a bare `...` in a template where R7RS requires `(... ...)`, and Kaappi's silent zero-expansion of a driver-less ellipsis (now #1791) is what made the result look like a comparison failure. With this fix plus the two in #1790, a full SRFI 148 port's combinators expand correctly end-to-end; details in #1787.

Issue #1799, opened 2026-07-27
Native backend miscompiles `apply` over any local binding: evaluated in the global environment


baijum:
In kaappi v0.21.0 (build 0687dc4), the LLVM native backend miscompiles `apply` whenever any of its arguments is a **local binding** — a procedure parameter or a `let`-bound name. The binary compiles and links cleanly and then fails at run time.

## Reproducer

```scheme
(define (s xs) (apply + xs))
(display (s (list 1 2 3)))
(newline)
```

```
$ kaappi t.scm
6
$ kaappi compile t.scm -o tbin && ./tbin
eval error: undefined variable 'xs'. Did you mean 's'? (UndefinedVariable)
```

## Characterisation

Verified on macOS aarch64.

| shape | result |
|---|---|
| `(define (s xs) (apply max xs))` | fails, names `xs` |
| `(define (s f xs) (apply f xs))` | fails, names `f` — the operator position too, not just the list |
| `(define (s) (let ((xs (list 1 2 3))) (apply + xs)))` | fails — locals generally, not just parameters |
| `(define g (list 1 2 3)) (define (s) (apply + g))` | **works** — globals are fine |
| `(display (apply + (list 1 2 3)))` at top level | **works** — literals are fine |
| `(fold + 0 xs)`, explicit recursion over a parameter | **work** |

`call/cc`, `call-with-current-continuation`, `call-with-values` and `eval` fail the same way:

```scheme
(define (s x) (call-with-values (lambda () (values x 2)) +))  ; undefined variable 'x'
(define (s x) (call/cc (lambda (k) (+ x 1))))                 ; undefined variable 'x'
```

A same-named global turns the error into a **silently wrong answer**:

```scheme
(define x 100)
(define (s x) (apply + (list x)))
(display (s 1))   ; interpreter: 1     compiled binary: 100
```

## Root cause

`apply` lowers to an IR `.passthrough` node, and `emitPassthrough` serializes those to source text for `kaappi_eval_cached` — an interpreter eval that resolves names in the **global** environment. All four native-compilation gates (both closure tiers, `tryCompileDefineFunction`, `emitLet`) exist to stop a lexical scope being split that way (#827); each consults `freevars.sexprNeedsEvalFallback`, which matches head keywords against the comptime-derived `ir.eval_fallback_form_names`.

[...]

Issue #1496, opened 2026-07-13
LLVM backend: lower cond/case/do natively instead of eval fallback


baijum:
Part of #1491.

## Problem

`cond`, `case`, and `do` currently fall back to `kaappi_eval`
(`ir.eval_fallback_form_names`; `src/llvm_emit.zig` table). But they desugar into
`if` / `let` / loops that the emitter already lowers natively and well.

## Proposed change

Lower `cond`, `case`, and `do` to native IR by desugaring to the already-native
primitives (`if`, `let`, self-tail-call loop for `do`) rather than routing them
through eval. Reuse the existing `emitIf` / `emitLet` / self-tail-call machinery.

Mind lexical scope: like `emitLet`, bail to the interpreter for a sub-form that
cannot be lowered in scope rather than emitting incorrect global-env code
(cf. #827 handling).

## Acceptance criteria

- `cond`/`case`/`do` produce native code (no `kaappi_eval` call) for the common
  cases; correctness preserved via `tests/e2e` diff + `tests/scheme` suites.
- Fallback still correct for cases that legitimately can't be lowered.

PR #1805, opened 2026-07-28
Lower `apply` natively in the LLVM backend (#1803)

### LINKED ISSUE #1803: Lower `apply` natively in the LLVM backend instead of declining the whole enclosing function
Follow-up to #1799 / PR #1798.

## Current state

#1799 fixed a miscompilation: `apply` lowers to an IR `.passthrough` node that `emitPassthrough` hands to `kaappi_eval` — which resolves names in the *global* environment — so `(define (s xs) (apply + xs))` compiled cleanly and then died with "undefined variable 'xs'". The fix adds `apply` (and `call/cc`, `call-with-current-continuation`, `call-with-values`, `eval`) to `ir.eval_fallback_form_names`, so every native-compilation gate declines the enclosing lambda frame or `let` and the interpreter — which owns the lexical scope — handles the whole form.

That is correct, and it matches what the `cond`/`case`/`do` gate (`isRejectedFormHead`, #1496) already did. But it is blunt.

## The cost

Declining is **all-or-nothing for the enclosing function**: one `apply` anywhere in a body sends *every other expression in that body* to the interpreter too, however hot it is.

```scheme
;; heavy-pure.scm — natively compiled
(define (work n acc)
  (if (= n 0) acc
      (work (- n 1) (+ acc (* n 3) (- n 1) (* n n)))))
(display (work 3000000 0))
```

```scheme
;; heavy-apply.scm — identical, plus ONE apply that runs ONCE, on exit
(define (work n acc)
  (if (= n 0) (apply + (list acc 0))
      (work (- n 1) (+ acc (* n 3) (- n 1) (* n n)))))
(display (work 3000000 0))
```

| program | native defines emitted | eval fallbacks | compiled binary (3 runs) |
|---|---:|---:|---|
| `heavy-pure` | 1 | 1 | 0.28 / 0.24 / 0.27 s |
| `heavy-apply` | 0 | 2 | 4.82 / 4.94 / 5.23 s |

**~19x slower**, from an `apply` that executes exactly once. The 3-million-iteration arithmetic loop — pure fixnum work the backend compiles well, including the self-tail-call-to-loop optimization — is what actually got de-optimized.

This is the shape that matters in practice: not that `apply` is slow, but that *containing* an `apply` is contagious to the whole function. `(apply + xs)` is the idiomatic list sum, so it turns up in exactly the numeric code that most wants the native backend. (A microbenchmark of `apply` in isolation shows nothing — `(apply + xs)` over a 100-element list spends its time inside the variadic `+` primitive, so the enclosing tier is invisible. The regression only appears when the function has other hot work.)

## Proposal

Lower `apply` natively so only `apply` itself goes through the runtime, and the enclosing frame keeps its native compilation.

`emitCallNode` already emits `call i64 @kaappi_call_scheme(ptr %vm, i64 %callee, ptr %args_alloca, i64 N)` with a fixed-size `alloca` of arguments. `apply` needs the same, except the final argument is a list to splice at run time, so N is not known at emit time. That needs one new C-ABI entry point in `runtime_exports.zig`, alongside `kaappi_call_scheme`:

```zig
// fixed_args[0..n_fixed] followed by the elements of `list`.
pub export fn kaappi_apply(vm: ?*vm_mod.VM, callee: u64,
                           fixed_args: ?[*]const u64, n_fixed: u64,
                           list: u64) callconv(.c) u64
```

The body is essentially `primitives.applyFn` (`src/primitives.zig`) minus the argument-shuffling, which already does the work this needs and is the semantics to match exactly:

- the improper-list check, using the same tortoise-and-hare cycle detection (a circular list must raise, not hang);
- `isProcedure` / `isNativeFn` validation on the callee;
- the `typeError("apply", …)` messages, so a native binary's diagnostics match the interpreter's verbatim.

Then `emitCallNode` grows an `apply` case: emit the callee and the fixed arguments into an `alloca` as it does today, emit the final argument as a Value, and call `@kaappi_apply` instead of `@kaappi_call_scheme`.

### Things to get right

- **GC rooting.** The splice allocates (`call_args` is an `ArrayList`), and the callee plus every already-evaluated fixed argument must be rooted across it. `emitCallNode`'s existing `emitRootPush`/`emitPopRoots` scaffolding covers the emit side; the runtime side needs the usual care from `.claude/rules/gc-safety.md`.
- **Tail position.** `emitCallNode` emits `tail call` and pops `frame_entry_roots + body_scope_roots` before the `ret`. A tail `(apply f xs)` should get the same treatment, or deep `apply`-driven loops will grow the native stack where the interpreter would not.
- **Removing `apply` from the fallback set is the actual gate.** It stays in `ir.other_special_forms` with payload `true` until `emitCallNode` handles every shape — including `(apply f)` with no list, a non-`global_ref` operator, and `apply` reached as a *value* rather than a call head (`(map apply …)`), which must keep falling back. Flipping the payload before all of those work reintroduces #1799, so the #1799 regression tests (`tests/scheme/compile/native-apply-lexical-scope-1799.sh`) are the guard — and they were written to compare compiled output against the interpreter rather than to assert a fallback, precisely so they stay valid across this change.
- **`call/cc` and friends are out of scope.** They were fixed by the same list change but need continuation machinery the native backend does not have; they should keep declining.

## Verification

`tests/scheme/compile/native-apply-lexical-scope-1799.sh` already covers ten shapes (parameter, `let`-bound name, bare top-level `let`, computed operator, fixed-args-plus-list, rest parameter, recursion, closure capture, global-not-clobbered, summary statistics), each asserting the compiled binary agrees with the interpreter. It should pass unchanged, with the emit-level tests in `src/tests_native.zig` updated to assert the new native lowering instead of the fallback.

The `heavy-pure` / `heavy-apply` pair above is the perf check: they should converge.

baijum:
Closes #1803. Follow-up to #1799/#1798.

## What

One `apply` anywhere in a body no longer de-optimizes the whole enclosing function. `emitPassthrough` routes an `(apply …)` head inside a lexical scope to a new `emitApplyForm`, which emits the callee and fixed arguments natively and hands the final operand to a new C-ABI entry point:

```zig
pub export fn kaappi_apply(vm: ?*vm_mod.VM, callee: u64,
                           fixed_args: ?[*]const u64, n_fixed: u64,
                           list: u64) callconv(.c) u64
```

`kaappi_apply` is `primitives.applyFn` minus the argument shuffling — same `isProcedure`/`isNativeFn` validation, same tortoise-and-hare cycle detection (a circular list raises, never hangs), same `typeError` texts verbatim.

## Dispatch mirrors the interpreter case for case

The interpreter's own `apply` handling is position-dependent (`compiler.zig`/`compileApplyTail`), and the native dispatch replicates it exactly — including its quirk:

| shape | interpreter | native |
|---|---|---|
| tail + unshadowed + ≥2 operands | `tail_apply` opcode (ignores a top-level rebinding of `apply`) | `@kaappi_apply`, builtin semantics |
| tail + unshadowed + <2 operands | compile-time `InvalidSyntax` | abandons native compilation → same error |
| shadowed / non-tail rebound / non-tail <2 operands | ordinary call through the current binding | indirect call via `emitGlobalRef`'s lexical order |

An earlier draft honored the rebinding in tail position too ("more correct" per R7RS) — parity testing caught it diverging from the interpreter (99 vs 6), so it was reverted to bug-compatible parity: **zero behavior change on every shape, only performance changes.**

## The latent free-variable trap

Both free-variable walks (`nodeHasFreeVars`/`collectNodeFreeVars`) treated every `.passthrough` node as capture-free — sound only while every passthrough keyword also declined the enclosing scope. With `apply` out of `eval_fallback_form_names`, `(lambda (x) (apply + base x))` would have compiled the closure with an incomplete upvalue set and read `base` as a global — #1799's failure mode in a new spot. Both arms now walk apply operands through the existing scope-aware `FreeNameWalk`.

## Results

The issue's heavy pair (fixnum-safe variant, 3M iterations), compiled, 3 runs each:

| program | before | after |
|---|---|---|
| heavy-pure | 0.12 s | 0.12 s |
| heavy-apply | ~19x slower (per issue) | **0.12 s** |

`(define (s xs) (apply + xs))` now emits 1 native define and 0 eval fallbacks (was 0 and 2).

## Verification

- `tests/scheme/compile/native-apply-lexical-scope-1799.sh` — **passes unchanged**, as the issue designed it to.
- New `tests/scheme/compile/native-apply-lowering-1803.sh` — 11 behavior-parity cases: the heavy reproducer + its structural convergence check (zero eval fallbacks in the emitted IR), tail-apply root balancing over 2000 iterations, fixed args under `KAAPPI_GC_THRESHOLD=1`, shadowed parameter, tail-vs-non-tail rebinding, apply-as-value, cond-arm, and four error shapes with verbatim `typeError` detail (not-a-procedure, improper, circular, arity).
- `src/tests_native.zig` — emit-level tests updated from assert-the-decline to assert-the-lowering; `must_have`/`must_not_have` name-set guards updated (`apply` now must stay OUT).
- `native-bootstrap-callbacks-1376.sh` case 8 relied on `apply` forcing the eval fallback to exercise bytecode `tail_apply` against a NativeClosure; a semantically transparent `(letrec ())` wrapper restores that documented coverage (`do` had already stopped forcing it in #1496).
- Full `zig build test` and `bash tests/scheme/run-all.sh`: **1998 pass, 0 fail**. Cross-compiles clean for x86_64-linux, aarch64-windows, and wasm.

## Notes

- `call/cc`/`call-with-values`/`eval` keep declining, per the issue's scope.
- A tail `(apply f xs)` is still not a real tail call across the runtime boundary (`callWithArgs` re-entry) — verified to be the same pre-existing divergence class as computed-operator tail calls through `kaappi_call_scheme`, which overflow at the same depths on unmodified main. Documented in `docs/dev/llvm-backend.md`.
- Two pre-existing bugs surfaced while validating against clean main (99dc2320), both reproduced without this change: macro uses inside natively-lowered `let` bodies miscompile to global lookups, and the issue's own `heavy-pure.scm` (no `apply` at all) segfaults at n=3M when the accumulator crosses into bignum range. Neither is touched here; both are being filed separately.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -175,8 +175,11 @@ FFI-`dlopen` directory, so an archive placed there is invisible to
 **Features compiled natively:** arithmetic, comparisons, if/and/or/when/unless,
 let/let*, cond/case/do, lambda (with closures and variadic parameters),
 self-tail-call optimization (compiled as loops), tail calls to other native
-functions. Forms not yet compiled natively (letrec, named-let, guard,
-quasiquote, `apply`, `call/cc`, `call-with-values`, `eval`, …) run through
+functions, and `apply` (lowered to the argument-splicing `@kaappi_apply`
+runtime call, so an enclosing function keeps its native compilation —
+kaappi#1803; the dispatch mirrors the interpreter's tail/non-tail/shadowed
+cases exactly). Forms not yet compiled natively (letrec, named-let, guard,
+quasiquote, `call/cc`, `call-with-values`, `eval`, …) run through
 `kaappi_eval` at runtime — and because that eval resolves names in the *global*
 environment, any lexical scope containing one (the enclosing `define`/`lambda`
 frame or `let`) declines native compilation as a whole rather than splitting
--- a/docs/dev/llvm-backend.md
+++ b/docs/dev/llvm-backend.md
@@ -394,14 +394,15 @@ past the fixed arity before branching, so variadic named functions loop too
 - contains an eval-fallback form (`letrec`, `guard`, named `let`, …), or a
   `cond`/`case`/`do` whose clauses themselves reach one (kaappi#1496);
 - contains a keyword-only special form that reaches the interpreter as a whole
-  `passthrough` — `apply`, `call/cc`,
-  `call-with-current-continuation`, `call-with-values`, `eval` (kaappi#1799);
+  `passthrough` — `call/cc`, `call-with-current-continuation`,
+  `call-with-values`, `eval` (kaappi#1799; `apply` left this set when it got
+  its own native lowering, kaappi#1803 — see below);
 - contains an internal `define` (the closure tier sets up no locals scope for
   it), or a rest parameter that is captured and mutated (no box model yet); or
 - captures an unmutated `let`-local, or a rest parameter, that has no copyable
   slot.
 
-### Why `apply` declines the whole enclosing scope (kaappi#1799)
+### Why `call/cc` & friends decline the whole enclosing scope (kaappi#1799)
 
 All four native-compilation gates — the two closure tiers,
 `tryCompileDefineFunction`, and `emitLet` — ask
@@ -414,16 +415,16 @@ compiles natively, its body's `passthrough` is serialized and run by
 `undefined variable 'xs'` — or, worse, silently read an unrelated global of the
 same name.
 
-The five keywords above were missing because the `passthrough` **node** is
-`.capability = .native` in `llvm_node_table`, which is right for the one shape
-`emitPassthrough` compiles (a `(define (f …) …)` shorthand) and wrong for every
-other. They are contributed instead by `ir.other_special_forms`, whose `bool`
-payload records exactly "an evaluated form headed by this keyword becomes a
-passthrough the interpreter runs". `else`, `=>`, `_`, `...`,
-`unquote`/`unquote-splicing` and the keywords `lowerFormWithMacros` intercepts
-earlier are deliberately `false`: `sexprNeedsEvalFallback` recurses blindly into
-sub-forms, so listing `else` would reject every natively lowered `cond`/`case`
-with an else clause.
+The five keywords (then including `apply`) were missing because the
+`passthrough` **node** is `.capability = .native` in `llvm_node_table`, which
+is right for the one shape `emitPassthrough` compiles (a `(define (f …) …)`
+shorthand) and wrong for every other. They are contributed instead by
+`ir.other_special_forms`, whose `bool` payload records exactly "an evaluated
+form headed by this keyword becomes a passthrough the interpreter runs".
+`else`, `=>`, `_`, `...`, `unquote`/`unquote-splicing` and the keywords
+`lowerFormWithMacros` intercepts earlier are deliberately `false`:
+`sexprNeedsEvalFallback` recurses blindly into sub-forms, so listing `else`
+would reject every natively lowered `cond`/`case` with an else clause.
 
 Publishing the frame's bindings as globals instead (`bindParamsAsGlobals`,
 the kaappi#1410 mechanism the `letrec`/`let` fallbacks use) is **not** an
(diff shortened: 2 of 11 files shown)


baijum:
@coderabbitai review

Issue #1831, opened 2026-07-29
cadar (specifically) fails when called from a helper function invoked during er-macro-transformer expansion


baijum:
## Summary

The composed accessor `cadar` fails silently when called from inside a
plain (non-macro) helper function that is itself invoked from an
`er-macro-transformer`'s expansion-time code (SRFI 211). The call never
completes, and the whole macro expansion is later reported as a generic
`invalid syntax` compile error with no other diagnostic. The two
semantically-identical unrolled spellings, `(car (cdr (car x)))` and
`(cadr (car x))`, both work correctly in the exact same position.

This is specific to `cadar` as a *named primitive called from a nested
helper function* -- calling `cadar` directly inline in the transformer
body works fine, and `caar`/`cadr`/`cddr`/`cdddr` used the same way (from
inside a nested helper) show no issue.

## Reproduction

```scheme
;; lib/srfi/repro.sld
(define-library (srfi repro)
  (import (scheme base) (srfi 211 explicit-renaming))
  (export probe probe-unrolled)
  (begin
    (define (lookup-second field alist)
      (cond
        ((null? alist) #f)
        ((equal? field (cadar alist)) (caar alist))
        (else (lookup-second field (cdr alist)))))

    (define (lookup-second-unrolled field alist)
      (cond
        ((null? alist) #f)
        ((equal? field (cadr (car alist))) (caar alist))
        (else (lookup-second-unrolled field (cdr alist)))))

    (define-syntax probe
      (er-macro-transformer
       (lambda (form rename compare)
         (list (rename 'quote) (lookup-second 'b '((x a) (y b)))))))

    (define-syntax probe-unrolled
      (er-macro-transformer
       (lambda (form rename compare)
         (list (rename 'quote) (lookup-second-unrolled 'b '((x a) (y b)))))))))
```

```scheme
;; main.scm -- fails
(import (scheme base) (scheme write) (srfi repro))
(write (probe))
```

```
main.scm:2:1: compile error[KP2001]: invalid syntax
```

```scheme
;; main2.scm -- works, prints y
(import (scheme base) (scheme write) (srfi repro))
(write (probe-unrolled))
```

Run with:
```
kaappi --lib-path <dir containing lib/srfi/repro.sld> main.scm
```

## Isolation notes

- `(lookup-second 'b '((x a) (y b)))` needs `equal?` to actually reach the
  `cadar` branch (i.e. the field being looked up isn't matched by the
[...]

Issue #1726, opened 2026-07-22
Library import silently lets colliding export names resolve to whichever import came last


baijum:
## Summary

Importing two libraries that export the same identifier does not error or warn — it silently resolves to whichever library was imported *last* in the import-list order, with zero diagnostic either way. Discovered independently by two different implementers this session while working on separate SRFI libraries (SRFI 29 and SRFI 94), both of which export a name (`format`, and separately a name matching another SRFI) that collides with a sibling library.

## Repro

\`\`\`scheme
(import (srfi 28) (srfi 29))   ; both export `format` -- (srfi 29)'s wins, no diagnostic
(import (srfi 29) (srfi 28))   ; reversed order -- (srfi 28)'s wins instead, still no diagnostic
\`\`\`

Verified empirically: reversing the import order flips which binding wins, confirming this is "last import wins" rather than, say, "first wins" or a deterministic tie-break.

## Root cause

\`importBinding\` in \`src/vm_library.zig\` is a plain, unconditional hashmap \`put\` — it has no check for whether the name already has a binding from an earlier import in the same form.

## Why this matters

R7RS's own semantics (7.1.4, library-declaration / import-set) are stricter: importing the same identifier from two different libraries with no explicit \`rename\`/\`only\`/\`except\` disambiguation is specified as *an error*, not "last one wins." Kaappi currently makes this a silent, order-dependent behavior instead — a real conformance gap, and one that's easy to trip over precisely in an ecosystem with many small libraries (like Kaappi's own 130+ portable SRFIs) that don't all coordinate their export names.

## Suggested fix

In \`importBinding\` (or its caller in \`vm_library.zig\`), detect when a name being imported already has a binding introduced by an *earlier import in the same import-list* (not a pre-existing global from the user's own code, which legitimately gets shadowed) and signal a clear compile/read-time error naming both source libraries and the colliding identifier, matching R7RS's "an error" language for this case.

## Notes

Not a blocker for either of the SRFIs that surfaced it (both worked around it by choosing distinguishable names / documenting the collision risk in their own header comments). Medium priority: a real, standards-mandated diagnostic gap, independently rediscovered twice in one session, in a codebase that increasingly builds on `(import ...)`-heavy composition of small libraries.

Issue #1715, opened 2026-07-21
let-values intermittently raises spurious apply error with variadic procedures


baijum:
## Summary

`let-values` intermittently raises a spurious `apply: expected proper list, got #<record_instance>` error when consuming values from a variadic (rest-arg) procedure, in a library that also defines other variadic procedures. `call-with-values` does not trigger the same failure in the same situations.

## Repro context

Found while implementing SRFI 101 (Purely Functional Random-Access Pairs and Lists, #1691). Consuming `list-ref/update`'s two return values via `let-values` raised the error; reproduced independently in a minimal, unrelated external library (a variadic `list`-alike plus a separate `values`-returning procedure — nothing SRFI-101-specific), suggesting a genuine interpreter bug rather than anything in that library's own logic.

## Impact

Silent-ish (raises an error rather than corrupting data), but surprising and hard to diagnose — the error message references `apply` and a record instance, giving no hint that `let-values` itself is the actual source, or that the triggering condition is "a variadic procedure is in scope."

## Suggested fix

Needs a minimal, isolated reproduction (not yet captured in this issue — the original repro lived inside a larger library) to bisect exactly which combination of variadic-procedure shape and `let-values` usage triggers it, then trace how `let-values`'s compiled expansion invokes `apply`/`call-with-values` internally.

## Workaround used

Used `call-with-values` directly instead of `let-values` throughout the affected library, with the caveat documented in the library's header comment for future maintainers.

Issue #1714, opened 2026-07-21
case-lambda's arity dispatch calls global length non-hygienically


baijum:
## Summary

`case-lambda`'s compiled arity-dispatch logic appears to call the top-level global `length` procedure directly rather than through a hygienic reference. If a library legitimately shadows `length` (e.g. to provide its own O(log n) `length` for a non-standard list-like type), any `case-lambda` defined in that library's scope breaks when called with a native argument list, since dispatch resolves to the *user's* shadowed `length` instead of the real list-length primitive `case-lambda` needs internally.

## Repro context

Found while implementing SRFI 101 (Purely Functional Random-Access Pairs and Lists, #1691), which necessarily shadows `length` (via `except`/local `define`) to provide its own random-access-list version. A `case-lambda`-based `make-list` in that library's scope failed until rewritten with plain rest-arg dispatch instead of `case-lambda`.

## Suggested fix

`case-lambda`'s generated dispatch code should reference the list-length operation through the same hygienic-rename mechanism used elsewhere in the expander/compiler, or use an internal-only primitive not reachable by user shadowing, rather than resolving `length` through the ambient global scope.

## Workaround used

Avoided `case-lambda` for the affected procedure, using a plain `. rest` lambda with manual arity dispatch instead.

PR #1859, opened 2026-07-30
Stop (scheme base) from exporting %-prefixed internals

### LINKED ISSUE #1856: (scheme base) exports 18 %-prefixed internals, so a user library defining one can no longer be imported
## Summary

`(scheme base)` exports `%length` — the internal alias added in v0.22.0 so `case-lambda`'s arity dispatch survives user shadowing of `length`. Because v0.22.0 also began enforcing R7RS 5.2's "same identifier from two libraries with different bindings is an error" (#1726), any user library that defines its own `%length` and imports `(scheme base)` now fails to load entirely.

The two changes are individually correct; together they took a name away from users without anybody noticing. `%length` is one of **18** `%`-prefixed primitives currently in `(scheme base)`'s export set.

## Reproduction

`lib/mylib/ffi.sld`:

```scheme
(define-library (mylib ffi)
  (import (scheme base))
  (export %length)
  (begin
    (define (%length s) (string-length s))))
```

`lib/mylib.sld`:

```scheme
(define-library (mylib)
  (import (scheme base) (mylib ffi))
  (export byte-length)
  (begin
    (define (byte-length s) (%length s))))
```

`main.scm`:

```scheme
(import (scheme base) (scheme write) (mylib))
(write (byte-length "hi")) (newline)
```

```
$ kaappi --lib-path ./lib main.scm
main.scm:1: error[KP2001]: identifier '%length' is imported with different bindings inside library (mylib)'s own import declaration, from both (scheme base) and (mylib ffi) -- R7RS 5.2: "it is an error to import the same identifier more than once with different bindings"; disambiguate with only/except/rename/prefix
main.scm:1:54: error[KP3001]: undefined variable 'byte-length'
$ echo $?
1
```

Control: renaming the binding `%length` -> `%strlen` in both files, changing nothing else, makes the identical program print `2` and exit 0. Verified on v0.22.0 (`build_id ca18ae6`, aarch64-macos).

## Root cause

`src/primitives.zig:225`:

```zig
.{ .name = "%length", .func = &length, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_base) },
```

`.libs = LS.initOne(.scheme_base)` puts it in the public export set rather than keeping it internal to the compiler. `src/compiler_advanced.zig:876` is the only consumer that needs it (`gc.allocSymbol("%length")` for the dispatch path).

## Scope

18 `%`-prefixed names are exported from `(scheme base)` today, so the same trap exists for each:

```
%default-random-source  %host-big-endian?  %length  %make-promise-lazy
%make-record  %make-record-type  %parameter-convert  %parameter-set!
%random-port-make-from-seed  %random-port-make-from-state
%random-port-make-randomized  %random-port-state  %random-port?
%record-ref  %record-set!  %record?  %rs-next-int  %rs-next-real
```

Found with:

```bash
grep -rhn '\.name = "%' src/primitives*.zig | grep 'scheme_base'
```

The `%` prefix is a long-standing convention for "private helper" across this codebase and the portable SRFI libraries, so user code has good reason to assume the namespace is theirs.

## Question to settle

Is membership in `(scheme base)`'s export set intended for these, or incidental?

- If **incidental** (likely, at least for `%length`): give the compiler a way to reach a primitive that is registered but not exported, so `%`-prefixed internals stop occupying user namespace. That fixes all 18 at once.
- If **intended** for some (the record and random-source helpers are plausibly reachable by portable library code on purpose): those should stay, and it is worth documenting that `%`-prefixed names in `(scheme base)` are reserved — the current failure mode gives a user no hint that the name was ever spoken for.

## Real-world impact

This is not hypothetical. It broke the documented C-extension walkthrough at https://kaappi-lang.org/guide/c-extensions/, whose example FFI library exported `%length`; the entire walkthrough failed to import on v0.22.0. The docs are unblocked — kaappi/kaappi.github.io@4d60e00 renamed the example binding to `%strlen` — so nothing here is waiting on this issue.

baijum:
Closes #1856.

## The bug

`(scheme base)` exported **22** `%`-prefixed internal primitives (the issue counted 18 — four more, `%record?/inherit` and friends, were tagged through a `BASE` alias in `primitives_srfi237.zig` and slipped the issue's grep). Because v0.22.0 also began enforcing R7RS 5.2 (#1726), a user library that defined any of those names and imported `(scheme base)` failed to load outright. `%` is this codebase's own private-helper marker, so user code has good reason to treat that namespace as its own — the documented C-extension walkthrough, whose example FFI library exported `%length`, was a real casualty.

## Answering the issue's question

The issue asked whether export-set membership was intended or incidental. **Incidental for all 22** — but they split into two groups by who names them, and the fix treats them differently:

| Who names it | Where it went |
|---|---|
| Only Zig-generated code — `%make-promise-lazy`, `%parameter-set!`, `%parameter-convert`, `%record?/inherit`, `%record-ref/inherit`, `%record-set!/inherit`, `%record-split-args` | `.internal`: registered in `vm.globals`, exported by nothing |
| A portable `.sld` names it in Scheme source — SRFI 27's random-source accessors, SRFI 74's endianness probe, SRFI 271's random ports, the record substrate SRFI 57/131/136/150/237 build on | `.internal` **plus** a new `(kaappi primitives)` library those `.sld`s now import |

A comptime check in `primitives.zig` now rejects any `%` name tagged with a `scheme.*` library, so this can't drift back.

`(kaappi primitives)` is deliberately importable and deliberately not a stability promise: it exposes exactly what `(scheme base)` already exposed, but namespaced and opt-in, and it lets each `.sld` declare the dependency it actually has instead of relying on ambient visibility.

## Unexporting alone would have been a silent regression

A library defining its own `%record-ref` would now load — and its `define-record-type` accessors would then call it:

```scheme
(define-library (u rec)
  (import (scheme base))
  (export get)
  (begin
    (define (%record-ref a b c) 'user-version)
    (define-record-type point (make-point x y) point? (x point-x) (y point-y))
    (define (get) (point-x (make-point 7 8)))))
;; without the second half of this PR: (get) => user-version
```

So compiler-synthesized references now go through `Compiler.trueBuiltinRefOrSymbol` / `globals_mod.baseBindingSymbol`, resolving against a pristine startup snapshot (`LibraryRegistry.internal_bindings`) instead of `vm.globals` — the same mechanism #1715 already used for `let-values`. The user's `%record-ref` stays theirs; `define-record-type` gets the real one. Same for `case-lambda`, `parameterize`, and `delay`.

`%length` is **deleted** rather than relocated: `case-lambda`'s arity dispatch references `length`'s pristine `(scheme base)` binding directly. That is what the alias existed to approximate (#1714) and is strictly stronger — a top-level `(define (%length x) ...)` could overwrite the alias but cannot touch the export table. The name is now entirely free for users.

## Two adjacent fixes

- **`kaappi check` KP4001 noise.** The lint judged base-binding-prefixed names against `ir.globals`, which cannot see them. Already wrong for `let-values`' synthesized `call-with-values`; `case-lambda`/`define-record-type`/`parameterize`/`delay` would have made it common. Now skipped, exactly like the hygienic-prefix case beside it.
- **Globals drift guard.** It assumed every `.internal` spec is purged from globals by `vm_bootstrap.install()`, which only held while the two sets happened to coincide. It now tests the real invariant against `vm_bootstrap.internal_helpers`.

## Found along the way, not fixed here

`kaappi check lib/srfi/27.sld` failing during development exposed a **pre-existing** gap: a library-body *top-level* reference to a global the library hasn't imported raises "undefined variable", while the identical reference inside a lambda in the same library resolves. It reproduces with `cadar` on `main` (nothing to do with this PR) — `compileExpressionInEnv` sets `restricted_globals` on the outer function of every library-body form, which gates off the `vm.globals` fallback #1831 added. Filed separately; this PR sidesteps it by giving the portable `.sld`s explicit imports, which is better style regardless.

## Testing

- New Scheme regression test `tests/scheme/smoke/percent-name-user-library-1856.scm` with `lib1856/` fixtures — the issue's own multi-file reproduction, plus the shadowing cases above at both library and program top level.
- New unit tests: no `scheme.*` library exports a `%` name; `(kaappi primitives)` exports what the `.sld`s import; a user library may define `%length` alongside `(scheme base)`.
- `tests_advanced.zig`'s #1714 test now also rebinds `%length` at top level, which the old alias could not survive.
- Full suite: **2014 Scheme tests pass, 0 fail** (was 2001 with 12 failing mid-refactor); `zig build test` green; `--sandbox`, `zig build wasm`, and `-Dtarget=x86_64-linux` all verified.

## Note for the docs repo

kaappi/kaappi.github.io@4d60e00 renamed the C-extension walkthrough's `%length` to `%strlen` to work around this. That workaround is no longer needed — `%length` is a free name again — though nothing breaks if it stays.

### DIFF
--- a/.claude/skills/add-builtin/SKILL.md
+++ b/.claude/skills/add-builtin/SKILL.md
@@ -25,13 +25,20 @@ fn myProc(args: []const Value) PrimitiveError!Value {
 .{ .name = "my-proc", .func = &myProc, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_base) },
 ```
 
-An **internal helper** — a `%`-prefixed name only compiler-generated code or
-a portable `.sld` calls — takes `.libs = primitives.INTERNAL` instead:
-registered in `vm.globals`, exported by nothing, so it doesn't reserve the
-name against every user library that imports `(scheme base)` (kaappi#1856; a
-comptime check rejects a `%` name tagged `scheme.*`). Synthesize *references*
-to one with `Compiler.trueBuiltinRefOrSymbol` /
-`globals_mod.baseBindingSymbol`, never a bare `allocSymbol("%foo")`.
+An **internal helper** — a `%`-prefixed name no user program should have to
+avoid — never takes a `scheme.*` tag, which reserves the name against every
+user library that imports it (kaappi#1856; a comptime check rejects this).
+Two cases:
+
+- Only compiler-generated code names it: `.libs = primitives.INTERNAL` —
+  registered in `vm.globals`, exported by nothing.
+- A portable `.sld` names it in its own Scheme source:
+  `.libs = primitives.INTERNAL_PUBLIC`, which adds `(kaappi primitives)` so
+  the `.sld` can import it (`lib/srfi/27.sld` and the record SRFIs do).
+
+Synthesize *references* to either with `Compiler.trueBuiltinRefOrSymbol` /
+`globals_mod.baseBindingSymbol`, never a bare `allocSymbol("%foo")` — that
+picks up a user binding of the same name.
 See `docs/dev/adding-features.md`.
 
 3. **Add a test** in `src/vm.zig` test section:
--- a/.claude/skills/add-builtin/SKILL.md
+++ b/.claude/skills/add-builtin/SKILL.md
@@ -17,11 +17,22 @@ fn myProc(args: []const Value) PrimitiveError!Value {
 }
 ```
 
-2. **Register it** in `registerAll()` in `src/primitives.zig`:
+2. **Add a spec entry** to the file's `specs` table. One entry carries the
+   name, function, arity, and the libraries that export it — `library.zig`
+   derives every export set from these tags, so there is no second list:
 
 ```zig
-try reg(vm, "my-proc", &myProc, .{ .exact = 1 });  // or .{ .variadic = N }
+.{ .name = "my-proc", .func = &myProc, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_base) },
 ```
+
+An **internal helper** — a `%`-prefixed name only compiler-generated code or
+a portable `.sld` calls — takes `.libs = primitives.INTERNAL` instead:
+registered in `vm.globals`, exported by nothing, so it doesn't reserve the
+name against every user library that imports `(scheme base)` (kaappi#1856; a
+comptime check rejects a `%` name tagged `scheme.*`). Synthesize *references*
+to one with `Compiler.trueBuiltinRefOrSymbol` /
+`globals_mod.baseBindingSymbol`, never a bare `allocSymbol("%foo")`.
+See `docs/dev/adding-features.md`.
 
 3. **Add a test** in `src/vm.zig` test section:
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -988,17 +988,18 @@ map.deinit();  // no allocator arg needed
    }
    ```
 
-2. Register in the file's `registerXxx` function:
+2. Add one entry to the file's `specs` table — name, function, arity, and the
+   libraries that export it. `registerAll` walks `all_specs` and
+   `library.zig` derives every export set from the same tags, so this is the
+   whole registration; there is no second list and no manual `reg()` call:
 
    ```zig
-   try primitives.reg(vm, "my-proc", &myProc, .{ .exact = 1 });
+   .{ .name = "my-proc", .func = &myProc, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_base) },
    ```
 
    Arity: `.{ .exact = N }` for fixed, `.{ .variadic = N }` for N minimum args.
 
-3. Add to the file's `specs` table with the exporting libraries in `.libs`
-   (`library.zig` derives every export set from these tags — there is no
-   second list). An internal helper — a `%`-prefixed name only
+3. An internal helper — a `%`-prefixed name only
    compiler-generated code or a portable `.sld` calls — takes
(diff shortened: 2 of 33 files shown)


baijum:
Thanks — two of three taken, one declined.

**Taken: `.claude/skills/add-builtin/SKILL.md` — `INTERNAL` vs `INTERNAL_PUBLIC`.** Correct, the skill only mentioned `INTERNAL`. It now spells out both cases (compiler-generated-only → `INTERNAL`; a portable `.sld` names it in Scheme source → `INTERNAL_PUBLIC`, which adds `(kaappi primitives)`), matching what `CLAUDE.md` and `docs/dev/adding-features.md` already say.

**Taken: `CLAUDE.md` — the stale `registerXxx` step.** Also correct, and worse than flagged: `reg()` has no caller outside `registerAll`/`registerSandboxed` looping over `all_specs`, so step 2 described a workflow that no longer exists and contradicted the step beside it. Steps 2 and 3 are now one spec-table step. `src/CLAUDE.md` had the same stale sentence ("Each `primitives_*.zig` has a `registerXxx` function that calls `primitives.reg()`") and is fixed too.

**Declined: migrating the two tests to `th.TestContext`.** Both files use the manual `memory.GC.init` + `th.makeTestVM` form exclusively — 49 occurrences in `tests_advanced.zig`, 48 in `tests_libraries.zig`, and zero `th.TestContext` in either. The `tests_advanced.zig` site is a pre-existing test I only added two `eval` lines to, so converting it would be unrelated churn, and converting only my new `tests_libraries.zig` tests would make them the odd ones out in their own file. Matching the surrounding code is the convention here; a file-wide migration would be its own change.

Docs-only follow-up pushed once the current CI run finishes, to avoid restarting it.

PR #1883, opened 2026-07-31
Let a record's constructor be named `fields` or `parent` (#1882)

### LINKED ISSUE #1882: R6RS clause-syntax detection misparses valid R7RS records whose constructor is named `fields`, `parent`, …
Found while working #1880/#1881, which touched these functions but did not
change this behavior — it predates that PR, arriving with the SRFI 237 slice.

## The bug

SRFI 237's R6RS clause syntax is **ambient**: `(scheme base)`'s
`define-record-type` accepts it with no `(import (srfi 237))`. That part is
consistent with house style — SRFI 46/149/213 recognition is ambient too, with
the `.sld` export as declaration of intent.

The problem is that it is **not a conservative extension**. The discriminator
tests *the constructor's name* against 8 clause keywords, so a valid
R7RS-small program is misparsed if its constructor happens to be called one of
them:

```scheme
;; Valid R7RS-small: a record whose constructor is named `fields`.
(define-record-type point (fields x y) point? (x point-x) (y point-y))
(display (point-x (fields 1 2)))
```

```
error[KP2001]: invalid syntax
error[KP3001]: undefined variable 'fields'. Did you mean 'yield'?
```

The affected names are the 8 in `isR6RSClauseKeyword`
(`src/vm_records.zig:275`): `fields`, `parent`, `protocol`, `sealed`,
`opaque`, `nongenerative`, `generative`, `parent-rtd`. Realistically `fields`
and `parent` are the ones anyone would actually write; verified reproducing
with `fields`, `parent` and `sealed`.

Two things make it worse than the hit rate suggests:

- It reaches programs that **never use SRFI 237 at all** — the import is not
  required, so nothing in the program signals that the R6RS grammar is in
  play.
- The diagnostic is a bare `KP2001: invalid syntax` with no hint that a name
  collision caused it. The follow-on `undefined variable 'fields'` and its
  `Did you mean 'yield'?` suggestion actively point away from the cause.

## Why it happens

`looksLikeR6RSClauseSyntax` (`src/vm_records.zig:288`) checks only whether the
head of the form's 2nd element is a clause keyword — and in R7RS syntax that
element is `(<constructor> <field name>*)`, so the head it reads is the
constructor's name. The existing comment names the case:

> a constructor literally named e.g. "parent" would misdetect, but that's an
> extreme, self-inflicted edge case

Fair that constructors are conventionally `make-X`, but `fields` and `parent`
are ordinary words, and there is a cheap fix that removes the judgment call.

## Proposed fix

The two grammars differ at **position 3**, unambiguously:

| | 3rd element (index 2) |
|---|---|
| R7RS-small | always `<predicate>` — a bare **symbol**, and the form always has ≥3 arguments |
| R6RS / SRFI 237 | a `<record clause>`, which is always a **list** — or absent, when there are ≤1 clauses |

So the fix is to **add** a second condition rather than replace the existing
one: treat the form as R6RS clause syntax only when the 2nd element's head is a
clause keyword **and** the 3rd element is absent or a list.

That is strictly narrower than today's rule, which is what makes it safe:

- No currently-accepted R6RS form changes — every clause is a list, so a
  genuine R6RS form always has a list-or-nothing there. Checked against every
  R6RS-clause use in `tests/scheme/srfi/srfi237.scm` and `srfi240.scm`; all of
  them qualify.
- The shadow case above flips back to the R7RS path, because `point?` is a
  symbol.
- Malformed input keeps its current routing, so no diagnostic regresses. In
  particular `(define-record-type point (make-point x y))` still fails on the
  R7RS path today and would continue to, since its head is not a keyword.

## Test

`tests/scheme/compliance/` — a record whose constructor is named `fields`
(and one named `parent`) constructs and reads back correctly under
`(scheme base)` alone, alongside an existing SRFI 237 form to pin that the
narrowing did not break detection.

Related: #1880 / #1881 (the diagnostics on these same two functions),
kaappi#1736 (the SRFI 237 slice this arrived with).

baijum:
Fixes #1882.

## The bug

SRFI 237's R6RS clause grammar is ambient — `(scheme base)`'s
`define-record-type` accepts it with no `(import (srfi 237))` — so the two
syntaxes have to be told apart structurally. `looksLikeR6RSClauseSyntax` read
only the head of the form's 2nd element, which in R7RS syntax is the
*constructor's* name:

```scheme
(define-record-type point (fields x y) point? (x point-x) (y point-y))
(display (point-x (fields 1 2)))
```

```
error[KP2001]: invalid syntax
error[KP3001]: undefined variable 'fields'. Did you mean 'yield'?
```

Nothing in that program signals that the R6RS grammar is in play, and the
follow-on `undefined variable` with its `Did you mean 'yield'?` points away
from the cause. All 8 keywords were affected (`fields`, `parent`, `protocol`,
`sealed`, `opaque`, `nongenerative`, `generative`, `parent-rtd`); `fields` and
`parent` are the ones anyone would realistically write.

## The fix

The 3rd element separates the grammars unambiguously:

| | 3rd element |
|---|---|
| R7RS-small | always the `<predicate>` — a bare **symbol**, always present |
| R6RS / SRFI 237 | a `<record clause>`, always a **list** — or absent, with ≤1 clause |

So the keyword test gets a second condition rather than a replacement: a
**symbol** there means R7RS and nothing else. That is a pure narrowing — only
the symbol case is taken away — which is what makes it safe.

Verified by A/B against a binary built from the parent commit:

- The repro above, all 8 keywords as constructor names, and a zero-field
  record whose *predicate* is `sealed`: rejected before, correct after.
- Every genuine R6RS form in `srfi237.scm`/`srfi240.scm` still routes to the
  R6RS path (single-clause, multi-clause, and `parent` inheritance are all
  covered by the new tests).
- 10 malformed forms — missing predicate, non-symbol predicate, improper
  tail, undeclared constructor field, `parent-rtd`, duplicate fields — produce
  byte-identical diagnostics before and after.

## Two paths the issue didn't mention

The same predicate guards two other call sites, and both were also affected:

- **In a body**, a misdetection made the internal-define scan `break`, so
  sibling `define`s written *after* the record lost their forward
  declarations — mutual recursion among them failed with `undefined variable`
  even though the record itself still worked.
- **In a library body**, where the R6RS grammar is rejected outright as a
  documented top-level-only feature, the whole library failed to load and
  every importer saw `undefined variable` for every name it exports.

Both are covered by the new tests.

## Tests

- `tests/scheme/compliance/record-ctor-clause-keyword-1882.scm` (18
  assertions) with the library fixture `lib1882/rec.sld` — all 8 keywords as
  constructor names, a mutable field, the body-scan case, the library-body
  case, and R6RS single-clause/multi-clause/inheritance forms to pin that
  detection did not regress. Fails 8 assertions on the parent commit.
- Two unit tests in `src/tests_records.zig`; the R7RS one fails without the
  fix, the R6RS one passes either way by design (it is the "still detected"
  pin).
- `zig build test` and `bash tests/scheme/run-all.sh` green: 623 Scheme files,
  1395 R7RS, 0 failures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   `parent`, or another R6RS clause keyword** (#1882). SRFI 237's R6RS clause
   grammar is ambient — `(scheme base)`'s `define-record-type` accepts it with no
   `(import (srfi 237))` — so the two syntaxes are told apart structurally, and
-  the test read only the head of the form's 2nd element, which in R7RS syntax is
-  the *constructor's* name. `(define-record-type point (fields x y) point? (x
-  point-x) (y point-y))` was therefore parsed as R6RS and rejected with a bare
+  the syntax detector inspected only the head of the form's 2nd element, which
+  in R7RS syntax is the *constructor's* name. `(define-record-type point (fields
+  x y) point? (x point-x) (y point-y))` was therefore parsed as R6RS and
+  rejected with a bare
   `KP2001: invalid syntax`, in a program with nothing to suggest the R6RS
   grammar was in play; the follow-on `undefined variable 'fields'` and its `Did
   you mean 'yield'?` pointed away from the cause. The 3rd element now decides:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- **A valid R7RS record was rejected when its constructor was named `fields`,
+  `parent`, or another R6RS clause keyword** (#1882). SRFI 237's R6RS clause
+  grammar is ambient — `(scheme base)`'s `define-record-type` accepts it with no
+  `(import (srfi 237))` — so the two syntaxes are told apart structurally, and
+  the test read only the head of the form's 2nd element, which in R7RS syntax is
+  the *constructor's* name. `(define-record-type point (fields x y) point? (x
+  point-x) (y point-y))` was therefore parsed as R6RS and rejected with a bare
+  `KP2001: invalid syntax`, in a program with nothing to suggest the R6RS
+  grammar was in play; the follow-on `undefined variable 'fields'` and its `Did
+  you mean 'yield'?` pointed away from the cause. The 3rd element now decides:
+  R7RS always has one and it is always the bare-symbol `<predicate>`, while an
+  R6RS `<record clause>` is always a list — or absent, when there is at most one
+  clause. That is a pure narrowing, so no R6RS form and no malformed form
+  changes path, and no diagnostic changes. In a body the same misdetection also
+  aborted the internal-define scan, so sibling `define`s written after the
+  record lost their mutual visibility; in a library body, where the R6RS grammar
+  is rejected outright, the whole library failed to load.
 - **Native backend: a nested `let` that fell back to the interpreter lost the
   enclosing `let`'s bindings** (#1862). When a `let` inside another `let` gave up
   on native compilation mid-emission — more than 32 bindings, more head
--- a/src/tests_records.zig
+++ b/src/tests_records.zig
@@ -3,6 +3,7 @@ const std = @import("std");
 const th = @import("testing_helpers.zig");
 const types = @import("types.zig");
 const memory = @import("memory.zig");
+const vm_mod = @import("vm.zig");
 
 test "define-record-type basic" {
     var gc = memory.GC.init(std.testing.allocator);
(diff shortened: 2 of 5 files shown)


baijum:
Took all three findings in `37dea4de`:

- **Malformed-form coverage** (the substantive one): added `malformed record forms still fail after the #1882 narrowing` — six shapes near the new boundary. Worth noting what it can and can't do: it passes with *and* without the fix, by design. The narrowing's hazard isn't that a malformed form changes diagnostic (both readings of the rule land on `KP2001` either way — verified byte-identical A/B on 10 forms against a binary built from the parent commit) but that one could now reach the other parser and be *accepted*. That's what these pin.
- **`th.TestContext`**: switched both new tests. The other 8 tests in the file use the older raw `GC.init` + `makeTestVM` pattern, but `src/CLAUDE.md` asks for `TestContext` on multi-eval tests and it's a strict simplification here.
- **Changelog wording**: "the syntax detector inspected only the head".

Mutation-checked after the refactor: reverting `src/vm_records.zig` to the parent commit still fails exactly the R7RS test and leaves the R6RS and malformed tests green.

Issue #1509, opened 2026-07-13
kaappi test: first-class test runner with SRFI-64 aggregation, --json, --seed


baijum:
## Context

Kaappi's tests today run through per-repo shell scripts (`tests/scheme/run-all.sh`, ecosystem `kaappi --lib-path ./lib tests/test-x.scm`). SRFI-64 is already the standard harness across the suite, and `--coverage` already reports per-procedure coverage. What's missing is a first-class runner that an agent (or CI) can drive and parse without scraping stdout.

## What

`kaappi test [paths...]`:

- **Discovery**: default to `tests/**/*.scm` (or explicit paths), honoring `--lib-path`.
- **SRFI-64 aware**: aggregate pass/fail/skip counts across files from the SRFI-64 runner itself, not by scraping output text.
- **`--json`**: one result object per test file plus a summary object — suite name, file, pass/fail/xfail counts, each failure with test name, expected/actual, and source location, plus duration. JSON Lines.
- **`--seed N`**: seed SRFI-27 random sources deterministically, and print the effective seed on *every* run so any failure is reproducible.
- **Exit code**: nonzero iff there are failures.
- Stretch (separate PR): parallel execution across files — a natural fibers (KEP-0001) dogfood.

## Acceptance criteria

- [ ] `kaappi test` discovers and runs SRFI-64 suites with an aggregated summary
- [ ] `--json` output documented and covered by a shell test using a real JSON parser
- [ ] `--seed` reproducibility test (same seed → same results)
- [ ] `tests/scheme/run-all.sh` can delegate to it (or is documented as the legacy path)
- [ ] Works unchanged on ecosystem repos (`kaappi test --lib-path ./lib`)

## Notes

Affected-test selection (`--changed`) is tracked separately — the runner must not block on dependency tracking.

Part of #1503.

PR #1887, opened 2026-07-31
Run the test suites in parallel, and stop sleeping through them

baijum:
## Summary

`bash tests/scheme/run-all.sh` took **17.8 minutes** on a 4-core box to do about
90 seconds of actual work. Two independent causes, both fixed here.

**The poll interval.** `wait_with_timeout` polled each spawned child with
`sleep 1`, so every one of the ~620 spawned units cost a full second of wall
clock no matter how fast it really was — and 381 of the 566 `.scm` files finish
in under 50ms. That single sleep was **~93% of the script's runtime**. The tick
is now 0.05s, counted in 20ths of a second so the `$secs` timeout keeps its
exact meaning, with an integer fallback if a platform's `sleep` rejects
fractions.

**No parallelism.** Each `.scm` file is already a fresh interpreter with no
shared state, so `run_suite` now dispatches `KAAPPI_TEST_JOBS` files at a time
(default: one per CPU; `=1` restores strictly-sequential). Workers write a
verdict to a slot and the parent tallies and prints afterwards in glob-sorted
order, so **output ordering and the counters are unchanged**. Before enabling
this I audited the corpus for shared state: every duplicated fixed `/tmp` path
lives inside a single file, no test binds a fixed port, and the shell suites use
`mktemp` in 50 places.

`kaappi test` gets the same treatment via `-j`/`--jobs`, which
`docs/dev/test-runner.md` already anticipated as the #1509 stretch goal.

### Measurements

4 cores, each run clean with nothing else on the CPU. All three produce
**identical results — 2019 pass, 0 fail**:

| Config | Time | vs before |
|---|---|---|
| `run-all.sh` before | 1070s (17.8 min) | — |
| `run-all.sh`, `KAAPPI_TEST_JOBS=1` (poll fix only) | 425s (7.1 min) | **2.5x** |
| `run-all.sh`, default (4 jobs) | **279s (4.7 min)** | **3.8x** |
| `kaappi test tests/scheme/srfi` | 18.5s → **4.6s** | **4.0x** |

### One real bug fixed along the way

`kaappi test` set the worker's `KAAPPI_TEST_EMIT` path via `setenv` on the
**parent** before each fork — a single mutable global shared by every in-flight
worker. Two concurrent spawns would have sent both children to the same emit
path and lost a result. It now travels in the child's own `envp`. Windows stays
at one job, where that env is still inherited from the parent
(`CreateProcessW` here builds no custom block); lifting that means threading an
environment block through `platform.winSpawnCaptureMerged`.

### Deliberately not done

**Shell suites stay sequential.** They are now the *larger* half of the
remaining wall time (~230s of the 279s). Several call `ensure_runtime_lib`,
which runs `zig build lib` into the shared `zig-out/lib`, so concurrent scripts
would race over one output archive. Parallelising them needs a build-once guard
in `tests/scheme/shell-common.sh` first, which affects every CI leg including
the BSD and Windows ones. Documented in `docs/dev/test-runner.md` as the next
opportunity.

Separately, profiling turned up that the bytecode cache is inert for any program
that imports — filed as #1888, not addressed here.

### Testing

`tests/scheme/test-runner/jobs.sh` pins the parity down by diffing **whole
transcripts** between `--jobs 1` and `--jobs 4`, not just totals. It includes a
deliberately slow-first fixture (`a.scm` is slowest and sorts first) so a
reporter that emitted in completion order would fail even though its counts
stayed correct, plus a vacuity guard so the diff cannot pass by comparing two
empty transcripts.

## Checklist

- [x] `zig build test` passes (1425 tests)
- [x] Scheme test suites pass (`bash tests/scheme/run-all.sh`) — 624 pass,
      0 fail, plus 1395 R7RS
- [x] New tests added for new behavior (`tests/scheme/test-runner/jobs.sh`)
- [x] Documentation updated — `docs/dev/test-runner.md`, `tests/scheme/CLAUDE.md`,
      root `CLAUDE.md`; `markdownlint-cli2` reports 0 issues across 82 files

> [!NOTE]
> Measured on Linux x86_64, 4 cores, ReleaseSafe, Zig 0.16.0. Speedup scales
> with core count, so CI runners with more cores should do better on the `.scm`
> half. macOS's bash 3.2 has no `wait -n`, so it uses a batch-draining fallback
> rather than a sliding pool — correct, marginally less efficient at batch
> boundaries.

https://claude.ai/code/session_01Qk6ip9ctt9dJANetqdwDMm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- **`kaappi test -j` / `--jobs <n>`** runs test files concurrently, defaulting
+  to one job per CPU (#1887). Every file was already an isolated worker
+  process, so this is a scheduling change only: verdicts, per-file output and
+  its ordering, and the summary counts are identical at any job count — worker
+  threads claim files from a shared counter while the main thread reports the
+  completed prefix in file order. `--jobs 1` keeps the old strictly-sequential
+  behaviour. Windows runs one job regardless, because there the worker's emit
+  path still reaches the child through the inherited parent environment.
+  Measured on 4 cores, `kaappi test tests/scheme/srfi` went from 18.5s to 4.6s.
+
+### Fixed
+
+- **`kaappi test` could have lost a file's results once it ran more than one
+  worker at a time** (#1887). The path each worker writes its JSON to was
+  published by setting `KAAPPI_TEST_EMIT` on the *parent* before each fork — a
+  single mutable global shared by every in-flight worker, so two concurrent
+  spawns would have sent both children to the same path. It now travels in the
+  child's own `envp`. Not reachable before `--jobs` existed, since spawns were
+  serialised.
+
 ## [0.22.1] - 2026-07-31
 
 ### Added
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -53,7 +53,10 @@ build id, target triple, build mode, compiled-in subsystems (the KEP-0004
 portable SRFIs, and initial VM/GC limits — all derived, no hardcoded second
 list; see `docs/dev/features.md`;
 `kaappi test [paths...]` runs SRFI-64 suites (`--json`, `--seed <n>`,
-`--lib-path`) aggregating from the runner's own counters; `--changed`
+`--lib-path`) aggregating from the runner's own counters; `-j`/`--jobs <n>`
+runs files concurrently (default: one per CPU, Windows always 1) with verdicts
+and output order identical at any job count, since each file was already its own
+worker process; `--changed`
 /`--list-affected` (with `--since <rev>`) select only suites whose R7RS import
 closure changed, falling back to a loud full run when the graph can't be trusted
 — see `docs/dev/test-runner.md`. `kaappi ast|expand|ir <file>` are read-only
--- a/docs/dev/test-runner.md
+++ b/docs/dev/test-runner.md
@@ -31,9 +31,11 @@ kaappi test [paths...]
 - **`-j` / `--jobs <n>`.** Run up to `n` files concurrently (default: one per
   CPU; `--jobs 1` forces the old strictly-sequential behaviour). Because every
   file is already an isolated worker process, this changes scheduling only —
-  verdicts, per-file output and its **ordering**, and the summary are identical
-  at any job count, which `tests/scheme/test-runner/jobs.sh` pins down by
-  diffing whole transcripts. Windows always runs one job; see below.
+  verdicts, per-file output and its **ordering**, and the summary counts are
+  identical at any job count. Durations are the one thing that legitimately
+  moves, so `tests/scheme/test-runner/jobs.sh` normalises the `…ms` fields and
+  then requires the whole transcript to match byte for byte. Windows always runs
+  one job; see below.
 - **Exit status** is nonzero iff a test failed, unexpectedly passed (`xpass`),
   or a file errored.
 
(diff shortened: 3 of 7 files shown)


baijum:
Addressed all three review findings in 2965584.

**`KAAPPI_TEST_JOBS` validation** — this one was a real hole, not a nitpick. I
reproduced it before fixing:

```console
$ bash -c 'running=0; JOBS="abc"; if [[ $running -ge $JOBS ]]; then echo "would wait (JOBS parsed as 0)"; fi'
would wait (JOBS parsed as 0)
```

`[[ ... -ge ... ]]` evaluates arithmetically, so a non-numeric value is silently
`0`. `KAAPPI_TEST_JOBS=abc` therefore did not fail — it quietly serialised the
run with a spurious `wait` per file and still reported success, which is the
worst shape for this bug: a typo in a CI env var costs a 4x-slower suite that
looks perfectly healthy. Bad values now exit 2 naming the offending value:

```console
$ KAAPPI_TEST_JOBS=abc bash tests/scheme/run-all.sh
run-all.sh: KAAPPI_TEST_JOBS must be a positive integer (got 'abc')
$ KAAPPI_TEST_JOBS='' bash tests/scheme/run-all.sh
Running .scm suites with 4 parallel job(s) (override: KAAPPI_TEST_JOBS).
```

Empty deliberately still means "auto" (`${VAR:-default}` treats unset and empty
alike). I also rejected `0` inside `detect_jobs`, which would otherwise have
passed a detected `0` straight through.

**`resolveJobs` unit test** — added, covering zero and one file, a request above
the file count, explicit `--jobs 1`, and auto. The Windows and
`single_threaded` expectations are gated on the same `comptime` conditions the
function uses, as suggested.

**Doc wording** — fair catch. The line claimed output is identical at any job
count without noting durations obviously are not. It now states that the test
normalises the `…ms` fields and then requires a byte-for-byte transcript match,
which is what `jobs.sh` actually does.

Verified after the change: `zig build test` passes (1426 tests now),
`tests/scheme/test-runner/jobs.sh` passes, `zig fmt --check src/` clean,
`markdownlint-cli2` reports 0 issues.

---

baijum:
Flagging that the 🟢s on the latest benchmark comment are runner noise, not this PR — please don't credit it with a 20% `call/cc` improvement.

This PR changes no interpreter code. Its only `src/` file is
`src/test_runner.zig`, the `kaappi test` orchestrator, which does not execute
during a benchmark run at all; everything else is docs, `tests/`, and
`run-all.sh`. And the two benchmarked commits differ by exactly one markdown
file:

```console
$ git diff --stat 2965584 a71f460
 CHANGELOG.md | 22 ++++++++++++++++++++++
 1 file changed, 22 insertions(+)
```

Yet the verdict moved from "all ⚪, 0.99–1.04x" to "5 🟢, `call_cc` 0.80x".

The clearest tell is the **base** column, which is re-measured each run against
the same commit `261fde5`:

| Benchmark | base @ 2965584 run | base @ a71f460 run | base-vs-base |
|---|--:|--:|--:|
| `call_cc` | 1.612578 | 0.994531 | **1.62x** |
| `fib` | 4.408113 | 3.360616 | 1.31x |
| `tak` | 2.967651 | 2.383651 | 1.25x |
| `vector` | 0.311692 | 0.252577 | 1.23x |
| `tailcall` | 1.228607 | 1.027044 | 1.20x |

Identical commit, nothing to explain a difference — so ~1.2–1.6x is this
harness's run-to-run spread on a shared runner. That is wider than the 20%
threshold the workflow uses to flag 🔴, which means the same variance could
manufacture a false regression on someone else's PR as easily as it
manufactured this false improvement. Worth knowing before anyone treats a
single ⚪/🟢/🔴 here as signal; `benchmarks/gate/` and the Kalibera-Jones
protocol in `docs/dev/kep-0002-phase7-envelope-benchmarks.md` exist precisely
because one paired run cannot resolve differences this small.

No action needed on this PR. Happy to file the harness-variance observation as
its own issue if it's useful.

---

baijum:
Filed the harness-variance observation as #1906, with the base-vs-base evidence
and the detail that `min`/`max`/`iterations` are already measured and serialised
into the benchmark JSON before being discarded at the comparison step. Nothing
further needed on this PR.

---

Issue #1911, opened 2026-07-31
Two divergent applyExactness implementations: the reader un-rounds an f64, string->number rebuilds from digits


baijum:
**Category:** Numeric tower · **Severity:** systemic · **Found by:** systematic audit v2, Phase 1A (#1890)

Umbrella for a family of reader-exactness defects that share one structural cause. Individually each looks like a small bug; together they have a single fix.

## The cause

`applyExactness` exists **twice**, with structurally different strategies:

| | `reader_tokens.zig:393` | `primitives_numeric.zig:979` |
|---|---|---|
| Used by | the reader (`#e`/`#i` literals) | `string->number`, `exact`, `inexact` |
| Strategy | parse to **f64**, then try to **un-round** it with a tolerance-based continued fraction | rebuild exactly from the decimal digits (mantissa × 10^scale via bignum) |

That is why the reader loses at **both ends** of the range while `string->number`
does not — it is not a missing edge case, it is a weaker algorithm.

R7RS §6.2.7 (p.40) makes the divergence a spec violation on its own terms:
*"The rules used by a particular implementation for `string->number` must also
be applied to `read` … in order to maintain consistency."*

## Instances

| issue | symptom | end of the range |
|---|---|---|
| #1907 | **panic, exit 134** on `#e` at exactly 2^63 — also kills `check`/`fmt`/`ast` | overflow |
| #1908 | `#i` on a radix bignum reinterprets digits as decimal | overflow |
| #1909 | `#e` below 1e-15 collapses to exact `0`, sign included | underflow |
| #1910 | no Complex arm in either implementation (reopens #751) | type gap |
| #1891 | `#e1e19` inexact, `#e1e400` ⟹ `+inf.0` | overflow |

Plus one parity gap with no separate issue: `#e+inf.0` / `#e-inf.0` / `#e+nan.0`
read as inexact inf/NaN, while `string->number` returns `#f`. **Closed #419
settled that `#f` is the intended answer**, so the reader is the wrong side.
Control: `#i+inf.0` agrees.

## Why past fixes kept missing

Each landed in one copy only:

| issue | fixed in | still live in |
|---|---|---|
| #79 | reader_tokens | guard added but **off by one** → #1907 |
| #604 | primitives_numeric | — |
| #419 | primitives_numeric | **reader_tokens** → the inf/nan parity gap above |
| #751 | primitives_numeric | fix landed but its target has no Complex arm → #1910 |
| #1891 | — | **reader_tokens** |

[...]

Issue #1910, opened 2026-07-31
applyExactness has no complex arm: #i1+2i is a no-op, #e1e19+1i writes an unreadable 0/0 (reopens #751's symptom)


baijum:
**Category:** Numeric tower · **Severity:** wrong-result · **Found by:** systematic audit v2, Phase 1A (#1890)

Neither `applyExactness` implementation has a `Complex` arm, in either direction. Three symptoms, one cause.

## 1. `#i` on a complex literal does nothing

```scheme
(list (exact? #i1+2i) (exact? #i1))
⟹ (#t #f)
```

`#i1+2i` stays **exact**. Control: `(exact? (inexact 1+2i))` ⟹ `#f`, so the procedure is right and only the prefix path is wrong. `reader_tokens.zig:443-444` — the inexact arm is `.flonum, .complex => tok`, an explicit pass-through.

## 2. `#e` on a complex past i64 produces an unreadable `0/0`

`#e1e19+1i` writes as `0/0+1i`, which reads back as `+nan.0+1i`; bare `0/0` is a read error. This was **the only round-trip failure in a 27-case `write`→`read` sweep**. Control: `#e1e18+1i` round-trips fine. `reader_tokens.zig:437-439` — the `.complex` arm sets both exactness flags with no range check at all.

## 3. #751 is closed but its own repro line still fails

```scheme
(exact? (string->number "#e1+2i"))   ⟹ #f
```

That is the reproduction from #751 ("string->number ignores #e/#i exactness prefix for complex numbers"), still failing.

**I initially misread this as "never fixed"; the truth is more useful.** #751's fix *did* land — all four `allocComplex` sites were wired through `applyExactness`. But `applyExactness` itself has no Complex arm: `primitives_numeric.zig:1000` and `:1008` both fall through to a bare `return val`. The plumbing was connected to a function that does nothing for this type.

**Recommend reopening #751** rather than treating this as new — the fix is incomplete at its destination, and reopening keeps that history attached.

## Suggested fix

Give `applyExactness` a real `Complex` arm in both implementations: recurse into
the real and imaginary parts and rebuild. That closes all three symptoms and
makes #751's own test meaningful.

**Verified:** fresh ReleaseSafe build at `12588782`, isolated `KAAPPI_HOME`; all three symptoms and both controls reproduced twice; both `applyExactness` bodies read directly.

Issue #1909, opened 2026-07-31
#e on a decimal below 1e-15 collapses to exact 0 — the entire value, including its sign, is lost


baijum:
**Category:** Reader · **Severity:** wrong-result · **Found by:** systematic audit v2, Phase 1A (#1890)

`#e` on a small decimal silently yields exact `0`. Not a rounding artefact — the whole value goes, sign included.

## Reproduction

```scheme
(list #e1e-16 #e1e-15 #e1e-14)
⟹ (0 1/999999999999999 1/100000000000000)
```

- `#e1e-16`, `#e1e-17`, `#e1e-19`, `#e2e-16`, `#e1.5e-18`, `#e1e-30`, `#e1e-400` → all **`0`**
- `#e-1e-20` → **`0`**, so the sign is lost too
- At the cliff itself, `#e1e-15` → `1/999999999999999` — an off-by-one denominator, not `1/1000000000000000`

## Discriminating controls

| control | result | tells us |
|---|---|---|
| `#e1e-14` | `1/100000000000000` — correct | cliff is at 1e-15, not "small decimals" generally |
| `#e0.1` | `1/10` | ordinary decimals fine |
| `#e3.14159` | correct | multi-digit fine |
| `(exact 1e-19)` | full exact value | the **procedure** is correct; only the reader is not |
| `string->number "#e1e-19"` | correct | reader-only, again |

## Mechanism

The continued-fraction rationalisation at `reader_tokens.zig:428/:430`
terminates on **absolute** tolerances (`< 1e-15`). Below that magnitude the very
first iteration already satisfies the tolerance at `0/1`, so it returns zero.
The cliff is precisely 1e-15, matching the constant.

This is the opposite end of the number line from #1891 and a completely
different mechanism — nothing overflows here. `(exact ...)` is unaffected
because `primitives_numeric.zig`'s own `applyExactness` rebuilds the value from
the decimal digits (mantissa × 10^scale via bignum) rather than un-rounding an
f64.

## Suggested fix

The reader should not be un-rounding an f64 at all. Reuse the digit-exact
strategy `primitives_numeric.zig:979` already implements — see the umbrella
issue on the two divergent `applyExactness` implementations.

**Verified:** fresh ReleaseSafe build at `12588782`, isolated `KAAPPI_HOME`, reproduced twice with all five controls.

Issue #1908, opened 2026-07-31
#i on a radix-prefixed bignum reinterprets the digits as decimal — silent wrong answer


baijum:
**Category:** Reader · **Severity:** wrong-result · **Found by:** systematic audit v2, Phase 1A (#1890)

`#i` applied to a radix-prefixed integer **beyond the i64 range** feeds the digit string to `parseFloat` without its radix, so the digits are reinterpreted as decimal. The result is a plausible-looking number that is wrong by orders of magnitude.

## Reproduction

| input | correct | actual |
|---|---|---|
| `#i#x1000000000000000000` | `4.7223664828696452e21` | **`1000000000000000000.0`** |
| `#i#b<65 ones>` | `3.6893488147419103e19` | **`1.1111111111111111e64`** |
| `#i#o100000000000000000000000` | `5.902958103587057e20` | **`1.0e23`** |
| `#i#xFFFFFFFFFFFFFFFFFF` | `4.722366482869645e21` | **spurious read error** |

Verbatim, showing the reader disagreeing with `string->number` on the same text:

```scheme
(list '#i#x1000000000000000000
      (string->number "#i#x1000000000000000000"))
⟹ (1000000000000000000.0 4.722366482869645e+21)
```

R7RS §6.2.7 (p.40) requires exactly this not to happen: *"The rules used by a
particular implementation for `string->number` must also be applied to `read`
… in order to maintain consistency."*

## Discriminating controls

| control | result | tells us |
|---|---|---|
| same magnitudes **without** `#i` | correct exact bignums | the bignum reader is fine |
| `#i#xFF`, `#i#b1010` (below i64) | correct | only the >i64 path is affected |
| `#i` on a **decimal** bignum | correct | radix is the discriminator, not size |
| `string->number` on all four | correct | the defect is reader-only |

## Mechanism

`reader_tokens.zig:447` — the `.bignum_str` arm calls `parseFloat(bs.str)` and
ignores `bs.radix`. The **adjacent** `.big_rational` arm (`:449-450`) *does*
check radix, and the comment at `:448` asserts the very invariant the line above
it fails to enforce.

## Note for #1891

#1891's stated root cause names this line (`:447-452`, the `parseFloat`
fallback). That is this bug, not #1891's — `#e1e19` dies at the `:400-402`
range guard in the `.flonum` arm. Fixing `:447` alone will not close #1891, and
fixing #1891 alone will not close this.

[...]

PR #1917, opened 2026-07-31
Phase 1A: reader exactness audit — 128 assertions, 5 new bugs including a process-abort panic

baijum:
Phase 1A of the v2 audit campaign (#1890). Adds `tests/scheme/compliance/reader-exactness-gaps.scm`.

## Result

**128 assertions pass, 0 fail.** 39 disabled behind `;; FAIL: #NNNN` markers, each sitting beside an **enabled discriminating control** so the disabled half re-enables with a one-line uncomment. Rest of `compliance/` still green (953 passed, 39 files).

## What was tested

R7RS §6.2.3–6.2.7 and §7.1.1 gave three oracles, most importantly §6.2.7 p.40 — *"The rules used by a particular implementation for `string->number` must also be applied to `read` … in order to maintain consistency."* That makes reader↔`string->number` parity a **spec requirement**, not a style preference, and it is where most of the findings live.

Matrix: `{#e,#i,none} × {#b,#o,#d,#x,none} × {fixnum, i64 boundary, beyond-i64, bignum, rational, decimal, exponent, tiny/huge exponent, inf/nan, complex} × {read, string->number}` — ~300 cells, plus a 27-case `write`→`read` round-trip sweep and a bisection around 2^63.

## New issues filed

| issue | severity | finding |
|---|---|---|
| #1907 | **crash** | `#e` at exactly 2^63 **panics, exit 134** — and aborts `check`, `fmt`, `ast` |
| #1908 | wrong-result | `#i` on a radix bignum reinterprets digits as decimal |
| #1909 | wrong-result | `#e` below 1e-15 collapses to exact `0`, sign included |
| #1910 | wrong-result | `applyExactness` has no Complex arm (reopens #751's symptom) |
| #1911 | systemic | umbrella: **two divergent `applyExactness` implementations** |

#1911 is the one that matters most. `applyExactness` exists twice: the reader parses to f64 and tries to **un-round** it with a tolerance-based continued fraction; `string->number` rebuilds exactly from the decimal digits. That is why the reader loses at *both* ends of the range and `string->number` does not — and why past fixes (#79, #419, #604, #751) each landed in one copy only.

## Two corrections to existing issues

- **#1891's stated root cause is wrong.** It names `reader_tokens.zig:447-452`, which is #1908's line. `#e1e19` actually dies at the `:400-402` range guard. Fixing `:447` alone will not close it. Commented on the issue.
- **#751 is closed but its own repro line still fails** — `(exact? (string->number "#e1+2i"))` ⟹ `#f`. Its fix *did* land (all four `allocComplex` sites wired through `applyExactness`), but that function has no Complex arm, so the plumbing connects to a no-op. Recommend reopening; detail in #1910.

## Deliberately not filed

Seven assertions marked `;; FAIL: #1892` cover a gap **adjacent to** that issue but not inside it: any prefix disables the trailing-delimiter check on the **decimal** path (`#e1.5abc` ⟹ `3/2`, and `'(#e1.5abc)` reads as two datums). #1892 is scoped to *radix* prefixes and `readIntegerWithRadix`; these need no radix prefix and go through `readNumber`. A fix confined to the radix path leaves them live — noted on #1892 for unit 1B.

## Confirmed correct

All eight `#e`/`#i` × radix orderings; `#x1e5` ⟹ 485 (hex digit, not exponent); SRFI 270 hex floats in both parsers; `#e`/`#i` on within-range rationals in every radix; `#i` on decimal bignums; `#e` on decimals ≥ 1e-14; round-trip on 26 of 27 cells. Pinned as enabled assertions so they are not re-audited.

## Note

Crash cases are written as `(read (open-input-string …))`, never source literals, so re-enabling one cannot abort the suite at read time.

### DIFF
--- a/tests/scheme/compliance/reader-exactness-gaps.scm
+++ b/tests/scheme/compliance/reader-exactness-gaps.scm
@@ -152,13 +152,13 @@
 (test-assert "string->number does not crash on the same text"
              (not (string->number "#e9223372036854775808.0")))
 
-;; FAIL: TBD (reader panics: #e<double == 2^63> overflows @intFromFloat)
+;; FAIL: #1907 (reader panics: #e<double == 2^63> overflows @intFromFloat)
 ;; (test-equal "#e2^63 as a decimal float" 9223372036854775808
 ;;             (rd "#e9223372036854775808.0"))
-;; FAIL: TBD (same panic -- lowest decimal that rounds up to 2^63)
+;; FAIL: #1907 (same panic -- lowest decimal that rounds up to 2^63)
 ;; (test-equal "#e lowest double rounding to 2^63" 9223372036854775808
 ;;             (rd "#e9223372036854775296.0"))
-;; FAIL: TBD (same panic via exponent notation)
+;; FAIL: #1907 (same panic via exponent notation)
 ;; (test-equal "#e2^63 via exponent" 9223372036854775808
 ;;             (rd "#e9.223372036854776e18"))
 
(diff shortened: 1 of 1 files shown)


baijum:
Both findings addressed in `6348a939`. One was right for a different reason than stated; the other was right outright and led to a new bug.

## 1. `(scheme inexact)` — fix correct, reasoning wrong

**Added.** `finite?`/`infinite?`/`nan?` are registered with `.libs = LS.initOne(.scheme_inexact)` (`primitives_numeric.zig:98-100`), so the import belongs there.

But the stated consequence — *"the file fails at expansion or evaluation and the whole suite reports zero results"* — did not happen. The file ran **128 assertions, 0 failed**, before the change:

```console
$ kaappi /tmp/f1.scm     # (import (scheme base) (scheme write)) only
(#t #t #t)               # (finite? 1.0) (infinite? +inf.0) (nan? +nan.0)
```

A top-level script reaches `vm.globals` directly whether or not the import is declared — the ambient-reachability behaviour documented in kaappi#1831/#1860. So the import is not load-bearing *here*, which is precisely the argument for declaring it: without it the file passes at top level while the identical code fails inside a library body. That reasoning is now a comment above the import so it does not get "tidied" away.

## 2. The control pinned non-conforming behaviour — correct, and it was hiding a bug

**Rewritten** to assert only the non-crash property, as suggested.

Investigating *why* `string->number` returned `#f` turned up something the control was masking. The **unprefixed** spelling fails too, and the reader gets it right:

| input | `string->number` | reader |
|---|---|---|
| `9223372036854775807.0` | correct | correct |
| **`9223372036854775808.0`** | **`#f`** | `9223372036854776000.0` |
| `-9223372036854775808.0` | correct | correct |
| `1e19`, `1e300` | correct | correct |

Same 2^63 boundary as #1907 and the same negative-side-is-fine asymmetry, but **no exactness prefix is involved**, so it is not an `applyExactness` defect. And the parity **inverts** — in #1908, #1909 and #1911 the reader is the wrong side; here it is the correct one.

Filed as **#1921**, with four enabled controls and one disabled assertion in the parity section.

On the second half of that comment — the cross-reference to section 8 — I dropped it rather than extending the table, since the rewritten control no longer asserts anything about `string->number`'s return value.

Issue #1922, opened 2026-07-31
Bytecode cache HIT loses the source line and snippet from runtime errors, breaking cache transparency


baijum:
**Category:** Bytecode cache · **Severity:** wrong-result (diagnostic) · **Found by:** systematic audit v2, Phase 4B (#1890)

`docs/dev/cache.md` states the cache is transparent — a HIT must behave exactly like a MISS. It does not, for a class of runtime errors: the source line and the code snippet disappear on the second run.

## Reproduction

```bash
printf '(display "start")\n(newline)\n(display (/ 1 0))\n' > dz.scm
kaappi dz.scm    # cold
kaappi dz.scm    # warm
```

```text
cold (MISS): dz.scm:3: error[KP3004]: division by zero
                 (display (/ 1 0))
warm (HIT):  dz.scm: error[KP3004]: division by zero
```

stdout and exit code agree; only the diagnostic degrades.

## Discriminating controls

| control | result | tells us |
|---|---|---|
| `(vector-ref (vector 1 2) 9)`, cached | `dz.scm:1:22:` **identical** cold and warm | the cache does **not** lose all debug info |
| `kaappi cache clear`, then rerun | full location restored | the cache is the responsible tier |
| `--no-ir-opt` (skips the cache both ways) | full location | same |

## Affected vs unaffected

Errors raised **without** a line-table entry covering the failing instruction lose their location: `(/ 1 0)`, `(error ...)`, `(raise ...)`.

Errors located via the line table are unaffected: `car '()`, `vector-ref`/`string-ref` out of range, undefined variable, non-procedure application, `assert`, wrong arity.

## Mechanism (inferred from source)

`toplevel_driver.vmErrorLocation` passes a hardcoded `fallback_line` of `0` on
the cache-HIT path (`src/main.zig:732`), where the fresh-compile path passes the
top-level datum's line (`src/main.zig:815`). The fallback only matters when
`vm.last_error_line == 0` — i.e. exactly the no-line-table-entry cases above,
which is why the split is invisible for everything else.

## A trap for whoever reproduces this

**Your repro must not `import` anything.** Programs that import are never cached
(`docs/dev/cache.md`; see also #1888), so cold and warm are the identical
configuration and the bug cannot appear. I hit this myself on the first attempt
and wrongly concluded it did not reproduce. Confirm with `--timings=text`, which
prints `cache: MISS (wrote …)` / `cache: HIT (…)`.

## Test

[...]

Issue #1516, opened 2026-07-13
Cache transparency: kaappi cache status|clear + build id in .sbc cache keys


baijum:
## Context

The `.sbc` bytecode cache is keyed on source hash + version *string*. During development the version string doesn't change between rebuilds, so a freshly rebuilt kaappi silently executes bytecode compiled by the **previous** binary. This has repeatedly manufactured phantom bugs and masked real fixes; the standing workaround — "delete the cache before testing compiler changes" — is tribal knowledge (currently documented as a footgun in project notes). `--no-ir-opt` already bypasses the cache entirely, which is itself a hint the key is too coarse.

For an agent, an invisible cache is a *correctness* hazard, not a convenience.

## What

1. **Honest keys**: include a build id (git commit hash + dirty flag embedded at build time; fallback to build timestamp) in the cache key, so a recompiled binary can never serve bytecode produced by a different compiler.
2. **`kaappi cache status`**: location, entry count, total size, per-entry source path + producing build id.
3. **`kaappi cache clear`**: the one supported way to wipe it — users shouldn't need to know the path.
4. Every run reports HIT/MISS under `--timings` (tracked in the timings issue).
5. **Docs**: one page on cache semantics — what the key contains, what invalidates, how to bypass.

## Acceptance criteria

- [ ] Dev rebuild (same version string, different commit/dirty state) → cache MISS, verified by an automated test
- [ ] `cache status` / `cache clear` subcommands
- [ ] Cache semantics documentation page
- [ ] CLAUDE.md / project-notes footgun entries updated to "fixed by build-id keys"

Part of #1503.

PR #1923, opened 2026-07-31
Phase 4B: differential harness for the opt-off and cache tiers — 557 files, one real divergence

baijum:
Phase 4B of the v2 audit campaign (#1890). Adds `tests/scheme/differential/run-differential.sh` plus 7 probe files, wired into `run-all.sh` as a shell suite.

## What it checks

Per file: cold `kaappi F` (oracle), warm `kaappi F` (tier d), `kaappi --no-ir-opt F` (tier b). Requires **byte-identical stdout, identical exit code, identical normalised stderr**. stdout and exit code are never normalised; only stderr is (absolute paths, `line:col`), and a raw-but-not-normalised stderr difference is reported as a non-failing `NOTE` so line-number drift stays visible.

Env: `KAAPPI_DIFF_FULL=1`, `KAAPPI_DIFF_TIMEOUT`, `KAAPPI_DIFF_VERBOSE`, `KAAPPI_DIFF_KEEP`.

## Results

| corpus | files | tier b | tier d | elapsed |
|---|---:|---|---|---:|
| default (probes+smoke+compliance+audit) | 330 | 0 differ | 0 differ | 106s |
| `KAAPPI_DIFF_FULL=1` (+continuations/hygiene/srfi) | 557 | 0 differ | 0 differ | 210s |

Default is the wired-in setting: full costs +90s and buys **zero** additional tier coverage (0 extra IR diffs, +7 cached files).

## The real finding — #1922

On a cache **HIT**, a runtime error loses its source line and snippet:

```
cold (MISS): dz.scm:3: error[KP3004]: division by zero
                 (display (/ 1 0))
warm (HIT):  dz.scm: error[KP3004]: division by zero
```

`fallback_line` is a hardcoded `0` on the HIT path (`main.zig:732`) where the fresh-compile path passes the datum's line (`main.zig:815`). It bites only when no line-table entry covers the failing instruction — `/ 1 0`, `(error ...)`, `(raise ...)`. Contradicts `docs/dev/cache.md`'s transparency contract (#1516).

Suppressed via `KNOWN_DIFFS` (the shell analogue of `;; FAIL: #1234`), which **reports STALE** when a listed entry stops diverging, so it cannot rot silently.

## The tier-(b) green is weak, and now quantified

Across all 550 pre-existing corpus files, `kaappi ir` and `ir --no-opt` are byte-identical on **547**. Only three files in the whole tree make the optimiser do anything — all three purpose-built optimiser regression tests. Across `compliance/`, `audit/`, `continuations/`, `hygiene/`, `srfi/` (288 files) the count is **zero**, because `define`/`lambda` bodies lower to opaque `passthrough` nodes.

The 7 added probes triple that (6 verified non-vacuous). The script's header says this prominently and the summary reports the census every run, so a green tier-b result is never mistaken for "the optimiser is correct".

## Two design choices beyond the brief

**The static skip list is empty.** Rather than guessing at nondeterminism, a mismatch must survive three controls: (1) *structural* — if the file wrote no `.sbc` entry, cold and warm are the identical configuration and a tier-d mismatch **cannot** be a cache effect; (2) two cold baselines must agree; (3) it must reproduce.

Control (1) earned its place immediately: `smoke/nested-wait-under-sleep-dirty-snapshot-1490.scm` reported a tier-d divergence that was a load-induced hang — the file writes zero cache entries across 5×3 runs, so there was nothing to diverge. Flagged as flaky under load, not a bug.

**`timed_out()` accepts 143**, not just 124/142 — the perl fallback exits SIGTERM, which the narrower check would have compared as a real exit status.

## Incidental finding, out of scope

`kaappi ir` cannot lower SRFI 17 generalized `set!`: `(set! (vector-ref v 0) 99)` emits `; lower error: InvalidSyntax` **inline in the dump** and exits 1 with nothing on stderr, while the same program runs correctly. Control: the same file without the generalized `set!` lowers cleanly. Hits `srfi123.scm` in the corpus. Not filed — belongs to a pipeline-dump unit.

### DIFF
--- a/tests/scheme/CLAUDE.md
+++ b/tests/scheme/CLAUDE.md
@@ -21,6 +21,7 @@
 | `fmt/` | `kaappi fmt` formatter | yes |
 | `cache/` | `.sbc` bytecode cache transparency | yes |
 | `timings/` | `--timings` stage reporting | yes |
+| `differential/` | Execution-tier differential harness (`--no-ir-opt`, cold-vs-warm cache) + its `probes/` | yes |
 | `coverage/` | Coverage gap-fillers (`zig build coverage-scheme`) | no |
 | `robustness/` | Stress tests | no (CI runs it separately) |
 | `sandbox/` | Sandbox isolation tests | no (CI runs it separately) |
--- a/tests/scheme/differential/probes/begin-simplify.scm
+++ b/tests/scheme/differential/probes/begin-simplify.scm
@@ -1,0 +1,35 @@
+;; Probe: begin simplification (ir.simplifyBegin) and the order of effects.
+;;
+;; Flattening `(begin A (begin B C))` into `(begin A B C)` is only sound if
+;; every effect keeps its position and the value of the whole form is still
+;; the LAST subform's value.  A single-subform `begin` collapsing to its
+;; subform, and a `begin` in value position, are the cases where an
+;; over-eager pass changes an answer rather than just a shape.
+
+;; --- effect order across nesting -----------------------------------------
+(begin (display "a") (begin (display "b") (display "c")) (display "d")) (newline)
+(begin (begin (begin (display "deep")))) (newline)
+(begin (display 1) (display 2) (display 3) (display 4) (display 5)) (newline)
+
+;; --- value of a `begin` is its last subform ------------------------------
+(display (begin 1 2 3)) (newline)
+(display (begin (display "[eff]") 42)) (newline)
+(display (begin 'only)) (newline)
+(write (begin "str")) (newline)
+
+;; --- a `begin` whose last subform is a `define`-free tail ----------------
+(display (begin (+ 1 1) (* 2 2) (- 9 3))) (newline)
+
+;; --- interleaving with `if`, so both passes see the same tree ------------
+(begin
+  (if #t (display "bi-then") (display "bi-else"))
+  (newline)
+  (begin
+    (if #f (display "bi2-then") (display "bi2-else"))
+    (newline)))
+
+;; --- a `begin` that raises partway: the prefix effects must already have
+;; happened, and the suffix must not.
+(display "pre") (newline)
+(begin (display "x") (newline) (car '()) (display "MUST-NOT-PRINT") (newline))
+(display "post") (newline)
--- a/tests/scheme/differential/probes/boolean-simplify.scm
+++ b/tests/scheme/differential/probes/boolean-simplify.scm
@@ -1,0 +1,40 @@
+;; Probe: boolean simplification (ir.simplifyBooleans).
+;;
+;; The pass rewrites `(if (not X) A B)` into `(if X B A)`.  Two things can go
+;; wrong and only show up against `--no-ir-opt`:
+;;   1. the arms get swapped the wrong way round;
+;;   2. `X` is evaluated a different number of times, or its side effects
+;;      move relative to the surrounding forms.
+;; The one-armed case `(if (not X) A)` is the sharp edge: rewriting it needs
+;; an alternate synthesised out of nothing.
+
+(define seen '())
+(define (note! tag v) (set! seen (cons tag seen)) v)
+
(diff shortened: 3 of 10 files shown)


baijum:
Rebased onto `main` after #1887 (parallel `run-all.sh`) and #1905, and re-verified on the rebased tree. Force-pushed `7ff0abe7`.

## The conflict was structural, not textual

- **`tests/scheme/CLAUDE.md`** — this branch added a `differential/` row where #1905 had added six others. Kept all seven.
- **`run-all.sh`** — merged textually, but #1887 rewrote how suites run, so I checked the registration still *fits* rather than assuming a clean textual merge meant a correct one. `run_shell_suite` survived the parallel rewrite unchanged, and `run_shell_suite "Differential tiers"` sits correctly among its siblings.

I rebuilt before re-verifying, since `main` also changed `src/test_runner.zig`.

## Re-verified numbers (these supersede the PR body)

| | before rebase | **after rebase** |
|---|---:|---:|
| corpus (default) | 330 | **331** (`main` added a file) |
| tier b / tier d | 0 differ | **0 differ** |
| standalone elapsed | 106s | **118s** |
| cache exercised | 40/330 | **40/331** |
| IR differs under `--no-ir-opt` | 9/330 | **9/331** |

## The timing question #1887 raised

The harness spawns ~1,000 `kaappi` processes, and `SHELL_TIMEOUT` is 300s — so a parallel `run-all.sh` could plausibly push it over and produce what looks like a flake rather than a budget problem. Measured rather than assumed:

```text
=== Differential tiers ===
  PASS  tests/scheme/differential/run-differential.sh

  Scheme files: 626 pass, 0 fail, 0 timeout, 0 skipped
  R7RS suite:   1395 pass, 0 fail
  Total:        2021 pass, 0 fail
```

Full `run-all.sh`: **6m35s total, 0 timeouts**, differential suite passing inside it.

**One honest limitation:** `run_shell_suite` swallows a passing suite's stdout, so I can confirm it finished under 300s but not the exact margin under contention. Standalone is 118s on a loaded laptop; CI hardware will differ. If it ever does get tight, `KAAPPI_DIFF_TIMEOUT` tunes the per-run bound and the default corpus is already the smaller of the two (`KAAPPI_DIFF_FULL=1` adds 227 files for zero extra optimiser coverage). Worth watching on the first few CI runs rather than pre-tuning blind.

baijum:
CI's Debug leg killed this at run-all.sh's 300s `SHELL_TIMEOUT`:

```text
TIMEOUT  tests/scheme/differential/run-differential.sh  (killed after 300s)
```

That is exactly the risk flagged when this was opened — *"CI hardware will differ… worth watching on the first few CI runs rather than pre-tuning blind."* It got tight on the one leg that was always going to be worst: Debug runs the whole pipeline unoptimised, ~500× slower on allocation-heavy work, against ~1000 interpreter invocations.

Fixed in `90add8e9` by **reducing the Debug corpus to `probes/`** rather than skipping the suite or raising the budget.

That is a principled reduction, and the harness's own census is the argument: **9 of 331 corpus files make the optimiser do anything, and 6 of those 9 are the probes.** So Debug keeps the files that discriminate and drops the ~322 whose comparison is vacuous by construction.

Measured against a real `-Doptimize=Debug` build, not assumed:

| | files | elapsed | optimiser-reaching |
|---|---:|---:|---|
| ReleaseSafe (unchanged) | 331 | 110s | 9 |
| **Debug (reduced)** | **7** | **15s** | **6 of 7** |

20× inside the budget, with essentially all the tier-(b) signal retained.

Two details worth noting:

- The build mode is read from **`kaappi features --json`**, not an env var, so it is correct whether the build came from CI, a local `-Doptimize=Debug`, or a stale `zig-out`.
- `KAAPPI_DIFF_FULL=1` still forces the whole corpus, so the Debug reduction is overridable when someone actually wants it.

The run prints why it reduced, so a future reader does not mistake a 7-file Debug run for the harness silently shrinking.

Issue #1959, opened 2026-07-31
The "port I/O abandoned" error names dynamic-wind as unparkable, contradicting README — it is bootstrapped Scheme and can park


baijum:
**Category:** Diagnostics / doc-truth · **Severity:** misleading user-visible text · **Found by:** systematic audit v2, Phase 5B (#1890)

`fiber.zig:1215` raises:

> `port I/O abandoned: fiber cannot suspend under re-entrant native frames (guard, dynamic-wind, callbacks) while an enclosing completed wait needs this thread`

**`dynamic-wind` does not belong in that list.** It is bootstrapped Scheme (`vm_bootstrap.zig:272`), runs in the bytecode dispatch loop, and a fiber *can* park inside it. Test-verified in this unit: `guard` drives, while `dynamic-wind` and a bare read **park**.

`README.md` § Fibers already says so, correctly:

> Callbacks driven by `map`, `for-each`, `vector-map`, `vector-for-each`, `string-map`, `string-for-each`, **`dynamic-wind`**, and `force` run in the bytecode dispatch loop, so a fiber can park inside them

So the README is right and the error message is the stale copy — the worse way round, since the error text is what a user sees while debugging exactly this situation. Someone hitting it will move blocking I/O out of a `dynamic-wind` for no reason, and leave it inside the `guard` that actually caused it.

## A second, quieter stale copy

`vm.zig:513` lists "map/for-each callbacks" among reentrant calls incrementing `native_reentry_depth`. Both are bootstrapped Scheme too, for the same reason. This one is a code comment rather than user-visible text.

## Suggested fix

Replace `dynamic-wind` in the error message with something accurate — `guard` and native higher-order drivers (SRFI-1 `fold`/`filter`/`find`, `sort`, `hash-table-walk`, `assoc`/`member` with a custom predicate, `string-index`, `eval`), which README already enumerates as the real native drivers. And correct the `vm.zig:513` comment.

`tests/…/tests_waitforfd.zig` (Phase 5B) pins the actual behaviour — `guard` drives, `map`/`dynamic-wind`/bare read park — so a corrected message can be checked against a test rather than against prose.

**Verified:** fresh ReleaseSafe build; the error text, `vm_bootstrap.zig`'s definition, and README's contradicting claim all read directly; the drive/park split asserted by committed tests.

Issue #1625, opened 2026-07-17
Parked guard-wrapped reader redispatched by a later fiber-join blocks in an unbounded reactor poll while a runnable sibling waits


baijum:
Found while hardening the close-port park/wake tests during the #1623 review round. **Independent of the new pipe backend**: it reproduces identically with an AF_UNIX socketpair (`makeBidiFdPair`) and with OS pipes, on macOS (kqueue), and involves only stage-1/KEP-0001 code paths.

## Repro (Zig test, hangs forever)

```zig
test "parked guard reader + later spawned closer + join" {
    var gc = memory.GC.init(std.testing.allocator);
    defer gc.deinit();
    var vm = try th.makeTestVM(&gc);
    defer vm.deinit();

    const pair = th.makeBidiFdPair(); // also hangs with th.makePipeFdPair()
    _ = try definePortGlobal(vm, "rp", pair[0], true, false);
    _ = try definePortGlobal(vm, "wp", pair[1], false, true);

    _ = try vm.eval("(import (kaappi fibers))");
    _ = try vm.eval(
        \\(define f (spawn (lambda ()
        \\  (guard (e (#t (list 'caught (error-object? e))))
        \\    (read-char rp)
        \\    'not-reached))))
    );
    // The load-bearing difference from the passing test: dispatch f so it
    // parks (.io_waiting, verified) BEFORE the closer exists...
    _ = try vm.eval("(fiber-join (spawn (lambda () 1)))");
    _ = try vm.eval("(define closer (spawn (lambda () (close-port rp))))");
    // ...then join. Hangs: closer never runs.
    _ = try vm.eval("(equal? (fiber-join f) '(caught #t))");
}
```

The existing suite's ordering — closer spawned before the join, both pending — passes: FIFO dispatch runs `f` first, it parks, and its in-place drive dispatches `closer`, whose `close-port` wakes it into the clean "port closed" raise.

## Observed state at the hang (`sample` on the test binary)

```
tests_port_io.test...
  vm.eval (fiber-join f)
    vm_calls.runWithScheduler → vm_dispatch.runUntil
      primitives_fiber.fiberJoinFn
        fiber.runSchedulerStep → vm_dispatch.runUntil     ← f dispatched, guard frames live
          primitives_control.callWithEscapeContinuation → vm.callHandler → vm_calls.callReentrant
            primitives_control.withExceptionHandlerFn → vm_calls.callThunk → vm_calls.callReentrant
[...]

PR #1960, opened 2026-07-31
Phase 5B: waitForFd park-vs-drive — 26 tests for a protocol that had zero direct coverage

baijum:
Phase 5B of the v2 audit campaign (#1890). Adds `src/tests_waitforfd.zig`.

**26 tests, all suite-safe, nothing probe-only.** `zig build test`: 1449 pass, 3 skip, 0 fail (exit 0). Full `run-all.sh` on the rebased tree: **632 pass, 0 fail; R7RS 1395/1395**. 33/33 across 6 runs, and 33/33 under `-Dgc-stress=true` with no behaviour change.

## No bugs found in `waitForFd` — and that is the result

`waitForFd`, `driving_waits` and `anyAncestorWaitResolved` had **zero direct references** in any test. The protocol turns out to be correct; what was missing was any way to notice if it stopped being.

## What is now pinned

**The park-vs-drive selector**, forced deliberately as a 2×2 with the two DRIVE cells as controls for the two conjuncts:

| `my_idx` | `dispatched_from_scheduler` | result |
|---|---|---|
| ≠0 | true | **park** — `error.Yielded`, `yield_retry`, `.io_waiting`, 1 reactor waiter |
| 0 / ≠0 | true / false | drive — normal return, `io_fd == null`, 0 waiters |

Plus: the park branch **ignores readiness** (parks identically on a readable and an empty fd, proving the branch is chosen by dispatch state rather than the fd), `.write` interest lands in `write_waiters`, and a failed `register` restores the prior status and leaves no waiter.

**The unwind asymmetry — its nine `false` rows had no coverage at all.** A comptime table over all ten production `Ctx` types now pins `IoWait` as the sole `unwind_on_resolved_ancestor = true`; an accidental change to `TargetWait`, `SleepWait`, `MutexWait`, `CondVarWait` or the five channel kinds would previously have been silent.

**The yield-retry contract holds at every parking path.** All fd parking funnels through `waitPortFd`'s three call sites; every read caller holding progress wraps in `propagateReadErr`, and the two bare `try`s are lead-byte reads with nothing consumed. The write side needs no stash — `portWriteBytes` drains before appending and progress lives in the durable `write_buf_start`. Five tests cover a park mid-item with an item-boundary control asserting `read_buf` stays empty.

## One correction to the unit's brief

I told this unit #1625 had **one** test; there are **three** (`tests_port_io.zig:172, 217, 248`). All three are the same shape and none touches `driving_waits`/`anyAncestorWaitResolved` or the asymmetry, so the gap stands — but the count was wrong.

## Filed: #1959

The user-visible `"port I/O abandoned"` error names **`dynamic-wind`** among frames a fiber cannot suspend under. It is bootstrapped Scheme (`vm_bootstrap.zig:272`) and *can* park — `README.md` § Fibers says so correctly, so the error text is the stale copy. That is the worse direction: someone debugging this will move I/O out of a `dynamic-wind` for nothing and leave it inside the `guard` that actually caused it. A second stale copy sits in a `vm.zig:513` comment.

## No Scheme-level tests, deliberately — and that is a finding

The park branch needs an fd that can `EAGAIN`. Core Scheme exposes no constructor for a pipe or socket; file ports never `EAGAIN`, and `create-fifo` needs two OS threads to rendezvous through a blocking `open(2)` — exactly the wall-clock construction that makes concurrency tests flaky. So this protocol is unreachable from Scheme except incidentally, which is why it had no coverage.

## One latent coupling, unreachable today

`anyAncestorWaitResolved`'s `w.fiber != me` filter would mask an outer drive if one fiber ever nested two drives. Every same-fiber re-entrant path (custom-port callbacks) is explicitly rejected first, so it is sound now — but a new same-fiber native driver would silently reopen #1625.

## Backend independence

Everything `waitForFd` itself does is backend-free; only `register→arm` and `parkOnReactor→wait` touch it. The comptime table and all `anyAncestorWaitResolved` tests are fully backend-free. **INFERRED and worth a `/do-linux-test`**: epoll's ONESHOT disarms the *whole* fd where kqueue's knotes are per-direction — the two-waiters-on-one-fd test is the one most likely to expose a difference. WASI's `arm` cannot fail at all, so the register-failure test is macOS/Linux/Windows only.

### DIFF
--- a/src/fiber.zig
+++ b/src/fiber.zig
@@ -1134,7 +1134,7 @@ pub fn wakeIoWaitersOnFd(sched: *FiberScheduler, fd: platform.fd_t) void {
 
 /// Wait until the current fiber's own fd readiness resolves: done as soon
 /// as something (reactor poll, close-port wake) flips it out of io_waiting.
-const IoWait = struct {
+pub const IoWait = struct {
     me: *Fiber,
     pub fn isDone(self: IoWait) bool {
         return self.me.status != .io_waiting;
--- a/src/primitives_fiber.zig
+++ b/src/primitives_fiber.zig
@@ -274,7 +274,7 @@ fn clearRvWaitDeadline(vm: *vm_mod.VM) void {
 /// a freed slot (for a rendezvous channel: unmatched receiver demand) or
 /// the channel becoming closed (channel-close! wakes both waiter roles via
 /// the same wakeChannelWaiters call).
-const ChannelSendWait = struct {
+pub const ChannelSendWait = struct {
     ch: *types.Channel,
     pub fn isDone(self: ChannelSendWait) bool {
         const cap = self.ch.capacity orelse return true;
@@ -414,14 +414,14 @@ fn channelSendLocal(ch: *types.Channel, ch_val: Value, payload: Value, deadline_
         "channel-send: deadlock — channel is full and no fibers can receive"));
 }
 
-const ChannelWait = struct {
+pub const ChannelWait = struct {
     ch: *types.Channel,
     pub fn isDone(self: ChannelWait) bool {
         return self.ch.head != types.NIL;
     }
 };
 
-const SharedChannelWait = struct {
+pub const SharedChannelWait = struct {
     me: *fiber_mod.Fiber,
     pub fn isDone(self: SharedChannelWait) bool {
         return self.me.status != .waiting;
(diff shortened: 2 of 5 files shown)


baijum:
FreeBSD failed on `tests/scheme/errors/error-format.sh` — *"runtime error has diagnostic — expected 'expected number' in output"*. Rebased onto current `main` (`cb2b61b6`) and re-running.

Worth recording what the failure was **not**, since the obvious readings are both wrong:

- **Not this PR's tests.** The unit suite passed on that FreeBSD run: `1457 passed; 3 skipped; 0 failed`. `error-format.sh` is a shell suite this PR does not touch — it adds one file, `src/tests_waitforfd.zig`.
- **Not the diagnostics change in #1958**, which was my first hypothesis and which I checked before acting on: that commit touches `primitives.zig`, `primitives_srfi160.zig`, `primitives_srfi181.zig` and one audit test, and contains **zero** occurrences of `expected number`.

The branch was four commits behind `main` (missing #1952, #1956, #1957, #1958). #1957 in particular changed how `run-all.sh` schedules its shell suites, so the old branch was running a different harness than the one `main` is green under.

`error-format.sh` passes locally on the rebased tree — **133 passed, 0 failed** — so if FreeBSD is still red after this, the cause is genuinely platform-specific rather than staleness, and I will chase it as such rather than rebasing again.

PR #1965, opened 2026-07-31
Stop the "port I/O abandoned" error naming dynamic-wind as unparkable

### LINKED ISSUE #1959: The "port I/O abandoned" error names dynamic-wind as unparkable, contradicting README — it is bootstrapped Scheme and can park

baijum:
Fixes #1959.

## The bug

`fiber.zig`'s `raiseIoWaitAbandoned` told users a fiber cannot suspend under
*"guard, dynamic-wind, callbacks"*. `dynamic-wind` is bootstrapped Scheme
(`vm_bootstrap.zig:272`) — its body runs in the bytecode dispatch loop and a
fiber parks inside it exactly like a bare read.

This is the worst way for a diagnostic to be wrong: the text is what a user
reads *while standing in this situation*, so it sends them to move blocking
I/O out of a `dynamic-wind` for no reason, leaving it inside the `guard` that
actually caused the drive.

## What changed

**The message** now names what genuinely opens a nested `runUntil`:

> port I/O abandoned: fiber cannot suspend under re-entrant native frames (guard, and native higher-order drivers such as SRFI-1 fold/filter/find, hash-table-walk, assoc/member with a custom predicate, string-index, eval) while an enclosing completed wait needs this thread

**Four code comments** carried the same stale claim, each naming `map`/`for-each`
callbacks as native frames — `vm.zig`'s `dispatched_from_scheduler` (the one the
issue flagged) and `in_custom_port_callback`, `vm_dispatch.zig`'s `runUntil`, and
`primitives_fiber.zig`'s `blockOrDeadlock`. Corrected alongside, since that is
how the error text acquired the claim in the first place.

One nuance kept rather than flattened: `dynamic-wind`'s before/after thunks **are**
native frames when a continuation transition re-enters them via `callThunk`
(`vm_continuations.zig:111/151/158`). That is the wind machinery, not the ordinary
`(dynamic-wind before thunk after)` call the message was read as indicting — so
`vm_dispatch.zig`'s `runUntil` doc, which was talking about exactly that path,
keeps its `dynamic-wind` reference, spelled out.

**One README correction beyond the issue's ask.** The issue holds up README §Fibers
as the correct copy, and it is — except it lists `sort` among the native drivers.
There is no native `sort`: it is portable Scheme in `lib/srfi/95.sld`, so it parks
too. Dropped, so README and the corrected message agree.

## Verification

New test `#1959: a reader inside dynamic-wind or map parks; only native frames
drive` in `tests_port_io.zig`, placed directly above the `#1625` test where `guard`
drives and unwinds, so the contrast is readable in one screen. Both readers are
dispatched *before any writer exists*, so a lucky write-first schedule cannot
satisfy it — a fiber that drove in place never reaches `.io_waiting`.

Mutation-checked: flipping the expectation to `.completed` fails with
`expected .completed, found .io_waiting`, confirming the test both runs and
discriminates.

- `zig build test` — 1457 pass, 3 skip
- `bash tests/scheme/run-all.sh` — 2029 pass, 0 fail
- `zig fmt --check`, `markdownlint-cli2` — clean

## Note for a follow-up (different repo)

`kaappi.github.io/docs/conformance.md:37` repeats the same two errors — it lists
`dynamic-wind` correctly as parkable but `sort` incorrectly as a native driver.
Not touched here since it is a separate repo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- **The "port I/O abandoned" error no longer blames `dynamic-wind`** (#1959).
+  It named "guard, dynamic-wind, callbacks" as the frames a fiber cannot
+  suspend under, but `dynamic-wind` is bootstrapped Scheme: its body runs in
+  the bytecode dispatch loop and a fiber parks inside it exactly like a bare
+  read. Anyone debugging this error was being sent to move blocking I/O out
+  of a `dynamic-wind` for no reason, leaving it inside the `guard` that
+  actually caused the drive. The message now names what genuinely opens a
+  nested native frame: `guard`, and the native higher-order drivers (SRFI-1
+  `fold`/`filter`/`find`, `hash-table-walk`, `assoc`/`member` with a custom
+  predicate, `string-index`, `eval`). README's matching list also dropped
+  `sort`, which is portable Scheme (SRFI 95) and parks too.
 - **Nested `guard` past 64 levels no longer returns a wrong answer** (#1886).
   The exception-handler and dynamic-wind stacks were fixed 64-entry arrays.
   Past that, `with-exception-handler` relabelled the overflow as
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,7 +57,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   nested native frame: `guard`, and the native higher-order drivers (SRFI-1
   `fold`/`filter`/`find`, `hash-table-walk`, `assoc`/`member` with a custom
   predicate, `string-index`, `eval`). README's matching list also dropped
-  `sort`, which is portable Scheme (SRFI 95) and parks too.
+  `sort`, which is portable Scheme (SRFI 95) and parks too — a fiber
+  blocking inside a `sort` comparator is now covered by the same test.
 - **Nested `guard` past 64 levels no longer returns a wrong answer** (#1886).
   The exception-handler and dynamic-wind stacks were fixed 64-entry arrays.
   Past that, `with-exception-handler` relabelled the overflow as
--- a/README.md
+++ b/README.md
@@ -420,8 +420,8 @@ Callbacks driven by `map`, `for-each`, `vector-map`, `vector-for-each`,
 bytecode dispatch loop, so a fiber can park inside them (e.g. block on an
 empty channel) and resume later. Other higher-order procedures are still
 native drivers — SRFI-1 (`fold`, `filter`, `find`, `any`, `every`, ...),
-`sort`, `hash-table-walk`/`hash-table-update!`, `assoc`/`member` with a
-custom predicate, `string-index`, `eval`, ... — and a fiber that blocks on
+`hash-table-walk`/`hash-table-update!`, `assoc`/`member` with a custom
+predicate, `string-index`, `eval`, ... — and a fiber that blocks on
 an empty channel inside one of those callbacks cannot be parked: the native
 call's state lives on the Zig stack and cannot be suspended. If other fibers
 are runnable the scheduler still makes progress, but if the blocked receive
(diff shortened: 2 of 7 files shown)


baijum:
Thanks — two of the three were real, and one of them found a copy of this PR's own defect that I'd missed. Addressed in 8f59011b.

**`src/vm_dispatch.zig` — `raiseDeadNativeReturn`'s "(map, for-each, sort, apply, ...)": valid, and a good catch.** This is a *fifth* instance of exactly what the PR is fixing, and my greps missed it because it spells the list differently from the `map/for-each callbacks` phrasing every other copy used. `returns_to_native` frames are pushed by `callReentrant` on behalf of `callWithArgs`/`callThunk`/`callHandler`, so `map`/`for-each`/`sort` cannot own one at all — and `sort` is precisely the name this PR added a test to prove parks. Rewritten to `apply, a native higher-order driver like SRFI-1 fold or hash-table-update!, eval, a wind thunk, ...`.

**`src/vm.zig` — ordinary unwinds also re-enter wind thunks: valid.** Confirmed two non-continuation call sites — a frame return leaving its own extent (`vm_dispatch.zig:808-816`) and an error unwinding `execute` (`vm_calls.zig:224-227`). My own note in `fiber.zig` already said "a continuation transition **or an unwind**", so the two comments disagreed; they now agree. I wrote my own wording rather than taking the committable suggestion, which appended the replacement sentence instead of replacing the original, leaving both.

**`README.md` — declining.** The guideline this cites scopes the docs repo to the *documentation site* (guide, procedures, cookbook, ecosystem); `README.md` is the core repo's own front page and has carried a §Fibers section since long before this PR. The diff here doesn't add end-user documentation — it deletes one incorrect word (`sort`) from a pre-existing paragraph, because `sort` is portable Scheme in `lib/srfi/95.sld` and parks. Moving the paragraph wholesale to the docs repo is a separate editorial decision, and leaving the wrong word in place meanwhile would keep README contradicting both the corrected error text and the new test.

The docs-site copy of the same list *was* wrong in the same way, and is fixed separately in kaappi/kaappi.github.io#27.

Unit suite green at 1519 pass / 3 skip; `zig fmt --check` clean. Comments only in this commit — no behaviour change.

Issue #1974, opened 2026-07-31
[SRFI-237] Four R6RS conformance defects: field-names returns a list, integer k is absolute, opaque unenforced, mutator accepts immutable fields


baijum:
**Category:** SRFI 237 / R6RS records · **Severity:** wrong-result · **Found by:** systematic audit v2, Phase 2.5 (#1890)

Four independent deviations from R6RS §6.3, each with a control showing the implementation already has what it needs.

## 1. `record-type-field-names` returns a list; R6RS requires a vector

> "Returns a **vector** of symbols naming the fields…"

```scheme
(vector? (record-type-field-names C))   ⟹ #f
(vector-length (record-type-field-names C))   ⟹ raises
```

**Control:** `make-record-type-descriptor` *requires* a vector for the same field list, so the library is asymmetric rather than uniformly list-based.

## 2. `record-accessor`/`record-mutator` integer `k` is absolute, not own-field-relative

> "Note that k cannot be used to specify a field of any type rtd extends."

With `P(a b)` ← `C(c d)` and instance `(1 2 3 4)`:

```scheme
((record-accessor C 0) i)   ⟹ 1     ; should be 3 (field c)
((record-accessor C 1) i)   ⟹ 2     ; should be 4 (field d)
((record-accessor C 'c) i)  ⟹ 3     ; by-name is correct
```

**Two controls.** The by-name path resolves per level correctly. And `record-field-mutable?` on the **same rtd with the same k** already uses own-field indexing — so absolute indexing is a defect here, not a house convention.

Secondary symptom: the integer path also skips rtd validation. `(record-accessor 5 0)` returns a procedure, while `(record-accessor 5 'x)` correctly raises.

## 3. The opaque flag is stored but never enforced

> "If passed an instance of the record type, `record?` returns `#f`… `record-rtd`… an exception is raised." … "The record type is also opaque if an opaque parent is supplied."

```scheme
(record? opaque-instance)        ⟹ #t     ; should be #f
(record-rtd opaque-instance)     ⟹ rtd    ; should raise
(record-type-opaque? child-of-opaque)  ⟹ #f   ; should inherit #t
```

**Controls:** the flag itself reads back correctly (`record-type-opaque?` on the parent ⟹ `#t`), a type declaring `opaque? #t` directly reports `#t`, and the parent link is intact and walkable. Everything needed to enforce it is present.

## 4. `record-mutator` does not reject an immutable field

[...]

Issue #1973, opened 2026-07-31
A record with 27 or more fields aborts the process on instantiation — u8 overflow in allocRecordInstance


baijum:
**Category:** GC / records · **Severity:** crash · **Found by:** systematic audit v2, Phase 2.5 (#1890)

Creating an *instance* of a record type with **≥27 fields** panics with an integer overflow. Not catchable by `guard`; exit 134.

**This is plain R7RS `define-record-type`.** No SRFI 237, no procedural layer, nothing exotic.

## Reproduction

```scheme
(import (scheme base) (scheme write))
(define-record-type big (mk f0 f1 ... f26) big? (f0 g0) ... (f26 g26))
(big? (mk 0 1 ... 26))
```

```text
thread 32696792 panic: integer overflow
src/gc_alloc.zig:473 in allocRecordInstance
exit 134
```

Wrapping the constructor in `guard` does not help — still exit 134.

## Bisected cliff

| fields | result |
|---:|---|
| 26 | `#t` |
| **27** | **panic** |
| 28 | panic |

## Mechanism

`gc_alloc.zig:473`:

```zig
self.finishAlloc(&ri.header, @sizeOf(RecordInstance) + record_type.num_fields * @sizeOf(Value));
```

`num_fields` is **`u8`** (`types_record.zig:11`), so `num_fields * @sizeOf(Value)` is evaluated in `u8` arithmetic. 40 + 8×27 = 256, which overflows.

## The dangerous part

**The record *type* is created without complaint.** Only the first instantiation dies. So a library can define a wide record type, load fine, pass its own type-level tests, and abort the first time anyone constructs one.

Reachable three ways, all verified: own fields, an inherited chain (20 + 7), and plain R7RS as above.

## Inferred, and worth checking before fixing

Under `ReleaseFast`, unsigned overflow is illegal behaviour rather than a trap — `finishAlloc` would receive a **wrapped** byte count and silently corrupt GC accounting instead of aborting. Not verified (the repro was run under ReleaseSafe); it changes the severity from "crash" to "silent heap corruption" if true, so it should be established before choosing a fix.

## Suggested fix

Widen the arithmetic at the call site (`@as(usize, record_type.num_fields) * @sizeOf(Value)`), and decide separately whether `num_fields: u8` is the right type at all — the 255-field limit it implies is enforced elsewhere with a proper `argError`, so the field *count* cap is deliberate, but the size *computation* should not inherit the narrow type.

[...]

Issue #1932, opened 2026-07-31
A record returned through thread-join! loses its type: prints correctly, but pt? is #f and every accessor raises


baijum:
**Category:** SRFI-18 threads / records · **Severity:** wrong-result · **Found by:** systematic audit v2, Phase 5C (#1890)

Returning a record from a worker thread is the plain, supported pattern — no shared globals, no mutation. It produces an instance of a **different type**.

## Reproduction

```scheme
(import (scheme base) (scheme write) (srfi 18))
(define-record-type <pt> (make-pt x y) pt? (x pt-x) (y pt-y))
(define t (make-thread (lambda () (make-pt 1 2))))
(thread-start! t)
(define r (thread-join! t))
(list r (pt? r) (pt-x r))
```

**3/3 runs**, ReleaseSafe and `-Dgc-stress=true` alike:

```text
value    #<<pt> 1 2>                                  <- prints correctly!
pt?      #f
accessor RAISED "type error in '%record-ref': expected <pt>, got #<record_instance>"
```

Control, same record built on the main thread: `pt?` ⟹ `#t`, `pt-x` ⟹ `1`.

## Why this is worse than it looks

The value **prints as a well-formed `<pt>`**. Nothing signals that it is a different type until a predicate quietly answers `#f` or an accessor raises somewhere downstream. A `cond` dispatching on `pt?` silently takes the wrong branch.

## Mechanism

`gc_deep_copy.zig:304-314` manufactures a **fresh `RecordType`** when copying a record instance that carries no SRFI 237 metadata. Record types are identity-based, so the copy is disjoint from the original by construction.

The same root cause shows from the other side: a **lexically captured** predicate answers `#f` for a genuine parent-heap instance, and a captured accessor raises.

## Discriminating control

A SRFI 237 `nongenerative` record type — one with a uid — **works**. The uid registry (`record_uid_registry`, shared via `VM.initForThread`) is the one identity-preserving path in `gc_deep_copy`. That is what pins the mechanism: it is not "records cannot cross threads", it is "records without a uid get a new type".

## Suggested direction

Give the plain path the same identity preservation the uid path already has — key on the parent's `RecordType` pointer during the copy, or synthesise a uid for every record type so the existing registry lookup applies. The machinery exists; only the no-metadata case bypasses it.

Distinct from #1924 (a child *mutating* a shared object). This one needs no sharing and no mutation at all — just `thread-join!`.

[...]

Issue #1915, opened 2026-07-31
%make-record never checks the field count: short calls pad with #<undefined>, long calls silently drop values


baijum:
**Category:** Records / internal primitives · **Severity:** wrong-result · **Found by:** systematic audit v2, Phase 2.1 (#1890)

`%make-record` does not compare the supplied field count against `rt.num_fields`.

```scheme
(%make-record (%make-record-type "T" 2) 1)        ⟹ #<T 1 #<undefined>>
(%make-record (%make-record-type "T" 2) 1 2 3 4)  ⟹ #<T 1 2>
```

Too few arguments pad with `types.UNDEFINED` (`gc_alloc.zig:454-460`); too many are dropped. Neither raises.

## Why the padding is the worse half

`#<undefined>` then escapes into ordinary user code as a **truthy**, printable, unreadable value:

```scheme
(if (%record-ref r 1 rt) 'true 'false)   ⟹ true
```

So an under-supplied constructor does not fail at construction, and does not fail at the first read either — it produces a value that passes a truth test and only misbehaves later, far from the cause.

`%make-record` is `INTERNAL`, but as established in #1912's unit, that means "exported by nothing", not "unreachable" — it is callable from a plain script with no import, and the portable record SRFIs build on it.

## Suggested fix

Compare `args.len - 1` against `rt.num_fields` and raise `argError` (KP3007) naming both counts. Add both directions to `tests/scheme/audit/internal-primitives-audit.scm`, where they currently sit disabled behind `;; FAIL: TBD` markers.

**Verified:** fresh ReleaseSafe build at `261fde5f`, isolated `KAAPPI_HOME`, reproduced twice.

PR #1975, opened 2026-07-31
Phase 2.5: SRFI 237 audit — 186 assertions, and a 27-field record aborts the process

### LINKED ISSUE #1199: [R7RS 5.5] record accessors/mutators do not check the record type — cross-type access silently reads/writes the wrong field

baijum:
Phase 2.5 of the v2 audit campaign (#1890). Adds `tests/scheme/audit/primitives_srfi237-audit.scm` for a file with **18 specs and no audit test**.

**186 enabled assertions, 26 disabled**, green on both runners. Full `run-all.sh` on the rebased tree: **636 pass, 0 fail; R7RS 1395/1395**.

## #1973 — critical, and not a SRFI 237 bug at all

A record with **≥27 fields aborts the process on first instantiation**. Exit 134, uncatchable by `guard`.

```scheme
(define-record-type big (mk f0 … f26) big? (f0 g0) … (f26 g26))
(mk 0 1 … 26)     ⟹ thread panic: integer overflow  (gc_alloc.zig:473)
```

**This is plain R7RS `define-record-type`.** `num_fields` is `u8`, so `40 + 8×27 = 256` overflows. Bisected: 26 fine, 27 panics. Reachable three ways — own fields, an inherited 20+7 chain, and plain R7RS.

The nastiest part: **the record *type* builds without complaint**, so a library can define a wide record, load cleanly, pass type-level tests, and abort the first time anyone constructs one.

Flagged as inferred-not-verified in the issue: under `ReleaseFast` unsigned overflow is illegal behaviour rather than a trap, so `finishAlloc` would receive a *wrapped* byte count — silent heap corruption instead of an abort. Worth establishing before choosing the fix.

## #1974 — four R6RS conformance defects, each with a control

| defect | control showing the machinery exists |
|---|---|
| `record-type-field-names` returns a list; R6RS says vector | `make-record-type-descriptor` *requires* a vector for the same list |
| integer `k` is absolute, not own-field-relative | **`record-field-mutable?` on the same rtd and same k already uses own-field indexing** |
| `opaque` stored but never enforced (`record?`, `record-rtd`, inheritance) | the flag reads back correctly and the parent link is walkable |
| `record-mutator` accepts an immutable field | `record-field-mutable?` correctly reports `#f` for it |

Plus three specified names missing and undocumented: `make-record-constructor-descriptor`, `record-constructor-descriptor?`, and the 7-argument `make-record-descriptor`.

**A trap worth knowing:** a test written from the R6RS spec fails on the *missing binding* rather than the defect under test. That confounded my first verification pass; it is noted in the issue.

## Confirmed correct — the larger half

**Protocol composition is exact: 16/16 masks** over a 4-level chain, with protocols running **once per `record-constructor` call, never per instance**. Field shadowing is fully correct — a child redeclaring an ancestor's name gets its own slot and mutability, the ancestor's stays readable through the parent rtd, and mutating one leaves the other untouched. Sealed enforcement is shared between both layers with identical wording, including a syntactically-sealed rtd rejected as a *procedural* parent. Nongenerative uid reuse is correct on all four R6RS axes, and the type *name* is correctly not among them.

Cross-type access control is right in all three directions — subtype→ancestor yes, reverse no, unrelated no — which closes #1199's class.

**The nongenerative cross-thread path is pinned as an enabled assertion**: a uid'd record survives `thread-join!` with `eqv?` rtd identity, a working predicate, and its uid preserved. #1932's generative failure is pinned as current behaviour, so its fix flips a test rather than going unnoticed.

**Zero panics across ~60 adversarial calls** apart from #1973, and `%record-split-args`' fixed 256-element stack buffer is correctly guarded (256 OK, 257 raises catchably, no overrun).

### DIFF
--- a/tests/scheme/audit/primitives_srfi237-audit.scm
+++ b/tests/scheme/audit/primitives_srfi237-audit.scm
@@ -1,0 +1,881 @@
+;;; Audit: src/primitives_srfi237.zig (SRFI 237 — R6RS Records, refined)
+;;; Systematic audit v2, Phase 2.5 (docs/audit-strategy.md, tracking #1890).
+;;;
+;;; Oracle: SRFI 237 defers to R6RS for every procedure ("Equivalent to the
+;;; procedure with the same name in R6RS"), so the authority for the
+;;; procedural and inspection layers is R6RS Library chapter 6 (Records).
+;;; Quoted R6RS sentences appear beside the assertions they justify.
+;;;
+;;; Covers all 18 specs in primitives_srfi237.zig, the R6RS `define-record-type`
+;;; desugarer that shares its enforcement helpers, and the portable layer in
+;;; lib/srfi/237.sld that is the only caller of the 13 `.srfi_237_primitives`
+;;; entries.
+;;;
+;;; Run: zig-out/bin/kaappi tests/scheme/audit/primitives_srfi237-audit.scm
+
+(import (scheme base) (scheme write) (scheme process-context)
+        (srfi 18) (srfi 64) (srfi 237))
+
+(test-begin "primitives_srfi237 audit")
+
+;; #t iff the thunk raises anything catchable.
+(define (raises? thunk)
+  (call-with-current-continuation
+    (lambda (k)
+      (with-exception-handler (lambda (c) (k #t))
+                              (lambda () (thunk) #f)))))
+
+;; The raised message, or 'no-raise.
+(define (message-of thunk)
+  (call-with-current-continuation
+    (lambda (k)
+      (with-exception-handler
+        (lambda (c) (k (if (error-object? c) (error-object-message c) 'non-error)))
+        (lambda () (thunk) 'no-raise)))))
+
+;; Naive substring search, so message assertions need no SRFI-13 import.
+(define (string-search haystack needle)
+  (let ((hn (string-length haystack)) (nn (string-length needle)))
+    (let loop ((i 0))
+      (cond ((> (+ i nn) hn) #f)
+            ((string=? (substring haystack i (+ i nn)) needle) #t)
+            (else (loop (+ i 1)))))))
+
+(define (fields-vector n)
+  (let loop ((i 0) (acc '()))
+    (if (= i n)
+        (list->vector (reverse acc))
+        (loop (+ i 1)
+              (cons (list 'immutable (string->symbol (string-append "f" (number->string i))))
+                    acc)))))
+
+(define (count-up n)
+  (let loop ((i 0) (acc '())) (if (= i n) (reverse acc) (loop (+ i 1) (cons i acc)))))
+
+(define (root-ctor rtd) (record-constructor (make-record-descriptor rtd #f #f)))
+
+
+;;; ===================================================================
+;;; D1 — the internal-primitive direct-call surface
+;;;
+;;; All 18 specs are registered in vm.globals, so every one is callable
+;;; from a plain script with no import at all. That makes each a reachable
+;;; surface that must reject bad input catchably rather than crash.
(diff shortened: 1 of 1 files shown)


baijum:
Reviewed this as a discovery-only audit test file (no source changes). Overall it's solid — the 26 disabled assertions I counted match the PR description exactly, each `FAIL:` marker points at a filed issue (#1973/#1974/#1914/#1915/#1932), and every disabled case has an enabled control demonstrating the related machinery does work, which is the right structure for a characterization suite like this.

Two minor nits, neither blocking:

1. **Flaky-by-construction assertion** (`tests/scheme/audit/primitives_srfi237-audit.scm:373-374`):
   ```scheme
   (test-assert "arity: too FEW constructor arguments leaks an uninitialized field (#1915)"
                (not (equal? 1 ((record-accessor B 0) (ctor 1)))))
   ```
   This pins "not equal to 1" as a proxy for "reads uninitialized memory." Since the value is genuinely undefined (whatever byte pattern happens to occupy that slot), this could in principle read back `1` by coincidence on a different GC/allocator and fail spuriously, or pass without actually detecting a regression if the leaked value just happens to differ from 1 for unrelated reasons. A stronger and less flaky check would assert a specific stable property (e.g. `record?`/type stays intact) rather than inequality with an arbitrary sentinel — though I recognize this is inherently testing UB-ish behavior, so there may not be a clean fix.

2. **Unused import**: `(scheme write)` at line 16 doesn't appear to be exercised anywhere in the file (no `write`/`display` calls) — looks like it can be dropped.

No issues found with the R6RS-quote-then-assert structure, the protocol-composition matrix math, the 255-field-cap vs. 27-field-instantiation-overflow distinction (correctly kept as separate sections despite both being field-count related), or the cross-thread/nongenerative coverage.

Issue #1978, opened 2026-07-31
[SRFI-170] errno is discarded at ~102 syscall sites; the spec's posix-error protocol is absent and range errors report as file errors


baijum:
**Category:** SRFI 170 · **Severity:** missing-feature + error taxonomy · **Found by:** systematic audit v2, Phase 2.12 (#1890)

## 1. The spec's error protocol is not implemented

SRFI 170 mandates `posix-error?`, `posix-error-name` (*"a symbol that is the name associated with errno"*) and `posix-error-message`. Kaappi exports none of the three, and discards errno at all ~102 syscall sites.

The consequence is visible from Scheme: ENOENT, ENOTDIR, EACCES, ELOOP and ENAMETOOLONG all collapse into **one generic string per procedure**. A program cannot distinguish "no such file" from "permission denied" — which is exactly the distinction this SRFI exists to expose.

## 2. Argument-range failures are reported as *file* errors

These never touch the filesystem, yet all satisfy `file-error?`:

```scheme
(set-file-mode p #o10000)   ⟹ file-error? #t
(set-umask! -1)             ⟹ file-error? #t
(nice 999999999999)         ⟹ file-error? #t
```

**Control:** a genuine type error correctly answers `file-error?` ⟹ `#f`. So the predicate does discriminate — these cases are simply classified wrongly.

Per dimension D2 of the audit strategy these are `argError` (KP3007): a value of acceptable type the procedure rejects anyway. Calling them file errors sends a handler looking at the filesystem for a fault that is in its own argument.

## Also worth recording

`file-info`, `user-info` and `group-info` cannot cross a thread boundary — `gc_deep_copy.zig:509-511` lists them as `UncopyableType` beside `.port`/`.continuation`/`.fiber`, and the message misdescribes them as *"uncopyable type (port, continuation, etc.)"*. All three are pure value records (scalars plus owned string bytes, like the copyable `SchemeString`).

**Control:** `posix-time`'s `Srfi18Time` round-trips fine, and `.directory_object` in the same list *is* correctly uncopyable — it holds a live `DIR*`.

Related: #1937 covers the uncopyable-tag list being bypassed by the globals path; this is the complementary problem of the list being *too broad*.

**Verified:** fresh ReleaseSafe build at `03bda165`; the `file-error?` classifications and their control reproduced; the missing exports checked against `lib/srfi/170.sld`.

baijum:
[...]

Issue #1977, opened 2026-07-31
[SRFI-170] nice, set-environment-variable! and four optional arguments accept wrongly-typed values silently


baijum:
**Category:** SRFI 170 filesystem · **Severity:** wrong-result (silent) · **Found by:** systematic audit v2, Phase 2.12 (#1890)

Three related defects where a bad argument is discarded rather than rejected. Each has a control showing the validation exists but is reached only on a different path.

## 1. `nice` renices the process on a wrongly-typed argument

```scheme
(nice "x")             ⟹ 1        ; and the level really moves
(nice (expt 2 100))    ⟹ 1
```

**Control:** `(nice 999999999999)` raises `"nice value out of range"`. So the range check exists; only the *type* test is missing — `if (args.len > 0 and types.isFixnum(args[0]))` silently treats a non-fixnum as "no argument supplied".

This one changes process scheduling state from a typo.

## 2. `set-environment-variable!` cannot fail, and silently sets nothing

`platform.zig:727` does `_ = setenv(name_z, value_z, 1)`, discarding the return — so `setEnvVarFn`'s `raiseFileError` branch is unreachable.

```scheme
(set-environment-variable! "KA=B" "v")   ⟹ returns normally
(get-environment-variable "KA=B")        ⟹ #f          ; nothing was set
```

Absent under `"KA"`, under `"KA=B"`, and from `(get-environment-variables)`. Same for an empty name. `unsetEnv` has the same shape.

**Control:** `"KA_OK"` sets and reads back `"v"` correctly.

## 3. Four ignored optional arguments

A mistyped optional is discarded rather than rejected:

| call | result |
|---|---|
| `(create-directory p "notanum")` | mode 0755 |
| `(create-fifo p "notanum")` | mode 0664 |
| `(create-temp-file 42)` | default prefix |
| `(set-file-times p "garbage")` or `(… 5.0 5.0)` | **both stamps set to now** |

**Control for each:** the same argument as an out-of-range *fixnum* raises correctly — e.g. `#o10000` ⟹ `"mode value out of range"`.

`set-file-times` is the worst of the four: a typo silently rewrites both timestamps to the current time rather than failing.

## Related, same file

Surplus arguments are also accepted — `.variadic = N` has no upper bound, so `(file-info p #t 'extra)` ignores the extra. Too *few* arguments is reported correctly.

**Verified:** fresh ReleaseSafe build at `03bda165`, isolated `KAAPPI_HOME`; every row and control reproduced independently.

Issue #1976, opened 2026-07-31
file-info aborts the process on any devfs path — signed st_dev @intCast into u64


baijum:
**Category:** SRFI 170 filesystem · **Severity:** crash · **Found by:** systematic audit v2, Phase 2.12 (#1890)

`file-info` on anything under `/dev` aborts. Uncatchable by `guard`; exit 134.

```scheme
(guard (e (#t 'caught)) (file-info "/dev/null" #t))
```

```text
thread 33127508 panic: integer does not fit in destination type
exit 134
```

## Mechanism

`src/primitives_filesystem.zig:134` — `.dev = @intCast(stat_buf.dev)` narrows a **signed** `dev_t` into `u64`. macOS devfs reports `st_dev = 2998493978`, which is `−1296473318` as `i32`.

## Discriminating controls

| path | result | what it isolates |
|---|---|---|
| `/tmp` | `directory` | positive `st_dev` is fine — not "file-info is broken" |
| `/dev/null` | **abort** | character device |
| **`/dev/fd`** | **abort** | **a *directory* with `st_rdev = 0`** |

`/dev/fd` is the useful one: it is not a device node and has no `rdev`, yet it still aborts — so the fault is `st_dev`, not `rdev`, and not the file *type*.

Identical under `-Dgc-stress=true`, so not a GC interaction.

## Why this matters more than a device-node edge case

`/dev/null` is an ordinary path for ordinary code — a build script asking whether it exists, a test fixture, a config default. Any `file-info` on a devfs path takes down the process with no catchable error.

It also makes device nodes **untestable**: the audit could not assert anything about `file-info` on them, because reaching them aborts first.

## Platform scope

- **macOS: verified affected.**
- **OpenBSD: inferred affected** — `dev_t = i32`. Read from the type declaration, not run.
- Linux (`statx` + `makedev`, u64), Windows (`c_uint`), FreeBSD and NetBSD (`u64`) are safe by construction — also inferred from declarations.

## Suggested fix

`@bitCast` to the unsigned width before widening, or store `dev`/`rdev` as signed. Whichever is chosen, `rdev` at `:135` needs the same treatment — it is the same signed type and simply happens not to be negative on the paths that currently reach it.

**Verified:** fresh ReleaseSafe build at `03bda165`, isolated `KAAPPI_HOME`, cache cleared; the abort, its exit code, the `guard` case and all three controls reproduced independently.

PR #1985, opened 2026-07-31
Phase 2.12: filesystem audit — 69 → 428 assertions, and file-info aborts on any /dev path

baijum:
Phase 2.12 of the v2 audit campaign (#1890). Extends `tests/scheme/audit/primitives_filesystem-audit.scm` — 68 specs (the strategy said 69; the table has 68) and 102 syscall sites against a 177-line test.

**69 → 427 enabled assertions**, 177 → 1097 lines. 18 disabled under 9 markers, each with an enabled control. Full `run-all.sh` on the rebased tree: **635 pass, 0 fail; R7RS 1395/1395**.

**Coverage: 10 specs had never been mentioned** by the old test (`file-info:device/inode/rdev/blksize/blocks/gid`, `file-info-socket?/-device?`, `set-current-directory!`, `user-info:full-name`). **After: 0 of 68 untouched.**

## #1976 — `file-info` aborts on any devfs path

```scheme
(guard (e (#t 'caught)) (file-info "/dev/null" #t))   ⟹ exit 134, uncatchable
```

`.dev = @intCast(stat_buf.dev)` narrows a **signed** `dev_t` into `u64`; macOS devfs reports a value that is negative as `i32`.

**The control that isolates it:** `/dev/fd` is a **directory** with `st_rdev = 0` and still aborts — so the fault is `st_dev`, not `rdev`, and not the file type. `/tmp` is fine. Identical under gc-stress.

This also made device nodes **untestable** — the audit could not assert anything about them, because reaching them aborts first.

## #1977 — silently discarded arguments

`(nice "x")` **renices the process** and returns 1; the range check exists (`(nice 999999999999)` raises) but the type test does not. `set-environment-variable!` discards `setenv`'s return, so its error branch is unreachable and `"KA=B"` sets **nothing** while returning normally. Four optional arguments are discarded when mistyped — worst is `set-file-times`, where a typo silently rewrites both stamps to *now*.

## #1978 — errno discarded at ~102 sites

SRFI 170 mandates `posix-error?`/`-name`/`-message`; none is exported. ENOENT, ENOTDIR, EACCES, ELOOP and ENAMETOOLONG all collapse to one generic string per procedure. Separately, pure argument-range failures report as `file-error?` — control: a genuine type error correctly answers `#f`.

## Confirmed correct — the larger half

The **NUL guard is uniform across all 24 path-taking procedures**, rejecting before the syscall. Symlink loops, self-loops, 5000-char paths, a directory deleted mid-traversal, a 2000-entry directory across `getdents` refills, and `truncate-file` to 2^47 all terminate cleanly. ~~3000 unclosed directory streams do not exhaust the fd table~~ — **retracted, see #1993.** That result held only because macOS and Linux CI run with `ulimit -n 1048576`. At a normal limit they are not reclaimed in time: at `ulimit -n 256` exactly **253** unclosed opens succeed, and `--gc-stats` reports `Collections: 0` — the process dies at the descriptor limit holding 5165 live objects, under the 8192-object threshold. `gc_sweep.zig:537` *does* finalise `DirectoryObject` correctly; the sweep is right, and nothing triggers it on descriptor pressure. Ports behave identically (253 / 1747), so this is every fd-holding object. The assertion exhausted the table on `openbsd-test` and `netbsd-test`, taking down the following block, and is now disabled behind a FAIL marker.

**`--sandbox` gating is fully coherent** — unlike `primitives_sysinfo.zig` in Phase 2.1, all 68 specs carry the same flags and 70/70 probes are present normally and absent under `--sandbox`. No per-name drift.

Also correct: 12-bit mode validation shared across all 4 call sites, umask vs explicit mode, `create-temp-file` unique and `#o600`, `user-info`/`group-info` returning `#f` for unknown identifiers exactly as the SRFI requires, `terminal?` under both a pty and redirection, non-ASCII filenames with codepoint-correct lengths, D4 (68 arities, zero mismatches), and gc-stress clean across 60 iterations with 60 live `file-info` objects retained.

## Stated limits

Socket `file-info` verified interactively but **not asserted** — no portable way to bind one from Scheme. Invalid-UTF-8 filenames untested because **APFS refuses to create them**; that needs a Linux/ext4 run. `read-symlink`'s "target too long" branch is unreachable where `PATH_MAX = 1024`. #1976 on OpenBSD is inferred from `dev_t = i32`, not run.

One design consequence worth recording: `set-file-times`' fixnum sentinels (−1 = now, −2 = unchanged) make timestamps −1 and −2 unrepresentable.

### DIFF
--- a/tests/scheme/audit/primitives_filesystem-audit.scm
+++ b/tests/scheme/audit/primitives_filesystem-audit.scm
@@ -572,7 +572,7 @@
    (test-equal 0 (file-info:rdev (file-info (string-append dir "/reg") #t)))
    (test-equal 0 (file-info:rdev (file-info (string-append dir "/fifo") #f)))))
 
-;; FAIL: TBD (file-info panics — SIGABRT, not catchable — on any path whose
+;; FAIL: #1976 (file-info panics — SIGABRT, not catchable — on any path whose
 ;; st_dev is negative when read as i32; on macOS that is every entry under
 ;; /dev, whose devfs st_dev is 2998493978 = -1296473318 as i32.  The cast is
 ;; `.dev = @intCast(stat_buf.dev)` at src/primitives_filesystem.zig:134.
@@ -691,7 +691,7 @@
 (with-scratch
  (lambda (dir)
    ;; -- create-directory mode --
-   ;; FAIL: TBD (a non-fixnum mode is silently discarded and 0o755 used:
+   ;; FAIL: #1977 (a non-fixnum mode is silently discarded and 0o755 used:
    ;; `if (args.len > 1 and types.isFixnum(args[1]))` in createDirectoryFn)
    ;; (test-equal #t (raised? (lambda () (create-directory (string-append dir "/m1") "notanum"))))
    (test-equal #t (raised? (lambda () (create-directory (string-append dir "/m2") #o10000))))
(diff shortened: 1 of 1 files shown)


baijum:
Named assertions did their job on the first run: `freebsd-test` printed `FAIL (memv (file-info:rdev fi) '(0 -1))` rather than a bare `#f`, naming the defect outright.

Three fixes in fdf9dbe5:

1. **rdev** — POSIX defines `st_rdev` only for character- and block-special files; elsewhere it is unspecified. Two guesses were both wrong (0, then {0, -1} for FreeBSD's NODEV), and FreeBSD 14.3 still failed for a *regular file* while passing for a *fifo* in the same run. Both value assertions dropped; `exact-integer?` is the portable contract.

2. **fd exhaustion (#1993)** — the 3000-unclosed-`open-directory` assertion was the openbsd/netbsd abort. The GC does reclaim these, but only on its allocation-count threshold, never on descriptor pressure: at `ulimit -n 256` exactly **253** unclosed opens succeed (no collection runs at all); at 1024, **1745**. Ports behave identically (253 / 1747), so it is every fd-holding object. Disabled behind a FAIL marker, replaced with the open/close cycle the program actually controls.

3. **Log noise** — the three stray `#f` lines were this suite's own top-level cleanup calls, echoed because kaappi prints non-void top-level values in scripts. Deliberate, but documented only in `docs/dev/fuzzing.md` — filed as #1994.

Verified at `ulimit -n` 256, 512, 1024 and default: **427 passes, 0 failures**, two lines of output.

Issue #2053, opened 2026-08-01
[SRFI-146] mapping-map and mapping-find run their whole fold twice — the first result is discarded


baijum:
**Category:** SRFI
**Spec reference:** SRFI 146, *Mapping and folding* — "`(mapping-map proc comparator mapping)` **Applies `proc`**, which returns two values, on two arguments, the key and value of each association of `mapping` in increasing order of the keys and returns a newly allocated mapping…". Note that the neighbouring `mapping-find`/`-count`/`-any?`/`-every?` entries each carry an explicit licence — "There are no guarantees how many times and with which keys and values `predicate` is invoked" — and the `mapping-map` entry deliberately does not.
**Severity:** edge-case

Both procedures contain a complete first copy of their fold whose result is discarded, then recompute it:

```scheme
;; lib/srfi/146.sld:332
(define (mapping-map proc comparator m)
  (let ((lt ...) (eq ...))
    (%rbt-fold (lambda (k v acc) ...) %empty (%mapping-tree m))    ; <- value thrown away
    (%make-mapping comparator
      (%rbt-fold (lambda (k v acc) ...) %empty (%mapping-tree m))))) ; <- same fold again

;; lib/srfi/146.sld:304
(define (mapping-find pred m failure)
  (%rbt-fold (lambda (k v acc) ...) 'not-found (%mapping-tree m))   ; <- value thrown away
  (let ((result (%rbt-fold ...))) ...))
```

**Minimal reproduction:**

```scheme
(import (scheme base) (srfi 128) (srfi 146))
(define c (make-default-comparator))
(define m (mapping c 1 'a 2 'b 3 'c 4 'd 5 'e))

(let ((n 0))
  (mapping-map (lambda (k v) (set! n (+ n 1)) (values k v)) c m)
  n)                                       ;; expected: 5   actual: 10

(let ((n 0))
  (mapping-find (lambda (k v) (set! n (+ n 1)) (= k 3)) m (lambda () (values #f #f)))
  n)                                       ;; 6 -- two passes of 3
```

**Discriminating control:** the neighbouring folds in the same file run their body exactly once per association, so this is dead code in two procedures rather than a design choice:

```scheme
(let ((n 0)) (mapping-for-each (lambda (k v) (set! n (+ n 1))) m) n)        ;; => 5
(let ((n 0)) (mapping-filter   (lambda (k v) (set! n (+ n 1)) #t) m) n)     ;; => 5
[...]

Issue #2052, opened 2026-08-01
[SRFI-146] the ten submapping comparison predicates reject the single-mapping form the signature allows


baijum:
**Category:** SRFI
**Spec reference:** SRFI 146, *Submappings* — the five signatures are `(mapping=? comparator mapping1 mapping2 ...)`, `(mapping<? …)`, `(mapping>? …)`, `(mapping<=? …)`, `(mapping>=? …)`. `mapping2 ...` is zero-or-more, so one mapping is a legal argument list, and the prose is vacuously satisfied by it: "Returns `#t` if each mapping `mapping` contains the same associations".
**Severity:** edge-case

Each `case-lambda` starts at three arguments, so the one-mapping form raises instead of returning `#t`:

```scheme
;; lib/srfi/146.sld:413
(define mapping=?
  (case-lambda
    ((vcmp m1 m2) ...)              ; no ((vcmp m1) #t) clause
    ((vcmp m1 m2 . rest) ...)))
```

**Minimal reproduction:**

```scheme
(import (scheme base) (srfi 128) (srfi 146) (srfi 146 hash))
(define c (make-default-comparator))
(define m (mapping c 1 'a))

(mapping=?  c m)      ;; expected: #t   actual: raises (no matching case-lambda clause)
(mapping<?  c m)      ;; expected: #f   actual: raises
(mapping<=? c m)      ;; expected: #t   actual: raises
(mapping>?  c m)      ;; expected: #f   actual: raises
(mapping>=? c m)      ;; expected: #t   actual: raises
(hashmap=?  c (hashmap c 1 'a))   ;; expected: #t   actual: raises
```

Ten exported names in total (five in each library).

**Discriminating control:** the variadic set-theory procedures in the same file *do* accept the one-mapping form, so the omission is specific to the comparison predicates rather than a house convention:

```scheme
(mapping-size (mapping-union m))          ;; => 1  (accepted)
(mapping-size (mapping-intersection m))   ;; => 1  (accepted)
(mapping-size (mapping-difference m))     ;; => 1  (accepted)
```

**Oracle:** the reference implementation has an explicit clause for it in all five:

```scheme
;; srfi-146.tgz, srfi/146.scm:438
(define mapping=?
  (case-lambda
    ((comparator mapping) (assume (mapping? mapping)) #t)
    ((comparator mapping1 mapping2) (%mapping=? comparator mapping1 mapping2))
    ...))
```

[...]

Issue #2050, opened 2026-08-01
[SRFI-146] mapping-any?/mapping-every? return the predicate's value, not a boolean — the hashmap twins do not


baijum:
**Category:** SRFI
**Spec reference:** SRFI 146, *The whole mapping* — "`(mapping-any? predicate mapping)` **Returns `#t`** if any association of the mapping `mapping` satisfies `predicate` (in the sense of `mapping-find`), **or `#f` otherwise**." — and identically for `mapping-every?`.
**Severity:** edge-case

The ordered library implements both as an accumulating fold that keeps the predicate's own return value:

```scheme
;; lib/srfi/146.sld:319
(define (mapping-any? pred m)
  (%rbt-fold (lambda (k v acc) (or acc (pred k v))) #f (%mapping-tree m)))
(define (mapping-every? pred m)
  (%rbt-fold (lambda (k v acc) (and acc (pred k v))) #t (%mapping-tree m)))
```

so a predicate returning a non-`#f`, non-`#t` value leaks that value out.

**Minimal reproduction:**

```scheme
(import (scheme base) (srfi 128) (srfi 146))
(define m (mapping (make-default-comparator) 1 'a 2 'b))

(mapping-any?   (lambda (k v) v) m)   ;; expected: #t   actual: a
(mapping-every? (lambda (k v) v) m)   ;; expected: #t   actual: b
```

**Discriminating control:** the hash siblings, which accumulate into a boolean flag, return exactly `#t`/`#f` for the identical predicate — so the two halves of the same SRFI disagree:

```scheme
(import (srfi 146 hash))
(define hm (hashmap (make-default-comparator) 1 'a 2 'b))
(hashmap-any?   (lambda (k v) v) hm)  ;; => #t  (correct)
(hashmap-every? (lambda (k v) v) hm)  ;; => #t  (correct)
```

and a boolean-returning predicate is correct on both, which is why the existing suite never caught it:

```scheme
(mapping-any? (lambda (k v) (eq? v 'a)) m)   ;; => #t  (correct)
(mapping-any? (lambda (k v) #f) m)           ;; => #f  (correct)
```

**Oracle:** the reference implementation returns literal booleans — `mapping-any?` escapes with `(return #t)` and falls through to `#f`, and `mapping-every?` is `(not (mapping-any? (lambda (k v) (not (predicate k v))) mapping))` (`srfi-146.tgz`, `srfi/146.scm:302` and `:313`).

[...]

Issue #2049, opened 2026-08-01
[SRFI-146] hashmap-ref/default calls a procedural default as a thunk instead of returning it


baijum:
**Category:** SRFI
**Spec reference:** SRFI 146, *Accessors* — "`(mapping-ref/default mapping key default)` Semantically equivalent to, but may be more efficient than, the following code: `(mapping-ref mapping key (lambda () default))`"
**Severity:** wrong-result

`hashmap-ref/default` forwards its `default` to SRFI 69's `hash-table-ref`, whose third argument is a **failure thunk**, not a default value:

```scheme
;; lib/srfi/146/hash.sld:72
(define (hashmap-ref/default m key default)
  (hash-table-ref (%hm-ht m) key default))
```

`hashTableRefFn` (`src/primitives_hashtable.zig:490`) calls the argument when `types.isProcedure` holds and only otherwise returns it, so a `default` that happens to be a procedure is invoked with no arguments and its *result* is returned in place of the procedure the caller asked for.

**Minimal reproduction:**

```scheme
(import (scheme base) (srfi 128) (srfi 146 hash))
(define hm (hashmap (make-default-comparator) 1 'a))

(hashmap-ref/default hm 99 (lambda () 'called))
;; expected: the procedure itself   actual: called
```

Realistically this bites any hashmap whose values are procedures — a dispatch table, a memo of thunks — where the natural default is another procedure:

```scheme
(define handlers (hashmap (make-default-comparator) 'get (lambda () 'GET)))
(define fallback (lambda () 'NOT-FOUND))
((hashmap-ref/default handlers 'post fallback))
;; expected: NOT-FOUND   actual: raises -- the default was already called and
;;                                'NOT-FOUND is not applicable
```

**Discriminating control:** the ordered sibling, which does not go through `hash-table-ref`, returns the procedure unchanged — so this is the SRFI-69 argument-shape confusion and not an intended reading of the spec:

```scheme
(import (srfi 146))
(procedure? (mapping-ref/default (mapping (make-default-comparator) 1 'a) 99 (lambda () 'called)))
;; => #t   (correct)
```

A second control shows non-procedure defaults are unaffected, which is why the existing suite never noticed:

```scheme
(hashmap-ref/default hm 99 'zzz)   ;; => zzz  (correct)
(hashmap-ref/default hm 99 0)      ;; => 0    (correct)
```

[...]

PR #2070, opened 2026-08-01
Phase 3.4: SRFI 146 audit — 668 assertions over all 161 exports, and (srfi 146 hash) never uses the comparator it is given

baijum:
Phase 3.4 of the systematic audit v2 (`docs/audit-strategy.md`, tracking #1890).

SRFI 146 exports **161 names** across `(srfi 146)` and `(srfi 146 hash)`. Its
one test file used manual pass/fail counters, reported **0 tests** to
`kaappi test`, and touched **53** of the 161. Two new SRFI-64 files bring that
to **668 assertions touching all 161** — every one with a name string.

| file | assertions | disabled | what it is |
|---|---:|---:|---|
| `tests/scheme/srfi/srfi146-reference.scm` | 167 | 7 | the SRFI's **own** reference suite, ported verbatim |
| `tests/scheme/srfi/srfi146-differential.scm` | 501 | 26 | ordered-vs-hash differential, comparator plumbing, edge cases, scale |

## The differential is what found most of it

The two libraries expose the same 66-name surface under two prefixes, and the
spec says they differ only in iteration order and in the 29 extra ordered-only
procedures. So one parameterised body runs over both through an operation
table and demands the same answers — an oracle needing nothing external.
Nothing asserts hashmap iteration order; only set equality, sizes, and
key/value correspondence.

It disagreed in four places, each of which turned out to be a real defect. The
sharpest is #2044: a comparator whose equality is `=` merges `1` and `1.0` in
the ordered library and splits them in the hash one, because `(srfi 146 hash)`
never uses the comparator it was handed.

The reference suite added three more of its own, including two assertions the
SRFI itself ships for `mapping=?`'s comparator-identity rule.

## Issues filed

| # | Finding |
|---|---|
| #2044 | **`(srfi 146 hash)` discards its comparator.** Every constructor builds a bare `(make-hash-table)`; the comparator is stored, returned by `hashmap-key-comparator`, and never consulted. A `=`-equality comparator gives `(hashmap-size (hashmap numc 1 'a 1.0 'b))` ⟹ **2**; the ordered sibling with the same equality ⟹ **1**. Case-insensitive string keys behave the same way. Knock-on: no comparator can route this library around #2023. |
| #2045 | **The constructors keep the LAST duplicate key**, where the spec says "Earlier associations with equal keys take precedence over later arguments" — the one distinction it explicitly contrasts with `mapping-set`. `mapping-adjoin` (first wins) and `mapping-set` (last wins) are both correct, which is what makes this a defect and not a house convention. Six names: `mapping`, `mapping/ordered`, `mapping-unfold`, `mapping-unfold/ordered`, `hashmap`, `hashmap-unfold`. Reference oracle: `mapping` ⟶ `mapping-unfold` ⟶ `mapping-adjoin`. |
| #2046 | **`mapping-key-predecessor`/`-successor` invoke `failure` unconditionally** — it is the fold's *seed*. A thunk that raises (the natural spelling of "this cannot happen") raises even when the answer exists: `(mapping-key-predecessor m 3 (lambda () (error "no predecessor")))` on `{1,2,3}` should be `2`. |
| #2047 | **`mapping=?`/`hashmap=?` omit the key-comparator identity check.** Spec: "it is explicitly not an error if `mapping=?` is invoked on mappings that do not share the same (key) comparator. In that case, `#f` is returned." Caught by two assertions in the SRFI's own suite; the reference opens `%mapping=?` with exactly the missing `eq?`. |
| #2048 | **`make-mapping-comparator` supplies no ordering predicate and `make-hashmap-comparator` no hash function** — both pass `#f`. The consequence is the one the spec names outright: mappings cannot key a mapping. Five reference assertions. |
| #2049 | **`hashmap-ref/default` calls a procedural default as a thunk** — it forwards to SRFI 69's `hash-table-ref`, whose third argument is a failure thunk. Bites any hashmap of procedures. `mapping-ref/default` returns it unchanged. |
| #2050 | **`mapping-any?`/`mapping-every?` return the predicate's value**, not `#t`/`#f` as the spec states. The hashmap twins return a boolean — a differential the reference confirms. |
| #2052 | **The ten submapping comparison predicates reject the single-mapping form** their signature allows. The variadic set-theory procedures in the same file accept it; the reference has an explicit clause for it in all five. |
| #2053 | **`mapping-map` and `mapping-find` run their whole fold twice** — the first result is discarded dead code, so `proc` is applied 2× per association. `for-each`/`filter`/`count` beside them are single-pass, and so is `hashmap-map`. |

Also commented on **#2023** rather than re-filing: `(srfi 146 hash)` has no
comparator that avoids the depth-8 `equal?`-hash cliff (both `make-equal-` and
`make-default-comparator` miss 100/100), and the damage is not only lookups —
`hashmap-union` of a hashmap with an **equal** hashmap returns 84–89 entries
instead of 50, `hashmap-xor` of a set with itself returns 58–73 instead of 0,
and `hashmap-delete` of a present key silently no-ops.

## What came back clean

The Phase 3 preamble is right that a high untested-export count is not itself
evidence of bugs. These are now instrumented rather than assumed:

- **All four `mapping-search` continuations** (`insert`, `ignore`, `update`,
  `remove`) on **both** libraries, including `success` receiving the mapping's
  own stored key, and the two-value return in every case.
- **Escaping continuations** out of `fold`, `for-each`, `filter`, `map`,
  `find` and `update` on both libraries — all unwind cleanly and leave the
  source mapping intact.
- **Purity.** Every pure variant is called and its argument re-measured; the
  result is then mutated and the source re-read. All eight pure/linear pairs
  behave.
- **The whole ordered-only surface** — min/max key/value/entry (and their
  six empty-mapping raises), predecessor/successor including absent `obj`,
  all five ranges and their five `!` twins, `split`, `catenate`,
  `map/monotone`, `fold/reverse`, and the four `/ordered` aliases.
- **Set algebra on disjoint, identical and overlapping inputs**, plus empty
  operands and three-way forms, for union/intersection/difference/xor and
  their `!` twins. `mapping-difference` against the union of the rest is
  correct; union/intersection draw from the first mapping as specified.
- **Comparator plumbing on the ordered side** — a custom equality is honoured,
  the key comparator survives filter/copy/partition/range, an inconsistent
  ordering predicate does not crash the tree, and the default comparator
  handles integer/string/symbol/char/list keys in one mapping.
- **Scale and tree structure.** The ordered library's delete reuses the
  *insertion* balance routine and so cannot restore black-height. 1000
  sequential inserts, deletion of every even key, and 1500 interleaved
  insert/delete rounds over a sliding window all keep every surviving key
  findable and the key list strictly increasing.

## Disabled tests

33 assertions carry `;; FAIL: #NNNN` markers (7 + 26). **Mutation-tested in
both files**: uncommenting them makes *exactly* those fail — 7 of 167 and 26
of 501 — with every enabled assertion still passing. Enabled controls sit
beside most of them (`adjoin`/`set`/`alist->` precedence, the ordered `=`
comparator, depth-5 and depth-7 hash keys, `mapping=?` on a shared
comparator), and several current-behaviour pins so a change is never silent.

## Verification

- Fresh ReleaseSafe build at `08828884`, isolated `KAAPPI_HOME` per shell.
- `kaappi test tests/scheme/srfi` — 31,322 passed, only the pre-existing
  `srfi150.scm` error (kaappi#1832, unrelated).
- `bash tests/scheme/run-all.sh` green.
- Both files run in 0.37s combined. `N` in the scale group is sized for the
  QEMU-emulated CI legs against the 60s per-file timeout, not for this
  machine.
- Every finding is deterministic and reproducible across runs; none is a
  crash or nondeterminism, so `-Dgc-stress=true` has nothing to localise here.
- Oracles used: the SRFI 146 spec text (quoted per issue), the SRFI's own
  reference implementation from `srfi-146.tgz`, and the ordered/hash
  differential.

### DIFF
--- a/tests/scheme/srfi/srfi146-differential.scm
+++ b/tests/scheme/srfi/srfi146-differential.scm
@@ -973,8 +973,22 @@
   ;;         (h2 (let loop ((i 0) (m (hashmap c)))
   ;;               (if (= i 40) m (loop (+ i 1) (hashmap-set m (deep 12 i) i))))))
   ;;     (hashmap-size (hashmap-union h1 h2))))
-  (test-assert "pins #2023: hash keys 12 levels deep go missing"
-    (> (hash-misses 12 40) 0)))
+  ;; #2023 is pinned by the two disabled assertions above, not by a miss
+  ;; count.  An earlier version asserted `(> (hash-misses 12 40) 0)` -- that
+  ;; the bug is still present -- and it FAILED on CI's `ubuntu-latest, Debug`
+  ;; leg.  Past MAX_HASH_DEPTH the hash degenerates to the key's pointer, so
+  ;; whether two structurally-equal keys collide is an accident of allocation:
+  ;; measured 31 and 32 misses of 40 at `deep 8` across two runs of one
+  ;; ReleaseSafe binary, and 3 vs 40 of 40 at `deep 9`.  No miss count is
+  ;; assertable past the limit, in either direction.
+  ;;
+  ;; What IS stable in every run and both build modes is the cliff itself, so
+  ;; that is what this pins.  Note `deep` builds a FLAT list of length depth+1,
+  ;; not a nested structure, and the hash walks the spine -- so the limit is
+  ;; reached at `deep 8` (length 9) and `deep 7` (length 8) is the last fully
+  ;; findable case.  Measured 0/40 at every depth from 1 to 7, both builds.
+  (test-equal "keys at the depth limit are all findable (the #2023 cliff)"
+              0 (hash-misses 7 40)))
 
 (let ((runner (test-runner-current)))
   (test-end "srfi146-differential")
--- a/tests/scheme/srfi/srfi146-differential.scm
+++ b/tests/scheme/srfi/srfi146-differential.scm
@@ -1,0 +1,981 @@
+;;; SRFI 146 — ordered/hash differential, comparator plumbing, edge cases.
+;;;
+;;; (srfi 146) and (srfi 146 hash) expose the same 66-name surface under two
+;;; prefixes.  The spec says they differ only in iteration order and in the 29
+;;; extra ordered-only procedures, so running one parameterised body over both
+;;; and demanding the same answers is an oracle that needs nothing external.
+;;; Nothing here asserts hashmap iteration order — only set equality and sizes.
+;;;
+;;; The second half covers the 29 ordered-only names, comparator plumbing, the
+;;; spec's own worked examples, and the edge cases the generic bodies skip.
+;;;
+;;; Every assertion carries a name string; the shared ones are built with
+;;; string-append from the library label, which SRFI-64 logs correctly.
+;;;
+;;; Added by Systematic audit v2, Phase 3.4.
+
+(import (scheme base) (scheme write) (scheme process-context)
+        (srfi 64) (srfi 128) (srfi 146) (srfi 146 hash))
+
+(define c (make-default-comparator))
+
+;;; ---------------------------------------------------------------- helpers
+
+(define (sortl lst)                     ; insertion sort on exact integers
+  (define (insert x sorted)
+    (cond ((null? sorted) (list x))
+          ((< x (car sorted)) (cons x sorted))
+          (else (cons (car sorted) (insert x (cdr sorted))))))
+  (let loop ((l lst) (acc '()))
+    (if (null? l) acc (loop (cdr l) (insert (car l) acc)))))
+
+(define (sorted-alist al)               ; ((k . v) ...) sorted by integer key
+  (map (lambda (k) (assv k al)) (sortl (map car al))))
+
+;;; ------------------------------------------------- the two operation tables
+
+(define (ordered-op sym)
+  (case sym
+    ((make) mapping)                    ((unfold) mapping-unfold)
+    ((pred) mapping?)                   ((contains?) mapping-contains?)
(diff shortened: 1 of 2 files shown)


baijum:
Local suite, final state after the branch was committed and the binary rebuilt at `d1ace726`:

```
Scheme files: 644 pass, 0 fail, 0 timeout, 0 skipped
R7RS suite:   1395 pass, 0 fail
Total:        2039 pass, 0 fail
```

One footgun worth recording for the other audit units, because it cost a diagnostic detour and names none of your own tests when it fires. Two `tests/scheme/compile/*.sh` suites (`compile-import-error-703.sh`, `compile-preamble-gc-700.sh`) failed on the run *before* this one, with the bundled fixture binary dying on `fatal: invalid embedded bytecode (wrong version or format)`.

It is not a regression, and it is not concurrency. `bundle_fixture_binary` compiles the fixture with `zig-out/bin/kaappi` — which stamps **that binary's** git build id into the `.sbc` — and then runs `zig build -Dbundle=$sbc`, which rebuilds the interpreter from the **current working tree**. Committing changes the working tree's build id, so a `zig-out/bin/kaappi` built before the commit produces a `.sbc` the freshly rebuilt interpreter rejects. `shell-common.sh`'s own comment predicts exactly this ("A cached .sbc of our own would instead go stale against the rebuilt binary's build id and fail with `invalid embedded bytecode`").

**`zig build` again after committing, before running the compile suite.** Both suites pass individually and in the full run once the binary and the tree agree — verified above.

Issue #2075, opened 2026-08-01
[R7RS 4.2.3] A begin-wrapped internal define in a let body escapes to the global environment unless an enclosing procedure exists


baijum:
**Category:** R7RS
**Spec reference:** R7RS §4.2.3, page 17 (Sequencing) — *"This form of begin can appear as part of a ⟨body⟩, or at the outermost level of a ⟨program⟩, or at the REPL, or directly nested in a begin that is itself of this form. It causes the contained expressions and definitions to be evaluated exactly as if the enclosing begin construct were not present."*
**Severity:** wrong-result (silent; clobbers an unrelated global)

A `begin`-wrapped internal definition in the body of a `let`/`let*`/`letrec`
does not become a local binding. It compiles to a **global** definition — so
the enclosing `let` binding of the same name is not shadowed (the spec says it
must be), and a top-level variable of that name is silently overwritten.

It only happens when there is no enclosing procedure: put the identical `let`
inside any `lambda` and it is correct. The unwrapped form is correct
everywhere.

**Minimal reproduction:**

```scheme
(import (scheme base) (scheme write))

(define g 'global)
(write (let ((g 'outer)) (begin (define g 'inner)) g))  ;; expected: inner   actual: outer
(newline)
(write g)                                               ;; expected: global  actual: inner
(newline)
```

```text
outer
inner
```

Both lines are wrong, and the second is the damaging one: a definition written
inside a `let` body has escaped the `let` and replaced an unrelated top-level
binding, with no diagnostic and exit 0.

**Discriminating controls** — three independent axes, each isolating the
`begin` + no-enclosing-procedure conjunction:

| probe | value of the `let` | global afterwards |
|---|---|---|
| `(let ((g 'outer)) (define g 'inner) g)` — no `begin` | `inner` ✓ | `global` ✓ |
| `(let ((g 'outer)) (begin (define g 'inner)) g)` — top level | **`outer`** ✗ | **`inner`** ✗ |
| `((lambda () (let ((g 'outer)) (begin (define g 'inner)) g)))` | `inner` ✓ | `global` ✓ |
| `(let ((z 1)) (let ((g 'outer)) (begin (define g 'inner)) g))` — nested, still no lambda | **`outer`** ✗ | **`inner`** ✗ |
| `((lambda () (let ((g 'outer)) (begin (begin (define g 'inner))) g)))` — nested `begin` | `inner` ✓ | `global` ✓ |

[...]

Issue #2060, opened 2026-08-01
[SRFI-158] SRFI 1 procedures are still native drivers: the spec's generator-unfold example raises, and gtake cannot be resumed inside fold/filter/any/every or hash-table-walk


baijum:
**Category:** SRFI / Doc-truth
**Spec reference:** [SRFI 158](https://srfi.schemers.org/srfi-158/srfi-158.html), `generator-unfold` — the entry's own worked example:

```scheme
(generator-unfold (make-for-each-generator string-for-each "abc") unfold)
  ⇒ (#\a #\b #\c)
```

and `make-for-each-generator` — "A generator constructor that converts **any collection** obj to a generator that returns its elements using a for-each procedure appropriate for obj. This must be a procedure that when called as `(for-each proc obj)` calls proc on each element of obj."
**Severity:** wrong-result (documented example does not run)

SRFI 158's `unfold` example raises. So does converting a hash table — or
anything else whose `for-each` is a native driver — into a generator, which is
the whole point of `make-for-each-generator`.

The root cause is one thing: SRFI 1's higher-order primitives, `hash-table-walk`,
and `assoc`/`member` with a custom predicate are still **native drivers**, so a
continuation captured inside their callback cannot be resumed once the native
frame has returned. #1347 trampolined the `map`/`for-each`/`vector-map`/
`vector-for-each`/`string-for-each`/`dynamic-wind`/`force` family and is closed;
the SRFI 1 family was never in its scope.

The blast radius inside SRFI 158 is wider than the one example, because **four**
of its constructors are built on `make-coroutine-generator` and therefore on
`call/cc`: `make-coroutine-generator`, `make-for-each-generator`,
`make-unfold-generator`, and — the common one — **`gtake`**.

**Minimal reproduction:**

```scheme
(import (scheme base) (srfi 158) (srfi 1) (srfi 69))

;; 1. The spec's own example.
(generator-unfold (make-for-each-generator string-for-each "abc") unfold)
;; expected: (#\a #\b #\c)
;; actual:   continuation cannot resume across a returned native call

;; 2. gtake — nothing exotic — inside any SRFI 1 driver.
(let ((g (gtake (make-range-generator 0 100) 5)))
  (fold (lambda (i acc) (cons (g) acc)) '() '(a b c)))
;; expected: (2 1 0)   actual: continuation cannot resume across a returned native call

;; 3. make-for-each-generator over a hash table.
(let ((h (make-hash-table equal?)))
[...]

Issue #682, opened 2026-07-01
syntax-rules: no ellipsis-depth validation — mismatched pattern-variable depths silently produce wrong output instead of a syntax error


baijum:
## Bug

R7RS requires that "each pattern variable... must be used in the template
at the same or lower ellipsis depth at which it was matched" and that
using pattern variables of different ellipsis depths together under a
single `...` in the template is a syntax violation. `src/expander.zig`
does not validate this at all: `instantiateEllipsis` mixes bindings of
different depths under the same repeat count without checking they are
compatible, and depth-mismatched variables silently evaluate to `()`
(nil) instead of raising an error.

## Location

`src/expander.zig`:
- `instantiateEllipsis` (lines 571-639) computes a single `repeat_count`
  from whichever list-bindings are referenced (lines 574-585), then for
  each iteration unpacks *every* list binding referenced in
  `elem_template`, regardless of whether its `depth` actually matches the
  ellipsis nesting used in the template (lines 600-630).
- The `is_list` binding used at depth 0 without its own `...` falls
  through in `instantiateTemplate` (lines 426-431) and returns `b.value`,
  which for a list binding was only ever initialized to `types.NIL`
  (`src/expander.zig:338` sets `.value = types.NIL` for ellipsis
  bindings) and never updated — so any depth mismatch silently yields
  `()` rather than an error.

```zig
        // 1. Pattern variable -- substitute with matched value (from use site)
        for (bindings) |b| {
            if (std.mem.eql(u8, b.name, name)) {
                if (!b.is_list) return b.value;
                // Shouldn't use a list binding at depth 0 without ellipsis
                return b.value;   // <-- always NIL, no error raised
            }
        }
```

## Repro

```scheme
;; a is bound at ellipsis depth 2 (nested list), b at depth 1.
;; Using them together under a single ... is a depth mismatch and
;; should be a syntax error.
(define-syntax bad-depth
  (syntax-rules ()
    ((_ ((a ...) ...) (b ...))
     (list (list a b) ...))))
(display (bad-depth ((1 2) (3 4)) (10 20)))
```

Run with:
```
zig build run -- repro.scm
```

**Actual:** prints `((() 10) (() 20))` — `a` silently becomes `()` on
each iteration instead of a compile-time error.

**Expected:** a syntax error at macro-expansion time (R7RS: pattern
variables must be used at a consistent ellipsis depth), or at minimum
correct behavior rather than silently wrong output.

## Impact

[...]

Issue #550, opened 2026-06-30
vm_eval: top-level define-values ignores arity mismatches


baijum:
## Summary

Top-level `(define-values ...)` forms are intercepted by `handleTopLevelForm` in `src/vm_eval.zig` and executed by `handleDefineValues` instead of using the normal compiler desugaring. The handwritten binding logic accepts arity mismatches silently: it binds as many values as it can, ignores extras, leaves missing bindings uncreated, and continues execution.

R7RS `define-values` uses lambda-style formals for the produced values, so fixed-arity mismatches should be errors, as with `call-with-values` and a consumer lambda.

## Reproduction

Build Kaappi, then run these snippets with `./zig-out/bin/kaappi <file>`.

Too few values for fixed formals:

```scheme
(import (scheme base) (scheme write))
(define-values (a b) (values 1))
(display "REACHED-too-few")
(newline)
```

Actual output:

```text
REACHED-too-few
```

Too many values for fixed formals:

```scheme
(import (scheme base) (scheme write))
(define-values (a b) (values 1 2 3))
(display "REACHED-too-many:")
(write a)
(write b)
(newline)
```

Actual output:

```text
REACHED-too-many:12
```

Zero formals with a single value:

```scheme
(import (scheme base) (scheme write))
(define-values () 42)
(display "REACHED-zero-formals")
(newline)
```

Actual output:

```text
REACHED-zero-formals
```

Single value for two fixed formals:

```scheme
(import (scheme base) (scheme write))
(define-values (a b) 1)
(display "REACHED-single-two-formals:")
(write a)
(newline)
```

Actual output:

```text
REACHED-single-two-formals:1
```

## Expected behavior

Each malformed fixed-arity case should signal an error before reaching the marker expression. For example, `(define-values (a b) (values 1))` and `(define-values (a b) (values 1 2 3))` should not continue.

## Code references

- `src/vm_eval.zig:44` dispatches top-level `define-values` to `handleDefineValues`.
- `src/vm_eval.zig:66-114` handles multiple values but stops when either the formal list or value list ends, so too few values and extra values are both accepted.
- `src/vm_eval.zig:115-135` handles a non-`MultipleValues` result by binding only the first formal and accepting empty/multi-formal fixed lists.
- `src/compiler_lambda.zig:298-420` has the normal compiler desugaring through `call-with-values` and a consumer lambda, which is the path that should enforce lambda-style arity.

## Notes

[...]

PR #2078, opened 2026-08-01
Phase 3.8: ten smoke-only SRFIs — 264 assertions, and two closed issues whose regression tests pinned the valid case

baijum:
Audit v2 unit **3.8** — the SMOKE-ONLY batch: ten SRFIs whose test files
existed but held ≤12 assertions each. **67 → 261 assertions**, 11 disabled
behind `;; FAIL:` markers, every disabled one mutation-tested.

Four of the ten are genuinely thin *by nature* and their smoke tests were
close to adequate for what the library itself contains — SRFI 23 is one
procedure whose spec declines to prescribe behaviour at all, 46 and 149 are
bare `(export syntax-rules)` conformance statements, 244 re-exports a core
special form. So the work went into the mechanism each one *claims* rather
than the wrapper, which is where all three findings came from.

## Findings

| | |
|---|---|
| [#682](https://github.com/kaappi/kaappi/issues/682) **reopened** | A pattern variable used at a **lower** ellipsis depth than it was matched silently expands to `()`. This issue's own repro runs character-for-character as filed. |
| [#550](https://github.com/kaappi/kaappi/issues/550) **reopened** | Top-level `define-values` ignores an arity mismatch. Three of this issue's four repros still exit 0. |
| [#2075](https://github.com/kaappi/kaappi/issues/2075) **new** | A `begin`-wrapped internal `define` in a `let` body escapes to the **global** environment when no enclosing procedure exists. |
| [#2082](https://github.com/kaappi/kaappi/issues/2082) **new** | `syntax-rules` accepts **two ellipses in one list pattern** — outside R7RS's grammar — and splits the input arbitrarily. |
| [#2060](https://github.com/kaappi/kaappi/issues/2060) cited | SRFI 190's `coroutine-generator` is `call/cc`-based, so it inherits the SRFI-1-native-driver limitation exactly. Not re-filed. |

**#682** is the striking one, because the whole chain is visible in the
history. `8c3dda42` (PR #730) closed it with one line —
`if (b.depth != 1) return ExpandError.EllipsisDepthMismatch;`. `7c600a87`
(#931) removed that line, correctly: it blocked the legitimate nested
reduction of a depth-2 binding at an outer ellipsis. Nothing replaced it with
the depth *comparison* that would have been right, so today there is no
under-use validation at all. The removal went unnoticed because the
regression test #682 shipped,
`tests/scheme/smoke/ellipsis-depth-mismatch.scm`, only ever exercised the
**valid** case — it has printed `PASS` throughout. The depth-0 half that
`8c3dda42` never touched still carries the comment admitting it
(`expander_instantiate.zig:97`).

**#550** has the same shape one level down: `153cefd8` added its checks
inside the `types.isMultipleValues(result)` arm only. The single-value arm
has none, and `(values 1)` is indistinguishable from `1`, so every mismatch
producing exactly one value is still accepted silently.

**#2082** is the pattern-side twin of #682. R7RS §4.3.2's grammar admits at
most one ⟨ellipsis⟩ per list or vector pattern, and SRFI 46's tail patterns
widen what may *follow* it rather than how many there may be. Kaappi accepts
`(_ a ... b ...)` and gives `b` the last **two** elements whatever the input
length — the surplus ellipsis tokens are counted as fixed tail elements — so
`(two-ell 1 2 3)` ⟹ `((1) (2 3))` and `(two-ell 1 2 3 4 5)` ⟹ `((1 2 3) (4 5))`,
while `(two-ell 1)` fails outright. chibi and Guile both reject the
`define-syntax`. Filed separately from #682 because the check belongs where
the rule is parsed, before any binding exists to compare depths against.

**#2075** came out of checking the claim `lib/srfi/188.sld`'s header rests
on. R7RS §4.2.3 says a definition-context `begin` is transparent; here it is
transparent inside a procedure and not at top level:

```scheme
(define g 'global)
(let ((g 'outer)) (begin (define g 'inner)) g)  ;; ⟹ outer   (chibi: inner)
g                                               ;; ⟹ inner   (chibi: global)
```

The unwrapped `define` is correct in every position, and the identical `let`
inside any `lambda` is correct — so the defect is exactly "`begin`-wrapped
**and** no enclosing procedure scope". `let`, `let*` and `letrec` all leak.
The .sld header states the top-level answer as if it were universal.

A side-effect worth knowing for the fix's tests: **SRFI-64's `test-equal`
evaluates its expression inside a lambda**, which supplies the very scope
whose absence is the bug — so the first cut of the pinning assertions all
reported `FAIL` because inside the assertion the answer is *right*. The
committed version evaluates each probe at true top level into a `define`.

## Per-SRFI

| SRFI | exports | before | after | outcome |
|---|---:|---:|---:|---|
| 23 error | 1 | 7 | 24 | clean. Irritant identity, every basic type, unwinding through `dynamic-wind` and nested `guard`, predicate disjointness |
| 46 basic syntax-rules extensions | 1 (re-export) | 6 | 27 | **#2082**. Custom ellipsis makes `...` ordinary data in pattern *and* template; `(::: :::)` escape; nested default-ellipsis; tail patterns with two fixed tails, compound tails, dotted patterns, nesting, vectors |
| 98 environment variables | 2 | 7 | 26 | clean. The alist is freshly allocated per call, so mutating a returned pair cannot corrupt the environment or the next lookup |
| 112 environment inquiry | 6 | 8 | 36 | clean. All six string-or-`#f`, correct arity, stable; `machine-name`/`os-version` `#f` pinned as deliberate scope |
| 139 syntax parameters | 2 | 5 | 19 | clean. The adjustment reaches a macro defined *before* the `syntax-parameterize` and is restored across a macro boundary — the property a lexical `let-syntax` binding could not provide |
| 149 syntax-rules template extensions | 1 (re-export) | 6 | 22 | **#682**. Rule 4 over a *compound* template element (no spec example) matches Guile 3.0; chibi 0.12 diverges |
| 188 splicing let-syntax | 2 | 7 | 18 | **#2075**. Transformer-environment half is exactly right in both directions |
| 190 coroutine generators | 2 | 8 | 33 | **#2060** (cited). `yield` binding correct through another macro's template *and* pattern variable, and an inner rebinding still shadows |
| 236 independently | 1 | 7 | 26 | clean. Every expression exactly once at 20-wide; no template capture of `tmp`; zero/one/many value counts |
| 244 define-values | 1 (re-export) | 6 | 33 | **#550**. Every formals shape × every definition context; letrec* region |
| **total** | | **67** | **264** | |

Order is asserted nowhere in SRFI 236 — this implementation evaluates
right-to-left (so does chibi), so an order-assuming test there would be a
portability trap, not a conformance test. Nothing in 98 or 112 asserts a
platform value.

## Verified

Fresh ReleaseSafe build at `1718a2e1`, isolated `KAAPPI_HOME` per shell, cold
cache. #2075 re-checked under `-Dgc-stress=true` (built to a separate prefix):
identical. chibi-scheme 0.12 and Guile 3.0 used as oracles throughout; where
chibi is the one that diverges (SRFI 149 rule 4) that is said so and **not**
filed against Kaappi.

`kaappi test tests/scheme/srfi`: 30654 → 30851 passed, 0 failed. The one
`ERROR tests/scheme/srfi/srfi150.scm` is the pre-existing F13/#1903
divergence and is present on a clean tree too.

Tracking: #1890

### DIFF
--- a/tests/scheme/srfi/srfi112.scm
+++ b/tests/scheme/srfi/srfi112.scm
@@ -6,41 +6,76 @@
 ;; with native primitives (plus the literal "kaappi" name for
 ;; implementation-name) and always answers #f for machine-name and
 ;; os-version -- a deliberate reduced scope (see the .sld header), not a bug.
+;;
 ;; Exact OS/CPU string values are platform-dependent and deliberately not
-;; asserted here (the SRFI itself disclaims any standard string vocabulary);
-;; only the string-vs-#f shape is checked.
+;; asserted here: the SRFI itself says "no attempt is made to standardize
+;; the string values they return", and this file runs on macOS, five Linux
+;; architectures, three BSDs and two Windows targets. Only the string-vs-#f
+;; shape, the "or #f" contract, arity, and call-to-call stability are
+;; checked. Grown in audit v2 Phase 3.8 from 8 assertions.
 
 (import (scheme base) (scheme process-context) (srfi 64) (srfi 112))
 
 (test-begin "srfi-112")
 
-;; --- implementation-name --------------------------------------------------
+;;; --- every procedure returns a string or #f, and nothing else.  This is
+;;; the one blanket contract the SRFI does state: each returns a string,
+;;; or #f "if the implementation cannot provide an appropriate and relevant
+;;; result". ---
+
+(define srfi112-all
+  (list (cons "implementation-name" implementation-name)
+        (cons "implementation-version" implementation-version)
+        (cons "cpu-architecture" cpu-architecture)
+        (cons "machine-name" machine-name)
+        (cons "os-name" os-name)
+        (cons "os-version" os-version)))
+
+(for-each
+  (lambda (entry)
+    (test-assert (string-append (car entry) " returns a string or #f")
+      (let ((v ((cdr entry)))) (or (string? v) (eq? v #f))))
+    (test-assert (string-append (car entry) " is a procedure")
+      (procedure? (cdr entry)))
+    (test-error (string-append (car entry) " takes no arguments")
+      ((cdr entry) 'surplus))
+    (test-assert (string-append (car entry) " is stable across calls")
+      (equal? ((cdr entry)) ((cdr entry)))))
+  srfi112-all)
+
+;;; --- the four that must answer with a real string on every supported
+;;; platform, since each is backed by a compile-time constant ---
 
 (test-assert "implementation-name returns a string" (string? (implementation-name)))
 (test-equal "implementation-name is \"kaappi\"" "kaappi" (implementation-name))
 
-;; --- implementation-version -----------------------------------------------
-
 (test-assert "implementation-version returns a string"
   (string? (implementation-version)))
-
-;; --- cpu-architecture -------------------------------------------------------
+(test-assert "implementation-version is non-empty"
+  (> (string-length (implementation-version)) 0))
 
 (test-assert "cpu-architecture returns a string" (string? (cpu-architecture)))
-
-;; --- os-name ----------------------------------------------------------------
+(test-assert "cpu-architecture is non-empty"
+  (> (string-length (cpu-architecture)) 0))
 
 (test-assert "os-name returns a string" (string? (os-name)))
+(test-assert "os-name is non-empty" (> (string-length (os-name)) 0))
 
-;; --- machine-name: deliberately unsupported, always #f ----------------------
+;;; --- machine-name and os-version: deliberately unsupported, always #f.
+;;; The SRFI sanctions #f explicitly; this pins the reduced scope so that
+;;; implementing either later is a deliberate change, not an accident. ---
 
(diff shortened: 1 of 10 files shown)


baijum:
Merged count correction: the squash commit's subject says **261 assertions**, which was the total before the last commit added the #2082 pinning set to `srfi46.scm`. The merged content is **264** (`srfi46.scm` is 27, not 24) — `kaappi test tests/scheme/srfi` goes 30654 → 30851.

Also, for anyone re-running the suites against this branch: the two `tests/scheme/compile/` failures I saw locally were **not** a regression. `compile-import-error-703.sh` and `compile-preamble-gc-700.sh` compile the fixture `.sbc` with whatever `zig-out/bin/kaappi` happens to be, then rebuild the interpreter from current sources — so a `zig-out` that predates your own commits embeds an `.sbc` whose build id no longer matches, and the bundled binary dies with `invalid embedded bytecode`, silently (the test discards its stderr). `zig build` after committing, and both pass.

Issue #2000, opened 2026-08-01
[SRFI-181] Channel/fiber blocking inside a custom-port callback bypasses in_custom_port_callback and aborts the process (SIGBUS)


baijum:
**Category:** SRFI 181 / KEP-0001 fibers · **Severity:** crash (uncatchable process abort) · **Found by:** Systematic audit v2, Phase 2.6 (#1890)

`vm.in_custom_port_callback` exists precisely to stop a custom-port callback from blocking, because a callback always runs with `dispatched_from_scheduler` forced false and so cannot park — it can only *recursively drive the scheduler on the native stack*. Its own doc comment (`src/vm.zig:518-531`) names the hazard: "a confirmed native-stack-overflow risk under concurrent fibers".

The counter is checked at **exactly two sites** — `fiber.waitForFd` (`src/fiber.zig:1225`) and `primitives_srfi18.threadSleepFn` (`src/primitives_srfi18.zig:644`). The three blocking primitives in `(kaappi fibers)` are not among them, and all three take the recursive-drive path.

## The drive is real, not theoretical

```scheme
(import (scheme base) (srfi 181) (kaappi fibers))
(define ch (make-channel))
(define log '())
(define port (make-custom-textual-input-port
              "drv"
              (lambda (str start count)
                (set! log (cons 'callback-entered log))
                (let ((v (channel-receive ch)))          ; empty at entry
                  (set! log (cons (list 'callback-got v) log))
                  (string-set! str start #\X) 1))
              #f #f #f))
(spawn (lambda () (set! log (cons 'sibling-ran log)) (channel-send ch 'from-sibling)))
(read-char port)   ;; ⟹ #\X
(reverse log)      ;; ⟹ (callback-entered sibling-ran (callback-got from-sibling))
```

A whole sibling fiber ran to completion **inside** the callback. `channel-send` (on a full or rendezvous channel) and `fiber-join` behave identically — verified separately.

## And it aborts the process

Nesting grows one native drive per level. With `n` fibers each reading the same custom port, whose `read!` blocks on a channel:

| n | result |
|---:|---|
| 500 / 1000 / 2000 / 2400 | ok — `max-native-nesting` = n+1 |
| **2500 / 3000 / 6000** | **`Bus error at address 0x…`, exit 134** |

[...]

Issue #1939, opened 2026-07-31
A custom-port read! callback that re-enters a read on its own port aborts the process (exit 134)


baijum:
**Category:** SRFI 181 custom ports · **Severity:** crash · **Found by:** systematic audit v2, Phase 2.2 (#1890)

A `read!` callback that reads from its own port trips `std.debug.assert(port.read_buf == null)`. `guard` cannot catch it; the process aborts.

```text
kaappi internal error — this is a bug in kaappi, not in your program.
thread 29215802 panic: reached unreachable code
src/primitives_io.zig:1098 in readOneByte
exit 134
```

## Reproduction

```scheme
(import (scheme base) (srfi 181))
(define self #f) (define depth 0) (define n 0)
(define p (make-custom-binary-input-port "reent"
  (lambda (bv start count)
    (set! depth (+ depth 1))
    (if (= depth 1) (read-u8 self))
    (set! depth (- depth 1))
    (if (< n 8) (begin (bytevector-u8-set! bv start 65)
                       (bytevector-u8-set! bv (+ start 1) 66)
                       (set! n (+ n 1)) 2) 0))
  #f #f #f))
(set! self p)
(guard (e (#t 'caught)) (read-u8 p))
```

Assert sites: `primitives_io.zig:1210` (`readOneByteFromCustomPort`) and `:1368` (`readOneByteFromTranscodedPort`). The transcoded variant is reached when the inner re-entry consumes only part of a multi-byte character.

## Discriminating control

The **identical** re-entrant shape with every `read!` returning exactly **1** byte completes cleanly, exit 0. So the trigger is specifically *"the inner call leaves `read_buf` non-empty **and** the outer call returns ≥ 2 bytes"* — not re-entrancy as such.

## Non-aborting sibling, same root cause

When the outer `read!` returns exactly one byte the assert never fires, but the outer's byte is served **before** bytes the inner call already buffered — bytes that are chronologically earlier. The stream is silently reordered: emits `Z` then `B`.

So the single-slot `read_buf` is both a crash and a correctness bug depending on the return size.

## Why the existing guard does not cover it

`vm.in_custom_port_callback` guards **blocking**, not **recursion**. That mechanism works correctly (verified separately: a callback attempting `thread-sleep!` or another port's I/O is rejected with a catchable error and a good message). Nothing prevents a callback from re-entering a *non-blocking* read on its own port.

## Suggested direction

[...]

PR #2090, opened 2026-08-01
Require a priority label on every issue

baijum:
## Summary

Adds an **Issue tracker** section to `CLAUDE.md` and a row to its enforcement map. Follow-up to #2032 and #2041, which documented the priority rubric in `docs/dev/github-issues.md` but left nothing telling a filer to apply it.

The evidence that a document alone is not enough: **41 issues were filed without a priority label while those two PRs were open**, arriving in bursts of 6–27 as each audit phase completed. Triaging after the fact means re-reading every issue body to recover a judgement the filer had already made and discarded.

All of them are now labeled. Every open issue carries exactly one priority label, modulo the `fuzz-finding` exemption.

## What the rule says

The decision table, at filing time rather than a link away:

| Level | The question it answers |
|-------|------------------------|
| critical | Does this compromise the process — memory unsafety, or an abort reachable from an ordinary program? |
| high | Does a legal program get a silently wrong answer, hang, or lose data in a path users actually reach? |
| medium | Is behaviour wrong against a spec or a documented guarantee, but loud, narrow, or recoverable? |
| low | Is the behaviour right and only its *description* or *diagnostic* wrong? |

Plus the four rules that settle the hard cases — `critical` is process-level unsafety only; reachability separates critical from high; an audit header's `Severity:` is an input, not the answer; and silence is an aggravator that moves an issue up *within* a level rather than between levels.

The full rubric, the four label axes, and the worked boundary cases stay in `docs/dev/github-issues.md`. This is the filing-time subset, and it points there.

## Advisory, not gated

Listed as **Advisory only** in the enforcement map, alongside the existing test and file-size rules. A CI gate would have to run against the tracker rather than the diff, and the failure mode is a missing label rather than broken code — not worth a workflow that can only ever fail after the fact.

`AGENTS.md` is a symlink to `CLAUDE.md`, so the rule reaches both without a second copy.

## Checklist

- [x] `zig build test` passes — n/a, no `src/` changes
- [x] `zig fmt --check src/` passes — n/a, no `.zig` files touched
- [x] Scheme test suites pass — n/a, no engine or library changes
- [ ] New tests added for new behavior — n/a, documentation only
- [x] Documentation updated (if applicable) — this PR is the documentation

`npx markdownlint-cli2` is clean, and the `gh` snippet embedded in the section was run verbatim before committing. The CHANGELOG gate does not apply (fires only on `src/` or `lib/srfi/`).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1258,6 +1258,44 @@ every release, across `ubuntu-latest`, `ubuntu-24.04-arm`, and `macos-latest`.
 Adding a platform means teaching its `detect_platform` the `uname` spelling
 and its `rt_artifact` case — `docs/dev/porting.md` Stage 6.
 
+## Issue tracker
+
+**Every issue you file or triage gets exactly one `priority:` label** —
+`critical`, `high`, `medium`, or `low`. Set it when filing; an issue that
+arrives without one is not triaged. `docs/dev/github-issues.md` is the full
+rubric (the four label axes, worked boundary cases, the triage commands);
+this is the part you need at filing time:
+
+| Level | The question it answers |
+|-------|------------------------|
+| critical | Does this compromise the process — memory unsafety, or an abort reachable from an ordinary program? |
+| high | Does a legal program get a silently wrong answer, hang, or lose data in a path users actually reach? |
+| medium | Is behaviour wrong against a spec or a documented guarantee, but loud, narrow, or recoverable? |
+| low | Is the behaviour right and only its *description* or *diagnostic* wrong? |
+
+Four rules decide the hard cases:
+
+- **`critical` is process-level unsafety only.** A correctness bug tops out
+  at `high` however broad or silent. All 13 issues ever marked critical are
+  memory unsafety or a process abort.
+- **Reachability separates critical from high.** An abort needing a stress
+  harness is `high`; one reachable from a five-line program is `critical`.
+- **An audit header's `Severity:` is an input, not the answer.**
+  `wrong-result` spans `high` to `medium` purely on blast radius.
+- **Silence is an aggravator, not a level.** A loud failure is safer than a
+  quiet one, so silence moves an issue up *within* its level.
+
+Auto-filed `fuzz-finding` CI reports are the one exemption — they are
+triage-and-close, and carry no priority. To find what still needs a label:
+
+```bash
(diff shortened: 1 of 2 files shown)


baijum:
Thanks — one accepted, one declined.

## Accepted: the audit command was half-blind (`6541be2d`)

Correct and worth catching. The prose requires *exactly one* `priority:` label and the filter only reported *zero*, so an issue carrying two was invisible to the check meant to enforce the rule. Verified the difference against a synthetic fixture before taking it:

```
old filter → 2 (none)
new filter → 1 (two labels), 2 (none)
```

Adopted the counting form, and fixed it in `docs/dev/github-issues.md` too — the snippet was copied from there, so the same blind spot shipped in #2032. Both copies were then run verbatim against the live tracker: 0 violations either way, which is expected, since no issue has ever carried two. That is the point of fixing it — a check that cannot observe half of what it asserts reports a clean tracker whether or not the tracker is clean.

## Declined: moving the guidance out of CLAUDE.md

`CLAUDE.md` is loaded into context automatically at the start of every session; `docs/dev/*.md` is not. A rule that must be applied *at filing time* has to be in the file that is already open, or in practice it will not be consulted — which is the failure this PR exists to fix. 41 issues were filed unlabeled while `docs/dev/github-issues.md` already existed and already said all of this.

This also follows the established house pattern rather than departing from it. `CLAUDE.md` carries the actionable summary and the detailed document carries the full treatment, in several places already:

| CLAUDE.md section | Full treatment |
|---|---|
| "GC safety" — 3 key rules | `.claude/rules/gc-safety.md` |
| "How to add a new built-in procedure" — the ~75-line checklist | `docs/dev/adding-features.md` |
| "How to add a new compiler form" | `.claude/rules/compiler-forms.md` |

The new section is 39 lines and points at `docs/dev/github-issues.md` in its second sentence, which is the same shape as those three. The label taxonomy, the worked boundary cases (#1939 vs #2000), the severity-vs-priority discussion and the corpus evidence all stayed in `docs/dev/` — only the decision table and the four tie-breaker rules were duplicated, and those are exactly what a filer needs in hand.

On the enforcement map: the row already cites both locations — `This file (Issue tracker), docs/dev/github-issues.md` — matching how the neighbouring "Bug fixes need tests" row cites its own section. No change needed there.

Issue #2096, opened 2026-08-01
[Tooling] kaappi --check runs the program: a fmt-scoped flag is accepted globally, one hyphen from the check subcommand that executes nothing


baijum:
**Category:** Tooling
**Severity:** wrong-result (silent code execution when the user asked for none)
**Found by:** Systematic audit v2, Phase 6C (completions ↔ flag-table drift gate)

## Summary

`--check` and `--no-opt` are *subcommand-scoped* flags — `--check` belongs to
`kaappi fmt`, `--no-opt` to `kaappi ir` — but both are parsed by the **global**
flag loop, which accepts them in any position with no scope check at all. For
`--no-opt` that is merely inert. For `--check` it is a live hazard, because the
spelling is one hyphen-pair away from the `check` **subcommand**, whose entire
documented contract is that it *executes nothing*:

> `kaappi check <file>` runs compile-only static analysis (reads, expands,
> compiles, executes nothing)

A user (or a script) that types `--check` instead of `check` gets the program
**run**.

## Minimal reproduction

```console
$ cat bad.scm
(display "ran\n")
(car 1)

$ kaappi check bad.scm            # the subcommand: nothing executes
2:1: error[KP4003]: 'car' expects a pair as argument 1, but an integer literal was given
check: 1 error, 0 warnings
$ echo $?
1

$ kaappi --check bad.scm          # the flag: the program RUNS
ran
bad.scm:2:1: error[KP3002]: type error in 'car': expected pair, got 1
    (car 1)
$ echo $?
1
```

Both exit 1, so a CI step that only inspects the exit code cannot tell the two
apart — it silently executed the file it meant to only analyse.

## Discriminating control

`kaappi --deny-warnings bad.scm` also runs the file, and that is *fine*: it is
documented as `(check)`-scoped but has no same-named subcommand to be confused
with, so nothing about the spelling promises otherwise. The defect is specific
to `--check`, where the flag and a subcommand with the opposite execution
semantics differ only by the leading `--`.

The same shape, inert rather than harmful:

```console
$ kaappi --no-opt bad.scm         # accepted; means nothing outside `kaappi ir`
ran
...
$ kaappi fmt --check --no-ir-opt ugly.scm   # --no-ir-opt accepted under fmt too
```

## Root cause

`src/cli.zig`'s parse loop dispatches every flag through one table with no
notion of which subcommand is in play. `cli_spec.zig` now records the intent —
[...]

Issue #552, opened 2026-06-30
zsh completion does not offer the `compile` subcommand


baijum:
## Summary

The zsh completion script generated by `kaappi --completions zsh` only completes the first positional argument as a `*.scm` file. Unlike the bash and fish completion scripts, it does not offer the `compile` subcommand, so users do not get tab completion for the documented native compile command.

## Affected code

- `src/completions.zig`: the zsh script uses `_arguments ... '1:command or file:_files -g "*.scm"'`, which treats the first argument as files only.
- The bash script includes `compgen -W "compile"` and the fish script includes `complete -c kaappi -a compile`, so this is a zsh-only mismatch.

## Reproduction

1. Build/run kaappi from the repo.
2. Generate the zsh completion script:
   ```sh
   zig build run -- --completions zsh
   ```
3. Inspect the first positional completion rule:
   ```zsh
   '1:command or file:_files -g "*.scm"'
   ```
4. Load the generated script in zsh and type:
   ```zsh
   kaappi c<TAB>
   ```

## Expected behavior

The zsh completion should offer `compile`, matching the documented CLI and the bash/fish completion scripts.

## Actual behavior

The zsh completion only offers Scheme files for the first positional argument, so `compile` is not suggested.

PR #2099, opened 2026-08-01
Phase 6C: derive shell completions from the flag table — --no-ir-opt and 5 test flags were missing, and zsh/fish offered 15 rejected flags inside every out-of-line subcommand

baijum:
Phase 6C of the v2 audit campaign (#1890). A **fix** unit, not a survey.

`kaappi --completions <shell>` emitted three hand-written string literals, structurally parallel to the argument parsers with nothing keeping them in step — the same shape as `isRejectedFormHead` (unit 4A) and the `%`-prefix incident. This makes them **generated at comptime from the table the parsers themselves dispatch on**, so the class of bug cannot recur.

Closed once before as a one-off: #552, "zsh completion does not offer the `compile` subcommand".

## The drift, measured in both directions

Against the real binary at `1718a2e1`, macOS aarch64, ReleaseSafe, isolated `KAAPPI_HOME`.

### Accepted by the CLI, missing from the completions

| Flag | Context | bash | zsh | fish |
|---|---|:--:|:--:|:--:|
| `--no-ir-opt` | global | **missing** | **missing** | **missing** |
| `-j` / `--jobs <n>` | `test` | **missing** | **missing** | **missing** |
| `--changed` | `test` | **missing** | **missing** | **missing** |
| `--list-affected` | `test` | **missing** | **missing** | **missing** |
| `--since <rev>` | `test` | **missing** | **missing** | **missing** |
| `--seed <n>` | `test` | ok | **missing** | **missing** |
| `--json` | `explain`/`features`/`test`/`doctor` | ok | **missing** | **missing** |
| `--all` | `explain` | ok | **missing** | **missing** |
| `-h` / `--help` | all 8 subcommand contexts | **missing** | (global) | (global) |
| bare `--timings` | global | ok | **missing** (only `--timings=`) | ok |
| `status` / `clear` | `cache` | ok | **missing** | ok |

The tracker named only `--no-ir-opt`. The `kaappi test` selection flags are the larger gap: five flags, none of them offered anywhere.

### Offered by the completions, rejected by the CLI

The worse direction, and the one nobody had looked at. `explain`, `features`, `test`, `doctor` and `cache` intercept argv in their own module **before** `cli.parse` runs, and reject anything outside their own small table with exit 2. zsh (`_arguments -s $flags`) and fish (`complete` with no `-n` guard) offered the entire global flag set in every one of those contexts:

```console
$ kaappi features --gc-stats      # fish/zsh completed this
kaappi features: unexpected argument '--gc-stats'      # exit 2
$ kaappi doctor --profile
kaappi doctor: unknown option '--profile'              # exit 2
$ kaappi cache --version
kaappi cache: unknown subcommand '--version'           # exit 2
```

Measured: of 16 global flags probed per context, **15 are rejected with exit 2 and 1 accepted** (`--help`) — identically for all five subcommands. bash was the only script that scoped, and had zero offered-but-invalid.

`thottam`'s three scripts had **no drift in either direction** — asserted so now rather than left to luck.

## The authoritative table

New `src/cli_spec.zig`, read by five consumers:

| Consumer | For |
|---|---|
| `cli.zig` — `parse` / `preScanSandbox` | global flags |
| `cli.zig` — `printUsage` | the `Options:` block, **generated** |
| `explain.zig`, `features.zig`, `doctor.zig`, `test_runner.zig`, `cache.zig` | each subcommand's own flags |
| `thottam.zig` | thottam's flags |
| `completions.zig` | all six shell scripts, **generated at comptime** |

Every parse loop was rewritten to dispatch on `switch (m.flag.id)` over its table's `Id` enum. `printUsage`'s hand-maintained `Options:` block — a *fourth* parallel list — is now generated too.

The `--diagnostics=` / `--timings[=fmt]` escape hatch is closed. Both were matched by `std.mem.startsWith` **outside** the old flag table, because it had no way to express GNU `=` syntax — which is exactly how they, and later `--no-ir-opt`, drifted out of the scripts. A new `ValueSyntax` (`none` / `separate` / `inline_eq` / `inline_eq_optional`) covers every spelling, so nothing reaches a parse loop without passing through the table. That also fixes the zsh specs: an attached-only value is `--opt=-` and `::` marks it optional, so zsh now completes `--diagnostics=json` instead of the separated `--diagnostics json` that the parser rejects.

## The gate

1. **Comptime, per parser** — an exhaustive `switch` over each table's `Id`. A new row is a compile error until the parser handles it.
2. **Comptime, per table** — every `Id` variant must appear exactly once; no duplicate spellings; `value_name` required on value-taking flags; descriptions restricted to characters needing no quoting inside a zsh `'…[desc]'` spec or fish `-d '…'` string.
3. **Generation** — the scripts cannot lag, because they *are* the table.

Verified to fire, by mutation:

| Mutation | Caught by |
|---|---|
| add a `GlobalId` variant with no table row | `cli_spec.zig`: `global_flags: frobnicate must appear exactly once in the table` |
| add a table row the parser does not handle | `cli.zig:183: error: switch must handle all possibilities` |
| delete `--seed` from `test_flags` | `cli_spec.zig`: `test_flags: seed must appear exactly once in the table` |
| drop `--no-ir-opt` from the top-level offer | 2 unit tests (`completions`, `cli`) + 3 shell assertions |

## Tests

`completions.zig` had **zero**. Now 11 unit tests plus a 41-assertion shell suite.

**Unit** (`src/completions.zig`, `src/cli_spec.zig`, `src/cli.zig`): completeness in both directions (every table flag in every script; every `--word` token in every script names a declared flag), the `test` selection flags by name, per-shell invariants (balanced quotes per line, ASCII only, the zsh `=-` spelling, the fish bare-`--timings` spelling), `match` over all four value syntaxes, `--timeout` not prefix-matching `--timings`, and `printUsage` documenting exactly the top-level flags.

**Shell** (`tests/scheme/completions/completions.sh`, new, wired into `run-all.sh`): sources the generated bash function and **drives it**, then feeds every offered flag back to the **real binary** per context, requiring none comes back as `unknown option '<flag>'` / `unexpected argument '<flag>'` / `unknown subcommand '<flag>'`. It carries its own discriminating control — `kaappi features --sandbox` must be rejected *and the matcher must see it* — so a broken matcher cannot make the soundness checks pass vacuously. bash and zsh are syntax-checked (`-n`); fish is gated on being installed, since it is absent from most CI legs.

Portability: ERE (`grep -oE`), never GNU `\|`; here-strings, never a pipe into `grep -q`; probes run from inside the temp dir because `-o` / `--compile` / `--emit-llvm` write a file named after their argument (caught during development — the first draft left two files in the working tree).

## Found while doing it

#2096 — `kaappi --check foo.scm` **runs the program**. `--check` is `fmt`'s flag but the global loop accepts it anywhere, and it is one hyphen-pair from the `check` subcommand whose contract is that nothing executes. Both exit 1, so a CI step reading only the exit code cannot tell them apart. Filed, not fixed here: the fix is a behaviour change worth deciding on its own.

## Verification

- `zig build test` — green
- `bash tests/scheme/run-all.sh` — 2036 pass. The first run showed 2 failures, both `bundle_fixture_binary` tests, both `fatal: invalid embedded bytecode`: I had built the binary from a dirty tree and then committed, so the `-Dbundle` rebuild's build id (`54b87942`) no longer matched the `.sbc` the stale binary (`1718a2e1-dirty`) produced — the documented `docs/dev/cache.md` footgun, not a regression. Both pass against a correctly-built binary; a full clean re-run is in flight and CI builds from the committed tree, where this cannot occur.
- `tests/scheme/completions/completions.sh` — 41/41, under both bash 5.3 and macOS's system bash 3.2
- `tests/scheme/sandbox/` — 4/4 (`preScanSandbox` was rewritten; 9 hand probes confirm `--sandbox` detection is unchanged after `--diagnostics=json`, `--timings`, `--timings=json`, `--no-ir-opt`, `--lib-path <dir>` and a bare subcommand)
- cross-compiles clean for `aarch64-windows`, `x86_64-linux`, `s390x-linux`, `wasm32-wasi` — the tables are pure comptime and target-independent
- `doctor` 20/20, `fmt` 12/12, `cache` 11/11, `timings` 42/42, `pipeline` 13/13, `errors/exit-code` 49/49, all 5 `test-runner` suites — green
- `markdownlint-cli2` — 0 issues
- all three scripts eyeballed by hand; `bash -n` and `zsh -n` clean

Docs: new `docs/dev/cli-surface.md` (indexed in `docs/dev/README.md`), plus `CLAUDE.md`, `tests/scheme/CLAUDE.md` and `docs/dev/explain.md`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -403,6 +403,8 @@ its struct lives outside `types.zig` entirely with no `types.Fiber` re-export.
 | `disassembler.zig` | Bytecode disassembler for `(disassemble proc)` |
 | `linenoise.zig` | Zig FFI wrapper for vendored linenoise C library |
 | `main.zig` | Entry point, REPL loop with linenoise, file execution, CLI flags, `pub const version`, `pub const panic` |
+| `cli_spec.zig` | **The** CLI flag/subcommand tables. Every parse loop (`cli.zig`, `explain`, `features`, `doctor`, `test_runner`, `cache`, `thottam`) dispatches on an exhaustive `switch` over one of its `Id` enums, `cli.printUsage` generates its `Options:` block from it, and `completions.zig` generates all six shell scripts from it — so a flag cannot reach a parser without the docs and completions following. See `docs/dev/cli-surface.md` |
+| `completions.zig` | bash/zsh/fish completion scripts for `kaappi` and `thottam`, generated at comptime from `cli_spec.zig`. Nothing here is hand-maintained |
 | `crash.zig` | Custom panic handler (`PanicHandler(name)`) + pipeline breadcrumb (`noteStage`/`noteFile`); prints version/target/build-mode + stage + report URL before the trace. See `docs/dev/crash-reporting.md` |
 | `native_compiler.zig` | LLVM IR emission, native binary compilation, C compiler discovery, linker invocation |
 | `thottam.zig` | Package manager binary (thottam): install, remove, list, update, verify |
--- a/docs/dev/README.md
+++ b/docs/dev/README.md
@@ -43,6 +43,7 @@ investigation produced analysis worth keeping.
 | [features.md](features.md) | `kaappi features [--json]`: machine-readable capability discovery — version/build id, subsystems (the shared `cond-expand` table), SRFIs, limits |
 | [doctor.md](doctor.md) | `kaappi doctor`: installation/environment self-check — the checks, the exit-code contract, the native-backend smoke link |
 | [cache.md](cache.md) | The `.sbc` bytecode cache: what the key contains (source hash + build id), where entries live, what invalidates them, how to inspect/clear/bypass |
+| [cli-surface.md](cli-surface.md) | `src/cli_spec.zig`, the one flag table every parser, `--help`, and all six shell completion scripts are derived from — and the comptime gate that keeps them from drifting |
 | [claude-code-harness.md](claude-code-harness.md) | Hooks, permissions, path-scoped rules, and skills for AI-assisted development |
 
 ## Design decisions
--- a/docs/dev/cli-surface.md
+++ b/docs/dev/cli-surface.md
@@ -1,0 +1,152 @@
+# The CLI surface: one table, five consumers
+
+`src/cli_spec.zig` is the single authoritative description of the `kaappi` and
+`thottam` command lines. Everything that needs to know how a flag is spelled,
+whether it takes a value, or what it does reads that file:
+
+| Consumer | What it reads it for |
+|---|---|
+| `src/cli.zig` — `parse` / `preScanSandbox` | recognizing global flags |
+| `src/cli.zig` — `printUsage` | the `Options:` block, generated |
+| `src/explain.zig`, `features.zig`, `doctor.zig`, `test_runner.zig`, `cache.zig` | each subcommand's own flags |
+| `src/thottam.zig` | thottam's flags |
+| `src/completions.zig` | all six shell scripts, generated at comptime |
+
+**Add a flag in one place and every one of those follows.** There is no second
+list to update and no golden file to regenerate.
+
+## Why it exists
+
+The completion scripts used to be hand-written string literals parallel to the
+parsers — the same structural hazard as `isRejectedFormHead`
+(`docs/dev/llvm-backend.md`) and the `%`-prefix incident: two lists that must
+agree, with nothing enforcing it. They had drifted in **both** directions:
+
+- `--no-ir-opt` was in the parser and in none of the three kaappi scripts.
+- `kaappi test` grew `-j`/`--jobs`, `--changed`, `--list-affected` and
+  `--since`; none of the three offered any of them, and only bash offered
+  `--seed`.
+- The zsh and fish scripts offered the whole global flag set inside
+  `explain`, `features`, `test`, `doctor` and `cache` — 15 of which those
+  subcommands' own parsers reject outright with exit 2. **Completing a flag
+  that then errors is the worse failure**, and it was the larger of the two.
+
+## The shape of the table
+
+```zig
+pub const GlobalId = enum { help, version_flag, /* … */ };
+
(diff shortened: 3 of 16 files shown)


baijum:
**The `benchmark-pr` failure here is not attributable to this diff.** Evidence, in the order that settles it:

**1. `call_cc` and `call_ec` come from the same process.** `benchmarks/run-benchmarks.sh:71-87` runs `zig build bench` **once** and greps both metrics out of that single output:

```bash
output=$($TIMEOUT_CMD zig build bench 2>&1 || true)
cc_line=$(echo "$output" | grep "^name: call_cc," || true)
ec_line=$(echo "$output" | grep "^name: call_ec," || true)
```

In this run `call_cc` is **1.00x** (1.664188 vs 1.665071) while `call_ec` is 1.22x. Any cost this PR could plausibly add — CLI parsing, startup, a larger binary, allocation — is paid by that one process and would move **both** numbers. It moved one.

**2. Nothing else moved.** All 14 other benchmarks are 0.98x–1.02x, including `string` at 0.0047s, which is **ten times smaller** than `call_ec` and would therefore show a fixed startup delta ten times more strongly. It shows nothing.

**3. `call_ec` is the only single-shot entry in the table.** Both `call_cc` and `call_ec` are emitted as `... 0 0 1 0 0 0` — min 0, max 0, **iterations 1**. Every other benchmark reports a median over repetitions with min/max. So the one metric that tripped the 1.20x gate is structurally the noisiest one measured.

I checked the alternative before concluding, rather than assuming noise: across four unrelated PRs (#2076, #2081, #2092, #2098) `call_ec` sits in 0.0435–0.0455, and this run's 0.0545 is outside that band. That is what made it worth investigating rather than waving off — but points 1 and 2 rule out a mechanism, and point 3 explains why this entry in particular produces the outlier.

Filed as a separate issue: a gate that can fail a PR on an unrepeated 45 ms measurement will keep doing this.

PR #2104, opened 2026-08-01
Re-derive the bytecode and IR docs from source, and gate the counts at comptime

### LINKED ISSUE #2102: docs/dev/bytecode.md and docs/dev/ir.md describe an earlier design: wrong opcode count, wrong operand widths, and two analysis passes that no longer exist
## Summary

`docs/dev/bytecode.md` and `docs/dev/ir.md` describe two subsystems inaccurately enough
that anyone working from them will write wrong code. Both document *counts, widths and
struct fields* — the kind of detail a reader takes on faith rather than re-deriving — and
in each case the source disagrees.

Found while comparing kaappi's IR and bytecode against a second implementation's, to
judge whether the two could share a representation. Every claim below was checked against
`src/` on `v0.22.1`.

## `docs/dev/bytecode.md`

**Opcode count is 29; the enum has 31.**

```
$ awk '/pub const OpCode = enum\(u8\)/,/^};/' src/types.zig | grep -cE '^\s+(@"[a-z_]+"|[a-z_]+),'
31
```

A naive count gives 30 and misses `@"return"` (types.zig:16 of the enum), which is
escaped because `return` is a Zig keyword. The two absent from the doc's table are
`tail_call_cc` and `tail_eval`.

**Register operands are documented as u8; the dispatch loop reads u16.**

> `docs/dev/bytecode.md:8` — "Operands are u8 (register/slot) or u16 (constant index, big-endian)."

But `vm_dispatch.zig` reads `readU16` for register operands too:

```zig
.load_const => {
    const dst = readU16(self, frame);   // dst is a register
    const idx = readU16(self, frame);
.move => {
    const dst = readU16(self, frame);
    const src = readU16(self, frame);
```

and the helper is `registerIndex(vm, base, reg: u16)`. Only `nargs` is u8. This makes
**every per-opcode byte count in the document's table off by one or more**, which is the
part that actually matters — a reader computing instruction widths from the doc gets a
stream that `validateCode` rejects. The authoritative table is the `fixed_operand_bytes`
switch in `vm_dispatch.zig`.

**Closure capture descriptors are 3 bytes, not 2** — `is_local: u8` + `index: u16`.
Confirmed on the reader side:

```zig
// bytecode_file_read.zig:305
const capture_bytes = @as(usize, inner.upvalue_count) * 3;
```

## `docs/dev/ir.md`

**Node-tag count is 33; `NodeTag` has 18.**

```
$ awk '/pub const NodeTag = enum/,/^};/' src/ir.zig | grep -cE '^\s+(@"[a-z_]+"|[a-z_]+),'
18
```

The document appears to predate the collapse of `named_let`, `do_form`, `cond`,
`case_form`, `guard`, `quasiquote` and the rest into the single `sexpr_form` tag with its
18 `FormKind`s. Describing them as top-level node tags gives entirely the wrong picture of
how lowered the IR is — which is the one thing a reader most needs from this document.

**Three documented `Annotations` fields do not exist.** The doc's table lists:

| field | attributed to |
|---|---|
| `is_primitive_call` | `identifyPrimitives` |
| `primitive_name` | `identifyPrimitives` |
| `is_constant` | `markConstants` |

The actual struct is:

```zig
pub const Annotations = struct {
    is_tail: bool = false,
    span: types.Span = .{},
};
```

**And the two analysis passes those fields came from are gone.** `identifyPrimitives` and
`markConstants` have zero occurrences anywhere in `src/`. The pipeline diagram at
`docs/dev/ir.md:24-25` still shows both, and §"identifyPrimitives(node)" / §"markConstants(node)"
document them in full. `markTailPositions` is the only analysis pass left.

There is knock-on text: `docs/dev/ir.md:259` says `node.ann.is_primitive_call` is "passed
through to call identification", and `:281` describes tests that "verify that each
analysis pass annotates nodes".

## Same claims elsewhere

Three more files repeat it, so a fix should sweep them rather than only the two above:

```
README.md:341                | **IR** | … (33 node types) with analysis passes
                               (tail positions, primitives, constants) …
CLAUDE.md:225                (33 node  (tail pos,
CLAUDE.md:236                … (33 node types). Runs 3 analysis passes (tail positions,
                               primitive identification, constant detection) …
docs/dev/architecture.md:31  … (33 node types). Runs 3 analysis passes (tail positions,
                               primitive identification, constant detection) …
```

"3 analysis passes" is the same error stated a third way: there is one,
`markTailPositions`.

## Why it is worth fixing rather than living with

These are not prose inaccuracies. The opcode count, operand widths and descriptor size are
exactly what an implementer copies into an encoder, and the IR's node-tag count and
annotation fields are what a reader uses to judge how much lowering has already happened.
Both documents currently describe an earlier design confidently enough that a careful
reader would not think to check.

Suggested guard: the `OpCode` and `NodeTag` counts could be asserted at comptime against
the numbers in the docs, the way `diagnostics.zig` already enforces its own table's
completeness — that is the mechanism that would have caught all of this.

baijum:
Fixes #2102.

Every claim in the issue was verified against `src/` before changing anything, and all of them held. The fix re-derives both documents from the source and adds the comptime guard the issue suggested.

## What was wrong

**`docs/dev/bytecode.md`** — 29 opcodes documented, 31 in the enum. Register operands documented as u8, but `vm_dispatch.zig` reads `readU16` for every one of them, so **every byte count in the table was wrong** — the part that actually matters, since a reader computing instruction widths gets a stream `validateCode` rejects. Closure capture descriptors are 3 bytes (`is_local:u8` + `index:u16`), not 2.

The table is now re-derived from the `OpCode` enum and the `fixed_operand_bytes` switch, with `tail_call_cc` and `tail_eval` added. I verified the finished table programmatically against both — 31 rows, correct numbering, `Bytes == 1 + fixed_operand_bytes` throughout.

The disassembly example was hand-computed with the old widths, so I replaced it with real `(disassemble fib)` output. It also showed a `; Bytecode (N bytes):` header the disassembler never emits, and described offsets as hex when the format is `{d:0>4}` — decimal.

**`docs/dev/ir.md`** — 33 node tags documented, `NodeTag` has 18. The doc predates the collapse of `cond`/`case`/`guard`/… into the single `sexpr_form` tag with its 18 `FormKind`s, which gives entirely the wrong picture of how lowered the IR is. Three documented `Annotations` fields don't exist, and the two passes that set them (`identifyPrimitives`, `markConstants`) were removed in v0.13.0 — `markTailPositions` is the only pass left. The standalone `Emitter` and its bytecode-parity tests went with them.

## Three findings the issue didn't list

- **`architecture.md` carried a second, undeclared opcode table** describing three opcodes that do not exist (`get_local`, `set_local`, `close_upvalue`) while omitting three that do (`self_tail_call`, `tail_call_cc`, `tail_eval`) — a wrong table whose row count happened to stay right at **31**, so a count check would have passed it. Removed rather than corrected: `bytecode.md` is the declared single source of truth, and duplicating it is how this drifted in the first place.
- **`adding-features.md` instructed the reader** to add a `NodeTag` variant, a `Data` union variant, and switch arms in `identifyPrimitives`/`markConstants` — the pre-`sexpr_form` procedure, contradicting `.claude/rules/compiler-forms.md`. Following it would not have compiled. Rewritten to the current `FormKind` + `sexpr_form_map` path (also `dst: u8` → `u16`, and dispatch is in `compiler_ir.zig`, not `compiler.zig`).
- **The `OpCode` enum's own inline comments** claimed `dst:u8` too, so the source read as wrong as the doc. Fixed.

## The guard

`OpCode`, `NodeTag` and `FormKind` counts are now pinned at comptime in the `diagnostics.zig` style — changing one fails the build with a message naming the files to update. **Each gate was mutation-tested** (count off by one → build fails with the intended message; restored → builds clean), since a guard that never fires is worthless.

This would have caught everything here except `architecture.md`'s table, whose count was coincidentally right — which is exactly why that table is now gone instead of corrected.

## Also swept

`README.md`, `CLAUDE.md`, `docs/dev/README.md`, `understanding-map.md`, `claude-code-harness.md`, and the `/bytecode-isa` skill all repeated the same counts. The skill said **32** — so the count existed as 29 (doc), 32 (skill), and 31 (enum) simultaneously. `CHANGELOG.md` entries are left alone as historical record.

## Verification

- `zig build` clean; `zig fmt --check` clean
- `zig build test` passes
- `npx markdownlint-cli2` — 0 issues in 84 files (the CI `format` job)
- Opcode table checked programmatically against the enum and `fixed_operand_bytes`
- Disassembly example is verbatim real output
- `markTailPositions` propagation rules re-checked against the code

No behavior change — docs, comments, and comptime assertions only.

### DIFF
--- a/.claude/skills/bytecode-isa/SKILL.md
+++ b/.claude/skills/bytecode-isa/SKILL.md
@@ -5,7 +5,7 @@ description: Reference for the Kaappi bytecode instruction set
 # Bytecode ISA
 
 The instruction set reference lives in `docs/dev/bytecode.md` — read that
-file for the full 32-opcode table, operand encodings, closure capture
+file for the full 31-opcode table, operand encodings, closure capture
 encoding, and disassembler output format. Do not duplicate the table here;
 that doc is the single source of truth.
 
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -73,9 +73,15 @@ reviews:
 
     - path: "src/ir.zig"
       instructions: |
-        Intermediate representation: 33 node types, 3 analysis passes,
-        5 optimization passes. New node types need: NodeTag variant, Data union
-        field, freeNode arm, markTailPositions arm, and llvm_node_table entry.
+        Intermediate representation: 18 node types, 1 analysis pass
+        (markTailPositions), 5 optimization passes. Most special forms are NOT
+        their own node type — they share the single `sexpr_form` tag and are
+        discriminated by one of 18 `FormKind`s, so a new form normally needs
+        only the FormKind + sexpr_form_map path described for src/compiler*.zig
+        above, with no NodeTag, Data union, freeNode or analysis-pass change.
+        Only a genuinely structured node (one with IR-level children) needs:
+        NodeTag variant, Data union field, lowering, freeNode arm,
+        markTailPositions arm, and llvm_node_table entry.
         Optimization passes use `else => return node` — only add arms if the
         new form should participate in constant folding or simplification.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -240,9 +240,9 @@ cannot resolve. `zig cc` includes these automatically.
 
 ```text
 Source → Reader → Expander → IR → Analysis → Optimization → Bytecode Emission → VM
-         (UTF-8    (syntax-    (33 node  (tail pos,    (const fold,     (register-   (generational
-          lexer)    rules)      types)    primitives,   dead branch,      based)       GC)
-                                          constants)    boolean, etc.)
+         (UTF-8    (syntax-    (18 node  (tail pos)    (const fold,     (register-   (generational
+          lexer)    rules)      types)                  dead branch,      based)       GC)
+                                                        boolean, etc.)
 ```
 
 ### Pipeline stages
@@ -251,7 +251,7 @@ Source → Reader → Expander → IR → Analysis → Optimization → Bytecode
 |-------|------|------|
 | Reader | `reader.zig` (+ `reader_tokens.zig`, `reader_datum.zig`) | Tokenizer + recursive descent parser. Handles R7RS lexical syntax including Unicode identifiers and `#\λ` character literals. |
 | Expander | `expander.zig` | `syntax-rules` pattern matching and template instantiation. Called by the compiler when a macro keyword is encountered. |
-| IR | `ir.zig` | Lowers S-expressions to tree-structured IR (33 node types). Runs 3 analysis passes (tail positions, primitive identification, constant detection) and 5 optimization passes (constant folding, dead branch elimination, boolean simplification, identity elimination, begin simplification). See `docs/dev/ir.md`. |
+| IR | `ir.zig` | Lowers S-expressions to tree-structured IR (18 node types; `sexpr_form` carries 18 `FormKind`s). Runs 1 analysis pass (tail positions) and 5 optimization passes (constant folding, dead branch elimination, boolean simplification, identity elimination, begin simplification). See `docs/dev/ir.md`. |
 | Compiler | `compiler.zig` | IR nodes → register-based bytecode via `compileFromNode()`. Retains `compileExpr()` for forms delegated via `passthrough`. Dispatches derived forms to sub-modules. |
 | VM | `vm.zig` | Executes bytecode. Growable register file + call frame stack (heap-allocated, double-on-overflow). Handles continuations (stack-copying), exception handler stack, dynamic-wind stack, stepping debugger. |
 | GC | `memory.zig` | Generational collector (young/old) with minor and full collections, write barrier for old→young references. Roots tracked via `gc.pushRoot`/`gc.popRoot`. Triggered after N allocations. |
@@ -328,8 +328,7 @@ its struct lives outside `types.zig` entirely with no `types.Fiber` re-export.
 
 | File | Responsibility |
 |------|---------------|
-| `ir.zig` | IR node types (33), AST→IR lowering, 3 analysis passes, 5 optimization passes |
-| `ir_emitter.zig` | Standalone IR → bytecode emitter (used by Stage 1 parity tests) |
+| `ir.zig` | IR node types (18), AST→IR lowering, 1 analysis pass, 5 optimization passes |
 | `compiler.zig` | Core: IR pipeline orchestration (`compile()` lowers to IR, runs passes), retains `compileExpr()` for passthrough forms, scope/register management, macro forms |
 | `compiler_ir.zig` | IR-to-bytecode: `compileFromNode()` dispatch, if, begin, call, lambda, define, set!, and, or, when, unless |
 | `compiler_lambda.zig` | lambda, define, set!, begin, delay, delay-force, body compilation |
--- a/README.md
+++ b/README.md
@@ -338,7 +338,7 @@ Source → Reader → Expander → IR → Bytecode emission → VM
 |-----------|------|
 | **Reader** | Tokenizer + recursive descent parser for the full R7RS lexical syntax, including Unicode identifiers and `#\λ` character literals. |
 | **Expander** | `syntax-rules` pattern matching and hygienic template instantiation. |
-| **IR** | Tree-structured intermediate representation (33 node types) with analysis passes (tail positions, primitives, constants) and optimization passes (constant folding, dead-branch elimination, and more). |
+| **IR** | Tree-structured intermediate representation (18 node types) with a tail-position analysis pass and 5 optimization passes (constant folding, dead-branch elimination, and more). |
 | **Compiler** | IR → register-based bytecode. |
 | **VM** | Bytecode interpreter with growable register file and frame stack, exception handler and dynamic-wind stacks, stack-copying continuations, and a stepping debugger. |
 | **GC** | Generational collector (young/old) with write barrier for old→young references. |
--- a/docs/dev/README.md
+++ b/docs/dev/README.md
@@ -72,7 +72,7 @@ investigations.
 
 | Document | Contents |
 |----------|----------|
-| [bytecode.md](bytecode.md) | Bytecode instruction set (32 opcodes), encoding, disassembler |
+| [bytecode.md](bytecode.md) | Bytecode instruction set (31 opcodes), encoding, disassembler |
 | [repl.md](repl.md) | REPL reference: line editing, comma commands, completion |
 | [unicode-case-mapping.md](unicode-case-mapping.md) | Case-conversion coverage by script |
 | [fuzzing-feasibility.md](fuzzing-feasibility.md) | Why neither Fuzzilli nor AFL++ is the tool, the existing `std.testing.fuzz` targets, and where fuzzing can improve |
--- a/docs/dev/README.md
+++ b/docs/dev/README.md
@@ -20,7 +20,7 @@ investigation produced analysis worth keeping.
 |----------|----------|
 | [vision.md](vision.md) | Why Kaappi exists, what it values, how those values guide decisions |
 | [architecture.md](architecture.md) | Major subsystems: pipeline, value representation, VM, GC, file organization |
-| [ir.md](ir.md) | Compiler IR: 33 node types, analysis passes, optimization passes |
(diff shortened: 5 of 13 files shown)


baijum (commented):
Reviewed as the reporter of #2102. I re-derived the claims from source rather than
reading the diff, and the substance holds — with two misses in the sweep, described at
the end.

## What I verified independently

**The opcode table is consistent.** I extracted the `OpCode` enum and the
`fixed_operand_bytes` switch from the PR branch and compared them against the 31 table
rows programmatically:

```
enum 31 | fixed_operand_bytes 31 | doc rows 31
CONSISTENT: 31 rows, order and numbering match the enum,
every Bytes == 1 + fixed_operand_bytes, all operand widths well-formed
```

`closure`'s `5 + 3n` is right too — 1 opcode byte + 4 fixed + 3 per capture descriptor.

**The gate numbers are correct**: `OpCode` 31, `NodeTag` 18, `FormKind` 18, each matching
`@typeInfo(...).fields.len` on the branch.

**The phantom opcodes are genuinely phantom.** `get_local`, `set_local` and
`close_upvalue` have zero occurrences in `src/`. Deleting `architecture.md`'s duplicate
table rather than correcting it is the right call, and the stated reason is the
interesting one: its row count was coincidentally 31, so a count gate would have waved it
through. That is a good argument for not having two tables, independent of this fix.

## Two misses, and they share a cause

### `docs/dev/README.md:74` still says 32 opcodes

```
| [bytecode.md](bytecode.md) | Bytecode instruction set (32 opcodes), encoding, disassembler |
```

The count now reads three ways on this branch:

```
docs/dev/README.md:74        (32 opcodes)      ← stale
docs/dev/architecture.md:249  31 opcodes
docs/dev/bytecode.md:8        31 opcodes
```

so the 29/32/31 split the PR set out to close is now a 32/31 split. The file *was* edited
— line 23's `ir.md` row went 33 → 18 node types — so this is one row missed in a file
that was open, not a file overlooked.

**And the guard will not catch it next time.** Compare the two gates:

```zig
// types.zig — OpCode
"OpCode count changed: update the table in docs/dev/bytecode.md " ++
"and the count in .claude/skills/bytecode-isa/SKILL.md, then update this number"

// ir.zig — NodeTag
"NodeTag count changed: update docs/dev/ir.md, docs/dev/architecture.md, " ++
"docs/dev/README.md, CLAUDE.md and README.md, then update this number"
```

`docs/dev/README.md` is named in the `NodeTag` list and absent from the `OpCode` one —
which is exactly why the node-type row on line 23 got fixed and the opcode row on line 74
did not. The gap in the file list and the miss are the same gap.

Worth adding `docs/dev/architecture.md:249` to the `OpCode` message as well, since it now
carries the count too.

### `.coderabbit.yaml:76` still describes the old IR

```yaml
- path: "src/ir.zig"
  instructions: |
    Intermediate representation: 33 node types, 3 analysis passes,
    5 optimization passes. New node types need: NodeTag variant, Data union
    field, freeNode arm, markTailPositions arm, and llvm_node_table entry.
```

Both numbers are the ones this PR is fixing, and the procedure is the pre-`sexpr_form` one
that `adding-features.md` was just rewritten *away from*. Since this is configuration for
the automated reviewer, the effect is that the bot carries the corrected-away model into
every future PR and would advise contradicting the new `adding-features.md`. That makes it
arguably the worst remaining copy, not the most harmless one.

It is in neither gate's file list either.

## Suggested fix

Two one-line edits, plus widening the gate messages so the guard covers its own blast
radius:

- `docs/dev/README.md:74` — 32 → 31
- `.coderabbit.yaml:76` — 33 node types / 3 analysis passes → 18 / 1, and the "new node
  types need …" sentence replaced with the `FormKind` + `sexpr_form_map` path
- `types.zig`'s `OpCode` message — add `docs/dev/README.md` and `docs/dev/architecture.md`
- `ir.zig`'s `NodeTag` message — add `.coderabbit.yaml`

The rest of the sweep is clean: the only other survivors are the historical notes at
`docs/dev/ir.md:172` and `src/ir.zig:1091`, which explain the v0.13.0 removal and should
stay.

baijum:
Both findings confirmed against source and fixed. The diagnosis was the valuable part, and following it turned up four more instances of the same gap.

## The two sites

- `docs/dev/README.md:74` — 32 → 31
- `.coderabbit.yaml:76` — 33 node types / 3 analysis passes → 18 / 1

I re-ran the sweep across **every file type** rather than `*.md` + `*.zig`, since that filter is precisely what hid `.coderabbit.yaml` from the original pass. These two were the only survivors; the rest of the tree is clean.

## Why they were missed, and why the fix isn't just two more filenames

Your diagnosis is right and generalizes further than the two sites. Checking all three gate lists against the sweep, they had **six** gaps, not two:

| gate | named | actually states the count | missing |
|---|---|---|---|
| `OpCode` | bytecode.md, SKILL.md | + architecture.md, docs/dev/README.md, claude-code-harness.md | **3** |
| `NodeTag` | 5 files | + .coderabbit.yaml | **1** |
| `FormKind` | ir.md | + CLAUDE.md, architecture.md | **2** |

`FormKind` is the one that makes the point: both `CLAUDE.md` and `architecture.md` say "carries 18 `FormKind`s", and neither was listed.

So adding two names would have left four gaps. **A list of filenames inside an error message is its own drift surface — the exact bug class this gate exists for.** Each message now carries a search alongside the (completed) list, and states that the list is the known set rather than a guarantee.

## The search greps the noun, not the number

Worth recording, because grepping the number fails in both directions:

- **Separator variance** — `31-opcode` and `31 opcodes` need different patterns. That is exactly what hid `README.md:74`: my original sweep matched `32-opcode`, and the file says `(32 opcodes)`. Same blind spot as the gate-list gap, one layer down.
- **Order variance** — `18 node types` vs `IR node types (18)` vs `**18** node tags`.
- **A bare number is unusable** — 43 hits for `31` across docs, and `18` collides with every SRFI number.

Grepping `opcode` / `node (type|tag)` / `form ?kind` catches all shapes. Each search was verified to be a superset of its known sites; `llvm-backend.md` surfaces in one but states no count, so it is noise, not a miss.

## One place I did not follow the suggestion

You proposed replacing `.coderabbit.yaml`'s "new node types need …" sentence with the `FormKind` + `sexpr_form_map` path. I kept both instead: that `NodeTag` / `Data union` / `freeNode` / `markTailPositions` / `llvm_node_table` list is still **correct** for a genuinely structured node with IR-level children — it matches `.claude/rules/compiler-forms.md`'s second procedure. Replacing it outright would trade a stale instruction for a wrong one. It is now labelled as the structured-node path, with the `FormKind` path named as the normal case, which also resolves its contradiction with the sibling `src/compiler*.zig` block that already had the `FormKind` procedure right.

Agreed on your framing that this was the worst remaining copy: it steers the automated reviewer, so it would have propagated the corrected-away model into every future PR.

## Verification

`zig build`, `zig build test`, `zig fmt --check` clean; `markdownlint-cli2` 0 issues / 84 files; `.coderabbit.yaml` parses. All three gates re-mutation-tested after the rewrite. Every count claim in the repo now matches the source.

baijum:
Rebased onto `origin/main` (now at `3c1a56be`), no longer behind. Both commits replayed cleanly — no conflicts, no fixups.

Four commits had landed: #2105 (freed-alias pin), #2098 (Phase 7A port-satellite tracing), #2099 (Phase 6C shell completions), #2093 (`kaappi fmt` line endings). Only two files overlapped with this branch, both docs, and both in unrelated sections:

- `docs/dev/README.md` — main added the `cli-surface.md` row at line 46; my edits are the `ir.md` row (23) and the `bytecode.md` row (75)
- `CLAUDE.md` — main's 3-line change is elsewhere in the file

Nothing touched `src/types.zig` or `src/ir.zig` (#2098 is in `types_port.zig` / `gc_collect.zig` / `gc_sweep.zig`), so the gates were untouched by the replay.

**Re-verified on the rebased tree**, not assumed:

- Swept every count claim again across all file types, including main's newly added `docs/dev/cli-surface.md` — it states no opcode or node-type count, so it is not a new site. No stale `29`/`32`/`33`/`3 analysis passes` anywhere.
- Re-checked the opcode table programmatically against the `OpCode` enum and `fixed_operand_bytes`: 31 rows, order and byte widths match.
- Re-mutation-tested the `OpCode` gate on the merged tree — still fires.
- `zig build`, `zig build test`, `zig fmt --check` clean; `markdownlint-cli2` 0 issues across 85 files (one more than before — main's new doc).

`format` has since passed on the rebased head.

Issue #1900, opened 2026-07-31
run-all.sh globs are non-recursive: tests/scheme/srfi/slow/ has never run


baijum:
**Category:** Test infrastructure · **Severity:** coverage gap · **Found by:** systematic audit v2 reconnaissance (#1890)

`tests/scheme/run-all.sh` uses non-recursive globs (`tests/scheme/srfi/*.scm` at `:181`, and the same shape for every other suite). `tests/scheme/srfi/slow/` therefore matches nothing and has never been executed by the runner or by CI.

## Contents

```console
$ ls tests/scheme/srfi/slow/
srfi257-full.scm  srfi257-rx-full.scm
```

These are the **full SRFI 257 and 257-rx reference suites** — 159 assertions that no scheduled job runs. `grep -rn 'slow/' ` finds no reference from `run-all.sh`, `ci.yml`, or any other workflow.

`kaappi test` *does* pick them up (it recurses), which is part of why the two runners disagree on totals.

## Why it matters beyond these two files

The directory is invisible rather than skipped: nothing prints "SKIP", no counter reports it, and the suite summary reads as complete. Any future test placed in a subdirectory of an existing suite silently joins it.

## Options

1. Make the globs recursive and give slow files an explicit time budget or a `KAAPPI_TEST_SKIP` default.
2. Add an explicit `run_suite "SRFI (slow)" tests/scheme/srfi/slow/*.scm` guarded by an env flag, so the default run stays fast but the files are visibly skipped rather than absent.
3. Schedule them in a nightly job alongside the fuzz targets.

Whichever is chosen, `tests/scheme/CLAUDE.md`'s directory table also needs updating — it currently omits `fmt/`, `doctor/`, `cache/`, `pipeline/`, `timings/`, `test-runner/` and `acceptance/`, all of which exist and all of which `run-all.sh` does run. That stale table caused a reconnaissance agent to report those suites as never-executed, which was wrong.

Phase 3.10 of the audit strategy.

**Verified:** directory listing and `run-all.sh` glob read directly at `74efe596`; the 2026-07-31 baseline run reported 624 Scheme files, none from `slow/`.

baijum:
**Already fixed — closing as stale-open.** Verified against `main`:

- `tests/scheme/srfi/slow/` no longer exists; both reference suites were moved up into `tests/scheme/srfi/` by Phase 3.10 (#2071).
[...]

Issue #438, opened 2026-06-29
bytecode_file: writeF64 omits endianness conversion unlike all integer write methods


baijum:
## Bug

In `src/bytecode_file.zig`, the `Writer.writeF64` method (line 88-91) writes f64 values using raw `@bitCast` without endianness conversion:

```zig
fn writeF64(self: *Writer, allocator: std.mem.Allocator, v: f64) !void {
    const bytes: [8]u8 = @bitCast(v);
    self.buf.appendSlice(allocator, &bytes) catch return BytecodeError.OutOfMemory;
}
```

Meanwhile, all integer write methods (`writeU16`, `writeU32`, `writeU64`, `writeI64`) explicitly convert to little-endian:

```zig
fn writeI64(self: *Writer, allocator: std.mem.Allocator, v: i64) !void {
    const bytes: [8]u8 = @bitCast(std.mem.nativeToLittle(i64, v));
    // ...
}
```

Similarly, `Reader.readF64` (line 145-149) uses raw `@bitCast` without `littleToNative`, while all integer read methods use `littleToNative`.

## Impact

On all currently supported platforms (macOS ARM, Linux x86_64/ARM/RISC-V, WASM), this is not a problem since they are all little-endian. However:

1. The .sbc file format is inconsistent: integers are stored in a defined byte order (little-endian), but floats are stored in host byte order.
2. An .sbc file written on a little-endian machine would produce wrong float values if read on a big-endian machine (and vice versa). This affects flonums (`TAG_FLONUM`) and complex numbers (`TAG_COMPLEX`), which serialize real and imag parts as f64.

## Fix

Apply the same endianness conversion to f64 by reinterpreting as u64:

```zig
fn writeF64(self: *Writer, allocator: std.mem.Allocator, v: f64) !void {
    const bits: u64 = @bitCast(v);
    const bytes: [8]u8 = @bitCast(std.mem.nativeToLittle(u64, bits));
    self.buf.appendSlice(allocator, &bytes) catch return BytecodeError.OutOfMemory;
}

fn readF64(self: *Reader) !f64 {
    if (self.pos + 8 > self.data.len) return BytecodeError.CorruptedFile;
    const bytes = self.data[self.pos..][0..8];
    self.pos += 8;
    const bits = std.mem.littleToNative(u64, @bitCast(bytes.*));
    return @bitCast(bits);
}
```

PR #2145, opened 2026-08-01
Phase 7C: the big-endian canary — it ran the unit suite and r7rs-tests.scm, and SRFI 74's one endian assertion compared a function to its own definition

### LINKED ISSUE #2139: [Portability] The big-endian s390x canary runs no byte-order-sensitive test, and SRFI 74's native-agreement assertion cannot fail
**Category:** Portability / Tooling
**Spec reference:** none — this is a coverage claim, not a conformance one. The
claim under audit is `docs/dev/porting.md`: *"the `s390x-test` CI job guards the
invariant on every PR … the s390x job is the canary that catches it if it
doesn't."*
**Severity:** coverage gap (no wrong answer observed on any host)

## The gap

s390x is the project's only big-endian target and exists precisely to catch
byte-order bugs (#1654). `ci.yml`'s `s390x-test` job runs exactly three steps:

```yaml
- name: Build (cross-compile for s390x)
  run: zig build -Dtarget=s390x-linux
- name: Unit tests (s390x via QEMU)
  run: zig build test -Dtarget=s390x-linux
- name: R7RS test suite (s390x via QEMU)
  run: zig-out/bin/kaappi tests/scheme/r7rs/r7rs-tests.scm
```

It does **not** run `tests/scheme/run-all.sh`. `r7rs-tests.scm` imports no SRFI
with a byte-order-sensitive surface, so **no `tests/scheme/**` file carrying a
byte-order assertion has ever executed on a big-endian machine**:

| File | Endian-sensitive content | Ran on s390x |
|---|---|---|
| `tests/scheme/srfi/srfi74.scm` | blob big/little/native layouts | no |
| `tests/scheme/srfi/srfi160.scm`, `srfi4.scm`, `srfi66.scm`, `srfi63.scm` | host-native `NumericVector` element storage | no |
| `tests/scheme/audit/primitives_srfi160-audit.scm` | encoder-vs-`ref`-vs-printer cross-decoder probe, whose own header cites this very gap | no |
| `tests/scheme/srfi/srfi135.scm` | — (**no UTF-16 coverage at all**, despite exporting `textual->utf16be`/`le`) | no |
| `tests/scheme/srfi/srfi178-audit.scm` | bit order, explicitly annotated "so the big-endian s390x CI leg sees identical values" | no |

`ppc64le-test` is **not** a second canary — POWER is little-endian in this
configuration — so s390x is the only leg where any of this could fail.

## The assertion that could not fail

Under the heading *"native accessors agree with explicit-endianness
accessors"*, `tests/scheme/srfi/srfi74.scm` asserted:

```scheme
(let ((b1 (make-blob 4)) (b2 (make-blob 4)))
  (blob-u32-native-set! b1 0 42)
  (blob-u32-set! (endianness native) b2 0 42)
  (test-equal (blob->u8-list b1) (blob->u8-list b2))   ;; <- tautology
  (test-equal 42 (blob-u32-native-ref b1 0)))
```

`lib/srfi/74.sld:109` defines the left-hand side as the right-hand side:

```scheme
(define (blob-u32-native-set! b k n) (blob-u32-set! (endianness native) b k n))
```

Both sides are the same call. The assertion cannot fail on any host, for any
value of `%host-big-endian?`.

**Reproduction (verified).** Break the native arm of the `endianness` macro so
it always answers `'big`:

```diff
-((_ native) (if (%host-big-endian?) 'big 'little))
+((_ native) 'big)
```

On little-endian macOS aarch64, ReleaseSafe at `94ebd5a0`:

* `tests/scheme/srfi/srfi74.scm` — **30 of 30 pass**, exit 0.
* the replacement suite added alongside — **7 failures**.

**Discriminating control:** with the mutation reverted, both files pass. The
30/30 is not a stale binary or a skipped file.

## Why no Scheme test can close it completely

`(endianness native)` is *defined* as `(if (%host-big-endian?) 'big 'little)`,
and `%host-big-endian?` is the only Scheme-visible witness of host byte order —
nothing else in `lib/` calls it, and SRFI 160 exposes no byte view of a
`NumericVector` (the six `%numeric-vector-*` primitives are create / predicate /
kind / length / ref / set!). So a Scheme assertion about native byte order can
only re-derive its expectation from the same primitive. The non-circular check
needs a memory probe, i.e. Zig — and the Zig unit suite is one of the three
things the s390x leg already runs.

## Also blind, and left open for Phase 7D

Every `.sbc` test writes and reads on the same host, so a byte-swap bug applied
to **both** directions cancels out and stays green everywhere. The nightly
`cross-diff` fuzz job (which does run s390x) does not close this either: each
side runs against its own arch's cache. No test writes a file on one endianness
and reads it on the other. Note #438 (`writeF64` omitted the endianness
conversion) was this exact class of bug and was found by inspection, not by a
test.

**Verified:** fresh ReleaseSafe build at `94ebd5a0`, isolated `KAAPPI_HOME`,
cache cleared. `zig build -Dtarget=s390x-linux` produces an `ELF 64-bit MSB
executable, IBM S/390` — the cross-build is fine; only the test selection is
the problem. **No big-endian execution was performed for this issue**; every
claim above is about which files run where, read off `ci.yml`, plus
little-endian mutation results.

**Found by:** Systematic audit v2, Phase 7C

baijum:
Audit v2, Phase 7C. Closes #2139.

## What `s390x-test` actually ran (verified from `ci.yml`)

Three steps: `zig build -Dtarget=s390x-linux`, `zig build test
-Dtarget=s390x-linux`, `zig-out/bin/kaappi tests/scheme/r7rs/r7rs-tests.scm`.
No `run-all.sh`. `r7rs-tests.scm` imports no SRFI with a byte-order-sensitive
surface, so **no `tests/scheme/**` file carrying a byte-order assertion had
ever executed on a big-endian machine** — including
`tests/scheme/audit/primitives_srfi160-audit.scm`, whose own section header
cites this gap by name.

`ppc64le-test` is **not** a second canary: POWER is little-endian in this
configuration. s390x is the only big-endian leg.

## The assertion that could not fail

`tests/scheme/srfi/srfi74.scm`, under *"native accessors agree with
explicit-endianness accessors"*:

```scheme
(blob-u32-native-set! b1 0 42)
(blob-u32-set! (endianness native) b2 0 42)
(test-equal (blob->u8-list b1) (blob->u8-list b2))
```

`lib/srfi/74.sld:109` defines the left as the right. **Measured:** hardcode
the `endianness` macro's native arm to `'big` and `srfi74.scm` still reports
**30 of 30 pass, exit 0**; the replacement suite reports **7 failures**.
Reverting the mutation makes both green — the 30/30 was not a stale binary.

## Why the fix is split between Zig and Scheme

`(endianness native)` is `(if (%host-big-endian?) 'big 'little)`, and that
primitive is the only Scheme-visible witness of host byte order — nothing else
in `lib/` calls it, and SRFI 160 exposes no byte view of a `NumericVector`
(its six primitives are create/predicate/kind/length/ref/set!). A Scheme
assertion can only re-derive its expectation from the same primitive. So:

**`src/tests_endian.zig`** (11 tests) holds what Scheme cannot express, and
needs **no CI change** — the unit suite is already one of the three things the
s390x leg runs.

* `%host-big-endian?` against an actual in-memory byte probe.
* The `.sbc` codec's canonical little-endian scalars. Every existing `.sbc`
  test writes and reads on the *same* host, so a paired byte-swap cancels out
  and stays green everywhere. These break the pairing one direction at a time:
  the writer (`writeU16/U32/U64/I64/F64/writeStr`) against **literal** expected
  bytes, and `readHeaderInfo` against a **hand-assembled literal-little-endian
  header** — with a discriminating control (the same header with only the two
  VERSION bytes swapped must be rejected) and a `comptime` guard that fails the
  build if a future VERSION makes that control vacuous.
  #438 (`writeF64` omitted the endianness conversion) was exactly this class of
  bug and was found by inspection, not by a test.
* A whole written `.sbc` file, checked at byte offsets 0–14.

**`tests/scheme/audit/endianness-audit.scm`** (99 assertions) covers the
Scheme surfaces:

* SRFI 74 explicit big/little at every width, signed and unsigned, plus
  `blob-uint/sint` at sizes 3 and 5 and the list conversions — byte-level, with
  every expected value a constant and every test value chosen so its
  byte-reverse is a *different* number.
* SRFI 74 native: the layout expectation is **derived from
  `%host-big-endian?`**, never hardcoded, plus the other half ("a native write
  differs from the opposite explicit order") so an arm that always answered
  `'big` cannot pass on a big-endian host either.
* SRFI 160: encoder ↔ `ref` ↔ printer agreement across all 10 multi-byte
  kinds. Those are two independent hand-written decoders; on little-endian they
  cannot disagree even if both were wrong.
* SRFI 135 UTF-16BE/LE including **surrogate pairs** and BOM dispatch.
  `tests/scheme/srfi/srfi135.scm` had **zero** utf16 coverage.
* Controls: UTF-8, bytevectors, `char->integer`, and two-limb bignums — all
  byte-oriented and required to be identical on every host.

**`tools/run-endian-suite.sh`** is the one command: the hub plus the nine
pre-existing suites that carry endian assertions. ~1s natively. Loud on a
missing file, since silently running nine of ten is how
`tests/scheme/srfi/slow/` went unnoticed for eleven days (#1900).

## CI change

One step added to `s390x-test` (the point) and to `riscv64-test` /
`ppc64le-test` as little-endian controls — a failure there means the suite is
wrong, not the host.

## Also fixed

`srfi74.scm`'s tautology is replaced, and **all 31 of its assertions are
named**: SRFI-64 prints a failed assertion's *value*, not its source, so an
unnamed failure on an emulated leg is a bare `#f` among 30 candidates. One
byte-level expectation I hand-computed there was wrong (`0x07` for `0x87`) and
the test caught it — which is the argument for byte-level assertions in
miniature.

`docs/dev/porting.md` claimed the s390x job "guards the invariant on every
PR". It now says what the job runs and what remains blind, including two
things worth stating outright:

* **A paired byte-swap still cancels**, and the nightly `cross-diff` fuzz job
  does not close it either — each side runs against its own arch's cache. Left
  for Phase 7D.
* **A local `zig build test -Dtarget=s390x-linux` proves nothing about
  behaviour.** `build.zig` sets `skip_foreign_checks = true`, so on macOS the
  run step is *skipped* and the command exits 0 with no output (`--summary all`
  shows `run test unit-tests skipped`). It executes on the CI leg only because
  `setup-qemu-action` registers binfmt_misc, so the direct spawn succeeds and
  Zig never consults its foreign-executor logic.

## Verification status

**Verified, little-endian macOS aarch64 ReleaseSafe at `94ebd5a0`:**
`zig build test` green; `run-all.sh` **2060 pass / 0 fail** on a clean tree
(an earlier run showed 2 compile-suite failures caused by cross-compiling
concurrently into the same `zig-out/`, which cleared on a native rebuild —
not attributable to the diff); the endian suite 10/10; the mutation results
above; `zig fmt --check` and markdownlint clean.

**Verified about building, not behaviour:** `zig build -Dtarget=s390x-linux`
produces an `ELF 64-bit MSB executable, IBM S/390`, and the test binary
cross-compiles for s390x.

**Not verified locally, by construction:** every claim about how this code
*behaves* big-endian. No s390x execution was performed on the authoring
machine, and none is possible there. **The `s390x-test` leg on this PR is the
first real big-endian run of any of it** — read that check before merging, and
treat a failure there as a finding rather than as a flake.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -201,6 +201,13 @@ jobs:
       - name: R7RS test suite (riscv64 via QEMU)
         run: zig-out/bin/kaappi tests/scheme/r7rs/r7rs-tests.scm
 
+      # The little-endian control for the same step on s390x below: every
+      # assertion in these suites is either host-independent or derived from
+      # `%host-big-endian?`, so a failure here means the suite itself is wrong,
+      # not that the host's byte order is.
+      - name: Endian-sensitive Scheme suites (riscv64 via QEMU)
+        run: bash tools/run-endian-suite.sh zig-out/bin/kaappi
+
   # s390x: interpreter tier like riscv64 (#1654), and the only big-endian
   # target in the matrix — a permanent canary for byte-order bugs in the
   # `.sbc` codec and any future serialization. Also validated on a
@@ -235,6 +242,17 @@ jobs:
       - name: R7RS test suite (s390x via QEMU)
         run: zig-out/bin/kaappi tests/scheme/r7rs/r7rs-tests.scm
 
+      # THE POINT OF THE CANARY (audit v2, Phase 7C). Until this step, the
+      # three things that ran here were the build, the unit suite and
+      # `r7rs-tests.scm` -- and `r7rs-tests.scm` imports no SRFI with a
+      # byte-order-sensitive surface, so SRFI 74 (blobs), 160 (host-native
+      # numeric-vector storage) and 135 (UTF-16) had never executed on a
+      # big-endian machine at all. `run-all.sh` is too heavy for an emulated
+      # leg; this runs just the files that carry endian assertions, in about
+      # a second natively. See tools/run-endian-suite.sh.
+      - name: Endian-sensitive Scheme suites (s390x via QEMU)
+        run: bash tools/run-endian-suite.sh zig-out/bin/kaappi
+
   # ppc64le: interpreter tier like riscv64 (#1654). Also validated on a
   # real-kernel Alpine ppc64le VM (unit + thottam + R7RS + full Scheme
   # suites; 128 TB = 2^47 default map window fits the NaN-box payload).
@@ -266,6 +284,13 @@ jobs:
 
       - name: R7RS test suite (ppc64le via QEMU)
         run: zig-out/bin/kaappi tests/scheme/r7rs/r7rs-tests.scm
+
+      # POWER is little-endian in this configuration despite the architecture
+      # family's big-endian history, so this is a control leg, not a second
+      # canary -- worth stating, because "ppc64le covers big-endian too" is an
+      # easy and wrong assumption to make from the job list.
+      - name: Endian-sensitive Scheme suites (ppc64le via QEMU)
+        run: bash tools/run-endian-suite.sh zig-out/bin/kaappi
 
   # FreeBSD: cross-compile the binaries and test executables from Linux
   # (guarding the cross-compilation path release.yml ships with), then
--- a/docs/dev/porting.md
+++ b/docs/dev/porting.md
@@ -309,6 +309,38 @@ all, only build/CI work — *provided* the preconditions hold.
       precondition now expresses: new serialization or byte-punning code
       must use the same explicit-conversion helpers — the s390x job is
       the canary that catches it if it doesn't.
+
+      **Exactly what that job runs, and what it therefore cannot see.**
+      Four steps: the cross-compile, `zig build test
+      -Dtarget=s390x-linux`, `tests/scheme/r7rs/r7rs-tests.scm`, and (as
+      of audit v2 Phase 7C) `bash tools/run-endian-suite.sh`. It does
+      **not** run `tests/scheme/run-all.sh`, which is why the endian
+      suite is a separate, deliberately small entry point rather than
+      the whole tree. Two blind spots remain, both structural:
+
+      * **A paired byte-swap cancels out.** Every `.sbc` test writes and
+        reads on the same host, so a bug that swapped *both* directions
+        stays green everywhere. `src/tests_endian.zig` breaks the pairing
+        one direction at a time — the writer against literal expected
+        bytes, the reader against a hand-assembled literal-little-endian
+        header — but no test yet writes a file on one endianness and
+        reads it on the other. The nightly `cross-diff` fuzz job does not
+        close this either: both sides run against their own arch's cache.
+      * **A local `zig build test -Dtarget=s390x-linux` proves nothing
+        about behaviour.** `build.zig` sets `skip_foreign_checks = true`,
+        so on a host with no emulator registered the run step is
+        *skipped* and the command still exits 0 (`--summary all` shows
+        `run test unit-tests skipped`). On the CI leg it does execute,
+        because `docker/setup-qemu-action` registers binfmt_misc and the
+        direct spawn succeeds. Locally that command is a compile gate,
+        which is exactly what it was added for.
+      * **`%host-big-endian?` is the only Scheme-visible witness of host
+        byte order.** SRFI 74's `(endianness native)` is defined as
+        `(if (%host-big-endian?) 'big 'little)`, so no Scheme test can
+        check that primitive non-circularly — a test can only ask it
+        again. It is pinned against an actual memory probe in
+        `src/tests_endian.zig`, which is why that pin is Zig and not
+        Scheme.
 - [ ] **f64 hardware or soft-float** with IEEE-754 semantics (flonums
       are NaN-boxed doubles; NaN canonicalization assumes IEEE bit
       patterns).
--- a/src/tests_endian.zig
+++ b/src/tests_endian.zig
@@ -52,7 +52,6 @@ const std = @import("std");
 const builtin = @import("builtin");
 const types = @import("types.zig");
 const memory = @import("memory.zig");
-const platform = @import("platform.zig");
 const file_utils = @import("file_utils.zig");
 const bf = @import("bytecode_file.zig");
 const bfw = @import("bytecode_file_write.zig");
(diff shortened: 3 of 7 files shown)


baijum:
Post-push addendum: a byte-punning sweep of `src/` (`asBytes`/`bytesAsSlice`/`@bitCast` to byte arrays/`mem.writeInt`/`mem.readInt`) turned up one surface the first pass missed, now in the suite.

**SRFI 271 determinized random ports.** `primitives_random_port.zig` and `types_port.zig:373` move u64s in and out of a bytevector with an explicit `.little`, so a seeded stream is meant to be byte-identical on every host. But `tests/scheme/srfi/srfi271.scm` asserts only `(test-equal (take-bytes a 128) (take-bytes b 128))` — two ports **on the same host** — which is the same cancelling pairing the `.sbc` tests have. Running it big-endian catches a write and read side disagreeing; a cross-host golden byte sequence is still missing, the same shape of gap 7D owns for `.sbc`.

For completeness, the rest of the sweep came back clean:

| Surface | Verdict |
|---|---|
| `bytecode_file_{write,read}.zig` | explicit `nativeToLittle`/`littleToNative`; now pinned both directions |
| `primitives_srfi160.zig` + `printer.zig` | host-native, and **not observable from Scheme** — no byte view of a `NumericVector` exists, so the only checkable property is that the two independent decoders agree with the encoder |
| `doctor.zig:719` | explicit `.little` |
| `primitives_srfi260.zig:77` | `asBytes` over a scalar filled with OS entropy — the hex differs by host, but the value is random, so it is endian-visible and semantically irrelevant |
| bytecode operands in `Function.code` | assembled and consumed a byte at a time (`v >> 8`, `v & 0xFF`), never byte-punned — endian-independent by construction |
| bignum limbs | `[]u64` in little-endian *limb* order, each limb serialized through `writeU64` |
| SRFI 178 bit order | spec-defined, not host-defined; the file already says so |
| strings / UTF-8 / bytevectors | octet-defined; included as the suite's discriminating controls |

Suite is now 11 files, still ~1s natively.

Issue #2144, opened 2026-08-01
[Tooling] thottam list/verify/update build paths from unvalidated package names — isValidPkgName guards install and remove only


baijum:
**Category:** Tooling
**Spec reference:** `src/thottam_state.zig:42` — `isValidPkgName` restricts a
package name to alphanumerics, `-` and `_`, which is what keeps
`joinPath(src_dir, name)` and `joinPath(lib_dir, name)` inside `$KAAPPI_HOME`.
**Severity:** edge-case (hardening — not exploitable on its own; see below)

`isValidPkgName` is applied at the top of `doInstall` (`src/thottam.zig:386`)
and `doRemove` (`src/thottam.zig:525`). It is **not** applied by `doList`,
`doVerify`, or `doUpdate`, which take a name straight out of `installed.txt` or
`thottam.lock` and build a filesystem path from it.

```console
$ NAME='../../../../../../tmp/kaappi-6e-outside'
$ printf '%s\n' "$NAME"          >  $KAAPPI_HOME/installed.txt
$ printf '%s deadbeef\n' "$NAME" >  $KAAPPI_HOME/thottam.lock

$ thottam list
Installed packages:
  ../../../../../../tmp/kaappi-6e-outside  deadbeef

$ thottam verify
  MISMATCH: ../../../../../../tmp/kaappi-6e-outside (locked: deadbeef, actual: 4f3c...)

$ thottam update "$NAME"
Updating ../../../../../../tmp/kaappi-6e-outside...
There is no tracking information for the current branch.
```

`verify` ran `git rev-parse HEAD` in an unrelated repository outside
`$KAAPPI_HOME` and reported its SHA; `update` got far enough into that
repository for git to complain about *its* branch configuration, i.e. `git -C
<escaped path> pull` executed there. Neither command noticed the name.

**Discriminating control — the guard works where it is applied.** The same name
through the two commands that do validate:

```console
$ thottam remove "$NAME"
error: invalid package name '../../../../../../tmp/kaappi-6e-outside' (only alphanumeric, '-', '_' allowed)
$ thottam install "$NAME"
error: invalid package name '../../../../../../tmp/kaappi-6e-outside' (only alphanumeric, '-', '_' allowed)
```

The escaped repository was intact afterwards. The guard also holds for a name
arriving from an untrusted manifest's `depends:` line, which is the case a user
never types — that path goes through `doInstall`, so it is checked.

**Why this is filed as hardening rather than a vulnerability.** Reaching it
[...]

Issue #2138, opened 2026-08-01
[Doc-truth] kaappi.pkg 'source:' and 'name:' are parsed and never read — the documented third-party hosting field is inert


baijum:
**Category:** Doc-truth
**Spec reference:** `CLAUDE.md` § "kaappi.pkg manifest (ecosystem repos)" —

> ```
> name: kaappi-json
> version: 0.2.0
> build: make          # only if csrc/ exists
> depends: kaappi-net  # space-separated specs, supports name[@ver][::url]
> source: https://github.com/yourname/kaappi-json  # optional, for third-party packages
> ```

and § "Package manager (thottam)":

> All fields except `name` are optional. The `source` field declares where
> this package is hosted (for third-party packages).

**Severity:** doc-truth

Of the fields documented above, `thottam` reads exactly two.

`parsePkgManifest` (`src/thottam_state.zig:57`) recognises four keys — `name:`,
`depends:`, `build:`, `source:` — and `getPkgManifest` dutifully heap-copies all
four. But grepping every use of the returned struct in `src/thottam.zig` finds
only `manifest.depends` (lines 491, 594) and `manifest.build_cmd` (500, 650).
**`manifest.name` and `manifest.source` are parsed, allocated, freed, and never
read.** `version:` is not in the parser's key list at all — it is silently
ignored like any unknown key.

**What that means in practice:**

- `source:` has no effect. It does not become the clone URL, and it does not
  reach the lockfile — only a `::url` typed on the command line or written into
  a `depends:` spec does (`parsed.source`, not `manifest.source`). So the field
  documented as declaring "where this package is hosted", for exactly the
  third-party case where it would matter, is inert.
- `name:` is never checked against the package being installed, so the doc's
  "all fields except `name` are optional" describes a requirement nothing
  enforces — a manifest may omit `name:` entirely, or name a different package,
  and the install succeeds either way.

**Minimal reproduction:**

```console
$ cat kaappi.pkg          # in the repo published as kaappi-srcfield
name: WRONG-NAME-ENTIRELY
source: https://evil.example.com/not-this-repo

$ thottam install kaappi-srcfield
  kaappi-srcfield installed (locked at 9635560ef01e...)
$ cat ~/.kaappi/thottam.lock
kaappi-srcfield 9635560ef01ed9919d4ce26668e14cd9ce2fc5f0     # no source column
$ thottam list
Installed packages:
  kaappi-srcfield  9635560ef01e                              # no "(from: ...)"
```

[...]

Issue #2137, opened 2026-08-01
[Tooling] thottam --locked checks the SHA but not the source URL, then overwrites the lockfile's recorded provenance with the one it was handed


baijum:
**Category:** Tooling
**Spec reference:** `thottam --help` — "`--locked`  Refuse to install packages
not in the lockfile"; `CLAUDE.md` § "Package manager (thottam)" — "The lockfile
(`~/.kaappi/thottam.lock`) records source URLs for provenance."
**Severity:** wrong-result (security-adjacent)

A lockfile line is `<pkg> <sha> [<source-url>]`. `--locked` enforces the SHA
column and ignores the source column entirely: the clone URL comes from
whatever this invocation was handed (`parsed.source`, else `$KAAPPI_ORG/<pkg>.git`
— `src/thottam.zig:449`), never from the lockfile. Worse, once the SHA check
passes, `updateLockfile` **overwrites the recorded source** with the URL that was
just used (`src/thottam.zig:518`), so the provenance record is destroyed by the
operation that was supposed to be checking it.

A fork that shares history satisfies the SHA check, which is the ordinary case
for a fork.

**Minimal reproduction** (offline; `FORK-kaappi-alpha.git` is a clone of
`kaappi-alpha.git`, so the SHA is reachable in both):

```console
$ thottam install "kaappi-alpha::$ORG/kaappi-alpha.git"
$ cat thottam.lock
kaappi-alpha 75f253e118c0976ac2caabec574af39394e7bc4e /tmp/org/kaappi-alpha.git

$ rm -rf src lib && : > installed.txt            # fresh machine, restore from lock
$ thottam --locked install "kaappi-alpha::$ORG/FORK-kaappi-alpha.git"
Installing kaappi-alpha@75f253e118c0976ac2caabec574af39394e7bc4e...
  Cloning /tmp/org/FORK-kaappi-alpha.git...
  Checking out 75f253e118c0976ac2caabec574af39394e7bc4e...
  Resolved: 75f253e118c0976ac2caabec574af39394e7bc4e
  kaappi-alpha installed (locked at 75f253e118c0976ac2caabec574af39394e7bc4e)
$ echo $?
0
$ git -C src/kaappi-alpha config remote.origin.url
/tmp/org/FORK-kaappi-alpha.git                   # not what the lockfile said

$ cat thottam.lock
kaappi-alpha 75f253e118c0976ac2caabec574af39394e7bc4e /tmp/org/FORK-kaappi-alpha.git
[...]

Issue #2136, opened 2026-08-01
[Tooling] thottam remove deletes library files by name with no ownership record — removing one package breaks another that is still installed


baijum:
**Category:** Tooling
**Spec reference:** `CLAUDE.md` § "Package manager (thottam)" — "Copies `.sld`
files to `~/.kaappi/lib/` (preserving directory structure)" and
`thottam remove kaappi-web  # uninstall`.
**Severity:** wrong-result

Every package's `lib/` tree is merged into one flat shared directory
(`$KAAPPI_HOME/lib`), and nothing records which package a given installed file
came from. `doRemove` therefore removes by *name*: it walks the package's own
source tree for `.sld` paths and unlinks each one from the shared directory
(`removeSldFiles`, `src/thottam.zig:286`). If another still-installed package
ships a file at the same relative path, it goes too.

**Minimal reproduction.** Two packages, each shipping its own library plus
`lib/kaappi/shared.sld`:

```console
$ thottam install kaappi-one && thottam install kaappi-two
$ ls ~/.kaappi/lib/kaappi/
one.sld  shared.sld  two.sld

$ thottam remove kaappi-two
Removing kaappi-two...
  kaappi-two removed

$ ls ~/.kaappi/lib/kaappi/
one.sld                      # shared.sld is gone

$ thottam list
Installed packages:
  kaappi-one  9dcab67aba64   # still installed, per thottam

$ echo '(import (kaappi shared))(display s)' > t.scm && kaappi t.scm
t.scm:1: error[KP2001]: library not found: (kaappi.shared)
```

`kaappi-one` is reported as installed and is broken. `thottam verify` also
reports `OK: kaappi-one` — it compares git SHAs, and the source checkout is
untouched; only the installed copy was deleted.

The install side of the same missing ownership is silent too: installing
`kaappi-two` **overwrote** `kaappi-one`'s `shared.sld` with no diagnostic, so the
running behaviour of an already-installed package changed as a side effect of
installing an unrelated one. `installLibTree` documents overwriting as intended
("existing files are overwritten (updates re-copy)") — which is right for an
update of the *same* package and wrong across packages, and there is currently no
way for the code to tell those apart.

**Discriminating control:** removing `kaappi-one` from a home where only
`kaappi-one` is installed correctly deletes both `one.sld` and `shared.sld` and
leaves nothing behind — so removal itself is not over-broad; it is specifically
missing the "does anyone else still need this file" question.

[...]

PR #2146, opened 2026-08-01
Phase 6E: thottam audit — 301 assertions, and a nightly tag installs as a release

baijum:
Phase 6E of the [systematic correctness audit v2](https://github.com/kaappi/kaappi/issues/1890): `thottam`, the package manager.

`thottam` is 932 lines that ship in every release for every platform, clone from GitHub, run a `build:` command out of a downloaded manifest, and copy files into `~/.kaappi/lib`. It had **3 tests**, all of them on the caret operator.

## What this adds

**`src/tests_thottam.zig` — 233 assertions, 46 test blocks, no network.** Version parsing, all seven constraint operators at their boundaries, the `name[@ver][::url]` spec grammar, `kaappi.pkg` manifest parsing, the package-name guard, and the lockfile/installed-list round trip. Runs on all 18 CI legs via `zig build test`.

**`tests/scheme/thottam/thottam-lifecycle.sh` — 68 assertions, still no network.** The whole install/list/verify/remove/update lifecycle against a directory of local **bare git repositories** pointed at by `KAAPPI_ORG`: git clones a plain path as happily as a URL, so clone, `ls-remote --tags` constraint resolution, checkout, lib copy and lockfile write all run unchanged offline. Wired into `run-all.sh` as a new "Package manager" suite.

Both are hermetic — `KAAPPI_HOME` and `KAAPPI_ORG` point at throwaway directories. That is load-bearing rather than tidy: `thottam remove` calls `removeTree`.

**Oracle.** SemVer 2.0.0 for what a *version* is; node-semver's "Ranges" for what `^` and `~` *mean*, since neither operator appears in the SemVer spec at all. Every assertion says which one it is pinning where they could disagree.

## Issues filed

Assertions documenting a filed defect are commented out with a `FAIL: #N` marker stating the property that *should* hold; the fix PR re-enables them.

| # | Finding |
|---|---|
| [#2130](https://github.com/kaappi/kaappi/issues/2130) | `Semver.parse` reads three dot-separated components and discards the rest, and hands each to `std.fmt.parseInt` — so Zig's integer-literal grammar becomes the version grammar. `v2.0.0.nightly-UNRELEASED` parses as 2.0.0 and `>=1.0.0` installs it; `v1_0.0.0` parses as **10.0.0** and outranks `v2.0.0`. |
| [#2131](https://github.com/kaappi/kaappi/issues/2131) | An omitted version component is filled with `0` and then treated as written, so `~1` means `1.0.x` instead of node-semver's `1.x`. |
| [#2132](https://github.com/kaappi/kaappi/issues/2132) | Six distinct constraint parse failures all surface as `no version matching X`. |
| [#2133](https://github.com/kaappi/kaappi/issues/2133) | A CRLF lockfile makes `verify` print `MISMATCH (locked: 75f253e118c0, actual: 75f253e118c0)`. |
| [#2134](https://github.com/kaappi/kaappi/issues/2134) | `install pkg@ver` on an installed package is a silent no-op, exit 0; `update` can never move a package off a pin. |
| [#2135](https://github.com/kaappi/kaappi/issues/2135) | `verify` iterates the lockfile, not `installed.txt` — an empty lockfile reports `All packages verified.` and exits 0. |
| [#2136](https://github.com/kaappi/kaappi/issues/2136) | `remove` unlinks `.sld` files by name with no ownership record, breaking a still-installed package. |
| [#2137](https://github.com/kaappi/kaappi/issues/2137) | `--locked` enforces the SHA but not the source URL, then overwrites the recorded provenance. |
| [#2138](https://github.com/kaappi/kaappi/issues/2138) | Doc-truth: `kaappi.pkg` `source:` and `name:` are parsed and never read. |
| [#2144](https://github.com/kaappi/kaappi/issues/2144) | `isValidPkgName` guards `install` and `remove` only; `list`/`verify`/`update` build paths from state-file names. |

### The most striking one

Two fixtures identical except for a single tag:

```console
$ git -C repo tag ; thottam install 'kaappi-odd@>=1.0.0'
v1.0.0
v2.0.0-rc1
  Resolved >=1.0.0 -> v1.0.0                      # correct: pre-releases excluded

$ git -C repo tag ; thottam install 'kaappi-nightly@>=1.0.0'
v1.0.0
v2.0.0.nightly-UNRELEASED
  Resolved >=1.0.0 -> v2.0.0.nightly-UNRELEASED   # a nightly marker, installed as a release
```

## Negative results, recorded so they are not re-tested

- **A package name cannot escape `$KAAPPI_HOME/lib`** through `install` or `remove`. The unit suite pins this as a property over the whole byte range rather than a spot check, so a future "allow dots in names" change has to come through that test first. The transitive `depends:` path is guarded too — a manifest naming `../../../../../../tmp/kaappi-ESCAPED` as a dependency fails the install.
- git refuses the `ext::` transport outright (`fatal: transport 'ext' not allowed`), so a `pkg::ext::sh -c …` spec is not a way in.
- The #614 option-injection guard on source URLs holds (`pkg::-upload-pack=…` falls back to the org default).
- The #403 lockfile prefix-collision fix holds (`kaappi-net` does not match `kaappi-net-extra`), as does #613's empty-after-quote-trim guard.

## What could not be tested offline

Nothing that mattered. The `KAAPPI_ORG` fixture reaches `install`, `update`, `remove`, `verify`, `list` and the full constraint resolver. Not covered: real HTTPS/SSH transport error handling, GitHub-specific behaviour, and `runBuildCommand` beyond an empty `build:` (running a real `make` would need a native fixture; the Windows refusal path is `comptime`-gated and unreachable natively). The shell suite skips on Windows — it builds POSIX-path git repositories — but the unit suite covers the parsing surface on all 18 legs, which is where the CRLF finding lives.

**Verified:** fresh ReleaseSafe build at `94ebd5a0`; `zig build test` 1638/1641 (3 skipped), `bash tests/scheme/run-all.sh` green including the new suite.

### DIFF
--- a/src/tests_thottam.zig
+++ b/src/tests_thottam.zig
@@ -314,7 +314,7 @@ test "parseConstraints accepts a four-part range (issue #2132)" {
     // try std.testing.expect(semver.parseConstraints(">= 1.0.0") != null);
 }
 
-test "isConstraintSpec routes exactly the four operator-led forms" {
+test "isConstraintSpec routes the six operator-led forms and nothing else" {
     try std.testing.expect(semver.isConstraintSpec(">=1.0.0"));
     try std.testing.expect(semver.isConstraintSpec(">1.0.0"));
     try std.testing.expect(semver.isConstraintSpec("<=1.0.0"));
@@ -571,11 +571,16 @@ test "updateLockfile records and replaces the provenance URL in place" {
     defer if (eb.source) |s| allocator.free(s);
     try std.testing.expectEqualStrings("sha-b", eb.sha);
 
-    // And must not duplicate the rewritten one.
+    // And must not duplicate the rewritten one, nor leave the superseded
+    // line behind. The old line is `a sha-a https://h/a` — matching on
+    // `sha-a` alone would also match `sha-a2`, so the whole line is the
+    // only spelling that can actually fail here.
     const content = try thottam.readFile(allocator, lock);
     defer allocator.free(content);
-    try std.testing.expectEqual(@as(usize, 1), std.mem.count(u8, content, "sha-a2"));
-    try std.testing.expect(std.mem.indexOf(u8, content, "sha-a\n") == null);
+    try std.testing.expectEqual(@as(usize, 1), std.mem.count(u8, content, "a sha-a2 https://h/a2\n"));
+    try std.testing.expect(std.mem.indexOf(u8, content, "a sha-a https://h/a\n") == null);
+    // Two entries in, two lines out — an in-place rewrite, not an append.
+    try std.testing.expectEqual(@as(usize, 2), std.mem.count(u8, content, "\n"));
 }
 
 test "lockfile lookup does not match a package whose name is a prefix (issue #403)" {
--- a/src/tests_thottam.zig
+++ b/src/tests_thottam.zig
@@ -1,0 +1,710 @@
+//! Audit-campaign v2 Phase 6E: `thottam`, the package manager.
+//!
+//! The three pre-existing tests here covered the caret operator only. This
+//! file covers the rest of the surface that needs no network: version
+//! parsing, the seven constraint operators and their boundaries, the
+//! `name[@ver][::url]` spec grammar, `kaappi.pkg` manifest parsing, the
+//! package-name guard that keeps an install inside `$KAAPPI_HOME/lib`, and
+//! the lockfile reader/writer round trip.
+//!
+//! **Oracle.** For the constraint operators the oracle is SemVer 2.0.0
+//! (semver.org) for what a *version* is, and node-semver's range grammar
+//! (github.com/npm/node-semver, "Ranges") for what `^` and `~` *mean* —
+//! `^` and `~` are not in the SemVer spec at all, they are npm's, and every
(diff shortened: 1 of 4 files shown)


baijum:
All four CodeRabbit findings were valid and are fixed in 8572939.

The `saved_lock` one was the most useful: using the variable as suggested turned up a sharper symptom of [#2137](https://github.com/kaappi/kaappi/issues/2137) — a `--locked` restore does not merely *overwrite* the lockfile's recorded source URL, it **erases** it, because `doInstall` rewrites the entry from `parsed.source`, which is `null` when no `::url` was typed. A committed lockfile therefore degrades to the `$KAAPPI_ORG` default after one round trip through the flag meant to preserve it. Pinned as a disabled assertion with an enabled control showing the SHA half does survive; [reported on the issue](https://github.com/kaappi/kaappi/issues/2137#issuecomment-5153270670).

The lockfile round-trip finding was a real hole in my own test: `indexOf(content, "sha-a\n")` could never fire, since an entry carrying a source URL does not end with its SHA. It now matches the whole superseded line and pins the line count, so an append-instead-of-replace regression is caught.

Assertion counts are unchanged (233 Zig + 69 shell); `zig build test` and the lifecycle suite are green.

Issue #2172, opened 2026-08-02
[SRFI-152/196] string-segment and range-segment hang on n=0 — the same missing check as #1949 and #2084


baijum:
**Category:** SRFI
**Severity:** crash (unbounded loop — the process never returns)

`(string-segment s 0)` and `(range-segment r 0)` **loop forever**. Verified with a 15-second alarm; `n=1` is the control and returns correctly:

```
(string-segment "abcdef" 1)                 -> (a b c d e f)      rc=0
(string-segment "abcdef" 0)                 -> <no output>        rc=142  (killed)
(range-segment (numeric-range 0 6) 0)       -> <no output>        rc=142  (killed)
```

Sources: `lib/srfi/152.sld:202` and `lib/srfi/196.sld:105`. Both advance by `n` per iteration, so `n = 0` never advances.

**This is a fourth and fifth member of a class already filed twice** — #1949 (`Uvector-segment` with `n=0` hangs) and #2084 (`bitvector-segment`, where `n=0` recurses without bound). Four `segment`-family procedures across four different SRFIs, all with the same missing precondition.

**The decisive evidence that this is a missing check rather than something inherent to the shape:** SRFI 171's `tsegment` **raises** on `n = 0` instead of hanging. So one implementation in this tree already gets it right, and the fix is to do what it does.

Worth fixing as one change rather than four: every `segment`/`chunk` procedure should reject a non-positive size at entry. A grep for the family is the whole audit — and it is worth doing now, because two of the four were found only by someone thinking to try `0` on a *different* SRFI than the one already filed.

**Verified:** fresh ReleaseSafe build at `0bba2a1f`, macOS aarch64, isolated `KAAPPI_HOME`. Each hang reproduced under a 15 s alarm; `rc=142` is the alarm, not a clean exit.

**Found by:** Systematic audit v2, Phase 8 (synthesis) — while checking whether two separately-filed issues were the same defect.

Issue #2157, opened 2026-08-01
[CI] Five ci.yml steps run r7rs-tests.scm bare — the (chibi test) shim exits 0 on a failed assertion, so 1,395 assertions gate nothing there


baijum:
**Category:** Tooling (CI)
**Severity:** coverage gap (silent — a wrong answer on five legs is reported green)

Five `ci.yml` steps invoke `tests/scheme/r7rs/r7rs-tests.scm` bare and take the
process exit status as the verdict. That file uses the `(chibi test)` shim,
which has **no exit-on-fail epilogue** — it prints its counts and exits 0 no
matter how many assertions failed. `run-all.sh` has always known this and
parses the counts instead (`tests/scheme/run-all.sh`, the `R7RS_FAIL` awk).
The five CI steps do not, so on those legs the suite's **1,395 assertions
catch only a crash, never a wrong answer.**

## Minimal reproduction

Copy the suite, break exactly one assertion, run it:

```console
$ cp tests/scheme/r7rs/r7rs-tests.scm /tmp/probe/
$ # insert `(test 999999 (+ 1 1))` right after the first (test-begin ...)
$ zig-out/bin/kaappi --lib-path lib /tmp/probe/r7rs-tests.scm | grep FAIL
FAIL [4.1 Primitive expression types]: expected 999999 got 2
$ echo $?
0
```

`EXIT=0` with a failed assertion in the log. Verified 2026-08-02 on a fresh
ReleaseSafe build of `910f7c75` (macOS aarch64, cache isolated).

**Discriminating control:** an ordinary corpus file — `(srfi 64)` with the
exit-on-fail epilogue — exits **1** on the same kind of mutation. The defect
is specific to `r7rs-tests.scm`'s runner, not to Scheme tests generally, which
is exactly why the bare spelling looks safe.

## The five steps

| job | step |
|---|---|
| `riscv64-test` | `R7RS test suite (riscv64 via QEMU)` |
| `s390x-test` | `R7RS test suite (s390x via QEMU)` |
| `ppc64le-test` | `R7RS test suite (ppc64le via QEMU)` |
| `windows-arm-test` | `R7RS test suite` |
| `windows-x64-test` | `R7RS test suite` |

All five are `run: <kaappi> tests/scheme/r7rs/r7rs-tests.scm` with nothing
downstream of it.

## Why it matters most on s390x

`s390x-test` is the project's **big-endian canary** (#1654) and its stated
[...]

Issue #2153, opened 2026-08-01
[Reactor] the WASI poll_oneoff backend's fd path is executed by no test, and the unit suite does not compile for wasm32-wasi


baijum:
**Category:** GC / Native — reactor backend parity
**Spec reference:** `docs/dev/porting.md` Stage 3 (fd readiness), the checklist item that defines a reactor backend as done:

> - [ ] The fd-readiness unit suites (`tests_reactor.zig`, `tests_scheduler.zig`, `tests_port_io.zig`) pass. If the OS lacks POSIX pipes/socketpairs, extend `testing_helpers.zig`'s `makeFdPair`/`makeBidiFdPair` (Windows substitutes loopback TCP pairs) rather than skipping the suites.

**Severity:** coverage gap (no live wrong answer found)

`src/reactor.zig` has four backends behind one `Reactor` API and they are required to be behaviourally interchangeable. CI executes the shared unit suite against three of them. The fourth — `WasiPollBackend` — is executed by nothing, and is not even compile-gated.

### 1. The unit-test module does not compile for wasm32-wasi

```
$ zig build test -Dtarget=wasm32-wasi
...
error: 16 compilation errors      # unit-tests
error: 4 compilation errors       # thottam-tests
```

The 20 errors are 32-bit-target arithmetic and platform-facade issues in code the test module pulls in, not in the reactor:

| count | file | error |
|--:|---|---|
| 4 | `src/platform.zig` | `no field named 'ACCMODE' in struct 'c.O__struct_13395'` |
| 2 | `src/platform.zig` | `type '*void' does not support field access` |
| 2 | `src/testing_helpers.zig` | `type 'void' has no members` |
| 1 | `src/thottam_proc.zig` | `type 'void' not a function` |
| 1 | `src/tests_deepcopy.zig` | `expected type 'usize', found 'u64'` |
| 1 | `src/ffi_callback.zig` | `type 'usize' cannot represent integer value '281474976710656'` |
| 2 | `src/ffi.zig` | `type 'usize' cannot represent integer value '20015998343868'` / `'140737488355327'` |

This is the same 32-bit-`usize` class as #1912, at different sites.

The contrast is with Windows, which `build.zig` deliberately keeps as a compile gate:

```zig
// Cross-compiled test binaries the host can't execute (no emulator
// registered — e.g. aarch64- or x86_64-windows) skip cleanly instead
// of failing, so `zig build test -Dtarget=<arch>-windows` is a
[...]

PR #2173, opened 2026-08-02
Close the campaign at 53 of 53, and correct the claim Phase 8 disproved

baijum:
**The audit campaign is complete.** 53 of 53 units, **186 issues filed**, grouped by Phase 8 into **35 root causes**.

## A correction of mine, first

Phase 8 was asked to *refute* the leads it was handed rather than confirm them. It refuted three — one of which was mine and had already landed in a merged tracker entry.

I wrote that SRFI 166's 11 issues "share close to one root cause". **Phase 8 disproved it by experiment rather than argument**: routing the supposedly-derived issues through the *working* procedural mechanism still produces wrong answers, so the failures are consumer-side. **Eleven distinct causes.** The 3.5 entry now carries the correction *above* the original reading rather than quietly replacing it, so the record shows what was believed and why it was wrong.

## The finding I would put in front of a maintainer

Phase 8's priority argument is the campaign's central result stated as an action:

> **Fix the green-but-tests-nothing class first — because until failure is detectable, every other fix's regression test is of unknown value.**

Five structurally different mechanisms found this campaign: regression tests exercising only the valid case (#682, #550 — both *closed* on that strength); 54 files with no exit path (#2116); `zig build test -Dtarget=s390x-linux` exiting 0 having run nothing (#2139); a WASI test suite that cannot compile (#2153); and **1,395 R7RS assertions gating nothing on five CI legs** because the `(chibi test)` shim exits 0 on failure (#2157).

The campaign opened against a **fully green** suite — 2,019 assertions passing. That gap is the whole story.

## Two structural items nobody had filed

- `isSpecialTopLevelForm` is a **fourth** hand-maintained parallel list.
- The `segment`-at-`n=0` class has **four** members, not the two already filed — #2172, verified here: `string-segment` and `range-segment` hang while `n=1` is clean, and SRFI 171's `tsegment` *raises* correctly, which is what proves a missing precondition rather than an inherent shape.

## Marker inventory

101 distinct `;; FAIL:` numbers, and **none cites a closed issue** — the convention held for the entire campaign. Two blind spots worth knowing: one marker cites a *merged PR* rather than an issue (a mechanical state audit reads that as done), and three use `;;;`, which a grep for the documented convention misses.

53 ticked, 0 unticked; markdownlint clean.


baijum:
Both review comments were right, and checking them found the reasons rather than just reconciling the digits.

**186 vs 188.** The campaign filed **188**; **186** carry the footer. The three that do not — #2129, #1870, #1920 — are precisely the ones Phase 8 flagged as recoverable only by label or by the `;; FAIL:` markers. So the discrepancy was carrying real information about the campaign's own bookkeeping, and flattening it to one number would have destroyed that. The status line now states both and names the three. Date moved to 2026-08-02.

**R25's 8 vs 3.5's 11.** Both correct; different populations; the doc said neither. Verified against the synthesis comment's own group membership lists:

| | |
|---|---|
| Phase 3.5 filed | **11** SRFI 166 issues |
| R25 holds | **8** — #2054, #2056, #2058, #2059, #2061, #2063, #2064, #2065 |
| #2062, #2066 → | **R21** *byte vs codepoint vs column* |
| #2067 → | **R11** |

That is the regrouping doing exactly what it exists for: three issues filed under one library turned out to belong to cross-cutting themes, and two of them to a measurement bug shared with #2100, #1996 and #1925. Both counts are now stated with what each measures.

No source changes; markdownlint clean.

Issue #2179, opened 2026-08-02
[SRFI-42] first-ec, any?-ec, every?-ec don't stop early — (first-ec #f (:integers i) i) hangs


baijum:
Found while working on #2177 (out of scope there, so filing separately).

SRFI 42 specifies that `first-ec` "stop[s] the loop after the first value", and that `any?-ec`/`every?-ec` stop "as soon as the result is known" — the spec's own examples rely on this to use them with infinite generators. The port in `lib/srfi/42.sld` instead materializes the *entire* comprehension first:

- `first-ec` expands to `(list-ec ...)` then takes `car`,
- `any?-ec`/`every?-ec` run the full `do-ec` loop and just record the answer.

## Repro

```scheme
(import (srfi 42))
(first-ec #f (:integers i) i)        ; hangs — spec says ⇒ 0
(any?-ec (:integers i) (> i 5))      ; hangs — spec says ⇒ #t
```

Finite comprehensions give the right answers, just eagerly — so this is silent extra work in the finite case and a hang in the legal infinite case.

## Fix sketch

The library already has the mechanism: `%do-ec` threads a mutable stop flag `s` that every generator loop checks before each iteration (it's what `:while`/`:until` set). All three collectors can allocate the flag themselves and set it from the body:

```scheme
(define-syntax first-ec
  (syntax-rules ()
    ((_ default qualifier ... expr)
     (let ((%s #f) (%result default))
       (%do-ec %s qualifier ...
               (begin (set! %result expr) (set! %s #t)))
       %result))))
```

with the analogous one-liner bodies for `any?-ec` (set result `#t` and stop when `expr` is true) and `every?-ec` (set `#f` and stop when false). `min-ec`/`max-ec`/`append-ec` also materialize a full list via `apply`, but those are spec-permitted (no early-exit semantics) — a separate, optional cleanup.

Priority `high` per the rubric: a legal program hangs, in a path the spec's own examples reach.

PR #2180, opened 2026-08-02
Complete SRFI 42: generic ':' dispatch, :real-range, :char-range, :dispatched

### LINKED ISSUE #2177: [SRFI-42] Generic `:` qualifier is missing, along with :real-range, :char-range, and :dispatched — (list-ec (: i 5) i) is invalid syntax
Found while verifying on-camera REPL demos for the @KaappiScheme channel against the shipped binary (v0.22.0, build `ca18ae6`, aarch64-macos). `kaappi features --json` lists 42 among the portable SRFIs.

`lib/srfi/42.sld` implements eleven typed qualifiers (`:list :string :vector :range :integers :port :do :let :parallel :while :until`) but not the generic dispatching qualifier `:` — the form every SRFI 42 example leads with — nor `:real-range`, `:char-range`, or `:dispatched`.

## Repro

```scheme
(import (srfi 42) (scheme base) (scheme write))
(display (list-ec (: i 5) i))
```

```
compile error[KP2001]: invalid syntax
```

Same failure for `(string-ec (:char-range c #\a #\e) c)` and `(list-ec (:real-range x 0 1 1/4) x)`. `:dispatched` is likewise absent from the export list (not tested beyond that).

## What already works

Typed qualifiers, guards, and the collectors behave:

```scheme
(list-ec (:range i 1 11) (* i i))                        ; ⇒ (1 4 9 16 25 36 49 64 81 100)
(string-ec (:string c "hello") c)                        ; ⇒ "hello"
(list-ec (:range i 1 31) (if (zero? (modulo i 15))) i)   ; ⇒ (15 30)
```

Since the qualifier simply isn't defined, `list-ec`'s macro finds no matching pattern and the whole comprehension collapses into `invalid syntax` — so this is a coverage gap in the port, not an expander failure. Whether the fix is the spec's full `:dispatched` machinery or a simpler compile-time type dispatch, `(list-ec (: i 5) i)` failing is the first thing anyone pasting a SRFI 42 tutorial into the REPL hits.

baijum:
Fixes #2177.

`(list-ec (: i 5) i)` — the first form in every SRFI 42 tutorial — was `KP2001 invalid syntax`: the port implemented eleven typed qualifiers but not the generic `:`, nor `:real-range`, `:char-range`, or `:dispatched`. Investigating also showed three qualifiers the library already **exported** — `:port`, `:do`, and `:parallel` — had no `%do-ec` rules at all and failed identically inside a comprehension. All of these work now.

## Design

The port is direct-style (nested loops around the body with a stop flag), not the reference implementation's CPS design, so `:` cannot dispatch by `:do` fusion. It instead uses the SRFI's own **runtime generator-procedure protocol** (the one `:dispatched`'s spec text defines): a dispatcher maps the evaluated argument list to a generator procedure `g`, stepped as `(g empty)` until it returns the eq?-unique sentinel.

- The initial dispatcher covers the spec's cases: lists, strings, vectors, 1–3 exact integers → `:range`, 1–3 reals → `:real-range`, 2 chars → `:char-range`, port `[+ reader]` → `:port`, plus the n>3 all-lists/strings/vectors cases.
- The extension surface is complete and exported: `:-dispatch-ref`, `:-dispatch-set!`, `make-initial-:-dispatch`, `dispatch-union`, `:generator-proc` — user dispatchers compose per spec (covered by a test that extends `:` to symbols).
- `:generator-proc` compiles the standard typed generators straight to closure constructors (no call/cc); any other generator form falls back to a call/cc coroutine over `do-ec`.
- `:real-range` follows the reference formula exactly (i counts 0,1,… while `i < (to−from)/step`, value `from + i·step`, with `from` turning inexact when `to` or `step` is), so `(:real-range x 0 1 1/4)` stays exact: `(0 1/4 1/2 3/4)`.
- `:char-range` is inclusive of both endpoints.
- Typed generators now accept multiple arguments (`(:list x '(1 2) '(3))` concatenates), matching the spec and the runtime dispatch path.

## Documented deviations (library header)

- The `(index i)` variable form remains unsupported (this port never had it).
- `:parallel` accepts only single-variable sub-generator forms; raw `:do` and `:while`/`:until`-wrapped generators are rejected at expansion time rather than merged.
- The reference implementation's `make-initial-:-dispatch` tests `(string? a1)` twice where `a2`/`a3` were meant — corrected here.

## Tests

`tests/scheme/srfi/srfi42.scm` grows from 40 to **86 assertions**: every issue repro (`(: i 5)`, `(string-ec (:char-range c #\a #\e) c)`, `(:real-range x 0 1 1/4)`), all dispatch arities and types, port dispatch with default (hygienically resolved) and custom readers, `:do` short and full six-slot forms, `:parallel` lockstep incl. an infinite `:integers` and a coroutine-backed `:let`, `:dispatched` with a hand-written dispatcher, the `:generator-proc` protocol on both the fast path and the coroutine fallback, `dispatch-union` identification and conflict conventions, and the catchable unrecognized-argument error.

Full local run: `bash tests/scheme/run-all.sh` → **2066 pass, 0 fail** (1 pre-existing WASM-differential skip); `kaappi check lib/srfi/42.sld` clean; suite also passes under `--no-ir-opt` and via `kaappi test`.

The pre-existing eager-termination gap in `first-ec`/`any?-ec`/`every?-ec` found during this work is out of scope here and tracked as #2179.

### DIFF
--- a/lib/srfi/42.sld
+++ b/lib/srfi/42.sld
@@ -1,17 +1,57 @@
 ;;; SRFI 42 — Eager Comprehensions
+;;;
+;;; Direct-style port: each qualifier expands to a loop nested around the
+;;; rest of the comprehension, with a mutable stop flag (s) threaded
+;;; through for :while/:until early exit.  This differs from the
+;;; reference implementation's CPS design (everything reduced to a :do
+;;; form), so the generic dispatching layer (: / :dispatched /
+;;; :generator-proc) uses the SRFI's *runtime* generator-procedure
+;;; protocol instead of :do fusion: a generator procedure g is called as
+;;; (g empty) and returns the next value of its sequence, or the
+;;; eq?-unique sentinel `empty' once exhausted — exactly the protocol
+;;; the :dispatched spec text defines.  :generator-proc compiles the
+;;; standard typed generators straight to closure constructors (%gen-*);
+;;; any other generator form falls back to a call/cc coroutine over
+;;; do-ec, which therefore shares the engine's general continuation
+;;; limits (stepping a coroutine-backed generator after the native frame
+;;; it was created under has returned is unsupported; the standard types
+;;; never touch call/cc).
+;;;
+;;; Deliberate deviations from the reference implementation:
+;;;   - The (index i) variable form is not supported (in any generator —
+;;;     as before this port never had it).
+;;;   - :parallel sub-generators must be plain single-variable generator
+;;;     forms (typed generators, :, :dispatched, :let, ...); raw :do
+;;;     forms and :while/:until-wrapped generators are rejected at
+;;;     expansion time rather than merged.
+;;;   - The reference's make-initial-:-dispatch tests (string? a1) twice
+;;;     in its 2- and 3-argument string cases where a2/a3 were meant —
+;;;     corrected here.
+;;;   - :parallel steps every sub-generator once per iteration and stops
+;;;     at the first exhausted one, so longer generators are advanced one
+;;;     step past the shortest — the same one-step-ahead the reference's
+;;;     fused loops perform.
 (define-library (srfi 42)
-  (import (scheme base))
+  (import (scheme base)
+          (scheme cxr)
+          (scheme read))
   (export list-ec string-ec vector-ec
           append-ec sum-ec product-ec
           min-ec max-ec
           first-ec last-ec any?-ec every?-ec
           do-ec fold-ec fold3-ec
-          :list :string :vector :range :integers
-          :port :do :let :parallel :while :until)
+          : :list :string :vector :range :real-range :char-range
+          :integers :port :do :let :parallel :while :until
+          :dispatched :generator-proc
+          :-dispatch-ref :-dispatch-set!
+          make-initial-:-dispatch dispatch-union)
   (begin
 
     ;; Generator keywords — error when used outside a comprehension.
     ;; Defined first so %do-ec can resolve them as syntax-rules literals.
+    (define-syntax :
+      (syntax-rules ()
+        ((_ rest ...) (error ": used outside a comprehension"))))
     (define-syntax :list
       (syntax-rules ()
         ((_ rest ...) (error ":list used outside a comprehension"))))
@@ -24,6 +64,12 @@
     (define-syntax :range
       (syntax-rules ()
         ((_ rest ...) (error ":range used outside a comprehension"))))
+    (define-syntax :real-range
+      (syntax-rules ()
+        ((_ rest ...) (error ":real-range used outside a comprehension"))))
+    (define-syntax :char-range
+      (syntax-rules ()
+        ((_ rest ...) (error ":char-range used outside a comprehension"))))
     (define-syntax :integers
       (syntax-rules ()
         ((_ rest ...) (error ":integers used outside a comprehension"))))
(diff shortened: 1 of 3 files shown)


lib/srfi/42.sld
```diff
@@ -45,13 +91,215 @@
...
+              empty))))
+
+    (define (%gen-real-range from to step)
+      (if (not (and (real? from) (real? to) (real? step)))
+          (error "arguments of :real-range are not real" from to step))
+      (let* ((a (if (and (exact? from)
+                         (not (and (exact? to) (exact? step))))
+                    (inexact from)
+                    from))
+             (istop (/ (- to a) step))
```
baijum, line 173: **[high] `:real-range` / `%gen-real-range` divide by zero when `step` is 0**

`istop` is computed as `(/ (- to a) step)` with no guard against `step = 0`. The matching `%do-ec` arm for `(:real-range var a b c)` does the same with `(%lim (/ (- %b %x) %c))` (see comment on that arm).

Any comprehension with a zero step fails at setup/first expansion, e.g.:
- `(list-ec (:real-range x 0 1 0) x)`
- `(list-ec (: x 0.0 1.0 0.0) x)`
- `(:generator-proc (:real-range 0 1 0))`

Please reject `step = 0` (or otherwise define the empty/error case) in both the runtime generator and the `%do-ec` macro arm.

lib/srfi/42.sld
```diff
@@ -45,13 +91,215 @@
...
+               (set! n (vector-length cur))
+               (set! k 0)
+               (set! rest (cdr rest))
+               (loop))
+              (else empty))))))
+
+    (define (%gen-range from to step)
+      (let ((x from))
+        (lambda (empty)
+          (if (if (positive? step) (< x to) (> x to))
```
baijum, line 160: **[high] Integer `:range` with `step` 0 hangs when `from > to`**

`%gen-range` treats a non-positive step with `(> x to)` instead of `(< x to)`. When `step` is `0`, `x` never advances (`(+ x 0)`), so if `from > to` the condition stays true forever and the generator yields the same value in an infinite loop.

The same pattern appears in the existing `%do-ec` `(:range var a b c)` macro arm (`(if (positive? %c) ...)` + `(+ var %c)`).

Examples: `(list-ec (:range i 5 3 0) i)` or `(list-ec (: i 5 3 0) i)` via the generic dispatcher.

Please reject `step = 0` (or treat it as an empty range) in both paths.

lib/srfi/42.sld
```diff
@@ -72,56 +320,116 @@
...
+         (%do-ec s (:real-range var a b 1) rest1 rest2 ...))
+        ((_ s (:real-range var a b c) rest1 rest2 ...)
+         (let ((%a a) (%b b) (%c c))
+           (if (not (and (real? %a) (real? %b) (real? %c)))
+               (error "arguments of :real-range are not real" %a %b %c))
+           (let* ((%x (if (and (exact? %a)
+                               (not (and (exact? %b) (exact? %c))))
+                          (inexact %a)
+                          %a))
+                  (%lim (/ (- %b %x) %c)))
```
baijum, line 358: **[high] Same divide-by-zero as `%gen-real-range`**

This `%do-ec` arm also computes `(%lim (/ (- %b %x) %c))` without rejecting `%c = 0`. Fix should stay in sync with `%gen-real-range`.

baijum (commented):
Bugbot review of this PR found **2 high-severity** issues around zero `step` in range generators. Inline comments below.

lib/srfi/42.sld
baijum: Confirmed, and fixed in d289716d. `%gen-range` is new in this PR (reachable via `(: i 5 3 0)` and `(:generator-proc (:range 5 3 0))`): with a zero step it never advances `x`, so `from > to` yielded the same value forever. The macro arm predates the PR but behaves identically — and the reference implementation rejects a zero step in `:range` explicitly, so the port had silently diverged from it. Both paths now raise the reference's exact message ("step size must not be zero in :range"); `from < to`, which was a quiet empty before, errors too. Regression tests cover the macro, generic-dispatch, and `:generator-proc` routes.

lib/srfi/42.sld
baijum: Fixed in d289716d, with one correction to the diagnosis. For an **exact** zero step this already failed *loudly* at setup — `(guard (e (#t (error-object-message e))) (list-ec (:real-range x 0 1 0) x))` returned "division by zero", a catchable error and byte-for-byte the reference implementation's behavior (its `:real-range` has no zero-step check either). The genuinely dangerous case is the one this comment's second example names but mischaracterizes: `(: x 0.0 1.0 0.0)` does **not** fail at setup — `(/ 1.0 0.0)` is `+inf.0`, so `istop = +inf.0` and the comprehension loops forever yielding `0.0` (in the reference too). Both `%gen-real-range` and the macro arm now reject a zero step explicitly ("step size must not be zero in :real-range"), which closes the silent inexact loop and upgrades the exact case's diagnostic; the library header documents this as a deliberate deviation beyond the reference. Tests pin the exact, inexact, and dispatch routes.

lib/srfi/42.sld
baijum: Fixed in d289716d, in sync with `%gen-real-range` (same check, same message). See the analysis on the `%gen-real-range` comment: the exact-step case was already a loud catchable error identical to the reference's behavior — the real hazard here was an **inexact** zero step, which produced `istop = +inf.0` and looped forever rather than dividing by zero.

lib/srfi/42.sld
```diff
@@ -67,61 +324,125 @@
...
-        ((_ s (:until (:string sargs ...) test) rest1 rest2 ...)
-         (%do-ec s (:string sargs ...) (:until test) rest1 rest2 ...))
-        ((_ s (:until (:vector vargs ...) test) rest1 rest2 ...)
-         (%do-ec s (:vector vargs ...) (:until test) rest1 rest2 ...))
-        ((_ s (:until (:integers iargs ...) test) rest1 rest2 ...)
-         (%do-ec s (:integers iargs ...) (:until test) rest1 rest2 ...))
+        ((_ s (:while (g garg ...) test) rest1 rest2 ...)
+         (%do-ec s (g garg ...) (:while test) rest1 rest2 ...))
+        ((_ s (:until (g garg ...) test) rest1 rest2 ...)
+         (%do-ec s (g garg ...) (:until test) rest1 rest2 ...))
```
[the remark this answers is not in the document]
baijum: Confirmed against the spec ("Runs `<generator>` **while** `<expression>` evaluates to non-#f" — the test scopes to the wrapped generator) and fixed in 6d947041. `(list-ec (:range j 2) (:while (:list x '(1 2 3)) (< x 2)) (cons j x))` returned `((0 . 1))`, dropping `j = 1` entirely; it now returns `((0 . 1) (1 . 1))`. The wrap rules give the wrapped generator a private stop flag, and a translator qualifier re-threads the enclosing flag for the rest of the chain while mirroring an ambient stop inward — so the bare `(:while test)`/`(:until test)` forms (this port's extension; the spec's grammar has no bare form) keep their stop-everything meaning even from inside a wrapped generator. Both semantics are documented in the header and pinned by four new tests.

One knock-on discovery: the new rules pushed `%do-ec` past the engine's 32-rules-per-`syntax-rules` cap, which rejects the whole definition as a bare `InvalidSyntax` — the qualifier processor is now split into `%do-ec` (generators) chaining to `%do-ec-more` (grouping/command/control/guards), and the cap's unhelpful diagnostic is filed as #2184.

lib/srfi/42.sld
```diff
@@ -1,17 +1,62 @@
...
+;;;     corrected here.
+;;;   - A zero step is rejected loudly in :real-range too, not just in
+;;;     :range (where the reference already rejects it).  The reference's
+;;;     :real-range instead raises division-by-zero for an exact zero
+;;;     step and loops forever on an inexact one (istop = +inf.0);
+;;;     silent non-termination is the failure mode worth closing.
+;;;   - :parallel steps every sub-generator once per iteration and stops
+;;;     at the first exhausted one, so longer generators are advanced one
+;;;     step past the shortest — the same one-step-ahead the reference's
+;;;     fused loops perform.
```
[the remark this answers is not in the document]
baijum: Implemented in 6d947041 rather than documented away — with one correction to the finding: SRFI 42 spells the grouping qualifier `(nested <qualifier>*)`, no colon; there is no `:nested` anywhere in the spec. The same grammar production also includes the `(begin <sequence>)` command qualifier, which this port was equally missing and the review didn't flag — each is a one-line rule in this direct-style design, so both now work: `(list-ec (nested (:range i 2) (:range j 2)) (list i j))` groups, `(begin ...)` runs its commands once per binding, and a body that is itself a `begin` form still parses as the body (all tested).

tests/scheme/srfi/srfi42.scm
```diff
@@ -90,6 +90,173 @@
...
+(define (upto-dispatcher args)
+  (if (and (= 1 (length args)) (exact-integer? (car args)))
+      (let ((n (car args)) (i 0))
+        (lambda (empty)
+          (if (< i n)
+              (let ((v i)) (set! i (+ i 1)) v)
+              empty)))
+      #f))
+(test-equal ":dispatched custom dispatcher" '(0 1 2)
+  (list-ec (:dispatched i upto-dispatcher 3) i))
```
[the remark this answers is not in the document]
baijum: Fixed in 6d947041, with a stronger test than the direct `(upto-dispatcher '())` call proposed: the new assertion drives the **library's** failure path — `%dispatch-gproc` reports `(d '())` as the second error irritant — so the identification is observed through `:dispatched` itself: `(cadr (error-object-irritants e))` ⇒ `upto`. The helper now follows the convention the second dispatcher in the file already did, so the two examples agree. (Small generation artifact worth fixing upstream: the committable suggestion writes `(else \`#f\`)` — a quasiquoted `#f`.)

tests/scheme/srfi/srfi42.scm
```diff
@@ -90,6 +90,173 @@
...
+ (dispatch-union
+  (:-dispatch-ref)
+  (lambda (args)
+    (if (and (= 1 (length args)) (symbol? (car args)))
+        (:generator-proc (:string (symbol->string (car args))))
+        #f))))
+(test-equal ": extended to symbols via dispatch-union" '(#\f #\o #\o)
+  (list-ec (: c 'foo) c))
+(test-equal ": still handles integers after extension" '(0 1 2)
+  (list-ec (: i 3) i))
```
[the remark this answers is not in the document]
baijum: Both points accepted in 6d947041. The "keep last" rot was introduced by my own second commit, and the analysis of why the zero-step tests still passed regardless — the raise happens inside the first dispatcher before the union ever consults the second — is exactly right. The mutating block now really is the last section of the file, so every earlier `:` test targets the unmutated initial dispatcher, and `dispatch-union`'s conflict branch has a regression test: two local dispatchers built on the exported `:generator-proc` (the `%gen-*` constructors are internal, as noted), both claiming one-symbol argument lists, asserting the "dispatching conflict" error.

tests/scheme/srfi/srfi42.scm
```diff
@@ -90,6 +90,173 @@
...
+(test-equal ":real-range zero step raises (dispatch path)"
+  "step size must not be zero in :real-range"
+  (caught-message (lambda () (list-ec (: x 0.0 1.0 0.0) x))))
+
+;;; --- multi-argument typed generators (concatenation) ---
+(test-equal ":list multi-arg" '(1 2 3) (list-ec (:list x '(1 2) '(3)) x))
+(test-equal ":string multi-arg" "abcd" (string-ec (:string c "ab" "cd") c))
+(test-equal ":vector multi-arg incl. empty" '(1 2 3 4)
+  (list-ec (:vector x #(1 2) #() #(3 4)) x))
+
```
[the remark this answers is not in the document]
baijum: Added in 6d947041 — plus one case the proposed tests would not catch: three well-typed strings pass under the reference's `(string? a1)` typo too, since every argument *is* a string, so a happy-path 3-string test cannot distinguish the corrected dispatcher from the buggy one. The discriminating regression is the mixed-type case `(: c "ab" #(1) "cd")`: the corrected `(string? a2)` check routes it to "unrecognized arguments in dispatching", where the typo'd dispatcher would have (wrongly) selected `:string`. All three tests are in — 3-string, 3-vector, mixed — along with the `:generator-proc` zero-step case for `:real-range`, and with the dispatcher-mutating block moved to the end of the file they all run against the unmutated global dispatcher.

Issue #419, opened 2026-06-29
string->number with #e prefix does not return #f for +inf.0 and +nan.0


baijum:
## Bug

`string->number` in `src/primitives_numeric.zig` silently ignores the `#e` (exact) prefix when parsing `+inf.0`, `-inf.0`, `+nan.0`, and `-nan.0`. It should return `#f` since these values have no exact representation.

```scheme
(string->number "#e+inf.0") ;; => +inf.0  WRONG, should be #f
(string->number "#e-inf.0") ;; => -inf.0  WRONG, should be #f
(string->number "#e+nan.0") ;; => +nan.0  WRONG, should be #f
(string->number "#e-nan.0") ;; => +nan.0  WRONG, should be #f
```

The R7RS spec (section 6.2.7) says: *"If string is not a syntactically valid notation for a number [...], then string->number returns #f."* Since `#e` demands an exact result and infinity/NaN have no exact representation, the result must be `#f`.

**Root cause:** The infinity/NaN string checks at lines 758-761 return immediately with `makeFlonum(...)` **before** the `applyExactness` function is called. Even though `applyExactness` (line 683) correctly returns `types.FALSE` for inf/NaN with exact prefix, it is never reached for these values.

**Fix:** Move the `+inf.0`/`-inf.0`/`+nan.0`/`-nan.0` checks after prefix parsing, or apply exactness to their results before returning. When `exactness == .exact` and the parsed value is inf or NaN, return `types.FALSE`.

Issue #1892, opened 2026-07-31
[R7RS 7.1.1] Radix-prefixed integers skip the trailing-delimiter check: '(#b1p4) reads as (1 p4)


baijum:
**Category:** R7RS conformance · **Severity:** wrong-result · **Found by:** systematic audit v2 reconnaissance (#1890)

A radix-prefixed integer is accepted and *terminated* as soon as the digits stop being valid for that radix, with no check that what follows is a delimiter. The remainder of the token is then read as a **second datum**, so one malformed token silently becomes two well-formed ones.

## Reproduction

```scheme
'(#b1p4)   ;; expected: read error   actual: (1 p4)
'(#o1e3)   ;; expected: read error   actual: (1 e3)
'(#o0o10)  ;; expected: read error   actual: (0 o10)
```

Verbatim at `74efe596`, v0.22.1:

```text
(1 p4)(1 e3)
```

## Discriminating control

The `#t`/`#f` path **does** perform this check (`src/reader_tokens.zig:501`), so this is a missing guard on one path rather than a global reader policy. That contrast is what localises it.

## Why it matters beyond malformed input

Silently splitting one token into two changes arity and structure rather than raising: `(f #b1p4)` becomes a two-argument call. A reader that rejects the input is strictly better than one that invents a plausible parse.

## Source location

`src/reader_tokens.zig` — `readIntegerWithRadix` returns `.{ .fixnum = n }` at ~`:822` (and the rational arm at ~`:817`) without the `Reader.isDelimiter` guard used at `:501`.

## Care needed in the fix

Two valid syntaxes legitimately continue past the digits and must keep working:

- **SRFI 270 hex floats** — `#x1p4` ⟹ `16.0` via `bignum.parseHexFloat` at ~`:825`
- **SRFI 169 digit separators** — `#x1_f` ⟹ `31` via `bignum.stripUnderscores` (and `1__0`/`1_` correctly error today)

Also `1/2+3i`-style rational and complex tails. A sweep of every radix × every invalid trailing character is the right test shape — see Phase 1B of the audit strategy.

**Verified:** fresh ReleaseSafe build at `74efe596`, cache cleared, reproduced twice.

baijum:
**Adjacent gap a radix-scoped fix would leave live** (found in Phase 1A, #1917).

[...]

Issue #751, opened 2026-07-02
Bug: string->number ignores #e/#i exactness prefix for complex numbers


baijum:
## Description

`string->number` does not apply the `#e` (exact) or `#i` (inexact) exactness prefix when parsing complex number literals. The exactness prefix is correctly parsed and stored, but the complex number parsing paths return results directly without calling `applyExactness`.

**File:** `src/primitives_numeric.zig`, lines 1037-1084

All four complex number return paths bypass `applyExactness`:
- Line 1042: pure imaginary `+i`
- Line 1045: pure imaginary `-i`
- Line 1076: general `a+bi`
- Line 1082: pure imaginary like `+3i`

Each returns `gc.allocComplex(...)` directly instead of routing through `applyExactness`.

## Reproduction

```scheme
(exact? (string->number "#e1+2i"))
; => #f (wrong, should be #t with #e prefix)

(string->number "#e1+2i")
; => 1.0+2.0i (wrong, should be exact 1+2i)

(inexact? (string->number "#i1+2i"))
; => #t (happens to be correct since complex is inexact by default,
;        but not by explicit application of the #i prefix)
```

## Expected behavior

Per R7RS 7.1.1, the exactness prefix `#e` / `#i` applies to the entire numeric literal:
- `(string->number "#e1+2i")` should return an exact complex number `1+2i`
- The `#i` prefix should explicitly mark the result as inexact

## Suggested fix

Apply `applyExactness` to complex results, or build exact complex when `exactness == .exact`. For example, wrap the complex return values:

```zig
const complex_val = gc.allocComplex(real_val, imag_val) catch return PrimitiveError.OutOfMemory;
return applyExactness(gc, complex_val, exactness);
```

Note: `applyExactness` may also need to handle complex values in its `.exact` branch.

baijum:
**Reopening: the fix landed but its target is a no-op for this type.** Found in Phase 1A of the v2 audit (#1917, #1890).

The original repro still fails at v0.22.1:

```scheme
(exact? (string->number "#e1+2i"))   ⟹ #f
```

I initially read this as "never fixed", but the truth is more useful. The fix **did** land — all four `allocComplex` sites were wired through `applyExactness`. But `applyExactness` has **no Complex arm in either direction**: `primitives_numeric.zig:1000` and `:1008` both fall through to a bare `return val`. The plumbing was connected to a function that does nothing for complex numbers.

Two related symptoms share the cause, detailed in #1910:
[...]

Issue #604, opened 2026-06-30
[primitives_numeric.zig] (string->number "#e<large>") aborts: unchecked @intFromFloat to i64 in applyExactness


baijum:
## Summary

`(string->number "#e<big>")` for an integer-valued flonum whose magnitude exceeds `i64` range (~9.22e18) **aborts the process** (ReleaseSafe panic: "integer part of floating point value out of bounds"). The crash is in the `.exact` branch of `applyExactness` in `src/primitives_numeric.zig`, which does an unchecked `@intFromFloat` into `i64`.

This is the same defect class as #79 Bug B, but in a **different file / code path**. #79 fixed the reader's `applyExactness` (`src/reader_tokens.zig:307`, which now guards `if (f > max_i64_f or f < min_i64_f) break :blk Token{ .flonum = f }`). The *separate* `applyExactness` used by the `string->number` **procedure** (`src/primitives_numeric.zig:894`) was not fixed and still panics.

## File/location

`src/primitives_numeric.zig`, `applyExactness`, `.exact` branch:

```zig
.exact => {
    if (types.isFlonum(val)) {
        const f = types.toFlonum(val);
        if (std.math.isNan(f) or std.math.isInf(f)) return types.FALSE;
        const trunc = @trunc(f);
        if (f == trunc) {
            const n: i64 = @intFromFloat(trunc);   // line ~923 — panics when |f| > i64 max
            return arith.makeFixnumChecked(n);
        }
        ...
```

(line numbers ~917-925). Reached from `stringToNumber` at line 1031: `return applyExactness(gc, types.makeFlonum(f), exactness);`.

The non-integer rational path just below (lines ~929-938: `@intFromFloat(num)` / `@intFromFloat(den)` into i64) is the same i64-overflow hazard for large non-integer values.

## Description

The guard checks only NaN/Inf, not whether `|f|` fits in `i64`. A finite integer-valued flonum with `|f| > maxInt(i64)` (~9.223e18) overflows the checked `@intFromFloat(i64)` cast → process abort. The boundary: `#e9e18` works, `#e1e19` / `#e9.5e18` panic.

## Impact

Process abort (DoS) on a value that should parse cleanly. `string->number` is a normal Scheme procedure, so a program parsing untrusted numeric text can be crashed. The correct result for `(string->number "#e1e20")` is the exact integer `100000000000000000000`.

Note the `exact` procedure itself (`exactFn`) was fixed (#28) and handles these correctly:
```scheme
[...]

PR #2181, opened 2026-08-02
Route reader #e/#i through string->number's digit-exact parser (#1911 family)

### LINKED ISSUE #1891: [R7RS 7.1.1] #e drops exactness on i64 overflow: #e1e19 is inexact, #e1e400 is +inf.0
**Category:** R7RS conformance · **Severity:** wrong-result · **Found by:** systematic audit v2 reconnaissance (#1890)

R7RS §7.1.1 makes `#e` a request for an *exact* number. Kaappi honours it only while the mantissa fits in an `i64`; past that the prefix is silently discarded.

## Reproduction

```scheme
(exact? #e1e18)   ;; expected #t   actual #t     <-- fine
(exact? #e1e19)   ;; expected #t   actual #f     <-- BUG
#e1e400           ;; expected 10^400 (bignum)    actual +inf.0   <-- BUG
```

Verbatim on a fresh ReleaseSafe build at `74efe596`, v0.22.1, cache cleared:

```text
(#t #f +inf.0 #t)
```

(the four values are `(exact? #e1e18)`, `(exact? #e1e19)`, `#e1e400`, `(exact? #e1+2i)`)

## Discriminating control

The cliff is exactly the `i64` range, not the magnitude of the literal: `#e1e18` is exact, `#e1e19` is not. And `(exact? #e1+2i)` ⟹ `#t`, so the complex path — the subject of #751 and #296 — is unaffected and this is a distinct defect.

`#e1e400` is the worse form of the same bug: an *exactness* prefix producing an **infinity**.

## Expected vs actual

| Input | Expected | Actual |
|---|---|---|
| `#e1e19` | exact `10000000000000000000` | inexact `1e19` |
| `#e1e400` | exact `10^400` (bignum) | `+inf.0` |
| `#e1.5e20` | exact | inexact |
| `#e#d1e20` | exact | inexact |

## Source location

`src/reader_tokens.zig` — the integer parse at ~`:214`/`:794` goes through `std.fmt.parseInt` into an `i64`; its `error.Overflow` arm falls back to `parseFloat` at ~`:447-452` **without consulting the `#e` prefix**, and never takes the bignum path.

The fix should route the overflow case to `bignum.parseBignumString` (and to a rational for the `#e1.5e20` shape) when `exact0` is set.

## Notes

Related but not duplicates: #751 (`string->number` ignored `#e`/`#i` for complex) and #296 (`#e` on complex, `#i` on bignums) — both closed, both a different path, as the control above shows. Worth checking whether `string->number` shares this decimal-exponent defect; it shares other reader helpers.

**Verified:** fresh ReleaseSafe build at `74efe596`, `kaappi cache clear` beforehand, reproduced twice.

### LINKED ISSUE #1907: Reader panics (exit 134) on #e with a decimal that rounds to exactly 2^63 — also aborts check, fmt, and ast
**Category:** Reader · **Severity:** crash · **Found by:** systematic audit v2, Phase 1A (#1890)

The reader aborts the process — not a catchable Scheme error, a Zig panic with exit 134 — on an `#e`-prefixed decimal that rounds to exactly 2^63.

## Reproduction

```scheme
(read (open-input-string "#e9223372036854775808.0"))
```

```text
kaappi internal error — this is a bug in kaappi, not in your program.
  version: v0.22.1 (aarch64-macos, ReleaseSafe)
thread 28178826 panic: integer part of floating point value out of bounds
src/reader_tokens.zig:435:17: applyExactness
src/reader_tokens.zig:482:26: readNumberPrefixed
src/reader_tokens.zig:546:38: readHash
```

Exit 134. Deterministic, 10/10 runs.

## It aborts tools that never execute the code

This is what makes it more than a curiosity. With `(define x #e9223372036854775808.0)` in a source file:

| command | exit |
|---|---|
| `kaappi check file.scm` | **134 (panic)** |
| `kaappi ast file.scm` | **134 (panic)** |
| `kaappi fmt --check file.scm` | **panic** |
| `kaappi fmt file.scm` | **panic** |

`check` is compile-only static analysis and `fmt` is a formatter — neither runs the program, and a linter or CI formatting gate that meets this literal dies rather than reporting.

## Discriminating controls

All four exit 0, which localises the fault precisely:

| input | exit | note |
|---|---|---|
| `#i9223372036854775808.0` | 0 | inexact prefix — fine |
| `#e9223372036854775808` | 0 | no `.0` — integer path, fine |
| `#e-9223372036854775808.0` | 0 | **negative side fine** |
| `#e9223372036854774784.0` | 0 | just below the window |
| `#e9223372036854775296.0` | **134** | first value in the window |

## Mechanism (inferred from source, corroborated by the exact window)

`reader_tokens.zig:400-402` guards with `f > max_i64_f` where
`max_i64_f = @floatFromInt(maxInt(i64))`. `maxInt(i64)` = 2^63−1 is **not**
representable as an f64, so that conversion rounds **up** to 2^63 — and the
guard therefore admits `f == 2^63`, which then overflows `@intFromFloat`.
`minInt(i64)` = −2^63 *is* exactly representable, which is exactly why the
negative side is safe. The crash window is precisely the doubles that round to
2^63, matching this reading.

**This is the guard added by #79 to fix this very panic. It is off by one value.**
A regression test pinning `maxInt(i64)` alone would not have caught it; the test
needs the first *representable double at or above* the boundary.

## Suggested fix

Compare against `9223372036854775808.0` (2^63) with `>=`, or convert via
`std.math.lossyCast`/an explicit range check that does not round the bound.
Pin both `9223372036854775296.0` (first crashing value) and
`9223372036854774784.0` (last safe value) in `src/tests_reader*.zig`, and add a
`kaappi check` case so the static-tool path is covered too.

## Related

Part of a family of reader-exactness defects sharing one root cause — see the
umbrella issue for the duplicated `applyExactness` implementations. #1891 is a
different defect on the same prefix; note that #1891's stated source location
points at the `#i` bignum path, not this one.

**Verified:** fresh ReleaseSafe build at `12588782`, isolated `KAAPPI_HOME`, reproduced twice plus all five controls.

### LINKED ISSUE #1908: #i on a radix-prefixed bignum reinterprets the digits as decimal — silent wrong answer

### LINKED ISSUE #1909: #e on a decimal below 1e-15 collapses to exact 0 — the entire value, including its sign, is lost

### LINKED ISSUE #1910: applyExactness has no complex arm: #i1+2i is a no-op, #e1e19+1i writes an unreadable 0/0 (reopens #751's symptom)

### LINKED ISSUE #1911: Two divergent applyExactness implementations: the reader un-rounds an f64, string->number rebuilds from digits

### LINKED ISSUE #1921: string->number returns #f for the unprefixed decimal 9223372036854775808.0, which the reader parses correctly
**Category:** Numeric tower · **Severity:** wrong-result · **Found by:** systematic audit v2, Phase 1A review (#1890, #1917)

`string->number` rejects a valid R7RS decimal — the plain, unprefixed spelling of 2^63 — while the reader accepts the identical text.

## Reproduction

```scheme
(string->number "9223372036854775808.0")            ⟹ #f
(read (open-input-string "9223372036854775808.0"))  ⟹ 9223372036854776000.0
```

## Discriminating controls — all four agree between the two parsers

| input | `string->number` | reader |
|---|---|---|
| `9223372036854775807.0` | correct | correct |
| **`9223372036854775808.0`** | **`#f`** | `9223372036854776000.0` |
| `-9223372036854775808.0` | correct | correct |
| `1e19` | correct | correct |
| `1e300` | correct | correct |

So it is not "large decimals" — it is precisely the positive decimal spelling of
2^63, with the negative side fine.

## Why it is worth filing separately

This is the **same 2^63 boundary** as #1907 and shows the same
negative-side-is-fine asymmetry (−2^63 is exactly representable as an f64;
2^63−1 is not, so `@floatFromInt(maxInt(i64))` rounds up). But:

- **No exactness prefix is involved**, so it is not an `applyExactness` defect.
  It is the plain decimal parse path.
- **The parity inverts.** In #1908, #1909, #1911 and the inf/nan gap, the reader
  is the wrong side and `string->number` is correct. Here it is the reverse.

That inversion strengthens #1911's case rather than weakening it: the two number
parsers do not merely differ in strength, they each have boundary handling the
other lacks. Unifying them has to take the union, not just adopt one side.

## How it was found

A review comment on #1917 flagged a test control asserting
`(not (string->number "#e9223372036854775808.0"))` as pinning non-conforming
behaviour. Investigating that showed the *unprefixed* form fails too — the
control had been masking a defect one layer down. The control has been rewritten
to assert only the non-crash property it exists to establish.

A disabled parity assertion in
`tests/scheme/compliance/reader-exactness-gaps.scm` is marked `;; FAIL: #1919`
and re-enables when this is fixed, alongside four enabled controls.

**Verified:** fresh ReleaseSafe build at `12588782`, isolated `KAAPPI_HOME`, reproduced twice with all four controls.

baijum:
Unify the two number parsers behind `#e`/`#i`: the reader now routes exactness through `string->number`'s digit-exact `parseNumberText`, deleting its own f64-unrounding `applyExactness`. One structural change closes the whole audit family.

Closes #1911, closes #1891, closes #1907, closes #1908, closes #1909, closes #1910, closes #1921.

## The root cause (#1911)

`applyExactness` existed twice with structurally different strategies: `string->number` rebuilt exact values from the decimal digits (mantissa × 10^scale via bignum), while the reader parsed to **f64 first** and tried to un-round it with an i64 continued fraction under an absolute 1e-15 tolerance. Every past fix (#79, #419, #604, #751, #1891) landed in one copy. The weaker algorithm is why the reader lost at both ends of the range — and R7RS §6.2.7 explicitly requires `read` and `string->number` to follow the same rules.

## What changed

**`primitives_numeric.zig`** — `stringToNumber`'s body is extracted into pub `parseNumberText(gc, text, radix, exactness)`; `string->number` is now a thin argument-unwrapping wrapper over it.

**Reader** — `readNumberPrefixed` still tokenizes exactly as before (token boundaries and incremental-input `UnexpectedEof` semantics are untouched), but when an exactness prefix is present the body span becomes a new `Token.prefixed_real{str, exact, radix}`, and `tokenToValue` re-parses that span through `parseNumberText`. The reader's 65-line token-level `applyExactness` is **deleted**. The two parsers cannot diverge on `#e`/`#i` again by construction.

`.complex` tokens are the one deliberate exception: `readNumber`'s complex grammar is wider than `string->number`'s (rational real/imag parts like `1/2+3/4i`, inf/nan parts like `3.0+inf.0i`), so re-parsing the span would reject valid literals. Exactness on a complex is exactly its two flags, set/cleared at token level, with `#e` refusing non-finite parts — the same rule the `parseNumberText` path applies.

## Per-issue effect

| issue | before | after |
|---|---|---|
| #1891 | `#e1e19` inexact, `#e1e400` ⟹ `+inf.0` | exact `10^19`, exact `10^400` |
| #1907 | **panic, exit 134** on `#e9223372036854775808.0` (also aborted `check`/`fmt`/`ast`) | exact `2^63`; all four tools exit 0 |
| #1908 | `#i#x1000000000000000000` ⟹ `1e18` (digits read as decimal) | `4.722366482869645e21`, identical to `string->number` (same limb walk) |
| #1909 | `#e1e-16` ⟹ exact `0`, sign lost; `#e1e-15` off-by-one denominator | full exact value at any magnitude; `1/10^15` exactly |
| #1910 | `#i1+2i` a no-op; `(exact? (string->number "#e1+2i"))` ⟹ `#f`; `#e1e19+1i` wrote unreadable `0/0+1i` | both `applyExactness` sides have Complex arms; `#e1e19+1i` writes `10000000000000000000+1i` and round-trips |
| #1921 | `string->number "9223372036854775808.0"` ⟹ `#f` | the flonum the reader gives — parseInt's `Overflow` no longer masks a valid decimal (`InvalidCharacter` from the bignum fallback now falls through to the decimal shapes) |
| inf/nan parity gap (#1911 body) | reader: `#e+inf.0` ⟹ `+inf.0`; `string->number` ⟹ `#f` | reader raises a read error (#419 settled `#f` as the intended answer) |

## The printer half of #1910

`formatComplexPart` converted exact-flagged components through i64 and rendered anything beyond as `0/0` — and its small-rational search silently printed exact-flagged tiny values as `0`. It is now writer-based `writeComplexPart`: integral values print exact digits at any magnitude (a fixed-buffer double-and-carry decimal printer; the i64 window bound is spelled as the literal `9223372036854775808.0` — `@floatFromInt(maxInt(i64))` rounds up, which was exactly #1907's off-by-one), the small-rational form is used only when it validates against the f64, and everything else prints the f64's own exact `m/2^k` value. Verified `string=?` against an independent oracle: `(number->string (exact f))`.

Integral and i64-rational spellings read back with the exact flag intact. The `m/2^k` fallback for tiny components (e.g. `#e1e-300+1i`) prints the correct exact value but does **not** read back yet — the reader's complex grammar stops at i64 rational parts (#2182), and extending it must wait for the scaled rational→f64 conversion (#2183) or the form would read back as a silently wrong `0.0+1i`. A loud read error beats the old silent collapse to `0`; the gap is pinned in both directions (written form matches the `exact` oracle; read-back is asserted to error, failing loudly when #2182/#2183 land so the pin gets flipped to a round-trip assertion). Surfaced by review.

## Semantics note: one corrected test expectation

The disabled cell for `#e9.223372036854776e18` expected `9223372036854775808` — the un-round-the-f64 answer. Digit-exact semantics (and `string->number`, unchanged here) give `9223372036854776000`: the literal denotes its own decimal value; that is what `#e` makes exact. The re-enabled cell asserts the digit-exact value plus explicit parity with `string->number`.

Similarly `#e9223372036854775296.0` is now `9223372036854775296` (its own digits), where the old expectation encoded the rounding to `2^63`.

## Deliberately unchanged

- **#1892 behavior** (trailing junk after a prefixed number reads as two datums) — closed NOT_PLANNED; the tokenize-first design preserves it, and section 9 of the parity test stays disabled as documentation.
- **Bare (unprefixed) complex exactness and the complex part grammar**: `1+2i` is exact via the reader but inexact via `string->number`; `string->number` cannot parse rational complex parts (`"1/2+3/4i"` ⟹ `#f`) or inf/nan parts; the reader's own rational parts stop at i64. Filed as #2182, sequenced after #2183 (the rational→f64 conversion bug that would otherwise turn the tiny-component read error into a silent `0.0`).
- #751 stays closed: its repro line is now a passing regression test here (#1910 covers the reopen recommendation).

## Tests

- `src/tests_numeric.zig`: 7 new tests pinning every issue, including the #1907 crash window as source literals (a regression aborts the run loudly), the corrected digit-exact expectations, write/round-trip of `#e1e19+1i`, and the two-direction pin of the tiny-exact-component gap (written form matches the `exact` oracle; read-back errors until #2182/#2183, then fails loudly to force the flip to a round-trip assertion).
- `src/tests_reader_incremental.zig`: the prefixed inf/nan scanner-path representative is now `#i+inf.0` (`#e+inf.0` is deliberately an error).
- `tests/scheme/compliance/reader-exactness-gaps.scm`: all 39 disabled cells of the #1911 family re-enabled (181 assertions pass), banners rewritten as regression-guard documentation, round-trip matrix extended with the previously-crashing and previously-wrong cells.
- Full unit suite, gc-stress on the touched paths, and `tests/scheme/run-all.sh` green; `kaappi check`/`ast`/`fmt`/`fmt --check` verified on the #1907 literal.

### DIFF
--- a/src/printer.zig
+++ b/src/printer.zig
@@ -368,10 +368,15 @@ fn writeBigPowerOfTwoDecimal(writer: anytype, mantissa: u64, shift: u16) !void {
 }
 
 /// Write one part of a complex number. Inexact parts print as flonums. An
-/// exact-flagged part prints a spelling that reads back to the same f64 with
-/// its exact flag: plain integer digits when integral (bignum-wide past i64),
-/// the small-rational form when it reproduces the f64 exactly, and the f64's
-/// own mantissa/2^k value otherwise -- never a value-destroying collapse.
+/// exact-flagged part prints its exact value, never a value-destroying
+/// collapse: plain integer digits when integral (bignum-wide past i64), the
+/// small-rational form when it reproduces the f64 exactly, and the f64's own
+/// mantissa/2^k value otherwise. The first two spellings read back with the
+/// exact flag intact; the mantissa/2^k fallback is exact but does NOT read
+/// back yet -- the reader's complex grammar stops at i64 rational parts
+/// (kaappi#2182), and closing that needs the scaled rational->f64
+/// conversion first or tiny components would read back as a silent 0.0
+/// (kaappi#2183).
 fn writeComplexPart(writer: anytype, f: f64, exact: bool) !void {
     var buf: [64]u8 = undefined;
     if (!exact or std.math.isNan(f) or std.math.isInf(f)) {
--- a/src/printer.zig
+++ b/src/printer.zig
@@ -328,28 +328,101 @@ fn symbolNeedsBars(name: []const u8) bool {
     return false;
 }
 
-/// Format a flonum in Scheme syntax into `buf`, returning the slice. Uses
-/// scientific notation for very large/small magnitudes so the output stays
-/// bounded (plain `{d}` expands denormals/huge values to hundreds of decimal
-/// digits, overflowing fixed buffers). Always includes a `.`/`e` so the result
-/// reads back as inexact.
-fn formatComplexPart(buf: []u8, f: f64, exact: bool) []const u8 {
-    if (std.math.isNan(f) or std.math.isInf(f)) return formatFlonum(buf, f);
-    if (exact) {
-        const trunc = @trunc(f);
-        if (f == trunc and @abs(f) < 4.5e18) {
-            const i: i64 = @intFromFloat(trunc);
-            return std.fmt.bufPrint(buf, "{d}", .{i}) catch return formatFlonum(buf, f);
-        }
-        // Exact non-integer: try rational notation
-        const numeric = @import("primitives_numeric.zig");
-        const rat = numeric.floatToRational(f);
-        if (rat.den != 1) {
-            return std.fmt.bufPrint(buf, "{d}/{d}", .{ rat.num, rat.den }) catch return formatFlonum(buf, f);
-        }
-        return std.fmt.bufPrint(buf, "{d}", .{rat.num}) catch return formatFlonum(buf, f);
+/// Write the exact decimal digits of `mantissa << shift` (unsigned). Any
+/// finite f64 decomposes into such a pair (mantissa < 2^53, shift <= 971 for
+/// integers; 2^k denominators use mantissa 1, k <= 1074), so this covers up
+/// to ~325 digits -- far past any fixed format buffer, which is why exact
+/// complex parts beyond i64 used to print as the unreadable `0/0` (#1910).
+fn writeBigPowerOfTwoDecimal(writer: anytype, mantissa: u64, shift: u16) !void {
+    // Callers pass f64 decompositions: mantissa < 2^53 with shift <= 971
+    // (integers), or mantissa 1 with shift <= 1074 (2^k denominators).
+    // 20 digits of u64 + 1100 doublings adding at most one digit each
+    // stays under 360.
+    var digits: [360]u8 = undefined; // little-endian, values 0..9
+    var len: usize = 0;
+    var m = mantissa;
+    while (m > 0) : (m /= 10) {
+        digits[len] = @intCast(m % 10);
+        len += 1;
     }
-    return formatFlonum(buf, f);
+    if (len == 0) {
+        digits[0] = 0;
+        len = 1;
+    }
+    var i: u16 = 0;
+    while (i < shift) : (i += 1) {
+        var carry: u8 = 0;
+        for (digits[0..len]) |*d| {
+            const v = d.* * 2 + carry;
+            d.* = v % 10;
+            carry = v / 10;
+        }
+        if (carry > 0 and len < digits.len) {
+            digits[len] = carry;
+            len += 1;
+        }
+    }
+    var out: [360]u8 = undefined;
+    for (0..len) |j| out[j] = '0' + digits[len - 1 - j];
+    try writer.writeAll(out[0..len]);
+}
+
+/// Write one part of a complex number. Inexact parts print as flonums. An
+/// exact-flagged part prints a spelling that reads back to the same f64 with
+/// its exact flag: plain integer digits when integral (bignum-wide past i64),
+/// the small-rational form when it reproduces the f64 exactly, and the f64's
+/// own mantissa/2^k value otherwise -- never a value-destroying collapse.
+fn writeComplexPart(writer: anytype, f: f64, exact: bool) !void {
+    var buf: [64]u8 = undefined;
+    if (!exact or std.math.isNan(f) or std.math.isInf(f)) {
+        try writer.writeAll(formatFlonum(&buf, f));
+        return;
+    }
+    if (f == @trunc(f)) {
+        // 2^63 spelled as a literal: maxInt(i64) is not representable as
+        // f64 and @floatFromInt would round the bound up to admit 2^63
+        // itself -- the exact off-by-one behind #1907's reader panic.
+        if (f >= -9223372036854775808.0 and f < 9223372036854775808.0) {
+            const i: i64 = @intFromFloat(f);
+            try writer.writeAll(std.fmt.bufPrint(&buf, "{d}", .{i}) catch unreachable);
+            return;
+        }
+        // Integral beyond i64: |f| >= 2^63 with a 53-bit mantissa forces a
+        // strictly positive binary exponent.
+        if (f < 0) try writer.writeByte('-');
+        const bits: u64 = @bitCast(@abs(f));
+        const mantissa = (bits & 0x000FFFFFFFFFFFFF) | 0x0010000000000000;
+        const exp: u16 = @intCast(@as(i16, @intCast((bits >> 52) & 0x7FF)) - 1023 - 52);
+        try writeBigPowerOfTwoDecimal(writer, mantissa, exp);
+        return;
+    }
+    // Exact non-integer: prefer the small-rational spelling, but only when
+    // it reads back to this exact f64 -- the search collapses magnitudes
+    // below its granularity to 0/1, which silently destroys the value.
+    const numeric = @import("primitives_numeric.zig");
+    const rat = numeric.floatToRational(f);
+    if (rat.den > 0 and
+        @as(f64, @floatFromInt(rat.num)) / @as(f64, @floatFromInt(rat.den)) == f)
+    {
+        try writer.print("{d}/{d}", .{ rat.num, rat.den });
+        return;
+    }
+    // Fall back to the f64's own exact value: odd-mantissa / 2^k.
+    if (f < 0) try writer.writeByte('-');
+    const bits: u64 = @bitCast(@abs(f));
+    const raw_exp: u11 = @intCast((bits >> 52) & 0x7FF);
+    var m: u64 = if (raw_exp == 0)
+        bits & 0x000FFFFFFFFFFFFF
+    else
+        (bits & 0x000FFFFFFFFFFFFF) | 0x0010000000000000;
+    var k: u16 = if (raw_exp == 0) 1074 else 1075 - @as(u16, raw_exp);
+    while (m & 1 == 0 and k > 0) {
+        m >>= 1;
+        k -= 1;
+    }
+    try writeBigPowerOfTwoDecimal(writer, m, 0);
+    try writer.writeByte('/');
+    try writeBigPowerOfTwoDecimal(writer, 1, k);
 }
 
 pub fn formatFlonum(buf: []u8, f: f64) []const u8 {
@@ -655,7 +728,7 @@ fn printValueWithDepth(writer: anytype, value: Value, mode: PrintMode, depth: u3
                     try writer.writeAll(formatFlonum(&buf, c.real));
                 } else {
                     const has_real = c.real != 0.0 or std.math.signbit(c.real);
-                    if (has_real) try writer.writeAll(formatComplexPart(&buf, c.real, c.exact_real));
+                    if (has_real) try writeComplexPart(writer, c.real, c.exact_real);
                     const im = c.imag;
                     if (std.math.isNan(im)) {
                         try writer.writeAll("+nan.0i");
@@ -664,7 +737,7 @@ fn printValueWithDepth(writer: anytype, value: Value, mode: PrintMode, depth: u3
                     } else {
                         try writer.writeByte(if (im < 0 or std.math.signbit(im)) '-' else '+');
                         const mag = @abs(im);
-                        if (mag != 1.0 or has_real) try writer.writeAll(formatComplexPart(&buf, mag, c.exact_imag));
+                        if (mag != 1.0 or has_real) try writeComplexPart(writer, mag, c.exact_imag);
                         try writer.writeByte('i');
                     }
                 }
--- a/src/reader_datum.zig
+++ b/src/reader_datum.zig
@@ -65,6 +65,12 @@ fn tokenToValue(self: *Reader, tok: Token) ReadError!Value {
         .prefixed_real => |p| {
             // #e/#i literal body: same parser as string->number, so the two
             // can never disagree on an exactness-prefixed number (#1911).
+            // Non-OOM errors all become InvalidNumber deliberately: in a
+            // literal context every parse-side failure IS a bad number
+            // literal, and the tokenizer has already validated the shape.
+            // If parseNumberText ever grows a failure that is not the
+            // text's fault, give it its own arm here rather than letting
+            // it masquerade as a user typo.
             const numeric = @import("primitives_numeric.zig");
             const v = numeric.parseNumberText(self.gc, p.str, p.radix, if (p.exact) .exact else .inexact) catch |err| switch (err) {
                 error.OutOfMemory => return ReadError.OutOfMemory,
--- a/src/reader_datum.zig
+++ b/src/reader_datum.zig
@@ -62,6 +62,17 @@ fn tokenToValue(self: *Reader, tok: Token) ReadError!Value {
             if (bignum_mod.isZero(den)) return ReadError.InvalidNumber;
             return arith.makeRationalReduced(self.gc, slot_num.get(), den) catch return ReadError.OutOfMemory;
         },
+        .prefixed_real => |p| {
+            // #e/#i literal body: same parser as string->number, so the two
+            // can never disagree on an exactness-prefixed number (#1911).
+            const numeric = @import("primitives_numeric.zig");
+            const v = numeric.parseNumberText(self.gc, p.str, p.radix, if (p.exact) .exact else .inexact) catch |err| switch (err) {
+                error.OutOfMemory => return ReadError.OutOfMemory,
+                else => return ReadError.InvalidNumber,
+            };
+            if (v == types.FALSE) return ReadError.InvalidNumber;
+            return v;
+        },
         .complex => |c| return self.gc.allocComplexEx(c.real, c.imag, c.exact_real, c.exact_imag) catch return ReadError.OutOfMemory,
         .boolean => |b| return if (b) types.TRUE else types.FALSE,
         .character => |c| return types.makeChar(c),
--- a/src/tests_numeric.zig
+++ b/src/tests_numeric.zig
@@ -524,10 +524,23 @@ test "#e/#i reach complex literals on both parsers (#1910, #751)" {
     );
     // An exact-flagged component below the rational search's granularity
     // used to print as 0 (value destroyed); it now prints its exact
-    // mantissa/2^k value.
-    try th.expectEvalBool(
-        "(let ((p (open-output-string))) (write #e1e-300+1i p) (string-prefix? \"0+\" (get-output-string p)))",
-        false,
+    // mantissa/2^k value, verified against the independent bignum printer
+    // behind (exact f).
+    try th.expectEvalTrue(
+        "(let ((p (open-output-string)) (ex (exact 1e-300))) (write #e1e-300+1i p)" ++
+            " (equal? (get-output-string p) (string-append (number->string (numerator ex))" ++
+            " \"/\" (number->string (denominator ex)) \"+1i\")))",
+    );
+    // KNOWN GAP, pinned: that spelling does not read back yet. The reader's
+    // complex grammar stops at i64 rational parts (kaappi#2182), and closing
+    // it needs the scaled rational->f64 conversion first or this would read
+    // back as a silent 0.0+1i (kaappi#2183). A loud error still beats the
+    // old silently-wrong 0. When those land, the read succeeds, this
+    // deliberately returns #f, and the failure tells you to replace it with
+    // a round-trip equal? assertion.
+    try th.expectEvalTrue(
+        "(let ((p (open-output-string))) (write #e1e-300+1i p)" ++
+            " (guard (e (#t #t)) (read (open-input-string (get-output-string p))) #f))",
     );
     // #e refuses non-finite components, like the flonum rule (#419).
     try th.expectEvalTrue("(guard (e (#t #t)) (read (open-input-string \"#e1e999+2i\")) #f)");
--- a/src/tests_numeric.zig
+++ b/src/tests_numeric.zig
@@ -452,3 +452,108 @@ test "eqv? discriminates exact and inexact complex (#2167)" {
         "(eq? 'hit (case (- (make-rectangular 3/2 1)) ((-3/2-1i) 'hit) (else 'miss)))",
     );
 }
+
+// ---------------------------------------------------------------------------
+// #1911 family: the reader's #e/#i path now routes through the same
+// digit-exact parseNumberText as string->number, replacing a token-level
+// f64-unrounding conversion that failed at both ends of the range.
+// ---------------------------------------------------------------------------
+
+test "#e keeps exactness past i64 (#1891)" {
+    try th.expectEvalBool("(exact? #e1e19)", true);
+    try th.expectEvalTrue("(= #e1e19 (expt 10 19))");
+    try th.expectEvalBool("(exact? #e1e400)", true);
+    try th.expectEvalTrue("(= #e1e400 (expt 10 400))");
+    try th.expectEvalTrue("(= #e-1e19 (- (expt 10 19)))");
+    try th.expectEvalTrue("(= #e1.5e20 (* 15 (expt 10 19)))");
+    try th.expectEvalBool("(exact? #e#d1e20)", true);
+}
+
+test "#e decimal at the 2^63 boundary parses exactly, no panic (#1907)" {
+    // Every double rounding to exactly 2^63 used to overflow @intFromFloat
+    // and abort the process (exit 134): maxInt(i64) is not representable as
+    // f64, so the old `f > @floatFromInt(maxInt(i64))` guard admitted 2^63
+    // itself. Source literals are deliberate — a regression aborts this
+    // test run loudly.
+    try th.expectEvalTrue("(= #e9223372036854775808.0 (expt 2 63))");
+    try th.expectEvalTrue("(= #e9223372036854775296.0 9223372036854775296)");
+    try th.expectEvalTrue("(= #e-9223372036854775808.0 (- (expt 2 63)))");
+    try th.expectEvalTrue("(= #e9223372036854774784.0 9223372036854774784)");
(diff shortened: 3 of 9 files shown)


src/printer.zig
```diff
@@ -328,28 +328,101 @@ fn symbolNeedsBars(name: []const u8) bool {
...
+        (bits & 0x000FFFFFFFFFFFFF) | 0x0010000000000000;
+    var k: u16 = if (raw_exp == 0) 1074 else 1075 - @as(u16, raw_exp);
+    while (m & 1 == 0 and k > 0) {
+        m >>= 1;
+        k -= 1;
     }
-    return formatFlonum(buf, f);
+    try writeBigPowerOfTwoDecimal(writer, m, 0);
+    try writer.writeByte('/');
+    try writeBigPowerOfTwoDecimal(writer, 1, k);
```
baijum, line 430: **High — exact complex with a bignum-denominator component doesn’t round-trip**

This fallback correctly avoids printing `0`, but when `k` is large enough that `2^k` overflows `i64`, the emitted `mantissa/2^k+Ni` form isn’t readable:

```scheme
(write #e1e-300+1i)
;; => <bignum>/<bignum>+1i
(read (open-input-string "…"))  ; => KP3000 invalid number
```

Root cause: `bigRationalToken` (in `readNumber`) rejects a trailing non-delimiter by design — “a bignum rational must end at a delimiter” — so a bignum-denominator rational inside a complex literal fails. This PR is the first producer of that shape.

**Suggestion:** extend complex-suffix detection after a bignum-denominator rational (mirroring the bignum-integer path), *or* document this as a known gap + file a follow-up (same pattern as #1892 here). Please don’t leave the “verified round-trip on every cell” claim standing for this case.

src/tests_numeric.zig
```diff
@@ -452,3 +452,108 @@ test "eqv? discriminates exact and inexact complex (#2167)" {
...
+    try th.expectEvalTrue(
+        "(let ((x #e1e19+1i) (p (open-output-string))) (write x p) (equal? x (read (open-input-string (get-output-string p)))))",
+    );
+    // An exact-flagged component below the rational search's granularity
+    // used to print as 0 (value destroyed); it now prints its exact
+    // mantissa/2^k value.
+    try th.expectEvalBool(
+        "(let ((p (open-output-string))) (write #e1e-300+1i p) (string-prefix? \"0+\" (get-output-string p)))",
+        false,
+    );
```
baijum, line 544: **High (test gap) — asserts “not `0+`” but never round-trips**

Unlike the `#e1e19+1i` case above (which does `write` → `read` → `equal?`), this only checks the print isn’t the old collapsed `0+…`. That misses the regression where the new spelling itself is unreadable:

```scheme
(let ((x #e1e-300+1i) (p (open-output-string)))
  (write x p)
  (equal? x (read (open-input-string (get-output-string p)))))
;; currently errors KP3000
```

Please add a round-trip assertion once the reader accepts bignum-rational + imag suffix (or `test-expect-fail` / document if deferred).

src/reader_datum.zig
```diff
@@ -62,6 +62,17 @@ fn tokenToValue(self: *Reader, tok: Token) ReadError!Value {
             if (bignum_mod.isZero(den)) return ReadError.InvalidNumber;
             return arith.makeRationalReduced(self.gc, slot_num.get(), den) catch return ReadError.OutOfMemory;
         },
+        .prefixed_real => |p| {
+            // #e/#i literal body: same parser as string->number, so the two
+            // can never disagree on an exactness-prefixed number (#1911).
+            const numeric = @import("primitives_numeric.zig");
+            const v = numeric.parseNumberText(self.gc, p.str, p.radix, if (p.exact) .exact else .inexact) catch |err| switch (err) {
+                error.OutOfMemory => return ReadError.OutOfMemory,
+                else => return ReadError.InvalidNumber,
```
baijum, line 77: **Low — broad `else` maps non-OOM errors to `InvalidNumber`**

Any `PrimitiveError` other than `OutOfMemory` becomes a catchable invalid-number read error. Unlikely to matter today (tokenizer already validated the text; matches the neighbouring `.bignum_str` style), but worth keeping an eye on if `parseNumberText` gains new fallible paths that shouldn’t look like bad literals.

baijum (commented):
## Review summary

**Verdict: request changes** *(posted as COMMENT — GitHub disallows REQUEST_CHANGES on your own PR).*

The core refactor — routing reader `#e`/`#i` through `parseNumberText` — is the right fix and the seven claimed issues (#1891/#1907/#1908/#1909/#1910/#1911/#1921) look genuinely fixed. Docs, GC stress, and the 2⁶³ boundary handling are in good shape.

However, the new exact-complex printer path introduces a **real write→read round-trip regression** for small-magnitude exact components (e.g. `#e1e-300+1i`). Please either fix the reader’s bignum-rational + imag-suffix path, or document it as a known gap (same pattern as #1892 in this PR) and file a follow-up — the current “round-trips” claim doesn’t hold for that shape.

### Positives
- Single `parseNumberText` eliminates `read`/`string->number` divergence by construction
- Careful `2^63` literal bound (avoids `@floatFromInt(maxInt(i64))` round-up)
- Out-of-scope items (#1892, bare complex exactness) called out explicitly
- Unit + Scheme compliance coverage for the claimed fixes is thorough

src/printer.zig
baijum: Confirmed — and investigating the suggested fix found it is **not safe to land on its own**. The root-cause reading is right (`bigRationalToken`'s delimiter guard, this PR the first producer of the shape), but extending complex-suffix detection past bignum-denominator rationals would route the real part through the rational→f64 conversion, which collapses to `0.0` whenever the denominator alone overflows f64:

```scheme
(inexact (/ 1 (expt 2 1049)))   ⟹ 0.0   ;; expected ~1.68e-316, pre-existing
```

`#e1e-300+1i`'s exact denominator is 2^1049 (316 digits) — inside that zone. So the grammar extension alone would read the printed form back as a silently wrong `0.0+1i`, trading today's loud error for silent corruption. Filed the conversion bug as #2183 (three manifestations: finite/inf → 0.0, inf/finite → +inf.0, and a parser-vs-procedure inf/inf divergence) and extended #2182 with the reader-side grammar work, sequenced after it.

Took the documentation path you offered, in a5603e01: `writeComplexPart`'s doc comment, the compliance file's section-7 banner, and the section-10 "every cell" claim now state the gap precisely, and section 7 pins it in both directions — written form must equal the independent `(exact f)` oracle spelling, read-back must error. The overbroad claims are gone from the PR body as well.

src/tests_numeric.zig
baijum: Right — worse, a naive "assert the read errors OR works" pin would be a check that tests nothing (the campaign's own top anti-pattern). a5603e01 replaces the not-`0+` check with a two-direction pin:

1. The written form must `equal?` the independent oracle spelling — `(number->string (numerator (exact 1e-300)))` `/` denominator — so the print is verified correct, not merely non-zero.
2. The read-back is asserted to **error**, with the guard body returning `#f` on success — so the moment #2182/#2183 make the shape readable, this assertion fails loudly and its comment says to replace it with the round-trip `equal?`.

Same pin duplicated in `reader-exactness-gaps.scm` section 7 with the same flip-me instruction.

src/reader_datum.zig
baijum: Agreed it's the right worry to note. Kept the broad mapping deliberately — in a literal context every parse-side failure *is* a bad number literal, and enumerating `PrimitiveError` variants here would go stale as the set grows — but a5603e01 documents the intent at the arm with an explicit instruction: if `parseNumberText` ever gains a failure that isn't the text's fault, it gets its own arm rather than masquerading as a user typo. (Today's reachable non-OOM errors — div-by-zero from `1/0`-shaped input, the unreachable gc/vm-instance guards — all map correctly.)

baijum:
Review response landed in a5603e01. Summary of the critical look:

**Printer round-trip (High)** — confirmed, with one important amendment: the suggested reader-grammar extension is **unsafe to land alone**. The rational→f64 conversion collapses to `0.0` when the denominator alone overflows f64 (`(inexact (/ 1 (expt 2 1049)))` ⟹ `0.0` today, pre-existing), and `#e1e-300+1i`'s exact denominator is 2^1049 — so the extension by itself would turn the loud read error into a silently wrong `0.0+1i`. Filed the conversion bug as #2183 (three verified manifestations, including a parser-vs-procedure inf/inf divergence where only `inexactFn` has a fallback), extended #2182 with the reader-side grammar work sequenced after it, and took the offered documentation path: every overbroad "reads back" claim corrected (doc comment, section-7/-10 banners, PR body), gap pinned in both directions.

**Test gap (High)** — fixed with a two-direction pin: written form must equal the independent `(exact f)` oracle spelling, and read-back must error via a guard that returns `#f` on success — so when #2182/#2183 land, the assertion fails loudly and instructs its own replacement with the round-trip `equal?`. Mirrored in the compliance file.

**Broad `else` (Low)** — kept deliberately (in a literal context every parse failure is a bad literal; enumerating variants would go stale), documented at the arm with the instruction to give any future not-the-text's-fault failure its own arm.

Verification on a5603e01: unit suite 1686 pass / 0 fail, full Scheme suite 2066 pass / 0 fail.

Issue #2155, opened 2026-08-01
[Cache] The .sbc key has no target component — all 17 release binaries share one compiler hash, and cond-expand is baked into the bytecode


baijum:
**Category:** Tooling / tier-divergence
**Spec reference:** `docs/dev/cache.md` § "The key: source **and** compiler" — "An entry is reused only when **both** halves of its key match the current run: 1. **Source hash** … 2. **Compiler hash** — a hash of the release version string **and the git build id**."
**Severity:** wrong-result (latent — see "How reachable is this today" for the honest exposure)

## Summary

The `.sbc` key identifies *which compiler binary* produced an entry, but not
*which target it compiles for*. Two binaries built from the same commit with a
clean tree — say `aarch64-macos` and `x86_64-windows` — have the **same**
`compilerHash`, so each will happily load the other's bytecode. That would be
harmless if the bytecode were target-independent. It is not: `cond-expand` is
resolved at compile time against `types.platform_features`, and only the taken
branch survives into the `.sbc`.

Nothing in the format or the load gate can notice. This is not the byte-order
question (the codec *is* endian-clean — audit v2 Phase 7D pins that with golden
bytes); it is the same file being *decodable* on a target it was not compiled
for.

## Evidence

**1. The key takes no target input.** `bytecode_file.compilerHashFor(version_str,
build_id)` (`src/bytecode_file.zig:121`) hashes exactly those two strings.
`build_options.git_build_id` comes from `gitBuildId` (`build.zig:576`), which is
`git rev-parse --short HEAD` plus a `-dirty` flag — no `-Dtarget` involvement.
So a cross-compiled build and a native build of the same clean commit produce
byte-identical keys.

**2. The load gate checks nothing else.** `deserializeFromBuffer`
(`src/bytecode_file_read.zig:336`) accepts on magic, `VERSION`, source hash and
compiler hash. The header has no target, OS, or feature field at all
(`writeFunctionsToBuffer`, `src/bytecode_file_write.zig:267`); the two strings it
does carry (build id, source path) are informational and are skipped on load.

**3. `cond-expand` is baked in.** Measured, macOS aarch64, ReleaseSafe at
`910f7c75`, isolated `KAAPPI_HOME`:

```scheme
;; ce.scm
(define (which)
  (cond-expand (windows "WINDOWS-BRANCH") (posix "POSIX-BRANCH")))
(display (which))
(newline)
```

```console
$ kaappi ce.scm
POSIX-BRANCH
[...]

Issue #2113, opened 2026-08-01
[Cache] The writer emits entries the reader always rejects — a permanent miss that cache status reports as 'current'


baijum:
**Category:** Tooling
**Spec reference:** `docs/dev/cache.md` — "`cache status` marks each entry **current** (produced by the running binary, **so a plain run of its still-unchanged source would hit**) or **stale**".
**Severity:** edge-case (wasted work, forever) + doc-truth (`cache status` misreports it)

## Summary

The two halves of the `.sbc` codec do not enforce the same limits. The **writer**
enforces almost nothing; the **reader** enforces six caps. A program whose
constants exceed any reader-side cap therefore writes an entry that can **never**
be loaded: every run recompiles from source, rewrites the same unusable `.sbc`,
and pays the write cost on top of the compile cost — forever, silently.

`kaappi cache status` cannot see it either. It reads only the header
(`bytecode_file.readHeaderInfo`, `src/cache.zig:286`) and never attempts the body,
so it reports the entry as **current** — which `docs/dev/cache.md` defines as "a
plain run of its still-unchanged source would hit". For these entries that is
false. The one tool provided for inspecting the cache actively misreports it.

The asymmetry:

| limit | writer | reader |
|---|---|---|
| `MAX_CONSTANT_DEPTH` (256) | **truncates to `TAG_NIL`** (`bytecode_file_write.zig:134`) | rejects (`bytecode_file_read.zig:110`) |
| `MAX_STRING_BYTES` (1 MiB) | — | rejects (:130) |
| `MAX_VECTOR_LEN` (262 144) | — | rejects (:165) |
| `MAX_BYTEVECTOR_LEN` (1 MiB) | — | rejects (:180) |
| `MAX_BIGNUM_LIMBS` (262 144) | — | rejects (:186) |
| `MAX_CONSTANTS_PER_FUNCTION` (65 535) | — | rejects |
| `MAX_SYMBOL_BYTES` (4 096) | rejects (:193) | rejects (:124) |

Only the last row agrees. The reader being the stricter half is what keeps the
depth row from being a *silent wrong answer* — the truncated tail is never
observed, because the whole file is rejected first — so this is a resource and
observability defect rather than a correctness one. But it is one guard away from
being a correctness one, and nothing pins the relationship.

## Minimal reproduction

The cliff is exact. A quoted proper list of N elements reaches depth N in
`writeConstant` (one level per cdr step):

```console
[...]

Issue #2112, opened 2026-08-01
[Cache] A .sbc HIT leaves the macro table empty — a top-level define-syntax is invisible to run-time eval


baijum:
**Category:** Tooling
**Spec reference:** `docs/dev/cache.md` — "The cache is transparent to correctness — a miss just recompiles — but an *invisible* cache is a hazard: a stale entry that silently runs the wrong bytecode manufactures phantom bugs and masks real fixes."
**Severity:** wrong-result (exit code and stdout both change)

## Summary

`define-syntax` is a **compiler** form: `compileDefineSyntax`
(`src/compiler_define_syntax.zig`) registers the transformer in the live macro
table as a side effect of compiling the file, and emits no bytecode that would
re-register it. A cache HIT skips compilation entirely
(`src/main.zig:673`–`:740`), so on the warm run `vm.macros` never learns the name.

Any *run-time* consumer of the macro table then disagrees between the two runs.
`eval` is the reachable one: it compiles its argument at run time against that
table, so a top-level `define-syntax` that works on the cold run is an undefined
variable on the warm one.

## Minimal reproduction

```scheme
;; m.scm
(define-syntax dbl (syntax-rules () ((_ x) (* 2 x))))
(display (dbl 21))(newline)
(display (eval '(dbl 5) (interaction-environment)))(newline)
```

```console
$ export KAAPPI_HOME=$(mktemp -d)

$ kaappi m.scm                        # cold — cache MISS
42
10
$ echo $?
0

$ kaappi m.scm                        # warm — cache HIT
42
m.scm:3:25: error[KP3001]: undefined variable 'dbl'. Did you mean 'abs'?
    (display (eval '(dbl 5) (interaction-environment)))(newline)
$ echo $?
1
```

Same for a macro that expands into another macro, and for one used inside a
nested form:

```scheme
(define-syntax outer (syntax-rules () ((_ x) (inner x))))
(define-syntax inner (syntax-rules () ((_ x) (+ x 100))))
(display (eval '(outer 2) (interaction-environment)))(newline)
;; cold: 102     warm: undefined variable 'outer'

(display (eval '(let () (dbl 3)) (interaction-environment)))(newline)
;; cold: 6       warm: undefined variable 'dbl'
```

`define-property` (SRFI 213) has the same shape — it writes into the VM-owned
`syntax_properties` table at compile time — and is equally invisible after a HIT.

## Discriminating control

Two controls, both of which **agree** cold and warm:

[...]

Issue #2111, opened 2026-08-01
[Cache] A .sbc HIT unshares datum-label structure — eq? flips, shared DAGs are exponential, cyclic literals never load


baijum:
**Category:** Tooling / R7RS
**Spec reference:** R7RS §2.4 (Datum labels), p. 9 — "The lexical syntax `#⟨n⟩#` serves as a reference to some object labelled by `#⟨n⟩=`; **the result is the same object as the `#⟨n⟩=`** (see section 6.1). Together, these syntaxes permit the notation of structures with shared or circular substructure."
**Severity:** wrong-result

## Summary

`writeConstant` (`src/bytecode_file_write.zig:133`) is a plain recursive walk with
no visited-set, so a constant reached through two datum-label references is
emitted **once per reference**, and `readConstant`
(`src/bytecode_file_read.zig:109`) allocates a fresh object for each. A cache HIT
therefore answers `#f` where the cold run answers `#t`.

One root cause, three symptoms:

1. **Sharing is silently lost** — `eq?` flips across a cache HIT (below).
2. **A shared DAG makes the `.sbc` exponential in the source** — 241 source bytes
   produce 4.7 MB of cache, doubling per nesting level.
3. **A cyclic literal is a permanent cache miss** — the walk recurses until
   `writeConstant`'s depth-256 guard truncates, and `readConstant`'s matching
   guard then rejects the whole file (see #2113 for that mechanism).

## Minimal reproduction

```scheme
;; shared.scm
(define v '(#1=(1 2) #1#))
(display (eq? (car v) (cadr v)))(newline)
```

```console
$ export KAAPPI_HOME=$(mktemp -d)
$ kaappi shared.scm     # cold — cache MISS
#t
$ kaappi shared.scm     # warm — cache HIT
#f
```

Pairs, vectors and strings are all affected, at any nesting depth:

```scheme
(define v '#(#1=(a) #1# #1#))
(display (list (eq? (vector-ref v 0) (vector-ref v 1))
               (eq? (vector-ref v 1) (vector-ref v 2))))(newline)
;; cold: (#t #t)     warm: (#f #f)

(define v '(#1="abc" #1#))
(display (eq? (car v) (cadr v)))(newline)
;; cold: #t          warm: #f

(define v '(#1=(1 2) (x #1#) #(#1#)))
;; cold: both #t     warm: both #f
```

`write-shared` sees the same thing:

```scheme
(write-shared '(#1=(1 2) #1#))(newline)
[...]

PR #2188, opened 2026-08-02
Make the .sbc cache transparent: format v11, honest write gating, target-keyed compiler hash

### LINKED ISSUE #2187: [Cache] A file with a top-level compile error is cached — the warm run executes the partial program with exit 0 and the diagnostic gone
**Category:** Tooling / tier-divergence
**Spec reference:** `docs/dev/cache.md` — "The cache is transparent to correctness — a miss just recompiles — but an *invisible* cache is a hazard: a stale entry that silently runs the wrong bytecode manufactures phantom bugs and masks real fixes."
**Severity:** wrong-result (exit code flips, and the diagnostic vanishes)

## Summary

`runFile` reports a top-level compile error, skips the failed form, and keeps
going — so `compiled_funcs` holds only the forms that did compile. The cache
write was gated only on `has_imports`, so the **partial program** was written
as a perfectly valid entry. The warm run then HITs, executes the partial
program, prints no diagnostic, and exits 0.

## Minimal reproduction

```console
$ export KAAPPI_HOME=$(mktemp -d)
$ printf '(display "a")(newline)\n(if)\n(display "b")(newline)\n' > ce.scm

$ kaappi ce.scm                      # cold — cache MISS
a
ce.scm:2:1: compile error[KP2001]: invalid syntax
b
$ echo $?
1

$ kaappi ce.scm                      # warm — cache HIT
a
b
$ echo $?
0
```

The compile error *vanishing on the second run* is the worst shape of cache
opacity: it makes a broken program silently "work", and masks the diagnostic
exactly when someone reruns the file to look at it. Runtime errors were never
affected (the failing form is in the cached bytecode and fails again warm);
this is specific to forms that fail *compilation* and are therefore absent
from what gets cached.

## Discriminating control

The same file with the bad form removed agrees cold/warm; and with the bad
form present, `kaappi cache clear` or `--no-ir-opt` restores the cold
behaviour — the cache is the responsible tier.

**Found:** while reworking the cache-write gating for #2112 (the
`define-syntax` refusal), by reading the write condition — the reproduction
above was then confirmed on a fresh ReleaseSafe build at `e77bbf7b`, macOS
aarch64, isolated `KAAPPI_HOME`.

**Fix:** in the same PR as #2110–#2113/#1922/#2155: `runFile` sets
`had_compile_error` and refuses to cache the file (`--timings` reports
`not cached: compile error`).

**Test:** `tests/scheme/differential/probes/cache-compile-error.scm` — the
differential harness fails on any recurrence.

**Priority rationale:** `high` — a legal-looking second run gets a silently
wrong exit status and a vanished diagnostic, on a path any user with a typo
reaches.

baijum:
One PR for seven cache defects with one root theme: **a HIT behaved differently from a MISS.** Six were filed (#1922, #2110, #2111, #2112, #2113, #2155); the seventh (#2187) was found while reworking the write gating. The codec fixes share a single `.sbc` format bump (v10 → v11), which also invalidates every pre-existing entry in the wild.

## Format v11 — the constant codec (`bytecode_file*`)

- **Immutability round-trips** (#2110): pair/string/vector/bytevector constants carry their `Object.flags.immutable` byte, so a `set-car!` on a literal raises KP3002 warm exactly as cold — instead of succeeding with exit 0 where the cold run exited 1.
- **Sharing round-trips** (#2111): a shareable constant reached twice is emitted once and referenced via a new `TAG_BACKREF`, registered in identical pre-order on both halves. Datum-label structure keeps `eq?` and `write-shared` across a HIT, a 20-level shared DAG drops from 4.7 MB to 474 B of `.sbc`, and cyclic literals (`'#0=(1 2 . #0#)`) terminate, cache, and HIT.
- **The two halves enforce the same limits** (#2113): the writer now *refuses* — never truncates — anything the reader would reject (`error.LimitExceeded` / `error.UnsupportedConstant`; the old `else => TAG_NIL` fallback that could serve a silently wrong constant is gone). List spines are also walked **iteratively on both halves**, so depth counts nesting only: the 257-element-list cliff is gone and long quoted data caches, while >256 *nesting* refuses loudly. `--compile` reports refusals with the reason instead of writing an artifact that fails at run time.

## Cache gating and reporting (`main.zig`, `cache.zig`)

- **`define-syntax` / `define-property` files are not cached** (#2112, the strictly-safe option the issue blessed): `runFile` snapshots `vm.macros.count()` and `vm.syntax_properties.count()` around each form's compilation, so detection is semantic — a macro *expanding into* `define-syntax` is covered too. `--timings` reports `not cached: define-syntax`.
- **Compile-error files are not cached** (#2187, found during this work): the warm run used to execute the partial program with exit 0 and the diagnostic gone. Now `not cached: compile error`.
- **Runtime diagnostics survive a HIT** (#1922): the HIT path passes `Function.source_line` — serialized since v7, equal to the datum line the fresh path uses — as the error fallback line, so `(/ 1 0)`-class errors keep their `file:line` and source snippet.
- **`cache status` verifies "current"** (#2113): each current-build entry's body is dry-run through the deserializer; one the reader rejects is reported as `unloadable` instead of `current` ("a plain run would hit" is now checked, not assumed).

## Cache key (#2155)

`compilerHashFor` gains a target component: the triple plus `types.platform_features` — the exact inputs `cond-expand` bakes into bytecode. The 17 release binaries built from one clean checkout no longer share a key, so a POSIX-compiled `.sbc` fed to `zig build -Dbundle= -Dtarget=<windows>` is a loud miss instead of silently running the wrong `cond-expand` branch, and the key reacts automatically if a feature identifier ever becomes arch-gated.

## Tests and docs

- New Zig unit tests: immutability round-trip (both flag values, all four types), sharing within and across constants, cyclic pair + self-referential vector, 400-element list, writer refusal at/past the depth cap and for oversized strings (including "no file appears on disk"), target-hash keying, `cache status` unloadable rendering.
- `src/tests_endian.zig`'s hand-spelled golden fixture extended to pin the v11 immutability byte and `TAG_BACKREF` encoding byte-for-byte (and the host-independence constant re-captured for the new key inputs).
- `tests/scheme/timings/timings-1515.sh` asserts the two new `--timings` reasons; `src/testdata/fuzz-seed.sbc` regenerated per its documented VERSION-bump procedure.
- **Differential harness: both suppression lists are now empty.** The five probes (`cache-error-location`, `sbc-literal-immutability`, `sbc-shared-structure`, `sbc-macro-table`, `sbc-constant-depth`) stay in the corpus as ordinary regression probes — any of these divergences coming back fails the run — and a new `cache-compile-error.scm` probe pins #2187. Harness header prose updated to describe the new gating.
- `docs/dev/cache.md`: the key's third component, the full refusal list with reasons, the `unloadable` state, and a new "Transparency guarantees" section.

## Verification

- Full unit suite: 1629 pass / 3 skip, including under the rebase onto #2186.
- Full `tests/scheme/run-all.sh`: 2067 pass, 0 fail (pre-rebase run; post-rebase run green on the unit suite, full suite re-running).
- Differential harness: 358 files, 0 divergences on all tiers, 44 cached files all HIT, 0 permanent misses.
- Every issue's reproduction verified cold/warm by hand, before and after.
- Cross-compiles clean: aarch64-windows, wasm32-wasi, s390x-linux (big-endian canary; the endian golden pins the new bytes independently).

Fixes #2187.


baijum:
Review response — all three CodeRabbit findings taken, in e9ff06fb:

1. **Bundle/preamble limit parity (Major)** — the sharpest catch: the v11 header comment claimed the writer enforces every limit the reader enforces, and that was false for `writeFileWithBundle`'s sections. Worth stating the blast radius precisely: it's **latent, not live** — the largest shipped `.sld` is 64 KB against the 1 MiB cap, and bundles are `--compile` artifacts, not the auto-run cache (which writes both sections empty), so the failure mode was a loud-but-late "invalid embedded bytecode" at bundle run time rather than #2113's invisible permanent miss. Fixed anyway, since it's exactly the drift class this PR exists to pin: the writer refuses over-cap counts/lengths with `LimitExceeded`, the reader's magic `4096`s are now shared `MAX_BUNDLED_FILES`/`MAX_PREAMBLE_FORMS` constants, the path cap makes the u16 cast unreachable, and a unit test covers refusal + nothing-on-disk + the in-bounds round-trip.

2. **`freeDeserializeResult` resets `funcs` (Trivial)** — taken as proposed (`&.{}`, not null — the field is a non-optional slice); the new bundle test calls it twice to pin the no-double-free behavior.

3. **`--timings` compile-error assert (Minor)** — added to timings-1515.sh with the same three checks as the other refusal reasons (note the fixture needs `|| true` inside the command substitution — the file exits 1 by design and the suite runs `set -euo pipefail`).

Full unit suite green at 1630 pass / 3 skip.

PR #2191, opened 2026-08-02
Reject non-positive segment sizes across SRFIs 152/160/178/196

### LINKED ISSUE #1949: [SRFI-160] Uvector-segment with n=0 never terminates — unbounded allocation, all 12 kinds
**Category:** SRFI 160 · **Severity:** hang (unbounded allocation) · **Found by:** systematic audit v2, Phase 2.4 (#1890)

```scheme
(s8vector-segment (s8vector 1 2 3 4) 0)
```

Never returns. Killed by `SIGALRM` at 20s (exit 142), consing a fresh zero-length copy every iteration.

## Mechanism

`%uvec-segment` (`lib/srfi/160/base.sld:357`) advances with `(min len (+ start n))`. With `n = 0`, `start` never moves, so the loop is infinite and allocates on every pass.

Affects **all 12 kinds** and **both dispatch branches** — verified on `u8` (bytevector branch) and `u16`/`c128`/`s8` (NumericVector branch).

## Discriminating controls

| input | result |
|---|---|
| `n = 0` | **hang** |
| `n = -1` | raises |
| `n = 1`, `n = 3` | correct |
| empty vector, `n = 0` | returns `()` |

The `n = -1` control is the important one: the guard for a non-positive `n` **exists** and catches negatives — zero slips past it into the loop.

## Spec

SRFI 160: *"It is an error if n is not an exact positive integer."* Zero is not positive, so raising conforms; hanging does not. The empty-vector case already returns rather than looping, which is why this is invisible unless the vector is non-empty.

## Suggested fix

Extend the existing non-positive guard to reject `0`, matching what the `-1` path already does.

**Verified:** fresh ReleaseSafe build, isolated `KAAPPI_HOME`; the hang and all four controls reproduced twice.

### LINKED ISSUE #2084: [SRFI-178] bitvector-segment recurses per segment: a 200,000-bit vector overflows the stack uncatchably, and n=0 never terminates
**Category:** SRFI
**Spec reference:** SRFI 178 §"Selectors" — "`(bitvector-segment bvec n)` Returns a list of bitvectors, each of which contains *n* consecutive elements of *bvec*. The last bitvector may be shorter than *n*. It is an error if *n* is not an exact positive integer."
**Severity:** crash (uncatchable stack overflow on a legal input)

`bitvector-segment` recurses once per segment with the `cons` outside the recursive call, so it is not a tail call and consumes one frame per piece. A perfectly legal call on a large bitvector therefore dies with `KP3008: stack overflow`, which per `CLAUDE.md` is **uncatchable** — a `guard` never sees it.

**Minimal reproduction:**

```scheme
(import (scheme base) (scheme write) (srfi 178))
(display (length (bitvector-segment (make-bitvector 200000 1) 1)))
;; expected: 200000
;; actual:   error[KP3008]: stack overflow  in __nlet_67_loop
```

The `n = 0` case is the same defect with no floor at all — it recurses forever on `(min len (+ i 0))`, again ending in an uncatchable `KP3008` rather than the error the spec calls for:

```scheme
(display (guard (e (#t 'caught)) (bitvector-segment (bitvector 1 0 1) 0)))
;; expected: an error the guard catches
;; actual:   error[KP3008]: stack overflow — the guard clause never runs
```

**Discriminating control:** `(length (bitvector-segment (make-bitvector 1000 1) 1))` returns `1000`. The same call shape succeeds at 1,000 segments and overflows at 200,000, so the failure is stack depth, not the arguments.

**The reference implementation does not have either half.** Its `bitvector-segment` is `(map W (u8vector-segment (U bvec) n))` guarded by an explicit `(unless (and (integer? n) (positive? n)) (error "bitvector-segment: not a positive integer" n))`, and loading it into this build gives a clean `KP3000` for `n = 0`.

Related but distinct: #1949 is the same *class* of bug in `(srfi 160)`'s `Uvector-segment` (unbounded allocation on `n = 0`), in a different library with different code. This one additionally fails on legal inputs.

**Source location:** `lib/srfi/178.sld`, `bitvector-segment` — `(cons (bitvector-copy bv i (min len (+ i n))) (loop (+ i n)))`.

**Disabled test:** `tests/scheme/srfi/srfi178-audit.scm` — 2 assertions marked `;; FAIL: #2084`.

### LINKED ISSUE #2172: [SRFI-152/196] string-segment and range-segment hang on n=0 — the same missing check as #1949 and #2084

baijum:
Fixes #1949, fixes #2084, fixes #2172.

Four `segment` procedures across four SRFIs shared the same missing precondition: none rejected a non-positive size, and their loops advance by that size per iteration, so `n = 0` never advanced — an unbounded-allocation hang (`%uvec-segment`, `string-segment`, `range-segment`) or unbounded recursion into an uncatchable `KP3008` (`bitvector-segment`). SRFI 171's `tsegment` was the one family member that already rejected `n = 0` at entry; this PR makes the other four do the same.

## Changes

**Guards** — each procedure now rejects a size that is not an exact positive integer at entry, with a catchable error, before touching the input:

- `lib/srfi/160/base.sld` `%uvec-segment` (#1949) — one guard covers all 12 kinds and both dispatch branches (bytevector and NumericVector). The error message names the user-facing procedure via the `kind` tag (`s8vector-segment: …`). SRFI 160: *"It is an error if n is not an exact positive integer."*
- `lib/srfi/178.sld` `bitvector-segment` (#2084) — same spec sentence in SRFI 178.
- `lib/srfi/152.sld` `string-segment` and `lib/srfi/196.sld` `range-segment` (#2172) — neither spec defines behavior for a non-positive size, and no finite list of length-0 pieces covers a non-empty input, so raising is the only sane behavior (matching `tsegment`).

**Stack safety** (#2084's second half) — `bitvector-segment` recursed once per segment with the `cons` outside the recursive call, so a legal call on a 200,000-bit vector died with an uncatchable `KP3008`. It now accumulates in tail position and reverses, the same shape the other three already had.

Behavior change worth noting: `(s8vector-segment (s8vector) 0)` previously returned `()` (the empty vector was the one input where `n = 0` terminated); it now raises like every other `n = 0` call, since the guard runs before the loop. The audit test asserting the old behavior was updated — SRFI 160 makes `n = 0` an error regardless of the vector.

## Tests

- `tests/scheme/srfi/srfi178-audit.scm` — the two assertions disabled with `;; FAIL: #2084` are re-enabled, plus an inexact-`n` case. The 200,000-segment depth case lives in a new file, `tests/scheme/srfi/srfi178-segment-stack-2084.scm`, listed in `KAAPPI_GC_STRESS_SKIP` (reason (a)): under stress collection, building 200k bitvectors is quadratic in the live heap (measured ~3 min at 40k; the first CI run timed out at 200k), and any count that still exceeds the 32,768 frame cap is already minutes. `run-all.sh` runs it plain on every PR (~0.1 s); the audit file's other 369 assertions — including the `n = 0` guard — stay stressed (5.3 s under the stress binary).
- `tests/scheme/audit/primitives_srfi160-audit.scm` — the assertion disabled with `;;; FAIL: #1949` is re-enabled, plus: the bytevector-branch (`u8`) case, the empty-vector `n = 0` case (now expects a raise), and an inexact-`n` case.
- `tests/scheme/srfi/srfi152.scm`, `tests/scheme/srfi/srfi152-audit.scm`, `tests/scheme/srfi/srfi196.scm` — new regression assertions that `n = 0` (and `n = -1`) raises catchably.

Verified: all five modified test files pass; family-wide grep for `segment`/`chunk` found no fifth member (`chunk` appears only in comments); full `tests/scheme/run-all.sh` green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -316,9 +316,15 @@ jobs:
           #              GC-owned strings, so a nongenerative uid stops
           #              resolving once its string is collected (19 assertions)
           #
-          # All eleven still run stressed in the nightly fuzz.yml legs' unit
-          # phase where applicable, and unstressed under run-all.sh on every
-          # other leg here, so nothing is untested -- only unstressed.
+          # Listing a file here removes its stressed Scheme run entirely --
+          # fuzz.yml's gc-stress legs run fuzz targets and the Zig unit
+          # suite, never this corpus. Every listed file still runs plain
+          # under run-all.sh on every other leg, so nothing is untested,
+          # only unstressed; where the guarded behavior has a counterpart,
+          # that half keeps stress coverage (tests_reader_incremental.zig in
+          # the gc-stress unit job for reader-port-refill-gaps.scm, and
+          # srfi178-audit.scm, still in this run, for the n = 0 half of
+          # srfi178-segment-stack-2084.scm).
           #
           # The script FAILS if a name here matches no file, so this list
           # cannot rot into a silent no-op.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -297,7 +297,12 @@ jobs:
           #     the 900 s stress timeout on this job's first run against it,
           #     PR #2174); the incomplete-input reader mode it guards gets its
           #     stress coverage from src/tests_reader_incremental.zig in the
-          #     gc-stress unit job instead.
+          #     gc-stress unit job instead. srfi178-segment-stack-2084.scm
+          #     builds a 200,000-element bitvector list to pin #2084's
+          #     tail-recursion fix past the 32,768 frame cap -- ~0.1 s plain,
+          #     exit 124 stressed (measured quadratic: ~3 min at 40k on a
+          #     fast machine); the n = 0 guard half of #2084 stays stressed
+          #     in srfi178-audit.scm.
           #
           # (b) REAL BUGS THIS GATE FOUND ON ITS FIRST RUN -- excluded so the
           #     gate is green for everything else, exactly as the campaign
@@ -311,7 +316,7 @@ jobs:
           #              GC-owned strings, so a nongenerative uid stops
           #              resolving once its string is collected (19 assertions)
           #
-          # All ten still run stressed in the nightly fuzz.yml legs' unit
+          # All eleven still run stressed in the nightly fuzz.yml legs' unit
           # phase where applicable, and unstressed under run-all.sh on every
           # other leg here, so nothing is untested -- only unstressed.
           #
@@ -319,7 +324,7 @@ jobs:
           # cannot rot into a silent no-op.
           KAAPPI_GC_STRESS_SKIP: >-
             trampoline-polish-1375.scm srfi14.scm srfi264.scm
-            reader-port-refill-gaps.scm
+            reader-port-refill-gaps.scm srfi178-segment-stack-2084.scm
             primitives_srfi1-audit.scm srfi1-ext.scm srfi1-gc-stress.scm
             primitives_srfi237-audit.scm srfi237.scm srfi240-audit.scm
 
--- a/lib/srfi/178.sld
+++ b/lib/srfi/178.sld
@@ -164,12 +164,17 @@
     (define (bitvector-drop-right bv n) (bitvector-copy bv 0 (- (%len bv) n)))
 
     (define (bitvector-segment bv n)
+      ;; "It is an error if n is not an exact positive integer." n = 0 would
+      ;; never advance the loop below (#2084), so reject it up front. The loop
+      ;; accumulates in tail position — a cons around the recursive call costs
+      ;; one frame per segment and overflows uncatchably on large inputs.
+      (unless (and (exact-integer? n) (positive? n))
+        (error "bitvector-segment: size must be an exact positive integer" n))
       (let ((len (%len bv)))
-        (let loop ((i 0))
+        (let loop ((i 0) (acc '()))
           (if (>= i len)
-              '()
-              (cons (bitvector-copy bv i (min len (+ i n)))
-                    (loop (+ i n)))))))
+              (reverse acc)
+              (loop (+ i n) (cons (bitvector-copy bv i (min len (+ i n))) acc))))))
 
     (define (%refs bvs i) (map (lambda (bv) (bitvector-ref/int bv i)) bvs))
 
--- a/tests/scheme/audit/primitives_srfi160-audit.scm
+++ b/tests/scheme/audit/primitives_srfi160-audit.scm
@@ -716,6 +716,17 @@
             'ERR (try (lambda () (s8vector-segment (s8vector) 0))))
 (test-equal "segment with an inexact n raises"
             'ERR (try (lambda () (s8vector-segment (s8vector 1 2) 2.0))))
+;; The guard lives once in %uvec-segment, so s8/u8 above already prove both
+;; dispatch branches — this sweep pins that every public wrapper reaches it,
+;; so a future per-kind split cannot drop the guard for one kind unnoticed.
+(for-each
+ (lambda (row seg)
+   (test-equal (nm (k-name row) "vector-segment rejects n = 0") 'ERR
+               (try (lambda () (seg ((k-make row) 2) 0)))))
+ all-kinds
+ (list s8vector-segment u8vector-segment u16vector-segment s16vector-segment
+       u32vector-segment s32vector-segment u64vector-segment s64vector-segment
+       f32vector-segment f64vector-segment c64vector-segment c128vector-segment))
 
 ;;; ------------------------------------------------------------------
 ;;; 11. Comparators
--- a/tests/scheme/audit/primitives_srfi160-audit.scm
+++ b/tests/scheme/audit/primitives_srfi160-audit.scm
@@ -702,21 +702,20 @@
 (test-equal '((1) (2)) (map s8vector->list (s8vector-segment (s8vector 1 2) 1)))
 (test-equal '() (s8vector-segment (s8vector) 2))
 ;; SRFI 160: "It is an error if n is not an exact positive integer."
-;; n = -1 raises; n = 0 does not — see the disabled case below.
-(test-equal 'ERR (try (lambda () (s8vector-segment (s8vector 1 2) -1))))
-;; Discriminating control for the hang below: with an EMPTY vector the
-;; loop's (>= start len) guard is true on entry, so n = 0 terminates.
-(test-equal '() (s8vector-segment (s8vector) 0))
-
-;;; FAIL: #1949 (Uvector-segment with n = 0 never terminates)
-;;; %uvec-segment (lib/srfi/160/base.sld:357) advances with
-;;;   (loop (min len (+ start n)) ...)
-;;; so n = 0 leaves `start` unchanged forever while consing a fresh
-;;; zero-length copy each iteration — an unbounded-allocation infinite
-;;; loop, on all 12 kinds. SRFI 160 makes n = 0 "an error", so any raise
-;;; would conform; a hang does not. Controls above: n = -1 raises, n = 1
-;;; and n = 3 work, and an empty vector with n = 0 returns ().
-;;; (test-equal 'ERR (try (lambda () (s8vector-segment (s8vector 1 2 3 4) 0))))
+;; Regression, #1949: %uvec-segment used to advance with
+;; (min len (+ start n)), so n = 0 looped forever on any non-empty vector
+;; while consing a fresh zero-length copy each iteration — on all 12 kinds.
+;; It now rejects a non-positive (or inexact, or non-integer) n at entry,
+;; before the loop, so even the empty vector raises rather than answering.
+(test-equal "segment n = -1 raises" 'ERR (try (lambda () (s8vector-segment (s8vector 1 2) -1))))
+(test-equal "segment n = 0 raises catchably (NumericVector branch)"
+            'ERR (try (lambda () (s8vector-segment (s8vector 1 2 3 4) 0))))
+(test-equal "segment n = 0 raises catchably (bytevector branch)"
+            'ERR (try (lambda () (u8vector-segment (u8vector 1 2 3 4) 0))))
+(test-equal "segment n = 0 raises even on the empty vector"
+            'ERR (try (lambda () (s8vector-segment (s8vector) 0))))
+(test-equal "segment with an inexact n raises"
+            'ERR (try (lambda () (s8vector-segment (s8vector 1 2) 2.0))))
 
 ;;; ------------------------------------------------------------------
 ;;; 11. Comparators
--- a/tests/scheme/srfi/srfi152.scm
+++ b/tests/scheme/srfi/srfi152.scm
@@ -175,6 +175,12 @@
 (test-equal '("ab" "cd" "e") (string-segment "abcde" 2))
 (test-equal '("abc") (string-segment "abc" 5))
 (test-equal '() (string-segment "" 3))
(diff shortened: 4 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -311,15 +316,15 @@ jobs:
           #              GC-owned strings, so a nongenerative uid stops
           #              resolving once its string is collected (19 assertions)
           #
-          # All ten still run stressed in the nightly fuzz.yml legs' unit
+          # All eleven still run stressed in the nightly fuzz.yml legs' unit
```
baijum: Nit on accuracy: `fuzz.yml`'s gc-stress legs only run the Zig unit suite, not this Scheme corpus — so none of the eleven `KAAPPI_GC_STRESS_SKIP` entries "still run stressed" there. That was already a bit loose for the previous ten, but it is especially misleading for this new reason-(a) file, which exists precisely because it cannot be stressed.

Suggest something closer to: the reason-(a) files (including this one) stay covered unstressed via `run-all.sh`, and the `n = 0` half of #2084 remains stressed inside `srfi178-audit.scm`. Drop or narrow the nightly-fuzz claim so a later reader does not assume the 200k case gets stress coverage elsewhere.

tests/scheme/srfi/srfi178-audit.scm
```diff
@@ -443,14 +443,16 @@
...
+;; full-depth 200,000-segment case lives in srfi178-segment-stack-2084.scm —
+;; its own file so the gc-stress leg can skip just it (building 200k
+;; bitvectors is quadratic in the live heap under stress collection) without
+;; skipping this file's assertions.
+(test-assert "segment with n=0 raises a catchable error"
+             (raises? (lambda () (bitvector-segment (bitvector 1 0 1) 0))))
+(test-assert "segment with an inexact n raises a catchable error"
+             (raises? (lambda () (bitvector-segment (bitvector 1 0 1) 2.0))))
 
 ;; Discriminating control: the same shape at a size the recursion survives.
```
baijum: Stale framing after the fix: this is no longer a "discriminating control" for recursion depth — with the accumulate-then-reverse loop every size is one frame, and the cap-exceeding case lives in `srfi178-segment-stack-2084.scm`.

Worth rewording to a plain smoke/regression note (e.g. "smaller n=1 case kept stressed here") so the next reader does not think the 1,000-bit case is still doing the stack-depth job.

tests/scheme/srfi/srfi178-audit.scm
```diff
@@ -443,14 +443,16 @@
...
+;; Regression, #2084: the loop now accumulates in tail position (one frame
+;; total, not one per segment) and rejects a non-positive n at entry. The
+;; full-depth 200,000-segment case lives in srfi178-segment-stack-2084.scm —
+;; its own file so the gc-stress leg can skip just it (building 200k
+;; bitvectors is quadratic in the live heap under stress collection) without
+;; skipping this file's assertions.
+(test-assert "segment with n=0 raises a catchable error"
+             (raises? (lambda () (bitvector-segment (bitvector 1 0 1) 0))))
+(test-assert "segment with an inexact n raises a catchable error"
+             (raises? (lambda () (bitvector-segment (bitvector 1 0 1) 2.0))))
```
baijum, line 457: Minor coverage asymmetry: 160's audit covers `n = -1`, `n = 0`, empty+`n = 0`, and inexact; here you have `n = 0` and inexact but not `n = -1` (and not empty+`n = 0`, which used to return `()` and now raises). Same shape as the hang — optional one-liner if you want the family uniform before merge; not blocking.

tests/scheme/srfi/srfi152.scm
```diff
@@ -175,6 +175,12 @@
 (test-equal '("ab" "cd" "e") (string-segment "abcde" 2))
 (test-equal '("abc") (string-segment "abc" 5))
 (test-equal '() (string-segment "" 3))
+;; Regression, #2172: k = 0 never advanced the loop, hanging forever on any
+;; non-empty string. It must raise a catchable error instead.
+(test-equal "string-segment k=0 raises catchably"
+            'caught (guard (e (#t 'caught)) (string-segment "abcdef" 0) 'no-error))
+(test-equal "string-segment k=-1 raises catchably"
+            'caught (guard (e (#t 'caught)) (string-segment "abcdef" -1) 'no-error))
```
baijum, line 183: Optional: 160/178 also pin an inexact size (`2.0`) against the `exact-integer?` half of the guard. 152/196 only cover `0`/`-1` here. Same guard expression, so behavior is already right — worth a one-liner only if you want the four SRFIs' regressions to match.

lib/srfi/178.sld
```diff
@@ -164,12 +164,17 @@
...
+        (error "bitvector-segment: size must be an exact positive integer" n))
       (let ((len (%len bv)))
-        (let loop ((i 0))
+        (let loop ((i 0) (acc '()))
           (if (>= i len)
-              '()
-              (cons (bitvector-copy bv i (min len (+ i n)))
-                    (loop (+ i n)))))))
+              (reverse acc)
+              (loop (+ i n) (cons (bitvector-copy bv i (min len (+ i n))) acc))))))
```
baijum, line 178: Tail shape looks correct — `loop` is in tail position with the `cons` in the accumulator arg, same as `%uvec-segment` / `string-segment` / `range-segment`. Combined with the new 200k file this should permanently close the uncatchable `KP3008` half of #2084.

baijum (commented):
**Review: looks good to merge** (would APPROVE, but GitHub blocks self-approve on this auth).

Solid, focused fix for a real hang/stack-overflow class across the four remaining `segment` cousins of SRFI 171's `tsegment`.

The guards match the SRFI 160/178 wording (`exact-integer?` + `positive?`) and are the only sane reading for 152/196; they run before the loop so empty inputs with `n = 0` raise too (the one intentional behavior change, covered for 160). `bitvector-segment`'s accumulate-then-reverse rewrite is the right shape, and the 200k case in its own stress-skipped file is the correct way to pin past the 32,768 frame cap without tanking `gc-stress-scheme`.

`%uvec-segment`'s `(string-append (symbol->string kind) "vector-segment: …")` naming is correct for the `'s8`/`'u8`/… kind tags.

Left a few non-blocking notes inline (stale CI prose about nightly fuzz stress, a leftover "recursion survives" comment, and a small test-symmetry gap for `n = -1` / inexact on 152/196/178). No correctness blockers.

tests/scheme/audit/primitives_srfi160-audit.scm
```diff
@@ -702,21 +702,20 @@
...
+;; before the loop, so even the empty vector raises rather than answering.
+(test-equal "segment n = -1 raises" 'ERR (try (lambda () (s8vector-segment (s8vector 1 2) -1))))
+(test-equal "segment n = 0 raises catchably (NumericVector branch)"
+            'ERR (try (lambda () (s8vector-segment (s8vector 1 2 3 4) 0))))
+(test-equal "segment n = 0 raises catchably (bytevector branch)"
+            'ERR (try (lambda () (u8vector-segment (u8vector 1 2 3 4) 0))))
+(test-equal "segment n = 0 raises even on the empty vector"
+            'ERR (try (lambda () (s8vector-segment (s8vector) 0))))
+(test-equal "segment with an inexact n raises"
+            'ERR (try (lambda () (s8vector-segment (s8vector 1 2) 2.0))))
```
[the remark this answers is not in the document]
baijum: Partially adopted in 8d5b73c0, deliberately not the full 12×3 matrix. The guard is one shared definition in `%uvec-segment` — all 12 wrappers are one-line delegations passing a kind tag — so 36 assertions would test the same line 36 times. The two axes that actually discriminate were already covered: both dispatch branches (s8 = NumericVector, u8 = bytevector, the axis #1949 itself verified on) and every predicate term (-1, 0, 2.0, empty+0) through s8. What the sweep adds is the third real axis: that every public wrapper *reaches* the shared guard, so a future per-kind split cannot drop it for one kind unnoticed — n = 0 through all 12 wrappers via the existing `all-kinds` table, 12 assertions.

tests/scheme/srfi/srfi152.scm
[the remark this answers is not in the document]
baijum: Adopted in 8d5b73c0, and went one further: each of the four libraries carries its own copy of the guard expression, so each site now covers the full matrix — n = 0, n = -1, inexact 2.0, and n = 0 on the empty input. The empty-input case is the one with real teeth: it pins guard-before-length-check order (empty inputs used to return `()` under n = 0), which is the one subtle placement decision in the fix.

.github/workflows/ci.yml
baijum: Right on both counts — I checked fuzz.yml before rewording: its gc-stress legs build fuzz targets and run the Zig unit suite, never this corpus, so the sentence was already loose for the previous ten and would have been outright wrong for this file. Rewritten in 8d5b73c0 to say what is true: listing a file removes its stressed Scheme run entirely; plain coverage stays via run-all.sh on every other leg; and the named counterparts keep the stressed half (tests_reader_incremental.zig for the reader file, srfi178-audit.scm — still in this run — for the n = 0 half of #2084).

tests/scheme/srfi/srfi178-audit.scm
baijum: Reworded in 8d5b73c0 — it now reads as the smoke check it is ("every size costs one frame") and points at srfi178-segment-stack-2084.scm for the depth job. Also dropped "does terminate" from the assertion name, which read as if termination were still in question.

tests/scheme/srfi/srfi178-audit.scm
baijum: Both added in 8d5b73c0 — n = -1 and empty+n = 0 (the latter with a comment noting it used to return `()`, same as the 160 case). The four sites now carry the identical matrix.

tests/scheme/srfi/srfi152.scm
baijum: Added in 8d5b73c0 — 2.0 at both the 152 and 196 sites (plus empty+0 for the same uniformity). Worth having despite the identical guard expression: each .sld carries its own copy, so per-library coverage of each predicate term is what keeps a future single-file edit honest.

Issue #1998, opened 2026-08-01
[R7RS 6.13.1] close-input-port on a bidirectional custom port closes the output side too (and vice versa)


baijum:
**Category:** R7RS I/O × SRFI 181 · **Severity:** edge-case · **Found by:** Systematic audit v2, Phase 2.3 (#1890)

`types.Port` carries a single `is_open` flag, and `closePortObj` sets it unconditionally. `close-input-port` and `close-output-port` therefore both close the whole port — and run the custom port's `close` callback — even when the port is simultaneously an input and an output port.

## Spec

R7RS-small § 6.13.1, `close-port` / `close-input-port` / `close-output-port`:

> Scheme implementations may provide ports which are simultaneously input and output ports, such as sockets; the `close-input-port` and `close-output-port` procedures can then be used to close the input and output sides of the port independently.

Kaappi does provide such a port — `make-custom-binary-input/output-port` is the only way to construct one — so the sentence applies.

## Minimal reproduction

```scheme
(import (scheme base) (scheme write) (srfi 181))
(define n 0)
(define io (make-custom-binary-input/output-port "x"
             (lambda (b s c) 0) (lambda (b s c) c) #f #f
             (lambda () (set! n (+ n 1)))))

(list (input-port-open? io) (output-port-open? io))   ;; ⟹ (#t #t)
(close-input-port io)
(list (input-port-open? io) (output-port-open? io) n) ;; expected: (#f #t 0)
                                                      ;; actual:   (#f #f 1)
```

The `close` callback fires on the *first* side closed, so a program that closes the write side and keeps reading loses the reader as well, and a program that closes both sides in sequence runs `close` after the first, not the second.

## Discriminating control

`close-output-port` first produces the mirror-image result — `(#f #f 1)` — so it is not an input-specific bug, and a subsequent `close-output-port` (or `close-port`) is a no-op that does **not** run the callback a second time. That last part is right; only the side isolation is missing.

`close-port` on the same port is fully correct: both sides close, callback once.

**Verified:** fresh ReleaseSafe build at `e6081c3b`, isolated `KAAPPI_HOME`, cache cleared. Deterministic (not a crash — gc-stress not applicable).
**Source location:** `src/primitives_io.zig` `closePortObj` (`port.is_open = false` with no direction argument); `types_port.Port` has no per-direction open state to set.

[...]

Issue #1997, opened 2026-08-01
[SRFI-181] Transcoded output ignores #\return as a line ending: a crlf transcoder turns one CRLF into CR CR LF, doubling every line break on round trip


baijum:
**Category:** SRFI 181 transcoded ports · **Severity:** wrong-result · **Found by:** Systematic audit v2, Phase 2.3 (#1890)

The encode half of the transcoder converts `#\newline` only. The decode half correctly treats all three line endings — this implementation's own input direction converts a lone `#\return` and a CRLF pair to `#\newline` — so the two halves disagree about what a line ending is, and a transcoder is no longer a round trip.

## Spec

SRFI 181, § *End-of-line styles*:

> In order to conform to this SRFI, implementations must support at least three kinds of line endings: a `#\newline` character, a `#\return` character, and a `#\return` followed by a `#\newline`, which is known as a CRLF sequence.

> On an output port, the symbol `crlf` causes **any line ending** to be output as a CRLF sequence, whereas the symbol `lf` causes **any line ending** to be output as a `#\newline` character, All other characters remain unchanged.

## Minimal reproduction

```scheme
(import (scheme base) (scheme write) (srfi 181))
(define t (make-transcoder (utf-8-codec) (quote crlf) (quote replace)))

(string->bytevector "a\rb" t)     ;; expected: #u8(97 13 10 98)      actual: #u8(97 13 98)
(string->bytevector "a\r\nb" t)   ;; expected: #u8(97 13 10 98)      actual: #u8(97 13 13 10 98)
```

The second line is the damaging one: the `#\return` passes through unconverted and the following `#\newline` is expanded to CRLF, producing **CR CR LF**.

## The round trip loses data

```scheme
(bytevector->string (string->bytevector "a\r\nb" t) t)
;; expected: "a\nb"  (3 chars, one line break)
;; actual:   "a\n\nb" (4 chars, two line breaks)
(string-length (bytevector->string (string->bytevector "a\r\nb" t) t))  ;; ⟹ 4, expected 3
```

Decoding is correct on its own — it collapses the spurious CR CR LF into two newlines exactly as the spec requires — so the corruption is entirely on the encode side.

`lf` has the same defect in the other direction: `(string->bytevector "a\rb" lf-transcoder)` ⟹ `#u8(97 13 98)` where the spec asks for `#u8(97 10 98)`.

## Discriminating control

[...]

Issue #1943, opened 2026-07-31
[SRFI-181] flush-output-port on a transcoded port is a silent no-op — data stays unwritten


baijum:
**Category:** SRFI 181 transcoded ports · **Severity:** wrong-result (durability) · **Found by:** systematic audit v2, Phase 2.2 (#1890)

`flushOutputPort` (`primitives_io.zig:1818`) has a `custom_backend` branch and an `isBufferedFdPort` branch, but **no `transcode` branch** — and a transcoded port carries `fd = -1`, so neither applies. The call succeeds and does nothing.

Write 5 bytes through a transcoded output port, call `flush-output-port`, and the file is **0 bytes**.

## Discriminating controls

- `flush-output-port` on the **wrapped** port → file is 5 bytes. The data exists; only the transcoded wrapper's flush is inert.
- `close-port` on the **transcoded** port → 5 bytes. The cascade works.

So flushing is the one operation in the group that silently does not.

## Why it matters

`flush-output-port` is the operation a program calls precisely when it needs durability *without* closing — a log line before a long computation, output before a fork, a checkpoint. Silently doing nothing defeats exactly that use.

Undocumented: no mention in `CONFORMANCE.md`, `README.md`, or `lib/srfi/181.sld`.

**Verified:** fresh ReleaseSafe build, isolated `KAAPPI_HOME`; the no-op and both controls reproduced.

baijum:
Fixed in `92e3133c` (PR #2192, squash-merged to main).

The PR's closing line listed all six issues after a single `Closes` keyword, which GitHub only applies to the first — hence closing this one by hand. The fix, its regression tests, and the re-enabled audit assertions are all in that commit.

Issue #1942, opened 2026-07-31
[SRFI-192] port-has-set-port-position!? is registered to the wrong function — wrong in both directions


baijum:
**Category:** SRFI 192 · **Severity:** wrong-result · **Found by:** systematic audit v2, Phase 2.2 (#1890)

`port-has-port-position?` and `port-has-set-port-position!?` are **both registered to `portHasPortPositionP`** (`primitives_io.zig:37-38`), which inspects only `get_position_proc`. So the second predicate answers a question about the *getter*.

SRFI 192: *"Returns `#t` if port supports the `set-port-position!` operation."*

Wrong in **both** directions:

| custom port has | `port-has-set-port-position!?` | reality |
|---|---|---|
| get-position, **no** set-position! | **`#t`** | `set-port-position!` **raises** |
| set-position!, **no** get-position | **`#f`** | `set-port-position!` **succeeds** |

## Discriminating controls

Ports with **both** answer `#t`/`#t`; ports with **neither** answer `#f`/`#f`. That is why the defect is invisible on every ordinary port — only a custom port with exactly one of the two exposes it, and both asymmetric cases are wrong.

This is a **D4 registration-table defect** — the kind the audit's new dimension list was added for. Two spec entries pointing at one function, where the names promise different questions.

**Verified:** fresh ReleaseSafe build, isolated `KAAPPI_HOME`; all four combinations reproduced.

baijum:
Fixed in `92e3133c` (PR #2192, squash-merged to main).

The PR's closing line listed all six issues after a single `Closes` keyword, which GitHub only applies to the first — hence closing this one by hand. The fix, its regression tests, and the re-enabled audit assertions are all in that commit.

PR #2192, opened 2026-08-02
Fix six port-layer defects across SRFI 181, SRFI 192 and R7RS 6.13.1

### LINKED ISSUE #1995: [SRFI-181] (read port) returns #<eof> on every custom and transcoded port — readDatumFn bypasses readOneByte
**Category:** SRFI 181 / R7RS I/O · **Severity:** wrong-result · **Found by:** Systematic audit v2, Phase 2.3 (#1890)

`read` is the one input primitive that does not go through `readOneByte`. `readDatumFn` (`src/primitives_io.zig:1590`) drains the port's software buffers (`peek_byte` → `peek_extra` → `read_buf`) and then refills straight from `portFdRead(port, ...)`. A custom port and a transcoded port both carry the `fd = -1` sentinel and have **no** `custom_backend`/`transcode` branch here, so `platform.read(-1, ...)` fails with `EBADF`, the loop treats that as end of input, and `read` reports EOF — the `read!` callback is never invoked even once.

Every other input primitive on the same port works: `read-char`, `peek-char`, `read-line`, `read-string`, `read-u8`, `read-bytevector`, `read-bytevector!` all funnel through `readOneByte`, which does have both branches. This is exactly the invariant CLAUDE.md states — "`readOneByte` / `portWriteBytes` in `src/primitives_io.zig` are the single byte source/sink for every textual *and* binary port primitive — hook new I/O through them, not around them" — and `read` is the one place it is broken.

## Minimal reproduction

```scheme
(import (scheme base) (scheme write) (srfi 181))
(define s "(1 2 3)") (define i 0)
(define (mk) (set! i 0)
  (make-custom-textual-input-port "d"
    (lambda (str st c)
      (if (>= i (string-length s)) 0
          (begin (string-set! str st (string-ref s i)) (set! i (+ i 1)) 1)))
    #f #f #f))

(read (mk))                                     ;; expected: (1 2 3)   actual: #<eof>
(read (transcoded-port (open-input-bytevector (string->utf8 "(1 2 3)"))
                       (native-transcoder)))    ;; expected: (1 2 3)   actual: #<eof>
```

A binary custom port behaves the same way (`#<eof>`).

## Second symptom: a spurious `KP3000: read error`

When the port's buffers hold only *part* of a datum, `read` drains them, cannot refill, and reports a read error instead of EOF:

```scheme
(let ((p (mk))) (peek-char p) (read p))
;; error[KP3000]: read error       (exit 1)
```

`peek-char` leaves `(` in `peek_byte`; `read` consumes it, gets `UnexpectedEof` from the reader, tries to refill, fails, and raises.

## Third symptom: it works by accident

If a single `read!` burst happens to buffer the whole datum into `read_buf` *before* `read` is called, `read` succeeds — so the failure is a function of the callback's chunk size, not of anything the program did:

```scheme
(define done #f)
(define p3 (make-custom-textual-input-port "d"
  (lambda (str st c)
    (if done 0 (begin (set! done #t)
                      (do ((k 0 (+ k 1))) ((= k 7))
                        (string-set! str (+ st k) (string-ref "(1 2 3)" k)))
                      7)))
  #f #f #f))
(peek-char p3)   ;; primes read_buf with the whole datum
(read p3)        ;; ⟹ (1 2 3)
```

## Discriminating controls

| port | `read-char` | `read` |
|---|---|---|
| `open-input-string` | works | **works** |
| `open-input-file` | works | **works** |
| custom textual | works | `#<eof>` |
| custom binary | works | `#<eof>` |
| transcoded | works | `#<eof>` |

Same datum, same bytes; only the port kind differs. So this is the `readOneByte` bypass, not a reader or datum problem.

**Verified:** fresh ReleaseSafe build at `e6081c3b`, isolated `KAAPPI_HOME`, cache cleared. Deterministic (not a crash — gc-stress not applicable).
**Source location:** `src/primitives_io.zig:1590` `readDatumFn` — the `is_string_port` fast path is handled, `custom_backend` and `transcode` are not; the refill at `portFdRead(port, &tmp, tmp.len)` is the bypass.
**Disabled test:** `tests/scheme/audit/primitives_srfi181-audit.scm` (marked `;; FAIL: #this`)

baijum:
Closes #1995, #1997, #1943, #1942, #1941, #1998.

All six are port-layer defects the SRFI 181 audit (#1890 phase 2.2/2.3) filed, and five of the six are the same shape: a branch that was never written for a port kind with no fd. A custom port and a transcoded port both carry the `fd = -1` sentinel, so a path missing their branch fell through to a real syscall on `-1`, failed with `EBADF`, and reported that as ordinary end of input — or as nothing at all.

## What each fix does

**#1995 — `read` returned `#<eof>` on every custom and transcoded port.** `readDatumFn` is the one input primitive that does not go through `readOneByte`; its refill called `portFdRead(-1, ...)` directly, so the `read!` callback was never invoked even once. It now refills through `readOneByte` for these two port kinds, restoring the invariant `CLAUDE.md` already states. That also removes the spurious `KP3000` the partially-drained case produced (`peek-char` then `read`), and the chunk-size dependence that made a whole-datum burst work by accident.

**#1997 — a `crlf` transcoder turned one CRLF into CR CR LF.** The encode half converted only `#\newline`, while the decode half already treated bare CR, bare LF, and CRLF alike — so the two halves disagreed about what a line ending is and a round trip doubled every line break. Encode now recognizes the same three endings decode does. A CRLF split across two writes (`write-char #\return` then `write-char #\newline`) stays one line ending via a new `TranscodeState.pending_cr`, committed only after the wrapped write succeeds so a parked write's from-scratch retry recomputes the identical translation.

**#1943 — `flush-output-port` on a transcoded port was a silent no-op.** The function had a `custom_backend` branch and an `isBufferedFdPort` branch and no `transcode` branch. The body is now `flushPortObj`, which cascades to the wrapped port — itself possibly a custom or buffered fd port, hence a recursive call rather than a drain.

**#1942 — `port-has-set-port-position!?` answered a question about the getter.** It was registered to `portHasPortPositionP`, which inspects `get_position_proc`. Wrong in both directions for a custom port carrying one of the two procedures without the other; invisible on every other port kind, which supports both or neither. It gets its own function.

**#1941 — string ports ignored peek read-ahead.** The string branch of `port-position`/`set-port-position!` neither subtracted software read-ahead nor discarded it on seek, though the fd branch has always done both and `read-line`'s CR handling pushes a byte back on string ports too. Position over-reported by one, and a seek served the stale byte first.

**#1998 — `close-input-port` closed the output side too.** R7RS 6.13.1 gives the two names independent semantics on a port that is simultaneously input and output, which `make-custom-binary-input/output-port` constructs. `Port` gains `input_closed`/`output_closed`; the shared backing (close callback, fd) is released only when the second side goes, so the callback still runs exactly once. Single-direction ports are unaffected — both names still behave exactly like `close-port` there. Both flags are set *after* the possibly-suspending flush completes, mirroring `closePortObj`'s own drain-then-mark order, so `waitPortFd`'s new half-close re-check cannot misread this close's own drain as a sibling's.

## Verification

A/B against the pre-fix **v0.22.1** binary on one script: every reported symptom reproduces there (`(3 b)`, `X`, `(#f #t)`, `#<eof>`, the `KP3000`, `(97 13 98)`, `(97 13 13 10 98)`, `4`, `(#f #f 1)`) and none here.

- `zig build test` — full unit suite green, including 12 new regression tests in `tests_srfi181.zig` and 2 in `tests_io.zig`.
- `bash tests/scheme/run-all.sh` — **2070 pass, 0 fail** (675 Scheme files + the 1395-test R7RS suite).
- 17 previously disabled audit assertions re-enabled across `primitives_io-audit.scm` and `primitives_srfi181-audit.scm`; both "current (wrong) behaviour" pins flipped to assert the correct value rather than deleted, so the fix moved a test rather than silently changing an untested behaviour. Audit files now at 227 and 215 assertions, all green.

`tests_gc_tracing.zig`'s field inventory is re-pinned for the three new fields — all plain bools, so they add no marking or sweeping obligation. `CONFORMANCE.md`'s SRFI 192 paragraph is corrected: it claimed the buffering correction applied to every port kind while describing only the fd branch, which #1941 was the counter-example to.


baijum:
Thanks — I looked at all three. Declining each, with reasoning; none is silently resolved.

**1. `CONFORMANCE.md:135` — move the file under `docs/dev/`.**

Declining: this flags a pre-existing, repo-wide condition as though this diff introduced it. `CONFORMANCE.md` has lived at the repository root since long before this PR, is referenced from `README.md` by that path, and this diff touches exactly one sentence inside it — correcting a claim that #1941 turned out to be the counter-example to (it said the read-ahead correction applied "for every port kind" while describing only the fd branch). Relocating the file and updating every reference is a separate change with a much larger blast radius than the bug fix it would be bolted onto. This is the same finding raised on the SRFI #1694 slice PR and declined there for the same reason.

**2. `src/tests_io.zig:544` — use `th.TestContext` instead of manual `memory.GC` + `th.makeTestVM`.**

Declining: it inverts the file's convention. `src/tests_io.zig` has 29 tests using the manual `memory.GC.init` + `th.makeTestVM` pattern and zero using `th.TestContext` — including the several existing multi-evaluation tests (`peek-char does not consume`, `read-line with multiple lines`, `parameterize current-input-port redirects read-line`). Adopting `TestContext` in exactly the two new tests would make them the only outliers in the file. `TestContext` is genuinely the better helper, but converting this file is its own mechanical change, not something to smuggle into a bug fix.

**3. `primitives_io-audit.scm:618` — the fixed `/tmp` path can collide.**

Declining: the premise doesn't hold here. `run-all.sh` parallelises across *files*, not within one — this audit file is a single sequential interpreter process — and `kaappi-audit-io-tcf.txt` appears in exactly one file in the tree (`grep -rn kaappi-audit-io-tcf tests/` → one hit), as does every other `kaappi-audit-io-*` path, all of which are already fixed paths with per-test suffixes (`-cr`, `-cr2`, `-tcw`, `-tcc`, `-hw`, …). `run-all.sh`'s own header records that the suite was audited for exactly this: "the suite's own audit found no cross-file collisions on fixed paths or ports." This particular line is also not new — it was present but commented out as the `FAIL: #1943` marker; this PR only re-enables it.

PR #2193, opened 2026-08-02
Stop two custom-port callback shapes from aborting the process

### LINKED ISSUE #1939: A custom-port read! callback that re-enters a read on its own port aborts the process (exit 134)

### LINKED ISSUE #2000: [SRFI-181] Channel/fiber blocking inside a custom-port callback bypasses in_custom_port_callback and aborts the process (SIGBUS)

baijum:
A SRFI 181 callback runs re-entrantly, under `vm.callWithArgs` with
`dispatched_from_scheduler` forced false. Two things the port layer assumed could
not happen from there both could, and both killed the process uncatchably from
ordinary Scheme.

## #1939 — a `read!` that reads from its own port

`port.read_buf` is a single slot, and the three fills that stash into it — the fd
burst, a custom `read!`'s result, a transcoded port's re-encoded character — each
did `std.debug.assert(port.read_buf == null)` first.

A `read!` that reads from its own port runs a whole *earlier* `read!` to
completion, so that slot already holds the earlier burst's leftovers when the
outer invocation returns. At ≥ 2 bytes the assert fired — exit 134, and `guard`
cannot catch it. At exactly 1 byte the assert did not fire and the later burst's
byte was served ahead of the earlier one's leftovers, so the stream was silently
reordered instead: same root cause, two very different symptoms.

All three fills now go through one `takeFirstBufferingRest`, which concatenates
in chronological order and hands out the front byte — not necessarily `bytes[0]`.
No abort, and bytes come out in the order `read!` produced them. Re-entrancy
stays supported, as it already was for `write!`, `flush` and `close`, and the fd
path keeps its allocation-free single-byte fast path.

Verified against the issue's own repro (exit 134 → `66`) and its one-byte
reordering sibling (`Z B` → `B Z`).

## #2000 — a callback blocking on a channel, join, mutex or condvar

`vm.in_custom_port_callback` was read at exactly two sites, `waitForFd` and
`thread-sleep!`. Every other blocking primitive — `channel-receive`,
`channel-send`, `fiber-join`, and SRFI-18's `thread-join!`, `mutex-lock!`,
`condition-variable-wait` — drove the scheduler recursively on the native stack
instead, running whole sibling fibers to completion *inside* the callback.

Nesting grows one drive per level, so n fibers reading one such port died with
SIGBUS at n = 2500 while n = 2400 was clean — far short of `callReentrant`'s
`max_native_depth = 3000`, which is calibrated for a plain re-entrant call rather
than a nested `runUntil` **plus** a drive.

The check now also sits in `fiber.runSchedulerStep`, the single shared body
behind every in-place drive, so a blocking primitive added later is covered
without anyone having to list it — the issue's own suggestion. The two early
checks stay, each having state (a registered fd, an armed timer) it is cheaper
never to arm than to unwind. The guard restores `me` the way the normal epilogue
does before raising; that part is defensive and the comment says so — no program
was found that observes its absence.

A/B on the issue's repro: pre-fix `Bus error`, exit 134; post-fix a catchable
error, exit 0.

## Tests

- `tests/scheme/audit/primitives_fiber-audit.scm` — the three assertions #2000
  disabled are re-enabled, plus `mutex-lock!` (the SRFI-18 side of the same choke
  point) and the two timed variants, which arm scheduler state before the wait.
  114 → 117.
- `tests/scheme/audit/primitives_srfi181-audit.scm` — #1939's disabled assertion
  is re-enabled and corrected (it asserted `'(65 66)` against a one-element
  result), plus the one-byte reordering case and a textual counterpart. 215 → 218.
- `tests/scheme/srfi/srfi181-callback-drive-2000.scm` (new) — the deep-nesting
  case the audit deliberately left out because it took the runner down with it.
  It still aborts pre-fix (exit 134) and passes post-fix in 17 ms.

`zig build test`, `bash tests/scheme/run-all.sh` (676 files, R7RS 1395) and the
`format` job's three checks all pass. The four affected suites also pass under
`-Dgc-stress=true`.

Closes #1939
Closes #2000


src/primitives_io.zig
```diff
@@ -1402,14 +1453,7 @@ fn readOneByteFromCustomPort(port: *types.Port, cb: *types.CustomBacking) Primit
...
 
-    const first = bytes[0];
-    if (bytes.len > 1) {
-        std.debug.assert(port.read_buf == null);
-        const rest = gc.allocator.dupe(u8, bytes[1..]) catch return PrimitiveError.OutOfMemory;
-        port.read_buf = rest;
-        port.read_buf_len = rest.len;
-    }
-    return first;
+    return takeFirstBufferingRest(port, bytes);
```
baijum, line 1472: **Bug (confidence ~90): EOF over re-entrant leftovers**

This call only runs when `n > 0`. Just above (unchanged, so not in the diff) the custom-port path still does:

```zig
if (n == 0) return null; // EOF
```

After a nested `read!` leaves bytes in `port.read_buf`, that early return reports EOF while chronologically earlier bytes sit in the buffer. The next call drains them — a spurious EOF / silent reorder, the same class of bug as the one-byte case this PR fixes for `n >= 1`.

Minimal repro (same nesting as the new one-byte test, but outer returns `0`):

```scheme
(let ((self #f) (depth 0))
  (set! self (make-custom-binary-input-port "eof-reent"
    (lambda (bv start count)
      (set! depth (+ depth 1))
      (cond ((= depth 1)
             (read-u8 self) ; inner fill leaves a leftover in read_buf
             (set! depth (- depth 1))
             0)            ; outer reports EOF
            (else
             (set! depth (- depth 1))
             (bytevector-u8-set! bv start 65)
             (bytevector-u8-set! bv (+ start 1) 66)
             2)))
    #f #f #f))
  (read-u8 self)) ; should be 66, currently EOF; a second read then yields 66
```

**Suggested fix:** route EOF through the helper too:

```zig
if (n == 0) return takeFirstBufferingRest(port, &.{});
```

and teach the helper that an empty `bytes` still serves `pending` before returning `null` (see sibling comment). Please add a Scheme assertion next to the existing one-byte reordering test.

src/primitives_io.zig
```diff
@@ -1225,6 +1225,68 @@ fn flushCustomPortIfNeeded(vm: *vm_mod.VM, cb: *types.CustomBacking) PrimitiveEr
...
+/// Scheme, uncatchable by `guard` (#1939) — and in the one-byte case where
+/// the assert happened not to fire, the later burst's byte was served
+/// ahead of the earlier burst's leftovers, silently reordering the stream.
+/// Concatenating handles both: no abort, and bytes come out in the order
+/// read! produced them.
+fn takeFirstBufferingRest(port: *types.Port, bytes: []const u8) PrimitiveError!?u8 {
+    // Every caller already maps its own empty-burst case to EOF before
+    // getting here; treating it as EOF rather than asserting keeps this
+    // helper total, which is the whole point of replacing the asserts.
+    if (bytes.len == 0) return null;
```
baijum: Related to the EOF-over-reentrancy bug on the custom-port path: this early exit is what would still make `takeFirstBufferingRest(port, &.{})` wrong if used as the fix without a change here.

Pending should be considered before treating an empty burst as EOF:

```zig
const old = port.read_buf;
const pending: []const u8 = if (old) |rb| rb[rb.len - port.read_buf_len ..] else &.{};
if (bytes.len == 0) {
    if (pending.len == 0) {
        // free stale empty buf if any, then EOF
        ...
        return null;
    }
    // serve pending[0], rebuffer pending[1..] with nothing appended
    ...
}
```

That keeps the helper total for every fill site and makes the custom-port EOF path a one-liner.

baijum (commented):
Reviewed the #1939 / #2000 fix end-to-end (buffer helper, scheduler choke-point guard, and the re-enabled / new Scheme coverage).

## Verdict: changes needed (posted as COMMENT — GitHub disallows REQUEST_CHANGES on own PRs)

The #2000 half looks solid: putting the guard in `runSchedulerStep` is the right choke point, the early `waitForFd` / `thread-sleep!` checks stay for good reason, and the defensive `me.status = .running` / timer unwind is warranted given callers that arm state before the drive. Tests for channel/join/mutex plus the timed unwind and the deep-nesting file are the right shape.

The #1939 half mostly lands — `takeFirstBufferingRest` correctly concatenates in chronological order and removes the uncatchable assert — but one sibling of the same root cause is still open (inline on the custom-port fill path).

## What looks good
- Shared helper used by all three fill sites (fd / custom / transcode)
- Allocation-free single-byte fast path preserved when `read_buf` is empty
- Audit assertions re-enabled and corrected; deep-nesting case isolated so a regression aborts loudly again

baijum:
Both review findings addressed in bbdce7aa.

**@baijum — EOF over re-entrant leftovers: confirmed, and it was a real miss.**
Verified against the committed state rather than the description: at `f92df40`
the repro gives `(a #<eof> eof? #t)` then `(b 66)`. The first commit routed every
*non-empty* fill through `takeFirstBufferingRest` but left all three `return null`
EOF exits alone, so it fixed two of the three sizes of one bug and shipped the third.

Extended past the custom-port path: the fd burst's `if (raw == 0) return null` and
the transcoded character's `orelse return null` had the same shape. I could not
construct a program that reaches either with a non-empty `read_buf` — both need
re-entrancy on their own port, and the transcoded fill is always byte-balanced
within one `read-char` — so I'm not claiming they were live. They route through the
helper anyway, and the comment says which two are unreached. Three fills that look
identical while one silently isn't is what produced this.

Also kept the explicit `rest_len == 0` branch: `alloc(u8, 0)` would leave `read_buf`
non-null at `read_buf_len == 0`, and the drain falls through that state into a
spurious extra `read!` call.

**@coderabbitai — `thread-join!` / `condition-variable-wait`: valid.** The doc named
both as covered and tested neither. Both are reachable, so I added tests rather than
removing the references:

- condvar is spelled `(mutex-unlock! m cv timeout)` in SRFI 18 — rejected correctly.
- Only `thread-join!`'s **fiber** path drives; the OS-thread path joins the pthread
  directly. That is exactly why this suite's existing control "a `read!` that joins a
  short-lived thread is NOT rejected" keeps passing, so both directions are now pinned.
  The target must have *started*: `thread-join!` on a `spawn`'d fiber still in
  `.created` spins in its own `sleepNs` poll without ever driving the scheduler and
  hangs. That reproduces with no custom port involved, so it is unrelated to this PR
  and left alone — noted at the test.

Audit suites 117 → 119 (fiber) and 218 → 219 (srfi181). The EOF test fails against
the pre-fix binary and passes now; the two D5 additions pass either way, being
coverage for an untested claim rather than regression tests. `zig build test`,
`bash tests/scheme/run-all.sh` (676 files, R7RS 1395), `zig fmt --check` and
markdownlint all clean.

Issue #1731, opened 2026-07-23
Split src/types.zig (1627 lines, over the 1500-line policy)


baijum:
## Summary

\`src/types.zig\` is 1627 lines, over the project's 1500-line file-size
policy (\`CLAUDE.md\`: "Keep source files under 1500 lines. When a file
grows past that, split it along natural seams."). Flagged by CodeRabbit
during review of #1730 (SRFI 181 custom ports), which added a further 22
lines (\`CustomBacking\` struct + \`Port.custom_backend\` field).

**This predates #1730**: \`types.zig\` was already 1605 lines on \`main\`
before that PR touched it — over the limit already. This issue tracks the
pre-existing violation as its own focused task rather than bundling a
structural refactor into a feature PR, matching this project's general
practice of scoping large/independent changes into their own PRs (see
e.g. #1729, split off the same way from #1730 for an unrelated reason).

## Why this wasn't fixed as part of #1730

- The violation predates that PR; SRFI 181 didn't introduce the underlying
  problem, just nudged an already-over-limit file slightly further over.
- A proper split is what CodeRabbit itself labeled a "heavy lift": every
  heap type in this file (\`Port\`, \`Continuation\`, \`HashTable\`,
  \`FfiLibrary\`, \`Channel\`, ~30 more) is referenced from primitives files,
  \`memory.zig\`, \`gc_collect.zig\`, every \`vm*.zig\` file, \`printer.zig\`, and
  more — dozens of call sites across the codebase. Doing this carefully
  (likely via a re-export hub in \`types.zig\` itself, mirroring
  \`compiler_forms.zig\`'s "thin re-export hub, don't add logic there"
  pattern, so far-flung callers don't all need updating) deserves its own
  focused review, not a rushed addition to an already-large feature PR's
  diff.

## Context for whoever picks this up

- \`types.zig\`'s own documented expected size in \`CLAUDE.md\`'s file
  organization table is "~500" lines — it has grown ~3x past that
  organically, one heap type addition at a time, across 105 commits in
  its history. This is the same pattern that will keep recurring unless
  addressed once, structurally.
- The file is currently a flat list of ~55 independent struct/enum
  definitions (\`ObjectTag\`, \`Object\`, \`Value\`-adjacent helpers, then one
  struct per heap type) plus the \`OpCode\` enum near the end. Most structs
  don't reference each other's internals, only via \`Value\`/pointer fields
  — similar in spirit to how "primitives files are intentionally broad...
[...]

PR #2195, opened 2026-08-03
Fix four SRFI-69 hashing defects: deep keys, a self-mutating hash procedure, and negative hashes

### LINKED ISSUE #2023: [SRFI-69] equal? tables cannot find any key nested deeper than 8 — the depth cutoff hashes a pointer (the unfixed half of #1180)
**Category:** SRFI
**Spec reference:** SRFI 69, *Hashing* — "a hash funtion `hash` is acceptable iff `(e obj1 obj2)` → `(= (hash obj1) (hash obj2))`." The default table's `e` is `equal?`.
**Severity:** wrong-result (silent data loss)

`valueHashDepth` (`src/primitives_hashtable.zig:333`) stops recursing at `MAX_HASH_DEPTH = 8` and returns **the key's pointer** for whatever sits at the cutoff:

```zig
if (depth >= MAX_HASH_DEPTH) return @truncate(key *% 2654435761);
if (types.isPair(key)) { ... }
```

`findKey`/`findSlot` compare with `primitives.deepEqual` (structural), so two `equal?` keys whose structure reaches depth 8 hash to unrelated buckets and the stored entry becomes unreachable. A list of **9** cons cells is enough: cell #9 sits at depth 8.

**Minimal reproduction:**

```scheme
(define h (make-hash-table))
(do ((i 0 (+ i 1))) ((= i 32)) (hash-table-set! h i i))   ; grow past capacity 8
(hash-table-set! h (list 1 2 3 4 5 6 7 8 9) 'v)
(hash-table-ref/default h (list 1 2 3 4 5 6 7 8 9) 'MISSING)
;; expected: v    actual: MISSING
(equal? (list 1 2 3 4 5 6 7 8 9) (list 1 2 3 4 5 6 7 8 9))  ;; => #t
```

At scale it is total, not occasional — 200 twelve-element keys inserted, then looked up with freshly-built `equal?` keys: **200 of 200 unfindable**.

**Discriminating controls** (three, all run):

1. **The cliff is exactly `MAX_HASH_DEPTH`.** Same program with an **8**-element key ⟹ found; 0 of 200 misses at scale. `(= (hash k1) (hash k2))` for `equal?` keys is `#t` at 1–8 elements and `#f` at 9, 10, 12, 20.
2. **The growth step is what unmasks it.** With one entry in a capacity-8 table the lookup succeeds: heap pointers are 8-byte aligned and `ptr *% odd` is ≡ 0 (mod 8), so every pointer-hashed key lands in the same low bucket. The first grow widens the mask past the alignment and the buckets diverge. (Same masking `#1180` described.)
3. **Only the native hash is wrong — the same table with a Scheme-side hash is correct.** `(srfi 125)`'s `make-equal-comparator` has `equal?` as its equality, so `detectMode` picks `.equal` and the table uses `valueHash`, ignoring the comparator's own `default-hash` ⟹ 13/100 misses. `make-default-comparator`'s equality is a Scheme lambda, so the table falls into `.custom` and actually calls `default-hash` ⟹ **0/100** misses. Identical keys, identical library, identical table type.

**Blast radius** (all measured on this build, 10-element keys):

| Entry point | misses |
|---|---|
| `(srfi 69)` `make-hash-table` — the default | 200/200 |
| `(srfi 146 hash)` `hashmap` over `make-equal-comparator` | 90/100 |
| `(srfi 126)` `(make-hashtable equal-hash equal?)` | 88/100 |
| `(srfi 125)` `make-hash-table` over `make-equal-comparator` | 13/100 |

**This is the unimplemented half of closed #1180.** That issue's own text named it — "The same fallback fires for pairs/vectors nested deeper than `MAX_HASH_DEPTH = 8`, so deep structural keys break identically" — and its own suggested fix said "make the depth cutoff return a constant instead of hashing the sub-pointer." The bignum/rational/complex half shipped (`valueHashDepth:308-332`); the depth cutoff did not. The regression test that re-enabled #1180 (`tests/scheme/audit/primitives_hashtable-audit.scm:65-83`) covers only bignum and rational keys, and every other hash-consistency assertion in that file stays at depth ≤ 2 — which is why a green suite never caught it.

Returning a constant at the cutoff costs collisions (bounded by `deepEqual` still deciding equality) and restores the invariant.

**Verified:** fresh ReleaseSafe build at `d0d03deb`, isolated `KAAPPI_HOME`; behaviour **unchanged** under `zig build -Dgc-stress=true` (same miss, same cliff) — it is a hashing bug, not a rooting bug.
**Source location:** `src/primitives_hashtable.zig:333` (`valueHashDepth`, the `depth >= MAX_HASH_DEPTH` arm)
**Disabled test:** `tests/scheme/audit/primitives_hashtable-audit.scm` (marked `;; FAIL: #this`)

**Found by:** Systematic audit v2, Phase 2.8

### LINKED ISSUE #2024: [SRFI-69] A custom hash procedure that inserts into its own table aborts the process silently — rehash holds a freed entry array across the callback
**Category:** SRFI / GC
**Spec reference:** SRFI 69 — `make-hash-table` takes "`hash`, a hash function" supplied by the caller; nothing restricts what that procedure may do. Memory safety is not the caller's to guarantee.
**Severity:** crash (silent process abort, no diagnostic at all)

`rehash` (`src/primitives_hashtable.zig:424`) captures the old entry array and then calls back into Scheme **inside the loop that iterates it**:

```zig
const old_entries = ht.entries;
const old_cap = ht.capacity;
for (old_entries[0..old_cap]) |entry| {
    if (entry.state == .occupied) {
        const h = hashForTable(proc, ht, entry.key) catch ...;   // <-- calls the user's hash proc
        ...
    }
}
ht.entries = new_entries;
memory.freeSliceNoFill(gc.allocator, HashEntry, old_entries);
```

`hashForTable`'s `.custom` arm runs `vm.callWithArgs(ht.hash_fn, ...)`. If that procedure inserts into **its own** table, the nested `hash-table-set!` reaches `growIfNeeded` → a nested `rehash`, which reads `ht.entries` (still `old_entries`, deliberately — see the comment at line 433), swaps in its own array, and **frees `old_entries`**. The outer loop then keeps iterating freed memory, finally overwrites the inner rehash's `ht.entries` with its own stale array and frees `old_entries` a **second** time.

**Minimal reproduction:**

```scheme
(import (scheme base) (scheme write) (srfi 69))
(define h #f) (define armed #t)
(define (myhash k)
  (when (and armed h)
    (set! armed #f)
    (do ((j 0 (+ j 1))) ((= j 5)) (hash-table-set! h (+ 100 j) j))
    (set! armed #t))
  (if (number? k) k 0))
(set! h (make-hash-table = myhash))
(do ((i 0 (+ i 1))) ((= i 5)) (hash-table-set! h i i))
(display "survived")
```

```text
$ kaappi repro.scm ; echo "exit=$?"
exit=133
```

**Empty stdout and empty stderr.** No `KP` code, no panic text, no trace. 5/5 runs. A 4x4 sweep of (outer inserts, injected inserts) over {5,10,20,40}x{5,10,20,60} aborts in 12 of 16 cells (exit 133 `SIGTRAP` or 134 `SIGABRT`).

**Discriminating controls** (three, all run, all exit 0):

1. **Same custom hash mutating a *different* table** ⟹ `h=5 g=5`, clean. So it is the self-reference, not custom hashes as such.
2. **A custom *equality* mutating its own table** ⟹ `5`, clean. `rehash` calls only `hashForTable`, never `equalForTable`, so the equality callback never runs from inside the loop that owns the freed slice.
3. **Unbounded re-entry is already handled correctly** — an unguarded `(hash-table-set! h -1 -1)` in the hash proc raises `KP3008: native re-entrancy too deep` and exits 1. So a depth guard exists; it just does not cover the case where the re-entry *returns* and the caller resumes over freed memory.

**Second manifestation of the same root cause — silent, no crash.** `hashTableMergeFn:870` iterates `ht2.entries[0..ht2.capacity]` across `findSlot` on `ht1`, which reaches the same two callbacks. With a custom hash that mutates `h2`, merging a 10-entry `h2` into an empty `h1` leaves `h1` with **1** entry and exits 0.

`hash-table-walk` and `hash-table-fold` are **not** affected: #1181's fix gave them `snapshotLiveEntries` plus `extra_roots`, and they were verified safe here under insert-during-walk, delete-during-walk, raise, `call/cc` escape, and nested re-entry. `rehash` and `hashTableMergeFn` are the two loops that still hold a live `entries` slice across a Scheme call — the same shape closed #690 fixed for walk/fold, at two sites it did not reach.

**Verified:** fresh ReleaseSafe build at `d0d03deb`, isolated `KAAPPI_HOME`; **unchanged** under `zig build -Dgc-stress=true` (3/3 exit 133, controls still clean) — consistent with an allocator double-free/UAF rather than a GC rooting fault.
**Source location:** `src/primitives_hashtable.zig:424-455` (`rehash`), `:870-885` (`hashTableMergeFn`)
**Disabled test:** `tests/scheme/audit/primitives_hashtable-audit.scm` (marked `;; FAIL: #this`; kept commented out because the failure mode is a process abort, which no `guard` can contain)

**Found by:** Systematic audit v2, Phase 2.8

### LINKED ISSUE #2025: [SRFI-69] hash / string-hash / string-ci-hash / hash-by-identity return negative integers — a 62-bit mask through a 48-bit makeFixnum
**Category:** SRFI
**Spec reference:** SRFI 69, `hash` — "Produces a hash value for `object` in the range **( 0, bound (**. If `bound` is not given, the implementation is free to choose any bound, given that the default bound is greater than the size of any imaginable hash table in a normal application." A negative result is outside every such range.
**Severity:** wrong-result

All four hash procedures end with the same two lines:

```zig
return types.makeFixnum(@intCast(h & 0x3FFFFFFFFFFFFFFF));
```

The mask keeps **62** bits; `makeFixnum` keeps **48** and `toFixnum` sign-extends from `i48`. Any hash with bit 47 set therefore comes back negative — roughly half of all inputs.

**Minimal reproduction:**

```scheme
(import (scheme base) (scheme write) (srfi 69))
(string-hash "aaaaaaaaaaaa")   ;; expected: >= 0    actual: -9255028371264
(string-hash "bbbbbbbbbb")     ;; actual: -137297362969280
```

`stringHashFn` computes `h` in an explicit `u64` (`h = h *% 31 +% c`) with no pointer and no `usize`, so these two values are the same on every supported target — they are safe to assert. 27 of the 59 strings `"a"`, `"aa"`, … `"a"x59` hash negative. Across a 2500-call sweep of `hash`, `string-hash`, `string-ci-hash` and `hash-by-identity`, **178** results were negative.

**Discriminating control:** the **bounded** form is correct. `@mod(h, bound)` is computed on the full `u64` before the fixnum conversion, so the result is always small enough to survive it — 0 out of 199 `(string-hash s 97)` and 0 out of 199 `(hash s 97)` calls fell outside `[0, 97)`. The defect is exactly the unbounded arm's 62-bit mask, not the hashing.

**Affected, one root cause, four sites:** `hashFn:757`, `stringHashFn:775`, `stringCiHashFn:807`, `hashByIdentityFn:819`.

**Blast radius beyond `(srfi 69)`:** `lib/srfi/126.sld:385,389` define `equal-hash` and `symbol-hash` as direct wrappers over this `hash`, so SRFI 126's own hash functions inherit it. `(srfi 128)` is **not** affected — it defines its own portable `string-hash` (`lib/srfi/128.sld:56`) rather than re-exporting the native one.

SRFI 69's own tables are unaffected: `findKey`/`findSlot` use the internal `usize` hash directly and never go through these procedures. The damage is to callers who use `hash` for their own bucketing, and to any comparator built on it — SRFI 128 requires a comparator's hash function to return an exact non-negative integer.

Prior art for the same mechanism, all closed: #46 (FFI returns > 2^47), #31 (`file-info` inode/device), #16 (gcd/lcm). The pattern is `makeFixnum` on a value the caller never range-checked.

**Verified:** fresh ReleaseSafe build at `d0d03deb`, isolated `KAAPPI_HOME`; gc-stress not applicable (pure arithmetic, no allocation).
**Source location:** `src/primitives_hashtable.zig:757, 775, 807, 819`
**Disabled test:** `tests/scheme/audit/primitives_hashtable-audit.scm` (marked `;; FAIL: #this`)

**Found by:** Systematic audit v2, Phase 2.8

baijum:
Four defects in `src/primitives_hashtable.zig`, all breaking the same SRFI 69
rule from different directions: a hash must agree with the table's equality,
and must land in `[0, bound)`.

Closes #2023.
Closes #2024.
Closes #2025.

## #2023 — the depth cutoff hashed a pointer

`valueHashDepth` stopped at `MAX_HASH_DEPTH = 8` and returned the *pointer* of
whatever sat there, while `findKey`/`findSlot` compare with `deepEqual`. Two
`equal?` keys whose structure reached depth 8 hashed to unrelated buckets and
the stored entry became unreachable — measured on this build, 10-element keys:

| entry point | before | after |
|---|---|---|
| `(srfi 69)` `make-hash-table` | 200/200 missing | 0/200 |
| `(srfi 146 hash)` over `make-equal-comparator` | 90/100 | 0/100 |
| `(srfi 126)` `(make-hashtable equal-hash equal?)` | 88/100 | 0/100 |
| `(srfi 125)` over `make-equal-comparator` | 13/100 | 0/100 |

The cutoff now folds in a fixed sentinel. Separately, a list spine is walked
iteratively, so *length* no longer spends the budget that only *nesting*
should — without that, every list longer than 8 would collapse to one value.
Collisions replace the pointer, and `deepEqual` still decides equality.

**A fourth instance, found while fixing this one.** `deepEqual` compares SRFI
160 numeric vectors structurally (kind + raw bytes), but `valueHashDepth` had
no arm for them at all, so they fell through to the same pointer hash at depth
0: `(equal? (f64vector 1 2 3) (f64vector 1 2 3))` was `#t` while the two hashed
apart and 200/200 keys were unfindable. Same function, same invariant, no depth
involved — fixed here rather than filed separately.

## #2024 — rehash held a freed entry array across a Scheme callback

`rehash` captured the old entry array and then called the table's own hash
procedure from inside the loop iterating it. A hash procedure that inserted
into its *own* table reached a nested `rehash`, which swapped in its own array
and freed the one the outer loop was still walking, then freed it a second
time. Silent abort: exit 133/134, empty stdout and stderr.

`HashEntry` now caches the hash computed at insertion (`findSlot` already had
it), so `rehash` re-buckets from that and runs **no Scheme code at all** —
nothing can re-enter while `old_entries` is live. The field deliberately has no
default, so every insertion site must supply it.

`hash-table-merge!`, the second site named in the issue, walked ht2's live
array across `findSlot` on ht1 and left ht1 holding **1 of 10** entries. It now
iterates a rooted snapshot — the shape walk/fold got in #1181. A merged entry
is also re-bucketed under *ht1's* hash function rather than carried over with
the hash ht2 cached for it.

`findKey`/`findSlot` derive the mask only after the hash procedure returns
(it can grow the table) and copy entries out by value.

## #2025 — negative hashes

`hash`, `string-hash`, `string-ci-hash` and `hash-by-identity` masked to 62
bits and handed the result to `makeFixnum`, which keeps 48 and sign-extends
from `i48` — so any hash with bit 47 set came back negative, roughly half of
all inputs. Masking to 47 bits is the widest value that survives the round
trip. The bounded arm was already correct and is untouched. `(srfi 126)`'s
`equal-hash`/`symbol-hash`, which wrap these directly, are fixed with them.

## Second commit: a duplicate struct definition

`types_hashtable.zig` was added by the #1731 domain split but never wired up —
nothing imported it, so the byte-identical copy in `types.zig` was what the
build compiled and the domain file was never type-checked. Adding
`HashEntry.hash` meant editing both by hand with no compile error if they
drifted. `types.zig` now re-exports the four names the way it does for every
other `types_*.zig`, which is what CLAUDE.md already documents. No behaviour
change.

## Tests

All four `;; FAIL: #this` markers in
`tests/scheme/audit/primitives_hashtable-audit.scm` are cleared and the
assertions behind them re-enabled, joined by new cells for the numeric-vector
instance, a discriminating merge repro, and coverage for the caching mechanism
itself (growth, copy, and re-bucketing under the target table's hash function).
216 pass, 0 fail.

Every re-enabled assertion was A/B'd against a binary built from the pre-fix
tree: the #2023 cells FAIL, the #2024 cell aborts the process mid-suite, merge
leaves h1 with 1 of 10, and the unbounded hashes come back negative.

- `zig build test` — green
- `zig build test -Dgc-stress=true` — green
- `bash tests/scheme/run-all.sh` — 2070 pass; the 3 WASM tier divergences it
  reports (`deep-print-depth.scm`, `peek-char-fd-utf8-798.scm`,
  `peek-char-utf8.scm`) reproduce identically with this branch's `src/` stashed
  and are not in the harness's `KNOWN_DIFFS` — pre-existing, unrelated, filed
  separately.
- Throughput unchanged: 3 runs each of 120k insert+lookup over string keys
  (40.3/40.7/40.3s before → 38.7/38.7/38.7s after) and 40k over list keys
  (77.2/77.1/77.0s → 73.2/73.2/73.3s). Collapsing the `.equal` fast path into
  the generic probe costs nothing; not re-hashing on growth pays slightly.

### DIFF
--- a/src/gc_alloc.zig
+++ b/src/gc_alloc.zig
@@ -1150,7 +1150,7 @@ pub fn allocHashTable(self: *GC, initial_capacity: usize) !Value {
     const entries = try allocSliceNoFill(self.allocator, HashEntry, cap);
     errdefer freeSliceNoFill(self.allocator, HashEntry, entries);
     for (entries) |*e| {
-        e.* = .{ .key = 0, .value = 0, .state = .empty };
+        e.* = .{ .key = 0, .value = 0, .hash = 0, .state = .empty };
     }
     const ht = try self.allocator.create(HashTable);
     ht.* = .{
--- a/src/gc_deep_copy.zig
+++ b/src/gc_deep_copy.zig
@@ -249,6 +249,9 @@ fn deepCopyValue(gc: *GC, src: Value, visited: *std.AutoHashMap(usize, Value)) D
                         new_ht.entries[i] = .{
                             .key = try deepCopyValue(gc, entry.key, visited),
                             .value = try deepCopyValue(gc, entry.value, visited),
+                            // Slot positions are preserved, so the source's
+                            // cached hash stays the right one for this slot.
+                            .hash = entry.hash,
                             .state = .occupied,
                         };
                         new_ht.count += 1;
@@ -261,11 +264,12 @@ fn deepCopyValue(gc: *GC, src: Value, visited: *std.AutoHashMap(usize, Value)) D
                     if (entry.state == .occupied) {
                         const nk = try deepCopyValue(gc, entry.key, visited);
                         const nv = try deepCopyValue(gc, entry.value, visited);
-                        var idx = hashtable.hashForMode(new_ht.compare_mode, nk) & (new_ht.capacity - 1);
+                        const nh = hashtable.hashForMode(new_ht.compare_mode, nk);
+                        var idx = nh & (new_ht.capacity - 1);
                         while (new_ht.entries[idx].state == .occupied) {
                             idx = (idx + 1) & (new_ht.capacity - 1);
                         }
-                        new_ht.entries[idx] = .{ .key = nk, .value = nv, .state = .occupied };
+                        new_ht.entries[idx] = .{ .key = nk, .value = nv, .hash = nh, .state = .occupied };
                         new_ht.count += 1;
                     }
                 }
--- a/src/gc_deep_copy.zig
+++ b/src/gc_deep_copy.zig
@@ -269,7 +269,7 @@ fn deepCopyValue(gc: *GC, src: Value, visited: *std.AutoHashMap(usize, Value)) D
                         while (new_ht.entries[idx].state == .occupied) {
                             idx = (idx + 1) & (new_ht.capacity - 1);
                         }
-                        new_ht.entries[idx] = .{ .key = nk, .value = nv, .hash = nh, .state = .occupied };
+                        new_ht.entries[idx] = .{ .key = nk, .value = nv, .hash = @truncate(nh), .state = .occupied };
                         new_ht.count += 1;
                     }
                 }
--- a/src/primitives_hashtable.zig
+++ b/src/primitives_hashtable.zig
@@ -387,48 +387,175 @@ fn valueHashDepth(key: Value, depth: usize) usize {
     return @truncate(key *% 2654435761);
 }
 
-fn findKey(proc: []const u8, ht: *HashTable, key: Value) PrimitiveError!?usize {
-    if (ht.capacity == 0) return null;
-    // A `.custom` hash procedure can re-enter and grow the table, so the mask
-    // is derived only after it has returned. `ht.entries` is re-read on every
-    // probe and entries are copied out by value, so an equality procedure that
-    // does the same cannot leave this loop reading a freed array.
-    const h = try hashForTable(proc, ht, key);
-    const mask = ht.capacity - 1;
-    var idx = h & mask;
-    var probes: usize = 0;
-    while (probes < ht.capacity) {
-        const entry = ht.entries[idx];
-        if (entry.state == .empty) return null;
-        if (entry.state == .occupied and try equalForTable(proc, ht, entry.key, key)) return idx;
-        idx = (idx + 1) & mask;
-        probes += 1;
+/// How many times a probe may restart because the table moved under it before
+/// we give up. A `.custom` procedure that mutates on *every* call would
+/// otherwise spin forever; anything sane settles in one.
+const MAX_PROBE_RESTARTS = 16;
+
+/// A probe's view of the table. `.custom` hash and equality procedures are
+/// arbitrary Scheme, and one that inserts into the table being probed makes
+/// `rehash` install a new, larger `entries` — which invalidates both the mask
+/// the probe is stepping with and any index it has already chosen. There is no
+/// use-after-free (`ht.entries` is re-read and entries are copied out by
+/// value); the defect is index staleness. A slot picked under the old layout
+/// names an arbitrary bucket in the new one, so the caller's write lands on a
+/// live entry for a different key while `count` is still incremented — one key
+/// silently destroyed, and `count` no longer matching the live set.
+const Layout = struct {
+    entries: [*]HashEntry,
+    capacity: usize,
+
+    fn of(ht: *HashTable) Layout {
+        return .{ .entries = ht.entries.ptr, .capacity = ht.capacity };
     }
-    return null;
+
+    fn movedSince(self: Layout, ht: *HashTable) bool {
+        return ht.entries.ptr != self.entries or ht.capacity != self.capacity;
+    }
+};
+
+fn probeRestartExhausted(proc: []const u8) PrimitiveError {
+    return primitives.argError(
+        proc,
+        "hash table kept being modified by its own hash or equality procedure during a single lookup",
+        .{},
+    );
 }
 
-const FindSlotResult = struct { idx: usize, found: bool, hash: usize };
+fn findKey(proc: []const u8, ht: *HashTable, key: Value) PrimitiveError!?usize {
+    if (ht.capacity == 0) return null;
+    // The mask is derived only after the hash procedure has returned, since it
+    // too can grow the table.
+    // `.equal` is the default mode and by far the hottest; calling `valueHash`
+    // and `deepEqual` directly rather than through the dispatchers keeps them
+    // inlinable. Routing it through `hashForTable`/`equalForTable` instead --
+    // whose `.custom` arms pull in the whole VM-call path -- cost 18% on
+    // insert and 10% on lookup in `benchmarks/hashtable.scm`.
+    if (ht.compare_mode == .equal) {
+        const mask = ht.capacity - 1;
+        var idx = valueHash(key) & mask;
+        var probes: usize = 0;
+        while (probes < ht.capacity) : (probes += 1) {
+            const entry = &ht.entries[idx];
+            if (entry.state == .empty) return null;
+            if (entry.state == .occupied and primitives.deepEqual(entry.key, key)) return idx;
+            idx = (idx + 1) & mask;
+        }
+        return null;
+    }
+    const h = try hashForTable(proc, ht, key);
+    if (ht.compare_mode != .custom) {
+        // No Scheme runs in this loop either, so the layout cannot move under
+        // it: probe in place and skip the restart bookkeeping.
+        const mask = ht.capacity - 1;
+        var idx = h & mask;
+        var probes: usize = 0;
+        while (probes < ht.capacity) : (probes += 1) {
+            const entry = &ht.entries[idx];
+            if (entry.state == .empty) return null;
+            if (entry.state == .occupied and try equalForTable(proc, ht, entry.key, key)) return idx;
+            idx = (idx + 1) & mask;
+        }
+        return null;
+    }
+    var restarts: usize = 0;
+    restart: while (true) {
+        const layout = Layout.of(ht);
+        const mask = layout.capacity - 1;
+        var idx = h & mask;
+        var probes: usize = 0;
+        while (probes < layout.capacity) : (probes += 1) {
+            const entry = ht.entries[idx];
+            if (entry.state == .empty) return null;
+            if (entry.state == .occupied) {
+                const eq = try equalForTable(proc, ht, entry.key, key);
+                if (layout.movedSince(ht)) {
+                    restarts += 1;
+                    if (restarts > MAX_PROBE_RESTARTS) return probeRestartExhausted(proc);
+                    continue :restart;
+                }
+                if (eq) return idx;
+            }
+            idx = (idx + 1) & mask;
+        }
+        return null;
+    }
+}
+
+/// `hash` is the `u32` an insertion caches (see `HashEntry.hash`), already
+/// narrowed here so every insertion site stores it without a cast of its own.
+const FindSlotResult = struct { idx: usize, found: bool, hash: u32 };
 
 fn findSlot(proc: []const u8, ht: *HashTable, key: Value) PrimitiveError!FindSlotResult {
-    const h = try hashForTable(proc, ht, key);
-    const mask = ht.capacity - 1;
-    var idx = h & mask;
-    var first_tombstone: ?usize = null;
-    var probes: usize = 0;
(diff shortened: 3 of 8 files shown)


baijum:
Thanks — five of six were right, and one of them found a defect worth more than the rest of the review put together.

**Stale `slot.idx` (`src/primitives_hashtable.zig`) — fixed.** I had reached the same conclusion independently and deferred it, which was the wrong call. Your framing is exactly right: no use-after-free, the defect is index staleness. I swept it as (prefill × inject × span) and it reproduced in **128 of 210** combinations — one pre-existing key silently destroyed, `count` no longer matching the live key list. All 210 are clean now. The probe restarts when it observes the layout move, bounded at 16 so a procedure that mutates on every call raises rather than spins. `findKey` had the same hole and got the same treatment. Regression test added.

**Neither cutoff was covered — fixed.** Correct, and it was the more embarrassing miss: `mk` builds a flat list, and the whole point of the fix is that a flat list *no longer* spends the nesting budget, so `DEEP_CUTOFF_HASH` was unreachable from any new assertion. Added a `nest` helper for keys deeper than `MAX_HASH_DEPTH`, plus a case whose only distinguishing element sits past `MAX_HASH_SPINE` so all 50 keys hash alike and `deepEqual` has to separate them.

**`(>= (hash-table-size h) 5)` too weak — fixed**, and it's the same defect as the first finding, so the strengthened assertion now fails without that fix.

**Duplicate test — fixed** by making it the nesting case rather than deleting it, which also closes part of finding 2.

**Test name — fixed.** You're right that 24 bytes and 3 bytes are not the same length, and that the body asserts retrieval rather than hash inequality. Renamed to what it actually checks.

**`(srfi 160 f64)` availability — declining.** `lib/srfi/160/f64.sld` has no `cond-expand` gating, and the file already imports `(srfi 18)`, which is native-only — so the suite's availability profile is unchanged by this import, and the WASM differential harness already excludes the file on that basis.

One thing your review did not flag but CI did, worth recording here: this PR regressed `benchmarks/hashtable.scm` by 1.32x. Cause was routing the default `.equal` mode through `hashForTable`/`equalForTable`, whose `.custom` arms pull in the whole VM-call path and stop `valueHash`/`deepEqual` inlining — 18% on insert, 10% on lookup. The `.equal` specialization is restored and `HashEntry.hash` is now `u32`, keeping the struct at its original 24 bytes. 0.0302s vs 0.0306s on main.

Your suggested deep-copy assertion I implemented differently: asserting the copy preserves the *source* hash fails, because the non-custom arm deliberately re-hashes as it re-buckets. The invariant that matters is that the cached hash agrees with the *destination* table, so that is what it asserts now.

PR #2199, opened 2026-08-03
Stop compile-only commands running program code, and name the real cache reason

### LINKED ISSUE #2114: [Doc-truth] Eight top-level heads disable the .sbc cache, not just import — and --timings blames 'imports' for all of them
**Category:** Doc-truth / Tooling
**Spec reference:** `docs/dev/cache.md` § "What is and isn't cached" — "**Cached:** a plain `kaappi file.scm` run of a program that does **not** `import`. (Programs that import are skipped: library loading can free collected function pointers, so their top-level functions are not safe to serialize after the fact.)"
**Severity:** doc-truth

## Summary

The documented rule names one construct. The implemented rule names **eight**, and
the stated rationale covers only three of them. `--timings` then blames `imports`
for a file containing no `import` at all.

`src/main.zig:829` gates the cache write on `has_imports`, which
`src/main.zig:785` sets whenever **any** top-level form is claimed by
`vm_eval.handleTopLevelForm` (`src/vm_eval.zig:72`). That is:

```text
import   define-library   define-record-type   define-values
include  include-ci       begin               cond-expand
```

One occurrence of any of them, anywhere at top level, makes the **whole file**
uncacheable — every other form in it included. `define-record-type`,
`define-values`, a top-level `begin` and a top-level `cond-expand` have nothing
to do with library loading, so the source comment's rationale ("GC may have freed
collected function pointers during library loading") does not explain them.

## Minimal reproduction

```console
$ export KAAPPI_HOME=$(mktemp -d)

$ printf '(define-record-type <p> (mk a) p? (a p-a))\n(display (p-a (mk 5)))(newline)\n' > r.scm
$ kaappi --timings r.scm 2>&1 | grep cache
cache: MISS (not cached: imports)

$ printf '(define-values (a b) (values 1 2))\n(display (+ a b))(newline)\n' > d.scm
$ kaappi --timings d.scm 2>&1 | grep cache
cache: MISS (not cached: imports)

$ printf '(begin (display 1))\n(newline)\n' > b.scm
$ kaappi --timings b.scm 2>&1 | grep cache
cache: MISS (not cached: imports)

$ printf '(cond-expand (else (display 1)))\n(newline)\n' > c.scm
$ kaappi --timings c.scm 2>&1 | grep cache
cache: MISS (not cached: imports)
```

None of the four contains an `import`. The reason string comes from
`src/main.zig:836` (`timings.cacheReason("imports")`), which is fed by the same
misnamed flag.

## Discriminating control

The rule is **top-level head position only**. The same four constructs *nested*
inside a body leave caching alive — verified, all four in one file, which HITs on
the warm run:

```scheme
(define (a) (begin 1 2 3))
(define (b) (cond-expand (else 'chosen)))
(define (c) (define-values (x y) (values 10 20)) (+ x y))
(define (d) (define-record-type <pt> (mk x) pt? (x pt-x)) (pt-x (mk 7)))
```

So this is `handleTopLevelForm`'s dispatch set, not "these forms are unserialisable".

## Measurement

Over the differential harness's default corpus (`probes` + `smoke` + `compliance`
+ `audit`, 345 files at the time of measuring), one isolated `KAAPPI_HOME` per
file:

| | files |
|---|---:|
| populate the cache | **40** |
| do not | 305 |
| …of those, have a top-level `import` | 303 |
| …disabled by a top-level `begin` only | 1 (`probes/begin-simplify.scm`) |
| …disabled by a top-level `define-values` only | 1 (`smoke/define-values-lambda-body.scm`) |
| cached files containing any of the eight at top level | **0** |

The last row is the check that the rule has no false positives: not one file
caches despite containing one of the eight.

`import` does dominate the corpus (#1888 quantifies the cost of that half and is
the enhancement issue for fixing it). This issue is narrower and separable: the
documented rule, the `--timings` reason string, and the flag's own name should
say what the code does — and the four non-library constructs deserve a stated
reason, or a fix, rather than being covered by an `import`-shaped explanation.

## Suggested shape of a fix

1. `docs/dev/cache.md` § "What is and isn't cached": list the eight heads.
2. `timings.cacheReason` should name the head that actually disabled it
   (`not cached: top-level define-record-type`), which is one string away —
   `handleTopLevelForm` already knows the name.
3. Rename `has_imports` to something true (`has_toplevel_declaration`).
4. Separately, decide whether the four non-library heads need the gate at all.

**Verified:** fresh ReleaseSafe build at `3c1a56be`, macOS aarch64, one isolated
`KAAPPI_HOME` per file, cold and warm.

**Source location:**
- `src/main.zig:752,785,829,836` (`has_imports` and the reason string)
- `src/vm_eval.zig:72` (`handleTopLevelForm` — the eight heads)
- `docs/dev/cache.md` § "What is and isn't cached"

**Test:** `tests/scheme/differential/probes/sbc-population.scm` pins the rule and
its nested-form control; `run-differential.sh`'s header carries the measurement
and its "cache exercised: N / M" line makes a regression visible as the count
collapsing.

**Found by:** Systematic audit v2, Phase 4E (`.sbc` cache coverage).

### LINKED ISSUE #2156: [Tooling] kaappi --compile executes top-level cond-expand, begin and define-values bodies — a compile-only command deletes files
**Category:** Tooling
**Spec reference:** `CLAUDE.md` — "`kaappi compile <file> [-o output]` compiles to a native binary via LLVM"; `docs/dev/cache.md` — "`kaappi --compile file.scm [-o out.sbc]` writes an *explicit* bytecode artifact you named". Neither says compiling runs the program, and `kaappi check` is documented as the compile-only command that "executes nothing".
**Severity:** wrong-result (a compile-only command performs real filesystem side effects)

## Summary

`kaappi --compile` evaluates the bodies of top-level `cond-expand`, `begin` and
`define-values` forms **for real**, at compile time. A `delete-file` inside any
of them deletes the file while producing the `.sbc`. The same form is then
recorded in the artifact's preamble and re-executed at run time, so across
compile + run the effect happens twice.

## Minimal reproduction

```console
$ export KAAPPI_HOME=$(mktemp -d)
$ echo hi > /tmp/victim.txt
$ printf '(cond-expand (posix (delete-file "/tmp/victim.txt")))\n(display "ran")(newline)\n' > c.scm
$ kaappi --compile -o c.sbc c.scm
Compiled c.scm -> c.sbc
$ ls /tmp/victim.txt
ls: /tmp/victim.txt: No such file or directory
```

## Discriminating control

The same side effect as a *bare* top-level form is not executed — so this is not
"compiling runs the program", it is specific to the heads
`vm_eval.handleTopLevelForm` claims. Measured, one probe per row, each starting
from a fresh `/tmp/probe.victim`:

| Top-level form | file deleted by `--compile`? |
|---|---|
| `(delete-file "…")` | no |
| `(cond-expand (posix (delete-file "…")))` | **yes** |
| `(begin (delete-file "…"))` | **yes** |
| `(define-values (a) (values (delete-file "…")))` | **yes** |
| `(define-record-type <p> (mk a) p? (a p-a))` | no |
| `(define (f) (delete-file "…"))` | no |

## Cause

`src/main.zig:1055`, in the `--compile` collection loop, calls
`vm.handleTopLevelForm(expr)` — which *evaluates* the form — and then appends the
form's **source** to the preamble at `src/main.zig:1061` for replay at run time.
The eight heads `handleTopLevelForm` claims are `import`, `define-library`,
`define-record-type`, `define-values`, `include`, `include-ci`, `begin` and
`cond-expand` (`src/vm_eval.zig:72`; enumerated in #2114). Three of them can run
arbitrary user code, and do.

For `import`/`include`/`define-library` the evaluation is the point — the
compiler needs the bindings. For `begin`, `cond-expand` and `define-values` it is
an unintended side effect of routing everything through one dispatcher: only the
branch *selection* (or the value binding) is needed at compile time, not the
body's effects.

Confirmed that the artifact replays it too: `zig build -Dbundle=c.sbc -p /tmp/b`
then `/tmp/b/bin/kaappi` prints the `cond-expand` body's output once more.

**Verified:** fresh ReleaseSafe build at `910f7c75`, macOS aarch64, isolated
`KAAPPI_HOME`, cache cleared; not GC-related (`-Dgc-stress` not applicable —
no allocation-dependent behaviour involved).

**Source location:** `src/main.zig:1055` and `:1061`; `src/vm_eval.zig:72`.

**Related:** #2114 (same eight-head dispatcher, different defect: which files the
cache skips and what `--timings` calls the reason). #2096 (`--check` runs the
program) is a different mechanism — a fmt-scoped flag accepted globally.

**Found by:** Systematic audit v2, Phase 7D.

baijum:
Both issues are the same root cause: `vm_eval.handleTopLevelForm` claims **eight**
top-level heads — `import`, `define-library`, `define-record-type`,
`define-values`, `include`, `include-ci`, `begin`, `cond-expand` — and three
consumers each carried a hand-kept copy of that list. Both defects are a copy
disagreeing with what the dispatcher actually does.

## #2156 — `--compile` deletes files (priority: high, wrong-result)

`--compile` and `--disassemble` routed every claimed head through the
*evaluator*. Three of the eight carry ordinary program code, so a `delete-file`
inside a top-level `begin`, `cond-expand` or `define-values` ran for real while
the `.sbc` was produced — and the form was also recorded in the preamble, so
across compile + run the effect happened **twice**. Reproduced against
`07cceb25`; a bare top-level `(delete-file ...)` was never executed, which is
what makes this the dispatcher's fault rather than "compiling runs the program".

The fix splices `begin` / `cond-expand` into the driver's form stream
(`toplevel_driver.TopLevelForms`) so their bodies are **compiled** — only a
`cond-expand`'s branch *selection* is a compile-time question — and evaluates
only `TopLevelHead.isEnvSetup()`, the five declarations later forms are compiled
*against*. `define-values` is recorded for replay without running its producer.
`check.zig` already worked this way; `--compile` and `--disassemble` now share
its classification instead of having their own.

**`--disassemble` was not named in the issue** but had the identical defect in
the adjacent loop, so it is fixed the same way. Its bodies are now disassembled
rather than silently executed.

### A second defect the same fix closes

The preamble replays *entirely before* the compiled forms, so a top-level
`begin` also ran **out of program order** in the artifact. A/B against a
pre-fix build:

| `one / (begin two) / three` | interpreted | standalone binary |
|---|---|---|
| before | `one two three` | `two one three` |
| after | `one two three` | `one two three` |

## #2114 — `--timings` blames `imports` for all eight (priority: low, doc-truth)

All eight heads legitimately disable the `.sbc` cache, but not for the reason
the code and docs gave. The real invariant: `handleTopLevelForm` consumes such a
form and appends no `Function`, so a HIT — which compiles nothing — would skip
its work entirely. The library-loading/GC rationale is real but covers only the
three heads that load libraries. `--timings` now names the head that fired:

```text
cache: MISS (not cached: top-level define-record-type)   # was: imports
```

`has_imports` is gone in favour of `first_toplevel_decl: ?TopLevelHead`, which
both gates the write and supplies the reason, so the two can't disagree.
`docs/dev/cache.md` states the shared reason and the top-level-head-position
rule; `docs/dev/timings.md`, the differential harness header and two probe
headers are corrected too.

## One list instead of four

`vm_eval.TopLevelHead` is now the single source of truth. Dispatch is an
exhaustive `switch`, so a ninth head cannot reach the VM without a handler.
`isSpecialTopLevelForm` — whose own comment read *"Keep this list in sync with
handleTopLevelForm above"* — is now one line over the enum, as is `check.zig`'s
`isEnvSetupForm`.

## Tests

Two new suites, both verified to **fail against a pre-fix build** and pass
after:

- `tests/scheme/compile/compile-toplevel-side-effects-2156.sh` — 12 assertions.
  Six probes (3 heads × `--compile`/`--disassemble`), the bare-`delete-file`
  control, the `include`-still-runs control, and an end-to-end standalone-binary
  check that the effect happens exactly once and output matches the interpreted
  run. Pre-fix: 7 failures.
- `tests/scheme/cache/cache-decline-reason-2114.sh` — 11 assertions: one per
  head, the nested-position control, and cold-MISS-then-warm-HIT. Pre-fix: 8
  failures.

`tests/scheme/timings/timings-1515.sh` updated for the new reason string.

Full suite: **2074 pass, 0 fail** (`zig build test`, `tests/scheme/run-all.sh`,
R7RS 1395/1395). Differential harness unchanged at 45/361 files writing an entry,
0 divergences. `zig fmt --check` and markdownlint clean.

Closes #2156.
Closes #2114.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -44,7 +44,10 @@ zero overhead when absent — see `docs/dev/timings.md`),
 `--completions <shell>`.
 Subcommands: `kaappi compile <file> [-o output]` compiles to a native binary
 via LLVM; `kaappi check <file>` runs compile-only static analysis (reads,
-expands, compiles, executes nothing) reporting read/compile errors plus the
+expands, compiles, executes no program code — only the five
+`vm_eval.TopLevelHead.isEnvSetup()` declarations later forms are compiled
+*against*, which `--compile` and `--disassemble` share since kaappi#2156)
+reporting read/compile errors plus the
 `KP4xxx` lint findings — unknown top-level variable (warning), and arity or
 wrong-type-literal on direct built-in calls (errors); honors
 `--diagnostics=json` and `--deny-warnings` — see `docs/dev/check.md`;
--- a/docs/dev/cache.md
+++ b/docs/dev/cache.md
@@ -124,9 +124,29 @@ to* the source as `file.sbc`. A central store is what makes `cache status` /
 - **Not cached — top-level forms the VM handles directly:** one occurrence of
   any of `import`, `define-library`, `include`, `include-ci`,
   `define-record-type`, `define-values`, `begin`, or `cond-expand` at top
-  level skips caching for the whole file (reason: `imports`). Library loading
-  can free collected function pointers, so these files' top-level functions
-  are not safe to serialize after the fact.
+  level skips caching for the whole file — every other form in it included.
+  These eight are `vm_eval.TopLevelHead`, the set `handleTopLevelForm`
+  claims; `--timings` names the one that fired (`not cached: top-level
+  cond-expand`). Until
+  [#2114](https://github.com/kaappi/kaappi/issues/2114) it reported all eight
+  as `imports`, so a file with no `import` in it was told an import was the
+  cause.
+
+  They are refused for one shared reason: `handleTopLevelForm` *interprets*
+  such a form and appends no `Function` to the run's compiled list, so a HIT —
+  which compiles nothing — would skip that form's work entirely. (The separate
+  hazard that library loading can free collected function pointers is real,
+  but specific to the three heads that load libraries.)
+
+  The rule is **top-level head position only**. The same constructs nested
+  inside a body are ordinary code and leave caching alive:
+
+  ```scheme
+  (define (a) (begin 1 2 3))                     ; still cacheable
+  (define (b) (cond-expand (else 'chosen)))
+  (define (c) (define-values (x y) (values 1 2)) (+ x y))
+  ```
+
 - **Not cached — compile-time registrations
   ([#2112](https://github.com/kaappi/kaappi/issues/2112)):** a top-level
   `define-syntax` or `define-property` (or a macro use expanding into one)
@@ -151,6 +171,21 @@ to* the source as `file.sbc`. A central store is what makes `cache status` /
   binary via `zig build -Dbundle=out.sbc`, not for the auto-run path. It is
   never read as the run-cache, so `--no-ir-opt --compile` can't poison a plain
   run.
+
+  `--compile` does **not** run the program. Of the eight heads above it
+  evaluates only the five `TopLevelHead.isEnvSetup()` names — the declarations
+  later forms are compiled *against* (`import`, `include`, `include-ci`,
+  `define-library`, `define-record-type`) — and records them in the artifact's
+  preamble for replay. `begin` and `cond-expand` are spliced into the form
+  stream so their bodies are compiled (only a `cond-expand`'s branch
+  *selection* is a compile-time question), and a `define-values` is recorded
+  without its producer expression being evaluated. Until
+  [#2156](https://github.com/kaappi/kaappi/issues/2156) all three were
+  evaluated for real, so `(begin (delete-file "x"))` deleted the file while
+  producing the `.sbc` and again at run time from the preamble — and, because
+  the preamble replays entirely *before* the compiled forms, a top-level
+  `begin` also ran out of program order in the artifact. `--disassemble`
+  follows the same discipline.
 - `.sld` library loads are never cached in either direction.
 
 ## Inspect, clear, bypass
--- a/docs/dev/cache.md
+++ b/docs/dev/cache.md
@@ -182,10 +182,30 @@ to* the source as `file.sbc`. A central store is what makes `cache status` /
   without its producer expression being evaluated. Until
   [#2156](https://github.com/kaappi/kaappi/issues/2156) all three were
   evaluated for real, so `(begin (delete-file "x"))` deleted the file while
-  producing the `.sbc` and again at run time from the preamble — and, because
-  the preamble replays entirely *before* the compiled forms, a top-level
-  `begin` also ran out of program order in the artifact. `--disassemble`
-  follows the same discipline.
+  producing the `.sbc` and again at run time from the preamble.
+  `--disassemble` follows the same discipline.
+
+  **The preamble replays in full before any compiled form**, so a recorded
+  declaration does not keep its position in the program. Splicing puts `begin`
+  and `cond-expand` bodies into the compiled stream, which is what restores
+  their order; the other six heads still replay first. For `import`,
+  `include`, `include-ci`, `define-library` and `define-record-type` that is
+  harmless — they are declarations, and hoisting them is what a preamble is
+  for. For `define-values` it is not, because its producer is arbitrary
+  program code that can depend on earlier forms:
+
+  ```scheme
+  (define x 1)
+  (define-values (a b) (values x 2))   ; replayed first, with x unbound
+  (display (list a b))
+  ```
+
+  The interpreter prints `(1 2)`; the standalone binary fails with
+  `preamble error[KP3001]: undefined variable 'x'` and exits 1. This
+  pre-dates #2156 and is unchanged by it — tracked separately as
+  [#2200](https://github.com/kaappi/kaappi/issues/2200), since fixing it needs
+  either an order-preserving preamble in the `.sbc` format or a compilable
+  `define-values`.
 - `.sld` library loads are never cached in either direction.
 
 ## Inspect, clear, bypass
--- a/docs/dev/timings.md
+++ b/docs/dev/timings.md
@@ -56,7 +56,7 @@ The cache line is never blank. When caching was not even attempted it says why:
 
 ```text
 cache: off (--no-ir-opt)     # or (sandbox), or (no home dir)
-cache: MISS (not cached: imports)   # imported programs are never cached (#1516)
+cache: MISS (not cached: top-level import)   # one of the eight top-level heads (#1516, #2114)
 ```
 
 ### Stages
--- a/src/check.zig
+++ b/src/check.zig
@@ -30,7 +30,6 @@ const reader = @import("reader.zig");
 const compiler = @import("compiler.zig");
 const vm_mod = @import("vm.zig");
 const vm_library = @import("vm_library.zig");
-const vm_eval = @import("vm_eval.zig");
 const ir_mod = @import("ir.zig");
 const diagnostics = @import("diagnostics.zig");
 const lsp_diagnostic = @import("lsp_diagnostic.zig");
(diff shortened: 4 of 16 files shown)


baijum:
Worked through all five. Every one was verified against the code before acting — four applied, one filed separately. Two turned out to be more serious than described.

### 1. `check.zig` — macro-shadowed `define-record-type` — **applied, and worse than reported**

Confirmed and fixed via `vm.topLevelHead`. But the report describes only the false-*negative* half ("neither compiled nor linted, no diagnostic produced"). Two corrections:

- **The lint channel can't show this at all.** `check` never lints through *any* macro use — `(define-syntax m ...)` + `(m undefined-thing)` reports nothing today, shadowed or not, because top-level define names are gathered structurally. The channel that actually discriminates is **compile errors**: a shadowed use the macro cannot match is a real syntax error, and `check` silently exited 0 on it.
- **The false-positive half is the damaging one, and wasn't mentioned.** Dropping the form also dropped whatever it was meant to bind, so later forms drew phantom diagnostics. I swept `check` over 880 files (every test `.scm` plus every shipped `.sld`), old binary vs new. Exactly two differ, both known-good files that pass their own suites:

  | file | before | after |
  |---|---|---|
  | `tests/scheme/srfi/srfi136.scm` | 3 × `KP2001 invalid syntax` + 1 phantom `KP4001` | clean |
  | `tests/scheme/srfi/srfi150.scm` | 1 × `KP2002 type error in 'caar'` | clean |

  Those are hard errors — `check` was failing valid files. SRFI 136 binds each type name to a CPS introspection macro that the dropped form never registered, which is where the `KP2001`s came from.

  This is also why the regression test uses the real SRFI files rather than a synthetic probe: a synthetic one can't reproduce it, for the same structural-gathering reason above. Four new assertions in `tests/scheme/errors/check.sh`, all four failing pre-fix.

### 2. `define-values` preamble ordering — **valid, filed as #2200, not fixed here**

Confirmed, and your `(define x 1)` example is sharper than reordering — the artifact doesn't just reorder, it **fails**:

```
interpreter:  (1 2)                                     exit 0
artifact:     preamble error[KP3001]: undefined variable 'x'   exit 1
```

Reproduced identically at `07cceb25` (before this PR) and `e179b302` (after), so this PR neither caused nor fixed it. Splicing is what restores order, and there is nothing to splice a `define-values` into — it has no compiler form. The two real fixes are an order-preserving preamble (`.sbc` format change) or a compilable `define-values` lowering; both are larger than this PR.

Filed as #2200 with the repro, and `docs/dev/cache.md` now states the limitation explicitly instead of leaving the order-preservation claim unqualified — that half of your comment is applied here.

Worth noting the PR did remove a spurious compile-time `error[KP3001]` that `--compile` printed for that same file, since it no longer evaluates the producer.

### 3. Three silent failure paths in `compileFile` — **applied**

Correct, and the consequence is as stated: an artifact missing an `import` or a top-level form, with `Compiled ... -> ...` printed and exit 0. All three now propagate `OutOfMemory`, matching `runFile`. (All three predate this PR; the loop rewrite is what put them in the diff.)

### 4. Deduplicate the include/include-ci assertion — **applied**

Extracted `assert_decline_reason` plus a `cache_line_for` variant for the on-disk case, since `include` needs a sibling fixture file.

### 5. Use `interp_stdout` / `assert_tiers_agree` — **applied**

Verified these exist in `shell-common.sh` and are the documented convention (`tests/scheme/CLAUDE.md`, "The interpreter is the oracle"). Adopted — and they also compare **exit status**, which my hand-rolled comparison ignored. The golden string stays as a second assertion against the interpreter, per that same doc. None of the three by-design tier differences apply to this program (no bare top-level expression, no top-level error).

---

Full suite after the changes: **2074 pass, 0 fail** (`zig build test`, `run-all.sh`, R7RS 1395/1395), `zig fmt --check` and markdownlint clean, differential harness unchanged at 45/361 caching with 0 divergences.

Issue #2108, opened 2026-08-01
wasm32: no file-backed .sld is importable even when the host mounts it — platform.openRead has no WASI branch, so resolveLibraryPath's probe always fails


baijum:
**Category:** Tier divergence (WASM) · **Severity:** missing-feature (whole library-loading path dead on one target) · **Found by:** systematic audit v2, Phase 4D (#1890)

No file-backed `.sld` can be imported on the WASM build — **even when the host has the directory mounted and the file is demonstrably readable**. That covers ~160 portable SRFIs, every ecosystem library, and any user-authored `.sld`. Only registry built-ins and the two `embedded_libraries` entries are reachable.

This contradicts the loader's own stated guarantee. `vm_library.zig:329-335`:

> WASM has no blanket block like sandbox's, but relies on whatever preopened directory the host (browser shim, wasmtime `--dir`) happens to mount at a given relative path [...] anything else still falls through to the normal disk check (**unaffected -- a host that does mount lib/ still serves those fine**).

Hosts that mount `lib/` do not in fact serve those fine. Kaappi's own CI is such a host: `.github/workflows/ci.yml` runs `wasmtime run --dir=. zig-out/bin/kaappi.wasm ...` from the repo root, with `lib/srfi/*.sld` sitting right there.

## Reproduction

```
$ wasmtime run --dir=. zig-out/bin/kaappi.wasm i2.scm      # (import (scheme base) (srfi 2))
i2.scm:1: error[KP2001]: library not found: (srfi.2)
```

Measured over the audit corpus (591 files in `tests/scheme/{smoke,compliance,audit,continuations,hygiene,srfi,differential/probes}`): **401 files fail on WASM for this reason alone** — native runs them fine, and the only WASM stderr is `KP2001 library not found`.

| import | native | WASM |
|---|---|---|
| `(scheme base)`, `(scheme write)`, `(scheme file)` | ok | ok (registry) |
| `(srfi 1/9/13/39/69/133/254/258/260)` | ok | ok (registry built-ins) |
| `(kaappi parallel)`, `(srfi 181)` | ok | ok (**embedded**) |
| `(srfi 2/8/26/41/64/128/158/231)` | ok | **KP2001 library not found** |
| a user `.sld` beside the script | ok | **KP2001 library not found** |

`cond-expand` agrees with the import in every one of these cases, so nothing is *inconsistent* — the capability is simply absent.

## Discriminating control — the file is readable; only the probe fails

[...]

Issue #2107, opened 2026-08-01
wasm32: write/display of an 848-deep nested structure aborts the module — MAX_PRINT_DEPTH's 1024 guard is unreachable there


baijum:
**Category:** Tier divergence (WASM) · **Severity:** crash (uncatchable module abort) · **Found by:** systematic audit v2, Phase 4D (#1890)

`write`/`display` of a **car**-nested structure deeper than **847** aborts the whole WASM module with an out-of-bounds memory access. Native prints the same value fine and exits 0.

The regression test for closed #49 — `tests/scheme/smoke/deep-nesting-print.scm`, whose whole point is "deep nesting should print without crashing (stack overflow)" — is itself one of the files that aborts. Nothing runs the corpus on WASM today, so this has been invisible.

## Reproduction

```scheme
(import (scheme base) (scheme write))
(define (nest n acc) (if (= n 0) acc (nest (- n 1) (list acc))))
(let ((p (open-output-string))) (write (nest 848 '()) p) (display "ok"))
```

| | native aarch64 | **wasm32 (wasmtime 46.0.0)** |
|---|---|---|
| depth 847 | `ok`, exit 0 | `ok`, exit 0 |
| depth 848 | `ok`, exit 0 | **exit 134**, no stdout, no Kaappi diagnostic |
| `deep-nesting-print.scm` (depth 200000) | `PASS: ... length=2051`, exit 0 | **exit 134** |

The trap, verbatim:

```
2: memory fault at wasm address 0xffdff9cf in linear memory of size 0x1190000
3: wasm trap: out of bounds memory access
```

`0xffdff9cf` is the shadow stack pointer having run off the bottom of the stack region and wrapped past address 0 — not a clean stack-overflow diagnostic. Deterministic: 3/3 runs at depth 848.

## It is uncatchable

```scheme
(display "before") (newline)
(display (guard (e (#t "CAUGHT"))
  (let ((p (open-output-string))) (write (nest 2000 '()) p) "no-error")))
(newline)
(display "after") (newline)
```

Native prints `before` / `no-error` / `after`, exit 0. WASM prints `before` and then aborts — `CAUGHT` never runs, `after` never runs. A `guard` with a `#t` clause cannot see it, so no program can defend against it.

## Discriminating controls

Two, and together they isolate this to car-recursion depth in the printer:

[...]

Issue #1912, opened 2026-07-31
wasm32: index arguments are truncated to u32 inside the bounds check — silent wrong reads AND writes


baijum:
**Category:** Indexing / portability · **Severity:** wrong-result (silent data corruption) · **Found by:** systematic audit v2, Phase 2.1 (#1890)

Index arguments are `@intCast` to `usize` **inside** the bounds comparison. On wasm32 `usize` is `u32`, so an index ≥ 2³² wraps to its low 32 bits *before* the check and then aliases an in-range element. The bounds check still passes — against the truncated value.

This is not confined to the `%` primitives the unit was auditing. It reproduces on plain R7RS `vector-ref` / `vector-set!`.

## Reproduction

```scheme
(define v (vector 10 11 12 13))
(vector-ref v 4294967297)                      ; 2^32+1
(vector-ref v 4294967305)                      ; 2^32+9  <- control
(vector-set! v 4294967297 99) (vector-ref v 1)
```

| | native aarch64 | **wasm32 (wasmtime 46.0.0)** |
|---|---|---|
| `(vector-ref v 4294967297)` | RAISED | **`11`** — silently reads element 1 |
| `(vector-ref v 4294967305)` | RAISED | RAISED |
| `(vector-set! v 4294967297 99)` → elt 1 | RAISED, stays `11` | **succeeds, elt 1 becomes `99`** |

## Discriminating control

`2³²+9` raises on **both** platforms, because its low 32 bits (9) still exceed the length 4. That is what proves the failure is exactly low-32-bit truncation and not a missing bounds check.

## Affected

Confirmed on wasm32: `vector-ref`, `vector-set!`, `string-ref`, `bytevector-u8-ref`, `%numeric-vector-ref`, `%numeric-vector-set!`, `%record-ref`. `list-ref` is unaffected — it walks the list.

Representative sites: `primitives.zig:1063`, `:1078`, `:540`, `:546`; `primitives_srfi160.zig:264`, `:276`. The shape is always `@as(usize, @intCast(raw_idx)) >= len`.

## The fix pattern is already in-tree

`makeNumericVectorFn` (`primitives_srfi160.zig:225-231`) guards this exact hazard, with a comment naming the platform:

> narrowing to usize — on wasm32 (usize = u32) a fixnum-range length (up to 2^47) would otherwise panic @intCast instead of raising a catchable error.

It compares in `u64` **before** narrowing. The ref/set paths do not. Applying the same pre-comparison at each site closes this.

[...]

Issue #1903, opened 2026-07-31
[Tooling] kaappi test and run-all.sh disagree on srfi150.scm; the summary contradicts itself


baijum:
**Category:** Tooling / test infrastructure · **Severity:** tooling correctness · **Found by:** systematic audit v2 reconnaissance (#1890)

The two runners return opposite verdicts on the same file, and `kaappi test`'s own output disagrees with itself.

## Reproduction

```console
$ kaappi tests/scheme/srfi/srfi150.scm        # what run-all.sh does
exit=0

$ kaappi test tests/scheme/srfi/srfi150.scm
exit=1
```

`kaappi test`'s summary, verbatim:

```text
Summary: 21 passed, 0 failed, 0 unexpected-pass, 4 expected-fail, 0 skipped
Files:   1 run, 1 failed, 1 errored (39ms)
```

Line 1 says nothing failed. Line 2 says the file failed. Both are printed by the same command.

JSON shows the mechanism — `fail: 0` beside `error: true`:

```json
{"type":"file","file":"tests/scheme/srfi/srfi150.scm","suite":"srfi-150",
 "tests":25,"pass":21,"fail":0,"xpass":0,"xfail":4,"skip":0,"error":true,
 "error_message":"tests/scheme/srfi/srfi150.scm:164: error[KP3000]: unknown field a",
 "failures":[]}
```

## Cause

A top-level form in the suite raises (`:164`, `KP3000: unknown field a`). `kaappi test` treats a tolerated top-level error as a **file** error and exits nonzero; the legacy path does not.

The file *deliberately* depends on the legacy behaviour — its own comment at `:158-163` says so:

> `kaappi`'s batch runner tolerates one top-level form erroring and continues to the next, so the rest of the suite still runs and reports correctly.

So this is a genuine semantic disagreement between two supported runners, not a stray bug in one of them.

## Why it matters

`docs/dev/test-runner.md` presents `kaappi test` as the first-class runner, and the v2 audit strategy's session protocol now tells auditors to use it. But **CI never runs it** — `run-all.sh` is what gates PRs — so the divergence is latent today and will surface the moment anyone tries to make `kaappi test` the gate.

It also means the two runners report different totals for the same tree, which undermines using either as a baseline.

## Decision needed

1. Is a tolerated top-level error a *file* error? If yes, `srfi150.scm` needs restructuring so the raise happens inside a test case rather than at top level. If no, `kaappi test` should record it without failing the file.
[...]

PR #2207, opened 2026-08-03
Make a failing test able to fail

### LINKED ISSUE #2116: [Tooling] 54 test files cannot fail: no exit path, and run-all.sh's stdout net matches neither output shape
**Category:** Tooling
**Severity:** coverage gap (silent — a failing regression test is reported green)

54 `.scm` files in `run-all.sh`-globbed suite directories have **no way to
signal failure**. Both runners report them as passing regardless of what they
compute. `tests/scheme/smoke/thread-sleep-876.scm` was demonstrably green under
the very regression it was written to catch.

## Minimal reproduction

`tests/scheme/smoke/thread-sleep-876.scm` (before the phase-5E fix) was:

```scheme
(import (srfi 18))
(define s0 (time->seconds (current-time)))
(thread-sleep! 0.1)
(define s1 (time->seconds (current-time)))
(display (>= (- s1 s0) 0.09))
(newline)
```

Substituting `(thread-sleep! 0)` — exactly the #876 regression — gives:

```console
$ kaappi tests/scheme/smoke/thread-sleep-876.scm    # mutated
#f
$ echo $?
0
```

It prints the answer and exits 0 either way.

## The safety net does not cover either output shape

`run-all.sh`'s `run_file_worker` has a net for files that forget the SRFI-64
epilogue (`tests/scheme/run-all.sh:125`):

```bash
if grep -Eq '(^|[^0-9])[1-9][0-9]* fail|unexpected (failures|errors) +[1-9]' "$slot.out"; then
```

It requires a failure **count**. Neither shape used by these 54 files has one:

| output | net verdict |
|---|---|
| `#f` | PASS |
| `FAIL: (odd? 2.5) did not raise error` | PASS |

End-to-end, with `tests/scheme/smoke/even-odd-non-integer.scm` mutated so its
first check reports failure:

```console
exit status: 0
FAIL: (odd? 2.5) did not raise error
PASS: (even? 2.5) raises error
...
--- run-all.sh verdict logic ---
VERDICT: PASS
```

`kaappi test` does not cover it either — it reports the file as passing and
says so explicitly:

```console
$ kaappi test tests/scheme/smoke/thread-sleep-876.scm
  PASS  tests/scheme/smoke/thread-sleep-876.scm  (0 tests, 0 skipped, 101ms)
Tests:   0 passed, 0 failed, ...
```

`0 tests` is the whole finding: the runner knows the file asserted nothing and
still calls it PASS.

## Scope

Files with no `(exit …)`, no `(error …)`, no `test-runner-fail-count` and no
`assert`, in directories `run-all.sh` globs:

| directory | files |
|---|---:|
| `smoke/` | 51 |
| `compliance/` | 1 (`scheme-repl.scm`) |
| `continuations/` | 1 (`coroutine-repl-echo.scm`) |
| `srfi/` | 1 (`test-srfi69-extras.scm`) |
| **total** | **54** |

Of those, 16 print a bare `FAIL:` line (missed by the net, as above) and 38
print nothing failure-shaped at all — they cannot signal a failure in any form.

Reproduce the list with:

```bash
for f in tests/scheme/{smoke,compliance,continuations,srfi,hygiene,audit}/*.scm; do
  grep -qE "\(exit |\(error |test-runner-fail-count|assert" "$f" || echo "$f"
done
```

## Why this matters now

The v2 audit campaign has twice reopened an issue that was closed on the
strength of a regression test which only ever exercised the valid case (#682,
#550). This is the same failure mode one level lower: the test exercises the
right case and then discards the verdict.

## Suggested fix

Two independent parts, either of which closes the `thread-sleep-876.scm` case:

1. **Convert the 54 files.** Mechanical for most: they already compute a
   boolean per check. `tests/scheme/CLAUDE.md` already documents the required
   SRFI-64 shape and the exit-on-fail epilogue.
2. **Make it unrepresentable.** `run-all.sh` already has a reachability check
   that fails the run when a `.scm` file under a suite *subdirectory* contains
   `test-begin` (kaappi#1900). The same shape works here: fail the run if a
   globbed `.scm` file contains neither `test-begin` nor `(exit`. That gate is
   what stops the count growing back.

Widening the net's grep to also match a bare `FAIL` token would catch 16 of the
54 and is worth doing regardless, but it cannot reach the 38 that print nothing.

**Verified:** fresh ReleaseSafe build at `3c1a56be`, `KAAPPI_HOME` isolated,
cache cleared. Counts are from the unmodified tree (`git stash`ed), and exclude
`thread-sleep-876.scm`, which this campaign's phase 5E converts.

**Found by:** Systematic audit v2, Phase 5E (de-flake and arm the timing tests)

### LINKED ISSUE #2157: [CI] Five ci.yml steps run r7rs-tests.scm bare — the (chibi test) shim exits 0 on a failed assertion, so 1,395 assertions gate nothing there

### LINKED ISSUE #2162: [Tooling] primitives_filesystem-audit.scm asserts the environment variable KAAPPI is unset — the name CI itself exports
**Category:** Tooling (test)
**Severity:** edge-case (a test fails for reasons outside the program under test)

`tests/scheme/audit/primitives_filesystem-audit.scm:850-852` asserts that an
environment variable literally named `KAAPPI` is **unset**:

```scheme
(test-equal "(get-environment-variable 'KAAPPI')" #f (get-environment-variable "KAAPPI"))
(test-equal "(get-environment-variable 'KAAPPI=A212')" #f (get-environment-variable "KAAPPI=A212"))
(test-equal "(assoc 'KAAPPI' (get-environment-variables))" #f (assoc "KAAPPI" (get-environment-variables)))
```

`KAAPPI` is not a name the test controls — it is a name the project's own
harness uses for "the binary under test". Both Windows CI jobs export it
(`.github/workflows/ci.yml`, the `Shell suites (#1612)` steps:
`export KAAPPI=bin/kaappi.exe`), and `tests/scheme/shell-common.sh` documents
it as the convention every shell suite reads.

## Minimal reproduction

```console
$ zig-out/bin/kaappi --lib-path lib tests/scheme/audit/primitives_filesystem-audit.scm | grep -c ^FAIL
0
$ KAAPPI=zig-out/bin/kaappi zig-out/bin/kaappi --lib-path lib tests/scheme/audit/primitives_filesystem-audit.scm | grep ^FAIL
FAIL (get-environment-variable 'KAAPPI')
FAIL (get-environment-variable 'KAAPPI=A212')
FAIL (assoc 'KAAPPI' (get-environment-variables))
```

**Discriminating control:** exporting any *other* name (`FOO`, `KAAPPI_A`,
`ZZZ_LAST`) changes nothing, and the set/delete round trip the surrounding
block tests works correctly in all of those cases — so this is not an
environment-handling bug in the interpreter, it is the assertion depending on
an uncontrolled fact.

## How it surfaced

Audit v2 Phase 7E's new `tools/run-gc-stress-suite.sh` exported `KAAPPI` for
its worker subshell — the obvious name for "the binary to run". Its first run
reported these three as gc-stress failures. They were not; the script now
exports `GC_STRESS_KAAPPI_BIN` instead, with a comment pointing here. Anyone
writing the next harness will reach for the same name.

Note also that the three assertion *labels* are truncated in a way that
misattributes them: they render as `'KAAPPI'` and `'KAAPPI=A212'`, which read
like the neighbouring `KAAPPI_AUDIT_VAR` / `KAAPPI_A212_OK` cases 700 lines
earlier and sent the first triage to the wrong block entirely.

## Suggested fix

Assert on a name the test owns and can guarantee. The block's intent is "a
name that was never set reads as `#f`", which a per-run unique name expresses
without depending on the ambient environment — e.g. `KAAPPI_A212_ABSENT`,
which the surrounding block's own naming convention already suggests, and
which nothing in the repo exports.

**Verified:** fresh ReleaseSafe build at `910f7c75`, macOS aarch64, cache
isolated; 3/3 assertions flip with the variable set and back without it.
**Source location:** `tests/scheme/audit/primitives_filesystem-audit.scm:850-852`;
`.github/workflows/ci.yml` (both Windows `Shell suites` steps);
`tests/scheme/shell-common.sh`

**Found by:** Systematic audit v2, Phase 7E

### LINKED ISSUE #2163: [Tooling] run-all.sh silently builds a plain kaappi when zig-out/bin/kaappi is missing — a suite 'run under -Dgc-stress' can be a plain run
**Category:** Tooling
**Severity:** wrong-result (silent — a suite run under a build flag can measure a binary built without it)

`tests/scheme/run-all.sh` opens with:

```bash
# Use pre-built binary if available, otherwise build once.
if [[ ! -x zig-out/bin/kaappi ]]; then
    zig build
fi
KAAPPI=zig-out/bin/kaappi
```

That `zig build` carries **no build options**. Combined with the fact that
`zig build test -Dgc-stress=true` does *not* install an executable — it builds
and runs the test binaries only — the natural two-command sequence for "run
everything under gc-stress"

```bash
zig build test -Dgc-stress=true      # stresses the unit suite
bash tests/scheme/run-all.sh         # silently builds and runs a PLAIN kaappi
```

reports a full green Scheme suite that stressed nothing. The same applies to
any other build option someone wants the Scheme corpus exercised under
(`-Doptimize=`, `-Dmax-registers=`, `-Dgc-threshold=`). Nothing in the run's
output names the configuration of the binary it used.

## Why this is more than theoretical

`kaappi features --json` reports the comptime options, so the substitution is
detectable in one command — but only if you know to look:

```console
$ rm -rf zig-out && zig build test -Dgc-stress=true >/dev/null
$ ls zig-out/bin/kaappi
ls: zig-out/bin/kaappi: No such file or directory
$ bash tests/scheme/run-all.sh >/dev/null 2>&1
$ zig-out/bin/kaappi features --json | tr ',' '\n' | grep gc_stress
"gc_stress":false
```

**Discriminating control:** with `zig build -Dgc-stress=true` run first, the
same command reports `"gc_stress":true` and run-all.sh reuses it.

## The measurement that raised it

Audit v2 Phase 7E timed the `run-all.sh` `.scm` corpus (605 files) against a
verified `"gc_stress":true` build on macOS aarch64 (12 cores, load average
8-13 from concurrent work):

| file | wall time under gc-stress |
|---|---|
| `tests/scheme/smoke/trampoline-polish-1375.scm` | 11 min |
| `tests/scheme/srfi/srfi14.scm` | 9 min |
| `tests/scheme/srfi/srfi264.scm` | 7 min |

`run-all.sh`'s default per-file timeout is **60 s**
(`KAAPPI_TEST_TIMEOUT:-60`). A genuinely-stressed corpus run cannot report
`0 timeout` with that budget — the three files above exceed it by two orders of
magnitude, and dividing out even a generous 5x load factor leaves them well
over a minute each.

Phase 5F's droplet run reported the Scheme suite under gc-stress as
`2061 pass, 0 fail, 0 timeout` (#2147, `docs/audit-strategy.md` 5F). That
result is **inconsistent with a stressed binary** and consistent with exactly
the substitution above. The droplet was destroyed, so this cannot be confirmed
directly — hence this issue rather than a correction — but the mechanism is
verified and the arithmetic does not close. 5F's *unit*-suite half is
unaffected: `zig build test -Dgc-stress=true` takes the flag directly, and its
control (50 s / `1567 pass, 3 skip` vs 5 m 07 s / `1570 pass`) is exactly the
evidence a Scheme-suite run has no equivalent of today.

## Suggested fix

Make the substitution impossible to make silently, in either of two ways:

1. **Refuse rather than build.** If `zig-out/bin/kaappi` is missing, print what
   to run and exit 2. The convenience build dates from when the script was the
   only runner; every CI caller already builds first.
2. **Report the configuration.** Print `kaappi features --json`'s
   version + build id + `gc_stress` in the run header, so the log says which
   binary produced the counts. This is cheap and helps well beyond gc-stress.

`tools/run-gc-stress-suite.sh` (added by Phase 7E) takes the stricter line for
its own case: it refuses to run at all unless `features --json` reports
`"gc_stress":true`, which is what makes it a gate rather than a decoration.

**Verified:** fresh ReleaseSafe build at `910f7c75`, cache isolated, macOS
aarch64. The per-file timings were taken under concurrent load and are upper
bounds on an idle machine; the conclusion does not depend on their precision.
**Source location:** `tests/scheme/run-all.sh` (the `zig build` fallback);
`build.zig` (the `test` step installs no executable)

**Found by:** Systematic audit v2, Phase 7E

baijum:
Four issues, one root cause: **a check that reports success having asserted nothing.**
`docs/audit-strategy.md` ranks this group (R10) first, ahead of every correctness
fix, because until it lands every other fix's regression test is of unknown value.

Closes #2116
Closes #2157
Closes #2162
Closes #2163

#1900 needed no work — it is already closed, and its reachability check was
re-verified here (a `test-begin` file planted in `srfi/slow/` is still caught).

## #2116 — 56 files that could not fail

They computed the right thing and threw the verdict away. `(display (= x 43))`
prints `#f` and exits 0; `(display "FAIL: ...")` exits 0 too. Both runners
reported every one of them as passing whatever they computed, and
`smoke/thread-sleep-876.scm` was demonstrably green under the very regression it
was written to catch. `run-all.sh`'s stdout net could not help — it matched a
failure *count*, which neither shape has.

All 56 are now SRFI-64 suites with the exit-on-fail epilogue. That is **55 from
the issue's own enumeration plus one it missed**: `deep-nesting-print-tier-margin.scm`
slipped through because the issue's predicate matched a bare `assert`, and the
word "assert" appears in one of its comments. It printed `FAIL:` and exited 0
like the rest.

Three files stay import-free with a hand-rolled `(exit 1)`, each saying why in
its own header — `(import (srfi 64))` fails at library load on WASM (#2108), and
`run-wasm-differential.sh` classifies that as LIBDIFF:

| File | Why |
|---|---|
| `smoke/large-index-bounds-1912.scm` | `KNOWN_DIFFS` probe for #1912 — LIBDIFF would retire its entry |
| `smoke/deep-nesting-print.scm` | `KNOWN_DIFFS` probe for #2107, same reason |
| `smoke/deep-nesting-print-tier-margin.scm` | the positive control that keeps #2107's margin under watch |
| `continuations/coroutine-repl-echo.scm` | its top-level forms must stay **bare** — consuming them in `test-equal` is exactly what hid the bug it exists to catch |

Two new gates so the count cannot grow back:

- a **verdict-channel check** in `run-all.sh`, failing the run if a globbed suite
  file carries none of `test-begin`, `(exit`, `(error`, `(assert`. It is a
  heuristic and says so — an `(error` inside a `guard` being *tested* is not a
  verdict channel — so it is a backstop, not the standard.
- the stdout net widened to a bare `FAIL` token, anchored on both sides so it
  cannot fire on `FAILURES` or lower-case prose.

**The predicted second finding appeared on the first run**, exactly as #2157
warned it might: `smoke/fiber-error-handling.scm` had been asserting the #551
behaviour that **#1155 deliberately reversed** (`spawn` must *accept* native
procedures). It printed `FAIL - should have raised` and exited 0 ever since
#1155 merged. Now asserts what #1155 specifies.

`docs/dev/testing.md`'s "Writing a Scheme test" section — which taught the
print-and-eyeball pattern that produced these files — was rewritten.

## #2157 — five CI steps where 1,395 assertions gated nothing

The `(chibi test)` shim has no exit-on-fail epilogue, so
`kaappi r7rs-tests.scm; echo $?` prints 0 with failures in the log. `run-all.sh`
always parsed the counts; five `ci.yml` steps did not.

`tools/run-r7rs-suite.sh` is now the single implementation behind all six callers
(`run-all.sh`, `run-gc-stress-suite.sh`, riscv64, s390x, ppc64le, both Windows
legs). `run-all.sh` reads the counts back through a file rather than re-parsing,
so there is still exactly one parser. It also refuses a run reporting 0 passing
assertions, so "did not run" cannot read as "no failures".

`runner-agreement.sh` (#1903) carries a **copy** of the net regex — precisely how
the two runners' rules would drift while that script kept certifying agreement —
so it now greps `run-all.sh` for the pattern verbatim and fails if it is gone.

## #2163 — a suite "run under -Dgc-stress" could be a plain run

`zig build test -Dgc-stress=true` installs no executable, so the natural
follow-up `bash tests/scheme/run-all.sh` silently built and measured a **plain**
binary. `run-all.sh` now refuses rather than building, and prints the binary's
`version` / `build_id` / `target` / `build_mode` / `gc_stress` in its header, so
the log always says which configuration produced the counts.

## #2162 — a test asserting that `KAAPPI` is unset

`primitives_filesystem-audit.scm` asserted that an environment variable named
`KAAPPI` is unset — the name every harness in this repo uses for "the binary
under test". Moved to `KAAPPI_A212_ABSENT`, which the block's own naming
convention already suggested.

## Measured cost, stated up front

Giving 52 files an `(import (srfi 64))` moves them into the #2108 LIBDIFF bucket:
`run-wasm-differential.sh`'s `agree` count goes **176 → 124**. That is a real,
temporary reduction in that harness's reach which reverses entirely when #2108
lands, and it is recorded in the harness header so a green run is not misread as
unchanged coverage.

## Verification

| Check | Result |
|---|---|
| `bash tests/scheme/run-all.sh` | 2075 pass, 0 fail, 0 timeout |
| `zig build test --summary all` | 1672 pass, 3 skip |
| `runner-agreement.sh` | 10/10 |
| `run-wasm-differential.sh` | PASS, 0 divergences |
| R7RS block under a real `-Dgc-stress=true` build | 1395 pass, 0 fail |

Mutation controls, one per new gate — each verified to fire and to stay quiet:

- a fresh verdictless file → `NO-VERDICT`, run fails
- the net regex removed from `run-all.sh` → `runner-agreement.sh` exits 1
- `zig-out/bin/kaappi` removed → `run-all.sh` exits 2 with instructions
- one broken assertion in `r7rs-tests.scm` → bare spelling exits **0**, the new
  wrapper exits **1** (the discriminating control from #2157)
- each hand-rolled `(exit 1)` file mutated → exits 1, stdout otherwise byte-identical
- the widened net checked against 6 must-catch and 6 must-not-catch strings

## Out of scope, reported not fixed

- **`kaappi test` still calls a 0-assertion file `PASS`.** It honours a nonzero
  exit correctly (verified: a 0-test file exiting 1 is reported `ERROR`, exit 1),
  and directory-mode selection filters on the `"srfi 64"` substring — so this
  conversion *adds* 52 files to its reach. Making "0 tests" itself a failure
  needs its own analysis; seven corpus files legitimately report 0.
- **Three pre-existing `-Dgc-stress=true` failures**, all in files this PR does
  not touch, reproduced on an unmodified tree: `srfi237.scm`,
  `primitives_srfi237-audit.scm`, `srfi240-audit.scm` all fail
  `record-uid->rtd` / nongenerative-rtd-reuse assertions, plus three SIGABRTs in
  `gc_collect` (`srfi1-ext.scm`, `srfi1-gc-stress.scm`,
  `primitives_srfi1-audit.scm`). Happy to file these if wanted.
- **`deep-nesting-print.scm`'s `KNOWN_DIFFS` entry is already STALE** at
  `e62b90eb` with the file unmodified — #2107 no longer reproduces for it on the
  current wasmtime. Confirmed pre-existing by re-running against `HEAD`'s version;
  left in place for #2107's owner to judge, and recorded in the harness header.


baijum:
Worked through all ten findings. **Eight were valid and are fixed in 43f8f934. The two marked 🔴 Critical were both wrong** — disproving them was the most useful part of the pass, so the evidence is below rather than just a "skipped".

## Declined, with evidence

**[run-all.sh:637] "does not fail the run when the R7RS suite reports failing assertions"** — it does. The finding traced the block but stopped at line 637 and did not check the summary condition eleven lines later, which tests `$R7RS_FAIL -gt 0` — and `R7RS_FAIL` *is* sourced from the counts file. (The comment asked "Confirm nothing below line 637 already compensates"; something does.)

Proved end to end rather than argued — one failing assertion injected into `r7rs-tests.scm`, full `run-all.sh`:

```
=== R7RS test suite ===
  1395 pass, 1 fail (exit 0)
        FAIL [4.1 Primitive expression types]: expected 999999 got 2
...
run-all.sh EXIT=1
```

The *suggestion* is still worth taking, so I took it: the block now gates on `R7RS_RUNNER_STATUS -ne 0`, one condition covering every runner exit instead of three that happen to cover today's two. `R7RS_FAIL` stays in the summary as an independent second signal — it is what keeps the counts load-bearing rather than decorative.

**[vector-negative-index.scm] "(srfi 133) is not imported, so all five assertions can pass vacuously"** — SRFI 133 is **built-in** in Kaappi; the three names resolve from `(scheme base)`:

```console
$ kaappi --lib-path lib -  # (import (scheme base) (scheme write)) only
(reverse!  reverse-copy #(3 2 1) unfold #(0 1 2))
```

Discriminating control the finding needed: the **valid** calls do not raise — they would if the names were unbound, which is exactly what a vacuous pass looks like — while a genuinely unbound name does, so `raises?` is sound:

```
valid vector-reverse! does NOT raise            PASS
valid vector-reverse-copy does NOT raise        PASS
valid vector-unfold does NOT raise              PASS
an unbound name raises                          PASS
```

The supporting argument — that `vector-unfold-empty-values-806.scm` imports `(srfi 133)`, so this file must need it — is a style observation, not evidence of necessity.

## Accepted

**The inventory finding was right, and understated.** The counts genuinely drifted, and two docs said "Three files are exempt" while omitting `deep-nesting-print.scm` — a file I converted and then reverted late without updating them. The true numbers:

| | files |
|---|--:|
| had no verdict channel | **56** |
| ↳ converted to SRFI-64 | 52 |
| ↳ kept a hand-rolled `(exit 1)` | 4 |
| plus `fiber-error-handling.scm` | 1 |
| **SRFI-64 conversions total** | **53** |

`tests/scheme/CLAUDE.md` now carries that as the single table; the other four sites point at it.

**Seven callers, not six** — fixed in all three places.

**`-x` not `-e`** — correct, and the exit-code taxonomy argument is the right one. Verified: a non-executable binary now gives exit 2 (harness failure), not 1 ("the suite reported failing assertions").

**The two `exit` imports** — accepted after verifying the concern I had about them. `(scheme process-context)` is a **built-in** library, not a file-backed `.sld`, so it does **not** move these files into the #2108 LIBDIFF bucket. Confirmed under wasmtime 46.0.0 that `deep-nesting-print-tier-margin.scm` produces byte-identical output and exit status with and without the import, and that `coroutine-repl-echo.scm`'s top-level echo — the whole point of that file — is unchanged. The two files that *are* `KNOWN_DIFFS` probes are deliberately left alone, since their divergence is the measurement.

**Three assertion-strength findings**, each fixed with a mutation control. The `syntax-rules-many-vars` one is the sharpest — addition is commutative, so a permuted expansion still totalled 153:

```
OLD sum-based,  permuted template: exit=0   (bug slips through)
NEW list-based, permuted template: exit=1   (caught)
```

`prettyprint-cycle-859` now matches `#0=` and `#0#` rather than any `#` (wrong label → exit 1), and the vacuous `(eq? '->string '->string)` — both literals truncate identically under #647, so `eq?` is `#t` either way — now asserts the symbol's text.

## Verification

`run-all.sh` 2075 pass / 0 fail; `runner-agreement.sh` 10/10 with its regex copy and drift check intact; markdownlint clean.

One note for anyone reproducing locally: three `compile/` tests will fail with `fatal: invalid embedded bytecode (wrong version or format)` if `zig-out/bin/kaappi` is stale relative to the working tree — `zig build -Dbundle=` stamps the *current* tree's build id (`<sha>-dirty`) while the installed binary still reports `<sha>`. Re-run `zig build` first. That is the documented `docs/dev/cache.md` hazard, not a change here; CI builds fresh and is unaffected.

PR #2221, opened 2026-08-04
Add paredit-style structural editing to the REPL

### LINKED ISSUE #2216: REPL: structural s-expression editing (barf/slurp/raise/rotate), portable from bestline
## What

Add paredit-style structural editing to the REPL line editor: barf, slurp,
raise, and rotate. These move a paren rather than a character — the edits a
Lisp programmer actually wants and that a character-oriented editor makes
tedious.

Given `(+ 1 (* 2 3))` with the cursor inside the inner form:

| Command | Effect |
|---|---|
| **slurp** | pull the next datum in past the closing paren — `(+ 1 (* 2 3 4))` |
| **barf** | push the last datum out past the closing paren — `(+ 1 (* 2) 3)` |
| **raise** | replace the enclosing form with the datum at point — `(+ 1 3)` |
| **rotate** | rotate the datums of the enclosing form |

## Why this is filed separately

It came out of evaluating line-editing libraries for the REPL (see the
isocline migration work). [bestline](https://github.com/jart/bestline) —
Justine Tunney's linenoise fork — is the only one of the three candidates that
implements these, and they are the most interesting thing in any of them for a
Scheme REPL. But bestline is not the right base for us overall: its
`bestlineBalanceModeEnable()` resets the edit buffer on every continuation line
(`bestline.c:3517-3523`), so it does not give in-place multi-line editing, its
`IsBalanced` counts `(` and `)` with no awareness of strings, comments, or
character literals (`(display "abc(")` strands the prompt), and it has no
highlight callback at all.

So: take the structural-editing commands, not the library.

## Source

`bestline.c`, BSD-2-Clause (compatible with Kaappi's MIT):

- `bestlineEditBarf` — line 3212
- `bestlineEditSlurp` — line 3286
- `bestlineEditRaise` — line 3335
- `bestlineEditRotate` — line 3040

Each is a self-contained transform over the edit buffer plus cursor position,
roughly 60-120 lines, with no dependency on the rest of bestline's state beyond
`l->buf` / `l->pos`. Standard keybindings there are `ALT-SHIFT-B` / `-S` / `-R`
and `ALT-y`.

## Dependency

Should land **after** the REPL moves to an editor with a real multi-line
buffer. Structural editing is much less useful when only the current physical
line is editable, which is the constraint today (each line is a separate
`linenoise()` call; see `src/repl.zig`). On a single multi-line buffer these
commands can operate across the whole form.

## Notes for whoever picks this up

- Kaappi's reader gives `[` and `]` no meaning (`0]` is `KP1002`), so unlike
  most paredit implementations these should pair round parens only. Note the
  highlighter's `findMatchingOpen` (`src/repl.zig`) *does* pair brackets —
  that divergence is pre-existing and worth resolving at the same time.
- Ported code must respect string literals, `;` line comments, `#| |#` block
  comments, `#;` datum comments, `#\(` character literals, and `|pipe|`
  symbols. bestline's originals do not; the port needs that logic, and
  `src/reader.zig` is the authority on where a datum ends.
- Attribution: keep bestline's copyright notice in the vendored/ported source.

baijum:
Closes #2216.

Four keys now move a paren rather than a character. `|` marks the cursor:

| Key | Command | Effect |
|---|---|---|
| alt+shift+S | slurp | `(a\| b) c d` → `(a\| b c) d` |
| alt+shift+B | barf | `(a\| b c)` → `(a\| b) c` |
| alt+shift+R | raise | `(+ 1 (* 2 \|3))` → `(+ 1 \|3)` |
| alt+y | rotate | `(if a\| b c)` → `(if b c a\|)` |

They are listed under `,help` and under isocline's own F1 help, the latter only
when the callback is set.

This had to wait for the isocline migration (#2219): with one editable physical
line per `linenoise()` call there was no whole form to restructure.

## Two corrections to the issue

The issue proposed porting the four commands from bestline, citing four line
numbers. **Two of them are not there to port:**

- `bestlineEditRaise` (line 3335, the line cited) is `(void)l;` — an empty stub.
- `bestlineEditRotate` (line 3040) rotates the **kill ring** and re-yanks —
  emacs `M-y` yank-pop — not the datums of a form.

Barf and slurp do exist, but they walk runes with a mirror stack and no notion
of strings, comments or character literals — precisely the bug class the issue
asked the port to avoid. So this is a reimplementation, not a port, and the
attribution in `PATCHES.md` and the module header says so: the design and the
keybindings are bestline's, the code is not.

## One judgment call worth a look

**Rotate keeps the head and cycles only the arguments.** The issue's table says
"rotate the datums of the enclosing form", which read literally turns
`(+ 1 2)` into `(1 2 +)` — never valid Scheme, so the command would be useless
in practice. Head-preserving rotation is still a pure cycle: repeating it n-1
times on n arguments restores the original. Easy to flip if the literal reading
was meant.

## What is where

**`src/repl_sexp.zig`** (new, 706 lines with tests) — the transforms as pure
functions over `(buffer, byte cursor)`, so they test without a terminal. The
scanner takes `Reader.isDelimiter` from `reader.zig` **directly rather than
copying it**, and skips strings, `;` and `#|…|#` comments, `#;` datum comments,
`#\(` character literals and `|pipe|` symbols. 31 unit tests, including the
cases the ported originals would have got wrong.

An unbalanced form has no paren to move, so every command declines and leaves
the buffer exactly as it was rather than guessing.

**KAAPPI PATCH 3** (`vendor/isocline/`) — `ic_set_default_sexp_edit`, the four
key bindings, and an F1 help section. isocline owns the keys and the buffer;
the host language owns the syntax and does the rewriting. One
`editor_start_modify` wraps the replacement so ctrl-Z undoes a structural edit
as a single step. With no callback set, behavior is exactly upstream's.

**Also closed: the bracket divergence the issue noted in passing.**
`findMatchingOpen` is gone (isocline does brace matching now, already limited
to `"()"`), but `scanHighlight` still painted `[`/`]` like parens while the
reader gives them no meaning (`0]` is KP1002). `repl.isDelimiter` is now an
alias of `Reader.isDelimiter`, so the colors, the brace matcher, the reader and
`repl_sexp` all agree.

## Tests

`tests/scheme/smoke/repl-structural-editing-2216.sh` drives all four keys over
a real pty — the C half is unreachable from Zig. Each case types a form, moves
the cursor inside it, presses one key, submits, and asserts on what the
*evaluator* printed, so the check does not depend on how the editor redraws.

**Mutation-tested:** with `edit_sexp` stubbed to return immediately, all four
cases fail. Two things that make it robust rather than timing-guessy are worth
knowing if it ever needs editing: it sets `TIOCSWINSZ` (a fresh pty is 0x0,
which leaves the editor no room to render), and it never uses the prompt string
as a post-submit sentinel — isocline redraws the prompt on every keystroke, so
waiting on it returns instantly and the next case then types into the tty's
canonical buffer. It waits for the printed result instead. No usable pty exits
77 (skip) rather than failing.

- `zig build test` — 1755/1758 pass (3 pre-existing skips)
- `bash tests/scheme/run-all.sh` — 2077 pass, 0 fail; the new pty test also
  passes under the suite's shell concurrency, and 6/6 standalone runs
- Cross-compiles clean for `x86_64-windows` and `wasm32-wasi`

## Not done

`src/repl.zig` is now 1571 lines, past the 1500-line policy. It was already at
1514 before this change; the natural fix is splitting the highlighter into its
own file, which would move ~340 lines and bury this diff. Worth a separate
pass.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -392,11 +392,13 @@ It holds a whole form in one buffer, which is why `repl.zig` no longer joins
 continuation lines: `ic_readline` returns a finished expression, newlines
 included, and every line of it stays editable until submit.
 
-**The copy is patched** — two changes, each marked `KAAPPI PATCH` in the source
-and documented in `vendor/isocline/PATCHES.md`: an input-completeness callback
-(upstream's Enter always submits) and a configurable history size (upstream
-caps at 200). Re-apply both when updating; `grep -rn 'KAAPPI PATCH'
-vendor/isocline/` finds every site.
+**The copy is patched** — three changes, each marked `KAAPPI PATCH` in the
+source and documented in `vendor/isocline/PATCHES.md`: an input-completeness
+callback (upstream's Enter always submits), a configurable history size
+(upstream caps at 200), and structural s-expression editing — slurp, barf,
+raise, rotate, whose transforms live in `src/repl_sexp.zig` (`docs/dev/repl.md`).
+Re-apply all three when updating; `grep -rn 'KAAPPI PATCH' vendor/isocline/`
+finds every site.
 
 ## Package manager (thottam)
 
--- a/docs/dev/repl.md
+++ b/docs/dev/repl.md
@@ -21,6 +21,7 @@ which `main.zig` never reaches.
 |---|---|---|
 | `ic_set_default_is_complete` | `isCompleteCallback` → `inputIncomplete` | Enter submits only when the **reader** says the form is finished. Not upstream isocline — see `vendor/isocline/PATCHES.md`. |
 | `ic_set_default_completer` | `completionCallback` | Comma commands match the whole line; everything else goes through `ic_complete_word` over `vm.globals`. `,load`/`,import` complete filenames. |
+| `ic_set_default_sexp_edit` | `sexpEditCallback` → `repl_sexp.apply` | The four structural-edit keys. Not upstream isocline — patch 3. |
 | `ic_set_default_highlighter` | `highlightCallback` → `scanHighlight` | Emits styled spans; `scanHighlight` is separated so the token rules are testable without a terminal. |
 | `ic_style_def` | `applyTheme` | `repl.color.*` names become isocline styles via `ansiToIcStyle`. `ic-prompt` and `ic-bracematch` are isocline's own names, redefined. |
 | `ic_enable_brace_matching` | — | Limited to `"()"`: the reader gives `[`/`]` no meaning (`0]` is KP1002). |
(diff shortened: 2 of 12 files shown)


baijum:
Went through the three review findings. Two taken, one declined on its stated rationale — and chasing the declined one found a worse bug than the one reported.

### Declined as stated: "exponential fork loop" (flagged 🔴 Critical)

> The child then reaches `pty.fork()` itself, and each new child repeats the failure. That is an exponential fork loop in CI.

This cannot happen here. `pty.fork()` is called **once at module top level, not inside any loop**. A raising `os.execv` propagates out of the `if pid == 0:` block, Python prints a traceback and the child exits — there is no control path back to `pty.fork()`.

Measured, with `KAAPPI=/nonexistent/kaappi`:

```
SKIP: no usable pty for the REPL here
EXIT=77
procs before=596 after=600
```

No growth, no stray processes.

I took the `os._exit(127)` guard anyway, for a different reason: `pty.fork()`'s contract is that the child execs or dies, and without it a traceback lands in the pty slave and pollutes the buffer the parent is matching against. Correct hygiene, wrong stated cause.

### The actual bug, which none of the three named

Look at that exit code again: **77**. `run-all.sh` reads that as a skip and reports the run green. A binary that never started made this test *silently not run* — much worse than a fork loop, which would at least be loud.

Fixed by separating two things that were being conflated:

- **no output at all** → the REPL did not start → exit 1
- **output but no prompt** → no usable terminal → exit 77 (still a skip; that is what keeps the emulated legs from flaking)

Each failure mode now returns the code it means:

| `KAAPPI` | before | after |
|---|---|---|
| `/nonexistent/kaappi` | 77 skip | **1 fail** |
| `README.md` (not executable) | 77 skip | **1 fail** |
| `/usr/bin/true` (execs, no output) | 77 skip | **1 fail** |
| bare name on `PATH` | 77 skip | **0 pass** |

### Taken: mark against the raw buffer

Valid, and the reasoning is right — stripped indices are not monotonic. A read ending mid-escape leaves bytes the regex cannot match yet; they vanish once the rest arrives, shifting every earlier index. Raw offsets only grow. `seen()` now slices `buf` first and strips the slice.

The practical impact was smaller than "intermittent failure" suggests — both needles sit far from the mark boundary, and `wait_for` re-evaluates in a loop — but the fix is two lines and removes the class.

### Taken, differently: bare command name

Rather than bail out, `$KAAPPI` is now resolved up front — no slash means a `PATH` lookup, anything else is a path as given — so the bare form works instead of erroring. The suggested `exit 77` would have deepened the silent-skip hole above, so this fails with a clear message instead.

### Re-verified

Mutation test still holds: with `edit_sexp` stubbed to return immediately, the suite exits 1 and all four cases report. Normal run passes 3/3 standalone.

PR #2225, opened 2026-08-05
Fix REPL comma-command TAB completion appending instead of replacing

### LINKED ISSUE #2224: REPL: TAB after a comma-command prefix appends instead of replacing (`,h` + TAB -> `,h,help`)
## Bug

In the interactive REPL, typing `,h` and pressing TAB to complete a comma-command produces `,h,help` instead of `,help`.

## Root cause

`completionCallback` in `src/repl.zig` handles comma-commands (`,help`, `,quit`, etc.) by calling `ic.addCompletion(cenv, cmd)` directly, bypassing `ic.completeWord`. `ic_add_completion` (vendor/isocline) defaults to `delete_before = 0, delete_after = 0` — it inserts the replacement at the cursor without deleting any of the already-typed text. The correct `delete_before` is only computed when completions are routed through `ic_complete_word`, which measures the word boundary and wraps the completion callback (`vendor/isocline/src/completers.c`, `token_add_completion_ex`). The filename-completion branch (`,load `, `,import `) already goes through `ic.completeFilename` and is unaffected; only the bare command-name branch used the direct `addCompletion` path.

## Fix

Route the command-name branch through `ic.completeWord` with a word-boundary rule that treats everything except a space as part of the "word" (so it includes the leading `,`), same as the existing scheme-identifier completion path does with `isIdentChar`.

baijum:
## Summary
- `,h` + TAB produced `,h,help` instead of `,help` — the comma-command completion branch in `completionCallback` (`src/repl.zig`) called `ic.addCompletion` directly, bypassing `ic.completeWord`. isocline's `ic_add_completion` defaults `delete_before` to 0, so it splices the replacement in at the cursor without deleting the already-typed prefix; that deletion is only computed correctly when routed through `ic_complete_word` (`vendor/isocline/src/completers.c`).
- Route the command-name branch through `ic.completeWord` with a word-boundary rule that treats everything but a space as part of the word (so it includes the leading `,`), mirroring how scheme-identifier completion already uses `isIdentChar`.
- Filename completion (`,load `, `,import `) already went through `ic.completeFilename` and was unaffected.

Fixes #2224

## Test plan
- [x] `zig build` / `zig build test` pass
- [x] New pty-driven regression test `tests/scheme/smoke/repl-comma-command-completion-2224.sh` — fails on the pre-fix binary (`,h,help`, `,d,d`) and passes on the fixed one
- [x] Existing `tests/scheme/smoke/repl-structural-editing-2216.sh` still passes (same pty-driving pattern, unaffected feature)

### DIFF
--- a/src/repl.zig
+++ b/src/repl.zig
@@ -135,6 +135,23 @@ fn completeWordCallback(cenv: ?*ic.CompletionEnv, prefix: [*c]const u8) callconv
     }
 }
 
+/// Word-char rule for comma-command names: only a space ends the "word", so
+/// the leading ',' is included and `ic.completeWord` deletes the whole typed
+/// command (not just the part after it) before splicing in the replacement.
+fn isCommandChar(s: [*c]const u8, len: c_long) callconv(.c) bool {
+    if (len != 1) return true;
+    return s[0] != ' ';
+}
+
+fn completeCommandNameCallback(cenv: ?*ic.CompletionEnv, prefix: [*c]const u8) callconv(.c) void {
+    const word = if (prefix) |p| std.mem.span(@as([*:0]const u8, @ptrCast(p))) else return;
+    for (&repl_commands) |cmd| {
+        if (std.mem.startsWith(u8, std.mem.span(cmd), word)) {
+            if (!ic.addCompletion(cenv, cmd)) return;
+        }
+    }
+}
+
 fn completionCallback(cenv: ?*ic.CompletionEnv, prefix: [*c]const u8) callconv(.c) void {
     const line = if (prefix) |p| std.mem.span(@as([*:0]const u8, @ptrCast(p))) else return;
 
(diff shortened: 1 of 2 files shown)


baijum:
@coderabbit review

baijum:
Responding to the two CodeRabbit findings:

**"Fail when the started REPL does not show a prompt" (exit 1 instead of 77).** Declining — this is intentional, not a bug. Both lines were copied from the already-merged `repl-structural-editing-2216.sh`, whose header explicitly documents the distinction: a REPL that wrote *nothing at all* is a genuine failure (exit 1), but one that wrote something and never prompted has no usable terminal here (exit 77/SKIP) — the alternative is flaking on every emulated CI leg with partial terminal support. Forcing exit 1 in both cases would reintroduce exactly the flakiness that design avoids.

**"Bound REPL shutdown and validate its exit status."** Good catch, applied in 9a7c44bc — added a bounded `waitpid(WNOHANG)` reap with a 10s deadline (kill + reap on timeout) and a check that `,quit` exits 0, so a future shutdown regression fails this test instead of hanging it. I verified empirically first that `,quit` currently exits cleanly in ~1s, so this is pure hardening, not a workaround for a real hang. Note the identical unbounded-`waitpid` shape also exists in `repl-structural-editing-2216.sh` (already merged); I left that file alone to keep this PR scoped to #2224, but it'd be a reasonable small follow-up.

PR #2227, opened 2026-08-05
Stop discarding pasted input when the REPL drops out of raw mode

### LINKED ISSUE #2226: REPL: pasting multiple top-level forms loses everything after the first
## Bug

Pasting a block of Scheme source containing more than one top-level form into
the REPL (e.g. copied from a browser/editor) only shows/evaluates the first
form. Everything after it silently disappears — no error, no partial output,
just gone.

## Repro

```
$ zig-out/bin/kaappi
kaappi> (define x 1)
(paste the rest of the block here, in one paste action:)
(define y 2)
(display (+ x y))
```

Only `(define x 1)` takes effect. `y` is left undefined:

```
kaappi> y
error[KP3001]: undefined variable 'y'. Did you mean '+'?
```

## Root cause

Many terminal emulators deliver a pasted newline as a literal `\r` (the same
byte a real Enter keypress sends), so isocline treats the boundary between
the first two pasted forms exactly like the user pressing Enter after typing
`(define x 1)`. Since that line alone is already a complete form,
`isCompleteCallback` submits right there and `ic_editline` returns — before
the rest of the pasted bytes have been `read()` at all; they're still sitting
in the pty's input queue.

`ic_editline` (`vendor/isocline/src/editline.c`) wraps every `ic_readline`
call in `tty_start_raw()` / `tty_end_raw()` (`vendor/isocline/src/tty.c`).
Both currently call `tcsetattr(..., TCSAFLUSH, ...)`. `TCSAFLUSH` discards
any input the kernel has already received but the process hasn't read yet —
so the moment the first form's `ic_editline` call returns and `tty_end_raw()`
runs, the rest of the paste (still unread, sitting in the tty's queue) is
thrown away. Confirmed with a pty-driven repro that pastes `(define x
1)\r(define y 2)\r(display (+ x y))\r` as one write(2) and observes `y`
undefined afterward.

`TCSADRAIN` (wait for pending output, but don't discard unread input) fixes
it without the discard side effect.

## Notes

- Present on the current isocline-backed REPL (post-#2219).
- Marked `priority: high`: not memory-unsafe, but it silently drops user
  input/data on a very common REPL workflow (paste-and-eval), with no
  indication anything was lost.

baijum:
## Summary

- Pasting a block of Scheme source with more than one top-level form into the REPL only evaluated the first form — everything after it silently vanished, with no error or indication anything was lost.
- Root cause: many terminals deliver a pasted newline as a literal CR (the same byte a real Enter keypress sends). isocline's `ic_editline` wraps every `ic_readline` call in `tty_start_raw()`/`tty_end_raw()`, both of which called `tcsetattr(..., TCSAFLUSH, ...)`. If the first pasted form is already complete on its own, `isCompleteCallback` submits right there and `ic_editline` returns *before* the rest of the still-unread paste has been read from the pty — and `tty_end_raw`'s `TCSAFLUSH` then discards it before the next `ic_readline` call gets a chance.
- Fix: switch both `tty_start_raw` and `tty_end_raw` (`vendor/isocline/src/tty.c`) from `TCSAFLUSH` to `TCSADRAIN`, which still waits for pending output but leaves unread input alone. Documented as Patch 4 in `vendor/isocline/PATCHES.md` and the root `CLAUDE.md`'s isocline section, per the existing convention for this vendored/patched copy.
- Left the `TCSAFLUSH` in the termination-signal handler (`sig_handler`) untouched — that path runs when the process is dying/suspending, where discarding stray input is correct and unrelated to this bug.

Fixes #2226

## Test plan

- [x] New pty-driven regression test `tests/scheme/smoke/repl-paste-multiple-forms-2226.sh`: pastes `(define x 1)\r(define y 2)\r(display (+ x y))\r` as a single `write(2)` (mirroring how a terminal delivers a clipboard paste) and asserts all three forms take effect.
- [x] Verified the new test **fails** against the pre-fix `tty.c` (`y` ends up undefined, `display` never prints `3`) and **passes** with the fix.
- [x] `zig build test` passes.
- [x] Existing REPL pty suites (`repl-structural-editing-2216.sh`, `repl-comma-command-completion-2224.sh`) still pass — no regression in raw-mode handling for typed (non-pasted) input.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -392,13 +392,15 @@ It holds a whole form in one buffer, which is why `repl.zig` no longer joins
 continuation lines: `ic_readline` returns a finished expression, newlines
 included, and every line of it stays editable until submit.
 
-**The copy is patched** — three changes, each marked `KAAPPI PATCH` in the
+**The copy is patched** — four changes, each marked `KAAPPI PATCH` in the
 source and documented in `vendor/isocline/PATCHES.md`: an input-completeness
 callback (upstream's Enter always submits), a configurable history size
-(upstream caps at 200), and structural s-expression editing — slurp, barf,
-raise, rotate, whose transforms live in `src/repl_sexp.zig` (`docs/dev/repl.md`).
-Re-apply all three when updating; `grep -rn 'KAAPPI PATCH' vendor/isocline/`
-finds every site.
+(upstream caps at 200), structural s-expression editing — slurp, barf,
+raise, rotate, whose transforms live in `src/repl_sexp.zig` (`docs/dev/repl.md`)
+— and `TCSADRAIN` instead of `TCSAFLUSH` around raw-mode transitions, so a
+multi-line paste that submits partway through doesn't get its still-unread
+tail silently discarded (kaappi#2226). Re-apply all four when updating;
+`grep -rn 'KAAPPI PATCH' vendor/isocline/` finds every site.
 
 ## Package manager (thottam)
 
--- a/tests/scheme/smoke/repl-paste-multiple-forms-2226.sh
+++ b/tests/scheme/smoke/repl-paste-multiple-forms-2226.sh
@@ -101,6 +101,11 @@ def idle(quiet=0.5, limit=5.0):
             return
 
 if not wait_for(b'kaappi> ', 0, 25):
+    # Two very different things look alike here, and conflating them is how a
+    # test goes quietly green: a REPL that wrote *nothing at all* did not run,
+    # which is a failure; a REPL that wrote something but never prompted has no
+    # usable terminal, which is a skip (the alternative is a flake on every
+    # emulated CI leg).
     if not buf:
         sys.stdout.write('the REPL produced no output at all; it did not start\n')
         sys.exit(1)
@@ -167,6 +172,8 @@ else:
 
 if exit_status is not None and os.WIFEXITED(exit_status) and os.WEXITSTATUS(exit_status) != 0:
     failures.append(('shutdown: REPL exited with status %d' % os.WEXITSTATUS(exit_status), b''))
+elif exit_status is not None and os.WIFSIGNALED(exit_status):
+    failures.append(('shutdown: REPL was killed by signal %d' % os.WTERMSIG(exit_status), b''))
 
 try:
     os.close(fd)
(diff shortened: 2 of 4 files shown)


baijum:
Went through CodeRabbit's three findings individually rather than applying them wholesale:

**1. "Move the isocline patch-maintenance documentation to `docs/dev/`" — declined.**
The flagged lines (`CLAUDE.md` 395-403) are the pre-existing "Dependencies" → isocline blurb; my diff only bumped "three changes" to "four" and appended one clause about the new patch. The actual maintenance detail already lives in `vendor/isocline/PATCHES.md` (linked from that paragraph) — the root `CLAUDE.md` text has only ever been a short summary + pointer, unchanged in structure since before this PR (`git show HEAD~2:CLAUDE.md` has the identical shape, just "three" instead of "four"). This finding would apply identically to `main` with no PR at all, so it's a pre-existing-condition flag rather than something introduced here, and restructuring doc placement is out of scope for a bug-fix PR.

**2. "Exit 1 instead of 77 when output was received but no prompt appeared" — declined, same as the identical finding raised (and declined) on #2225.**
This is byte-for-byte the same branch as `repl-comma-command-completion-2224.sh:103-108` and `repl-structural-editing-2216.sh:136-141` (both already merged), copied from the same template. `2216.sh` carries the rationale inline: a REPL that wrote *nothing* didn't start (fail), but one that wrote something and never prompted has no usable terminal in this environment (skip) — the alternative is flaking on emulated CI legs (QEMU riscv64/s390x/ppc64le) where terminal timing differs from native. Applying this suggestion here would make the new test behave inconsistently with its own siblings for no behavioral gain. I did carry the explanatory comment over from `2216.sh` into this file (it was missing here and in `2224.sh`) so the next reviewer doesn't have to rediscover the reasoning.

**3. "Also fail on signal-terminated shutdown" — accepted, fixed in 24064880.**
Real gap: the shutdown check only tested `os.WIFEXITED`, so a REPL that crashed via signal (SIGSEGV/SIGABRT/etc.) after `,quit` instead of exiting cleanly wouldn't be flagged. Added an `os.WIFSIGNALED`/`os.WTERMSIG` branch. Note the identical gap exists in the already-merged `repl-comma-command-completion-2224.sh` (`os.WIFEXITED`-only check added in #2225) — left that one alone to keep this PR scoped to its own new file, consistent with how #2225 left `2216.sh`'s pattern alone.

Verified after the fix: the new test still passes, and I re-ran it against the pre-fix `tty.c` with a full clean rebuild (`rm -rf zig-out .zig-cache`) to confirm it still fails there — a stale incremental build had briefly masked the regression during my own verification, which a from-scratch build resolved.

PR #2229, opened 2026-08-05
Give records and FFI handles an identity that survives a heap copy

### LINKED ISSUE #1932: A record returned through thread-join! loses its type: prints correctly, but pt? is #f and every accessor raises

### LINKED ISSUE #2027: Deep copy aliases FFI handles across heaps: a child-created ffi-fn/ffi-library is freed under the receiver and reads back as (0.0 . 0.0)
**Category:** GC / FFI / SRFI-18 threads · **Severity:** crash (use-after-free, silent type confusion) · **Found by:** Systematic audit v2, Phase 2.7 (#1890)

`gc_deep_copy.zig` aliases FFI handles instead of copying or refusing them. The receiving heap gets a raw pointer into the *sending* heap, and neither collector accounts for it — so the object is freed while the receiver still holds it.

```zig
// src/gc_deep_copy.zig:461-465
// FFI objects wrap process-global dlopen handles that cannot be
// duplicated per-heap, so they are aliased. Known limitation: an FFI
// handle created inside a child thread and returned through
// thread-join! dangles once the child heap is freed.
.ffi_library, .ffi_function => src,
```

The comment anticipates the hazard but understates it in three ways: it is **not** limited to `thread-join!` (all three copy boundaries are affected), it does **not** require the child heap to be freed (a live child's own collector frees the object first), and the result is **not** a clean dangle — the recycled slot is read as a well-formed value of a different type.

## Minimal reproduction

```scheme
  (ffi-fn (ffi-open "libc.dylib") "abs" '(int) 'int))))
(thread-start! t)
(define r (thread-join! t))
(list 'procedure? (procedure? r) 'raw r)
```

**3/3 runs**, ReleaseSafe and `-Dgc-stress=true` alike:

```text
procedure?  #f
raw         (0.0 . 0.0)      <- the freed slot, re-read as a pair of flonums
calling it  KP3005: not a procedure
```

## Discriminating control

The cliff is exactly **which heap owns the object**, not FFI itself:

| # | who allocates the handle | how it crosses | result |
|---|---|---|---|
| A | **parent** | child calls it via lexical capture | `5` — correct |
| A' | **parent** | parent re-calls it after the join | `6` — correct |
| B | **parent** | child captures **and returns** it | `procedure? #t`, callable, `8` — correct |
| C | **child** | child returns it | `procedure? #f`, `(0.0 . 0.0)` |

A and B alias a *parent*-heap object, which outlives the child. Only C aliases a *child*-heap object. Same code path, same arm, opposite outcome.

## All three copy boundaries, not just thread-join!

`channel-send` is affected too, and there the child is **still running** when the parent receives — so this is not "the child heap was freed at join":

```scheme
(let ((ch (make-channel)))
  (let ((t (make-thread (lambda ()
             (channel-send ch (ffi-fn (ffi-open "libc.dylib") "abs" '(int) 'int))
             'sent))))
    (thread-start! t)
    (let ((got (channel-receive ch)))    ; child still alive here
      (thread-join! t)
      (list 'procedure? (procedure? got) 'raw got))))
;; => (procedure? #f raw (0.0 . 0.0))   3/3
;; control on the same channel: (channel-send ch (list 1 2)) => (1 2)   3/3
```

## Mechanism

`gc_collect.zig` skips foreign-owner objects while marking (`:446`, `:731` — `if (obj.owner != gc.id) return`). The alias therefore creates a cross-heap reference **neither** collector can see: the parent never marks it (foreign), and the child never finds a root for it (the parent's reference is invisible to the child). The child's own collector reclaims it on its next cycle, whether or not the child has exited. Marked INFERRED from the source; the observable behaviour above is VERIFIED.

This is why the failure is deterministic rather than timing-dependent, and why `-Dgc-stress=true` changes nothing.

## Why it is worse than a dangle

The value passes `write` and every non-FFI type check without complaint — it reads back as `(0.0 . 0.0)`, an ordinary pair. Code that stores the handle and calls it later fails at an arbitrary distance from the boundary, with `KP3005: not a procedure` rather than anything naming a thread or the FFI.

The same recycled-slot shape was seen from a different root cause in Phase 5D (`mutex-state` handing out a child fiber, noted in `tests/scheme/audit/primitives_srfi18-audit.scm`'s epilogue, not filed). Related but distinct: that is a primitive handing out a foreign pointer, this is the deep-copy arm creating one.

## Suggested direction (not applied — audit campaign files, does not fix)

Two of the three obvious options are already precedented in this file:

- **Refuse**, joining the 14-tag list — consistent, and matches `ffi_callback`, which is *already* refused. Breaks control B, which works today.
- **Promote and refcount**, the `.channel` arm's shape (`shared_channel.promoteChannel` + `retain`) — preserves both controls, but needs a process-global owner for the `dlopen` handle.

A third, cheapest option: keep the alias but only when `obj.owner == gc.id` for the *destination*, refusing otherwise. That preserves A/B exactly and turns C into a clean catchable refusal.

**Verified:** fresh ReleaseSafe build at `d0d03deb` (v0.22.1), macOS aarch64, isolated `KAAPPI_HOME`, `kaappi cache clear`; 3/3 at each of the three boundaries; `-Dgc-stress=true` **unchanged**.
**Source location:** `src/gc_deep_copy.zig:465` (`.ffi_library, .ffi_function => src`); foreign-owner mark skip at `src/gc_collect.zig:446,731`
**Disabled test:** `tests/scheme/audit/srfi18-deepcopy-matrix-audit.scm` (two assertions marked `;; FAIL: #this`; an enabled assertion pins `(not (procedure? ...))` so the fix flips it)
**Platform note:** reproduced on macOS via `libc.dylib`. The arm is platform-independent; the committed test probes a list of library names and skips where none loads.

baijum:
Closes #1932
Closes #2027

Both issues are the same mistake in `src/gc_deep_copy.zig`, reached from opposite directions: a value's identity was its **address**, and every thread boundary copies into a heap where the address is necessarily different.

## #1932 — a record loses its type through `thread-join!`

Record type identity *was* the `RecordType` pointer. The copy is a new object, so it was a disjoint type: the value printed as a well-formed `#<<pt> 1 2>` while `pt?` answered `#f` and every accessor raised `expected <pt>, got #<record_instance>`. A `cond` dispatching on the predicate silently took the wrong branch.

`RecordType` gains an `identity: u64` from a process-global atomic counter — minted at definition by the two allocators, carried verbatim by both `.record_type` copy paths. `types.sameRecordType` replaces the four pointer comparisons behind `%record?`, `%record-ref`, `%record-set!` and SRFI 237's `isOrDescendsFrom`.

Generativity is untouched: two evaluations of a `define-record-type` form still mint two identities. That is the control the tests pin, in both the local and the post-copy direction — a fix keyed on name or shape would pass every positive assertion and fail it.

All four boundaries are fixed, not just the one in the repro: thunk capture (a captured predicate rejecting a genuine parent-heap instance), the join result, the uncaught-exception path, and a channel message.

## #2027 — a child-created FFI handle is freed under the receiver

The `.ffi_library, .ffi_function => src` arm aliased, on the reasoning that a dlopen handle cannot be duplicated per-heap. True of the handle — but the *wrapper* is an ordinary object owned by one GC, and `gc_collect` skips foreign-owner objects while marking. The receiver held a reference neither collector could see; the sender's own collector reclaimed it, **whether or not the sender had exited** (so `channel-send` was affected too, with the child still running), and the recycled slot read back as `(0.0 . 0.0)` — an ordinary pair that passes `write` and every non-FFI type check, so a stored handle failed later with `KP3005: not a procedure` at an arbitrary distance from the thread that produced it.

The wrapper is now copied, exactly as `.native_fn` copies the object around a static function pointer; the process-global handle and symbol address are shared by value. No second `dlopen`, nothing process-global duplicated.

The issue's other option — refusing the tag, joining the 14-tag list — is why controls A/A'/B are in the test file: a parent-owned handle captured (and returned) by a child works today and a refusal would have broken it. `ffi_callback` stays refused; it wraps a live Scheme closure, not a process-global address.

**One behaviour change worth flagging:** `ffi-close` nulls the handle in one wrapper only, so a copy on another heap no longer reports the library as closed. `dlclose` while another thread is calling into the library was already undefined, so this moves where it is diagnosed rather than whether it is safe. Documented in the arm's comment, in `docs/dev/thread-value-sharing.md`, and in the changelog.

## Tests

New:
- `tests/scheme/srfi/srfi18-record-identity-1932.scm` — 12 assertions over all four boundaries plus the generativity controls.
- `tests/scheme/ffi/thread-boundary-2027.scm` — 11 assertions: the issue's full A/A'/B/C matrix, the channel boundary with the child still alive, and the `ffi-callback` refusal.
- 5 Zig unit tests in `src/tests_deepcopy.zig`, including the "source heap is freed right after the copy" shape the `#958` `native_fn` test already uses.

Both new `.scm` files were A/B'd against a binary built from the parent commit: `srfi18-record-identity-1932.scm` goes 12 pass → 4 pass / 4 fail with hard errors, `thread-boundary-2027.scm` 11 pass → 8 pass / 1 fail (it aborts at cell C).

Re-enabled, in four files:
- `audit/srfi18-deepcopy-matrix-audit.scm` — three `#1932` rows and two `#2027` rows; the class table at the top moves `ffi_library`/`ffi_function` from **aliased** to **copied** (26 + 1 + 14 = 41).
- `audit/cross-thread-boundary-audit.scm` — the plain-record join and the captured-predicate rows.
- `audit/primitives_srfi237-audit.scm`, `audit/primitives_srfi181-audit.scm`, `srfi/srfi-notest-batch.scm` — a generative rtd, a transcoder, and an SRFI 111 box (a spec-mandated record type reached by the same defect).

Two bug-presence pins are **replaced, not inverted**, per `docs/audit-strategy.md`'s "never assert that a bug is still present" — #2027's own pin (`the aliased handle arrives unusable`, green on macOS, red on FreeBSD after merge) is the cautionary case that rule was written from.

## Verification

- `zig build test` — 1701 pass, 7 skip
- `zig build test -Dgc-stress=true` — 1701 pass, 0 fail
- `bash tests/scheme/run-all.sh` — 2080 pass, 0 fail (685 Scheme files + the 1395-test R7RS suite)
- `zig fmt --check src/`, markdownlint — clean

macOS aarch64. The FFI arm is platform-independent; the new `.scm` test uses `libm`/`ucrtbase` the way `ffi/basic.scm` does, so it runs on every leg.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,41 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- **A record returned from a thread keeps its type** (#1932). `(thread-join!
+  t)` on a thunk returning a `define-record-type` instance handed back a value
+  that *printed* as a well-formed `#<<pt> 1 2>` while its predicate answered
+  `#f` and every accessor raised `expected <pt>, got #<record_instance>` — so
+  a `cond` dispatching on the predicate silently took the wrong branch. Record
+  type identity was the `RecordType` address, and every thread boundary deep-
+  copies into a separate heap, which necessarily changes it; a lexically
+  captured predicate failed the same way on the way *in*, and the uncaught-
+  exception path a third time. Identity is now a process-global counter
+  carried across the copy, so all four boundaries (thunk capture, join result,
+  raised object, channel message) preserve it. Generativity is untouched: two
+  evaluations of a `define-record-type` form still make two distinct types.
+  A `nongenerative` (SRFI 237 uid) type was the one working case before, and
+  still works.
+
+- **An FFI handle created on a child thread is no longer freed under the
+  receiver** (#2027). `gc_deep_copy.zig` aliased `ffi-library` and
+  `ffi-function` across heaps instead of copying them, on the reasoning that a
+  dlopen handle cannot be duplicated per-heap — true of the handle, but the
+  *wrapper* is an ordinary object owned by one GC, and marking skips
+  foreign-owner objects. The receiver therefore held a reference neither
+  collector could see: the sender's own collector reclaimed it, whether or not
+  the sender had exited, and the recycled slot read back as `(0.0 . 0.0)` — an
+  ordinary pair that passes `write` and every non-FFI type check, so a stored
+  handle failed later with `KP3005: not a procedure`, at an arbitrary distance
+  from the thread that produced it. The wrapper is now copied and the
+  process-global handle and symbol address shared by value, at all three copy
+  boundaries. A handle created on the *parent* and captured by a child kept
+  working throughout and still does — which is why refusing FFI handles
+  outright was not the fix. `ffi-callback` remains refused: it wraps a live
+  Scheme closure, not a process-global address. One consequence: `ffi-close`
+  now nulls the handle in one wrapper only, so a copy on another heap does not
+  see the library as closed — closing a library another thread is calling was
+  already undefined behaviour, and this moves where it is diagnosed.
+
 - **SRFI 42's generic `:` qualifier works, along with `:real-range`,
   `:char-range`, `:dispatched`, and the full dispatch machinery** (#2177).
   `(list-ec (: i 5) i)` — the first form in every SRFI 42 tutorial — was
--- a/docs/dev/thread-value-sharing.md
+++ b/docs/dev/thread-value-sharing.md
@@ -66,6 +66,27 @@ its unpromoted twin failed, so a thread that read one out of a shared
 global could still hand it down to a child, and that child got a working
 stub for a channel nobody gave it.
 
+FFI handles are not on that list either — `ffi-library` and `ffi-function`
+are **copied**, and only the callback is refused. The dlopen handle and the
+symbol address are process-global and are shared by value; it is the
+wrapper object that gets duplicated into the receiving heap, exactly as
+`native-fn` duplicates the object around a static function pointer. Until
+kaappi#2027 the arm aliased them instead, which handed the receiver a
+pointer neither collector accounts for — a child-created handle was freed
+by the child's own collector, running or not, and read back in the parent
+as `(0.0 . 0.0)`. One consequence to know: `ffi-close` nulls the handle in
+one wrapper only, so a copy on another heap no longer reports the library
+as closed. (Closing a library another thread is calling was already
+undefined; this moves where it is diagnosed, not whether it is safe.)
+
+Record types are copied, and the copy is **the same type**. Type identity
+is `RecordType.identity`, a process-global counter carried across every
+copy boundary, not the RecordType address — which each copy necessarily
+changes. Until kaappi#1932 it was the address, so a record returned by
+`thread-join!` printed as a well-formed `#<<pt> 1 2>` while `pt?` answered
+`#f` and every accessor raised. Generativity is unaffected: two evaluations
+of a `define-record-type` form still mint two identities.
+
 A refusal surfaces as `error.UncopyableType`, which the boundary turns
 into a catchable error naming neither the type nor the offending value:
 
@@ -126,6 +147,8 @@ with a top-level `define` and named by the thunk.
 | `scheme-environment` | refused | works (`eval` in it succeeds) | unchecked |
 | `file-info` / `user-info` / `group-info` | refused | works | unchecked |
 | directory object | refused | works | unchecked |
+| ffi-library / ffi-function | **copied** | works | **coherent** since kaappi#2027 — the wrapper crosses, the process-global handle is shared by value |
+| record type / instance | **copied, same type** | works | **coherent** since kaappi#1932 — identity is a counter carried by the copy, not the address |
 | guardian | refused | refused (kaappi#2008) | **coherent** — the raw-container mutation made a check mandatory |
 | ephemeron | refused | works | unchecked — bound to one GC's collection cycle, but nothing mutates a raw container through it |
 | transport cell | refused | — | unreachable from Scheme: on this non-moving collector a transport-cell guardian always yields `#f` |
@@ -257,3 +280,12 @@ duplicating: `src/tests_deepcopy.zig` asserts the port and continuation
 refusals at the `GC.deepCopy` level directly, and
 `src/tests_shared_channel.zig` asserts that a channel *message* carrying a
 port is refused the same way.
+
+The two copied rows added above have their own regression suites, each
+covering all four boundaries (into the child, out of it, the raise path, a
+channel message): `tests/scheme/srfi/srfi18-record-identity-1932.scm` and
+`tests/scheme/ffi/thread-boundary-2027.scm`. Both carry the control that
+constrains the fix — a second look-alike record type that must stay
+disjoint, and the parent-owned FFI handle a blanket refusal would have
+broken. `tests/scheme/audit/srfi18-deepcopy-matrix-audit.scm` remains the
+per-tag enumeration.
--- a/docs/dev/thread-value-sharing.md
+++ b/docs/dev/thread-value-sharing.md
@@ -131,7 +131,10 @@ usable from a child thread through a global.
 
 ## The actual matrix
 
-Verified at `e24e594e`, ReleaseSafe, isolated `KAAPPI_HOME`. "capture"
+Verified at `e24e594e`, ReleaseSafe, isolated `KAAPPI_HOME` — except the
+two **copied** rows, which describe behaviour that commit predates and
+which were verified on kaappi#1932 / kaappi#2027's own branch. Their
+regression suites are named under "Keeping it honest" below. "capture"
 means bound in a `let` and closed over by the thunk; "global" means bound
 with a top-level `define` and named by the thunk.
 
--- a/src/gc_alloc.zig
+++ b/src/gc_alloc.zig
@@ -361,6 +361,7 @@ pub fn allocRecordType(self: *GC, name: []const u8, num_fields: u8) !Value {
         .header = .{ .tag = .record_type },
         .name = owned_name,
         .num_fields = num_fields,
+        .identity = types.nextRecordTypeIdentity(),
     };
     self.finishAlloc(&rt.header, @sizeOf(RecordType) + name.len);
     return types.makePointer(&rt.header);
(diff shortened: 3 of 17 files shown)


baijum:
Went through all seven findings and verified each against the code rather than taking it on trust. **Five accepted, two declined.** One was a genuine use-after-free in my own test.

### Accepted

**`src/tests_deepcopy.zig` — root `rt_val` (Major).** Correct, and worse than "a rule says to". `rt` is dereferenced for `identity` after two further `gc1` allocations; the bare `GC.init` in this test has no root marker, and `allocRecordTypeExtended` roots only its `parent` argument. I probed it directly under `-Dgc-stress=true`:

```
PROBE tag=<garbage> ... panic: invalid enum value
```

The header is gone by the last assertion — the test was passing only because the garbage `identity` happened not to collide with a real one. Rooted `rt_val` and `ext_val` (strictly nested, and `arg_roots` is a separate mechanism from the `pushRoot` stack, so the `defer`s are safe here). Re-ran `-Dgc-stress=true -Doptimize=Debug`, where freed memory is poisoned: 36/36 pass.

**Two assertions that could not fail.** Both confirmed by feeding them a deliberately raising thunk:

| shape | on a failing join |
|---|---|
| `(not (mrec2? (join-thunk ...)))` | **passes** — `mrec2?` says `#f` to the `(RAISED ...)` list |
| `(and (mrec? r) (not (mrec2? r)))` | fails, correctly |
| `(on-thread ...)` as a truth value | **passes** — `(raised . msg)` is truthy |
| `(eq? #t (on-thread ...))` | fails, correctly |

The look-alike row is the control that constrains the entire fix, so a vacuous pass there was the worst place for one. Both tightened. Re-ran the two audit files against a pre-fix binary: the look-alike row now fails there, where before it passed silently.

**Section E header (outside-diff).** Right, and a straight miss — I updated the cells but not the section header, which still described `=> src` and "the parent is left holding a freed object", contradicting this file's own class table twelve lines above. Rewritten.

**Channel handshake.** Agreed, and it is more than cosmetic: `thread-sleep!` guarantees nothing, so on a loaded runner that cell degrades into cell C and stops testing the thing its comment claims. Replaced with a second channel. Deadlock-free from both sides: the child's send is unconditional (a `guard` sends a sentinel on failure) and the parent's `go` is sent after a total `guard`, so no assertion can escape before it.

**Doc provenance line.** Fair — the two new rows post-date `e24e594e`. Noted separately rather than restating the whole table's provenance.

### Declined

**Split `tests_gc_tracing.zig`.** It was 1730 lines before this branch touched it; this PR adds **one line** to a pinned field inventory. Splitting it is unrelated to either issue and would bury a two-issue fix under an unreviewable file move. Worth its own issue.

**Equate identities on the uid-reuse path.** The premise is accurate — `existing.identity != rt.identity` — but it has no observable consequence, and the proposed fix would introduce a real bug.

No consequence, because the `.record_instance` arm retypes the instance to whatever `deepCopyValue` returned for its type, which *is* `existing`. So the destination sees one coherent type. I probed all five nongenerative shapes, including the one that would expose a split (the rtd and an instance crossing together in a single value):

```
(OUT-pred #t field 1 rtd-eq-parent #t)
(IN (#t 3))
(PAIRED-rtd-eq #t pred #t accessor-via-crossed-rtd 5)
(EXC #t)
(CHAN #t 10)
(cross-type #f #f)
```

All correct. And overwriting `existing.identity` with the source's would **silently retype every instance already living under `existing`** in the destination heap — turning a benign internal divergence into exactly the class of bug #1932 was. Not carrying identity there is deliberate.

Widening `sameRecordType` to compare uids is the other option offered. R6RS does say same-uid-plus-same-fields is the same type, but `sameRecordType` cannot cheaply check field equivalence, and the registry already guarantees one `RecordType` object per uid within a process — so it would add a weaker rule with no case to serve.

### Verification after the changes

- `zig build test` — 1701 pass
- `zig build test -Dgc-stress=true -Doptimize=Debug` (deepCopy filter) — 36 pass
- `bash tests/scheme/run-all.sh` — 2081 pass, 0 fail, 0 timeout
- markdownlint clean

Previous head was green on all 24 checks.

Issue #2129, opened 2026-08-01
[SRFI-18] thread-join! frees the joined thread's GC/VM while a thread it spawned is still starting — SIGSEGV; makes (srfi 120) unusable from a thread


baijum:
Found by audit v2 Phase 5A (validate-or-retire the SRFI 120 corruption claim, #1890).
It is *not* the claim under audit — that one is genuinely retired — but it is a live
crash reached from the same library, through an entry path nobody had written down.

## Summary

`thread-join!` unconditionally destroys the joined thread's `GC` and `VM`
(`freeChildResources`). If that thread had itself called `thread-start!`, and the
thread it started is still inside `threadEntryFn`'s prologue, the grandchild is
using the very `GC`/`VM` being destroyed — it holds `parent_vm` and
`parent_vm.gc` for `GC.initForThread` / `VM.initForThread` and for the deep copy
of its own thunk. The process dies with a SIGSEGV or a Zig panic.

No uncopyable value, no channel, and no `(srfi 120)` is required. A thread that
spawns a thread and returns is enough.

## Reproduction

```scheme
;; m7.scm
(import (scheme base) (scheme write) (srfi 18))
(let ((t (make-thread
          (lambda ()
            (let ((g (make-thread (lambda () (thread-sleep! 0.5) 'g))))
              (thread-start! g)
              'plain)))))          ; <- middle thread returns immediately
  (thread-start! t)
  (write (thread-join! t)) (newline))
(display "M7-DONE\n")
```

`v0.22.1` (aarch64-macos), fresh `zig build`, isolated `KAAPPI_HOME`:

| build | crash | clean |
|---|---|---|
| ReleaseSafe | **22** | 3 |
| `-Dgc-stress=true` | **13** | 2 |

(25 and 15 runs respectively.) `M7-DONE` is never printed on a crashing run; the
value `plain` usually is, so the join has already returned.

Two crash signatures alternate run to run:

```
thread 40358816 panic: incorrect alignment
std/hash_map.zig:784:44: in header (kaappi)
std/hash_map.zig:1088:30: in getAdapted__anon_115050 (kaappi)
```

```
Segmentation fault at address 0x0
std/mem/Allocator.zig:142:26: in rawAlloc (kaappi)     <- null vtable
std/hash_map.zig:1295:30: in growIfNeeded (kaappi)
std/hash_map.zig:1026:52: in putContext (kaappi)
[...]

Issue #2127, opened 2026-08-01
[GC] The gc-stress UAF detector is blind to child-heap teardown: GC.deinit drains the quarantine, so a dangling cross-heap value is never caught


baijum:
**Category:** GC / Tooling
**Severity:** coverage gap (silent — a green stress run proves nothing for this class)

The deterministic use-after-free detection added by #1687 / #1690 —
`FREED_OWNER` stamping plus the gc-stress free-quarantine — cannot see a UAF
that arises from **child-heap teardown**, which is the single most
UAF-prone area in the codebase (#1924, #1933, #2008, #2027 are all cross-heap
criticals).

The mechanism is one line, and it is deliberate: `GC.deinit` drains the
quarantine before returning.

```zig
// src/memory.zig:299-302
for (self.foreign_symbols.items) |o| gc_collect.freeObject(self, o);
// The freeObject calls above append to the quarantine in gc-stress builds;
// give every slot back before the allocator is torn down.
self.quarantineDrain();
```

`quarantineDrain` releases every withheld slot back to the allocator
(`src/memory.zig:585`). So for a child heap torn down by `thread-join!`:

1. `freeChildResources` (`src/primitives_srfi18.zig:933`) calls
   `child_gc.deinit()`.
2. `deinit` runs `freeObject` over every child object →
   `poisonAndDestroy` stamps `FREED_OWNER` and quarantines the slot.
3. `deinit` then **drains** the quarantine — every slot goes back to the
   allocator.
4. The next parent allocation reuses the slot and overwrites the sentinel.
5. The parent's next mark sees a valid, live-looking object. The
   `markValueInner` panic that #1687 exists to raise never fires.

The drain itself is correct — those slots would otherwise leak when the
allocator is torn down. The gap is that nothing distinguishes "this GC is
being destroyed at process exit" from "this GC is a joined child whose
objects a **live parent** may still reference".

## Reproduction

Any dangling cross-heap value works; #2125 is a live one. Capture a child's
fiber via `mutex-state`, join, then touch it:

```scheme
(import (scheme base) (scheme write) (srfi 18))
(define m (make-mutex 'probe))
(define t (make-thread (lambda () (mutex-lock! m) (thread-sleep! 0.4))))
(thread-start! t)
(thread-sleep! 0.15)
(define owner (mutex-state m))
(thread-join! t)
(write (list 'pair? (pair? owner))) (newline)
(set-car! owner 'clobbered)
(write owner) (newline)
```

```console
[...]

Issue #878, opened 2026-07-02
SRFI-18: thread-join! ignores timeout/timeout-val for OS threads (blocks until completion or forever); never-started threads return void immediately


baijum:
## What's wrong

`thread-join!`'s `timeout` and `timeout-val` arguments are completely ignored for started (OS-backed) threads: the call blocks until the child finishes, no matter what timeout was given, and blocks **forever** if the child never finishes. For never-started threads, `thread-join!` returns void immediately instead of waiting.

## Affected code

`src/primitives_srfi18.zig`, `threadJoinFn` (lines 361–452):

- The OS-thread path (lines 376–419) calls `thread.join()` unconditionally — **before** the timeout arguments are even parsed. Parsing at lines 422–431 only happens on the fiber path.
- For a never-started thread (`os_thread == null`, status `.created`), `runSchedulerUntilDone` (line 480) breaks immediately because the fiber was never added to any scheduler (`makeThreadFn` does not call `addFiber`), and control falls through to `threadJoinResult`, returning the fiber's initial `result` (void).

## Reproduction

```scheme
(import (scheme base) (scheme write) (srfi 18))
(define (spawn-slow)                       ; ~2s busy loop (thread-sleep! is broken in children, see #876)
  (thread-start! (make-thread (lambda ()
    (let loop ((i 0)) (if (< i 30000000) (loop (+ i 1)) 'done))))))

(define t (spawn-slow))
(define s0 (time->seconds (current-time)))
(define r (thread-join! t 0.2 'timed-out))
(define s1 (time->seconds (current-time)))
(write r) (newline)                        ; prints: done      (expected: timed-out)
(display (- s1 s0)) (newline)              ; prints: 2.125     (expected: ~0.2)

(define t2 (spawn-slow))
(guard (e ((join-timeout-exception? e) (display "join-timeout") (newline)))
  (write (thread-join! t2 0.2)))           ; prints: done      (expected: join-timeout raised)

;; never-started thread:
(define t3 (make-thread (lambda () 42)))
(write (thread-join! t3 0.3 'timed-out))   ; prints void immediately (expected: 'timed-out after ~0.3s)
```

Observed on macOS aarch64, v0.11.1 (`d46986f`). If the child never terminates (e.g. an infinite loop), `(thread-join! t 0.5 'x)` hangs forever.

## Expected (SRFI-18)

> "If the timeout is reached, thread-join! returns timeout-val if it is supplied, otherwise a 'join timeout exception' is raised."

[...]

PR #2230, opened 2026-08-05
Fix four SRFI-18 concurrency bugs from the v0.22.2 audit (#2129, #2194, #1982, #2125)

### LINKED ISSUE #1982: thread-terminate! cannot interrupt a thread parked in a native SRFI-18 wait — thread-join! then hangs forever (residual of closed #880)
**Category:** SRFI-18 threads · **Severity:** hang · **Found by:** systematic audit v2, Phase 5D (#1890)

```scheme
(define t (make-thread (lambda () (thread-sleep! 60))))
(thread-start! t) (thread-sleep! 0.2) (thread-terminate! t)
(thread-join! t)          ;; never returns
```

**Hangs 3/3** (SIGALRM at 12s). The **control completes 3/3**: an identical program whose child runs a Scheme spin loop instead of `thread-sleep!` terminates and joins normally.

## Every native wait is affected

| shape | result |
|---|---|
| `thread-sleep! 60` + terminate + join | **hang 3/3** |
| `mutex-lock! m 60` + terminate + join | **hang 3/3** |
| `mutex-unlock! m cv 60` + terminate + join | **hang 3/3** |
| untimed `mutex-lock!` + terminate + join | **hangs forever 3/3** |
| CONTROL: Scheme spin loop | completes 3/3 |
| CONTROL: `mutex-lock! 60` released by parent unlock | completes 3/3 |
| CONTROL: condvar released by broadcast | completes 3/3 |

## This is a residual of closed #880

#880 — *"thread-terminate! cannot stop a running OS thread; thread-join! blocks forever"* — was closed by #933, which polls `VM.terminate_flag` at the 1024-instruction safepoint in `vm_dispatch.runUntil`. That safepoint is reached **only while executing bytecode**.

`threadSleepFn`, `mutexLockFn` and `mutexUnlockFn`'s condvar branch each run their own native `while (true)` / `runSchedulerStep` loop, and **none consults `fiber.terminated`**.

**#880's own verification used a `(thread-yield!)` busy loop** — precisely the case #933 handles. The native-wait case was never covered, which is why it survived the fix.

## The latency ladder, measured

Per terminate + join:

| child is doing | latency |
|---|---|
| Scheme busy loop | 0.17 ms |
| `(thread-yield!)` loop | 0.30 ms |
| `(thread-sleep! 0.001)` loop | 394 ms |
| `(thread-sleep! 0.01)` loop | **3,758 ms** |
| single native wait | **never** |

That gradient is itself worth knowing: termination latency is proportional to how much *bytecode* the child executes, so a mostly-sleeping thread is nearly uninterruptible even before it hits the never case.

## Suggested direction

The native wait loops need the same `terminated` check the bytecode safepoint has — most cheaply by having `runSchedulerStep`'s wait predicates consult `fiber.terminated` alongside their own completion condition.

**Verified:** fresh ReleaseSafe build, isolated `KAAPPI_HOME`; all four hang shapes and all three controls reproduced with run counts.

### LINKED ISSUE #2125: [SRFI-18] mutex-state reports the owning thread's internal fiber, not the thread — the owner it returns is not eq? to any thread the caller holds
**Category:** SRFI
**Spec reference:** [SRFI 18](https://srfi.schemers.org/srfi-18/srfi-18.html), `mutex-state` — "returns … **the thread** that owns the mutex, if the mutex is in the locked/owned state"
**Severity:** wrong-result (spec deviation)

`mutex-state` on a mutex locked by another SRFI-18 thread returns that thread's
internal **fiber** object, not the thread handle `make-thread` returned. The
value is not `eq?` to any thread the caller holds, so the owner it reports
cannot be used for anything — `thread-join!`, `thread-terminate!` and identity
comparison against a known thread all fail on it.

## Minimal reproduction

```scheme
(define m (make-mutex 'probe))
(define t (make-thread (lambda () (mutex-lock! m) (let loop () (loop)))))
(thread-start! t)
(thread-sleep! 0.2)
(write (mutex-state m))          ;; expected: t   actual: #<fiber 0 running>
(write (eq? (mutex-state m) t))  ;; expected: #t  actual: #f
```

```console
unlocked:      not-abandoned
held by child: #<fiber 0 running>
eq? to the thread object: #f
after terminate: abandoned
```

Note the printed identity: the parent's own thread handle writes as
`#<fiber 1 running>`, and the owner comes back as `#<fiber 0 running>` — fiber 0
of the **child's** heap. So this is not merely the wrong *kind* of object; the
parent is handed a value that belongs to another heap.

## Discriminating controls

- The two unowned states are correct: `'not-abandoned` before the lock and
  `'abandoned` after `thread-terminate!` (verified in the same run above).
- The deviation is specific to the owned state, and specific to a *cross-thread*
  owner — the value is only reachable at all because the owning thread is not
  the caller.

## Why it matters beyond conformance

It silently defeats the obvious way to write "wait until the child holds the
lock", which is exactly what a test synchronising on an event rather than a
`thread-sleep!` wants to do:

```scheme
(eq? (mutex-state m) t)   ;; never true, so this spins to its retry budget
```

`tests/scheme/srfi/srfi18-cross-heap-abandoned-mutex.scm` therefore has to ask
"is the state neither of the two unowned symbols" instead, with a comment
pointing here. That workaround is the reason this was found.

**Verified:** fresh ReleaseSafe build at `18b4bb1a`, isolated `KAAPPI_HOME`,
cache cleared. macOS aarch64.
**Source location:** `src/primitives_srfi18.zig` (`mutex-state`), reporting
`Mutex.owner` — which holds a `*Fiber` — rather than the owning thread handle.
**Related:** #1984 (`mutex-unlock!` never clears the abandoned flag) is a
different defect in the same procedure's state machine; #1154 (closed) fixed the
owner assignment for an explicit `#f` thread argument.

**Found by:** Systematic audit v2, Phase 5E

### LINKED ISSUE #2194: SRFI-18: thread-join! hangs forever on a spawn'd fiber that has not been dispatched yet
**Category:** SRFI 18 × (kaappi fibers) · **Severity:** hang (unbounded) · **Found by:** working on #2000 ([PR #2193](https://github.com/kaappi/kaappi/pull/2193))

`thread-join!` on a `(kaappi fibers)` fiber that has been `spawn`ed but never dispatched blocks forever. The fiber is perfectly runnable — `fiber-join` on the very same object returns immediately — but `thread-join!` busy-polls a status only the local scheduler can change, and never drives that scheduler.

## Reproduction

Hangs, printing `about-to-join` and nothing further:

```scheme
(import (scheme base) (scheme write) (srfi 18) (kaappi fibers))
(define f (spawn (lambda () 'v)))
(display 'about-to-join) (newline)
(display (thread-join! f)) (newline)
```

No custom port, no SRFI 181, no OS threads involved.

## Discriminating controls

All four run against the same `spawn`'d fiber; only the second fails.

| Probe | Result |
|---|---|
| `(fiber-join f)`, fiber never dispatched | `v` — **works** |
| `(thread-join! f)`, fiber never dispatched | **hangs** (killed at 10s) |
| `(thread-join! f)` after the fiber has started and parked | `v` — works |
| `(thread-join! f 0.2 'timed-out)`, never dispatched | `timed-out` |

So it is specific to `thread-join!` reaching a fiber in `.created`. `fiber-join` proves the fiber can run; a *started* fiber proves the join machinery is fine; and the timeout row shows the second, quieter symptom — with a deadline the program does not hang, but it reports a timeout for a fiber that would have completed instantly had anything dispatched it. That is a wrong answer, not just a slow one.

## Cause

`threadJoinFn`, `src/primitives_srfi18.zig:810-826`:

```zig
// Never-started thread: poll until started+finished or timeout
if (@atomicLoad(fiber_mod.FiberStatus, &target.status, .acquire) == .created) {
    while (status != .completed and status != .errored) {
        if (deadline_ns) |deadline| { ...timeout... }
        sleepNs(1_000_000);
    }
    ...
}
```

The loop calls `sleepNs`, never `fiber_mod.runSchedulerStep`. For a SRFI-18 `make-thread` handle awaiting `thread-start!` from *another* OS thread that is exactly right — the status is changed from outside, so polling observes it. For a `spawn`'d fiber, only this thread's cooperative scheduler can move it out of `.created`, and the poll loop is precisely what prevents that from ever happening. With no timeout the loop is unbounded.

The fiber path further down (line 828, "Fiber path (cooperative scheduling)") does drive the scheduler correctly — it is just unreachable for a target still in `.created`.

Worth noting: this branch is the fix for #878, which reported the opposite defect ("never-started threads return void immediately"). Returning void was wrong, but the polling replacement assumed every never-started target is externally startable.

## Suggested direction

In the `.created` branch, distinguish a local fiber from an OS-thread handle — `target.os_thread == null` is already the discriminator used two lines later — and fall through to the fiber path so the scheduler dispatches it, rather than polling. The `os_thread != null` case should keep polling.

Worth deciding explicitly while there: what a never-started *SRFI-18* `make-thread` handle should do. Blocking forever is arguably correct (such a thread genuinely never terminates), but it is currently indistinguishable from this bug and neither behaviour is documented.

## Notes

- `tests/scheme/audit/primitives_fiber-audit.scm` has a comment at the "a custom-port read! that blocks in thread-join! on a started fiber is rejected" test explaining why it joins a *started* fiber. Update it if this is fixed.
- Needs a regression test per the repo policy — the `fiber-join`-works/`thread-join!`-hangs pair above is the discriminating shape, and any such test needs a bound so a regression fails loudly instead of wedging the runner.

**Verified:** ReleaseSafe build at `bbdce7aa`, isolated `KAAPPI_HOME`, each probe bounded at 10s; the hang reproduced on every run.

baijum:
## Summary

Four SRFI-18 / concurrency audit bugs from the v0.22.2 milestone, one commit each, in the shared `primitives_srfi18.zig` / `fiber_wait.zig` / `fiber.zig` wait-and-thread-lifecycle machinery.

### #2194 — `thread-join!` hangs forever on a spawn'd fiber that has not been dispatched yet (`af743b06`)
`threadJoinFn`'s "never-started" branch busy-polled `fiber.status` without ever driving the scheduler. Correct for a `make-thread` handle awaiting `thread-start!` from outside (#878) — wrong for a `(kaappi fibers)` spawn'd fiber, which only this thread's cooperative scheduler can dispatch: the poll loop *was* the starvation. Discriminate by `sched_idx` (set only by `addFiber`): make-thread handles keep polling; spawn'd fibers fall through to the fiber path, which drives the scheduler. `os_thread` alone is not a safe discriminator — a handle about to be started has `os_thread == null` for the whole window before `thread-start!`'s `std.Thread.spawn`.

### #1982 — `thread-terminate!` cannot interrupt a thread parked in a native SRFI-18 wait (`40df26c8`)
The bytecode safepoint polls `VM.terminate_flag` every 1024 instructions (#933), but only while executing bytecode. `thread-sleep!`, `mutex-lock!` and `mutex-unlock!'s` condvar branch wait inside `runSchedulerStep`'s native loop, so a parked thread never observed the flag: `thread-terminate!` flipped the handle's status, the join's poll exited immediately, and `reapOsThread`'s `thread.join()` blocked forever on a child that would never unwind. `runSchedulerStep` now checks termination (the VM's `terminate_flag` for an OS-thread child — the same flag the safepoint reads — plus the fiber's own flag for a local fiber) at the top of every loop iteration and unwinds `VMError.Terminated`. `SleepWait` gains `pollCapNs` (1ms when another OS thread could terminate this one) so a sleeping thread actually wakes to observe the flag; solo sleeps stay a single reactor block. The wait-context duck-type comptime test moves from 2 to 3 poll caps with the justification.

### #2125 — `mutex-state` reports the owning thread's internal fiber, not the thread (`ce78e10e`)
`Mutex.owner` tracks the *fiber* that acquired the lock (needed by `abandonFiberMutexes`), which for an OS-thread child is the child-heap current fiber — a value not `eq?` to any thread the caller holds, and a dangling pointer once the join frees the child heap (the #2127 quarantine was the detector-side mitigation, not the fix). A new `Mutex.owner_thread` records the thread handle alongside the owner at every write site; `threadEntryFn` stashes the handle on the child VM (`VM.thread_handle`, foreign to the child GC and rooted by the parent), and `mutex-lock!`'s fast/slow paths resolve it exactly like the owner, honouring an explicit SRFI-18 owner argument. `mutex-unlock!` and `abandonFiberMutexes` clear both together. `mutex-state` returns `owner_thread`; the `(eq? (mutex-state m) t)` synchronisation idiom finally works. The existing cross-heap-abandoned-mutex test's workaround ("exclude the two unowned symbols") becomes the real `eq?` probe.

### #2129 — `thread-join!` frees the joined thread's GC/VM while a thread it spawned is still starting (`d1d35353`)
`threadEntryFn`'s prologue dereferences the spawning thread's VM/GC: `GC.initForThread` reads `parent_vm.gc` for the shared symbol tables (`shared_symbols = &parent.symbols`, used by *every* symbol interning for the thread's whole life), and `VM.initForThread` reads the shared maps. `freeChildResources` had no interlock, so joining a thread that had itself spawned a thread freed its GC/VM out from under the grandchild — mid-prologue (the thunk `deepCopy` interns symbols into the freed table) or at its next interning. Reproduced at 18/20 (ReleaseSafe) and 13/15 (gc-stress). `threadStartImpl` now passes the **root VM** (new `VM.root_vm`, resolved in `initForThread` by walking the parent chain) to `threadEntryFn`, so every descendant chains its symbol tables, `foreign_symbols` and shared maps to the root's — which lives for the whole process. A middle thread's own tables stay empty and its GC/VM can be freed at its join. The `thread_handle` from #2125 is recorded only when root-heap.

## Known residual — #2129 stays open

This PR fixes the prologue/symbol-table half of #2129 only. The grandchild's **own handle** — the `*Fiber` `make-thread` allocated on the middle thread, in the middle's heap — is still freed by `freeChildResources` at the middle's join, while the live grandchild reads and writes it for its whole life:

- `child_vm.terminate_flag = &fiber.terminated`, read by the bytecode safepoint every 1024 instructions and by the new `waitTerminated`
- the final `@atomicStore(&fiber.status, .completed, .release)` in `threadEntryFn`

Pre-existing (reproduces identically at `b02ecb25` under Guard Malloc — SIGSEGV in `runUntil`), not introduced here. Silent under the default allocator (the #2129 regression test passes only because the corruption is silent), a live use-after-free under Guard Malloc. A full fix (allocating the handle in the root heap, or gating `freeChildResources` on "no live descendant" the way `hasLiveChildThreads` gates main's teardown) is a larger design change — registry-based deferred frees with deep-nesting cases — and deserves its own PR. Tracked in #2129.

Also documented: grandchildren of a joined thread remain un-joinable (their handle lives in the middle's heap; `checkThreadOwner` refuses foreign joins), so their child resources leak as before.

## Verification

- Crash/hang repros: #2129 prologue 0/25 crashes ReleaseSafe + 0/15 gc-stress (was 14–18/20, 13–15/15); #2194 repro completes; #1982 all four hang shapes terminate promptly; #2125 `eq?` probe is `#t`
- Regression tests (all deadline-bounded so a regression fails loudly instead of wedging the runner):
  - `tests/scheme/srfi/srfi18-join-created-fiber-2194.scm` (incl. the #878 `sched_idx == 0` probe)
  - `tests/scheme/srfi/srfi18-terminate-native-wait-1982.scm`
  - `tests/scheme/srfi/srfi18-mutex-state-owner-2125.scm`
  - `tests/scheme/srfi/srfi18-join-spawn-grandchild-2129.scm`
  - `tests/scheme/srfi/srfi18-cross-heap-abandoned-mutex.scm` updated (workaround → real `eq?` probe)
- Full `run-all.sh`: 2081 pass / 3 fail / 0 timeout (the 3 `compile/` tier failures were a stale dirty-build artifact in my workflow — the interpreter had been built pre-commit with a `-dirty` build id, so its `.sbc` mismatched the bundle's; after rebuilding from the committed tree all three pass)
- `zig build test`, `zig fmt --check`, and the affected suites under `-Dgc-stress=true` all green

## Review comments addressed (`7851bca0`)

- `mutex-lock!` publishes `owner_thread` before `owner`; `mutex-state` gates on `owner_thread` alone (no cross-field race)
- local-fiber termination mid-wait now surfaces "thread terminated" instead of a contentless error
- `waitTerminated` moved out of `runSchedulerStep`'s doc block; orphaned loop-top comment relocated to the check it describes
- `SleepWait.pollCapNs` documents the measured 1ms-poll cost on child threads and the notifier-based follow-up
- #878 `sched_idx` probe test added; #2129 regression test reduced to 12 iterations with the residual documented in its header
- doc drift fixed: `docs/dev/thread-value-sharing.md` and `CLAUDE.md` now say shared tables/maps chain to the root's

Fixes #2194
Fixes #1982
Fixes #2125

### DIFF
--- a/src/fiber_wait.zig
+++ b/src/fiber_wait.zig
@@ -245,6 +245,20 @@ pub fn parkOnReactor(vm: *VM, sched: *FiberScheduler, cap_ns: ?u64) VMError!bool
     return true;
 }
 
+/// thread-terminate! sets the *handle's* `terminated` flag, which an
+/// OS-thread child's VM reaches through `terminate_flag` (the bytecode
+/// safepoint's own check) and a local fiber reaches directly (`me` IS the
+/// handle). A native wait must consult both: it never executes bytecode, so
+/// without this a thread parked in thread-sleep!/a mutex/a condvar wait
+/// never observes termination, never unwinds, and the joining thread hangs
+/// forever in reapOsThread's thread.join() (#1982).
+fn waitTerminated(vm: *VM, me: *Fiber) bool {
+    if (vm.terminate_flag) |flag| {
+        if (@atomicLoad(bool, flag, .monotonic)) return true;
+    }
+    return @atomicLoad(bool, &me.terminated, .monotonic);
+}
+
 /// Drives the scheduler — dispatching other fibers, saving/restoring their
 /// state exactly as switchTo would — until `ctx.isDone()` becomes true or
 /// `me`'s own timed wait expires (`me.timed_out`), parking on the reactor
@@ -288,26 +302,7 @@ pub fn parkOnReactor(vm: *VM, sched: *FiberScheduler, cap_ns: ?u64) VMError!bool
 /// independent progress while a descendant is active regardless, so
 /// excluding them from selection changes no genuine liveness outcome —
 /// only the parked fiber's own loop, right here, ever consumes its wake).
-/// thread-terminate! sets the *handle's* `terminated` flag, which an
-/// OS-thread child's VM reaches through `terminate_flag` (the bytecode
-/// safepoint's own check) and a local fiber reaches directly (`me` IS the
-/// handle). A native wait must consult both: it never executes bytecode, so
-/// without this a thread parked in thread-sleep!/a mutex/a condvar wait
-/// never observes termination, never unwinds, and the joining thread hangs
-/// forever in reapOsThread's thread.join() (#1982).
-fn waitTerminated(vm: *VM, me: *Fiber) bool {
-    if (vm.terminate_flag) |flag| {
-        if (@atomicLoad(bool, flag, .monotonic)) return true;
-    }
-    return @atomicLoad(bool, &me.terminated, .monotonic);
-}
-
 pub fn runSchedulerStep(comptime Ctx: type, ctx: Ctx, vm: *VM, sched: *FiberScheduler, me: *Fiber) VMError!bool {
-    // Checked at the top of every loop iteration: a capped park (pollCapNs,
-    // which the SRFI-18 waits set whenever another OS thread could terminate
-    // this one) re-enters this loop at that cadence, and a local sibling
-    // terminator runs inside this very drive and is observed on the next
-    // iteration.
     // SRFI 181: reaching here from inside a custom port callback means that
     // callback is about to block, which it may not do -- see
     // vm.in_custom_port_callback's doc comment. The guard lives at this
@@ -382,7 +377,15 @@ pub fn runSchedulerStep(comptime Ctx: type, ctx: Ctx, vm: *VM, sched: *FiberSche
     defer _ = sched.driving_waits.pop();
 
     while (!ctx.isDone() and !me.timed_out) {
-        if (waitTerminated(vm, me)) return VMError.Terminated;
+        // Checked at the top of every loop iteration: a capped park
+        // (pollCapNs, which the SRFI-18 waits set whenever another OS thread
+        // could terminate this one) re-enters this loop at that cadence, and
+        // a local sibling terminator runs inside this very drive and is
+        // observed on the next iteration.
+        if (waitTerminated(vm, me)) {
+            vm.setErrorDetail("thread terminated", .{});
+            return VMError.Terminated;
+        }
         const next_idx = sched.scheduleForDispatch() orelse {
             // Nothing dispatchable — but scheduleForDispatch's own per-tick
             // runReactorTick may have just resolved *this* wait, after the
--- a/src/fiber_wait.zig
+++ b/src/fiber_wait.zig
@@ -288,7 +288,26 @@ pub fn parkOnReactor(vm: *VM, sched: *FiberScheduler, cap_ns: ?u64) VMError!bool
 /// independent progress while a descendant is active regardless, so
 /// excluding them from selection changes no genuine liveness outcome —
 /// only the parked fiber's own loop, right here, ever consumes its wake).
+/// thread-terminate! sets the *handle's* `terminated` flag, which an
+/// OS-thread child's VM reaches through `terminate_flag` (the bytecode
+/// safepoint's own check) and a local fiber reaches directly (`me` IS the
+/// handle). A native wait must consult both: it never executes bytecode, so
+/// without this a thread parked in thread-sleep!/a mutex/a condvar wait
+/// never observes termination, never unwinds, and the joining thread hangs
+/// forever in reapOsThread's thread.join() (#1982).
+fn waitTerminated(vm: *VM, me: *Fiber) bool {
+    if (vm.terminate_flag) |flag| {
+        if (@atomicLoad(bool, flag, .monotonic)) return true;
+    }
+    return @atomicLoad(bool, &me.terminated, .monotonic);
+}
+
 pub fn runSchedulerStep(comptime Ctx: type, ctx: Ctx, vm: *VM, sched: *FiberScheduler, me: *Fiber) VMError!bool {
+    // Checked at the top of every loop iteration: a capped park (pollCapNs,
+    // which the SRFI-18 waits set whenever another OS thread could terminate
+    // this one) re-enters this loop at that cadence, and a local sibling
+    // terminator runs inside this very drive and is observed on the next
+    // iteration.
     // SRFI 181: reaching here from inside a custom port callback means that
     // callback is about to block, which it may not do -- see
     // vm.in_custom_port_callback's doc comment. The guard lives at this
@@ -363,6 +382,7 @@ pub fn runSchedulerStep(comptime Ctx: type, ctx: Ctx, vm: *VM, sched: *FiberSche
     defer _ = sched.driving_waits.pop();
 
     while (!ctx.isDone() and !me.timed_out) {
+        if (waitTerminated(vm, me)) return VMError.Terminated;
         const next_idx = sched.scheduleForDispatch() orelse {
             // Nothing dispatchable — but scheduleForDispatch's own per-tick
             // runReactorTick may have just resolved *this* wait, after the
--- a/src/primitives_srfi18.zig
+++ b/src/primitives_srfi18.zig
@@ -807,8 +807,27 @@ fn threadJoinFn(args: []const Value) PrimitiveError!Value {
         return reapOsThread(target, args[0]);
     }
 
-    // Never-started thread: poll until started+finished or timeout
-    if (@atomicLoad(fiber_mod.FiberStatus, &target.status, .acquire) == .created) {
+    // Never-started target, in two kinds discriminated by sched_idx (set only
+    // by addFiber; a make-thread object is never added to any scheduler and
+    // leaves it at 0):
+    //
+    //  * A make-thread handle awaiting thread-start! (sched_idx == 0): poll.
+    //    The status is changed from outside this loop, so polling observes it
+    //    (#878). os_thread alone is NOT a safe discriminator here -- a handle
+    //    about to be started has os_thread == null for the whole window before
+    //    thread-start!'s std.Thread.spawn, and must keep polling, not fall
+    //    through to the cooperative path.
+    //
+    //  * A (kaappi fibers) spawn'd fiber (sched_idx != 0) still in .created:
+    //    only THIS thread's cooperative scheduler can dispatch it, and the
+    //    poll loop below is exactly what starves it -- the status can never
+    //    change from outside, so without a timeout the loop is unbounded and
+    //    thread-join! hangs on a fiber that would have completed instantly
+    //    (#2194). Fall through to the fiber path below, which drives the
+    //    scheduler.
+    if (target.sched_idx == 0 and
+        @atomicLoad(fiber_mod.FiberStatus, &target.status, .acquire) == .created)
+    {
         while (@atomicLoad(fiber_mod.FiberStatus, &target.status, .acquire) != .completed and
             @atomicLoad(fiber_mod.FiberStatus, &target.status, .acquire) != .errored)
         {
--- a/src/primitives_srfi18.zig
+++ b/src/primitives_srfi18.zig
@@ -450,8 +450,15 @@ fn threadStartImpl(args: []const Value) PrimitiveError!Value {
     // it one below the true count with other children alive), letting
     // crossThreadWaitPossible wrongly conclude no other thread exists.
     _ = @atomicRmw(usize, &live_child_threads, .Add, 1, .release);
+    // Pass the ROOT VM, not this thread's own: threadEntryFn's prologue
+    // dereferences it (GC.initForThread's shared symbol tables, and the
+    // shared maps), and a middle thread's VM/GC are freed when *it* is
+    // joined -- a grandchild still starting (or interning symbols) would
+    // then read freed memory (#2129). The root VM lives for the whole
+    // process, so every descendant chains its shared state to it.
+    const root_vm = vm.root_vm orelse vm;
     fiber.os_thread = std.Thread.spawn(.{}, threadEntryFn, .{
-        fiber, gc.allocator, vm, envelope,
+        fiber, gc.allocator, root_vm, envelope,
     }) catch {
         _ = @atomicRmw(usize, &live_child_threads, .Sub, 1, .release);
         envelope.deinit();
@@ -461,7 +468,11 @@ fn threadStartImpl(args: []const Value) PrimitiveError!Value {
     return args[0];
 }
 
-fn threadEntryFn(fiber: *fiber_mod.Fiber, allocator: std.mem.Allocator, parent_vm: *vm_mod.VM, envelope: *shared_channel.Envelope) void {
+fn threadEntryFn(fiber: *fiber_mod.Fiber, allocator: std.mem.Allocator, root_vm: *vm_mod.VM, envelope: *shared_channel.Envelope) void {
+    // `root_vm` is the ROOT VM (see threadStartImpl): its struct, GC and
+    // shared maps are never freed, so the whole prologue below -- and this
+    // thread's later symbol interning into GC.initForThread's shared tables
+    // -- cannot race a join of whatever thread spawned this one (#2129).
     // Balances the increment in threadStartFn on every exit path (including
     // the early GC/VM-init failures below), so crossThreadWaitPossible's "is
     // another OS thread still alive" check stays accurate.
@@ -475,7 +486,7 @@ fn threadEntryFn(fiber: *fiber_mod.Fiber, allocator: std.mem.Allocator, parent_v
         @atomicStore(fiber_mod.FiberStatus, &fiber.status, .errored, .release);
         return;
     };
-    child_gc.* = memory.GC.initForThread(allocator, parent_vm.gc);
+    child_gc.* = memory.GC.initForThread(allocator, root_vm.gc);
 
     const child_vm = allocator.create(vm_mod.VM) catch {
         child_gc.deinit();
@@ -484,7 +495,7 @@ fn threadEntryFn(fiber: *fiber_mod.Fiber, allocator: std.mem.Allocator, parent_v
         return;
     };
     @memset(std.mem.asBytes(child_vm), 0);
-    child_vm.* = vm_mod.VM.initForThread(child_gc, parent_vm) catch {
+    child_vm.* = vm_mod.VM.initForThread(child_gc, root_vm) catch {
         allocator.destroy(child_vm);
         child_gc.deinit();
         allocator.destroy(child_gc);
@@ -502,10 +513,18 @@ fn threadEntryFn(fiber: *fiber_mod.Fiber, allocator: std.mem.Allocator, parent_v
     // The thread handle (the parent-heap fiber make-thread returned) this
     // child runs for, so mutex-state on a mutex this thread locks reports
     // the thread the caller holds -- not the child's internal current
-    // fiber, which lives in this child heap (#2125). Foreign to this GC;
-    // the parent roots the handle while the child runs, and isYoungPointer
-    // returns false for it, so no write barrier is needed.
-    child_vm.thread_handle = types.makePointer(&fiber.header);
+    // fiber, which lives in this child heap (#2125). Recorded only when the
+    // handle is ROOT-heap (this thread's creator was the root thread): a
+    // middle thread's handle lives in the middle's heap and is freed when
+    // the middle is joined, while this child -- and any mutex-state query on
+    // a mutex it locked -- can outlive that join, so a recorded middle-heap
+    // handle would dangle (#2129's grandchild shape). Such a child falls
+    // back to its own (never-freed, un-joinable) current fiber, the
+    // pre-#2125 behaviour. Foreign to this GC either way; the owner roots
+    // the handle, and isYoungPointer returns false for it, so no write
+    // barrier is needed.
+    if (fiber.header.owner == root_vm.gc.id)
+        child_vm.thread_handle = types.makePointer(&fiber.header);
 
     child_registry.put(@intFromPtr(fiber), .{ .child_gc = child_gc, .child_vm = child_vm }) catch {
         fiber.result = types.VOID;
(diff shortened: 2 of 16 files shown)


baijum (commented):
Built the branch, ran `zig build test` (green), `zig fmt --check` (clean), and all 34 SRFI-18 / fiber / audit Scheme suites (all pass), plus targeted probes under Guard Malloc (`DYLD_INSERT_LIBRARIES=/usr/lib/libgmalloc.dylib`).

Four fixes, each well-diagnosed, and the commit messages are unusually good. Four inline findings below; one is substantive (#2129 is only half fixed). Plus one that has no diff line to hang on:

### Doc drift (low)

`docs/dev/thread-value-sharing.md` is the canonical model per `CLAUDE.md` and still describes the pre-#2129 chaining:

- line 253 — "`GC.initForThread` ... Per-thread GC, sharing **the parent's** symbol table"
- line 255 — "`VM.initForThread` ... sharing **the parent's** globals and libraries by pointer"
- lines 100-102 (the "globals route" prose) — same framing

After this PR it is the **root's**, for every descendant — which is precisely the fact #2129 turned on, so this is the one doc where the distinction is load-bearing. The same sentence in `kaappi/CLAUDE.md`'s Concurrency section is stale too. `mutex-state`'s observable return value also changed; the user-facing page lives in the other repo, but worth a note there.

### What's good

The `sched_idx` discriminator is sound — `make-thread` never calls `addFiber`, and `retireSlot`/`popFreeSlot` never hand out slot 0, so `sched_idx == 0` really does partition the two kinds. `owner`/`owner_thread` stay consistent at every write site including the `#f`-owner case, so `mutex-state` can never return a bare VOID for a locked mutex. The GC side is complete (`markObjectContents`, `markValueInner`, `referencesYoung`, plus the field-inventory guard). The comptime poll-cap count test was updated with its justification rather than just bumped. And every new test is deadline-bounded with real negative controls.

tests/scheme/srfi/srfi18-join-spawn-grandchild-2129.scm
```diff
@@ -0,0 +1,67 @@
...
+;; (srfi 120) unusable from a thread.
+;;
+;; threadEntryFn's prologue dereferences the spawning thread's VM/GC
+;; (GC.initForThread's shared symbol tables, then the shared maps), and every
+;; later symbol interning goes through those tables for the thread's whole
+;; life. freeChildResources had no interlock: joining a thread that had
+;; itself spawned a thread freed its GC/VM out from under the grandchild,
+;; which was still (or forever after) reading freed memory. The fix chains
+;; every thread's shared state to the ROOT VM/GC -- which lives for the whole
+;; process -- so a join can never free anything a descendant references.
```
baijum, line 12: **#2129 is only half fixed — this claim does not hold (high)**

Moving the *shared* state to the root VM does close the prologue crash. But the grandchild's **own handle** — the `*Fiber` that `make-thread` allocated *on the middle thread*, so in the middle's heap — is still freed by `freeChildResources` → `res.child_gc.deinit()` at the middle's join. The live grandchild reads and writes it for its whole life:

- `child_vm.terminate_flag = &fiber.terminated`, read by the bytecode safepoint every 1024 instructions (`src/vm_dispatch.zig:100`) **and now also by the new `waitTerminated`** (`src/fiber_wait.zig:300`)
- `@atomicStore(&fiber.status, .completed, .release)` at the end of `threadEntryFn`

Repro — grandchild outlives the middle's join, no native wait needed:

```scheme
(define (spin n) (let loop ((i 0) (a 0)) (if (>= i n) a (loop (+ i 1) (+ a i)))))
            (let ((g (make-thread (lambda () (spin 30000000) 'g)))) (thread-start! g) 'plain))))
(thread-start! t) (display (thread-join! t)) (newline) (thread-sleep! 5.0)
```

```
DYLD_INSERT_LIBRARIES=/usr/lib/libgmalloc.dylib ./zig-out/bin/kaappi /tmp/gc2129b.scm
```

`Segmentation fault ... in runUntil (vm_dispatch.zig:100)` — **identical at `b02ecb25` and at `d1d35353`**. It also reproduces through the new `waitTerminated` when the grandchild sleeps instead of spinning, so the #1982 commit adds a second, hotter dereference of the same freed object.

So this is pre-existing, not introduced — but it is the same issue. `run-shape` passes 30/30 only because the corruption is silent under `c_allocator`; the shape is not actually safe.

Suggested: either extend the fix (allocate the handle in the root heap, or gate `freeChildResources` on "no live descendant", the way `hasLiveChildThreads` already gates main's teardown), or — if that's out of scope here — keep #2129 open, move this to the Known-residual section, and soften the claim in this comment. The residual section currently mentions only the leak and un-joinability, not a live use-after-free.

src/fiber_wait.zig
```diff
@@ -288,7 +288,26 @@ pub fn parkOnReactor(vm: *VM, sched: *FiberScheduler, cap_ns: ?u64) VMError!bool
...
 /// excluding them from selection changes no genuine liveness outcome —
 /// only the parked fiber's own loop, right here, ever consumes its wake).
+/// thread-terminate! sets the *handle's* `terminated` flag, which an
+/// OS-thread child's VM reaches through `terminate_flag` (the bytecode
+/// safepoint's own check) and a local fiber reaches directly (`me` IS the
+/// handle). A native wait must consult both: it never executes bytecode, so
+/// without this a thread parked in thread-sleep!/a mutex/a condvar wait
+/// never observes termination, never unwinds, and the joining thread hangs
+/// forever in reapOsThread's thread.join() (#1982).
+fn waitTerminated(vm: *VM, me: *Fiber) bool {
```
baijum: **`runSchedulerStep` lost its doc comment (medium)**

`waitTerminated` was inserted *between* the existing ~40-line `///` block and the function it documents. Zig binds a doc comment to the next declaration, so that whole block — the `Ctx` duck type, `pollCapNs`, the entire `me.driving` / #1487 rationale — now documents this five-line boolean helper, and `runSchedulerStep`, the single shared body behind every wait, has no doc comment at all. It shows up that way in generated docs and in editor hover.

Fix: move `waitTerminated` above the doc block (or below `runSchedulerStep`), and keep just its own three or four lines as its doc.

src/fiber_wait.zig
```diff
@@ -288,7 +288,26 @@ pub fn parkOnReactor(vm: *VM, sched: *FiberScheduler, cap_ns: ?u64) VMError!bool
...
+    }
+    return @atomicLoad(bool, &me.terminated, .monotonic);
+}
+
 pub fn runSchedulerStep(comptime Ctx: type, ctx: Ctx, vm: *VM, sched: *FiberScheduler, me: *Fiber) VMError!bool {
+    // Checked at the top of every loop iteration: a capped park (pollCapNs,
+    // which the SRFI-18 waits set whenever another OS thread could terminate
+    // this one) re-enters this loop at that cadence, and a local sibling
+    // terminator runs inside this very drive and is observed on the next
+    // iteration.
```
baijum: This comment is orphaned from what it explains: the check it describes is ~75 lines below, inside the `while` loop, and the very next line here starts an unrelated SRFI-181 comment. Reads as if the SRFI-181 guard is the termination check. Better placed immediately above `if (waitTerminated(vm, me)) ...`.

src/primitives_srfi18.zig
```diff
@@ -594,6 +621,16 @@ pub const SleepWait = struct {
...
+    // See MutexWait.pollCapNs: thread-terminate! from another OS thread is
+    // observed only by polling -- nothing wakes a sleep park on termination
+    // -- so a sleep in a program with other live OS threads must not block
+    // for its whole duration on the offchance the terminator is one of
+    // them (#1982); the runSchedulerStep loop re-checks the termination
+    // flag at this cadence. Solo (no other OS thread can exist to
+    // terminate this one) it stays a single true reactor block.
+    pub fn pollCapNs(_: SleepWait) ?u64 {
+        return if (crossThreadWaitPossible()) CROSS_THREAD_POLL_NS else null;
+    }
```
baijum, line 642: **Every child-thread sleep becomes a 1 ms poll (low-medium)**

`crossThreadWaitPossible()` is unconditionally true on any child thread (`!vm.owns_globals` → `return true`), so a child's `(thread-sleep! 60)` now wakes 60,000 times. Measured on a two-thread shape (2.5 s + 3.0 s sleeps), main vs. this branch:

| | involuntary ctx switches | user+sys |
|---|---|---|
| `b02ecb25` | 33 | 0.00 s |
| this PR | 4953 | 0.04 s |

CPU is still small, but it scales linearly with sleep duration and thread count, and the PR description doesn't quantify the tradeoff.

Worth a follow-up: `ThreadNotifier.notify()` is already documented thread-safe and has the refcounted `alive`/`releaseNotifier` lifetime built for exactly a possibly-dying reactor, and `thread-terminate!` can reach the victim's reactor through `child_registry`'s `child_vm`. Notifying instead of polling would make termination *immediate* and remove the wakeups entirely. Not blocking.

src/fiber_wait.zig
```diff
@@ -363,6 +382,7 @@ pub fn runSchedulerStep(comptime Ctx: type, ctx: Ctx, vm: *VM, sched: *FiberSche
     defer _ = sched.driving_waits.pop();
 
     while (!ctx.isDone() and !me.timed_out) {
+        if (waitTerminated(vm, me)) return VMError.Terminated;
```
baijum: **New reachable path yields a contentless error (low)**

A local sibling terminating the fiber currently inside `runSchedulerStep` is newly observable here — the safepoint never caught it, because `vm.terminate_flag` is null on the main VM and the safepoint doesn't consult `me.terminated`.

```scheme
(import (scheme base) (scheme write) (srfi 18) (kaappi fibers))
(define main-t (current-thread))
(define ch (make-channel))
(define blocker (spawn (lambda () (channel-receive ch))))
(define f (spawn (lambda () (thread-terminate! main-t) 'b-done)))
(display (guard (e (#t (list 'caught e))) (fiber-join blocker)))
```

- `b02ecb25`: `(caught #<error fiber-join: deadlock — joined fiber can never complete (all fibers blocked)>)`
- this PR: `error[KP9000]: error` — uncatchable (correctly, `Terminated` is uncatchable) but with no message at all

The unwind is more correct than the old hang-then-deadlock-error. The diagnostic is the regression: compare the OS-thread path, which surfaces a proper `terminated-thread-exception`. A local fiber reaching the top level this way deserves the same.

tests/scheme/srfi/srfi18-join-created-fiber-2194.scm
```diff
@@ -0,0 +1,56 @@
...
+(test-equal "thread-join! on an uncompletable never-dispatched fiber times out"
+  'timed-out
+  (thread-join! (spawn (lambda () (channel-receive (make-channel)))) 0.5 'timed-out))
+
+;; The make-thread side of the sched_idx discriminator must keep working: a
+;; handle whose OS thread has completed still joins normally.
+(test-equal "thread-join! on a completed OS thread still joins"
+  7 (let ((t (make-thread (lambda () 7))))
+      (thread-start! t)
+      (thread-join! t)))
```
baijum, line 70: Minor / test coverage: this pins the *completed* OS-thread case, but nothing here pins the branch the discriminator actually exists to preserve — a `make-thread` handle joined **before** something else starts it (the #878 shape, `sched_idx == 0` + `.created` → keep polling). The existing suites cover it incidentally, but an explicit probe (join with a deadline while another fiber does the `thread-start!`) would make both halves of the `sched_idx` split regression-proof, not just the new one.

baijum:
Thanks both for the careful review. `7851bca0` addresses everything actionable; the one substantive finding (#2129 half-fix) is documented as a known residual and stays tracked in #2129.

**baijum** — thanks for the gmalloc verification; you are right that the handle lifetime is the same issue, and pre-existing. I chose the documented-residual path you offered rather than the registry-gated deferred-free design, which deserves its own PR (deep-nesting record lookup, join-semantics questions). The PR body now states the residual as a *live* use-after-free (not just the leak), the #2129 test header says it pins only the symbol-table half, and `Fixes #2129` is removed. Also fixed: the `waitTerminated` doc-block theft, the orphaned loop-top comment, the contentless-error diagnostic (now "thread terminated"), the #878 probe test, the doc drift in `thread-value-sharing.md`/`CLAUDE.md`, and the SleepWait polling tradeoff is quantified in the code comment (notifier-based interrupt noted as the follow-up).

**CodeRabbit** — both lock paths now publish `owner_thread` before `owner` (computed up front), and `mutex-state` gates on `owner_thread` alone, which removes the cross-field race entirely rather than just narrowing it; the #2129 test runs 12 iterations instead of 30 with the leak documented.

PR #2232, opened 2026-08-05
Fix bundled-binary argv, stale-bundle failures, and -dirty build id (#2010, #1930, #2097)

### LINKED ISSUE #1930: Two compile tests fail spuriously after any commit: bundled .sbc is built by the old binary, bundler by the new one
**Category:** Test harness · **Severity:** flaky failure with a misleading diagnostic · **Found by:** systematic audit v2 — hit independently by **two of six** units (Phases 1B and 3.1) (#1890)

`tests/scheme/compile/compile-import-error-703.sh` and `compile-preamble-gc-700.sh` fail with:

```text
fatal: invalid embedded bytecode (wrong version or format)
```

That message reads like a serialisation bug. It is not — it is a staleness mismatch, and it is triggered by the ordinary edit-commit-test workflow.

## Mechanism

Both scripts are the two that use `-Dbundle=<pre-compiled .sbc>`. They:

1. produce the `.sbc` using the **pre-built** `zig-out/bin/kaappi`, then
2. rebuild the bundler from **current source** (`zig build -Dbundle=… --prefix …`).

The `.sbc` cache key folds in the git build id — the commit hash plus a bare `-dirty` flag. So any build-id change *between those two steps* makes the bundled binary reject its own payload. A new commit does it. So does a clean↔dirty flip. Reproduced down to a hello-world `.sbc`.

## Why it keeps catching people

The audit workflow is: build → write tests → **commit** → run the suite. The commit flips the build id between the binary on disk and the rebuild the script performs, so the failure appears at the *end* of a unit's work, attributed to whatever that unit changed. Both units that hit it initially suspected their own diff.

Both scripts pass after a plain `zig build` at the current commit, and pass individually. Neither failure was real.

## Suggested fixes, cheapest first

1. **Make the diagnostic honest.** When the embedded bytecode's build id differs from the running binary's, say so — *"bundled bytecode was produced by build id X, this binary is Y; rebuild"* — instead of "wrong version or format". This alone converts a 20-minute investigation into a one-line read, and helps outside the test harness too.
2. **Have the scripts rebuild the interpreter first**, or generate the `.sbc` with the same binary they then bundle with, so the two steps cannot disagree.
3. Optionally have `run-all.sh` fail fast with a clear message when `zig-out/bin/kaappi`'s build id does not match the working tree.

## Related

#1926 covers these same two scripts from the runtime angle (two full `zig build -Dbundle` rebuilds dominating the shell-suite time). Fixing (2) here may interact with that, so worth doing together.

The general class — the `-dirty` flag being a bare flag rather than a hash of the changes, so two different uncommitted edits at one commit alias to the same id — is documented in `docs/dev/cache.md` and is why the audit protocol's step 3 requires `kaappi cache clear` after every rebuild. That guard does not help here, because the mismatch is inside a bundled binary rather than in the cache.

**Verified:** reproduced independently by two units; in both cases the scripts failed before a rebuild and passed after, with no `.zig` file changed in between.

### LINKED ISSUE #2010: A -Dbundle standalone binary silently swallows arguments that collide with kaappi subcommand names
## Summary

A standalone binary built with `zig build -Dbundle=prog.sbc` silently **consumes
some of its own arguments** before the bundled program sees them. The bundled
program's `(command-line)` is missing the first argument whenever that argument
happens to name one of kaappi's subcommands.

Some subcommands run *instead of* the bundled program (`explain`, `doctor`,
`test`). Others are swallowed without running anything (`check`, `fmt`, `ast`,
`compile`) — the bundled program starts, but with an argument list that has
quietly lost its first element.

The effect is that a bundled program cannot define a CLI of its own: any
subcommand name it picks that collides with kaappi's is unreachable, and the
collision is invisible — no error, just a shorter argument list.

## Repro

```scheme
;; av.scm
(import (scheme base) (scheme write) (scheme process-context))
(write (command-line))
(newline)
```

```
kaappi --compile av.scm -o av.sbc
zig build -Dbundle=$PWD/av.sbc
cp zig-out/bin/kaappi ./avbin
```

```
./avbin hello z.scm     => ("hello" "z.scm")     correct
./avbin a b c           => ("a" "b" "c")         correct
./avbin eval z.scm      => ("eval" "z.scm")      correct
./avbin repl z.scm      => ("repl" "z.scm")      correct

./avbin check z.scm     => ("z.scm")             `check` swallowed
./avbin fmt z.scm       => ("z.scm")             `fmt` swallowed
./avbin ast z.scm       => ("z.scm")             `ast` swallowed
./avbin compile z.scm   => ("z.scm")             `compile` swallowed

./avbin explain z.scm   => kaappi explain: unknown diagnostic code 'z.scm'
./avbin doctor z.scm    => kaappi doctor: unexpected argument 'z.scm'
./avbin test z.scm      => kaappi test: seed 733812 (…)
```

The last three never reach the bundled program at all.

## Expected

With bytecode bundled, the binary *is* the bundled program. Its argument vector
belongs to that program, so `(command-line)` should be the full list in every
case above — the same `("check" "z.scm")` shape that `hello` and `eval`
already produce.

The inconsistency is the clearest signal that this is unintended: `eval` and
`repl` are kaappi subcommands too and pass through fine, while `check`, `fmt`,
`ast` and `compile` do not.

## How it was found

[paal](https://github.com/kaappi/paal) bundles itself this way and has its own
`check` and `fmt` subcommands. In a bundled paal, `paal check f.scm` *ran* the
program instead of compiling it: `check` was consumed, so paal's dispatcher saw
only `("f.scm")` and treated it as a positional file. The same commands are
correct when paal runs on kaappi as a script, which is what made it look like a
paal bug at first.

The repro above has no paal in it — it is a four-line program.

## Versions

- kaappi `main` @ 321da93a (v0.22.1), macOS arm64, zig 0.16.0.

### LINKED ISSUE #2097: [Tooling] An untracked file marks the build id -dirty, breaking bundle tests with 'invalid embedded bytecode'
**Category:** Tooling / build
**Severity:** edge-case (developer-facing; presents as a spurious test failure with a misleading message)

**Adding a new, untracked file makes the build id `-dirty`**, which silently invalidates any `zig-out/bin/kaappi` built moments earlier — and `tests/scheme/compile/`'s bundle tests then fail with `fatal: invalid embedded bytecode`, naming nothing that would lead you to the cause.

`gitBuildId` (`build.zig:587`) decides dirtiness with:

```zig
const status = b.runAllowFail(&.{ "git", "-C", cwd, "status", "--porcelain" }, &code, .ignore) catch "";
const dirty = std.mem.trim(u8, status, " \t\r\n").len != 0;
```

`git status --porcelain` reports untracked files by default. Demonstrated on a throwaway repo:

```console
clean tree      : porcelain=[]                 -> dirty=no
$ echo b > newfile.scm
after untracked : porcelain=[?? newfile.scm]   -> dirty=yes
with -uno       : porcelain=[]                 -> dirty=no
```

**Why it produces a confusing failure.** The bundle tests compile a fixture `.sbc` using whatever `zig-out/bin/kaappi` already exists, then embed it in a freshly built binary. Create a test file, run `zig build`, and the two now disagree: the old binary stamped `<sha>`, the new one expects `<sha>-dirty`. The `.sbc` is rejected as foreign. Nothing in the message mentions the build id, git, or the file you just created.

**Discriminating control:** the same file, `git add`-ed but not committed, produces the identical `-dirty` id — so the flag is not wrong in general, only in counting *untracked* files as a change to the build. And `--porcelain -uno` reports clean with the untracked file present.

**Cost.** Five separate audit units hit this today (Phases 3.4, 3.7, 3.8, 3.9, 6A) and each spent time deciding whether they had caused a real regression in `tests/scheme/compile/`. Every one concluded "my own artifact" independently. Three of them wrote it up in their PR body as a footgun to warn the next person, which is the tell that the tooling should just not do this.

**Suggested fix.** `git status --porcelain -uno` (or `--untracked-files=no`). An untracked file is by definition not part of the build's source of truth: it is not in the commit and not staged, so it cannot change what a *tracked-source* build produces. Committed-but-modified and staged files should keep marking the tree dirty, and `-uno` preserves both.

If untracked files are deliberately counted (e.g. a new `.zig` file *would* change the build once referenced), then the fix is instead to make the failure legible — the bundle mismatch should say the two ids and where they came from. But note the current behaviour is already inconsistent with that rationale, since a new untracked `.scm` test cannot affect the compiled binary at all.

**Verified:** the git behaviour on a throwaway repo (transcript above); `build.zig:576-590` read directly. macOS aarch64, git 2.x.

**Found by:** Systematic audit v2 — reported independently by Phases 3.4, 3.7, 3.8, 3.9 and 6A; diagnosed precisely by 3.7.

baijum:
Closes #2010, closes #1930, closes #2097.

Three issues in the `-Dbundle` standalone path and the build-id machinery
that feeds it, fixed together because they are the same class.

## #2010 — a bundled binary swallows subcommand-named arguments

With bytecode bundled the binary *is* the bundled program, so its whole
argv belongs to that program's `(command-line)`. `cli.parse` silently
dropped subcommand words (`check`, `fmt`, `ast`, `compile`) from the list,
and the pre-VM dispatchers ran `explain`/`doctor`/`test` *instead of* the
bundled program. Bundled mode now bypasses kaappi's argument parsing
entirely: `cli.parseBundled` hands every argv element after `argv[0]` to
`(command-line)` verbatim, the pre-VM subcommands and the `--sandbox`
pre-scan are skipped, and `--help`/`--version`/flags all belong to the
program. New regression test `tests/scheme/compile/bundle-args-2010.sh`
reuses the shared bundle fixture (a cache hit, not a third full rebuild).

## #1930 — spurious `invalid embedded bytecode` failures + misleading diagnostic

The `.sbc`'s compiler hash folds in the producing binary's git build id, so
a tree that moved (a new commit, or clean↔dirty) between producing a `.sbc`
and building the bundler made the binary reject its own payload as foreign —
a message that read like a serialisation bug. Two fixes:

- **Honest diagnostic.** When the embedded `.sbc`'s header parses but names
  a different build, the fatal message now says both build ids and "rebuild
  the bundle from current source" (`classifyEmbeddedRejection` in
  `bytecode_file.zig` + unit test; `main.zig`).
- **The harness can no longer trip on the mismatch.** `bundle_fixture_binary`
  (and the inline `-Dbundle` section of
  `compile-toplevel-side-effects-2156.sh`, which had the same latent hazard)
  now build the interpreter into an isolated prefix from the same source as
  the bundler and produce the `.sbc` with that binary. The caller's
  `zig-out/bin/kaappi` is never touched (kaappi#2163), and the `-Dbundle`
  rebuild that follows is a cache hit (kaappi#1926) — verified: the second
  and third bundle scripts run in <1s after the first.

I deliberately skipped the issue's optional suggestion #3 (fail-fast in
`run-all.sh` on a stale binary): with the scripts fixed there is no spurious
failure left to guard against, and a hard fail would false-positive on
legitimate release-binary/dirty-tree workflows.

## #2097 — an untracked file marks the build id `-dirty`

`gitBuildId` counted `git status --porcelain`'s untracked files, so a
brand-new file (e.g. a test being written) silently flipped every later
build id to `-dirty`, invalidating an existing `zig-out/bin/kaappi` and
tripping the bundle tests. Now `--porcelain -uno`: an untracked file is not
part of a tracked-source build's output. Committed-but-modified and staged
files still mark the tree dirty.

## Verification

- `zig build test` (incl. `-Dgc-stress=true`) — green, incl. the two new
  unit tests (`classifyEmbeddedRejection`, `parseBundled`)
- All 26 `tests/scheme/compile/*.sh` — green (23 existing + new
  `bundle-args-2010.sh`); the bundle scripts pass even when handed a
  deliberately stale binary (build id from a different commit)
- End-to-end repro of the old failure: a `.sbc` from a `65bc8ff8` build
  embedded in a `006b263e` bundler now prints
  `fatal: embedded bytecode was produced by a different build (build id
  65bc8ff8-dirty); this binary is 006b263e-dirty. Rebuild the bundle from
  current source.` (exit 1)
- `markdownlint-cli2` clean; `zig fmt --check` clean


baijum (commented):
Reviewed the three fixes; the logic is sound and I verified it locally.

**Verification performed**
- New unit tests pass — `zig build test` filtered to `parseBundled` + `classifyEmbeddedRejection`: 8/8, 0 fail.
- End-to-end `bundle-args-2010.sh` passes on a freshly built standalone binary: all 13 subcommand-named words plus `--gc-stats` reach `(command-line)`.

**Correctness notes**
- **#2010** is fixed on *both* interception paths — the pre-VM dispatchers (`explain`/`doctor`/`test`/`cache`/`features`) via the `embedded_bytecode.bytecode == null` comptime guard, and the in-parse subcommands (`check`/`fmt`/`ast`/`compile`/`expand`/`ir`) via the `parse` → `parseBundled` swap. The standalone block `return`s before the lib-path fold and subcommand dispatch, so `parseBundled` leaving `file_path`/`lib_paths` at default is safe, and `vm.lib_paths` was never set in the bundled path anyway — no library-resolution regression.
- **#1930**: the classifier is reachable because `deserializeFromBuffer` returns `null` (not an error) on a compiler-hash mismatch, so the `orelse` branch runs. Producing the `.sbc` with an interpreter built from the same tree as the bundler is the right harness fix; a cross-*version* `.sbc` correctly falls back to the generic message. Checked the other `.sbc`-producing scripts (`compile-preamble-699.sh`, `native-external-library-import-1743.sh`) — neither runs `-Dbundle=<precompiled.sbc>`, so they don't share the hazard.
- **#2097**: `--porcelain -uno` still flags staged and modified/deleted *tracked* files as dirty; only untracked files stop counting. Correct and minimal.

**One suggestion (non-blocking)**

`parseBundled` disables *all* kaappi runtime flags on a bundled binary, so the `--gc-stats`/`--profile`/`--coverage` handling in the standalone block (`src/main.zig`, the `defer if (opts.gc_stats_mode …)` group) is now unreachable — that block is only entered when bundled, and there `opts` always comes from `parseBundled`, which never sets those fields. Consider either deleting the now-dead handling or commenting that these diagnostics are deliberately no longer honored in bundled mode. It's also worth a CHANGELOG clause noting `--profile`/`--coverage`/`--gc-stats` no longer apply to a bundled binary — the entry currently frames `--gc-stats` only as correct passthrough, so the capability removal is otherwise invisible.

**Nit**: the PR description says "All 26 … (23 existing + new)" — 23 + 1 ≠ 26; the directory has 26 files, so it should read "25 existing + 1 new." Description only.

baijum:
Addressed both CodeRabbit findings (cc5c6e7e, a35f8580):

1. **`bundle-args-2010.sh` (Major)** — `KAAPPI` was genuinely unused (SC2034). Added the interpreter-as-oracle no-argument baseline via `interp_stdout` + `assert_tiers_agree`: with no args the bundled `(command-line)` is `()` and the interpreter's is `("main.scm")`, so the fixture's conditional `cmdline` print is silent on both tiers and the two must agree exactly. The per-argument golden assertions stay golden on purpose — bundled `(command-line)` intentionally differs from direct source execution (no script path). Note `a35f8580`: the first attempt exposed a real bug in the baseline itself (relative `zig-out/bin/kaappi` broke once `interp_stdout` cd'd into the fixture — exit 127), which is exactly what an oracle comparison is for; fixed by resolving an absolute path up front.

2. **`gitBuildId` dirty-state regression (Minor)** — the #2097 contract had no test. New `tests/scheme/compile/build-id-untracked-2097.sh` exercises `gitBuildId` end-to-end through `kaappi features --json` on three isolated-prefix builds: pristine → `<hash>`, +untracked file → unchanged (the regression — fails pre-fix), +staged → `<hash>-dirty`. Skips (77) when the working tree isn't pristine, since the contract is unobservable then. Warm-cache cost 1.3s; cold ~83s. It lives in `compile/` rather than the bytecode_file unit tests because `gitBuildId` is configure-time code in `build.zig`, outside the Zig test module.

Issue #1929, opened 2026-07-31
Reader.readIntegerWithRadix is dead code: its delimiter guard has never executed since it was written


baijum:
**Category:** Reader · **Severity:** wrong-result · **Found by:** systematic audit v2, Phase 1B (#1890)

#1892 reported that radix-prefixed integers skip the trailing-delimiter check. Phase 1B swept the whole space — **19 prefix spellings × 26 trailing characters = 494 cells** — and found the cause is a single bypassed call, wider than #1892's stated scope, with a striking history.

## The measurement

```text
cells=494   agree=112   split=382   strict=0
```

Splits per prefix (of 26): `#b`-family **25** · `#o`-family **24** · `#d`/`#e`/`#i`-family **21** · `#x`-family **16** · **no prefix: 0**.

A cell splits iff the trailing character is not a valid continuation *in that radix* **and any `#` prefix is present**. `strict=0` — the reader is never over-strict, only under-strict.

## Root cause: one bypassed call site

`readNumberPrefixed` (`reader_tokens.zig:481`) calls the **file-local** `readNumber` / `readIntegerWithRadix`, bypassing the delimiter-checking wrappers `Reader.readNumber` (`reader.zig:366`) and `Reader.readIntegerWithRadix` (`reader.zig:646`).

That single bypass subsumes both #1892's radix cases *and* the decimal-path widening found in #1917 — same defect, one fix.

## The archaeology: the guard was born dead

`Reader.readIntegerWithRadix` has **zero callers**, today and at the commit that created it.

Commit `9e8cf95a` (2026-07-02, *"Require delimiter after numeric tokens in the reader"*) added the guard to it. At that commit the only caller was already `reader_tokens.zig:388` — `readIntegerWithRadix(self, radix)` unqualified, which inside `reader_tokens.zig` resolves to the file-local `:508`, not the `Reader` method.

**So the radix half of that fix has never executed.** Its sibling change to `Reader.readNumber` *did* take effect, because `nextToken` calls `self.readNumber()` — which is exactly why the un-prefixed row is the only correct one in the matrix. Zig does not warn on an uncalled private function, so it compiled silently and the accompanying tests passed via the working half.

## Three unguarded returns #1892 does not name

| path | example | result |
|---|---|---|
| SRFI 270 hex-float suffix (`readHexFloatSuffix`, `:829`) | `#x1p4z` | `16.0` + `z` |
[...]

PR #2245, opened 2026-08-06
Read radix-prefixed complex numbers per R7RS <complex R> (#2243)

### LINKED ISSUE #2243: Reader rejects the valid R7RS radix-16 complex #x1/2+3i (guile reads it; chibi parity is the blocker)
**Category:** Reader · **Severity:** conformance · **Found by:** follow-up to #1929 verification

#2241 closed the #1929 delimiter-gap (all 382 split cells now raise). During independent verification of that fix, one cell in the pinned failure set turned out to be a *valid* R7RS number that the fix now rejects — and the issue texts oracle claim ("Guile 3 raises on every cell pinned as a failure") is inaccurate for it.

## The cell

`#x1/2+3i` is a radix-16 complex with a rational real part. R7RS 7.1.1:

```text
<complex R> → <real R> + <ureal R> i | …
<real R>   → <sign> <ureal R>
<ureal R>  → <uinteger R> | <uinteger R> / <uinteger R> | <decimal R>
```

So `#x1/2+3i` → `<real 16> + <ureal 16> i` → `1/2 + 3 i` — grammatically valid.

## Measured behaviour (all three on the same cells)

| cell | kaappi (post-#2241) | chibi 0.12 | guile 3.0.11 |
|---|---|---|---|
| `#x1/2+3i` | `KP1004 invalid number literal` | read error | `0.5+3.0i` |
| `#o1+2i` | read error | read error | `1.0+2.0i` |
| `#d1/2+3i` | `1/2+3i` | `1/2+3i` | `0.5+3.0i` |
| `#b1+2i` | read error | read error | read error |

kaappi and chibi agree on every cell (the fix is at exact parity with the projects pinned differential oracle); guile is the grammar-conformant outlier. `#x1/2+3i` is not a regression from #2241 — pre-fix it silently split into `1/2` + symbol `+3i`, post-fix it is a loud read error — but the *current* behaviour still refuses a legal R7RS program.

## The decision point

Should kaappi read radix-prefixed complex suffixes with rational real parts (`#x1/2+3i` → exact `1/2+3i`, matching guiles acceptance), or keep chibi-parity rejection?

- **For accepting:** the token is valid R7RS; a conforming program that uses it today gets a loud error. The radix-10 path already consumes `+...i` (`#d1/2+3i` reads), so the grammar machinery exists.
- **For rejecting:** chibi 0.12 (the campaigns pinned differential-fuzz oracle) rejects it; the R7RS suite (1395/1395) passes with the current behaviour, and the split damage model of #1929 is fully closed either way.

## Suggested fix (if accepted)

Extend the post-body delimiter check in `readNumberPrefixed` (src/reader_tokens.zig) so a non-decimal radix consumes a complex `+<ureal R> i` / `-<ureal R> i` tail, mirroring the radix-10 path — then verify `#x1/2+3i` → `1/2+3i` and re-run the delimiter sweep to confirm no pinned cell flips. **Not implemented or tested** — this is a decision-first issue.

**Verified:** fresh ReleaseSafe build at `ff9fd4c0`; cells cross-checked against chibi-scheme 0.12 and guile 3.0.11; error path confirmed via both `kaappi check` (KP1004) and `read` (KP3000).

baijum:
Closes #2243.

## Summary

The #1929 delimiter fix turned every non-decimal radix complex spelling into a read error, but `#x1/2+3i`, `#x1+2i`, `#x1+i`, `#b1+1i` are **valid R7RS 7.1.1**: `<complex R> → <real R> + <ureal R> i` (and the `-`/`+i`/`-i` twins) holds in every radix. guile, Chez 10.4.1, and kaappi's own radix-10 path all read them — the non-decimal paths were the odd ones out.

- **Reader** (`readIntegerWithRadix`): consumes a complex tail with radix-valid digits and optional rational parts, producing an exact complex token exactly like the decimal path does for `1/2+3i`. The bare-sign pure imaginary `#x+i` → `0+1i` is grammar too (Chez agrees).
- **`string->number`** (`parseNumberText`): the complex branch is no longer radix-10-only; split forms parse in every radix. This also closes the documented `FAIL: TBD` where `string->number "1/2+3i"` returned `#f` for a valid R7RS complex (Chibi and guile both accept it).

The guard rails stay, in both parsers: `#b1+2i` (2 is not a binary digit), `#o1+8i`, `#x1+2` (no `i`), `#x1+2iz` (glued tail), and the signless `#x3i`/`#xi` still error; bignum components stay a loud error (kaappi#2182 stance — an exact bignum part has no honest f64 value).

## Differential evidence

Fresh ReleaseSafe build vs the three oracles on the full matrix:

| cell | kaappi | chez 10.4.1 | guile 3.0.11 | chibi 0.12 |
|---|---|---|---|---|
| `#x1/2+3i` | `1/2+3i` | `1/2+3i` | `0.5+3.0i` | error |
| `#x1+2i` | `1+2i` | `1+2i` | `1.0+2.0i` | error |
| `#x1+i` | `1+1i` | `1+1i` | `1.0+1.0i` | error |
| `#b1+1i` | `1+1i` | `1+1i` | `1.0+1.0i` | error |
| `#x1+1/2i` | `1+1/2i` | `1+1/2i` | `1.0+0.5i` | error |
| `#x+i` | `+i` | `0+1i` | — | error |
| `#b1+2i` | error | error | error | error |
| `#x1+2` | error | error | error | error |
| `#x3i` | error | error | error | error |
| `1/2+3i` (s->n) | `0.5+3.0i` | — | `0.5+3.0i` | `1/2+3i` |

kaappi now matches Chez on every cell (Chez is the closest oracle: it agrees on both acceptance *and* exactness); guile accepts but reads inexact; chibi rejects all radix complex (the parity we deliberately gave up per the #2243 decision).

## Tests

- **Enables the group-7 TBD** (`s->n accepts 1/2+3i`) and **flips the `#x1/2+3i` pins** in `reader-delimiter-gaps.scm` from reject to read-whole, keeping the glued-tail rejections (`#x1/2+3z`, `#x1/2+3iz`).
- **New group 8** in `reader-delimiter-gaps.scm`: 38 assertions covering the accepted grammar, radix-valid-digit rejections, string->number agreement, bignum-component rejection, and the one-datum-in-a-list guarantee.
- **Zig unit test** (`prefixed numeric tokens require a trailing delimiter`) extended: `#x1/2+3i`/`#x1+2i`/`#b1+1i`/`#x+i` read, `#b1+2i`/`#o1+8i`/`#x1+2`/`#x1+2iz`/`#x3i`/`#xi`/bignum error.
- **CHANGELOG** entry under Unreleased → Fixed.

## Verification

- `zig build test` — full unit suite green (1708 tests: 1701 passed, 7 skipped, 0 failed)
- Reader tests green under `-Dgc-stress=true` (all 74 reader tests)
- `tests/scheme/run-all.sh` — 691 scheme files + 1395 R7RS = **2086 pass, 0 fail**
- `zig build wasm` green
- `kaappi check` on `(define x (quote (#x1/2+3i)))` → no issues; on `(#x1/2+3iz)` → `KP1004`

### DIFF
--- a/src/bignum.zig
+++ b/src/bignum.zig
@@ -857,6 +857,70 @@ pub fn stripUnderscores(s: []const u8, radix: u8, buf: []u8) ?[]const u8 {
     return buf[0..n];
 }
 
+/// True when converting the i64 to f64 is lossless. Complex-number
+/// components claim exactness only inside this range: values in
+/// (2^53, 2^63] round to a different integer, so an exact-flagged token
+/// would silently carry the wrong value (kaappi#2182/#2243 -- loud
+/// rejection over silent rounding).
+pub fn f64ExactI64(n: i64) bool {
+    const f: f64 = @floatFromInt(n);
+    return @as(i64, @intFromFloat(f)) == n;
+}
+
+/// True when a bignum Value's magnitude is exactly representable in f64:
+/// at most 53 significant bits after stripping trailing binary zeros
+/// (10^19 = 5^19*2^19 has 45 bits and qualifies; 2^53+1 does not). The
+/// exact-complex machinery carries components as f64 plus an exactness
+/// flag, so a non-representable integer would silently round (kaappi#2243).
+pub fn bignumExactInF64(v: Value) bool {
+    const bn = types.toBignum(v);
+    if (bn.len == 0) return true;
+    var trailing: u64 = 0;
+    var i: usize = 0;
+    while (i < bn.len and bn.limbs[i] == 0) : (i += 1) trailing += 64;
+    if (i >= bn.len) return true; // zero
+    trailing += @ctz(bn.limbs[i]);
+    const top = bn.limbs[bn.len - 1];
+    const bit_len: u64 = 64 * @as(u64, @intCast(bn.len - 1)) + 64 - @clz(top);
+    return bit_len - trailing <= 53;
+}
+
+/// Largest |numerator|/|denominator| an exact-flagged complex component
+/// may carry. The exact-complex printer recovers rationals via
+/// `floatToRational`'s brute-force search, which tries denominators up to
+/// this limit; a component within it round-trips exactly, while a larger
+/// one would print the rounded f64's mantissa/2^k value -- a silently
+/// wrong number claiming exactness. Values beyond the limit are rejected
+/// loudly, the same policy as i64-overflowing components (kaappi#2182/
+/// #2243).
+pub const complex_rational_limit: i64 = 1_000_000;
+
+/// Parse a radix-R <ureal> slice (radix digits with optional SRFI-169 `_`
+/// separators, optionally `N/D`) to f64. No sign -- callers handle signs.
+/// i64 overflow, an invalid character, a zero denominator, or an i64 part
+/// that does not round-trip through f64 (see `f64ExactI64`) yields null.
+/// This is the single implementation shared by the reader and
+/// `string->number` so the two complex grammars cannot drift (kaappi#2243).
+pub fn parseRadixUrealToF64(s: []const u8, radix: u8) ?f64 {
+    if (std.mem.indexOfScalar(u8, s, '/')) |sp| {
+        if (sp == 0 or sp + 1 >= s.len) return null;
+        var nb: [256]u8 = undefined;
+        var db: [256]u8 = undefined;
+        const num = stripUnderscores(s[0..sp], radix, &nb) orelse return null;
+        const den = stripUnderscores(s[sp + 1 ..], radix, &db) orelse return null;
+        const n = std.fmt.parseInt(i64, num, radix) catch return null;
+        const d = std.fmt.parseInt(i64, den, radix) catch return null;
+        if (d == 0) return null;
+        if (@abs(n) > complex_rational_limit or @abs(d) > complex_rational_limit) return null;
+        return @as(f64, @floatFromInt(n)) / @as(f64, @floatFromInt(d));
+    }
+    var b: [256]u8 = undefined;
+    const clean = stripUnderscores(s, radix, &b) orelse return null;
+    const n = std.fmt.parseInt(i64, clean, radix) catch return null;
+    if (!f64ExactI64(n)) return null;
+    return @as(f64, @floatFromInt(n));
+}
+
 // ---------------------------------------------------------------------------
 // Helpers
 // ---------------------------------------------------------------------------
--- a/src/bignum.zig
+++ b/src/bignum.zig
@@ -864,6 +864,11 @@ pub fn stripUnderscores(s: []const u8, radix: u8, buf: []u8) ?[]const u8 {
 /// rejection over silent rounding).
 pub fn f64ExactI64(n: i64) bool {
     const f: f64 = @floatFromInt(n);
+    // The top 512 i64 values (2^63-512 .. 2^63-1) round UP to 2^63, which
+    // no longer fits the i64 destination of @intFromFloat -- a ReleaseSafe
+    // panic reachable from a one-line program (kaappi#2243 review). They
+    // never round-trip, so reject before the conversion.
+    if (f >= 9223372036854775808.0) return false;
     return @as(i64, @intFromFloat(f)) == n;
 }
 
--- a/src/primitives_numeric.zig
+++ b/src/primitives_numeric.zig
@@ -1128,6 +1128,38 @@ fn stringToNumber(args: []const Value) PrimitiveError!Value {
     return parseNumberText(gc, str.data[0..str.len], radix, .unspecified);
 }
 
+/// Parse a signed radix-R <ureal> (integer, N/D rational, or -- radix 10
+/// only -- a decimal) to f64, for use as a complex component. Mirrors the
+/// reader's complex-component grammar so `string->number` and `read` agree
+/// on `#x1+2i`, `#x1/2+3i` and friends (kaappi#2243): i64 overflow yields
+/// null (an exact bignum component has no honest f64 value -- loud
+/// rejection over silent rounding, cf. kaappi#2182).
+fn parseComplexPartToF64(part: []const u8, radix: u8) ?f64 {
+    // Rational N/D (any radix): R7RS <ureal R>.
+    if (std.mem.indexOfScalar(u8, part, '/')) |sp| {
+        if (sp == 0 or sp + 1 >= part.len) return null;
+        var nb: [256]u8 = undefined;
+        var db: [256]u8 = undefined;
+        const num = bignum_mod.stripUnderscores(part[0..sp], radix, &nb) orelse return null;
+        const den = bignum_mod.stripUnderscores(part[sp + 1 ..], radix, &db) orelse return null;
+        const n = std.fmt.parseInt(i64, num, radix) catch return null;
+        const d = std.fmt.parseInt(i64, den, radix) catch return null;
+        if (d == 0) return null;
+        return @as(f64, @floatFromInt(n)) / @as(f64, @floatFromInt(d));
+    }
+    if (radix == 10) {
+        // Decimal parts (1.5+2i, 1e5+2i) and inf/nan spellings: parseFloat
+        // accepts them with the sign included. Unchanged acceptance from
+        // before #2243.
+        return std.fmt.parseFloat(f64, part) catch null;
+    }
+    // Radix 2/8/16: integer part only -- <decimal R> does not exist there.
+    var b: [256]u8 = undefined;
+    const clean = bignum_mod.stripUnderscores(part, radix, &b) orelse return null;
+    const n = std.fmt.parseInt(i64, clean, radix) catch return null;
+    return @as(f64, @floatFromInt(n));
+}
+
 /// Parse a complete number text into a Value, or `types.FALSE` when the text
 /// is not a number. This is the single number grammar behind BOTH
 /// `string->number` and the reader's `#e`/`#i` literals: R7RS 6.2.7 requires
@@ -1201,7 +1233,13 @@ pub fn parseNumberText(gc: *@import("memory.zig").GC, text: []const u8, radix_in
 
     // Rational: num/den
     if (std.mem.indexOfScalar(u8, s, '/')) |slash_pos| {
-        if (slash_pos > 0 and slash_pos + 1 < s.len) {
+        // A trailing 'i' means the '/' is a complex-component separator
+        // (1/2+3i, 1/2+3/4i, #x1/2+3i): the complex branch below owns
+        // those, parsing each component as a <ureal R> (kaappi#2243).
+        // Without this guard, 1/2+3i would try "2+3i" as a denominator,
+        // fail, and return #f for a valid R7RS complex.
+        const complex_shaped = s.len >= 2 and s[s.len - 1] == 'i';
+        if (!complex_shaped and slash_pos > 0 and slash_pos + 1 < s.len) {
             const num_str = s[0..slash_pos];
             const den_str = s[slash_pos + 1 ..];
             const num_val: Value = if (std.fmt.parseInt(i64, num_str, radix)) |num|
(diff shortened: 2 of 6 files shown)


src/reader_tokens.zig
```diff
@@ -844,7 +944,23 @@ pub fn readIntegerWithRadix(self: *Reader, radix: u8) ReadError!Token {
 
     const num_str = self.source[start..self.pos];
     if (num_str.len > Reader.MAX_TOKEN_BYTES) return ReadError.TokenTooLong;
-    if (num_str.len == 0 or (num_str.len == 1 and (num_str[0] == '+' or num_str[0] == '-'))) return invalidNumberOrEof(self);
+    if (num_str.len == 0) return invalidNumberOrEof(self);
+    if (num_str.len == 1 and (num_str[0] == '+' or num_str[0] == '-')) {
```
baijum, line 990: **Signed pure-imaginary with a magnitude (`+<ureal R> i`) is left unimplemented for non-decimal radices — and the result is a surprising split.**

This branch reads `#x+i` / `#x-i` (the R7RS `+ i` / `- i` productions), and `tryComplexTail` reads `<real R> + <ureal R> i`. But the productions `<complex R> → + <ureal R> i | - <ureal R> i` (a *signed* pure imaginary with an explicit magnitude) are never accepted:

| spelling | value | result |
|---|---|---|
| `#x+i` | 0+1i | reads ✓ |
| `#x+1i` | 0+1i | **read error** ✗ |
| `#x+3i` | 0+3i | **read error** ✗ |
| `#x0+3i` | 0+3i | reads ✓ |

`readIntegerWithRadix` consumes `+3` as the real integer `3`, then hits a bare `i` with no sign for `tryComplexTail`, so the delimiter check rejects it. Net effect: `#x+i` works but the identical-valued `#x+1i` doesn't, and `#x0+3i` works but the identical-valued `#x+3i` doesn't.

These are valid R7RS in every radix (Chez/guile both read `#x+3i` as `0+3i`), and none appear in the new differential matrix or tests. Is dropping this production intentional? If so, the CHANGELOG/PR text should say the `+<ureal R> i` form stays radix-10-only — not just the truly-signless `2i` extension. If not, `readIntegerWithRadix` can special-case "signed integer/rational immediately followed by `i`+delimiter" into a pure-imaginary complex, mirroring the radix-10 pure-imaginary branch.

src/primitives_numeric.zig
```diff
@@ -1276,59 +1314,71 @@ pub fn parseNumberText(gc: *@import("memory.zig").GC, text: []const u8, radix_in
...
-            }
-            if (std.mem.eql(u8, body, "-")) {
-                const c = gc.allocComplex(0.0, -1.0) catch return PrimitiveError.OutOfMemory;
-                return applyExactness(gc, c, exactness);
-            }
+    // Try parsing as complex: a+bi, a-bi, +bi, -bi, +i, -i (R7RS 7.1.1
+    // <complex R>). Radix-prefixed spellings (#x1+2i, #x1/2+3i) are valid
+    // R7RS and the reader now accepts them too (kaappi#2243), so this
+    // branch is no longer radix-10-only: the split forms work in every
+    // radix, while decimal real/imag parts and the signless pure-imaginary
```
baijum: This characterization is inaccurate and is what drives the reader-side gap. `+3i` is **not** "the signless pure-imaginary extension" — it carries a `+` sign and is the core R7RS production `<complex R> → + <ureal R> i`, valid in every radix. The genuine kaappi extension is `2i` (no sign, absent from the R7RS grammar). Lumping `+3i` with `2i` as radix-10-only is precisely what makes `#x+i` read while `#x+1i` / `#x+3i` error (see the reader_tokens.zig comment). Suggest separating the two here, and deciding whether `+3i` should actually be accepted per R7RS.

src/primitives_numeric.zig
```diff
@@ -1276,59 +1314,71 @@ pub fn parseNumberText(gc: *@import("memory.zig").GC, text: []const u8, radix_in
...
+    // extension (2i, +3i) stay radix-10-only -- exactly mirroring the
+    // reader's grammar (which rejects `#x3i` and `#x2/3i`).
+    if (s.len >= 2 and s[s.len - 1] == 'i') {
```
baijum: **Case divergence between `read` and `string->number`.** The reader's `tryComplexTail` and bare-sign branch accept both `i` and `I` (`... == 'i' or ... == 'I'`), but this branch only tests lowercase `'i'`. So `(read (open-input-string "#x1+2I"))` yields a complex while `(string->number "#x1+2I")` returns `#f` — the two disagree, which is exactly the agreement property this PR advertises. It's pre-existing for radix 10 (the radix-10 reader already accepts `1+2I`), but the new code extends the disagreement to every radix. Fold-case the marker here (or reject uppercase `I` in the reader) so the two parsers stay in lock-step.

src/reader_tokens.zig
```diff
@@ -813,6 +813,106 @@ fn appendByteStringByte(self: *Reader, b: u8) ReadError!void {
     self.token_buf.append(self.gc.allocator, b) catch return ReadError.OutOfMemory;
 }
 
+/// Parse a radix-R <ureal> slice (radix digits, optional SRFI-169 `_`
+/// separators, optionally `N/D`) to f64. i64 overflow or any invalid
+/// character yields null: an exact bignum component has no honest f64
+/// value, and a loud error beats a silently rounded one (the kaappi#2182
+/// stance, same as the radix-10 complex grammar's i64 bound).
+fn parseRadixUrealToF64(s: []const u8, radix: u8) ?f64 {
```
baijum, line 848: Nit / maintainability: `parseRadixUrealToF64` here and `parseComplexPartToF64` in `primitives_numeric.zig` are near-duplicates (radix `N/D`-or-integer → f64, i64-overflow→null, same kaappi#2182 rationale); the `primitives_numeric` copy only adds a radix-10 float branch. Since the entire point of #2243 is that the two parsers agree, a future fix to one could silently skew the other. Consider factoring the shared radix-N ureal→f64 logic into a single helper (e.g. in `bignum`/`arith`). Not blocking.

baijum (commented):
Solid, careful change — the token-level exact-complex parse mirrors the radix-10 path faithfully, the delimiter/glued-tail guard rails are preserved, the differential table against three oracles is convincing, and the test coverage (group 8 + the extended Zig unit test) is thorough. Traced the reader and `string->number` paths against each other on the accepted grammar, the radix-valid-digit rejections, and the bignum-component path; they line up on everything the new tests assert.

Two behavioral notes and one maintainability note below. The main one is a genuine R7RS-grammar gap the PR silently leaves open (`#x+3i` / `#x+1i`), which sits oddly next to the `#x+i` the PR *does* now accept. None are blocking; the first is worth a decision before merge since the PR frames itself as "per R7RS `<complex R>`".

baijum:
Review round addressed in `3b87e2ea` — all five CodeRabbit findings and both of the human review threads are closed. Findings-by-finding:

**CodeRabbit**
1. *radix 19–36 `i`-as-digit* (major, real regression) — confirmed: `(string->number "1/2i" 19)` was `#f` instead of `1/56`. Fixed: the trailing-`i` complex detection is gated on `radix <= 18` in both the rational-branch guard and the complex branch. New group 9 pins it.
2. *allocComplexEx with derived exactness* (minor) — done. `string->number` now derives component exactness from the text (integer/rational parts exact, decimals/exponents inexact), matching the reader's exact-flagged tokens: `(string->number "1+2i")` → exact `1+2i`, `"1/2+3i"` → exact `1/2+3i`, `"1.5+2i"` → inexact real. `#e`/`#i` still override via `applyExactness`. No existing test asserted the old inexact default.
3. *unify radix `<ureal>` parsing* (trivial) — done. `bignum.parseRadixUrealToF64` is now the single implementation; both readers and `parseComplexComponent` delegate to it.
4. *2^53–2^63 exact-complex claim* (minor, real) — done and extended beyond the ask. Integer components that don't round-trip through f64 (`#x20000000000001+2i`, `9007199254740993+2i`) are rejected loudly in the new radix paths **and** the radix-10 `readNumber` paths (all five complex sites), and in `string->number` — consistent with the existing bignum rejection, so the two parsers cannot diverge. Non-representable bignums are caught by a significant-bit test (`bignumExactInF64`), so `#e1e19+1i` (45 significant bits) still round-trips. Rational parts beyond the exact-complex printer's recovery granularity (`floatToRational` searches denominators to 1e6) are rejected too — `1/20000000000001+3i` would otherwise print a wrong mantissa/2^k fraction.
5. *`complex?` too weak* (trivial) — the three assertions now check `real-part`/`imag-part` numerically, plus new s->n cells for `#x+i`, `#x-i`, `#b+i` and the bignum rejections.

**baijum**
1. *`#x+3i` / `#x+1i` signed pure-imaginary gap* — fixed in full: `+ <ureal R> i` / `- <ureal R> i` (integer *and* rational magnitudes) now read in every radix in both readers and `string->number` — including the pre-existing radix-10 `+3/4i` gap the agreement property demanded. Chez reads all of these (`#x+3i` → `0+3i`, `#x+3/4i` → `0+3/4i`). The signless `#x3i`/`#x3/4i` stay rejected.
2. *characterization of `+3i` as "signless extension"* — corrected: the comment now separates the R7RS `+ <ureal R> i` production from the genuine signless `2i` extension.
3. *`I`/`i` case divergence* — fixed: `string->number` folds the marker case, so `(string->number "#x1+2I")` agrees with the reader.
4. *duplicate helper* — closed by the `bignum.parseRadixUrealToF64` unification (CodeRabbit #3).

Validation: full unit suite 0 failures (1700+ tests), `tests/scheme/run-all.sh` **2088 pass / 0 fail**, reader tests green under `-Dgc-stress=true`, `zig build wasm` green, and the differential matrix re-run against Chez 10.4.1 on every new cell.

One deliberate divergence to flag: chez reads `1/20000000000001+3i` exactly, kaappi rejects it — kaappi's exact-complex representation is f64+exactness-flag (not a true rational component, kaappi#2182 family), so a ratio the printer cannot recover would print a wrong value. Loud rejection is the project's documented stance for that. If the exact-complex type ever grows a true rational component, the granularity bound disappears.

src/bignum.zig
```diff
@@ -857,6 +857,70 @@ pub fn stripUnderscores(s: []const u8, radix: u8, buf: []u8) ?[]const u8 {
...
 }
 
+/// True when converting the i64 to f64 is lossless. Complex-number
+/// components claim exactness only inside this range: values in
+/// (2^53, 2^63] round to a different integer, so an exact-flagged token
+/// would silently carry the wrong value (kaappi#2182/#2243 -- loud
+/// rejection over silent rounding).
+pub fn f64ExactI64(n: i64) bool {
+    const f: f64 = @floatFromInt(n);
+    return @as(i64, @intFromFloat(f)) == n;
```
baijum, line 872: **Critical — reachable abort: `f64ExactI64` panics on the top 512 `i64` values instead of rejecting them.**

`@intFromFloat` is checked illegal behavior when the source float is out of the destination range. For `n` in `[2^63−512, 2^63−1]` (i.e. `9223372036854775296 … 9223372036854775807`), `@floatFromInt(n)` rounds **up** to `2^63`, which no longer fits in `i64`, so `@as(i64, @intFromFloat(f))` panics in ReleaseSafe. The value should have been rejected (`#f`) exactly like `9007199254740993+2i` already is — instead the interpreter aborts.

Reachable from a one-line ordinary program (fresh ReleaseSafe build of `3b87e2ea`, aarch64-macos):

```
$ echo '(string->number "9223372036854775807+2i")' | kaappi /dev/stdin
thread ... panic: integer part of floating point value out of bounds
  bignum.zig:867            f64ExactI64
  primitives_numeric.zig:1412  parseComplexComponent
  primitives_numeric.zig:1128  stringToNumber
```

Also via the reader: `(read (open-input-string "#x7fffffffffffffff+2i"))`. I bisected the boundary: `9223372036854775295` is fine, `9223372036854775296` and up crash (the 512 values whose f64 rounds to `2^63`). `i64` min is safe (`−2^63` is representable and round-trips), and bignum magnitudes are safe (they route through `bignumExactInF64`, which avoids `@intFromFloat`). Per the project rubric this is `critical` — a process abort reachable from a five-line program.

Fix: guard the top edge before the float→int round-trip, e.g.

```zig
pub fn f64ExactI64(n: i64) bool {
    const f: f64 = @floatFromInt(n);
    if (f >= 9223372036854775808.0) return false; // rounds to/past 2^63 — not exact, and out of i64 range
    return @as(i64, @intFromFloat(f)) == n;
}
```

(or derive exactness from the significant-bit count, like `bignumExactInF64`, and skip the round-trip entirely). Worth a regression cell for the `2^63` band — the new Zig unit test exercises `2^53±` but not this edge, which is why it slipped through the green suite.

baijum (commented):
Re-reviewed `3b87e2ea`. All nine prior findings are genuinely closed — verified on a fresh ReleaseSafe build: the signed pure-imaginary production (`#x+3i` → `+3i`, `#x+3/4i`), case-folding (`(string->number "#x1+2I")` → `1+2i`), the radix-19 `i`-digit gate (`(string->number "1/2i" 19)` → `1/56`), derived exactness, and the `2^53+1` loud rejection (`9007199254740993+2i` → `#f`) all behave as described. The `bignum.parseRadixUrealToF64` unification is clean.

However, the exactness-honesty fix introduced a **critical, reachable crash** — details inline on `f64ExactI64`. One blocker; everything else is good to merge once it's addressed.

baijum:
Round 3 addressed in `c2bb3f85`. Both reviewers found the same critical bug — thanks for the catch, and the boundary bisection (2^63−512 / 2^63−5295/5296) was exactly right.

**Critical: `f64ExactI64` panic on the top 512 i64 values** — confirmed and fixed. `@floatFromInt(n)` for n ∈ [2^63−512, 2^63−1] rounds up to 2^63, and `@intFromFloat(2^63)` is checked illegal behavior in ReleaseSafe — `(string->number "9223372036854775807+2i")` aborted the interpreter. `f64ExactI64` now rejects `f >= 2^63` before the conversion; those values never round-trip anyway. Verified: the top-512 cells reject cleanly (#f / read error, no abort), `−2^63+2i` and `2^63` (power of two, still exact) keep working.

**The rest of the round:**
- *Signless pure-imaginary integer path lacks bignum fallback* (CodeRabbit, major) — `(string->number "10000000000000000000i")` was #f while the reader read `0+1e19i` exactly. The signless branch now falls back to `parseBignumString` + `bignumExactInF64` like every other integer-component path. Note: I did *not* take the proposed shortcut of delegating to `parseComplexComponent`, because that would also accept the signless rational `2/3i`, which the reader still rejects — the integer-only structure is load-bearing.
- *Special-float imaginary parts* (CodeRabbit, minor) — `3.0+inf.0i` / `+inf.0i` read in the reader but `string->number` returned #f, because Zig's `parseFloat` rejects `+inf.0` (the old radix-10 code had the same gap). `parseComplexComponent` now names the four special spellings explicitly, matching the reader's grammar; all four cells agree in both parsers.
- *CHANGELOG contradiction* (CodeRabbit, minor) — the #1929 entry's read-error list still named `#x1/2+3i`; it now lists `#x1zzz` and cross-references #2243.
- *Weak `string->number` assertions* (CodeRabbit, minor) — the remaining `complex?`-wrapped cells are replaced with `real-part`/`imag-part` equality checks, plus new cells pinning the inf/nan and bignum-magnitude agreement.

Validation: full unit suite 0 failures, `run-all.sh` **2087 pass / 0 fail**, reader tests green under `-Dgc-stress=true`, `zig build wasm` green, both differential matrices re-run.

baijum:
Verified `c2bb3f85` on a fresh ReleaseSafe build — the critical panic is resolved and the four divergences are closed.

**Crash fix confirmed:** the `f >= 2^63` guard in `f64ExactI64` is the right shape (`@intFromFloat` is now only reached for `f < 2^63`, and every i64's `@floatFromInt` is `>= -2^63`, so both destination edges are in range). Re-ran the whole former crash band — all reject cleanly, no abort:

- `(string->number "9223372036854775807+2i")` → `#f`
- `(string->number "1+9223372036854775807i")` → `#f`
- `(string->number "9223372036854775296+2i")` (2^63−512) → `#f`
- `(read (open-input-string "#x7fffffffffffffff+2i"))` → read error
- `9223372036854774784+2i` (2^63−1024, representable) → still accepted
- `-9223372036854775808+2i` (i64 min) → still accepted

**Divergences confirmed closed:** `10000000000000000000i` now reads `+10000000000000000000i` in both parsers; `+inf.0i` and `3.0+inf.0i` agree; and `2/3i` correctly stays `#f` — good call keeping the integer-only structure rather than delegating to `parseComplexComponent`, which would have leaked the signless rational.

`reader-delimiter-gaps.scm` → 277 passes, and the `prefixed numeric tokens require a trailing delimiter` unit test is green. Nothing further from me — LGTM.

PR #2247, opened 2026-08-07
Fix cross-thread heap safety: reject cross-heap stores (#1924) and mark live children's roots (#1933)

### LINKED ISSUE #1924: Use-after-free: a child thread mutating a shared heap object installs a child-heap pointer that dangles after thread-join!
**Category:** SRFI-18 threads / GC · **Severity:** crash / silent memory corruption · **Found by:** systematic audit v2, Phase 3.1 (#1890)

A child thread that mutates a heap object reachable from a top-level binding stores a **child-heap pointer** into a **parent-heap** object. The child heap is freed at `thread-join!`, so the parent is then reading freed memory. Deterministic, not a race.

## Reproduction

```scheme
(define-record-type <box> (make-box v) box? (v box-v set-box-v!))
(define b (make-box #f))
(define t (make-thread (lambda () (set-box-v! b (list 1 2 3)))))
(thread-start! t) (thread-join! t)
(write (box-v b))
```

Expected `(1 2 3)`. Actual, **6 of 6 runs**:

```text
((box-v b))
```

That is the program's **own source text**, read back out of a recycled slot.

## It is general, and it type-confuses

| child writes | parent reads back |
|---|---|
| pair, via record field `set!` | `((box-v b))` — recycled source text |
| pair, via `vector-set!` | `("vector-set! pair  -> ")` — a recycled string |
| string, via `set-car!` | **`#<procedure>`** |
| **fixnum (control)** | **`7` — correct** |

The `#<procedure>` case is the alarming one: a freed slot is being read as a *callable*. A program that then applies it is calling into freed memory.

## Discriminating control

Immediates survive; heap values do not. A fixnum written by the child reads back correctly every time. That is exactly the expected signature of a dangling pointer rather than a logic bug — fixnums are NaN-boxed inline and carry no pointer.

## Why the architecture allows it

`CLAUDE.md` documents the model: each thread gets its own VM and GC with an independent heap, and values are **deep-copied at the two boundaries** — the thunk closure at `thread-start!`, the result at `thread-join!`. It also documents that **top-level bindings are shared by pointer** (`initForThread` shares `globals`), unlike lexical captures.

Those two facts together leave a gap: a mutation *through* a shared pointer is neither of the two copied boundaries. Nothing deep-copies the assigned value, and nothing keeps the child heap alive for it. `lib/srfi/120.sld`'s header already warns that a top-level binding "reaches the ORIGINAL object from the wrong heap" — this is the write direction of that same hazard, and unlike the channel case there is no primitive-level check to catch it.

## Consequence beyond the direct case

**All portable memoisation is unsafe across threads**, including R7RS `delay`/`force` and any `(define cache #f)` + fill-on-first-use idiom, whenever the filling happens on a child thread. This was found while implementing lazily-built SRFI 14 constants: a child thread forcing the constant first left the parent reading freed memory (`type error in 'cdar': expected pair, got 0`). That is why #1895's implementation uses generated literal tables instead — the lazy design is not safe in this engine today.

## Suggested direction

Options, roughly in increasing cost: reject the write (as channels already do for cross-thread use), deep-copy the assigned value into the owning heap at the write barrier, or keep a child heap alive while any parent object references it. The first is the smallest change and matches the existing channel precedent; it would need the write barrier to know the owning GC, which `Object.owner` already records (#958).

**Verified:** fresh ReleaseSafe build at `12588782`, isolated `KAAPPI_HOME`; the record case 6/6, the vector/`set-car!`/fixnum matrix reproduced twice. Not tested under `-Dgc-stress=true`.

### LINKED ISSUE #1933: The parent's collector reclaims objects a live child thread still references (gc-stress: hard use-after-free panic)
**Category:** GC / SRFI-18 threads · **Severity:** crash / memory corruption · **Found by:** systematic audit v2, Phase 5C (#1890)

The parent's garbage collector cannot see a live child thread's roots, so it reclaims objects the child is still using. **This is the opposite direction from #1924** — no mutation, no `thread-join!`, no cross-heap write is involved.

## Mechanism

`markVMRoots` (`vm.zig`) returns early unless `vm.gc == gc`, so a child's registers are invisible to the parent's root marker. The child's own marker independently skips foreign-owner objects (#958). Between them, an object owned by the parent and referenced **only** from a running child's registers is unreachable to both markers.

## Reproduction shape

Parent holds the object only via a hash-table entry filled inside a procedure; child reads it into a local through a channel handshake; parent deletes the entry and churns the heap; then releases the child.

| configuration | result |
|---|---|
| ReleaseSafe | **6/6 raise** — `type error in 'cadr': expected pair, got 0` / `got ()`, varying by recycled slot |
| ReleaseSafe, **control** (entry not dropped) | correct read, **3/3** |
| `-Dgc-stress=true` | **hard panic 3/3** — `GC: marking freed object (use-after-free)` in `markValueInner` via `markRoots`' `arg_roots` |
| `-Dgc-stress=true`, control | clean, **3/3** |

The gc-stress escalation from wrong-value to a detected UAF is what identifies this as GC-localised rather than a logic error.

## A retention detail worth knowing before reproducing

A top-level `(define g …)` followed by `(set! g #f)` is **not** collectable — stale registers in the base frame retain it (verified with a SRFI 254 guardian: an object never bound at all *is* collected; one that was bound to a global is not).

That is why this needs the hash-table shape to trigger, and why it can look unreproducible from a REPL transcript. Any attempt to reproduce via a plain global will silently fail to drop the object.

## Suggested direction

The parent's collector needs to see live children's roots — either by having `markVMRoots` walk registered child VMs, or by pausing children at a safepoint during a parent collection. Both are substantial; the invariant to restore is that no live thread's registers are invisible to whichever collector owns the object.

**Verified:** fresh ReleaseSafe build at `1c3dec04` plus a separate `-Dgc-stress=true` build (isolated `--prefix`); all run counts above measured, each with its no-drop control.

baijum:
Closes #1924, closes #1933.

Two halves of the same gap: each SRFI-18 OS thread has its own GC heap, top-level bindings are shared by pointer, and nothing coordinated the two collectors on the mutation path. Both were deterministic use-after-frees — #1924 on every run, #1933 wrong values or a hard `GC: marking freed object` panic under `-Dgc-stress`.

## #1924 — reject cross-heap stores

A child storing one of its own heap's objects into a shared parent-heap container (record field, vector slot, pair, hash-table entry, a promise's memoised value, or the globals map) left a pointer the parent's collector skips as foreign and the child's collector cannot see a reference to. The store is now **rejected before it happens** (`memory.crossHeapStoreViolation`, checked at every general mutation site), unless the value belongs to the container's own heap. The `mutex-lock!` owner pair is the one sanctioned exception (a child locking a global mutex must record its own fiber as owner).

## #1933 — the parent's collector marks live children's roots

A parent-heap object referenced only from a live child's registers was unreachable to both markers. The root's collector now stops every live child at the dispatch-loop safepoint (or finds it already parked / in an FFI call) and marks its roots with the root's gc (`markLiveChildRoots`, wired as `gc.child_marker`). Details:

- `CollectionState` (running/parked/stopped/in_native) reported from the safepoint, the park (`parkOnReactor`) and `callFfi` — with function-scope defers (a block-scoped defer fires before the call, leaving the child reported `.running` for the whole blocking wait).
- Children spawned mid-collection wait on `collection_in_progress` before their first shared-globals read.
- Dead/retired children are skipped via a registry `thread_exited` flag — marking a dead child's stale registers is itself a UAF (found via the #2129 grandchild test under gc-stress).
- The `symbol_mutex` section of `markRoots` no longer wraps the child-marking — it would deadlock a child blocked in `allocSymbol` while deep-copying its thunk.

## Tests

- `tests/scheme/srfi/srfi18-cross-heap-mutation-1924.scm` — the full rejection matrix (record/vector/pair/hash-table/promise/global-set!) plus the immediate / Direction-B / mutex controls. 15/15.
- `tests/scheme/srfi/srfi18-child-registers-1933.scm` — the issue's hash-table + channel-handshake shape (deterministic: a second handshake makes the child read only after the parent's churn recycled the slot), with a guardian control proving the churn collects. Both tests fail on the baseline.
- `srfi18-sharing-model.scm` pins the new mutation-refused rows; `docs/dev/thread-value-sharing.md` documents both fixes and the residuals.

Verified: `zig build test` (incl. `-Dgc-stress`), full Scheme suite (695 files + R7RS 1395/1395), the srfi18 suite under gc-stress, WASM build, markdownlint.

### DIFF
--- a/src/fiber_wait.zig
+++ b/src/fiber_wait.zig
@@ -231,6 +231,16 @@ pub fn parkOnReactor(vm: *VM, sched: *FiberScheduler, cap_ns: ?u64) VMError!bool
 
     var ready: std.ArrayList(*Fiber) = .empty;
     defer ready.deinit(vm.gc.allocator);
+    // #1933: a child OS thread blocked in the reactor poll is quiescent; a
+    // collecting parent may mark its registers/frames without waiting for a
+    // safepoint. The main thread (owns_globals) never reports a state — it is
+    // the collector's own thread. The defer must be at FUNCTION scope: a
+    // defer inside the if-block would fire when the block exits — before the
+    // poll — leaving the child reported .running (and unmarkable) for the
+    // whole blocking wait.
+    const report_park_state = !vm.owns_globals;
+    if (report_park_state) vm.setCollectionParked();
+    defer if (report_park_state) vm.setCollectionRunning();
     reactor.poll(cap_ns, &ready) catch return VMError.OutOfMemory;
     // A notify arriving *during* the blocking poll() above is what actually
     // interrupted it; its wake_pending flag needs consuming here even though
--- a/src/fiber_wait.zig
+++ b/src/fiber_wait.zig
@@ -234,14 +234,16 @@ pub fn parkOnReactor(vm: *VM, sched: *FiberScheduler, cap_ns: ?u64) VMError!bool
     // #1933: a child OS thread blocked in the reactor poll is quiescent; a
     // collecting parent may mark its registers/frames without waiting for a
     // safepoint. The main thread (owns_globals) never reports a state — it is
-    // the collector's own thread. The defer must be at FUNCTION scope: a
-    // defer inside the if-block would fire when the block exits — before the
-    // poll — leaving the child reported .running (and unmarkable) for the
-    // whole blocking wait.
+    // the collector's own thread. The window covers the POLL ONLY: the code
+    // after it mutates scheduler state (sweepSharedWaiters, wakeReadyFiber,
+    // markRunnable) that markVmRoots traverses, so reporting quiescence there
+    // would race the parent's mark. setCollectionRunning is the guarded
+    // resume (honours collection_stop).
     const report_park_state = !vm.owns_globals;
     if (report_park_state) vm.setCollectionParked();
-    defer if (report_park_state) vm.setCollectionRunning();
-    reactor.poll(cap_ns, &ready) catch return VMError.OutOfMemory;
+    const poll_result = reactor.poll(cap_ns, &ready);
+    if (report_park_state) vm.setCollectionRunning();
+    poll_result catch return VMError.OutOfMemory;
     // A notify arriving *during* the blocking poll() above is what actually
     // interrupted it; its wake_pending flag needs consuming here even though
     // poll()'s own ReadyEvent list never surfaces the notifier's own event
--- a/src/primitives_hashtable.zig
+++ b/src/primitives_hashtable.zig
@@ -653,6 +653,10 @@ fn hashTableSetFn(args: []const Value) PrimitiveError!Value {
     const ht = try getHashTable("hash-table-set!", args[0]);
     try growIfNeeded(ht);
     const slot = try findSlot("hash-table-set!", ht, args[1]);
+    // #1924: reject before the store — a shared parent-heap table must not
+    // come to hold a child-heap key or value.
+    if (memory.crossHeapStoreViolation(types.toObject(args[0]), args[1])) return primitives.raiseCrossHeapStore("hash-table-set!");
+    if (memory.crossHeapStoreViolation(types.toObject(args[0]), args[2])) return primitives.raiseCrossHeapStore("hash-table-set!");
     if (memory.gc_instance) |gc| {
         gc.writeBarrier(types.toObject(args[0]), args[1]);
         gc.writeBarrier(types.toObject(args[0]), args[2]);
--- a/src/primitives_hashtable.zig
+++ b/src/primitives_hashtable.zig
@@ -852,6 +852,10 @@ fn hashTableUpdateFn(args: []const Value) PrimitiveError!Value {
 
     try growIfNeeded(ht);
     const slot = try findSlot("hash-table-update!", ht, key);
+    // #1924: a shared parent-heap table must not come to hold a child-heap
+    // key or value (see hash-table-set!).
+    if (memory.crossHeapStoreViolation(types.toObject(args[0]), key)) return primitives.raiseCrossHeapStore("hash-table-update!");
+    if (memory.crossHeapStoreViolation(types.toObject(args[0]), new_val)) return primitives.raiseCrossHeapStore("hash-table-update!");
     if (memory.gc_instance) |gc| {
         gc.writeBarrier(types.toObject(args[0]), key);
         gc.writeBarrier(types.toObject(args[0]), new_val);
@@ -885,6 +889,9 @@ fn hashTableUpdateDefaultFn(args: []const Value) PrimitiveError!Value {
 
     try growIfNeeded(ht);
     const slot = try findSlot("hash-table-update!/default", ht, key);
+    // #1924: see hash-table-update!.
+    if (memory.crossHeapStoreViolation(types.toObject(args[0]), key)) return primitives.raiseCrossHeapStore("hash-table-update!/default");
+    if (memory.crossHeapStoreViolation(types.toObject(args[0]), new_val)) return primitives.raiseCrossHeapStore("hash-table-update!/default");
     if (memory.gc_instance) |gc| {
         gc.writeBarrier(types.toObject(args[0]), key);
         gc.writeBarrier(types.toObject(args[0]), new_val);
@@ -1048,6 +1055,11 @@ fn hashTableMergeFn(args: []const Value) PrimitiveError!Value {
     }
 
     for (snapshot) |entry| {
+        // #1924: a shared parent-heap destination table must not come to
+        // hold keys/values from another heap (see hash-table-set!); checked
+        // per entry before any store.
+        if (memory.crossHeapStoreViolation(types.toObject(args[0]), entry.key)) return primitives.raiseCrossHeapStore("hash-table-merge!");
+        if (memory.crossHeapStoreViolation(types.toObject(args[0]), entry.value)) return primitives.raiseCrossHeapStore("hash-table-merge!");
         const slot = try findSlot("hash-table-merge!", ht1, entry.key);
         gc.writeBarrier(types.toObject(args[0]), entry.key);
         gc.writeBarrier(types.toObject(args[0]), entry.value);
--- a/src/primitives_vector.zig
+++ b/src/primitives_vector.zig
@@ -245,8 +245,14 @@ fn vectorCopyBangFn(args: []const Value) PrimitiveError!Value {
     const count = end - start;
     if (at + count > to_vec.data.len) return primitives.typeError("vector-copy!", "valid index range", args[1]);
 
+    // #1924: a shared parent-heap destination must not come to hold a
+    // pointer from this child's heap (or need the owner's remembered-set
+    // barrier for a young value) — reject before any element is copied.
+    // Only pointer elements trigger the rejection; an all-immediate copy
+    // installs no pointers and is fine.
     if (memory.gc_instance) |gc| {
         for (from_vec.data[start..end]) |val| {
+            if (memory.crossHeapStoreViolation(types.toObject(args[0]), val)) return primitives.raiseCrossHeapStore("vector-copy!");
             gc.writeBarrier(types.toObject(args[0]), val);
         }
     }
@@ -1003,8 +1009,11 @@ fn vectorReverseCopyBangFn(args: []const Value) PrimitiveError!Value {
     const count = end - start;
     if (at + count > to_vec.data.len) return primitives.typeError("vector-reverse-copy!", "valid index range", args[1]);
 
+    // #1924: see vector-copy! — same per-element rejection for a shared
+    // destination.
     if (memory.gc_instance) |gc| {
         for (from_vec.data[start..end]) |val| {
+            if (memory.crossHeapStoreViolation(types.toObject(args[0]), val)) return primitives.raiseCrossHeapStore("vector-reverse-copy!");
             gc.writeBarrier(types.toObject(args[0]), val);
         }
     }
--- a/src/primitives_vector.zig
+++ b/src/primitives_vector.zig
@@ -123,6 +123,7 @@ fn vectorSetFn(args: []const Value) PrimitiveError!Value {
     const k = types.toFixnum(args[1]);
     // u64 comparison before narrowing (kaappi#1912): see fixnumIndexInBounds.
     if (!primitives.fixnumIndexInBounds(k, vec.data.len)) return primitives.indexError("vector-set!", k, vec.data.len);
+    if (memory.crossHeapStoreViolation(types.toObject(args[0]), args[2])) return primitives.raiseCrossHeapStore("vector-set!");
     if (memory.gc_instance) |gc| gc.writeBarrier(types.toObject(args[0]), args[2]);
     vec.data[@intCast(k)] = args[2];
     return types.VOID;
@@ -193,6 +194,7 @@ fn vectorFillFn(args: []const Value) PrimitiveError!Value {
     const range = try primitives.parseOptionalRange(args, 2, len, "vector-fill!");
     const start = range.start;
     const end = range.end;
+    if (memory.crossHeapStoreViolation(types.toObject(args[0]), args[1])) return primitives.raiseCrossHeapStore("vector-fill!");
     if (memory.gc_instance) |gc| gc.writeBarrier(types.toObject(args[0]), args[1]);
     @memset(vec.data[start..end], args[1]);
     return types.VOID;
--- a/src/primitives_vector.zig
+++ b/src/primitives_vector.zig
@@ -979,6 +979,9 @@ fn vectorMapBangFn(args: []const Value) PrimitiveError!Value {
             call_args[vi] = types.toVector(args[1 + vi]).data[i];
         }
         const result = try callVM(f, call_args);
+        // #1924: a shared parent-heap destination must not come to hold a
+        // pointer from this child's heap (see vector-copy!).
+        if (memory.crossHeapStoreViolation(types.toObject(args[1]), result)) return primitives.raiseCrossHeapStore("vector-map!");
         gc.writeBarrier(types.toObject(args[1]), result);
         target.data[i] = result;
     }
@@ -1071,6 +1074,11 @@ fn vectorUnfoldBangFn(args: []const Value) PrimitiveError!Value {
         if (types.isMultipleValues(result)) {
             const mv = types.toObject(result).as(types.MultipleValues);
             if (mv.values.len == 0) return primitives.typeError("vector-unfold!", "at least one return value from step procedure", result);
+            // #1924: a shared parent-heap destination must not come to hold a
+            // pointer from this child's heap (see vector-copy!).
+            if (memory.crossHeapStoreViolation(types.toObject(args[1]), mv.values[0])) return primitives.raiseCrossHeapStore("vector-unfold!");
+            // #1924: see vector-unfold!.
+            if (memory.crossHeapStoreViolation(types.toObject(args[1]), mv.values[0])) return primitives.raiseCrossHeapStore("vector-unfold-right!");
             gc.writeBarrier(types.toObject(args[1]), mv.values[0]);
             vec.data[i] = mv.values[0];
             for (0..seeds.items.len) |j| {
@@ -1079,6 +1087,8 @@ fn vectorUnfoldBangFn(args: []const Value) PrimitiveError!Value {
                 }
             }
         } else {
+            if (memory.crossHeapStoreViolation(types.toObject(args[1]), result)) return primitives.raiseCrossHeapStore("vector-unfold!");
+            if (memory.crossHeapStoreViolation(types.toObject(args[1]), result)) return primitives.raiseCrossHeapStore("vector-unfold-right!");
             gc.writeBarrier(types.toObject(args[1]), result);
             vec.data[i] = result;
         }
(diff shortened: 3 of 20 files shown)


baijum (commented):
## Review — cross-thread heap safety (#1924 / #1933)

Solid, well-documented work, and the two mechanisms are the right shape: a per-*store* (not per-type) rejection for #1924, and a stop-the-world child-root mark for #1933. The tests are properly discriminating rather than vacuous (`refused?` matches the specific hazard string, the negative controls assert concrete values, and the #1933 test carries a guardian control proving the churn actually collects) — good, given how easy it is for a thread test to pass on a truthy failure sentinel.

Two things I'd want addressed before merge — one concrete, one a design question.

---

### 1. `crossHeapStoreViolation` is missing at several general mutation sites (same #1924 UAF)

Both the PR description and the commit message say the check runs *"at every general mutation site"* and enumerate them. But the guard was added per-primitive, and at least **six** Value-storing mutators that a child can reach through the globals route have **no** check. Each installs a child-heap pointer into a possibly-shared parent-heap container — the exact deterministic UAF #1924 is about — and none is covered by the new tests:

| Primitive | File / line | Store |
|---|---|---|
| `list-set!` | `src/primitives_list.zig:81` | `setCar(current, args[2])` into a (possibly shared) pair |
| `vector-copy!` | `src/primitives_vector.zig:254,260` | copies `from_vec` elements into `to_vec` (`args[0]`) |
| `vector-reverse-copy!` | `src/primitives_vector.zig:1010` | same, reversed |
| `hash-table-update!` | `src/primitives_hashtable.zig:856,858` | stores computed `new_val` into `args[0]` |
| `hash-table-update!/default` | `src/primitives_hashtable.zig:889,891` | same |
| `hash-table-merge!` / `-union!` | `src/primitives_hashtable.zig:1051,1055` | stores `ht2`'s entries into `ht1` (`args[0]`) |

The asymmetry is stark within a single file: `hash-table-set!` gets the guard (`primitives_hashtable.zig`), but `hash-table-update!`, `hash-table-update!/default`, and `hash-table-merge!` — which write into the same container the same way — do not. A child that reads a shared table from a global and calls `(hash-table-update! shared 'k (lambda (_) (list 1 2 3)))` reproduces #1924 just as `hash-table-set!` did; likewise `(vector-copy! shared-vec 0 child-vec)` and `(list-set! shared-list 0 (list 1 2 3))`.

`alistToHashTableFn` (`primitives_hashtable.zig:803`) is *not* a gap — its container is freshly allocated on the running thread — so the fix is genuinely per-site, which is exactly why the enumeration needs to be exhaustive.

Suggest: add `crossHeapStoreViolation` (→ `raiseCrossHeapStore`) at each of the six sites, and extend `srfi18-cross-heap-mutation-1924.scm`'s rejection matrix to cover them so the "every mutation site" claim is test-enforced rather than asserted. It'd also be worth a comment on *why* the check lives at each primitive rather than one central choke point, since the recurring failure mode here is a newly-added mutator silently reopening the hole.

---

### 2. Are `.parked` / `.in_native` children actually pinned through the parent's mark? (please confirm)

`markLiveChildRoots` waits for each child to leave `.running`, then — after the wait loop finishes for *all* children — calls `markVmRoots(gc, child_vm)`. Only a child that stops at the **dispatch-loop safepoint** (`stopForCollection`) is truly pinned: it sets `.stopped` and spins on `collection_stop` until the parent clears it, so its frames are frozen across the mark.

A `.parked` child (e.g. a mutex/condvar/channel waiter) is different. In `mutexLockFn` the wait loop calls `runSchedulerStep` → `parkOnReactor` (reports `.parked`, `defer` restores `.running`) and then `sleepNs(CROSS_THREAD_POLL_NS)` — so a contended waiter cycles `.parked` → `.running` roughly every ms. The parent's wait (`while (state == .running) sleep`) can observe `.parked` and move on, but nothing re-checks or re-pins the child immediately before `markVmRoots`. Between the parent observing non-`.running` and actually marking, a woken child runs scheduler/wake code as `.running` and can push/pop frames (`frame_count`), so the parent may read `frames[0..frame_count]` while the child mutates it. The same window exists for `.in_native` on FFI return. It's re-caught at the *next* safepoint, but the racy mark has already happened by then.

This is hard to hit and may well be why gc-stress stays green (the srfi18 suite doesn't obviously exercise a mutex-waiting child concurrent with heavy parent churn), so I'm flagging it as a question rather than a confirmed bug: is the intent that `.parked`/`.in_native` children re-verify `collection_stop` and stay quiescent until released (like the safepoint path), or is reading a momentarily-`.running` child's frames considered safe here? If the former, the parked/native sites may need to spin on `collection_stop` before returning to bytecode, mirroring `stopForCollection`.

---

### Minor

- `CollectionState`'s doc comment says `.parked` means *"no bytecode runs until it wakes"* — the *"until it wakes"* is exactly the window in (2); worth tightening once (2) is resolved.
- `%promise-merge!` guards the `outer.value = inner.value` store but not the back-pointer `inner.value = makePointer(&outer.header)`. Almost certainly fine (both are promises in a force chain), but a one-line note on why the reverse direction can't be a cross-heap store would save the next reader the trace.

src/primitives_vector.zig
```diff
@@ -123,6 +123,7 @@ fn vectorSetFn(args: []const Value) PrimitiveError!Value {
     const k = types.toFixnum(args[1]);
     // u64 comparison before narrowing (kaappi#1912): see fixnumIndexInBounds.
     if (!primitives.fixnumIndexInBounds(k, vec.data.len)) return primitives.indexError("vector-set!", k, vec.data.len);
+    if (memory.crossHeapStoreViolation(types.toObject(args[0]), args[2])) return primitives.raiseCrossHeapStore("vector-set!");
```
baijum, line 126: `vector-set!` is guarded here — but `vector-copy!` (`vectorCopyBangFn`, ~L246) and `vector-reverse-copy!` (~L1010) copy `from_vec` elements into `args[0]` with only a write barrier, no `crossHeapStoreViolation` check. A child doing `(vector-copy! shared-vec 0 child-vec)` installs child-heap pointers into a shared vector — the same #1924 UAF. Both need the guard, plus a rejection-matrix row.

src/primitives_hashtable.zig
```diff
@@ -653,6 +653,10 @@ fn hashTableSetFn(args: []const Value) PrimitiveError!Value {
     const ht = try getHashTable("hash-table-set!", args[0]);
     try growIfNeeded(ht);
     const slot = try findSlot("hash-table-set!", ht, args[1]);
+    // #1924: reject before the store — a shared parent-heap table must not
+    // come to hold a child-heap key or value.
```
baijum, line 657: `hash-table-set!` is guarded here, but three sibling mutators in this file are not: `hash-table-update!` (~L856), `hash-table-update!/default` (~L889), and `hash-table-merge!`/`-union!` (~L1051) all store a value (computed `new_val`, or `ht2`'s entries) into the same shared container `args[0]` with only a write barrier. `(hash-table-update! shared 'k (lambda (_) (list 1 2 3)))` from a child reproduces #1924. The commit says the check runs at 'every general mutation site' — these need it too, plus rejection-matrix rows. (`alistToHashTableFn` at ~L803 is fine — fresh container.)

src/fiber_wait.zig
```diff
@@ -231,6 +231,16 @@ pub fn parkOnReactor(vm: *VM, sched: *FiberScheduler, cap_ns: ?u64) VMError!bool
...
     defer ready.deinit(vm.gc.allocator);
+    // #1933: a child OS thread blocked in the reactor poll is quiescent; a
+    // collecting parent may mark its registers/frames without waiting for a
+    // safepoint. The main thread (owns_globals) never reports a state — it is
+    // the collector's own thread. The defer must be at FUNCTION scope: a
+    // defer inside the if-block would fire when the block exits — before the
+    // poll — leaving the child reported .running (and unmarkable) for the
+    // whole blocking wait.
+    const report_park_state = !vm.owns_globals;
+    if (report_park_state) vm.setCollectionParked();
```
baijum, line 243: TOCTOU question about the `.parked` state reported here. `markLiveChildRoots` waits for a child to leave `.running`, then marks it later. Only a `.stopped` child (safepoint) is truly pinned — it spins on `collection_stop`. A `.parked` mutex/condvar/channel waiter cycles `.parked`→`.running` ~every ms (`parkOnReactor` `defer` restores `.running`, then `mutexLockFn` runs scheduler code + `sleepNs` before re-parking). Nothing re-checks the child right before `markVmRoots`, so between the parent observing non-`.running` and marking, a woken child can run wake code as `.running` and change `frame_count` — the parent then reads `frames[0..frame_count]` concurrently. Re-caught at the next safepoint, but the racy mark already happened (same for `.in_native` on FFI return). Intended safe, or should the park/native sites spin on `collection_stop` before resuming like `stopForCollection`? Detail in the top-level review.

baijum:
Addressed all points in 8c69527c. Summary:

**1. Missing mutation sites** — added `crossHeapStoreViolation` at all six: `list-set!`, `vector-copy!`, `vector-reverse-copy!`, `hash-table-update!`, `hash-table-update!/default`, `hash-table-merge!`. The rejection matrix in `srfi18-cross-heap-mutation-1924.scm` now covers them (plus `set-cdr!`, a child-allocated hash key, `define` via eval, and the parent-value-into-parent-container case). `alistToHashTableFn` confirmed not a gap (fresh container).

**2. `.parked` / `.in_native` pinning (the TOCTOU)** — you were right, and the review also surfaced the actual CI failure. `setCollectionRunning` is now guarded: resuming from a quiescent state re-checks `collection_stop` and, if armed, publishes `.stopped` and spins until the parent clears it — mirroring `stopForCollection` — so a parent that observed `.parked`/`.in_native` can always mark the child frozen. `callWithArgs`' FFI-callback guard and the `threadEntryFn` startup handshake route through it. The `.parked` doc wording is tightened accordingly, and `%promise-merge!`'s back-pointer note added.

**3. The gc-stress CI hang** (bonus, found while validating 2): three tests timed out. Two were the stop-the-world itself: a child joining its own child blocks in a raw pthread join that never reaches a safepoint (parent waited forever, grandchild spun on `collection_in_progress` — a livelock), and the 1ms wait poll made every parent collection with a live child cost ~1ms. `reapOsThread`'s `thread.join()` now reports the in-native quiescent state (function-scope defer), and the wait spin-yields. `channel-promoted-owner-1934` went 0.15s → >40min → 1s. The third (`srfi146-differential`) is pre-existing — 3m on base, no threads, unaffected by this PR.

**Also** (from CodeRabbit + self-review): the predicate now rejects *all* foreign-container stores (a parent-owned value into a parent-owned container needs the owner's generational barrier for a young value — the owner's remembered set can't be touched cross-thread, with interned symbols kept allowed (permanent); exited children's  is always released;  is atomic. The audit characterisation tests that asserted the old store-anything behavior are updated.
EOF
)

baijum:
Re-reviewed `8c69527c` — all three points resolved, and CI (incl. `gc-stress` / `gc-stress-scheme`, the previous timeouts) is green. LGTM.

**1. Mutation sites** — all six now guarded (`list-set!`, `vector-copy!`, `vector-reverse-copy!`, `hash-table-update!`, `hash-table-update!/default`, `hash-table-merge!`), with matching rejection-matrix rows. Confirmed `alistToHashTableFn` is correctly left alone (fresh container).

**2. Broadened predicate** — this is the right call, and sharper than what I'd flagged. I only reasoned about dangling pointers (foreign-heap value); you caught the second hazard I'd missed — a value owned by the container's *own* foreign heap, young, stored into an old foreign container, still needs the **owner's** remembered-set edge, and the writing thread's `writeBarrier` records into the wrong gc's set. Rejecting every foreign-container store is the sound rule. The interned-symbol exception is safe (permanent + promoted, no old→young edge) and keeps symbol-keyed writes working.

**3. TOCTOU** — `setCollectionRunning` re-checking `collection_stop` and re-parking as `.stopped` closes it. I traced the one residual window: `parkOnReactor` runs `sweepSharedWaiters`/`wakeReadyFiber`/`markRunnable` after `poll()` while still reporting `.parked` (before the defer). That only mutates scheduler queues and fiber status — not `vm.frames`/registers/`frame_count`, which is all `markVmRoots` reads — so a concurrent parent mark races nothing there. Agreed it's benign.

Nice bonus catches on the way: the self-join livelock (`reapOsThread` now reporting `.in_native`), the 1ms-per-collection cost (spin-yield), and the separate `armed` list so an exited child's `collection_stop` is always released rather than stranded.

One caveat on my sign-off: this is a read-through plus CI, not a local `-Dgc-stress` run on my end — the green `gc-stress-scheme` job is the real evidence for the concurrency paths.

baijum:
Addressed the still-valid findings in 83ac8b9a (the rest of the 8 were already handled in 8c69527c):

- **`.parked` window too wide (the critical one)** — correct: the post-poll code (`sweepSharedWaiters`, `wakeReadyFiber`, `markRunnable`) mutates scheduler state that `markVmRoots` traverses via `sched.markRoots`. `parkOnReactor` now reports `.parked` across the blocking `reactor.poll` only, and resumes through the guarded `setCollectionRunning` (which honours `collection_stop`) immediately after.
- **vector-map!/vector-unfold!/vector-unfold-right!** — these write the step procedure's results into a caller-supplied destination; a shared parent-heap destination now gets the same per-result cross-heap store rejection as `vector-copy!` (all-immediate runs stay legal). Added to the rejection matrix.
- **define_global on non-globals envs** — the child-store rejection now runs before the env branch (matching `set_global`), covering a child eval-define into a shared library/eval environment, not just the globals map. (R7RS `(environment ...)` is immutable, so the reachable case is the globals map, covered by the eval-define test row.)
- **test churn** — the child-registers-1933 churn is reduced 4x (100k iterations; verified against the pre-fix engine that the freed slot is still recycled and the read turns to garbage), bringing the gc-stress run from ~160s to ~40s.
- **doc row** — reworded to match the strengthened predicate (any store into a not-owned container is rejected; interned symbols excepted).

Already addressed in 8c69527c: the owner-aware-barrier concern (resolved by rejecting all foreign-container stores — the owner's remembered set cannot be touched cross-thread, so no store into a foreign container is allowed at all); the armed-list release of `collection_stop`; atomic `child_marker`; the guarded `setCollectionRunning`/`callWithArgs`/startup transitions; `gc_instance = null` test cleanup; and the set-cdr!/hash-key/define test rows.

CI re-running; gc-stress-scheme was green on the previous round.

baijum:
Correction to my previous LGTM — I was wrong about the `.parked` post-poll window, and `83ac8b9a` fixes a real race, not a cosmetic one.

I claimed the post-poll code "only mutates scheduler queues … not `vm.frames`/registers, which is **all** `markVmRoots` reads." That "all" was false: `markVmRoots` also calls `sched.markRoots(gc)` and `reactor.markRoots(gc)` (vm.zig:77–82) — which traverse exactly the scheduler state `sweepSharedWaiters`/`wakeReadyFiber`/`markRunnable` mutate. So a parent marking a child reported `.parked` while that post-poll code ran was a genuine data race on the scheduler roots. My error was mechanical and the kind I should've caught: I read `markVmRoots` with a 60-line cap and the `sched.markRoots` call sits at line ~77, just past where I stopped. I then asserted "benign" from the truncated view instead of reading the whole function. Apologies — that's exactly the sort of overconfident dismissal a review shouldn't ship.

The `83ac8b9a` fix is correct: reporting `.parked` across `reactor.poll` only, then resuming through the guarded `setCollectionRunning()` *before* the scheduler mutations, means the child is `.running` (or re-stopped at the safepoint if `collection_stop` is armed) during `sweepSharedWaiters`/`wakeReadyFiber`/`markRunnable`, so the parent never marks it mid-mutation.

The other two round-3 changes also check out:
- **`vector-map!` / `vector-unfold!` / `vector-unfold-right!`** — correct to guard these; they write step results into a caller-supplied destination, same shape as `vector-copy!`. These were extra mutation sites beyond the six I listed — I missed them in my sweep, good that the second CodeRabbit pass caught them. Per-result check keeps all-immediate runs legal.
- **`define_global`** — moving the child-store rejection ahead of the `env == self.globals` branch (matching `set_global`) correctly extends it to a child eval-define into a shared library/eval env, not just the globals map.

CI is green including `gc-stress` (5m) and `gc-stress-scheme` (20m).

One residual I'll flag *tentatively* (low confidence, having just been wrong once): during the poll-only `.parked` window, `reactor.poll` can still pop the timer heap while the parent runs `reactor.markRoots` over it. It looks practically harmless — the child's reactor holds child-heap Values, which the parent's `markValue` skips as foreign-owner, so no bad mark results; at worst it's a racy read of the heap array with a no-op outcome. Worth a glance to confirm `reactor.markRoots` can't index out of bounds if the heap length shrinks mid-iteration, but I don't think it's blocking. LGTM stands on `83ac8b9a`.

baijum:
Agreed on both counts — thanks for the correction and the careful re-check.

On the timer-heap residual you flagged: `reactor.markRoots` iterates a snapshot of `self.timers.items` (Zig `for (slice)` captures ptr+len once), and the child's `reactor.poll` pops by decrementing `len` without reallocating or freeing the backing array — so the parent's snapshot never indexes freed memory, and the worst case is reading a stale popped entry whose fiber is still alive (it is in the scheduler's `fibers` array, so the child's own GC cannot have freed it, and the parent's `markValue` skips it as foreign-owner). No OOB, no freed-header read. Non-blocking confirmed.

Also pushed f9c5dfe2 for the one remaining CodeRabbit item: `%promise-merge!` now validates the inner back-pointer store (`inner.value = outer`) through the same `crossHeapStoreViolation` as the outer store — the comment claimed the predicate covered that direction but the check was never actually called there. Both directions are now checked before any store or barrier.

CI re-running on f9c5dfe2.

baijum:
`f9c5dfe2` resolves the last open item from my original review (the `%promise-merge!` back-pointer), and the fix is correct: both `crossHeapStoreViolation` checks now run **before** either store or barrier, so a rejection can't leave the promise pair half-mutated, and `outer_val` is computed once and reused for the check and the store. The real value here is that the code now matches its comment — the previous comment asserted the container-owner rule "covered" the inner direction, but that check was never actually wired, so the claimed coverage was fictional. Good to close that gap even though, as you note, both promises share a heap in practice so it rarely fires.

With the broadened predicate this second check also correctly rejects *any* child merge into a shared parent-heap inner promise regardless of `outer`'s owner — consistent with the "no store into a foreign container" rule.

That clears everything I raised across the four rounds. CI is still running on this commit; given it's a two-line addition inside one primitive and untouched concurrency paths, I don't expect surprises, but the gc-stress jobs are the arbiter as before. LGTM.

Issue #1812, opened 2026-07-28
Hygiene: a use-site top-level redefinition of a name a macro's expansion references reaches the expansion (syntax-rules and ER alike)


baijum:
Found while verifying SRFI 211's ER macros had hygiene parity with syntax-rules (PR #1811). Both macro kinds share this pre-existing hole; it is engine-wide, not new.

Reproduction (any library exporting a macro whose expansion references a library helper):

```scheme
;; lib: (demo srmac) — exports tw; helper2 is library-internal
;;   (define (helper2 x) (* x 2))
;;   (define-syntax tw (syntax-rules () ((_ e) (helper2 e))))

(import (scheme base) (scheme write) (demo srmac))
(display (tw 21))          ; 42 — correct
(define helper2 'shadowed) ; user redefines the helper's NAME at top level
(display (tw 5))           ; KP3005 "not a procedure" — expansion now sees 'shadowed
```

R7RS 4.3.1 referential transparency says the template's `helper2` should keep denoting the macro-definition-site binding. The engine's existing protection (the `bound_free_refs` VOID-sentinel dance + `injectHygienicCapturedLocals` in `expandAndCompileMacroUse`) covers non-procedure globals and definition-site locals, but a *procedure* global referenced by name compiles to a by-name `call_global`/`get_global`, so whatever that name resolves to at run time wins — including a later same-file top-level redefinition.

Notes for whoever picks this up:
- ER macros (`expander.expandProceduralMacro`) inherit this through `renameForHygiene`'s keep-global-procedure-names-unrenamed branch — fixing it for syntax-rules fixes both.
- The import-time value copy (`vm_library.copyTransformerFreeRefs` / `copyOneDefEnvBinding`) already has the def-site *values* in hand; a fix might bake those values (constant or a def-site alias) instead of emitting a by-name global reference, but the tradeoff with intentional late rebinding of globals (e.g. test seams that redefine procedures) needs thought — plain top-level programs redefine imported names legally.
- Debugging trap from the discovery session: `.sbc`-cached library loads currently put library-internal defines into `vm.globals` while fresh loads keep them lib_env-only, so this reproduction can appear/disappear across otherwise-identical runs. Run `kaappi cache clear` between experiments. That cached-vs-fresh globals divergence may deserve its own look.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2074, opened 2026-08-01
[R7RS 4.3.2] A use-site local named after a syntactic keyword captures it inside any macro template — 15 of 17 keywords, incl. begin/lambda/quote/letrec


baijum:
**Category:** R7RS macro hygiene
**Spec reference:** R7RS-small §4.3.2 (errata-corrected):

> "If a macro transformer inserts a free reference to an identifier, the reference refers to the binding that was visible where the transformer was specified, regardless of any local bindings that surround the use of the macro."

**Severity:** wrong-result / spurious error (silent at the definition site; the failure surfaces at an unrelated use site)

A **local variable at the use site whose name collides with a syntactic keyword** captures that keyword inside any macro template expanded there. 15 of the 17 keywords probed are affected. This is a *different* mechanism from #2003 — that one is about a template's free reference to a global holding a **procedure**; this one is about special forms, which are not globals at all, and it has its own source location and its own discriminator (see below).

## Minimal reproduction

```scheme
(import (scheme base) (scheme write))
(define-syntax m (syntax-rules () ((_ e) (begin e))))
(display (let ((begin 5)) (m 7)))
;; expected 7
;; actual   error[KP3005]: not a procedure     <- the template's `begin` became a call to the local
```

The template's `(begin e)` is compiled as the *procedure call* `(5 7)`. Same with `letrec`:

```scheme
(define-syntax m (syntax-rules () ((_ e) (letrec ((t e)) t))))
(let ((letrec 5)) (m 7))
;; error[KP3001]: undefined variable '__hyg_2_t'
```

That error message is itself the proof: the template's *binding* `t` **was** hygienically renamed to `__hyg_2_t`, so the renamer ran — it simply does not rename the free keyword `letrec`, so `letrec` is left as a bare symbol that the use-site local then shadows, and the `letrec` body's own binding never gets created.

## Full sweep (Kaappi vs. Chibi 0.12.0)

Each row shadows the keyword with `(let ((<kw> 5)) ...)` and calls a macro whose template uses that keyword; the correct answer is `7` everywhere.

| keyword | Kaappi | Chibi | | keyword | Kaappi | Chibi |
|---|---|---|---|---|---|---|
| `let` | 7 ✅ | 7 | | `and` | **raises** ❌ | 7 |
| `if` | 7 ✅ | 7 | | `or` | **raises** ❌ | 7 |
| `letrec` | **raises** ❌ | 7 | | `do` | **raises** ❌ | 7 |
[...]

PR #2251, opened 2026-08-08
Fix macro hygiene: use-site bindings can no longer capture template free references (#2003, #2074)

### LINKED ISSUE #2003: [R7RS 4.3.2] A local binding at the use site captures a macro template's free reference to a global procedure — (let ((car ...)) ...) hijacks car inside any macro
**Category:** R7RS macro hygiene (`syntax-rules` and SRFI 211 ER alike)
**Spec reference:** R7RS 4.3.2 (Pattern language), errata-corrected text:

> "If a macro transformer inserts a free reference to an identifier, the reference refers to the binding that was visible where the transformer was specified, regardless of any local bindings that surround the use of the macro."

**Severity:** wrong-result (silent, wrong answer; no diagnostic)

## Minimal reproduction

```scheme
(define (helper x) (* x 10))
(define-syntax use (syntax-rules () ((_ n) (helper n))))
(write (let ((helper (lambda (x) (- x)))) (use 3)))   ; expected 30, actual -3
```

The use site's *ordinary local binding* — not a redefinition, just a `let` — captures the template's free reference. Builtins are captured the same way, including R7RS 4.3.2's own canonical example:

```scheme
(define-syntax usecar (syntax-rules () ((_ l) (car l))))
(let ((car (lambda (x) 'HIJACKED))) (usecar (list 1 2)))   ; expected 1, actual HIJACKED

(define-syntax add (syntax-rules () ((_ a b) (+ a b))))
(let ((+ (lambda (a b) (* a b)))) (add 3 4))               ; expected 7, actual 12
```

Every binding form does it — `let`, a `lambda` parameter, an internal `define`, named `let`, `do` — and nesting depth makes no difference.

## Discriminating controls

The failure is confined to a referenced global that holds a **procedure**. Three near-identical programs separate the cases:

| | Kaappi | Chibi 0.11 | Guile 3.0 `--r7rs` |
|---|---|---|---|
| (a) def-site **local** variable (R7RS 4.3's own `let-syntax` example) | `outer` ✅ | `outer` | `outer` |
| (b) global **non-procedure** (`(define gv 'global-value)`) | `global-value` ✅ | `global-value` | `global-value` |
| (c) global **procedure** (`(define (gp) 'global-proc)`) | **`shadowed`** ❌ | `global-proc` | `global-proc` |
| (d) builtin `car` | **`HIJACKED`** ❌ | `1` | `1` |
| (e) builtin `+` | **`12`** ❌ | `7` | `7` |
| (f) shadow of an **unrelated** name | correct ✅ | correct | correct |

(a) and (b) passing is what makes this a *targeted* defect rather than "hygiene is off": the machinery works, and stops working exactly where the referenced global is a procedure. (f) rules out a blanket expansion-in-the-wrong-scope explanation.

## Relationship to closed #1812

#1812 is the **top-level-redefinition** half of this and its notes already name the mechanism:

> "a *procedure* global referenced by name compiles to a by-name `call_global`/`get_global`, so whatever that name resolves to at run time wins"
> "ER macros (`expander.expandProceduralMacro`) inherit this through `renameForHygiene`'s keep-global-procedure-names-unrenamed branch — fixing it for syntax-rules fixes both."

This issue is the **local-scope** half, and it is materially worse: #1812 needs the user to redefine a top-level name after importing the macro, which is at least a deliberate act. This one needs only an ordinary `let` naming a variable `car`, `+`, `list`, `min`, `length` — legal, common, and something no caller can be expected to avoid, since a macro's template contents are not part of its interface. It also fires at *compile* time rather than run time, so the unrenamed name resolves against the enclosing lexical scope, not just against whatever the global later holds.

ER macros behave identically, so the parity claim in `lib/srfi/211/explicit-renaming.sld`'s header ("Hygiene strength equals this engine's own syntax-rules hygiene exactly") is **accurate** — it is the shared strength that is wrong. `docs/dev/`/`README.md` document neither half: `README.md § Known limitations § Macros` says only "Only `syntax-rules` is supported."

## Source location

`src/expander.zig` — `renameForHygiene`'s branch that leaves a name bound to a global procedure unrenamed. Not `#1812`'s suggested import-time value copy: that fix targets the *global* binding, and here the capturing binding is lexical, so a fix has to make the emitted reference itself scope-proof (a hygienic alias, or a reference form that skips local scopes).

**Verified:** fresh ReleaseSafe build at `e6081c3b`, isolated `KAAPPI_HOME`, cache cleared, reproduced twice; identical under `--no-ir-opt`, so it is not an IR-optimisation artefact. Cross-checked against `chibi-scheme` and `guile --r7rs` on the byte-identical file.

**Disabled test:** `tests/scheme/audit/primitives_smallbatch-audit.scm`, section "hygiene (e)", marked `;; FAIL:`. Its neighbours (a)–(d) and (f) are enabled and pass; a PARITY assertion asserting syntax-rules and ER agree under the shadow is enabled too, so a fix that lands on only one path fails there.

**Found by:** Systematic audit v2, Phase 2.11

baijum:
Fixes #2003 and #2074 — the two sibling R7RS 4.3.2 macro-hygiene defects from audit v2 (Phase 2.11 / 3.7). Both let an ordinary use-site local binding capture part of a macro template; both are marked `priority: high`, milestone 0.23.

## #2003 — use-site local captures a template's free reference to a global procedure

```scheme
(define (helper x) (* x 10))
(define-syntax use (syntax-rules () ((_ n) (helper n))))
(write (let ((helper (lambda (x) (- x)))) (use 3)))   ; was -3, now 30
(define-syntax usecar (syntax-rules () ((_ l) (car l))))
(write (let ((car (lambda (x) 'HIJACKED))) (usecar (list 1 2))))  ; was HIJACKED, now 1
```

The template's free reference to a global *procedure* was left unrenamed, so it compiled through the use site's lexical scope. It is now hygiene-renamed like any other template-introduced identifier; the run-time global lookup's existing hygienic-prefix fallback resolves the rename to the base global by name — immune to use-site locals, while still observing a same-environment top-level redefinition (the semantics chibi and guile implement). Companion changes: `isContinuationBarrier` and the four legacy tail-position fast paths (`apply` / `call-with-values` / `call/cc` / `eval`) now recognize a renamed spelling — SRFI 248's guard re-raise needs the `call-with-values` one for correct multiple-value passing (the exact case #1812's notes warned about).

**Deliberate exception:** a template *lambda formal* colliding with a global procedure keeps its bare spelling (identity rename) — SRFI 190's coroutine body binds to the template's `yield` formal by name. This is *not* extended to let variables (#681 pins the opposite).

## #2074 — use-site local named after a syntactic keyword captures it in templates

```scheme
(define-syntax m (syntax-rules () ((_ e) (begin e))))
(display (let ((begin 5)) (m 7)))   ; was "not a procedure", now 7
```

15 of 17 swept operator keywords (`let`/`if` were already renamed) were inserted bare, so a use-site local of the same spelling captured them. The operator keywords among the well-known forms are now hygiene-renamed; the compiler already recognized renamed special forms via effective-name stripping. Kept bare: the definition/library forms, `syntax-rules`, the aux `else`, the pattern markers `...`/`_`, and the quote/quasiquote *value* symbols. `quote`/`quasiquote` *form* heads are renamed too, with strip-aware quasiquote depth handling in `compileQQ` (and hygiene-stripped rebuilt heads). `=>` in cond/case clauses is renamed and recognized through the strip; cond-expand feature combinators likewise.

## Tests

- Re-enabled the previously disabled audit `hygiene (e)` tests and the `srfi-notest-batch` #2003/#2074 tests.
- Updated two SRFI 190 tests that pinned the old anaphoric capture through *another* macro's template — under correct hygiene a free `yield` there resolves at its own definition site (chibi and the SRFI reference implementation's syntax-parameter design agree); the direct-anaphora tests still pass.
- Expand snapshot tests now compare through a gensym-id normalizer (the exact `__hyg_N_` id is a process-global counter).
- Full verification: unit suite + `-Dgc-stress`, R7RS suite (1395), all 694 Scheme test files, and the native `compile` tier — all green.

## Notes

- A library macro's template reference to a *base* binding still resolves by name at the use site (so a use-site top-level redefinition of e.g. `car` reaches it) — the pre-existing, documented #1812 leftover, deliberately out of scope here; chibi/guile bind it to the library's own env. Same-file top-level redefinition stays visible, matching both oracles.
- Template *definition* keywords (`define`, `define-syntax`, `let-syntax`, …) are deliberately kept bare (#2074's sweep does not include them and renaming them would interact with the body scanner / transformer-spec resolver) — a documented residual.

---

## Review follow-ups (commit 7d9970f8)

- **#2252** — the `FORMAL_FLAG` comments claimed the bare-spelling anaphoric exception covers *lambda or case-lambda* formals, but only `lambda` is handled. Per the issue's Option 1 (nothing depends on case-lambda anaphora; renaming case-lambda formals hygienically like let-variables is the more consistent behaviour), the comments now state the lambda-only scope and cite #2252. `tests/scheme/hygiene/template-binds-builtin-name.scm` gains a pair of tests pinning the distinction: a case-lambda formal does *not* capture a spliced body (resolves to the global), a lambda formal keeps its anaphoric spelling.
- **Review comment (types.zig)** — dropped the `__nlet_` alternative in `isContinuationBarrier`: `stripHygienicPrefix` does not strip `__nlet_`, so the branch was dead (named-let loop names are always locals, resolved by the call path before the barrier check is reached).
- **Review FYI (llvm_emit_freevars.zig)** — acknowledged: the native free-var scanners still match `lambda`/`quote`/`define` by bare spelling, but macro-expanded forms are eval'd (not natively emitted) in the native tier, so `__hyg_N_` heads never reach them; left unchanged as a latent-consistency note.

### DIFF
--- a/src/expander_instantiate.zig
+++ b/src/expander_instantiate.zig
@@ -41,7 +41,7 @@ const QUOTE_FLAG: u32 = 0x40000000; // inside (quote ...): substitute, hygiene-r
 const BINDING_FLAG: u32 = 0x20000000; // identifier is in binding position
 const NESTED_SR_FLAG: u32 = 0x10000000; // inside a nested syntax-rules template
 const LET_PAIR_FLAG: u32 = 0x08000000; // template is a single let-binding (var init) pair
-const FORMAL_FLAG: u32 = 0x04000000; // identifier is a lambda/case-lambda formal
+const FORMAL_FLAG: u32 = 0x04000000; // identifier is a lambda formal (anaphoric-binding position)
 // Re-walking a usertext-marker-protected splice (an enclosing expansion's
 // pattern-var value, spliced verbatim into a nested syntax-rules template) in
 // substitute-only mode. Reuses QUOTE_FLAG's "substitute, expand ellipses"
@@ -404,18 +404,22 @@ pub fn instantiateTemplate(gc: *GC, template: Value, bindings: []Binding, intro_
             }
         }
         // Function formals are binding positions too. Marked with
-        // FORMAL_FLAG so renameForHygiene treats a formal colliding with a
-        // global procedure as an anaphoric binding (kept bare — SRFI 190's
-        // coroutine body binds to the template's `yield` formal) rather than
-        // as a free reference (renamed, #2003). Distinct from BINDING_FLAG
-        // (let variables): #681 pins that a template LET variable named
-        // after a built-in must NOT capture use-site text, so let-vars stay
-        // hygiene-renamed; only function formals keep the pre-#2003
-        // anaphoric spelling. The formals are walked through the ordinary
-        // instantiateTemplate path (not a custom walk) so ellipsis in a
-        // formals list — `(lambda (slot ...) ...)`, SRFI 26's cut — still
-        // expands through the regular ellipsis machinery; the flag rides
-        // along on every formal symbol.
+        // FORMAL_FLAG so renameForHygiene treats a lambda formal colliding
+        // with a global procedure as an anaphoric binding (kept bare — SRFI
+        // 190's coroutine body binds to the template's `yield` formal)
+        // rather than as a free reference (renamed, #2003). Distinct from
+        // BINDING_FLAG (let variables): #681 pins that a template LET
+        // variable named after a built-in must NOT capture use-site text, so
+        // let-vars stay hygiene-renamed; only LAMBDA formals keep the
+        // pre-#2003 anaphoric spelling (#2252 — case-lambda formals are NOT
+        // included: nothing depends on case-lambda anaphora, and renaming
+        // them hygienically like let-variables is the more consistent
+        // behaviour; a case-lambda clause's formals go through the ordinary
+        // pair recursion below without the flag). The formals are walked
+        // through the ordinary instantiateTemplate path (not a custom walk)
+        // so ellipsis in a formals list — `(lambda (slot ...) ...)`, SRFI
+        // 26's cut — still expands through the regular ellipsis machinery;
+        // the flag rides along on every formal symbol.
         if (std.mem.eql(u8, form_name, "lambda")) {
             const new_car = try instantiateTemplate(gc, elem, bindings, intro_scope, literals, macro_keyword, globals, macros);
             var car_root = new_car;
@@ -956,16 +960,14 @@ pub fn renameForHygiene(gc: *GC, name: []const u8, scope: u32, globals: ?*std.St
                 // falls through and is hygiene-renamed like any other
                 // template-introduced identifier, so a use-site local of the
                 // same name cannot capture it (R7RS 4.3.2). The one
-                // exception: a LAMBDA/CASE-LAMBDA FORMAL (FORMAL_FLAG)
-                // colliding with a global procedure keeps its bare spelling,
-                // recorded as an identity rename — the deliberate
-                // anaphoric-binding pattern (SRFI 190's coroutine body, whose
-                // `yield` must reach the template's formal by name). This
-                // mirrors the pre-#2003 behaviour for function formals
-                // exactly, and is deliberately NOT extended to let variables:
-                // #681 pins that a template let variable named after a
-                // built-in must not capture use-site text, so let-vars stay
-                // hygiene-renamed.
+                // exception: a LAMBDA FORMAL (FORMAL_FLAG) colliding with a
+                // global procedure keeps its bare spelling, recorded as an
+                // identity rename — the deliberate anaphoric-binding pattern
+                // (SRFI 190's coroutine body, whose `yield` must reach the
+                // template's formal by name). This mirrors the pre-#2003
+                // behaviour for lambda formals exactly, and is deliberately
+                // NOT extended to let variables (#681) or to case-lambda
+                // formals (#2252): those stay hygiene-renamed.
                 const is_formal = (scope & FORMAL_FLAG) != 0;
                 if (is_formal and !in_binding and !scopeTableContains(clean_scope, name)) {
                     if (expander.scope_table_count >= MAX_SCOPE_ENTRIES) return ExpandError.ScopeTableFull;
--- a/src/expander_instantiate.zig
+++ b/src/expander_instantiate.zig
@@ -41,6 +41,7 @@ const QUOTE_FLAG: u32 = 0x40000000; // inside (quote ...): substitute, hygiene-r
 const BINDING_FLAG: u32 = 0x20000000; // identifier is in binding position
 const NESTED_SR_FLAG: u32 = 0x10000000; // inside a nested syntax-rules template
 const LET_PAIR_FLAG: u32 = 0x08000000; // template is a single let-binding (var init) pair
+const FORMAL_FLAG: u32 = 0x04000000; // identifier is a lambda/case-lambda formal
 // Re-walking a usertext-marker-protected splice (an enclosing expansion's
 // pattern-var value, spliced verbatim into a nested syntax-rules template) in
 // substitute-only mode. Reuses QUOTE_FLAG's "substitute, expand ellipses"
@@ -106,8 +107,12 @@ pub fn instantiateTemplate(gc: *GC, template: Value, bindings: []Binding, intro_
             }
         }
 
-        // 3. Well-known form or built-in -- keep as-is
-        if (isWellKnown(name)) {
+        // 3. Well-known form or built-in -- keep as-is. Only the reserved
+        //    subset stays bare (see isTemplateReserved): the operator
+        //    keywords among well_known_forms fall through to the hygienic
+        //    rename below so a use-site local of the same spelling cannot
+        //    capture them (kaappi#2074).
+        if (expander.isTemplateReserved(name)) {
             return template;
         }
 
@@ -208,8 +213,20 @@ pub fn instantiateTemplate(gc: *GC, template: Value, bindings: []Binding, intro_
                 var nq_root = new_quoted;
                 gc.pushRoot(&nq_root);
                 defer gc.popRoot();
+                // The FORM head is hygiene-renamed too (#2074): a use-site
+                // local named `quote` must not capture the template's quote
+                // special form. The compiler recognizes the renamed head via
+                // effective-name stripping and strips renames from the datum
+                // (stripHygieneFromDatum), so a template `(quote x)` still
+                // yields the symbol x; a BARE `quote` used as a VALUE (e.g.
+                // `(list quote)`) stays unrenamed because the symbol walk
+                // keeps reserved_template_forms bare.
+                const new_head = try renameForHygiene(gc, "quote", intro_scope, globals);
+                var head_root = new_head;
+                gc.pushRoot(&head_root);
+                defer gc.popRoot();
                 const tail = try gc.allocPair(nq_root, types.NIL);
-                return gc.allocPair(tmpl_head, tail);
+                return gc.allocPair(head_root, tail);
             }
             // More than 2 elements: `quote` is not forming a quote special
             // form here, just an ordinary (possibly data) value in head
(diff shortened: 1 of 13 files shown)


src/expander_instantiate.zig
```diff
@@ -378,6 +403,38 @@ pub fn instantiateTemplate(gc: *GC, template: Value, bindings: []Binding, intro_
...
+        // as a free reference (renamed, #2003). Distinct from BINDING_FLAG
+        // (let variables): #681 pins that a template LET variable named
+        // after a built-in must NOT capture use-site text, so let-vars stay
+        // hygiene-renamed; only function formals keep the pre-#2003
+        // anaphoric spelling. The formals are walked through the ordinary
+        // instantiateTemplate path (not a custom walk) so ellipsis in a
+        // formals list — `(lambda (slot ...) ...)`, SRFI 26's cut — still
+        // expands through the regular ellipsis machinery; the flag rides
+        // along on every formal symbol.
+        if (std.mem.eql(u8, form_name, "lambda")) {
```
baijum, line 423: The FORMAL_FLAG comments (this block, the field decl at line 44, and the `isProcedure` branch at ~959) say the bare-spelling anaphoric exception applies to a **"lambda/case-lambda formal"**, but only `lambda` is handled here — a `case-lambda` clause's formal falls through the regular pair recursion without `FORMAL_FLAG`, so a case-lambda formal colliding with a global procedure is hygiene-renamed, not kept bare.

Verified against the branch binary:
```scheme
(define-syntax mk (syntax-rules () ((_) (case-lambda ((car) (car)) ((car x) (+ car x))))))
;; expands to: (__hyg_1_case-lambda ((__hyg_2_car) (__hyg_2_car)) ...)
```
whereas the same formal under `lambda` stays bare. It runs correctly either way (binding + reference rename consistently), so this isn't a miscompile — but it diverges from the documented `lambda` exception, and the cross-macro anaphoric pattern the comment describes would not reach a case-lambda formal. Suggest either extending the `FORMAL_FLAG` walk to `case-lambda`, or dropping "case-lambda" from these comments so they match the code.

src/types.zig
```diff
@@ -1084,14 +1084,22 @@ pub fn isContinuationBarrier(name: []const u8) bool {
...
+    // call-with-values would take the call_global fast path, which is
+    // exactly what this gate exists to prevent.
     const n = if (std.mem.startsWith(u8, name, base_binding_prefix))
         name[base_binding_prefix.len..]
     else if (std.mem.startsWith(u8, name, def_env_binding_prefix))
         (if (std.mem.indexOf(u8, name, def_env_binding_sep)) |sep|
             name[sep + def_env_binding_sep.len ..]
         else
             name)
+    else if (std.mem.startsWith(u8, name, "__hyg_") or std.mem.startsWith(u8, name, "__nlet_"))
```
baijum: `stripHygienicPrefix` only strips `__hyg_` and `def_env_binding_prefix` (see its body at ~1030) — it does **not** handle `__nlet_`. So for a `__nlet_`-prefixed name this branch calls `stripHygienicPrefix` which returns the name unchanged, i.e. identical to the final `else`, making `or std.mem.startsWith(u8, name, "__nlet_")` a no-op.

Harmless in practice (a named-let loop name is always a local binding and never reaches this barrier check), but the `__nlet_` alternative is dead and slightly misleading. Either drop it, or teach `stripHygienicPrefix` about `__nlet_` if a renamed named-let head could ever need barrier recognition.

baijum (commented):
Reviewed the hygiene changes closely and exercised them against the built branch binary (interpreter + `kaappi compile` native tier). The two headline fixes and a broad set of edge cases all behave correctly:

- #2003 examples (`use`/`helper`, `usecar`/`car`), #2074 (`begin`), and `call-with-values`/`call/cc`/`error` capture-resistance ✅
- Template `=>` in `cond`/`case` (immune to a use-site `=>`), nested quasiquote, `unquote-splicing`, `do`, `when`, `cond-expand` combinators, and macro-generating macros (transformer refs kept bare) ✅
- Native `compile` tier: macro-introduced capturing lambdas inside a user `let` (`(let ((v k)) (capture v))`) capture correctly across multiple escaping closures ✅

The design (rename operator keywords + reserved-form allowlist + effective-name stripping at the dispatchers, with the lambda-formal anaphoric exception for SRFI 190) is coherent and the comments are excellent. Two small non-blocking points below, plus one FYI.

**FYI (not a bug — verified):** several native-backend scanners in `src/llvm_emit_freevars.zig` (`exprHasCapturingLambda`, `sexprReferencesNames`, `sexprContainsDefine`) still match `lambda`/`quote`/`define` by *bare* spelling, and templates now emit `__hyg_N_lambda`/`__hyg_N_quote` heads (`define` stays reserved, so it's unaffected). This PR patched many analogous bare-name sites (cond/case `=>`, qq depth, cond-expand, `resolveTransformerSpecRec`'s `begin`) but not these. I tried hard to turn the `exprHasCapturingLambda` miss into a native miscompile and could not — the native closure tier captures correctly regardless, and CI is green — so this is only a latent-consistency note in case future hygiene work makes those scanners load-bearing again.

src/expander_instantiate.zig
baijum: Filed as kaappi/kaappi#2252 (priority: low, doc-truth) so this doesn't block the PR — it can be resolved separately, most likely by trimming "case-lambda" from the comments.

baijum:
Review comments addressed in 7d9970f8:

1. **case-lambda / FORMAL_FLAG** (CodeRabbit Major + baijum non-blocking, now #2252): the comments claimed the anaphoric exception covers lambda *or* case-lambda formals, but only `lambda` is handled. Chose #2252's Option 1 — the behaviour is correct and more consistent (case-lambda formals renamed hygienically like let-variables), nothing depends on case-lambda anaphora. Comments aligned to lambda-only scope; regression tests added in `tests/scheme/hygiene/template-binds-builtin-name.scm` pinning both directions (case-lambda formal → spliced body resolves to the global; lambda formal → keeps anaphoric spelling).

2. **Dead `__nlet_` branch** (baijum, types.zig): `stripHygienicPrefix` doesn't strip `__nlet_`, so the alternative was a no-op. Dropped it (named-let loop names are always locals, resolved by resolveLocal before the barrier check is reached).

3. **FYI (llvm_emit_freevars.zig bare-name scanners)**: agreed it's latent — macro-expanded forms are eval'd rather than natively emitted in the native tier, so `__hyg_N_` heads never reach those scanners. Left unchanged to avoid churn without a failing test.

src/expander_instantiate.zig
baijum: Confirmed resolved in 7d9970f8. The comments now scope the exception to lambda formals and cite #2252, and `tests/scheme/hygiene/template-binds-builtin-name.scm` pins the contrast precisely — case-lambda formal hygiene-renamed (spliced body reaches the global `car`, check 5) vs. lambda formal kept bare/anaphoric (spliced body binds the formal, check 6). All 7 checks pass locally.

For the record, I endorse the fix-the-comments direction over extending FORMAL_FLAG to case-lambda (CodeRabbit's suggestion above): nothing depends on case-lambda anaphora — SRFI 190's coroutine-generator uses plain `lambda` — and renaming case-lambda formals hygienically like let-variables is the more consistent R7RS behaviour. Resolving.

src/types.zig
baijum: Confirmed resolved in 7d9970f8 — the `__nlet_` alternative is dropped and the remaining `__hyg_`-only branch now carries a comment explaining why (named-let loop names are locals, resolved by the call path before the barrier check). Resolving.

baijum:
**LGTM** ✅ (GitHub blocks a formal approval here since I'm reviewing as the PR author's session, so leaving this as an explicit sign-off instead.)

Reviewed the hygiene rework end to end and exercised it against the built branch binary — interpreter *and* the `kaappi compile` native tier:

- Both headline fixes (#2003 procedure-reference capture, #2074 operator-keyword capture) behave correctly, along with `=>` in cond/case, nested quasiquote, `unquote-splicing`, `do`/`when`/`cond-expand` combinators, tail-position `call-with-values`/`call/cc`, macro-generating macros (transformer refs stay bare), and lambda-formal bareness not leaking use-site capture.
- I specifically tried to break the native tier with macro-introduced capturing lambdas inside a user `let` (renamed `__hyg_N_lambda` heads capturing a let-local across escaping closures) and could not — capture stays correct, and CI is green.
- The design is coherent (rename operator keywords + reserved-form allowlist + effective-name stripping at every dispatcher, with the deliberate lambda-formal anaphoric exception), and the comments are unusually thorough.

The two review points I raised were both resolved in 7d9970f8: the `FORMAL_FLAG` comments now match the lambda-only scope (with a new `tests/scheme/hygiene/template-binds-builtin-name.scm` pinning the case-lambda-vs-lambda contrast, all checks passing), and the dead `__nlet_` branch in `isContinuationBarrier` is gone. Nothing blocking remains from my side — good to merge.

Issue #1980, opened 2026-07-31
[LSP] Six protocol and lifecycle defects: no response on bad params, no shutdown state, wrong exit code, id coercion


baijum:
**Category:** Language server · **Severity:** protocol conformance · **Found by:** systematic audit v2, Phase 6D (#1890)

Six independent defects found by driving the real binary over framed JSON-RPC. Each has a control.

## 1. A request with unusable `params` gets no response at all

`hover`, `definition`, `references`, `completion` and `documentSymbol` all `orelse return` on a missing field. No reply frame is emitted, so **the client blocks on that id forever**.

**Control:** a well-formed request in the same session is answered normally.

This is the worst of the six — LSP clients keep a pending-request table, and an id that never resolves is a hang in the editor, not a visible error.

## 2. Malformed *headers* are fatal to the whole session

`Content-Length: abc`, a missing `Content-Length`, `Content-Length: 0`, and `Content-Length:` with no space each end the session silently, dropping every later message.

**Control:** malformed *bodies* are handled well — unparseable JSON, non-object JSON and a missing `method` are all skipped and the session continues. So the recovery path exists; it just isn't reached for header faults.

## 3. No shutdown state

`shutdown` before `initialize` returns success instead of `-32002` (it is dispatched above the `!initialized` guard). Requests *after* `shutdown` are served normally instead of `-32600`.

## 4. Exit code contract violated

LSP 3.17 requires status **1** when `exit` arrives without a prior `shutdown`. Kaappi exits **0** in both cases, so a supervising client cannot tell a clean shutdown from a crash-exit.

## 5. End-of-line overrun

`lineColToOffset` clamps to `text.len` rather than the line end, so hover at line 0 / character 200 resolves a symbol on **line 1**.

**Control:** the same text *with* a trailing newline returns `null`.

## 6. Unrepresentable ids become `id: 0`

A `null` or float request id is answered with `0`, matching no outstanding request.

**Control:** integer, negative and string ids all round-trip correctly.

## Confirmed correct

Nothing is advertised that isn't implemented (`formattingProvider`/`renameProvider`/`codeActionProvider` are all correctly absent). `Content-Length` is byte-exact — a lone `shutdown` reply is exactly 60 bytes. Notifications produce no frame; `jsonrpc: "2.0"` is on every reply; unopened URIs, `didChange` on an unopened document, negative and huge positions all degrade cleanly. `references` correctly skips occurrences inside strings and comments.

[...]

Issue #1979, opened 2026-07-31
[LSP] A define-syntax in one open document leaks into every other document's diagnostics


baijum:
**Category:** Language server · **Severity:** wrong-result · **Found by:** systematic audit v2, Phase 6D (#1890)

Macros defined in one open document leak through the shared `vm.macros` into the diagnostics of every other open document. The same file is reported clean or broken depending on what else the editor happens to have open.

## Reproduction

`a.scm`:
```scheme
(define-syntax m (syntax-rules () ((m) 1)))
```
`b.scm`:
```scheme
(m 1 2 3)
```

| session | `b.scm` diagnostics |
|---|---|
| open `b.scm` alone | `[]` — clean |
| open `a.scm`, then `b.scm` | **`KP2001`** |

Byte-identical `b.scm`, opposite results.

## Discriminating controls

- `b.scm` alone is clean — so it is not a latent error in the file.
- Macro definition *and* misuse in one file **is** correctly `KP2001` — the diagnostic itself is right.
- The leak **survives `didClose`** of the defining document.
- A plain `(define m …)` of the same name does **not** leak — so it is the macro table specifically, not globals. Globals were separately confirmed not to leak.

## Why it matters

An editor typically has several files open. The direction of the bug is the harmful one: a *false positive* appears in a file the user is not editing, attributed to a macro they cannot see from there. And because it survives `didClose`, closing the offending file does not clear it — only restarting the server does.

**Verified:** fresh ReleaseSafe build, isolated `KAAPPI_HOME`; both sessions driven against the real `kaappi-lsp` binary over framed JSON-RPC, reproduced independently.

PR #2253, opened 2026-08-08
Expand imported macros when computing LSP diagnostics

### LINKED ISSUE #2254: kaappi-lsp reports a false one-armed-if error on valid SRFI 42 comprehension guards
## Summary

`kaappi-lsp` flags valid SRFI 42 code with a false-positive error. The program compiles and runs correctly and `kaappi check` accepts it, so this is a pure false positive — but it shows a red "Problems: 1" under correct code (bad for teaching/demo material).

## Repro

With the `vscode-kaappi` extension + `kaappi-lsp` on kaappi v0.22.3:

```scheme
(import (scheme base) (scheme write) (srfi 42))
(display (list-ec (: i 1 11) (if (even? i)) i))   ; <- red squiggle (KP2001) here
(newline)
```

The LSP flags the `(if (even? i))` guard as a malformed one-armed `if`. The program runs fine and prints `(2 4 6 8 10)`. The same false error appears on the Pythagorean-triples guard `(if (= (* c c) (+ (* a a) (* b b))))`.

Only the bare `(if test)` guard is flagged; `(: i 1 11)` and the other SRFI 42 keywords are not.

## Root cause

`runDiagnostics` in `src/kaappi_lsp.zig` compiled each top-level form in isolation and never processed the file's `import` / `define-library` / `include` / `define-record-type` declarations. So the imported comprehension macros (`list-ec`, `sum-ec`, `vector-ec`, …) were never in scope when a later form was diagnosed. With the macro unexpanded, the compiler saw the comprehension's `(if test)` *filter qualifier* as a bare one-armed R7RS `if` and reported KP2001.

`kaappi check` does not have this bug because `check.zig` runs those environment-establishing forms for their effect first. A second latent gap: the server never set `vm.lib_paths`, so a file-based `.sld` like `(srfi 42)` couldn't be resolved at all.

## Environment

- kaappi v0.22.3
- Client: `vscode-kaappi` + `kaappi-lsp`
- Found while filming SRFI editor material for @KaappiScheme.

## Fix

See PR — `runDiagnostics` now runs env-setup forms via the shared `TopLevelHead` API (as `kaappi check` does) and the server seeds `vm.lib_paths`.

baijum:
Closes #2254

## Problem

`kaappi-lsp` flags valid SRFI 42 code with a false-positive error. Given:

```scheme
(import (scheme base) (scheme write) (srfi 42))
(display (list-ec (: i 1 11) (if (even? i)) i))   ; <- red squiggle (KP2001)
(newline)
```

the server paints a red squiggle under the `(if (even? i))` guard, reporting it as a malformed one-armed `if`. The program runs correctly (prints `(2 4 6 8 10)`), and `kaappi check` on the same file reports no issues — so this is a pure false positive. It shows a "Problems: 1" under correct code, which is bad for teaching material. The same appears on the Pythagorean-triples guard `(if (= (* c c) (+ (* a a) (* b b))))`.

## Root cause

`runDiagnostics` compiled every top-level form in isolation and **never processed the file's `import` / `define-library` / `include` / `define-record-type` declarations**. So the imported comprehension macros (`list-ec`, `sum-ec`, `vector-ec`, …) were never in scope when a later form was diagnosed. With the macro unexpanded, the compiler sees the comprehension's `(if test)` *filter qualifier* as a bare one-armed R7RS `if` and emits KP2001.

`kaappi check` gets this right because [`check.zig`](https://github.com/kaappi/kaappi/blob/main/src/check.zig) deliberately runs those environment-establishing forms for their effect first. The LSP was a cruder second implementation of the same read+compile loop that skipped that step.

A second, latent gap: the server never set `vm.lib_paths`, so a file-based `.sld` like `(srfi 42)` couldn't be resolved at all.

## Fix

- **`runDiagnostics`** now classifies each top-level form through the same `TopLevelHead` machinery `kaappi check` and the runtime share (#2114), so the three cannot drift:
  - top-level `begin` and the selected `cond-expand` clause splice and are recursed into;
  - the environment-establishing heads (`import`/`define-library`/`include`/`define-record-type`) are **run for their effect** so later forms see the bindings and macros they introduce;
  - everything else is compiled but not executed, exactly as before.
- **`vm.lib_paths`** is seeded with `~/.kaappi/lib` and the exe-relative `../lib` fallback, via the same `kaappi_paths` helpers `main.zig` uses (#1523), so file-based libraries resolve.

Deliberately preserved (all still covered by the LSP suite): the per-document macro reset (#1979), the first-error-only publish (#1980), and the whole-line `0..999` range sentinel.

## Tests

Added coverage to `tests/scheme/lsp/lsp.sh`:

- `list-ec` and Pythagorean-triples `if`-guards are diagnosed clean (cross-checked against `kaappi check`);
- a genuine top-level one-armed `if` still reports KP2001 (the fix suppresses nothing real).

`kaappi-lsp: 160 passed, 0 failed` (including the isolation/leak and range-sentinel guards). `zig build test` green; `zig fmt --check` clean.

### DIFF
--- a/src/kaappi_lsp.zig
+++ b/src/kaappi_lsp.zig
@@ -713,13 +713,19 @@ fn runDiagnostics(allocator: std.mem.Allocator, vm: *vm_mod.VM, uri: []const u8,
 
     // Retract the value bindings a previous document's `import` left in
     // `vm.globals` (the globals analogue of the macro reset above): remove every
-    // global not present at startup. Without this, `(import (srfi 1))` in one
-    // document would leave `fold` etc. resolvable while diagnosing another that
+    // global not present at startup. Without this, `(import (mylib))` in one
+    // document would leave its exports resolvable while diagnosing another that
     // imports nothing. Compilation never depends on a global's *value* here so
     // diagnostics were already correct, but hover/completion leaked across
-    // documents and the asymmetry with the macro reset was surprising. Held
-    // under the same write lock `importBinding` uses, with a `global_version`
-    // bump so any cached global lookups are invalidated.
+    // documents and the asymmetry with the macro reset was surprising.
+    //
+    // Because this runs at the *start* of every diagnostics run, one shared
+    // `vm.globals` means a document's imported names resolve for hover/completion
+    // only until another document is opened or edited, self-healing on that
+    // document's next run. That is the intended, bounded behaviour (matching the
+    // per-document macro reset) — strictly better than the pre-change state where
+    // imported names resolved nowhere — not a bug to chase into per-document
+    // environments.
     pruneImportedGlobals(vm, allocator);
 
     // Resolve library and `include` paths against the document's own directory
@@ -729,20 +735,22 @@ fn runDiagnostics(allocator: std.mem.Allocator, vm: *vm_mod.VM, uri: []const u8,
     // sibling `.sld` is resolved through `vm.lib_paths` — which never consults
     // `current_lib_dir` — so the document's directory must be prepended there
     // too, exactly as `main.zig` seeds the file's directory for `kaappi check`.
+    // `path_buf` backs the decoded path for the whole run; the slices below
+    // point into it, so it must outlive them (it does — it is this frame's).
     const saved_lib_dir = vm.current_lib_dir;
     const saved_lib_paths = vm.lib_paths;
     defer {
         vm.current_lib_dir = saved_lib_dir;
         vm.lib_paths = saved_lib_paths;
     }
+    var path_buf: [2048]u8 = undefined;
     var lib_paths_run: [8][]const u8 = undefined;
-    if (std.mem.startsWith(u8, uri, "file://")) {
-        const p = uri["file://".len..];
-        if (std.mem.lastIndexOfScalar(u8, p, '/')) |pos| {
-            vm.current_lib_dir = p[0 .. pos + 1];
+    if (fileUriToPath(uri, &path_buf)) |doc_path| {
+        if (std.mem.lastIndexOfScalar(u8, doc_path, '/')) |pos| {
+            vm.current_lib_dir = doc_path[0 .. pos + 1];
             // Prepend the document's directory (no trailing slash, like
             // main.zig) ahead of the auto-discovered paths, capacity permitting.
-            const dir = if (pos == 0) p[0..1] else p[0..pos];
+            const dir = if (pos == 0) doc_path[0..1] else doc_path[0..pos];
             if (base_lib_paths.len + 1 <= lib_paths_run.len) {
                 lib_paths_run[0] = dir;
                 @memcpy(lib_paths_run[1 .. base_lib_paths.len + 1], base_lib_paths);
@@ -799,12 +807,16 @@ fn runDiagnostics(allocator: std.mem.Allocator, vm: *vm_mod.VM, uri: []const u8,
 
 // Remove every global not present at startup — the value bindings a previous
 // document's `import` wrote into `vm.globals` — so imported names do not leak
-// across documents. Keys are collected first (mutating a map mid-iteration is
-// unsound), then removed under the write lock `importBinding` also takes, with a
-// `global_version` bump to invalidate any cached global lookups. The collected
-// key slices point at externally-owned strings, so they stay valid across the
-// removals.
+// across documents. The write lock `importBinding` also takes is held across the
+// whole operation — iteration, key collection, and removal — so a concurrent
+// child-thread reader (#958) never observes a half-pruned map; `global_version`
+// is bumped to invalidate any cached global lookups. Keys are collected before
+// removing (mutating a map mid-iteration is unsound) into a list; the collected
+// slices point at externally-owned strings, so they stay valid across removal.
+// Nothing in the loop re-acquires `globals_lock`, so holding it cannot deadlock.
 fn pruneImportedGlobals(vm: *vm_mod.VM, allocator: std.mem.Allocator) void {
+    vm.globals_lock.lock();
+    defer vm.globals_lock.unlock();
     var to_remove: std.ArrayList([]const u8) = .empty;
     defer to_remove.deinit(allocator);
     var it = vm.globals.iterator();
@@ -813,10 +825,49 @@ fn pruneImportedGlobals(vm: *vm_mod.VM, allocator: std.mem.Allocator) void {
             to_remove.append(allocator, entry.key_ptr.*) catch return;
     }
     if (to_remove.items.len == 0) return;
-    vm.globals_lock.lock();
-    defer vm.globals_lock.unlock();
     for (to_remove.items) |k| _ = vm.globals.remove(k);
     vm.global_version +%= 1;
+}
+
+// Convert a `file://` URI to a native filesystem path in `buf`, percent-decoding
+// `%XX` escapes and handling an empty or `localhost` authority and a Windows
+// drive-letter path (`file:///C:/x` -> `C:/x`). Returns null for a non-`file:`
+// URI (an editor may send `untitled:`/`vscode-vfs:` — no on-disk directory to
+// resolve against) or when the decoded path does not fit `buf`. Not exhaustive
+// URL parsing: only what document paths need so `include`/import resolution
+// matches the real paths `kaappi check` sees (baijum review, coderabbit review).
+fn fileUriToPath(uri: []const u8, buf: []u8) ?[]const u8 {
+    if (!std.mem.startsWith(u8, uri, "file://")) return null;
+    var rest = uri["file://".len..];
+    // Authority runs up to the next '/'. Accept only empty (`file:///…`) or
+    // `localhost`; a real remote host has no local path to offer.
+    if (rest.len > 0 and rest[0] != '/') {
+        const slash = std.mem.indexOfScalar(u8, rest, '/') orelse return null;
+        if (!std.mem.eql(u8, rest[0..slash], "localhost")) return null;
+        rest = rest[slash..];
+    }
+    var n: usize = 0;
+    var i: usize = 0;
+    while (i < rest.len) {
+        var c = rest[i];
+        if (c == '%' and i + 2 < rest.len) {
+            const hi = std.fmt.charToDigit(rest[i + 1], 16) catch 0xff;
+            const lo = std.fmt.charToDigit(rest[i + 2], 16) catch 0xff;
+            if (hi != 0xff and lo != 0xff) {
+                c = @intCast(hi * 16 + lo);
+                i += 3;
+            } else i += 1; // malformed escape: keep the '%' literally
+        } else i += 1;
+        if (n >= buf.len) return null;
+        buf[n] = c;
+        n += 1;
+    }
+    var path = buf[0..n];
+    // `file:///C:/x` decodes to `/C:/x`; drop the slash before the drive letter.
+    if (path.len >= 3 and path[0] == '/' and path[2] == ':' and
+        ((path[1] >= 'A' and path[1] <= 'Z') or (path[1] >= 'a' and path[1] <= 'z')))
+        path = path[1..];
+    return path;
 }
 
 // State captured by beginOutputRedirect so endOutputRedirect can restore it.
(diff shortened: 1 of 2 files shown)


baijum:
Reviewed the diff locally (build, `zig build test`, the full `tests/scheme/lsp/lsp.sh` suite — 160 passed including the four new cases — plus `-Dgc-stress=true` build and `zig fmt --check`).

The direction is right: routing the LSP's per-form classification through the shared `TopLevelHead` machinery (instead of re-implementing a third copy), running env-setup heads for their effect, and seeding `vm.lib_paths` are exactly what the false positive needs. Rooting is correct (`expr` is rooted across the recursion by the caller; `runTopLevelHead` re-roots internally), the error-code mapping mirrors `check.zig`'s `addRuntimeOrCompileError` (I verified both surfaces report KP2001 for a missing library), macro isolation still holds (imported macros land in `vm.macros`, which is reset per document — the leak tests stay green), and the negative control (a real one-armed `if` still KP2001) is good discipline.

Findings, in decreasing order of severity:

**1. Document-local libraries still diverge from `kaappi check` (the "cannot drift" goal isn't fully met).**
The new comment in `runDiagnostics` claims it resolves "file-relative import paths against the document's own directory" via `current_lib_dir` — but `resolveLibraryPath` never consults `current_lib_dir`; only `openIncludeFile` (i.e. `include`) does. `kaappi check` resolves a `.sld` next to the file because `main.zig` pushes the file's own directory into `vm.lib_paths`; the LSP never adds it. Repro (verified):
- `/tmp/…/mylib.sld` defining `(library (mylib) …)`, document `(import (mylib))\n(display my-const)` beside it.
- `kaappi check` → `no issues found` (exit 0); the LSP → **KP2001 "invalid syntax"** phantom error, from any CWD.

Suggest adding the document's directory to `vm.lib_paths` for the duration of the run (save/restore like `current_lib_dir`), mirroring `main.zig`'s file-dir entry, or extending the `.sld` search to consult `current_lib_dir`.

**2. The server now *executes* user code — repeatedly, in a long-lived process.**
Env-setup heads run on every `didOpen`/`didChange` (full sync). Two consequences:
- **Wire-protocol hazard**: `vm.stdout_port` is fd 1 and nothing in the LSP redirects it, so top-level `(display …)`/`(write …)` inside an included file, a `define-library` `begin` body, or an imported library's top-level code writes straight to the LSP's stdout between framed responses, corrupting the framing for the client. `kaappi check` executes the same forms, but it's one-shot and owns its whole stdout. Worth redirecting the VM's stdout/stderr ports to a sink (the server's own `log` already uses fd 2) so diagnostics stay side-effect-free.
- **Repeated execution cost**: an `(include "setup.scm")` or `define-library` body re-runs on every keystroke. (Plain `(import …)` of an already-loaded library is a cheap registry hit, so the common case is fine.)

**3. Imported *value* bindings now leak across documents (hover/completion only).**
`importBinding` puts value exports into `vm.globals`, which is *not* reset per run (only `vm.macros` is). So after doc A does `(import (srfi 1))`, hover on `fold` in doc B — which imports nothing — resolves (verified). Diagnostics are unaffected (macros are the only thing that changes compilation, and they're isolated), but this contradicts the "no globals leak" test's intent and is asymmetric with it: a plain `define` doesn't leak because it's compiled-not-run, while an `import` now does. At minimum worth an explicit test/acknowledgment; ideally roll back the globals additions at the end of each document's run (harder, since globals are the import target).

**4. Minor: the new tests don't pin the `lib_paths` setup.**
From the repo root (where `run-all.sh` runs), `lib/srfi/42.sld` resolves via the built-in `lib/` prefix in `resolveLibraryPath` *before* `vm.lib_paths` is consulted — so a regression in the new lib_paths block wouldn't fail CI. I verified the suite passes from a foreign CWD (which does exercise it), but the test itself doesn't force that. Consider `cd "$TMP"` (or similar) for the srfi-42 cases so the lib_paths code is load-bearing.

**5. Nits:**
- Error *messages* for run/import failures are now visible to users but generic — the LSP reports the registry template (KP2001 "invalid syntax") where `check` reports the specific detail ("library not found: (no.such.lib)"). Code parity holds; message parity doesn't (pre-existing LSP design, but import errors now occur much more often).
- `current_lib_dir` is derived from the raw URI with no percent-decoding, so documents in paths containing spaces/non-ASCII get a `%20`-laden directory and their file-relative `include`/imports won't resolve. `check` never has this problem (real paths). Also `file://host/…` URIs misparse.

Overall: the fix is correct for the reported SRFI-42 case and a real improvement (LSP and `check` finally agree on imported-macro expansion), but items 1–3 are worth addressing before merge — 1 is a same-shaped phantom error the PR is explicitly trying to eliminate, and 2 is a new failure mode in a long-lived server.

baijum:
Thanks for the thorough review — all of it landed. Pushed `aae86f1b` addressing findings 1–4; notes on 5 below.

**1. Document-local libraries now resolve like `check`.** You were right that `current_lib_dir` only feeds `include`, never `resolveLibraryPath`. The document's own directory is now prepended to `vm.lib_paths` for the duration of each run (save/restore), mirroring the file-dir entry `main.zig` adds — so `(import (mylib))` of a sibling `.sld` resolves instead of throwing a phantom KP2001. The overclaiming comment is fixed to distinguish the two mechanisms. Verified against your exact repro; new test `sibling-sld` is load-bearing precisely because `(mylib)` is not a built-in prefix, so only the doc-dir entry can resolve it.

**2. Executed side effects can no longer corrupt the wire.** Confirmed the hazard is real and reachable: importing a library runs its `begin` body, and `kaappi check` demonstrably prints it. The VM's current-output-port (both the `current_output_port_param` and `stdout_port` fallback) is now redirected to a discarding in-memory string port for each run and restored after; its cursor/length reset per run so the buffer never grows across the server's life. New `stdout-guard` test asserts library-body output stays off the wire, with a control confirming `check` really does execute that body. (The repeated-execution *cost* remains inherent to running env-setup for effect — same as `check` — so I left that as-is; the redirect is the part that was unsafe in a long-lived server.)

**3. Imported value bindings no longer leak across documents.** Every global not present at startup is retracted at the start of each run — the globals analogue of the existing macro reset — under the same write lock `importBinding` uses, with a `global_version` bump. `my-const` from doc A's import is no longer hoverable in doc B (new `globals-isolation` test), while it stays resolvable within A itself. Diagnostics were already correct; this closes the hover/completion asymmetry you flagged.

**4. lib_paths is now pinned.** The `sibling-sld` case fails without the doc-dir lib_paths change (the library isn't a built-in prefix), so it exercises the new plumbing regardless of CWD.

**5. Nits — acknowledged, not yet changed:**
- *Generic import-error messages*: still the registry template rather than the specific detail `check` carries. This is the pre-existing whole-line/template LSP design (`addDiagnostic` uses `code.message()`), and now that import errors surface more often it's more visible — worth a follow-up to thread the detail through, but out of scope here.
- *URI percent-decoding / `file://host/…`*: `current_lib_dir` and the doc-dir entry are still derived from the raw URI, so a path with `%20`/non-ASCII won't resolve. Doing this correctly (percent + UTF-8 + Windows `file:///C:/`) is more than a one-liner, so I've left it as a documented limitation rather than half-fix it.

Full suite: `kaappi-lsp: 167 passed, 0 failed`; `zig build test` green; `zig fmt --check` clean.

baijum:
Re-verified `aae86f1b` locally: build + `zig build test` + `zig fmt --check` + `-Dgc-stress=true` build all clean; full LSP suite 167/167, including the three new cases.

All four findings confirmed addressed:

1. **Sibling `.sld`** — re-ran the exact `/tmp/…/mylib.sld` repro: the LSP now publishes `"diagnostics":[]` where it previously threw KP2001. The two-mechanism split (`current_lib_dir` for `include`, doc-dir prepended to `vm.lib_paths` for imports) matches what `main.zig` actually does, and the corrected comment now describes it accurately. The `sibling-sld` test is genuinely load-bearing: `(mylib)` isn't a builtin prefix and isn't under any base path, so only the doc-dir entry can resolve it (verified it would fail from a CWD where `lib/` also wouldn't resolve).
2. **stdout containment** — `beginOutputRedirect`/`endOutputRedirect` cover both resolution paths (`current_output_port_param` first, `stdout_port` fallback), the sink buffer is truncated per run, and the `stdout-guard` test's check-side control proves the library body really executes. Framing stays intact after the noisy document (the subsequent clean-diagnostic assertion on the same stream would fail otherwise).
3. **Globals isolation** — the prune-under-lock with a `global_version` bump is the right shape. I confirmed the retraction works with a non-built-in library (my original `fold` probe was invalid: SRFI 1 is one of the 12 built-in SRFIs, so `fold` is a startup global, not an import leak — `my-const` via `(mylib)` is the correct probe and it's now null across documents).
4. **lib_paths pinned** — yes, `sibling-sld` covers it as analyzed above.

One interaction worth being a *conscious* decision (not a blocker): because the prune runs at the *start of every diagnostics run*, the imported bindings of any document are retracted as soon as *another* document is opened/edited. So in a multi-file session, hover/completion on an imported name in file A goes null after you touch file B — it self-heals only on A's next `didChange`/`didOpen`. The new `globals-isolation` control test masks this by using a fresh session (only the importing doc open); in the same session with both docs open, hovering A's own `my-const` after B opens returns null (verified). This is the inherent cost of sharing one `vm.globals` without per-document environments, and it's strictly better than the pre-PR state (imported names never resolved anywhere) — just worth a note in the PR description or a comment so the behavior is intentional rather than discovered.

Two residual edges, both minor: C-FFI native code that writes to fd 1 during a library load isn't covered by the port redirect (Scheme-level ports only), and the deferred nits (generic import-error message, URI percent-decoding) remain as documented.

This is good to go from my side. Thanks for the fast, complete turnaround.

baijum:
Second review pass addressed in `c4cadd9e` (CodeRabbit + local review).

**Applied:**

- **Globals prune lock ordering** — `pruneImportedGlobals` now holds `vm.globals_lock` across the whole operation (iterate → collect → remove), not just the removals, so a concurrent child-thread reader can't observe a half-pruned map. Nothing in the loop re-acquires the lock, so no deadlock.
- **File-URI → native path** — new `fileUriToPath` percent-decodes `%XX` (verified: a document under a path with a space now resolves its sibling `.sld`), accepts an empty or `localhost` authority, and strips the leading slash before a Windows drive letter (`file:///C:/x` → `C:/x`). Both `current_lib_dir` and the doc-dir `lib_paths` entry use the decoded path.
- **Test: vacuous control** — the globals-isolation control now asserts the positive hover payload (`"result":{"contents"`) instead of merely lacking `null`, so an error or dropped response can't pass it vacuously.
- **Test: lib_paths isolation negative control** — a document in a *different* directory imports a *fresh* library that lives only under the first document's directory, and asserts it's unresolved (KP2001, cross-checked against `kaappi check`). This proves the per-run `lib_paths` restore holds. It must use a fresh library: an already-loaded one resolves from `vm.libraries` regardless (see note below).

**Not applied, with reasons:**

- **Enable `sandbox_mode` during env-setup** — this would *reintroduce the bug*. `tryLoadLibraryFromFile` rejects every file-backed load under sandbox (only embedded libraries pass), so `(import (srfi 42))` and every ecosystem/FFI import would fail → the comprehension macro wouldn't expand → the KP2001 false positive returns, and the LSP would diverge from `kaappi check` (which never sandboxes). The re-execution *cost* CodeRabbit also flagged is inherent to running env-setup for effect (same as `check`); an import-set cache is a possible future optimization but not a correctness issue.
- **Second doc importing the *same* `(mylib)` from another dir** (the suggested negative control) — that resolves via the library cache (`vm.libraries`), not a lib_paths leak, so it wouldn't test what it claims. I used a fresh library instead (see above).

**Worth recording (from the local review):**

- *Library-cache leniency* — because `vm.libraries` persists across documents, a library first loaded while diagnosing doc A stays resolvable for doc B even if B couldn't reach it on disk. Benign for diagnostics (macro/value isolation still holds via the per-run resets) and only makes the LSP *more lenient* than `check` in that edge; a per-run library-cache reset would fix it at real re-load cost, so it's left as-is.
- *Prune-at-start hover behavior* — documented in a code comment: a document's imported names resolve for hover/completion until another document is opened/edited, self-healing on that document's next run. Intended and bounded, matching the macro reset.
- *C-FFI residual edge* — a comment records that native code writing to fd 1 during a library load bypasses the Scheme-port redirect, and why closing it (OS-level `dup2`) is left out.

169 LSP checks pass; `zig build test` green (`-Dgc-stress=true` build clean too); `zig fmt --check` clean.

Issue #2082, opened 2026-08-01
[R7RS 4.3.2] syntax-rules accepts two ellipses in one list pattern and splits the input arbitrarily — b always takes the last two elements


baijum:
**Category:** R7RS
**Spec reference:** R7RS §4.3.2, page 22 — the ⟨pattern⟩ grammar admits **at most one** ⟨ellipsis⟩ per list or vector pattern:

```text
(⟨pattern⟩ ...)
(⟨pattern⟩ ⟨pattern⟩ ... . ⟨pattern⟩)
(⟨pattern⟩ ... ⟨pattern⟩ ⟨ellipsis⟩ ⟨pattern⟩ ...)
(⟨pattern⟩ ... ⟨pattern⟩ ⟨ellipsis⟩ ⟨pattern⟩ ... . ⟨pattern⟩)
#(⟨pattern⟩ ...)
#(⟨pattern⟩ ... ⟨pattern⟩ ⟨ellipsis⟩ ⟨pattern⟩ ...)
```

SRFI 46's tail-pattern extension (implemented, `lib/srfi/46.sld`) widens
what may follow the ellipsis; it does not add a second one.
**Severity:** wrong-result (silent, and the split is arbitrary)

**Minimal reproduction:**

```scheme
(import (scheme base) (scheme write))

(define-syntax two-ell
  (syntax-rules ()
    ((_ a ... b ...) (list (list a ...) (list b ...)))))

(two-ell 1 2 3)   ;; expected: a syntax error at the define-syntax
                  ;; actual:   ((1) (2 3))
```

The pattern is not in the grammar, so there is no right answer — but the
answer given is not a defensible reading of anything. `b` collects the last
**two** elements regardless of input length, because the two extra ellipsis
tokens are counted as fixed tail elements by the SRFI 46 tail-pattern
arithmetic:

| call | `a` | `b` |
|---|---|---|
| `(two-ell)` | `KP2001: invalid syntax` | |
| `(two-ell 1)` | `KP2001: invalid syntax` | |
| `(two-ell 1 2)` | `()` | `(1 2)` |
| `(two-ell 1 2 3)` | `(1)` | `(2 3)` |
| `(two-ell 1 2 3 4 5)` | `(1 2 3)` | `(4 5)` |

So the *short* calls are rejected and the longer ones silently succeed, which
is the worst of both: the mistake surfaces as a confusing arity-shaped error
in one test case and as a plausible-looking wrong value in the next.

[...]

PR #2256, opened 2026-08-08
Validate syntax-rules ellipsis usage: template depth (#682) and pattern grammar (#2082)

baijum:
## Summary

Fixes **#682** (template-side ellipsis depth validation) and its pattern-side sibling **#2082** (at most one ellipsis per list/vector pattern). Both are R7RS 4.3.2 conformance defects where `syntax-rules` silently produced wrong output instead of a syntax error.

> Note: #2003 (and its sibling #2074) were already fixed on `main` by #2251 — I verified both fixes on a fresh build (including the full 17-keyword sweep from #2074) and they hold. This PR is the remaining open ellipsis-validation work.

### #682 — ellipsis depth under-use now errors instead of substituting `()`

```scheme
(define-syntax bad-depth
  (syntax-rules ()
    ((_ ((a ...) ...) (b ...))
     (list (list a b) ...))))
(bad-depth ((1 2) (3 4)) (10 20))
;; before: ((() 10) (() 20))   — silent wrong output
;; after:  compile error[KP2001]: invalid syntax
```

- `instantiateEllipsis` rejects a **directly-referenced** list binding whose depth exceeds the ellipsis run consuming it (`1 + extra_ellipsis`, so `(x ... ...)` consecutive-ellipsis flattening of a depth-2 binding stays legal).
- `instantiateTemplate` rejects a list binding used at depth 0 (no ellipsis at all).
- Legitimate shapes — nested `((a ...) ...)`, `(x ... ...)` flattening, SRFI 149's excess-ellipsis innermost replication with a deeper sibling — are untouched (all 21 srfi149 assertions pass).

### #2082 — multiple ellipses in one pattern now rejected at define-syntax

```scheme
(define-syntax two-ell
  (syntax-rules () ((_ a ... b ...) (list (list a ...) (list b ...)))))
;; before: accepted; (two-ell 1 2 3) => ((1) (2 3))  — arbitrary split
;; after:  compile error[KP2001]: invalid syntax at the define-syntax
```

- `parseSyntaxRules` validates every rule's pattern via a new `validPatternGrammar` walk (list + vector spellings, nested sub-patterns, custom ellipsis identifiers, ellipsis-as-literal carve-out).
- A first-position `...` stays a plain pattern variable (per the matcher), keeping `srfi136`'s `(cname field (... ...))` guard legal — 37 srfi136 assertions pass.

### Tests

- `tests/scheme/smoke/ellipsis-depth-mismatch.scm` — grew the invalid cases it never had; asserts the mismatch **raises** (shown to FAIL against a build with the fix reverted).
- `tests/scheme/srfi/srfi149.scm` / `srfi46.scm` — flipped the disabled `;; FAIL:` assertions to enabled `test-error`s, removed the enabled pins of the wrong answers.
- `src/tests_macros.zig` — 6 new Zig unit tests (both issues plus control shapes).

## Checklist

- [x] `zig build test` passes (also under `-Dgc-stress=true`)
- [x] Scheme test suites pass (`bash tests/scheme/run-all.sh` — 2089 pass, 0 fail; R7RS suite 1395 pass)
- [x] CHANGELOG updated


src/compiler_define_syntax.zig
```diff
@@ -788,6 +797,85 @@ pub fn parseSyntaxRules(self: *Compiler, spec: Value, extra_bound: []const []con
...
+            first = false;
+            cur = types.cdr(cur);
+        }
+        // Dotted tail: a plain pattern, not a list element. An ellipsis
+        // token there (`(a ... . ...)`) is outside the grammar too.
+        if (cur != types.NIL) {
+            if (types.isSymbol(cur)) {
+                const name = types.symbolName(cur);
+                if (std.mem.eql(u8, name, ellipsis_name) and !literalNamed(literals, name)) return false;
+            } else if (types.isPair(cur)) {
```
baijum: Minor: this `else if (types.isPair(cur))` branch is unreachable. The `while (types.isPair(cur))` loop above only exits once `cur` is *not* a pair, so `types.isPair(cur)` here is always false and the recursive call can never run. Harmless as defensive code, but it reads as if it handles a case that can't occur — worth dropping (or converting to an assert) for clarity.

src/compiler_define_syntax.zig
```diff
@@ -788,6 +797,85 @@ pub fn parseSyntaxRules(self: *Compiler, spec: Value, extra_bound: []const []con
...
+                return validPatternGrammar(cur, ellipsis_name, literals);
+            }
+        }
+        return true;
+    }
+    if (types.isVector(v)) {
```
baijum, line 852: Minor / optional: the vector branch duplicates the whole per-element ellipsis-counting scan from the list branch (the `seen_ellipsis` / `first` / symbol-vs-nested logic is identical). Since both spellings must stay in lockstep for correctness, a small shared helper — e.g. `scanLevel(elems_iter)` or a routine taking a slice of already-collected elements — would avoid the two copies drifting apart later. Not required for this PR.

baijum (commented):
Reviewed the full diff and verified it against a fresh build cherry-picked onto `main` (kept the PR worktree untouched).

**What I checked and confirmed:**
- `zig build test -Dtest-filter=macros` — passes (including the 6 new unit tests).
- The three modified Scheme test files run green: `ellipsis-depth-mismatch.scm` (6 PASS), `srfi149.scm` (21), `srfi46.scm` (28).
- The tricky control cases hold: `srfi136` (37 — first-position `(... ...)` guard), `srfi257` (34) and `srfi41` (93) and `srfi35` (macro-generating / condition macros), so the depth-0 `instantiateTemplate` change doesn't regress nested-`syntax-rules` templates.
- Full R7RS suite: 1395 pass, 0 fail. Hygiene + continuations suites clean.

**Logic review — both fixes look correct:**
- The template-side depth check keys off `templateReferencesVarDirectly`, so a deeper binding that's only reachable through an inner ellipsis is classified `indirect` and never trips the `b.depth > 1 + extra_ellipsis` guard — that's why `((a (b c) ...) ...)` (the SRFI-35 shape) and `(x ... ...)` flattening stay legal. Moving the `extra_ellipsis` scan above the count loop is side-effect-free, so the reordering is safe.
- `validPatternGrammar`'s first-position skip mirrors the matcher's `i+1` lookahead exactly, which is what preserves the ellipsis-as-literal and `(... ...)` escape cases.

Nice, well-commented change with genuinely discriminating tests (the reverted-fix FAIL demonstration and the removed wrong-answer pins are exactly right). Two minor, non-blocking notes inline.

baijum:
## Review response — all findings addressed (commit `5b3fd934`)

**CodeRabbit + baijum (converged on the same spot): the dotted-tail branch.**
Both were right, in complementary ways: the `else if (types.isPair(cur))` branch was *unreachable* (the `while (types.isPair(cur))` loop only exits once `cur` is not a pair — baijum's point), and a **vector** dotted tail — `(_ . #(a ... b ...))` — was therefore never validated, accepting two ellipses in one vector pattern (CodeRabbit's point). The branch now recurses into `types.isVector(cur)` instead, with a regression test (`two ellipses in a vector dotted tail` in `tests_ellipsis.zig`) verified to fail against the pre-fix code.

**CodeRabbit: tests_macros.zig over the 1500-line policy.** The 7 ellipsis tests moved to a dedicated `src/tests_ellipsis.zig` (wired through `vm_tests.zig`, same pattern as `tests_macros_nested_sr.zig`). `tests_macros.zig` is back to its pre-PR size (1963 lines — still over the limit, but that's pre-existing; I didn't grow it further).

**CodeRabbit: smoke test harness.** The file was written in the old verdictless `display`/`exit` style (`docs/dev/testing.md` explicitly discourages this — "do not write a test that prints its answer"). Converted to the documented SRFI-64 shape: imports `(scheme process-context)` and `(srfi 64)`, `test-begin`/`test-end`, grabs the runner before `test-end`, exits 1 on fail-count. Verified to exit **1** against the unfixed build (212d2428) and pass with the fix.

**baijum (optional, skipped): dedupe the list/vector scans in `validPatternGrammar`.** Agreed the two copies could drift, but the loops genuinely differ (spine-walk vs slice) and you marked it not-required; kept the change minimal.

Validation: `zig build test` + `-Dgc-stress=true` green; `run-all.sh` 2089 pass / 0 fail; R7RS suite 1395 pass; the three modified Scheme files fail on 212d2428 and pass on the fix.

baijum:
LGTM ✅

Re-reviewed commit `5b3fd934` on a fresh build — both of my notes are handled well, and the dotted-tail one turned out to matter:

- **Vector dotted-tail fix is correct.** Replacing the unreachable `isPair(cur)` branch with `isVector(cur)` closes the real gap it was masking. Confirmed `(_ . #(a ... b ...))` now errors with `KP2001` (was silently accepted), and the new `two ellipses in a vector dotted tail` regression test covers it.
- **Test split + SRFI-64 harness** look good — `tests_ellipsis.zig` wired through `vm_tests.zig`, smoke test now uses the documented `test-begin`/`test-end` shape and exits 1 on failure.
- Skipping the list/vector scan dedupe is a fair call — the two loops genuinely differ.

Verified green locally: macro unit tests, `srfi149/46/136/257`, and the full R7RS suite (1395/0). Nice work.

Issue #2166, opened 2026-08-02
[R7RS 6.2] Exact complex components are stored as f64 + flag: arithmetic drops exactness, make-rectangular 2^53+1 is silently off by one, write and real-part disagree about the same object


baijum:
**Found by:** external testing (paal's reader test suite against kaappi v0.22.0) · **Verified:** main @ 98195459 and v0.22.1

`Complex` (`src/types.zig:533`) stores its components as two f64s plus two exactness *flags*:

```zig
pub const Complex = struct {
    header: Object,
    real: f64,
    imag: f64,
    exact_real: bool = false,
    exact_imag: bool = false,
};
```

An exactness flag over an already-rounded double cannot carry the exact value it claims, so every consumer is forced to choose between honest-but-inexact and exact-but-wrong — and the codebase currently does both, in different places, about the same object. Kaappi advertises `exact-complex` **and** `exact-closed` in `(features)` (#1177 made the claims agree with each other; this issue is about the claims agreeing with the arithmetic).

## Instances

All verified with the current main build. `z` is `(make-rectangular 3/2 1)`; `(exact? z)` → `#t`.

**1. Every arithmetic operation on an exact complex returns inexact** — including the identity and a self-subtraction that collapses the type:

```scheme
(- z)     ; => -1.5-1.0i   R7RS requires exact -3/2-1i
(+ z z)   ; => 3.0+2.0i    should be exact 3+2i
(* z z)   ; => 1.25+3.0i   should be exact 5/4+3i
(- z z)   ; => 0.0         should be exact 0
(+ z)     ; => 1.5+1.0i    the *identity* drops exactness
(/ z)     ; => 0.46153846153846156-0.3076923076923077i   should be 6/13-4/13i
```

Cause: `+`/`-`/`*`/`/` route any complex operand through `toComplexParts` (f64 pairs) and rebuild with `makeComplexOrReal(real, imag)` — the flag-less constructor (`src/primitives_arithmetic.zig:302-510`, `src/primitives_numeric.zig:962`).

**2. `make-rectangular` silently corrupts exact values at construction:**

```scheme
(make-rectangular 9007199254740993 1)   ; => 9007199254740992+1i  — 2^53+1, silently off by one, still claims exact
(make-rectangular (expt 10 25) 1)       ; => 0/0+1i               — unreadable printed form
(= (real-part (make-rectangular (expt 10 25) 1)) (expt 10 25))    ; => #f — you don't get back what you put in
[...]

PR #2257, opened 2026-08-08
Fix rational→flonum conversion past f64 range; close the m/2^k complex read-back gap (#2183, #2182)

### LINKED ISSUE #2182: string->number's complex grammar is narrower than the reader's: rational and inf/nan parts rejected, bare 1+2i exactness diverges
**Category:** Numeric tower · **Severity:** wrong-result · **Found by:** #1911's fix (PR #2181), which unified `#e`/`#i` handling and deliberately left these pre-existing complex-*grammar* gaps out of scope.

R7RS 7.1.1's `<complex R> → <real R> + <ureal R> i` production admits rational and inf/nan parts, and the reader accepts them. `string->number` parses complex parts with `parseFloat` only, so it rejects valid syntax the reader accepts — and on the shapes both sides accept, the two disagree about exactness.

## Reproduction (verbatim at `6d60e659`, fresh build, isolated `KAAPPI_HOME`)

```scheme
(exact? '1+2i)                        ⟹ #t     ;; reader: integer parts are exact
(exact? (string->number "1+2i"))      ⟹ #f     ;; string->number: always inexact  <-- silent divergence
(string->number "1/2+3/4i")           ⟹ #f     ;; reader reads 1/2+3/4i fine
(string->number "3.0+inf.0i")         ⟹ #f     ;; reader reads 3.0+inf.0i fine
(string->number "#i1/2+3i")           ⟹ #f     ;; reader gives 0.5+3.0i
```

R7RS §6.2.7 (p.40): *"The rules used by a particular implementation for `string->number` must also be applied to `read`… in order to maintain consistency."* These are the same class of violation as the #1911 family, on the complex-part grammar axis rather than the exactness axis.

## Why these survived #1911's fix

PR #2181 routes the reader's `#e`/`#i` bodies through the shared `parseNumberText`, but `.complex` tokens deliberately keep their token-level parse — precisely because `readNumber`'s complex grammar is *wider* than `parseNumberText`'s, and re-parsing would have regressed valid literals like `#e1/2+3/4i`. That preserved the reader's correct behavior and left `string->number`'s narrower grammar as-is.

## Existing disabled tests that re-enable on the fix

`tests/scheme/compliance/reader-delimiter-gaps.scm`, section 7, carries two `FAIL: TBD` cells for exactly this (`s->n accepts 1/2+3i`), plus the adjacent note that `1/2+3i` prints an exact-looking real part while `real-part` hands back an inexact 0.5.

## Suggested direction

Extend `parseNumberText`'s complex branch (`src/primitives_numeric.zig`) with a real part-parser: `[+-]inf.0/nan.0`, `n/d` rationals (i64, matching the reader's `@floatFromInt(n) / @floatFromInt(d)` computation exactly), and decimal floats — returning per-part syntax-based exactness flags the way `readNumber` computes them (`!has_dot and !has_exp`). Because #1911 made `parseNumberText` the single number grammar, fixing it here fixes `string->number` and every future prefixed-reader interaction in one place. The bare `1+2i` exactness change should be called out loudly when it lands: `(string->number "3+4i")` flipping from inexact to exact is visible to `eqv?`/`write`.

**Verified:** fresh ReleaseSafe build at `6d60e659` (PR #2181's commit), all five reproductions plus the reader-side controls, reproduced twice.

---

## Addendum (from PR #2181 review): the reader has its own rational-part ceiling, and #2181's printer is the first producer of a shape behind it

The reader's complex grammar accepts rational parts only while both digit runs fit i64: a **bignum-denominator** rational followed by a complex suffix is a read error, because `bigRationalToken` requires a delimiter after the rational (by design, to avoid silent datum splits) and the complex-suffix detection only exists on the i64 paths.

```scheme
(write #e1e-300+1i)   ;; since #2181: "<16-digit mantissa>/<316-digit 2^k>+1i" — the honest exact value
(read (open-input-string <that text>))   ⟹ read error: invalid number literal
```

So the full part-grammar unification has three pieces, in this order:

1. **#2183 first** — the scaled rational→f64 conversion. Extending the reader's grammar before it would read the form above back as a silently wrong `0.0+1i` (the true value's denominator exceeds 2^1023, where today's `toF64(num)/toF64(den)` collapses to 0), trading #2181's loud error for silent corruption.
2. **Reader**: complex-suffix detection after the bignum-rational paths in `readNumber` (mirroring the i64-rational path at `reader_tokens.zig:266`), converting the real part via the fixed conversion.
3. **`string->number`**: the shared part-parser described above, so both parsers gain the full grammar together and `#e1e-300+1i` round-trips end to end.

The gap is pinned in both directions by `tests/scheme/compliance/reader-exactness-gaps.scm` section 7 and a paired unit test in `src/tests_numeric.zig` — each fails loudly when the reader learns the shape, pointing at the round-trip assertion to flip to.

### LINKED ISSUE #2183: Rational->flonum conversion yields 0.0 when the denominator alone overflows f64: (inexact (/ 1 (expt 2 1049))) is 0.0, not the subnormal
**Category:** Numeric tower · **Severity:** wrong-result (silent) · **Found by:** review of PR #2181 (the suggested fix for a printer round-trip gap would have routed through this conversion, turning a loud read error into a silent 0.0)

Rational→flonum conversion is `toF64(numerator) / toF64(denominator)` everywhere: each side saturates to ±inf **independently**, so the quotient is wrong whenever either side alone leaves f64 range while the true quotient is representable. One of the three call paths carries a partial band-aid, which makes the failure surface inconsistent as well as wrong.

## The three call paths

| path | used by | band-aid |
|---|---|---|
| `inexactFn` rational arm (`primitives_numeric.zig:337`) | the `inexact` procedure | NaN result → bignum quotient/remainder retry — catches **inf/inf only** |
| `applyExactness` `.inexact` rational arm (`primitives_numeric.zig`) | `string->number`, and since #2181 the reader's `#i` | none |
| `types.toF64` / `toF64Ext` rational arms | everything else (`toComplexParts`, comparisons, …) | none |

## Reproductions (verbatim at `6d60e659`, fresh ReleaseSafe build, isolated `KAAPPI_HOME`)

**1. finite/inf → 0.0** (denominator alone overflows; true value a representable subnormal):

```scheme
(inexact (/ 1 (expt 2 1074)))   ⟹ 0.0    ;; expected 4.9406564584124654e-324 (min subnormal)
(inexact (/ 1 (expt 2 1049)))   ⟹ 0.0
(inexact (exact 1e-320))        ⟹ 0.0    ;; exact→inexact round-trip destroyed
```

**2. inf/finite → +inf.0** (numerator alone overflows; true value an ordinary integer):

```scheme
(inexact (/ (+ (expt 2 1030) 1) (expt 2 1000)))   ⟹ +inf.0    ;; expected 1073741824.0 (2^30)
```

**3. inf/inf → +nan.0 on the parser path only** (both overflow; `inexactFn`'s band-aid catches it, the parsers' arm has none — a procedure-vs-parser divergence):

```scheme
(inexact (/ (+ (expt 2 1100) 1) (expt 2 1050)))   ⟹ 1125899906842624.0   ;; correct, 2^50
(string->number "#i<same 332 digits>/<same 317 digits>")  ⟹ +nan.0
(read (open-input-string "#i<same literal>"))             ⟹ +nan.0        ;; read and string->number agree (6.2.7 holds) — both disagree with the procedure
```

## Discriminating controls

| control | result | tells us |
|---|---|---|
| `(inexact (/ 1 (expt 2 1000)))` | `9.332636185032189e-302` — correct | cliff is where a **side** exits f64 range, not "small values" |
| `(inexact (/ (expt 2 1100) (expt 2 1050)))` | correct `2^50` | reducible pairs are saved by gcd reduction before conversion — the pair must be irreducible to reproduce |
| `(exact 1e-320)` | correct exact rational | the →exact direction is fine; only →inexact loses |

## Suggested fix

One shared scaled conversion used by all three sites (replacing `inexactFn`'s special-case retry): compare the bit lengths of numerator and denominator, shift the larger down so both convert in range (53+ε significant bits suffice), divide, then apply the removed exponent with `std.math.ldexp`, which rounds into subnormals correctly.

## Why it matters beyond the exotic magnitudes

PR #2181's exact-complex printer emits `mantissa/2^k` components with k up to 1074 — the honest exact value of an f64. The reader-side grammar extension that would make those round-trip (see #2182) must route through this conversion — landing that extension **before** this fix would read `#e1e-300+1i`'s written form back as a silently wrong `0.0+1i` instead of today's loud read error. Sequencing: this first, then #2182.

**Verified:** fresh ReleaseSafe build at `6d60e659`, isolated `KAAPPI_HOME`; all three manifestations, the round-trip case, and all three controls reproduced twice; all three call-path bodies read directly.

baijum:
Closes #2183, closes #2182.

## #2183 — rational→flonum conversion collapses when a side alone leaves f64 range

`(inexact (/ 1 (expt 2 1074)))` was `0.0` instead of the minimum subnormal, `(inexact (/ (+ (expt 2 1030) 1) (expt 2 1000)))` was `+inf.0` instead of `2^30`, and the `#i`/`#e` parser paths produced `+nan.0` when both sides overflowed — every rational→f64 path computed `toF64(num)/toF64(den)`, saturating each side independently. `inexact` survived inf/inf only via a special quotient/remainder retry; `string->number` and `read` had no band-aid at all.

**Fix:** a shared `types.rationalToF64` used by all five call sites (`types.toF64`, `primitives.toF64`, `toF64Ext`, `inexactFn`'s rational arm — band-aid deleted — and `applyExactness`'s `.inexact` arm, which also fixes SRFI-18's `getSleepSeconds`):

1. both magnitudes normalized to their top 64 significant bits,
2. quotient computed to 64+ significant bits with a u128 division (a plain f64/f64 ratio of truncated operands is off by 1–2 ulp, and a short quotient loses mantissa precision),
3. rounded with round-half-to-even,
4. removed power of two re-applied via frexp + exact-power multiply, which rounds through the subnormal range correctly — including the exact tie at 2^-1075, which `std.math.ldexp` mishandles (rounds it up to the min subnormal).

Verified bit-for-bit against a correctly-rounded Python oracle over 3301 rationals spanning powers of two from 2^-1100 to 2^1100, random bignum ratios, lopsided numerator/denominator sizes, and the m/2^k round-trip shape.

## #2182 (remaining scope) — the m/2^k read-back gap closes

#2181's exact-complex printer emits `m/2^k` spellings (denominator up to 2^1074, a bignum) for tiny exact-flagged components, but the reader's complex grammar stopped at i64 rational parts, so `(write #e1e-300+1i)` could not be read back — a loud error, pinned in both directions by tests. With #2183's conversion in place the read-back is exact, so the grammar opens up:

- **Reader:** all four bignum-rational paths (`readNumber` and `readIntegerWithRadix`, numerator- and denominator-overflow) gain a complex tail mirroring the i64-rational path — real, imaginary, and signed pure-imaginary bignum rational parts — converting through the shared scaled conversion.
- **string->number:** `parseComplexComponent` routes through the same gated helper, so the two parsers cannot drift (R7RS 6.2.7).
- **The gate:** a bignum rational whose value is *not* exactly representable in f64 (`10^25/3`) stays a loud read error / `string->number` `#f`, preserving the never-masquerade policy of #2243. `m/2^k` forms (the printer's honest output) round-trip exactly.
- **Adjacent parity gap:** the i64-rational-real path's imaginary part now runs the same round-trip exactness check as the main complex path, so `1/2+123456789012345678901234567890i` is a loud error instead of silently rounding an exact-flagged component.

Pins flipped: `reader-exactness-gaps` section 7 and the paired `tests_numeric` cell now assert the round-trip; the m/2^k shapes join the round-trip matrix; gate rejections pinned. New scheme + unit tests cover #2183's conversions end to end.

**Sequencing note:** #2182's addendum called for exactly this order — #2183 first, then the grammar. Landing the grammar before the conversion would have read the honest `m/2^k` spelling back as a silently wrong `0.0+1i`.

## Known remaining gaps (pre-existing, out of scope)

- `+inf.0i` exactness-flag parity: the reader's symbol-path special case marks the real part inexact (`0.0`), string->number's pure-imaginary arm marks it exact (`0`). `eqv?`-visible, predates both issues.
- The i64-rational-real path's imaginary scan does not accept exponent spellings (`3/4+1e-300i` read-errors while string->number accepts); predates both issues.
- #2166 (exact complex components stored as f64+flag) remains open — `real-part` of an exact complex still hands back the rounded f64.

### DIFF
--- a/src/bignum.zig
+++ b/src/bignum.zig
@@ -977,13 +977,14 @@ pub fn valueBitLen(v: Value) usize {
 /// and k <= 1074); a value that would silently round must stay a loud
 /// error instead of masquerading as exact (kaappi#2182/#2183).
 pub fn rationalExactInF64(num: Value, den: Value) bool {
-    if (types.isFixnum(num) and types.toFixnum(num) == 0) return true; // 0/den == 0.0
+    const bl_raw = valueBitLen(num);
+    if (bl_raw == 0) return true; // 0/den == 0.0, exactly representable
     const k = powerOfTwoExponent(den) orelse return false;
     // Reduce: value == num_odd * 2^(v2 - k), num_odd odd with valueBitLen -
     // v2 significant bits.
     const v2 = trailingZeroBits(num);
     const k_eff: i64 = @as(i64, @intCast(k)) - @as(i64, @intCast(v2));
-    const bl = valueBitLen(num) - v2; // significant bits of the reduced value
+    const bl = bl_raw - v2; // significant bits of the reduced value
     if (k_eff > 1074) return false; // value < 2^-1075: rounds to 0
     if (k_eff <= 1022) return bl <= 53; // value >= 2^-1022: normal range
     // Value in (2^-1074, 2^-1022): subnormal iff num_odd < 2^(k_eff - 1022),
--- a/src/bignum.zig
+++ b/src/bignum.zig
@@ -900,6 +900,135 @@ pub fn bignumExactInF64(v: Value) bool {
 /// #2243).
 pub const complex_rational_limit: i64 = 1_000_000;
 
+// ---------------------------------------------------------------------------
+// Bignum rational complex parts (kaappi#2182/#2183)
+//
+// The reader's complex grammar stops at i64 rational parts whose magnitude
+// stays within complex_rational_limit. The printer's exact-component
+// fallback emits m/2^k spellings (odd mantissa <= 2^53, k <= 1074) whose
+// denominator is a bignum, and reading those back requires accepting bignum
+// rational parts -- but only when the value is exactly representable in f64,
+// so an exact-flagged component never silently rounds (the same
+// never-masquerade policy as the i64 limit). These helpers are the shared
+// gate + conversion behind both the reader and string->number.
+// ---------------------------------------------------------------------------
+
+/// The exponent k when v == 2^k (v a positive fixnum or bignum), else null.
+/// A rational's f64-exactness requires its denominator to be a power of two.
+pub fn powerOfTwoExponent(v: Value) ?u64 {
+    if (types.isFixnum(v)) {
+        const n = types.toFixnum(v);
+        if (n <= 0) return null;
+        const mag = @as(u64, @intCast(n));
+        if (@popCount(mag) != 1) return null;
+        return 64 - @clz(mag) - 1;
+    }
+    if (!types.isBignum(v)) return null;
+    const bn = types.toBignum(v);
+    if (!bn.positive or bn.len == 0) return null;
+    var k: u64 = 0;
+    var found = false;
+    for (bn.limbs[0..bn.len], 0..) |limb, i| {
+        if (limb == 0) continue;
+        if (found or @popCount(limb) != 1) return null;
+        found = true;
+        k = @as(u64, @intCast(i)) * 64 + (64 - @clz(limb) - 1);
+    }
+    return if (found) k else null;
+}
+
+/// Number of trailing zero bits of a nonzero value's magnitude (v2(v)).
+fn trailingZeroBits(v: Value) u64 {
+    if (types.isFixnum(v)) {
+        const n = types.toFixnum(v);
+        const mag: u64 = if (n < 0) @as(u64, @intCast(-(n + 1))) + 1 else @intCast(n);
+        return @ctz(mag);
+    }
+    const bn = types.toBignum(v);
+    var zeros: u64 = 0;
+    for (bn.limbs[0..bn.len]) |limb| {
+        if (limb == 0) {
+            zeros += 64;
+        } else {
+            return zeros + @ctz(limb);
+        }
+    }
+    return zeros;
+}
+
+/// Significant-bit count of a fixnum or bignum magnitude; 0 for zero.
+pub fn valueBitLen(v: Value) usize {
+    if (types.isFixnum(v)) {
+        const n = types.toFixnum(v);
+        const mag: u64 = if (n < 0) @as(u64, @intCast(-(n + 1))) + 1 else @intCast(n);
+        return if (mag == 0) 0 else 64 - @clz(mag);
+    }
+    const bn = types.toBignum(v);
+    var i = bn.len;
+    while (i > 0 and bn.limbs[i - 1] == 0) i -= 1;
+    if (i == 0) return 0;
+    return (i - 1) * 64 + (64 - @clz(bn.limbs[i - 1]));
+}
+
+/// True when the rational num/den (num signed, den positive) is exactly
+/// representable in f64 -- i.e. its own value round-trips losslessly
+/// through the f64 conversion. This is exactly the shape the printer emits
+/// for exact complex components (m/2^k with the reduced mantissa <= 2^53
+/// and k <= 1074); a value that would silently round must stay a loud
+/// error instead of masquerading as exact (kaappi#2182/#2183).
+pub fn rationalExactInF64(num: Value, den: Value) bool {
+    if (types.isFixnum(num) and types.toFixnum(num) == 0) return true; // 0/den == 0.0
+    const k = powerOfTwoExponent(den) orelse return false;
+    // Reduce: value == num_odd * 2^(v2 - k), num_odd odd with valueBitLen -
+    // v2 significant bits.
+    const v2 = trailingZeroBits(num);
+    const k_eff: i64 = @as(i64, @intCast(k)) - @as(i64, @intCast(v2));
+    const bl = valueBitLen(num) - v2; // significant bits of the reduced value
+    if (k_eff > 1074) return false; // value < 2^-1075: rounds to 0
+    if (k_eff <= 1022) return bl <= 53; // value >= 2^-1022: normal range
+    // Value in (2^-1074, 2^-1022): subnormal iff num_odd < 2^(k_eff - 1022),
+    // i.e. bl <= k_eff - 1022 -- then the mantissa num_odd << (1074 -
+    // k_eff) must fit 52 bits. Otherwise the value is back in normal range.
+    if (bl <= @as(usize, @intCast(k_eff - 1022))) {
+        return bl <= @as(usize, @intCast(52 - (1074 - k_eff)));
+    }
+    return bl <= 53;
+}
+
+/// Parse a radix-R rational pair (num with optional sign, den positive) to
+/// f64. i64 parts within complex_rational_limit divide directly; a part
+/// that overflows i64 goes through the bignum path, gated on exact f64
+/// representability (`rationalExactInF64`) so a non-representable bignum
+/// rational (10^25/3) stays a loud error. A zero denominator, invalid
+/// digits, or a gated-out value yields null. This is the shared
+/// implementation for bignum rational complex parts in both the reader and
+/// string->number (kaappi#2182/#2183).
+pub fn parseRationalToF64(gc: *memory.GC, num_str: []const u8, den_str: []const u8, radix: u8) ?f64 {
+    var nb: [256]u8 = undefined;
+    var db: [256]u8 = undefined;
+    const num = stripUnderscores(num_str, radix, &nb) orelse return null;
+    const den = stripUnderscores(den_str, radix, &db) orelse return null;
+    if (std.fmt.parseInt(i64, num, radix)) |n| {
+        if (std.fmt.parseInt(i64, den, radix)) |d| {
+            if (d == 0) return null;
+            if (@abs(n) > complex_rational_limit or @abs(d) > complex_rational_limit) return null;
+            return @as(f64, @floatFromInt(n)) / @as(f64, @floatFromInt(d));
+        } else |derr| {
+            if (derr != error.Overflow) return null;
+        }
+    } else |nerr| {
+        if (nerr != error.Overflow) return null;
+    }
+    // At least one part overflows i64: bignum path, gated on exactness.
+    const num_v = parseBignumString(gc, num_str, radix) catch return null;
+    var slot = gc.rootedSlot(num_v) catch return null;
+    defer slot.release();
+    const den_v = parseBignumString(gc, den_str, radix) catch return null;
+    if (isZero(den_v)) return null;
+    if (!rationalExactInF64(slot.get(), den_v)) return null;
+    return types.rationalToF64(slot.get(), den_v);
+}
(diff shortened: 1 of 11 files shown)


baijum (commented):
Reviewed the numeric conversion and reader/parser changes. The core `rationalToF64` rewrite is well-reasoned and I confirmed it against the built binary: the #2183 subnormal/overflow cases, the 2^-1075 tie, and the m/2^k bignum-rational round-trips all pass, and the never-masquerade gate (10^25/3) correctly stays loud. Tests are thorough on both parsers. One gap below; otherwise looks solid.

src/bignum.zig
```diff
@@ -900,6 +900,136 @@ pub fn bignumExactInF64(v: Value) bool {
...
+/// string->number (kaappi#2182/#2183).
+pub fn parseRationalToF64(gc: *memory.GC, num_str: []const u8, den_str: []const u8, radix: u8) ?f64 {
+    var nb: [256]u8 = undefined;
+    var db: [256]u8 = undefined;
+    const num = stripUnderscores(num_str, radix, &nb) orelse return null;
+    const den = stripUnderscores(den_str, radix, &db) orelse return null;
+    if (std.fmt.parseInt(i64, num, radix)) |n| {
+        if (std.fmt.parseInt(i64, den, radix)) |d| {
+            if (d == 0) return null;
+            if (@abs(n) > complex_rational_limit or @abs(d) > complex_rational_limit) return null;
```
baijum: The `i64` fast path in `parseRationalToF64` rejects a rational the moment either side exceeds `complex_rational_limit` (1e6), *without* consulting the exactness gate. So a value whose denominator is a power of two in the range `(10^6, i64max]` — exactly representable in f64, i.e. a legitimate `m/2^k` printer form — is rejected as a read error even though the bignum path a few magnitudes higher accepts the same shape.

This leaves a "dead zone" between the i64-limit and the bignum path. Reachable counterexamples (all exact complex numbers whose *own printed spelling* the reader refuses):

```scheme
(define (rt x) (equal? x (read (open-input-string (number->string x)))))
(rt (make-rectangular (/ 1 (expt 2 30)) 1))  ; writes 1/1073741824+1i  -> READ ERROR
(rt (make-rectangular (/ 1 (expt 2 23)) 1))  ; writes 1/8388608+1i     -> READ ERROR
(rt (make-rectangular (/ 1 (expt 2 40)) 1))  ; writes 1/1099511627776+1i -> READ ERROR
```

`1/1073741824` is `m/2^k` with `m=1, k=30` — precisely the printer-honest shape this PR claims round-trips exactly, and the "Known remaining gaps" list doesn't mention it. Only the `k >= 63` (bignum-denominator) subset was closed.

Suggested fix: when an i64 part exceeds `complex_rational_limit`, fall through to the same `rationalExactInF64` gate the bignum path uses instead of returning `null`. That keeps the within-limit exact-rational policy (`1/3` still stored exact), accepts the f64-exact power-of-two denominators, and still rejects over-limit non-representable ratios (`1/2000001` -> `#f`). Same reasoning applies to the sibling line at bignum.zig:1049 if it's ever fed printer output.

Not a regression (the i64 gate predates this PR), but it's squarely in this PR's stated scope of closing the `m/2^k` complex read-back gap, so worth either fixing here or explicitly listing as a remaining gap.

baijum:
Re-reviewed commit f38ba9f — the dead-zone finding is fully addressed, and the additional fixes look correct. Verified against the freshly built binary:

- **Dead zone closed:** `(make-rectangular (/ 1 (expt 2 30)) 1)`, `2^23`, `2^40` all round-trip now via the shared `i64RationalExactInF64` gate, while non-representable ratios (`1/3`, `10^25/3`) stay exact-rational / loud as intended.
- **Top-binade overflow:** `(inexact (/ (+ (expt 2 1024) 1) 2))` = `8.98846567431158e307` ✓ (was `+inf.0`). The `e == 1024` two-step scale through `2^1023` is exact.
- **Correct rounding (independent check):** ran a differential test of `rationalToF64` against a Python `Fraction`→`float` oracle over 1018 rationals — random bignum ratios to 400 bits, lopsided num/den, the subnormal tail (2^-1074..2^-1076), the top binade, and **adversarial exact half-ulp tie constructions** `(2m+1)·2^e`. **0 mismatches.** The u192 operands + division-remainder sticky bit hold up.

Test suites green: `exact-inexact-842-848` (12), `reader-exactness-gaps` (202), `tests_numeric` unit filter. Nice work — no further blocking concerns from me.

One tiny non-blocking note: in `i64RationalExactInF64`, since `den` is a positive i64 power of two, `k = ctz(den) <= 62`, so `k_eff <= 62` always — the `k_eff > 1074` and subnormal branches are unreachable for this path. Harmless (they mirror the bignum version for clarity), just noting they're dead code here.

baijum:
Thanks for the thorough review — all six findings are addressed in f38ba9fe:

1. **Off-by-one overflow guard (Critical)** — fixed: the guard is now `e > 1024` with an exact two-step scale through 2^1023 for `e == 1024`, so the whole top binade [2^1023, 2^1024) stays finite. `(inexact (/ (+ (expt 2 1024) 1) 2))` is now 8.98846567431158e307.
2. **Gate missing upper bound (Major)** — fixed: the normal-range branches require `bl <= k_eff + 1024`, so `2^2001/2` is rejected loudly instead of converting to an exact-flagged +inf.0.
3. **Radix not honored in the imag scan (Critical)** — fixed: radix-aware digit predicate + `parseRadixUrealToF64` for non-decimal magnitudes; `#x<bignum>/2+12i` now reads the imag as hex 18.
4. **Imag not gated by exactIntegerRoundTrips (Major)** — fixed: the bignum-rational tail's imag is now gated, so a bignum integer imag past 2^53 is a loud error.
5. **Dead zone between i64 limit and bignum path (Major)** — fixed: the i64 path shares the same exactness gate in both the reader and `parseRationalToF64`, so `1/2^40+1i` (the printer's own m/2^k output) reads exactly, while `1/20000000000001` stays a loud error.
6. **Sticky-bit tie misrounding (Major)** — fixed: operands are now reduced to 128 significant bits (u192 division) and the division remainder is used as a sticky bit for half-ties, making the conversion correctly rounded for all but the measure-zero case of a true value within ~2^-128 of a tie.

The oracle sweep was extended to 5607 rationals (top binade, subnormal tail, adversarial half-ulp tie constructions) with zero mismatches, and regression tests cover every finding. Also added the radix-prefixed bignum-rational complex cells to the round-trip matrix as suggested. All 22 CI checks pass.

PR #2258, opened 2026-08-08
Fix SRFI-178 bitvector-logical-shift shifting the wrong way (#2083)

### LINKED ISSUE #2083: [SRFI-178] bitvector-logical-shift moves bits the wrong way for both signs — fails the SRFI's own test suite
**Category:** SRFI
**Spec reference:** SRFI 178 §"Quasi-integer operations" — "Returns a bitvector equal in length to *bvec* containing the logical left shift (toward lower indices) when *count*>=0 or the right shift (toward upper indices) when *count*<0. Newly vacated elements are filled with *bit*."
**Severity:** wrong-result

`bitvector-logical-shift` moves bits in the opposite direction from the spec, for **both** signs of `count`. The two assertions below are lifted verbatim from SRFI 178's own test suite (`test/quasi-ints.scm`, lines 8–13 of the reference repository).

**Minimal reproduction:**

```scheme
(import (scheme base) (scheme write) (srfi 178))
(write (bitvector->list/int (bitvector-logical-shift (bitvector 1 0 1 1) 2 0)))
;; SRFI 178's own test suite: (1 1 0 0)     actual: (0 0 1 0)
(newline)
(write (bitvector->list/int (bitvector-logical-shift (bitvector 1 0 1 1) -2 #t)))
;; SRFI 178's own test suite: (1 1 1 0)     actual: (1 1 1 1)
```

**Discriminating control:** `(bitvector-logical-shift (bitvector 1 0 1 1) 0 0)` returns `(1 0 1 1)` — the `count = 0` path is correct, so the defect is exactly the two sign branches. The two neighbouring procedures in the same source region, `bitvector-count-run` and `bitvector-first-bit`, are byte-for-byte equivalent to the reference implementation.

**Not a reference-implementation defect.** The reference implementation's `%bitvector-left-shift` (positive count) computes `out[i] = bvec[i + count]` — toward lower indices, agreeing with the spec prose — while `lib/srfi/178.sld` computes `out[i] = bv[i - count]`. The reference was loaded into this build as an alternate library and diffed against `(srfi 178)` over a 523,361-check corpus; `bitvector-logical-shift` was one of only three procedures to diverge (the other two are `bitvector-pad`/`bitvector-pad-right` in the shrinking case, which the spec leaves ambiguous and which is deliberately not filed).

**Source location:** `lib/srfi/178.sld`, `bitvector-logical-shift` — the `(>= count 0)` branch writes `out[i] = bv[i - count]` and the negative branch writes `out[i] = bv[i + shift]`; both are inverted relative to the reference's `%bitvector-left-shift` / `%bitvector-right-shift`.

**Disabled test:** `tests/scheme/srfi/srfi178-audit.scm` — 4 assertions marked `;; FAIL: #2083`.

baijum:
## Summary

Fixes #2083: `bitvector-logical-shift` in `lib/srfi/178.sld` moved bits the wrong way for **both** signs, failing SRFI 178's own test suite.

The spec says `count>=0` is a logical **left** shift (toward lower indices, `out[i] = bvec[i + count]`) and `count<0` a logical **right** shift (toward upper indices, `out[i] = bvec[i - |count|]`), with vacated elements filled with `bit`. Both branches were inverted relative to the reference implementation's `%bitvector-left-shift` / `%bitvector-right-shift`:

- left branch wrote `out[i] = bv[i - count]` instead of `bv[i + count]`
- right branch wrote `out[i] = bv[i + |count|]` instead of `bv[i - |count|]`

The loop bounds were coupled to the wrong formulas, so they moved with the fix (left: `i` in `[0, n-count)`; right: `i` in `[|count|, n)`).

## Changes

- `lib/srfi/178.sld` — corrected both branches (the `count=0` path was already correct and is unchanged).
- `tests/scheme/srfi/srfi178-audit.scm` — enabled the 4 assertions that were disabled pending this fix (the first two are the SRFI's own `test/quasi-ints.scm` cases verbatim). The 2 discriminating controls (identity at count 0, full-length shift) still pass.
- `tests/scheme/srfi/srfi178.scm` — corrected the two `bitvector-logical-shift` values that had pinned the buggy behavior.

`bitvector-count-run` / `bitvector-first-bit` were verified byte-for-byte correct against the reference and left untouched, as were the spec-ambiguous `bitvector-pad` / `bitvector-pad-right` shrinking cases (not part of this issue).

## Verification

- Repros from the issue now produce the expected `(1 1 0 0)` and `(1 1 1 0)`.
- 1024 differential cases (16 patterns x 16 counts x 4 fill bits) agree with a direct translation of the reference implementation.
- `zig-out/bin/kaappi --lib-path lib tests/scheme/srfi/srfi178-audit.scm` -> 375 passes.
- `bash tests/scheme/run-all.sh` -> 2089 pass, 0 fail.
- `zig build test` passes; `zig fmt --check src/` passes.

## Checklist

- [x] `zig build test` passes
- [x] Scheme test suites pass (`bash tests/scheme/run-all.sh`)
- [x] New tests added for new behavior (4 previously-disabled audit assertions enabled; 2 conformance values corrected)
- [x] Documentation updated (if applicable) — none needed; `docs/audit-strategy.md` only references the issue number

### DIFF
--- a/lib/srfi/178.sld
+++ b/lib/srfi/178.sld
@@ -497,14 +497,17 @@
 
     ;; --- quasi-integer ops ------------------------------------------------
 
+    ;; A logical left shift moves bits toward lower indices (out[i] = bv[i + count])
+    ;; and a right shift toward upper indices (out[i] = bv[i - shift]); indices
+    ;; read outside the source are filled with bit (#2083).
     (define (bitvector-logical-shift bv count bit)
       (let* ((n (%len bv)) (b (%norm-bit bit)) (out (make-bytevector n b)))
         (if (>= count 0)
-            (do ((i (- n 1) (- i 1))) ((< i count) (%raw-make-bitvector out))
-              (bytevector-u8-set! out i (bitvector-ref/int bv (- i count))))
+            (do ((i 0 (+ i 1))) ((>= i (- n count)) (%raw-make-bitvector out))
+              (bytevector-u8-set! out i (bitvector-ref/int bv (+ i count))))
             (let ((shift (- count)))
-              (do ((i 0 (+ i 1))) ((>= i (- n shift)) (%raw-make-bitvector out))
-                (bytevector-u8-set! out i (bitvector-ref/int bv (+ i shift))))))))
+              (do ((i shift (+ i 1))) ((>= i n) (%raw-make-bitvector out))
+                (bytevector-u8-set! out i (bitvector-ref/int bv (- i shift))))))))
 
     (define (bitvector-count bit bv)
       (let ((b (%norm-bit bit)) (n (%len bv)))
--- a/tests/scheme/srfi/srfi178-audit.scm
+++ b/tests/scheme/srfi/srfi178-audit.scm
@@ -168,24 +168,22 @@
 ;;;  left shift (toward lower indices) when count>=0 or the right shift
 ;;;  (toward upper indices) when count<0. Newly vacated elements are filled
 ;;;  with bit."
-;;; The two assertions below are the SRFI's own test suite verbatim
-;;; (test/quasi-ints.scm lines 8-13).
 ;;; ------------------------------------------------------------------
 
-;; FAIL: #2083 (bitvector-logical-shift moves bits the wrong way for both signs)
-;; (test-equal "logical-shift +2 moves toward lower indices"
-;;             '(1 1 0 0) (L (bitvector-logical-shift (bitvector 1 0 1 1) 2 0)))
-;; FAIL: #2083
-;; (test-equal "logical-shift -2 moves toward upper indices, filling with the bit"
-;;             '(1 1 1 0) (L (bitvector-logical-shift (bitvector 1 0 1 1) -2 #t)))
-;; FAIL: #2083
-;; (test-equal "logical-shift +1 of (0 0 0 1)"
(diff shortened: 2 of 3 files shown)


baijum (commented):
Reviewed the fix against the SRFI-178 specification and ran the suites — this is correct and well-covered.

**Correctness.** The spec reads: *"the logical left shift (toward lower indices) when count>=0 or the right shift (toward upper indices) when count<0. Newly vacated elements are filled with bit."* The new bodies implement exactly that: the `count>=0` branch computes `out[i] = bv[i+count]` (bits move to lower indices), the negative branch `out[i] = bv[i-shift]` (toward upper indices), with the untouched region left as the `make-bytevector n b` fill. The direction comment on the definition matches. The prior code had both branches inverted.

**Edge cases hold.** `count > n` (either sign) terminates the `do` loop immediately, leaving an all-fill result — matching the "full length leaves only the fill bit" test. Empty bitvector is a no-op. `count 0` is preserved as the identity control.

**Tests.** Good move restoring the four SRFI-verbatim assertions in `srfi178-audit.scm` from commented-`FAIL` back to live `test-equal`, and correcting the two expected values in `srfi178.scm` to what the reference implementation produces. I ran both locally: 85/85 and 375/375 pass.

LGTM.

PR #2259, opened 2026-08-08
Persist #!fold-case across read calls on the same port

### LINKED ISSUE #2175: #!fold-case does not persist across separate read calls on the same port
**Category:** Reader / port I/O · **Severity:** wrong-result (silent, narrow) · **Found while:** fixing #1893/#1920/#1940/#1945 (PR #2174)

R7RS 7.1.1 says a `#!fold-case` directive affects reading "from the same port" from that point on. Kaappi applies it only within the *current* `read` call's parse buffer: `readDatumFn` constructs a fresh `Reader` per call, and the `fold_case` flag lives (and dies) on that Reader. Once the call that consumed the directive returns its datum, later calls never see the directive again.

## Reproduction

```scheme
(import (scheme base) (scheme read) (scheme file) (scheme write))
;; file contains: #!fold-case\nFOO\nBAR\n
(define p (open-input-file "fold2.scm"))
(write (list (read p) (read p)))
;; prints: (foo BAR)   — expected (foo bar)
```

Not boundary-related — a 15-byte file reproduces. Verified at PR #2174's head; the same happens on `main`.

Note the boundary interaction is already correct as of PR #2174: a directive separated from its datum by the 4096-byte chunk boundary *within one call* folds correctly (the loop no longer discards a buffer that held a directive — `Reader.saw_directive`), and `tests/scheme/compliance/reader-port-refill-gaps.scm` pins that. This issue is only about state crossing from one `(read p)` call to the next.

## Suggested direction

Persist the mode on the `Port` (`types_port.zig`): seed each call's fresh Reader from `port.fold_case`, and write the Reader's final flag back after a successful parse. The string-port path needs the same treatment. `include` and whole-file loading are unaffected (single Reader over the whole source).

**Priority rationale:** spec deviation with a silently wrong answer, but the path is narrow — `#!fold-case` via the `read` procedure across multiple calls — and recoverable, so medium per the rubric; silence is the aggravator that keeps it at the top of medium rather than low.

baijum:
Fixes #2175.

R7RS 7.1.1 says a `#!fold-case` directive affects reading "from the same port" from the point it appears on. Kaappi applied it only within the current read call: `readDatumFn` builds a fresh `Reader` per call, so `Reader.fold_case` lived and died with it.

## Change

- Add `Port.fold_case` (`src/types_port.zig`) — a plain bool, no GC tracing or write-barrier obligation.
- In `readDatumFn` (`src/primitives_io.zig`), seed each call's fresh Reader from `port.fold_case` and write the Reader's final flag back after a successful parse, at all four Reader sites:
  - string-port path (seeds + writes back),
  - incremental fd path (seeds + writes back on datum return),
  - post-EOF path (seeds + writes back on datum return),
  - peek-byte-only path (seeds only — a lone byte cannot hold a directive).
- `#!no-fold-case` resets the same field, so it persists too.
- Re-pin the `Port` field inventory in `src/tests_gc_tracing.zig` (compiler-enforced; bool adds no GC obligation).

The within-one-call chunk-boundary handling (`Reader.saw_directive`) is untouched: a directive split across the 4096-byte boundary is still re-parsed from the kept buffer, and the write-back only fires once a datum parse succeeds.

## Tests

New `tests/scheme/compliance/fold-case-persists.scm` (SRFI-64): file port, string port, `#!no-fold-case` reset, mid-stream directive, and a boundary-straddling directive persisting to the next read call. The three discriminating cases fail without the fix and pass with it.

- `zig build test` — 1710 pass, 7 skip
- `zig build test -Dgc-stress=true` — 1788/1788 pass
- `bash tests/scheme/run-all.sh` — 695 scheme files + 1395 R7RS tests, 0 fail
- Existing `fold-case-unicode.scm` (8) and `reader-port-refill-gaps.scm` (53) still pass.

Signed-off-by: Baiju Muthukadan <baiju.m.mail@gmail.com>

### DIFF
--- a/src/primitives_io.zig
+++ b/src/primitives_io.zig
@@ -1854,6 +1854,7 @@ fn readFromPeekByteOnly(gc: *@import("memory.zig").GC, port: *types.Port) Primit
     const source: [1]u8 = .{b};
     var reader = reader_mod.Reader.init(gc, &source);
     reader.mark_immutable = false;
+    reader.fold_case = port.fold_case; // a lone byte can't hold a #! directive, so no write-back
     defer reader.deinit();
     const maybe_datum = parseDatumForRead(&reader) catch |err| {
         if (err == reader_mod.ReadError.OutOfMemory) return PrimitiveError.OutOfMemory;
@@ -1889,12 +1890,17 @@ fn readDatumFn(args: []const Value) PrimitiveError!Value {
 
         var reader = reader_mod.Reader.init(gc, source);
         reader.mark_immutable = false;
+        reader.fold_case = port.fold_case;
         defer reader.deinit();
         const maybe_datum = parseDatumForRead(&reader) catch |err| {
             if (err == reader_mod.ReadError.OutOfMemory) return PrimitiveError.OutOfMemory;
             return raiseReadError(gc, err);
         };
         const datum = maybe_datum orelse return types.EOF;
+        // The parse saw the whole remaining source, so its final flag is
+        // the port's state from this point on (a directive may have been
+        // consumed before the datum) — persist it for the next call (#2175).
+        port.fold_case = reader.fold_case;
         // Advance string_pos by amount consumed
         port.string_pos += reader.pos;
         if (combined.items.len > 0 and reader.pos > 0) {
@@ -1948,10 +1954,15 @@ fn readDatumFn(args: []const Value) PrimitiveError!Value {
             var reader = reader_mod.Reader.init(gc, buf.items);
             reader.mark_immutable = false;
             reader.incomplete_input = true;
+            reader.fold_case = port.fold_case;
             defer reader.deinit();
             const result = parseDatumForRead(&reader);
             if (result) |maybe_datum| {
                 if (maybe_datum) |datum| {
+                    // The datum parse succeeded, so any directive was fully
+                    // inside the consumed span: persist the final flag for
+                    // the next call on this port (#2175).
+                    port.fold_case = reader.fold_case;
                     // Save unconsumed bytes back to port buffer.
                     const remaining = buf.items[reader.pos..];
                     if (remaining.len > 0) {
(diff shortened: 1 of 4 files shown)


baijum (commented):
## Review — persist `#!fold-case` across `read` calls

The fix is correct and cleanly scoped. `Port.fold_case` is the right home for state that R7RS 7.1.1 says lives on the port, and seeding each fresh `Reader` from it + writing back after a successful parse threads the flag through every `readDatumFn` path. I traced all four Reader sites and the semantics hold:

- **String-port whole-source path** — write-back at end; because `parseDatumForRead` stops at the datum boundary (`reader.pos`) and doesn't consume trailing whitespace/directives, a directive *after* the datum correctly does **not** leak into `port.fold_case`. The `mid-stream directive` test pins exactly this.
- **Incremental fd path** — write-back only on datum return; the `saw_directive` buffer-keep re-seeds each refill from the (unchanged) `port.fold_case` and re-parses from buffer start, so directive application stays idempotent. Good.
- **Post-EOF path** — write-back before returning the datum.
- **Peek-byte-only path** — seed only, no write-back. Correct: a lone byte can't hold a full `#!…` directive.

The GC reasoning is sound — a plain `bool` adds no marking/sweeping/barrier obligation, and the `tests_gc_tracing.zig` re-pin keeps the field inventory compiler-enforced.

The test file is thorough: it covers both port kinds (the two distinct fresh-Reader sites), the `#!no-fold-case` reset, mid-stream application, and the 4089-byte boundary straddle that exercises the `saw_directive` refill machinery. The three discriminating cases genuinely fail pre-fix.

### Minor (non-blocking)

One path drops the flag: when the port's remaining content is a directive followed immediately by EOF, the post-EOF parse returns `null` and hits `const datum = maybe_datum orelse return types.EOF;` (`src/primitives_io.zig:2047`) **before** the `port.fold_case = reader.fold_case` write-back on line 2050. In practice this is harmless — there's nothing left to read on that port, so the lost flag can never be observed — but it's the one place a directive is parsed without its effect persisting. Worth a one-line comment noting the early-return skips the write-back deliberately, in case a future interleaved read/peek path makes it reachable.

LGTM.

PR #2263, opened 2026-08-08
Honor top-level redefinitions of the five tail fast-path names (Fixes #2033)

### LINKED ISSUE #2033: [R7RS 5.3.1] A top-level redefinition of call/cc, apply, eval or call-with-values is ignored in tail position
**Category:** R7RS
**Spec reference:** R7RS §5.3.1, "Top level definitions" — "At the top level of a program, a definition ... has essentially the same effect as the assignment expression ... if ⟨variable⟩ is bound to a non-syntax value." Nothing in R7RS makes `call/cc`, `apply`, `eval` or `call-with-values` immune to that.
**Severity:** wrong-result

`compiler.zig`'s tail-position dispatch sends `(apply …)`, `(eval …)`, `(call/cc …)`, `(call-with-current-continuation …)` and `(call-with-values …)` to hand-written superinstructions. The guard around each one is

```zig
if (self.resolveLocal(sym_name) == null and (try self.resolveUpvalue(sym_name)) == null)
```

— a *local* or *upvalue* binding is consulted, a **global** rebinding is not. So a program that redefines one of these five names at top level gets its own definition everywhere except in tail position, where the builtin runs instead and the user's procedure is silently discarded.

**Minimal reproduction:**

```scheme
(define (call/cc f) 'user-callcc)
(define (tail-use)    (call/cc 42))        ; tail position
(define (nontail-use) (list (call/cc 42))) ; not tail position
(write (guard (e (#t 'ERR)) (tail-use)))    (newline)  ; expected: user-callcc  actual: ERR
(write (guard (e (#t 'ERR)) (nontail-use))) (newline)  ; expected: (user-callcc) actual: (user-callcc)
```

All five names, each called in tail position after a top-level redefinition:

| form | expected | Kaappi | chibi-scheme |
|---|---|---|---|
| `(apply + (list 1 2))` | `user-apply` | `3` | `user-apply` |
| `(eval '(+ 1 2) (interaction-environment))` | `user-eval` | `3` | honours it |
| `(call/cc (lambda (k) 1))` | `user-callcc` | `1` | `user-callcc` |
| `(call-with-current-continuation (lambda (k) 1))` | `user-ccc` | `1` | `user-ccc` |
| `(call-with-values (lambda () 1) list)` | `user-cwv` | `(1)` | `user-cwv` |

**Discriminating control:** a **local** binding of the same name *is* honoured in tail position —

```scheme
(let ((call/cc (lambda (f) 'local))) ((lambda () (call/cc 42))))  ;; => local, correct
```

so the mechanism is specifically the missing globals check, not tail-position compilation in general. The non-tail column above is the second control: the identical call one position away resolves correctly.

**Verified:** fresh ReleaseSafe build at `d0d03deb`, isolated `KAAPPI_HOME`, cache cleared. Cross-checked against chibi-scheme, which honours the redefinition for all five (its `eval` differs only because `interaction-environment` is unbound there).

**Source location:** `src/compiler.zig:~875-905` (the four `std.mem.eql(u8, sym_name, …)` blocks dispatching to `compiler_passthrough.compileApplyTail` / `compileCallWithValuesTail` / `compileCallCCTail` / `compileEvalTail`).

**Disabled test:** `tests/scheme/audit/primitives_control-audit.scm` (marked `;; FAIL: #this`); the three local-shadowing controls are enabled next to it.

baijum:
Fixes #2033 — [R7RS 5.3.1] A top-level redefinition of `call/cc`, `apply`, `eval` or `call-with-values` was ignored in tail position.

## The bug

`compiler.zig`'s tail-position dispatch sent `(apply …)`, `(eval …)`, `(call/cc …)`, `(call-with-current-continuation …)` and `(call-with-values …)` to hand-written superinstructions guarded only by `resolveLocal`/`resolveUpvalue` — a **global** rebinding of the name was not consulted. A program that redefined one of these five names at top level got its own definition everywhere except in tail position, where the builtin ran instead and the user's procedure was silently discarded. R7RS 5.3.1: *"At the top level of a program, a definition … has essentially the same effect as the assignment expression."*

```scheme
(define (call/cc f) 'user-callcc)
(define (tail-use)    (call/cc 42))        ; tail position — ran the builtin (ERR)
(define (nontail-use) (list (call/cc 42))) ; not tail — correct
```

## The fix

Each fast path is now gated on the compile-time global binding, mirroring `IR.isRedefined` and `vm_dispatch_helpers.lookupGlobalLocked`'s resolution order (raw name, then the hygienic-prefix fallback to the bare name):

- a `set!` target in the enclosing form, or a truncated pre-scan (`set_targets_all`), declines the fast path;
- a name unbound in a library/restricted environment declines it (the builtin no longer silently runs for a name the program never bound);
- everything else (the 99.9% case) keeps the structural superinstruction — **zero runtime cost**, the gate only decides which bytecode the compiler emits.

## Adjacent fixes the gate forced

- **Synthesized references stay pristine (#1715):** the let-values / let*-values / define-values / case-lambda / define-record-type desugarings minted bare `apply`/`call-with-values` symbols that were indistinguishable from user text. They now carry the `base_binding_prefix` so they remain bound to the `(scheme base)` procedures under a redefinition; `ir.zig` lowers a base-prefixed special-form head as a passthrough so it still reaches `compileForm`'s dispatch instead of bypassing it as a plain call (fast path preserved). The define-values single-name `list` consumer is base-prefixed too.
- **Native backend parity (#1803):** `emitApplyForm` mirrored the interpreter's old "tail ignores rebinding" behavior; it now gates on the module's rebound/native redefinitions the same way.
- **Docs:** `docs/dev/llvm-backend.md` updated.

## Tests

- New `tests/scheme/compliance/top-level-redefinition-2033.scm` — all five names (tail + non-tail controls), non-procedure redefinitions, hygiene-renamed macro-template references, `set!`-in-form suppression, and desugaring immunity under redefinition.
- The previously disabled audit assertions in `primitives_control-audit.scm` are re-enabled with real top-level redefinitions (the old proxy used an internal `let` define, which shadows lexically and never exercised the bug).
- New Zig unit test in `src/tests_core_eval.zig`.
- `native-apply-lowering-1803.sh`'s "rebound" case updated to the corrected expectation (it documented the bug as expected behavior).
- Verified: both new test files fail on `main` and pass here; full suite is green — **700 Scheme files, 1395 R7RS assertions, unit tests (incl. `-Dgc-stress=true`)**, WASM build, and the native tier matches the interpreter on the repro.

## Known limitation (unchanged, out of scope)

A redefinition compiled **after** the caller is baked in stays stale in tail position — the same compile-time-baking trade-off the constant folder already makes for `(+ 1 2)` (only ordinary calls re-resolve globals at run time). No new divergence was introduced.

---

## Review follow-ups (commit `8e09371e`)

- **CodeRabbit:** the original `define-record-type` test used R7RS constructor-first syntax, where `(parent …)` parses as a *field spec* — it never exercised the changed R6RS inherited-constructor paths in `vm_records.zig`. Replaced with top-level R6RS clause-syntax records covering **both** inherited-ctor variants (protocol-less split-args path and protocol path); both fail on `main` and pass here. While in those paths, the remaining bare synthesized refs (`list`/`append`/`car`/`cdr`) got the same base-binding-prefix treatment as `apply`.
- **baijum:** `emitApplyForm`'s `<2 operands` early return fired before the rebound check, needlessly abandoning native compilation for a *rebound* tail `apply` with one operand. Gated on `!rebound` so that shape takes the generic indirect call (matching the interpreter's #2033-gated ordinary call path); docs updated.
- **baijum:** trimmed the stale `buildLetValues` doc comment that described the removed `is_tail`-dependent reasoning.

### DIFF
--- a/src/compiler_bindings.zig
+++ b/src/compiler_bindings.zig
@@ -636,21 +636,15 @@ pub fn compileLetValues(self: *Compiler, args: Value, dst: u16, is_tail: bool) C
         var j = count;
         while (j > 0) {
             j -= 1;
-            // Every inner apply (j > 0) sits in the tail position of its
-            // enclosing consumer lambda, so it always compiles through
-            // compileApplyTail's structural bytecode fast path (matched on
-            // a literal `apply` symbol, checked only against local/upvalue
-            // shadowing) regardless of this form's own is_tail — that path
-            // never looks `apply` up by name, so a bare symbol is already
-            // safe there. Only the outermost (j == 0) apply's tail-ness
-            // follows this let-values's own is_tail; when that's false, it
-            // falls through to ordinary call compilation, which resolves
-            // `apply` like any other identifier and can be shadowed by a
-            // top-level redefinition (#1715).
-            const apply_sym = if (j == 0 and !is_tail)
-                try compiler_mod.Compiler.trueBuiltinRefOrSymbol(gc, "apply")
-            else
-                gc.allocSymbol("apply") catch return CompileError.OutOfMemory;
+            // Every apply here is a compiler-synthesized reference and must
+            // mean the (scheme base) apply regardless of any top-level
+            // redefinition (#1715, #2033): the tail-position superinstruction
+            // recognizes the base-binding-prefixed spelling and skips its
+            // redefinition gate, and the prefixed name resolves through the
+            // base registry if the fast path is unavailable (non-tail). A
+            // bare symbol would be indistinguishable from user text and get
+            // routed to the user's redefined `apply` under #2033's gate.
+            const apply_sym = try compiler_mod.Compiler.trueBuiltinRefOrSymbol(gc, "apply");
             const inner_body = gc.allocPair(inner, types.NIL) catch return CompileError.OutOfMemory;
             const consumer = gc.allocPair(lambda_sym, gc.allocPair(formals_arr[j], inner_body) catch return CompileError.OutOfMemory) catch return CompileError.OutOfMemory;
             const temp_sym = temp_syms[j];
@@ -671,7 +665,7 @@ pub fn compileLetStarValues(self: *Compiler, args: Value, dst: u16, is_tail: boo
     const body = types.cdr(args);
     if (body == types.NIL) return CompileError.InvalidSyntax;
 
-    const desugared = buildLetValues(self, bindings, body, is_tail) catch |err| return switch (err) {
+    const desugared = buildLetValues(self, bindings, body) catch |err| return switch (err) {
         error.InvalidSyntax => CompileError.InvalidSyntax,
         else => CompileError.OutOfMemory,
     };
@@ -697,7 +691,7 @@ pub fn compileLetStarValues(self: *Compiler, args: Value, dst: u16, is_tail: boo
 /// identifier resolution can be shadowed by a top-level redefinition of
 /// `call-with-values` (#1715) — so it takes the same true-binding reference
 /// unconditionally rather than relying on which path the compiler picks.
-pub fn buildLetValues(self: *Compiler, bindings: Value, body: Value, is_tail: bool) !Value {
+pub fn buildLetValues(self: *Compiler, bindings: Value, body: Value) !Value {
     const gc = self.gc;
     gc.no_collect += 1;
     defer gc.no_collect -= 1;
@@ -718,7 +712,7 @@ pub fn buildLetValues(self: *Compiler, bindings: Value, body: Value, is_tail: bo
     if (!types.isPair(expr_rest)) return error.InvalidSyntax;
     const expr = types.car(expr_rest);
 
-    const inner = try buildLetValues(self, rest_bindings, body, true);
+    const inner = try buildLetValues(self, rest_bindings, body);
 
     // (lambda () expr)
     const producer_body = try gc.allocPair(expr, types.NIL);
@@ -729,9 +723,9 @@ pub fn buildLetValues(self: *Compiler, bindings: Value, body: Value, is_tail: bo
     const consumer_lambda = try gc.allocPair(lambda_sym, try gc.allocPair(formals, consumer_body));
 
     // (call-with-values producer consumer)
-    const cwv_sym = if (is_tail)
-        try gc.allocSymbol("call-with-values")
-    else
-        try compiler_mod.Compiler.trueBuiltinRefOrSymbol(gc, "call-with-values");
+    // A compiler-synthesized reference: always the pristine (scheme base)
+    // binding, so no top-level redefinition of `call-with-values` can change
+    // the desugaring's meaning (#1715, #2033).
+    const cwv_sym = try compiler_mod.Compiler.trueBuiltinRefOrSymbol(gc, "call-with-values");
     return try gc.allocPair(cwv_sym, try gc.allocPair(producer_lambda, try gc.allocPair(consumer_lambda, types.NIL)));
 }
--- a/src/compiler_bindings.zig
+++ b/src/compiler_bindings.zig
@@ -679,18 +679,14 @@ pub fn compileLetStarValues(self: *Compiler, args: Value, dst: u16, is_tail: boo
 ///   (lambda (a b)
 ///     (call-with-values (lambda () e2) (lambda (c) (begin body)))))
 ///
-/// Each *nested* call-with-values (built by the recursive call below) is
-/// the sole body of its enclosing consumer lambda, so it is always in tail
-/// position, hence always compiled through compileCallWithValuesTail's
-/// bytecode fast path — which avoids both native re-entrancy and (since
-/// #1715) shadowing, as it now loads the true `(scheme base)` binding
-/// directly instead of resolving `call-with-values` by name. Only the
-/// OUTERMOST call-with-values (returned to the original, non-recursive
-/// caller) has tail-ness that depends on this let*-values's own is_tail;
-/// when that's false, it falls through to ordinary call compilation, whose
-/// identifier resolution can be shadowed by a top-level redefinition of
-/// `call-with-values` (#1715) — so it takes the same true-binding reference
-/// unconditionally rather than relying on which path the compiler picks.
+/// Every call-with-values built here is a compiler-synthesized reference to
+/// the true `(scheme base)` binding (base-binding-prefixed, #1715): the
+/// nested ones sit in tail position and compile through
+/// compileCallWithValuesTail's bytecode fast path — which avoids both
+/// native re-entrancy and (since #2033) the redefinition gate, as the
+/// dispatch recognizes the prefixed spelling as immune — and any that fall
+/// through to ordinary call compilation resolve the pristine binding
+/// regardless of how the program rebinds `call-with-values`.
 pub fn buildLetValues(self: *Compiler, bindings: Value, body: Value) !Value {
     const gc = self.gc;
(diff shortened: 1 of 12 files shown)


src/llvm_emit_forms.zig
```diff
@@ -1044,7 +1046,7 @@ pub fn emitApplyForm(self: *LLVMEmitter, expr: Value, is_tail: bool) EmitError![
     if (!shadowed and is_tail and operands.items.len < 2)
         return error.UnsupportedNodeType;
 
-    const is_builtin_apply = !shadowed and (is_tail or !rebound);
+    const is_builtin_apply = !shadowed and !rebound;
```
baijum, line 1060: Minor (behavior-preserving, non-blocking): the `<2 operands` early return three lines up —

```zig
if (!shadowed and is_tail and operands.items.len < 2)
    return error.UnsupportedNodeType;
```

fires *before* the `rebound` check, so it also triggers for a **rebound** `apply` in tail position with a single operand (e.g. `(lambda (f) (apply f))` where `apply` was redefined to a 1-arg proc). In that case the updated doc/comment says "a rebound apply (tail or non-tail) → ordinary indirect call," but the code instead abandons native compilation of the whole enclosing scope. The result stays correct because the eval fallback hands it to the (now redefinition-honoring) interpreter, so this is purely a stale-rationale + needless de-opt, not a correctness bug. If you want the comment/doc to match the code exactly, either gate the early return on `!rebound` (letting the generic indirect path at ~L1096 handle it) or narrow the doc bullet to the unrebound case.

src/compiler_bindings.zig
```diff
@@ -697,7 +691,7 @@ pub fn compileLetStarValues(self: *Compiler, args: Value, dst: u16, is_tail: boo
 /// identifier resolution can be shadowed by a top-level redefinition of
 /// `call-with-values` (#1715) — so it takes the same true-binding reference
 /// unconditionally rather than relying on which path the compiler picks.
-pub fn buildLetValues(self: *Compiler, bindings: Value, body: Value, is_tail: bool) !Value {
+pub fn buildLetValues(self: *Compiler, bindings: Value, body: Value) !Value {
```
baijum, line 690: Nit: this doc comment still explains the old `is_tail`-dependent reasoning ("Only the OUTERMOST call-with-values ... has tail-ness that depends on this let*-values's own is_tail; when that's false, it falls through to ordinary call compilation, whose identifier resolution can be shadowed ..."), but `buildLetValues` no longer takes `is_tail` and now emits the base-prefixed true-binding reference unconditionally. The final sentence captures the new behavior, but the middle paragraph describes a code path that no longer exists — worth trimming so it doesn't mislead a future reader.

baijum (commented):
Reviewed the full diff. This is a solid, well-scoped fix with strong parity discipline between the interpreter gate and the native backend, and thorough tests (five names × tail/non-tail, non-procedure rebind, hygiene-renamed template, `set!`-in-form suppression, and desugaring immunity). A few things I verified and two minor notes below.

**Verified sound:**
- `globalBindingStillGenuine` faithfully mirrors `IR.isRedefined`'s gate order (`set_targets_all` → `set_targets` → null-globals-optimistic → restricted-env-unbound → native_fn name match), and its `!restricted_env` unbound arm is the correct logical inverse of `isRedefined`'s `restricted_env`. `call/cc` and `call-with-current-continuation` are registered as separate specs, so each `native_fn.name` matches the `eff_name` passed in — no spurious fast-path decline for the common case.
- The `is_synth` bypass of the local/upvalue + global gate is correct: a `__kaappi_base__` symbol is never a lexical binding and always means the pristine `(scheme base)` procedure. The `ir.zig` passthrough-for-base-prefixed-special-form-head change is what keeps those synthesized references reaching `compileForm` instead of bypassing the fast path as a plain call.
- Native side: `(define apply myapply)` (symbol form) leaves `apply` in `rebound_globals` (the `fetchRemove` at `emitDefine` only fires for the lambda shape), so `is_builtin_apply = !shadowed and !rebound` correctly routes the `rebound.scm` case to an indirect call → `99/99`. The synthesized base-prefixed `apply`/`call-with-values` never reach `emitApplyForm` because every form that mints them (`let-values`/`let*-values`/`define-values`/`case-lambda`/`define-record-type`) is an eval-fallback form in the native backend — so the literal `"apply"` match at `emitPassthrough` and `emitApplyForm` stays correct.

Note: I reviewed statically and did not rebuild; trusting the PR's stated green run (700 Scheme files / 1395 R7RS / unit + gc-stress / WASM / native tier).

tests/scheme/compliance/top-level-redefinition-2033.scm
```diff
@@ -0,0 +1,192 @@
...
+(test-equal "case-lambda desugaring keeps pristine apply (rest-arg arm)"
+            '(5) ((case-lambda ((a b) (+ a b)) (a a)) 5))
+(test-equal "define-record-type desugaring keeps pristine apply"
+            3 (let ()
+                (define-record-type <t2033-pt> (make-t2033-pt x y) t2033-pt?
+                  (x t2033-pt-x) (y t2033-pt-y))
+                (define-record-type <t2033-pt3> (make-t2033-pt3 x y z) t2033-pt3?
+                  (x t2033-pt3-x) (y t2033-pt3-y) (z t2033-pt3-z)
+                  (parent <t2033-pt>))
+                (t2033-pt3-z (make-t2033-pt3 1 2 3))))
```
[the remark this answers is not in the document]
baijum: Addressed in 8e09371e. You were right: the original test used R7RS constructor-first syntax, so `(parent <t2033-pt>)` parsed as a field spec and the changed R6RS inherited-constructor paths in vm_records (L673 protocol, L709 protocol-less) were never exercised. Replaced it with top-level R6RS clause-syntax records covering both inherited-ctor variants, verified failing on main (bare `apply` resolves to the redefinition and record construction breaks with a %record-ref type error) and passing here. Also converted the remaining bare synthesized `list`/`append`/`car`/`cdr` refs in those same paths to base-binding-prefixed, same bug class.

baijum:
Reviewed commit `8e09371e` — all three findings resolved cleanly:

1. **Native `<2 operands` gate** — now `if (!shadowed and !rebound and is_tail and operands.items.len < 2)`, so a rebound `apply` with too few operands in tail position falls through to the generic indirect call (matching the interpreter's #2033-gated ordinary-call path) instead of needlessly abandoning native compilation. The `llvm_emit_forms.zig` comment and `llvm-backend.md` bullet now spell out the rebound sub-case. ✅

2. **`buildLetValues` doc** — trimmed to the base-prefixed unconditional-reference behavior; the stale `is_tail`-dependent outermost-vs-nested reasoning is gone. ✅

3. **(CodeRabbit) R6RS record coverage** — good catch that the old R7RS constructor-first record parsed `(parent …)` as a *field* spec and never reached `handleDefineRecordTypeR6RS`. The replacement R6RS clause-syntax records now exercise both inherited-ctor paths (protocol-less split-args and protocol), and I confirmed the accompanying audit is complete: every procedure-valued synthesized reference in those paths (`list`/`append`/`car`/`cdr`/`apply`/`%make-record`/`%record-ref`/`%record-split-args`) is now base-binding-prefixed. The only remaining bare `allocSymbol` calls are internal gensyms and the `lambda`/`let`/`let*` keywords, which are structural and out of scope for #2033. ✅

LGTM. (Static review only — not rebuilt; relying on the stated green CI run.)

PR #2265, opened 2026-08-09
REPL: click inside the input to move the edit cursor (#2264)

### LINKED ISSUE #2264: REPL: click to reposition the edit cursor (mouse support)
## Summary

Allow a left mouse click in the REPL to reposition the edit cursor within the
current input, instead of only supporting arrow keys / structural editing.

## Why this is a patch to isocline, not `repl.zig`

The REPL delegates all input reading to vendored isocline (`ic_readline`).
Isocline owns raw mode (`tty.c`), the CSI escape decoder (`tty_esc.c`), the edit
buffer, redraw, and cursor placement (`editline.c`). There is a single `read()`
on stdin and isocline owns it, so `repl.zig` cannot run a competing read loop.
Mouse support has to be a **fifth isocline patch** (joining the four in
`vendor/isocline/PATCHES.md`), re-applied on every isocline upgrade.

## What the patch involves

1. **Enable mouse tracking** on raw-mode entry / disable on exit (`tty.c`):
   emit `\x1b[?1000h` (button-press tracking; deliberately *not* `?1002h`/`?1003h`
   motion) and `\x1b[?1006h` (SGR extended coordinates).
2. **Decode SGR mouse in `tty_esc.c`** — `\x1b[<b;x;yM`/`m` starts with `[` then
   `<`, which the current CSI decoder treats as a "special" byte and mis-handles.
   Intercept `<` and produce a new mouse event carrying (button, x, y) rather than
   forcing it into the `code_t` keycode space (mind the virtual-key range in
   `tty.h`).
3. **Handle it in the edit loop** — on left-press, translate the click to an
   editor-relative (row, col) and call the existing
   `edit_set_pos_at_rowcol(env, eb, row, col)` (`editline.c:151`). The hard part
   — mapping (row, col) → buffer byte position with prompt width, continuation
   prompt, and line wrapping — is already implemented (`sbuf_get_pos_at_rc`).

## The one hard part: absolute → relative anchor

The mouse reports **absolute** screen coordinates; isocline works entirely in
coordinates **relative to the prompt** (`eb->cur_row` is 0-based relative to the
prompt, `editline.c:34`) and never queries the absolute cursor position — `term.c`
has no `ESC[6n`/DSR reader. So the patch must anchor the input's on-screen start,
via a one-time DSR query at readline start (new machinery). This anchor is the
only piece that must be correct; everything downstream is already safe.

## Clicks outside the editing area

`edit_set_pos_at_rowcol` already guards `if (pos < 0) return;`, and the mapper
returns `-1` for out-of-range rows, so most out-of-area clicks are safe no-ops:

| Click location | Result |
|---|---|
| Below the input (blank rows) | row out of range → `-1` → cursor doesn't move |
| Above the input (prior output / scrollback) | negative relative row → `-1` → no-op *(iff the anchor is right)* |
| On the prompt glyphs | clamps to start of that logical line's content |
| Past end of a line's text | `i < end` guard clamps to line end |
| Inside text (incl. wrapped lines) | maps exactly (mapper is wrap-aware) |

Clicking into scrollback can never reposition there: that region is terminal
history, not an edit buffer, and isocline has no model of it. The whole risk
concentrates in the absolute→relative anchor — a stale/off-by-one anchor could
map an above-input click to a valid in-range row and jump the cursor
unexpectedly, instead of no-op'ing.

## Terminal selection tradeoff

`?1000h` captures clicks, so drag-to-select-and-copy in the REPL stops working.
This is the usual reason REPLs don't do this. Mitigation: mainstream terminals
bypass app mouse mode with a modifier for one-off native selection (Option-drag
in Terminal.app / iTerm2, Shift-drag in most Linux terminals), so copy still
works, modifier-gated. Preserving un-gated selection would require shift-click
passthrough logic — out of scope for a first cut.

## Proposal

Ship behind a `repl.mouse` config setting (default **off**), so the selection
behavior only changes for users who opt in.

## Portability

SGR mouse (`?1006h`) covers the SGR-capable terminals. The Windows console needs
the separate console mouse-input path, not `?1006h` — the four existing patches
already carry a Windows burden; this adds another.

## Key references

- `vendor/isocline/PATCHES.md` — the four existing patches; this becomes #5
- `vendor/isocline/src/editline.c:151` — `edit_set_pos_at_rowcol`
- `vendor/isocline/src/stringbuf.c:534` — `str_get_pos_at_rc` (clamping/fail-safe)
- `vendor/isocline/src/tty_esc.c` — CSI decoder to extend
- `docs/dev/repl.md` — REPL / isocline integration overview

baijum:
## Summary

Fixes #2264 — click inside the REPL input to reposition the edit cursor (SGR mouse support).

Shipped behind **`repl.mouse: true`** in `~/.kaappi/config` (default **off**), so the selection tradeoff — `?1000h` captures clicks, which stops un-gated drag-to-select while enabled — only applies to users who opt in. This is the **fifth Kaappi patch to vendored isocline** (`vendor/isocline/PATCHES.md`), following the pattern of the other four (re-applied on every isocline upgrade; each site marked `KAAPPI PATCH 5`).

## What it does

- **Tracking** (`ic_editline`): emits `\x1b[?1000h\x1b[?1006h` on raw-mode entry, `l` sequences on exit. Deliberately `?1000h` (button presses), not `?1002h`/`?1003h` (motion).
- **Decode** (`tty_esc.c`): intercepts `ESC[<b;x;yM|m` right after the CSI decoder's special-byte check (the generic parser only handles two parameters). The event is stashed on the tty via `tty_set_mouse_event` and surfaced as `KEY_EVENT_MOUSE`, since two coordinates cannot fit the `code_t` keycode space.
- **Anchor** (`editline.c`): the mouse reports absolute screen coordinates while isocline works relative to the prompt, so the editor issues a one-time `ESC[6n` query at edit-session start and maps clicks through the existing `edit_set_pos_at_rowcol` / `sbuf_get_pos_at_rc` machinery (prompt width, continuation prompt, line wrapping). Terminals that don't answer the DSR leave the anchor unset → clicks are no-ops.
- **Safety**: clicks above (scrollback), below, or on the prompt are no-ops or clamp to line content; a delayed DSR response now decodes to `KEY_NONE`, which the edit loop ignores instead of inserting a NUL.

## Windows (not in this PR)

Per the issue and its author comment: Windows reads `INPUT_RECORD` structs, not a byte stream, so SGR has nothing to decode there. It needs a second, parallel Console-API implementation (`MOUSE_EVENT` arm + `ENABLE_MOUSE_INPUT`, with `GetConsoleScreenBufferInfo` for the anchor). That path is documented in `PATCHES.md` as the follow-up; tracking is `#if !defined(_WIN32)` gated until then. It also has no test harness here (REPL pty tests skip on Windows), which is the other reason to land POSIX first.

## Tests

- New `tests/scheme/smoke/repl-mouse-click-2264.sh` — pty test that plays the terminal emulator: answers the DSR query, feeds SGR presses, and asserts on what the *evaluator* prints (single-line click, continuation-row click, plus a mouse-off run proving the SGR bytes are inert by default).
- `src/config.zig` unit tests for `repl.mouse` parsing/defaults.
- Also fixes the stale "three patches" count in the `src/isocline.zig` module doc (PATCHES.md had already outgrown it).

Full suite green locally: `zig build test` + `tests/scheme/run-all.sh` (2095 pass, 0 fail, including the new test).

## Docs follow-up (separate repo)

The end-user guide (`kaappi.github.io/docs/guide/repl.md`) should get a short `repl.mouse` mention in a follow-up PR to the docs repo.

### DIFF
--- a/src/repl.zig
+++ b/src/repl.zig
@@ -615,6 +615,10 @@ pub fn repl(vm: *vm_mod.VM) !void {
         ic.enableBraceInsertion(false);
         ic.enableHighlight(highlight_enabled);
         ic.enableHint(false); // no hint callback is wired up
+        // Opt-in SGR mouse tracking: click inside the input to move the edit
+        // cursor (kaappi#2264). Off by default — while tracking is on, the
+        // terminal stops reporting drag-to-select to the application.
+        ic.enableMouse(cfg.mouse);
         applyTheme(cfg.theme);
 
         ic.setIsComplete(&isCompleteCallback, null);
--- a/vendor/isocline/src/editline.c
+++ b/vendor/isocline/src/editline.c
@@ -194,7 +194,7 @@ static void edit_mouse_click(ic_env_t* env, editor_t* eb) {
   tty_mouse_t m;
   if (!tty_last_mouse(env->tty, &m)) return;
   if (!eb->has_anchor) return;
-  if (m.button != 0) return;                    // only a plain left press moves the cursor
+  if (m.button != 0 || !m.press) return;   // only a plain left press moves the cursor
   if (m.row < 1 || m.col < 1) return;
   const ssize_t row = m.row - eb->anchor_row;   // 0-based, relative to the prompt
   if (row < 0) return;                          // click above the input: scrollback, no-op
@@ -944,22 +944,17 @@ static char* edit_line( ic_env_t* env, const char* prompt_text )
   // query (ESC[6n). The mouse reports absolute screen coordinates; isocline
   // works relative to the prompt, so this anchor is the only mapping piece —
   // everything downstream (prompt width, continuation prompt, line wrapping)
-  // is already handled by `sbuf_get_pos_at_rc`. A terminal that does not
-  // answer (some emulators, SSH without mouse support) just leaves the anchor
-  // unset and clicks become no-ops. See PATCHES.md.
+  // is already handled by `sbuf_get_pos_at_rc`. The response reader never
+  // consumes input: typing ahead between forms is common in a REPL, and any
+  // bytes that are not a well-formed response are pushed back so the edit
+  // loop still reads them as keys. A terminal that does not answer just
+  // leaves the anchor unset and clicks become no-ops. See PATCHES.md.
   #if !defined(_WIN32)
   if (env->mouse) {
     term_writef(env->term, "\x1B[6n");
     term_flush(env->term);
-    char dsrbuf[128];
-    if (tty_read_esc_response(env->tty, '[', false, dsrbuf, sizeof(dsrbuf))) {
-      ssize_t crow = 0;
-      ssize_t ccol = 0;
-      if (ic_atoz2(dsrbuf, &crow, &ccol)) {
-        eb.anchor_row = crow;
-        eb.anchor_col = ccol;
-        eb.has_anchor = true;
-      }
+    if (tty_read_dsr_response(env->tty, &eb.anchor_row, &eb.anchor_col)) {
+      eb.has_anchor = true;
     }
   }
   #endif
--- a/vendor/isocline/src/editline.c
+++ b/vendor/isocline/src/editline.c
@@ -40,6 +40,13 @@ typedef struct editor_s {
   editstate_t*  redo;         // redo buffer
   const char*   prompt_text;  // text of the prompt before the prompt marker    
   alloc_t*      mem;          // allocator
+  // KAAPPI PATCH 5: absolute anchor of the input's on-screen start, queried
+  // once per edit session via ESC[6n. The mouse reports absolute screen
+  // coordinates; everything else in isocline is relative to the prompt, so
+  // this anchor is the only mapping piece. See PATCHES.md.
+  ssize_t       anchor_row;   // absolute 1-based row where the prompt starts
+  ssize_t       anchor_col;   // absolute 1-based col where row 0's content starts
+  bool          has_anchor;   // true once the DSR query has answered
   // caches
   attrbuf_t*    attrs;        // reuse attribute buffers 
   attrbuf_t*    attrs_extra; 
@@ -58,7 +65,24 @@ static void edit_refresh(ic_env_t* env, editor_t* eb);
 ic_private char* ic_editline(ic_env_t* env, const char* prompt_text) {
   tty_start_raw(env->tty);
   term_start_raw(env->term);
+  // KAAPPI PATCH 5: opt-in mouse tracking for click-to-position (see
+  // PATCHES.md). ?1000h = button presses (deliberately not ?1002h/?1003h
+  // motion), ?1006h = SGR extended coordinates. The Windows console needs its
+  // own mouse-input path, not SGR — out of scope, so tracking stays off there.
+  #if !defined(_WIN32)
+  const bool mouse_tracking = env->mouse;
+  if (mouse_tracking) {
+    term_writef(env->term, "\x1B[?1000h\x1B[?1006h");
+    term_flush(env->term);
+  }
+  #endif
   char* line = edit_line(env,prompt_text);
+  #if !defined(_WIN32)
+  if (mouse_tracking) {
+    term_writef(env->term, "\x1B[?1000l\x1B[?1006l");
+    term_flush(env->term);
+  }
+  #endif
   term_end_raw(env->term,false);
   tty_end_raw(env->tty);
   term_writeln(env->term,"");
@@ -155,6 +179,31 @@ static void edit_set_pos_at_rowcol( ic_env_t* env, editor_t* eb, ssize_t row, ss
   if (pos < 0) return;
   eb->pos = pos;
   edit_refresh(env, eb);
+}
+
+// KAAPPI PATCH 5: click to reposition the cursor (see PATCHES.md). The mouse
+// reports absolute screen coordinates; the anchor (queried once at edit
+// session start) converts them to prompt-relative row/col, which
+// `sbuf_get_pos_at_rc` maps to a buffer position with prompt width,
+// continuation prompt, and line wrapping all taken into account. Clicks
+// outside the editing area are safe no-ops: rows above the input are negative
+// (scrollback), rows below are out of range, both of which `sbuf_get_pos_at_rc`
+// answers with -1.
+static void edit_mouse_click(ic_env_t* env, editor_t* eb) {
+  if (!env->mouse) return;
+  tty_mouse_t m;
+  if (!tty_last_mouse(env->tty, &m)) return;
+  if (!eb->has_anchor) return;
+  if (m.button != 0) return;                    // only a plain left press moves the cursor
+  if (m.row < 1 || m.col < 1) return;
+  const ssize_t row = m.row - eb->anchor_row;   // 0-based, relative to the prompt
+  if (row < 0) return;                          // click above the input: scrollback, no-op
+  ssize_t promptw, cpromptw;
+  edit_get_prompt_width(env, eb, false, &promptw, &cpromptw);
+  ssize_t col = m.col - eb->anchor_col;
+  if (row > 0) col += (promptw - cpromptw);     // continuation rows start at the cpromptw column
+  if (col < 0) col = 0;                         // click on the prompt glyphs: clamp to content start
+  edit_set_pos_at_rowcol(env, eb, row, col);
 }
 
 static bool edit_pos_is_at_row_end( ic_env_t* env, editor_t* eb ) {
@@ -891,6 +940,30 @@ static char* edit_line( ic_env_t* env, const char* prompt_text )
   // show prompt
   edit_write_prompt(env, &eb, 0, false);   
 
+  // KAAPPI PATCH 5: anchor the input's on-screen start with a one-time DSR
+  // query (ESC[6n). The mouse reports absolute screen coordinates; isocline
+  // works relative to the prompt, so this anchor is the only mapping piece —
+  // everything downstream (prompt width, continuation prompt, line wrapping)
+  // is already handled by `sbuf_get_pos_at_rc`. A terminal that does not
+  // answer (some emulators, SSH without mouse support) just leaves the anchor
+  // unset and clicks become no-ops. See PATCHES.md.
+  #if !defined(_WIN32)
+  if (env->mouse) {
+    term_writef(env->term, "\x1B[6n");
+    term_flush(env->term);
+    char dsrbuf[128];
+    if (tty_read_esc_response(env->tty, '[', false, dsrbuf, sizeof(dsrbuf))) {
+      ssize_t crow = 0;
+      ssize_t ccol = 0;
+      if (ic_atoz2(dsrbuf, &crow, &ccol)) {
+        eb.anchor_row = crow;
+        eb.anchor_col = ccol;
+        eb.has_anchor = true;
+      }
+    }
+  }
+  #endif
+
   // always a history entry for the current input
   history_push(env->history, "");
 
@@ -982,6 +1055,10 @@ static char* edit_line( ic_env_t* env, const char* prompt_text )
         break;
       case KEY_EVENT_AUTOTAB:
         edit_generate_completions(env, &eb, true);
+        break;
+      // KAAPPI PATCH 5: click to reposition the cursor (see PATCHES.md)
+      case KEY_EVENT_MOUSE:
+        edit_mouse_click(env,&eb);
         break;
 
       // completion, history, help, undo
@@ -1123,7 +1200,12 @@ static char* edit_line( ic_env_t* env, const char* prompt_text )
       default: {
         char chr;
         unicode_t uchr;
-        if (code_is_ascii_char(c,&chr)) {
+        if (c == KEY_NONE) {
+          // KAAPPI PATCH 5: a DSR or other query response that arrived after
+          // its query timed out decodes to KEY_NONE; without this guard it
+          // would fall through to `code_is_unicode(0)` and insert a NUL.
+        }
+        else if (code_is_ascii_char(c,&chr)) {
           edit_insert_char(env,&eb,chr);
         }
         else if (code_is_unicode(c, &uchr)) {
--- a/vendor/isocline/src/tty.c
+++ b/vendor/isocline/src/tty.c
@@ -11,6 +11,7 @@
 #include <locale.h>
 
 #include "tty.h"
+#include "stringbuf.h"  // ic_atoz2 (KAAPPI PATCH 5)
 
 #if defined(_WIN32)
 #include <windows.h>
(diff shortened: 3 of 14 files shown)


src/repl.zig
```diff
@@ -615,6 +615,10 @@ pub fn repl(vm: *vm_mod.VM) !void {
         ic.enableBraceInsertion(false);
         ic.enableHighlight(highlight_enabled);
         ic.enableHint(false); // no hint callback is wired up
+        // Opt-in SGR mouse tracking: click inside the input to move the edit
+        // cursor (kaappi#2264). Off by default — while tracking is on, the
+        // terminal stops reporting drag-to-select to the application.
+        ic.enableMouse(cfg.mouse);
```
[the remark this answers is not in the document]
baijum: Thanks — both test comments are fixed in 3e197678: the `typed`→`send_bytes` NameError on the timeout path, and a new narrow-pty (20-column) run that clicks an automatically wrapped row of `(list 1 2 3 4 5 6)` and asserts the wrapped-row mapping (`(1 2 3 94 5 6)` vs a plain end-insert).

On the `repl.zig` split: agreed it's over the 1,500-line guideline, but that predates this PR (1,588 lines before my +4) and splitting the REPL loop + its interwoven callbacks is a refactor with its own review surface. I'd rather land it as a focused follow-up than bundle it into a feature PR. Tracked here: #2266.

vendor/isocline/src/editline.c
```diff
@@ -891,6 +940,30 @@ static char* edit_line( ic_env_t* env, const char* prompt_text )
...
+  // everything downstream (prompt width, continuation prompt, line wrapping)
+  // is already handled by `sbuf_get_pos_at_rc`. A terminal that does not
+  // answer (some emulators, SSH without mouse support) just leaves the anchor
+  // unset and clicks become no-ops. See PATCHES.md.
+  #if !defined(_WIN32)
+  if (env->mouse) {
+    term_writef(env->term, "\x1B[6n");
+    term_flush(env->term);
+    char dsrbuf[128];
+    if (tty_read_esc_response(env->tty, '[', false, dsrbuf, sizeof(dsrbuf))) {
```
baijum: **Type-ahead here is silently lost (main concern).** This DSR query runs on every `ic_readline`. If the user typed ahead between forms (common in a REPL — hit Enter, start typing the next form while the previous evaluates), those bytes are already in the tty buffer. `tty_read_esc_response` reads the first one, sees it isn't `ESC`, and `return false` — but the byte is already consumed (tty.c:260), not pushed back. So the first character of the next input is silently dropped, and since the call returned false the anchor is left unset for that line too. The leftover real `ESC[...R` then sits in the queue and decodes to `KEY_NONE` (which your new guard now swallows — so this is partly why that guard was needed).

This is the one behavior change for `repl.mouse: true` users that isn't in the documented limitations. Options: push the non-`ESC` leading byte back with `tty_cpush_char` before returning (in a mouse-local reader), or only issue the query when the input queue is empty. Secondary: on terminals that never answer, this blocks up to `2*esc_initial_timeout` on *every* prompt — worth a mention even though it's opt-in.

vendor/isocline/src/tty_esc.c
```diff
@@ -257,6 +257,35 @@ static code_t tty_read_csi(tty_t* tty, uint8_t c1, uint8_t peek, code_t mods0, l
...
+    if (peek == ';') {
+      if (!tty_readc_noblock(tty,&peek,esc_timeout)) return KEY_NONE;
+      tty_read_csi_num(tty,&peek,&x,esc_timeout);
+    }
+    if (peek == ';') {
+      if (!tty_readc_noblock(tty,&peek,esc_timeout)) return KEY_NONE;
+      tty_read_csi_num(tty,&peek,&y,esc_timeout);
+    }
+    uint8_t final = peek;   // 'M' = press, 'm' = release
+    if (final == 'M' || final == 'm') {
```
baijum, line 280: `?1000h` delivers both press (`M`) and release (`m`), and both are surfaced as `KEY_EVENT_MOUSE` with button 0 (SGR release keeps the button number, only the final flips to `m`). Since the `M`/`m` distinction is decoded but not stored, `edit_mouse_click` repositions on both — a single click runs the move twice. Idempotent, so harmless, but you could return early / mark it a release for `m` so only the press moves the cursor and you skip the redundant `edit_refresh`.

vendor/isocline/src/editline.c
```diff
@@ -891,6 +940,30 @@ static char* edit_line( ic_env_t* env, const char* prompt_text )
...
+  // KAAPPI PATCH 5: anchor the input's on-screen start with a one-time DSR
+  // query (ESC[6n). The mouse reports absolute screen coordinates; isocline
+  // works relative to the prompt, so this anchor is the only mapping piece —
+  // everything downstream (prompt width, continuation prompt, line wrapping)
+  // is already handled by `sbuf_get_pos_at_rc`. A terminal that does not
+  // answer (some emulators, SSH without mouse support) just leaves the anchor
+  // unset and clicks become no-ops. See PATCHES.md.
+  #if !defined(_WIN32)
+  if (env->mouse) {
+    term_writef(env->term, "\x1B[6n");
```
baijum, line 954: Minor / maintainability: this inline write + `tty_read_esc_response` + `ic_atoz2` reimplements what `term_esc_query_raw` + `ic_atoz2` already do in term.c. `term_esc_query_raw` (unlike `term_esc_query`) does *not* toggle raw mode, so it's usable here — the PATCHES.md note about `term_esc_query` being unusable is true of the toggling wrapper, not the `_raw` one. Exposing a small non-toggling cursor-pos helper and calling it would keep the DSR round-trip in one place to re-apply on isocline upgrades. Non-blocking.

baijum (commented):
Reviewed the full diff against the local worktree. This is a well-scoped, well-documented patch — default-off gating is right, `PATCHES.md` is excellent (it even records the scroll-staleness limitation and the full Windows follow-up plan), and the pty test is genuinely clever: asserting on the *evaluator's* output rather than the redraw sidesteps the usual terminal-test brittleness, and the mouse-off run proving the SGR bytes are inert is exactly the right negative case.

One issue I'd want addressed before merge, plus two minor notes — all inline.

**1. Type-ahead is silently eaten by the per-prompt `ESC[6n` (main concern).** The anchor query runs on *every* `ic_readline`, and `tty_read_esc_response` consumes the first queued byte and bails if it isn't `ESC`. In a REPL, typing ahead between forms is common — press Enter, start typing the next form while the previous one evaluates. Those bytes sit in the tty buffer; the DSR reader then eats the first one (and, because it returns false, also leaves the anchor unset for that line). Result: silent loss of the first character of the next input, for `repl.mouse: true` users. This is the one behavior change not covered in the documented limitations. Detail on the editline.c line.

**2. Press *and* release both reposition** (minor) — `?1000h` reports release too, and the `M`/`m` final is decoded but discarded, so a single click runs the move twice. Harmless (idempotent), but acting only on `M` would avoid the redundant refresh.

**3. The anchor reimplements existing DSR machinery** (minor / maintainability) — `term_esc_query_raw` already does write + `tty_read_esc_response` without toggling raw mode, which is exactly what's needed here. Reusing it keeps the DSR path in one place across isocline upgrades.

Nothing here blocks the design — (1) is the one I'd fix before merge.

vendor/isocline/src/editline.c
baijum: Fixed in 508cd427. The anchor query no longer uses `tty_read_esc_response`; a new `tty_read_dsr_response` (tty.c) accepts only a well-formed `ESC [ digits ; digits R` and **pushes back every byte it read** on any failure path, in order. A queued keystroke or an ESC-sent key sequence (arrows, Alt+key) still reaches the edit loop; the only consequence of an unreadable response is an unset anchor (clicks no-op for that line) — never lost input. A response that arrives after the reader gave up decodes to `KEY_NONE` and is ignored by the guard added earlier.

The pty test has a fourth run for exactly this: it submits `(+ 1 2)`, immediately types `(list 9 8)` with no idle in between (while the first form is still evaluating), and asserts both results print. Verified it fails without the push-back.

vendor/isocline/src/tty_esc.c
baijum: Fixed in 508cd427: the mouse event now carries a `press` flag (set from the `M`/`m` final in tty_esc.c) and `edit_mouse_click` repositions on press only — a single click under `?1000h` moves the cursor exactly once instead of twice.

vendor/isocline/src/editline.c
baijum: Agreed the write+read was a reimplementation — 508cd427 replaces it with the dedicated `tty_read_dsr_response`. I deliberately did not route through `term_esc_query_raw`: it can't validate the `R` terminator or push back bytes consumed on a failed read, which is exactly what the type-ahead fix (the main review point) needs. `term_esc_query_raw` also toggles nothing, but its caller `term_esc_query` pairs it with `tty_start_raw`/`tty_end_raw`, and `tty_end_raw` clears the low-level pushback buffer — so a push-back inside the shared path would be thrown away there. The dedicated reader keeps DSR parsing in one place (tty.c, next to `tty_read_esc_response`) and leaves the shared response reader untouched for term.c's other queries.

vendor/isocline/src/tty.c
```diff
@@ -227,6 +233,78 @@ ic_private code_t tty_read(tty_t* tty)
...
+      }
+      break;                               // malformed payload: restore below
+    }
+    if ((c < '0' || c > '9') && c != ';') break;  // not a DSR: restore below
+    if (n >= (ssize_t)sizeof(buf) - 1) break;     // absurdly long: restore
+    buf[n++] = (char)c;
+  }
+  // push back everything we read, in order (the low-level buffer pops LIFO)
+  tty_cpush_char(tty, c);
+  while (n > 0) { n--; tty_cpush_char(tty, buf[n]); }
```
baijum: **This push-back can overflow `cpushbuf` (out-of-bounds write).** `cpushbuf` is `TTY_PUSH_MAX` = 32 bytes (tty.c:37,48), but this path pushes back up to `c` + `n` digits + `[` + `ESC`, and `n` is bounded only by `buf[64]` (`n >= sizeof(buf)-1` → up to 63 at line 297). So a restore can push **up to 66 bytes** into a 32-byte buffer.

Normally `tty_cpush` would guard this, but its check is on the **wrong counter**:

```c
if (tty->push_count + len > TTY_PUSH_MAX) { ... return; }  // push_count = high-level pushbuf
tty->cpushbuf[tty->cpush_count + i] = ...;                 // but writes cpushbuf via cpush_count
```

`push_count` (the `pushbuf` code buffer) is ~0 during a DSR read, so the guard never trips while `cpush_count` runs past 32 — the writes land in the `tty_s` struct past `cpushbuf` (into `esc_*`/mouse fields). And because the guard tests the wrong counter, the `assert(false)` never fires, so there's no debug safety net either. (The mismatched guard is pre-existing upstream, but this reader is the first caller that can push back more than a couple of bytes, so it's exposed here.)

**Trigger:** any CSI with ~29+ `0-9`/`;` bytes not terminated by `R`, seen while `repl.mouse` is on — a garbled/hostile terminal report, a long device report from a multiplexer, or pasted bytes containing such a sequence. Narrow, but it's an unconditional OOB write once hit.

**Fix:** cap the digit buffer and total push-back to stay within `TTY_PUSH_MAX` — e.g. size `buf` so `1 + n + 2 <= TTY_PUSH_MAX` (n ≤ 29), and bail to the restore path at that bound. A real cursor report is only a handful of digits, so a tight cap costs nothing. Optionally also fix the `tty_cpush` guard to test `cpush_count`, but that's the broader upstream bug.

baijum (commented):
Re-reviewed commits `508cd42` and `1a0eb9c`. Both earlier points are addressed well:

- **Type-ahead** — the new `tty_read_dsr_response` accepts only a well-formed `ESC [ digits ; digits R` and pushes back everything else. I checked the push-back ordering against `tty_cpop`: it's LIFO (`cpush_count--`), and the reverse-order re-push (`c`, digits reversed, `[`, ESC) reconstructs the original byte order correctly. The dedicated type-ahead pty run (`ahead`) is a good regression.
- **Press-only** — `m.press` is now carried and `edit_mouse_click` gates on it. Correct.

**But the push-back path can overflow `cpushbuf` (new issue, memory safety).** Details inline — I'd fix this before merge; it's the one thing here that's a safety bug rather than a behavior nit.

Everything else in the two commits looks right.

vendor/isocline/src/tty.c
baijum: Good catch — fixed in 9be6bfe8. Two changes in tty.c:

1. `tty_read_dsr_response` caps the digit buffer at `TTY_PUSH_MAX - 2` (29 digits) so a restore is at most 1 + 29 + 2 = 32 bytes, and skips the restore entirely when `cpush_count + n + 3` would exceed `TTY_PUSH_MAX` (defensive against leftover bytes). A hostile 30+ digit CSI just fails the read; nothing is corrupted or lost beyond the cap.
2. `tty_cpush`'s guard now checks `cpush_count`, the counter the writes actually use — noted as a KAAPPI PATCH 5 change since the mismatch is pre-existing upstream and this reader is the first caller that can exercise it.

The pty test has a fifth run: a 40-byte CSI of `0;` with no `R` terminator delivered during the DSR window, asserting the REPL survives and a later form evaluates. I verified it crashes the child (OOB write corrupting the tty struct, SIGTRAP) without the fix and passes with it.

One mea culpa: I caught mid-verification that zig's build cache was serving me a stale object in an earlier round of local testing (my first "green" runs of the type-ahead test were against a binary from before the push-back fix — the regression test caught it when the garble case crashed). Everything above was re-verified against a clean full rebuild (`rm -rf .zig-cache && zig build`), and the full suite (unit + 2095 Scheme tests) is green.

baijum:
LGTM ✅

Re-reviewed `9be6bfe` and `48c414b`. The round-2 memory-safety finding is fully resolved:

- `buf` is capped at `TTY_PUSH_MAX - 2`, so `n ≤ 29` and a full restore is at most `1 (c) + 29 + 2 ('[' + ESC) = 32` bytes — exactly `cpushbuf`'s capacity, no overflow.
- The restore is additionally gated on `cpush_count + n + 3 <= TTY_PUSH_MAX`, so it stays safe even with leftover bytes in the buffer.
- `tty_cpush`'s guard now checks `cpush_count` (the counter the writes use), so the `assert` is a genuine backstop instead of dead code.
- The new `garble` pty run (40-byte digit/`;` CSI with no `R` during the pending DSR query, asserting no crash + a later form still evaluates) is exactly the right regression, and it's confirmed to have crashed pre-fix.

I traced the push-back bounds on every exit path and they hold. All three review findings across the two rounds — type-ahead loss, press/release double-move, and the `cpushbuf` overflow — are now addressed with matching regressions. Nice work.

(One negligible, pre-existing nit, explicitly **non-blocking**: on the loop's timeout path `c` retains the last-read digit, so a partial CSI that stalls mid-parse re-pushes a duplicated terminator byte. It only ever perturbs an already-garbled terminal report — never user input or memory — so not worth a change.)

PR #2267, opened 2026-08-09
Splice definition-context begins in body scans (Fixes #2075)

### LINKED ISSUE #2075: [R7RS 4.2.3] A begin-wrapped internal define in a let body escapes to the global environment unless an enclosing procedure exists

baijum:
Fixes #2075.

## What

A `begin`-wrapped internal definition in a `let`/`let*`/`letrec` body escaped to the global environment unless an enclosing procedure existed. Per R7RS 4.2.3 a definition-context `begin` behaves exactly as if the wrapper were absent, so

```scheme
(let ((g 'outer)) (begin (define g 'inner)) g)
```

must answer `inner` at top level — it answered `outer`, and the definition escaped the `let` and silently overwrote an unrelated global.

## Root cause (two halves)

1. **`scanBodyDefs` never descended into `begin`** (`src/compiler_lambda.zig`) — a begin-wrapped define was neither pre-declared into the body's letrec* region nor compiled as a body definition. The scan now unwraps spliceable begins (recursively, mirroring the IR lowerer's shadow test for a `begin` head) before its three passes run, so definitions inside them join the letrec* region like unwrapped ones — mutual recursion, `define-record-type`, `define-values` and `define-syntax` included.

2. **`compileDefine`'s `in_body_scope` gate was only set by the procedure-body paths** — the let-family bodies (`compileBodyForms`) were missing it, so a definition reached at compile time (a macro expansion producing one, or a begin the scan couldn't splice) became a global at top level while the identical text inside a lambda bound a local. `compileBodyForms` now sets `in_body_scope` like `compileBody`/`compileSyntaxBody` do.

## Also in this PR

- **SRFI 188 `.sld` header** — its flagship claim ("a hand-written `(let ((x 'outer)) (begin (define x 'inner) #f) x)` evaluates to `outer` at top level") was the doc-truth half of this bug; rewritten. The library's documented non-splicing delegation behavior is unchanged (pinned by tests).
- **`tests/scheme/srfi/srfi188.scm`** — the 4 assertions pinning the wrong answers are flipped to the R7RS-required answers; the 4 disabled assertions are enabled.
- **`tests/scheme/compliance/begin-spliced-define-2075.scm`** — new regression suite (21 assertions): the issue's repro across let/let*/letrec/letrec*, nested and doubly-nested begins, fresh names, mutual recursion through a begin, all definition forms inside a begin, macro-produced defines in let bodies, and the shadowed-`begin` control. Evaluated at true top level per the SRFI-64 lambda-wrapping caveat from the issue.
- **`tests/scheme/compile/native-let-internal-define-root-1854.sh`** — three new cases pin that the native tier (which routes begin-spliced defines through the interpreter fallback) agrees with the interpreter's now-correct answers.
- **`src/tests_core_eval.zig`** — Zig unit tests for the shadowing, letrec*-region, and macro-produced halves.

## Verification

- `zig build test` — pass (also under `-Dgc-stress=true`)
- `bash tests/scheme/run-all.sh` — 2095 pass, 0 fail (incl. 1395/1395 R7RS)
- New compliance test: 13 failures pre-fix, 21/21 post-fix

### DIFF
--- a/src/compiler_lambda.zig
+++ b/src/compiler_lambda.zig
@@ -173,6 +173,154 @@ pub const BodyScan = struct {
     }
 };
 
+/// Is `form` a definition-context `begin` — a literal `(begin ...)` whose
+/// contents splice into the surrounding body per R7RS 4.2.3, "evaluated
+/// exactly as if the enclosing begin construct were not present"?
+///
+/// The shadow test mirrors the IR lowerer (ir.zig, lowerFormWithMacros): a
+/// lexical binding or macro shadowing `begin` makes the form an ordinary
+/// procedure call, not a splice, while a hygienically renamed `begin` from a
+/// macro template keeps its special-form meaning and is never shadowed.
+fn isSpliceableBegin(compiler: *const Compiler, form: Value) bool {
+    if (!types.isPair(form)) return false;
+    const head = types.car(form);
+    if (!types.isSymbol(head)) return false;
+    const name = types.symbolName(head);
+    const effective = types.stripHygienicPrefix(name);
+    if (!std.mem.eql(u8, effective, "begin")) return false;
+    if (std.mem.eql(u8, effective, name)) {
+        if (compiler.isLexicallyBound(name)) return false;
+        if (compiler.lookupMacro(name) != null) return false;
+    }
+    return true;
+}
+
+/// Recursively append the effective body elements of `form` to `elems`: a
+/// spliceable `begin` contributes its (recursively spliced) children, any
+/// other form contributes itself. Called under no_collect (see
+/// spliceLeadingBegins).
+fn appendSplicedBodyElement(compiler: *Compiler, elems: *std.ArrayList(Value), form: Value) CompileError!void {
+    if (isSpliceableBegin(compiler, form)) {
+        var child = types.cdr(form);
+        while (types.isPair(child)) {
+            try appendSplicedBodyElement(compiler, elems, types.car(child));
+            child = types.cdr(child);
+        }
+        // An improper begin `(begin e1 . tail)` is invalid syntax — reject
+        // it rather than silently dropping the tail from the spliced body.
+        if (child != types.NIL) return CompileError.InvalidSyntax;
+    } else {
+        elems.append(compiler.gc.allocator, form) catch return CompileError.OutOfMemory;
+    }
+}
+
+/// Does the body's leading definition region bind the name `begin` with a
+/// `define-syntax`? The scanner installs body-local macro bindings only in
+/// its second pass — after this splice would have run — and the IR lowerer
+/// gives a macro shadowing `begin` precedence over the special form, so a
+/// body that binds its own `begin` must not have its `(begin ...)` forms
+/// spliced as if the wrapper were absent: they are macro uses. Walks the
+/// already-flattened leading forms (the scan's own recognition rules: a
+/// literal define-family head continues the region, anything else ends it);
+/// the list is already spliced, so no descent is needed. The keyword is
+/// compared after stripping a hygiene prefix, so a template-introduced
+/// `define-syntax begin` shadows the (identically renamed) uses too.
+fn bodyBindsBeginMacro(flattened: Value) bool {
+    var cur = flattened;
+    while (types.isPair(cur)) {
+        const form = types.car(cur);
+        if (!types.isPair(form)) break;
+        const head = types.car(form);
+        if (!types.isSymbol(head)) break;
+        const hn = types.symbolName(head);
+        if (std.mem.eql(u8, hn, "define-syntax")) {
+            const rest = types.cdr(form);
+            if (types.isPair(rest) and types.isSymbol(types.car(rest)) and
+                std.mem.eql(u8, types.stripHygienicPrefix(types.symbolName(types.car(rest))), "begin"))
+            {
+                return true;
+            }
+            cur = types.cdr(cur);
+            continue;
+        }
+        if (std.mem.eql(u8, hn, "define") or std.mem.eql(u8, hn, "define-record-type") or
+            std.mem.eql(u8, hn, "define-values"))
+        {
+            cur = types.cdr(cur);
+            continue;
+        }
+        break;
+    }
+    return false;
+}
+
+/// R7RS 4.2.3 definition-context `begin` splicing (#2075): a literal
+/// `(begin ...)` among a body's leading forms — or directly nested in
+/// another such begin — behaves exactly as if the wrapper were absent, so a
+/// definition inside it joins the body's letrec* region: it must shadow an
+/// enclosing binding and must not escape into the global environment. The
+/// body scanners only recognize literal `define`-family heads as body
+/// elements, so this helper unwraps every spliceable `begin` anywhere in
+/// `body` before scanning.
+///
+/// Returns a freshly allocated list equivalent to `body` with every
+/// spliceable begin removed (the caller must root it), or null when there
+/// is nothing to splice — the common case allocates nothing. The result
+/// shares all element and tail structure with `body`; only the cons cells
+/// connecting spliced elements are new. Built under no_collect so the
+/// fresh cells cannot be swept before the caller roots the result; the
+/// recursion mirrors the IR's own unbounded lowering of nested begins.
+fn spliceLeadingBegins(compiler: *Compiler, body: Value) CompileError!?Value {
+    const gc = compiler.gc;
+
+    // Fast path: a body with no spliceable begin element allocates nothing.
+    var any_begin = false;
+    {
+        var cur = body;
+        while (types.isPair(cur)) {
+            if (isSpliceableBegin(compiler, types.car(cur))) {
+                any_begin = true;
+                break;
+            }
+            cur = types.cdr(cur);
+        }
+    }
+    if (!any_begin) return null;
+
+    gc.no_collect += 1;
+    errdefer gc.no_collect -= 1;
+
+    var elems: std.ArrayList(Value) = .empty;
+    defer elems.deinit(gc.allocator);
+    var cur = body;
+    while (types.isPair(cur)) {
+        try appendSplicedBodyElement(compiler, &elems, types.car(cur));
+        cur = types.cdr(cur);
+    }
+    // An improper body tail must not be silently dropped either.
+    if (cur != types.NIL) return CompileError.InvalidSyntax;
+
+    var result = types.NIL;
+    var i = elems.items.len;
+    while (i > 0) {
+        i -= 1;
+        result = gc.allocPair(elems.items[i], result) catch return CompileError.OutOfMemory;
+    }
+
+    // A body-local `(define-syntax begin ...)` shadows the special form for
+    // the whole body (IR precedence: macro over special form), and the scan
+    // installs that binding only after this splice would have run — so
+    // decline to splice and let the scan + IR handle every `(begin ...)` as
+    // a macro use, exactly as they would in non-leading position.
+    if (bodyBindsBeginMacro(result)) {
+        gc.no_collect -= 1;
+        return null;
+    }
+
+    gc.no_collect -= 1;
+    return result;
+}
+
 pub fn scanBodyDefs(compiler: *Compiler, body: Value, handle_define_syntax: bool) CompileError!BodyScan {
     var scan_result = BodyScan{};
     scan_result.compiler = compiler;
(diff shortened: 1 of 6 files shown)


src/compiler_lambda.zig
```diff
@@ -173,6 +173,98 @@ pub const BodyScan = struct {
...
+fn isSpliceableBegin(compiler: *const Compiler, form: Value) bool {
+    if (!types.isPair(form)) return false;
+    const head = types.car(form);
+    if (!types.isSymbol(head)) return false;
+    const name = types.symbolName(head);
+    const effective = types.stripHygienicPrefix(name);
+    if (!std.mem.eql(u8, effective, "begin")) return false;
+    if (std.mem.eql(u8, effective, name)) {
+        if (compiler.isLexicallyBound(name)) return false;
+        if (compiler.lookupMacro(name) != null) return false;
```
baijum, line 193: **Body-local `(define-syntax begin ...)` is spliced as the special form → silently wrong.**

`isSpliceableBegin` consults `compiler.lookupMacro("begin")`, but this splice runs at the very start of `scanBodyDefs`, *before* the body's own `define-syntax` bindings are installed (that happens in the later pass). The IR lowerer checks macros before special forms ("macro wins"), so:

```scheme
(let ()
  (define-syntax begin (syntax-rules () ((_ . r) (list 'm . r))))
  (begin 1 2))
```

should evaluate to `(m 1 2)`, but the splice doesn't see the not-yet-installed local macro, treats `(begin 1 2)` as the special form, and flattens it to `1 2` → the body returns `2`. Silently wrong for a legal program.

This appears already handled in the latest local edits (a `bodyBindsBeginMacro` guard on the flattened region) — flagging so it's captured against the pushed revision.

baijum (commented):
## Review — Fixes #2075

The diagnosis is correct and the two-halves framing is exactly right: `scanBodyDefs` never descended into `begin`, and `compileBodyForms` never set `in_body_scope`. The fix addresses both, and the test coverage is genuinely thorough (the SRFI-64 lambda-wrapping caveat is handled correctly by evaluating the shadowing probes at true top level into a `define`).

**Things I verified and am happy with:**

- **GC rooting is correct.** `spliceLeadingBegins` builds the fresh spine under `no_collect`, and the caller roots the head into `extra_roots` before the next heap allocation. `scanBodyDefs.roots_base` is captured *before* the splice append, so `BodyScan.deinit`'s `shrinkRetainingCapacity(roots_base)` truncates the splice root only after the caller has compiled `scan.remaining` (which aliases the fresh spine) — the #1010 pattern the comment cites. The `errdefer gc.no_collect -= 1` vs. the manual decrements are balanced on every path (no double-decrement).
- **`in_body_scope` change is safe.** The only two callers of `compileBodyForms` are `compileBody` (already set it) and `compileLetBody` (the one that was missing it); no top-level-program path routes through it, so top-level `define`s stay global.
- **`isSpliceableBegin` mirrors the IR `is_shadowed` test** (`stripHygienicPrefix`, only shadow-checking a non-hygienic head).

**Two correctness bugs in this pushed revision** (both narrow, both silent — and I can see they're already addressed in the latest local worktree edits, so this is confirmation the fixes are the right call rather than new work):

1. A body-local `(define-syntax begin ...)` is spliced as the special form → silently wrong answer (inline comment).
2. An improper `begin` tail is silently dropped instead of raising `InvalidSyntax` (inline comment).

**Minor / non-blocking:**

- The helper is named `spliceLeadingBegins` and its doc says "among a body's leading forms", but the loop actually splices a spliceable `begin` at *any* position in the body. That's harmless (and arguably more correct), but the name/comment overstate the restriction — worth aligning the wording.
- Extreme corner, likely pre-existing and not a regression: a body-local *variable* `(define begin ...)` has the same install-ordering blind spot as bug #1 (the local isn't in `locals` when `isSpliceableBegin` runs, so `isLexicallyBound` misses it). Not worth fixing in this PR unless it's trivial alongside the macro case; just noting it.

Overall this is a solid, well-tested fix. Once the two silent-behavior cases above are covered (which the local edits appear to do), I'd be happy to see it land.

src/compiler_lambda.zig
```diff
@@ -173,6 +173,98 @@ pub const BodyScan = struct {
...
+}
+
+/// Recursively append the effective body elements of `form` to `elems`: a
+/// spliceable `begin` contributes its (recursively spliced) children, any
+/// other form contributes itself. Called under no_collect (see
+/// spliceLeadingBegins).
+fn appendSplicedBodyElement(compiler: *Compiler, elems: *std.ArrayList(Value), form: Value) CompileError!void {
+    if (isSpliceableBegin(compiler, form)) {
+        var child = types.cdr(form);
+        while (types.isPair(child)) {
```
baijum, line 205: **Improper `begin` tail is silently dropped.** `while (types.isPair(child))` stops at a non-nil tail without checking it, so a malformed `(begin (define a 1) . 2)` compiles (dropping the `. 2`) instead of raising `InvalidSyntax`. The same applies to the outer body loop in `spliceLeadingBegins`. Also appears fixed in the latest local edits (`if (child != NIL) return CompileError.InvalidSyntax`) — noting it against the pushed code.

baijum:
All three CodeRabbit findings addressed in the latest push (`278709f3`):

1. **Improper list tails** — `spliceLeadingBegins` now returns `InvalidSyntax` when a spliceable `begin`'s children or the body itself ends in a non-NIL (improper) tail, instead of silently dropping it. Verified: `(let () (begin (define x 1) . bad) x)` now reports `compile error[KP2001]: invalid syntax`.

2. **Body-local macro named `begin`** — a leading `(define-syntax begin ...)` now suppresses splicing for the whole body (`bodyBindsBeginMacro`), so `(begin ...)` forms compile as macro uses exactly as the IR would treat them in non-leading position. The scanner installs body-local macro bindings only after the splice would have run, hence the up-front guard. Regression test added: `(let () (define-syntax begin (syntax-rules () ((_ . rest) (list 'macro . rest)))) (begin 1 2))` → `(macro 1 2)`.

3. **Golden output assertions** — `check()` in `native-let-internal-define-root-1854.sh` gained an optional 4th argument pinning the exact expected output, and all three new #2075 cases assert it, so interpreter and native can no longer pass by agreeing on a wrong answer.

baijum:
LGTM on `278709f3`. ✅

Both correctness bugs I flagged in the previous revision are fixed and re-verified:

1. **Body-local `(define-syntax begin ...)`** — `bodyBindsBeginMacro` now declines splicing when the body binds its own `begin` macro, so `(begin 1 2)` compiles as a macro use (`(macro 1 2)`) rather than being flattened as the special form. The `no_collect` accounting on the new early-`return null` path is balanced (manual decrement; the `errdefer` correctly does not double-fire on a non-error return).
2. **Improper `begin` tail** — now returns `InvalidSyntax` in both `appendSplicedBodyElement` (inner `begin` children) and `spliceLeadingBegins` (the body tail) instead of silently dropping the tail. Confirmed `(let () (begin (define x 1) . bad) x)` → `KP2001: invalid syntax`.
3. The golden-output assertions in `native-let-internal-define-root-1854.sh` close the interpreter/native "wrong in the same way" gap.

What I re-checked and am happy with:

- **GC rooting** — the fresh spliced spine is built under `no_collect`, rooted into `extra_roots` before the next heap allocation, and `roots_base` (captured before the splice append) means `BodyScan.deinit`'s `shrinkRetainingCapacity` releases it only after the caller has compiled `scan.remaining`. Balanced on every path, including the two new error returns.
- **`in_body_scope`** — only `compileBody` and `compileLetBody` reach `compileBodyForms`; no top-level-program path does, so top-level `define`s stay global.
- **`isSpliceableBegin`** faithfully mirrors the IR lowerer's `is_shadowed` / hygiene test.
- **Tests** — the compliance suite exercises let/let\*/letrec, nested and doubly-nested begins, fresh names, mutual recursion, all definition forms inside a begin, macro-produced defines, and the shadowed-`begin` control, all at true top level per the SRFI-64 lambda-wrapping caveat. Benchmark comment shows no meaningful perf change (all within noise).

Two non-blocking nits remain, neither worth holding the PR for:
- `spliceLeadingBegins`'s summary line / name still says "leading", though the loop splices a spliceable `begin` at any body position (the detailed doc paragraph does say "anywhere" — only the one-liner is loose). Cosmetic.
- CodeRabbit's still-open suggestion to add `(test-error "a fresh name does not escape the let" fresh2075)` is a fair coverage add if you want the fresh-name probe to also assert the name is globally unbound, not just locally `100`.

Nice, well-documented fix.

Issue #2167, opened 2026-08-02
[R7RS 6.1] eqv?, equal?, memv, assv, case, and eqv-keyed hash tables equate an exact and an inexact complex — exactness flags are never compared


baijum:
**Found by:** external testing (paal's reader test suite against kaappi v0.22.0) · **Verified:** main @ 98195459 and v0.22.1

R7RS §6.1 requires `(eqv? a b)` to be `#f` whenever one number is exact and the other inexact. For complex numbers, every eqv?-semantics comparator in the codebase compares the two f64 components bitwise and never consults the `exact_real`/`exact_imag` flags:

```scheme
(eqv? (make-rectangular -3/2 -1) -1.5-1.0i)     ; => #t   must be #f
(equal? (make-rectangular -3/2 -1) -1.5-1.0i)   ; => #t   must be #f (equal? on numbers = eqv?)
(memv (make-rectangular -3/2 -1) (list -1.5-1.0i))            ; => (-1.5-1.0i)   must be #f
(assv (make-rectangular -3/2 -1) (list (cons -1.5-1.0i 'x)))  ; => (-1.5-1.0i . x)   must be #f
(case (make-rectangular -3/2 -1) ((-1.5-1.0i) 'inexact) (else 'else))  ; => inexact   must be else
(let ((t (make-hash-table eqv?)))                ; SRFI 69
  (hash-table-set! t -1.5-1.0i 'inexact-key)
  (hash-table-ref/default t (make-rectangular -3/2 -1) 'miss))  ; => inexact-key   must be miss

(= (make-rectangular -3/2 -1) -1.5-1.0i)        ; => #t — correct; = ignores exactness
(eqv? 1 1.0)                                    ; => #f — correct; only the complex arm is affected
```

## Cause

The comparison is duplicated four times, and every copy has the same gap — components bitwise, flags ignored:

| site | serves |
|---|---|
| `src/primitives.zig:835` (`eqvP`) | `eqv?`, and `case` (compiled `case` emits an `eqv?` call) |
| `src/primitives.zig:879` (`deepEqualWithVisited`) | `equal?`, `member`/`assoc`, equal-keyed tables |
| `src/primitives_list.zig:209` (`isEqv`) | `memv`, `assv` |
| `src/primitives_hashtable.zig:140` (`eqvEqual`) | SRFI 69 eqv-keyed hash tables |

This kind of drift between copies is how the gap survived — the same shape as #1911's two `applyExactness` implementations.

[...]

PR #2268, opened 2026-08-09
Store complex components as Values: exact complex arithmetic, make-rectangular, write/real-part, reader

### LINKED ISSUE #2166: [R7RS 6.2] Exact complex components are stored as f64 + flag: arithmetic drops exactness, make-rectangular 2^53+1 is silently off by one, write and real-part disagree about the same object

baijum:
## Summary

Fixes #2166: exact complex components are stored as `Value`s (fixnum/bignum/rational/flonum) instead of f64 + exactness flags. Every instance in the issue is fixed:

```scheme
(define z (make-rectangular 3/2 1))
(+ z z)     ; => 3+2i        (was inexact 3.0+2.0i)
(* z z)     ; => 5/4+3i      (was inexact 1.25+3.0i)
(- z z)     ; => 0           (was 0.0)
(/ z)       ; => 6/13-4/13i  (was inexact 0.4615…-0.3076…i)
(make-rectangular 9007199254740993 1)   ; => 9007199254740993+1i  (was off-by-one)
(make-rectangular (expt 10 400) 1)      ; => exact, not +inf.0
(real-part (make-rectangular 1/3 1))    ; => 1/3  (was 0.333…; write said 1/3+1i)
```

## What changed

- **Representation** — `Complex.real`/`Complex.imag` are `Value`s; the exactness flags are gone. Complex is now a Value-bearing heap type: mark/sweep/deep-copy arms added in `gc_collect.zig`/`gc_sweep.zig`/`gc_deep_copy.zig`, the tracing field pin re-pinned.
- **Arithmetic** — `+ - * /` and `expt` (integer exponent) run componentwise over the exact tower and are exact-closed; the interim unary-negation special case dissolves. Division by an *exact* zero complex now raises (`(/ 1+2i 0+0i)`), matching `(/ 1+2i 0)`.
- **make-rectangular** — never touches an f64; `real-part`/`imag-part` return the stored component.
- **Printer** — components print through the normal numeric printer; the f64 un-rounding path (`writeComplexPart`/`floatToRational`) is deleted. `write` and `real-part` agree.
- **Reader / string->number** — components are built digit-exactly at any size; the #2182/#2243 f64 round-trip gates dissolve (`9007199254740993+1i`, `#x20000000000001+2i`, `10000000000000000000000000/3+1i` all read exactly). The scanner roots scanned components until the datum constructor converts them.
- **eqv? family** — `types.complexEqv` compares components with numeric eqv?; eqv-keyed hash tables hash components by value.
- **Semantics** — per R7RS 6.2.2 a stored complex is never mixed-exactness; a zero imaginary part demotes to the real, except that a literal's *inexact* zero imag stays complex (`(real? -2.5+0.0i)` => `#f`) while an exact one demotes (`(integer? 3+0i)` => `#t`).
- **.sbc** — `TAG_COMPLEX` writes the two component constants; the golden byte test is updated.

## Testing

- Full scheme suite: **2097 pass, 0 fail** (incl. 1395/1395 R7RS) plus the compile/differential/probe suites.
- `zig build test` and `zig build test -Dgc-stress=true` green.
- Compliance suite expanded to pin the full #2166 behavior (arithmetic exactness, make-rectangular precision, write/real-part agreement, reader/string->number digit-exactness, expt, division by zero, hash tables); the interim-slice and gate assertions updated.

This also resolves the `eqv?`-semantics gap from #2167 in the new representation and supersedes the #2182/#2243 round-trip gates.


baijum (commented):
Reviewed the complex-as-Values rework end to end (representation, GC, arithmetic, printer, reader, `string->number`, `.sbc`, tests). The core design is sound and the test coverage is genuinely strong — the reader/`string->number` digit-exactness pins and the eqv?/hash-table tests are exactly right. One finding is a real memory-safety bug that I'd fix before merge; the rest are correctness/robustness items and nits.

**High — fix before merge**
- **GC use-after-free in the radix-prefixed complex reader.** `rootComplexImag` writes `complex_root[1]` but never registers the root slots — only `rootComplexReal` does the lazy `pushRoot`. In `readIntegerWithRadix`, `tryComplexTail` stores a *heap* imaginary via `rootComplexImag` first, and the caller only calls `rootComplexReal` **after** allocating the real part, so the imaginary is unrooted across that allocation. `(read (open-input-string "#x1/2+3/4i"))` (or the same literal in source) frees the `3/4` imaginary while `1/2` is being built. Deterministic panic under `-Dgc-stress=true`; latent otherwise. Details inline.

**Medium**
- Printer emits the *exact* `+i`/`-i` spelling for an **inexact** unit imaginary, so `write` and `imag-part` disagree and a write→read loses inexactness — the exact defect class #2166 set out to close.
- `complexExptExact` runs an unbounded O(n) loop; a large fixnum exponent hangs even when the result is trivial (`(expt +i 1000000000)` ⇒ `1`).
- `make-rectangular` and the reader disagree on an inexact zero imaginary: `(make-rectangular 1.5 0.0)` ⇒ real, but the literal `1.5+0.0i` ⇒ complex. Worth confirming intent — it's internally inconsistent and breaks make-rectangular/real-part/imag-part round-tripping.

**Low / nits**
- `.sbc` `TAG_COMPLEX` wire format changed without bumping `bytecode_file.VERSION` (still 11); only `compilerHash` saves it.
- The `.sbc` round-trip unit test exercises only flonum components — the new exact (fixnum/bignum/rational) component paths get no coverage.
- `toComplexParts` (f64) is now dead code.

Nice work overall — the digit-exact reader and the componentwise exact tower are clean.

src/reader_tokens.zig
```diff
@@ -85,32 +86,102 @@ fn bigRationalToken(self: *Reader, num_str: []const u8, den_str: []const u8, rad
...
+fn rootComplexReal(self: *Reader, real: Value) void {
+    self.complex_root[0] = real;
+    if (!self.complex_roots_pushed) {
+        self.complex_roots_pushed = true;
+        self.gc.pushRoot(&self.complex_root[0]);
+        self.gc.pushRoot(&self.complex_root[1]);
+    }
+}
+
+fn rootComplexImag(self: *Reader, imag: Value) void {
```
baijum, line 107: **High — GC use-after-free.** `rootComplexImag` only writes `self.complex_root[1]`; the two root slots are registered with the GC **exclusively** inside `rootComplexReal` (the lazy `pushRoot` gated on `complex_roots_pushed`). So a component stored here is GC-visible only if `rootComplexReal` has already run on this Reader.

That ordering is violated in `readIntegerWithRadix`: `tryComplexTail` builds and stores the *imaginary* first (`rootComplexImag`), and the caller calls `rootComplexReal` **after** allocating the real part — e.g. line 1240 (`tryRationalComponent` → `makeRationalReduced`), line 1239 (`parseBignumString`), line 1302 (`exactIntComponent` → `allocBignumFromI64`). While `complex_roots_pushed` is still false, the heap imaginary sits unrooted in `complex_root[1]` across that allocation.

Reproducer (both components heap-allocated): `(read (open-input-string "#x1/2+3/4i"))` — building `1/2` collects the already-parsed `3/4`, leaving `tail.imag` dangling in the returned token. Also `#x99999999999999999999+FFFFFFFFFFFFFFFFFFi` (bignum real + bignum imag). Deterministic “marking freed object” panic under `-Dgc-stress=true`; latent (GC-threshold-dependent) in normal builds. Reachable from an ordinary `read`.

Fix: have `rootComplexImag` push the slots lazily the same way `rootComplexReal` does (both slots default to `NIL`, so pushing before slot[0] is set is safe). A regression test should read one of these radix-prefixed heap-component literals under `-Dgc-stress=true` — the existing Scheme-level `#x1/2+3/4i` test at `tests/scheme/compliance/reader-delimiter-gaps.scm` doesn't catch it because that suite isn't run under gc-stress and the new unit tests use only fixnum imaginaries.

src/printer.zig
```diff
@@ -583,21 +583,32 @@ fn printValueOnce(
...
-                    } else if (std.math.isInf(im)) {
-                        try writer.writeAll(if (im > 0) "+inf.0i" else "-inf.0i");
+                    } else if (isInfComponent(imag)) {
+                        try writer.writeAll(if (componentNegative(imag)) "-inf.0i" else "+inf.0i");
                     } else {
-                        try writer.writeByte(if (im < 0 or std.math.signbit(im)) '-' else '+');
-                        const mag = @abs(im);
-                        if (mag != 1.0 or has_real) try writeComplexPart(writer, mag, c.exact_imag);
+                        try writer.writeByte(if (componentNegative(imag)) '-' else '+');
+                        if (!componentUnit(imag) or has_real) try writeComponentMagnitude(allocator, writer, imag);
```
baijum, line 611: This gates the unit spelling on `componentUnit(imag)` (|v| == 1 for a flonum too), so an **inexact** unit imaginary with no real part prints as `+i`/`-i`. But kaappi's reader reads `+i` as **exact**, so for `(make-rectangular 0.0 1.0)`: `write` ⇒ `+i` (reads back exact) while `(imag-part z)` ⇒ inexact `1.0`, and a write→read round-trip silently flips exactness. That's exactly the `write`/`real-part` disagreement #2166 set out to eliminate (it only bites when there's no real part; with a real part `has_real` forces the magnitude and `3.0+1.0i` prints correctly). Suggest gating the unit spelling on the component being *exact* ±1 (`isFixnum` ±1), not flonum ±1.0.

src/primitives_numeric.zig
```diff
@@ -453,7 +490,27 @@ fn complexPowGeneral(gc: *memory.GC, zr: f64, zi: f64, wr: f64, wi: f64) Primiti
...
+/// Exact complex exponentiation: (a+bi)^n for an exact complex base and an
+/// exact integer exponent, by repeated multiplication over the exact tower
+/// (which is exact-closed, R7RS 6.2.2 / kaappi#2166). A negative exponent
+/// reciprocates the base-first power; a zero exponent yields exact 1.
+fn complexExptExact(gc: *memory.GC, base: Value, n: i64) PrimitiveError!Value {
+    var count = if (n < 0) -n else n;
+    var result: Value = types.makeFixnum(1);
+    var slot = gc.rootedSlot(result) catch return PrimitiveError.OutOfMemory;
+    defer slot.release();
+    while (count > 0) : (count -= 1) {
```
baijum: This loop has no bound on `count` (a fixnum exponent up to ~1.4e14), unlike the f64 path which caps repeated multiplication at `@abs(wr) < 100` and otherwise uses the closed-form. `(expt +i 1000000000)` loops a billion times even though the true result is `1`, and `(expt 2+1i 100000)` grows a bignum quadratically. The real integer `expt` uses `bignum_mod.expt` (square-and-multiply); this should do the same (fast exponentiation over the exact complex tower), so cost is O(log n) multiplications rather than O(n).

src/primitives_numeric.zig
```diff
@@ -949,19 +1009,65 @@ pub fn toComplexParts(v: Value) PrimitiveError!struct { real: f64, imag: f64 } {
...
+/// Reader/string->number complex construction: like makeComplexOrRealV, but
+/// only an EXACT zero imaginary part demotes to the real component. The R7RS
+/// conformance suite pins both sides: (real? -2.5+0.0i) => #f (an inexact
+/// zero imag written in a literal stays a complex) and (integer? 3+0i) => #t
+/// (an exact zero imag demotes). Inexactness still propagates whole-number
+/// (R7RS 6.2.2), so the two parsers agree (R7RS 6.2.7, kaappi#2166).
+pub fn makeComplexOrRealLiteral(gc: *memory.GC, real_in: Value, imag_in: Value) PrimitiveError!Value {
+    var real = real_in;
+    var imag = imag_in;
+    const imag_exact_zero = types.isExactNumber(imag) and isZeroValue(imag);
```
baijum, line 1057: `makeComplexOrRealLiteral` keeps an *inexact* zero imaginary complex (`1.5+0.0i` ⇒ `(real? …)` #f), but `makeComplexOrRealV` (used by `make-rectangular`) demotes it: `(make-rectangular 1.5 0.0)` ⇒ real `1.5`. Both are pinned in the compliance suite, so this is deliberate — but it's internally inconsistent (Racket keeps both complex; Chez demotes both) and it breaks round-tripping: `(make-rectangular (real-part 1.5+0.0i) (imag-part 1.5+0.0i))` loses complexness. Worth a second look at whether `make-rectangular` should share the literal path's inexact-zero rule.

src/bytecode_file_write.zig
```diff
@@ -314,11 +314,13 @@ fn writeConstant(w: *Writer, allocator: std.mem.Allocator, val: Value, all_funcs
...
+                // Components are written as nested constants (fixnum /
+                // bignum / rational / flonum), digit-exact like every other
+                // value — the old f64+flags encoding could not carry an
+                // exact component (kaappi#2166).
                 try w.writeU8(allocator, bf.TAG_COMPLEX);
-                try w.writeF64(allocator, cx.real);
-                try w.writeF64(allocator, cx.imag);
-                try w.writeU8(allocator, if (cx.exact_real) @as(u8, 1) else @as(u8, 0));
-                try w.writeU8(allocator, if (cx.exact_imag) @as(u8, 1) else @as(u8, 0));
+                try writeConstant(w, allocator, cx.real, all_funcs, seen, depth + 1);
```
baijum, line 325: The on-disk `TAG_COMPLEX` layout changed (f64+f64+u8+u8 → two nested constants) but `bytecode_file.VERSION` stayed at 11 — and the immediately prior format change (#2188) *did* bump to 11. In practice `compilerHash()` (which folds in `git_build_id`) invalidates a stale cache from any other commit, so no real misread occurs; but the explicit version signal is the defense-in-depth if two builds ever share a hash (e.g. dirty-vs-dirty at the same tree). Consider bumping to 12 for consistency.

src/bytecode_file.zig
```diff
@@ -989,7 +989,7 @@ test "bytecode round-trip: vector pair bignum rational complex constants" {
     const rat = try gc.allocRational(rat_num, rat_den);
     func.constants.append(allocator, rat) catch unreachable;
 
-    const cx = try gc.allocComplexEx(3.0, 4.0, false, false);
+    const cx = try gc.allocComplex(types.makeFlonum(3.0), types.makeFlonum(4.0));
```
baijum, line 995: The round-trip test only constructs flonum components, so the new recursive serializer's exact paths (fixnum / bignum / rational components) — the whole point of this PR — get no `.sbc` regression coverage. Suggest adding a complex with a bignum and a rational component (e.g. `100000000000000000000000+1/2i`) to this test.

baijum:
Thanks for the review — all actionable items are addressed in 8b7c034f.

**High (fixed)**
- GC use-after-free in the radix-prefixed complex reader: `rootComplexImag` now performs the lazy root registration too (only `rootComplexReal` did before, but `readIntegerWithRadix` parses the imaginary part first via `tryComplexTail`). Reproduced as `GC: marking freed object` under `-Dgc-stress=true` — and it was the `gc-stress-scheme` CI failure — then verified fixed with regression tests for `#x1/2+3/4i`, `#x800000000000+99999999999999999999i`, and `#x20000000000001+3/4i` in the compliance suite.

**Medium (fixed)**
- Printer unit-imag spelling: `+i`/`-i` is now only emitted for an *exact* ±1 fixnum; an inexact ±1.0 prints its magnitude, so `(write (make-rectangular 0.0 1.0))` → `+1.0i` and write→read preserves inexactness.
- `complexExptExact` now uses square-and-multiply — `(expt +i 1000000000)` returns instantly instead of looping a billion times.
- `inexact` on an all-inexact complex returns it unchanged, so `-2.5+0.0i` keeps its complexness.
- Unary `(- z)` and the `(/ z)` conjugation use IEEE negation (`0.0` → `-0.0`), so `(- 0.0+1.0i)` ⇒ `-0.0-1.0i` as IEEE requires.

**On the make-rectangular vs reader inexact-zero discrepancy**: this is deliberate and pinned by two suites — the R7RS conformance suite requires `(real? -2.5+0.0i)` ⇒ `#f` (the reader keeps an inexact-zero imag complex), while the srfi160 audit requires `(real? (make-rectangular 1.5 0.0))` ⇒ `#t` (make-rectangular demotes). The two are different construction sites with different rules, documented in the `Complex` doc comment. If you'd prefer them unified, that's a separate semantic decision that would require changing one of the pinned suites.

**Low / nits (fixed)**
- `.sbc` `VERSION` bumped 11 → 12 (the `TAG_COMPLEX` payload is wire-incompatible); the writer now validates components are real before serializing; the checked-in fuzz-seed fixture is regenerated.
- `.sbc` round-trip coverage: the bytecode round-trip unit test and the `sbc-constants-numeric` probe now include bignum/rational component complexes (cold/warm compared).
- `toComplexParts` removed (dead after the representation change); `complexPowGeneral`'s unused gc parameter dropped; `string->number` propagates `OutOfMemory` instead of returning `#f` on the signless pure-imaginary path.
- GC tests: `gc tracing: complex traces real and imaginary components` and the remembered-set twin were added, plus a deep-copy test with a bignum component.

Full suite: 2096 pass / 0 fail (incl. 1395/1395 R7RS), `zig build test` and `-Dgc-stress=true` green, all platforms' CI jobs pass.

baijum:
LGTM ✅

All review findings are addressed in `8b7c034f`, verified against the diff:

- **High** — the radix-complex reader GC use-after-free is fixed (`rootComplexImag` now does the lazy root registration), with gc-stress regression tests for `#x1/2+3/4i`, `#x800000000000+99999999999999999999i`, and `#x20000000000001+3/4i`. This was also the red `gc-stress-scheme` CI job, now green.
- **Medium** — printer only uses the `+i`/`-i` spelling for an exact ±1 (so `write`/`imag-part` agree and write→read preserves exactness); `complexExptExact` uses square-and-multiply; the make-rectangular vs reader inexact-zero split is deliberate and documented (pinned by the R7RS conformance and srfi160 suites).
- **Low** — `.sbc` `VERSION` bumped to 12 with a real-component write guard and extended round-trip/probe coverage; dead `toComplexParts` removed.

The IEEE `negate2` and `makeFixnumChecked` i48 short-circuit are nice touches and check out. Suite green (2096/0, 1395/1395 R7RS) including `-Dgc-stress=true`. Nice work.

src/bytecode_file.zig
```diff
@@ -989,9 +992,15 @@ test "bytecode round-trip: vector pair bignum rational complex constants" {
...
     func.constants.append(allocator, cx) catch unreachable;
 
+    // Exact components (bignum real, rational imag) must survive the .sbc
+    // round trip digit-exactly (kaappi#2166).
+    const cx_bn = try gc.allocBignumFromI64(9007199254740993);
+    const cx_exact = try gc.allocComplex(cx_bn, try gc.allocRational(types.makeFixnum(3), types.makeFixnum(4)));
```
baijum: **gc-stress use-after-free — test bug, not production.** This line is the `gc-stress-scheme`/gc-stress-unit CI abort (`GC: marking freed object (use-after-free)`, trace bottoms out at `allocComplex` → `maybeCollect` here).

**Mechanism, step by step:**
1. `cx_bn` (line 1000) is a freshly-allocated **bignum** held only in a Zig local — **not rooted**, and not yet reachable from anything the GC scans (it hasn't been appended to `func.constants` yet).
2. Zig evaluates call arguments **left-to-right before** the call, so `try gc.allocRational(3, 4)` runs **first**. Under `-Dgc-stress=true` every allocation forces a full collection.
3. That inner `allocRational` collection runs while `cx_bn` is unrooted → the collector **frees** `cx_bn` (and gc-stress poisons/quarantines the slot).
4. `gc.allocComplex(cx_bn, rat)` is then called with `cx_bn` **dangling**. `allocComplex` auto-roots its Value args via `rootArgs2` **before** its own `maybeCollect()`, so it pushes the dangling `cx_bn` onto the root set.
5. `allocComplex`'s own `maybeCollect()` collects again, `markRoots` walks the root set, reaches the freed `cx_bn`, and panics.

**Why the other constants are fine:** `bn`/`rat`/`vec`/`pair` above are each appended to `func.constants` (reachable) *before* the next allocation; `cx_bn` is the only value left in a bare local across an intervening allocation. The flonum-component `cx` on line 995 is immune because flonum components are immediates — no heap object to free. This is exactly the `.claude/rules/gc-safety.md` rule *“Root fresh results between allocations.”*

**Why it passed locally / plain `zig build test` but fails in CI:** only `-Dgc-stress=true` collects on that inner `allocRational`; a normal build almost never collects at that instant, so `cx_bn` survives.

**Fix (test-only)** — root `cx_bn` across the rational allocation:
```zig
const cx_bn = try gc.allocBignumFromI64(9007199254740993);
var cx_bn_root = try gc.rootedSlot(cx_bn);
defer cx_bn_root.release();
const cx_rat = try gc.allocRational(types.makeFixnum(3), types.makeFixnum(4));
const cx_exact = try gc.allocComplex(cx_bn_root.get(), cx_rat);
```
(Allocating the rational into its own local first, then calling `allocComplex`, works too — the point is that no allocation may run while `cx_bn` is only in an unrooted local.) The production `allocComplex` path is correct; it was just handed an already-freed value here.

baijum (commented):
The `gc-stress` CI failure (`bytecode round-trip: vector pair bignum rational complex constants` → `GC: marking freed object (use-after-free)`) is a bug in **this new test**, not in the production complex code. It's an unrooted local across a nested allocation, so it only manifests under `-Dgc-stress=true`. Details inline on the exact line.

baijum:
The make-rectangular vs reader inexact-zero-imag design decision is now tracked in #2269 (opened from this PR's review discussion).

Issue #1061, opened 2026-07-04
Extract the shared read->compile->execute->report driver: the same ~100-line loop is copy-pasted 6 times across main.zig, repl.zig, kaappi_lsp.zig


baijum:
## Problem
Six copies with small variations:
1. `runFile` (`src/main.zig:849-961`, with source snippet + stack trace)
2. `runStdin` (`:972-1066` — already silently diverged: no snippet, no trace)
3. `disassembleFile` (`:1068-1139`)
4. `compileFile` (`:1141-1234`)
5. `evalInputInner` (`src/repl.zig:969-1115`, prettyPrint + `_` binding)
6. `runDiagnostics` (`src/kaappi_lsp.zig:773-802`, read+compile only)

Verbatim duplications inside the family: the stack-trace printer is byte-identical in `main.zig:942-954` and `repl.zig:1068-1080`; the `getErrorDetail()`-or-fallback block appears 9x in main.zig + 2x in repl.zig; the read-error format block 8x + 4x; preamble replay twice in main.zig alone (`:447-478` vs `:802-819`).

## Fix
A `src/toplevel_driver.zig` with a config struct (source name; on-result action: print/prettyPrint/disassemble/collect; error sink; snippet/trace toggles; `script_had_error` out-param) plus free `reportRuntimeError` / `printStackTrace` / `replayPreamble`. Cuts ~400 lines and ends the drift. LSP diagnostics get real error messages for free.

Size: L. Risk: medium — **error message text is load-bearing**; `tests/scheme/errors/*.sh` (error-format, exit-code, reader-token-errors, reader-hasmore-errors, test-source-snippet) pin exact stderr and exit codes. The REPL path has zero unit tests — keep its behavior byte-identical and verify manually (`printf '(car 1)' | kaappi`).

As of c0d4f10. From the 2026-07-04 refactoring analysis. Pairs with the cli.zig split.

PR #2270, opened 2026-08-09
Split repl.zig along its natural seams (#2266)

### LINKED ISSUE #2266: repl.zig is over the 1,500-line guideline; split along natural seams
`src/repl.zig` is at ~1,590 lines, above the repo's 1,500-line file-size
guideline (AGENTS.md). The REPL loop, the isocline callbacks (completeness,
completion, highlighting, structural editing), and the comma-command
dispatch are all in one file — genuine tangled coupling, not breadth.

Noted in review of #2265 (the mouse-support PR, which added 4 lines and did
not create the condition). A split is a refactor with its own review
surface, so it should land as a focused follow-up rather than be bundled
into a feature PR.

Natural seams to consider: the highlighting/scanning code and its tests
(scanHighlight, highlightCallback, theme application), and the comma
commands (dispatch + handlers), both of which are mostly independent of the
main loop. Behavior must stay identical — the unit tests in the file cover
the pure functions; the pty smoke tests cover the loop.

baijum:
Fixes #2266 — `repl.zig` was ~1,590 lines, over the repo's 1,500-line guideline, with genuine tangled coupling: the REPL loop, the isocline callbacks, and the comma-command dispatch all in one file.

This is a **pure-motion refactor** — no behavior changes. The unit tests that cover the pure functions and the pty smoke tests that cover the loop pass unchanged.

## The split (three seams → three new files)

| File | Contents |
|---|---|
| `src/repl_highlight.zig` | Token scanner (`scanHighlight`), the isocline highlighter callback, and the theme→isocline style bridge (`ansiToIcStyle`, `applyTheme`), with all of its unit tests. Driven by `Reader.isDelimiter` and `config.zig`'s theme escapes, so colors cannot disagree with the parse. |
| `src/repl_commands.zig` | Comma-command dispatch (`handleCommand`, tri-state so `,quit` can *end* the REPL rather than continue it), the handlers, the command-name completion helpers, the usage table. |
| `src/repl_eval.zig` | The read → compile → execute → print driver (`evalInputInner` and friends, `EvalMode`), shared by the main loop and the commands, plus the pretty-print terminal width. |

`src/repl.zig` keeps the main loop, line editing, the completeness/completion/sexp-edit callbacks, and `inputIncomplete`'s tests — **1,592 → 489 lines**. Every file is well under the guideline. The dependency graph is acyclic: `repl → {highlight, eval, commands}`, `commands → eval`.

Also: dropped the write-only `theme` global (assigned at startup, never read), and updated the REPL file map in `docs/dev/repl.md` plus the `repl.zig` file references in `docs/dev/crash-reporting.md` / `docs/dev/porting.md`.

## Verification (behavior-identical vs. the pre-split binary)

- `zig build`, `zig build test`, `zig build test -Dgc-stress=true`, `zig build wasm` — all green
- Full Scheme suite: **702 files + 1,395 R7RS tests = 2,097 pass, 0 fail**
- All six `repl-*` pty smoke tests pass (completion, multi-line history, paste, structural editing, mouse, multiple values)
- A pty-driven pass over every comma command (`,time` `,type` `,env` `,version` `,gc` `,help` `,describe` `,apropos` `,expand` `,import` `,break` `,breakpoints` `,condition` `,delete all` `,profile` `,dis` `,step` `,quit` + unknown-command) produced **byte-identical output to the original binary** (the 19-command matrix ran against both builds)

## Related issues

Scanned open issues for anything that could fold in cleanly: the only REPL-adjacent candidates (e.g. #1061, the toplevel-driver extraction that created `toplevel_driver.zig`) are closed, and there are no other open file-size issues. Per #2266's own note — *"a split is a refactor with its own review surface, so it should land as a focused follow-up"* — this PR stays scoped to the split.


baijum:
Thanks for the review — I went through every comment critically. Six of the seven inline comments target code that was **moved verbatim** from the original `repl.zig` (confirmed against base `55708cfa`), so I evaluated each on its technical merits rather than treating the PR as new code. Here's the disposition:

**Fixed (in the new commit `bfabc4c8`):**

1. **`writeBarrier` in `,import` (Major)** — valid. Every other pair-spine mutation in the codebase barriers (`primitives_fiber.zig`, `bytecode_file_read.zig`, `expander.zig`), and `.claude/rules/gc-safety.md` is explicit: *"Omitting this corrupts the generational GC during minor collections."* Added `vm.gc.writeBarrier(tail_obj, pair)` after the `cdr` write. Behavior-identical (only records an old→young edge). I also made `pair` `const` and dropped the `_ = &pair` suppression (comment 3) — same lines.

2. **Unrooted datum/expansion in `,expand` (Major)** — valid. `evalInputInner` (the REPL's own eval path) roots read datums with `pushRoot(&expr); defer popRoot()`; the `,expand` handler was the exception. Rooted `expr` before `expandMacro` and `expanded` before `stripUsertextMarkers`, matching that pattern (LIFO-safe: `expanded`'s pop fires before `expr`'s). Behavior-identical.

3. **`containsSubstring` → `std.mem.indexOf` (Trivial)** — done; empty-needle semantics match (`indexOf` returns 0, the old loop returned true).

4. **`describeSymbol` unused `allocator` param (Trivial)** — done; dropped the parameter and the `_ = allocator` discard.

5. **`docs/dev/repl.md` intro (Minor)** — done; the intro now names all four REPL files and links to the Key files table. This one was about my change, and the point was correct.

**Deliberately not done here — with reasons:**

6. **`,load` path escaping (Major)** — real, but pre-existing, and fixing it *changes observable behavior* (paths containing `"`/`\` start working), which contradicts this PR's "behavior must stay identical" mandate, and per repo convention a bug fix needs its own regression test. I've filed it as a separate issue: [#2273](https://github.com/kaappi/kaappi/issues/2271).

7. **DRY-extract the duplicated print blocks in `repl_eval.zig` (Trivial)** — the two copies are pre-existing and identical; extracting them restructures the eval loop's `catch continue` control flow (your own caveat notes the helper must return a signal). That's a rewrite, not a split — the risk belongs in a dedicated follow-up, not a pure-motion PR whose whole point is that behavior is byte-identical.

For what it's worth, I also tried to reproduce the two GC issues empirically before fixing them: under `-Dgc-stress=true` (collection on every allocation) neither the original nor the fixed binary crashes in a pty-driven `,import`/`,expand` session — they're latent, which is exactly why they're cheap to harden now and why the barrier/rooting follow the documented rules regardless.

PR #2271, opened 2026-08-09
Keep an inexact zero imaginary part complex in make-rectangular, write, and c64/c128 decode (Fixes #2269)

### LINKED ISSUE #2269: make-rectangular demotes an inexact zero imaginary part but the reader keeps it complex — R7RS favors the reader (keep complex, Option B)
## Summary

Since #2166 stores complex components as `Value`s, the two construction sites
disagree about an **inexact zero imaginary part**: `make-rectangular` demotes it
to a real, but the reader keeps it complex.

```scheme
(real? (make-rectangular 1.5 0.0))              ; => #t   demoted to the real 1.5
(real? (read (open-input-string "1.5+0.0i")))   ; => #f   stays Complex{1.5, 0.0}
```

An **exact** zero imag demotes in both places, so the split is specifically the
*inexact* zero.

**Per R7RS, the reader is correct and `make-rectangular` is the outlier.** Adopt
**Option B**: `make-rectangular` keeps an inexact-zero-imag value complex,
demoting only an exact zero — plus a printer fix so the value round-trips through
`write`/`read`.

## What R7RS says

Two passages in §6.2.6 are in tension (a known spec defect, resolved by the
worked examples):

- **Prose:** *"If z is a complex number, then `(real? z)` is true if and only if
  `(zero? (imag-part z))` is true."* By this alone `1.5+0.0i` would be real.
- **Worked examples contradict the prose for the inexact case:**

  ```scheme
  (real? -2.5+0i)    => #t     ; exact zero imag -> real
  (real? -2.5+0.0i)  => #f     ; inexact zero imag -> NOT real
  ```

The concrete, testable example is authoritative: **an explicitly inexact zero
imaginary part keeps the value complex; only an exact zero demotes.** This is
representation-agnostic — it holds whether the value came from the reader or from
`make-rectangular`. The official suite pins it (`r7rs-tests.scm:760`).

The **`make-rectangular` contract** (§6.2.6) reinforces this: it says
`real-part`/`imag-part` *"may return exact real numbers when applied to an
inexact complex number **if the corresponding argument passed to
make-rectangular was exact**"* — the exactness of each component must be
preserved. `(make-rectangular 1.5 0.0)` is the same kind of value as the literal
`-2.5+0.0i`, and the spec draws no distinction between constructor and literal.
So `real?` of it must be `#f`, matching the reader.

## Why Option A ("keep the split") is not spec-clean

Beyond the constructor/reader inconsistency, the status quo also violates
**§6.2.7**: `(write 1.5+0.0i)` prints `1.5` (the printer's `+0.0`-imag collapse
branch), which reads back as the real. §6.2.7 requires
`(eqv? number (string->number (number->string number)))`, and `"1.5+0.0i"`
*does* satisfy it — so emitting `"1.5"` is a conformance bug, present today for
any reader-produced `1.5+0.0i`, independent of `make-rectangular`.

## Cross-implementation check

Ran the same probes on four R7RS implementations. **They are unanimous**, and
all four already implement the recommended Option-B + printer-honesty behavior;
kaappi is the lone outlier, in exactly two spots (⚠️).

| Probe | Chez 10.4 | Guile 3.0 | chibi 0.12 | Gambit 4.9 | **kaappi (now)** | kaappi (Option B) |
|---|---|---|---|---|---|---|
| `real? (make-rectangular 1.5 0.0)` | `#f` | `#f` | `#f` | `#f` | **`#t`** ⚠️ | `#f` |
| `real? -2.5+0.0i` (reader) | `#f` | `#f` | `#f` | `#f` | `#f` | `#f` |
| `real? (make-rectangular 1.5 0)` (exact) | `#t` | `#t` | `#t` | `#t` | `#t` | `#t` |
| `real? -2.5+0i` (exact reader) | `#t` | `#t` | `#t` | `#t` | `#t` | `#t` |
| `imag-part (make-rectangular 1.5 0.0)` | `0.0` | `0.0` | `0.0` | `0.` | `0.0` | `0.0` |
| `number->string (make-rectangular 1.5 0.0)` | `"1.5+0.0i"` | `"1.5+0.0i"` | `"1.5+0.0i"` | `"1.5+0.i"` | **`"1.5"`** ⚠️ | `"1.5+0.0i"` |
| `number->string 1.5+0.0i` (reader) | `"1.5+0.0i"` | `"1.5+0.0i"` | `"1.5+0.0i"` | `"1.5+0.i"` | **`"1.5"`** ⚠️ | `"1.5+0.0i"` |
| `eqv? 1.5+0.0i 1.5` | `#f` | `#f` | `#f` | `#f` | `#f` | `#f` |
| `= 1.5+0.0i 1.5` | `#t` | `#t` | `#t` | `#t` | `#t` | `#t` |
| decomp round-trip `(eqv? (make-rect (real-part z) (imag-part z)) z)`, z = reader `1.5+0.0i` | `#t` | `#t` | `#t` | `#t` | **`#f`** ⚠️ | `#t` |

Takeaways:

- All four keep the inexact-zero-imag value complex through `make-rectangular`
  (`real?` => `#f`) — confirming Option B on top of the R7RS `real?` example.
- All four print it with the imaginary part intact (`1.5+0.0i`, Gambit
  `1.5+0.i`), never collapsing to `1.5` — confirming the §6.2.7 printer fix.
- All four give `(eqv? 1.5+0.0i 1.5)` => `#f` with `(= …)` => `#t`, so kaappi's
  `eqv?` `#f` is already correct and stays. This makes the four tested
  implementations *unanimous against* Racket's `#t` on that specific point —
  Racket is the outlier there, not Guile.
- The decomposition round-trip only *fails* in kaappi when `z` comes from the
  **reader**; starting from `(make-rectangular 1.5 0.0)` kaappi has already
  demoted `z` to a real, so it passes trivially for the wrong reason. The
  discriminating probe must start from the reader value `1.5+0.0i` (last row).

## Recommended fix (Option B + printer honesty)

1. **`make-rectangular`** — route through the exact-zero-only demotion
   (`makeComplexOrRealLiteral`) that the reader already uses, instead of
   `makeComplexOrRealV`. — `src/primitives_numeric.zig:1598`. This does **not**
   break the `make-rectangular` round-trip: `real-part` still returns `1.5` and
   `imag-part` still returns `0.0` (`imag-part` of an inexact real already
   returns `0.0`, `src/primitives_numeric.zig:1624`).
2. **Printer** — emit an explicit `1.5+0.0i` for an inexact-zero-imag complex
   instead of collapsing to `1.5`, so `number->string`/`read` round-trips
   (§6.2.7).
3. **Tests** — re-pin the kaappi-authored srfi160 *control* assertion (and the
   c128 zero-imag decode expectations) to `#f`
   (`tests/scheme/audit/primitives_srfi160-audit.scm:242-253`); this assertion
   is not a SRFI-160 requirement (SRFI 160 governs vector storage, not `real?`).
   Keep `(real? -2.5+0.0i)` => `#f` green (`r7rs-tests.scm:760`). Add regressions
   for the decomposition round-trip
   `(eqv? (make-rectangular (real-part z) (imag-part z)) z)` and the
   `write`/`read` round-trip, for `z = 1.5+0.0i`.

After this, four properties hold at once: the R7RS `real?` example (`#f`), the
`make-rectangular` component round-trip, the §6.2.7 `number->string` round-trip,
and the decomposition round-trip.

## Ruled out

- **Option C (reader demotes inexact-zero imag)** — contradicts the R7RS `real?`
  example and breaks the conformance pin `(real? -2.5+0.0i)` => `#f`.
- **`eqv?` latitude is fine, not a bug:** `(eqv? 1.5+0.0i 1.5)` => `#f` while
  `(= ...)` => `#t` is permitted by §6.1 (numerically-equal inexacts
  distinguishable by `real?`), and is the same philosophy as the pinned
  `(eqv? 0.0+1i -0.0+1i)` => `#f`. Chez, Guile, chibi, and Gambit all agree with
  kaappi here (`#f`); Racket is the outlier (`#t`). Leave as-is.

baijum:
## Summary

Fixes #2269. Since #2166 stored complex components as `Value`s, construction
sites disagreed about an **inexact zero imaginary part**: `make-rectangular`
demoted `(make-rectangular 1.5 0.0)` to the real `1.5`, while the reader kept
`1.5+0.0i` complex; the arithmetic tower and `make-polar` demoted too, and the
printer collapsed `(write 1.5+0.0i)` to `"1.5"` — which reads back as a
different value, violating R7RS §6.2.7's `number->string` round-trip.

Per R7RS §6.2.6's worked examples, an explicitly **inexact** zero imaginary
part keeps the value complex (`(real? -2.5+0.0i)` => `#f`); only an **exact**
zero demotes (`(real? -2.5+0i)` => `#t`). I verified against Chez 10.4, Guile
3.0, chibi 0.12, and Gambit 4.9 — all four keep the complex everywhere
(construction, arithmetic, `make-polar`, and the transcendentals); kaappi was
the lone outlier.

## Changes

1. **One exact-zero-only construction rule.** `makeComplexOrRealV` and
   `makeComplexOrRealLiteral` differed only in their demotion condition and are
   now collapsed into a single `makeComplexOrRealV` (`src/primitives_numeric.zig`)
   — the one Value-component complex builder used by the reader,
   `string->number`, `make-rectangular`, the arithmetic tower (`+ - * /`), and
   `exact`/`inexact` conversion. An inexact zero imag stays complex; only an
   exact zero demotes.
2. **`make-polar`** (`src/primitives_numeric.zig`) — demotes only an exact
   zero angle (`(make-polar 1.5 0)` => `1.5`); an inexact zero imag stays
   complex (`(make-polar 1.5 0.0)` => `1.5+0.0i`), `-0.0` preserved
   (`(make-polar 1.5 -0.0)` => `1.5-0.0i`), and `(make-polar 0.0 1.0)` stays
   `0.0+0.0i` — all matching the references.
3. **Printer** (`src/printer.zig`) — the `.complex` arm collapses only an
   *exact* zero imag; an inexact ±0.0 prints in full (`"1.5+0.0i"` /
   `"1.5-0.0i"`), so `write` and `number->string` round-trip through
   `read`/`string->number`.
4. **SRFI-160 decode** (`src/primitives_srfi160.zig`) — c64/c128
   `decodeElement` preserves a `+0.0` imaginary part instead of demoting to a
   plain real, so vector refs agree with the standalone constructor.

## Tests

- Regressions in `tests/scheme/compliance/complex-exactness-2166-2167.scm`
  for the decomposition round-trip and `write`/`read` +
  `number->string`/`string->number` round-trips (starting from the reader
  value, the discriminating probe), plus the arithmetic (`+ - * /`) and
  `make-polar` cases from the review.
- Re-pinned the srfi160 audit control/decode assertions, `complex-neg-zero.scm`,
  and the Zig unit test for the new constructor/printer behavior.
- Exact-zero survival pins: `(- z z)` => `0`, `(exact 1.5+0.0i)` => `3/2`,
  `(make-polar 1.5 0)` => `1.5`, `(integer? 3+0i)` => `#t` — all still demote.
- `(real? -2.5+0.0i)` => `#f` stays green in `r7rs-tests.scm:760`.

## Known remaining divergence (unchanged, deliberate)

The f64 transcendental and power paths — `sin`/`cos`/`tan`, `sqrt`, `expt`,
`exp`, `log`, `asin` — still demote an exactly-zero (or, for `exp`/`log`/`asin`,
below-1e-15-noise) imaginary result via the pre-existing `makeComplexOrReal`
and its epsilon guard. That noise-suppression design is orthogonal to the
construction-site rule this PR fixes, the references themselves disagree about
noise (e.g. Guile zeroes `(expt 1.0+2.0i 2)` to `-3.0+4.0i` where Chez keeps
`-2.9999999999999987+4.0i`), and the reviewer's recommendation scoped to
construction + arithmetic + `make-polar`. Flagging it here so it is not shipped
silently; a follow-up could revisit it.

## Checklist

- [x] `zig build test` passes (also under `-Dgc-stress=true`)
- [x] Scheme test suites pass (`bash tests/scheme/run-all.sh`): 701 scheme
  files + 1395 R7RS tests, 0 failures
- [x] Documentation updated (CHANGELOG.md)

### DIFF
--- a/src/primitives_numeric.zig
+++ b/src/primitives_numeric.zig
@@ -1033,7 +1033,9 @@ pub fn isZeroValue(v: Value) bool {
 /// inexact, so a stored complex is never mixed-exactness (kaappi#2166) — and
 /// demoting to the real component when the imaginary part is zero, so every
 /// complex constructed here has a nonzero imaginary part. The construction
-/// site behind make-rectangular and exact/inexact conversions.
+/// site behind exact/inexact conversions and the arithmetic tower
+/// (kaappi#2269 moved make-rectangular to makeComplexOrRealLiteral so an
+/// inexact zero imag stays complex there).
 pub fn makeComplexOrRealV(gc: *memory.GC, real_in: Value, imag_in: Value) PrimitiveError!Value {
     var real = real_in;
     var imag = imag_in;
@@ -1592,10 +1594,13 @@ fn makeRectangular(args: []const Value) PrimitiveError!Value {
     for (args) |a| {
         if (!isComponentValue(a)) return primitives.typeError("make-rectangular", "real number", a);
     }
-    // The exactness rule (R7RS 6.2.2) and the zero-imag demotion live in
-    // makeComplexOrRealV, so make-rectangular never touches an f64:
-    // 2^53+1 and 10^25 survive digit-exactly (kaappi#2166).
-    return makeComplexOrRealV(gc, args[0], args[1]);
+    // Components are never forced through an f64: 2^53+1 and 10^25 survive
+    // digit-exactly (kaappi#2166). The exactness rule (R7RS 6.2.2) and the
+    // zero-imag demotion live in makeComplexOrRealLiteral — the same
+    // construction the reader uses, so an INEXACT zero imaginary part stays
+    // complex ((real? (make-rectangular 1.5 0.0)) => #f, kaappi#2269) and
+    // only an exact zero demotes, matching the literal 1.5+0.0i.
+    return makeComplexOrRealLiteral(gc, args[0], args[1]);
 }
 
 fn makePolar(args: []const Value) PrimitiveError!Value {
--- a/src/primitives_numeric.zig
+++ b/src/primitives_numeric.zig
@@ -364,10 +364,11 @@ pub fn inexactFn(args: []const Value) PrimitiveError!Value {
     }
     if (types.isComplex(args[0])) {
         const c = types.toComplex(args[0]);
-        // Already all-inexact: return unchanged. Rebuilding through
-        // makeComplexOrRealV would demote a stored inexact zero imag
-        // (-2.5+0.0i) to the real -2.5, losing the complexness the reader
-        // deliberately keeps ((real? -2.5+0.0i) => #f, kaappi#2166).
+        // Already all-inexact: return unchanged rather than rebuild. With
+        // makeComplexOrRealV's exact-zero-only demotion (kaappi#2269) a
+        // rebuild would preserve the stored inexact zero imag anyway
+        // ((real? -2.5+0.0i) => #f); this is a no-alloc shortcut, not a
+        // correctness guard.
         if (!types.isExactNumber(c.real)) return args[0];
         const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
         const real_i = try inexactFn(&[1]Value{c.real});
@@ -1031,31 +1032,20 @@ pub fn isZeroValue(v: Value) bool {
 /// Build a complex number from two real component Values, applying the
 /// R7RS 6.2.2 inexactness rule — if either component is inexact both become
 /// inexact, so a stored complex is never mixed-exactness (kaappi#2166) — and
-/// demoting to the real component when the imaginary part is zero, so every
-/// complex constructed here has a nonzero imaginary part. The construction
-/// site behind exact/inexact conversions and the arithmetic tower
-/// (kaappi#2269 moved make-rectangular to makeComplexOrRealLiteral so an
-/// inexact zero imag stays complex there).
+/// demoting to the real component only when the imaginary part is an EXACT
+/// zero. An inexact zero imag keeps the value complex ((real? -2.5+0.0i)
+/// => #f, kaappi#2269) at every construction site — the reader,
+/// string->number, make-rectangular, the arithmetic tower (+ - * /), and
+/// exact/inexact conversion — matching Chez, Guile, chibi, and Gambit.
+/// This single rule replaced makeComplexOrRealLiteral (kaappi#2269); an
+/// exact zero imag demotes ((integer? 3+0i) => #t).
 pub fn makeComplexOrRealV(gc: *memory.GC, real_in: Value, imag_in: Value) PrimitiveError!Value {
     var real = real_in;
     var imag = imag_in;
-    if (!types.isExactNumber(real) or !types.isExactNumber(imag)) {
-        real = try inexactFn(&[1]Value{real});
-        imag = try inexactFn(&[1]Value{imag});
-    }
-    if (isZeroValue(imag)) return real;
-    return gc.allocComplex(real, imag) catch return PrimitiveError.OutOfMemory;
-}
-
-/// Reader/string->number complex construction: like makeComplexOrRealV, but
-/// only an EXACT zero imaginary part demotes to the real component. The R7RS
-/// conformance suite pins both sides: (real? -2.5+0.0i) => #f (an inexact
-/// zero imag written in a literal stays a complex) and (integer? 3+0i) => #t
-/// (an exact zero imag demotes). Inexactness still propagates whole-number
-/// (R7RS 6.2.2), so the two parsers agree (R7RS 6.2.7, kaappi#2166).
-pub fn makeComplexOrRealLiteral(gc: *memory.GC, real_in: Value, imag_in: Value) PrimitiveError!Value {
-    var real = real_in;
-    var imag = imag_in;
+    // Exactness of the zero is read off the INPUT imag: an inexact 0.0
+    // must stay complex, while an exact 0 (fixnum/bignum/rational zero)
+    // demotes — after the whole-number inexactness propagation below the
+    // imag would be an indistinguishable 0.0 flonum.
     const imag_exact_zero = types.isExactNumber(imag) and isZeroValue(imag);
     if (!types.isExactNumber(real) or !types.isExactNumber(imag)) {
         real = try inexactFn(&[1]Value{real});
@@ -1513,7 +1503,7 @@ pub fn parseNumberText(gc: *@import("memory.zig").GC, text: []const u8, radix_in
             const imag_str = body[sp..];
             // Components are built digit-exactly; exactness is carried by
             // each component's own type (integer/rational parts are exact,
-            // decimals and exponents inexact), and makeComplexOrRealLiteral
+            // decimals and exponents inexact), and makeComplexOrRealV
             // normalizes the pair to a single exactness. `#e`/`#i` still
             // override everything via applyExactness below (#2243 review).
             const real_v = if (real_str.len == 0)
@@ -1528,7 +1518,7 @@ pub fn parseNumberText(gc: *@import("memory.zig").GC, text: []const u8, radix_in
                 types.makeFixnum(-1)
             else
                 parseComplexComponent(gc, imag_str, radix) orelse return types.FALSE;
-            const c = makeComplexOrRealLiteral(gc, slot.get(), imag_v) catch return PrimitiveError.OutOfMemory;
+            const c = makeComplexOrRealV(gc, slot.get(), imag_v) catch return PrimitiveError.OutOfMemory;
             return applyExactness(gc, c, exactness);
         } else {
             // No split found -- pure imaginary with a magnitude: +3i,
@@ -1577,7 +1567,7 @@ pub fn parseNumberText(gc: *@import("memory.zig").GC, text: []const u8, radix_in
                     imag_v = types.makeFlonum(f);
                 }
             }
-            const c = makeComplexOrRealLiteral(gc, types.makeFixnum(0), imag_v) catch return PrimitiveError.OutOfMemory;
+            const c = makeComplexOrRealV(gc, types.makeFixnum(0), imag_v) catch return PrimitiveError.OutOfMemory;
             return applyExactness(gc, c, exactness);
         }
     }
@@ -1596,19 +1586,27 @@ fn makeRectangular(args: []const Value) PrimitiveError!Value {
     }
     // Components are never forced through an f64: 2^53+1 and 10^25 survive
     // digit-exactly (kaappi#2166). The exactness rule (R7RS 6.2.2) and the
-    // zero-imag demotion live in makeComplexOrRealLiteral — the same
-    // construction the reader uses, so an INEXACT zero imaginary part stays
-    // complex ((real? (make-rectangular 1.5 0.0)) => #f, kaappi#2269) and
-    // only an exact zero demotes, matching the literal 1.5+0.0i.
-    return makeComplexOrRealLiteral(gc, args[0], args[1]);
+    // exact-zero-only demotion live in makeComplexOrRealV — the same
+    // construction the reader and arithmetic use, so an INEXACT zero
+    // imaginary part stays complex ((real? (make-rectangular 1.5 0.0))
+    // => #f, kaappi#2269) and only an exact zero demotes, matching the
+    // literal 1.5+0.0i.
+    return makeComplexOrRealV(gc, args[0], args[1]);
 }
 
 fn makePolar(args: []const Value) PrimitiveError!Value {
+    const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
     const mag = try toF64(args[0]);
     const ang = try toF64(args[1]);
     const real = mag * @cos(ang);
     const imag = mag * @sin(ang);
-    return makeComplexOrReal(real, imag);
+    // An inexact zero imaginary part keeps the value complex
+    // ((make-polar 1.5 0.0) => 1.5+0.0i, kaappi#2269), matching the reader,
+    // make-rectangular, and arithmetic. Only an EXACT zero angle — which
+    // produces an exact zero imag — demotes ((make-polar 1.5 0) => 1.5).
+    // Verified against Chez, Guile, chibi, and Gambit.
+    if (imag == 0.0 and types.isExactNumber(args[1])) return types.makeFlonum(real);
+    return gc.allocComplex(types.makeFlonum(real), types.makeFlonum(imag)) catch return PrimitiveError.OutOfMemory;
 }
 
 fn realPart(args: []const Value) PrimitiveError!Value {
--- a/src/primitives_numeric.zig
+++ b/src/primitives_numeric.zig
@@ -1603,9 +1603,12 @@ fn makePolar(args: []const Value) PrimitiveError!Value {
     // An inexact zero imaginary part keeps the value complex
     // ((make-polar 1.5 0.0) => 1.5+0.0i, kaappi#2269), matching the reader,
     // make-rectangular, and arithmetic. Only an EXACT zero angle — which
-    // produces an exact zero imag — demotes ((make-polar 1.5 0) => 1.5).
+    // produces an exact zero imag — demotes ((make-polar 1.5 0) => 1.5);
+    // the angle must be numerically zero, not merely exact, so a tiny exact
+    // nonzero angle whose sin underflows to 0.0 stays complex
+    // ((make-polar 1.5 (/ 1 (expt 10 400))) => 1.5+0.0i).
     // Verified against Chez, Guile, chibi, and Gambit.
-    if (imag == 0.0 and types.isExactNumber(args[1])) return types.makeFlonum(real);
+    if (imag == 0.0 and types.isExactNumber(args[1]) and isZeroValue(args[1])) return types.makeFlonum(real);
     return gc.allocComplex(types.makeFlonum(real), types.makeFlonum(imag)) catch return PrimitiveError.OutOfMemory;
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,14 +34,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   through the normal numeric printer, so `write` and `real-part` agree;
   and the reader and `string->number` build components digit-exactly at
   any size. Per R7RS 6.2.2 a stored complex is never mixed-exactness (an
-  inexact operand makes both components inexact), and a zero imaginary
-  part demotes to the real component — except that a literal's *inexact*
-  zero imag stays complex (`(real? -2.5+0.0i)` => #f) while an exact one
-  demotes (`(integer? 3+0i)` => #t). The `.sbc` constant encoding stores
+  inexact operand makes both components inexact), and an *exact* zero
+  imaginary part demotes to the real component (`(integer? 3+0i)` => #t)
+  while an *inexact* zero imag keeps the value complex (`(real?
+  -2.5+0.0i)` => #f). The `.sbc` constant encoding stores
   the two component constants instead of f64+flags. This dissolves the
   #2182/#2243 f64 round-trip gates in the reader and `string->number`:
   `9007199254740993+1i`, `#x20000000000001+2i`, and
   `10000000000000000000000000/3+1i` all read digit-exactly now.
+
+- **`make-rectangular` keeps an inexact zero imaginary part complex, and
+  `write`/`number->string` print it in full** (#2269). The reader already
+  kept `1.5+0.0i` complex (`(real? 1.5+0.0i)` => #f), but
+  `(make-rectangular 1.5 0.0)` demoted to the real `1.5` and the printer
+  collapsed an inexact-zero-imag complex to its bare real part, so
+  `(write 1.5+0.0i)` printed `1.5` — which reads back as a different
+  value, violating R7RS 6.2.7's `number->string` round-trip. Both
+  construction and printing now match the reader: only an exact zero imag
+  demotes, `(write 1.5+0.0i)` emits `"1.5+0.0i"`, and the decomposition
+  round-trip `(eqv? (make-rectangular (real-part z) (imag-part z)) z)`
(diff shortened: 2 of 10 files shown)


src/primitives_numeric.zig
```diff
@@ -1033,7 +1033,9 @@ pub fn isZeroValue(v: Value) bool {
 /// inexact, so a stored complex is never mixed-exactness (kaappi#2166) — and
 /// demoting to the real component when the imaginary part is zero, so every
 /// complex constructed here has a nonzero imaginary part. The construction
-/// site behind make-rectangular and exact/inexact conversions.
+/// site behind exact/inexact conversions and the arithmetic tower
+/// (kaappi#2269 moved make-rectangular to makeComplexOrRealLiteral so an
+/// inexact zero imag stays complex there).
 pub fn makeComplexOrRealV(gc: *memory.GC, real_in: Value, imag_in: Value) PrimitiveError!Value {
```
baijum, line 1042: `makeComplexOrRealV` still demotes **any** zero imag (exact, or inexact after the force-to-inexact). With `make-rectangular` moved off it, its remaining callers are the arithmetic tower (`primitives_arithmetic.zig:367/446/542/659`) and `exact`/`inexact` conversion. For arithmetic that disagrees with all four references: `(+ 1.0+2.0i 1.0-2.0i)` => `2.0` here vs `2.0+0.0i` in Chez/Guile/chibi/Gambit; `(real? (* 1.5+0.0i 2.0))` => `#t` here vs `#f` everywhere else.

Exact-zero-only is now the correct rule for *every* remaining caller (`exact` already produces an exact 0, which should still demote). At that point this function becomes identical to `makeComplexOrRealLiteral` — consider collapsing the two. Note `inexactFn` (lines 365-377) already works around this very demotion for all-inexact complexes.

src/primitives_numeric.zig
```diff
@@ -1592,10 +1594,13 @@ fn makeRectangular(args: []const Value) PrimitiveError!Value {
...
-    // makeComplexOrRealV, so make-rectangular never touches an f64:
-    // 2^53+1 and 10^25 survive digit-exactly (kaappi#2166).
-    return makeComplexOrRealV(gc, args[0], args[1]);
+    // Components are never forced through an f64: 2^53+1 and 10^25 survive
+    // digit-exactly (kaappi#2166). The exactness rule (R7RS 6.2.2) and the
+    // zero-imag demotion live in makeComplexOrRealLiteral — the same
+    // construction the reader uses, so an INEXACT zero imaginary part stays
+    // complex ((real? (make-rectangular 1.5 0.0)) => #f, kaappi#2269) and
+    // only an exact zero demotes, matching the literal 1.5+0.0i.
+    return makeComplexOrRealLiteral(gc, args[0], args[1]);
```
baijum: `make-rectangular` is fixed, but its sibling `make-polar` (just below, via the f64 `makeComplexOrReal`) still demotes a `0.0` imaginary result: `(make-polar 1.5 0.0)` => `1.5` here vs `1.5+0.0i` in Chez/Guile/Gambit. Same bug class — worth fixing in the same pass so all construction sites agree.

baijum (commented):
## Review: correct as far as it goes, but the fix is incomplete — an inexact zero imag still demotes in the arithmetic tower and `make-polar`

I verified the four targeted fixes against Chez 10.4, Guile 3.0, chibi 0.12, and Gambit 4.9, and built this branch to check behavior directly.

**What's right (confirmed):** `make-rectangular`, the reader, the printer, and c64/c128 decode now all keep an inexact zero imaginary part complex and only demote an *exact* zero — matching all four reference implementations and R7RS §6.2.6's `real?` examples. The updated suites pass here (`complex-exactness-2166-2167.scm`, `primitives_srfi160-audit.scm`, `complex-neg-zero.scm`, `srfi231-storage-classes.scm`), and `(real? -2.5+0.0i)` => `#f` stays green. The §6.2.7 printer argument is correct and the regression tests start from the discriminating reader value. Good work.

### Blocking concern: the same demotion survives in two other construction sites

The PR routes `make-rectangular` through `makeComplexOrRealLiteral` (exact-zero-only), but leaves `makeComplexOrRealV` (demotes *any* zero, including an inexact `0.0`) in place for the **arithmetic tower** (`+ - * /`, `primitives_arithmetic.zig:367,446,542,659`), and leaves **`make-polar`** on the f64 `makeComplexOrReal` (also demotes `imag == 0.0`). So the exact inconsistency #2269 set out to remove — two construction sites disagreeing about an inexact zero imag — still exists after this PR; it has just moved from *(reader vs make-rectangular)* to *(make-rectangular/reader vs arithmetic/make-polar)*.

Built from this branch (`a1aee10`), against the four references:

| Expression | Chez | Guile | chibi | Gambit | **this PR** |
|---|---|---|---|---|---|
| `real? (make-rectangular 1.5 0.0)` | `#f` | `#f` | `#f` | `#f` | `#f` ✅ |
| `real? (+ 1.0+2.0i 1.0-2.0i)` (imag cancels) | `#f` | `#f` | `#f` | `#f` | **`#t`** ❌ |
| `(+ 1.0+2.0i 1.0-2.0i)` | `2.0+0.0i` | `2.0+0.0i` | `2.0+0.0i` | `2.+0.i` | **`2.0`** ❌ |
| `real? (* 1.5+0.0i 2.0)` | `#f` | `#f` | `#f` | `#f` | **`#t`** ❌ |
| `real? (- 1.5+2.0i 0.0+2.0i)` | `#f` | `#f` | `#f` | `#f` | **`#t`** ❌ |
| `real? (+ 1.5+0.0i 0)` (add exact 0) | `#f` | `#f` | `#f` | `#f` | **`#t`** ❌ |
| `real? (make-polar 1.5 0.0)` | `#f` | `#f` | `#f` | `#f` | **`#t`** ❌ |
| `real? (exact 1.5+0.0i)` (exact→demote) | `#t` | — | `#t` | `#t` | `#t` ✅ |

The references demote **only an exact zero** everywhere — construction, arithmetic, and `make-polar` alike. `(exact 1.5+0.0i)` => `#t` because the conversion turns the `0.0` into an exact `0`; that path already behaves correctly here and should keep doing so.

The codebase already knows `makeComplexOrRealV` mis-demotes an inexact zero: `inexactFn` (`primitives_numeric.zig:365-377`) special-cases an all-inexact complex specifically *"[because] rebuilding through makeComplexOrRealV would demote a stored inexact zero imag (-2.5+0.0i) to the real -2.5"*. The four arithmetic sites and `make-polar` have no such guard.

### Suggested resolution

`makeComplexOrRealV` and `makeComplexOrRealLiteral` differ *only* in that the former demotes an inexact zero and the latter demotes an exact zero. Once `make-rectangular` moves off it, the exact-zero-only rule is the correct one for **every** remaining caller (arithmetic and `exact`/`inexact`), so the cleanest fix is to make `makeComplexOrRealV` demote exact-zero-only too — at which point the two functions are identical and can be collapsed into one. `make-polar`'s f64 path needs the analogous change (don't demote a `0.0` imaginary result). I confirmed the R7RS suite pins nothing that requires arithmetic to demote an inexact zero (only `(real? -2.5+0.0i)`, `(= 1 1.0 1.0+0.0i)`, `(zero? 0.0+0.0i)`, none of which this would break), so fixing it here is low-risk.

If you'd rather keep this PR narrowly scoped, that's defensible — but then the remaining split should be called out explicitly (CHANGELOG + a follow-up issue) rather than shipped silently, since a reader hitting `(+ 1.0+2.0i 1.0-2.0i)` => `2.0` will reasonably see it as the same bug. My recommendation is to finish it here, plus regression tests for the arithmetic and `make-polar` cases (the reader-decomposition style you already used).

### Minor

- `printer.zig` — with every construction site demoting an *exact* zero, a stored `Complex` can never hold an exact-zero imaginary part, so the new `isZeroComponent(imag) and types.isExactNumber(imag)` collapse branch is effectively unreachable for stored values. That's fine as defensive code and the comment is accurate; just flagging that it now documents an invariant (exact-zero never stored) rather than a live path.

baijum:
Thanks for the thorough review — the claims check out. I verified all of them
directly against the installed Chez 10.4, Guile 3.0, chibi 0.12, and Gambit 4.9,
and they're unanimous: `(+ 1.0+2.0i 1.0-2.0i)` => `2.0+0.0i`, `(* 1.5+0.0i 2.0)`,
`(- 1.5+2.0i 0.0+2.0i)`, `(+ 1.5+0.0i 0)`, and `(make-polar 1.5 0.0)` are all
complex (`real?` => `#f`) in all four, while `(make-polar 1.5 0)` (exact angle)
demotes to `1.5` and `(exact 1.5+0.0i)` demotes to `3/2`.

Implemented your suggested resolution in 0ae22ea8:

- **Collapsed `makeComplexOrRealV` / `makeComplexOrRealLiteral` into one
  exact-zero-only `makeComplexOrRealV`** — now the single Value-component
  construction site for the reader, `string->number`, `make-rectangular`, the
  arithmetic tower, and `exact`/`inexact` conversion. The exactness of the
  zero is read off the input imag (before whole-number inexactness
  propagation), so an inexact `0.0` stays complex and an exact `0` demotes.
- **`make-polar` now demotes only an exact zero angle** — `(make-polar 1.5
  0.0)` => `1.5+0.0i` (real? #f), `(make-polar 1.5 0)` => `1.5`, `-0.0`
  preserved (`1.5-0.0i`), and `(make-polar 0.0 1.0)` stays `0.0+0.0i` — the
  exact-angle-vs-inexact-angle split the references show.
- `(exact 1.5+0.0i)` still demotes to `3/2` and `(- z z)` for an exact `z`
  still gives exact `0`, as pinned by the new tests.

Full suite is green: 701 scheme files + 1395 R7RS tests, 0 failures; `zig
build test` and `-Dgc-stress=true` both pass.

One deliberate boundary, called out in the PR body and CHANGELOG rather than
shipped silently: the **f64 transcendental paths** (`sin`/`cos`/`tan`, `sqrt`,
`expt`, `exp`, `log`, `asin`) still demote an exactly-zero or below-1e-15-noise
imaginary result via `makeComplexOrReal` and its pre-existing epsilon guard.
That's a separate noise-suppression design (the references themselves disagree
about noise — e.g. Guile zeroes `(expt 1.0+2.0i 2)` to `-3.0+4.0i` while Chez
keeps `-2.9999999999999987+4.0i`), and your recommendation scoped to
construction + arithmetic + `make-polar`. Happy to follow up separately if
you'd like that revisited.

On the minor point about the printer's exact-zero collapse branch: agreed it's
now unreachable defensive code — with every construction site on the
exact-zero-only rule a stored `Complex` can never hold an exact-zero imag. The
comment documents the invariant; keeping the branch guards against a future
`.sbc` constant or hand-built object slipping through.

src/primitives_numeric.zig
```diff
@@ -1592,18 +1584,29 @@ fn makeRectangular(args: []const Value) PrimitiveError!Value {
...
     const ang = try toF64(args[1]);
     const real = mag * @cos(ang);
     const imag = mag * @sin(ang);
-    return makeComplexOrReal(real, imag);
+    // An inexact zero imaginary part keeps the value complex
+    // ((make-polar 1.5 0.0) => 1.5+0.0i, kaappi#2269), matching the reader,
+    // make-rectangular, and arithmetic. Only an EXACT zero angle — which
+    // produces an exact zero imag — demotes ((make-polar 1.5 0) => 1.5).
+    // Verified against Chez, Guile, chibi, and Gambit.
+    if (imag == 0.0 and types.isExactNumber(args[1])) return types.makeFlonum(real);
```
baijum: **Confirmed bug (agrees with CodeRabbit):** this guard checks the angle's *exactness*, not that it is *zero*. A tiny exact **non-zero** angle whose `sin` underflows to `0.0` is still an exact number, so it demotes:

`(make-polar 1.5 (/ 1 (expt 10 400)))` => `1.5` here (built at `0ae22ea`), but `1.5+0.0i` in Chez/Guile/chibi/Gambit; `real?` is `#t` here vs `#f` in all four.

Add the zero check:
```zig
if (imag == 0.0 and types.isExactNumber(args[1]) and isZeroValue(args[1]))
    return types.makeFlonum(real);
```
and a regression for the underflow angle — the current `make-polar` pins don't exercise this path.

baijum (commented):
## Follow-up review of `0ae22ea` — main finding resolved; one real edge-case bug remains in `make-polar`

Thanks — the follow-up commit resolves my main finding cleanly, and collapsing `makeComplexOrRealLiteral` into a single exact-zero-only `makeComplexOrRealV` is exactly the right shape (one construction site, one rule). I rebuilt at `0ae22ea` and confirmed the arithmetic tower now matches all four references:

```
real? (+ 0.5+0.0i 1.0)        => #f
real? (* 1.5+0.0i 2.0)        => #f
real? (- 1.5+2.0i 0.0+2.0i)   => #f
real? (+ 1.0+2.0i 1.0-2.0i)   => #f     (writes 2.0+0.0i)
real? (+ 1.5+0.0i 0)          => #f
real? (exact 1.5+0.0i)        => #t     (exact zero still demotes ✔)
```

### Confirmed bug: `make-polar`'s guard demotes a non-zero angle that underflows

CodeRabbit's `make-polar` comment is correct, and I verified it against the references and a fresh build. The guard on line 1608 tests the angle's **exactness**, not that it is **zero**:

```zig
if (imag == 0.0 and types.isExactNumber(args[1])) return types.makeFlonum(real);
```

A tiny exact *non-zero* angle whose `sin` underflows to `0.0` is exact, so it demotes:

| | Chez | Guile | chibi | Gambit | **this branch** |
|---|---|---|---|---|---|
| `(make-polar 1.5 (/ 1 (expt 10 400)))` | `1.5+0.0i` | `1.5+0.0i` | `1.5+0.0i` | `1.5+0.i` | **`1.5`** ❌ |
| `real?` of it | `#f` | `#f` | `#f` | `#f` | **`#t`** ❌ |

All four references keep it complex; this branch returns the real `1.5`. CodeRabbit's fix is right — also require the angle to be numerically zero:

```zig
if (imag == 0.0 and types.isExactNumber(args[1]) and isZeroValue(args[1]))
    return types.makeFlonum(real);
```

Please add `(make-polar 1.5 (/ 1 (expt 10 400)))` => not `real?` as a regression alongside the existing `make-polar` pins — none of the current cases exercise the underflow path, which is why it slipped through green.

### One thing worth knowing about the exact-zero-angle case (not a change request)

Implementations actually *split* on `(make-polar 1.5 0)` (exact zero angle): Chez, Guile, and Gambit demote to real `1.5` (`real?` => `#t`), but **chibi keeps it complex** (`1.5+0.0i`, `#f`). This branch demotes, matching the 3-of-4 majority and its own "exact zero demotes" rule — a defensible choice, so no change needed; just flagging that this specific pin (`(make-polar 1.5 0)` => `1.5`) is not universal, unlike everything else in this PR where the four are unanimous.

### The other two bot comments (both low, agree)

- `tests_advanced.zig` using `th.TestContext` instead of a raw `GC`/VM — reasonable per the repo's test-helper convention; take it or leave it, it doesn't affect coverage.
- `complex-neg-zero.scm` importing `(scheme process-context)` for `exit` — matches the smoke/audit convention and is a safe tidy-up; the test passes today because kaappi registers `exit` ambiently in script mode.

With the `make-polar` guard fixed and its regression added, this looks complete and correct to me — every construction site would then agree with the reference set (modulo the exact-zero-angle case above, which the implementations themselves disagree on).

baijum:
Thanks for the follow-up — all three comments addressed in `81cc6d1d`, and I
verified each against the references before fixing.

1. **`make-polar` underflow (confirmed bug)** — you and CodeRabbit are both
   right. My guard tested the angle's *exactness*, not that it is *zero*: a
   tiny exact nonzero angle underflows through `toF64`/`sin` to `0.0`, so
   `(make-polar 1.5 (/ 1 (expt 10 400)))` demoted to the real `1.5` while
   Guile, Chez, and Gambit (and chibi, per your run) all keep `1.5+0.0i`
   (real? => #f). Fixed with `isZeroValue(args[1])` alongside the exactness
   check — the exact-zero angle pin `(make-polar 1.5 0)` => `1.5` is
   unchanged (matching the 3-of-4 majority; chibi's split is noted). Added the
   underflow regression, which the existing pins didn't exercise.
2. **`tests_advanced.zig`** — converted the make-rectangular test to
   `th.TestContext`, per `docs/dev/testing.md`'s documented multi-evaluation
   helper.
3. **`complex-neg-zero.scm`** — added the explicit `(scheme process-context)`
   import for `exit`.

Full verification on `81cc6d1d`: `zig build test` and `-Dgc-stress=true`
pass; `bash tests/scheme/run-all.sh` green (701 files + 1395 R7RS tests, 0
failures); `zig fmt --check` clean. CI is green across freebsd/netbsd/openbsd/
riscv64/s390x/ppc64le/gc-stress/gc-stress-scheme, and CodeRabbit passed.

Thanks for the sharp catch on the exactness-vs-zero distinction — that's
exactly the kind of edge that slips through when the pins only use `0`/`0.0`.

baijum (commented):
## Verified `81cc6d1d` — all findings resolved

Rebuilt at `81cc6d1d` and confirmed the `make-polar` guard fix directly:

```
(make-polar 1.5 (/ 1 (expt 10 400)))  => 1.5+0.0i   real? => #f   ✔ (was 1.5 / #t)
(make-polar 1.5 0.0)                  => 1.5+0.0i   real? => #f   ✔
(make-polar 1.5 0)   (exact angle)    => 1.5        real? => #t   ✔ (unchanged)
```

That now agrees with Chez/Guile/chibi/Gambit on the underflow case, and the exact-zero-angle pin is untouched. The compliance/smoke/audit suites and the converted `tests_advanced` unit test all pass here too.

Both the exactness-vs-zero distinction and the deliberate f64 transcendental-noise boundary are handled and documented. Nothing outstanding from my side — LGTM. Nice, tidy convergence across the three commits.

PR #2274, opened 2026-08-09
,load: build the form as Values so paths with quotes or backslashes load (Fixes #2273)

### LINKED ISSUE #2273: [REPL] ,load does not escape the path when building (load "...") — quotes and backslashes break it
## Problem

`,load <path>` builds a Scheme `(load "...")` form by splicing the path into a
string literal with no escaping (`src/repl_commands.zig`, moved verbatim from
`repl.zig` by #2270):

```zig
const load_expr = std.fmt.bufPrint(&load_buf, "(load \"{s}\")", .{load_path}) catch { ... };
```

The reader's string escapes (`\"`, `\\`, `\t`, …) mean a path containing a
double quote or a backslash breaks or changes the path:

- `,load /tmp/a"b.scm` → `(load "/tmp/a"b.scm")` → reader syntax error.
- `,load C:\tmp\file.scm` → `(load "C:\tmp\file.scm")` → `\t` becomes a TAB →
  wrong path → load fails.

On POSIX quotes/backslashes in filenames are rare, but on **Windows every path
uses backslashes**, so `,load` is effectively broken there for normal use.
The failure is loud (an error is reported), never silent.

Found by CodeRabbit review of #2270 (comment on `src/repl_commands.zig:191`);
deliberately not fixed there because the fix changes observable behavior and
the PR is a pure-motion split.

## Fix

Escape the path (`\` → `\\`, `"` → `\"`, maybe newline) before embedding it
in the `(load "...")` form, or better, load the file by raw bytes without
round-tripping through a Scheme literal. A regression test (pty smoke test,
like `tests/scheme/smoke/repl-*.sh`) should cover a path with an embedded
quote and a backslash.

baijum:
Fixes #2273 — `,load <path>` spliced the path into a `(load "...")` string literal, so the reader's string escapes broke or changed it:

- `,load /tmp/a"b.scm` → `(load "/tmp/a"b.scm")` → `read error[KP1006]: unterminated string literal`
- `,load C:\tmp\file.scm` → `\t` decodes to a TAB → `error[KP3000]: cannot open file ...` (a *different* file); `\s` in `back\slash.scm` → `read error[KP1007]: invalid escape sequence`

On Windows every path uses backslashes, so `,load` was effectively broken there for normal use.

## Fix

Build `(load "path")` as Values instead of text: a `load` symbol, a Scheme string holding the **raw path bytes**, and the two pairs — no escaping to get wrong, and the old 1024-byte path limit is gone. Evaluation goes through a new `repl_eval.evalInputValue`, which shares the compile/execute/print driver with the text path: the loop body of `evalInputInner` was extracted into `evalExpr` (returning whether the batch should continue, preserving the stop-at-first-error behavior) and the duplicated print/store logic into `printResult`, so error reporting, the runtime stack trace, multiple-values printing, and the `_` binding are unchanged.

GC-safety: the two fresh values (`path_val`, `sym_val`) are explicitly rooted across the pair allocations; `allocPair` roots its Value arguments internally per `.claude/rules/gc-safety.md`. Verified under `-Dgc-stress=true`.

## Regression test

`tests/scheme/smoke/repl-load-path-2273.sh` (pty-driven, like the other `repl-*.sh` tests) creates files whose names contain a double quote, a backslash, and the literal `\t` sequence, plus a normal-path control, and asserts each loads (its marker prints, no `error[` line). All three tricky cases fail on the old code with exactly the errors above.

## Verification

- `zig build` ✓, `zig fmt --check` ✓
- `zig build test` ✓, also green under `-Dgc-stress=true`
- `bash tests/scheme/run-all.sh`: 2097 pass, 0 fail (R7RS 1395/1395)
- New smoke test passes with the fix, fails without it

### DIFF
--- a/src/repl_commands.zig
+++ b/src/repl_commands.zig
@@ -185,12 +185,38 @@ pub fn handleCommand(vm: *vm_mod.VM, allocator: std.mem.Allocator, input: []cons
         if (load_path.len == 0) {
             writeStderr(",load requires a file path\n");
         } else {
-            var load_buf: [1024]u8 = undefined;
-            const load_expr = std.fmt.bufPrint(&load_buf, "(load \"{s}\")", .{load_path}) catch {
-                writeStderr("path too long\n");
+            // Build `(load "path")` as Values instead of splicing the path
+            // into a string literal: the reader's escapes (`\"`, `\\`,
+            // `\t`, ...) would mangle a path containing a quote or a
+            // backslash — on Windows every path does (kaappi#2273).
+            // evalInputValue evaluates the form with the same driver as the
+            // text path, errors included.
+            var path_val: types.Value = undefined;
+            vm.gc.pushRoot(&path_val);
+            defer vm.gc.popRoot();
+            path_val = vm.gc.allocString(load_path) catch {
+                writeStderr("out of memory\n");
                 return .handled;
             };
-            repl_eval.evalInput(vm, allocator, load_expr);
+            var sym_val: types.Value = undefined;
+            vm.gc.pushRoot(&sym_val);
+            defer vm.gc.popRoot();
+            sym_val = vm.gc.allocSymbol("load") catch {
+                writeStderr("out of memory\n");
+                return .handled;
+            };
+            // allocPair roots its Value arguments internally, so only the
+            // two locals above need explicit roots across the calls below
+            // (see .claude/rules/gc-safety.md).
+            const args = vm.gc.allocPair(path_val, types.NIL) catch {
+                writeStderr("out of memory\n");
+                return .handled;
+            };
+            const form = vm.gc.allocPair(sym_val, args) catch {
+                writeStderr("out of memory\n");
+                return .handled;
+            };
+            repl_eval.evalInputValue(vm, allocator, form, .normal);
         }
         return .handled;
     }
--- a/src/repl_commands.zig
+++ b/src/repl_commands.zig
@@ -191,23 +191,28 @@ pub fn handleCommand(vm: *vm_mod.VM, allocator: std.mem.Allocator, input: []cons
             // backslash — on Windows every path does (kaappi#2273).
             // evalInputValue evaluates the form with the same driver as the
             // text path, errors included.
-            var path_val: types.Value = undefined;
+            //
+            // Assign before rooting (matching `,expand`/`,import` below): the
+            // slot must hold a valid value before any later call can collect
+            // — allocString/allocSymbol copy their bytes, then maybeCollect()
+            // *before* returning, and markRoots dereferences every rooted
+            // slot, so rooting an undefined slot would let garbage bits that
+            // happen to look like a pointer corrupt the heap (kaappi#2274
+            // review). path_val is rooted before allocSymbol (the next
+            // allocation) and sym_val before the allocPair calls; allocPair
+            // roots its Value arguments internally.
+            var path_val = vm.gc.allocString(load_path) catch {
+                writeStderr("out of memory\n");
+                return .handled;
+            };
             vm.gc.pushRoot(&path_val);
             defer vm.gc.popRoot();
-            path_val = vm.gc.allocString(load_path) catch {
+            var sym_val = vm.gc.allocSymbol("load") catch {
                 writeStderr("out of memory\n");
                 return .handled;
             };
-            var sym_val: types.Value = undefined;
             vm.gc.pushRoot(&sym_val);
             defer vm.gc.popRoot();
-            sym_val = vm.gc.allocSymbol("load") catch {
-                writeStderr("out of memory\n");
-                return .handled;
-            };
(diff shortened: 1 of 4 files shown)


src/repl_commands.zig
```diff
@@ -185,12 +185,38 @@ pub fn handleCommand(vm: *vm_mod.VM, allocator: std.mem.Allocator, input: []cons
...
+            // Build `(load "path")` as Values instead of splicing the path
+            // into a string literal: the reader's escapes (`\"`, `\\`,
+            // `\t`, ...) would mangle a path containing a quote or a
+            // backslash — on Windows every path does (kaappi#2273).
+            // evalInputValue evaluates the form with the same driver as the
+            // text path, errors included.
+            var path_val: types.Value = undefined;
+            vm.gc.pushRoot(&path_val);
+            defer vm.gc.popRoot();
+            path_val = vm.gc.allocString(load_path) catch {
```
baijum: `path_val` is rooted **while still `undefined`**, and the very next line's `allocString` can collect before the assignment lands:

```zig
var path_val: types.Value = undefined;
vm.gc.pushRoot(&path_val);   // roots an undefined slot
defer vm.gc.popRoot();
path_val = vm.gc.allocString(load_path) catch { ... };  // maybeCollect() marks &path_val here, still undefined
```

`allocString` copies the bytes, then calls `maybeCollect()` *before* returning, and `markRoots` reads `root.*` for every rooted slot. During that collection `path_val` holds garbage. If the bits happen to satisfy `isPointer` (top 16 == `0xFFFC`), `markValueInner` dereferences a bogus 48-bit address and writes a mark bit into it → memory corruption / UAF. Same issue for `sym_val` at line 201–204.

It doesn't reproduce in the test/release builds because ReleaseSafe and Debug 0xAA-fill `= undefined` (`0xAAAA… >> 48 != 0xFFFC`, so `isPointer` is false) — which is exactly why `zig build test`, `-Dgc-stress`, and the shipped ReleaseSafe artifacts are all safe. Under `-Doptimize=ReleaseFast` (documented "for maximum throughput") there is no fill and the slot is genuinely undefined.

Every other root in this same file assigns a valid value *before* `pushRoot` (`var expr = er.readDatum() catch …; pushRoot(&expr)`, `var datum = ir.readDatum() catch …; pushRoot(&datum)`, `var import_root = import_list; …`). Matching that convention fixes it at zero cost — assign first, then root:

```zig
var path_val = vm.gc.allocString(load_path) catch {
    writeStderr("out of memory\n");
    return .handled;
};
vm.gc.pushRoot(&path_val);
defer vm.gc.popRoot();
var sym_val = vm.gc.allocSymbol("load") catch {
    writeStderr("out of memory\n");
    return .handled;
};
vm.gc.pushRoot(&sym_val);
defer vm.gc.popRoot();
```

`path_val` is now valid before it is rooted, and it is rooted before `allocSymbol` (the next allocation), so it's protected exactly when a collection can occur.

baijum (commented):
The fix is the right shape — building `(load "path")` as Values sidesteps the reader-escape mangling entirely, drops the 1024-byte cap, and the pty smoke test genuinely exercises all three failure modes (I confirmed the `,`-command bypass in `isCompleteCallback` lets the unterminated-quote case submit, so the QUOTE case is real coverage, not an accident). The `evalInputInner` → `evalExpr`/`printResult` extraction reads as behavior-preserving: the two former print blocks were identical, and `catch continue` → `catch return` + `return true` keeps the reader loop's continue-on-print-failure semantics. The intra-command rooting is also correct: every intervening allocation (`allocString`, both `allocPair`s, `evalInputValue`) is balanced, so the two sibling `defer popRoot()`s pop strictly LIFO (`sym_val` then `path_val`).

One latent GC-safety nit on the two `= undefined` roots — details inline. Not a blocker for the shipped ReleaseSafe binaries (0xAA-fill masks it), but it deviates from the file's own convention and bites under `-Doptimize=ReleaseFast`.

src/repl_commands.zig
baijum: Good catch — fixed in e997794e. The `,load` handler now assigns before rooting, matching the `,expand`/`,import` convention in this file (`path_val` is valid before `pushRoot`, and rooted before `allocSymbol`; `sym_val` before the `allocPair`s). Verified: the smoke test passes against a `-Doptimize=ReleaseFast` build, which is exactly the mode where the undefined slot was dangerous.

baijum:
LGTM. The follow-up commit e997794e resolves the GC-safety nit exactly — assign-before-root now matches the `,expand`/`,import` handlers, so no undefined slot is ever live on the root stack when a collection can occur, and the assign→root ordering leaves no allocation gap. Combined with the Values-based `,load` (no reader-escape round-trip), the behavior-preserving `evalExpr`/`printResult` extraction, and the pty smoke test that genuinely exercises the quote/backslash/TAB cases, this is good to merge. 🚀

Issue #2252, opened 2026-08-08
FORMAL_FLAG comments claim case-lambda formals are kept bare, but only lambda is handled


baijum:
Introduced in #2251 (macro hygiene, #2003/#2074).

## What the comments claim

Three comments around `FORMAL_FLAG` in `src/expander_instantiate.zig` say the bare-spelling anaphoric exception applies to a **lambda *or* case-lambda** formal:

- the field declaration: `const FORMAL_FLAG: u32 = 0x04000000; // identifier is a lambda/case-lambda formal`
- the block comment above the lambda handling (~line 404): "Marked with FORMAL_FLAG so renameForHygiene treats a formal colliding with a global procedure as an anaphoric binding …"
- the `isProcedure` branch (~line 959): "The one exception: a **LAMBDA/CASE-LAMBDA FORMAL** (FORMAL_FLAG) colliding with a global procedure keeps its bare spelling …"

## What the code does

Only `lambda` is handled. `instantiateTemplate` sets `FORMAL_FLAG` on the formals list solely under:

```zig
if (std.mem.eql(u8, form_name, "lambda")) {
    ...
    intro_scope | FORMAL_FLAG   // formals walked with the flag
```

A `case-lambda` head (`form_name == "case-lambda"`) never matches, so each clause's formals go through the ordinary pair recursion **without** `FORMAL_FLAG`. A case-lambda formal that collides with a global procedure is therefore hygiene-renamed, not kept bare — the opposite of what the comments describe.

## Reproduction (branch binary)

```scheme
(define-syntax mk
  (syntax-rules () ((_) (case-lambda ((car) (car)) ((car x) (+ car x))))))
(define f (mk))
```

`kaappi expand` shows the formal renamed:

```
(define f (__hyg_1_case-lambda ((__hyg_2_car) (__hyg_2_car))
                               ((__hyg_2_car __hyg_3_x) (__hyg_4_+ __hyg_2_car __hyg_3_x))))
```

whereas the identical formal under a plain `lambda` stays bare (`car`).

## Impact

[...]

PR #2275, opened 2026-08-09
Remove syntax-rules rule/literal caps; fix define-property in templates (Fixes #2184, #2089)

### LINKED ISSUE #2089: [Expander] define-property is missing from well_known_forms: any syntax-rules template using it expands to an undefined variable
**Category:** R7RS macro hygiene / expander
**Severity:** wrong-result — a `syntax-rules` template that uses `define-property` cannot work at all
**Found by:** Systematic audit v2, Phase 4A (#1890), while deriving `isRejectedFormHead` (#1896)

`define-property` (SRFI 213) is missing from `expander.well_known_forms`
(`src/expander.zig:47`) — the hand-maintained list of identifiers the hygiene
renamer must never rename. Every other `sexpr_form_map` keyword that has no
global binding is in that list; `define-property` is the only one that is not.

Consequence: a template-introduced `define-property` is renamed to
`__hyg_N_define-property`, which `ir.sexpr_form_map` does not recognize, so the
form compiles as an ordinary **procedure call** to an undefined name.

## Minimal reproduction

```scheme
(define color 'red)
(define hue 'hue)
(define-syntax defprop
  (syntax-rules () ((_ id key val) (define-property id key val))))
(defprop color hue 'BRIGHT)
```

```
error[KP3001]: undefined variable '__hyg_2_define-property'
    (defprop color hue 'BRIGHT)
```

`kaappi expand` names the mechanism outright — the head, and only the head, was
renamed:

```
(__hyg_1_define-property color hue (quote BRIGHT))
```

## Discriminating control

The identical template shape around a keyword that **is** in
`well_known_forms` expands and runs correctly:

```scheme
(define-syntax defvals
  (syntax-rules () ((_ formals expr) (define-values formals expr))))
(defvals (a b) (values 1 2))
(display (list a b))   ;; => (1 2)
```

And `define-property` used **directly**, with no macro, works:

```scheme
(define color 'red)
(define-property color hue 'BRIGHT)   ;; fine
```

So the difference is exactly membership in that list, not `define-property`'s
own compilation.

## Not a duplicate of #2074

[#2074](https://github.com/kaappi/kaappi/issues/2074) is the **inverse**
mechanism: keywords that *are* in `well_known_forms` are left un-renamed, so a
use-site **local** of the same name shadows them. This one needs no shadowing
and no use-site binding at all — the keyword is renamed when it should not be,
and then resolves to nothing. Different direction, different fix (one name
added to a list, vs. changing how free keywords resolve). A fix for #2074 that
made keyword resolution hygiene-aware in general could subsume this; a fix that
only widens the un-renamed set would still need this entry.

## Sweep of the other absentees

Every other name in `ir.other_special_forms` / `ir.sexpr_form_map` that is
absent from `well_known_forms` was probed in the same template position and is
**fine**, so this is one name, not a class:

| absent name | via template | why it survives |
|---|---|---|
| `call/cc`, `call-with-current-continuation`, `call-with-values`, `apply`, `eval` | works | also bound globals, so the renamed head still resolves |
| `if`, `let` | works | documented deliberate omissions (the R7RS suite rebinds them as variables) |
| **`define-property`** | **raises** | special form with no global binding — nothing to fall back to |

## Suggested fix

Add `"define-property"` to `well_known_forms`. The durable version is the same
treatment #1896 just got for the native backend's own parallel list: derive the
special-form half of `well_known_forms` from `ir.sexpr_form_map` +
`ir.other_special_forms` with the two documented omissions (`if`, `let`) as an
explicit exclusion list, so a new `FormKind` cannot drift out of it. That is a
larger change than this bug needs and is filed here as the option, not the
requirement.

**Verified:** fresh ReleaseSafe build at `1718a2e1` + the #1896 derivation,
cache cleared; `kaappi expand` output quoted above.
**Source location:** `src/expander.zig:47` (`well_known_forms`), consumed by
`isWellKnown`.

### LINKED ISSUE #2184: syntax-rules silently caps at 32 rules (and 32 literals) with a bare KP2001 and no hint
Found while extending `(srfi 42)`'s `%do-ec` for PR #2180's review: adding four rules to a 31-rule `syntax-rules` made the whole `define-library` fail with

```
error[KP2001]: InvalidSyntax while compiling library definition '%do-ec'
```

and nothing else. The macro was structurally valid R7RS; the only problem was its size.

## Cause

`parseSyntaxRules` in `src/compiler_define_syntax.zig` uses fixed buffers:

```zig
var literals_buf: [32]Value = undefined;
...
if (lit_count >= 32) return CompileError.InvalidSyntax;
...
var patterns_buf: [32]Value = undefined;
var templates_buf: [32]Value = undefined;
...
if (rule_count >= 32) return CompileError.InvalidSyntax;
```

So the 33rd rule (or 33rd literal) turns the whole definition into `InvalidSyntax` — indistinguishable from a real syntax error, with no mention of a limit and no pointer to which rule tripped it.

## Why it matters

R7RS places no bound on rule count, and 33+ rules is not exotic — a qualifier- or opcode-style dispatcher macro gets there naturally (`(srfi 42)`'s qualifier processor reached 35 before being split into two chained macros as a workaround). The failure is loud but **wrongly attributed**: it reads as "your macro is malformed," which sends the author paren-hunting through a correct definition. Diagnosis took bisecting the definition down to a rule-count experiment.

## Suggested fix (either suffices, both are better)

1. Make the buffers growable (heap-allocate past 32), removing the limit.
2. If the cap stays, report it as its own diagnostic: `syntax-rules: too many rules (limit 32)` / `too many literals (limit 32)` with the form's location — a `KP2xxx` with an `explain` entry, so the failure names the actual constraint.

A regression test would be a `syntax-rules` with 33 trivial rules: today it must fail with the specific new diagnostic (or, with fix 1, compile and expand correctly).

Priority `medium` per the rubric: behaviour is wrong against the spec (a legal program is rejected) but loud, narrow, and recoverable by splitting the macro.

baijum:
Closes #2184, closes #2089.

## #2184 — `syntax-rules` silently capped at 32 rules (and 32 literals)

`parseSyntaxRules` (`src/compiler_define_syntax.zig`) parsed into fixed `[32]Value` stack buffers and rejected the 33rd rule or literal with a bare `KP2001 InvalidSyntax` — a legal, well-formed macro failed as if it were malformed, with no mention of the cap and no pointer to the tripping rule. R7RS places no bound on either count, and a 33+ rule dispatcher macro is a normal shape (`(srfi 42)`'s qualifier processor reached 35 and was split into two chained macros as a workaround).

**Fix:** the buffers are now growable `std.ArrayList`s (raw allocator, never GC-triggering; every held `Value` is a subpart of the rooted spec). The limit is gone. A `u16` guard replaces the old cap where `Transformer.num_rules`'s width would otherwise make `allocTransformer`'s `@intCast` trap on a 65k+-rule macro (previously unreachable).

**Bonus — the workaround is gone:** `lib/srfi/42.sld`'s `%do-ec`/`%do-ec-more` split is merged back into a single 34-rule `%do-ec`, exactly the shape that exposed the bug. All 119 srfi-42 assertions pass.

## #2089 — `define-property` unusable from a `syntax-rules` template

Two independent gaps, both fixed:

1. The keyword was missing from the expander's `reserved_template_forms`, so a template-introduced `define-property` was hygiene-renamed to `__hyg_N_define-property`. (`well_known_forms` alone — the issue's suggested fix — does not stop the rename: `instantiateTemplate` consults `isTemplateReserved`.)
2. Even left bare, the legacy `compileForm` path (which compiles macro expansions, via `compileExpr`) had no `define-property` dispatch — only the IR path did — so the expanded form compiled as a call to an undefined name.

`define-property` now sits in both `reserved_template_forms` and `well_known_forms` (mirroring `define-values`), and `compileForm` dispatches it like the IR path.

## Verification

- New regression tests (both fail on `main` with the exact issue symptoms, pass here):
  - `tests/scheme/hygiene/syntax-rules-unbounded-2184.scm` — 40-rule, 40-literal, and 33-rule-in-library macros
  - `tests/scheme/srfi/srfi213.scm` — template-emitted `define-property` at top level and inside a `define-library`
- `bash tests/scheme/run-all.sh`: **702 scheme files + 1395 R7RS tests, 0 failures**
- `zig build test` and `zig build test -Dgc-stress=true`: pass

Note: #2252 (FORMAL_FLAG comments) appears already resolved — the merged #2251 commit's comments now say "lambda formal… deliberately NOT extended to… case-lambda formals (#2252)", matching the code.

### DIFF
--- a/src/compiler_define_syntax.zig
+++ b/src/compiler_define_syntax.zig
@@ -719,29 +719,35 @@ pub fn parseSyntaxRules(self: *Compiler, spec: Value, extra_bound: []const []con
     const literals_list = types.car(after_ellipsis);
     const rules = types.cdr(after_ellipsis);
 
-    var literals_buf: [32]Value = undefined;
-    var lit_count: usize = 0;
+    // R7RS places no bound on a syntax-rules' literal or rule count, so
+    // these grow past the old fixed 32-slot stack buffers (kaappi#2184): a
+    // 33-rule dispatcher macro is legal and must compile, not fail with a
+    // bare KP2001 that reads as "malformed macro". The ArrayList backing
+    // stores use the raw allocator (never GC-triggering), and every Value
+    // they hold is a subpart of `spec`, which resolveTransformerSpecRec
+    // roots for the duration of this call — safe against collection at
+    // any point.
+    var literals: std.ArrayList(Value) = .empty;
+    defer literals.deinit(self.gc.allocator);
     var lit = literals_list;
     while (lit != types.NIL) {
         if (!types.isPair(lit)) return CompileError.InvalidSyntax;
-        if (lit_count >= 32) return CompileError.InvalidSyntax;
         // A generating macro may splice a user identifier into this spec's
         // literal list (SRFI 257's if-new-var) — unwrap the provenance
         // marker so the literal is the bare identifier.
-        literals_buf[lit_count] = expander.unwrapUsertext(types.car(lit));
-        lit_count += 1;
+        literals.append(self.gc.allocator, expander.unwrapUsertext(types.car(lit))) catch return CompileError.OutOfMemory;
         lit = types.cdr(lit);
     }
 
-    var patterns_buf: [32]Value = undefined;
-    var templates_buf: [32]Value = undefined;
-    var rule_count: usize = 0;
+    var patterns: std.ArrayList(Value) = .empty;
+    defer patterns.deinit(self.gc.allocator);
+    var templates: std.ArrayList(Value) = .empty;
+    defer templates.deinit(self.gc.allocator);
     var rule = rules;
     while (rule != types.NIL) {
         if (!types.isPair(rule)) return CompileError.InvalidSyntax;
         const r = types.car(rule);
         if (!types.isPair(r)) return CompileError.InvalidSyntax;
-        if (rule_count >= 32) return CompileError.InvalidSyntax;
         const rule_pattern = types.car(r);
         // R7RS 4.3.2 (kaappi#2082): the <pattern> grammar admits at
         // most one <ellipsis> per list or vector pattern. A second one
@@ -749,22 +755,28 @@ pub fn parseSyntaxRules(self: *Compiler, spec: Value, extra_bound: []const []con
         // what may FOLLOW the single ellipsis); accepting it previously
         // split the input arbitrarily because the surplus ellipsis
         // tokens were counted as fixed tail elements by the matcher.
-        if (!validPatternGrammar(rule_pattern, custom_ellipsis orelse "...", literals_buf[0..lit_count]))
+        if (!validPatternGrammar(rule_pattern, custom_ellipsis orelse "...", literals.items))
             return CompileError.InvalidSyntax;
-        patterns_buf[rule_count] = rule_pattern;
+        patterns.append(self.gc.allocator, rule_pattern) catch return CompileError.OutOfMemory;
         const r_rest = types.cdr(r);
         if (r_rest == types.NIL) return CompileError.InvalidSyntax;
-        templates_buf[rule_count] = types.car(r_rest);
-        rule_count += 1;
+        templates.append(self.gc.allocator, types.car(r_rest)) catch return CompileError.OutOfMemory;
         rule = types.cdr(rule);
     }
 
-    if (rule_count == 0) return CompileError.InvalidSyntax;
+    if (patterns.items.len == 0) return CompileError.InvalidSyntax;
+
+    // Transformer.num_rules is a u16; the old 32-slot cap made an overflow
+    // unreachable, so guard the new unbounded path explicitly rather than
+    // let allocTransformer's @intCast trap (ReleaseSafe panic) on a
+    // 65k+-rule macro. No real program reaches this; it exists only to
+    // keep the now-unbounded rule count from becoming a crash.
+    if (patterns.items.len > std.math.maxInt(u16)) return CompileError.InvalidSyntax;
 
     const tx_val = self.gc.allocTransformer(
-        literals_buf[0..lit_count],
-        patterns_buf[0..rule_count],
-        templates_buf[0..rule_count],
+        literals.items,
+        patterns.items,
+        templates.items,
     ) catch return CompileError.OutOfMemory;
     const tx = types.toObject(tx_val).as(types.Transformer);
     if (custom_ellipsis) |ce| {
@@ -773,9 +785,9 @@ pub fn parseSyntaxRules(self: *Compiler, spec: Value, extra_bound: []const []con
     // R7RS 4.3.2: record each literal's def-site binding slot (0xFFFF = unbound).
     // Binding identity — not just bound/unbound — is needed so that two
     // different bindings with the same name don't falsely match.
-    if (lit_count > 0) {
-        const slots = self.gc.allocator.alloc(u32, lit_count) catch return CompileError.OutOfMemory;
-        for (literals_buf[0..lit_count], 0..) |lv, li| {
+    if (literals.items.len > 0) {
+        const slots = self.gc.allocator.alloc(u32, literals.items.len) catch return CompileError.OutOfMemory;
+        for (literals.items, 0..) |lv, li| {
             slots[li] = if (types.isSymbol(lv)) blk: {
                 const lname = types.symbolName(lv);
                 // Resolve through the full lexical chain with the SAME rules
--- a/src/compiler_define_syntax.zig
+++ b/src/compiler_define_syntax.zig
@@ -722,11 +722,16 @@ pub fn parseSyntaxRules(self: *Compiler, spec: Value, extra_bound: []const []con
     // R7RS places no bound on a syntax-rules' literal or rule count, so
     // these grow past the old fixed 32-slot stack buffers (kaappi#2184): a
     // 33-rule dispatcher macro is legal and must compile, not fail with a
-    // bare KP2001 that reads as "malformed macro". The ArrayList backing
-    // stores use the raw allocator (never GC-triggering), and every Value
-    // they hold is a subpart of `spec`, which resolveTransformerSpecRec
-    // roots for the duration of this call — safe against collection at
-    // any point.
+    // bare KP2001 that reads as "malformed macro". GC safety: the backing
+    // stores use the raw allocator (never GC-triggering), nothing between
+    // here and allocTransformer triggers a collection (unwrapUsertext /
+    // validPatternGrammar / append allocate nothing on the GC heap), and
+    // allocTransformer dupes all three slices with the raw allocator
+    // before it can collect — so the element Values only need to survive
+    // as subparts of the form being parsed. resolveTransformerSpecRec
+    // additionally roots `spec`, but the body-scan caller in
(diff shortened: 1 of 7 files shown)


src/compiler_define_syntax.zig
```diff
@@ -719,52 +719,64 @@ pub fn parseSyntaxRules(self: *Compiler, spec: Value, extra_bound: []const []con
...
-    var literals_buf: [32]Value = undefined;
-    var lit_count: usize = 0;
+    // R7RS places no bound on a syntax-rules' literal or rule count, so
+    // these grow past the old fixed 32-slot stack buffers (kaappi#2184): a
+    // 33-rule dispatcher macro is legal and must compile, not fail with a
+    // bare KP2001 that reads as "malformed macro". The ArrayList backing
+    // stores use the raw allocator (never GC-triggering), and every Value
+    // they hold is a subpart of `spec`, which resolveTransformerSpecRec
+    // roots for the duration of this call — safe against collection at
+    // any point.
```
baijum: Nit (doc accuracy): this comment credits `resolveTransformerSpecRec` for rooting `spec`, which is exactly right for the primary caller (it does `pushRoot(&spec)` at line 527 and calls into here at 559). But `parseSyntaxRules` has a second caller that bypasses `resolveTransformerSpecRec` entirely — the `define-syntax` branch of the lambda/let body scan (`compiler_lambda.zig:581`), which calls `parseSyntaxRules(compiler, transformer_spec, …)` directly without pushing a root for `transformer_spec`.

Safety still holds on that path, but for a different reason than the comment states: nothing in these two loops GC-allocates, and `allocTransformer` `dupe`s the slices before it can collect, with the elements surviving as subparts of the body form the compiler is actively processing. Since GC-safety comments here are load-bearing, consider broadening it so a future editor who adds a GC-triggering call to the parse loop isn't misled into thinking `spec` is independently rooted on *every* path — e.g. "…every Value they hold is a subpart of the spec being parsed; nothing between here and `allocTransformer` triggers a collection, and `allocTransformer` copies the slices before it can. On the primary path the spec is additionally rooted by `resolveTransformerSpecRec`."

src/compiler_define_syntax.zig
```diff
@@ -719,52 +719,64 @@ pub fn parseSyntaxRules(self: *Compiler, spec: Value, extra_bound: []const []con
...
 
-    if (rule_count == 0) return CompileError.InvalidSyntax;
+    if (patterns.items.len == 0) return CompileError.InvalidSyntax;
+
+    // Transformer.num_rules is a u16; the old 32-slot cap made an overflow
+    // unreachable, so guard the new unbounded path explicitly rather than
+    // let allocTransformer's @intCast trap (ReleaseSafe panic) on a
+    // 65k+-rule macro. No real program reaches this; it exists only to
+    // keep the now-unbounded rule count from becoming a crash.
+    if (patterns.items.len > std.math.maxInt(u16)) return CompileError.InvalidSyntax;
```
baijum, line 779: Optional: this guard returns a bare `CompileError.InvalidSyntax` (KP2001) — the very "reads as malformed" diagnostic this PR is removing — for a 65536+-rule macro. It's genuinely unreachable (that's ~hundreds of KB of macro source) and the comment says as much, so this is fine to leave. Just flagging that if you wanted full consistency with the issue's spirit, a dedicated "too many rules" message (`argError`/a KP with the count) would read better than the generic one. Not worth blocking on.

baijum (commented):
Reviewed both fixes against the code. This is a clean, well-reasoned PR — the reasoning in the comments and description is unusually thorough and matches what the code does. I verified the GC-safety story, the u16 boundary, the dispatch placement, and the srfi-42 merge. No blocking issues; two small notes below.

**#2184 — unbounded rules/literals**
- The GC argument holds: `append` uses the raw allocator (no collection), `unwrapUsertext`/`validPatternGrammar` don't GC-allocate, and `allocTransformer` `dupe`s all three slices before `finishAlloc` can collect — with the element Values kept alive as subparts of the rooted `spec`. Confirmed `resolveTransformerSpecRec` does `pushRoot(&spec)` around the `parseSyntaxRules(self, spec, ...)` call (compiler_define_syntax.zig:527/559).
- The `u16` guard boundary is correct: `> maxInt(u16)` rejects, so `len == 65535` still passes and `allocTransformer`'s `@intCast` is safe. `num_rules` bounds only `patterns`/`templates` (appended in lockstep, so equal length); `literals` is a separate slice with no width cap — all consistent.
- srfi-42 merge is order-preserving (generator rules first, then the former `%do-ec-more` qualifier rules, base case last) and the added literals (`nested begin :while :until … if not and or`) are exactly what the inlined rules match literally. 119 assertions passing backs this up.

**#2089 — define-property in templates**
- Correct on both axes: added to `reserved_template_forms` (stops the `__hyg_N_` rename via `isTemplateReserved`) and to `well_known_forms`, mirroring `define-values`. Both list lookups are linear scans, so the reformatting/ordering is immaterial.
- The new legacy-path dispatch sits inside the `!is_shadowed` block, and `compileDefineProperty` still guards `self.in_body_scope` first, so the body-scope rejection is preserved regardless of which path reaches it.

Tests, CHANGELOG, and the two regression files all look right.

src/compiler_define_syntax.zig
baijum: Fixed in 9bee75dc. Confirmed the second caller: compiler_lambda.zig:581 (the define-syntax branch of the lambda/let body scan) calls parseSyntaxRules directly with no root for transformer_spec. The comment now states the guarantee in terms of what this code itself ensures (nothing in the parse loops GC-allocates; allocTransformer dupes all three slices with the raw allocator before it can collect) rather than crediting resolveTransformerSpecRec, and carries an explicit warning not to add a GC-triggering call to the loops without rooting the spec.

src/compiler_define_syntax.zig
baijum: Agreed it is unreachable in practice, and agreed that a bare KP2001 is not ideal for it. I looked at giving it a distinct diagnostic: CompileError.InvalidSyntax maps straight to diagnostics.invalid_syntax (KP2001) with no per-error message mechanism, so a dedicated "too many rules" message would mean a new CompileError variant plus a diagnostics code/stage/explain entry for an edge no real program can hit (~65k rules). That felt disproportionate, so I left the guard as-is; the comment already documents why it exists. Happy to revisit if you would rather have the diagnostic.

baijum:
LGTM ✅

The follow-up commit (9bee75dc) resolves my one substantive note — the GC-safety comment now describes the guarantee the function itself enforces (raw-allocator backing, no GC-triggering call before `allocTransformer` dupes the slices) and explicitly warns that the `compiler_lambda.zig` body-scan caller doesn't root `spec`. Clean fix, comment-only.

Both changes verified:
- **#2184** — unbounded rules/literals is GC-safe; the `u16` guard boundary is correct (65535 still passes `@intCast`); `num_rules` bounds only patterns/templates (appended in lockstep), literals is a separate uncapped slice; the srfi-42 merge back into a single 34-rule `%do-ec` is order-preserving and test-covered.
- **#2089** — `define-property` added to both `reserved_template_forms` and `well_known_forms` (mirroring `define-values`); legacy-path dispatch sits inside `!is_shadowed` and still honors the `in_body_scope` rejection.

Regression tests, CHANGELOG, and the srfi-42 workaround removal all look right. Nice work.

PR #2276, opened 2026-08-09
Fix all eight SRFI-146 audit findings (2045, 2046, 2047, 2048, 2049, 2050, 2052, 2053)

baijum:
Closes the SRFI-146 backlog from Systematic audit v2, Phase 3.4: **#2045 #2046 #2047 #2048 #2049 #2050 #2052 #2053** — the complete set of open SRFI-146 issues. Every fix aligns the implementation with the SRFI's reference implementation (`srfi-146.tgz`), and each commit re-enables the assertions the audit had disabled behind a `;; FAIL: #NNNN` marker.

## Changes (one commit per issue)

| Issue | Severity | Fix |
|---|---|---|
| #2045 | medium | `mapping`/`mapping-unfold`/`hashmap`/`hashmap-unfold` (and `/ordered` aliases) now keep the **first** of equal keys, per the Constructors spec, using the same first-wins semantics as `mapping-adjoin`/`alist->mapping` |
| #2046 | medium | `mapping-key-predecessor`/`-successor` tail-call `failure` only when no key exists, instead of evaluating it as the fold seed on every call |
| #2047 | medium | `mapping=?`/`hashmap=?` return `#f` for mappings with **different key comparators** (`eq?` identity check, as the reference does) |
| #2048 | medium | `make-mapping-comparator` now supplies the spec's **lexicographic ordering** (`mapping-ordering`); `make-hashmap-comparator` supplies a hash function (the constant the reference ships, whose TODO for a real hash is inherited) — mappings can now key mappings, the spec's stated purpose |
| #2049 | medium | `hashmap-ref/default` returns a **procedural default unchanged** via `hash-table-ref/default`, instead of invoking it as a failure thunk |
| #2050 | low | `mapping-any?`/`mapping-every?` return literal `#t`/`#f`, not the predicate's value, matching the hashmap twins |
| #2052 | low | all **ten** submapping comparison predicates accept the single-mapping form (`(mapping=? c m)` → `#t`) the signature allows |
| #2053 | low | deleted the discarded first fold in `mapping-map` (side-effecting `proc` was applied **twice** per association) and `mapping-find` |

## Tests

- Re-enabled every disabled assertion for these issues in `srfi146-differential.scm` and `srfi146-reference.scm`; dropped the pins that recorded the wrong answers. Includes the reference suite's `hashmap-keyed hashmap`, which was pinned under #2044 (closed separately) but actually needs this PR's hash function.
- Left the unrelated `#2023` (deep-keyed hash cliff) markers disabled.
- Full scheme suite: **703 files pass, 0 fail**; R7RS suite **1395 pass, 0 fail**. `srfi146.scm` 82/82, reference 174 passes, differential 526 passes.
- Also verified the SRFI-146 dependents (srfi 167, 168) still pass.

### DIFF
--- a/lib/srfi/146.sld
+++ b/lib/srfi/146.sld
@@ -428,26 +428,31 @@
 
     (define mapping=?
       (case-lambda
+        ((vcmp m1) #t)               ; one mapping: vacuously true (#2052)
         ((vcmp m1 m2) (%mapping=? vcmp m1 m2))
         ((vcmp m1 m2 . rest) (and (%mapping=? vcmp m1 m2) (apply mapping=? vcmp m2 rest)))))
 
     (define mapping<?
       (case-lambda
+        ((vcmp m1) #t)
         ((vcmp m1 m2) (and (%mapping<=? vcmp m1 m2) (not (%mapping=? vcmp m1 m2))))
         ((vcmp m1 m2 . rest) (and (mapping<? vcmp m1 m2) (apply mapping<? vcmp m2 rest)))))
 
     (define mapping>?
       (case-lambda
+        ((vcmp m1) #t)
         ((vcmp m1 m2) (mapping<? vcmp m2 m1))
         ((vcmp m1 m2 . rest) (and (mapping>? vcmp m1 m2) (apply mapping>? vcmp m2 rest)))))
 
     (define mapping<=?
       (case-lambda
+        ((vcmp m1) #t)
         ((vcmp m1 m2) (%mapping<=? vcmp m1 m2))
         ((vcmp m1 m2 . rest) (and (%mapping<=? vcmp m1 m2) (apply mapping<=? vcmp m2 rest)))))
 
     (define mapping>=?
       (case-lambda
+        ((vcmp m1) #t)
         ((vcmp m1 m2) (%mapping<=? vcmp m2 m1))
         ((vcmp m1 m2 . rest) (and (mapping>=? vcmp m1 m2) (apply mapping>=? vcmp m2 rest)))))
 
--- a/lib/srfi/146.sld
+++ b/lib/srfi/146.sld
@@ -430,7 +430,16 @@
     (define mapping<?
       (case-lambda
         ((vcmp m1) #t)
-        ((vcmp m1 m2) (and (%mapping<=? vcmp m1 m2) (not (%mapping=? vcmp m1 m2))))
+        ((vcmp m1 m2)
+         ;; The strict predicates must not inherit #2047's identity guard
+         ;; through =? only: %mapping<=? is structural, so with different
+         ;; (but structurally identical) comparator objects both m1<m2 and
+         ;; m2<m1 would hold.  Guard the strict comparison on identity too,
+         ;; keeping the mixed-comparator path antisymmetric like the
+         ;; reference (whose <? is defined without consulting =?).
+         (and (eq? (%mapping-comparator m1) (%mapping-comparator m2))
+              (%mapping<=? vcmp m1 m2)
+              (not (%mapping=? vcmp m1 m2))))
         ((vcmp m1 m2 . rest) (and (mapping<? vcmp m1 m2) (apply mapping<? vcmp m2 rest)))))
 
     (define mapping>?
--- a/lib/srfi/146.sld
+++ b/lib/srfi/146.sld
@@ -594,8 +594,30 @@
 
     ;;; Comparator
 
+    ;; The spec's ordering for mapping-comparator: "the lexicographic ordering
+    ;; with respect to the keys (and, in case a tiebreak is necessary, with
+    ;; respect to the ordering of the values)", for pairs of mappings sharing
+    ;; the same key comparator.  The reference implements it by walking two
+    ;; tree generators in parallel; the sorted alists are the same walk (#2048).
+    (define (mapping-ordering vcmp)
+      (let ((veq (comparator-equality-predicate vcmp))
+            (vlt (comparator-ordering-predicate vcmp)))
+        (lambda (m1 m2)
+          (let ((eq (%cmp= m1))
+                (lt (%cmp< m1)))
+            (let loop ((a1 (mapping->alist m1)) (a2 (mapping->alist m2)))
+              (cond
+                ((null? a1) (not (null? a2)))
+                ((null? a2) #f)
+                ((eq (caar a1) (caar a2))
+                 (if (veq (cdar a1) (cdar a2))
+                     (loop (cdr a1) (cdr a2))
+                     (vlt (cdar a1) (cdar a2))))
+                (else (lt (caar a1) (caar a2)))))))))
+
     (define (make-mapping-comparator vcmp)
-      (make-comparator mapping? (lambda (m1 m2) (%mapping=? vcmp m1 m2)) #f #f))
+      (make-comparator mapping? (lambda (m1 m2) (%mapping=? vcmp m1 m2))
+                       (mapping-ordering vcmp) #f))
 
     (define mapping-comparator (make-mapping-comparator (make-default-comparator)))
 
--- a/lib/srfi/146.sld
+++ b/lib/srfi/146.sld
@@ -397,14 +397,18 @@
     ;;; Comparisons
 
     (define (%mapping=? vcmp m1 m2)
-      (let ((veq (comparator-equality-predicate vcmp))
-            (a1 (mapping->alist m1)) (a2 (mapping->alist m2)))
-        (and (= (length a1) (length a2))
-             (let loop ((a a1) (b a2))
-               (or (null? a)
-                   (and ((%cmp= m1) (caar a) (caar b))
-                        (veq (cdar a) (cdar b))
-                        (loop (cdr a) (cdr b))))))))
+      ;; The spec: it is "explicitly not an error" to compare mappings with
+      ;; different key comparators -- in that case #f is returned.  "Share the
+      ;; same comparator" means object identity, so this is eq? (#2047).
+      (and (eq? (%mapping-comparator m1) (%mapping-comparator m2))
+           (let ((veq (comparator-equality-predicate vcmp))
+                 (a1 (mapping->alist m1)) (a2 (mapping->alist m2)))
+             (and (= (length a1) (length a2))
+                  (let loop ((a a1) (b a2))
+                    (or (null? a)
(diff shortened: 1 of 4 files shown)


lib/srfi/146.sld
```diff
@@ -412,26 +423,31 @@
...
     (define mapping=?
       (case-lambda
+        ((vcmp m1) #t)               ; one mapping: vacuously true (#2052)
         ((vcmp m1 m2) (%mapping=? vcmp m1 m2))
         ((vcmp m1 m2 . rest) (and (%mapping=? vcmp m1 m2) (apply mapping=? vcmp m2 rest)))))
 
     (define mapping<?
       (case-lambda
+        ((vcmp m1) #t)
         ((vcmp m1 m2) (and (%mapping<=? vcmp m1 m2) (not (%mapping=? vcmp m1 m2))))
```
baijum: The #2047 comparator-identity guard on `%mapping=?` now leaks into `mapping<?`/`mapping>?`, which are defined here as `(and <=? (not =?))`. Concretely, two **same-content mappings built with different (but structurally identical) comparator objects** now compare as:

```scheme
(define c (make-default-comparator))
(define d (make-default-comparator))
(define m1 (mapping c 1 'a))
(define m2 (mapping d 1 'a))
(mapping<? c m1 m2)  ; => #t   (was #f before this PR)
(mapping<? c m2 m1)  ; => #t   <-- non-antisymmetric!
```

Both directions return `#t` because `%mapping<=?` ignores comparator identity while `%mapping=?` now returns `#f`. The reference sidesteps this: it defines `%mapping<?` as `(not (%mapping<=? m2 m1))` (structural only), so it returns `#f` for both here. So this PR changes `<?`/`>?` behaviour on the mixed-comparator path from `#f` to a non-antisymmetric `#t`.

It's an edge case (mixing comparators, which the spec only explicitly blesses for `=?`), and none of the suites exercise it — so I'm not treating it as a blocker. But since #2047 is what introduced it, it's worth a deliberate decision: either guard `<?`/`>?` on comparator identity too, or define them structurally like the reference. `(srfi 146 hash)` has the identical shape at `%hm=?`/`hashmap<?`.

lib/srfi/146/hash.sld
```diff
@@ -280,22 +296,27 @@
 
     (define hashmap=?
       (case-lambda
+        ((vcmp m1) #t)               ; one hashmap: vacuously true (#2052)
         ((vcmp m1 m2) (%hm=? vcmp m1 m2))
         ((vcmp m1 m2 . rest) (and (%hm=? vcmp m1 m2) (apply hashmap=? vcmp m2 rest)))))
     (define hashmap<?
       (case-lambda
+        ((vcmp m1) #t)
         ((vcmp m1 m2) (and (%hm<=? vcmp m1 m2) (not (%hm=? vcmp m1 m2))))
```
baijum: Same non-antisymmetry as the mapping side (see the note on `lib/srfi/146.sld` line 433): `(hashmap<? c h1 h2)` and `(hashmap<? c h2 h1)` both return `#t` for same-content hashmaps with different comparator objects, since `%hm<=?` ignores comparator identity but `%hm=?` (post-#2047) does not. Whatever you decide for the mapping side, mirror it here.

lib/srfi/146.sld
```diff
@@ -412,26 +423,31 @@
 
     (define mapping=?
       (case-lambda
+        ((vcmp m1) #t)               ; one mapping: vacuously true (#2052)
```
baijum, line 426: Nit (optional): the reference's 1-arg clauses do `(assume (mapping? mapping))`, so `(mapping=? c 42)` errors there but returns `#t` here. It's consistent with the rest of this file (the 2-arg clauses don't `assume` either), so purely a note — no change needed unless you want the parity.

baijum (commented):
Reviewed all eight fixes against the SRFI-146 reference implementation (`srfi/146.scm`, `srfi/146/hash.scm` from `srfi-146.tgz`). Every fix matches the reference's semantics, and I independently re-ran the three suites in the worktree: `srfi146-reference` **174 passes**, `srfi146-differential` **526 passes**, `srfi146.scm` **82/82** — all with the re-enabled assertions. The one-commit-per-issue structure with re-enabled `;; FAIL:` assertions is exactly the regression coverage CLAUDE.md asks for. Nice work.

One behavioral wrinkle worth a decision (non-blocking), plus a tiny nit — both inline. Summary of the verification per issue:

- **#2045** `%rbt-adjoin` / `hash-table-exists?`-guarded set → first-wins, matching the reference's `mapping-adjoin` / `hashmap-adjoin`. ✓
- **#2046** predecessor/successor now carry a `(cons found? key)` seed and only call `failure` when unfound — matches `tree-key-predecessor`. ✓
- **#2047** `eq?` on the key comparator — matches `(eq? (mapping-key-comparator m1) ...)` in the reference `%mapping=?`. ✓ (see inline note on how this leaks into `<?`)
- **#2048** lexicographic `mapping-ordering` (sorted-alist walk == the reference's parallel generator walk) and the constant-`0` hash (the reference's shipped TODO value). ✓
- **#2049** `hash-table-ref/default` returns a procedural default unchanged. ✓
- **#2050** `#t`/`#f` booleans, matching the hashmap twins and the reference's `call/cc`+`not`. ✓
- **#2052** single-mapping form → `#t`, matching the reference's 1-arg `case-lambda` clause. ✓
- **#2053** dead first fold removed; `mapping-map` still `%rbt-insert` (last-wins), matching the reference's `mapping-set`. ✓

baijum:
@coderabbitai review

lib/srfi/146.sld
baijum: Agreed, and thanks for the precise diagnosis — the pre-PR `#f`/`#f` came from the buggy `=?`, so #2047 flipped this path to a non-antisymmetric `#t`/`#t`. Fixed in 30f8da0f by giving `mapping<?`/`mapping>?` (and the hash twins) their own `(eq? ...)` key-comparator guard, so the mixed-comparator path is `#f` both ways again — matching both pre-PR behaviour and the reference (whose `%mapping<?` is defined without consulting `=?`).

I kept `%mapping<=?` structural rather than guarding it too: the reference's `<=?` is also structural (it only returns `#t` here through `(?`, so no observable difference), and the spec only blesses `=?` with the identity check — guarding `<=?` would have changed same-content subset results across comparators from `#t` to `#f` beyond what this issue needs. The result is the family is consistent: `<?`/`>?` are `#f` across comparators while `<=?`/`>=?` stay subset-shaped, `<?` remains a proper subset of `<=?`, and same-comparator semantics are untouched.

Two differential regression tests (one per library) pin the antisymmetry and were verified to fail without the guard.

lib/srfi/146/hash.sld
baijum: Mirrored on the hash side in the same commit (30f8da0f): `hashmap<?`/`hashmap>?` carry their own `(eq? (%hm-comparator m1) (%hm-comparator m2))` guard, with a matching antisymmetry regression test verified to fail without it.

lib/srfi/146.sld
baijum: Fair — left as-is, deliberately. This file uses no `assume` anywhere and the two-argument clauses don't type-check either, so an `assume` in only the one-argument clauses would be the odd one out. The reference's `assume` calls are part of its house style; kaappi's SRFI-146 port checks types only where the spec makes it an error (e.g. the empty-mapping procedures), not in the comparison family.

tests/scheme/srfi/srfi146-differential.scm
```diff
@@ -541,40 +541,28 @@
...
+             ((op 'map) (lambda (k v) (set! n (+ n 1)) (values k v)) c
+                        ((op 'make) c 1 'a 2 'b 3 'c 4 'd 5 'e))
+             n)))
+
+  (test-assert "mapping=? accepts a single mapping"
+    (guard (e (#t #f)) (mapping=? c (mapping c 1 'a))))
+  (test-assert "mapping<=? accepts a single mapping"
+    (guard (e (#t #f)) (mapping<=? c (mapping c 1 'a))))
+  (test-assert "hashmap=? accepts a single hashmap"
+    (guard (e (#t #f)) (hashmap=? c (hashmap c 1 'a)))))
```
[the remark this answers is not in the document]
baijum: Two of the three asks were already satisfied by the existing tests; one was a real gap. Addressed in 30f8da0f:

1. **"agree only checks sibling agreement"** — true in the abstract, but each of these `agree()` checks demonstrably fails without its fix: the hashmap side is the correct oracle, so the disagreement IS the failure signal (that's the differential suite's stated design — "running one parameterised body over both and demanding the same answers"). The audit verified each one fails against the pre-fix build. Still, to make each defect independent of the sibling oracle, I added direct spec-value assertions on the defective side: `mapping-any?`/`mapping-every?` return exactly `#t` for a truthy predicate value (#2050), `hashmap-ref/default` returns a procedural default unchanged (#2049), and `mapping-map` invokes its proc exactly 5 times (#2053) — each fails even if both libraries regress together.

2. **Single-argument coverage** — valid, this was a genuine gap. The re-enabled #2052 assertions covered only `mapping=?`, `mapping<=?`, and `hashmap=?`; the fix touched all ten predicates. Added one-argument assertions for the remaining seven (`mapping<?`/`>?`/`>=?`, `hashmap<?`/`>?`/`<=?`/`>=?`), all in the same `guard` style.

baijum:
Confirmed 30f8da0f resolves the antisymmetry finding. Verified in the worktree:

```
mapping<? both dirs: (#f #f)     ; was (#t #t)
mapping>? both dirs: (#f #f)
hashmap<? both dirs: (#f #f)
control proper-subset:  #t       ; shared-comparator semantics unchanged
```

The own-`eq?`-guard approach on the strict predicates lands in the same place as the reference's `(not (<=? m2 m1))` for the mixed-comparator case, while keeping kaappi's proper-subset semantics on the shared-comparator path. Leaving `<=?`/`>=?` structural is the right call — that's exactly the reference's shape (structural `<=?` + `eq?`-guarded `=?`), so no new deviation there. `mapping>?`/`hashmap>?` inherit the guard through their `(<? m2 m1)` definition. 

The extra coverage is a nice touch: the direct spec-value pins for #2050/#2049/#2053 (independent of the differential oracle), all-ten single-arg clauses, and the antisymmetry regressions with a shared-comparator control. Re-ran locally: differential **540 passes**, reference **174 passes**. LGTM.

Issue #2086, opened 2026-08-01
[SRFI-113] set->bag! silently drops the set's contribution for an element the bag already holds


baijum:
**Category:** SRFI
**Spec reference:** SRFI 113 §"Set and bag conversion" — "`(set->bag! bag set)` The `set->bag!` procedure returns a bag containing the elements of both *bag* and *set*."
**Severity:** wrong-result

`set->bag!` only inserts a set element that the bag does not already contain. An element the bag already holds keeps its old count, so the set's contribution is silently dropped.

**Minimal reproduction:**

```scheme
(import (scheme base) (scheme write) (srfi 113) (srfi 128))
(define cmp (make-default-comparator))
(define b (bag cmp 1 1))                 ; multiplicity of 1 is 2
(set->bag! b (set cmp 1 2))
(display (bag-element-count b 1))        ; expected: 3   actual: 2
```

**Discriminating control:** `(bag-element-count (set->bag! (bag cmp 1 1) (set cmp 1 2)) 2)` is `1` — the element the bag did *not* already hold is added correctly, so only the already-present branch is wrong. `set->bag` (the non-destructive twin, which always starts from an empty bag) is correct.

**Both independent oracles agree with the spec, not with Kaappi.** The reference implementation is

```scheme
(define (set->bag! bag set)
  ...
  (hash-table-for-each (lambda (key value) (sob-increment! bag key value))
                       (sob-hash-table set))
  bag)
```

— an unconditional increment — and chibi-scheme 0.12.0 returns `3`.

**Source location:** `lib/srfi/113.sld`, `set->bag!`:

```scheme
(hash-table-walk (%set-ht s)
  (lambda (k v)
    (if (not (hash-table-exists? (%bag-ht b) k))
        (hash-table-set! (%bag-ht b) k 1))))
```

**Verified:** fresh ReleaseSafe build at `1718a2e1`, macOS aarch64, isolated `KAAPPI_HOME`.

**Disabled test:** `tests/scheme/srfi/srfi113-audit.scm` — 1 assertion marked `;; FAIL: #2086`.

**Found by:** Systematic audit v2, Phase 3.9.

baijum:
Fixed by #2277 ("Fix four SRFI audit findings: and-let*, SRFI 222, set->bag!, random seed") but never closed. Verified against a fresh build at ad52ca4c:

```scheme
(define c (make-default-comparator))
(define b (bag c 1 1))
(set->bag! b (set c 1 2))
[...]

Issue #2072, opened 2026-08-01
[SRFI-222] Compound objects: 5 of 10 procedures missing, make-compound does not flatten, compound-subobjects raises on a non-compound


baijum:
**Category:** SRFI
**Spec reference:** [SRFI 222 — Compound Objects](https://srfi.schemers.org/srfi-222/srfi-222.html), Specification
**Severity:** missing-feature + wrong-result

`lib/srfi/222.sld` implements **5 of the 10 procedures** SRFI 222 specifies, and two of the five it does implement do the wrong thing. `(cond-expand (srfi-222 ...))` and `(cond-expand ((library (srfi 222)) ...))` both answer `#t`, so portable code has no way to detect the shortfall.

## 1. Five of ten procedures are not exported

| Spec procedure | Exported by `(srfi 222)` |
|---|---|
| `make-compound` | yes |
| `compound?` | yes |
| `compound-subobjects` | yes (but see 3.) |
| `compound-length` | yes |
| `compound-ref` | yes |
| `compound-map` | **no** |
| `compound-map->list` | **no** |
| `compound-filter` | **no** |
| `compound-predicate` | **no** |
| `compound-access` | **no** |

```scheme
(import (scheme base) (srfi 222))
compound-map          ;; error[KP3001]: undefined variable 'compound-map'. Did you mean 'compound-ref'?
compound-map->list    ;; error[KP3001]: undefined variable 'compound-map->list'
compound-filter       ;; error[KP3001]: undefined variable 'compound-filter'
compound-predicate    ;; error[KP3001]: undefined variable 'compound-predicate'
compound-access       ;; error[KP3001]: undefined variable 'compound-access'
```

`compound-predicate` and `compound-access` are the two procedures the SRFI's own rationale is built around (its worked example is `(define (uni-member-hire-year obj) (compound-access teacher? hire-year #f obj))`), so what is missing is not a fringe of the spec.

## 2. `make-compound` does not flatten nested compound objects

> "Create a compound object containing the objects in *objs* in the specified order. If any object in *objs* is itself a compound object, it is flattened into its subobjects, which are then added to the compound object in sequence."

```scheme
(compound-length     (make-compound 1 2 (make-compound 3 4)))
;; expected: 4        actual: 3
(compound-subobjects (make-compound 1 2 (make-compound 3 4)))
;; expected: (1 2 3 4) actual: (1 2 #<<compound> (3 4)>)
```

[...]

Issue #1913, opened 2026-07-31
[SRFI-271] %random-port-make-from-seed accepts an all-zero seed; the port then emits only zeros forever


baijum:
**Category:** SRFI 271 (random ports) · **Severity:** wrong-result · **Found by:** systematic audit v2, Phase 2.1 (#1890)

```scheme
(read-bytevector 16 (%random-port-make-from-seed (make-bytevector 32 0)))
⟹ #u8(0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0)
```

An all-zero seed puts the generator in its degenerate fixed point, and it never leaves — every byte drawn from the port is zero, silently.

## Discriminating controls

1. **The byte-identical degenerate state is rejected by the sibling entry point:**
   ```scheme
   (%random-port-make-from-state
     (bytevector-append (bytevector 83 50 55 49 1 0) (make-bytevector 40 0)))
   ⟹ raises
   ```
2. **A non-degenerate seed works:** seed of 7 yields non-zero bytes, so the zeros come from the seed rather than from the port machinery.

## The defense already exists in two of the three places

- `isStateBytevector` (`primitives_random_port.zig:119-122`) has the check, **with a comment explaining why it is needed**.
- `lib/srfi/271/determinized.sld:87` also raises for it.
- `makeFromSeedFn` (`:58-68`) has neither.

So this is a missed site in an otherwise-defended invariant, not an undiscovered hazard.

## Suggested fix

Reject an all-zero seed in `makeFromSeedFn` the same way `isStateBytevector` does, or expand the seed through a mixing step that cannot produce the fixed point. Add the all-zero seed to `tests/scheme/srfi/srfi271.scm`.

**Verified:** fresh ReleaseSafe build at `261fde5f`, isolated `KAAPPI_HOME`, reproduced twice with both controls.

baijum:
Phase 2.11 (the small-file batch) re-reproduced this and found a **second, worse half of the same root cause**: the degenerate port has no valid state, so it violates SRFI 271's state round trip.

```scheme
(import (scheme base) (scheme write) (srfi 271 determinized))
(define z (%random-port-make-from-seed (make-bytevector 32 0)))

(random-port? z)                                  ⟹ #t     ; it IS a determinized port
(random-port-state? (%random-port-state z))       ⟹ #f     ; …whose own state is invalid
(make-random-port (%random-port-state z))         ⟹ raises random-port-initialization-error
```

[...]

PR #2277, opened 2026-08-10
Fix four SRFI audit findings: and-let*, SRFI 222, set->bag!, random seed

### LINKED ISSUE #2073: [SRFI-2] and-let* with claws and an empty body returns #t instead of the last claw's value
**Category:** SRFI
**Spec reference:** [SRFI 2 — `AND-LET*`](https://srfi.schemers.org/srfi-2/srfi-2.html), *Formal (Denotational) Semantics*:

> ```
> eval[ (AND-LET* (CLAW) ), env] = eval_claw[ CLAW, env ]
> ...
> eval_claw[ BOUND-VARIABLE, env ] = eval[ BOUND-VARIABLE, env ]
> eval_claw[ (EXPRESSION), env ]   = eval[ EXPRESSION, env ]
> eval_claw[ (VARIABLE EXPRESSION), env ] = eval[ EXPRESSION, env ]
> ```

**Severity:** wrong-result (silent; a true value is replaced by a different true value)

An `and-let*` with at least one CLAW and **no BODY** must return the value of the last claw's expression. Kaappi returns `#t` in every such case, discarding it.

## Minimal reproduction

```scheme
(import (scheme base) (srfi 2))
(and-let* ((x 1)))          ;; expected 1        actual #t
(and-let* (((+ 1 2))))      ;; expected 3        actual #t
(let ((v 5)) (and-let* (v)))  ;; expected 5      actual #t
(and-let* ((x 1) (y 2)))    ;; expected 2        actual #t
```

All three CLAW shapes the grammar admits — `(VARIABLE EXPRESSION)`, `(EXPRESSION)`, and `BOUND-VARIABLE` — are affected, at every claw count. Chibi 0.12.0 returns `1`, `3`, `5`, `2` respectively.

This is the idiom SRFI 2 exists for: `(and-let* ((x (assv key alist)) ((pair? (cdr x))) (v (cadr x))))` as a *value-producing* guarded lookup collapses to a bare `#t`, so the caller loses the result and cannot tell success-with-value from success.

## Discriminating controls

- `(and-let* ((x #f)))` ⟹ `#f` — the failure path is correct; only the success value is lost.
- `(and-let* ())` ⟹ `#t` — correct, and specified separately (`eval[(AND-LET* ()), env] = #t`).
- `(and-let* ((x 1)) x)` ⟹ `1` — with a body present the value flows normally, so the defect is confined to the empty-BODY case.
- `(and-let* () 1 2)` ⟹ `2`, `(and-let* ((x 2) (y (* x 3))) (+ x y))` ⟹ `8`, short-circuiting, and all four spec examples are correct.

## Source location

`lib/srfi/2.sld`. The first rule is

```scheme
((and-let* () body ...) (begin #t body ...))
```

so the recursion's base case yields `(begin #t)` ⟹ `#t` whenever the body is empty, regardless of how many claws were consumed on the way in. A conforming shape needs the two base cases the SRFI's own reference implementation has — `((_ ()) #t)` for the no-claw case and a separate last-claw rule that yields the claw's value — rather than one rule that folds both together.

## Verified

Fresh ReleaseSafe build at `1718a2e1`, macOS aarch64, isolated `KAAPPI_HOME`; unchanged under `--no-ir-opt`. Chibi 0.12.0 was used as the oracle and matches the spec on all 7 probes; **chibi does not reproduce this**, so it is a Kaappi porting defect, not a defect inherited from the reference implementation.

## Disabled test

`tests/scheme/srfi/srfi2.scm` (marked `;; FAIL: #2073`)

**Found by:** Systematic audit v2, Phase 3.7

baijum:
Fixes #2073, #2072, #2086, #1913 — four independent SRFI conformance defects from the systematic audit, one PR.

## #2073 — and-let* with claws and an empty body returns #t instead of the last claw's value

SRFI 2's formal semantics (`eval[(AND-LET* (CLAW)), env] = eval_claw[CLAW, env]`) require the last claw's value when there is no body; the base case expanded to `(begin #t body ...)`, collapsing the value-producing guarded-lookup idiom to a bare `#t`. The expansion now ends on the last claw and returns its value for all three claw shapes, keeping the `#f` short-circuit (`(and-let* ())` is still `#t`).

## #2072 — SRFI 222 ships 5 of 10 procedures; make-compound doesn't flatten; compound-subobjects raises on non-compounds

`lib/srfi/222.sld` now exports all ten spec procedures. `make-compound` flattens nested compounds recursively, `compound-subobjects` returns a one-element list for non-compounds (per spec), and the five missing procedures are implemented: `compound-map` (flattening compound results), `compound-map->list`, `compound-filter`, `compound-predicate`, `compound-access` — the two the SRFI's rationale is built around.

## #2086 — set->bag! silently drops the set's contribution for elements the bag already holds

`set->bag!` only inserted elements the bag didn't already contain, so an element the bag held kept its old count. It now unconditionally increments (`bag-increment! b k 1`), matching the reference implementation and chibi-scheme.

## #1913 — %random-port-make-from-seed accepts an all-zero seed; the port then emits only zeros forever

An all-zero seed puts the xoshiro256** state at its degenerate fixed point. The sibling entry points already rejected it (`isStateBytevector`, `determinized.sld`); `makeFromSeedFn` now does too. Unit test added in `src/tests_random_port.zig` plus Scheme-level tests in `srfi271.scm`.

## Testing

- All four previously-disabled regression suites are enabled and pass: `srfi2.scm` (45), `srfi222.scm` (63), `srfi113-audit.scm` (246), `srfi271.scm` (37)
- `zig build test` — full unit suite green (including the new #1913 test)
- `bash tests/scheme/run-all.sh` — 703 Scheme files pass; R7RS suite 1395/1395
- The single failing item (WASM differential: `sbc-constants-numeric`, `complex-neg-zero`, `large-index-bounds-1912`) is **pre-existing** — verified identical failures on the clean tree via `git stash`

### DIFF
--- a/lib/srfi/222.sld
+++ b/lib/srfi/222.sld
@@ -1,15 +1,37 @@
 (define-library (srfi 222)
   (import (scheme base))
   (export make-compound compound? compound-subobjects
-          compound-length compound-ref)
+          compound-length compound-ref
+          compound-map compound-map->list compound-filter
+          compound-predicate compound-access)
   (begin
     (define-record-type <compound>
       (make-compound-record subobjects)
       compound?
-      (subobjects compound-subobjects))
+      (subobjects %compound-subobjects))
 
-    (define (make-compound . subobjects)
-      (make-compound-record subobjects))
+    ;; "If obj is a compound object, returns a list of its subobjects;
+    ;; otherwise, returns a list containing only obj." (#2072: the bare
+    ;; record accessor used to raise on a non-compound.)
+    (define (compound-subobjects obj)
+      (if (compound? obj)
+          (%compound-subobjects obj)
+          (list obj)))
+
+    ;; Subobjects are never compounds: make-compound (and compound-map, which
+    ;; builds through it) flattens nested compound objects, so a compound's
+    ;; subobject list is always flat (#2072).
+    (define (flatten objs)
+      (let loop ((in objs) (out '()))
+        (if (null? in)
+            (reverse out)
+            (loop (cdr in)
+                  (if (compound? (car in))
+                      (append (reverse (compound-subobjects (car in))) out)
+                      (cons (car in) out))))))
+
+    (define (make-compound . objs)
+      (make-compound-record (flatten objs)))
 
     (define (compound-length obj)
       (if (compound? obj)
@@ -21,4 +43,40 @@
           (list-ref (compound-subobjects obj) k)
           (if (= k 0)
               obj
-              (error "compound-ref: index out of range" k))))))
+              (error "compound-ref: index out of range" k))))
+
+    (define (compound-map mapper obj)
+      ;; Mapper results that are themselves compounds are flattened into the
+      ;; result's subobjects, in sequence.
+      (make-compound-record
+       (append-map (lambda (o)
+                     (let ((r (mapper o)))
+                       (if (compound? r)
+                           (compound-subobjects r)
+                           (list r))))
+                   (compound-subobjects obj))))
+
+    (define (compound-map->list mapper obj)
+      (map mapper (compound-subobjects obj)))
+
+    (define (compound-filter pred obj)
+      (make-compound-record (filter pred (compound-subobjects obj))))
+
+    (define (compound-predicate pred obj)
+      (if (or (pred obj)
+              (and (compound? obj)
+                   (let loop ((os (compound-subobjects obj)))
+                     (cond ((null? os) #f)
+                           ((pred (car os)) #t)
+                           (else (loop (cdr os)))))))
+          #t
+          #f))
+
+    (define (compound-access predicate accessor default obj)
+      (cond ((predicate obj) (accessor obj))
+            ((compound? obj)
+             (let loop ((os (compound-subobjects obj)))
+               (cond ((null? os) default)
+                     ((predicate (car os)) (accessor (car os)))
+                     (else (loop (cdr os))))))
+            (else default)))))
--- a/lib/srfi/222.sld
+++ b/lib/srfi/222.sld
@@ -10,6 +10,25 @@
       compound?
       (subobjects %compound-subobjects))
 
+    ;; Local helpers: `filter' and `append-map' are not (scheme base)
+    ;; bindings in R7RS — kaappi resolves them only through its vm.globals
+    ;; fallback for library bodies, and other R7RS implementations would not
+    ;; load the library at all. The SRFI 222 reference implementation defines
+    ;; its own `filter', as does lib/srfi/217.sld, so this library stays
+    ;; loadable by any R7RS implementation.
+    (define (filter pred lst)
+      (let loop ((in lst) (out '()))
+        (cond ((null? in) (reverse out))
+              ((pred (car in)) (loop (cdr in) (cons (car in) out)))
+              (else (loop (cdr in) out)))))
+
+    ;; Apply f to every element and append the results, one level deep.
+    (define (append-map f lst)
+      (let loop ((in lst) (out '()))
+        (if (null? in)
+            (reverse out)
+            (loop (cdr in) (append (reverse (f (car in))) out)))))
+
     ;; "If obj is a compound object, returns a list of its subobjects;
     ;; otherwise, returns a list containing only obj." (#2072: the bare
     ;; record accessor used to raise on a non-compound.)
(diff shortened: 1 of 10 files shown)


baijum (commented):
Reviewed all four fixes against the SRFI specs and verified behaviorally against a build of this branch. **All four are spec-conformant and well-tested** — the enabled suites pass at exactly the counts claimed (srfi2 45, srfi222 63, srfi113-audit 246, srfi271 37), and I re-ran the SRFI 222 spec's own worked examples (`uni-member-hire-year`, `compound-predicate`) which all match.

Specific things I checked and confirmed correct:

- **#2073 (and-let*)** — The three new last-claw-no-body rules are placed *before* the recursion rules, so a single-claw form always takes them; the three claw shapes are structurally disjoint, so ordering among them is unambiguous. Crucially the `(EXPRESSION)` and `(VAR EXPR)` shapes bind with `let` rather than re-referencing the expression, so the value-producing expression runs exactly once — I verified a side-effecting last claw evaluates once (`n` = 1), not twice. Hygiene of the introduced `t` is fine under Kaappi's expander.
- **#2072 (SRFI 222)** — `compound-access` correctly checks `(predicate obj)` first (matching the spec's "If obj satisfies predicate, accessor is applied to obj" — including when obj is itself a compound), then falls to the first matching subobject, then default. `compound-predicate`, `compound-map` result-flattening, and the recursive `make-compound` flatten all match the spec. Relying on `filter`/`append-map` from `(scheme base)` is consistent with the existing repo pattern (134/158/189/217/247.sld), so no portability regression. The mapper-returns-a-compound flattening path is covered by a test.
- **#2086 (set->bag!)** — Now unconditionally `bag-increment! b k 1`; verified an element the bag already holds goes from 2 to 3 after merging a set that contains it.
- **#1913 (zero seed)** — The guard only inspects the first 32 (seeded) bytes, which is exactly right since trailing bytes never reach the xoshiro state. The `guard`-based regression tests are structured to fail (not pass vacuously) if the primitive stops raising.

One non-blocking observation inline. Nice, tightly-scoped PR.

src/primitives_random_port.zig
```diff
@@ -59,6 +59,16 @@ fn makeFromSeedFn(args: []const Value) PrimitiveError!Value {
...
     if (bv.data.len < seed_len) return primitives.typeError("%random-port-make-from-seed", "bytevector of length >= 32", args[0]);
+    // An all-zero seed puts the xoshiro256** state at its degenerate fixed
+    // point, which emits only zero bytes forever. The sibling entry points
+    // already reject it (isStateBytevector above; bytevector-all-zero? in
+    // determinized.sld); without the same guard here the raw primitive
+    // silently built the broken generator (#1913).
+    for (bv.data[0..seed_len]) |b| {
+        if (b != 0) break;
+    } else {
+        return primitives.argError("%random-port-make-from-seed", "seed must contain at least one nonzero byte", .{});
```
baijum, line 70: Non-blocking / for confirmation: this rejects an all-zero seed with `argError` (KP3007), whereas the sibling all-zero-*state* rejection goes through `isStateBytevector` → `typeError` (KP3002) at `makeFromStateFn`. `argError` is arguably the *more* accurate classification here (a correctly-typed, correctly-sized bytevector whose value is degenerate is an invalid-argument, not a type error), so I'd keep it — just flagging that the two degenerate-input paths now report different error codes. Both are `%`-internal and both are catchable by `guard`, so impact is nil; only worth a second look if the SRFI-271 library layer ever depends on which code the raw primitive raises.

lib/srfi/222.sld
```diff
@@ -1,15 +1,37 @@
 (define-library (srfi 222)
   (import (scheme base))
   (export make-compound compound? compound-subobjects
-          compound-length compound-ref)
+          compound-length compound-ref
+          compound-map compound-map->list compound-filter
+          compound-predicate compound-access)
```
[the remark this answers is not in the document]
baijum: Valid finding — fixed. `filter` and `append-map` are tagged `.srfi_1` (not `.scheme_base`) in `src/primitives_srfi1.zig`, so they reach this library only through kaappi's `vm.globals` fallback for library bodies (`lookupGlobalLocked` in `src/vm_dispatch_helpers.zig`), which other R7RS implementations don't have. My initial "they resolve from (scheme base)" check was misleading: scripts/REPL evaluate against the full global map, unlike a library env.

The library now defines both helpers locally using only R7RS base functionality, matching the SRFI 222 reference implementation (which defines its own `filter`) and `lib/srfi/217.sld` (same pattern). Verified by poisoning the global `filter`/`append-map` bindings to non-procedures before `(import (srfi 222))`: the library still loads and all 63 srfi222 tests pass.

src/primitives_random_port.zig
baijum: Agreed, keeping `argError` (KP3007). The two paths reject different things: the state path folds several structural checks (length, magic, version, out-position, non-zero state words) into one `isStateBytevector` predicate, so a type-style error is the natural classification for "not a valid state object". The seed path's only invalid *value* — the bytevector is already type- and length-checked — is the all-zero seed, an invalid argument rather than a type error. This also matches the codebase guidance (AGENTS.md): don't stretch `typeError` over failures that aren't type errors. Both are `%`-internal and catchable by `guard`, and neither the `.sld` layer nor any test depends on the specific code, so the divergence is cosmetic.

PR #2279, opened 2026-08-10
Fix SRFI-170 validation gaps and implement the posix-error protocol (Fixes #1977, #1978)

### LINKED ISSUE #1977: [SRFI-170] nice, set-environment-variable! and four optional arguments accept wrongly-typed values silently

baijum:
Two SRFI-170 audit findings (systematic audit v2, Phase 2.12, #1890), fixed together because they touch the same file and the same raise helpers.

## #1977 — mistyped arguments were silently discarded

| call | before | after |
|---|---|---|
| `(nice "x")` | renice'd the process by +1 (type test missing) | type error |
| `(set-environment-variable! "KA=B" "v")` | returned normally, set nothing (setenv return discarded) | posix error `EINVAL` |
| `(create-directory p "notanum")` | mode 0755 silently | type error |
| `(create-fifo p "notanum")` | mode 0664 silently | type error |
| `(create-temp-file 42)` | default prefix silently | type error |
| `(set-file-times p "garbage" "garbage")` | both stamps rewritten to now | type error |
| `(set-file-times p 1000000)` | one-time accepted | KP3007 (spec: exactly one time is an error) |
| `(file-info p #t 'extra)` | surplus arg ignored | arity mismatch |

The seven variadic SRFI-170 specs now declare a new `.range` arity (`min..max`) so surplus arguments are rejected at the arity layer (enforced in `vm_calls`/`vm_dispatch`/lint/LSP/REPL display).

## #1978 — posix-error protocol, error taxonomy, cross-thread records

- **`posix-error?` / `posix-error-name` / `posix-error-message`** (SRFI-170 §3.1) are exported from `(srfi 170)`. Every SRFI-170 syscall failure now stamps the captured `errno` onto the condition object (`ErrorObject.posix_errno`); `posix-error-name` scans Zig's per-OS `std.c.E` enums so names (`ENOENT`, `ENOTDIR`, `EACCES`, `ELOOP`, `ENAMETOOLONG`, …) are portable even though the numbers differ; `posix-error-message` is `strerror(3)`. Non-syscall raises pass errno 0 explicitly, so `posix-error?` stays `#f` for them.
- **Argument-range failures are no longer file errors.** `(set-file-mode p #o10000)`, `(set-umask! -1)`, `(nice 999999999999)`, out-of-range uid/gid and an over-long temp prefix are raised as KP3007 invalid-argument conditions (messages unchanged) and answer `#f` to `file-error?`.
- **`file-info` / `user-info` / `group-info` now cross the SRFI-18 thread boundary.** They are pure value records (scalars + owned string bytes) but were listed as `UncopyableType` beside ports/continuations/fibers; `gc_deep_copy.zig` copies them by value now (`directory-object` stays uncopyable — live `DIR*`). The "uncopyable type (port, continuation, etc.)" messages name the actual uncopyable set.

## Verification

- Audit suite's disabled `FAIL` rows for both issues enabled; bug-pinning controls removed
- New section G2: posix-error protocol matrix incl. errno survival through the thread boundary (`uncaught-exception-reason`)
- `srfi18-deepcopy-matrix-audit.scm` flips the three SRFI-170 rows from refused → cross
- `tests_deepcopy.zig`: unit tests for the three copyable types, the `directory-object` refusal, and `posix_errno` preservation on error-object deep copy
- `zig build test` green incl. `-Dgc-stress=true`; all 2099 Scheme files + 1395 R7RS tests pass; wasm + all 8 cross-compile targets build

### DIFF
--- a/src/primitives_filesystem.zig
+++ b/src/primitives_filesystem.zig
@@ -45,8 +45,8 @@ fn errnoName(errno_val: c_int) []const u8 {
     return "EUNKNOWN";
 }
 
-fn validateMode(gc: *GC, val: Value) PrimitiveError!std.c.mode_t {
-    if (!types.isFixnum(val)) return primitives.typeError("set-file-mode", "integer", val);
+fn validateMode(gc: *GC, proc: []const u8, val: Value) PrimitiveError!std.c.mode_t {
+    if (!types.isFixnum(val)) return primitives.typeError(proc, "integer", val);
     const n = types.toFixnum(val);
     // POSIX permission modes use only the low 12 bits (07777). Bounding by
     // maxInt(mode_t) is platform-dependent: mode_t is u16 on macOS but u32
@@ -118,7 +118,9 @@ fn doStat(path: [*:0]const u8, follow: bool, errno_out: *c_int) ?StatResult {
         // as zero, exactly like SRFI-170 suggests for hosts without them.
         var wbuf: platform.WPathBuf = undefined;
         const wpath = platform.widen(&wbuf, std.mem.sliceTo(path, 0)) orelse {
-            errno_out.* = std.c._errno().*;
+            // widen is a UTF-8→UTF-16 conversion, not a syscall — errno is
+            // stale here, so report no errno.
+            errno_out.* = 0;
             return null;
         };
         var st: platform.win.Stat64 = undefined;
@@ -250,7 +252,7 @@ pub const specs = [_]primitives.PrimSpec{
     .{ .name = "user-effective-uid", .func = &userEffectiveUidFn, .arity = .{ .exact = 0 }, .libs = LS.initOne(.srfi_170), .sandbox = false, .wasm = false },
     .{ .name = "user-effective-gid", .func = &userEffectiveGidFn, .arity = .{ .exact = 0 }, .libs = LS.initOne(.srfi_170), .sandbox = false, .wasm = false },
     .{ .name = "user-supplementary-gids", .func = &userSupplementaryGidsFn, .arity = .{ .exact = 0 }, .libs = LS.initOne(.srfi_170), .sandbox = false, .wasm = false },
-    .{ .name = "nice", .func = &niceFn, .arity = .{ .variadic = 0 }, .libs = LS.initOne(.srfi_170), .sandbox = false, .wasm = false },
+    .{ .name = "nice", .func = &niceFn, .arity = .{ .range = .{ .min = 0, .max = 1 } }, .libs = LS.initOne(.srfi_170), .sandbox = false, .wasm = false },
     .{ .name = "set-environment-variable!", .func = &setEnvVarFn, .arity = .{ .exact = 2 }, .libs = LS.initOne(.srfi_170), .sandbox = false, .wasm = false },
     .{ .name = "delete-environment-variable!", .func = &deleteEnvVarFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_170), .sandbox = false, .wasm = false },
     .{ .name = "terminal?", .func = &terminalP, .arity = .{ .exact = 1 }, .libs = LS.initOne(.srfi_170), .sandbox = false, .wasm = false },
@@ -343,10 +345,10 @@ fn posixErrorP(args: []const Value) PrimitiveError!Value {
     return if (err.posix_errno != 0) types.TRUE else types.FALSE;
 }
 
-fn expectPosixError(args: []const Value) PrimitiveError!*types.ErrorObject {
-    if (!types.isErrorObject(args[0])) return primitives.typeError("posix-error-name", "posix-error", args[0]);
+fn expectPosixError(comptime proc: []const u8, args: []const Value) PrimitiveError!*types.ErrorObject {
+    if (!types.isErrorObject(args[0])) return primitives.typeError(proc, "posix-error", args[0]);
     const err = types.toObject(args[0]).as(types.ErrorObject);
-    if (err.posix_errno == 0) return primitives.typeError("posix-error-name", "posix-error", args[0]);
+    if (err.posix_errno == 0) return primitives.typeError(proc, "posix-error", args[0]);
     return err;
 }
 
@@ -354,16 +356,20 @@ fn expectPosixError(args: []const Value) PrimitiveError!*types.ErrorObject {
 /// errno numbers differ across POSIX systems but the names are shared, so
 /// the spec returns the name rather than the code.
 fn posixErrorName(args: []const Value) PrimitiveError!Value {
-    const err = try expectPosixError(args);
+    // Copy the scalar before allocating: allocSymbol may collect, and the
+    // raw ErrorObject pointer must not survive an allocation unrooted
+    // (the args register file roots the value itself, but the gc-safety
+    // rule is pointer-across-allocation, so read the field out first).
+    const errno_val = (try expectPosixError("posix-error-name", args)).posix_errno;
     const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
-    return gc.allocSymbol(errnoName(err.posix_errno)) catch return PrimitiveError.OutOfMemory;
+    return gc.allocSymbol(errnoName(errno_val)) catch return PrimitiveError.OutOfMemory;
 }
 
 /// `posix-error-message` — the strerror(3) text for the captured errno.
 fn posixErrorMessage(args: []const Value) PrimitiveError!Value {
-    const err = try expectPosixError(args);
+    const errno_val = (try expectPosixError("posix-error-message", args)).posix_errno;
     const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
-    const text = strerror(err.posix_errno);
+    const text = strerror(errno_val);
     return gc.allocString(std.mem.span(text)) catch return PrimitiveError.OutOfMemory;
 }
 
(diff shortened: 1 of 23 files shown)


src/primitives_filesystem.zig
```diff
@@ -239,17 +299,80 @@ fn raiseFileError(gc: *GC, msg_text: []const u8, irritant: Value) PrimitiveError
...
+fn posixErrorP(args: []const Value) PrimitiveError!Value {
+    if (!types.isErrorObject(args[0])) return types.FALSE;
+    const err = types.toObject(args[0]).as(types.ErrorObject);
+    return if (err.posix_errno != 0) types.TRUE else types.FALSE;
+}
+
+fn expectPosixError(args: []const Value) PrimitiveError!*types.ErrorObject {
+    if (!types.isErrorObject(args[0])) return primitives.typeError("posix-error-name", "posix-error", args[0]);
+    const err = types.toObject(args[0]).as(types.ErrorObject);
+    if (err.posix_errno == 0) return primitives.typeError("posix-error-name", "posix-error", args[0]);
```
baijum: `expectPosixError` hard-codes `"posix-error-name"` as the procedure name in both `typeError` calls, but it also backs `posix-error-message`. So `(posix-error-message 42)` and `(posix-error-message (error "x"))` raise `type error in 'posix-error-name': ...`, naming the wrong procedure. The G2 audit only checks `(error-object? e)` on these paths, so the mislabel isn't caught.

Suggest passing the caller's name in (e.g. `fn expectPosixError(comptime proc: []const u8, args: ...)`, called with `"posix-error-name"` / `"posix-error-message"`). Low severity — diagnostic text only.

baijum (commented):
Reviewed the full diff (src + audit/scheme tests + docs). This is a careful, well-scoped PR: the errno-at-the-syscall discipline is right (the `-errno` recovery from raw `statx` is exactly the trap the CI leg caught, and every `raiseFileError` reads `std.c._errno()` before any allocation intervenes), the `.range` arity is threaded through all six exhaustive switch sites plus lint/LSP/REPL display, the taxonomy split (KP3007 argError vs `.file` posix error) matches the spec intent, and the deep-copy + posix-error protocol are backed by real cross-thread regression tests. `zig`'s exhaustive-switch checking guarantees no arity site was missed. Three minor observations below; none is blocking.

**1. `nice` is left unbounded (minor consistency, #1977).** `nice` stays `.arity = .{ .variadic = 0 }` (src/primitives_filesystem.zig:253), so `(nice 1 2 3)` is still silently accepted with the surplus ignored — the exact silent-surplus class this PR converts everywhere else. SRFI-170's signature is `(nice [delta])`, so `.range = .{ .min = 0, .max = 1 }` would make it consistent with the seven procedures you did bound (and the audit has no `(nice 1 2)` arity probe, so the gap is untested). Was leaving it out deliberate?

**2. Duplicated uncopyable-type message string (minor maintainability).** The full enumerated list — "port, continuation, fiber, mutex, condition variable, FFI callback, directory object, environment, ephemeron, guardian, or transport cell" — is now hand-written in three sites (src/primitives_fiber.zig:1345, src/primitives_srfi18.zig:610 and :1295) with nothing tying it to the actual `gc_deep_copy.zig` refusal switch. It's correct and complete today, but a future tag added to/removed from that switch won't force these strings to update — the same drift shape called out for hand-kept parallel lists elsewhere in the codebase. A shared `const` or a comptime-built list would remove the risk. Optional.

The inline note on `expectPosixError` is the only one I'd suggest actually changing.

baijum:
Thanks for the thorough reviews (CodeRabbit + @baijum). I've gone through every comment; here's the disposition. Commit `4531963c` addresses the actionable ones.

## Fixed

- **`nice` unbounded arity** (CodeRabbit ×2 + baijum) — genuine oversight, the one variadic SRFI-170 signature I didn't bound. Now `.range = .{ .min = 0, .max = 1 }`; audit pins `(nice 0 'extra)`.
- **`validateMode`/`expectPosixError` hardcode one caller's proc name** (CodeRabbit + baijum) — `(create-directory d "x")` reported `type error in 'set-file-mode'` and `(posix-error-message 42)` named `posix-error-name`. Both helpers now take the caller's name; all four `validateMode` sites and both accessors pass their own.
- **Unrooted `ErrorObject*` across allocation in `posix-error-name`/`-message`** (CodeRabbit, high) — copy `posix_errno` out before `allocSymbol`/`allocString`.
- **Monotonic time objects accepted by `set-file-times`** (CodeRabbit) — `(set-file-times p (monotonic-time) …)` wrote an arbitrary-epoch clock as wall-clock time. Now rejected; audit adds the rejection plus a `(posix-time)` round-trip control.
- **`directory-object` refusal test couldn't detect a regression** (CodeRabbit) — the old `out-of-thread` helper returned `#t` for both a successful copy and a clean refusal. Split into `copies?`/`refused?` so each direction is pinned.
- **Locale-dependent strerror literal in the G2 audit** (CodeRabbit) — now asserts `(string? …)` like the sibling checks.
- **`doStat` Windows `widen` failure read stale errno** (CodeRabbit, 2nd pass) — it's a UTF conversion, not a syscall; reports 0 now.
- **vm_dispatch arity-switch triplication** (CodeRabbit) — `checkNativeArity` is now `pub` and the two dispatch sites call it; one copy to maintain.
- **check_lint plural for a `0..1` range** (CodeRabbit) — renders "expects 0 to 1 arguments".
- **Stale docs/comment counts**: `thread-value-sharing.md:287` 14→11-tag; `platform.zig` "We ignore the return either way" note.
- **`tests_deepcopy` partial field assertions** (CodeRabbit) — every `FileInfo`/`UserInfo` field is now checked.

## Evaluated and skipped, with reasons

- **Split `tests_gc_tracing.zig`** (CodeRabbit, "heavy lift") — that file is 1,760 lines *before* this PR; I only added one name to the field inventory. Splitting a pre-existing file is unrelated scope for a bug-fix PR.
- **`strerror` → `strerror_r`** (CodeRabbit, "Major") — the race window is a few instructions (the text is copied into a GC string immediately), the spec explicitly says the message "may be, or may include, the output of the strerror() function", and `strerror_r` doesn't exist on Windows (`strerror_s` has a different signature), so the "fix" adds a platform matrix for a theoretical race. Left as-is deliberately.
- **Range-checked `@intCast`s in `timeArgToTimespec`** (CodeRabbit, "Major") — every supported target has i64 `timespec` fields, fixnums top out at 2^47, and `Srfi18Time` seconds/nanoseconds are i64, so none of these casts can overflow; the kernel rejects out-of-range nsec with EINVAL → a clean posix error. The "panic at runtime" premise doesn't hold for this value representation.
- **Handle `setEnv`/`unsetEnv` status in `test_runner.zig`/`tests_ffi.zig`** (CodeRabbit, "high") — those call sites use fixed literal names (`KAAPPI_TEST_SEED`, `KAAPPI_TEST_EMIT`, `KAAPPI_HOME`), for which EINVAL is unreachable and ENOMEM is already propagated by the existing `catch`; the status carries no information there.
- **Import `(scheme process-context)` in the matrix audit** (CodeRabbit) — Kaappi's `(scheme base)` already exports `exit` as an extension (verified: `(procedure? exit)` → `#t` with only `(scheme base)` imported), so the file loads and exits correctly today.

## Verification

Full suite green after the fixes: 2099 Scheme files + 1395 R7RS tests, unit + `-Dgc-stress`, wasm build, all cross-compile targets, and the filesystem audit re-run in an Alpine container (x86_64 Linux).

Issue #1832, opened 2026-07-29
syntax-rules template literal loses hygienic rename on 2nd internal expansion when it collides with a pre-existing top-level binding


baijum:
## Summary

A `syntax-rules` macro whose template contains a literal identifier (not
a pattern variable) that is *also* used, at the same spelling, as a
pattern-variable argument at a particular use site, and where that
spelling *also* already has a top-level binding before the macro is
used, gets expanded more than once internally for that one use (visible
by having the expansion call into a procedural macro that prints the raw
form it receives). The first internal expansion correctly hygienically
renames the template's own literal. The second does not -- it leaks
through unrenamed, becoming indistinguishable from the pattern-variable
argument of the same spelling.

Without the pre-existing top-level binding of that name, both internal
expansions correctly and consistently rename the template literal (with
different, but equally fresh, rename counters) -- no collision.

## Reproduction

```scheme
;; lib/srfi/repro2.sld
(define-library (srfi repro2)
  (import (scheme base) (scheme write) (srfi 211 explicit-renaming))
  (export probe)
  (begin
    (define-syntax probe
      (er-macro-transformer
       (lambda (form rename compare)
         (write (list 'SAW form))
         (newline)
         (list (rename 'quote) 'ok))))))
```

```scheme
;; main.scm
(import (scheme base) (scheme write) (srfi repro2))

(define a #f)                 ; <-- pre-existing top-level binding

(define-syntax def
  (syntax-rules ()
    ((def b)
     (probe a b))))            ; `a` is def's own template literal; `b` is
                                ; a pattern variable

(def a)                        ; called with the SAME spelling, `a`
```

Output:
```
(SAW (probe __hyg_1_a a))
(SAW (probe a a))
```

The second line should read `(probe __hyg_N_a a)` for some fresh `N`,
consistently distinct from the plain `a` (the pattern-variable argument,
correctly referring to the top-level binding). Instead the template's
own `a` collapses to bare `a`, becoming indistinguishable from the
argument.

Removing the top-level `(define a #f)` (or changing the macro's use site
to `(def a2)`, so the two identifiers no longer share a spelling) makes
both lines come out correctly and consistently renamed.

## Isolation notes

- Reproduces with no records, no ER-macro-generated `define`s, and no
  SRFI 150/213/237 involved -- pure `(srfi 211 explicit-renaming)` plus a
  `syntax-rules` wrapper macro and `(scheme base)`.
- The corruption is present in the raw `form` argument the transformer
[...]

PR #2280, opened 2026-08-10
Fix SRFI 150 hygiene collapse: resolve field identity at expansion time (Fixes #2051)

### LINKED ISSUE #2051: [SRFI-150] All 4 expected failures are one defect: hygienic field names are round-tripped through quote, which strips the rename — not #1832
**Category:** SRFI (implementation strategy) + Doc-truth
**Spec reference:** [SRFI 150](https://srfi.schemers.org/srfi-150/srfi-150.html) — "Hygienic ERR5RS Record Syntax"; its own reference test suite asserts `1` and `2` for the two `Hygiene 1` accessors. Underlying rule is R7RS 4.3 (macro-introduced identifiers are distinct from same-spelled identifiers at the use site).
**Severity:** wrong-result (silent; no diagnostic in 2 of the 4 cases)

All four `test-expect-fail` cases in `tests/scheme/srfi/srfi150.scm` share **one** root cause, and it is **not** the closed #1832 they currently cite, nor #2003.

## Root cause

`lib/srfi/150.sld` carries field names from macro-expansion time to run time **inside `quote`**, then looks them up by `equal?` on the raw symbol. Hygiene renames do not survive `quote`:

```scheme
(eq? '__hyg_2_a 'a)   ;; => #t
```

So two identifiers the expander correctly kept distinct collapse into one field name before the runtime lookup ever sees them.

The expansion itself is **correct** — `kaappi expand` on the failing program shows the two field names properly distinguished:

```scheme
(list->vector (quote ((immutable __hyg_2_a) (immutable a))))
(record-accessor __hyg_1_<r> (quote __hyg_2_a))
(record-accessor __hyg_1_<r> (quote a))
```

Both `quote`d symbols then strip to `a`. The strip is *not* at read time — `kaappi ast` shows `(quote __hyg_2_a)` intact — so it happens in the compiler's quote handling, where it is correct and required in general (a `syntax-rules` template's `'foo` must yield `foo`, not `__hyg_1_foo`). The defect is that SRFI 150's strategy is unsound on top of it.

## Minimal reproduction

```scheme
(import (scheme base) (scheme write) (srfi 150))
(define-syntax def
  (syntax-rules ()
    ((def b mk ga gb)
     (define-record-type <r> (mk a b) r? (a ga) (b gb)))))
(def a mk ga gb)                                  ; use site supplies `a`
(write (list (ga (mk 1 2)) (gb (mk 1 2))))        ; expected (1 2), actual (2 2)
```

## Discriminating controls

| # | Program | Expected | Actual | Reading |
|---|---|---|---|---|
| C4 | above, **with** a preceding `(define a #f)` | `(1 2)` | `(2 2)` | fails |
| **C5** | above, **no** top-level `a` at all | `(1 2)` | `(2 2)` | **fails identically — so a pre-existing top-level binding is _not_ required** |
| C6 | above, use site passes `q` instead of `a` | `(1 2)` | `(1 2)` | passes — the trigger is purely the spelling collision |
| C1 | C4's exact shape with **plain `(scheme base)` `define-record-type`** | `(1 2)` | `(1 2)` | passes, and matches chibi 0.12 |

**C5 is the one that matters.** #1832's defining condition is a pre-existing top-level binding of the colliding spelling; C5 has none and fails the same way. #1832's own regression test (`tests/scheme/hygiene/free-global-vs-arg-1832.scm`) passes 6/6 on this build, and C1 shows its exact shape works under plain R7RS records. **#1832's fix is complete; the annotations citing it are misattributed.**

It is also not #2003: that one needs the captured global to hold a **procedure** (its own table shows non-procedure globals behaving correctly), and here nothing is bound at all.

The same mechanism explains the other two cases:

- **Hygiene 2** — same collision via an inherited field; the no-top-level-binding control fails identically, and renaming the template's own field to a non-colliding spelling makes it pass.
- **Alex Shinn's tuple example** — `deftuple` emits `__hyg_1_tmp` and `__hyg_2_tmp` (distinct at expansion, confirmed via `kaappi expand`); both strip to `tmp`, so the fields collapse and `(make-point 1 2)` reads back `(2 2)`.

## Doc-truth component

`lib/srfi/150.sld`'s header states the opposite of the observed behaviour:

> "a stored field-name/accessor-name symbol retains its full hygienic spelling (rename prefix included) across the definition/lookup boundary. Plain `equal?` on that raw symbol therefore already implements both bound-identifier=? ... and free-identifier=? ... — verified against the exact scenario each SRFI 150 hygiene test group requires"

The rename prefix does not survive `quote`, and the scenario it claims to have verified is exactly the failing one. `tests/scheme/srfi/srfi150.scm`'s per-case annotations inherit the same wrong mechanism ("collides in spelling with the pre-existing top-level `(define a #f)` above"). Both are corrected in the PR that files this.

A fix needs SRFI 150 to stop round-tripping hygienic identity through `quote` — e.g. keying the SRFI 213 property table on a value that is not a quoted symbol, or resolving field names entirely at expansion time.

## Note (minor, same mechanism)

The strip is by spelling, not by provenance: a symbol a user literally types as `'__hyg_2_a` also reads back as `a`. Not worth its own issue, but it is why the collapse is unconditional.

**Verified:** fresh ReleaseSafe build at `08828884`, isolated `KAAPPI_HOME`, cache cold. All four cases reproduce deterministically; oracle for C1 is chibi-scheme 0.12 (which has no SRFI 150, so the oracle for the SRFI-specific cases is the SRFI's own reference test suite).
**Source location:** `lib/srfi/150.sld` (`field-alist-ref` / `resolve-field-name` and the `(quote ...)` field lists its `er-macro-transformer` emits)
**Disabled test:** `tests/scheme/srfi/srfi150.scm` — 4 `test-expect-fail` cases, re-annotated to cite this issue

**Found by:** Systematic audit v2, Phase 3.10

baijum:
## Summary

Fixes #2051 — all four `test-expect-fail` cases in `tests/scheme/srfi/srfi150.scm` share one root cause: `lib/srfi/150.sld` carried field names from macro-expansion time to run time inside `quote`, and the compiler strips a `__hyg_N_` rename from any quoted datum (`compileQuote`'s `stripHygieneFromDatum`, correct and required for `syntax-rules` templates). Two field identifiers the expander had correctly distinguished (e.g. `__hyg_2_a` and `a`) therefore stripped to one runtime name and collapsed into a single field — `(get-a (mk 1 2))` and `(get-b (mk 1 2))` both read back `2`.

The attribution to #1832 (a pre-existing top-level binding of the colliding spelling) was wrong on both halves: #1832 is fixed and its exact shape passes under plain `(scheme base)` `define-record-type`, and the no-binding control (C5) fails identically.

## Fix

Resolve field identity **entirely at macro-expansion time**, while the renamed symbols are still in hand — nothing hygienic crosses the expansion/runtime boundary:

- **Constructor specs resolve to numeric absolute layout indices.** A spec entry matches the current form's own fields by full spelling (bound-identifier=? in this engine's rename-by-spelling model), then inherited fields by hygiene-stripped spelling against the parent's stored property (free-identifier=?). `named-constructor` fills the field vector by index; no runtime by-name lookup happens at all.
- **Own fields get deduped runtime names** for the rtd and accessor/mutator creation: the stripped spelling, with a numeric suffix when two own fields strip to the same spelling (the Hygiene 1 shape). An own field matching an inherited field's spelling is deliberately *not* deduped — that is ordinary shadowing.
- **The property table stores keys and indices only** — total field count plus stripped-spelling keys to absolute indices, never renamed symbols.
- **The emitted type-name binding is hygiene-stripped too.** A template-introduced `__hyg_N_<t>` reference whose base `<t>` is already bound is intercepted by the #1832 referential-transparency alias (it loads the pre-existing global's value even inside the same expansion that defines it), so accessors used to bind against the *old* record type whenever a macro redefined an already-bound type name. This was the separate hard-error at the Hygiene 1 definition site.
- **Constant field names now actually work** (the header promised them): a non-identifier field name gets a generated `field-<index>` runtime name, making SRFI 150's "field names may be constants" conformant (previously errored in `symbol->string`).

## Tests

- Reference suite passes in full: **25/25** (previously 21 passes + 4 expected failures). The four `test-expect-fail` annotations and their obsolete mechanism comments are removed.
- New regression tests: the issue's C5 (no pre-existing binding) and C6 (non-colliding spelling) controls, the minimal repro verbatim, and constant field names (numbers, strings, inherited constants).
- Verified: `tests/scheme/run-all.sh` (703 scheme files + 1395 R7RS tests, 0 failures), `zig build test`, `zig build test -Dgc-stress=true`, markdownlint, and the `kaappi test` runner.
- `docs/dev/srfi-implementation-notes.md` updated (the SRFI 150 section described the old, unsound mechanism and the "21 of 25" gap).

## Files

- `lib/srfi/150.sld` — the redesign + header/doc rewrite
- `tests/scheme/srfi/srfi150.scm` — unmark the 4 failures, add regression controls
- `docs/dev/srfi-implementation-notes.md` — SRFI 150 section rewrite

### DIFF
--- a/lib/srfi/150.sld
+++ b/lib/srfi/150.sld
@@ -163,7 +163,11 @@
   ;; below. They used to arrive with (scheme base), which reserved their
   ;; names against every user library (kaappi#1856).
   (import (scheme base) (srfi 211 explicit-renaming) (srfi 213)
-          (srfi 237) (srfi 237 primitives) (kaappi primitives))
+          (srfi 237) (srfi 237 primitives) (kaappi primitives)
+          ;; string-prefix?/string-index are (srfi 13) exports; imported
+          ;; explicitly rather than relying on the vm.globals fallback
+          ;; (the #1831 hazard this file's header documents below).
+          (only (srfi 13) string-prefix? string-index))
   (export define-record-type)
   (begin
 
@@ -246,14 +250,20 @@
     ;; Per-own-field expansion-time data, walked once. Returns three
     ;; values: the runtime names (one per field, deduped), the own alist
     ;; keyed by FULL spelling (for in-form, bound-identifier=-style
-    ;; matching; field-name key before accessor-name key per SRFI 150's
-    ;; precedence rule), and the property alist keyed by STRIPPED
-    ;; spelling (for a child's cross-form matching). `base` is the number
-    ;; of inherited fields, so own field i sits at absolute index base+i.
+    ;; matching), and the property alist keyed by STRIPPED spelling (for a
+    ;; child's cross-form matching). In both alists every field-name key
+    ;; precedes every accessor-name key (SRFI 150: a label that is both a
+    ;; field name and an accessor name resolves to the field, whatever the
+    ;; two fields' relative order), preserving field order within each
+    ;; group. `base` is the number of inherited fields, so own field i
+    ;; sits at absolute index base+i.
     (define (own-field-data specs base)
-      (let loop ((specs specs) (i 0) (used '()) (rnames '()) (okeys '()) (oprops '()))
+      (let loop ((specs specs) (i 0) (used '()) (rnames '())
+                 (fkeys '()) (akeys '()) (fprops '()) (aprops '()))
         (if (null? specs)
-            (values (reverse rnames) (reverse okeys) (reverse oprops))
+            (values (reverse rnames)
+                    (append (reverse fkeys) (reverse akeys))
+                    (append (reverse fprops) (reverse aprops)))
             (let* ((fs (car specs))
                    (abs (+ base i))
                    (cand (if (symbol? (car fs))
@@ -267,17 +277,18 @@
                     (+ i 1)
                     (cons nm used)
                     (cons nm rnames)
-                    (cons (cons field-key abs) (cons (cons acc-key abs) okeys))
-                    (cons (cons (hygiene-strip-key field-key) abs)
-                          (cons (cons (hygiene-strip-key acc-key) abs) oprops)))))))
+                    (cons (cons field-key abs) fkeys)
+                    (cons (cons acc-key abs) akeys)
+                    (cons (cons (hygiene-strip-key field-key) abs) fprops)
+                    (cons (cons (hygiene-strip-key acc-key) abs) aprops))))))
 
     ;; Resolve one constructor-spec entry to its absolute index: own
     ;; fields first (full spelling), then inherited fields (stripped
     ;; spelling against the parent's stored property). The precedence
     ;; rule -- a name that is both a field name and an accessor name
-    ;; resolves to the FIELD -- falls out of the alists' ordering:
-    ;; within each field's pair of keys the field name comes first, and
-    ;; the entries are in field order.
+    ;; resolves to the FIELD -- falls out of the alists' ordering: every
+    ;; field-name key precedes every accessor-name key, in field order
+    ;; (see own-field-data).
     (define (field-alist-ref key alist)
       (cond
         ((null? alist) #f)
--- a/lib/srfi/150.sld
+++ b/lib/srfi/150.sld
@@ -75,39 +75,75 @@
 ;;;     211's `capture-lookup` convention; `em-syntax-rules`/plain
 ;;;     syntax-rules transformers have no way to receive a procedure), so
 ;;;     define-record-type itself is an er-macro-transformer, not an
-;;;     em-syntax-rules macro. This turns out to simplify the hygienic
-;;;     comparison too: since this engine represents every identifier as
-;;;     a plain, hygiene-renamed symbol (see lib/srfi/211/explicit-
-;;;     renaming.sld's own header), and SRFI 213's property storage keeps
-;;;     a stored value exactly as given (it is not itself macro-expanded
-;;;     or re-processed), a stored field-name/accessor-name symbol was
-;;;     believed to retain its full hygienic spelling (rename prefix
-;;;     included) across the definition/lookup boundary, so that plain
-;;;     `equal?` on that raw symbol would already implement both
-;;;     bound-identifier=? and free-identifier=? with no
-;;;     em-bound-identifier=?-style machinery at all.
+;;;     em-syntax-rules macro.
 ;;;
-;;;     *** THAT IS FALSE -- see kaappi#2051. *** The rename does NOT
-;;;     survive the boundary, because this library carries field names
-;;;     across it inside `quote`, and quoting strips the hygiene prefix:
-;;;     (eq? '__hyg_2_a 'a) is #t. Expansion is correct (`kaappi expand`
-;;;     shows the two names properly distinguished), and both then strip
-;;;     to the same bare symbol before the runtime by-name lookup below
-;;;     ever sees them, so two hygienically-distinct fields collapse into
-;;;     one. This is what all four KNOWN FAILURES in
-;;;     tests/scheme/srfi/srfi150.scm are; the claim above was previously
-;;;     recorded as "verified against the exact scenario each SRFI 150
-;;;     hygiene test group requires", which is precisely the scenario
-;;;     that fails. A fix has to stop round-tripping hygienic identity
-;;;     through `quote` -- resolve field names entirely at expansion
-;;;     time, or key the property table on something that is not a quoted
-;;;     symbol. The rename-by-spelling representation described here is
-;;;     otherwise accurate, and is specific to this
-;;;     engine's rename-by-spelling hygiene representation, noted as a
-;;;     `(srfi 213)` conformance property already ("nominal... matching
-;;;     this symbol-based expander's identity model") -- a portable
-;;;     implementation on a syntax-object-based Scheme could not assume
-;;;     this and would need real bound-identifier=?/free-identifier=?.
+;;; The hygienic comparison and the constructor's field-name resolution
+;;; BOTH happen entirely at macro-expansion time, against the already-
+;;; hygiene-renamed form the expander hands the transformer. Nothing
+;;; hygienic is carried into runtime data at all, which is the point of
+;;; the redesign (kaappi#2051):
+;;;
+;;;   * Field identity is resolved to a numeric ABSOLUTE index into the
+;;;     full (inherited-then-own) field layout while the transformer still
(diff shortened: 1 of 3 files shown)


lib/srfi/150.sld
```diff
@@ -133,71 +169,131 @@
...
+                    (cons (cons field-key abs) (cons (cons acc-key abs) okeys))
+                    (cons (cons (hygiene-strip-key field-key) abs)
+                          (cons (cons (hygiene-strip-key acc-key) abs) oprops)))))))
+
+    ;; Resolve one constructor-spec entry to its absolute index: own
+    ;; fields first (full spelling), then inherited fields (stripped
+    ;; spelling against the parent's stored property). The precedence
+    ;; rule -- a name that is both a field name and an accessor name
+    ;; resolves to the FIELD -- falls out of the alists' ordering:
+    ;; within each field's pair of keys the field name comes first, and
```
baijum: The stated mechanism doesn't match what's built. `own-field-data` prepends `field-key` then `acc-key` per iteration and the whole list is `(reverse ...)`d, so each field's pair ends up **accessor-key before field-key** in `own-keys` — the opposite of "the field name comes first." It happens not to matter for results, because both keys of one field map to the *same* absolute index, so `field-alist-ref` returns the same index either way; new and old resolution are behaviorally identical. But the comment (here and at line 249) explains the precedence via an ordering that isn't actually present.

More substantively, the SRFI-150 rule this claims to guarantee — "a label that is both a field name and an accessor name resolves to the field" — is only honored when the coinciding field sits at a *lower* index than the accessor. When the field name belongs to a *later* field than the matching accessor name, resolution picks the accessor's (earlier) index:

```scheme
(define-record-type <prec>
  (mk-prec bar)   ; `bar` is field1's name AND field0's accessor
  prec?
  (x bar)         ; field0, accessor `bar`
  (bar get-bar))  ; field1, field name `bar`
(get-bar (mk-prec 99)) ; => unspecified, but spec says `bar` is the field => 99
```

This is pre-existing (the `main` code resolves it the same way via `field-alist-ref` checking field-then-accessor per entry in field order), so it's not introduced here and not a blocker. But since the new comment newly asserts the precedence "falls out of the alists' ordering," it'd be worth either (a) softening the comment to reflect that precedence is really field-index order (both keys share an index, so field-vs-accessor within a pair is moot), or (b) if you want true conformance, matching all field-name keys before any accessor-name key across the whole layout — and adding this shape to the regression controls.

lib/srfi/150.sld
```diff
@@ -133,71 +169,131 @@
...
+    ;; hygiene-strip-name mirrors types.stripHygienicPrefix: remove a
+    ;; `__hyg_<n>_` gensym prefix (and a `__kaappi_defenv__<lib>\x1f`
+    ;; definition-environment prefix), looping, leaving the underlying
+    ;; name. A name that merely starts with one of the markers but has no
+    ;; separator (e.g. `__hyg_foo` or `__kaappi_defenv__foo`) is left
+    ;; untouched, exactly as the engine's own quote-time stripping does.
+    (define (hygiene-strip-name s)
+      (let loop ((s s))
+        (cond
+          ((and (string-prefix? "__hyg_" s)
```
baijum, line 215: `hygiene-strip-name` uses `string-prefix?` and `string-index`, which are exported **only** by `(srfi 13)` (see `primitives_string_ext.zig` specs: `.libs = LS.initOne(.srfi_13)`), but `(srfi 13)` isn't in this import list. They resolve today purely through the library-body `vm.globals` fallback — the exact non-tail-library-body path this file's own header (lines 148-160) documents as the #1831 hazard, and `hygiene-strip-name` calls them in non-tail (`cond`-test / operand) position. It works now that #1831 is fixed, but relying on the fallback for a genuine cross-library dependency is implicit and fragile. Suggest adding `(srfi 13)` to the import list so the dependency is explicit and doesn't hinge on the fallback (none of those names should collide with the existing imports).

baijum (commented):
Reviewed the redesign end-to-end. The core approach — resolving field identity to absolute layout indices at expansion time and never round-tripping a hygienic symbol through `quote` — is sound and correctly targets the root cause. I reproduced the fix locally: the reference suite plus the new controls all pass (32/32), and I ran extra adversarial cases (permuted/omitted named-ctor args, 3-level shadowing, cross-form referral by a parent's accessor name, mutable/boolean/char constant fields) — all behave correctly. The docs and test additions are thorough.

Two minor, non-blocking points inline. Neither is a regression versus `main`; the first is a comment-accuracy issue with a pre-existing (obscure) conformance edge behind it, the second is a latent-fragility nit.

Nice work overall.

baijum:
Thanks for the review — all three findings are addressed in `ae5c6452`:

1. **Field-name precedence (both reviewers, inline on `own-field-data`)** — fixed. `own-field-data` now accumulates field-name keys and accessor-name keys separately and emits every field-name key before every accessor-name key (in field order within each group), for both the full-spelling own alist and the stripped-spelling property alist. The discriminating reversed-order shape (field at a *higher* index than the coinciding accessor) is added as a regression test:
   ```scheme
   (define-record-type <prec>
     (mk-prec y)
     prec?
     (a y)        ; field 0: accessor `y`
     (y get-y))   ; field 1: field name `y`
   (get-y (mk-prec 99))  ; => 99 (was the unfilled field before the fix)
   ```
   Note on the intra-pair claim: with the reverse-then-append construction, each field's pair was actually already field-key-first within the pair — the defect was the *inter-field* interleaving (field *j*'s field name losing to field *i*'s accessor when i < j), which is what the split-accumulator version fixes.

2. **`(srfi 13)` import (both reviewers, inline on `hygiene-strip-name`)** — `string-prefix?`/`string-index` are now imported explicitly via `(only (srfi 13) string-prefix? string-index)` instead of resolving through the `vm.globals` fallback.

3. **C5 control (CodeRabbit, inline on `def-no-binding`)** — the control reused the spelling `a`, which the reference suite binds at top level just above it, so it never tested the "no pre-existing binding" claim. Both the template literal and the use-site spelling are now `nb`, which nothing in the file binds.

Verified: `tests/scheme/srfi/srfi150.scm` 33/33, the related SRFI suites (131/136/213/237/57/240) and the #1832/#1829 hygiene regressions pass, `tests/scheme/run-all.sh` (703 files + 1395 R7RS tests) and `zig build test` are green.

Issue #1855, opened 2026-07-30
GC root stack is not unwound on OutOfMemory during compile/expansion (canonical push/pop pattern)


baijum:
Surfaced by CodeRabbit review on PR #1853; splitting out the codebase-wide half that a spot-fix would misrepresent.

## The property

The documented canonical rooting pattern (`.claude/rules/gc-safety.md`, the "Safe" example):

```zig
var a = try gc.allocSomething(...);
gc.pushRoot(&a);
const result = try gc.allocOther(a, ...);
gc.popRoot();
```

leaks the pushed root if the *protected* call itself fails: the `try` unwinds before `popRoot`, and nothing anywhere resets `root_count` on eval/compile error recovery (verified: `root_count` is only ever touched by `pushRoot`/`popRoot`). After an OOM-unwind the root stack permanently holds a dead stack-frame address; the next collection dereferences it (usually reads as a garbage flonum under NaN-boxing, occasionally as a fake heap pointer → UB). A leaked entry also breaks LIFO pairing for every *registered* `defer popRoot()` above it on the unwind path, so pops start removing the wrong entries.

This is a property of essentially every push/pop site in the codebase, not of any one function. The three `expander_instantiate.zig` sites flagged in the PR #1853 review (the `synth` build in `instantiateEllipsis` and the per-iteration pair) are instances: every **user-reachable** `ExpandError` path there unwinds balanced — the exposure is OutOfMemory-only. (The one flagged site that *was* user-reachable — `compileLetSyntax`'s defer placement, reachable via a malformed later binding — was fixed directly in PR #1853 with a regression test.)

## Options

1. **Accept and document**: treat OOM during expansion/compilation as effectively unrecoverable for GC-rooting purposes; note it in gc-safety.md. Cheapest, matches today's de facto behavior.
2. **Boundary reset**: save `root_count` at the eval/compile entry points (REPL loop, `vm.eval`, library load) and restore it on error return. One mechanism, covers every site, no per-site churn. Probably the right shape — mirrors how `extra_roots` is truncated per compile scope.
3. **Per-site `errdefer` discipline**: hundreds of sites, and `defer`/`errdefer` near loops is exactly the LIFO footgun gc-safety.md warns about (the SRFI 147 lesson). Least attractive.

FailingAllocator can't reach these deep-pipeline sites for a direct regression test (see the documented deep-pipeline limitation), so option 2's test would assert `root_count` restoration at the boundary instead.

[...]

PR #2284, opened 2026-08-22
Add bounded-step, resumable execution entry point (Fixes #2283)

### LINKED ISSUE #2283: WASM/VM: bounded-step entry point for cooperative, interruptible execution
## Summary

Add a bounded-step, resumable entry point to the WASM (and interpreter) so a host can run Scheme code in chunks, handing control back periodically instead of blocking until the program completes. This enables an interruptible, streaming playground and is generally useful for any embedding that needs cooperative execution (REPL-over-WASM, interruptible LSP, etc.).

Motivating case (from kaappi.github.io#32): the browser playground can only run programs via a synchronous WASI `_start` that blocks until completion, so its only runaway guard is a hard 5 s timeout that `terminate()`s the Web Worker — killing legitimately long-running, constant-space programs and discarding any output already produced. Example that should be able to run indefinitely in bounded memory:

```scheme
((call/cc call/cc) (call/cc call/cc))
```

## The good news: the hard part already exists

The VM is already a cooperatively-resumable, safepoint-based machine, because the SRFI-18 scheduler and the GC need it (`src/vm.zig`, `src/vm_dispatch.zig`):

- **Dispatch-loop safepoint "between instructions"** with an explicit `.stopped` state.
- **Yield/rewind protocol**: `error.Yielded` + `maybeRewindRetry`, which rewinds `ip` to the start of the current instruction so it re-runs cleanly on resume.
- **`terminated` flag polled at the safepoint**; GC already stops-the-world by parking there.
- **`runUntil(target_frame_count, target_wind_count)`** already runs the VM to a bounded target and returns; the scheduler already drives the VM cooperatively.

All execution state (frames, wind stack, IP) lives in the VM struct, not on the host C stack across yields — kaappi reifies its own control stack for `call/cc`. That property is exactly what makes a step entry point clean, and makes Asyncify / JSPI unnecessary here (see the approach comparison in kaappi.github.io#32).

## Proposed shape

- Split **setup** (read/compile the program, initialize VM) from **run**, so the program compiles once and the run loop can be pumped.
- Export a bounded-step run entry, e.g. `kaappi_step(budget) -> { running | done | error }`, that runs at most `budget` instructions (or to the next safepoint) and returns, preserving VM state in linear memory for the next call.
- Expose a **stop** flag / poll so the host can request cooperative termination between chunks (piggyback on the existing `terminated` path).
- Ensure **stdout/stderr stream** as produced (so a host can drain output between chunks) rather than only being observable after `_start` returns.

The batch WASI `_start` path stays as-is for CLI/file execution; the step entry point is additive.

## Downstream consumer

- **kaappi.github.io#32** — playground front-end: replace the synchronous `wasi.start()` + 5 s hard-kill with a pump loop that drains streamed stdout, checks a Stop flag, and enforces a wall-clock budget. Blocked by this issue.

## References

- `src/vm.zig`, `src/vm_dispatch.zig` — safepoint, `error.Yielded` / `maybeRewindRetry`, `runUntil`, `terminated`
- Origin: kaappi.github.io#32, and r/KaappiScheme "Timeout in the playground"

baijum:
## Summary

Adds a bounded-step, resumable entry point so a host can run Scheme code in instruction-budgeted chunks — handing control back periodically — instead of blocking until the program completes.

Motivating case (kaappi.github.io#32): the browser playground could only run programs through a synchronous WASI `_start` that blocks until completion, so its only runaway guard was a hard 5 s `terminate()` on the Web Worker. That kills legitimately long-running, constant-space programs and discards output already produced. The canonical example that should run indefinitely in bounded memory:

```scheme
((call/cc call/cc) (call/cc call/cc))
```

The batch WASI `_start` path (`main.runFile`) is unchanged; stepping is purely additive.

## Approach

Reuses the machinery the SRFI-18 scheduler and GC already need — the dispatch-loop safepoint, `error.Yielded`, and frames that live in the VM struct rather than on the host C stack across a yield (kaappi reifies its own control stack for `call/cc`, which is why Asyncify/JSPI are unnecessary).

A step budget is one more thing the safepoint checks. When the outermost stepped loop reaches its deadline, it returns `error.Yielded` with `step_paused` set — *between* instructions, so every stack (frames, wind, handler) and the `ip` are consistent and no rewind is needed.

**The one safety invariant:** a pause fires only in the outermost stepped `runUntil`, guarded by `step_active`, which `runUntil` save/restores exactly as it does `dispatched_from_scheduler`. A nested `runUntil` (eval, a native higher-order driver's callback, a scheduler fiber slice, a file-backed library load) runs with `step_active == false` and cannot pause, so a mid-form pause never strands a half-finished native frame. `beginStep`/`resumeStep` arm stepping only when no scheduler exists, so a fibered program runs its scheduler slice to completion within a step rather than pausing mid-fiber (a deliberate, documented limitation, not a correctness gap).

## What's in it

- **VM** (`vm_calls.zig`): `beginStep`/`resumeStep`, sharing `prepareTopLevelFrame` and the success/error tails (the #1855 root-boundary reset and dynamic-wind after-thunk unwinding) with `execute`; the pause is intercepted *before* those tails so nothing is torn down while the form is merely suspended. Five VM fields carry the state (`vm.zig`); the safepoint gate and `step_active` save/restore live in `vm_dispatch.zig`.
- **Driver** (`vm_step.zig`): `Stepper` iterates top-level forms like `runFile` — quick forms and library declarations run to completion, ordinary forms are stepped — under one shared budget, with matching result echoing and error reporting, plus a cooperative stop flag wired through `vm.terminate_flag`.
- **WASM** (`wasm_step.zig`, built `rdynamic`): `kaappi_step_alloc` / `kaappi_step_setup` / `kaappi_step_run` / `kaappi_step_stop` / `kaappi_step_reset` — the C-ABI the playground drives instead of `_start`. stdout/stderr already stream through WASI `fd_write` as produced (fd 1/2 are unbuffered).

## Testing

- `src/tests_step.zig` (18 cases via the shared harness): a finite program finishes and its side effects persist; a long form pauses and resumes to the same result a batch run produces; the constant-space infinite program keeps stepping forever without completing; the stop flag ends a running program cleanly; a form error is reported but execution continues to the next form; and the direct `beginStep`/`resumeStep` API pauses/resumes. Loop bounds scale down under `-Dgc-stress=true` (kaappi#1401).
- Full unit suite green (`zig build test`, and `-Dgc-stress=true` for the step tests).
- R7RS suite: 1395 pass, 0 fail. Continuation Scheme tests pass.
- End-to-end: a Node WASI harness driving the `kaappi_step_*` exports confirms the finite program streams `42\n` and returns DONE, the infinite program stays RUNNING across 15 steps, the stop flag ends a runaway loop, and a runtime error returns ERROR with prior output already streamed.

## Downstream

- kaappi.github.io#32 — the playground can now replace the synchronous `wasi.start()` + 5 s hard-kill with a pump loop that drains streamed stdout, checks a Stop flag, and enforces a wall-clock budget.

Docs: `docs/dev/bounded-step.md`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -382,6 +382,15 @@ library and vendored code are excluded).
   lexically** — exactly inverted. Mutating shared state through a global is a
   live hazard (kaappi#1924), not a supported idiom.
 
+- **Bounded-step execution (kaappi#2283)** — `docs/dev/bounded-step.md`. A
+  resumable, instruction-budgeted entry point (`VM.beginStep`/`resumeStep`,
+  driven by `vm_step.Stepper`; exported to WASM as `kaappi_step_*`) so a host —
+  the browser playground — runs Scheme in chunks instead of blocking until it
+  completes. Reuses the same dispatch-loop safepoint and `error.Yielded` the
+  scheduler and GC need; a pause fires only in the outermost stepped `runUntil`
+  (guarded by `step_active`, save/restored like `dispatched_from_scheduler`), so
+  it never strands a nested native frame.
+
 ## Dependencies
 
 **isocline** (vendored in `vendor/isocline/`): MIT-licensed C library for REPL
--- a/build.zig
+++ b/build.zig
@@ -293,6 +293,11 @@ pub fn build(b: *std.Build) void {
     // before any depth guard calibrated for the native stack could fire
     // (kaappi#2107).
     wasm_exe.stack_size = 64 * 1024 * 1024;
+    // Export the bounded-step C-ABI (kaappi_step_*, kaappi#2283) from the
+    // module alongside `_start`, so the playground host can call them. A WASI
+    // executable otherwise exports only `_start` and `memory`; `rdynamic`
+    // tells wasm-ld to keep every `export fn` symbol.
+    wasm_exe.rdynamic = true;
     const wasm_install = b.addInstallArtifact(wasm_exe, .{});
     wasm_step.dependOn(&wasm_install.step);
 
--- a/docs/dev/README.md
+++ b/docs/dev/README.md
@@ -30,6 +30,7 @@ investigation produced analysis worth keeping.
 | [porting.md](porting.md) | Porting to a new OS or CPU architecture: the support matrix, where portability lives, the degradation ladder, staged checklists, what "supported" means |
 | [adding-features.md](adding-features.md) | Step-by-step guides for the most common extension tasks |
 | [fibers-and-reactor.md](fibers-and-reactor.md) | KEP-0001: the per-thread reactor, parking vs. driving in place, lazy non-blocking mode as the platform probe, write buffering |
+| [bounded-step.md](bounded-step.md) | kaappi#2283: the resumable, instruction-budgeted step entry point — the safepoint mechanism, the outermost-loop invariant, the `beginStep`/`resumeStep` and `kaappi_step_*` WASM APIs |
 | [srfi-implementation-notes.md](srfi-implementation-notes.md) | How each non-trivial SRFI is implemented: what needed engine changes, resolved spec ambiguities, bugs each port surfaced |
 | [thottam.md](thottam.md) | The package manager: commands, `kaappi.pkg`, auto-discovery, the ecosystem library layout |
 | [testing.md](testing.md) | The four test layers, how to run them, where new tests go |
--- a/docs/dev/bounded-step.md
+++ b/docs/dev/bounded-step.md
@@ -1,0 +1,124 @@
+# Bounded-step execution (kaappi#2283)
+
+A resumable, instruction-budgeted entry point so a host can run Scheme code in
+chunks — handing control back periodically — instead of blocking until the
+program completes. The browser playground is the motivating consumer
+(kaappi.github.io#32): a synchronous WASI `_start` can only be stopped by a hard
+`terminate()` on a wall-clock timeout, which kills legitimately long-running,
+constant-space programs and throws away output already produced. The canonical
+example that should run indefinitely in bounded memory:
+
+```scheme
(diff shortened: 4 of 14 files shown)


baijum:
Thanks — addressed the review in b1d1f929.

- **Root boundary (major, `vm_calls.zig`/`vm_step.zig`).** Confirmed: the Stepper pushed a redundant GC root for the compiled func and popped it after `beginStep`, so `beginStep` recorded the form's root-stack base one slot too high; a *resumed* form that raised then truncated to that stale depth, leaving a bogus root. The func is already kept alive by `gc.extra_roots` (`Compiler.init` leaves it there on compile success), so the manual root was both unnecessary and harmful. Dropped it — the recorded base now matches the real one. Added a regression test that pauses a form, raises on loop exit, then allocates in the next form; it fails under `-Dgc-stress` with the old boundary and passes now.

- **Zero budget (minor, `vm_step.zig`).** `step` now clamps the deadline with `@max(budget, 1024)`, so `kaappi_step_run(0)` makes progress instead of spinning. Sub-1024 budgets can't pause before the safepoint anyway. Added a budget-0 progress test.

- **Setup leak (minor, `wasm_step.zig`).** The stepper-allocation failure path now frees the source buffer `ownedSource` had taken ownership of. (Left the latent `setGCInstance`-points-at-freed-GC note as-is: no exported entry dereferences it before the next `kaappi_step_setup` re-points it, and clearing it in teardown would add a null path every allocation site would have to guard.)

- **Test root (minor, `tests_step.zig`).** Removed the redundant `func_val` root in the direct-API test for the same `extra_roots` reason.

Full unit suite green (normal and `-Dgc-stress`), R7RS 1395/0, and the Node WASI end-to-end harness still green.

PR #2287, opened 2026-08-23
thottam: strict SemVer tag parsing, npm-style ^/~ ranges, and constraint diagnostics

### LINKED ISSUE #2130: [Tooling] thottam accepts a non-version git tag as a release — v2.0.0.nightly-UNRELEASED parses as 2.0.0, and v1_0.0.0 as 10.0.0
**Category:** Tooling
**Spec reference:** SemVer 2.0.0 §2 — "A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative integers, and MUST NOT contain leading zeroes."
**Severity:** wrong-result

`thottam`'s `Semver.parse` (`src/thottam_semver.zig:8`) splits on `.`, takes the
first three fields, and hands each to `std.fmt.parseInt(u32, …, 10)`. Two
consequences, both reachable from an ordinary `git ls-remote --tags` sweep:

1. **A fourth (and fifth, …) dot-separated component is silently discarded**, so
   a tag that is not a version becomes a release candidate.
2. **Each component is parsed with Zig's integer-literal grammar**, not SemVer's.
   `parseInt` accepts a leading `+` and `_` digit separators, so `v1_0.0.0`
   parses as **10.0.0** — a tag that outranks `v2.0.0`.

Leading zeros are accepted too (`v01.02.03` → 1.2.3), against §2's explicit
prohibition.

`resolveVersion` (`src/thottam.zig:187`) applies the constraint to every tag the
remote advertises, so anything a repo happens to have tagged — a nightly marker,
a branch snapshot, a build-id tag — is a candidate release.

**Minimal reproduction** (fully offline; `KAAPPI_ORG` may be a directory of bare
repos, which is how the new suite runs this):

```console
$ git -C repo tag                    # exactly two tags
v1.0.0
v2.0.0.nightly-UNRELEASED

$ thottam install 'kaappi-gamma@>=1.0.0'
  Resolved >=1.0.0 -> v2.0.0.nightly-UNRELEASED
  Checking out v2.0.0.nightly-UNRELEASED...
```

and the underscore case, which is the sharper one because it *reorders* releases:

```console
$ git -C repo tag
v1_0.0.0
v2.0.0

$ thottam install 'kaappi-zeta@>=2.0.0'
  Resolved >=2.0.0 -> v1_0.0.0        # read as 10.0.0, so it beats v2.0.0
  Checking out v1_0.0.0...
```

**Discriminating control.** The adjacent malformed spellings are correctly
rejected, so this is not general leniency:

| tag | `Semver.parse` | correct? |
|---|---|---|
| `v2.0.0-rc1` | rejected | yes — a pre-release, excluded from a plain range |
| `v2.0.0nightly` | rejected | yes |
| `v2.0.0/4` | rejected | yes |
| `v1-0.0.0` | rejected | yes |
| `v0x10.0.0` | rejected | yes |
| **`v2.0.0.nightly-UNRELEASED`** | **2.0.0** | **no** |
| **`v1.2.3.4`** | **1.2.3** | **no** |
| **`v01.02.03`** | **1.2.3** | **no** |
| **`v1_0.0.0`** | **10.0.0** | **no** |
| **`v+1.2.3`** | **1.2.3** | **no** |

Two identical fixtures differing only in that one tag (`kaappi-odd` carries
`v2.0.0-rc1`, `kaappi-nightly` carries `v2.0.0.nightly-UNRELEASED`) resolve
`>=1.0.0` to `v1.0.0` and to the nightly respectively.

A second-order effect: because a discarded suffix makes two distinct tags compare
`.eq`, `resolveVersion`'s `Semver.order(sv, b) == .gt` keeps whichever the remote
listed first, so `v1.2.3` and `v1.2.3.4` are separated by `git ls-remote`'s ref
ordering rather than by anything semantic.

**Suggested shape of a fix:** reject a version with more than three components,
with a leading zero on any component, or with any byte outside `0-9` in a
component (i.e. validate before `parseInt`, or hand-roll the digit loop).
Rejecting is the safe direction — an unparsed tag is simply not a candidate.

**Verified:** fresh ReleaseSafe build at `94ebd5a0`, `zig build`; every command
above run against local bare repositories with `KAAPPI_HOME` and `KAAPPI_ORG`
pointed at throwaway directories (no network).
**Source location:** `src/thottam_semver.zig:8` (`Semver.parse`), consumed by
`src/thottam.zig:206` (`resolveVersion`'s tag loop).
**Disabled tests:** `src/tests_thottam.zig` ("Semver.parse rejects a tag that is
not a version"), `tests/scheme/thottam/thottam-lifecycle.sh` ("a four-component
tag is not treated as a release") — both marked `FAIL: #NNNN`.

### LINKED ISSUE #2131: [Tooling] thottam: an abbreviated ~1 means 1.0.x, not 1.x — an omitted version component is indistinguishable from an explicit 0
**Category:** Tooling
**Spec reference:** node-semver, "Ranges" — `~1` := `>=1.0.0 <2.0.0` ("allows
minor-level changes when only the major version is specified"); `^0.0` :=
`>=0.0.0 <0.1.0`. `^`/`~` are not in the SemVer spec at all — they are npm's,
and every ecosystem that spells them this way inherited npm's meanings, which
`CLAUDE.md`'s "`^` (compatible), `~` (patch-level)" is naming.
**Severity:** wrong-result (edge-case: only the abbreviated forms)

`Semver.parse` fills an omitted component with `0` and returns a plain
`{major, minor, patch}` with no record of how many components were actually
written. `Constraint.matches` then treats the filled-in zero as if the author
had typed it, so an abbreviated range collapses to the fully-spelled one.

Under npm's grammar the *number of components written* is what selects the
range width, which is exactly the information `Semver.parse` throws away:

| range | node-semver | thottam | agree? |
|---|---|---|:---:|
| `^1.2.3` | `>=1.2.3 <2.0.0` | `>=1.2.3 <2.0.0` | yes |
| `^0.2.3` | `>=0.2.3 <0.3.0` | `>=0.2.3 <0.3.0` | yes |
| `^0.0.3` | `>=0.0.3 <0.0.4` | exactly `0.0.3` | yes |
| `^1` | `>=1.0.0 <2.0.0` | `>=1.0.0 <2.0.0` | yes |
| `^0.2` | `>=0.2.0 <0.3.0` | `>=0.2.0 <0.3.0` | yes |
| **`^0.0`** | **`>=0.0.0 <0.1.0`** | **exactly `0.0.0`** | **no** |
| `~1.2.3` | `>=1.2.3 <1.3.0` | `>=1.2.3 <1.3.0` | yes |
| `~1.2` | `>=1.2.0 <1.3.0` | `>=1.2.0 <1.3.0` | yes |
| **`~1`** | **`>=1.0.0 <2.0.0`** | **`>=1.0.0 <1.1.0`** | **no** |

`~1` is the one that will actually bite: it is a common way to write "track the
1.x line", and here it silently means "track 1.0.x".

**Minimal reproduction** (offline, tags `v0.1.0 v0.2.0 v0.2.5 v1.0.0 v1.2.3
v1.9.0 v2.0.0`):

```console
$ thottam install 'kaappi-alpha@~1'
  Resolved ~1 -> v1.0.0        # node-semver would give v1.9.0

$ thottam install 'kaappi-alpha@~1.0.0'
  Resolved ~1.0.0 -> v1.0.0    # same answer for a DIFFERENT range
```

**Discriminating control:** `~1` and `~1.0.0` are different ranges in
node-semver and identical here, while `~1.2` and `~1.2.0` are correctly
identical in both — so the defect is precisely "an omitted component is
indistinguishable from an explicit `0`", not a general tilde bug. The caret
row confirms it from the other side: `^1` (omitted minor+patch) is *right*
only because filling with zero happens to give the same answer there, and
`^0.0` — where it does not — is wrong.

**Suggested shape of a fix:** carry the written-component count out of
`Semver.parse` (or parse a range into an explicit `[lower, upper)` pair at
`parseSingleConstraint` time, which also makes `^`/`~` uniform and removes the
three-way branch in `Constraint.matches`).

**Verified:** fresh ReleaseSafe build at `94ebd5a0`; resolutions run against a
local bare repository with `KAAPPI_HOME`/`KAAPPI_ORG` isolated (no network).
The `^` rows are additionally pinned as unit tests.
**Source location:** `src/thottam_semver.zig:8` (`Semver.parse` fills with 0),
`src/thottam_semver.zig:37-43` (`Constraint.matches` `.caret`/`.tilde` arms).
**Disabled tests:** `src/tests_thottam.zig` ("caret on an abbreviated range",
"tilde on an abbreviated range") — marked `FAIL: #NNNN`.

### LINKED ISSUE #2132: [Tooling] thottam reports a malformed version constraint as 'no version matching' — six distinct parse failures share one message
**Category:** Tooling
**Spec reference:** `CLAUDE.md` § "kaappi.pkg manifest" — "Version constraints
are supported: `depends: kaappi-net@">=0.2.0"` with operators `>=`, `>`, `<=`,
`<`, `^` (compatible), `~` (patch-level), and comma-separated ranges
(`>=1.0.0,<2.0.0`)."
**Severity:** edge-case (diagnostics)

`isConstraintSpec` decides *that* a version string is a constraint by looking at
one byte (`> < ^ ~`); `parseConstraints` then decides *whether it parses*, and
`resolveVersion` collapses both "the constraint is malformed" and "no tag
satisfies it" into a single `?[]const u8` return. The user gets one message for
both:

```console
$ thottam install 'kaappi-alpha@>= 1.0.0'
error: no version matching >= 1.0.0 for kaappi-alpha
```

There *are* seven tags in that repository and four of them satisfy `>=1.0.0`.
The range was rejected because node-semver allows whitespace between an operator
and its version and this parser does not — but nothing in the output says so, so
the natural next step is to go looking at the remote's tags.

Every one of these produces that same message, with no way to tell them apart:

| spec | why it fails | reported as |
|---|---|---|
| `>= 1.0.0` | space between operator and version (legal in node-semver) | no version matching |
| `>=1.0.0-rc1` | pre-release in the constraint | no version matching |
| `>=1.0.0,` | trailing empty part | no version matching |
| `>=>1.0.0` | two operators | no version matching |
| `>=` | operator with no version | no version matching |
| `>=1.0.0,<2.0.0,>0.5.0,<=1.9.0,>0.1.0` | a fifth part (the array is `[4]?Constraint`) | no version matching |
| `>=9.0.0` | genuinely unsatisfiable | no version matching |

The four-constraint ceiling is undocumented and silent; the fifth part makes
`parseConstraints` return `null` for the whole range.

**Discriminating control:** `>=1.0.0, <2.0.0` (space after the *comma*) parses
and resolves to `v1.9.0`, because `parseConstraints` trims each part with
`std.mem.trim(u8, part, " ")` — so whitespace is tolerated in one position and
fatal in the adjacent one, which is what makes the shared message actively
misleading rather than merely terse.

**Three smaller diagnostics in the same family:**

1. **A rejected package name is reported twice, the second time as a raw Zig
   error name.** `main` special-cases `GitFailed`/`BuildFailed`/`CopyFailed` and
   lets everything else escape, so `InvalidPackageName` reaches Zig's default
   `main` error printer:

   ```console
   $ thottam install ../../evil
   error: invalid package name '../../evil' (only alphanumeric, '-', '_' allowed)
   error: InvalidPackageName
   ```

   Exit status is 1, which is right; the second line is an implementation
   detail. Same for `thottam remove ../../evil`.

2. **`pkg@` (a trailing `@` with nothing after it) yields an empty version
   rather than none.** It prints `Installing pkg@...`, then fails inside
   `checkoutVersion`'s own `v.len == 0` guard. Omitting the `@` entirely
   installs the default branch, so the two should agree.

3. **An empty `build:` line in `kaappi.pkg` is a command, not an absence.** It
   prints `Building <pkg>...` and runs `/bin/sh -c ""`. Verified end to end
   with a manifest whose whole content is `depends:\nbuild:\n`.

**Suggested shape of a fix:** have `resolveVersion` distinguish "did not parse"
from "no tag matched" (two error cases, not one `null`), and name the offending
part; add `InvalidPackageName` to `main`'s handled set alongside the other three;
treat an empty `@` version and an empty `build:` value as absent.

**Verified:** fresh ReleaseSafe build at `94ebd5a0`; all commands run offline
against local bare repositories with `KAAPPI_HOME`/`KAAPPI_ORG` isolated.
**Source location:** `src/thottam.zig:415-435` (the `isConstraintSpec` /
`resolveVersion` pair), `src/thottam.zig:880-892` (`main`'s error filter),
`src/thottam_semver.zig:48` (`[4]?Constraint`), `src/thottam_state.zig:36`
(`parsePkgSpec`'s `@` split), `src/thottam_state.zig:65` (`build:`).
**Disabled tests:** `src/tests_thottam.zig` ("parseConstraints accepts four
parts and rejects a fifth", "parsePkgSpec on degenerate specs",
"parsePkgManifest on missing and empty values") — marked `FAIL: #NNNN`.

baijum:
Fixes #2130, fixes #2131, fixes #2132.

Three thottam version-resolution defects from the Phase 6E audit, fixed together because they all live in the same `Semver.parse` → constraint-matcher → `resolveVersion` path and share the disabled-test fixtures.

## #2130 — a non-version git tag became a release candidate

`Semver.parse` split on `.`, took the first three fields, and handed each to `std.fmt.parseInt` — so `v2.0.0.nightly-UNRELEASED` parsed as `2.0.0`, and `v1_0.0.0` (via Zig's integer-literal `_` separators) as `10.0.0`, outranking `v2.0.0`.

- Components are now parsed by a hand-rolled digit loop (digits only, no leading zeroes, u32-bounded).
- A fourth dot-separated component disqualifies the tag.
- Empty *intermediate* components (`1..3`) are rejected while genuinely omitted trailing components (`v1.2`) stay lenient.

## #2131 — abbreviated `^`/`~` ranges collapsed to their filled-in zeros

`Semver.parse` filled omitted components with `0` and discarded how much was written, so `~1` meant `~1.0.0` (`>=1.0.0 <1.1.0`) and `^0.0` meant exactly `0.0.0`.

- `Semver` now carries a `written` component count (1–3).
- `^` and `~` match against it per node-semver: `^1`/`~1` are `>=1.0.0 <2.0.0`, `^0.0` is the whole `0.0.x` line, `~1.2` is unchanged.

## #2132 — malformed constraints reported as "no version matching"

`resolveVersion` collapsed "didn't parse" and "no tag matched" into one `null`, and six distinct parse failures shared one message.

- `resolveVersion` returns a `ResolveOutcome` union; a malformed range now reports `invalid version constraint '…' for …: part N is not a valid constraint (<op><version>)` (and the undocumented 4-part ceiling is diagnosed as `too_many_parts`).
- Operator/version whitespace (`>= 1.0.0`) is accepted, matching node-semver.
- `InvalidPackageName` is handled in `main` (install and remove) so the raw `error: InvalidPackageName` line no longer leaks after the real message.
- `pkg@` (trailing `@`) is now an absent version, not an empty string that failed at checkout.
- An empty `build:` line is treated as absent, not a command run via `/bin/sh -c ""`.

## Tests

- Re-enabled every `FAIL: #N` assertion across `src/tests_thottam.zig` and `tests/scheme/thottam/thottam-lifecycle.sh`, plus new unit tests for the component-count semantics, `too_many_parts`, and part-index diagnostics, and new lifecycle checks for the underscore-tag reorder and malformed-vs-unmatched messaging.
- Docs (`docs/dev/thottam.md`) updated for the node-semver `^`/`~` abbreviations and the strict tag rules.

`zig build test` and the thottam lifecycle suite (81 checks) both pass.

### DIFF
--- a/src/thottam.zig
+++ b/src/thottam.zig
@@ -183,15 +183,19 @@ fn joinPath3(allocator: std.mem.Allocator, a: []const u8, b: []const u8, c: []co
     return result;
 }
 
-/// Outcome of constraint resolution. The three cases were collapsed into a
-/// single `null` before kaappi#2132, which reported "no version matching" for
-/// a malformed range — indistinguishable from an unsatisfiable one.
+/// Outcome of constraint resolution. The cases were collapsed into a single
+/// `null` before kaappi#2132, which reported "no version matching" for a
+/// malformed range and for a failed `git ls-remote` alike — indistinguishable
+/// from an unsatisfiable one.
 const ResolveOutcome = union(enum) {
     resolved: []const u8,
     /// The range parsed but no tag satisfies it.
     no_match,
     /// The range itself does not parse; carries where and why.
     invalid_constraint: semver.ConstraintParseError,
+    /// `git ls-remote --tags` itself failed (missing/private repo, no
+    /// network). Not the same as "no matching tag": the range may be fine.
+    git_failed,
 };
 
 fn resolveVersion(allocator: std.mem.Allocator, clone_url: []const u8, constraint_str: []const u8) ResolveOutcome {
@@ -199,7 +203,7 @@ fn resolveVersion(allocator: std.mem.Allocator, clone_url: []const u8, constrain
     const constraints = semver.parseConstraintsDiag(constraint_str, &diag) orelse
         return .{ .invalid_constraint = diag };
 
-    const output = runGitCapture(allocator, &.{ "ls-remote", "--tags", "--", clone_url }) catch return .no_match;
+    const output = runGitCapture(allocator, &.{ "ls-remote", "--tags", "--", clone_url }) catch return .git_failed;
     defer allocator.free(output);
 
     var best: ?Semver = null;
@@ -454,6 +458,13 @@ fn doInstall(
                         .too_many_parts => std.fmt.bufPrint(&buf, "invalid version constraint '{s}' for {s}: a comma-separated range has at most 4 parts\n", .{ v, pkg }) catch "invalid version constraint\n",
                     };
                     writeStderr(detail);
+                    return error.GitFailed;
+                },
+                .git_failed => {
+                    var buf: [320]u8 = undefined;
+                    printErrColor(Color.red, "error: ");
+                    const msg = std.fmt.bufPrint(&buf, "failed to list tags for {s} (cannot resolve {s})\n", .{ clone_url, v }) catch "failed to resolve version\n";
+                    writeStderr(msg);
                     return error.GitFailed;
                 },
             }
--- a/src/thottam.zig
+++ b/src/thottam.zig
@@ -25,7 +25,6 @@ const state = @import("thottam_state.zig");
 const tfs = @import("thottam_fs.zig");
 
 const Semver = semver.Semver;
-const parseConstraints = semver.parseConstraints;
 const matchesAllConstraints = semver.matchesAllConstraints;
 const isConstraintSpec = semver.isConstraintSpec;
 
@@ -184,10 +183,23 @@ fn joinPath3(allocator: std.mem.Allocator, a: []const u8, b: []const u8, c: []co
     return result;
 }
 
-fn resolveVersion(allocator: std.mem.Allocator, clone_url: []const u8, constraint_str: []const u8) ?[]const u8 {
-    const constraints = parseConstraints(constraint_str) orelse return null;
+/// Outcome of constraint resolution. The three cases were collapsed into a
+/// single `null` before kaappi#2132, which reported "no version matching" for
+/// a malformed range — indistinguishable from an unsatisfiable one.
+const ResolveOutcome = union(enum) {
+    resolved: []const u8,
+    /// The range parsed but no tag satisfies it.
+    no_match,
+    /// The range itself does not parse; carries where and why.
+    invalid_constraint: semver.ConstraintParseError,
+};
 
-    const output = runGitCapture(allocator, &.{ "ls-remote", "--tags", "--", clone_url }) catch return null;
+fn resolveVersion(allocator: std.mem.Allocator, clone_url: []const u8, constraint_str: []const u8) ResolveOutcome {
+    var diag: semver.ConstraintParseError = undefined;
+    const constraints = semver.parseConstraintsDiag(constraint_str, &diag) orelse
+        return .{ .invalid_constraint = diag };
+
+    const output = runGitCapture(allocator, &.{ "ls-remote", "--tags", "--", clone_url }) catch return .no_match;
     defer allocator.free(output);
 
     var best: ?Semver = null;
@@ -218,9 +230,10 @@ fn resolveVersion(allocator: std.mem.Allocator, clone_url: []const u8, constrain
     }
 
     if (best_tag) |bt| {
-        return allocator.dupe(u8, bt) catch return null;
+        const dup = allocator.dupe(u8, bt) catch return .no_match;
+        return .{ .resolved = dup };
     }
-    return null;
+    return .no_match;
 }
 
 fn getPkgSha(allocator: std.mem.Allocator, src_dir: []const u8, pkg: []const u8) ?[]const u8 {
@@ -417,19 +430,32 @@ fn doInstall(
             const clone_url = parsed.source orelse blk: {
                 break :blk std.fmt.allocPrint(allocator, "{s}/{s}.git", .{ config.org, pkg }) catch return error.OutOfMemory;
             };
-            if (resolveVersion(allocator, clone_url, v)) |resolved| {
-                writeStdout("  Resolved ");
-                writeStdout(v);
-                writeStdout(" -> ");
-                writeStdout(resolved);
-                writeStdout("\n");
-                install_version = resolved;
-            } else {
-                var buf: [256]u8 = undefined;
-                printErrColor(Color.red, "error: ");
-                const msg = std.fmt.bufPrint(&buf, "no version matching {s} for {s}\n", .{ v, pkg }) catch "no matching version\n";
-                writeStderr(msg);
-                return error.GitFailed;
+            switch (resolveVersion(allocator, clone_url, v)) {
+                .resolved => |resolved| {
+                    writeStdout("  Resolved ");
+                    writeStdout(v);
+                    writeStdout(" -> ");
+                    writeStdout(resolved);
+                    writeStdout("\n");
+                    install_version = resolved;
+                },
+                .no_match => {
+                    var buf: [256]u8 = undefined;
+                    printErrColor(Color.red, "error: ");
+                    const msg = std.fmt.bufPrint(&buf, "no version matching {s} for {s}\n", .{ v, pkg }) catch "no matching version\n";
+                    writeStderr(msg);
+                    return error.GitFailed;
+                },
+                .invalid_constraint => |diag| {
+                    var buf: [320]u8 = undefined;
+                    printErrColor(Color.red, "error: ");
+                    const detail = switch (diag.kind) {
+                        .bad_part => std.fmt.bufPrint(&buf, "invalid version constraint '{s}' for {s}: part {d} is not a valid constraint (<op><version>)\n", .{ v, pkg, diag.part_index + 1 }) catch "invalid version constraint\n",
+                        .too_many_parts => std.fmt.bufPrint(&buf, "invalid version constraint '{s}' for {s}: a comma-separated range has at most 4 parts\n", .{ v, pkg }) catch "invalid version constraint\n",
+                    };
+                    writeStderr(detail);
+                    return error.GitFailed;
+                },
             }
         }
     }
@@ -878,7 +904,10 @@ pub fn main(init: std.process.Init.Minimal) !void {
         var visited = std.StringHashMap(void).init(allocator);
         defer freeVisited(allocator, &visited);
         doInstall(allocator, config, spec, locked_mode, &visited) catch |err| {
-            if (err == error.GitFailed or err == error.BuildFailed or err == error.CopyFailed) std.process.exit(1);
+            // InvalidPackageName already printed its own message; exiting here
+            // keeps Zig's default handler from printing the raw error name as a
+            // second line (kaappi#2132).
+            if (err == error.GitFailed or err == error.BuildFailed or err == error.CopyFailed or err == error.InvalidPackageName) std.process.exit(1);
             return err;
         };
     } else if (std.mem.eql(u8, cmd, "remove")) {
@@ -887,7 +916,7 @@ pub fn main(init: std.process.Init.Minimal) !void {
             std.process.exit(1);
         };
         doRemove(allocator, config, pkg) catch |err| {
-            if (err == error.NotInstalled) std.process.exit(1);
+            if (err == error.NotInstalled or err == error.InvalidPackageName) std.process.exit(1);
             return err;
         };
     } else if (std.mem.eql(u8, cmd, "list")) {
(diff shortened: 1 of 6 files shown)


src/thottam.zig
```diff
@@ -184,10 +183,23 @@ fn joinPath3(allocator: std.mem.Allocator, a: []const u8, b: []const u8, c: []co
...
+    invalid_constraint: semver.ConstraintParseError,
+};
+
+fn resolveVersion(allocator: std.mem.Allocator, clone_url: []const u8, constraint_str: []const u8) ResolveOutcome {
+    var diag: semver.ConstraintParseError = undefined;
+    const constraints = semver.parseConstraintsDiag(constraint_str, &diag) orelse
+        return .{ .invalid_constraint = diag };
 
-    const output = runGitCapture(allocator, &.{ "ls-remote", "--tags", "--", clone_url }) catch return null;
+    const output = runGitCapture(allocator, &.{ "ls-remote", "--tags", "--", clone_url }) catch return .no_match;
```
baijum: #2132 is about no longer collapsing distinct failures into one "no version matching" message, but a `git ls-remote` failure (network down, missing/private repo, auth failure) is still folded into `.no_match` here — so it reaches the user as `no version matching <v> for <pkg>`, which sends them hunting through tags for a range that was fine while the real problem was the fetch. This is pre-existing behavior, but since this PR is reworking exactly this function's outcome taxonomy, it'd be a natural place to add a third `.git_failed` arm (or surface the underlying error) so an IO failure reads as an IO failure. Optional.

src/thottam_semver.zig
```diff
@@ -36,54 +75,110 @@ pub const Constraint = struct {
...
         const trimmed = std.mem.trim(u8, part, " ");
-        result[i] = parseSingleConstraint(trimmed) orelse return null;
+        result[i] = parseSingleConstraint(trimmed) orelse {
+            diag.* = .{ .kind = .bad_part, .part_index = i };
+            return null;
+        };
         i += 1;
     }
-    if (i == 0) return null;
+    if (i == 0) {
```
baijum: This `if (i == 0)` guard looks unreachable: `splitScalar` always yields at least one token, and for the only case that would leave `i == 0` (an empty/whitespace spec) the single loop iteration hits `parseSingleConstraint("")` → `null` and returns `bad_part`/index 0 first. So the loop can never exit with `i == 0`. Harmless defensive code, but worth a comment noting it's belt-and-suspenders, or dropping it. (The `parseConstraintsDiag("", ...)` test passes via the loop's `bad_part` path, not this branch.)

baijum (commented):
Reviewed all six files. The SemVer/constraint logic is correct end-to-end — I checked the `^`/`~` abbreviations against node-semver's Ranges grammar (`^1`, `~1`, `^0`, `^0.0`, `~0.2`, `~1.2` all line up), the strict component parser rejects Zig-integer-literal spellings (`+`, `_`, leading zeroes, overflow) without panicking, and the `ResolveOutcome` split cleanly separates malformed-constraint from no-match. Test coverage is thorough and the re-enabled `FAIL: #N` assertions are a nice audit trail.

Two minor observations inline — neither blocks merge.

src/thottam.zig
baijum: Addressed in c4b7066 — resolveVersion now returns a distinct `.git_failed` outcome when `git ls-remote` fails, and doInstall reports it as "failed to list tags for <url>". Added a lifecycle check (`kaappi-missing@>=1.0.0`) asserting it's a fetch error, not "no version matching".

src/thottam_semver.zig
baijum: Addressed in c4b7066 — dropped the unreachable `i == 0` guard and left a comment explaining that `splitScalar` always yields a token, so the empty-spec path returns via `parseSingleConstraint("") → bad_part` in the loop body.

PR #2288, opened 2026-08-23
thottam: verify the installation, enforce --locked provenance, tolerate CRLF

### LINKED ISSUE #2133: [Tooling] A CRLF thottam.lock makes verify print 'MISMATCH (locked: X, actual: X)' — the two reported SHAs are identical
**Category:** Tooling
**Spec reference:** `CLAUDE.md` § "Package manager (thottam)" — "The lockfile
(`~/.kaappi/thottam.lock`) records source URLs for provenance." A lockfile
exists to be committed and shared between machines; `thottam` ships on all 18
CI legs including Windows under Git Bash.
**Severity:** wrong-result

`getLockedEntry` splits the lockfile on `'\n'` and never strips a trailing
`'\r'`, so on a checkout normalised to CRLF the carriage return becomes the last
byte of the recorded SHA. `doVerify` then compares that against `git rev-parse
HEAD` and reports a mismatch — but it prints only the first 12 characters of
each, and the `\r` is well outside that window, so **the two values it shows as
different are identical**:

```console
$ printf 'kaappi-alpha %s\r\n' "$(git -C src/kaappi-alpha rev-parse HEAD)" > thottam.lock
$ thottam verify
Verifying installed packages against lockfile...
  MISMATCH: kaappi-alpha (locked: 75f253e118c0, actual: 75f253e118c0)
Verification failed.
$ echo $?
1
```

There is no way to act on that message. The obvious readings — "the checkout
moved", "someone force-pushed" — are all wrong, and the diff the user would
reach for shows nothing.

**Discriminating control:** byte-for-byte the same line with `\n` instead of
`\r\n` verifies clean and exits 0. The only difference is the line terminator.

The same unstripped `\r` reaches `--locked` mode, where it turns the restore
into a hard failure: `install_version` is set to `<sha>\r`, `checkoutVersion`
passes it to `git checkout`, and git rejects the ref.

`thottam` itself always writes `\n`, so this needs an external touch — which is
exactly the lockfile's stated job: a file you commit, and that a colleague's
`core.autocrlf=true` checkout, or a Windows editor, normalises on the way past.

**Suggested shape of a fix:** trim `\r` where the lockfile and `installed.txt`
are read (`getLockedEntry`, `isInstalled`, `updateLockfile`,
`removeFromLockfile`, `removeFromInstalled`, and `doVerify`'s own line walk) —
the manifest parser already does this, `parseField` trims `" \t\r"`, so the
codebase has the convention, just not on this path. Independently, the 12-char
truncation in the MISMATCH message should not be able to render two different
values identically.

**Verified:** fresh ReleaseSafe build at `94ebd5a0`, run offline against a local
bare repository with `KAAPPI_HOME` isolated. Reproduced on macOS; the CRLF
lockfile is written explicitly, so it does not depend on the host's line-ending
convention and will reproduce on any leg.
**Source location:** `src/thottam_state.zig:89` (`getLockedEntry`),
`src/thottam_state.zig:74` (`isInstalled`), `src/thottam.zig:713-739`
(`doVerify`'s line walk and its `@min(12, …)` display).
**Disabled test:** `src/tests_thottam.zig` ("lockfile reader tolerates CRLF line
endings") — marked `FAIL: #NNNN`.

### LINKED ISSUE #2135: [Tooling] thottam verify iterates the lockfile, not the installation — an empty lockfile reports 'All packages verified.' and exits 0
**Category:** Tooling
**Spec reference:** `thottam --help` — "verify: Check installs match the
lockfile"; `src/thottam.zig:710` prints "Verifying installed packages against
lockfile...".
**Severity:** wrong-result

`doVerify` iterates the **lockfile** and asks, for each line, whether the
corresponding checkout is at that SHA. It never consults `installed.txt`. So a
package that is installed but absent from the lockfile is not verified, not
reported, and not counted — and the run still says everything is fine:

```console
$ thottam install kaappi-alpha && thottam install kaappi-one
$ thottam verify
Verifying installed packages against lockfile...
  OK: kaappi-alpha (75f253e118c0)
  OK: kaappi-one (9dcab67aba64)
All packages verified.

$ grep -v '^kaappi-one ' thottam.lock > t && mv t thottam.lock   # drop one line
$ thottam verify
Verifying installed packages against lockfile...
  OK: kaappi-alpha (75f253e118c0)
All packages verified.
$ echo $?
0
```

`kaappi-one` is still installed, still in `installed.txt`, still listed by
`thottam list` — and now unverifiable by construction. Truncating the lockfile to
zero bytes generalises this: **every** package becomes unchecked and the command
reports success.

```console
$ : > thottam.lock
$ thottam verify
Verifying installed packages against lockfile...
All packages verified.
$ echo $?
0
```

A lockfile of binary garbage does the same, because a line with no `' '` hits
`orelse continue` (`src/thottam.zig:716`) and `ok` is never cleared.

**Discriminating control — the machinery works; only its iteration domain is
wrong.** With the entry present, a wrong SHA is caught exactly as intended:

```console
$ sed 's/^kaappi-one .*/kaappi-one 0000.../' thottam.lock > t && mv t thottam.lock
$ thottam verify ; echo $?
  MISMATCH: kaappi-one (locked: 000000000000, actual: 9dcab67aba64)
Verification failed.
1
```

And a *missing* lockfile is caught (`No lockfile found`, exit 1) — so the blind
spot is specific to a lockfile that exists and says less than it should, which is
the shape a bad merge, a partial write, or a hand-edit produces.

The same iteration choice makes `verify` blind to the filesystem half of an
install: after `thottam remove` deletes a library file another installed package
was also using (see the sibling issue on `.sld` ownership), `verify` still reports
`OK` for the broken package, because the SHA it checks is unaffected.

**Suggested shape of a fix:** iterate `installed.txt` and treat "no lockfile
entry for an installed package" as a failure, keeping the current per-entry SHA
check; skip nothing silently — a malformed lockfile line should fail the run
rather than be `continue`d past. Optionally also check that each package's
`.sld` files are actually present under `$KAAPPI_HOME/lib`, which is what
"installs match the lockfile" reads as promising.

**Verified:** fresh ReleaseSafe build at `94ebd5a0`; all commands run offline
against local bare repositories with `KAAPPI_HOME`/`KAAPPI_ORG` isolated.
**Source location:** `src/thottam.zig:703-748` (`doVerify` — reads
`config.lockfile`, never `config.installed`; `orelse continue` at 716).
**Disabled tests:** `tests/scheme/thottam/thottam-lifecycle.sh` ("verify fails
when an installed package has no lockfile entry", "verify fails on an empty
lockfile with packages installed") — marked `FAIL: #NNNN`.

### LINKED ISSUE #2137: [Tooling] thottam --locked checks the SHA but not the source URL, then overwrites the lockfile's recorded provenance with the one it was handed

baijum:
## Summary

Three related audit findings in the package manager (`thottam`), fixed together because they share the same files and the same root cause: thottam trusting a lockfile it did not write, and reading the wrong half of the install state.

- **Closes #2135** — `verify` iterated the lockfile, not `installed.txt`, so an installed-but-unlocked package (or an empty/binary-garbage lockfile) silently passed. `doVerify` now walks `installed.txt`; every installed package must have a lockfile entry at the SHA it is checked out at, and a malformed lockfile line fails the run instead of being skipped.
- **Closes #2137** — `--locked` checked only the SHA, cloned from whatever URL the invocation was handed, then overwrote the lockfile's recorded provenance. The lockfile entry is now read once: the recorded source is the clone URL, an explicit `::url` that disagrees is refused before cloning, and a `--locked` install never rewrites the recorded source.
- **Closes #2133** — a CRLF-normalised lockfile left a trailing `\r` in the SHA, so `verify` printed `MISMATCH (locked: X, actual: X)` and `--locked` handed `<sha>\r` to `git checkout`. Every reader of `thottam.lock` / `installed.txt` now strips the trailing `\r`, and the mismatch message prints full SHAs so it cannot render two different values identically.

## Changes

- `src/thottam.zig`: `doVerify` rewrite (iterate `installed.txt`, validate lockfile structure, full-SHA mismatch output); `--locked` source-URL enforcement in `doInstall`; `trimStateLine` helper; CRLF trimming in `doList`/`doUpdate`.
- `src/thottam_state.zig`: strip trailing `\r` in `getLockedEntry`, `isInstalled`, `updateLockfile`, `removeFromLockfile`, `removeFromInstalled`.
- `src/tests_thottam.zig`: re-enable the #2133 CRLF assertion.
- `tests/scheme/thottam/thottam-lifecycle.sh`: re-enable the #2135 and #2137 regression checks.

## Testing

- `zig build test`: 1795/1802 tests pass (7 skipped, 0 failed).
- `thottam-tests`: 72/72 pass.
- `bash tests/scheme/thottam/thottam-lifecycle.sh`: 89 passed, 0 failed.

### DIFF
--- a/src/thottam.zig
+++ b/src/thottam.zig
@@ -41,7 +41,6 @@ const parsePkgManifest = state.parsePkgManifest;
 const isInstalled = state.isInstalled;
 const LockEntry = state.LockEntry;
 const getLockedEntry = state.getLockedEntry;
-const getLockedSha = state.getLockedSha;
 const appendLockEntry = state.appendLockEntry;
 const updateLockfile = state.updateLockfile;
 const removeFromLockfile = state.removeFromLockfile;
@@ -101,6 +100,19 @@ fn fatal(msg: []const u8) noreturn {
     if (use_color) writeStderr(Color.reset);
     writeStderr("\n");
     std.process.exit(1);
+}
+
+/// Strip a trailing carriage return from a line read out of a state file
+/// (`thottam.lock` / `installed.txt`). These files exist to be committed and
+/// shared between machines, so a checkout that a colleague's
+/// `core.autocrlf=true` (or a Windows editor) normalised to CRLF must read
+/// back the same values as the `\n` that thottam itself writes (#2133).
+///
+/// Only `\r` is stripped, not surrounding whitespace: a trailing space is
+/// load-bearing — it is the delimiter between a package name and an (empty)
+/// SHA in a truncated line.
+pub fn trimStateLine(line: []const u8) []const u8 {
+    return std.mem.trimEnd(u8, line, "\r");
 }
 
 pub fn readFile(allocator: std.mem.Allocator, path: []const u8) ![]u8 {
@@ -400,6 +412,15 @@ fn doInstall(
     const pkg = parsed.name;
     var install_version = parsed.ver;
 
+    // --locked mode: the lockfile is authoritative for both the SHA and the
+    // clone URL. Both are read up front so the source is enforced before any
+    // clone happens, and the recorded provenance is never rewritten by the
+    // very operation that is checking it (#2137).
+    var locked_sha: ?[]const u8 = null;
+    var locked_source: ?[]const u8 = null;
+    defer if (locked_sha) |s| allocator.free(s);
+    defer if (locked_source) |s| allocator.free(s);
+
     if (!isValidPkgName(pkg)) {
         printErrColor(Color.red, "error: ");
         writeStderr("invalid package name '");
@@ -418,15 +439,33 @@ fn doInstall(
     }
 
     if (locked_mode) {
-        const locked_sha = getLockedSha(allocator, config.lockfile, pkg);
-        if (locked_sha == null) {
+        const entry = getLockedEntry(allocator, config.lockfile, pkg) orelse {
             var buf: [256]u8 = undefined;
             printErrColor(Color.red, "error: ");
             const msg = std.fmt.bufPrint(&buf, "{s} is not in the lockfile (running in --locked mode)\n", .{pkg}) catch "package not in lockfile\n";
             writeStderr(msg);
             std.process.exit(1);
+        };
+        locked_sha = entry.sha;
+        locked_source = entry.source;
+        install_version = entry.sha;
+
+        // A `::url` handed to a --locked install must agree with the source
+        // the lockfile recorded (or, absent one, the org default). Refuse
+        // before cloning so a fork or mirror that merely shares history
+        // cannot redirect a locked install (#2137).
+        if (parsed.source) |supplied| {
+            var url_buf: [512]u8 = undefined;
+            const expected = locked_source orelse
+                (std.fmt.bufPrint(&url_buf, "{s}/{s}.git", .{ config.org, pkg }) catch return error.OutOfMemory);
+            if (!std.mem.eql(u8, supplied, expected)) {
+                var msg_buf: [512]u8 = undefined;
+                printErrColor(Color.red, "error: ");
+                const msg = std.fmt.bufPrint(&msg_buf, "source URL mismatch for {s}: lockfile records '{s}', got '{s}' (running in --locked mode)\n", .{ pkg, expected, supplied }) catch "source URL mismatch\n";
+                writeStderr(msg);
+                std.process.exit(1);
+            }
         }
-        install_version = locked_sha;
     }
 
     if (install_version) |v| {
@@ -483,7 +522,14 @@ fn doInstall(
     const pkg_dir = try joinPath(allocator, config.src_dir, pkg);
     defer allocator.free(pkg_dir);
 
-    const clone_url = if (parsed.source) |s| s else blk: {
+    // In --locked mode the lockfile's recorded source is the clone URL (the
+    // org default when it recorded none); otherwise the URL handed to this
+    // invocation, or the org default.
+    const clone_url = if (locked_mode)
+        (locked_source orelse blk: {
+            break :blk std.fmt.bufPrint(&buf, "{s}/{s}.git", .{ config.org, pkg }) catch return error.OutOfMemory;
+        })
+    else if (parsed.source) |s| s else blk: {
         break :blk std.fmt.bufPrint(&buf, "{s}/{s}.git", .{ config.org, pkg }) catch return error.OutOfMemory;
     };
 
@@ -512,14 +558,11 @@ fn doInstall(
     printBuf(&buf, "  Resolved: {s}\n", .{resolved_sha});
 
     if (locked_mode) {
-        if (getLockedSha(allocator, config.lockfile, pkg)) |locked_sha| {
-            defer allocator.free(locked_sha);
-            if (!std.mem.eql(u8, resolved_sha, locked_sha)) {
-                printErrColor(Color.red, "error: ");
-                const msg = std.fmt.bufPrint(&buf, "SHA mismatch for {s} (locked: {s}, got: {s})\n", .{ pkg, locked_sha, resolved_sha }) catch "SHA mismatch\n";
-                writeStderr(msg);
-                std.process.exit(1);
-            }
+        if (!std.mem.eql(u8, resolved_sha, locked_sha.?)) {
+            printErrColor(Color.red, "error: ");
+            const msg = std.fmt.bufPrint(&buf, "SHA mismatch for {s} (locked: {s}, got: {s})\n", .{ pkg, locked_sha.?, resolved_sha }) catch "SHA mismatch\n";
+            writeStderr(msg);
+            std.process.exit(1);
         }
     }
 
@@ -552,7 +595,9 @@ fn doInstall(
     }
 
     try appendFile(allocator, config.installed, pkg);
-    try updateLockfile(allocator, config.lockfile, pkg, resolved_sha, parsed.source);
+    // A --locked install must not rewrite the lockfile's recorded source: the
+    // provenance it is checking must survive the check (#2137).
+    try updateLockfile(allocator, config.lockfile, pkg, resolved_sha, if (locked_mode) locked_source else parsed.source);
     writeStdout("  ");
     printColor(Color.green, pkg);
     printBuf(&buf, " installed (locked at {s})\n", .{resolved_sha});
@@ -611,7 +656,8 @@ fn doList(allocator: std.mem.Allocator, config: Config) !void {
     writeStdout("Installed packages:\n");
     var buf: [512]u8 = undefined;
     var lines = std.mem.splitScalar(u8, content, '\n');
-    while (lines.next()) |pkg| {
+    while (lines.next()) |line| {
+        const pkg = trimStateLine(line);
         if (pkg.len == 0) continue;
         const entry = getLockedEntry(allocator, config.lockfile, pkg);
         const sha_short = if (entry) |e| e.sha[0..@min(12, e.sha.len)] else "unknown";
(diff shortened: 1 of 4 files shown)


baijum (commented):
Reviewed the three audit fixes. The core logic is correct and the change is well-reasoned — walking `installed.txt` instead of the lockfile is the right inversion for #2135, reading the lockfile entry once up front for #2137 is clean, and centralizing the CRLF strip in `trimStateLine` (with the note about why a trailing space must survive) is exactly right. A few minor points below; none are blocking.

**Dead code:** `getLockedSha` in `src/thottam_state.zig` (lines 118–122) is now unused — its only caller was the removed import in `thottam.zig`. Since it's `pub`, the compiler won't flag it. Worth deleting so a future reader doesn't reach for the SHA-only accessor that the #2137 fix deliberately moved away from.

**Test-coverage nit:** the new `--locked` enforcement is tested on the *reject* path (fork URL) and on the no-`::url` restore path, but not on the *agree* path — supplying a `::url` that matches the recorded/default source should be accepted. A control asserting that would pin the boundary from both sides and guard against the enforcement becoming accidentally over-strict.

src/thottam_state.zig
```diff
@@ -98,8 +98,9 @@ pub fn getLockedEntry(allocator: std.mem.Allocator, lockfile_path: []const u8, p
...
     var lines = std.mem.splitScalar(u8, content, '\n');
     while (lines.next()) |line| {
-        if (std.mem.startsWith(u8, line, pkg) and line.len > pkg.len and line[pkg.len] == ' ') {
-            const rest = line[pkg.len + 1 ..];
+        const l = thottam.trimStateLine(line);
+        if (std.mem.startsWith(u8, l, pkg) and l.len > pkg.len and l[pkg.len] == ' ') {
+            const rest = l[pkg.len + 1 ..];
             if (std.mem.indexOfScalar(u8, rest, ' ')) |sp| {
                 return .{
                     .sha = allocator.dupe(u8, rest[0..sp]) catch return null,
```
baijum, line 106: Edge case: a lockfile line with a trailing space after the SHA (`pkg <sha> `) makes `rest` end in a space, so `indexOfScalar` finds it and `source` becomes `dupe("")` — a **non-null empty string**, not `null`.

Downstream in `doInstall`'s `--locked` path that empty source is truthy, so `clone_url` / `expected` both become `""`, and the install fails with a confusing `Failed to clone repository` (or enforces `::url` against an empty string). Note `doVerify`'s new MALFORMED pass does *not* catch this — it only looks at the first space, and `pkg <sha> ` has a valid first delimiter.

Consider treating an empty source column as absent: `.source = if (rest[sp+1..].len == 0) null else (allocator.dupe(...) catch null)`. Narrow (needs a hand-edited/corrupt lockfile), so low priority.

src/thottam.zig
```diff
@@ -738,35 +785,85 @@ fn doUpdate(allocator: std.mem.Allocator, config: Config, pkg: ?[]const u8) !voi
...
+        if (entry == null) {
+            writeStdout("  ");
+            printColor(Color.red, "UNLOCKED");
+            printBuf(&buf, ": {s} (no lockfile entry)\n", .{pkg});
+            ok = false;
+            continue;
+        }
+
+        const locked_sha = entry.?.sha;
         const current_sha = getPkgSha(allocator, config.src_dir, pkg);
```
baijum, line 863: Minor: `current_sha` (from `getPkgSha` → `runGitCapture`) is heap-allocated and never freed, so this leaks once per installed package. Pre-existing behavior, and `thottam` is a short-lived process so it's harmless in practice — but the verify loop is being rewritten here anyway, so a `defer if (current_sha) |s| allocator.free(s);` would close it cheaply.

PR #2289, opened 2026-08-23
thottam: fix version re-pins, ownership-aware removal, and state-file name validation

### LINKED ISSUE #2134: [Tooling] thottam version pinning is a dead end: install pkg@ver on an installed package is a silent no-op (exit 0), and update can never move a pin
**Category:** Tooling
**Spec reference:** `CLAUDE.md` § "Package manager (thottam)" —
`thottam install kaappi-web@v1.0.0  # pinned version` and
`thottam update  # pull + rebuild all`.
**Severity:** wrong-result

`thottam` has no concept of "the version this package is currently installed
at", and both ends of the pinning workflow break on that.

**1. `install pkg@<version>` on an already-installed package is a silent no-op
that exits 0.** `doInstall` short-circuits on `isInstalled` before it looks at
the requested version at all (`src/thottam.zig:396`, ahead of the version
handling at 415):

```console
$ thottam install kaappi-multi              # HEAD is v2.0.0
  kaappi-multi installed (locked at db55be379e19...)

$ thottam install kaappi-multi@v1.0.0
  kaappi-multi already installed
$ echo $?
0
$ grep -o 'define v "[^"]*"' ~/.kaappi/lib/kaappi/multi.sld
define v "2.0.0"                            # still v2.0.0
```

A provisioning script that pins a version and checks the exit status is told it
succeeded. Nothing anywhere reports that the pin was ignored — the lockfile is
unchanged, `verify` passes (it is comparing against the SHA that *is* installed),
and `list` shows the wrong version with no annotation. `--locked` behaves the
same way, so the mode whose whole purpose is reproducibility short-circuits
before its own lockfile check runs.

**2. `update` can never move a package off a pin.** A pinned install leaves the
package repo on a detached HEAD, and `doUpdate` runs `git pull --quiet`
unconditionally (`src/thottam.zig:643`), so git's own advice reaches the user
verbatim:

```console
$ thottam install kaappi-alpha@v1.0.0
$ thottam update kaappi-alpha
Updating kaappi-alpha...
You are not currently on a branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.

    git pull <remote> <branch>

  Failed to pull
$ echo $?
1
```

`git pull <remote> <branch>` is not advice a `thottam` user can act on — the
repo is thottam's, under `$KAAPPI_HOME/src`, and running git in it by hand puts
the checkout out of step with the lockfile.

This is not confined to the named-package form. `thottam update` with no
argument loops over every installed package and exits 1 with "Some packages
failed to update" if any fails, so **one pinned package makes the whole-tree
update fail** — and the ecosystem's documented pattern (`depends:
kaappi-net@">=0.2.0"`) resolves a constraint to a tag and checks it out, which
is exactly a pin.

**Discriminating control:** the same `thottam update kaappi-alpha` on an
*unpinned* install of the same package succeeds and prints
`updated (now at <sha>)`. The difference is only whether the previous install
carried a version.

**Suggested shape of a fix:** for (1), compare the requested version against
what is checked out before short-circuiting — re-checkout when they differ, or
refuse with a message naming both. For (2), detect the detached HEAD and either
say plainly that the package is pinned (and to which ref) or re-resolve the
original constraint, which is the operation the user meant.

**Verified:** fresh ReleaseSafe build at `94ebd5a0`; run offline against local
bare repositories with `KAAPPI_HOME`/`KAAPPI_ORG` isolated. The fixture gives
each tag its own commit, so the "still v2.0.0" reading is not an artefact of
several tags sharing a SHA.
**Source location:** `src/thottam.zig:396` (the `isInstalled` short-circuit,
ahead of version handling and of the `locked_mode` check at 403),
`src/thottam.zig:643` (`git pull`), `src/thottam.zig:690-699` (the all-packages
loop's failure aggregation).
**Disabled tests:** `tests/scheme/thottam/thottam-lifecycle.sh` ("installing at a
pin changes the recorded SHA", "updating a pinned package does not leak raw git
advice") — marked `FAIL: #NNNN`.

baijum:
Closes #2134, #2136, #2138, #2144 — four thottam audit findings from Phase 6E, in one PR since they all touch the same module (`src/thottam.zig`, `src/thottam_state.zig`) and the same state-file surface.

## #2134 — version pinning is a dead end

- **`install pkg@ver` on an already-installed package** now re-checkouts when the requested version differs from the checkout (rebuild, re-copy libs, re-record lockfile) instead of short-circuiting on `isInstalled` and exiting 0 while leaving the old version in place. In `--locked` mode the lockfile is the requested version, so a locked install also re-checkouts when the installed SHA differs. The constraint-resolution block moved *before* the already-installed check so `install pkg@">=1.0.0"` compares against the resolved tag too. `installed.txt` recording is now idempotent.
- **`update` on a pinned package** detects the detached HEAD (`git symbolic-ref -q HEAD`) and says plainly what the package is pinned to (`git describe --tags --exact-match`, falling back to the lockfile SHA) instead of surfacing git's raw "You are not currently on a branch" advice. The named form is a clean no-op (exit 0) and the all-packages form skips the pin, so **one pinned package no longer fails a whole-tree update**. To move a pin: `thottam install pkg@<version>` — which now works (above).

## #2136 — removal has no ownership record

- New state file **`~/.kaappi/thottam.files`**: one `<pkg> <relpath>` line per installed file, written at install time.
- **`remove`** unlinks only files no other installed package's manifest claims — removing `kaappi-two` no longer takes `lib/kaappi/shared.sld` away from a still-installed `kaappi-one`. The manifest is the source of truth, so removal also stops orphaning copies of files a package renamed upstream (legacy installs fall back to the old source-tree walk, still claim-guarded).
- **`install`** warns (to stderr, yellow) when it is about to overwrite a file another package owns, making the cross-package collision audible at the point it's introduced.
- Empty directory skeletons left behind by removal are pruned (`$KAAPPI_HOME/lib/kaappi/deep/deeper/` no longer survives a full removal).

## #2138 — kaappi.pkg `name:`/`source:` parsed but never read

Per the design decision in the issue, the inert fields are **deleted from the parser and the documented grammar** (Option 2): `parsePkgManifest` now reads only `depends:` and `build:`; `name:`, `version:`, `source:` and any other key are ignored by construction. `manifest.source` could never be the clone URL — the manifest is only readable after cloning — and the docs said it declared "where this package is hosted". `docs/dev/thottam.md` updated; third-party sourcing is `pkg::url` or `KAAPPI_ORG`.

## #2144 — list/verify/update build paths from unvalidated names

The `isValidPkgName` guard now holds for every consumer, not just install/remove:
- `update <name>` validates its argument at entry (same message as install/remove).
- `list` and `update` (all-packages form) skip lines whose name fails the guard.
- `verify` names invalid names in `installed.txt` and `thottam.lock` **MALFORMED** and fails the run (consistent with the existing #2135 structure checks) — the escaped repo outside `$KAAPPI_HOME` is never touched.

## Tests

- Zig unit suite: `thottam-tests` 77 pass (manifest tests rewritten for the two-field grammar; new `addToInstalled`, `updateFileManifest`, `removeFromFileManifest`, `fileClaimedBy` round-trips).
- `tests/scheme/thottam/thottam-lifecycle.sh`: the four disabled `FAIL: #NNNN` checks are re-enabled and extended (pin re-install records the pin's SHA + copies the pinned lib, same-pin no-op, pin move, pinned-update no-op names the ref, all-update with one pin, shared-file survival + last-claimant deletion + overwrite warning, empty-dir pruning, corrupted-state handling across list/update/verify, inert-manifest installs). **133 assertions, 0 failed**, offline against local bare repos.

## Follow-ups outside this repo

- `kaappi.github.io/docs/ecosystem/thottam.md` documents the same manifest grammar (`source:` field table row, manifest example) — needs the same update in the docs-site repo.
- Workspace-root `CLAUDE.md` § "kaappi.pkg manifest" quotes `source:` as meaningful — that file is outside this repo (workspace root is not a git repo), so it needs a manual edit.

### DIFF
--- a/src/thottam.zig
+++ b/src/thottam.zig
@@ -39,6 +39,10 @@ const parsePkgSpec = state.parsePkgSpec;
 const isValidPkgName = state.isValidPkgName;
 const parsePkgManifest = state.parsePkgManifest;
 const isInstalled = state.isInstalled;
+const addToInstalled = state.addToInstalled;
+const updateFileManifest = state.updateFileManifest;
+const removeFromFileManifest = state.removeFromFileManifest;
+const fileClaimedBy = state.fileClaimedBy;
 const LockEntry = state.LockEntry;
 const getLockedEntry = state.getLockedEntry;
 const appendLockEntry = state.appendLockEntry;
@@ -69,6 +73,7 @@ const Config = struct {
     src_dir: []const u8,
     installed: []const u8,
     lockfile: []const u8,
+    files: []const u8,
 };
 
 fn writeToFd(fd: platform.fd_t, bytes: []const u8) void {
@@ -149,7 +154,7 @@ pub fn writeFile(allocator: std.mem.Allocator, path: []const u8, content: []cons
     }
 }
 
-fn appendFile(allocator: std.mem.Allocator, path: []const u8, line: []const u8) !void {
+pub fn appendFile(allocator: std.mem.Allocator, path: []const u8, line: []const u8) !void {
     const path_z = try allocator.dupeZ(u8, path);
     defer allocator.free(path_z);
     const fd = platform.openAppend(path_z, 0o644) catch return error.CannotOpen;
@@ -267,10 +272,8 @@ fn getPkgManifest(allocator: std.mem.Allocator, src_dir: []const u8, pkg: []cons
     const content = readFile(allocator, pkg_file) catch return null;
     defer allocator.free(content);
     var m = parsePkgManifest(content);
-    m.name = if (m.name) |n| allocator.dupe(u8, n) catch null else null;
     m.depends = if (m.depends) |d| allocator.dupe(u8, d) catch null else null;
     m.build_cmd = if (m.build_cmd) |b| allocator.dupe(u8, b) catch null else null;
-    m.source = if (m.source) |s| allocator.dupe(u8, s) catch null else null;
     m.owned = true;
     return m;
 }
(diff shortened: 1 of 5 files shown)


src/thottam.zig
```diff
@@ -584,17 +709,38 @@ fn doInstall(
...
 
-    try appendFile(allocator, config.installed, pkg);
+    // Record the installed files so `remove` can tell what belongs to whom
+    // (issue #2136). Rewrites this package's own lines: a re-install at a
+    // different version must not leave stale entries behind.
+    var files_list: std.ArrayList([]const u8) = .empty;
+    defer files_list.deinit(allocator);
+    for (sld_files.items) |rel| files_list.append(allocator, rel) catch return error.OutOfMemory;
+    for (dylib_files.items) |name| files_list.append(allocator, name) catch return error.OutOfMemory;
+    try updateFileManifest(allocator, config.files, pkg, files_list.items);
```
baijum: `install` records the file-ownership manifest here, but `doUpdate` (the `git pull` path around lines 923-929) copies the refreshed lib tree without a matching `updateFileManifest` call. After any `update` the manifest is stale: a file gained on update isn't recorded, so `fileClaimedBy` can't see this package as a claimant, and a subsequent `remove` of a *different* package will delete a now-shared file — reintroducing #2136 via the update path. A file dropped on update also lingers on disk unrecorded. Consider extracting 713-741 into a helper both `install` and `update` call so copy and record never drift.

baijum (commented):
Reviewed the four fixes end-to-end. The parser deletion (#2138), the name-validation sweep (#2144), and the removal ownership record (#2136) all look correct and are well tested; the re-pin fall-through (#2134) is a clean fix and the `wanted_sha` option-injection guard mirrors `checkoutVersion` nicely. One real gap below plus a smaller follow-on, both around the same root cause: the `thottam.files` manifest is written by `install` but never refreshed by `update`.

**Primary — `update` does not maintain `thottam.files`, so the #2136 guarantee lapses after any pull.**

`doUpdate` copies the pulled lib tree via `installLibTree` / `copyDylibsFromPkg` but never calls `updateFileManifest`. Concretely: install kaappi-A and kaappi-B (no overlap), then an upstream release of kaappi-A adds `lib/kaappi/shared.sld` (already owned by B). `thottam update kaappi-A` copies `shared.sld` but records nothing, so the manifest still shows only B owning it. A later `thottam remove kaappi-B` then runs `fileClaimedBy(shared.sld, except=B)`, finds no other claimant, and unlinks the file out from under the now-dependent kaappi-A — exactly the bug #2136 set out to close, reachable via the update path. The lifecycle suite doesn't cover update-then-shared-remove, so it passes green. Fix is to run the same collect + `warnIfClaimed` + `updateFileManifest` after the update copy that install does (factoring the install block at 713-741 into a shared helper would avoid a third copy of it).

**Secondary — re-checkout (#2134) orphans files the new version dropped.** On a re-pin that moves to a version where upstream *deleted* a lib file, the old copy in `~/.kaappi/lib` is never unlinked, and `updateFileManifest` rewrites the package's lines to the new (smaller) set — so the stale file is now owned by nobody and the source-tree fallback in `removeInstalledFiles` won't find it either (the checkout no longer contains it). It becomes permanently un-removable. Lower impact than the primary since it needs an upstream deletion across a re-pin, but same shape: on-disk state and the manifest diverge because copy and record aren't kept in lockstep.

Neither blocks the core fixes — they're completeness gaps in keeping `thottam.files` authoritative. Everything else LGTM.

baijum:
All review feedback addressed in `e76d4c9a`:

**Primary (update doesn't maintain `thottam.files`)** — fixed as suggested. The install copy+record block is factored into a shared `syncInstalledFiles` used by **both** `install` and `update` (`doUpdate` now calls it in place of the bare `installLibTree`/`copyDylibsFromPkg` pair). It collects the new file set, warns on files another package claims, copies, then records. The exact scenario from the review (upstream release of A adds a file B owns → `update A` → `remove B` keeps it) is now a lifecycle regression and passes.

**Secondary (re-pin orphans files the new version dropped)** — fixed in the same helper. `syncInstalledFiles` diffs the package's previous manifest lines against the new file set and unlinks dropped files that nobody else claims (pruning empty parents), so a re-pin or update that deletes a lib file no longer leaves a permanently un-removable stale copy. A regression covers the update-drops-a-shared-path variant: the other package's copy survives, the record is dropped, and removing the updated package (then the last claimant) behaves correctly.

**CodeRabbit items** — `docs/dev/thottam.md` now documents that the newest copy of a shared file stays authoritative and removal does not restore the previous contents; the unused `sha_v100` is wired into the v1.0.0 pin control assertion.

Verification: lifecycle suite now **145 assertions, 0 failed** (was 133); unit suite 1723 pass / 7 skip; thottam tests 77/77.

PR #2290, opened 2026-08-23
thottam: resolve git through PATH instead of hardcoding /usr/bin/git (Fixes #2152)

### LINKED ISSUE #2152: [Tooling] thottam hardcodes /usr/bin/git — every git operation fails on FreeBSD, OpenBSD and NetBSD
**Category:** Tooling / portability
**Severity:** missing-feature (thottam is unusable on three supported platforms)

`thottam` hardcodes the absolute path `/usr/bin/git` on every non-Windows platform (`src/thottam_proc.zig:148`, in both `runGit` and `runGitCapture`):

```zig
argv.append(allocator, if (platform.is_windows) "git" else "/usr/bin/git")
```

That path exists on macOS and on the Linux distributions CI uses. It does **not** exist on any of the three BSDs Kaappi supports and ships binaries for:

| platform | where git actually installs |
|---|---|
| FreeBSD (ports/pkg) | `/usr/local/bin/git` |
| OpenBSD (`pkg_add`) | `/usr/local/bin/git` |
| NetBSD (pkgsrc) | `/usr/pkg/bin/git` |

So **every git-backed thottam operation fails on those platforms** — `install`, `update`, and the `ls-remote` version resolution. Since cloning is the first thing `install` does, the package manager is effectively non-functional there.

**Discriminating control, and how this was found.** Phase 6E's new lifecycle suite failed on all three BSD legs while passing on macOS and Linux. My first hypothesis was that git could not clone a local bare repository there, so I added a capability probe that does exactly that — create a bare repo, clone it — and skips if it fails. **The probe passed on OpenBSD and the suite still failed.** The probe invokes `git` through `PATH`; thottam invokes `/usr/bin/git`. That difference is the whole bug, and the failed hypothesis is what isolated it.

**A second, compounding defect:** `runGit` discards git's stderr entirely —

```zig
runGit(allocator, &.{ "clone", "--quiet", "--", url_copy, pkg_dir }) catch {
    printErrColor(Color.red, "  Failed to clone repository\n");
```

so the user (and CI) sees `Failed to clone repository` with no cause. A `FileNotFound` on the git binary and a genuine clone failure are indistinguishable. Three CI runs' worth of logs contain **no** git diagnostic at all, which is why this took a falsified hypothesis to find rather than a glance at a log. Worth fixing alongside: surfacing the spawn error would have made this self-evident.

**Suggested fix:** invoke `git` by name and let the OS search `PATH`, as the Windows branch already does. If an absolute path is wanted for robustness, resolve it once at startup by searching `PATH` (the same way `kaappi compile` already discovers a C compiler) rather than assuming a location. The comment on the Windows branch — *"no fixed install path; CreateProcessW searches PATH"* — states the assumption that turns out to be false for POSIX too.

**Verified:** source read at `94ebd5a0`; the probe-vs-thottam contrast is from CI runs 30716991774 and 30722942738. The BSD install paths are stated from the platforms' packaging conventions, **not** executed — I have no BSD box in this session, so treat that row as the inference and the `/usr/bin/git` hardcode plus the PATH-probe contrast as the verified part.

**Found by:** Systematic audit v2, Phase 6E.

baijum:
## Summary

`thottam` hardcoded `/usr/bin/git` on every non-Windows platform, which exists on macOS and CI's Linux images but on **none of the three supported BSDs** — FreeBSD and OpenBSD install git in `/usr/local/bin`, NetBSD in `/usr/pkg/bin`. Every git-backed operation (`install`, `update`, `ls-remote` version resolution) therefore failed there, leaving the package manager non-functional on platforms Kaappi ships binaries for.

### The fix

- **Resolve `git` through `PATH`** (`findGit`/`findInPath` in `src/thottam_proc.zig`), the same way `kaappi compile` discovers a C compiler (`native_compiler.zig`) and `test_selection` locates its git. The resolved absolute path is handed to `execve`, so the child never depends on PATH resolution. The Windows branch's assumption — "no fixed install path; CreateProcessW searches PATH" — turns out to hold for POSIX too.
- **A missing git is now a distinct `error.GitNotFound`** that `install`/`update` report with a cause (`git not found in PATH — thottam requires git to …`), instead of the old silent 127 surfacing as the unhelpful `Failed to clone repository`.
- **Stop swallowing the spawn failure**: `runPassthrough`'s child now prints `cannot execute <argv[0]>: <errno>` before exiting 127, so a `FileNotFound` on the git binary and a genuine clone failure are no longer indistinguishable in the logs — the second defect that hid the first across three CI runs.

### Tests

- New Zig unit test `findInPath resolves an executable through PATH, not a fixed location` — asserts the resolver finds a fixture `git` in a directory preceding `/usr/bin` (first-match-wins), and returns `null` for an absent name.
- Updated the existing `checkoutVersion` precondition probe from `/usr/bin/git` to the PATH-based `findGit`.
- `tests/scheme/thottam/thottam-lifecycle.sh`: removed the `/usr/bin/git` skip-gate; the suite now runs wherever `command -v git` succeeds (which is the regression test for the BSD legs). **145 assertions, 0 failed**.

## Checklist

- [x] `zig build test` passes (1801/1808, 7 skipped)
- [x] `bash tests/scheme/thottam/thottam-lifecycle.sh` passes (145/145)
- [x] Documentation updated (not applicable — no doc referenced the git binary path)

Closes #2152.

### DIFF
--- a/src/thottam.zig
+++ b/src/thottam.zig
@@ -744,9 +744,18 @@ fn doInstall(
         writeStdout("...\n");
         const url_copy = try allocator.dupe(u8, clone_url);
         defer allocator.free(url_copy);
-        runGit(allocator, &.{ "clone", "--quiet", "--", url_copy, pkg_dir }) catch {
-            printErrColor(Color.red, "  Failed to clone repository\n");
-            return error.GitFailed;
+        runGit(allocator, &.{ "clone", "--quiet", "--", url_copy, pkg_dir }) catch |err| switch (err) {
+            error.GitNotFound => {
+                // Distinguish a missing git binary from a clone failure: the
+                // old silent 127 made every /usr/bin/git-less platform report
+                // "Failed to clone repository" with no cause (#2152).
+                printErrColor(Color.red, "  git not found in PATH — thottam requires git to install packages\n");
+                return error.GitNotFound;
+            },
+            else => {
+                printErrColor(Color.red, "  Failed to clone repository\n");
+                return error.GitFailed;
+            },
         };
     }
 
@@ -960,9 +969,15 @@ fn doUpdate(allocator: std.mem.Allocator, config: Config, pkg: ?[]const u8) !voi
             return;
         }
 
-        runGit(allocator, &.{ "-C", pkg_dir, "pull", "--quiet" }) catch {
-            printErrColor(Color.red, "  Failed to pull\n");
-            return error.GitFailed;
+        runGit(allocator, &.{ "-C", pkg_dir, "pull", "--quiet" }) catch |err| switch (err) {
+            error.GitNotFound => {
+                printErrColor(Color.red, "  git not found in PATH — thottam requires git to update packages\n");
+                return error.GitNotFound;
+            },
+            else => {
+                printErrColor(Color.red, "  Failed to pull\n");
+                return error.GitFailed;
+            },
         };
 
         if (getPkgManifest(allocator, config.src_dir, p)) |manifest| {
@@ -1274,8 +1289,9 @@ pub fn main(init: std.process.Init.Minimal) !void {
         doInstall(allocator, config, spec, locked_mode, &visited) catch |err| {
             // InvalidPackageName already printed its own message; exiting here
             // keeps Zig's default handler from printing the raw error name as a
-            // second line (kaappi#2132).
-            if (err == error.GitFailed or err == error.BuildFailed or err == error.CopyFailed or err == error.InvalidPackageName) std.process.exit(1);
+            // second line (kaappi#2132). GitNotFound likewise prints its own
+            // message at the clone site (kaappi#2152).
+            if (err == error.GitFailed or err == error.BuildFailed or err == error.CopyFailed or err == error.InvalidPackageName or err == error.GitNotFound) std.process.exit(1);
             return err;
         };
     } else if (std.mem.eql(u8, cmd, "remove")) {
@@ -1291,7 +1307,7 @@ pub fn main(init: std.process.Init.Minimal) !void {
         try doList(allocator, config);
     } else if (std.mem.eql(u8, cmd, "update")) {
         doUpdate(allocator, config, sub_arg) catch |err| {
-            if (err == error.NotInstalled or err == error.GitFailed or err == error.CopyFailed) std.process.exit(1);
+            if (err == error.NotInstalled or err == error.GitFailed or err == error.CopyFailed or err == error.GitNotFound) std.process.exit(1);
             return err;
         };
     } else if (std.mem.eql(u8, cmd, "verify")) {
--- a/src/thottam.zig
+++ b/src/thottam.zig
@@ -213,6 +213,8 @@ const ResolveOutcome = union(enum) {
     /// `git ls-remote --tags` itself failed (missing/private repo, no
     /// network). Not the same as "no matching tag": the range may be fine.
     git_failed,
+    /// `git` is not on PATH — not a repo/network problem (#2152).
+    git_not_found,
 };
 
 fn resolveVersion(allocator: std.mem.Allocator, clone_url: []const u8, constraint_str: []const u8) ResolveOutcome {
@@ -220,7 +222,10 @@ fn resolveVersion(allocator: std.mem.Allocator, clone_url: []const u8, constrain
     const constraints = semver.parseConstraintsDiag(constraint_str, &diag) orelse
         return .{ .invalid_constraint = diag };
 
-    const output = runGitCapture(allocator, &.{ "ls-remote", "--tags", "--", clone_url }) catch return .git_failed;
+    const output = runGitCapture(allocator, &.{ "ls-remote", "--tags", "--", clone_url }) catch |err| switch (err) {
+        error.GitNotFound => return .git_not_found,
+        else => return .git_failed,
+    };
     defer allocator.free(output);
 
     var best: ?Semver = null;
@@ -536,6 +541,19 @@ fn printErrColor(comptime color: []const u8, text: []const u8) void {
     if (use_color) writeStderr(Color.reset);
 }
 
+/// The missing-git diagnostic, shared by every call site that can surface
+/// `error.GitNotFound`. One place to print it, so the distinction never folds
+/// back into a generic "clone/list/checkout failed" message the way it did
+/// before #2152. `verb` names the operation, e.g. "install packages". Returns
+/// the error so callers can `return missingGit(...)` in one statement.
+fn missingGit(verb: []const u8) error{GitNotFound} {
+    var buf: [96]u8 = undefined;
+    const msg = std.fmt.bufPrint(&buf, "  git not found in PATH — thottam requires git to {s}\n", .{verb}) catch
+        "  git not found in PATH\n";
+    printErrColor(Color.red, msg);
+    return error.GitNotFound;
+}
+
 /// Record `pkg` in the install cycle/dedup guard. Returns true if it was newly
 /// inserted, false if already present.
 ///
@@ -661,6 +679,7 @@ fn doInstall(
                     writeStderr(msg);
                     return error.GitFailed;
                 },
+                .git_not_found => return missingGit("install packages"),
             }
         }
     }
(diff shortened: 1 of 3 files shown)


src/thottam.zig
```diff
@@ -744,9 +744,18 @@ fn doInstall(
         writeStdout("...\n");
         const url_copy = try allocator.dupe(u8, clone_url);
         defer allocator.free(url_copy);
-        runGit(allocator, &.{ "clone", "--quiet", "--", url_copy, pkg_dir }) catch {
-            printErrColor(Color.red, "  Failed to clone repository\n");
-            return error.GitFailed;
+        runGit(allocator, &.{ "clone", "--quiet", "--", url_copy, pkg_dir }) catch |err| switch (err) {
+            error.GitNotFound => {
```
baijum: The `GitNotFound` distinction added here is reachable only for a **fresh, unversioned** `thottam install <pkg>` — the one path that clones before running any other git command. Every other common invocation reaches a git-missing state through an *earlier* sibling `runGit*` call that still collapses `GitNotFound` into the generic "real failure" message this PR is trying to retire:

- `thottam install <pkg>@">=1.0.0"` → `resolveVersion` (line 223) does `runGitCapture(...) catch return .git_failed`, so a missing git surfaces as **"failed to list tags for <url> (cannot resolve …)"** (line 660) — i.e. "your repo/network is broken", the precise misdiagnosis #2152 is about.
- `thottam install <pkg>@<tag>` on an already-installed pkg → `checkoutVersion(...) catch { … return error.GitFailed }` (line 765) → **"Failed to checkout version"**.
- `thottam update <pkg>` → the `symbolic-ref` probe at line 957 swallows `GitNotFound` into its `else |_|` "pinned; skipping" branch and returns *before* the guarded `pull` below — so the update path never shows the new message either.

So the friendly "git not found in PATH" line only fires in the narrowest case. Consider funnelling the distinction into `runGit`/`runGitCapture` themselves (they already return `GitNotFound`) — e.g. a single helper the call sites route their errors through, or have `resolveVersion`/the `symbolic-ref` probe re-raise `GitNotFound` instead of folding it into `.git_failed`/`null`. Otherwise the fix's diagnostic intent is mostly unobservable in practice.

src/thottam.zig
```diff
@@ -960,9 +969,15 @@ fn doUpdate(allocator: std.mem.Allocator, config: Config, pkg: ?[]const u8) !voi
             return;
         }
 
-        runGit(allocator, &.{ "-C", pkg_dir, "pull", "--quiet" }) catch {
-            printErrColor(Color.red, "  Failed to pull\n");
-            return error.GitFailed;
+        runGit(allocator, &.{ "-C", pkg_dir, "pull", "--quiet" }) catch |err| switch (err) {
+            error.GitNotFound => {
```
baijum: This `error.GitNotFound` arm is effectively dead code when git is genuinely missing: the `symbolic-ref` capture at line 957 runs first, and its `else |_|` branch treats *any* failure (including `GitNotFound`) as "not on a branch → pinned", printing "… is pinned at unknown; skipping" and returning before the `pull` here is ever reached. So `thottam update` on a git-less box reports a bogus pin, not a missing git.

src/thottam_proc.zig
```diff
@@ -130,6 +130,17 @@ pub fn runPassthrough(allocator: std.mem.Allocator, argv: []const []const u8, cw
             @ptrCast(argv_z.ptr),
             @ptrCast(std.c.environ),
         );
+        // execve only returns on failure. Passthrough inherits the child's
+        // stderr, so say why instead of exiting 127 silently: a missing or
+        // unexecutable git used to surface as "Failed to clone repository"
+        // with no cause, indistinguishable from a real clone failure (#2152).
+        var exec_err: [256]u8 = undefined;
+        const exec_msg = std.fmt.bufPrint(&exec_err, "thottam: cannot execute {s}: {s}\n", .{ argv_z[0].?, @tagName(platform.errno(-1)) }) catch {
```
baijum: Nice — but `runCapture`'s child (the `execve` at ~line 57-62) still does a bare `std.process.exit(127)` with no diagnostic. `runCapture` backs `runGitCapture`, which is what `ls-remote` version resolution uses, so the "stop swallowing the spawn failure" rationale applies there too. With `findGit` now gating the missing-binary case up front, the residual exposure is a git that resolves but won't exec (see the note on `findInPath`); when that happens via `runGitCapture` it's back to a silent 127. Worth mirroring the same `cannot execute …` message into `runCapture`'s child for symmetry.

src/thottam_proc.zig
```diff
@@ -141,11 +152,54 @@ pub fn runPassthrough(allocator: std.mem.Allocator, argv: []const []const u8, cw
...
+    var iter = std.mem.splitScalar(u8, path_str, platform.path_list_sep);
+    while (iter.next()) |dir| {
+        if (dir.len == 0) continue;
+        const full = std.fmt.allocPrint(allocator, "{s}/{s}{s}", .{ dir, name, platform.exe_suffix }) catch continue;
+        const full_z = allocator.dupeZ(u8, full) catch {
+            allocator.free(full);
+            continue;
+        };
+        allocator.free(full);
+        const fd = platform.openRead(full_z) catch {
```
baijum: `openRead` checks *readability*, not *executability* — on POSIX it succeeds on a non-executable regular file and even on a directory named `git`. So a readable-but-unexecutable `git` earlier in PATH shadows the real one later, and `execve` then fails with the generic clone/list error rather than falling through to the next PATH entry the way `execvp` would. This matches the `native_compiler.zig` / `test_selection.zig` precedents, so it's consistent rather than a regression — but since the whole point here is locating a *runnable* git, an `access(X_OK)` / mode check (there's `platform.statPath` returning mode) would make the resolver actually pick the executable, not just the first readable file.

Minor, same line: `allocator.free(full_z[0 .. full_z.len + 1])` (and the caller's `git[0 .. git.len + 1]`) — `Allocator.free` already accounts for the sentinel on a `[:0]` slice, so `allocator.free(full_z)` frees the correct len+1 bytes and matches the two cited precedents, which avoids having to document the `path[0 .. path.len + 1]` free convention callers must remember.

baijum (commented):
Reviewed the fix — the core change (resolve `git` through PATH, hand the absolute path to `execve`) is correct and the right shape, and it matches the `native_compiler.zig` / `test_selection.zig` precedents it cites. The BSD breakage is genuinely fixed and the new unit test pins the first-match-wins behaviour well.

My main concern is that the *diagnostic* half of the PR — the whole `error.GitNotFound` distinction — is only wired into 2 of the ~6 git call sites, and those two happen to be shadowed by earlier, still-unguarded git calls on the exact commands users run. So for a missing git, most invocations still print the old misleading message the PR set out to eliminate. Details inline. A couple of smaller notes on `runCapture` symmetry and the PATH probe. Nothing here blocks the BSD fix itself.

baijum:
Thanks for the thorough review — both the diagnostic-coverage gap and the executability point were right. Addressed in `5767ee10`:

**Windows test failure (real, not just style):** the `findInPath` test hardcoded `:`, so it split `PATH` wrong on Windows and failed with `TestUnexpectedResult`. Now built from `platform.path_list_sep`.

**Executability, not readability** (`findInPath`): it now requires an *executable regular file* via `access(X_OK)` + `statPath().is_file` (Windows: `is_file`, no X bit). A non-executable file or a directory named `git` no longer shadows a later real git. The test asserts the negative case (non-executable fixture → skipped) and that the resolved fixture actually runs.

**`GitNotFound` now routed through every call site** — this was the main gap. Previously only the fresh-`clone` and `pull` paths recognized it; `resolveVersion` folded it into `.git_failed`, `checkoutVersion` into a generic failure, and the update flow's `symbolic-ref` probe into the "pinned" branch (so `update` could exit 0 on a git-less box). Now:
- `ResolveOutcome` gains `git_not_found`; `install pkg@">=1.0.0"` reports the missing git.
- `checkoutVersion` re-raises it; `install pkg@tag` on an installed package reports it.
- `symbolic-ref` distinguishes `GitNotFound` from a detached HEAD; `update pkg` reports it (and the previously-dead `pull` arm is now reachable).
- A shared `missingGit()` helper prints the single message, so the distinction can't silently fold back.

**`runCapture` symmetry:** its child now saves the real stderr before `/dev/null`'ing it and shares the `execveOrReport` diagnostic with `runPassthrough`, so a git that resolves but won't exec is no longer a silent 127 through `ls-remote`.

Also switched the `free` idiom to `allocator.free(slice)` on the `[:0]` slices (Zig accounts for the sentinel), matching `native_compiler.zig`/`test_selection.zig`.

Verified locally: `zig build test` (1801/1808), `thottam-lifecycle.sh` (145/145), `zig fmt --check`, and cross-compiles for `x86_64-windows`, `freebsd`, `netbsd`, and `openbsd`. The Windows `thottam unit tests` step should now be green.

Issue #2067, opened 2026-08-01
[SRFI-166] 15 documented names are not exported (joined/dot, numeric/fitted, trimmed/lazy, make-state-variable, 7 state variables, 4 sub-library names) and (srfi 166 base) does not exist


baijum:
**Category:** SRFI
**Spec reference:** SRFI 166 §"Index" enumerates the library's names; §"State Variables" says they are "All exported by `(srfi 166 base)`".
**Severity:** missing-feature

Fifteen names the SRFI's own index lists are not exported by any of the five libraries, and `(srfi 166 base)` does not exist. Meanwhile `(import (srfi 166))` succeeds and `(cond-expand (srfi-166 ...))` answers yes, so portable code has no way to detect the gap — the same shape as the SRFI 14 finding in #1895.

**Procedures (4):**

| name | what it is |
|---|---|
| `joined/dot` | `(joined/dot mapper dot-mapper list [sep])` — formats a dotted list's tail with `dot-mapper`; the spec's example is `(show #f "(" (joined/dot displayed (lambda (x) (each ". " x)) '(1 2 . 3) " ") ")")` ⟹ `"(1 2 . 3)"` |
| `numeric/fitted` | `(numeric/fitted width n . args)` — emits hashes rather than a truncated number; `(with ((precision 2)) (numeric/fitted 4 12.345))` ⟹ `"#.##"` |
| `trimmed/lazy` | truncates as it generates, so it is "safe to use with an infinite amount of output, e.g. from `written-simply` on an infinite list" — the SRFI's only mechanism for formatting an unbounded datum |
| `make-state-variable` | `(make-state-variable name default [immutable])` — the extension point the whole "maximally expressive and extensible" rationale rests on |

**State variables (7):** `writer`, `substring/width`, `substring/preserve`, `decimal-align`, `word-separator?`, `ambiguous-is-wide?`, `pretty-environment`.

**Sub-library names (4):** `pretty-with-color` from `(srfi 166 pretty)`; `string-terminal-width/wide`, `substring-terminal-width/wide` and `substring-terminal-preserve` from `(srfi 166 unicode)`.

Three of these are load-bearing for other defects rather than merely absent: `terminal-aware` is *specified* as a `with` of `string-width`, `substring/width` and `substring/preserve` (#2066), `wrapped`'s tokenizer is specified in terms of `word-separator?` (#2065), and `trimmed`'s ellipsis handling is specified in terms of `substring/preserve` (#2062).

[...]

Issue #2066, opened 2026-08-01
[SRFI-166] (srfi 166 unicode) is a stub: substring-terminal-width returns an integer instead of a substring, string-terminal-width is string-length, terminal-aware is a no-op


baijum:
**Category:** SRFI
**Spec reference:** SRFI 166 §"Unicode".
`(string-terminal-width str)` — "returns the integer number of columns str would require in a terminal, according to the following rules: non-spacing characters (format control characters with the property Cf, or non-spacing marks with the property Mn) count as **0** columns; characters with the East Asian Wide (W) or East Asian Fullwidth (F) properties … count as **2** columns; … ANSI terminal control sequences, as output by the color formatters above, count as **0** columns."
`(substring-terminal-width str from to)` — "**Returns the substring of str**, starting from the first index where the total string width exceeds from width, inclusive, to the first index exceeding to width, exclusive."
`(terminal-aware fmt ...)` — "Equivalent to `(fn (ambiguous-is-wide?) (with ((string-width …) (substring/width …) (substring/preserve …)) fmt ...))`."
**Severity:** wrong-result (one procedure returns the wrong *type*)

`lib/srfi/166/unicode.sld` is 29 lines and implements none of the width model:

```scheme
(define (string-terminal-width str) (string-length str))
(define (substring-terminal-width str from to) (- to from))
(define (terminal-aware . fmts) (apply each fmts))
```

1. **`substring-terminal-width` returns an integer, not a string.** Every one of the spec's eight worked examples expects a substring. This is a type error a caller cannot recover from, and it is silent.
2. **`string-terminal-width` is `string-length`.** East Asian Wide/Fullwidth characters count 1 instead of 2, combining marks and ZWJ count 1 instead of 0, and an ANSI colour sequence counts its full byte length instead of 0 — which is exactly the classic column-misalignment bug, reachable through the sibling `(srfi 166 color)` library.
3. **`terminal-aware` is `each`.** The spec's contrasted pair collapses to one answer.

Kaappi indexes strings by codepoint, so this is a live risk rather than a theoretical one, and the Unicode data needed is already in the tree (`src/unicode_tables.zig`, `tools/gen_srfi_charsets.py`).

| expression | spec | actual |
|---|---|---|
| `(string-terminal-width "日本語")` | `6` | `3` |
| `(string-terminal-width "ａｂｃ")` | `6` | `3` |
| `(string-terminal-width "e\x301;")` | `1` | `2` |
[...]

Issue #2065, opened 2026-08-01
[SRFI-166] (srfi 166 columnar): columnar and tabular do not align, wrapped hardcodes width 78, wrapped/char and justified are aliases, line-numbers is not a stream


baijum:
**Category:** SRFI
**Spec reference:** SRFI 166 §"Columnar Formatting".
`(columnar column ...)` — "The current line width (**from the width state variable**) is divided evenly among the columns (setting their state variables accordingly), and all but the last column are right-padded"; "columnar treats **raw strings as literals inserted into the given location on every line**, to be used as borders"; "**The degenerate case of no columns produces a single blank line.**"
`(tabular column ...)` — "Equivalent to columnar except that each column is padded at least to the minimum width required on any of its lines."
`(wrapped fmt ...)` — "lines are wrapped to fit in **the current width** as in the Unix fmt(1) command … words are tokenized by splitting on all characters which satisfy the predicate in the parameter `word-separator?`".
`(wrapped/char fmt ...)` — "splits simply on individual characters as the current width is reached on each line."
`(justified fmt ...)` — "Like wrapped except the lines are **full-justified**."
`(line-numbers [start])` — "formats an **infinite stream** of numbers (in the current radix) beginning with start."
**Severity:** wrong-result / missing-feature

Six of this library's eight exports do not implement their specification (`lib/srfi/166/columnar.sld`, 90 lines):

- **`columnar`** materialises each column to a whole string, splits it on newlines, and joins the row-th line of each with a single `" "`. It never reads `width`, never pads any column, and treats a raw string as a one-line column rather than a per-line border. Zero columns raises `KP3003: 'max': expected at least 1 arguments, got 0` from `(apply max '())` instead of producing a blank line. It also cannot stream: `(show p c)` per column is the opposite of the spec's "builds its output incrementally, interleaving calls to the column formatters until each has produced a line", which is what the `from-file`/`line-numbers` design depends on.
- **`tabular`** is `(apply columnar cols)` — no minimum-width computation.
- **A column whose output ends in a newline gains a spurious blank output row.** `%string-split-lines` on `"abc\ndef\n"` yields `("abc" "def" "")`, so `max-lines` is 3 and a third, empty row is emitted — visible as the trailing `" \n"` in every actual above. Independent of the alignment defects and probably the cheapest of them to fix.
- **`wrapped`** hardcodes `(w 78)` in its own `let` and never reads the `width` state variable or `word-separator?`.
[...]

Issue #2064, opened 2026-08-01
[SRFI-166] pretty is write: it never breaks a line at any width, and written-shared/pretty-shared do not label shared structure


baijum:
**Category:** SRFI
**Spec reference:** SRFI 166, `(pretty obj)` — "Pretty-prints obj. The result should be identical to `written` except possibly for differences in whitespace **to make the output resemble formatted source code**. Implementations should print vectors and data lists … in a tabular format when possible to reduce vertical space."
`(written-shared obj)` / `(pretty-shared obj)` — "using data labels for shared structures among all pairs and vectors, **analogous to write-shared**."
`(written-simply obj)` / `(pretty-simply obj)` — "doesn't handle shared structures, analogous to `write-simply`" / "without using any datum labels."
**Severity:** missing-feature

There is no pretty-printer and no shared-structure walker. Six exported names all reduce to R7RS `write`:

```scheme
;; lib/srfi/166.sld
(define written-shared written)
(define written-simply written)
;; lib/srfi/166/pretty.sld — the entire library body
(define (pretty obj) (lambda (st) (let ((p (open-output-string))) (write obj p) ((displayed (get-output-string p)) st))))
(define pretty-shared pretty)
(define pretty-simply pretty)
```

Consequences:

1. **`pretty` never emits a line break, at any width.** A 175-character datum stays on one line whether `width` is its default 78 or an explicit 20. The `width` state variable is not read. The SRFI's own worked example — `(show #t (columnar (pretty func) " ; " (justified doc)))`, which lays a 6-line pretty-printed `fold` beside a justified docstring — cannot be reproduced.
2. **`written-shared` and `pretty-shared` do not label non-cyclic sharing.** `write` labels only cycles, so the `write-shared` behaviour these two names exist to provide is unavailable.

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 166) (srfi 166 pretty))
(string-length
  (show #f (with (list (list width 20))
                 (pretty '(define (long-procedure-name alpha beta gamma)
                            (list alpha beta gamma alpha beta gamma))))))
;; expected: several lines, none over 20 columns   actual: one 88-column line

(let ((x (list 1 2))) (show #f (written-shared (list x x))))
;; expected: "(#0=(1 2) #0#)"   actual: "((1 2) (1 2))"
(let ((x (list 1 2))) (show #f (pretty-shared  (list x x))))
[...]

PR #2292, opened 2026-08-24
Rewrite SRFI 166 to fix the v2 audit findings

### LINKED ISSUE #2054: [SRFI-166] fn and with are procedures where the spec defines macros, fn binds nothing, and the output state variable has no slot
**Category:** SRFI
**Spec reference:** SRFI 166 §"Higher Order Formatters and State" — `(fn ((id state-var) ...) expr ... fmt)` and `(with ((state-var value) ...) fmt ...)` are given as *syntax*, with the worked examples `(show #f "column: " (fn (col) col))` ⟹ `"column: 8"` and `(show #f 1.5 " " (with ((precision 0)) 1.5))` ⟹ `"1.5 2"`. §"State Variables" defines `output` as "the underlying formatter used for writing strings."
**Severity:** missing-feature (blocks every documented use of the state-variable API)

`lib/srfi/166.sld` exports `fn`, `with` and `with!` as **procedures** taking an already-evaluated list of bindings. Written in the spec's own syntax, `(with ((precision 0)) 1.5)` evaluates `(precision 0)` — a call to the list `precision` — and raises `not a procedure`. Same for `(fn (col) col)`.

Worse, even under this implementation's own procedural convention `fn` **binds nothing**: it computes `bound` and threads an `env` alist through a loop, then discards both and runs `fmts` unchanged (`lib/srfi/166.sld:385-396`). There is no code path by which a state variable's value can reach a formatter body. That makes `col`, `row`, `width` and `port` unreadable from Scheme at all — they are write-only.

Separately, `%var->idx` (`lib/srfi/166.sld:121-136`) has no arm for `output`, and `%state-size` has no slot for it, so `(with (list (list output ...)) ...)` is silently discarded rather than erroring.

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 166))
(show #f 1.5 " " (with ((precision 0)) 1.5))   ;; expected: "1.5 2"   actual: raises "not a procedure"
(show #f "column: " (fn (col) col))            ;; expected: "column: 8" actual: raises "not a procedure"
(show #f (with (list (list output (lambda (s) (displayed "XXX")))) "abc"))
                                               ;; expected: "XXX"     actual: "abc"
```

**Discriminating control:** the same file's `with` *does* work when handed the procedural form it expects — `(show #f (with (list (list pad-char #\.)) (padded 4 "hi")))` ⟹ `"..hi"` — so the binding machinery exists; only the interface and `fn`'s body are wrong. And `(with (list (list radix 16)) (numeric 255))` ⟹ `"ff"` shows a state variable genuinely reaching a formatter, which is exactly what `fn` cannot do in the other direction.

**Verified:** fresh ReleaseSafe build at `08828884` (macOS aarch64), isolated `KAAPPI_HOME`, cache cleared.
**Source location:** `lib/srfi/166.sld:385-422` (`fn`, `with`, `with!`), `lib/srfi/166.sld:121-136` (`%var->idx`, no `output` arm)
**Disabled test:** `tests/scheme/srfi/srfi166-audit.scm` (5 assertions marked `;; FAIL:`). Two *enabled* assertions pin the current failure so the fix flips them: "fn: the spec's own (fn (col) col) syntax currently raises" and "with: the spec's own ((state-var value)) syntax currently raises".

**Found by:** Systematic audit v2, Phase 3.5

baijum:
Fixes #2054, #2056, #2058, #2059, #2061, #2062, #2063, #2064, #2065, #2066, #2067.

Complete reimplementation of the SRFI 166 (Monadic Formatting) library against
the spec, resolving all eleven Phase 3.5 audit findings in one change.

### Core (`lib/srfi/166/base.sld`, new)
- `fn` and `with` are now **macros** — `fn` binds state variables into a lexical
  environment, `with` dynamically binds them for its body and restores only the
  variables it bound (so `col`/`row` output position survives) (#2054, #2056).
- Extensible, first-class state variables via `make-state-variable`, with an
  `output` slot; `displayed` returns a formatter argument as-is (#2054, #2063).
- `numeric` honours radix+precision, `sign-rule`, `comma-rule`, `comma-sep`,
  `decimal-sep` and their state defaults; `numeric/comma` and `numeric/si` work
  (#2061).
- `escaped` drops the spurious delimiters, honours `esc-ch`/`renamer`;
  `maybe-escaped` quotes on embedded quote/escape (#2059).
- `tab-to` is a no-op on a tab stop and no longer divides by zero (#2058).
- padding/trimming/fitted honour `string-width` and `ellipsis` (#2062).
- `written-shared`/`pretty-shared` label non-cyclic sharing (#2064).

### Sub-libraries
- `(srfi 166 pretty)`: real line-breaking `pretty`, sharing-aware variants (#2064).
- `(srfi 166 columnar)`: columnar/tabular alignment, width/`word-separator?`-
  aware `wrapped`, character-wrapping `wrapped/char`, full-justified `justified`,
  streaming `line-numbers`, blank-line degenerate `columnar` (#2065).
- `(srfi 166 unicode)`: terminal-width model (wide=2, combining=0, ANSI=0),
  `substring-terminal-width` returning substrings, `terminal-aware` (#2066).
- `(srfi 166 base)` now exists and exports the 15 previously-missing names (#2067).

### Tests
`tests/scheme/srfi/srfi166-audit.scm` flips all 56 disabled `;; FAIL:` assertions
to enabled, pins the two macro-syntax cases as now-passing, and flips the 14
export-absence probes to presence probes.  `srfi166.scm` and `srfi166-sub.scm`
are updated for the corrected `escaped` and `line-numbers` behaviour.

All three SRFI-166 suites pass (244 + 35 + 11 assertions).


lib/srfi/166/columnar.sld
```diff
@@ -1,74 +1,229 @@
...
+          (else (error "columnar: invalid column argument" (car ls))))))
+
+    (define (%border-total cells)
+      (fold (lambda (c n) (if (eq? 'b (car c)) (+ n (string-length (cadr c))) n)) 0 cells))
+
+    (define (column-cell-width c)
+      (list-ref c 2))
+
+    (define (%pad-line line w align pad-char)
+      (let ((needed (- w (string-length line))))
```
baijum: `columnar`/`tabular` measure and pad cells with `string-length` (codepoint count), not the `string-width` state variable. `%pad-line` here computes `(- w (string-length line))`, and `tabular`'s min-width fold at line ~145 uses `(string-length line)` too. Both `columnar` (line 99) and `tabular` (line 133) capture `string-width` via `fn` but never pass it down, so the binding is dead.

Failure scenario: `(show #f (terminal-aware (columnar (displayed "日本\n") (displayed "x\n"))))` — each CJK cell is treated as width 2 instead of 4, so the first column is under-padded and the second column no longer aligns in a real terminal. Any custom `string-width` (or wide/zero-width/ANSI content) misaligns. Thread `string-width` through `%render-string`/`%render-row`/`%pad-line` and use it in place of `string-length`.

---

lib/srfi/166/base.sld
```diff
@@ -0,0 +1,797 @@
...
+    ;; Flatten obj to a string, labelling shared structure, with numbers
+    ;; formatted according to radix/precision.
+    (define (%write-flat obj shares radix precision)
+      (define (write-number n)
+        (let ((cell (and (not precision)
+                         (assv radix '((16 . "#x") (10 . "") (8 . "#o") (2 . "#b"))))))
+          (cond
+            ((and cell (eqv? radix 10)) (number->string n))
+            ((and cell (exact? n)) (string-append (cdr cell) (number->string n (car cell))))
+            (else (%number->string n 10 precision #f #f #f #f #f)))))
```
baijum, line 751: `write-number` only honours the `radix` state variable for radices in {2, 8, 10, 16}. For any other radix, and whenever `precision` is set (which makes `cell` `#f`), it falls through to this else branch which hardcodes radix `10`.

Failure scenario: `(show #f (with ((radix 3)) (written 5)))` yields `"5"` instead of a base-3 rendering, while `(numeric 5 3)` renders correctly — an inconsistency between `written` and `numeric` for the same `radix`. Likewise `(with ((radix 16) (precision 2)) (written 255))` prints `"255.00"` (decimal), whereas `(numeric 255 16 2)` gives `"ff.00"`. If decimal fallback for non-readable radices is intentional, at least pass the real `radix` (not `10`) so `numeric` and `written` agree where the radix is representable.

---

lib/srfi/166/columnar.sld
```diff
@@ -79,12 +234,15 @@
...
+    ;; `columnar` an 'infinite column is applied once per line, so this streams
+    ;; the next number on each call.
     (define (line-numbers . rest)
       (let ((start (if (null? rest) 1 (car rest))))
         (let ((n start))
           (lambda (st)
             (let ((s (number->string n)))
               (set! n (+ n 1))
-              ((each (padded 5 (displayed s)) (displayed " ")) st))))))
+              ((padded 5 (displayed s)) st))))))
```
baijum: `line-numbers` hardcodes `(padded 5 ...)` and `number->string` (radix 10), ignoring both the enclosing column's width/alignment and the `radix` state variable. The spec says it "just formats an infinite stream of numbers (in the current radix)" and lets `columnar` handle width/alignment. It passes the current tests only because the sample uses a column narrower than 5 (so the 5-wide pad is a no-op) and radix 10.

Failure scenario: `(columnar 6 'right 'infinite (line-numbers) " " (from-file f))` produces 5-wide left-padded numbers inside a 6-wide right-aligned column, so the numbers are misaligned by a column; `(with ((radix 16)) (columnar ... (line-numbers) ...))` still prints decimal. The comment also says "right-padded" but `padded` pads on the left. Format the number in the current radix and let the column pad/align it, rather than baking in width 5.

---

lib/srfi/166/pretty.sld
```diff
@@ -1,15 +1,79 @@
...
+             (display (show #f (written obj)) out))))
+        (get-output-string out)))
+
     (define (pretty obj)
-      (lambda (st)
-        (let ((p (open-output-string)))
-          (write obj p)
-          ((displayed (get-output-string p)) st))))
+      (fn (width string-width)
+        (displayed (%pp obj width string-width (lambda (o) (show #f (written o)))))))
```
baijum: `pretty` (and `pretty-shared`/`pretty-simply`) obtain the flat form via `(show #f (written o))`, which runs `written` in a brand-new show state with default `radix`/`precision`. The current dynamic state is lost.

Failure scenario: `(show #f (with ((radix 16)) (pretty 255)))` prints `"255"`, but `(show #f (with ((radix 16)) (written 255)))` prints `"#xff"`. The spec says `pretty`'s result "should be identical to written except possibly for differences in whitespace", so the two must agree on number formatting. `%break`'s atom arm (line ~60) similarly calls `(show #f (written obj))` with a fresh state.

---

lib/srfi/166/columnar.sld
```diff
@@ -79,12 +234,15 @@
                   (begin (close-input-port p) st)
```
baijum: `from-file` closes the input port only on this normal EOF path. If a formatter in the streaming loop raises (a downstream `output` override errors, or `show` is aborted mid-stream), `close-input-port` is never reached and the file descriptor leaks.

Failure scenario: streaming a file inside a `guard` that catches a mid-stream error leaves the port open; repeated over a long-running program this exhausts descriptors. Wrap the loop in `dynamic-wind`, or use `call-with-input-file`, so the port is closed on every exit path.

---

baijum (commented):
Code review of PR #2292 (SRFI 166 rewrite) at high effort. The reimplementation is thorough and the sharing/cycle detection, macro `fn`/`with`, numeric formatting, and terminal-width tables all check out against the audit suite. Five findings below, none blocking correctness of the covered tests but each a real defect in an uncovered path (mostly wide-character / state-propagation edges). I could not build in this environment (no `zig`), so these rest on static analysis against the spec and test suite.

---

lib/srfi/166/pretty.sld
```diff
@@ -1,15 +1,87 @@
...
+    ;; radix/precision so the result matches `written` (modulo whitespace).
+    ;;
+    ;; Shared or cyclic data is never broken: a manual break cannot preserve
+    ;; the datum labels, and iterating a cyclic pair would not terminate.  Such
+    ;; data is emitted flat (with its labels) even if it overflows the width.
+    (define (%pp obj width sw radix precision shares)
+      (let ((flat (%write-flat obj shares radix precision)))
+        (if (<= (sw flat) width)
+            flat
+            (if (positive? (hash-table-size (car (extract-shared-objects obj #f))))
```
baijum: Low severity, non-blocking — follow-up on the shared/cyclic fix.

This guard uses `(extract-shared-objects obj #f)` (all sharing, cyclic *and* acyclic) for all three entry points, but `pretty` passes `#t` (cyclic-only) as `shares` on line 71. So for `pretty`, the flat form only carries datum labels for genuine cycles — an acyclic DAG (a pair/vector reachable by two `eq?` paths, no cycle) renders fully expanded with **no** labels. `%break` could reproduce that safely (it re-renders the shared sub-structure at each occurrence, exactly as the flat form does — no labels to preserve), yet this guard sees the sharing and refuses to break, leaving a long overflowing line unformatted:

```scheme
(show #f (pretty (let ((sub (make-list 30 'x))) (list sub sub))))
;; one ~130-char line at width 78, never broken
```

Two consequences:
- The code comment on lines 11–13 ("emitted flat (with its labels)") is inaccurate for this case — `pretty` on acyclic sharing produces no labels.
- The block is stricter than it needs to be *for `pretty`*.

Suggestion: make the guard's cyclic-only? flag match the variant (cyclic-only for `pretty`, all-sharing for `pretty-shared`) — e.g. thread the same flag used to build `shares` — or gate on whether the produced `flat` actually contains a `#n=` label rather than re-detecting sharing. `pretty-shared`/`pretty-simply` are already correct with `#f`.

---

baijum (commented):
Re-reviewed the follow-up commit `0593f4e` ("Address SRFI 166 review findings"). All five findings from my previous review are correctly resolved, and the `wrapped/char` non-termination fix is sound. Verified in detail:

- **write-number radix + precision** (`base.sld`) — non-decimal readable radices now keep the `#x`/`#o`/`#b` prefix and drop precision; radix 10 keeps precision; unsupported radices fall back to decimal. Matches the spec and the new regression tests.
- **columnar/tabular string-width threading** — `%pad-line`, `%render-row`, and the tabular min-width fold all take the `string-width` state variable now; `terminal-aware` alignment tests pass.
- **line-numbers** — formats in the current radix and leaves width/alignment to `columnar`; the `fn` + closed-over `set!` correctly streams one number per application.
- **from-file** — `call-with-input-file` closes the port on every exit path.
- **justified** — subtracting the mandatory per-gap space from the budget makes lines land exactly on `width` (confirmed by the arithmetic and the new test).
- **pretty** — radix/precision now thread through both the flat and broken paths; cyclic data is left flat with labels rather than looping.

One new low-severity, non-blocking observation on the `pretty` shared-object guard (inline). Also a minor note: the newly added fractional column-width resolution in `%column-widths` (real widths in (0,1)) has no test exercising it, so its behavior isn't pinned — worth a case if fractional widths are a supported path.

Nothing here blocks the PR.

---

PR #2295, opened 2026-08-24
Make equal? recurse into record fields (structural record equality)

### LINKED ISSUE #2293: Make equal? recurse into record fields (structural record equality)
## Summary

Make `equal?` recurse into record fields, so two distinct instances of the same
record type whose fields are pairwise `equal?` compare as `#t`. Today Kaappi
compares record instances by **identity only** — the same behavior as `eq?` and
`eqv?` — so two structurally-identical records are `#f`.

This is a **feature request, not a conformance bug**: R7RS explicitly leaves
`equal?` on records unspecified (see analysis below). But every mainstream
implementation that *natively targets R7RS* recurses into fields, so Kaappi is
the odd one out, which surprises users porting code.

Prompted by a user report on r/KaappiScheme:
"Better if record types with equal? components were equal?"
https://www.reddit.com/r/KaappiScheme/comments/1vwodyg/better_if_record_types_with_equal_components_were/

## Reproduction

```scheme
(define-record-type foo (mkfoo x) foo? (x bar))

(equal? (mkfoo 1) (mkfoo 1))   ; Kaappi => #f   ; wanted => #t
(eqv?  (mkfoo 1) (mkfoo 1))    ; => #f  (correct, stays #f)
(let ((r (mkfoo 1))) (equal? r r)) ; => #t  (correct, unchanged)
```

(The original Reddit example reuses `foo` as both the type name and the
constructor name; that form happens to work on Kaappi and Gambit but is rejected
by Chibi and Guile. Unrelated to this issue — noted only so the example above
uses distinct names.)

## Current behavior — where it lives

`equal?` is registered at `src/primitives.zig:300` and implemented by `equalP`
(`src/primitives.zig:906`), which delegates to `deepEqualWithVisited`
(`src/primitives.zig:913`). That helper has explicit branches for flonums,
bignums, complex, rationals, pairs, strings, vectors, bytevectors, and numeric
vectors — but **no branch for `record_instance`** (`ObjectTag` 13,
`src/types.zig:178`; struct `types_record.RecordInstance`,
`src/types_record.zig:104`). So two records only pass the identity fast-path at
`src/primitives.zig:914` (`if (a == b) return true;`); otherwise control falls
through to the catch-all `return false;` at `src/primitives.zig:969`.

The only uses of `types.isRecordInstance` in `primitives.zig` are the record
accessors at lines 1104–1129, never the equality path — confirming records get
no structural treatment. `eqv?`/`eq?` are identity-only for records too, which
**is** spec-required and must stay unchanged.

## Spec analysis (docs/errata-corrected-r7rs.pdf)

- **§6.1 `equal?` (p. 32).** `equal?` recursively compares *pairs, vectors,
  strings, and bytevectors*; returns the same as `eqv?` for *booleans, symbols,
  numbers, characters, ports, procedures, and the empty list*; and then: "In all
  other cases, equal? may return either #t or #f." **Records are in neither
  list**, so they land in "all other cases" — the result is
  **implementation-defined**. Both Kaappi's `#f` and the field-recursing `#t` are
  conformant.
- **§6.1 `eqv?` (p. 30).** For records, `eqv?` is same-location (identity)
  equality; distinct records are never `eqv?`. The constraint "if two objects are
  `eqv?`, they must be `equal?` as well" therefore imposes nothing here (distinct
  records aren't `eqv?` to begin with), so we're free to choose.
- **§6.1 rule of thumb (p. 32).** "A rule of thumb is that objects are generally
  equal? if they print the same." Kaappi prints `#<foo 1>` for both instances, so
  by the report's own (non-normative) heuristic they "ought" to be `equal?`. A
  mild point in favor of structural.
- **§5.5 (p. 27).** Records are generative *at the type level*; nothing there
  constrains instance equality.

## Cross-implementation survey

Tested locally with the same record definition:

| Implementation | Targets | `equal?` on distinct-but-equal records |
|---|---|---|
| **Kaappi** v0.22.3 | R7RS | `#f` (identity) |
| Chibi 0.12 | R7RS | `#t` (structural) |
| Guile 3.0 | R7RS | `#t` (structural) |
| Gambit (per report) | R7RS | `#t` (structural) |
| Chez | **R6RS** | `#f` (identity) |

Among **native-R7RS** implementations it's 3-of-3 structural; Kaappi would be
the lone holdout. Chez also uses identity, but Chez is an R6RS implementation
(its default distribution ships no R7RS libraries), and R6RS's `equal?` doesn't
recurse into records — so it's an R6RS precedent, not an R7RS peer. `eqv?` is
identity-only everywhere, as required.

## Recommendation

**Support it** — make `equal?` structural on records — for an implementation
that bills itself as R7RS. Interop with R7RS peers outweighs the R6RS-flavored
argument for identity, and doing it while the ecosystem is young keeps the
breaking-direction cost (some `#f` results become `#t`) at its smallest.

### Trade-off to acknowledge

Some records model *identity* (unique tokens, resource wrappers), and R7RS gives
no transparency flag to separate "data" records from "opaque" ones, so structural
`equal?` is all-or-nothing. This is contained in Kaappi: built-in resource types
(mutexes, ports, threads) are native heap types, **not** `record_instance`, so
only user-defined records are affected — and those users can always fall back to
`eqv?`/`eq?` for identity. The three R7RS peers already live with this.

## Design decisions to settle

1. **Same-type requirement.** Compare only when both operands share the same
   `record_type` descriptor. (Recommended: yes — matches Gambit/Guile/Chibi.)
2. **Field comparison depth.** Recurse each field through the existing
   `deepEqualWithVisited`, so nested pairs/vectors/records compare deeply and
   leaf procedures/opaque values fall to the existing `eqv?`-style leaf handling
   (a record holding the same closure → `#t`, a different one → `#f`).
3. **Cycles.** Route through the shared `VisitedMap` (as pairs/vectors already
   do) so cyclic records terminate.
4. **Opaque records.** Decide whether to unconditionally recurse (simplest,
   matches peers) or add an opt-out later. Recommended: recurse unconditionally
   for now; revisit only if a transparency mechanism is introduced.

## Proposed implementation sketch

Add one arm to `deepEqualWithVisited` (`src/primitives.zig:913`), before the
final `return false;`:

```zig
if (types.isRecordInstance(a) and types.isRecordInstance(b)) {
    const ra = types.toObject(a).as(types.RecordInstance);
    const rb = types.toObject(b).as(types.RecordInstance);
    if (ra.record_type != rb.record_type) return false; // same type required
    if (ra.fields.len != rb.fields.len) return false;
    const key = VisitedKey{ .a = a, .b = b };
    if (visited.get(key) != null) return true;
    visited.put(key, {}) catch {};
    for (ra.fields, rb.fields) |fa, fb| {
        if (!deepEqualWithVisited(fa, fb, visited)) return false;
    }
    return true;
}
```

(Field/`record_type` access spelled to match `types_record.RecordInstance`;
confirm exact field names against `src/types_record.zig:104` and the accessors
at `src/primitives.zig:1104`.)

## Action items

- [ ] Confirm the design decisions above (esp. same-type requirement and
      opaque-record handling).
- [ ] Add the `record_instance` arm to `deepEqualWithVisited`
      (`src/primitives.zig`).
- [ ] Regression tests:
  - [ ] Zig unit test in `src/tests_*.zig`: distinct-but-equal records → `#t`;
        different-type same-shape → `#f`; different-field-value → `#f`;
        `eqv?`/`eq?` on the same pairs still `#f`.
  - [ ] Scheme test under `tests/scheme/` (the Reddit example, using distinct
        type/constructor names).
  - [ ] Cyclic record test (a record field pointing back to the record) —
        must terminate.
  - [ ] Record containing a procedure field — same closure `#t`, different
        closure `#f`.
- [ ] Verify green under `zig build test -Dgc-stress=true` (`deepEqualWithVisited`
      allocates a `VisitedMap`; the new arm allocates no heap Values, but confirm).
- [ ] Update `CONFORMANCE.md` — the SRFI 9 section (line 23) currently says
      nothing about equality; add a short "`equal?` on records" note recording
      the decision and the R7RS §6.1 basis.
- [ ] Consider a brief mention in end-user docs (kaappi.github.io) if record
      equality is documented there.
- [ ] Reply to the r/KaappiScheme poster once shipped.

## Priority

`low` — current behavior is correct and conformant; this is an
interoperability/ergonomics enhancement, not a defect.

baijum:
## Summary

Make `equal?` recurse into record fields, so two distinct instances of the same record type whose fields are pairwise `equal?` compare as `#t`. Today Kaappi compares records by identity only (same as `eq?`/`eqv?`).

R7RS §6.1 leaves records in the "all other cases" clause for `equal?`, so the result is implementation-defined — this is a permitted extension, not a conformance fix. But every native-R7RS implementation (Gambit, Guile, Chibi) recurses into fields, and Kaappi was the lone identity-only holdout. `eq?`/`eqv?` stay identity-based, as §6.1 requires.

```scheme
(define-record-type point (make-point x y) point? (x point-x) (y point-y))

(equal? (make-point 1 2) (make-point 1 2))   ; now #t (was #f)
(eqv?  (make-point 1 2) (make-point 1 2))    ; #f  (unchanged)
```

## Changes

- Add a `record_instance` arm to `deepEqualWithVisited` (`src/primitives.zig`):
  - Same record type required, compared by `types.sameRecordType` (identity, not pointer — a type survives an SRFI-18 thread hop at a new address, kaappi#1932).
  - Fields compared pairwise through the existing `deepEqualWithVisited`, so nested pairs/vectors/records/strings recurse and leaf procedures fall to identity.
  - Routes through the shared `VisitedMap` (as pairs/vectors do), so cyclic records terminate.
- Zig unit tests in `src/tests_records.zig`: distinct-but-equal, different-types-same-shape, differing fields, recursive fields, procedure-bearing fields, and isomorphic cyclic records.
- Scheme smoke test `tests/scheme/smoke/record-equal-2293.scm` (auto-discovered by `run-all.sh`).
- `CONFORMANCE.md` SRFI 9 section records the decision and its §6.1 basis.

## Verification

- `zig build test` — green
- `zig build test -Dgc-stress=true -Dtest-filter=records` — green
- `./zig-out/bin/kaappi tests/scheme/smoke/record-equal-2293.scm` — 9 passes

Fixes #2293

### DIFF
--- a/src/primitives.zig
+++ b/src/primitives.zig
@@ -966,6 +966,24 @@ fn deepEqualWithVisited(a: Value, b: Value, visited: *VisitedMap) bool {
         const nb = types.toNumericVector(b);
         return na.kind == nb.kind and std.mem.eql(u8, na.data, nb.data);
     }
+    if (types.isRecordInstance(a) and types.isRecordInstance(b)) {
+        const ra = types.toObject(a).as(types.RecordInstance);
+        const rb = types.toObject(b).as(types.RecordInstance);
+        // Same record type required (identity, not pointer -- a type
+        // survives an SRFI-18 thread hop at a new address, kaappi#1932), then
+        // fields compared pairwise. R7RS 6.1 leaves records in the "all other
+        // cases" clause, so this structural choice is a permitted extension
+        // that matches Gambit/Guile/Chibi (kaappi#2293).
+        if (!types.sameRecordType(ra.record_type, rb.record_type)) return false;
+        if (ra.fields.len != rb.fields.len) return false;
+        const key = VisitedKey{ .a = a, .b = b };
+        if (visited.get(key) != null) return true;
+        visited.put(key, {}) catch {};
+        for (ra.fields, rb.fields) |fa, fb| {
+            if (!deepEqualWithVisited(fa, fb, visited)) return false;
+        }
+        return true;
+    }
     return false;
 }
 
--- a/src/primitives.zig
+++ b/src/primitives.zig
@@ -975,6 +975,9 @@ fn deepEqualWithVisited(a: Value, b: Value, visited: *VisitedMap) bool {
         // cases" clause, so this structural choice is a permitted extension
         // that matches Gambit/Guile/Chibi (kaappi#2293).
         if (!types.sameRecordType(ra.record_type, rb.record_type)) return false;
+        // Defensive: sameRecordType already implies equal field counts (both
+        // instances are sized from record_type.num_fields), so this is a cheap
+        // belt-and-suspenders guard, not a load-bearing check.
         if (ra.fields.len != rb.fields.len) return false;
         const key = VisitedKey{ .a = a, .b = b };
         if (visited.get(key) != null) return true;
--- a/tests/scheme/smoke/record-equal-2293.scm
+++ b/tests/scheme/smoke/record-equal-2293.scm
@@ -2,7 +2,7 @@
 ;; distinct instances of the same record type with equal? fields are equal?.
 ;; eq?/eqv? stay identity-based (R7RS 6.1 requires that; the equal? behavior
 ;; is a permitted extension under the "all other cases" clause).
-(import (scheme base) (scheme write) (scheme process-context) (srfi 64))
+(import (scheme base) (scheme write) (scheme process-context) (srfi 64) (srfi 69))
 
 (test-begin "record-equal-2293")
 
@@ -36,6 +36,29 @@
 ;; Records of different types are never equal?, even with equal fields.
 (test-assert "different record types not equal?"
   (not (equal? (make-point 1 2) (make-other 1))))
+
+;; member/assoc default to equal? and share deepEqual, so they now find
+;; records structurally — every equal? consumer moved together.
+(test-assert "member finds a structurally-equal record"
+  (pair? (member (make-point 1 2) (list (make-point 1 2)))))
+(test-assert "assoc finds a structurally-equal record key"
+  (pair? (assoc (make-point 1 2) (list (cons (make-point 1 2) 'v)))))
+
+;; The hash/equality contract: deepEqual is structural on records, so valueHash
+;; must be too — two equal? records have to hash alike, or a default SRFI 69
+;; table (equal? + valueHash) silently loses record entries (kaappi#2293).
+(test-assert "equal? records hash equal"
+  (= (hash (make-point 1 2)) (hash (make-point 1 2))))
+
+(test-assert "populated equal?-table round-trips structurally-equal record keys"
+  (let ((ht (make-hash-table)))
+    (do ((i 0 (+ i 1))) ((= i 200))
+      (hash-table-set! ht (make-point i i) i))
+    (let loop ((i 0) (ok #t))
+      (if (= i 200)
+          ok
+          (loop (+ i 1)
+                (and ok (= (hash-table-ref/default ht (make-point i i) -1) i)))))))
 
 ;; Cyclic records must terminate (the cycle-tracking visited set handles it).
 (define-record-type node (make-node v) node? (v node-v set-node-v!))
--- a/tests/scheme/smoke/record-equal-2293.scm
+++ b/tests/scheme/smoke/record-equal-2293.scm
@@ -1,0 +1,50 @@
+;; Regression test for #2293: equal? recurses into record fields, so two
+;; distinct instances of the same record type with equal? fields are equal?.
+;; eq?/eqv? stay identity-based (R7RS 6.1 requires that; the equal? behavior
+;; is a permitted extension under the "all other cases" clause).
+(import (scheme base) (scheme write) (scheme process-context) (srfi 64))
+
+(test-begin "record-equal-2293")
+
+(define-record-type point (make-point x y) point? (x point-x) (y point-y))
+(define-record-type other (make-other x)   other? (x other-x))
+
+;; The headline case from the Reddit report.
+(test-assert "distinct-but-equal records are equal?"
+  (equal? (make-point 1 2) (make-point 1 2)))
+
+;; eq?/eqv? must remain identity-based.
+(test-assert "eqv? on distinct records is #f"
+  (not (eqv? (make-point 1 2) (make-point 1 2))))
+(test-assert "eq? on distinct records is #f"
+  (not (eq? (make-point 1 2) (make-point 1 2))))
+
+;; A record is equal? to itself.
+(test-assert "record equal? to itself"
+  (let ((p (make-point 1 2))) (equal? p p)))
+
+;; Differing fields are not equal?.
+(test-assert "differing field values not equal?"
+  (not (equal? (make-point 1 2) (make-point 1 9))))
+
+;; Fields compared recursively.
+(test-assert "recursive field comparison (equal? contents)"
+  (equal? (make-point (list 1 2) "ab") (make-point (list 1 2) "ab")))
+(test-assert "recursive field comparison (differing contents)"
+  (not (equal? (make-point (list 1 2) "ab") (make-point (list 1 2) "ac"))))
+
+;; Records of different types are never equal?, even with equal fields.
+(test-assert "different record types not equal?"
+  (not (equal? (make-point 1 2) (make-other 1))))
+
+;; Cyclic records must terminate (the cycle-tracking visited set handles it).
+(define-record-type node (make-node v) node? (v node-v set-node-v!))
+(test-assert "isomorphic cyclic records are equal?"
+  (let ((x (make-node 1)) (y (make-node 1)))
+    (set-node-v! x x)
+    (set-node-v! y y)
+    (equal? x y)))
+
+(let ((runner (test-runner-current)))
+  (test-end "record-equal-2293")
+  (when (> (test-runner-fail-count runner) 0) (exit 1)))
--- a/CONFORMANCE.md
+++ b/CONFORMANCE.md
@@ -24,6 +24,15 @@ Implemented: `cons*`, `xcons`, `list-tabulate`, `circular-list`, `iota`, `proper
 
 **Coverage: 100%.** `define-record-type` is implemented as R7RS compiler syntax.
 
+`equal?` recurses into record fields: two distinct instances of the same
+record type whose fields are pairwise `equal?` compare `#t`. R7RS §6.1 leaves
+records in the "all other cases" clause (the result is implementation-defined),
+so this is a permitted extension — matching Gambit, Guile, and Chibi — rather
+than a conformance requirement. Records of *different* types are never
+`equal?`, even with equal field values. `eq?`/`eqv?` remain identity-based on
+records, as §6.1 requires. Cyclic records terminate via the same visited-map
+used for pairs and vectors (kaappi#2293).
+
 ### SRFI 13 — String Library
 
 **Coverage: 97%** (30 of 31 non-mutating spec procedures)
--- a/src/primitives_hashtable.zig
+++ b/src/primitives_hashtable.zig
@@ -386,8 +386,24 @@ fn valueHashDepth(key: Value, depth: usize) usize {
         for (nv.data) |b| h = h *% 31 +% b;
         return h;
     }
-    // Everything left (procedures, ports, records, ...) is compared by identity
-    // by `deepEqual`, so hashing the address is consistent with equality.
+    if (types.isRecordInstance(key)) {
+        if (depth >= MAX_HASH_DEPTH) return DEEP_CUTOFF_HASH;
+        const rec = types.toObject(key).as(types.RecordInstance);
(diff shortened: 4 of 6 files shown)


src/primitives.zig
```diff
@@ -966,6 +966,24 @@ fn deepEqualWithVisited(a: Value, b: Value, visited: *VisitedMap) bool {
         const nb = types.toNumericVector(b);
         return na.kind == nb.kind and std.mem.eql(u8, na.data, nb.data);
     }
+    if (types.isRecordInstance(a) and types.isRecordInstance(b)) {
```
baijum, line 969: **Blocking:** this arm breaks the hash/equality contract for records. `deepEqual` is now structural on records, but `valueHash` still hashes them by address — `primitives_hashtable.zig:389`:

```zig
// Everything left (procedures, ports, records, ...) is compared by identity
// by `deepEqual`, so hashing the address is consistent with equality.
return @truncate(key *% 2654435761);
```

That comment was true before this PR and is false after it. Verified on this branch — a default SRFI 69 table (`equal?` + `valueHash`) silently loses every record entry:

```scheme
(define-record-type point (make-point x y) point? (x point-x) (y point-y))
(define ht (make-hash-table))
(do ((i 0 (+ i 1))) ((= i 200)) (hash-table-set! ht (make-point i i) i))
;; re-lookup each key with a fresh equal? record => 200 of 200 miss
```

A one-entry table still finds its key — the two address-derived hashes collide modulo the tiny initial mask — which is exactly what makes this easy to miss; it takes a populated table to surface. `(= (hash (make-point 1 2)) (hash (make-point 1 2)))` is already `#f`, directly visible.

This is the same bug class fixed for f64vector in kaappi#2023, documented in the comment block right above that line. Under the triage rubric it is a silently wrong answer on a path users reach → `high`; I would not merge without it.

Suggested fix: add a `record_instance` arm to `valueHashDepth` that folds `record_type.identity` — that is what `sameRecordType` gates on, so it is the only type discriminator the hash may use — with per-field `valueHashDepth` hashes, capped like the vector arm (first few fields). `valueHash` has no visited set, so cyclic record fields are absorbed by the `MAX_HASH_DEPTH` sentinel, same trade the pair arm already makes. A structural record hash is also safe for `.eqv`-mode tables: `eqv?` stays identity on records, so structural hashing can only add collisions there, never misses. Please also flip the stale comment at the bottom of `valueHash`, and add a regression test pair: equal records hash equal, and the populated-table round-trip above.

src/primitives.zig
```diff
@@ -966,6 +966,24 @@ fn deepEqualWithVisited(a: Value, b: Value, visited: *VisitedMap) bool {
...
+        const ra = types.toObject(a).as(types.RecordInstance);
+        const rb = types.toObject(b).as(types.RecordInstance);
+        // Same record type required (identity, not pointer -- a type
+        // survives an SRFI-18 thread hop at a new address, kaappi#1932), then
+        // fields compared pairwise. R7RS 6.1 leaves records in the "all other
+        // cases" clause, so this structural choice is a permitted extension
+        // that matches Gambit/Guile/Chibi (kaappi#2293).
+        if (!types.sameRecordType(ra.record_type, rb.record_type)) return false;
+        if (ra.fields.len != rb.fields.len) return false;
```
baijum, line 981: [nit] This is unreachable: `allocRecordInstance` sizes the slice from `record_type.num_fields` (`gc_alloc.zig:454`), and the `sameRecordType` gate two lines up already established both instances carry the same type. Fine as defense-in-depth — just flagging it so it doesn't read as a load-bearing check.

tests/scheme/smoke/record-equal-2293.scm
```diff
@@ -0,0 +1,50 @@
...
+
+;; A record is equal? to itself.
+(test-assert "record equal? to itself"
+  (let ((p (make-point 1 2))) (equal? p p)))
+
+;; Differing fields are not equal?.
+(test-assert "differing field values not equal?"
+  (not (equal? (make-point 1 2) (make-point 1 9))))
+
+;; Fields compared recursively.
```
baijum, line 30: Worth one more assertion here pinning the propagation: `member`/`assoc` default to `equal?` and route through the same `deepEqual` (`primitives_list.zig:163`), so they now find records structurally — e.g. `(member (make-point 1 2) (list (make-point 1 2)))` returns the list. Cheap line to add, and it documents that every `equal?` consumer moved together — the same shared path that makes the hash-table contract in the `primitives.zig` comment part of this change's blast radius.

baijum:
Reviewed and verified on this branch (built locally; `zig build test -Dtest-filter=records`, the smoke test, and `srfi9.scm` all green as claimed).

**One blocking finding** (inline on the new `record_instance` arm): `deepEqual` becoming structural for records breaks the hash/equality contract — `valueHash` still hashes records by address, so a default SRFI 69 `equal?` table silently loses record entries. Reproduced: 200 keys inserted, 200 lookups miss; `(hash (make-point 1 2))` differs across two `equal?` instances. Same class as the f64vector address-hash bug (kaappi#2023), and `primitives_hashtable.zig:389`'s justification comment is now stale. Needs a matching `record_instance` arm in `valueHashDepth` before merge.

Everything else checks out:

- The `sameRecordType` gate (identity, not pointer) is the right call and correctly handles the SRFI-18 deep-copied-type case from kaappi#1932; SRFI 237 inheritance and nongenerative uids fall out correctly since instances of sibling/parent types carry distinct `identity` values.
- Every other `deepEqual` consumer moves consistently and needs no changes: `member`/`assoc` (`primitives_list.zig`, `primitives_srfi1.zig`), hashtable lookups themselves, and `fmt.zig` (round-trip verification only; records are not reader-reachable). No optimizer/IR path folds `equal?` separately.
- Routing through the `VisitedMap` matches the pair/vector arms, and the cyclic-record tests cover the termination case.
- `eq?`/`eqv?` staying identity-based, the `CONFORMANCE.md` note, and the `srfi9.scm` expectation flip are all correct.

Two non-blocking notes inline: the `fields.len` check is unreachable given `allocRecordInstance` sizing, and the smoke test could pin `member`/`assoc` propagation for free.

src/primitives.zig
baijum: Agreed — added a comment marking it as a cheap belt-and-suspenders guard rather than a load-bearing check.

tests/scheme/smoke/record-equal-2293.scm
baijum: Done — added member and assoc assertions to record-equal-2293.scm confirming both now find records structurally.

src/primitives.zig
baijum: Fixed in 242125ea. Added a record_instance arm to valueHashDepth that folds record_type.identity (the discriminator sameRecordType gates on, so it survives an SRFI-18 thread hop) with the first few field hashes, capped like the vector arm; cyclic fields hit the MAX_HASH_DEPTH sentinel. Flipped the stale identity-fallback comment. Regression tests: `(= (hash (make-point 1 2)) (hash (make-point 1 2)))` and the 200-entry populated-table round-trip in record-equal-2293.scm, plus a direct valueHash unit test.

baijum:
**LGTM** 🚢 — 242125ea resolves the blocking finding correctly, verified on this branch.

**The hash fix checks out.** Re-ran the repro from my earlier comment: 200 record keys inserted and re-looked-up in a default `equal?` table → **0 misses** (was 200), and `(hash (make-point 1 2))` now agrees across `equal?` instances. The arm itself is the right shape: it folds `record_type.identity` — the exact discriminator `sameRecordType` gates on, so it stays correct across SRFI-18 thread hops — caps field hashing like the vector arm, and lets `MAX_HASH_DEPTH` absorb cyclic fields. Safe for `.eqv`-mode tables too (identity `eqv?` + structural hash only adds collisions, never misses). No allocation in the hash path, so no GC-rooting concerns. Stale comment flipped, and both regression tiers added (the Zig hash unit test and the populated-table smoke round-trip). The `member`/`assoc` smoke assertions and the belt-and-suspenders comment were folded in as well.

**Verification on 242125ea:** build green; `zig build test -Dtest-filter=records` and `-Dtest-filter=hash` green; smoke test 13/13; `srfi69.scm`, `srfi69-ext.scm`, `srfi9.scm` all green; and the **full** unit suite under `zig build test -Dgc-stress=true` green (exit 0) — that completes the one #2293 action item the PR description had only covered with the filtered records run.

**Against #2293's action-item list:** design decisions settled and implemented as recommended (same-type via `sameRecordType`; opaque records recurse unconditionally, documented in `CONFORMANCE.md`); the `record_instance` arm added; all four requested regression tests present on both tiers (distinct-but-equal, different-type same-shape, differing field, `eqv?`/`eq?` identity, cyclic, procedure fields); gc-stress green; `CONFORMANCE.md` updated with the §6.1 basis. Two items remain, both explicitly post-merge: the end-user docs mention (kaappi.github.io, "consider") and the reply to the r/KaappiScheme poster once shipped — worth a note somewhere that survives the auto-close of `Fixes #2293`, since that checklist lives in this issue.

PR #2296, opened 2026-08-24
Document non-portability of reusing one identifier as record type name and constructor name

### LINKED ISSUE #2294: Document non-portability of reusing one identifier as both record type name and constructor name
## Summary

`define-record-type` accepts the same identifier for both the type `⟨name⟩` and
the `⟨constructor name⟩`, e.g.:

```scheme
(define-record-type foo (foo x) foo? (x bar))
```

Kaappi (and Gambit) accept this; **Chibi and Guile reject it**. R7RS neither
explicitly permits nor explicitly forbids the reuse, so the outcome is
**implementation-dependent** — meaning code written this way is **not portable**.
Kaappi's acceptance is not a rule violation, but it leans on unspecified
territory and is *semantically lossy* (the type-name binding is shadowed by the
constructor). This issue is to **document the non-portability** so users don't
mistake "works on Kaappi" for "portable."

Surfaced while investigating the `equal?`-on-records report (#2293), whose
original snippet used this form:
https://www.reddit.com/r/KaappiScheme/comments/1vwodyg/better_if_record_types_with_equal_components_were/

## Reproduction

```scheme
;; Accepted on Kaappi and Gambit:
(define-record-type foo (foo x) foo? (x bar))
(foo 1)            ; => #<foo 1>   (foo is now the constructor)

;; Rejected on Chibi 0.12:
;;   ERROR: same binding for record rtd and constructor: foo
;; Rejected on Guile 3.0:
;;   foo is bound to the record type; (foo 1) => "Wrong type to apply: #<record-type foo>"
```

Distinct names work everywhere and are the portable form:

```scheme
(define-record-type foo (make-foo x) foo? (x bar))
```

## Spec analysis (docs/errata-corrected-r7rs.pdf, §5.5, p. 27)

The grammar treats `⟨name⟩` and `⟨constructor name⟩` as two separate identifier
positions with two separate bindings:

- `⟨name⟩` "is bound to a representation of the record type itself."
- `⟨constructor name⟩` "is bound to a procedure" that constructs records.

The only explicit distinctness ("it is an error") rules cover **field names** and
**accessor/mutator names** — the spec is **silent** on whether `⟨name⟩` and
`⟨constructor name⟩` may coincide.

The decisive clause is that the type-name representation "may be a run-time object
or a purely syntactic representation." R7RS deliberately leaves that
implementation-defined, which is what makes the collision's outcome
implementation-dependent:

- If an implementation binds `⟨name⟩` as a **syntactic keyword**, it cannot also
  bind that identifier as a **variable** (the constructor) in the same scope →
  conflict → Chibi/Guile reject.
- If it binds `⟨name⟩` as an ordinary **run-time variable**, the collision is
  tolerable → Kaappi/Gambit accept, with `foo` ending up bound to the constructor
  (the type-name binding is lost/shadowed).

Both behaviors are conformant. Rejecting is arguably the more faithful reading;
accepting is not a rule violation but is non-portable and lossy.

Contrast with #2293: there, §6.1 *explicitly* blesses both `#t` and `#f` for
`equal?` on records. Here the spec simply doesn't define the outcome — a weaker
footing.

## Cross-implementation summary

| Implementation | `(define-record-type foo (foo x) foo? (x bar))` |
|---|---|
| Kaappi v0.22.3 | accepted (`foo` becomes the constructor) |
| Gambit | accepted |
| Chibi 0.12 | rejected — "same binding for record rtd and constructor" |
| Guile 3.0 | rejected — `foo` is the record type; `(foo 1)` fails to apply |

## Recommendation

Primary: **document it.** This is a "behavior is defensible, only its
description/discoverability is lacking" case → `priority: low`.

Optional stretch (separate decision, not required by this issue): consider
emitting a warning — or matching the stricter implementations by rejecting the
collision — so `define-record-type` nudges users toward portable code. Rejecting
would be a small behavior change; a warning is non-breaking. Deferred pending
maintainer preference; the safe, in-scope action here is documentation only.

## Action items

- [ ] Add a short note to end-user docs (kaappi.github.io) on `define-record-type`:
      the type name and constructor name should be distinct identifiers; reusing
      one identifier for both is non-portable (accepted by Kaappi/Gambit, rejected
      by Chibi/Guile) and shadows the type-name binding.
- [ ] Optionally add a one-line note to `CONFORMANCE.md` under SRFI 9 / records
      recording this as a known portability caveat.
- [ ] (Stretch, needs a decision) Evaluate emitting a compile-time warning when
      `⟨name⟩` and `⟨constructor name⟩` are the same identifier; if pursued, file
      as its own issue with a regression test under `tests/scheme/`.

## Priority

`low` — Kaappi's behavior is conformant; only its portability implications are
undocumented. No correctness defect.

baijum:
Fixes #2294

R7RS §5.5 is silent on whether `<name>` and `<constructor name>` may be the same identifier. Kaappi (like Gambit) accepts the collision, with the constructor winning the name; Chibi and Guile reject the same form. The outcome is implementation-dependent, so code that reuses one identifier for both positions is not portable.

## Changes

- **`CONFORMANCE.md`** — the SRFI 9 section now documents the collision and recommends the portable distinct-names form (following the precedent set by #2295 for the `equal?`-on-records decision).
- **`src/vm_records.zig`** — comment at the R7RS desugarer's constructor generation (`handleDefineRecordType`) recording the collision semantics.
- **`src/compiler_lambda.zig`** — matching comment at `compileDefineRecordType`, cross-referencing `vm_records`.
- **`tests/scheme/smoke/record-name-ctor-collision-2294.scm`** — SRFI-64 smoke test pinning Kaappi's side of the contract on both the R7RS and SRFI 237 clause-syntax paths, so the documentation cannot drift.

## Verification

- `zig fmt --check` and `zig build` clean
- Unit suite: exit 0
- R7RS suite: 1395 pass, 0 fail
- Full `tests/scheme/run-all.sh`: 2102 pass, 0 fail
- markdownlint: 0 issues

### DIFF
--- a/CONFORMANCE.md
+++ b/CONFORMANCE.md
@@ -33,6 +33,17 @@ than a conformance requirement. Records of *different* types are never
 records, as §6.1 requires. Cyclic records terminate via the same visited-map
 used for pairs and vectors (kaappi#2293).
 
+R7RS §5.5 is silent on whether `<name>` and `<constructor name>` may be the
+same identifier. Kaappi accepts the collision:
+`(define-record-type foo (foo x) foo? (x bar))` binds `foo` to the
+*constructor* — `(foo 1)` constructs a record, and the record type itself is
+not reachable under that name (it lives only behind the internal
+`__record_type_<name>` alias). Chibi and Guile reject the same form, so the
+outcome is implementation-dependent: code that reuses one identifier for both
+positions is not portable. Use distinct names —
+`(define-record-type foo (make-foo x) foo? (x bar))` — which every
+implementation accepts (kaappi#2294).
+
 ### SRFI 13 — String Library
 
 **Coverage: 97%** (30 of 31 non-mutating spec procedures)
--- a/CONFORMANCE.md
+++ b/CONFORMANCE.md
@@ -34,15 +34,19 @@ records, as §6.1 requires. Cyclic records terminate via the same visited-map
 used for pairs and vectors (kaappi#2293).
 
 R7RS §5.5 is silent on whether `<name>` and `<constructor name>` may be the
-same identifier. Kaappi accepts the collision:
+same identifier. Kaappi accepts the collision on the R7RS form —
 `(define-record-type foo (foo x) foo? (x bar))` binds `foo` to the
-*constructor* — `(foo 1)` constructs a record, and the record type itself is
-not reachable under that name (it lives only behind the internal
-`__record_type_<name>` alias). Chibi and Guile reject the same form, so the
-outcome is implementation-dependent: code that reuses one identifier for both
-positions is not portable. Use distinct names —
+*constructor*, so `(foo 1)` constructs a record; the record type itself is
+not reachable under that name (on this path it is only ever bound behind the
+internal `__record_type_<name>` alias). Chibi and Guile reject the same form,
+so the outcome is implementation-dependent: code that reuses one identifier
+for both positions is not portable. Use distinct names —
 `(define-record-type foo (make-foo x) foo? (x bar))` — which every
-implementation accepts (kaappi#2294).
+implementation accepts. On the SRFI 237 clause form the collision is a
+deviation rather than an unspecified outcome: SRFI 237 specifies that the
+record name evaluates to the underlying record descriptor, and Kaappi's
+constructor-wins handling overwrites that name→descriptor binding
+(kaappi#2294).
 
 ### SRFI 13 — String Library
 
--- a/src/vm_records.zig
+++ b/src/vm_records.zig
@@ -651,6 +651,14 @@ pub fn handleDefineRecordTypeR6RS(vm: *VM, args: Value) VMError!Value {
         // SRFI 237: "As an expression, this keyword evaluates to the
         // underlying record descriptor" -- the declared name itself, not
         // just the hidden redefinition-stable alias above, must be bound.
+        //
+        // Collision note (kaappi#2294): when <name> also names the
+        // constructor, the constructor's alias define below overwrites
+        // this name->descriptor binding, so the identifier ends up the
+        // constructor and SRFI 237's name-as-expression guarantee is lost
+        // for it -- a documented deviation (CONFORMANCE.md SRFI 9). The
+        // ordering is load-bearing: this binding must stay BEFORE the
+        // constructor defines, or the rtd would silently win instead.
         lib_env.put(spec.type_name, rt_val) catch return VMError.OutOfMemory;
     } else {
         vm.defineGlobal(internal_name, rt_val) catch return VMError.OutOfMemory;
@@ -1044,6 +1052,9 @@ pub fn expandRecordTypeDefines(
     }
 
     // 2. (define ctor (let (( __rt __record_type_X)) (lambda (f1 f2) (%make-record  __rt ...))))
+    // Collision note (kaappi#2294): a <name>/<constructor name> collision is
+    // accepted here too — the ctor define wins the name. See
+    // handleDefineRecordType's constructor block for the full discussion.
     {
         const rt_local = gc.allocSymbol(" __rt") catch return CompileError.OutOfMemory;
         const lambda_sym = gc.allocSymbol("lambda") catch return CompileError.OutOfMemory;
--- a/src/vm_records.zig
+++ b/src/vm_records.zig
@@ -48,6 +48,15 @@ pub fn handleDefineRecordType(vm: *VM, args: Value) VMError!Value {
     // does not retarget previously created constructors (#1203):
     // (define ctor (let (( __rt __record_type_X))
     //               (lambda (f1 f2) (%make-record  __rt f_for_0 ...))))
+    //
+    // Collision note (kaappi#2294): the R7RS grammar lets <name> and
+    // <constructor name> be the same identifier, and this desugarer accepts
+    // it — the constructor's define below wins the name, so after
+    // (define-record-type foo (foo x) foo? (x bar)) the identifier `foo`
+    // names the constructor and the record type is reachable only through
+    // the internal __record_type_foo alias (never bound under `foo` itself
+    // on this R7RS path). R7RS §5.5 leaves the collision unspecified; Chibi
+    // and Guile reject it, so such code is not portable.
     {
         vm.gc.no_collect += 1;
         defer vm.gc.no_collect -= 1;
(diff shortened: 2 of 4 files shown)


CONFORMANCE.md
```diff
@@ -33,6 +33,17 @@ than a conformance requirement. Records of *different* types are never
 records, as §6.1 requires. Cyclic records terminate via the same visited-map
 used for pairs and vectors (kaappi#2293).
 
+R7RS §5.5 is silent on whether `<name>` and `<constructor name>` may be the
```
baijum, line 36: The R7RS-form half of this paragraph is exactly right, but as written it also covers the clause form (the pin test exercises it), and for that form "R7RS §5.5 is silent" isn't the whole story. SRFI 237 itself specifies the name binding — "As an expression, this keyword evaluates to the underlying record descriptor" — the same sentence `vm_records.zig` cites at the `lib_env.put(spec.type_name, rt_val)` binding in `handleDefineRecordTypeR6RS`, which exists for exactly that reason. On that path the rtd *is* bound to the declared name and then overwritten by the constructor define, so the collision doesn't choose among unspecified alternatives — it overrides a specified binding (verified on current main: `(procedure? r6foo)` → `#t`, `(record-type-descriptor? r6foo)` → `#f` after `(define-record-type (r6foo r6foo r6foo?) (fields (immutable v)))`).

Since CONFORMANCE.md is the document that records deviations from SRFI specs, consider one extra sentence scoping this: on the SRFI 237 clause form, the collision replaces the name→descriptor binding SRFI 237 specifies (a minor, documented deviation); on the R7RS form the outcome falls within §5.5's silence. Relatedly, "(it lives only behind the internal `__record_type_<name>` alias)" describes the R7RS path's mechanism — on the clause path the rtd is bound to the name first and then clobbered — so as a general statement it reads slightly too broad.

src/vm_records.zig
```diff
@@ -48,6 +48,15 @@ pub fn handleDefineRecordType(vm: *VM, args: Value) VMError!Value {
     // does not retarget previously created constructors (#1203):
     // (define ctor (let (( __rt __record_type_X))
     //               (lambda (f1 f2) (%make-record  __rt f_for_0 ...))))
+    //
+    // Collision note (kaappi#2294): the R7RS grammar lets <name> and
```
baijum, line 52: Accurate for this path. One placement gap: the path where a real shadow happens is the R6RS desugarer. `handleDefineRecordTypeR6RS` binds the rtd to the declared name (per SRFI 237's name-as-expression rule, ~line 654/657) and the user-ctor define at ~line 770 then overwrites it — and that ordering is load-bearing: move the name→rtd binding after the ctor define and the rtd silently wins instead. The smoke test pins the outcome on both paths, but the only collision comment sits here, on the path that never binds the type name at all. A one-line cross-ref at the ctor alias define (or the name binding) in `handleDefineRecordTypeR6RS` would put the note where the shadowing actually occurs.

Minor: this is 2 of 3 R7RS-path constructor generators — `expandRecordTypeDefines` (library-body path) accepts the collision the same way and carries no note. The `See vm_records.handleDefineRecordType` cross-ref style already used in compiler_lambda.zig might age better there than duplicating the full block.

tests/scheme/smoke/record-name-ctor-collision-2294.scm
```diff
@@ -0,0 +1,34 @@
...
+(test-assert "constructor wins the name (foo is a procedure)"
+  (procedure? foo))
+(test-assert "foo constructs a record"
+  (foo? (foo 1)))
+(test-equal "accessor works on the colliding constructor's record"
+  1 (bar (foo 1)))
+
+;; R6RS clause-syntax (SRFI 237) name spec: same collision, same outcome —
+;; the constructor define rebinds the name over the rtd binding.
+(define-record-type (r6foo r6foo r6foo?) (fields (immutable v)))
```
baijum, line 23: Good that both syntaxes are pinned. Suggested extra assertion for the clause path, pinning the non-obvious half of the outcome — that the SRFI 237 name-as-expression binding does *not* survive the collision:

```scheme
(test-assert "R6RS clause syntax: colliding name is not the record descriptor"
  (not (record-type-descriptor? r6foo)))
```

Both values verified on current main (`procedure?` → `#t`, `record-type-descriptor?` → `#f`), and `record-type-descriptor?` is already imported via `(srfi 237)`. This is what makes the SRFI 237-side deviation explicit in the test, rather than only in prose.

baijum (commented):
Verification (against PR head 6e1d89e6):

- `zig fmt --check` and `zig ast-check` clean on both Zig files (comment-only diff, so no rebuild needed for evidence).
- `markdownlint-cli2` with the repo config: 0 issues.
- Ran the new smoke test against a main-built binary (the diff changes no behavior): 6 expected passes, exit 0. `run-all.sh` globs `tests/scheme/smoke/*.scm`, so it is auto-discovered.
- Both new comments match their code paths: on the R7RS desugarers the type is bound only under the space-prefixed internal alias, so the constructor define is the sole binding of the name — nothing is actually shadowed, which the comments correctly acknowledge.
- `define-record-type` is eval-fallback in the native tier (`ir.zig` `other_special_forms`), so both tiers share these desugarers — no `tests/scheme/compile/` counterpart needed for this pin.

Findings are inline. Summary: the core deliverable — documenting the R7RS-form non-portability and pinning it — is correct and verified. The one substantive point is that the SRFI 237 clause-syntax path is not covered by "§5.5 is silent": SRFI 237 itself specifies the record name evaluates to the record descriptor, and Kaappi's constructor-wins outcome overrides that specified binding, which is worth naming in CONFORMANCE.md since that document exists to record deviations. The other two comments are a comment-placement suggestion and an optional extra test assertion. None of these block merge.

CONFORMANCE.md
baijum: Addressed in b986ab24. The paragraph now splits the two forms: the R7RS form stays scoped to §5.5's silence (with the internal-alias claim explicitly "on this path"), and a new sentence records that on the SRFI 237 clause form the collision overrides SRFI 237's specified name→record-descriptor binding — a documented deviation, as CONFORMANCE.md exists to record.

src/vm_records.zig
baijum: Addressed in b986ab24. Added the collision note at the R6RS name→descriptor binding in handleDefineRecordTypeR6RS, explicitly calling out that the ordering is load-bearing (the binding must stay before the constructor defines or the rtd silently wins). Also added the short 'See handleDefineRecordType's constructor block' cross-ref at expandRecordTypeDefines (the 3rd R7RS-path generator) rather than duplicating the block.

tests/scheme/smoke/record-name-ctor-collision-2294.scm
baijum: Addressed in b986ab24. Added the assertion pinning that the colliding name is not the record descriptor on the clause path — (not (record-type-descriptor? r6foo)) — with a comment explaining that SRFI 237's name-as-expression binding does not survive the collision. Smoke test now 7 expected passes.

baijum:
LGTM

Issue #1981, opened 2026-07-31
[LSP] Whole files go undiagnosed, only the first error is shown, KP4xxx lints never appear, and ranges are fabricated


baijum:
**Category:** Language server / diagnostics · **Severity:** wrong-result · **Found by:** systematic audit v2, Phase 6D (#1890)

`docs/dev/diagnostics-json.md` states the CLI and the LSP share the `Diagnostic` shape so they "can never drift". They agree on **shape**; they diverge on **values** four ways. Cross-checked by running the same file through `kaappi check --diagnostics=json` and the LSP.

Agreement is real where it holds: on read and compile errors both agree exactly on `code`, `severity` and start `line` (4 cases pinned as enabled assertions).

## 1. Whole files go undiagnosed — the serious one

A file beginning with `#| unterminated` publishes **`[]`** — and `documentSymbol` also returns `[]` — hiding a real `(if)` error further down. `check` reports `KP1001` on the same file. Same for a trailing `#|` and a dangling `#;`.

**Control:** an unterminated *string* — a `readDatum()` error rather than a skip-phase one — is reported identically by both.

Mechanism (inferred from source, consistent with every probe): `while (r.hasMore() catch false)`. `hasMore()` skips intertoken space and can fail *there*; `catch false` converts that failure into "end of input", so the document silently looks empty.

## 2. Only the first diagnostic is ever reported

`(if)\n(let)\n` → LSP shows 1, `check` shows 2. `runDiagnostics` breaks on the first error.

## 3. `KP4xxx` lints never appear

`(car 1 2 3)` → `check` reports `KP4002` and exits 1; the LSP shows a clean file. Consequence: the LSP can **never** emit `severity: 2`, so `KP4001` warnings are invisible in an editor entirely.

## 4. Ranges are fabricated whole-line spans

On `(define (f) (if))`, `check` pinpoints characters 12–16; the LSP highlights `0..999` — the whole line.

## Also: a documentation gap

`check --diagnostics=json` writes to **stdout** (documented in `check.md:26`), while `diagnostics-json.md` says stderr without noting the exception.

**Verified:** fresh ReleaseSafe build, isolated `KAAPPI_HOME`; all four divergences and the unterminated-string control reproduced against the real binary, cross-checked against `check --diagnostics=json` on identical files.

baijum:
Fixed by #2297 (merged as 57c6869).

PR #2297, opened 2026-08-24
Fix LSP protocol/lifecycle and diagnostics drift

### LINKED ISSUE #1980: [LSP] Six protocol and lifecycle defects: no response on bad params, no shutdown state, wrong exit code, id coercion

baijum:
Fixes #1980 and #1981 — the six protocol/lifecycle defects and the four diagnostics divergences from the Phase 6D LSP audit.

## Protocol and lifecycle (kaappi#1980)

- **Unusable params get a reply** — `hover`/`definition`/`references`/`completion`/`documentSymbol` now answer `-32602 InvalidParams` instead of silently stranding the client on that id.
- **Malformed headers are survivable** — a non-numeric, missing, or zero `Content-Length` skips that frame and resynchronises on the next, instead of ending the whole session.
- **Shutdown state exists** — `shutdown` before `initialize` errors `-32002`; requests after `shutdown` error `-32600`.
- **Exit code contract** — `exit` without a prior `shutdown` exits status 1.
- **End-of-line overrun** — `lineColToOffset` clamps to the line end, not the document end.
- **Unrepresentable ids** — a `null`/float request id answers an Invalid Request with `id: null` (`-32600`), never a fabricated `id: 0`.

## Diagnostics (kaappi#1981)

- `runDiagnostics` now drives the exact `kaappi check` analysis (a new `src/check.zig` `analyzeSource`, shared by `run`, `analyzeForTest`, and the LSP), so the two surfaces share the analysis, not just the serializer:
  - a whole-file read error (e.g. a leading unterminated block comment) is reported, not swallowed
  - every failing form is reported, not just the first
  - KP4xxx lints (severity-2 warnings included) reach the editor
  - ranges carry the real span, not a whole-line `0..999` sentinel

## Tests & docs

- `tests/scheme/lsp/lsp.sh` enables all previously disabled `# FAIL: #1980`/`#1981` assertions (now 184, up from 152), each beside a control.
- `docs/dev/diagnostics-json.md` notes the `check --diagnostics=json` stdout exception and the shared analysis.

### DIFF
--- a/src/check.zig
+++ b/src/check.zig
@@ -82,19 +82,9 @@ pub fn run(vm: *VM, path: []const u8, opts: Options) u8 {
 
     var ctx: check_lint.Context = .{ .arena = arena, .user_defined = &user_defined };
 
-    // `check` discards bytecode, so folding buys nothing — and disabling it keeps
-    // every call visible to the lint (no `(car 5 6)` folded away before the walk).
-    const saved_opt = ir_mod.optimize_enabled;
-    ir_mod.optimize_enabled = false;
-    check_lint.active = &ctx;
-    defer {
-        check_lint.active = null;
-        ir_mod.optimize_enabled = saved_opt;
-    }
+    analyzeSource(vm, &ctx, source, path);
 
-    analyze(vm, &ctx, arena, source, path);
-
-    std.mem.sort(check_lint.Finding, ctx.findings.items, {}, findingLess);
+    sortFindings(ctx.findings.items);
     report(arena, ctx.findings.items, opts);
 
     var errors: usize = 0;
@@ -119,6 +109,17 @@ fn isError(code: Code, deny_warnings: bool) bool {
 /// off, lint active, no file I/O or reporting). The caller owns `ctx.arena` and
 /// inspects `ctx.findings` directly.
 pub fn analyzeForTest(vm: *VM, ctx: *check_lint.Context, source: []const u8) void {
+    analyzeSource(vm, ctx, source, "<test>");
+}
+
+/// Analyse `source` (read + env-setup + compile + lint) into `ctx.findings`, with
+/// IR optimization off and the lint collector installed for the duration. Shared
+/// by `run`, `analyzeForTest`, and the language server (src/kaappi_lsp.zig), so
+/// the three surfaces diagnose identical source identically — the LSP must not
+/// drift from `kaappi check` (kaappi#1981).
+pub fn analyzeSource(vm: *VM, ctx: *check_lint.Context, source: []const u8, path: []const u8) void {
+    // Optimization is off so folding never hides a call from the lint walk (no
+    // `(car 5 6)` folded away before it is judged).
     const saved_opt = ir_mod.optimize_enabled;
     ir_mod.optimize_enabled = false;
     check_lint.active = ctx;
@@ -126,7 +127,12 @@ pub fn analyzeForTest(vm: *VM, ctx: *check_lint.Context, source: []const u8) voi
         check_lint.active = null;
         ir_mod.optimize_enabled = saved_opt;
     }
-    analyze(vm, ctx, ctx.arena, source, "<test>");
+    analyze(vm, ctx, ctx.arena, source, path);
+}
+
+/// Sort findings by (line, col) so every report surface agrees on ordering.
+pub fn sortFindings(items: []check_lint.Finding) void {
+    std.mem.sort(check_lint.Finding, items, {}, findingLess);
 }
 
 fn analyze(vm: *VM, ctx: *check_lint.Context, arena: std.mem.Allocator, source: []const u8, path: []const u8) void {
--- a/src/check.zig
+++ b/src/check.zig
@@ -84,7 +84,6 @@ pub fn run(vm: *VM, path: []const u8, opts: Options) u8 {
 
     analyzeSource(vm, &ctx, source, path);
 
-    sortFindings(ctx.findings.items);
     report(arena, ctx.findings.items, opts);
 
     var errors: usize = 0;
@@ -116,7 +115,9 @@ pub fn analyzeForTest(vm: *VM, ctx: *check_lint.Context, source: []const u8) voi
 /// IR optimization off and the lint collector installed for the duration. Shared
 /// by `run`, `analyzeForTest`, and the language server (src/kaappi_lsp.zig), so
 /// the three surfaces diagnose identical source identically — the LSP must not
-/// drift from `kaappi check` (kaappi#1981).
+/// drift from `kaappi check` (kaappi#1981). Findings are left sorted by
+/// (line, col) so every caller gets a deterministic order without having to
+/// remember to sort.
 pub fn analyzeSource(vm: *VM, ctx: *check_lint.Context, source: []const u8, path: []const u8) void {
     // Optimization is off so folding never hides a call from the lint walk (no
     // `(car 5 6)` folded away before it is judged).
@@ -128,11 +129,7 @@ pub fn analyzeSource(vm: *VM, ctx: *check_lint.Context, source: []const u8, path
         ir_mod.optimize_enabled = saved_opt;
     }
     analyze(vm, ctx, ctx.arena, source, path);
-}
-
-/// Sort findings by (line, col) so every report surface agrees on ordering.
-pub fn sortFindings(items: []check_lint.Finding) void {
-    std.mem.sort(check_lint.Finding, items, {}, findingLess);
+    std.mem.sort(check_lint.Finding, ctx.findings.items, {}, findingLess);
 }
 
 fn analyze(vm: *VM, ctx: *check_lint.Context, arena: std.mem.Allocator, source: []const u8, path: []const u8) void {
--- a/src/kaappi_lsp.zig
+++ b/src/kaappi_lsp.zig
@@ -166,8 +166,11 @@ fn readMessage(allocator: std.mem.Allocator) ?[]u8 {
                 while (cl_end < headers.len and headers[cl_end] >= '0' and headers[cl_end] <= '9') : (cl_end += 1) {}
                 // A non-numeric Content-Length corrupts only this frame's own
                 // framing. Return an empty (unparseable) body so the caller
-                // skips it and resynchronises on the next header instead of
-                // ending the whole session (kaappi#1980).
+                // skips it instead of ending the whole session; the next
+                // readMessage then re-reads the leftover bytes and best-effort
+                // resynchronises on the next `Content-Length` header (a body
+                // that itself contains that literal can still mislead the
+                // resync) (kaappi#1980).
                 content_length = std.fmt.parseInt(usize, headers[cl_start..cl_end], 10) catch {
                     return allocator.alloc(u8, 0) catch return null;
                 };
@@ -807,12 +810,29 @@ fn runDiagnostics(allocator: std.mem.Allocator, vm: *vm_mod.VM, uri: []const u8,
 
     var ctx: check_lint.Context = .{ .arena = arena, .user_defined = &user_defined };
     check.analyzeSource(vm, &ctx, text, uri);
-    check.sortFindings(ctx.findings.items);
+
+    // Serialize every finding through the shared LSP `Diagnostic` writer into an
+    // allocating writer (no fixed buffer), exactly as `check.zig`'s `reportJson`
+    // does, so an arbitrarily long finding message is emitted rather than
+    // silently dropped. The comma separator is gated on the buffer being past the
+    // opening `[`, so a finding that fails to serialize can never leave a `[,` or
+    // `,]` that would corrupt the whole array (kaappi#1981 review).
+    var aw: std.Io.Writer.Allocating = .init(arena);
+    defer aw.deinit();
 
     diag_buf.append(allocator, '[') catch return;
-    for (ctx.findings.items, 0..) |finding, i| {
-        if (i > 0) diag_buf.append(allocator, ',') catch {};
-        appendFinding(&diag_buf, allocator, finding);
+    for (ctx.findings.items) |finding| {
+        if (diag_buf.items.len > 1) diag_buf.append(allocator, ',') catch {};
+        aw.clearRetainingCapacity();
+        var cbuf: [diagnostics.Code.render_width]u8 = undefined;
+        const diag: lsp_diagnostic.Diagnostic = .{
+            .range = lsp_diagnostic.spanRange(finding.span),
+            .severity = lsp_diagnostic.severityOf(finding.code.info().severity),
+            .code = finding.code.render(&cbuf),
+            .message = finding.message,
+        };
+        diag.writeJson(&aw.writer) catch continue;
+        diag_buf.appendSlice(allocator, aw.written()) catch return;
     }
     diag_buf.append(allocator, ']') catch return;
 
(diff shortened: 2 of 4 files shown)


src/kaappi_lsp.zig
```diff
@@ -918,117 +937,19 @@ fn endOutputRedirect(vm: *vm_mod.VM, rd: OutputRedirect) void {
...
 // (src/lsp_diagnostic.zig) — the same code path `--diagnostics=json` uses, so
-// the CLI and the server can never drift (kaappi#1505). The range spans the
-// whole line (character 0..999) to highlight it in an editor; the `code` is the
-// stable KP code and `message` its registry template.
-fn addDiagnostic(buf: *std.ArrayList(u8), allocator: std.mem.Allocator, line0: u32, code: diagnostics.Code) void {
+// the CLI and the server can never drift (kaappi#1505). The range is the
+// finding's real span, converted to LSP's zero-based coordinates via
+// `spanRange`, so a nested error is pinpointed to its characters rather than
+// painted as a whole-line 0..999 sentinel (kaappi#1981).
+fn appendFinding(buf: *std.ArrayList(u8), allocator: std.mem.Allocator, f: check_lint.Finding) void {
```
baijum: **[Blocking] Fixed 1024-byte serializer silently drops long findings — and corrupts the array.**

`appendFinding` serializes through `std.Io.Writer.fixed(&tmp)` with a 1024-byte buffer and `catch return`s on overflow. Before this PR `message` was the short registry template; now it is the finding's real message, whose length is unbounded: KP4001 embeds the identifier verbatim (`"unknown variable '{s}' at top level"` in `check_lint.zig:177`), and compile-error details run up to 512 bytes (`compiler.syntax_error_detail`) with JSON escaping on top. A legal ~900+ character identifier overflows the buffer (envelope is ~120 bytes).

Both failure modes reproduced against this branch:

1. **Single finding dropped** → `"diagnostics":[]`. DidOpen `(display aaa…1000 a)` publishes an empty array while `kaappi check --diagnostics=json` reports the KP4001 warning — the #1981 defect class again, at a new boundary.
2. **Two findings, either one dropped** → invalid JSON. The comma is appended at line 814 *before* this function runs, so a drop leaves `[,` or `,]`. DidOpen `(display aaa…1000 a)\n(display short-undef)` produces `"diagnostics":[,{"range":…` — the whole frame fails to parse, so the editor shows *nothing* for the document.

Suggested fix: serialize into an allocating writer the way `check.zig`'s `reportJson` already does — `std.Io.Writer.Allocating` on the per-run arena, then `buf.appendSlice(allocator, aw.written())` — so the LSP path has the same unbounded guarantee the CLI path has. The `code` slice into `cbuf` stays valid since `writeJson` consumes it synchronously.

src/kaappi_lsp.zig
```diff
@@ -770,31 +791,29 @@ fn runDiagnostics(allocator: std.mem.Allocator, vm: *vm_mod.VM, uri: []const u8,
...
-        // (kaappi#1980, deliberately covered in the LSP suite).
-        if (stop) break;
-    }
+    var ctx: check_lint.Context = .{ .arena = arena, .user_defined = &user_defined };
+    check.analyzeSource(vm, &ctx, text, uri);
+    check.sortFindings(ctx.findings.items);
 
+    diag_buf.append(allocator, '[') catch return;
+    for (ctx.findings.items, 0..) |finding, i| {
+        if (i > 0) diag_buf.append(allocator, ',') catch {};
```
baijum: Worth hardening alongside the writer fix: the comma is appended before `appendFinding` has proven the finding will serialize, so *any* append failure (overflow today, OOM after the fix) yields `[,` or `,]` and the whole frame becomes unparseable. Gating on the buffer instead of the index — `if (diag_buf.items.len > 1) append(',')` — keeps the array well-formed no matter what a single append does.

src/kaappi_lsp.zig
```diff
@@ -149,13 +164,21 @@ fn readMessage(allocator: std.mem.Allocator) ?[]u8 {
                 const cl_start = cl_pos + 16;
                 var cl_end = cl_start;
                 while (cl_end < headers.len and headers[cl_end] >= '0' and headers[cl_end] <= '9') : (cl_end += 1) {}
-                content_length = std.fmt.parseInt(usize, headers[cl_start..cl_end], 10) catch return null;
+                // A non-numeric Content-Length corrupts only this frame's own
+                // framing. Return an empty (unparseable) body so the caller
+                // skips it and resynchronises on the next header instead of
+                // ending the whole session (kaappi#1980).
+                content_length = std.fmt.parseInt(usize, headers[cl_start..cl_end], 10) catch {
```
baijum, line 174: Nit: the comment's "resynchronises on the next header" is a little stronger than what's implemented. Recovery works by leaving the skipped frame's body bytes in the stream and letting the next `readMessage` `indexOf("Content-Length: ")` over them — so a body containing that literal hijacks the resync. Verified on this branch: `Content-Length: abc\r\n\r\n{"evil":"Content-Length: 999"}` followed by a well-formed `initialize` made the server consume 999 bytes as a body; the real initialize never parsed and nothing was answered. The suite's `Content-Length: abc` case survives only because its `{}` body carries no such text. If it's worth hardening: accept `Content-Length:` only at a line start (preceded by `\n` or the buffer start). Otherwise I'd soften the comment to "best-effort resync" so the guarantee isn't oversold.

src/kaappi_lsp.zig
```diff
@@ -1185,13 +1112,24 @@ pub fn main(init: std.process.Init.Minimal) !void {
...
         if (root.get("id")) |id_val| {
-            _ = formatIdValue(&id_buf, allocator, id_val);
+            id_valid = formatIdValue(&id_buf, allocator, id_val);
+        }
+        const id: ?[]const u8 = if (id_valid) id_buf.items else null;
+
+        // A request whose id we cannot echo is answered immediately with an
+        // Invalid Request (id null) and never dispatched.
+        if (root.get("id") != null and !id_valid) {
+            sendError(allocator, "null", -32600, "Invalid Request");
```
baijum, line 1132: Nit: the two `-32600` spellings differ — `"Invalid Request"` here, `"InvalidRequest"` in the post-shutdown branches (lines 1144/1169). JSON-RPC's spec examples use "Invalid Request"; either works, but a client string-matching on the message would appreciate one spelling across all three sites.

baijum (commented):
Reviewed, built, and exercised this branch: `zig build` clean, `zig fmt --check` clean, `tests/scheme/lsp/lsp.sh` 184/184, `zig build test` green (one first run failed mid-suite with no test failure — re-ran clean twice; looks like machine contention, not the PR).

**One blocking issue**, found by probing the new shared-diagnostics path: a finding whose serialized JSON exceeds 1024 bytes is silently dropped, and with two or more findings the drop corrupts the `publishDiagnostics` array into unparseable JSON. Reproduced against this branch's `kaappi-lsp` — details and a concrete trigger inline. It's reachable from ordinary Scheme: a legal identifier of ~900+ characters does it, because KP4001 embeds the identifier in the message. Ironically the single-finding variant is exactly the #1981 defect class this PR closes ("editor shows clean where `check` reports"), reintroduced at a different boundary.

Everything else holds up well. The `analyzeSource` consolidation is the structurally right fix for #1981 — it deletes the hand-mirrored `diagnoseTopLevelForm` classifier rather than patching it a third time, so check-side fixes can no longer miss the LSP. The lifecycle state machine now matches LSP 3.17 on all the edges the suite pins, and enabling the previously-disabled assertions each beside a control is exactly the house style. Remaining comments are nits.

src/kaappi_lsp.zig
```diff
@@ -1202,18 +1140,33 @@ pub fn main(init: std.process.Init.Minimal) !void {
             null;
 
         if (std.mem.eql(u8, method, "initialize")) {
-            handleInitialize(allocator, id orelse "0");
-            initialized = true;
+            if (shutdown_requested) {
+                if (id) |req_id| sendError(allocator, req_id, -32600, "InvalidRequest");
+            } else {
+                handleInitialize(allocator, id orelse "0");
```
baijum: Pre-existing, but it cuts against this PR's own "never a fabricated id" principle: an `initialize` with no id at all (i.e. sent as a notification) still gets a response with a fabricated id `0` — a reply to something that asked no question. The `shutdown` branch already handles its notification case by skipping `sendResponse`. Consider dropping the `orelse "0"` and skipping the reply when `id == null`, or leaving as-is with a note. Not blocking.

src/check.zig
```diff
@@ -119,14 +109,30 @@ fn isError(code: Code, deny_warnings: bool) bool {
...
 pub fn analyzeForTest(vm: *VM, ctx: *check_lint.Context, source: []const u8) void {
+    analyzeSource(vm, ctx, source, "<test>");
+}
+
+/// Analyse `source` (read + env-setup + compile + lint) into `ctx.findings`, with
+/// IR optimization off and the lint collector installed for the duration. Shared
+/// by `run`, `analyzeForTest`, and the language server (src/kaappi_lsp.zig), so
+/// the three surfaces diagnose identical source identically — the LSP must not
+/// drift from `kaappi check` (kaappi#1981).
+pub fn analyzeSource(vm: *VM, ctx: *check_lint.Context, source: []const u8, path: []const u8) void {
```
baijum, line 121: This is the right shape for the fix — the old `diagnoseTopLevelForm` was a hand-maintained mirror of `checkForm` (its own comment said so), so every check-side fix was one more chance for the LSP to drift. Sharing `analyzeSource` deletes the whole class, and the ~110-line deletion is the proof.

One optional thought: both `run` and the LSP must remember to call `sortFindings` after `analyzeSource` — forgetting it would silently produce unsorted diagnostics on one surface. Folding the sort into `analyzeSource` would make the ordering invariant impossible to forget; `tests_check.zig` asserts which findings appear, not their order, so nothing breaks.

tests/scheme/lsp/lsp.sh
```diff
@@ -538,15 +539,14 @@ did_open "file://$TMP/two.scm" '(if)\n(let)\n'
 msg "$EXITN"
 lsp_run
 TWO_CHK="$(run_timeout 30 "$KAAPPI" check --diagnostics=json "$TMP/two.scm" 2> /dev/null)"
-# Control: `check` really does see both errors, so the LSP's count is the outlier.
+# Control: `check` really does see both errors, and so does the LSP.
 assert_eq "multi-error control: check reports 2 diagnostics" "$(count "$TWO_CHK" '"severity":')" "2"
 assert_eq "multi-error control: lsp reports the first error's code" "$(first_code "$OUT")" "KP2001"
-# FAIL: #1980 (LSP stops at the first bad form; check reports every one)
-# assert_eq "multi-error: lsp reports both diagnostics" "$(count "$OUT" '"severity":')" "2"
+assert_eq "multi-error: lsp reports both diagnostics" "$(count "$OUT" '"severity":')" "2"
```
baijum, line 545: Could this section gain a sibling regression for the long-message case (see my blocking comment on `appendFinding`)? Concretely: a document with an undefined identifier of ~1000 chars plus a second short undefined name, asserting the `publishDiagnostics` payload is parseable JSON carrying 2 diagnostics — e.g. by piping `$OUT` through a JSON parse of the frame rather than `count`. On this branch that input emits `"diagnostics":[,{…`, so today's `count '"severity":'`-style assertions can't catch it (they'd pass on the dropped-finding count of 1 and never notice the frame doesn't parse).

baijum:
Thanks for the review. Addressed in `475f5a7c`:

**Blocking (fixed):** the fixed 1024-byte serializer is gone. `runDiagnostics` now serializes each finding through `std.Io.Writer.Allocating` on the per-run arena (the same shape `check.zig`'s `reportJson` uses), and the comma separator is gated on `diag_buf.items.len > 1` instead of the index — so an arbitrarily long finding is emitted, and a finding that fails to serialize can never leave a `[,`/`,]`. Added the `long-message` regression in `tests/scheme/lsp/lsp.sh` (a ~1000-char identifier + a short one, asserting both KP4001 findings are published and the array isn't corrupted). Suite is now 187/187.

**Nits (fixed):** folded the finding sort into `analyzeSource` (callers can't forget it); unified the `-32600` message on `"Invalid Request"`; no fabricated `id: 0` for an `initialize` sent as a notification; softened the header-resync comment to "best-effort"; documented that `kaappi check` adds severity-2 warnings (KP4001).

**Skipped (CodeRabbit `requestParams` position validation):** rejecting negative positions with `-32602` would contradict #1980's "Confirmed correct" note that "negative and huge positions all degrade cleanly", which `lsp.sh` already pins (`position: negative line/character clamps to null, no crash`). I left the structural validation (`textDocument`/`uri`/`position` object present) as-is and kept value-level degradation to `0`.

**CI note:** the cancelled job is `test (macos-latest, ReleaseSafe)` hitting its 30-minute `timeout-minutes` during the "Differential tiers" suite — *after* `lsp.sh` had already passed. It's the known macOS recurrence noted in `ci.yml` ("the macOS/PR #1635 recurrence"); the base commit `d4b5e7e`'s own CI run also failed (on `test (ubuntu-latest, Debug)`). Not caused by this change.

baijum:
LGTM — verified on 475f5a7c.

**Blocking issue confirmed fixed.** Re-ran both repros against the rebuilt binary: the two-finding document (1000-char identifier + short-undef) now publishes a *parseable* frame with both KP4001 diagnostics — the 1032-byte message included, which is exactly what overflowed the old fixed writer — and the single-finding document publishes its warning instead of an empty array. Serializing through an allocating writer is the same guarantee `reportJson` has, so the two surfaces now match on long messages too.

All the nits are addressed and spot-checked: notification-`initialize` gets no reply and does not complete the handshake (a following hover correctly answers `-32002`); the `-32600` message is unified on "Invalid Request"; the sort folded into `analyzeSource`; the resync comment now says best-effort and names the misleading-body caveat; the docs severity row matches what `check` actually emits.

Suites: `tests/scheme/lsp/lsp.sh` 187/187 (including the new long-message regression), `zig build test` 1810/1817 (7 skipped), `zig fmt --check` clean.

One non-blocking footnote: the new comment in `runDiagnostics` says a failed serialize "can never leave a `[,` or `,]`" — `[,` is indeed impossible now (the comma requires a finding already in the buffer), but a trailing `,]` is still reachable if the *last* finding's `writeJson` OOMs after its comma was appended. That's an OOM-only path and under OOM the publish is likely lost anyway, so fine to ship — appending the comma *after* the successful `appendSlice` would close it if this is ever touched again.

PR #2298, opened 2026-08-24
Fix WASM file-backed .sld loading and command-line/lib-path setup (Fixes #2108, #2109)

### LINKED ISSUE #2108: wasm32: no file-backed .sld is importable even when the host mounts it — platform.openRead has no WASI branch, so resolveLibraryPath's probe always fails

### LINKED ISSUE #2109: [R7RS 6.14] wasm32: (command-line) is '() and vm.lib_paths is empty — the WASM entry returns before both are set
**Category:** Tier divergence (WASM) / R7RS conformance · **Severity:** wrong-result · **Found by:** systematic audit v2, Phase 4D (#1890)

`main.zig`'s WASM entry path returns before the shared post-CLI initialisation block, so two `VM` fields that every other target populates keep their empty defaults on wasm32: `vm.command_line_args` and `vm.lib_paths`.

## 1. `(command-line)` returns the empty list

```scheme
(import (scheme base) (scheme write) (scheme process-context))
(write (command-line)) (newline)
```

| | native aarch64 | **wasm32 (wasmtime 46.0.0)** |
|---|---|---|
| `(command-line)` | `("_t5.scm")` | **`()`** |
| `(length (command-line))` | `1` | **`0`** |

R7RS §6.14, page 59, exact text:

> **`(command-line)`** — process-context library procedure. Returns the command line passed to the process as a list of strings. **The first string corresponds to the command name**, and is implementation-dependent. It is an error to mutate any of these strings.

The command name is implementation-dependent, but the spec describes a list that *has* a first string. `()` has none, so `(car (command-line))` — the ordinary way to reach the program name, and the shape SRFI 193 builds on — errors on wasm32 alone.

This already breaks a corpus file: `tests/scheme/smoke/command-line-o-flag.scm` (the regression test for #602) prints `all passed` natively and `FAIL: (command-line) too short` on WASM.

## 2. `vm.lib_paths` is never populated

The script's own directory is never added to the library search path, so a `.sld` sitting beside the program is invisible:

```
_libtest/user.scm          (import (scheme base) (scheme write) (mylib hello))
_libtest/mylib/hello.sld   (define-library (mylib hello) ...)
```

```
$ ./zig-out/bin/kaappi _libtest/user.scm
hello-from-sld
$ wasmtime run --dir=. zig-out/bin/kaappi.wasm _libtest/user.scm
_libtest/user.scm:1: error[KP2001]: library not found: (mylib.hello)
```

**Discriminating control:** neither `mylib/hello.sld` nor `lib/mylib/hello.sld` exists relative to the cwd, so `resolveLibraryPath`'s built-in prefixes `{"", "lib/"}` cannot be what finds it natively — the script-directory entry in `vm.lib_paths` is the only route, and it is exactly the entry WASM never gets.

This is **separate from #2108** and survives it: #2108 is the `platform.openRead` WASI gap, which once fixed makes the cwd-relative `{"", "lib/"}` prefixes work. A sibling `.sld` reached only via the script's own directory still needs this.

## Root cause

`main.zig:242-257` handles WASM and returns:

```zig
if (comptime is_wasm) {
    try primitives.registerAll(vm);
    ...
    const file_path = wasi_args.next() orelse { ... };
    try runFile(vm, file_path);
    return;                                  // <- everything below is skipped
}
```

Both assignments live below it:

- `vm.command_line_args = opts.scriptArgs();` — `main.zig:305`
- `vm.lib_paths = lib_paths[0..lib_path_count];` — `main.zig:499`

Their defaults are `&.{}` (`vm.zig:378-379`), which is what wasm32 keeps.

**This looks like an oversight rather than a design choice**, because the skipped block already contains code written specifically for the WASM case. `main.zig:442` opens with:

```zig
if (!is_wasm) {
    // ~/.kaappi/lib and the DYLD/LD_LIBRARY_PATH export
```

A `!is_wasm` guard *inside* a block that WASM can never reach is dead by construction — it only makes sense if the author expected the surrounding block to run on this target. The script-directory append immediately above it (`main.zig:435-439`) carries no guard at all.

## Suggested direction

Populate both fields on the WASM path from the WASI argv the branch already iterates, rather than duplicating the CLI parser: the file path is in hand at `main.zig:251`, which is all the script-directory entry and a one-element `command-line` need.

**Verified:** fresh `zig build` + `zig build wasm` at `3c1a56be` (v0.22.1), macOS aarch64, wasmtime 46.0.0, `KAAPPI_HOME` isolated, run from the repo root with `--dir=.`.
**Source location:** `src/main.zig:242-257` (early return), `:305` (`command_line_args`), `:499` (`lib_paths`), `:442` (the dead `if (!is_wasm)`); defaults at `src/vm.zig:378-379`.
**Related:** #2108 (the other, independent reason no `.sld` loads on WASM).

**Found by:** Systematic audit v2, Phase 4D (WASM cross-tier)

baijum:
Fixes #2108 and fixes #2109 — two independent wasm32 tier divergences from the audit v2 Phase 4D sweep, fixed together in one PR because both break library loading on the WASM tier and share the same regression harness.

## #2108 — no file-backed `.sld` is importable

`platform.openRead` had no WASI branch, so `resolveLibraryPath`'s existence probe failed for every candidate path and no file-backed `.sld` could be imported — even when the host had the directory mounted and the file was demonstrably readable (the same path read fine as a *script* but would not resolve as a *library*).

The fix gives `openRead` the same preopened-dir (fd 3) `path_open` branch `file_utils.readWholeFile` already had, so the probe finds `lib/srfi/*.sld` under the existing `"lib/"` prefix with no other change.

## #2109 — `(command-line)` is `'()` and `vm.lib_paths` is empty

`main.zig`'s WASM entry returned before the shared post-CLI block that populates `vm.command_line_args` and `vm.lib_paths`, so both kept their empty defaults. `(command-line)` — which R7RS 6.14 says has a first string — returned `()`, and a `.sld` beside the program was invisible (the script's own directory is the only route to it).

The fix repopulates both fields from the WASI argv the branch already iterates, rather than duplicating the CLI parser.

## Tests

- `tests/wasm/library-load.scm` — imports `(srfi 2)` (file-backed `.sld` via `lib/`) and a sibling fixture `.sld` (via the script-directory `lib_paths` entry), wired into the CI `wasm` job.
- `tests/wasm/command-line.scm` — asserts `(command-line)` is non-empty and names the script, wired into the CI `wasm` job.
- Removed the `command-line-o-flag.scm` and `condexpand-include-lib-868-879.scm` `KNOWN_DIFFS` entries from `run-wasm-differential.sh` now that the tiers agree again.
- Updated the stale #2108/#2109 references in the differential header, `docs/dev/testing.md`, `tests/scheme/CLAUDE.md`, and the three exempt probe-file headers.

## Verification

- `zig build wasm` + `zig build` clean; `zig fmt --check` clean.
- `zig build test` green.
- `wasmtime run --dir=.` passes `library-load.scm`, `command-line.scm`, and all four pre-existing `tests/wasm/*` programs.
- `run-wasm-differential.sh` — **0 divergences, 0 hangs** across 368 corpus files; `LIBDIFF` drops from ~231 to 48, where the remaining 48 are the libraries deliberately refused on this tier by `Lib.wasmAvailable()` (`(srfi 18)`, `(srfi 170)`, `(srfi 192)`, `(kaappi ffi)` — no OS threads/filesystem/FFI), not file-backed `.sld` failures.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -289,7 +289,7 @@ fn mainImpl(init: std.process.Init.Minimal) !void {
         var cmd_args: std.ArrayList([]const u8) = .empty;
         defer cmd_args.deinit(allocator);
         while (wasi_args.next()) |arg| {
-            cmd_args.append(allocator, arg) catch return;
+            try cmd_args.append(allocator, arg);
         }
         if (cmd_args.items.len == 0) {
             writeStderr("kaappi-wasm: no file specified\n");
--- a/src/main.zig
+++ b/src/main.zig
@@ -278,10 +278,34 @@ fn mainImpl(init: std.process.Init.Minimal) !void {
         var wasi_args = try init.args.iterateAllocator(allocator);
         defer wasi_args.deinit();
         _ = wasi_args.skip(); // skip argv[0]
-        const file_path = wasi_args.next() orelse {
+
+        // The non-WASM path below populates vm.command_line_args (from
+        // opts.scriptArgs()) and vm.lib_paths (script dir + --lib-path +
+        // auto-discovered dirs), but this branch returns before either, so
+        // both keep their empty defaults. Repopulate them from the WASI argv
+        // the branch already iterates: (command-line) must report the script
+        // path (R7RS 6.14) and a sibling .sld must resolve via the script's
+        // own directory. (kaappi#2109)
+        var cmd_args: std.ArrayList([]const u8) = .empty;
+        defer cmd_args.deinit(allocator);
+        while (wasi_args.next()) |arg| {
+            cmd_args.append(allocator, arg) catch return;
+        }
+        if (cmd_args.items.len == 0) {
             writeStderr("kaappi-wasm: no file specified\n");
             return;
-        };
+        }
+        const file_path = cmd_args.items[0];
+        vm.command_line_args = cmd_args.items;
+
+        var script_dir_buf: [1][]const u8 = undefined;
+        var script_dir_count: usize = 0;
+        if (std.mem.lastIndexOfScalar(u8, file_path, '/')) |pos| {
+            script_dir_buf[0] = if (pos == 0) file_path[0..1] else file_path[0..pos];
+            script_dir_count = 1;
+        }
+        vm.lib_paths = script_dir_buf[0..script_dir_count];
+
         try runFile(vm, file_path);
         return;
     }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1232,7 +1232,13 @@ jobs:
       - name: Platform gate diagnostics (kaappi#1972)
         run: wasmtime run --dir=. zig-out/bin/kaappi.wasm tests/wasm/platform-gates.scm
 
-      # The four steps above are hand-written WASM-only programs that assert
+      - name: File-backed library loading (kaappi#2108, #2109)
+        run: wasmtime run --dir=. zig-out/bin/kaappi.wasm tests/wasm/library-load.scm
+
+      - name: (command-line) reports the script path (kaappi#2109)
+        run: wasmtime run --dir=. zig-out/bin/kaappi.wasm tests/wasm/command-line.scm
+
+      # The steps above are hand-written WASM-only programs that assert
       # their own results. The differential below instead runs the SHARED
       # corpus on both tiers and requires the WASM output to match the
       # interpreter's byte-for-byte, which is what catches a WASM build that
--- a/docs/dev/testing.md
+++ b/docs/dev/testing.md
@@ -276,12 +276,12 @@ above.
 **Four** files in the corpus are exempt and say so in their own headers —
 `smoke/large-index-bounds-1912.scm`, `smoke/deep-nesting-print.scm`,
 `smoke/deep-nesting-print-tier-margin.scm` and
-`continuations/coroutine-repl-echo.scm`. The first three must stay free of
-file-backed `.sld` imports because `(import (srfi 64))` fails at library load
-on WASM (kaappi#2108) and `run-wasm-differential.sh` would reclassify them as
-LIBDIFF; the fourth must leave its top-level forms bare because consuming their
-values is what hid the bug it exists to catch. All four carry a hand-rolled
-`(exit 1)` instead.
+`continuations/coroutine-repl-echo.scm`. The first three are cross-tier probes
+whose bare top-level forms are the measurement, so they stay import-free to run
+identically on every tier rather than being wrapped in a SRFI-64 suite; the
+fourth must leave its top-level forms bare because consuming their values is
+what hid the bug it exists to catch. All four carry a hand-rolled `(exit 1)`
+instead.
 
 `tests/scheme/CLAUDE.md` holds the single inventory table (56 verdictless
 files → 52 converted + 4 exempt, plus `fiber-error-handling.scm` = 53
--- a/src/platform.zig
+++ b/src/platform.zig
@@ -218,6 +218,19 @@ fn winOpen(path: []const u8, oflag: c_int, pmode: c_int) OpenError!fd_t {
 
 pub fn openRead(path: [:0]const u8) OpenError!fd_t {
     if (comptime is_windows) return winOpen(path, win.O_RDONLY, 0);
+    if (comptime is_wasm) {
+        // WASI has no AT.FDCWD: paths resolve only relative to a preopened
+        // directory fd. fd 3 is the first preopened dir — the working
+        // directory that both wasmtime `--dir=.` and the browser shim mount —
+        // the same convention file_utils.readWholeFile already uses. Without
+        // this branch, resolveLibraryPath's existence probe failed for every
+        // candidate path and no file-backed .sld was importable on wasm32
+        // even when the host mounted the directory (kaappi#2108).
+        var result_fd: std.os.wasi.fd_t = undefined;
+        const rc = std.os.wasi.path_open(3, .{ .SYMLINK_FOLLOW = true }, path.ptr, path.len, .{}, .{ .FD_READ = true, .FD_SEEK = true }, .{ .FD_READ = true }, .{}, &result_fd);
+        if (rc != .SUCCESS) return error.OpenFailed;
+        return @as(fd_t, @intCast(result_fd));
+    }
     return std.posix.openatZ(std.posix.AT.FDCWD, path, .{}, 0) catch error.OpenFailed;
 }
 
--- a/src/vm_library.zig
+++ b/src/vm_library.zig
@@ -339,8 +339,10 @@ pub fn libraryIsAvailable(vm: *VM, lib_name: []const u8, lib_name_list: Value) b
     // mount at a given relative path -- unlike a normal install, there's no
     // guarantee lib/ is reachable there. Preferring the embedded copy when
     // there is one sidesteps that packaging dependency for exactly this
-    // library; anything else still falls through to the normal disk check
-    // (unaffected -- a host that does mount lib/ still serves those fine).
+    // library; anything else still falls through to the normal disk check,
+    // which resolves against the preopened directory since platform.openRead
+    // grew its WASI branch (kaappi#2108) -- so a host that does mount lib/
+    // serves those fine.
     if (is_wasm) {
         var path_buf: [512]u8 = undefined;
         if (buildLibRelPath(lib_name_list, &path_buf) catch null) |rel_path| {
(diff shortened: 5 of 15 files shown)


src/main.zig
```diff
@@ -278,10 +278,34 @@ fn mainImpl(init: std.process.Init.Minimal) !void {
...
+        // opts.scriptArgs()) and vm.lib_paths (script dir + --lib-path +
+        // auto-discovered dirs), but this branch returns before either, so
+        // both keep their empty defaults. Repopulate them from the WASI argv
+        // the branch already iterates: (command-line) must report the script
+        // path (R7RS 6.14) and a sibling .sld must resolve via the script's
+        // own directory. (kaappi#2109)
+        var cmd_args: std.ArrayList([]const u8) = .empty;
+        defer cmd_args.deinit(allocator);
+        while (wasi_args.next()) |arg| {
+            cmd_args.append(allocator, arg) catch return;
```
baijum: `catch return` here swallows OOM and returns from `mainImpl` **normally** — exit code 0 — without ever calling `runFile`. The process then reports success while never running the requested script, which is the one outcome a script runner must never produce silently.

Every other allocation in this branch propagates with `try` (`iterateAllocator` two lines up, `registerAll`, `install`, etc.), so this is also inconsistent with the surrounding style. Since `mainImpl` is `!void`, propagate the error instead:

```zig
while (wasi_args.next()) |arg| {
    try cmd_args.append(allocator, arg);
}
```

That way an allocation failure surfaces as a non-zero exit like every other failure path, rather than a green run with no output.

---

baijum (commented):
Reviewed the WASM library-loading and command-line fixes. The two core fixes are correct and well-targeted:

- **`platform.openRead` WASI branch** (`src/platform.zig`) faithfully mirrors the existing `file_utils.readWholeFile` pattern (same fd-3 preopened-dir `path_open` call, same rights), so file-backed `.sld` resolution now works through the mounted directory. The `resolveLibraryPath` probe at `vm_library.zig:307-308` correctly closes the returned fd, so no descriptor leak.
- **`vm.lib_paths` script-dir logic** in the WASM branch matches the non-WASM path at `main.zig:495-500` exactly (same `lastIndexOfScalar('/')` + `pos == 0` root handling).
- **Tests** are solid: `library-load.scm` exercises both routes ((srfi 2) via the `lib/` prefix, sibling `.sld` via the script-dir lib path), and the `wasm-fixtures/sibling.sld` fixture deliberately has no `lib/` copy so it can only resolve through the new script-directory entry. Both are wired into the CI `wasm` job. The documentation/comment sweep across the differential harness and probe headers is consistent.

One correctness issue inline (`main.zig`, the `catch return` in the argv loop): OOM during argument collection silently returns exit 0 without running the script. Worth fixing before merge — it's the same concern CodeRabbit's merge-risk note raised.

Minor, non-blocking: `vm.command_line_args`/`vm.lib_paths` are assigned slices backed by the block-local `cmd_args` (freed by its `defer`) and the stack-local `script_dir_buf`. Both stay valid for the duration of `runFile` since it's called within the same block, so this is fine as written — just flagging that these are intentionally shorter-lived than the arena-backed non-WASM equivalents in case future code reads them after `runFile` returns.

---

Issue #2314, opened 2026-08-25
SRFI 231: array-packed? wrongly requires zero body offset — wrong #f on extracted views, reshape fallout


baijum:
### Summary

`array-packed?` requires a specialized array's lexicographic traversal to visit body indices `0, 1, 2, …, volume-1`. The spec and reference only require **increasing and consecutive** indices — any base offset is fine. The result is a silently wrong `#f` for legal arrays whose elements start at a nonzero body offset (e.g. `array-extract` views), and a downstream behavioral divergence in `specialized-array-reshape`.

### Reproduction

```scheme
kaappi> (import (srfi 231))
kaappi> (define a (make-specialized-array (make-interval '#(2) '#(4))))
kaappi> (array-packed? a)                                    ; => #t  (correct)
kaappi> (array-packed? (array-extract a (make-interval '#(3) '#(4))))  ; => #f  (should be #t)
```

The extracted view's single element sits at body index 1; Kaappi demands index 0. Reference: `%%compute-array-packed?` (generic-arrays.scm) checks only stride-1 differences between lexicographic neighbors and treats a length-1 axis as trivially packed → `#t`. The reference's in-order cache starts as a non-boolean `%%order-unknown` sentinel, so this value is genuinely computed, not overridden.

Downstream consequence — reshape of an offset view is rejected where the reference reshapes in place:

```scheme
kaappi> (array-ref (specialized-array-reshape
                     (array-extract a (make-interval '#(3) '#(4)))
                     (make-interval '#(1)))
                    0)
ERROR: specialized-array-reshape: requested reshape is not affinely representable without copying
```

The reference reshapes this view in place (it is packed by its definition; base = first body index).

### Analysis

`lib/srfi/231/arrays.sld:184-187` initializes `expected` to 0 and requires each successive indexer value to equal `expected` then increments — i.e. a zero base is mandatory. The comment at `arrays.sld:171-179` documents that stricter reading ("visits body positions 0,1,2,...,volume-1"), which does not match the spec: "returns `#t` if the elements of array, taken in lexicographical order, are stored in `(array-body array)` with **increasing and consecutive indices**" (SRFI 231, `array-packed?` entry).

Two corroborating points:
- The port's own `specialized-array-reshape` fast path at `lib/srfi/231/views.sld:350` already computes `base` from the first indexer value — the rest of the port assumes the correct semantics; fixing `array-packed?` aligns reshape too (offset views then reshape in place, like the reference, instead of erroring/copying).
[...]

PR #2319, opened 2026-08-25
Validate SRFI 231 entry-point arguments to match the reference (#2312 #2313 #2315 #2316 #2317 #2318)

### LINKED ISSUE #2312: SRFI 231: interval-contains-multi-index? silently returns #f on dimension mismatch (reference errors)
### Summary

`interval-contains-multi-index?` silently returns `#f` when the number of indices does not match the interval's dimension, instead of signaling an error. The SRFI 231 spec makes this an error situation, and the reference implementation raises for it — this exact expression is an error test in the reference test suite (test-arrays.scm).

### Reproduction

```scheme
kaappi> (import (srfi 231))
kaappi> (interval-contains-multi-index? (make-interval '#(1 2 3) '#(4 5 6)) 1)
#f
```

Reference test suite expects an error with message `"interval-contains-multi-index?: The dimension of the first argument (an interval) does not match number of indices: "` (test-arrays.scm, "interval-contains-multi-index?  error tests").

### Analysis

`lib/srfi/231/intervals.sld:130` folds the dimension check into the truth chain:

```scheme
(and (= (length multi-index) (vector-length lo)) ...)
```

so a mismatch falls through to `#f`. The spec text says: "Assumes that interval is an interval with dimension $d$ and multi-index is a multi-index (a sequence of exact integers) of length $d$ … It is an error to call interval-contains-multi-index? if interval and multi-index do not satisfy this condition." The reference implementation (generic-arrays.scm) raises explicitly via `cond` checks before computing.

This is also internally inconsistent with the same file's treatment of the sibling case: `interval-subset?` at `lib/srfi/231/intervals.sld:119` deliberately signals an error on a dimension mismatch, with a comment explaining that a mismatch is an error, not a normal `#f` result. The procedure itself already raises for non-exact-integer index components (`intervals.sld:134`), so only the dimension check is missing.

Priority: medium — the calling program is non-conforming (an error situation per spec, any behavior permitted), but the port's neighbors and the reference all treat this as a loud error, and a silent `#f` looks like a legitimate "index not in interval" answer.

### Fix sketch

Replace the `and`-chain dimension check with an explicit check that calls `error` with a message mirroring the reference, e.g. `"interval-contains-multi-index?: dimension of interval does not match number of indices"`. Add a guard-based regression test next to the existing ones in `tests/scheme/srfi/srfi231-intervals.scm:90-96`.

Source: reported via [r/KaappiScheme](https://www.reddit.com/r/KaappiScheme/comments/1vxk61h/some_srfi_231_anomalies/); verified against the [SRFI 231 spec](https://srfi.schemers.org/srfi-231/srfi-231.html) and [reference implementation](https://github.com/scheme-requests-for-implementation/srfi-231).

### LINKED ISSUE #2313: SRFI 231: storage-class-data->body is unvalidated identity for every built-in storage class
### Summary

Every built-in storage class in the SRFI 231 port installs `values` (the identity function) as its `data->body` entry, so `((storage-class-data->body sc) data)` returns any object unchanged — including objects of the wrong type — where the reference implementation validates the data type and raises.

### Reproduction

```scheme
kaappi> (import (srfi 231))
kaappi> ((storage-class-data->body u8-storage-class) 'a)
a
kaappi> ((storage-class-data->body generic-storage-class) 'a)
a
```

Reference raises: `"Expecting a u8vector passed to (storage-class-data->body u8-storage-class): "` (generic-arrays.scm, homogeneous storage-class macro — every built-in class wraps `data->body` in a type check).

### Analysis

`lib/srfi/231/storage-classes.sld` passes `values` as the `data->body` argument for all built-ins (generic at line 68, char 73, s8 80, … u8 97, … c128 123).

The reference source comments the contract explicitly: `data->body: (lambda (data) ...) converts data to a body, raising an exception if needed`, and the homogeneous-class macro generates `(if (,? data) data (error "Expecting a <type> passed to ..."))` for every class. The spec's `make-specialized-array-from-data` entry backs this: `data` "must be an object for which `((storage-class-data? storage-class) data)` returns `#t` … It is an error if the arguments do not satisfy these conditions."

Mitigating: `make-specialized-array-from-data` in the port does its own validation (`arrays.sld:158` — "data is not compatible with storage-class"), so the hole is confined to calling the `storage-class-data->body` accessor's returned procedure directly.

Priority: medium — the calling program is non-conforming per spec; the direct `data->body` escape hatch is a low-level path. But returning garbage that poses as a "body" is worse than the reference's loud error, and custom `make-storage-class` users writing FFI-style code are exactly the audience that reaches for these low-level procedures.

### Fix sketch

Replace `values` with a validating wrapper per class, e.g. built from the class's own `data?` predicate and name:

```scheme
(define (%checked-data->body pred class-name type-name)
  (lambda (data)
    (if (pred data) data (error (string-append "Expecting a " type-name " passed to (storage-class-data->body " class-name "): ") data))))
```

(In this port the body *is* the data for homogeneous classes, so validation is the only missing piece.) Add regression tests in the storage-classes section of `tests/scheme/srfi/srfi231-*.scm`.

Source: reported via [r/KaappiScheme](https://www.reddit.com/r/KaappiScheme/comments/1vxk61h/some_srfi_231_anomalies/); verified against the [SRFI 231 spec](https://srfi.schemers.org/srfi-231/srfi-231.html) and [reference implementation](https://github.com/scheme-requests-for-implementation/srfi-231).

### LINKED ISSUE #2315: SRFI 231: make-array accepts a non-procedure setter (mutable-array? then lies, failure deferred)
### Summary

`make-array` accepts a setter that is neither a procedure nor `#f`. The reference implementation validates the third argument at construction time; the port stores it as-is, so `mutable-array?` reports `#t`, `array-setter` returns the raw non-procedure, and the failure is deferred until (and unless) the setter is ever invoked.

Found by running the SRFI 231 reference test suite's 330 error-expectation tests against Kaappi (this is T70/T255; 322 of 330 pass).

### Reproduction

```scheme
kaappi> (import (srfi 231))
kaappi> (mutable-array? (make-array (make-interval '#(3) '#(4)) list 1))
#t
kaappi> (array-setter (make-array (make-interval '#(3) '#(4)) list 1))
1
```

Reference errors at construction: `"make-array: The third argument is not a procedure: "` (test-arrays.scm error tests T70/T255; generic-arrays.scm `make-array` checks).

### Analysis

`lib/srfi/231/arrays.sld:93-97` validates `interval` and `getter` but stores `maybe-setter` unchecked. Consequences: `mutable-array?` and `array-setter` misrepresent the object (spec: "mutable-array? assumes that array is an array; returns #t if array has a setter" — a non-procedure setter is not a setter in any usable sense); `array-assign!` accepts such an array as a destination and then fails at first write with an unrelated "not a procedure" error; an array that is only ever read never fails at all.

Priority: medium — construction with a non-procedure setter is an error situation per the reference, so any behavior is spec-permitted; but the port's own `getter` check two lines up shows the intent is construction-time validation, and the delayed/wrong-signature failure mode is the kind users can ship on top of without noticing.

### Fix sketch

In `make-array`, after extracting `setter`: `(unless (or (not setter) (procedure? setter)) (error "make-array: setter must be a procedure or #f" setter))`. Regression test beside the existing make-array tests in `tests/scheme/srfi/srfi231-*.scm`, guarded, asserting the error.

Source: found by sweeping the [reference test suite](https://github.com/scheme-requests-for-implementation/srfi-231) error tests against Kaappi.

### LINKED ISSUE #2316: SRFI 231: non-boolean safe?/mutable? accepted by make-specialized-array, make-specialized-array-from-data, and both default parameters
### Summary

The `safe?` and `mutable?` options of `make-specialized-array` and `make-specialized-array-from-data`, and the two parameters `specialized-array-default-safe?` / `specialized-array-default-mutable?`, all accept non-boolean values without error. The reference implementation raises for every one of these four entry points. One defect class, four sites.

Found by sweeping the SRFI 231 reference test suite's 330 error-expectation tests against Kaappi (T83, T109, T110; 322 pass).

### Reproduction

```scheme
kaappi> (import (srfi 231))
kaappi> (array-safe? (make-specialized-array (make-interval '#(0) '#(10)) generic-storage-class 'a 'a))
a                       ; wrong type stored, returned by array-safe?
kaappi> (make-specialized-array-from-data '#(1 2 3) generic-storage-class 'a 'a)
#<<array> ...>          ; constructed with no error
kaappi> (specialized-array-default-safe? 'a)
                        ; no error; 'a installed as the parameter value
```

Reference messages:
- `"make-specialized-array: The fourth argument is not a boolean: "` (T83)
- `"specialized-array-default-safe?: The argument is not a boolean: "` (T109)
- `"specialized-array-default-mutable?: The argument is not a boolean: "` (T110)
- plus the corresponding `make-specialized-array-from-data` checks.

### Analysis

- `lib/srfi/231/arrays.sld:135-148` (`make-specialized-array`) reads `safe?` via `%opt` and uses it truthily; a non-boolean lands verbatim in the record, so `array-safe?` returns it — an accessor documented to return a boolean returning `'a`. A truthy non-boolean also silently *enables* bounds-checked accessors (divergence from the requested mode).
- `lib/srfi/231/arrays.sld:154-169` (`make-specialized-array-from-data`) treats `mutable?` and `safe?` the same way.
- `lib/srfi/231/arrays.sld:99-100` defines both parameters as plain `make-parameter`s with no converter, so `(specialized-array-default-safe? 'a)` installs `'a` as the global default — and then every subsequently created specialized array inherits a non-boolean `safe?`.

Priority: medium — non-boolean arguments are an error situation per the reference, so behavior is unconstrained by the spec; but the parameter case mutates global default state and every later constructor compounds it, which is the silent-failure end of medium.

### Fix sketch

Add a shared `%check-boolean!` helper and apply it in both constructors; define the two parameters with converters (`(make-parameter #f %check-boolean!)` style — note R7RS converters receive both filter and set values, so reject non-booleans there). Regression tests for all four sites.

Source: found by sweeping the [reference test suite](https://github.com/scheme-requests-for-implementation/srfi-231) error tests against Kaappi.

### LINKED ISSUE #2317: SRFI 231: specialized-array-share accepts a non-procedure mapper (never detected for empty domains)
### Summary

`specialized-array-share` accepts a third argument (`new-domain->old-domain`) that is not a procedure. The reference raises immediately; the port constructs the array anyway, deferring the failure to the first access — and for an empty `new-domain` the mapper is never called, so a bogus array is constructed with no error at any point.

Found by sweeping the SRFI 231 reference test suite's 330 error-expectation tests against Kaappi (T150; 322 pass).

### Reproduction

```scheme
kaappi> (import (srfi 231))
;; non-procedure mapper, nonempty domain: array built, fails only on access
kaappi> (specialized-array-share (make-specialized-array (make-interval '#(1) '#(2)))
                                 (make-interval '#(0) '#(1)) 1)
#<<array> ...>
;; empty new-domain: the mapper is never invoked, nothing ever fails
kaappi> (array-domain (specialized-array-share (make-specialized-array (make-interval '#(1) '#(2)))
                                               (make-interval '#(0) '#(0)) 5))
#<<interval> #(0) #(0)>
```

Reference: `"specialized-array-share: The third argument is not a procedure: "` (generic-arrays.scm; test-arrays.scm T150).

### Analysis

`lib/srfi/231/views.sld:69-73` validates the first argument (specialized array), second (interval), and the volume inequality, but not that the mapper is a procedure. `specialized-array-share` is the foundational primitive every view procedure routes through (`array-extract`, `array-translate`, `array-permute`, `array-reverse`, `array-sample`, `array-curry` internals), and it is also a public export — the natural entry point for users building custom shared views.

Priority: medium — a non-procedure mapper is an error situation per the reference; the delayed failure is loud for nonempty domains, but the empty-domain variant produces a permanently bogus specialized array (a getter/setter pair that can never be invoked, yet is `specialized-array?`, has a body, and would pass through `array-copy`/`array-assign!` plumbing).

### Fix sketch

Add `(unless (procedure? new-domain->old-domain) (error "specialized-array-share: new-domain->old-domain must be a procedure" new-domain->old-domain))` beside the existing checks at `views.sld:69-73`. Regression test: guarded error for the non-procedure case plus the empty-domain case.

Source: found by sweeping the [reference test suite](https://github.com/scheme-requests-for-implementation/srfi-231) error tests against Kaappi.

### LINKED ISSUE #2318: SRFI 231: array-tile accepts a scalar slice-width on a zero-width axis (reference errors)
### Summary

`array-tile` accepts a scalar slice-width on an axis whose width is 0. The reference implementation requires the axis width to be positive when the slice-width entry is a scalar; a zero-width axis may only be tiled with the explicit-vector form (a nonempty vector of nonnegative exact integers summing to 0, e.g. `#(#(0))`).

Found by sweeping the SRFI 231 reference test suite's 330 error-expectation tests against Kaappi (T236; 322 pass).

### Reproduction

```scheme
kaappi> (import (srfi 231))
kaappi> (array-domain (array-tile (make-array (make-interval '#(0)) list) '#(2)))
#<<interval> #(0) #(0)>      ; silently an empty outer array; reference errors
kaappi> (array-domain (array-tile (make-array (make-interval '#(0)) list) '#(#(0))))
#<<interval> #(0) #(1)>      ; correct: vector form is the sanctioned way
```

Reference message: `"array-tile: Element 0 of the second argument is neither a positive exact integer (allowed if the width of the first argument's corresponding axis is positive) nor a nonempty vector of nonnegative exact integers summing to the width of axis 0 of the first argument: "` (generic-arrays.scm `slice-widths-check`; test-arrays.scm T236).

### Analysis

`lib/srfi/231/views.sld` (`%tile-cut-offsets`, ~line 246) accepts any positive exact integer `Sk` regardless of the axis width; with width 0 it computes zero slices and the result is an empty outer domain with no error. The reference's `slice-widths-check` adds the `(not (eqv? width 0))` conjunct to the scalar branch. The port's vector branch (nonempty, nonnegative entries, summing to the width) already matches the reference.

Priority: medium — the calling program is non-conforming (the reference treats it as an error), and the port's silent result (an empty outer array) is a plausible-looking answer rather than an error, which is the misleading end of medium. The window is narrow: empty axis + scalar slice-width.

### Fix sketch

In `%tile-cut-offsets`'s scalar branch, take the axis width as it already does and raise when `(eqv? width 0)`, mirroring the reference message shape. Regression test beside the existing array-tile tests in `tests/scheme/srfi/srfi231-*.scm`: scalar-on-empty-axis errors, `#(#(0))` still works.

Source: found by sweeping the [reference test suite](https://github.com/scheme-requests-for-implementation/srfi-231) error tests against Kaappi.

baijum:
Group A of the SRFI 231 audit fixes — six reference-parity argument-validation guards, one per issue, each with a guarded regression test in its per-module suite. Found by extracting every `(test expr "expected-error-message")` form from the [SRFI 231 reference test suite](https://github.com/scheme-requests-for-implementation/srfi-231) (330 of them — the message means the reference *raises*) and running each against kaappi.

**Method note:** these are all error situations per the SRFI ("It is an error to call … if the arguments do not satisfy this condition"), so any behavior is spec-permitted — but the reference implementation raises explicitly at each site, and half of these expressions are literal error tests in its suite. The port's own neighbors already treated the sibling cases as errors (e.g. `interval-subset?` on dimension mismatch, non-exact-integer multi-index components); these six were the gaps.

| Issue | Site | Fix |
|---|---|---|
| #2312 | `interval-contains-multi-index?` | dimension mismatch now errors instead of silently returning `#f` |
| #2313 | built-in storage classes | `data->body` validates the data type instead of being the identity (`values`) |
| #2315 | `make-array` | setter must be a procedure or `#f`, checked at construction |
| #2316 | both specialized constructors + both default parameters | non-boolean `safe?`/`mutable?` rejected; parameters get boolean converters |
| #2317 | `specialized-array-share` | mapper must be a procedure — checked even for empty new-domains where it is never invoked |
| #2318 | `array-tile` | scalar slice-width only on a positive-width axis (reference's own rule); explicit-vector form unchanged |

**Verification**

- Reference error-test sweep: 322/330 before → **330/330 after** (no flags, no hangs).
- All seven `tests/scheme/srfi/srfi231-*.scm` suites pass (arrays 71, assembly 63, combinators 56, intervals 82, storage-classes 142, views 79, hub 17).
- The three Reddit-reported repros: the two covered here now raise with reference-shaped messages; the third (`array-packed?` on an extracted view) is unchanged as intended.

**Out of scope:** #2314 (`array-packed?` zero-offset semantics, priority: high) is a behavior change with `specialized-array-reshape` fallout and a dev-notes update — it ships as its own PR (group B of the audit plan), rebranching trivially off this one.

Closes #2312
Closes #2313
Closes #2315
Closes #2316
Closes #2317
Closes #2318

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- **SRFI 231 argument validation now matches the reference implementation at
+  six entry points** (#2312, #2313, #2315, #2316, #2317, #2318), found by
+  running the reference test suite's 330 error-expectation tests against
+  Kaappi (all 330 now pass; 322 did before).
+  `interval-contains-multi-index?` signaled nothing when the number of
+  indices differed from the interval's dimension, silently returning `#f`
+  (#2312); every built-in storage class's `data->body` was the unvalidated
+  identity, handing wrong-typed data back as a would-be body (#2313);
+  `make-array` accepted a non-procedure setter, so `mutable-array?` and
+  `array-setter` then misreported the object (#2315); non-boolean
+  `safe?`/`mutable?` values were accepted by both specialized constructors
+  and both default parameters — the parameter case installing garbage as a
+  process-wide default every later array would inherit (#2316);
+  `specialized-array-share` accepted a non-procedure mapper, undetectable
+  for empty domains where the mapper is never invoked (#2317); and
+  `array-tile` accepted a scalar slice-width on a zero-width axis, where the
+  spec permits only the explicit-vector form (#2318).
+
 ## [0.23.0] - 2026-08-23
 
 ### Added
--- a/lib/srfi/231/arrays.sld
+++ b/lib/srfi/231/arrays.sld
@@ -94,10 +94,24 @@
       (unless (interval? interval) (error "make-array: not an interval" interval))
       (unless (procedure? getter) (error "make-array: getter must be a procedure" getter))
       (let ((setter (if (null? maybe-setter) #f (car maybe-setter))))
+        (unless (or (not setter) (procedure? setter))
+          (error "make-array: setter must be a procedure or #f" setter))
         (%make-array interval getter setter #f #f #f #f)))
 
-    (define specialized-array-default-safe? (make-parameter #f))
-    (define specialized-array-default-mutable? (make-parameter #t))
+    ;; safe?/mutable? are booleans everywhere they appear (the two default
+    ;; parameters and both specialized constructors' options) -- the
+    ;; reference implementation rejects non-booleans at each site rather
+    ;; than letting a truthy wrong-typed value silently flip a mode.
+    (define (%check-boolean! x who)
+      (unless (boolean? x) (error (string-append who ": argument must be a boolean") x)))
+
+    (define (%boolean-parameter default name)
+      (make-parameter default (lambda (x) (%check-boolean! x name) x)))
+
+    (define specialized-array-default-safe?
+      (%boolean-parameter #f "specialized-array-default-safe?"))
+    (define specialized-array-default-mutable?
+      (%boolean-parameter #t "specialized-array-default-mutable?"))
 
     ;; The lexicographical mapping of interval onto [0, interval-volume) --
     ;; per spec, a specialized array's own indexer. Subtracts each axis's
@@ -134,6 +148,8 @@
 
     (define (make-specialized-array interval . opts)
       (unless (interval? interval) (error "make-specialized-array: not an interval" interval))
+      (%check-boolean! (%opt opts 2 (specialized-array-default-safe?))
+                       "make-specialized-array: safe?")
       (let* ((storage-class (%opt opts 0 generic-storage-class))
              (initial-value (%opt opts 1 (storage-class-default storage-class)))
              (safe? (%opt opts 2 (specialized-array-default-safe?)))
@@ -152,6 +168,10 @@
     ;; new 1-D array WITHOUT copying -- data must already be exactly
     ;; body-shaped for storage-class (per storage-class-data?).
     (define (make-specialized-array-from-data data . opts)
+      (%check-boolean! (%opt opts 1 (specialized-array-default-mutable?))
+                       "make-specialized-array-from-data: mutable?")
+      (%check-boolean! (%opt opts 2 (specialized-array-default-safe?))
+                       "make-specialized-array-from-data: safe?")
       (let* ((storage-class (%opt opts 0 generic-storage-class))
              (mutable? (%opt opts 1 (specialized-array-default-mutable?)))
              (safe? (%opt opts 2 (specialized-array-default-safe?))))
--- a/lib/srfi/231/intervals.sld
+++ b/lib/srfi/231/intervals.sld
@@ -125,16 +125,22 @@
                  (<= (interval-upper-bound i1 k) (interval-upper-bound i2 k))
                  (loop (+ k 1))))))
 
+    ;; Per spec, the multi-index must have exactly the interval's
+    ;; dimension -- a mismatch is an error, not a normal #f result (#f is
+    ;; reserved for a well-formed multi-index that fails the bound check),
+    ;; mirroring interval-subset? above.
     (define (interval-contains-multi-index? interval . multi-index)
+      (unless (= (length multi-index) (interval-dimension interval))
+        (error "interval-contains-multi-index?: the dimension of the interval does not match the number of indices"
+               interval multi-index))
       (let ((lo (interval-lower-vec interval)) (up (interval-upper-vec interval)))
-        (and (= (length multi-index) (vector-length lo))
-             (let loop ((i 0) (xs multi-index))
-               (or (null? xs)
-                   (begin
-                     (unless (and (integer? (car xs)) (exact? (car xs)))
-                       (error "interval-contains-multi-index?: multi-index entries must be exact integers" (car xs)))
-                     (and (<= (vector-ref lo i) (car xs)) (< (car xs) (vector-ref up i))
-                          (loop (+ i 1) (cdr xs)))))))))
+        (let loop ((i 0) (xs multi-index))
+          (or (null? xs)
+              (begin
+                (unless (and (integer? (car xs)) (exact? (car xs)))
+                  (error "interval-contains-multi-index?: multi-index entries must be exact integers" (car xs)))
+                (and (<= (vector-ref lo i) (car xs)) (< (car xs) (vector-ref up i))
+                     (loop (+ i 1) (cdr xs))))))))
 
     ;; General d-dimensional nested traversal in lexicographic order. proc
     ;; receives each multi-index as a LIST; public callers apply it to their
--- a/lib/srfi/231/storage-classes.sld
+++ b/lib/srfi/231/storage-classes.sld
@@ -57,6 +57,20 @@
       (data? storage-class-data?)
       (data->body storage-class-data->body))
 
+    ;; data->body: the body IS the data for every built-in class (no
+    ;; offset/scale to install), but the contract still requires rejecting
+    ;; wrong-typed data instead of handing it back as a would-be body
+    ;; (reference implementation: "converts data to a body, raising an
+    ;; exception if needed").
+    (define (%checked-data->body pred class-name type-name)
+      (lambda (data)
+        (if (pred data)
+            data
+            (error (string-append "Expecting a " type-name
+                                  " passed to (storage-class-data->body "
+                                  class-name "): ")
+                   data))))
+
     ;; Reference definitions from the spec text itself, reused verbatim --
     ;; Kaappi's native vector/string already satisfy every one of
     ;; make-storage-class's operational contracts (O(1) getter/setter,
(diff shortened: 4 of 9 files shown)


baijum:
LGTM ✅

Reviewed the full diff, the six referenced issues (#2312–#2318), the surrounding `lib/srfi/231/*.sld` context, and the two existing PR comments. Verified locally: all seven SRFI 231 suites pass with counts matching the PR (arrays 71, assembly 63, combinators 56, intervals 82, storage-classes 142, views 79, hub 17); `format` and DCO are green; the benchmark comment shows no regression.

**What's correct**

- Each of the six guards is a faithful reference-parity fix that reuses the port's existing conventions: `interval-subset?`'s error-on-dimension-mismatch for #2312, each storage class's own `data?` predicate for #2313, and `make-parameter` converters for #2316.
- The regression tests genuinely fail without the fixes. The parameter tests asserting the defaults are *unchanged after a rejected set* are a nice touch — they prove the converter errors before storing, so a failed set can't corrupt process-wide state.
- `specialized-array-share`'s mapper check sits before the volume guard, so the empty-domain case is covered — which is the whole point of #2317.

**Non-blocking notes** (no change required):

- Wrong-dimension `array-ref`/`array-set!` on a safe array now raises the interval's dimension-mismatch error instead of `array getter/setter: multi-index not in domain`. Message-only; both are errors either way.
- `%opt` re-evaluates the parameter default twice in each constructor (harmless).
- Tests use catch-all `guard (e (#t #t))`, matching the file's existing style rather than asserting the exact message.
- `bytevector` vs the reference's `u8vector` in the #2313 message is deliberate and accurate for this port (u8 *is* a bytevector here).

The #2314 group-B rebase note (same files touched) is worth keeping in mind, but nothing needs to change in this PR.

Issue #2321, opened 2026-08-25
SRFI 231: combinator family accepts non-procedure function arguments (array-map, array-outer-product, folds, any/every, for-each, inner-product)


baijum:
### Summary

The combinator family does not validate that its function arguments are procedures: `array-map`, `array-for-each`, `array-fold-left`, `array-fold-right`, `array-any`, `array-every`, `array-outer-product`, and `array-inner-product` (its `f` and `g`) all accept a non-procedure first/second/third argument. Only `array-reduce` checks (it already errors "array-reduce: not a procedure"). The reference implementation validates every one of them at call time with a specific message.

Reported by Brad Lucier (SRFI 231's author) in [a follow-up comment](https://www.reddit.com/r/KaappiScheme/comments/1vxk61h/comment/p5pn0ry/) — his example was `(array-outer-product 'a test-array test-array)`.

### Reproduction

```scheme
kaappi> (import (srfi 231))
kaappi> (define t (make-specialized-array (make-interval '#(1) '#(2))))
kaappi> (array? (array-outer-product 'a t t))
#t                                    ; reference: "The first argument is not a procedure"
kaappi> (array? (array-map 'a t))
#t                                    ; ditto
;; worst case: an eager combinator over an EMPTY domain never applies f at all
kaappi> (array-for-each 'a (make-array (make-interval '#(0)) list))
                                      ; returns silently — no error, ever
```

For the lazy constructors (`array-map`, `array-outer-product`), the failure is deferred to the first element access and may never happen; for the eager ones the failure surfaces at the first `apply` as a generic "not a procedure" — except over empty domains, where it is never detected at all.

Reference messages (generic-arrays.scm, one per constructor): `"array-map: The first argument is not a procedure: "`, `"array-for-each: …"`, `"array-fold-left: …"`, `"array-fold-right: …"`, `"array-any: …"`, `"array-every: …"`, `"array-outer-product: The first argument is not a procedure: "`, `"array-inner-product: The second/third argument is not a procedure: "`.

### Analysis

`lib/srfi/231/combinators.sld`: none of these constructors guard their function argument; the eager ones rely on the eventual `apply` failing, the lazy ones build a `make-array` whose getter would fail per-access. Same defect class as #2315/#2317 (non-procedure accepted where the reference rejects), but with the empty-domain/never-accessed silent-success aggravation for the lazy and empty cases.

[...]

Issue #2320, opened 2026-08-25
SRFI 231: array-copy/array-copy! accept non-boolean mutable?/safe? options (the #2316 gap)


baijum:
### Summary

`array-copy` / `array-copy!` do not validate their own `mutable?` and `safe?` options. This is the one gap #2316's boolean validation left: `array-copy` parses its options separately (`%array-copy-impl` in views.sld), and while `safe?` happens to flow into `make-specialized-array` (which now validates it, though the error is attributed to the constructor rather than to `array-copy`), `mutable?` never reaches any validating code — it only gates the final `array-freeze!`.

Reported by Brad Lucier (SRFI 231's author) in [a follow-up comment](https://www.reddit.com/r/KaappiScheme/comments/1vxk61h/comment/p5pn0ry/) on the r/KaappiScheme anomalies post.

### Reproduction

```scheme
kaappi> (import (srfi 231))
kaappi> (define src (make-array (make-interval '#(1) '#(2)) list))
kaappi> (mutable-array? (array-copy src generic-storage-class 'a))
#t                        ; reference: error "'a is not a boolean"
kaappi> (array-copy src generic-storage-class #f 'a)
ERROR: make-specialized-array: safe?: argument must be a boolean   ; errors, but attributed to the constructor
```

Reference: `array-copy`/`array-copy!` validate all four arguments through an explicit chain (`four-args` → `three-args` → `two-args` → `one-arg` in generic-arrays.scm): array is an array, storage-class is a storage class, `mutable?` and `safe?` are booleans, with `array-copy`-attributed messages.

### Analysis

`lib/srfi/231/views.sld` (`%array-copy-impl`, ~line 308): `mutable?` is read via `%opt` and used only truthily (`(if mutable? dest (array-freeze! dest))`), so `'a` silently yields an unfrozen — mutable — array. The `safe?` path errors transitively via `make-specialized-array`'s new check from #2316, so the observable gap is `mutable?` plus error attribution for `safe?`.

Priority: medium — same class as #2316 (error situation per the reference, silent-or-misattributed here), and this is the exact constructor most user code funnels through.

### Fix sketch

[...]

PR #2322, opened 2026-08-25
Fix array-packed?: consecutive increasing indices from any body base (#2314)

### LINKED ISSUE #2314: SRFI 231: array-packed? wrongly requires zero body offset — wrong #f on extracted views, reshape fallout

baijum:
Group B of the SRFI 231 audit: the semantic fix for `array-packed?` (#2314, priority: high — the one audit finding where a *legal* program got a silently wrong answer), plus its behavioral ripple into `specialized-array-reshape`.

**The fix.** The spec: "`#t` if the elements of array, taken in lexicographical order, are stored in `(array-body array)` with **increasing and consecutive indices**" — any base offset qualifies. The port demanded the traversal start at body index 0. The new implementation captures the first visited index as the base and requires base+1, base+2, … thereafter (same single-pass `call/cc` early exit). Empty arrays remain vacuously `#t`, matching the reference.

```scheme
(define a (make-specialized-array (make-interval '#(2) '#(4))))
(array-packed? (array-extract a (make-interval '#(3) '#(4))))   ; was #f, now #t
```

**The ripple.** `specialized-array-reshape`'s fast path already computed its `base` from the first indexer value (`views.sld`) — it was written for the correct semantics. With `array-packed?` fixed, an offset view now reshapes **in place, sharing the body** (write-through verified in tests), exactly like the reference, instead of erroring "not affinely representable" or forcing a copy.

**Also in line with the reference, verified:** `array-reverse` → `#f` (decreasing), `array-sample` → `#f` (stride ≠ 1), empty and zero-dimensional → `#t` — the spec's own `array-packed?` example block passes unchanged.

**Verification:** all seven `tests/scheme/srfi/srfi231-*.scm` suites pass (the packed-view cases live in the views suite, which imports the view constructors); the spec document's full worked-example corpus still passes (66 automated checks); the reference test suite's 330 error-expectation tests remain 330/330. Docs: the reshape-simplification note in `srfi-implementation-notes.md` now records the corrected semantics.

Follow-ups from the same audit still open: #2320 (`array-copy` option booleans) and #2321 (combinator function-argument validation) — a small same-class PR like #2319 whenever wanted.

Closes #2314

### DIFF
--- a/tests/scheme/srfi/srfi231-views.scm
+++ b/tests/scheme/srfi/srfi231-views.scm
@@ -182,6 +182,28 @@
   (let ((forced (specialized-array-reshape b (make-interval (vector 8)) #t)))
     (test-equal #t (interval= (array-domain forced) (make-interval (vector 8))))))
 
+;;; --- an offset view (non-zero body base) IS packed (#2314), so reshape
+;;; flattens it in place -- sharing the body, not copying ---
+(let* ((base (make-specialized-array (make-interval (vector 2) (vector 4))))
+       (view (array-extract base (make-interval (vector 3) (vector 4))))
+       (r (specialized-array-reshape view (make-interval (vector 1)))))
+  (array-set! base 'x 3)
+  (test-equal 'x (array-ref r 0))
+  (array-set! r 'y 0)  ;; write-through proves body sharing
+  (test-equal 'y (array-ref base 3)))
+
+;;; --- array-packed? on views: consecutive-from-any-base per spec (#2314);
+;;; the first visited body position may have ANY offset, only stride-1
+;;; increasing order matters ---
+(let* ((a (make-specialized-array (make-interval (vector 2) (vector 4))))
+       (A (make-specialized-array-from-data (vector 0 1 2 3))))
+  (test-equal #t (array-packed? (array-extract a (make-interval (vector 3) (vector 4)))))
+  (test-equal #t (array-packed? (array-extract a (make-interval (vector 2) (vector 4)))))
+  (test-equal #t (array-packed? (array-translate a (vector 1))))
+  ;; the spec's own example: decreasing order and stride-2 are not packed
+  (test-equal #f (array-packed? (array-reverse A)))
+  (test-equal #f (array-packed? (array-sample A (vector 2)))))
+
 (test-equal #t (guard (e (#t #t))
                   (specialized-array-reshape (make-array (make-interval (vector 3)) (lambda (i) i)) (make-interval (vector 3)))
                   #f))  ;; not specialized
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- **`array-packed?` now follows the spec's definition: consecutive increasing
+  body indices from any base, not from zero** (#2314). It previously required
+  the lexicographic traversal to start at body index 0, so any view with a
+  non-zero offset — `(array-extract A interval)` being the common one —
+  wrongly reported `#f` (the spec says "increasing and consecutive indices";
+  the reference checks only stride-1 between neighbors). Side effect:
+  `specialized-array-reshape` now reshapes such offset views in place,
+  sharing the body like the reference, instead of erroring or copying.
+
 - **SRFI 231 argument validation now matches the reference implementation at
   six entry points** (#2312, #2313, #2315, #2316, #2317, #2318), found by
   running the reference test suite's 330 error-expectation tests against
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -828,7 +828,10 @@ pre-materialized before the fill, a distinction observable only under
 multi-shot-continuation re-entry, which the spec itself declares undefined).
 `specialized-array-reshape` uses a deliberate packed-check-based
 affine-detection simplification instead of the reference's full multi-group
-algorithm, verified identical on the spec's own worked examples. `array-block`
+algorithm, verified identical on the spec's own worked examples. (`array-packed?`
+itself means consecutive-increasing from any body base — not a zero base —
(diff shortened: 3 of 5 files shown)


tests/scheme/srfi/srfi231-views.scm
```diff
@@ -182,6 +182,28 @@
...
+  (test-equal 'x (array-ref r 0))
+  (array-set! r 'y 0)  ;; write-through proves body sharing
+  (test-equal 'y (array-ref base 3)))
+
+;;; --- array-packed? on views: consecutive-from-any-base per spec (#2314);
+;;; the first visited body position may have ANY offset, only stride-1
+;;; increasing order matters ---
+(let* ((a (make-specialized-array (make-interval (vector 2) (vector 4))))
+       (A (make-specialized-array-from-data (vector 0 1 2 3))))
+  (test-equal #t (array-packed? (array-extract a (make-interval (vector 3) (vector 4)))))
```
baijum, line 200: Non-blocking: the positive offset cases here only exercise a *single-element* extract (`[3,4)` -> body base 1) and a base-0 `array-translate`; `array-extract a [2,4)` covers two elements but from base 0. Suggest adding one *multi-element* non-zero-base slice — e.g. `(array-packed? (array-extract A (make-interval '#(3) '#(7))))` with `A` spanning `[0,10)` — which is the strongest form of #2314 (first element at a non-zero base plus several consecutive successors), and it would also pin the reshape fast path for a multi-element offset view (the write-through test above only checks one element).

baijum:
**LGTM** — correct fix, well-scoped, and it aligns `array-packed?` with the spec and the reference implementation.

**Verified:**

- **Implementation** (`lib/srfi/231/arrays.sld:205-216`): the first visited body index is captured as the base and each successor must equal `base+1, base+2, …`; `call/cc` still short-circuits on the first mismatch. The `#f` sentinel is safe because `expected` becomes an integer after the first element, so `(not expected)` cleanly distinguishes the first visit. Empty (zero `interval-for-each` calls) and zero-dimensional (one call) arrays still return `#t`.
- **Reshape ripple** needs no code change — `views.sld:358` already computed `base` from the first indexer value, so the fast path was written for the corrected semantics and now takes effect for offset views. The new write-through test proves body sharing in both directions.
- **Tests**: ran locally — `srfi231.scm` (17), `srfi231-arrays.scm` (74), `-assembly` (63), `-combinators` (56), `-intervals` (82), `-storage-classes` (142), `-views` (86): all pass.

**Non-blocking suggestion** (see inline comment): add a *multi-element* non-zero-base positive case to fully exercise the consecutive-from-any-base branch.

**Trivial:** the PR description says "all seven `srfi231-*.scm` suites" — six files match that glob; the seventh is `srfi231.scm`.

PR #2324, opened 2026-08-25
Reclaim descriptors on EMFILE before failing an open (#1993)

### LINKED ISSUE #1993: GC: open-input-file and open-directory raise on EMFILE without collecting and retrying
**Category:** GC / R7RS
**Severity:** wrong-result (spurious failure on a legal program)

`open-input-file`, `open-output-file` and `open-directory` raise as soon as the
OS says `EMFILE`. They never force a collection and retry — yet the objects
holding those descriptors are ordinarily unreachable and *are* reclaimable, so
the retry would succeed. A program that abandons fd-holding objects faster than
the GC's allocation-count threshold trips will fail on a machine with a normal
`ulimit -n`, and succeed on one with a large one. Nothing about the program
changes; only the descriptor limit does.

**Minimal reproduction** — count how many unclosed streams fit before failure:

```scheme
(import (scheme base) (scheme write) (scheme file) (scheme process-context) (srfi 170))
(define d (string-append "/tmp/fdcount-" (number->string (pid))))
(create-directory d)
(call-with-output-file (string-append d "/seed") (lambda (p) (write-string "x" p)))
(display (call/cc (lambda (k)
  (let loop ((i 0))
    (if (= i 20000) i
        (begin (guard (e (#t (k i))) (open-directory d)) (loop (+ i 1))))))))
```

| `ulimit -n` | `open-directory` | `open-input-file` |
|---|---:|---:|
| 256 | **253** | **253** |
| 1024 | 1745 | 1747 |

At 256 the count is the descriptor limit itself — **no collection ran at all**
before the table was exhausted. At 1024 the count exceeds the limit, which
proves the sweep arm does work and the objects *are* collectable; the GC simply
has no idea the process is under descriptor pressure. Ports and directory
streams behave identically, so this is a property of every fd-holding heap
object, not of one primitive.

**Discriminating control:** the byte-identical loop with `close-directory` added
completes 3000 iterations and leaves the next `call-with-output-file` working,
at every limit tested. So the fault is the missing reclamation path, not the
loop, the directory, or the filesystem.

**Why it is not merely "the program should close its streams":** the finalizer
already exists and is correct — `gc_sweep.zig:537` closes the `DIR*`, and ports
close their fd the same way. The design intent is clearly that an abandoned
stream is reclaimed. It is reclaimed *eventually*, on a schedule driven by
object-allocation count, which is uncorrelated with descriptor consumption: a
`DirectoryObject` is 24 bytes, so thousands fit under the 8192-object threshold.

**Suggested fix:** on `EMFILE`/`ENFILE` specifically, run a full collection and
retry the open once before raising. This is the standard technique for
GC-managed OS handles. The failing paths are:

- `src/primitives_io.zig:720` — `platform.openRead(path_z) catch { return raiseFileError(...) }`
- `src/primitives_io.zig:743` — same shape for the output path
- `src/primitives_filesystem.zig:992` — `platform.dirIterCreate(path_z) orelse { return raiseFileError(...) }`
- `src/primitives_filesystem.zig:267` — the second `open-directory` entry point

**Verified:** fresh ReleaseSafe build at 5c3e3776, macOS aarch64, isolated
`KAAPPI_HOME`. Numbers above are from direct runs, not inferred.

**How it surfaced:** `tests/scheme/audit/primitives_filesystem-audit.scm` opened
3000 directory streams without closing them, asserting they were reclaimed. That
assertion passes on macOS and Linux CI (`ulimit -n` 1048576) and **aborts the
whole suite** on the `openbsd-test` and `netbsd-test` legs, where the next block
cannot create its own scratch file:

```
error[KP3000]: cannot open output file "/tmp/kaappi-a212-68521-14/f0"
error[KP3000]: cannot create temp file #f
```

The assertion is disabled in that file pending this issue — it asserts a
property the runtime does not currently provide.

**Found by:** Systematic audit v2, Phase 2.12

baijum:
## Problem

`open-input-file`, `open-output-file`, and `open-directory` raised as soon as the OS reported `EMFILE` (per-process) / `ENFILE` (system-wide), without forcing a collection and retrying. The fd-holding objects (ports, SRFI-170 directory streams) are ordinarily unreachable and their finalizers close the descriptor — so a collection would free table slots and the retry would succeed. A program that abandons fd-holders faster than the GC's allocation-count threshold trips fails at a normal `ulimit -n` and succeeds at a larger one; nothing about the program changes, only the descriptor limit.

At `ulimit -n 256` the reproduction fit only 253 opens (the descriptor limit itself), proving no collection ran before failing.

## Fix

- `src/platform.zig`: new `OpenError.FdExhausted` distinguishes EMFILE/ENFILE from every other open failure (`ProcessFdQuotaExceeded`/`SystemFdQuotaExceeded` on POSIX, `_errno` on the Windows `_wopen` path; `errnoIsFdExhausted()` for the `opendir` path, read immediately after the failing call). `dirIterCreate` now returns `OpenError!*DirIter`.
- `src/memory.zig`: `GC.collectFull()` forces a full (both-generation) collection — a plain `collect()` runs a minor sweep, which misses an fd-holder already promoted to the old generation.
- `src/primitives_io.zig` / `src/primitives_filesystem.zig`: `open{Input,Output}File`, `open-directory`, and `directory-files` now force a `collectFull()` and retry **once** on `FdExhausted`. Any other failure propagates unchanged.

## Tests

`tests/scheme/smoke/fd-reclaim-emfile-1993.sh` opens-and-abandons fd-holders past a low descriptor cap and asserts the loop completes (the GC reclaims fds) rather than failing. `zig build test` passes; `zig build -Dtarget=x86_64-windows` cross-compiles (platform.zig touched); gc-stress running.

Closes #1993

### DIFF
--- a/src/gc_collect.zig
+++ b/src/gc_collect.zig
@@ -328,7 +328,7 @@ fn isYoungPointer(gc: *GC, val: Value) bool {
     return obj.flags.generation == 0;
 }
 
-fn fullCollect(gc: *GC) void {
+pub fn fullCollect(gc: *GC) void {
     // #2196: drain the remembered_set up front. A full collect marks from
     // roots over both generations, so it never consults the set — and doing
     // this before sweepOld frees any old object means every entry is still
--- a/src/memory.zig
+++ b/src/memory.zig
@@ -534,6 +534,17 @@ pub const GC = struct {
         gc_collect.collect(self);
     }
 
+    /// Force a full (both-generation) collection regardless of the
+    /// minor-cycle schedule. Used to reclaim OS descriptors held by
+    /// unreachable objects — abandoned file ports and directory streams —
+    /// before reporting descriptor exhaustion to a program (kaappi#1993).
+    /// A plain `collect()` usually runs a minor sweep, which misses any
+    /// fd-holder already promoted to the old generation.
+    pub fn collectFull(self: *GC) void {
+        self.stats.collections += 1;
+        gc_collect.fullCollect(self);
+    }
+
     pub fn markValue(self: *GC, v: Value) void {
         gc_collect.markValue(self, v);
     }
--- a/src/platform.zig
+++ b/src/platform.zig
@@ -206,13 +206,41 @@ pub const sockPollReady = win_sock.sockPollReady;
 // is platform-specific and the runtime only ever uses these five shapes.
 // ---------------------------------------------------------------------------
 
-pub const OpenError = error{OpenFailed};
+/// `FdExhausted` singles out EMFILE (per-process) / ENFILE (system-wide)
+/// descriptor exhaustion so a caller can force a GC — reclaiming descriptors
+/// held by unreachable ports / directory streams — and retry the open before
+/// raising (kaappi#1993). Every other open failure stays `OpenFailed`.
+pub const OpenError = error{ OpenFailed, FdExhausted };
+
+/// Fold std.posix's open error set onto OpenError, mapping the two
+/// descriptor-exhaustion errnos (EMFILE → ProcessFdQuotaExceeded,
+/// ENFILE → SystemFdQuotaExceeded) to `FdExhausted`.
+fn classifyOpenError(e: anytype) OpenError {
+    return switch (e) {
+        error.ProcessFdQuotaExceeded, error.SystemFdQuotaExceeded => error.FdExhausted,
+        else => error.OpenFailed,
+    };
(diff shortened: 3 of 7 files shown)


baijum:
gc-stress verification: `zig build test -Dgc-stress=true` passed (exit 0) — the collect-and-retry path is clean under collection-on-every-allocation.

PR #2326, opened 2026-08-25
Validate array-copy options and combinator function arguments (#2320 #2321)

### LINKED ISSUE #2320: SRFI 231: array-copy/array-copy! accept non-boolean mutable?/safe? options (the #2316 gap)

### LINKED ISSUE #2321: SRFI 231: combinator family accepts non-procedure function arguments (array-map, array-outer-product, folds, any/every, for-each, inner-product)

baijum:
The final PR of the SRFI 231 audit: the two remaining findings from the SRFI author's follow-up report, both the same guard-clause shape as #2319.

**#2320 — `array-copy`/`array-copy!` validate their own boolean options.** `array-copy` parses its options separately from the constructors; its `mutable?` only gates the final `array-freeze!`, so a truthy wrong-typed value silently produced an unfrozen (mutable) array — and `safe?` errors were attributed to the inner `make-specialized-array` rather than to `array-copy`. Both options are now checked up front with the `array-copy:` prefix (the reference validates all four arguments through an explicit chain). `%check-boolean!` moved to `arrays.sld`'s internal `%`-helper exports so `views.sld` can reuse it — the hub's public 118-binding surface is unchanged.

**#2321 — every combinator rejects a non-procedure function argument at call time.** `array-map`, `array-for-each`, `array-fold-left`, `array-fold-right`, `array-any`, `array-every`, `array-outer-product`, and `array-inner-product` (both `f` and `g`) now guard their function argument like `array-reduce` already did. Previously the lazy combinators (`array-map`, `array-outer-product`) deferred the failure to the first element access — possibly never — and the eager ones could **succeed silently over an empty domain**, where `f` is never invoked at all (covered by a dedicated test).

**Verification:** all seven `tests/scheme/srfi/srfi231-*.scm` suites pass; the spec document's full worked-example corpus still passes (66 automated checks); the reference test suite's 330 error-expectation tests remain 330/330.

With this and the merged #2319/#2322, every finding from the audit — the Reddit post, the author's follow-up comment, the reference-suite sweep, and the spec-document example run — is resolved except the deliberately deferred `u1`/`f8`/`f16` storage classes (documented in `docs/dev/srfi-implementation-notes.md`).

Closes #2320
Closes #2321

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- **`array-copy`/`array-copy!` reject non-boolean `mutable?`/`safe?` options, and
+  every SRFI 231 combinator rejects a non-procedure function argument**
+  (#2320, #2321), closing the two gaps surfaced by the SRFI author's follow-up
+  report. `array-copy`'s `mutable?` never flowed through a validating
+  constructor, so a truthy wrong-typed value silently yielded an unfrozen
+  array; `array-map`, `array-for-each`, both folds, `array-any`, `array-every`,
+  `array-outer-product`, and `array-inner-product` now check at call time —
+  previously the lazy combinators deferred the failure to first element access
+  (possibly never) and the eager ones could succeed silently over an empty
+  domain.
+
 - **`array-packed?` now follows the spec's definition: consecutive increasing
   body indices from any base, not from zero** (#2314). It previously required
   the lexicographic traversal to start at body index 0, so any view with a
--- a/lib/srfi/231/arrays.sld
+++ b/lib/srfi/231/arrays.sld
@@ -40,7 +40,7 @@
           ;; and later phases) to build genuinely specialized arrays with
           ;; custom indexers, the same way (srfi 160 base)'s %uvec-* helpers
           ;; exist only for that package's own per-tag files to consume.
-          %make-array %safe-getter %safe-setter %make-lex-indexer)
+          %make-array %safe-getter %safe-setter %make-lex-indexer %check-boolean!)
   (begin
 
     (define-record-type <array>
--- a/lib/srfi/231/combinators.sld
+++ b/lib/srfi/231/combinators.sld
@@ -64,6 +64,15 @@
 
     (define (%opt lst i default) (if (> (length lst) i) (list-ref lst i) default))
 
+    ;; The reference implementation validates the function argument of
+    ;; every combinator at call time -- without this, the lazy ones
+    ;; (array-map, array-outer-product) would defer the failure to the
+    ;; first element access (possibly never), and the eager ones would
+    ;; fail with a generic apply error -- or succeed silently over an
+    ;; empty domain, where f is never invoked at all.
+    (define (%check-procedure! x who)
+      (unless (procedure? x) (error (string-append who ": the function argument must be a procedure") x)))
+
     (define (%check-same-domain! all who)
       (let ((domain (array-domain (car all))))
         (for-each (lambda (a)
@@ -78,6 +87,7 @@
     ;; definition, which builds an ordinary (immutable, non-specialized)
     ;; make-array rather than a specialized/precomputed one.
     (define (array-map f array . arrays)
+      (%check-procedure! f "array-map")
       (let ((all (cons array arrays)))
         (%check-same-domain! all "array-map")
         (make-array (array-domain array)
@@ -85,12 +95,14 @@
                       (apply f (map (lambda (a) (apply (array-getter a) multi-index)) all))))))
 
     (define (array-for-each f array . arrays)
+      (%check-procedure! f "array-for-each")
       (let ((all (cons array arrays)))
         (%check-same-domain! all "array-for-each")
         (interval-for-each (lambda multi-index (apply f (map (lambda (a) (apply (array-getter a) multi-index)) all)))
                             (array-domain array))))
 
     (define (array-fold-left operator identity array . arrays)
+      (%check-procedure! operator "array-fold-left")
       (let ((all (cons array arrays)) (acc identity))
         (%check-same-domain! all "array-fold-left")
         (interval-for-each (lambda multi-index
@@ -104,6 +116,7 @@
     ;; plain left-walk applying (operator e0 e1 ... acc), matching
     ;; interval-fold-right's own already-verified technique.
     (define (array-fold-right operator identity array . arrays)
+      (%check-procedure! operator "array-fold-right")
       (let ((all (cons array arrays)) (results '()))
         (%check-same-domain! all "array-fold-right")
         (interval-for-each (lambda multi-index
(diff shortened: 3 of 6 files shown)


baijum:
LGTM ✅

Reviewed the diff and ran the suites locally on `46225c75`. This cleanly finishes the SRFI 231 validation work, following the same guard-clause shape as the merged #2319.

**Correctness**
- `%check-procedure!` is applied to every function-taking combinator: `array-map`, `array-for-each`, `array-fold-left`, `array-fold-right`, `array-any`, `array-every`, `array-outer-product`, and both `f`/`g` of `array-inner-product`. `array-reduce` already had its own check — so the family is now fully covered, with none missed.
- The empty-domain guard is the important one: checking `f` up front (rather than at first element access) is what turns the previously silent success over an empty interval into an error. The dedicated `array-for-each ... empty` test pins exactly that regression.
- `array-copy`/`array-copy!` validate `mutable?`/`safe?` before the constructor path, using the default-parameter value as the `%opt` fallback — so an omitted option stays a boolean and passes, while a user-supplied wrong-typed value is rejected and attributed to `array-copy` itself.
- Moving `%check-boolean!` into `arrays.sld`s `%`-helper export list is the right seam for `views.sld` to reuse it; the public 118-binding surface is unchanged, and `%opt` is already defined locally in `views.sld`.

**Verification (local, `zig-out/bin/kaappi`)**
- `srfi231-combinators.scm` → 67 passes, `srfi231-views.scm` → 90 passes.
- Full family green: arrays 74, assembly 63, intervals 82, storage-classes 142, plus the `srfi231.scm` hub 17 — all exit 0.

**Minor, non-blocking**
- `array-copy!` errors are attributed to the `array-copy:` prefix (both funcs route through `%array-copy-impl`). This is consistent with the pre-existing `"array-copy: not an array"` message on line 317, so it is a continuation of the existing convention rather than something new — worth a follow-up only if per-name attribution is ever wanted.
- Message-style drift: the new `%check-procedure!` says `"... : the function argument must be a procedure"` while the older `array-reduce` inline check says `"array-reduce: not a procedure"`. Purely cosmetic.
- PR description says "seven `tests/scheme/srfi/srfi231-*.scm` suites" — the dash glob actually matches six; the seventh is `srfi231.scm` (no dash). Trivial wording.

Nothing here blocks merge.

PR #2328, opened 2026-08-25
Guard the WASM differential against a stale kaappi.wasm

### LINKED ISSUE #2197: [Tooling] The WASM differential silently accepts a stale kaappi.wasm — run-all.sh never rebuilds it
**Category:** Tooling / test harness
**Severity:** a differential check that can validate the wrong binary

`tests/scheme/differential/run-wasm-differential.sh` compares the interpreter
against `zig-out/bin/kaappi.wasm`, but nothing establishes that the module was
built from the tree under test. The script guards only *existence*:

```bash
if [ ! -f "$WASM" ]; then
    echo "SKIP: no wasm module at $WASM (run 'zig build wasm' first)"
    exit 77
fi
```

`tests/scheme/run-all.sh` never runs `zig build wasm` (it has no `wasm` step at
all), so a local `run-all.sh` compares today's interpreter against whatever
module happens to be sitting in `zig-out/` — possibly from a different commit,
a different branch, or weeks earlier. CI is unaffected: its `wasm` job runs
`zig build wasm` immediately before the harness.

**How it presented.** A stale module dated 4 days earlier (built before #2016)
produced three confident, specific failures naming real corpus files:

```text
FAIL: 3 WASM tier divergence(s):
  [wasm] tests/scheme/differential/probes/deep-print-depth.scm
  [wasm] tests/scheme/smoke/peek-char-fd-utf8-798.scm
  [wasm] tests/scheme/smoke/peek-char-utf8.scm
```

Nothing in that output suggests the module is old. The two `peek-char` files
diverged only because the stale module still emits the pre-#2016 wording:

```text
error[KP3002]: type error in 'open-input-file': expected non-WASM platform, got #<string>
```

That text no longer exists in the tree — `primitives_io.zig:780` now raises a
`file-error?` with `": this WebAssembly build has no filesystem access"`, which
is exactly what `PLATFORM_SIGNATURE` (line 255) matches. So the harness's
message-text matching worked correctly; it was matching against a binary that
predated the message. `deep-print-depth.scm` is a probe added by #2190, newer
than the module, and traps at nesting depth 848.

After `zig build wasm`: **0 divergences**, 360 files compared.

**Why this is worth fixing rather than filing under "user error".** The
harness's header already argues at length that a green run must not be trusted
blindly, and reports its own exclusion counts for that reason. The one thing it
does not report is which binary it measured. The failure direction seen here is
the loud one; the dangerous direction is silent — a stale module that still
agrees with the interpreter yields a clean **PASS** while testing nothing about
the current tree, which is precisely the "checks that look green while testing
nothing" class the v2 audit (#1890) put first.

**Suggested fix**, cheapest first:

1. Print the module's path, size and mtime in the harness preamble beside the
   `native:`/`wasm:`/`runtime:` lines it already prints. Costs nothing and
   makes staleness visible in any transcript.
2. Fail (or warn loudly) when the module is older than the newest tracked file
   under `src/` — the same freshness question `.sbc` cache keys answer with a
   build id (`docs/dev/cache.md`).
3. Or have `run-all.sh` run `zig build wasm` before the differential, the way
   CI does. Heavier, but removes the question entirely.

**Also, unrelated to staleness:** with a fresh module the harness reports

```text
  STALE: these KNOWN_DIFFS entries no longer diverge — delete them
         from run-wasm-differential.sh (not a failure):
  deep-nesting-print.scm
```

`deep-nesting-print.scm` is listed against #2107 (shadow-stack underflow on
deeply nested printing). It no longer diverges — plausibly fixed by #2190's
iterative printer. Worth confirming and then removing the entry, and closing
#2107 if it is genuinely fixed.

**Verified:** macOS arm64, wasmtime 46.0.0, at `ec29e955`. Stale module
Jul 31 15:34 -> 3 divergences; `zig build wasm` -> 0 divergences, same tree,
same corpus.

baijum:
## Problem

`tests/scheme/differential/run-wasm-differential.sh` compared today's
interpreter against `zig-out/bin/kaappi.wasm` but only guarded the module's
*existence* — never that it was built from the tree under test. Because
`run-all.sh` has no `zig build wasm` step, a local `run-all.sh` measured the
current interpreter against whatever module happened to be sitting in
`zig-out/` — possibly from another commit or weeks earlier. That produced
confident, specific **FALSE** tier-divergence FAILs naming real corpus files
(the harness's message-text matching worked correctly; it was matching against
an engine that predated the message, e.g. the pre-#2016 file-error wording).
The silent direction is worse: a stale module that still agrees yields a clean
**PASS** while testing nothing about the current tree. CI was unaffected — its
`wasm` job runs `zig build wasm` first.

## Fix

The key invariant: **a divergence FAIL can only be produced against a module
known to be built from this tree; otherwise SKIP with a clear reason.**

- **`run-wasm-differential.sh` is now self-protecting (correct however
  invoked).** A freshness gate runs after the existence check: if any
  interpreter source compiled *into* the module (`src/`, `build.zig{,.zon}`,
  `vendor/`) is newer than the module, it cannot reflect this tree, so the
  harness SKIPs (77) with a message telling you to run `zig build wasm` —
  instead of reporting divergences against a module of unknown provenance. Only
  the binary's build inputs are checked, so editing a test or a doc does not
  trip it (both tiers read those live from disk). The check is plain POSIX
  (`find -newer`; no GNU-only predicate, so the strict OpenBSD grep leg is
  unaffected — it sees no regex here). The preamble also now prints the module
  size as a cheap staleness tell in transcripts.

- **`run-all.sh` builds the module up front** (like CI) when both `zig` and
  `wasmtime` are present, so the common local path actually *runs* the leg
  instead of SKIPping. Best-effort and gated: without the toolchain the build
  is skipped and the differential's own freshness guard degrades the leg to a
  clear SKIP rather than a false FAIL.

The module cannot report its own `build_id` (the wasm entry point only runs a
`.scm` file — no `features`/`--version` subcommand), so an mtime-vs-sources
freshness check is the robust signal, matching the `make`-style freshness model
and issue suggestion (2).

## Verification (macOS arm64, wasmtime 48.0.0, zig 0.16.0)

**Fresh module** — `zig build wasm`, then run the differential: behaves as
before.

```
=== differential: WASM tier vs interpreter oracle ===
native:  .../zig-out/bin/kaappi (Kaappi Scheme v0.23.0)
wasm:    zig-out/bin/kaappi.wasm (1510794 bytes, verified newer than src/)
runtime: wasmtime 48.0.0 ...
...
  DIVERGENCES:                     0
  wasm hangs:                      0
PASS: no unknown WASM tier divergence across 368 files      # exit 0
```

**Missing module** — SKIPs 77 (unchanged message):

```
SKIP: no wasm module at zig-out/bin/does-not-exist.wasm (run 'zig build wasm' first)
# exit 77
```

**Stale module** — a source newer than the module (simulated with
`touch src/main.zig`) now SKIPs 77 instead of FAILing:

```
SKIP: wasm module zig-out/bin/kaappi.wasm is stale — interpreter sources are newer
      than it (e.g. src/main.zig), so it was not built from this tree.
      Run 'zig build wasm' to refresh it.  Refusing to report WASM
      tier divergences against a module of unknown provenance
      (kaappi#2197).
# exit 77
```

Both scripts pass `bash -n`; `shellcheck -x` reports no new findings.

Closes #2197

### DIFF
--- a/tests/scheme/differential/run-wasm-differential.sh
+++ b/tests/scheme/differential/run-wasm-differential.sh
@@ -168,7 +168,13 @@ for p in src build.zig build.zig.zon vendor; do
 done
 if [ -n "$FRESH_INPUTS" ]; then
     # shellcheck disable=SC2086  # deliberate word splitting over the path list
-    NEWER_SRC="$(find $FRESH_INPUTS -type f -newer "$WASM" 2> /dev/null)"
+    # Fail closed: if the scan itself fails (an input vanished mid-scan or is
+    # unreadable), freshness cannot be established — SKIP rather than proceed
+    # on an empty NEWER_SRC and measure a module of unknown provenance.
+    if ! NEWER_SRC="$(find $FRESH_INPUTS -type f -newer "$WASM" 2> /dev/null)"; then
+        echo "SKIP: cannot verify freshness of wasm module $WASM"
+        exit 77
+    fi
     if [ -n "$NEWER_SRC" ]; then
         NEWER_ONE="$(printf '%s\n' "$NEWER_SRC" | head -1)"
         echo "SKIP: wasm module $WASM is stale — interpreter sources are newer"
@@ -461,7 +467,7 @@ echo "=== differential: WASM tier vs interpreter oracle ==="
 echo "native:  $KAAPPI ($("$KAAPPI" --version 2> /dev/null | head -1))"
 # Size is a cheap staleness tell in a transcript; the freshness gate above has
 # already refused to run against a module older than the sources (kaappi#2197).
-echo "wasm:    $WASM ($(wc -c < "$WASM" 2> /dev/null | tr -d ' ') bytes, verified newer than src/)"
+echo "wasm:    $WASM ($(wc -c < "$WASM" 2> /dev/null | tr -d ' ') bytes, verified newer than interpreter build inputs)"
 echo "runtime: $(wasmtime --version 2> /dev/null | head -1)"
 printf 'corpus: '
 # shellcheck disable=SC2086  # deliberate word splitting over the dir list
--- a/tests/scheme/differential/run-wasm-differential.sh
+++ b/tests/scheme/differential/run-wasm-differential.sh
@@ -18,8 +18,10 @@
 #   KAAPPI_WASM_DIFF_VERBOSE=1 print a line per file instead of only diffs
 #   KAAPPI_WASM_DIFF_KEEP=1    keep the temp tree of captured outputs
 #
-# Exits 77 (SKIP) when there is no WASM runtime or no module to run — most
-# dev boxes have neither, and CI's `wasm` job has both.
+# Exits 77 (SKIP) when there is no WASM runtime, no module to run, or the module
+# predates the interpreter sources (i.e. was not built from this tree — see the
+# freshness gate below, kaappi#2197). Most dev boxes have neither runtime nor a
+# fresh module; CI's `wasm` job has both.
 #
 # ---------------------------------------------------------------------------
 # READ THIS BEFORE TRUSTING A GREEN RUN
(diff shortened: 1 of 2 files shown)


baijum:
Review follow-up on the CodeRabbit comments:

- **Fail closed when `find` cannot establish freshness** — applied (b421c069).
- **Report the full freshness scope** — applied (b421c069): now reads "verified newer than interpreter build inputs".
- **Require a verifiable fingerprint instead of mtimes** — not applied. The mtime gate is a deliberate, documented heuristic that catches the realistic case (a module built from an earlier commit, with newer sources present). A full fingerprint system would require emitting module metadata from `build.zig`, which is disproportionate for a dev-facing differential harness and is documented as a limitation in the gate's comment block.
- **Add regression coverage for the freshness gate** — not applied (nitpick; the gate is a dozen lines of straightforward shell, and the differential harness is itself test infrastructure).
- **Honor the selected binary in run-all.sh** — not applied. The hardcoded `KAAPPI=zig-out/bin/kaappi` is intentional (kaappi#2163): run-all.sh deliberately refuses to substitute a default build, and this PR only adds the `zig build wasm` step without touching that contract.

PR #2339, opened 2026-08-25
srfi-158: begin range generators with start; gflatten yields nothing for empty lists

### LINKED ISSUE #2055: [SRFI-158] make-range-generator does not begin the sequence with an inexact start — a large step annihilates it to 0.0
**Category:** SRFI
**Spec reference:** [SRFI 158](https://srfi.schemers.org/srfi-158/srfi-158.html), `make-range-generator` — "Creates a generator of a sequence of numbers. **The sequence begins with start**, increases by step (default 1), and continues while the number is less than end, or forever if end is omitted. If both start and step are exact, it generates exact numbers; otherwise it generates inexact numbers."
**Severity:** wrong-result

`make-range-generator`'s three-argument case coerces `start` with a round trip
through `step` — `(set! start (- (+ start step) step))`. That achieves the
exactness contagion the spec's second sentence requires (an exact `start` with
an inexact `step` becomes inexact), but it does not leave an already-inexact
`start` alone: the sequence then begins with something other than `start`.

The error is not bounded by one ulp. When `step` is much larger than `start`,
the addition rounds `start` away entirely and the subtraction returns zero, so
the sequence begins with `0.0` regardless of what `start` was.

**Minimal reproduction:**

```scheme
(import (scheme base) (srfi 158))

(car (generator->list (make-range-generator 0.1 3)))
;; expected: 0.1                    actual: 0.10000000000000009

(car (generator->list (make-range-generator 0.1 0.5 0.2)))
;; expected: 0.1                    actual: 0.10000000000000003

(car (generator->list (gtake (make-range-generator 1e-20 1.0 1.0) 1)))
;; expected: 1e-20                  actual: 0.0

(car (generator->list (gtake (make-range-generator 5.0 1e30 1e17) 1)))
;; expected: 5.0                    actual: 0.0
```

**Discriminating controls:**

- `(make-range-generator 3 8 2)` ⟹ `(3 5 7)` and `(make-range-generator 1/3 1 1/7)` ⟹ `(1/3 10/21 ...)` — exact arithmetic makes the round trip an identity, so only inexact `start` is affected.
- `(car (generator->list (gtake (make-range-generator 0.1) 1)))` ⟹ `0.1` — **the same procedure with the same `start`**, one argument instead of three. The one-argument form goes through `make-infinite-range-generator`, which has no round trip.
- `make-iota-generator` carries the identical "The sequence begins with start" wording and yields `start` untouched at every input above: `(make-iota-generator 3 0.1 0.2)` ⟹ `(0.1 0.30000000000000004 0.5)`, `(make-iota-generator 1 1e-20 1.0)` ⟹ `(1e-20)`.

The contagion the round trip exists for must be preserved by any fix — these
must keep holding:

```scheme
(exact? (car (generator->list (make-range-generator 0 5 0.5))))  ;; #f
(exact? (car (generator->list (make-range-generator 0.5 5 1))))  ;; #f
(exact? (car (generator->list (make-range-generator 0 5 1))))    ;; #t
```

An `(if (and (exact? start) (exact? step)) start (exact->inexact start))` does
both.

**Verified:** fresh ReleaseSafe build at `08828884`, isolated `KAAPPI_HOME`, cache cold. chibi-scheme 0.11 reproduces every line identically, so this is inherited from the SRFI's own sample implementation rather than introduced here.
**Source location:** `lib/srfi/158-impl.scm:73` (inside `make-range-generator`, defined at line 69)
**Disabled tests:** `tests/scheme/srfi/srfi158-audit.scm`, 4 assertions marked `;; FAIL: #this`

**Found by:** Systematic audit v2, Phase 3.6

### LINKED ISSUE #2057: [SRFI-158] gflatten raises on an empty list from its source instead of yielding nothing
**Category:** SRFI
**Spec reference:** [SRFI 158](https://srfi.schemers.org/srfi-158/srfi-158.html), `gflatten` — "Returns a generator that yields **the elements of the lists** produced by the given generator."
**Severity:** wrong-result (raises where it must yield)

A list with no elements contributes no elements. `gflatten` instead raises a
`car` type error the moment its source produces `'()`.

The refill in the generator body runs **once**, not until it has a non-empty
list:

```scheme
(define (gflatten gen)
  (let ((state '()))
    (lambda ()
      (if (null? state) (set! state (gen)))   ; <- one refill, not a loop
      (if (eof-object? state)
        state
        (let ((obj (car state)))              ; <- (car '()) when gen gave '()
          (set! state (cdr state))
          obj)))))
```

**Minimal reproduction:**

```scheme
(import (scheme base) (srfi 158))

(generator->list (gflatten (generator '(1) '() '(2))))
;; expected: (1 2)
;; actual:   type error in 'car': expected pair, got ()

(generator->list (gflatten (generator '() '(1))))   ;; expected (1)
(generator->list (gflatten (generator '(1) '())))   ;; expected (1)
```

This is not a contrived input. The natural way to write a filtering map over a
generator produces `'()` for every rejected element:

```scheme
(generator->list (gflatten (gmap (lambda (x) (if (even? x) (list x) '()))
                                 (make-range-generator 0 4))))
;; expected: (0 2)   actual: type error in 'car'
```

**Discriminating controls:** `(generator->list (gflatten (generator '(1) '(2))))` ⟹ `(1 2)` and `(generator->list (gflatten (generator '(1 2 3))))` ⟹ `(1 2 3)` — the same shapes with no empty sub-list are correct, so the defect is the empty-list arm specifically, not `gflatten` generally. `(generator->list (gflatten (generator)))` ⟹ `()` — an empty *source* is handled; only an empty *element* is not.

**Verified:** fresh ReleaseSafe build at `08828884`, isolated `KAAPPI_HOME`, cache cold. chibi-scheme 0.11 raises `car: not a pair` on all four inputs, so this is inherited from the SRFI's own sample implementation rather than introduced here.
**Source location:** `lib/srfi/158-impl.scm:198-206`, the refill at line 201
**Disabled tests:** `tests/scheme/srfi/srfi158-audit.scm`, 4 assertions marked `;; FAIL: #this` (the raising behaviour is pinned by three enabled `raises?` assertions in the meantime, so the fix flips both sets)

**Found by:** Systematic audit v2, Phase 3.6

baijum:
## What

Two fixes in `lib/srfi/158-impl.scm`, both deviations the SRFI 158 spec explicitly forbids and both inherited from the SRFI's own reference implementation (chibi-scheme 0.11 reproduces each identically, per the verification notes on both issues).

### #2057 — `gflatten` raised on an empty list from its source

The reference implementation's refill ran exactly once, so an empty list from the source reached `car` and raised a type error instead of contributing no elements. The spec: *"Returns a generator that yields the elements of the lists produced by the given generator"* — a list with no elements contributes none. This is the natural shape of a filtering map (`gmap` returning `'()` for every rejected element), which previously could not be flattened at all:

```scheme
(generator->list (gflatten (generator '(1) '() '(2))))
;; before: type error in 'car': expected pair, got ()   after: (1 2)
```

The refill is now a named-let loop that runs until state holds a non-empty list or eof. Exhaustion stickiness is preserved (pinned by the pre-existing audit assertions).

### #2055 — `make-range-generator` did not begin the sequence with an inexact start

The three-argument case coerced `start` with `(- (+ start step) step)`. That achieved the spec's exactness contagion but did not leave an already-inexact start alone — and the error is not bounded by one ulp: when `step` dwarfs `start` the addition rounds `start` away entirely and the subtraction returns `0.0`. The spec: *"The sequence begins with start"*:

```scheme
(car (generator->list (gtake (make-range-generator 1e-20 1.0 1.0) 1)))
;; before: 0.0   after: 1e-20
(car (generator->list (make-range-generator 0.1 3)))
;; before: 0.10000000000000009   after: 0.1
```

Now `start` is coerced with `exact->inexact` only when `step` is inexact; an inexact start passes through untouched, and exact/exact stays exact. All three contagion controls from the issue keep holding (`0 5 0.5` → inexact, `0.5 5 1` → inexact, `0 5 1` → exact).

Both fixes diverge deliberately from the reference implementation and from every port of it; the divergence and why is recorded in comments at both sites, as the issue discussions recommended.

## Tests

`tests/scheme/srfi/srfi158-audit.scm` (discovered by `tests/scheme/run-all.sh`'s `tests/scheme/srfi/*.scm` glob):

- The eight assertions the audit had parked under `;; FAIL: #2055` / `;; FAIL: #2057` markers are enabled, and the three `raises?` assertions that had pinned `gflatten`'s raising behavior in the meantime are removed.
- One new regression added: `gflatten` over consecutive empty sub-lists — the case a naive fix that refills only once more still misses.

## Evidence

- All nine regressions **fail on the old library** (reverted `lib/srfi/158-impl.scm` via `git stash`, rebuilt, reran: 9 FAIL lines, 346 passes).
- With the fix: `srfi158-audit.scm` reports **355 expected passes, 0 failures, exit 0**.
- All ten other suites importing `(srfi 158)` pass (`srfi158`, `srfi221`, `srfi168`, `srfi190`, `srfi194`, `srfi214`, `srfi225-audit`, `srfi252`, `srfi-import-order`, `environment-file-sld-2012`).
- `zig build test` green (exit 0).

No `.zig` files touched — pure portable-Scheme library change.

Closes #2057
Closes #2055

### DIFF
--- a/lib/srfi/158-impl.scm
+++ b/lib/srfi/158-impl.scm
@@ -66,11 +66,19 @@
 
 
 ;; make-range-generator
+;; "The sequence begins with start": coerce an exact start to inexact with
+;; exact->inexact when step is inexact, never with the reference
+;; implementation's (- (+ start step) step) round trip.  The round trip
+;; achieves the exactness contagion but does not leave an already-inexact
+;; start alone — it perturbs it by a rounding error, and when step dwarfs
+;; start it annihilates it to 0.0 entirely (#2055; chibi-scheme, which
+;; carries the reference code, reproduces both).
 (define make-range-generator
   (case-lambda ((start end) (make-range-generator start end 1))
                ((start) (make-infinite-range-generator start))
                ((start end step)
-                (set! start (- (+ start step) step))
+                (unless (and (exact? start) (exact? step))
+                  (set! start (exact->inexact start)))
                 (lambda () (if (< start end)
                              (let ((v start))
                               (set! start (+ start step))
@@ -195,15 +203,24 @@
                   v)))))
 
 ;; gflatten
+;; The refill loops until state holds a non-empty list (#2057): the
+;; reference implementation refills exactly once, so an empty list from the
+;; source reaches car and raises (chibi-scheme reproduces this).  The spec —
+;; "yields the elements of the lists produced by the given generator" —
+;; means a list with no elements contributes no elements.
 (define (gflatten gen)
   (let ((state '()))
     (lambda ()
-      (if (null? state) (set! state (gen)))
-      (if (eof-object? state)
-        state
-        (let ((obj (car state)))
-          (set! state (cdr state))
-          obj)))))
+      (let refill ()
+        (cond ((null? state)
+               (set! state (gen))
+               (refill))
+              ((eof-object? state)
+               state)
+              (else
+                (let ((obj (car state)))
+                  (set! state (cdr state))
+                  obj)))))))
 
 ;; ggroup
 (define ggroup
--- a/tests/scheme/srfi/srfi158-audit.scm
+++ b/tests/scheme/srfi/srfi158-audit.scm
@@ -13,9 +13,11 @@
 ;;; The three areas ranked highest going in — laziness/effect ordering,
 ;;; exhaustion stickiness, and infinite generators under bounded consumers —
 ;;; came back CLEAN, and are pinned below with counter-instrumented sources
-;;; rather than left as prose.  The bugs are elsewhere: an inexact `start`
-;;; that the sequence does not begin with, an empty sub-list that raises, and
-;;; a documented spec example that cannot run at all.
+;;; rather than left as prose.  The bugs were elsewhere: an inexact `start`
+;;; that the sequence did not begin with (#2055) and an empty sub-list that
+;;; raised (#2057) are fixed and pinned as passing regressions in sections 6
+;;; and 7; a documented spec example that cannot run at all (#2060) is still
+;;; pinned as raises? in section 15.
 ;;;
 ;;; Portability notes for the emulated CI legs:
 ;;;   * Never write (list (g) (g)) — R7RS leaves argument evaluation order
@@ -480,8 +482,12 @@
 ;;    "The sequence begins with start, increases by step (default 1) ...
 ;;     If both start and step are exact, it generates exact numbers;
 ;;     otherwise it generates inexact numbers."
-;;    The impl coerces start with (- (+ start step) step), which achieves the
-;;    exactness contagion but does not leave an already-inexact start alone.
+;;    Contagion without perturbation (#2055): an exact start is coerced with
+;;    exact->inexact only when step is inexact, so the sequence begins with
+;;    start itself.  The reference implementation's (- (+ start step) step)
+;;    round trip — which this port and chibi-scheme both carried — achieves
+;;    the contagion but perturbs an already-inexact start, annihilating it
+;;    to 0.0 when step dwarfs it.
 ;; ---------------------------------------------------------------------------
 
 (test-equal "range: exact start and step stay exact"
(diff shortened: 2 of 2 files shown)


baijum:

Both pure-library fixes are correct and minimal.

- **make-range-generator (#2055):** the coercion `(unless (and (exact? start) (exact? step)) (set! start (exact->inexact start)))` handles all four exactness combinations — exact/exact untouched; an inexact start passes through *exactly* (fixing both the 1-ulp perturbation and the 0.0 annihilation); exact start + inexact step gets the required contagion. Contagion controls and both annihilation repros now begin with `start`.
- **gflatten (#2057):** the `refill` named-let loops on `(null? state)` until a non-empty list or eof, so empty sub-lists (leading/interior/trailing/consecutive) contribute nothing; both `(refill)` calls are in tail position (no stack growth), and exhaustion stays sticky.

Tests flip three `raises?` pins to real `test-equal` regressions plus a new consecutive-empty-sub-lists case and four inexact-start assertions — all fail pre-fix. DCO + Co-Authored-By present. Low-risk `.sld` change; trivially splittable per-issue if you'd prefer, but one PR is fine.

PR #2340, opened 2026-08-25
fmt: pin the #2143 lexeme-glue and #2080 diagnostic fixes in fmt-adversarial

### LINKED ISSUE #2080: [Tooling] kaappi fmt reports a user syntax error as "internal error: formatting would change the program"
**Category:** Tooling · **Severity:** diagnostic quality · **Found by:** systematic audit v2, Phase 6A (#1890)

`kaappi fmt` reports an ordinary **user** syntax error as an **internal** error,
and throws away a precise diagnostic it already has.

## Reproduction

```bash
printf '(display #\\qqq)\n' > bad.scm

kaappi bad.scm
# bad.scm:1:15: read error[KP1005]: invalid character name

kaappi fmt --check bad.scm
# kaappi fmt: bad.scm: internal error: formatting would change the program; file left unchanged
# exit 1
```

Same for `fmt` without `--check`. The file is genuinely malformed; nothing about
the formatter is wrong.

## Why it happens

`formatFile` (`src/fmt.zig`) has two failure paths that both funnel into one
message:

```zig
if (!verifyRoundTrip(gc, source, formatted)) {
    reportFileError(path, "internal error: formatting would change the program; file left unchanged");
    return .failed;
}
```

`verifyRoundTrip` returns `false` for two structurally different reasons
(`readAllRooted` returns `false` on any read error):

1. the **original** does not read — a user syntax error the CST lexer happened
   to tolerate; and
2. the original and the formatted text read to *different* datum sequences — the
   genuine formatter bug this net exists to catch.

Only (2) is an internal error. (1) is the common case and is misattributed.

## Reproduces across every reader error class

| Input | `kaappi <file>` says | `kaappi fmt` says |
|---|---|---|
| `(display #\qqq)` | `KP1005: invalid character name` | internal error |
| `(display #\xZZ)` | `KP1005: invalid character name` | internal error |
| `(a . . b)` | `KP1008: '.' outside of a list` | internal error |
| `(display '(a<FF>b))` (U+000C) | `KP1002: unexpected character` | internal error |

The last one also shows the CST lexer and the real reader disagreeing on whether
form feed is whitespace (`fmt.Lexer.isSpace` includes `0x0c`,
`reader.isDelimiter` and `skipWhitespaceAndCommentsChecked` do not). The
round-trip guard correctly refuses to write, so nothing is corrupted — but the
user is told it is our fault.

## Suggested fix

Split the check: read the original first, and on failure report the reader's own
`KP1xxx` message with its line:column, exactly as `kaappi check` does. Reserve
the `internal error` wording for a real mismatch between two successfully-read
datum sequences. `docs/dev/fmt.md` § *The round-trip safety net* describes the
two cases as one and should be split alongside it.

**Verified:** fresh ReleaseSafe build at `1718a2e1`, macOS aarch64, cache
cleared. `fmt` and `fmt --check` both exit 1; the file is left untouched in both
(no data loss — this is a message-quality defect only).
**Source location:** `src/fmt.zig` `formatFile` / `runStdin` / `verifyRoundTrip` / `readAllRooted`.

### LINKED ISSUE #2143: [Tooling] kaappi fmt refuses (or reports a bogus syntax error on) legal Scheme where a #-led lexeme is glued to an identifier
**Category:** Tooling
**Spec reference:** `src/fmt.zig:146` — "A boundary between atoms. Mirrors `reader.Reader.isDelimiter` so the formatter carves the same lexemes the real reader does."
**Severity:** wrong-result (a readable file cannot be formatted; one spelling is reported as a syntax error in the user's file)

It mirrors `isDelimiter`, but `isDelimiter` is not what ends an identifier in
the real reader. `Reader.readSymbol` (`src/reader.zig:428`) runs
`while (isSubsequent(...))`, and `isSubsequent` is
`isInitial(c) or isDigit(c) or c ∈ {+ - . @}` — so the reader ends an
identifier at the first non-⟨subsequent⟩ byte. `fmt.Lexer.scanAtom`
(`src/fmt.zig:293`) runs `while (!isDelimiter(...))`.

`#` is a delimiter for neither and a subsequent for neither, so every `#`-led
lexeme glued to the end of an identifier splits differently: the reader sees
`ident` + `#…`, `fmt` sees `ident#` + whatever follows.

**Minimal reproductions.** All three are read without complaint by `kaappi`
itself; none can be formatted:

```bash
$ printf '(list a#;(b) c)\n' > x.scm
$ kaappi ast x.scm
(list a c)
$ kaappi fmt x.scm
kaappi fmt: x.scm: syntax error: unterminated list          # exit 1
```

`fmt` lexes `a#` as an atom, then `;(b) c)` as a **line comment**, and the
list never closes — so it reports a syntax error in a file that has none.

```bash
$ printf '(a b#|c|# d)\n' > y.scm
$ kaappi ast y.scm
(a b d)
$ kaappi fmt y.scm
kaappi fmt: y.scm: internal error: formatting would change the program; file left unchanged
```

Here `fmt` lexes `b#`, then `|c|` as a **piped symbol**, then `#` — five datums
where the reader sees three. Emitting them with its usual single spaces would
change the program, so the round-trip guard (correctly) refuses and the file is
simply unformattable.

```bash
$ printf '(a b#(1) d)\n' > z.scm && kaappi ast z.scm
(a b #(1) d)
$ kaappi fmt z.scm
kaappi fmt: z.scm: internal error: formatting would change the program; file left unchanged
```

**Discriminating control — one space fixes every case:**

```bash
$ printf '(a b #|c|# d)\n' | kaappi fmt
(a b #|c|# d)
$ printf '(list a #;(b) c)\n' | kaappi fmt
(list a #;(b) c)
```

Same datums, same comments, only the space differs.

**A silent variant.** `b#f` and `b#t` are *not* refused — `fmt` treats
`b#f` as one atom and emits it byte-for-byte, so no datum changes and the guard
passes — but its structural view is still wrong (`(a b#f)` is three datums to
the reader, two items to `fmt`), which only shows up as layout. `b'c`, ``b`c``
and `b,c` behave the same way.

**How it was found and how common it is.** A fuzzer that inserts comments at
token boundaries in the repo's own `.scm`/`.sld` files produced 604 refusals in
3 000 mutants; **100 % of them contain an identifier character immediately
followed by `#`**, and re-running with a space before every inserted comment
gave **0 refusals in 6 000**. Independently, a byte-level mutation run over the
same corpus (30 000 mutants) produced exactly 3 guard trips — all three the same
glue, e.g. `(import#(scheme base) …)` and
`(define shared-pair#(#2=(a) #2# #2#))`. One cause explains every interesting
outcome of both campaigns.

**Note on which lexing is "right".** Three implementations disagree about
`(a b#|c|# d)`, so this is not a straightforward conformance bug:

| | reads as |
|---|---|
| kaappi | `(a b d)` — `#\|` ends the identifier and opens a block comment |
| chibi 0.11 | `(a \|b#\\\|c\\\|#\| d)` — one symbol |
| guile 3.0 | `(a #{b#\|c\|#}# d)` — one symbol |
| `kaappi fmt` | `b#`, `\|c\|`, `#` — three lexemes, matching none of them |

R7RS 7.1.1 lists only whitespace, `\|`, `(`, `)`, `"` and `;` as ⟨delimiter⟩s
and does not admit `#` as a ⟨subsequent⟩, so the input is outside the grammar
either way. The bug is not that kaappi picked an answer — it is that `fmt`
picked a *different* one while documenting that it does not.

**Suggested shape of the fix:** end an atom the way `readSymbol` does — stop at
the first non-⟨subsequent⟩ byte — while keeping the existing carve-outs that the
comment above `scanAtom` already names (`#0#`, `#e#xFF`), which are `#`-**led**
lexemes and so are decided by `scanHash` before `scanAtom` is entered at all.

**Verified:** fresh ReleaseSafe build at 94ebd5a0, isolated `KAAPPI_HOME`;
`kaappi ast` exit codes checked without a pipeline.
**Source location:** `src/fmt.zig:146` (`Lexer.isDelimiter`), `src/fmt.zig:293`
(`scanAtom`) vs `src/reader.zig:270` (`isSubsequent`), `src/reader.zig:428`
(`readSymbol`).
**Disabled test:** `tests/scheme/fmt/fmt-adversarial.sh` (marked `# FAIL: #this`)

**Found by:** Systematic audit v2, Phase 6F (mutation + structural fuzzing of `kaappi fmt`)

baijum:
## What

Group H task for #2143 + #2080. **Both issues were already fixed on main by #2291 (commit `e5bd7747`) before this branch was cut** — this PR is **verify-and-close** for the code fixes and adds the regression coverage the issues' own bodies call for, which `fmt-adversarial.sh` (the file both issues name) did not yet have.

## Verify: neither issue reproduces on main @ 645d2bba

Every reproduction from both issue bodies, run against a fresh ReleaseSafe build of this branch:

| Issue repro | Pre-fix (as filed) | Now |
|---|---|---|
| `(list a#;(b) c)` | `syntax error: unterminated list` | formats to `(list a #;(b) c)`, exit 0 |
| `(a b#|c|# d)` | `internal error: formatting would change the program` | formats to `(a b #|c|# d)`, exit 0 |
| `(a b#(1) d)` | `internal error: …` | formats to `(a b #(1) d)`, exit 0 |
| g2.scm (`(display (list a#;(b) c))`) | `fmt --check`: `syntax error: unterminated list` | reported as merely unformatted; formats to `(display (list a #;(b) c))`, second pass clean |
| `(display #\qqq)` | `internal error: formatting would change the program` | `:1:15: read error[KP1005]: invalid character name` (fmt, fmt --check, stdin) |
| `#\xZZ` / `(a . . b)` / form feed U+000C | `internal error: …` | KP1005 / KP1008 / KP1002 with line:column |

## The coupling

#2143's mis-lex is what produced #2080's bogus message in half these cases: the CST lexer glued `ident#` into one atom, carved a different lexeme stream than the reader, and the resulting divergence surfaced through the single shared "internal error" funnel. The two fixes in `src/fmt.zig` (#2291) are two halves of one behavior: `scanIdentifier` now ends at the first non-⟨subsequent⟩ byte — delegating to `Reader.isSubsequent`/`isUnicodeSubsequent`, the reader's own functions rather than a copy — and `verifyRoundTrip` returns a `RoundTrip` union whose `original_unreadable` arm reports the reader's KP1xxx diagnostic instead of claiming an internal fault.

## Tests added

`tests/scheme/fmt/fmt-adversarial.sh` (+6 cases):

- #2143 — the two guard-tripping shapes from the issue's byte-mutation campaign, absent from the six existing glue cases: `(import#(scheme base))` and `(define shared-pair#(#2=(a) #2# #2#))` (a `#(` glued after an identifier once carved `…#` as one atom, leaving a list where the reader opens a vector).
- #2080 — the diagnostic path, previously untested in this suite (fmt.sh had one case, `#\qqq`): an `assert_reader_error` helper asserting fmt exits 1 with the reader's own `KP1xxx` diagnostic and **never** "internal error", over all four error classes of the issue's table (`#\qqq` KP1005, `#\xZZ` KP1005, `(a . . b)` KP1008, form feed KP1002).

Each new case was proven to fail without the fix by rebuilding with the pre-fix `src/fmt.zig` (`220e31a2`): all six fail with exactly the filed symptom (`kaappi fmt: internal error: formatting would change the program`); `fmt-adversarial: 72 passed, 9 failed` pre-fix vs **81 passed, 0 failed** post-fix. Also green: `tests/scheme/fmt/fmt.sh` (38/38) and `zig build test`.

Closes #2143
Closes #2080

### DIFF
--- a/tests/scheme/fmt/fmt-adversarial.sh
+++ b/tests/scheme/fmt/fmt-adversarial.sh
@@ -15,7 +15,7 @@
 #
 # The generator that found the failures below is `tools/fmt_fuzz.py` — it is
 # deliberately *not* run here (it takes minutes; this file takes ~2 seconds).
-# See kaappi#2141, kaappi#2142, kaappi#2143, kaappi#2149.
+# See kaappi#2141, kaappi#2142, kaappi#2143, kaappi#2080, kaappi#2149.
 
 set -uo pipefail
 
@@ -147,6 +147,39 @@ case_ "vector after an identifier, spaced"           '(a b #(1) d)\n'
 case_ "block comment glued to an identifier"         '(a b#|c|# d)\n'
 case_ "datum comment glued to an identifier"         '(list a#;(b) c)\n'
 case_ "vector glued to an identifier"                '(a b#(1) d)\n'
+# The two shapes from #2143's byte-mutation campaign that tripped the round-trip
+# guard outright: a `#(` glued after an identifier once carved `…#` as one atom,
+# leaving a bare `(` that opens a *list* where the reader opens a vector — a
+# different program, so the guard refused and the file was unformattable.
+case_ "vector open glued after import"               '(import#(scheme base))\n'
+case_ "datum labels inside a glued vector"           '(define shared-pair#(#2=(a) #2# #2#))\n'
+
+# ── A user syntax error is the reader's error, not an internal one ─────────
+#
+# The round-trip guard reads the *original* first: when that read fails — a
+# syntax error the CST lexer happened to tolerate — `fmt` must report the
+# reader's own KP1xxx diagnostic with its position, exactly as `kaappi check`
+# does. The "internal error: formatting would change the program" wording is
+# reserved for a genuine mismatch between two successfully-read texts and must
+# never appear on this path (kaappi#2080). All four are the error classes of
+# that issue's table; the file is left untouched in every case.
+
+# assert_reader_error <label> <expected-kp-code> <source-with-\n-escapes>
+assert_reader_error() {
+    local label="$1" kp="$2" src="$3" status=0 err
+    printf '%b' "$src" > "$TMP/e.scm"
+    err="$("$KAAPPI" fmt < "$TMP/e.scm" 2>&1)" || status=$?
+    if [[ "$status" -eq 1 && "$err" == *"read error[$kp]"* && "$err" != *"internal error"* ]]; then
+        pass "$label"
+    else
+        fail "$label" "exit $status, stderr: [$err]"
+    fi
+}
+
+assert_reader_error "user error: invalid character name"       KP1005 '(display #\\qqq)\n'
+assert_reader_error "user error: invalid hex char literal"     KP1005 '(display #\\xZZ)\n'
+assert_reader_error "user error: '.' outside of a list"        KP1008 '(a . . b)\n'
+assert_reader_error "user error: form feed is not whitespace"  KP1002 "(display '(a\fb))\n"
 
 # ── Fit-to-width is measured in columns (Unicode code points) ────────────
 


baijum:
**Review (Claude Code, built + ran the script)** — ✅ approve

This is a **test-only** PR, and correctly so: both #2143 and #2080 are still open, but the underlying `fmt.zig` behavior was already fixed by #2291 (on main). This PR pins that behavior with regression cases and closes the issues.

I ran `fmt-adversarial.sh` against a current-main `kaappi`: **81 passed, 0 failed**. The added cases genuinely match the issue repros:
- #2143: `(import#(scheme base))` and `(define shared-pair#(#2=(a) #2# #2#))` — the `#(`-glued-to-identifier shapes that previously carved `…#` as one atom and left a bare `(`, making the file unformattable.
- #2080: four `assert_reader_error` cases confirm a user syntax error is reported as `read error[KP1xxx]` with position and **never** as `internal error: formatting would change the program` (that wording stays reserved for a genuine two-text mismatch).

No code change, so no convention/GC/size concerns. Legitimate close. _(Worth noting for the record: because the fix rode in on #2291, the value here is purely the regression guard — which is real and now green.)_

Issue #1886, opened 2026-07-31
Nested `guard` past 64 levels silently returns a wrong result — MAX_HANDLERS overflow is catchable and misreported as KP9002


baijum:
**Kaappi version**

`v0.22.1` — local `zig build` of `main` @ `b3d13a08`. Also reproduces identically on the `v0.22.0` release binary.

**Platform**

macOS aarch64 (Darwin arm64)

**Build mode**

ReleaseSafe (default `zig build`)

**Description**

`MAX_HANDLERS` ([`src/types_continuation.zig:49`](https://github.com/kaappi/kaappi/blob/main/src/types_continuation.zig#L49)) caps the exception-handler stack at 64. `VM.pushHandler` ([`src/vm.zig:959`](https://github.com/kaappi/kaappi/blob/main/src/vm.zig#L959)) correctly returns `VMError.StackOverflow` past that — but the overflow never reaches the user *as* a stack overflow. It arrives as an ordinary, catchable Scheme error object whose message is the bare string `"error"`.

Because it is catchable, an **enclosing `guard` in the user's own program swallows it**, and the program returns a plausible-looking wrong value and exits 0. Nothing indicates a limit was hit.

Three separable problems, in descending severity:

1. **Silent wrong answer.** Past 64 nested handlers, `guard` returns a value from the wrong handler. No error, exit 0. A program that nests `guard` in a recursive function is silently incorrect rather than failing.
2. **Wrong diagnostic when it does surface.** It is reported as `KP9002 out-of-memory`, which is misleading — no allocation failed. `diagnostics.zig:639` already maps `error.StackOverflow => .stack_overflow` (KP3008), so something upstream converts the error into a raised Scheme error object and loses its identity before that mapping applies.
3. **The cap is low and hard-coded.** `MAX_FRAME_LIMIT` (32768) and `MAX_REGISTER_LIMIT` (65536) are configurable through `build_options` (`max_frames`, `max_registers`); `MAX_HANDLERS` and `MAX_WINDS` are fixed at 64 with no build option. 64 dynamically nested `guard`s is not an unreasonable amount for recursive code.

**Minimal reproduction**

```scheme
(import (scheme base) (scheme write))

;; Each level wraps the recursive call in its own guard.  The innermost call
;; (n = 0) raises, and its OWN guard should always catch it — so every depth
;; must return 0.
(define (f n)
  (guard (e (#t n))
    (if (= n 0) (raise 'b) (f (- n 1)))))

[...]

PR #2342, opened 2026-08-25
Preserve fiber fault identity and fix (kaappi fibers) argument diagnostics

### LINKED ISSUE #2002: [Fibers] Argument diagnostics misidentify the problem: make-channel's range error claims a type error, and a bad channel timeout is blamed on 'thread'
**Category:** KEP-0002 fibers / error taxonomy (audit dimension D2–D3) · **Severity:** diagnostic quality · **Found by:** Systematic audit v2, Phase 2.6 (#1890)

Two independent sites where a `(kaappi fibers)` argument rejection describes the wrong thing. Grouped because they are the whole of this library's argument-diagnostic surface and a taxonomy sweep (Phase 2.13) will want them together; the fixes are separate.

## 1. `make-channel` reports a range rejection as a type error, and its "expected" text describes exactly the value it got

```scheme
(make-channel 4294967296)
;; error[KP3002]: type error in 'make-channel': expected non-negative exact integer, got 4294967296
```

`4294967296` **is** a non-negative exact integer. The message is self-contradictory as written, and gives the reader nothing to act on — the real constraint is "≤ 2³²−1", which appears nowhere. Same for a bignum (`got #<bignum>`) and for `-1`.

Per `CLAUDE.md`'s own rule, `argError` (`InvalidArgument`, KP3007) is "for a value of acceptable type that the procedure rejects anyway"; `-1` and `4294967296` are exactly that. `typeError`/KP3002 is right only for `1.0` and `'x`.

**Discriminating control:** `(make-channel 4294967295)` ⟹ a channel. The cliff is the u32 bound, i.e. a range check reported as a type check.

**Source:** `makeChannelFn`, `src/primitives_fiber.zig:111-115` — one `typeError` covers the non-fixnum case *and* both range failures.

## 2. A bad timeout to `channel-send` / `channel-receive` is blamed on a procedure called `thread`

```scheme
(channel-receive (make-channel) 'bad)
;; error[KP3002]: type error in 'thread': expected time object or number, got #<symbol>
(channel-send (make-channel) 1 'bad)
;; error[KP3002]: type error in 'thread': expected time object or number, got #<symbol>
```

There is no procedure named `thread` — not in `(kaappi fibers)`, not in `(srfi 18)`. The name is hardcoded in the shared helper `timeoutToDeadlineNs` (`src/primitives_srfi18.zig:242`), which `channel-send`/`channel-receive` call for their optional timeout argument. A user of `(kaappi fibers)` who has never imported SRFI 18 gets a diagnostic naming a procedure that does not exist, in a library they are not using.

**Discriminating control:** `(thread-sleep! 'bad)` ⟹ `type error in 'thread-sleep!': got bad` — the sibling site passes its own name through correctly, so this is a missing parameter, not a house style.

**Scope:** `thread-join!`'s timeout argument reports `'thread'` too, so a fix should thread the caller's name through `timeoutToDeadlineNs` rather than patching the fiber call sites.

**Adjacent, different root cause (not filed here, for Phase 2.7):** `(mutex-lock! (make-mutex) 'bad)` returns `#t` — it accepts a symbol as a timeout and silently ignores it, rather than reaching this helper at all.

## Verified

Fresh ReleaseSafe build at `3812d735`, isolated `KAAPPI_HOME`; both messages quoted verbatim from the binary.

**Tests:** `tests/scheme/audit/primitives_fiber-audit.scm` pins the rejections themselves as enabled `raises?` assertions (the behaviour is right; only the wording is wrong), so a fix to either site keeps the suite green.

### LINKED ISSUE #2204: [Fibers] A VM-level fault in a fiber body loses its code and message: fiber-join reports KP3007 "fiber error (no exception value)"
**Category:** KEP-0001 fibers · **Severity:** wrong-result (a substituted condition, not merely a reworded one) · **Priority:** `medium` — see the reasoning at the bottom

`fiber-join` re-raises a fiber's Scheme-level `raise`/`error` faithfully, but a
**VM-level fault** (type error, unbound variable, bad index, arity mismatch)
arrives with both its error code and its message replaced by a placeholder. The
real diagnostic is destroyed, not just reworded — there is no way to recover it
from Scheme.

## Minimal reproduction

```scheme
(import (scheme base) (kaappi fibers))
(define five 5)
(fiber-join (spawn (lambda () (car five))))
```

```text
error[KP3007]: fiber error (no exception value)
```

Expected: the same diagnostic the identical expression produces outside a fiber
— `error[KP3002]: type error in 'car': expected pair, got 5`.

## Which faults survive, and which do not

| fault in the fiber body | reported |
|---|---|
| `(car five)` | `KP3007` "fiber error (no exception value)" |
| `(no-such-var)` | `KP3007` "fiber error (no exception value)" |
| `(vector-ref (vector 1) 9)` | `KP3007` "fiber error (no exception value)" |
| `((lambda (x) x))` (arity) | `KP3007` "fiber error (no exception value)" |
| `(/ 1 zero)` | `KP3004` "division by zero" — **correct** |
| `(error "boom" 1 2)` | "boom", irritants `(1 2)` — **correct** |
| `(raise 'sym)` | `sym`, still a symbol — **correct** |

The split is not arbitrary, and it is the whole diagnosis (see below): a fault
that travels as a **condition object** survives the fiber boundary; a fault that
travels as a bare `VMError` tag plus `vm.last_error_detail` does not.
`(/ 1 zero)` is in the first group only because
`primitives_arithmetic.raiseDivByZero` (`src/primitives_arithmetic.zig:177`)
deliberately allocates a real coded `ErrorObject` and sets
`vm.current_exception`.

## Discriminating controls

```scheme
;; 1. The identical fault outside a fiber keeps everything:
(car five)          ;; error[KP3002]: type error in 'car': expected pair, got 5

;; 2. Scheme-level conditions cross the fiber boundary intact (table above).

;; 3. It is genuinely the .errored branch, not a one-off race — joining the
;;    SAME fiber a second time can only reach reraiseFiberError, and repeats it:
(let ((f (spawn (lambda () (car five)))))
  (list (guard (e (#t (error-object-message e))) (fiber-join f))
        (guard (e (#t (error-object-message e))) (fiber-join f))))
;; ⟹ ("fiber error (no exception value)" "fiber error (no exception value)")
```

An interposed `(yield)` before the fault changes nothing, so this is not
specific to whether the fiber was dispatched by the scheduler loop or driven in
place.

## Cause

Three steps, no single obvious culprit:

1. **The fault is discarded.** The dispatch loop's error arm
   (`src/fiber_wait.zig:411-429`) catches `err` from `vm.runUntil`, and after
   the `Yielded` case it never looks at `err` again — it only sets
   `.errored` via `retireSlot`. The `VMError` tag, and with it the error
   *code*, is dropped on the floor.
2. **Only a condition object is carried across.**
   `FiberScheduler.saveCurrentFiber` copies exactly one error channel —
   `fiber.current_exception = vm.current_exception` (`src/fiber.zig:596`).
   `vm.last_error_detail`, where a type error's entire text lives, is not
   part of the fiber's saved state.
3. **The joiner substitutes a placeholder.** `reraiseFiberError`
   (`src/primitives_fiber.zig:116-124`) re-raises `fiber.current_exception`
   when it is non-null; otherwise it overwrites the detail with
   `"fiber error (no exception value)"` and returns
   `PrimitiveError.InvalidArgument` — which is why the code becomes `KP3007`
   (invalid argument) regardless of what actually happened.

## Suggested fix

`primitives_control.nativeErrorToErrorObject` is the existing precedent: it
turns a `VMError` into a coded `ErrorObject` from `vm.getErrorDetail()` plus
`diagnostics.runtimeErrorCode(err)`, and it is what makes `guard` able to catch
`(car 5)` at all. The dispatch loop's error arm wants the same conversion, so
that step 1 stores a real condition instead of discarding one and
`reraiseFiberError`'s fallback becomes genuinely unreachable.

Two constraints for whoever takes it:

- The detail has to be captured **at the point of failure**, before anything
  else runs — `vm.last_error_detail` is cleared by the next native call, and
  `retireSlot`/`abandonFiberMutexes`/`wakeWaiters` all run in between.
- The conversion allocates, and it must land in the **joining thread's** heap
  for `fiber-join` to hand it back (a fiber cannot cross an OS-thread boundary
  at all, so this is the same heap — but the rooting still has to be right;
  `nativeErrorToErrorObject` returns an *unrooted* object by contract).

An uncatchable error (`errors.isUncatchable` — `StackOverflow`,
`ExecutionTimeout`, `Terminated`, `Yielded`) must not be converted into a
catchable condition on the way through.

## History

#564 (closed) fixed the *swallowing* — errors in spawned fibers used to be
discarded entirely and `fiber-join` returned VOID. This is the residual: the
error now propagates, but its identity does not.

Noticed while fixing #1999, which cited this message as one of its four
symptoms (a pure-variadic `spawn` thunk failed inside the fiber and reported
it). #1999's fix removes that trigger, so the placeholder no longer appears
for wrong-arity thunks — the reporting hole itself is independent and
unchanged.

## Test location

`tests/scheme/audit/primitives_fiber-audit.scm` already asserts the half that
works ("fiber-join re-raises the fiber's exception with the message intact",
for `(error "fiber-boom")`); the VM-fault half has no assertion at all, which
is why this survived the audit. A regression test belongs beside it, plus a
unit test in `src/tests_fibers.zig`.

## Verified

Fresh ReleaseSafe build at `a8d70363`, isolated `KAAPPI_HOME`, cache cleared.
Deterministic across runs and unaffected by an interposed `yield`.

## Why `medium`

Not `critical`: no memory unsafety and no abort. Not `high`: no legal program
gets a silently wrong *answer*, no hang, no data loss — the error is still
raised and still catchable, so control flow is correct. `medium` rather than
`low` because more than the wording is wrong: the condition object handed to
`guard` is a **different, substituted one** carrying a different error code
(`KP3007` for what was a `KP3002`), so a clause discriminating on the
condition cannot match, and the original is unrecoverable. It is loud rather
than silent, which per the rubric keeps it from moving up.

baijum:
## What

Two issues in fiber error/argument reporting, one PR:

### #2204 — a VM-level fault in a fiber body loses its code and message

`(fiber-join (spawn (lambda () (car five))))` reported `KP3007 "fiber error (no exception value)"` — a *substituted* condition, not the real `KP3002 "type error in 'car': expected pair, got 5"` the identical expression produces outside a fiber. Three steps, per the issue's diagnosis: the dispatch loop's error arm (`src/fiber_wait.zig`) discarded the `VMError` tag; only `vm.current_exception` (the condition-object channel) crosses the fiber boundary via `saveCurrentFiber`; and `reraiseFiberError` substituted the placeholder.

**Fix:** the dispatch loop's error arm now converts the fault into a coded `ErrorObject` — using the same `nativeErrorToErrorObject` (newly `pub`) that `withExceptionHandlerFn` uses, so the error-coding boundaries cannot drift — *at the point of failure*, before `retireSlot`/`abandonFiberMutexes`/`wakeWaiters` or any later native call can overwrite `vm.last_error_detail`, and stages it in `vm.current_exception`, the one error channel `saveCurrentFiber` already transports into `fiber.current_exception` for the joiner to re-raise. GC-safe: the assignment lands directly in a marked VM root with no allocating call in between.

Deliberately unchanged: Scheme-level raises (already carried intact), `ContinuationInvoked` (control flow, not a fault), and uncatchable errors (`StackOverflow`, `ExecutionTimeout`, `Terminated` via `errors.isUncatchable`) which must not become catchable conditions (#1886).

| fault in the fiber body | before | after |
|---|---|---|
| `(car five)` | KP3007 placeholder | KP3002 `type error in 'car': expected pair, got 5` |
| unbound variable | KP3007 placeholder | KP3001 `undefined variable '...'` |
| `(vector-ref (vector 1) 9)` | KP3007 placeholder | KP3006 `vector-ref: index 9 out of range for length 1` |
| call arity mismatch | KP3007 placeholder | KP3003 `expected 1 arguments, got 0` |
| `(/ 1 0)`, `(error ...)`, `(raise ...)` | already correct | unchanged |

### #2002 — argument diagnostics misidentify the problem

1. `make-channel`'s u32 capacity range rejection was a *type* error whose text ("expected non-negative exact integer") described exactly the value it got, hiding the real bound. Now `argError` (KP3007): `"make-channel: capacity must be an exact integer between 0 and 4294967295, got 4294967296"`. A bignum is handled as the range case it is; only genuinely non-integer arguments (`1.0`, `'x`, `1/2`) stay `typeError`/KP3002.
2. A bad timeout to `channel-send`/`channel-receive` was blamed on a nonexistent procedure `'thread'` — `timeoutToDeadlineNs`'s hardcoded name. The caller now passes its own name through the helper, which also fixes the same message from `thread-join!`/`mutex-lock!`/`mutex-unlock!` timeouts. (`mutex-lock!`'s silently-accepted bad timeout remains, as the issue notes, a separate Phase 2.7 item.)

Also made `primitives.safeValueDescription` pub so the range rejection can name the offending value without duplicating the defensive renderer.

## Tests

- `tests/scheme/audit/primitives_fiber-audit.scm`: +20 assertions (129 → 149 passes) — fault codes/messages per class, interposed-yield, join-twice, guard-inside-fiber control, outside-fiber control, make-channel taxonomy/bound, timeout naming.
- `src/tests_fibers.zig`: +5 tests — exact code+message fidelity for `(car five)` (direct vs joined), per-class codes outside vs inside a fiber, join-twice, make-channel codes + bound substrings, timeout naming (including a negative check that `'thread'` no longer appears).

**Fail-without-fix verified** by stashing the five source fixes and keeping the tests: 15/20 new audit assertions and all 5 new Zig tests fail on the pre-fix tree; all pass with the fixes.

## Evidence

- `zig build test`: **1827/1834 pass, 7 skipped, 0 fail** (full suite, single foreground run)
- fiber filter under `-Dgc-stress=true`: 36/36
- Audit suites with the worktree binary: `primitives_fiber-audit` 149, `primitives_srfi18-audit` 287, `srfi18-deepcopy-matrix-audit` 127, `cross-thread-boundary-audit` 51 — all green
- All 22 `tests/scheme/smoke/fiber-*.scm` green
- `zig fmt --check` clean on every touched file

## BASELINE note (for reviewers)

The bare error-taxonomy gate moves **28 → 27**: `reraiseFiberError`'s fallback return is now annotated `bare-ok` — it sets its own detail via `setErrorDetail`, and since #2204 it is only reachable for uncatchable faults and conversion OOM. `.github/workflows/ci.yml` updated. Expected trivial conflict with sibling PRs touching the same line; resolve by re-counting on the rebased tree.

Closes #2204
Closes #2002

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,7 +149,7 @@ jobs:
       # write `error.TypeError` in `expectError`, so src/tests_*.zig is excluded.
       - name: Check bare error-taxonomy regression
         run: |
-          BASELINE=28
+          BASELINE=27
           hits=$(grep -rnE 'return (PrimitiveError|VMError|error)\.(TypeError|IndexOutOfBounds|InvalidArgument)' src/*.zig \
                  | grep -v '^src/tests_' | grep -v 'bare-ok' || true)
           count=$(printf '%s' "$hits" | grep -c . || true)
--- a/src/fiber_wait.zig
+++ b/src/fiber_wait.zig
@@ -12,6 +12,8 @@ const types = @import("types.zig");
 const vm_mod = @import("vm.zig");
 const reactor_mod = @import("reactor.zig");
 const fiber_mod = @import("fiber.zig");
+const errors = @import("errors.zig");
+const primitives_control = @import("primitives_control.zig");
 const Value = types.Value;
 const VM = vm_mod.VM;
 const VMError = vm_mod.VMError;
@@ -451,6 +453,40 @@ pub fn runSchedulerStep(comptime Ctx: type, ctx: Ctx, vm: *VM, sched: *FiberSche
                     sched.markRunnable(fiber);
                 }
                 continue;
+            }
+            // #2204: a VM-level fault (type error, unbound variable, bad
+            // index, arity mismatch) carries its identity in two places the
+            // fiber's saved state never transports — the VMError tag and
+            // vm.last_error_detail — so fiber-join used to substitute a
+            // contentless KP3007 "fiber error (no exception value)" for the
+            // real diagnostic. Convert the fault HERE, before anything else
+            // runs (retireSlot/abandonFiberMutexes/wakeWaiters, or any later
+            // native call, can overwrite the detail), into the same coded
+            // ErrorObject withExceptionHandlerFn hands a guard, and stage it
+            // in vm.current_exception — the one error channel
+            // saveCurrentFiber copies into fiber.current_exception below,
+            // which is what the joiner's reraiseFiberError re-raises. A
+            // Scheme-level raise (ExceptionRaised) already carries its
+            // condition that way; a continuation jump (ContinuationInvoked)
+            // is control flow, not a fault; and an uncatchable error
+            // (StackOverflow, ExecutionTimeout, Terminated —
+            // errors.isUncatchable) must not become a catchable condition
+            // on its way through (#1886), so none of those convert.
+            if (err != VMError.ExceptionRaised and
+                err != VMError.ContinuationInvoked and
+                !errors.isUncatchable(err))
+            {
+                if (primitives_control.nativeErrorToErrorObject(vm, vm.gc, err)) |exc| {
+                    // Assigning straight into vm.current_exception (a GC
+                    // root, vm.zig markVmRoots) with no allocating call in
+                    // between is what roots the otherwise-unrooted object
+                    // the conversion contract warns about.
+                    vm.current_exception = exc;
+                } else |_| {
+                    // The conversion itself failed (OOM): keep the retire
+                    // bookkeeping and let the joiner see the fallback
+                    // placeholder — better a vague error than a lost fiber.
+                }
             }
             // Fiber 0 is the main fiber: finishing or aborting one
             // top-level form is not thread death, so its mutexes stay
--- a/src/primitives.zig
+++ b/src/primitives.zig
@@ -622,7 +622,10 @@ pub fn parseOptionalRange(args: []const Value, arg_offset: usize, max_len: usize
 ///       one-level summary (a vector/bytevector reports its length; a rational
 ///       renders its two exact-integer components), never a recursive print —
 ///       so a cyclic structure cannot make this loop.
-fn safeValueDescription(buf: *[128]u8, value: Value) []const u8 {
+/// `pub` since #2002: primitives_fiber.makeChannelFn's range rejection needs
+/// the offending value in its argError message ("…between 0 and 4294967295,
+/// got <value>") without duplicating this deliberately-defensive renderer.
+pub fn safeValueDescription(buf: *[128]u8, value: Value) []const u8 {
     var w: std.Io.Writer = .fixed(buf);
     describeValue(&w, value);
     const out = w.buffered();
--- a/src/primitives_control.zig
+++ b/src/primitives_control.zig
@@ -168,7 +168,12 @@ pub fn raiseContinuable(vm: *vm_mod.VM, obj: Value) PrimitiveError!Value {
 /// two error-coding boundaries cannot drift apart.
 ///
 /// The returned object is *unrooted* — root it before allocating again.
-fn nativeErrorToErrorObject(vm: *vm_mod.VM, gc: *memory.GC, err: anyerror) PrimitiveError!Value {
+///
+/// `pub` since #2204: the fiber dispatch loop's error arm (fiber_wait.zig)
+/// runs the same conversion so a VM-level fault inside a fiber keeps its
+/// code and message across the fiber boundary — the fourth error-coding
+/// boundary, deliberately the same function so none of them can drift.
+pub fn nativeErrorToErrorObject(vm: *vm_mod.VM, gc: *memory.GC, err: anyerror) PrimitiveError!Value {
     const detail = vm.getErrorDetail();
     var msg_str = if (detail.len > 0)
         gc.allocString(detail) catch return PrimitiveError.OutOfMemory
--- a/src/primitives_fiber.zig
+++ b/src/primitives_fiber.zig
@@ -119,8 +119,16 @@ fn reraiseFiberError(fiber: *fiber_mod.Fiber) PrimitiveError {
         vm.current_exception = exc;
         return PrimitiveError.ExceptionRaised;
     }
+    // Since #2204 the dispatch loop converts every catchable VM-level fault
+    // into a coded ErrorObject before the fiber is retired, so this fallback
+    // is reached only for faults that deliberately carry no condition across
+    // the boundary: an uncatchable one (StackOverflow, ExecutionTimeout,
+    // Terminated — errors.isUncatchable, which must stay uncatchable) or the
+    // conversion itself failing. The detail below is set by hand, so this is
+    // a described error, not the synthesized blame-a-real-argument kind the
+    // bare-return gate exists to catch.
     vm.setErrorDetail("fiber error (no exception value)", .{});
-    return PrimitiveError.InvalidArgument;
+    return PrimitiveError.InvalidArgument; // bare-ok: detail set above; uncatchable-fault fallback (#2204)
 }
 
 fn fiberPredFn(args: []const Value) PrimitiveError!Value {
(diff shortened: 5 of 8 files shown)


baijum:

Both issues fully addressed. #2204: a catchable VM fault in a fiber body is now converted to a coded `ErrorObject` via the same `nativeErrorToErrorObject` a `guard` uses, staged in `vm.current_exception` *before* `retireSlot`/`wakeWaiters` can clobber `last_error_detail`, so the join preserves the original CODE **and** MESSAGE (tests assert exact KP3002/KP3001/KP3006/KP3003 identity, both driven-in-place and post-yield paths). #2002: `make-channel` range error → `argError` (KP3007) naming the real bound; `timeoutToDeadlineNs` gained a `proc` param and all 5 callers updated, fixing the 'thread' mislabel.

- Fibers share one VM/GC/heap (single OS thread), so the reraised error object is same-heap — no deep-copy/`Object.owner` boundary crossed. GC-root assignment path checks out.
- **Ratchet:** `BASELINE` 28→27 (the reraise fallback gained `// bare-ok:`). ⚠️ **Conflicts with #2347** on the same `ci.yml` `BASELINE=` line (that PR →10) — see note there; merge order matters.
- `primitives_fiber.zig` 1409→1430, under the 1500 cap.

Nit: an *uncatchable* fault (StackOverflow/Timeout) in a fiber still surfaces as a catchable KP3007 "no exception value" at the join — pre-existing, unchanged here, but worth a follow-up issue.

_Recommend confirming the gc-stress fiber leg is green before merge (the fix adds an allocating conversion on a root-assignment path); rooting reasoning holds on inspection._

Issue #1988, opened 2026-07-31
[R7RS 4.2.7] A guard that declines leaves its dynamic environment installed, so the enclosing guard's clauses run in the wrong one


baijum:
## Summary

When a `guard` declines to handle a condition, the *enclosing* `guard` evaluates its
clauses in the **declining guard's** dynamic environment rather than its own. R7RS
4.2.7 requires the clauses to run in the dynamic environment of their own `guard`
expression.

It is visible two ways: a `parameterize` between the two guards leaks its value into
the outer guard's clauses, and a `dynamic-wind` in the same position runs its `after`
thunk *after* the outer clauses instead of before them.

A `guard` that *handles* is fine, and a declining guard with no dynamic extent between
it and the outer guard is fine — which is why this survives the obvious tests.

## Repro

```scheme
(define p (make-parameter 1))

;; The outer guard is outside the parameterize, so its clause must see p = 1.
(guard (e (#t (p)))
  (parameterize ((p 2))
    (raise 'boom)))                          ; => 1   correct

;; Adding a guard that declines changes the answer.
(guard (e (#t (p)))
  (parameterize ((p 2))
    (guard (e ((number? e) 'no-match))       ; declines: 'boom is not a number
      (raise 'boom))))                       ; => 2   should be 1
```

Same defect through `dynamic-wind`:

```scheme
(define log '())
(guard (e (#t (set! log (cons 'clause log))))
  (dynamic-wind
    (lambda () (set! log (cons 'before log)))
    (lambda () (guard (e ((number? e) 'no-match)) (raise 'x)))
    (lambda () (set! log (cons 'after log)))))
(reverse log)
; => (before clause after)   should be (before after clause)
```

The full script is attached below; on `main` it prints:

```
ok   no inner guard                expected 1, got 1
FAIL declining inner guard         expected 1, got 2
FAIL declining, raise deeper       expected 1, got 2
FAIL two declining guards          expected 1, got 2
FAIL declining, raise-continuable  expected 1, got 2
ok   declining, no extent between  expected 1, got 1
ok   wind order, no inner guard    expected (before after clause), got (before after clause)
FAIL wind order, declining guard   expected (before after clause), got (before clause after)
ok   restored after the guard      expected 1, got 1
```

## Spec

R7RS-small 4.2.7:

> That implicit `cond` expression is evaluated with the continuation and dynamic
> environment of the `guard` expression. If every `⟨cond clause⟩`'s `⟨test⟩` evaluates
[...]

Issue #1375, opened 2026-07-09
Follow-up polish for the map/for-each/dynamic-wind/force trampoline (PR #1374)


baijum:
Follow-up to PR #1374 (which trampolined the `map`/`for-each`/`dynamic-wind`/`force` callback family via `src/vm_bootstrap.zig`, closing #1347). The PR's blocking/major findings were all resolved before merge; this issue tracks the remaining **non-blocking polish** items surfaced during review. None are correctness blockers — the full suite is green.

### 1. Confusing error messages for zero-sequence misuse (`priority: low`)

Calling the Scheme-bootstrapped iteration procedures with too few sequence arguments leaks the implementation instead of a clean arity/type error (all still **catchable** — only diagnostics degrade):

```scheme
(map car)          ; ⇒ "type error in 'cdr': expected pair, got ()"   (want: "'map': expected at least 2 arguments, got 1")
(vector-map +)     ; ⇒ "type error in 'make-vector': expected exact non-negative integer, got #f"
(map 5 '(1 2 3))   ; ⇒ "not a procedure"                              (want: "map: procedure")
```

Root cause: the bootstrap sources in `src/vm_bootstrap.zig` (e.g. `(if (null? (cdr lists)) …)`) surface internal errors. Consider up-front arg validation in the Scheme sources, or keeping a thin native arity/type guard.

### 2. Internal `%`-helpers still globally callable (`priority: low`)

PR #1374 moved `%push-wind`/`%pop-wind`/`%promise-forced?`/`%promise-merge!`/… to the `.internal` lib so they're no longer **exported** from `(scheme base)`/`(scheme lazy)`. But `registerAll` pre-binds every primitive into `vm.globals`, so they remain directly callable without any import:

```scheme
(procedure? %push-wind)      ; ⇒ #t   (no import needed)
```

`%push-wind` without a matching `%pop-wind` corrupts the wind stack; the `%promise-*` mutators can corrupt promise state. Fully hiding them would require gating global visibility by import (a broader change to Kaappi's global model) — filing as low-priority for consideration.

### 3. Iteration procedures now late-bind to mutable globals (`priority: low`)

`map`/`for-each`/`vector-map`/… now resolve `apply`, `reverse`, `cons`, `error`, `car`/`cdr`, arithmetic as globals at call time, so a top-level redefinition silently breaks them (native C impls were immune):

```scheme
(define reverse (lambda (x) 'HIJACKED))
[...]

PR #2343, opened 2026-08-25
control: align tail-position/dynamic-wind error taxonomy; purge %unwind-to-escape (#2036, #2037)

### LINKED ISSUE #2036: [Diagnostics] Three control-flow error paths diverge: tail-position call/cc reports KP3005 'error', dynamic-wind's type errors are uncoded, and a tail-position arity mismatch is KP2001
**Category:** Tooling / diagnostics (audit dimension D2 — error taxonomy, D3 — diagnostic fidelity)
**Spec reference:** KEP-0005 §4 — `error-object-code` recovers "the stable diagnostic code the implementation stamped on `obj`". `CLAUDE.md` (adding-features): report type errors with `primitives.typeError(proc, expected, got)`, "never a bare `return PrimitiveError.TypeError` — it names the expected type and the offending value".
**Severity:** edge-case (diagnostic quality), but two of the three make a *correctly written* `guard` dispatch on `error-object-code` take the wrong branch

Three control-flow error paths report differently from every other primitive in the file.

### 1. `call/cc` in tail position: KP3005 and the message `"error"`

```scheme
(import (scheme base) (scheme write) (kaappi diagnostics))
(define (show e) (list (error-object-code e) (error-object-message e)))
(write (guard (e (#t (show e)))       (call/cc 42)))  (newline) ; (KP3005 "error")
(write (guard (e (#t (show e))) (list (call/cc 42)))) (newline) ; (KP3002 "type error in 'call/cc': expected procedure, got 42")
```

One expression, two positions, two codes — and in tail position the message carries no information at all. `tail_call_cc`'s final `else` arm (`vm_dispatch.zig:1299`) is `return VMError.NotAProcedure;` with no `setErrorDetail`, so `nativeErrorToErrorObject` falls back to the literal string `"error"`. The non-tail path goes through `callWithCurrentContinuation`, which calls `primitives.typeError("call/cc", "procedure", args[0])`.

**Discriminating control:** `call/ec` has no tail superinstruction and answers `KP3002 type error in 'call/ec': expected procedure, got 42` in *both* positions.

### 2. `dynamic-wind` argument type errors carry no code and drop the value

```scheme
(write (guard (e (#t (show e))) (dynamic-wind 1 (lambda () 2) (lambda () 3))))
;; (#f "type error in 'dynamic-wind': expected procedure, got")
```

`error-object-code` is `#f` where every sibling gives `KP3002`, and the offending value is missing from the message. `dynamic-wind` is bootstrapped Scheme (`vm_bootstrap.dynamic_wind_src`) and raises with `(error "type error in 'dynamic-wind': expected procedure, got" before)`, so the value is an *irritant*, not part of the message, and a user `error` object is `.uncategorized` by construction. A program doing `(eq? (error-object-code e) 'KP3002)` — the dispatch KEP-0005 §4 exists to enable — misses `dynamic-wind` alone among the control primitives.

### 3. A wrong-arity call to a tail-dispatched name is `KP2001 invalid syntax`

```scheme
(define (f) (call/cc))            ;; compile error[KP2001]: invalid syntax
(define (g) (list (call/cc)))     ;; runtime: 'call/cc': expected 1 arguments, got 0  (KP3003)
```

`compileCallCCTail` / `compileCallWithValuesTail` / `compileApplyTail` return `CompileError.InvalidSyntax` for an argument-count mismatch, so `(call/cc)`, `(call/cc a b)`, `(call-with-values p)`, `(call-with-values a b c)` and `(apply)` in tail position are reported as *syntax* errors — not catchable by a `guard`, and the whole top-level form is abandoned. Every other builtin, including these same names one position away, gives a runtime arity error. (Checking arity at compile time is an improvement; reporting it as "invalid syntax" is not.)

**Verified:** fresh ReleaseSafe build at `d0d03deb`, isolated `KAAPPI_HOME`, cache cleared.

**Source location:** `src/vm_dispatch.zig:1299` (`tail_call_cc`'s `else => return VMError.NotAProcedure`); `src/vm_bootstrap.zig` `dynamic_wind_src`; `src/compiler_passthrough.zig` `compileCallCCTail` / `compileCallWithValuesTail` / `compileApplyTail`.

**Disabled test:** `tests/scheme/audit/primitives_control-audit.scm` (4 assertions marked `;; FAIL: #this`). The non-tail `call/cc` control and the 10 other D2 code assertions are enabled beside them.

### LINKED ISSUE #2037: [D1] %unwind-to-escape is missing from internal_helpers: user code can pop the wind stack, and the underflow reports as "type error in '%pop-wind'"
**Category:** VM (audit dimension D1 — internal-primitive direct-call surface)
**Spec reference:** `src/vm_bootstrap.zig`, `internal_helpers`: "The `.internal` specs that must not merely be unexported but unreachable: removed from vm.globals by `install()` above. ... `%push-wind` and the `%promise-*` mutators corrupt VM state if called out of sequence, so they must be unreachable, not merely unexported."
**Severity:** edge-case

`%unwind-to-escape` (`primitives_control.zig`, tag `.internal`) pops dynamic-wind records off `vm.wind_stack` and runs their after-thunks. It exists for one caller — `compiler_advanced.compileGuard`'s desugaring (#1988). It is **not** in `vm_bootstrap.internal_helpers`, so unlike its two siblings `%push-wind` and `%pop-wind` it stays in `vm.globals` and is callable from a plain script with no import at all.

Calling it desynchronises the wind stack from the bootstrapped `dynamic-wind` that pushed the records, and the resulting underflow surfaces as a **type error naming a primitive the user never wrote**.

**Minimal reproduction** (no imports beyond `(scheme base)`):

```scheme
(define log '())
(define (add s) (set! log (cons s log)))
(write (guard (e (#t (list 'ERR (error-object-message e))))
         (call/ec
           (lambda (k)
             (dynamic-wind
               (lambda () (add 'B1))
               (lambda ()
                 (dynamic-wind (lambda () (add 'B2))
                               (lambda () (%unwind-to-escape k) 'inner)
                               (lambda () (add 'A2))))
               (lambda () (add 'A1)))))))
(newline)
(write (reverse log)) (newline)
```

```text
(ERR "type error in '%pop-wind'")
(B1 B2 A2 A1)
```

Both after-thunks ran early (that is `%unwind-to-escape` doing its documented job), and then the inner `dynamic-wind` reached its own `(%pop-wind)` on an empty stack. `popWindFn` guards the underflow with a bare `return PrimitiveError.TypeError; // bare-ok: underflow`, and `mapNativeError` dresses that up as `type error in '%pop-wind'` — a type error about an internal name, for what is a stack underflow.

**Discriminating control:** the two siblings *are* purged, and report cleanly as undefined variables —

```scheme
(guard (e (#t (error-object-message e))) (eval '(%push-wind) (interaction-environment)))
;; => "undefined variable '%push-wind'"   (KP3001)
(guard (e (#t (procedure? %unwind-to-escape))) 'no)
;; => #t                                   -- reachable
```

so the omission is specific to this one name, not a property of the `.internal` tag.

**Fix shape:** add `"%unwind-to-escape"` to `vm_bootstrap.internal_helpers`. It is referenced from compiler-generated code through `Compiler.trueBuiltinRefOrSymbol`, which resolves against `LibraryRegistry.internal_bindings` (the pristine snapshot taken before the purge), so `guard` keeps working — the same arrangement `%push-wind` already has for `dynamic-wind`. Worth checking the underflow tag in `popWindFn` at the same time: `PrimitiveError.InvalidBytecode` (KP9001) matches what `docs/dev/gc-safety-and-error-handling.md` prescribes for an internal-invariant failure, where `TypeError` blames an argument.

**Verified:** fresh ReleaseSafe build at `d0d03deb`, isolated `KAAPPI_HOME`, cache cleared. Deterministic; unchanged under `-Dgc-stress=true`.

**Source location:** `src/vm_bootstrap.zig:51` (`internal_helpers`); `src/primitives_control.zig:39` (the spec) and `:536` (`popWindFn`'s underflow guard).

**Disabled test:** `tests/scheme/audit/primitives_control-audit.scm` (`;; FAIL: #this` — the purge assertion). The two sibling-purge assertions and a test pinning the current after-thunk behaviour are enabled beside it.

baijum:
## What

Two audit findings in the control-flow error paths, one PR:

### #2036 — three divergences

1. **`call/cc` in tail position** reported a non-procedure receiver as `KP3005` with the literal message `"error"` (`tail_call_cc`'s bare `return VMError.NotAProcedure`). The else arm now mirrors the non-tail path's `primitives.typeError`: `KP3002`, `"type error in 'call/cc': expected procedure, got 42"` — identical in both positions, like `call/ec` always was.
2. **`dynamic-wind`'s type errors** were bootstrapped Scheme `(error ...)`, so `error-object-code` answered `#f` and the offending value was an irritant, invisible in the message. The three checks now go through a new internal native `%check-procedure` (`primitives.typeError`), giving `KP3002` and `"type error in 'dynamic-wind': expected procedure, got 1"`. The native raise propagates through the same `nativeErrorToErrorObject` boundary every other primitive uses, so the code is stamped. Validation-before-`(before)` semantics (no leaked side effects, #1375) are unchanged.
3. **A wrong-arity call to a tail-dispatched name** — `compileCallCCTail` / `compileCallWithValuesTail` / `compileApplyTail` returned `CompileError.InvalidSyntax` for a *proper* operand list of the wrong length, so `(call/cc)`, `(call/cc a b)`, `(call-with-values p)`, `(apply)` in tail position were uncatchable KP2001 compile errors that abandoned the whole top-level form. Such lists now route through the ordinary call path (`compileCall`), so the runtime native arity check reports KP3003 (`"'call/cc': expected 1 arguments, got 0"`) exactly as the same form one position away always did. Improper lists (`(call/cc . 5)`) remain genuine KP2001 syntax errors. The native tier already lowered these names as ordinary calls, so this *removes* an interpreter/native divergence rather than adding one; the stale mirror comment in `llvm_emit_forms.zig` (`emitApplyForm`'s `<2 operands` case) is updated — its eval-fallback abandon produces the identical runtime diagnostic.

### #2037 — `%unwind-to-escape` reachability

Added to `vm_bootstrap.internal_helpers` and purged, so user code can no longer pop the wind stack out of sequence (the underflow used to surface as `type error in '%pop-wind'`). `popWindFn`'s underflow guard now reports KP9001 with `"wind stack underflow in '%pop-wind'"` — the internal-invariant tag `docs/dev/gc-safety-and-error-handling.md` prescribes — instead of a TypeError blaming an argument.

**Deviation from the issue's fix shape, found the hard way:** the issue claims compiler-generated `%unwind-to-escape` references resolve against `internal_bindings` as "the pristine snapshot taken before the purge". That ordering does not exist: `registerStandardLibraries` (which snapshots globals into `internal_bindings`) runs **after** `vm_bootstrap.install` purges, on every init path (`main.zig` ×2, `testing_helpers`, benches, LSP). Purging alone breaks every `guard` with `undefined variable '__kaappi_base__%unwind-to-escape'` — verified by temporarily removing the seed. `install()` therefore seeds `internal_bindings` itself, immediately before the remove; this is order-independent and also covers init paths that never register libraries.

## Tests

- `tests/scheme/audit/primitives_control-audit.scm`: the 4 disabled `FAIL: #2036` assertions enabled (dynamic-wind code+message ×2 incl. a new `after` variant, call/cc tail code+message), the disabled `#2037` purge assertion enabled, plus new `%check-procedure` purge and guard-driven after-thunk-ordering assertions, and 6 new tail-arity assertions (KP3003 code + exact messages for `(call/cc)`, `(call/cc 1 2)`, `(call-with-values 42)`, `(apply)` via `eval`, so a residual compile error surfaces as a catchable object instead of aborting the file). The three old direct-call `%unwind-to-escape` tests are gone with the reachability they pinned; the fourth (after-thunk depth) was passing vacuously after the purge and is rewritten to exercise `%unwind-to-escape` through its one legitimate caller (`guard`), where it distinguishes #1988's clause ordering from plain unwinding. Every enabled assertion's pre-fix value was captured on the baseline binary (KP3005 / `#f` / KP2001 / reachable) — each fails without its fix.
- `src/tests_libraries.zig`: new unit test locking the purge+seed pair (purged from globals, present in `internal_bindings`, `guard` works end-to-end with correct wind ordering). Removing the seed fails it.
- Evidence: audit file 200/200 pass; `zig build test` green (1823 passed, 7 skipped); R7RS suite 1395/0; `error-format.sh` 144/0 (its dynamic-wind message assertion still matches — the value moved from irritant into the message, same printed text); `error-object-code.sh` 12/0; all continuation suites; `guard-clause-dynamic-env-1988.scm` 23/23; `exit-wind.sh` 8/0; native tier exercised via `kaappi compile` (apply/guard/call/cc/dynamic-wind program) plus `native-apply-lowering-1803.sh` and `native-bootstrap-callbacks-1376.sh`.

## BASELINE note

Unchanged (28). The one bare `PrimitiveError.TypeError` this PR touches (`popWindFn`'s underflow) was already excluded via `bare-ok` and became a non-gated `InvalidBytecode`; the new `VMError.TypeError` in `tail_call_cc` carries `// bare-ok: detail set above`, matching the adjacent `tail_eval` precedent.

Closes #2036
Closes #2037

### DIFF
--- a/src/compiler_passthrough.zig
+++ b/src/compiler_passthrough.zig
@@ -266,7 +266,18 @@ fn compileSelfTailCall(self: *Compiler, expr: Value, dst: u16, nargs: u8) Compil
 
 pub fn compileApplyTail(self: *Compiler, expr: Value, dst: u16) CompileError!void {
     var arg_list = types.cdr(expr);
-    if (arg_list == types.NIL) return CompileError.InvalidSyntax;
+    // A *proper* operand list of the wrong length (< 2) is an arity question,
+    // not a syntax question: route the form through the ordinary call path so
+    // the native apply's runtime arity check reports KP3003 exactly as the
+    // same form one position away does (#2036). Only an improper list is
+    // malformed syntax.
+    {
+        var count: usize = 0;
+        var cur = arg_list;
+        while (types.isPair(cur)) : (cur = types.cdr(cur)) count += 1;
+        if (cur != types.NIL) return CompileError.InvalidSyntax;
+        if (count < 2) return compileCall(self, expr, dst, true);
+    }
 
     const needs_rebase = (dst + 1 != self.next_register);
     const base = if (needs_rebase) try self.allocReg() else dst;
@@ -276,14 +287,12 @@ pub fn compileApplyTail(self: *Compiler, expr: Value, dst: u16) CompileError!voi
 
     var nargs_count: usize = 0;
     while (arg_list != types.NIL) {
-        if (!types.isPair(arg_list)) return CompileError.InvalidSyntax;
         const arg_reg = try self.allocReg();
         try self.compileExprViaIR(types.car(arg_list), arg_reg, false);
         nargs_count += 1;
         arg_list = types.cdr(arg_list);
     }
 
-    if (nargs_count < 1) return CompileError.InvalidSyntax;
     if (nargs_count > 255) return CompileError.InternalLimit;
     const nargs: u8 = @intCast(nargs_count);
 
@@ -311,13 +320,21 @@ pub fn compileCallWithValuesTail(self: *Compiler, expr: Value, dst: u16) Compile
     // top-level redefinition of either name can affect this call (#1715;
     // the previous plain get_global/call_global-by-name approach only
     // ever protected against the former).
+    // A *proper* argument list of the wrong length is an arity question, not a
+    // syntax question: route the form through the ordinary call path so the
+    // runtime arity check reports KP3003 exactly as the same form one position
+    // away does (#2036). Only an improper list is malformed syntax.
+    {
+        var count: usize = 0;
+        var cur = types.cdr(expr);
+        while (types.isPair(cur)) : (cur = types.cdr(cur)) count += 1;
+        if (cur != types.NIL) return CompileError.InvalidSyntax;
+        if (count != 2) return compileCall(self, expr, dst, true);
+    }
     const args = types.cdr(expr);
-    if (args == types.NIL or !types.isPair(args)) return CompileError.InvalidSyntax;
     const producer = types.car(args);
     const rest = types.cdr(args);
-    if (rest == types.NIL or !types.isPair(rest)) return CompileError.InvalidSyntax;
     const consumer = types.car(rest);
-    if (types.cdr(rest) != types.NIL) return CompileError.InvalidSyntax;
 
     const needs_rebase = (dst + 1 != self.next_register);
     const base = if (needs_rebase) try self.allocReg() else dst;
@@ -348,10 +365,18 @@ pub fn compileCallWithValuesTail(self: *Compiler, expr: Value, dst: u16) Compile
 }
 
 pub fn compileCallCCTail(self: *Compiler, expr: Value, dst: u16) CompileError!void {
-    const args = types.cdr(expr);
-    if (args == types.NIL or !types.isPair(args)) return CompileError.InvalidSyntax;
-    const receiver = types.car(args);
-    if (types.cdr(args) != types.NIL) return CompileError.InvalidSyntax;
+    // A *proper* argument list of the wrong length is an arity question, not a
+    // syntax question: route the form through the ordinary call path so the
+    // runtime arity check reports KP3003 exactly as the same form one position
+    // away does (#2036). Only an improper list is malformed syntax.
+    {
+        var count: usize = 0;
+        var cur = types.cdr(expr);
+        while (types.isPair(cur)) : (cur = types.cdr(cur)) count += 1;
+        if (cur != types.NIL) return CompileError.InvalidSyntax;
+        if (count != 1) return compileCall(self, expr, dst, true);
+    }
+    const receiver = types.car(types.cdr(expr));
 
     const needs_rebase = (dst + 1 != self.next_register);
     const base = if (needs_rebase) try self.allocReg() else dst;
--- a/src/llvm_emit_forms.zig
+++ b/src/llvm_emit_forms.zig
@@ -1016,13 +1016,14 @@ pub fn emitPassthrough(self: *LLVMEmitter, expr: Value, is_tail: bool) EmitError
 //     define/set! of `apply` in this module marks it rebound and routes the
 //     form through an ordinary indirect call instead.
 //   - tail + unshadowed + unrebound + <2 operands → the interpreter's
-//     compileApplyTail raises InvalidSyntax at compile time; abandoning
-//     native compilation routes the enclosing form to that exact error. A
-//     REBOUND `apply` with too few operands is not this case: the
-//     interpreter routes it through the ordinary call path (its #2033
-//     gate), so the generic indirect call below is the matching behavior,
-//     and the user's own procedure (or the built-in arity check) raises at
-//     run time.
+//     compileApplyTail routes the form through the ordinary call path
+//     (#2036), so the built-in arity check raises at run time. Abandoning
+//     native compilation here stays the safe mirror: the eval fallback
+//     recompiles the enclosing form through that same ordinary path and
+//     produces the identical runtime error, so only the executing tier
+//     differs, never the diagnostic. A REBOUND `apply` with too few
+//     operands is the same story through the #2033 gate and the generic
+//     indirect call below.
 //   - everything else — lexically shadowed `apply` (any shape), a rebound
 //     `apply` (tail or non-tail), or a non-tail form with <2 operands — is
 //     an ordinary indirect call through whatever `apply` resolves to in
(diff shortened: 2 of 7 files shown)


baijum:
**Review (Claude Code)** — ✅ approve-with-nits

#2036 (three diverging control-flow error paths) and #2037 (`%unwind-to-escape` reachable) both fully addressed. Tail `call/cc` non-procedure now mirrors the non-tail path (→KP3002 with the value named); `dynamic-wind` type errors route through native `%check-procedure` (→KP3002, coded); tail-dispatched wrong-arity on a *proper* operand list routes to `compileCall` (→runtime KP3003) while an *improper* list stays `InvalidSyntax`. `%unwind-to-escape` is purged from globals and seeded into `internal_bindings` — and the ordering argument (purge runs before `snapshotInternalBindings`, so the manual seed is genuinely required and stays rooted) is correct. Reference sweep for both `%`-names is clean.

Nits (non-blocking):
- `vm_dispatch.zig` tail path formats the receiver with `printer.valueToString` (heap alloc, `.write`) whereas the sibling `primitives.typeError` uses `safeValueDescription` (128-byte buffer) — identical for ordinary values, could diverge for exotic/cyclic receivers. Calling `primitives.typeError("call/cc", "procedure", receiver)` directly would be byte-identical and drop the alloc.
- No `tests/scheme/compile/*.sh` for native-tier `apply` arity — low value here since the native path emits nothing for the erroring form (falls back to eval → KP3003 automatically).

_Not built locally; exact runtime message strings are asserted by the new audit tests but unrun on my side._

PR #2344, opened 2026-08-25
Finish the SRFI 166 rework started in #2292: lazy trimmed/lazy, immutable state variables

### LINKED ISSUE #2056: [SRFI-166] with restores the whole state vector, so output written inside it does not advance col or row
**Category:** SRFI
**Spec reference:** SRFI 166, `(with ((state-var value) ...) fmt ...)` — "Applies each of the formatters fmt with each state-var bound to the corresponding value. The resulting state is then updated to restore **each state-var** to its original value."
**Severity:** wrong-result

`with` snapshots the **entire** 13-slot state vector on entry and copies every slot back on exit (`lib/srfi/166.sld:398-414`), rather than restoring only the state variables it bound. `col` and `row` are not settings — they track where output has actually reached — so any formatter run inside a `with` becomes invisible to everything that follows it. `space-to`, `tab-to` and `fl` all then compute against a stale column.

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 166))
(show #f (with (list (list pad-char #\.)) "abcd") (space-to 6) "x")
;; expected: "abcd  x"     actual: "abcd      x"   (col restored to 0, so space-to pads 6)

(show #f (with (list (list pad-char #\.)) "a" nl "b") fl "Z")
;; expected: "a\nb\nZ"     actual: "a\nbZ"         (col restored to 0, so fl emits nothing)
```

**Discriminating control:** remove the `with` wrapper and the identical output is tracked correctly — `(show #f "abcd" (space-to 6) "x")` ⟹ `"abcd  x"` and `(show #f "a" nl "b" fl "Z")` ⟹ `"a\nb\nZ"`. Both controls are enabled assertions in the suite, so the defect is isolated to the restore loop and not to `space-to`/`fl`.

`port` is restored the same way. That one is arguably harmless today, but the same loop is what would silently undo a `with`-scoped port redirection's side effects on position.

**Verified:** fresh ReleaseSafe build at `08828884` (macOS aarch64), isolated `KAAPPI_HOME`, cache cleared.
**Source location:** `lib/srfi/166.sld:398-414` — `(let restore ((i 0)) (when (< i %state-size) (vector-set! st i (vector-ref saved i)) ...))` copies back all 13 slots including `%idx-col` and `%idx-row`.
**Disabled test:** `tests/scheme/srfi/srfi166-audit.scm`, 2 assertions marked `;; FAIL:` ("with: output position survives the form (col not restored)", "with: fl after a with sees the real column"), each paired with an enabled control.

**Found by:** Systematic audit v2, Phase 3.5

### LINKED ISSUE #2058: [SRFI-166] tab-to emits a full tab width when already on a tab stop, and divides by zero on tab-width 0
**Category:** SRFI
**Spec reference:** SRFI 166, `(tab-to [tab-width])` — "Outputs spaces up to the next tab stop, using tab stops of width tab-width, which defaults to 8. **If already on a tab stop, does nothing.** If you want to ensure you always tab at least one space, you can use `(each " " (tab-to width))`. Columns are zero-based." The spec's first worked example is `(show #f (tab-to 5) "b")` ⟹ `"b"`.
**Severity:** wrong-result

`tab-to` computes the next stop as `(* (+ (quotient c tw) 1) tw)` unconditionally (`lib/srfi/166.sld:226-232`), so when the column is already a multiple of the tab width it advances a whole extra stop instead of doing nothing. Column 0 is a tab stop, so `(tab-to n)` as the very first formatter always emits `n` spaces — the exact case the spec documents as emitting none, and the exact case the spec's "if you want at least one space, write `(each " " (tab-to width))`" advice exists to work around.

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 166))
(show #f (tab-to 5) "b")           ;; expected: "b"        actual: "     b"
(show #f "abcde" (tab-to 5) "|")   ;; expected: "abcde|"   actual: "abcde     |"
```

**Discriminating control:** the spec's two *other* `tab-to` examples both pass — `(show #f "a" (tab-to 5) "b")` ⟹ `"a    b"` and `(show #f "abcdefghi" (tab-to 5) "b")` ⟹ `"abcdefghi b"` — so the tab arithmetic is right everywhere except exactly on a stop. All three are assertions in the suite; only the on-a-stop pair is disabled.

**Also, from the same expression:** `(tab-to 0)` raises `KP3004: division by zero` from the `quotient`. A zero tab width has no next stop, so "does nothing" is the natural answer; a hard error out of a formatter that is otherwise total is at minimum surprising. Third disabled assertion.

**Verified:** fresh ReleaseSafe build at `08828884` (macOS aarch64), isolated `KAAPPI_HOME`, cache cleared.
**Source location:** `lib/srfi/166.sld:226-232`
**Disabled test:** `tests/scheme/srfi/srfi166-audit.scm`, 3 assertions marked `;; FAIL:`

**Found by:** Systematic audit v2, Phase 3.5

### LINKED ISSUE #2059: [SRFI-166] escaped wraps its output in delimiters and ignores esc-ch/renamer; maybe-escaped never quotes on an embedded quote
**Category:** SRFI
**Spec reference:** SRFI 166, `(escaped str [quote-ch esc-ch renamer])` — "Outputs the string str, escaping any quote or escape characters. If esc-ch, which defaults to `#\\`, is `#f`, escapes only the quote-ch, which defaults to `#\"`, by doubling it, as in SQL strings and CSV values. If renamer is provided, it should be a procedure of one character which maps that character to its escape value." Worked example: `(show #f (escaped "hi, bob!"))` ⟹ `"hi, bob!"`.
`(maybe-escaped str pred [quote-ch esc-ch renamer])` — "Like escaped, but first checks if any quoting is required (**by the existence of either any quote or escape characters, or any character matching pred**), and if so outputs the string in quotes and with escapes. Otherwise outputs the string as is."
**Severity:** wrong-result

Three separate deviations in the same 17-line region (`lib/srfi/166.sld:151-167`):

1. **`escaped` adds surrounding delimiters the spec does not.** The spec's own first example is a string containing nothing to escape, and its expected output is the string unchanged — no quote characters. This implementation emits `qch` before and after unconditionally. `escaped` escapes; `maybe-escaped` is the one that quotes.
2. **`escaped` ignores its `esc-ch` and `renamer` arguments.** It takes `(escaped str . rest)`, reads only `(car rest)` as the quote char, and always escapes with a literal `#\\`. The documented `esc-ch` = `#f` mode (double the quote, "as in SQL strings and CSV values") is unreachable, and `renamer` is unreachable.
3. **`maybe-escaped` consults only `pred`.** The spec requires quoting when the string contains a quote or escape character *even if* `pred` matches nothing — which is exactly the CSV/symbol-quoting case the procedure exists for.

**Minimal reproduction:**
```scheme
(import (scheme base) (scheme char) (srfi 166))
(show #f (escaped "hi, bob!"))                              ;; expected: "hi, bob!"  actual: "\"hi, bob!\""
(show #f (escaped "a\"b" #\" #f))                           ;; expected: "a\"\"b"    actual: "\"a\\\"b\""
(show #f (maybe-escaped "foo\"bar\"baz" char-whitespace? #\"))
                                          ;; expected: a quoted, escaped string   actual: "foo\"bar\"baz" unchanged
```

**Discriminating control:** `maybe-escaped`'s other two documented examples both pass — `(maybe-escaped "foo" char-whitespace? #\")` ⟹ `"foo"` and `(maybe-escaped "foo bar" char-whitespace? #\")` ⟹ `"\"foo bar\""` — so the quoting path itself is right; only the *decision* to take it is too narrow. Both are enabled assertions in the suite. That pair also isolates deviation 1: `maybe-escaped` correctly adds delimiters only in the quoting case, while `escaped` adds them always.

**Verified:** fresh ReleaseSafe build at `08828884` (macOS aarch64), isolated `KAAPPI_HOME`, cache cleared.
**Source location:** `lib/srfi/166.sld:151-160` (`escaped`), `lib/srfi/166.sld:162-167` (`maybe-escaped`)
**Disabled test:** `tests/scheme/srfi/srfi166-audit.scm`, 4 assertions marked `;; FAIL:`

**Found by:** Systematic audit v2, Phase 3.5

### LINKED ISSUE #2061: [SRFI-166] the numeric family ignores sign-rule, comma-rule, comma-sep and decimal-sep; numeric/comma inserts no commas and numeric/si ignores its base
**Category:** SRFI
**Spec reference:** SRFI 166 §"Formatting Numbers". `(numeric num [radix precision sign-rule comma-rule comma-sep decimal-sep])` — "You can optionally specify any radix from 2 to 36 (even if num isn't an integer)"; "A sign-rule of `#t` indicates to output a plus sign (+) for positive integers. However, if sign-rule is a pair of two strings, it means to wrap negative numbers with the two strings … `-1.99 => (1.99)`"; "comma-rule is an integer specifying the number of digits between commas, or a list of integers…"; "decimal-sep is the character to use for decimals"; "These parameters may seem unwieldy, but they can also take their defaults from state variables."
`(numeric/si num [base separator])` — "The base defaults to 1000 … If the base is 1024, the suffixes are Ki, Mi, Gi, etc. … If separator is provided, it is inserted after the number, before any suffix."
**Severity:** wrong-result (every documented example of two of the three procedures produces the wrong string)

Of `numeric`'s seven documented parameters, **only `num`, `radix` and `precision` are read at all** (`lib/srfi/166.sld:171-180`), and `radix` is silently dropped as soon as `precision` is supplied, because the precision branch routes through `%format-float`, which is decimal-only. Of the six state variables that are supposed to supply defaults, only `radix` is consulted.

`numeric/comma` (`lib/srfi/166.sld:200-201`) is a stub: `(lambda (st) (%output-string st (number->string num)))`. It ignores its `comma-rule`, `radix`, `precision` and `sign-rule` arguments and the `comma-rule`/`comma-sep` state variables, and inserts no separators at all. All three of the spec's examples are wrong.

`numeric/si` (`lib/srfi/166.sld:203-212`) ignores its `base` and `separator` arguments, has no sub-unit (µ, m, n) prefixes, and emits a spurious `.0` on values that abbreviate exactly.

| expression | spec | actual |
|---|---|---|
| `(show #f (numeric 255 16 2))` | `"ff.00"` | `"255.00"` |
| `(show #f (numeric 5 10 #f #t))` | `"+5"` | `"5"` |
| `(show #f (numeric -1.99 10 2 '("(" . ")")))` | `"(1.99)"` | `"-1.99"` |
| `(show #f (numeric 1234567 10 #f #f 3))` | `"1,234,567"` | `"1234567"` |
| `(show #f (with ((decimal-sep ",")) (numeric 1.5 10 1)))` | `"1,5"` | `"1.5"` |
| `(show #f (with ((precision 0)) 1.5))` | `"2"` | `"1.5"` |
| `(show #f (with ((precision 3)) (written 1/7)))` | `"0.143"` | `"1/7"` |
| `(show #f (numeric/comma 123456789))` | `"123,456,789"` | `"123456789"` |
| `(show #f (numeric/comma 123456789 2))` | `"1,23,45,67,89"` | `"123456789"` |
| `(show #f (numeric/comma 123456789 '(3 2)))` | `"12,34,56,789"` | `"123456789"` |
| `(show #f (numeric/si 3986))` | `"4k"` | `"4.0k"` |
| `(show #f (numeric/si 3986 1024) "B")` | `"3.9KiB"` | `"4.0kB"` |
| `(show #f (numeric/si 608 1000 " ") "B")` | `"608 B"` | `"608B"` |
| `(show #f (numeric/si 1.23e-6) "m")` | `"1.2µm"` | `"0.0m"` |

(The `with` rows are written in the spec's macro syntax; reproduced here through this implementation's procedural `with` — see #2054 — they give the same wrong answers.)

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 166))
(show #f (numeric/comma 123456789))   ;; expected: "123,456,789"   actual: "123456789"
(show #f (numeric 255 16 2))          ;; expected: "ff.00"         actual: "255.00"
```

**Discriminating control:** the parameters that *are* wired up work — `(numeric 255 16)` ⟹ `"ff"`, `(numeric 3.14159 10 2)` ⟹ `"3.14"`, `(with (list (list radix 16)) (numeric 255))` ⟹ `"ff"`, `(numeric/si 608)` ⟹ `"608"`, `(numeric (expt 2 100))` ⟹ the full bignum. So this is not "numeric is broken"; it is precisely the arguments after `precision`, plus the radix/precision interaction and every state-variable default except `radix`. All the working cases are enabled assertions in the suite.

**Verified:** fresh ReleaseSafe build at `08828884` (macOS aarch64), isolated `KAAPPI_HOME`, cache cleared.
**Source location:** `lib/srfi/166.sld:171-212` (`numeric`, `%format-float`, `numeric/comma`, `numeric/si`)
**Disabled test:** `tests/scheme/srfi/srfi166-audit.scm`, 14 assertions marked `;; FAIL:`

**Found by:** Systematic audit v2, Phase 3.5

### LINKED ISSUE #2062: [SRFI-166] padded/trimmed/fitted measure with string-length: the ellipsis and string-width state variables are never read
**Category:** SRFI
**Spec reference:** SRFI 166 §"Padding and Trimming" — "If a truncation ellipsis is set, then when any truncation occurs, `trimmed` and `trimmed/right` will prepend and append the ellipsis, respectively. `trimmed/both` will both prepend and append. The length of the ellipsis will be considered when truncating the original string, so that the total width will never be longer than width."
§"State Variables", `string-width` — "Procedure `(string [start end])` returning the column width the string occupies (default: `string-length`)"; `wrapped` is likewise specified to break "ensuring no line exceeds width (**as measured with the string-width state variable**)".
**Severity:** wrong-result

`padded`, `padded/right`, `padded/both`, `trimmed`, `trimmed/right`, `trimmed/both`, `fitted`, `fitted/right` and `fitted/both` all call `string-length` on the captured output directly (`lib/srfi/166.sld:297-381`). None reads `%idx-ellipsis` and none reads `%idx-strwidth`. The `ellipsis` state variable is therefore write-only, and `string-width` is initialised to `string-length` (`lib/srfi/166.sld:50`) and then never consulted by anything.

This is also the mechanism that makes `(srfi 166 unicode)`'s `terminal-aware` unimplementable — see the linked unicode issue — since `terminal-aware` is specified as nothing more than a `with` of `string-width`/`substring/width`/`substring/preserve`.

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 166))
(show #f (with (list (list ellipsis "...")) (trimmed 5 "abcdef")))
;; expected: "...ef"   actual: "bcdef"
(show #f (with (list (list ellipsis "...")) (trimmed/right 5 "abcdef")))
;; expected: "ab..."   actual: "abcde"
(show #f (with (list (list ellipsis "_")) (trimmed/both 5 "abcdef")))
;; expected: "_bcd_"   actual: "abcde"
(show #f (with (list (list string-width (lambda (s) (* 2 (string-length s)))))
               (padded/right 4 "ab")) "|")
;; expected: "ab|"  (a doubling width function makes "ab" measure 4)   actual: "ab  |"
```

**Discriminating control:** the ellipsis-free truncation arithmetic is entirely correct — all six of the spec's non-ellipsis trimming examples pass, including the odd/even asymmetry `(trimmed/both 5 "abcdef")` ⟹ `"abcde"` versus `(trimmed/both 4 "abcdef")` ⟹ `"bcde"`, and `(show #f (with (list (list ellipsis "...")) (trimmed 5 "abc")))` ⟹ `"abc"` correctly leaves an unused ellipsis alone. For `string-width`, the enabled control `(show #f (padded/right 4 "ab") "|")` ⟹ `"ab  |"` gives byte-identical output with and without the binding, which is what shows the binding is not read rather than read-and-ignored.

**Verified:** fresh ReleaseSafe build at `08828884` (macOS aarch64), isolated `KAAPPI_HOME`, cache cleared.
**Source location:** `lib/srfi/166.sld:297-381`
**Disabled test:** `tests/scheme/srfi/srfi166-audit.scm`, 4 assertions marked `;; FAIL:`

**Found by:** Systematic audit v2, Phase 3.5

### LINKED ISSUE #2063: [SRFI-166] displayed renders a formatter argument as #<procedure> instead of returning it as is
**Category:** SRFI
**Spec reference:** SRFI 166, `(displayed obj)` — "**If obj is a formatter, returns obj as is.** Otherwise, outputs obj using display semantics. … It is an error if obj is a procedure which is not a formatter."
**Severity:** wrong-result

`displayed` (`lib/srfi/166.sld:140-142`) unconditionally routes its argument through `display` to a string port. A formatter is a procedure, so `display` renders it as `#<procedure>` and that text is emitted instead of the formatter's own output.

This matters because `displayed` is the natural mapper to hand to `joined` and friends, and the spec's own `columnar` examples pass `(displayed x)` where `x` may itself be a formatter. Any code that composes `displayed` over a value that might be a formatter silently prints `#<procedure>` — a wrong result with no error.

**Minimal reproduction:**
```scheme
(import (scheme base) (srfi 166))
(show #f (displayed (displayed "inner")))   ;; expected: "inner"   actual: "#<procedure>"
(show #f (joined displayed (list "a" (each "b" "c")) "-"))
                                            ;; expected: "a-bc"    actual: "a-#<procedure>"
```

**Discriminating control:** `show`'s own top-level dispatch does make this distinction — `%run-fmt` (`lib/srfi/166.sld:80-88`) has a `(procedure? fmt)` arm — so `(show #f (each "b" "c"))` ⟹ `"bc"` correctly. The defect is that `displayed` does not reuse it. Both `(show #f (displayed "hello"))` ⟹ `"hello"` and `(show #f (displayed #\x))` ⟹ `"x"` are enabled assertions and pass, so the non-formatter path is right.

**Verified:** fresh ReleaseSafe build at `08828884` (macOS aarch64), isolated `KAAPPI_HOME`, cache cleared.
**Source location:** `lib/srfi/166.sld:140-142`
**Disabled test:** `tests/scheme/srfi/srfi166-audit.scm`, 1 assertion marked `;; FAIL:`

**Found by:** Systematic audit v2, Phase 3.5

### LINKED ISSUE #2064: [SRFI-166] pretty is write: it never breaks a line at any width, and written-shared/pretty-shared do not label shared structure

### LINKED ISSUE #2065: [SRFI-166] (srfi 166 columnar): columnar and tabular do not align, wrapped hardcodes width 78, wrapped/char and justified are aliases, line-numbers is not a stream

### LINKED ISSUE #2066: [SRFI-166] (srfi 166 unicode) is a stub: substring-terminal-width returns an integer instead of a substring, string-terminal-width is string-length, terminal-aware is a no-op

### LINKED ISSUE #2067: [SRFI-166] 15 documented names are not exported (joined/dot, numeric/fitted, trimmed/lazy, make-state-variable, 7 state variables, 4 sub-library names) and (srfi 166 base) does not exist

baijum:
## What remained broken after #2292, and what this fixes

#2292 restructured SRFI 166 into `lib/srfi/166/{base,pretty,columnar,unicode,color}.sld` and fixed the bulk of the v2 audit findings. Re-verifying all ten wave-2 issues against that tree (`origin/main` @ `645d2bba`, fresh ReleaseSafe build, isolated `KAAPPI_HOME`, cache cleared) shows **the audit already ran 259/259 green there** — the reported "~105 failures" reflects a pre-#2292 tree, not current main. Direct probes of every claim table in the ten issues found **two specified behaviors still missing**, both fixed here, plus under-covered spec examples now pinned.

### Fixed in this PR

1. **`trimmed/lazy` was a non-lazy alias of `trimmed/right`** (`lib/srfi/166/base.sld` had a comment admitting it). The spec's defining property — "safe to use with an infinite amount of output, e.g. from written-simply on an infinite list" — did not hold: `(show #f (trimmed/lazy 7 (written-simply circular-list)))` hit a hard, uncatchable **KP3008 stack overflow** inside the output capture. Now:
   - `%write-flat` was rewritten as a streaming `%write-stream` that emits one token at a time (`(`, each element, separators, `)`), and `written`/`written-shared`/`written-simply` thread each chunk through the `output` state variable. `%write-flat` survives as an accumulating wrapper for `pretty`'s fitting decision.
   - `trimmed/lazy` installs a counting `output` hook and, once the width budget is spent, **terminates the generator itself** via a `call/cc` escape — restoring the hook binding by hand on both exit paths (a plain `with` would skip its restore on the escape). No ellipsis is output, per spec.
   - Stack safety came with it: `extract-shared-objects` is now an explicit enter/exit worklist and the writer's spines are tail-recursive, so a 50,000-element list no longer overflows either (it did, pre-fix). The exit-event timing preserves the cycle-vs-sharing distinction exactly — `written` still labels only cycles, `written-shared` labels sharing (both now pinned by tests).
2. **`make-state-variable`'s `immutable` flag was stored but never enforced.** The spec: an immutable state variable "can only be dynamically bound with `with`, and not set with `with!`" — `with!` on one is now an error (and `with` still binds it, also pinned).

### Verify-and-close (already fixed by #2292 on main; evidence cited)

Every claim in these issues was re-verified against the new `lib/srfi/166/*` files before touching anything — via the 259 already-enabled audit assertions plus direct probes of each issue's full claim table:

- **#2066** (unicode stub): full EAW-wide/fullwidth=2, Mn/Cf=0, ANSI=0 model; `substring-terminal-width` returns a substring (all 8 spec examples); `terminal-aware` is a real `with` of `string-width`/`substring/width`/`substring/preserve`. This PR adds the `/wide` variant assertions the inventory-only checks missed.
- **#2065** (columnar): even width division, per-line borders, degenerate no-column case, `tabular` minimum widths, `wrapped` reads `width` and `word-separator?`, `wrapped/char` splits on characters, `justified` full-justifies, `line-numbers` streams through `columnar`/`from-file` (the nl(1) example). This PR adds a custom-`word-separator?` tokenization case.
- **#2064** (pretty): breaks datagrams across lines at `width`, round-trips, `written-shared`/`pretty-shared` label non-cyclic sharing. This PR adds the SRFI's own `columnar`+`pretty`+`justified` worked example (border on every line, multi-line layout).
- **#2063** (`displayed` on a formatter): returns it as is — audit case existed.
- **#2062** (ellipsis/string-width): all three ellipsis forms and the `string-width`-measured padding pass — audit cases existed.
- **#2061** (numeric family): radix+precision (`ff.00`), sign-rule `#t` and pair, comma-rule int/list, `decimal-sep`/`precision` state vars, all `numeric/comma` and `numeric/si` examples pass. This PR adds the `comma-sep` and `decimal-align` state-variable cases.
- **#2059** (escaped): no spurious delimiters, `esc-ch` `#f` doubling, `maybe-escaped` quoting on an embedded quote — all audit-pinned. This PR adds the `renamer` case.
- **#2058** (tab-to): on-a-stop no-op and zero width — audit cases existed.
- **#2056** (`with` restoring col/row): audit cases existed (with paired controls).
- **#2067**'s export inventory: all 15 names and `(srfi 166 base)` existed post-#2292; this PR completes the issue by making `trimmed/lazy` and `make-state-variable` behave as specified and pinning `numeric/fitted`'s three spec hash-examples, `joined/dot`'s spec example, the `writer` state variable, and `pretty-environment` reachability.

## Regression evidence

`tests/scheme/srfi/srfi166-audit.scm` grows 259 → **285 assertions**. On the pre-fix tree (`origin/main` @ `645d2bba` + this PR's tests, library change stashed) the audit **fails hard**: it aborts with an uncatchable KP3008 stack overflow at `trimmed/lazy: terminates an infinite written-simply stream`, after counted failures `trimmed/lazy: no ellipsis is output even when one is set` and `make-state-variable: with! on an immutable variable is an error` (the `written: a 50,000-element list` overflow abort follows the same way). With the fix: 285/285.

## Test results

- `tests/scheme/srfi/srfi166-audit.scm`: **285 pass, 0 fail**
- `bash tests/scheme/run-all.sh`: **717 scheme files pass, 0 fail** + R7RS suite **1395 pass, 0 fail**
- `zig build test`: **1744 pass, 7 skip** (unit) + 78 pass (thottam)
- `npx markdownlint-cli2` on the touched doc: clean

## Docs

`docs/dev/srfi-implementation-notes.md` gains a SRFI 166 section recording the constraints that are invisible from the code: the token-streaming writer as the foundation of `trimmed/lazy`, the exit-event timing that distinguishes cycles from sharing, emission-time label numbering, and the library search-order trap (`zig-out/lib` / `~/.kaappi/lib` shadow the source tree — rerun `zig build` after every `.sld` edit).

Closes #2067
Closes #2066
Closes #2065
Closes #2064
Closes #2063
Closes #2062
Closes #2061
Closes #2059
Closes #2058
Closes #2056

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -456,7 +456,13 @@ jobs:
           #     tail-recursion fix past the 32,768 frame cap -- ~0.1 s plain,
           #     exit 124 stressed (measured quadratic: ~3 min at 40k on a
           #     fast machine); the n = 0 guard half of #2084 stays stressed
-          #     in srfi178-audit.scm.
+          #     in srfi178-audit.scm. srfi166-audit.scm streams a
+          #     50,000-element list through the token writer to pin #2344's
+          #     iterative (non-recursing) written/pretty stack-safety -- ~0.1 s
+          #     plain, exit 124 stressed since each emitted token allocates
+          #     against the whole live list (and its parallel lane starves a
+          #     sibling into a timeout too); the full 285-assertion audit still
+          #     runs unstressed on every other leg.
           #
           # (b) REAL BUGS THIS GATE FOUND ON ITS FIRST RUN -- excluded so the
           #     gate is green for everything else, exactly as the campaign
@@ -493,8 +499,8 @@ jobs:
           # cannot rot into a silent no-op.
           KAAPPI_GC_STRESS_SKIP: >-
             trampoline-polish-1375.scm srfi14.scm srfi264.scm
-            reader-port-refill-gaps.scm srfi178-segment-stack-2084.scm
-            srfi18-deepcopy-matrix-audit.scm
+            srfi166-audit.scm reader-port-refill-gaps.scm
+            srfi178-segment-stack-2084.scm srfi18-deepcopy-matrix-audit.scm
 
   riscv64-test:
     needs: format
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -747,6 +747,63 @@ view there too, which a literal R7RS vector cannot provide in portable Scheme
 documented scope reduction, unlike `array-flatten`, which the spec itself
 mandates as always a fresh copy regardless of source mode.
 
+### SRFI 166 — Monadic Formatting
+
+SRFI 166 ships as six libraries: `(srfi 166 base)` holds the entire state
+model and formatting core, and `(srfi 166)` re-exports base plus the four
+sub-libraries (`pretty`, `columnar`, `unicode`, `color`) from
+`lib/srfi/166/`. The #2292 rewrite put every state variable, `fn`/`with`
+macro, and formatter on one architecture: a state variable is a record
+carrying name/default/immutable (so `make-state-variable` is the real spec
+extension point, and `with!` errors on an immutable one because the spec
+allows "only dynamically bound with with"), the formatting state is a hash
+table keyed by the state-variable *object*, formatters mutate the state in
+place, and `with` restores exactly the variables it bound — which is why
+`col`/`row` survive a `with` while `forked`/`call-with-output` snapshot the
+whole table with `hash-table-copy`. Three constraints from that file that
+are invisible from the code and easy to break:
+
+1. **The writer streams token by token, and that is load-bearing.**
+   `%write-stream` calls `(emit string)` for every token ("(", each element,
+   separators, ")") rather than building one string; `written` and friends
+   thread each chunk through the `output` state variable. This is what makes
+   `trimmed/lazy` implementable at all: it installs a counting `output` hook
+   and, when the width budget is spent, unwinds the *generator itself* via a
+   `call/cc` escape — the spec's only mechanism that is "safe to use with an
+   infinite amount of output, e.g. from written-simply on an infinite
+   (circular) list". Regressing `%write-stream` back to whole-string
+   accumulation reinstates a hard KP3008 stack overflow on that case (the
+   audit pins it). Two related details: the list/vector spines are written
+   tail-recursively (a 50k-element list must not overflow), and
+   `trimmed/lazy` must restore the `output` binding on its escape path by
+   hand — a plain `with` skips its restore on a non-local exit, which would
+   strand the counting hook in the state for every later formatter.
+2. **`extract-shared-objects`' exit-event timing is the cycle/sharing
+   distinction.** The walk is an explicit enter/exit worklist (again so long
+   lists don't consume stack). An entry survives the `cyclic-only?` deletion
+   only if it was revisited *before* its first visit completed — that is a
+   cycle, kept for `written`'s datum labels; plain acyclic sharing re-visits
+   only after the exit event deleted the entry, so it re-counts from one and
+   `written` prints it duplicated, exactly like `write`, while
+   `written-shared` keeps it. Moving the delete earlier (e.g. to make the
+   cdr walk a tail call) misses cycles and hangs the writer on circular
+   data; moving it later labels sharing under plain `written`, which the
+   spec reserves for `-shared`.
+3. **Label numbers are assigned at first emission, not by the walk.** The
+   count stored in the shared table is a placeholder; `%gen-shared-ref`
+   overwrites it from a counter at the moment the label is first printed,
+   so numbering follows output order (`#0=` then `#0#`) regardless of hash
+   iteration order.
+
+A test-harness trap worth knowing: the library search order is explicit
+`--lib-path`, then the auto-added script dir, `~/.kaappi/lib`, and the
+exe-relative `zig-out/lib`, all *before* the cwd `lib/` fallback — so an
+installed or previously-built copy silently shadows edits to the source
+tree, and `zig build` (which re-installs `lib/` into `zig-out/lib`) must be
+rerun after every `.sld` edit. The shell suites isolate this with a temp
+`KAAPPI_HOME`; do the same when running
+`tests/scheme/srfi/srfi166-audit.scm` by hand.
+
 ### SRFI 63 — homogeneous and heterogeneous arrays
 
 SRFI 63 (homogeneous and heterogeneous arrays), the third piece of #1694's
--- a/lib/srfi/166/base.sld
+++ b/lib/srfi/166/base.sld
@@ -233,13 +233,18 @@
              st2)))))
 
     ;; with! is a procedure taking flat (state-var value) pairs and setting
-    ;; them permanently (the state variables are first-class values).
+    ;; them permanently (the state variables are first-class values).  An
+    ;; immutable state variable "can only be dynamically bound with with, and
+    ;; not set with with!" — with! on one is an error.
     (define (with! . bindings)
       (lambda (st)
         (let loop ((bs bindings))
           (if (null? bs)
               st
               (begin
+                (if (state-variable-immutable? (car bs))
+                    (error "with!: immutable state variable"
+                           (state-variable-name (car bs))))
                 (%st-set! st (car bs) (cadr bs))
                 (loop (cddr bs)))))))
 
@@ -449,10 +454,45 @@
     (define (fitted/both w . fmts)
       (padded/both w (apply trimmed/both (cons w fmts))))
 
-    ;; A non-lazy stand-in for trimmed/lazy: correct for finite output (the
-    ;; infinite-stream laziness has no observable effect for the finite tests).
+    ;; Spec: "a variant of trimmed which generates each fmt in left-to-right
+    ;; order, and truncates and terminates immediately if more than width
+    ;; characters are generated.  It does not output ellipsis.  Thus this is
+    ;; safe to use with an infinite amount of output, e.g. from
+    ;; written-simply on an infinite list."
+    ;;
+    ;; Laziness is realized through the `output` state variable: the writer
+    ;; streams one token at a time (see %write-stream), so a counting output
+    ;; hook sees each chunk as it is generated and, once the budget is spent,
+    ;; terminates the whole walk by invoking an escape continuation — no
+    ;; amount of further generation can follow.  The output binding is
+    ;; restored on both exit paths (a plain `with` would skip its restore on
+    ;; the escape path), and chunks are delegated to the output binding found
+    ;; on entry so nested uses compose.
     (define (trimmed/lazy w . fmts)
-      (apply trimmed/right (cons w fmts)))
+      (lambda (st)
+        (call/cc
+          (lambda (return)
+            (let ((kept 0)
+                  (saved (%st-ref st output)))
+              (define (finish st2)
+                (%st-set! st2 output saved)
+                (return st2))
+              (define (lazy-output str)
+                (lambda (st2)
+                  (let ((room (- w kept)))
+                    (cond
+                      ((<= room 0) (finish st2))
+                      ((<= (string-length str) room)
+                       (set! kept (+ kept (string-length str)))
+                       ((saved str) st2))
+                      (else
+                        (set! kept w)
+                        (finish ((saved (substring str 0 room)) st2)))))))
+              (let ((st2 (begin
+                           (%st-set! st output lazy-output)
+                           ((apply each fmts) st))))
+                (%st-set! st2 output saved)
+                st2))))))
 
     ;;; ================================================== escaping
 
@@ -689,27 +729,52 @@
 
     ;;; =============================================== shared structures
 
+    ;; Iterative depth-first walk (explicit enter/exit worklist) so a long
+    ;; list does not consume stack.  The `exit` event runs the same
+    ;; delete-if-not-revisited check the recursive original ran after a
+    ;; subtree completed -- and the timing of that check is exactly what
+    ;; distinguishes a cycle (revisited before its first visit completes, so
+    ;; the count is already 2 and the entry survives `cyclic-only?` deletion)
+    ;; from plain sharing (revisited only after the entry was deleted, so it
+    ;; is re-counted from 1 and is kept only when `cyclic-only?` is #f).
     (define (extract-shared-objects x cyclic-only?)
       (let ((seen (make-hash-table eq?)))
-        (let find ((x x))
-          (cond
-            ((or (pair? x) (vector? x))
-             (hash-table-update!/default seen x (lambda (n) (+ n 1)) 0)
-             (cond
-               ((> (hash-table-ref seen x) 1))
-               ((pair? x) (find (car x)) (find (cdr x)))
-               ((vector? x)
-                (let ((len (vector-length x)))
-                  (do ((i 0 (+ i 1))) ((= i len)) (find (vector-ref x i))))))
-             (if (and cyclic-only? (<= (hash-table-ref/default seen x 0) 1))
-                 (hash-table-delete! seen x)))))
-        (let ((res (make-hash-table eq?)) (count 0))
-          (hash-table-walk seen
-            (lambda (k v)
-              (cond ((> v 1)
-                     (hash-table-set! res k (cons count #f))
-                     (set! count (+ count 1))))))
-          (cons res 0))))
+        (let loop ((todo (list (cons 'enter x))))
+          (if (null? todo)
+              (let ((res (make-hash-table eq?)) (count 0))
+                (hash-table-walk seen
+                  (lambda (k v)
+                    (cond ((> v 1)
+                           (hash-table-set! res k (cons count #f))
+                           (set! count (+ count 1))))))
+                (cons res 0))
+              (let ((ev (car todo))
+                    (rest (cdr todo)))
+                (cond
+                  ((eq? (car ev) 'exit)
+                   (if (and cyclic-only?
+                            (<= (hash-table-ref/default seen (cdr ev) 0) 1))
+                       (hash-table-delete! seen (cdr ev)))
+                   (loop rest))
+                  ((not (or (pair? (cdr ev)) (vector? (cdr ev)))) (loop rest))
+                  (else
+                   (let ((x (cdr ev)))
+                     (hash-table-update!/default seen x (lambda (n) (+ n 1)) 0)
+                     (cond
+                       ((> (hash-table-ref seen x) 1) (loop rest))
+                       ((pair? x)
+                        (loop (cons (cons 'enter (car x))
+                                    (cons (cons 'enter (cdr x))
+                                          (cons (cons 'exit x) rest)))))
+                       (else
+                        (loop (let ((len (vector-length x)))
+                                (let build ((i (- len 1))
+                                            (acc (cons (cons 'exit x) rest)))
+                                  (if (< i 0)
+                                      acc
+                                      (build (- i 1)
+                                             (cons (cons 'enter (vector-ref x i))
+                                                   acc))))))))))))))))
 
     (define (%gen-shared-ref cell shares)
       (set-car! cell (cdr shares))
@@ -717,27 +782,16 @@
       (set-cdr! shares (+ (cdr shares) 1))
       (number->string (car cell)))
 
-    (define (%shared-ref-prefix obj shares proc)
-      (let ((cell (hash-table-ref/default (car shares) obj #f)))
-        (cond
-          ((and (pair? cell) (cdr cell))
-           (string-append "#" (number->string (car cell)) "#"))
-          ((pair? cell)
-           (string-append "#" (%gen-shared-ref cell shares) "=" (proc)))
-          (else (proc)))))
-
-    (define (%shared-ref-cdr obj shares proc)
-      (let ((cell (hash-table-ref/default (car shares) obj #f)))
-        (cond
-          ((and (pair? cell) (cdr cell))
-           (string-append ". #" (number->string (car cell)) "#"))
-          ((pair? cell)
-           (string-append ". #" (%gen-shared-ref cell shares) "=(" (proc) ")"))
-          (else (proc)))))
-
-    ;; Flatten obj to a string, labelling shared structure, with numbers
-    ;; formatted according to radix/precision.
-    (define (%write-flat obj shares radix precision)
+    ;; Stream obj to (emit string) one token at a time ("(", each element,
+    ;; separators, ")"), labelling shared structure per the `shares` table,
+    ;; with numbers formatted per radix/precision.  Chunked emission is what
+    ;; makes `trimmed/lazy` able to truncate and *terminate* an infinite
+    ;; generator such as written-simply on a circular list: emit runs after
+    ;; every token, so an `output` override observing the chunks can stop the
+    ;; walk.  The list/vector spines iterate tail-recursively so a long list
+    ;; does not consume stack (the previous whole-string builder recursed
+    ;; non-tail once per element).
+    (define (%write-stream obj shares radix precision emit)
       ;; Per the spec, `written` uses the radix only for 2/8/10/16 (the
       ;; readable radices), and fixed-point precision only when the radix is
       ;; 10.  Precision must not disable the radix branch for non-decimal
@@ -749,56 +803,84 @@
              (if precision (%number->string n 10 precision #f #f #f #f #f) (number->string n)))
             ((and cell (exact? n)) (string-append (cdr cell) (number->string n (car cell))))
             (else (%number->string n 10 precision #f #f #f #f #f)))))
-      (let wr ((obj obj))
-        (%shared-ref-prefix
-          obj shares
-          (lambda ()
-            (cond
-              ((pair? obj)
-               (string-append
-                 "("
-                 (let lp ((ls obj))
-                   (let ((rest (cdr ls)))
-                     (string-append
-                       (wr (car ls))
-                       (cond
-                         ((null? rest) "")
-                         ((pair? rest)
-                          (string-append " " (%shared-ref-cdr rest shares (lambda () (lp rest)))))
-                         (else (string-append " . " (wr rest)))))))
-                 ")"))
-              ((vector? obj)
-               (let ((len (vector-length obj)))
-                 (if (zero? len)
-                     "#()"
-                     (let loop ((i 0) (acc "#("))
-                       (if (= i len)
-                           (string-append acc ")")
-                           (loop (+ i 1)
-                                 (string-append acc (if (zero? i) "" " ")
-                                                (wr (vector-ref obj i)))))))))
-              ((number? obj) (write-number obj))
-              (else (%write-to-string obj)))))))
+      (define (emit-shared-cdr rest k)
+        ;; k continues the spine walk from rest's elements.
+        (let ((cell (hash-table-ref/default (car shares) rest #f)))
+          (cond
+            ((and (pair? cell) (cdr cell))
+             (emit (string-append ". #" (number->string (car cell)) "#")))
+            ((pair? cell)
+             (emit (string-append ". #" (%gen-shared-ref cell shares) "=("))
+             (k)
+             (emit ")"))
+            (else (k)))))
+      (define (wr obj)
+        (let ((cell (hash-table-ref/default (car shares) obj #f)))
+          (cond
+            ((and (pair? cell) (cdr cell))
+             (emit (string-append "#" (number->string (car cell)) "#")))
+            ((pair? cell)
+             (emit (string-append "#" (%gen-shared-ref cell shares) "="))
+             (wr-body obj))
+            (else (wr-body obj)))))
+      (define (wr-body obj)
+        (cond
+          ((pair? obj)
+           (emit "(")
+           (let lp ((ls obj))
+             (wr (car ls))
+             (let ((rest (cdr ls)))
+               (cond
+                 ((null? rest))
+                 ((pair? rest)
+                  (emit " ")
+                  (emit-shared-cdr rest (lambda () (lp rest))))
+                 (else (emit " . ") (wr rest)))))
+           (emit ")"))
+          ((vector? obj)
+           (emit "#(")
+           (let ((len (vector-length obj)))
+             (do ((i 0 (+ i 1)))
+                 ((= i len))
+               (if (positive? i) (emit " "))
+               (wr (vector-ref obj i))))
+           (emit ")"))
+          ((number? obj) (emit (write-number obj)))
+          (else (emit (%write-to-string obj)))))
+      (wr obj))
+
+    ;; Flatten obj to a single string: %write-stream accumulated in order.
+    ;; Used where a whole string is genuinely needed, e.g. pretty's
+    ;; does-it-fit decision.
+    (define (%write-flat obj shares radix precision)
+      (let ((acc '()))
+        (%write-stream obj shares radix precision
+                       (lambda (s) (set! acc (cons s acc))))
+        (apply string-append (reverse acc))))
+
+    ;; Stream obj's tokens through the `output` state variable, threading the
+    ;; state through each chunk.  `written` alone consults the `writer` state
+    ;; variable (it is the mapper for *automatic* formatting of non-string
+    ;; values via displayed's fallback; -shared/-simply are explicit).
+    (define (%write-to-state st obj shares)
+      (let ((cur (list st)))
+        (%write-stream obj shares (%st-ref st radix) (%st-ref st precision)
+                       (lambda (s) (set-car! cur (%output-string (car cur) s))))
+        (car cur)))
 
     (define (written obj)
       (lambda (st)
         (let ((w (%st-ref st writer)))
           (if w
               ((w obj) st)
-              (%output-string st
-                (%write-flat obj (extract-shared-objects obj #t)
-                             (%st-ref st radix) (%st-ref st precision)))))))
+              (%write-to-state st obj (extract-shared-objects obj #t))))))
 
     (define (written-shared obj)
       (lambda (st)
-        (%output-string st
-          (%write-flat obj (extract-shared-objects obj #f)
-                       (%st-ref st radix) (%st-ref st precision)))))
+        (%write-to-state st obj (extract-shared-objects obj #f))))
 
     (define (written-simply obj)
       (lambda (st)
-        (%output-string st
-          (%write-flat obj (extract-shared-objects #f #f)
-                       (%st-ref st radix) (%st-ref st precision)))))
+        (%write-to-state st obj (extract-shared-objects #f #f))))
 
     ))
--- a/tests/scheme/srfi/srfi166-audit.scm
+++ b/tests/scheme/srfi/srfi166-audit.scm
@@ -908,6 +908,154 @@
         (if (= i (string-length s)) n
             (loop (+ i 1) (if (char=? (string-ref s i) #\newline) (+ n 1) n)))))))
 
+;;; ================================ completing the #2292 rework
+
+;; Spec, (trimmed/lazy width fmt ...): "a variant of trimmed which generates
+;; each fmt in left-to-right order, and truncates and terminates immediately
+;; if more than width characters are generated.  It does not output ellipsis.
+;; Thus this is safe to use with an infinite amount of output, e.g. from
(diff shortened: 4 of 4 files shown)


baijum:
**Review (Claude Code, built + ran the full audit)** — ✅ approve

Substantial and correct. `base.sld` gets a genuine rework, not a stand-in: `trimmed/lazy` is now truly lazy (a `call/cc` escape + a streaming `output` state variable that terminates the walk once the budget is spent), `%write-flat` is re-expressed as a token-streaming `%write-stream` (so an infinite `written-simply` can be truncated *and terminated*), `extract-shared-objects` becomes an explicit enter/exit worklist (no stack growth on long lists), and immutable state variables are enforced in `with!`.

I built this branch and ran the full SRFI-166 audit: **285 passed, 0 failed.** I also hand-verified the flagship claims — an infinite circular `written-simply` under `trimmed/lazy 7` terminates (`(1 2 1 `), a 50,000-element `written` does not overflow, `with!` on an immutable variable raises, and the ellipsis is suppressed even when set.

⚠️ **Local-verification footgun (not a defect in this PR):** the SRFI-166 library bytecode cache in `~/.kaappi/cache` can be stale from an earlier build and silently load the *pre-#2292* library — I first saw 2 failures + 2 stack overflows that vanished entirely after clearing the cache (fresh `KAAPPI_HOME`). Anyone verifying this PR (or #2292) locally should clear that cache first; CI is unaffected (fresh env).

The 10 issues span base/pretty/columnar/unicode sub-libraries; #2292 (on main) fixed the sub-libraries and this PR finishes the base-level gaps, so the full-audit-green result substantiates closing all ten.

baijum:
**Follow-up (Claude Code) — ⛔ merge blocker: `gc-stress-scheme` fails (timeout), not a flake**

The `gc-stress-scheme` check failed on this PR ([run](https://github.com/kaappi/kaappi/actions/runs/32895055723)). Two files hit the 900 s per-file cap (exit 124):

- `tests/scheme/srfi/srfi166-audit.scm` — this PR's `(written (make-list 50000 'x))` stack-safety assertion (line ~994). Under `-Dgc-stress=true` (a full collection on *every* allocation) a 50 000-element streamed write is astronomically slow — far past 900 s.
- `tests/scheme/srfi/srfi146-differential.scm` — collateral: the suite runs `jobs: 4` in parallel, so the runaway above starves a sibling lane and tips it over too.

Evidence it's this PR and not a pre-existing flake: every other Wave-2 PR's `gc-stress-scheme` ran green against the *original* (smaller) srfi166-audit; the timeout appears only here, where this PR enlarged it. Re-running won't help — the 50k write genuinely can't finish in 900 s under stress.

`CLAUDE.md`: *"loop-heavy tests should scale counts down under gc_stress."* There's no `gc-stress` cond-expand feature, so a `.scm` file can't self-scale — the precedented fix is to add the file to `KAAPPI_GC_STRESS_SKIP`, exactly as `tests/scheme/srfi/srfi178-segment-stack-2084.scm` does (its header: *"KAAPPI_GC_STRESS_SKIP lists this file (reason (a), too slow under stress)"*). Add `srfi166-audit.scm` to that env list in `.github/workflows/ci.yml` (with a one-line reason), and annotate the test. The normal `zig build test` / non-stress corpus still exercises the 50k assertion, so stack-safety coverage is retained.

Everything else on this PR verified clean (full SRFI-166 audit 285/0 on a fresh cache). Just this gc-stress gate to clear.

PR #2345, opened 2026-08-25
records: carry field names on R7RS define-record-type rtds (#2088)

### LINKED ISSUE #2088: [SRFI-240/237] An R7RS define-record-type produces an rtd with zero field names, so record-accessor/mutator are unusable on it
**Category:** SRFI
**Spec reference:** SRFI 240 (Reconciled Records), <https://srfi.schemers.org/srfi-240/>, whose entire content is a `define-record-type` accepting **either** R6RS clause syntax **or** the R7RS/SRFI-9 positional syntax, "both producing interoperable record types", with everything else inherited from SRFI 237. `lib/srfi/240.sld`'s own header repeats that claim verbatim. SRFI 237 §"Inspection": "`(record-type-field-names rtd)` Returns a vector of symbols naming the fields of the record type."
**Severity:** wrong-result

A record type created by R7RS positional `define-record-type` reports **zero** own field names to the SRFI 237/240 inspection layer. `record-type-field-names` does not error — it returns `#()`, a well-formed but false answer — and `record-accessor`, `record-mutator` and `record-field-mutable?` are consequently unusable on such a type, all three failing with `index 0 out of range for length 0`.

**Minimal reproduction:**

```scheme
(import (scheme base) (scheme write) (srfi 240))
(define-record-type <pt> (mk-pt x y) pt? (x pt-x) (y pt-y set-pt-y!))
(define o (mk-pt 1 2))
(write (record-type-field-names (record-rtd o)))   ; expected: #(x y)   actual: #()
(newline)
((record-accessor (record-rtd o) 0) o)
;; expected: 1
;; actual:   error[KP3006]: %record-field-mutable?: index 0 out of range for length 0
```

**Discriminating control:** the same two-field, one-mutable shape built the two other ways is fully introspectable, so the defect is specific to the R7RS positional desugarer rather than to the inspection layer:

```scheme
(define-record-type <r6> (fields (immutable a) (mutable b)))
(write (record-type-field-names (record-rtd (make-<r6> 1 2))))          ; #(a b)
(write (record-type-field-names
        (make-record-type-descriptor 'p #f #f #f #f '#((immutable a) (mutable b)))))  ; #(a b)
```

A second control shows the field *count* is known somewhere: a constructor derived from the R7RS record's own rtd takes two arguments and builds a valid instance —

```scheme
(record-predicate (record-rtd o))                                 ; works
((record-constructor (make-record-constructor-descriptor (record-rtd o) #f #f)) 1 2)  ; a valid <pt>
```

— so `record?`, `record-rtd`, `record-type-name`, `record-predicate` and `record-constructor` all work on the same rtd whose field metadata is empty.

**Source location:** `src/types_record.zig` documents the cause directly: `own_field_count` is "`0` for a plain R7RS record type, which doesn't track individual field metadata." The R7RS branch of `src/vm_records.zig`'s `define-record-type` desugarer never populates `own_field_names` / `own_field_mutable`; the R6RS branch and `make-record-type-descriptor` (`src/primitives_srfi237.zig`) both do.

Distinct from #1974 (closed), which was about `record-type-field-names` returning a *list* rather than a vector, plus three other R6RS conformance defects, all on rtds that did have field metadata.

**Disabled test:** `tests/scheme/srfi/srfi240-audit.scm` — 4 assertions marked `;; FAIL: #2088`, with the current behaviour pinned enabled alongside them so the fix flips both sets.

baijum:
## What

An R7RS positional `define-record-type` produced an rtd with **zero own field names**, so the SRFI 237/240 inspection layer could not see its fields: `record-type-field-names` returned `#()` (a well-formed but false answer) and `record-accessor` / `record-mutator` / `record-field-mutable?` all failed with `index 0 out of range for length 0`. SRFI 240's entire content is a `define-record-type` accepting either syntax "both producing interoperable record types" — the R6RS-clause and `make-record-type-descriptor` paths already carried the metadata; only the R7RS positional path dropped it.

## Why it happened

`RecordType.own_field_names`/`own_field_mutable` were added for SRFI 237 and populated by the R6RS-clause desugarer and the procedural layer, but all three R7RS emit paths created the rtd count-only (`allocRecordType`) or emitted a count-only `(%make-record-type "name" n)` call.

## The fix

All three R7RS emit paths now record each field's name and mutability (mutable iff the field's clause names a mutator, R7RS 5.5.1):

- `handleDefineRecordType` (top level and library bodies): allocates via `allocRecordTypeExtended` — parentless, generative, transparent, i.e. exactly the shape `allocRecordType` built, plus the field metadata.
- `expandRecordTypeDefines` (leading-define body scanning) and `compileDefineRecordType` (general dispatch): emit the metadata through `%make-record-type`, which gains an **optional third argument** — a list of `(name-string . mutable?)` pairs, the same convention `%make-record-type-descriptor`'s field-specs already use. Count-only callers keep the two-argument form.

The rtd representation is unchanged: `own_field_names`/`own_field_mutable` already existed and are already handled by every GC switch (raw owned bytes, same category as `RecordType.name`); no new heap fields, no new tracing. `gc_deep_copy`'s metadata-ful slow path now carries R7RS field names across SRFI-18 thread boundaries (verified by the cross-thread-boundary audit). Stale comments documenting the old behavior (`types_record.zig`'s "0 for a plain R7RS record type", `gc_deep_copy.zig`'s fast-path rationale) are updated.

## Tests added

- `tests/scheme/srfi/srfi240-audit.scm` — the four assertions disabled under `;; FAIL: #2088` are enabled and the pinned broken-behavior set flipped to assert the fixed behavior; plus new coverage: by-name `record-accessor`, immutable-field mutator rejection, out-of-range index rejection, a zero-field type staying legitimately `#()`, and a body-local `define-record-type` carrying its names. **9 of these fail without the fix** (verified by stashing the `src/` changes and rebuilding).
- `src/tests_records.zig` — two unit tests asserting `own_field_names`/`own_field_mutable` on the rtd for the top-level and body-local desugarer paths.
- `tests/scheme/audit/internal-primitives-audit.scm` — `%make-record-type` arity assertions updated for the optional third argument (the old `rt: too many` 3-arg call now pins the specs-not-a-list rejection instead), plus positive/mismatch tests for the field-specs form.

## Evidence

- `srfi240-audit.scm`: 81/81 pass (72 pass + 9 unexpected failures without the fix)
- `internal-primitives-audit.scm`: 255/255
- srfi 9 / 57 / 131 / 136 / 137 / 150 / 237 / 240 suites: all pass
- R7RS suite: 1395 pass, 0 fail
- `zig build test`: green
- Native tier: a no-import record program through `kaappi compile` prints its field names (`(x y)`)

Closes #2088

### DIFF
--- a/src/compiler_lambda.zig
+++ b/src/compiler_lambda.zig
@@ -907,12 +907,28 @@ pub fn compileDefineRecordType(self: *Compiler, args: Value, dst: u16) CompileEr
         forms = gc.allocPair(gc.makeList(&[_]Value{ define_sym, np, body }) catch return CompileError.OutOfMemory, forms) catch return CompileError.OutOfMemory;
     }
 
-    // Internal record type: (define __rt (%make-record-type "name" n))
+    // Internal record type: (define __rt (%make-record-type "name" n
+    //   (quote (("fname" . mutable?) ...)))) -- the quoted field-specs list
+    // carries the per-field metadata (#2088) in the same (name-string .
+    // mutable?) shape %make-record-type-descriptor uses, so the SRFI 237/240
+    // inspection layer sees this type's field names. Mutable iff the field's
+    // clause names a mutator (R7RS 5.5.1).
     {
         const mrt = globals_mod.baseBindingSymbol(gc, "%make-record-type") catch return CompileError.OutOfMemory;
         const ns = gc.allocString(spec.type_name) catch return CompileError.OutOfMemory;
         const nf = types.makeFixnum(@intCast(spec.field_count));
-        const init = gc.makeList(&[_]Value{ mrt, ns, nf }) catch return CompileError.OutOfMemory;
+
+        var spec_elems: [256]Value = undefined;
+        for (0..spec.field_count) |fi| {
+            const fname_str = gc.allocString(spec.field_names[fi]) catch return CompileError.OutOfMemory;
+            const fmut = if (spec.mutator_names[fi] != null) types.TRUE else types.FALSE;
+            spec_elems[fi] = gc.allocPair(fname_str, fmut) catch return CompileError.OutOfMemory;
+        }
+        const specs_list = gc.makeList(spec_elems[0..spec.field_count]) catch return CompileError.OutOfMemory;
+        const quote_sym = gc.allocSymbol("quote") catch return CompileError.OutOfMemory;
+        const quoted_specs = gc.makeList(&[_]Value{ quote_sym, specs_list }) catch return CompileError.OutOfMemory;
+
+        const init = gc.makeList(&[_]Value{ mrt, ns, nf, quoted_specs }) catch return CompileError.OutOfMemory;
         const rt_sym = gc.allocSymbol(internal_name) catch return CompileError.OutOfMemory;
         forms = gc.allocPair(gc.makeList(&[_]Value{ define_sym, rt_sym, init }) catch return CompileError.OutOfMemory, forms) catch return CompileError.OutOfMemory;
     }
--- a/src/gc_deep_copy.zig
+++ b/src/gc_deep_copy.zig
@@ -324,11 +324,14 @@ fn deepCopyValue(gc: *GC, src: Value, visited: *std.AutoHashMap(usize, Value)) D
             if (rt.parent == null and rt.own_field_count == 0 and rt.uid == null and
                 !rt.sealed and !rt.is_opaque and !rt.has_protocol)
             {
-                // Plain R7RS record type (the common case) -- no SRFI 237
-                // metadata to carry across. `has_protocol` joins the guard
-                // rather than being copied afterwards: allocRecordType makes
-                // a type with no SRFI 237 metadata at all, so a type that has
-                // any must take the slow path below.
+                // Count-only record type (%make-record-type with no field
+                // specs) -- no metadata of any kind to carry across. R7RS
+                // define-record-type types carry field names since #2088,
+                // so they take the slow path below, as SRFI 237 types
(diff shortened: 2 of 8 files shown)


baijum:

The skeptical worry for a record-type change — "a new Value field on `RecordType` must be threaded through every GC switch" — **does not apply here**: `own_field_names` (`[][]const u8`) and `own_field_mutable` (`[]bool`) already existed on the struct (used by the R6RS-clause / procedural `%make-record-type-descriptor` paths) and are *raw owned memory, not Values*. The R7RS path now simply populates them via the pre-existing `allocRecordTypeExtended` (which dupes every slice before `maybeCollect`). So `markObjectContents`/`referencesYoung`/printer are correctly untouched, and `objectSize`/`freeObject`/deep-copy already handle them — the switches that look "missing" are correctly absent. All three R7RS desugar paths (top-level handler, body-local `expandRecordTypeDefines`, compiled `compileDefineRecordType`) now carry metadata consistently, mutable iff a mutator is named. `compiler_lambda.zig` is not scope creep — `compileDefineRecordType` already lived there.

Coverage is strong: `srfi240-audit` flips the four `;; FAIL: #2088` assertions (field-names → `#(x y)`, accessor by index **and** name, mutator, immutable-field rejection), keeps a zero-field type legitimately `#()`, and adds unit tests asserting the slices directly.

Nit: the `>255` specs guard returns bare `TypeError` while the count/spec disagreement uses `argError` — minor taxonomy inconsistency, and `>255` is unreachable via the desugarer. Recommend confirming the `kaappi compile` native-record path once (the only path with no unit test; shares spec-emission code with the tested ones, so low risk).

PR #2346, opened 2026-08-25
thottam: regression tests for state-file name validation (#2144)

### LINKED ISSUE #2144: [Tooling] thottam list/verify/update build paths from unvalidated package names — isValidPkgName guards install and remove only

baijum:
## What

**Verify-and-close for #2144.** The code fix already landed on `main` with #2289 (2026-08-23): `isValidPkgName` now guards every read path out of `installed.txt` / `thottam.lock` in `doList`, `doUpdate` (named and all-packages forms) and `doVerify` (lockfile structure pass and installed walk), in addition to the pre-existing install/remove guards. This PR pins that behaviour with the regression tests the issue demands, plus one adjacent one-line fix found while reproducing.

## Why

Nothing covered the #2289 guards: reverting all five would have passed the entire suite. Per the audit-campaign protocol (a filed defect needs an assertion that states the property that should hold), these tests hold the line "a package name cannot name a path outside `$KAAPPI_HOME`" against future refactors of the read paths.

## Changes

- `src/tests_thottam.zig` — five regression tests driving the real commands against a throwaway `$KAAPPI_HOME` whose state files carry a traversal-shaped name (`../../<dir>`, the issue's exact repro shape). A `captureOutput` helper swaps fd 1/2 to a pipe around the call (the commands print straight to the descriptors; `pipe`/`dup`/`dup2` are CRT calls on Windows too, so the swap is portable). `TmpHome` builds the home, `src`/`lib`, and the escaped target directory as a plain non-repo, so a regression fails by assertion rather than by killing the test runner.
- `src/thottam.zig` — `Config`, `doList`, `doUpdate`, `doVerify` made `pub` so the test file can exercise them (no behaviour change), and `main()`'s `update` handler now lists `InvalidPackageName` alongside the errors that already print their own message: `thottam update <bad-name>` used to print the proper `invalid package name` error and then a second raw `error: InvalidPackageName` line from Zig's default handler — install and remove have suppressed exactly that since kaappi#2132.

## Tests

Each test fails with the #2289 guards stripped and passes with them (verified by stripping all five guards, running, and restoring):

| Test | Pre-fix behaviour it pins out |
|------|-------------------------------|
| `doList omits a state-file name that escapes $KAAPPI_HOME` | list printed the traversal name and read `kaappi.pkg` through it |
| `doVerify reports a hostile lockfile name as corruption and fails` | verify exited **clean** — the line was structurally sound, nothing installed meant nothing else to check |
| `doVerify reports a hostile installed.txt name as corruption, not a lockfile finding` | reported `UNLOCKED` after joining the name onto `src_dir` for `getPkgSha` |
| `doUpdate rejects a hostile name given on the command line` | returned `NotInstalled` instead of the loud `InvalidPackageName` |
| `doUpdate of every package skips hostile names read back from installed.txt` | announced `Updating <traversal>...` and ran git inside the escaped directory |

End-to-end repro of the issue's scenario (hostile lines in `installed.txt` + `thottam.lock`, scratch git repo outside `$KAAPPI_HOME`):

```console
$ thottam list            # hostile line skipped entirely
$ thottam verify          # MALFORMED: ../../repo deadbeef / ../../repo — Verification failed. (exit 1)
$ thottam update ../../repo   # error: invalid package name '../../repo' (...) — exit 1, one line
$ thottam update          # nothing announced, exit 0; escaped repo SHA untouched
```

`zig build test`: green (1827/1834 pass, 7 skip; the 7 are platform skips). One note: `tests_fuzz`'s `portable-subset generator: programs evaluate without error` flaked twice under full-suite load during development (its own comment allows for deadline jitter on a loaded machine) — it passes in isolation on both this tree and pristine `main`, and the full suite is green on this branch, so it is pre-existing load flakiness, unrelated to this change.

Closes #2144

### DIFF
--- a/src/tests_thottam.zig
+++ b/src/tests_thottam.zig
@@ -24,6 +24,7 @@ const platform = @import("platform.zig");
 const thottam = @import("thottam.zig");
 const semver = @import("thottam_semver.zig");
 const state = @import("thottam_state.zig");
+const tfs = @import("thottam_fs.zig");
 
 const Semver = semver.Semver;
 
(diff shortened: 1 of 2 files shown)


baijum:

The validator `isValidPkgName` (thottam_state.zig) is a strict allowlist (`[A-Za-z0-9_-]`, empty rejected), airtight against path traversal — `..`, `.`, `/`, `\`, `:`, absolute paths, NUL, and whitespace are all rejected. Every consumer that joins a name onto `src_dir`/`lib_dir` (getPkgSha/getPkgManifest/joinPath) is reachable only downstream of a guard, across install/remove/list/update/verify. Fails safe (verify → `MALFORMED`+`VerifyFailed`; the command entries → clear `InvalidPackageName`). The five regression tests drive a real `../../<outside>` name through list/verify/update and each fails pre-fix.

Framing worth stating in the PR: the security fix itself landed in #2289; this PR adds the pinning tests + one functional fix (`InvalidPackageName` wired into the `update` command's exit handler) + makes `Config`/`doList`/`doUpdate`/`doVerify` `pub`.

Nits (non-blocking): validation is per-call-site, not the type-level "validate-on-read" the issue suggested — all current consumers are covered, but a future reader of `installed.txt`/`thottam.lock` must remember to guard. `list`/`update-all` silently `continue` past a hostile line while `verify` fails loud (defensible, but a corrupted line vanishes from `list` with no signal). Neither verify test drives `getPkgSha` on the hostile name (guard sits before it, so the guard is still proven).

PR #2347, opened 2026-08-25
Error taxonomy sweep: bounds -> KP3006, value rejections -> KP3007, real procedure names (#2020/#2021/#2022)

### LINKED ISSUE #2020: Error taxonomy: 33 bounds failures across 8 primitives files report KP3002 — substring and string-copy are one operation with two codes
**Category:** Diagnostics / error taxonomy · **Severity:** diagnostic quality · **Found by:** Systematic audit v2, Phase 2.13 (#1890)

33 bounds failures across 8 `primitives_*.zig` files report `KP3002 type error` instead of `KP3006 index-out-of-bounds`. The single largest contributor is `primitives.parseOptionalRange`, the shared `[start end]` parser behind **20 procedures**.

This is the D2 sweep the strategy doc predicted, minus the parts already filed: #1944/#1972/#2016 (io), #1978 (filesystem), #2002 (fibers), #1914 (`%` internals).

## The finding in one line

R7RS §6.7 defines `(substring string start end)` as *"exactly the same as `string-copy`"*. They are the same operation under two names, and they disagree on the code:

```scheme
(substring   "ab" 0 99)   ;; KP3006  substring: index 99 out of range for length 2
(string-copy "ab" 0 99)   ;; KP3002  type error in 'string-copy': expected valid index, got 99
```

`99` is an exact integer, which is the type `string-copy` wants. Nothing about the type is wrong.

## Discriminating controls

Every one of these is a *sibling that gets it right*, so this is a defect and not a house convention:

| Correct (KP3006) | Wrong (KP3002) | Relationship |
|---|---|---|
| `(substring "ab" 0 99)` | `(string-copy "ab" 0 99)` | R7RS defines them as the same operation |
| `(vector-ref v 5)` | `(bytevector-u8-ref bv 5)` | sibling accessors; **KP3006's own `explain` text names `bytevector-u8-ref` as its example** |
| `(list-ref l 5)` | `(list-tail l 5)` | sibling list accessors |
| `(s8vector-ref v 9)` | `(u8vector-ref v 9)` | **same SRFI, same operation, two element kinds** |
| `(vector-set! v 5 0)` | `(vector-swap! v 0 9)` | both index a vector |

The SRFI 160 row is the sharpest: `u8vector` is a plain bytevector per SRFI 160's own recommendation, so `u8vector-ref` routes to `bytevector-u8-ref` while all 11 other element kinds route to `%numeric-vector-ref`. Unit 2.4 verified the seam's *values* agree across all 12 kinds; the diagnostic codes do not.

## The documentation says KP3006, twice

`docs/dev/adding-features.md:64-68`:

> | `indexError(proc, index, len)` | `IndexOutOfBounds` (KP3006) | an index outside `0..len` |

`src/diagnostics.zig:426` — the text `kaappi explain KP3006` prints:

> An index passed to a sequence operation (vector-ref, string-ref, **list-ref, bytevector-u8-ref**, ...) was negative or not less than the length of the sequence.

Both `list-ref`-adjacent `list-tail` and the literally-named `bytevector-u8-ref` report KP3002.

For the `start > end` cases, `kaappi explain KP3007` names the case verbatim — *"for example a start index greater than an end index"* — and no procedure produces it: `substring` gives KP3006, everything else gives KP3002.

## Root cause 1 — `parseOptionalRange` (20 procedures)

`src/primitives.zig:543`:

```zig
pub fn parseOptionalRange(args, arg_offset, max_len, proc_name) PrimitiveError!Range {
    ...
    if (s < 0 or @as(usize, @intCast(s)) > max_len) return typeError(proc_name, "valid index", args[arg_offset]);   // :549
    ...
    if (e < 0 or @as(usize, @intCast(e)) > max_len) return typeError(proc_name, "valid index", args[arg_offset+1]); // :555
    if (start > end)                                return typeError(proc_name, "start <= end", args[arg_offset]);  // :558
}
```

Line 547 immediately above `:549` is the *real* type check (`!isFixnum` → `"exact integer"`), and it is correct. Lines 549/555/558 run only after the value is known to be a fixnum, so by construction they can never be type failures.

The 21 call sites cover 20 procedures:

`bytevector-copy` · `bytevector-copy!` · `read-bytevector!` · `reverse-vector->list` · `string->list` · `string->utf8` · `string->vector` · `string-copy` · `string-copy!` · `string-fill!` · `utf8->string` · `vector->list` · `vector->string` · `vector-copy` · `vector-copy!` · `vector-fill!` · `vector-reverse!` · `vector-reverse-copy` · `vector-reverse-copy!` · `write-bytevector`

## Root cause 2 — direct sites

| File | Sites | Procedures |
|---|--:|---|
| `primitives_string.zig` | 9 | `string-copy`, `string-copy!`, `string->list`, `string->vector` (post-`parseOptionalRange` UTF-8 offset conversions) |
| `primitives_vector.zig` | 7 | `vector-copy!`, `vector-swap!`, `vector-reverse-copy!`, `vector-unfold!`, `vector-unfold-right!`, `vector-append-subvectors` |
| `primitives_srfi160.zig` | 6 | `%numeric-vector-set!` element range, `%make-numeric-vector` length |
| `primitives_bytevector.zig` | 5 | `bytevector-u8-ref`, `bytevector-u8-set!`, `bytevector-copy!`, `string->utf8` |
| `primitives.zig` | 3 | `parseOptionalRange` itself |
| `primitives_srfi1.zig` | 2 | `take-right`, `drop-right` |
| `primitives_string_ext.zig` | 1 | `string-replace` |

Separately, `list-tail`/`take`/`drop` report a range overrun as `expected pair, got ()` — honest about the walk, but `list-ref` reports KP3006 for the identical input, so the spelling is a choice rather than a constraint.

## Related: `src/ffi.zig` says "out of range" under KP3002

18 of `src/ffi.zig`'s 27 `return error.TypeError` sites are narrow-integer *range* checks. The message already knows:

```scheme
(abs-int32 4294967296)  ;; KP3002  'abs': argument 1 out of range for int32
(abs-int32 'sym)        ;; KP3002  'abs': argument 1 must be int32, got symbol
```

Same code for both; only the prose distinguishes them. Filed separately as the CI-gate issue, since those sites are also invisible to the bare-TypeError grep gate — but the taxonomy half belongs here.

## Suggested fix

Swap the helper. `parseOptionalRange`'s three sites become `indexError(proc_name, s, max_len)`; the direct sites likewise. The `start > end` case is the one judgment call: `indexError`'s wording (`index N out of range for length L`) does not describe it, so `argError(proc, "start {d} is greater than end {d}", .{s, e})` (KP3007) matches both the failure and `explain KP3007`'s own text.

Note the fix must not over-reach: the *type* branch immediately preceding each of these is correct and must stay KP3002. `tests/scheme/audit/error-taxonomy-audit.scm` pins 7 such genuine type errors for exactly that reason.

## Disabled test

`tests/scheme/audit/error-taxonomy-audit.scm` — 29 assertions marked `;; FAIL: #THIS`, plus 5 `TODAY (#THIS)` assertions pinning current behaviour so the fix is a visible change rather than a silent one. All 29 verified to fail when enabled and pass nowhere else.

**Verified:** fresh ReleaseSafe build at `d0d03deb` (macOS aarch64, isolated `KAAPPI_HOME`); every message above reproduced from that binary, not inferred from format strings. Site counts from `grep` over `src/primitives*.zig`.

**Found by:** Systematic audit v2, Phase 2.13

### LINKED ISSUE #2021: Error taxonomy: 72 sites reject a correctly-typed value as KP3002 — the type branch and the range branch share one message
**Category:** Diagnostics / error taxonomy · **Severity:** diagnostic quality · **Found by:** Systematic audit v2, Phase 2.13 (#1890)

72 sites across 16 `primitives_*.zig` files reject a value of the **correct type** and report `KP3002 type error`. This is `KP3007 invalid-argument`'s own definition. Companion to #2020 (bounds → KP3006); different mechanism, different fix.

## The mechanism: one message serves two different failures

The dominant shape is a type check immediately followed by a range check, sharing a single message string and a single code:

`src/primitives_bytevector.zig:51-53`

```zig
if (!types.isFixnum(args[1])) return typeError("make-bytevector", "exact integer 0-255", args[1]);  // type
const f = types.toFixnum(args[1]);
if (f < 0 or f > 255)         return typeError("make-bytevector", "exact integer 0-255", args[1]);  // range
```

The consequence is that a caller cannot tell the two apart programmatically *or* from the prose:

```scheme
(bytevector 'x)   ;; KP3002  type error in 'bytevector': expected exact integer 0-255, got #<symbol>
(bytevector 256)  ;; KP3002  type error in 'bytevector': expected exact integer 0-255, got 256
```

`256` is an exact integer. The message says it expected an exact integer. The message argues against itself.

## Discriminating control

`primitives_io.zig` already reports this exact class correctly, after #1944/#1969 and #1972/#2016:

```scheme
(let ((p (open-input-string "a"))) (close-port p) (read-char p))
;; KP3007  read-char: input port is closed
```

The port is a port — right type, rejected anyway — and it reports KP3007. So the helper, the house phrasing, and the precedent all exist; these 72 sites predate or skipped them. One file was fixed; the sweep was never finished.

## The documentation says KP3007

`docs/dev/adding-features.md:68`:

> | `argError(proc, fmt, args)` | `InvalidArgument` (KP3007) | the type is fine and the procedure rejects the value anyway |

`src/diagnostics.zig:437` — what `kaappi explain KP3007` prints:

> An argument was of an acceptable type but outside the range or shape the procedure allows — for example a start index greater than an end index, or **a value a procedure explicitly rejects**.

## Reproductions, by kind

**Byte range** (the value is an exact integer):

```scheme
(bytevector 256)                          ;; KP3002  expected exact integer 0-255, got 256
(bytevector-u8-set! (bytevector 1) 0 256) ;; KP3002
(make-bytevector 1 256)                   ;; KP3002
```

**Negative lengths** (the value is an exact integer):

```scheme
(make-vector -1)      ;; KP3002  expected exact non-negative integer, got -1
(make-string -1)      ;; KP3002
(make-bytevector -1)  ;; KP3002
(make-list -1)        ;; KP3002
(make-s8vector -1)    ;; KP3002  expected non-negative length, got -1
```

**Unicode domain** — R7RS §6.6: *"It is an error if n is not a Unicode scalar value."* The argument is an exact integer in all three:

```scheme
(integer->char #x110000)  ;; KP3002  expected valid Unicode scalar value (0..#xD7FF, #xE000..#x10FFFF), got 1114112
(integer->char #xD800)    ;; KP3002  (a surrogate)
(integer->char -1)        ;; KP3002
```

**Enumerated-value rejections** — the message text does not even claim a type:

```scheme
(number->string 10 1)     ;; KP3002  expected radix between 2 and 36, got 1
(null-environment 6)      ;; KP3002  expected 5 or 7, got 6
(hash 'k 0)               ;; KP3002  expected positive integer, got 0
(s8vector 200)            ;; KP3002  expected in-range signed integer, got 200
```

**Immutability** — a property of the object, not its type. The value *is* a pair / string / vector:

```scheme
(set-car! '(1 2) 9)         ;; KP3002  expected mutable pair, got #<pair>
(string-set! "ab" 0 #\x)    ;; KP3002  expected mutable string, got #<string>
(vector-set! '#(1 2) 0 9)   ;; KP3002  expected mutable vector, got #<vector>
```

Note `got #<pair>` beside `expected mutable pair`: the message reports the value's type as the very type it says it wanted.

**A lookup miss is not a type failure.** SRFI 69 says `hash-table-ref` signals an error when the key is absent and no thunk was given. The key is a perfectly good key; it is simply not in the table:

```scheme
(hash-table-ref (make-hash-table) 'nope)
;; KP3002  type error in 'hash-table-ref': expected key to be present or default, got #<symbol>
```

`#<symbol>` (see #1899) means the message cannot even say *which* key was missing.

## Distribution

| File | Sites | Representative |
|---|--:|---|
| `primitives_bytevector.zig` | 13 | `0-255` element/fill range |
| `primitives_vector.zig` | 12 | mutability, `valid character` |
| `primitives_string.zig` | 11 | `valid character`, `integer->char` |
| `primitives_hashtable.zig` | 6 | `positive integer` bound, absent key |
| `primitives_ffi.zig` | 5 | `at most 8 parameter types`, name length |
| `primitives_string_ext.zig` | 4 | `valid character` |
| `primitives_r7rs.zig` | 3 | `5 or 7`, `nanosecond in [0, 999999999]` |
| `primitives_random.zig` | 3 | `positive integer`, `non-all-zero state` |
| `primitives_srfi181.zig` | 3 | `recognized codec` / `eol-style` / `error-handling-mode` |
| `primitives_numeric.zig` | 2 | `radix between 2 and 36`, `finite number` |
| `primitives_filesystem.zig` | 2 | `valid user ID` / `valid group ID` (overlaps #1978) |
| `primitives_random_port.zig` | 2 | `bytevector of length >= 32`, `determinized random port` |
| `primitives.zig` | 2 | `openable library`, `open ffi-library` |
| `primitives_srfi160.zig` | 1 | `known element kind` |
| `primitives_srfi18.zig` | 1 | `new thread` |
| `primitives_list.zig` | 1 | `key to be present` |
| **Total** | **72** | |

Already-filed overlaps, excluded from any fix scoped here: #1978 (filesystem errno), #2002 (fibers), #1914 (`%` internals).

## Files that are clean on this axis

13 of 31 have zero sites of either kind (this issue or #2020): `primitives_arithmetic` · `primitives_char` · `primitives_control` · `primitives_cxr` · `primitives_fiber` · `primitives_lazy` · `primitives_parallel` · `primitives_srfi211` · `primitives_srfi237` · `primitives_srfi254` · `primitives_srfi258` · `primitives_srfi260` · `primitives_sysinfo`.

`primitives_srfi237.zig` is the model — it introduced `argError` and routes sealed-parent and uid-collision rejections through it (see its own `:138-190` comments).

## Suggested fix

Split each conflated branch. The type branch keeps `typeError`; the range/domain branch becomes `argError`, whose comptime format string can also name the value properly — sidestepping #1899 for free, as `docs/dev/adding-features.md:75-77` already points out:

```zig
if (!types.isFixnum(args[1])) return typeError("make-bytevector", "exact integer", args[1]);
const f = types.toFixnum(args[1]);
if (f < 0 or f > 255) return argError("make-bytevector", "fill byte {d} is outside 0-255", .{f});
```

The immutability group may warrant its own phrasing (`argError(proc, "cannot mutate a literal {s}", ...)`) rather than being squeezed into "invalid argument".

## Disabled test

`tests/scheme/audit/error-taxonomy-audit.scm` — 22 assertions marked `;; FAIL: 2021`, plus 10 `TODAY (2021)` assertions pinning current behaviour, including one asserting that the type branch and the range branch currently emit *the same* `expected` text.

**Verified:** fresh ReleaseSafe build at `d0d03deb` (macOS aarch64, isolated `KAAPPI_HOME`); every message reproduced from that binary. Site counts from `grep` over `src/primitives*.zig`.

**Found by:** Systematic audit v2, Phase 2.13

### LINKED ISSUE #2022: 31 procedures misreport which procedure failed: shared helpers hardcode 'string', 'arithmetic', 'string operation'
**Category:** Diagnostics · **Severity:** diagnostic quality · **Found by:** Systematic audit v2, Phase 2.13 (#1890)

20 SRFI-13 procedures report their `[start end]` range failure as coming from a procedure called `string`. `string` is a real, working, entirely unrelated `(scheme base)` procedure, so the message names something that exists and is innocent.

## Reproduction

```scheme
(import (srfi 13))
(string-index    "ab" #\a 0 9)   ;; type error in 'string': expected valid index, got 9
(string-contains "ab" "a"  0 9)  ;; type error in 'string': expected valid index, got 9
(string-count    "ab" #\a 0 9)   ;; type error in 'string': expected valid index, got 9
(string-trim     "ab" #\a 0 9)   ;; type error in 'string': expected valid index, got 9
(string-prefix?  "a" "ab" 0 9)   ;; type error in 'string': expected valid index, got 9
```

Nothing in any of those five calls mentions `string`. A user reading the message, or grepping their source for the named procedure, is sent to the wrong place — and `string` really is callable, so there is no cue that the name is a placeholder.

## Discriminating controls

1. **`string` works.** `(string #\a #\b)` ⟹ `"ab"`. The name is not a stand-in the reader can recognise as one.
2. **A sibling in the same file names itself correctly.** `string-take` does not route through the shared helper:

   ```scheme
   (string-take "ab" 9)   ;; index out of bounds in 'string-take'
   ```

   So the correct name is available at these call sites; it is simply not passed.

## Root cause

`src/primitives_string_ext.zig:137`

```zig
fn parseStartEnd(data: []const u8, args: []const Value, start_arg_idx: usize) PrimitiveError!Range {
    const cp_count = utf8CodepointCount(data);
    const range = try primitives.parseOptionalRange(args, start_arg_idx, cp_count, "string");
    //                                                                            ^^^^^^^^
```

The helper takes no procedure-name parameter, so every one of its 23 call sites inherits the literal.

## Affected procedures (20)

`string-any` · `string-contains` · `string-count` · `string-delete` · `string-every` · `string-filter` · `string-index` · `string-index-right` · `string-pad` · `string-pad-right` · `string-prefix?` · `string-replace` · `string-reverse` · `string-skip` · `string-skip-right` · `string-suffix?` · `string-titlecase` · `string-trim` · `string-trim-both` · `string-trim-right`

10 of the 20 verified live against the fresh build; the other 10 share the identical single code path.

## A second wrong name in the same file

`src/primitives_string_ext.zig:127` reports `"string operation"`, which is not a procedure name at all:

```scheme
(string-index "ab" 42)   ;; type error in 'string operation': expected procedure or char-set, got 42
```

## Why this matters beyond aesthetics

`--diagnostics=json` emits these messages as LSP `Diagnostic` objects, and the LSP server (`src/kaappi_lsp.zig`) surfaces them in an editor. A diagnostic that names the wrong procedure is worse than one that names none: it is confidently wrong, and tooling has no way to detect that.

## Suggested fix

Give `parseStartEnd` a `proc: []const u8` parameter and thread the real name from each of its 23 call sites; likewise for the `"string operation"` site at `:127`. Mechanical, and independently useful for #2020 — the fix there changes what these messages *say*, and this one changes *who they blame*; both should land before either is considered done.

A cheap guard against recurrence: assert in `tests/scheme/audit/error-taxonomy-audit.scm` that each procedure's own name appears in its own message. That form of assertion catches every future shared-helper hardcode, not just this one.

## Disabled test

`tests/scheme/audit/error-taxonomy-audit.scm` — 5 assertions marked `;; FAIL: #2022` asserting each procedure names itself, plus 11 `TODAY (#2022)` assertions pinning the current wrong name, plus both controls.

**Verified:** fresh ReleaseSafe build at `d0d03deb` (macOS aarch64, isolated `KAAPPI_HOME`); all 10 messages and both controls reproduced from that binary. Call-site count from `grep -c 'parseStartEnd(' src/primitives_string_ext.zig`.

**Found by:** Systematic audit v2, Phase 2.13

---

## Update: the same root cause has a second, larger instance — `'arithmetic'`

Broadening this issue after finding the identical pattern in `primitives_arithmetic.zig` and `primitives_numeric.zig`. Same root cause (**a shared helper hardcodes a placeholder where the caller's own name belongs**), same one-line fix shape, so it belongs here rather than in a separate issue.

`src/primitives_arithmetic.zig:40-42`

```zig
fn numberTypeError(v: Value) PrimitiveError {
    return primitives.typeError("arithmetic", "number", v);
}
```

20 call sites in that file, plus `primitives_arithmetic.zig:77` (`"rational number"`) and `primitives_numeric.zig:959`.

**11 procedures report as `arithmetic`:**

```scheme
(+ 1 'x)      ;; type error in 'arithmetic': expected number, got #<symbol>
(- 1 'x)      ;; 'arithmetic'
(* 1 'x)      ;; 'arithmetic'
(< 1 'x)      ;; 'arithmetic'
(> 1 'x)      ;; 'arithmetic'
(= 1 'x)      ;; 'arithmetic'
(<= 1 'x)     ;; 'arithmetic'
(max 1 'x)    ;; 'arithmetic'
(min 1 'x)    ;; 'arithmetic'
(abs 'x)      ;; 'arithmetic'
(expt 'x 2)   ;; 'arithmetic'
```

**Seven sibling controls in the same two files get it right**, which is what makes this a defect rather than a convention:

```scheme
(/ 1 'x)             ;; type error in '/':         expected number, got #<symbol>
(gcd 1 'x)           ;; 'gcd'
(lcm 1 'x)           ;; 'lcm'
(numerator 'x)       ;; 'numerator'
(square 'x)          ;; 'square'
(floor 'x)           ;; 'floor'
(truncate 'x)        ;; 'truncate'
```

`+` and `/` are adjacent entries in the same `specs` table and disagree.

`arithmetic` is at least not a bound name, unlike `string` — but `+` is by a wide margin the most-called procedure in the language, and its type error names nothing the user wrote.

## Full scope of this issue

| Helper | Placeholder reported | Procedures | File |
|---|---|--:|---|
| `parseStartEnd` (`:137`) | `string` (a real, working procedure) | 20 | `primitives_string_ext.zig` |
| `numberTypeError` (`:40`) + 2 siblings | `arithmetic` | 11 | `primitives_arithmetic.zig`, `primitives_numeric.zig` |
| char-set predicate check (`:127`) | `string operation` | — | `primitives_string_ext.zig` |

**31 procedures** in total misreport which procedure failed.

## Fix, unchanged in shape

Thread a `proc: []const u8` parameter through each helper from its call sites. `numberTypeError(v)` becomes `numberTypeError(proc, v)`; `parseStartEnd(data, args, i)` becomes `parseStartEnd(proc, data, args, i)`.

**Verified:** fresh ReleaseSafe build at `d0d03deb` (macOS aarch64, isolated `KAAPPI_HOME`); all 11 misreports and all 7 controls reproduced from that binary.

baijum:
## What

Error-taxonomy sweep fixing three coupled diagnostics issues. All three share two helpers (`primitives.parseOptionalRange`, `primitives_string_ext.parseStartEnd`) and one audit file, so they land as one PR.

### #2020 — bounds failures reported KP3002 (33 sites)

`parseOptionalRange` (behind 20 procedures) reported an out-of-range `[start end]` through `typeError`, so `(substring "ab" 0 99)` was KP3006 while `(string-copy "ab" 0 99)` — the same operation per R7RS 6.7 — was KP3002. The three range branches now use `indexError` (KP3006), and `start > end` uses `argError` (KP3007, the case `kaappi explain KP3007` names verbatim). Direct sites likewise fixed: `bytevector-u8-ref`/`-set!`, `bytevector-copy!`, `string->utf8`, `vector-copy!`, `vector-swap!`, `vector-reverse-copy!`, `vector-unfold!(-right)!`, `vector-append-subvectors`, `string-copy(!)`/`string->list`/`string->vector` UTF-8 offset conversions, `string-replace`, and the list-walk family (`list-tail`, `take`, `drop`, `take-right`, `drop-right` report KP3006 when k walks off a proper list, like `list-ref`; a non-pair element is still KP3002). The u8/s8 SRFI 160 seam now agrees across all element kinds.

### #2021 — well-typed values rejected as KP3002 (72 sites)

The dominant shape was a type check and a range check sharing one message string: `(bytevector 256)` said "expected exact integer 0-255" about an exact integer. Each conflated branch is split — the type branch keeps `typeError`, the domain branch becomes `argError` naming the value. Covers byte ranges (0-255), negative lengths (`make-vector`/`-string`/`-bytevector`/`-list`/`-s8vector`), the `integer->char` Unicode domain, enumerated rejections (`number->string` radix, `null-environment` version, `hash` bound, `s8vector` element range, `transcoded-port` codec/eol-style/error-mode), immutability ("got #<pair>" beside "expected mutable pair" argued against itself), absent `hash-table-ref` keys, FFI name-length/param-count guards, `make-time` nanosecond range, random bounds/seeds, `thread-start!` on a started thread, and `exact` on inf/nan.

### #2022 — 31 procedures misreported which procedure failed

Shared helpers hardcoded a placeholder where the caller's name belongs. `parseStartEnd` and `callPredOrCharset` (`primitives_string_ext.zig`) now take the procedure name from each of their 23/13 call sites — 20 SRFI-13 procedures used to blame the real, unrelated, working procedure `string`, and the predicate check blamed `string operation`. `numberTypeError`/`ratPartsVal`/`complexPartsOf`/`cmpPair`/`toF64Ext` (`primitives_arithmetic.zig`, `primitives_numeric.zig`) thread the real name from every call site, so `+ - * / < > <= >= = max min abs quotient remainder modulo gcd lcm expt sqrt sin cos tan asin acos exp log magnitude angle numerator denominator` name themselves — `+` now agrees with its adjacent specs-table neighbour `/`.

Also: the 18 bare `PrimitiveError.IndexOutOfBounds`/`InvalidArgument` returns in `primitives_string_ext.zig` are now `indexError`/`argError` calls — the code was right but no detail was set, so the message lost the index and length.

## ⚠ BASELINE change (CI ratchet)

`.github/workflows/ci.yml` bare error-taxonomy gate: **BASELINE lowered 28 → 10**. All 18 `primitives_string_ext.zig` sites are cleared; the remaining 10 are VM-infrastructure guards (`vm_calls`, `vm_dispatch_helpers`, `vm_continuations`, `fiber_wait`) plus two `primitives_io`/`primitives_fiber` guards, outside this PR's scope. Verified with the gate's exact grep from the worktree root.

## Tests

- `tests/scheme/audit/error-taxonomy-audit.scm` rewritten: every previously-disabled `;; FAIL:` assertion is live, the `TODAY` pins of the buggy behaviour are removed, Part 6 re-pins the now-detailed `string-take` messages, and a self-naming sweep asserts all 20 SRFI-13 procedures plus the arithmetic operators name themselves (the recurrence guard #2022 asked for). **164 assertions, 0 failures.**
- Regression proof: the pre-fix audit file (from `HEAD~1`) run against the fixed binary fails exactly its 31 `TODAY` pins — the assertions are real regression tests, and the fix is a visible change.
- Pinned wordings updated where messages legitimately changed: `internal-primitives-audit.scm` (SRFI 160 element-range wording), `primitives_srfi181-audit.scm` (transcoded-port codec symbols); two stale prose comments corrected (`sbc-literal-immutability.scm`, `primitives_srfi160-audit.scm`).

## Evidence

- `zig build test`: **1744 pass, 7 skip** (1751 total)
- `bash tests/scheme/run-all.sh`: **717 pass / 0 fail**, R7RS suite **1395 pass / 0 fail**
- `bash tools/run-r7rs-suite.sh`: 1395 pass, 0 fail
- Error-taxonomy audit: 164 expected passes, 0 failures
- Spot checks: `(string-copy "ab" 0 99)` → `KP3006 string-copy: index 99 out of range for length 2`; `(string-copy "abc" 2 1)` → `KP3007 string-copy: start 2 is greater than end 1`; `(bytevector 256)` → `KP3007 bytevector: byte 256 is outside 0-255` while `(bytevector 'x)` stays KP3002; `(+ 1 'x)` → `type error in '+': expected number, got x`; `(string-index "ab" #\a 0 9)` names `string-index`.

Closes #2020
Closes #2021
Closes #2022

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,12 +144,16 @@ jobs:
       # is clean; a backlog of bare IndexOutOfBounds/InvalidArgument in the
       # primitives (kaappi#2020/#2021/#2022) is not, so this is once again a
       # ratchet, exactly as it was for TypeError (91 -> 20 -> 0, cleared by
-      # kaappi#1868). BASELINE may only go DOWN: lower it whenever you remove a
+      # kaappi#1868). The #2020/#2021/#2022 sweep cleared all 18
+      # primitives_string_ext.zig sites (28 -> 10); what remains is
+      # VM-infrastructure guards in vm_calls/vm_dispatch_helpers/
+      # vm_continuations/fiber_wait plus two primitives_io/primitives_fiber
+      # guards. BASELINE may only go DOWN: lower it whenever you remove a
       # bare return, and delete it when it reaches 0. Test files legitimately
       # write `error.TypeError` in `expectError`, so src/tests_*.zig is excluded.
       - name: Check bare error-taxonomy regression
         run: |
-          BASELINE=27
+          BASELINE=9
           hits=$(grep -rnE 'return (PrimitiveError|VMError|error)\.(TypeError|IndexOutOfBounds|InvalidArgument)' src/*.zig \
                  | grep -v '^src/tests_' | grep -v 'bare-ok' || true)
           count=$(printf '%s' "$hits" | grep -c . || true)
--- a/src/primitives.zig
+++ b/src/primitives.zig
@@ -600,16 +600,16 @@ pub fn parseOptionalRange(args: []const Value, arg_offset: usize, max_len: usize
     if (args.len > arg_offset) {
         if (!types.isFixnum(args[arg_offset])) return typeError(proc_name, "exact integer", args[arg_offset]);
         const s = types.toFixnum(args[arg_offset]);
-        if (!fixnumIndexInBoundsInclusive(s, max_len)) return typeError(proc_name, "valid index", args[arg_offset]);
+        if (!fixnumIndexInBoundsInclusive(s, max_len)) return indexError(proc_name, s, max_len);
         start = @intCast(s);
     }
     if (args.len > arg_offset + 1) {
         if (!types.isFixnum(args[arg_offset + 1])) return typeError(proc_name, "exact integer", args[arg_offset + 1]);
         const e = types.toFixnum(args[arg_offset + 1]);
-        if (!fixnumIndexInBoundsInclusive(e, max_len)) return typeError(proc_name, "valid index", args[arg_offset + 1]);
+        if (!fixnumIndexInBoundsInclusive(e, max_len)) return indexError(proc_name, e, max_len);
         end = @intCast(e);
     }
-    if (start > end) return typeError(proc_name, "start <= end", args[arg_offset]);
+    if (start > end) return argError(proc_name, "start {d} is greater than end {d}", .{ start, end });
     return .{ .start = start, .end = end };
 }
 
@@ -823,7 +823,7 @@ fn cdr(args: []const Value) PrimitiveError!Value {
 
 fn setCar(args: []const Value) PrimitiveError!Value {
     if (!types.isPair(args[0])) return typeError("set-car!", "pair", args[0]);
-    if (types.toObject(args[0]).flags.immutable) return typeError("set-car!", "mutable pair", args[0]);
+    if (types.toObject(args[0]).flags.immutable) return argError("set-car!", "cannot mutate an immutable pair", .{});
     // #1924: reject before the store — a shared parent-heap pair must not
     // come to hold a child-heap object.
     if (memory.crossHeapStoreViolation(types.toObject(args[0]), args[1])) return raiseCrossHeapStore("set-car!");
@@ -834,7 +834,7 @@ fn setCar(args: []const Value) PrimitiveError!Value {
 
 fn setCdr(args: []const Value) PrimitiveError!Value {
     if (!types.isPair(args[0])) return typeError("set-cdr!", "pair", args[0]);
-    if (types.toObject(args[0]).flags.immutable) return typeError("set-cdr!", "mutable pair", args[0]);
+    if (types.toObject(args[0]).flags.immutable) return argError("set-cdr!", "cannot mutate an immutable pair", .{});
     if (memory.crossHeapStoreViolation(types.toObject(args[0]), args[1])) return raiseCrossHeapStore("set-cdr!");
     if (memory.gc_instance) |gc| gc.writeBarrier(types.toObject(args[0]), args[1]);
     types.setCdr(args[0], args[1]);
--- a/src/primitives_arithmetic.zig
+++ b/src/primitives_arithmetic.zig
@@ -18,7 +18,7 @@ const isZeroValue = numeric.isZeroValue;
 
 /// Split a number into (real, imag) component Values. A non-complex number
 /// is (v, fixnum 0). Components are never complex themselves (kaappi#2166).
-fn complexPartsOf(v: Value) PrimitiveError!struct { real: Value, imag: Value } {
+fn complexPartsOf(proc: []const u8, v: Value) PrimitiveError!struct { real: Value, imag: Value } {
     if (types.isComplex(v)) {
         const c = types.toComplex(v);
         return .{ .real = c.real, .imag = c.imag };
@@ -26,7 +26,7 @@ fn complexPartsOf(v: Value) PrimitiveError!struct { real: Value, imag: Value } {
     if (types.isFixnum(v) or types.isBignum(v) or types.isRationalObj(v) or types.isFlonum(v)) {
         return .{ .real = v, .imag = types.makeFixnum(0) };
     }
-    return numberTypeError(v);
+    return numberTypeError(proc, v);
 }
 
 /// Componentwise scalar arithmetic over the real tower. Each operand is a
@@ -76,7 +76,7 @@ pub fn makeFixnumCheckedGc(gc: *@import("memory.zig").GC, n: i64) PrimitiveError
 }
 
 /// Extended toF64 that also handles bignums and rationals.
-pub fn toF64Ext(v: Value) PrimitiveError!f64 {
+pub fn toF64Ext(proc: []const u8, v: Value) PrimitiveError!f64 {
     if (types.isFixnum(v)) return @floatFromInt(types.toFixnum(v));
     if (types.isFlonum(v)) return types.toFlonum(v);
     if (types.isBignum(v)) return bignum_mod.toF64(v);
@@ -84,11 +84,11 @@ pub fn toF64Ext(v: Value) PrimitiveError!f64 {
         const r = types.toRational(v);
         return types.rationalToF64(r.numerator, r.denominator);
     }
-    return numberTypeError(v);
+    return numberTypeError(proc, v);
 }
 
-fn numberTypeError(v: Value) PrimitiveError {
-    return primitives.typeError("arithmetic", "number", v);
+fn numberTypeError(proc: []const u8, v: Value) PrimitiveError {
+    return primitives.typeError(proc, "number", v);
 }
 
 // ---------------------------------------------------------------------------
@@ -118,13 +118,13 @@ fn toRationalParts(v: Value) ?RatParts {
 
 const RatPartsVal = struct { num: Value, den: Value };
 
-fn ratPartsVal(v: Value) PrimitiveError!RatPartsVal {
+fn ratPartsVal(proc: []const u8, v: Value) PrimitiveError!RatPartsVal {
     if (types.isFixnum(v) or types.isBignum(v)) return .{ .num = v, .den = types.makeFixnum(1) };
     if (types.isRationalObj(v)) {
         const r = types.toRational(v);
         return .{ .num = r.numerator, .den = r.denominator };
     }
-    return primitives.typeError("arithmetic", "rational number", v);
+    return primitives.typeError(proc, "rational number", v);
 }
 
 fn allocRationalRooted(gc: *@import("memory.zig").GC, n: i64, d: i64) PrimitiveError!Value {
@@ -358,7 +358,7 @@ fn add(args: []const Value) PrimitiveError!Value {
         var slot_i = gc.rootedSlot(acc_imag) catch return PrimitiveError.OutOfMemory;
         defer slot_i.release();
         for (args) |a| {
-            const parts = try complexPartsOf(a);
+            const parts = try complexPartsOf("+", a);
             const nr = try add2(slot_r.get(), parts.real);
             slot_r.set(nr);
             const ni = try add2(slot_i.get(), parts.imag);
@@ -369,7 +369,7 @@ fn add(args: []const Value) PrimitiveError!Value {
     if (anyFlonum(args)) {
         var sum: f64 = 0;
         for (args) |a| {
-            sum += try toF64Ext(a);
+            sum += try toF64Ext("+", a);
         }
         return makeFlonumVal(sum);
     }
@@ -387,10 +387,10 @@ fn add(args: []const Value) PrimitiveError!Value {
         defer slot_t2.release();
         for (args) |a| {
             if (types.isFlonum(a)) {
-                const acc_f = try toF64Ext(acc_num) / try toF64Ext(acc_den);
+                const acc_f = try toF64Ext("+", acc_num) / try toF64Ext("+", acc_den);
                 return makeFlonumVal(acc_f + types.toFlonum(a));
             }
-            const parts = try ratPartsVal(a);
+            const parts = try ratPartsVal("+", a);
             const t1 = bignum_mod.mul(gc, acc_num, parts.den) catch return PrimitiveError.OutOfMemory;
             slot_t1.set(t1);
             const t2 = bignum_mod.mul(gc, parts.num, acc_den) catch return PrimitiveError.OutOfMemory;
@@ -406,7 +406,7 @@ fn add(args: []const Value) PrimitiveError!Value {
     // Fixnum path with overflow detection
     var sum: i64 = 0;
     for (args) |a| {
-        if (!types.isFixnum(a)) return numberTypeError(a);
+        if (!types.isFixnum(a)) return numberTypeError("+", a);
         const r = @addWithOverflow(sum, types.toFixnum(a));
         if (r[1] != 0) return bignumAddAll(args);
         sum = r[0];
@@ -418,7 +418,7 @@ fn sub(args: []const Value) PrimitiveError!Value {
     std.debug.assert(args.len > 0);
     if (isAnyComplex(args)) {
         const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
-        const first = try complexPartsOf(args[0]);
+        const first = try complexPartsOf("-", args[0]);
         const acc_real = first.real;
         const acc_imag = first.imag;
         var slot_r = gc.rootedSlot(acc_real) catch return PrimitiveError.OutOfMemory;
@@ -436,7 +436,7 @@ fn sub(args: []const Value) PrimitiveError!Value {
             slot_i.set(ni);
         } else {
             for (args[1..]) |a| {
-                const parts = try complexPartsOf(a);
+                const parts = try complexPartsOf("-", a);
                 const nr = try sub2(slot_r.get(), parts.real);
                 slot_r.set(nr);
                 const ni = try sub2(slot_i.get(), parts.imag);
@@ -446,16 +446,16 @@ fn sub(args: []const Value) PrimitiveError!Value {
         return makeComplexOrRealV(gc, slot_r.get(), slot_i.get());
     }
     if (anyFlonum(args)) {
-        if (args.len == 1) return makeFlonumVal(-(try toF64Ext(args[0])));
-        var result = try toF64Ext(args[0]);
+        if (args.len == 1) return makeFlonumVal(-(try toF64Ext("-", args[0])));
+        var result = try toF64Ext("-", args[0]);
         for (args[1..]) |a| {
-            result -= try toF64Ext(a);
+            result -= try toF64Ext("-", a);
         }
         return makeFlonumVal(result);
     }
     if (anyRational(args) or (anyBignum(args) and !anyFlonum(args))) {
         const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
-        const init = try ratPartsVal(args[0]);
+        const init = try ratPartsVal("-", args[0]);
         var acc_num: Value = init.num;
         var acc_den: Value = init.den;
         if (args.len == 1) {
@@ -472,10 +472,10 @@ fn sub(args: []const Value) PrimitiveError!Value {
         defer slot_t2.release();
         for (args[1..]) |a| {
             if (types.isFlonum(a)) {
-                const acc_f = try toF64Ext(acc_num) / try toF64Ext(acc_den);
+                const acc_f = try toF64Ext("-", acc_num) / try toF64Ext("-", acc_den);
                 return makeFlonumVal(acc_f - types.toFlonum(a));
             }
-            const parts = try ratPartsVal(a);
+            const parts = try ratPartsVal("-", a);
             const t1 = bignum_mod.mul(gc, acc_num, parts.den) catch return PrimitiveError.OutOfMemory;
             slot_t1.set(t1);
             const t2 = bignum_mod.mul(gc, parts.num, acc_den) catch return PrimitiveError.OutOfMemory;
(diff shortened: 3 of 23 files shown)


baijum:
**Review (Claude Code, read-only + static verification)** — ✅ approve

A large but uniform, mechanical sweep. The three shared helpers at the root of the issues (`parseOptionalRange`, `parseStartEnd`/`callPredOrCharset`, `numberTypeError` family) are fixed at the source and threaded consistently to every call site; the previously-disabled `;; FAIL:` assertions in the audit files are flipped live and would fail pre-fix. Coverage of #2020 (bounds→KP3006), #2021 (value rejection→KP3007), #2022 (real proc name) all check out.

- **Ratchet:** `BASELINE` lowered 28→10; the 10 remaining bare returns are all VM-infra/io/fiber guards (vm_calls, vm_dispatch_helpers, vm_continuations, fiber_wait, primitives_io, primitives_fiber) — matches the updated comment.
- ⚠️ **Cross-PR merge conflict with #2342:** both PRs rewrite the same `BASELINE=` line in `.github/workflows/ci.yml` (this one →10, #2342 →27). They will conflict. Merge one, then rebase the other and recompute the real count against the merged tree — the fiber bare-return #2342 removes is inside this PR's remaining 10, so the true post-merge baseline is ~9, not 10 or 27.

Nits (non-blocking): `primitives_numeric.zig`/`primitives_srfi1.zig`/`primitives_srfi18.zig` are >1500 lines but were already over before this PR (near-zero net delta). A negative `at` to `string-copy!`/`vector-copy!` now reports KP3006 (index) rather than KP3007 — defensible judgment call.

_Not built locally (shared cache); signature-consistency and the ratchet count were verified by grep._

PR #2348, opened 2026-08-25
SRFI-254: guardian weak resurrection and transport-cell weak keys (#2011, #2006)

### LINKED ISSUE #2006: [SRFI-254] Transport cell keys are held strongly: cells never break and every registration is permanent
**Category:** SRFI / GC · **Severity:** wrong-result + unbounded retention · **Found by:** Systematic audit v2, Phase 2.10 (#1890)

SRFI 254 specifies a transport cell's key field as **weakly holding**, and specifies that the cell **breaks** when its key is reclaimed:

> "A transport cell is a record structure with two fields called the key and value fields. **The content of the key field is stored in a weakly holding location.**"

> "Transport cells can be broken. If the content of its key field denotes a location or a sequence of locations, **the transport cell is broken in the course of the computation when the garbage collector reclaims the location** or sequence of locations."

Kaappi holds both strongly and never breaks a cell. `gc_collect.markGuardianStrong`'s `is_transport` arm marks every registered cell during the **strong** mark phase, and `markObjectContents`' `.transport_cell` arm then marks the cell's `key` and `value`; `processWeakRefs` skips transport cell guardians entirely.

## What this costs

`(tg key value)` pins `key` and `value` for the lifetime of the guardian, and there is no way to get them back — `(tg)` correctly always returns `#f` on a non-moving collector, so a registered cell is **never removed from `registered`**. Every registration is permanent. That is the opposite of what the SRFI names as the feature's purpose: *"Use cases for transport cell guardians are efficient portable implementations of `eq?`-based hashing data structures."* A weak `eq?` hash table built on this retains every key ever inserted.

The retention is also **observable from Scheme**, because it defeats an unrelated object guardian:

## Minimal reproduction

```scheme
(import (scheme base) (scheme write) (srfi 254))
(define (churn!) (do ((i 0 (+ i 1))) ((= i 20000)) (list i i)))

(define g (make-guardian))
(define tg (make-transport-cell-guardian))
(let ((o (list 'both))) (g o) (tg o 'v))   ; same object: guarded AND a tcg key
(churn!) (churn!)
(write (g))    ;; expected: (both)   actual: #f
```

## Discriminating control

```scheme
(define g2 (make-guardian))
(let ((o (list 'only-g))) (g2 o))          ; identical, minus the tcg registration
(churn!) (churn!)
(write (g2))   ;; => (only-g)
```

The only difference is the `(tg o 'v)` call, so the transport cell is what pins `o`. Per the spec the tcg key is weakly holding and must not affect `g` at all.

Two further consequences, both directly assertable:

| | expected | actual |
|---|---|---|
| `(transport-cell-broken? c)` after the key becomes unreachable | `#t` | `#f`, always |
| `(transport-cell-key c)` after that | `#f` | the key, verbatim |

## Doc-truth component

`CLAUDE.md`, `src/primitives_srfi254.zig`'s header and `src/types_weakrefs.zig`'s `Guardian`/`TransportCell` doc comments all describe transport cell guardians as "degenerate: keys never move, so `(tg)` always returns `#f`". That part is correct and conformant. What none of them says is that the cells are *also* pinned forever and never break — which is the part that is not conformant. Whatever the fix, the docs need the second half.

**Verified:** fresh ReleaseSafe build at `e6081c3b` (v0.22.1), isolated `KAAPPI_HOME`; behaviour is **unchanged** under a separate `zig build -Dgc-stress=true` build — the retention is structural, not a collection-timing artefact.
**Source location:** `src/gc_collect.zig` — `markGuardianStrong` (`is_transport` arm) and `markObjectContents`' `.transport_cell` arm; `processWeakRefs` has no transport-cell phase at all.
**Disabled tests:** `tests/scheme/audit/primitives_srfi254-audit.scm`, section 12 (two `;; FAIL:` blocks); current behaviour is pinned by three enabled assertions beside them, so a fix flips visible tests.

### LINKED ISSUE #2011: [SRFI-254] A second guardian watching the same object never resurrects it — the first guardian's ready queue is marked strongly
**Category:** SRFI / GC · **Severity:** wrong-result (a finalizer silently never fires) · **Found by:** Systematic audit v2, Phase 2.10 (#1890)

SRFI 254's resurrection rule makes the *ready queues of every guardian in the system* part of the hypothetical, not part of the live set:

> "If the content of a guarded element denotes a location or a sequence of locations, the element is resurrected in the course of the computation when the garbage collector would reclaim the location or sequence of locations **if all locations denoted by the fields of guarded elements or by resurrected elements of all guardians were weakly resurrected.**"

So an object registered with two independent guardians must be resurrected by *both*. In Kaappi the first guardian to be processed wins and the second starves for as long as the first holds it — permanently, if the first is never drained.

## Minimal reproduction

```scheme
(import (scheme base) (scheme write) (srfi 254))
(define (churn!) (do ((i 0 (+ i 1))) ((= i 20000)) (list i i)))
(define ga (make-guardian))
(define gb (make-guardian))
(let ((o (list 'shared))) (ga o) (gb o))
(do ((i 0 (+ i 1))) ((= i 30)) (churn!))
(write (gb))              ;; expected: (shared)   actual: #f
(display "-- drain ga --") (newline)
(write (ga))              ;; => (shared)
(do ((i 0 (+ i 1))) ((= i 3)) (churn!))
(write (gb))              ;; => (shared)  — only now
```

Thirty collections do not help. The instant `ga`'s ready queue is drained, `gb` fires.

## Discriminating controls

1. **Single guardian**, same object, same churn: `(gc2)` ⟹ `(solo)` immediately. So the object *is* unreachable; the second guardian is what fails.
2. **Draining the winner** unblocks the loser, which identifies the blocker as the ready queue rather than the registration itself.
3. Which of `ga`/`gb` wins is mark-order dependent, so the assertion in the suite counts how many of the two fire (currently `1`, spec says `2`) rather than naming one.

## Mechanism

Two sites in `src/gc_collect.zig` conspire, both marking the watched object **strongly** where the spec's hypothetical requires it to be weak:

1. `markGuardianStrong` runs in the strong mark phase and does `for (g.ready.items) |e| { markValue(gc, e.watched); markValue(gc, e.payload); }`. By the time `processWeakRefs` runs, every other guardian's `weakReachable(gc, e.watched)` probe already answers `true`.
2. Within a single `processWeakRefs` pass, the resurrect branch itself does `markValue(gc, e.watched)` before appending to `ready`, so a *later* entry watching the same object in the same pass also sees it reachable.

Site 2 alone produces the same defect inside one guardian: registering one object twice in one guardian resurrects one element per collection round rather than both.

## Why it matters

This is the SRFI's own headline use case — *"management of externally allocated resources that eventually need to be freed"*. Two independent libraries each guarding the same handle is exactly the shape that breaks: one library's cleanup runs, the other's never does, with no error anywhere. The failure is silent and permanent, not merely delayed, whenever the winning guardian's queue is drained lazily or not at all.

**Verified:** fresh ReleaseSafe build at `e6081c3b` (v0.22.1), isolated `KAAPPI_HOME`, 3/3 runs; **identical under a separate `zig build -Dgc-stress=true` build** — collection frequency does not change the outcome, which rules out a timing artefact.
**Source location:** `src/gc_collect.zig` — `markGuardianStrong` (the `ready` loop) and `processWeakRefs` (the resurrect branch's `markValue(gc, e.watched)`).
**Disabled test:** `tests/scheme/audit/primitives_srfi254-audit.scm`, section 11 (`;; FAIL:` block); the current one-of-two behaviour is pinned enabled beside it, so a fix flips a visible test.

baijum:
## What

Two SRFI-254 collector conformance fixes in `processWeakRefs`, one PR because they touch the same weak-resolution machinery and the same test file.

### #2011 — a second guardian watching the same object never resurrects it

The spec's resurrection hypothetical makes the ready queues of *every* guardian part of the hypothetical, not part of the live set:

> "the element is resurrected in the course of the computation when the garbage collector would reclaim the location or sequence of locations if all locations denoted by the fields of guarded elements or by resurrected elements of all guardians were **weakly resurrected**."

Two sites marked the watched object **strongly** where the hypothetical requires weak:

1. `markGuardianStrong` marked every ready-queue element during the strong mark phase, so every other guardian's `weakReachable` probe already answered *true* — cross-collection starvation (the issue's 30-churn-round repro).
2. The resurrect branch inside `processWeakRefs` marked the watched object on the spot, so a later entry in the same pass saw it reachable — one resurrection per collection round per guardian even for duplicate registrations in a *single* guardian.

**Fix.** `processWeakRefs` now implements "weakly resurrected = kept alive without being reachable" directly:

- **Frozen probes**: each round partitions every guardian's `registered` entries (probe sub-pass) *before* any element is resurrected (apply sub-pass), so N guardians — or N registrations in one guardian — watching the same object all fire in the same collection.
- **`weak_resurrected` set** (new, per collection): ready-queue contents, freshly resurrected elements, and retained representatives are recorded in the set instead of being marked. A new *settle* pass materializes them into real marks once every weak decision of a round is made (with a `settled` flag so marking-only discoveries get one more resolution round and the loop still terminates).
- **Two probes by design**: guardian probes use plain `weakReachable` (marked bits only — never sees a weak resurrection); ephemeron and transport-cell keys use the new `keptAlive` (marked ∨ weakly resurrected), preserving the spec's kept-alive reading — a key held only by a guardian's ready queue neither breaks its ephemeron nor breaks its cell.

The ephemerons-before-guardians fixpoint and its pinned interactions are unchanged: an ephemeron keyed on a resurrected object still survives, live chains still survive, dead chains still break.

### #2006 — transport cell keys are held strongly; cells never break

The spec:

> "The content of the key field is stored in a **weakly holding location**." / "the transport cell is broken in the course of the computation when the garbage collector reclaims the location or sequence of locations."

Kaappi marked each registered cell **and its key** strongly (`markGuardianStrong`'s `is_transport` arm + the `.transport_cell` mark arms), so no cell ever broke and every registration pinned its key forever — defeating the SRFI's named use case (weak `eq?` hash tables) and observably blocking an unrelated object guardian watching the same object.

**Fix.** The `.transport_cell` arms of `markObjectContents`/`markValueInner` now mark only the **value** (strong, per "Except as noted, all newly chosen locations are strongly holding") and defer the key to a new `pending_transport_cells` list; after the fixpoint, every cell whose key is neither reachable nor kept alive breaks: `transport-cell-broken?` → `#t`, `transport-cell-key` → `#f` (already implemented in the primitive via the `broken` flag), value intact. A weak key no longer blocks an object guardian. Registration stays permanent — the non-moving collector never transports a cell, so `(tg)` always returns `#f`; that half of the degeneracy is conformant and unchanged. Doc-truth updated (`CONFORMANCE.md`, SRFI notes, type/primitive header comments) to say both halves.

## Tests

- `tests/scheme/audit/primitives_srfi254-audit.scm`: the pinned one-of-two and never-breaks assertions are flipped to the spec answers; new coverage for same-guardian double registration (site 2), both-fire-and-re-arm across a second cycle, cross-cycle ready-queue hold vs a fresh registration, and stays-broken. The "cell alongside an object guardian is intact" assertion moved before the `(g-vs-tcg)` drain — after the drain the key is genuinely dead and a collection may rightly break the cell (only observable under `-Dgc-stress=true`).
- `src/tests_srfi254.zig`: 5 new/rewritten deterministic GC-semantics tests (bare GC, `enabled = false`, explicit `collect()`).
- `src/tests_gc_tracing.zig`: transport-cell tracing tests rewritten for the weak key (value traced, key dies and the cell breaks — including through the remembered-set walk); the ready-elements test now documents that liveness comes from the settle pass.

Every new/flipped test fails on the unfixed collector (verified by reverting the source changes and rebuilding): 9 audit assertions + 8 Zig tests.

## Evidence

- Audit: 189/189 on ReleaseSafe; 189/189 three times under a `-Dgc-stress=true` build.
- `zig build test` green (also green before the wrap-up rerun).
- R7RS suite 1395/1395 under the gc-stress build; `srfi-254.scm`, `srfi18-child-registers-1933.scm`, `srfi18-sharing-model.scm`, `cross-thread-boundary-audit.scm` green.
- Native tier: both repro shapes compiled with `kaappi compile` (after `zig build lib`) pass.
- Both issues' minimal reproductions now print the spec answers (`(gb)` → `(shared)` immediately; orphan cell broken with `#f` key and intact value).

Closes #2011
Closes #2006


baijum:
**Review (Claude Code, built + tested locally)** — ✅ approve

High-risk GC-internals, and it holds up. I built this branch and ran the guardian/tracing unit tests (`zig build test -Dtest-filter=srfi254 -Dtest-filter=gc_tracing`) — **green**.

Design is sound: the old code marked guardian ready-queue elements *strongly* and resurrected inline, so a second watcher saw "reachable" and starved (#2011). The new flow probes every guardian's elements against the *frozen* mark state (`weakReachable`, marked-bit only), records resurrect/ready/retain in a new `weak_resurrected` set (never marked inline), then a settle pass `markValue`s them after every weak decision — before sweep. The deliberate split (guardians probe `weakReachable`; ephemeron + transport-cell keys probe `keptAlive` = marked **or** resurrected) is what preserves the spec's kept-alive-vs-reachable distinction. #2006: transport-cell key is now weak, cell breaks with `key=FALSE` (no dangling Value), value stays strong. Minor/full-collection liveness, ordering, and gc-stress re-entrancy (uses the raw backing allocator, no nested `maybeCollect`) all check out.

- One documented narrow deviation: a guardian *discovered during* settle marking probes next round, so it keeps-rather-than-fires for an already-resurrected object. Clearly commented; narrows rather than widens the old bounded delay.
- `tests_gc_tracing.zig` is 2017 lines (>1500) — pre-existing breadth, independent tests, not this PR's doing.

_Recommend the gc-stress CI leg (required check) be confirmed green before merge — it's the one that would expose any settle-pass rooting/ordering mistake._

PR #2349, opened 2026-08-25
wasi: real compile gate for the wasm32-wasi unit suite, executed under wasmtime (kaappi#2153)

### LINKED ISSUE #2153: [Reactor] the WASI poll_oneoff backend's fd path is executed by no test, and the unit suite does not compile for wasm32-wasi

baijum:
## Direction chosen

The issue's **(a)** — fix the compile errors so `zig build test -Dtarget=wasm32-wasi` becomes a compile gate like Windows and run the produced binary under wasmtime in the existing `wasm` job — with **(b)**'s documentation for the one piece no runtime can host: the fd-readiness criterion.

I probed the boundary empirically on wasmtime 48 (the CI runner is v46.0.1, same p1 implementation): `sock_open` is an unresolvable import even with `-S inherit-network`/`-S tcp`, the legacy p1 (`-S preview2=n`) is gone, and `poll_oneoff` with an FD subscription fails the whole call with `BADF` on every fd a guest *can* obtain (closed fd, stdout pipe end, preopened dir). So a WASI guest cannot construct any EAGAIN-capable fd under the CI runtime — the fd suites cannot execute there at any cost, which is exactly what (b)'s Stage 3 note is for. Everything else the issue asks for is delivered as execution, not documentation.

## What changed

**The compile gate (the headline).** `zig build test -Dtarget=wasm32-wasi` exits 0 and installs `zig-out/bin/unit-tests.wasm`:
- WASI arms across the platform facade: write-open family + `openNullSink` (`path_open`), `DirIter` (`fd_readdir`), the `dl*` family, `argsIterate`; process-spawn sites (`thottam_proc`, `test_runner`, `test_selection`, `doctor`, `native_compiler`) return their spawn errors comptime-gated so the module never imports wasi-libc's missing `pipe`/`fork`/`execve` symbols; `testing_helpers`' fd-pair family grows honest WASI arms; FFI test constants made 32-bit-safe.
- `build.zig`: the wasm test module is single-threaded (matching the wasm executable — std's Io/futex plumbing otherwise emits `memory.atomic.wait` inline asm), carries the atomics CPU feature (only waits need shared memory; this module never waits), defaults to ReleaseSmall (Debug exceeds wasmtime's per-function locals limit in the comptime-generated FFI dispatchers; ReleaseSafe crashes the LLVM wasm32 backend selecting a float constant pool — explicit `-Doptimize` is always respected), and installs the test binary for a wasm runtime.

**Execution under wasmtime.** `wasmtime run --dir=. --dir=/tmp zig-out/bin/unit-tests.wasm` → **1542 passed, 209 skipped, 0 failed, of 1751 — the same test count as the native suite**. This genuinely executes `WasiPollBackend`'s CLOCK path (`addTimer`, `removeTimer`, `popExpiredTimers`, `msFromNs`, the scheduler/fiber halves), which previously ran only in `tests/wasm/timers.scm` via the CLI.

**The fd boundary, documented.** fd-pair/thread/ffi/Scheme-file-I/O tests skip on wasm with per-test comptime gates naming their reason; `testing_helpers.wasmNoFdPairs` panics if an fd test reaches a pair constructor without its skip (a missing gate is loud, never silent). `docs/dev/porting.md` Stage 3 states the boundary and what IS required of a WASI leg, replacing the self-contradiction the issue called out.

**A real bug the gate found immediately** (the issue's own 32-bit-`usize` class, sibling of #1912): `(make-bytevector 100000000000000)` and `(make-string ...)` with absurd lengths **silently allocated a truncated, far smaller object** on wasm32 — the i64 length truncated inside `@intCast` before the GC payload cap could see it (verified on the shipped `kaappi.wasm`: the call returned normally). `primitives.fixnumFitsUsize` now compares in u64 before narrowing at the three `make-*` sites and raises the same `OutOfMemory` the 64-bit payload cap would, so `guard` sees one cross-platform condition. The pre-existing absurd-payload regression test now *runs on wasm and fails without the fix* — it could never run before this PR.

**Also:**
- `platform.zig`'s WASI opens now resolve through the **preopen table** the way wasi-libc does (relative → CWD preopen, absolute → longest preopen-name prefix match) instead of hardcoding fd 3 — the unit tests' `/tmp/...` paths resolve under `--dir=/tmp`, and Zig-level reads keep the #2108 relative convention.
- Three assertions that assumed "unit tests never build for WASM" (`kaappi-threads`, `(library (srfi 18))`, `features`' `sandbox_available`) now expect each platform's correct answer instead of an unconditional native one.
- CI: the `wasm` job gains `WASI unit tests (compile gate + run under wasmtime)` — the runner was already installed there. The bare-error BASELINE line is untouched.

## Verification

- `zig build test -Dtarget=wasm32-wasi` → exit 0 (compile gate; run auto-skips as foreign), installs `unit-tests.wasm`.
- `wasmtime run --dir=. --dir=/tmp zig-out/bin/unit-tests.wasm` → exit 0, **1542 passed; 209 skipped; 0 failed** (wasmtime 48.0.0, macOS aarch64 host).
- Native `zig build test` → exit 0, **Build Summary: 11/11 steps succeeded; 1822/1829 tests passed (7 skipped)** (unit-tests 1744 pass/7 skip, thottam-tests 78 pass).
- `zig build wasm` (shipped binary) → exit 0; its smoke surface unchanged.
- `zig fmt --check` clean on every touched file; `ci.yml` parses; skips are visible in the wasm run output (`...SKIP` per test), not silent.

Closes #2153

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1273,6 +1273,20 @@ jobs:
       - name: Timer test (reactor poll_oneoff backend)
         run: wasmtime run --dir=. zig-out/bin/kaappi.wasm tests/wasm/timers.scm
 
+      # kaappi#2153: the wasm32-wasi unit suite as a compile gate (like
+      # `-Dtarget=<arch>-windows` — the foreign run skips, compilation must
+      # not) plus execution of the produced binary under wasmtime. The step
+      # runs everything the WASI target can run: VM/GC/reader/printer, the
+      # reactor's CLOCK path (timer tests in tests_reactor*), scheduler and
+      # fiber logic. fd-pair/thread/ffi/Scheme-file-I/O tests skip there
+      # with per-test comptime gates — WASI p1 under wasmtime cannot
+      # construct an EAGAIN-capable fd (no pipe/socketpair syscalls,
+      # sock_open unimplemented) — see docs/dev/porting.md Stage 3.
+      - name: WASI unit tests (compile gate + run under wasmtime, kaappi#2153)
+        run: |
+          zig build test -Dtarget=wasm32-wasi
+          wasmtime run --dir=. --dir=/tmp zig-out/bin/unit-tests.wasm
+
       - name: Parallel pool test (fiber-degraded, KEP-0002 Phase 5)
         run: wasmtime run --dir=. zig-out/bin/kaappi.wasm tests/wasm/parallel.scm
 
--- a/build.zig
+++ b/build.zig
@@ -8,11 +8,12 @@ pub fn build(b: *std.Build) void {
     // heavy workloads. ReleaseSafe matches ReleaseFast in throughput here while
     // keeping bounds/safety checks (fixnum overflow auto-promotes to bignum).
     // Override for development with `-Doptimize=Debug`.
-    const optimize = b.option(
+    const optimize_opt = b.option(
         std.builtin.OptimizeMode,
         "optimize",
         "Prioritize performance, safety, or binary size (default: ReleaseSafe)",
-    ) orelse .ReleaseSafe;
+    );
+    const optimize = optimize_opt orelse .ReleaseSafe;
 
     // Standalone binary: embed compiled bytecode via -Dbundle=path/to/program.sbc
     const bundle = b.option([]const u8, "bundle", "Path to .sbc bytecode file to embed for standalone binary");
@@ -419,10 +420,39 @@ pub fn build(b: *std.Build) void {
     fuzz_gen_step.dependOn(&b.addInstallArtifact(fuzz_gen_exe, .{}).step);
 
     // Unit tests
+    //
+    // wasm32-wasi test-module specifics (kaappi#2153), making
+    // `zig build test -Dtarget=wasm32-wasi` a compile gate like Windows:
+    //
+    //  * single_threaded, matching the wasm executable (wasm32-wasi has no
+    //    wasi-threads here; std's Io/poll plumbing otherwise analyzes its
+    //    futex paths, whose inline asm requires the atomics feature).
+    //  * The atomics CPU feature on the test target: that same std code
+    //    emits `memory.atomic.*` even single-threaded. Only atomic *waits*
+    //    require shared memory, and a single-threaded module never waits,
+    //    so wasmtime runs the binary unmodified. The shipped kaappi.wasm
+    //    keeps its plain baseline CPU — only the test module needs this.
+    //  * ReleaseSmall when `-Doptimize` was not passed explicitly: Debug
+    //    exceeds wasmtime's per-function locals limit in the comptime-
+    //    generated ffi.callFfiGeneric dispatchers (wasm32 halves the slot
+    //    count per local), and ReleaseSafe crashes the LLVM wasm32 backend
+    //    selecting a float constant-pool entry ("Cannot select: i32 =
+    //    ConstantPool<float ...>") in test-only code. ReleaseSmall compiles
+    //    the same Sema surface — every error class the gate exists for is
+    //    semantic, not codegen — and is what CI runs under wasmtime. An
+    //    explicit -Doptimize is always respected.
+    const wasm_test_target = if (is_wasm_target) b.resolveTargetQuery(.{
+        .cpu_arch = .wasm32,
+        .os_tag = .wasi,
+        .cpu_features_add = std.Target.wasm.featureSet(&.{.atomics}),
+    }) else target;
     const test_mod = kaappiModule(b, options_mod, .{
-        .target = target,
-        .optimize = optimize,
+        .target = wasm_test_target,
+        .optimize = if (is_wasm_target and optimize_opt == null) .ReleaseSmall else optimize,
         .embed = null_embed,
+        // Match the wasm executable's own threading model (see `main_mod`
+        // above): wasm32-wasi without wasi-threads is single-threaded.
+        .single_threaded = if (is_wasm_target) true else null,
     });
     const unit_tests = b.addTest(.{
         .name = "unit-tests",
@@ -439,6 +469,14 @@ pub fn build(b: *std.Build) void {
     run_unit_tests.skip_foreign_checks = true;
     const test_step = b.step("test", "Run unit tests");
     test_step.dependOn(&run_unit_tests.step);
+    if (is_wasm_target) {
+        // Install the cross-compiled test binary so a wasm runtime can
+        // execute it (the run step above skipped as foreign):
+        //   wasmtime run --dir=. zig-out/bin/unit-tests.wasm
+        // The wasm CI job does exactly that (kaappi#2153).
+        const install_wasm_tests = b.addInstallBinFile(unit_tests.getEmittedBin(), "unit-tests.wasm");
+        test_step.dependOn(&install_wasm_tests.step);
+    }
 
     const thottam_test_mod = kaappiModule(b, options_mod, .{
         .root = "src/thottam.zig",
--- a/docs/dev/porting.md
+++ b/docs/dev/porting.md
@@ -215,10 +215,25 @@ thottam → #1608 readiness), and it kept every intermediate PR shippable.
       `tests_reactor_parity.zig` is the one written *as* the
       cross-backend comparison: it names no backend, so a property that
       holds on kqueue/epoll/Windows and fails on the new one is a parity
-      defect by construction. Note that this criterion is currently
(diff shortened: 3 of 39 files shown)


baijum:
**Review (Claude Code)** — ✅ approve-with-nits

The gate genuinely **compiles and executes** — not a compile-only gate dressed up as a run. CI does `zig build test -Dtarget=wasm32-wasi` (compile gate; `skip_foreign_checks=true` skips only host *execution*, compilation must still succeed) then `wasmtime run … unit-tests.wasm`, and Zig's test-runner exit code propagates through wasmtime, so a failing test fails the step. The wasm test module is `single_threaded` + `.atomics`, defaults to `ReleaseSmall` (Debug exceeds wasmtime's per-function locals limit; ReleaseSafe crashes the LLVM wasm32 backend) — all documented inline and sound. `kaappi.wasm` is untouched, so the existing wasm timer/parallel steps still work.

**Coverage of #2153 — full as-framed, with one honest caveat.** The issue offered "(a) compile+run or (b) document the boundary"; this does (a) for what WASI can run and (b) for what it can't. Complaint 1 (suite didn't compile for wasm32-wasi) is fully fixed — compiles and runs (reported 1542 passed / 209 skipped / 0 failed, same test count as native). Complaint 2 (poll_oneoff **fd path** covered by nothing): the CLOCK path now genuinely runs against `WasiPollBackend` under wasmtime, but the `arm`/`subFd`/`clearInterest`/`wait` fd branch stays uncovered because WASI p1 under wasmtime can't construct any EAGAIN-capable fd (no pipe/socketpair; `sock_open` unimplemented). That's now a **documented** platform limitation (`docs/dev/porting.md` Stage 3, `tests_reactor_parity.zig`) rather than an oversight — so closing #2153 is legitimate, but be aware the title's literal first clause persists as a known boundary.

Nice touches: every wasm skip is a `comptime` `error.SkipZigTest` with a reason tied to a real pre-existing gate (no fd pairs, no OS threads, bytecode-writes/FFI/SRFI-170/file-I/O gated), and `testing_helpers.wasmNoFdPairs()` **panics** if a test reaches a pair constructor without its skip — a forgotten skip fails loudly, not vacuously. Real regression fix rides along: `fixnumFitsUsize` guards `make-vector/string/bytevector` against wasm32 `@intCast` truncation (absurd length silently allocating a tiny object), covered by `tests_exceptions.zig` now running on wasm.

Nits (non-blocking):
- ⚠️ `src/platform.zig` grows **1414 → 1601 lines**, crossing the 1500 policy. The WASI arm (preopen resolution, `fd_readdir` DirIter, `dl*` stubs) is a natural seam for a `platform_wasi.zig`. Advisory-only, but this is the one thing I'd act on.
- A few `tests_srfi18.zig` skip-reason comments say "OS threads are unregistered on wasm" when the real reason is the whole `(srfi 18)` surface (incl. `seconds->time`) is gated — accurate outcome, imprecise wording.

_Not built for wasm locally (shared cache); the gate is self-verifying on merge since CI runs this exact target._

Issue #1927, opened 2026-07-31
--lib-path cannot shadow a bundled (srfi N); it falls through silently, so A/B tests look like they ran


baijum:
**Category:** Library loader / tooling · **Severity:** wrong-result (silent) · **Found by:** systematic audit v2, Phase 3.1 (#1890)

`--lib-path` works for a *new* library but cannot shadow a bundled `(srfi N)`. The bundled one wins, with no diagnostic — so an A/B comparison appears to run and silently measures the wrong thing.

## Reproduction

```bash
mkdir -p shadow/srfi
echo '(define-library (srfi 14) (export char-set?)
        (begin (define (char-set? x) (quote SHADOWED))))' > shadow/srfi/14.sld
echo '(import (scheme base) (scheme write) (srfi 14)) (write (char-set? 1))' > p.scm

kaappi --lib-path shadow p.scm     # expected SHADOWED, actual #f
```

## Discriminating control — `--lib-path` itself is fine

```bash
mkdir -p mylib/my
echo '(define-library (my thing) (export greet)
        (begin (define (greet) (quote from-libpath))))' > mylib/my/thing.sld
kaappi --lib-path mylib q.scm      # from-libpath   <- works
```

So the flag resolves new libraries correctly. Only a name that already exists as a bundled library is unshadowable.

## Why this is worth fixing rather than documenting

`kaappi --help` describes `--lib-path <path>` as adding a library search path, and `CLAUDE.md` states the auto-added directories are appended *"after any `--lib-path` entries"* — both imply `--lib-path` has priority. It does not, for bundled SRFIs.

Worse, it **fails silently**. A comparison of an old and new implementation of the same SRFI produces a plausible-looking result computed entirely from the bundled library. This bit this very campaign: a Phase 3.1 control intended to run a new conformance suite against the *old* `(srfi 14)` reported the same 172 passes as the new one. The real answer, obtained by physically swapping `lib/srfi/14.sld`, is **17 pass / 147 fail** — the control had been vacuous, and the finding it was meant to support was unverified until re-done a different way.

## Suggested fix

Either give `--lib-path` genuine precedence over bundled libraries (matching the documented ordering), or reject/warn when a `--lib-path` entry contains a file that shadows a bundled library, so the fall-through is visible. Silence is the part that causes damage.

## Workaround for anyone doing this today

[...]

PR #2351, opened 2026-08-25
SRFI 231: validate boolean options and reshape strided views

### LINKED ISSUE #2350: SRFI 231: missing argument validation + specialized-array-reshape rejects valid strided-view reshapes
Follow-up to the SRFI 231 anomalies reported by @gambiteer (Brad Lucier) on the
r/KaappiScheme thread, after v0.24.0. Two themes, both checked against the SRFI 231
text and reference implementation (`generic-arrays.scm`).

## 1. Missing "it is an error" argument validation

These `it is an error` conditions returned a wrong-typed result instead of raising:

```scheme
;; the mutable?/safe? options must be booleans
(list->array   (make-interval '#(0) '#(1)) '(1000)   generic-storage-class 'a)  ; => array, should error
(vector->array (make-interval '#(0) '#(1)) '#(1000)  generic-storage-class 'a)  ; => array, should error
;; copy-on-failure? must be a boolean
(specialized-array-reshape A-ref (make-interval '#(100)) 'a)                     ; => array, should error
```

## 2. `specialized-array-reshape` rejects valid reshapes of strided views

The port only handled the `array-packed?` case, so it raised
`requested reshape is not affinely representable without copying` for a reversed
(negatively-strided) view — even though the elements are still affinely reachable
(step the body backwards):

```scheme
(define array (array-reverse (array-copy (make-array (make-interval '#(2 1 3 1)) list))))
(specialized-array-reshape array (make-interval '#(6)))     ; wrongly errors — should succeed
(specialized-array-reshape array (make-interval '#(3 2)))   ; wrongly errors — should succeed
```

Per the SRFI, the affine-detectability test is "modeled on the corresponding code
in the Python library NumPy" (`_attempt_nocopy_reshape`), which is more general
than a packed-only check.

## Fix

- Add `%check-boolean!` on the mutable?/safe? opts of `list->array`/`vector->array`
  and on `copy-on-failure?` of `specialized-array-reshape`.
- Replace the packed-only reshape shortcut with a faithful port of the reference's
  NumPy `_attempt_nocopy_reshape`: probe the affine indexer for base + per-axis
  strides, drop size-1 axes, greedily match adjacent-axis volume groups, verify
  C-contiguity within each group. Genuinely non-affine reshapes still raise
  (no false positives); body sharing (no copy) is preserved.
- Regression tests mirroring Brad's reference test cases (reversed / per-axis-flipped
  / `array-sample`d reshapes, plus the non-affine `test-error` cases).

baijum:
Fixes the SRFI 231 anomalies @gambiteer (Brad Lucier) reported on r/KaappiScheme after v0.24.0. Closes #2350.

## Theme 1 — missing "it is an error" argument validation

`list->array` / `vector->array` accepted a non-boolean `mutable?`/`safe?`, and `specialized-array-reshape` accepted a non-boolean `copy-on-failure?`, silently returning a wrong-typed array. Now each site calls `%check-boolean!`, matching the reference implementation's per-argument checks.

```scheme
(list->array (make-interval '#(0) '#(1)) '(1000) generic-storage-class 'a)   ; now raises
(specialized-array-reshape A-ref (make-interval '#(100)) 'a)                 ; now raises
```

## Theme 2 — `specialized-array-reshape` rejected valid strided-view reshapes

The old port only handled `array-packed?`, so it raised "not affinely representable" for a reversed (negatively-strided) view even though its elements are affinely reachable by stepping the body backwards. Replaced with a faithful port of the reference's NumPy `_attempt_nocopy_reshape` (probe the affine indexer for base + strides, drop size-1 axes, greedily match adjacent-axis volume groups, verify C-contiguity per group). Genuinely non-affine reshapes still raise; body sharing (no copy) is preserved.

```scheme
(define a (array-reverse (array-copy (make-array (make-interval '#(2 1 3 1)) list))))
(specialized-array-reshape a (make-interval '#(6)))    ; now succeeds, reversed order preserved
(specialized-array-reshape a (make-interval '#(3 2)))  ; now succeeds
```

## Tests

Regression tests mirror the reference suite's reversed / per-axis-flipped / `array-sample`d reshapes and its non-affine `test-error` cases, plus the boolean-option rejections. All 7 SRFI-231 suites and the full 235-file SRFI suite pass.


baijum:
**Review (Claude Code, fresh-cache binary spot-checks)** — ✅ approve

Both #2350 anomalies fully addressed and verified against the running library (fresh `KAAPPI_HOME`, worktree `--lib-path`):

1. **Boolean-option validation** — `%check-boolean!` added on `mutable?`/`safe?` of `list->array`/`vector->array` (combinators.sld) and `copy-on-failure?` of `specialized-array-reshape` (views.sld). `'a` for those options now raises; legitimate `#t`/`#f` and defaulted/absent options still succeed (default flows through the boolean-typed parameter — no over-rejection).
2. **Strided-view reshape** — the packed-only shortcut is replaced by a faithful port of NumPy's `_attempt_nocopy_reshape`. Reversed / per-axis-flipped / `array-sample`d views now reshape zero-copy preserving logical order; genuinely non-affine reshapes still raise; `copy-on-failure?` retries via `array-copy` (packed → always succeeds, no infinite recursion).

The stride math was traced against the canonical C-order routine and matches: index-increment recovery of base+strides, size-1-axis filtering, the loop-2 `np!=op` grouping, the C-contiguity check `oldstrides[ok]==olddims[ok+1]*oldstrides[ok+1]`, new-stride derivation and group advance are all faithful (negative strides satisfy the C-contiguity relation, which is why the reversed case works). Empirically PASS: reversed 4d→1d/2d, last-axis-reversed →2d–5d, sampled+reversed, non-affine raises, copy-on-failure retry, empty-array, volume-mismatch raises. **Both suites green: srfi231-combinators 71/71, srfi231-views 106/106.** DCO + Co-Authored-By present.

Nits (non-blocking): the port omits NumPy's `last_stride` propagation for trailing size-1 new axes — *correct* here (a size-1 axis index is always its lower bound, so the term is always `stride*0`), verified on `->3x2x1x1x1`, but a one-line comment would help a future reader diffing against the reference. And `(%opt opts 1 …)` is read once for the check and again in the following `let*` — purely cosmetic.

lib/srfi/231/views.sld
```diff
@@ -336,14 +332,50 @@
     ;; than the reference implementation's full NumPy-derived multi-group
     ;; stride-matching algorithm). ---
```
baijum: nit (docs): this section banner still describes the *previous* implementation — "see the file header for the documented conservative simplification (packed-check based rather than the reference implementation's full NumPy-derived multi-group stride-matching algorithm)" — but the rewritten file header now says the opposite (this *is* the full algorithm), so the two contradict each other. Suggest trimming it to something like:

```scheme
;; --- specialized-array-reshape: see the file header for the
;; algorithm description. ---
```

lib/srfi/231/views.sld
```diff
@@ -353,23 +385,75 @@
...
+                                             (cdr increments))))
+                 ;; drop the size-1 axes: they contribute no stride and
+                 ;; would derail the adjacent-group matching below.
+                 (keep? (lambda (i) (not (eqv? 1 (vector-ref sides i)))))
+                 (olddims (%vector-filter-by-index keep? sides))
+                 (oldstrides (%vector-filter-by-index keep? strides))
+                 (newdims (interval-widths new-domain))
+                 (newnd (vector-length newdims))
+                 (oldnd (vector-length olddims))
+                 (newstrides (make-vector newnd 0))
```
baijum, line 416: nit: worth a half-line here (or at the `make-vector`) saying why leaving unassigned strides at `0` is safe — the only axes that can still hold `0` at `build` time are width-1 new axes, whose `(index - lower)` term is always 0, so the stride value is never observed. The reference implementation calls this divergence out explicitly ("The NumPy code then sets the strides of the last dimensions with side-length 1 to a value, we leave it zero"); without a note, the zeros read like an uninitialized-vector bug that a future cleanup might "fix".

lib/srfi/231/views.sld
```diff
@@ -353,23 +385,75 @@
...
+                           (raw-setter (lambda (val . multi-index)
+                                         ((storage-class-setter storage-class) body (apply new-indexer multi-index) val)))
+                           (getter (if safe? (%safe-getter new-domain raw-getter) raw-getter))
+                           (setter (and mutable?
+                                        (if safe? (%safe-setter new-domain (storage-class-checker storage-class) raw-setter) raw-setter))))
+                      (%make-array new-domain getter setter body new-indexer storage-class safe?)))))
+            ;; NumPy's _attempt_nocopy_reshape, transcribed. Walk minimal
+            ;; adjacent-axis groups of equal volume in the old and new
+            ;; shapes; within each old group require C-contiguity so one
+            ;; affine stride covers it, then derive the new group's strides.
```
baijum, line 443: consider (non-blocking): the reference implementation carries an explicit NumPy BSD 3-clause attribution for exactly this code — "The code for specialized-array-reshape is derived from _attempt_nocopy_reshape in .../numpy/core/src/multiarray/shape.c" plus the license text — because loops 1–4 are a line-by-line translation. We follow the equivalent convention for other ported SRFI code (e.g. `lib/srfi/64.sld`: "Reference implementation by Per Bothner, adapted for Kaappi. SPDX-License-Identifier: MIT"), and this transcription is close enough to verbatim (the probe helper even notes it mirrors `%%compute-multi-index-increments`) that a short attribution note in the file header would match both the reference and our own house style.

baijum (commented):
Verified against both the reference implementation (gambiteer's `generic-arrays.scm`) and NumPy's `_attempt_nocopy_reshape`, and the substance checks out:

- **Faithful transcription.** The stride probing (`%reshape-index-increments` ≡ the reference's `%%compute-multi-index-increments`), size-1-axis filtering by index, the four matching loops, the per-group C-contiguity check, and the stride derivation all match the reference exactly, and the reference in turn matches current NumPy `shape.c`. The only divergences are ones the reference also makes (zero strides for trailing width-1 new axes instead of NumPy's `last_stride`; generic instead of fixnum arithmetic).
- **Termination/bounds.** Since width-1 old axes are dropped and volumes are pre-checked equal, loop-2 can never read past either dims vector (`np`/`op` are prefix products bounded by the equal total volumes) and both sides exhaust simultaneously — no pathological shape can send it out of range or spinning.
- **Tests.** All 7 SRFI-231 suites pass (views 106, combinators 71 including all the new tests), and the full 235-file SRFI directory passes. I also probed beyond the tests: 0-d ↔ 1-d volume-1 reshapes in both directions, non-zero lower bounds on old *and* new domains, interleaved/trailing width-1 new axes, a transposed view correctly failing without `copy-on-failure?` and succeeding with the right values with `#t`, body sharing preserved (writing through a reshape of a reversed view hits the original at the mirrored index), and immutability inherited by the result.

One local-validation hazard I hit while checking (not a defect of this PR): an older copy of srfi 231 in `~/.kaappi/lib/srfi/231/` (from an earlier install) shadows the repo's `lib/srfi/231` when running with the default `KAAPPI_HOME`, so the new tests fail locally against the stale library. A fresh `KAAPPI_HOME` (or removing that copy) makes everything green. Same hazard class as #1927; possibly worth a separate issue.

Minor and pre-existing, fine to leave as-is: the port validates fewer argument types than the reference for these procedures (the reference also checks `list?`/`vector?`/`storage-class?`); this PR adds exactly the boolean checks the anomalies thread was about.

lib/srfi/231/views.sld
baijum: Fixed in 0c0f7956 — trimmed the banner to `;; --- specialized-array-reshape: see the file header for the algorithm description. ---` so it no longer contradicts the rewritten header.

lib/srfi/231/views.sld
baijum: Added in 0c0f7956: a comment above `make-vector` noting that only width-1 new axes keep stride 0, and their `(index - lower)` term is always 0 so the value is never observed — with the reference's "we leave it zero" note.

lib/srfi/231/views.sld
baijum: Added in 0c0f7956: a NumPy BSD 3-Clause attribution in the file header, noting loops 1–4 are a line-by-line translation of `_attempt_nocopy_reshape` via the SRFI 231 reference, matching our house style (e.g. `lib/srfi/64.sld`).

baijum:
Thanks for the thorough verification. All three review nits are addressed in 0c0f7956 (comments/attribution only, no behavior change).

Filed the `~/.kaappi/lib` shadowing hazard as #2352 (same class as #1927). Left the broader argument-type validation (`list?`/`vector?`/`storage-class?`) out of scope here, as suggested — this PR sticks to the boolean checks the anomalies thread raised.

PR #2363, opened 2026-08-26
SRFI 231: fix the spec-audit deviations (#2353-#2362) and vendor the official test suite

### LINKED ISSUE #2353: SRFI 231: u1-storage-class is #f but the spec requires it (u16vector bit-packing is available)
### Summary

`u1-storage-class` is exported as `#f`, but the SRFI 231 spec **requires** implementations to define it: "Implementations shall define … uX-storage-class for X=1, 8, 16, 32, and 64" (srfi-231.html, Storage Classes). The `#f` fallback the spec allows ("Implementations with an appropriate homogeneous vector type should define the associated global variable using make-storage-class. Otherwise, they shall define the associated global variable to #f.") applies only when no appropriate homogeneous vector type exists — and Kaappi ships SRFI 160 `u16vector`, which is exactly the representation the spec documents for u1:

> the sample implementation uses `(vector n (u16vector ...))` to represent the body of an array with a u1-storage-class, where n represents the valid number of bits

The reference implementation (`generic-arrays.scm` lines 1329–1375) implements u1 portably by bit-packing over `u16vector`. `lib/srfi/231/storage-classes.sld:141` defines `(define u1-storage-class #f)`, and its own header comment acknowledges the class is "fully portable-implementable via bit-packing over a u16vector, per the reference implementation's own code".

Consequences: the spec's own worked examples that use `u1-storage-class` (e.g. the Eratosthenes sieve example, the 3×3 bit-board `#b111100110111` example) cannot run, and every conversion/procedure taking a storage class fails with an internal error (`type error in '%record-ref': expected <storage-class>, got #f`) instead of a u1-backed array.

`f8-storage-class`/`f16-storage-class` are also `#f`. For f8 that is the sanctioned fallback (no implementation has f8 vectors); for f16 the reference implements software half-floats over u16vectors — worth a deliberate maintainers' call, but u1 is the clear spec violation since the spec names the u16vector representation explicitly.

### Reproduction

```scheme
kaappi> (import (srfi 231))
kaappi> u1-storage-class
=> #f                                        ; spec requires a storage class
kaappi> (make-specialized-array (make-interval '#(8)) u1-storage-class)
ERROR: type error in '%record-ref': expected <storage-class>, got #f
```

### References

- Spec: Storage Classes section — "Implementations shall define … uX-storage-class for X=1, …"; body-representation note naming the u16vector bit-packing.
- Reference: `generic-arrays.scm` 1329–1375 (u1 via `(vector n u16vector)`), tested by the official suite (e.g. the `#b111100110111` board tests in `test-arrays.scm`).
- Kaappi: `lib/srfi/231/storage-classes.sld:141-143`.

### LINKED ISSUE #2354: SRFI 231: char-storage-class default is #\0 (digit zero, U+0030) — reference and official suite use #\null (U+0000)
### Summary

The default fill value of `char-storage-class` is `#\0` — the **digit zero, U+0030** — while the reference implementation and the official test suite both use `#\null` (NUL, U+0000). A legal program that reads any element of an uninitialized char array gets a silently wrong character.

- Kaappi: `lib/srfi/231/storage-classes.sld:87` (`#\0` in the defaults). Verified: `(char->integer (array-ref (make-specialized-array (make-interval '#(4)) char-storage-class) 0))` → `48`.
- Reference: `generic-arrays.scm:1276` — defaults list ends `#\null`.
- Official test suite: `test-arrays.scm:1101` asserts the default fill is `#\null`.
- Spec prose (Storage Classes, the `char-storage-class` definition) shows `#\0` — the prose is stale relative to its own reference implementation and test suite; per this project's established rule for SRFI 231 ("when the SRFI's prose and its reference implementation disagree, trust the code"), the reference is authoritative.

### Reproduction

```scheme
kaappi> (import (srfi 231))
kaappi> (char->integer (array-ref (make-specialized-array (make-interval '#(4)) char-storage-class) 0))
=> 48                                  ; digit zero #\0
;; reference/official suite: #\null => 0
```

### Fix

Change the default at `lib/srfi/231/storage-classes.sld:87` from `#\0` to the NUL character (`#\x0`), matching `generic-arrays.scm:1276` and `test-arrays.scm:1101`. A port of the official test suite currently fails exactly this assertion.

### LINKED ISSUE #2355: SRFI 231: f32/f64/c64/c128 checkers accept exact values — safe-mode array-set! silently coerces (c64 rational → f32 precision loss); reference requires inexact
### Summary

The checkers of `f32/f64-storage-class` accept any `real?`, and `c64/c128-storage-class` accept any `complex?` — including **exact** integers and rationals. The reference requires inexact values: `flonum?` for fX (`generic-arrays.scm:1322-1323`) and inexact real and imaginary parts for cX (`generic-arrays.scm:1398-1402`). The spec says fX classes "manipulate … floating-point numbers" and cX classes "complex numbers with … floating-point … real and imaginary parts".

Because the checker is the documented value-validation mechanism ("checked for correctness" on `array-set!` for safe arrays, and on `make-specialized-array`'s initial value), this diverges observably:

- **Safe-mode `array-set!` silently coerces exact values** where the reference raises an error:
  - `(array-set! A 3 1)` on a safe `f64` array → accepted, reads back `3.0`. Reference: error.
  - `(array-set! C 1/3 1)` on a safe `c64` array → accepted, reads back `0.3333333432674408+0.0i` — a **silent precision loss through f32 narrowing on a checker-blessed value**.
  - `(array-set! C (make-rectangular 3 4) 1)` on a safe `c64` array → accepted, reads `3.0+4.0i`. Reference: error (exact parts).
- **Exact initial values are accepted at construction** where the reference errors (`make-specialized-array` validates the initializer via the checker, `generic-arrays.scm:2599-2601`): `(make-specialized-array (make-interval '#(1)) f64-storage-class 0)` fills `0.0` silently.

Kaappi sites: `lib/srfi/231/storage-classes.sld:123,126` (`real?` checkers for f32/f64), `:133,137` (`complex?` checkers for c64/c128).

All valid inexact paths are correct; the divergence is the checker's domain.

### Reproduction

```scheme
kaappi> (import (srfi 231))
kaappi> (define C (make-specialized-array (make-interval '#(2)) c64-storage-class 0.0+0.0i #t))  ; safe
kaappi> (array-set! C 1/3 1)
kaappi> (array-ref C 1)
=> 0.3333333432674408+0.0i    ; accepted + silently narrowed; reference: error "value cannot be stored in body"
```

### Fix shape

Tighten the four checkers: fX → `(lambda (x) (and (real? x) (inexact? x)))` (equivalently `flonum?`), cX → `(and (complex? x) (inexact? (real-part x)) (inexact? (imag-part x)))` (`+nan.0`/`-0.0` included, per the reference).

### LINKED ISSUE #2356: SRFI 231: array-decurry re-invokes the outer getter unboundedly (9x vs the spec's once-per-element guarantee); array-block double-accesses AofA
### Summary

`array-decurry` (and its `!` alias) violates the spec's once-per-element access guarantee:

> "array-decurry evaluates each array element of AofA once, and evaluates each element of AofA's array elements once" — srfi-231.html, array-decurry

Kaappi's implementation (`lib/srfi/231/assembly.sld:106-127`) samples one element and validates all inner domains via `interval-for-each` over the **original** getter, and the virtual getter (`assembly.sld:120-124`) re-fetches the outer element for **every** result element. The outer getter therefore fires `1 + outer-volume + result-volume` times instead of `outer-volume`:

```scheme
;; outer = 1-D volume 2 whose (shared) inner is 1-D volume 3, getters counting invocations:
outer-calls 9    ;; spec: 2
```

With effectful or expensive generators (the documented `make-array` use case) this is an unbounded multiplier, and with escaping continuations the re-fetch can observe mutated state between accesses. Values are correct for pure getters. The reference copies `AofA` once up front (`generic-arrays.scm:4799`, plus optional re-specialization at 4808-4816) and block-copies from the copy — each element exactly once.

Related, same guarantee class: `array-block`/`array-block!` access each outer element **twice** (validation pass over the original getter at `assembly.sld:233-236`, then the copy at `:238`) where the spec says "Each element of AofA is accessed once" and the reference validates the copy (`generic-arrays.scm:4864-4866`). Constant factor 2, values correct.

### Reproduction

```scheme
(import (scheme base) (srfi 231))
(define outer-calls 0)
(define (make-inner)
  (make-array (make-interval '#(3)) (lambda (j) j)))
(define outer
  (make-array (make-interval '#(2))
    (lambda (i) (set! outer-calls (+ outer-calls 1)) (make-inner))))
(define d (array-decurry outer))
(array->list d)
;; => outer-calls is 9 (spec: 2: one per outer element)
```

### Fix shape

`(array-copy AofA)` before validating/consuming, as the reference does — the decurry path then reads the copy's materialized inner arrays exactly once per element, and the access-count guarantee holds.

### LINKED ISSUE #2357: SRFI 231: interval-scale silently accepts non-positive and non-integer (rational) scales — no positive-exact-integer validation
### Summary

`interval-scale` never validates that each entry of `scales` is a **positive exact integer**. The spec: "scales is a length-d vector of positive exact integers … It is an error if interval and scales do not satisfy these conditions" (srfi-231.html, interval-scale). The reference rejects bad scales up front with a dedicated check (`generic-arrays.scm:611-614`). Kaappi (`lib/srfi/231/intervals.sld:234-244`) checks only dimension and zero lower bounds, so two classes of invalid input **succeed silently with plausible-looking results**:

1. **Negative scale on a zero-width axis** — fails downstream only when `ceiling(u/s) < 0`, which never happens for `u=0`:
   `(interval-scale (make-interval '#(0)) '#(-1))` → `[0,0)`, no error.
2. **Rational (non-integer) scale** — arithmetic proceeds:
   `(interval-scale (make-interval '#(1)) '#(1/2))` → `[0,2)`, no error.

(Non-empty intervals with s<0 do error via the lower≤upper check, and s=0 errors via division by zero, so the gap is narrow — same shape as the #2318 `array-tile` fix.)

### Reproduction

```scheme
kaappi> (import (srfi 231))
kaappi> (interval-scale (make-interval '#(0)) '#(-1))
=> #<<interval> #(0) #(0)>            ; reference: error "not a vector of positive, exact, integers"
kaappi> (interval-scale (make-interval '#(1)) '#(1/2))
=> #<<interval> #(0) #(2)>            ; reference: error
```

### Fix shape

Add the missing guard, mirroring the existing `%vector-every` validations in the file:

```scheme
(unless (%vector-every (lambda (s) (and (integer? s) (exact? s) (positive? s))) scales)
  (error "interval-scale: scales must be a vector of positive exact integers" scales))
```

### LINKED ISSUE #2358: SRFI 231: unsafe specialized-array getters silently drop extra trailing multi-index arguments (reference errors on wrong arity)
### Summary

Calling the getter of an **unsafe** specialized array with **more multi-index arguments than the array's dimension** silently drops the extra trailing arguments and returns an element, instead of erroring. The spec makes the wrong index count an error ("It is an error if the number of indices does not equal the dimension of the array"), and the reference's fixed-arity getters raise "Wrong number of arguments" even in unsafe mode. Too-few indices *does* error in kaappi; only the too-many direction is silently lenient.

This matches the class of #2312 (silent wrong behavior on invalid input where the reference errors): a typo'd index list yields a plausible value instead of a diagnostic.

Site: the raw getters built for unsafe specialized arrays in `lib/srfi/231/arrays.sld` accept `(lambda multi-index ...)` variadically and index with the first d entries, ignoring the rest. (Safe arrays check arity correctly.)

### Reproduction

```scheme
kaappi> (import (srfi 231))
kaappi> (define u (make-specialized-array (make-interval '#(2 2)) u8-storage-class 7 #f))
kaappi> ((array-getter u) 0 1 999)
=> 7                                   ; third index silently dropped
;; reference: error (wrong number of arguments)
```

### Fix shape

After the in-domain check path (or in the variadic getter), reject `(not (= (length multi-index) d))` regardless of `safe?`, matching the reference's fixed-arity behavior.

### LINKED ISSUE #2359: SRFI 231: invalid-argument diagnostics surface internal primitive errors across entry points (consolidated audit findings)
### Summary

Consolidated low-severity audit findings: several SRFI 231 entry points surface **internal primitive errors** instead of procedure-prefixed validation messages when given invalid arguments, and a few lenient paths diverge from the reference's checking. Behavior is conformant on all valid input (verified against the reference across the full surface); these are diagnostics/validation-quality gaps found in a systematic spec audit. Grouped by class:

**1. Non-array / non-procedure arguments → raw `%record-ref` errors instead of contextual messages** (reference: per-procedure checks, e.g. `generic-arrays.scm:4064-4065` "array-map: Not all arguments after the first are arrays"): `array-map`, `array-for-each`, `array-fold-left/right`, `array-any`, `array-every`, `array-outer-product`, `array-inner-product`, `array->list`, `array->vector`, `array->list*`, `array->vector*`, and non-list/`length`-type inputs to `list->array`/`vector->array` (3rd-arg storage-class likewise). E.g. `(array-map + 5)` → `type error in '%record-ref': expected <array>, got 5`. `array-reduce` and the boolean-option checks show the intended pattern is achievable in this codebase.

**2. `list*->array` / `vector*->array` don't validate `d`** is a nonnegative fixnum up front (reference `generic-arrays.scm:2499-2501`/`2569-2571`): `(list*->array -1 '())` → `type error in 'make-list': …`; all invalid inputs do error, just from the wrong layer.

**3. Unsafe-mode element-check failures report the primitive, not the procedure** (spec: values not storable "shall be detected and raised"): `(array-set! <safe?#f u8 array> … 300)` → `type error in 'bytevector-u8-set!'` instead of a message naming the array procedure. Same for `array-assign!` into an unsafe destination (`assembly.sld:62-69`) — the official suite's commentary (test-arrays.scm:4361 "should check anyway") expects the reference's own `array-assign!: Not all elements of the source can be stored in destination`.

**4. `make-array` lenience vs reference**: (a) `(make-array I g #f)` is accepted as immutable — the reference's 3-argument form requires a procedure (`generic-arrays.scm:1131-1139`); (b) kaappi stores plain-array getters/setters raw, so out-of-domain access on a plain (non-specialized) array runs the closure unchecked (`((array-getter a) 11 0)` on the spec's own example array returns `0`; the reference wraps in index checks via `%%make-safer-array`, `generic-arrays.scm:1099-1124`). The spec's Note ("checking … is up to the programmer") supports kaappi's reading — flagged as a prose/reference conflict, trusting the reference per project rule.

**5. `make-specialized-array` doesn't run the storage-class checker on an explicit initial-value up front** (reference validates, `generic-arrays.scm:2599-2601`); for u8/u16/char the maker still rejects with a storage-layer message, but see #2355 for the float/complex silent-coercion case.

### Reproduction

```scheme
kaappi> (import (srfi 231))
kaappi> (array-map + 5)
ERROR: type error in '%record-ref': expected <array>, got 5
;; reference: "array-map: Not all arguments after the first are arrays: (5)"
```

Fix shape: per-procedure up-front checks at each entry point (the `array-reduce` / `%check-boolean!` pattern already in the files).

### LINKED ISSUE #2360: Nested do + guard in inner body: spurious KP3002 'arithmetic expected number, got pair' at the loop termination test (legal (scheme base) program)
### Summary

A legal pure-`(scheme base)` program fails with a spurious `KP3002` **type error in 'arithmetic': expected number, got #<pair>** at an inner `do` loop's termination test when a `guard` form appears anywhere in that inner loop's body. The loop aborts; in script mode execution then *continues* past the error, so the program silently produces wrong results while only printing the error. Found during the SRFI 231 spec audit (it silently aborted a probe loop) but it involves no SRFI — plain nested `do` + `guard`.

Isolation matrix (all on current main, `zig-out/bin/kaappi` rebuilt today):

| Variant | Result |
|---|---|
| single `do` + `guard` in a `let` binding | ok |
| nested `do`, no `guard` | ok |
| nested `do`, `guard` inside the inner body (binding or via `set!`) | **spurious KP3002 at inner termination test** |
| nested `do` + `guard`, with/without `(srfi 1)` imports | fails either way |

The error is reported at the **termination test** `(= k (+ n 1))` — "arithmetic … got #<pair>" suggests a loop variable (or the test's operand register) is clobbered with a pair, plausibly by the `guard` dynamic-state save/restore interacting with the `do` lowering's variable handling.

### Reproduction

```scheme
(import (scheme base))
(define fails '())
(do ((n 0 (+ n 1))) ((= n 8))
  (do ((k 0 (+ k 1))) ((= k (+ n 1)))
    (let ((got (guard (e (#t 'err)) (* n k))))
      (if (not (number? got)) (set! fails (cons got fails))))))
(display "G fails: ")(write (length fails))(newline)
```

```
error[KP3002]: type error in 'arithmetic': expected number, got #<pair>
      (do ((k 0 (+ k 1))) ((= k (+ n 1)))
G fails: 0        ;; loops aborted; script mode continued
```

Moving the identical `guard` usage out of the nested-loop inner body makes it pass. The original, more SRFI-1-heavy variant also shows the guard *catching* the bogus error once (corrupting the counted result), i.e. the error can originate inside the guarded region's dynamic extent: `/tmp/srfi231/dl-d.scm` (`D fails: 1` plus an uncaught KP3002 at top level).

### Impact

Ordinary defensive code (error handling inside nested loops) errors spuriously; with script-mode continuation the program then completes with silently wrong output. Not yet tested against the native tier (`kaappi compile`) — worth checking whether the bug is in the IR lowering of `do`/`guard` (shared by both tiers) or in VM execution only.

### LINKED ISSUE #2361: SRFI 4: (srfi 4) is missing the u64/s64 homogeneous vector family (u64vector, s64vector, predicates, accessors)
### Summary

Kaappi's `(srfi 4)` library does not export the **u64/s64 homogeneous vector family** — `u64vector`, `u64vector?`, `make-u64vector`, `u64vector-ref`, `u64vector-set!`, `u64vector-length`, `u64->vector`/`vector->u64vector` (and the whole `s64*` set). SRFI 4's specification covers all of u8/u16/u32/**u64**, s8/s16/s32/**s64**, f32/f64, so a conforming `(srfi 4)` must provide them.

The tags exist elsewhere in the implementation — `(srfi 160 u64)` and `(srfi 160 s64)` work, and SRFI 231's `u64-storage-class`/`s64-storage-class` are built on them — so this is a re-export/surface gap of the `(srfi 4)` library itself, not missing functionality.

Found while porting the official SRFI 231 test suite (its imports of `(srfi 160 u64)`/`(srfi 160 s64)` had to be substituted because the natural `(srfi 4)` path lacks them).

### Reproduction

```scheme
kaappi> (import (scheme base) (srfi 4))
kaappi> (u64vector 1 2 3)
ERROR[KP3001]: undefined variable 'u64vector'. Did you mean 'f64vector'?
kaappi> (f64vector 1.0 2.0)   ;; other families fine
=> #f64(1.0 2.0)
```

### Fix shape

Add the u64/s64 bindings to the `(srfi 4)` re-export hub (same source they already exist under for `(srfi 160 u64)`/`(srfi 160 s64)`).

### LINKED ISSUE #2362: SRFI 231: non-specialized view/combinator getters and setters never validate multi-indices (124 official-suite failures: out-of-domain, wrong arity, empty domains silently accepted)
### Summary

Getters and setters of **non-specialized** arrays produced by the view/combinator family never validate the multi-index: out-of-domain, wrong-arity, and calls on empty domains are silently accepted, running the underlying closure on arbitrary inputs. The reference implementation wraps *every* generalized array's getter/setter in index checks (`%%make-safer-array`, `generic-arrays.scm:1099-1124` via `%%wrap-getter-in-index-checks:2149-2195`), so all of these raise errors there. Kaappi checks only on the *safe specialized* path (verified: safe specialized translate/permute/curry DO error correctly).

This is the single largest divergence found by a full adapted run of the official SRFI 231 test suite: **124 of 138 failing evaluations** trace to this one gap. All positive (valid-index) tests pass — values are right; only invalid-input checking is missing. The spec makes out-of-domain multi-indices and wrong index counts errors, and the reference enforces them unconditionally; Kaappi's dev notes follow the spec's Note ("checking … is up to the programmer" for `make-array`), but the reference's behavior is the project's tie-breaker for this SRFI, and the official suite asserts it.

Affected procedures (non-specialized result paths, plus their setters): `array-curry`, `array-map`, `array-translate`, `array-permute`, `array-reverse`, `array-sample`, `array-outer-product`, `specialized-array-share` (plain-array input), `make-array` views generally, and the empty-domain case ("Array domain is empty" in the reference).

### Reproduction (official-suite lines)

```scheme
kaappi> (import (scheme base) (srfi 231))
;; test-arrays.scm:2387 — index 3 ∉ [0,3):
kaappi> (apply (array-getter (array-map length (make-array (make-interval '#(3)) list))) '(3))
=> 3                                  ; reference: error "domain does not contain multi-index"
;; test-arrays.scm:2891 — curry outer, index 100 ∉ [0,3):
kaappi> (array-ref (array-curry (make-array (make-interval '#(3 3)) list) 1) 100)
=> #<array …>                         ; reference: error
;; test-arrays.scm:3332 — translate a plain array out of reach:
kaappi> (array-ref (array-translate (make-array (make-interval '#(3 3)) list) '#(-4 -4)) 0 0)
=> (-4 -4)                            ; reference: error
;; test-arrays.scm:2906 — wrong arity on 2-D plain array:
kaappi> (array-ref (make-array (make-interval '#(3 3)) list) 1 1 1 1)
=> ((1 1 1) 1)                        ; reference: error
;; test-arrays.scm:4626 — empty outer-product domain:
kaappi> ((array-getter (array-outer-product cons A B)) 0 0 0 0)   ; A,B with empty domain
=> ((0 0) 0 0)                        ; reference: error "Array domain is empty"
```

### Fix shape

Wrap plain-array getters/setters at construction in `arrays.sld`'s `make-array` (single choke point — every affected procedure builds its non-specialized results through it), mirroring the reference's `%%wrap-getter-in-index-checks`: check exact-integer arity-matching multi-index against the domain, and always error for empty domains. The file's existing `%safe-getter`/`%safe-setter` already implement this for specialized arrays and can be reused.

Evidence: adapted official-suite run (`/tmp/agent-suite/adapted.scm`, 7,154 evaluations): 124 failures all in this family; each expected an error string from the reference and got a value.

### LINKED ISSUE #2365: Nested do + let-values/let*-values/parameterize/receive: same unboxed-capture KP3002 as #2360 (whitelist still incomplete)
### Summary

A `do`-loop variable captured inside `let-values`, `let*-values`, `parameterize`, or `receive` (SRFI 8) stays **unboxed**, so on the second loop iteration the box object itself is passed to the loop's termination test (or other arithmetic), producing a spurious `KP3002` `type error in '=': expected number, got #<pair>` on a legal `(scheme base)` program.

This is the **same root cause as #2360** (nested `do` + `guard`). PR #2363 fixes the `guard` instance by adding `"guard"` to the `CaptureScan` closure-form whitelist in `src/compiler_bindings.zig`, but the sibling forms that also desugar to internal lambdas were already missing from that list and remain broken.

### Reproductions

All fail on `main` and on the #2363 branch. Each should evaluate to `6`:

```scheme
;; let-values
(let ((acc 0))
  (do ((n 0 (+ n 1))) ((= n 3))
    (do ((i 0 (+ i 1))) ((= i n))
      (set! acc (+ acc (let-values (((x) (values (+ n i)))) x)))))
  acc)

;; let*-values — same
(let ((acc 0))
  (do ((n 0 (+ n 1))) ((= n 3))
    (do ((i 0 (+ i 1))) ((= i n))
      (set! acc (+ acc (let*-values (((x) (values (+ n i)))) x)))))
  acc)

;; parameterize
(define p (make-parameter 0))
(let ((acc 0))
  (do ((n 0 (+ n 1))) ((= n 3))
    (do ((i 0 (+ i 1))) ((= i n))
      (set! acc (+ acc (parameterize ((p 0)) (+ n i))))))
  acc)

;; receive (import (srfi 8))
(let ((acc 0))
  (do ((n 0 (+ n 1))) ((= n 3))
    (do ((i 0 (+ i 1))) ((= i n))
      (set! acc (+ acc (receive (x) (values (+ n i)) x)))))
  acc)
```

Actual output for each:

```
error[KP3002]: type error in '=': expected number, got #<pair>
```

### Root cause

`compileDo` runs `CaptureScan` over the loop body to decide which loop variables must be boxed *before* the loop back-edge (so a closure that captures them sees mutations, and every access compiles as box-aware). `CaptureScan.walk` only flips into "inside a closure" mode for a hardcoded whitelist of head symbols:

```
lambda / case-lambda / delay / delay-force   (+ guard, added by #2363)
```

But these forms also lower into internal lambdas that capture free references:

- `compileLetValues` desugars to `(apply (lambda (formals) body) temp)` — the consumer lambda captures the loop var in `body`.
- `compileParameterize` wraps its body in a thunk.
- `let*-values` builds on the same `let-values` desugaring.
- `receive` (SRFI 8) expands to `call-with-values` + a lambda consumer.

Because none is in the whitelist, `CaptureScan` walks their sub-expressions with `in_closure = false`, misses the capture, and the loop variable is left unboxed. The `box_local` is then emitted mid-loop (when the internal lambda is compiled), while the loop test — compiled earlier — still reads the raw register, so iteration 2 sees the box object. This is exactly the failure mode #2360 describes for `guard`.

### Suggested fix

Two options (noted on the #2363 review):

1. **Extend the whitelist** with `let-values`, `let*-values`, `parameterize` (and confirm `receive` — as a macro it may need the fallback path to actually work; empirically it does not today). Cheap, but perpetuates the hand-maintained-list fragility the #2363 comment itself warns about.
2. **Make `CaptureScan` structural** rather than a name whitelist, so any form lowered through an internal lambda is covered automatically. More robust; closes the class rather than the instances.

A regression test per form belongs in `src/tests_derived_forms.zig` (alongside the `guard` test #2363 added) and/or a Scheme test under `tests/scheme/`.

### Notes

- Pre-existing; not introduced by #2363. Filed separately so the remaining instances aren't lost when #2363 merges with only `guard` fixed.
- Severity mirrors #2360 (`priority: high`): legal `(scheme base)` program, reachable pattern, wrong outcome.

baijum:
## Summary

Fixes all ten deviations filed by the systematic SRFI 231 spec audit (spec + reference implementation + the official 6,970-line test suite, each finding verified by reproduction before fixing), then vendors that official suite as a permanent conformance asset so this class of drift stays caught.

### Fixes (one commit each)

| Commit | Issues | What |
|---|---|---|
| `0e077fb5` | #2353 #2354 #2355 | **u1-storage-class implemented** — port of the reference's u16vector bit-packing (the spec mandates uX for X=1 and documents exactly this representation; the spec's own board example now reproduces); char default → `#\null` per the reference + official suite; f32/f64/c64/c128 checkers require inexact values (safe-mode `array-set!` no longer silently coerces — `1/3` into c64 lost f32 precision) |
| `6d7c5184` | #2357 | `interval-scale` rejects non-positive/non-integer scales |
| `176fb39b` | #2358 #2362 #2359 | `make-array` wraps getters/setters with index checks as the reference's `%%make-safer-array` does — **the one gap behind 124 of 138 official-suite failures**; unsafe getters reject over-long multi-indices (were silently dropped); constructor argument/initial-value validation |
| `537150ec` | #2361 | `(srfi 4)` re-exports the missing u64/s64 family (SRFI 4 covers all ten kinds) |
| `44ed1980` | #2359 | Per-procedure argument validation across the combinators + upfront element-vs-checker validation in `list->array`/`vector->array` (unchecked fills silently corrupted u1 bits) |
| `e5d2fc64` | #2356 #2359 | `array-decurry`/`array-block` copy-first, honoring the spec's once-per-element access guarantee (was 9× on a 2×3 case, unbounded in result volume); `array-assign!` validates values even for unsafe destinations |
| `ed12b477` | #2360 | **Compiler fix, root-caused by bytecode disassembly**: `guard` desugars to internal lambdas but `CaptureScan`'s closure-form whitelist didn't include it — a do-variable captured by a guard stayed unboxed until mid-loop, and the second loop iteration passed the box object itself to `=`. Legal `(scheme base)` program, silently wrong under script mode |
| `77ecb37c` | (asset for #2362) | Vendor the **official SRFI 231 test suite** as `tests/scheme/srfi/srfi231-official.scm` — generated, never hand-edited, from the pristine upstream source also vendored in `srfi231-official-fixtures/`; known kaappi-vs-reference divergences are accounted in a table that fails CI when an entry goes stale; wired into `run-all.sh` with a per-file timeout override (~150s cold-cache). `docs/dev/testing.md` documents it |

### Validation

- Official SRFI 231 suite (adapted): **138 failures → 9** (all documented residuals: f16 deferral ×6, unsafe-view UB per spec text ×2, Gambit string-mutability ×1), now executed as 8,778 evaluations (u1 unlocked 1,600+ previously-blocked tests)
- `bash tests/scheme/run-all.sh`: **718/718 Scheme files, 1,395/1,395 R7RS, exit 0** (includes the new vendored suite)
- `zig build test`: **1,842/1,849** (7 skipped); `zig fmt --check src/` clean
- Regression tests accompany every fix (Scheme + one Zig unit test for the compiler fix in `tests_derived_forms.zig`)

### Deliberately left as-is (noted for reviewers)

- **f16-storage-class stays `#f`** — the reference implements software half-floats; a port of its own to make, tracked in #2353's comments rather than silently skipped.
- Two knowingly-lenient paths from #2359 remain (plain `array-set!` on unsafe specialized arrays surfaces the storage-layer message; `(make-array I g #f)` accepted as immutable) — both defensible under the spec's "it is an error" umbrella.

Fixes #2353
Fixes #2354
Fixes #2355
Fixes #2356
Fixes #2357
Fixes #2358
Fixes #2359
Fixes #2360
Fixes #2361
Fixes #2362

## Checklist

- [x] `zig build test` passes
- [x] Scheme test suites pass (`bash tests/scheme/run-all.sh`)

Fixes #2365 (the structural fix, commit 60f87d23, closes the whole class — see the issue for the four named forms and the review thread for the macro case)


src/compiler_bindings.zig
```diff
@@ -351,10 +351,23 @@ const CaptureScan = struct {
...
+            // the loop while the loop test still read the register raw —
+            // the second iteration then passed the box object itself to
+            // arithmetic (kaappi#2360). Any new compiler form that wraps
+            // sub-expressions in an internal lambda MUST be added here.
             if (std.mem.eql(u8, h, "lambda") or
                 std.mem.eql(u8, h, "case-lambda") or
                 std.mem.eql(u8, h, "delay") or
-                std.mem.eql(u8, h, "delay-force"))
+                std.mem.eql(u8, h, "delay-force") or
+                std.mem.eql(u8, h, "guard"))
```
baijum: The `guard` fix here is correct and I reproduced it (the two new unit tests pass, and a hand build with only this one-line addition flips the `do`+`guard` repro from a `KP3002` crash to the right answer). But the new comment asserts an invariant this whitelist still doesn't satisfy: **`guard` is not the only `(scheme base)` form that desugars to an internal lambda, and the siblings remain broken with the identical failure.**

`compileLetValues` builds `(apply (lambda (formals) body) temp)` and `compileParameterize` wraps its body in a thunk — both capture free references the same way `guard` does. Neither is in this list, so a `do`-loop variable captured inside them stays unboxed and the second iteration passes the box object to arithmetic. All four of these fail *today* on this branch (`git checkout` at `93f4cab` + `zig build`) with the same `type error in '=': expected number, got #<pair>`:

```scheme
;; let-values — expect 6, errors
(let ((acc 0))
  (do ((n 0 (+ n 1))) ((= n 3))
    (do ((i 0 (+ i 1))) ((= i n))
      (set! acc (+ acc (let-values (((x) (values (+ n i)))) x)))))
  acc)

;; parameterize — expect 6, errors
(define p (make-parameter 0))
(let ((acc 0))
  (do ((n 0 (+ n 1))) ((= n 3))
    (do ((i 0 (+ i 1))) ((= i n))
      (set! acc (+ acc (parameterize ((p 0)) (+ n i))))))
  acc)
```

`let*-values` and `receive` (SRFI 8) reproduce it too. This is pre-existing (the whitelist was already incomplete before this PR), but the PR is the right place to note it because it touches exactly this code and adds a comment asserting completeness. Two options:

1. Also add `let-values`, `let*-values`, `parameterize` (and audit `receive`/`case-lambda`, already present) to the whitelist — cheap, but keeps the hand-maintained-list fragility the comment itself warns about.
2. Make `CaptureScan` structural instead of a name whitelist, so any form lowered through an internal lambda is covered automatically (matches the "MUST be added here" concern the comment raises).

At minimum I'd fix `let-values`/`let*-values`/`parameterize` in this PR since they're trivially reachable from legal `(scheme base)` code and silently wrong.

---

Filed as #2365 to track the remaining instances separately, so they aren't lost when this PR merges with only `guard` fixed.

baijum:
## Review

Went through the SRFI 231/4 library changes and the compiler fix. The SRFI work is solid — I built the branch's `.sld` tree against a local binary and confirmed the behavioral claims hold:

- **u1 storage class** round-trips correctly (`array-set!` value/index order respected; bits 0,2,4 → `(1 0 1 0 1)`), and `storage-class-length` reports the valid-bit count, not the padded u16 width.
- **Float/complex checkers** reject exact values in **safe** mode (`1/3` into f32 → error; `1.0`/`make-rectangular` into c64 behave as the comment describes), while the unsafe path staying lenient matches the "left as-is" note.
- **`interval-scale`** rejects negative and rational scales.
- **SRFI 4** now resolves `u64vector`/`s64vector` (was a hard failure).
- The `array-decurry`/`array-block` copy-first refactors and the upfront element/argument validations in the combinators read correctly and match the reference’s ordering.

One substantive issue, left inline on `src/compiler_bindings.zig`: the `guard` fix is correct but the `CaptureScan` whitelist is **still incomplete** — `let-values`, `let*-values`, and `parameterize` desugar to internal lambdas the same way and reproduce the *identical* `type error in '=': expected number, got #<pair>` on a `do`-loop variable, today, on this branch. It's pre-existing, but the PR both touches this exact list and adds a comment asserting the invariant, so it's the right place to either extend the list or make the scan structural. Details + repros in the inline comment.

Everything else LGTM; the vendored official suite as a conformance asset is a good call.

src/compiler_bindings.zig
baijum: Went with option 2 (structural), pushed as `60f87d23` — you were right that the whitelist was unenumerable, and it's worse than the three siblings: **any user macro expanding to a capturing lambda fails identically**, because the scan runs pre-expansion and can't see the lambda `receive` (SRFI 8) expands to. Verified failing on the branch before the fix:

```scheme
(define-syntax capture-n (syntax-rules () ((_ e) (lambda () e))))
;; ... do-loop capturing n through (capture-n (+ n i)) => same KP3002
```

The scan now collects **every** symbol in the do's test/commands/steps/results and boxes each resolvable local before the loop (globals fail `resolveLocal` and drop out). Over-boxing is safe because `get_box_local`/`set_box_local` lazily box raw registers — verified the edge you'd worry about (a conditionally-skipped do whose captured variable is read afterwards).

All four of your repros (let-values, let*-values via the same path, parameterize, receive) now return 6; your two repros plus the macro case are pinned as unit tests in `tests_derived_forms.zig`.

Measured cost: full benchmark suite flat-to-faster (medians: fib 2.67→2.56s, nqueens 2.80→2.77s, primes 0.359→0.343s, tak 1.93→1.80s — noise, no regression). The honest worst case is a pure-counting `do` whose only references are its own step/test: ~2x on that micro (147ns vs 74ns/iter against the equivalent named let) — one box per variable per loop entry, off the back-edge. A sound way to recover it later would be lowering `guard`/`let-values`/`parameterize` through the expander so a scan over *fully-expanded* code sees explicit lambdas again; noted in the commit message. Corpus: 1,842/1,849 unit (7 skipped), 718/718 Scheme files, 1,395/1,395 R7RS.

PR #2364, opened 2026-08-26
thottam: unit-tier regression guard for ownership-aware removal (#2136)

### LINKED ISSUE #2136: [Tooling] thottam remove deletes library files by name with no ownership record — removing one package breaks another that is still installed

baijum:
## Summary

Issue #2136 — `thottam remove` deleting shared library files by name — was **already fixed on `main`** by #2289, which added the per-file ownership manifest (`thottam.files`): install records which package owns each installed file, and removal unlinks only files no other installed package still claims. That fix ships with unit coverage for the manifest primitives in `thottam_state.zig` and an end-to-end scenario in `tests/scheme/thottam/thottam-lifecycle.sh`.

The gap this PR closes: the shell test needs a real git remote and is **skipped by `zig build test`**, so the removal path had no guard in the unit suite that runs on every build. This PR adds a **network-free Zig-tier regression test** and closes the tracking issue.

## What the test does

`tests_thottam.zig` gains `doRemove keeps a file a still-installed package still claims (issue #2136)`:

- Lays out two packages, each shipping its own library plus a shared `lib/kaappi/shared.sld`, in a throwaway `$KAAPPI_HOME/src` exactly as a clone would.
- Installs both through the **real** install-side file sync (`syncInstalledFiles`) — copies into the shared lib dir and records ownership.
- Drives the **real** `doRemove`: removing `kaappi-two` must keep `kaappi-one`'s files *and* the shared file it still claims, while dropping `kaappi-two`'s own unique file.
- Removing the last claimant finally deletes the shared file — confirming removal is claim-guarded, not over-broad.

This fails against the pre-#2289 removal-by-name (which walked the removed package's own source tree and unlinked `shared.sld` unconditionally) and passes against the ownership-aware code.

`doRemove` and `syncInstalledFiles` are made `pub` so the test can drive them, matching `doList`/`doUpdate`/`doVerify`, which are already `pub` for the same reason.

## Verification

- `zig build test` — 1842/1849 pass (7 skipped), including `thottam-tests` 84 pass.
- `bash tests/scheme/thottam/thottam-lifecycle.sh` — 145 pass, 0 fail.
- `src/thottam.zig` stays at 1379 lines (well under the 1500 cap).

Closes #2136

### DIFF
--- a/src/tests_thottam.zig
+++ b/src/tests_thottam.zig
@@ -1022,14 +1022,18 @@ fn installedFileExists(allocator: std.mem.Allocator, lib_dir: []const u8, rel: [
 }
 
 /// Run the real install-side file sync for `pkg` (copy its lib tree into the
-/// shared lib dir and record ownership in thottam.files), capturing its
-/// output so the suite stays quiet.
-fn syncPkg(allocator: std.mem.Allocator, config: thottam.Config, pkg: []const u8) !void {
+/// shared lib dir and record ownership in thottam.files), returning the
+/// captured stdout+stderr (owned by the caller) so a collision warning can be
+/// asserted — and so the suite stays quiet when it is not.
+fn syncPkg(allocator: std.mem.Allocator, config: thottam.Config, pkg: []const u8) ![]const u8 {
     const pkg_dir = try std.mem.concat(allocator, u8, &.{ config.src_dir, "/", pkg });
     defer allocator.free(pkg_dir);
     const cap = try captureOutput(true, thottam.syncInstalledFiles, .{ allocator, config, pkg, pkg_dir });
-    defer allocator.free(cap.out);
-    if (cap.err) |e| return e;
+    if (cap.err) |e| {
+        allocator.free(cap.out);
+        return e;
+    }
+    return cap.out;
 }
 
 test "doRemove keeps a file a still-installed package still claims (issue #2136)" {
@@ -1048,8 +1052,14 @@ test "doRemove keeps a file a still-installed package still claims (issue #2136)
 
     // Install both through the real file-sync path: copies into the shared
     // lib dir and records ownership in thottam.files.
-    try syncPkg(allocator, env.config, "kaappi-one");
-    try syncPkg(allocator, env.config, "kaappi-two");
+    allocator.free(try syncPkg(allocator, env.config, "kaappi-one"));
+    // The second install overwrites a file kaappi-one's manifest already
+    // claims. warnIfClaimed must make that audible on stderr — the "loud, not
+    // silent" half of the #2136 fix, and this is the only unit-tier place that
+    // sees it.
+    const two_out = try syncPkg(allocator, env.config, "kaappi-two");
+    defer allocator.free(two_out);
+    try std.testing.expect(std.mem.indexOf(u8, two_out, "also provided by kaappi-one") != null);
     try thottam.writeFile(allocator, env.config.installed, "kaappi-one\nkaappi-two\n");
 
     // All three files landed in the shared lib dir.
@@ -1069,9 +1079,19 @@ test "doRemove keeps a file a still-installed package still claims (issue #2136)
     try std.testing.expect(installedFileExists(allocator, env.config.lib_dir, "kaappi/one.sld"));
     try std.testing.expect(installedFileExists(allocator, env.config.lib_dir, "kaappi/shared.sld"));
 
-    // doRemove dropped kaappi-two from installed.txt and thottam.files itself.
+    // doRemove dropped kaappi-two from installed.txt.
     try std.testing.expect(!state.isInstalled(allocator, env.config.installed, "kaappi-two"));
     try std.testing.expect(state.isInstalled(allocator, env.config.installed, "kaappi-one"));
+
+    // And it dropped kaappi-two's lines from thottam.files, while kaappi-one's
+    // claim on the shared file — the reason removal kept it — survives. This
+    // is what removal consulted, asserted directly on the manifest.
+    {
+        const owner = state.fileClaimedBy(allocator, env.config.files, "kaappi/shared.sld", "kaappi-two");
+        defer if (owner) |o| allocator.free(o);
+        try std.testing.expect(owner != null);
+        try std.testing.expectEqualStrings("kaappi-one", owner.?);
+    }
 
     // Removing the last claimant finally deletes the shared file — removal is
     // not over-broad, only claim-guarded.
@@ -1082,4 +1102,11 @@ test "doRemove keeps a file a still-installed package still claims (issue #2136)
     }
     try std.testing.expect(!installedFileExists(allocator, env.config.lib_dir, "kaappi/one.sld"));
     try std.testing.expect(!installedFileExists(allocator, env.config.lib_dir, "kaappi/shared.sld"));
+
+    // With kaappi-one gone too, nobody claims the shared file in thottam.files.
+    {
+        const owner = state.fileClaimedBy(allocator, env.config.files, "kaappi/shared.sld", "kaappi-two");
+        defer if (owner) |o| allocator.free(o);
+        try std.testing.expect(owner == null);
+    }
 }
--- a/src/tests_thottam.zig
+++ b/src/tests_thottam.zig
@@ -990,3 +990,96 @@ test "doUpdate of every package skips hostile names read back from installed.txt
     try std.testing.expect(std.mem.indexOf(u8, cap.out, evil) == null);
     try std.testing.expect(std.mem.indexOf(u8, cap.out, "../") == null);
 }
+
+// ---------------------------------------------------------------------------
+// Ownership-aware removal (issue #2136)
+// ---------------------------------------------------------------------------
+//
+// A network-free end-to-end guard for the bug that motivated the file-
+// ownership manifest: `thottam remove` deleted installed library files by
+// name, so removing one package took a shared file another still-installed
+// package relied on. These drive the real install-side file sync
+// (`syncInstalledFiles`) and the real `doRemove`, both over a throwaway
+// $KAAPPI_HOME with source trees laid out as a clone would leave them — no
+// git, so they run under `zig build test` (the thottam-lifecycle.sh cover of
+// the same scenario needs a git remote and is skipped there).
+
+/// Write `<src_dir>/<pkg>/lib/<rel>` with a trivial library body, creating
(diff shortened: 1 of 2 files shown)


baijum (commented):
Reviewed as a test-only close-out of #2136. Everything I could verify checks out; the three inline notes are polish (P3), none blocking.

**Independently verified:**

- `zig build test` green at the PR head; `thottam-tests` reports **84/84** (matches the PR body) and the full suite exits 0.
- `src/thottam.zig` stays at **1379 lines** (matches).
- The central claim — *this test fails against pre-#2289 removal-by-name* — holds by construction. At `59a65520^` (one commit before #2289), `doRemove` called `removeSldFiles`, which collected every `.sld` under the removed package's own `lib/` and unlinked each **unconditionally** (no manifest existed yet). This test plants `shared.sld` in `kaappi-two`'s source tree, so the old code deletes the surviving `lib/kaappi/shared.sld` and the assertion at line 1070 fails. (I read the pre-fix code directly; I did not re-run it.)
- The `pub` widening of `doRemove`/`syncInstalledFiles` matches the `doList`/`doUpdate`/`doVerify` precedent the #2144 tests already rely on.
- The scenario never writes `thottam.lock`, but that's safe: `removeFromLockfile` bails on read failure (`catch return`), so `doRemove` still exits clean.

The test drives the **real** `syncInstalledFiles` and the **real** `doRemove` over the same `TmpHome`/`captureOutput` machinery the #2144 tests use (portable to the Windows legs), and `std.testing.allocator` proves it leak-clean. Good close of the gap: with this, the claim-guarded removal has a guard in the suite that runs on every build.

src/tests_thottam.zig
```diff
@@ -990,3 +990,96 @@ test "doUpdate of every package skips hostile names read back from installed.txt
...
+    return thottam.fileExists(allocator, full);
+}
+
+/// Run the real install-side file sync for `pkg` (copy its lib tree into the
+/// shared lib dir and record ownership in thottam.files), capturing its
+/// output so the suite stays quiet.
+fn syncPkg(allocator: std.mem.Allocator, config: thottam.Config, pkg: []const u8) !void {
+    const pkg_dir = try std.mem.concat(allocator, u8, &.{ config.src_dir, "/", pkg });
+    defer allocator.free(pkg_dir);
+    const cap = try captureOutput(true, thottam.syncInstalledFiles, .{ allocator, config, pkg, pkg_dir });
```
baijum, line 1031: **[P3] Consider asserting the collision warning instead of discarding it.** The second `syncPkg` call (`kaappi-two`, line 1052) makes `warnIfClaimed` print `kaappi/shared.sld is also provided by kaappi-one; overwriting its installed copy` to stderr — the audible half of the #2136 fix — and this capture is the only unit-tier place that sees it. Right now the only guard on that warning is `thottam-lifecycle.sh:500`, which is exactly the git-dependent suite this PR exists to complement (skipped by `zig build test` and on Windows). Having `syncPkg` return the captured output (or asserting inline after the second install: `expect(std.mem.indexOf(u8, out, "also provided by kaappi-one") != null)`) would put the warning's emission under the every-build suite too — same motivation as the PR itself. `captureOutput(true, ...)` already captures stderr, so nothing else changes.

src/tests_thottam.zig
```diff
@@ -990,3 +990,96 @@ test "doUpdate of every package skips hostile names read back from installed.txt
...
+    {
+        const cap = try captureOutput(true, thottam.doRemove, .{ allocator, env.config, "kaappi-two" });
+        defer allocator.free(cap.out);
+        try std.testing.expect(cap.err == null);
+    }
+    try std.testing.expect(!installedFileExists(allocator, env.config.lib_dir, "kaappi/two.sld"));
+    try std.testing.expect(installedFileExists(allocator, env.config.lib_dir, "kaappi/one.sld"));
+    try std.testing.expect(installedFileExists(allocator, env.config.lib_dir, "kaappi/shared.sld"));
+
+    // doRemove dropped kaappi-two from installed.txt and thottam.files itself.
```
baijum: **[P3] The comment promises more than the assertions check.** It says `doRemove` dropped kaappi-two from "installed.txt **and thottam.files itself**", but the two assertions below observe only `installed.txt` — the manifest drop is proven only implicitly, two phases later, by the last-claimant deletion. That proof is sound, but a regression in `removeFromFileManifest` specifically would surface as a confusing failure at line 1084 rather than here. Suggest asserting the claim set directly with the already-imported `state.fileClaimedBy`:

```zig
const owner = state.fileClaimedBy(allocator, env.config.files, "kaappi/shared.sld", "kaappi-two");
try std.testing.expect(owner != null);
defer if (owner) |o| allocator.free(o);
```

(and `== null` after the final remove). Each phase then fails at the exact step that regressed. `fileClaimedBy` is already `pub` in `thottam_state.zig` and returns an owned string, so it needs the `defer free`.

src/tests_thottam.zig
```diff
@@ -990,3 +990,96 @@ test "doUpdate of every package skips hostile names read back from installed.txt
...
+test "doRemove keeps a file a still-installed package still claims (issue #2136)" {
+    const allocator = std.testing.allocator;
+    var env = try TmpHome.init(allocator, "rmshared");
+    defer env.deinit(allocator);
+
+    // Two packages, each shipping its own library plus a shared one, laid out
+    // in $KAAPPI_HOME/src exactly as a clone would. The shared file lives in
+    // BOTH source trees — the shape that let removal-by-name (which walked the
+    // removed package's own source tree) delete it out from under the other.
+    try writeSrcLib(allocator, env.config.src_dir, "kaappi-one", "kaappi/one.sld", "(define-library (kaappi one) (export a) (import (scheme base)) (begin (define a 1)))");
```
baijum, line 1048: **[P3, optional] The legacy discovery half of `removeInstalledFiles` never contributes a file here.** Both packages have manifest lines, so when `doRemove` runs, the source-tree walk (the pre-manifest compatibility path, `thottam.zig:338-351`) finds exactly what the manifest already recorded and `addFileToList` dedupes it away. Neither this test nor `thottam-lifecycle.sh` removes a package with *no* manifest lines — the shape a real user hits after installing with a pre-#2289 thottam and removing after an upgrade. A third package whose files are hand-copied into `lib_dir` with no `thottam.files` lines (removed while kaappi-one still claims the shared file) would exercise that path's claim-guarding directly. Happy to see this deferred to a follow-up; the mainline scenario is well covered.

baijum:
Addressed the review in `c56e6ef4`:

- **P3 (collision warning)** — `syncPkg` now returns the captured stdout+stderr; the kaappi-two install asserts `indexOf(out, "also provided by kaappi-one") != null`, so the audible half of the fix is under the every-build suite too.
- **P3 (assert `thottam.files` directly)** — added `state.fileClaimedBy(...)` checks: `owner == "kaappi-one"` while kaappi-two is gone but kaappi-one remains, `owner == null` after the last claimant is removed (each freed via `defer`), and reworded the comment to match.
- **P3 (legacy no-manifest discovery path)** — deferred per your note; filed as a follow-up so it isn't lost.

Verified locally: `zig build test` green (1758 unit + 84 thottam pass, 7 skipped). CI re-running on the new commit.

Issue #1972, opened 2026-07-31
primitives_io.zig: five typeError sites that are not about a type — four platform gates and one range check


baijum:
**Category:** Diagnostics / error taxonomy · **Severity:** diagnostic quality · **Found by:** follow-on from #1944 (systematic audit v2, Phase 2.2, #1890) — same class, deliberately left out of scope there because #1944 did not name these sites

#1944 fixed the `set-port-position!`, `write-string`, and closed-port sites in `primitives_io.zig`. Five sites in the same file report a non-type failure as a type error and were left alone. Both groups put something in the "expected type" slot that is not a type.

## 1. Four platform gates blame the argument for the platform

`openInputFile:711`, `openOutputFile:733`, `fdToPort:780`, and `deleteFile:1835` all read:

```zig
if (comptime is_wasm) return primitives.typeError("open-input-file", "non-WASM platform", args[0]);
```

On a `wasm32-wasi` build:

```
error[KP3002]: type error in 'open-input-file': expected non-WASM platform, got #<string>
error[KP3002]: type error in 'delete-file': expected non-WASM platform, got #<string>
```

The filename is valid, and "non-WASM platform" is not a type. This is `argError` (KP3007) territory — or arguably its own "unsupported on this platform" phrasing, since it is not really about the argument at all.

Reachable only on the WASM tier, which is the playground at kaappi-lang.org — so it is the tier where a confusing message is *most* likely to reach someone who cannot read the source to work out what happened.

## 2. `fd->port` reports a range constraint as a type error

`fdToPort:785`:

```zig
if (fd_i < 3 or fd_i > std.math.maxInt(i32)) return primitives.typeError("fd->port", "socket/pipe file descriptor (> 2)", args[0]);
```

```
error[KP3002]: type error in 'fd->port': expected socket/pipe file descriptor (> 2), got 0
```

`0` is a fixnum, which is the type this procedure wants; the fault is that it is out of the accepted range. Same shape as the `set-port-position!` case #1944 moved to `indexError`. Whether this is better as `argError` ("descriptors 0-2 are the standard streams") or `indexError` is a judgment call worth making explicitly — the lower bound is a semantic rule about stdio, not an array length, which argues for `argError`.

## Internal control

After #1944 the file has **6 `indexError` and 9 `argError` sites**, so the machinery and the house phrasing are both present and these five skip them. #1944's own control — a genuine non-port still reporting `KP3002` — stays valid either way, since neither group here is about a wrong type.

[...]

PR #2366, opened 2026-08-26
io: gate fd->port off on WASM like its ffi-* siblings (#2018)

### LINKED ISSUE #2018: fd->port is registered on WASM but can never work there — the only `.kaappi_ffi` spec without `.wasm = false`
**Category:** Tier divergence / registration hygiene · **Found by:** follow-on from #1972 (PR #2016), which fixed the *message* this gate produces without touching why the gate is reachable at all

`fd->port` is the only spec tagged `.kaappi_ffi` that does not set `.wasm = false`.

[`src/primitives_io.zig:68`](https://github.com/kaappi/kaappi/blob/main/src/primitives_io.zig#L68):

```zig
.{ .name = "fd->port", .func = &fdToPort, .arity = .{ .exact = 1 }, .libs = LS.initOne(.kaappi_ffi), .sandbox = false },
```

Its seven siblings in [`src/primitives_ffi.zig:14-20`](https://github.com/kaappi/kaappi/blob/main/src/primitives_ffi.zig#L14-L20) — `ffi-open`, `ffi-fn`, `ffi-close`, `ffi-callback`, `ffi-callback-release`, `ffi-callback?`, `ffi-bytevector-ptr` — all carry `.wasm = false`. `PrimSpec.wasm` defaults to `true`, so `fd->port` is registered in `vm.globals` on the WASM build while the other seven are not.

Meanwhile [`primitives.Lib.wasmAvailable`](https://github.com/kaappi/kaappi/blob/main/src/primitives.zig#L217-L222) lists `.kaappi_ffi` as unavailable on WASM. So the library cannot be imported, but the procedure still resolves through the `vm.globals` fallback:

```
$ wasmtime run --dir=. zig-out/bin/kaappi.wasm f.scm
f.scm:1: error[KP2001]: library not found: (kaappi.ffi)
fd->port: raw file descriptors are unavailable in this WebAssembly build
```

That second line is #1972's fix doing its job — the message is now honest. The question this issue raises is whether the name should be there to answer at all.

## Why `fd->port` is not the same as its `wasmAvailable = false` neighbours

`.srfi_18` and `.srfi_192` are in that same list and have the same *shape* — library unimportable, primitives registered as globals — so at a glance this looks like an established pattern that `fd->port` merely joins. It is not the same thing, and this is the trap worth naming before anyone "fixes" all three together.

For those two, the registered globals **work**:

```scheme
(define p (open-input-string "abc"))
(display (port-position p))   ;; => 0     (srfi 192)
(thread-sleep! 0)             ;; => ok    (srfi 18)
```

Both verified on a `zig build wasm` artifact under wasmtime. For SRFI 18 this is deliberate and already documented in `CLAUDE.md`: `thread-sleep!` is "the one SRFI-18 primitive registered on WASM, as a global; the `(srfi 18)` library itself stays native-only", because the OS-thread machinery cannot exist there but the fiber-safe subset can. The library is native-only; the WASM-viable subset stays reachable.

`fd->port` has no such subset. Its gate is unconditional:

```zig
if (comptime is_wasm) return primitives.argError("fd->port", "raw file descriptors are unavailable in this WebAssembly build", .{});
```

It is the one member of the pattern whose registered global exists **solely to refuse**. Nothing behind it can ever succeed on that target.

## Options

1. **Add `.wasm = false`**, matching its seven siblings. `(fd->port 5)` on WASM becomes `KP3001 undefined variable`, which is structurally honest — the procedure genuinely does not exist there — and consistent with `ffi-open` et al. The `comptime is_wasm` gate in `fdToPort` then becomes unreachable.

   Before deleting that gate, check whether `zig build wasm` still succeeds without it. A comptime-true early return is what keeps subsequent POSIX code out of the WASM build's analysis, and `primitives_srfi18.zig`'s `threadStartImpl` guard is documented as exactly that: unreachable, but kept because the spelling is load-bearing for comptime analysis. The same may hold here.

2. **Leave registration as is and document why** in the spec's comment — i.e. decide that a name that explains itself beats a name that is missing. This is defensible; #1972 is what made the refusal informative in the first place.

## Not in scope, but noticed while checking

`(import (srfi 192))` fails on WASM with `KP2001 library not found` even though `port-position` and friends work perfectly there through the globals fallback. Unlike SRFI 18, there is no documented reason for `.srfi_192` to be in `wasmAvailable`'s exclusion list. That looks like a separate, real gap — worth its own issue rather than being folded into this one.

## If option 1 is chosen

`tests/wasm/platform-gates.scm` has two assertions that expect the current KP3007 message — the `fd->port` case and its `CONTROL: fd->port still type-checks its argument first` companion. Both need updating, and the file is run by the `wasm` CI job:

```bash
wasmtime run --dir=. zig-out/bin/kaappi.wasm tests/wasm/platform-gates.scm
```

**Verified:** against `d0d03deb` (main), with a fresh `zig build wasm` artifact run under wasmtime 46. Spec counts and the `wasmAvailable` list from the merged source; the `port-position` / `thread-sleep!` results are from a real run, not inferred.

baijum:
## Summary

`fd->port` was the only `.kaappi_ffi` spec without `.wasm = false`. Because `PrimSpec.wasm` defaults to `true`, it was registered in `vm.globals` on the `wasm32-wasi` build while its seven siblings (`ffi-open`, `ffi-fn`, `ffi-close`, `ffi-callback`, `ffi-callback-release`, `ffi-callback?`, `ffi-bytevector-ptr`) were not. The `(kaappi ffi)` library is itself `wasmAvailable = false`, and `fdToPort`'s `comptime is_wasm` gate can only refuse — so the registered global existed solely to say "no".

Unlike SRFI 18 (whose OS-thread library is WASM-unavailable but keeps the fiber-safe `thread-sleep!` as a working global), `(kaappi ffi)` has **no** WASM-viable subset. The honest state is that the name does not exist there, which is what `.wasm = false` gives — `(fd->port …)` becomes an undefined-variable fault, consistent with the siblings.

## Changes

- `src/primitives_io.zig`: add `.wasm = false` to the `fd->port` spec.
- `src/primitives.zig`: comptime assertion that **every** `.kaappi_ffi` spec sets `.wasm = false`. Comptime blocks run on every target, so the native build alone now fails to compile if a future FFI primitive reintroduces the divergence.
- `tests/wasm/platform-gates.scm` (run under wasmtime in the `wasm` CI job): the two `fd->port` assertions now check that the name is undefined on WASM rather than a refusing global.

The in-body `comptime is_wasm` gate in `fdToPort` is kept intentionally: `.wasm = false` only skips *runtime registration*, so `fdToPort` is still compiled into the wasm binary and the gate remains load-bearing for comptime analysis (mirrors `primitives_srfi18.zig`'s `threadStartImpl` guard).

## Verification

- `zig build` and `zig build test` (native): pass — 1841 tests. The successful native build already proves the comptime invariant holds (fd->port has `.wasm = false`).
- `zig build test -Dtarget=wasm32-wasi`: compiles (`unit-tests.wasm` builds).
- `zig build wasm` + `wasmtime run zig-out/bin/kaappi.wasm tests/wasm/platform-gates.scm`: all 9 platform-gate checks pass, including the new "fd->port is not registered on WASM" assertion.
- Native `fd->port` still registered and type-checks (`type error in 'fd->port': expected file descriptor, got x`).

Note: `zig build -Dtarget=wasm32-wasi` (the ReleaseSafe exe) fails locally on a pre-existing LLVM "Cannot select" float-constant bug in `primitives_numeric.angleFn` (reproduces on clean `origin/main`, unrelated to this change; `zig build wasm` uses ReleaseSmall and is unaffected).

Closes #2018

### DIFF
--- a/src/primitives.zig
+++ b/src/primitives.zig
@@ -385,6 +385,18 @@ comptime {
                     lib.canonicalName() ++ "; use `.internal` (see Lib.internal) or a *_primitives sub-library");
         }
     }
+    // Every `(kaappi ffi)` primitive is native-only. Unlike SRFI 18 — whose
+    // OS-thread library is WASM-unavailable yet keeps the fiber-safe
+    // `thread-sleep!` registered as a working global — (kaappi ffi) has no
+    // WASM-viable subset: nothing behind these names can ever succeed on
+    // wasm32-wasi. A spec that omits `.wasm = false` would register a global
+    // that exists solely to refuse (kaappi#2018), so require it here rather
+    // than let the next FFI primitive silently reintroduce the divergence.
+    for (all_specs) |spec| {
+        if (spec.libs.contains(.kaappi_ffi) and spec.wasm)
+            @compileError("`.kaappi_ffi` spec \"" ++ spec.name ++
+                "\" must set `.wasm = false`: (kaappi ffi) has no WASM-viable subset (kaappi#2018)");
+    }
 }
 
 pub fn registerAll(vm: *vm_mod.VM) !void {
--- a/src/primitives_io.zig
+++ b/src/primitives_io.zig
@@ -65,7 +65,7 @@ pub const specs = [_]primitives.PrimSpec{
     // already imports — so a raw fd from an FFI call can be given reactor-
     // integrated non-blocking I/O (#1478). Not in sandbox: it is a raw
     // capability over an arbitrary descriptor.
-    .{ .name = "fd->port", .func = &fdToPort, .arity = .{ .exact = 1 }, .libs = LS.initOne(.kaappi_ffi), .sandbox = false },
+    .{ .name = "fd->port", .func = &fdToPort, .arity = .{ .exact = 1 }, .libs = LS.initOne(.kaappi_ffi), .sandbox = false, .wasm = false },
 };
 
 // ---------------------------------------------------------------------------
--- a/src/tests_libraries.zig
+++ b/src/tests_libraries.zig
@@ -698,6 +698,12 @@ test "every spec name resolves in globals (drift guard)" {
     defer vm.deinit();
 
     for (&primitives_mod.all_specs) |spec| {
+        // A `.wasm = false` spec is deliberately not registered on wasm32-wasi
+        // — registerAll gates it with `if (!is_wasm or spec.wasm)`, so on this
+        // target its absence is the rule, not drift (kaappi#2018).
+        if (comptime platform.is_wasm) {
+            if (!spec.wasm) continue;
+        }
         // The helpers vm_bootstrap.install() removes from globals after the
         // bootstrapped closures capture them (#1375) must NOT resolve —
         // calling %push-wind or a %promise-* mutator out of sequence
--- a/tests/wasm/platform-gates.scm
+++ b/tests/wasm/platform-gates.scm
@@ -23,6 +23,10 @@
   (newline)
   (if (not ok) (set! failures (+ failures 1))))
 
+(define (starts-with? s prefix)
+  (and (>= (string-length s) (string-length prefix))
+       (string=? (substring s 0 (string-length prefix)) prefix)))
+
 (define (classify thunk)
   (guard (e (#t (list (cond ((file-error? e) 'file-error)
                             ((error-object? e) 'error)
@@ -69,14 +73,19 @@
               (open-input-file "gate.txt"))))
 
 ;; --- fd->port ---------------------------------------------------------------
-;; A Kaappi extension over (kaappi ffi) — which is itself unavailable on WASM,
-;; though the primitive stays registered as a global, so the gate is reachable.
-;; No file is involved and nothing here satisfies file-error?, so this one is an
-;; argument fault (KP3007) rather than a borrowed file error.
+;; A Kaappi extension over (kaappi ffi), which has no WASM-viable subset — so,
+;; unlike SRFI 18's thread-sleep!, fd->port carries `.wasm = false` like its
+;; seven ffi-* siblings and is not registered at all on this target
+;; (kaappi#2018). The name genuinely does not exist here, so a reference is an
+;; undefined-variable fault, not a global that exists solely to refuse. (A
+;; "Did you mean …?" tail may follow, so match the prefix rather than the
+;; whole message.)
 
-(check "fd->port reports the platform, not the descriptor"
-       (equal? (classify (lambda () (fd->port 5)))
-               '(error "fd->port: raw file descriptors are unavailable in this WebAssembly build" ())))
+(check "fd->port is not registered on WASM (undefined, not a refusing global)"
+       (let ((c (classify (lambda () (fd->port 5)))))
+         (and (eq? (car c) 'error)
+              (string? (cadr c))
+              (starts-with? (cadr c) "undefined variable 'fd->port'"))))
 
 ;; --- controls ---------------------------------------------------------------
 ;; Without these the fix could have flattened every failure into a file error
@@ -85,10 +94,6 @@
 (check "CONTROL: a non-string argument is still a type error"
        (equal? (classify (lambda () (open-input-file 42)))
                '(error "type error in 'open-input-file': expected string, got 42" ())))
-
-(check "CONTROL: fd->port still type-checks its argument first"
-       (equal? (classify (lambda () (fd->port 'not-a-fixnum)))
-               '(error "type error in 'fd->port': expected file descriptor, got not-a-fixnum" ())))
 
 ;; file-exists? is the one filesystem procedure with a value to degrade to, so
 ;; it answers #f rather than raising anything at all — deliberately unlike the


src/primitives_io.zig
```diff
@@ -65,7 +65,7 @@ pub const specs = [_]primitives.PrimSpec{
     // already imports — so a raw fd from an FFI call can be given reactor-
     // integrated non-blocking I/O (#1478). Not in sandbox: it is a raw
     // capability over an arbitrary descriptor.
-    .{ .name = "fd->port", .func = &fdToPort, .arity = .{ .exact = 1 }, .libs = LS.initOne(.kaappi_ffi), .sandbox = false },
+    .{ .name = "fd->port", .func = &fdToPort, .arity = .{ .exact = 1 }, .libs = LS.initOne(.kaappi_ffi), .sandbox = false, .wasm = false },
```
baijum, line 68: Verified from the code: this was the only `.kaappi_ffi` spec relying on the `wasm = true` default — all seven `ffi-*` siblings set `.wasm = false` — and `(kaappi ffi)` itself is never registered on WASM (`library.zig` gates on `wasmAvailable()` at both sites), so `vm.globals` was indeed the only leaky surface. Nothing in the engine synthesizes `fd->port` references, so unregistration breaks nothing else.

Supporting data point: `tests/scheme/differential/run-wasm-differential.sh:319`'s `PLATFORM_SIGNATURE` has listed `undefined variable .(…|fd->port).` since #2122 — the WASM tier now behaves the way the differential runner already classified.

Keeping the in-body `comptime is_wasm` gate in `fdToPort` is right too: the spec table still references `&fdToPort` on the WASM target, so the body is compiled even though unregistered, and the gate makes a hypothetical future re-registration refuse loudly instead of half-working.

baijum (commented):
Reviewed against the code rather than the description — the change is correct, and every claim checks out:

- `fd->port` was the only `.kaappi_ffi` spec on the `wasm = true` default; all seven `ffi-*` siblings set `.wasm = false`, and `(kaappi ffi)` is itself never registered on WASM (`library.zig` gates both registration sites on `wasmAvailable()`), so `vm.globals` really was the only surface where the name leaked through.
- The comptime assertion is complete: on the WASM target `primitives_ffi.specs` is swapped for `no_specs` in `all_specs`, but comptime blocks analyze on every target, so the native build alone enforces the invariant for all eight specs.
- The drift-guard exemption is necessary, not cosmetic — the wasm CI leg executes `unit-tests.wasm` under wasmtime (kaappi#2153), `primitives_io.specs` is in `all_specs` on every target, and without the skip that leg would fail its own drift guard on this change.
- Nothing else on the WASI leg breaks: both fd->port Zig tests (`tests_port_io.zig:572`/`:602`) are already comptime-gated `SkipZigTest`.
- The undefined-variable fault is raised at runtime (`raiseUndefinedVariable` in `vm_dispatch_helpers.zig`), so the guard-based check is structurally sound, and prefix-matching is the right call given the optional `. Did you mean '…'?` suffix.
- Nice confirmation: `tests/scheme/differential/run-wasm-differential.sh`'s `PLATFORM_SIGNATURE` has expected `undefined variable 'fd->port'` since #2122 — this lands exactly the behavior the differential runner already classified as platform-gated.

The only findings are documentation/comment drift (inline below): the #1972 section in `docs/dev/gc-safety-and-error-handling.md`, the cross-reference at `tests/scheme/audit/primitives_io-audit.scm:757`, and this test file's header all still describe fd->port's WASM refusal as reachable. All nits — nothing blocking.

src/primitives.zig
```diff
@@ -385,6 +385,18 @@ comptime {
                     lib.canonicalName() ++ "; use `.internal` (see Lib.internal) or a *_primitives sub-library");
         }
     }
+    // Every `(kaappi ffi)` primitive is native-only. Unlike SRFI 18 — whose
```
baijum, line 388: The assertion is sound and complete. One subtlety a future reader may trip on: on the WASM target `primitives_ffi.specs` is swapped for `no_specs` in `all_specs`, so those seven specs are only ever checked by the native build — fine, since comptime blocks analyze on every target and the native build must compile, but it's the answer to "how do the ffi-* specs pass this on wasm?" and lives three screens away in the `all_specs` definition. A half-sentence here pointing at the `no_specs` swap would save that search.

Follow-up this change owes the docs (nit, non-blocking): the #1972 section of `docs/dev/gc-safety-and-error-handling.md` still documents the platform-gate split with the row `fd->port → argError (KP3007)` and closes with "They are covered by `tests/wasm/platform-gates.scm`" — both now describe an unreachable state: on WASM the name is an undefined variable, and the gate message is asserted nowhere (platform-gates.scm now asserts the *absence*). The table row should become "not registered — the name does not exist on WASM", and the closing paragraph needs the fd->port carve-out. Same staleness at `tests/scheme/audit/primitives_io-audit.scm:757`, whose "tests/wasm/platform-gates.scm covers those under wasmtime" is no longer true for fd->port's gate specifically.

src/tests_libraries.zig
```diff
@@ -698,6 +698,12 @@ test "every spec name resolves in globals (drift guard)" {
     defer vm.deinit();
 
     for (&primitives_mod.all_specs) |spec| {
+        // A `.wasm = false` spec is deliberately not registered on wasm32-wasi
```
baijum, line 701: This exemption is load-bearing rather than tidy: the wasm CI leg executes `unit-tests.wasm` under wasmtime (kaappi#2153), `primitives_io.specs` is in `all_specs` on every target, and `registerAll` now skips `fd->port` on WASM — without this `continue` the drift guard itself would flag fd->port as drift on that leg. Mirroring `registerAll`'s exact condition (`!is_wasm or spec.wasm`) is the right shape: the test keeps asserting everything that should register, on both tiers.

tests/wasm/platform-gates.scm
```diff
@@ -69,14 +73,19 @@
...
-               '(error "fd->port: raw file descriptors are unavailable in this WebAssembly build" ())))
+;; A Kaappi extension over (kaappi ffi), which has no WASM-viable subset — so,
+;; unlike SRFI 18's thread-sleep!, fd->port carries `.wasm = false` like its
+;; seven ffi-* siblings and is not registered at all on this target
+;; (kaappi#2018). The name genuinely does not exist here, so a reference is an
+;; undefined-variable fault, not a global that exists solely to refuse. (A
+;; "Did you mean …?" tail may follow, so match the prefix rather than the
+;; whole message.)
+
+(check "fd->port is not registered on WASM (undefined, not a refusing global)"
```
baijum, line 84: This is a genuine regression test for the change — on main under wasmtime `(fd->port 5)` returns the KP3007 refusal, so the prefix match fails there and passes here. Matching a prefix rather than the whole message is correct: `raiseUndefinedVariable` (`vm_dispatch_helpers.zig`) appends `. Did you mean '…'?` whenever `findSimilarName` produces a suggestion, and which name it suggests can drift independently of this file.

nit: the file header still opens with "a handful of procedures stay registered but cannot do their job" — present-tense framing this change makes untrue for fd->port. The section comment right above explains it well; one sentence in the header distinguishing the three refusing `(scheme file)` globals from the one ffi name that doesn't exist would keep it accurate.

Issue #2200, opened 2026-08-03
[Tooling] A compiled artifact replays top-level define-values before every compiled form, so it can fail where the interpreter succeeds


baijum:
**Category:** Tooling / tier divergence
**Spec reference:** `docs/dev/cache.md` § "What is and isn't cached" — the `.sbc`
preamble; R7RS 5.1 (top-level forms are evaluated in order).
**Severity:** wrong-result (a valid program's standalone binary exits 1)

## Summary

`kaappi --compile` records every top-level form claimed by
`vm_eval.handleTopLevelForm` into the artifact's **preamble**, and the artifact
replays the whole preamble *before* any compiled form. For the five
`TopLevelHead.isEnvSetup()` declarations that is harmless — hoisting them is
what a preamble is for. `define-values` is different: its producer is arbitrary
program code that can depend on earlier forms, so hoisting it changes results
and can fail outright.

## Minimal reproduction

```scheme
(define x 1)
(define-values (a b) (values x 2))
(display (list a b))
(newline)
```

```console
$ kaappi dvdep.scm
(1 2)
$ echo $?
0

$ kaappi --compile -o dvdep.sbc dvdep.scm
$ zig build -Dbundle=dvdep.sbc -p /tmp/b && /tmp/b/bin/kaappi
preamble error[KP3001]: undefined variable 'x'. Did you mean '+'?
error[KP3001]: undefined variable 'a'. Did you mean '+'?
$ echo $?
1
```

The second error is the knock-on: `a` was never bound because the preamble form
raised.

## Discriminating control

Ordering, not the head itself. With no dependency on an earlier form, only the
*order of effects* differs:

```scheme
(display "one")(newline)
(define-values (a) (values (begin (display "producer") (newline) 1)))
(display a)(newline)
```

| | output |
|---|---|
| interpreter | `one` / `producer` / `1` |
| standalone binary | `producer` / `one` / `1` |

## Cause

`src/main.zig`'s `compileFile` appends the form's source to `preamble` for every
classified head; `bytecode_file_write.writeFileWithBundle` stores the preamble as
its own section, and the bundle runner (`src/main.zig`, the `loaded.preamble`
loop) evaluates all of it before executing `loaded.funcs`.

`begin` and `cond-expand` had the same defect until
[#2156](https://github.com/kaappi/kaappi/issues/2156), which fixed them by
*splicing* — their bodies now go into the compiled stream and keep their
position. That is not available for `define-values`: it has no compiler form, so
[...]

PR #2370, opened 2026-08-26
Cache .sld libraries and importing programs (#1888)

### LINKED ISSUE #1888: Bytecode cache is inert for any program that imports: `.sld` files are never cached, and importing disables caching of the main file too
## Summary

The `.sbc` bytecode cache never engages for a program that uses `import`. Two
independent gates cause this, and together they mean **523 of the 566 `.scm`
test files (92%) recompile everything they touch on every single run** — as does
essentially every real Kaappi program, since almost all of them import
something.

This surfaced while profiling the test suite (the parallel-execution work on
branch `claude/kaappi-test-performance-kkoo2l`). Parallelism divides this cost
across cores but does not remove it; caching would remove it.

## The two gates

**1. `.sld` library files are never cached, in either direction** —
`src/vm_library.zig:516`:

```zig
// No .sbc caching for .sld files, in either direction. Writing was
// disabled because collected function pointers can be freed by GC during
// library loading (use-after-free in the serializer). The old cache-read
// path reconstructed the export table by re-parsing the .sld top level,
// silently dropping exports from include-library-declarations and
// cond-expand — if caching is ever reintroduced, serialize the export
// table into the .sbc instead of re-deriving it from source.
```

**2. A main file that imports anything is never cached** —
`src/main.zig:829`:

```zig
// Cache compiled bytecode (skip when imports are used — GC may have freed
// collected function pointers during library loading). Best-effort: a
// failed write (read-only home, etc.) just means the next run recompiles.
if (!has_imports and compiled_funcs.items.len > 0) {
```

Note both comments cite the *same* root cause: collected `Function` pointers can
be freed by GC during library loading, so the serializer walks freed memory.
Gate 2 looks like a conservative blanket workaround for that hazard rather than
a separate design decision.

## Evidence

`--timings` reports the miss reason directly:

```console
$ kaappi --timings=text bare.scm          # (display 1), no imports
timings: read 0.0ms | expand 0.0ms | lower 0.0ms | optimize 0.0ms | emit 0.0ms | execute 0.0ms
cache: MISS (wrote /tmp/.../a29a96428d339e1f.sbc)

$ kaappi --timings=text i64.scm           # (import (scheme base) (srfi 64))
timings: read 0.0ms | expand 2.6ms | lower 0.7ms | optimize 0.1ms | emit 2.7ms | execute 26.8ms
cache: MISS (not cached: imports)

$ kaappi --timings=text i148.scm          # (import (scheme base) (srfi 148))
timings: read 0.0ms | expand 39.9ms | lower 0.1ms | optimize 0.0ms | emit 12.8ms | execute 95.8ms
cache: MISS (not cached: imports)
```

Library compilation lands in `execute` because imports are processed at run
time. Bare startup is ~7ms wall; `(import (srfi 64))` takes it to ~38ms, i.e.
**~30ms per process to recompile one 712-line `.sld`**. SRFI 148 (1285 lines)
costs ~148ms per process.

Prevalence in `tests/scheme/{smoke,compliance,continuations,hygiene,srfi,ffi,audit}`:

| Measure | Count |
|---|---|
| `.scm` test files | 566 |
| …that use `import` (never cached) | **523** |
| …that import `(srfi 64)` | 333 |
| Sampled smoke files rejected from cache, per `--timings` | 51 of 60 |

Lower bound from `(srfi 64)` alone: 333 × ~30ms ≈ **10s per full suite run**,
recompiling one identical library 333 times. Every other imported library adds
to that. Total measured work for the whole `.scm` corpus is ~50s, so library
recompilation is a large fraction of it.

The effect is not test-specific. Any user program that imports a library pays
the same cost on every run, and `kaappi --timings` will report
`cache: MISS (not cached: imports)` forever — the cache silently does nothing
for the overwhelming majority of real programs.

## Why this is not a quick fix

The disabled code paths have two documented hazards that a naive re-enable would
reintroduce:

1. **GC use-after-free in the serializer.** Collected `Function` pointers can be
   freed during library loading, so anything writing them out afterwards may walk
   freed memory. Whatever fixes this needs to root the collected functions for
   the duration, or collect them differently.
2. **Export-table correctness.** The old cache-*read* path re-derived a library's
   export table by re-parsing its `.sld` top level, which silently dropped
   exports coming from `include-library-declarations` and `cond-expand`. The
   comment states the requirement plainly: serialize the export table into the
   `.sbc` rather than re-deriving it.

Both are real and both are why this is filed rather than fixed in passing.

## Suggested approach

- Serialize the resolved export table into the `.sbc` (hazard 2), rather than
  reconstructing it from source on load.
- Root or copy collected `Function` pointers so the serializer cannot observe
  freed objects (hazard 1) — then re-enable `.sld` writing.
- Re-examine gate 2 (`!has_imports`) once hazard 1 is fixed; it likely exists
  only because of it, and lifting it is what makes caching benefit whole
  programs rather than just import-free scripts.
- Cache keys must fold in each imported library's own source hash, or a program
  will happily run against a stale library. `docs/dev/cache.md` documents the
  existing source+compiler key; this adds a transitive dimension.
- `cond-expand` inside a library makes its compiled form depend on the feature
  set, so the key needs to cover that too (or such libraries stay uncached).

## Verification when done

- `kaappi --timings=text` on a file that imports a library reports a cache
  `HIT` on the second run, and stays a `MISS` after editing either the program
  or any library in its import closure.
- A library whose exports come from `include-library-declarations` or
  `cond-expand` exports exactly the same set warm as cold — this is the specific
  regression the old read path had, so it wants a dedicated test.
- The suite stays green under `-Dgc-stress=true`, which is where the
  use-after-free hazard would show up.

## Environment

- Measured on Linux x86_64, 4 cores, ReleaseSafe, Zig 0.16.0, at v0.22.1
  (`40f04a6`).

baijum:
Closes #1888.

## The problem

The `.sbc` bytecode cache never engaged for a program that imports. Two gates caused it: `.sld` library files were never cached in either direction, and any main file containing one of the eight top-level declaration heads was refused outright — so **523 of 566 `.scm` test files (92%) recompiled everything they touched on every run**, and so did essentially every real program.

## The design: structure from source, code from cache

The issue's design analysis framed a fork — serialize the export table (closures over `lib_env`, record-type descriptors, macro transformers: a large net-new serialization surface) vs. replay-and-recompile (loses the macro-compile savings that are the whole point). This takes the road between: **a library entry stores the compiled body functions, the `define-syntax` transformers as data, and an ordered event log; a HIT re-parses the (hash-validated) `.sld` and walks its declarations through the ordinary loader, but replays cached code wherever the cold path compiled.**

- Imports really load (dependencies hit their own entries), export lists are re-derived by name, `cond-expand` re-selects, `define-record-type` runs as data — the *structure*.
- Body functions are **executed**, not serialized as values: closures capture the reconstructed `lib_env`, record types / hash tables / every other runtime value are created exactly as cold, and the export table comes out of the normal export-name lookup — so exports contributed by `include-library-declarations` or `cond-expand` are present warm for the same reason they are present cold (the specific regression the old pre-#1888 cache-read path had).
- The two hazards that forced caching off are addressed: collected functions are rooted in `gc.extra_roots` for the load's duration (the serializer can no longer walk freed memory), and the export-table reconstruction hazard is gone by construction.

## Gate 2 lifted positionally

Main files no longer refuse the eight top-level heads. Each becomes a **positional declaration slot** — the verbatim source span, re-dispatched through `handleTopLevelForm` on a HIT, in exact top-level order. An `(import ...)` between two defines stays between them (the #2200 reorder class cannot arise — no preamble hoisting).

## Invalidation

Layered on the existing key (resolved path, source hash, compiler hash): each library entry also records its include files (path + content hash) and file-backed dependencies (relative path, resolved path, content hash), re-validated before a warm replay starts. Editing a dependency stales every entry that transitively imported it; a `--lib-path` change that re-resolves a dependency elsewhere misses. Libraries whose `cond-expand` consulted *library availability* (`(library …)`, `srfi-<n>` feature ids) are never cached — that answer depends on the live registry, not on anything a key can hash (platform features are compile-time constants already in the compiler hash).

## The surprise finding: the reader, not the compiler

Profiling the warm path showed re-*parsing* the `.sld` dominated, not recompiling: `lineColAt` rescanned from byte 0 per datum, so one large nested datum (every `.sld`) cost **O(n²)**. A monotone line/col cursor fixes it for cold and warm alike — cold srfi-64 loads drop 27.5ms → 11.4ms and warm runs to 8.5ms on this machine (macOS arm64, ReleaseSafe).

## Format and safety

- `.sbc` v13: entry-kind byte (program / bundle / library), program slot section, library transformer/event/include/dependency sections, `TAG_CLOSURE`. Older entries read back as ordinary misses; golden/endian fixtures updated deliberately.
- `kaappi check` / the LSP never write library entries (analysis-mode placeholder transformers must not poison one); `--sandbox`, `--no-ir-opt`, WASM, and embedded/bundled loads stay outside the cache, mirroring the program cache's gates.
- GC discipline: collected functions explicitly rooted (then removed *by pointer*, never by truncation, so `rootedSlot`'s permanent entries survive); deserialized transformers rooted until the replay registers them; `lib_env` keys interned so no slice outlives its entry.

## Tests

- `tests/scheme/cache/library-cache-1888.sh` (new, 22 checks): the issue's verification checklist — HIT on second run; program / library / include / dependency edits each miss appropriately; `--lib-path` re-resolution misses; export-set completeness (include + cond-expand + macro) cold vs warm; exit-code parity; inline `define-library` slots.
- Differential corpus: cache exercised goes from **40/345 files to 354/368**, cold==warm byte-identical on all 368, tier (b) unchanged.
- Unit + gc-stress suites green; full `run-all.sh` green (2115 pass / 0 fail); `zig build wasm` compiles; regenerated `fuzz-seed.sbc` for v13.
- `cache-decline-reason-2114.sh` and `timings-1515.sh` updated to pin the new contract (the eight heads now cache; the remaining refusals are `define-syntax` and `compile error`).
- Docs: `docs/dev/cache.md` rewritten for the new contract; `docs/dev/timings.md` documents the `libcache:` line.


src/main.zig
```diff
@@ -947,24 +1033,25 @@ fn runFile(vm: *vm_mod.VM, path: []const u8) !void {
...
+    // (kaappi#1888): each becomes a declaration slot — its verbatim source
+    // span, replayed positionally through the same dispatch on a HIT — while
+    // `import`/`define-library`/`include` inside them hit the per-.sld library
+    // cache. Best-effort: a failed write (read-only home, etc.) just means the
+    // next run recompiles.
+    if (!defines_syntax and !had_compile_error and (compiled_funcs.items.len > 0 or slots.items.len > 0)) {
         if (sbc_path) |sp| {
             cache.ensureDir();
-            if (bytecode_file.writeFileWithTopLevel(allocator, compiled_funcs.items, source_hash, path, sp)) |_| {
+            if (bytecode_file.writeFileWithSlots(allocator, compiled_funcs.items, slots.items, source_hash, path, sp)) |_| {
```
baijum: **Stale macro expansions replay on HIT: ENTRY_PROGRAM entries carry no dependency/include records.**

`prog.scm` = `(import (lib m)) (display (mac))`. Cold run: the import registers the library's transformer into `vm.macros` (`vm_imports.zig` `importBinding`), the `display` form is compiled with `mac`'s expansion baked in, and the entry is written keyed only on prog.scm's hash. Edit `mac`'s template in `lib/m.sld` and rerun: the library's own entry misses and re-registers the new macro, but the program entry still HITs and replays the compiled slot containing the **old** expansion — silently wrong output, no diagnostic.

Same for `(include "macros.scm")` defining syntax: the registration happens inside `runTopLevelHead`, so the `defines_syntax` guard (line 998) never fires and the include file is not hashed for program entries.

The pre-#1888 eight-head refusal prevented exactly this; #1888's dep tracking (`noteDepLoaded`/`recordsValid`) only covers ENTRY_LIBRARY entries. Program entries need the same include/dependency records — or must decline to cache when the program imports/includes anything that registers syntax.

---

src/vm_library_cache.zig
```diff
@@ -0,0 +1,711 @@
...
+        return VMError.CompileError;
+    }
+
+    timings.libCacheHit();
+    pop(vm);
+    // Drop the deserialize roots: live objects stayed reachable through the
+    // registered library (lib_env values, transformers and their procs); the
+    // executed top-level form wrappers are garbage now, exactly like the cold
+    // path's collected functions after endColdLoad.
+    vm.gc.extra_roots.shrinkRetainingCapacity(roots_base);
```
baijum: **Use-after-free: `endWarmLoad`/`abortWarmLoad` drop roots by truncating `gc.extra_roots` to `roots_base`, discarding unrelated longer-lived roots appended during the library body's execution.**

A library whose body form calls `(thread-start! ...)` at load time: `threadStartFn` (`primitives_srfi18.zig:627`) appends the fiber to `gc.extra_roots` to keep it alive until `thread-join!`. When that library is loaded WARM, the append lands above the `WarmCursor.roots_base` captured in `beginWarmLoad`, and `endWarmLoad`'s `shrinkRetainingCapacity(roots_base)` (also `abortWarmLoad`, line 645) silently removes the running thread's fiber root → the next GC frees the live fiber while its OS thread still writes `fiber.status` → use-after-free/crash.

The module doc itself states roots must be dropped "by pointer identity, never by truncation, so rootedSlot's permanent ... entries interleaved into the same list survive" — `endColdLoad` honors that rule; the warm paths violate it.

---

src/vm_library.zig
```diff
@@ -534,22 +544,53 @@ pub fn tryLoadLibraryFromFile(vm: *VM, name_list: Value) !void {
...
+        if (!cold_ok) return error.UndefinedVariable;
+    }
+
+    // Record this .sld as a dependency of every enclosing load (not of this
+    // one — our own file is the key): an enclosing library's compiled body
+    // embeds expansions of the macros we exported, so it must miss when we
+    // change.
+    if (library_mod.libraryNameToString(allocator, name_list)) |dep_name| {
+        defer allocator.free(dep_name);
+        lcc.noteDepLoaded(vm, rel_path, sld_path, source_hash, dep_name, depth_before);
```
baijum: **Load-order-dependent stale HIT: `noteDepLoaded` never fires for a dependency already registered in `vm.libraries`, so the entry omits that dep.**

`prog.scm` imports `(a)` then `(b)`, where `b.sld` imports `(a)` and uses a macro from it. Cold: `(a)` loads file-backed first; during `(b)`'s cold load, `processImportSet` hits the registry short-circuit (`vm_imports.zig:345` — `if (vm.libraries.get(lib_name)) |lib| ... return`) and never reaches `tryLoadLibraryFromFile`, so `noteDepLoaded` is never called and `(b)`'s entry records **no** dep on `(a)`.

Edit `(a)`'s macro; warm run: `(a)` misses on its own hash and re-registers the new macro, but `(b)`'s `recordsValid` passes (no record of `(a)`) → `(b)` HITs and replays body functions with `(a)`'s old expansions — silently wrong.

`library-cache-1888.sh` only covers the ordering where the dep is first loaded from inside the parent's load; this ordering needs a dep record (e.g. note the dep at the registry short-circuit too, when the registered library is file-backed) and a test.

---

src/main.zig
```diff
@@ -814,31 +843,67 @@ fn runFile(vm: *vm_mod.VM, path: []const u8) !void {
...
+                                timings.end();
+                                try runCachedTopLevelFunc(vm, func, source, path);
+                            }
+                        },
+                    }
+                }
+                return;
+            }
+
+            const top_count = @min(loaded.top_level_count, @as(u32, @intCast(loaded.funcs.len)));
```
baijum: **Kind confusion: `runFile`'s HIT path never checks the entry kind, so a v13 LIBRARY entry read back for a directly-run `.sld` executes library body functions as a program.**

`kaappi prog.scm` imports `(user1888 u)` → writes an ENTRY_LIBRARY `.sbc` at `cache.pathForSource(u.sld)`. Then `kaappi u.sld` (running the library file directly — cold correctly interprets the `define-library` and exits 0): `sbc_path` is the same cache key, the source hash matches, deserialize succeeds, `loaded.slots` is null (only ENTRY_PROGRAM gets slots) → the `top_count` fallback loop runs the library's body functions directly with `func.env == null` and no transformer registration/imports → undefined-variable errors / wrong exit code where cold ran cleanly.

Note the fallback's stated purpose ("older-format entries predate slots") is dead code — pre-v13 files are rejected by the VERSION check — so its only reachable use is this misfire. A kind check treating non-PROGRAM entries as a miss fixes both.

---

src/vm_library_cache.zig
```diff
@@ -0,0 +1,711 @@
...
+    if (depth == 0 or depth > vm.lib_cache_stack.len) return; // overflowed frame
+    const c = &vm.lib_cache_stack[depth - 1];
+    const roots_base = if (c.warm) |*w| w.roots_base else vm.gc.extra_roots.items.len;
+
+    if (c.replay_desync or (c.warm != null and c.warm.?.pos != c.warm.?.events.len)) {
+        // Structure and log disagree. The validated source/include/dep
+        // equality makes this near-impossible, so it indicates a
+        // loader/serializer bug — loud, never silent.
+        vm.setErrorDetail("library cache: replay event log desync for {s}", .{sld_path});
+        return VMError.CompileError;
```
baijum, line 679: **Collector-stack leak/desync: `endWarmLoad`'s desync path returns an error without `pop(vm)`.**

A replay desync inside an included form is swallowed by `evalIncludedForm` (`lcc.replayGlobalForm(vm, c) catch ... continue` in `vm_library.zig`), so `loadLibrarySource` returns success with `c.replay_desync` set — reachable e.g. via a TOCTOU edit of an include file between `recordsValid` and the warm structure walk changing the form count. `endWarmLoad` then returns `VMError.CompileError` before `pop(vm)`: the warm frame stays on `lib_cache_stack` and `lib_cache_depth` stays raised.

The enclosing cold load's `endColdLoad` subsequently operates on `stack[depth-1]` — the leaked warm frame, not its own collector — skipping its cache write, and every outer frame shifts by one; the leaked frame is only freed at VM deinit. The error path needs the same pop+root-drop as `abortWarmLoad`.

---

src/vm_library_cache.zig
```diff
@@ -0,0 +1,711 @@
...
+    if (result.library == null) {
+        bf.freeDeserializeResult(allocator, &result);
+        timings.libCacheMiss();
+        return false;
+    }
+    const lib = &result.library.?;
+
+    if (!recordsValid(vm, lib)) {
+        bf.freeDeserializeResult(allocator, &result);
+        timings.libCacheStale();
```
baijum: **Permanently-rooted garbage: `beginWarmLoad`'s miss paths never shrink `gc.extra_roots` back to `roots_base`.**

When a dependency was edited, `recordsValid` fails here (also the `result.library == null` path at line 579 and the `!pushed` path at line 602). `deserializeFromBuffer` had already succeeded, so `keep_roots=true` left every loaded `Function` and `Transformer` rooted in `gc.extra_roots` (`bytecode_file_read.zig:560`/`791`). These paths call `freeDeserializeResult` (freeing only host-side slices) and return `false` without dropping those roots — the GC objects stay rooted for the process lifetime even though the library is then recompiled cold.

A long-running process with several stale libraries pins one full compiled body per stale entry. (Per the module's own rule, drop them by pointer identity, not truncation.)

---

src/reader.zig
```diff
@@ -208,9 +250,11 @@ pub const Reader = struct {
         if (!self.record_spans) return;
         if (!types.isPair(val) and !types.isVector(val)) return;
         const end_pos = self.pos;
-        const end = lineColAt(self.source, end_pos);
-        // A start at or past the end (an empty span) collapses to the end point.
-        const start = if (start_pos >= end_pos) end else lineColAt(self.source, start_pos);
+        // Query start BEFORE end so the monotone cursor only ever advances
+        // (see lineColMonotone). A start at or past the end (an empty span)
+        // collapses to the end point.
+        const start = if (start_pos >= end_pos) self.lineColMonotone(end_pos) else self.lineColMonotone(start_pos);
```
baijum: **Span recording stays quadratic for nested data: any datum containing a nested pair/vector takes `lineColMonotone`'s backward-jump path (a full O(offset) rescan), so the "no caller does this today" comment is false.**

Reading `(define (f x) (+ x 1))` at byte offset p: the children's `recordSpan` calls advance the cursor to their *end* positions first; the enclosing list's `recordSpan` then queries its start p, which is behind the cursor → the "defensive backward jump (no caller does this today)" fallback runs `lineColAt` from byte 0, O(p). Every list with a nested recorded child pays this, so a large `.sld` remains ~O(n·m) in span recording — the O(distance) rationale doesn't hold for the dominant nested shape.

Cheaper: capture the monotone cursor's (pos, line, col) at datum *start* in `readDatumOrEof` (a per-depth save, mirroring the existing depth counter) and pass it to `recordSpan`, instead of re-deriving the start position at datum end.

---

src/main.zig
```diff
@@ -814,31 +843,67 @@ fn runFile(vm: *vm_mod.VM, path: []const u8) !void {
...
+            if (loaded.slots) |slots| {
+                for (slots) |slot| {
+                    switch (slot) {
+                        .function => |idx| {
+                            if (idx >= loaded.top_level_count) continue;
+                            const func = loaded.funcs[idx];
+                            try runCachedTopLevelFunc(vm, func, source, path);
+                        },
+                        .declaration => |decl| {
+                            var dr = reader.Reader.init(vm.gc, decl.src);
```
baijum: **Cold/warm divergence: declaration-slot replay parses each slot with a fresh `Reader`, losing `#!fold-case` / `#!no-fold-case` state carried from earlier top-level forms.**

A program beginning `#!fold-case` followed by `(define x 1)` (a function slot — the directive falls inside *its* span, which is never re-parsed) and later `(IMPORT (SCHEME BASE))`: cold, the single file-wide `Reader` folds the symbols and `handleTopLevelForm` claims the import. Warm, the declaration slot is re-parsed by a fresh `reader.Reader.init(vm.gc, decl.src)` with `fold_case=false`, `IMPORT` is not recognized as a top-level head, falls through to the compile-and-run fallback, and fails at runtime where the cold run succeeded.

Directive state (fold-case at minimum) needs to be recorded per slot or carried across the per-slot readers.

---

src/reader.zig
```diff
@@ -190,11 +194,49 @@ pub const Reader = struct {
         return r;
     }
 
-    /// Compute line and column from the current position by scanning from the
-    /// start of the source. O(n) per call but only used on error paths and
-    /// datum boundaries, not every character advance.
+    /// Compute `getLineCol` at `self.pos`, advancing the reader's monotone
+    /// line/col cursor (see `lineColMonotone`).
+    /// line/col cursor (see `lineColMonotone`).
```
baijum, line 199: Nit: this doc-comment line is duplicated — lines 198–199 are byte-identical. Delete one.

---

baijum (commented):
Reviewed at head `b07448d`. 9 findings, ranked most-severe first as inline comments:

- **Two silent-wrong-answer invalidation gaps**: ENTRY_PROGRAM entries carry no dependency/include records, so stale imported/included macro expansions replay on HIT (`src/main.zig:1054`); and the `vm.libraries` registry short-circuit means an already-loaded dep is never recorded into a library entry, so staleness is load-order dependent (`src/vm_library.zig:592`).
- **One memory-safety bug**: `endWarmLoad`/`abortWarmLoad` truncate `gc.extra_roots`, stripping e.g. `thread-start!`'s until-join fiber root appended during a warm replay → use-after-free (`src/vm_library_cache.zig:632`).
- **One kind-confusion bug**: `runFile`'s HIT path executes a LIBRARY entry's body functions as a program when a `.sld` is run directly (`src/main.zig:904`).
- Plus: a collector-stack leak/desync on `endWarmLoad`'s error path (`vm_library_cache.zig:623`), permanently-rooted garbage on stale-entry misses (`vm_library_cache.zig:586`), a still-quadratic backward-jump path in the reader's new monotone line/col cursor (`reader.zig:256`), a `#!fold-case` cold/warm divergence in declaration-slot replay (`main.zig:864`), and a duplicated doc line (`reader.zig:199`).

Overall the structure-from-source / code-from-cache design is sound and well-documented; the findings are concentrated in invalidation completeness for the newly-cacheable program entries and in root-lifetime discipline on the warm paths.

---

baijum:
Thank you for the thorough review — all eight findings are legitimate and addressed in 142170d4, each with a regression check added to `tests/scheme/cache/library-cache-1888.sh` (now 34 assertions):

1. **Program-entry staleness** — program entries now carry the same include/dependency records as library entries (the sections are shared in the v13 layout), validated before a HIT replays anything. Verified: edit an imported library's macro → `cache: MISS (wrote …)` for the program with the new expansion. You're right that the old eight-head refusal was the accidental guard here.

2. **Root truncation UAF** — all warm-path root drops (`endWarmLoad` success, desync, `abortWarmLoad`, and the `beginWarmLoad` miss paths — your finding 6) now remove exactly the deserialized funcs + transformers **by pointer**, the same rule `endColdLoad` follows; a fiber root appended by the body's own execution survives. Ordering matters too: roots are dropped *before* the result is freed/popped, since `freeDeserializeResult` resets the pointer fields the walk reads.

3. **Registry short-circuit deps** — file-backed libraries now stamp `Library.source_path`/`source_hash` provenance at load, and `processImportSet`'s registry branch records the dependency into every enclosing collector plus the run recorder. Verified with your exact ordering: main imports `(a)` first, `(b)` imports `(a)` from the registry; editing `a`'s macro stales both `(b)` and the program entry.

4. **Kind confusion on a directly-run `.sld`** — `runFile` now requires `ENTRY_PROGRAM` *and* a slot stream; anything else is an ordinary miss. The "legacy no-slots" fallback is gone (dead code, as you noted — the VERSION check rejects pre-v13).

5. **Desync frame leak** — `endWarmLoad`'s desync path pops (and drops roots by pointer) before returning the error, so a swallowed include-form desync can no longer shift enclosing collectors.

6. **Miss-path root pinning** — folded into the fix for 2: every `beginWarmLoad` fall-through drops the deserialize roots before returning false.

7. **Nested-span quadratic** — the cursor is now advanced to the datum's **start** in `readDatumOrEof` before reading (per-depth save, exactly your suggested shape); `recordSpan` takes the precomputed `(line, col)`, so only the strictly-forward end query remains. The backward path survives as a defensive fallback only.

8. **`#!fold-case` divergence** — declaration slots carry a fold-case flag recorded from the reader state at the form; the per-slot re-parse applies it. Verified with your exact shape (`#!fold-case` … `(IMPORT (SCHEME BASE))` — cold and warm both 8; without the directive it still fails, so the test discriminates).

The include/dependency sections moving to a shared position changed the v13 layout again (no shipped v13 format exists), so the golden/endian fixture and the fuzz seed were regenerated. All suites re-run green: unit, gc-stress, differential (368/368 cold==warm, cache exercised 354/368), full `run-all.sh` (2114 pass / 0 fail), wasm build.

src/bytecode_file_read.zig
```diff
@@ -542,104 +636,333 @@ pub fn deserializeFromBuffer(gc: *GC, data: []const u8, expected_hash: ?u64) !?D
         validateFunctionBytecode(func) catch return null;
     }
 
+    // Program slot section (v13): the positional replay stream.
+    var result: DeserializeResult = .{ .funcs = &.{}, .top_level_count = top_level_count, .entry_kind = entry_kind };
+    var ok = false;
+    // One cleanup funnel for every corrupt-cache `return null` below: frees
+    // whichever sections were already read (kaappi#2113 discipline — nothing
+    // partial leaks past a miss).
+    defer if (!ok) freeDeserializeResult(allocator, &result);
```
baijum, line 658: **New in this commit: the corrupt-file `return null` paths leak — `errdefer` never fires on `return null`.**

`return null` is a *success-value* return of `!?DeserializeResult`, so the `errdefer` blocks in the section readers are skipped, and this funnel only frees sections already **attached** to `result`. Every in-flight section leaks on its own corrupt paths:

- Slots: `if (filled != slot_count) return null` (line 676) leaks `entries` plus every duped `declaration.src` — the old code freed them explicitly here.
- Includes: `const hash = r.readU64() catch return null` (line 704) leaks the just-duped `path`, the `includes` array, and `includes[0..inc_filled]`.
- Deps: same shape — `res_len`/`hash`/`name_len` `catch return null` after `rel`/`resolved` were duped; their `errdefer allocator.free(rel)` also never fires.
- Bundled files and preamble were rewritten into the same pattern (`errdefer` + `catch return null`), where the old code called `freeBundledFiles`/`freePreambleEntries` on each exit.

Consequence: a truncated or corrupt cache entry leaks its partially-read sections on every read attempt (then misses and recompiles). Not unsafe, but it contradicts the "nothing partial leaks past a miss" claim in this comment and the kaappi#2113 discipline.

Simplest fix: inside these sections, replace `catch return null` / `break`-then-`return null` with `return BytecodeError.CorruptedFile` so the `errdefer`s actually fire (the funnel then frees the attached sections) — `readLibrarySections` already works exactly this way. Callers like `readFileWithTopLevel` already `catch null`, so behavior is unchanged; only check whether `cache status` (#2113) distinguishes error from null in its loadability report before collapsing the two.

---

baijum (commented):
Re-reviewed at head `142170d` ("Address review: program-entry staleness, root safety, registry deps"). **8 of the 9 findings are fixed correctly, each with a matching regression test.** Verified in detail:

- **Program-entry staleness** — `.sbc` program entries now carry the shared include/dep sections, fed by the VM-level run recorder (`noteRunDep`/`noteRunInclude`), validated by the shared `recordsValid` before replay. Top-level includes of the main file route to the run recorder via the `lib_cache_depth` split in `openIncludeFile`, which also correctly covers inline `define-library` includes. Test E covers the macro-edit staleness.
- **Registry-order deps** — `Library.source_path`/`source_hash` provenance is stamped at load, freed in `deinit` (including the replace path in `register`), and the `processImportSet` short-circuit records through `noteDepLoadedAllFrames` into every armed frame plus the run recorder. Test F pins the exact ordering from the review.
- **Root safety** — `dropDeserializeRoots` removes by pointer identity on every warm/miss/abort path; I verified its coverage is exact: the deserializer's `keep_roots` pins precisely the functions (all of `result.funcs`) and the transformers, and the writer only serializes transformers reachable from `register_tx` events, so nothing stays pinned. The header-first-field address equality makes the pointer match sound for both types.
- **Kind confusion** — checked in both directions (`runFile` rejects non-PROGRAM/slot-less entries; `beginWarmLoad` rejects non-LIBRARY), with test G for the direct `.sld` run. The alternating direct-run/import scenario now just clobbers the shared cache slot back and forth — correct, merely a cache-thrash curiosity not worth code.
- **`endWarmLoad` desync** now pops before erroring; **miss paths** drop roots via the shared funnel; **reader quadratic** fixed the way suggested (start line/col captured in `readDatumOrEof` before reading children — all call sites updated, and the monotone-cursor comment corrected); **fold-case** carried per slot with test H.

**One new finding** (inline, `bytecode_file_read.zig:645`): the deserializer rework mixes `errdefer` with `catch return null`, but `errdefer` doesn't fire on `return null` — so the corrupt-cache-file paths in the slots/includes/deps/bundled/preamble sections leak their in-flight allocations, which the old explicit-free code did not. Low severity (corrupt entry → bounded leak, then an ordinary miss), but it contradicts the funnel comment's own claim.

Two leftovers, no new comment threads:
- The duplicated doc line at `reader.zig:198–199` from the first review is still present.
- Optional: the v13 layout changed within this PR (declaration `flags` byte, program include/dep sections) without bumping `VERSION`. Fine since v13 never shipped — intermediate-commit entries exist only on dev machines and near-certainly fail the section bounds or the final `r.pos != data.len` check, reading as a miss — but a bump is cheap insurance if any v13 fixture escaped.

With the leak fixed and the doc-line dedup, this looks ready from my side.

---

baijum:
Second round addressed in 28b9e545 — the new inline finding plus the CodeRabbit sweep. Two of the CodeRabbit items turned out to be the tip of real correctness icebergs, so I went past the letter of each finding where the fix demanded it:

**The corrupt-read leak (your finding + CodeRabbit's)** — fixed by making the deserializer tail's corrupt exits real errors (`BytecodeError.CorruptedFile`), so the `errdefer`s and the cleanup funnel both run; a thin wrapper maps non-OOM errors back to a plain miss so every caller (`readFileWithTopLevel`, `cache status`'s dry-run, the fuzz loader) sees identical behavior. `cache status` collapses error and null the same way (`catch return false`), so no distinction was needed there. Two extras fell out of writing the test: the include section's read order stranded its in-flight `path` dupe between the allocation and the assignment that made the errdefer own it (reordered: borrowed reads first, dupe last), and a truncation-sweep unit test now pins leak-freedom — it fails on the pre-fix code.

**CodeRabbit's registry/include items led to a bigger hole**: a program's compiled slots don't just depend on its libraries' `.sld` files — they embed the libraries' macro *expansions*, which can change through any file in a library's transitive closure. Program entries now **inherit each loaded library's own include/dependency records** (folded in where the records exist: `endWarmLoad`, and `endColdLoad` on a successful write), so an include-file edit stales the importing program too. When a dependency is unrecordable — a library that declined caching, or whose entry write failed — the run is **poisoned** and the program entry declines outright (`--timings`: `not cached: uncacheable dependency`) rather than serving stale slots forever. Main-file top-level includes also reach the run recorder now (the structure-depth flag wraps runFile's head dispatch; runtime `eval` includes still record nothing).

**The overflow-depth marker leak** — `beginWarmLoad`'s `!pushed` path pops the marker `push()` left behind; your scenario (9-deep chain shifting every enclosing collector) was exactly right.

**Program/library key collision** — CodeRabbit's `.sld`-directly test hardening exposed that a direct run and an import share one cache key, so each access pattern *clobbered* the other's entry. Library entries moved to their own namespace (`cache.pathForLibrary`); the shell test now warms a side-effecting library and checks the direct run, the re-run, and the importer's HIT separately.

**Also fixed**: `writeTransformer`'s missing per-slice caps (writer-refuses-what-reader-rejects, kaappi#2113); `tests_fuzz`'s `freeLoaded` double-free of bundled/preamble after `freeDeserializeResult`; declaration-only programs (import-only scripts) being uncacheable because the reader rejected `func_count == 0`; the JSON `libcache` reason field; and the unit test now roots its fixture across allocations and executes both event functions through a real VM asserting `5` and `6`.

`library-cache-1888.sh` is at 39 assertions, including the requested `cache: MISS (wrote` on every transitive invalidation and a discriminating direct-`.sld` check (body side effect must appear; a thunk misfire would die on undefined variables instead). All suites green: unit, gc-stress, differential (368/368, cache exercised 354/368), full run-all (2114 / 0), wasm build.

src/vm_library_cache.zig
```diff
@@ -0,0 +1,907 @@
...
+                timings.libCacheWrote();
+                // Record the dependency for the MAIN run's entry, inheriting
+                // this library's own include/dependency records (#1888
+                // review): a program's slots embed this library's macro
+                // expansions, which can change through an edit to any file
+                // in the transitive closure.
+                noteRunDep(vm, rel_path, sld_path, source_hash, lib_name);
+                inheritRunIncludes(vm, c.includes.items);
+                inheritRunDeps(vm, c.deps.items);
+            } else |err| switch (err) {
```
baijum: **The write-failure branch neither records the dep nor poisons the run — the program entry then omits this library entirely and serves stale slots after its edits.**

When `writeFileWithLibrary` fails (`LimitExceeded`, `UnsupportedConstant`, or the `else` arm), this branch only records a timings reason. `noteRunDep`/`inherit*` live in the success arm, `noteRunPoison` only fires on the `pathForLibrary == null` path and the "not cacheable" branch — so here `run_cache_ok` stays true and the run recorder never hears about this library. The registry hook doesn't save the first import either: `processImportSet`'s post-load re-lookup (line 397) imports bindings directly without re-entering the registry branch.

Result: the program entry is written with **no record** of the failed-write library → edit that library's macros → the library itself cold-loads with the new definitions, but the program entry HITs (`recordsValid` passes vacuously) and replays slots with the old expansions — exactly the silent-staleness class this commit closes elsewhere. And this commit widens the trigger: the four new `LimitExceeded` guards in `writeTransformer` turn oversized transformer fields into precisely this failure.

Fix: the run records don't depend on the entry existing — they validate against the filesystem (re-resolve + hash). So on write failure, still run

```zig
noteRunDep(vm, rel_path, sld_path, source_hash, lib_name);
inheritRunIncludes(vm, c.includes.items);
inheritRunDeps(vm, c.deps.items);
```

(`c.bail` is already excluded from this branch, so the collector's records are complete; the bail case correctly poisons via the "not cacheable" branch below.) That keeps program caching enabled and correct; `noteRunPoison(vm)` here is the blunter but also-correct alternative. A test twin of section E with a deliberately oversized transformer would pin it.

---

baijum (commented):
Re-reviewed at head `28b9e54` ("Address second review round: corrupt-read leaks, run-record completeness"; the force-push is a clean rebase — the earlier commits' content is unchanged apart from absorbing #2369 from main). This round both fixed the outstanding finding and closed three gaps beyond what I had flagged. Verified in detail:

**Fixed from last round:**
- **Corrupt-read leaks** — `deserializeFromBuffer` is now a thin wrapper over an `Impl` whose corrupt exits are `BytecodeError.CorruptedFile`, so the `errdefer`s actually fire; the wrapper maps them back to `null` (miss) and preserves `OutOfMemory` for callers. The includes loop also reorders borrowed reads before the dupe so nothing can fail between the allocation and `inc_filled` taking ownership. The new truncation-sweep unit test (every prefix of a real entry, under the testing allocator) pins both the miss and the no-leak property.

**Gaps closed proactively — all correct:**
- **Separate key namespaces** (`pathForLibrary`, `l`-prefixed, distinct seed): a `.sld` run directly and the same file imported no longer clobber each other's entries; the kind checks remain as defense. The upgraded section-G test (a load-time side effect print) now genuinely discriminates cold load / warm library replay / kind-confused program replay.
- **Include-closure inheritance**: a program's entry now inherits each loaded library's own include/dep records (`inheritRunIncludes`/`inheritRunDeps` in `endColdLoad`/`endWarmLoad`), closing a staleness hole I had missed — editing a file *included by* an imported library changes the library's macros without changing its `.sld` hash, so the program's plain dep record alone would have validated stale slots. The registry short-circuit not inheriting is fine: with one `runFile` per process, the first load in the run always records the closure.
- **Poisoning** (`run_cache_ok`): a deliberately-uncached library (availability-dependent `cond-expand`) or one with no cache path now declines the program entry too — right call, since no record could ever validate those expansions.
- Also verified: the `pop(vm)` on `beginWarmLoad`'s `!pushed` path correctly balances `push()`'s overflow depth marker (without it, ≥8-deep import chains would shift every enclosing collector); the `lib_structure_depth` bump in `runFile` is what actually makes main-file top-level includes reach `noteRunInclude` (and it also makes caching of include-defined macros *correct*, since the include file is now hash-validated and the warm replay re-registers through the re-dispatched slot); import-only programs (zero compiled functions) are now readable rather than write-only; the `writeTransformer` limit guards close a writer/reader asymmetry per #2113; and the `tests_fuzz` `freeLoaded` simplification removes what had become a double-free of `bundled_files`/`preamble`.

**One new finding** (inline, `vm_library_cache.zig:866`): the library write-*failure* branch (`LimitExceeded` / `UnsupportedConstant` / `else`) neither records the dep into the run recorder nor poisons the run — so the program entry omits that library entirely and later HITs with stale expansions after the library is edited. The main.zig comment ("could not write its entry") describes the intended behavior; the code only covers the no-path case. The fix is small: record (or poison) in the error arm too.

**Leftover nit**: the duplicated doc line at `reader.zig:198–199` (open thread from round 1) is still present.

With those two, this is an LGTM from my side — the invalidation story is now genuinely closed over programs, libraries, includes, and registry ordering, and the test suite grew a check for every finding raised across the three rounds.

---

baijum:
Round-3 finding fixed in fb71843a, exactly as you suggested — record, don't poison.

`endColdLoad` now records the run dependency and inherits the library's include/dependency closure on a **failed** write too (the records validate against the filesystem, not against an entry, so they are fully usable without one). Poisoning would also have been correct but would have needlessly declined the program entry — an importer of an unserializable library still caches correctly, it just re-validates that library by hash on every warm run.

The test twin is in place (`library-cache-1888.sh`, section I, 44 assertions total): a library whose body carries a literal nested past the `.sbc` depth cap, so its entry write is refused with `library exceeds .sbc limits`. The importer's program entry still HITs warm (dependency recorded, run not poisoned), and editing the library's macro then misses the program entry and runs the new expansion. I used the over-deep literal rather than an oversized transformer field since it reaches the same branch — and as you noted, the new `writeTransformer` caps route into it as well.

All suites re-run green: unit, gc-stress, differential (368/368, cache exercised 354/368), full run-all (2114 / 0).

tests/scheme/cache/library-cache-1888.sh
```diff
@@ -0,0 +1,387 @@
...
+} > "$LIBDIR/deep1888/d.sld"
+P4="$PROGDIR/p4.scm"
+cat > "$P4" <<'SCM'
+(import (deep1888 d) (scheme base))
+(display (mac))
+(newline)
+SCM
+i1="$(run_timed "$P4")"
+check "uncacheable-library program: cold output" "111" "$i1"
+check "uncacheable-library program: library write refused with a reason" "libcache: 0 hits, 1 miss" "$i1"
```
baijum, line 369: Nit (non-blocking): `check` is a substring match, so `"libcache: 0 hits, 1 miss"` also passes when the line reads `... 1 miss, 1 written` — this assertion can't actually tell that the write was refused. If the writer's depth cap is ever raised past 300, the write starts succeeding and section I silently stops exercising the failure path it exists for, while every check still passes. Asserting `"0 written"` (or the reason text `"library exceeds .sbc limits"`) pins it.

---

baijum (commented):
**LGTM.** Re-reviewed at head `fb71843` ("Record run dependencies even when the library entry write fails").

The commit resolves the round-3 finding exactly as proposed: `endColdLoad` now records the run dependency and inherits the library's include/dep closure on **both** arms of the write — the shared `record_run_dep` helper runs on success and on every failure (`LimitExceeded`, `UnsupportedConstant`, and the catch-all), which is correct because the records validate against the filesystem, not against the entry. I verified the collector's records are still intact at both call sites, and that the `c.bail` case stays excluded from this branch (incomplete records correctly poison via the "not cacheable" branch instead of being recorded). The registry re-lookup gap I traced last round (a first import whose write fails records nothing anywhere) is closed by this.

Section I of the shell test pins the scenario end to end: a 300-deep literal forces the write to fail, the program still HITs warm (dependency recorded, not poisoned), and a macro edit in the never-cached library correctly misses the program entry and runs the new expansion — the exact silent-staleness sequence the finding described. One non-blocking nit inline: the "write refused" check is a substring match that would also pass if the write succeeded, so it doesn't guard the scenario's precondition against a future depth-cap change.

**Final state across the four rounds:** all 10 substantive findings are fixed and each carries a regression test — program-entry dependency/include invalidation (with registry-order provenance and include-closure inheritance), the root-truncation use-after-free (by-pointer removal everywhere), the kind confusion (separate key namespaces plus checks both ways), the collector-frame leak on desync and the overflow-marker imbalance, miss-path root pinning, the corrupt-read leak class (errdefer-reachable exits plus the truncation-sweep test), the reader's quadratic span recording, the fold-case divergence, and the failed-write invisibility fixed here. The invalidation story is now closed over programs, libraries, includes, dependency ordering, and uncacheable/unwritable libraries alike.

Only cosmetic leftovers, neither blocking: the duplicated doc line at `reader.zig:198–199` (open thread from round 1), and the substring-match nit above. From my side this is ready — a human approval is still required, since I don't approve or merge.

---

baijum:
Good catch — pinned in 8160a4bd: the check now asserts `libcache: 0 hits, 1 miss (library exceeds .sbc limits)`, so it fails if the write ever starts succeeding (e.g. the depth cap moving past the fixture's 300) instead of passing vacuously. 44/44 green.

baijum:
Correction to my previous comment: the commit is **12b89b59**, not 8160a4bd (I cited it before the hash existed). It landed on the branch right after the merge, so it is not in the PR — it is now up as #2371 (cherry-picked onto main). The PR itself is unaffected: the nit was explicitly non-blocking and the merged fb71843a contains all substantive fixes.

Issue #2196, opened 2026-08-03
[GC] Undeduplicated remembered set makes minor collections quadratic in writes to one container — 3s stalls filling a hash table


baijum:
**Category:** GC / performance
**Severity:** performance cliff (correct results, multi-second stalls)

`GC.writeBarrier` (`src/memory.zig:315`) appends to `remembered_set`
unconditionally, with no membership check:

```zig
if (child.owner == self.id and child.flags.generation == 0) {
    self.remembered_set.append(self.allocator, container) catch @panic(...);
}
```

So a single old-generation container mutated *n* times contributes up to *n*
entries, all the same pointer. `gc_collect.zig:59` then marks every entry, and
marking a hash table is O(capacity) — so a minor collection costs
O(barrier_count x capacity) rather than O(distinct_containers x capacity).

`hash-table-set!` fires the barrier **twice** per insert (key and value), so
filling one table with 150k entries can queue ~300k entries for a table whose
capacity is 262144.

**Minimal reproduction** — 13 fills of a fresh table in one process:

```scheme
(import (scheme base) (scheme write) (scheme time) (srfi 69))
(define (ms t0 t1) (round (/ (* 1000 (- t1 t0)) (jiffies-per-second))))
(define (trial n)
  (let ((t (make-hash-table)) (t0 (current-jiffy)))
    (do ((i 0 (+ i 1))) ((= i n)) (hash-table-set! t (number->string i) i))
    (display n) (display " -> ") (display (ms t0 (current-jiffy))) (display " ms") (newline)))
(for-each trial '(10000 20000 30000 40000 50000 55000 60000 65000 70000 80000 100000 120000 150000))
```

```text
$ kaappi --gc-stats repro.scm
10000 -> 4 ms      55000 -> 2491 ms   <-- 33x the 50000 cost for 1.1x the work
20000 -> 16 ms     60000 -> 76 ms
30000 -> 30 ms     65000 -> 80 ms
40000 -> 47 ms     70000 -> 3113 ms   <-- and again
50000 -> 67 ms     80000 -> 95 ms ... 150000 -> 159 ms

  Collections:       8
  Live objects:      151026 (peak: 415200)
  Mark time:         5463.95 ms total
  Sweep time:        0.00 ms total
```

The 5464 ms of mark time is the whole anomaly (2491 + 3113 ~= 5.6 s of visible
stalls). Sweep is 0.00 ms, so this is the mark phase alone.

**Discriminating controls** (three, all run):

[...]

PR #2372, opened 2026-08-27
Make the minor collection's mark generational

### LINKED ISSUE #1961: A minor collection performs a full transitive mark — only the sweep is generational, so minor and full cost the same to mark
**Category:** GC architecture · **Severity:** performance + latent-correctness framing · **Found by:** systematic audit v2, Phase 7B (#1890)

`minorCollect` calls `clearOldMarks(gc)` then `markRoots(gc)`, and `markValueInner` **never stops at the young/old boundary** — it has zero generation checks. So a minor collection marks the entire live graph. Only `sweepYoung` is generational.

## Verified

```text
minorCollect:  clearOldMarks → markRoots → remembered_set walk → sweepYoung
markValueInner: 0 occurrences of generation / is_old / young
markObjectContents: exactly ONE production caller (gc_collect.zig:59, the remembered-set walk)
```

## Three consequences

**1. Minor and full collections have identical mark cost.** They differ only in whether old objects are swept. The usual reason to run a minor collection — that it touches only the young set — does not hold here. That is a real performance issue on a heap with a large old generation.

**2. A forgotten `writeBarrier` is a conservative-retention miss, not a lost live object.** This is worth stating explicitly because it changes how urgent barrier bugs are: an old→young store with no remembered-set entry still has its referent marked, via the root walk. Today that is a leak-shaped bug, not a use-after-free.

**3. `markObjectContents`'s 41 arms are retention-only.** Its sole caller is the remembered-set walk, so the whole switch exists to *prune* the remembered set rather than to keep objects alive.

## The part that makes this worth tracking rather than just noting

If `minorCollect` is ever made a true generational mark — the obvious optimisation, and the fix for consequence 1 — then consequences 2 and 3 invert. Every `writeBarrier` call site becomes load-bearing, and a missing one becomes a **use-after-free** rather than a retention miss.

Phase 7B pins this with a dedicated test that **flips to failing the moment `minorCollect` stops marking through old objects**, so whoever makes that change is told that every mutation site now needs auditing. That is the intended signal, not a spurious failure.

## Suggested direction

Two separable pieces of work: (a) make the minor mark actually generational, and (b) before doing so, audit every `writeBarrier` call site, because (a) is what makes them matter. `.claude/rules/gc-safety.md` should say which world it is describing — it currently reads as though the barrier is already load-bearing.

**Verified:** `minorCollect`'s body, `markValueInner`'s absence of generation checks, and `markObjectContents`' single caller all read directly at `7083b655`; the marks-through-old property asserted by a committed test.

baijum:
Fixes #1961.

## What changed

A minor collection used to perform a **full transitive mark** — `markValueInner` had no generation check, so minor and full collections cost the same to mark (only the sweep differed). This PR makes the minor mark actually generational, following the design in the issue's audit comment, plus the additional edge-audit the `-Dgc-stress=true` suite demanded.

### Core

- **`GC.minor_marking`**: while set, `markValueInner` treats the old generation as opaque — old objects are never swept by `sweepYoung`, so they need no mark. The minor mark now costs O(live young) instead of O(live heap). `clearOldMarks` is dropped from `minorCollect` (nothing marks an old object during a minor; every sweep clears the marks it observes).
- **Weak predicates** (`keptAlive`/`weakReachable`): an old object answers *alive* during a minor — the old generation survives every minor, so an ephemeron with an old key must not break (destructive!) and a guardian watching an old object must not resurrect early. The decision defers to the next full collection, where old garbage is unmarked and visible.

### Remembered-set completeness — what makes the above safe

The issue's comment identified two gaps (guardian registration barrier, promotion scan) and one coverage question (running-fiber fields). All three are here, plus three more the stress suite surfaced:

1. **Promotion scan** (`sweepYoung`): the write barrier only fires for containers *already* old, so an edge created while the container was young is recorded when the container is promoted, via a `referencesYoung` scan.
2. **Full-collect re-scan** (`sweepOld`): a full collection drains the remembered set up front and never promotes, so an old container whose young referent just survived the full would face the next minor with no entry. Surviving old→young edges are re-recorded while the old heap is walked anyway — found by gc-stress; every container type is affected, not just guardians.
3. **Missing write barriers** found by the audit and fixed:
   - guardian registration in `invokeGuardian` (`registered` grows at every registration — the old comment claiming weak structures never enter the remembered set was wrong for guardians),
   - `Function.constants` appends into a possibly-promoted function (`compiler.addConstant`, the `.sbc` deserializer, gc-stress test helpers),
   - `define`/`set!`/`define-syntax` into a **mutable `SchemeEnvironment`**'s map (opcode handlers + `compileDefineSyntax`; the root-marked globals/lib_env maps need no barrier),
   - the running fiber's `param_overrides` put (`VM.setParameterValue`),
   - `trackOwnedMutex`'s append (belt-and-braces; the explicit fiber marking already covers it).
4. **`FiberScheduler.markRoots`** now calls `markFiberState` for *every* resident fiber, the running one included. A promoted fiber is opaque to the minor mark, and a running fiber's stale saved snapshot must stay *traced* because full collections still walk it through `markValueInner`'s `.fiber` arm — a value freed by an intervening minor would turn that walk into a use-after-free. *(Deviation from the issue comment's design, which proposed marking only the non-swapped fields — that variant crashes under gc-stress; the full explicit pass is both simpler and matches the pre-#1961 retention semantics for the snapshot.)*

### Tests

- The old sentinel (`a minor collection marks through an old object with no remembered-set entry`) flips to pin the opposite — exactly the signal it was written to give. Its companion proves the remembered walk keeps the referent alive.
- New pins: the promotion scan (promote-then-kept-young), the full-collect re-scan, the weak-predicate deferral (old ephemeron key not broken / old watched object not resurrected during a minor, both decided at the next full), the running old fiber, and `stats.minor_old_skips` proving the minor mark stops at the generational boundary while a full mark does not.
- End-to-end smoke test (`tests/scheme/smoke/gc-generational-minor-1961.scm`): an old guardian's registered entry across churn, a `parameterize`d value on the promoted main fiber, and a large old heap under young churn.

### Docs

`.claude/rules/gc-safety.md` and `docs/dev/gc-safety-and-error-handling.md` previously read as though the barrier were already load-bearing; they now describe the mechanisms that actually make it so (barrier + promotion scan + full-collect re-scan, the two wholesale-marked exceptions, and the mutable-environment case).

## Validation

- `zig build test` — green
- `zig build test -Dgc-stress=true` — green (this is the suite that caught gaps 2 and 3)
- R7RS suite 1395/1395 — including under a `-Dgc-stress=true` binary
- `bash tests/scheme/run-all.sh` — 2115 pass, 0 fail
- markdownlint — clean

### DIFF
--- a/docs/dev/gc-safety-and-error-handling.md
+++ b/docs/dev/gc-safety-and-error-handling.md
@@ -99,24 +99,64 @@ keeps it alive.
 
 ### The write barrier
 
-Because the collector is generational, minor collections scan only the
-young generation plus a remembered set of old objects that point into
-it. Mutating a field of a heap object (set-car!, set-cdr!, vector-set!,
-hash-table-set!, record field mutation) can create an old→young
-reference the minor collection would otherwise never see — the young
-object would be freed while still reachable.
+Minor collections are generational (#1961): the minor mark traces roots
+into the young generation and treats old objects as opaque (`GC.minor_marking`
+gates the check in `markValueInner`), so it costs O(live young) rather than
+O(live heap). An old→young reference — a field of an old heap object holding
+a young object — is therefore invisible to that mark unless it is recorded.
+The remembered set is that record, and it is fed two ways:
 
-After storing a Value into a heap object field, call:
+- `gc.writeBarrier(container, new_val)` after storing a Value into a heap
+  object field (set-car!, set-cdr!, vector-set!, hash-table-set!, record
+  field mutation, guardian registration, …). The barrier is load-bearing:
+  a missing one is a **use-after-free** — the young referent is swept while
+  the old container still points at it — not a leak.
+- The **promotion scan** in `sweepYoung`: the barrier only fires for a
+  container that is *already* old, so an edge created while the container
+  was young (a barrier there is a correct no-op) is recorded when the
+  container is promoted, by scanning it with `referencesYoung` at that
+  moment. The dual **full-collect re-scan** in `sweepOld` re-records every
+  surviving old→young edge after a full collection, whose up-front drain
+  would otherwise leave the next minor without the entry (full collections
+  never promote, so the referent is still young).
 
-```zig
-gc.writeBarrier(container, new_val);
-```
+During the minor mark, the remembered-set walk (`markObjectContents`) marks
+the young referents of every remembered container; `pruneRememberedSet`
+drops entries whose referents have all aged into the old generation, so a
+later old→young write re-queues the container (the write barrier does).
 
-where `container` is the heap object being mutated. The barrier records
-the old→young edge in the remembered set. Omitting it does not fail
-immediately — it corrupts the heap only when a minor collection happens
-to run before the next full collection, which is why these bugs surface
-as rare, allocation-pattern-sensitive crashes.
+Two things deliberately do **not** rely on the barrier:
+
+- Bulk state the GC re-traces wholesale: every resident fiber — the running
+  one included — gets an explicit `markFiberState` pass from
+  `FiberScheduler.markRoots` each collection. That covers a suspended
+  fiber's saved registers/frames (written by `saveCurrentFiber`'s memcpy,
+  with no per-slot barriers) and, for a running fiber, both its mutable
+  fields (`waiting_on`, parameter overrides, held mutexes, …) and its
+  *stale* saved snapshot — which must stay traced because full collections
+  still walk it through `markValueInner`'s `.fiber` arm, and a value freed
+  by an intervening minor would turn that walk into a use-after-free.
+- The root-marked maps (`vm.globals`, library `lib_env`s): their values are
+  marked directly as roots every collection, so `set!`/`define` into them
+  needs no barrier. A *mutable* `SchemeEnvironment` (the `eval`/
+  `environment` objects) is a heap object, not a root-marked map — the
+  `define`/`set!`/`define-syntax` stores into its map carry a write barrier
+  on the wrapper object.
+
+Weak structures follow the same rule with one addition: a guardian's
+`registered` list grows at every registration, so `invokeGuardian` barriers
+the append, and `processWeakRefs` re-queues an old guardian that still
+references young objects after a full collection (which drains the set).
+The weak probes (`keptAlive`/`weakReachable`) treat every old object as
+alive during a minor collection — the old generation is never swept by one,
+so an ephemeron with an old key must not break and a guardian watching an
+old object must not resurrect; the decision is deferred to the next full
+collection, where old garbage is unmarked and visible.
+
+Omitting a barrier does not fail immediately — it corrupts the heap only
+when a minor collection happens to run before the next full collection,
+which is why these bugs surface as rare, allocation-pattern-sensitive
+crashes.
 
 ### Stress testing
 
--- a/docs/dev/gc-safety-and-error-handling.md
+++ b/docs/dev/gc-safety-and-error-handling.md
@@ -108,9 +108,15 @@ The remembered set is that record, and it is fed two ways:
 
 - `gc.writeBarrier(container, new_val)` after storing a Value into a heap
   object field (set-car!, set-cdr!, vector-set!, hash-table-set!, record
-  field mutation, guardian registration, …). The barrier is load-bearing:
-  a missing one is a **use-after-free** — the young referent is swept while
-  the old container still points at it — not a leak.
+  field mutation, guardian registration, a function's `global_cache`
+  slots, …). The barrier is load-bearing: a missing one is a
+  **use-after-free** — the young referent is swept while the old container
+  still points at it — not a leak. Two shared funnels keep the discipline
+  in one place: `GC.appendFunctionConstant` (every `Function.constants`
+  append — compiler, `.sbc` deserializer, tests) and `GC.rememberObject`
+  (every wholesale enrollment of an owned container outside the
+  store+barrier pattern — `writeBarrier`'s owned branch, both sweep-phase
+  scans, `saveCurrentFiber`'s bulk snapshot).
 - The **promotion scan** in `sweepYoung`: the barrier only fires for a
   container that is *already* old, so an edge created while the container
   was young (a barrier there is a correct no-op) is recorded when the
@@ -129,29 +135,38 @@ Two things deliberately do **not** rely on the barrier:
 
 - Bulk state the GC re-traces wholesale: every resident fiber — the running
   one included — gets an explicit `markFiberState` pass from
-  `FiberScheduler.markRoots` each collection. That covers a suspended
-  fiber's saved registers/frames (written by `saveCurrentFiber`'s memcpy,
-  with no per-slot barriers) and, for a running fiber, both its mutable
-  fields (`waiting_on`, parameter overrides, held mutexes, …) and its
-  *stale* saved snapshot — which must stay traced because full collections
-  still walk it through `markValueInner`'s `.fiber` arm, and a value freed
-  by an intervening minor would turn that walk into a use-after-free.
+  `FiberScheduler.markRoots` each collection (authoritative; the fiber-field
+  barriers are belt-and-braces). That covers a suspended fiber's saved
+  registers/frames and, for a running fiber, both its mutable fields
+  (`waiting_on`, parameter overrides, held mutexes, …) and its *stale* saved
+  snapshot — which must stay traced because full collections still walk it
+  through `markValueInner`'s `.fiber` arm, and a value freed by an
+  intervening minor would turn that walk into a use-after-free. A fiber
+  retired from its scheduler slot loses that pass while staying
+  heap-reachable, so `saveCurrentFiber` also enrolls an old fiber in the
+  remembered set wholesale (`GC.rememberObject`): its bulk snapshot has no
+  per-field barriers, and `pruneRememberedSet` drops the entry once the
+  snapshot has aged.
 - The root-marked maps (`vm.globals`, library `lib_env`s): their values are
   marked directly as roots every collection, so `set!`/`define` into them
-  needs no barrier. A *mutable* `SchemeEnvironment` (the `eval`/
-  `environment` objects) is a heap object, not a root-marked map — the
-  `define`/`set!`/`define-syntax` stores into its map carry a write barrier
-  on the wrapper object.
+  needs no barrier — nor into the interaction-environment wrapper, whose
+  map *is* `vm.globals` (the `.owned == false` case the handlers skip). A
+  *mutable* `SchemeEnvironment` (the `eval`/`environment` objects) is a
+  heap object, not a root-marked map — the `define`/`set!`/`define-syntax`
+  stores into its map carry a write barrier on the wrapper object, and the
+  compiler's `def_env_val`/`let_syntax_peer_vals` stores into a
+  possibly-promoted aliased transformer (SRFI 147/211) carry one on the
+  transformer.
 
 Weak structures follow the same rule with one addition: a guardian's
 `registered` list grows at every registration, so `invokeGuardian` barriers
-the append, and `processWeakRefs` re-queues an old guardian that still
-references young objects after a full collection (which drains the set).
-The weak probes (`keptAlive`/`weakReachable`) treat every old object as
-alive during a minor collection — the old generation is never swept by one,
-so an ephemeron with an old key must not break and a guardian watching an
-old object must not resurrect; the decision is deferred to the next full
-collection, where old garbage is unmarked and visible.
+the append (the old "weak structures never enter the remembered set" claim
+was wrong for guardians). The weak probes (`keptAlive`/`weakReachable`)
+treat every old object as alive during a minor collection — the old
+generation is never swept by one, so an ephemeron with an old key must not
+break and a guardian watching an old object must not resurrect; the
+decision is deferred to the next full collection, where old garbage is
+unmarked and visible.
 
 Omitting a barrier does not fail immediately — it corrupts the heap only
 when a minor collection happens to run before the next full collection,
--- a/docs/dev/gc-safety-and-error-handling.md
+++ b/docs/dev/gc-safety-and-error-handling.md
@@ -150,10 +150,11 @@ Two things deliberately do **not** rely on the barrier:
 - The root-marked maps (`vm.globals`, library `lib_env`s): their values are
   marked directly as roots every collection, so `set!`/`define` into them
   needs no barrier — nor into the interaction-environment wrapper, whose
-  map *is* `vm.globals` (the `.owned == false` case the handlers skip). A
-  *mutable* `SchemeEnvironment` (the `eval`/`environment` objects) is a
-  heap object, not a root-marked map — the `define`/`set!`/`define-syntax`
-  stores into its map carry a write barrier on the wrapper object, and the
+  map *is* `vm.globals`. That exclusion and the barrier it gates live once,
+  in `GC.envStoreBarrier`: a *mutable* `SchemeEnvironment` (the `eval`/
+  `environment` objects) is a heap object, not a root-marked map, so the
+  `define`/`set!`/`define-syntax` stores into its map carry a write barrier
+  on the wrapper object — the `.owned == false` wrapper does not. The
   compiler's `def_env_val`/`let_syntax_peer_vals` stores into a
   possibly-promoted aliased transformer (SRFI 147/211) carry one on the
   transformer.
--- a/docs/dev/gc-safety-and-error-handling.md
+++ b/docs/dev/gc-safety-and-error-handling.md
@@ -101,8 +101,10 @@ keeps it alive.
 
 Minor collections are generational (#1961): the minor mark traces roots
 into the young generation and treats old objects as opaque (`GC.minor_marking`
-gates the check in `markValueInner`), so it costs O(live young) rather than
-O(live heap). An old→young reference — a field of an old heap object holding
+gates the check in `markValueInner`). Its cost is O(live young) plus the
+fields scanned in remembered containers — a large old container with one
+young referent still costs its full capacity to re-scan each minor, which
+is why the dedup and pruning below matter — not O(live heap). An old→young reference — a field of an old heap object holding
 a young object — is therefore invisible to that mark unless it is recorded.
 The remembered set is that record, and it is fed two ways:
 
--- a/docs/dev/gc-safety-and-error-handling.md
+++ b/docs/dev/gc-safety-and-error-handling.md
@@ -154,8 +154,9 @@ Two things deliberately do **not** rely on the barrier:
   in `GC.envStoreBarrier`: a *mutable* `SchemeEnvironment` (the `eval`/
   `environment` objects) is a heap object, not a root-marked map, so the
   `define`/`set!`/`define-syntax` stores into its map carry a write barrier
-  on the wrapper object — the `.owned == false` wrapper does not. The
-  compiler's `def_env_val`/`let_syntax_peer_vals` stores into a
+  on the wrapper object — the `.owned == false` wrapper does not, and
+  `referencesYoung` answers false for it too, so neither scan enrolls it.
+  The compiler's `def_env_val`/`let_syntax_peer_vals` stores into a
   possibly-promoted aliased transformer (SRFI 147/211) carry one on the
   transformer.
 
--- a/src/bytecode_file.zig
+++ b/src/bytecode_file.zig
@@ -381,23 +381,31 @@ test "bytecode round-trip: various constant types" {
     // Add various constant types
     func.constants.append(allocator, types.makeFixnum(-100)) catch unreachable;
     func.constants.append(allocator, types.TRUE) catch unreachable;
+    gc.writeBarrier(&func.header, types.TRUE); // #1961: func may be promoted already
     func.constants.append(allocator, types.FALSE) catch unreachable;
+    gc.writeBarrier(&func.header, types.FALSE); // #1961: func may be promoted already
     func.constants.append(allocator, types.NIL) catch unreachable;
+    gc.writeBarrier(&func.header, types.NIL); // #1961: func may be promoted already
     func.constants.append(allocator, types.VOID) catch unreachable;
+    gc.writeBarrier(&func.header, types.VOID); // #1961: func may be promoted already
     func.constants.append(allocator, types.makeChar('Z')) catch unreachable;
 
     const sym = try gc.allocSymbol("hello");
     func.constants.append(allocator, sym) catch unreachable;
+    gc.writeBarrier(&func.header, sym); // #1961: func may be promoted already
 
     const str = try gc.allocString("world");
     func.constants.append(allocator, str) catch unreachable;
+    gc.writeBarrier(&func.header, str); // #1961: func may be promoted already
 
     const flo = types.makeFlonum(3.14);
     func.constants.append(allocator, flo) catch unreachable;
+    gc.writeBarrier(&func.header, flo); // #1961: func may be promoted already
 
     const bv_data = [_]u8{ 1, 2, 3 };
     const bv = try gc.allocBytevector(&bv_data);
     func.constants.append(allocator, bv) catch unreachable;
+    gc.writeBarrier(&func.header, bv); // #1961: func may be promoted already
 
     func.arity = 0;
     func.locals_count = 1;
@@ -467,6 +475,7 @@ test "bytecode round-trip: nested functions" {
     parent_func.code.append(allocator, 0) catch unreachable; // src high
     parent_func.code.append(allocator, 0) catch unreachable; // src low
     parent_func.constants.append(allocator, types.makePointer(&child_func.header)) catch unreachable;
+    gc.writeBarrier(&parent_func.header, types.makePointer(&child_func.header)); // #1961: func may be promoted already
     parent_func.arity = 0;
     parent_func.locals_count = 1;
 
@@ -592,6 +601,7 @@ test "bytecode round-trip: datum-label sharing preserved via backrefs (kaappi#21
     gc.pushRoot(&tail);
     const outer = try gc.allocPair(shared_pair, tail);
     func.constants.append(allocator, outer) catch unreachable;
+    gc.writeBarrier(&func.header, outer); // #1961: func may be promoted already
 
     var shared_str = try gc.allocString("s");
     gc.pushRoot(&shared_str);
@@ -603,6 +613,7 @@ test "bytecode round-trip: datum-label sharing preserved via backrefs (kaappi#21
     gc.writeBarrier(types.toObject(vec), shared_pair);
     gc.writeBarrier(types.toObject(vec), shared_str);
     func.constants.append(allocator, vec) catch unreachable;
+    gc.writeBarrier(&func.header, vec); // #1961: func may be promoted already
     gc.popRoot(); // shared_str
     gc.popRoot(); // tail
     gc.popRoot(); // shared_pair
@@ -642,6 +653,7 @@ test "bytecode round-trip: cyclic literal terminates and re-ties the knot (kaapp
     types.toObject(head).as(types.Pair).cdr = second;
     gc.writeBarrier(types.toObject(head), second);
     func.constants.append(allocator, head) catch unreachable;
+    gc.writeBarrier(&func.header, head); // #1961: func may be promoted already
     gc.popRoot();
 
     // A self-referential vector too: #0=#(1 #0#).
@@ -650,6 +662,7 @@ test "bytecode round-trip: cyclic literal terminates and re-ties the knot (kaapp
     types.toVector(cyc_vec).data[1] = cyc_vec;
     gc.writeBarrier(types.toObject(cyc_vec), cyc_vec);
     func.constants.append(allocator, cyc_vec) catch unreachable;
+    gc.writeBarrier(&func.header, cyc_vec); // #1961: func may be promoted already
     gc.popRoot();
 
     var loaded = try roundTrip(&gc, func, "/tmp/kaappi_test_cycle.sbc");
@@ -684,6 +697,7 @@ test "bytecode round-trip: a long quoted list costs no nesting depth (kaappi#211
         list = try gc.allocPair(types.makeFixnum(i), list);
     }
     func.constants.append(allocator, list) catch unreachable;
+    gc.writeBarrier(&func.header, list); // #1961: func may be promoted already
     gc.popRoot();
 
     var loaded = try roundTrip(&gc, func, "/tmp/kaappi_test_longlist.sbc");
@@ -778,6 +792,7 @@ test "bytecode write: refuses what the reader would reject (kaappi#2113)" {
         nested = try gc.allocPair(nested, types.NIL);
     }
     func.constants.append(allocator, nested) catch unreachable;
+    gc.writeBarrier(&func.header, nested); // #1961: func may be promoted already
     gc.popRoot();
 
     var funcs_arr = [_]*Function{func};
@@ -792,7 +807,9 @@ test "bytecode write: refuses what the reader would reject (kaappi#2113)" {
     const big = allocator.alloc(u8, MAX_STRING_BYTES + 1) catch unreachable;
     defer allocator.free(big);
     @memset(big, 'x');
-    func.constants.append(allocator, try gc.allocString(big)) catch unreachable;
+    const big_str = try gc.allocString(big);
+    func.constants.append(allocator, big_str) catch unreachable;
+    gc.writeBarrier(&func.header, big_str); // #1961: func may be promoted already
     try std.testing.expectError(BytecodeError.LimitExceeded, writeFileWithTopLevel(allocator, &funcs_arr, 0x1234, "test.scm", path));
 
     // And exactly AT the cap still round-trips — the two halves agree on the
@@ -805,6 +822,7 @@ test "bytecode write: refuses what the reader would reject (kaappi#2113)" {
         at_cap = try gc.allocPair(at_cap, types.NIL);
     }
     func.constants.append(allocator, at_cap) catch unreachable;
+    gc.writeBarrier(&func.header, at_cap); // #1961: func may be promoted already
     gc.popRoot();
     var loaded = try roundTrip(&gc, func, path);
     read.freeDeserializeResult(allocator, &loaded);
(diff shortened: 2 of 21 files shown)


baijum:
Thanks @coderabbitai — disposition of the four findings:

1. **Pre-store barrier order (gc-safety.md)** — skipped. The repo's established convention is store-*then*-barrier (`types.setCdr(tail, rest); gc.writeBarrier(...)` in reader_datum.zig, primitives.zig, primitives_vector.zig, …) and `.claude/rules/gc-safety.md` documents exactly that order. Both orders are safe here because no allocation — hence no collection — can occur between the store and the barrier; I kept the code consistent with the surrounding convention rather than introduce a second style.

2. **Begin-wrapped `define-syntax` env store (compiler_define_syntax.zig:702)** — fixed, good catch. The SRFI-148 helper-definition path had the same `env.put` I had only fixed in `compileDefineSyntax`; it now carries the same conditional barrier (b5735538).

3. **Scope `set_global`'s barrier to the successful store (vm_dispatch.zig)** — taken in part. The barrier now sits inside the non-null `ptr` branch, so a failed lookup records nothing. I kept it *after* the store per the convention in (1); the comment there spells out why that order is equivalent when nothing alloculates in between.

4. **SRFI-64 + `(exit 1)` for the smoke test** — skipped. The dominant convention in `tests/scheme/smoke/` is the hand-rolled `check` harness ending in `(if (> fail 0) (error ...))` — see `gc-rooting-stress.scm`, which is the closest neighbor (same GC-stress category). `(error ...)` exits non-zero, which is what `run-all.sh`'s exit-code contract needs, and the suite's output scraping expects the `N passed, M failed` line this shape prints.

src/vm_dispatch.zig
```diff
@@ -336,7 +336,19 @@ pub fn runUntil(self: *VM, target_frame_count: usize, target_wind_count: usize)
...
+                    p.* = val;
+                    // #1961: when the target map belongs to a mutable
+                    // SchemeEnvironment (func.env_val — the eval/environment
+                    // objects), this store is an old→young edge on that heap
+                    // object once it promotes. The VM-rooted maps (globals,
+                    // library lib_envs — env_val NIL there) are marked as roots
+                    // every collection and need no barrier. Nothing allocates
+                    // between the store and the barrier, so this order is as
+                    // safe as barrier-first.
+                    if (types.isEnvironment(func.env_val))
```
baijum: **[correctness — confirmed] `Function.global_cache` stores are missing write barriers (use-after-free under the new generational mark)**

The barrier audit covered `constants.append` but missed the other Value store into a `Function`: the global-cache fills at lines 266/270 (`get_global`), 367 (`set_global`), 417 (`define_global`), 997/1021 (`call_global`), and 1113/1119 (`tail_call_global`). All of them store a Value into a possibly-old `Function` with no `writeBarrier(&func.header, val)`, while `markObjectContents`, `markValueInner`'s `.function` arm, and `referencesYoung` all walk `global_cache` unconditionally — there is no `cache_version` guard on the trace side, and a stale cache is only memset when *that* function next executes a global op, so a function that never runs again keeps the stale slot forever.

Failure sequence: old promoted `F` executes `get_global g`, caching young closure `C` (alive via the root-marked globals map). `g` is rebound — `global_version` bumps but `F`'s cache is physically untouched. `C`'s only remaining reference is `F`'s untraced cache slot; the next minor sweeps `C` (old `F` is opaque and not remembered); the next full collection walks `F.global_cache` into freed memory — `GC: marking freed object` panic under gc-stress/Debug, silent heap corruption in release. Pre-#1961 the full transitive minor mark absorbed exactly this.

Fix: `self.gc.writeBarrier(&func.header, val)` at each cache store (the fresh-cache-allocation stores too, since a function can promote before its first global op). Cheap and deduplicated via `in_remembered_set`.

src/fiber.zig
```diff
@@ -1032,7 +1032,17 @@ pub const FiberScheduler = struct {
...
+                // resident fiber gets this explicit pass, the running one
+                // included. For a running fiber the saved registers/frames are
+                // stale (the VM holds the live copies, marked by markVmRoots),
+                // but they must stay *traced* anyway: markValueInner's .fiber
+                // arm still walks them during full collections, and a value
+                // freed by a minor in between would turn that walk into a
+                // use-after-free. Retaining the stale snapshot until the next
+                // save overwrites it is the same conservatism the pre-#1961
+                // full transitive minor mark provided.
                 markFiberState(gc, fiber);
```
baijum, line 1062: **[correctness — confirmed] The resident-fiber exemption ends at slot reuse, but the retire paths store into the fiber *after* `retireSlot` with no barriers**

Both retire paths run `saveCurrentFiber` **after** `retireSlot` has queued the slot for reuse (`vm_calls.zig:386-389`, `fiber_wait.zig:495-505`), and `saveCurrentFiber` (fiber.zig:584-609) memcpys registers/frames/handlers/winds and assigns `continuation_value`/`current_exception` with zero barriers — only `fiber.result` is barriered. On the errored path a freshly allocated (young) ErrorObject is staged into `vm.current_exception` precisely so the save copies it into `fiber.current_exception`.

While the terminal fiber still occupies its slot this pass covers it — but `addFiber`/`popFreeSlot` overwrite the slot on the next `spawn`, ending residency while the fiber stays heap-reachable (a joiner's `waiting_on`, or any variable holding the handle). The promotion scan can't help (it ran when the fiber promoted, before these final stores), and `pruneRememberedSet` has already dropped a long-lived fiber whose old snapshot went all-old. Result: the next minor sweeps the young exception/snapshot values; `reraiseFiberError` re-raises a freed object, and the next full collection's `.fiber` arm walks freed values — the exact hazard this comment block names for the resident case, now on a non-resident fiber.

Fix: at the end of `saveCurrentFiber`, if `fiber.header.flags.generation == 1`, enter the fiber into the remembered set (a shared `rememberObject` helper). That covers every save site uniformly and `pruneRememberedSet` drops the entry once the snapshot ages. Barriering only `current_exception`/`continuation_value` at the retire sites would leave the saved registers/frames unprotected.

src/compiler_define_syntax.zig
```diff
@@ -84,6 +84,13 @@ pub fn compileDefineSyntax(self: *Compiler, args: Value, dst: u16) CompileError!
     if (self.body_macro_depth == 0) {
         if (self.lib_env) |env| {
             env.put(name, transformer) catch return CompileError.OutOfMemory;
+            // #1961: for a mutable SchemeEnvironment (eval/environment
+            // objects — assertEnvMapInvariant's env_val case) this store is
+            // an old→young edge on that heap object once it promotes; the
+            // VM-rooted library lib_envs (env_val NIL) are marked as roots
+            // every collection, where the barrier is simply a no-op.
+            if (types.isEnvironment(self.lib_env_val))
```
baijum: **[correctness — confirmed] `tx.let_syntax_peer_vals` snapshot store (lines 270-271) is unbarriered, and the SRFI 147 alias path hands it an arbitrarily old transformer**

`resolveTransformerSpecRec` can return a *pre-existing* transformer for a bare-symbol spec (merged_macros hit at :548, SRFI 211 globals-held transformer at :561). A transformer defined by plain `define-syntax` never sets `peers_computed`, so an old aliased tx passes the :235 guard and takes the snapshot store at :270-271 — an old→young edge with no barrier. `let_syntax_peer_vals` is traced by all three collector arms and dereferenced at expansion time (`compiler_macro.zig:439-440`).

Repro shape, in a lambda body: `(define-syntax helper …)` (young, held only by `self.macros`/extra_roots, both unwound at the compile boundary), then `(let-syntax ((k old-m)) …)` where `old-m` is an old transformer whose template references `helper`. After the compile, `old-m.let_syntax_peer_vals` is helper's only reference; the next minor sweeps it; a later expansion of `k` or the next full-collect trace of `old-m` hits freed memory.

Fix — after line 271:
```zig
for (new_peer_vals) |pv|
    self.gc.writeBarrier(types.toObject(transformer), pv);
```

Root cause worth a note in the file: the alias alternatives make `resolveTransformerSpec` return possibly-old transformers, and every post-resolution field store here was written assuming a fresh (young) tx the promotion scan would cover.

src/compiler_define_syntax.zig
```diff
@@ -693,6 +700,11 @@ fn resolveTransformerSpecRec(self: *Compiler, spec_in: Value, merged_macros: *st
                 if (self.body_macro_depth == 0) {
                     if (self.lib_env) |env| {
                         env.put(def_name, def_transformer) catch return CompileError.OutOfMemory;
+                        // #1961: same old→young edge on a mutable
+                        // SchemeEnvironment as compileDefineSyntax's own
+                        // lib_env store above.
+                        if (types.isEnvironment(self.lib_env_val))
```
baijum: **[correctness — confirmed] `tx.def_env_val = self.lib_env_val` (line 67, and the begin-wrapped path at ~:694) mutates a possibly-old transformer with no barrier on the *transformer* as container**

b5735538 barriered the opposite direction only (env-object → transformer, after `env.put`); the transformer-side edge is still unrecorded. `def_env_val` is GC-traced in `markObjectContents`, `markValueInner`, and `referencesYoung`, and the transformer can be pre-existing and old via the same SRFI 147/211 alias paths (`resolveTransformerSpecRec` :548/:561).

Failure: `(eval '(define-syntax k m) my-env)` with old transformer `m` and a fresh mutable environment — old `m` receives `def_env_val` = young env, unbarriered. Per the comment at :88-91, mutable eval/environment objects are *not* VM-rooted, so once the caller drops the env, `m.def_env_val` is its only reference; the next minor sweeps it, leaving hygiene resolution and the next full-collect trace pointing at a freed environment. (The fresh-transformer common case is safe — tx is young at the store and the promotion scan records the edge later — which is why tests don't catch this.)

Fix: `self.gc.writeBarrier(types.toObject(transformer), self.lib_env_val);` after line 67, and the same for `def_transformer` after ~:694. No `isEnvironment` guard needed — `writeBarrier` no-ops on non-pointer values, and a NIL `lib_env_val` is an immediate.

src/vm_dispatch.zig
```diff
@@ -336,7 +336,19 @@ pub fn runUntil(self: *VM, target_frame_count: usize, target_wind_count: usize)
...
+                        self.gc.writeBarrier(types.toObject(func.env_val), val);
```
baijum: **[efficiency] This guard also fires for the `interaction-environment` wrapper, contradicting the comment's own exemption**

`interaction-environment` is `allocEnvironment(vm.globals, false, false)` (primitives_r7rs.zig:458) — a mutable heap `SchemeEnvironment` *borrowing the root-marked globals map*. Eval'ing through it sets `func.env == vm.globals` and `func.env_val` = the wrapper, so `isEnvironment(func.env_val)` is true and this barrier enrolls the wrapper in the remembered set — even though the stored-to map is exactly the "VM-rooted maps … marked as roots every collection" case the adjacent comment exempts. Once the wrapper is old and enrolled, `markObjectContents` + `referencesYoung` re-walk the entire globals map every minor (mark + prune), and the entry never prunes while any global value is young — a permanent redundant O(#globals) walk per minor on top of `markVmRoots`' own walk. Notably the `define` path below takes the `env == self.globals` branch and skips the barrier for the same wrapper — the two opcodes disagree.

Since `environment`/`null-environment`/`scheme-report-environment` are all immutable (rejected before the store), the interaction-environment wrapper may be the only mutable `SchemeEnvironment` reachable from Scheme today — i.e. these barriers may only ever fire in the redundant case.

Suggested guard at both sites: `func.env != self.globals` (or `!types.toEnvironment(func.env_val).owned`, the flag that uniquely marks the borrowed-globals wrapper).

src/vm.zig
```diff
@@ -946,6 +946,11 @@ pub const VM = struct {
         const key = @intFromPtr(param);
         if (self.current_fiber) |fiber| {
             fiber.param_overrides.put(key, val) catch return VMError.OutOfMemory;
+            // #1961: an old running fiber's override map holding a young
+            // value is an old→young edge the generational minor mark reaches
+            // only through the remembered set (FiberScheduler.markRoots'
+            // explicit markFiberState pass is the belt-and-braces backstop).
```
baijum: **[comment accuracy] This comment inverts which mechanism is load-bearing — the other three sites (and this PR's own test) say the opposite**

Here: the edge "is reached **only through the remembered set** (FiberScheduler.markRoots' explicit markFiberState pass is the belt-and-braces backstop)". But `gc_collect.zig:257-266` states flatly "For fibers the whole remembered-set path is belt-and-braces, not load-bearing: every scheduler-resident fiber is marked as an unconditional root each collection"; `trackOwnedMutex`'s comment agrees (markFiberState "covers it as a root"); and the new `tests_scheduler.zig` test deliberately skips the barrier on `waiting_on` and proves markFiberState alone keeps the young value alive. `markFiberState` does mark `param_overrides` — so it is the authoritative mechanism and this comment has primary/backstop reversed.

Why it matters: a future refactorer trusting this wording could remove the running-fiber markFiberState pass ("the barrier covers it") and reintroduce a UAF for the many barrier-less fiber mutation sites (saveCurrentFiber's memcpy, `waiting_on` stores) that only markFiberState protects. Suggest rewording to match gc_collect.zig, and stating the invariant once: markFiberState is authoritative for resident fibers; fiber-field barriers are belt-and-braces.

src/compiler.zig
```diff
@@ -288,6 +288,11 @@ pub const Compiler = struct {
         }
         if (self.func.constants.items.len >= 65536) return CompileError.TooManyConstants;
         self.func.constants.append(self.gc.allocator, value) catch return CompileError.OutOfMemory;
+        // #1961: compiling a large body spans collections, so `func` may
```
baijum: **[test-coverage] Several of the PR's load-bearing barrier fixes have no regression test (kaappi/CLAUDE.md: "Every bug fix MUST include a regression test that fails without the fix")**

Removing this `addConstant` barrier fails nothing in the suite — the barrier edits in `bytecode_file.zig`/`disassembler.zig` are test *setup* code applying their own barriers, not tests of this path. Same for the deserializer barrier (`bytecode_file_read.zig:636`) and the mutable-`SchemeEnvironment` `set!`/`define`/`define-syntax` barriers (no test creates a mutable environment, promotes it, stores through the real opcode path, and survives a minor). The `invokeGuardian` barrier is pinned only by a test that *mirrors* the store+barrier pair rather than calling `invokeGuardian`, so deleting the primitive's barriers leaves the deterministic test green (only the churn-based smoke test might catch it).

Deterministic pins are feasible with the pattern the new `tests_gc_tracing.zig` tests already use: promote the container, drive a young value through the real code path, `forceMinor`, assert liveness by list membership.

src/bytecode_file.zig
```diff
@@ -381,23 +381,31 @@ test "bytecode round-trip: various constant types" {
     // Add various constant types
     func.constants.append(allocator, types.makeFixnum(-100)) catch unreachable;
     func.constants.append(allocator, types.TRUE) catch unreachable;
+    gc.writeBarrier(&func.header, types.TRUE); // #1961: func may be promoted already
```
baijum: **[altitude] ~20 hand-copied append+barrier pairs in tests — five of them provable no-ops — instead of one `appendConstant` helper**

Production got the barrier at the right altitude (inside `Compiler.addConstant` and the deserializer loop), but every test must remember the two-line incantation manually. Within this very block the pattern is applied inconsistently: `writeBarrier` begins `if (container.flags.generation == 1 and types.isPointer(new_val))`, so the barriers on `types.TRUE`, `types.FALSE`, `types.NIL`, `types.VOID`, and the flonum are unconditional no-ops, while the fixnum `-100` and char `'Z'` appends beside them get no barrier — correct either way, but it teaches the next reader that immediates need barriers (they never do, anywhere).

Suggest a `Function.appendConstant(gc, val)` (or a shared free function) used by `compiler.addConstant`, `bytecode_file_read.zig`'s loop, and these tests: the barrier becomes a property of "adding a constant" rather than of each call site, and the five no-op calls disappear. Otherwise the next test appending a heap constant without the barrier is a gc-stress-only UAF flake.

src/gc_sweep.zig
```diff
@@ -68,6 +69,20 @@ pub fn sweepYoung(gc: *GC) void {
...
+                // container that is *already* old, so an old→young edge
+                // created while the container was young (a barrier there is
+                // a correct no-op) would go unrecorded at promotion — and
+                // since the minor mark stopped tracing through old objects,
+                // that edge is the referent's only route to survival. Scan
+                // at promotion and remember the container. The flag is
+                // provably clear here: this object was young a moment ago,
+                // and writeBarrier only ever enqueues old containers.
+                if (gc_collect.referencesYoung(gc, o)) {
+                    o.flags.in_remembered_set = true;
```
baijum: **[reuse] Remembered-set enrollment is now hand-rolled in three places — worth one `gc.rememberObject(obj)` helper**

This promotion-scan block, the `sweepOld` re-scan below (:126-129), and `writeBarrier`'s owned branch (memory.zig:405-410) each implement check/set `flags.in_remembered_set` + `remembered_set.append` + `catch @panic`, with three different panic strings and deliberately different flag handling (this site skips the flag check on the "provably clear" argument; the re-scan checks it). The #2196 dedup contract (in-set ⟺ flag set, for owned objects) is now enforced by three independently edited blocks: a future change to dedup semantics, the flag's clearing sites, or the OOM strategy that touches `writeBarrier` but not `gc_sweep.zig` silently splits the invariant — and missing the sweepOld copy drops re-scan entries, which the next minor turns into a swept live referent.

A flag-checked `pub fn rememberObject(self: *GC, obj: *Object) void` is a safe superset of this site's invariant (the check is a no-op when provably clear) and serves all three owned-path call sites; `writeBarrier` keeps its foreign-container special case.

src/gc_collect.zig
```diff
@@ -54,7 +54,14 @@ pub fn collect(gc: *GC) void {
...
 fn minorCollect(gc: *GC) void {
-    clearOldMarks(gc);
+    // #1961: the minor mark is generational. `minor_marking` makes
+    // markValueInner treat the old generation as opaque, so this mark costs
+    // O(live young) — roots plus remembered-set entries — instead of
+    // O(live heap). No clearOldMarks here: nothing marks an old object while
+    // minor_marking is set, and every sweep (sweepYoung/sweep/sweepOld)
+    // clears the marks it observes, so old marks are always false between
+    // collections.
+    gc.minor_marking = true;
```
baijum, line 66: **[consistency] This comment's invariant makes `fullCollect`'s remaining `clearOldMarks` call redundant — one of the two should go**

The claim here — "every sweep (sweepYoung/sweep/sweepOld) clears the marks it observes, so old marks are always false between collections" — checks out (the only `flags.marked = true` site is behind the `minor_marking` gate, and sweepYoung clears the mark before the promotion branch). But under that invariant the `clearOldMarks(gc)` call still made in `fullCollect` (~:359) is equally dead, and keeping it while this comment asserts the invariant is internally inconsistent: the next reader can't tell whether old marks can be dirty at fullCollect entry. Relatedly, fullCollect's "defensive reset if a panic path ever left the flag set" describes an impossible path — `minorCollect` is straight-line and every early exit is an aborting `@panic`.

Suggest either deleting the fullCollect `clearOldMarks` call (making the function dead → delete it) or scoping this comment's claim; and trimming the "defensive" justification (or using `defer gc.minor_marking = false;` here so the invariant is structural rather than by convention).

src/vm_dispatch.zig
baijum: Fixed in f78f3ee4 — all six cache-store sites barriered (get/set/define/call ×2/tail_call), including the fresh-cache-allocation paths, with a comment at each spelling out the orphaned-slot window. Mechanism pin added in `tests_gc_tracing.zig` (`global_cache is a barriered container`, with a no-barrier control); the dispatch paths themselves are exercised end-to-end by the gc-stress suite.

baijum: Fixed in f78f3ee4 — all six cache-store sites barriered (get/set/define/call ×2/tail_call), including the fresh-cache-allocation paths, with a comment at each spelling out the orphaned-slot window. Mechanism pin added in `tests_gc_tracing.zig` (`global_cache is a barriered container`, with a no-barrier control); the dispatch paths themselves are exercised end-to-end by the gc-stress suite.

src/fiber.zig
baijum: Fixed exactly as suggested, in f78f3ee4 — `saveCurrentFiber` ends with `if (fiber.header.flags.generation == 1) vm.gc.rememberObject(&fiber.header);`, covering every save site uniformly (resident fibers keep their authoritative markFiberState pass; this is additive for them). Mechanism pin added in `tests_gc_tracing.zig`: an unbarriered snapshot store into a non-resident old fiber survives via `rememberObject`, and the same store without it is swept — the discriminating control.

src/compiler_define_syntax.zig
baijum: Fixed in f78f3ee4 — `for (new_peer_vals) |pv| self.gc.writeBarrier(types.toObject(transformer), pv);` after the snapshot stores, with a comment noting the root cause you called out: the alias alternatives make `resolveTransformerSpec` return possibly-old transformers, and the post-resolution stores were written assuming a fresh one. The same note now sits on the def_env_val fix.

src/compiler_define_syntax.zig
baijum: Fixed in f78f3ee4 — `self.gc.writeBarrier(types.toObject(transformer), self.lib_env_val);` after line 67 and the same for `def_transformer` on the begin-wrapped path, unguarded as you suggested (writeBarrier no-ops the immediate NIL). The comment records why: the fresh-transformer common case is promotion-scan-covered, the aliased old-transformer case is not.

src/vm_dispatch.zig
baijum: Fixed in f78f3ee4 with the `.owned` guard you suggested — `types.isEnvironment(func.env_val) and types.toEnvironment(func.env_val).owned` at both opcodes, so the interaction-environment wrapper (whose map IS the root-marked globals map) is never enrolled, and define/set! now apply the identical rule at both sites. The comments spell out the borrowed-globals exemption.

src/vm.zig
baijum: Agreed — fixed in f78f3ee4. The vm.zig comment now states markFiberState as the authoritative mechanism for resident fibers, with this barrier (like the other fiber-field barriers) as belt-and-braces, and points at gc_collect.zig's own statement of the invariant plus saveCurrentFiber's enrollment for the non-resident window.

src/compiler.zig
baijum: Addressed in f78f3ee4. Real-path pins added: (1) invokeGuardian — eval `(g (vector 'a 'b 'c))` against a manually promoted guardian, force a minor, and the entry must be probed and resurrected (without the primitive's barrier the old guardian is never enrolled, never pending, and the guardian yields nothing); (2) define/set! into a promoted mutable SchemeEnvironment through the real opcode path, minor, read-back intact (`tests_core_eval.zig`, reusing the #1269 test's env construction). For the constants funnel, production now goes through `GC.appendFunctionConstant`, and the new pin drives that helper against a promoted function with a raw-append control — removing the helper's barrier flips it. The deserializer/compiler dispatch paths remain covered end-to-end by gc-stress (that is what caught them), which the test comments note explicitly.

src/bytecode_file.zig
baijum: Done in f78f3ee4 — `GC.appendFunctionConstant(gc, func, v)` in gc_alloc.zig (append + barrier, no-op on immediates), used by compiler.addConstant, the deserializer loop, and all 30 test sites; the five no-op barriers on immediates are gone along with every hand-copied pair. Pinned by the new constants test (barriered vs raw append against a promoted function).

src/gc_sweep.zig
baijum: Done in f78f3ee4 — `pub fn rememberObject` on GC (flag-checked, one panic string), now used by writeBarrier's owned branch, the promotion scan, the full-collect re-scan, and saveCurrentFiber's new enrollment. The foreign-container special case stays in writeBarrier as you noted. The #2196 contract (in-set ⟺ flag, owned) now has exactly one enforcement point.

src/gc_collect.zig
baijum: Both taken in f78f3ee4: `clearOldMarks` is deleted (call and function — under the invariant it was dead in fullCollect too, and minorCollect's comment now states the full mark-bit invariant so the deletion is self-explaining), `minor_marking` resets via `defer` so the structure is local, and fullCollect's comment now says 'makes the requirement local instead of transitive' rather than claiming a defensive panic path.

src/compiler_define_syntax.zig
```diff
@@ -84,6 +92,13 @@ pub fn compileDefineSyntax(self: *Compiler, args: Value, dst: u16) CompileError!
     if (self.body_macro_depth == 0) {
         if (self.lib_env) |env| {
             env.put(name, transformer) catch return CompileError.OutOfMemory;
+            // #1961: for a mutable SchemeEnvironment (eval/environment
+            // objects — assertEnvMapInvariant's env_val case) this store is
+            // an old→young edge on that heap object once it promotes; the
+            // VM-rooted library lib_envs (env_val NIL) are marked as roots
+            // every collection, where the barrier is simply a no-op.
+            if (types.isEnvironment(self.lib_env_val))
```
baijum: **[nit — consistency] These two `lib_env.put` barriers (here and :726) still lack the `.owned` guard the opcodes now apply**

f78f3ee4 guards the `set_global`/`define_global` env barriers with `types.toEnvironment(func.env_val).owned` so the interaction-environment wrapper (the only `.owned == false` environment — its map *is* the root-marked `vm.globals`) is never enrolled in the remembered set. But this compile-time store and its begin-wrapped sibling at :726 kept the bare `isEnvironment(self.lib_env_val)` test, and the path is reachable: `(eval '(define-syntax k rules) (interaction-environment))` passes the `:38` immutable check (the wrapper is mutable), reaches `body_macro_depth == 0` with `lib_env` = `vm.globals` and `lib_env_val` = the wrapper, and enrolls it — after which every minor re-walks the whole globals map (mark + prune) until process exit, the exact cost the opcode guard now avoids.

Rare path, safety unaffected (over-approximation is safe) — but the rule the doc now states ("nor into the interaction-environment wrapper, whose map is vm.globals — the `.owned == false` case the handlers skip") is contradicted by these two sites. Suggest the same `and types.toEnvironment(self.lib_env_val).owned` at both, or a tiny shared `envStoreBarrier(gc, env_val, val)` so the rule lives once. Fine as a follow-up; not worth blocking the merge.

baijum:
**LGTM** — f78f3ee4 resolves all 10 review findings, and I verified each fix against the code rather than the commit message:

**Correctness (the four confirmed UAFs)**
- **`Function.global_cache`**: barriers added at all six store sites — `get_global` (placed after the if/else so the fresh-cache path is covered too), `set_global`/`define_global` cache fills, both `call_global` forms, and `tail_call_global`. Verified placement covers both cache-hit and fresh-cache branches at each site.
- **Retired-fiber window**: `saveCurrentFiber` now enrolls an old fiber wholesale via `GC.rememberObject`, which covers every save site (park, switch, and the retire-after-`retireSlot` paths) uniformly; `pruneRememberedSet` ages the entry out. The comment correctly frames it as additive to the resident `markFiberState` pass.
- **`tx.def_env_val`** (both the plain and begin-wrapped paths) and **`tx.let_syntax_peer_vals`**: transformer-side barriers added, with comments capturing the real root cause (SRFI 147/211 aliasing hands back pre-existing, possibly promoted transformers).

**Efficiency / consistency**
- The `set_global`/`define_global` env barriers now skip the interaction-environment wrapper via `.owned == false` — verified that `primitives_r7rs.zig:458` is the *only* `owned=false` allocation site, so the discriminator is exact. One residual: the two compile-time `lib_env.put` barriers in `compiler_define_syntax.zig` (:100, :726) kept the unguarded form — see the inline nit; fine as a follow-up.
- The `setParameterValue` comment no longer inverts the authoritative mechanism, and now gives the barrier a real job (the non-resident window) consistent with `gc_collect.zig`'s `.fiber` arm and the doc.

**Altitude / reuse**
- `GC.appendFunctionConstant` funnels compiler + deserializer + all ~28 test sites; the five no-op immediate barriers are gone with it.
- `GC.rememberObject` is now the single owned-path enrollment (writeBarrier's owned branch, promotion scan, sweepOld re-scan, saveCurrentFiber) — the #2196 dedup contract lives in one place. The sweepOld re-scan's dropped flag pre-check is behavior-identical (flags are clear post-drain; the helper checks anyway).
- `clearOldMarks` deleted outright, the mark-bit invariant is stated once in `minorCollect`, and `minor_marking` is now cleared via `defer` — the invariant is structural rather than by convention. `fullCollect`'s reset keeps a local justification instead of the impossible "panic path" one.

**Tests** — the every-fix-needs-a-test rule is met with the right split: real-path pins where the path is drivable from tests (`invokeGuardian` through `(g obj)` on a promoted guardian; `define`/`set!` through the real opcodes into a promoted mutable env), and mechanism pins with discriminating controls (barrier-vs-no-barrier) for the constants funnel, the global_cache store shape, and the retired-fiber enrollment. Each control proves the test fails without the fix.

Branch compiles clean here; commit reports `zig build test` + `-Dgc-stress=true` green and R7RS 1395/1395. Nothing blocking — good to merge.

src/compiler_define_syntax.zig
baijum: Fixed in eed30fab, with your second suggestion: `GC.envStoreBarrier(gc, env_val, val)` now owns the rule (`isEnvironment and .owned` → `writeBarrier` on the wrapper), and all four sites — both opcode handlers and both compiler stores — call it, so the exclusion lives exactly once. Pinned by your exact repro in `tests_core_eval.zig`: a promoted interaction-environment wrapper takes a define-syntax through eval, must not appear in the remembered set, and the macro must still expand afterwards (the store lands; only the enrollment is skipped).

src/tests_core_eval.zig
```diff
@@ -922,3 +922,86 @@ test "255-argument calls work on every dispatch path; 256 params reject cleanly
...
+    )));
+}
+
+// #1961 (review follow-up): define-syntax through the interaction-
+// environment must NOT enroll its wrapper in the remembered set — the
+// wrapper's map is the root-marked globals map, and enrolling it would
+// re-walk every global once per minor until process exit. GC.envStoreBarrier
+// is the one place that exclusion lives; this is the reviewer's reachable
+// path (mutable wrapper, top-level depth, lib_env = vm.globals).
+test "define-syntax via interaction-environment does not enroll the wrapper (#1961)" {
```
baijum, line 986: **[nit — optional follow-up] The scans can still enroll the wrapper; this pin holds by a hair**

`envStoreBarrier` closes the *barrier* route, but the wrapper has two other routes into the remembered set that don't consult `.owned`: the promotion scan (`sweepYoung` → `referencesYoung`) and the full-collect re-scan (`sweepOld`). `referencesYoung`'s `.scheme_environment` arm (gc_collect.zig:295-301) walks the whole map with no `.owned` check — so a promoted wrapper is enrolled the moment *any* globals value is young at promotion or full-collect time, and the entry never prunes while any global stays young. In a real program that's the common case (young globals are constantly being defined), so the O(#globals)-per-minor walk this commit exists to prevent still arrives through the scans; `markVmRoots` already marks every globals value each collection, so the walk is pure redundancy.

That's also why this test is a bit fragile: it passes because at the wrapper's promotion moment every other globals value is already old (the builtins promoted during init, `ie`'s own entry reads as old because `sweepYoung` sets `generation = 1` before running the scan). Add one young global binding before the two manual collections and the promotion scan enrolls the wrapper, failing the assertion for a reason unrelated to `envStoreBarrier`.

Suggested completion, if the cost is deemed worth it: have `referencesYoung`'s `.scheme_environment` arm return `false` for `.owned == false` wrappers (their map is root-marked by the owning VM; for a child-thread wrapper the values are foreign and untraced by that GC anyway), which stops both scan enrollments — and optionally skip the map walk in the two mark arms for the same reason. That would also make this test's pin robust rather than order-dependent. Fine as a separate PR; nothing here is unsafe — enrollment is over-approximation, cost only.

baijum:
**LGTM on eed30fab** — the nit is resolved exactly as suggested, and better: the rule now lives once in `GC.envStoreBarrier` (isEnvironment + `.owned` → `writeBarrier` on the wrapper), called from all four sites — both opcode handlers and both compiler `lib_env.put` stores — so the next env-map writer can't get the exclusion wrong. Verified:

- The helper's semantics are identical to the previous inline guard at the opcode sites (no behavior change there), and the compiler sites' change is exclusion-only — the wrapper's map is the root-marked globals, so skipping enrollment there is safe by the same argument as the opcodes.
- The `set_global` barrier stays scoped to the successful-store branch, as before.
- The regression test drives my exact repro — a promoted interaction-environment wrapper takes a `define-syntax` through `eval`, must not appear in the remembered set, and the macro still expands afterwards — and the doc now states the single-owner rule.

One optional follow-up noted inline on the test: the promotion scan and the `sweepOld` re-scan don't consult `.owned` (`referencesYoung`'s `.scheme_environment` arm walks the map unconditionally), so a promoted wrapper is still enrolled whenever any globals value is young at those moments — the common case in a live program — which re-opens the same O(#globals)-per-minor cost through a different door, and is also what makes the new test's pin order-dependent rather than structural. Cost only, never unsafe; fine as a separate PR.

Nothing blocking. With three rounds now converged — the generational mark itself, the six-site barrier audit closure, the shared helpers (`rememberObject`, `appendFunctionConstant`, `envStoreBarrier`), and real-path regression pins for each fix — this is good to merge.

src/tests_core_eval.zig
baijum: Done in 4960a534 — folded into this PR rather than a follow-up, since the fragile test was mine. `referencesYoung`'s `.scheme_environment` arm now returns false for `.owned == false` (verified that interaction-environment at primitives_r7rs.zig:458 is the only such constructor), which stops both scan enrollments and also drops any stale wrapper entry at the next prune. I left the two mark arms walking the map as before — as you note it is pure redundancy, but it is idempotent and bounded (a young wrapper for at most two minors; an old one only in fulls), and skipping them buys little next to the enrollment fix. The test now defines a young global between the wrapper's two promotion collections, so the promotion-scan moment genuinely sees a young value in the wrapper's map — the previous order-luck is gone, and reverting either the scan fix or envStoreBarrier flips it.

docs/dev/gc-safety-and-error-handling.md
```diff
@@ -99,24 +99,81 @@ keeps it alive.
...
-
-where `container` is the heap object being mutated. The barrier records
-the old→young edge in the remembered set. Omitting it does not fail
-immediately — it corrupts the heap only when a minor collection happens
-to run before the next full collection, which is why these bugs surface
-as rare, allocation-pattern-sensitive crashes.
+Minor collections are generational (#1961): the minor mark traces roots
+into the young generation and treats old objects as opaque (`GC.minor_marking`
+gates the check in `markValueInner`), so it costs O(live young) rather than
+O(live heap). An old→young reference — a field of an old heap object holding
```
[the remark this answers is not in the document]
baijum: Fixed in ad22b52f, in all three places the claim appeared (the dev doc's write-barrier section, `GC.minor_marking`'s field comment, and minorCollect's own comment): the cost is now stated as O(live young) plus the remembered containers' scanned fields, with the doc noting why a large old container with one young referent still costs its capacity to re-scan — which is exactly what the #2196 dedup and the prune protect against.

baijum:
Thanks! One note on the LGTM's "optional follow-up": the promotion-scan/re-scan enrollment route it describes is already closed in 4960a534 (the commit after the eed30fab this review was written against) — `referencesYoung`'s `.scheme_environment` arm returns `false` for `.owned == false` wrappers, and the wrapper-exclusion test now plants a young global between the two promotion collections so the pin is structural, not order-dependent (reverting either the scan fix or `envStoreBarrier` flips it).

The only piece left open from the whole thread is the declined micro-optimization of skipping the map walk in the two mark arms for the same wrapper class — recorded as #2377 (priority: low) so it isn't lost. Nothing else outstanding; CI is 22/22 green on ad22b52f (the one openbsd-test failure was a QEMU flake — a comment-only commit that re-ran green on identical code).

PR #2375, opened 2026-08-27
Document that ~/.kaappi/lib shadows a checkout's lib/ for .sld work

### LINKED ISSUE #2352: Installed ~/.kaappi/lib silently shadows a checkout's lib/, so local .sld edits look like no-ops
## Summary

When developing a `.sld` library in a checkout (e.g. editing `lib/srfi/231/views.sld`)
and testing it with `zig-out/bin/kaappi file.scm`, an existing install at
`~/.kaappi/lib` **silently shadows the checkout's own `lib/`**, so the edits appear
to have no effect. Rebuilding and `kaappi cache clear` don't help, because the wrong
source file is being loaded — not a cached copy of the right one.

## Mechanism

The default library search order (`src/main.zig`, ~L495–580) is:

1. the script's own directory
2. `~/.kaappi/lib`   ← the last installed release
3. exe-relative `<exe>/../lib` (= `zig-out/lib`, which `zig build` populates from the checkout)

Step 2 is checked before step 3 so that a from-source binary never shadows an
existing install (#1523) — good for users, a trap for library development: the
exe-relative fallback *does* carry your edits, but the stale install wins.

## Reproduce

```
# with an install present at ~/.kaappi/lib from a previous release
$ echo '(import (srfi 231 views)) (display "edited?")' > /tmp/t.scm
# edit lib/srfi/231/views.sld, then:
$ zig build && zig-out/bin/kaappi /tmp/t.scm     # runs the OLD ~/.kaappi/lib copy
$ KAAPPI_HOME=$(mktemp -d) zig-out/bin/kaappi /tmp/t.scm   # runs the checkout's copy
```

`tests/scheme/run-all.sh` does not set `KAAPPI_HOME`, so locally it, too, tests the
stale install rather than the working tree. CI is unaffected (no `~/.kaappi/lib`
there), which is exactly why this only bites local `.sld` work and can go unnoticed.

## Suggested fixes (pick one)

- Have `run-all.sh` (and the docs' local-dev instructions) run with an isolated
  `KAAPPI_HOME` so the suite always exercises the working tree.
- Emit a one-line notice (or `kaappi doctor` check) when a `(srfi N)` / `(kaappi …)`
  import resolves from `~/.kaappi/lib` while an exe-relative `lib/` also has it and
  the two differ.
- Document the hazard next to the build/test instructions.

Same hazard class as #1927 (a shadow that falls through silently so an A/B test
looks like it ran); surfaced while verifying #2350 / PR #2351.

baijum:
## The hazard (#2352)
The library search order checks `$KAAPPI_HOME/lib` (default `~/.kaappi/lib`) **before** the exe-relative `<exe>/../lib` (= `zig-out/lib`, populated from the checkout by `zig build`). That order is deliberate — a from-source binary must never shadow a user's install (#1523) — but it means a developer who has ever installed Kaappi has their working-tree `.sld` edits silently shadowed by the last release. A bare `zig-out/bin/kaappi file.scm` loads the *old* installed copy, so the edit looks like a no-op, and neither a rebuild nor `kaappi cache clear` helps, because it's the wrong source file being read — not a stale cache of the right one.

## Scope of the fix
This is a **doc-truth** close. The issue's suggested option 1 (isolate the suite's `KAAPPI_HOME`) turned out to be **already done** on `main`: `run-all.sh` exports a fresh `mktemp` home for the whole run (added for cache isolation, #1516), so the suite already exercises the working tree. The remaining gap was documentation of the hazard for ad-hoc interactive `.sld` work, plus making sure the suite's existing isolation isn't later narrowed to cache-only.

Changes:
- **`CLAUDE.md`** and **`docs/dev/testing.md`** — document the search order, the trap, and the `KAAPPI_HOME=$(mktemp -d) zig-out/bin/kaappi file.scm` workaround; note that the suite already isolates so only ad-hoc invocations need the prefix, and that CI never has a `~/.kaappi/lib` (why it's local-only and easy to miss).
- **`tests/scheme/run-all.sh`** — comment only: pin the existing isolation as load-bearing for #2352 ("Do NOT narrow this to a cache-only isolation").

The search order itself is **not** changed — reordering would reintroduce #1523. No `.zig` touched.

## Verification
- `markdownlint-cli2` (the CI `format` gate): 0 issues.
- Confirmed against `origin/main` that `run-all.sh` already exports a throwaway `KAAPPI_HOME` with a cleanup trap, so the suite half of the issue is already resolved.

Closes #2352

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -326,6 +326,25 @@ bash tools/run-r7rs-suite.sh zig-out/bin/kaappi         # R7RS suite (1,395 test
 bash tests/scheme/run-all.sh                            # everything (build FIRST)
 ```
 
+**An installed `~/.kaappi/lib` shadows the checkout's `lib/` for `.sld`
+resolution.** The library search order is (1) the script's own dir, (2)
+`$KAAPPI_HOME/lib` (default `~/.kaappi/lib`), (3) the exe-relative
+`<exe>/../lib` (= `zig-out/lib`, which `zig build` populates from the checkout).
+Step 2 before step 3 is deliberate — a from-source binary must never shadow a
+user's install (kaappi#1523) — but it means that if you have ever installed
+Kaappi, `~/.kaappi/lib` wins over your working-tree edits, so a change to
+`lib/**/*.sld` tested with a bare `zig-out/bin/kaappi file.scm` can look like a
+no-op no matter how many times you rebuild or `kaappi cache clear` (kaappi#2352).
+Run local `.sld` work with an isolated home so the checkout wins:
+
+```bash
+KAAPPI_HOME=$(mktemp -d) zig-out/bin/kaappi file.scm
+```
+
+`tests/scheme/run-all.sh` already does this (a fresh `mktemp` home per run), so
+the suite always exercises the working tree; only ad-hoc invocations need the
+prefix.
+
 The unit suite must also stay green under `zig build test -Dgc-stress=true`
 (collection on every allocation — kaappi#1401). Tests holding heap values in
 Zig locals across allocations must root them; loop-heavy tests should scale
--- a/docs/dev/testing.md
+++ b/docs/dev/testing.md
@@ -219,6 +219,37 @@ bash tests/scheme/run-all.sh
 bash tools/run-r7rs-suite.sh zig-out/bin/kaappi
 ```
 
+### An installed `~/.kaappi/lib` shadows the checkout's `lib/`
+
+The library search order is (1) the script's own directory, (2)
+`$KAAPPI_HOME/lib` (default `~/.kaappi/lib`), (3) the exe-relative
+`<exe>/../lib` — which is `zig-out/lib`, populated from the checkout by
+`zig build`. Step 2 is checked before step 3 on purpose, so a from-source
+binary never shadows a real install (kaappi#1523).
+
+The trap for library development (kaappi#2352) falls straight out of that
+order: if you have ever installed Kaappi, `~/.kaappi/lib` holds the last
+release's `.sld` files, and it wins over your working tree. So editing, say,
+`lib/srfi/231/views.sld` and running
+
+```bash
+zig build && zig-out/bin/kaappi /tmp/t.scm
+```
+
+loads the **old** installed copy — the edit looks like a no-op, and neither
+another rebuild nor `kaappi cache clear` helps, because it is the wrong source
+file being read, not a stale cache of the right one. Run local `.sld` work with
+an isolated home so the checkout's `lib/` (via `zig-out/lib`) wins:
+
+```bash
+KAAPPI_HOME=$(mktemp -d) zig-out/bin/kaappi /tmp/t.scm
+```
+
+`run-all.sh` already exports a fresh `mktemp` `KAAPPI_HOME` for the whole run,
+so the suite always exercises the working tree — this only bites ad-hoc
+`zig-out/bin/kaappi file.scm` invocations. CI never has a `~/.kaappi/lib`, which
+is why the hazard is local-only and easy to miss.
+
 ### The official SRFI 231 conformance suite
 
 `tests/scheme/srfi/srfi231-official.scm` is the SRFI's **official** test
--- a/tests/scheme/run-all.sh
+++ b/tests/scheme/run-all.sh
@@ -42,6 +42,17 @@ fi
 # Isolate ~/.kaappi so the suite is hermetic: plain `kaappi <file>` runs write a
 # bytecode cache under $KAAPPI_HOME/cache (kaappi#1516), and we don't want the
 # suite reading or polluting the developer's real cache (or history/config).
+#
+# The fresh (empty) home is also what keeps the suite exercising THIS working
+# tree's `.sld` libraries rather than a stale install: the library search order
+# is (1) the script's own dir, (2) $KAAPPI_HOME/lib, (3) the exe-relative
+# <exe>/../lib (= zig-out/lib, which `zig build` populates from the checkout).
+# Step 2 is checked before step 3 by design, so a from-source binary never
+# shadows a real install (kaappi#1523) — but that means a developer's own
+# ~/.kaappi/lib would silently shadow their checkout edits (kaappi#2352). A
+# throwaway home has no lib/, so step 3 wins and the corpus runs against the
+# tree under test. Do NOT narrow this to a cache-only isolation.
+#
 # Tests that need their own HOME (e.g. exe-relative-lib-1523) override this.
 KAAPPI_HOME_TMP=$(mktemp -d /tmp/kaappi-test-home-XXXXXX)
 export KAAPPI_HOME="$KAAPPI_HOME_TMP"


docs/dev/testing.md
```diff
@@ -219,6 +219,37 @@ bash tests/scheme/run-all.sh
 bash tools/run-r7rs-suite.sh zig-out/bin/kaappi
 ```
 
+### An installed `~/.kaappi/lib` shadows the checkout's `lib/`
+
+The library search order is (1) the script's own directory, (2)
```
baijum, line 224: This states a three-step order, but two rungs are missing: explicit `--lib-path` entries are searched *before* all three of these (`main.zig` prepends them ahead of the auto-discovered dirs), and the cwd-relative `""`/`lib/` fallbacks come last (`resolveLibraryPath`'s `builtin_prefixes`). `doctor.zig`'s `collectLibrary` numbers the real order as four steps. Since this paragraph is what people will cite for resolution behavior, worth either qualifying it ("the auto-added entries — after any `--lib-path`, before the cwd fallbacks") or listing everything.

Related suggestion: `--lib-path` is arguably the better workaround for this exact trap than a throwaway home —

```bash
zig-out/bin/kaappi --lib-path zig-out/lib /tmp/t.scm
```

puts the checkout ahead of the install with no temp dir to leak, and there is no stale-cache risk (cache entries are keyed by resolved path + source hash, so the checkout's file gets its own entry). The CI thottam step already uses precisely this idiom to make `$KAAPPI_HOME/lib` win. Documenting both and letting the reader pick seems strictly better.

(The same three-step paragraph appears in `CLAUDE.md` L329–331 and in the new `run-all.sh` comment — if you reword here, those two copies should follow.)

docs/dev/testing.md
```diff
@@ -219,6 +219,37 @@ bash tests/scheme/run-all.sh
 bash tools/run-r7rs-suite.sh zig-out/bin/kaappi
 ```
 
+### An installed `~/.kaappi/lib` shadows the checkout's `lib/`
```
baijum, line 222: Since this closes a `priority: medium` **bug** with documentation alone, the issue's detection option stays unimplemented and unfiled: warn when an import resolves from `~/.kaappi/lib` while the exe-relative `lib/` also has the file and the two differ (issue option 2). `kaappi doctor` already walks `lib_paths` and prints each entry (`collectLibrary`), so it is the natural home for that probe. Could you file a follow-up issue for it so the detection half doesn't evaporate when #2352 auto-closes? The trap remains exactly as silent for the next person who skips this section.

docs/dev/testing.md
```diff
@@ -219,6 +219,37 @@ bash tests/scheme/run-all.sh
...
+another rebuild nor `kaappi cache clear` helps, because it is the wrong source
+file being read, not a stale cache of the right one. Run local `.sld` work with
+an isolated home so the checkout's `lib/` (via `zig-out/lib`) wins:
+
+```bash
+KAAPPI_HOME=$(mktemp -d) zig-out/bin/kaappi /tmp/t.scm
+```
+
+`run-all.sh` already exports a fresh `mktemp` `KAAPPI_HOME` for the whole run,
+so the suite always exercises the working tree — this only bites ad-hoc
```
baijum, line 249: One suite command escapes this reassurance: `tools/run-r7rs-suite.sh` — listed as an equal-status command in the code block just above — does **not** isolate `KAAPPI_HOME`, unlike its siblings `tools/run-endian-suite.sh` and `tools/run-gc-stress-suite.sh`, which both carry the `mktemp` + trap guard for the same #1516 reason. Run directly (rather than via `run-all.sh`, whose export it inherits), it uses the developer's real home. The r7rs corpus itself imports only built-in `(scheme ...)` libraries, so the *shadowing* half cannot bite it — but "the suite always exercises the working tree" silently excludes this one wrapper, and the cache-pollution half of the isolation rationale does apply. Either scope the sentence ("`run-all.sh` and everything it invokes") or give the wrapper the same three-line guard its siblings have — happy to do the latter as a follow-up PR.

CLAUDE.md
```diff
@@ -326,6 +326,25 @@ bash tools/run-r7rs-suite.sh zig-out/bin/kaappi         # R7RS suite (1,395 test
...
+`lib/**/*.sld` tested with a bare `zig-out/bin/kaappi file.scm` can look like a
+no-op no matter how many times you rebuild or `kaappi cache clear` (kaappi#2352).
+Run local `.sld` work with an isolated home so the checkout wins:
+
+```bash
+KAAPPI_HOME=$(mktemp -d) zig-out/bin/kaappi file.scm
+```
+
+`tests/scheme/run-all.sh` already does this (a fresh `mktemp` home per run), so
+the suite always exercises the working tree; only ad-hoc invocations need the
```
baijum, line 345: Same claim as docs/dev/testing.md L249 — my comment there applies to this copy too: `tools/run-r7rs-suite.sh`, listed just above, is the one suite command that doesn't isolate `KAAPPI_HOME` when run directly.

CLAUDE.md
```diff
@@ -326,6 +326,25 @@ bash tools/run-r7rs-suite.sh zig-out/bin/kaappi         # R7RS suite (1,395 test
...
+`<exe>/../lib` (= `zig-out/lib`, which `zig build` populates from the checkout).
+Step 2 before step 3 is deliberate — a from-source binary must never shadow a
+user's install (kaappi#1523) — but it means that if you have ever installed
+Kaappi, `~/.kaappi/lib` wins over your working-tree edits, so a change to
+`lib/**/*.sld` tested with a bare `zig-out/bin/kaappi file.scm` can look like a
+no-op no matter how many times you rebuild or `kaappi cache clear` (kaappi#2352).
+Run local `.sld` work with an isolated home so the checkout wins:
+
+```bash
+KAAPPI_HOME=$(mktemp -d) zig-out/bin/kaappi file.scm
```
baijum, line 341: Nit: this incantation leaks one temp dir (with a bytecode cache) per ad-hoc run — every suite script cleans its `mktemp` home up with a trap. If the `--lib-path` alternative from my testing.md comment lands, it sidesteps this entirely; otherwise maybe a parenthetical acknowledging the leak is deliberate for one-off use. Same for the block in testing.md L245.

tests/scheme/run-all.sh
```diff
@@ -42,6 +42,17 @@ fi
...
+#
+# The fresh (empty) home is also what keeps the suite exercising THIS working
+# tree's `.sld` libraries rather than a stale install: the library search order
+# is (1) the script's own dir, (2) $KAAPPI_HOME/lib, (3) the exe-relative
+# <exe>/../lib (= zig-out/lib, which `zig build` populates from the checkout).
+# Step 2 is checked before step 3 by design, so a from-source binary never
+# shadows a real install (kaappi#1523) — but that means a developer's own
+# ~/.kaappi/lib would silently shadow their checkout edits (kaappi#2352). A
+# throwaway home has no lib/, so step 3 wins and the corpus runs against the
+# tree under test. Do NOT narrow this to a cache-only isolation.
```
baijum, line 54: Checked the invariants this comment pins against the code: the order and the #1523 rationale match `main.zig`'s `lib_paths` seeding, and an empty throwaway home does make the `$KAAPPI_HOME/lib` probe fall through to `zig-out/lib`. Good pin — this is exactly the comment that stops a future "it's only about the cache" cleanup from silently narrowing it.

baijum (commented):
Verified every factual claim against the code before judging the prose — it all checks out:

- **Search order**: `src/main.zig` seeds `lib_paths` as script dir → `$KAAPPI_HOME/lib` → `<exe>/../lib`, and the in-code rationale matches verbatim ("Checked after `~/.kaappi/lib` so an existing install is never shadowed", #1523). `build.zig:193` (`installDirectory`, `.sld`/`.scm`) confirms `zig build` populates `zig-out/lib`.
- **"Already done on `main`"**: correct — `run-all.sh` has exported a throwaway `KAAPPI_HOME` (with cleanup trap) since #1576/#1516, so issue #2352's "`run-all.sh` does not set `KAAPPI_HOME`" was stale. Pinning that isolation as load-bearing is the right move.
- **Cache claim**: correct — `.sld` cache entries key on the resolved path + source hash (`vm_library_cache.zig` header), so `kaappi cache clear` genuinely cannot help when resolution picks the installed file.
- **CI claim**: holds — the thottam step runs under its own `mktemp` `KAAPPI_HOME` and removes it; nothing else in the PR matrix installs into `~/.kaappi`.
- **`markdownlint-cli2`** with the repo's `.markdownlint-cli2.jsonc`: 0 issues on both markdown files.

The comments below are suggestions (one scoping nit, one follow-up, a documented alternative) — nothing blocking.

PR #2376, opened 2026-08-27
Wire up kaappi.pkg name: and source:; drop version: from the grammar

### LINKED ISSUE #2138: [Doc-truth] kaappi.pkg 'source:' and 'name:' are parsed and never read — the documented third-party hosting field is inert

baijum:
## The gap (#2138)
The `kaappi.pkg` manifest documented four fields, but `thottam` read only `depends:` and `build:`. `name:` and `source:` were **inert** — a manifest could name a different package, or declare a canonical source, and the install neither checked nor recorded either. `version:` was never a field at all.

(Note: the issue described `name:`/`source:` as "parsed but never read"; on current `main` the parser had since been narrowed to two keys, so they weren't even parsed. Same doc-truth gap, addressed the same way.)

## What this does — the "wire up" direction
Per the design call in the issue's **option 1**, respecting the clone-first architecture and the #2137 provenance model:

- **`name:`** — consistency-checked against the package being installed. A manifest naming a different package is a packaging error; the install **refuses before recording anything**. Still optional (a manifest may omit it), but a mismatching `name:` is never accepted.

  ```
  $ thottam install kaappi-badname     # manifest says name: WRONG-NAME-ENTIRELY
  error: manifest declares name 'WRONG-NAME-ENTIRELY' but installing 'kaappi-badname'
  ```

- **`source:`** — declares the canonical host. On a **bare-name install** (no `::url`) it's recorded as the lockfile's provenance, so `thottam list` shows `(from: …)` and a later `--locked` install fetches from it — **surfaced with a one-line note**, never silent, because the manifest is then steering where the package comes from (the interaction the issue flags). When a `::url` is given and disagrees, thottam **warns** (a fork/mirror is legitimate) and records the URL it actually fetched, not the manifest's claim. `--locked` mode is untouched: the lockfile stays authoritative and `source:` is ignored (#2137).

  **Honest limitation, documented:** `source:` **cannot redirect the first clone** — thottam reads the manifest only after cloning the repo, so the initial fetch URL is always the command-line `::url` or the `KAAPPI_ORG` default.

- **`version:`** — dropped from the documented grammar. thottam locks by git SHA, so a manifest version means nothing; it keeps being ignored like any unknown key.

## Regression tests
- `src/tests_thottam.zig`: the parser tests now pin all four read fields plus the empty-value-is-absent rule.
- `tests/scheme/thottam/thottam-lifecycle.sh`: scenario 10 rewritten from "these fields are inert" to cover the name-mismatch refusal (with lockfile untouched), provenance recording + its note, the `::url`-vs-`source:` warning (fetched URL wins), and the no-`name:` install. **152 checks pass** (11 new).

## Verification
- `zig build test` — passes (full suite).
- `bash tests/scheme/thottam/thottam-lifecycle.sh` — 152/0.
- markdownlint (CI `format` gate) on `docs/dev/thottam.md` — 0 issues.
- Manual: verified name-mismatch error, bare-name source recording + `(from:)` in `list`, `::url` mismatch warning, and `--locked` fetching from the lockfile source (authority preserved).

Docs: `docs/dev/thottam.md` rewritten to describe the four read fields, the provenance/first-clone semantics, and the absence of `version:`. (The workspace-root `CLAUDE.md` the issue also quotes is not in this repo — it's a local, untracked orientation file — so it can't be included here; it's the same grammar and should be updated wherever it's maintained.)

Closes #2138

### DIFF
--- a/docs/dev/thottam.md
+++ b/docs/dev/thottam.md
@@ -74,7 +74,12 @@ manifest would mean nothing; it is dropped like any unknown key (kaappi#2138).
   trailing `/` and a `.git` suffix — those are cosmetic spellings of one
   remote — but a scheme difference (`http` vs `https`) is a real downgrade and
   still warns. In `--locked` mode the lockfile is authoritative and the
-  manifest's `source:` is ignored (#2137).
+  manifest's `source:` is ignored (#2137). Because a manifest is
+  attacker-controlled and its `source:` is later handed to `git clone`, a
+  declared source that uses a git remote-helper transport (`ext::<cmd>` runs a
+  command) or is option-shaped (leading `-`) is refused rather than recorded;
+  the clone itself also runs with `protocol.ext.allow=never`, so the
+  OS-command-injection vector is closed at both ends (CWE-78).
 - `depends` is space-separated, and may carry an inline custom URL:
   `depends: kaappi-net kaappi-auth::https://github.com/bob/kaappi-auth`.
 - Version constraints use `>=`, `>`, `<=`, `<`, `^` (compatible), `~`
--- a/docs/dev/thottam.md
+++ b/docs/dev/thottam.md
@@ -20,7 +20,8 @@ thottam remove kaappi-web                                        # uninstall
 
 - Clones from `github.com/kaappi/<package>` (or a custom `::url`) into
   `$KAAPPI_HOME/src/` (`~/.kaappi/src/` by default).
-- Reads `kaappi.pkg` for dependencies and build commands.
+- Reads `kaappi.pkg` for the package name, dependencies, build command, and
+  declared source.
 - Copies `.sld` files to `~/.kaappi/lib/`, preserving directory structure.
 - Copies `.dylib`/`.so` to `~/.kaappi/lib/`.
 - Records source URLs in the lockfile `~/.kaappi/thottam.lock`, for provenance.
@@ -43,17 +44,31 @@ directory. See the LLVM backend section of `CLAUDE.md`.
 ## The `kaappi.pkg` manifest
 
 ```text
+name: kaappi-web
 depends: kaappi-http kaappi-json
 build: make
+source: https://github.com/someone/kaappi-web
 ```
 
-Only these two fields are read. `name:`, `version:`, `source:` and any other
-key are ignored — a package is identified by the name it is installed under,
-the version is whatever tag or SHA the install requested, and a third-party
-source is given on the command line (`thottam install pkg::url`) or in a
-`depends:` spec, never in the package's own manifest (which thottam can only
-read after it has already cloned the repo) (kaappi#2138).
+Four keys are read: `name:`, `depends:`, `build:`, and `source:`. There is no
+`version:` field — thottam locks by git SHA, so a version written in the
+manifest would mean nothing; it is dropped like any unknown key (kaappi#2138).
 
+- `name`, when present, must match the package being installed. A manifest that
+  names a different package is a packaging error, and the install refuses
+  before recording anything. It is optional — a manifest may omit it — but a
+  mismatching `name:` is never accepted.
+- `source` declares where the package is canonically hosted. On a bare-name
+  install (no `::url` on the command line) it is recorded as the lockfile's
+  provenance, so `thottam list` shows `(from: …)` and a later `--locked`
+  install fetches from it — surfaced with a one-line note at record time, since
+  the manifest is then steering where the package comes from. It **cannot**
+  redirect the *first* clone: thottam can only read the manifest after cloning
+  the repo, so the initial fetch URL is always the `::url` given on the command
+  line or the `KAAPPI_ORG` default. When a `::url` *is* given and disagrees with
+  `source:`, thottam warns (a fork or mirror is legitimate) and records the URL
+  it actually fetched from, never the manifest's claim. In `--locked` mode the
+  lockfile is authoritative and the manifest's `source:` is ignored (#2137).
 - `depends` is space-separated, and may carry an inline custom URL:
   `depends: kaappi-net kaappi-auth::https://github.com/bob/kaappi-auth`.
 - Version constraints use `>=`, `>`, `<=`, `<`, `^` (compatible), `~`
--- a/docs/dev/thottam.md
+++ b/docs/dev/thottam.md
@@ -56,8 +56,11 @@ manifest would mean nothing; it is dropped like any unknown key (kaappi#2138).
 
 - `name`, when present, must match the package being installed. A manifest that
   names a different package is a packaging error, and the install refuses
-  before recording anything. It is optional — a manifest may omit it — but a
-  mismatching `name:` is never accepted.
+  before recording anything (removing the checkout it just cloned, so nothing
+  unrecorded is left for a later install to reuse). It is optional — a manifest
+  may omit it — but a mismatching `name:` is never accepted. The check is scoped
+  to unlocked installs: in `--locked` mode the lockfile already vouches for the
+  exact SHA, so the manifest at that SHA is not re-litigated.
 - `source` declares where the package is canonically hosted. On a bare-name
   install (no `::url` on the command line) it is recorded as the lockfile's
   provenance, so `thottam list` shows `(from: …)` and a later `--locked`
@@ -67,8 +70,11 @@ manifest would mean nothing; it is dropped like any unknown key (kaappi#2138).
   the repo, so the initial fetch URL is always the `::url` given on the command
   line or the `KAAPPI_ORG` default. When a `::url` *is* given and disagrees with
   `source:`, thottam warns (a fork or mirror is legitimate) and records the URL
-  it actually fetched from, never the manifest's claim. In `--locked` mode the
-  lockfile is authoritative and the manifest's `source:` is ignored (#2137).
+  it actually fetched from, never the manifest's claim. The comparison ignores a
+  trailing `/` and a `.git` suffix — those are cosmetic spellings of one
+  remote — but a scheme difference (`http` vs `https`) is a real downgrade and
+  still warns. In `--locked` mode the lockfile is authoritative and the
+  manifest's `source:` is ignored (#2137).
 - `depends` is space-separated, and may carry an inline custom URL:
   `depends: kaappi-net kaappi-auth::https://github.com/bob/kaappi-auth`.
 - Version constraints use `>=`, `>`, `<=`, `<`, `^` (compatible), `~`
--- a/src/tests_thottam.zig
+++ b/src/tests_thottam.zig
@@ -509,6 +509,22 @@ test "sourcesEquivalent ignores trailing slash and .git, keeps scheme (#2138)" {
     try std.testing.expect(!eq("https://h/p", "https://h/q"));
 }
 
+test "isRecordableSource refuses remote-helper transports and option-shaped URLs (#2138)" {
+    const ok = thottam.isRecordableSource;
+    // Ordinary remotes and local paths are recordable.
+    try std.testing.expect(ok("https://github.com/foo/bar.git"));
+    try std.testing.expect(ok("git@github.com:foo/bar.git"));
+    try std.testing.expect(ok("/srv/git/bar.git"));
+    // A `<transport>::<address>` remote helper runs a command -> refused.
+    try std.testing.expect(!ok("ext::sh -c 'touch /tmp/pwned'"));
+    try std.testing.expect(!ok("fd::17"));
+    try std.testing.expect(!ok("transport::whatever"));
+    // A leading '-' would be read by `git clone` as an option -> refused.
+    try std.testing.expect(!ok("--upload-pack=evil"));
+    // Empty is not a source.
+    try std.testing.expect(!ok(""));
+}
+
 test "parsePkgManifest tolerates CRLF line endings" {
     // thottam runs on Windows under Git Bash; a manifest checked out with
     // core.autocrlf=true has \r before every \n.
--- a/src/tests_thottam.zig
+++ b/src/tests_thottam.zig
@@ -496,6 +496,19 @@ test "parsePkgManifest treats empty name: and source: as absent (#2138)" {
     try std.testing.expectEqualStrings("dep", m.depends.?);
 }
 
+test "sourcesEquivalent ignores trailing slash and .git, keeps scheme (#2138)" {
+    const eq = thottam.sourcesEquivalent;
+    // Cosmetic differences — the same remote — do not warn.
+    try std.testing.expect(eq("https://h/p", "https://h/p.git"));
+    try std.testing.expect(eq("https://h/p", "https://h/p/"));
+    try std.testing.expect(eq("https://h/p.git", "https://h/p/"));
+    try std.testing.expect(eq("https://h/p", "https://h/p"));
+    // A scheme downgrade is NOT cosmetic — it still warns.
+    try std.testing.expect(!eq("https://h/p", "http://h/p"));
+    // A genuinely different repository still warns.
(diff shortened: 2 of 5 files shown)


src/thottam.zig
```diff
@@ -799,9 +801,65 @@ fn doInstall(
...
+        // installed is a packaging error, not a fetch to silently record
+        // (kaappi#2138). `name:` was documented as the one required field;
+        // this is what finally enforces it.
+        if (m.name) |declared| {
+            if (!std.mem.eql(u8, declared, pkg)) {
+                var msg_buf: [512]u8 = undefined;
+                printErrColor(Color.red, "error: ");
+                const msg = std.fmt.bufPrint(&msg_buf, "manifest declares name '{s}' but installing '{s}'\n", .{ declared, pkg }) catch "manifest name mismatch\n";
+                writeStderr(msg);
+                return error.ManifestNameMismatch;
```
baijum, line 879: **[P2] `ManifestNameMismatch` is missing from the top-level exit list.** This error is not in the `catch` at line 1365 (`GitFailed or BuildFailed or CopyFailed or InvalidPackageName or GitNotFound`), so it falls through to `return err` and Zig's default handler prints a second line — the raw `error.ManifestNameMismatch` — exactly the duplicate-message pattern kaappi#2132 eliminated for `InvalidPackageName`. The lifecycle test passes anyway because it greps a substring out of combined output; run the failing install without `2>&1` filtering and you'll see both lines. Add it to the `std.process.exit(1)` set there, and ideally pin the absence of the raw error name in the test.

src/thottam.zig
```diff
@@ -799,9 +801,65 @@ fn doInstall(
...
+        // Reconcile the declared `source:` with how we actually fetched
+        // (kaappi#2138). A command-line `::url` is authoritative for the
+        // fetch; the manifest only informs provenance.
+        if (!locked_mode) {
+            if (m.source) |declared| {
+                if (parsed.source) |supplied| {
+                    // A fork or mirror sharing history is legitimate, so a
+                    // divergence warns rather than refuses — but it is never
+                    // silent, which is the whole point of the field.
+                    if (!std.mem.eql(u8, supplied, declared)) {
```
baijum: **[P3] Strict string compare fires on cosmetic URL differences.** A manifest declaring `https://github.com/foo/bar` vs an install of `pkg::https://github.com/foo/bar.git` (trailing slash, `.git` suffix, http-vs-https) warns about a "divergence" that is the same host and repo. Since a fork/mirror is exactly what this warning exists for, cosmetic mismatches are noise that trains users to ignore it. Either normalize before comparing (strip trailing `/` and `.git`) or document in `docs/dev/thottam.md` that `source:` must be written in the exact canonical form (`…/name.git`).

src/thottam.zig
```diff
@@ -799,9 +801,65 @@ fn doInstall(
...
+    // hosted (kaappi#2138).
+    var record_source: ?[]const u8 = if (locked_mode) locked_source else parsed.source;
+
+    if (manifest) |m| {
+        // A manifest that names a different package than the one being
+        // installed is a packaging error, not a fetch to silently record
+        // (kaappi#2138). `name:` was documented as the one required field;
+        // this is what finally enforces it.
+        if (m.name) |declared| {
+            if (!std.mem.eql(u8, declared, pkg)) {
```
baijum, line 869: **Question: is enforcing `name:` in `--locked` mode intended?** This check runs in every mode, so a `--locked` restore of a previously-installed package now fails if upstream later edits the manifest's `name:` — even though the lockfile vouches for the exact SHA, and `source:` is deliberately ignored in locked mode (#2137). The PR description says "--locked mode is untouched", which reads as though this were scoped to unlocked installs. Refusing here is defensible (a manifest that lies is a lie at any SHA), but it should be a stated decision — and worth a lifecycle check pinning the locked-mode behavior either way.

src/thottam.zig
```diff
@@ -799,9 +801,65 @@ fn doInstall(
...
+                        writeStderr(msg);
+                    }
+                } else {
+                    // Installed by bare name: record the declared home as
+                    // provenance so `thottam list` shows it and a later
+                    // --locked install fetches from it. Surfaced, not silent,
+                    // because the manifest now steers where the package comes
+                    // from — the interaction #2138 flags.
+                    record_source = declared;
+                    writeStdout("  note: recording declared source ");
```
baijum: **[P4] The note goes to stdout, the warning to stderr.** They tell the same reconciliation story at different severities, so they should reach the same stream — I'd send the note to stderr so `thottam install … > log` keeps the provenance notice with the diagnostics. Also minor: the note prints before `updateLockfile` runs, so a lockfile write failure leaves a note about something that didn't happen.

baijum (commented):
Overall the direction looks right, and the aliasing/lifetime handling for `record_source` (manifest kept alive past `updateLockfile` because the recorded provenance may alias `manifest.source`) is correctly reasoned and commented. The one thing I'd want fixed before merge is the missing `ManifestNameMismatch` arm in the top-level exit path.

src/thottam.zig
baijum, line 879: **[P4] A refused install leaves the clone behind.** By this point the repo is already in `$KAAPPI_HOME/src/<pkg>` but nothing is recorded in the installed list or lockfile — `thottam list`/`verify` don't know about it, and a later fixed install will silently reuse this checkout (the `dirExists` guard at the clone skips re-cloning). Harmless in the common case, but consider removing the checkout on this path or noting the leftover in the error message.

baijum: Fixed in 0c1fba8e — added `error.ManifestNameMismatch` to the install command's `std.process.exit(1)` set, so it exits cleanly with no raw second line. Pinned the raw name's absence in the lifecycle test ("the refusal prints no raw Zig error name").

src/thottam.zig
baijum: Fixed in 0c1fba8e — added `sourcesEquivalent`/`canonicalSource`: the compare now strips a trailing `/` and a `.git` suffix, so those cosmetic spellings no longer warn. A scheme difference (http vs https) stays significant — that's a downgrade, not cosmetic. Unit-tested in tests_thottam.zig and pinned in the lifecycle suite.

src/thottam.zig
baijum: Good catch — the description was wrong. Scoped both the name and source reconciliation to `!locked_mode` in 0c1fba8e, so "--locked mode is untouched" is now literally true: in locked mode the lockfile vouches for the SHA and the manifest at that SHA isn't re-litigated. Added a lifecycle check pinning that a name-mismatching manifest does NOT refuse a --locked restore, and documented the scoping in thottam.md.

src/thottam.zig
baijum: Fixed in 0c1fba8e — the note now goes to stderr, alongside the warning. Left it just before updateLockfile: a lockfile write failure returns an error that surfaces loudly, so the note being marginally early is cosmetic — happy to move it past updateLockfile if you'd prefer.

src/thottam.zig
baijum: Fixed in 0c1fba8e — the refusal path now removes the checkout, but only when this invocation created it (a `did_clone` flag), so an already-installed package's tree is never touched. Pinned in the lifecycle test ("the refused install leaves no checkout behind").

src/thottam.zig
```diff
@@ -799,9 +821,77 @@ fn doInstall(
...
+                    const msg = std.fmt.bufPrint(&w, "{s} declares source '{s}' but was fetched from '{s}'\n", .{ pkg, declared, supplied }) catch "declared source differs from fetch URL\n";
+                    writeStderr(msg);
+                }
+            } else {
+                // Installed by bare name: record the declared home as
+                // provenance so `thottam list` shows it and a later --locked
+                // install fetches from it. Surfaced on stderr, alongside the
+                // warning above, not silent — the manifest now steers where the
+                // package comes from, the interaction #2138 flags.
+                record_source = declared;
```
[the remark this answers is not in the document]
baijum: Fixed in 62523766 — this is the real one. `source:` is attacker-controlled and reaches `git clone` under --locked, so: (1) `isRecordableSource` refuses to record a source containing `::` (a `<transport>::<address>` remote helper like `ext::<cmd>`) or a leading `-`; (2) the clone runs with `-c protocol.ext.allow=never` so even a hand-edited lockfile can't execute a helper. Unit-tested (`isRecordableSource`) and a lifecycle check proves the `ext::` command never runs (a marker file stays absent).

tests/scheme/thottam/thottam-lifecycle.sh
```diff
@@ -688,34 +688,111 @@ check "--locked matching-source restore records the same provenance" \
...
+git clone -q --bare "$WORK/kaappi-declared" "$ORG/kaappi-declared.git"
+fresh_home
+out="$("$THOTTAM" install kaappi-declared 2>&1)" && ec=0 || ec=$?
+check_exit "a manifest with a valid name and a source installs" 0 "$ec"
+check "the declared source is surfaced with a note at record time" \
+    "note: recording declared source https://forge.example.org/kaappi-declared" "$out"
+check "the declared source reaches the lockfile as provenance" \
+    "https://forge.example.org/kaappi-declared" "$(cat "$KAAPPI_HOME/thottam.lock")"
+check "the declared source shows in list" \
+    "(from: https://forge.example.org/kaappi-declared)" "$("$THOTTAM" list)"
```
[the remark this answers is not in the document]
baijum: Fixed in 62523766 — added a lifecycle check that a --locked restore actually clones from the recorded source: the org default is moved out of the way and the restore still succeeds from the declared alternate remote (a byte-identical bare clone so the SHA matches), then asserts the library was fetched. The earlier checks only proved recording + list output.

docs/dev/thottam.md
```diff
@@ -20,7 +20,8 @@ thottam remove kaappi-web                                        # uninstall
 
 - Clones from `github.com/kaappi/<package>` (or a custom `::url`) into
   `$KAAPPI_HOME/src/` (`~/.kaappi/src/` by default).
-- Reads `kaappi.pkg` for dependencies and build commands.
+- Reads `kaappi.pkg` for the package name, dependencies, build command, and
+  declared source.
```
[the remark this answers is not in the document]
baijum: Declining — this is a false positive. `docs/dev/thottam.md` is the *existing* contributor doc for the package manager (CLAUDE.md lists it as "The package manager" under developer docs, and it already documented the manifest grammar before this PR). I edited existing sections in place; I didn't add end-user docs. The user-facing thottam reference lives separately in kaappi.github.io/docs/ecosystem/. Moving this file's manifest/behavior sections there would be out of scope and would strip the contributor doc of the grammar its own parser implements.

src/tests_thottam.zig
```diff
@@ -466,20 +466,47 @@ test "isValidPkgName accepts the shapes the ecosystem actually uses" {
...
+    const eq = thottam.sourcesEquivalent;
+    // Cosmetic differences — the same remote — do not warn.
+    try std.testing.expect(eq("https://h/p", "https://h/p.git"));
+    try std.testing.expect(eq("https://h/p", "https://h/p/"));
+    try std.testing.expect(eq("https://h/p.git", "https://h/p/"));
+    try std.testing.expect(eq("https://h/p", "https://h/p"));
+    // A scheme downgrade is NOT cosmetic — it still warns.
+    try std.testing.expect(!eq("https://h/p", "http://h/p"));
+    // A genuinely different repository still warns.
+    try std.testing.expect(!eq("https://h/p", "https://h/q"));
```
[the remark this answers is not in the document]
baijum: Declining — false positive. These are pure-function tests: `state.parsePkgManifest(...)` and `thottam.sourcesEquivalent(...)`/`isRecordableSource(...)`. There's no Scheme evaluation, so `th.expectEval*` (which compiles+runs Scheme source) doesn't apply, and there's no VM to wrap in `th.TestContext`. Every sibling test in tests_thottam.zig follows this exact shape — `state.parseX(...)` + `std.testing.expect*` — because it's the pure parsing/semver/URL surface. Matching the file's established convention.

Issue #2382, opened 2026-08-27
SRFI 231: c64/c128 storage classes reject the reference's f32/f64vector data representation in make-specialized-array-from-data


baijum:
### Summary

Surfaced while resolving #2379 (pruning the f16 divergence entries from `srfi231-official-transform.py`): the known-divergence entry for official-suite test id 150 was hiding **two unrelated failures** — the c64 and c128 rows — alongside the f16 row it was written for. With f16 now a real class (#2379), those two rows fail on their own:

```scheme
(make-specialized-array-from-data (make-f32vector 10) c64-storage-class)
;; ERROR: make-specialized-array-from-data: data is not compatible with storage-class
;; reference: returns a 5-element c64 array sharing the f32vector
```

### Cause

The SRFI 231 reference implementation has no native complex homogeneous vector, so it backs `c64-storage-class`/`c128-storage-class` with an **f32vector/f64vector of length 2n** (getter/setter split `real-part`/`imag-part` across adjacent elements; `data?` is `(and (f32vector? data) (even? (f32vector-length data)))` — `generic-arrays.scm`, `make-complex-storage-classes`). Kaappi ships real `c64vector`/`c128vector` (SRFI 160), and `lib/srfi/231/storage-classes.sld` deliberately builds the classes on them (`data?` = `c64vector?`/`c128vector?`).

The behavior is spec-legitimate — the spec lets `data?` define what the maker's own output type is, and the native representation is arguably better — but it is reference-coupled-test-divergent: the official suite's test 150 feeds even-length `f32vector`/`f64vector` fixtures as *good* data (and odd-length ones as bad, which still error correctly on our side).

### Current mitigation

`srfi231-official-transform.py`'s known-divergence table carries id 150 with this reason (re-scoped from the old f16 wording in #2379's PR).

### Options considered

- **Switch the classes to the reference's pair-of-floats representation** — closes the divergence fully but changes `array-body`'s user-visible type for every c64/c128 array and throws away the native SRFI 160 types; a design change, not a bug fix.
- **Accept both representations in `data?`/`data->body`** — impossible without dual-representation getters; the getter cannot index an f32vector with `c64vector-ref`.

Leaving as a documented divergence unless a concrete portability need (reference-coupled user code) shows up.

PR #2383, opened 2026-08-27
Implement SRFI 231 f16-storage-class as software half-floats

### LINKED ISSUE #2379: SRFI 231: f16-storage-class is #f though the u16vector substrate is shipped — port the reference's software half-floats
### Summary

`f16-storage-class` is bound to `#f` at `lib/srfi/231/storage-classes.sld:206` — a documented scope reduction carried over from #2353 (now closed; that issue shipped `u1` and deliberately deferred `f16`). The spec's escape clause — "Implementations with an appropriate homogeneous vector type should define the associated global variable using `make-storage-class`. Otherwise, they shall define the associated global variable to `#f`" (srfi-231.html, Storage classes) — is written for implementations *lacking* the substrate. That does not describe Kaappi: `(srfi 160 u16)` is shipped, and the reference implementation (`generic-arrays.scm` lines 1449–1601) implements `f16-storage-class` as **software half-floats over a u16vector** with a purely *arithmetic* codec — it uses no host bit-reinterpret facility anywhere, so it ports to R7RS-small directly.

Impact today: official-suite tests 98/148/149/150 sit in the known-divergence table (`srfi231-official-transform.py:732-735`), and the f16 rows at `srfi231-official.scm` lines 849, 861, 1116, 1150, 1428, 2199, 2268, 2461, 3819, 3909 are unexercised (the residue of the "12 direct failures + 33 unreachable sites" quantified in the #2353 discussion, of which u1's share was fixed there).

Context: in the srfi-231 mailing-list announcement of Kaappi's implementation (26 Aug 2026), the `f16-storage-class = #f` stub is one of exactly two limitations named — and the follow-up notes that Chibi, the other portable implementation, ships f16. Closing this closes the only substantive storage-class gap (f8 stays `#f`: the reference itself leaves it `#f`, since no standard 8-bit float type exists).

### Reproduction

```scheme
kaappi> (import (srfi 231))
kaappi> f16-storage-class
#f
kaappi> (make-specialized-array (make-interval '#(4 4)) f16-storage-class)
ERROR: error[KP3002]: type error in '%record-ref': expected <storage-class>, got #f
;; reference: returns a 4x4 array of half-floats; getter/setter round-trip
;; through IEEE 754 binary16 with round-to-nearest-even
```

### Fix shape

Faithful transliteration of the reference codec (`generic-arrays.scm` 1463–1566), hand-expanded for the one instantiation `(mantissa-width 10, exponent-width 5, bias 15)` — not a from-scratch codec. `%f16-decode`/`%f16-encode` as library-internal defines in `storage-classes.sld` (`%` names are free since #1856). Everything the codec needs is already imported there (`(srfi 160 u16)`, `(srfi 60)`).

Substitution table for the Gambit-isms (none is a bit-reinterpret):

| Gambit | pure R7RS |
|---|---|
| `##flonum->fixnum (flround ...)` | `(exact (round ...))` — R7RS `round` **is** ties-to-even, exactly the required semantics |
| `flscalbn x n` | `(* x (expt 2.0 n))` — n ∈ [−24, 24], power-of-two scaling is exact in f64, so the codec keeps its scale-exactly-round-**once** shape (no double rounding) |
| `flilogb` | small `%ilogb` loop — the only routine to write fresh; only finite nonzero x, only classification vs [−14, 15] plus the exact e matter |
| `##flcopysign` sign test | `(if (or (< x 0.0) (eqv? x -0.0)) 1 0)` — `eqv?` distinguishes −0.0 per R7RS; encode NaN as `#x7FFF` (NaN sign unobservable portably — benign deviation, decode yields `+nan.0` either way) |
| `fxand`/`fxior`/`fxarithmetic-shift-*` | `(srfi 60)` ops, already imported for u1 |
| `flfinite?`/`flnan?` | `(scheme inexact)` `finite?`/`nan?` |

Structure to preserve from the reference (this is what makes it correct):

- **separate subnormal branch** scaling by 2^24 directly — one rounding at the correct ulp (a shift-after-normal-form formula rounds twice);
- **mantissa-carry after rounding**: significand can round to exactly 2048 → bump exponent (exact, no re-round); at the top binade → ±inf. This is what sends 65520 (the tie) **up** to `+inf.0` while 65519.x stays 65504;
- **−0.0 preserved** both ways (`#x8000` ↔ `-0.0`);
- values |x| < 2^−25 round to **signed** zero; 2^−25 exactly ties to even → 0.

Class wiring is a mechanical clone of the existing `f32`/`u16` entries: checker = existing `%flonum-checker` (identical to the reference's `flonum?` acceptance — the setter *rounds*, never rejects), maker = `(make-u16vector n (%f16-encode fill))` (encode once), copier = `u16vector-copy!`, length = `u16vector-length`, default `0.`, data? = `u16vector?`, data->body = `(%checked-data->body u16vector? ...)`.

Tests:

1. **Exhaustive 65536-pattern round-trip** in `tests/scheme/srfi/srfi231-storage-classes.scm` — `(%f16-encode (%f16-decode p)) = p` for every pattern except NaN payloads (the reference documents this exact strategy in its own comment block, lines 1607–1656). Directed edges: `1.0↔#x3C00`, `-0.0↔#x8000`, ±inf, NaN, 65504 max finite, 65519.9→65504 vs 65520.0→`+inf.0`, 2^−25→0.0 tie, subnormal tie 1.5·2^−24 → mantissa 2 (ties-to-even), a mantissa-carry case (2047.5/1024-style).
2. Flip `srfi231-storage-classes.scm:118` from `(test-equal #f f16-storage-class)` to real class assertions (`check-storage-class` pattern at lines 60–69), and update the lines 75–81 comment.
3. **Prune divergence ids 98/148/149/150 from `srfi231-official-transform.py` and regenerate `srfi231-official.scm`** — the generated file must never be hand-edited, and the suite's `DIVERGENCE-RESOLVED` epilogue exits 1 on stale entries, so the prune is enforced, not optional. The pristine fixture's f16-specific `array-copy` expectations (`test-arrays.scm.gambit` lines 1871, 1960–2007 — branches on `(eq? destination-storage-class f16-storage-class)` with half-float rounding expectations) become live and are the external correctness oracle for the codec.
4. Check `arrays.sld` against the reference's storage-class widening lists (`generic-arrays.scm` 2659, 2715–2719: f16 → f32 → f64 in `array-append` etc.) — f16 becoming non-`#f` may need to join a widening table.

Doc/comment updates in the same PR: `storage-classes.sld` header (lines 14/19/26–29 — the 14-real/3-deferred arithmetic becomes 16/1), `docs/dev/testing.md:279` (drop "f16 deferral" from the divergence-reason examples), `docs/dev/srfi-implementation-notes.md:880` (see companion #2380 — that line is already stale today), repoint the in-tree "(#2353 tracks the call)" comments at this issue, and a CHANGELOG `### Added` entry under `[Unreleased]`.

Out of scope, considered and rejected:

- **Chibi's portable fallback** aliases `f16-storage-class` to `f32-storage-class` (spec-permitted) — rejected: 2× storage and no f16 rounding semantics, and Kaappi has the substrate to do it properly.
- **A Zig-side codec primitive** (`@floatCast` to `f16` gives correct RNE for free) — rejected as a correctness path: it would break the property that `(srfi 231)` is pure R7RS-small, which is the implementation's announced selling point. Worth revisiting only as a later performance optimization behind the same interface.
- **f8**: stays `#f` — sanctioned fallback even in the reference; no standard 8-bit float format to conform to.

Estimated size: ~80–100 lines added to `storage-classes.sld` (decode ~20, encode ~45–55 incl. `%ilogb`, class ~15).

### References

- Reference implementation: `generic-arrays.scm` lines 1449–1601 (codec + class), 1607–1656 (validation strategy) — https://github.com/scheme-requests-for-implementation/srfi-231
- #2353 (closed) — u1 shipped there; its discussion quantifies the f16 gap and rebuts the "sanctioned fallback" reading; commit 5bf94e6e / PR #2363
- srfi-231 mailing-list thread, Aug 2026: https://srfi-email.schemers.org/srfi-231/msg/43018594/
- Chibi's `lib/srfi/231.sld` cond-expand (native `f16vector` via `(srfi 160 mini)` on Chibi; f32 alias elsewhere)

### LINKED ISSUE #2384: SRFI 231 official suite: known-divergence matching by test id alone can absorb a regression in other rows of a multi-fire id (test 150 masks f16)
### Summary

Surfaced during review of #2383 (also flagged by CodeRabbit there): the official suite's known-divergence table (`srfi231-official-transform.py`, embedded into the generated `srfi231-official.scm`) matches divergences by **numeric test id alone**, but some ids fire many times. Test id 150 runs once per storage class — 16 evaluations over the `test-values` table — while its divergence entry documents only the two c64/c128 rows (#2382). Any *other* row failing under id 150 is therefore printed as `DIVERGENT-EXPECTED 150 (c64/c128 …)` and does not fail the suite.

The row that matters most is the one #2383 just made live: `f16-storage-class` with a `(make-u16vector 10)` good-data fixture. If `make-specialized-array-from-data` or `storage-class-data?` regressed for f16, the suite would stay green.

### Why the verdict doesn't catch it

The epilogue gates the exit status on exactly two things:

```scheme
(exit (if (and (= failed-tests 0) (= resolved-divergences 0)) 0 1))
```

- `failed-tests` never increments for a known id — `report-failure` routes it to `divergent-tests` first.
- `resolved-divergences` only fires when an id **stops diverging entirely** (`diverged-ids` never set). Since the c64/c128 rows always fail, id 150 is always marked diverged, so that guard can't see an *additional* row joining the failure set.
- `divergent-tests` (currently 5) is printed in the summary line but never asserted — 5 silently becoming 6 changes nothing.

### Why this is new pressure, not a new hole

The id-granularity limitation predates #2383 — the old entry lumped the f16+c64+c128 rows the same way. But before that PR every row under id 150 was *expected* to fail, so there was nothing to hide. Now some rows under an expected-fail id are expected to **pass**, which the accounting was never built for.

### Proposed fix

Stay inside the existing architecture (no renumbering of upstream ids; the pristine fixture is never hand-edited):

1. Carry an expected multiplicity per entry — e.g. `(150 2 "c64/c128 …")` — in the transform script's table.
2. Count observed divergences per id in `report-failure` (a second vector next to `diverged-ids`).
3. In the epilogue, treat **observed > expected** as a failure with its own message (`DIVERGENCE-OVERCOUNT`), symmetric with the existing `DIVERGENCE-RESOLVED` staleness check. Tightening `DIVERGENCE-RESOLVED` to fire on observed < expected (rather than only observed = 0) falls out for free.

A cheaper alternative — asserting the total `(= divergent-tests 5)` in the final verdict — also closes the hole, but produces a less pointed diagnostic and needs retouching on every table change anyway, so the per-entry count seems worth the few extra lines.

Whatever shape lands, remember both copies: the table lives in `srfi231-official-transform.py` **and** the generated `srfi231-official.scm` must be regenerated, never hand-edited (`docs/dev/testing.md`).

### References

- #2383 — the PR whose review surfaced this (see its review thread on `srfi231-official-transform.py`)
- #2382 — the c64/c128 representation divergence id 150 now documents
- #2379 — the f16 port that made rows under id 150 expected-to-pass
- `tests/scheme/srfi/srfi231-official.scm` lines ~112–142 (accounting), ~1148–1162 (the 16-row id-150 loop), epilogue at the end of the file

baijum:
Closes #2379. Closes #2384 (the divergence-accounting hardening landed here, in d546cb1, rather than as a follow-up).

## What

`f16-storage-class` was bound to `#f` — a scope reduction carried over from #2353 — even though the `u16vector` substrate it needs is shipped and the reference implementation's codec (generic-arrays.scm 1449–1601) is **pure arithmetic** over it: no host bit-reinterpretation anywhere, so it ports to R7RS-small directly. The spec's `#f` escape clause is written for implementations *lacking* the substrate, which this is not.

## The codec

A faithful transliteration of the reference's `f16->double` / `double->f16`, hand-expanded from its defining macro for the single instantiation the class needs (mantissa-width 10, exponent-width 5, bias 15): `%f16-decode` / `%f16-encode` as library-internal defines plus a fresh `%ilogb` loop (halving/doubling, exact at every step). Gambit-ism substitutions, each semantics-preserving:

| Gambit | pure R7RS |
|---|---|
| `##flonum->fixnum (flround x)` | `(exact (round x))` — R7RS round **is** ties-to-even |
| `flscalbn x n` | `(* x (expt 2.0 n))` — every scale factor is a power of two within f64's exact range, so the scale-exactly-then-round-**once** shape survives |
| `##flcopysign` sign test | `(or (< x 0.0) (eqv? x -0.0))` |
| `flfinite?` / `flnan?` | `finite?` / `nan?` from `(scheme inexact)` |

The structure that makes the reference correct is preserved: the subnormal branch scales by 2^24 directly (one rounding at the subnormal ulp), mantissa carry after rounding bumps the exponent exactly — sending the 65520.0 tie **up** to `+inf.0` while 65519.9 stays 65504 — `-0.0` round-trips both ways, and 2^-25 ties to even → 0. NaN payloads canonicalize to `#x7FFF` (sign/payload not portable; decode yields `+nan.0` either way). Class wiring mirrors the f32/u16 entries (`%flonum-checker`, fill encoded once by the maker, `u16vector-copy!`, `%checked-data->body`).

## Tests

- **Exhaustive 65536-pattern round-trip** in `srfi231-storage-classes.scm`, driven through the public getter/setter (the reference's own validation strategy, tightened to a full sweep): every non-NaN pattern re-encodes exactly; 2046 NaN payloads identified.
- Directed edges: `1.0↔#x3C00`, `-0.0↔#x8000`, ±inf, NaN, max finite 65504, the 65520.0 overflow tie, subnormal/carry ties, negatives.
- Old `(test-equal #f f16-storage-class)` replaced with `check-storage-class` coverage (202 passes).

## Official suite

Divergence ids **98/148/149 pruned** and the suite regenerated (never hand-edited). One discovery along the way: **id 150 was overloaded** — its entry was hiding two *pre-existing* c64/c128 row failures (the reference backs complex classes with even-length f32/f64vector pairs; Kaappi uses native `c64vector`/`c128vector`) alongside its f16 row. It stays in the table, re-scoped to that accurate reason — tracked in #2382, filed for it.

Result: `10922 passed, 231 error-message-only, 5 known divergences, 0 unexpected failures, 0 resolved divergences` — evaluation count up from 8879, because f16 paths now complete: the random-builder blocks give `array-copy`/`array-append` real half-float rounding as the external correctness oracle. `arrays.sld` needs no widening table — this port's `array-copy` goes through the generic checker/setter path.

Review hardening (d546cb1, #2384): the known-divergence table now records the exact expected divergence count per id and the verdict epilogue fails on any mismatch in either direction, so an undocumented failure under a shared test id can no longer hide behind a documented divergence; `%ilogb` additionally clamps to the [-15, 16] classification range.

## Docs

`storage-classes.sld` header and `srfi-implementation-notes.md` now say 16 real / 1 deferred (`f8`); `testing.md`'s divergence-reason examples updated; CHANGELOG `### Added` entry.

## Validation

- `tests/scheme/srfi/srfi231-storage-classes.scm`: 202 passes
- Official suite: green as above
- `tests/scheme/run-all.sh`: 720 Scheme files + R7RS 1395/1395 → **2115 pass, 0 fail**

### DIFF
--- a/docs/dev/testing.md
+++ b/docs/dev/testing.md
@@ -276,11 +276,15 @@ Conventions specific to this suite:
 
 - **Known divergences are accounted, not failed.** A table of test ids in
   the generated file maps each documented kaappi-vs-reference divergence to
-  its reason and issue reference (the c64/c128 native-vector representation
-  choice, the unsafe-view UB choice, the Gambit string-mutability
-  expectation). The suite exits nonzero only on *unexpected* failures — or
-  when a known divergence stops diverging, which means its entry is stale
-  and hiding real coverage: prune it.
+  its reason, issue reference, and the *exact* number of evaluations
+  expected to diverge under it (shared test forms run once per storage
+  class, so one id can account several rows) — the c64/c128
+  native-vector representation choice, the unsafe-view UB choice, the
+  Gambit string-mutability expectation. The suite exits nonzero only on
+  *unexpected* failures — or when an entry's divergence count doesn't match
+  exactly: zero observed means the entry is stale and hiding real coverage
+  (prune it), and more than recorded means an undocumented failure is
+  absorbing into the entry.
 - **Error-expecting tests pass on any error** — only the Gambit message
   text differs from kaappi's (counted separately as "error-message-only").
 - **It runs ~150 s** (the isolated `KAAPPI_HOME` compiles the SRFI's
--- a/docs/dev/testing.md
+++ b/docs/dev/testing.md
@@ -276,10 +276,11 @@ Conventions specific to this suite:
 
 - **Known divergences are accounted, not failed.** A table of test ids in
   the generated file maps each documented kaappi-vs-reference divergence to
-  its reason and issue reference (f16 deferral, the unsafe-view UB choice,
-  the Gambit string-mutability expectation). The suite exits nonzero only
-  on *unexpected* failures — or when a known divergence stops diverging,
-  which means its entry is stale and hiding real coverage: prune it.
+  its reason and issue reference (the c64/c128 native-vector representation
+  choice, the unsafe-view UB choice, the Gambit string-mutability
+  expectation). The suite exits nonzero only on *unexpected* failures — or
+  when a known divergence stops diverging, which means its entry is stale
+  and hiding real coverage: prune it.
 - **Error-expecting tests pass on any error** — only the Gambit message
   text differs from kaappi's (counted separately as "error-message-only").
 - **It runs ~150 s** (the isolated `KAAPPI_HOME` compiles the SRFI's
--- a/lib/srfi/231/storage-classes.sld
+++ b/lib/srfi/231/storage-classes.sld
@@ -23,10 +23,11 @@
 ;;; spec gives their exact reference definitions verbatim, reused here
 ;;; unchanged. u1 is a direct port of the reference's own bit-packing over
 ;;; u16vector (the representation the spec itself documents for bit
-;;; arrays). f8 is left #f even in the SRFI's OWN reference implementation
-;;; (no standard 8-bit-float Scheme type exists anywhere), and f16 -- which
-;;; the reference implements as software half-floats -- remains #f as a
-;;; documented scope reduction (#2353).
+;;; arrays), and f16 a faithful port of its software half-floats over
+;;; u16vector (#2379) -- the spec's #f escape clause is for implementations
+;;; lacking the substrate, which neither port is. f8 alone is left #f,
+;;; even in the SRFI's OWN reference implementation (no standard 8-bit
+;;; float Scheme type exists anywhere).
 (define-library (srfi 231 storage-classes)
   (import (scheme base)
           (srfi 160 s8) (srfi 160 s16) (srfi 160 s32) (srfi 160 s64)
@@ -34,7 +35,9 @@
           (srfi 160 f32) (srfi 160 f64) (srfi 160 c64) (srfi 160 c128)
           ;; bitwise-and/ior/not for u1's bit-packing (R7RS-small has no
           ;; bitwise primitives of its own)
-          (srfi 60))
+          (srfi 60)
+          ;; finite?/nan? for the f16 codec's special-value classification
+          (scheme inexact))
   (export make-storage-class storage-class?
           storage-class-getter storage-class-setter storage-class-checker
           storage-class-maker storage-class-copier storage-class-length
@@ -199,8 +202,142 @@
              (error "Expecting a u16vector passed to (storage-class-data->body u1-storage-class): " data)
              (vector (* 16 (u16vector-length data)) data)))))
 
+    ;; f8 -- #f even in the SRFI's OWN reference implementation: no
+    ;; standard 8-bit float format exists to conform to.
     (define f8-storage-class #f)
-    ;; f16 stays #f for now: the reference implements software half-floats
-    ;; over u16vectors, which is a deliberate port of its own to make, not
-    ;; a mechanical substrate mapping like u1's (#2353 tracks the call).
-    (define f16-storage-class #f)))
+
+    ;; f16 -- software half-floats over a u16vector: a faithful
+    ;; transliteration of the reference implementation's own codec
+    ;; (generic-arrays.scm's f16->double / double->f16), hand-expanded
+    ;; from its defining macro for the single instantiation (mantissa-width
+    ;; 10, exponent-width 5, bias 15) the class needs (#2379). The codec
+    ;; is pure arithmetic -- no host bit-reinterpretation anywhere -- which
+    ;; is exactly why it ports to R7RS-small. Gambit-ism substitutions,
+    ;; each semantics-preserving:
+    ;;   ##flonum->fixnum (flround x) -> (exact (round x))
+    ;;     (R7RS round IS ties-to-even, the required rounding mode)
+    ;;   flscalbn x n -> (* x (expt 2.0 n))
+    ;;     (every scale factor here is a power of two within f64's exact
+    ;;     range, so the scale-exactly-then-round-ONCE shape -- no double
+    ;;     rounding -- is preserved)
+    ;;   ##flcopysign sign test -> (or (< x 0.0) (eqv? x -0.0))
+    ;;     (comparisons don't distinguish -0.0; eqv? does, per R7RS)
+    ;;   flfinite?/flnan? -> finite?/nan? from (scheme inexact)
+    ;; The two structural properties that make this codec correct, both
+    ;; from the reference and both preserved:
+    ;;   - the subnormal branch scales by 2^24 directly: ONE rounding at
+    ;;     the subnormal ulp (a normalize-then-shift formula would round
+    ;;     twice, at the wrong ulp);
+    ;;   - mantissa carry after rounding: a significand that rounds up to
+    ;;     exactly 2048 bumps the exponent -- exact, no re-round; at the
+    ;;     top binade it overflows to infinity. This is what rounds the
+    ;;     tie 65520.0 UP to +inf.0 while 65519.9 stays 65504.0.
+    ;; -0.0 round-trips both ways (#x8000 <-> -0.0); |x| below 2^-25
+    ;; rounds to signed zero, with the tie 2^-25 exactly going to 0 (even)
+    ;; and the subnormal tie 1.5*2^-24 to mantissa 2. NaN payloads are
+    ;; not preserved -- decode yields +nan.0 and encode canonicalizes
+    ;; every NaN to #x7FFF -- since a NaN's sign and payload are not
+    ;; portable across implementations (decode yields +nan.0 either way).
+    (define (%f16-decode x)   ;; exact integer in [0,65536) -> real
+      (let ((e (bitwise-and 31 (arithmetic-shift x -10)))
+            (m (bitwise-and 1023 x))
+            (s (arithmetic-shift x -15)))
+        (cond ((= e 31)
+               (if (= m 0)
+                   (if (= s 0) +inf.0 -inf.0)
+                   +nan.0))
+              ((> e 0)
+               (let ((n (if (= s 0) (+ 1024 m) (- (+ 1024 m)))))
+                 (* n (expt 2.0 (- e 25)))))
+              ((= m 0)
+               (if (= s 0) +0.0 -0.0))
+              (else
+               (let ((n (if (= s 0) m (- m))))
+                 (* n (expt 2.0 -24)))))))
+
+    ;; floor(log2|x|) for a finite, nonzero real x -- the reference's
+    ;; flilogb, written fresh as a halving/doubling loop (exact at every
+    ;; step: multiplying by 2.0 or 0.5 never rounds). Only the
+    ;; classification against [-14,15] and the exact exponent feeding the
+    ;; scale factor matter to the codec.
+    (define (%ilogb x)
+      (let ((ax (abs x)))
+        (if (>= ax 1.0)
+            (let loop ((e 0) (v ax))
+              (if (< v 2.0) e (loop (+ e 1) (* v 0.5))))
+            (let loop ((e -1) (v (* ax 2.0)))
+              (if (>= v 1.0) e (loop (- e 1) (* v 2.0)))))))
+
+    (define (%f16-encode x)   ;; real -> exact integer in [0,65536)
+      (define (%construct sign-bit biased-exponent mantissa)
+        (bitwise-ior (arithmetic-shift sign-bit 15)
+                     (bitwise-ior (arithmetic-shift biased-exponent 10)
+                                  mantissa)))
+      (let ((sign-bit (if (or (< x 0.0) (eqv? x -0.0)) 1 0)))
+        (cond ((not (finite? x))
+               (if (nan? x)
+                   ;; canonical NaN: the sign bit of a NaN is unobservable
+                   (%construct 0 31 1023)
+                   ;; an infinity
+                   (%construct sign-bit 31 0)))
+              ((zero? x)
+               ;; a zero (sign preserved)
+               (%construct sign-bit 0 0))
+              (else
+               (let ((exponent (%ilogb x)))
+                 (cond ((<= 16 exponent)
+                        ;; infinity: the exponent is too large
+                        (%construct sign-bit 31 0))
+                       ((< -15 exponent)
+                        ;; probably normal, finite in representation,
+                        ;; unless the rounded mantissa carries
+                        (let ((possible-mantissa
+                               (exact (round (* (abs x)
+                                                (expt 2.0 (- 10 exponent)))))))
+                          (if (< possible-mantissa 2048)
+                              ;; no overflow
+                              (%construct sign-bit
+                                          (+ exponent 15)
+                                          (bitwise-and possible-mantissa 1023))
+                              (if (= exponent 15)
+                                  ;; maximum finite exponent: overflow to
+                                  ;; infinity
+                                  (%construct sign-bit 31 0)
+                                  ;; increase exponent by 1, mantissa is
+                                  ;; zero, no double rounding
+                                  (%construct sign-bit (+ exponent 16) 0)))))
+
+                       (else
+                        ;; usually subnormal
+                        (let ((possible-mantissa
+                               (exact (round (* (abs x) (expt 2.0 24))))))
+                          (if (< possible-mantissa 1024)
+                              ;; doesn't overflow to normal
+                              (%construct sign-bit 0 possible-mantissa)
+                              ;; overflow to smallest normal
+                              (%construct sign-bit 1 0))))))))))
+
+    ;; Class wiring mirrors the reference's f16 entry exactly: the same
+    ;; checker acceptance as f32/f64 (flonum? -- the setter ROUNDS, never
+    ;; rejects), the fill encoded once by the maker, and the identity
+    ;; data->body contract of every u16vector-backed class.
+    (define f16-storage-class
+      (make-storage-class
+       ;; getter
+       (lambda (body i) (%f16-decode (u16vector-ref body i)))
+       ;; setter
+       (lambda (body i obj) (u16vector-set! body i (%f16-encode obj)))
+       ;; checker
+       %flonum-checker
+       ;; maker
+       (lambda (n val) (make-u16vector n (%f16-encode val)))
+       ;; copier
+       u16vector-copy!
+       ;; length
+       u16vector-length
+       ;; default
+       0.0
+       ;; data?
+       u16vector?
+       ;; data->body
+       (%checked-data->body u16vector? "f16-storage-class" "u16vector")))))
--- a/lib/srfi/231/storage-classes.sld
+++ b/lib/srfi/231/storage-classes.sld
@@ -255,18 +255,24 @@
                (let ((n (if (= s 0) m (- m))))
                  (* n (expt 2.0 -24)))))))
 
-    ;; floor(log2|x|) for a finite, nonzero real x -- the reference's
-    ;; flilogb, written fresh as a halving/doubling loop (exact at every
-    ;; step: multiplying by 2.0 or 0.5 never rounds). Only the
-    ;; classification against [-14,15] and the exact exponent feeding the
-    ;; scale factor matter to the codec.
+    ;; floor(log2|x|), clamped to [-15, 16], for a finite, nonzero real
+    ;; x -- the reference's flilogb, written fresh as a halving/doubling
+    ;; loop (exact at every step: multiplying by 2.0 or 0.5 never
+    ;; rounds). The clamp loses nothing: %f16-encode only ever consumes
+    ;; the classification ("<= -15", "in [-14,15]", ">= 16") plus the
+    ;; exact exponent inside the normal range, and it bounds the loop to
+    ;; ~31 iterations even for astronomically large or tiny doubles.
     (define (%ilogb x)
       (let ((ax (abs x)))
         (if (>= ax 1.0)
             (let loop ((e 0) (v ax))
-              (if (< v 2.0) e (loop (+ e 1) (* v 0.5))))
+              (cond ((< v 2.0) e)
+                    ((>= e 16) 16)
+                    (else (loop (+ e 1) (* v 0.5)))))
             (let loop ((e -1) (v (* ax 2.0)))
-              (if (>= v 1.0) e (loop (- e 1) (* v 2.0)))))))
+              (cond ((>= v 1.0) e)
+                    ((<= e -15) -15)
+                    (else (loop (- e 1) (* v 2.0))))))))
 
     (define (%f16-encode x)   ;; real -> exact integer in [0,65536)
       (define (%construct sign-bit biased-exponent mantissa)
--- a/tests/scheme/srfi/srfi231-official-transform.py
+++ b/tests/scheme/srfi/srfi231-official-transform.py
@@ -729,11 +729,8 @@ NEW_REPORT = ''';;; --- kaappi vendoring: known-divergence accounting ----------
 (define divergent-tests 0)
 (define diverged-ids (make-vector 10000 #f))
 (define known-divergences
-  (list '(98 . "f16-storage-class is #f (documented deferral, kaappi#2353)")
-        '(148 . "f16-storage-class is #f (documented deferral, kaappi#2353)")
-        '(149 . "f16-storage-class is #f (documented deferral, kaappi#2353)")
-        '(150 . "f16-storage-class is #f (documented deferral, kaappi#2353)")
-        '(147 . "R7RS strings are mutable; the suite encodes Gambit's immutable-string expectation")
+  (list '(147 . "R7RS strings are mutable; the suite encodes Gambit's immutable-string expectation")
+        '(150 . "c64/c128 are backed by native (srfi 160) c64vector/c128vector, not the reference's even-length f32/f64vector pairs, so those fixtures are rejected (kaappi#2382)")
         '(351 . "unsafe specialized views are unchecked per the spec text; the reference happens to check (see kaappi#2362)")))
 (define (known-divergence id) (assq id known-divergences))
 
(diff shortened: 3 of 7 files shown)


tests/scheme/srfi/srfi231-official-transform.py
```diff
@@ -729,11 +729,8 @@ def strip_time(src):
 (define divergent-tests 0)
 (define diverged-ids (make-vector 10000 #f))
 (define known-divergences
-  (list '(98 . "f16-storage-class is #f (documented deferral, kaappi#2353)")
-        '(148 . "f16-storage-class is #f (documented deferral, kaappi#2353)")
-        '(149 . "f16-storage-class is #f (documented deferral, kaappi#2353)")
-        '(150 . "f16-storage-class is #f (documented deferral, kaappi#2353)")
-        '(147 . "R7RS strings are mutable; the suite encodes Gambit's immutable-string expectation")
+  (list '(147 . "R7RS strings are mutable; the suite encodes Gambit's immutable-string expectation")
+        '(150 . "c64/c128 are backed by native (srfi 160) c64vector/c128vector, not the reference's even-length f32/f64vector pairs, so those fixtures are rejected (kaappi#2382)")
```
baijum: Test id 150 fires 16 times — once per storage-class row in the `test-values` table — and `known-divergence` matches by numeric id alone, so this entry absorbs a failure in *any* of those 16 rows, not just the two c64/c128 rows it documents. The f16 row (`f16-storage-class` with a `(make-u16vector 10)` good-data fixture) is exactly the coverage this PR just made live: if `make-specialized-array-from-data` or `storage-class-data?` regressed for f16, the suite would print `DIVERGENT-EXPECTED 150 (c64/c128 …)` and still exit 0. The verdict epilogue only gates on `failed-tests` and `resolved-divergences`; `divergent-tests` is printed but never asserted, so the count silently going 5 → 6 wouldn't fail anything. (CodeRabbit flagged the same shape; adding the mechanics here.)

The granularity limitation is pre-existing — the old entry lumped f16+c64+c128 the same way — but before this PR every row under the id was expected to fail, so nothing could hide. Now some rows are expected to *pass*, which is a new situation for the accounting.

A minimal fix that stays inside the existing architecture (no renumbering of upstream ids): carry an expected multiplicity per entry — `(150 2 "c64/c128 …")` — count per-id divergences in `report-failure`, and have the epilogue treat observed > expected as a failure (observed < expected is already covered by the resolved-divergence check only when the count hits zero, so the multiplicity check would tighten that side too). Even just asserting the total (`(= divergent-tests 5)`) in the final `exit` form would close the hole, at the cost of a less pointed message.

Fine as a follow-up rather than in this PR, but worth an issue reference so it doesn't evaporate — it's the accounting guard for precisely the coverage this PR fought to make real.

---

lib/srfi/231/storage-classes.sld
```diff
@@ -199,8 +202,142 @@
...
+    ;; step: multiplying by 2.0 or 0.5 never rounds). Only the
+    ;; classification against [-14,15] and the exact exponent feeding the
+    ;; scale factor matter to the codec.
+    (define (%ilogb x)
+      (let ((ax (abs x)))
+        (if (>= ax 1.0)
+            (let loop ((e 0) (v ax))
+              (if (< v 2.0) e (loop (+ e 1) (* v 0.5))))
+            (let loop ((e -1) (v (* ax 2.0)))
+              (if (>= v 1.0) e (loop (- e 1) (* v 2.0)))))))
```
baijum: Nit, optional: the halving loop runs once per binade, so `(%f16-encode 1e300)` does ~1000 exact multiply-compare iterations before `%f16-encode`'s `(<= 16 exponent)` discards everything beyond "≥ 16". Since the codec only ever consumes the exact exponent within [-14, 15] plus the two overflow/underflow classifications, the `ax >= 1.0` arm could bail out as soon as `e` reaches 16 (returning 16 directly) with no observable difference. Only matters for setters fed astronomically large doubles, and the loop is correct and terminating as written — take it or leave it.

---

baijum (commented):
Reviewed the codec, the class wiring, the test strategy, and the divergence-table changes. Two inline comments (one accounting gap worth a follow-up issue, one optional nit); the implementation itself I could not fault.

**Independent verification of the codec.** Since I couldn't build kaappi in this environment, I transliterated `%f16-decode` / `%ilogb` / `%f16-encode` line-for-line into Python (whose `round` is also ties-to-even on a double, matching R7RS `round`) and checked it against hardware IEEE 754 binary16 (`struct` format `e`):

- **decode**: agrees with hardware on all 65 536 bit patterns, including NaN classification, ±0.0 signs, and subnormals;
- **round-trip**: `encode(decode(p)) = p` for all patterns except the 2 046 NaN payloads, which canonicalize to `#x7FFF` — matching the test suite's expected count;
- **encode**: agrees with hardware round-to-nearest-even on ~440 000 doubles — every midpoint between adjacent representable f16 values plus its two neighboring doubles (the complete tie sweep), the 65520.0 overflow tie and neighbors, `2^-25` and neighbors, extremes (`±1e300`, `±5e-324`), and 250 000 random values across the normal and subnormal ranges. Zero mismatches.

So the transliteration's exactness claims hold as stated: the power-of-two scalings never round, `%ilogb` is exact, and each branch rounds exactly once at the correct ulp.

**Other checks:**

- Class wiring is field-for-field consistent with the neighboring f32/u16 entries (checker shared, fill encoded once in the maker, identity `data->body` with the type guard).
- Confirmed the issue #2379 item 4 concern is moot as claimed: no file in `lib/srfi/231/` dispatches on storage-class identity (no widening table, no `eq?` fast paths), so f16 arrays ride the generic getter/setter path everywhere.
- Doc arithmetic is consistent across all touched files (17 singletons / 16 real / f8 only `#f`; "15 real classes below that have a copier" = 16 − u1).
- `srfi231-official.scm` and the transform script's embedded divergence tables match exactly, and the re-scoped id-150 reason matches what #2382 documents.
- Test-file epilogue follows the repo convention (runner grabbed before `test-end`, `exit 1` on failures).

---

tests/scheme/srfi/srfi231-official-transform.py
baijum: Implemented in d546cb16, per your suggested shape (stays inside the existing architecture, no renumbering of upstream ids):

- entries are now `(id expected-count . reason)` — `(147 1 …)`, `(150 2 …)`, `(351 2 …)`;
- `report-failure` counts per-id observations into `diverged-counts` (replacing the boolean vector);
- the epilogue fails the suite on any mismatch in either direction: zero observed is the existing `DIVERGENCE-RESOLVED` stale-entry signal; **more** than expected is a new `DIVERGENCE-COUNT-MISMATCH` ("an undocumented failure is hiding under this id"); **fewer** is the same mismatch with an over-accounting message ("re-count it") — so the partial-resolution case you noted (2 → 1) is also caught, not just the zero case.

Negative test on a scratch copy (not committed): corrupting the f16 row's good-data fixture to an `s16vector` — the exact hiding scenario — now prints `DIVERGENCE-COUNT-MISMATCH 150: expected 2 diverging evaluations, observed 3` and exits 1; before the fix that exited 0 silently. The suite is deterministic (fixed SRFI 27 seeds), so the counts are stable run to run — verified across the baseline, post-f16, and post-fix runs.

Since the fix landed in this PR rather than as a follow-up, I didn't file the separate issue — this comment thread (and the commit message) carries the rationale. `testing.md` and the generated-file header now describe the exact-count contract too.

lib/srfi/231/storage-classes.sld
baijum: Taken in d546cb16 — with the symmetric clamp on the doubling arm as well: the loop now bails at `e >= 16` upward and `e <= -15` downward, bounding it to ~31 iterations for any input. The clamp is safe for the reason you gave (encode only consumes the classification "≤ −15 / in [−14,15] / ≥ 16" plus the exact exponent inside the normal range), and the doc comment now states that contract explicitly instead of claiming an unclamped `floor(log2|x|)`.

No observable change: the exhaustive 65 536-pattern sweep and all directed edges are identical, and `encode` of `1e300`/`-1e300`/`5e-324`/`-5e-324` lands on `±inf`/`±0` as before.

tests/scheme/srfi/srfi231-official.scm
```diff
@@ -117,11 +117,8 @@ OTHER DEALINGS IN THE SOFTWARE.
 (define divergent-tests 0)
 (define diverged-ids (make-vector 10000 #f))
 (define known-divergences
-  (list '(98 . "f16-storage-class is #f (documented deferral, kaappi#2353)")
-        '(148 . "f16-storage-class is #f (documented deferral, kaappi#2353)")
-        '(149 . "f16-storage-class is #f (documented deferral, kaappi#2353)")
-        '(150 . "f16-storage-class is #f (documented deferral, kaappi#2353)")
-        '(147 . "R7RS strings are mutable; the suite encodes Gambit's immutable-string expectation")
+  (list '(147 . "R7RS strings are mutable; the suite encodes Gambit's immutable-string expectation")
+        '(150 . "c64/c128 are backed by native (srfi 160) c64vector/c128vector, not the reference's even-length f32/f64vector pairs, so those fixtures are rejected (kaappi#2382)")
```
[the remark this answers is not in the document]
baijum: Addressed in d546cb16 via the expected-multiplicity route (no distinct ids, so upstream ids stay stable): each known-divergence entry is now `(id expected-count . reason)`, `report-failure` counts observations per id, and the verdict epilogue exits nonzero on any count mismatch — more than expected reports `DIVERGENCE-COUNT-MISMATCH … an undocumented failure is hiding under this id`, which makes an f16 (or any other row's) regression under test 150 a hard failure again. Verified by a scratch-copy negative test: breaking only the f16 row flips the suite from exit 0 (before) to exit 1 with the mismatch message.

baijum:
## Reviewer verification of d546cb1 + merge-readiness assessment

Follow-up to my review of `b855bde`: I re-verified the head commit end to end. Summary of where the PR stands and my verdicts on the three CodeRabbit threads still open on the current head.

### d546cb1 verification

**Divergence accounting.** The implementation matches the shape my review thread proposed, and I checked the pieces independently:

- The per-id expected counts (147 → 1, 150 → 2, 351 → 2) sum to the 5 known divergences every prior run reported, so the table is self-consistent with the suite's own history. Id 351's count of 2 from a single textual `test-err` form is plausible for the same reason id 150's is — the form sits inside a nested loop (the `do` over `left-dim` × `for-each` over three array kinds), so one textual site legitimately accounts multiple evaluations.
- The transform script's embedded `NEW_REPORT`/`NEW_SUM` blocks and the generated `srfi231-official.scm` match exactly — the generated file was regenerated, not hand-edited.
- The epilogue now fails on both directions of a mismatch (observed > expected: `DIVERGENCE-COUNT-MISMATCH`, an undocumented failure hiding under the id; 0 < observed < expected: the over-accounting message; observed = 0: the pre-existing `DIVERGENCE-RESOLVED` staleness signal), and the negative test described in the commit message exercises exactly the hiding scenario the review flagged.

**`%ilogb` clamp.** I traced the boundary cases: a true exponent of exactly 16 or −15 still returns exactly, because each loop's value check precedes its clamp check — the clamp only fires when the true exponent is already outside what `%f16-encode` distinguishes ("≤ −15", "in [−14, 15]", "≥ 16"). I also re-ran my independent Python transliteration of the codec (the oracle from my original review: hardware IEEE 754 binary16 via `struct`) with the clamped `%ilogb` substituted in: all 65 536 decode patterns, the full round-trip (2 046 NaN payloads canonicalizing to `#x7FFF`), and the ~440 000-double encode sweep — every representable-value midpoint tie ± one ulp, the 65520.0 overflow tie, `2^-25` and neighbors, `±1e300`, `±5e-324`, plus 250 000 random values — are all still zero-mismatch. The "no observable change" claim in the commit message holds.

Both of my review threads are addressed; I've resolved them. Issue #2384 (the follow-up I had filed for the accounting gap) is fixed by this commit — the PR description now carries `Closes #2384` so it closes on merge.

### Verdicts on the three open CodeRabbit threads (head `d546cb1`)

1. **`testing.md` omits the fewer-but-nonzero mismatch case** — **valid**, trivial: the epilogue reports "the entry over-accounts (re-count it)" for `0 < observed < expected`, and the doc currently describes only the zero and greater-than cases. Two-line doc addition.

2. **The `(when (vector-ref executed-tests id) …)` guard skips entries whose id never executes** — **valid hardening**, pre-existing shape: if a future regeneration drops a form entirely, its table entry passes silently today. Treating an unexecuted id as a failure (ideally with its own "never executed" wording rather than reusing the misleading "stops diverging" message) closes the last accounting gap. Fits naturally in the same commit as (1), with the `.scm` regenerated.

3. **Signed-zero `eqv?` reliance is non-portable R7RS** — **incorrect, no change needed**, for three independent reasons:
   - R7RS-small §6.1 requires `eqv?` to return `#f` for two inexact numbers that are operationally distinguishable by standard arithmetic. `+0.0` and `-0.0` are exactly that wherever both exist: `(/ 1.0 0.0)` ⇒ `+inf.0` vs `(/ 1.0 -0.0)` ⇒ `-inf.0`. So in any R7RS implementation *with* distinct signed zeros, `(eqv? 0.0 -0.0)` ⇒ `#f` is required, not merely permitted — the code's comment ("eqv? does, per R7RS") is accurate for every implementation where the distinction is observable at all.
   - In a hypothetical implementation with a single zero, the `-0.0` literal denotes that same zero on *both* sides of the codec — encode's `(eqv? x -0.0)` test and decode's `-0.0` return value collapse together — so the round-trip stays self-consistent there too; only the stored bit pattern differs, which that implementation cannot observe.
   - The codec is a transliteration of the reference's `##flcopysign` sign test, which has identical semantics; and the concrete host is Kaappi, whose NaN-boxed flonums have fully distinguishable signed zeros (`-0.0 ↔ #x8000` is a tested directed edge in this very PR).

### Merge readiness

- **CI**: 19/21 checks green on `d546cb1` at last look — including format, gc-stress, s390x, all four BSDs, riscv64/ppc64le, wasm, windows-x64; `windows-arm-test` and `test (macos-latest, ReleaseSafe)` still running, no failures anywhere.
- **Remaining before merge**: the two running CI legs; an approving review (all existing reviews are comments); and disposition of threads (1)–(3) above — (1)+(2) as one small commit, (3) declinable with the rationale here.

---

baijum:
@coderabbitai Regarding the outside-diff finding on `lib/srfi/231/storage-classes.sld` lines 223–224 ("Make the signed-zero precondition explicit" — the claim that `%f16-encode`'s `(eqv? x -0.0)` sign test relies on a distinction R7RS does not guarantee): declining this one as not applicable, for three independent reasons.

1. **R7RS-small §6.1 does guarantee it wherever it matters.** `eqv?` must return `#f` for two inexact numbers that yield different results under some finite composition of standard arithmetic procedures. `+0.0` and `-0.0` are operationally distinguishable wherever both exist — `(/ 1.0 0.0)` ⇒ `+inf.0` vs `(/ 1.0 -0.0)` ⇒ `-inf.0` — so in any R7RS implementation with distinct signed zeros, `(eqv? 0.0 -0.0)` ⇒ `#f` is *required*, not merely permitted. The code comment's "eqv? does, per R7RS" is accurate for every implementation where the distinction is observable at all.

2. **The degenerate case is self-consistent.** In a hypothetical implementation with a single zero, the `-0.0` literal denotes that same zero on *both* sides of the codec — the encode-side `(eqv? x -0.0)` test and the decode-side `-0.0` return value collapse together — so zero round-trips to zero either way. Only the stored bit pattern differs, which such an implementation cannot observe.

3. **It is the reference's own semantics on the actual host.** The sign test transliterates the reference implementation's `##flcopysign` behavior, and the concrete host is Kaappi, whose NaN-boxed flonums have fully distinguishable signed zeros — `-0.0 ↔ #x8000` is a tested directed edge in this PR's own suite, and the exhaustive 65536-pattern round-trip pins it.

No code or comment change needed.

---

docs/dev/testing.md
```diff
@@ -276,10 +276,15 @@ Conventions specific to this suite:
...
-  which means its entry is stale and hiding real coverage: prune it.
+  its reason, issue reference, and the *exact* number of evaluations
+  expected to diverge under it (shared test forms run once per storage
+  class, so one id can account several rows) — the c64/c128
+  native-vector representation choice, the unsafe-view UB choice, the
+  Gambit string-mutability expectation. The suite exits nonzero only on
+  *unexpected* failures — or when an entry's divergence count doesn't match
+  exactly: zero observed means the entry is stale and hiding real coverage
+  (prune it), and more than recorded means an undocumented failure is
+  absorbing into the entry.
```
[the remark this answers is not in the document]
baijum: Landed in #2385 (this PR merged before disposition): the fewer-but-nonzero case — "fewer than recorded but nonzero means the entry over-accounts and must be re-counted" — is now documented in the same testing.md bullet, alongside the never-executed case from the sibling thread, so the guide matches the epilogue's verdict logic case-for-case.

tests/scheme/srfi/srfi231-official-transform.py
```diff
@@ -797,27 +804,50 @@ def strip_time(src):
...
-                         (not (vector-ref diverged-ids id)))
-                (set! resolved-divergences (+ resolved-divergences 1))
-                (display "DIVERGENCE-RESOLVED ") (display id)
-                (display " -- prune it from known-divergences (")
-                (display (cdr entry)) (display ")") (newline))))
+            (let ((id (car entry))
+                  (expected (cadr entry)))
+              (when (vector-ref executed-tests id)
+                (let ((observed (vector-ref diverged-counts id)))
+                  (cond ((= observed 0)
```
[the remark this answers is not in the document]
baijum: Landed in #2385 (this PR merged before disposition), taking the "count an unexecuted entry as a mismatch" route with its own wording rather than reusing the stale-entry message, per the review assessment: an id that never executes now prints DIVERGENCE-NEVER-EXECUTED — "no evaluation ran under this id (its test form is gone from the suite; prune the entry)" — and counts into the mismatch total that gates the exit status. Verified by a scratch-copy negative test (bogus `'(9999 2 . …)` entry → exit 1, previously silent); the real suite is unchanged-green since all three real ids execute.

PR #2385, opened 2026-08-27
Close the last srfi231-official divergence-accounting gaps (#2383 review)

baijum:
Follow-up to #2383 (merged): the two CodeRabbit threads the reviewer's post-merge assessment verdicted **valid** landed after the merge, so they come in here. The third open thread — the signed-zero `eqv?` finding — was declined on the PR with rationale (R7RS §6.1 requires the distinction wherever observable; the degenerate single-zero case is self-consistent on both sides of the codec; it's the reference's own `##flcopysign` semantics on a host with distinguishable signed zeros) and needs no change.

## 1. Document the fewer-than-expected mismatch case (`docs/dev/testing.md`)

The guide described only the zero-observed and greater-than-recorded cases; the epilogue's third case — `0 < observed < expected`, reported as "the entry over-accounts (re-count it)" — is now in the same bullet, together with the never-executed case added here, so the doc matches the verdict logic exactly.

## 2. Fail when a known-divergence id does not execute (`srfi231-official-transform.py`, suite regenerated)

The epilogue's `(when (vector-ref executed-tests id) …)` guard silently skipped an entry whose id never executes — if a future regeneration drops a test form entirely, its table entry passes with observed count zero and neither `DIVERGENCE-RESOLVED` nor a count mismatch fires. That was the last accounting gap. An unexecuted id is now a hard failure with its own wording — `DIVERGENCE-NEVER-EXECUTED … its test form is gone from the suite; prune the entry` — counted into the mismatch total that gates the exit status. The generated file was regenerated from the transform, never hand-edited.

## Validation

- Real suite unchanged-green: `10922 passed, 5 known divergences, 0 unexpected, 0 resolved, 0 count mismatches` (all three real ids execute).
- Negative test (scratch copy, not committed): a correctly-quoted bogus entry `'(9999 2 . …)` prints `DIVERGENCE-NEVER-EXECUTED 9999` and **exits 1**, where the old guard exited 0 silently.
- fmt corpus: 938 files, zero semantic drift, idempotent.

### DIFF
--- a/tests/scheme/srfi/srfi231-official-transform.py
+++ b/tests/scheme/srfi/srfi231-official-transform.py
@@ -857,6 +857,19 @@ NEW_SUM = '''
           0 1))'''
 src = src.replace(OLD_SUM, NEW_SUM)
 
+# --- P6: generation-time guard (complement to the runtime epilogue): every
+# known-divergence id must belong to a test form the scanner actually found
+# and numbered. A dropped/renamed upstream form would otherwise surface only
+# as the committed suite's DIVERGENCE-NEVER-EXECUTED after a ~150 s run;
+# here it fails the regeneration itself, instantly. The runtime check stays
+# authoritative for CI -- it guards the committed artifact, not the step.
+div_ids = [int(m) for m in re.findall(r"'\((\d+) \d+ \. \"", NEW_REPORT)]
+assert div_ids, 'no known-divergence ids parsed from NEW_REPORT (regex drifted?)'
+missing_ids = [i for i in div_ids if i not in {m['id'] for m in test_meta}]
+assert not missing_ids, (
+    'known-divergence ids with no test form in the generated suite '
+    '(dropped or renumbered upstream?): %r' % missing_ids)
+
 open(OUT, 'w').write(src)
 # The id->original-line map is debugging metadata for triaging failures;
 # write it only on request so regeneration leaves no untracked droppings.
--- a/tests/scheme/srfi/srfi231-official-transform.py
+++ b/tests/scheme/srfi/srfi231-official-transform.py
@@ -690,10 +690,11 @@ HEADER = ''';;; srfi231-official.scm -- the OFFICIAL SRFI 231 test suite (test-a
 ;;; ids whose failure encodes a documented kaappi-vs-reference divergence,
 ;;; each with an issue reference and the exact number of evaluations expected
 ;;; to diverge under it (shared test forms run once per storage class). The
-;;; suite exits nonzero only on UNEXPECTED failures -- or when an entry's
-;;; divergence count does not match exactly: zero observed means the entry is
-;;; stale and hiding real coverage (prune it), and more than recorded means
-;;; an undocumented failure is absorbing into the entry.
+;;; suite exits nonzero only on UNEXPECTED failures -- or when an entry does
+;;; not account exactly: an id that never executes, zero observed (stale
+;;; entry hiding real coverage -- prune it), more than recorded (an
+;;; undocumented failure absorbing into the entry), or fewer but nonzero
+;;; (over-accounting -- re-count it).
 ;;; Error-EXPECTING tests count as passes when any error is raised; only the
 ;;; Gambit message text differs.
 ;;;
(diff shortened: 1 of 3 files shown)


tests/scheme/srfi/srfi231-official-transform.py
```diff
@@ -805,35 +806,41 @@ def strip_time(src):
...
-                         (display (if (> observed expected)
-                                      "an undocumented failure is hiding under this id"
-                                      "the entry over-accounts (re-count it)"))
-                         (display " (")
-                         (display (cddr entry)) (display ")") (newline)))))))
+              (if (not (vector-ref executed-tests id))
+                  (begin
+                    (set! divergence-count-mismatches
+                          (+ divergence-count-mismatches 1))
+                    (display "DIVERGENCE-NEVER-EXECUTED ") (display id)
```
baijum, line 823: Non-blocking follow-up idea: the transform already collects every generated test id in `test_meta`, and the known-divergence ids live in this file's own `MID` constant a few lines up. A generation-time assertion — every table id must appear in `{m['id'] for m in test_meta}` — would catch the dropped-form case at regeneration time (instant, at the `wrote …` line) instead of one ~150 s suite run later. The runtime check added here is still the one that matters for CI (it guards the committed artifact, not just the regeneration step), so this would be a complement, not a replacement. Fine to land as-is.

baijum (commented):
Reviewed with independent verification — the change is correct and every claim in the PR description reproduces.

**What I verified locally (fresh `main` binary at a6cf52db):**

- **Regeneration is faithful**: ran `srfi231-official-transform.py` against `srfi231-official-fixtures/test-arrays.scm.gambit` on this branch — output is byte-identical to the committed `srfi231-official.scm`.
- **Real suite unchanged-green**: `10922 passed, 231 error-message-only, 5 known divergences, 0 unexpected failures, 0 resolved divergences, 0 count mismatches, out of 10927 evaluations`, exit 0 — exactly the numbers in the description.
- **Negative test reproduces**: a scratch copy with a bogus `'(9999 2 . …)` entry prints `DIVERGENCE-NEVER-EXECUTED 9999 -- no evaluation ran under this id …` and exits 1, where the old `when` guard skipped it silently.
- **No false-positive risk from conditional execution**: every `test`/`test-err`/`test-mv` form in the generated suite runs unconditionally (no `cond-expand`/feature gating), so a `DIVERGENCE-NEVER-EXECUTED` failure can only mean what it says — the form is gone.
- The new counter feeds `divergence-count-mismatches`, which both the summary line and the `(exit …)` predicate already read, so the gate is real, and the doc bullet in `testing.md` now matches all four verdict cases exactly.

One minor cosmetic note, not worth a change: the never-executed case shows up in the summary line under the "count mismatches" bucket, which slightly under-describes it — but the labeled `DIVERGENCE-NEVER-EXECUTED` diagnostic line precedes it and a third counter would just bloat the summary for a state that should never persist.

One non-blocking suggestion inline. LGTM.

tests/scheme/srfi/srfi231-official-transform.py
baijum: Addressed in cda948fc — implemented as suggested, as a complement to the runtime epilogue:

- the ids are parsed out of `NEW_REPORT`'s own table with a strict regex (`'\((\d+) \d+ \. \"`), so the table stays the single source of truth and no parallel Python constant can drift from it; an empty parse is itself an assertion failure, so the guard cannot rot silently if the entry format changes;
- every parsed id must appear in `{m['id'] for m in test_meta}` — the set the scanner already collects — failing regeneration instantly at the dropped/renamed-upstream-form case instead of a ~150 s suite run later;
- the assertion runs **before** `open(OUT, 'w')`, so a failed regeneration writes nothing and cannot clobber a good artifact.

Validated: normal regeneration exits 0 and leaves the committed `srfi231-official.scm` byte-identical (`git diff` empty); a scratch copy of the transform with id 351 renumbered to 9999 aborts with `known-divergence ids with no test form in the generated suite (dropped or renumbered upstream?): [9999]`, exit 1, and writes no output file.

Agreed on the division of labor you stated: the runtime `DIVERGENCE-NEVER-EXECUTED` check remains the one that matters for CI, since it guards the committed artifact even when nobody regenerates.

PR #2387, opened 2026-08-27
Switch SRFI 231 c64/c128 bodies to the reference representation

### LINKED ISSUE #2382: SRFI 231: c64/c128 storage classes reject the reference's f32/f64vector data representation in make-specialized-array-from-data

baijum:
Closes #2382.

## The problem

`c64-storage-class`/`c128-storage-class` were backed by native `c64vector`/`c128vector`, so the reference implementation's data shape — an **even-length `f32vector`/`f64vector` of interleaved re/im pairs** — was rejected by `make-specialized-array-from-data`:

```scheme
(make-specialized-array-from-data (make-f32vector 10) c64-storage-class)
;; ERROR: data is not compatible with storage-class
;; reference: returns a 5-element c64 array sharing the f32vector
```

That was known-divergence id 150 on the official suite (filed as #2382 when the f16 work surfaced it).

## Two findings settled the design

1. **A converting `data->body` is spec-illegal.** SRFI 231's spec says `data?` returns `#t` *iff* `data->body` returns a body **sharing the data, without copying**. So the "accept both representations, convert on the way in" middle path is off the table — accepting the reference's data shape is possible only by actually using it as the body. (The spec explicitly allows either representation: "another implementation … might make another choice.")
2. **Kaappi's `c64vector` already uses the reference's exact byte layout** — `src/types_numeric.zig` documents "2 consecutive f32s/f64s (real, imag) packed contiguously — never boxed". The switch is a **change of type tag, not of memory shape**, which removes the usual argument against it (throwing away a richer representation — there isn't one).

With the port's reference-fidelity philosophy (u1, f16, the #2355 checker), switching is the right call.

## What changed

- `lib/srfi/231/storage-classes.sld`: c64/c128 rebuilt through one `%complex-storage-class` helper — a faithful port of the reference's `make-complex-storage-classes`: getter reassembles the interleaved pair (an inexact zero imag stays complex, kaappi#2269, exactly like the native decode), setter explodes into the two float slots, maker fills alternating re/im, length halves the physical float count, copier is the float block copy, `data?`/`data->body` accept exactly the even-length float vectors (zero-copy identity). The now-unused `(srfi 160 c64)`/`(srfi 160 c128)` imports are dropped.
- **User-visible consequences** (CHANGELOG `### Changed`): `(array-body A)` for a c64/c128 array now reports the float vector; the storage-class copier counts floats (2 per complex element); `c64vector`/`c128vector` data is no longer accepted directly.

## Tests

- `check-storage-class` learns a body-units-per-element option (2 for complex classes) so its copier exercise stays meaningful.
- New representation section in `srfi231-storage-classes.scm`: `data?` acceptance/rejection (even / odd / `c64vector`), `data->body` identity, interleave round-trips, setter explosion, maker fill.
- End-to-end zero-copy proof in `srfi231-arrays.scm`: `make-specialized-array-from-data` over an `f32vector` shares so thoroughly that mutating the caller's vector is visible through the array.

## Official suite

Divergence id 150 pruned, suite regenerated. Pleasing detail: run *before* pruning, the accounting from #2385 flagged the resolution on its own — `DIVERGENCE-RESOLVED 150 -- prune it from known-divergences` — the enforcement working exactly as designed. The table is now down to the two genuinely unavoidable entries: **147** (Gambit string mutability) and **351** (unsafe-view checking).

## Validation

- `srfi231-storage-classes.scm` 219 passes; `srfi231-arrays.scm` 90; all seven other SRFI 231 suites green.
- Official suite: **10,924 passed, 3 known divergences, 0 unexpected failures, 0 resolved, 0 count mismatches**.
- fmt corpus: 938 files, zero semantic drift, idempotent.
- `run-all.sh`: 720 Scheme files + R7RS 1395/1395 → **2,115 pass, 0 fail**.

Rebased onto `main` at `fb703515` (post-#2385); the two #2385 review commits dropped as already-upstream, validation re-run on the rebased tree.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   test id (16 storage-class rows re-use one id) can no longer hide behind a
   documented divergence.
 
+### Changed
+
+- **SRFI 231 c64/c128 array bodies use the reference implementation's
+  interleaved-float representation** (#2382) — an `f32vector`/`f64vector`
+  of twice the logical length holding real/imaginary pairs, instead of
+  native `c64vector`/`c128vector` (the byte layout is identical either
+  way: 2 consecutive f32s/f64s per element). Even-length float vectors
+  are now accepted zero-copy by `make-specialized-array-from-data`, so
+  reference-coupled portable code and the official suite's fixtures work
+  unchanged — the spec's `data?` contract (shares, never copies) makes
+  this the only spec-legal way to accept that data shape. Consequences:
+  `(array-body A)` for a c64/c128 array now reports the float vector,
+  the storage-class copier counts floats (2 per complex element), and
+  `c64vector`/`c128vector` data is no longer accepted directly.
+  Official-suite divergence id 150 pruned; the known-divergence table is
+  down to the two unavoidable entries (string mutability, unsafe-view
+  checking).
+
 ## [0.25.0] - 2026-08-27
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,8 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   unchanged — the spec's `data?` contract (shares, never copies) makes
   this the only spec-legal way to accept that data shape. Consequences:
   `(array-body A)` for a c64/c128 array now reports the float vector,
-  the storage-class copier counts floats (2 per complex element), and
-  `c64vector`/`c128vector` data is no longer accepted directly.
+  and `c64vector`/`c128vector` data is no longer accepted directly.
   Official-suite divergence id 150 pruned; the known-divergence table is
   down to the two unavoidable entries (string mutability, unsafe-view
   checking).
--- a/lib/srfi/231/storage-classes.sld
+++ b/lib/srfi/231/storage-classes.sld
@@ -161,53 +161,59 @@
     ;; never boxed -- so this is a change of type tag, not of memory
     ;; shape; the spec explicitly allows either representation ("another
     ;; implementation ... might make another choice"), and reference
-    ;; fidelity is what interoperates. Consequence: the copier's element
-    ;; granularity is FLOATS (2 per complex element), and
+    ;; fidelity is what interoperates. Consequence:
     ;; c64vector/c128vector data is no longer accepted -- convert with
-    ;; make-specialized-array's maker or a copy loop if needed.
+    ;; make-specialized-array's maker or a copy loop if needed. Like the
+    ;; reference's own c64vector-copy!/c128vector-copy! wrappers, the
+    ;; copier takes LOGICAL complex-element offsets and scales by 2
+    ;; internally -- the same units as every other storage-class field.
     (define (%complex-storage-class float-ref float-set! make-float-vector
                                     float-copy! float-length vec?
                                     class-name type-name)
-      (make-storage-class
-       ;; getter -- reassemble the interleaved pair (an inexact zero
-       ;; imag stays complex, kaappi#2269, exactly like the native
-       ;; c64vector decode)
-       (lambda (body i)
-         (make-rectangular (float-ref body (* 2 i))
-                           (float-ref body (+ (* 2 i) 1))))
-       ;; setter -- explode into the interleaved pair (f32 storage rounds)
-       (lambda (body i obj)
-         (float-set! body (* 2 i) (real-part obj))
-         (float-set! body (+ (* 2 i) 1) (imag-part obj)))
-       ;; checker
-       %inexact-complex-checker
-       ;; maker -- the fill exploded into alternating re/im
-       (lambda (n val)
-         (let* ((l (* 2 n))
-                (re (real-part val))
-                (im (imag-part val))
-                (result (make-float-vector l)))
-           (do ((i 0 (+ i 2)))
-               ((= i l) result)
-             (float-set! result i re)
-             (float-set! result (+ i 1) im))))
-       ;; copier
-       float-copy!
-       ;; length -- half the physical float count
-       (lambda (body) (quotient (float-length body) 2))
-       ;; default
-       (make-rectangular 0.0 0.0)
-       ;; data?
-       (lambda (data)
-         (and (vec? data) (even? (float-length data))))
-       ;; data->body -- identity on even-length float vectors
-       (lambda (data)
-         (if (and (vec? data) (even? (float-length data)))
-             data
-             (error (string-append "Expecting a " type-name
-                                   " with an even number of elements passed to (storage-class-data->body "
-                                   class-name "): ")
-                    data)))))
+      (let ((complex-data?
+             (lambda (data) (and (vec? data) (even? (float-length data))))))
+        (make-storage-class
+         ;; getter -- reassemble the interleaved pair (an inexact zero
+         ;; imag stays complex, kaappi#2269, exactly like the native
+         ;; c64vector decode)
+         (lambda (body i)
+           (make-rectangular (float-ref body (* 2 i))
+                             (float-ref body (+ (* 2 i) 1))))
+         ;; setter -- explode into the interleaved pair (f32 storage rounds)
+         (lambda (body i obj)
+           (float-set! body (* 2 i) (real-part obj))
+           (float-set! body (+ (* 2 i) 1) (imag-part obj)))
+         ;; checker
+         %inexact-complex-checker
+         ;; maker -- the fill exploded into alternating re/im; a uniform
+         ;; fill (eqv?: the default 0.0+0.0i is the overwhelmingly common
+         ;; case, and -0.0/0.0 or NaN mismatches still take the loop)
+         ;; collapses to one native fill
+         (lambda (n val)
+           (let* ((l (* 2 n))
+                  (re (real-part val))
+                  (im (imag-part val)))
+             (if (eqv? re im)
+                 (make-float-vector l re)
+                 (let ((result (make-float-vector l)))
+                   (do ((i 0 (+ i 2)))
+                       ((= i l) result)
+                     (float-set! result i re)
+                     (float-set! result (+ i 1) im))))))
+         ;; copier -- the reference's c64vector-copy! wrapper: logical
+         ;; element offsets, scaled by 2 onto the float block copy
+         (lambda (to at from start end)
+           (float-copy! to (* 2 at) from (* 2 start) (* 2 end)))
+         ;; length -- half the physical float count
+         (lambda (body) (quotient (float-length body) 2))
+         ;; default
+         (make-rectangular 0.0 0.0)
+         ;; data?
+         complex-data?
+         ;; data->body -- identity on even-length float vectors
+         (%checked-data->body complex-data? class-name
+                              (string-append type-name
+                                             " with an even number of elements")))))
 
     (define c64-storage-class
       (%complex-storage-class f32vector-ref f32vector-set! make-f32vector
--- a/lib/srfi/231/storage-classes.sld
+++ b/lib/srfi/231/storage-classes.sld
@@ -32,7 +32,7 @@
   (import (scheme base)
           (srfi 160 s8) (srfi 160 s16) (srfi 160 s32) (srfi 160 s64)
           (srfi 160 u16) (srfi 160 u32) (srfi 160 u64)
-          (srfi 160 f32) (srfi 160 f64) (srfi 160 c64) (srfi 160 c128)
+          (srfi 160 f32) (srfi 160 f64)
           ;; bitwise-and/ior/not for u1's bit-packing (R7RS-small has no
           ;; bitwise primitives of its own)
           (srfi 60)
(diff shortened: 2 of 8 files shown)


lib/srfi/231/storage-classes.sld
```diff
@@ -144,14 +144,79 @@
...
+         (let* ((l (* 2 n))
+                (re (real-part val))
+                (im (imag-part val))
+                (result (make-float-vector l)))
+           (do ((i 0 (+ i 2)))
+               ((= i l) result)
+             (float-set! result i re)
+             (float-set! result (+ i 1) im))))
+       ;; copier
+       float-copy!
```
baijum: **Correctness — the copier field is not a faithful port: it counts floats, the reference's counts complex elements.**

The reference does **not** pass the raw float copier here. `generic-arrays.scm:137-140` defines a ×2-scaling wrapper and installs *that* as the copier (line 1417 of the macro):

```scheme
(define (c64vector-copy! to at from start end)
  (f32vector-copy! to (fx* 2 at) from (fx* 2 start) (fx* 2 end)))
```

So in the reference, `(copier to at from start end)` takes **logical complex-element** offsets for c64/c128 — the same units as getter/setter/length/maker, and the same units its `%%move-array-elements` passes (indexer-derived offsets). As written here, kaappi's c64/c128 copier is the only storage-class field in the record — and the only copier among all 16 classes — indexed in a different unit.

Failure: reference-coupled code (the exact audience #2382 targets) doing `((storage-class-copier c64-storage-class) to 1 from 0 2)` — "copy elements 0–1 into slot 1" — copies floats 0–1 to float offset 1: half the data, at an odd offset, so real parts silently land in imaginary slots. No error. Nothing in-tree calls the copier (per the u1 comment below), so no suite catches it.

```suggestion
       (lambda (to at from start end)
         (float-copy! to (* 2 at) from (* 2 start) (* 2 end)))
```

tests/scheme/srfi/srfi231-storage-classes.scm
```diff
@@ -40,15 +46,16 @@
     (setter body 1 value)
     (test-assert (%same? value (getter body 1)))
     ;; exercise the copier field too -- otherwise a wrong copy procedure
-    ;; would pass every other assertion here undetected
+    ;; would pass every other assertion here undetected; the copier counts
+    ;; BODY units, so complex classes copy 2 floats per logical element
     (let ((copied (maker 3 (storage-class-default sc))))
-      ((storage-class-copier sc) copied 0 body 0 3)
+      ((storage-class-copier sc) copied 0 body 0 (* 3 units-per-element))
```
baijum: **Test-coverage — `units-per-element` adapts the test to the copier bug instead of detecting it.**

Had this stayed `3` (logical elements, the reference contract), it would have failed and exposed the granularity divergence flagged on `storage-classes.sld:195`. Once the copier gets the reference's ×2 wrapper, all 16 classes copy logical units again — this knob and the `#t 2` at the c64/c128 call sites can be reverted wholesale.

Separately: the copier exercise copies an aligned full range (`copied 0 body 0 n`), which can't distinguish the two granularities' offset misalignment even in principle. Consider a non-zero `at`/`start` so a future unit regression actually trips it.

CHANGELOG.md
```diff
@@ -22,6 +22,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
...
+  interleaved-float representation** (#2382) — an `f32vector`/`f64vector`
+  of twice the logical length holding real/imaginary pairs, instead of
+  native `c64vector`/`c128vector` (the byte layout is identical either
+  way: 2 consecutive f32s/f64s per element). Even-length float vectors
+  are now accepted zero-copy by `make-specialized-array-from-data`, so
+  reference-coupled portable code and the official suite's fixtures work
+  unchanged — the spec's `data?` contract (shares, never copies) makes
+  this the only spec-legal way to accept that data shape. Consequences:
+  `(array-body A)` for a c64/c128 array now reports the float vector,
+  the storage-class copier counts floats (2 per complex element), and
```
baijum: **Doc/claim consistency — "the storage-class copier counts floats" is a divergence from the reference, not part of the faithful port.**

The reference's complex copier counts complex elements (`generic-arrays.scm:137-140` wraps the float copier with ×2 scaling on `at`/`start`/`end`; the macro installs that wrapper at line 1417). If the copier is fixed to match, this line — and the matching claims in the `storage-classes.sld` comment block and `docs/dev/srfi-implementation-notes.md` — should say the copier still counts complex elements, with the ×2 living inside it.

tests/scheme/srfi/srfi231-official-transform.py
```diff
@@ -741,7 +741,6 @@ def strip_time(src):
 (define diverged-counts (make-vector 10000 0))
 (define known-divergences
   (list '(147 1 . "R7RS strings are mutable; the suite encodes Gambit's immutable-string expectation")
```
baijum, line 744: **Stale comment — the example names the entry this PR deletes.**

The preamble a few lines up still says "e.g. an f16 regression absorbed by the c64/c128 entry", but id 150 is gone as of this PR, so the example points at a nonexistent entry (here in `NEW_REPORT` and in the generated `srfi231-official.scm:126`). Fix it in the `.py` and regenerate — the `.scm` header forbids hand edits.

lib/srfi/231/storage-classes.sld
```diff
@@ -144,14 +144,79 @@
...
+       ;; length -- half the physical float count
+       (lambda (body) (quotient (float-length body) 2))
+       ;; default
+       (make-rectangular 0.0 0.0)
+       ;; data?
+       (lambda (data)
+         (and (vec? data) (even? (float-length data))))
+       ;; data->body -- identity on even-length float vectors
+       (lambda (data)
```
baijum: **Simplification — this re-inlines both the `data?` predicate and `%checked-data->body`.**

The `(and (vec? data) (even? (float-length data)))` test appears verbatim in `data?` (line 202) and again here, and the identity-or-error shape is exactly `%checked-data->body` (line 70) with two extra words in the message. Hoisting one predicate and reusing the helper produces a byte-identical error:

```scheme
(let ((ok? (lambda (d) (and (vec? d) (even? (float-length d))))))
  (make-storage-class
   ...
   ok?
   (%checked-data->body ok? class-name
     (string-append type-name " with an even number of elements"))))
```

The concrete risk of the duplication: if one copy is later changed and the other missed, `data?` and `data->body` disagree — which is precisely the spec contract the comment above quotes ("#t if and only if data->body returns a body sharing data").

tests/scheme/srfi/srfi231-storage-classes.scm
```diff
@@ -24,10 +25,15 @@
...
-;;; assertion is a separate, optional check. ---
+;;; assertion is a separate, optional check. A second optional is the
+;;; copier's body-units per logical element -- 1 everywhere except the
+;;; complex classes, whose interleaved-float bodies hold 2 (see the c64
+;;; section below). ---
 (define (%same? a b) (if (and (number? a) (number? b)) (= a b) (equal? a b)))
 
-(define (check-storage-class sc value bad-value expected-default . rejects-bad?)
+(define (check-storage-class sc value bad-value expected-default . opts)
+  (let ((rejects-bad? (or (null? opts) (car opts)))
```
baijum: **Conventions/simplification — the wrapping `let` leaves the whole body at its old indentation.**

The enclosed `let*` (and everything under it) now sits at the same column as the binding form it's nested inside, against the repo's Scheme style rule (root `CLAUDE.md`: "Scheme (libraries): 2-space indentation, R7RS style"). Internal defines would avoid both the nesting and the null-chains, with no re-indent:

```scheme
(define (check-storage-class sc value bad-value expected-default . opts)
  (define rejects-bad? (if (pair? opts) (car opts) #t))
  (define units-per-element
    (if (and (pair? opts) (pair? (cdr opts))) (cadr opts) 1))
  (let* ((maker (storage-class-maker sc))
   ...
```

(And if the copier gets the reference's ×2 wrapper per the `storage-classes.sld:195` comment, `units-per-element` disappears entirely.)

lib/srfi/231/storage-classes.sld
```diff
@@ -144,14 +144,79 @@
...
+         (make-rectangular (float-ref body (* 2 i))
+                           (float-ref body (+ (* 2 i) 1))))
+       ;; setter -- explode into the interleaved pair (f32 storage rounds)
+       (lambda (body i obj)
+         (float-set! body (* 2 i) (real-part obj))
+         (float-set! body (+ (* 2 i) 1) (imag-part obj)))
+       ;; checker
+       %inexact-complex-checker
+       ;; maker -- the fill exploded into alternating re/im
+       (lambda (n val)
```
baijum: **Efficiency (minor) — the interpreted 2n-store loop runs even for uniform fills.**

Every `make-specialized-array` without an explicit initial value calls this maker with the default `0.0+0.0i`, where `re` and `im` are the same float — a 1M-element c128 array pays 1M interpreted iterations and 2M `f64vector-set!` calls where one native fill suffices:

```scheme
(if (eqv? re im)
    (make-float-vector l re)
    (do ...))
```

`eqv?`, not `=`, so `-0.0`/`0.0` and NaN fills stay correct.

lib/srfi/231/storage-classes.sld
baijum: Confirmed and fixed in 0e3b39a1 — you found a real bug in the port. Verified against the reference source: `generic-arrays.scm:137-140` defines `c64vector-copy!`/`c128vector-copy!` as ×2-scaling wrappers, and the class macro installs those, so the copier takes **logical** complex-element offsets like every other field. My raw `float-copy!` installation made the complex classes the only field in the only class indexed in a different unit — exactly the silent half-data/odd-offset corruption you described, and nothing in-tree calls the copier so no suite caught it.

Fixed with the reference's own wrapper shape (the suggested code, with `fx*` → `*` for R7RS-small):
```scheme
(lambda (to at from start end)
  (float-copy! to (* 2 at) from (* 2 start) (* 2 end)))
```
Your exact repro `(copier to 1 from 0 2)` verified by hand: `1.0+2.0i` lands at slot 1, `3.0+4.0i` at slot 2, slots 0/3 untouched.

tests/scheme/srfi/srfi231-storage-classes.scm
baijum: You're right that the knob was adapting the test to the bug — reverted wholesale in 0e3b39a1: `check-storage-class` is back to its original `. rejects-bad?` shape with the plain `3`-element copy, the `#t 2` call sites are gone, and with the ×2 wrapper all 16 classes copy logical units again.

And taken on the offset point too: the copier exercise now runs twice — the aligned full range plus a non-zero-offset copy, `(copier shifted 1 body 1 3)`, asserting value at slot 1 and default at slot 2. That one discriminates the two granularities exactly as you reasoned: a raw-float copier passes the aligned copy while a slot-1 destination at float offset 1 mispairs re/im, so `(getter shifted 1)` no longer returns the value. Pinned across all 15 copier-bearing classes (the suite went 219 → 249 passes).

CHANGELOG.md
baijum: Fixed in 0e3b39a1 — with the ×2 wrapper there is no copier divergence at all, so the CHANGELOG consequence sentence is simply dropped (leaving the two real ones: `array-body` reports the float vector, and `c64vector`/`c128vector` data is no longer accepted directly). The `.sld` comment block now states the reference-wrapper shape explicitly ("takes LOGICAL complex-element offsets and scales by 2 internally — the same units as every other storage-class field"). The notes file only names `copier` in the 9-field list, no granularity claim there — checked, nothing to fix.

tests/scheme/srfi/srfi231-official-transform.py
baijum: Fixed in 0e3b39a1 — the example now points at a live entry: "a third unsafe-view evaluation absorbed by the 351 entry, whose site legitimately fires twice." Regenerated from the transform, not hand-edited (the header's own rule).

lib/srfi/231/storage-classes.sld
baijum: Done in 0e3b39a1, essentially as suggested: the even-length predicate is hoisted once as `complex-data?`, installed as `data?`, and `data->body` is `(%checked-data->body complex-data? class-name (string-append type-name " with an even number of elements"))` — byte-identical error message, one predicate, no drift surface. (Kept the `let` over internal-define style to match the surrounding file's idiom rather than the suggested `let ((ok? ...))` spelling of the same shape.)

tests/scheme/srfi/srfi231-storage-classes.scm
baijum: Mooted the way you anticipated: with the copier fixed to logical units, `units-per-element` (and the `#t 2` call sites) disappeared entirely, and `check-storage-class` reverted to its original `. rejects-bad?` + single `let*` shape — so the re-indentation problem dissolved rather than needed solving. The only addition to the original is the second, offset-sensitive copier exercise described in the sibling thread.

lib/srfi/231/storage-classes.sld
baijum: Taken in 0e3b39a1, exactly as suggested: `(if (eqv? re im) (make-float-vector l re) (do ...))` — `eqv?` so a `-0.0`/`0.0` or NaN fill mismatch still takes the interleaved loop, and the default `0.0+0.0i` (the no-initial-value `make-specialized-array` path) collapses to one native fill. Same reasoning on `=`: it would wrongly merge `0.0`/`-0.0` fills and clobber the sign.

baijum:
Re-reviewed `0e3b39a1` — **all 7 review findings are addressed, and the changes look good.** No new issues found.

Verified per finding:

1. **Copier granularity** — now the reference's wrapper (logical element offsets, ×2 onto the float block copy). Runtime-checked: `(copier to 1 from 1 3)` on `[1+2i, 3+4i, 5+6i]` lands `3+4i, 5+6i` at slots 1–2 with slot 0 untouched — exactly the reference contract.
2. **Test knob** — `units-per-element` gone, signature reverted, and the new second copier exercise with non-zero `at`/`start` genuinely discriminates: a float-granular copier would put `1.0+0.0i` where `1.0+2.0i` belongs. Nice touch making it run for all 16 classes.
3. **CHANGELOG / comment claims** — "copier counts floats" removed; the `.sld` comment now correctly cites the reference's own `c64vector-copy!` wrappers.
4. **Stale divergence example** — replaced with the 351-based example in the `.py`; I regenerated `srfi231-official.scm` from the updated transform and it's byte-identical to the committed file.
5. **`data->body` dedup** — `complex-data?` shared between `data?` and `%checked-data->body`; error message stays byte-identical to the reference's.
6. **Indentation** — moot with the knob removed; the `%complex-storage-class` let wrapper body is properly re-indented.
7. **Maker fast path** — `eqv?`-guarded native fill; verified the uniform path (`7.0+7.0i`) and the loop path (`-1.5+2.5i`) both produce correct interleaving.

Also re-ran on this branch with an isolated `KAAPPI_HOME`: `srfi231-storage-classes.scm` **249 passes**, `srfi231-arrays.scm` **90 passes**, zero failures.

PR #2397, opened 2026-08-27
Add channel identity comparators to (kaappi fibers)

### LINKED ISSUE #2394: eqv?/equal? don't compare the shared pointer for promoted channel stubs (KEP-0002 §2 divergence)
KEP-0002 §2 (Identity) specifies:

> Two stubs for one `SharedChannel` are distinct heap objects, so raw `eq?` would report `#f` for "the same channel" seen from two threads. `eqv?`/`equal?` (and the scheduler's waiter matching, §5) compare the `shared` pointer when both operands are promoted channels (Unresolved question 4 confirms the exact predicate set).

None of that shipped:

- `eqvP` (src/primitives.zig) — no channel branch
- `deepEqual` (src/primitives.zig) — no channel branch
- `isEqv` (src/primitives_list.zig) — no channel branch; also backs `memv`/`member`/`assv` and hashtable `eqv?` key comparison
- `write` prints a bare `#<channel>` (src/printer.zig) — no stable shared id (the other half of KEP-0002 UQ4)
- `deepCopy`'s `.channel` arm allocates a fresh stub per copy with no per-heap dedup (src/gc_deep_copy.zig), so receiving the same channel in two messages yields two stubs that are not `eqv?`/`equal?` to each other even within one thread:

```scheme
(define ch1 (channel-receive shared))   ; the same channel arrives twice
(define ch2 (channel-receive shared))
(eqv? ch1 ch2)  ; => #f   (KEP-0002 §2 specifies #t)
```

Decide one of:

1. implement the `shared`-pointer compare in `eqv?`/`equal?`, and settle the rest of the UQ4 predicate set (`eq?` transparency, stable `write` id for shared channels), or
2. drop the promise via a KEP-0002 amendment and document stub identity as the semantics.

Context: KEP-0002 Unresolved question 4; the divergence was recorded in the KEP's 2026-08-27 as-implemented amendment. Not user-visible until channels round-trip through threads, which is exactly the KEP-0002 use case.

baijum:
Fixes #2394.

## Decision

Implements **option 1b** from the issue discussion — expose channel identity as a comparator surface rather than extending the global predicates. Two stubs for one `SharedChannel` are observationally indistinguishable but `eqv?`/`equal?` stay stub-identity, as the KEP-0002 2026-08-27 as-implemented amendment records; the SRFI-128 idiom is the R7RS-large answer for exactly this (per-type equality+hash), and it leaves the `eqvP`/`deepEqual`/`valueHash` coupling untouched. A future transparent-equality layer (option 1) can be added on top later — both would compute the same answer.

## What's added to `(kaappi fibers)`

- **`channel=? a b`** — `#t` iff both operands are channels backing the same `SharedChannel` (or the same unpromoted local channel; the mixed case only arises for genuinely different channels, since promotion rewrites the original in place and every stub is born promoted). Carries the same foreign-owner check as `channel-send`, like `channel-closed?` — it reads channel fields; `channel?` stays exempt as a total predicate.
- **`channel-hash ch [bound]`** — hashes the `shared` pointer, consistent with `channel=?` by construction (the comparator contract); mirrors `(hash obj [bound])` (SRFI 69) including the KP3002/KP3007 argument discipline. An unshared channel hashes identically here and under plain `(hash ch)`.
- **`channel-comparator`** — `(make-comparator channel? channel=? #f channel-hash)` built by calling the **real** `(srfi 128)` `make-comparator`, so `comparator?` and the field accessors agree with every hand-built comparator and the native `extractComparatorFns` bridge recognizes the record (`(make-hash-table (channel-comparator))` works). The library lazy-loads on demand — the `environment` procedure's own route — from a new `embedded_libraries` entry (`srfi/128.sld`, same build.zig pattern as `(srfi 181)`), so `--sandbox` and WASM keep working. The three procedures are the `(kaappi fibers)` registry's pristine exports, immune to top-level shadowing of those names.

The issue's core scenario, now green:

```scheme
(define registry (make-hash-table channel=? channel-hash)) ; or (channel-comparator)
;; worker thread receives ch from to-w, sends it back twice through to-p:
(let ((a (channel-receive to-p)) (b (channel-receive to-p)))
  (list (channel=? a b) (channel=? a ch))       ; => (#t #t)
  (hash-table-set! registry a 'reply-to)
  (hash-table-ref/default registry ch 'MISS))   ; => reply-to — any stub finds it
```

## Tests

- `src/tests_fibers.zig`: local identity/hash, the cross-thread stub-unification scenario (asserts `eq?`/`eqv?` stay `#f` while `channel=?` is `#t` — the divergence this PR addresses), hash-table dedup via both table flavors, and lazy comparator construction.
- `tests/scheme/smoke/channel-identity-2394.scm`: the end-to-end suite (11 SRFI-64 tests).
- Full verification run from the worktree: unit suite 1871 pass / 7 skipped; fiber tests also green under `-Dgc-stress=true`; `tests/scheme/run-all.sh` = **2116 pass / 0 fail** (R7RS 1395/1395, compile + wasm-differential included); `zig fmt --check`, `markdownlint-cli2`, and `zig build wasm` all clean.

## Caveats (documented in `docs/dev/thread-value-sharing.md`)

- **Hash stability across promotion**: a channel hashed before its first cross-thread send hashes its local address, after it the `SharedChannel`'s. Promotion is sticky — share before keying. This is the edge the issue's option-1b comment said to document.
- **`(srfi 113)` sets don't honor comparators in this port** (pre-existing): `113.sld`'s `%make-empty-set` builds its backing table with a zero-argument `make-hash-table`, so the issue comment's `(set channel-comparator …)` example does **not** dedup — `make-hash-table` (SRFI 69/125) is the working route. Fixing 113 is deliberately out of scope: routing its tables through comparator lambdas would drop every SRFI-113 set off the native `equal?` fast path.
- `unboundedHash` is now `pub` so `channel-hash` shares the exact unbounded hash rendering.
- Side effect of the embedding: `(import (srfi 128))\) now also works under `--sandbox` (consistently — `cond-expand` availability flips with it, per the #1649 derive-don't-list principle).

The KEP-0002 §2 amendment wording (promise replaced by the comparator API) lives in the `kaappi/keps` repo and is left for a follow-up there.

### DIFF
--- a/docs/dev/thread-value-sharing.md
+++ b/docs/dev/thread-value-sharing.md
@@ -141,17 +141,16 @@ is what SRFI-128 comparators exist for:
   `SharedChannel` (or the same unpromoted local channel; the mixed case only
   arises for genuinely different channels, since promotion rewrites the
   original in place and every stub is born promoted).
-- `channel-hash ch [bound]` — hashes the `shared` pointer, consistent with
+- `channel-hash ch [bound]` — hashes the channel's identity, consistent with
   `channel=?` by construction; mirrors `(hash obj [bound])` (SRFI 69).
 - `channel-comparator` — `(make-comparator channel? channel=? #f
   channel-hash)`, built by calling the **real** `(srfi 128)`
-  `make-comparator`. That library lazy-loads on first use (the `environment`
-  procedure's on-demand route; an `embedded_libraries` entry keeps it
-  loadable under `--sandbox` and on WASM, like `(srfi 181)`), so
-  `comparator?` and the field accessors agree with every hand-built
-  comparator, and the native `extractComparatorFns` bridge recognizes the
-  record. The three procedures are `(kaappi fibers)`'s pristine registry
-  exports, immune to top-level shadowing of those names.
+  `make-comparator`, so `comparator?` and the field accessors agree with
+  every hand-built comparator and the native `extractComparatorFns` bridge
+  recognizes the record. It is cached on the VM like
+  `default_random_source` — repeat calls return the same record — and its
+  three procedures are `(kaappi fibers)`'s pristine registry exports,
+  immune to top-level shadowing of those names.
 
 The reply-channel/registry idiom:
 
@@ -160,21 +159,45 @@ The reply-channel/registry idiom:
 (hash-table-set! registry stub 'reply-to) ; any stub of the channel finds it
 ```
 
-Two caveats:
-
-- **Hash stability across promotion.** A channel hashed before its first
-  cross-thread send hashes its local address; after promotion (sticky) the
-  `SharedChannel`'s. Key a channel into a table only after sharing it.
-- **`(srfi 113)` sets do not honor comparators in this port**: `113.sld`'s
-  `%make-empty-set` builds its backing table with a zero-argument
-  `make-hash-table` (default `equal?`), so `(set channel-comparator …)`
-  still dedups by pointer. Use `make-hash-table` (SRFI 69/125) directly for
-  channel-keyed tables.
+The hash input is the channel's identity **for its whole life**, not the
+representation-of-the-moment: its own tagged Value while unpromoted, and —
+once promoted — the `SharedChannel.identity_seed` `promoteChannel` preserved
+from exactly that Value before publishing. A table keyed before the first
+cross-thread send therefore still finds its entry after promotion rewrites
+the original in place, and every stub hashes the one seed. If the original
+object is later collected and its address reused, the collision is benign —
+equality still compares `shared` pointers. (Unpromoted channels hash through
+the same `valueHash` as plain `(hash ch)`, so the two agree on every target,
+wasm32's 32-bit `usize` included.)
 
 `channel=?` and `channel-hash` read channel fields, so they carry the same
-foreign-owner check as `channel-send` (a channel reached through a shared
+foreign-owner check as `channel-send`, now factored as
+`primitives_fiber.checkChannelOwner` (a channel reached through a shared
 global is a diagnosis, not a comparison); `channel?` stays exempt as a total
 predicate.
+
+`(srfi 113)` sets honor their comparator since kaappi#2394: `113.sld`'s
+`%make-empty-set`/`%make-empty-bag` pass the comparator to
+`make-hash-table`, which unwraps the record into its equality/hash pair —
+and the native `detectMode` keeps the fast path for the standard
+comparators, whose equality fields hold the real `eq?`/`eqv?`/`equal?`, so
+`make-eq-comparator` sets pay nothing for the fix. Before it, every set's
+backing table was built with a zero-argument `make-hash-table` (default
+`equal?`) and any non-`equal?` comparator silently deduped by pointer.
+
+**Loading `(srfi 128)` safely across threads.** `vm.libraries` reaches
+child VMs as a struct copy (unlike `globals`, which got the pointer+lock
+treatment — see `vm.zig`'s `initForThread`), and `markVmRoots` gates
+library marking behind `owns_globals`. A lazy library load from a worker
+thread would therefore put() into bucket storage the root reads unlocked
+and leave child-heap exports unmarked by every collector.
+`threadStartImpl` closes this by pre-loading `(srfi 128)` on the spawning
+VM before any child exists (`ensureComparatorLibraryLoaded` — best-effort;
+the process's first `make-thread` is necessarily the root, so the load is
+always pre-children and read-only afterwards), and an
+`embedded_libraries` entry keeps the library loadable under `--sandbox`,
+on WASM, and — as the last resort after a disk miss — on a binary with no
+lib tree to read at all.
 
 ## The globals route
 
@@ -366,6 +389,8 @@ thread gets its own VM and GC with an independent heap.
 | `markLiveChildRoots` | `src/primitives_srfi18.zig` | kaappi#1933: registered on the **root** GC as `gc.child_marker`; the root's collector stops every live child at a dispatch-loop safepoint (or finds it already parked / in an FFI call) and marks its roots with the root's gc, so a parent-heap object referenced only from a live child's registers is never freed under it. Children spawned mid-collection spin on `collection_in_progress` before their first shared-globals read. The child side reports `collection_state` (`.running`/`.parked`/`.stopped`/`.in_native`) from the safepoint (`stopForCollection`), the park (`parkOnReactor`) and FFI (`callFfi`) sites |
 | `crossHeapStoreViolation` | `src/memory.zig` | kaappi#1924: the rejection predicate checked before every general store into a heap object — a store into a container owned by another GC is allowed only when the value is owned by that same GC. The `mutex-lock!` owner pair is the deliberate exception |
 | `Object.owner` vs `GC.id` | `src/gc_collect.zig` | Every heap object records its owning GC; marking skips objects owned by another GC, so a child's collections never write mark bits on parent-heap objects reached through shared globals (kaappi#958) |
+| `SharedChannel.identity_seed` | `src/shared_channel.zig` | kaappi#2394: the promoting channel's tagged Value, written before `ch.shared` publishes — the hash identity every stub of the channel shares for its whole life |
+| `ensureComparatorLibraryLoaded` | `src/primitives_fiber.zig`, called from `threadStartImpl` | kaappi#2394: best-effort pre-load of `(srfi 128)` on the spawning VM, so no child VM ever lazy-loads into the struct-copied registry (see the channel-identity section) |
 
 The deep copy happens at exactly three boundaries: the thunk closure at
 `thread-start!`, the result (or uncaught exception) at `thread-join!`, and a
--- a/docs/dev/thread-value-sharing.md
+++ b/docs/dev/thread-value-sharing.md
@@ -124,6 +124,58 @@ directory object, environment, ephemeron, guardian, or transport cell),
 or a channel owned by another thread
 ```
 
+## Channel identity across stubs (kaappi#2394)
+
+Every cross-thread hand-off of a channel mints a **stub** on the receiving
+heap (`gc_deep_copy.zig`'s `.channel` arm → `allocChannelStub`): a fresh
+`types.Channel` whose `shared` pointer targets the one refcounted
+`SharedChannel`. Receiving the same channel twice therefore yields two stubs
+that are *not* `eq?`/`eqv?`/`equal?` — distinct heap objects — even though no
+channel operation can tell them apart. KEP-0002 §2 once promised an `eqv?`
+that compares the `shared` pointer; the KEP's 2026-08-27 as-implemented
+amendment dropped that, and kaappi#2394 settled the promise as an explicit
+comparator surface in `(kaappi fibers)` instead — extending `eqv?` per-type
+is what SRFI-128 comparators exist for:
+
+- `channel=? a b` — `#t` iff both are channels backing the same
+  `SharedChannel` (or the same unpromoted local channel; the mixed case only
+  arises for genuinely different channels, since promotion rewrites the
+  original in place and every stub is born promoted).
+- `channel-hash ch [bound]` — hashes the `shared` pointer, consistent with
+  `channel=?` by construction; mirrors `(hash obj [bound])` (SRFI 69).
+- `channel-comparator` — `(make-comparator channel? channel=? #f
+  channel-hash)`, built by calling the **real** `(srfi 128)`
+  `make-comparator`. That library lazy-loads on first use (the `environment`
+  procedure's on-demand route; an `embedded_libraries` entry keeps it
+  loadable under `--sandbox` and on WASM, like `(srfi 181)`), so
+  `comparator?` and the field accessors agree with every hand-built
+  comparator, and the native `extractComparatorFns` bridge recognizes the
+  record. The three procedures are `(kaappi fibers)`'s pristine registry
+  exports, immune to top-level shadowing of those names.
+
+The reply-channel/registry idiom:
+
+```scheme
+(define registry (make-hash-table channel=? channel-hash)) ; or (channel-comparator)
+(hash-table-set! registry stub 'reply-to) ; any stub of the channel finds it
+```
+
+Two caveats:
+
+- **Hash stability across promotion.** A channel hashed before its first
+  cross-thread send hashes its local address; after promotion (sticky) the
+  `SharedChannel`'s. Key a channel into a table only after sharing it.
+- **`(srfi 113)` sets do not honor comparators in this port**: `113.sld`'s
+  `%make-empty-set` builds its backing table with a zero-argument
+  `make-hash-table` (default `equal?`), so `(set channel-comparator …)`
+  still dedups by pointer. Use `make-hash-table` (SRFI 69/125) directly for
+  channel-keyed tables.
+
+`channel=?` and `channel-hash` read channel fields, so they carry the same
+foreign-owner check as `channel-send` (a channel reached through a shared
+global is a diagnosis, not a comparison); `channel?` stays exempt as a total
+predicate.
+
 ## The globals route
 
 `VM.initForThread` shares the **root** VM's `globals` map **by pointer**,
@@ -137,7 +189,7 @@ Four types defend themselves at the primitive level, by comparing
 
 | type | check | sites |
 |---|---|---|
-| channel | `channel belongs to another thread; pass it through the thread thunk to share it` | `primitives_fiber.zig` — `channel-send`, `channel-receive`, `channel-close!`, `channel-closed?` |
+| channel | `channel belongs to another thread; pass it through the thread thunk to share it` | `primitives_fiber.zig` — `channel-send`, `channel-receive`, `channel-close!`, `channel-closed?`, `channel=?`, `channel-hash` |
 | thread handle | `thread belongs to another OS thread; a thread handle may only be used by the thread that created it` | `primitives_srfi18.zig` `checkThreadOwner` — `thread-name`, `thread-specific`, `thread-specific-set!`, `thread-start!`, `thread-terminate!`, `thread-join!` |
 | fiber | `fiber-join: fiber belongs to another thread; a fiber may only be joined by the thread that spawned it` | `primitives_fiber.zig` — `fiber-join` (kaappi#2001) |
 | guardian | `guardian belongs to another thread; a guardian may only be used by the thread that created it` | `vm_calls.zig` `invokeGuardian` — both the register and the retrieve call shapes, object and transport-cell guardians alike (kaappi#2008) |
@@ -366,6 +418,13 @@ duplicating: `src/tests_deepcopy.zig` asserts the port and continuation
 refusals at the `GC.deepCopy` level directly, and
 `src/tests_shared_channel.zig` asserts that a channel *message* carrying a
 port is refused the same way.
+
+Channel identity across stubs (previous section) has its own coverage:
+`tests/scheme/smoke/channel-identity-2394.scm` runs the
+two-stubs-one-channel scenario end to end (both `make-hash-table` flavors,
+the SRFI-128 comparator contract), and `src/tests_fibers.zig` carries the
+unit-level versions (local identity, stub unification, table dedup, lazy
+comparator construction).
 
 The two copied rows added above have their own regression suites, each
 covering all four boundaries (into the child, out of it, the raise path, a
--- a/docs/dev/thread-value-sharing.md
+++ b/docs/dev/thread-value-sharing.md
@@ -164,11 +164,15 @@ representation-of-the-moment: its own tagged Value while unpromoted, and —
 once promoted — the `SharedChannel.identity_seed` `promoteChannel` preserved
 from exactly that Value before publishing. A table keyed before the first
 cross-thread send therefore still finds its entry after promotion rewrites
-the original in place, and every stub hashes the one seed. If the original
-object is later collected and its address reused, the collision is benign —
-equality still compares `shared` pointers. (Unpromoted channels hash through
-the same `valueHash` as plain `(hash ch)`, so the two agree on every target,
-wasm32's 32-bit `usize` included.)
+the original in place, and every stub hashes the one seed. The seed is
+hashed with `identityHash` — a pure function of the bits that never
+dereferences them (the original object may since have been swept or live on
+another heap; `valueHash` would be wrong there, its dispatch reads `.tag`
+through pointer bits) — so if the original is later collected and its
+address reused, the collision is benign and deterministic: equality still
+compares `shared` pointers. For a live unpromoted channel the result equals
+plain `(hash ch)`'s identity arm on every target, wasm32's 32-bit `usize`
+included.
 
 `channel=?` and `channel-hash` read channel fields, so they carry the same
 foreign-owner check as `channel-send`, now factored as
@@ -188,16 +192,17 @@ backing table was built with a zero-argument `make-hash-table` (default
 **Loading `(srfi 128)` safely across threads.** `vm.libraries` reaches
 child VMs as a struct copy (unlike `globals`, which got the pointer+lock
 treatment — see `vm.zig`'s `initForThread`), and `markVmRoots` gates
-library marking behind `owns_globals`. A lazy library load from a worker
-thread would therefore put() into bucket storage the root reads unlocked
-and leave child-heap exports unmarked by every collector.
-`threadStartImpl` closes this by pre-loading `(srfi 128)` on the spawning
-VM before any child exists (`ensureComparatorLibraryLoaded` — best-effort;
-the process's first `make-thread` is necessarily the root, so the load is
-always pre-children and read-only afterwards), and an
-`embedded_libraries` entry keeps the library loadable under `--sandbox`,
-on WASM, and — as the last resort after a disk miss — on a binary with no
-lib tree to read at all.
+library marking behind `owns_globals`. A library load from a non-root
+VM would therefore put() into bucket storage the root reads unlocked
+and leave non-root exports unmarked by every collector.
+`threadStartImpl` closes this by pre-loading `(srfi 128)` on the root
+VM before any child exists (`ensureComparatorLibraryLoaded`, which
+refuses to load off-root; best-effort — the process's first
+`thread-start!` is necessarily the root, so the load is always
+pre-children, and a read-only no-op afterwards once it succeeded), and
+an `embedded_libraries` entry keeps the library loadable under
+`--sandbox`, on WASM, and — as the last resort after a disk miss — on a
+binary with no lib tree to read at all.
 
 ## The globals route
 
--- a/lib/srfi/113.sld
+++ b/lib/srfi/113.sld
@@ -70,10 +70,15 @@
     ;; The backing table uses the comparator's own equality/hash (kaappi#2394):
     ;; the native make-hash-table unwraps a <comparator> record into its two
     ;; procedures, and detectMode keeps the native fast path for the standard
-    ;; comparators, whose equality fields hold the real eq?/eqv?/equal?. Until
+    ;; comparators, whose equality fields hold the real eq?/eqv?/equal? --
+    ;; make-default-comparator included while nothing is registered via
+    ;; comparator-register-default! (the hashtable bridge recognizes that
+    ;; case natively). A user comparator with closure equality runs the
+    ;; generic custom mode: correct, with a procedure call per probe. Until
     ;; this fix the comparator was stored in the record but every membership
-    ;; test ran through the default equal? table, so any non-equal? comparator
-    ;; — (set (channel-comparator) a b) included — silently deduped by pointer.
+    ;; test ran through the default equal? table, so any non-equal?
+    ;; comparator -- (set (channel-comparator) a b) included -- silently
+    ;; deduped by pointer.
     (define (%make-empty-set comparator)
       (%make-set comparator (make-hash-table comparator)))
 
--- a/lib/srfi/113.sld
+++ b/lib/srfi/113.sld
@@ -67,11 +67,18 @@
 
     ;;; Internal helpers
 
+    ;; The backing table uses the comparator's own equality/hash (kaappi#2394):
+    ;; the native make-hash-table unwraps a <comparator> record into its two
+    ;; procedures, and detectMode keeps the native fast path for the standard
+    ;; comparators, whose equality fields hold the real eq?/eqv?/equal?. Until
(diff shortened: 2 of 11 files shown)


src/primitives_fiber.zig
```diff
@@ -1425,6 +1433,138 @@ fn channelClosedFn(args: []const Value) PrimitiveError!Value {
...
+        gc.pushRoot(&srfi_sym);
+        defer gc.popRoot();
+        // allocPair auto-roots its Value args, so the tail needs no separate
+        // root before it is handed to the outer allocPair (the rooting shape
+        // of vm_library.zig's buildSrfiNameList).
+        const tail = gc.allocPair(types.makeFixnum(128), types.NIL) catch return PrimitiveError.OutOfMemory;
+        var iset = gc.allocPair(srfi_sym, tail) catch return PrimitiveError.OutOfMemory;
+        gc.pushRoot(&iset);
+        defer gc.popRoot();
+        vm_imports.ensureLibraryLoaded(vm, iset, "srfi.128") catch {
```
baijum: **[correctness] Lazy `(srfi 128)` load from a worker thread races the struct-copied library registry.**

`vm.libraries` reaches SRFI-18 child VMs as an unlocked **struct copy** (`vm.zig:774` `.libraries = parent.libraries` — unlike `globals`, which got the heap-pointer + `globals_lock` treatment), and `LibraryRegistry.register` takes no lock. So when a worker thread is the first to call `(channel-comparator)` — the PR's own headline idiom is `(make-hash-table (channel-comparator))` inside a worker — this `ensureLibraryLoaded` does an unsynchronized `StringHashMap.put` into bucket storage shared with the parent: a data race, and a child-side rehash frees the bucket array the parent still reads (use-after-free on the parent's next import). Even without a rehash, the registration is invisible to the parent, so a later parent-side load mints a *second* `<comparator>` record type and `comparator?` answers `#f` for comparators built on the other thread.

This is the pre-existing gap documented at `vm.zig:387-390` (reachable via `environment`/eval-import), but this primitive turns worker-thread-first loading from exotic into the expected case. Worth either eagerly loading `(srfi 128)` on the root VM (e.g. at fibers registration or first `make-thread`), routing the lazy load through the root VM, or at least giving `libraries` the #958-style pointer+lock treatment before shipping this surface.

src/primitives_fiber.zig
```diff
@@ -1425,6 +1433,138 @@ fn channelClosedFn(args: []const Value) PrimitiveError!Value {
...
+/// top-level shadowing of those names. (srfi 128) loads on demand — the
+/// `environment` procedure's own lazy route — from the embedded copy under
+/// --sandbox and on WASM.
+fn channelComparatorFn(_: []const Value) PrimitiveError!Value {
+    const vm = vm_mod.vm_instance orelse return PrimitiveError.InvalidBytecode; // no VM: internal invariant
+    const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
+
+    // Load (srfi 128) before fetching any Values that must survive a GC the
+    // load can trigger: library loading runs arbitrary begin-block code.
+    if (vm.libraries.get("srfi.128") == null) {
```
baijum: **[correctness] A child-VM lazy load's `(srfi 128)` exports are never GC-marked → use-after-free on the second call.**

`markVmRoots` gates library marking behind `if (vm.owns_globals)` (`vm.zig:276`), and cross-heap marking skips foreign-owned objects (`gc_collect.zig:975`). So when a worker thread performs this lazy load, the `make-comparator` closure and `<comparator>` record type land on the **child heap**, reachable only through the child's *unmarked* registry copy once this function's pushRoots pop. Sequence: worker calls `(channel-comparator)`, allocates enough to trigger a child-heap collection (the loaded library gets swept), calls it again — this fast-path guard sees the registry entry, fetches the swept `make-comparator` Value, and `callWithArgs` executes a freed Function. Panic under `-Duaf-detection`; silent UAF in release.

Related nit: the rooting comment at L1549-1550 ("otherwise reachable only through this frame's locals") is factually wrong on the root VM — `lib.exports` is marked on every collection there — while the rooting is genuinely load-bearing on child VMs *only because of this marking gap*. If the gap gets closed (non-owner VMs marking self-owned registry values), the four-root ceremony becomes removable; until then it must stay, so the comment should state the real reason.

src/vm_library.zig
```diff
@@ -389,6 +389,11 @@ const embedded_libraries = [_]struct { rel_path: []const u8, source: []const u8
     // .srfi_181_primitives) -- without this entry, --sandbox and WASM would
     // silently lose access to the already-shipped custom-port constructors.
     .{ .rel_path = "srfi/181.sld", .source = @import("kaappi_srfi_181_sld").source },
+    // (srfi 128): (channel-comparator) (kaappi#2394) builds the channel
+    // identity comparator through the real make-comparator, on demand, so
+    // the library must stay loadable where file loads are blocked (--sandbox)
+    // or unreliable (WASM).
+    .{ .rel_path = "srfi/128.sld", .source = @import("kaappi_srfi_128_sld").source },
```
baijum, line 399: **[correctness] The embedded `128.sld` is never consulted on native non-sandbox builds, so `(channel-comparator)` still requires a disk `lib/` tree.**

`findEmbeddedLibrary` is only consulted under `sandbox_mode` (L498-501) and `is_wasm` (L508-510); the native path ends at `resolveLibraryPath(...) orelse return error.UndefinedVariable` with no embedded fallback. Two reachable failures:

1. A release binary copied to `/usr/local/bin` on a machine with no exe-relative `lib/`, no `~/.kaappi/lib`, no `--lib-path`, and no cwd `lib/`: `(channel-comparator)` raises "failed to load (srfi 128)" even though the identical source is compiled into that very binary.
2. A `-Dbundle-src` standalone whose app never imported `(srfi 128)` at bundle time (`bundled_files` records only files read during the compile run): the same failure on a lib-tree-less deployment machine — the standalone binary's primary use case.

A fallthrough to `findEmbeddedLibrary` after the `resolveLibraryPath` miss keeps disk-first precedence (the documented native trade-off) and closes both. If the disk-only behavior is intentional for the comparator too, the doc's "the library must stay loadable" comment here oversells what the embed achieves on native.

docs/dev/thread-value-sharing.md
```diff
@@ -124,6 +124,58 @@ directory object, environment, ephemeron, guardian, or transport cell),
...
+The reply-channel/registry idiom:
+
+```scheme
+(define registry (make-hash-table channel=? channel-hash)) ; or (channel-comparator)
+(hash-table-set! registry stub 'reply-to) ; any stub of the channel finds it
+```
+
+Two caveats:
+
+- **Hash stability across promotion.** A channel hashed before its first
```
baijum: **[correctness] The promotion caveat undersells the failure: promotion retroactively and silently corrupts the *sender's own* tables — and a cheap fix exists.**

Promotion rewrites the sender's channel in place (`shared_channel.zig:484` sets `ch.shared` on the existing object), flipping `channel-hash`'s branch. An entry keyed *before* the first cross-thread send now lives in the old-address bucket while lookups probe the SharedChannel-pointer bucket: `hash-table-ref` misses silently, and a subsequent `hash-table-set!` inserts a duplicate "equal" key. "Key a channel only after sharing it" reads as advice for a stable-but-different hash; the actual consequence is retroactive lookup misses in a table you already built, with no error. At minimum this paragraph should name that consequence.

But the mechanism can absorb it: `promoteChannel` already computes the pre-promotion tagged Value (`const ch_val = types.makePointer(&ch.header)` in the step-4 migration), which is exactly what the unpromoted branch hashes. Storing it as a `SharedChannel.identity_seed` and having `channel-hash` hash the seed when promoted makes the hash stable for the channel's whole life — `channel=?` untouched, stubs consistent (they all hash the one seed), address-reuse collisions benign since equality is by `shared` pointer. If this was considered and rejected, the doc should record why; otherwise it looks like the better fix.

src/primitives_fiber.zig
```diff
@@ -1425,6 +1433,138 @@ fn channelClosedFn(args: []const Value) PrimitiveError!Value {
...
+        return raiseFiberError("channel belongs to another thread; pass it through the thread thunk to share it");
+
+    const ch = ch_obj.as(types.Channel);
+    // Promoted: the one pointer every stub of this channel shares. Unpromoted:
+    // the boxed value itself, so an unshared channel hashes identically here
+    // and under plain (hash ch). Same mixing constant as identityHash.
+    const h: u64 = if (ch.shared) |raw|
+        @as(u64, @intFromPtr(raw)) *% 2654435761
+    else
+        args[0] *% 2654435761;
```
baijum: **[correctness] The claimed invariant "an unshared channel hashes identically here and under plain `(hash ch)`" is false on wasm32.**

`valueHash` returns `usize` and its identity fallback is `@truncate(key *% 2654435761)` — on wasm32 that keeps only the low 32 bits, and SRFI-69 `hashFn` re-widens with zero high bits before masking. This branch keeps the full 64-bit product masked to 47 bits, so on the WASM build `(= (channel-hash ch) (hash ch))` is `#f` for essentially every heap address (bits 32..46 differ). The smoke assertion "an unshared channel hashes like plain (hash ch)" would fail there — it's currently masked only because the smoke file imports `(srfi 18)`, which is unavailable on wasm, so `run-wasm-differential.sh` drops the whole file into the LIBDIFF exclusion bucket.

Suggested fix: route the unpromoted branch through `primitives_hashtable.valueHash(args[0])` (already `pub`, already imported) so both sides compute the same value on every target. That also removes the copied mixing constant — right now the invariant is enforced only by two literals staying in sync.

src/primitives_fiber.zig
```diff
@@ -1425,6 +1433,138 @@ fn channelClosedFn(args: []const Value) PrimitiveError!Value {
...
+        defer gc.popRoot();
+        // allocPair auto-roots its Value args, so the tail needs no separate
+        // root before it is handed to the outer allocPair (the rooting shape
+        // of vm_library.zig's buildSrfiNameList).
+        const tail = gc.allocPair(types.makeFixnum(128), types.NIL) catch return PrimitiveError.OutOfMemory;
+        var iset = gc.allocPair(srfi_sym, tail) catch return PrimitiveError.OutOfMemory;
+        gc.pushRoot(&iset);
+        defer gc.popRoot();
+        vm_imports.ensureLibraryLoaded(vm, iset, "srfi.128") catch {
+            if (vm.last_error_detail_len == 0)
```
baijum: **[correctness] Stale `last_error_detail` misroutes this branch: reset it before the load, per the `ffi.zig:473` precedent.**

`last_error_detail` is only cleared at eval/pipeline boundaries and on *some* native dispatch paths — not `.call_global`/`.tail_call_global` (non-profile), which is how a global like `channel-comparator` is invoked — and guard-catching an error doesn't clear it. Meanwhile the loader's own `setErrorDetail` sites are all guarded by `if (vm.last_error_detail_len == 0)`, so they never overwrite a stale value, and the plain miss path (`resolveLibraryPath ... orelse return error.UndefinedVariable`) writes nothing at all.

Concretely: `(begin (guard (e (#t 'ok)) (car 5)) (channel-comparator))` with `(srfi 128)` unresolvable reports **"type error in 'car': expected pair, got 5"** as the reason the library failed to load — and the `// bare-ok: the loader's own detail` comment is false on that path. Fix: `vm.last_error_detail_len = 0;` immediately before the `ensureLibraryLoaded` call, exactly as `callFfi` does (`ffi.zig:473`).

docs/dev/thread-value-sharing.md
```diff
@@ -124,6 +124,58 @@ directory object, environment, ephemeron, guardian, or transport cell),
...
+  cross-thread send hashes its local address; after promotion (sticky) the
+  `SharedChannel`'s. Key a channel into a table only after sharing it.
+- **`(srfi 113)` sets do not honor comparators in this port**: `113.sld`'s
```
baijum: **[conventions] This caveat documents a live SRFI-113 spec bug with no tracker issue — and the fix looks like two one-liners.**

`docs/dev/CLAUDE.md`: "Open bugs belong in the issue tracker, not here." `%make-empty-set`/`%make-empty-bag` ignoring their **required** comparator argument is a spec violation for every non-`equal?` comparator, not just channels — `(set (channel-comparator) a b)`, the exact idiom this PR ships a comparator for, silently dedups by pointer with the only warning living in a dev doc.

The stated reason for not fixing it ("routing its tables through comparator lambdas would drop every SRFI-113 set off the native `equal?` fast path") doesn't hold up against the code: `113.sld` already imports `(srfi 128)`, the native bridge accepts comparator records directly (this PR's own test proves `(make-hash-table (channel-comparator))` works), and `configureHashTable` runs `detectMode` on the *extracted* equality — so a comparator whose equality is the real `equal?` keeps the fast mode automatically, no gating needed. `set-copy` already preserves table modes, and every fresh table routes through these two constructors. So the fix is `(make-hash-table comparator)` at 113.sld:71 and :74, plus a set-dedup test — either do that (and delete this caveat) or file the issue and link it here.

src/primitives_fiber.zig
```diff
@@ -1425,6 +1433,138 @@ fn channelClosedFn(args: []const Value) PrimitiveError!Value {
...
+/// (channel-comparator) — (make-comparator channel? channel=? #f
+/// channel-hash) built by the real (srfi 128), so `comparator?` and the
+/// field accessors agree with every other comparator in the image and
+/// (make-hash-table channel-comparator) works through the native comparator
+/// bridge (kaappi#2394, KEP-0002 §2). The three procedures come from
+/// (kaappi fibers)'s pristine registry exports, immune to any later
+/// top-level shadowing of those names. (srfi 128) loads on demand — the
+/// `environment` procedure's own lazy route — from the embedded copy under
+/// --sandbox and on WASM.
+fn channelComparatorFn(_: []const Value) PrimitiveError!Value {
```
baijum, line 1537: **[efficiency] Consider caching the comparator — it's a per-VM constant rebuilt on every call.**

Each `(channel-comparator)` call does 2 library lookups + 4 export-map string lookups + 4 root pushes + a bytecode call into the Scheme `make-comparator` closure + a fresh `<comparator>` record allocation. The codebase already has the exact pattern for this: `default_random_source` (`vm.zig:572`), lazily built, cached on the VM, marked in `markVmRoots` (`vm.zig:252`), lazy-init in `primitives_random.zig`. The only observable difference is `(eq? (channel-comparator) (channel-comparator))` flipping `#f`→`#t`, which SRFI-128 permits and no test relies on.

Bonus: a root-VM-cached, GC-marked singleton also shrinks exposure to the child-thread lazy-load hazards flagged above whenever the root VM touches it first.

src/primitives_fiber.zig
```diff
@@ -1425,6 +1433,138 @@ fn channelClosedFn(args: []const Value) PrimitiveError!Value {
...
+/// dereferences the channel, so it carries the foreign-owner check rather
+/// than channel?'s total-predicate convention.
+fn channelEqFn(args: []const Value) PrimitiveError!Value {
+    if (!types.isChannel(args[0]))
+        return primitives.typeError("channel=?", "channel", args[0]);
+    if (!types.isChannel(args[1]))
+        return primitives.typeError("channel=?", "channel", args[1]);
+
+    const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
+    for (args) |a| {
```
baijum: **[reuse] 5th and 6th verbatim copies of the foreign-owner check — time to extract `checkChannelOwner`.**

`channelEqFn` and `channelHashFn` bring this file to six inline copies of `owner != gc.id` + the same "channel belongs to another thread…" string (lines 182, 1087, 1399, 1426, 1455, 1482). The sibling type already models the right shape: `primitives_srfi18.zig:498` `checkThreadOwner(val, proc)`, used by six thread primitives. The string is load-bearing beyond this file (asserted in `tests_shared_channel.zig`, quoted in the `thread-value-sharing.md` table), so a wording change today means six coordinated edits.

Minor, same site: the `for (args) |a|` loop computes `types.toObject(a)` for both args and then L1458-1459 re-derive `toObject(args[0])`/`toObject(args[1])` — every other owner-check site in the file binds the object once and reuses it for both the check and the `.as()` cast.

src/primitives_fiber.zig
```diff
@@ -1425,6 +1433,138 @@ fn channelClosedFn(args: []const Value) PrimitiveError!Value {
...
+    // Load (srfi 128) before fetching any Values that must survive a GC the
+    // load can trigger: library loading runs arbitrary begin-block code.
+    if (vm.libraries.get("srfi.128") == null) {
+        var srfi_sym = gc.allocSymbol("srfi") catch return PrimitiveError.OutOfMemory;
+        gc.pushRoot(&srfi_sym);
+        defer gc.popRoot();
+        // allocPair auto-roots its Value args, so the tail needs no separate
+        // root before it is handed to the outer allocPair (the rooting shape
+        // of vm_library.zig's buildSrfiNameList).
+        const tail = gc.allocPair(types.makeFixnum(128), types.NIL) catch return PrimitiveError.OutOfMemory;
```
baijum: **[reuse] This block hand-duplicates `buildSrfiNameList` — the comment even cites it.**

`vm_library.zig:198` `buildSrfiNameList(gc, n)` builds exactly this list with exactly this rooting shape (the comment above literally references "the rooting shape of vm_library.zig's buildSrfiNameList"). Making it `pub` and calling it costs nothing — no import cycle (`vm_library.zig` doesn't import `primitives_fiber.zig`) — and means a future change to `allocPair`'s auto-rooting contract gets fixed in one place instead of leaving this copy as a GC-window bug that only fires when the lazy load triggers a collection.

Two smaller dedups in the same block: the outer `if (vm.libraries.get("srfi.128") == null)` duplicates `ensureLibraryLoaded`'s own first-line early return (calling it unconditionally leaves one guard; the warm-path cost is just the name-list allocation, which the helper call makes trivial), and "channel-comparator: failed to load (srfi 128)" appears twice (L1530, L1536).

baijum:
Thanks — every finding verified against the code before fixing, all addressed in a8d66a2d.

**Promotion-stable hash (CodeRabbit + the identity_seed suggestion).** Took the suggested mechanism: `promoteChannel` preserves the promoting channel's tagged Value as `SharedChannel.identity_seed` (written before `ch.shared` publishes `sc`), and `channel-hash` hashes that seed in every representation. A table keyed before the first cross-thread send stays reachable after promotion, and all stubs hash the one seed. The unpromoted branch also routes through `valueHash` now, which fixes the wasm32 invariant claim **and** deletes the copied mixing constant — `(= (channel-hash ch) (hash ch))` holds by construction on every target. Added the insert → promote → lookup regression in both suites.

**Worker-thread lazy-load hazards (registry race + unmarked exports).** Went with the eager root-side option: `threadStartImpl` calls `ensureComparatorLibraryLoaded(vm)` before spawning — the process's first `make-thread` is necessarily the root (children exist only inside this function), so the load always happens pre-children and is a read-only no-op for every later spawn, whose registry copy postdates it. No child ever puts into the shared bucket storage, and the exports land on the root heap where `markVmRoots` marks them. Best-effort by design; the lazy path remains as the error-reporting fallback. Verified worker-thread-first `(channel-comparator)` end to end (child builds its own cached comparator; root's still works after the join).

**Comparator caching.** `vm.default_channel_comparator`, marked unconditionally next to `default_random_source` (markValue skips foreign-heap values, so the unconditional mark is safe for a child VM too). `(eq? (channel-comparator) (channel-comparator))` is now `#t` — pinned by a test.

**Embedded fallback on native disk-miss.** `tryLoadLibraryFromFile` now falls through to `findEmbeddedLibrary` after `resolveLibraryPath` misses — covers both the no-lib-tree release binary and the `-Dbundle-src` standalone whose bundle never recorded the file. Disk keeps precedence. Verified by copying the binary to a scratch dir with an empty `KAAPPI_HOME`: `(make-hash-table (channel-comparator))` works with no lib tree in sight.

**Stale `last_error_detail`.** Reset before the load (ffi.zig:473 precedent), and the swallow path clears again so a failed load's detail can't leak into the next unrelated error.

**`checkChannelOwner` / `buildSrfiNameList`.** Both extracted; the six inline owner checks became helper call sites (channel=? now binds each operand once), and the hand-rolled `(srfi 128)` name list is `vm_library.buildSrfiNameList(gc, 128)`. The double "failed to load" message and the redundant outer presence check are gone (the cache means the whole body runs once per VM anyway).

**SRFI-113.** Took the fix rather than the issue: `%make-empty-set`/`%make-empty-bag` pass the comparator to `make-hash-table`, your fast-path analysis confirmed (`detectMode` sees the extracted equality, and the standard comparators' equality fields hold the real `eq?`/`eqv?`/`equal?` — checked against 128.sld:216-218). `(set (channel-comparator) a b ch)` → size 1; both SRFI-113 suites green (133 + 246), full run-all 2116/0. The caveat paragraph in thread-value-sharing.md now documents the fix instead of the gap.

**CodeRabbit's TestContext nits.** Both multi-step tests converted.

One fix the review didn't predict: the first `u64` field in `SharedChannel` broke the wasm build — on wasm32-baseline a natural `u64` demands align 8 against `Header`'s 4, and `destroyHook`'s `@fieldParentPtr` refuses to widen. `identity_seed` carries an explicit header-matching alignment.

src/primitives_fiber.zig
```diff
@@ -1425,6 +1415,169 @@ fn channelClosedFn(args: []const Value) PrimitiveError!Value {
...
+    // therefore still finds its entry afterwards, and every stub of the
+    // channel hashes the one seed. Routed through valueHash, not a local
+    // mixing constant, so an unpromoted channel hashes identically here
+    // and under plain (hash ch) on every target — wasm32's 32-bit usize
+    // truncation included.
+    const seed: Value = if (ch.shared) |raw| blk: {
+        const sc: *shared_channel.SharedChannel = @ptrCast(@alignCast(raw));
+        break :blk @bitCast(sc.identity_seed);
+    } else args[0];
+    const h: u64 = primitives_hashtable.valueHash(seed);
```
baijum: **[correctness] `valueHash(seed)` dereferences the seed — a racy foreign read now, a use-after-free once the original is collected.**

`identity_seed` is the pre-promotion pointer-tagged Value, and the `SharedChannel` doc correctly says it's "never dereferenced — a hash input only"… but `valueHash` *does* dereference it: the dispatch chain (`isString`, `isSymbol`, `isPair`, `isVector`, …) reads `toObject(v).tag` — ~10 header reads — before a channel Value can fall through to the identity arm. Nothing keeps the original channel object alive for the SharedChannel's lifetime (the refcount flows the other way: stubs retain the SharedChannel; when the sender drops its references, its GC sweeps the original while receiver stubs live on). Your own comment anticipates address reuse ("the collision is benign") — which concedes the original can be dead. Three grades of failure on a receiver-thread `channel-hash`: cross-thread racy header read while the original lives; freed-memory read after it's swept; and if the address is reused by, say, a string, `valueHash` *content-hashes the unrelated object*, so the "stable for its whole life" hash silently changes as that object mutates — breaking tables again, this time non-deterministically.

The seed design itself is sound — only the routing is wrong. A live channel matches no arm and falls through to `@truncate(key *% 2654435761)`, a pure function of the bits. Expose that as `pub fn identityHash(bits: u64) usize` in `primitives_hashtable.zig` and compute `identityHash(if (ch.shared) sc.identity_seed else args[0])` — bit-identical to today's unpromoted result on every target (wasm32's `usize` truncate included), zero dereference of the seed.

lib/srfi/113.sld
```diff
@@ -67,11 +67,18 @@
...
+    ;; The backing table uses the comparator's own equality/hash (kaappi#2394):
+    ;; the native make-hash-table unwraps a <comparator> record into its two
+    ;; procedures, and detectMode keeps the native fast path for the standard
+    ;; comparators, whose equality fields hold the real eq?/eqv?/equal?. Until
+    ;; this fix the comparator was stored in the record but every membership
+    ;; test ran through the default equal? table, so any non-equal? comparator
+    ;; — (set (channel-comparator) a b) included — silently deduped by pointer.
     (define (%make-empty-set comparator)
-      (%make-set comparator (make-hash-table)))
+      (%make-set comparator (make-hash-table comparator)))
```
baijum, line 83: **[efficiency] The fast-path claim doesn't hold for `make-default-comparator` — the common portable case just got slower.**

The comment (and commit message) say the standard comparators' "equality fields hold the real eq?/eqv?/equal?" — true for `make-eq-comparator`/`make-eqv-comparator`/`make-equal-comparator`, but **not** `make-default-comparator`: its equality is `default-equality`, a Scheme closure (necessarily — SRFI 128 requires it to consult `comparator-register-default!`). `detectMode` recognizes fast modes via `isNativeFn` + name match, so a closure lands in `.custom` — and `extractComparatorFns` installs `default-hash` (also a closure) as the hash. Net: every `(set (make-default-comparator) …)` — previously a native `equal?` table — now pays **two** `vm.callWithArgs` bridges per probe (equality *and* hash), with unchanged observable behavior whenever nothing is registered, which is the near-universal case since `default-equality` falls back to `equal?`.

Worth either a mitigation (teach `extractComparatorFns`/`detectMode` to recognize the default comparator while the registry is empty and keep `.equal` mode) or at least correcting this comment so the regression is a recorded trade-off rather than an overclaim.

src/primitives_fiber.zig
```diff
@@ -1425,6 +1415,169 @@ fn channelClosedFn(args: []const Value) PrimitiveError!Value {
...
+/// globals, which got the pointer+lock treatment — so a lazy load from a
+/// worker thread would put() into bucket storage the parent reads
+/// unlocked, and a child-side rehash would free buckets the parent's copy
+/// still names; the child-side exports would also be unmarked by every
+/// collector (markVmRoots gates library marking behind owns_globals).
+/// Loading at the first make-thread is always pre-children — children
+/// exist only inside threadStartImpl — and read-only afterwards. Failures
+/// are swallowed: channel-comparator's own lazy load then reports the
+/// described error instead.
+pub fn ensureComparatorLibraryLoaded(vm: *vm_mod.VM) void {
```
baijum: **[correctness] The worker-side lazy load is still reachable — consider a root-VM guard, plus two comment fixes.**

The pre-load is best-effort, and its failure path is reachable without OOM: the embedded fallback in `tryLoadLibraryFromFile` triggers only when `resolveLibraryPath` returns *null* — a **found but broken** `128.sld` (corrupted, stale, or shadowed by a user `--lib-path`) fails via `readWholeFile catch return error.UndefinedVariable` with no embedded fallthrough, and the failure is silently swallowed here. After that, a worker's `(channel-comparator)` runs this function on the *child* VM and performs exactly the lazy load the pre-load exists to prevent (registry race + unmarked child-heap exports); a child spawning a grandchild re-runs it from `threadStartImpl` too (it passes `vm`, not `root_vm`). A `if (vm.root_vm != null) return;`-style guard (skip on non-root; let `channel-comparator` report the load failure) closes the residual path cheaply — or extend the embedded fallback to the read-failure case so the root pre-load can't fail short of OOM.

Two comment nits in the same area: the hook runs at `thread-start!` (`threadStartImpl`), not `make-thread` as the comments and doc say; and "a worker thread never performs this load" / "read-only no-op for every later one" are only true when the root load succeeded.

src/primitives_fiber.zig
```diff
@@ -1425,6 +1415,169 @@ fn channelClosedFn(args: []const Value) PrimitiveError!Value {
...
+    var iset = vm_library.buildSrfiNameList(gc, 128) catch return;
+    gc.pushRoot(&iset);
+    defer gc.popRoot();
+    // Clear-then-clear-again: the loader's setErrorDetail sites only write
+    // when no detail is set, so a stale one would both suppress the
+    // loader's own message and — once swallowed here — mis-attach to the
+    // next unrelated error (the ffi.zig callFfi precedent for the reset).
+    vm.last_error_detail_len = 0;
+    vm_imports.ensureLibraryLoaded(vm, iset, "srfi.128") catch {
+        vm.last_error_detail_len = 0;
```
baijum: **[minor] The swallow now discards the loader's diagnostics entirely.**

Previously a load failure could propagate the loader's specific detail ("ParseError while loading library from …/srfi/128.sld"). Now this catch zeroes `last_error_detail_len`, so `channel-comparator`'s fallback always raises the generic "failed to load (srfi 128)" — a user with a broken `128.sld` gets no hint which file or error is responsible. The clear-on-catch is right (stale detail must not mis-attach to the next error), but copy the detail into the raised message (or log it) before clearing rather than dropping it.

baijum:
Round 2 addressed in 9b8b2f87 — again, every finding verified against the code first, and all four were real.

**`valueHash(seed)` dereferencing the seed.** Completely right, and a sharper catch than my "never dereferenced" comment deserved — `isString`/`isSymbol`/… all read `.tag` through the bits, so the seed was a racy foreign read while the original lived, a freed-memory read after, and content-hashing of whatever reused the address. `channel-hash` now feeds the bits to `primitives_hashtable.identityHash` (made `pub`, param `u64`): the pure fall-through arm, no dereference, bit-identical to `valueHash` for a live channel on every target (wasm32 included). The doc's "collision is benign and deterministic" claim is now actually true.

**`make-default-comparator` regression.** Took the mitigation rather than the comment: the hashtable bridge recognizes an *unregistered* default comparator — equality is the `default-equality` binding from `(srfi 128)`'s lib_env and `registered-comparators` is `'()` — and installs the native `equal?`/`hash` pair, restoring the `.equal` fast path for the common case. A non-empty registry opts out (late registration changes the default comparator's meaning; a table capturing its construction-time behavior is SRFI-128-permissible), and any rename of the library internals degrades silently to `.custom` — slower, never wrong. Smoke tests pin both directions: equal?-table behavior unregistered, and a *registered* `channel-comparator` extending the default comparator (channel stubs unify through it) after `comparator-register-default!`.

**Residual off-root load.** `ensureComparatorLibraryLoaded` now has the root guard (`vm.root_vm != null` → refuse) and returns success, so no VM ever loads `(srfi 128)` off-root — a child's `channel-comparator` gets the described error instead of racing the registry. The broken-disk-file hole (found-but-unloadable `128.sld`) is verified end to end with a `--lib-path` shadow: the root reports `channel-comparator: failed to load (srfi 128): LibrarySourceReadError while loading library from …/128.sld`, the child reports the clean generic message, and the next unrelated error stays uncontaminated. I left the embedded fallback at *path-miss only* — substituting the embedded copy for a present-but-broken user file felt like masking user edits; with the root guard the failure mode is now a described error rather than a hazard, so the stricter choice costs nothing.

**Detail folding.** As above — the loader's diagnosis is folded into the raised message (the broken-file run shows the file and error verbatim), then cleared so it cannot attach to the next error.

Comment nits fixed too (`thread-start!`, not `make-thread`; the "read-only no-op" claim now qualified on the root load having succeeded). Verification: unit 1873/7-skipped, fibers green under gc-stress, SRFI-113 (133) and SRFI-128 (94) suites green, full `run-all.sh` **2116/0**, wasm builds, fmt + markdownlint clean.

Issue #2388, opened 2026-08-27
[SRFI 211] compare is name-based — implement binding-aware free-identifier=? or ratify the deviation; land the four-quadrant test


baijum:
**Context**

`er-macro-transformer` shipped in v0.22.0 (#1811). [KEP-0006](https://github.com/kaappi/keps/blob/main/keps/0006-explicit-renaming-macros.md) (amended 2026-08-27, keps#47) resolved Unresolved question 2 against SRFI 211's contract: `compare` is `free-identifier=?` — "these two identifiers denote the same binding, or both are unbound." The shipped `compare` is not that.

**Current behavior**

`erCompareFn` (`src/expander.zig:530`) is hygiene-stripped *name* equality: `stripHygienicPrefix` on both sides, then `std.mem.eql`. It answers the classic auxiliary-keyword checks (`else`, `=>`) but cannot distinguish two same-named bindings, so a use-site local rebinding of a keyword still compares true:

```scheme
(let ((else 1)) (er-is-else else))  ; => #t — the keyword "shadowing" is invisible
```

That limit is self-documented (`lib/srfi/211/explicit-renaming.sld:43`) and pinned in the audit suite (`tests/scheme/audit/primitives_smallbatch-audit.scm:761`) precisely so a stronger `compare` shows up as a test change.

**Decision needed**

- **(a) Implement binding-aware `compare`** — reuse the literal-matching machinery `syntax-rules` already has: `literal_bound` (definition-site binding slot) + `use_check.resolve` (use-site slot) in `matchPattern`'s symbol-literal branch (`src/expander.zig:565–606`). KEP-0006's Reference-level design sketches this as "the same machinery, exposed a second way."
- **(b) Ratify name-based `compare` as a documented deviation** (KEP-0004 discoverable-deviations spirit) and keep the audit pin as the contract.

**Either way: the four-quadrant acceptance test**

KEP-0006 defined it and it was never written. Against an er-macro `cond`-style transformer:

1. `(cond (else X))` — else clause fires (positive detection).
2. `(let ((else 1)) (cond (else X)))` — must **not** fire (keyword shadowed). *Fails today.*
3. An outer macro that *introduces* `else` into a `cond` it emits still fires, by hygiene.
4. The `=>` variants — including the shadowed-`=>` analogue of (2), which fails the same way (2) does under name-based equality.

**Evidence to inform the decision**

[...]

PR #2398, opened 2026-08-28
Re-port SRFI 241 and 202 on er-macro-transformer (KEP-0006 step 5)

### LINKED ISSUE #2391: [SRFI 241/202] Re-port match and and-let* on er-macro-transformer (KEP-0006 step 5 — the acceptance test)
**Context**

These two libraries were the *motivation* for `er-macro-transformer` (KEP-0006), and re-porting them was its designated acceptance test (Implementation-plan step 5): "if the documented limitations can now be lifted, that is the concrete, falsifiable signal that this delivered what the KEP set out to fix."

Both files have had **zero commits since `7949e497` (2026-07-21)** — the pre-implementation snapshot the KEP was written against. Every workaround the KEP's Motivation catalogued is still standing in `lib/srfi/241.sld`'s header:

- a **custom ellipsis identifier (`%%%`)** in every helper macro, purely so the literal `...` token can be pattern-matched as ordinary data;
- the SRFI's **ellipsis-aware `quasiquote** inside `match` bodies is not provided;
- **ellipsis-repeated sub-patterns** restricted to `,var`, `,_`, and repeated `,(var)`;
- **vector patterns**: fixed-length and whole-vector-ellipsis only — no mixed mandatory prefix/suffix around an ellipsis.

`lib/srfi/202.sld` is likewise still the self-contained pure-`syntax-rules` port.

**Deliverable**

Re-port one or both on `er-macro-transformer` (`(srfi 211 explicit-renaming)`): ordinary list processing replaces the template gymnastics, and the four limitations above are the lift targets. Keep the shipped reference-test suites green (`tests/scheme/srfi/`).

**Second purpose: evidence for #2388**

These macros check pattern literals (`_`, `...`, `->`) through `compare`, which is currently name-based (#2388). The re-port is the real-world exercise that decides whether name-based `compare` suffices in practice or whether binding-aware comparison is required — port first, then make the #2388 call on the evidence.

**Also fold in**

`lib/srfi/148.sld:10–14` still asserts Kaappi "has no er-macro-transformer support (that mechanism is SRFI 72's, a much larger, deferred undertaking)" — false since v0.22.0. Fix the header comment (the code itself correctly takes the reference's portable branch).

**References**

- KEP-0006 as amended: Motivation, Implementation-plan step 5 annotation, and "As implemented → Remaining work" (keps#47)
- Prior art: SRFI 150's `define-record-type` is an er-macro-transformer (`lib/srfi/150.sld:319`) — the one shipped consumer so far

baijum:
KEP-0006 step 5 — the designated acceptance test for `er-macro-transformer`. Both libraries are re-ported from pure `syntax-rules` onto `(srfi 211 explicit-renaming)`: each is now a single procedural transformer that compiles the pattern language by ordinary list processing, replacing the `%%%` custom-ellipsis template gymnastics wholesale.

## Limitations lifted (all four)

1. **Arbitrary ellipsis sub-patterns** — `(subpat ...)` now accepts any pattern: compound lists, dotted pairs, vectors, literals, nested ellipses, and catas, e.g. `(((,a ,b) ...)`, `((,x ...) ...)`, `((,(f -> b)) ...)`. The old port allowed only `,var`, `,_`, and repeated `,(var)`.
2. **Mandatory patterns after the ellipsis** — `(,x ... ,y)`, `(,x ... ,y . ,rest)`, and the full vector grammar `#(p1 ... pk pe ellipsis pm+1 ... pn)` with mandatory prefix and suffix. The split is deterministic (count the fixed patterns after the ellipsis), no backtracking.
3. **Ellipsis-aware quasiquote in clause bodies** — each clause body is wrapped in `(let-syntax ((quasiquote %match-qq)) ...)`; a subtemplate followed by ellipses iterates its unquoted expressions the way a syntax-rules template iterates pattern variables (`` `((,a . ,b) ...) ``, `` `(,x ... ...) ``, the spec's `` `(a ,(+ 1 2) ,(map abs '(4 -5 6)) ... b) `` example), with `(... tpl)` escape, `,@`, nested levels, and vector templates. Ordinary templates behave exactly as before; quasiquote outside match bodies is untouched.
4. **Spec cata evaluation order** — cata operators now run only after the clause's guard passes (the old port evaluated them during the structural match and documented the deviation). The structural match binds hidden temporaries; `%match-cata` applies operators afterwards and transposes per-ellipsis-level results, so one cata can bind several variables under `...` at any depth.

SRFI 202 came along cleanly on the same pattern and additionally gains SRFI 2's bare **bound-variable claw** (missing from the old port) and **vector patterns** in quasiquoted claws.

## Also in this PR

- `lib/srfi/148.sld` header: the claim that Kaappi "has no er-macro-transformer support" has been false since v0.22.0 — corrected. The code still (correctly) takes the reference's portable branch: the Chibi `er-macro-transformer` branch implements `free-identifier=?`/`bound-identifier=?` on top of a compare it assumes is binding-aware, which Kaappi's name-based compare is not (#2388).
- `docs/dev/srfi-implementation-notes.md`: new SRFI 241/202 section documenting the engine facts the port depends on (let-syntax with a bare-symbol transformer spec across import, counter-distinct renamed temporaries, the deliberate bare-`quasiquote` capture).

## Remaining scope notes (documented in the 241 header)

- Square-bracket clause notation — reader has no `[...]` syntax (pre-existing, unrelated to this port).
- A match form produced by a *syntax-rules* template is subject to that template's own ellipsis processing before match sees it (escape with `(... ...)`) — inherent to syntax-rules, not liftable here. Likewise a template-renamed `quasiquote` in a clause body resolves to the built-in quasiquote rather than the match-body binding.
- Only `match` is exported (as before). With a name-based compare, exporting bindings for the aux keywords (`...`, `->`, `_`, `guard`) would not change what is matched — see below.

## Evidence for #2388

Recorded while porting; **no compare change is made here** — that is #2388's decision.

**Where name-based compare was sufficient (everything these two libraries do):**

- Every literal check — `...`, `_`, `->`, `guard`, `unquote`, `unquote-splicing`, `quasiquote` (241), `values`, `quasiquote`, `unquote`, `_` (202) — is `(compare x (rename 'kw))`. Keywords arrive either as the plain interned symbol (direct use) or hygiene-renamed (`__hyg_N_kw` when the match form is emitted by another er-macro); name-stripping equality answers both correctly. Regression-covered: `srfi241.scm`'s "match emitted by an er-macro with renamed keywords" builds the whole pattern out of `(rename 'unquote)` / `(rename '...)` and passes.
- For the unbound aux keywords (`...`, `_`, `->`, `guard` — no `(scheme base)` binding to compare against), a binding-aware `free-identifier=?` degenerates to name equality on unbound identifiers anyway, so name-based compare is not an approximation there — it is the same answer.

**Where name-based compare is observably weaker (none blocked the port):**

- A use-site local rebinding of a keyword name cannot opt out of keyword-hood. The one literal in these libraries with a real binding is SRFI 202's `values`: a claw `((values a b) e)` written where the user has shadowed `values` is still parsed as the values-claw keyword; a binding-aware compare could distinguish. Same shape for `->` used as a local variable inside a cata position.
- Symmetrically, because compare strips renames, a macro-*generated* identifier that merely spells like a keyword is treated as the keyword (e.g. an er-macro emitting `(rename '_)` as a would-be pattern variable gets the wildcard instead). Harmless in practice — nothing binds `_` — but it is the general shape of the weakness.
- SRFI 148's Chibi fast path stays unusable: with a name-based compare, its derived `free-identifier=?` and `bound-identifier=?` collapse into the same predicate, so the portable branch remains (header now says exactly this).

**Conclusion offered to #2388:** for aux-syntax literals with no binding — the overwhelmingly common case, including R7RS's own `else`/`=>` — name-based compare produces the answers a binding-aware `free-identifier=?` would. The observable divergence requires either a use-site rebinding of a *bound* keyword name or macro-generated same-spelled identifiers; neither arose in re-porting two real macro libraries.

## Testing

- `tests/scheme/srfi/srfi241.scm`: 18 → **54** passes (every lifted limitation has dedicated tests, including the deferred-cata side-effect ordering and depth-2 cata transposition).
- `tests/scheme/srfi/srfi202.scm`: 25 → **31** passes (bare bound-variable claw, vector patterns).
- `tests/scheme/srfi/srfi148.scm`: 142 passes; `srfi211.scm`: 29 passes (unchanged).
- `zig build test`: all unit suites pass (no Zig changes).
- All runs with `KAAPPI_HOME=$(mktemp -d)` and a rebuilt `zig-out/lib` (the exe-relative library dir shadows worktree `.sld` edits until refreshed — now noted in the implementation notes).

Closes #2391

### DIFF
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -1456,8 +1456,12 @@ result lists so one cata can bind several variables under `...`). The
 202 re-port additionally gained SRFI 2's bare bound-variable claw and
 vector patterns in quasiquoted claws.
 
-Engine facts the port depends on (each probed before use, all
-regression-covered by `tests/scheme/srfi/srfi241.scm` / `srfi202.scm`):
+Engine facts the port depends on (each probed before use; the macro
+behavior in the first three bullets is regression-covered by
+`tests/scheme/srfi/srfi241.scm` / `srfi202.scm`, and the expansion-time
+diagnostics by `tests/scheme/errors/srfi-241-expansion-errors-2391.sh`;
+the last bullet is a testing-workflow note, not something those suites
+check):
 
 - **A macro expansion can rebind `quasiquote` via `let-syntax` with a
   bare-symbol transformer spec.** `match` wraps each clause body in
@@ -1485,4 +1489,8 @@ regression-covered by `tests/scheme/srfi/srfi241.scm` / `srfi202.scm`):
 - **A worktree's `.sld` edit is invisible until `zig-out/lib` is
   refreshed** — the exe-relative library dir is populated at `zig build`
   time, so rebuild (or re-copy) after editing, on top of the usual
-  `KAAPPI_HOME=$(mktemp -d)` isolation (kaappi#2352).
+  `KAAPPI_HOME=$(mktemp -d)` isolation (kaappi#2352). The
+  `KAAPPI_HOME` isolation and `.sld` staleness invalidation themselves
+  are regression-covered by `tests/scheme/cache/library-cache-1888.sh`;
+  the zig-out/lib refresh step is not covered by any test — it is a
+  local-workflow footgun to remember.
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -1435,3 +1435,54 @@ identifier macros, or output-provenance tracking a symbol-based expander
 cannot honestly provide — implicit renaming specifically cannot
 distinguish injected from macro-generated symbols when both are the same
 interned object.
+
+### SRFI 241 and SRFI 202 — match and pattern and-let*, on explicit renaming
+
+Both libraries were re-ported from pure `syntax-rules` onto
+`er-macro-transformer` (kaappi#2391, KEP-0006 step 5 — the acceptance
+test for SRFI 211). The old ports' template gymnastics — a custom `%%%`
+ellipsis identifier in every helper macro so the literal `...` token
+could be matched as data — are gone; each library is now one procedural
+transformer that compiles the pattern language by ordinary list
+processing. The re-port lifted all four of the old 241 port's documented
+limitations: arbitrary (compound/nested/cata) sub-patterns under an
+ellipsis, mandatory patterns after the ellipsis in lists and vectors
+(`(,x ... ,y . ,r)`, `#(,a ,m ... ,z)`), the SRFI's ellipsis-aware
+quasiquote inside clause bodies, and the spec's cata evaluation order
+(cata operators run only after the guard passes — hidden temporaries are
+bound during the structural match, then applied through the
+`%match-cata` runtime helper, which also transposes per-ellipsis-level
+result lists so one cata can bind several variables under `...`). The
+202 re-port additionally gained SRFI 2's bare bound-variable claw and
+vector patterns in quasiquoted claws.
+
+Engine facts the port depends on (each probed before use, all
+regression-covered by `tests/scheme/srfi/srfi241.scm` / `srfi202.scm`):
+
+- **A macro expansion can rebind `quasiquote` via `let-syntax` with a
+  bare-symbol transformer spec.** `match` wraps each clause body in
+  `(let-syntax ((quasiquote <renamed %match-qq>)) body ...)`, where
+  `%match-qq` is a library-level `(define %match-qq
+  (er-macro-transformer ...))` global holding a Transformer value; the
+  renamed reference resolves at the use site through the whole-def-env
+  import copy, and the binding shadows the built-in quasiquote for
+  exactly the body's scope. The binding NAME is the bare symbol
+  `quasiquote` (deliberately unhygienic — it must capture the body's
+  use-site backticks).
+- **Fresh temporaries are `rename` of counter-distinct symbols.**
+  Renaming the same symbol twice in one expansion yields the same
+  identifier, so every temporary is minted as `%mN.<base>` with a
+  per-expansion counter, then renamed.
+- **Keyword recognition (`...`, `_`, `->`, `guard`, `unquote`,
+  `quasiquote`, `values`) is `compare` against `rename` of the keyword —
+  hygiene-stripped name equality.** That strength suffices for these
+  macros (kaappi#2388 records the evidence); a use inside another
+  er-macro's output whose keywords arrive renamed still compares equal.
+  The known edge: a match form produced by a *syntax-rules* template is
+  subject to that template's own ellipsis processing first, and a
+  template-renamed `quasiquote` in a clause body resolves to the
+  built-in quasiquote, not the match-body one.
+- **A worktree's `.sld` edit is invisible until `zig-out/lib` is
+  refreshed** — the exe-relative library dir is populated at `zig build`
+  time, so rebuild (or re-copy) after editing, on top of the usual
+  `KAAPPI_HOME=$(mktemp -d)` isolation (kaappi#2352).
--- a/lib/srfi/241.sld
+++ b/lib/srfi/241.sld
@@ -1,135 +1,467 @@
 ;;; SRFI 241 — Match
 ;;;
-;;; A syntax-rules-only port of the Wright-Cartwright-Shinn matcher (Kaappi
-;;; has no syntax-case). The "trick" throughout is giving every helper macro
-;;; a custom ellipsis identifier (%%%) so the literal three-dot token can be
-;;; matched as ordinary data instead of triggering syntax-rules' own
-;;; repetition — that is what lets a pattern like (,x ... . ,y) be recognized
-;;; by shape rather than misinterpreted as "repeat the sub-pattern ,x".
+;;; An er-macro-transformer port of the Wright-Cartwright-Shinn matcher
+;;; (KEP-0006 step 5). The previous port was pure syntax-rules and had to
+;;; give every helper macro a custom ellipsis identifier (%%%) so the
+;;; literal three-dot token could be pattern-matched as data; the whole
+;;; pattern language is now compiled by ordinary list processing inside one
+;;; procedural transformer, which lifted the port's four documented
+;;; limitations:
 ;;;
-;;; Scope, relative to the full SRFI:
-;;;  - Ellipsis-repeated sub-patterns are supported only in the two shapes
-;;;    (subpat ...) and (subpat ... . tailpat), and only when subpat is a
-;;;    plain ,var, ,_, or a repeated default-cata ,(var) (which collects
-;;;    (map self ...) — the shape used by the SRFI's own "(+ ,[x*] ...)"
-;;;    example) — not an arbitrary compound pattern. This covers the
-;;;    overwhelmingly common "collect the rest" use.
-;;;  - Vector patterns support fixed length (#(p1 p2 ...)) and whole-vector
-;;;    ellipsis (#(p ...)), not a mixed mandatory-prefix/suffix segment.
-;;;  - Square-bracket clause notation isn't available (Kaappi's reader has no
-;;;    bracket syntax); write clauses and cata patterns with plain parens,
-;;;    e.g. (,(f -> x y) ...) rather than [,[f -> x y] ...].
-;;;  - The ellipsis-aware quasiquote that SRFI 241 binds inside match bodies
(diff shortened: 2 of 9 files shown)


tests/scheme/srfi/srfi241.scm
```diff
@@ -100,6 +101,188 @@
...
+;;; compare is hygiene-stripped name equality, so a renamed ... / unquote
+;;; emitted by another macro is still recognized) ---
+(define-syntax collect-all
+  (er-macro-transformer
+   (lambda (form rename compare)
+     (list (rename 'match) (cadr form)
+           (list (list (list (rename 'unquote) (rename 'x)) (rename '...))
+                 (rename 'x))))))
+
+(test-equal "match emitted by an er-macro with renamed keywords"
```
baijum, line 282: One gap relative to the repo's own evidence standard: everything here is interpreter-only, and per the LLVM-backend notes a `.scm` regression test is not native-tier evidence. Since this PR *is* the designated acceptance test for `er-macro-transformer`, a small `tests/scheme/compile/*.sh` smoke that bundle-compiles a program importing `(srfi 241)`/`(srfi 202)` would close that (plain `kaappi compile` refuses disk-resolved libraries by design, so `zig build -Dbundle-src=` is the route).

I verified the bundle route works today — mandatory-suffix ellipsis, compound subpattern under ellipsis, the ellipsis-aware quasiquote body, and a 202 vector claw all produce byte-identical output to the interpreter — so this is just the missing checked script, not a latent failure. Non-blocking; happy to see it as a follow-up if you'd rather keep this PR library-only.

lib/srfi/241.sld
```diff
@@ -1,135 +1,467 @@
...
+                                                                (list (car p)
+                                                                      (list r-reverse (cdr p))))
+                                                              accs)
+                                                         rest-code)
+                                                   (list r-let
+                                                         (list (list e (list r-car in)))
+                                                         elem-code))))
+                                 fk)))))))
+
+             ;; Success: guard first, then catas, then the body with the
```
baijum, line 419: Confirmed against the SRFI text: this ordering — pattern variables bound first, guards evaluated in the pattern-variable environment, then cata operators, then the body — is precisely the spec's evaluation sequence (the environment is extended separately for the cata bindings), so the deferred-cata design isn't just an improvement over the old port, it's the conforming order. The traced-cata tests pinning the side-effect ordering are exactly the right regression for it. Guards referencing a cata variable being a compile-time `undefined variable` error is also the correct loud failure given that order.

baijum (commented):
Extensive verification, all green — this is a really clean port. What I checked:

- **Suites**: srfi241 54, srfi202 31, srfi148 142, srfi211 29 — exactly the claimed counts, with an isolated `KAAPPI_HOME` and refreshed `zig-out/lib` (the exe-relative shadowing trap the notes now document).
- **Docs**: `markdownlint-cli2` clean on the implementation-notes change; the new section after SRFI 211 reads naturally; the 148.sld header's "since v0.22.0" claim is right (the SRFI 211 commit is contained in tag v0.22.0).
- **Adversarial probing beyond the suite** — all correct: nested `match` inside a clause body (the `let-syntax` quasiquote capture doesn't leak inward), depth-3 ellipsis, a literal-symbol dotted tail after an ellipsis (`(,x ... . end)`), a vector-with-nested-ellipsis under an ellipsis, guards seeing ellipsis-bound vars, guards keeping the ordinary quasiquote (`` `(a ,x ...) `` stays literal there), `values` claws with dotted collectors, internal `define` inside a clause body, `set!` of an ellipsis-bound variable, and multiple return values from a body.
- **Guard vs. cata variables**: referencing a cata variable from a guard is a clean compile-time error (`undefined variable 'a'`) — which is exactly what the spec's evaluation order implies (guards see only pattern-variable bindings; cata vars are bound later). Bare `_` being a literal (only `,_` wildcards) also matches the spec's symbol rule and the old port.
- **Malformed input**: bad claws/patterns hit the clear `match:` / `and-let*:` `verr` paths rather than misbehaving silently.
- **Old vs. new**: ran the same adversarial file against the pre-PR libraries — identical behavior on everything the old port supported (and the old port additionally rejects the lifted-limitation shapes, as expected). I specifically re-checked a quasiquoted pattern leading a general multi-value claw on both ports, since the rewrites of the `%match-vals*` dispatch are where that could have regressed — both give the same answer.
- **Native tier**: bundle-compiled a program exercising mandatory-suffix ellipses, compound subpatterns under ellipsis, the ellipsis-aware quasiquote, and 202 vector claws; output is byte-identical to the interpreter. One non-blocking suggestion inline about locking that in.

The `cp-ellipsis` compiler (accumulate-then-reverse with `vars-since` marks) and the deferred-cata design with `%match-cata`'s depth transposition are genuinely elegant, and the #2388 evidence section in the description is the most useful kind of acceptance-test writeup. Nothing blocking.

docs/dev/srfi-implementation-notes.md
```diff
@@ -1435,3 +1435,54 @@ identifier macros, or output-provenance tracking a symbol-based expander
...
+  macros (kaappi#2388 records the evidence); a use inside another
+  er-macro's output whose keywords arrive renamed still compares equal.
+  The known edge: a match form produced by a *syntax-rules* template is
+  subject to that template's own ellipsis processing first, and a
+  template-renamed `quasiquote` in a clause body resolves to the
+  built-in quasiquote, not the match-body one.
+- **A worktree's `.sld` edit is invisible until `zig-out/lib` is
+  refreshed** — the exe-relative library dir is populated at `zig build`
+  time, so rebuild (or re-copy) after editing, on top of the usual
+  `KAAPPI_HOME=$(mktemp -d)` isolation (kaappi#2352).
```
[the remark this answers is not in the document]
baijum: Reworded in e615ed20: the engine-facts intro now attributes the macro behavior to the SRFI suites plus the new `tests/scheme/errors/srfi-241-expansion-errors-2391.sh`, credits `tests/scheme/cache/library-cache-1888.sh` with the KAAPPI_HOME isolation and .sld staleness invalidation (verified against that script's contract comment), and explicitly marks the zig-out/lib refresh step as an uncovered local-workflow footgun rather than something any test checks.

tests/scheme/srfi/srfi241.scm
```diff
@@ -100,6 +101,188 @@
...
+  '(1 2 3)
+  (collect-all '(1 2 3)))
```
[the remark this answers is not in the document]
baijum: Added in e615ed20: `tests/scheme/errors/srfi-241-expansion-errors-2391.sh`, following the errors-suite shell pattern (reader-token-errors.sh shape rather than diagnostics-json.sh, since these are text-diagnostic assertions, not JSON-schema ones). One malformed form per diagnostic for all six verr branches — misplaced ellipsis, multiple ellipses in one list, invalid cata pattern, invalid cata variable, and %match-qq's unquote-splicing-outside-list-context and no-unquoted-expressions-under-ellipsis — each asserting nonzero exit plus the KP2002 message text, hermetic via throwaway KAAPPI_HOME and the checkout's --lib-path. 6/6 pass.

tests/scheme/srfi/srfi241.scm
baijum: Added in this PR (e615ed20) rather than as a follow-up: `tests/scheme/compile/srfi-241-202-bundle-2391.sh` with its own fixture at `fixtures/srfi-241-202/main.scm`. It follows bundle_fixture_binary's same-source/isolated-prefix discipline (kaappi#1930) but deliberately uses a separate fixture and lock: srfi imports would make the shared bundle-replay fixture's .sbc bytes depend on the srfi library search path, and its three scripts share one binary. The .sbc is compiled from inside the fixture dir with a throwaway KAAPPI_HOME (kaappi#2352) so the bytes stay deterministic and repeat -Dbundle builds are content-cache hits. The three output lines cover exactly the capabilities you listed — mandatory-suffix ellipsis with a compound subpattern under it, the ellipsis-aware quasiquote body, and a 202 vector claw (plus a values claw) — and the whole bundled-binary output is diffed against the interpreter run, per the oracle doctrine. Passes locally, both tiers byte-identical, matching your verification.

lib/srfi/241.sld
baijum: Thanks for checking the ordering against the SRFI text — good to have it confirmed that pattern-bindings → guard → cata operators → body is the conforming sequence and not just a fidelity improvement over the old port. No change needed; the traced-cata tests stay as the pin, and the loud compile-time undefined-variable for a guard touching a cata variable is intended.

PR #2399, opened 2026-08-28
Add the -Dgc-stress rooting tests for procedural-transformer calls

### LINKED ISSUE #2390: [GC] Add the -Dgc-stress rooting test for procedural-transformer calls (KEP-0006 step 1 exit criterion)
**Context**

KEP-0006 (amended 2026-08-27, keps#47) Implementation-plan step 1 defined an exit criterion for the reentrant-VM-during-compile machinery that `er-macro-transformer` (v0.22.0, #1811) rests on:

> a hand-written test in `tests_macros.zig` that defines a "transformer" this way and confirms rooting survives a forced GC (`-Dgc-stress=true`) during the call.

The mechanism shipped without that test. The rooting exists — the use path opens a no-collect window (`src/compiler_macro.zig:390`), keeps `pushRoot`/`popRoot` discipline inside the expander (`src/expander.zig:414–456`), and appends the expanded result to `gc.extra_roots` (`src/compiler_macro.zig:406`); the define-time path roots around `evalDatumForMacro` (`src/vm.zig:68–80`) — but nothing proves it under forced collection.

**Deliverable**

A test (in `src/tests_macros.zig`, alongside the existing SRFI 211 tests at `:1754+`) that:

1. defines an `er-macro-transformer` whose body **allocates** (enough to trigger collection under stress — cons up a list, build strings),
2. uses the macro so the transformer runs reentrantly from inside `compile()`,
3. verifies the expansion and the input form survive intact — the failure mode being a dangling reference into freed IR/AST data, not a clean assertion failure.

Run under `-Dgc-stress=true` (the flag exists: `build.zig:32,60,71`); the test should also pass without it. If rooting holes surface, fix them here — this is the correctness net for every future change to the expansion/compilation boundary.

**References**

- KEP-0006 as amended: Implementation-plan step 1 annotation and "As implemented" Divergence 7 (keps#47)
- Related precedent: `src/tests_gc_tracing.zig:356` (transformer trace coverage — that test covers *marking*, not reentrancy survival)

baijum:
KEP-0006 step 1's exit criterion — a hand-written test proving the reentrant-VM-during-compile machinery under `er-macro-transformer` keeps everything rooted when a GC fires around the transformer call — was never implemented. The rooting itself exists (the no-collect window in `compiler_macro.zig`, the `pushRoot`/`popRoot` discipline in `expander.expandProceduralMacro`, the `extra_roots` append, and the define-time rooting around `vm.evalDatumForMacro`), but nothing exercised it under forced collection. This PR adds that correctness net.

## New tests

Three tests in the new `src/tests_macros_procedural.zig`, each driving an allocation storm through a distinct rooting seam. Counts scale via `@import("build_options").gc_stress` (300 loop iterations under stress, 10,000 otherwise — each iteration conses a pair and builds a fresh string, so normal builds comfortably exceed the 8192-object collection threshold):

1. **Use path** (`SRFI 211 gc-stress: heavily-allocating er transformer keeps input form and expansion rooted (#2390)`) — the transformer conses up a large list of fresh strings, then only *after* the storm re-reads the use-site form, splices the raw `(+ 1 2)` input subform into the expansion for the compiler to compile afterwards, and embeds the freshly-built list as quoted literal data the compiler must keep alive while the deferred collections (the no-collect window defers them; they fire on the first allocation after it closes) run over the rest of compilation. Expands twice to prove reentrancy survival, not a one-shot fluke.
2. **Define-time path** (`SRFI 211 gc-stress: transformer-spec eval allocating heavily at define time stays rooted (#2390)`) — the transformer-spec expression itself allocates heavily before producing the procedure. This path (`vm.evalDatumForMacro`) runs with collections *live* (no no-collect window), so collections genuinely fire mid-eval; the resulting closure's captured list is read back at use time through the Transformer's GC trace.
3. **SRFI 213 re-entry hop** (`SRFI 213 gc-stress: allocation across the lookup re-entry hop keeps form and lookup rooted (#2390)`) — both the transformer body and the procedure-result re-entry hop allocate storms; the second hop re-reads the original input form (rooted across every hop) and consults the property table through the `lookup` NativeFn (`res_root`/`lookup_root` in the hop loop).

Under `-Dgc-stress=true` every allocation collects (or defers a collection). Per the review's mutation testing: bypassing the no-collect window makes all three tests panic with "GC: marking freed object", so the window — the primary protection — and the overall invariant (no in-flight intermediate left exposed while collections are live) are genuinely pinned. The individually-named explicit roots are defense-in-depth these tests corroborate but do not isolate: with the window intact, each value is also reachable through a redundant cover at the instant collections fire (the compile boundary's source-tree rooting for the input/spec datums, plus the allocators' argument auto-rooting as the expansion passes between allocating steps — not the VM register file, which `markVmRoots` only marks for live frames), so removing one of those roots alone keeps this suite green. An exclusive-hold test would need a value provably unreachable from every cover at the collection instant, which depends on compiler-internal allocation ordering and is not constructible from Scheme-level test code.

## File split

`src/tests_macros.zig` was already at 1963 lines, over the repo's 1500-line policy, so the new tests must not grow it. The SRFI 211/213 engine-seam block (previously at `:1754+`) moves verbatim into `src/tests_macros_procedural.zig` (registered in `src/vm_tests.zig` next to `tests_macros_nested_sr.zig`, the previous split from the same file), and the new tests live there. `tests_macros.zig` drops to 1777 lines and loses its now-unused `compiler` import; a pointer comment marks where the block went.

## Rooting holes found

None. All three tests pass both with and without `-Dgc-stress=true` — the existing rooting (no-collect window, expander root discipline, `extra_roots`, define-time roots) holds under forced collection. The machinery was correct; it just had no proof.

## Verification

- `zig build test`: 1782 pass, 7 skip, 0 fail (plus 88 thottam tests).
- `zig build test -Dgc-stress=true -Dtest-filter="#2390"`: all 3 new tests confirmed executed by name and passing.
- `zig build test -Dgc-stress=true -Dtest-filter="SRFI 211"`: all 13 named SRFI 211 tests (11 moved + 2 new) confirmed executed under stress and passing.
- Note: `zig build test` emits a `failed command: .../unit-tests --listen=-` line plus two KP3002 diagnostics on stderr while still exiting 0 with all tests passing; verified identical on a clean origin/main checkout — pre-existing on main (the diagnostics come from `tests_step`'s error-at-end programs), unrelated to this change.

Closes #2390


src/tests_macros_procedural.zig
```diff
@@ -0,0 +1,321 @@
...
+// machinery must keep every live value rooted when collections fire around a
+// procedural-transformer call. The load-bearing pieces are the no-collect
+// window compiler_macro.zig opens across expandMacro (a use-path transformer
+// call defers collections; the deferred storm then fires on the first
+// allocation after the window closes, with the expansion held only by the
+// extra_roots append), the pushRoot discipline inside
+// expander.expandProceduralMacro (input form + rename/compare/lookup
+// NativeFns across the SRFI 213 re-entry hops), and the define-time rooting
+// around vm.evalDatumForMacro. Under -Dgc-stress=true every allocation
+// collects (or defers a collection), so a single unrooted value in any of
```
baijum: I mutation-tested this claim and it holds for the no-collect window, but not for the individually-named roots. I removed three roots — (a) the `extra_roots.append` in `compiler_macro.zig` (replaced with a no-op shrink), (b) `gc.pushRoot(&input_root)` in `expander.expandProceduralMacro`, (c) the `expr_root` push in `vm.evalDatumForMacro` — **all three at once** — and `zig build test -Dgc-stress=true -Dtest-filter="#2390"` still passed. In exactly these scenarios each value evidently stays reachable through a redundant cover — most plausibly the re-entrant VM's marked register file still holding the transformer-call result, plus caller-side roots for the input/spec datums.

The flip side does hold, and is worth knowing: with all roots intact, bypassing the window (`if (self.no_collect > 0)` in `maybeCollect` → constant false) makes all three tests panic with `GC: marking freed object (use-after-free)`. So the tests genuinely pin the no-collect window plus the overall invariant — any change that leaves in-flight intermediates unrooted while collections are live gets caught.

Consequence for this paragraph: a future refactor that drops any *one* of those three explicit roots keeps this suite green, so "a single unrooted value in any of these stretches panics deterministically" overstates what's pinned — the values aren't singly-rooted in these stretches. If pinning the individual roots matters for the step 1 exit criterion, the tests need a scenario where a value is held exclusively through one seam (e.g. proving the expansion is not incidentally reachable from a marked register at the instant the deferred collection fires). Otherwise I'd reword this to present the window as the primary protection and the explicit roots as defense-in-depth that these tests corroborate but don't isolate.

src/tests_macros.zig
```diff
@@ -1751,195 +1750,10 @@ test "SRFI 147: a bare symbol resolving to nothing is a compile error" {
...
-    const result = ctx.vm.eval(
-        \\(define (f)
-        \\  (define-property f f 1)
-        \\  1)
-    );
-    try std.testing.expectError(error.CompileError, result);
-}
+// The SRFI 211 (procedural macro transformers) + SRFI 213 (identifier
+// properties) engine-seam tests — including the KEP-0006 step 1 forced-GC
+// rooting tests (#2390) — live in tests_macros_procedural.zig (split out to
```
baijum, line 1755: Nit since this comment now names the policy: after the split the file sits at 1777 lines, still ~280 over the 1500 target. Fine as an advisory stopgap (and the PR description says as much) — but if one more natural seam is ever worth taking, the SRFI 147 block just above or the let-syntax/root-hygiene tests would be the candidates. Totally optional, not asking for it here.

baijum (commented):
Verified before commenting:

- The move is verbatim — I diffed the removed block against the new file; the only deltas are the extended header comment and the new file's own imports. Line counts match the description (tests_macros.zig → 1777, new file 321), registration sits next to `tests_macros_nested_sr.zig`.
- `zig build test -Dtest-filter="#2390"` passes both normally and under `-Dgc-stress=true`; the three new tests are confirmed executed by name (8 tests matched the filter).

I also mutation-tested the load-bearing claim (details inline). Short version: the tests are a genuine net for rooting-regime breakage — bypassing the no-collect window makes all three panic — but each of the three individually-named roots survives removal, even all three removed at once. Not blocking since the tests strictly add coverage, but the determinism paragraph promises more pinning than the tests deliver; either soften it or chase the exclusively-held scenario.

src/tests_macros_procedural.zig
baijum: Agreed on both halves — reworded in d4a5c58d (banner comment) and in the PR description: the no-collect window is presented as the primary protection (your window-bypass mutation panics all three tests, which is the invariant these pin), and the named explicit roots as defense-in-depth the tests corroborate but do not isolate.

One refinement to the redundant-cover hypothesis, verified against the code: it is probably not the re-entrant VM's register file. markVmRoots (src/vm.zig:232) marks registers only per live frame (frames[0..frame_count], each frame's window), and by the time the deferred storm fires the transformer call has returned and those frames are popped. The covers that fit the observation are (a) the compile boundary's rooting of the source tree, which reaches the input/spec datums, and (b) the allocators' argument auto-rooting (arg_roots/slice_roots), which keeps the expansion reachable at each allocating step as it is handed through stripUsertextMarkers/injection/compilation — a custody chain, not a lingering register.

I looked at whether a clean exclusive-hold test falls out and stopped, per your 'otherwise reword' branch: flipping on any ONE root alone needs the value provably unreachable from every cover at the exact collection instant, and that chain is a property of compiler-internal allocation ordering — nothing Scheme-level test code can arrange, and a Zig-level construction would pin today's allocation order rather than the rooting contract. The existing tests are unchanged; the banner now also says explicitly that the roots stay load-bearing as the contract even though the covers make them non-isolatable here.

src/tests_macros.zig
baijum: Acknowledged — leaving it at 1777 in this PR as you suggest. Agreed on the candidate seams for whenever the next split is worth taking: the SRFI 147 custom-transformer block is the most self-contained (it already has a sibling precedent in tests_macros_nested_sr.zig), and the let-syntax/root-hygiene tests at the tail are the other natural unit. Noting both here so the next person picking at this file has the seams named.

Issue #2119, opened 2026-08-01
[Native] Re-invoking a top-level continuation from a native callback silently keeps the stale value instead of resuming


baijum:
**Category:** Native
**Severity:** wrong-result (silent — no error, no diagnostic, the wrong number)

`docs/dev/decisions/continuation-strategy.md` § "Conformance plan" commits to:

> **Behavioral equivalence:** programs that use `call/cc` must produce identical results whether compiled natively (with VM fallback) or executed purely on the VM

and lists "continuation captured across function boundaries" as *deferred to VM fallback* — deferred, i.e. it should still give the right answer, more slowly. It does not: invoking a continuation captured at top level from inside a **natively compiled callback** silently abandons the resumption and leaves the pre-capture value in place.

**Minimal reproduction:**

```scheme
(define k1 #f)
(define result 0)
(set! result (+ 100 (call-with-current-continuation (lambda (c) (set! k1 c) 0))))
(define (go) (for-each (lambda (x) (when (= x 2) (k1 42))) (list 1 2 3)) (display "reached\n"))
(go)
(display result)
(newline)
```

```text
interpreter: 142      native: 100
```

Neither tier prints `reached`, so the `(k1 42)` call *does* abort the `for-each` on both. The native binary just never re-enters the top-level form the continuation was captured in, so `result` keeps the `100` from the original pass — and exits 0 with no diagnostic.

**Discriminating control:** the same escape used *within* its dynamic extent —

```scheme
(define (esc)
  (call-with-current-continuation
    (lambda (k) (for-each (lambda (x) (if (= x 2) (k 42))) (list 1 2 3)) 100)))
(display (esc)) (newline)
```

prints `42` on **both** tiers. So the native `call/ec`-through-a-callback path is fine; what fails is re-entry after the capturing frame returned.

**Blast radius, measured:** `tests/scheme/smoke/call-global-continuation.scm` fails 2 of its 5 assertions natively (`for-each-continuation got=100 expected=142`, `map-continuation got=200 expected=250`) while passing under the interpreter. No suite runs it through `kaappi compile`.

A silent wrong number is the worst available outcome here. If the re-entry genuinely cannot be supported natively, the runtime should raise rather than continue with a stale value — the same argument that made VM limit overflows uncatchable-but-loud in #1886.

**Verified:** fresh ReleaseSafe build at 3c1a56be, isolated `KAAPPI_HOME`, macOS aarch64, `zig cc` link.

[...]

PR #2401, opened 2026-08-28
Make er-macro compare binding-aware free-identifier=? and land the four-quadrant test

### LINKED ISSUE #2388: [SRFI 211] compare is name-based — implement binding-aware free-identifier=? or ratify the deviation; land the four-quadrant test

### LINKED ISSUE #2404: [SRFI 211] `compare` hangs forever on a circular macro-use input (#2401)
**Category:** Expander (SRFI 211 ER macros) · **Severity:** hang · **Found by:** review of #2401

> **Not on `main`.** This defect lives only in the `issue-2388-er-compare` branch of #2401 (commit `030cc6a8`). Filed separately so it is tracked independently of that PR's review thread; it must be fixed before #2401 merges.

`compare` walks the macro-use input looking for a symbol, with no cycle detection. R7RS datum labels (§7.1.2) produce a genuinely circular structure, so the walk never returns.

```scheme
(import (scheme base) (scheme write) (srfi 211 explicit-renaming))
(define-syntax self-else?
  (er-macro-transformer
   (lambda (form rename compare)
     (list (rename 'quote) (compare (rename 'else) (rename 'else))))))
(display "before") (newline)
(display (self-else? #0=(zz . #0#)))   ;; never returns
```

| build | result |
|-------|--------|
| `origin/main` (5e1680f2) | prints `before`, then `#t`; exits 0 in <1s |
| #2401 @ `030cc6a8` | prints `before`, then spins; killed by SIGALRM at 20s (rc=142) |

## Where

`erFormMentionsSymbol` (`src/expander.zig`), added in `030cc6a8`:

```zig
var cur = form;
while (types.isPair(cur)) : (cur = types.cdr(cur)) {
    if (erFormMentionsSymbol(types.car(cur), sym)) return true;
}
```

Nothing bounds the cdr spine. It is reached from `erCompareFn`'s bare/bare branch via the new `er_input_slot`, which is why `compare` never had this exposure before — prior to `030cc6a8` it only inspected its two arguments.

## Discriminating controls

- **The cycle itself is real, not a printed alias**: `(define x '#0=(a . #0#))` ⟹ `(eq? x (cdr x))` is `#t`. So this is not a reader artifact.
- **A plain circular datum is fine**: `(write '#0=(zz . #0#))` terminates; cycle handling elsewhere works.
- **The searched spelling must be absent from the input**, or the walk short-circuits on the first match before reaching the cycle. `(kw-probe else #0=(zz . #0#))` returns `#t` promptly; the repro above compares two `rename` products of a spelling nowhere in the input, forcing the full traversal.
- **An ER transformer that never triggers the walk is unaffected** — the `stripUsertextMarkers` pre-pass in `expandProceduralMacro` handles the same input fine.

## Why `high` and not `critical`

It is an unbounded spin, not an abort: no panic, no root-stack exhaustion. It happens in the expander rather than at a VM safepoint, so the execution-timeout machinery cannot interrupt it and no `guard` can see it — the program simply never continues. That is the `high` row ("a legal program … hangs, in a path users actually reach"), not the process-level-unsafety row.

Closely related, and **already on `main`**: #2403 — `rename` on the same circular input aborts the process. Same root cause (ER helpers walk the macro-use input assuming it is acyclic); a shared cycle-aware traversal would fix both.

## Suggested fix

The walk only ever answers "does this one interned symbol occur here", so it does not need to be exact. Either carry a visited set, or bound the traversal the way the reader bounds nesting (`KP1009`) and treat exhaustion as "not found" — a conservative `false` there degrades to the pre-`030cc6a8` reflexive answer rather than to a wrong refusal.

A regression test belongs in `tests/scheme/srfi/srfi211.scm` next to the reflexivity assertions the same commit added.

baijum:
Closes #2388

## The decision: option (a) — implement binding-aware `compare`

KEP-0006 (amended 2026-08-27) already resolved Unresolved question 2 against SRFI 211's contract — "`compare` is `free-identifier=?`" — with a falsifiable four-quadrant test; ratifying the name-based deviation (option b) would have contradicted the amended KEP. The #2398 evidence step (SRFI 241/202 re-port, #2398) completed in the meantime: name-based compare held up in real library code, but recorded concrete observable weaknesses — SRFI 202's `values` claw parsing as the keyword even where the user shadowed `values`, and macro-generated same-spelled identifiers counting as keywords. Nothing blocked the stronger semantics, and the reuse the KEP's design section sketched ("the same machinery, exposed a second way") turned out to be genuinely small.

## What changed

**`src/expander.zig` — `erCompareFn` is now free-identifier=?**, mirroring `matchPattern`'s symbol-literal branch outcome-for-outcome, reusing the machinery it already runs on:

- Each argument is classified as a **use-site local slot** (`UseSiteBindingCheck.resolve` — the same callback literal matching uses), a **def-env binding** (`def_env_binding_prefix`-marked spelling as identity), or **free** (aux syntax, gensym'd rename/template products, unshadowed globals). Same classification ⇒ same binding; free/free ⇒ both unbound ⇒ equal.
- The one wrinkle symbols-as-syntax impose: **symbols are interned**, so `(rename 'else)` (reserved forms rename to themselves) is *the same object* as a user-typed `else` — no per-side classification is possible for a bare/bare pair. `erRenameSymbol` therefore records identity entries in the per-invocation scope table, and a bare/bare pair where that spelling was bare-renamed this invocation answers "do the def-site and use-site views agree" (a local rebinding ⇒ `#f`); any other bare/bare pair is two use-site tokens — the same identifier — and stays reflexive, so the pairwise input-token comparison idiom (duplicate-key checks) keeps working. The rule is keyed on the pair, not the argument position, so it is order-independent.

**The KEP-0006 four-quadrant acceptance test, finally written** — as an ER/syntax-rules *parity* suite in `tests/scheme/srfi/srfi211.scm`: a cond-style `erq-cond` transformer and a `srq-cond` `syntax-rules` twin with `else`/`=>` literals, every quadrant asserted against the **same expected value** for both systems. That pins KEP-0018 UQ6 — "an ER macro is exactly as hygienic as a syntax-rules one" — as a documented guarantee (the issue's References asked for exactly that fold-in). Also covers reflexivity, reversed-argument compare, and the `values` claw improvement.

**The audit pin flipped** — `primitives_smallbatch-audit.scm` asserted name-based compare *precisely so a stronger compare shows up as a test change*; it now asserts the refusal for both `else` and `=>`, plus the quadrant coverage moved into the parity suite.

**Docs**: the SRFI 211 `.sld` header's "compare is name-based" deviation paragraph is replaced by the binding-aware contract, the shared-limitation statement (reserved-form spellings kept bare are shadowed for macro-introduced occurrences in *both* systems; spellings the engine can mark, like `=>`, stay hygienic), and the verified-by pointers. `docs/dev/srfi-implementation-notes.md`'s SRFI 211 mechanism section and the 241/202 "keyword recognition" bullet updated likewise.

## Behavior change, precisely

`(compare x (rename 'kw))` where a use-site **local** shadows `kw`'s spelling now answers `#f` (was `#t`) — for reserved-form keywords (`else`), renamed global-name keywords (`values`), and non-reserved ones (`=>`) alike; everything else answers as before. SRFI 211/213/241/202/148 suites pass unchanged (29→48, 17, 54, 31, 142).

## Also in this PR

**Closes #2404 fully** — the `compare` input-walk hang (fixed in 5160768c) *and* the pre-existing `compiler.collectSetTargets` instance of the same class found while verifying it: the scan's depth cap counts only car recursion, so a cyclic macro operand re-emitted into a body position (e.g. a `test-assert` operand) hung the compiler on `origin/main`. Both spine loops now carry a 1M-step cap; exhaustion takes the scanner's ordinary loses-optimization-never-correctness truncation path (`set_targets_all` boxing, Part B correct-late). Regression suite: `tests/scheme/hygiene/cyclic-macro-input-2404.scm`. (#2403, `erRenameDatum` on circular input, stays out: no finite rename exists for an unfoldable cycle, so it needs a semantic decision.)

- **kaappi#2400** (filed, not fixed here): a *pre-existing* quirk found while writing the Zig tests — a `vm.eval`'d `define-syntax` re-using a macro keyword name from an earlier eval in the same process can fail with a bare `CompileError`. Reproduces on `origin/main` with the old compare (verified by stashing this change). The new Zig tests use `q2388-`-prefixed unique names with a comment pointing at the issue.
- KEP-0006's "As implemented" section (Divergence 1) and the four-quadrant record live in the keps repo — they'll need a small as-built amendment there once this lands; happy to open the keps PR as a follow-up.

## Testing

- `zig build test` — all unit suites pass, including four new engine-seam tests in `tests_macros_procedural.zig`.
- `zig build test -Dgc-stress=true` — passes (compare allocates nothing; the identity entries follow the existing scope-table slice lifetime pattern).
- `bash tests/scheme/run-all.sh` — **2118 pass, 0 fail** (723 Scheme files + the 1395-test R7RS suite).
- `bash tests/scheme/compile/srfi-241-202-bundle-2391.sh` — native tier agrees with the interpreter on the re-ported er-macro libraries.
- `npx markdownlint-cli2` — clean.

### DIFF
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -1422,12 +1422,19 @@ interned symbols as syntax: a bare-rename product (reserved forms and
 keywords rename to themselves) is the same object as a use-site token of
 that spelling, so compare recognizes the classic
 `(compare <token> (rename 'kw))` shape from the invocation's rename
-record plus whether the spelling occurs in the macro-use input —
-order-independent, reflexive for the invocation's own rename products
-(the hoisted-rename self-compare) and for two plain use-site tokens, and
-non-reflexive only for the quadrant-2 shape: a spelling that occurs in
-the input, bare-renamed this invocation, compared under a use-site local
-shadow of it. A
+record plus whether the spelling occurs in the macro-use input (a walk
+bounded by a node budget and depth cap — datum labels make inputs
+genuinely circular, kaappi#2404 — with exhaustion counting
+conservatively as occurrence) — order-independent, reflexive for two
+plain use-site tokens and for the invocation's own rename products
+whenever the spelling is absent from the input. The unsettled shape,
+stated plainly: a spelling that occurs in the input AND was bare-renamed
+this invocation, compared under a use-site local shadow — the refusal is
+free-identifier=?'s demanded answer when one argument is that input
+token, and known-wrong (broken reflexivity) when both arguments were the
+invocation's own rename products; a distinguishable wrapper for bare
+rename products would break the compiler's bare matching of the reserved
+forms macros emit. A
 bare-symbol
 spec falls back to a globals lookup holding a Transformer value, so
 `(define t (er-macro-transformer p))` + `(define-syntax m t)` works. Two
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -1405,19 +1405,29 @@ reaches the expansion (verified equivalent on both paths). Since kaappi#2388
 free-identifier=? built from the same machinery syntax-rules literal
 matching uses — `UseSiteBindingCheck.resolve` for use-site binding slots,
 `rename`'s per-invocation scope-table identity entries to recognize the
-definition-side bare spelling, def-env-prefixed names as binding identity
-for renamed library references — which makes "an ER macro is exactly as
-hygienic as a syntax-rules one" (KEP-0018 unresolved question 6) a pinned
-guarantee rather than an aspiration: the KEP-0006 four-quadrant test in
-`tests/scheme/srfi/srfi211.scm` asserts every quadrant against BOTH
-systems with the same expected value. One approximation is inherent to
+definition-side bare spelling, def-env-prefixed names agreeing with a
+use-site reference the use site can resolve to the same exported binding
+— which makes "an ER macro is exactly as hygienic as a syntax-rules one"
+(KEP-0018 unresolved question 6) a pinned guarantee for the
+auxiliary-keyword spellings (reserved forms and macro keywords, plus
+gensym-marked renames of any other spelling): the KEP-0006 four-quadrant
+test in `tests/scheme/srfi/srfi211.scm` asserts every quadrant against
+BOTH systems with the same expected value. The guarantee's boundary is
+pinned there too: spellings whose bare rename comes from
+renameForHygiene's other bare-returning branches (the VOID sentinel for a
+name defined later in the use-site body) keep compare's reflexive
+use-token view where a literal refuses — a pre-existing divergence.
+One approximation is inherent to
 interned symbols as syntax: a bare-rename product (reserved forms and
 keywords rename to themselves) is the same object as a use-site token of
 that spelling, so compare recognizes the classic
-`(compare <token> (rename 'kw))` shape by the invocation's rename record
-— order-independent, but not reflexive for that one spelling when a
-use-site local shadows it (two plain use-site tokens of a spelling this
-invocation did not bare-rename stay reflexive). A
+`(compare <token> (rename 'kw))` shape from the invocation's rename
+record plus whether the spelling occurs in the macro-use input —
+order-independent, reflexive for the invocation's own rename products
+(the hoisted-rename self-compare) and for two plain use-site tokens, and
+non-reflexive only for the quadrant-2 shape: a spelling that occurs in
+the input, bare-renamed this invocation, compared under a use-site local
+shadow of it. A
 bare-symbol
 spec falls back to a globals lookup holding a Transformer value, so
 `(define t (er-macro-transformer p))` + `(define-syntax m t)` works. Two
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -1400,7 +1400,24 @@ save/restore discipline as the syntax-rules `active_*` context), and
 `rename` reuses `renameForHygiene` — so ER macros get exactly the hygiene
 strength syntax-rules templates have, including the shared pre-existing
 limitation that a use-site top-level redefinition of a referenced name
-reaches the expansion (verified equivalent on both paths). A bare-symbol
+reaches the expansion (verified equivalent on both paths). Since kaappi#2388
+(KEP-0006's resolved direction), `compare` is binding-aware
+free-identifier=? built from the same machinery syntax-rules literal
+matching uses — `UseSiteBindingCheck.resolve` for use-site binding slots,
+`rename`'s per-invocation scope-table identity entries to recognize the
+definition-side bare spelling, def-env-prefixed names as binding identity
+for renamed library references — which makes "an ER macro is exactly as
+hygienic as a syntax-rules one" (KEP-0018 unresolved question 6) a pinned
+guarantee rather than an aspiration: the KEP-0006 four-quadrant test in
+`tests/scheme/srfi/srfi211.scm` asserts every quadrant against BOTH
+systems with the same expected value. One approximation is inherent to
+interned symbols as syntax: a bare-rename product (reserved forms and
+keywords rename to themselves) is the same object as a use-site token of
+that spelling, so compare recognizes the classic
+`(compare <token> (rename 'kw))` shape by the invocation's rename record
+— order-independent, but not reflexive for that one spelling when a
+use-site local shadows it (two plain use-site tokens stay reflexive). A
+bare-symbol
 spec falls back to a globals lookup holding a Transformer value, so
 `(define t (er-macro-transformer p))` + `(define-syntax m t)` works. Two
 non-obvious integration points: (1) `vm_imports.copyTransformerFreeRefs`
@@ -1479,9 +1496,14 @@ check):
   per-expansion counter, then renamed.
 - **Keyword recognition (`...`, `_`, `->`, `guard`, `unquote`,
   `quasiquote`, `values`) is `compare` against `rename` of the keyword —
-  hygiene-stripped name equality.** That strength suffices for these
-  macros (kaappi#2388 records the evidence); a use inside another
-  er-macro's output whose keywords arrive renamed still compares equal.
+  binding-aware free-identifier=? since kaappi#2388 (name-stripped
+  equality before it; that strength is what the #2398 re-port evidence
+  validated).** A use inside another er-macro's output whose keywords
+  arrive renamed still compares equal (both sides unbound); a use-site
+  local rebinding of a keyword spelling now opts out of keyword-hood
+  exactly as a syntax-rules literal refuses the same shadowing — nothing
+  in either library's own test suite does that, so the port is unaffected
+  (the one bound keyword, 202's `values` claw, gains the refusal).
   The known edge: a match form produced by a *syntax-rules* template is
   subject to that template's own ellipsis processing first, and a
   template-renamed `quasiquote` in a clause body resolves to the
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -1416,7 +1416,8 @@ keywords rename to themselves) is the same object as a use-site token of
 that spelling, so compare recognizes the classic
 `(compare <token> (rename 'kw))` shape by the invocation's rename record
 — order-independent, but not reflexive for that one spelling when a
-use-site local shadows it (two plain use-site tokens stay reflexive). A
+use-site local shadows it (two plain use-site tokens of a spelling this
+invocation did not bare-rename stay reflexive). A
 bare-symbol
 spec falls back to a globals lookup holding a Transformer value, so
 `(define t (er-macro-transformer p))` + `(define-syntax m t)` works. Two
--- a/lib/srfi/211/explicit-renaming.sld
+++ b/lib/srfi/211/explicit-renaming.sld
@@ -62,8 +62,9 @@
 ;;;     the same object as a use-site token of that spelling; compare
 ;;;     recognizes the classic (compare <token> (rename 'kw)) shape by the
 ;;;     invocation's rename record, which makes the answer independent of
-;;;     argument order, and stays reflexive for two plain use-site tokens
-;;;     (the pairwise input-comparison idiom).
+;;;     argument order. Two plain use-site tokens stay reflexive unless
+;;;     this invocation also renamed that spelling bare and a use-site
+;;;     local shadows it (the pairwise input-comparison idiom).
 ;;;
 ;;;   * `(rename 'x)` used to BIND x when x names a global procedure
 ;;;     returns x unrenamed (reference semantics win); rename fresh names
--- a/lib/srfi/211/explicit-renaming.sld
+++ b/lib/srfi/211/explicit-renaming.sld
@@ -74,14 +74,22 @@
 ;;;     the same object as a use-site token of that spelling; compare
 ;;;     recognizes the classic (compare <token> (rename 'kw)) shape from
 ;;;     the invocation's rename record together with whether the spelling
-;;;     occurs in the macro-use input — which makes the answer independent
-;;;     of argument order, and reflexive both for two of the invocation's
-;;;     own rename products (the hoisted-rename style) and for two plain
-;;;     use-site tokens (the pairwise input-comparison idiom). The one
-;;;     non-reflexive shape: a spelling that occurs in the input, renamed
-;;;     bare this invocation, compared under a use-site local shadow of
-;;;     it — the quadrant-2 case, where the shadowing local is a
-;;;     different binding.
+;;;     occurs in the macro-use input (a bounded walk — datum labels make
+;;;     inputs genuinely circular, and exhaustion conservatively counts as
+;;;     occurrence) — which makes the answer independent of argument
+;;;     order, and reflexive for two plain use-site tokens (the pairwise
+;;;     input-comparison idiom) and for two of the invocation's own
+;;;     rename products whenever the spelling is absent from the input.
+;;;     The shape compare cannot settle, stated plainly: a spelling that
+;;;     occurs in the input AND was bare-renamed this invocation, compared
+;;;     under a use-site local shadow — if one argument is that input
+;;;     token, the refusal is exactly the answer free-identifier=? demands
+;;;     (the shadowing local is a different binding); if both arguments
+;;;     were the invocation's own rename products, the #f is known-wrong
+;;;     (a broken reflexivity), because interned symbols make the two
+;;;     cases representationally identical and a distinguishable wrapper
+;;;     for bare rename products would break the compiler's bare matching
+;;;     of the reserved forms macros emit.
 ;;;
 ;;;   * `(rename 'x)` used to BIND x when x names a global procedure
 ;;;     returns x unrenamed (reference semantics win); rename fresh names
--- a/lib/srfi/211/explicit-renaming.sld
+++ b/lib/srfi/211/explicit-renaming.sld
@@ -11,8 +11,11 @@
 ;;;                      that cannot capture (or be captured by) use-site
 ;;;                      names. Renaming the same symbol twice within one
 ;;;                      expansion yields the same identifier.
-;;;   (compare a b)    — free-identifier=? to the strength this engine can
-;;;                      answer (see limitations below).
+;;;   (compare a b)    — free-identifier=?: the two identifiers denote the
+;;;                      same binding, or both are unbound — the rule a
+;;;                      syntax-rules literal matches by (R7RS 4.3.2), at
+;;;                      the same strength this engine answers it for
+;;;                      literals (see the compare paragraph below).
 ;;;
 ;;; SRFI 211 permits providing only some of its libraries but each provided
 ;;; one whole; this library's full export set is er-macro-transformer plus
@@ -35,22 +38,42 @@
 ;;;     whole environment is the honest static over-approximation).
 ;;;
 ;;;   * Hygiene strength equals this engine's own syntax-rules hygiene
-;;;     exactly — rename routes through the same renameForHygiene the
-;;;     template instantiator uses. The shared, pre-existing limitation:
-;;;     a use-site TOP-LEVEL redefinition of a name the macro's output
-;;;     references (e.g. redefining a library helper's name after import)
-;;;     reaches the expansion, for procedural and syntax-rules macros
-;;;     alike. compare is name-based (hygiene-stripped equality), which
-;;;     answers the classic literal checks (else, =>) correctly but cannot
-;;;     distinguish two same-named bindings.
+;;;     exactly — the documented guarantee (KEP-0018 unresolved question 6,
+;;;     decided with kaappi#2388): rename routes through the same
+;;;     renameForHygiene the template instantiator uses, and compare
+;;;     through the same binding machinery literal matching uses, so the
+;;;     two macro systems answer every keyword-check quadrant identically
+;;;     (regression-pinned, quadrant for quadrant, by the KEP-0006
+;;;     four-quadrant test in tests/scheme/srfi/srfi211.scm). Shared,
+;;;     pre-existing limitations, identical on both paths: a use-site
+;;;     TOP-LEVEL redefinition of a name the macro's output references
+;;;     reaches the expansion; and a reserved-form spelling the hygiene
+;;;     engine keeps bare (else, _, ...) is shadowed by a use-site local of
+;;;     that spelling for a macro-INTRODUCED occurrence just as surely as
+;;;     for a user-typed one, while identifiers the engine CAN mark (any
+;;;     non-reserved spelling, e.g. =>) stay hygienic.
+;;;
+;;;   * compare is binding-aware (kaappi#2388): (compare x (rename 'kw))
+;;;     answers whether the use-site identifier x and the definition-side
+;;;     keyword denote the same binding — an unshadowed use is #t, a
+;;;     use-site local rebinding of the spelling is #f, and a renamed
+;;;     (macro-introduced or global-name) keyword never resolves to a
+;;;     use-site local. Symbols are interned, so a bare-rename product is
+;;;     the same object as a use-site token of that spelling; compare
+;;;     recognizes the classic (compare <token> (rename 'kw)) shape by the
+;;;     invocation's rename record, which makes the answer independent of
+;;;     argument order, and stays reflexive for two plain use-site tokens
+;;;     (the pairwise input-comparison idiom).
 ;;;
 ;;;   * `(rename 'x)` used to BIND x when x names a global procedure
 ;;;     returns x unrenamed (reference semantics win); rename fresh names
 ;;;     for binders — the classic ER idiom — and they gensym correctly.
 ;;;
-;;; Verified: tests/scheme/srfi/srfi211.scm and the "SRFI 211" tests in
-;;; src/tests_macros.zig (swap!/my-or hygiene, compare on else, rename of
-;;; whole trees, library-helper resolution, let-syntax bodies).
+;;; Verified: tests/scheme/srfi/srfi211.scm (including the KEP-0006
+;;; four-quadrant acceptance test and the ER/syntax-rules parity suite)
+;;; and the "SRFI 211" tests in src/tests_macros_procedural.zig
+;;; (swap!/my-or hygiene, compare quadrants, rename of whole trees,
+;;; library-helper resolution, let-syntax bodies).
 (define-library (srfi 211 explicit-renaming)
   (import (scheme base)
           (srfi 211 primitives))
--- a/lib/srfi/211/explicit-renaming.sld
+++ b/lib/srfi/211/explicit-renaming.sld
@@ -38,33 +38,50 @@
 ;;;     whole environment is the honest static over-approximation).
 ;;;
 ;;;   * Hygiene strength equals this engine's own syntax-rules hygiene
-;;;     exactly — the documented guarantee (KEP-0018 unresolved question 6,
-;;;     decided with kaappi#2388): rename routes through the same
-;;;     renameForHygiene the template instantiator uses, and compare
-;;;     through the same binding machinery literal matching uses, so the
-;;;     two macro systems answer every keyword-check quadrant identically
+;;;     exactly for the auxiliary-keyword spellings — the documented
+;;;     guarantee (KEP-0018 unresolved question 6, decided with
+;;;     kaappi#2388): rename routes through the same renameForHygiene the
+;;;     template instantiator uses, and compare through the same binding
+;;;     machinery literal matching uses, so for reserved forms and macro
+;;;     keywords (the spellings rename keeps bare and records) and for
+;;;     gensym-marked renames of any other spelling, the two macro systems
+;;;     answer every keyword-check quadrant identically
 ;;;     (regression-pinned, quadrant for quadrant, by the KEP-0006
-;;;     four-quadrant test in tests/scheme/srfi/srfi211.scm). Shared,
-;;;     pre-existing limitations, identical on both paths: a use-site
-;;;     TOP-LEVEL redefinition of a name the macro's output references
-;;;     reaches the expansion; and a reserved-form spelling the hygiene
-;;;     engine keeps bare (else, _, ...) is shadowed by a use-site local of
-;;;     that spelling for a macro-INTRODUCED occurrence just as surely as
-;;;     for a user-typed one, while identifiers the engine CAN mark (any
-;;;     non-reserved spelling, e.g. =>) stay hygienic.
+;;;     four-quadrant test in tests/scheme/srfi/srfi211.scm). The
+;;;     guarantee does NOT extend to spellings whose bare rename comes
+;;;     from renameForHygiene's other bare-returning branches (e.g. the
+;;;     VOID sentinel for a name defined later in the use-site body):
+;;;     there compare keeps the reflexive use-token view while a
+;;;     syntax-rules literal refuses — a pre-existing divergence, pinned
+;;;     as such by the same suite. Shared, pre-existing limitations,
+;;;     identical on both paths: a use-site TOP-LEVEL redefinition of a
+;;;     name the macro's output references reaches the expansion; and a
+;;;     reserved-form spelling the hygiene engine keeps bare (else, _,
+;;;     ...) is shadowed by a use-site local of that spelling for a
+;;;     macro-INTRODUCED occurrence just as surely as for a user-typed
+;;;     one, while identifiers the engine CAN mark (any non-reserved
+;;;     spelling, e.g. =>) stay hygienic.
 ;;;
 ;;;   * compare is binding-aware (kaappi#2388): (compare x (rename 'kw))
 ;;;     answers whether the use-site identifier x and the definition-side
 ;;;     keyword denote the same binding — an unshadowed use is #t, a
 ;;;     use-site local rebinding of the spelling is #f, and a renamed
 ;;;     (macro-introduced or global-name) keyword never resolves to a
-;;;     use-site local. Symbols are interned, so a bare-rename product is
+;;;     use-site local. A rename of a name bound in the transformer's own
+;;;     library (def-env-marked outside it) compares equal to a use-site
+;;;     reference of the same exported binding, refusing only under a
+;;;     local shadow. Symbols are interned, so a bare-rename product is
 ;;;     the same object as a use-site token of that spelling; compare
-;;;     recognizes the classic (compare <token> (rename 'kw)) shape by the
-;;;     invocation's rename record, which makes the answer independent of
-;;;     argument order. Two plain use-site tokens stay reflexive unless
-;;;     this invocation also renamed that spelling bare and a use-site
-;;;     local shadows it (the pairwise input-comparison idiom).
+;;;     recognizes the classic (compare <token> (rename 'kw)) shape from
+;;;     the invocation's rename record together with whether the spelling
+;;;     occurs in the macro-use input — which makes the answer independent
+;;;     of argument order, and reflexive both for two of the invocation's
+;;;     own rename products (the hoisted-rename style) and for two plain
+;;;     use-site tokens (the pairwise input-comparison idiom). The one
+;;;     non-reflexive shape: a spelling that occurs in the input, renamed
+;;;     bare this invocation, compared under a use-site local shadow of
+;;;     it — the quadrant-2 case, where the shadowing local is a
+;;;     different binding.
 ;;;
 ;;;   * `(rename 'x)` used to BIND x when x names a global procedure
 ;;;     returns x unrenamed (reference semantics win); rename fresh names
--- a/src/compiler.zig
+++ b/src/compiler.zig
@@ -1104,6 +1104,21 @@ const SetScanBudget = struct {
     truncated: bool = false,
 };
 
+/// Per-list-level cap on cdr-spine steps in collectSetTargets (#2404).
+/// The depth cap bounds only car recursion; a datum-label cycle
+/// (`#0=(zz . #0#)`, R7RS §7.1.2) lives in the spine, where the walk used
+/// to spin forever — reached from a cyclic macro operand re-emitted into a
+/// body position. One million steps per level is far beyond any real form
+/// (the whole repo's suites stay under ~1.7k expansions per top-level
+/// form); exhausting it stops that level's scan, and on a budgeted path
+/// marks `truncated` — the same loses-optimization-never-correctness
+/// degradation the expansion budget already takes, with Part B
+/// (scanSetTargets in expandAndCompileMacroUse) catching anything missed
+/// at real-expansion time. The null-budget callers (define-syntax specs,
+/// the native backend) have no flag to set and just stop: their misses
+/// take the same correct-late path.
+const SET_SCAN_SPINE_CAP: usize = 1_000_000;
+
 /// Recursively collect the symbol names that appear as the target of a
 /// `(set! <name> ...)` anywhere in `expr` into `out`. Used to suppress
 /// constant folding of those names within the enclosing form (see
@@ -1133,7 +1148,16 @@ fn collectSetTargets(self: ?*Compiler, expr: Value, out: *std.StringHashMap(void
         return;
     }
     var cur = expr;
+    var spine_steps: usize = 0;
     while (types.isPair(cur)) {
+        // #2404: a datum-label cycle makes this spine infinite; the step
+        // cap turns it into the scan's ordinary loses-optimization
+        // truncation (see SET_SCAN_SPINE_CAP).
+        spine_steps += 1;
+        if (spine_steps > SET_SCAN_SPINE_CAP) {
+            if (budget) |b| b.truncated = true;
+            return;
+        }
         const head = types.car(cur);
         if (types.isSymbol(head)) {
             const hname = types.stripHygienicPrefix(types.symbolName(head));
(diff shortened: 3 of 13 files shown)


src/expander.zig
```diff
@@ -517,21 +517,112 @@ fn erRenameDatum(gc: *GC, v: Value) anyerror!Value {
...
+/// products, unshadowed global references.
+const ErIdentifierBinding = union(enum) {
+    free,
+    local: u32,
+    def_env: []const u8,
+};
+
+fn erIdentifierBinding(full: []const u8) ErIdentifierBinding {
+    if (std.mem.startsWith(u8, full, types.def_env_binding_prefix))
+        return .{ .def_env = full };
```
baijum, line 574: **A `def_env`-marked rename never compares equal to a bare use-site reference — with no shadowing involved.**

`erIdentifierBinding` classifies a `__kaappi_defenv__`-prefixed name as `.def_env`, and a bare use-site symbol resolves to `.free` (globals are not use-site local slots). `erBindingsAgree` then answers `#f` for the `def_env`/`free` pair unconditionally. But `renameForHygiene` takes the `active_def_env` branch for *any* name bound in the transformer's own library env that isn't a `(scheme base)` name (`expander_instantiate.zig:1013`), whenever the macro is used **outside** that library (`denv != globals`). So a library-defined ER transformer whose keyword spelling is also a library binding now refuses a use-site reference to that very same binding.

This contradicts the PR description's "everything else answers as before" — no use-site local is required, and `#t` was the *correct* free-identifier=? answer (both identifiers denote the same, imported binding).

Repro (`lib/testlib3.sld`):

```scheme
(define-library (testlib3)
  (import (scheme base) (srfi 211 explicit-renaming))
  (export in-lib? not-in-lib? lib-bound)
  (begin
    (define lib-bound 'marker)
    (define-syntax in-lib?
      (er-macro-transformer
        (lambda (form rename compare)
          (list (rename 'quote) (compare (cadr form) (rename 'lib-bound))))))
    (define-syntax not-in-lib?
      (er-macro-transformer
        (lambda (form rename compare)
          (list (rename 'quote) (compare (cadr form) (rename 'usesite-only))))))))
```

```scheme
(import (scheme base) (scheme write) (testlib3))
(define usesite-only 1)
(display (list 'lib-bound-name  (in-lib? lib-bound)
               'usesite-only-name (not-in-lib? usesite-only)))
```

```
origin/main : (lib-bound-name #t usesite-only-name #t)
this branch : (lib-bound-name #f usesite-only-name #t)
```

The second column isolates the cause: only the name that takes the `def_env` branch flips.

No shipped library hits this (241/202 only rename reserved forms and `(scheme base)` names, which skip the `def_env` branch via `lookupBaseBinding`), and no new test covers it either — every new Zig and Scheme test defines its transformer at top level/REPL, where `transformer.def_env` is `null`. A `def_env` identity probably needs to compare equal to a free reference of the same effective spelling (or at least to another reference that resolves to the same library binding), plus a library-defined-transformer regression test.

src/expander.zig
```diff
@@ -517,21 +517,112 @@ fn erRenameDatum(gc: *GC, v: Value) anyerror!Value {
...
+        // object, so "which side is the rename product" is unanswerable
+        // from the values. When this invocation renamed the spelling bare,
+        // assume the classic shape — one side the definition-site keyword,
+        // the other a use-site token — and answer whether the two sites'
+        // views agree: a use-site local rebinding of the spelling is a
+        // different binding, exactly as a syntax-rules literal refuses the
+        // same shadowing. Otherwise the arguments are two use-site tokens
+        // — the same identifier — and compare is reflexive (the pairwise
+        // input-token idiom, e.g. duplicate-key checks, keeps working).
+        if (erRenamedBareThisInvocation(eff_a))
```
baijum: **Reflexivity break: `(compare (rename 'kw) (rename 'kw))` answers `#f`.**

The bare/bare branch keys only on "was this spelling bare-renamed in this invocation", not on the *arguments*. When both arguments are the transformer's own definition-side rename product — no use-site token involved at all — the branch still consults `active_use_check.resolve` and returns `#f` under a use-site local.

```scheme
(import (scheme base) (scheme write) (srfi 211 explicit-renaming))
(define-syntax self-else?
  (er-macro-transformer
   (lambda (form rename compare)
     (list (rename 'quote) (compare (rename 'else) (rename 'else))))))
(display (list 'plain (self-else?) 'shadowed (let ((else 1)) (self-else?))))
```

```
origin/main : (plain #t shadowed #t)
this branch : (plain #t shadowed #f)
```

SRFI 211's `compare` is `free-identifier=?`, which has to be an equivalence relation; `(compare x x)` on one and the same identifier must be `#t` for every `x`. This matters for the `(define r-else (rename 'else))` style the shipped ports use (241/202 hoist all their renames into `r-*` bindings up front): any helper that ends up comparing a hoisted rename against another rename of the same keyword — a keyword-set membership test, a normalization pass over an already-renamed sub-form — silently gets `#f` under a use-site shadow.

The `matchPattern` literal branch this is said to mirror is never asked this question (its two sides are structurally def-site and use-site), so the mirroring argument doesn't cover it. A cheap guard would be to return `#t` up front when `args[0] == args[1]` (they are the same interned object), before any of the classification below.

lib/srfi/211/explicit-renaming.sld
```diff
@@ -35,22 +38,43 @@
...
-;;;     references (e.g. redefining a library helper's name after import)
-;;;     reaches the expansion, for procedural and syntax-rules macros
-;;;     alike. compare is name-based (hygiene-stripped equality), which
-;;;     answers the classic literal checks (else, =>) correctly but cannot
-;;;     distinguish two same-named bindings.
+;;;     exactly — the documented guarantee (KEP-0018 unresolved question 6,
+;;;     decided with kaappi#2388): rename routes through the same
+;;;     renameForHygiene the template instantiator uses, and compare
+;;;     through the same binding machinery literal matching uses, so the
+;;;     two macro systems answer every keyword-check quadrant identically
```
baijum: **"the two macro systems answer every keyword-check quadrant identically" is stated unconditionally, but the parity holds only for spellings `erRenameSymbol` classifies as `keeps_name`.**

`erRenameSymbol` records the identity entry (the thing `erRenamedBareThisInvocation` reads) only on the `isTemplateReserved or er_macros.contains` path. `renameForHygiene` has *other* branches that also return a bare symbol and record nothing — notably the VOID-sentinel branch for a name that is a later internal define in the use-site body (`expander_instantiate.zig:1073`). For those spellings `compare` takes the `return types.TRUE` fall-through and ignores use-site binding entirely, while a `syntax-rules` literal of the same spelling refuses in the same position:

```scheme
(import (scheme base) (scheme write) (srfi 211 explicit-renaming))
(define-syntax is-later-er?
  (er-macro-transformer
   (lambda (form rename compare)
     (list (rename 'quote) (compare (cadr form) (rename 'laterdef))))))
(define-syntax is-later-sr?
  (syntax-rules (laterdef) ((_ laterdef) #t) ((_ x) #f)))
(define (probe)
  (define r (list (list 'er (is-later-er? laterdef)
                        (let ((laterdef 1)) (is-later-er? laterdef)))
                  (list 'sr (is-later-sr? laterdef)
                        (let ((laterdef 1)) (is-later-sr? laterdef)))))
  (define laterdef 9)
  r)
(display (probe))
```

```
both builds: ((er #t #t) (sr #f #f))
```

The divergence is pre-existing, so this is not a regression — but this PR is what promotes the parity claim from an aspiration to a documented guarantee ("pinned as a contract", KEP-0018 UQ6) in this header, in `srfi-implementation-notes.md`, and in the new `srfi211.scm` suite. Either qualify the guarantee to the reserved/macro-keyword spellings it actually covers, or extend the identity-entry recording to every branch where `rename` hands back a bare spelling.

docs/dev/srfi-implementation-notes.md
```diff
@@ -1479,9 +1497,14 @@ check):
   per-expansion counter, then renamed.
 - **Keyword recognition (`...`, `_`, `->`, `guard`, `unquote`,
   `quasiquote`, `values`) is `compare` against `rename` of the keyword —
-  hygiene-stripped name equality.** That strength suffices for these
-  macros (kaappi#2388 records the evidence); a use inside another
-  er-macro's output whose keywords arrive renamed still compares equal.
+  binding-aware free-identifier=? since kaappi#2388 (name-stripped
```
baijum, line 1517: This bullet is updated, but the three in-tree copies of the same claim — in the libraries the bullet is *about* — still describe the old semantics and cite this very issue number:

- `lib/srfi/241.sld:32-34` — "recognized by hygiene-stripped name comparison (the strength of this engine's ER compare — see kaappi#2388)"
- `lib/srfi/241.sld:263` — `;; Keyword recognition is name-based compare (kaappi#2388).`
- `lib/srfi/202.sld:41` — `;; Keyword recognition is name-based compare (kaappi#2388).`

After this PR `kaappi#2388` means the opposite of what those comments say, so a reader who greps for the issue number lands on three statements that contradict the two that were updated. Worth flipping them in the same commit.

`241.sld`'s follow-on sentence needs a second look too: "exporting bindings for [the auxiliary keywords] would not change what is matched" was reasoning from name equality. It still happens to hold (an exported binding is a global, and globals are not use-site local slots), but it is no longer self-evident from the stated mechanism — either drop it or say why it survives.

src/expander.zig
baijum: Fixed in 030cc6a8, exactly along the suggested line: a `def_env`-marked rename now agrees with a bare reference when the bare side is not locally shadowed and the spelling resolves in the use-site globals (i.e. the library was imported and planted the binding) — so your `in-lib?` repro is back to `#t` unshadowed and `#f` under a local shadow. Renamed (gensym) sides still never agree with a def-env name. Regression-tested with a library-defined transformer added to srfi211.scm's `t211 helperlib` (`is-lib-bound?`), which is also the library-defined-transformer coverage you noted was missing. Thanks for the isolated two-column repro — it made the diagnosis immediate.

src/expander.zig
baijum: Fixed in 030cc6a8, and you were right that reflexivity must hold — but the cheap `args[0] == args[1]` guard cannot work here, for the reason at the heart of this PR: the use-site token and the bare rename product are the *same interned object*, so quadrant 2's pair `(compare <input-else> (rename 'else))` is also args-equal, and an up-front `#t` would reintroduce exactly the shadowing failure #2388 exists to fix. The discriminator that distinguishes your repro from quadrant 2 is the macro-use input: in `(self-else?)` the spelling `else` occurs nowhere in the input, so both arguments must be the invocation's own rename products (reflexive `#t`), while in `(er-is-else else)` it occurs as a use-site token (quadrant rule applies). `erCompareFn` now walks the input for the bare/bare-renamed shape only — read-only, reached through a pointer to `expandProceduralMacro`'s rooted slot so a moving GC updates it. Your repro now prints `(plain #t shadowed #t)`; quadrant 2 and the parity suite are unchanged; both pinned in `srfi211.scm` (`er-self-else?`, `er-self-arrow?`).

lib/srfi/211/explicit-renaming.sld
baijum: Went with the first option — the guarantee is now stated for the spellings it actually covers (reserved forms and macro keywords via the identity-entry path, plus gensym-marked renames of any other spelling via the classification path), in this header, in `srfi-implementation-notes.md`, and in the parity suite's own header comment. Your `laterdef` divergence is pinned as the boundary in `srfi211.scm` (`er-is-laterdef?` → `(#t #t)`, with a comment saying so), so the qualified claim is enforced, not just asserted. I deliberately did not extend the identity-entry recording into `renameForHygiene`'s other bare-returning branches: matching the literal's exact `BOUND_PENDING`/prescan answer there means resolving the name at rename time with `parseSyntaxRules`'s `extra_bound` logic, which drags definition-site machinery through a function shared with syntax-rules templates — out of scope for this PR, and no shipped library renames such spellings. If it ever matters, that would be its own change (and your probe is the ready-made test).

docs/dev/srfi-implementation-notes.md
baijum: All three flipped in 030cc6a8 (both `Keyword recognition is ... compare` comments now say binding-aware, and 241's limitation bullet describes free-identifier=? compare). For the follow-on sentence I kept it with the re-reasoning you asked for: an exported binding is planted as a *global*, and globals are not use-site local slots — so the use-site token of that spelling still classifies free and compares equal; only a *local* rebinding now opts out of keyword-hood. That reasoning is now written into the bullet itself so it stays self-evident from the stated mechanism.

src/expander.zig
```diff
@@ -517,21 +531,172 @@ fn erRenameDatum(gc: *GC, v: Value) anyerror!Value {
...
+/// (pairs, dotted tails, vectors)? Read-only and allocation-free, so it is
+/// safe from compare at any point of the transformer call. Used only for
+/// bare spellings this invocation also renamed: occurrence in the input
+/// is what separates a use-site token from the invocation's own rename
+/// products when the two are representationally identical (interned).
+fn erFormMentionsSymbol(form: Value, sym: Value) bool {
+    if (types.isSymbol(form)) return form == sym;
+    if (types.isPair(form)) {
+        var cur = form;
+        while (types.isPair(cur)) : (cur = types.cdr(cur)) {
```
baijum, line 644: **New hang: `erFormMentionsSymbol` has no cycle detection, and the macro-use input can be genuinely circular.**

R7RS datum labels (§7.1.2) are supported and produce a real cycle, not a copy:

```scheme
(define x '#0=(a . #0#))
(write (list (car x) (car (cdr x)) (eq? x (cdr x))))   ;=> (a a #t)
```

So `while (types.isPair(cur)) : (cur = types.cdr(cur))` never terminates once the walk reaches such a tail. Before this commit `compare` never touched the input, so this is a regression introduced here.

Repro — the searched spelling must be absent from the input so the walk can't short-circuit before reaching the cycle:

```scheme
(import (scheme base) (scheme write) (srfi 211 explicit-renaming))
(define-syntax self-else?
  (er-macro-transformer
   (lambda (form rename compare)
     (list (rename 'quote) (compare (rename 'else) (rename 'else))))))
(display "before") (newline)
(display (self-else? #0=(zz . #0#)))
```

```
origin/main  : prints `before` then `#t`, exits 0 in <1s
this commit  : prints `before`, then spins forever (killed by SIGALRM at 20s, rc=142)
```

It spins inside the expander, not at a VM safepoint, so `--time-limit` / interrupt machinery won't reach it either. Per `docs/dev/github-issues.md` a hang reachable from a legal program is `priority: high`.

A depth cap or a visited-set is needed — or, since the walk only ever answers "does this one interned symbol occur here", bound it the way the reader bounds nesting (`KP1009`) and treat exhaustion as "not found". Note the recursion on `car` is separately unbounded, though the reader's own nesting cap makes that unreachable from source (checked at depth 50000 → `KP1009` on both builds).

src/expander.zig
```diff
@@ -517,21 +531,172 @@ fn erRenameDatum(gc: *GC, v: Value) anyerror!Value {
...
+        // a use-site local rebinding of the spelling is a different
+        // binding, exactly as a syntax-rules literal refuses the same
+        // shadowing. When the spelling does NOT occur in the input, both
+        // arguments are the invocation's own rename products — the same
+        // identifier — and compare is reflexive. With no bare rename of
+        // the spelling at all, the arguments are two plain use-site
+        // tokens, likewise reflexive (the pairwise input-token idiom,
+        // e.g. duplicate-key checks, keeps working).
+        if (erRenamedBareThisInvocation(eff_a)) {
+            const in_input = if (er_input_slot) |slot| erFormMentionsSymbol(slot.*, args[0]) else false;
```
baijum: The occurrence test fixes the self-compare only when the spelling is *absent* from the input — which is the case the motivating idiom does not hit.

```scheme
(define-syntax self-else?
  (er-macro-transformer
   (lambda (form rename compare)
     (list (rename 'quote) (compare (rename 'else) (rename 'else))))))
(display (list 'no-mention/plain    (self-else?)
               'no-mention/shadowed (let ((else 1)) (self-else?))
               'mention/plain       (self-else? else)
               'mention/shadowed    (let ((else 1)) (self-else? else))))
```

```
this commit: (no-mention/plain #t no-mention/shadowed #t mention/plain #t mention/shadowed #f)
```

The last column is still a reflexivity break: neither argument is a use-site token — both are this invocation's own `rename` products — yet `compare` answers `#f`.

The commit message motivates the fix as protecting "the hoisted-rename style the 241/202 ports use". But a matcher checks keywords precisely because they appear in the form it was handed, so in 241/202 the spelling is in the input essentially whenever the check matters, and this branch stays on the non-reflexive path. Concretely, `241.sld`'s `(kw? x 'guard)` under `(let ((guard …)) (match …(,a (guard …))…))` has `guard` in the input; any helper there that ends up comparing two rename products of `guard` still gets `#f`.

Two things worth separating in the write-up: the `.sld` header calls the surviving shape "the quadrant-2 case", but quadrant 2 is `(compare <use-token> (rename 'kw))` — here no use-site token participates, and the answer is wrong rather than deliberately refused. Either say that plainly, or reach for a discriminator that actually distinguishes the two (e.g. having `rename` hand back a distinguishable wrapper for bare products, rather than inferring provenance from the input).

src/expander.zig
```diff
@@ -517,21 +531,172 @@ fn erRenameDatum(gc: *GC, v: Value) anyerror!Value {
...
+/// review: an exported library binding must still compare equal to a
+/// use-site reference of that spelling; a shadowing local, or a spelling
+/// the use site never imported, is a different binding or none).
+fn erDefEnvAgreesWithBare(bare_full: []const u8) bool {
+    if (active_use_check.resolve(bare_full) != LITERAL_UNBOUND) return false;
+    const g = er_globals orelse return false;
+    const gmod = @import("globals.zig");
+    const glk = gmod.acquireGlobalsRead(g);
+    defer gmod.releaseGlobalsRead(glk);
+    return g.contains(bare_full);
```
baijum, line 598: `g.contains(bare_full)` tests only that *some* global of that spelling exists at the use site, so the doc comment's inference — "the library was imported, planting the binding in the use-site globals" — doesn't hold. A use site that never imported the name and has an unrelated global of its own answers `#t`:

```scheme
(define-library (testlib4)
  (import (scheme base) (srfi 211 explicit-renaming))
  (export same-as-libname?)
  (begin
    (define lib-thing 'library-binding)
    (define-syntax same-as-libname?
      (er-macro-transformer
       (lambda (form rename compare)
         (list (rename 'quote) (compare (cadr form) (rename 'lib-thing))))))))
```

```scheme
(import (scheme base) (only (testlib4) same-as-libname?))
(define lib-thing 'totally-different-binding)   ; never imported lib-thing
(display (same-as-libname? lib-thing))          ;=> #t, should be #f
```

Two genuinely different bindings, `#t`. This is not a regression — the pre-PR name equality answered `#t` too — so it is fine to ship, but it is the same class of over-approximation the header already owns for `rename` ("the whole environment is the honest static over-approximation"). Worth saying so here rather than asserting an import that may not have happened; otherwise the next reader takes the comment at face value.

tests/scheme/srfi/srfi211.scm
```diff
@@ -191,19 +355,33 @@
...
+      (er-macro-transformer
+       (lambda (form rename compare)
+         (list (rename 'quote) (compare (cadr form) (rename 'lib-bound-var))))))
     (define-syntax combo
       (syntax-rules ()
         ((_ e) (list (lib-twice e) e))))))
 (import (t211 helperlib))
 (test-equal 42 (lib-twice 21))
 (test-equal '(20 10) (combo 10))
+(test-assert "a def-env-marked rename compares equal to the use-site reference of the same exported binding"
```
baijum, line 381: The label says "the same **exported** binding", but `lib-bound-var` is not in `(export lib-twice combo is-lib-bound?)` — it is a private library binding. The assertion still passes, and for a real reason: `copyTransformerFreeRefs` plants the transformer's free references into the use-site globals at import time, which is exactly the path the header's "a library's own non-exported helper at the use site" note describes. Verified:

```scheme
(import (t211b helperlib))            ; only is-lib-bound? exported
(display (is-lib-bound? lib-bound-var))  ;=> #t
(write lib-bound-var)                    ;=> marker   <- reachable, though unexported
```

So the test pins the non-exported/free-ref path, not the exported-binding path that motivated the fix. Either rename the binding to something that is actually exported (which would then cover the case the review raised), or keep it and relabel — as written, a later reader will take it as coverage for exports it doesn't provide.

src/expander.zig
baijum: Filed as #2404 (`priority: high`), so it is tracked independently of this PR.

While confirming the `origin/main` baseline for it I hit a second, more severe instance of the same assumption that is already shipping: `(rename <circular-datum>)` recurses with two `pushRoot`s per level until the root stack panics — an uncatchable process abort from five lines of ordinary R7RS. Filed as #2403 (`priority: critical`). Both come from ER helpers walking the macro-use input as if it were acyclic, so one cycle-aware traversal would close both.

baijum: Fixed in 5160768c — good catch, and the repro (absent spelling, so the walk can't short-circuit) was exactly right. The walk now carries a node budget (1M) and a depth cap (10k), both non-allocating; exhausting either counts conservatively as *occurs*, so the quadrant rule applies and compare never wrongly accepts. Your repro now terminates with `#t` (matching origin/main), pinned in srfi211.scm as a define-initializer use. Two notes on the shape: (1) the test deliberately avoids `test-assert` as the wrapper — your #2404 verification led me to sample the suite hang and find a **third member of this class, pre-existing on origin/main**: `compiler.collectSetTargets` hangs the same way on a cyclic macro operand re-emitted into a body position (`(test-assert "..." (m #0=(zz . #0#)))` spins there on the main binary); details commented on #2404. (2) #2403 (`erRenameDatum` root-stack abort) I left to its own issue deliberately — unlike the walk, `(rename <cycle>)` has no finite answer, so the fix needs a semantic decision (a catchable syntax error, presumably) rather than a bound.

src/expander.zig
baijum: You're right on all three counts — the occurrence test protects only the absent-spelling case, the motivating hoisted-rename idiom in 241/202 does co-occur with the spelling being in the input, and calling the surviving shape "the quadrant-2 case" blurred a real distinction. 5160768c takes the say-it-plainly option: the .sld header, the implementation notes, and erCompareFn's doc now all state that a spelling that occurs in the input AND was bare-renamed this invocation, compared under a use-site shadow, gets the demanded refusal when one argument is that input token and a **known-wrong (broken reflexivity)** answer when both arguments were the invocation's own rename products. On the wrapper alternative: I agree it would be the real fix, but a distinguishable representation for bare rename products is not survivable here — `(rename 'else)`'s product flows into macro *output* (an er transformer building a `cond` clause emits it), and the built-in clause processors and the expander match reserved forms by bare spelling; every consumer would need to learn the wrapper, which is the deep-expander change #2388 deliberately scoped out. If that ever becomes worth doing, the parity suite now pins the exact semantics to migrate.

src/expander.zig
baijum: Fair — reworded in 5160768c. The comment now says what the check actually establishes: "the use-site globals hold some binding of this spelling", explicitly named as the same class of over-approximation the whole-def-env import copy makes for rename, with your testlib4 case called out (unrelated same-named global answers #t) and the shadowing-local exclusion kept exact. Verified your probe against the fixed build: #t, i.e. unchanged from origin/main, now as a documented fact rather than an accidental one.

tests/scheme/srfi/srfi211.scm
baijum: Good catch on the label — `lib-bound-var` is now actually exported from `(t211 helperlib)` (5160768c), so the test covers the exported-binding path it claims; the unexported free-ref planting path that the assertion was really exercising stays covered by the neighboring `lib-twice` test, which is the one the header's non-exported-helper note describes. Thanks also for verifying the unexported case reaches the value (`marker`) — that confirmed the fix works through both import routes.

src/expander.zig
baijum: Acknowledged both filings. The #2404 instance introduced by this PR is fixed in 5160768c (node budget + depth cap, exhaustion = conservative occurrence; repro and termination pinned in srfi211.scm). While re-running the suite I also hit a third, pre-existing member of the class and left the details on #2404: `compiler.collectSetTargets` spins the same way on a cyclic macro operand in a body position, on origin/main. #2403 stays untouched here on purpose — `(rename <cycle>)` has no finite renamed copy, so it needs a semantic decision (catchable syntax error?) rather than a bound, which belongs with that issue.

baijum: Per the maintainer's call we're fixing the `collectSetTargets` instance in this PR too — fce810f4. Both spine loops (main walk + let-syntax bindings) carry a 1M-step cap per list level; exhaustion on a budgeted path marks the scan truncated, which is the scanner's designed loses-optimization-never-correctness degradation (set_targets_all boxing, Part B correct-late), and null-budget callers just stop. Your repro `(test-assert "..." (m #0=(zz . #0#)))` now passes; the ellipsis-pattern variant terminates at matchEllipsis's MAX_ELLIPSIS_VALUES cap (probed, no change needed). New regression suite: tests/scheme/hygiene/cyclic-macro-input-2404.scm. So #2404 is now closed entirely by this PR (erFormMentionsSymbol in 5160768c, collectSetTargets here); #2403 remains separate on purpose — rename on an unfoldable cycle has no finite answer, so that one is a semantic decision, not a bound.

src/compiler.zig
```diff
@@ -1235,7 +1259,13 @@ fn collectSetTargets(self: ?*Compiler, expr: Value, out: *std.StringHashMap(void
                     const rest = types.cdr(cur);
                     if (!types.isPair(rest)) return;
                     var bindings_cur = types.car(rest);
+                    var binding_steps: usize = 0;
                     while (types.isPair(bindings_cur)) {
+                        binding_steps += 1;
+                        if (binding_steps > SET_SCAN_SPINE_CAP) {
+                            budget.?.truncated = true;
```
baijum, line 1290: `budget.?` is an unconditional unwrap, two lines after the sibling cap in the same function correctly writes `if (budget) |b| b.truncated = true;`. The asymmetry looks unintended, and `budget` is genuinely null on this path:

- the `define-syntax` branch above recurses into the spec with `null` — and a `let-syntax` inside a `define-syntax` template is ordinary code (SRFI 241 emits exactly that);
- this branch recurses into each binding spec with `null`;
- `scanSetTargetsWithoutMacros` (the native backend) passes `null`.

So whenever this cap fires under a null budget, the walk panics on a null optional instead of taking the "stops, correct-late" path the doc comment promises — turning this commit's own degrade-never-panic contract into the failure mode it was written to remove.

I could not reach it on this build, and said so rather than claiming a repro: the cyclic route hangs earlier in `collectFreeRefsWithLocals` (see the other comment), and the >1,000,000-literal-bindings route trips the source size limit (`File too large` at a 1.9 MB input). Latent, not live — but a one-line fix:

```suggestion
                            if (budget) |b| b.truncated = true;
```

tests/scheme/hygiene/cyclic-macro-input-2404.scm
```diff
@@ -0,0 +1,52 @@
+;; Regression test for kaappi#2404: walkers that touch a macro-use input
+;; as if it were acyclic spin forever on the genuine cycles R7RS datum
+;; labels (§7.1.2) produce — `(eq? x (cdr x))` is #t for
+;; '#0=(zz . #0#), so this is a real cycle, not a reader artifact.
+;;
+;; Two walkers are exercised:
```
baijum, line 6: Both walkers here are genuinely fixed — verified: my #2404 repro now prints `#t` and exits 0, this file passes in <1s, and the bound design is sound (the depth cap of 10,000 sits well above the reader's `MAX_NESTING_DEPTH = 1024`, so only the node budget is reachable from source; 200 `compare` calls over a cyclic input cost <1s).

But "Fix #2404 **completely**" and `Closes #2404` overstate it — the class has at least two more members, both pre-existing on `main` and both reachable from ordinary source.

**1. `compiler_macro.collectFreeRefsWithLocals`.** Three lines, and the macro is never even used — defining it is enough:

```scheme
(define-syntax m (syntax-rules () ((_) (quote #0=(a . #0#)))))
(display "ok")
```

`rc=142` (SIGALRM) on both `origin/main` and `fce810f4`. `sample` puts it squarely in the same class:

```
compiler_define_syntax.compileDefineSyntax   compiler_define_syntax.zig:84
  compiler_define_syntax.finalizeTransformer  compiler_define_syntax.zig:354
    compiler_macro.collectFreeRefsWithLocals  compiler_macro.zig:782   <- spins
```

**2. `compiler_lambda.compileBodyForms` / `compileExprSequence`.** A cyclic `let-syntax` bindings list in a function body:

```scheme
(define (f) (let-syntax #0=((a (syntax-rules () ((_) 1))) . #0#) 1))
```

`rc=142` on both builds. `collectSetTargets` now terminates on this input (the new cap does its job); the spin has simply moved to the body compiler behind it. Interestingly `kaappi compile` on the same file succeeds — this is an interpreter-tier hang.

Neither is a regression and neither should block this PR. But `Closes #2404` will retire the issue with the class still open, and the file header's framing ("walkers that touch a macro-use input as if it were acyclic") is exactly the right frame for a class that isn't closed yet. Either narrow the closing keyword to the two sites actually fixed and file the rest, or keep #2404 open as the class tracker and reference it. Same for `#2403`'s note here — that one is correctly scoped as its own semantic decision.

src/compiler.zig
```diff
@@ -1104,6 +1104,21 @@ const SetScanBudget = struct {
...
+/// body position. One million steps per level is far beyond any real form
+/// (the whole repo's suites stay under ~1.7k expansions per top-level
+/// form); exhausting it stops that level's scan, and on a budgeted path
+/// marks `truncated` — the same loses-optimization-never-correctness
+/// degradation the expansion budget already takes, with Part B
+/// (scanSetTargets in expandAndCompileMacroUse) catching anything missed
+/// at real-expansion time. The null-budget callers (define-syntax specs,
+/// the native backend) have no flag to set and just stop: their misses
+/// take the same correct-late path.
+const SET_SCAN_SPINE_CAP: usize = 1_000_000;
```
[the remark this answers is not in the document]
baijum: Valid finding, fixed in bd889ba3 — and it was bigger than the spine cap: the **depth cap had the identical exposure on null-budget paths**, pre-existing, so the fix closes the class rather than the instance. `SetScanBudget` gained an `expand` flag so a non-null budget no longer implies "speculatively expanding"; structure-only scans walk literally (the define-syntax/let-syntax shortcuts are outcome-equivalent there) and now *report* depth/spine truncation. Part B (`Compiler.scanSetTargets`) propagates to `set_targets_all` — the pre-scan's own every-name-is-a-target conservatism. For `scanSetTargetsWithoutMacros` the return value now carries the signal and the LLVM read loop eval-fallbacks the whole form via `makePassthrough` (the tier has no `set_targets_all` switch; a passthrough form is compiled by the full VM compiler with its own conservative machinery — the same action #2119 takes for continuation-capturing forms). One deliberate exception, documented: the define-syntax/let-syntax **spec** walks keep a non-propagating structure-only budget — a `set!` that only materializes when the spec's own macros run is caught at the macro's real use site, the pre-existing correct-late path. Tests in `tests_prescan.zig`: truncation reported on a cyclic form, not hair-triggered on an ordinary one, and a wrapper-shape cyclic operand (verified as the minimal `origin/main` hang — a plain top-level use never reaches the scan) keeps set! boxing correct under continuation capture.

src/compiler.zig
```diff
@@ -1147,10 +1189,16 @@ fn collectSetTargets(self: ?*Compiler, expr: Value, out: *std.StringHashMap(void
...
             } else if (budget) |b| {
-                // Only a budgeted scan expands, and only a budgeted scan is
-                // ever handed a Compiler — so `self.?` below is reached only
-                // when `maybe_macro` was non-null, which requires one.
-                const maybe_macro = if (self) |c| c.lookupMacro(types.symbolName(head)) else null;
+                // Only an expanding scan looks macros up, and only an
+                // expanding scan is ever handed a Compiler — so `self.?`
+                // below is reached only when `maybe_macro` was non-null,
+                // which requires one. Structure-only scans (b.expand ==
+                // false) never take this branch and keep walking the form
```
baijum, line 1196: Nit, comment-only: both claims here became false with this commit, and the second one inverts what the code now does.

- "Structure-only scans … never take this branch" — they do. The branch is `else if (budget) |b|`, and Part B and `scanSetTargetsWithoutMacros` now pass a non-null budget (`.expand = false`). What they skip is the macro *lookup*, not the branch.
- "only an expanding scan is ever handed a Compiler" — `Compiler.scanSetTargets` passes `self` with `expand = false`. Harmless, because the new `self orelse break :blk null` guards it, but the stated invariant no longer holds.

The consequence is worth spelling out where a reader will look for it: taking the branch means these two callers now reach the `define-syntax` / `let-syntax` shortcuts, which they did not when they passed `null` and walked generically. I checked and agree that is outcome-equivalent-or-better — the `define-syntax` arm still walks the spec (`cdr(rest)`), and the `let-syntax` arm skipping binding *names* is strictly safer than the old generic walk (a binding named `quote` used to derail it). So the behavior is right; only the comment says otherwise.

Something like: "Structure-only scans (`b.expand == false`) take this branch but never look a macro up, so they walk the form literally and fall through to the `define-syntax`/`let-syntax` shortcuts below."

baijum:
LGTM. Checked `bd889ba3` on a fresh build; everything substantive is verified and I have no blocking concerns.

**The truncation-propagation change is the right fix, and better than the one-line guard I suggested.** A silently-partial target set leaving a `set!`-ed local unboxed is a real correctness exposure, not a lost optimization, and closing it on *every* caller — rather than only where a budget happened to be non-null — is the correct shape. Two details I specifically checked and agree with:

- `set_targets_all` genuinely reaches both consumers (fold suppression in `compiler.zig:410`/`528`, `ir.isRedefined` at `ir.zig:388`), so Part B's propagation is conservative in the right direction. The `set_targets_all = false` at `:608` is teardown inside the `defer`, not a clobber of a Part-B-set flag.
- The native tier has no `set_targets_all` switch, so `makePassthrough` really is the analogous action — same one #2119 takes.

**Verification on `bd889ba3`:**

| check | result |
|---|---|
| `zig build test` | 1794 passed, 7 skipped, **0 failed** (+2 = the new `tests_prescan` cases) |
| `tests/scheme/compile/*.sh` (native tier) | **30/30 pass** — incl. `native-macro-set-tracking-2212`, `native-set-capture-divergence-1422`, `native-boxed-shadow-divergence-1584`, `native-rebind-stale-call-822`, `native-param-shadow-fold-790` |
| `cyclic-macro-input-2404.scm` | ok, <1s |
| srfi211 / srfi241 / srfi202 / smallbatch-audit / quote-shadow-boxing | 55 / 54 / 31 / 236 / 2, all pass |
| no eval-fallback hair-trigger | `#1775: ordinary code does not trigger the pre-scan fallback` passes; ordinary program compiles native and runs identically to `main` |
| every earlier review probe | answers unchanged |

**Three open items, none blocking — happy for these to land as follow-ups:**

1. https://github.com/kaappi/kaappi/pull/2401#discussion_r3879471641 — `budget.?` at `compiler.zig:1290` is still an unconditional unwrap where the sibling caps use `if (budget) |b|`. The two recursions at `:1263` and `:1295` still pass `null`, so the null path is unchanged. Latent (I could not reach it: the cyclic route hangs earlier in `collectFreeRefsWithLocals`, the >1M-literal route trips the 1.9 MB source-size limit), but it is the one place left where this scan can panic instead of degrading.
2. https://github.com/kaappi/kaappi/pull/2401#discussion_r3879911341 — comment-only: the `expand` flag made both claims in the `else if (budget)` comment false, and the behavior change it hides (these callers now reach the `define-syntax`/`let-syntax` shortcuts) is the thing a reader would want stated. Behavior itself is fine — I checked it is outcome-equivalent or better.
3. https://github.com/kaappi/kaappi/pull/2401#discussion_r3879473688 — `Closes #2404` will retire the issue with two pre-existing members of the class still open, both hanging on `main` today: `collectFreeRefsWithLocals` (three lines, macro never used) and `compileBodyForms` on a cyclic `let-syntax` bindings list. Worth either narrowing the closing keyword or keeping #2404 as the class tracker.

On merge order: #2403 is independent — I verified its repro is byte-identical on `main` and on this branch, and the new `compare` walk is allocation- and root-free, so this adds no second path into that abort. No need to sequence them.

src/native_compiler.zig
```diff
@@ -244,8 +244,17 @@ pub fn emitLlvmFile(vm: *vm_mod.VM, path: []const u8, output_path: ?[]const u8)
...
+        // or spine cap — e.g. a datum-label cycle in the form) means the
+        // partial map cannot gate folding or inline dispatch either, and
+        // this tier has no set_targets_all switch: eval-fallback the whole
+        // form (same action as #2119's continuation forms), where the full
+        // compiler's own conservative machinery applies (#2401 review).
+        if (try compiler.scanSetTargetsWithoutMacros(expr, &redefined_names)) {
+            const passthrough_node = ir_instance.makePassthrough(expr) catch continue;
+            ir_nodes.append(allocator, passthrough_node) catch continue;
+            continue;
+        }
```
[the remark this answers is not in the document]
baijum: Right again — fixed in c1e4e735 with your first option: from the first truncated scan to the end of the file, every top-level form is a passthrough (VM-evaluated). Rationale for whole-rest-of-program rather than a propagated redefinition state: the truncated form is about to be VM-executed, so its rebindings happen at run time and no static map can enumerate them; execution order is preserved, so forms lowered *before* the truncation stay temporally correct; and only pathological inputs reach the caps (structure-only scan of raw source: >256 raw nesting or a datum-label cycle), so ordinary files keep their native lowering. The regression you asked for is in: `native-truncated-scan-fallback-2404.sh` case 1 is exactly your scenario — cyclic `wrap` operand, then a macro use expanding to `(set! + *)`, then `(+ 5 2)` — and both tiers print 10 (before the fix the native binary printed 7). Two controls pin engagement-gating and post-truncation correctness. One test-shape note: the wrapper template yields void (`(begin e (if #f #f))`) so the test doesn't depend on top-level value echoing, which the interpreter CLI does and native programs don't — a pre-existing tier difference unrelated to this change.

Issue #1489, opened 2026-07-13
Lost cross-thread wakeup: local sibling send+receive during the SharedChannelPoll drive disarms the notifier registration before park (permanent hang)


baijum:
## Summary

A fiber receiving on a promoted (shared) channel can park **permanently** even though a live peer thread later sends: a local sibling fiber's `channel-send` + `channel-receive` pair, executed during the receiver's `SharedChannelPoll` drive, consumes the receiver's one-shot notifier registration — and nothing re-registers it before the park. The later remote send finds `recv_waiters` empty and rings nothing. The parked fiber's own `shared_waiters` registry entry keeps `hasRunnableFibers()` true, so the deadlock detector is (by design) suppressed, and the process hangs silently forever.

Confirmed by repro on `main` (e4510f96) and on PR #1486's head (95c6d752). Found by triaging [CodeRabbit's inline comment on PR #1486](https://github.com/kaappi/kaappi/pull/1486) (anchored there only because that PR's stale base re-showed #1485's diff — the code is #1485's, already merged).

## Mechanism

`channelReceiveShared` (`src/primitives_fiber.zig`), per loop iteration:

1. `shared_channel.receive()` → empty → **registers this scheduler's notifier** in `sc.recv_waiters` (one-shot: every send `snapshotAndClearRecvWaiters`s the list) → `.would_park`.
2. `SharedChannelPoll` drive runs local siblings. A sibling sends: `send()` **clears `recv_waiters` and rings** our own notifier (`wake_pending = true`). The same sibling (or another) then receives, draining the value. Inside the drive, `parkOnReactor`'s consume-protocol swallows `wake_pending` and sweeps an **empty** `shared_waiters` registry (the victim isn't enrolled yet).
3. Fall-through check is `sc.peekReady()` — a plain emptiness peek that **does not re-register**. Channel empty → proceed to park.
4. `sharedWakeupPossible()` is true (peer thread holds a stub) → enroll + park. `recv_waiters` is empty; `wake_pending` is false.
5. Peer sends: `snapshotAndClearRecvWaiters` finds nothing, rings nothing. The receiver blocks in `reactor.poll()` forever; `hasRunnableFibers()` stays true (the registry entry), so `parkOnReactor` never returns the genuine-deadlock `false`.

The `if (sc.peekReady()) continue;` path is fine (it loops back through `receive()`, which re-registers). Only the **empty-after-drive → park** path loses the registration.

## Repro

[...]

PR #2402, opened 2026-08-28
Add the kaappi-shared-channels cond-expand feature identifier (KEP-0004 Phase 2)

baijum:
KEP-0004 Phase 2 was blocked on KEP-0002's cross-thread wakeup landing on `main` **with its review findings resolved**. Both halves have been true for six weeks — [#1487](https://github.com/kaappi/kaappi/issues/1487) and [#1489](https://github.com/kaappi/kaappi/issues/1489) closed 2026-07-13/14 and were fixed in v0.15.0, and KEP-0002 as a whole shipped in v0.15.0/v0.16.0 — but the identifier itself never landed. This adds it.

## Why the non-wasm branch

`kaappi-shared-channels` joins `kaappi-threads` in the non-wasm suffix of `platform_features`, not the always-on base array next to `kaappi-fibers`/`kaappi-reactor`/`kaappi-diagnostics`.

Cross-thread channel promotion requires OS threads, which wasm32-wasi does not have. KEP-0002 §5 spells out the consequence: on WASI the reactor's `notify` is a no-op and nothing ever calls it, and its cross-platform section states no promotion ever happens there — channels stay fiber-local. Advertising the identifier on a target where the capability cannot occur is exactly the failure mode KEP-0004 exists to prevent.

## Changes

- `src/types.zig` — the identifier, plus a doc comment recording the gate and the KEP-0002 §5 rationale for the platform split.
- `src/tests_advanced.zig` — a `cond-expand` test asserting `yes` natively and `no` on wasm, mirroring the existing `kaappi-threads` test (the unit suite builds for wasm since #2153, so both answers are assertable).
- `src/tests_libraries.zig` — the `srfiFeatureNumber` parser test confirms the new identifier never parses as an SRFI id.
- `tests/scheme/smoke/features-consistency.scm` — Scheme-level parity per #1177: `(features)` membership and expression-level `cond-expand`.
- `docs/dev/features.md` — the example JSON matches real output again.

## Verification

- `zig build test` — 13/13 steps, 1877/1884 passed (7 skipped, 0 failed).
- `zig build wasm` — cross-compiles clean, so the comptime branch holds on the target that omits the identifier.
- `tests/scheme/errors/features.sh` — 45/45, including a derived-from-the-binary `cond-expand recognizes 'kaappi-shared-channels'`. This suite is what keeps `kaappi features`, `(features)`, and `cond-expand` from drifting, so it is the load-bearing check here.
- `features-consistency.scm` — 16 passes, exit 0.

Companion PRs: [kaappi/keps#53](https://github.com/kaappi/keps/pull/53) records Phase 2 as shipped; [kaappi/kaappi.github.io](https://github.com/kaappi/kaappi.github.io) updates the conformance table's "Cross-thread channels" row, which until now told readers the fixes had shipped but the identifier had not.

### DIFF
--- a/src/types.zig
+++ b/src/types.zig
@@ -1532,7 +1532,10 @@ const is_wasm_target = builtin.os.tag == .wasi;
 /// accessor, KEP-0005 §4) is likewise always compiled in — no platform
 /// carve-out. `kaappi-threads` is omitted on wasm, matching
 /// Lib.wasmAvailable()'s `srfi_18 => false` gate (primitives.zig) — no OS
-/// threads there. (KEP-0004)
+/// threads there. `kaappi-shared-channels` (KEP-0002 cross-thread channel
+/// promotion; gate cleared when kaappi#1487/#1489 closed) rides the same
+/// non-wasm branch: promotion needs OS threads, and on wasi `notify` is a
+/// no-op that nothing ever calls (KEP-0002 §5). (KEP-0004)
 const is_windows_target = builtin.os.tag == .windows;
 const base_platform_features = [_][]const u8{ "r7rs", "kaappi", "ieee-float", "exact-closed", "exact-complex", "kaappi-fibers", "kaappi-reactor", "kaappi-diagnostics" };
 // R7RS appendix B feature identifiers: exactly one OS-class identifier —
@@ -1542,7 +1545,7 @@ const os_feature = [_][]const u8{if (is_windows_target) "windows" else "posix"};
 pub const platform_features = if (is_wasm_target)
     base_platform_features ++ os_feature
 else
-    base_platform_features ++ os_feature ++ [_][]const u8{"kaappi-threads"};
+    base_platform_features ++ os_feature ++ [_][]const u8{ "kaappi-threads", "kaappi-shared-channels" };
 
 test "nil is not a pointer" {
     try std.testing.expect(!isPointer(NIL));
--- a/docs/dev/features.md
+++ b/docs/dev/features.md
@@ -41,7 +41,8 @@ native-only; WASM's entry point just runs a file.
   "sandbox_available": true,
   "features": ["r7rs", "kaappi", "ieee-float", "posix", "exact-closed",
                "exact-complex", "kaappi-fibers", "kaappi-reactor",
-               "kaappi-diagnostics", "kaappi-threads"],
+               "kaappi-diagnostics", "kaappi-threads",
+               "kaappi-shared-channels"],
   "srfis": { "builtin": [1, 9, 13, …], "portable": [0, 2, 4, …] },
   "limits": {
     "initial_frame_capacity": 480,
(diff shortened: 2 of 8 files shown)


src/types.zig
```diff
@@ -1542,7 +1545,7 @@ const os_feature = [_][]const u8{if (is_windows_target) "windows" else "posix"};
 pub const platform_features = if (is_wasm_target)
     base_platform_features ++ os_feature
 else
-    base_platform_features ++ os_feature ++ [_][]const u8{"kaappi-threads"};
+    base_platform_features ++ os_feature ++ [_][]const u8{ "kaappi-threads", "kaappi-shared-channels" };
```
baijum, line 1548: Verified the platform gate against the code, not just the comment: `ThreadNotifier.notify()`'s `.wasi => {}` arm (`src/reactor.zig:143`) and `WasiPollBackend.notifierBackend` returning an empty struct (`src/reactor.zig:787`-792, "nothing for notify() to ring") mean the cross-thread promotion path genuinely cannot exist on wasm — so advertising the identifier there really would be the KEP-0004 failure mode. Non-wasm branch is the right call.

Nit (non-blocking): `docs/dev/freebsd.md:35`, `docs/dev/netbsd.md:157`, and `docs/dev/openbsd.md:126` each enumerate "All capability identifiers (`kaappi-threads`, `kaappi-fibers`, `kaappi-reactor`, `kaappi-diagnostics`)" as present — with a fifth capability identifier that list now under-counts on all three BSD pages. Those files are outside this PR's scope, but worth a one-line follow-up (or folding in here if preferred).

baijum (commented):
Reviewed and verified locally on this branch (macOS aarch64):

- `zig build test` — exit 0 across the full unit suite; the new `cond-expand kaappi-shared-channels feature` test and the extended `srfi-N feature` parser test both pass.
- `zig build wasm` — clean, so the comptime branch holds on the target that omits the identifier.
- `tests/scheme/errors/features.sh` — 45/45, including the derived `cond-expand recognizes 'kaappi-shared-channels'`. Confirmed the suite needs no edit in this PR: it iterates the binary's own advertised features at runtime, so the new check appears automatically.
- `KAAPPI_HOME=$(mktemp -d) zig-out/bin/kaappi tests/scheme/smoke/features-consistency.scm` — 16 passes, exit 0.
- The `docs/dev/features.md` example now matches live `kaappi features --json` output exactly (`[... "kaappi-threads", "kaappi-shared-channels"]`).

The single-table property holds: `types.platform_features` is still the only list, `kaappi features` / `(features)` / `cond-expand` all read it, and `features.sh` asserts the three stay identical. The WASI rationale in the doc comment checks out against the reactor code (inline). One non-blocking doc-drift nit inline. Branch is up to date with main.

src/types.zig
baijum: Folded in — [9e2ca09c](https://github.com/kaappi/kaappi/commit/9e2ca09c). All three BSD pages now list `kaappi-shared-channels` as the fifth capability identifier.

Worth noting the claim those sentences make ("are present — nothing is gated") stayed *true* on all three throughout: BSDs are non-wasm, so they get the identifier. It was only the enumeration that would have under-counted, which is the more insidious shape — a list that reads as exhaustive and quietly is not.

Swept for the same drift elsewhere: `docs/dev/porting.md:74` writes the list open-ended (`kaappi-threads`, …) and needs nothing, `docs/dev/diagnostics.md` names only its own identifier, and `docs/dev/features.md`'s JSON example was already updated in the first commit. Those three BSD pages were the only closed enumerations. `markdownlint-cli2` clean across all 90 files.

PR #2410, opened 2026-08-28
Implement SRFI 273: extensions to data (type-)checking

### LINKED ISSUE #2408: [SRFI 273] Implement Extensions to Data (Type-)Checking (extends SRFI 253)
## Summary

Implement [SRFI 273: Extensions to Data (Type-)Checking](https://srfi.schemers.org/srfi-273/srfi-273.html) (finalized 2026-08-03, Artyom Bologov).

SRFI 273 is a set of quality-of-life extensions layered on top of **SRFI 253**, which Kaappi already ships as a portable library (`lib/srfi/253.sld`). SRFI 253 gives us the base forms (`check-arg`, `check-case`, `lambda-checked`, `case-lambda-checked`, `define-checked`); 273 adds aliasing, pre/post declaration, and **return-value checks**.

## What 273 adds over 253

- **`define-check`** — alias a checking predicate under a new name (helps static-typing implementations that need the check in a separate module for inference).
- **`declare-checked`** — pre- and post-declare the type/check for a value or procedure.
- **Return-value checks** via a `=>` arrow clause in `lambda-checked`, `case-lambda-checked`, and `define-checked`, e.g. `(define-checked (f (x integer?)) => (integer?) ...)`. This is an extension to the **existing 253 forms**, not net-new syntax — so implementing 273 means teaching `lib/srfi/253.sld`'s macros the `=>` clause, or providing them from `(srfi 273)`.
- **`check-case`** enhanced with arrow clauses.
- **`check-impl?`** — escape hatch for implementation-specific type checks (macro-context only; not guaranteed inside `check-arg`).
- **`define-values-checked`** — multi-value binding with per-value checks.

Primary library: `(srfi 273)`. The return-value additions also extend the `(srfi 253)` namespace.

| 253 | Base checked-data forms (extended) | ✅ portable (`lib/srfi/253.sld`) |
| 235 | Combinators (`disjoin`/`conjoin`/`complement`) — optimizable patterns | ✅ portable |
| 26 | `cut` currying — optimizable patterns | ✅ portable |
| 43 | `vector-every` — optimizable patterns | ✅ portable |
| 1 | `every` — optimizable patterns | ✅ built-in |

The 235/26/43/1 references are for *suggested optimizable patterns*, not hard requirements — the checks themselves only need `syntax-rules` and error signalling, both of which Kaappi has.

## Implementation notes

- Pure `syntax-rules` port; no engine changes expected. Fits the standard portable `.sld` pattern.
- The reference implementation is inline in the spec (minimal `define-syntax`/`syntax-rules` for `define-check`, `define-values-checked`, `declare-checked`); the return-value-checking sample lives in the SRFI 253 repository.
- Since the `=>` return-value clause extends forms Kaappi already exports from `(srfi 253)`, decide whether to fold it into `253.sld` or export the extended forms from a new `(srfi 273)` that re-exports/renames — record the decision in `docs/dev/srfi-implementation-notes.md`.
- Add a test suite under `tests/scheme/srfi/` and, once shipped, confirm it registers as the `srfi-273` `cond-expand` feature.

Reference: https://srfi.schemers.org/srfi-273/srfi-273.html

baijum:
Closes #2408

## What

Ships [SRFI 273](https://srfi.schemers.org/srfi-273/srfi-273.html) as a portable `lib/srfi/273.sld` layered on `(srfi 253)`, plus the fixes the `=>` return-value checking needed in `253.sld` — the SRFI 273-specified extension to the 253 forms was already folded into `253.sld` by the original port, but two shapes and all multi-value cases were broken.

## `(srfi 273)` contents

- **`define-check`** — alias a checking predicate under a new name.
- **`declare-checked`** — advisory declaration (all three shapes); expands to `(when #f #f)` exactly as the reference implementation does. Declarations for imported identifiers are accepted, as the spec highly recommends.
- **`define-values-checked`** — like `define-values`, but each returned value is checked against its predicate. Routes the form through `call-with-values` so the values are received exactly once (a `values-checked` splice would evaluate the form once per value). Unlike the minimal reference implementation, the checks are real.
- **`check-impl?`** — auxiliary syntax that strips to its datum. Kaappi defines no implementation-specific check datatypes, so a bound datum behaves as itself and an unknown one (`uint`, `pointer`) is an unbound variable — the spec's `(values-checked ((check-impl? uint)) -1)` is an error here too. Since Kaappi's `check-arg` is a macro, it also works in operand position.
- Re-exports the whole `(srfi 253)` vocabulary, so importing `(srfi 273)` alone suffices.
- `check-case` arrow clauses are deliberately absent — the SRFI waives them.
- The optimizable patterns (SRFI 235/26/43/1) are not statically recognized; they work as ordinary predicates, which is all compliance requires.

## `253.sld` fixes (pre-existing bugs the port surfaced)

- `(lambda-checked () => (pred) body…)` and the rest-formals equivalent silently dropped the clause and compiled to a body with a bare `=>` variable reference. Both shapes now check their returns.
- **Multi-value `=>` checks never worked**: the terminal expansions spliced N predicates into `values-checked` against a single value expression — an ellipsis-count mismatch that compiled to garbage (KP2001) or silently mis-checked. `values-checked` checks N value *expressions* pairwise; it cannot check the N values of one expression.
- New `%check-results` helper: each predicate wraps one `(lambda (v . more))` layer (hygiene gives each recursive level fresh `v`/`more`), checks its layer's first value, and rotates it to the tail on the way out, with the predicate list reversed first so the innermost layer carries the first predicate. The rotation lines value *k* up with predicate *k* and restores the original order at the outermost layer; a count mismatch errors, matching `values-checked`'s documented "number of values and predicates should match".
  - A first cut of this layering without the reversal checked *every* predicate against the *first* value and left the rest unchecked — the new suite pins the pairing (`(=> (integer? string?) (values 7 "x"))` passes, the swapped predicate list errors) so that cannot return.
- Return-check failure errors now name the caller as a quoted symbol (`'f`), consistent with the argument checks.

## Tests

- New `tests/scheme/srfi/srfi273.scm` — 102 assertions: availability (`srfi-273` feature id and `(library (srfi 273))`), re-exports, `define-check`, `declare-checked` (incl. foreign declarations), `define-values-checked` (incl. evaluated-exactly-once and error paths), `check-impl?` (bound datum, unknown datum, argument position, `check-case`), `=>` return checks across `lambda-checked`/`define-checked`/`case-lambda-checked` (empty/rest formals, multi-value pairing, count mismatches), and the optimizable patterns through the checking forms.
- `tests/scheme/srfi/srfi253.scm` — adds the previously untested `define-checked` `=>` path.
- Full `tests/scheme/run-all.sh`: 724 scheme files pass, R7RS suite 1395/1395. `zig build test` green. `tools/check-srfi-status.sh`: OK (SRFI 273 is final).

## Docs

- `docs/dev/srfi-implementation-notes.md` — new SRFI 273 section recording the fold-vs-re-export decision (re-export; `253.sld` stays the owner of the extended forms) and the `values-checked` trap.
- Counts updated: 179 SRFIs / 163 portable in `CLAUDE.md` (AGENTS.md symlink), `README.md`, `CONFORMANCE.md` (+ table row), and the notes' shipped list.
- `srfi-273` registers as a `cond-expand` feature identifier automatically (derived from the `.sld`).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,7 +1,7 @@
 # Kaappi — R7RS Scheme in Zig
 
 Complete R7RS-small Scheme implementation. Zig 0.16, ~100k lines, 695 built-in
-procedures, 178 SRFIs.
+procedures, 179 SRFIs.
 
 This file is the orientation map. Detail lives in `docs/dev/` — every section
 below names the document that owns it. `docs/dev/README.md` is the full index.
@@ -186,14 +186,14 @@ Exceptions: auto-generated data files (`unicode_tables.zig`) are exempt.
 
 ## Libraries and SRFIs
 
-178 SRFIs supported: 12 built-in as Zig primitives (1, 9, 13, 18, 39, 69, 133,
-170, 192, 254, 258, 260), 162 portable `.sld` files under `lib/srfi/` loaded on
+179 SRFIs supported: 12 built-in as Zig primitives (1, 9, 13, 18, 39, 69, 133,
+170, 192, 254, 258, 260), 163 portable `.sld` files under `lib/srfi/` loaded on
 demand, plus SRFI 261 (an import-resolver convention with no library file) and
 the sub-library-only 160, 211 and 226. Every supported SRFI is also a
 `cond-expand` feature identifier `srfi-<n>`, derived rather than listed.
 
 Re-derive the counts after each release with `kaappi features --json` — it
-reports only the 12 + 162, since it scans just the top-level `lib/srfi/N.sld`
+reports only the 12 + 163, since it scans just the top-level `lib/srfi/N.sld`
 files; add the other four by hand.
 
 The library loader in `vm_library.zig` supports `cond-expand`, `include` (paths
--- a/CONFORMANCE.md
+++ b/CONFORMANCE.md
@@ -16,7 +16,7 @@ must stay clean. See README.md "Known limitations → Script output".
 
 ## SRFI conformance
 
-178 SRFIs supported. 12 built-in (native Zig), 162 portable (.sld files), plus SRFI 261 (Portable SRFI Library Reference) as an import-resolver convention with no library file, and SRFI 226, SRFI 160, and SRFI 211 as sub-libraries only with no bare `(srfi 226)`/`(srfi 160)`/`(srfi 211)` file (so none appears as a bare number in `kaappi features`' scan, unlike every other portable SRFI). `(srfi srfi-<n>)` and `(srfi <mnemonic>-<n>)` — e.g. `(srfi srfi-1)`, `(srfi lists-1)`, `(srfi vectors-133)` — resolve to `(srfi <n>)`, with literal names winning when they exist. Coverage details for the built-in SRFIs follow.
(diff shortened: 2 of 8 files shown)


baijum:
Superseded by #2411 — same change, reopened from a DCO-signed branch after this PR's branch had to be replaced.

PR #2411, opened 2026-08-28
Implement SRFI 273: extensions to data (type-)checking

### LINKED ISSUE #2408: [SRFI 273] Implement Extensions to Data (Type-)Checking (extends SRFI 253)

baijum:
Closes #2408

Supersedes #2410 (closed only because its branch was replaced to add the DCO sign-off; identical change plus `Signed-off-by`).

## What

Ships [SRFI 273](https://srfi.schemers.org/srfi-273/srfi-273.html) as a portable `lib/srfi/273.sld` layered on `(srfi 253)`, plus the fixes the `=>` return-value checking needed in `253.sld` — the SRFI 273-specified extension to the 253 forms was already folded into `253.sld` by the original port, but two shapes and all multi-value cases were broken.

## `(srfi 273)` contents

- **`define-check`** — alias a checking predicate under a new name.
- **`declare-checked`** — advisory declaration (all three shapes); expands to `(when #f #f)` exactly as the reference implementation does. Declarations for imported identifiers are accepted, as the spec highly recommends.
- **`define-values-checked`** — like `define-values`, but each returned value is checked against its predicate. Routes the form through `call-with-values` so the values are received exactly once (a `values-checked` splice would evaluate the form once per value). Unlike the minimal reference implementation, the checks are real.
- **`check-impl?`** — auxiliary syntax that strips to its datum. Kaappi defines no implementation-specific check datatypes, so a bound datum behaves as itself and an unknown one (`uint`, `pointer`) is an unbound variable — the spec's `(values-checked ((check-impl? uint)) -1)` is an error here too. Since Kaappi's `check-arg` is a macro, it also works in operand position.
- Re-exports the whole `(srfi 253)` vocabulary, so importing `(srfi 273)` alone suffices.
- `check-case` arrow clauses are deliberately absent — the SRFI waives them.
- The optimizable patterns (SRFI 235/26/43/1) are not statically recognized; they work as ordinary predicates, which is all compliance requires.

## `253.sld` fixes (pre-existing bugs the port surfaced)

- `(lambda-checked () => (pred) body…)` and the rest-formals equivalent silently dropped the clause and compiled to a body with a bare `=>` variable reference. Both shapes now check their returns.
- **Multi-value `=>` checks never worked**: the terminal expansions spliced N predicates into `values-checked` against a single value expression — an ellipsis-count mismatch that compiled to garbage (KP2001) or silently mis-checked. `values-checked` checks N value *expressions* pairwise; it cannot check the N values of one expression.
- New `%check-results` helper: each predicate wraps one `(lambda (v . more))` layer (hygiene gives each recursive level fresh `v`/`more`), checks its layer's first value, and rotates it to the tail on the way out, with the predicate list reversed first so the innermost layer carries the first predicate. The rotation lines value *k* up with predicate *k* and restores the original order at the outermost layer; a count mismatch errors, matching `values-checked`'s documented "number of values and predicates should match".
  - A first cut of this layering without the reversal checked *every* predicate against the *first* value and left the rest unchecked — the new suite pins the pairing (`(=> (integer? string?) (values 7 "x"))` passes, the swapped predicate list errors) so that cannot return.
- Return-check failure errors now name the caller as a quoted symbol (`'f`), consistent with the argument checks.

## Tests

- New `tests/scheme/srfi/srfi273.scm` — 102 assertions: availability (`srfi-273` feature id and `(library (srfi 273))`), re-exports, `define-check`, `declare-checked` (incl. foreign declarations), `define-values-checked` (incl. evaluated-exactly-once and error paths), `check-impl?` (bound datum, unknown datum, argument position, `check-case`), `=>` return checks across `lambda-checked`/`define-checked`/`case-lambda-checked` (empty/rest formals, multi-value pairing, count mismatches), and the optimizable patterns through the checking forms.
- `tests/scheme/srfi/srfi253.scm` — adds the previously untested `define-checked` `=>` path.
- Full `tests/scheme/run-all.sh`: 724 scheme files pass, R7RS suite 1395/1395. `zig build test` green. `tools/check-srfi-status.sh`: OK (SRFI 273 is final).

## Docs

- `docs/dev/srfi-implementation-notes.md` — new SRFI 273 section recording the fold-vs-re-export decision (re-export; `253.sld` stays the owner of the extended forms) and the `values-checked` trap.
- Counts updated: 179 SRFIs / 163 portable in `CLAUDE.md` (AGENTS.md symlink), `README.md`, `CONFORMANCE.md` (+ table row), and the notes' shipped list.
- `srfi-273` registers as a `cond-expand` feature identifier automatically (derived from the `.sld`).

### DIFF
--- a/lib/srfi/253.sld
+++ b/lib/srfi/253.sld
@@ -9,6 +9,10 @@
 ;;; Reference: https://srfi.schemers.org/srfi-253/srfi-253.html
 ;;; License: MIT
 ;;; Author: Artyom Bologov (reference impl); ported to Kaappi
+;;;
+;;; The `=>` return-value checking that SRFI 273 adds to these forms lives
+;;; here too (the SRFI 253 sample implementation carries it); (srfi 273)
+;;; re-exports this library and layers its own forms on top.
 
 (define-library (srfi 253)
   (import (scheme base)
@@ -46,6 +50,53 @@
         ((_ (predicate ...) value ...)
          (values (values-checked (predicate) value) ...))))
 
+    ;; %check-results — check each value returned by a body against the
+    ;; corresponding predicate, preserving count and order (SRFI 273's =>
+    ;; clauses). values-checked cannot be reused here: it pairs N predicate
+    ;; *expressions* with N value expressions, while a body is one
+    ;; expression returning N values. syntax-rules cannot mint N distinct
+    ;; identifiers for a fixed-arity receiver, so instead each predicate
+    ;; wraps one value-passing layer ((v . more) — hygiene gives every
+    ;; recursive expansion's v/more a distinct binding) that checks its
+    ;; layer's first value and rotates it to the tail on the way out. The
+    ;; rotation lines value k up with predicate k and restores the original
+    ;; order at the outermost layer — which is why %cr-rev first reverses
+    ;; the predicates: the innermost layer must carry the first predicate.
+    ;; Rotation is a full cycle (identity) only when the counts match, so
+    ;; %cr-finish rejects any other count, as values-checked's "number of
+    ;; values and predicates should match" already does.
+    (define-syntax %check-results
+      (syntax-rules ()
+        ((_ caller (predicate ...) body ...)
+         (call-with-values
+             (lambda () (%cr-rev caller () (predicate ...) body ...))
+           (%cr-finish caller (predicate ...))))))
+    (define-syntax %cr-rev
+      (syntax-rules ()
+        ((_ caller (rev ...) () body ...)
+         (%cr-check caller (rev ...) body ...))
+        ((_ caller (rev ...) (predicate . remaining) body ...)
+         (%cr-rev caller (predicate rev ...) remaining body ...))))
+    (define-syntax %cr-check
+      (syntax-rules ()
+        ((_ caller () body ...)
+         (begin body ...))
+        ((_ caller (predicate . remaining) body ...)
+         (call-with-values
+             (lambda () (%cr-check caller remaining body ...))
+           (lambda (v . more)
+             (check-arg predicate v caller)
+             (apply values (append more (list v))))))))
+    (define-syntax %cr-finish
+      (syntax-rules ()
+        ((_ caller (predicate ...))
+         (let ((expected (length '(predicate ...))))
+           (lambda received
+             (if (= (length received) expected)
+                 (apply values received)
+                 (error "number of values and predicates should match"
+                        caller)))))))
+
     ;; check-case — predicate dispatch.
     (define-syntax %check-case
       (syntax-rules (else)
@@ -80,9 +131,7 @@
         ((_ name (=> (returns ...) body ...) args (checks ...))
          (lambda args
            checks ...
-           (values-checked
-            (returns ...)
-            (begin body ...))))
+           (%check-results 'name (returns ...) body ...)))
         ;; Terminal: without return check
         ((_ name (body ...) args (checks ...))
          (lambda args
@@ -106,11 +155,17 @@
 
     ;; lambda-checked — lambda with type-checked arguments.
     (define-syntax lambda-checked
-      (syntax-rules ()
+      (syntax-rules (=>)
+        ;; Empty formals with a return-value check (SRFI 273)
+        ((_ () => (returns ...) body ...)
+         (lambda () (%check-results 'lambda-checked (returns ...) body ...)))
         ((_ () body ...)
          (lambda () body ...))
         ((_ (arg . args) body ...)
          (%lambda-checked lambda-checked (body ...) () () arg . args))
+        ;; Rest-arg lambda with a return-value check (SRFI 273)
+        ((_ arg => (returns ...) body ...)
+         (lambda arg (%check-results 'lambda-checked (returns ...) body ...)))
         ;; Rest-arg lambda (no-op, can't check)
         ((_ arg body ...)
          (lambda arg body ...))))
@@ -146,7 +201,7 @@
           (done ...
            (formals
             checks ...
-            (values-checked (returns ...) (begin body ...))))
+            (%check-results 'case-lambda-checked (returns ...) body ...)))
           remaining ...))
         ;; All args consumed — plain body
         ((_ (done ...) (remaining ...)
@@ -199,7 +254,7 @@
             remaining ...)
          (%clc-clauses
           (done ...
-           (() (values-checked (returns ...) (begin body ...))))
+           (() (%check-results 'case-lambda-checked (returns ...) body ...)))
           remaining ...))
         ;; Clause with empty formals, plain body
         ((_ (done ...)
--- a/lib/srfi/253.sld
+++ b/lib/srfi/253.sld
@@ -54,23 +54,51 @@
     ;; corresponding predicate, preserving count and order (SRFI 273's =>
     ;; clauses). values-checked cannot be reused here: it pairs N predicate
     ;; *expressions* with N value expressions, while a body is one
-    ;; expression returning N values. syntax-rules cannot mint N distinct
-    ;; identifiers for a fixed-arity receiver, so instead each predicate
-    ;; wraps one value-passing layer ((v . more) — hygiene gives every
-    ;; recursive expansion's v/more a distinct binding) that checks its
-    ;; layer's first value and rotates it to the tail on the way out. The
-    ;; rotation lines value k up with predicate k and restores the original
-    ;; order at the outermost layer — which is why %cr-rev first reverses
-    ;; the predicates: the innermost layer must carry the first predicate.
-    ;; Rotation is a full cycle (identity) only when the counts match, so
-    ;; %cr-finish rejects any other count, as values-checked's "number of
-    ;; values and predicates should match" already does.
+    ;; expression returning N values.
+    ;;
+    ;; A single predicate — by far the common shape — takes the same cheap
+    ;; let shape values-checked uses: no call-with-values, no closures per
+    ;; call. Like values-checked, it does not itself count values; a body
+    ;; returning zero or several still fails the predicate, because Kaappi
+    ;; propagates multiple values through a single-variable let binding and
+    ;; no predicate matches the resulting values object.
+    ;;
+    ;; Several predicates go through %cr-gate and a layer tower.
+    ;; syntax-rules cannot mint N distinct identifiers for a fixed-arity
+    ;; receiver, so each predicate wraps one value-passing layer
+    ;; ((v . more) — hygiene gives every recursive expansion's v/more a
+    ;; distinct binding) that checks its layer's first value and rotates it
+    ;; to the tail on the way out. The rotation lines value k up with
+    ;; predicate k and restores the original order at the outermost layer —
+    ;; which is why %cr-rev first reverses the predicates: the innermost
+    ;; layer must carry the first predicate. Rotation is a full cycle
+    ;; (identity) only when the counts match, and a short value list would
+    ;; otherwise pair predicate k with value k-M and fail with a misleading
+    ;; predicate message, so %cr-gate checks the count up front and rejects
+    ;; a mismatch, as values-checked's "number of values and predicates
+    ;; should match" already does.
     (define-syntax %check-results
       (syntax-rules ()
+        ((_ caller (predicate) body ...)
+         (let ((v (begin body ...)))
+           (check-arg predicate v caller)
+           v))
         ((_ caller (predicate ...) body ...)
          (call-with-values
-             (lambda () (%cr-rev caller () (predicate ...) body ...))
-           (%cr-finish caller (predicate ...))))))
+             (lambda () body ...)
+           (%cr-gate caller (predicate ...))))))
+    (define-syntax %cr-gate
+      (syntax-rules ()
+        ((_ caller (predicate ...))
+         (let ((expected (length '(predicate ...))))
+           (lambda received
+             (if (= (length received) expected)
+                 ;; The tower rides in the gate's body — as a receiver
+                 ;; operand it would be evaluated eagerly, running the
+                 ;; checks before any values exist.
+                 (%cr-rev caller () (predicate ...) (apply values received))
+                 (error "number of values and predicates should match"
+                        caller)))))))
     (define-syntax %cr-rev
       (syntax-rules ()
         ((_ caller (rev ...) () body ...)
@@ -87,15 +115,6 @@
            (lambda (v . more)
              (check-arg predicate v caller)
              (apply values (append more (list v))))))))
-    (define-syntax %cr-finish
-      (syntax-rules ()
-        ((_ caller (predicate ...))
-         (let ((expected (length '(predicate ...))))
-           (lambda received
-             (if (= (length received) expected)
-                 (apply values received)
-                 (error "number of values and predicates should match"
-                        caller)))))))
 
     ;; check-case — predicate dispatch.
     (define-syntax %check-case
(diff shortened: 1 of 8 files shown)


lib/srfi/253.sld
```diff
@@ -46,6 +50,53 @@
...
+    ;; wraps one value-passing layer ((v . more) — hygiene gives every
+    ;; recursive expansion's v/more a distinct binding) that checks its
+    ;; layer's first value and rotates it to the tail on the way out. The
+    ;; rotation lines value k up with predicate k and restores the original
+    ;; order at the outermost layer — which is why %cr-rev first reverses
+    ;; the predicates: the innermost layer must carry the first predicate.
+    ;; Rotation is a full cycle (identity) only when the counts match, so
+    ;; %cr-finish rejects any other count, as values-checked's "number of
+    ;; values and predicates should match" already does.
+    (define-syntax %check-results
```
baijum, line 80: **The single-predicate `=>` gets ~21x slower than it is on `main`.**

`%check-results` routes *every* `=>` through the `call-with-values` tower, including the one-predicate case that `main` handled with `values-checked`'s cheap `(let ((v value)) (check-arg ...) v)` clause. Per call that now costs two `call-with-values`, N+1 closure allocations for the thunk/receiver layers, a fresh `%cr-finish` closure, a runtime `(length '(predicate ...))`, a rest-args list, plus `append`/`list`/`apply values`.

Measured on this branch, 300k calls of `(define-checked (f (n integer?)) => (integer?) (+ n 1))`, best of two runs:

```
plain (no => at all)                          38,339 jiffies
main's expansion: (values-checked (integer?) (+ n 1))     46,586
this PR's %check-results                     989,306
```

A single-predicate fast path recovers most of it and still enforces exactly-one-value (the fixed-arity receiver does it):

```scheme
((_ caller (predicate) body ...)
 (call-with-values (lambda () body ...)
   (lambda (v) (check-arg predicate v caller) v)))
```

That measured 313k–358k here — 3x better than the current expansion. If you want the full 46k you'd use the `let` form instead, at the cost of the exactly-one-value check (which is what `values-checked` already trades away). Either way, one clause added ahead of the general one keeps the general machinery for the multi-value case where it's actually needed.

lib/srfi/253.sld
```diff
@@ -46,6 +50,53 @@
...
+    (define-syntax %cr-check
+      (syntax-rules ()
+        ((_ caller () body ...)
+         (begin body ...))
+        ((_ caller (predicate . remaining) body ...)
+         (call-with-values
+             (lambda () (%cr-check caller remaining body ...))
+           (lambda (v . more)
+             (check-arg predicate v caller)
+             (apply values (append more (list v))))))))
```
baijum, line 117: **When there are fewer values than predicates, a mispaired check fires before `%cr-finish` can report the count mismatch.**

The rotation is only order-preserving once the cycle completes, so with M values and N > M predicates the layers past M apply predicate k to value k-M rather than to nothing, and one of those usually fails first. `%cr-finish` never runs.

```scheme
(define-checked (mis) => (integer? string?) 7)
(mis)
;; => assumption violated  (string? v)  ... 7  mis
;; expected: number of values and predicates should match
```

No wrong result escapes — it's an error either way — but the message points at `string?` and value `7`, which is a real diagnostic wrong turn for what is actually an arity mistake. It also leaks the internal binding name `v` into the irritants (`(string? v)`), where the argument-check path shows the user's own parameter name.

The zero-value case is a third message again:

```scheme
(define-checked (f) => (integer?) (values))
(f)  ;; => expected at least 1 arguments, got 0
```

Checking the count first would collapse all three into the documented one — receive the body's values once, compare against the expected count, and only then rotate.

Worth noting this is exactly the failure mode the `%check-results` comment says the design guards against; it guards the *pairing*, not the *diagnosis*.

tests/scheme/srfi/srfi273.scm
```diff
@@ -0,0 +1,332 @@
...
+ ((lambda-checked ((n integer?)) => (integer? string?) (values n "x")) 7))
+(test-error
+ ((lambda-checked ((n integer?)) => (string? integer?) (values n "x")) 7))
+(test-error
+ ((lambda-checked ((n string?)) => (integer? string?) (values n "x")) "n"))
+;; The number of values must match the number of predicates — in either
+;; direction, as with values-checked.
+(test-error
+ ((lambda-checked () => (integer?) (values 1 'two "three"))))
+(test-error ((lambda-checked () => (integer? integer?) (values 1))))
```
baijum: This pair reads as covering the count mismatch "in either direction", but only the first one does.

Line 258 (too many values) genuinely reaches `%cr-finish`. Line 259 reaches it only because both predicates are `integer?`, so the mispaired second check happens to pass. Swap one — `(=> (integer? string?) (values 1))` — and the test still goes green, but from a `string?`-vs-`1` predicate failure, not from the count check (see the `%cr-check` comment). So neither assertion pins the behaviour the comment above them describes.

Adding the distinct-predicate case, and asserting on the error message rather than just `test-error`, would pin it.

tests/scheme/srfi/srfi273.scm
```diff
@@ -0,0 +1,332 @@
...
+;; call time.
+(define-checked (ffi-ref (data (check-impl? pointer)) (idx integer?)) idx)
+(test-error (ffi-ref 'anything 0))
+(define-checked (str-ref (s (check-impl? string?)) (idx integer?))
+  (string-ref s idx))
+(test-equal #\k (str-ref "kaappi" 0))
+
+;; In check-case clauses.
+(test-assert (check-case 3 ((check-impl? integer?) 'int) (else 'other)))
+(test-assert (check-case "x" ((check-impl? integer?) 'int) (else 'other)))
```
baijum: These two `check-case` assertions pass regardless of which clause fires — `'int` and `'other` are both truthy, so `test-assert` can't tell them apart. Line 213 in particular is meant to show that a non-matching `(check-impl? integer?)` falls through to `else`, but it would pass just as well if it dispatched to `'int`.

Same shape at line 41 (`re-exports`) and lines 322–323 (`optimizable-patterns`, where 322/323 are the matching/non-matching pair and neither distinguishes). `test-equal` with the expected symbol makes all five real:

```scheme
(test-equal 'int   (check-case 3   ((check-impl? integer?) 'int) (else 'other)))
(test-equal 'other (check-case "x" ((check-impl? integer?) 'int) (else 'other)))
```

lib/srfi/273.sld
```diff
@@ -0,0 +1,81 @@
...
+
+    ;; declare-checked — advisory declaration of the checks for a value or
+    ;; for a separately defined procedure. Expands to nothing, as in the
+    ;; SRFI's reference implementation: Kaappi has no static checker to
+    ;; inform, and re-wrapping an already-defined procedure — possibly an
+    ;; imported one, e.g. (declare-checked (negative? (x real?)) =>
+    ;; (boolean?)) — cannot be done portably from syntax-rules. Later
+    ;; modifications of a declared value are likewise left unchecked.
+    (define-syntax declare-checked
+      (syntax-rules ()
```
baijum: Nit: `=>` isn't declared as a literal here, so in the third clause it is a pattern variable matching anything — `(declare-checked (f (x integer?)) whatever (boolean?))` silently matches the return-check shape.

Harmless today because all three clauses expand to `(when #f #f)`, and the reference implementation has the same `(syntax-rules ())`. But every other `=>`-aware macro in this port declares it (`lambda-checked`, `%lambda-checked`, `%clc-process`, `%clc-clauses` all use `(syntax-rules (=>))`), and this one would become a real trap the moment `declare-checked` grows an expansion that isn't a no-op. Cheap to make consistent now.

baijum (commented):
Reviewed on the branch: built it, ran `tests/scheme/srfi/srfi273.scm` (102 pass) and `srfi253.scm` (108 pass), `tools/check-srfi-status.sh` (OK, 176 final), and `kaappi features --json` confirms 163 portable + 12 built-in, so the 179/163 count edits across `CLAUDE.md`/`README.md`/`CONFORMANCE.md`/the notes are right.

I also probed the semantics directly. The things the PR description claims are true: 3-value ordering is preserved (`(=> (integer? string? symbol?) (values 1 "a" 'b))` -> `(1 a b)`), the swapped predicate list errors, `define-values-checked` evaluates its form exactly once, the return-check failure names the caller (`'g` shows up in the irritants), `(check-impl? p)` works in `check-arg`'s operand position, and all three `declare-checked` shapes expand cleanly — including before an internal `define`, which is the SRFI's primary use case. I additionally checked two expansion sites the suite doesn't reach — `define-checked` with dotted formals + `=>`, and `case-lambda-checked` with a multi-value `=>` on both a fixed and a rest-formals clause — and they work too.

The rotation trick is sound where it matters: I could not construct a case where a wrong value/predicate pairing *passes*. Two things I would still change before merging — one measured performance regression on the common shape, and one place where the count check is reached too late to produce its own message. Details inline.

lib/srfi/253.sld
baijum: Went one step further than the suggested clause: the single-predicate case now expands to the same `let` shape `values-checked` uses —

```scheme
((_ caller (predicate) body ...)
 (let ((v (begin body ...)))
   (check-arg predicate v caller)
   v))
```

That is full parity with main rather than 3x: re-running your benchmark shape on this machine, 300k calls measure 46,316 / 46,387 jiffies for the macro against 38,930 for the hand-written `values-checked` shape (the tower version measured 989k on the first cut, 437k with the fixed-arity receiver).

I took the `let` variant of the trade you laid out, and the exactly-one-value check turns out not to be lost in practice: Kaappi propagates multiple values through a single-variable `let` binding (`(let ((v (values 1 2))) v)` ⇒ `#<values 1 2>`), so a body returning zero or several values binds `v` to something no predicate matches and the check still fails — the same behavior `values-checked` itself has always had on this path. Pinned by the `single predicate … any count other than one is still an error` test.

lib/srfi/253.sld
baijum: Restructured exactly along those lines: `%cr-gate` now receives the body's values once, compares the count, and only then re-emits them into the tower (`(%cr-rev caller () (predicate ...) (apply values received))`). All three of your cases now collapse to the documented arity error — `(mis)` reports `number of values and predicates should match`, the zero-value case too (for N≥2; the single-predicate fast path reports the predicate failure instead, per the first comment).

Your diagnosis also caught the bug in my own first cut of the gate: I initially put the tower in the gate's *receiver* operand, where evaluating it as an argument ran the whole check cascade eagerly before any values existed — the innermost `(begin)` handed `v` the unspecified value and `integer?` failed on it. The suite went 12-for-12 red on the multi-value cases, which is exactly the pinning working. The tower now rides in the gate's body, with a comment recording why.

tests/scheme/srfi/srfi273.scm
baijum: Done, with a twist on the mechanism: Kaappi's `test-error` matches its second argument as a condition type / predicate (`condition-type?` / `procedure?` / `#t`), not a message, so a string there always falls through to `#t` and pins nothing. The count-mismatch tests instead go through a small `guard` helper, `count-mismatch-error?`, that compares `error-object-message` against the exact string, with distinct predicates in both directions:

```scheme
(test-assert (count-mismatch-error?
  (lambda () ((lambda-checked () => (integer? string?) (values 1))))))
(test-assert (count-mismatch-error?
  (lambda () ((lambda-checked () => (integer? string? symbol?) (values 1 'two "three" 'four))))))
```

Neither can pass via a predicate failure any more — and with the count-first gate they couldn't even if the predicates were all `integer?`, since the gate fires before any check runs.

tests/scheme/srfi/srfi273.scm
baijum: All five converted to `test-equal` on the dispatched symbol — the re-exports `check-case`, both `check-impl?` clauses, and the `'small`/`'other` pair in optimizable-patterns — so a fall-through that picks the wrong clause now fails the assertion.

lib/srfi/273.sld
baijum: Done — `(syntax-rules (=>))`, with a comment noting why it matches the rest of the port.

baijum:
Review feedback addressed in 925f6b65, all five points (details in the inline replies):

1. **Perf** — single-predicate `=>` now expands to `values-checked`'s `let` shape: 300k calls measure 46k jiffies vs 39k for the hand-written main shape (was 989k). Full parity instead of the 3x partial recovery.
2. **Count diagnosis** — `%cr-gate` compares value/predicate count before the rotation tower runs, so every mismatch reports `number of values and predicates should match` instead of a mispaired predicate firing first.
3. **Count tests** — pinned on the error message (via a `guard` helper, since `test-error`'s matcher is condition-typed, not message-typed) with distinct predicates in both directions.
4. **check-case tests** — all five dispatch assertions now `test-equal` the selected clause's symbol.
5. **`declare-checked`** — `=>` declared as a literal.

My first cut of the gate put the tower in the receiver operand, which evaluated it eagerly and failed every check on an unspecified value — the suite caught it immediately (12 failures), which is a good sign the multi-value pins are load-bearing. Final state: `run-all.sh` 724 scheme files + R7RS 1395 pass, srfi273 105, srfi253 108.

PR #2412, opened 2026-08-28
Implement SRFI 274: extended list conversion procedures

### LINKED ISSUE #2409: [SRFI 274] Implement Extended List Conversion Procedures (start/end + improper lists)
## Summary

Implement [SRFI 274: Extended List Conversion Procedures](https://srfi.schemers.org/srfi-274/srfi-274.html) (finalized 2026-08-20, Peter McGoron).

Modest extensions to the list-conversion and list-copying procedures that (a) align them with the other R7RS conversion/copy procedures (`vector-copy`, `string->list`) by adding optional `start`/`end` index arguments, and (b) let them operate on **improper lists** (dotted and circular — collectively "im-lists").

## Procedures

Each takes `(im-list [start [end]])` with `0 ≤ start ≤ end ≤ length`:

| Procedure | Target | Library |
|-----------|--------|---------|
| `list-copy` | list / improper list | `(srfi 274 base)` |
| `list->string` | string | `(srfi 274 base)` |
| `list->vector` | vector | `(srfi 274 base)` |
| `list->stream` | SRFI 41 stream | `(srfi 274 41)` |
| `list->ideque` | SRFI 134 ideque | `(srfi 274 134)` |
| `list->generator` | SRFI 158 generator | `(srfi 274 158)` |
| `list->⟨type⟩vector` | SRFI 160 homogeneous vector | `(srfi 274 160 ⟨type⟩)` / `(srfi 274 160 base)` |

Key semantics:
- Optional `start`/`end` select a range; the cdr is never inspected beyond `end`.
- It is an error to pass an improper list **without** an explicit `end` argument — **except** `list-copy`, which may copy an improper list wholesale.

## Library layout note

274 deliberately extends procedures that already live in `(scheme base)` (`list-copy`, `list->string`, `list->vector`). The SRFI exposes them under `(srfi 274 base)` (and the per-collection `(srfi 274 41 | 134 | 158 | 160 …)` sub-libraries) so the extended variants don't clash with the built-ins. Kaappi's port should follow the same sub-library structure — mirror the `(srfi 160 <type>)` sub-library pattern already in `lib/srfi/160/`.

| 1 | `take`/`drop` for range handling | ✅ built-in |
| 41 | streams (`list->stream`) | ✅ portable |
| 134 | ideques (`list->ideque`) | ✅ portable |
| 158 | generators (`list->generator`) | ✅ portable |
| 160 | homogeneous vectors (`list->⟨type⟩vector`) | ✅ sub-library-only (`lib/srfi/160/`) |

## Implementation notes

- Portable R7RS reference implementation is available in the SRFI repo (tested with Chibi 0.11.0) — good starting point.
- Circular-list handling needs a bounded traversal (the `end` argument bounds it); guard the no-`end` improper-list case as an error per spec, except for `list-copy`.
- Add a test suite under `tests/scheme/srfi/` covering: range args, dotted lists, circular lists (with `end`), the error cases, and each target collection type. Record any spec ambiguities in `docs/dev/srfi-implementation-notes.md`.
- Once shipped, confirm the `srfi-274` `cond-expand` feature registers (note: `features --json` only scans top-level `lib/srfi/N.sld`, so the top-level `274.sld` should re-export or `cond-expand` the sub-libraries as appropriate).

Reference: https://srfi.schemers.org/srfi-274/srfi-274.html

baijum:
Fixes #2409

## Summary

Portable port of [SRFI 274](https://srfi.schemers.org/srfi-274/srfi-274.html) (Peter McGoron, MIT; finalized 2026-08-20): every list-conversion procedure gains optional `start`/`end` range arguments and operates on **improper lists** (dotted and circular) whenever `end` is supplied — the cdr of the endth pair is never inspected. No engine changes; 180 SRFIs now supported.

## Library layout

Mirrors the SRFI's own sub-library structure (the extended names deliberately shadow `(scheme base)` / underlying-library bindings, and Kaappi enforces R7RS 5.2's double-import rule, so the variants must live in opt-in libraries):

| Library | Exports |
|---|---|
| `(srfi 274)` | bare alias re-exporting `(srfi 274 base)` — makes `srfi-274` answer as a cond-expand feature id and counts in `kaappi features` |
| `(srfi 274 base)` | `list-copy`, `list->string`, `list->vector` |
| `(srfi 274 41)` / `(srfi 274 134)` / `(srfi 274 158)` | `list->stream`, `list->ideque`, `list->generator` |
| `(srfi 274 160 base)` + twelve `(srfi 274 160 <type>)` | all twelve `list-><type>vector`, mirroring the `(srfi 160 <type>)` layout |
| `(srfi 274 internal)` | shared `argcheck!` (bounded by `end`, so it terminates on circular lists) |

## Port decisions (recorded in `docs/dev/srfi-implementation-notes.md`)

- **No `ideque-unfold` / `<type>vector-unfold`**: Kaappi's `(srfi 134)` doesn't export `ideque-unfold`, and importing all twelve full `(srfi 160 <type>)` surfaces just for their unfold would be heavy. Both start+end cases instead hand the bounded, always-proper range from `(srfi 274 base)`'s `list-copy` to the underlying one-argument converter — same results, one intermediate list each (the allocation the SRFI asks *native* implementations to avoid, acceptable in a portable port).
- **Improper lists without `end`** are diagnosed through the underlying builtins' proper-list checks (and through stream/generator laziness). One documented leniency: `(srfi 274 134)`'s start-only case is silent, because Kaappi's simplified `(srfi 134)` builds an ideque as `(cons lst '())` and never walks its input — inherited from that port, not from SRFI 274.

## Testing

- `tests/scheme/srfi/srfi274.scm` — 164 tests adapted from the SRFI's official suite plus additions: sub-library availability, the spec's inverse properties for `list->string`, freshness/car-sharing/tail-sharing of `list-copy`, the eof boundary of bounded generators, and the error cases (improper without `end`, `end` past length, `start` > `end`, inexact/non-integer indices, out-of-range elements).
- Full local verification: `zig build test` ✅, `tests/scheme/run-all.sh` — 727 Scheme files + R7RS 1395, **0 fail** ✅, `markdownlint-cli2` clean ✅, `tools/check-srfi-status.sh` reports all implemented SRFIs final (274 included) ✅, `kaappi features --json` reports 164 portable SRFIs with 274 present ✅.

## Docs updated

Counts 179→180 / 163→164 in `README.md`, `CLAUDE.md`/`AGENTS.md`, `CONFORMANCE.md` (sub-library list + table row), the notes header, and a `CHANGELOG.md` `Unreleased/Added` entry.


baijum (commented):
Reviewed on the branch: built it, ran `tests/scheme/srfi/srfi274.scm` (164 passes), and probed the edges against the [SRFI 274 spec](https://srfi.schemers.org/srfi-274/srfi-274.html). The port itself looks right — the `end`-bounded walk, the `case-lambda` shapes, the lazy stream/generator constructions and the twelve `(srfi 160 <type>)` re-exports all behave as specified, and the `(do be do be doo . dah) 2 5` example from the spec gives `#(do be doo)`.

Eight things below, none of them blocking correctness of the happy paths: one user-visible diagnostic defect, one wrong error attribution, one stale doc count, two test gaps, one doc claim that's narrower than stated, and two nits.

One design question that doesn't fit an inline anchor — **the bare `(srfi 274)` alias**. I confirmed against the spec that SRFI 274 defines no such library (only `(srfi 274 base)`, `(srfi 274 41|134|158)`, `(srfi 274 160 base)` and `(srfi 274 160 <type>)`). Kaappi's precedent for exactly this shape points the other way: `CONFORMANCE.md` records 160, 211 and 226 as "sub-libraries only with no bare `(srfi 226)`/`(srfi 160)`/`(srfi 211)` file (so none appears as a bare number in `kaappi features`' scan, unlike every other portable SRFI)." So 274 is counted in `kaappi features` where three structurally identical SRFIs deliberately are not, and the mechanism that counts it is a library name no portable SRFI-274 program will ever import. That may well be the right call — a bare alias is friendlier than a scan exception — but the notes should say *why* 274 diverges from the 160/211/226 precedent, since the next person adding a sub-library-only SRFI will look at both and find them contradictory.

lib/srfi/274/internal.sld
```diff
@@ -0,0 +1,27 @@
...
+    (define (length>=? im-list end)
+      (let loop ((lst im-list) (n end))
+        (or (zero? n)
+            (and (pair? lst)
+                 (loop (cdr lst) (- n 1))))))
+
+    (define (argcheck! who start end im-list)
+      (cond
+        ((not (exact-integer? start))
+         (error who "start must be an exact integer" start))
```
baijum: **`error` is called with a symbol as the message**, so `who` lands in `error-object-message` and the actual message becomes an irritant. What a user sees:

```
$ kaappi -e '(list-copy (list 1 2 3) 0 9)'
error[KP3000]: list-copy "list is not long enough" (1 2 3) 9
```

The message string is rendered *in quotes, as data*. Compare the sibling diagnostic from the same run:

```
error[KP3006]: list-tail: index 5 out of range for length 2
```

Two reasons to change it:

1. R7RS §6.11 specifies `(error message obj ...)` with *message* a string; `error-object-message` returning the symbol `list-copy` is a deviation a portable program can observe.
2. It is the only place in the whole portable library tree that does this — I grepped all 575 `(error ...)` call sites under `lib/` and **zero** others pass a symbol; every one of them uses `(error "<proc>: <message>" irritant ...)`.

Suggested shape, keeping the `who` parameter:

```scheme
(define (fail who msg . irritants)
  (apply error (string-append (symbol->string who) ": " msg) irritants))
```

and then `(fail who "list is not long enough" im-list end)`, which renders as `list-copy: list is not long enough (1 2 3) 9`. All four arms in this `cond` are affected.

lib/srfi/274/base.sld
```diff
@@ -0,0 +1,63 @@
...
+             (set-car! head (car lst)))))))
+
+    (define list->string
+      (case-lambda
+        ((lst) (r7rs:list->string lst))
+        ((lst start) (r7rs:list->string (list-tail lst start)))
+        ((lst start end)
+         ;; Bounded copy first (terminates on any im-list given end), then
+         ;; the builtin does the character validation and string build.
+         (r7rs:list->string (list-copy lst start end)))))
```
baijum: **Range errors from `list->string` are attributed to `list-copy`.** This arm delegates without calling `argcheck!` itself, so the `who` that reaches the user is the delegate's:

```scheme
(list->string '(#\a #\b) 0 5)
;; => error-object-message: list-copy
;;    irritants: ("list is not long enough" (#\a #\b) 5)

(list->vector '(1 2) 0 5)
;; => error-object-message: list->vector   ; correct
```

The sibling `list->vector` two definitions down, and every one of the twelve converters in `(srfi 274 160 base)`, pass their own `who` before delegating. This is the one conversion that doesn't, and the test at `tests/scheme/srfi/srfi274.scm:151` (`"end beyond length"`) passes precisely because `test-error` doesn't inspect which procedure raised.

Fix is one line:

```scheme
((lst start end)
 (argcheck! 'list->string start end lst)
 (r7rs:list->string (list-copy lst start end)))
```

(That does re-walk `end` pairs — see the note on `160/base.sld:55` about the same redundancy, which a `who`-taking range helper in `(srfi 274 internal)` would fix for all three call sites at once.)

CONFORMANCE.md
```diff
@@ -170,7 +170,7 @@ String ports track their own position directly (the read cursor and the SRFI 192
 
 ### Portable SRFIs (166 SRFIs: 163 importable as bare `(srfi N)`, plus SRFI 160, 211, and 226 as sub-libraries only)
```
baijum: **Stale count — this heading was missed by the 179→180 / 163→164 sweep.** Every other count in the PR was bumped (line 19 here, `README.md` ×3, `CLAUDE.md` ×2, the notes header), but this one still reads `166 SRFIs: 163 importable`.

The table immediately below now has 167 rows (`awk 'NR>171 && /^\| [0-9]/{c++}'` → 167), which is 164 importable + 160 + 211 + 226. So the heading should be:

```
### Portable SRFIs (167 SRFIs: 164 importable as bare `(srfi N)`, plus SRFI 160, 211, and 226 as sub-libraries only)
```

It's the only stale count left — I swept the tree for `179 SRFI|163 portable|163 importable|166 SRFIs` and this line is the sole hit.

tests/scheme/srfi/srfi274.scm
```diff
@@ -0,0 +1,240 @@
...
+(test-equal "library (srfi 274 134)" #t (cond-expand ((library (srfi 274 134)) #t) (else #f)))
+(test-equal "library (srfi 274 158)" #t (cond-expand ((library (srfi 274 158)) #t) (else #f)))
+(test-equal "library (srfi 274 160 base)" #t
+            (cond-expand ((library (srfi 274 160 base)) #t) (else #f)))
+(test-equal "library (srfi 274 160 u8)" #t
+            (cond-expand ((library (srfi 274 160 u8)) #t) (else #f)))
+
+;;; The per-type libraries carry the same binding as (srfi 274 160 base),
+;;; so both can be imported side by side (R7RS 5.2) and give equal results.
+(test-equal "per-type library re-exports base binding"
```
baijum: **This assertion is vacuous** — both sides evaluate the *same* `list->s8vector` identifier, so it reduces to `(equal? X X)` and holds no matter what the per-type libraries export. Wrapping one side in `(lambda args (apply ...))` changes the call mechanism, not the binding being called.

The comment above it is right about what should be proven, and the proof is already sitting in the import list at the top of the file: importing `(srfi 274 160 base)` and `(srfi 274 160 s8)` together *only compiles* if they carry the same binding — R7RS 5.2 makes two different bindings for one name an error, and Kaappi enforces it (KP2001). So the load-time success is the test, and this `test-equal` adds nothing on top.

Either drop it and let a comment point at the import list, or make it actually discriminating — e.g. import one of the per-type libraries under a prefix and compare the two procedure objects with `eqv?`.

tests/scheme/srfi/srfi274.scm
```diff
@@ -0,0 +1,240 @@
...
+(test-begin "srfi-274")
+
+(define clist '#1= (1 2 . #1#))
+
+;;; ====================================================================
+;;; Availability: the cond-expand feature id and every sub-library.
+;;; ====================================================================
+
+(test-equal "cond-expand srfi-274" #t (cond-expand (srfi-274 #t) (else #f)))
+(test-equal "library (srfi 274)" #t (cond-expand ((library (srfi 274)) #t) (else #f)))
```
baijum, line 63: **The bare `(srfi 274)` library is never actually loaded by this suite.** It's probed only through `cond-expand`'s `library` clause, and that clause resolves via `globals.libraryExists` (`src/compiler_conditionals.zig:319-333`), which checks the live registry and the `.sld` on disk — it never reads, expands, or compiles the library body. So a typo in `274.sld`'s `export` list, or a wrong sub-library name in its `import`, would sail through CI green.

That matters more here than for the other seven availability checks, because the other libraries *are* imported at the top of the file, and because `(srfi 274)` is this PR's own invention (the spec defines no bare library) and the thing that makes 274 count in `kaappi features`.

The fix is one word — add `(srfi 274)` to the import list. I verified on this branch that it coexists with `(srfi 274 base)`: since `274.sld` re-exports the same bindings, R7RS 5.2 is satisfied and the program runs:

```scheme
(import (except (scheme base) list-copy list->string list->vector)
        (srfi 274) (srfi 274 base))
(list-copy '(1 2 3 4) 1 3)   ;; => (2 3)
```

docs/dev/srfi-implementation-notes.md
```diff
@@ -1602,3 +1607,44 @@ syntax) plus a re-export of the whole `(srfi 253)` vocabulary, so importing
...
+- **The extended conversions deliberately shadow existing names**, so the
+  sub-libraries are separate rather than extensions of `(scheme base)` etc. —
+  a program imports `(srfi 274 base)` alongside `(except (scheme base)
+  list-copy list->string list->vector)`, exactly as the SRFI's own test
+  suite does. Kaappi diagnoses a bare double import (KP2001), which is what
+  forces the `except` pattern.
+- **Improper lists without an explicit `end` are an error** everywhere except
+  wholesale `list-copy`. Kaappi diagnoses this through the underlying
+  builtins' proper-list checks (`list->vector`, `list->string`, and every
+  `list-><type>vector` walks the list) and through the laziness of streams
```
baijum, line 1635: **"through the laziness of streams and generators" holds for dotted lists only, not circular ones** — so `(srfi 274 134)` is not the only silent path, and the spec violation being missed is the one the SRFI states outright ("It is an error if im-list is circular").

Measured on this branch:

| expression | result |
|---|---|
| `(list->stream '(1 2 . 3) 1)` forced | raises ✅ |
| `(list->generator '(1 2 . 3) 1)` drained | raises ✅ |
| `(list->stream clist)` / `(list->stream clist 1)` | **infinite stream, no error** |
| `(list->generator clist)` / `(list->generator clist 1)` | **infinite generator, no error** — `(list (g) (g) (g))` → `(1 2 1)` |

Laziness is exactly *why* the circular cases are silent: nothing ever walks off the end to hit the `car` that would raise. For the no-argument cases this is inherited from `(srfi 41)` / `(srfi 158)` rather than introduced here, but the start-only cases are this library's own `(list-tail lst start)` handoff.

Suggest widening the bullet to name all three silent paths (134 start-only + 41/158 on circular input), and adding tests that pin the current behaviour — the suite has no circular-without-`end` case for any of the three, which is presumably how the gap stayed invisible.

lib/srfi/274/160/base.sld
```diff
@@ -0,0 +1,69 @@
...
+          list->f64vector
+          list->c64vector
+          list->c128vector)
+  (begin
+    (define (converter who basic)
+      (case-lambda
+        ((lst) (basic lst))
+        ((lst start) (basic (list-tail lst start)))
+        ((lst start end)
+         (argcheck! who start end lst)
```
baijum: Nit: `argcheck!` runs twice per call. This line validates with the correct `who`, then `list-copy` on the next line runs `(argcheck! 'list-copy start end lst)` again — two full bounded walks of `end` pairs before a third walk copies them. Same shape in `(srfi 274 134)` at `134.sld:28-29`.

Harmless, but it's paid on every ranged conversion, and on a circular list with a large `end` it's a real 2× on the check. Exporting a `who`-taking range builder from `(srfi 274 internal)` — say `(range-list who lst start end)`, checking once and copying — would collapse this to one walk here, one in `134.sld`, and would also give `list->string` the correct `who` (see the note on `base.sld:51`) without adding a third check.

CONFORMANCE.md
```diff
@@ -170,7 +170,7 @@ String ports track their own position directly (the read cursor and the SRFI 192
...
+Loaded on demand from `.sld` files via `(import (srfi N))`. Sub-libraries: (srfi 146 hash), (srfi 166 pretty), (srfi 166 columnar), (srfi 166 unicode), (srfi 166 color), (srfi 171 meta), (srfi 211 explicit-renaming), (srfi 211 define-macro), (srfi 211 syntax-parameter), (srfi 226 control prompts), (srfi 226 control continuations), (srfi 226 control times), (srfi 248 primitives), (srfi 254 ephemerons), (srfi 254 guardians), (srfi 254 transport-cell-guardians), (srfi 254 ephemerons-and-guardians), (srfi 257 misc), (srfi 257 box), (srfi 257 rx), (srfi 263 syntax), (srfi 271 randomized), (srfi 271 determinized), (srfi 274 base), (srfi 274 internal), (srfi 274 41), (srfi 274 134), (srfi 274 158), (srfi 274 160 base), (srfi 274 160 u8), (srfi 274 160 s8), (srfi 274 160 u16), (srfi 274 160 s16), (srfi 274 160 u32), (srfi 274 160 s32), (srfi 274 160 u64), (srfi 274 160 s64), (srfi 274 160 f32), (srfi 274 160 f64), (srfi 274 160 c64), (srfi 274 160 c128).
```
baijum: Nit: `(srfi 274 internal)` is listed in a user-facing catalogue of importable sub-libraries, but its own header says the opposite:

> Not part of the SRFI's public surface; do not import it from user code.

It's the only entry in this list that isn't meant to be imported. Consider dropping it here (and at `docs/dev/srfi-implementation-notes.md:40`, which has the same issue) — the dev-notes prose at line 1622 already records that it exists and what it holds, which is where a contributor would look for it.

lib/srfi/274/internal.sld
baijum: Fixed exactly as suggested: a `fail` helper builds `(string-append (symbol->string who) ": " msg)` and all four arms go through it. Rendered now as `error[KP3000]: list-copy: list is not long enough (1 2 3) 9`, and `error-object-message` returns the string. Added two `guard`-based tests pinning the message text.

lib/srfi/274/base.sld
baijum: Fixed via the `range-list` route you suggested in the 160/base note: the start+end arm is now `(r7rs:list->string (range-list 'list->string lst start end))`, so the check runs once under `list->string`'s own name. Pinned with a `guard` test asserting `error-object-message` is `"list->string: list is not long enough"`.

CONFORMANCE.md
baijum: Fixed — the heading now reads `(167 SRFIs: 164 importable as bare \`(srfi N)\`, plus SRFI 160, 211, and 226 as sub-libraries only)`, matching the 167 table rows. Good catch on the awk cross-check; I'd missed that this heading carried its own count.

tests/scheme/srfi/srfi274.scm
baijum: Agreed it was vacuous. Replaced with your discriminating version: `(prefix (only (srfi 274 160 u8) list->u8vector) u8lib:)` is now imported and the test asserts `(eqv? u8lib:list->u8vector list->u8vector)` — the prefixed per-type binding against the plain base one, so it fails if the re-export ever drifts. (You're right that the co-import alone already proves binding identity at load time; this makes the intent explicit rather than incidental.)

tests/scheme/srfi/srfi274.scm
baijum: Done — `(srfi 274)` is in the import list now (verified it coexists with `(srfi 274 base)` as you did), so the alias's import/export wiring is actually expanded and compiled, not just probed for existence.

docs/dev/srfi-implementation-notes.md
baijum: You're right — 'laziness' only catches dotted input; on circular input it's precisely why the error never fires. The bullet now names all three silent paths (134 start-only improper; 41/158 circular without `end`, no-arg inherited from the underlying libraries, start-only from this port's `list-tail` handoff) and says so explicitly. Tests pin the behavior: `stream-ref` at index 6 for the stream cases (stream->list would not terminate), a 5-element drain for the generator cases, and a no-raise assertion for 134's start-only case.

lib/srfi/274/160/base.sld
baijum: Done as suggested: `(srfi 274 internal)` now exports `range-list`, which checks once under the caller's `who` and copies. The twelve converters here, `(srfi 274 134)`, and `list->string`'s start+end arm all go through it, so each ranged conversion does one bounded walk for the check plus one for the copy — no more double `(argcheck! ...)` + `list-copy` re-check, and the 160 converters no longer need the `(only (srfi 274 base) list-copy)` import at all.

CONFORMANCE.md
baijum: Dropped from this catalogue and from the notes header's sub-library list. The SRFI 274 notes section still records that `internal` exists and holds `argcheck!`/`range-list` (now with an explicit "not itself in the importable sub-library catalogue — it is plumbing" parenthetical), which is where a contributor would look.

baijum:
Thanks for the thorough review — all eight inline points are addressed in af33e728 (each thread has a reply with the specifics), with the full suite re-run locally (727 Scheme files + R7RS 1395, 0 fail) and the SRFI 274 suite at 171 passes.

On the design question — **the bare `(srfi 274)` alias** — I'd like to keep it, for three reasons:

1. **Issue #2409 asks for it explicitly**: *"the top-level `274.sld` should re-export or `cond-expand` the sub-libraries as appropriate… Once shipped, confirm the `srfi-274` `cond-expand` feature registers"*. The `cond-expand` alternative (a file that exists but exports nothing) would satisfy the feature id, but "re-export" is listed first and makes the library do something useful rather than merely exist.
2. **The asymmetry with 160/211/226 points the other way for this SRFI.** Those three have no natural single entry point — 160's types are peers, 211's and 226's sub-libraries serve different purposes — so a bare file would be arbitrary. 274's own structure funnels everything through `(srfi 274 base)`; a bare alias is exactly the "give me the extended conversions" import, the same role `(srfi 271)` plays for its randomized library.
3. **The counting mechanism depends on it.** `features --json` and `tools/check-srfi-status.sh` (the CI guard that verifies every implemented SRFI is `final`) scan top-level `lib/srfi/N.sld` only. Without the bare file, 274 would be the fourth SRFI invisible to the status guard — and unlike 160/211/226, whose sub-libraries can each be checked by number-free names, 274 would have no story for why it's unguarded.

The one real cost you identified — a library name no portable SRFI-274 program will import — is now mitigated: the test suite actually loads the alias (your comment on that line), so a broken `export`/`import` in `274.sld` fails CI rather than sailing through.

CLAUDE.md
```diff
@@ -186,8 +186,8 @@ Exceptions: auto-generated data files (`unicode_tables.zig`) are exempt.
 
 ## Libraries and SRFIs
 
-179 SRFIs supported: 12 built-in as Zig primitives (1, 9, 13, 18, 39, 69, 133,
-170, 192, 254, 258, 260), 163 portable `.sld` files under `lib/srfi/` loaded on
+180 SRFIs supported: 12 built-in as Zig primitives (1, 9, 13, 18, 39, 69, 133,
+170, 192, 254, 258, 260), 164 portable `.sld` files under `lib/srfi/` loaded on
```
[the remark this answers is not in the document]
baijum: Good catch — fixed in b2e9cafe. My 179→180 / 163→164 sweep grepped for `163 portable`, which missed this `12 + 163` spelling. CLAUDE.md now reads `12 + 164` (AGENTS.md is a symlink to it, so both are covered), and a fresh sweep for every remaining `163`/`179` variant across the docs confirms this was the last one.

Issue #2403, opened 2026-08-28
[SRFI 211] `rename` on a circular datum aborts the process — GC root stack overflow


baijum:
**Category:** Expander (SRFI 211 ER macros) · **Severity:** crash (process abort) · **Found by:** review of #2401 (baseline control)

`rename` on a circular datum recurses forever, pushing a GC root per level, until the root stack panics. The panic is uncatchable — no `guard` sees it, the process is gone.

Five lines, all ordinary R7RS:

```scheme
(import (scheme base) (srfi 211 explicit-renaming))
(define-syntax m
  (er-macro-transformer
   (lambda (form rename compare) (rename (cadr form)))))
(m #0=(zz . #0#))
```

```
kaappi internal error — this is a bug in kaappi, not in your program.
  version: v0.25.0 (aarch64-macos, ReleaseSafe)
  while:   compiling min-er.scm

thread 22598722 panic: GC root stack overflow
  src/memory.zig:626  growRootBuffer
  src/expander.zig:481  erRenameDatum   <- gc.pushRoot(&car_new)
  src/expander.zig:482  erRenameDatum   <- recursion on types.cdr(v)
  ... (repeats to MAX_ROOT_CAPACITY)
```

Reproduces on `main` @ 5e1680f2, default `ReleaseSafe`. No stress build, no flag, no extreme input.

## Where

`erRenameDatum` (`src/expander.zig:477-509`) walks pairs by recursing on car and cdr with no cycle guard and no depth bound, pushing two roots per level. `SRFI 211 permits any datum here` ("replacing the symbols at the leaves by their renamings"), so passing the whole form to `rename` is the documented idiom, not an abuse.

## Discriminating controls

- **The cycle is real**: `(define x '#0=(a . #0#))` ⟹ `(eq? x (cdr x))` is `#t`.
- **A plain circular datum is fine**: `(write '#0=(zz . #0#))` prints `#0=(zz . #0#)` and returns. Cycle handling in the printer works (#1954, #1955).
- **`syntax-rules` is unaffected**: `(define-syntax m (syntax-rules () ((_ x) 'done)))` then `(m #0=(zz . #0#))` ⟹ `done`. The pattern binds the form without traversing it.
- **An ER transformer that never calls `rename` on the circular part is unaffected**: returning `(list (rename 'quote) 'ok)` prints `ok`. So `expandProceduralMacro`'s `stripUsertextMarkers` pre-pass is not the trigger — `erRenameDatum` is.

[...]

Issue #1954, opened 2026-07-31
All four output procedures hang forever on a cycle reached through an error-object irritant


baijum:
**Category:** Printer · **Severity:** hang · **Found by:** systematic audit v2, Phase 1D (#1890)

```scheme
(define c (list 1)) (set-cdr! c c)
(define e (guard (x (#t x)) (error "boom" c)))
(display e)          ;; never returns
```

Exit 142 (SIGALRM) at 12s. Affects `write`, `display`, `write-shared` and `write-simple` alike.

## Discriminating controls

- **The same cycle in ordinary position terminates**: `(write c)` ⟹ `#0=(1 . #0#)`. So cycle handling itself works.
- **A cyclic *vector* irritant terminates** — that arm recurses with `depth+1`.

Together those isolate it to container arms that recurse **without** a depth increment or cycle guard.

## Affected containers

Error-object irritants, and mutex / condition-variable names. `printListWithDepth` (`:889-900`) walks the cdr spine with no depth increment and no cycle guard; nothing but the pre-pass stops it, and the pre-pass does not walk into these containers.

`.multiple_values` is a fourth such arm, currently latent — `call-with-values` consumes the object before it can reach the printer.

## `display` hanging is its own violation

R7RS requires `display` to terminate on cyclic structure. An error object with a cyclic irritant is not exotic: any `(error "..." x)` where `x` is a cyclic user structure produces one, and the natural thing to do with a caught error is display it.

## Related

**#1713 fixed exactly this class for record instances** (joining them to `markCyclesRec` via `vectorLikeChildren`) and left these four arms behind. The fix shape is known and already in-tree.

**Verified:** fresh ReleaseSafe build, isolated `KAAPPI_HOME`; the hang and both controls reproduced twice.

PR #2413, opened 2026-08-28
Diagnose circular datums in code position instead of aborting or hanging (#2405)

### LINKED ISSUE #2405: [compiler] Circular datum in code position: hang or stack-overflow abort — several walks assume acyclic forms
**Category:** Compiler · **Severity:** crash (stack-overflow abort) and hang · **Found by:** work on #2403 (baseline controls)

A circular datum placed in **code position** — no ER macro, no `rename`, just an R7RS datum label (`§7.1.2`) — is walked by several compiler passes that assume forms are acyclic. This is the macro-free remainder of the "walk the input assuming it is acyclic" family that #2403's expander fix (and its `collectSetTargets` guards) deliberately did not chase: the three repros below reproduce on `main` @ 5e1680f2 **unchanged by that PR's branch**, and were verified there only to move past the prescan before failing further along.

## Repro A — car-side cycle: process abort

```scheme
(display #1=(p #1# q))
```

Bus error on `main` (ReleaseSafe, macOS ARM): the `set!` pre-scan's sub-form recursion (`collectSetTargets`, `src/compiler.zig`) recurses into the same car edge forever at non-incremented depth until the native stack overflows. With #2403's depth-charged recursion the prescan survives, and the abort moves to `ir.tryFoldFromAST` (`src/ir.zig`), i.e. the IR constant-fold walk has the same assumption.

## Repro B — spine cycle: hang

```scheme
#0=(display 1 . #0#)
```

Prints nothing, spins forever on `main`. With #2403's tortoise-and-hare spine guard the prescan terminates; compilation then ends in `compile error[KP9001]: internal error` — terminating, but the diagnosis should name the improper/circular form rather than "internal error".

## Repro C — let-syntax body cycle: hang in the real compiler

```scheme
#0=(let-syntax () . #0#)
```

Spins on `main` in `compileSyntaxBody` (`src/compiler_define_syntax.zig:379` recursion on its own body); unchanged by #2403, which only bounds the pre-scan.

## Controls

- Quoted circular data is fine everywhere: `(define x '#0=(zz . #0#))` + `(write x)` prints `#0=(zz . #0#)` (#1954/#1955 printer cycle handling).
- `syntax-rules` macro uses with circular arguments are fine: `(define-syntax m (syntax-rules () ((_ x) 'done)))` then `(m #0=(zz . #0#))` ⟹ `done`.
- ER `rename` on circular input is now a clean catchable `KP2002` (#2403).

## Priority

`critical` per `docs/dev/github-issues.md`: Repro A is a process abort (stack overflow ⟹ SIGBUS/SIGSEGV) reachable from a one-line ordinary program, and no `guard` can intercept it.

## Suggested direction

A shared cycle-aware "walk a form" discipline for the compile-side passes (the IR folder, `compileSyntaxBody`'s body recursion, and wherever Repro B bottoms out after the prescan), mirroring what #2403 did in the expander: exact cycle detection (path/visited set, or tortoise-and-hare where the walk is a single cdr iteration), and a diagnosable compile error — ideally "circular/improper form in code position" rather than `KP9001 internal error` — instead of a hang or an abort.

baijum:
Fixes #2405.

## What broke

A datum label (R7RS §7.1.2 `#n=`/`#n#`) in **code position** — no ER macro, no `rename`, just the reader datum — put a genuine cycle into forms that several compile-side walks assumed were acyclic. The three repros from the issue, verified on `main` @ c92284a6:

| Repro | On `main` |
|---|---|
| `(display #1=(p #1# q))` | **SIGBUS abort** — `lowerCall` recursed into the same car edge through `lowerWithMacros` until the native stack overflowed (uncatchable) |
| `#0=(display 1 . #0#)` | terminated, but as `KP9001 internal error` — the arg-spine walk spun until the 256-argument cap |
| `#0=(let-syntax () . #0#)` | **hang** — `compileSyntaxBody`'s body walk spun forever |

Probing further, the same family also hung `and`/`or`/`when` bodies (eager IR spine walks), `lambda` and `let`-family bodies (`scanBodyDefs`'s splice scan), `do` (the capture scan), quasiquote templates, and the `apply`/`call-with-values`/`eval` tail fast paths; `(cond (1 . #0#))` aborted like Repro A.

## The fix

The cycle-aware walk discipline #2403 established in the expander and the `set!` pre-scan, now shared across the compile side:

- **`compiler.SpineWalk`** — Floyd tortoise-and-hare guard for walks that advance exactly one cdr per iteration. Exact and allocation-free; generalizes the inline guard #2403 left in `collectSetTargets`. Applied at the IR arg/body walks (`lowerCall`, `lowerList`, `lowerCondBody`, `lowerBegin`), the body scanners (`scanBodyDefs`, `spliceLeadingBegins`, `compileExprSequence`), `compileSyntaxBody`, the let/do binding walks, `hasUnquoteSplicing`/`compileQQSplicing`, and the passthrough tail fast paths.
- **Active-path sets keyed on Object address** (stable in this non-moving GC; they hold no Values) for walks that recurse through arbitrary sub-form positions: IR lowering's `lowerWithMacros` funnel (one path per `IR` instance — every caller already scopes an instance to one expression tree), begin splicing, named-let renaming, `do`'s capture scan, and `compileQQ` (keying vectors too, since its fresh list wrapper gets a new address per visit). Path **membership**, not prior visitation: shared-but-acyclic structure — the same `#1=` sub-form as siblings — renews legitimately, and the expander's shared pattern-variable subtrees keep compiling unchanged.
- **`compiler.circularFormError`** — the shared diagnosis: catchable `InvalidSyntax` whose recorded `syntax_error_detail` the reporter renders as

  ```
  file:1:10: syntax-error[KP2002]: circular form in code position: the form contains itself (datum-label cycle)
  ```

  instead of an abort, a hang, or `KP9001`.

## Controls (from the issue, all still green)

- `(define x '#0=(zz . #0#))` + `(write x)` ⟹ `#0=(zz . #0#)` — quoting makes the datum a constant no code walk enters; printer cycle detection (#1954) untouched.
- `(m #0=(zz . #0#))` with `syntax-rules` ⟹ `done`.
- Shared-but-acyclic code (`(begin #1=(+ 1 2) #1#)`) still compiles and runs.

The native tier improves for free: these forms still compile (eval-fallback forms), and the binary's embedded eval now reports the compile error and exits 1 instead of aborting or hanging at run time.

## Tests

- Unit: `src/tests_circular_code.zig` — the three repros, the walk family, quasiquote, the tail fast paths, plus controls for quoted data, shared structure, and the detail-channel reset.
- Scheme: `tests/scheme/errors/circular-code-2405.sh` — message + exit pins for nine shapes and the issue's controls.
- Native tier: `tests/scheme/compile/native-circular-code-2405.sh` — interpreter oracle (exit 1, KP2002) and the compiled binary terminates.

Full suites green on the rebased tree (main @ f0937b21): unit **1900/1907** (+7 skip), gc-stress **1907/1907**, run-all **2125/2125**, R7RS **1395/1395**.

### DIFF
--- a/src/compiler_advanced.zig
+++ b/src/compiler_advanced.zig
@@ -153,13 +153,13 @@ pub fn compileCase(self: *Compiler, args: Value, dst: u16, is_tail: bool) Compil
 
     var end_jumps: std.ArrayList(usize) = .empty;
     defer end_jumps.deinit(self.gc.allocator);
-    var current = clauses;
+    var current = compiler_mod.SpineWalk.init(clauses);
     var had_else = false;
 
-    while (current != types.NIL) {
-        if (!types.isPair(current)) return CompileError.InvalidSyntax;
-        const clause = types.car(current);
-        current = types.cdr(current);
+    while (current.cur != types.NIL) : (current.next()) {
+        if (!types.isPair(current.cur)) return CompileError.InvalidSyntax;
+        if (current.cyclic()) return compiler_mod.circularFormError();
+        const clause = types.car(current.cur);
         if (!types.isPair(clause)) return CompileError.InvalidSyntax;
 
         const datums = types.car(clause);
--- a/src/compiler_advanced.zig
+++ b/src/compiler_advanced.zig
@@ -315,12 +315,13 @@ pub fn compileQuasiquote(self: *Compiler, args: Value, dst: u16) CompileError!vo
     const template = types.car(args);
     // #2405: a datum-label cycle in the template recursed through compileQQ's
     // car/cdr (and vector re-wrapping) forever — the template is rebuilt with
-    // runtime conses, which a cycle has no finite form of. Active-path set
-    // keyed on Object address (stable in this non-moving GC, holds no Values),
-    // the erRenameDatum discipline from #2403.
-    var path: std.AutoHashMapUnmanaged(usize, void) = .empty;
-    defer path.deinit(self.gc.allocator);
-    try compileQQ(self, template, dst, 0, &path);
+    // runtime conses, which a cycle has no finite form of. The path is the
+    // Compiler's SHARED set (see Compiler.code_path): the splicing desugar
+    // re-wraps a template element as a fresh `(quasiquote elem)` pair and
+    // compiles it through compileExprViaIR, which re-enters compileQuasiquote
+    // with a new wrapper each round — only an address set that survives the
+    // re-wrap catches the element coming back around (review of PR #2413).
+    try compileQQ(self, template, dst, 0);
 }
 
 /// True when `head` is the quasiquote keyword `kw` — recognized through the
@@ -332,15 +333,15 @@ fn isQQKeyword(head: Value, comptime kw: []const u8) bool {
     return std.mem.eql(u8, types.stripHygienicPrefix(types.symbolName(head)), kw);
 }
 
-fn compileQQ(self: *Compiler, tmpl: Value, dst: u16, depth: u8, path: *std.AutoHashMapUnmanaged(usize, void)) CompileError!void {
+fn compileQQ(self: *Compiler, tmpl: Value, dst: u16, depth: u8) CompileError!void {
     if (types.isVector(tmpl)) {
         // #2405: key the path on the VECTOR too — its fresh list wrapper gets
         // a new address each visit, so only the vector's own address catches
         // the vector↔pair cycle (`#0=#((x . #0#))`).
         const vec_key = @intFromPtr(types.toObject(tmpl));
-        if (path.contains(vec_key)) return compiler_mod.circularFormError();
-        try path.put(self.gc.allocator, vec_key, {});
-        defer _ = path.remove(vec_key);
+        if (self.codePathContains(vec_key)) return compiler_mod.circularFormError();
+        self.codePathPush(vec_key);
+        defer self.codePathPop(vec_key);
         // Vector quasiquote: compile elements as a list at the current depth,
         // then call list->vector. This preserves the nesting level so that
         // inner unquotes at depth > 0 remain literal (#850).
@@ -365,7 +366,7 @@ fn compileQQ(self: *Compiler, tmpl: Value, dst: u16, depth: u8, path: *std.AutoH
         // Compile the list at the current quasiquote depth
         const fn_reg = try self.allocReg();
         const arg_reg = try self.allocReg();
-        try compileQQ(self, list, arg_reg, depth, path);
+        try compileQQ(self, list, arg_reg, depth);
         // Call list->vector on the result
         const l2v_sym = gc.allocSymbol("list->vector") catch return CompileError.OutOfMemory;
         const l2v_idx = try self.addConstant(l2v_sym);
@@ -398,9 +399,9 @@ fn compileQQ(self: *Compiler, tmpl: Value, dst: u16, depth: u8, path: *std.AutoH
     // #2405: the active-path membership test for the pair branches below —
     // see compileQuasiquote.
     const tmpl_key = @intFromPtr(types.toObject(tmpl));
-    if (path.contains(tmpl_key)) return compiler_mod.circularFormError();
-    try path.put(self.gc.allocator, tmpl_key, {});
-    defer _ = path.remove(tmpl_key);
+    if (self.codePathContains(tmpl_key)) return compiler_mod.circularFormError();
+    self.codePathPush(tmpl_key);
+    defer self.codePathPop(tmpl_key);
 
     const head = types.car(tmpl);
 
@@ -427,7 +428,7 @@ fn compileQQ(self: *Compiler, tmpl: Value, dst: u16, depth: u8, path: *std.AutoH
             try self.emitU16(unquote_sym_idx);
 
             const inner_reg = try self.allocReg();
-            try compileQQ(self, types.car(rest), inner_reg, depth - 1, path);
+            try compileQQ(self, types.car(rest), inner_reg, depth - 1);
 
             // Build (inner . ())
             const nil_reg = try self.allocReg();
@@ -470,7 +471,7 @@ fn compileQQ(self: *Compiler, tmpl: Value, dst: u16, depth: u8, path: *std.AutoH
         try self.emitU16(us_sym_idx);
 
         const inner_reg = try self.allocReg();
-        try compileQQ(self, types.car(rest), inner_reg, depth - 1, path);
+        try compileQQ(self, types.car(rest), inner_reg, depth - 1);
 
         // Build (inner . ())
         const nil_reg = try self.allocReg();
@@ -511,7 +512,7 @@ fn compileQQ(self: *Compiler, tmpl: Value, dst: u16, depth: u8, path: *std.AutoH
         try self.emitU16(qq_sym_idx);
 
         const inner_reg = try self.allocReg();
-        try compileQQ(self, types.car(rest), inner_reg, depth + 1, path);
+        try compileQQ(self, types.car(rest), inner_reg, depth + 1);
 
         const nil_reg = try self.allocReg();
         try self.emitOp(.load_nil);
@@ -542,10 +543,10 @@ fn compileQQ(self: *Compiler, tmpl: Value, dst: u16, depth: u8, path: *std.AutoH
 
     // Regular pair: cons car and cdr
     const car_reg = try self.allocReg();
-    try compileQQ(self, types.car(tmpl), car_reg, depth, path);
+    try compileQQ(self, types.car(tmpl), car_reg, depth);
 
     const cdr_reg = try self.allocReg();
-    try compileQQ(self, types.cdr(tmpl), cdr_reg, depth, path);
+    try compileQQ(self, types.cdr(tmpl), cdr_reg, depth);
 
     try self.emitOp(.cons);
     try self.emitU16(dst);
--- a/src/compiler_advanced.zig
+++ b/src/compiler_advanced.zig
@@ -313,7 +313,14 @@ pub fn compileCase(self: *Compiler, args: Value, dst: u16, is_tail: bool) Compil
 pub fn compileQuasiquote(self: *Compiler, args: Value, dst: u16) CompileError!void {
     if (args == types.NIL) return CompileError.InvalidSyntax;
     const template = types.car(args);
-    try compileQQ(self, template, dst, 0);
+    // #2405: a datum-label cycle in the template recursed through compileQQ's
+    // car/cdr (and vector re-wrapping) forever — the template is rebuilt with
+    // runtime conses, which a cycle has no finite form of. Active-path set
+    // keyed on Object address (stable in this non-moving GC, holds no Values),
+    // the erRenameDatum discipline from #2403.
+    var path: std.AutoHashMapUnmanaged(usize, void) = .empty;
+    defer path.deinit(self.gc.allocator);
+    try compileQQ(self, template, dst, 0, &path);
 }
 
 /// True when `head` is the quasiquote keyword `kw` — recognized through the
@@ -325,8 +332,15 @@ fn isQQKeyword(head: Value, comptime kw: []const u8) bool {
     return std.mem.eql(u8, types.stripHygienicPrefix(types.symbolName(head)), kw);
 }
 
-fn compileQQ(self: *Compiler, tmpl: Value, dst: u16, depth: u8) CompileError!void {
+fn compileQQ(self: *Compiler, tmpl: Value, dst: u16, depth: u8, path: *std.AutoHashMapUnmanaged(usize, void)) CompileError!void {
     if (types.isVector(tmpl)) {
+        // #2405: key the path on the VECTOR too — its fresh list wrapper gets
+        // a new address each visit, so only the vector's own address catches
+        // the vector↔pair cycle (`#0=#((x . #0#))`).
+        const vec_key = @intFromPtr(types.toObject(tmpl));
+        if (path.contains(vec_key)) return compiler_mod.circularFormError();
+        try path.put(self.gc.allocator, vec_key, {});
+        defer _ = path.remove(vec_key);
         // Vector quasiquote: compile elements as a list at the current depth,
         // then call list->vector. This preserves the nesting level so that
         // inner unquotes at depth > 0 remain literal (#850).
@@ -351,7 +365,7 @@ fn compileQQ(self: *Compiler, tmpl: Value, dst: u16, depth: u8) CompileError!voi
         // Compile the list at the current quasiquote depth
         const fn_reg = try self.allocReg();
         const arg_reg = try self.allocReg();
-        try compileQQ(self, list, arg_reg, depth);
+        try compileQQ(self, list, arg_reg, depth, path);
         // Call list->vector on the result
         const l2v_sym = gc.allocSymbol("list->vector") catch return CompileError.OutOfMemory;
         const l2v_idx = try self.addConstant(l2v_sym);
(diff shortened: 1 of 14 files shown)


src/ir.zig
```diff
@@ -626,6 +639,19 @@ pub fn lowerWithMacros(ir: *IR, expr: Value, macros: ?*std.StringHashMap(Value))
...
+        // re-enters this function on the same pair at ever-increasing depth
+        // — every recursive position routes through this one funnel — until
+        // the native stack aborts, uncatchably. Membership on the active
+        // path means exactly that back-edge (see IR.lower_path); a cycle
+        // has no leaves to bottom out at, so it is a named compile error,
+        // not an internal one. Quoted data never passes through here —
+        // lowerQuote makes the whole datum a constant without walking it —
+        // so circular *data* keeps working (#1954 controls).
+        const path_key = @intFromPtr(types.toObject(expr));
+        if (ir.lower_path.contains(path_key)) return compiler_mod.circularFormError();
```
baijum: `lower_path` lives on the `IR` instance, but the compile pipeline re-enters lowering through `Compiler.compileExprViaIR` ([compiler.zig:685](https://github.com/kaappi/kaappi/blob/35aa78f5/src/compiler.zig#L685)) for **every sub-expression of an `sexpr_form`** — let/letrec/do binding inits, lambda and clause bodies, `let-values` inits — and that builds a *fresh* `IR` with an empty path. The active path is dropped at that boundary, so any cycle whose back-edge crosses it is invisible to this guard.

That is exactly the abort set from the review summary. `#0=(let ((x #0#)) x)`:

```
compileLet -> compileExprViaIR(#0#)   [fresh IR, path = {}]
  -> lowerWithMacros(#0=(let ...))    [pushes the let form]
    -> sexpr_form .let -> compileLet -> compileExprViaIR(#0#)   [fresh IR again]
```

Each level pushes a native frame and never re-meets itself, so it SIGBUSes uncatchably — verified on this branch:

```
$ echo '(display #0=(let ((x #0#)) x))' > /tmp/t.scm && zig-out/bin/kaappi /tmp/t.scm
Bus error at address 0x16f487912
.../src/compiler_bindings.zig:116:5: 0x102ea02c4 in compileLet (kaappi)
```

Same for `letrec`, `lambda`, `define`, `cond`, `do`, `let-values`; `let*` and `case` bottom out in `KP9001` instead.

The comment above the field ("every caller already scopes an instance to one expression tree") is accurate about *IR* instances but is the reason the guard misses: one expression tree is exactly the wrong granularity, because a datum-label cycle in code position spans compile-side re-entry, not just one lowering.

Suggestion: hang the active-path set on the `Compiler` rather than the `IR` and have `compileExprViaIR` push/pop against it (or thread the `IR`'s map into the nested instance). Soundness is unchanged — path membership still means "this exact form object is an ancestor of itself", which is a real cycle for any legitimate program, since the only way a form is nested under itself is a back-edge or an unbounded macro regress. It would also subsume the per-walk sets in `compiler_bindings.CaptureScan`, `renameInBody`, and `appendSplicedBodyElement`.

src/compiler_advanced.zig
```diff
@@ -514,17 +535,17 @@ fn compileQQ(self: *Compiler, tmpl: Value, dst: u16, depth: u8) CompileError!voi
     }
 
     // Check if any element uses unquote-splicing
-    if (hasUnquoteSplicing(tmpl, depth)) {
+    if (try hasUnquoteSplicing(tmpl, depth)) {
```
baijum, line 539: `compileQQSplicing` is the one quasiquote path that doesn't take `path`, and it can't just be given one: its compound elements are re-wrapped as a fresh `(quasiquote elem)` pair and compiled via `compileExprViaIR` ([compiler_advanced.zig:733](https://github.com/kaappi/kaappi/blob/35aa78f5/src/compiler_advanced.zig#L733) in `buildQQListExpr`), which re-enters `compileQuasiquote` and allocates a **new empty path**. So a cycle whose back-edge is a template *element* under a splice escapes the guard:

```
$ echo '(display `#0=(x ,@(list 1) #0#))' > /tmp/t.scm && zig-out/bin/kaappi /tmp/t.scm
/tmp/t.scm:1:1: compile error[KP9001]: internal error
```

That's the same `KP9001` symptom repro B was filed for (better than v0.25.0, which SIGBUSes here, but not the diagnosis this PR promises). The spine guard you added a few lines down does its job — it's the element recursion that leaks, because the wrapper pair is a new object each round so address-keying never re-meets it.

A `Compiler`-level path (see the `src/ir.zig` comment) fixes this one for free: the element being re-wrapped is `tmpl` itself, and `compileQQ` would find `tmpl`'s own address still on the path when it comes back around.

Worth a `` `#0=(x ,@(list 1) #0#) `` case in `tests_circular_code.zig` alongside the two spine templates already there.

tests/scheme/errors/circular-code-2405.sh
```diff
@@ -0,0 +1,97 @@
...
+cat > "$CONTROL" <<'EOF'
+(define x '#0=(zz . #0#))
+(write x)
+(newline)
+(define-syntax m (syntax-rules () ((_ x) 'done)))
+(display (m #0=(zz . #0#)))
+(newline)
+EOF
+
+control_output=$("$KAAPPI" "$CONTROL" 2>&1)
```
baijum: `set -euo pipefail` makes this FAIL branch unreachable. If `kaappi` exits non-zero on the controls, the `control_output=$(...)` assignment trips `set -e` and the script dies right here — `control_status=$?` never runs, no `FAIL:` message is printed, and `run-all.sh` sees kaappi's raw exit code instead of the diagnostic. When the branch *is* reached, `control_status` is always 0, so `-eq 0` is vacuous.

`check_circular` above already has the right idiom:

```bash
control_output=$("$KAAPPI" "$CONTROL" 2>&1) && control_status=0 || control_status=$?
```

Reproduces standalone:

```
$ bash -c 'set -euo pipefail; out=$(bash -c "echo hi; exit 3"); st=$?; echo "reached st=$st"'; echo "outer=$?"
outer=3
```

src/compiler_passthrough.zig
```diff
@@ -115,12 +117,11 @@ pub fn compileCall(self: *Compiler, expr: Value, dst: u16, is_tail: bool) Compil
 
     try self.compileExprViaIR(operator, base, false);
 
-    arg_list = types.cdr(expr);
-    while (arg_list != types.NIL) {
-        const arg = types.car(arg_list);
+    var arg_walk = compiler_mod.SpineWalk.init(types.cdr(expr));
```
baijum, line 120: Nit: this loop and the one in `compileSelfTailCall` (line 246) use `SpineWalk` purely as an iterator — no `cyclic()` call, and no `isPair` check either. That's safe today only because the counting loop above returns `circularFormError` before either can run, but a `SpineWalk` with no `cyclic()` reads like a guard that is one. A plain `var arg_list = types.cdr(expr)` with a comment pointing at the counting loop would say what's actually true; or call `cyclic()` here too and let it be cheap redundancy.

baijum (commented):
Reviewed on the PR head (35aa78f5) rebased on `main` @ f0937b21, built ReleaseSafe on aarch64-macos.

**The discipline is right and the mechanics check out.** `SpineWalk` is a correct Floyd guard — the tortoise only ever advances to a position the head already occupied and survived, so its `cdr` never dereferences a non-pair; `next()` is the sole advance at every call site I checked; the `break`-only loops in `scanBodyDefs` keep their old semantics (no `continue` that would now double-advance); `compileQQSplicing`'s `continue`s do run the loop's continue-expression, matching the old advance-at-top; `compileExprSequence`'s `defer current.* = walk.cur` is safe (neither of the two callers reads it back). `syntax_error_detail` is 512 bytes against a 77-byte message. The three repros, the new shell suite (10/10), and the unit binary (1812 passed, 7 skipped, 0 failed) are all green here.

**But the headline claim — "instead of an abort, a hang, or `KP9001`" — does not hold for the family.** Nine shapes that are one keystroke away from the issue's own repros still abort or still report `KP9001`, on this branch:

| Form | PR head | v0.25.0 |
|---|---|---|
| `#0=(let ((x #0#)) x)` | **SIGBUS** (134) | SIGBUS |
| `#0=(letrec ((x #0#)) x)` | **SIGBUS** | SIGBUS |
| `#0=(lambda () #0#)` | **SIGBUS** (recursive panic) | SIGBUS |
| `#0=(define (f) #0#)` | **SIGBUS** | SIGBUS |
| `#0=(cond (1 #0#))` | **SIGBUS** | SIGBUS |
| `#0=(do ((i #0#)) (#t))` | **SIGBUS** | SIGBUS |
| `#0=(let-values (((a) #0#)) a)` | **SIGBUS** | SIGBUS |
| `#0=(let* ((x #0#)) x)` | `KP9001` | SIGBUS |
| `#0=(case 1 ((1) #0#))` | `KP9001` | SIGBUS |
| `` `#0=(x ,@(list 1) #0#) `` | `KP9001` | SIGBUS |

None of these are regressions — they all abort on v0.25.0 too — and `(guard (e (#t 1)) #0=(let ((x #0#)) x))` still aborts uncatchably, so by the `docs/dev/github-issues.md` rubric this is still `critical`: an abort reachable from a five-character variation on the issue's own repro. The dotted form of each of these *is* fixed (`#0=(cond (1 . #0#))` → KP2002, `#0=(let ((x 1)) . #0#)` → KP2002), which is exactly the shape of the gap: **the spine guards landed everywhere, the recursion guard only covers what one `IR` instance lowers.** Details in the inline comments on `src/ir.zig` and `src/compiler_advanced.zig`.

I'd rather see this land complete than land as-is with a follow-up issue — the two gaps have one shared cause and the tests are already written in the shape that would cover them (just swap `. #0#` for ` #0#`). If you'd rather split it, the PR body and the test-file header should stop claiming the family is covered, since a reader following `docs/dev/testing.md` would take `tests_circular_code.zig`'s "the body/binding walk family terminates" as the family being closed.

**Test coverage note:** every case in `tests_circular_code.zig` and `circular-code-2405.sh` except repro A uses a spine cycle (`. #0#`). The car-position variant is the one that aborts, and it's untested for every form but `lowerCall`. Adding `#0=(let ((x #0#)) x)`, `#0=(lambda () #0#)`, `#0=(cond (1 #0#))` and `` `#0=(x ,@(list 1) #0#) `` to `expectCircular` would have caught both findings.

baijum:
Thanks @coderabbitai — both findings verified against the binary before fixing (3e07b075):

**Major (code-position cycles not recursing through `lowerWithMacros`): confirmed, and wider than the two examples.** `(set! #0=(f 1 . #0#) 3)` did hang — the SRFI-17 setter desugar walks the target's operand spine raw, in both `ir.zig` and its legacy twin. Probing the same class found four more unguarded walks: the `cond`/`cond-expand`/`case` clause walks (a clause list whose own spine cycles hangs whenever no clause matches first — the common cases only terminated because a match or a non-pair clause exited early), the top-level cond-expand selector `vm_eval.selectCondExpandBody` (a separate walk outside the compiler, found by sampling the hung process), and `validPatternGrammar` (a cyclic *pattern* — `(_ #0=(a . #0#))` — spun the validator). All now carry the shared guard; the cyclic pattern is bounded (tortoise + depth cap) and reports as invalid grammar, which is the truthful verdict for a pattern with no finite form.

For the `if` example: confirmed it compiled the `if` symbol as the alternate (`#0=(if #t 1 . #0#)` printed 1). `if` has always silently accepted extra forms after the alternate, so this is fixed **detection-only** (`compiler.spineCyclic` on the tail): a cyclic tail now names itself via the circular-form diagnosis, while the pre-existing acceptance of non-cyclic extras is unchanged — that laxity felt out of scope to tighten in a cycle-safety PR.

**Minor (bound each interpreter invocation): fixed** — `check_circular` now runs each program through the same background + bounded-wait pattern as the native test; a hung run reports as a failed case instead of stalling the suite.

New regressions pin all of it: `(set! #0=(f 1 . #0#) 3)` and `#0=(if #t 1 . #0#)` as message+exit pins in the shell test, plus four new unit tests (set! spine, cyclic pattern, clause cycles, if tail, and the top-level cond-expand selector via the VM detail channel). Suites: unit 1903/1910 (+7 skip), gc-stress 1910/1910, run-all 2125/2125, R7RS 1395/1395.

baijum:
@baijum Thank you — all four comments verified against the binary before fixing, and all addressed in dc4bd141 (both of your repros now report `syntax-error[KP2002]: circular form in code position`, and your full family — letrec/lambda/define/cond/do/let-values — is pinned as regression tests):

**1. The cross-boundary abort (`#0=(let ((x #0#)) x)`)** — confirmed the SIGBUS exactly as you traced it, including that my "one expression tree per IR instance" comment was the reason for the miss, not a safety argument. The guard now lives on the Compiler as you suggested, with one refinement: **two maps with two distinct spans**, because a single path set false-positives on every compile unit (a unit's own root is legitimately on it while lowering descends):

- `code_roots` spans each compile unit's lower+emit (pushed by `compile`, `compileMultiple`, `compileExprViaIR`, and the define/set value compiles). Re-entering a live root is the back-edge, diagnosed at the entry itself.
- `code_path` spans one lowering/qq-walk tree (`lowerWithMacros`, `compileQQ`) for back-edges that never reach an entry point.

Both are consulted **through the child-compiler chain** (`initChild`'s `.parent`) — that was the second half of the bug: a lambda body compiles in a child Compiler whose maps were empty, so the let-values producer's re-entry through a wrapper lambda was invisible even with Compiler-level maps. A lambda body is a descendant of every enclosing unit, so the chain walk is exactly the ancestor relation the guard needs. I did not consolidate `CaptureScan`/`renameInBody`/`appendSplicedBodyElement` onto it (they're sound as local sets; happy to do that as a follow-up if you'd rather have one mechanism).

**2. The quasiquote splicing leak (`` `#0=(x ,@(list 1) #0#) ``)** — confirmed the KP9001. `compileQQ` now pushes its templates on the shared `code_path` (chain-aware), so the element coming back around through the fresh `(quasiquote elem)` wrapper finds its own address still on the path — fixed "for free" by the Compiler path as you predicted. Your case is in `tests_circular_code.zig` alongside the two spine templates.

One honest note from this round: my first draft also spanned the let-values *producer* itself, which false-positived the producer's legitimate first descent — `(let-values (((a b) (values 1 2))) ...)` broke, caught immediately by the existing let-values unit tests. The wrapper embeds the producer in a lambda body, whose `compileExprViaIR` already makes it a live root; the cycle's second descent trips over that, so no producer span is needed. That's documented in the commit message.

**3. The `set -e` unreachable FAIL branch** — fixed with the `|| control_status=$?` idiom, plus a comment.

**4. The bare-iterator SpineWalk nit** — both loops now call `cyclic()` with a comment noting the counting walk above already proves the spine acyclic (kept as cheap, self-documenting redundancy).

Verification: unit 1905/1912 (+7 skip), gc-stress 1912/1912, run-all 2125/2125, R7RS 1395/1395, and a 61-shape probe matrix (the issue's repros, both review repros, the CodeRabbit family, and sharing controls like `(let ((x #1=(+ 1 2))) #1#)`) shows no hang and no abort anywhere.

baijum:
Round-2 disposition (this PR was merged as b7b9cc8a before this landed; the fixes are in #2420):

**Fixed in #2420 — three real hangs, each verified on the merged tree first:**
- `case` datum list: `(case 1 (#0=(1 . #0#) 'a))` emitted eqv? comparisons forever. (Note: the auto-marker on that comment credited dc4bd141 — incorrect; it still hung there.)
- `parseSyntaxRules` literals/rules loops — cyclic literals list, before pattern validation ever runs.
- `handleTopLevelBegin` — top-level begin compiled *and executed* its rotation forever; same guard + VM detail channel as `selectCondExpandBody`.

**Not reproducible — left unchanged, with the reason:**
- and/or/when/unless spines: those forms lower through the guarded IR paths (`lowerList`/`lowerCondBody`); `(and #0=(#t . #0#))`, `(or ...)`, `(when ...)`, `(unless ...)` all return KP2002 already.
- `evalFeatureReq` and/or operand walks: a feature requirement's datum-label cycle always rotates through the requirement's own operator symbol, which evaluates as an unrecognized (false) feature, so the walk terminates by construction — every constructed shape tested terminates.

PR #2420, opened 2026-08-28
Guard the case datum list, syntax-rules outer lists, and top-level begin (#2405 follow-up)

baijum:
Follow-up to #2413 (merged as b7b9cc8a): three more walks of the #2405 family, found by CodeRabbit's re-review of the merged PR and each verified to hang on the merged tree before fixing.

## The three real hangs (now terminating with the circular-form diagnosis)

> **Diagnostic code note (baijum, review):** the two *compile-time* walks report `syntax-error[KP2002]`; the top-level `begin` splicer fires **during execution**, so the runtime reporter maps it to `error[KP2001]` — same message (`circular form in code position: …`), different code by construction. Both codes are pinned explicitly in the shell test so neither can drift unnoticed.

| Repro (hangs on b7b9cc8a) | Walk |
|---|---|
| `(case 1 (#0=(1 . #0#) 'a))` | `compileCase`'s **datum list** — the clause walk's guard does not reach the datums inside one clause; it emitted `eqv?` comparisons forever. (CodeRabbit's auto-marker credited dc4bd141 with this one; that was wrong — it still hung there.) |
| `(define-syntax m (syntax-rules #0=((). #0#) ((_ x) x)))` | `parseSyntaxRules`'s **literals/rules collection loops**, which run before pattern validation, so a cyclic literals list never reached `validPatternGrammarDepth`'s bounds. `SpineWalk` is allocation-free, keeping the no-GC-call constraint those loops document. |
| `(begin . #0=(1 . #0#))` | `handleTopLevelBegin` — the top-level begin splicer **compiled and executed** its rotation forever; same guard + VM detail channel as `selectCondExpandBody`. |

## Two findings from the same round that do not reproduce (left unchanged, with the reason)

- **`and`/`or`/`when`/`unless` spines**: those forms lower through the guarded IR paths (`lowerList`/`lowerCondBody`); all four probed shapes return KP2002 (e.g. `(and #0=(#t . #0#))` → `1:6: syntax-error[KP2002]`).
- **`evalFeatureReq` and/or operand walks**: a cond-expand feature requirement's datum-label cycle always rotates through the requirement's own operator symbol, which evaluates as an unrecognized (false) feature — so the operand walk terminates by construction. Every constructed shape tested terminates (e.g. `(cond-expand ((and #0=(srfi-1 . #0#)) 1) (else 2))` → `2`).

## Tests

One new unit test covering all three fixed shapes (the top-level begin case pinned through the VM detail channel). Suites green on this branch: unit filter 22/22, errors shell test 16/16; on the identical tree before cherry-pick: unit 1906/1913 (+7 skip), gc-stress 1913/1913, run-all 2125/2125, R7RS 1395/1395, all 70 probe shapes terminate with no hang and no abort.

Fixes the remaining findings from #2413's review round 2.

### DIFF
--- a/src/compiler_advanced.zig
+++ b/src/compiler_advanced.zig
@@ -794,13 +794,17 @@ pub fn compileParameterize(self: *Compiler, args: Value, dst: u16, is_tail: bool
     const gc = self.gc;
     const begin_sym = gc.allocSymbol("begin") catch return CompileError.OutOfMemory;
 
-    // Count bindings
+    // Count bindings. #2405 (review of PR #2420): every other binding-list
+    // walk took SpineWalk in #2413; parameterize was the odd one out — a
+    // cyclic bindings list (`(parameterize #0=((p 1) . #0#) ...)`) spun this
+    // count forever. The collect loop below is bounded by binding_count once
+    // this walk terminates.
     var binding_count: usize = 0;
-    var b = bindings;
-    while (b != types.NIL) {
-        if (!types.isPair(b)) return CompileError.InvalidSyntax;
+    var b = compiler_mod.SpineWalk.init(bindings);
+    while (b.cur != types.NIL) : (b.next()) {
+        if (b.cyclic()) return compiler_mod.circularFormError();
+        if (!types.isPair(b.cur)) return CompileError.InvalidSyntax;
         binding_count += 1;
-        b = types.cdr(b);
     }
 
     if (binding_count == 0) {
@@ -817,10 +821,10 @@ pub fn compileParameterize(self: *Compiler, args: Value, dst: u16, is_tail: bool
     var val_exprs: [32]Value = undefined;
     if (binding_count > 32) return CompileError.TooManyLocals;
 
-    b = bindings;
+    b = compiler_mod.SpineWalk.init(bindings);
     var idx: usize = 0;
-    while (b != types.NIL) : (idx += 1) {
-        const binding = types.car(b);
+    while (b.cur != types.NIL) : (b.next()) {
+        const binding = types.car(b.cur);
         if (!types.isPair(binding)) return CompileError.InvalidSyntax;
         param_exprs[idx] = types.car(binding);
         const val_rest = types.cdr(binding);
@@ -843,7 +847,7 @@ pub fn compileParameterize(self: *Compiler, args: Value, dst: u16, is_tail: bool
         const new_name = std.fmt.bufPrint(&new_buf, "%pnew{d}", .{idx}) catch return CompileError.OutOfMemory;
         new_syms[idx] = gc.allocSymbol(new_name) catch return CompileError.OutOfMemory;
 
-        b = types.cdr(b);
+        idx += 1;
     }
 
     // Build the desugared form with collection disabled: every intermediate
--- a/src/compiler_advanced.zig
+++ b/src/compiler_advanced.zig
@@ -202,7 +202,10 @@ pub fn compileCase(self: *Compiler, args: Value, dst: u16, is_tail: bool) Compil
 
         var body_jumps: std.ArrayList(usize) = .empty;
         defer body_jumps.deinit(self.gc.allocator);
-        var datum_list = datums;
+        // #2405 (CodeRabbit re-review): a cyclic datum list emitted eqv?
+        // comparisons forever — the clause walk's guard does not reach the
+        // datums inside one clause.
+        var datum_list = compiler_mod.SpineWalk.init(datums);
 
         // Allocate 3 contiguous temp registers for eqv? calls:
         // cmp_base (function + result), cmp_base+1 (arg1), cmp_base+2 (arg2)
@@ -212,10 +215,10 @@ pub fn compileCase(self: *Compiler, args: Value, dst: u16, is_tail: bool) Compil
         const cmp_arg1 = try self.allocReg();
         const cmp_arg2 = try self.allocReg();
 
-        while (datum_list != types.NIL) {
-            if (!types.isPair(datum_list)) return CompileError.InvalidSyntax;
-            const datum = types.car(datum_list);
-            datum_list = types.cdr(datum_list);
+        while (datum_list.cur != types.NIL) : (datum_list.next()) {
+            if (!types.isPair(datum_list.cur)) return CompileError.InvalidSyntax;
+            if (datum_list.cyclic()) return compiler_mod.circularFormError();
+            const datum = types.car(datum_list.cur);
 
             const datum_idx = try self.addConstant(datum);
 
--- a/src/compiler_define_syntax.zig
+++ b/src/compiler_define_syntax.zig
@@ -251,7 +251,7 @@ pub fn compileLetSyntax(self: *Compiler, args: Value, dst: u16, is_tail: bool) C
         // transformer free-references; on overflow, fall back to all siblings.
         var free_names: [64][]const u8 = undefined;
         var free_count: usize = 0;
-        const have_free = collectTransformerFreeRefs(transformer, &free_names, &free_count);
+        const have_free = try collectTransformerFreeRefs(transformer, &free_names, &free_count);
         var peer_names_f: std.ArrayList([]const u8) = .empty;
         defer peer_names_f.deinit(self.gc.allocator);
         var peer_vals_f: std.ArrayList(Value) = .empty;
@@ -401,7 +401,7 @@ fn restoreMacros(self: *Compiler, names: [][]const u8, values: []?Value) void {
 /// Collect the free-reference identifier names in a transformer's templates —
 /// identifiers that are neither the rule's pattern variables nor literals.
 /// Returns false on overflow (caller should then treat the set as unknown).
-fn collectTransformerFreeRefs(transformer: Value, out: *[64][]const u8, count: *usize) bool {
+fn collectTransformerFreeRefs(transformer: Value, out: *[64][]const u8, count: *usize) CompileError!bool {
     const tx = types.toObject(transformer).as(types.Transformer);
     var pv_names: [64][]const u8 = undefined;
     var pv_count: usize = 0;
@@ -409,7 +409,7 @@ fn collectTransformerFreeRefs(transformer: Value, out: *[64][]const u8, count: *
         if (!collectSymbols(pat, &pv_names, &pv_count)) return false;
     }
     for (tx.templates[0..tx.num_rules]) |tmpl| {
-        if (!collectFreeRefs(tmpl, pv_names[0..pv_count], tx.literals, out, count)) return false;
+        if (!(try collectFreeRefs(tmpl, pv_names[0..pv_count], tx.literals, out, count))) return false;
     }
     return true;
 }
@@ -424,7 +424,7 @@ fn computeBoundFreeRefs(self: *Compiler, transformer: Value) CompileError!void {
     var cand_names: [64][]const u8 = undefined;
     var cand_count: usize = 0;
     for (tx.templates[0..tx.num_rules]) |tmpl| {
-        if (!collectFreeRefs(tmpl, pv_names[0..pv_count], tx.literals, &cand_names, &cand_count))
+        if (!(try collectFreeRefs(tmpl, pv_names[0..pv_count], tx.literals, &cand_names, &cand_count)))
             return;
     }
     if (cand_count == 0) return;
--- a/src/compiler_define_syntax.zig
+++ b/src/compiler_define_syntax.zig
@@ -834,24 +834,28 @@ pub fn parseSyntaxRules(self: *Compiler, spec: Value, extra_bound: []const []con
     // these loops without rooting the spec first.
     var literals: std.ArrayList(Value) = .empty;
     defer literals.deinit(self.gc.allocator);
-    var lit = literals_list;
-    while (lit != types.NIL) {
-        if (!types.isPair(lit)) return CompileError.InvalidSyntax;
+    // #2405 (CodeRabbit re-review): a cyclic literal or rule list spun these
+    // collection loops forever. SpineWalk is allocation-free, keeping the
+    // no-GC-call constraint documented above.
+    var lit = compiler_mod.SpineWalk.init(literals_list);
+    while (lit.cur != types.NIL) : (lit.next()) {
+        if (!types.isPair(lit.cur)) return CompileError.InvalidSyntax;
+        if (lit.cyclic()) return compiler_mod.circularFormError();
         // A generating macro may splice a user identifier into this spec's
         // literal list (SRFI 257's if-new-var) — unwrap the provenance
         // marker so the literal is the bare identifier.
-        literals.append(self.gc.allocator, expander.unwrapUsertext(types.car(lit))) catch return CompileError.OutOfMemory;
-        lit = types.cdr(lit);
+        literals.append(self.gc.allocator, expander.unwrapUsertext(types.car(lit.cur))) catch return CompileError.OutOfMemory;
     }
 
     var patterns: std.ArrayList(Value) = .empty;
     defer patterns.deinit(self.gc.allocator);
     var templates: std.ArrayList(Value) = .empty;
     defer templates.deinit(self.gc.allocator);
-    var rule = rules;
-    while (rule != types.NIL) {
-        if (!types.isPair(rule)) return CompileError.InvalidSyntax;
-        const r = types.car(rule);
+    var rule = compiler_mod.SpineWalk.init(rules);
+    while (rule.cur != types.NIL) : (rule.next()) {
+        if (!types.isPair(rule.cur)) return CompileError.InvalidSyntax;
+        if (rule.cyclic()) return compiler_mod.circularFormError();
+        const r = types.car(rule.cur);
         if (!types.isPair(r)) return CompileError.InvalidSyntax;
         const rule_pattern = types.car(r);
         // R7RS 4.3.2 (kaappi#2082): the <pattern> grammar admits at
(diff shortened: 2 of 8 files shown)


baijum:
@coderabbitai review

src/vm_eval.zig
```diff
@@ -293,9 +293,16 @@ pub fn runCachedForm(vm: *VM, func_val: Value) VMError!Value {
...
-    var rest = body;
-    while (types.isPair(rest)) {
-        const form = types.car(rest);
+    // #2405 (CodeRabbit re-review): a cyclic top-level begin compiled and
+    // executed its rotation forever — same guard as selectCondExpandBody,
+    // reported through the VM's detail channel.
+    var rest = compiler_mod.SpineWalk.init(body);
+    while (types.isPair(rest.cur)) : (rest.next()) {
+        if (rest.cyclic()) {
+            vm.setErrorDetail("circular form in code position: the form contains itself (datum-label cycle)", .{});
```
baijum: `handleTopLevelBegin` gets the guard, but `handleDefineValues` — the sibling arm of the very same `runTopLevelHead` switch (line 189), ~180 lines below in this file — has the identical unguarded shape and still hangs:

```zig
// src/vm_eval.zig:479-489
var formal = formals;
while (formal != types.NIL) {
    if (types.isSymbol(formal)) { has_rest = true; break; }
    if (!types.isPair(formal)) return VMError.CompileError;
    if (!types.isSymbol(types.car(formal))) return VMError.CompileError;
    fixed_count += 1;
    formal = types.cdr(formal);      // <- no tortoise
}
```

```
$ echo '(define-values #0=(a . #0#) (values 1))' > /tmp/t.scm && zig-out/bin/kaappi /tmp/t.scm
<hangs; killed after 8s>
$ echo '(define-values (a . #0=(b . #0#)) (values 1 2))' > /tmp/t.scm && zig-out/bin/kaappi /tmp/t.scm
<hangs>
```

Both hang on `main` too, so it isn't a regression from this PR — but it is the same walk, in the same file, reached the same way, and it's the one arm of that switch a reader would assume this PR covered. The second binding loop at :503-507 walks the same spine again and would need the same treatment (or can rely on `fixed_count` once the first loop is bounded).

Two smaller notes on the guard as written:

- The diagnosis here surfaces as **`error[KP2001]`**, not `syntax-error[KP2002]`:
  ```
  $ echo '(begin . #0=(1 . #0#))' > /tmp/t.scm && zig-out/bin/kaappi /tmp/t.scm; echo "exit=$?"
  /tmp/t.scm:1: error[KP2001]: circular form in code position: the form contains itself (datum-label cycle)
  exit=1
  ```
  The PR body's table says all three become `syntax-error[KP2002]`. Worth either fixing the body or reconciling the codes — one condition reporting under two codes is the kind of thing that makes the next `grep -F KP2002` assertion quietly wrong. Relatedly, the message string is now copy-pasted in three places (`compiler.zig:1248`, `vm_eval.zig:302`, `vm_eval.zig:376`); a shared `const` next to `circularFormError` would keep them from drifting.
- This guard runs *during* execution, so up to two rotations of the cycle execute before it fires — `(begin . #0=((display 1) . #0#))` prints `1` and then reports. That's inherent to a splicer that evaluates as it walks and I'd leave it, but it's a real behavioural difference from every compile-time member of the family and deserves a line in the comment so nobody later reads the guard as "nothing ran".

src/compiler_advanced.zig
```diff
@@ -212,10 +215,10 @@ pub fn compileCase(self: *Compiler, args: Value, dst: u16, is_tail: bool) Compil
         const cmp_arg1 = try self.allocReg();
         const cmp_arg2 = try self.allocReg();
 
-        while (datum_list != types.NIL) {
-            if (!types.isPair(datum_list)) return CompileError.InvalidSyntax;
-            const datum = types.car(datum_list);
-            datum_list = types.cdr(datum_list);
+        while (datum_list.cur != types.NIL) : (datum_list.next()) {
+            if (!types.isPair(datum_list.cur)) return CompileError.InvalidSyntax;
+            if (datum_list.cyclic()) return compiler_mod.circularFormError();
```
baijum, line 220: The `case` datum list is guarded correctly (and I like that the guard is on the spine only, so a circular *datum* still works). But `compileParameterize` in this same file, ~590 lines below, has the untouched twin and still hangs:

```zig
// src/compiler_advanced.zig:798-804
var b = bindings;
while (b != types.NIL) {
    if (!types.isPair(b)) return CompileError.InvalidSyntax;
    binding_count += 1;
    b = types.cdr(b);        // <- no tortoise
}
```

```
$ echo '(parameterize #0=((p 1) . #0#) 1)' > /tmp/t.scm && zig-out/bin/kaappi /tmp/t.scm
<hangs; killed after 8s>
```

`sample` puts the whole stack in `compiler_advanced.compileParameterize` at :800-803. Pre-existing on `main`, not a regression here — but every other binding-list walk in the tree (`compileLet`, `compileLetStar`, `compileLetrecImpl`, `compileNamedLet`, `compileDo`, `compileLetSyntax`, `compileLetrecSyntax`) took `SpineWalk` in #2413, so `parameterize` is now the odd one out rather than part of a consistent unguarded set. It's a three-line change identical to the ones already made.

src/compiler_define_syntax.zig
```diff
@@ -834,24 +834,28 @@ pub fn parseSyntaxRules(self: *Compiler, spec: Value, extra_bound: []const []con
...
     var templates: std.ArrayList(Value) = .empty;
     defer templates.deinit(self.gc.allocator);
-    var rule = rules;
-    while (rule != types.NIL) {
-        if (!types.isPair(rule)) return CompileError.InvalidSyntax;
-        const r = types.car(rule);
+    var rule = compiler_mod.SpineWalk.init(rules);
+    while (rule.cur != types.NIL) : (rule.next()) {
+        if (!types.isPair(rule.cur)) return CompileError.InvalidSyntax;
+        if (rule.cyclic()) return compiler_mod.circularFormError();
```
baijum, line 857: These two loops are the right fix for a cyclic literals/rules list, and the allocation-free point is well made. The gap is one level in: **the template** this loop collects (`templates.append(..., types.car(r_rest))` at :872) is never spine-guarded, and it is walked at *definition* time by `compiler_macro.collectFreeRefsWithLocals`, whose last line is `return collectFreeRefsWithLocals(rest, ...)` on the cdr with no bound (`src/compiler_macro.zig:782`, car recursion at :781):

```
$ echo '(define-syntax m (syntax-rules () ((_ x) #0=(begin . #0#))))' > /tmp/t.scm && zig-out/bin/kaappi /tmp/t.scm
<hangs; killed after 8s>

$ echo '(define-syntax m (syntax-rules () ((_ x) #0=(f #0#))))' > /tmp/t.scm && zig-out/bin/kaappi /tmp/t.scm
Bus error at address 0x16d543f20
```

Note the second one is a **SIGBUS**, not a hang — the uncatchable abort class #2405 was filed for, from a one-line `define-syntax` that never even gets used. `sample` on the first shows the entire stack in `collectFreeRefsWithLocals` cycling :689 → :782. That function needs both halves of the #2413 discipline, not just `SpineWalk`: the cdr tail-recursion takes the tortoise, the car recursion at :781 needs the active-path set, exactly as `lowerWithMacros` does.

There's a third unguarded walk in the same function — the nested `sr_rules` loop at :768-774, which re-walks a `syntax-rules` spec appearing *inside* a template and so bypasses the two loops you guarded here.

Since this PR is the `syntax-rules` one, I'd fold the template walk in rather than leave `(syntax-rules () ((_ x) <cyclic>))` aborting while `(syntax-rules <cyclic> ((_ x) x))` is diagnosed.

src/tests_circular_code.zig
```diff
@@ -190,6 +190,28 @@ test "#2405 review: cycles crossing the lower/emit re-entry boundary are diagnos
...
+    // unrecognized (false) feature.
+    try expectCircular("(case 1 (#0=(1 . #0#) (quote a)))");
+    try expectCircular("(define-syntax m (syntax-rules #0=((). #0#) ((_ x) x)))");
+    // The top-level begin reports through the VM's detail channel.
+    {
+        var ctx: th.TestContext = undefined;
+        try ctx.init();
+        defer ctx.deinit();
+        try std.testing.expectError(error.CompileError, ctx.vm.eval("(begin . #0=(1 . #0#))"));
+        try std.testing.expect(std.mem.indexOf(u8, ctx.vm.getErrorDetail(), "circular form in code position") != null);
```
baijum, line 211: All three shapes fixed here were **hangs**, and they're pinned only in the unit suite — which has no time bound. If one of these guards regresses, `zig build test` will hang until the CI job timeout with no indication of which test stalled.

`tests/scheme/errors/circular-code-2405.sh` grew a bounded runner in #2413's round 2 for exactly this reason (`check_circular` kills at 10s and reports `FAIL: <name>: HUNG (killed after 10s) — the cycle guard regressed`). Adding the case and `syntax-rules` shapes there costs two lines and gets that property:

```bash
check_circular round2-case "(case 1 (#0=(1 . #0#) 'a))"
check_circular round2-syntax-rules '(define-syntax m (syntax-rules #0=((). #0#) ((_ x) x)))'
```

The top-level begin shape can't go in as-is, because `check_circular` greps for `KP2002` and that path reports `KP2001` — which is itself the argument for reconciling the two codes (see the `vm_eval.zig` comment). Either way this test should say *why* the begin case is asserted through `vm.getErrorDetail()` instead of joining its nine siblings in the shell suite; right now the asymmetry looks incidental rather than forced.

baijum (commented):
Reviewed at 1fe529e4 on top of `main` @ 25756921, ReleaseSafe / aarch64-macos, against a comparison build of `main` in a separate tree.

**All three fixes verified end to end.** Each shape hangs on `main` and terminates with the diagnosis here, so the "verified to hang on the merged tree before fixing" claim holds — including the `case` one CodeRabbit's auto-marker mis-credited to dc4bd141. Both "does not reproduce" claims also check out on this build: `(and #0=(#t . #0#))` → `1:6: syntax-error[KP2002]`, and `(cond-expand ((and #0=(srfi-1 . #0#)) 1) (else 2))` → `2` for both `and` and `or`. Suites green: unit 1818 passed / 7 skipped / 0 failed, `circular-code-2405.sh` 16/16.

Two details worth calling out as *right*, since they're the easy things to get wrong here:

- `compileCase`'s guard sits after the `else` check, the `datums == NIL` skip and the `!isPair` reject, and it guards the datum **list spine** only — so `(case 1 ((#0=(1 . #0#)) 'a))`, a genuinely circular datum compared by `eqv?` and never traversed, still compiles and exits 0. That's the correct line to draw between circular *code* and circular *data*.
- `SpineWalk` is three scalar fields with no heap use, so the no-GC-call constraint documented above the `parseSyntaxRules` loops really is preserved.

**But the family is still open, and one of the gaps is inside this PR's own subsystem.** Five more shapes, all pre-existing on `main` (none is a regression from this PR), all still hanging or aborting at 1fe529e4:

| Repro | 1fe529e4 | `main` |
|---|---|---|
| `(define-values #0=(a . #0#) (values 1))` | **HANG** | HANG |
| `(define-values (a . #0=(b . #0#)) (values 1 2))` | **HANG** | HANG |
| `(define-syntax m (syntax-rules () ((_ x) #0=(begin . #0#))))` | **HANG** | HANG |
| `(define-syntax m (syntax-rules () ((_ x) #0=(f #0#))))` | **SIGBUS** | SIGBUS |
| `(parameterize #0=((p 1) . #0#) 1)` | **HANG** | HANG |

Sites located by `sample`; details in the three inline comments. The shape of the misses is what concerns me more than the count: one is the sibling arm of the same `switch` in the same file you just fixed, one is 600 lines below the `case` fix in the same file, and one is the `syntax-rules` template — this PR guarded that spec's literals list and rules list and left the template it collects unguarded.

**Process:** this is round 3 of fixing the #2405 family one reviewer-reported shape at a time, and each round has found the next one by hand. A rough grep counts ~80 remaining raw `x = types.cdr(x)` spine advances across `compiler*.zig`, `vm_eval.zig`, `vm_library.zig`, `expander*.zig` and `ir.zig` — most are on already-validated finite lists and are fine, but nothing distinguishes those from the next `handleDefineValues`. I'd suggest one of: a sweep of the walks reachable from a user-supplied datum before this closes, or a short rule in `docs/dev/` (and ideally a CI grep like the bare-`TypeError` ratchet) saying that a walk over a reader-produced form takes `SpineWalk`. Otherwise round 4 is just waiting for the next reviewer.

baijum:
@baijum All four comments verified against the tree before fixing, then fixed in 06528dd8 — every repro you posted now terminates with the named diagnosis:

**`handleDefineValues`** — confirmed both hangs; the formals walk takes the tortoise, and the binding loop is left as-is since it is bounded by `fixed_count` once the walk terminates (your second option). Reports through the VM detail channel like its sibling arms.

**`compileParameterize`** — confirmed the hang; the count walk is now the standard `SpineWalk` idiom (parameterize was indeed the one binding-list walk the tree had missed), and the collect loop below is bounded by `binding_count`.

**Cyclic templates — the important one.** Both confirmed: the hang *and* the SIGBUS from an unused `define-syntax`. `collectFreeRefsWithLocals` now carries the full discipline, with one simplification over the prescription: since its recursion funnels through the same function on both the car and the cdr, a **single active-path membership test at the pair entry catches both cycle shapes** — no separate tortoise needed on the tail recursion. Tortoises went where the function does *not* recur per pair: the `binds`, `params`, and nested-`sr_rules` spine loops. `collectSymbols` also gained a depth cap, because the patterns of a syntax-rules spec *nested inside a template* reach it unvalidated (the spec's own patterns were already bounded by `validPatternGrammar`). The named diagnosis propagates: `collectFreeRefs`/`collectTransformerFreeRefs` became `CompileError!bool`, overflow keeps the existing set-unknown `false` degradation, and `vm_imports`' best-effort env scan degrades (`catch break`) rather than propagating — the definition site is what reports. Folded into this PR as you suggested.

**The smaller notes:** the message is now the single shared `compiler.circular_form_message` (all three copies gone); `handleTopLevelBegin`'s comment states the execution-time semantics explicitly — up to two rotations genuinely run before the tortoise meets, not a "nothing ran" guarantee. On the **KP2001/KP2002** split: reconciled by documenting and pinning rather than by changing the mapping — the two channels are different by construction (compile-time reporters vs the runtime reporter's `error.CompileError → invalid_syntax`), so the PR body now says which path reports which code, and `check_circular` grew an expected-code parameter with `check_circular_runtime` pinning KP2001 for the begin and define-values shapes — a future `grep -F KP2002` can no longer quietly miss the execution-time half. The two round-2 shapes you wanted out of the unbounded unit suite (`round2-case`, `round2-syntax-rules`) are in the shell test now too.

Tests: new unit test covering all seven shapes; seven new shell pins (23 total in the file, all passing). Full suites on 06528dd8: unit 1907/1914 (+7 skip), gc-stress 1914/1914, run-all 2125/2125, R7RS 1395/1395, native tier clean — and all 77 probe shapes terminate with no hang and no abort.

baijum (commented):
LGTM — all four review comments are addressed, verified at 06528dd8 with a fresh ReleaseSafe build (aarch64-macos) against a comparison build of `main`.

**Every shape I reported now terminates with the diagnosis:**

| Repro | 1fe529e4 | 06528dd8 |
|---|---|---|
| `(define-values #0=(a . #0#) (values 1))` | HANG | `error[KP2001]` |
| `(define-values (a . #0=(b . #0#)) (values 1 2))` | HANG | `error[KP2001]` |
| `(parameterize #0=((p 1) . #0#) 1)` | HANG | `syntax-error[KP2002]` |
| `(define-syntax m (syntax-rules () ((_ x) #0=(begin . #0#))))` | HANG | `syntax-error[KP2002]` |
| `(define-syntax m (syntax-rules () ((_ x) #0=(f #0#))))` | **SIGBUS** | `syntax-error[KP2002]` |

All 33 shapes across my three probe matrices terminate; no hang, no abort, and the controls still hold (`(case 1 ((#0=(1 . #0#)) 'a))` → 0, quoted circular data → 0, the two `cond-expand` non-reproducers → `2`).

Specifically on each comment:

- **`handleDefineValues`** — guarded, and the comment correctly notes the second binding loop is bounded by `fixed_count` once the first walk terminates. Same for `compileParameterize`'s collect loop under `binding_count`; the `idx += 1` move out of the continue-expression is safe (no `continue` in either body).
- **`collectFreeRefsWithLocals`** — the right call to use the active-path set rather than a depth cap: one membership test at the single funnel does catch both the cdr and the car shape, which is why the SIGBUS variant is fixed too. Good that the three nested spines (`binds`, `params`, `sr_rules`) each got a tortoise as well.
- **KP2001/KP2002** — reconciled the way I'd have argued for: one shared `circular_form_message`, the split documented as a channel property rather than papered over, `check_circular` taught an expected-code parameter, and both codes pinned. The PR body note is accurate now.
- **Test placement** — all round-2 and round-3 shapes moved into the bounded shell suite (23/23), so a guard regression reports as `HUNG (killed after 10s)` instead of stalling the unbounded unit suite. That was the substantive half of that comment.

**Suites, this branch:** unit 1819 passed / 7 skipped / 0 failed; `run-all.sh` 2126 pass / 0 fail / 0 timeout (R7RS 1395/1395); `circular-code-2405.sh` 23/23; `zig fmt --check src/` clean.

One **non-blocking** observation, recorded rather than requested — `TEMPLATE_WALK_DEPTH_CAP = 256` decrements on *both* car and cdr, so the bound is the pair-chain depth, and for a proper list that equals its length: a `syntax-rules` template or pattern whose spine exceeds ~256 elements now takes the set-unknown degradation where it previously completed. Both consumers already document and handle that return (`out` is capped at 64 names regardless), every suite above is green, and I could not construct an observable difference against `main` in five attempts — a 600-form template still resolves its free refs and still keeps referential transparency under a shadowing use site. So I'd ship it as-is; just worth knowing that the constant is a list-length bound, not a nesting bound, if a large generated template ever turns up with a hygiene oddity. If it does, `collectSymbols` is the one walk still on the heuristic — `TemplatePath` next door is the exact device.

Nice work closing this out — three rounds, and the family is the widest it has been probed.

PR #2432, opened 2026-08-30
Restore fiber state on every wait error exit; cut the 2200 compile test to one build (#2429, #2430, #2431)

### LINKED ISSUE #2429: A catchable OOM out of runSchedulerStep resumes the waiting fiber on a sibling's register window
Found while reviewing PR #2428 (kaappi#2395). Pre-existing on `main`; not introduced by that PR and not fixed by it.

## What happens

`fiber_wait.runSchedulerStep` restores the waiting fiber (`me`) in its **epilogue** only:

```zig
// src/fiber_wait.zig:513-518 (main)
const done = ctx.isDone();
try sched.restoreFiber(my_idx);
sched.current_idx = my_idx;
me.status = .running;
vm.current_fiber = me;
return done;
```

Five `try`s inside the dispatch loop can return before that epilogue ever runs:

| line (main) | call |
|---|---|
| 428 | `try parkOnReactor(vm, sched, poll_cap_ns)` |
| 438 | `try sched.restoreFiber(next_idx)` |
| 450 | `try sched.saveCurrentFiber()` (Yielded path) |
| 497 | `try sched.saveCurrentFiber()` (errored-dispatch path) |
| 505 | `try sched.saveCurrentFiber()` (completed-dispatch path) |

All five surface `VMError.OutOfMemory`, which is **catchable** — unlike the `return VMError.Terminated` at line 401, which `errors.isUncatchable` unwinds past every handler.

By the time lines 450/497/505 run, `restoreFiber(next_idx)` (438) has already loaded a *sibling's* registers, frames, handler stack and wind stack into the VM. Line 428 is reached at the top of a later iteration, with the sibling's state still loaded (`saveCurrentFiber` copies VM→fiber; it does not restore `me`). Returning OOM from any of them leaves:

- `vm.current_fiber` = the sibling
- `sched.current_idx` = `next_idx`
- `vm.registers` / `vm.frames` / `vm.handler_stack` = the sibling's window

`me`'s bytecode then resumes against that window. A Scheme `guard` that catches the OOM continues executing the waiting fiber's code with another fiber's registers.

## Why it matters

This is the #1487 corruption shape reached by a different route. #1487 was about `scheduleForDispatch` selecting a `driving` fiber and resuming it from a stale snapshot; here the *same* fiber resumes from a stale VM window because the restore was skipped. `runSchedulerStep` is the single shared body behind `channel-receive`, `channel-send`, `fiber-join`, `thread-join!`, `mutex-lock!`, condition-variable waits and `thread-sleep!`, so every one of those inherits it.

The custom-port-callback guard at the top of the same function already documents the principle it violates here:

> "running fiber marked parked" is the precondition for the #1487 dispatch-from-stale-snapshot corruption, and four lines here is cheaper than proving no future caller reaches it.

## Reachability

OOM-gated, which is why this is `medium` rather than `high`: it needs an allocation failure inside a scheduler drive *and* a `guard` that catches it and keeps going. No reproducer exists today. It is constructible, though — `gc.oom_countdown` (see `.claude/rules/gc-safety.md` and `src/tests_gc_root_boundary.zig`) exists precisely to drive a failure at every allocation a form performs, and #1855 fixed a structurally identical deep-OOM boundary problem. **If an `oom_countdown` sweep produces a reproducer, this should move to `priority: high`** — silent wrong answers from a legal program.

## Suggested fix

Restore `me` on every exit, not just the epilogue — e.g. an `errdefer` after `saveCurrentFiber()` (line 365) that performs the epilogue's restore. Note `restoreFiber` is itself fallible, so the error path needs a best-effort variant (the capacities it grows are already reserved for `me`, since `me` was current on entry, so a re-restore of `me` specifically cannot need to grow them).

An `oom_countdown` sweep over a drive with at least one dispatched sibling would make a regression test for this.

## Related

- Companion issue for the caller-side half of the same class: the three SRFI-18 park sites that leave `me` `.waiting` on an error exit.
- PR #2428 fixed this shape for the one function it added (`parkForThreadStatus`), on review feedback.

### LINKED ISSUE #2430: Three SRFI-18 park sites leave the fiber marked .waiting (with a live timer and waiter-index entry) on a catchable error
Found while reviewing PR #2428 (kaappi#2395). Pre-existing on `main`; that PR fixed the shape for the one function it added, on review feedback, and deliberately left these three alone.

## What happens

Three park sites in `src/primitives_srfi18.zig` arm the parked state and then use bare `try` for both the timer registration and the drive, with nothing to undo it:

```zig
me.waiting_on = ...;          // no write barrier — see below
me.status = .waiting;
me.timed_out = false;
ctx.sched.enrollWaiter(me);   // #1530 index entry
if (deadline) |d| {
    me.deadline_ns = d;
    try ctx.reactor.addTimer(d, me);   // <-- fails: state stays armed
}
const done = try fiber_mod.runSchedulerStep(...); // <-- fails: same
```

| site (main) | function |
|---|---|
| `primitives_srfi18.zig:1218` | `threadJoinFn`, fiber path |
| `primitives_srfi18.zig:1654` | `mutexLockFn` |
| `primitives_srfi18.zig:1825` | `mutexUnlockFn`, condition-variable branch |

`threadJoinFn`'s site additionally never calls `removeTimer` on the error path, so a pending timer can survive and later fire against whatever fiber `addFiber` reuses that slot for.

Both failure sources return `VMError.OutOfMemory`, which is catchable. A Scheme `guard` can therefore resume on a fiber the scheduler still believes is parked — the "running fiber marked parked" state that `runSchedulerStep`'s own custom-port-callback guard exists to avoid, and the precondition for the #1487 dispatch-from-stale-snapshot corruption. The `waiter_index` tolerates a stale entry by design (`indexWakeOn` revalidates `status == .waiting`), but that tolerance is exactly what a lying `.waiting` defeats.

## Suggested fix

One `errdefer` per site, following the channel waits in `primitives_fiber.zig` (e.g. `:658-659`, `:677-678`, `:683-684`, `:856-857`), which already restore `.running` and clear `waiting_on` in their `catch` blocks. PR #2428 uses this shape in `parkForThreadStatus`:

```zig
errdefer {
    if (me.deadline_ns != null) { ctx.reactor.removeTimer(me); me.deadline_ns = null; }
    me.status = .running;
    me.timed_out = false;
    me.waiting_on = types.VOID;
}
```

## Secondary, same lines: the `waiting_on` write barrier

The channel waits barrier their `waiting_on` stores (`primitives_fiber.zig:401-402`, `425-426`, `650-651`, `669-670`, `853-854`); these three SRFI-18 sites do not. This is **not** currently a bug — a scheduler-resident fiber is marked as an unconditional root every collection via `markVMRoots` → `FiberScheduler.markRoots` → `markFiberState`, which traces `waiting_on`, and `gc_collect.zig`'s `referencesYoung` fiber arm documents the remembered-set path as "belt-and-braces, not load-bearing" for fibers. Worth closing anyway while touching these lines: the precedent is mixed, the barrier is a deduplicated append, and residency ends the moment `retireSlot` runs.

## Reachability

OOM-gated and with no known reproducer, hence `medium` rather than `high`. `gc.oom_countdown` can drive a failure at a chosen allocation (`.claude/rules/gc-safety.md`), which is the route to a regression test.

## Related

- #2429 — the same failure class one level down, in `runSchedulerStep` itself, where the error exits skip the epilogue that restores the waiting fiber's register window. Fixing that one does not fix these, and vice versa.

### LINKED ISSUE #2431: compile-define-values-order-2200.sh intermittently exceeds the 600s shell-test timeout in the Debug CI job
Noticed while triaging CI on PR #2428; unrelated to that PR's changes, which touch nothing in the compile or native tier.

## What happens

`test (ubuntu-latest, Debug)` — a **required** check — intermittently fails with:

```
TIMEOUT  tests/scheme/compile/compile-define-values-order-2200.sh  (killed after 600s)
Scheme files: 725 pass, 0 fail, 1 timeout, 6 skipped
Total:        2120 pass, 1 fail (1 from timeouts, 6 skipped)
```

Observed on:

- **main**, run [33227886876](https://github.com/kaappi/kaappi/actions/runs/33227886876) at `13fdfd04`
- PR #2428, run [33286266439](https://github.com/kaappi/kaappi/actions/runs/33286266439) at `f76c418b`

and passing on the runs either side of both, including a later run of the same PR branch. Roughly 1 in 14 recent `main` CI runs.

## Not a hang — a budget problem

The test is genuinely slow rather than nondeterministically stuck. It drives **14 separate `kaappi compile` invocations**, each of which runs the LLVM backend and links against `libkaappi_rt.a`. Measured locally on Apple Silicon in **ReleaseSafe**:

```
elapsed 74.0s   (PASS)
```

`KAAPPI_SHELL_TEST_TIMEOUT` is 600s. A Debug build on a shared GitHub runner is comfortably an order of magnitude slower than a local ReleaseSafe build, which puts this test close enough to the ceiling that ordinary runner variance decides it. Nothing in the log suggests a deadlock: the failure is the harness kill, and the same binary passes the other 30-odd `tests/scheme/compile/*.sh` scripts in the same run.

## Options

1. **Trim the invocation count.** Several of the 14 compiles likely re-verify the same property with different spellings; a handful of representative cases plus one interpreter-oracle comparison may cover #2200's regression just as well.
2. **Raise `KAAPPI_SHELL_TEST_TIMEOUT` for the Debug job specifically.** Cheapest, but only moves the ceiling — the test would still be the slowest in the suite by a wide margin.
3. **Skip or reduce it under Debug**, the way `native-loop-alloca-stack-growth-1808.sh` already SKIPs in that job.

(1) or (3) seem better than (2): the point of the Debug job is safety-check coverage, not native-backend throughput, and a required check that fails ~7% of the time trains people to re-run rather than read.

## Reproducing

```bash
zig build -Doptimize=Debug
time bash tests/scheme/compile/compile-define-values-order-2200.sh zig-out/bin/kaappi
```

baijum:
## Summary

Three issues. The first two are one bug at two levels of the same call path; the third is unrelated, and rides along because it is the intermittent failure on a *required* check that would otherwise decide this PR's own CI.

### #2429 — a drive that errors out leaves the waiting fiber on a sibling's register window

`runSchedulerStep` restored `me` in its epilogue only, and five `try`s inside the dispatch loop return without reaching it: `parkOnReactor`, `restoreFiber(next_idx)`, and the three `saveCurrentFiber` calls on the Yielded, errored-dispatch and completed-dispatch paths. All five surface `VMError.OutOfMemory`, which is *catchable* — unlike the `Terminated` return above them.

By the time those three run, `restoreFiber` has already loaded a **sibling's** registers, frames, handler stack and wind stack into the VM, and `saveCurrentFiber` only copies VM→fiber. So an OOM left `vm.current_fiber`, `sched.current_idx` and the whole window belonging to the sibling, and a Scheme `guard` catching it resumed `me`'s bytecode against another fiber's registers — the #1487 dispatch-from-stale-snapshot corruption by a different route, in the single shared body behind `channel-receive`/`-send`, `fiber-join`, `thread-join!`, `mutex-lock!`, condition-variable waits and `thread-sleep!`.

One `errdefer` after the entry `saveCurrentFiber` now runs the epilogue's restore on every error exit. Its `catch {}` cannot fire: `restoreFiber` is fallible only through four `ensureXxxCapacity` calls that return early when the need is met and never shrink, and `me` was current on entry.

### #2430 — three SRFI-18 park sites stay `.waiting` on a catchable error

`threadJoinFn`'s fiber path, `mutexLockFn` and `mutexUnlockFn`'s condvar branch arm the parked state and then reach both `addTimer` and `runSchedulerStep` with nothing to undo it. `threadJoinFn` additionally never called `removeTimer` on its error path — and was missing the *success*-path one its two siblings have, so a surviving timer could be stranded by the `deadline_ns = null` that follows and later fire against a reused slot.

PR #2428's local `Unpark` struct becomes the shared `unparkOnError` helper; the three sites `errdefer` it. Their `waiting_on` stores are also barriered, matching the channel waits (not a live bug — `markFiberState` marks resident fibers unconditionally — but residency ends at `retireSlot`).

### #2431 — the diagnosis in the issue is wrong about the mechanism

The issue attributes the 600s timeout to a Debug build being slow. **The script never builds Debug.** `build.zig` defaults to ReleaseSafe and the script ran a bare `zig build`, so on `test (ubuntu-latest, Debug)` it built ReleaseSafe artefacts into a cache holding only Debug ones — three full **cold** builds. The ReleaseSafe legs prime that cache with the job's own build first, which is exactly why only the Debug leg hit the ceiling and why it was intermittent.

Fixed by build count, not by the timeout: a shared `fixture_interpreter` helper (extracted from `bundle_fixture_binary`), one bundled program instead of two, and `-Doptimize=ReleaseSafe` named explicitly so every caller lands on one cache key. Three builds → one unique. Measured on a 4-core box: **1m28s**, all assertions passing.

## Notes for review

- **`gc.oom_countdown` cannot reach #2429's sites**, contrary to the issue's suggestion: the OOM there comes from `ensureXxxCapacity`/`growFiberXxx`, which use the raw allocator the injector does not count. The test forces the identical exit with `Terminated` instead — deterministic, and the fix is one `errdefer` covering every error return rather than a per-error patch.
- **#2430's tests reach the error through a SRFI 181 custom port whose `read!` callback blocks** — a real program raising a real catchable error, where the OOM has no reproducer. `waiting_on` is the discriminator; `status` is deliberately not asserted because `prepareTopLevelFrame` leaves the main fiber `.completed` when a form finishes, so it reads the same either way by the time `eval` returns.
- All four tests were confirmed to **fail against the pre-fix sources** with the tests in place: `waiting_on` still holding the joined fiber / mutex / condition variable, and `current_idx=1 (want 0) current_fiber_is_me=false vm.frame_count=1 (want 0, sibling has 1)`.
- The #2429 test needs **two** spinning siblings: `thread-yield!` is advisory and no-ops unless another fiber is runnable, and `me` is excluded for the whole drive — a lone sibling never yields back and spins inside `runUntil` forever.

## Checklist

- [x] `zig build test` passes
- [ ] Scheme test suites pass (`bash tests/scheme/run-all.sh`) — **not run in full.** Verified the changed script (`compile-define-values-order-2200.sh`, 3/3 pass) and `compile-preamble-gc-700.sh`, which exercises the refactored `bundle_fixture_binary`. The full suite is left to CI.
- [x] New tests added for new behavior — one in `tests_fibers.zig` (#2429), three in `tests_srfi18.zig` (#2430)
- [x] Documentation updated — `docs/dev/test-runner.md` gains `fixture_interpreter` and the cache-mode finding

Closes #2429
Closes #2430
Closes #2431

https://claude.ai/code/session_016mKdjyytpEKCSU2hepD21M

---
_Generated by [Claude Code](https://claude.ai/code/session_016mKdjyytpEKCSU2hepD21M)_

### DIFF
--- a/src/primitives_srfi18.zig
+++ b/src/primitives_srfi18.zig
@@ -355,6 +355,43 @@ fn ensureScheduler() @TypeOf(fiber_mod.ensureScheduler(undefined)) {
     return fiber_mod.ensureScheduler(vm);
 }
 
+/// Undo the parked state a SRFI-18 wait site arms (`.waiting` status, the
+/// `timed_out` flag, `waiting_on`, and any reactor deadline timer) on an
+/// error exit, so every such site can `errdefer` it.
+///
+/// Every error exit from a park must leave `me` runnable again (#2430, and
+/// PR #2428 review for the one site that already did). Both failure sources
+/// at these sites -- `reactor.addTimer` and `runSchedulerStep`'s own
+/// allocations -- return `VMError.OutOfMemory`, which is *catchable*: a
+/// Scheme `guard` can swallow it, and the fiber then keeps executing while
+/// the scheduler still believes it is parked. "Running fiber marked parked"
+/// is the precondition for the #1487 dispatch-from-stale-snapshot
+/// corruption, and the `waiter_index`'s deliberate tolerance of a stale
+/// entry (`indexWakeOn` revalidates `status == .waiting`, which is also how
+/// the success path disposes of one) is exactly what a lying `.waiting`
+/// defeats -- so clearing the status is what retires the index entry too,
+/// and no explicit de-enrolment is needed.
+///
+/// The timer is the part that outlives the fiber: `threadJoinFn` used to
+/// leave a pending deadline behind on its error path, and a timer that
+/// survives its fiber can later fire against whatever `addFiber` reuses that
+/// slot for. `removeTimer` is remove-first, so a caller's own cleanup after
+/// catching the error stays correct.
+///
+/// Mirrors the channel waits' `catch` blocks in `primitives_fiber.zig` and
+/// `runSchedulerStep`'s custom-port-callback guard, which undo the same
+/// fields. It is the caller-side half of #2429's fix, which restores the
+/// register window one level down; neither fixes the other.
+fn unparkOnError(reactor: *reactor_mod.Reactor, f: *fiber_mod.Fiber) void {
+    if (f.deadline_ns != null) {
+        reactor.removeTimer(f);
+        f.deadline_ns = null;
+    }
+    f.status = .running;
+    f.timed_out = false;
+    f.waiting_on = types.VOID;
+}
+
 // 1ms. Since #2395 this is no longer how a cross-thread wait normally
 // resolves -- the resolving thread rings this one's reactor notifier
 // (reactor.zig's cross-thread wake registry) and the wait blocks on that
@@ -1248,29 +1285,7 @@ fn parkForThreadStatus(
     // enrolment below is for.
     ctx.sched.enrollWaiter(me);
 
-    // Every error exit below must leave `me` runnable again (PR #2428
-    // review). A primitive that returns a *catchable* error -- OutOfMemory
-    // from addTimer or from runSchedulerStep's own allocations -- can be
-    // caught by a Scheme `guard`, and this fiber then keeps executing while
-    // the scheduler still believes it is parked. "Running fiber marked
-    // parked" is the precondition for the #1487 dispatch-from-stale-snapshot
-    // corruption, and the waiter_index's tolerance of a stale entry
-    // (indexWakeOn revalidates `status == .waiting`) is exactly what a lying
-    // `.waiting` defeats. Mirrors the channel waits' catch blocks in
-    // primitives_fiber.zig and runSchedulerStep's own custom-port-callback
-    // guard, which undo the same three fields.
-    const Unpark = struct {
-        fn run(c: @TypeOf(ctx), f: *fiber_mod.Fiber) void {
-            if (f.deadline_ns != null) {
-                c.reactor.removeTimer(f);
-                f.deadline_ns = null;
-            }
-            f.status = .running;
-            f.timed_out = false;
-            f.waiting_on = types.VOID;
-        }
-    };
-    errdefer Unpark.run(ctx, me);
+    errdefer unparkOnError(ctx.reactor, me); // PR #2428 review; see the helper
 
     var enrolment: fiber_mod.CrossThreadEnrolment = .{};
     defer enrolment.release();
@@ -1416,15 +1431,29 @@ fn threadJoinFn(args: []const Value) PrimitiveError!Value {
         const me = ctx.vm.current_fiber orelse return PrimitiveError.OutOfMemory;
 
         me.waiting_on = fiber_val;
+        // #2430: barriered like the channel waits' own `waiting_on` stores
+        // (primitives_fiber.zig) and parkForThreadStatus's. Belt-and-braces
+        // for a scheduler-resident fiber, which markFiberState re-traces as a
+        // root every collection -- but residency ends the moment retireSlot
+        // runs, and the barrier costs a deduplicated remembered-set append.
+        ctx.vm.gc.writeBarrier(&me.header, fiber_val);
         me.status = .waiting;
         me.timed_out = false;
         ctx.sched.enrollWaiter(me); // #1530: O(1) wake when the joined fiber completes
+        errdefer unparkOnError(ctx.reactor, me); // #2430
         if (deadline_ns) |d| {
             me.deadline_ns = d;
             try ctx.reactor.addTimer(d, me);
         }
 
         const done = try fiber_mod.runSchedulerStep(fiber_mod.TargetWait, .{ .target = target }, ctx.vm, ctx.sched, me);
+        // Matches parkForThreadStatus and mutexLockFn, which this site alone
+        // was missing: a local wake cancels the timer for us, but clearing
+        // `deadline_ns` without it would strand any that survived — and a
+        // null `deadline_ns` is exactly what stops a later unparkOnError from
+        // finding it, leaving it to fire against a reused fiber slot.
+        // removeTimer is remove-first, so it is a no-op in the normal case.
+        if (deadline_ns != null) ctx.reactor.removeTimer(me);
         me.deadline_ns = null;
 
         if (me.timed_out) {
@@ -1855,9 +1884,11 @@ fn mutexLockFn(args: []const Value) PrimitiveError!Value {
     const me = ctx.vm.current_fiber orelse return PrimitiveError.OutOfMemory;
 
     me.waiting_on = mutex_val;
+    ctx.vm.gc.writeBarrier(&me.header, mutex_val); // #2430, see threadJoinFn
     me.status = .waiting;
     me.timed_out = false;
     ctx.sched.enrollWaiter(me); // #1530: O(1) wake on mutex unlock / abandonment
+    errdefer unparkOnError(ctx.reactor, me); // #2430
     if (deadline) |d| {
         me.deadline_ns = d;
         try ctx.reactor.addTimer(d, me);
@@ -2055,9 +2086,11 @@ fn mutexUnlockFn(args: []const Value) PrimitiveError!Value {
 
         const me = ctx.vm.current_fiber orelse return PrimitiveError.OutOfMemory;
         me.waiting_on = args[1];
+        ctx.vm.gc.writeBarrier(&me.header, args[1]); // #2430, see threadJoinFn
         me.status = .waiting;
         me.timed_out = false;
         ctx.sched.enrollWaiter(me); // #1530: O(1) wake on signal / broadcast
+        errdefer unparkOnError(ctx.reactor, me); // #2430
         if (deadline) |d| {
             me.deadline_ns = d;
             try ctx.reactor.addTimer(d, me);
--- a/src/primitives_srfi18.zig
+++ b/src/primitives_srfi18.zig
@@ -369,8 +369,14 @@ fn ensureScheduler() @TypeOf(fiber_mod.ensureScheduler(undefined)) {
 /// corruption, and the `waiter_index`'s deliberate tolerance of a stale
 /// entry (`indexWakeOn` revalidates `status == .waiting`, which is also how
 /// the success path disposes of one) is exactly what a lying `.waiting`
-/// defeats -- so clearing the status is what retires the index entry too,
-/// and no explicit de-enrolment is needed.
+/// defeats.
+///
+/// That tolerance stops a stale entry waking the wrong fiber; it does not
+/// remove it. Only a later wake naming the same key would, so the enrolment
+/// is withdrawn explicitly here -- before `waiting_on` is cleared, since that
+/// field IS the key -- or an abandoned park on an object nothing ever wakes
+/// would strand a map key and list for the scheduler's lifetime, one per
+/// failure on a fresh object (PR #2432 review).
 ///
 /// The timer is the part that outlives the fiber: `threadJoinFn` used to
 /// leave a pending deadline behind on its error path, and a timer that
@@ -382,13 +388,14 @@ fn ensureScheduler() @TypeOf(fiber_mod.ensureScheduler(undefined)) {
 /// `runSchedulerStep`'s custom-port-callback guard, which undo the same
 /// fields. It is the caller-side half of #2429's fix, which restores the
 /// register window one level down; neither fixes the other.
-fn unparkOnError(reactor: *reactor_mod.Reactor, f: *fiber_mod.Fiber) void {
+fn unparkOnError(sched: *fiber_mod.FiberScheduler, reactor: *reactor_mod.Reactor, f: *fiber_mod.Fiber) void {
     if (f.deadline_ns != null) {
         reactor.removeTimer(f);
         f.deadline_ns = null;
     }
     f.status = .running;
     f.timed_out = false;
+    sched.withdrawWaiter(f, f.waiting_on);
     f.waiting_on = types.VOID;
 }
 
@@ -1285,7 +1292,7 @@ fn parkForThreadStatus(
     // enrolment below is for.
     ctx.sched.enrollWaiter(me);
 
-    errdefer unparkOnError(ctx.reactor, me); // PR #2428 review; see the helper
+    errdefer unparkOnError(ctx.sched, ctx.reactor, me); // PR #2428 review; see the helper
 
     var enrolment: fiber_mod.CrossThreadEnrolment = .{};
     defer enrolment.release();
@@ -1440,7 +1447,7 @@ fn threadJoinFn(args: []const Value) PrimitiveError!Value {
         me.status = .waiting;
         me.timed_out = false;
         ctx.sched.enrollWaiter(me); // #1530: O(1) wake when the joined fiber completes
-        errdefer unparkOnError(ctx.reactor, me); // #2430
+        errdefer unparkOnError(ctx.sched, ctx.reactor, me); // #2430
         if (deadline_ns) |d| {
             me.deadline_ns = d;
             try ctx.reactor.addTimer(d, me);
@@ -1888,7 +1895,7 @@ fn mutexLockFn(args: []const Value) PrimitiveError!Value {
     me.status = .waiting;
     me.timed_out = false;
     ctx.sched.enrollWaiter(me); // #1530: O(1) wake on mutex unlock / abandonment
-    errdefer unparkOnError(ctx.reactor, me); // #2430
+    errdefer unparkOnError(ctx.sched, ctx.reactor, me); // #2430
     if (deadline) |d| {
         me.deadline_ns = d;
         try ctx.reactor.addTimer(d, me);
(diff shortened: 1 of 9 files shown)


tests/scheme/shell-common.sh
```diff
@@ -234,6 +234,55 @@ ensure_runtime_lib() {
...
+# is what happens otherwise: `test (ubuntu-latest, Debug)` fills the cache with
+# Debug artefacts and then runs these ReleaseSafe builds stone cold, which is
+# why that leg alone kept hitting the 600s shell-test timeout.
+fixture_interpreter() {
+    local repo="$1" out interp log rc
+    out="$repo/.zig-cache/kaappi-test-fixtures"
+    interp="$out/interp/bin/kaappi"
+    log=$(mktemp)
+
+    build_lock "$repo" fixture-interp
```
baijum, line 306: `fixture_interpreter` is always called through command substitution, where Bash keeps `$$` equal to the outer script PID rather than the actual holder process (`$BASHPID`). If `run-all.sh` times out and kills that outer PID, this subshell and its `zig build` can keep running, but another waiter sees the recorded PID as dead and steals the still-live lock. The old holder can then reach `build_unlock` and remove the new holder’s lock, allowing concurrent writes to the shared prefix. Please record the actual holder identity and make unlock ownership-checked (or ensure timeout kills the complete worker tree) before sharing this install path.

src/primitives_srfi18.zig
```diff
@@ -355,6 +355,43 @@ fn ensureScheduler() @TypeOf(fiber_mod.ensureScheduler(undefined)) {
...
+/// fields. It is the caller-side half of #2429's fix, which restores the
+/// register window one level down; neither fixes the other.
+fn unparkOnError(reactor: *reactor_mod.Reactor, f: *fiber_mod.Fiber) void {
+    if (f.deadline_ns != null) {
+        reactor.removeTimer(f);
+        f.deadline_ns = null;
+    }
+    f.status = .running;
+    f.timed_out = false;
+    f.waiting_on = types.VOID;
```
baijum, line 399: Clearing `status` makes this enrollment invalid, but it does not retire it from `waiter_index`; `indexWakeOn` only compacts stale entries if this same key is later awakened. Repeated guarded failures on fresh join targets/mutexes/condition variables therefore leave one map key/list per failure indefinitely, and after `waiting_on` is cleared the key is no longer rooted by the fiber. Please explicitly withdraw this fiber/index entry using the original key before clearing `waiting_on`, and assert the waiter-index count returns to its baseline in these regression tests.

tests/scheme/shell-common.sh
```diff
@@ -234,6 +234,55 @@ ensure_runtime_lib() {
...
+# also build.zig's default, so a bare `zig build` compiles the same thing — but
+# only pinning it here documents that every caller, and every -Dbundle build
+# that follows, must name the same mode to land on one cache key. kaappi#2431
+# is what happens otherwise: `test (ubuntu-latest, Debug)` fills the cache with
+# Debug artefacts and then runs these ReleaseSafe builds stone cold, which is
+# why that leg alone kept hitting the 600s shell-test timeout.
+fixture_interpreter() {
+    local repo="$1" out interp log rc
+    out="$repo/.zig-cache/kaappi-test-fixtures"
+    interp="$out/interp/bin/kaappi"
```
baijum: This literal path is wrong for a native Windows build: `zig build --prefix` installs `kaappi.exe`, and the later `[ -x "$interp" ]` checks the literal filename rather than applying Git Bash’s execution-time `.exe` lookup. A standalone run of a bundle test on Windows with Zig installed will report a failed fixture build despite a valid executable. Please append `.exe` when `is_windows` (the `sibling_tool` documentation above calls out this exact `-x` distinction).

tests/scheme/compile/compile-define-values-order-2200.sh
```diff
@@ -57,22 +73,31 @@ bad() {
     FAIL=$((FAIL + 1))
 }
 
-# The producer of the define-values depends on an earlier top-level define, so
-# any hoisting into the preamble breaks it.
+# `(define-values (a b) ...)`'s producer reads `x`, bound by an earlier
+# top-level form, so any hoisting into the preamble breaks it. The `import`
+# is the control: it must still reach the preamble and be replayed, which is
+# what a preamble is for, and `display` resolving at all is the proof.
```
baijum: `display` does not prove that `(scheme write)` was replayed: it is also exported from `(scheme base)`, which this same program imports, so dropping the write import still produces the expected output. The folded fixture therefore no longer has an effective import-preamble control despite the comment’s claim. Please consume something unique to `(scheme write)` such as `write-simple`, or import a fixture library binding/macro whose absence changes the result.

src/primitives_srfi18.zig
baijum: Confirmed and fixed in fdd2f96.

`FiberScheduler.withdrawWaiter` is now `enrollWaiter`'s inverse, and `unparkOnError` calls it *before* clearing `waiting_on`, since that field is the key. It removes every occurrence of the slot, not just the first — `enrollWaiter`'s dedup only skips a duplicate at the tail — and frees the key once the list empties, matching what `indexWakeOn` does on the wake path.

Your reading of the rooting is right: `FiberScheduler.markRoots` traces the fibers and their state, never the `waiter_index` keys, so once the unpark cleared `waiting_on` the key was unrooted garbage. It was not reachable as a use-after-free — the map hashes the `Value` and never dereferences it, and a recycled address still has to pass `indexWakeOn`'s per-entry validation — but it could never be matched again either, so it was pure leak.

The regression tests assert both halves you asked for: that no key lists this fiber, and that the failed park added no net key. The second is a before/after delta rather than an empty-index check, because `mutex-lock!`'s holder fiber is legitimately parked on a channel by then and owns a key of its own.

Verified the assertion has teeth — with the withdrawal disabled all three fail with `indexed=true` and `waiting_on` already VOID, which is exactly the stale-unrooted-entry state you described.

---

tests/scheme/compile/compile-define-values-order-2200.sh
baijum: You're right that `display` proves nothing — it carries `.libs = LS.initMany(&.{ .scheme_base, .scheme_r5rs, .scheme_write })` in `src/primitives_io.zig`, so the `(scheme write)` import is dead weight.

But `write-simple` would not have fixed it either, and I only found that by trying it: **every built-in library's bindings are ambient in script mode.** The folded program runs identically with the import line deleted *outright* — I checked `write-simple` specifically, then the whole import line, and both still print `(1 2)` / `30`. That is also why the original pre-fold `prog.scm` worked with no import at all. So no procedure from `(scheme base)` or `(scheme write)` can serve as this control.

Your second suggestion is the one that works. Fixed in fdd2f96 by importing `(srfi 8)` and using `receive`: it comes only from `lib/srfi/8.sld`, and without its import the program dies with `undefined variable 'b'`. The bundled binary resolves the portable `.sld` through its own installed prefix, so the standalone tier still agrees with the interpreter — 3/3 passing.

---

tests/scheme/shell-common.sh
baijum: Fixed in fdd2f96, both halves.

The `$$` problem is gone by removing the reason callers used command substitution at all: `fixture_interpreter` now builds and returns status only, and `fixture_interpreter_path` is a separate pure function (no build, no lock) that is safe to substitute anywhere. Called directly, `build_lock` records the pid of the shell that actually holds it.

`build_unlock` is ownership-checked to match — it compares the recorded pid before removing, so a late holder whose lock had already been stolen can no longer delete the new holder's lock and admit a third writer to the shared prefix. That was the concrete corruption path.

One thing I want to be honest about rather than claim as solved: the residual race is still there. If `run-all.sh` kills a script on `SHELL_TIMEOUT` and its `zig build` is orphaned rather than reaped, the next waiter sees a dead pid, steals a lock that a live process is still writing under, and two builds can install into the shared prefix concurrently. That predates this PR — `build_lock`'s steal has always trusted "recorded pid is dead" to mean "no work in flight" — and closing it properly is the "kill the complete worker tree" half of your suggestion, which belongs in `run-all.sh`'s timeout handling rather than here. Happy to do it in this PR if you'd prefer it not wait; otherwise I'd file it separately.

---

tests/scheme/shell-common.sh
baijum: Fixed in fdd2f96 — `fixture_interpreter_path` appends `.exe` when `is_windows`.

`is_windows` rather than `sibling_tool`'s mirror-the-caller trick is deliberate, and I've noted why in the comment: `sibling_tool` derives its suffix from the path it was handed because runners spell that both ways, whereas this path names a file *we* produce with our own `zig build --prefix`, so the platform is the right source of truth.

Worth recording that this predates the PR — `bundle_fixture_binary` had the same literal `bin/kaappi` — but it was latent there because all four of its callers `skip_on_windows` first. `compile-define-values-order-2200.sh` only gates on `skip_without_zig`, so it is the one that would actually have reached it, exactly as you describe.

---

tests/scheme/compile/compile-define-values-order-2200.sh
```diff
@@ -57,22 +73,31 @@ bad() {
...
+(display (+ p q))
+(newline)
 SCM
 
+EXPECTED="(1 2)
+30"
+
 echo "=== interpreter (the tier oracle) ==="
 interp_status=0
 interp_out="$(interp_stdout "$KAAPPI_ABS" "$WORK" "$WORK/prog.scm")" || interp_status=$?
```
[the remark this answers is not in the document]
baijum: This one doesn't reproduce — there are no duplicate commands. Lines 96 and 120 are the *only* run of each tier:

```
$ grep -n 'interp_out=\|native_out=' tests/scheme/compile/compile-define-values-order-2200.sh
96:interp_out="$(interp_stdout "$KAAPPI_ABS" "$WORK" "$WORK/prog.scm")" || interp_status=$?
120:native_out="$("$WORK/bundle/bin/kaappi" 2> /dev/null)" || native_status=$?
```

Two matches total, one per tier. The rewrite went the other way — it *removed* a tier run, folding two bundled programs into one so the script does a single `zig build -Dbundle=` instead of two, which is the point of the #2431 commit.

I think this is the diff being read as additive: the pre-image also had an `interp_out=` and a `native_out=` at nearby line numbers, so the same variables appear on both sides of the hunk. No change made.

---

Issue #2435, opened 2026-08-30
gc.oom_countdown cannot reach raw-allocator allocations, so whole classes of OOM error paths are untestable


baijum:
Found while writing the regression test for #2429 in PR #2432. Line numbers are against `4740b54`.

## What happens

`gc.oom_countdown` fires from `GC.maybeCollect` (`src/memory.zig:568`), so it only ever intercepts allocations that go through the **GC**. Every allocation made straight from the raw allocator is invisible to it.

That exclusion is much wider than the docs suggest. `memory.zig`'s own comment names one case —

> Counted per `maybeCollect` call, so the few allocators that skip it (`allocFunction`) are not injectable

— but the actual uncovered surface includes the VM's four growable stacks (`ensureRegisterCapacity` and siblings, via `memory.allocSliceNoFill`), the fiber snapshot buffers (`growFiberRegisters`/`Frames`/`Handlers`/`Winds`), the reactor's timer heap (`addTimer`), and the scheduler's `driving_waits` list. Those are precisely the allocations behind the fiber scheduler's OOM paths.

Meanwhile `.claude/rules/gc-safety.md` and `docs/dev/testing.md` both present the injector as exhaustive:

> Sweeping `n` over a range drives a failure at every allocation a form performs.

It does not. It drives a failure at every *GC* allocation a form performs.

## Why it mattered

#2429 is an OOM bug: five `try`s inside `runSchedulerStep`'s dispatch loop skipped the epilogue that restores the waiting fiber's register window. Its own issue text suggested the natural test —

> An `oom_countdown` sweep over a drive with at least one dispatched sibling would make a regression test for this.

— and that cannot work, because all five failure sources allocate from the raw allocator. PR #2432 had to force the identical exit with `VMError.Terminated` instead. That is a sound proxy (the fix is one `errdefer` covering every error return, not a per-error patch) but it is not the OOM the bug is about, and the same wall stands in front of #2433 and any future fix in this area.

`std.testing.FailingAllocator` is not a drop-in substitute: swapping the GC's backing allocator fails allocations indiscriminately across the whole VM including construction, and `docs/dev/testing.md` already records why it is unsuitable deep in the pipeline.

## Suggested fix

Either widen the injector or stop claiming it is exhaustive; preferably both.

[...]

Issue #1530, opened 2026-07-14
FiberScheduler wake paths still O(fiber count) after #1477 (wakeWaiters / wakeChannelWaiters / mutex+condvar wakes)


baijum:
Follow-up to #1477 (fixed by #1525). That work made the **dispatch** hot path O(1) — `FiberScheduler.schedule()` (ready ring) and `addFiber()` (free-slot list) no longer scan `sched.fibers.items`. But the **wake** paths still linearly scan every fiber slot on each call, so workloads with many concurrently-live fibers remain super-linear.

## Remaining O(fiber count) scans

All at `src/fiber.zig` (a3f36183), each `for (self.fibers.items) |f|` over the full slot array:

| Fn | line | called on | wakes |
|----|------|-----------|-------|
| `wakeWaiters` | ~635 | every fiber completion (fiber-join) | all matching → full scan always |
| `wakeChannelWaiters` | ~653 | every local channel send/dequeue/close | all matching → full scan always |
| `wakeAllCondVarWaiters` | ~747 | `condition-variable-broadcast!` | all matching → full scan always |
| `wakeMutexWaiters` | ~721 | mutex unlock | first match (early-exit, so O(pos)) |
| `wakeOneCondVarWaiter` | ~734 | `condition-variable-signal!` | first match (early-exit) |
| `hasRunnableFibers` | ~612 | every idle tick before blocking | early-exits on first runnable; O(n) only when nothing runnable |

(KEP-0002 §5's `shared_waiters` registry already makes the *cross-thread* `sweepSharedWaiters` O(shared-waiters), and `markRoots` at ~759 is necessarily O(all fibers) for GC — both out of scope here.)

## Evidence

`zig build bench-fibers` at a3f36183 (post-#1525):

**Per-switch (N fibers, 50 yield rounds each, then join all):**

| fibers | ns/switch |
|---|---|
| 100 | 324 |
| 1,000 | 486 |
| 10,000 | 4,891 |

**Spawn+join, 0 yield rounds (addFiber is now O(1), so this isolates the join/wake path):**

| fibers | ns/(spawn+join) |
|---|---|
| 100 | 22,950 |
| 1,000 | 28,682 |
| 10,000 | 209,550 |

[...]

PR #2439, opened 2026-08-30
Unpark and withdraw local-channel waits on every error exit (#2433)

### LINKED ISSUE #2433: Local channel waits: channel-send's timed park has no error-path unpark, and every local park leaks its waiter_index entry
Found while fixing #2430 in PR #2432. The same two defects that PR fixed for the three SRFI-18 park sites are still present in `primitives_fiber.zig`'s local-channel waits, which that PR did not touch. Line numbers are against `4740b54` (main, post-merge).

## Gap 1 — `channelSendLocal` arms the park and then uses bare `try` twice

```zig
// src/primitives_fiber.zig:425-431
        me.status = .waiting;
        me.waiting_on = ch_val;
        vm.gc.writeBarrier(&me.header, ch_val);
        ctx.sched.enrollWaiter(me); // #1530: O(1) wake on a freed slot / close
        me.deadline_ns = d;
        try ctx.reactor.addTimer(d, me);          // <-- armed state, bare try
    }
    _ = try fiber_mod.runSchedulerStep(ChannelSendWait, .{ .ch = ch }, ctx.vm, ctx.sched, me);  // <-- same
```

This is exactly the #2430 shape, verbatim: `.waiting`, `waiting_on`, a `waiter_index` enrolment and a live reactor timer are all armed, both failure sources return catchable `VMError.OutOfMemory`, and nothing undoes any of it. A Scheme `guard` that swallows the OOM resumes on a fiber the scheduler still believes is parked — the "running fiber marked parked" state that is the precondition for the #1487 dispatch-from-stale-snapshot corruption.

Note the contrast with the *receive* side, which already gets this right: `channelReceiveLocal` uses `catch |err| { ... }` at both `:1217` and `:1225`, restoring `.running`, clearing `waiting_on`, and detaching the timer. The send side is the odd one out.

## Gap 2 — no local-channel cleanup path withdraws the waiter_index entry

PR #2432 added `FiberScheduler.withdrawWaiter` and made `unparkOnError` call it, because clearing `status` stops a stale entry waking the wrong fiber but does not *remove* it — only a later wake naming the same key would, and a park abandoned on an object nothing ever wakes strands one map key and list for the scheduler's lifetime.

The channel sites have the same problem and none of them call it:

| site | enrols at | cleanup | withdraws? |
|---|---|---|---|
| `channelSendLocal` | `:427` | none (Gap 1) | no |
| `channelReceiveLocal` | `:1215` | `:1217`, `:1225` | no |
| `channelReceiveLocal` (dispatched re-park) | `:1178` | `:1182` | no |

Each existing `catch` block clears `waiting_on` — which, as in #2430, is what removes the last thing rooting the key, since `FiberScheduler.markRoots` traces fibers and their state but never the `waiter_index` keys.

The **promoted-channel** paths are fine and need no change: they use `enrollSharedWaiter`/`removeSharedWaiter` (`:665`, `:674`, `:684`, `:880`, `:1052`), a different registry that their `catch` blocks already unwind correctly. `enrollWaiter`'s own doc comment is explicit that a promoted-channel park must not touch the local index.

## Reachability

OOM-gated with no known reproducer, so `medium` on the same reasoning as #2430 — and with the same caveat: if a reproducer appears, Gap 1 is a silent wrong answer from a legal program and belongs at `high`.

Worth noting that `gc.oom_countdown` cannot drive either of these, for the reason recorded separately: `reactor.addTimer` pushes onto a heap owned by the raw allocator, which the injector does not count.

## Suggested fix

Gap 1: an `errdefer` after `enrollWaiter` at `:427`, mirroring `primitives_srfi18.unparkOnError`. Gap 2: a `sched.withdrawWaiter(me, me.waiting_on)` before each `me.waiting_on = types.VOID` in the existing cleanup blocks. Both sides would be less error-prone as one shared helper rather than the two parallel idioms the two files now carry.

For tests, the SRFI 181 custom-port probe used in `tests_srfi18.zig` (a `read!` callback that blocks, raising `runSchedulerStep`'s custom-port-callback error) reaches these sites deterministically too, and `expectUnparked` there already asserts the exact post-conditions.

## Related

- #2430 — the same two defects at the SRFI-18 park sites, fixed in PR #2432
- #2429 — the register-window half, one level down in `runSchedulerStep`
- #1530 — the `waiter_index` this leaks into
- #1487 — the corruption a lying `.waiting` is the precondition for

baijum:
## Summary

The two defects PR #2432 fixed for the three SRFI-18 park sites were still
present in `primitives_fiber.zig`'s local-channel waits, which that PR did not
touch. This mirrors that fix (`FiberScheduler.withdrawWaiter`, the
`unparkOnError` shape) at the local-channel sites.

### Gap 1 — `channelSendLocal` armed the park then used bare `try`

`.waiting`, `waiting_on`, a #1530 `waiter_index` enrolment and a live reactor
timer were all armed, and the site then reached both `try addTimer` and
`try runSchedulerStep` with nothing to undo it. Both sources return a
*catchable* `OutOfMemory`, so a Scheme `guard` swallowing it resumes on a fiber
the scheduler still believes is parked — the #1487 dispatch-from-stale-snapshot
precondition. The receive side already had `catch` blocks; the send side was
the odd one out, and now mirrors them with explicit `catch`.

**The issue's site table is one entry short.** It lists `:427`, `:1178`,
`:1215`. There is a **third** Gap-1 instance at the dispatched yield-retry
re-park (`enrollWaiter` at `:403`, bare `try ctx.reactor.addTimer` at `:407`) —
confirmed against current source and covered here, mirroring the receive side's
dispatched re-park.

### Gap 2 — no local cleanup path withdrew the `waiter_index` entry

Clearing `status` stops a stale entry waking the wrong fiber but does not
*remove* it; only a later wake naming the same key would (`indexWakeOn`
compacts as a side effect). A park abandoned on an object nothing ever wakes
stranded one map key and list for the scheduler's lifetime, one per failure on
a fresh object. Every local error path now calls `sched.withdrawWaiter(me,
me.waiting_on)` *before* clearing `waiting_on` (which is the key). The receive
side's `runSchedulerStep` `catch` additionally never cleared `waiting_on` at
all (runSchedulerStep's #2429 epilogue restores `status` but not `waiting_on`);
it does now.

### `blockOrDeadlock` (`:1330`) — the remaining local `enrollWaiter`, left as is

It parks **deliberately** and returns the *uncatchable* `Yielded`
(`errors.isUncatchable`), so no `guard` can resume `me` while the scheduler
holds it parked — this is a live waiter a wake will compact, not an abandoned
one to withdraw. It needs neither an unpark nor a withdraw; documented inline.

### Not touched — promoted/shared-channel paths

`enrollSharedWaiter` / `removeSharedWaiter` (`:656`, `:675`, `:855`, `:873`,
`:1006`, `:1042`) use a different registry, their `catch` blocks already unwind
correctly, and `enrollWaiter`'s doc comment is explicit that a promoted-channel
park must not touch the local index.

## Tests

Two SRFI-181 custom-port probes (the mechanism the #2430 tests use) whose
`read!` callback blocks on a timed local `channel-receive` / `channel-send`,
raising `runSchedulerStep`'s custom-port-callback error from inside the armed
in-call park. Both reach the main-fiber in-call sites (`channelSendLocal` Site
B, `channelReceiveLocal` Site D) deterministically, and both **fail against the
pre-fix source** — `waiting_on` still set, entry still `indexed=true`. They
reuse `expectUnparkedAfterBlockedCallback` / `expectUnparked`, which assert
**both halves**: status/`waiting_on`/timer restored, AND the `waiter_index`
entry withdrawn (plus that the index did not grow across the failed park).

The dispatched flat-park re-park sites (Gap-1 instance at `:403`, and the
receive equivalent) arm **no** `runSchedulerStep` — only `addTimer` — so their
error path is OOM-only, which `gc.oom_countdown` cannot reach (`addTimer`
allocates from the raw allocator the injector does not see, #2435). Their fix
mirrors the tested in-call sites exactly.

## Notes

- `zig build test` passes; `zig fmt --check` on both changed files passes.
- `src/primitives_fiber.zig` is already 1600 lines, over the repo's 1500-line
  policy. This PR adds to it deliberately rather than splitting mid-fix; if a
  split is ever forced, the file's real fault line is local-channel vs.
  promoted/shared-channel paths.

## Related

- #2432 — the same two defects at the SRFI-18 park sites
- #2430 / #2429 — the caller-side and register-window halves one level down
- #1530 — the `waiter_index` this leaks into
- #1487 — the corruption a lying `.waiting` is the precondition for
- #2435 — why `gc.oom_countdown` cannot drive these sites

Closes #2433

### DIFF
--- a/src/primitives_fiber.zig
+++ b/src/primitives_fiber.zig
@@ -404,7 +404,17 @@ fn channelSendLocal(ch: *types.Channel, ch_val: Value, payload: Value, deadline_
         if (me.deadline_ns orelse deadline_ns) |d| {
             ctx.reactor.removeTimer(me);
             me.deadline_ns = d;
-            try ctx.reactor.addTimer(d, me);
+            // #2433: this was a bare `try` — the yield-retry re-park armed
+            // .waiting/waiting_on and an enrolment, then returned a catchable
+            // OutOfMemory with nothing to undo it. Mirror the receive side's
+            // dispatched re-park (channelReceiveLocal, below).
+            ctx.reactor.addTimer(d, me) catch |err| {
+                ctx.sched.withdrawWaiter(me, me.waiting_on);
+                me.status = .running;
+                me.waiting_on = types.VOID;
+                me.deadline_ns = null;
+                return err;
+            };
         }
         vm.yield_retry = true;
         return PrimitiveError.Yielded;
@@ -426,9 +436,31 @@ fn channelSendLocal(ch: *types.Channel, ch_val: Value, payload: Value, deadline_
         vm.gc.writeBarrier(&me.header, ch_val);
         ctx.sched.enrollWaiter(me); // #1530: O(1) wake on a freed slot / close
         me.deadline_ns = d;
-        try ctx.reactor.addTimer(d, me);
+        // #2433: both `try`s here armed the park and returned a *catchable*
+        // OutOfMemory with nothing to undo it — a `guard` swallowing it would
+        // resume on a fiber the scheduler still believes is parked (the #1487
+        // precondition). channelReceiveLocal already handles both; the send
+        // side was the odd one out. Mirror it with explicit catch blocks.
+        ctx.reactor.addTimer(d, me) catch |err| {
+            ctx.sched.withdrawWaiter(me, me.waiting_on);
+            me.status = .running;
+            me.waiting_on = types.VOID;
+            me.deadline_ns = null;
+            return err;
+        };
     }
-    _ = try fiber_mod.runSchedulerStep(ChannelSendWait, .{ .ch = ch }, ctx.vm, ctx.sched, me);
+    _ = fiber_mod.runSchedulerStep(ChannelSendWait, .{ .ch = ch }, ctx.vm, ctx.sched, me) catch |err| {
+        // runSchedulerStep's own epilogue (#2429) already restored `me`'s
+        // window and .running status; undo the park state armed above that it
+        // does not touch. Guarded on deadline_ns: with none, nothing above ran.
+        if (deadline_ns != null) {
+            ctx.reactor.removeTimer(me);
+            ctx.sched.withdrawWaiter(me, me.waiting_on);
+            me.waiting_on = types.VOID;
+            me.deadline_ns = null;
+        }
+        return err;
+    };
     if (deadline_ns != null) {
         ctx.reactor.removeTimer(me);
         me.deadline_ns = null;
@@ -1182,6 +1214,7 @@ fn channelReceiveFn(args: []const Value) PrimitiveError!Value {
                     ctx.reactor.addTimer(d, me) catch |err| {
                         // Terminal for this wait: don't leak the committed
                         // demand or leave park state behind (#1604 review).
+                        ctx.sched.withdrawWaiter(me, me.waiting_on); // #2433
                         me.status = .running;
                         me.waiting_on = types.VOID;
                         me.deadline_ns = null;
@@ -1215,6 +1248,7 @@ fn channelReceiveFn(args: []const Value) PrimitiveError!Value {
         ctx.sched.enrollWaiter(me); // #1530: O(1) wake on a channel send / close
         me.deadline_ns = d;
         ctx.reactor.addTimer(d, me) catch |err| {
+            ctx.sched.withdrawWaiter(me, me.waiting_on); // #2433
             me.status = .running;
             me.waiting_on = types.VOID;
             me.deadline_ns = null;
@@ -1228,6 +1262,11 @@ fn channelReceiveFn(args: []const Value) PrimitiveError!Value {
         releaseRvToken(ch, ch_val, me);
         if (deadline_ns != null) {
             ctx.reactor.removeTimer(me);
+            // #2433: withdraw the #1530 enrolment armed above. runSchedulerStep
+            // (#2429) restored status but never clears waiting_on, and only a
+            // later wake naming this key would otherwise compact the entry.
+            ctx.sched.withdrawWaiter(me, me.waiting_on);
+            me.waiting_on = types.VOID;
             me.deadline_ns = null;
         }
         return err;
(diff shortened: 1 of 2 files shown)


src/primitives_fiber.zig
```diff
@@ -426,9 +436,31 @@ fn channelSendLocal(ch: *types.Channel, ch_val: Value, payload: Value, deadline_
...
+        ctx.reactor.addTimer(d, me) catch |err| {
+            ctx.sched.withdrawWaiter(me, me.waiting_on);
+            me.status = .running;
+            me.waiting_on = types.VOID;
+            me.deadline_ns = null;
+            return err;
+        };
     }
-    _ = try fiber_mod.runSchedulerStep(ChannelSendWait, .{ .ch = ch }, ctx.vm, ctx.sched, me);
+    _ = fiber_mod.runSchedulerStep(ChannelSendWait, .{ .ch = ch }, ctx.vm, ctx.sched, me) catch |err| {
```
baijum, line 452: **Medium:** Please restore every park field here instead of relying on `runSchedulerStep`’s epilogue. Its initial `try sched.saveCurrentFiber()` in `fiber_wait.zig` executes before the #2429 `errdefer`; if that allocation fails, this handler clears the index/timer but leaves `me.status == .waiting`. An error after a timer wake can likewise leave `me.timed_out == true`, because the epilogue only forces `.running`. Since `OutOfMemory` is catchable, a Scheme `guard` can then continue with stale park state. Set `me.status = .running` and `me.timed_out = false` here as well, matching `primitives_srfi18.unparkOnError`.

src/primitives_fiber.zig
```diff
@@ -1228,6 +1262,11 @@ fn channelReceiveFn(args: []const Value) PrimitiveError!Value {
         releaseRvToken(ch, ch_val, me);
         if (deadline_ns != null) {
             ctx.reactor.removeTimer(me);
+            // #2433: withdraw the #1530 enrolment armed above. runSchedulerStep
+            // (#2429) restored status but never clears waiting_on, and only a
+            // later wake naming this key would otherwise compact the entry.
+            ctx.sched.withdrawWaiter(me, me.waiting_on);
```
baijum, line 1283: **Medium:** The receive-side handler has the same gap. `runSchedulerStep` can fail in its first `saveCurrentFiber()` before its restoration `errdefer` exists, so withdrawing the entry and clearing `waiting_on` still leaves this running fiber marked `.waiting`; the generic epilogue also does not reset a fired `timed_out` flag. Please explicitly restore `.running` and clear `timed_out` on this error path, as the SRFI-18 cleanup helper does.

src/primitives_fiber.zig
baijum: Good catch — fixed in 37158b2c. Confirmed the ordering in fiber_wait.zig: `try sched.saveCurrentFiber()` runs before the #2429 `errdefer` is armed, so an OOM there returns with `me` still `.waiting`, and the epilogue forces only `.running` (never resetting a mid-drive `timed_out`). The send-side handler now sets `me.status = .running` and `me.timed_out = false` explicitly, matching `primitives_srfi18.unparkOnError`. Not test-coverable: the custom-port probe hits runSchedulerStep's own callback-blocked guard, which pre-restores those fields before returning; the leak is on the deeper first-`saveCurrentFiber` OOM path, which the injector can't reach (#2435).

src/primitives_fiber.zig
baijum: Fixed in 37158b2c — the receive-side handler now also restores `me.status = .running` and `me.timed_out = false` before the index/timer cleanup, matching `unparkOnError`. Same root cause and same test-coverage caveat as the send side.

PR #2440, opened 2026-08-30
Give the OOM injector a raw-allocator counterpart the fiber scheduler paths need (#2435)

### LINKED ISSUE #2435: gc.oom_countdown cannot reach raw-allocator allocations, so whole classes of OOM error paths are untestable

baijum:
Closes #2435.

## Problem

`gc.oom_countdown` fires from `GC.maybeCollect`, so it only ever intercepts allocations that go through the **GC**. Every allocation made straight from the raw allocator is invisible to it — much wider than the docs implied:

- the VM's growable register/frame/handler/wind stacks (`ensureRegisterCapacity` and siblings, via `memory.allocSliceNoFill`)
- the fiber snapshot buffers (`growFiberRegisters`/`Frames`/`Handlers`/`Winds`)
- the reactor timer heap (`addTimer`)
- the scheduler's `driving_waits` list
- bignum limb scratch and the bytecode/IR/constant pools

Those are precisely the allocations behind the fiber scheduler's OOM error paths (#2429, #2433), which is why #2432 had to force the #2429 exit with `VMError.Terminated` instead of the OOM the bug is actually about. Meanwhile `docs/dev/testing.md`, `docs/dev/gc-safety-and-error-handling.md`, and `.claude/rules/gc-safety.md` all presented the injector as reaching *every* allocation a form performs.

## Fix

Both halves the issue asked for — widen the injector **and** stop overclaiming.

**`memory.OomAllocator`** — a test-only wrapper over a backing allocator with its own `countdown`, independent of `oom_countdown` (a form's raw and GC allocation sequences are unrelated, so they arm and sweep separately). Hand `allocator()` to `GC.init` and every raw allocation becomes injectable. It ticks on `alloc` and refuses `remap` once the budget is spent, so an `ArrayList`/timer-heap growth (`driving_waits.append`, `addTimer`) can't slip past an armed injector; `resize` (pure in-place) forwards untouched. `null` — the default — forwards everything, so a VM built on it constructs normally until a test arms it (unlike `std.testing.FailingAllocator`, which fails construction indiscriminately). A `builtin.is_test` guard makes a production reference a compile error.

**Docs corrected** — `oom_countdown` reaches every *GC* allocation, not every allocation; each of the three docs (and the `memory.zig` field comment) now names the raw-allocator surface and points at `OomAllocator`.

## Tests

- `src/memory.zig`: two allocator-mechanism tests — a raw `allocSliceNoFill`, and `ArrayList` growth (the `addTimer`/`driving_waits` shape).
- `src/tests_gc_root_boundary.zig`: an end-to-end test that builds a real VM on `OomAllocator` and shows its register-file growth is reachable by the raw injector — and, armed to fail its very next GC allocation, is **not** reachable by `oom_countdown`.

`zig build` (production) compiles, `zig build test` green, `zig build test -Dgc-stress=true -Dtest-filter=OomAllocator` green, `zig fmt --check` clean.

### DIFF
--- a/docs/dev/testing.md
+++ b/docs/dev/testing.md
@@ -145,6 +145,38 @@ the first handful of allocations and never reaches the expander at all.
 `oom_countdown` is gated on `builtin.is_test`, so it compiles out of every
 non-test build.
 
+**`oom_countdown` sees only GC allocations.** It fires from
+`GC.maybeCollect`, so its sweep covers every `allocXxx` (pairs, vectors,
+strings, closures, …) — and nothing else. An allocation made straight from
+the raw allocator is invisible to it: the VM's growable register/frame/
+handler/wind stacks (`ensureRegisterCapacity` and siblings), the fiber
+snapshot buffers (`growFiberRegisters`/`Frames`/`Handlers`/`Winds`), the
+reactor timer heap (`addTimer`), the scheduler's `driving_waits` list, bignum
+limb scratch, and the bytecode/IR/constant pools all bottom out in
+`memory.allocSliceNoFill` or an `ArrayList` over `gc.allocator`, never in
+`maybeCollect`. Those are exactly the allocations behind the fiber
+scheduler's OOM paths (#2429, #2433) — an `oom_countdown` sweep cannot drive
+a failure at any of them (#2435).
+
+For the raw-allocator surface, use `memory.OomAllocator`: a test-only wrapper
+with its own countdown, independent of `oom_countdown` (a form's raw and GC
+allocation sequences are unrelated). Wrap the backing allocator, hand
+`allocator()` to `GC.init`, and arm `countdown` after construction — `null`
+forwards untouched, so the VM builds normally until the sweep begins:
+
+```zig
+var oom = memory.OomAllocator.init(std.testing.allocator);
+var gc = memory.GC.init(oom.allocator());
+const vm = try th.makeTestVM(&gc);
+defer { vm.deinit(); gc.deinit(); }
+
+oom.countdown = 0;               // fail the next raw acquisition
+try std.testing.expectError(error.OutOfMemory, vm.ensureRegisterCapacity(n));
+oom.countdown = null;            // disarm
+```
+
+Both are gated to test builds and compile out of every non-test binary.
+
 ---
 
 ## Scheme Integration Tests
--- a/docs/dev/testing.md
+++ b/docs/dev/testing.md
@@ -145,18 +145,19 @@ the first handful of allocations and never reaches the expander at all.
 `oom_countdown` is gated on `builtin.is_test`, so it compiles out of every
 non-test build.
 
-**`oom_countdown` sees only GC allocations.** It fires from
-`GC.maybeCollect`, so its sweep covers every `allocXxx` (pairs, vectors,
-strings, closures, …) — and nothing else. An allocation made straight from
-the raw allocator is invisible to it: the VM's growable register/frame/
-handler/wind stacks (`ensureRegisterCapacity` and siblings), the fiber
-snapshot buffers (`growFiberRegisters`/`Frames`/`Handlers`/`Winds`), the
-reactor timer heap (`addTimer`), the scheduler's `driving_waits` list, bignum
-limb scratch, and the bytecode/IR/constant pools all bottom out in
-`memory.allocSliceNoFill` or an `ArrayList` over `gc.allocator`, never in
-`maybeCollect`. Those are exactly the allocations behind the fiber
-scheduler's OOM paths (#2429, #2433) — an `oom_countdown` sweep cannot drive
-a failure at any of them (#2435).
+**`oom_countdown` sees only allocations mediated by `maybeCollect`.** It fires
+from `GC.maybeCollect`, so its sweep covers every allocation that routes
+through it — most `allocXxx` (pairs, vectors, strings, closures, …), but *not*
+`allocSymbol` or `allocFunction`, which deliberately skip `maybeCollect`. An
+allocation made straight from the raw allocator is likewise invisible to it:
+the VM's growable register/frame/handler/wind stacks (`ensureRegisterCapacity`
+and siblings), the fiber snapshot buffers (`growFiberRegisters`/`Frames`/
+`Handlers`/`Winds`), the reactor timer heap (`addTimer`), the scheduler's
+`driving_waits` list, bignum limb scratch, and the bytecode/IR/constant pools
+all bottom out in `memory.allocSliceNoFill` or an `ArrayList` over
+`gc.allocator`, never in `maybeCollect`. Those are exactly the allocations
+behind the fiber scheduler's OOM paths (#2429, #2433) — an `oom_countdown`
+sweep cannot drive a failure at any of them (#2435).
 
 For the raw-allocator surface, use `memory.OomAllocator`: a test-only wrapper
 with its own countdown, independent of `oom_countdown` (a form's raw and GC
--- a/src/memory.zig
+++ b/src/memory.zig
@@ -152,6 +152,99 @@ pub fn dupeSliceNoFill(allocator: std.mem.Allocator, comptime T: type, data: []c
     return buf;
 }
 
+/// Test-only OOM injector for *raw-allocator* allocations — the counterpart
+/// to `GC.oom_countdown`, which only ever sees GC (`allocXxx`) allocations
+/// because it fires from `maybeCollect`. Wrap the backing allocator with this
+/// and hand `allocator()` to `GC.init`, and every allocation that goes
+/// straight to the raw allocator becomes injectable: the VM's growable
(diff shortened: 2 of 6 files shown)


src/memory.zig
```diff
@@ -152,6 +152,99 @@ pub fn dupeSliceNoFill(allocator: std.mem.Allocator, comptime T: type, data: []c
...
+    }
+    fn oomResize(ctx: *anyopaque, mem: []u8, alignment: std.mem.Alignment, new_len: usize, ret_addr: usize) bool {
+        // A resize is a pure in-place change — it acquires no new backing
+        // memory, so it is never an OOM injection point: forward untouched.
+        const self: *OomAllocator = @ptrCast(@alignCast(ctx));
+        return self.backing.vtable.resize(self.backing.ptr, mem, alignment, new_len, ret_addr);
+    }
+    fn oomRemap(ctx: *anyopaque, mem: []u8, alignment: std.mem.Alignment, new_len: usize, ret_addr: usize) ?[*]u8 {
+        const self: *OomAllocator = @ptrCast(@alignCast(ctx));
+        if (self.spent()) return null;
```
baijum: **Medium:** Successful growth through `remap`/`resize` must participate in the countdown. Zig’s allocator contract allows both operations to expand an allocation, and a successful `remap` means the fallback `alloc` is never called. With `countdown = 1`, this forwards a successful remap without decrementing, so subsequent `ArrayList`/`PriorityQueue` growth can keep bypassing injection; `resize` can even grow with `countdown = 0`. I reproduced this with a `FixedBufferAllocator`-backed `ArrayList`: after the first successful growth, the countdown was still `1`. Gate only growth (`new_len > mem.len`), fail it at zero, decrement after a successful growth, and always forward shrink/no-op operations. Please add a nonzero-countdown test starting from an existing buffer.

docs/dev/testing.md
```diff
@@ -145,6 +145,38 @@ the first handful of allocations and never reaches the expander at all.
 `oom_countdown` is gated on `builtin.is_test`, so it compiles out of every
 non-test build.
 
+**`oom_countdown` sees only GC allocations.** It fires from
+`GC.maybeCollect`, so its sweep covers every `allocXxx` (pairs, vectors,
```
baijum: **Low:** `oom_countdown` still does not cover every `allocXxx`: `allocSymbol` and `allocFunction` skip `maybeCollect` (the updated field comment in `memory.zig` already acknowledges `allocFunction`). Since this PR’s purpose includes removing the overclaim, please say “every allocation mediated by `maybeCollect`” rather than “every `allocXxx`”, and make the equivalent wording consistent in the other updated docs.

src/memory.zig
```diff
@@ -152,6 +152,99 @@ pub fn dupeSliceNoFill(allocator: std.mem.Allocator, comptime T: type, data: []c
...
+    pub fn allocator(self: *OomAllocator) std.mem.Allocator {
+        return .{ .ptr = self, .vtable = &vtable };
+    }
+
+    /// Consume one unit of budget on a new-buffer allocation. Returns true if
+    /// the budget is spent and this allocation must fail.
+    fn tick(self: *OomAllocator) bool {
+        if (self.countdown) |n| {
+            if (n == 0) return true;
+            self.countdown = n - 1;
```
baijum: **Medium:** This mutable countdown is shared unsafely with SRFI-18 children. `thread-start!` passes the parent’s `gc.allocator` into `GC.initForThread`, so every child allocator retains this same `OomAllocator` pointer. If a threaded test arms the injector while a child is allocating, `tick()` performs concurrent unsynchronized reads/writes of `?usize`, creating a data race and nondeterministic injection. Please use atomic/locked state with arm/disarm methods, or explicitly prevent this allocator from being used by threaded tests.

src/memory.zig
baijum: Fixed in 475c1e0d. `resize`/`remap` now gate on growth (`new_len > mem.len`), fail a growth once the budget is spent, and consume one unit only when the growth actually succeeds — so a successful in-place `remap` ticks the counter just like a fresh `alloc`, and `resize` can no longer grow at `countdown == 0`. Shrink/no-op forwards untouched. Added the requested nonzero-countdown test starting from an existing buffer (`OomAllocator: a successful in-place growth consumes a nonzero budget`, using a `FixedBufferAllocator` so the growth is a real in-place `remap`), plus a gating test.

docs/dev/testing.md
baijum: Fixed in 475c1e0d. All four sites (the two docs, `.claude/rules/gc-safety.md`, and the `oom_countdown` field comment) now say `oom_countdown` reaches only allocations *mediated by `maybeCollect`* — most `allocXxx`, but not `allocSymbol`/`allocFunction`, which skip it — rather than "every `allocXxx`".

src/memory.zig
baijum: Fixed in 475c1e0d by making injection thread-affine. The wrapper captures the constructing thread's id in `init`; only that thread is ever injected, and any allocation arriving on another thread (an SRFI-18 child that inherited the parent's `gc.allocator`) forwards untouched and never reads or writes `countdown`. So the mutable state is only ever touched by the owning thread — no data race — without needing atomics. Deterministic injection is single-threaded anyway ("the n-th acquisition" is only well defined on one thread). Added a test (`OomAllocator: only the owning thread is injected`) that arms on the owner, spawns a child using the same allocator, and confirms the child is never failed while the owner stays armed.

baijum:
Pushed 475c1e0d addressing the three review comments (growth now participates in the countdown; injection is thread-affine to kill the SRFI-18 child race; docs say "mediated by \`maybeCollect\`").

It also lands the payoff this tool was built for: **the #2429 OOM regression test**, `a raw-allocator OOM mid-dispatch restores the waiting fiber's window (#2429, #2435)`. It's the sibling of the existing `Terminated`-driven probe, but drives the *actual* `OutOfMemory` — failing the `growFiberXxx` snapshot save that runs while a sibling's window is loaded into the VM, exactly what `gc.oom_countdown` could never reach. Verified it fails without the errdefer:

```
drive OOM'd out on a sibling's window — current_idx=2 (want 0) current_fiber_is_me=false vm.frame_count=681 (want 0)
```

and passes with it, in both normal and `-Dgc-stress` builds.

Issue #2422, opened 2026-08-29
CLOEXEC audit leftovers: test/dev fd pairs (testing_helpers, bench_reactor) stay inheritable


baijum:
Found during the KEP-0022 Phase 1 CLOEXEC audit (#2414) and deliberately left out of that PR's scope: the production fd-opening sites were all fixed (platform.zig open helpers, `platform.pipe`, the reactor's kqueue fd, opendir streams, mkstemp, thottam's saved-stderr dup), but the test/dev-only pair constructors still create descriptors without close-on-exec:

| Site | What it opens |
|------|---------------|
| `src/testing_helpers.zig:152` | `std.c.pipe` directly (makeFdPair) — bypasses `platform.pipe`, which is CLOEXEC since Phase 1 |
| `src/testing_helpers.zig:163` | `socketpair(AF_UNIX, SOCK_STREAM, 0, ...)` (makeBidiFdPair) — type argument `0`, no `SOCK_CLOEXEC` (also absent from the macOS SDK; would need `fcntl(F_SETFD)` via `platform.setFdCloexec`) |
| `src/bench_reactor.zig:36` | `std.c.pipe` directly (makePipe) |

## Why it was deferred

These fds live only for the lifetime of a test or benchmark in the same process and never intentionally cross an exec, so nothing is wrong today. The two consequences worth fixing anyway:

1. **fd-hygiene tests can see them.** `tests_process.zig`'s child-sees-only-stdio assertion already has to snapshot the parent's own fd set and diff against it, precisely because a leaky helper from an earlier suite would otherwise fail the test spuriously. Making these CLOEXEC tightens that test toward the strict form the fresh-process Scheme test (`tests/scheme/process/process-test.scm`) already asserts.
2. **They bypass `platform.pipe`**, so a future change to pipe semantics (buffer sizing, flags) has two call sites that can drift.

## Fix

Route all three through `platform.pipe` (which sets `FD_CLOEXEC` on both ends) and add `platform.setFdCloexec` on the socketpair ends; anything that genuinely needs an inheritable pair can clear it explicitly.

## Acceptance

- No direct `std.c.pipe`/`socketpair` calls left outside `platform.zig` in test/dev code.
- The fd-hygiene unit test's pre-snapshot diff remains green under the full suite (and becomes empty when run alone).

Issue #2414, opened 2026-08-28
KEP-0022 Phase 1: (kaappi process) — Process heap type, posix_spawn, pipe ports (POSIX)


baijum:
Phase 1 of [KEP-0022: Native Subprocess Support](https://github.com/kaappi/keps/blob/main/keps/0022-subprocess-support.md) (§Implementation plan). POSIX platforms only; blocking `process-wait`; no reactor work yet (that is Phase 2).

## Scope

- **New heap type `Process`** in a new `src/types_process.zig` (re-exported from `types.zig`), per the KEP's Reference-level design: pid, `wait_handle`, encoded `status`, three port Values, pgid, waiters. Full five-switch checklist (mark/worklist/referencesYoung; objectSize/freeObject; typeName; printer — port fields are traversable children). Cross-heap defense via the header `Object.owner` check in each primitive (channel precedent, NOT ports). Add `.process` to the `gc_deep_copy.zig` refusal list (11 → 12 tags).
- **`%process-spawn`** in a new `src/primitives_process.zig`, registered `.sandbox = false, .wasm = false` (same gating as the SRFI 170 filesystem specs). `posix_spawnp(3)` via libc; redirections as `posix_spawn_file_actions_t`; `new-group:` via `POSIX_SPAWN_SETPGROUP`; `POSIX_SPAWN_CLOEXEC_DEFAULT` on macOS. No fork anywhere; no pre-exec hook.
- **Pipe ports**: `pipe2(O_CLOEXEC)`; parent ends wrapped by refactoring the internals of `fdToPort` (`src/primitives_io.zig:868`) into a shared constructor used by both the `(kaappi ffi)` primitive and spawn. Child ends stay blocking.
- **CLOEXEC audit** of every fd-opening site in `platform.zig` (and any other open sites) so a child inherits only the three stdio slots — the Linux/musl answer, since `addclosefrom_np` is glibc-2.34+ only.
- **SIGPIPE**: set `SIG_IGN` explicitly at runtime init instead of relying on Zig `std.start`'s `keep_sigpipe == false` default (embedders linking the runtime get no `std.start`). Dead-child stdin writes must surface as the ordinary KEP-0005 I/O-error condition carrying `EPIPE`.
[...]

PR #2441, opened 2026-08-30
Add (kaappi process): posix_spawn subprocess support, POSIX Phase 1 (#2414)

baijum:
## Summary

Implements **Phase 1 of KEP-0022** (#2414): a spawn-based (never fork-based) subprocess API in the core, with child stdin/stdout/stderr exposed as ordinary reactor-integrated pipe ports. **POSIX only; blocking `process-wait`; reactor reaping is Phase 2 and Windows is Phase 3.**

### New heap type — `Process`
- `src/types_process.zig` (ObjectTag `.process`), re-exported from `types.zig`. Full five-switch checklist (`markObjectContents` / `markValueInner` worklist / `referencesYoung`; `objectSize` / `freeObject`; `typeName`; printer) plus `expectedTag` and `allocProcess`. The three pipe-port `Value` fields are traced by the GC; the printer arm is atomic (`#<process N>`), so `isTraversable`/`childAt` are untouched.
- Thread-affine: every accessor checks the header `Object.owner` against the current GC id (the channel precedent), and `gc_deep_copy.zig` refuses to copy a `Process` across a channel (refusal list 11 → 12 tags).

### Library `(kaappi process)`
`.kaappi_process` Lib enum entry, exports derived from the specs table (registered `.sandbox = false, .wasm = false`). Present on POSIX; absent on WASM and, for Phase 1, Windows (new `Lib.availableOnTarget`), so `(cond-expand ((library (kaappi process)) …))` gates correctly on every target.

`spawn-process` (options `stdin:`/`stdout:`/`stderr:` with `inherit`/`pipe`/`null`/a caller port/`'stdout` merge, `directory:`, `env:` wholesale replacement, `new-group:`), `process?`, `process-pid`, `process-group`, `process-stdin`/`stdout`/`stderr` (`#f` for every non-`'pipe` spec), `process-status`, blocking `process-wait`, `process-kill` (`signal:`/`group:`, no-op after reap), `process-environment`.

### Settled Phase-1 decisions (from the issue)
- **Status encoding:** a fixnum exit code, or `(signaled . signo)`; `#f` while running.
- **Zombie discipline:** `process-wait`/`process-status` reap and store status (the normal path leaves no zombie). The background `waitpid(WNOHANG)` sweep is **deferred to Phase 2**, whose reactor reaps every exited child regardless of waiters and closes the window entirely; a per-thread registry sweep in the blocking model needs GC-integrated lifecycle tracking that belongs with that work.
- `pass-fds:` is parsed but rejected (deferred).

### CLOEXEC audit + SIGPIPE
Fixes the production fd-opening sites (the audit that also motivated #2422 for the test/dev leftovers): `platform.pipe` now sets `FD_CLOEXEC` on both ends (`pipe2` where available, else the new `platform.setFdCloexec`), the open helpers pass `O_CLOEXEC` / `O_NOINHERIT`, and the kqueue queue fd is marked close-on-exec. `platform.ignoreSigpipe` sets `SIGPIPE` to `SIG_IGN` at VM init so a write to a dead child's stdin surfaces as an EPIPE I/O error rather than killing an embedder. `fdToPort`'s core is factored into a shared `primitives_io.rawFdToPort` used by both `(kaappi ffi)` and spawn.

### Tests
- `src/tests_process.zig`: exit/status matrix, redirection matrix, signal death, `directory:`/`env:`, kill + kill-after-reap no-op, `new-group:` group kill, and direct `platform.pipe`/`setFdCloexec` CLOEXEC checks. Green under `-Dgc-stress=true`.
- `src/tests_gc_tracing.zig`: `Process` trace + remembered-set cases (all three mark switches); `src/tests_deepcopy.zig`: refusal case.
- `tests/scheme/process/process-test.scm`: end-to-end, including the child-sees-only-0/1/2 fd-hygiene assertion (run against a clean binary; the Zig harness's own inherited fds make a raw count meaningless there).
- `tests/scheme/compile/process-spawn-2414.sh`: native-tier coverage (`spawn-process` is reachable from compiled programs).

Verified compiling on `aarch64-macos`, `x86_64-windows`, and `wasm32-wasi` (the library is correctly excluded from the latter two).

Docs (`docs/dev` + a guide page) and the high-level `run-process`/timeouts are Phase 4 per the KEP.

## Checklist

- [x] `zig build test` passes (1944 tests; also green under `-Dgc-stress=true` for the process suite)
- [x] Scheme test suites pass (`bash tests/scheme/run-all.sh`: 2129 pass, 0 fail, 1 skipped — the wasm-differential, expected on Linux)
- [ ] Documentation updated (Phase 4 per the KEP)

### DIFF
--- a/src/gc_sweep.zig
+++ b/src/gc_sweep.zig
@@ -264,6 +264,7 @@ pub fn objectSize(obj: *Object) usize {
                 fiber.wind_stack.len * @sizeOf(types.WindRecord);
         },
         .channel => @sizeOf(types.Channel),
+        .process => @sizeOf(types.Process),
         .mutex => @sizeOf(types.Mutex),
         .condition_variable => @sizeOf(types.ConditionVariable),
         .srfi18_time => @sizeOf(types.Srfi18Time),
@@ -577,6 +578,14 @@ pub fn freeObject(gc: *GC, obj: *Object) void {
                 sc.release();
             }
             poisonAndDestroy(gc, types.Channel, ch);
+        },
+        .process => {
+            // Phase 1 holds no OS handle to release: the pipe ports are
+            // separate heap objects that close their own fds, and
+            // `wait_handle` is -1 (the reactor pidfd/HANDLE that Phase 2
+            // adds will be closed here). A never-waited child is reaped by
+            // the zombie sweep, not the GC finalizer.
+            poisonAndDestroy(gc, types.Process, obj.as(types.Process));
         },
         .mutex => {
             poisonAndDestroy(gc, types.Mutex, obj.as(types.Mutex));
--- a/src/primitives_process.zig
+++ b/src/primitives_process.zig
@@ -1,0 +1,693 @@
+//! `(kaappi process)` — native subprocess support, Phase 1 (KEP-0022,
+//! kaappi#2414). POSIX only: registered `.sandbox = false, .wasm = false`,
+//! and the whole file is excluded from the WASM and Windows builds (see the
+//! `all_specs` concatenation in `primitives.zig` — neither has `posix_spawn`;
+//! Windows' CreateProcess backend is Phase 3).
+//!
+//! Phase 1 spawns via `posix_spawnp(3)` with `posix_spawn_file_actions_t`
+//! redirections and offers a *blocking* `process-wait`. There is no reactor
+//! reaping yet (Phase 2), so a fiber never parks on child exit; the child's
+//! pipe ports, however, are ordinary reactor-integrated fd ports and a read
+//! that would block already parks the calling fiber.
+//!
+//! A `Process` is thread-affine: every accessor checks the header
+//! `Object.owner` against the current GC id (the channel precedent), and
+//! `gc_deep_copy.zig` refuses to copy one across a channel.
+
+const std = @import("std");
+const builtin = @import("builtin");
+const types = @import("types.zig");
+const memory = @import("memory.zig");
+const vm_mod = @import("vm.zig");
+const primitives = @import("primitives.zig");
+const platform = @import("platform.zig");
+const primitives_io = @import("primitives_io.zig");
+
+const Value = types.Value;
+const GC = memory.GC;
+const PrimitiveError = primitives.PrimitiveError;
+const LS = primitives.LibSet;
+
+// ---------------------------------------------------------------------------
+// libc process surface
+//
+// Zig's `std.c` only exposes the posix_spawn family on Darwin, so declare it
+// here (Kaappi links libc). `posix_spawn_file_actions_t` and
+// `posix_spawnattr_t` are opaque objects whose size is platform-specific; the
+// _init/_destroy pair manages their contents in place, so an over-allocated,
+// suitably aligned byte buffer works on every target (glibc's spawnattr is
+// the largest at ~336 bytes — 512 is a safe ceiling).
+// ---------------------------------------------------------------------------
+
+// Opaque posix_spawn objects — over-allocate a byte buffer large enough for
+// any target's real struct (glibc's spawnattr is the largest at ~336 bytes),
+// suitably aligned; _init/_destroy manage the contents in place. Both use the
+// same 512-byte ceiling so an undersized buffer can never silently corrupt the
+// stack in _init.
+const FileActions = extern struct {
+    buf: [512]u8 align(@alignOf(usize)) = [_]u8{0} ** 512,
+};
+const SpawnAttr = extern struct {
+    buf: [512]u8 align(@alignOf(usize)) = [_]u8{0} ** 512,
+};
+
+/// Universal across glibc/musl/macOS/BSD.
+const POSIX_SPAWN_SETPGROUP: c_short = 0x02;
+
+const O_RDONLY: c_int = 0;
+const O_WRONLY: c_int = 1;
+const WNOHANG: c_int = 1;
+const SIGTERM: i64 = 15;
+
+/// `posix_spawn_file_actions_addchdir_np` exists on glibc (>=2.29), musl
+/// (>=1.2.3), macOS and FreeBSD, but NOT on NetBSD or OpenBSD — their libc
+/// has no chdir file action at all. Reference the symbol only where it links,
+/// so `directory:` degrades to a clear error there rather than breaking the
+/// build (kaappi#2414). A portable chdir for those two BSDs is a later phase.
+const have_addchdir_np = switch (builtin.os.tag) {
+    .linux, .macos, .freebsd => true,
+    else => false,
+};
+
+extern "c" fn posix_spawnp(
+    pid: *c_int,
+    file: [*:0]const u8,
+    file_actions: ?*const FileActions,
+    attrp: ?*const SpawnAttr,
+    argv: [*]const ?[*:0]const u8,
+    envp: [*]const ?[*:0]const u8,
+) c_int;
+extern "c" fn posix_spawn_file_actions_init(fa: *FileActions) c_int;
+extern "c" fn posix_spawn_file_actions_destroy(fa: *FileActions) c_int;
+extern "c" fn posix_spawn_file_actions_adddup2(fa: *FileActions, fd: c_int, newfd: c_int) c_int;
+extern "c" fn posix_spawn_file_actions_addopen(fa: *FileActions, fd: c_int, path: [*:0]const u8, oflag: c_int, mode: c_uint) c_int;
+extern "c" fn posix_spawn_file_actions_addchdir_np(fa: *FileActions, path: [*:0]const u8) c_int;
+extern "c" fn posix_spawnattr_init(attr: *SpawnAttr) c_int;
+extern "c" fn posix_spawnattr_destroy(attr: *SpawnAttr) c_int;
+extern "c" fn posix_spawnattr_setflags(attr: *SpawnAttr, flags: c_short) c_int;
+extern "c" fn posix_spawnattr_setpgroup(attr: *SpawnAttr, pgroup: c_int) c_int;
+extern "c" fn kill(pid: c_int, sig: c_int) c_int;
+extern "c" fn killpg(pgrp: c_int, sig: c_int) c_int;
+
+// ---------------------------------------------------------------------------
+// Error raising (KEP-0005 taxonomy)
+// ---------------------------------------------------------------------------
+
+/// A spawn failure raises a file-error-family condition carrying the errno,
+/// so `(file-error? e)` fires and `posix-error-name` is meaningful — program
+/// not found vs. permission denied, etc. Mirrors
+/// `primitives_filesystem.raiseFileErrorCode`.
+fn raiseSpawnErrorCode(gc: *GC, msg_text: []const u8, irritant: Value, errno_val: c_int) PrimitiveError!Value {
+    const vm = vm_mod.vm_instance orelse return PrimitiveError.InvalidBytecode; // no VM: internal invariant
+    var msg = gc.allocString(msg_text) catch return PrimitiveError.OutOfMemory;
+    gc.pushRoot(&msg);
+    defer gc.popRoot();
+    const irritants = gc.allocPair(irritant, types.NIL) catch return PrimitiveError.OutOfMemory;
+    var irritants_root = irritants;
+    gc.pushRoot(&irritants_root);
+    defer gc.popRoot();
+    const err_obj = gc.allocErrorObject(msg, irritants_root) catch return PrimitiveError.OutOfMemory;
+    const err = types.toObject(err_obj).as(types.ErrorObject);
+    err.error_type = .file;
+    err.posix_errno = errno_val;
+    vm.current_exception = err_obj;
+    return PrimitiveError.ExceptionRaised;
+}
+
+fn raiseSpawnError(gc: *GC, msg_text: []const u8, irritant: Value) PrimitiveError!Value {
+    return raiseSpawnErrorCode(gc, msg_text, irritant, std.c._errno().*);
+}
+
+/// Check the return of a posix_spawn setup call (init/add*/setflags/setpgroup).
+/// These APIs return the error number directly rather than setting the global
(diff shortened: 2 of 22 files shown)


src/primitives_process.zig
```diff
@@ -0,0 +1,616 @@
...
+
+/// `posix_spawn_file_actions_addchdir_np` exists on glibc (>=2.29), musl
+/// (>=1.2.3), macOS and FreeBSD, but NOT on NetBSD or OpenBSD — their libc
+/// has no chdir file action at all. Reference the symbol only where it links,
+/// so `directory:` degrades to a clear error there rather than breaking the
+/// build (kaappi#2414). A portable chdir for those two BSDs is a later phase.
+const have_addchdir_np = switch (builtin.os.tag) {
+    .linux, .macos, .freebsd => true,
+    else => false,
+};
```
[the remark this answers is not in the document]
baijum: Thanks — I've addressed the NetBSD/OpenBSD and test halves, but I'm keeping FreeBSD in `have_addchdir_np`, with CI as the evidence:

- On the first commit (`fe80cf5`), which referenced `posix_spawn_file_actions_addchdir_np` **unconditionally**, `netbsd-test` and `openbsd-test` failed to link with `undefined symbol`, but **`freebsd-test` passed** (it cross-compiles + links `x86_64-freebsd` against Zig's bundled FreeBSD libc, which provides the symbol; modern FreeBSD ≥13.1 has it too). So excluding FreeBSD would be gating out a target that actually links.
- `have_addchdir_np` now excludes only NetBSD/OpenBSD, and `directory:` raises a clean, catchable "directory: is not supported on this platform" error there rather than breaking the build.
- The `directory:` Scheme test now tolerates *that specific* error (via a substring check) so it stays strict on Linux/macOS/FreeBSD while passing on NetBSD/OpenBSD — and the corresponding Zig unit test skips on those two targets.

Re-verified with clean cross-compiles for `x86_64-freebsd`, `x86_64-netbsd`, and `x86_64-openbsd` on `54ca826`.

---

baijum (commented):
I found nine blocking correctness and portability issues. Eight are reproduced locally below; the ninth is confirmed by the required NetBSD check. I left these as comments rather than duplicating the existing CodeRabbit threads.

src/primitives_process.zig
```diff
@@ -0,0 +1,696 @@
...
+
+    // directory: — change cwd first so a relative program/paths resolve
+    // against it. addchdir_np needs no pre-exec hook (glibc ≥2.29, musl
+    // ≥1.2.3, macOS, BSD).
+    if (dir_opt != null) {
+        if (comptime have_addchdir_np) {
+            const d = dir_opt.?;
+            const ds = types.toObject(d).as(types.SchemeString);
+            const dz = arena.dupeZ(u8, ds.data[0..ds.len]) catch return PrimitiveError.OutOfMemory;
+            try spawnStep(posix_spawn_file_actions_addchdir_np(&fa, dz.ptr), gc, "cannot set directory: for spawn", d);
```
baijum, line 386: **[P1] Keep the Linux release glibc floor linkable**

This symbol was introduced in glibc 2.29, but the release workflow targets glibc 2.28. I reproduced the release failure with `zig build -Dtarget=x86_64-linux-gnu.2.28`: both `kaappi` and `kaappi-lsp` fail to link with `undefined symbol: posix_spawn_file_actions_addchdir_np`, even when `directory:` is unused. Please avoid referencing this symbol for the 2.28 target (or provide a compatible fallback), and add the actual release target to CI.

src/primitives_process.zig
```diff
@@ -0,0 +1,696 @@
...
+        .null_dev => try spawnStep(posix_spawn_file_actions_addopen(&fa, 2, "/dev/null", O_WRONLY, 0), gc, "cannot open /dev/null for stderr", args[0]),
+        .caller_fd => |fd| try spawnStep(posix_spawn_file_actions_adddup2(&fa, fd, 2), gc, "file action for stderr failed", args[0]),
+        .merge => try spawnStep(posix_spawn_file_actions_adddup2(&fa, 1, 2), gc, "file action for stderr merge failed", args[0]),
+    }
+
+    // Process group: SETPGROUP + setpgroup(0) makes the child its own group
+    // leader. fd hygiene is handled entirely by the CLOEXEC audit (every fd
+    // Kaappi opens is close-on-exec) — we deliberately do NOT set
+    // POSIX_SPAWN_CLOEXEC_DEFAULT on macOS: it would also close inherited
+    // stdio (0/1/2, which carry no file action), breaking 'inherit
```
baijum, line 455: **[P1] Close inherited descriptors, not only descriptors Kaappi opens**

The KEP requires close-by-default and says a child inherits only stdio, but this relies entirely on Kaappi-created fds already having `FD_CLOEXEC`. Any descriptor inherited by Kaappi from its launcher or introduced through FFI is passed to every child. On macOS I ran Kaappi with `9</dev/null`; a spawned `sh -c 'test -e /dev/fd/9'` returned 0. This can leak sockets, lock files, and pipe ends and can keep unrelated pipes from reaching EOF. Please use a close-all/default-close spawn facility with an explicit allowlist, and test with a deliberately inheritable fd present before Kaappi starts.

src/primitives_process.zig
```diff
@@ -0,0 +1,696 @@
...
+    // (kaappi#2414 review).
+    if (new_group) {
+        try spawnStep(posix_spawnattr_setflags(&attr, POSIX_SPAWN_SETPGROUP), gc, "posix_spawnattr_setflags failed", args[0]);
+        try spawnStep(posix_spawnattr_setpgroup(&attr, 0), gc, "posix_spawnattr_setpgroup failed", args[0]);
+    }
+
+    // --- spawn ---
+    var pid: c_int = -1;
+    const rc = posix_spawnp(&pid, file, &fa, &attr, argv.ptr, envp);
```
baijum, line 464: **[P1] Restore SIGPIPE's default disposition in spawned programs**

`VM.init` now sets process-wide `SIGPIPE` to `SIG_IGN`, and ignored dispositions survive `exec`. Because these spawn attributes never use `POSIX_SPAWN_SETSIGDEF`, every child silently inherits Kaappi's runtime policy. I reproduced this with `sh -c 'kill -PIPE $$; exit 42'`: `process-wait` returns 42 instead of `(signaled . 13)`. Please add SIGPIPE to the spawn default-signal set and combine `POSIX_SPAWN_SETSIGDEF` with the optional process-group flag.

src/gc_sweep.zig
```diff
@@ -578,6 +579,14 @@ pub fn freeObject(gc: *GC, obj: *Object) void {
             }
             poisonAndDestroy(gc, types.Channel, ch);
         },
+        .process => {
+            // Phase 1 holds no OS handle to release: the pipe ports are
+            // separate heap objects that close their own fds, and
+            // `wait_handle` is -1 (the reactor pidfd/HANDLE that Phase 2
+            // adds will be closed here). A never-waited child is reaped by
+            // the zombie sweep, not the GC finalizer.
```
baijum, line 587: **[P1] Do not discard the only handle capable of reaping this child**

There is no zombie sweep in this phase. Once this object is collected, its PID is forgotten and no later code can call `waitpid` for it. I spawned ten short-lived `true` children without retaining their Process values, then kept Kaappi alive by waiting on `sleep 3`; `ps` showed all ten as `<defunct>` children. Repeating this in a service grows zombies without bound. The Phase-1 issue explicitly calls for a bounded `waitpid(WNOHANG)` sweep on blocking spawn/wait paths, so the PID needs registry-backed lifecycle tracking (or another nonblocking reaping design) before the Process object can be freed.

src/primitives_process.zig
```diff
@@ -0,0 +1,696 @@
...
+        return .{ .caller_fd = port.fd };
+    }
+    return primitives.typeError(proc, "redirection spec (a symbol or an fd-backed port)", v);
+}
+
+/// Build a null-terminated C `argv` from a Scheme list of strings, owned by
+/// `arena`. The argv the OS receives must be Zig-owned memory, never an alias
+/// into the GC heap (which a collection could move data out from under).
+fn buildCArgv(arena: std.mem.Allocator, list: Value, proc: []const u8) PrimitiveError![]?[*:0]const u8 {
+    const n = types.listLength(list) orelse return primitives.typeError(proc, "proper list of strings", list);
```
baijum, line 245: **[P1] Reject cyclic argv and env lists instead of hanging**

`types.listLength` detects improper tails but has no cycle detection, so a legal cyclic Scheme pair spins forever inside this native primitive. For example, `(let ((x (list "true"))) (set-cdr! x x) (spawn-process x))` never returns, and `--timeout 100` cannot interrupt the native loop. `buildCEnvp` has the same problem at line 265. Please use a cycle-aware proper-list walk and add regressions for both arguments.

src/primitives_process.zig
```diff
@@ -0,0 +1,696 @@
...
+    if (n == 0) return primitives.argError(proc, "argv must have at least one element (the program to run)", .{});
+    const arr = arena.alloc(?[*:0]const u8, n + 1) catch return PrimitiveError.OutOfMemory;
+    var cur = list;
+    var i: usize = 0;
+    while (cur != types.NIL) : (i += 1) {
+        const elem = types.car(cur);
+        if (!types.isString(elem)) return primitives.typeError(proc, "string", elem);
+        const s = types.toObject(elem).as(types.SchemeString);
+        const z = arena.dupeZ(u8, s.data[0..s.len]) catch return PrimitiveError.OutOfMemory;
```
baijum, line 254: **[P1] Reject embedded NULs before constructing C strings**

`dupeZ` appends a terminator but does not reject an existing NUL, so the OS receives silently truncated argv. I passed a Scheme string containing `a\0b` to `printf`; the child received only `a` and exited successfully. The same truncation exists for `directory:` at line 385 and for environment names/values at line 278. This can execute or target something different from the value the Scheme program supplied. Please reject embedded NULs in all three surfaces; environment names should also reject empty names and `=` so entries cannot be reinterpreted.

src/primitives_process.zig
```diff
@@ -0,0 +1,696 @@
...
+    const p = try checkProcessOwner(args[0], "process-wait");
+    if (args.len > 1) {
+        const kw = if (types.isSymbol(args[1])) types.symbolName(args[1]) else "";
+        if (std.mem.eql(u8, kw, "timeout:"))
+            return primitives.argError("process-wait", "timeout: needs the reactor-driven wait (a later phase); this build only supports a blocking wait", .{});
+        return primitives.argError("process-wait", "unknown option '{s}'", .{kw});
+    }
+    if (p.status == null) {
+        var status: c_int = 0;
+        const r = waitpidRetry(p.pid, &status, 0);
```
baijum, line 575: **[P1] Publish the child VM as quiescent while blocking in waitpid**

When an SRFI-18 VM calls `process-wait`, it remains in collection state `.running` for the entire blocking syscall. A parent full collection then spins in `markLiveChildRoots` until the subprocess exits. I reproduced this with a child thread waiting on `sleep 2`: allocating 20,000 pairs on the parent took `1964453/1000000` seconds, essentially the whole remaining sleep. Other indefinitely blocking native operations wrap the blocking call with `setCollectionInNative` / `setCollectionRunning`; this path needs the same treatment when `!vm.owns_globals`.

src/vm.zig
```diff
@@ -721,6 +722,12 @@ pub const VM = struct {
     collection_state: std.atomic.Value(CollectionState) = .init(.running),
 
     pub fn init(gc: *memory.GC) !VM {
+        // Establish the process-wide SIGPIPE=SIG_IGN invariant at runtime
+        // init so a write to a dead child's stdin surfaces as an EPIPE I/O
+        // error rather than killing the process — load-bearing for
+        // (kaappi process), and needed explicitly for embedders that link
+        // the runtime without std.start (kaappi#2414). Idempotent.
+        platform.ignoreSigpipe();
```
baijum, line 730: **[P1] Actually surface EPIPE after ignoring SIGPIPE**

This establishes only the signal disposition; `drainWriteBuffer` in `primitives_io.zig` explicitly swallows EPIPE/EIO, clears the pending buffer, and reports success. I spawned an immediately exiting child with `stdin: 'pipe`, waited, then wrote and flushed under a `guard`; it returned `not-raised`. That violates the KEP's stated catchable I/O-error contract and silently loses input. Please preserve errno and raise the normal I/O condition for explicit writes/flushes (while retaining any non-raising finalizer cleanup path), with a dead-child stdin regression.

src/primitives_process.zig
```diff
@@ -0,0 +1,696 @@
...
+            return primitives.argError("process-kill", "group: #t requires spawning with new-group: #t", .{});
+        const rc = killpg(@intCast(p.pgid), target_sig);
+        netbsdKillProbe("killpg", p.pgid, target_sig, rc);
+        if (rc != 0) {
+            const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
+            return raiseSpawnError(gc, "cannot signal process group", args[0]);
+        }
+    } else {
+        const rc = kill(@intCast(p.pid), target_sig);
+        netbsdKillProbe("kill", p.pid, target_sig, rc);
```
baijum: **[P1] Fix process-kill on NetBSD before exposing the library there**

The required NetBSD check now fails on this implementation. The log shows `kill(..., 9)` and `kill(..., 15)` returning 0, but the `sleep 60` children remain alive for about 63 seconds and `process-wait` reports a normal exit instead of `(signaled . n)`; two process tests fail. `killpg` succeeds. This is functional breakage beyond the existing comment about removing the diagnostic probe, so the PID-signaling path needs to be diagnosed and fixed (or the library must remain unavailable on NetBSD for Phase 1).

baijum:
Superseded by #2442 — a fresh implementation of KEP-0022 Phase 1 that resolves all nine blocking review findings from this PR (each with a regression test), root-causes the NetBSD kill breakage (the kernel drops signals posted before a spawned child's exec completes; fixed with a kqueue NOTE_EXEC barrier), and adds the OpenBSD exec-failure pre-flight. Verified on the FreeBSD/OpenBSD/NetBSD/s390x reference VMs.

PR #2442, opened 2026-08-31
Add (kaappi process): posix_spawn subprocess support, POSIX Phase 1 (#2414)

### LINKED ISSUE #2414: KEP-0022 Phase 1: (kaappi process) — Process heap type, posix_spawn, pipe ports (POSIX)

baijum:
Implements **Phase 1 of KEP-0022** (#2414): spawn-based (never fork-based) subprocess support in the core, with child stdin/stdout/stderr exposed as ordinary reactor-integrated pipe ports. **POSIX only; blocking `process-wait`; reactor reaping is Phase 2, Windows is Phase 3.**

This is a fresh PR drawing on both prior attempts (#2441 and a second local iteration), and it addresses all nine blocking findings from the #2441 review — each with a regression test.

### New heap type — `Process`
- `src/types_process.zig` (ObjectTag `.process`, 41 → 42), re-exported from `types.zig`. Full five-switch checklist (`markObjectContents` / `markValueInner` worklist / `referencesYoung`; `objectSize` / `freeObject`; `typeName`; printer) plus the `expectFields` inventory pin and a `tests_gc_tracing` trace case. The three pipe-port `Value` fields are GC-traced; the printer arm is atomic (`#<process PID running|exited N|signaled N>`), so `isTraversable`/`childAt` are untouched.
- Thread-affine: every accessor owner-checks against the current GC (the channel precedent), and `gc_deep_copy.zig` refuses to copy a `Process` across a channel/thread boundary (refusal test in `tests_deepcopy.zig`).

### Library `(kaappi process)`
New `.kaappi_process` Lib enum entry; exports derived from the specs table (`.sandbox = false, .wasm = false`, with a comptime check enforcing the wasm gate on every spec). Registered on POSIX only — absent on WASM and (via the new `Lib.windowsAvailable`) on Windows, so `(cond-expand ((library (kaappi process)) …))` gates correctly everywhere.

`spawn-process` (options `stdin:`/`stdout:`/`stderr:` with `inherit`/`pipe`/`null`/a caller fd-port/`'stdout` merge, `directory:`, `env:` wholesale replacement, `new-group:`), `process?`, `process-pid`, `process-group`, `process-stdin`/`stdout`/`stderr` (`#f` for every non-`'pipe` spec), `process-status` (now reaps non-blockingly itself), blocking `process-wait`, `process-kill` (`signal:`/`group:`, no-op after reap), `process-environment`, plus the internal positional `%process-spawn`.

### Settled Phase-1 decisions (KEP unresolved questions)
- **Status encoding:** a fixnum exit code, or `(signaled . signo)`; `#f` while running.
- **Zombie discipline:** a bounded `waitpid(WNOHANG)` sweep of the per-heap unreaped list runs on the blocking spawn/wait paths; `process-status` reaps its own target; `freeObject` makes a last-resort non-blocking reap when a Process is collected. The only remaining window (a Process collected while its child still runs) is documented on `GC.unreaped_processes` and closes fully with Phase 2's reactor reaping.

## The nine review blockers, each fixed + tested

1. **glibc 2.28 link failure** — `posix_spawn_file_actions_addchdir_np` is glibc ≥ 2.29; `has_addchdir` now gates comptime on the target glibc version, so `zig build -Dtarget=x86_64-linux-gnu.2.28` (the release floor) links again. `directory:` raises a clean catchable error where the symbol is absent (gnu < 2.29, NetBSD, OpenBSD).
2. **Inherited descriptors leaked into children** — close-by-default is now real: macOS uses `POSIX_SPAWN_CLOEXEC_DEFAULT`; everywhere else the spawner enumerates open fds (`/proc/self/fd` on Linux; an `fcntl(F_GETFD)` probe on the BSDs, whose `/dev/fd` entries are static nodes) and adds a close action for every non-CLOEXEC fd > 2. Regression: a deliberately inheritable fd (no `O_CLOEXEC`, standing in for `kaappi 9</dev/null`) is first proven open+inheritable in the parent, then proven closed in the child.
3. **Children inherited SIGPIPE=SIG_IGN** — spawn attrs now set `POSIX_SPAWN_SETSIGDEF` with SIGPIPE in the default set. Fixing this exposed that **the `POSIX_SPAWN_SETSIGDEF` flag value is not portable**: it is 0x04 on macOS/glibc/musl but 0x10 on FreeBSD/NetBSD/OpenBSD (0x04 there is `SETSCHEDPARAM`). The constant is now per-OS, and the regression (`sh -c "kill -PIPE $$; exit 42"` must die `(signaled . 13)`) pins the value on every platform. The sigset is built with Zig's own `std.posix` helpers — NetBSD's libc sigset accessors are not portably linkable symbols.
4. **Zombies from abandoned Process objects** — the sweep/registry described above; `gc_sweep.freeObject` reaps a collected unreaped Process non-blockingly instead of forgetting the pid.
5. **Cyclic argv/env hung the VM** — both list walks are now Floyd-cycle-guarded and raise a catchable error; the second env pass is additionally bounded by the first pass's count.
6. **Embedded NUL truncation** — argv elements, env names/values, and `directory:` all reject interior NULs; env names additionally reject empty and `'='`-carrying names.
7. **Blocking `process-wait` starved parent GC** — a child-thread VM (`!owns_globals`) publishes itself via `setCollectionInNative`/`setCollectionRunning` around the blocking `waitpid` (the #1933 pattern), with the error path restructured so nothing allocates while the VM is marked in-native.
8. **EPIPE was swallowed** — `drainWriteBuffer` now raises a KEP-0005 `file-error` carrying the errno on the explicit write/flush paths (`write-*`, `flush-output-port`, `set-port-position!`), while close-port and the flush-before-read stay non-raising cleanup. Regression: write+flush to a dead child's stdin under `guard` must be caught by `file-error?`.
9. **NetBSD kill breakage** — diagnosed live on the NetBSD 10.1 reference VM. NetBSD's `posix_spawn` is an in-kernel syscall whose *child* completes its exec after the parent's call has returned, and a signal posted before that exec fully finishes is **dropped outright**: 5/40 immediate SIGTERMs to a spawned `sleep` were lost (children ran to completion), a shell-issued `kill` of the same children never failed, and a 50 ms-delayed native kill was 0/15 — pinning the mechanism to exec-completion timing. A CLOEXEC-pipe EOF barrier measured *insufficient* (8/40 still lost: the kernel closes close-on-exec fds partway through exec, before the child's signal state is final). The fix is kqueue's **`NOTE_EXEC`** — the kernel's own exec-completed notification (the reactor already speaks NetBSD kqueue) — with `NOTE_EXIT` and a 20 ms timeout as fallbacks: spawn returns only once the child is signalable. Re-measured: **40/40 immediate kills deliver, 0.77 s total** (vs. 253 s of lost-kill sleeps before). Every other target has no window by construction (vfork-semantics libcs resume the parent post-exec; macOS spawns atomically in-kernel).

A tenth platform gap the VM matrix surfaced: OpenBSD's userland `posix_spawn` cannot report the child's exec failure, so spawning a missing program "succeeded" with exit 127 (POSIX-sanctioned, but it broke the catchable-file-error contract every other platform honors). A pre-flight `access(X_OK)` on slash-containing program paths raises the same errno-carrying file-error there; PATH-searched bare names keep the POSIX fallback.

An eleventh find, courtesy of the first CI round: the `openbsd-test` leg's low `ulimit -n` exposed that spawn's pipe creation lacked the EMFILE/ENFILE recovery `open(2)` has had since #2324 — a program abandoning process handles faster than the GC threshold trips spuriously failed. `pipe(2)` in the spawner now collects and retries once, with a mutation-verified regression that lowers `RLIMIT_NOFILE` deterministically.

Review fixes beyond the nine: caller-port redirects flush their pending buffered output before the spawn and reject ports the fiber scheduler has flipped non-blocking (dup2 shares the O_NONBLOCK file-description flag); the group-kill ESRCH retry now sleeps 500 µs per attempt (~100 ms total) instead of the 500 ms that contradicted its own comment; `platform.pipe` marks both ends CLOEXEC; the kqueue reactor fd, `mkstemp` fds, opendir fds, and every `platform.open*` are CLOEXEC; `thottam_proc`/`test_selection` child helpers audited.

## Tests

- `src/tests_process.zig` (~20 tests): spawn/wait/status matrix, signaled death, kill + kill-after-reap no-op, redirection matrix, env wholesale/empty, pipe round trip, both fd-hygiene tests (strong empty assertion via dup-probe — `/dev/fd` stat lies on the BSDs — plus the deliberate-inheritable-fd regression), gc-stress-scaled spawn loop, cyclic argv/env, NUL surfaces, dead-child EPIPE, child SIGPIPE disposition, printer shape, sweep semantics. Green under `-Dgc-stress=true`.
- `src/tests_gc_tracing.zig`: Process trace case + `expectFields` pin + ObjectTag count; `src/tests_deepcopy.zig`: refusal case.
- `tests/scheme/process/process-test.scm` + `process-child-script.scm` (bidirectional line protocol with a Kaappi child), wired into `run-all.sh`.
- `tests/scheme/compile/process-spawn-2414.sh`: native-tier evidence (`kaappi compile` of a spawn/pipe/wait program, interpreter as oracle).

## Verified — on the real reference machines, not just CI

| Where | What | Result |
|-------|------|--------|
| macOS ARM (host) | `zig build test`; `run-all.sh`; process suite under `-Dgc-stress=true` | 1858 pass / 0 fail (×4 runs); **2130 pass / 0 fail**; 28/28 |
| FreeBSD 15.1 aarch64 (UTM VM) | unit + thottam + run-all | **1858/0** + 88/0; run-all 2122 pass / 1 fail — the one fail is a pre-existing filesystem-audit issue (FreeBSD rejects `chmod 7777` on a regular file with EFTYPE for non-root; even system chmod fails), flagged separately |
| OpenBSD 7.9 aarch64 (UTM VM) | unit + thottam + run-all | **1858/0** + 88/0; run-all 2105 pass / 13 fail, every residual environmental: the native-`compile/` suite (compile-on-box needs the BTCFI/toolchain setup this VM lacks), the same sticky-bit audit (OpenBSD's chmod rejects it too — noted on the tracking issue), and two load timeouts (`callcc-bench` passes standalone). Process + R7RS suites fully pass |
| NetBSD 10.1 aarch64 (UTM VM) | unit + thottam + run-all; 40× immediate-kill race loop | **1858/0** + 87/0; race loop **0/40 lost** (was 5–8/40); run-all 2095 pass / 27 fail, every residual environmental and accounted: 25× native-`compile/` (base GCC cannot consume `.ll`; box lacks pkgsrc clang), 1× the same sticky-bit audit, 1× `--changed` (no git on box). Process suites pass |
| Alpine s390x (big-endian, musl, QEMU VM) | unit + thottam + run-all | **1853 pass / 0 fail** (12 platform skips) + 88/0; both process Scheme suites pass. run-all's residual failures are that tier's known scope limits (static musl has no dlopen for the FFI suites; no native backend on s390x) — CI runs units + R7RS there by design |
| Windows 11 aarch64 (UTM VM) | library-absence gate on the real box | `(cond-expand ((library (kaappi process)) …))` prints `process-absent` |
| Alpine ppc64le (musl, QEMU VM) | unit + thottam | **1853 pass / 0 fail** (12 platform skips) + 88/0 — same shape as the s390x leg (the VM needed two restarts to come back on SSH, hence this row landing after the others) |
| Cross-compile | x86_64-linux-gnu.2.28 / gnu / musl, aarch64-linux-gnu, riscv64, s390x-musl, ppc64le, x86_64-{freebsd,netbsd,openbsd}, {x86_64,aarch64}-windows, wasm32-wasi | all build; the process library is registered on none of windows/wasi |

Docs (`docs/dev` + guide page) and the high-level `run-process`/timeouts are Phase 4 per the KEP; `docs/dev/architecture.md` file tables updated.

Closes #2414

### DIFF
--- a/docs/dev/architecture.md
+++ b/docs/dev/architecture.md
@@ -57,7 +57,7 @@ Source code
 | `printer.zig` | ~1250 | Value → string: iterative label-aware print engine + hashmap cycle/sharing detection (write/display/write-shared/write-simple; exact at any depth) and the bounded diagnostic `printValue` |
 | `printer_pretty.zig` | ~320 | REPL pretty-printer (fits-or-wraps layout over the bounded diagnostic printer; re-exported as `printer.prettyPrint`) |
 
-### Heap-type domain files (split into 11 files, kaappi#1731)
+### Heap-type domain files (split into 12 files, kaappi#1731)
 
 `types.zig` re-exports every name below (`pub const Foo = types_x.Foo;`), so
 existing `types.Foo` call sites across the codebase are unaffected by which
@@ -79,6 +79,7 @@ domain-mate (`Pair`, `Symbol`, `SchemeString`, `Closure`, `Function`,
 | `types_hashtable.zig` | `HashTable`, `HashEntry`, `HashEntryState`, `CompareMode` (SRFI 69) |
 | `types_filesystem.zig` | `FileInfo`, `UserInfo`, `GroupInfo`, `DirectoryObject` (SRFI 170) |
 | `types_weakrefs.zig` | `Ephemeron`, `Guardian`, `GuardEntry`, `TransportCell` (SRFI 254) |
+| `types_process.zig` | `Process` (KEP-0022, `(kaappi process)`), waitpid status decoders |
 
 `Fiber` (`fiber.zig`) predates this split and follows neither convention:
 its struct lives outside `types.zig` entirely with no `types.Fiber` re-export.
@@ -147,6 +148,7 @@ its struct lives outside `types.zig` entirely with no `types.Fiber` re-export.
 | `primitives_srfi237.zig` | SRFI-237: R6RS record procedural layer (`(srfi 237 primitives)`) |
 | `primitives_srfi254.zig` | SRFI-254: ephemeron/guardian constructors, predicates, accessors (GC half lives in `gc_collect.zig`) |
 | `primitives_fiber.zig` | `(kaappi fibers)`: spawn, yield, fiber-join, channels |
+| `primitives_process.zig` | `(kaappi process)` (KEP-0022): posix_spawnp-based subprocess support — spawn-process, pipe ports, blocking process-wait, process-kill, zombie sweep. POSIX-only (Windows is Phase 3, WASM registers nothing) |
 | `primitives_parallel.zig` | KEP-0002: the single native primitive backing `lib/kaappi/parallel.sld` |
 | `primitives_random_port.zig` | SRFI-271 random port `%`-prefixed internals |
 | `primitives_sysinfo.zig` | System inquiry shared by SRFI 59 (vicinity), 112 (environment), 193 (command line) |
--- a/docs/dev/architecture.md
+++ b/docs/dev/architecture.md
@@ -33,7 +33,7 @@ Source code
 | **Compiler** | `compiler.zig` + 9 sub-modules | Emits register-based bytecode from IR nodes via `compileFromNode()` (in `compiler_ir.zig`). Retains `compileExpr()` for forms delegated via `passthrough`. See the [Compiler & IR](#compiler--ir-11-files) table for the per-file split. |
 | **VM** | `vm.zig` + 10 sub-modules | Executes bytecode with a growable register file, call frame stack, exception handler stack, and dynamic-wind stack (all heap-allocated, double-on-overflow; exceeding a hard cap is an uncatchable KP3008). First-class continuations via stack copying, plus a stepping debugger. |
 | **GC** | `memory.zig` | Generational (young/old) mark-and-sweep collector over an intrusive linked list, with a write barrier and remembered set for old→young references. Root tracking via `pushRoot`/`popRoot`. Triggered after N allocations. |
-| **Primitives** | 31 `primitives_*.zig` files | 696 built-in procedures organized by domain. |
+| **Primitives** | 32 `primitives_*.zig` files | 696 built-in procedures organized by domain. |
 
 ---
 
@@ -116,7 +116,7 @@ its struct lives outside `types.zig` entirely with no `types.Fiber` re-export.
 | `vm_continuations.zig` | captureContinuation, restoreContinuation, performWindTransition, callWithCC |
 | `vm_debug.zig` | Stepping debugger: breakpoints (with conditions), watch expressions, step/next/step-out/continue, up/down frame navigation, locals, backtrace |
 
-### Primitives (split into 31 files)
+### Primitives (split into 32 files)
 
 | File | Procedures |
 |------|-----------|
--- a/src/gc_alloc.zig
+++ b/src/gc_alloc.zig
@@ -1085,6 +1085,11 @@ pub fn allocChannel(self: *GC) !Value {
 /// pointer in it is valid for exactly as long as the object is.
 pub fn allocProcess(self: *GC, pid: i32, pgid: i32) !*types.Process {
     try self.maybeCollect();
+    // Reserve the tracking slot first: spawn-process is a user-driven path
+    // whose callers already handle error.OutOfMemory, so a list-growth
+    // failure must raise, not abort — and reserving before the create means
+    // no tracked-but-unenrolled (or enrolled-then-failed) Process can exist.
+    try self.unreaped_processes.ensureUnusedCapacity(self.allocator, 1);
     const proc = try self.allocator.create(types.Process);
     proc.* = .{
         .header = .{ .tag = .process },
@@ -1092,8 +1097,7 @@ pub fn allocProcess(self: *GC, pid: i32, pgid: i32) !*types.Process {
         .pgid = pgid,
     };
     self.finishAlloc(&proc.header, @sizeOf(types.Process));
-    self.unreaped_processes.append(self.allocator, proc) catch
-        @panic("GC: process tracking list OOM");
+    self.unreaped_processes.appendAssumeCapacity(proc);
     return proc;
 }
 
--- a/src/gc_alloc.zig
+++ b/src/gc_alloc.zig
@@ -1073,6 +1073,30 @@ pub fn allocChannel(self: *GC) !Value {
     return types.makePointer(&ch.header);
 }
 
+/// KEP-0022: a spawned-subprocess handle. The three port fields are FALSE
+/// here and are stored by the spawner right after; those stores happen while
+/// the Process is young, but the write barrier is applied anyway per the
+/// gc-safety rule ("when in doubt").
+///
+/// Also enrolls the Process in `unreaped_processes` (zombie discipline,
+/// KEP-0022 unresolved question 3): every unreaped Process of this heap is
+/// swept with waitpid(WNOHANG) on the blocking spawn/wait paths, and
+/// gc_sweep.freeObject removes a collected Process from the list -- so a
+/// pointer in it is valid for exactly as long as the object is.
+pub fn allocProcess(self: *GC, pid: i32, pgid: i32) !*types.Process {
+    try self.maybeCollect();
+    const proc = try self.allocator.create(types.Process);
+    proc.* = .{
+        .header = .{ .tag = .process },
+        .pid = pid,
+        .pgid = pgid,
+    };
+    self.finishAlloc(&proc.header, @sizeOf(types.Process));
+    self.unreaped_processes.append(self.allocator, proc) catch
+        @panic("GC: process tracking list OOM");
+    return proc;
+}
+
 /// KEP-0002 §6: a bounded channel, `(make-channel capacity)`. Separate
 /// from allocChannel (rather than an optional parameter) so the
 /// pre-Phase-4 call sites that construct an unbounded channel need no
--- a/src/gc_sweep.zig
+++ b/src/gc_sweep.zig
@@ -427,6 +427,19 @@ pub fn freeObject(gc: *GC, obj: *Object) void {
                 // (EAGAIN) or any other failure just drops the remainder —
                 // programs that need the data call flush-output-port or
                 // close-port instead of leaking the port to the collector.
+                //
+                // For that promise to hold, the fd must actually RETURN
+                // EAGAIN: a port the scheduler never touched (a fresh
+                // process pipe, KEP-0022) is still in blocking mode, and a
+                // blocking write into a full pipe whose reader never drains
+                // would wedge the collector itself (kaappi#2442 review).
+                // Flip it non-blocking first — a no-op for regular files,
+                // whose flush still completes in full, and irrelevant to
+                // the fd's future since it is closed right below.
+                if (comptime !platform.is_windows and !platform.is_wasm) {
+                    if (!port.nonblocking and port.write_buf != null)
+                        _ = platform.setFdNonblockingForTeardown(port.fd);
+                }
                 if (port.write_buf) |wb| {
                     var start = port.write_buf_start;
                     while (start < port.write_buf_len) {
--- a/src/gc_sweep.zig
+++ b/src/gc_sweep.zig
@@ -434,33 +434,45 @@ pub fn freeObject(gc: *GC, obj: *Object) void {
                 // blocking write into a full pipe whose reader never drains
                 // would wedge the collector itself (kaappi#2442 review).
                 // Flip it non-blocking first — a no-op for regular files,
-                // whose flush still completes in full, and irrelevant to
-                // the fd's future since it is closed right below.
+                // whose flush still completes in full. O_NONBLOCK lives on
+                // the SHARED open-file description, so the prior flags are
+                // restored before the close — dup/dup2 aliases (a child's
+                // stdio, an fd->port twin) must not come out of a
+                // collection permanently non-blocking — and if the flip
+                // itself fails, the blocking-capable drain is skipped
+                // entirely rather than risked (kaappi#2442 review).
+                var teardown_flags: ?c_int = null;
+                var drain_ok = true;
                 if (comptime !platform.is_windows and !platform.is_wasm) {
-                    if (!port.nonblocking and port.write_buf != null)
-                        _ = platform.setFdNonblockingForTeardown(port.fd);
-                }
-                if (port.write_buf) |wb| {
-                    var start = port.write_buf_start;
-                    while (start < port.write_buf_len) {
-                        // A Windows socket port must send() — CRT _write
-                        // cannot operate on a SOCKET handle — and a pipe
-                        // port in emulated non-blocking mode must keep its
-                        // quota gate: plain _write on a full pipe would
-                        // block this sweep forever, exactly the would-block
-                        // this loop promises to drop (#1608).
-                        const remaining = port.write_buf_len - start;
-                        const rc = if (platform.is_windows and port.fd_state.is_socket)
-                            platform.sockSend(port.fd, wb.ptr + start, remaining)
-                        else if (platform.is_windows and port.fd_state.is_pipe and port.nonblocking)
-                            platform.pipeWrite(port.fd, wb.ptr + start, remaining)
-                        else
-                            platform.write(port.fd, wb.ptr + start, remaining);
-                        if (rc < 0 and platform.errno(rc) == .INTR) continue;
-                        if (rc <= 0) break;
-                        start += @as(usize, @intCast(rc));
+                    if (!port.nonblocking and port.write_buf != null) {
+                        teardown_flags = platform.setFdNonblockingForTeardown(port.fd);
+                        if (teardown_flags == null) drain_ok = false;
                     }
                 }
+                if (drain_ok) {
+                    if (port.write_buf) |wb| {
+                        var start = port.write_buf_start;
+                        while (start < port.write_buf_len) {
+                            // A Windows socket port must send() — CRT _write
+                            // cannot operate on a SOCKET handle — and a pipe
+                            // port in emulated non-blocking mode must keep its
+                            // quota gate: plain _write on a full pipe would
+                            // block this sweep forever, exactly the would-block
+                            // this loop promises to drop (#1608).
+                            const remaining = port.write_buf_len - start;
+                            const rc = if (platform.is_windows and port.fd_state.is_socket)
+                                platform.sockSend(port.fd, wb.ptr + start, remaining)
+                            else if (platform.is_windows and port.fd_state.is_pipe and port.nonblocking)
+                                platform.pipeWrite(port.fd, wb.ptr + start, remaining)
+                            else
+                                platform.write(port.fd, wb.ptr + start, remaining);
+                            if (rc < 0 and platform.errno(rc) == .INTR) continue;
+                            if (rc <= 0) break;
+                            start += @as(usize, @intCast(rc));
+                        }
+                    }
+                }
+                if (teardown_flags) |fl| platform.restoreFdStatusFlags(port.fd, fl);
                 _ = platform.close(port.fd);
             }
             if (port.write_buf) |wb| {
--- a/src/gc_sweep.zig
+++ b/src/gc_sweep.zig
@@ -428,27 +428,27 @@ pub fn freeObject(gc: *GC, obj: *Object) void {
                 // programs that need the data call flush-output-port or
                 // close-port instead of leaking the port to the collector.
                 //
-                // For that promise to hold, the fd must actually RETURN
-                // EAGAIN: a port the scheduler never touched (a fresh
-                // process pipe, KEP-0022) is still in blocking mode, and a
+                // For that promise to hold, the drain must be unable to
+                // block. A port the scheduler never touched (a fresh
+                // process pipe, KEP-0022) is still in BLOCKING mode, and a
                 // blocking write into a full pipe whose reader never drains
-                // would wedge the collector itself (kaappi#2442 review).
-                // Flip it non-blocking first — a no-op for regular files,
-                // whose flush still completes in full. O_NONBLOCK lives on
-                // the SHARED open-file description, so the prior flags are
-                // restored before the close — dup/dup2 aliases (a child's
-                // stdio, an fd->port twin) must not come out of a
-                // collection permanently non-blocking — and if the flip
-                // itself fails, the blocking-capable drain is skipped
-                // entirely rather than risked (kaappi#2442 review).
-                var teardown_flags: ?c_int = null;
-                var drain_ok = true;
-                if (comptime !platform.is_windows and !platform.is_wasm) {
-                    if (!port.nonblocking and port.write_buf != null) {
-                        teardown_flags = platform.setFdNonblockingForTeardown(port.fd);
-                        if (teardown_flags == null) drain_ok = false;
-                    }
-                }
+                // would wedge the collector itself — while toggling
+                // O_NONBLOCK for the drain is no better: the flag lives on
+                // the SHARED open-file description, so every dup/dup2
+                // alias (a child's stdio, an fd->port twin) transiently —
+                // or, on a failed restore, permanently — turns non-blocking
+                // (kaappi#2442 review, twice). So no flag is ever touched;
+                // instead the drain is gated on seekability: a seekable fd
+                // is a regular file, whose blocking write completes without
+                // waiting on a peer (the flush-on-collect file ports have
+                // always had), while an unseekable one (pipe, socket, FIFO,
+                // tty — everything with a peer or flow control) simply
+                // drops its remainder, exactly the best-effort contract.
+                const drain_ok = if (comptime platform.is_windows or platform.is_wasm)
+                    true // the Windows arms below have their own non-blocking gates
+                else
+                    port.nonblocking or port.write_buf == null or
+                        platform.seek(port.fd, 0, platform.SEEK_CUR) >= 0;
                 if (drain_ok) {
                     if (port.write_buf) |wb| {
                         var start = port.write_buf_start;
@@ -472,7 +472,6 @@ pub fn freeObject(gc: *GC, obj: *Object) void {
                         }
                     }
                 }
-                if (teardown_flags) |fl| platform.restoreFdStatusFlags(port.fd, fl);
                 _ = platform.close(port.fd);
             }
             if (port.write_buf) |wb| {
--- a/src/gc_sweep.zig
+++ b/src/gc_sweep.zig
@@ -264,6 +264,7 @@ pub fn objectSize(obj: *Object) usize {
                 fiber.wind_stack.len * @sizeOf(types.WindRecord);
         },
         .channel => @sizeOf(types.Channel),
+        .process => @sizeOf(types.Process),
         .mutex => @sizeOf(types.Mutex),
         .condition_variable => @sizeOf(types.ConditionVariable),
         .srfi18_time => @sizeOf(types.Srfi18Time),
@@ -577,6 +578,30 @@ pub fn freeObject(gc: *GC, obj: *Object) void {
                 sc.release();
             }
             poisonAndDestroy(gc, types.Channel, ch);
+        },
+        .process => {
+            const proc = obj.as(types.Process);
+            // Zombie discipline (KEP-0022 Phase 1): drop the collected
+            // Process from the heap's unreaped list first -- after
+            // poisonAndDestroy the pointer is invalid -- then make one
+            // last-resort non-blocking reap attempt so an already-exited
+            // child of an abandoned Process does not linger as a zombie.
+            // Still-running children are simply no longer tracked; their
+            // zombies (if any) are reparented to init when this process
+            // exits. A BLOCKING wait is never acceptable inside a sweep.
+            for (gc.unreaped_processes.items, 0..) |p, i| {
+                if (p == proc) {
+                    _ = gc.unreaped_processes.swapRemove(i);
+                    break;
+                }
+            }
+            if (proc.status == null) {
+                if (comptime !platform.is_wasm and !platform.is_windows) {
+                    var st: c_int = 0;
+                    _ = platform.waitPid(proc.pid, &st, platform.WNOHANG);
+                }
+            }
+            poisonAndDestroy(gc, types.Process, proc);
         },
         .mutex => {
             poisonAndDestroy(gc, types.Mutex, obj.as(types.Mutex));
--- a/src/memory.zig
+++ b/src/memory.zig
@@ -414,6 +414,18 @@ pub const GC = struct {
     /// and the FREED_OWNER sentinel the whole #1687 machinery rests on is
     /// overwritten before the parent's next mark can read it.
     quarantine_heir: ?*GC = null,
+    /// KEP-0022 Phase 1 (zombie discipline): every Process this heap spawned
+    /// and has not reaped. The blocking spawn/wait primitives sweep it with
+    /// waitpid(WNOHANG); gc_sweep.freeObject removes a collected Process from
+    /// it (after a last-resort non-blocking reap), so entries are valid
+    /// pointers for exactly as long as the objects are. Deliberately NOT a
+    /// root list: an unreachable unreaped Process must be collectable, so
+    /// that its pipe ports close and a child blocked on a full pipe gets
+    /// EOF/EPIPE instead of hanging forever (Python's Popen semantics).
+    /// Residual window, documented and accepted: a Process collected while
+    /// its child still runs leaves a zombie if the child exits later and
+    /// this Kaappi process keeps running.
+    unreaped_processes: std.ArrayList(*types.Process) = .empty,
 
     pub const QuarantineEntry = struct {
         ptr: [*]u8,
@@ -499,6 +511,7 @@ pub const GC = struct {
         self.symbols.deinit();
         self.allocator.free(self.root_buffer);
         self.extra_roots.deinit(self.allocator);
+        self.unreaped_processes.deinit(self.allocator);
         self.remembered_set.deinit(self.allocator);
         self.mark_worklist.deinit(self.allocator);
         self.pending_ephemerons.deinit(self.allocator);
@@ -657,6 +670,7 @@ pub const GC = struct {
     pub const allocChannel = gc_alloc.allocChannel;
     pub const allocChannelBounded = gc_alloc.allocChannelBounded;
     pub const allocChannelStub = gc_alloc.allocChannelStub;
+    pub const allocProcess = gc_alloc.allocProcess;
     pub const allocMutex = gc_alloc.allocMutex;
     pub const allocConditionVariable = gc_alloc.allocConditionVariable;
     pub const allocSrfi18Time = gc_alloc.allocSrfi18Time;
--- a/src/platform.zig
+++ b/src/platform.zig
@@ -193,15 +193,27 @@ pub fn getFdFlags(fd: fd_t) c_int {
     return fcntlRaw(fd, F_GETFD, 0);
 }
 
-/// Set O_NONBLOCK on an fd about to be closed, so a final best-effort drain
-/// gets EAGAIN instead of blocking (the GC sweep's port flush — a collector
-/// must never wedge on a full pipe, kaappi#2442). POSIX only; best-effort.
-pub fn setFdNonblockingForTeardown(fd: fd_t) bool {
-    if (comptime is_wasm or is_windows) return false;
+/// Set O_NONBLOCK on an fd so a final best-effort drain gets EAGAIN instead
+/// of blocking (the GC sweep's port flush — a collector must never wedge on
+/// a full pipe, kaappi#2442). Returns the PRIOR status flags so the caller
+/// can restore them: O_NONBLOCK lives on the shared open-file description,
+/// and dup/dup2 aliases of this fd — a child's stdio, an `fd->port` twin —
+/// would otherwise be left permanently non-blocking (kaappi#2442 review).
+/// Null when the flags cannot be read or set; the caller must then skip its
+/// blocking-capable drain rather than fall through.
+pub fn setFdNonblockingForTeardown(fd: fd_t) ?c_int {
+    if (comptime is_wasm or is_windows) return null;
     const flags = std.c.fcntl(fd, std.posix.F.GETFL, @as(c_int, 0));
-    if (flags < 0) return false;
+    if (flags < 0) return null;
     const nonblock: c_int = @intCast(@as(u32, @bitCast(std.posix.O{ .NONBLOCK = true })));
-    return std.c.fcntl(fd, std.posix.F.SETFL, flags | nonblock) >= 0;
+    if (std.c.fcntl(fd, std.posix.F.SETFL, flags | nonblock) < 0) return null;
+    return flags;
+}
+
+/// Restore fcntl status flags saved by setFdNonblockingForTeardown.
+pub fn restoreFdStatusFlags(fd: fd_t, flags: c_int) void {
+    if (comptime is_wasm or is_windows) return;
+    _ = std.c.fcntl(fd, std.posix.F.SETFL, flags);
 }
 
 /// dup(2), except the copy is close-on-exec (fcntl F_DUPFD_CLOEXEC, the
--- a/src/platform.zig
+++ b/src/platform.zig
@@ -195,10 +195,24 @@ pub fn getFdFlags(fd: fd_t) c_int {
 
 /// dup(2), except the copy is close-on-exec (fcntl F_DUPFD_CLOEXEC, the
 /// POSIX 2008 replacement for dup-then-F_SETFD that cannot be interrupted
-/// between the two steps). Returns -1 on failure.
+/// between the two steps). Returns -1 on failure. The command value is
+/// per-OS (POSIX names it but does not fix it: 1030 on Linux, 67 on macOS,
+/// 17/12/10 on the BSDs), so take it from std.c's per-target table — a
+/// hardcoded number lands on a different fcntl command entirely on every
+/// other libc (kaappi#2442 review: 6 was macOS's F_SETOWN and Linux's
+/// F_SETLK, so the dup silently never happened).
 pub fn fcntlDupCloexec(fd: fd_t) fd_t {
-    const F_DUPFD_CLOEXEC: c_int = 6;
-    return fcntlRaw(fd, F_DUPFD_CLOEXEC, 0);
+    if (comptime is_wasm or is_windows) return -1;
+    // std.c's per-target F table is the authority where it declares the
+    // command; OpenBSD's entry omits it (an upstream gap — the OS has had
+    // it since 5.0, value 10 per its <fcntl.h>).
+    const cmd: c_int = comptime if (@hasDecl(std.c.F, "DUPFD_CLOEXEC"))
+        @intCast(std.c.F.DUPFD_CLOEXEC)
+    else if (builtin.os.tag == .openbsd)
+        10
+    else
+        @compileError("F_DUPFD_CLOEXEC value unknown for this target");
+    return fcntlRaw(fd, cmd, 0);
 }
 
 /// fcntl(2) MUST be declared variadic — its C prototype is
--- a/src/platform.zig
+++ b/src/platform.zig
@@ -193,29 +193,6 @@ pub fn getFdFlags(fd: fd_t) c_int {
     return fcntlRaw(fd, F_GETFD, 0);
 }
 
-/// Set O_NONBLOCK on an fd so a final best-effort drain gets EAGAIN instead
-/// of blocking (the GC sweep's port flush — a collector must never wedge on
-/// a full pipe, kaappi#2442). Returns the PRIOR status flags so the caller
-/// can restore them: O_NONBLOCK lives on the shared open-file description,
-/// and dup/dup2 aliases of this fd — a child's stdio, an `fd->port` twin —
-/// would otherwise be left permanently non-blocking (kaappi#2442 review).
-/// Null when the flags cannot be read or set; the caller must then skip its
-/// blocking-capable drain rather than fall through.
-pub fn setFdNonblockingForTeardown(fd: fd_t) ?c_int {
-    if (comptime is_wasm or is_windows) return null;
-    const flags = std.c.fcntl(fd, std.posix.F.GETFL, @as(c_int, 0));
-    if (flags < 0) return null;
-    const nonblock: c_int = @intCast(@as(u32, @bitCast(std.posix.O{ .NONBLOCK = true })));
-    if (std.c.fcntl(fd, std.posix.F.SETFL, flags | nonblock) < 0) return null;
-    return flags;
-}
-
-/// Restore fcntl status flags saved by setFdNonblockingForTeardown.
-pub fn restoreFdStatusFlags(fd: fd_t, flags: c_int) void {
-    if (comptime is_wasm or is_windows) return;
-    _ = std.c.fcntl(fd, std.posix.F.SETFL, flags);
-}
-
 /// dup(2), except the copy is close-on-exec (fcntl F_DUPFD_CLOEXEC, the
 /// POSIX 2008 replacement for dup-then-F_SETFD that cannot be interrupted
 /// between the two steps). Returns -1 on failure. The command value is
--- a/src/platform.zig
+++ b/src/platform.zig
@@ -193,6 +193,17 @@ pub fn getFdFlags(fd: fd_t) c_int {
     return fcntlRaw(fd, F_GETFD, 0);
 }
 
+/// Set O_NONBLOCK on an fd about to be closed, so a final best-effort drain
+/// gets EAGAIN instead of blocking (the GC sweep's port flush — a collector
+/// must never wedge on a full pipe, kaappi#2442). POSIX only; best-effort.
+pub fn setFdNonblockingForTeardown(fd: fd_t) bool {
+    if (comptime is_wasm or is_windows) return false;
+    const flags = std.c.fcntl(fd, std.posix.F.GETFL, @as(c_int, 0));
+    if (flags < 0) return false;
+    const nonblock: c_int = @intCast(@as(u32, @bitCast(std.posix.O{ .NONBLOCK = true })));
+    return std.c.fcntl(fd, std.posix.F.SETFL, flags | nonblock) >= 0;
+}
+
 /// dup(2), except the copy is close-on-exec (fcntl F_DUPFD_CLOEXEC, the
 /// POSIX 2008 replacement for dup-then-F_SETFD that cannot be interrupted
 /// between the two steps). Returns -1 on failure. The command value is
@@ -202,6 +213,16 @@ pub fn getFdFlags(fd: fd_t) c_int {
 /// other libc (kaappi#2442 review: 6 was macOS's F_SETOWN and Linux's
 /// F_SETLK, so the dup silently never happened).
 pub fn fcntlDupCloexec(fd: fd_t) fd_t {
+    return dupCloexecAtLeast(fd, 0);
+}
+
+/// F_DUPFD_CLOEXEC with a minimum descriptor number: the returned copy is
+/// the lowest free fd >= `min`. The subprocess spawner uses `min = 3` to
+/// lift pipe ends and redirect sources out of the stdio range, where
+/// posix_spawn file actions would otherwise clobber them (kaappi#2442
+/// review: a launcher that closed fd 0 makes pipe(2) hand back 0, and a
+/// `stdout:`/`stderr:` swap names slots the earlier dup2 already rewrote).
+pub fn dupCloexecAtLeast(fd: fd_t, min: fd_t) fd_t {
     if (comptime is_wasm or is_windows) return -1;
     // std.c's per-target F table is the authority where it declares the
     // command; OpenBSD's entry omits it (an upstream gap — the OS has had
@@ -212,7 +233,7 @@ pub fn fcntlDupCloexec(fd: fd_t) fd_t {
         10
     else
         @compileError("F_DUPFD_CLOEXEC value unknown for this target");
-    return fcntlRaw(fd, cmd, 0);
+    return fcntlRaw(fd, cmd, @intCast(min));
 }
 
 /// fcntl(2) MUST be declared variadic — its C prototype is
--- a/src/platform.zig
+++ b/src/platform.zig
@@ -158,7 +158,86 @@ pub fn isatty(fd: fd_t) bool {
 pub fn pipe(fds: *[2]fd_t) c_int {
     if (comptime is_windows) return win._pipe(fds, 65536, win.O_BINARY | win.O_NOINHERIT);
     if (comptime is_wasm) return -1;
-    return std.c.pipe(fds);
+    const rc = std.c.pipe(fds);
+    if (rc == 0) {
+        // CLOEXEC audit (KEP-0022 Phase 1): pipe(2) takes no flags, so the
+        // descriptors are made close-on-exec here. Every existing caller
+        // installs child ends with dup2 (which clears CLOEXEC on the child's
+        // slot) and keeps parent ends for its own lifetime, which exec never
+        // interrupts -- so this is invisible to them and stops both ends
+        // leaking into a spawned child.
+        _ = setFdCloexec(fds[0]);
+        _ = setFdCloexec(fds[1]);
+    }
+    return rc;
+}
+
+/// Set FD_CLOEXEC on `fd` (the fcntl(2) F_SETFD form; O_CLOEXEC at open is
+/// unavailable for fds other calls hand back -- kqueue(), pipe(2) on
+/// platforms without pipe2). Returns false on failure; callers treat that
+/// as best-effort.
+pub fn setFdCloexec(fd: fd_t) bool {
+    if (comptime is_wasm) return false;
+    const F_SETFD: c_int = 2;
+    const FD_CLOEXEC: usize = 1;
+    return fcntlRaw(fd, F_SETFD, FD_CLOEXEC) != -1;
+}
+
+/// fcntl(F_GETFD): the fd-flags word (bit 0 = FD_CLOEXEC), or -1 when `fd`
+/// is not open. The subprocess spawner's close-by-default scan uses it both
+/// as an is-open probe and to skip descriptors exec will close anyway
+/// (KEP-0022).
+pub fn getFdFlags(fd: fd_t) c_int {
+    if (comptime is_wasm) return -1;
+    const F_GETFD: c_int = 1;
+    return fcntlRaw(fd, F_GETFD, 0);
+}
+
+/// dup(2), except the copy is close-on-exec (fcntl F_DUPFD_CLOEXEC, the
+/// POSIX 2008 replacement for dup-then-F_SETFD that cannot be interrupted
+/// between the two steps). Returns -1 on failure.
+pub fn fcntlDupCloexec(fd: fd_t) fd_t {
+    const F_DUPFD_CLOEXEC: c_int = 6;
+    return fcntlRaw(fd, F_DUPFD_CLOEXEC, 0);
+}
+
+/// fcntl(2) MUST be declared variadic — its C prototype is
+/// `int fcntl(int, int, ...)`. A fixed-arity extern happens to link but is
+/// not the same call: on arm64 macOS the variadic calling convention differs
+/// and the third argument never reaches the kernel, so F_SETFD (and
+/// F_DUPFD_CLOEXEC's min-fd) silently no-ops while returning success — the
+/// exact bug that made every CLOEXEC set here a no-op on that platform
+/// (KEP-0022 audit). Every call site passes a typed argument, as Zig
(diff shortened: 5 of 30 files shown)


docs/dev/architecture.md
```diff
@@ -147,6 +148,7 @@ its struct lives outside `types.zig` entirely with no `types.Fiber` re-export.
 | `primitives_srfi237.zig` | SRFI-237: R6RS record procedural layer (`(srfi 237 primitives)`) |
 | `primitives_srfi254.zig` | SRFI-254: ephemeron/guardian constructors, predicates, accessors (GC half lives in `gc_collect.zig`) |
 | `primitives_fiber.zig` | `(kaappi fibers)`: spawn, yield, fiber-join, channels |
+| `primitives_process.zig` | `(kaappi process)` (KEP-0022): posix_spawnp-based subprocess support — spawn-process, pipe ports, blocking process-wait, process-kill, zombie sweep. POSIX-only (Windows is Phase 3, WASM registers nothing) |
```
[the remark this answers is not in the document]
baijum: Fixed in c679752f — both the overview table and the section heading now say 32.

---

src/gc_alloc.zig
```diff
@@ -1073,6 +1073,30 @@ pub fn allocChannel(self: *GC) !Value {
...
+    try self.maybeCollect();
+    const proc = try self.allocator.create(types.Process);
+    proc.* = .{
+        .header = .{ .tag = .process },
+        .pid = pid,
+        .pgid = pgid,
+    };
+    self.finishAlloc(&proc.header, @sizeOf(types.Process));
+    self.unreaped_processes.append(self.allocator, proc) catch
+        @panic("GC: process tracking list OOM");
```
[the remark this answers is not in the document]
baijum: Fixed in c679752f as proposed: the tracking slot is reserved with ensureUnusedCapacity between maybeCollect and the create, and the append is infallible — a list-growth failure now raises error.OutOfMemory through the existing error union instead of aborting.

---

src/primitives_io.zig
```diff
@@ -355,7 +355,37 @@ fn unparkCurrentFiber(port: *types.Port) void {
...
+/// KEP-0022: flush a caller port's buffered output before its fd is handed
+/// to a spawned child as a redirection target, so bytes the parent wrote
+/// before the spawn reach the file ahead of anything the child writes.
+/// Raising flavor — a redirect onto a port whose backing fd is already dead
+/// should fail the spawn loudly, not silently drop the parent's bytes.
+pub fn drainPortWriteBufferForSpawn(port: *types.Port) PrimitiveError!void {
+    if (port.write_buf_len > port.write_buf_start) return drainWriteBufferInner(port, true);
+}
+
+fn drainWriteBufferInner(port: *types.Port, comptime raise_write_errors: bool) PrimitiveError!void {
```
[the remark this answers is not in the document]
baijum: Declining this one: the repo's file-size policy (CLAUDE.md, File size policy) explicitly exempts primitives files — "Do NOT split flat lists of independent functions (e.g. primitives files) — size from breadth is fine; size from tangled coupling is not." primitives_io.zig was 2391 lines before this PR and gains 59 here, all of it cohesive port-layer plumbing (the drain split and one shared constructor), so a split would be pre-existing-debt work out of scope for this feature PR.

---

src/tests_process.zig
```diff
@@ -0,0 +1,549 @@
...
+    if (comptime !is_posix) return error.SkipZigTest;
+    var ctx: th.TestContext = undefined;
+    try ctx.init();
+    defer ctx.deinit();
+    const vm = ctx.vm;
+    _ = try vm.eval("(import (scheme file))");
+
+    // 'null: the child's stdout goes to the null device — the pipe
+    // accessor still exists (stdout was 'pipe), and reads EOF.
+    // (Two 'stdout: options: last wins — the null one.)
```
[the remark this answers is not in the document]
baijum: Fixed in c679752f with your proposed wording — the comment now matches the assertion (last 'stdout: wins, no pipe, accessor is #f).

---

src/platform.zig
```diff
@@ -158,7 +158,86 @@ pub fn isatty(fd: fd_t) bool {
...
+    if (comptime is_wasm) return -1;
+    const F_GETFD: c_int = 1;
+    return fcntlRaw(fd, F_GETFD, 0);
+}
+
+/// dup(2), except the copy is close-on-exec (fcntl F_DUPFD_CLOEXEC, the
+/// POSIX 2008 replacement for dup-then-F_SETFD that cannot be interrupted
+/// between the two steps). Returns -1 on failure.
+pub fn fcntlDupCloexec(fd: fd_t) fd_t {
+    const F_DUPFD_CLOEXEC: c_int = 6;
```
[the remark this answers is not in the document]
baijum: Confirmed and fixed in c679752f — and it was worse than Linux-only: 6 is not F_DUPFD_CLOEXEC on any supported target (macOS: F_SETOWN, which even *returns 0*, silently handing back a non-descriptor; Linux: F_SETLK), so the #2152 diagnostic preservation never actually happened anywhere. The command now comes from std.c's per-target F table; OpenBSD's std.c entry omits the constant (an upstream gap — the OS has had it since 5.0), so its value 10 is supplied explicitly behind a @hasDecl check, with a @compileError backstop for any future target neither source covers. Verified compiling on linux-gnu/musl, macOS, all three BSDs, s390x, windows, and wasi.

---

src/primitives_process.zig
```diff
@@ -0,0 +1,1066 @@
...
+            // rlim_t is signed on some libcs (FreeBSD's i64); a negative
+            // cur (RLIM_INFINITY representations aside) means "no info".
+            const cur = std.math.cast(u64, rl.cur) orelse break :blk cap;
+            break :blk @min(cur, cap);
+        };
+        var fd: platform.fd_t = 3;
+        while (fd < limit) : (fd += 1) {
+            fds.append(gc.allocator, fd) catch return PrimitiveError.OutOfMemory;
+        }
+    }
```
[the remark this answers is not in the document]
baijum: Fixed in c679752f — the dense BSD fallback range is now probed inline (shared closeFdIfInheritable helper), so no per-spawn allocation and no artificial OutOfMemory mode on that path. The ArrayList remains only for the genuinely sparse /proc/self/fd listing, where collecting before probing is load-bearing: the directory's own fd must be closed before the F_GETFD re-probe so it self-excludes.

---

src/vm.zig
```diff
@@ -764,6 +765,28 @@ pub const VM = struct {
...
+    /// signals).
+    fn ignoreSigpipe() void {
+        const builtin = @import("builtin");
+        if (comptime builtin.os.tag == .windows or builtin.os.tag == .wasi) return;
+        var act: std.posix.Sigaction = .{
+            .handler = .{ .handler = std.posix.SIG.IGN },
+            .mask = std.posix.sigemptyset(),
+            .flags = 0,
+        };
+        std.posix.sigaction(std.posix.SIG.PIPE, &act, null);
```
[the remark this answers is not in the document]
baijum: Fixed in c679752f as proposed: the disposition is queried first and only SIG_DFL is upgraded to SIG_IGN, so an embedder's own SIGPIPE handler survives every VM.init — their handler still runs and the write still returns EPIPE, which is all the KEP-0022 contract needs.

---

src/primitives_process.zig
```diff
@@ -0,0 +1,1087 @@
...
+    const O_RDONLY: c_int = 0;
+    const O_WRONLY: c_int = 1;
+    const null_z: [*:0]const u8 = "/dev/null";
+
+    // stdin (slot 0)
+    switch (cfg.stdin) {
+        .inherit => {},
+        .pipe => {
+            if (spawn_c.posix_spawn_file_actions_adddup2(actions, stdin_pipe[0], 0) != 0 or
+                spawn_c.posix_spawn_file_actions_addclose(actions, stdin_pipe[0]) != 0)
```
baijum: `pipe()` may return 0, 1, or 2 when the launcher/embedder closed a standard descriptor. If `stdin_pipe[0] == 0`, this queues `dup2(0, 0)` followed by `close(0)`, so the child loses the stream that was just installed. I reproduced this on macOS and the FreeBSD VM by launching Kaappi with fd 0 closed; spawned `cat` reports `stdin: Bad file descriptor`, exits 1, and the parent gets EPIPE. Parent pipe ends in 0..2 also conflict with the port close discipline. Please normalize every pipe end above 2 before building actions and test with each standard descriptor initially closed.

src/primitives_process.zig
```diff
@@ -0,0 +1,1087 @@
...
+            const X_OK: c_int = 1;
+            if (spawn_c.access(prog, X_OK) != 0)
+                return raiseProcessError(gc, "cannot spawn process", cfg.argv[0], std.c._errno().*);
+        }
+    }
+
+    // -- spawn
+    var pid: c_int = -1;
+    const envp: [*:null]const ?[*:0]const u8 = env_block orelse @ptrCast(std.c.environ);
+    const spawn_rc = spawn_c.posix_spawnp(&pid, argv[0].?, actions, attr, argv, envp);
```
baijum: `posix_spawnp` is a potentially blocking native operation (the vfork-style implementations wait through exec and filesystem lookup), but a child-thread VM remains `.running` here. If it stalls, a concurrent parent collection spins forever in `markLiveChildRoots` waiting for a safepoint this VM cannot reach. Please use the same `setCollectionInNative`/`setCollectionRunning` protocol as `process-wait` around the spawn syscall and NetBSD barrier, restoring `.running` before any allocation or VM mutation. A blocking test shim from an SRFI-18 child can make the regression deterministic.

src/primitives_process.zig
```diff
@@ -0,0 +1,1087 @@
...
+    // pointer-flavor platforms).
+    defer _ = spawn_c.posix_spawn_file_actions_destroy(actions);
+
+    const O_RDONLY: c_int = 0;
+    const O_WRONLY: c_int = 1;
+    const null_z: [*:0]const u8 = "/dev/null";
+
+    // stdin (slot 0)
+    switch (cfg.stdin) {
+        .inherit => {},
```
baijum: `POSIX_SPAWN_CLOEXEC_DEFAULT` also suppresses descriptors 0/1/2 unless a file action explicitly manipulates them, so this no-op breaks the default `inherit` behavior on Darwin. Reproduced on the PR head: spawning `sh -c "echo CHILD-OUT; echo CHILD-ERR >&2"` produced neither line and exited 1. The Darwin man page specifically requires `posix_spawn_file_actions_addinherit_np` for pre-existing descriptors under this flag. Please add inherit actions for each open inherited stdio slot and a regression covering stdin, stdout, and stderr.

src/primitives_process.zig
```diff
@@ -0,0 +1,1087 @@
...
+        return raiseProcessError(gc, "cannot create stdin pipe", types.FALSE, std.c._errno().*);
+    if (cfg.stdout == .pipe and pipeWithFdRetry(gc, &stdout_pipe) != 0)
+        return raiseProcessError(gc, "cannot create stdout pipe", types.FALSE, std.c._errno().*);
+    if (cfg.stderr == .pipe and pipeWithFdRetry(gc, &stderr_pipe) != 0)
+        return raiseProcessError(gc, "cannot create stderr pipe", types.FALSE, std.c._errno().*);
+
+    // -- file actions + attrs
+    var actions_storage = spawn_c.zeroStorage();
+    const actions: *spawn_c.FileActionsPtr = @ptrCast(&actions_storage);
+    _ = spawn_c.posix_spawn_file_actions_init(actions);
```
baijum: Both this initializer and `posix_spawnattr_init` below return error numbers, but their results are ignored and the destroy defers are armed unconditionally. On pointer-backed implementations these initializers may fail with ENOMEM, leaving an uninitialized/null object that subsequent add/set/destroy calls use. Please capture each init return code, report it directly, and arm destruction only after successful initialization; the pre-spawn pipe cleanup should remain active on either failure. An injected-init-failure test would pin the lifecycle.

src/primitives_process.zig
```diff
@@ -0,0 +1,1087 @@
...
+        // — O_NONBLOCK included. A port the fiber scheduler already flipped
+        // non-blocking would hand the child non-blocking stdio, and the
+        // parent's next I/O would re-flip it even if cleared here; reject
+        // rather than corrupt (kaappi#2414 review).
+        if (port.nonblocking)
+            return primitives.argError(proc, "redirection port is in non-blocking use by the fiber scheduler; pass a port not yet driven by fibers", .{});
+        // Buffered parent output must land in the file before the child's
+        // own writes; the drain can park a fiber, and a re-executed spawn
+        // re-parses the options idempotently.
+        if (port.is_output) try primitives_io.drainPortWriteBufferForSpawn(port);
```
baijum: Output buffering is reconciled here, but input read-ahead is not. Passing an fd-backed input port after any read gives the child the kernel offset rather than the port logical offset, silently dropping buffered bytes. Reproduced by writing `AB` plus 5000 bytes, reading `A` in Kaappi, then passing the port to `cat`: the child begins at `x`, not `B`. Please seek back/clear `peek_byte`, `peek_extra`, and `read_buf` where possible, or reject a port with pending input state when it cannot be synchronized, with a read-then-spawn regression.

src/primitives_process.zig
```diff
@@ -0,0 +1,1087 @@
...
+    const envp: [*:null]const ?[*:0]const u8 = env_block orelse @ptrCast(std.c.environ);
+    const spawn_rc = spawn_c.posix_spawnp(&pid, argv[0].?, actions, attr, argv, envp);
+    if (spawn_rc != 0) {
+        // posix_spawn returns the errno itself, not -1. ENOENT is the common
+        // case (program not on PATH); the errno rides the condition so
+        // posix-error-name reports it precisely. The `spawned` defer closes
+        // the pipes; the destroy defers release actions/attr.
+        return raiseProcessError(gc, "cannot spawn process", cfg.argv[0], spawn_rc);
+    }
+    spawned = true;
```
baijum: This disarms all pipe cleanup before the fallible `allocProcess` and three `makeFdPort` calls below. An OOM at any of those boundaries returns with a live, untracked child and leaked parent pipe ends; a child reading piped stdin can then remain blocked indefinitely. Keep cleanup armed until every fd has transferred ownership, and ensure a post-spawn failure terminates/reaps or otherwise tracks the child. Please add OOM injection at `allocProcess` and each port allocation and assert no fd/child remains.

tests/scheme/run-all.sh
```diff
@@ -657,6 +660,7 @@ if command -v zig >/dev/null 2>&1; then
     fi
 fi
 run_suite "FFI tests" tests/scheme/ffi/*.scm
+run_suite "Process tests" tests/scheme/process/*.scm
```
baijum, line 663: This runs the new directory, but `SCM_SUITE_DIRS` above and the corresponding inventory in `tools/run-gc-stress-suite.sh` still omit `process`. The process files therefore bypass the reachability/verdict checks, and `gc-stress-scheme` does not execute them at all despite the PR reporting that gate green. Please add the directory to both inventories; the stress runner also needs an appropriate `KAAPPI` value so `process-child-script.scm` does not silently skip its round trip.

src/primitives_process.zig
```diff
@@ -0,0 +1,1087 @@
...
+            it.close();
+            for (fds.items) |fd| try closeFdIfInheritable(gc, actions, fd);
+            return;
+        }
+    }
+    // BSD fallback (and Linux without /proc): probe the descriptor range
+    // inline — materializing up to 65k entries per spawn would cost real
+    // allocation on the hot path for nothing, since the range is dense by
+    // construction (kaappi#2442 review).
+    const cap: u64 = 65536;
```
baijum: This hard cap breaks the stated close-by-default guarantee whenever `RLIMIT_NOFILE` exceeds 65536. On the FreeBSD reference VM (`ulimit -n` = 117153), I launched Kaappi with inheritable fd 70000 and a spawned child reported it still `OPEN`. Such descriptors can retain sockets, locks, or pipe writers. Please scan the full finite soft limit or use an OS close-range/closefrom facility without silently exempting valid high descriptors, and add a high-fd regression on FreeBSD.

src/primitives_process.zig
```diff
@@ -0,0 +1,1087 @@
...
+            if (spawn_c.posix_spawn_file_actions_adddup2(actions, stdout_pipe[1], 1) != 0 or
+                spawn_c.posix_spawn_file_actions_addclose(actions, stdout_pipe[1]) != 0)
+                return spawnFileActionError(gc);
+        },
+        .null_sink => {
+            if (spawn_c.posix_spawn_file_actions_addopen(actions, 1, null_z, O_WRONLY, 0o666) != 0)
+                return spawnFileActionError(gc);
+        },
+        .fd => |fd| {
+            if (spawn_c.posix_spawn_file_actions_adddup2(actions, fd, 1) != 0)
```
baijum: The actions are evaluated sequentially in the child fd namespace, so later fd-port redirects may no longer refer to the parent descriptor they named. For example, `stdout: (current-error-port)` followed by `stderr: (current-output-port)` queues `dup2(2,1); dup2(1,2)` and sends both streams to the original stderr instead of swapping them. Reproduced on the PR head: stdout was empty and stderr contained both `OUT` and `ERR`. Please stage stdio sources in temporary descriptors or otherwise order the action graph safely; test swaps and cross-references alongside the intentional stderr-to-stdout merge.

src/memory.zig
```diff
@@ -414,6 +414,18 @@ pub const GC = struct {
     /// and the FREED_OWNER sentinel the whole #1687 machinery rests on is
     /// overwritten before the parent's next mark can read it.
     quarantine_heir: ?*GC = null,
+    /// KEP-0022 Phase 1 (zombie discipline): every Process this heap spawned
+    /// and has not reaped. The blocking spawn/wait primitives sweep it with
+    /// waitpid(WNOHANG); gc_sweep.freeObject removes a collected Process from
+    /// it (after a last-resort non-blocking reap), so entries are valid
+    /// pointers for exactly as long as the objects are. Deliberately NOT a
+    /// root list: an unreachable unreaped Process must be collectable, so
+    /// that its pipe ports close and a child blocked on a full pipe gets
```
baijum, line 423: The current port finalizer does not reliably close these pipes: `gc_sweep.zig:430-449` first flushes the whole pending write buffer with blocking `platform.write`, and a fresh process pipe stays blocking until scheduler I/O touches it. Reproduced by buffering 1 MiB to the stdin of `sleep 5`, dropping the process/port, and forcing collection; the collector blocked past a 3-second bound. A sweep must not perform a potentially blocking flush. Please discard pending output and close immediately during finalization (explicit flush/close paths can retain their normal semantics), with a bounded regression.

src/primitives_process.zig
baijum: Fixed in 2f05452b: both ends of every created pipe are lifted above the stdio range with F_DUPFD_CLOEXEC (`normalizePipeAboveStdio`) before any file action names them, so a launcher-closed standard descriptor can no longer make a pipe end collide with the slot it is being installed into. Regression: `tests/scheme/errors/process-stdio-2442.sh` runs a cat round trip with the parent's fd 0 closed (`0<&-`) — the harness-level arrangement a .scm suite cannot make.

---

src/primitives_process.zig
baijum: Confirmed and fixed in 2f05452b: each `'inherit` slot whose parent descriptor is open now gets `posix_spawn_file_actions_addinherit_np`, exactly per the Darwin man page; a slot the parent itself has closed is skipped so the child correctly sees it closed. Regression: the same shell suite asserts CHILD-OUT/CHILD-ERR actually arrive on the parent's streams. A fun knock-on: fixing this broke the fd-hygiene scan test — with stdin now genuinely inherited, the probe shell's own `<&N` redirection parks the live stdin at bash's fd-10 save slot, and the scan detected the shell's own bookkeeping. The probes now close stdin up front (`exec 0<&-`).

---

src/primitives_process.zig
baijum: Fixed in 2f05452b: redirect sources in 0..2 are staged into >= 3 CLOEXEC duplicates before the actions are built, so the sequential child-side evaluation can no longer read a slot an earlier dup2 rewrote. The staged copies are parent-side, released on every exit path; exec drops the child's view of them. Regression: the shell suite runs your exact repro — `stdout: (current-error-port)`, `stderr: (current-output-port)` — and asserts OUT and ERR truly swap (and only swap).

---

src/primitives_process.zig
baijum: Fixed in 2f05452b: both `posix_spawn_file_actions_init` and `posix_spawnattr_init` return codes are checked and raised as the errno-carrying setup error, and each destroy defer is armed only after its init succeeded. The pre-spawn pipe cleanup is the function-wide defer, so it stays active across either failure. (An injected-init-failure test would need a libc shim we don't have a seam for — the lifecycle is pinned by inspection plus the OOM sweep below, which drives every *reachable* failure path.)

---

src/primitives_process.zig
baijum: Fixed in 2f05452b: an input redirect port's pending read-ahead (peek bytes + read_buf) is reconciled before the fd is handed over — `rewindPortReadAheadForSpawn` seeks the fd back by the pending count and discards the buffers; an unseekable descriptor with pending bytes is rejected with a message carrying the byte count. Regression in `tests_process.zig`: write AB+5000 filler, read one char (pulling a whole chunk into read-ahead), pass the port as `stdin:` to `dd bs=1 count=1` — the child must see B.

---

src/primitives_process.zig
baijum: Fixed in 2f05452b: the disarm-everything flag is gone. The cleanup defer is function-wide and always armed; ownership leaves it slot by slot (child ends -1 after their explicit close, each parent end -1 only once its Port exists). The one path where the child exists but nothing can ever track it — `allocProcess` itself failing — kills and reaps the fresh child before raising OOM. Regression: a 400-point `gc.oom_countdown` sweep over a three-pipe spawn asserts the process's open-descriptor count is identical before and after all failure injections.

---

tests/scheme/run-all.sh
baijum: Fixed in 2f05452b: `process` is in `SCM_SUITE_DIRS` in both `tests/scheme/run-all.sh` and `tools/run-gc-stress-suite.sh`, and the stress runner now `export`s KAAPPI so `process-child-script.scm` actually performs its round trip there instead of silently skipping. Verified locally: run-all's reachability/verdict checks pass with the process files enrolled (2131/0), and the stress-filtered process suite is green under `-Dgc-stress=true`.

---

src/primitives_process.zig
baijum: Fixed in 2f05452b: no more cap. FreeBSD — the platform where six-figure soft limits are the default — now uses `posix_spawn_file_actions_addclosefrom_np(actions, 3)`: one kernel-side action after the slot dup2s, unbounded by construction. The NetBSD/OpenBSD probe scan runs to the full soft limit (their defaults are modest), and Linux's /proc enumeration was never capped. Regression: a Zig test parks an inheritable fd at 70000 (skipped where the limit forbids it) and requires the child to find it closed, with an fd-2 control proving the probe can detect open descriptors.

---

src/primitives_process.zig
baijum: Fixed in 2f05452b: the `posix_spawnp` call and the NetBSD NOTE_EXEC barrier now run under the same `setCollectionInNative`/`setCollectionRunning` protocol as the blocking `process-wait`, with `.running` restored before anything below the block allocates. (I didn't add the blocking-shim regression — there's no seam to make posix_spawnp stall deterministically without a libc interposer; the pattern is the one #1933 pinned for thread-join and process-wait already exercises.)

---

src/memory.zig
baijum: Fixed in 2f05452b: before the sweep's best-effort drain, the fd is flipped O_NONBLOCK (`platform.setFdNonblockingForTeardown`) when the scheduler never did — so the drain's existing drop-on-EAGAIN promise actually holds. Regular files ignore O_NONBLOCK, so the long-standing flush-on-collect behavior for file ports is unchanged, and the fd is closed immediately after so the flip has no other observers. Regression: fill a `sleep 3` child's stdin pipe near capacity (flushed), buffer more below the high-water mark, abandon everything, and require `collectFull` to return well under the child's lifetime.

---

src/primitives_process.zig
```diff
@@ -0,0 +1,1087 @@
...
+}
+
+fn closeFdIfInheritable(gc: *GC, actions: *spawn_c.FileActionsPtr, fd: platform.fd_t) PrimitiveError!void {
+    const FD_CLOEXEC: c_int = 1;
+    if (fd < 3) return;
+    const flags = platform.getFdFlags(fd);
+    if (flags < 0) return; // closed (or the enumeration dir's own fd)
+    if ((flags & FD_CLOEXEC) != 0) return; // exec closes it anyway
+    if (spawn_c.posix_spawn_file_actions_addclose(actions, fd) != 0) {
+        _ = try spawnFileActionError(gc);
```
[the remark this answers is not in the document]
baijum: Fixed in 2f05452b: every posix_spawn file-action and attr call site captures the RETURNED error number and passes it to the (renamed) `spawnSetupError`, which forwards it as the condition's errno — ambient errno is no longer consulted for these.

---

src/primitives_process.zig
```diff
@@ -0,0 +1,1210 @@
...
+            if (std.c._errno().* != @intFromEnum(std.c.E.INTR)) break;
+        }
+        return PrimitiveError.OutOfMemory;
+    };
+    var proc_val = types.makePointer(&proc.header);
+    gc.pushRoot(&proc_val);
+    defer gc.popRoot();
+
+    if (stdin_pipe[1] >= 0) {
+        const port_val = try primitives_io.makeFdPort(gc, stdin_pipe[1], false, true, "process-stdin");
```
baijum, line 670: The new kill/reap cleanup covers only `allocProcess` failure. If this or either later `makeFdPort` fails, the function returns without killing the child; the Process is never returned, and the next GC collects it, removes it from `unreaped_processes`, and leaves a still-running child permanently untracked. The OOM test uses `true` and checks only fd counts, so the child exits before this becomes visible. Keep post-spawn kill/reap cleanup armed until all port construction succeeds, and test with a long-lived child while asserting its pid is gone and no registry entry is lost.

src/primitives_process.zig
```diff
@@ -0,0 +1,1210 @@
...
+        // — O_NONBLOCK included. A port the fiber scheduler already flipped
+        // non-blocking would hand the child non-blocking stdio, and the
+        // parent's next I/O would re-flip it even if cleared here; reject
+        // rather than corrupt (kaappi#2414 review).
+        if (port.nonblocking)
+            return primitives.argError(proc, "redirection port is in non-blocking use by the fiber scheduler; pass a port not yet driven by fibers", .{});
+        // Buffered parent output must land in the file before the child's
+        // own writes; the drain can park a fiber, and a re-executed spawn
+        // re-parses the options idempotently.
+        if (port.is_output) try primitives_io.drainPortWriteBufferForSpawn(port);
```
baijum: For a bidirectional fd port this order flushes pending output at the raw read-ahead offset, then rewinds afterward. The child gets the logical position, but the parent buffer has already been written to the wrong place. Reproduced with an O_RDWR file containing `ABCDEF`: read `A`, buffer `X`, then use the port as child stdin; the file becomes `ABCDEFX` rather than `AXCDEF`. Reconcile/discard input read-ahead before draining output, and add a bidirectional-port regression.

src/gc_sweep.zig
```diff
@@ -426,6 +427,19 @@ pub fn freeObject(gc: *GC, obj: *Object) void {
...
+                // EAGAIN: a port the scheduler never touched (a fresh
+                // process pipe, KEP-0022) is still in blocking mode, and a
+                // blocking write into a full pipe whose reader never drains
+                // would wedge the collector itself (kaappi#2442 review).
+                // Flip it non-blocking first — a no-op for regular files,
+                // whose flush still completes in full, and irrelevant to
+                // the fd's future since it is closed right below.
+                if (comptime !platform.is_windows and !platform.is_wasm) {
+                    if (!port.nonblocking and port.write_buf != null)
+                        _ = platform.setFdNonblockingForTeardown(port.fd);
```
baijum: `F_SETFL(O_NONBLOCK)` changes the shared open-file description, not just `port.fd`, and closing this descriptor does not undo it for `dup` aliases or child descriptors created with `dup2`. I reproduced this through `fd->port`: after writing/flushing once, dropping the port, and forcing GC, `F_GETFL` on a retained duplicate changed from `1` to `65541`; future blocking users of that alias now receive `EAGAIN`. Please avoid permanently mutating the open-file description during finalization (or preserve/restore the original flags before close), and add a duplicate-alias regression. If the transition itself fails, skip the drain rather than falling through to the original blocking write.

src/tests_process.zig
```diff
@@ -0,0 +1,788 @@
...
+
+    // Control: the probe must DETECT an open fd, or "closed" verdicts below
+    // are vacuous (a shell that rejects multi-digit redirection fds would
+    // report every fd closed). fd 2 is inherited and certainly open.
+    try expectTrue(vm,
+        \\(let ((p (spawn-process '("/bin/sh" "-c" "if (eval \": <&2\") 2>/dev/null; then exit 0; else exit 1; fi"))))
+        \\  (= (process-wait p) 0))
+    );
+    try expectTrue(vm,
+        \\(let ((p (spawn-process '("/bin/sh" "-c" "if (eval \": <&70000\") 2>/dev/null; then exit 1; else exit 0; fi"))))
```
baijum: This probe is vacuous on the FreeBSD reference VM: `/bin/sh` rejects `<&70000` as an invalid descriptor operand and takes the `else` branch even when fd 70000 is demonstrably open. I verified that by launching the shell with an inheritable fd 70000; this exact expression printed `CLOSED`. The fd-2 control only proves single-digit parsing, not that this multi-digit form can detect openness, so removing `addclosefrom_np` still leaves the test green. Use a child helper that calls `fcntl(70000, F_GETFD)` directly and mutation-check the closefrom path.

src/primitives_process.zig
```diff
@@ -0,0 +1,1210 @@
...
+    // there — `stdout: (current-error-port) stderr: (current-output-port)`
+    // queued dup2(2,1); dup2(1,2) and sent BOTH streams to the original
+    // stderr instead of swapping (kaappi#2442 review). The staged copies
+    // are parent-side CLOEXEC dups: actions reference them, exec drops
+    // them in the child, and the defer above releases the parent's copies.
+    var redirs = [3]Redir{ cfg.stdin, cfg.stdout, cfg.stderr };
+    for (&redirs, 0..) |*r, slot| {
+        if (r.* != .fd) continue;
+        if (r.*.fd > 2) continue;
+        const staged = platform.dupCloexecAtLeast(r.*.fd, 3);
```
baijum, line 438: Staging is unnecessary when the redirect source already equals its destination, such as `stdout: (current-output-port)`: the eventual `dup2(1,1)` is collision-safe and itself names the descriptor for Darwin CLOEXEC_DEFAULT. The unconditional duplicate makes this explicit equivalent of inheritance fail under `EMFILE`. Reproduced after filling a lowered fd limit: default `spawn-process` returned 0, while adding `stdout: (current-output-port)` raised. Skip staging when `r.fd == slot`, with an exhaustion regression.

src/primitives_process.zig
```diff
@@ -0,0 +1,1210 @@
...
+    }
+
+    // -- file actions + attrs. Both initializers can fail (ENOMEM on the
+    // pointer-backed libcs), and their destroy must run only on a
+    // successfully initialized object (kaappi#2442 review).
+    var actions_storage = spawn_c.zeroStorage();
+    const actions: *spawn_c.FileActionsPtr = @ptrCast(&actions_storage);
+    {
+        const rc = spawn_c.posix_spawn_file_actions_init(actions);
+        if (rc != 0) return spawnSetupError(gc, rc);
```
baijum, line 457: FreeBSD is an exception to the “returned errno” rule for this initializer: its libc implementation returns `-1` when the internal `malloc` fails, leaving `ENOMEM` in `errno` (the other file-action calls and `posix_spawnattr_init` return errno directly). Passing `-1` here produces an invalid `posix_errno`/`EUNKNOWN` diagnostic under memory pressure. Normalize a negative init result from `errno` on FreeBSD while continuing to use direct return codes elsewhere; an interposed allocation-failure test would pin the distinction.

src/primitives_process.zig
baijum: Fixed in 90c8a2f5: an errdefer now covers the whole port-construction stretch — any failure after the spawn kills and reaps the fresh child (shared killAndReapFresh), records the SIGKILL status on the Process so freeObject can never re-reap a reused pid, and drops the registry entry. Regression: the OOM sweep's child is now a self-killed `sleep 30` instead of `true` (an instantly-exiting child made an abandoned-live-child invisible, as you noted), and the sweep's epilogue asserts the unreaped registry is empty AND `waitpid(-1, WNOHANG)` finds no zombie.

---

src/primitives_process.zig
baijum: Fixed in 90c8a2f5: parseRedir reconciles the input read-ahead BEFORE draining buffered output, so the drain lands at the logical position. Regression: your exact repro — O_RDWR ABCDEF via fd->port, read A, buffer X, hand to `dd bs=1 count=1` — asserts the file becomes AXCDEF and the child reads C. Mutation-verified: swapping the two steps back reproduces your ABCDEFX (the test fails with a 7-byte file).

---

src/gc_sweep.zig
baijum: Fixed in 90c8a2f5: setFdNonblockingForTeardown now returns the prior status flags, the sweep restores them before the close, and a failed flip skips the blocking-capable drain entirely instead of falling through. Regression: a dup alias of a pipe write end is checked for O_NONBLOCK after the wrapped original is collected with pending buffered output — the alias must come out of the collection still blocking.

---

src/tests_process.zig
baijum: Confirmed and fixed in 90c8a2f5 — and verified on the FreeBSD reference VM both ways. The probe now asks the OS's own fd inventory per platform (procstat on FreeBSD, /proc/self/fd on Linux, /dev/fd on macOS; the shell-dup form is gone), each validated by an fd-2 control that must report OPEN first. Mutation check on the FreeBSD box: with addclosefrom_np disabled AND the old 65536 cap restored, exactly this test fails (1865/1873 + 1 failed); with the real code, 1866/0.

---

src/primitives_process.zig
baijum: Fixed in 90c8a2f5: staging is skipped when the source already equals its destination slot — dup2(slot, slot) is collision-free and still names the fd for Darwin's CLOEXEC_DEFAULT. Regression: the parent's fd table is filled to EMFILE (soft limit first lowered to a still-comfortable 256 — lowering to zero-free starves the CHILD's exec instead: dyld aborts, measured as (signaled . 6)), a control asserts a staging dup is impossible at that moment, and `stdout: (current-output-port)` must still spawn and wait 0. Mutation-verified: without the skip, the test fails with the staging EMFILE raise.

---

src/primitives_process.zig
baijum: Fixed in 90c8a2f5: both initializers' results are normalized — a negative return falls back to errno before being raised — so FreeBSD's -1/ENOMEM convention reports ENOMEM rather than an invalid error number, while the errno-returning libcs are untouched (they never return negatives).

---

src/primitives_process.zig
```diff
@@ -0,0 +1,1210 @@
...
+    // targets that commonly raise RLIMIT_NOFILE into six figures (FreeBSD,
+    // Linux) never reach this loop — they use addclosefrom_np and /proc
+    // respectively.
+    const fallback: u64 = 65536;
+    const limit: u64 = blk: {
+        const rl = std.posix.getrlimit(.NOFILE) catch break :blk fallback;
+        // rlim_t is signed on some libcs (FreeBSD's i64); a negative
+        // cur (RLIM_INFINITY representations aside) means "no info".
+        break :blk std.math.cast(u64, rl.cur) orelse fallback;
+    };
```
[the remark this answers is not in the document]
baijum: Fixed in 90c8a2f5: the scan limit is clamped to maxInt(fd_t) — an RLIM_INFINITY-shaped soft limit can no longer overflow the counter, and the clamp exempts nothing (no descriptor can exceed it). On the scan-visible platforms the loop stays: closefrom(2) exists as a *syscall* on the BSDs but not as a posix_spawn file action there, and the child-side close must happen between the dup2 actions and the exec, which only a file action can do. FreeBSD — the platform with six-figure default limits — bypasses the scan entirely via addclosefrom_np.

---

src/tests_process.zig
```diff
@@ -0,0 +1,767 @@
...
+    // more below the high-water mark (unflushed), abandon everything, and
+    // require the full collection to come back promptly.
+    try expectTrue(vm,
+        \\(let* ((p (spawn-process '("/bin/sleep" "3") 'stdin: 'pipe))
+        \\       (in (process-stdin p)))
+        \\  (write-string (make-string 60000 #\x) in)
+        \\  (flush-output-port in)
+        \\  (write-string (make-string 8100 #\y) in)
+        \\  #t)
+    );
```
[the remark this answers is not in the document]
baijum: Fixed in 90c8a2f5: the flushed portion is now 12000 bytes (fits even OpenBSD's fixed 16 KiB pipe, so the flush itself can never block) and the unflushed tail is 60000 bytes — larger than the remaining capacity on both 16 KiB and 64 KiB systems, so the sweep's drain meets a full pipe partway through on every platform.

---

tests/scheme/errors/process-stdio-2442.sh
```diff
@@ -0,0 +1,92 @@
...
+
+ok()   { echo "PASS: $1"; PASS=$((PASS + 1)); }
+bad()  { echo "FAIL: $1"; FAIL=$((FAIL + 1)); }
+
+# The library gate: exit 77 (suite SKIP) on targets without (kaappi process).
+printf '(cond-expand ((library (kaappi process)) (exit 0)) (else (exit 1)))\n' > "$DIR/gate.scm"
+if ! "$KAAPPI" "$DIR/gate.scm" >/dev/null 2>&1; then
+    echo "SKIP: (kaappi process) not available on this target"
+    exit 77
+fi
```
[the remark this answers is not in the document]
baijum: Fixed in 90c8a2f5: the gate program imports (scheme base)/(scheme process-context) and exits with a dedicated status 9 for "library absent"; only that status skips (77), 0 proceeds, and anything else fails the suite loudly.

---

src/tests_process.zig
```diff
@@ -0,0 +1,970 @@
...
+            break;
+        };
+    }
+    if (!filled_up) return error.SkipZigTest; // could not exhaust the table
+
+    // Control: a staged duplicate is impossible right now.
+    try std.testing.expect(platform.dupCloexecAtLeast(1, 3) < 0);
+
+    try expectTrue(vm,
+        \\(let ((p (spawn-process '("true") 'stdout: (current-output-port))))
```
baijum, line 976: This regression currently breaks both emulated architecture gates: `ppc64le-test` and `s390x-test` fail in this `spawn-process` call. Filling the parent table completely is not transparent under QEMU/binfmt; the interpreter needs a spare descriptor before the child reaches exec-time CLOEXEC cleanup, so even the fixed self-redirect path raises `cannot spawn process "true"`. Please test the no-staging decision without requiring an exec under a zero-free table, or leave enough descriptors for binfmt while arranging that an unwanted staging duplicate still fails. CI evidence: jobs 99546383549 and 99546383648.

src/tests_process.zig
```diff
@@ -0,0 +1,970 @@
...
+    // reap in killAndReapFresh is a BLOCKING waitpid, so those paths
+    // structurally cannot leave a zombie — and the registry must be empty.
+    // A countdown that fires in the *caller's* kill/wait after a
+    // successful spawn abandons the Process by the caller's own hand (the
+    // documented GC.unreaped_processes window, not spawnImpl's contract);
+    // its SIGKILLed child may still be mid-death on a slow box, so such
+    // stragglers are drained rather than asserted against.
+    try std.testing.expectEqual(@as(usize, 0), gc.unreaped_processes.items.len);
+    var zst: c_int = 0;
+    while (platform.waitPid(-1, &zst, platform.WNOHANG) > 0) {}
```
baijum: This does not assert that no child was abandoned. `waitpid(-1, WNOHANG) == 0` means a `sleep 30` child is still running, but the loop simply exits and the test passes; positive results are silently reaped rather than reported as leaked zombies. The registry assertion is also after `collectFull`, which removes an unreachable Process entry even when its child is still live. Mutation-checking out the new post-spawn `errdefer` can therefore leave live sleepers while this epilogue stays green. Record the spawned PIDs or require a bounded transition to `-1/ECHILD`, failing on a persistent 0, then clean up any leak after recording the failure.

src/primitives_process.zig
```diff
@@ -0,0 +1,1248 @@
...
+
+/// Terminate and reap a child spawned moments ago that no caller will ever
+/// be able to wait on (a post-spawn allocation failed). SIGKILL cannot be
+/// caught, so the blocking reap is bounded by the process teardown itself;
+/// EINTR retries, any other waitpid failure gives up (nothing more can be
+/// done on an OOM path).
+fn killAndReapFresh(pid: c_int) void {
+    _ = platform.procKill(pid, 9);
+    var st: c_int = 0;
+    while (platform.waitPid(pid, &st, 0) < 0) {
```
baijum: This new blocking wait runs while an SRFI-18 child VM still reports `.running`, unlike both `process-wait` and the spawn syscall. If child teardown is delayed, a concurrent parent collection spins in `markLiveChildRoots` waiting for a safepoint this VM cannot reach. Also, line 690 discards a failed `kill`; SIGKILL is only bounded after successful delivery, and an uninterruptible child can delay exit even then. Please apply the same `setCollectionInNative`/`setCollectionRunning` protocol here, check the kill result before entering an unconditional blocking wait, and retain a reap record if cleanup cannot complete synchronously.

src/platform.zig
```diff
@@ -158,7 +158,133 @@ pub fn isatty(fd: fd_t) bool {
...
+    if (flags < 0) return null;
+    const nonblock: c_int = @intCast(@as(u32, @bitCast(std.posix.O{ .NONBLOCK = true })));
+    if (std.c.fcntl(fd, std.posix.F.SETFL, flags | nonblock) < 0) return null;
+    return flags;
+}
+
+/// Restore fcntl status flags saved by setFdNonblockingForTeardown.
+pub fn restoreFdStatusFlags(fd: fd_t, flags: c_int) void {
+    if (comptime is_wasm or is_windows) return;
+    _ = std.c.fcntl(fd, std.posix.F.SETFL, flags);
```
baijum: Saving and restoring the whole status word does not make this safe for shared open-file descriptions. During the drain, every dup/child alias transiently observes O_NONBLOCK and can fail with EAGAIN; if an alias concurrently changes O_APPEND/O_ASYNC/O_NONBLOCK, this stale snapshot silently overwrites that change. The unchecked restoration can also leave aliases permanently nonblocking on EINTR or another failure. Since GC flushing is best-effort, avoid toggling a blocking shared description during finalization (skip that drain); at minimum only attempt it with pending bytes, merge current flags, and retry/check restoration. Add a synchronized alias test, not only a final-state check.

src/tests_process.zig
```diff
@@ -0,0 +1,970 @@
...
+    // require the full collection to come back promptly.
+    // 12000 flushed fits even the smallest fixed pipe (OpenBSD's 16 KiB),
+    // so the flush itself can never block; the 60000-byte unflushed tail
+    // then exceeds the REMAINING capacity on every platform (16 KiB fixed
+    // and 64 KiB big-pipe alike), so the sweep's drain always meets a full
+    // pipe partway through.
+    try expectTrue(vm,
+        \\(let* ((p (spawn-process '("/bin/sleep" "3") 'stdin: 'pipe))
+        \\       (in (process-stdin p)))
+        \\  (write-string (make-string 12000 #\x) in)
```
baijum: 12,000 bytes is still not a portable guaranteed prefill on Linux. After the per-user pipe-page soft limit is exceeded, Linux may create pipes with only two pages (commonly 8,192 bytes), so this explicit flush can block against `sleep` before the timer starts and mask the collector behavior again. Keep the prefill at or below one page/PIPE_BUF (for example 4 KiB) and make the unflushed tail larger than the normal 64 KiB capacity.

src/tests_process.zig
baijum: Fixed in 2e8072f8: the test now runs a plain-spawn control under the full table first and skips when even that cannot exec — which is exactly the QEMU/binfmt environments (the interpreter needs a spare descriptor before the child reaches its CLOEXEC cleanup), so the ppc64le/s390x legs skip while real kernels keep the full assertion. The staging-impossible control (dupCloexecAtLeast must fail) still runs before the skip decision, so the environments that do run it genuinely prove the no-staging path.

---

src/tests_process.zig
baijum: Fixed in 2e8072f8, both halves. The registry assertion now runs BEFORE collectFull (right after the WNOHANG sweep), so an unreachable-but-live-child Process cannot be dropped out of sight first — mutation-verified: no-op'ing the errdefer fails it immediately with 'expected 0, found 3'. And the epilogue now requires a bounded transition to -1/ECHILD (up to 5 s of WNOHANG polling): a persistent 0 — a live abandoned child — fails the test, positive returns are killed stragglers mid-death, reaped and retried. On the pid-recording alternative: the caller's kill runs before its first post-spawn GC allocation, so a countdown cannot abandon an un-killed child by the caller's hand — every live survivor is spawnImpl's fault, which is what the ECHILD requirement pins without a pid list.

---

src/primitives_process.zig
baijum: Fixed in 2e8072f8: killAndReapFresh now (1) checks the kill result and never enters the blocking wait unless the SIGKILL demonstrably went out (a failed kill falls back to one WNOHANG probe), (2) publishes a child-thread VM via the same setCollectionInNative/setCollectionRunning protocol as process-wait around the blocking reap, and (3) returns whether the reap completed — the errdefer keeps the Process's registry entry when it did not, so the ordinary WNOHANG sweeps finish the job instead of the pid being forgotten. The uninterruptible-sleep stretch is noted in the doc comment; with the in-native state published it is GC-safe and bounded by the child's own teardown.

---

src/platform.zig
baijum: Fixed in 2e8072f8 by taking your primary option: the sweep no longer touches file status flags at all. The drain is gated on seekability instead — a seekable fd is a regular file whose blocking write completes without waiting on any peer (preserving the flush-on-collect semantics file ports have always had), while an unseekable one (pipe/socket/FIFO/tty) drops its remainder outright, exactly the best-effort contract. No transient O_NONBLOCK window for aliases, no restore to fail, no stale-snapshot overwrite of concurrent flag changes. The dup-alias regression still stands (the alias must come out of a collection blocking) and the bounded-collect regression now exercises the skip path.

---

src/tests_process.zig
baijum: Fixed in 2e8072f8: the flushed prefill is 4096 bytes — at or below one pipe page even under Linux's per-user pipe-page soft limit — and the unflushed tail is 68000 bytes, larger than a full 64 KiB capacity, so the abandoned buffer always exceeds anything a drain could sink. (With the seekability redesign above the sweep never writes to the pipe at all; the sizes keep the test meaningful as a bounded-collect guard either way.)

---

src/primitives_process.zig
```diff
@@ -0,0 +1,1276 @@
...
+    // (kaappi#2442 review — the kill/reap must cover the port
+    // constructions, not just allocProcess). Terminate and reap the fresh
+    // child, and record the status so freeObject does not reap a reused
+    // pid later.
+    errdefer {
+        if (killAndReapFresh(pid)) {
+            proc.status = 9; // (signaled . 9): the raw wait status of SIGKILL
+            removeFromUnreaped(gc, proc);
+        }
+        // else: keep the registry entry — the WNOHANG sweeps reap it when
```
baijum, line 665: This fallback is not durable. `unreaped_processes` is deliberately non-rooting, so after this error unwinds the Process is unreachable; the next collection removes this entry and destroys the object even when its WNOHANG probe returns 0. The `allocProcess` failure path is worse: no Process/entry exists at all and its false cleanup result is ignored. Thus any synchronous cleanup failure still permanently forgets a live PID. Transfer unreaped survivors into a pre-reserved PID-only orphan list independent of Process lifetime, and test a forced kill failure across `collectFull` through eventual reap.

src/primitives_process.zig
```diff
@@ -0,0 +1,1276 @@
...
+    }
+    const wait_vm: ?*vm_mod.VM = if (vm_mod.vm_instance) |vm|
+        (if (!vm.owns_globals) vm else null)
+    else
+        null;
+    if (wait_vm) |vm| vm.setCollectionInNative();
+    defer if (wait_vm) |vm| vm.setCollectionRunning();
+    var st: c_int = 0;
+    while (true) {
+        const r = platform.waitPid(pid, &st, 0);
```
baijum, line 718: Publishing `.in_native` protects the parent collector, but this remains an unbounded wait in the failing `spawn-process` call itself. Successful `kill(SIGKILL)` only queues the signal; a child in uninterruptible kernel/filesystem I/O can remain alive indefinitely, contrary to the comment above, and this OOM path never returns. Use WNOHANG/bounded polling and transfer a surviving PID into durable orphan-reap state rather than blocking until teardown completes.

src/tests_process.zig
```diff
@@ -0,0 +1,998 @@
...
+    // 8 KiB pipes — kaappi#2442 review), so the explicit flush can never
+    // block; the 68000-byte unflushed tail then exceeds even a 64 KiB
+    // pipe's remaining capacity, so the abandoned buffer is guaranteed
+    // larger than what any drain could sink.
+    try expectTrue(vm,
+        \\(let* ((p (spawn-process '("/bin/sleep" "3") 'stdin: 'pipe))
+        \\       (in (process-stdin p)))
+        \\  (write-string (make-string 4096 #\x) in)
+        \\  (flush-output-port in)
+        \\  (write-string (make-string 68000 #\y) in)
```
baijum, line 641: 68,000 bytes does not exceed every supported Linux pipe capacity. Pipe size is page-based; a 64 KiB-page AArch64 kernel can provide a 1 MiB default pipe, so restoring the broken unconditional drain can consume this whole tail and keep the test green. Fill the actual pipe to EAGAIN before collection (then restore blocking state), or use a tail above the maximum supported default with a longer-lived child and explicit cleanup.

src/tests_process.zig
```diff
@@ -0,0 +1,998 @@
...
+    // is a LIVE abandoned child (the mutation shape: spawnImpl failing
+    // without killing) and fails the test; positive returns are killed
+    // stragglers mid-death on a slow box, reaped and retried. The caller's
+    // own kill runs before its first post-spawn allocation, so a countdown
+    // cannot abandon an un-killed child by the caller's hand here.
+    var reached_echild = false;
+    var tries: u32 = 0;
+    while (tries < 500) : (tries += 1) {
+        var zst: c_int = 0;
+        const r = platform.waitPid(-1, &zst, platform.WNOHANG);
```
baijum, line 748: This still masks the exact zombie regression. A positive result means production failed to reap a terminated child, but the test silently performs that reap and can later pass at ECHILD; any negative result is also accepted as ECHILD, including EINTR while a child remains. Record any positive result as a failure while reaping only for hygiene, retry EINTR, and accept only `errno == ECHILD`. `sweepUnreapedForTest` above should likewise report rather than hide a positive reap when validating production cleanup.

src/primitives_process.zig
```diff
@@ -0,0 +1,1276 @@
...
+/// rather than a pid being forgotten (kaappi#2442 review). The blocking
+/// reap runs only after the SIGKILL demonstrably went out (an unchecked
+/// kill followed by an unconditional wait could block on a child that was
+/// never signaled), under the same markable-in-native protocol as
+/// process-wait, since a child-thread VM blocking here would otherwise
+/// starve a concurrent parent collection (#1933 shape). SIGKILL cannot be
+/// caught, so the wait is bounded by the child's own teardown (an
+/// uninterruptible-sleep child can stretch that, but never indefinitely).
+fn killAndReapFresh(pid: c_int) bool {
+    if (platform.procKill(pid, 9) != 0) {
```
baijum, line 703: Cleanup always signals only `pid`, even when this spawn used `new-group: #t`. By the time post-spawn allocation fails, the child can already have started descendants in its new group; killing only the leader leaves those descendants running after `spawn-process` returns OOM and exposes no Process handle. Pass the group choice into cleanup and signal `-pid` for a successfully created new group, with a child-plus-sleeper regression at a port-allocation failure point.

src/gc_sweep.zig
```diff
@@ -426,25 +427,49 @@ pub fn freeObject(gc: *GC, obj: *Object) void {
...
+                // instead the drain is gated on seekability: a seekable fd
+                // is a regular file, whose blocking write completes without
+                // waiting on a peer (the flush-on-collect file ports have
+                // always had), while an unseekable one (pipe, socket, FIFO,
+                // tty — everything with a peer or flow control) simply
+                // drops its remainder, exactly the best-effort contract.
+                const drain_ok = if (comptime platform.is_windows or platform.is_wasm)
+                    true // the Windows arms below have their own non-blocking gates
+                else
+                    port.nonblocking or port.write_buf == null or
```
baijum, line 450: `port.nonblocking` is only cached per Port, while O_NONBLOCK belongs to the shared open-file description and can be cleared through any dup/child alias. If an alias clears it after the Port entered scheduler mode, this branch still drains an unseekable full pipe with blocking `write`, wedging GC. Do not use the cached flag to justify finalizer writes to unseekable descriptors; always drop pending pipe/socket/FIFO/TTY output during sweep. A regression should set the Port nonblocking, clear the flag through a duplicate, fill the pipe, then bound collection.

Issue #1608, opened 2026-07-16
Windows: fd readiness backend (IOCP/overlapped) to lift the blocking-port degradation


baijum:
The Windows aarch64 port (#1606) deliberately ships without fd readiness, mirroring the WASI degraded mode: CRT fds have no `O_NONBLOCK`, so `maybeSetNonblocking` (src/primitives_io.zig) is a comptime no-op on Windows, no read ever EAGAINs, nothing registers with the reactor, and fiber I/O degrades to single-fiber blocking. The reactor there (`WindowsEventBackend`, src/reactor.zig) is timer+notify only — `thread-sleep!`, timed channel/mutex waits, and KEP-0002 cross-thread wakeups all work, but a fiber blocked on a pipe/socket read blocks its whole OS thread.

Lifting this needs a real Windows readiness design, which does not map 1:1 onto the KEP-0001 readiness model:

- Win32 has no unified readiness API for arbitrary handles. IOCP is *completion*-based (submit an overlapped read, get notified when data has landed), which inverts the port layer's "wait until readable, then retry the syscall" protocol (`fiber.waitForFd` + the yield-retry re-execution contract, partial progress stashed in `port.read_buf`).
- CRT integer fds would need `_get_osfhandle` + overlapped I/O on the underlying HANDLE, or the port layer grows a per-platform descriptor type.
- Sockets (the main beneficiary, via kaappi-net's `fd->port` bridge, #1478) could alternatively use WSAEventSelect/`select` — a smaller readiness-shaped first step that leaves files/pipes blocking.

Suggested staging: (1) sockets-only readiness via WSAEventSelect wired into the existing auto-reset-event wait, (2) evaluate whether pipes/files justify a completion-based rework of the retry protocol.

Design constraints and the current backend's contract are documented in docs/dev/windows.md ("Deliberate degradations") and the WindowsEventBackend comments.

baijum:
[...]

PR #2455, opened 2026-09-01
Add run-process and the process-timeout condition (KEP-0022 Phase 4)

### LINKED ISSUE #2417: KEP-0022 Phase 4: run-process high-level API, process-timeout condition, docs, native-tier coverage
Phase 4 of [KEP-0022: Native Subprocess Support](https://github.com/kaappi/keps/blob/main/keps/0022-subprocess-support.md). Depends on Phases 1–3 (#2414, #2415, #2416).

## Scope

- **`run-process`**: one-shot spawn → feed optional `input:` → drain stdout/stderr concurrently via internal fibers (the fiber-native answer to Python's `communicate()` deadlock) → wait → `(values status out err)`.
- **`timeout:`**: kill the process group, drain what was produced, raise a `process-timeout` condition carrying the partial output (KEP-0005 taxonomy; spawn failures raise the file-error family with errno detail — program-not-found vs. permission).
- **Native-tier evidence**: `tests/scheme/compile/*.sh` scripts exercising `spawn-process`/`run-process` through `kaappi compile` — .scm tests are interpreter-only evidence.
- **Docs**: `docs/dev/` document for the subsystem (indexed in `docs/dev/README.md`); kaappi.github.io guide page + procedures page + a cookbook recipe ("Running external programs"); note the `(cond-expand ((library (kaappi process)) …))` gate and WASI absence in the conformance/stability pages as appropriate.
- **KEP hygiene**: reconcile KEP-0022 with what shipped (as-built amendments if anything drifted), settle remaining unresolved questions (pty deferral note, cross-thread wait, status encoding), and move the KEP toward Final per the keps process.

## Acceptance

- Cookbook example runs as written on macOS + Linux.
- `mkdocs build --strict` green on the docs repo.
- CHANGELOG entry; `kaappi features`/library listing reflects `(kaappi process)` where applicable.

baijum:
Closes #2417.

`spawn-process` gives you a child and three ports; getting a captured string out of it correctly does not follow. Feed a child's stdin and then read its stdout and you deadlock the moment either side passes the pipe buffer; read stdout to EOF before stderr and you deadlock on a child that talks on both. That is the bug Python's `communicate()` exists to work around, with a thread per stream and a `select` loop.

Kaappi already has the better answer, so `run-process` uses it: a feeder fiber writes `input:` and two drain fibers read the pipes, all three running while `process-wait` parks.

```scheme
(call-with-values
    (lambda () (run-process '("git" "log" "--oneline" "-5") 'directory: repo))
  (lambda (status out err) ...))
```

Options: `input:` (string or bytevector), `timeout:` (seconds), `output:` (`'string` default, or `'bytevector`), `directory:`, `env:`, `new-group:`.

## Why it is Scheme

Spawning and joining fibers is dispatch-loop work a native frame cannot do, so `run-process` is a `bootstrapStub` whose body `vm_bootstrap.install` evaluates — the same route `map` and `for-each` take, and `%process-spawn` was added in Phase 1 as the stable entry point for exactly this.

The install is gated on **that primitive being registered**, not on a mode flag, because the library is absent in two places for two unrelated reasons: WASM omits the module outright, and `--sandbox` skips every `.sandbox = false` spec. (`main.zig` also sets `vm.sandbox_mode` only *after* `install` runs, so a flag check would have been wrong as well as less direct.) Evaluating the definition in either would have taken VM startup down on an undefined variable — which is exactly what the first draft did.

## The condition

`timeout:` needs somewhere to put what the child produced before it stalled, since the values return never happens. It raises a condition carrying it — `process-timeout?`, `process-timeout-stdout`, `process-timeout-stderr` — on a new `ErrorType`, following the `channel_timeout` precedent. The partial output rides `uncaught_reason` rather than the irritants list: an uncaught timeout prints its irritants, and a child that wrote megabytes before stalling should not print them.

```
$ kaappi stall.scm
stall.scm:2: error[KP3000]: run-process: timed out ("sh" "-c" "printf a; sleep 5") 0.2
```

Spawn failures are unchanged — still the file-error family with errno, so program-not-found and permission-denied stay distinguishable.

## Four choices worth naming

- **`timeout:` implies `new-group: #t`.** `process-kill` refuses `'group:` on a child sharing the parent's own group, and the group kill is also what lets the drains reach EOF when a grandchild inherited the pipe. An explicit `new-group: #f` gets a child-only kill and the risk that comes with it.
- **The timeout kill is SIGKILL.** A timeout is a bound; a child ignoring SIGTERM would make it a suggestion. Python's `run()` kills for the same reason.
- **stdin is `'null` unless `input:` is given,** so a one-shot capture never blocks on the terminal (Go's `exec.Cmd` default).
- **`output: 'bytevector` exists** because the default path ends in `utf8->string`, which would otherwise fail a binary-producing child with a type error naming a procedure the caller never called.

One implementation note: the two places `run-process` swallows an error (EPIPE on the feed, best-effort joins after a kill) use `call/cc` + `with-exception-handler` rather than `guard`. `guard`'s desugaring reaches `%unwind-to-escape`, which `install` purges from globals immediately after evaluating the definition.

## Tests

- `src/tests_process_run.zig` (new, 13 tests) — the three values, `input:` as string and bytevector, EPIPE on a deaf child, all three streams past the pipe buffer at once, `output:`, `directory:`/`env:`, the timeout condition and its accessors, the grandchild tree kill, option errors, spawn failure staying a file error, and two calls overlapping in sibling fibers. Split from `tests_process.zig`, which is already near the 1500-line ceiling. Green under `-Dgc-stress=true`.
- `tests/scheme/process/process-run.scm` (new, 16 assertions) — the same contract with kaappi itself as the child, so it runs on the Windows tier too.
- `tests/scheme/compile/process-spawn-2414.sh` extended rather than duplicated. `run-process` has its own reason to be there: a compiled binary runs `vm_bootstrap.install` from `runtime_exports.zig`, not from `main.zig`.

Full suite green (2134 pass), `zig build test` green, cross-compiles clean for `x86_64-windows`, `aarch64-windows`, `x86_64-linux`, and `wasm32-wasi` (both the binary and the test-compile gate).

## Docs

`docs/dev/subprocess.md` is new and covers all four phases — file layout, the three park tiers, the thread-affinity rule, the two redirection rejections that are easy to trip over, the error taxonomy, and what the subsystem deliberately does not have.

The end-user pages (guide, procedures, cookbook recipe) and the KEP's as-built amendments live in the other two repos and follow as separate PRs; `kaappi features` needs no change, since KEP-0022 deliberately adds no `kaappi-process` feature identifier and portable code gates with `(cond-expand ((library (kaappi process)) ...))`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -530,7 +530,14 @@ jobs:
     needs: format
     if: needs.format.outputs.docs_only != 'true'
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    # 30 stopped being a hang-bound here, exactly as it did for the Debug and
+    # macOS legs above: healthy runs already sat at 22-26 min, and KEP-0022's
+    # subprocess suites pushed one to a cancellation at 30:21 with no error in
+    # the log. Every child a process test spawns is a whole /bin/sh emulated
+    # through binfmt QEMU, so this job pays a multiple of what the same tests
+    # cost natively. 40 restores real headroom; the timeout exists to catch
+    # hangs, not to race the suite.
+    timeout-minutes: 40
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Added
 
+- **`run-process` and the `process-timeout` condition (KEP-0022 Phase 4,
+  #2417)** — the one-shot layer over `spawn-process`:
+  `(run-process argv opt…)` spawns, feeds an optional `input:`, drains
+  stdout and stderr, waits, and returns `(values status out err)`. All three
+  streams move at once, through internal fibers that run while
+  `process-wait` parks — so a child that fills its stdin, stdout *and*
+  stderr buffers simultaneously completes instead of deadlocking, which is
+  the bug Python's `communicate()` exists to work around with threads.
+  Options are `input:` (string or bytevector), `timeout:` (seconds),
+  `output:` (`'string`, the default, or `'bytevector`, for children whose
+  output is not text), `directory:`, `env:` and `new-group:`; stdin is
+  `'null` unless `input:` is given, so a one-shot capture never blocks on
+  the terminal. `timeout:` kills the process group with SIGKILL, drains what
+  the child managed to produce, and raises a condition that `process-timeout?`
+  recognizes and `process-timeout-stdout`/`process-timeout-stderr` unpack —
+  the partial output has nowhere else to go, since the values return never
+  happens. It implies `new-group: #t` for that kill to be safe and to reach
+  a grandchild holding the same pipe. Spawn failures stay in the file-error
+  family with errno detail, so program-not-found and permission-denied
+  remain distinguishable. See `docs/dev/subprocess.md`, new with this
+  release and covering all four phases of the subsystem.
 - **`(kaappi process)` on Windows (KEP-0022 Phase 3, #2416)** — subprocess
   support now covers every hosted platform, with the same Scheme surface as
   POSIX. None of the machinery under it is a translation: spawn is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   the child managed to produce, and raises a condition that `process-timeout?`
   recognizes and `process-timeout-stdout`/`process-timeout-stderr` unpack —
   the partial output has nowhere else to go, since the values return never
-  happens. It implies `new-group: #t` for that kill to be safe and to reach
-  a grandchild holding the same pipe. Spawn failures stay in the file-error
+  happens. It implies `new-group: #t`, because only a group kill reaches a
+  grandchild holding the same pipes; an explicit `new-group: #f` alongside
+  `timeout:` is refused rather than accepted as an unbounded wait. Spawn failures stay in the file-error
   family with errno detail, so program-not-found and permission-denied
   remain distinguishable. See `docs/dev/subprocess.md`, new with this
   release and covering all four phases of the subsystem.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -401,6 +401,17 @@ library and vendored code are excluded).
   lexically** — exactly inverted. Mutating shared state through a global is a
   live hazard (kaappi#1924), not a supported idiom.
 
+- **Subprocesses (KEP-0022)** — `docs/dev/subprocess.md`. `(kaappi process)`
+  spawns without `fork` (`posix_spawnp` / `CreateProcessW`), hands back pipe
+  ends as ordinary reactor-integrated fd ports, and reaps child exit *at the
+  reactor* (kqueue `EVFILT_PROC` / Linux pidfd / a Windows process HANDLE) so
+  `process-wait` parks a fiber instead of blocking the thread. `run-process`
+  is the one-shot layer on top, and the one part written in Scheme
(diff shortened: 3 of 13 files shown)


baijum:
## Review findings

All five verified against the code and acted on.

**`timeout:` with an explicit `new-group: #f` could hang forever** — the one that mattered. `grouped` honored the override, so the timeout path issued a child-only SIGKILL; a grandchild holding the child's stdout keeps the write end open, `drain-into` never sees EOF, and `fiber-join` blocks with no bound (`quietly` catches exceptions, it does not bound time). The call then neither returned nor raised, so the bound `timeout:` promises was void. The tree-kill test already asserted that mechanism; nothing covered the override. The combination is now **refused at option-parse time**, and the kill branch is unconditionally the group form.

**Three tests did not test what they claimed:**

- The EPIPE feed used 65536 bytes — *exactly* Linux's default pipe capacity, not past it — so the write could land in the buffer whole and never break, and `st = 0` passed either way. Now 256 KiB, in both suites.
- `process-run.scm` used `current-jiffy`/`jiffies-per-second` without importing `(scheme time)`. They resolved anyway; that is luck, not a contract.
- The bulk-drain test scaled to 4096 bytes under `-Dgc-stress`, below the pipe buffer, so the stress leg ran a backpressure test that could not fail. The scale-down is gone — the cost under stress is per allocation, and a bulk read is a handful of them, so the byte count was never what made that leg slow.

**Test helpers:** six single-assertion tests now call `th.expectEvalTrue` directly with no local VM. The local `expectTrue` stays for the tests that share one interaction environment across several evals, with a comment saying so — mixing the two inside one test leaves `gc_instance` dangling when the helper's context deinits.

The one part I did not take is the concurrency critique of the shell-based flood test: that test no longer exists. It used `yes | head -c`, and `head -c` is a GNU/FreeBSD extension OpenBSD's `head` lacks, so it was replaced with two `/bin/cat` cases (stdin→stdout and stdin→stderr, each past the pipe buffer) before this review landed.

## Also: a Windows hang this PR's suite exposed — #2459

`windows-x64-test` cancelled at 45 minutes with `process-run.scm` hung and no error in the log. Reproduced on the Windows 11 ARM64 reference VM and bisected exactly: a stdin feed of **4096 bytes completes, 4097 hangs forever**, and only with a fiber scheduler running.

The cause is not in this subsystem. `platform_win_pipe.pipeWrite` treats `NtQueryInformationFile`'s `WriteQuotaAvailable` as a would-block oracle, and it is not one — instrumented on the VM, a freshly connected, completely *empty* child-stdin pipe reports 0 whenever the child is parked in a read on the far end, the exact state in which the write would succeed instantly. The writer parks on a number that never moves. `run-process` merely makes it easy to reach, since it always runs a scheduler and always feeds through a pipe.

I implemented the obvious fix (never refuse; let the blocking write run). It fixes this case completely — and hangs the opposite one, `tests_port_io`'s `#1608` writer whose only reader is a sibling fiber on the same thread. So it is reverted, the dead end is written up on the issue, and the real fix (overlapped pipe handles, libuv's route) belongs there rather than here.

This PR does not paper over it: `process-run.scm` caps only its **stdin** leg at the pipe buffer on Windows, both output legs stay full size on every platform, and the limitation is stated in `docs/dev/subprocess.md` and on the guide page. 18/18 assertions verified on the real Windows ARM64 box.

## Other CI

- `riscv64-test` was cancelled at 30:21 with nothing but orphaned zig processes in the log — the signature the `test` matrix comment already describes twice. Healthy runs there sat at 22–26 min against a 30-min cap before this branch, and every child a process test spawns is a whole `/bin/sh` under binfmt QEMU. Raised to 40, same fix as #1635 (macOS) and #1728 (Debug).
- The OpenBSD spawn-failure failure was mine: POSIX leaves it unspecified whether a failed PATH search fails at `posix_spawnp` or lets the child exec fail and exit 127, and OpenBSD takes the second. The strict assertion now uses a path; the bare-name case is pinned to either permitted outcome, and #2456 tracks closing the gap with an exec error pipe.
- `netbsd-test` fails on `main` independently of this branch.

Issue #2447, opened 2026-09-01
tests_step crashes the unit-test binary under podman aarch64 (Virtualization.framework VM)


baijum:
## Symptom

Inside the `kaappi-builder` aarch64 Linux container (podman on macOS, Virtualization.framework VM), the bounded-step suite kills the whole unit-test binary:

```
zig build test -Dtest-filter=tests_step
...
451/1887 tests_step.test.stepper: a form error is reported but the program still finishes...<test>:1:1: error[KP3002]: type error in 'car': expected pair, got 5
452/1887 tests_step.test.stepper: a form that pauses then raises unwinds roots cleanly...<test>:1:34: error[KP3002]: type error in 'car': expected pair, got 5
failed command: ./.zig-cache/o/<hash>/unit-tests --cache-dir=... --listen=-
```

The two KP3002 lines are those tests' *expected* stderr output; the binary then dies with no pass/fail summary — a crash (segfault or ReleaseSafe safety abort), not a test failure. Because the runner dies, the podman leg loses the rest of the unit suite with it.

## Scope

- Reproduces **identically on unmodified `main`** (ffb7d3c9) and on the kaappi#2415 branch — pre-existing, surfaced while running the `/linux-test` battery for PR #2445.
- Deterministic in that environment (three for three across seeds).
- Does **not** reproduce: macOS natively, ubuntu-arm CI (same suite green on every `main` run), FreeBSD/OpenBSD/NetBSD reference VMs, or the x86_64 Rosetta container (which fails elsewhere but runs `tests_step` past this point).

## Repro

```
podman run --rm -v /path/to/kaappi:/src:ro kaappi-builder bash -c '
  mkdir -p /workspace && cp -r /src/* /workspace/ && cd /workspace &&
  rm -rf .zig-cache zig-out && zig build test -Dtest-filter=tests_step'
```

## Where to look

The failing neighborhood is the `vm_step.Stepper` error paths (kaappi#2283): the crash lands in or right after the two erroring-form tests (`a form error is reported but the program still finishes`, `a form that pauses then raises unwinds roots cleanly`), each of which drives a `(do ... 40000 ...)` loop to a raise under a small step budget. A first bisect should establish whether the crash is in the raise-unwind bookkeeping under that VM's environment (signal delivery, stack sizing) or an environment artifact of Virtualization.framework itself — the same class of quirk that already makes the JIT execution tests skip in this container.

## Priority

[...]

PR #2461, opened 2026-09-01
Make SRFI 231 array-copy continuation-safe

### LINKED ISSUE #2454: SRFI 231 array-copy's direct fill overwrites an array already returned when a getter re-invokes a continuation
**Kaappi version**

HEAD with #2452 applied. Not observable before it — see "How it surfaced".

**Platform**

macOS aarch64, ReleaseSafe

**Description**

`%array-copy-impl` in `lib/srfi/231/views.sld` fills the destination array
directly: one `interval-for-each` pass that calls the source's getter and the
destination's setter per index, into a `dest` allocated before the loop starts.

If a getter captures a continuation and that continuation is re-invoked *after*
`array-copy` has returned, the resumed fill writes into **that same `dest`** —
the array the first copy already handed back. The caller's array mutates under
it, and the re-entry produces no fresh array at all.

SRFI 231 requires `array-copy` (the non-`!` one) to be safe under exactly this;
it is the whole documented difference between `array-copy` and `array-copy!`.
The reference implementation accumulates the source's values into a list first
and only then materializes a destination, so a re-entry re-runs the accumulation
and gets its own destination.

`array-append`, `array-stack`, `array-block` and `array-decurry` all delegate to
`array-copy` for the fill, so one loop accounts for five procedures.

**This is a known, deliberate scope reduction, not an oversight.** The comment
above `%array-copy-impl` says so in as many words — "a direct fill loop, no
accumulate-to-a-list-first indirection ... getters that escape and re-invoke a
captured continuation mid-copy are exotic enough that the extra allocation and
complexity isn't justified for this phase". This issue is to revisit that call
now that the scenario is reachable and testable.

**Minimal reproduction**

The official suite's own case (`tests/scheme/srfi/srfi231-official.scm:5904`):

```scheme
(let* ((cont #f)
       (call-cont #t)
       (domain (make-interval '#(2 2)))
       (A_ (lambda (i j)
              (lambda (c) (if (= i j 0) (set! cont c)) 1))))
       (A (make-array domain A_))
       (array-list '()))
  (let ((temp (array-copy A)))
    (set! array-list (cons temp array-list)))
  (if call-cont (begin (set! call-cont #f) (cont 4)))
  (for-each (lambda (result truth) (test 737 6727 (array->list* result) truth))
            array-list
            '(((4 1) (1 1))
              ((1 1) (1 1)))))
```

**Full output**

```text
FAIL 737 (line 6727) result=((4 1) (1 1)) expected=((1 1) (1 1))
```

The re-entry's array is correct; the *first* array is not. It should still read
`((1 1) (1 1))` and instead reads `((4 1) (1 1))` — the resumed fill overwrote
element (0,0) of the array the first `array-copy` returned.

**How it surfaced**

These five forms could not run at all until #2451. Their getters re-invoke a
continuation captured under a non-tail `apply` (SRFI 231's multi-indices are
variadic, so every getter/setter call goes through `apply`), which raised
`KP3000: continuation cannot resume across a returned native call` — 11
diagnostics per suite run. #2452 removed that engine limit, the forms now
execute, and they reach this library behaviour instead. Recorded there as
known-divergence ids 737-741 with a pointer back to `views.sld`, so the suite
accounts for them rather than failing; prune those five entries when this lands.

Note the exchange: a loud error became a silently wrong array. That is the
argument for revisiting the scope reduction rather than leaving it documented.

**Does it panic or raise a Scheme error?**

Neither — it returns a wrong answer, silently.

**Suggested resolution**

Collect the source's values before allocating the destination, then fill:

- The collection must be **continuation-safe itself**, so it cannot use a
  `set!` accumulator — a re-entry does not undo a `set!`. A functional
  accumulator threaded through the nested-loop walk (build reversed, `reverse`
  once at the end) keeps lexicographic order and re-runs cleanly. Note that
  `interval-fold-left`/`interval-fold-right` in `lib/srfi/231/intervals.sld`
  both use `set!` accumulators, so neither can back this as written.
- `array-copy!` may keep the direct fill — the spec permits the `!` variant to
  skip precisely this guarantee, and `assembly.sld` already documents the four
  `!` twins as aliases on that basis.
- Cost is one cons per element on top of a loop that already allocates a rest
  list per `apply getter` and per `apply dest-setter`, so the "extra allocation"
  the original comment weighed is marginal against what the loop spends already
  — but it is a real regression for large copies and worth measuring before
  committing to it.

**Priority rationale**

`priority: medium` per `docs/dev/github-issues.md`: behaviour is wrong against a
spec, and silently — which under the rubric aggravates *within* the level rather
than promoting it. It does not reach `high`, because the path is not one users
actually reach: it needs an array getter that captures a continuation and
re-invokes it after the copy has returned, which SRFI 231 itself describes as
exotic, and which no ordinary program does.

baijum:
Closes #2454

## What

`%array-copy-impl` filled the destination directly — one `interval-for-each`
pass calling the source's getter and the destination's setter per index, into
a `dest` allocated *before* the loop started. A getter that captured a
continuation and re-invoked it after `array-copy` returned kept writing into
the array the first copy had already handed back, so the caller's array
silently mutated under it and the re-entry produced no fresh array at all.
That is precisely the guarantee the spec gives `array-copy` over `array-copy!`.

`array-copy` now collects every source value through a **functional**
accumulator (threaded through a nested-loop walk, built reversed, `reverse`d
once — no `set!`, and `interval-fold-left`/`right` were unusable here because
both accumulate via `set!`) *before* the destination exists, then fills it in
a second pass. A re-entry re-runs the collection from its capture point and
materializes its own destination. `array-append`, `array-stack`, `array-block`
and `array-decurry` delegate to `array-copy`, so one change resolves all five
known-divergence ids (737–741), pruned from the official suite's ledger here.
`array-copy!` keeps the direct fill, which the spec explicitly permits, and
the `assembly.sld` header note about the four `!` twins is updated (they stay
pure aliases of the non-`!` forms and now inherit the safety — more than the
spec asks of them, never less).

## Evidence

- Repro (the official suite's own case): main returns `(((4 1) (1 1))
  ((4 1) (1 1)))`; this branch returns `(((4 1) (1 1)) ((1 1) (1 1)))` — the
  re-entry gets its fresh array, the first copy stays intact.
- `tests/scheme/srfi/srfi231-official.scm`: **10936 passed, 1 known
  divergence (the unrelated id 147), 0 unexpected failures, 0 count
  mismatches** — on main it was 10931 passed + 6 divergent rows under ids
  737–741.
- All seven SRFI 231 suite files green (views 114, assembly 69, arrays 105,
  combinators 88, intervals 87, storage-classes 249, hub 17).
- Perf, measured as the issue asked: 200×300 copies at 0.327s → 0.327s
  (specialized u8) and 0.403s → 0.395s (generic getter) per copy — the extra
  cons per element is noise against the rest-list each `apply` already
  allocates.

## Note

`zig build test` still stops at the pre-existing `tests_step` crash (#2447,
reproduced on clean main and unchanged by this PR — no Zig source is touched
here). The Scheme suites above were run with isolated `KAAPPI_HOME` homes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,10 +74,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   spec's whole documented difference between `array-copy` and `array-copy!`
   is that the non-`!` one must be safe under exactly this. `array-copy` (and
   `array-append`, `array-stack`, `array-block`, `array-decurry`, which
-  delegate to it) now collect every source value through a functional
-  accumulator — no `set!`, so a re-entry re-runs the collection and
-  materializes its own destination — before the destination exists.
-  `array-copy!` keeps the direct fill, which the spec explicitly permits.
+  delegate to it) now collect every source value — before the destination
+  exists — into a pre-sized scratch vector indexed by a position threaded
+  functionally through the walk (a cons-list accumulator instead holds N
+  pairs live and allocates 2N more across the reverse: 18.0M total pair
+  allocations for that shape against this one's 16.0M at 1M elements; the
+  churn that remains is the library apply shape every fill loop shares,
+  tracked in kaappi#2464), so a re-entry re-runs the collection from its
+  capture point and materializes its own destination. `array-copy!`
+  keeps the direct fill, which the spec explicitly permits. One residual
+  exposure remains, shared with the reference implementation: a destination
+  storage class from `make-storage-class` contributes user-written setter
+  code to the copy-out, where a capture re-enters with the destination
+  already allocated and hands back the same object (with identical
+  contents).
 - **Continuations captured under a non-tail `apply` or `call-with-values`
   (#2451)** — resuming one raised `KP3000: continuation cannot resume across a
   returned native call`. Both reached their callee through a native
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- **SRFI 231 `array-copy` is continuation-safe** (#2454) — the fill wrote
+  directly into a destination allocated before the loop started, so a getter
+  that captured a continuation and re-invoked it after `array-copy` returned
+  kept writing into the array the first copy had already handed back: the
+  caller's array silently mutated under it (the official suite's own case —
+  the first copy reading `((4 1) (1 1))` instead of `((1 1) (1 1))`). The
+  spec's whole documented difference between `array-copy` and `array-copy!`
+  is that the non-`!` one must be safe under exactly this. `array-copy` (and
+  `array-append`, `array-stack`, `array-block`, `array-decurry`, which
+  delegate to it) now collect every source value through a functional
+  accumulator — no `set!`, so a re-entry re-runs the collection and
+  materializes its own destination — before the destination exists.
+  `array-copy!` keeps the direct fill, which the spec explicitly permits.
 - **Continuations captured under a non-tail `apply` or `call-with-values`
   (#2451)** — resuming one raised `KP3000: continuation cannot resume across a
   returned native call`. Both reached their callee through a native
--- a/lib/srfi/231/assembly.sld
+++ b/lib/srfi/231/assembly.sld
@@ -37,7 +37,11 @@
 ;;; source's values before the destination exists, so the aliases
 ;;; inherit continuation-re-entry safety -- more than the spec asks of
 ;;; the `!` variants (which may skip exactly that guarantee), never
-;;; less.
+;;; less. The `!` twins do pay the same scratch cost as the non-! ones
+;;; (one pre-sized vector, N words -- the aliasing trades that memory
+;;; for one code path; splitting them apart to reclaim it would
+;;; re-diverge five fill loops for a cost the spec's `!` exemption
+;;; exists precisely to allow skipping).
 (define-library (srfi 231 assembly)
   (import (scheme base) (srfi 1)
           (srfi 231 misc) (srfi 231 intervals) (srfi 231 storage-classes)
--- a/lib/srfi/231/assembly.sld
+++ b/lib/srfi/231/assembly.sld
@@ -28,15 +28,16 @@
 ;;; entirely to array-copy (already handling storage-class/mutable?/
 ;;; safe? option parsing and the materializing fill loop) rather than
 ;;; hand-rolling a separate fill mechanism per procedure. Their `!`
-;;; twins are implemented as pure aliases of the non-! version -- a
-;;; scope reduction already established for array-copy/array-copy! in
-;;; phase 3, now confirmed for these four too by reading the reference
-;;; implementation directly: every one of these four's `!` and non-`!`
-;;; entry points are two one-line wrappers around one shared internal
-;;; helper differing only in whether input arrays are eagerly
-;;; materialized before the fill runs, a distinction that is only
-;;; observable under the exotic multi-shot-continuation-re-entry
-;;; scenario the spec itself declares "an error" to trigger.
+;;; twins are implemented as pure aliases of the non-! version --
+;;; confirmed by reading the reference implementation directly: every
+;;; one of these four's `!` and non-`!` entry points are two one-line
+;;; wrappers around one shared internal helper differing only in
+;;; whether input arrays are eagerly materialized before the fill runs.
+;;; Since #2454 the shared path underneath (array-copy) collects the
+;;; source's values before the destination exists, so the aliases
+;;; inherit continuation-re-entry safety -- more than the spec asks of
+;;; the `!` variants (which may skip exactly that guarantee), never
+;;; less.
 (define-library (srfi 231 assembly)
   (import (scheme base) (srfi 1)
           (srfi 231 misc) (srfi 231 intervals) (srfi 231 storage-classes)
--- a/lib/srfi/231/views.sld
+++ b/lib/srfi/231/views.sld
@@ -314,55 +314,44 @@
     ;; documented difference from array-copy!): the re-entry gets a FRESH
     ;; array, and the array the first call already returned never mutates
     ;; under its holder. That requires collecting every source value
-    ;; BEFORE the destination exists, threading the values through a
-    ;; FUNCTIONAL accumulator -- a set! accumulator (or
-    ;; interval-fold-left/right, which use one) lets the re-entry keep
-    ;; appending to the first run's partial list, and a direct fill into
+    ;; BEFORE the destination exists, into a pre-sized scratch vector
+    ;; indexed by a position threaded FUNCTIONALLY through the walk --
+    ;; the threading is the safety mechanism (a set! position counter, or
+    ;; interval-fold-left/right, which use one, lets the re-entry keep
+    ;; counting on the first run's progress). The scratch itself is
+    ;; shared mutable state, deliberately: a re-entry resumes from the
+    ;; position captured at its capture point, so positions below it
+    ;; retain the first run's values and positions above are re-fetched
+    ;; -- observably identical to a functional cons accumulator, without
+    ;; that shape's cost (at 1M elements: 18.0M total pair allocations
+    ;; for a cons-list collector against 16.0M here, and its N live
+    ;; pairs gone; the churn that remains is the library apply shape
+    ;; every fill loop already shares -- kaappi#2464). A direct fill into
     ;; a pre-allocated destination (the pre-#2454 shape, and what
     ;; array-copy! still is -- the spec explicitly permits the `!`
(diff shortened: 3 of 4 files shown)


baijum:
## Review

The fix is correct and the evidence in the description reproduces exactly. One
measured concern (peak memory, not time) and one comment-accuracy point below;
neither touches the correctness of what this PR set out to fix.

### Verified

Built this branch on macOS aarch64 (ReleaseSafe), isolated `KAAPPI_HOME` per run.

- `tests/scheme/srfi/srfi231-official.scm`: **10936 passed, 231
  error-message-only, 1 known divergence, 0 unexpected failures, 0 count
  mismatches, out of 10937 evaluations** — matches the description.
- Swapping only `zig-out/lib/srfi/231/views.sld` back to `main`'s copy (same
  binary, same test file with the pruned ledger) gives **10931 passed, 5
  unexpected failures**. So the ledger prune really is the regression test for
  ids 737–741: it fails without the fix and passes with it, and it runs in CI
  on the ReleaseSafe legs. Good — that satisfies the "every bug fix needs a
  regression test" rule without a hand-written duplicate.
- All seven hand-written suites green at exactly the claimed counts: views 114,
  assembly 69, arrays 105, combinators 88, intervals 87, storage-classes 249,
  hub 17.
- Walk order in `%array-collect` / `%array-fill-from-list` matches
  `%interval-for-each-indices` (`lib/srfi/231/intervals.sld:154`) on every shape
  I probed — lex order `(0 0) (0 1) (0 2) (1 0) (1 1) (1 2)`, 0-dimensional
  domain yields exactly one element (`getter` applied to `'()`), an axis with
  `lo = hi` yields none. Both passes share the walk, so the copy is
  order-consistent by construction; the value checker now fires *before* the
  destination is allocated, which is a small improvement over the old
  half-filled-`dest`-then-error path.
- `markdownlint-cli2` clean on the CHANGELOG entry.

### 1. Peak memory regresses ~60× on large copies, and it scales linearly

The 200×300 benchmark in the description (60k elements) is large enough to show
that *time* is fine, but too small to show what actually changed:
`array-copy` now holds `N` cons cells live — `2N` transiently across the
`reverse` — before the destination exists. Same binary, only `views.sld`
swapped, `u8-storage-class` → `u8-storage-class`:

| copy | main peak RSS | this branch | time |
|---|---|---|---|
| 1000×1000 (1M elems) | 12.8 MB | **411 MB** | 5.48s → 5.47s |
| 2000×2000 (4M elems) | 21.8 MB | **1.26 GB** | 21.5s → 22.1s |

(Most of that is GC headroom the collector takes once the live set grows —
strictly-live is ~2 pairs/element — but peak RSS is what OOMs a process, and
it is linear in element count. A 100M-element copy that works today would want
tens of GB.)

Two things make this worth addressing before merge rather than after:

- Bulk `array-copy` is *the* common SRFI 231 operation, so this is a path users
  reach, not an exotic one — the opposite of the case being fixed.
- Via the `assembly.sld` aliases, `array-append!`, `array-stack!`,
  `array-block!` and `array-decurry!` all pay it too, for a guarantee the spec
  explicitly frees the `!` variants from. The header note calls that "more than
  the spec asks ... never less", which is true of the *semantics* but not of the
  cost.

A minimal change keeps the exact continuation semantics and drops both the
`reverse` and most of the memory: thread a lexicographic **index into a
pre-sized scratch vector** instead of a cons accumulator — `(cons val acc)`
becomes `(begin (vector-set! buf k val) (+ k 1))`, threaded through the walk
the same way `acc` is now. The re-entry semantics are identical (positions
below the capture point keep the first run's values; positions above are
re-fetched), the returned array is still filled by copying values *out* of the
scratch, and each re-entry still allocates its own `dest`. Cost goes from `2N`
pairs to `N` words.

A source-side fast path is the other obvious lever, but note it can't key on
`specialized-array?` alone: `specialized-array-share` (`views.sld:82`) builds
`new-indexer` from the *user's* `new-domain->old-domain`, so a shared view's
unsafe getter runs user code. The scratch vector avoids needing to classify
that at all.

### 2. `%array-fill-from-list`'s "no continuation can be captured mid-fill" isn't quite true

> only library code runs here (the destination's unsafe setter -- the value
> check already happened at collection), so no continuation can be captured
> mid-fill

`make-storage-class` is exported from `(srfi 231 storage-classes)`, and
`make-specialized-array` builds `raw-setter` as
`((storage-class-setter storage-class) body idx val)`
(`arrays.sld:235`) — so `dest-setter` is user code whenever the destination's
storage class is user-supplied. It is capturable:

```scheme
(define cont #f) (define call-cont #t)
(define sc (make-storage-class vector-ref
             (lambda (body i v)
                 (lambda (c) (if (= i 0) (set! cont c)) #f))
               (vector-set! body i v))
             (lambda (x) #t) make-vector vector-copy! vector-length 0 vector?
             (lambda (d) d)))
(define src (make-array (make-interval '#(2)) (lambda (i) (+ i 1))))
(define saved '())
(let ((t (array-copy src sc))) (set! saved (cons t saved)))
(if call-cont (begin (set! call-cont #f) (cont 99)))
(list (length saved) (eq? (car saved) (cadr saved)))
;; => (2 #t)   <- same array object handed back twice, and re-mutated
```

The impact is small — the values come from the already-collected list, so both
runs write identical contents and nothing value-visible breaks (which is why
the official suite can't see it) — but the returned object is not fresh, and
the comment states the opposite as an absolute. Worth either softening the
claim to name the user-storage-class case, or closing it: with the scratch-
vector shape from §1, `dest` is allocated after the scratch fill, so a capture
in the *scratch* setter already re-enters ahead of `dest` and only the final
copy-out remains exposed.

### 3. Nit

`%array-collect` and `%array-fill-from-list` are the same nested walk with a
different leaf and a different threaded value. Once §1 turns the accumulator
into an index, they differ only in the leaf body — worth folding into one
`%domain-fold`-style helper taking a leaf procedure, so the two can't drift out
of order-agreement later.

### Summary

Correctness, tests, docs and the ledger prune all check out — the fix does what
it says and the divergence really is resolved. §1 is the one I'd want resolved
before merge; §2 and §3 are cheap follow-ons that the same edit mostly covers.

baijum:
Thank you — all three points are addressed in 6ad52484.

**§1 — scratch vector, exactly the suggested shape.** `(cons val acc)` is
now `(begin (vector-set! buf k val) (+ k 1))` with `k` threaded
functionally through the walk; the collection still runs strictly before
the destination is allocated (inside the `let*` initializer — moving it
into the body would have let a getter capture re-enter *with* `dest`
already existing and re-mutate the returned array, so that placement is
load-bearing). The `reverse` and the N live pairs are gone.

Measured at 1M elements (u8→u8, warm homes, `--gc-stats`, this machine):

| | pairs allocated | closures | live heap peak |
|---|---:|---:|---:|
| main (direct fill) | 15.0M | 2.00M | ~10.7 MB |
| first cut (cons) | 18.0M | 2.00M | ~10.7 MB + N live pairs |
| 6ad52484 (scratch) | 16.0M | 2.00M | ~10.7 MB |

The delta over main's 15.0M is the second walk's leaf (no getter call in
the copy-out), which the two-pass design inherently needs.

One thing I have to flag honestly: **your main baseline does not
reproduce on this machine.** For the identical script/binary (only
`views.sld` at main's copy), I measure **669–671 MB peak RSS on main
itself**, stably across repeated runs — not 12.8 MB. `--gc-stats` on the
same run reports a heap that never exceeds ~10.7 MB, and the ranking
here is inverted relative to yours (main ~670 MB @ 1187 collections,
cons 256 MB @ 24, scratch ~670 MB @ 1224): peak RSS on both of our
machines is governed by the interaction between GC cadence and how freed
memory is returned to the OS, not by the collector's data structure.
That behavior — RSS at ~65× the live heap, freed memory never returned —
predates this PR and is unaffected by it; I've filed it as #2464 with
the reproduction and both suggested directions (returning pages after
collections, and the ~16-objects-per-element library `apply` shape that
feeds the churn). The scratch change delivers the part that is portable
and inherent: strictly fewer allocations and no live accumulator.

**§2 — comment corrected rather than closed.** You're right that
`dest-setter` is user code under `make-storage-class`, and your repro
gives `(2 #t)` on this branch too, unchanged by the scratch shape (the
capture is inside the copy-out, after `dest` exists — exactly the
residual you predicted). The comments and CHANGELOG now state it as the
same residual the reference implementation's accumulate-then-materialize
has, instead of the false absolute.

**§3 — one helper.** Both passes now go through `%domain-walk lowers
uppers leaf k` (the two leaves differ only in body), so they cannot
drift out of order-agreement.

Re-verified after the change: official suite 10936 passed / 0 unexpected
/ 0 count mismatches (ledger unchanged), all seven hand-written suites
green, the issue repro returns `(((4 1) (1 1)) ((1 1) (1 1)))`, and the
200×300 timing stays at parity (0.318s / 0.389s per copy specialized /
generic).

Issue #2446, opened 2026-09-01
NetBSD: srfi18-join-spawn-grandchild-2129.scm hangs intermittently (cross-thread join lost wakeup)


baijum:
## Symptom

`tests/scheme/srfi/srfi18-join-spawn-grandchild-2129.scm` — normally a sub-second test — intermittently runs until killed on NetBSD 10.1 aarch64. Measured on the reference UTM VM with a 90 s bound, 10 runs each:

| binary | timeouts |
|---|---|
| unmodified `main` (ffb7d3c9) | **7/10** |
| kaappi#2415 branch (fc8d66da) | 2/10 |

So this is pre-existing on `main`, not introduced by the Phase-2 work — the branch column is just the run that surfaced it. It also killed the `netbsd-test` CI job on PR #2445 (run 33437584832, killed after the suite's 60 s per-test bound), where `tests/scheme/srfi/srfi18-terminate-native-wait-1982.scm` timed out in the same pass; a rerun of the job was green. Not observed on FreeBSD/OpenBSD/macOS/Linux in the same session's full-suite runs.

## Why this looks like a lost wakeup, not slowness

A 60–90 s timeout on a test that completes in well under a second when it passes is a hang that only the kill ends. The test exercises exactly the machinery with known NetBSD-shaped history: OS-thread grandchildren (`thread-start!` from a started thread, kaappi#2129's handle/`live_descendants` accounting) resolving through the cross-thread notifier waits (kaappi#2395), on the kqueue backend where NetBSD needs the versioned `__kevent50` binding (`src/reactor.zig`). `srfi18-terminate-native-wait-1982.scm` timing out in the same CI pass points the same direction — both are waits that only another OS thread can resolve.

## Repro

On the NetBSD reference VM (ssh alias `netbsd`; the `/vm-test` skill has the cross-compile-and-ship recipe):

```
cd ~/kaappi && for i in 1 2 3 4 5 6 7 8 9 10; do
  KAAPPI_HOME=$(mktemp -d) timeout 90 ./zig-out/bin/kaappi \
    tests/scheme/srfi/srfi18-join-spawn-grandchild-2129.scm >/dev/null 2>&1 \
    || echo "iter $i TIMEOUT"
done
```

Failure rate is load-sensitive (the 7/10 baseline run followed a full suite on the same box), which fits a timing-window lost wakeup.

## Priority

[...]

PR #2462, opened 2026-09-01
Report OpenBSD exec failures at the spawn; close wrapped and test fds by default

### LINKED ISSUE #2422: CLOEXEC audit leftovers: test/dev fd pairs (testing_helpers, bench_reactor) stay inheritable

### LINKED ISSUE #2424: fd->port leaves foreign (FFI) descriptors inheritable — Linux children can inherit kaappi-net sockets
Found during the KEP-0022 Phase 1 CLOEXEC audit (#2414); deliberately deferred there because the audit's Phase 1 scope was the fd-opening sites in `platform.zig`, and `fd->port` wraps descriptors Kaappi did not open.

`fd->port` (`src/primitives_io.zig`, `(kaappi ffi)`) adopts whatever fd an FFI call handed back — a kaappi-net socket, an FFI library's pipe — without touching its `FD_CLOEXEC` state. On macOS the spawn path's `POSIX_SPAWN_CLOEXEC_DEFAULT` closes everything not installed by file actions, so this is masked there. On Linux (where the audit is the whole story, per KEP-0022), a child spawned with `spawn-process` inherits every FFI-held descriptor: a live listening socket or an accepted connection reaches the child, which is precisely the "bug factory" class the KEP's close-fds-by-default lesson calls out.

## Fix options

- `fd->port` sets `FD_CLOEXEC` via `platform.setFdCloexec` at wrap time — Kaappi never execs except through `spawn-process`, whose stdio install path (`posix_spawn_file_actions_adddup2`) clears CLOEXEC on exactly the child's slots, so this cannot break a legitimate pass-through; or
- the spawner takes a snapshot of port-wrapped fds and `addclose`s them in the file actions.

The first is one line and makes the wrap point the ownership boundary ("this port owns the fd; nobody else gets it"), matching how `platform.pipe` now behaves.

## Caveat to record either way

An FFI user who *wants* a foreign fd to survive into a child has no escape hatch after this change — worth a sentence in the KEP-0022 docs pass (Phase 4) saying pass-through fds must be prepared by the foreign side itself.

## Acceptance

- With a kaappi-net socket wrapped via `fd->port`, a `spawn-process` child on Linux reports no inherited fds beyond stdio (the `sh -c '[ -e /dev/fd/N ]'` probe from the Phase 1 fd-hygiene tests).

### LINKED ISSUE #2456: spawn-process with a bare program name reports a failed exec as exit 127 on OpenBSD, not a file error
`(kaappi process)` documents — in `docs/dev/subprocess.md`, the guide page, and KEP-0022's error taxonomy — that failing to *start* a program raises a file-error-family condition carrying errno, so "not found" and "not allowed" stay distinguishable from anything the program itself did.

That holds on every platform when the argv head is a path. It does **not** hold on OpenBSD when it is a bare name:

```scheme
(import (scheme base) (kaappi process))

;; macOS / Linux / FreeBSD: raises a file error, errno ENOENT
;; OpenBSD:                 returns (values 127 "" "")
(run-process '("kaappi-no-such-program"))
```

Observed on the OpenBSD CI VM in [#2455](https://github.com/kaappi/kaappi/pull/2455) — `tests/scheme/process/process-run.scm`'s spawn-failure assertion failed there and only there, while the absolute-path form in `src/tests_process_run.zig` passed.

## Why

POSIX explicitly permits both: when `posix_spawnp`'s PATH search finds nothing, it is unspecified whether the call returns an error or the child is created and exits 127. OpenBSD's libc does the PATH search in the child, so the parent's `posix_spawnp` succeeds and the exec failure arrives as an ordinary exit status.

## Why it matters

127 is a plausible exit code for a program that really ran. A mistyped command therefore looks like a completed run that happened to fail, rather than a spawn failure — the exact distinction the file-error family exists to preserve, and one a caller cannot recover by inspection. It is a wrong answer that is silent about being wrong.

## Fix

The standard remedy is the one CPython's `subprocess` uses: a `CLOEXEC` error pipe between spawn and exec. The child writes its errno and closes; the parent reads, and if anything arrives it raises the file error instead of returning a status. `process_posix.zig` already owns the file-actions plumbing this would hang off.

Cheaper alternatives, both worse: resolve the PATH search in the parent before spawning (duplicates libc's rules, and races), or treat 127 as a spawn failure (indistinguishable from a program that legitimately exits 127).

## Meanwhile

`tests/scheme/process/process-run.scm` asserts the strict contract with a path, and pins the bare-name case to "an error, or a non-zero status with no output" — so the platform difference is recorded rather than invisible, and the test will tighten when this is fixed.

## Acceptance

- `(run-process '("<bare name that does not exist>"))` raises a file error with ENOENT on every supported POSIX platform, OpenBSD and NetBSD included.
- The bare-name assertion in `process-run.scm` becomes the strict one.

baijum:
Closes #2456
Closes #2424
Closes #2422

Three fd-hygiene fixes around exec, one per issue.

## #2456 — bare-name exec failure is a file error on OpenBSD

OpenBSD's vfork-based userland `posix_spawn` has no channel to report the
child's exec failure, so `(spawn-process '("missing-command"))` "succeeded"
and the child exited 127 — indistinguishable from a program that ran and
failed. That platform now spawns through **fork + execvp with a CLOEXEC
error pipe** (CPython `subprocess`'s mechanism, as the issue prescribes):

- the forked child installs the same redirections the file actions carried
  (dup2 per slot, `/dev/null` opens with addopen's close-then-open
  semantics, staged fd sources), closes every open non-CLOEXEC fd ≥ 3 (the
  fork twin of the actions scan), does the optional `setpgid(0,0)`, resets
  SIGPIPE to default, points `environ` at a caller-supplied `env:` block,
  and `execvp`s;
- on any failure it writes the exec errno (one byte — OpenBSD errnos fit)
  to the pipe and only then `_exit`s 127; a successful exec closes the
  pipe via FD_CLOEXEC, so the parent's read returns **exactly at the exec**;
- the parent raises the errno-carrying file error every other libc reports
  synchronously — and resuming only at the exec restores, for free, the
  signalability-from-return property the NetBSD exec barrier exists for.

Only syscalls run between fork and exec (arena-copied args; OpenBSD's
`execvp` builds PATH candidates in a stack buffer — the shell's own
external-command route). The advisory `access()` pre-flight for
slash-paths is superseded and removed. The bare-name assertions in
`process-run.scm` (pinned leniently for this gap by #2455) and
`tests_process_run.zig` are strict again.

Reproduced on the OpenBSD 7.9 aarch64 reference VM — main: `NO-ERROR
status=127`; this branch: `"cannot spawn process"` for both the bare-name
and slash-path forms. Suites there (fresh `KAAPPI_HOME` homes):
process-test 19/19, process-wait-test 10/10, process-run 18/18,
process-portable 18/18, process-child-script 3/3.

Two child-side details worth flagging from that verification: the
`'stdout: 'null` install originally closed the slot it had just opened
(`open()` lands on the just-freed slot number; the follow-up close undid
the install — `cat: stdout: Bad file descriptor`), fixed by recognizing
`opened == slot` as its own branch; and the wait-test failures that
first looked like a regression were bytecode-cache poisoning from a
lib-less first run against a reused `KAAPPI_HOME` — clean homes are
green.

## #2424 — `fd->port` marks the wrapped descriptor close-on-exec

An FFI descriptor (kaappi-net socket, a foreign pipe) wrapped as a port
kept whatever `FD_CLOEXEC` state it arrived with, so on Linux and the
BSDs every `spawn-process` child inherited it. `makeFdPort` now sets the
flag at the wrap — the ownership boundary — which cannot break a
legitimate pass-through: Kaappi execs only through `spawn-process`, whose
stdio install clears the flag on exactly the child's slots. New unit test
wraps a raw `open("/dev/null")` (no O_CLOEXEC) and asserts the flag
lands. The Linux-only acceptance leg (socket inherited into a spawned
child) is covered by the existing fd-hygiene suites on CI; the Linux
containers are not runnable from this dev machine today.

## #2422 — test/dev fd pairs route through `platform.pipe`

`testing_helpers.makeFdPair`/`makeBidiFdPair`, `bench_reactor`'s
`makePipe`, and the hand-rolled fork/exec plumbing in `test_runner.zig`
and `test_selection.zig` called `pipe(2)`/`socketpair(2)` directly,
bypassing `platform.pipe`'s CLOEXEC discipline (the two extra
`std.c.pipe` sites beyond the issue's table are included; `thottam_proc`
was already Phase-1-audited). No direct `std.c.pipe`/`std.c.socketpair`
call remains outside `platform.zig` in test/dev code.

## Verification

- macOS: all five process suites green (19/10/18/18/3), and
  `tests_process`, `tests_process_run`, `tests_port_io`, `tests_reactor`,
  `tests_scheduler`, `tests_waitforfd` all pass.
- Cross-compile clean for `aarch64-openbsd` (where the fork path is
  analyzed) and native.
- `zig build test` stops at the pre-existing `tests_step` crash (#2447 —
  reproduced on clean main, unrelated to this PR); the per-container runs
  above stand in for it.

Sequencing note: also touches `tests/scheme/process/process-run.scm` and
`CHANGELOG.md`; #2461 (SRFI 231) is disjoint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   code to the copy-out, where a capture re-enters with the destination
   already allocated and hands back the same object (with identical
   contents).
+- **A bare program name that resolves to nothing is a file error on
+  OpenBSD** (#2456) — `(spawn-process '("some-missing-command"))` there
+  "succeeded" and the child exited 127: OpenBSD's vfork-based userland
+  `posix_spawn` has no channel to report the child's exec failure, and
+  POSIX permits exactly that. 127 is also a plausible exit code of a
+  program that really ran, so a mistyped command looked like a completed
+  run. That platform now spawns through `fork` + `execvp` with a CLOEXEC
+  error pipe (CPython `subprocess`'s mechanism): the child writes its
+  exec errno and only then exits, so a miss raises the same
+  errno-carrying file error every other platform's libc reports — and
+  the parent resuming only at the exec restores the signalability
+  property the NetBSD exec barrier exists for. The bare-name assertions
+  in `process-run.scm` and `tests_process_run.zig` are strict again.
+- **`fd->port` leaves the wrapped descriptor inheritable** (#2424) — an
+  FFI descriptor (a kaappi-net socket, a foreign library's pipe) wrapped
+  as a port kept whatever close-on-exec state it arrived with, so on
+  Linux and the BSDs every `spawn-process` child inherited it — a live
+  listening socket could reach the child. The wrap is now the ownership
+  boundary: `makeFdPort` sets `FD_CLOEXEC`, which cannot break a
+  legitimate pass-through (Kaappi execs only through `spawn-process`,
+  whose stdio install clears the flag on exactly the child's slots); a
+  foreign side that wants a child to inherit its descriptor must prepare
+  the fd itself.
+- **Test/dev fd pairs are close-on-exec** (#2422) — the pair constructors
+  in `testing_helpers.zig` and `bench_reactor.zig` called `pipe(2)` and
+  `socketpair(2)` directly, bypassing `platform.pipe`'s CLOEXEC
+  discipline (the hand-rolled fork/exec plumbing in `test_runner.zig`
+  and `test_selection.zig` did the same). These descriptors never
+  intentionally cross an exec, but the fd-hygiene suites had to diff
+  against the parent's own fd set to tolerate them.
 - **Continuations captured under a non-tail `apply` or `call-with-values`
   (#2451)** — resuming one raised `KP3000: continuation cannot resume across a
   returned native call`. Both reached their callee through a native
--- a/src/bench_reactor.zig
+++ b/src/bench_reactor.zig
@@ -33,7 +33,10 @@ fn nowNs() u64 {
 
 fn makePipe() [2]std.c.fd_t {
     var fds: [2]std.c.fd_t = undefined;
-    if (std.c.pipe(&fds) != 0) unreachable;
+    // platform.pipe, not std.c.pipe: same CLOEXEC discipline as every
+    // other test/dev pair constructor (kaappi#2422), and no second call
+    // site to drift if pipe flags ever change.
+    std.debug.assert(platform.pipe(&fds) == 0);
     return fds;
 }
 
--- a/src/platform.zig
+++ b/src/platform.zig
@@ -175,8 +175,14 @@ pub fn pipe(fds: *[2]fd_t) c_int {
 /// Set FD_CLOEXEC on `fd` (the fcntl(2) F_SETFD form; O_CLOEXEC at open is
 /// unavailable for fds other calls hand back -- kqueue(), pipe(2) on
 /// platforms without pipe2). Returns false on failure; callers treat that
-/// as best-effort.
+/// as best-effort. A no-op on Windows: FD_CLOEXEC is a POSIX exec concept
+/// with no analog on CRT fds, and needs none -- process_win confines child
+/// inheritance to an explicit stdio handle list, a stronger guarantee than
+/// any close-by-default scan. No-op, not link-error: makeFdPort calls this
+/// unconditionally at the fd->port wrap (kaappi#2424), and the plain
+/// extern `fcntl` does not exist for Windows targets.
 pub fn setFdCloexec(fd: fd_t) bool {
+    if (comptime is_windows) return false;
     if (comptime is_wasm) return false;
     const F_SETFD: c_int = 2;
     const FD_CLOEXEC: usize = 1;
@@ -186,8 +192,9 @@ pub fn setFdCloexec(fd: fd_t) bool {
 /// fcntl(F_GETFD): the fd-flags word (bit 0 = FD_CLOEXEC), or -1 when `fd`
 /// is not open. The subprocess spawner's close-by-default scan uses it both
 /// as an is-open probe and to skip descriptors exec will close anyway
-/// (KEP-0022).
+/// (KEP-0022). POSIX-only for the same linking reason as setFdCloexec.
 pub fn getFdFlags(fd: fd_t) c_int {
+    if (comptime is_windows) return -1;
     if (comptime is_wasm) return -1;
     const F_GETFD: c_int = 1;
     return fcntlRaw(fd, F_GETFD, 0);
--- a/src/primitives_io.zig
+++ b/src/primitives_io.zig
@@ -946,6 +946,20 @@ fn openBinaryOutputFile(args: []const Value) PrimitiveError!Value {
 /// on it, and unregisters it from the reactor. The caller must not also
 /// close the fd through its own path.
 pub fn makeFdPort(gc: *memory.GC, fd: platform.fd_t, is_input: bool, is_output: bool, name: []const u8) PrimitiveError!Value {
+    // The wrap is the ownership boundary for exec too: a descriptor that
+    // arrives here without FD_CLOEXEC — an FFI socket from kaappi-net, a
+    // foreign library's pipe — would otherwise be inherited by every
+    // spawn-process child on Linux and the BSDs, where close-by-default is
+    // the CLOEXEC audit rather than macOS's wholesale
+    // POSIX_SPAWN_CLOEXEC_DEFAULT (kaappi#2424). Kaappi execs only through
+    // spawn-process, whose stdio install (dup2 into 0..2) clears the flag
+    // on exactly the child's slots, so this cannot break a legitimate
+    // pass-through; a foreign side that genuinely wants a child to inherit
+    // its descriptor must prepare the fd itself. Best-effort on failure: a
+    // descriptor too broken for F_SETFD is the port's problem to report on
+    // first I/O, not a wrap-time error the caller cannot distinguish from
+    // a bad fd number.
+    _ = platform.setFdCloexec(fd);
     const port_val = gc.allocPort(fd, is_input, is_output, name, false) catch return PrimitiveError.OutOfMemory;
     types.toObject(port_val).as(types.Port).is_binary = true;
     return port_val;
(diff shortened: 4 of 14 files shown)


baijum:
Triage of the four red legs on 5b5cfd69:

**windows-cross (aarch64 + x86_64) — real defect, fixed in 6c4df737.**
`lld-link: undefined symbol: fcntl`. #2424's `makeFdPort` change calls
`platform.setFdCloexec` unconditionally, which pulled its plain `extern
fcntl` into the Windows link for the first time — nothing had referenced
it there before (the spawner's fd-flag helpers are POSIX-gated, wasm
already had a no-op guard, Windows did not). Both helpers now no-op for
Windows targets like they already did for wasm: `FD_CLOEXEC` is a POSIX
exec concept, and `process_win` confines child inheritance to an
explicit stdio handle list — a stronger guarantee than any
close-by-default scan — so there is nothing for the flag to do there.
Verified locally before pushing this time: `aarch64-windows`,
`x86_64-windows`, `aarch64-openbsd` and native all build clean (the
macOS-native build I validated the original commit with cannot catch a
Windows-only link error).

**netbsd-test — the kaappi#2446 flake, not this diff.**
`srfi18-cross-thread-wait.scm` failed two assertions
("condition-variable-broadcast! actually woke waiter 1/2") — a lost
wakeup in the cross-thread wait machinery. That suite **passed on this
PR's exact base commit** (run 33510007536, 4aa2bab1), and #2446
documents this family as pre-existing and intermittent on NetBSD (its
own CI occurrence on #2445 was green on rerun). Nothing in this PR
touches condition variables, thread wakeups, or the notifier; the fd
changes are close-on-exec flags and test/dev pair constructors. The
proper fix is the wave-3 diagnosis work #2446 already tracks. Expecting
green on the rerun; if it recurs I will dig further rather than assume.

**riscv64-test — the kaappi#2458 timing flake, again.**
45m29s with the build/test processes still alive at the cap — the
QEMU-emulated leg ran past the 45-minute timeout #2460 raised from 30.
Same shape as before: no test failure, just a slow runner day. If it
recurs on this rerun I will bump the cap again in this PR.

baijum:
Follow-up on the rerun of 6c4df737, including the new openbsd failure it
surfaced — **all four legs of the original report are now accounted for, and
the fix for each is on the branch**:

| leg | verdict |
|---|---|
| windows-cross (aarch64, x86_64) | real defect — `fcntl` link error — **fixed in 6c4df737, green on rerun** |
| netbsd-test | kaappi#2446 cross-thread flake — **green on rerun** (25m33s) |
| riscv64-test | kaappi#2458 QEMU timing flake — **green on rerun** (31m02s) |
| openbsd-test: `process-spawn-2414.sh` | pre-existing test race under load — **fixed in 97e5076c** |

The OpenBSD one looked like it could implicate this PR's fork+exec spawn
path, so I reproduced instead of assuming: the failing assertion is
`process-timeout-stdout` coming back empty where `"partial"` was
expected — the kill landed before the child ever got a scheduling slot to
printf. On the OpenBSD reference VM under four CPU hogs, that shape fails
**11/50 on unmodified main (4aa2bab1)** and 4/50 on this branch — a
pre-existing race in the tests' 250–500 ms deadlines, not a spawn-path
change (the engine's empty drain is the honest answer to a child that
never wrote). 97e5076c gives every timeout-content assertion seconds of
headroom (2 s; 3 s where a grandchild interpreter must start too), which
is what the deadline was only ever supposed to contribute — firing
eventually, with the assertion about recovered content. Re-measured under
the same load: **0/50**. The same tight-deadline pattern existed in
`process-run.scm` and `tests_process_run.zig`, so they are fixed in the
same commit rather than waiting to flake another leg.

baijum:
Rebased onto e8070ab0 (#2461) — the only conflict was the two CHANGELOG insertions at the same `### Fixed` anchor; resolved by keeping both entry sets, #2461's first. All commits otherwise unchanged; re-verified on the rebased head: native + `aarch64-windows` + `aarch64-openbsd` builds clean, the five process suites (19/10/18/18/3), `process-spawn-2414.sh` PASS, `tests_process`/`tests_process_run`/`tests_port_io` green, and the SRFI 231 official suite unchanged at 10936/0. Also confirms the CI event pipeline is delivering again (this push triggered runs).

PR #2463, opened 2026-09-01
Stop treating a zero Windows pipe write quota as would-block

### LINKED ISSUE #2459: Writing more than 4096 bytes to a child's stdin hangs forever on Windows — WriteQuotaAvailable is not a would-block oracle
On Windows, writing past the pipe buffer to a pipe port **hangs the program forever** whenever a fiber scheduler is running. Reproduced on the Windows 11 ARM64 reference VM against `main` + [#2455](https://github.com/kaappi/kaappi/pull/2455).

```scheme
(import (scheme base) (kaappi process))

;; hangs; no error, no timeout, no CPU
(run-process (list "kaappi" "sink.scm") 'input: (make-string 4097 #\i))
```

The threshold is exact — measured by bisection on the VM:

| bytes written | outcome |
|---|---|
| 1000, 4000, **4096** | completes |
| **4097**, 5000, 8192, 16384, 40000 | hangs forever |

4096 is the pipe buffer. Under 4096 the write fits and never has to wait; over it, the second write parks and never wakes.

It needs a scheduler: the same 40 000-byte write from a program with no fibers completes, because a sequential program keeps blocking CRT writes and never enters the emulated non-blocking path.

## Cause

`platform_win_pipe.pipeWrite` treats `WriteQuotaAvailable` — from `NtQueryInformationFile(FilePipeLocalInformation)` — as "can I write without blocking", synthesizing `EAGAIN` when it is 0. `pipePollReady` then re-reads the same field to decide when to wake the parked fiber.

**That field is not a would-block oracle on these handles.** Instrumented output from the VM, on a freshly connected, completely empty child-stdin pipe:

```
[W] fd=3 cfg=2 end=0 outq=4096 wq=0 state=3 inq=4096 avail=0 len=4097
[P] fd=3 cfg=2 end=0 outq=4096 wq=0 state=3 inq=4096 avail=0
[P] fd=3 cfg=2 end=0 outq=4096 wq=0 state=3 inq=4096 avail=0
... forever
```

Nothing had been written yet — `state=3` is `FILE_PIPE_CONNECTED_STATE`, the buffer is empty, and a write would have succeeded immediately. `WriteQuotaAvailable` still reads 0. Whether it reads 0 or 4096 tracks whether the child happens to be parked in a read on the far end, not whether the buffer has room: the same fd reported 4096 in a run where the parent wrote before the child reached its read, and 0 in a run where the parent first waited for the child to announce itself.

So the writer parks on a number that never moves. There is no timeout on a pipe-write park, so the program is wedged: no error, no diagnostic, no CPU.

libuv reads the same field, but only to *size* a `uv_try_write` — never as the sole readiness signal behind a park. That distinction is the bug.

## Fix

Stop using the quota as a refusal. `pipeWrite` still clamps the request to the quota when it is non-zero (a useful short-write hint), but a zero quota falls through to the plain CRT write instead of synthesizing `EAGAIN`, and `pipePollReady` reports writability unconditionally.

The cost is that a pipe write to a genuinely full pipe blocks that OS thread until the reader drains — exactly the graceful degradation the file's query-failure and non-writable-end paths already take, and the behavior every pipe write had before #1608 stage 2. Reads are untouched and keep their gate: `PeekNamedPipe` answers "are there bytes right now" exactly, which is why `pipeRead` parks a fiber correctly and why draining two full pipes concurrently already works.

A readiness signal that would let writes park again needs an overlapped handle (libuv's route), which is a larger change than this hang deserves.

## Scope

Introduced by #1608 stage 2 (emulated non-blocking pipe I/O). Not reachable through sockets, which use `WSAEventSelect`. It became easy to reach with KEP-0022 Phase 4, since `run-process` always runs with a scheduler and always feeds `input:` through a pipe — which is how it was found.

Fixed with a Windows regression test in #2455.

baijum:
Closes #2459

## What

Writing more than 4096 bytes (the pipe size) through a pipe port hung the
program forever on Windows whenever a fiber scheduler was running.
`pipeWrite` treated `WriteQuotaAvailable` — from
`NtQueryInformationFile(FilePipeLocalInformation)` — as "can I write without
blocking" and synthesized `EAGAIN` at zero; `pipePollReady` keyed its wakeup
on the same field. As the issue measured, that field is not a would-block
oracle on these handles: whether it reads 0 or the buffer size tracks whether
the peer happens to be parked in a read on the far end, not whether the
buffer has room (0 on a freshly connected, *empty* child-stdin pipe,
`state=CONNECTED`). The parked writer waited on a number that never moves.

- A zero quota now **falls through to the plain CRT write**, which blocks the
  OS thread until the reader drains — the graceful degradation the
  query-failure path already took, and the pre-#1608-stage-2 behavior. A
  non-zero quota still usefully clamps the request (short-write hint).
- `pipePollReady` reports pipe writability **unconditionally** — a spurious
  wake is safe under park-and-retry, a missed one is not.
- The collector's abandoned-port flush must never block (nothing will ever
  drain an abandoned pipe — the GC would hang), so it gets
  **`pipeWriteNoBlock`**, the old clamp-and-EAGAIN flavor; the fill-to-full
  fd test uses it too, and the write-park fiber test is POSIX-only now.
- The Windows stdin cap in the concurrent-drain test is gone, and a new
  larger-than-pipe-buffer `input:` feed is asserted on every platform
  (with `timeout:` so the pre-fix wedge fails loudly instead of hanging a
  CI leg).

## Evidence (Windows 11 ARM64 reference VM, cross-built aarch64-windows)

- Main: the issue's 4097-byte `input:` feed hangs — killed at 15 s with no
  output. This branch: `status=0`, feed completes.
- `process-run.scm` is **19/19** there, including the uncapped
  concurrent-drain test and the new large-feed test.
- Shape boundary mapped on the VM while designing the test: a **drain-only**
  child fed 16384 bytes completes (`ok 0 0 0`); a small interleaved echoer
  (4000 bytes) round-trips (`ok 0 4000 0`); an interleaved **echoer whose
  echo outgrows its own 4 KiB output pipe while the feed is blocked**
  deadlocks — the documented cost of the blocking fallback, called out in
  the CHANGELOG and the test comment. A parked write with a *truthful*
  readiness signal needs overlapped pipe handles (libuv's route), the
  change #2459 explicitly defers.

## Verification

- macOS: all five process suites green (18/18/10/18/3 — process-test has 18
  here because #2462 adds a 19th on its own branch), `tests_port_io` green,
  native + `aarch64-windows` cross-builds clean.
- `zig build test` stops at the pre-existing `tests_step` crash (#2447,
  reproduced on clean main, unrelated).

Sequencing note: touches `tests/scheme/process/process-run.scm` and
`CHANGELOG.md` in different hunks than #2462; if both land close together
the second needs only a trivial CHANGELOG rebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   and `test_selection.zig` did the same). These descriptors never
   intentionally cross an exec, but the fd-hygiene suites had to diff
   against the parent's own fd set to tolerate them.
+- **Writing more than the pipe buffer to a child's stdin no longer hangs
+  forever on Windows** (#2459) — past 4096 bytes (the pipe size), a feed
+  through `input:` wedged with no error and no CPU whenever a fiber
+  scheduler was running. `pipeWrite` treated
+  `NtQueryInformationFile`'s `WriteQuotaAvailable` as "can I write
+  without blocking" and synthesized `EAGAIN` at zero — but that field
+  tracks whether the peer happens to be parked in a read on the far end,
+  not whether the buffer has room (it reads 0 on a freshly connected,
+  completely empty pipe), so the parked writer waited on a number that
+  never moves. A zero quota now falls through to the plain CRT write,
+  which blocks the OS thread until the reader drains — the graceful
+  degradation the query-failure path already took, and the pre-#1608
+  behavior; a non-zero quota still usefully clamps the request. The
+  collector's abandoned-port flush keeps a never-block flavor
+  (`pipeWriteNoBlock`), since nothing will ever drain an abandoned pipe.
+  Known residual cost, same as before #1608 stage 2: a child that echoes
+  its stdin back while the feed is still blocked can fill its own output
+  pipe and wedge both sides — a parked write with a truthful readiness
+  signal needs overlapped pipe handles, deliberately out of scope. The
+  Windows stdin cap in the concurrent-drain test is gone, and a
+  larger-than-pipe-buffer `input:` feed is asserted on every platform.
 - **Continuations captured under a non-tail `apply` or `call-with-values`
   (#2451)** — resuming one raised `KP3000: continuation cannot resume across a
   returned native call`. Both reached their callee through a native
--- a/src/gc_sweep.zig
+++ b/src/gc_sweep.zig
@@ -456,15 +456,17 @@ pub fn freeObject(gc: *GC, obj: *Object) void {
                         while (start < port.write_buf_len) {
                             // A Windows socket port must send() — CRT _write
                             // cannot operate on a SOCKET handle — and a pipe
-                            // port in emulated non-blocking mode must keep its
-                            // quota gate: plain _write on a full pipe would
-                            // block this sweep forever, exactly the would-block
-                            // this loop promises to drop (#1608).
+                            // port in emulated non-blocking mode needs the
+                            // never-block pipe flavor: pipeWrite's zero-quota
+                            // fall-through is a plain _write that would block
+                            // this sweep forever on a full pipe nobody will
+                            // ever drain, exactly the would-block this loop
+                            // promises to drop (#1608; the split is #2459).
                             const remaining = port.write_buf_len - start;
                             const rc = if (platform.is_windows and port.fd_state.is_socket)
                                 platform.sockSend(port.fd, wb.ptr + start, remaining)
                             else if (platform.is_windows and port.fd_state.is_pipe and port.nonblocking)
-                                platform.pipeWrite(port.fd, wb.ptr + start, remaining)
+                                platform.pipeWriteNoBlock(port.fd, wb.ptr + start, remaining)
                             else
                                 platform.write(port.fd, wb.ptr + start, remaining);
                             if (rc < 0 and platform.errno(rc) == .INTR) continue;
--- a/src/platform.zig
+++ b/src/platform.zig
@@ -15,12 +15,15 @@
 //! event-driven fiber suspension through WSAEventSelect (reactor.zig's
 //! WindowsEventBackend), reading/writing through sockRecv/sockSend.
 //! Pipe ports have no OS-level would-block mode, so they enter *emulated*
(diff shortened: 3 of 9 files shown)


baijum:
Rebased onto `e8070ab0` (#2461) to clear the CHANGELOG.md conflict — same resolution as #2462's rebase: both `### Fixed` entry sets kept, the array-copy entry first, then this PR's #2459 entry. The rebase changed nothing else (`git diff` of the commit content vs the old head `943396b9` is exactly the new base's files).

Verified on the rebased head `32e22099`: native build, `tests_port_io` 34/34, `tests/scheme/process/process-run.scm` 19/19 (fresh `KAAPPI_HOME`), and both Windows cross-builds (`aarch64-windows`, `x86_64-windows`) compile clean.

For the record: the previous head's CI run (33515785025) finished **success** including the rerun netbsd leg — the only red it ever had was the known #2446 flake.

baijum:
Rebased again, this time onto `997a2a25` (#2462's merge) — CHANGELOG.md was the only textual conflict (same resolution: #2462's three `### Fixed` entries first, then this PR's #2459 entry).

The four files both PRs touch (`src/platform.zig`, `src/primitives_io.zig`, `src/tests_port_io.zig`, `tests/scheme/process/process-run.scm`) auto-merged, so I checked the seams rather than trusting that: `makeFdPort` still sets `FD_CLOEXEC` at the wrap (#2424) and is a no-op on Windows where this PR's pipe changes live; `tests_port_io.zig` carries both the #2424 wrap-time test and the #2459 Windows skip; `process-run.scm` has both the deadline bumps and the larger-than-pipe-buffer `input:` feed.

Verified on the rebased head `777a362e`: native build; `tests_port_io` 36/36, `tests_process` 55/67 (12 platform skips), `tests_process_run` 20/20; `process-run.scm` and `process-test.scm` 19/19 each; `tests/scheme/compile/process-spawn-2414.sh` PASS; both Windows cross-builds clean. Also ran the full `process-run.scm` suite on a Windows 11 ARM64 VM with the combined binary: 19/19, exited on its own (the pre-fix hang shape would still trip the 180 s watchdog).

PR #2465, opened 2026-09-01
Discard expected-error stderr in the stepper tests (#2447)

### LINKED ISSUE #2447: tests_step crashes the unit-test binary under podman aarch64 (Virtualization.framework VM)

baijum:
Closes #2447

## Root cause: there is no crash

The unit-test binary never dies in the reported environment. What the
issue read as "the runner dies with no pass/fail summary" is Zig 0.16's
build-runner *warning* display for a **successful** step:

1. The two expected-error stepper tests (`a form error is reported but
   the program still finishes`, `a form that pauses then raises unwinds
   roots cleanly`) legitimately drive the `runFile`-style error path, so
   their KP3002 diagnostics are written to the process's real stderr —
   by design, matching batch output.
2. Under `zig build test` the build runner captures that fd as the run
   step's `result_stderr`. A step with non-empty captured stderr — even
   a fully green one — is rendered with a red ` w` warning marker, the
   stderr itself, and a red `failed command: <argv>` provenance line
   (`lib/compiler/build_runner.zig`: the `" w\n"` branch of
   `printStepFailure`, and the `result_failed_command` echo — set for
   every run step, printed as context).
3. The default summary mode prints **no pass/fail summary at all for a
   successful build**.

Expected stderr + red "failed command" + no summary reads exactly like
a segfault. That is the entire symptom.

## Evidence (kaappi-builder-arm64, podman applehv, unmodified main)

| Run | Result |
|---|---|
| Issue's exact repro (`cp` + `rm -rf .zig-cache zig-out` + filtered build), 4 runs | warning block as described — **exit 0** every time |
| Same under a pty (`script -qec`) | reproduces the reporter's display byte-for-byte (progress lines, red ` w` block, `failed command:`) — **exit 0** |
| `--summary all` on the filtered run | `13 pass (13 total)` |
| Test binary run directly (no `--listen`) | `All 13 tests passed.` |
| **Full** `zig build test --summary all`, one-shot container | `13/13 steps succeeded; 1990/2010 tests passed (20 skipped)` — **exit 0**; the rest of the unit suite is not lost |

So both the "crash" and "the podman leg loses the rest of the unit
suite" do not occur: the suite completes green in the exact repro
environment.

## What changed

`th.QuietStderr` (new in `src/testing_helpers.zig`): redirects fd 2 onto
the null sink for its lifetime and restores it — the same
dup/dup2/openNullSink juggling the fuzz harness already uses for stdout
(`tests_fuzz.evalNormalized`), inert on WASI (no dup, no null device,
kaappi#2153). The two expected-error stepper tests now use it, so their
expected diagnostics no longer land in the build runner's captured
stderr. A green `zig build test -Dtest-filter=tests_step` is now
**completely silent**, in every environment — nothing left to misread.

Unit test included (`QuietStderr routes fd 2 onto the sink and restores
it`): asserts fd 2 writes reach the sink while active and that fd 2 is
restored after. Mutation-checked: breaking the dup2 round-trip fails the
test (an earlier draft that *skipped* on redirect failure was rewritten
to assert, exactly so a broken helper cannot hide behind a skip).

## After the fix, in the container

- Issue's exact repro command: **no output, exit 0** (was: warning block
  + `failed command:` line).
- Full `zig build test --summary all` (one-shot container):
  `13/13 steps succeeded; 1991/2011 tests passed (20 skipped)` — exit 0,
  and the `run test unit-tests` line no longer carries the ` w` warning
  marker.
- macOS worktree: full suite green (1990/2011 passed, 21 skipped),
  `-Dgc-stress=true` filtered runs green, `zig fmt --check` clean.

## Side finding (out of scope, filed separately)

While verifying with a *persistent* container I hit the one test that
can genuinely fail in a podman harness: `tests_process` "process-wait
phase2: group kill reaches the child's own child" polls `kill(gpid, 0)`
until ESRCH, but under a PID 1 that never reaps adopted orphans
(`podman run -d ... sleep infinity`) the killed grandchild stays a
zombie and the poll times out — the group kill itself works (state `Z`,
reparented to PID 1). The issue's one-shot `bash -c` PID 1 reaps, so
this does not occur in the reported environment. Follow-up: #2466

Since the diagnosis is "no Kaappi defect, no VM defect — a display
misread", the issue's own downgrade rule applies: priority lowered from
`high` to `low` (only the diagnostic clarity was wrong).

### DIFF
--- a/src/testing_helpers.zig
+++ b/src/testing_helpers.zig
@@ -330,7 +330,13 @@ pub const QuietStderr = struct {
     }
 
     /// Point fd 2 back at the original stderr. A no-op on an inert receiver.
+    ///
+    /// On WASI the comptime guard must be the first statement of EVERY
+    /// method that names `dup`/`dup2`: semantic analysis of even an
+    /// unreachable-at-runtime call generates the `env::dup2` import, which
+    /// wasmtime then refuses (the wasm-job failure of kaappi#2447).
     pub fn deinit(self: *QuietStderr) void {
+        if (comptime platform.is_wasm) return;
         if (!self.active) return;
         _ = platform.dup2(self.saved, 2);
         _ = platform.close(self.saved);
--- a/src/testing_helpers.zig
+++ b/src/testing_helpers.zig
@@ -1,4 +1,5 @@
 const std = @import("std");
+const testing = std.testing;
 const types = @import("types.zig");
 const platform = @import("platform.zig");
 pub const memory = @import("memory.zig");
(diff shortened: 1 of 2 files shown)


baijum:
Reviewed the diagnosis, the helper, and the two call sites; ran everything locally at fcb60c7e.

**Diagnosis and native change: good.** `zig build test -Dtest-filter=tests_step` is now completely silent and green (13/13), the `QuietStderr` unit test passes, and the `init` / `initOnto` / `deinit` fd juggling is correct (dup2 before the `defer close(devnull)`, saved fd closed on both paths, inert receiver on any failure). The write-up of the build runner's ` w` / `failed command:` rendering matches what the runner actually does.

**Blocking: the `wasm` CI job fails on this PR and was green on the base commit (8f093f5a).**

```
Error: failed to run main module `zig-out/bin/unit-tests.wasm`
Caused by:
    0: failed to instantiate "zig-out/bin/unit-tests.wasm"
    1: unknown import: `env::dup2` has not been defined
```

Cause: `init` and `initOnto` bail with `if (comptime platform.is_wasm)`, so their bodies are never analyzed on wasm32-wasi. `deinit` is guarded only by the *runtime* `if (!self.active) return;`, so its `platform.dup2(self.saved, 2)` is still analyzed, `std.c.dup2` becomes an extern the WASI test module imports from `env`, and wasmtime refuses to instantiate the binary. `testing_helpers.zig` is compiled into the WASI unit-test build (unlike `tests_thottam.zig` / `tests_process.zig`, whose `dup2` calls never reach that target), which is why this shows up here and not for the other `dup2` users.

Fix is one line, verified locally (`zig build test -Dtarget=wasm32-wasi` + `wasmtime run --dir=. --dir=/tmp zig-out/bin/unit-tests.wasm` → `1628 passed; 295 skipped; 0 failed.`):

```zig
    pub fn deinit(self: *QuietStderr) void {
        if (comptime platform.is_wasm) return;
        if (!self.active) return;
```

Worth a word in the doc comment too: on WASI the guard has to be in *every* method that names `dup`/`dup2`, not just the ones that could set `active` — an unreachable-at-runtime call still produces the import.

Not LGTM yet; will approve once the wasm job is green.

baijum:
LGTM.

06f495b7 is the one-line comptime guard in `deinit` plus the doc note on why it has to be in every method that names `dup`/`dup2`. Re-verified locally at that head:

- `zig build test -Dtarget=wasm32-wasi` + `wasmtime run --dir=. --dir=/tmp zig-out/bin/unit-tests.wasm` → `1628 passed; 295 skipped; 0 failed.`
- `zig build test -Dtest-filter=tests_step` → silent, 13/13 green.

CI: every job green including `wasm`; only `netbsd-test` still in progress at time of writing (it is also the one job that failed on the base commit's own `main` run, unrelated to this change).

Issue #15, opened 2026-06-26
jit: box_local/get_box_local/set_box_local miscompiled as plain copies (box leaks into ops)


baijum:
## Summary

Both JIT backends compile the box opcodes (`box_local`, `get_box_local`, `set_box_local`) as plain register copies / no-ops, diverging from the VM's real box semantics (`src/vm_dispatch.zig:916-948`). When a function that boxes a mutable captured local gets JIT-compiled, a **box object leaks where a value is expected**, producing wrong results or a runtime type error.

## Reproduced

```scheme
(define (run)
  (let ((n 0))
    (let ((getter (lambda () n)))   ; capturing n forces boxing
      (set! n (+ n 1))              ; set_box_local
      (+ n (getter)))))             ; get_box_local + closure call; expect 2
(let loop ((i 0) (acc 0))
  (if (< i 200000)
      (loop (+ i 1) (+ acc (run)))
      (begin (display acc) (newline))))
```

```
$ kaappi run.scm           # JIT (default)
run.scm:6: error: expected number, got (0 . )    <-- a box (pair with cdr=VOID) reached `+`
$ kaappi --no-jit run.scm
400000                                            <-- correct
```

## Location

- x86_64: `src/jit_compile_x86_64.zig:214-225` — `get_box_local`/`set_box_local` share the plain-`move` handler; `box_local` only invalidates the cache slot and emits no allocation.
- aarch64: `src/jit_compile_aarch64.zig:326-344` — `box_local` is a literal no-op (`_ = reg;`); `get_box_local`/`set_box_local` emit `emitRegCopy` (a plain register copy).

Both opcodes are **JIT-eligible** (`src/jit.zig:186-187` return operand sizes, not `return false`), so they are compiled rather than side-exited.

## Why it's a bug

The VM (`src/vm_dispatch.zig:916-948`) implements real boxing:
- `box_local reg` → `registers[reg] = allocPair(registers[reg], VOID)` (heap-allocate a box).
- `get_box_local dst, reg` → if `registers[reg]` is a box (`isPair && cdr==VOID`), `dst = car(box)`; else box-on-demand.
- `set_box_local reg, src` → if box, `setCar(box, registers[src])` (mutate the shared cell); else box-on-demand.

The JIT instead leaves the raw value in the register (`box_local` no-op) and then copies the box/value around verbatim. As the repro shows, the boxed pair leaks into subsequent operations.

## Impact

[...]

PR #2468, opened 2026-09-02
Detach SRFI-18 OS threads and back off in cross-thread waits (#2446)

### LINKED ISSUE #2446: NetBSD: srfi18-join-spawn-grandchild-2129.scm hangs intermittently (cross-thread join lost wakeup)

baijum:
Closes #2446.

## What was actually happening

The `blocked-upstream` diagnosis on the issue was wrong. gdb showed a child LWP with a frozen PC and no syscall, which reads as "the kernel never scheduled it" — but a starved runnable thread looks exactly the same from userspace. The kernel's per-LWP view (`ps -s -o pid,lid,lstate,wchan,cpuid,pri,time,pcpu`) showed the hung process with 23 LWPs, 18 runnable, burning three CPUs' worth of time: **17 threads leaving `thread-sleep!` spinning in `memory.spinLock` inside `withdrawCrossThreadWaiter` at priority 25–27, while the holder — preempted inside the `kevent` ring `wakeCrossThreadWaiters` issues under the lock — sat runnable at priority 0 and was never scheduled again.**

Two NetBSD 4BSD-scheduler facts combine into that:

1. **`pthread_join` poisons the joiner's priority.** A nice-0 LWP's priority is `63 − estcpu/2048 − 20`, and the per-tick accumulation clamps `estcpu` at `18 × 2048`, so running can take a thread no lower than 25. But `sched_lwp_collect` — run in `_lwp_wait` on the *joining* LWP — adds the reaped LWP's `estcpu` with no cap, and `sched_lwp_fork` copies the joiner's estimate into every LWP it creates. The interpreter thread joins every middle thread; after three joins it is at priority 0 and so is every thread it spawns from then on — born below every ordinary process, which under CPU contention starves them outright. That is the "child parked at `pthread__create_tramp+0` forever" of the original report. A 40-line C program pins it: join three threads that each spin 1.5 s, then spawn a trivial child under four burners — the child's join took 0.8 s; detach the three instead and it took 0 ms, like the no-join control.
2. **`sched_yield` only requeues behind LWPs of the same priority**, so a crowd spinning — or yielding — on a lock held by a preempted thread never lets it back in.

Linux CFS and macOS give every runnable thread a share regardless of history, which is why four platforms only ever paid latency for the same code.

## The fix

- **Detach SRFI-18 OS threads at spawn** (a detached LWP is freed at exit with no `estcpu` transfer) and have `thread-join!` wait on a per-spawn exit flag (`Fiber.os_exit`) that the child raises after its outermost defer — covering exactly what `pthread_join` covered, descendant drain included.
- **`platform.spinBackoff`** — 32 spins, 64 yields, then sleeps doubling from 32 µs to 1 ms — replaces every pure spin in a cross-thread wait: `memory.spinLock`, `VM.stopForCollection`, `markLiveChildRoots`' wait for children to stop, `GlobalsRwLock`, and the new reap wait. The sleep is what takes a waiter off the run queue. The uncontended path is unchanged; no sleep ever happens on a healthy machine.

The backoff alone took the hang from 5/8 to 0/10 in a first sample and then hung 4 of the next 8 — with the process at *zero* CPU and every LWP at priority 0, which is what exposed the join-inheritance mechanism.

## Measurements

NetBSD 10.1 aarch64 reference VM (4 vCPUs), 40 s bound per run, `srfi18-join-spawn-grandchild-2129.scm`:

| build | 4 burners | 8 burners |
|-------|-----------|-----------|
| unmodified `main` | 5/8 hung | 4/10 hung |
| backoff only | 0/10, then 4/8 hung | — |
| backoff + detach | **0/20 hung** (and 0/10 of `srfi18-terminate-native-wait-1982.scm`, the other test the CI kill hit) | **0/30 hung** |

## Tests

- `tests_srfi18.zig` "kaappi#2446: contenders on a held spin lock sleep instead of burning CPU" — fails against the old lock (four contenders burned 518 ms of CPU across a 120 ms hold), passes with the backoff.
- `tests_srfi18.zig` "kaappi#2446: thread-join! returns only after the child's whole epilogue" — pins the exit-flag ordering the reap relies on.
- Host: full unit suite green (1992 passed, 21 skipped), all 33 `srfi18*.scm`, a `-Dgc-stress=true` subset over the thread/join/child tests, `zig build wasm` and the `aarch64-netbsd` cross-build.

## Docs

Postmortem `docs/dev/postmortems/2026-09-02-netbsd-spinlock-starvation.md` (with the wrong-diagnosis analysis and the `ps -s` recipe), a new scheduler section in `docs/dev/netbsd.md`, a fifth load-bearing bullet in `docs/dev/fibers-and-reactor.md`, and lessons-learned #15.

## Follow-ups (not in this PR)

- #2471 — `sched_lwp_collect`'s missing `ESTCPULIM` turns out to be fixed upstream already (trunk rev 1.48, 2026-03-01; pulled up to netbsd-11, **not** netbsd-10), so 10.x releases keep needing the detach; the issue now tracks the netbsd-10 pull-up.
- #2470 — `wakeCrossThreadWaiters` still rings under the registry lock; with the backoff that costs waiters latency, never liveness. Moving the ring outside the lock needs per-entry retain/release on a path that must not allocate — a separate trade-off.
- #2472 — the one remaining bare spin, WinSock init on Windows.
- #2473 — `thread-start!` leaves the fiber rooted in `extra_roots` when the spawn fails (pre-existing, noticed on the way).

### DIFF
--- a/docs/dev/fibers-and-reactor.md
+++ b/docs/dev/fibers-and-reactor.md
@@ -108,7 +108,10 @@ Five things load-bearing enough to be worth knowing before touching this:
   `platform.spinBackoff`: a short pure spin, then `sched_yield`, then
   `nanosleep` doubling from 32 µs to 1 ms — and that includes the registry
   lock itself (`memory.spinLock`), the stop-the-world handshake on both sides
-  (`VM.stopForCollection`, `markLiveChildRoots`) and the globals RW lock. A
+  (`VM.stopForCollection`, `markLiveChildRoots`), the globals RW lock, and
+  `thread-join!`'s wait for a detached OS thread's exit flag (`reapOsThread`,
+  which replaced `pthread_join` — the joining LWP would otherwise inherit the
+  dead thread's CPU-usage estimate on NetBSD). A
   pure spin only works while the thread being waited for is *running*; once
   the kernel preempts it, the spinners compete with it for CPU, and on a
   priority-decay scheduler they win outright. That was kaappi#2446: on
--- a/docs/dev/fibers-and-reactor.md
+++ b/docs/dev/fibers-and-reactor.md
@@ -67,7 +67,7 @@ of these waits instead re-checked their own state every millisecond
 (`sleepNs(CROSS_THREAD_POLL_NS)` and the `pollCapNs` caps), which is why a
 `(thread-sleep! 60)` on a child thread used to wake 60,000 times.
 
-Four things load-bearing enough to be worth knowing before touching this:
+Five things load-bearing enough to be worth knowing before touching this:
 
 - **Enrolment is unconditional, and must stay that way.** Gating it on
   `crossThreadWaitPossible()` at entry looks like an obvious saving and is a
@@ -104,6 +104,22 @@ Four things load-bearing enough to be worth knowing before touching this:
   An atomic length gate read outside the lock is a store-buffering pattern
   against the enroller — both sides can miss the other's store — and with no
   poll cap left as a backstop that is a hang, not a latency blip.
+- **Every wait on another OS thread backs off to a sleep** —
+  `platform.spinBackoff`: a short pure spin, then `sched_yield`, then
+  `nanosleep` doubling from 32 µs to 1 ms — and that includes the registry
+  lock itself (`memory.spinLock`), the stop-the-world handshake on both sides
+  (`VM.stopForCollection`, `markLiveChildRoots`) and the globals RW lock. A
+  pure spin only works while the thread being waited for is *running*; once
+  the kernel preempts it, the spinners compete with it for CPU, and on a
+  priority-decay scheduler they win outright. That was kaappi#2446: on
+  NetBSD's 4BSD scheduler, 17 threads leaving `thread-sleep!` spun on the
+  registry lock at priority 25–27 while the holder — preempted inside the
+  `kevent` ring, at priority 0 after the work it had done — stayed runnable
+  and never ran again. Yielding does not help (it requeues behind LWPs of the
+  *same* priority); only a sleep takes the spinner off the run queue. The
+  holder's ring is still issued under the lock (one `notify()` per parked
+  thread); with the backoff that costs the ringed threads latency, never
+  liveness.
 
 A "never" deadline is a real one and reaches the backends: SRFI-18 reads
 `+inf.0` as "never times out", `saturatedNsFromSeconds` turns that into
--- a/docs/dev/netbsd.md
+++ b/docs/dev/netbsd.md
@@ -123,8 +123,11 @@ the call is a comptime no-op there and everywhere else.
 ## Scheduler: a preempted lock holder starves under a pure spin
 
 NetBSD's default scheduler is 4BSD (`sys/kern/sched_4bsd.c`): a user LWP's
-priority is `63 − estcpu/2048`, `estcpu` grows by 1024 on every clock tick
-the LWP is running and decays only slowly (90% in `5 × loadavg` seconds),
+priority is `63 − estcpu/2048 − p_nice` (`p_nice` is the nice value offset by
+20, so 20 for an ordinary process), `estcpu` grows by 1024 on every clock
+tick the LWP is running, is clamped at `18 × 2048` by that per-tick path —
+so running alone floors a nice-0 thread at 25 — and decays only slowly
+(90% in `5 × loadavg` seconds),
 and a new LWP **inherits its spawner's `estcpu`**. Each CPU always runs the
 highest-priority runnable LWP, and `sched_yield` requeues an LWP behind the
 others *of its own priority* — never behind a lower one.
--- a/docs/dev/netbsd.md
+++ b/docs/dev/netbsd.md
@@ -120,6 +120,65 @@ the call is a comptime no-op there and everywhere else.
   is a test-harness concern only — the interpreter itself runs fine in the
   default limits.
 
+## Scheduler: a preempted lock holder starves under a pure spin
+
+NetBSD's default scheduler is 4BSD (`sys/kern/sched_4bsd.c`): a user LWP's
+priority is `63 − estcpu/2048`, `estcpu` grows by 1024 on every clock tick
+the LWP is running and decays only slowly (90% in `5 × loadavg` seconds),
+and a new LWP **inherits its spawner's `estcpu`**. Each CPU always runs the
+highest-priority runnable LWP, and `sched_yield` requeues an LWP behind the
+others *of its own priority* — never behind a lower one.
+
+That combination turns any userspace pure spin into a hang under CPU
+contention (kaappi#2446): a thread that has been doing real work sits at
+priority 0; the moment the kernel preempts it while it holds a lock — say,
+inside the `kevent` ring `wakeCrossThreadWaiters` issues per parked thread —
+every thread that then arrives at that lock spins at a *higher* priority
+(fresh from a sleep, or a new LWP whose spawner had been idle), and with
+more spinners than CPUs the holder is never scheduled again. Observed as
+`srfi18-join-spawn-grandchild-2129.scm` running until killed at ~290% CPU:
+17 LWPs in `memory.spinLock` from `withdrawCrossThreadWaiter`, one LWP
+runnable (`R`, no wchan) at priority 0 inside `_sys___kevent50`. Linux CFS
+and macOS never reproduce it — a fair scheduler always gives the holder its
+share, so the same spin is a latency cost there, not a liveness one.
+
+Two kaappi-side rules follow, both of which hold everywhere but were only
+ever *needed* here:
+
+* **Never `pthread_join` a `thread-start!`ed thread.** `_lwp_wait` makes the
+  joining LWP absorb the dead LWP's `estcpu` with no cap
+  (`sched_lwp_collect` lacks the `ESTCPULIM` every other path applies), and
+  `sched_lwp_fork` copies the estimate into every LWP the joiner creates. A
+  nice-0 thread cannot get below priority 25 by running, but three joins of
+  busy threads put the joiner — and all its future children — at 0, below
+  every ordinary process, where CPU contention starves them outright. SRFI-18
+  threads are therefore detached at spawn and joined through a per-spawn
+  exit flag (`Fiber.os_exit`, raised after the child's outermost defer).
+* **Every cross-thread wait loop goes through `platform.spinBackoff`** (spin,
+  then yield, then sleep 32 µs doubling to 1 ms): `memory.spinLock`, the GC
+  stop-the-world handshake on both sides, the globals RW lock and the reap
+  wait above. A sleep is what a yield is not — it takes the waiter off the
+  run queue, so a preempted holder at the same priority gets its turn. A
+  bare `spinLoopHint` loop reintroduces the convoy on this platform.
+
+**Diagnosing a hang here.** gdb alone misleads: a starved LWP shows a frozen
+PC and no syscall, which reads as "the OS never scheduled this thread" (the
+original diagnosis of kaappi#2446). The kernel-side view is what settles
+it — per-LWP state, wait channel, priority and whether the process's CPU
+time is still climbing:
+
+```bash
+ps -s -o pid,lid,lstate,wchan,cpuid,pri,time,pcpu -p <pid>   # twice, a few seconds apart
+gdb -batch -p <pid> -ex 'info threads' -ex 'thread apply all bt 8'
+```
+
+`R` with no wchan and `pri 0` while other LWPs sit at 25+ and `TIME` grows by
+seconds per second is starvation by spinners; `S` with a wchan is a kernel
+sleep; a lone `R` LWP in an otherwise idle process would be the scheduler's
+fault. Measure under load (`sh -c 'while :; do :; done' &` × ncpu) — the
+hang needs more runnable LWPs than CPUs — and kill the burners by PID
+afterwards; orphaned ones poison every later measurement.
+
 ## Self-exe path
 
 NetBSD has `KERN_PROC_PATHNAME`, but filed under the `KERN_PROC_ARGS`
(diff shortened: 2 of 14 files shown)


baijum (commented):
Reviewed the threading changes in detail and verified what I could locally (macOS aarch64, ReleaseSafe build).

**Verified correct:**

- **Defer ordering in `threadEntryFn`**: the exit-flag defer is declared first, so it runs last — after the `live_child_threads` decrement, `markExited`, and the descendant drain. Everything `pthread_join` used to cover is still covered, and the decrement-before-flag ordering is exactly what the new epilogue test pins (join returns ⇒ `hasLiveChildThreads()` is false).
- **`OsThreadExit` lifetime**: two refs (the child's very last action, the joiner after observing `exited`); the `.release` store / `.acquire` load pairing gives the joiner the same happens-before over `target.result` that `pthread_join` provided. The never-joined leak is documented and consistent with the pre-existing child GC/VM leak.
- **No stragglers**: no remaining `join()` on `os_thread` anywhere (the joins in `memory.zig`/`main.zig` are unrelated threads), and every reader of `os_thread` (`threadJoinFn`, `threadTerminateFn`) only null-tests it, so the marker-only semantics hold.
- **No cross-heap aliasing**: `.fiber` is refused by `gc_deep_copy` and the globals route is owner-checked (`checkThreadOwner`), so the raw `os_exit` pointer can't be duplicated into another heap.
- **GC latency**: `markLiveChildRoots` clears `collection_stop` without waiting for children to resume, so the backoff in `stopForCollection` costs at most ~1 ms of child-side resume latency (and only when the parent's mark outlasts the spin+yield window) — it does not lengthen the parent's collection.
- **Failure paths** in `threadStartImpl` unwind `live_child_threads`/`live_descendants`/envelope/flag correctly on both the flag-allocation and spawn OOM paths.

**Local runs:** both new tests pass; the full `srfi18` unit filter passes (70 tests); the spin-lock CPU test ×5 stable at ~130 ms/run, comfortably under the 60 ms CPU bound; all 70 `srfi18` tests also pass under `-Dgc-stress=true`; `zig fmt --check` and markdownlint are clean. I also confirmed `CLOCK_PROCESS_CPUTIME_ID` works on macOS (returns 0), so the CPU measurement is real here.

Four non-blocking nits inline. One more outside the diff: `src/fiber_wait.zig:362` still says the joiner "hangs forever in reapOsThread's thread.join() (#1982)" — the reasoning holds for the exit-flag wait, but the comment names a call that no longer exists; worth a follow-up touch-up.

The two-part fix is well argued, the postmortem is excellent (especially the wrong-diagnosis section), and the regression tests pin real properties rather than trying to reproduce the scheduler-specific hang. Non-blocking from my side.

src/tests_srfi18.zig
```diff
@@ -1000,3 +1001,80 @@ test "a channel-send park that errors out leaves the fiber runnable (#2433)" {
...
     , "channel-send");
 }
+
+// ---------------------------------------------------------------------------
+// kaappi#2446: waiting on a held spin lock must not burn CPU
+// ---------------------------------------------------------------------------
+
+fn processCpuNs() u64 {
+    var ts: std.c.timespec = undefined;
+    _ = std.c.clock_gettime(.PROCESS_CPUTIME_ID, &ts);
```
baijum: nit: the `clock_gettime` return value is discarded. On every current POSIX target `CLOCK_PROCESS_CPUTIME_ID` works (verified it returns 0 on macOS), but if a future platform rejects it, `ts` stays `undefined` and the 60 ms comparison below runs on stack garbage — flaky rather than failing loudly. Propagating the error (`if (std.c.clock_gettime(...) != 0) return error.SkipZigTest;`) makes that failure mode impossible.

src/platform.zig
```diff
@@ -955,6 +955,53 @@ pub fn dirIterDestroy(it: *DirIter) void {
 // sleep
 // ---------------------------------------------------------------------------
 
+/// How many iterations `spinBackoff` spends in its pure-spin phase, then in
+/// its yield phase, before it starts sleeping. `pub` so a test can drive a
```
baijum: nit: the stated reason for `pub` — "so a test can drive a wait loop past both" — isn't realized: no test references these constants (the CPU-burn test drives past both phases with a real 120 ms hold instead). Either wire a test to them or drop the claim (and the `pub`), so the comment doesn't promise a consumer that doesn't exist.

docs/dev/netbsd.md
```diff
@@ -120,6 +120,68 @@ the call is a comptime no-op there and everywhere else.
   is a test-harness concern only — the interpreter itself runs fine in the
   default limits.
 
+## Scheduler: a preempted lock holder starves under a pure spin
+
+NetBSD's default scheduler is 4BSD (`sys/kern/sched_4bsd.c`): a user LWP's
+priority is `63 − estcpu/2048`, `estcpu` grows by 1024 on every clock tick
```
baijum: nit (doc): this short form omits the nice offset that the later derivation on this page needs. "A nice-0 thread cannot get below priority 25 by running" follows from `63 − estcpu/2048 − 20` with `estcpu` clamped at `18 × 2048` (63 − 18 − 20 = 25); `63 − estcpu/2048` alone floors at 45, which makes the 25-floor claim two sections down look wrong. The postmortem's second-round section has the full form — using it here too would keep the page self-consistent.

docs/dev/postmortems/2026-09-02-netbsd-spinlock-starvation.md
```diff
@@ -0,0 +1,226 @@
...
+burners competing for 4 CPUs, never picked it again. The test's 12-iteration
+loop spawns a middle that returns without joining its grandchild, so a dozen
+sleeping grandchildren accumulate and all wake within the same few
+milliseconds, which is why this test and not another.
+
+## Why NetBSD and only NetBSD
+
+NetBSD's default scheduler is 4BSD (`sys/kern/sched_4bsd.c`):
+
+- A user LWP's priority is `63 − estcpu/2048`. `estcpu` grows by 1024 on
```
baijum: nit: same formula mismatch as in netbsd.md — this bullet's `63 − estcpu/2048` can't reach the priority-25 floor that this same document derives later ("For a nice-0 process `pri = 63 − estcpu/2048 − 20" with the `18 × 2048` clamp → 25). Harmonizing on the full form in both spots keeps the postmortem self-consistent.

docs/dev/fibers-and-reactor.md
```diff
@@ -104,6 +104,22 @@ Four things load-bearing enough to be worth knowing before touching this:
   An atomic length gate read outside the lock is a store-buffering pattern
   against the enroller — both sides can miss the other's store — and with no
   poll cap left as a backstop that is a hang, not a latency blip.
+- **Every wait on another OS thread backs off to a sleep** —
+  `platform.spinBackoff`: a short pure spin, then `sched_yield`, then
+  `nanosleep` doubling from 32 µs to 1 ms — and that includes the registry
+  lock itself (`memory.spinLock`), the stop-the-world handshake on both sides
+  (`VM.stopForCollection`, `markLiveChildRoots`) and the globals RW lock. A
```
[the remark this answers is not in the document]
baijum: Done in a4021f6d — the inventory now names `reapOsThread`'s exit-flag wait alongside the lock, GC and globals waits.

docs/dev/postmortems/2026-09-02-netbsd-spinlock-starvation.md
```diff
@@ -0,0 +1,219 @@
+# NetBSD: `thread-join!` hangs — uncapped priority inheritance through `pthread_join`, amplified by pure spin locks
+
+## Status
+
+**Fixed** (2026-09-02, kaappi#2446), in two parts. (1) SRFI-18 OS threads
```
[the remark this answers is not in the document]
baijum: Done in a4021f6d — Status now reads "commit 96934bde via kaappi#2468, closing kaappi#2446".

docs/dev/postmortems/2026-09-02-netbsd-spinlock-starvation.md
```diff
@@ -0,0 +1,219 @@
...
+ PID   LID STAT WCHAN   CPUID PRI    TIME %CPU
+3808 13115 R    -           0  26 2:04.32  290
+3808 28593 S    nanoslp     1  64 2:04.32  290
+3808 26365 I    lwpwait     0  64 2:04.32  290
+3808 12010 O    -           2  25 2:04.32  290
+3808 20262 R    -           0   0 2:04.32  290
+3808 18430 R    -           0   0 2:04.32  290     <- kevent, holds the lock
+3808  3455 R    -           0  26 2:04.32  290
+   ... 13 more R/O LWPs at PRI 25-27 ...
+3808  3808 I    lwpwait     2  85 2:04.32  290
```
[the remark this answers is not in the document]
baijum: Done in a4021f6d — the elision line now accounts for the two rows it hid (one `Z`, one more `lwpwait`), so 8 shown + 13 + 2 = 23.

docs/dev/postmortems/2026-09-02-netbsd-spinlock-starvation.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Lessons
+
+- **A frozen PC in gdb is not evidence about the scheduler.** It is
+  compatible with "the OS lost this thread", "this thread is runnable and
+  starved", and "this thread is asleep in a page fault". Only the kernel's
+  per-thread state distinguishes them. On NetBSD:
+  `ps -s -o pid,lid,lstate,wchan,cpuid,pri,time,pcpu -p <pid>`, sampled
+  twice — climbing `TIME` in a "hung" process means it is spinning, not
+  stuck. Linux: `/proc/<pid>/task/*/stat` (state, priority, utime);
```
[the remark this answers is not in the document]
baijum: Done in a4021f6d — the lesson now says process `TIME` only shows that *something* burns CPU, and classifies the hang from the per-LWP columns (many R/O rows with no wchan while TIME climbs = spin; an R row that never goes O = starvation; S + wchan = kernel sleep), noting that under a fair scheduler the thread under investigation can be blocked while a sibling burns.

src/tests_srfi18.zig
```diff
@@ -1000,3 +1001,80 @@ test "a channel-send park that errors out leaves the fiber runnable (#2433)" {
...
+            _ = self.acquired.fetchAdd(1, .acq_rel);
+            memory.spinUnlock(&self.lock);
+        }
+    };
+    var ctx: Ctx = .{};
+    const hold_ns: u64 = 120_000_000;
+    memory.spinLock(&ctx.lock);
+    const cpu_before = processCpuNs();
+    var threads: [4]std.Thread = undefined;
+    for (&threads) |*t| t.* = try std.Thread.spawn(.{}, Ctx.contend, .{&ctx});
```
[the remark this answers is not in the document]
baijum: Done in a4021f6d — added the readiness counter exactly as proposed; `cpu_before` is now taken only after all four workers have reached the lock call.

src/tests_srfi18.zig
baijum: Done in a4021f6d — `processCpuNs` now returns `!u64` and skips the test (`error.SkipZigTest`) if `clock_gettime` rejects the clock, so an unsupported platform fails loudly rather than comparing against an undefined `ts`.

src/platform.zig
baijum: Done in a4021f6d — wired the consumer rather than dropping the claim: tests_platform.zig "spinBackoff spins, then yields, then sleeps" drives the schedule against both constants (the spin+yield phases must finish in under a sleep quantum; the first iteration past them must sleep ≥ 32 µs; the capped one ≥ 1 ms). The comment now names that test.

docs/dev/netbsd.md
baijum: Done in a4021f6d — netbsd.md now states the full form `63 − estcpu/2048 − p_nice` with the 20 offset and the `18 × 2048` clamp, so the 25 floor derives on the page.

docs/dev/postmortems/2026-09-02-netbsd-spinlock-starvation.md
baijum: Done in a4021f6d — same full form in the postmortem's 4BSD bullet, so the second-round derivation no longer contradicts it.

PR #2478, opened 2026-09-02
Treat a zombie grandchild as success in the group-kill test

### LINKED ISSUE #2466: tests_process group-kill test fails under a container PID 1 that never reaps adopted orphans
## Symptom

`tests_process.test.process-wait phase2: group kill reaches the child's own child` (src/tests_process.zig) fails in a podman container started with a non-reaping PID 1, e.g.

```
podman run -d --name X localhost/kaappi-builder-arm64 sleep infinity
podman exec X bash -c 'cd /workspace && zig build test -Dtest-filter="group kill reaches"'
```

→ `1 fail`, `error.GrandchildSurvivedGroupKill` after the 10 s deadline.

Found while verifying #2447 in a persistent container (see PR #2465 for the full story). It does **not** fail in the one-shot `podman run --rm ... bash -c '...'` shape (whose bash PID 1 reaps adopted orphans), nor on macOS native, ubuntu-arm CI, or the BSD reference VMs.

## Root cause

The test kills the child's process group and then polls `kill(gpid, 0)` until it returns ESRCH, with the comment "once init reaps it, signaling it reports ESRCH". The group kill itself works — verified manually that the grandchild (`sleep 30`) is dead, state `Z`, reparented to PID 1. But under a PID 1 that never calls `wait()` (`sleep infinity` is a common debugging harness; other minimal inits behave the same), the zombie is never reaped, `kill(zombie_pid, 0)` keeps succeeding, and the poll times out.

So the product behaviour is correct; only the test's "init will reap" assumption is environment-sensitive.

## Possible directions

- Treat "grandchild is a zombie" as success on Linux by checking `/proc/<gpid>/stat` state `Z` when `kill(gpid, 0)` keeps succeeding (a dead-but-unreaped grandchild is exactly what a successful group kill looks like there).
- Or probe once up front whether PID 1 reaps (spawn+kill a scratch orphan) and skip (77-style, `error.SkipZigTest`) when it does not.

## Priority

`priority: low` per the rubric: the behaviour is right and only the test's environment assumption is wrong; the failure is loud and narrow (needs a specific container harness shape).

baijum:
Closes #2466

## Problem

`tests_process.test."process-wait phase2: group kill reaches the child's own child"` polls `kill(gpid, 0)` until ESRCH, assuming PID 1 reaps the reparented grandchild. Under a container PID 1 that never calls `wait()` (e.g. a podman container started with `sleep infinity`), the zombie is never reaped, `kill(zombie_pid, 0)` keeps succeeding, and the poll times out after 10 s with `error.GrandchildSurvivedGroupKill` — even though the group kill itself worked (verified manually: the grandchild is dead, state `Z`, reparented to PID 1).

Product behaviour is correct; only the test's "init will reap" assumption is environment-sensitive.

## Fix

Implements the issue's first direction: on Linux (`builtin.os.tag == .linux`), when `kill(gpid, 0)` keeps succeeding, read `/proc/<gpid>/stat` and treat state `Z` (zombie) as success — a dead-but-unreaped grandchild is exactly what a successful group kill looks like there. Other platforms keep the existing ESRCH poll, and the total deadline behaviour for a genuinely-alive grandchild is unchanged.

The `comm` field is parenthesized and may contain spaces or a closing paren, so the state letter is parsed after the *last* `)\' rather than by naive whitespace splitting.

## Regression test

The bug is in the test itself, so this test correction **is** the fix — no additional regression test is needed and no product code changed.

## Validation

- `zig build test -Dtest-filter="group kill reaches"` — passes
- `zig build test` (full unit suite) — passes

### DIFF
--- a/src/tests_process.zig
+++ b/src/tests_process.zig
@@ -1230,10 +1230,12 @@ fn grandchildIsZombie(pid: i32) bool {
     if (comptime builtin.os.tag != .linux) return false;
     var path_buf: [32]u8 = undefined;
     const path = std.fmt.bufPrint(&path_buf, "/proc/{d}/stat", .{pid}) catch return false;
-    var file = std.fs.openFileAbsolute(path, .{}) catch return false;
-    defer file.close();
+    // std.fs file APIs were reworked in 0.16; raw posix (as in platform.zig)
+    // compiles identically on every target.
+    const fd = std.posix.openat(std.posix.AT.FDCWD, path, .{ .ACCMODE = .RDONLY, .CLOEXEC = true }, 0) catch return false;
+    defer _ = std.posix.system.close(fd);
     var buf: [256]u8 = undefined;
-    const n = file.read(&buf) catch return false;
+    const n = std.posix.read(fd, &buf) catch return false;
     // Fields: pid (comm) state ... — skip past the last ')' so a comm
     // containing ')' or spaces can't fool the parse.
     var i: usize = 0;
--- a/src/tests_process.zig
+++ b/src/tests_process.zig
@@ -1208,11 +1208,39 @@ test "process-wait phase2: group kill reaches the child's own child" {
     // The grandchild dies with the group; once init reaps it, signaling it
     // reports ESRCH. Bounded: a surviving grandchild fails the test at the
     // deadline (and its 30-second sleep ends it soon after regardless).
+    //
+    // The ESRCH poll assumes an init that reaps the reparented zombie. Under
+    // a container PID 1 that never wait()s (e.g. `sleep infinity`), the zombie
+    // persists and kill(gpid, 0) keeps succeeding forever (#2466). On Linux a
+    // dead-but-unreaped grandchild shows state `Z` in /proc/<pid>/stat —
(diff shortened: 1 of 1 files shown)


baijum:
CI fix-up: all eight failing legs (gc-stress, ppc64le/riscv64/s390x-test, all four ubuntu test matrix entries) were the same compile error — `std.fs.openFileAbsolute` does not exist in Zig 0.16, introduced by 8ac51632 in `grandchildIsZombie`. macOS passed only because the function body is comptime-gated to Linux, so the bad line was never analyzed on darwin — which is also why the author's local `zig build test` was green.

Fix (6feecdfd): read `/proc/<pid>/stat` via raw `std.posix.openat` / `std.posix.read` / `std.posix.system.close`, the same pattern `platform.zig` already uses; compiles identically on every target. No behavior change — this is a build fix to a test-only helper.

Verified: `zig build test` (macOS host) green, `zig build test -Dtarget=` x86_64-linux, riscv64-linux, s390x-linux, powerpc64le-linux all compile the unit-test binary cleanly (they failed to compile before), `zig fmt --check` clean.

PR #2485, opened 2026-09-02
Stop the mark worklist from leaving freed-but-dirty pages behind every full collection (#2464)

### LINKED ISSUE #2464: Peak RSS runs ~65x the live heap on allocation-churning loops; freed memory is not returned
## Kaappi version

main (4aa2bab1), also reproduced on 0.25-era builds. macOS aarch64,
ReleaseSafe.

## Description

A bulk loop that churns transient objects (the SRFI 231 `array-copy` fill
loop is a convenient driver) leaves peak RSS at roughly **65x the live
heap**, and the process never returns the difference:

```bash
mkdir -p /tmp/kh && cat > /tmp/mem-n.scm <<'SCM'
(import (scheme base) (scheme write) (srfi 231))
(define d (string->number (car (cdr (command-line)))))
(define a (make-specialized-array (make-interval (vector d d)) u8-storage-class 0 #f))
(define b (array-copy a))
(write (array-ref b (- d 1) (- d 1))) (newline)
SCM
KAAPPI_HOME=/tmp/kh kaappi /tmp/mem-n.scm 1000   # warm the sld cache once
for i in 1 2 3; do KAAPPI_HOME=/tmp/kh /usr/bin/time -l kaappi /tmp/mem-n.scm 1000 2>&1 | grep maximum; done
```

Measured (4 runs, stable): **669-671 MB peak RSS**, every run. With
`--gc-stats` on the same run: `Heap size: ~10.6 MB (peak ~10.7 MB)`,
`Live objects: ~14.7k (peak ~25k)`, `Freed: ~18M objects, ~641 MB` —
i.e. the GC's own accounting says the heap never exceeds ~11 MB while the
process holds ~670 MB. The churn is ~16 pairs + 2 closures per element
in library bulk loops (`interval-for-each` + per-element `apply` of
getter/setter closures); a positional-lambda loop over the same domain
allocates ~4 pairs/element, so most of it is the `apply`-with-rest-list
shape inside library code.

Scaling: 250x250 (62.5k elems) ~13.8 MB, 500x500 (250k) ~13.8 MB,
1000x1000 (1M) ~670 MB, 1414x1414 (2M) ~1.56 GB — a cliff between 250k
and 1M elements, then linear.

## Why it matters

Any long-running program whose steady state is small but whose bulk
operations churn (which is exactly what SRFI 231 bulk operations, `map`
over big lists, etc. do) presents an RSS footprint tens of times its
live set, with no way to shrink it — on memory-constrained hosts that is
an OOM the program's own data does not justify. The per-element
allocation in the library `apply` shape is the multiplier feeding it.

## Suggested directions (not prescriptions)

- Return freed pages after (full) collections — `free`/`madvise(
  MADV_DONTNEED)` on now-empty allocator arenas, or a slab/region
  allocator that can shrink.
- Look at why the `apply`-of-a-rest-list shape inside library code
  allocates ~16 pairs + 2 closures per element where the equivalent
  positional call allocates ~4 — if the tail-apply zero-copy path is not
  reaching library-compiled code, fixing that shrinks the churn itself.

## Does it panic or raise a Scheme error?

Neither — it is resource behavior, silently.

baijum:
Closes #2464.

## Root cause

The issue's repro (SRFI 231 `array-copy` of a 1000×1000 u8 array) left peak RSS at ~670 MB against a ~10.6 MB live heap. Tracing the allocations showed the memory was neither the churn nor a leak in Kaappi's own accounting — it was the **mark worklist**:

1. `array-copy` keeps a 1M-element scratch vector of *fixnums* live through the whole copy (`views.sld`'s accumulate-then-materialize shape). Every full collection marks it, and `markValueInner`'s container arms pushed **every element — immediates included** onto the worklist, even though the drain discards an immediate at its `isPointer` check. Worklist peak: ~1.3M entries ≈ 10.4 MB, grown through libc `realloc`.
2. After the drain, `markValue` released the whole buffer (`max_retained` was 64K entries), so the *next* full collection regrew the identical realloc chain — 155 times in this run.
3. macOS libmalloc does not decommit freed large blocks in this pattern. Verified with a standalone C reproduction (60 rounds of realloc-grow-to-10.4MB/free/touch): **691 MB peak RSS**, matching Kaappi's. Each chain left ~10 MB of resident-dirty pages behind: 155 chains ≈ 640 MB ≈ the observed 65×.

vmmap mid-run confirmed it: 59 `MALLOC_LARGE (empty)` regions of exactly 10.0 MB each — freed by malloc, still resident — while `MALLOC_TINY` (the actual pair/closure churn) held only 96 KB.

## The fix (two changes, both in the mark path)

- **`wlPush` skips immediates at every one of the 56 worklist push sites** (all container arms in `markValueInner` plus the port visitor). Semantically identical — the drain's `markValueInner` returns at its `isPointer` check, so an immediate on the worklist could only ever be popped and discarded. A wide container of immediates now needs no worklist at all.
- **The post-drain retention floor rises from 64K to 1M entries (512 KB → 8 MB)**, so a genuinely pointer-wide frontier no longer frees and regrows its buffer every full collection. A frontier *wider* than the floor releases the grown buffer but re-reserves the 8 MB floor, so the next regrowth starts at 8 MB rather than zero; the above-floor portion is still rebuilt each wide collection, and on mallocs that keep freed large blocks dirty (macOS) those tops stay resident — bounded retention over unbounded churn, with a decommitting backing allocator as the follow-up #2464 itself suggests.

## Measurements (macOS aarch64, ReleaseSafe)

| workload | before | after |
|---|---|---|
| issue repro, 1000×1000 (1M elems) | **701 MB** | **19.3 MB** |
| 250×250 | 13.8 MB | 9.8 MB |
| 500×500 | 13.8 MB | 11.8 MB |
| 1414×1414 (2M) | 1.56 GB | 29.3 MB |
| 2000×2000 (4M) | (cliff, ~3 GB by scaling) | 49.4 MB |
| 300k live pairs + 3M-pair churn | 48.3 MB | 46.4 MB |
| 2M-element live vector of pairs + 5M-pair churn | 212 MB | 173 MB |

The cliff between 500k and 1M elements is gone; RSS now tracks the live heap. The repro also runs **~15% faster** (total mark time 1648 ms → 432 ms — a million skipped push/pop pairs per full collection were real work).

## Tests

- New `src/tests_gc_worklist.zig`, registered in `vm_tests.zig`. Both discriminating tests verified to **fail on the pre-fix code and pass after**:
  - a 40k-element immediate vector leaves the worklist untouched (40k appends grew capacity to ~52k — under the old 64K floor, so pre-fix the growth was retained and visible; sizes above the old floor were cleared to zero pre-fix and would pass vacuously);
  - a 70k-pair vector — above the old floor, below the new one — retains its buffer across collections (pre-fix: cleared every time).
  - a third test pins that pointer elements still reach the worklist (the skip must never widen to pointers).
- `zig build test`: 2001/2022 pass (21 pre-existing skips).
- `zig build test -Dgc-stress=true`: 2005/2022 pass (17 skips).
- `bash tests/scheme/run-all.sh`: **2140 pass, 0 fail** (an earlier concurrent run showed 2 failures that were load flakes from sharing the machine with the gc-stress suite; clean serial run is fully green).
- All six SRFI 231 suites pass individually (the heaviest `array-copy` users).

## Not addressed here (follow-up material, per the issue's own framing)

The second suggested direction — the `apply`-of-a-rest-list shape inside library bulk loops allocating ~16 pairs + 2 closures per element — is allocation *volume*, not retention. This PR fixes the retention symptom (RSS ≈ live heap). Reducing the churn itself is a separate optimization with its own trade-offs.

### DIFF
--- a/src/gc_collect.zig
+++ b/src/gc_collect.zig
@@ -807,7 +807,7 @@ fn markObjectContents(gc: *GC, obj: *Object) void {
 //
 // One enumeration of a port's Values, three different things to do with
 // each. This is what replaced the hand-kept triplicate: markPortValues
-// (recursive, for markObjectContents), an inline worklist.append list in
+// (recursive, for markObjectContents), an inline wlPush list in
 // markValueInner, and an inline isYoungPointer list in referencesYoung.
 // Each visitor keeps its own arm's exact previous semantics -- in
 // particular markValueInner still appends to the caller's worklist rather
@@ -825,16 +825,35 @@ fn markVisitor(gc: *GC, v: Value) bool {
     return false;
 }
 
-/// markValueInner: push every referent onto the caller's worklist.
+/// markValueInner: push every pointer referent onto the caller's worklist
+/// (immediates skipped — see wlPush).
 const WorklistVisitor = struct {
     gc: *GC,
     worklist: *std.ArrayList(Value),
 
     fn visit(self: WorklistVisitor, v: Value) bool {
-        self.worklist.append(self.gc.allocator, v) catch @panic("GC mark: worklist OOM");
+        wlPush(self.gc, self.worklist, v);
         return false;
     }
 };
+
+/// Push one referent onto the mark worklist, skipping immediates. The
+/// drain's markValueInner returns at its `isPointer` check, so an immediate
+/// on the worklist could only ever be popped and discarded — pushing one is
+/// pure transient growth. #2464: a wide container of immediates (the
+/// array-copy scratch vector of fixnums that motivated the issue) peaked
+/// the worklist at the container's full width every full collection, and
+/// the post-drain buffer release made each collection regrow through the
+/// same realloc chain — the backing libc malloc never decommits those freed
+/// large blocks, so peak RSS accumulated ~10 MB of resident-dirty pages per
+/// full collection (~65x the live heap) until the program ended. The pair
+/// arm's car push keeps its own car_is_ptr guard (it also steers the cdr
+/// tail-iteration); wlPush's check is what covers every container arm's
+/// fields.
+inline fn wlPush(gc: *GC, worklist: *std.ArrayList(Value), v: Value) void {
+    if (!types.isPointer(v)) return;
+    worklist.append(gc.allocator, v) catch @panic("GC mark: worklist OOM");
+}
 
 /// Record a reachable ephemeron for the post-mark weak fixpoint without tracing
 /// its key or value — those are decided by processWeakRefs.
@@ -955,8 +974,18 @@ pub fn markValue(gc: *GC, v: Value) void {
     gc.marking = false;
 
     // Cap retained capacity so one pathologically wide object (e.g. a
-    // 10M-element vector) doesn't keep ~80 MB allocated forever.
-    const max_retained = 64 * 1024;
+    // 10M-element vector of pointers) doesn't keep ~80 MB allocated
+    // forever. The floor is generous (#2464): releasing the buffer after
+    // every collection made each full mark regrow it through the same
+    // realloc chain, and the backing libc malloc does not decommit the
+    // freed large blocks — on macOS each freed ~10 MB top block stayed
+    // resident-dirty, so peak RSS accumulated roughly one chain per full
+    // collection (observed ~65x the live heap) even though every byte was
+    // freed. Retaining up to 8 MB (1M entries) of a buffer whose growth
+    // was driven by live pointer-reachability costs at most a fraction of
+    // the structure that caused it; only the excess past the floor is
+    // released.
+    const max_retained = 1024 * 1024;
     if (gc.mark_worklist.capacity > max_retained)
         gc.mark_worklist.clearAndFree(gc.allocator);
 }
@@ -1008,7 +1037,7 @@ fn markValueInner(gc: *GC, v: Value, worklist: *std.ArrayList(Value)) void {
                 // Push car onto worklist instead of recursing -- this is
                 // the key change that prevents stack overflow on deep
                 // structures like (((((...)))))).
-                worklist.append(gc.allocator, car) catch @panic("GC mark: worklist OOM");
+                wlPush(gc, worklist, car);
                 cur = cdr;
             } else if (cdr_is_ptr) {
                 cur = cdr;
@@ -1026,82 +1055,82 @@ fn markValueInner(gc: *GC, v: Value, worklist: *std.ArrayList(Value)) void {
         .pair => unreachable,
         .closure => {
             const cls = obj.as(Closure);
-            worklist.append(gc.allocator, types.makePointer(&cls.func.header)) catch @panic("GC mark: worklist OOM");
+            wlPush(gc, worklist, types.makePointer(&cls.func.header));
             for (cls.upvalues) |uv| {
-                worklist.append(gc.allocator, uv) catch @panic("GC mark: worklist OOM");
+                wlPush(gc, worklist, uv);
             }
         },
         .function => {
             const func = obj.as(Function);
             for (func.constants.items) |c| {
-                worklist.append(gc.allocator, c) catch @panic("GC mark: worklist OOM");
+                wlPush(gc, worklist, c);
             }
             if (func.global_cache) |cache| {
-                for (cache) |c| worklist.append(gc.allocator, c) catch @panic("GC mark: worklist OOM");
+                for (cache) |c| wlPush(gc, worklist, c);
             }
-            worklist.append(gc.allocator, func.env_val) catch @panic("GC mark: worklist OOM");
+            wlPush(gc, worklist, func.env_val);
         },
         .transformer => {
             const tx = obj.as(Transformer);
             for (tx.literals) |lit| {
-                worklist.append(gc.allocator, lit) catch @panic("GC mark: worklist OOM");
+                wlPush(gc, worklist, lit);
             }
             for (tx.patterns) |pat| {
-                worklist.append(gc.allocator, pat) catch @panic("GC mark: worklist OOM");
+                wlPush(gc, worklist, pat);
             }
             for (tx.templates) |tmpl| {
-                worklist.append(gc.allocator, tmpl) catch @panic("GC mark: worklist OOM");
+                wlPush(gc, worklist, tmpl);
             }
-            worklist.append(gc.allocator, tx.def_env_val) catch @panic("GC mark: worklist OOM");
-            worklist.append(gc.allocator, tx.proc) catch @panic("GC mark: worklist OOM");
+            wlPush(gc, worklist, tx.def_env_val);
+            wlPush(gc, worklist, tx.proc);
             for (tx.let_syntax_peer_vals) |pv| {
-                worklist.append(gc.allocator, pv) catch @panic("GC mark: worklist OOM");
+                wlPush(gc, worklist, pv);
             }
         },
         .error_object => {
             const err = obj.as(types.ErrorObject);
-            worklist.append(gc.allocator, err.message) catch @panic("GC mark: worklist OOM");
-            worklist.append(gc.allocator, err.irritants) catch @panic("GC mark: worklist OOM");
-            worklist.append(gc.allocator, err.uncaught_reason) catch @panic("GC mark: worklist OOM");
+            wlPush(gc, worklist, err.message);
+            wlPush(gc, worklist, err.irritants);
+            wlPush(gc, worklist, err.uncaught_reason);
         },
         .record_type => {
             const rt = obj.as(RecordType);
             if (rt.parent) |p| {
-                worklist.append(gc.allocator, types.makePointer(&p.header)) catch @panic("GC mark: worklist OOM");
+                wlPush(gc, worklist, types.makePointer(&p.header));
             }
         },
         .record_instance => {
             const ri = obj.as(RecordInstance);
-            worklist.append(gc.allocator, types.makePointer(&ri.record_type.header)) catch @panic("GC mark: worklist OOM");
+            wlPush(gc, worklist, types.makePointer(&ri.record_type.header));
             for (ri.fields) |field| {
-                worklist.append(gc.allocator, field) catch @panic("GC mark: worklist OOM");
+                wlPush(gc, worklist, field);
             }
         },
         .continuation => {
             const cont = obj.as(Continuation);
             for (cont.registers) |reg| {
-                worklist.append(gc.allocator, reg) catch @panic("GC mark: worklist OOM");
+                wlPush(gc, worklist, reg);
             }
             for (cont.frames[0..cont.frame_count]) |frame| {
                 if (frame.closure) |cls| {
-                    worklist.append(gc.allocator, types.makePointer(&cls.header)) catch @panic("GC mark: worklist OOM");
+                    wlPush(gc, worklist, types.makePointer(&cls.header));
                 }
                 if (frame.native) |nf| {
-                    worklist.append(gc.allocator, types.makePointer(&nf.header)) catch @panic("GC mark: worklist OOM");
+                    wlPush(gc, worklist, types.makePointer(&nf.header));
                 }
             }
             for (cont.handlers[0..cont.handler_count]) |handler| {
-                worklist.append(gc.allocator, handler.handler) catch @panic("GC mark: worklist OOM");
+                wlPush(gc, worklist, handler.handler);
             }
             for (cont.wind_records[0..cont.wind_count]) |wr| {
-                worklist.append(gc.allocator, wr.before) catch @panic("GC mark: worklist OOM");
-                worklist.append(gc.allocator, wr.after) catch @panic("GC mark: worklist OOM");
+                wlPush(gc, worklist, wr.before);
+                wlPush(gc, worklist, wr.after);
             }
         },
         .multiple_values => {
             const mv = obj.as(MultipleValues);
             for (mv.values) |val| {
-                worklist.append(gc.allocator, val) catch @panic("GC mark: worklist OOM");
+                wlPush(gc, worklist, val);
             }
         },
         .vector => {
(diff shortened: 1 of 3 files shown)


src/gc_collect.zig
```diff
@@ -955,8 +974,18 @@ pub fn markValue(gc: *GC, v: Value) void {
...
+    // freed large blocks — on macOS each freed ~10 MB top block stayed
+    // resident-dirty, so peak RSS accumulated roughly one chain per full
+    // collection (observed ~65x the live heap) even though every byte was
+    // freed. Retaining up to 8 MB (1M entries) of a buffer whose growth
+    // was driven by live pointer-reachability costs at most a fraction of
+    // the structure that caused it; only the excess past the floor is
+    // released.
+    const max_retained = 1024 * 1024;
     if (gc.mark_worklist.capacity > max_retained)
         gc.mark_worklist.clearAndFree(gc.allocator);
```
baijum, line 994: The comment above (and the PR body: "only the excess past the floor is released") describes a shrink-to-floor, but `clearAndFree` drops the *whole* buffer, so any frontier wider than 1M pointers is back to the pre-fix behaviour, just 16x further out: every full collection regrows from zero through the entire realloc chain (0 → 8 MB → … → 80 MB for the 10M-vector case), and by the reasoning in this very comment each of those chains leaves its freed large-block tops resident-dirty on macOS. So the cliff hasn't been removed for that case, it's been moved — and each chain is ~8x bigger than the one #2464 measured.

Two ways to make the code match the comment:

```zig
if (gc.mark_worklist.capacity > max_retained) {
    gc.mark_worklist.clearAndFree(gc.allocator);
    gc.mark_worklist.ensureTotalCapacityPrecise(gc.allocator, max_retained) catch {};
}
```

(one alloc of a fixed 8 MB block that is then retained, so the next chain starts at 8 MB instead of 0 — the `catch {}` is fine because a failed re-reserve just means the next collection grows from zero as today), or `shrinkAndFreePrecise` via `remap` if you'd rather keep the same block — but that API also sets `items.len`, so it needs a `clearRetainingCapacity` after it.

If you'd rather keep the release-all behaviour for the >1M case on purpose, that's defensible too (bounded retention over bounded churn), but then the comment and PR description should say so, and it would be worth one measurement of the 10M-pointer shape to confirm libmalloc doesn't do to 80 MB chains what it did to 10 MB ones.

src/gc_collect.zig
```diff
@@ -1134,33 +1163,33 @@ fn markValueInner(gc: *GC, v: Value, worklist: *std.ArrayList(Value)) void {
         ),
         .hash_table => {
             const ht = obj.as(HashTable);
-            if (ht.equiv_fn != 0) worklist.append(gc.allocator, ht.equiv_fn) catch @panic("GC mark: worklist OOM");
-            if (ht.hash_fn != 0) worklist.append(gc.allocator, ht.hash_fn) catch @panic("GC mark: worklist OOM");
+            if (ht.equiv_fn != 0) wlPush(gc, worklist, ht.equiv_fn);
+            if (ht.hash_fn != 0) wlPush(gc, worklist, ht.hash_fn);
```
baijum: Optional nit: these two `!= 0` guards are now subsumed by `wlPush`'s `isPointer` check — a raw `0` Value is the flonum `+0.0`, which `isPointer` rejects. Harmless to keep, but if you want the hash-table arm to read like every other arm they can go.

src/tests_gc_worklist.zig
```diff
@@ -0,0 +1,116 @@
...
+
+// Sizes chosen against the pre-fix release floor (64K entries), so each
+// test's discriminating assertion actually observes the retained buffer:
+// 40k pushes grew capacity to ~52k — *under* the old floor, hence retained
+// and visible; 70k grew past it, so the old code cleared the buffer to zero
+// after every collection. Under gc-stress the collection-per-allocation
+// makes a 70k-pair build quadratic, so only the single-allocation
+// immediate-only case runs there.
+const immediate_len = 40_000;
+const wide_len = if (build_options.gc_stress) 50_000 else 70_000;
```
baijum: The `gc_stress` branch here is dead: the only user of `wide_len` is the retention test, which `return error.SkipZigTest`s under gc-stress before touching it. Either drop the conditional (`const wide_len = 70_000;`) or drop the skip and let the 50k variant run — as written it reads as though the stress build exercises a smaller case, and it doesn't.

Relatedly, the sentence above ("so only the single-allocation immediate-only case runs there") is inaccurate: the pointer-elements test also runs under gc-stress. That's fine — with `enabled = false` its 1000 `allocPair`s never collect — but the comment should say two of the three run, not one.

baijum (commented):
Reviewed the diff against `main`, and independently re-ran the new tests both ways: all three pass on the branch, and the two discriminating ones (`immediates never grow the worklist`, `retained below the 1M-entry floor`) fail with `main`'s `gc_collect.zig` swapped in, so the regression coverage is real.

The `wlPush` change is semantically identical to the old pushes — the drain returns at `isPointer` before anything else, and nothing outside the drain inspects worklist contents — so the 56-site rewrite is safe. The re-entrant `markFiberState → gc.markValue` path goes through the same `markValueInner`, so it inherits the skip too.

One substantive point (the retention branch doesn't do what its comment and the PR body say it does) and a few test nits inline. Nothing blocking.

src/tests_gc_worklist.zig
```diff
@@ -0,0 +1,116 @@
...
+    var items: [n]types.Value = undefined;
+    for (&items, 0..) |*slot, i| {
+        slot.* = try gc.allocPair(types.makeFixnum(@intCast(i)), types.NIL);
+    }
+    var vec = try gc.allocVector(&items);
+    gc.pushRoot(&vec);
+    defer gc.popRoot();
+
+    gc.collect();
+    try std.testing.expect(gc.mark_worklist.capacity >= n);
```
baijum: Nit: the vector arm tail-calls the *last* element rather than pushing it, so the mark path only ever pushes `n - 1` entries. `>= n` holds today because `growCapacity` adds ~50% slack on top of the minimum, but `>= n - 1` is the bound the code actually guarantees, and the comment on the line above ("pushes one entry per pair") is off by one for the same reason.

src/gc_collect.zig
baijum: Done — implemented your first option in a219b8ac: `clearAndFree` + `ensureTotalCapacityPrecise(max_retained)` with `catch {}`, so the floor block survives and the next wide collection's chain starts at 8 MB. The comment now states plainly that the above-floor portion is rebuilt each wide collection and stays dirty on mallocs that don't decommit freed large blocks, and names the decommitting-backing-allocator follow-up.

I also ran the measurement you asked for — a 2M-element live vector of pairs (frontier ≈ 2M entries > 1M) plus 5M-pair churn:

- pre-fix (main): 212 MB
- committed (`clearAndFree` only): 172.7 MB
- floor-retained: 172.8 MB

The two post-fix variants are identical in practice, and the reason is instructive: `--gc-stats` shows the run reaches only **6 collections** total — the threshold scales to 4× the ~4.7M live objects (~19M entries), so the churn never triggers another one. Sustained many-full-collections churn against a >1M-wide live structure costs roughly 32× live in allocations per full collection; the 65× blowup only reached its cadence because the immediates bug let a 15k-object heap wear a 1M-wide worklist. So the moved cliff you describe is real but self-limiting to reach; the floor re-reserve is still worth having (it's free) and the comment no longer overclaims.

src/gc_collect.zig
baijum: Dropped in a219b8ac, with a one-line comment noting why no guard is needed (raw `0` is the flonum +0.0, which `wlPush`'s `isPointer` check rejects) so the next reader doesn't reintroduce it as an "obvious" fix.

src/tests_gc_worklist.zig
baijum: Dropped the conditional (`const wide_len = 70_000;`) and rewrote the comment: under gc-stress **two** of the three tests run — the immediate-only case (single allocation) and the pointer-elements case (its 1000 `allocPair`s run with `enabled = false`, so nothing collects); only the retention test skips.

src/tests_gc_worklist.zig
baijum: Good catch on the tail-call — the bound is now `>= n - 1` with a comment saying exactly why (the vector arm pushes every element except the last, which is iterated directly), and the immediate-case comment above was fixed to match the same off-by-one ("all but the tail-iterated last").

src/tests_gc_worklist.zig
```diff
@@ -0,0 +1,116 @@
...
+//!
+//! Both are asserted through `mark_worklist.capacity`, which persists across
+//! collections (see also the older "mark worklist retains capacity across
+//! collections" test in memory.zig — that one pins capacity *persistence*,
+//! these pin what may grow it and when it is released).
+
+const std = @import("std");
+const memory = @import("memory.zig");
+const types = @import("types.zig");
+const build_options = @import("build_options");
```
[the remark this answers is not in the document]
baijum: Declining this one. `testing_helpers.zig`'s `TestContext` boots a fully-bootstrapped VM (`makeTestVM` registers ~700 primitives, installs the bootstrap, registers every standard library) and its `expectEval*` helpers evaluate Scheme source — none of which can express what these tests assert: `gc.mark_worklist.capacity`, a field of the GC struct itself, with no VM involved. The sibling convention agrees: the GC-struct tests in `tests_gc_tracing.zig` and in `memory.zig` (including the pre-existing "mark worklist retains capacity across collections" test these three sit next to) all use raw `memory.GC.init(std.testing.allocator)` + `std.testing.expect*`; `th` appears in those files only for tests that genuinely need a VM. Spinning up a VM here would just slow the suite (especially under `-Dgc-stress=true`, where VM construction collects per allocation) without changing a single assertion.

src/gc_collect.zig
```diff
@@ -955,10 +974,28 @@ pub fn markValue(gc: *GC, v: Value) void {
...
+    // next collection's regrowth starts at 8 MB rather than zero — the
+    // above-floor portion is still rebuilt (and, on mallocs that keep
+    // freed large blocks dirty, still accrues); a decommitting backing
+    // allocator is the follow-up #2464 suggests.
+    const max_retained = 1024 * 1024;
+    if (gc.mark_worklist.capacity > max_retained) {
         gc.mark_worklist.clearAndFree(gc.allocator);
+        // A failed re-reserve just means the next collection grows from
+        // zero, exactly the pre-fix behaviour — not worth failing over.
+        gc.mark_worklist.ensureTotalCapacityPrecise(gc.allocator, max_retained) catch {};
```
baijum, line 997: This new branch (over-floor → release → re-reserve exactly the floor) is the one behaviour change in a219b8ac, and none of the three tests reach it — the widest one stops at 70k. It can be pinned cheaply: a *single* pair referenced 1M+64 times from one `allocVectorFill` makes the vector arm push every slot before the drain pops any, so the worklist peaks above the floor with one heap object behind it. I ran this against your commit and it passes in both the default build and under `-Dgc-stress=true` (it's one allocation, so no quadratic stress cost):

```zig
test "#2464: an over-floor frontier re-reserves exactly the floor" {
    var gc = memory.GC.init(std.testing.allocator);
    defer gc.deinit();
    gc.enabled = false;

    const floor = 1024 * 1024;
    var pair = try gc.allocPair(types.makeFixnum(1), types.NIL);
    gc.pushRoot(&pair);
    defer gc.popRoot();
    // One pair, referenced floor+64 times: the vector arm pushes every slot
    // before the drain pops any, so the worklist peaks above the floor with
    // a single heap object behind it.
    var vec = try gc.allocVectorFill(floor + 64, pair);
    gc.pushRoot(&vec);
    defer gc.popRoot();

    gc.collect();
    try std.testing.expectEqual(@as(usize, floor), gc.mark_worklist.capacity);
    gc.collect();
    try std.testing.expectEqual(@as(usize, floor), gc.mark_worklist.capacity);
    try std.testing.expect(types.isPair(types.toVector(vec).data[floor]));
}
```

Without it, a future edit that drops the `ensureTotalCapacityPrecise` line (or changes `max_retained`) would pass the suite unchanged. Non-blocking.

baijum (commented):
Re-reviewed a219b8ac. All four earlier comments are addressed, the comment now matches what the code does, and the `#2464` follow-up it points at (a decommitting/shrinking backing allocator) is indeed the direction the issue proposes. Re-ran the three worklist tests plus the older `memory.zig` retention test on this commit: green in both the default build and `-Dgc-stress=true` (the wide test skips there as intended). One optional suggestion inline for the branch this commit added, which no test currently reaches.

src/gc_collect.zig
baijum: Added verbatim in 7b2d368e — good catch that a219b8ac's one behaviour change had no coverage; the widest prior test stopped at 70k, 15x under the floor.

Verification on my side matches yours: the test passes in both the default build and `-Dgc-stress=true` (the single-allocation shape keeps it quadratic-free), and it **fails** with the `ensureTotalCapacityPrecise` line deleted, which is the regression it exists to catch. I hardcoded `floor = 1024 * 1024` rather than referencing a shared constant on purpose, per your reasoning — moving `max_retained` should also fail the suite, not silently re-point the pin. Full unit suite: 2002/2023 (21 pre-existing skips).

baijum:
The `test (macos-latest, ReleaseSafe)` failure is not this PR's change — filed as #2489 with the full evidence. Summary:

- The unit binary is being **externally SIGKILLed** mid-suite: `terminated with signal KILL`, a **different victim test each time** across three attempts of 7b2d368e (`tests_process` phase2, `tests_process_run` pipe-drain, `tests_process_run` grandchild-kill). A content-triggered kill would be deterministic, not a different victim per attempt.
- The same signature hit **two other branches earlier today** (14:18Z `tests_endian` — a pure byte-order test with no children/signals/pipes — and 15:24Z #2481), while this PR's parent commit a219b8ac passed the same leg green at 16:14Z. The only difference between green-parent and killed-head is one test that changes the suite's peak RSS by nothing (MaxRSS 4G measured with and without), runs in ~100 ms, and passes locally 25/25 in isolation plus two full-suite runs on the same macOS/ReleaseSafe configuration.
- Every other leg of this commit is green — ubuntu Debug/ReleaseSafe/ReleaseFast, gc-stress, gc-stress-scheme, all BSDs, WASM, Windows.

#2484 (merged this morning, for #2481's 40:04Z cancellation) already documents this leg's runners being slow today. #2489 lists diagnostics for the next recurrence. I've kicked one more rerun of the failed job; if the fleet is still unhappy it will keep dying regardless of content, in which case this leg needs the #2489 follow-up rather than a code change here.

Issue #2469, opened 2026-09-02
Make the builtin-superinstruction gate a run-time decision: #2467's whole-unit scan still bakes the builtin in for load, eval, the REPL, and macro-materialized redefinitions


baijum:
**Kaappi version**

PR #2467 head (`746c5f4d`), on top of `8f093f5a`.

**Platform**

macOS aarch64, ReleaseSafe

**Description**

#2467 closes #2457 with the issue's *interim* option: drivers that know the whole compilation unit (file runs, stdin, `kaappi compile`, `kaappi check`, the native backend) pre-scan the unit's literal top-level `define` / `define-values` / `set!` targets, and `globalBindingStillGenuine` declines the `apply` / `call-with-values` / `eval` / `call/cc` / `call-with-current-continuation` superinstructions for those names everywhere in the unit. #2457's *honest* fix — making the decision at run time, the way `car` and `map` already resolve — was left as the recorded follow-up. This is that record, so the residual routes stay tracked once #2457 is closed.

Every entry point without whole-unit knowledge, and every redefinition the structure-only scan cannot see, still bakes the builtin into a body compiled before the redefinition runs. Verified on the #2467 head:

| route | program | prints | expected (R7RS 5.3.1) |
|---|---|---|---|
| `load` — the loaded file's own use-before-define | `(load "b.scm")` where `b.scm` is `(define (nt) (apply + (list 1 2)))` `(define (apply f xs) 'loaded-user)` `(write (nt))` | `3` | `loaded-user` |
| `eval` | `(eval '(begin (define (nt2) (apply + (list 1 2))) (define (apply f xs) 'eval-user)) (interaction-environment))` `(write (nt2))` | `3` | `eval-user` |
| macro-materialized `set!` with a use-site name | `(define (nt) (let ((v (apply + (list 1 2)))) v))` `(define-syntax redef (syntax-rules () ((_ n) (set! n (lambda (f xs) 'macro-set)))))` `(redef apply)` `(write (list (nt) (apply + '(1 2))))` | `(3 macro-set)` | `(macro-set macro-set)` |
| REPL | the #2457 repro typed form by form | builtin | user's binding |

[...]

PR #2486, opened 2026-09-02
Share the globals generation counter with child-thread VMs (#2483)

### LINKED ISSUE #2483: Child-thread VMs never see the root's global_version bumps, so their per-function global caches serve stale bindings
**Kaappi version**

`main` at 62705249; surfaced by CodeRabbit on PR #2481 (#2469).

**Platform**

All.

**Description**

`VM.global_version` is a per-VM field (`src/vm.zig`, default 0). `VM.initForThread` shares the root's `globals` map by pointer (kaappi#2129) but does not share, copy, or ever re-sync the version counter. Every per-function global cache — filled by `get_global`, `call_global`/`tail_call_global`, and since #2469 `guard_builtin` (all through `vm_dispatch_helpers.lookupGlobalCached`) — is invalidated only by *its own* VM's `global_version` moving.

So a child thread that has cached a global keeps serving the cached value after the root (or any other thread) rebinds that global:

```scheme
(import (scheme base) (srfi 18))
(define (f) 'first)
  (let loop ((i 0) (last #f))
    (if (< i 2000000) (loop (+ i 1) (f)) last)))))
(thread-start! t)
(define (f) 'second)          ; root bumps ONLY its own global_version
(thread-join! t)              ; child's cache still holds the first `f`
```

The child's function cached `f` under its own version 0; the root's `define_global` bumped the root's counter; nothing tells the child. With #2469 the same staleness also keeps a `guard_builtin` on its fast path after the root redefines `apply` — the cached value is still the pristine primitive, so the compare succeeds. Before #2469 the analogous case was "calls a stale procedure"; now it is "silently keeps the superinstruction". Same root cause, same fix.

`docs/dev/thread-value-sharing.md` already calls mutating shared state through a global a live hazard (kaappi#1924), which is why this is `medium` rather than `high`: the path is narrow and documented as unsupported, but the answer is silently stale rather than an error.

**Suggested resolution**

Make the generation a property of the shared globals state rather than of the VM: either store it alongside the globals map (bump under the globals write lock, read by every VM chained to that map), or have the root's bump walk the child VMs. Add a threaded regression test: cache a guarded builtin in a child, rebind it in the root, invoke again in the child.

**Priority rationale**

`priority: medium` per `docs/dev/github-issues.md`: silently wrong against a documented guarantee, but on a path the concurrency docs already mark as a hazard.

baijum:
## Summary

`VM.global_version` was a per-VM field that `VM.initForThread` neither copied nor shared, while the globals map itself is shared by pointer (kaappi#2129). Every per-function global cache (`get_global`, `call_global`/`tail_call_global`, and since #2469 `guard_builtin`, all via `lookupGlobalCached`) was invalidated only by its **own** VM's rebindings. So a child thread that had cached a global kept calling the old binding after the root redefined it, and a child's `guard_builtin` silently stayed on its superinstruction fast path after the root redefined `apply`.

**Fix:** the generation counter now lives on the `GlobalsRwLock` — the one object every VM chained to the map already holds by pointer — as an atomic `version`, read through `VM.globalVersion()` and advanced through `VM.bumpGlobalVersion()`. The per-VM field is gone.

Two ordering rules make the shared counter sound across threads (documented on the two methods and in `docs/dev/thread-value-sharing.md`):

- **Writers** (`set_global`, `define_global`, `defineGlobal`, import merges) bump inside the locked region that made the store, and re-validate their own cache slot with the value **that bump returned** — a fresh load could already reflect another thread's later rebinding and would bless a slot the map no longer matches.
- **Readers** (`lookupGlobalCached`, `call_global`, `tail_call_global`) snapshot the version **once, before** the map read, and stamp the snapshot. Previously the fresh-cache paths stamped a load taken *after* the read.

## Test plan

- [x] New Zig unit test in `src/tests_srfi18.zig`: a child OS thread caches a user procedure and the pristine `apply` (via `guard_builtin`), a mutex/condition-variable handoff (shared as globals, the supported route) guarantees the root rebinds both only afterwards, and the child re-reads only after the signal. Every wait has a timeout so a lost wakeup fails rather than hangs.
- [x] New Scheme suite file `tests/scheme/srfi/srfi18-global-cache-2483.scm` with the same handoff. Run against the pre-fix `main` binary it fails with exactly the stale result: `expected (first 3 second user-apply) got (first 3 first 3)`. With this branch: `2 passed, 0 failed`.
- [x] `zig build test`: 1913 pass, 21 skip, 0 fail (+ 88 thottam tests).
- [x] `srfi18.scm`, `srfi18-cross-thread-wait.scm`, `srfi18-sharing-model.scm`, `thread-notifier-waits-2395.scm` all pass with the fixed binary.
- [x] `zig build -Dtarget=x86_64-windows` and `zig build wasm` succeed; `zig fmt --check` and `markdownlint-cli2` clean.

Closes #2483

### DIFF
--- a/src/globals.zig
+++ b/src/globals.zig
@@ -15,6 +15,18 @@ const Value = types.Value;
 pub const GlobalsRwLock = struct {
     /// Bit 31 = writer holds/wants the lock; low 31 bits = active readers.
     state: std.atomic.Value(u32) = .init(0),
+    /// Generation of the globals map this lock guards: bumped by every
+    /// rebinding that can invalidate a per-function global cache
+    /// (`Function.cache_version` is compared against it). It lives on the
+    /// lock rather than on the VM because the lock is the one object
+    /// `VM.initForThread` shares by pointer with every SRFI-18 child: a
+    /// per-VM counter meant a child's caches were invalidated only by the
+    /// child's own rebindings, so after the root redefined a global the
+    /// child kept serving the old binding -- and kept a `guard_builtin`
+    /// on its fast path after the root redefined the builtin (kaappi#2483).
+    /// Read through `VM.globalVersion`, advanced through
+    /// `VM.bumpGlobalVersion`; see those for the ordering rules.
+    version: std.atomic.Value(u32) = .init(0),
 
     const WRITER: u32 = 0x8000_0000;
 
--- a/src/runtime_exports.zig
+++ b/src/runtime_exports.zig
@@ -123,7 +123,7 @@ pub export fn kaappi_set_global(vm: ?*vm_mod.VM, name_ptr: [*]const u8, name_len
         // fallback, say) that cached this global — including the pristine
         // primitive a guard_builtin compares against (kaappi#2469) — must
         // not keep serving the old value.
-        v.global_version +%= 1;
+        _ = v.bumpGlobalVersion();
     } else {
         _ = platform.write(2, "set!: unbound variable '", 24);
         _ = platform.write(2, name_ptr, len);
--- a/src/tests_srfi18.zig
+++ b/src/tests_srfi18.zig
@@ -1127,3 +1127,63 @@ test "kaappi#2473: thread-start! unwinds extra_roots when the spawn setup fails"
     // leaves one extra entry, which stays for the life of the GC.
     try std.testing.expectEqual(before, gc.extra_roots.items.len);
 }
+
+// kaappi#2483: the globals generation counter was a per-VM field, and
+// VM.initForThread neither copied nor shared it, so a child thread's
+// per-function global caches were invalidated only by the child's OWN
+// rebindings. After the root redefined a global, a child that had already
+// cached it kept calling the old binding -- and, since kaappi#2469, a child's
+// guard_builtin kept its fast path after the root redefined the builtin,
+// because its cached value still compared equal to the pristine primitive.
+// The counter now lives on the GlobalsRwLock the child shares by pointer.
+//
+// The handoff is fully ordered through a mutex and two condition variables
+// shared as globals (the supported route for sync primitives), so the child
+// provably fills its caches BEFORE the root rebinds and re-reads AFTER. Every
+// wait carries a timeout, so a lost wakeup fails the test instead of hanging
+// it. Without the fix the child returns (first 3 first 3).
+test "child thread sees the root's rebinding of a cached global and a guarded builtin (#2483)" {
+    if (comptime platform.is_wasm) return error.SkipZigTest; // OS threads are unregistered on wasm (thread-start! is native-only)
+    var gc = memory.GC.init(std.testing.allocator);
+    defer gc.deinit();
+    var vm = try th.makeTestVM(&gc);
+    defer vm.deinit();
+
+    _ = try vm.eval("(define m (make-mutex))");
+    _ = try vm.eval("(define cv-cached (make-condition-variable))");
+    _ = try vm.eval("(define cv-rebound (make-condition-variable))");
+    _ = try vm.eval("(define (f) 'first)");
+    _ = try vm.eval(
+        \\(define (child)
+        \\  ;; call_global caches f; guard_builtin caches the pristine apply.
+        \\  (let ((f-before (f))
+        \\        (apply-before (apply + (list 1 2))))
+        \\    (mutex-lock! m)
+        \\    (condition-variable-signal! cv-cached)
+        \\    ;; Atomically release m and wait for the root's rebinding.
+        \\    (mutex-unlock! m cv-rebound 60)
+        \\    (list f-before apply-before (f) (apply + (list 1 2)))))
+    );
+    _ = try vm.eval("(define t (make-thread child))");
+    _ = try vm.eval("(mutex-lock! m)");
+    _ = try vm.eval("(thread-start! t)");
+    // Blocks until the child has filled its caches; the child cannot signal
+    // before it holds m, which this wait is what releases.
+    _ = try vm.eval("(mutex-unlock! m cv-cached 60)");
+    // Reacquire m: the child releases it only by entering its own wait, so
+    // once this returns the child is parked on cv-rebound.
+    _ = try vm.eval("(mutex-lock! m)");
+    _ = try vm.eval("(define (f) 'second)");
+    _ = try vm.eval("(define (apply proc args) 'user-apply)");
+    _ = try vm.eval("(condition-variable-signal! cv-rebound)");
+    _ = try vm.eval("(mutex-unlock! m)");
+    const result = try vm.eval(
+        \\(let ((r (thread-join! t 60 'join-timeout)))
+        \\  (if (equal? r '(first 3 second user-apply))
+        \\      'ok
+        \\      (string->symbol
+        \\        (let ((p (open-output-string))) (write r p) (get-output-string p)))))
+    );
+    try std.testing.expect(types.isSymbol(result));
+    try std.testing.expectEqualStrings("ok", types.symbolName(result));
+}
--- a/src/vm.zig
+++ b/src/vm.zig
@@ -399,6 +399,10 @@ pub const VM = struct {
     ///     every map read;
     ///   - the owner thread reads lock-free: it is the only thread expected
     ///     to structurally mutate the map, so its own reads cannot race.
+    /// The map's generation counter (`GlobalsRwLock.version`, read via
+    /// `globalVersion`) rides on the same shared lock object, so every
+    /// thread's per-function global caches are invalidated by every
+    /// thread's rebindings (kaappi#2483).
     /// Known gap: a child that defines globals via `eval` takes the exclusive
     /// lock (protecting other children), but the owner's lock-free reads can
     /// still race such writes — same limitation PR #968 documented for child
@@ -553,7 +557,6 @@ pub const VM = struct {
     watches: [16]WatchEntry = undefined,
     watch_count: usize = 0,
     inspect_frame: usize = 0,
-    global_version: u32 = 0,
     profile_mode: bool = false,
     coverage_mode: bool = false,
     coverage_xml_path: ?[]const u8 = null,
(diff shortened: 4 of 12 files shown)


src/globals.zig
```diff
@@ -15,6 +15,18 @@ const Value = types.Value;
...
+    /// (`Function.cache_version` is compared against it). It lives on the
+    /// lock rather than on the VM because the lock is the one object
+    /// `VM.initForThread` shares by pointer with every SRFI-18 child: a
+    /// per-VM counter meant a child's caches were invalidated only by the
+    /// child's own rebindings, so after the root redefined a global the
+    /// child kept serving the old binding -- and kept a `guard_builtin`
+    /// on its fast path after the root redefined the builtin (kaappi#2483).
+    /// Read through `VM.globalVersion`, advanced through
+    /// `VM.bumpGlobalVersion`; see those for the ordering rules.
+    version: std.atomic.Value(u32) = .init(0),
```
baijum, line 29: Placement is right — the lock object is the one thing `initForThread` provably shares (`globals_lock = parent.globals_lock`), so the counter cannot drift from the map it versions. One inherited caveat worth having in mind: the counter stays u32 with equality-based validation, so a wraparound ABA (2^32 rebindings returning to a stale cache's stamped version) remains theoretically reachable. Same exposure as the per-VM field before — and the counter now advances on every thread's rebinding rather than one thread's, so it wraps sooner in wall-clock terms — but 2^32 is comfortably out of reach for real workloads. Fine to leave as is.

src/vm.zig
```diff
@@ -1342,8 +1345,32 @@ pub const VM = struct {
...
+
+    /// Advance the shared generation after a rebinding, returning the new
+    /// value. A writer that then re-validates its own cache slot must stamp
+    /// exactly this value (not a fresh `globalVersion`): a concurrent
+    /// rebinding by another thread may already have moved the counter past
+    /// it, and stamping the later value would bless the writer's slot
+    /// against a map it no longer matches. Call it inside the same locked
+    /// region as the store, so the store and its bump are ordered together.
+    pub inline fn bumpGlobalVersion(self: *VM) u32 {
+        return self.globals_lock.version.fetchAdd(1, .acq_rel) +% 1;
```
baijum, line 1373: Verified both documented rules are actually upheld: all eight `cache_version` stamp sites use either the writer's own captured bump result (`nv`) or the reader's pre-read snapshot (`gv`) — no post-read fresh load survives anywhere. Also checked the subtler consequence: in `set_global`/`define_global` the memset + stamp run *after* the lock is released, so another thread may have bumped past `nv` in between — that is safe precisely because the stamp uses the older own value (the slot just gets re-invalidated on its next use; one wasted refill, never staleness). And the `guard_builtin` inheritance claim checks out: `guardBuiltin` resolves through `lookupGlobalCached`, so the #2469 fast-path re-validation is covered by the same snapshot. The doc comments on these two methods are unusually precise — thank you for writing the rule down where the next writer will read it.

src/vm_dispatch.zig
```diff
@@ -320,23 +320,27 @@ pub fn runUntil(self: *VM, target_frame_count: usize, target_wind_count: usize)
                     // barrier, so this order is as safe as barrier-first.
                     self.gc.envStoreBarrier(func.env_val, val);
                 }
+                // kaappi#2483: the version is shared with every child VM, so
+                // bump it inside the locked region that made the store, and
+                // stamp the cache below with the value THIS bump produced —
+                // see VM.bumpGlobalVersion.
+                const new_version: ?u32 = if (ptr != null and func.env == null) self.bumpGlobalVersion() else null;
```
baijum, line 327: Confirming this is sound as written, since it looks off at a glance: the bump runs under the *shared* lock, not the exclusive one. That's fine — the rule is "inside the locked region that made the store", and the shared region does cover this in-place store. Two concurrent `set_global` writers can hold the shared lock simultaneously and interleave stores/bumps, but a same-name cross-thread `set!` is the already-documented #1924 hazard, and with distinct names each stamp's `nv` still describes a map the writer itself produced. No change needed — leaving the reasoning here for the next reader.

src/vm_imports.zig
```diff
@@ -65,7 +65,7 @@ fn importBinding(vm: *VM, target: *std.StringHashMap(Value), name: []const u8, v
         return;
     }
     if (target == vm.globals) {
-        vm.global_version +%= 1;
+        _ = vm.bumpGlobalVersion();
```
baijum: This bump sits *outside* the locked region that made the put, which contradicts the ordering rule this same PR writes on `VM.bumpGlobalVersion` ("call it inside the same locked region as the store"). I convinced myself it's benign as written: the return value is discarded, `importBinding` never stamps a cache slot, and the reader rule is safe against a delayed bump — the worst case is a spurious cache invalidation. But the doc now states the rule as absolute, and the next writer that *does* stamp a slot will copy this shape. Since `types.isTransformer(val)` is already known before the first block, the bump can move inside the lock while preserving exact current semantics:

```zig
if (target == vm.globals) {
    vm.globals_lock.lock();
    defer vm.globals_lock.unlock();
    target.put(name, val) catch return error.OutOfMemory;
    if (!types.isTransformer(val)) _ = vm.bumpGlobalVersion();
}
```

(The `!isTransformer` guard keeps today's deliberate behavior of not bumping for transformers — safe only because a cache entry is only ever a closure or native procedure, so a transformer put can never strand a stale cached value. Bumping there unconditionally would also be fine, just mildly conservative.)

src/runtime_exports.zig
```diff
@@ -123,7 +123,7 @@ pub export fn kaappi_set_global(vm: ?*vm_mod.VM, name_ptr: [*]const u8, name_len
         // fallback, say) that cached this global — including the pristine
         // primitive a guard_builtin compares against (kaappi#2469) — must
         // not keep serving the old value.
-        v.global_version +%= 1;
+        _ = v.bumpGlobalVersion();
```
baijum, line 126: Pre-existing, but worth a follow-up issue now that the counter is formally shared: `kaappi_set_global` does its `getPtr` + store with no lock at all — neither the shared lock the interpreter's `set_global` opcode takes around exactly this sequence (its own comment: a concurrent rehash would leave `ptr` dangling) nor the exclusive one. The bump swap itself is correct; the surrounding store can still race a structural mutation from another thread (an `import` or `define` under the write lock), and that is the memory-unsafety flavor, not just stale-cache. If the embedding contract is "host calls this only from the owner thread with no concurrent Scheme running", that's worth stating here; otherwise this wants the same `lockGlobalsShared`/`unlockGlobalsShared` pair the opcode uses. Suggest filing separately rather than widening this PR.

src/tests_srfi18.zig
```diff
@@ -1127,3 +1127,63 @@ test "kaappi#2473: thread-start! unwinds extra_roots when the spawn setup fails"
...
+// guard_builtin kept its fast path after the root redefined the builtin,
+// because its cached value still compared equal to the pristine primitive.
+// The counter now lives on the GlobalsRwLock the child shares by pointer.
+//
+// The handoff is fully ordered through a mutex and two condition variables
+// shared as globals (the supported route for sync primitives), so the child
+// provably fills its caches BEFORE the root rebinds and re-reads AFTER. Every
+// wait carries a timeout, so a lost wakeup fails the test instead of hanging
+// it. Without the fix the child returns (first 3 first 3).
+test "child thread sees the root's rebinding of a cached global and a guarded builtin (#2483)" {
```
baijum, line 1145: Verified end-to-end on this branch vs pre-fix `main`: the Scheme twin fails on `main` with exactly `expected (first 3 second user-apply) got (first 3 first 3)` and passes here (`2 passed, 0 failed`); this Zig test plus the whole `tests_srfi18` file also pass under `-Dgc-stress=true`, and the full `zig build test` is green. The mutex/cv handoff is the right shape — provably ordered caches-then-rebind, with timeouts so a lost wakeup fails CI instead of hanging it. Nice touch that the failure diagnostic round-trips the wrong result through `write`, so a regression prints the actual stale list rather than a bare assertion failure.

baijum (commented):
Reviewed the ordering claims against the code and re-ran the evidence. The design is right: the lock object is the one thing `initForThread` provably shares by pointer, so the generation counter can no longer drift from the map it versions.

**Verified by inspection:**
- All eight former `global_version +%= 1` sites migrated 1:1 to `bumpGlobalVersion`; no per-VM field references remain (only comments).
- Every `cache_version` stamp site stamps either the writer's own captured bump result or the reader's pre-read snapshot — the two documented rules hold everywhere; no post-read fresh load survives.
- `guardBuiltin` resolves through `lookupGlobalCached`, so the #2469 superinstruction fast path inherits the fix as claimed.
- The LLVM/native tier has no separate version usage — nothing to migrate there.

**Reproduced the evidence** (built this branch and pre-fix `main` side by side):
- New Scheme test fails on `main` with exactly the predicted message — `expected (first 3 second user-apply) got (first 3 first 3)` — and passes here (`2 passed, 0 failed`).
- Full `zig build test` green; `tests_srfi18` also green under `-Dgc-stress=true`.
- `srfi18.scm` (44), `srfi18-cross-thread-wait.scm` (6), `srfi18-sharing-model.scm` (27), `thread-notifier-waits-2395.scm` (17) all pass on this branch.
- `zig fmt --check` clean on every changed file; `run-all.sh`'s `srfi/*.scm` glob picks the new suite file up automatically.

Two non-blocking notes inline: the `importBinding` bump sits outside its locked region, contradicting the rule this PR itself documents (one-line fix), and `kaappi_set_global` still does its store fully unlocked (pre-existing; suggest a follow-up issue rather than widening this PR).

src/vm_imports.zig
baijum: Moved in 0a04b4a6: the bump now sits inside the locked block, guarded by `!types.isTransformer(val)` so the transformer case keeps today's no-bump behaviour, with a comment on why that is safe (a cache entry is only ever a closure or native procedure). The trailing `if (target == vm.globals)` block is gone. Full `zig build test` green after the change (1913 pass, 21 skip), and the new Scheme test still passes.

src/runtime_exports.zig
baijum: Agreed, and filed separately rather than widening this PR: #2487 (`bug`, `priority: high`, `native-backend`, `concurrency`). The issue records both resolutions you outlined: take the same `lockGlobalsShared`/`unlockGlobalsShared` pair the opcode uses around getPtr + store + bump, or state and verify an owner-thread-only contract on `kaappi_set_global`. Rated `high` rather than `critical` because native-tier reachability from a child thread is unconfirmed and the path is one the sharing docs already mark as a hazard.

Issue #2488, opened 2026-09-02
CI: riscv64-test hangs in the QEMU unit-test step in ~1 of 4 runs since 2026-09-02, cancelled at the 45-minute cap


baijum:
**Kaappi version**

`main` from 17fe3fbe (2026-09-02) onward; also reproduced on PR #2486's branch.

**Platform**

The `riscv64-test` CI leg only (riscv64-linux unit tests under QEMU user-mode via binfmt).

**Description**

`riscv64-test` is being cancelled at its 45-minute cap in roughly one run in four, and the shape of those runs says it is a hang, not the slow tail #2458 diagnosed and fixed by raising the cap from 30 to 45 minutes.

Per-step timings for the "Unit tests (riscv64 via QEMU)" step, from the GitHub Actions job API:

| Outcome | Unit-test step | Runs (last 15 on main + PR #2486's 2) |
|---|---|---|
| success | 23–28 min | 11 |
| cancelled at 45 min | 41–43 min, still running | 4 |

Every cancellation was still inside the unit-test step, with the `unit-tests` binary alive when the runner killed orphan processes (`Terminate orphan process: pid (...) (unit-tests)`), and with no output between the step start and the cancel — the step only prints on completion. Every success finished that step in 28 minutes or less. Nothing lands in between: a slow tail would populate 29–44 minutes, and no run does. The four cancellations:

- main 17fe3fbe — run 33597653628 (first occurrence; parent 8f093f5a passed)
- main ebc418ed
- main 2381114a
- PR #2486 branch 0a04b4a6 — run 33660775463 (the same branch's previous commit passed the leg in 30 min)

The one earlier cancellation, 3fce40d2 on 2026-09-01, was a 27.5-minute unit step killed at the old 30-minute cap — #2458's slowness case, a different failure with the same symptom.

**When it started**

[...]

PR #2491, opened 2026-09-03
CI: bound each emulated unit test with --test-timeout so a hang names its test (#2488)

baijum:
Follow-up to #2490 for #2488 (the intermittent riscv64 hang). Does not close the issue: the hang itself is still open; this makes the next recurrence name the **test**, not just the chunk.

## Why

The eight-way chunking narrows a hang to a step, but a hung step still ends as a silent cancel. `zig build test` drives the test binary over its stdin/stdout (`--listen=-`), asks for one test at a time, and by default waits for that test's result **forever** — that is the whole 43-minute silence in the cancelled runs. Chunking further would mean a step per file.

## What

- **`tools/run-unit-test-chunk.sh` passes `zig build test --test-timeout 8m`** (`$KAAPPI_TEST_TIMEOUT` overrides). Zig 0.16 reports a test that overruns the bound by name (`'<file>.test.<title>' timed out after 8m`), kills its process, respawns the binary and continues from the next test — so one hang costs eight minutes, names its test, and every other test still gets a verdict.
- **Step caps grow** so the per-test bound fires before the step cap does (a step cap firing first would lose the name again). Each cap is the measured healthy time from run 33684320694 plus the 8m bound plus margin; the 55m job cap still holds setup + build + all chunks healthy (30m) + one hung test (8m).
- **s390x and ppc64le get the same flag** — they run the suite as one unfiltered step under QEMU and share the failure mode.
- `docs/dev/testing.md` CI table updated.

The same bound also covers the other way this protocol goes silent: a test that writes to fd 1 corrupts the IPC stream (the runner parses stray bytes as a message header and waits for a body that never comes). That is now a named per-test timeout instead of a step cancel.

## Verified

- Natively, `KAAPPI_TEST_TIMEOUT=40ms bash tools/run-unit-test-chunk.sh gc` reports `167 pass, 6 skip, 47 timeout`, one `error: '<test>' timed out after 40ms` line per test, and exits 1; `bash tools/run-unit-test-chunk.sh native` with the default passes (141 tests).
- `bash -n`, shellcheck, YAML load, and markdownlint are clean.
- Sizing: the fuzz chunk's 20 tests take ~4.5m *together* under QEMU, so no single test is near 8m.

## Still open

The hang has not recurred since #2490 landed (2 main runs + 2 PR runs green). A local repro loop of the old 1646-test rest chunk under a podman riscv64 container has completed 2 of 16 passes so far without a hang; if it catches one, the fix follows separately with the test named.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -600,14 +600,16 @@ jobs:
       # on top, so a hung test is killed and NAMED by `zig build` while the
       # step is still alive (a step cap that fires first would lose the
       # name again). Measured healthy times on the eight-way split under
-      # QEMU (run 33684320694): process 1.8m, concurrency 3.3m, io 2.1m,
-      # fuzz 6.0m, gc 2.9m, native 1.7m, tooling 2.1m, rest 6.9m.
+      # QEMU, two runs (33684320694 and PR #2491's), slowest of each:
+      # process 1.8m, concurrency 7.0m (3.3m on the other run -- the
+      # timing-sensitive chunk varies most with runner load), io 2.1m,
+      # fuzz 6.0m, gc 6.1m (2.9m), native 1.7m, tooling 2.1m, rest 6.9m.
       - name: Unit tests, process chunk (riscv64 via QEMU)
         timeout-minutes: 14
         run: bash tools/run-unit-test-chunk.sh process -Dtarget=riscv64-linux
 
       - name: Unit tests, concurrency chunk (riscv64 via QEMU)
-        timeout-minutes: 16
+        timeout-minutes: 20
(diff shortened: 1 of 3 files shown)


baijum:
First CI run on this branch: all legs green, riscv64 job 35m. Chunk times under the new caps: process 1.8m, concurrency **7.0m** (3.3m on the earlier measurement), io 2.0m, fuzz 5.3m, gc **6.1m** (2.9m earlier), native 1.8m, tooling 2.1m, rest 5.5m. Concurrency and gc vary with runner load enough that their caps (16/15) left under a minute over healthy + the 8m per-test bound, so a hang there could still have hit the step cap first. Pushed a second commit widening them to 20 and 18.

Issue #2489, opened 2026-09-02
macos-latest ReleaseSafe unit leg intermittently dies with 'terminated with signal KILL' (random victim test, multiple branches)


baijum:
**Kaappi version**

`main` (02984593) and PR branches on top of it; observed 2026-09-02 between ~14:18Z and ~18:50Z.

**Platform**

GitHub Actions, `macos-latest` (arm64), the `test (macos-latest, ReleaseSafe)` matrix leg, `zig build test -Doptimize=ReleaseSafe` step.

**Description**

The unit-test binary is being killed externally mid-suite: the zig runner reports `'<test>' terminated with signal KILL`, a different victim test each time, always deep into the suite (~12–16 min into the step; a healthy run of the step is ~10 min), always with every other leg of the same commit green (ubuntu Debug/ReleaseSafe/ReleaseFast, gc-stress, gc-stress-scheme, BSDs, WASM, Windows).

Occurrences observed today:

| time (Z) | branch | victim test |
|---|---|---|
| 14:18 | Ring cross-thread waiters PR | `tests_endian.test.endian: %host-big-endian? ...` |
| 15:24 | fix-2469-runtime-builtin-gate (#2481) | `tests_process_run.test.run-process: runs inside a spawned fiber...` |
| 17:19 | fix-2464-return-freed-pages (#2485) | `tests_process.test.process-wait phase2: a child reaped behind kaappi's back...` |
| 18:21 | same commit, rerun | `tests_process_run.test.run-process: a pipe past its buffer drains...` |
| 18:50 | same commit, rerun 2 | `tests_process_run.test.run-process: the timeout kill reaches a grandchild...` |

The `tests_endian` victim is the tell that attribution is incidental: it is a pure byte-order assertion with no children, no signals, no pipes — the binary simply happened to be running it when the SIGKILL arrived. Three different victims within the *same commit* rules out a deterministic content trigger.

Context: this leg was already strained today — #2484 (merged 13:44Z) raised this leg's timeout to 50 min after #2481's re-run was cancelled at 40:04Z on a slow runner. These failures are a different signature (a `fail` with SIGKILL, not a timeout `cancel`), but the same leg, the same day, suggests fleet-level trouble (slow/loaded runners) rather than any one PR's change.

**What I ruled out on #2485**

[...]

PR #2493, opened 2026-09-03
CI: record memory pressure and kernel kill reasons when the macOS unit step dies (#2489)

baijum:
Diagnostics only, for #2489 (macOS ReleaseSafe unit binary SIGKILLed mid-suite). Does not close it.

## What the data says so far

- Nine occurrences; every kill 11–13 min into a step a healthy run finishes in 10; every victim in the last ~2% of the suite (indices 1893–1921 of 1934: the process tests, twice the `tests_endian` file right after them).
- Locally the suite's memory peak is **4.7 GB at test 409** (the fixed-seed fuzz gate, 40 s in); the process tests run at ~1 GB. A plain RSS watermark would kill at the peak, never at the tail.
- A stray group kill from the process tests is ruled out by the build runner surviving: Zig's spawn leaves the test binary in the runner's process group, so `kill(-pgid)` there would take `zig build` down too.
- The runner's debug log (`gh run rerun --debug`) shows no termination event.

That leaves a kernel memorystatus kill on a swapping runner (fits "slow day" and "late in the run", not the victim position) or a userland `kill(pid)` from somewhere unexpected. They leave different traces on the runner, and only the runner can show them.

## What this adds (macOS leg only)

- **Sampler** before the unit step: every 5 s, `vm.swapusage`, `vm_stat` free/compressed pages, and the six largest processes by RSS, to `$RUNNER_TEMP`.
- **Failure-only diagnostics** after it: sampler tail, current memory, any `JetsamEvent*.ips` report, kernel `memorystatus`/`jetsam`/`kill` log lines from the last 40 min, and any log line naming `unit-tests`; the sampler log is uploaded as an artifact.
- **Stop** step (`always()`), so the loop does not linger as an orphan.

A green run pays one background loop and a small temp file. Both steps carry a "remove once #2489 is closed" note. The kill rate today has been high enough (3 of 3 attempts on one commit) that a few re-runs of this PR should capture one.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -288,6 +288,36 @@ jobs:
           fi
           exit "$code"
 
+      # kaappi#2489 diagnostics. On some macOS runs the unit binary is
+      # SIGKILLed mid-suite ("'<test>' terminated with signal KILL") with no
+      # runner-side event in the debug log, always 11-13 min into a step a
+      # healthy run finishes in 10, and always in the last ~2% of the suite
+      # (the process tests). The two live hypotheses are a kernel memory
+      # kill (macOS memorystatus, which writes a JetsamEvent report and a
+      # kernel log line naming the reason) and a stray kill(2) from inside
+      # the suite (which leaves neither). Nothing in a plain step log can
+      # tell them apart, so: a sampler records swap usage, free/compressed
+      # pages, and the largest processes every 5 s for the whole step, and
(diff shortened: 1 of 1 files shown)


baijum:
Superseded: the diagnostics captured the kernel's "no paging space" kill on their first run (see #2489), and #2494 (`-Dtest-strip` on the macOS leg) removed the cause. Not needed as a permanent step.

PR #2498, opened 2026-09-03
Drop the load-sensitive yield bound from the spinBackoff test

### LINKED ISSUE #2496: tests_platform spinBackoff test's 1 ms bound on the spin+yield phases is load-sensitive: 12.3 ms on macos-latest
**Kaappi version**

`main` at a2ca8504 merged into PR #2491's branch (e7ba4848); Zig 0.16.0, macos-latest ReleaseSafe, `-Dtest-strip=true`.

**Description**

`tests_platform.test.kaappi#2446: spinBackoff spins, then yields, then sleeps` failed once with

```
spin+yield phases took 12258 us
```

The test runs `spin_backoff_hint_iters + spin_backoff_yield_iters` (32 + 64) iterations of `platform.spinBackoff` and asserts the whole stretch stays under 1 ms, on the reasoning that "96 hints and yields are microseconds". The 32 hints are; the 64 `sched_yield` calls are not bounded by anything the test controls. A yield hands the CPU to any other runnable thread, and on a shared 3-vCPU CI VM that can be a full scheduler quantum per call — here ~190 µs each on average. The sleep phases of the same test already carry the right comment ("deliberately no upper bound (a loaded box oversleeps freely)"); the non-sleeping bound needs the same treatment for its yield half.

Run 33735999621, job `test (macos-latest, ReleaseSafe)`, 2000/2022 otherwise passed. A re-run of the failed job (attempt 2, fresh runner, cached build) failed the same assertion again: `spin+yield phases took 53284 us`. The identical stripped suite had passed this test on #2494's run half an hour earlier, so it is intermittent and load-dependent, not a regression; the strip only makes the leg fast enough to be scheduled against whatever else the host is doing.

**Suggested fix**

Keep the upper bound on the pure-spin phase only (32 `spinLoopHint`s, genuinely microseconds), and for the yield phase assert that it completes and is not *slower* than the first sleep iteration by more than some generous factor — or drop the upper bound on it entirely, as the sleep phases do. The assertions the test exists for (kaappi#2446: the sleep phases really sleep, with exact lower bounds) are unaffected.

**Priority rationale**

`priority: low` per `docs/dev/github-issues.md`: the behaviour under test is correct; only the test's diagnostic is wrong (a false red on a loaded runner).

baijum:
Closes #2496

## What

`tests_platform.test.kaappi#2446: spinBackoff spins, then yields, then sleeps` timed the spin and yield phases together (32 `spinLoopHint` + 64 `sched_yield` iterations of `platform.spinBackoff`) and asserted the whole stretch stays under 1 ms, on the reasoning that "96 hints and yields are microseconds". The 32 hints are; the 64 yields are not bounded by anything the test controls — each `sched_yield` hands the CPU to any other runnable thread, so on a loaded or shared runner each can cost a full scheduler quantum.

The behaviour under test is correct; only the test's bound is wrong. This keeps a tight upper bound on the pure-spin phase only and drops it from the yield phase.

## Why drop the yield bound rather than loosen it

- A finite bound cannot both survive load and detect misrouting: a yield phase that collapsed into the sleep path would sleep tens of milliseconds itself (its iterations double up to the 1 ms cap), which is indistinguishable from the tens of milliseconds a loaded box already adds. That is exactly the reasoning the sleep phases' existing "deliberately no upper bound (a loaded box oversleeps freely)" comment encodes; the yield half now gets the same treatment.
- The issue's alternative — bound the yield phase relative to the first sleep iteration by a generous factor — has no stable ceiling either: the first sleep oversleeps freely under the same load, so the ratio of two load-sensitive quantities converges to nothing. CI's 53.3 ms yield phase against a 32 µs floor sleep is already a factor of ~1650.

## The change (`src/tests_platform.zig`)

- The single timed loop is split. Iterations `0..spin_backoff_hint_iters` — pure spin, `spinLoopHint` never enters the kernel, genuinely microseconds — keep the tight `< 1 ms` bound; the whole phase collapsing into the sleep path (32 × 32 µs > 1 ms) still crosses it.
- The 64 yield iterations now only have to complete; no upper bound, with a comment (kaappi#2496) stating the rationale.
- The exact lower bounds the test exists for are untouched: `first_sleep_ns >= 32_000`, `capped_sleep_ns >= 1_000_000`.

## Verification

Test-only fix: the modified test is itself the regression proof — it false-reds on `main`'s bound under load and passes with this change under the same load. All runs: `zig build test -Dtest-filter=tests_platform` (ReleaseSafe default), on a 12-CPU machine whose ambient load average was ~300.

| test | runs | result |
|------|------|--------|
| fixed (this branch), no extra load | 1 | green, 21/21 |
| fixed (this branch), 8 busy loops + ambient | 3 | green, 21/21 each (14–205 ms suite wall time) |
| old (main, via stash round-trip), same load | 1 | **false red**: `spin+yield phases took 262323 us` |

For context, CI observed 12.3 ms and 53.3 ms (macos-latest, run 33735999621, both attempts), and the issue session reproduced the false red locally at 2899 µs and 19647 µs under ambient load alone.

### DIFF
--- a/src/tests_platform.zig
+++ b/src/tests_platform.zig
@@ -151,15 +151,28 @@ test "denormal arithmetic survives after normalizeFpEnvBestEffort" {
 // memory.spinLock, VM.stopForCollection, markLiveChildRoots and reapOsThread
 // all rely on. The cap iteration (five doublings past the phases) sleeps a
 // full millisecond.
+//
+// kaappi#2496: only the pure-spin half of the non-sleeping iterations can
+// carry an upper time bound. Each yield iteration is a sched_yield, which
+// hands the CPU to any other runnable thread and can therefore cost a full
+// scheduler quantum on a loaded or shared runner -- CI observed the 64-yield
+// phase at 12.3 ms and 53.3 ms. No bound survives that: a yield phase that
+// collapsed into the sleep path would sleep tens of milliseconds too, which
+// is indistinguishable from load. The yield half is therefore unbounded from
+// above, like the sleep phases below.
 test "kaappi#2446: spinBackoff spins, then yields, then sleeps" {
     if (comptime is_wasm or builtin.single_threaded) return error.SkipZigTest;
-    const phases = platform.spin_backoff_hint_iters + platform.spin_backoff_yield_iters;
-    // Every iteration inside the two non-sleeping phases together must stay
-    // far below one sleep quantum: 96 hints and yields are microseconds.
+    const hints = platform.spin_backoff_hint_iters;
+    const phases = hints + platform.spin_backoff_yield_iters;
+    // Pure-spin phase: 32 spinLoopHint iterations, none of which enter the
+    // kernel -- genuinely microseconds, so the whole phase collapsing into
+    // the sleep path (32 x 32 us > 1 ms) still crosses the bound.
     const t0 = platform.monotonicNs();
     var i: u32 = 0;
+    while (i < hints) : (i += 1) platform.spinBackoff(i);
+    const spin_ns = platform.monotonicNs() - t0;
+    // Yield phase: the 64 sched_yield iterations only have to complete.
     while (i < phases) : (i += 1) platform.spinBackoff(i);
-    const non_sleeping_ns = platform.monotonicNs() - t0;
     // The first sleeping iteration is 32 us; the capped one (>= 5 doublings
     // past the phases) is 1 ms. nanosleep never returns early, so the lower
     // bounds are exact; there is deliberately no upper bound (a loaded box
@@ -172,8 +185,8 @@ test "kaappi#2446: spinBackoff spins, then yields, then sleeps" {
     const capped_sleep_ns = platform.monotonicNs() - t2;
     try std.testing.expect(first_sleep_ns >= 32_000);
     try std.testing.expect(capped_sleep_ns >= 1_000_000);
-    if (non_sleeping_ns >= 1_000_000) {
-        std.debug.print("spin+yield phases took {d} us\n", .{non_sleeping_ns / 1000});
-        return error.NonSleepingPhasesSlept;
+    if (spin_ns >= 1_000_000) {
+        std.debug.print("spin phase took {d} us\n", .{spin_ns / 1000});
+        return error.SpinPhaseSlept;
     }
 }
--- a/src/tests_platform.zig
+++ b/src/tests_platform.zig
@@ -144,9 +144,11 @@ test "denormal arithmetic survives after normalizeFpEnvBestEffort" {
     try std.testing.expect(q == std.math.floatTrueMin(f64));
 }
 
-// kaappi#2446: the backoff's phase schedule. Iterations inside the spin and
-// yield phases must return without sleeping, and the first iteration past
-// them must actually sleep -- that sleep is the whole point (it is what takes
+// kaappi#2446: the backoff's phase schedule. Iterations inside the spin
+// phase must return without sleeping (the yield phase is run but, per
+// kaappi#2496, cannot be bounded from above), and the first iteration past
+// the phases must actually sleep -- that sleep is the whole point (it is
+// what takes
 // a waiter off the run queue), and the phase constants are what
 // memory.spinLock, VM.stopForCollection, markLiveChildRoots and reapOsThread
 // all rely on. The cap iteration (five doublings past the phases) sleeps a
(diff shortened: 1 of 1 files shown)


src/tests_platform.zig
```diff
@@ -151,15 +151,28 @@ test "denormal arithmetic survives after normalizeFpEnvBestEffort" {
...
-    // far below one sleep quantum: 96 hints and yields are microseconds.
+    const hints = platform.spin_backoff_hint_iters;
+    const phases = hints + platform.spin_backoff_yield_iters;
+    // Pure-spin phase: 32 spinLoopHint iterations, none of which enter the
+    // kernel -- genuinely microseconds, so the whole phase collapsing into
+    // the sleep path (32 x 32 us > 1 ms) still crosses the bound.
     const t0 = platform.monotonicNs();
     var i: u32 = 0;
+    while (i < hints) : (i += 1) platform.spinBackoff(i);
+    const spin_ns = platform.monotonicNs() - t0;
```
baijum: The spin phase is still wall-clock bounded on a preemptible thread, so this keeps a smaller version of the #2496 flake. Between `t0` and this read the thread can be descheduled at any instruction; on macOS that costs a full ~10 ms quantum, which crosses 1 ms exactly the way the yield phase did. "Genuinely microseconds" is true of CPU time, not wall time.

The window is sub-microsecond (32 `spinLoopHint`s plus two `clock_gettime` commpage reads), so per run this is roughly window/quantum, on the order of 1e-4 to 1e-5 on a saturated runner. That is orders of magnitude better than 64 explicit yields, but the suite runs this on every leg of every PR, and this is the test #2496 is closing as flaky.

`spinBackoff(i)` for `i < hints` is stateless, so the phase can be repeated and the minimum taken. Three consecutive preemptions inside a sub-microsecond window do not happen:

```suggestion
    // Preemption inside even this window costs a full quantum, so take the
    // best of three: a descheduled attempt cannot mask an unpreempted one.
    var spin_ns: u64 = std.math.maxInt(u64);
    var attempt: u8 = 0;
    while (attempt < 3) : (attempt += 1) {
        const t0 = platform.monotonicNs();
        var i: u32 = 0;
        while (i < hints) : (i += 1) platform.spinBackoff(i);
        spin_ns = @min(spin_ns, platform.monotonicNs() - t0);
    }
    var i: u32 = hints;
```

(The suggestion replaces lines 170–173; `i` then starts the yield loop at `hints` as before.)

src/tests_platform.zig
```diff
@@ -151,15 +151,28 @@ test "denormal arithmetic survives after normalizeFpEnvBestEffort" {
...
 test "kaappi#2446: spinBackoff spins, then yields, then sleeps" {
     if (comptime is_wasm or builtin.single_threaded) return error.SkipZigTest;
-    const phases = platform.spin_backoff_hint_iters + platform.spin_backoff_yield_iters;
-    // Every iteration inside the two non-sleeping phases together must stay
-    // far below one sleep quantum: 96 hints and yields are microseconds.
+    const hints = platform.spin_backoff_hint_iters;
+    const phases = hints + platform.spin_backoff_yield_iters;
+    // Pure-spin phase: 32 spinLoopHint iterations, none of which enter the
+    // kernel -- genuinely microseconds, so the whole phase collapsing into
+    // the sleep path (32 x 32 us > 1 ms) still crosses the bound.
```
baijum: Two things about the "collapsing into the sleep path (32 x 32 us > 1 ms)" justification.

1. The margin is 2.4%: 32 × 32 µs is 1.024 ms. It clears the bound in practice only because `nanosleep` never returns early and typically oversleeps by tens of µs per call, which the comment does not say. Since an unpreempted spin phase finishes in single-digit µs and a preempted one blows past *any* bound under ~10 ms, the bound could be tightened (say `>= 100_000`) at zero extra flake cost, and the collapse argument would then hold by 10× rather than by 24 µs.

2. In the current `spinBackoff` nothing can route `i < 32` to the sleep branch: dropping the hint guard sends those iterations to `yield`, and dropping both guards makes `spins - 96` overflow a `u32`, which panics in ReleaseSafe before reaching `sleepNs`. So what this bound actually guards is a future rewrite of the schedule, not a reachable state of today's code. Worth saying so, or the next reader will look for the path and not find it.

src/tests_platform.zig
```diff
@@ -151,15 +151,28 @@ test "denormal arithmetic survives after normalizeFpEnvBestEffort" {
...
 // full millisecond.
+//
+// kaappi#2496: only the pure-spin half of the non-sleeping iterations can
+// carry an upper time bound. Each yield iteration is a sched_yield, which
+// hands the CPU to any other runnable thread and can therefore cost a full
+// scheduler quantum on a loaded or shared runner -- CI observed the 64-yield
+// phase at 12.3 ms and 53.3 ms. No bound survives that: a yield phase that
+// collapsed into the sleep path would sleep tens of milliseconds too, which
+// is indistinguishable from load. The yield half is therefore unbounded from
+// above, like the sleep phases below.
```
baijum, line 164: Nit: the test's opening paragraph (lines 148–149, outside the hunk) still states "Iterations inside the spin and yield phases must return without sleeping" as what this test checks. After this change only the spin half is asserted; the yield half is run but nothing about it is verified. This paragraph qualifies that, but the first sentence reads as a live claim. Suggest rewording it to "Iterations inside the spin phase must return without sleeping (the yield phase is run but, per kaappi#2496, cannot be bounded)" or similar, so the two paragraphs do not contradict each other.

baijum (commented):
Read the diff against `platform.spinBackoff` / `sleepNs` / `monotonicNs` on `main` and the CI history in #2496. The split is the right shape: the yield half genuinely has no load-proof ceiling (misrouted yields would sleep ~2–60 ms, the same range CI already measured under load), and the ratio-to-first-sleep alternative in the issue divides two load-sensitive numbers, so dropping the bound rather than loosening it is correct. The two lower bounds the test exists for are untouched, and CI is green on every leg.

Three inline notes. The first is the only one I'd act on before merging: the spin phase carries the same *kind* of flake the PR is closing, just at much lower probability, and a min-of-3 makes it effectively zero for free. The other two are comment accuracy.

src/tests_platform.zig
baijum: Applied in 337280ac, essentially verbatim: the spin phase is timed three times and the minimum taken (it is stateless for i < hints), then `i` resumes at `hints` for the yield loop. Filtered suite green at ambient load and 5/5 green under 24 synthetic busy loops — the same setup where the pre-fix test false-reded 2 of 3.

src/tests_platform.zig
baijum: Both points taken and applied in 337280ac. The bound is now `>= 100_000`, so the collapsed-sleep margin is 10x instead of 24 us, and the comment states the nanosleep-never-returns-early (and oversleeps) premise. It also now says what the bound guards: per the routing in spinBackoff (hint guard, then yield guard, then `spins - 96` which underflows u32 in ReleaseSafe), nothing can send i < hints to the sleep branch today — the bound is a tripwire for a future rewrite of the phase schedule, not a reachable state.

src/tests_platform.zig
baijum: Reworded in 337280ac, along your suggested lines: the opening paragraph now says iterations inside the spin phase must return without sleeping, with the yield phase run-but-unbounded per kaappi#2496.

PR #2500, opened 2026-09-03
Lock the native tier's globals-map accesses against a concurrent rehash (#2487)

### LINKED ISSUE #2487: kaappi_set_global stores into the shared globals map with no lock, racing a concurrent rehash
**Kaappi version**

`main` after PR #2486 (kaappi#2483); noted in review of that PR.

**Platform**

All native-backend targets (aarch64/x86_64).

**Description**

`kaappi_set_global` in `src/runtime_exports.zig` — the native tier's `set!` on a top-level variable, emitted by `llvm_emit.zig` for every global assignment — does its `globals.getPtr` and the store through that pointer with **no lock at all**:

```zig
if (v.globals.getPtr(name)) |ptr| {
    ptr.* = val;
    _ = v.bumpGlobalVersion();
}
```

The interpreter's `set_global` opcode (`vm_dispatch.zig`) wraps exactly this sequence in `lockGlobalsShared`/`unlockGlobalsShared`, and its comment says why: a structural mutation of the shared map on another thread (an `import` or `define` under the exclusive lock, which may rehash and free the bucket array) between the `getPtr` and the store leaves `ptr` dangling. The native path has the same window with nothing guarding it. #2486 fixed only the version-bump half (the counter is now shared and the bump is correct); the surrounding store is the pre-existing gap.

This is the memory-unsafety flavour, not stale-cache: a store through a pointer into a freed bucket array.

**Reachability, and why `high` rather than `critical`**

The race needs a second OS thread performing a structural mutation of the globals map (a `define` or `import` via `eval`) concurrently with a native-compiled `set!`. Two things temper it:

- If the embedding contract is "native code runs only on the owner thread, with no concurrent child mutating globals", the store is covered by the same owner-thread lock-free protocol the `globals` field doc already describes (with the same documented gap for child-side `eval` defines). That contract is not stated anywhere today.
- Whether LLVM-emitted code can execute on a child SRFI-18 thread at all (a native procedure reaching a child through the thread thunk or a global) is not confirmed either way. `runtime_exports.zig` elsewhere (`kaappi_get_constant`-family caches) already assumes the eval fallback *can* run on a spawned thread.

Per `docs/dev/github-issues.md`: memory unsafety is the finding, but it sits behind a concurrency path the sharing docs mark as a hazard (kaappi#1924) and whose native-tier reachability is unconfirmed, so `high` rather than `critical`.

**Suggested resolution**

Either take the same `lockGlobalsShared`/`unlockGlobalsShared` pair around `getPtr` + store + bump that the opcode uses (the bump then also sits inside the locked region, per the rule on `VM.bumpGlobalVersion`), or state the owner-thread-only contract on `kaappi_set_global` and verify it holds (a native closure must never run on a child VM). The first is a three-line change and is strictly safer; add a `tests/scheme/compile/*.sh` case if a native procedure can be shown to run on a child thread.

**Priority rationale**

`priority: high`: a dangling store into the shared globals map, but only via a concurrent cross-thread structural mutation on a path already documented as a hazard, with native-tier reachability unconfirmed.

baijum:
## Summary

`kaappi_set_global` — the LLVM backend's entry point for every natively compiled `set!` on a top-level global — did its `globals.getPtr` and the store through that pointer with **no lock**. The interpreter's `set_global` opcode wraps exactly this sequence in `lockGlobalsShared`/`unlockGlobalsShared`; the native path had nothing, so a structural mutation on another thread (a `define`/`import` under the exclusive lock, which may rehash and free the bucket array) between the `getPtr` and the store left `ptr` dangling — a store into freed memory.

**Reachability (confirmed):** a native closure **can** run on an SRFI-18 child thread. `.native_closure` is on the deep-copy allow-list (`gc_deep_copy.zig` copies the object around the static `fn_ptr`), and `callNativeClosure` passes the **calling** VM — so on a child thread the emitted `@kaappi_set_global(ptr %vm, …)` receives the child VM, where `owns_globals == false` and the shared lock actually matters.

**Fix:** take the same `lockGlobalsShared`/`unlockGlobalsShared` pair around `getPtr` + store + `bumpGlobalVersion` (the bump inside the locked region, per the rule on `VM.bumpGlobalVersion`) — exactly as the issue's suggested resolution. The same protocol is applied to the two sibling unlocked reads of the shared map in the same file:

- `kaappi_global_lookup` — the emitter's global-reference call for every non-lexical global read, mirroring the interpreter's `lookupGlobalLocked`;
- `callPrimitive` — the fixnum-overflow fallbacks' primitive lookup (the thread-local `vm_instance` is the child VM on a child thread).

All three are no-ops on the owner VM (it reads lock-free by the documented protocol), so single-threaded native programs pay only two branch checks. No nesting hazard: every existing holder of a globals lock holds it only around short map operations, never across a call into Scheme/native code. `kaappi_define_global` was already safe (`defineGlobal` takes the exclusive lock).

## Checklist

- [x] `zig build test` passes — 1918 pass, 21 skip (+ 88 thottam tests)
- [x] Scheme test suites pass: full `tests/scheme/compile/` suite (36 scripts) and the SRFI-18 suites (`srfi18.scm`, `srfi18-cross-thread-wait.scm`, `srfi18-sharing-model.scm`, `srfi18-global-cache-2483.scm`) with an isolated `KAAPPI_HOME`
- [x] New tests added (below)
- [x] Documentation updated: `docs/dev/thread-value-sharing.md` — "The native tier follows the same protocol" paragraph + implementation-map row
- [x] Commit body explains *why* — no `CHANGELOG.md` edit

## Test plan

- [x] **Deterministic regression test** (`src/tests_srfi18.zig`): holds the exclusive globals lock on the root VM, calls `kaappi_set_global` on a `VM.initForThread` child from a real OS thread, and asserts the call **cannot complete** until the lock is released (then the value lands and the shared generation advanced). Verified it **fails without the fix** (the pre-fix store completes while the writer holds the lock) and passes with it.
- [x] **End-to-end compile-suite script** (`tests/scheme/compile/native-set-global-child-thread-2487.sh`): natively compiles a program where a child thread runs 200k native `set!`s on a shared global while the root evals 10k fresh defines (structural mutations that rehash the shared map); interpreter is the oracle, both tiers must print the child's last store. The crash/lost-store detection pre-fix is probabilistic (it passed one pre-fix run) — the deterministic gate is the unit test above; this script pins the reachability path (native procedure on a child thread) and tier agreement.

Closes #2487


src/runtime_exports.zig
```diff
@@ -116,15 +122,31 @@ pub export fn kaappi_set_global(vm: ?*vm_mod.VM, name_ptr: [*]const u8, name_len
...
+    // thread (define/import under the exclusive lock) may rehash and free
+    // the bucket array between them, leaving `ptr` dangling. The store
+    // itself is in-place (no rehash), so the shared lock is sufficient; it
+    // is a no-op on the owner VM, which reads lock-free. The version bump
+    // stays inside the locked region so the store and its bump are ordered
+    // together (VM.bumpGlobalVersion).
+    v.lockGlobalsShared();
+    const ptr = v.globals.getPtr(name);
+    if (ptr) |p| {
+        p.* = val;
```
baijum, line 156: The interpreter's `set_global` opcode does one more thing before it takes the lock that this mirror omits: the #1924 cross-heap guard (`vm_dispatch.zig` ~L270, `raiseCrossHeapStoreVM`). A child (`owns_globals == false`) storing a *child-heap* object into the shared map leaves a pointer the root collector cannot trace, dangling once the child's heap is freed at join.

This PR is what confirms a native closure runs on a child VM, so the gap is now a reachable UAF. Reproduced on this branch (`KAAPPI_HOME` isolated):

```scheme
(define (h) 1)
(define (worker) (set! h (list 1 2 3)))   ; target is a procedure name → "known" global → compiled natively
(define t (make-thread worker))
(define s (thread-start! t))
(thread-join! t)
(display h) (newline)
```

* interpreter: `uncaught exception in thread: set!: cannot store an object created on this thread into a heap object shared with another thread …`, `h` unchanged.
* `kaappi compile` binary: prints `(0.0 . 0.0)` (three runs, same) — the child-heap pair read after the child heap was freed. `--emit-llvm` shows the store is `call void @kaappi_set_global` inside `@r1.fast` (worker).

Suggest the same check, before taking the lock exactly as the opcode does, with the native tier's usual fatal-diagnostic shape (like the `set!: unbound variable` path just below):

```zig
if (!v.owns_globals and types.isPointer(val)) {
    const root_vm = v.root_vm orelse v;
    if (types.toObject(val).owner != root_vm.gc.id) {
        // same text as vm_dispatch_helpers.raiseCrossHeapStoreVM
        _ = platform.write(2, "set!: cannot store an object created on this thread into a heap object shared with another thread ...\n", ...);
        std.process.exit(1);
    }
}
```

`kaappi_define_global` does not need it: internal defines are rejected in native lambda bodies (#819), so it only ever runs at top level on the owner thread. If you'd rather keep this PR to the rehash race, please file it as `priority: critical` (memory unsafety from an ordinary short program, per `docs/dev/github-issues.md`) and soften the "follows the same protocol" doc paragraph — but it is a few lines and this is the natural place.

tests/scheme/compile/native-set-global-child-thread-2487.sh
```diff
@@ -0,0 +1,97 @@
...
+export KAAPPI_HOME="$DIR/home"
+mkdir -p "$KAAPPI_HOME"
+
+name=native-set-global-child-thread-2487
+cat > "$DIR/$name.scm" << 'EOF'
+(import (scheme base) (srfi 18))
+
+(define x 0)
+
+(define (child-loop n)
```
baijum, line 58: This script does not reach `kaappi_set_global` at all. `kaappi --emit-llvm` on the embedded program gives:

```
call void @kaappi_set_global : 0
call i64 @kaappi_eval_cached : 5   (every top-level form, including this define)
```

Two things force `child-loop` to the eval fallback: the named `let` (`sexprNeedsEvalFallback` rejects `(let <sym> …)`), and `x` being a *value* global — `nodeHasFreeVars` treats a `set!`/ref of a name that is not `isKnownOrReservedGlobal` (builtin or a top-level *procedure* name) as a free variable, so the lambda is rejected. The child's 200k `set!`s therefore run in the interpreter's `set_global` opcode, which was already locked; the script is a VM-vs-VM comparison inside a native binary, which is consistent with it having passed a pre-fix run. (The `import` line also makes every form after it a passthrough, but that is not the limiting factor here.)

A shape that does emit the native store inside the child's function, and prints `200000` on both tiers in ~0.7 s here:

```scheme
(define (x) 0)                 ; procedure-named global → the lambda's set! target is "known"
(define (child-loop n)
  (do ((i 1 (+ i 1))) ((> i n))   ; do is lowered natively (#1496); named let is not
    (set! x i)))
(define (run) (child-loop 200000))  ; a defined thunk; an inline (lambda …) arg to make-thread fell back in my probe
(define t (make-thread run))
(define started (thread-start! t))
(let loop ((i 0))
  (when (< i 10000)
    (eval (list 'define (string->symbol (string-append "pad-" (number->string i))) i))
    (loop (+ i 1))))
(thread-join! t)
(display x) (newline)
```

`--emit-llvm` for that: `kaappi_set_global` appears in `child-loop`'s fast entry. Since the point of a `compile/*.sh` case is native-tier evidence, I'd also have the script assert on the IR the way ten sibling scripts already do (`grep -l emit-llvm tests/scheme/compile/*.sh`): e.g. require `call void @kaappi_set_global` to appear *after* the first `define tailcc` line and not only in `@main`, so a future emitter change that quietly re-routes the body to eval turns the test red instead of hollow. The header comment's "a child thread running natively compiled code does 200k set!s" should then be true.

docs/dev/thread-value-sharing.md
```diff
@@ -257,6 +257,21 @@ and a reader snapshots it once *before* the map read and stamps the
 snapshot — a fresh load after the read could bless a value another thread
 had rebound in between (`VM.bumpGlobalVersion`, `VM.globalVersion`).
 
+**The native tier follows the same protocol.** The LLVM backend's global
```
baijum, line 260: Until `kaappi_set_global` carries the #1924 owner check (see the `runtime_exports.zig` comment), "follows the same protocol" overstates it: the native tier now follows the same *locking* protocol, but a child-side native `set!` of a child-heap object is accepted where the opcode raises. Either land the guard and keep this wording, or say "the same locking protocol" and name the remaining gap with its issue number.

src/tests_srfi18.zig
```diff
@@ -1187,3 +1187,81 @@ test "child thread sees the root's rebinding of a cached global and a guarded bu
...
+    // after this sleep can only mean the lock was not taken.
+    var waited_ms: usize = 0;
+    while (!entered.load(.acquire) and waited_ms < 5_000) : (waited_ms += 1) {
+        platform.sleepNs(std.time.ns_per_ms);
+    }
+    try std.testing.expect(entered.load(.acquire));
+    platform.sleepNs(100 * std.time.ns_per_ms);
+    try std.testing.expect(done.load(.acquire) == false);
+
+    vm.globals_lock.unlock();
```
baijum, line 1273: +1 to CodeRabbit's point on this: put `vm.globals_lock.unlock()` + `worker.join()` in a `defer` right after a successful spawn (unlock registered last so it runs first). As written, a failing `expect(entered …)` returns with the lock held and the worker never joined; the deferred `vm.deinit()` then destroys `globals_lock` under a thread still spinning in `lockShared`. In the regression case (`done == true`) the worker has already finished, so there is no hang — but the teardown UAF on the other path is worth a few lines.

Minor wording: the comment at L1202 calls this "deterministic". It is deterministic *with* the fix (the call provably blocks) and probabilistic against a regression (a worker descheduled across the 100 ms window would pass unfixed). That is fine for a regression gate; just say so, since the PR description leans on the word.

baijum (commented):
The locking change itself is correct and I verified it both ways: the new unit test fails on unfixed `main` (dropped only `src/tests_srfi18.zig` onto `c12e9e7b`: `'kaappi_set_global on a child VM holds the shared globals lock (#2487)' failed`) and passes on this branch; the new compile script passes here (~34 s wall). `callPrimitive` correctly releases before `callWithArgs`, and the unbound-variable exit paths run after the unlock.

Two things I think need to change before merge, and two minor ones — details inline:

1. **`kaappi_set_global` still lacks the interpreter's #1924 cross-heap guard**, and the reachability this PR establishes (native closure → child VM) turns that into a demonstrable use-after-free: a 7-line program prints garbage from the freed child heap in the native tier where the interpreter raises. See the `runtime_exports.zig` comment.
2. **The compile script never executes `kaappi_set_global`.** `--emit-llvm` on its program emits zero calls; every top-level form is `kaappi_eval_cached` (named `let` and a value-global `set!` target both force the fallback), so the child's 200k `set!`s run in the already-locked interpreter opcode. That is why it "passed one pre-fix run". A shape that does reach the native store, and prints 200000 on both tiers, is in the script comment.

On CodeRabbit's three findings: the third (defer unlock/join in the test) is worth taking; the second is a fair wording point but not a defect; I disagree with the first — a concurrent 64-bit aligned `Value` store under the shared lock is exactly the interpreter's `set_global` protocol, and taking the exclusive lock for every native `set!` is not warranted.

baijum:
Review addressed in dc9b401d — thank you for the depth here, especially the two blocking findings.

**1. #1924 cross-heap guard** — landed exactly as suggested: the owner check on `kaappi_set_global` before the lock, fatal diagnostic carrying the `raiseCrossHeapStoreVM` text (minus the remediation tail, which names channel/join routes that reach the same refusal). Verified with your repro on this branch: the interpreter raises with `h` unchanged, the native binary now exits 1 with the diagnostic instead of printing `(0.0 . 0.0)` — and your repro is now the script's second case, with explicit per-tier expectations (they differ by design on the erroring program, so no `assert_tiers_agree` there). `kaappi_define_global` left alone per your #819 reasoning, and the doc paragraph now names the guard.

**2. Compile script** — rewritten to your shape (procedure-named target, `do` loop, defined thunk); `--emit-llvm` shows `call void @kaappi_set_global` inside `@r1.fast`, and the script now asserts that as an IR gate (first `define tailcc` < call line < `@main`) so a future emitter change that re-routes the body to eval turns it red instead of hollow. Both tiers print `200000`; the whole script runs in ~5 s.

**3. Test teardown (CodeRabbit's third, your +1)** — restructured: after a successful spawn a single guarded defer does unlock-then-join on every exit path, and the happy path runs the same pair before the post-join assertions. The `deterministic` wording is also corrected per your note: deterministic against the fix, a high-probability detector against a regression.

On CodeRabbit's other two I followed your lead: no exclusive lock for native stores (the concurrent aligned `Value` store under the shared lock is the opcode's own protocol), and no acquisition-observation hook — the wording point is acknowledged in the comment instead.

Re-validated on this branch: the unit test (7/7 in its filter), the rewritten compile script (both cases), `native-rebind-stale-call-822`, `set-define-lexical-scope-819`, `srfi18.scm` (44/44), `srfi18-global-cache-2483.scm`, `zig fmt --check`.

Issue #1412, opened 2026-07-10
sqrt returns inexact result for exact rational perfect squares


baijum:
## Problem

`(sqrt 4)` returns exact `2`, but `(sqrt 9/4)` returns inexact `1.5` even though the exact result `3/2` is representable:

```scheme
(sqrt 4)    ; => 2      (exact) ✓
(sqrt 9/4)  ; => 1.5    (inexact) — should be 3/2
(sqrt 1/4)  ; => 0.5    (inexact) — should be 1/2
```

R7RS (§6.2.6, `sqrt`): "Returns the principal square root of z. […] if the argument is an exact number for which the result can be represented as an exact number, implementations are encouraged to return an exact result." Chez, Chibi, Gauche, and Guile all return `3/2` for `(sqrt 9/4)`.

## Suggested fix

In the sqrt primitive (`src/primitives_arithmetic.zig`): when the argument is an exact `Rational`, take exact integer square roots of numerator and denominator; if both are perfect squares, return the exact rational, otherwise fall back to the current inexact path. Verify exact `Bignum` perfect squares are covered as well.

Add Scheme-level tests under `tests/scheme/` following the SRFI-64 conventions used there (exit-on-fail epilogue, see `tests/scheme/hygiene/*.scm`).

## Why it matters

The new **kaappi-mpl** symbolic math library (R7RS port of dharmatech/mpl) defines its symbolic `sqrt` as:

```scheme
(define (sqrt x)
  (if (and (number? x) (exact? (rnrs:sqrt x)))
      (rnrs:sqrt x)
      (^ x 1/2)))
```

Because Kaappi's `sqrt` is inexact for exact rationals, `(sqrt 9/4)` yields the symbolic `(^ 9/4 1/2)` instead of `3/2`, diverging from upstream mpl behavior on Chez/Racket.

PR #2505, opened 2026-09-03
Return an exact complex from sqrt of a negative exact perfect square

### LINKED ISSUE #2503: sqrt of a negative exact perfect square returns an inexact complex: (sqrt -1) => +1.0i, R7RS says +i
## Problem

R7RS §6.2.6 (`sqrt`, errata-corrected PDF p. 38) gives the example `(sqrt -1) ⟹ +i`. Under the convention stated at the top of §6.2.6 ("any numerical constant written using an exact notation is indeed represented as an exact number"), that `+i` is exact: zero real part, exact `1` imaginary part.

Kaappi returns the right value with the wrong exactness:

```scheme
(sqrt -1)                ; => +1.0i   expected +i
(exact? (sqrt -1))       ; => #f      expected #t
(eqv? (sqrt -1) +i)      ; => #f      expected #t
(real-part (sqrt -1))    ; => 0.0     expected 0
(sqrt -4)                ; => +2.0i   expected +2i
(sqrt -1/4)              ; => +0.5i   expected +1/2i
(* (sqrt -1) (sqrt -1))  ; => -1.0+0.0i   expected -1
```

`(sqrt -1)` and `(sqrt -1.0)` are indistinguishable. The sign convention is correct in every case (`(sqrt -0.0)` → `-0.0`, `(sqrt -inf.0)` → `+inf.0i`, `(sqrt -3-4i)` → `1.0-2.0i`); only exactness is lost.

## Cause

In `sqrtFn` (`src/primitives_numeric.zig`, ~line 562) the exact bignum and rational paths are guarded by `!isNegative`, and the fixnum perfect-square check runs only *after* the `f < 0.0` branch has already returned. Every negative exact argument therefore falls into the f64 path and is built with `makeComplexOrReal`, which always produces flonum components.

kaappi#1412 made the positive side exact (`(sqrt 9/4)` → `3/2`); the negative side never got the same treatment. The gap is recorded, not accidental: `tests/scheme/compliance/complex-exactness-2166-2167.scm:123` asserts `(inexact? (sqrt -4))` with the label "out of scope". The R7RS suite (`tests/scheme/r7rs/r7rs-tests.scm:1016`) wraps its check in `(inexact (sqrt -1))`, so it cannot catch this.

## Suggested fix

Exact complex values already exist (kaappi#2166, Value-typed components). In `sqrtFn`, when the argument is a negative exact number (fixnum, bignum, or rational): negate it, reuse the existing exact perfect-square paths (`isqrtNonNegative`, the rational num/den split), and on success return `gc.allocComplex(makeFixnum(0), root)`. Otherwise fall through to the inexact path as today.

Tests: flip the "out of scope" assertion to `(eqv? (sqrt -4) +2i)`, and add the negative cases (`-1`, `-4`, `-1/4`, a negative bignum square, and a negative non-square staying inexact) to `tests/scheme/compliance/sqrt-exact.scm`.

## Priority

`medium`: wrong against the spec, but the numeric value is right and `(exact ...)` recovers it. Silent, so it sits at the top of medium.

baijum:
## Summary

R7RS §6.2.6 (errata-corrected PDF p. 38) gives `(sqrt -1) ⟹ +i`, exact under the section's stated convention that exact notation denotes an exact number. Kaappi returned inexact `+1.0i` for every negative exact argument, because the exact bignum/rational paths in `sqrtFn` were guarded by `!isNegative` and the fixnum perfect-square check ran only after the `f < 0.0` branch had returned.

This PR:

- Folds the three exact paths in `sqrtFn` into one `exactSqrt` helper (fixnum, bignum, rational; returns `null` for a non-square).
- For a negative exact argument, takes the exact root of its negation and returns `0+root*i` with an exact fixnum zero real part.
- Leaves negative non-squares on the inexact complex path and negative inexact arguments untouched.

| Expression | Before | After |
|---|---|---|
| `(sqrt -1)` | `+1.0i` | `+i` |
| `(eqv? (sqrt -1) +i)` | `#f` | `#t` |
| `(sqrt -9/4)` | `+1.5i` | `+3/2i` |
| `(* (sqrt -1) (sqrt -1))` | `-1.0+0.0i` | `-1` |
| `(sqrt -2)` | `+1.4142135623730951i` | unchanged |
| `(sqrt -1.0)` | `+1.0i` | unchanged |

## Tests

- `tests/scheme/compliance/sqrt-exact.scm`: negative fixnum, bignum, rational, the 2^47 fixnum boundary, non-squares staying inexact, inexact arguments unchanged (46 passes, was 33).
- `tests/scheme/compliance/complex-exactness-2166-2167.scm`: the "out of scope" `(inexact? (sqrt -4))` assertion flips to `(eqv? (sqrt -4) +2i)`.
- R7RS suite 1395/0, numeric unit tests 58/58, and `sqrt-exact.scm` green under a `-Dgc-stress=true` build.

Fixes #2503

### DIFF
--- a/src/primitives_numeric.zig
+++ b/src/primitives_numeric.zig
@@ -570,24 +570,27 @@ fn sqrtFn(args: []const Value) PrimitiveError!Value {
         const i = i_sign * @sqrt((mag - re) / 2.0);
         return makeComplexOrReal(r, i);
     }
-    if (types.isBignum(args[0]) and !bignum_mod.isNegative(args[0])) {
+    if (types.isFixnum(args[0]) or types.isBignum(args[0]) or types.isRationalObj(args[0])) {
         const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
-        const r = try isqrtNonNegative(gc, args[0]);
-        if (bignum_mod.isZero(r.rem)) return r.root;
-    }
-    if (types.isRationalObj(args[0])) {
-        const rat = types.toRational(args[0]);
-        if (!bignum_mod.isNegative(rat.numerator)) {
-            const gc = memory.gc_instance orelse return PrimitiveError.OutOfMemory;
-            const num_r = try isqrtNonNegative(gc, rat.numerator);
-            if (bignum_mod.isZero(num_r.rem)) {
-                var slot_root = gc.rootedSlot(num_r.root) catch return PrimitiveError.OutOfMemory;
-                defer slot_root.release();
-                const den_r = try isqrtNonNegative(gc, rat.denominator);
-                if (bignum_mod.isZero(den_r.rem)) {
-                    return arith.makeRationalReduced(gc, num_r.root, den_r.root);
-                }
+        // Exact argument: an exact perfect square (integer or rational,
+        // either sign) yields an exact root. R7RS 6.2.6 gives
+        // (sqrt -1) => +i, an EXACT complex under the section's own
+        // convention that exact notation denotes an exact number
+        // (kaappi#2503) -- so a negative perfect square becomes the exact
+        // complex 0+root*i rather than falling through to the f64 path.
+        const negative = if (types.isRationalObj(args[0]))
+            bignum_mod.isNegative(types.toRational(args[0]).numerator)
+        else
+            bignum_mod.isNegative(args[0]);
+        if (negative) {
+            const pos = try negateExact(gc, args[0]);
+            var slot_pos = gc.rootedSlot(pos) catch return PrimitiveError.OutOfMemory;
+            defer slot_pos.release();
+            if (try exactSqrt(gc, slot_pos.get())) |root| {
+                return gc.allocComplex(types.makeFixnum(0), root) catch return PrimitiveError.OutOfMemory;
             }
+        } else if (try exactSqrt(gc, args[0])) |root| {
+            return root;
         }
     }
     const f = try toF64(args[0]);
@@ -595,12 +598,41 @@ fn sqrtFn(args: []const Value) PrimitiveError!Value {
         const imag = @sqrt(-f);
         return makeComplexOrReal(0.0, imag);
     }
-    const result = @sqrt(f);
-    if (types.isFixnum(args[0])) {
-        const ri: i64 = @intFromFloat(result);
-        if (ri * ri == types.toFixnum(args[0])) return types.makeFixnum(ri);
+    return makeFlonumVal(@sqrt(f));
+}
+
+/// Negate an exact number (fixnum, bignum, or rational). The result is
+/// unrooted; callers must root it before allocating.
+fn negateExact(gc: *memory.GC, v: Value) PrimitiveError!Value {
+    if (types.isRationalObj(v)) {
+        const rat = types.toRational(v);
+        const neg_num = bignum_mod.negate(gc, rat.numerator) catch return PrimitiveError.OutOfMemory;
+        var slot = gc.rootedSlot(neg_num) catch return PrimitiveError.OutOfMemory;
+        defer slot.release();
+        return arith.makeRationalReduced(gc, slot.get(), rat.denominator);
     }
-    return makeFlonumVal(result);
+    return bignum_mod.negate(gc, v) catch return PrimitiveError.OutOfMemory;
+}
+
+/// Exact square root of a NON-NEGATIVE exact number, or null when the
+/// argument is not a perfect square (so the caller falls back to the inexact
+/// path). Integers use the shared exact-integer-sqrt helper; a rational is a
+/// perfect square iff numerator and denominator both are. The result is
+/// unrooted.
+fn exactSqrt(gc: *memory.GC, v: Value) PrimitiveError!?Value {
+    if (types.isRationalObj(v)) {
+        const rat = types.toRational(v);
+        const num_r = try isqrtNonNegative(gc, rat.numerator);
+        if (!bignum_mod.isZero(num_r.rem)) return null;
+        var slot_root = gc.rootedSlot(num_r.root) catch return PrimitiveError.OutOfMemory;
+        defer slot_root.release();
+        const den_r = try isqrtNonNegative(gc, rat.denominator);
+        if (!bignum_mod.isZero(den_r.rem)) return null;
+        return try arith.makeRationalReduced(gc, slot_root.get(), den_r.root);
+    }
+    const r = try isqrtNonNegative(gc, v);
+    if (!bignum_mod.isZero(r.rem)) return null;
+    return r.root;
 }
 
 const IsqrtResult = struct { root: Value, rem: Value };
--- a/src/primitives_numeric.zig
+++ b/src/primitives_numeric.zig
@@ -578,6 +578,12 @@ fn sqrtFn(args: []const Value) PrimitiveError!Value {
         // convention that exact notation denotes an exact number
         // (kaappi#2503) -- so a negative perfect square becomes the exact
         // complex 0+root*i rather than falling through to the f64 path.
+        // The rational arm is load-bearing, not cosmetic: bignum_mod.isNegative
+        // answers false for any Value that is not a fixnum or bignum, so a
+        // flattened isNegative(args[0]) would classify every negative rational
+        // as non-negative and hand exactSqrt a negative numerator -- where
+        // isqrtNonNegative evaluates @intFromFloat(@sqrt(-9.0)), a NaN, and
+        // panics in ReleaseSafe on plain (sqrt -9/4).
         const negative = if (types.isRationalObj(args[0]))
             bignum_mod.isNegative(types.toRational(args[0]).numerator)
         else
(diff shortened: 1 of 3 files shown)


src/primitives_numeric.zig
```diff
@@ -570,37 +570,69 @@ fn sqrtFn(args: []const Value) PrimitiveError!Value {
...
+
+/// Negate an exact number (fixnum, bignum, or rational). The result is
+/// unrooted; callers must root it before allocating.
+fn negateExact(gc: *memory.GC, v: Value) PrimitiveError!Value {
+    if (types.isRationalObj(v)) {
+        const rat = types.toRational(v);
+        const neg_num = bignum_mod.negate(gc, rat.numerator) catch return PrimitiveError.OutOfMemory;
+        var slot = gc.rootedSlot(neg_num) catch return PrimitiveError.OutOfMemory;
+        defer slot.release();
+        return arith.makeRationalReduced(gc, slot.get(), rat.denominator);
```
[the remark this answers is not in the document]
baijum: False positive, I believe — two separate facts make this safe:

1. **This GC is non-moving.** A `toRational` result holds raw field pointers into the rational object, but addresses are stable across collections (the tree is explicit about this — e.g. `compiler_bindings.zig:24`, `gc_collect.zig:433`). Holding `rat.denominator` across an allocating call can't go stale; it can only be *swept* if the parent becomes unreachable, and here it can't: `v` is `args[0]`, which lives in the VM register file when `sqrtFn` calls `negateExact` directly, and `slot_pos` in the negative path. This exact pattern (grab `toRational`, allocate, then read the other field) is already established in this file — `squareFn` at src/primitives_numeric.zig:552-557.

2. The values that really are unrooted here — the results `negateExact` and `exactSqrt` return — are documented as such and are only ever handed to callees that root their own arguments before allocating: `allocComplex` does `rootArgs2(real, imag)` before `maybeCollect` (gc_alloc.zig:924-927), and `makeRationalReduced` roots both arguments in its bignum path (primitives_arithmetic.zig:170-175). I traced both.

Also empirically: the PR ran `sqrt-exact.scm` under `-Dgc-stress=true` (collection on every allocation), and I re-ran the full suite plus a stress-free pass locally — a dangling-root window in these paths would not survive that.

src/primitives_numeric.zig
```diff
@@ -570,37 +570,69 @@ fn sqrtFn(args: []const Value) PrimitiveError!Value {
...
+        const negative = if (types.isRationalObj(args[0]))
+            bignum_mod.isNegative(types.toRational(args[0]).numerator)
+        else
+            bignum_mod.isNegative(args[0]);
+        if (negative) {
+            const pos = try negateExact(gc, args[0]);
+            var slot_pos = gc.rootedSlot(pos) catch return PrimitiveError.OutOfMemory;
+            defer slot_pos.release();
+            if (try exactSqrt(gc, slot_pos.get())) |root| {
+                return gc.allocComplex(types.makeFixnum(0), root) catch return PrimitiveError.OutOfMemory;
```
baijum, line 601: Nit (non-blocking): the safety here is correct but non-local. `exactSqrt` documents that its result is **unrooted**, and this line passes that unrooted `root` straight into `allocComplex` — which happens to root its arguments callee-side (`rootArgs2(real, imag)` before `maybeCollect`, gc_alloc.zig:924-927). That holds today; it just means any future allocating call inserted between the `exactSqrt` call and this line (or swapping in a constructor that doesn't self-root) dangles.

Since the project rule is "root before allocating", consider either rooting `root` in a slot the way `pos` is one line up, or extending the comment to say the allocation is safe because `allocComplex` roots its arguments — so the dependency is visible at this call site rather than inside gc_alloc.zig.

src/primitives_numeric.zig
```diff
@@ -570,37 +570,69 @@ fn sqrtFn(args: []const Value) PrimitiveError!Value {
...
-                if (bignum_mod.isZero(den_r.rem)) {
-                    return arith.makeRationalReduced(gc, num_r.root, den_r.root);
-                }
+        // Exact argument: an exact perfect square (integer or rational,
+        // either sign) yields an exact root. R7RS 6.2.6 gives
+        // (sqrt -1) => +i, an EXACT complex under the section's own
+        // convention that exact notation denotes an exact number
+        // (kaappi#2503) -- so a negative perfect square becomes the exact
+        // complex 0+root*i rather than falling through to the f64 path.
+        const negative = if (types.isRationalObj(args[0]))
```
baijum, line 587: Nit: worth one pinning comment, because the `isRationalObj` special case is load-bearing rather than cosmetic. `bignum_mod.isNegative` returns `false` for anything that isn't a fixnum or bignum (bignum.zig:1158-1164), so a future "simplification" to a bare `bignum_mod.isNegative(args[0])` would silently classify every negative rational as non-negative — and then the `else` branch's `exactSqrt` → `isqrtNonNegative(rat.numerator)` receives a negative fixnum and evaluates `@intFromFloat(@sqrt(-9.0))`, i.e. `@intFromFloat(NaN)`, which is a safety-checked panic in ReleaseSafe reachable from plain `(sqrt -9/4)`.

Something like `// isNegative is false for rational Values; the numerator test is what actually classifies them` would keep the next reader from flattening this.

baijum (commented):
## Review: verified, LGTM — two non-blocking nits posted inline

**Verification performed** (on a build of ea10802):

- **Behavior**: every row of the PR's table reproduces — `(sqrt -1)` → `+i` exact, `(eqv? (sqrt -4) +2i)` → `#t`, `(sqrt -9/4)` → `+3/2i`, `(* (sqrt -1) (sqrt -1))` → `-1` exact, `(sqrt -2)` / `(sqrt -1.0)` / `(sqrt -0.0)` unchanged, and the non-negative paths (`(sqrt 16)`, `(sqrt 2)`, `(sqrt 16/9)`, `(sqrt 10)`, flonums, exact complex args) are all bit-identical to before.
- **Suites**: `bash tests/scheme/run-all.sh` → 749 scheme files + R7RS 1395/0, zero failures; `zig build test -Dtest-filter=tests_numeric` green; the two modified compliance files pass standalone (46 and 123).
- **No collateral expectations**: grepped `tests/` and `lib/` for other negative-`sqrt` expectations — the only hit outside this PR is `(sqrt -8.0)` (flonum, untouched path). The R7RS suite's own `(inexact (sqrt -1))` stays green because it wraps the result in `inexact`.
- **Housekeeping**: `zig fmt --check` clean; no new bare `PrimitiveError.TypeError`; benchmark comparison is within noise.

**GC-safety trace** (the interesting part of this diff): the unrooted results of `negateExact`/`exactSqrt` are only ever passed to callees that root their own arguments before allocating — `allocComplex` (`rootArgs2` before `maybeCollect`, gc_alloc.zig:924) and `makeRationalReduced` (primitives_arithmetic.zig:170). `bignum_mod.negate`/`isNegative` both handle fixnums, including the `minInt(i48)` → bignum promotion. The CodeRabbit flag on `negateExact` is a false positive under this non-moving collector — replied on that thread with the evidence. I also like that the old float-based fixnum perfect-square check (`@intFromFloat(@sqrt(f))`, `ri*ri == n`) is gone in favor of the exact `isqrtNonNegative` — strictly more robust and one fewer duplicated path.

**Spec**: agree with the reading. R7RS §6.2.6's `(sqrt -1) ⟹ +i` example plus the section's exact-notation convention implies an exact result, and Chez returns exact `+1i` here. The PR is also upfront about flipping the #2166 suite's deliberately-recorded "out of scope" assertion, which is the right way to make that change visible.

**Inline nits** (comment-only, neither blocks merge):
1. src/primitives_numeric.zig:590 — the unrooted `root` → `allocComplex` handoff is safe only because `allocComplex` self-roots; a local slot or a half-line of comment would make that visible.
2. src/primitives_numeric.zig:581 — the `isRationalObj` special case in the sign test is load-bearing (`isNegative` returns `false` for rationals; flattening it makes `(sqrt -9/4)` hit `@intFromFloat(NaN)` in `isqrtNonNegative`); worth a pinning comment.

**Two follow-up notes**, not for this PR:
- `src/primitives_numeric.zig` is now 1959 lines vs. the 1500-line advisory policy (1927 before). If/when it's split, `isqrtNonNegative` + `exactSqrt`/`negateExact` are a natural seam — the integer-sqrt machinery has essentially no coupling to the rest of the file.
- The end-user procedures reference lives in kaappi/kaappi.github.io; if it documents `sqrt`'s exactness behavior, this change deserves a matching line there.

src/primitives_numeric.zig
baijum: Done in c4b3034e: a comment at the call site now states that `root` is unrooted and safe only because `allocComplex` roots both arguments (`rootArgs2`) before `maybeCollect`, and that anything allocating inserted before this line must root it first. Kept it as a comment rather than a slot: the GC rules already say allocator Value arguments are auto-rooted, so a slot here would be redundant and would read as if the callee did not root.

src/primitives_numeric.zig
baijum: Done in c4b3034e: pinned with a comment spelling out that `bignum_mod.isNegative` answers false for any non-fixnum/bignum Value, so flattening the sign test would classify every negative rational as non-negative and route `(sqrt -9/4)` into `isqrtNonNegative` with a negative numerator, where `@intFromFloat(@sqrt(-9.0))` is a ReleaseSafe panic.

baijum:
Both inline nits addressed in c4b3034e (comments only, no behaviour change); replied on each thread.

On the docs follow-up: the end-user reference at kaappi.github.io `docs/procedures/numbers.md` already shows `(sqrt -1) ;=> +i` in its example block, so it was ahead of the implementation (which printed `+1.0i`) and this PR makes that line true rather than needing a new one. Its prose says exact *integer* perfect squares give exact results, which has been narrower than reality since #1412 (rationals too), but that predates this PR.

The file-size note is acknowledged: `primitives_numeric.zig` is at 1970 lines; the integer-sqrt cluster (`isqrtNonNegative`, `exactSqrt`, `negateExact`, `exact-integer-sqrt`) is the natural seam for a later split.

PR #2506, opened 2026-09-03
Stop exporting sqrt from (scheme base)

### LINKED ISSUE #2504: (scheme base) exports sqrt, which R7RS reserves to (scheme inexact): user libraries exporting their own sqrt fail to load
## Problem

R7RS marks `sqrt` as an "inexact library procedure" (§6.2.6, errata-corrected PDF p. 38): it is exported by `(scheme inexact)`, not `(scheme base)`. Appendix A confirms `(scheme base)` exports `exact-integer-sqrt` but not `sqrt`.

Kaappi's spec table exports `sqrt` from `(scheme base)` as well:

```zig
// src/primitives_numeric.zig:97
.{ .name = "sqrt", ..., .libs = LS.initMany(&.{ .scheme_base, .scheme_inexact, .scheme_r5rs }) },
```

It is the only `(scheme inexact)` procedure with this extra tag; `exp`, `log`, `sin`, `cos`, `tan`, `asin`, `acos`, `atan`, `finite?`, `infinite?`, `nan?` are all base-free. The `(scheme r5rs)` tag is correct (R5RS does define `sqrt`).

## Why it matters

Since v0.22.0 Kaappi enforces R7RS 5.2 (importing one identifier from two libraries with different bindings is an error). A portable program that is legal under R7RS therefore fails to load:

```scheme
;; mylib.sld — legal: (scheme base) does not export sqrt, so no redefinition
(define-library (mylib)
  (import (scheme base))
  (export sqrt)
  (begin (define (sqrt x) 'lib-mine)))

;; program
(import (scheme base) (mylib))
(sqrt 4)
```

```text
error[KP2001]: identifier 'sqrt' is imported with different bindings from both
(scheme base) and (mylib) -- R7RS 5.2: ...
```

The program runs on any conforming implementation. This is the same collision class as kaappi#1856 (the 22 `%`-prefixed `(scheme base)` exports), with a name a math library is far more likely to define — kaappi-mpl already defines a symbolic `sqrt` and works around the clash today by importing the built-in under a `rnrs:` prefix.

Redefining `sqrt` at a program's top level after importing only `(scheme base)` happens to work (Kaappi permits the redefinition), so the failure surfaces only through the library route.

## Suggested fix

Drop `.scheme_base` from the `sqrt` spec entry. Three in-repo tests currently reach `sqrt` through `(scheme base)` alone and would need `(scheme inexact)` added to their imports:

- `tests/scheme/smoke/expt-negative-base-1725.scm`
- `tests/scheme/srfi/srfi231-official.scm` (via `srfi231-official-transform.py`)
- `tests/scheme/compliance/complex-exactness-2166-2167.scm`

Ecosystem code that imports only `(scheme base)` and calls `sqrt` would break the same way, so this is a `breaking` change to a non-standard extension; the release note should say so. Add a regression test that loads a library exporting its own `sqrt` alongside `(scheme base)`.

## Priority

`medium`: wrong against the spec, but loud (fails at load with a message naming both libraries) and recoverable with `except`/`rename`.

baijum:
Closes #2504

## What

Drop the `.scheme_base` tag from the `sqrt` spec entry in `src/primitives_numeric.zig`. R7RS §6.2.6 makes `sqrt` an inexact library procedure: `(scheme inexact)` exports it, `(scheme base)` does not (Appendix A lists `exact-integer-sqrt` among the base exports, not `sqrt`). It was the only `(scheme inexact)` procedure carrying the extra tag; the `(scheme r5rs)` tag stays.

## Why

Since v0.22.0 Kaappi enforces R7RS 5.2, so a portable, spec-legal program that imports `(scheme base)` next to a user library exporting its own `sqrt` failed to load with KP2001. Same collision class as #1856, with a name a math library is far more likely to define (kaappi-mpl works around it today).

## Behaviour change (`breaking`, non-standard extension)

- `(scheme inexact)` and `(scheme r5rs)` still export `sqrt`, and library/program bodies still resolve an unimported `sqrt` through the globals map. Code that just calls `sqrt` after importing only `(scheme base)` keeps running.
- `sqrt` is no longer visible in a restricted environment built from `(scheme base)` alone: `(eval '(sqrt 4) (environment '(scheme base)))` is now KP3001, as R7RS specifies.
- Any import selector naming `sqrt` out of `(scheme base)` — `only`, `except`, or `rename` — is now KP2001 ("not found in import set"), which R7RS 5.2 also specifies. kaappi-mpl has four such sites (three libraries and its test file), all workarounds for exactly this collision; kaappi/kaappi-mpl#3 removes them. That PR needs this change, so the nightly `pure-libs` group is red for kaappi-mpl between this merging and that merging, and kaappi-mpl's own CI (pinned to the latest release) is red for it until the next release.

## Tests

- New `tests/scheme/smoke/sqrt-user-library-2504.scm` + `lib2504/mathlib.sld`: loads a library exporting its own `sqrt` alongside `(scheme base)` (the issue's reproduction, KP2001 before this change), checks `(scheme inexact)`/`(scheme r5rs)` still hand out the same built-in, `(scheme base)` keeps `exact-integer-sqrt`, and a `prefix` import gives a program both.
- `(scheme inexact)` added to the four in-repo tests that reached `sqrt` through base alone (`expt-negative-base-1725`, `complex-exactness-2166-2167`, `r7rs-control-io-gaps`, `srfi231-official` and its generator script). The smoke test also pins that `(environment '(scheme base))` no longer resolves `sqrt`. The SRFI 231 generator now emits the #2461 known-divergence note it had drifted from, so regeneration is a no-op again.
- `zig build test`: 1918 pass, 21 skip. `tests/scheme/run-all.sh`: 2144 pass, 0 fail (R7RS suite 1395/1395). kaappi-mpl's suite with kaappi/kaappi-mpl#3 applied: 226/226 against this branch's binary.

### DIFF
--- a/tests/scheme/smoke/sqrt-user-library-2504.scm
+++ b/tests/scheme/smoke/sqrt-user-library-2504.scm
@@ -30,6 +30,13 @@
              (eq? (eval 'sqrt (environment '(scheme inexact)))
                   (eval 'sqrt (environment '(scheme r5rs)))))
 
+;; The other half of the change: a (scheme base) environment no longer
+;; resolves sqrt at all. Nothing else in the suite would notice the tag
+;; silently coming back — library and program bodies fall through to the
+;; globals map, so only a restricted environment can tell.
+(test-error "(scheme base) no longer resolves sqrt"
+            (eval 'sqrt (environment '(scheme base))))
+
 ;; (scheme base) keeps exact-integer-sqrt — that one really is a base export.
 (test-equal "(scheme base) exports exact-integer-sqrt"
             '(4 1) (eval '(call-with-values (lambda () (exact-integer-sqrt 17)) list)
--- a/tests/scheme/smoke/sqrt-user-library-2504.scm
+++ b/tests/scheme/smoke/sqrt-user-library-2504.scm
@@ -1,0 +1,49 @@
+;; Regression test for #2504: (scheme base) exported sqrt, which R7RS reserves
+;; to (scheme inexact) (§6.2.6; Appendix A lists exact-integer-sqrt, not sqrt,
+;; among the base exports). Since kaappi enforces R7RS 5.2 (importing one
+;; identifier from two libraries with different bindings is an error), a
+;; portable program that imported (scheme base) together with a user library
+;; exporting its own sqrt failed to load with KP2001 — the same collision
+;; class as #1856's %-prefixed base exports, with a name a math library is far
+;; more likely to define.
+;;
+;; The fixture library lives in lib2504/ and is found via the script's own
+;; directory on the library search path.
+(import (scheme base) (scheme write) (scheme process-context) (srfi 64)
+        (lib2504 mathlib))
+
+(test-begin "sqrt-user-library-2504")
+
+;; The program loads at all — this is the issue's own reproduction: with
+;; (scheme base) no longer exporting sqrt, the two imports do not collide.
+(test-equal "user sqrt is the user's" '(lib-mine 4) (sqrt 4))
+(test-equal "library-internal caller sees the library's sqrt"
+            '(lib-mine 25) (hypot-ish 3 4))
+
+;; The built-in is still reachable from the libraries R7RS and R5RS assign it
+;; to, and is the same procedure via both.
+(test-equal "(scheme inexact) exports sqrt"
+            2 (eval '(sqrt 4) (environment '(scheme inexact))))
+(test-equal "(scheme r5rs) exports sqrt"
+            3 (eval '(sqrt 9) (environment '(scheme r5rs))))
+(test-assert "inexact and r5rs sqrt are one binding"
+             (eq? (eval 'sqrt (environment '(scheme inexact)))
+                  (eval 'sqrt (environment '(scheme r5rs)))))
+
+;; (scheme base) keeps exact-integer-sqrt — that one really is a base export.
+(test-equal "(scheme base) exports exact-integer-sqrt"
+            '(4 1) (eval '(call-with-values (lambda () (exact-integer-sqrt 17)) list)
+                         (environment '(scheme base))))
+
+;; Renaming around the user library still reaches the built-in, so a program
+;; that wants both can have both.
+(test-equal "prefixed built-in alongside the user's"
+            '(2 (lib-mine 4))
+            (eval '(list (in:sqrt 4) (sqrt 4))
+                  (environment '(scheme base)
+                               '(prefix (scheme inexact) in:)
+                               '(lib2504 mathlib))))
+
+(let ((runner (test-runner-current)))
+  (test-end "sqrt-user-library-2504")
+  (when (> (test-runner-fail-count runner) 0) (exit 1)))
--- a/tests/scheme/srfi/srfi231-official.scm
+++ b/tests/scheme/srfi/srfi231-official.scm
@@ -66,7 +66,7 @@ OTHER DEALINGS IN THE SOFTWARE.
 
 (import (srfi 231) (srfi 27) (srfi 143) (srfi 4)
         (srfi 160 u64) (srfi 160 s64)
-        (scheme base) (scheme write) (scheme char) (scheme cxr)
+        (scheme base) (scheme inexact) (scheme write) (scheme char) (scheme cxr)
         (scheme file) (scheme read) (scheme lazy)
         (scheme process-context))
 
--- a/src/primitives_numeric.zig
+++ b/src/primitives_numeric.zig
@@ -94,7 +94,7 @@ pub const specs = [_]primitives.PrimSpec{
     .{ .name = "inexact", .func = &inexactFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_base) },
     .{ .name = "expt", .func = &exptFn, .arity = .{ .exact = 2 }, .libs = LS.initMany(&.{ .scheme_base, .scheme_r5rs }) },
     .{ .name = "square", .func = &squareFn, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_base) },
-    .{ .name = "sqrt", .func = &sqrtFn, .arity = .{ .exact = 1 }, .libs = LS.initMany(&.{ .scheme_base, .scheme_inexact, .scheme_r5rs }) },
+    .{ .name = "sqrt", .func = &sqrtFn, .arity = .{ .exact = 1 }, .libs = LS.initMany(&.{ .scheme_inexact, .scheme_r5rs }) },
     .{ .name = "exact-integer-sqrt", .func = &exactIntegerSqrt, .arity = .{ .exact = 1 }, .libs = LS.initOne(.scheme_base) },
     .{ .name = "sin", .func = &sinFn, .arity = .{ .exact = 1 }, .libs = LS.initMany(&.{ .scheme_inexact, .scheme_r5rs }) },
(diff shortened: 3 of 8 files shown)


tests/scheme/smoke/sqrt-user-library-2504.scm
```diff
@@ -0,0 +1,49 @@
...
+(test-begin "sqrt-user-library-2504")
+
+;; The program loads at all — this is the issue's own reproduction: with
+;; (scheme base) no longer exporting sqrt, the two imports do not collide.
+(test-equal "user sqrt is the user's" '(lib-mine 4) (sqrt 4))
+(test-equal "library-internal caller sees the library's sqrt"
+            '(lib-mine 25) (hypot-ish 3 4))
+
+;; The built-in is still reachable from the libraries R7RS and R5RS assign it
+;; to, and is the same procedure via both.
```
baijum, line 24: Nice coverage of the positive side. Consider also pinning the *negative* — that a `(scheme base)` environment no longer resolves `sqrt` at all — since that's the other half of this change and nothing else in the suite would catch a regression that silently re-adds the tag:

```scheme
(test-error "(scheme base) no longer resolves sqrt"
            (eval 'sqrt (environment '(scheme base))))
```

Verified against this branch: passes (KP3001 from `eval`), and fails on `main` where it resolves to the builtin — so it's a genuine regression pin, not a tautology.

tests/scheme/srfi/srfi231-official.scm
```diff
@@ -66,7 +66,7 @@ OTHER DEALINGS IN THE SOFTWARE.
 
 (import (srfi 231) (srfi 27) (srfi 143) (srfi 4)
         (srfi 160 u64) (srfi 160 s64)
-        (scheme base) (scheme write) (scheme char) (scheme cxr)
+        (scheme base) (scheme inexact) (scheme write) (scheme char) (scheme cxr)
```
baijum, line 69: Nit: this line in the generated file was hand-edited rather than regenerated — the transform script's own header says "The generated file must never be hand-edited — every patch belongs here... commit the regenerated suite together with the change."

The edit is content-identical to what the generator produces (I re-ran `python3 tests/scheme/srfi/srfi231-official-transform.py` on this branch; the only difference vs the committed file is an unrelated 13-line block, not this line), so nothing is wrong with the change itself. But re-running the generator surfaced a pre-existing drift worth knowing about: the committed `.scm` contains a known-divergence note ("Entries 737-741 lived here until kaappi#2454...") that the current script does not emit — #2461 added it directly to the `.scm` without a matching generator change (e8070ab0 touches the `.scm`, not the `.py`). So since #2461, regenerating drops that note.

Suggest either regenerating here (accepting the note drop as exposing the drift) or leaving it and reconciling generator↔file in a follow-up — but worth doing one of the two before the next person trusts `git diff` after a regeneration.

baijum (commented):
## Review: verified, LGTM — one follow-up test import + two nits (inline)

**Spec**: agree. R7RS Appendix A puts `sqrt` in `(scheme inexact)` and only `exact-integer-sqrt` in `(scheme base)`; the `scheme_r5rs` tag is correctly kept (R5RS does define `sqrt`). After the change the entry is uniform with its siblings — `sin`/`cos`/`tan`/`asin`/`acos`/`atan`/`exp`/`log` are all `{inexact, r5rs}`, and I grepped every `scheme_inexact` tag in the tree to confirm `sqrt` was indeed the only one carrying `scheme_base`. The one-token diff is the whole mechanism: `library.zig:163-171` derives export sets from the spec tags, no second list.

**Verification** (local builds of both `main` and 975c913):

- **Issue repro**: user library exporting its own `sqrt` alongside `(import (scheme base))` — KP2001 collision on `main`, loads and returns `(lib-mine 4)` here. The new smoke test is a genuine regression test.
- **Behavior changes, all as described**: `(eval '(sqrt 4) (environment '(scheme base)))` goes `2` → KP3001 (correct per R7RS); `(scheme inexact)` → 2 and `(scheme r5rs)` → 3 still work and are the same binding; `exact-integer-sqrt` remains a base export; the globals fallback survives (unimported `sqrt` resolves in program bodies and the REPL).
- **Suites**: `tests/scheme/run-all.sh` → **2145 pass, 0 fail** (750 scheme files — the +1 over `main` is the new smoke test — plus R7RS 1395/0); `zig build test` green; `zig fmt --check` clean.
- **Collateral audit**: every other `sqrt` reference in `tests/` and `lib/` is fine — shipped SRFIs 94/144/194 import `(scheme inexact)` explicitly, SRFI 70 only mentions it in a comment, the FFI tests bind libm's `"sqrt"` through dlsym (a string, not the Scheme binding), and the rest use `exact-integer-sqrt`/`real-sqrt`/`flsqrt`/locally-defined names.

**One missed in-repo test** (same class as the three this PR fixed): `tests/scheme/compliance/r7rs-control-io-gaps.scm:43` runs the R7RS spec example `((compose sqrt *) 12 75)` while importing only `(scheme base)` + friends — no `(scheme inexact)`. It still passes via the globals fallback, which is exactly why it's easy to miss, but it now tests nothing about proper imports. Suggest adding `(scheme inexact)` to its import line in this PR (one line) — happy to see it deferred to a follow-up otherwise.

**Breaking-change note for the release note**: the commit body singles out `(except (scheme base) sqrt)`, but the surface is slightly wider — `(only (scheme base) sqrt)` also goes from working → KP2001 (verified on both builds), and `(rename (scheme base) (sqrt ...))` likewise. Since the release note is generated from the commit body, consider rewording to "any import selector naming `sqrt` out of `(scheme base)` (`only`/`except`/`rename`)" so kaappi-mpl-style code isn't surprised by the `only` case specifically.

**Inline nits** (neither blocks):
1. tests/scheme/smoke/sqrt-user-library-2504.scm:24 — suggest adding a `test-error` pin that `(environment '(scheme base))` no longer resolves `sqrt` (verified: passes here, fails on `main`).
2. tests/scheme/srfi/srfi231-official.scm:69 — this generated file was hand-edited instead of regenerated; re-running the generator is content-identical for this line but surfaces a pre-existing #2461 drift (a 13-line known-divergence note the generator no longer emits).

tests/scheme/smoke/sqrt-user-library-2504.scm
baijum: Added in 25db322b — with a comment explaining why only a restricted environment can catch the tag coming back (bodies fall through to the globals map). Test count goes 7 → 8; passes here.

tests/scheme/srfi/srfi231-official.scm
baijum: Went with reconciling here rather than dropping the note: ef107c53 ports the 13-line #2461 block into the generator's postlude, and `python3 tests/scheme/srfi/srfi231-official-transform.py` now reproduces the committed `.scm` byte for byte (verified: empty `git diff` after regeneration on this branch). The hand-edited import line stays as-is since it is identical to the generator's output.

baijum:
Addressed all four points in two follow-up commits (no amends):

- **Missed test** — `r7rs-control-io-gaps.scm` now imports `(scheme inexact)` (first commit).
- **Negative pin** — the smoke test asserts `(environment '(scheme base))` no longer resolves `sqrt` (first commit).
- **Release-note wording** — the first commit's body states the wider surface explicitly: any import selector naming `sqrt` out of `(scheme base)` (`only`/`except`/`rename`) goes from working to KP2001. I verified `only` and `rename` on this branch as you did. The squash message is built from the commit messages, so the correction rides along into the release note.
- **Generator drift** — reconciled in this PR (second commit): the #2461 note now comes from the generator, and regeneration is a byte-for-byte no-op.

Issue #820, opened 2026-07-02
Library re-registration frees lib_env still referenced by compiled closures — use-after-free crash on next call


baijum:
## Summary

Re-registering a library (a second `define-library` with the same name — e.g. reloading a library definition in a program or REPL session) frees the old library's per-library environment (`lib_env`) while closures compiled against that environment are still live and reachable through `vm.globals`. Calling such a closure afterwards dereferences the freed `StringHashMap` and crashes the interpreter.

This is distinct from #17 (dangling registry *key*, fixed): the current `LibraryRegistry.register` correctly replaces the key, but `gop.value_ptr.deinit()` also destroys the old library's `lib_env`, which outstanding closures still reference via `Function.env`.

## Location

Root cause in `src/library.zig`:

- `LibraryRegistry.register` (src/library.zig:71-78) calls `gop.value_ptr.deinit()` on replacement.
- `Library.deinit` (src/library.zig:34-43) then deinits and destroys `self.lib_env`.

The freed pointer is consumed in `vm_dispatch.zig` `.get_global` (line 132: `const env = func.env orelse &self.globals;` → `env.get(name)`), because `handleDefineLibrary` compiles library bodies with `compileExpressionInEnv`, storing `lib_env` into each compiled `Function`.

## Reproduction

```scheme
(define-library (foo bar)
  (import (scheme base))
  (export f)
  (begin
    (define secret 42)
    (define (f) secret)))
(import (foo bar))
(display (f)) (newline)          ; 42

(define-library (foo bar)        ; re-register: old lib_env is freed here
  (import (scheme base))
  (export g)
  (begin (define (g) 99)))

(display (f)) (newline)          ; old closure looks up "secret" in freed env
```

Actual (v0.11.1, ReleaseSafe):

```
42
thread ... panic: integer overflow
/opt/homebrew/Cellar/zig/0.16.0_1/lib/zig/std/hash_map.zig:977:42: ... in getIndex...
            const mask = self.capacity() - 1;
src/vm.zig:901:36: ... in runUntil
```

Expected: no crash — calling the previously imported `f` should still return 42 (or at minimum raise a Scheme-level error).

## Analysis

[...]

PR #2518, opened 2026-09-04
Roll back library registration when a .sld load fails (#2510)

### LINKED ISSUE #2510: A library that fails KP2001 on first import succeeds on the second import in the same process
## What happens

A library whose source has a stray trailing `)` after the `define-library`
form fails to load on the **first** `import` and loads successfully on the
**second** `import` in the same process.

## Reproduction

`lib/demo/broken.sld` — a well-formed library with one extra `)` at the end
of the file:

```scheme
(define-library (demo broken)
  (import (scheme base))
  (export answer)
  (begin
    (define (answer) 42)))
)
```

`run2.scm`:

```scheme
(import (scheme base) (scheme write) (demo broken))
(import (demo broken))
(display "answer = ")
(display (answer))
(newline)
```

```
$ kaappi --lib-path ./lib run2.scm
run2.scm:1: error[KP2001]: LibrarySourceReadError while loading library from ./lib/demo/broken.sld
answer = 42
```

The first import errors; the second one binds `answer` and the program runs
to completion. With only the first import, `answer` is unbound and the
program fails with `KP3001`.

## Why this matters beyond the stray paren

Whichever behaviour is correct, the two imports disagree, and the failing
one is the one a normal single-import consumer gets. A library can therefore
be broken for every real caller while a test file that happens to import it
twice reports success.

That is not hypothetical — it is how I found it. An automated coding agent
submitted a library with this exact defect plus a duplicate import in the
test file, ran the suite, saw

```
  4 passed, 0 failed
```

and honestly reported the tests as green. The error text went to stderr and
the file's own `(exit 0)` on zero failures set the exit code, so nothing in
the pipeline noticed. The library did not load for any single-import
consumer.

## Suggested resolution

Either is fine and both are better than the current state:

- the first import should succeed if the second one does (the reader appears
  to hold state across attempts); or
- the second import should fail the same way the first one did.

A `KP2001` arguably also deserves to be fatal rather than skippable, but the
inconsistency between attempts is the part that hides real breakage.

## Version

Kaappi Scheme v0.26.1, macOS ARM (Darwin 25.6.0).

baijum:
Closes #2510

## The bug

A `.sld` whose `define-library` form is well-formed but is followed by a read error (a stray `)`) failed the **first** `import` with `KP2001` and **succeeded** on the **second** `import` in the same process. Whichever behavior is correct, the two attempts disagreed — and the failing one is what every normal single-import consumer gets, so a library could be broken for all real callers while a test file that happened to import it twice reported green (exactly how the issue was found).

## Mechanism

`loadLibrarySource` (`src/vm_library.zig`) reads datums one at a time and dispatches each through `vm.handleTopLevelForm`. The `define-library` form was therefore dispatched — and the library **registered in `vm.libraries`** by `handleDefineLibrary`, exports and body included — *before* the reader reached the trailing garbage and the load returned `error.LibrarySourceReadError`.

The `.sbc` disk-cache layer already handled failure correctly (`endColdLoad(false, …)` writes nothing; `abortWarmLoad` drops the replay state and roots). The leak was the **in-memory registry**: the second import's registry short-circuit in `processImportSet` found the partial entry and served it as a good warm hit, never re-reading the source.

## What gets rolled back

- `loadLibrarySource` now opens a **rollback frame** (`beginLibSourceLoad`/`endLibSourceLoad`): `handleDefineLibrary` records each library name it registers while any frame is in flight (`vm.lib_rollback_names`/`lib_rollback_frames`, owned duped names). A load that **fails** unregisters everything recorded above its mark, so the next import re-attempts the load from disk and fails **identically** — no half-registered exports leak.
- Frames nest **one per `loadLibrarySource` invocation**: a nested load that *succeeded* commits its own registrations by closing its deeper frame before returning, so an outer failure never rolls back a good dependency it happened to pull in (a failing `(a)` that imported `(b)` leaves `(b)` intact and importable).
- **Successful loads are untouched** — the kaappi#1888 warm-replay caching and cold-capture paths behave exactly as before; the rollback fires only on error paths.
- `LibraryRegistry.unregister` (`src/library.zig`) deinit's the `Library` exactly as a replaced one is, **retiring the `lib_env` into `retired_envs` first** (the #820 rule): closures compiled in the failed load's begin blocks can escape into live structures and keep raw `Function.env` pointers into the env. Nothing is imported from a library on a failed load, so dropping its exports table strands no live binding.

The reader still rejects trailing garbage after `define-library` — deliberately; the fix makes the two attempts agree on that failure instead of tolerating it.

## Regression tests (fail without the fix, verified)

- **Zig unit test** `src/tests_libraries.zig` — "failed .sld load rolls back its registration so both imports fail (#2510)": broken `.sld` imported twice → `error.CompileError` both times and `vm.libraries.get(...) == null` after each; a clean sibling library still imports fine twice with its exports live. Verified failing without the fix (second import succeeded, registry entry present).
- **Shell suite** `tests/scheme/errors/library-load-rollback-2510.sh` (picked up by `run-all.sh`): single and double import of the broken library each report exactly one `LibrarySourceReadError` per attempt and exit non-zero with no `42` leaking; a clean library double-imports and prints `answer = 42`; a *good dependency nested inside* the broken file stays committed and usable after the sibling rollback. Verified failing without the fix (one error then `answer = 42`).

Judgment call: the broken `.sld` fixtures are **generated at runtime** into a temp dir (hermetic `KAAPPI_HOME` + `--lib-path`, per kaappi#2352) rather than committed — the fmt corpus test walks every `.scm`/`.sld` under `tests/scheme`, and an intentionally unparseable file would fail it. This matches the convention `fmt.sh` and the reader-error suites already use for malformed sources.

File sizes: `src/vm_library.zig` 1239 → 1292 (under the 1500 limit, no split needed). `src/vm.zig` was already over the advisory limit before this change (1581 → 1605) — pre-existing, noted here rather than widening this PR with a split.

## Test evidence (all green)

- `zig build test` — full unit suite, exit 0
- `zig build test -Dgc-stress=true` — full unit suite under collection-on-every-allocation, exit 0 (rollback path is allocation-light but exercises `unregister`'s retire-before-remove ordering)
- `bash tests/scheme/run-all.sh` — **750 pass, 0 fail** (1 skipped), R7RS **1395 pass, 0 fail**, exit 0; the new suite appears as `PASS tests/scheme/errors/library-load-rollback-2510.sh`
- `bash tools/run-r7rs-suite.sh zig-out/bin/kaappi` — 1395 pass, 0 fail
- `bash tests/scheme/fmt/fmt.sh` — 38 pass, 0 fail (corpus re-checked after moving fixtures out of the tree)
- Issue repro with the fix: both imports report `error[KP2001]: LibrarySourceReadError …`, `answer` is `KP3001` undefined, exit 1.


baijum (commented):
## Review

Nice fix — the diagnosis (registry short-circuit serving a half-registered library as a warm hit) is exactly right, and the implementation gets the hard parts correct:

- The retire-before-remove ordering in `unregister` respects the #820 rule, and the append-failure leak is called out deliberately rather than "fixed" into a dangling pointer.
- Remove-before-deinit with the key-aliasing comment is correct — the map key *is* `lib.owned_name`, and `StringHashMap.remove` doesn't free keys, so ordering is the only thing that matters here.
- Nested-frame semantics are right: a nested load resolves its own fate (commit or rollback) before returning, so an outer failure never destroys a good dependency — and both the Zig test and the shell suite actually cover that case, not just the headline regression.
- Generating the broken `.sld` fixtures at runtime instead of committing an unparseable file (fmt corpus walks everything under `tests/scheme/`) is the right call and matches the reader-error suite convention.

One residual gap in the stated invariant and a few nits inline.

src/vm.zig
```diff
@@ -492,6 +492,24 @@ pub const VM = struct {
...
+    /// .sld — and the second import's registry short-circuit then served
+    /// that half-loaded library as a success. `lib_rollback_frames` holds
+    /// one mark per in-flight `loadLibrarySource` invocation (the length of
+    /// `lib_rollback_names` when it started); `handleDefineLibrary` appends
+    /// each name it registers while any frame exists. A failed load
+    /// unregisters everything above its mark; a successful one merely
+    /// forgets the names — nested loads pushed their own, deeper marks and
+    /// resolved their own fate before returning, so their committed
+    /// registrations are never above a still-live outer mark. The name
+    /// strings are owned (duped) so the list never borrows from a Library it
```
baijum: Nit: the `if (items.len > 0)` guard before popping silently absorbs frame/name imbalance instead of surfacing it. `beginLibSourceLoad`/`endLibSourceLoad` are strictly paired by construction (the defer is installed immediately after a successful append), so an empty stack here would mean a real bug — an unconditional pop (or an `std.debug.assert`) would catch that; the guard turns it into quiet, hard-to-notice drift.

src/library.zig
```diff
@@ -134,6 +134,38 @@ pub const LibraryRegistry = struct {
...
+        // entry is gone it keeps the env (and exports) marked; after the
+        // append, retired_envs does. An append failure (OOM) leaks the env
+        // deliberately — freeing it would dangle the Function.env pointers
+        // above.
+        if (lib.lib_env) |env| {
+            lib.lib_env = null;
+            self.retired_envs.append(self.allocator, env) catch {};
+        }
+        // Remove before deinit: the map key aliases lib.owned_name, so once
+        // deinit frees it nothing (not even remove's key comparison) may read
```
baijum, line 245: Worth a line in this doc comment or in `docs/dev/thread-value-sharing.md`: `VM.initForThread` struct-copies the registry, so `libraries` and `retired_envs` are shared with SRFI-18 child threads — a failed load in one thread now removes entries and appends to `retired_envs` (which can realloc) while another thread may hold a `*Library` from `get`/`processImportSet`'s short-circuit. `register`-replacement already had that hazard class, so this isn't a regression, but `unregister` is the first shared-mutator that *deinit's a Library* another thread can be holding a pointer into, so it deserves an explicit note.

tests/scheme/errors/library-load-rollback-2510.sh
```diff
@@ -0,0 +1,175 @@
...
+
+# --- the regression itself: double import must fail IDENTICALLY ---
+# Before the fix the second import succeeded via the registry short-circuit
+# and the program printed "answer = 42" after the error.
+cat > "$TMPDIR_TESTS/broken-twice.scm" << 'EOF'
+(import (rollback2510 broken))
+(import (rollback2510 broken))
+(display "answer = ") (display (answer)) (newline)
+EOF
+assert_fails_with "broken .sld, double import: both attempts fail KP2001" \
```
baijum, line 157: Nit: this scenario runs the binary twice — once inside `assert_fails_with`, once here for stdout. Could capture once with stdout/stderr split (`out`/`err`) and assert both from a single run; not worth blocking on, but it halves the runtime and removes the (small) chance the two runs disagree.

src/vm_library.zig
```diff
@@ -278,6 +316,7 @@ fn loadLibrarySource(vm: *VM, source: []const u8) !void {
             _ = try vm.runTopLevelFunction(func);
         }
     }
+    load_ok = true;
```
baijum: Residual window of the exact #2510 shape, one frame up: the rollback frame covers `loadLibrarySource`, but `tryLoadLibraryFromFile` can still fail *after* a successful load — `try lcc.endWarmLoad(...)` (currently line 589) returns `CompileError` on a replay event-log desync (`endWarmLoad` in vm_library_cache.zig). At that point the library is registered, the rollback names were already freed by this successful load's `endLibSourceLoad`, and `abortWarmLoad` is not called (its catch only wraps `loadLibrarySource`). A retrying import then short-circuits off the registry and succeeds — first attempt fails, second succeeds: the very inconsistency this PR exists to kill.

The desync path is documented as near-impossible (validated hashes; it means a loader/serializer bug), so this is narrow — but the PR description states "both attempts fail identically" as the invariant. Either cover it (e.g. treat a post-load failure as `endLibSourceLoad(vm, mark, false)` before the defer commits, by lifting the frame to `tryLoadLibraryFromFile`) or add a comment at the `endWarmLoad` call naming it as the one known residual path, so the invariant claim is accurate.

baijum: Took the covering fix. The frame now opens at the top of `tryLoadLibraryFromFile` and commits only past the last fallible step — `load_ok = true` sits immediately after the warm/cold if/else (commit 989e9e5f) — so the `try lcc.endWarmLoad(...)` desync commits as a failed frame and unregisters the load, as does every other error return from the function. The nesting semantics did not break, so the alternative was not needed: each `tryLoadLibraryFromFile` invocation pushes its own frame exactly as each `loadLibrarySource` did before, and a nested successful load still commits before returning — an outer failure still never rolls back a good dependency (the shell suite's nested-dependency scenario re-verifies that). One consequence of the lift worth naming: the embedded/bundled early-exit success paths now set `load_ok` before their `return`; otherwise those clean loads would have been rolled back. The good-library-twice scenarios in both the unit test and the shell suite pin that commit path (a forgotten `load_ok` on the disk exit fails the registry-get assertion).

Testing story: I could not drive `endWarmLoad`'s failure from a test, and I believe it is genuinely not drivable. Every input the structure walk consumes is hash-validated — `beginWarmLoad` checks the source hash, `recordsValid` the include/dep hashes, and feature-dependent `cond-expand` bails the entry — so the number of events the walk consumes is by design a function of hash-gated inputs. The only remaining way to get `pos != events.len` without a mid-load desync error is a `.sbc` whose event section legitimately parses but carries an extra event, which means hand-editing bytes past the functions/transformers sections, and no writer API produces such a file. So the fix is verified by inspection of the frame ordering: the `defer endLibSourceLoad` fires on every return from `tryLoadLibraryFromFile`, and `load_ok` is set only after `endWarmLoad` succeeds. What I could pin in `src/tests_libraries.zig`: the frame/name stacks return to empty after both the failed and the successful imports (a leaked frame would make later top-level `define-library` forms rollback-able by unrelated future failures), and a successful load still leaves the registry entry in place. One more property, for the invariant claim: the desync'd cache entry stays on disk, so a retry replays it and fails identically — "both attempts fail identically" now holds on this path too.

src/library.zig
baijum: Added in both places (commit 989e9e5f). `LibraryRegistry.unregister`'s doc comment now carries a cross-thread hazard paragraph, and `docs/dev/thread-value-sharing.md` gained a "**Library registry mutation across threads (kaappi#2510).**" paragraph right after the "Loading `(srfi 128)` safely across threads" one — same section, same struct-copied-registry subject, so it reuses that terminology. It states exactly the hazard you describe: the map removal, the `retired_envs` append (which can realloc), and the `deinit` all mutate storage shared via `VM.initForThread`'s by-value copy, while another thread may hold a `*Library` from `get`/`processImportSet`'s short-circuit; `register`-replacement already had this hazard class, and `unregister` is the first shared mutator that *deinit's a `Library`* another thread can be holding a pointer into. The practical rule stays the one the srfi-128 paragraph already gives: pre-load on the root before spawning, don't lazy-load from children.

src/vm.zig
baijum: Done — `endLibSourceLoad` now does `std.debug.assert(vm.lib_rollback_frames.items.len > 0)` followed by an unconditional pop (commit 989e9e5f). The pop itself stays memory-safe even where the assert is compiled out (`ArrayList.pop` returns null on empty), and the assert keeps Debug and ReleaseSafe — the default build — loud, per the "empty stack is a real bug" reasoning. The balance is if anything easier to see now: with the frame lifted to `tryLoadLibraryFromFile` (the other thread), begin and the defer-end are installed adjacently in one function, so every return path pops exactly the frame it pushed.

tests/scheme/errors/library-load-rollback-2510.sh
baijum: Done — the nested-dependency scenario now runs the binary exactly once (commit 989e9e5f): stdout is captured into a variable, stderr is redirected to a file under the test's temp dir (the same convention `unjoined-thread-teardown.sh` uses), and both the exit/KP2001-count assertions and the "dep answer = 42" stdout assertion read from that single run. Kept as two labeled checks so a failure still names which half broke. Suite result unchanged: 5/5.

src/vm_library.zig
```diff
@@ -228,7 +228,48 @@ pub fn srfiFeatureAvailable(vm: *VM, name: []const u8) bool {
...
+    // frame append in beginLibSourceLoad — so an empty stack here is a real
+    // bug, not a state to absorb. Assert rather than silently skipping the
+    // pop and letting frame/name imbalance drift (#2518 review).
+    std.debug.assert(vm.lib_rollback_frames.items.len > 0);
+    _ = vm.lib_rollback_frames.pop();
+    const allocator = vm.gc.allocator;
+    while (vm.lib_rollback_names.items.len > mark) {
+        const name = vm.lib_rollback_names.pop() orelse break;
+        if (!ok) vm.libraries.unregister(name);
+        allocator.free(name);
```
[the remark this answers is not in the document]
baijum: Valid and reachable — fixed in aa194af6.

**Reachability (confirmed, not hypothetical):** `processImportSet` only short-circuits for the *requested* name; `loadLibrarySource` then dispatches **every datum in the file**, and a `.sld` may contain a `define-library` for any name whatsoever. `handleDefineLibrary` has no already-registered guard, so its `register` replaces the existing entry (and `register`'s replacement branch deinits the prior immediately — it was unrecoverable). The sharpest drivable shape is exactly your example: a file that redefines `(scheme base)` and then has trailing syntax — and `(scheme base)` has no `.sld` to reload from (built-in only; `lib/` has no `scheme/` tree), so the plain-unregister rollback left the name permanently unregistered and every later `(scheme base)` import reported "library not found".

**Fix:** the journal entry is now `{ name, prior }` (`LibRollbackEntry`). `LibraryRegistry.take` detaches the displaced entry *before* the replacement registers (so `register`'s replacement branch never sees or destroys it); on rollback `LibraryRegistry.restore` puts the prior back — same map slot, its exports and `lib_env` live again, never retired — after releasing the replacement exactly as `unregister` releases one (`lib_env` **retired** per #820, since closures from the failed load's begin blocks can escape and hold `Function.env` pointers into it; exports and owned name freed). On commit the prior gets the release a committed replacement warrants (`releaseDetached`: env retired, rest freed) — the same end state `register`-replacement produces, just deferred to frame close. Records unwind LIFO, so each restore returns the registry to the state just before that registration, which also unwinds a replacement *chain* inside one file correctly. One wrinkle worth naming: a detached prior lives outside the map until its frame resolves, so `markVmRoots` and `isGcRootedEnvMap` now trace the journal — otherwise a mid-load collection would sweep exactly the exports a rollback must put back (verified green under `-Dgc-stress=true`).

**Regressions:** end-to-end unit test (fixture whose failing redefine-then-garbage sequence is exactly the shape — asserts both imports fail identically AND the built-in stays registered with its own exports, importable afterwards), a registry-level take/restore/releaseDetached ownership test, and a new shell scenario in `library-load-rollback-2510.sh` (driven via `(environment ...)` inside `guard`, since `import` is not an expression, so the program continues past the failure and proves the restored `(scheme base)` imports). Both the unit test and the shell scenario were verified to fail with restore neutered.

src/vm_library.zig
```diff
@@ -969,6 +1049,22 @@ pub fn handleDefineLibrary(vm: *VM, args: Value) VMError!Value {
...
+    // isn't rollback-able. With no load in flight (a top-level
+    // define-library in a program file or eval) nothing is recorded — a
+    // program file is never re-imported, so there is no second attempt whose
+    // answer must agree.
+    if (vm.lib_rollback_frames.items.len > 0) {
+        if (vm.gc.allocator.dupe(u8, lib_name)) |owned| {
+            vm.lib_rollback_names.append(vm.gc.allocator, owned) catch {
+                vm.gc.allocator.free(owned);
+            };
+        } else |_| {}
```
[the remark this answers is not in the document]
baijum: Fixed in aa194af6.

One location note first: the suppressed failure was at the *registration-time* journaling inside `handleDefineLibrary` (the `dupe`/`append` around lines 1061-1065) — `beginLibSourceLoad` itself has been fallible since the first round (`try lib_rollback_frames.append`, `try`-ed by the caller before its `defer endLibSourceLoad` exists, so begin-failure can never double-pop or skip a pop; the Debug assert added then guards that invariant).

The registration-site journaling is now atomic in `registerAndJournal`: the owned name copy and the journal-list capacity are reserved **before** the registry is touched, and the displaced entry (if any) is detached first, so the only fallible step remaining is the insert itself. On failure the registry is left exactly as it was (a displaced prior back in place, the new library unregistered and still caller-owned) and `OutOfMemory` propagates, failing the load — no registration can exist without a journal record that covers it. The old best-effort suppression is gone.

**Regression:** the journal allocates from the raw allocator, which `gc.oom_countdown` cannot see (#2435), so the test sweeps `memory.OomAllocator` across the import's raw allocation sequence (`tests_libraries.zig`, "rollback journaling is atomic under allocation failure"): every countdown position fails the import and must leave the registry exactly as before — the displaced built-in restored with its own exports, the broken library absent, no leaked frame or record. The sweep's backing allocator is `page_allocator`, deliberately: the load's OOM paths absorb some failures by design (e.g. `retired_envs` appends per #820), which the leak checker would misreport as failures of behavior this test doesn't claim; the ownership moves themselves are covered leak-checked by the registry-level unit test.

src/tests_libraries.zig
```diff
@@ -1284,3 +1284,73 @@ test "restricted environment does not leak globals into a closure (#1860)" {
...
+    // now wraps the whole tryLoadLibraryFromFile (#2518 review), so a
+    // success path that forgot to commit (load_ok left false) would roll
+    // the library back out and the final registry get would turn up null.
+    _ = try vm.eval("(import (rollback2510 good))");
+    _ = try vm.eval("(import (rollback2510 good))");
+    try std.testing.expectEqual(@as(i64, 42), types.toFixnum(try vm.eval("(answer)")));
+    try std.testing.expect(vm.libraries.get("rollback2510.good") != null);
+    try std.testing.expectEqual(@as(usize, 0), vm.lib_rollback_frames.items.len);
+    try std.testing.expectEqual(@as(usize, 0), vm.lib_rollback_names.items.len);
+}
```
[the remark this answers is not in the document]
baijum: Done in aa194af6 — the flagged test is ours (added in 702a35d4, extended in 989e9e5f). It now runs through `th.TestContext` (`ctx.init()`/`defer ctx.deinit()`, `ctx.vm.…`); the redundant explicit `setVMInstance` call went with it (`makeTestVM` already registers the instance). Direct registry/rollback-stack state checks are kept as-is, since those assert Zig-side internals `th.expectEval*` cannot express. The two new tests from this round (`"failed load that redefines a registered library restores the prior"`, `"rollback journaling is atomic under allocation failure"`) use `th.TestContext` from the start; the OOM sweep builds its VM manually because it needs the `OomAllocator`-backed GC rather than the helper's default.

src/vm.zig
```diff
@@ -492,6 +492,26 @@ pub const VM = struct {
...
+    /// `handleDefineLibrary` appends each name it registers while any frame
+    /// exists. A failed load unregisters everything above its mark; a
+    /// successful one merely forgets the names — nested loads pushed their
+    /// own, deeper marks and resolved their own fate before returning, so
+    /// their committed registrations are never above a still-live outer
+    /// mark. The name strings are owned (duped) so the list never borrows
+    /// from a Library it is about to remove. Per-VM, like the collector
+    /// stack (the shared `libraries` map is what gets rolled back).
+    lib_rollback_names: std.ArrayList([]const u8) = .empty,
+    lib_rollback_frames: std.ArrayList(usize) = .empty,
```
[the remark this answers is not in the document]
baijum: Not splitting in this PR — tracked as #2522 instead.

- The overage is pre-existing: `src/vm.zig` is 1581 lines on `main` (verified against `61be406f`); this PR's rollback fields add a few dozen on top of an already-over file, they did not push it over.
- The file-size policy is advisory by AGENTS.md's own enforcement map ("Files ≤ 1500 lines — Advisory only — this file, File size policy"), and vm.zig is *not* a flat list of independent functions, so the primitives-file exemption doesn't apply either — which is exactly why the split needs its own careful pass along real seams, not a mechanical carve landed inside a bugfix PR.
- A VM-core split is its own project: vm.zig publicly re-exports several `vm_*` modules, holds the struct definition every one of them reaches back into, and the move would touch import paths across the VM family — the kind of change that must not ride a ~500-line library-loader bugfix.

#2522 names the natural seams visible today (compiler/expander callback shims, GC root marking, error reporting/stack traces, the rollback journal, debugger support types) with acceptance criteria and a no-behavior-change requirement.

baijum:
The Debug-only red on the `test (ubuntu-latest, Debug)` leg was the new scenario in `tests/scheme/errors/library-load-rollback-2510.sh` ("failed redefine of `(scheme base)` caught as a read error, registry intact"). That scenario drives a failing library load through `(environment '(rollback2510 broken-redefine))` inside `guard` and asserts that **stderr stays empty** — and in Debug builds kaappi uses Zig's DebugAllocator, which detects and reports leaks at exit.

The leak is pre-existing on main (this branch never touched the function); the new test legitimately exposed it. `environmentFn` in `src/primitives_r7rs.zig` allocated a `StringHashMap` for the new environment and handed it to the GC'd `SchemeEnvironment` only on success — every error return abandoned it, namely the `importSetChecked` failure path this scenario hits and the `allocEnvironment` OOM path. DebugAllocator's `error(DebugAllocator): memory address ... leaked` report on stderr tripped the assertion. Release builds use `c_allocator` (no leak detection), which is why only the Debug leg was red.

Fixed with `errdefer` cleanup mirroring the sweep's exact order (deinit the entries, then destroy the map — gc_sweep.zig does the same once the Environment owns it); ownership still transfers to the Environment on the success return, so the success path is unchanged. The scenario itself is the regression test: it failed on the pre-fix Debug build with the leak report and now passes with stderr byte-for-byte empty — 7/7 in Debug, plus `tests/scheme/errors/exit-code.sh` (56/56, since the merge brings #2519). Full ReleaseSafe bar also green: unit tests, unit tests under `-Dgc-stress=true`, R7RS suite (1395/1395), and `tests/scheme/run-all.sh` (2148 pass, 0 fail — including #2519's `compile-failure-signals-2513.sh` and #2516's `char-ready-pipe-2511.scm`).

The branch now merges current main (#2516, #2519, #2520) — merge commit cd8fb65c, fix commit 3c470dc1.

PR #2519, opened 2026-09-04
Fail a script run that already reported an error, whatever it prints after

### LINKED ISSUE #2512: A script continues after an uncaught error, so a later (exit 0) reports success
## What happens

An uncaught runtime error while running a **script file** prints a
diagnostic, and then execution continues with the next top-level form. If
the script later calls `(exit 0)`, the process exits `0` and nothing in a
pipeline can tell that anything went wrong.

## Reproduction

```scheme
(display "before\n")
(define bad (vector-ref (vector 1 2) 99))
(display "after the error — this line still runs\n")
(exit 0)
```

```
$ kaappi ex.scm; echo "EXIT=$?"
before
ex.scm:3:25: error[KP3006]: vector-ref: index 99 out of range for length 2
    (define bad (vector-ref (vector 1 2) 99))
after the error — this line still runs
EXIT=0
```

I could not find a flag to change this. `--deny-warnings` covers `check`
lint findings, not runtime errors.

## Why it matters

Test runners are the common case, because the R7RS idiom is a file of
top-level `check` forms ending in `(if (> fail 0) (exit 1) (exit 0))`. An
error raised part-way through skips the rest of *that* form — so the
assertions in it never run and the failure counter never sees them — and
then the trailing `(exit 0)` reports success.

Two concrete costs, both from this week:

1. Our own suite once reported `0 failed` while the check it was named for
   had not executed: a procedure was called with its arguments swapped, the
   error scrolled past in stderr, and the file still exited 0. We now grep
   stderr for `error[KP` in the test runner and fail the suite on a hit.

2. Worse, because a machine acted on it. An automated coding agent
   submitted a library that fails to load (kaappi#2510), ran the project's
   test command, saw

   ```
     4 passed, 0 failed
   ```

   and exit 0, and honestly reported the tests as green. A pull request was
   opened on a branch whose library did not load for any consumer. The
   error was on stderr the whole time.

## Suggested resolution

Any of these would remove the trap:

- a non-zero exit status when any uncaught error was reported during the
  run, even if the script later exits 0 (an explicit `(exit 0)` arguably
  should still win, in which case:)
- an `--abort-on-error` / strict flag that stops the run at the first
  uncaught top-level error; or
- making script execution abort by default, with the current
  keep-going behaviour reserved for the REPL, where it is clearly right.

The REPL behaviour is not in question — carrying on after an error is what
a REPL is for. It is running a file non-interactively that surprises.

## Version

Kaappi Scheme v0.26.1, macOS ARM (Darwin 25.6.0).

### LINKED ISSUE #2513: A failed --compile prints "Compiled X -> Y" and leaves a truncated .sbc behind
## What happens

When `--compile` fails, kaappi still writes an output file — a partial one —
**and prints a success message**, while exiting non-zero.

## Reproduction

Any program that imports a C-FFI library whose shared object is not
installed. Here, a program importing `(kaappi http)` with
`libkaappi_net.dylib` moved out of `~/.kaappi/lib/`:

```
$ kaappi --lib-path ./lib --lib-path ../kaappi-json/lib \
         --lib-path ../kaappi-http/lib --lib-path ../kaappi-net/lib \
         --compile -o /tmp/repro.sbc mesthiri.scm
mesthiri.scm:7: error[KP2001]: ffi-open: dlopen(libkaappi_net, 0x0001): tried: ...
Compiled mesthiri.scm -> /tmp/repro.sbc
$ echo $?
1
```

Note the second line: `Compiled mesthiri.scm -> /tmp/repro.sbc`. That is
the success message, printed after the error, on a run that exits 1.

The file it names exists and is truncated:

```
$ ls -l /tmp/repro.sbc          # FFI libraries missing
104640 /tmp/repro.sbc
$ ls -l /tmp/good.sbc           # same source, libraries restored
307853 /tmp/good.sbc
```

Restoring the shared objects and recompiling gives exit 0 and 307,853
bytes. So the failed run wrote roughly a third of the bundle — the entry
point and whatever resolved before the failure.

## Why it matters

Both halves are traps, and they compound.

**The success message.** Anything reading stdout to decide whether the
build worked is told it worked. The exit code is right, so a `set -e`
script is safe, but a human watching the log sees "Compiled … -> …" scroll
past directly under an error and reasonably concludes the error was a
warning.

**The partial file.** A stale `.sbc` from a previous good build is silently
replaced by a broken one. A build step that compiles and then embeds the
bytecode — which is how a standalone binary is produced —
gets a plausible-looking artifact rather than a missing one, so the failure
surfaces much later, as a binary that starts and misbehaves rather than a
build that stopped.

Our release process defends against this by checking the byte size of the
`.sbc` as well as the exit code, which is not a check anyone should have to
invent.

## Suggested resolution

- Do not print `Compiled X -> Y` on a run that failed.
- Write the output atomically, or delete it on failure: a failed compile
  should leave no file rather than a partial one.

## Version

Kaappi Scheme v0.26.1, macOS ARM (Darwin 25.6.0).

baijum:
## Summary

The top-level driver recovers from uncaught read/compile/runtime errors and keeps going with the next top-level form, recording the failure in `script_had_error` — but two success signals ignored that flag, so a run could fail loudly on stderr and still look successful to whatever consumed it. This PR makes the flag win at both exit points.

Closes #2512
Closes #2513

## The invariant

A script run (file or stdin; never the REPL) that reported an uncaught read/compile/runtime error must not be able to signal success afterwards. The flag is set by the driver at every recovery site; the fix makes both success paths consume it.

## #2512 — `(exit 0)` after an error

Decided semantics (the flag wins, narrowly): when `script_had_error` is set, an explicit `(exit 0)` exits **1**; any **non-zero** explicit status stays exactly as given. Implemented in `primitives_r7rs.exitFn` via `effectiveExitCode` (0 upgraded to 1 only), after the `suppress_exit` check. REPL sessions never set the flag and are unaffected (guarded-error scripts still exit 0; clean `(exit 0)` still exits 0; `(exit 5)` after an error still exits 5).

## #2513 — failed `--compile` writing an artifact and printing success

`compileFile` now decides the artifact target up front; a run that ends with `script_had_error` set writes no artifact, prints no `Compiled X -> Y` line, and **unlinks any stale file at the exact `-o` / derived-`.sbc` target** (LIFO defer, best-effort, only that path). A failed compile leaves no file — not a truncated one, not a stale one a bundling step would embed. The native `kaappi compile` path was audited: it propagates read/compile errors before the link step and gates `Compiled` on the linker's exit status, so it needed no change.

## The `kaappi test` worker interaction (verified end to end)

`suppress_exit` returns **before** the flag upgrade, so the worker records the code the file actually asked for (`vm.exit_code` stays raw); the worker consumes the flag per file — `runWorkerFile` clears it before `runFile`, passes it into `emitResult`, and clears it after — and `resolveVerdict` decides the verdict, so workers still always exit 0 with pass/fail in the emitted JSON. `resolveVerdict` flips with the new rule (an `(exit 0)` can no longer waive a reported top-level error; the note now names both halves on the transcript), which keeps `kaappi test` agreeing with the plain-run exit status that `runner-agreement.sh` exists to hold together (kaappi#1903). Verified manually: an errored file with an `(exit 0)` epilogue reports `ERROR` + note under `kaappi test`, exit 1 for the run; a clean sibling still passes.

## Driver extraction

main.zig was 1508 lines (over the 1500-line policy) and the fix grew it to 1610, so the extraction landed as its own commit: `runFile` (with the bytecode-cache replay), `runStdin`, `compileFile`, `disassembleFile`, `runWorkerFile`, their private helpers, and the `compileFile` regression tests moved to `src/toplevel_driver.zig` (the module that already owned the diagnostics they report through and the flag they set). Pure code motion plus the namespace edits it forces; dev-doc references follow. **main.zig is now 676 lines**, toplevel_driver.zig 1325.

## Fail-without-fix evidence

With `git stash push -- src/` (fix removed, tests kept) and a rebuild, all three regression scripts fail; after `git stash pop` and rebuild, all pass:

- `tests/scheme/errors/compile-failure-signals-2513.sh` (new): without fix **3 pass, 4 fail** (failed compile prints `Compiled`, stale artifact at `-o` and derived target survives); with fix **7 pass, 0 fail**.
- `tests/scheme/errors/exit-code.sh`: without fix **3 fail** — `(exit 0)` after uncaught error exits 0 instead of 1, mid-file variant, stdin variant; with fix all green.
- `tests/scheme/test-runner/runner-agreement.sh`: without fix **2 fail** — both runners PASS a top-level-error-then-`(exit 0)` file instead of FAILing, and the new note is absent; with fix all green.

Also two Zig unit tests (run through `zig build test`'s binary): `compileFile` leaves no artifact / removes a stale one / control-writes on success, and `effectiveExitCode` upgrades only an explicit 0.

## Test evidence

- Both issue repros: `--compile` of a file importing a nonexistent library exits 1, prints no `Compiled` line, leaves no `/tmp/out.sbc`; the #2512 script exits 1 (was 0).
- Unit suite: 1942 tests in the `zig build test` binary — **1922 passed, 20 skipped, 0 failed**. (`zig build test`'s build-runner ↔ test-binary server protocol stalls deterministically in this macOS environment — the spawned binary idles in `readv` while the runner waits in `waitZigTest`, zero CPU, reproduced with piped and file-redirected output — so the compiled binary was executed directly; that binary is exactly the one the test step builds.)
- `bash tests/scheme/run-all.sh`: **750 scheme files pass, 0 fail; R7RS suite 1395 pass, 0 fail** (includes all three scripts above).
- `zig fmt --check` clean on all changed Zig files; markdownlint clean; no bare `PrimitiveError.TypeError` added.


src/primitives_r7rs.zig
```diff
@@ -129,6 +142,12 @@ fn exitFn(args: []const Value) PrimitiveError!Value {
             vm.exit_code = code;
             return types.VOID;
         }
+        // The flag wins over an explicit `(exit 0)` (kaappi#2512). Consulted
+        // here — process state the driver set, same as `suppress_exit` above —
+        // and after the suppress check, so the worker's recorded request keeps
+        // the code the file actually asked for and `resolveVerdict` applies
+        // its own semantics. REPL sessions never set the flag.
+        code = effectiveExitCode(code, toplevel_driver.script_had_error);
```
baijum, line 150: **`emergency-exit` still lets a flagged run signal success.** Verified on this branch:

```
$ printf '(import (scheme process-context))\n(car 1)\n(emergency-exit 0)\n' > ee.scm
$ kaappi ee.scm; echo EXIT=$?
ee.scm:2:1: error[KP3002]: type error in 'car': expected pair, got 1
EXIT=0
```

The PR's stated invariant is that a script run which reported an uncaught error must not be able to signal success afterwards, and this is the one remaining path that can. R7RS 6.14 says `emergency-exit` "communicates an exit value to the operating system in the same manner as `exit`", so applying the same rule here is consistent with the spec, not a stretch of it:

```zig
fn emergencyExitFn(args: []const Value) PrimitiveError!Value {
    std.process.exit(effectiveExitCode(exitCode(args), toplevel_driver.script_had_error));
}
```

Plus an `(emergency-exit 0)`-after-error case in `tests/scheme/errors/exit-code.sh`. The `effectiveExitCode` unit test can't catch this because the gap is the call site, not the function.

(Aside, pre-existing and out of scope: `emergency-exit` also ignores `suppress_exit`, so it terminates a `kaappi test` worker before `emitResult`. Not this PR's problem, but worth an issue if it isn't already one.)

src/test_runner.zig
```diff
@@ -176,14 +177,20 @@ const Verdict = struct {
...
-    if (exit_code == 0) return .{
-        .errored = false,
-        .note = if (toplevel_error)
-            "uncaught top-level error, acknowledged by an explicit (exit 0)"
+    // A reported top-level error makes the file errored, full stop: a plain
+    // run exits non-zero for it (the flag, or the explicit non-zero exit the
+    // file then made), and this runner must agree (kaappi#1903, kaappi#2512).
+    // An explicit `(exit 0)` used to waive it; since kaappi#2512 it cannot,
+    // but it still earns a note so the transcript names both halves.
+    if (toplevel_error) return .{
```
baijum, line 185: **Two consequences of attaching the note to an `errored` verdict.**

*The `noted` tally is now dead.* `resolveVerdict` produces a note in exactly two places — here and the unexplained-nonzero-exit branch below — and both now return `errored = true`. `accumulateAndReport` only counts a note when `!errored` (line 922), so `Totals.noted` can never be non-zero again, while the summary line and the `noted` JSON field still print it, and `docs/dev/test-runner.md` line 155 (edited in this PR) still says a note "is tallied as `noted` in the summary". Verified: the `ack.scm` fixture reports `Files: 1 run, 1 failed, 1 errored, 0 noted`.

*The note displaces the diagnostic in JSON.* `buildFileJson` writes `err_msg orelse verdict.note` into `error_message`, and the orchestrator's captured-output override only fires when `error_message == null`. So for the same top-level error:

```
unack.scm (no epilogue): "error":true,"error_message":"…:3: error[KP3000]: uncaught exception: deliberate-uncaught-top-level-error\n    (raise …)"
ack.scm   ((exit 0)):    "error":true,"error_message":"uncaught top-level error; an explicit (exit 0) cannot waive it (kaappi#2512)"
```

A JSON consumer reading `error_message` to learn *what* failed gets the KP code for one file and a sentence about `(exit 0)` for the other; the diagnostic itself is not in the JSON at all for the second. Text mode is fine because `reportFileText` prints both. The schema doc (`docs/dev/test-runner.md` line 191, "the diagnostic for an errored file") is also no longer always true.

Both go away with a separate `note` field on the file record: `error_message` keeps the diagnostic, `note` carries the verdict's sentence, `accumulateAndReport` tallies `note != null` regardless of `errored`, and the text reporter prints it as it does now. That is also what the CodeRabbit thread on this hunk is asking for from the tally side. If you'd rather not widen the schema in this PR, the minimum is to drop the "tallied as `noted`" claim from the doc and note the displacement at line 191 — but the summary column then stays permanently zero, which argues for removing it too.

docs/dev/test-runner.md
```diff
@@ -125,32 +126,35 @@ failure.
...
-A **note** is the channel for something a verdict deliberately tolerates. It
-travels in `error_message` (with `"error": false`), prints under the file's
-`PASS`/`FAIL` line in text mode along with whatever the worker wrote to
-stdout/stderr, and is tallied as `noted` in the summary. So a waived top-level
-error is visible without being fatal, rather than either failing the run or
-disappearing from it.
+A **note** is the channel for something a verdict wants on the record. It
+travels in `error_message`, prints under the file's `PASS`/`FAIL`/`ERROR` line
+in text mode along with whatever the worker wrote to stdout/stderr, and is
+tallied as `noted` in the summary. So the `(exit 0)` that failed to waive a
```
baijum: After this PR no note can reach the `noted` tally (see the `resolveVerdict` thread) — this sentence and the `error_message` bullet at line 191 both describe behaviour the runner no longer has.

src/toplevel_driver.zig
```diff
@@ -341,3 +377,949 @@ pub fn printSourceSnippet(source: []const u8, line: u32) void {
...
+        getSbcPath(allocator, path) catch {
+            writeStderr("Error creating output path\n");
+            script_had_error = true;
+            return;
+        };
+    defer allocator.free(sbc_path);
+    // Runs before the free above (LIFO): best-effort, and deliberately only
+    // the exact target — the error itself was already reported at its site,
+    // so a removal failure (permissions) has nothing further to add.
+    defer if (script_had_error) {
```
baijum: **Nit:** keying the unlink on `script_had_error` misses the paths that leave `compileFile` by error return — the `try printer.valueToString` at 1176, `return error.OutOfMemory` at 1179 and 1197. Those propagate to `mainInner`, which exits 1, but the flag is still false so this `defer` keeps the stale artifact at the target: exactly the "previous good build left looking current" case the comment above describes. It also makes the cleanup depend on process-wide state being false on entry rather than on what this call did.

A local set at the one success point covers every exit path uniformly:

```zig
var wrote_artifact = false;
defer if (!wrote_artifact) { … unlink … };
…
// after the bytecode_file.write* call succeeds:
wrote_artifact = true;
```

and the `if (script_had_error) return;` gate at 1210 stays as the thing that decides whether to write at all.

baijum (commented):
Reviewed both commits at 1e88575. Verified locally on macOS ARM: `zig build` clean; `compile-failure-signals-2513.sh` 7/7, `exit-code.sh` 53/53, `runner-agreement.sh` 10/10. I did not get a local `zig build test` result — the filtered run was still in its Debug build when I posted, so the unit-test claim rests on the PR's own evidence and CI. The extraction commit is pure code motion — I diffed the removed and added line multisets and the only deltas are the `compiler_mod`/`reader_mod` namespace renames, `pub` markers, imports, and the new header comment. The `suppress_exit`-before-upgrade ordering in `exitFn` is right and the worker's clear-around-`emitResult` matches the description.

Three things inline, in priority order:

1. **The invariant has a third exit point.** `emergency-exit` bypasses `effectiveExitCode`, so `(car 1)` followed by `(emergency-exit 0)` still exits 0. One-line fix plus a case in `exit-code.sh`.
2. **`Totals.noted` is now dead and the note displaces the diagnostic in JSON.** After this PR every note `resolveVerdict` produces rides on an `errored` verdict, and `accumulateAndReport` only tallies notes on non-errored files, so the summary's `noted` column can never be non-zero again. Separately, for the `ack.scm` shape the JSON `error_message` is now the note and the KP3000 diagnostic is gone from the JSON entirely (the no-`(exit 0)` sibling carries it). A separate `note` field fixes both — CodeRabbit's thread is asking for the same thing from the tally side.
3. **Nit:** the unlink `defer` keys on `script_had_error`, so the `try`-propagated OOM paths in `compileFile` exit 1 but leave the stale artifact. A local `wrote_artifact` flag covers every path.

One non-blocking note: this is a user-visible change to what `(exit 0)` means in a script, and nothing on kaappi-lang.org documents exit-status semantics today. Worth a line in the site's guide when the release goes out, alongside the changelog entry the commit body already provides.

baijum:
The "macOS environment/std anomaly" caveat in the description was wrong, and CI proved it: this was the new `compileFile` control test wedging the build runner itself, not a platform quirk.

**Root cause.** Under `zig build test` the unit-test binary speaks the build runner's server protocol (`std.zig.Server`, `--listen=-`) **over fd 1**. The control test drives a successful `compileFile`, whose `Compiled <src> -> <dst>` line is a raw `write(1, ...)`; the runner parses those bytes as a message header — `"Comp"` as the tag, `"iled"` as a ~1.7 GB body length — and blocks reading a body that never arrives. The plain test legs set no per-test timeout, so the block is forever at zero CPU (the QEMU legs surfaced it at their `--test-timeout 8m`). Running the binary *directly* — as the original session did locally, and as the windows/freebsd/openbsd/netbsd CI legs do — has no `--listen=-`, so fd 1 is an ordinary terminal and the same bytes are harmless. That direct-binary pass is exactly what masked the bug.

**Fix** (`b367020b`): `testing_helpers` gains `QuietStdout`, the fd-1 twin of `QuietStderr` (kaappi#2447); the control test parks fd 1 on the null sink for the compile. The `Compiled` line itself remains pinned end-to-end by `tests/scheme/errors/compile-failure-signals-2513.sh`; the unit test asserts the artifact and the flag. The failure-path test takes `QuietStderr` too, so its expected KP2001 diagnostics stop rendering the green run like a crash in every CI log.

Verified locally: full `zig build test` green **end-to-end through the runner protocol** (2014/2035 passed, 21 skipped), `zig build test -Dgc-stress=true` green (2018/2035, 17 skipped), the wasm test target compiles, and the three shell suites pass (`compile-failure-signals-2513.sh` 7/0, `exit-code.sh` 56/0, `runner-agreement.sh` 10/0).

The two follow-up commits address the review comments — `ced5de8b` (emergency-exit and the #2512 rule) and `de011ca1` (the verdict note as its own JSON field, tally fixed) — with details in the replies to each thread. The OOM sweep added for the `wrote_artifact` comment also caught and fixed a real leak in `GC.allocSymbol` (the interned name dupe leaked when the `Symbol` allocation OOM'd right after it).

src/primitives_r7rs.zig
baijum: Done in `ced5de8b`: `emergencyExitFn` now goes through `effectiveExitCode(exitCode(args), toplevel_driver.script_had_error)` — your snippet's shape exactly. `(car 1)` then `(emergency-exit 0)` exits 1; a clean `(emergency-exit 7)` stays 7, and `exit-code.sh` gained the after-error case plus two clean-run controls.

Re your aside on `suppress_exit`: filed as kaappi#2521 (`priority: medium`) — emergency-exit terminating a `kaappi test` worker before `emitResult` is pre-existing and out of scope here; the code comment in `emergencyExitFn` points at the issue.

src/test_runner.zig
baijum: Implemented your preferred fix, not the minimum (`de011ca1`): `note` is a field of the file record on both sides (`buildFileJson` and the parent's `writeFileObject`), `error_message` keeps the diagnostic — and since an errored file's worker no longer spends its message slot on the note, the orchestrator's captured-output tail now lands in `error_message` for it, so the KP diagnostic is present instead of displaced (your `ack.scm`/`unack.scm` asymmetry is gone). `accumulateAndReport`'s counting folded into `tallyFileResult`, which tallies `note != null` regardless of `errored`; the fixture now reports `"noted":1`. The text reporter prints the worker message and the note as separate lines where it printed one before.

Regression tests: `writeFileObject keeps the note out of error_message`, a `buildFileJson` round-trip through a real VM (`test collector setup failed` as diagnostic + the (exit 0) sentence as note, in their own fields), and `tallyFileResult counts a noted file whether or not it errored` — which fails against the old rule with `expected 1, found 0`.

src/test_runner.zig
```diff
@@ -176,14 +177,20 @@ const Verdict = struct {
...
+        .errored = true,
+        .note = if (exit_requested and exit_code == 0)
+            "uncaught top-level error; an explicit (exit 0) cannot waive it (kaappi#2512)"
```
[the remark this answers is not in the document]
baijum: Addressed via the note-field change in `de011ca1`: `note` is a separate field on the file record (worker and parent serializers), and `tallyFileResult` counts it even when `errored` — exactly the distinct-marker + tally fix asked for here, with ordinary error diagnostics staying in `error_message` and out of the count. Covered by `tallyFileResult counts a noted file whether or not it errored`, which fails against the old rule.

src/toplevel_driver.zig
baijum: Done with your shape in `b367020b`: local `var wrote_artifact = false;` with `defer if (!wrote_artifact) unlinkQuiet(sbc_path)` (the bufPrintZ/unlink extracted into a helper), set true only after the `bytecode_file` write succeeds; the `if (script_had_error) return;` gate stays deciding whether to write at all.

On coverage: `gc.oom_countdown` cannot drive those exits — all three are raw-allocator allocations, and its own doc comment notes "a compile with no macro use in it never fails here". The test reaches them anyway by wrapping `vm.gc.allocator` (a plain field) in `OomAllocator` (#2435) and sweeping its countdown with a stale artifact planted before every iteration: "compileFile: an error-return exit still leaves no artifact (#2513 OOM sweep)" asserts the target holds a file exactly when the run completed cleanly, and I verified it fails against the old flag-keyed defer (the stale 14-byte file survives an OOM error-return). Two things the sweep forced out beyond the nit itself: the `-o` arm's dupe-failure return happens before the defer is armed, so it now unlinks the known target itself (the derived-path arm genuinely has no name to unlink); and `GC.allocSymbol` leaked the interned name when the `Symbol` create OOM'd right after the dupe — fixed by allocating the object first, matching `allocUninternedSymbol`.

src/test_runner.zig
baijum: (Also resolves the docs comment on the "tallied as noted" sentence, id 3937714539 — no longer directly replyable since `de011ca1` rewrote that exact paragraph: the doc now describes the `note` field and the regardless-of-`errored` tally, as above.)

PR #2523, opened 2026-09-05
Honor suppress_exit in emergency-exit so a test worker still emits

### LINKED ISSUE #2521: emergency-exit ignores suppress_exit and kills a kaappi test worker before it can emit
Found in review of #2519 (inline thread `3937714536` on `src/primitives_r7rs.zig`, maintainer aside).

`(exit ...)` consults `vm.suppress_exit`, so a `kaappi test` worker turns a test file's `(exit 1)` failure epilogue into a recorded no-op and still reaches `emitResult`. `emergency-exit` (R7RS 6.14) does not — `emergencyExitFn` calls `std.process.exit` directly:

- A test file that calls `(emergency-exit 0)` / `(emergency-exit 1)` terminates the worker subprocess before it writes its `KAAPPI_TEST_EMIT` result.
- The orchestrator then sees a missing/empty result and reports `ERROR <file> (worker produced no result)` instead of whatever verdict the counters plus the emergency exit would have justified.

This is a documented-guarantee violation (the worker contract in `docs/dev/test-runner.md`: the worker always emits exactly one JSON object), but loud and narrow: nothing silently passes that should fail — the file reports as an error — and no user report of it so far. Out of scope for #2519, which only makes `emergency-exit` honor `script_had_error` like `exit` does (#2512).

Possible directions, whichever is decided should note that "skip everything, exit now" is emergency-exit's contract: either honor `suppress_exit` the way `exit` does (record the request, keep the VM usable, let the verdict machinery weigh it), or make `runWorkerFile`/the orchestrator treat a worker death by emergency-exit as a distinct outcome rather than "no result".

baijum:
Closes #2521

## The bug

`exit` consults `vm.suppress_exit`, so a `kaappi test` worker turns a file's `(exit 1)` failure epilogue into a recorded no-op and still reaches `emitResult` — the documented worker contract (docs/dev/test-runner.md): the worker always emits exactly one JSON object. `emergency-exit` did not — `emergencyExitFn` called `std.process.exit` directly, killing the worker subprocess before it wrote its `KAAPPI_TEST_EMIT` result. The orchestrator then reported:

```
ERROR file.scm  (worker produced no result)
```

losing every SRFI-64 count the file had already collected. A passing suite ending in `(emergency-exit 0)` was red under `kaappi test` and green under `tests/scheme/run-all.sh` — exactly the two-runner disagreement `runner-agreement.sh` exists to forbid (kaappi#1903). Reproduced on a684eeb3.

## The fix

`emergencyExitFn` (src/primitives_r7rs.zig) now consults `vm.suppress_exit` the same way `exitFn` does:

- **Under suppression** (a `kaappi test` worker): record the request (`vm.exit_requested` / `vm.exit_code` — the *raw* code, before the kaappi#2512 upgrade, so the worker's `resolveVerdict` weighs what the file actually asked for) and return `VOID`, keeping the VM usable so the worker reaches `emitResult`. A requested emergency exit code therefore influences the file's verdict exactly like a requested `(exit N)` does today.
- **Outside suppression**: unchanged — a real `std.process.exit` through `effectiveExitCode` (still honoring a reported script error per #2512), still skipping dynamic-wind cleanup and the port flush; that skip is emergency-exit's own contract (R7RS 6.14). Verified: a plain run of `(emergency-exit 3)` inside `dynamic-wind` exits 3 immediately, without running the after-thunk.

`src/test_runner.zig` and `src/toplevel_driver.zig` got only in-place comment rewraps (both are near/over the 1500-line cap); the diff stays in `src/primitives_r7rs.zig` plus docs and tests.

## Test evidence

- New unit test beside the primitive, `suppress_exit turns (emergency-exit) into a recorded no-op, VM stays usable` (src/primitives_r7rs.zig — the twin of the `exitFn` guard in test_runner.zig, placed there because test_runner.zig is already over the cap).
  - Without the fix: `zig build test -Dtest-filter="emergency-exit"` → build fails, "1 crashed" (the primitive killed the test binary).
  - With the fix: passes, and the full `zig build test` is green (exit 0).
- New end-to-end test `tests/scheme/test-runner/emergency-exit.sh` (picked up by `tests/scheme/run-all.sh`'s "Test runner" suite): three fixtures — a green suite ending `(emergency-exit 0)`, a failing suite ending `(emergency-exit 1)`, a bare unexplained `(emergency-exit 1)` — each run through a plain `kaappi <file>` and `kaappi test`, requiring both verdict agreement and the expected verdict, surviving counts on the transcript, and no "worker produced no result".
  - Without the fix: 2 pass, 6 fail (the transcripts show `ERROR ee-pass.scm (worker produced no result)`).
  - With the fix: 8 pass, 0 fail.
- Original repro after the fix: `kaappi test pass.scm` (passing SRFI-64 suite + `(emergency-exit 0)`) now prints `PASS ... (1 tests, 0 skipped, 1ms)` and exits 0; a bare `(emergency-exit 1)` yields `error: true` with the note `file requested a nonzero exit with no failing test to explain it` — the same verdict a suppressed `(exit 1)` gets.

## Docs

`docs/dev/test-runner.md` updated where it documents the suppression step (worker walkthrough step 2), the verdict inputs, and the test inventory.

### DIFF
--- a/docs/dev/test-runner.md
+++ b/docs/dev/test-runner.md
@@ -95,9 +95,11 @@ Inside the worker (`src/toplevel_driver.zig` `runWorkerFile` →
 2. **Suppress `(exit)`.** The worker sets `vm.suppress_exit`, so a file's
    `(exit 1)` failure epilogue becomes a *recorded* no-op (`vm.exit_requested`
    / `vm.exit_code`) instead of terminating the worker before it can emit its
-   result. Because the call never happens, its *semantics* are reapplied when
-   the verdict is resolved — see [Verdicts](#verdicts-and-what-errored-means)
-   below.
+   result. `(emergency-exit …)` records through the same channel — a plain
+   `emergencyExitFn` used to kill the worker mid-file, and the orchestrator
+   reported the file as a missing result (kaappi#2521). Because the call never
+   happens, its *semantics* are reapplied when the verdict is resolved — see
+   [Verdicts](#verdicts-and-what-errored-means) below.
 3. **Run the file** via the normal `runFile` path (so the `.sbc` cache, imports,
    and error diagnostics all behave exactly as a plain run).
 4. **Emit one JSON object** for the file to `KAAPPI_TEST_EMIT`, built by walking
@@ -119,7 +121,8 @@ to a failure of a test. SRFI-64 catches ordinary test failures internally via
 
 Three inputs decide it (`resolveVerdict` in `src/test_runner.zig`): whether an
 uncaught read/compile/runtime error was reported at top level, what the file's
-suppressed `(exit)` asked for, and whether the SRFI-64 counters already show a
+suppressed `(exit)` — or `(emergency-exit …)`, the same channel since
+kaappi#2521 — asked for, and whether the SRFI-64 counters already show a
 failure.
 
 | The file… | `error` | Why |
@@ -361,7 +364,8 @@ a transcript diff between two runs stays meaningful at any job count.
 
 - `src/test_runner.zig` unit tests — JSON serialization round-trips, the
   SRFI-64 discovery gate, and the `suppress_exit` behaviour (the guard on the
-  `exitFn` change in `src/primitives_r7rs.zig`).
+  `exitFn` change in `src/primitives_r7rs.zig`; the `emergencyExitFn` twin of
+  that guard lives beside the primitive it protects).
 - `src/test_selection.zig` unit tests — import-spec unwrapping, the
   native-artifact classifier, lexical path canonicalisation, and the closure
   BFS over synthetic graphs (diamond deps, incomplete edges, cycles).
@@ -391,6 +395,12 @@ a transcript diff between two runs stays meaningful at any job count.
   bug class #2116 exists to forbid, and every file in the corpus that can print
   `FAIL` also exits nonzero when it does — which is why all ten fixtures still
   agree. If `kaappi test` ever grows the same net, the asymmetry closes.
+- `tests/scheme/test-runner/emergency-exit.sh` — `(emergency-exit …)` under a
+  worker honors `suppress_exit` (kaappi#2521): each of three fixtures (a green
+  suite exiting 0, a failing suite exiting 1, a bare unexplained exit 1) runs
+  through a plain `kaappi <file>` and `kaappi test`, which must agree and land
+  on the expected verdict, with the counts on the transcript and no
+  "worker produced no result".
 - `tests/scheme/test-runner/changed.sh` — `--changed`/`--list-affected` over a
   throwaway git repo with a known dependency shape: diamond import, `include`,
   a `(load …)` escape hatch, native-artifact and unknown-revision full-run
--- a/src/primitives_r7rs.zig
+++ b/src/primitives_r7rs.zig
@@ -170,14 +170,26 @@ fn exitFn(args: []const Value) PrimitiveError!Value {
 }
 
 fn emergencyExitFn(args: []const Value) PrimitiveError!Value {
+    const code = exitCode(args);
+    if (vm_mod.vm_instance) |vm| {
+        // `kaappi test` worker: same suppression contract as `exit` (kaappi#2521).
+        // Terminating here would kill the worker before it emits its one JSON
+        // result, so the orchestrator reports the file as "worker produced no
+        // result" — losing the counts the file already collected. Record the
+        // request and return instead: the raw code, before the #2512 upgrade
+        // below, so the worker's `resolveVerdict` weighs what the file actually
+        // asked for, exactly as it does a suppressed `(exit N)`.
+        if (vm.suppress_exit) {
+            vm.exit_requested = true;
+            vm.exit_code = code;
+            return types.VOID;
+        }
+    }
     // Same #2512 rule as `exit`: an emergency exit from a run that already
     // reported an uncaught top-level error may not mask it with an explicit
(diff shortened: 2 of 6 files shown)


src/primitives_r7rs.zig
```diff
@@ -170,14 +170,26 @@ fn exitFn(args: []const Value) PrimitiveError!Value {
...
+    const code = exitCode(args);
+    if (vm_mod.vm_instance) |vm| {
+        // `kaappi test` worker: same suppression contract as `exit` (kaappi#2521).
+        // Terminating here would kill the worker before it emits its one JSON
+        // result, so the orchestrator reports the file as "worker produced no
+        // result" — losing the counts the file already collected. Record the
+        // request and return instead: the raw code, before the #2512 upgrade
+        // below, so the worker's `resolveVerdict` weighs what the file actually
+        // asked for, exactly as it does a suppressed `(exit N)`.
+        if (vm.suppress_exit) {
```
baijum, line 182: Non-blocking, but this comment should say the one thing that is *different* about suppressing `emergency-exit` versus `exit`: under suppression the call returns, so control leaves any enclosing `dynamic-wind` extent normally and the **after-thunks run** — the exact thing a real `emergency-exit` skips (and `tests/scheme/errors/exit-wind.sh` guards for the plain run). Verified on this head:

```scheme
(import (scheme base) (scheme process-context) (srfi 64))
(test-begin "wind")
(dynamic-wind (lambda () #f)
  (lambda () (emergency-exit 0))
  (lambda () (test-assert "after-thunk ran" #f)))
(test-end "wind")
```

Plain run: exit 0 (after-thunk skipped). `kaappi test`: `FAIL wind.scm (1/1 failed) - after-thunk ran: got #f`.

That's the same class of divergence `exit` already has for forms *after* the call (they run in the worker, not in a plain run), so I don't think it needs fixing — a worker that must reach `emitResult` cannot honor "skip everything" — but a sentence here saying "under suppression the winds do run; that's accepted, not a bug" will stop someone from reading the R7RS 6.14 comment below and 'fixing' it.

docs/dev/test-runner.md
```diff
@@ -95,9 +95,11 @@ Inside the worker (`src/toplevel_driver.zig` `runWorkerFile` →
 2. **Suppress `(exit)`.** The worker sets `vm.suppress_exit`, so a file's
    `(exit 1)` failure epilogue becomes a *recorded* no-op (`vm.exit_requested`
    / `vm.exit_code`) instead of terminating the worker before it can emit its
-   result. Because the call never happens, its *semantics* are reapplied when
-   the verdict is resolved — see [Verdicts](#verdicts-and-what-errored-means)
-   below.
+   result. `(emergency-exit …)` records through the same channel — a plain
+   `emergencyExitFn` used to kill the worker mid-file, and the orchestrator
+   reported the file as a missing result (kaappi#2521). Because the call never
+   happens, its *semantics* are reapplied when the verdict is resolved — see
```
baijum, line 101: Same line CodeRabbit flagged, and I'd go a bit further than its rewording. "The call never happens" was already loose for `exit`; for `emergency-exit` it's misleading, because the *termination* doesn't happen but the call does, and the parts of its contract that depend on not returning are lost: the file's remaining forms run, and — new with this PR — `dynamic-wind` after-thunks fire on the way out of their extent, which a plain `(emergency-exit …)` skips (repro in my comment on `emergencyExitFn`). Suggest something like:

> Because the process does not actually terminate, the request's *semantics* — the exit code — are reapplied when the verdict is resolved (see Verdicts). What is not reproduced is the non-return: forms after the call still run, and for `emergency-exit` the `dynamic-wind` after-thunks a plain run would skip do run. Both are accepted worker-vs-plain divergences; `runner-agreement.sh` covers the shapes that matter.

Purely docs, no code change asked for.

baijum (commented):
Reviewed on 03d1619b. The fix is the right shape — the exact twin of the `exitFn` guard, raw code recorded before the #2512 upgrade so `resolveVerdict` sees what the file asked for — and everything the PR claims reproduces locally. Two non-blocking doc comments inline, plus one pre-existing gap I found while probing, filed separately.

**Verified**

| Check | Result |
|---|---|
| `tests/scheme/test-runner/emergency-exit.sh`, this head | 8 pass, 0 fail |
| same script against a base (a684eeb3) binary | 2 pass, 6 fail — transcripts show `ERROR ee-pass.scm (worker produced no result)`, matching the PR text exactly |
| `zig build test -Dtest-filter="emergency-exit" -Dtest-filter="suppress_exit"` | 7 pass |
| `runner-agreement.sh` / `errors/exit-wind.sh` / `errors/exit-code.sh` | 10/0, 8/0, 56/0 — the plain-run contract (skips winds, honors #2512) is untouched |

One note on the new script's matrix: the `assert_agree` for `ee-fail-epilogue.scm` passes on the base binary too (a crashed worker is also `FAIL`), so it's the two transcript assertions after it that actually discriminate. Fine as written since they're there; just don't count that row as evidence on its own.

**Pre-existing, out of scope — filed as #2525**

`VM.initForThread` doesn't copy `suppress_exit`, so `(exit …)` *or* `(emergency-exit …)` from a SRFI-18 child thread still runs the real `std.process.exit` and the worker dies with `worker produced no result`. Reproduced on this head with a `thread-join!`'d `(make-thread (lambda () (exit 0)))`; identical for the emergency variant, and identical on main, so nothing this PR introduced. Left it out of the PR deliberately — the request would also have to cross the thread boundary to reach the root VM's `emitResult`, which is a different change.

I can't approve my own PR from this account, so this is a comment review; from my side it's mergeable once the two doc nits are in (or explicitly declined).

PR #2524, opened 2026-09-05
Name the differing compiler-key component in bundle rejection

### LINKED ISSUE #2514: Cross-target bundle rejection prints the same build id twice and advises the wrong fix
## What happens

When an embedded `.sbc` is refused because it was built for a **different
target**, the diagnostic prints the same build id twice and asserts they are
different:

```
fatal: embedded bytecode was produced by a different build (build id abc1234);
this binary is abc1234. Rebuild the bundle from current source.
```

"Rebuild the bundle from current source" is also the wrong advice in that
case — the bundle *is* from current source. The fix is to rebuild it with a
kaappi built for the same target.

## Why the ids match

The compiler key has three inputs (`src/bytecode_file.zig:220`):

```zig
pub fn compilerHashFor(version_str: []const u8, build_id: []const u8, target: []const u8) u64
```

The message reports one of them (`src/main.zig:406-412`):

```zig
.foreign_build => {
    const info = bytecode_file.readHeaderInfo(bytecode_data).?;
    writeStderr("fatal: embedded bytecode was produced by a different build (build id ");
    writeStderr(info.build_id);
    writeStderr("); this binary is ");
    writeStderr(build_options.git_build_id);
    writeStderr(". Rebuild the bundle from current source.\n");
},
```

And this project's own test says why the other two are equal in exactly the
case that matters (`src/bytecode_file.zig:851`):

> kaappi#2155: all release binaries are built from one clean checkout, so
> version and build id are identical across all 17 targets. Only the target
> component separates them

So for a cross-target bundle the version matches, the build id matches, and
the *only* differing input is the one the message does not name.

`HeaderInfo` (`src/bytecode_file_read.zig:1026`) cannot report it either —
it carries `source_hash`, `compiler_hash`, `build_id`, `source_path` and
`current_build`, but neither the version nor the target. Fixing the message
means recording the target in the header.

## Why it matters

This is the failure a release pipeline hits, and it reads as impossible.
Building mesthiri's Linux release taught us that a `.sbc` must be compiled
by a kaappi built for the same target as the binary embedding it — so
releases are a matrix of native builds rather than cross-compilation. That
constraint is fine. What cost the time was the diagnostic: the two sides
print identical and it still refuses, which reads as a kaappi bug rather
than as "you cross-built this".

kaappi#1930 clearly went in to make this message specific. It succeeded for
the stale-bundle case it was aimed at; the cross-target case slipped
through because it is the one where `build_id` is uninformative.

## Suggested resolution

Record the target in the `.sbc` header and name it in the message, e.g.:

```
fatal: embedded bytecode was produced by a different build
  bundle:  0.26.1  abc1234  aarch64-macos-none;r7rs;posix
  binary:  0.26.1  abc1234  x86_64-linux-gnu;r7rs;posix
The target differs: recompile the .sbc with a kaappi built for this target.
```

Printing all three components and letting the reader see which line differs
would also cover the version case, which is likewise unreported today.

## Version

Kaappi Scheme v0.26.1; source read at current main.

baijum:
Closes #2514

## The problem

When a bundled standalone refuses its embedded `.sbc` because it was built for a **different target**, the fatal diagnostic printed the build id on both sides — identical, because all release binaries come from one clean checkout — and still refused, advising "Rebuild the bundle from current source". The bundle *is* from current source; the real fix is recompiling the `.sbc` with a kaappi built for the embedding binary's target. The target is one of the three inputs of `compilerHashFor` (version, build id, target) and the only one the message never named.

## The fix

- **Format v14**: the `.sbc` header records the producing binary's compile target (`compile_target_id`, the exact string `compilerHashFor` hashes) and release version string after the source path. Purely informational, never load-gating — same contract as the v10 build id.
- **`HeaderInfo`** carries `target: ?[]const u8` and `version: ?[]const u8` (null for pre-v14 headers).
- **`renderForeignBuildDiagnostic`** (in `bytecode_file_read.zig`, re-exported by the hub; pure in its inputs) prints all three components for both sides and one advice line per differing component:
  - target differs → "The target differs: recompile the .sbc with a kaappi built for this target."
  - version or build id differs → the kaappi#1930 stale-bundle advice, unchanged.
  - nothing recorded differs (pre-v14 header) → says the differing component is not recorded, prints `?` for the unrecorded components, and never claims equality of a component it never read.
- `main.zig`'s `.foreign_build` arm renders through it.

New message, cross-target case (end-to-end via `-Dbundle`):

```
fatal: embedded bytecode was produced by a different build
  bundle:  0.26.1  a684eeb3-dirty  aarch64-linux-none;r7rs;kaappi;…
  binary:  0.26.1  a684eeb3-dirty  aarch64-macos-none;r7rs;kaappi;…
The target differs: recompile the .sbc with a kaappi built for this target.
```

## Backward compatibility

The reader now accepts `MIN_READ_VERSION..VERSION` (13..14) instead of exactly `VERSION`. A v13 file merely lacks the two new strings and shares v14's tail layout, so it still **loads** (cache safety is unchanged — the v13 compiler hash already folded version, build id and target), still classifies as `.foreign_build` when foreign, and reads the new fields back as null. v12 and older are rejected exactly as before. The checked-in v13 fuzz-seed fixture therefore still loads through its comptime compiler-hash patch and was not regenerated.

## Tests

- `classifyEmbeddedRejection: cross-target bundle names the target` — a v14 header at this version + build id but a different target classifies as `.foreign_build`, the loader refuses it, and the rendered diagnostic names the target with the recompile advice (and not the stale-bundle one).
- `renderForeignBuildDiagnostic: advice matches the differing component` — build-id, version, and pre-v14-unknown cases each get the right advice; the legacy case never claims a component differs.
- `pre-v14 .sbc still classifies and loads` — a v13 header classifies with `target`/`version` reading back null, and a complete v13 program entry with a current compiler hash still deserializes.
- Extended: the kaappi#2155 area tests, the `readHeaderInfo` round-trip (now asserts target + version), the golden-byte endian fixtures, and the hand-assembled header fixtures (now well-formed v14).
- Sabotage-checked: breaking one assertion in the cross-target test fails the suite, so the tests execute, not skip.

Evidence: `zig build test` — **unit-tests 1933 pass, 21 skip, 0 fail; thottam-tests 88 pass**. End-to-end: forged cross-target bundle prints the message above and exits 1; the un-forged same-target bundle prints its program output and exits 0.

### DIFF
--- a/docs/dev/cache.md
+++ b/docs/dev/cache.md
@@ -101,11 +101,20 @@ embedded bundle can say *which* of the three compiler-key components differed
 (`renderForeignBuildDiagnostic` in `src/bytecode_file_read.zig`, re-exported by
 `src/bytecode_file.zig`; for a cross-target bundle the build id matches on both
 sides, so it cannot carry the diagnosis alone). Older headers lack the last two
-fields and read them back as absent:
-the reader accepts `MIN_READ_VERSION..VERSION` (v13 files load and classify
-unchanged — a bump that only *adds* header fields never invalidates a cache),
-and anything older reads back as a miss. A format bump that changes the entry
-*layout* still invalidates every older entry.
+fields and read them back as absent: the reader accepts
+`MIN_READ_VERSION..VERSION`, and anything older reads back as a miss.
+
+That read window is narrower than it looks. Loading still requires the
+compiler hash to match, and the hash folds in the version string and the build
+id — so the only v13 entry that can *hit* under the v14 reader is one written
+at the same commit and tree state, in practice a dirty build on the author's
+own machine; for every released or per-commit binary a v13 entry is a miss
+either way. What the window actually buys: a v13 embedded bundle still
+classifies as a foreign build (and gets the "not recorded" advice rather than
+"invalid"), `cache status` still parses v13 entries, and the checked-in
+fuzz-seed fixture did not need regenerating. A format bump that changes the
+entry *layout* must raise `MIN_READ_VERSION` (see the `VERSION` doc-comment)
+and so invalidates every older entry.
 
 ## Location
 
--- a/docs/dev/cache.md
+++ b/docs/dev/cache.md
@@ -94,9 +94,18 @@ case above is different in kind — the key genuinely matches — which is why i
 needs the manual step.
 
 The header also records, purely for `cache status` to display, the **producing
-build id** and the **source path** (see `src/bytecode_file.zig`, format
-`VERSION`). Bumping the on-disk format `VERSION` invalidates every older entry —
-a version mismatch reads back as a miss.
+build id** and the **source path**, and since format v14
+([#2514](https://github.com/kaappi/kaappi/issues/2514)) the producing binary's
+**compile target** and **version string** too — the last two so a rejected
+embedded bundle can say *which* of the three compiler-key components differed
+(`renderForeignBuildDiagnostic` in `src/bytecode_file_read.zig`, re-exported by
+`src/bytecode_file.zig`; for a cross-target bundle the build id matches on both
+sides, so it cannot carry the diagnosis alone). Older headers lack the last two
+fields and read them back as absent:
+the reader accepts `MIN_READ_VERSION..VERSION` (v13 files load and classify
+unchanged — a bump that only *adds* header fields never invalidates a cache),
+and anything older reads back as a miss. A format bump that changes the entry
+*layout* still invalidates every older entry.
 
 ## Location
 
--- a/src/bytecode_file.zig
+++ b/src/bytecode_file.zig
@@ -54,6 +54,13 @@ pub const MAGIC = [4]u8{ 'K', 'P', 'B', 'C' };
 /// order), and library entries carry transformer, event, include and
 /// dependency sections. v12 (kaappi#2166): TAG_COMPLEX payload changed from
 /// two f64s + two exactness bytes to two nested component constants.
+///
+/// Bumping rule: a bump that changes the entry *tail* (anything after the
+/// header strings) must raise `MIN_READ_VERSION` to the new value — nothing
+/// else stops the reader from parsing an older tail under the new layout. A
+/// bump that only appends header fields may leave `MIN_READ_VERSION` alone,
+/// provided the readers gate the new fields on `ver >=` a named threshold
+/// like `TARGET_FIELDS_SINCE`.
 pub const VERSION: u16 = 14;
 /// Oldest header version this binary reads. v13 files lack the target and
 /// version strings but share v14's tail layout (entry kind, slots, sections),
@@ -61,6 +68,14 @@ pub const VERSION: u16 = 14;
 /// diagnostics must not claim equality of an unrecorded component. Anything
 /// older is rejected as before: a version mismatch reads as a plain miss.
 pub const MIN_READ_VERSION: u16 = 13;
+/// First format version whose header carries the compile target and version
+/// strings after the source path (kaappi#2514). Both readers gate those two
+/// reads on it; a header below it reads them back as null.
+pub const TARGET_FIELDS_SINCE: u16 = 14;
+comptime {
+    std.debug.assert(MIN_READ_VERSION <= VERSION);
+    std.debug.assert(MIN_READ_VERSION <= TARGET_FIELDS_SINCE and TARGET_FIELDS_SINCE <= VERSION);
+}
 pub const MAX_FUNCTIONS: u32 = 16_384;
 pub const MAX_TOP_LEVEL_FUNCTIONS: u32 = 4_096;
 pub const MAX_CODE_BYTES: u32 = 4_194_304;
@@ -979,8 +994,9 @@ test "classifyEmbeddedRejection: cross-target bundle names the target (kaappi#25
     try std.testing.expect(!std.mem.eql(u8, other_target, compile_target_id));
 
     // A v14 header produced by a binary for `other_target` at this very
-    // version and build id. The literal 14 pins the format this test targets;
-    // the reader must keep accepting it across future bumps.
+    // version and build id. The literal 14 pins the format this fixture is
+    // written in; when MIN_READ_VERSION passes 14, rewrite the fixture in the
+    // new format rather than the assertion.
     var w = Writer{ .buf = .empty };
     defer w.buf.deinit(allocator);
     const foreign_key = compilerHashFor(main.version, build_options.git_build_id, other_target);
--- a/src/bytecode_file.zig
+++ b/src/bytecode_file.zig
@@ -42,13 +42,25 @@ pub const MAGIC = [4]u8{ 'K', 'P', 'B', 'C' };
 // runtime errors can report `file:line:col`. A version mismatch makes
 // `readFileWithTopLevel` return null, so older `.sbc` caches are ignored and
 // silently recompiled — no stale-cache hazard.
-/// Format version. v13 (kaappi#1888): the header carries an entry-kind byte
-/// after the source path (program / bundle / library), program entries carry a
-/// positional slot section (compiled function vs. declaration source text, in
-/// top-level order), and library entries carry transformer, event, include and
-/// dependency sections. v12 (kaappi#2166): TAG_COMPLEX payload changed from two
-/// f64s + two exactness bytes to two nested component constants.
-pub const VERSION: u16 = 13;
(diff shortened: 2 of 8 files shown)


src/bytecode_file_read.zig
```diff
@@ -1028,33 +1035,146 @@ pub const HeaderInfo = struct {
...
+        .{},
+    );
+    if (target_differs) try w.print(
+        "The target differs: recompile the .sbc with a kaappi built for this target.\n",
+        .{},
+    );
+    if (!version_differs and !build_id_differs and !target_differs) {
+        if (bundle.version == null or bundle.target == null) {
+            // Pre-v14 header: the differing input is one this header never
+            // wrote down (target/version), or two dirty builds sharing one
```
baijum: This clause isn't reachable. Two dirty builds at the same commit share the build id *and* the version *and* the target, so `compilerHashFor` — pure in those three — gives them the **same** hash; they never classify as `.foreign_build`, they silently load each other's entries (the kaappi#1516 footgun the `compilerHashFor` doc describes). The only way a v13 header with a matching build id lands here is a differing version or target that the header didn't record — which is exactly what the printed message says. Suggest dropping "or two dirty builds sharing one build id" so the comment matches the text.

src/bytecode_file.zig
```diff
@@ -42,13 +42,25 @@ pub const MAGIC = [4]u8{ 'K', 'P', 'B', 'C' };
...
+/// order), and library entries carry transformer, event, include and
+/// dependency sections. v12 (kaappi#2166): TAG_COMPLEX payload changed from
+/// two f64s + two exactness bytes to two nested component constants.
+pub const VERSION: u16 = 14;
+/// Oldest header version this binary reads. v13 files lack the target and
+/// version strings but share v14's tail layout (entry kind, slots, sections),
+/// so they still load and classify — the new fields read back as null and
+/// diagnostics must not claim equality of an unrecorded component. Anything
+/// older is rejected as before: a version mismatch reads as a plain miss.
+pub const MIN_READ_VERSION: u16 = 13;
```
baijum, line 70: Nothing enforces that a *layout-changing* bump also raises this. The only canary is the fuzz-seed sanity test, and it fires only if the stale v13 tail actually fails to parse under the new layout — a layout change that keeps the old bytes parseable (e.g. a widened count field that happens to read as small) would sail through with the old fixture silently misinterpreted. Two cheap hardenings:

1. Name the literal `14` that appears in both readers (`bytecode_file_read.zig:584` and `:1068`) — something like `pub const TARGET_FIELDS_SINCE: u16 = 14;` — so a future reader has one place to look, and add a `comptime { std.debug.assert(MIN_READ_VERSION <= VERSION); }` next to it.
2. One sentence in the `VERSION` doc-comment along the lines of "a bump that changes the entry *tail* must raise `MIN_READ_VERSION` to the new value; a bump that only appends header fields may leave it" — that's the rule this PR establishes, and right now it lives only in `cache.md`.

src/bytecode_file.zig
```diff
@@ -949,7 +965,191 @@ test "classifyEmbeddedRejection: foreign build id is distinguished from corrupt"
...
+    var gc = GC.init(allocator);
+    defer gc.deinit();
+
+    // A target guaranteed different from this binary's on every host.
+    const other_target = "fakearch-fakeos-none;r7rs;fake";
+    try std.testing.expect(!std.mem.eql(u8, other_target, compile_target_id));
+
+    // A v14 header produced by a binary for `other_target` at this very
+    // version and build id. The literal 14 pins the format this test targets;
+    // the reader must keep accepting it across future bumps.
```
baijum: This states a policy the design doesn't have: `MIN_READ_VERSION` exists precisely so a layout bump *can* stop accepting 14. When that happens this fixture legitimately fails and should be rewritten as a header of the new format, not kept. Suggest: "The literal 14 pins the format this fixture is written in; when MIN_READ_VERSION passes 14, rewrite the fixture rather than the assertion."

docs/dev/cache.md
```diff
@@ -94,9 +94,18 @@ case above is different in kind — the key genuinely matches — which is why i
...
+build id** and the **source path**, and since format v14
+([#2514](https://github.com/kaappi/kaappi/issues/2514)) the producing binary's
+**compile target** and **version string** too — the last two so a rejected
+embedded bundle can say *which* of the three compiler-key components differed
+(`renderForeignBuildDiagnostic` in `src/bytecode_file_read.zig`, re-exported by
+`src/bytecode_file.zig`; for a cross-target bundle the build id matches on both
+sides, so it cannot carry the diagnosis alone). Older headers lack the last two
+fields and read them back as absent:
+the reader accepts `MIN_READ_VERSION..VERSION` (v13 files load and classify
+unchanged — a bump that only *adds* header fields never invalidates a cache),
```
baijum: Slight overclaim. Loading a v13 entry still requires its compiler hash to match, and that hash folds in the version string and the build id — so the only v13 entries that can *hit* under the v14 reader are ones written at this same commit and tree state (in practice: a dirty build on the author's machine). For every released or per-commit binary, a v13 entry is a miss either way. The window buys three real things: v13 bundles classify as `.foreign_build` (and get the "not recorded" message instead of "invalid"), `cache status` can still parse them, and the fuzz-seed fixture didn't need regenerating. Worth saying that here rather than "never invalidates a cache", so nobody later relies on the window for something it can't deliver.

baijum (commented):
Reviewed against `a684eeb3`; ran the affected unit tests locally on the branch (`-Dtest-filter` over the 2514 / readHeaderInfo / endian / fuzz-seed / classify / validation tests): **40/40 pass**. I found no correctness problem — the reader window, the v13 tail-layout claim, the `.foreign_build` classification of a v14 cross-target header, and the null-safe `main.zig` arm all check out. Four comment-level notes inline; two more that sit on unchanged lines:

- **`tests_endian.zig` comptime guard (line ~228) is now incomplete.** It asserts `lo != hi` so that byte-swapping `VERSION` is *visible*. With a read window, visibility is no longer sufficient — the swapped value must also fall **outside** `MIN_READ_VERSION..VERSION`. Today 14 → 0x0E00 = 3584 is safely outside, so nothing is wrong, but the invariant the guard states is the wrong one. Suggest replacing it with `const swapped = @byteSwap(bf.VERSION); if (swapped >= bf.MIN_READ_VERSION and swapped <= bf.VERSION) @compileError(...)`, which subsumes the `lo == hi` case.
- **Follow-up, not this PR:** `kaappi cache status` still shows only build id + source path for stale entries; now that the header carries the target, showing it for `current_build == false` entries would make a cross-target cache entry self-explaining the same way the bundle is.

src/bytecode_file_read.zig
baijum: Done in ccfc68b0: the clause is gone, and the comment now says explicitly that two dirty builds at one commit do *not* reach this branch (they share all three hash inputs, kaappi#1516).

src/bytecode_file.zig
baijum: Done in ccfc68b0: both readers now gate on `bf.TARGET_FIELDS_SINCE`, a comptime block asserts `MIN_READ_VERSION <= TARGET_FIELDS_SINCE <= VERSION`, and the `VERSION` doc-comment carries the bumping rule (tail change ⇒ raise `MIN_READ_VERSION`; header-only append ⇒ gate on a named threshold).

src/bytecode_file.zig
baijum: Done in ccfc68b0: reworded to "the literal 14 pins the format this fixture is written in; when MIN_READ_VERSION passes 14, rewrite the fixture in the new format rather than the assertion."

docs/dev/cache.md
baijum: Done in ccfc68b0: the "never invalidates a cache" sentence is replaced by a paragraph saying the window only lets a same-commit, same-tree-state v13 entry hit, and listing what it actually buys (v13 bundles classify as foreign, `cache status` parses them, fuzz-seed fixture untouched). markdownlint clean.

baijum:
Addressed the two review-body items in ccfc68b0 as well: the `tests_endian.zig` comptime guard now checks that `@byteSwap(VERSION)` falls outside `MIN_READ_VERSION..VERSION` (subsumes the old `lo == hi` check). The `cache status` target column is left as a follow-up. Filtered unit tests (2514 / readHeaderInfo / endian / fuzz seed / classify / validation / round-trip): 40/40 pass.

PR #2528, opened 2026-09-05
Honor spawn-process directory: at run time on no-addchdir builds

### LINKED ISSUE #2517: Linux release build: spawn-process directory: raises KP3007 on every Linux host (gnu.2.28 floor is the build target's)
### kaappi version

`v0.26.1` release binary (`kaappi-x86_64-linux`), on **any** Linux host — reproduced on `ubuntu-24.04` (glibc 2.39), well above the 2.29 that matters below.

### What happens

`spawn-process` with `directory:` refuses to spawn, regardless of the host's glibc:

```
error[KP3007]: spawn-process: directory: is not supported on this platform
```

### Minimal reproduction

`repro.scm`:

```scheme
(import (kaappi process))
(display (process-wait (spawn-process '("true") 'directory: "/tmp")))
(newline)
```

```
$ kaappi repro.scm
error[KP3007]: spawn-process: directory: is not supported on this platform
```

The same file on the macOS arm64 release prints `0`.

### Why the release build does this

`src/process_posix.zig` gates `posix_spawn_file_actions_addchdir_np` behind a comptime check of the target's glibc version, because referencing the symbol unconditionally would fail the `x86_64-linux-gnu.2.28` release link (kaappi#2414 review, per the comment there). That is right as a link constraint — but the floor is the *build target's*, not the host's, so the released binary loses `directory:` on every Linux machine, including brand-new runners. Downstream, mesthiri pins the release for CI and its agent spawner needs a working directory; it hit this on every Linux job (mesthiri/mesthiri#4) and is currently working around it with a fixed `/bin/sh -c 'cd "$1" && shift && exec "$@"'` wrapper.

### Ask

Either or both:

1. ship a Linux release artifact on a target where the gate is true — musl ≥ 1.1.24 already returns `true` in `has_addchdir`, so an `x86_64-linux-musl` binary would honour `directory:` as-is;
2. on the gnu.2.28 target, fall back at run time (e.g. fork + chdir + exec, or the vfork equivalent) when `directory:` is asked of a build whose comptime gate is false, instead of rejecting it.

Happy to test a candidate artifact — mesthiri's CI would exercise it on every push.

baijum:
Closes #2517

## What

`spawn-process` with `directory:` raised `KP3007 "directory: is not supported on this platform"` on **every** spawn by a build whose comptime gate for `posix_spawn_file_actions_addchdir_np` is false — which is every Linux release binary, because the release workflow targets `x86_64`/`aarch64-linux-gnu.2.28` and the floor is the *build target's*, not the host's (reproduced on ubuntu-24.04 / glibc 2.39, kaappi#2517).

The comptime gate has to stay for a **strong** extern: it is a link constraint (referencing the symbol unconditionally fails the gnu.2.28 release link, kaappi#2414 review). But a link constraint is not a capability verdict, so this PR honors the option instead of rejecting it — the issue's option 2, reworked per review around a **weak extern** bound at run time.

## How

- **Runtime availability, two layers.** Where the comptime link gate allows (macOS, FreeBSD, musl, gnu-ABI ≥ 2.29), the plain extern is used. On a comptime-false build, a **weak extern** asks at run time: it links on the gnu.2.28 floor and the dynamic linker binds it on every host with glibc ≥ 2.29, so a release binary keeps the `posix_spawnp` fast path for `directory:` essentially everywhere. Only where the symbol is genuinely absent — a glibc 2.28 host, NetBSD, OpenBSD — does the route divert to the fork path. The weak extern is bound in a **`var`**, which is load-bearing: with a `const`, Zig comptime-knows the optional non-null, the null test and unwrap fold, and an unbound host calls through null (measured: segfault on debian:10 / glibc 2.28); a `var` puts the pointer in `.data` behind a load-time relocation, so availability is a real runtime load.
- `process_posix.routeFor` is the route decision, `pub inline fn`: the OpenBSD fork key is a `comptime` parameter and **folds** — on OpenBSD the posix route is not analyzed at all, restoring the pre-fallback link protection; on comptime-true builds the availability argument is comptime-known `true`, likewise folding; on gnu.2.28/NetBSD builds the weak binding is runtime *by design*, both routes are emitted, and the branch is taken per spawn.
- The **fork + exec route** (OpenBSD's since #2456, and the fallback since this PR) runs the chdir between fork and exec — what glibc's `addchdir_np` performs on the vforked child inside `posix_spawn`; only async-signal-safe calls in the child, all fd/pipe/stdio setup parent-side. Its error pipe is lifted above the stdio range so a closed-stdio launcher cannot let a redirection replace the report path (CodeRabbit). A bad `directory:` raises a synchronous ENOENT file error, matching macOS/FreeBSD posix_spawn; on glibc the posix route instead reports it as a child exiting 127 — glibc's own action-failure behavior, unchanged from main.
- **PATH parity** with `posix_spawnp`: the child walks the *parent*-captured PATH with `execve` (CPython's mechanism; `execvpe` is not an option on OpenBSD), preserving execvp's rules — under `env:` the routes used to disagree for the same request.
- The child's fd sweep uses the raw `close_range(2)` syscall with `CLOSE_RANGE_CLOEXEC` on Linux (kernel ≥ 5.11; raw syscall, so no glibc-floor concern and async-signal-safe; ~80 ms/spawn of fcntl probes at soft nofile 524288 otherwise), falling back to the probe for older kernels and the BSDs. Marking CLOEXEC is the correct shape: everything ≥ 3 the child still holds is either close-on-exec by design or must not survive the exec — a plain close_range would have killed the error pipe.
- `supports_directory` and its KP3007 raise are removed from both backends. Windows was already native (`lpCurrentDirectory`).

## Verification

**Native (macOS arm64):** the issue's repro prints `0`; a bad directory raises `(file-error ENOENT)`; full `zig build test`: **2033 passed, 21 skipped, 0 failed**; process suites green under `-Dgc-stress=true`; `zig build wasm`, and the `x86_64-linux-gnu.2.28`, `x86_64-linux` (test binary), `x86_64-openbsd`, `x86_64-netbsd` cross-builds all compile and link.

**Cross-build proof** — `zig build -Dtarget=x86_64-linux-gnu.2.28`, run under podman (`--platform linux/amd64` via Rosetta) on **both** sides of the 2.29 boundary:

Before (unmodified main @ d6ca2229), on ubuntu-24.04 (glibc 2.39):

```
$ kaappi repro.scm
repro.scm:2:56: error[KP3007]: spawn-process: directory: is not supported on this platform
exit=1
```

After, on ubuntu-24.04 (glibc 2.39 — weak reference binds, **posix route**):

```
$ kaappi repro.scm
0
```

The child's `pwd` is captured through a pipe as `/tmp`; the `env:` + bare-name + `directory:` parity case returns `0` on both forms; 50 `directory:` spawns time the same as 50 plain spawns (~12 ms/spawn each — the fork route's ~92 ms/spawn measured cost is out of the common path). `readelf -sW` shows `WEAK DEFAULT UND posix_spawn_file_actions_addchdir_np`.

After, on debian:10 (glibc **2.28** — the reference stays unbound, **fork route**):

```
$ ldd --version | head -1
ldd (Debian GLIBC 2.28-10+deb10u3) 2.28
$ kaappi deb10.scm
0                                   ;; the issue's repro
(0 "/tmp\n" "")                     ;; pwd captured through pipes
(file-error ENOENT)                 ;; bad directory: raises through the error pipe
```

The fork route is additionally covered by the route-table unit test, the direct `spawnChildForkExec` tests (chdir, error-pipe lift, parent PATH, close-by-default — the last exercising `close_range` on the Linux CI legs), and the NetBSD/OpenBSD CI legs where the dispatch really forks.

**Tests** (new suite `src/tests_process_fork.zig` — the fork-route cluster split out of `tests_process.zig` at the 1500-line policy ceiling):

- routing table for `routeFor`, including the runtime weak-binding rows — "posix wherever addchdir is available, `directory:` included";
- the fork route chdirs the child before exec (direct call: captured `pwd` is `/`, exit 0);
- the error-pipe lift regression: closed stdio + conflicting `'null` redirection + invalid `directory:` must raise ENOENT, not "succeed" (verified to fail with the lift removed);
- bare names resolve on the **parent's** PATH under `env:` (direct call; verified to fail when the PATH source is wrong), plus the Scheme-level `env:`/`directory:` parity test;
- the fork route's close-by-default removes an inherited descriptor (fcntl probe on macOS, `close_range(CLOEXEC)` on the Linux CI legs).

Chose Zig unit tests + the Scheme test over a podman-gated script: the repo's `.sh` suites gate on host OS against the locally built binary, with no podman/cross-build precedent, and a gnu.2.28+podman script could not run in CI. The seam (`routeFor` as a pure decision + `pub spawnChildForkExec`) gives deterministic on-host coverage of the exact mechanism instead.

## Out of scope

The issue's option 1 — shipping an additional musl release artifact — is a release-workflow change independent of this fix and stays out of this PR.


src/process_posix.zig
```diff
@@ -641,6 +730,15 @@ fn childExecSide(
         if (!ok) childExecFail(err_fd, lastError());
     }
 
+    // `directory:`, performed parent-side by addchdir_np on comptime-true
+    // builds. chdir(2) between fork and exec is async-signal-safe, and a
+    // failure reports through the error pipe exactly like a failed exec —
+    // the same errno a comptime-true build's addchdir_np surfaces as the
+    // posix_spawnp return, so both routes raise the identical condition.
+    if (directory) |dir| {
+        if (std.c.chdir(dir) != 0) childExecFail(err_fd, lastError());
```
baijum, line 838: **PATH search diverges between the two routes under `env:`** (the actual line is 778, `std.c.environ = …; execvp`, outside the diff hunk).

The fork route installs the caller's `env:` block into `environ` and then `execvp`s, so a bare program name is searched in the *child's* PATH. glibc's `posix_spawnp` searches the *parent's* PATH (it resolves `getenv("PATH")` in the caller before the vfork). So the same release binary now answers differently depending on whether `directory:` was passed.

Measured on this branch's `x86_64-linux-gnu.2.28` build under podman `ubuntu:24.04` (glibc 2.39):

```scheme
(spawn-process '("true") 'env: '(("PATH" . "/nonexistent")))                    ; => 0
(spawn-process '("true") 'env: '(("PATH" . "/nonexistent")) 'directory: "/tmp") ; => (file-error ENOENT)
```

On macOS (comptime-true) both return 0. The reverse holds too: a program only on the `env:` PATH spawns via the fork route and fails ENOENT via `posix_spawnp`. So "both routes raise the identical condition" does not hold for `env:` + bare program name.

Options: search the parent's PATH in the child (capture `getenv("PATH")` before the fork and walk it with `execve`, which is what CPython does), or use `execvpe` where libc has it (glibc, musl, NetBSD ≥ 8; OpenBSD lacks it). Either way an `env:` + `directory:` parity test in `tests_process.zig` would pin it.

src/process_posix.zig
```diff
@@ -641,6 +730,15 @@ fn childExecSide(
...
+    }
+
     // The fork twin of addInheritedFdCloses: every open non-CLOEXEC fd
```
baijum: **The child's fd sweep (line 755, `while (fd < limit)`) now runs on Linux release binaries, and it is expensive there.**

This loop fcntl-probes every descriptor up to the soft `RLIMIT_NOFILE`. It was written for OpenBSD's small limits; Linux containers default to 524288–1048576. Measured on this branch's gnu.2.28 build in podman `ubuntu:24.04` (soft nofile 524288), 20 spawns of `true` each:

| | ms/spawn |
|---|---|
| `posix_spawnp` (no `directory:`) | 11.0 |
| fork route (`directory: "/tmp"`) | 92.6 |
| both, after `ulimit -Sn 1024` | 10.8 |

So ~80 ms of pure `fcntl` syscalls per `directory:` spawn; mesthiri's Linux CI spawner would pay it on every child.

Linux has `close_range(2)` (kernel ≥ 5.9), exposed as the raw syscall `std.os.linux.close_range` — async-signal-safe and no glibc-floor concern since it bypasses libc. Try it first (`close_range(3, maxInt, .{})`) and fall back to the probe on ENOSYS/EINVAL. Alternatively, compute the highest open fd parent-side from `/proc/self/fd` before the fork (allocation is fine there) and pass it in as the sweep bound. The rationale comments here ("no /proc on OpenBSD") and in `addInheritedFdCloses` ("Linux never reaches this loop", line 1013) are stale either way.

src/process_posix.zig
```diff
@@ -296,178 +339,18 @@ pub fn spawnChild(gc: *GC, cfg: SpawnConfig, redirs_in: [3]Redir) PrimitiveError
...
+    // -- spawn. Route it (see routeFor): the posix_spawnp fast path takes
+    // every spawn on a comptime-true addchdir build and every directory-less
+    // spawn anywhere; the fork + exec route takes OpenBSD and — the
+    // kaappi#2517 fallback — every `directory:` spawn on a comptime-false
+    // build, whose child chdirs between fork and exec because file actions
+    // cannot express the change of directory there.
     var pid: c_int = -1;
-    if (comptime spawn_c.spawn_via_fork_exec) {
-        pid = try spawnChildForkExec(gc, argv, env_block, redirs, child_pipe_ends, cfg);
+    if (routeFor(spawn_c.spawn_via_fork_exec, spawn_c.has_addchdir, directory_z != null) == .fork_exec) {
```
baijum: **Altitude: a weak reference removes the link constraint without forking at all.**

The comptime gate exists because an unconditional reference fails the gnu.2.28 link. A weak extern does not:

```zig
const addchdir_np = @extern(?*const fn (*FileActionsPtr, [*:0]const u8) callconv(.c) c_int, .{
    .name = "posix_spawn_file_actions_addchdir_np",
    .linkage = .weak,
});
```

I verified this: a test program built with `-target x86_64-linux-gnu.2.28` links cleanly and `addchdir_np != null` at run time under `ubuntu:24.04` (glibc 2.39). The dynamic linker binds the unversioned weak reference to the default `@@GLIBC_2.29` version.

With `routeFor` keyed on `addchdir_np != null` (a runtime fact, like `directory_requested`), release binaries keep the `posix_spawnp` fast path on every Linux host with glibc ≥ 2.29 — essentially all of them — and the fork route narrows to genuinely old hosts, NetBSD and OpenBSD. That also keeps the two issues above (PATH divergence, fd-sweep cost) out of the common path instead of adding them to every Linux release `directory:` spawn. The fork fallback is still needed, so this is a scoping change, not a rewrite. Worth considering before this lands, since it changes which path the tests need to pin.

src/process_posix.zig
```diff
@@ -204,6 +215,38 @@ pub const signal_zero_is_probe = true;
...
+/// - Everything else keeps `posix_spawnp`, so the fast path is untouched on
+///   comptime-true builds and for every directory-less spawn.
+///
+/// The comptime facts are `comptime` parameters so the call site's
+/// build-known arguments fold the decision — the untaken route is never
+/// emitted, and comptime-true builds behave exactly as before the fallback
+/// — while keeping them parameters (rather than reading `spawn_c` directly)
+/// makes the routing table unit-testable on hosts where the false-flag
+/// combinations never occur natively (`tests_process.zig`).
+pub fn routeFor(comptime spawn_via_fork_exec: bool, comptime has_addchdir: bool, directory_requested: bool) SpawnRoute {
```
baijum: **`routeFor` does not fold — both routes are analyzed on every target.**

`routeFor` is a plain `fn` called with a runtime third argument (`directory_z != null`), so the call is a runtime call and Zig analyzes both branches of `if (routeFor(...) == .fork_exec)`. The doc comment's "the untaken route is never emitted" and the PR's "comptime-true builds emit exactly as before" describe LLVM's DCE at best, not what Sema does.

Minimal repro: putting `@compileError` in the else branch of `if (routeFor(true, true, rt) == .fork_exec)` fires with the plain `fn`, and does not fire when `routeFor` is declared `inline fn`.

Concretely: on OpenBSD `spawnChildPosixSpawn` is now analyzed where `if (comptime spawn_c.spawn_via_fork_exec)` used to exclude it. It links today (I cross-compiled x86_64-openbsd, x86_64-netbsd and x86_64-linux-gnu.2.28 on this branch; all build), but the link protection the comment claims is gone. One keyword restores it:

```suggestion
pub inline fn routeFor(comptime spawn_via_fork_exec: bool, comptime has_addchdir: bool, directory_requested: bool) SpawnRoute {
```

baijum (commented):
Reviewed the diff plus the enclosing spawn code, cross-compiled `x86_64-linux-gnu.2.28`, `x86_64-openbsd` and `x86_64-netbsd` from this branch, and ran the gnu.2.28 binary under podman `ubuntu:24.04` to check the fallback empirically.

The fix does what it says: `directory:` works on the gnu.2.28 binary and a bad directory raises a file error with ENOENT. Four findings, most severe first:

1. **PATH lookup differs between the two routes under `env:`** — confirmed by running both on the same binary.
2. **The fork route's fd sweep costs ~80 ms per spawn at container fd limits** — measured 92.6 vs 11.0 ms/spawn.
3. **Altitude:** a weak `@extern` for `addchdir_np` links at gnu.2.28 and resolves on newer hosts (verified), which would keep the fast path almost everywhere and make 1 and 2 rare-path issues.
4. **`routeFor` does not actually fold** — `inline fn` makes the doc comment true.

Details inline.

baijum:
Thanks — all five addressed (68571101, plus d2380bed fixing two Linux-only defects the first push left). Each comment was verified against the code before fixing; none skipped.

**1. coderabbitai — error-pipe write end on fd 0/1/2 (Major).** Confirmed: `spawnChildForkExec` created the pipe and never lifted it, so with stdio closed the pipe landed in the stdio range and a non-inherit redirection could replace the child's write end before a failing `chdir` reported — parent reads EOF, spawn "succeeds", child dies 127. Fix is your diff verbatim: `try normalizePipeAboveStdio(gc, &err_pipe);` right after creation. Regression test (`tests_process_fork.zig`, "fork-route chdir failure reports even from a replaced stdio slot"): saves/dups fds 0-2, closes them, spawns with `'stdout: 'null` (the conflicting non-inherit redirection) and `directory: "/no/such/directory/kaappi-2517"`, and requires the `.file`/ENOENT raise. Verified it fails with the lift removed.

**2. PATH divergence under `env:`.** Confirmed at the old `environ = …; execvp` tail. Went with your libc-agnostic option since OpenBSD lacks `execvpe`: the parent captures `getenv("PATH")` before the fork, and the child walks it with `execve` (`childPathExec`) — CPython's mechanism, stack buffers only, with execvp's exact rules (slash short-circuits the search, empty entry = cwd, EACCES remembered while the search continues, unset PATH → `/bin:/usr/bin`). The global `environ` mutation is gone entirely. Pinned twice: a direct fork-route test (bare `sh` + `env: PATH=/nonexistent` must spawn and run — verified to fail when the PATH source is wrong) and a Scheme-level `env:`/`directory:` parity test that both routes answer identically. Also re-verified end-to-end on the gnu.2.28 build in podman: both forms now return `0` where the `directory:` form raised ENOENT before.

**3. fd sweep cost on Linux.** Replaced the probe with the raw `std.os.linux.close_range(3, ~0U, CLOSE_RANGE_CLOEXEC)` — raw syscall, so the gnu.2.28 floor is irrelevant and async-signal-safety holds; ENOSYS (pre-5.9), EINVAL (5.9-5.10) and, defensively, anything else fall back to the existing probe (kept for OpenBSD/NetBSD). I picked the **CLOEXEC variant, not a plain close**: at sweep time every fd ≥ 3 the child still holds is either close-on-exec *by design* (staged redirect sources; the error pipe, whose EOF-at-exec is the parent's very success signal) or an inherited descriptor close-by-default must remove — marking the whole range CLOEXEC gives exactly the probe's end state at exec, whereas a plain `close_range(3, max, .{})` would have killed the error pipe before a chdir/exec failure could report (your comment 1's failure mode, reintroduced by accident). The rationale comment defends this in the code; the stale comments at both probes ("no /proc on OpenBSD", "Linux never reaches this loop") are rewritten. Regression test: a parked inheritable fd must be absent from the fork-route child (`[ -e /dev/fd/N ]` probe with an fd-2 control) — fcntl probe on macOS, close_range on the Linux CI legs. Timing on ubuntu-24.04: 50 `directory:` spawns now time the same as 50 plain spawns (~12 ms/spawn each, vs your 92.6 ms fork-route measurement).

**4. weak extern (adopted — this reshaped the PR), with one correction to the suggestion.** Availability is now the runtime binding of `@extern(..., .{ .linkage = .weak })` on comptime-false builds; the strong extern remains for comptime-true builds. Release binaries keep the posix_spawnp fast path for `directory:` on every glibc ≥ 2.29 host. **But the binding must be a `var`, not a `const`:** with `const`, Zig comptime-knows the optional non-null, so `addchdir_np != null` and the `if (...) |f|` unwrap fold at compile time and the call goes through the symbol unconditionally — on an unbound host that is a jump to null. I measured exactly that: the `const` form segfaults on debian:10 (glibc 2.28 — the very floor the release targets), while on glibc ≥ 2.29 a folded check is indistinguishable from a real one (a verification on ubuntu-24.04 alone cannot tell the difference). With `var`, the pointer lives in `.data` behind a load-time relocation and every null test is a runtime load. End-to-end proof on **both** sides of the 2.29 boundary with the gnu.2.28 build: debian:10 takes the fork route (repro prints `0`, `pwd` captures `/tmp`, bad `directory:` raises `(file-error ENOENT)` through the error pipe); ubuntu-24.04 takes the posix route (identical results, `readelf -sW` shows `WEAK DEFAULT UND posix_spawn_file_actions_addchdir_np`, and the spawn timings match plain spawns). One more documented nuance: glibc's posix route reports a bad `directory:` as a child exiting 127 (its own file-action-failure behavior, identical to main on any glibc ≥ 2.29 build), while the fork route raises the ENOENT file error synchronously like macOS — earlier "identical condition" wording overstated this and has been corrected everywhere.

**5. `routeFor` doesn't fold.** Your one-keyword suggestion applied: `pub inline fn routeFor`. The doc comment now says exactly what folds and what doesn't: `spawn_via_fork_exec` stays a `comptime` parameter, so on OpenBSD the decision folds to `.fork_exec` and the posix route is not analyzed — the pre-fallback link protection restored; on comptime-true builds the availability argument is comptime-known `true`, likewise folding; on gnu.2.28/NetBSD the availability is the runtime weak binding, so both routes are emitted and the branch is genuinely per-spawn. The routing-table test pins the new runtime rows. (The folding asymmetry is also why the first push's `close_range` flags literal — missing the non-defaulted `UNSHARE` field — only failed to compile in the CI test binaries, where the tests force `spawnChildForkExec` into analysis on a CI-native Linux build; fixed in d2380bed, and the Linux test binary now compiles.)

Housekeeping: `tests_process.zig` hit the 1500-line ceiling, so the fork-route cluster moved to `tests_process_fork.zig` (registered in `vm_tests.zig`) — the same seam move `tests_process_run.zig` once made. Evidence: `zig build test` 2033 passed / 21 skipped / 0 failed; process suites green under `-Dgc-stress=true`; `x86_64-linux-gnu.2.28`, the `x86_64-linux` test binary, `x86_64-openbsd`, `x86_64-netbsd` cross-builds and `zig build wasm` all compile and link; updated podman outputs (ubuntu-24.04 *and* debian:10) are in the PR body.

baijum:
These three comments landed at 2026-09-05T09:10Z, after the round addressed in the previous comment — so this maps them separately. All three are fixed in 45ed6d6d; each was verified against the code and toolchain before fixing, none skipped.

**1. [Trivial] Silent no-op if the routing invariant breaks.** Adopted your guard verbatim: the posix route's weak-addchdir arm now has an `else` returning `spawnSetupError(gc, @intFromEnum(std.c.E.NOSYS))`, with the comment noting `routeFor` must have diverted this spawn to the fork route. The comment above the branch was reworded to point at the belt instead of claiming the invariant enforces itself.

**2. [Major] CLOSE_RANGE flag encoding — confirmed real in our toolchain, fixed, and it was worse than a wrong flag.** Verified first, three independent ways:

- *Toolchain*: our Zig 0.16.0 `lib/std/os/linux.zig` has `CLOSE_RANGE` as `packed struct(u32) { UNSHARE: bool, CLOEXEC: bool, _: u30 = 0 }` — no leading pad. A local printout confirms `.{ .UNSHARE = false, .CLOEXEC = true }` packs to **2**, the kernel's `CLOSE_RANGE_UNSHARE` (`1<<1`), while real `CLOSE_RANGE_CLOEXEC` is `1<<2`.
- *Kernel probe*: on a real 6.12 kernel (native arm64 container, no Rosetta), the packed form **closes** a test fd; the raw `4` leaves it open with `FD_CLOEXEC`; raw `2` closes it. So the sweep was closing fds ≥ 3 outright — killing the error pipe, making any post-sweep exec failure unreportable: the spawn "succeeded" and the child died 127 (the #2456 regression, reintroduced on the fallback path).
- *strace* (native arm64 debian:10, glibc 2.28 → fork route; old strace prints the raw form `syscall_0x1b4`):

```
BEFORE: 222 syscall_0x1b4(0x3, 0xffffffffffffffff, 0x2, ...) = 0   # 2 = CLOSE_RANGE_UNSHARE (wrong)
AFTER:  222 syscall_0x1b4(0x3, 0xffffffffffffffff, 0x4, ...) = 0   # 4 = CLOSE_RANGE_CLOEXEC (right)
```

End-to-end symptom, same container: a missing program through `directory:` **"succeeded"** before the fix (`no-error`, child silently 127) and raises `(file-error ENOENT)` after. My earlier debian:10 x86_64 checks missed this because **Rosetta does not translate syscall 436** — `close_range` returned ENOSYS there and the probe fallback silently ran — and CI never combines the fork route with an exec failure on Linux. The fix passes the raw bit (`@bitCast(@as(u32, 1 << 2))`), keeps `first = 3` / `last = -1`, and carries a comment pinning the 0.16.0 encoding bug and the upstream fix (5ceec001, "std.os.linux: pad CLOSE_RANGE by one bit", 2026-08-08) so a Zig bump can revert to `.{ .CLOEXEC = true }`.

**3. [Major] childPathExec errno semantics.** The walk now matches the glibc `execvpe`/`posix_spawnp` contract exactly: the search continues **only** for `ENOENT`, `ESTALE`, `ENOTDIR`, `ENODEV`, `ETIMEDOUT`; `EACCES` is remembered while the search continues and is the reported errno for a fully failed search; every other error — `ELOOP`, `EIO`, `ENOMEM`, `E2BIG`, `ENOEXEC`, … — goes straight to `childExecFail` as the spawn's verdict instead of being masked by a fabricated `ENOENT`. `ENOEXEC` is deliberate pass-through, per your research: the default binding passes only `SPAWN_XFLAGS_USE_PATH` and the `/bin/sh` fallback is the compat `TRY_SHELL` path — so no shell fallback and no continued search. Two new tests in `tests_process_fork.zig` pin it through the real walk (PATH steered via a scoped `setenv`, original restored after): an executable file with no recognized format raises **ENOEXEC** — verified to fail with a fabricated `ENOENT` (expected 8, found 2) when the old continue-on-everything logic is restored — and an all-`EACCES` PATH search reports **EACCES**, not `ENOENT`. The strace above also shows the fixed walk doing its job: `chdir("/tmp")`, then the PATH candidates in order (`/usr/local/sbin/true` → … `ENOENT` … → `/bin/true` succeeding).

**Evidence:** `zig build test` 2035 passed / 21 skipped / 0 failed (two new tests); process suites green under `-Dgc-stress`; wasm and the aarch64 + x86_64 `gnu.2.28` cross-builds link; on native arm64 debian:10 the full matrix passes through the fork route (repro `0`, pwd `/tmp`, bad directory `(file-error ENOENT)`, missing program `(file-error ENOENT)`, `env:`+`directory:` parity `0`); the glibc-2.39 posix route is byte-for-byte unchanged (matrix identical, 50 `directory:` spawns still ~12 ms/spawn).

PR #2529, opened 2026-09-05
Default -Dbundle builds to the portable baseline CPU

### LINKED ISSUE #2515: zig build with no -Dtarget tunes to the host CPU, so -Dbundle= binaries SIGILL elsewhere
## What happens

`zig build` with no `-Dtarget` compiles kaappi for the **host's exact CPU
model**, not a portable baseline. A binary produced that way — including a
standalone binary made with `-Dbundle=` — can die with SIGILL on another
machine of the same architecture.

## Demonstration

`build.zig:5` uses `b.standardTargetOptions(.{})`, so with no `-Dtarget` the
target is native, and native includes the host CPU model and its feature
set. A three-line program shows what that resolves to:

```zig
const std = @import("std");
const builtin = @import("builtin");
pub fn main() !void {
    std.debug.print("cpu model: {s}\n", .{builtin.cpu.model.name});
}
```

```
$ zig build-exe m.zig && ./m                                  # no -target
cpu model: apple_m3

$ zig build-exe m.zig -target aarch64-macos && ./m            # -target, no cpu
cpu model: apple_m1

$ zig build-exe m.zig -target aarch64-macos -mcpu=baseline && ./m
cpu model: apple_m1
```

The first is the host. The other two are the portable baseline for the
target.

## Why it matters for kaappi specifically

kaappi's own release workflow always passes `-Dtarget=`
(`.github/workflows/release.yml:179`), so **published kaappi binaries are
not affected**. This is about downstream projects using the standalone
binary feature, which is a kaappi feature and the documented way to ship a
Kaappi program.

The natural way to build one for the machine you are on is:

```
kaappi --compile -o app.sbc app.scm
zig build -Dbundle=app.sbc -Doptimize=ReleaseSafe
```

No `-Dtarget`, because you want the host. That produces a binary tuned to
the exact CPU of the build machine.

We hit this shipping mesthiri. Release binaries built on GitHub-hosted
runners crashed with SIGILL on *other* GitHub-hosted runners, because the
x86_64 pool is heterogeneous — a binary built on one microarchitecture met
a machine without those instructions. It cost two wrong hypotheses first (a
faulty container, then a suspected `set-car!` on a literal) because SIGILL
in a bytecode VM reads as a VM bug, not a build-flags bug. The fix was
`-Dcpu=baseline` on every `zig build` invocation.

## Suggested resolution

Any of these; the first two are cheap.

- **Document it** where `-Dbundle=` is documented: a standalone binary
  intended to run anywhere but this machine needs `-Dcpu=baseline` (or an
  explicit `-Dtarget`).
- **Warn** when `-Dbundle=` is used with a native target and non-baseline
  CPU features — that combination is nearly always meant for distribution.
- **Default `-Dbundle=` builds to baseline**, with an opt-out for people who
  really are building only for themselves. The performance kaappi gives up
  is small next to a crash that presents as a VM bug.

## Version

Kaappi Scheme v0.26.1, Zig 0.16, macOS ARM. SIGILL observed on
GitHub-hosted `ubuntu-latest` x86_64 runners.

baijum:
## Summary

`zig build` with no `-Dtarget` tunes for the **build host's exact CPU model**, so a standalone binary built with `-Dbundle=`/`-Dbundle-src=` — the documented way to ship a Kaappi program — could SIGILL on another machine of the same architecture, which presents as a VM bug rather than a build-flags bug (#2515). Published kaappi binaries are unaffected (the release workflow always passes `-Dtarget`); this hits downstream users of the standalone-binary feature.

This implements the issue's preferred resolution (option 3): **bundle builds default to the portable baseline CPU model**; plain `zig build` without bundle options stays host-tuned.

- `-Dbundle=` / `-Dbundle-src=` (with or without an explicit `-Dtarget=`) resolve `.cpu_model = .baseline` when the user did not name a CPU. An explicit `-Dtarget` arch/os/abi is preserved verbatim (Zig already defaults explicit targets to baseline without `-Dcpu`, so pinning is a no-op there).
- `-Dcpu=<model>` (including `-Dcpu=baseline`) is always respected. **`-Dcpu=native` is the opt-out** that restores host tuning for a binary that really will only run on the machine that built it — Zig's own spelling rather than a bundle-only flag, per review: it is the pre-fix tuning bit for bit (`.native` → the same `detectNativeCpuAndFeatures` call the unpinned query makes) and, unlike a flag that merely skipped the pin, it also delivers host tuning under an explicit `-Dtarget=`.
- Both `-Dbundle` build-option help strings state the default and the `-Dcpu=native` opt-out, so `zig build --help` communicates it.
- The CPU model is **not** part of the bytecode compiler key (`bytecode_file.compile_target_id` hashes arch/os/abi + `cond-expand` features only), so an `.sbc` produced by a host-tuned `kaappi --compile` still embeds in a baseline bundler from the same tree — the documented two-command bundle flow keeps working, and the #2514 foreign-build diagnostics never fire over a CPU difference.

Closes #2515

## Scope / follow-up

This PR fixes the **bundle** path only. The same hazard lives in the runtime archive: the `lib` step builds `libkaappi_rt.a` with the same host-tuned `target`, so a source-built `kaappi compile app.scm -o app` ships a host-tuned VM/GC inside every native executable (released archives are baseline — release.yml passes `-Dtarget`). That gap is tracked in #2531 rather than extended into this PR; `Closes #2515` above is not meant to close that hazard class.

## Verification

CPU models on this machine (the issue's probe technique — host is an Apple M3, baseline for aarch64-macos is apple_m1):

```
$ zig build-exe probe.zig && ./probe        # prints builtin.cpu.model vs Target.Cpu.baseline
host cpu model: apple_m3
baseline cpu model: apple_m1
```

Differential proof (Zig builds are reproducible within one build-graph state, so equal inputs ⇒ byte-identical binaries):

```
zig build -Dbundle=hello.sbc --prefix P-default            # no flags
zig build -Dbundle=hello.sbc -Dcpu=baseline --prefix P-baseline
zig build -Dbundle=hello.sbc -Dcpu=native --prefix P-native
zig build -Dcpu=baseline --prefix P-plain-baseline          # plain, no bundle

cmp P-default/bin/kaappi P-baseline/bin/kaappi     -> IDENTICAL  (default IS baseline)
cmp P-default/bin/kaappi P-native/bin/kaappi       -> DIFFER, ~144 KB of bytes (host tuning restored)
cmp zig-out/bin/kaappi P-plain-baseline/bin/kaappi -> DIFFER     (plain build stays host-tuned)
```

All bundled binaries run correctly (`hello from a bundled binary`, exit 0), including the default one whose `.sbc` was produced by the host-tuned plain binary — the compiler-key compatibility property above. `zig build -Dbundle-src=hello.scm` (single-step) also builds and runs.

Existing bundle flows (re-run after the review round, commit 45e92e09):

```
tests/scheme/compile/bundle-cpu-baseline-2515.sh    PASS
tests/scheme/compile/bundle-args-2010.sh            PASS
tests/scheme/compile/srfi-241-202-bundle-2391.sh    PASS
tests/scheme/compile/compile-preamble-gc-700.sh     exit 0
tests/scheme/compile/compile-import-error-703.sh    exit 0
tests/scheme/compile/compile-define-values-order-2200.sh    PASS (3/3)
tests/scheme/compile/compile-toplevel-side-effects-2156.sh  PASS (12/12)
tests/scheme/compile/native-external-library-import-1743.sh PASS
tests/scheme/cache/build-id-untracked-2097.sh       SKIP (77): its own pristine-tree
   premise does not hold in a dirty worktree; runs on a clean CI checkout
zig build test                                       exit 0 (macOS)
```

`zig fmt --check build.zig` clean; `npx markdownlint-cli2` 0 issues.

## Regression test

`tests/scheme/compile/bundle-cpu-baseline-2515.sh` — a build-system default has no Scheme-visible surface and the binary carries no CPU-model tag, so the test is differential: the default bundle binary must be byte-identical to an explicit `-Dcpu=baseline` one (fails without the fix wherever host ≠ baseline — CI's x86_64 legs, Apple Silicon above M1 — and is vacuously true where host = baseline, where the bug cannot manifest), and must differ from a `-Dcpu=native` one wherever the two tunings differ, detected with the issue's own CPU-model probe so the script knows which world it is in. Verified both ways after the review round: PASS with the fix; against pre-fix `build.zig` it exits 1 at the core assertion (`default -Dbundle build is not the -Dcpu=baseline build`).

## Follow-up

- #2531 tracks the `libkaappi_rt.a` / `kaappi compile` half of the hazard (see Scope above).
- The end-user standalone-binary guide lives in kaappi/kaappi.github.io (out of scope here by design) and needs a matching note there that bundles default to the baseline CPU and that `-Dcpu=native` is the opt-out.

🤖 This PR was prepared with ZCode assistance.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -2,7 +2,46 @@ const std = @import("std");
 const zon = @import("build.zig.zon");
 
 pub fn build(b: *std.Build) void {
-    const target = b.standardTargetOptions(.{});
+    // Standard -Dtarget/-Dcpu/-Dofmt/-Ddynamic-linker, parsed to a query but
+    // NOT yet resolved: the bundle path below may retune the CPU model first
+    // (kaappi#2515). Same option registration `standardTargetOptions` performs,
+    // so the non-bundle path below resolves to exactly what it always did.
+    const target_query = b.standardTargetOptionsQueryOnly(.{});
+
+    // Standalone binary: embed compiled bytecode via -Dbundle=path/to/program.sbc.
+    // Bundled builds default to the portable baseline CPU (kaappi#2515): a
+    // standalone binary is a thing you *ship*, and `zig build` with no -Dtarget
+    // otherwise tunes for the build host's exact CPU model, so the binary can
+    // SIGILL on another machine of the same architecture — which presents as a
+    // VM bug, not a build-flags bug. -Dbundle-cpu-native opts back into host
+    // tuning for a binary that really will only ever run here.
+    const bundle = b.option([]const u8, "bundle", "Path to .sbc bytecode file to embed for standalone binary (defaults to the portable baseline CPU so it runs on other machines of the same arch; -Dbundle-cpu-native restores host tuning)");
+    // Single-step: compile and embed in one build
+    const bundle_src = b.option([]const u8, "bundle-src", "Path to .scm source file to compile and embed for standalone binary (defaults to the portable baseline CPU so it runs on other machines of the same arch; -Dbundle-cpu-native restores host tuning)");
+    const bundle_cpu_native = b.option(bool, "bundle-cpu-native", "Tune a -Dbundle/-Dbundle-src binary for this machine's exact CPU model instead of the portable baseline default (kaappi#2515); such a binary can SIGILL on other machines of the same architecture") orelse false;
+
+    // kaappi#2515: when bundling, retune the query to the baseline CPU model
+    // unless the user named a CPU themselves. `.determined_by_arch_os` is the
+    // "user didn't say" state — Zig resolves it to the *host* model when no
+    // -Dtarget was given (that is the bug) and to baseline for an explicit
+    // -Dtarget, so pinning it to `.baseline` makes both spellings land on
+    // baseline while `-Dcpu=<model>`/`-Dcpu=native` (`.explicit`/`.native`) and
+    // `-Dcpu=baseline` are left exactly as passed. This tunes every artifact of
+    // a bundling configure (the embedded binary, its build-time
+    // kaappi-compiler, and the always-installed thottam/kaappi-lsp) — one
+    // configure, one target. Plain `zig build` without bundle options is
+    // untouched and stays host-tuned: that binary stays on the machine that
+    // built it. The CPU model is not part of the bytecode compiler key
+    // (`bytecode_file.compile_target_id` hashes arch/os/abi + cond-expand
+    // features only), so a baseline bundler still accepts an .sbc produced by
+    // a host-tuned kaappi from the same tree.
+    const bundling = bundle != null or bundle_src != null;
+    var effective_query = target_query;
+    if (bundling and !bundle_cpu_native and target_query.cpu_model == .determined_by_arch_os) {
+        effective_query.cpu_model = .baseline;
+    }
+    const target = b.resolveTargetQuery(effective_query);
+
     // Default to ReleaseSafe rather than Debug: the interpreter exists to *run*
     // Scheme programs, and Debug is ~500x slower for allocation/continuation-
     // heavy workloads. ReleaseSafe matches ReleaseFast in throughput here while
@@ -14,11 +53,6 @@ pub fn build(b: *std.Build) void {
         "Prioritize performance, safety, or binary size (default: ReleaseSafe)",
     );
     const optimize = optimize_opt orelse .ReleaseSafe;
-
-    // Standalone binary: embed compiled bytecode via -Dbundle=path/to/program.sbc
-    const bundle = b.option([]const u8, "bundle", "Path to .sbc bytecode file to embed for standalone binary");
-    // Single-step: compile and embed in one build
-    const bundle_src = b.option([]const u8, "bundle-src", "Path to .scm source file to compile and embed for standalone binary");
 
     const do_strip = b.option(bool, "strip", "Strip debug info from binaries (for release builds)") orelse false;
     const strip: ?bool = if (do_strip) true else null;
--- a/build.zig
+++ b/build.zig
@@ -13,12 +13,11 @@ pub fn build(b: *std.Build) void {
     // standalone binary is a thing you *ship*, and `zig build` with no -Dtarget
     // otherwise tunes for the build host's exact CPU model, so the binary can
     // SIGILL on another machine of the same architecture — which presents as a
-    // VM bug, not a build-flags bug. -Dbundle-cpu-native opts back into host
-    // tuning for a binary that really will only ever run here.
-    const bundle = b.option([]const u8, "bundle", "Path to .sbc bytecode file to embed for standalone binary (defaults to the portable baseline CPU so it runs on other machines of the same arch; -Dbundle-cpu-native restores host tuning)");
+    // VM bug, not a build-flags bug. -Dcpu=native opts back into host tuning
+    // for a binary that really will only ever run here.
+    const bundle = b.option([]const u8, "bundle", "Path to .sbc bytecode file to embed for standalone binary (defaults to the portable baseline CPU so it runs on other machines of the same arch; -Dcpu=native restores host tuning)");
     // Single-step: compile and embed in one build
-    const bundle_src = b.option([]const u8, "bundle-src", "Path to .scm source file to compile and embed for standalone binary (defaults to the portable baseline CPU so it runs on other machines of the same arch; -Dbundle-cpu-native restores host tuning)");
-    const bundle_cpu_native = b.option(bool, "bundle-cpu-native", "Tune a -Dbundle/-Dbundle-src binary for this machine's exact CPU model instead of the portable baseline default (kaappi#2515); such a binary can SIGILL on other machines of the same architecture") orelse false;
+    const bundle_src = b.option([]const u8, "bundle-src", "Path to .scm source file to compile and embed for standalone binary (defaults to the portable baseline CPU so it runs on other machines of the same arch; -Dcpu=native restores host tuning)");
 
     // kaappi#2515: when bundling, retune the query to the baseline CPU model
     // unless the user named a CPU themselves. `.determined_by_arch_os` is the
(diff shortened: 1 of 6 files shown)


build.zig
```diff
@@ -2,7 +2,46 @@ const std = @import("std");
...
+    // otherwise tunes for the build host's exact CPU model, so the binary can
+    // SIGILL on another machine of the same architecture — which presents as a
+    // VM bug, not a build-flags bug. -Dbundle-cpu-native opts back into host
+    // tuning for a binary that really will only ever run here.
+    const bundle = b.option([]const u8, "bundle", "Path to .sbc bytecode file to embed for standalone binary (defaults to the portable baseline CPU so it runs on other machines of the same arch; -Dbundle-cpu-native restores host tuning)");
+    // Single-step: compile and embed in one build
+    const bundle_src = b.option([]const u8, "bundle-src", "Path to .scm source file to compile and embed for standalone binary (defaults to the portable baseline CPU so it runs on other machines of the same arch; -Dbundle-cpu-native restores host tuning)");
+    const bundle_cpu_native = b.option(bool, "bundle-cpu-native", "Tune a -Dbundle/-Dbundle-src binary for this machine's exact CPU model instead of the portable baseline default (kaappi#2515); such a binary can SIGILL on other machines of the same architecture") orelse false;
```
baijum: **`-Dbundle-cpu-native` duplicates `-Dcpu=native`, and does not do what its help says under an explicit `-Dtarget`.**

In `std.zig.system.resolveTargetQuery` (Zig 0.16):

```zig
var cpu = switch (query.cpu_model) {
    .native => detectNativeCpuAndFeatures(io, query_cpu_arch, os, query),
    .baseline => Target.Cpu.baseline(query_cpu_arch, os),
    .determined_by_arch_os => if (query.cpu_arch == null)
        detectNativeCpuAndFeatures(io, query_cpu_arch, os, query)
    else
        Target.Cpu.baseline(query_cpu_arch, os),
    ...
```

So `-Dcpu=native` is the pre-fix tuning bit for bit — the same call with the same arguments — and the PR already guarantees `-Dcpu=native` is respected. The new flag buys nothing over the standard Zig spelling, and it costs a build option, a help string, a CLAUDE.md sentence and a test-runner rule to maintain, plus it is silently ignored when no bundle option is present.

It also has a wrinkle the standard spelling does not: `zig build -Dbundle=x.sbc -Dtarget=aarch64-macos -Dbundle-cpu-native` leaves `cpu_model == .determined_by_arch_os` with an explicit arch, which the table above resolves to **baseline** — the flag's help ("tune for this machine's exact CPU model") is false in that spelling, while `-Dcpu=native` delivers host tuning in both.

Suggest dropping the flag: keep the `.determined_by_arch_os → .baseline` pin, document `-Dcpu=native` as the opt-out (CLAUDE.md, README, the two help strings, test-runner.md), and make the regression script's `native` variant `build_variant native -Dcpu=native`. The differential test discriminates exactly the same way.

build.zig
```diff
@@ -2,7 +2,46 @@ const std = @import("std");
...
+    // `-Dcpu=baseline` are left exactly as passed. This tunes every artifact of
+    // a bundling configure (the embedded binary, its build-time
+    // kaappi-compiler, and the always-installed thottam/kaappi-lsp) — one
+    // configure, one target. Plain `zig build` without bundle options is
+    // untouched and stays host-tuned: that binary stays on the machine that
+    // built it. The CPU model is not part of the bytecode compiler key
+    // (`bytecode_file.compile_target_id` hashes arch/os/abi + cond-expand
+    // features only), so a baseline bundler still accepts an .sbc produced by
+    // a host-tuned kaappi from the same tree.
+    const bundling = bundle != null or bundle_src != null;
```
baijum, line 42: **Altitude: the same hazard is still in the runtime archive that `kaappi compile` links.**

The pin is gated on `bundling`, but the *other* documented way to ship a Kaappi program — `kaappi compile program.scm -o program` — links `libkaappi_rt.a`, and the `lib` step builds that archive with the same `target` (`build.zig:264`, `.target = target`). A from-source `zig build lib` therefore produces a host-tuned runtime (the whole VM and GC) inside every native executable a source-built kaappi emits.

Scenario: build kaappi from source on an AVX-512 x86_64 host, `kaappi compile app.scm -o app`, ship `app` to another x86_64 machine — SIGILL in the runtime, presenting as a VM bug exactly as in #2515. Released `libkaappi_rt.a` archives are baseline (release.yml passes `-Dtarget`), so like the bundle case this hits source builds only — the same population this fix targets.

Two options: extend the baseline default to the `lib` step too (or to the whole build, with `-Dcpu=native` as the dev opt-in — plain `zig-out/bin/kaappi` also gets copied between machines in practice, see the A/B testing protocol), or file the follow-up so the gap is tracked rather than read as closed by "Closes #2515". Not blocking; the bundle fix stands on its own.

baijum (commented):
Reviewed the diff against `std.Build` and `std.zig.system` in Zig 0.16 to check the removed-behavior claims. Both hold: `standardTargetOptions` is literally `standardTargetOptionsQueryOnly` + `resolveTargetQuery`, so the non-bundle path is unchanged, and `compile_target_id` hashes arch/os/abi + features only, so the compiler-key compatibility claim is right. The differential test is sound and the runner's longest-first grep (`zig build -D`) picks it up.

Two comments inline, neither a correctness bug in what the PR does:

1. **`-Dbundle-cpu-native` is redundant with `-Dcpu=native`** (same std resolution path), and unlike `-Dcpu=native` it resolves to baseline when combined with an explicit `-Dtarget`. Suggest dropping it.
2. **Altitude:** `libkaappi_rt.a` built from source is still host-tuned, so `kaappi compile` output from a source build carries the same SIGILL hazard. Extend the default or track it as a follow-up.

baijum:
Both review comments addressed in 45e92e09.

**1. Dropped `-Dbundle-cpu-native`; `-Dcpu=native` is the documented opt-out.** Verified your `resolveTargetQuery` reading against the 0.16 std source before changing anything — `.native` and an unpinned `.determined_by_arch_os` query (native arch) both land in the same `detectNativeCpuAndFeatures` call with the same arguments, so the standard spelling is the pre-fix tuning bit for bit, and under an explicit `-Dtarget=` only `.native` keeps host detection. Concret:

- `build.zig`: the `bundle-cpu-native` option is gone; the `.determined_by_arch_os → .baseline` pin is kept verbatim (`if (bundling and target_query.cpu_model == .determined_by_arch_os)`), and the comment block now explains why the opt-out is Zig's own flag rather than a bespoke one.
- Help strings on `-Dbundle`/`-Dbundle-src` now say `…; -Dcpu=native restores host tuning`.
- Docs: CLAUDE.md (AGENTS.md symlinks to it), README feature bullet, and the test-runner.md cache-key rule (now "no script may pass `-Dcpu=native` (or any host-naming `-Dcpu=<model>`) to a `-Dbundle=` build").
- Regression script: `build_variant native -Dcpu=native`. The differential test discriminates exactly as before — verified both directions after the change: PASS with the fix (`default == -Dcpu=baseline` byte-identical; `default` vs `-Dcpu=native` differ by ~144 KB of bytes), and FAIL against pre-fix `build.zig`, now at the core assertion (`default -Dbundle build is not the -Dcpu=baseline build`), which is the better failure since the old flag used to fail earlier on the unknown option.

One measurement note from checking the "bit for bit" claim empirically: cross-revision byte comparison of binaries carries ~180 bytes of noise (same flags rebuilt after a build.zig text change differ by 183 bytes — generated-file cache paths shift in DWARF), while different CPU tunings differ by ~144 KB; so the tuning equivalence is demonstrated within one build-graph state, not by comparing binaries across revisions.

**2. Runtime-archive gap tracked, not extended.** Filed #2531 with your scenario (AVX-512 host → `kaappi compile` → ship → SIGILL reading as a VM bug), the released-archive non-exposure (release.yml passes `-Dtarget`), and both options you raised — extending the baseline default to the `lib` step, or to the whole build with `-Dcpu=native` as the dev opt-in, noting plain `zig-out/bin/kaappi` also travels between machines per the A/B protocol in `docs/dev/performance.md`. Labels: `bug`, `tooling`, one priority label `priority: medium` (same class as #2515). The PR body now has a "Scope / follow-up" section referencing #2531 so `Closes #2515` doesn't read as closing the whole hazard class.

Re-verified after the change: all bundle shell tests listed in the PR body pass, `zig build test` exit 0, `zig fmt --check build.zig` clean, markdownlint 0 issues.

PR #2534, opened 2026-09-05
Show the interpreter oracle's stderr on unexpected failure

### LINKED ISSUE #2532: ubuntu-Debug CI leg: interpreter oracle of native-set-global-child-thread-2487.sh aborted once (exit 134, stderr dropped by harness)
### What happened

Run [33958780190](https://github.com/kaappi/kaappi/actions/runs/33958780190) (PR #2528, head `65e8d13d`), job `test (ubuntu-latest, Debug)`, 2026-09-05T10:15:09Z:

```
FAIL  tests/scheme/compile/native-set-global-child-thread-2487.sh
FAIL: native-set-global-child-thread-2487 — interpreter stdout '' exit 134, expected '200000' exit 0
```

Exit 134 is SIGABRT. Empty stdout is consistent with an abort **anywhere** in the run — stdout to a pipe is block-buffered and the buffer is lost on abort — so it does not localize the crash. The same test passed in the immediate re-run of the same job (749 pass / 0 fail on the leg) and in every other leg of the same commit.

### Why this is tracked rather than absorbed

- **Not PR-caused.** #2528's changes are spawn-path only (spawn routing, `close_range`, PATH walk, error-pipe lift); the test program spawns no processes, and the oracle runs the interpreter in its own fresh process, so nothing earlier in the suite can contaminate it.
- **One occurrence, unreproduced.** The same script was green on every other leg of the same head (ubuntu ReleaseSafe/ReleaseFast/arm, macOS, FreeBSD/OpenBSD/NetBSD, the QEMU legs) and in three earlier CI runs of sibling heads; 10/10 consecutive local Debug runs of the exact oracle program (macOS arm64) all print `200000` exit 0.
- **The abort reason is unrecoverable.** `interp_stdout` (`tests/scheme/shell-common.sh:127-135`) drops stderr **by design** — the two tiers' diagnostics differ in framing — so a Zig panic or allocator report never reaches the job log.

Two hypotheses, indistinguishable with the evidence we have:

1. **A timing-dependent abort in a race path.** The test is a probabilistic race stress by construction: a child thread doing 200k native `set!`s on a shared global while the root evals 10k fresh defines that rehash the shared globals map. Under Debug on a loaded x86_64 runner some window may still abort (the #2487 lock fixed the known one; this would be a residual).
2. **An abort at interpreter exit** — e.g. a `DebugAllocator` report on that timing. The same leg's known failure mode on `main` (run [33811986882](https://github.com/kaappi/kaappi/actions/runs/33811986882), 2026-09-03) was exactly a `error(DebugAllocator): memory leaked` report inside this suite.

### Catching a recurrence

When the interpreter oracle of a compile script fails unexpectedly, its stderr is the single most valuable datum and we currently throw it away. `interp_stdout` could redirect stderr to a temp file and callers could `cat` it only on the unexpected-failure branch — the parity rationale (never compare diagnostic text across tiers) is untouched, since the text would be shown, not asserted on. If this failure recurs, the panic text will say which hypothesis is right.

**Priority rationale:** `priority: medium` per `docs/dev/github-issues.md` — loud (a red CI leg) but narrow (once, one leg, unreproduced, no demonstrated user-facing defect).

baijum:
Closes #2532

## What

Implements the issue's "Catching a recurrence" prescription: when an interpreter oracle run in a compile script fails unexpectedly, its stderr — the one datum that can explain an abort — now lands in the CI log instead of `/dev/null`.

- `interp_stdout` (tests/scheme/shell-common.sh) takes an optional fourth argument, an errfile to capture the interpreter's stderr into; the default stays `/dev/null`.
- New `show_interp_stderr <errfile>` prints the capture with a `--- interpreter stderr ---` lead-in, and is a no-op when the file is empty — so a green run stays silent and a mismatch that is only the native tier's fault adds nothing.
- All 18 oracle call sites across the 16 compile scripts now pass an errfile (under each script's mktemp workdir, never inside the repo tree — a dirty tree changes the build id baked into `.sbc` files) and call `show_interp_stderr` on their interpreter-failure FAIL branches.
- `tests/scheme/smoke/interp-oracle-stderr-2532.sh` is a deterministic harness self-test (the `build-lock-orphan-2434.sh` shape, no real kaappi): a stub interpreter reproduces the CI failure's shape — panic line on stderr, no stdout, exit 134 — and the test asserts the capture and the showing, that the 3-arg form still buries stderr (proven with the noisy stub), and that quiet runs behave exactly as before.
- `tests/scheme/CLAUDE.md`'s oracle section documents the errfile convention.

## Why

Run 33958780190 aborted once in the interpreter oracle of `native-set-global-child-thread-2487.sh` (exit 134, empty stdout) and the two hypotheses — a residual race abort vs. an exit-time allocator report — were indistinguishable because `interp_stdout` dropped stderr by design. An aborting process loses its piped stdout to block buffering, so the captured stderr is the only thing that says *why* it died. If the failure recurs, the panic text now reaches the job log.

The tier-parity rule (never *compare* diagnostic text across tiers) is untouched: the text is shown on the failure branch, never asserted on.

## Verification

- Self-test verified both ways: red against `origin/main`'s `shell-common.sh` (a single clean FAIL line), green with the fix.
- End-to-end: pointing `native-macro-set-tracking-2212.sh` at a stub interpreter that aborts with exit 134 + panic text — every FAIL branch now carries `--- interpreter stderr ---` and the panic lines.
- All 16 touched compile scripts pass against a fresh ReleaseSafe build (run with an isolated `KAAPPI_HOME`, matching `run-all.sh`), plus `bash -n` on every touched script and `markdownlint-cli2` clean.

If the abort recurs on a Debug leg after this lands, file a fresh issue with the now-visible panic text — it should pick between the issue's two hypotheses directly.

### DIFF
--- a/tests/scheme/compile/bundle-args-2010.sh
+++ b/tests/scheme/compile/bundle-args-2010.sh
@@ -44,11 +44,16 @@ FAIL=0
 # golden on purpose: bundled (command-line) intentionally differs from direct
 # source execution (no script path, kaappi#2010).
 interp_status=0
-interp_out="$(interp_stdout "$KAAPPI_ABS" "$FIXTURE" "$FIXTURE/main.scm")" || interp_status=$?
+# The errfile goes under $DIR, never the fixture tree: interp_stdout would
+# happily write it in the workdir, and a stray file inside the repo changes
+# the build id baked into every .sbc.
+interp_err="$DIR/interp-oracle.err"
+interp_out="$(interp_stdout "$KAAPPI_ABS" "$FIXTURE" "$FIXTURE/main.scm" "$interp_err")" || interp_status=$?
 bundle_status=0
 bundle_out="$("$BUNDLE_BIN" 2>&1)" || bundle_status=$?
 if ! assert_tiers_agree "bundled vs interpreter (no args)" \
     "$interp_out" "$interp_status" "$bundle_out" "$bundle_status"; then
+    show_interp_stderr "$interp_err"
     exit 1
 fi
 
--- a/tests/scheme/compile/bundle-args-2010.sh
+++ b/tests/scheme/compile/bundle-args-2010.sh
@@ -44,9 +44,10 @@ FAIL=0
 # golden on purpose: bundled (command-line) intentionally differs from direct
 # source execution (no script path, kaappi#2010).
 interp_status=0
-# The errfile goes under $DIR, never the fixture tree: interp_stdout would
-# happily write it in the workdir, and a stray file inside the repo changes
-# the build id baked into every .sbc.
+# The errfile goes under $DIR, never the fixture tree: interp_stdout applies
+# the redirect before its cd, so a relative path would land in the script's
+# own cwd (the repo root under run-all.sh) — and a stray file inside the
+# repo changes the build id baked into every .sbc.
 interp_err="$DIR/interp-oracle.err"
 interp_out="$(interp_stdout "$KAAPPI_ABS" "$FIXTURE" "$FIXTURE/main.scm" "$interp_err")" || interp_status=$?
 bundle_status=0
--- a/tests/scheme/shell-common.sh
+++ b/tests/scheme/shell-common.sh
@@ -124,14 +124,35 @@ skip_on_debug_build() {
 #   3. A procedure value prints as `#<procedure name>` in the VM and
 #      `#<procedure>` natively.
 #
-# interp_stdout <kaappi> <workdir> <src>: run <src> through the interpreter
-# from <workdir>, print its stdout, and return its own exit status. stderr is
-# dropped deliberately: the two tiers' diagnostics differ in framing by design
-# (the VM prefixes file:line and a source excerpt, the native runtime does
-# not), so a script asserting on a diagnostic's text does that separately,
-# against whichever tier it means.
+# interp_stdout <kaappi> <workdir> <src> [<errfile>]: run <src> through the
+# interpreter from <workdir>, print its stdout, and return its own exit
+# status. stderr goes to <errfile> when given, /dev/null otherwise.
+#
+# An oracle caller should pass an errfile — under its own mktemp workdir,
+# never inside the repo tree (a dirty tree changes the build id baked into
+# every .sbc, see the lock comment below) — and hand it to show_interp_stderr
+# on its unexpected-failure branch. An aborting interpreter loses its piped
+# stdout entirely (block buffering), so the captured stderr — a Zig panic
+# trace, an allocator leak report — is the one datum that says WHY it died,
+# and dropping it left a one-off CI abort unlocalizable (kaappi#2532).
+#
+# The tier-parity rationale is untouched by this: the two tiers' diagnostics
+# differ in framing by design (the VM prefixes file:line and a source
+# excerpt, the native runtime does not), so a script asserting on a
+# diagnostic's text does that separately, against whichever tier it means.
+# Capturing stderr to SHOW it on failure asserts nothing.
 interp_stdout() {
-    (cd "$2" && "$1" "$3" 2> /dev/null)
+    (cd "$2" && "$1" "$3") 2> "${4:-/dev/null}"
+}
+
+# show_interp_stderr <errfile>: print what interp_stdout captured, so a FAIL
+# branch carries the interpreter's own diagnostics into the CI log. No-op
+# when the file is empty or missing — the expected case on a green run, and
+# for a mismatch that is the native tier's fault alone.
+show_interp_stderr() {
+    [ -s "$1" ] || return 0
+    echo "--- interpreter stderr ---" >&2
+    cat "$1" >&2
 }
 
 # assert_tiers_agree <label> <interp-out> <interp-status> <native-out> <native-status>
--- a/tests/scheme/shell-common.sh
+++ b/tests/scheme/shell-common.sh
@@ -141,6 +141,15 @@ skip_on_debug_build() {
 # excerpt, the native runtime does not), so a script asserting on a
 # diagnostic's text does that separately, against whichever tier it means.
 # Capturing stderr to SHOW it on failure asserts nothing.
+#
+# The redirect sits on the subshell rather than on the kaappi command: it is
+# set up in the caller's directory (a relative errfile resolves against the
+# script's cwd, not the workdir interp_stdout cds into), and a failed `cd`'s
+# message lands in the errfile too. For errfile callers that is an
+# improvement — the FAIL branch shows it — but the 3-arg form now buries it
+# in /dev/null where it used to reach the script's stderr. Only the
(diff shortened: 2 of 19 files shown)


tests/scheme/compile/bundle-args-2010.sh
```diff
@@ -44,11 +44,16 @@ FAIL=0
 # golden on purpose: bundled (command-line) intentionally differs from direct
 # source execution (no script path, kaappi#2010).
 interp_status=0
-interp_out="$(interp_stdout "$KAAPPI_ABS" "$FIXTURE" "$FIXTURE/main.scm")" || interp_status=$?
+# The errfile goes under $DIR, never the fixture tree: interp_stdout would
+# happily write it in the workdir, and a stray file inside the repo changes
```
baijum: Nit (comment accuracy): with the redirect now *outside* the subshell, a relative errfile resolves against the script's own cwd, not the workdir `interp_stdout` cds into. Under `run-all.sh` that cwd is the repo root, so the conclusion ("a stray file inside the repo changes the build id") still holds — it just lands somewhere other than where this comment says. Suggest: "a relative errfile would land in the script's cwd (the repo root under run-all.sh), and a stray file inside the repo changes …".

tests/scheme/shell-common.sh
```diff
@@ -124,14 +124,35 @@ skip_on_debug_build() {
...
+# and dropping it left a one-off CI abort unlocalizable (kaappi#2532).
+#
+# The tier-parity rationale is untouched by this: the two tiers' diagnostics
+# differ in framing by design (the VM prefixes file:line and a source
+# excerpt, the native runtime does not), so a script asserting on a
+# diagnostic's text does that separately, against whichever tier it means.
+# Capturing stderr to SHOW it on failure asserts nothing.
 interp_stdout() {
-    (cd "$2" && "$1" "$3" 2> /dev/null)
+    (cd "$2" && "$1" "$3") 2> "${4:-/dev/null}"
```
baijum, line 154: Worth a line in the comment above: moving `2>` outside the subshell also routes a failed `cd`'s message into the errfile. For the 4-arg callers that is an improvement (it gets shown on the FAIL branch), but the 3-arg form now sends it to `/dev/null`, where before it reached the script's stderr. No real caller uses the 3-arg form after this PR (only the self-test), so this is fine as-is — just documenting the trade so the next person doesn't move the redirect back "to fix" it.

tests/scheme/smoke/interp-oracle-stderr-2532.sh
```diff
@@ -0,0 +1,99 @@
...
+    echo "FAIL: interp_stdout leaked stderr text into stdout: '$out'" >&2
+    exit 1
+fi
+if [ "$status" -ne 134 ]; then
+    echo "FAIL: interp_stdout returned $status, expected 134" >&2
+    exit 1
+fi
+
+# The point of the fix: the stderr text survives in the errfile...
+if ! grep -q "stub panic" "$ERR"; then
```
baijum: Nit: on the red path (old `shell-common.sh`, which never creates the errfile) this prints a stray `grep: …/oracle.interp.err: No such file or directory` before the FAIL line. `grep -qs` (or `[ -s "$ERR" ] && grep -q …`) keeps the red diagnostic to the one line that matters. Verified by running the script against main's `shell-common.sh`:

```
grep: /var/folders/…/oracle.interp.err: No such file or directory
FAIL: interp_stdout dropped the interpreter's stderr (kaappi#2532)
```

baijum (commented):
Reviewed against the PR head (9eb7270). Looks good to merge; three inline nits, none blocking.

**Verified locally**

- Self-test is red against main's `shell-common.sh` (`FAIL: interp_stdout dropped the interpreter's stderr`) and green on the branch.
- All 18 `interp_stdout` call sites under `tests/scheme/compile/` pass an errfile and call `show_interp_stderr` on their interpreter-failure branches (the PR text says 19; `git grep` finds 18 — cosmetic).
- End-to-end: pointing `native-macro-set-tracking-2212.sh` at a stub that prints a panic trace and exits 134 puts `--- interpreter stderr ---` plus the trace under every FAIL line.
- All 16 touched compile scripts pass against a fresh ReleaseSafe build of the branch with an isolated `KAAPPI_HOME`; `bash -n` clean on every touched script.
- The shown text does reach the CI log: both `run_shell_worker` (run-all.sh) and the Windows shell-suite loop capture scripts with `2>&1` and print the capture on FAIL.
- The new smoke script will run on both Windows legs under Git Bash. `errors/exit-code.sh` already execs a `chmod +x` stub there without a skip, so that shape has precedent.

The tier-parity rationale is preserved as claimed: nothing asserts on the captured text.

baijum:
Pushed `d9d81f90` — CI root cause fixed and all five review comments addressed.

**CI failures (all 8 legs):** the new smoke self-test shipped mode `644`; `run-all.sh` classifies a non-executable script as `NOTEXEC` and fails the leg. It never fired locally because my verification ran `bash <script>`, which never consults the bit — now `755` like every sibling `.sh`. The `netbsd-test` `run-differential.sh` DIFF (`#<fiber 32 running>` vs `errored`) is unrelated to this change: the differential harness never calls `interp_stdout`, and the diff touches no `.scm` execution path — it reads as a load-dependent fiber-state flake on the QEMU runner.

**@baijum's three nits** — all applied:
- *bundle-args comment:* reworded — the redirect is applied before the `cd`, so a relative errfile would land in the script's own cwd (repo root under `run-all.sh`), not the workdir.
- *shell-common.sh:* the comment above `interp_stdout` now documents the trade explicitly, including the "don't move the redirect back inside to fix the 3-arg form" warning.
- *self-test red path:* `grep -qs`, and re-verified against `origin/main`'s `shell-common.sh` — the red diagnostic is now exactly the one FAIL line. (My earlier stash-based red check had silently become a no-op once the fix was committed; this one restores main's file.)

**CodeRabbit's two comments:**
- *compile-toplevel-side-effects-2156.sh:* `show_interp_stderr` added to the `assert_tiers_agree` failure branch, with a comment naming the scenario it covers — an exit-time abort after stdout flushed passes the oracle's stdout check and surfaces only as the status mismatch. On the suggested extra regression case: skipped deliberately. Simulating an aborting interpreter inside this compile script means a stub binary and a bypassed real-binary path; the harness contract (capture + show + default-to-`/dev/null`) is what's testable deterministically, and the strengthened self-test covers all of it.
- *Self-test:* now asserts the `--- interpreter stderr ---` header in the shown output, and proves the 3-arg form still hides stderr using the **noisy** stub (capturing the caller's stderr and requiring it empty) — a quiet stub couldn't tell a redirect regression from silence. Verified the new check fires by sourcing a deliberately leaky `interp_stdout` override.

Re-verified locally: self-test red (main) / green (branch), leak-simulation fires, `compile-toplevel-side-effects-2156.sh` passes 12/12, `bash -n` clean on everything touched.

PR #2535, opened 2026-09-06
Default zig build lib to the portable baseline CPU

### LINKED ISSUE #2531: zig build lib tunes libkaappi_rt.a to the host CPU, so source-built kaappi compile binaries can SIGILL elsewhere
## What happens

#2529 pinned the *bundle* builds (`-Dbundle=`/`-Dbundle-src=`) to the portable baseline CPU, fixing #2515 for the standalone-binary path. The other documented way to ship a Kaappi program — `kaappi compile program.scm -o program` — links `libkaappi_rt.a`, and the `lib` step builds that archive with the same `target` the rest of the build resolves (build.zig, the `lib` step's `lib_mod ... .target = target`), which without `-Dtarget` is tuned to the build host's exact CPU model. A from-source

```bash
zig build           # populates zig-out/lib with a host-tuned archive
kaappi compile app.scm -o app
```

therefore ships the whole VM and GC, tuned to the build host's exact CPU, inside every native executable a source-built kaappi emits.

## Scenario

Build kaappi from source on an AVX-512 x86_64 host, `kaappi compile app.scm -o app`, copy `app` to another x86_64 machine → SIGILL inside the runtime, presenting as a VM bug exactly as in #2515 — the same misdiagnosis cost (it reads as a VM/GC bug, not a build-flags bug).

## Exposure

Released `libkaappi_rt.a` archives are baseline (the release workflow passes `-Dtarget`), so like the bundle case this hits **source builds only** — the same population #2515 targeted. Spotted in review of #2529: the baseline pin there is gated on `bundling`, which the `lib` step does not set.

## Options raised in that review

- **Extend the baseline default to the `lib` step too**, or to the whole build, with `-Dcpu=native` as the explicit dev opt-in for intentional host tuning. The whole-build variant also covers the plain `zig-out/bin/kaappi`, which travels between machines in practice (the A/B protocol in `docs/dev/performance.md` is built around copying builds out of `zig-out/` and reusing them).
- Keep `zig build lib` host-tuned and document it — cheapest, but leaves the trap armed.

## Relation to #2515 / #2529

#2529 deliberately fixes the bundle path only; this issue tracks the runtime-archive path so that `Closes #2515` on #2529 is not read as closing the whole hazard class.

### LINKED ISSUE #2536: kaappi compile via zig cc tunes the emitted object to the host CPU, so native binaries can still SIGILL elsewhere after #2531
## What happens

`kaappi compile program.scm -o program` emits LLVM IR and hands it to the first C compiler it finds. `native_compiler.cc_search_order` puts `zig` first, and `tryLink` runs `zig cc -w -O2 program.ll …` with no `-mcpu`. **`zig cc` with no `-mcpu` compiles for the build host's exact CPU**, not the triple's baseline (unlike `clang`/`gcc`, whose default is the generic model). So the object built from the emitted IR — the program's own code — is host-tuned in every native executable a `kaappi compile` produces through the preferred route, on released binaries and source builds alike.

#2529 pinned the bundle path (#2515) and #2535 pins `libkaappi_rt.a` (#2531). This is the third shipped component, and the only one that is still host-tuned after both.

## Evidence

Zig 0.16.0, macOS/aarch64 M3 host, cold cache (`ZIG_GLOBAL_CACHE_DIR` pointed at an empty dir — a cache hit prints only the cached object path, no argv):

```bash
zig cc -v -O2 -c t.c -o t.o 2>&1 | grep -o -- '-target-feature [^ ]*' | grep -c '+'
```

| invocation | enabled `-target-feature`s |
|---|---|
| `zig cc -O2` (default) | 102 |
| `zig cc -O2 -mcpu=native` | 102, list identical to default |
| `zig cc -O2 -mcpu=baseline` | 92 |

Enabled by default but absent under `-mcpu=baseline` here: `+amvs +bf16 +bti +ecv +fgt +fpac +fuse-adrp-add +hcx +i8mm +v8.5a +v8.6a`. On an AVX-512 x86_64 host the same mechanism hands clang `+avx512f,+avx512bw,…`, which the `-O2` SLP/loop vectorizers and the `llvm.s{add,sub,mul}.with.overflow.i64` lowering the emitter relies on are free to use. (The `-target-cpu apple-m1` name `zig cc` prints alongside is a model-name mapping quirk; the feature list is what the backend acts on.)

## Scenario

Same as #2515/#2531 minus the fix that landed for each: build or install kaappi on an AVX-512 x86_64 host, `kaappi compile app.scm -o app`, copy `app` to another x86_64 machine → SIGILL, presenting as a VM bug. After #2535 the runtime inside `app` is baseline, but `app.o` is not.

## Exposure

- **Every** `kaappi compile` whose search finds `zig` first — the route the docs tell users to prefer ("Always use `zig cc`, not `clang`"). Released binaries included: the pin in #2529/#2531 lives in `build.zig`, but this compile happens on the user's machine.
- The `native` step in `build.zig` (`zig build native -Dnative-src=…`) spells the same `zig cc -w -O2` with no `-mcpu`.
- `clang`/`gcc` routes (FreeBSD's system cc, NetBSD) are unaffected: their default is the generic model for the triple.

## Fix

Append `-mcpu=baseline` on the `is_zig` branch of `native_compiler.tryLink` (no flag on the clang/gcc branches — `-mcpu=baseline` is a Zig spelling) and in the `native` step's `cc_run`. If host tuning is worth keeping reachable, the same shape as #2529: an opt-in (`kaappi compile --cpu native`, or honour an explicit `-mcpu` from a future `--cc-flags`) rather than a default.

Regression test: `tests/scheme/compile/*.sh` twin of `bundle-cpu-baseline-2515.sh` / `lib-cpu-baseline-2531.sh` — differential on the linked object (`--emit-llvm`, link the `.ll` with the default invocation and with an explicit `-mcpu=baseline`, compare), gated on the same host≠baseline probe, plus an `--emit-llvm` IR assertion is not needed here since the IR itself carries no CPU (`target triple` only; the tuning enters at the cc step).

Found while reviewing #2535 (the review comment there has the same numbers).

`kaappi --version`: built from `main` at 05e6a544 (v0.26.1 + 5).

baijum:
Fixes #2531
Fixes #2536
Supersedes the runtime-archive half that #2529 deliberately left open when closing #2515.

## The bug

`zig build` with no `-Dtarget` tunes for the build host's exact CPU model, and the `lib` step built `libkaappi_rt.a` from that same host-tuned target. That archive is linked into **every executable `kaappi compile` emits** — the other documented way to ship a Kaappi program — so building kaappi from source on, say, an AVX-512 x86_64 host and running `kaappi compile app.scm -o app` shipped a host-tuned VM/GC inside `app`, which SIGILLs on another x86_64 machine and presents as a VM bug. Released archives were never affected (the release workflow passes an explicit `-Dtarget`, which already resolved baseline); only source builds — the same population #2515 targeted.

## The fix, part 1: the archive

The `lib` module now resolves its own target with the **same pin #2529 applies to bundles** (`.determined_by_arch_os → .baseline`), but unconditional rather than gated on `bundling`:

- Any explicit `-Dcpu=<model>` is respected verbatim; `-Dcpu=native` is the opt-out that restores the pre-fix tuning bit for bit.
- An explicit `-Dtarget` already meant baseline in Zig, so the release workflow's archives are **bit-identical to before**.
- Under a bundling configure the effective query is already pinned, so the lib and main targets are the same resolved target — one cache key, no forking.
- The dev side of the same configure is untouched: plain `zig build`'s REPL binary, the unit-test modules, and the benchmarks stay **host-tuned by default** (an explicit `-Dcpu` is respected everywhere); only the artifacts that ship are pinned.

## The fix, part 2: the program code (#2536)

The archive is only half of what `kaappi compile` ships. `zig` is first in the C-compiler search order, and `zig cc` with no `-mcpu` ends up on the **host** CPU (its effective `-target-cpu` is the detected model, byte-for-byte the `-mcpu=native` list — measured: default and `-mcpu=native` both resolve to `apple-m3`, `-mcpu=baseline` to `apple-m1`), while clang/gcc default to the generic model for the triple. So the SIGILL hazard survived on the preferred route even with a baseline archive.

`tryLink`'s `is_zig` branch now passes `-mcpu=baseline` (matching the archive's default model), and `zig build native` passes its own resolved CPU model, so `-Dcpu=native` there host-tunes the whole output consistently. No new opt-out flag: the manual three-step flow remains the escape hatch for hand-picked codegen (its documented example now includes `-mcpu=baseline`).

## Regression test

`tests/scheme/compile/lib-cpu-baseline-2531.sh` — the archive twin of `bundle-cpu-baseline-2515.sh`:

1. The default `zig build lib` archive must be byte-identical to an explicit `-Dcpu=baseline` one (fails without the fix wherever host ≠ baseline).
2. Where the CPU-model probe says host ≠ baseline, the `-Dcpu=native` archive must differ from the default.
3. A program compiled and linked against the default archive via `KAAPPI_LIB_DIR` runs **and exits 0** (a teardown crash after the expected output is exactly what this catches).
4. A logging `zig` shim first on PATH captures the exact argv the link step hands the compiler; the captured `zig cc` line must contain `-mcpu=baseline` — meaningful on **every** host, baseline or not, and verified to fail against pre-fix `native_compiler.zig`.

Verified on this M3 host (host=`apple_m3`, baseline=`apple_m1`): all four assertions pass with the fix and fail without it; `zig build test` and `bundle-cpu-baseline-2515.sh` (its exit-status check fixed too) pass; `zig build native -Dnative-src=` exercised end-to-end.

Docs: `AGENTS.md`/`CLAUDE.md`, `README.md`, `docs/dev/llvm-backend.md`, `docs/dev/test-runner.md` (the CPU one-key cache discipline now covers `zig build lib`, with the sanctioned `--prefix`ed differential-probe exception). The end-user native-compilation guide in kaappi/kaappi.github.io needs a matching follow-up, as #2529's did.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ symbols, and **multi-line input** with automatic paren balancing.
 ### Beyond the standard
 
 - **180 SRFIs** — 12 built-in, 164 as portable `.sld` libraries, plus SRFI 261 portable library references (`(srfi srfi-1)`, `(srfi lists-1)`) resolved in the importer and SRFI 226/160/211 as sub-libraries only (full list in [CONFORMANCE.md](CONFORMANCE.md))
-- **Native binaries** — `kaappi compile program.scm -o program` compiles Scheme to a native executable via LLVM, with self-tail-calls compiled as loops, linking a runtime archive tuned to the portable baseline CPU so the result runs on other machines of the same architecture ([details](docs/dev/llvm-backend.md))
+- **Native binaries** — `kaappi compile program.scm -o program` compiles Scheme to a native executable via LLVM, with self-tail-calls compiled as loops, tuned to the portable baseline CPU so the result runs on other machines of the same architecture ([details](docs/dev/llvm-backend.md))
 - **Standalone bundles** — `zig build -Dbundle-src=program.scm` embeds bytecode + libraries in a single executable, tuned to the portable baseline CPU so it runs on other machines of the same architecture (`-Dcpu=native` restores host tuning)
 - **C FFI** — call shared libraries from Scheme via `(kaappi ffi)`; 18 marshalled types, callbacks for passing Scheme procedures to C
 - **Concurrency** — green threads with channels via `(kaappi fibers)`, plus real OS threads via SRFI-18
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ symbols, and **multi-line input** with automatic paren balancing.
 ### Beyond the standard
 
 - **180 SRFIs** — 12 built-in, 164 as portable `.sld` libraries, plus SRFI 261 portable library references (`(srfi srfi-1)`, `(srfi lists-1)`) resolved in the importer and SRFI 226/160/211 as sub-libraries only (full list in [CONFORMANCE.md](CONFORMANCE.md))
-- **Native binaries** — `kaappi compile program.scm -o program` compiles Scheme to a native executable via LLVM, with self-tail-calls compiled as loops ([details](docs/dev/llvm-backend.md))
+- **Native binaries** — `kaappi compile program.scm -o program` compiles Scheme to a native executable via LLVM, with self-tail-calls compiled as loops, linking a runtime archive tuned to the portable baseline CPU so the result runs on other machines of the same architecture ([details](docs/dev/llvm-backend.md))
 - **Standalone bundles** — `zig build -Dbundle-src=program.scm` embeds bytecode + libraries in a single executable, tuned to the portable baseline CPU so it runs on other machines of the same architecture (`-Dcpu=native` restores host tuning)
 - **C FFI** — call shared libraries from Scheme via `(kaappi ffi)`; 18 marshalled types, callbacks for passing Scheme procedures to C
 - **Concurrency** — green threads with channels via `(kaappi fibers)`, plus real OS threads via SRFI-18
--- a/docs/dev/llvm-backend.md
+++ b/docs/dev/llvm-backend.md
@@ -80,10 +80,14 @@ NaN-boxed representation crosses C ABI trivially).
 ```bash
 zig build lib                                        # build libkaappi_rt.a
 kaappi --emit-llvm -o program.ll program.scm         # emit LLVM IR
-zig cc -w -O2 program.ll -o program \
+zig cc -w -O2 -mcpu=baseline program.ll -o program \
     -Lzig-out/lib -lkaappi_rt -lc -lm -lpthread      # link native binary
 ./program                                            # run
 ```
+
+(`zig cc` with no `-mcpu` tunes for the build host — see the baseline-CPU
+note above. This manual flow is also the escape hatch for hand-picking
+codegen the pinned routes don't offer.)
 
 ### Single-step build
 
@@ -96,19 +100,28 @@ zig build native -Dnative-src=program.scm            # all-in-one
 library references `__zig_probe_stack` and other Zig compiler-rt intrinsics
 that `clang` cannot resolve.
 
-**The archive defaults to the portable baseline CPU** (kaappi#2531):
-`libkaappi_rt.a` is linked into every binary `kaappi compile` emits, so it is
-a thing you ship, and without `-Dtarget` a host-tuned archive put a
-host-tuned VM/GC inside binaries that could SIGILL on another machine of the
-same architecture. `zig build lib` therefore resolves the baseline CPU model
-unless you name one — `-Dcpu=native` restores host tuning for archives that
-really will only be linked on the machine that built them. Released archives
-were never affected (the release workflow passes an explicit `-Dtarget`,
-which already meant baseline); only source builds. The dev binary from the
-same `zig build` configure stays host-tuned (kaappi#2529); only the archive
-is pinned. Note the emitted IR's own tuning comes from whichever C compiler
-links it — on the user's machine, with that compiler's default CPU — which
-is outside the build's control either way.
+**Everything `kaappi compile` ships defaults to the portable baseline CPU**
+(kaappi#2531). The runtime archive is one half: `libkaappi_rt.a` is linked
+into every binary the command emits, so it is a thing you ship, and without
+`-Dtarget` a host-tuned archive put a host-tuned VM/GC inside binaries that
+could SIGILL on another machine of the same architecture. `zig build lib`
+therefore resolves the baseline CPU model unless you name one —
+`-Dcpu=native` restores host tuning for artifacts that really will only run
+on the machine that built them. Released archives were never affected (the
+release workflow passes an explicit `-Dtarget`, which already meant
+baseline); only source builds.
+
+The program code is the other half, and the C compiler decides its default:
+`zig cc` with no `-mcpu` ends up on the *host* CPU (its effective
+`-target-cpu` is the detected model, byte-for-byte the `-mcpu=native` list),
+while clang and gcc default to the generic model for the target triple.
+Since `zig` is first in `kaappi compile`'s C-compiler search order, the
+`zig cc` route pins the emitted IR with `-mcpu=baseline`, matching the
+model of the archive it links against; `zig build native` does the same
+with its own resolved CPU model, so `-Dcpu=native` there host-tunes the
+whole output consistently. The dev binary built by the same `zig build`
+configure stays host-tuned by default (kaappi#2529) — an explicit
+`-Dcpu=<model>` is respected everywhere.
 
 ### Where `kaappi compile` looks for `libkaappi_rt.a`
 
--- a/docs/dev/llvm-backend.md
+++ b/docs/dev/llvm-backend.md
@@ -96,6 +96,20 @@ zig build native -Dnative-src=program.scm            # all-in-one
 library references `__zig_probe_stack` and other Zig compiler-rt intrinsics
 that `clang` cannot resolve.
 
+**The archive defaults to the portable baseline CPU** (kaappi#2531):
+`libkaappi_rt.a` is linked into every binary `kaappi compile` emits, so it is
+a thing you ship, and without `-Dtarget` a host-tuned archive put a
+host-tuned VM/GC inside binaries that could SIGILL on another machine of the
+same architecture. `zig build lib` therefore resolves the baseline CPU model
+unless you name one — `-Dcpu=native` restores host tuning for archives that
+really will only be linked on the machine that built them. Released archives
+were never affected (the release workflow passes an explicit `-Dtarget`,
+which already meant baseline); only source builds. The dev binary from the
+same `zig build` configure stays host-tuned (kaappi#2529); only the archive
+is pinned. Note the emitted IR's own tuning comes from whichever C compiler
+links it — on the user's machine, with that compiler's default CPU — which
+is outside the build's control either way.
+
 ### Where `kaappi compile` looks for `libkaappi_rt.a`
 
 In order: the `KAAPPI_LIB_DIR` environment variable, `<exe_dir>/../lib/`,
--- a/docs/dev/test-runner.md
+++ b/docs/dev/test-runner.md
@@ -368,17 +368,19 @@ Three things make that safe and worthwhile (kaappi#1926):
   `-Dbundle=` build resolves the portable **baseline** CPU model by default
   (a bundled binary is shipped, and a host-tuned one SIGILLs on other
   machines of the same arch), while `fixture_interpreter`'s plain `zig build`
-  stays host-tuned. That asymmetry is deliberate and free — the embedded
-  bytecode already puts the bundler on its own cache key, and the CPU model
-  is not part of the `.sbc` compiler hash (see [cache.md](cache.md)) — but it
-  means no script may pass `-Dcpu=native` (or any host-naming
-  `-Dcpu=<model>`) to a `-Dbundle=` build: it would fork the shared cache key
-  *and* reintroduce the portability bug the default exists to prevent. Since
-  kaappi#2531 the same holds for `zig build lib`: the runtime archive — the
-  one `ensure_runtime_lib` builds up front and every compile script links
-  against — also defaults to baseline (it ships inside `kaappi compile`
-  output), so no script may pass `-Dcpu=native` to it either, for both the
-  same reasons.
+  stays host-tuned; since kaappi#2531 the same holds for `zig build lib` —
+  the runtime archive `ensure_runtime_lib` builds up front and every compile
+  script links against ships inside `kaappi compile` output. That asymmetry
+  is deliberate and free — the embedded bytecode already puts the bundler on
+  its own cache key, and the CPU model is not part of the `.sbc` compiler
+  hash (see [cache.md](cache.md)) — so no script may pass a host-naming
+  `-Dcpu` to the shared builders, or to any build that installs into
+  `zig-out/`: it would fork the shared cache key *and* reintroduce the
+  portability bug the defaults exist to prevent. The sanctioned exception is
+  a differential probe that builds into a `--prefix` of its own, as the two
+  CPU-baseline regression scripts do with `-Dcpu=native`
+  (`bundle-cpu-baseline-2515.sh`, `lib-cpu-baseline-2531.sh`) — deliberately,
+  as their one cold build each.
 
 Dispatch inside a suite is longest-first — scripts that shell out to a full
 `zig build -D…`, or to either shared builder, go first, found by grep rather
--- a/docs/dev/test-runner.md
+++ b/docs/dev/test-runner.md
@@ -373,7 +373,12 @@ Three things make that safe and worthwhile (kaappi#1926):
   is not part of the `.sbc` compiler hash (see [cache.md](cache.md)) — but it
   means no script may pass `-Dcpu=native` (or any host-naming
   `-Dcpu=<model>`) to a `-Dbundle=` build: it would fork the shared cache key
-  *and* reintroduce the portability bug the default exists to prevent.
+  *and* reintroduce the portability bug the default exists to prevent. Since
+  kaappi#2531 the same holds for `zig build lib`: the runtime archive — the
+  one `ensure_runtime_lib` builds up front and every compile script links
+  against — also defaults to baseline (it ships inside `kaappi compile`
+  output), so no script may pass `-Dcpu=native` to it either, for both the
+  same reasons.
 
 Dispatch inside a suite is longest-first — scripts that shell out to a full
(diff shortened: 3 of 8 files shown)


docs/dev/llvm-backend.md
```diff
@@ -96,6 +96,20 @@ zig build native -Dnative-src=program.scm            # all-in-one
...
+same architecture. `zig build lib` therefore resolves the baseline CPU model
+unless you name one — `-Dcpu=native` restores host tuning for archives that
+really will only be linked on the machine that built them. Released archives
+were never affected (the release workflow passes an explicit `-Dtarget`,
+which already meant baseline); only source builds. The dev binary from the
+same `zig build` configure stays host-tuned (kaappi#2529); only the archive
+is pinned. Note the emitted IR's own tuning comes from whichever C compiler
+links it — on the user's machine, with that compiler's default CPU — which
+is outside the build's control either way.
+
```
baijum, line 125: This sentence is what the finding in the review body is about: on the preferred `zig cc` route the tuning *is* in the build's control, and today it is host tuning. `zig cc` with no `-mcpu` resolves the native CPU (its default `-target-feature` list is byte-for-byte the `-mcpu=native` list and enables features `-mcpu=baseline` drops), while `native_compiler.tryLink` and the `cc_run` in `build.zig`'s `native` step both pass `zig cc -w -O2 <ll>` with no `-mcpu`. So the archive is now baseline but the object it links against is not, and the SIGILL scenario in #2531 remains reachable through `kaappi compile` on an AVX-512 host.

Either add `-mcpu=baseline` on the `is_zig` branch of `tryLink` (clang/gcc already default to the generic model, so nothing is needed there) plus the same flag in `cc_run`, or file the follow-up and reword this paragraph and the README bullet so they promise only what is pinned.

README.md
```diff
@@ -208,7 +208,7 @@ symbols, and **multi-line input** with automatic paren balancing.
 ### Beyond the standard
 
 - **180 SRFIs** — 12 built-in, 164 as portable `.sld` libraries, plus SRFI 261 portable library references (`(srfi srfi-1)`, `(srfi lists-1)`) resolved in the importer and SRFI 226/160/211 as sub-libraries only (full list in [CONFORMANCE.md](CONFORMANCE.md))
-- **Native binaries** — `kaappi compile program.scm -o program` compiles Scheme to a native executable via LLVM, with self-tail-calls compiled as loops ([details](docs/dev/llvm-backend.md))
+- **Native binaries** — `kaappi compile program.scm -o program` compiles Scheme to a native executable via LLVM, with self-tail-calls compiled as loops, linking a runtime archive tuned to the portable baseline CPU so the result runs on other machines of the same architecture ([details](docs/dev/llvm-backend.md))
```
baijum: Over-promises until the emitted object is also pinned (see the llvm-backend.md comment): with `zig cc` first in the search order, the runtime archive is baseline after this PR but `app.o` is still compiled for the host CPU. Suggest "linking a runtime archive tuned to the portable baseline CPU" and stop there, or land the `-mcpu=baseline` on the `zig cc` invocation and keep the sentence.

docs/dev/test-runner.md
```diff
@@ -373,7 +373,12 @@ Three things make that safe and worthwhile (kaappi#1926):
   is not part of the `.sbc` compiler hash (see [cache.md](cache.md)) — but it
   means no script may pass `-Dcpu=native` (or any host-naming
   `-Dcpu=<model>`) to a `-Dbundle=` build: it would fork the shared cache key
-  *and* reintroduce the portability bug the default exists to prevent.
+  *and* reintroduce the portability bug the default exists to prevent. Since
+  kaappi#2531 the same holds for `zig build lib`: the runtime archive — the
+  one `ensure_runtime_lib` builds up front and every compile script links
+  against — also defaults to baseline (it ships inside `kaappi compile`
+  output), so no script may pass `-Dcpu=native` to it either, for both the
+  same reasons.
```
baijum: As written, both regression scripts violate this rule: `bundle-cpu-baseline-2515.sh` passes `-Dcpu=native` to a `-Dbundle=` build and `lib-cpu-baseline-2531.sh` passes it to `zig build lib`, deliberately, as their one cold build into a `--prefix` of their own. The rule is really about builds that are meant to share the up-front archive or the shared bundler's cache key, so it would be clearer to say that: no script may pass a host-naming `-Dcpu` to the shared builders or to a build that installs into `zig-out/`; a `--prefix`ed differential probe like the two CPU-baseline scripts is the sanctioned exception and costs one cold build.

tests/scheme/compile/lib-cpu-baseline-2531.sh
```diff
@@ -0,0 +1,162 @@
...
+(display "2531: baseline archive links and runs") (newline)
+EOF
+COMPILE_LOG="$DIR/compile.log"
+if ! (cd "$DIR" && KAAPPI_LIB_DIR="$DIR/prefix-default/lib" \
+        "$INTERP" compile prog.scm -o prog) > "$COMPILE_LOG" 2>&1; then
+    echo "FAIL: kaappi compile against the default (baseline) archive failed" >&2
+    cat "$COMPILE_LOG" >&2
+    exit 1
+fi
+OUTPUT="$("$DIR/prog" 2>&1)" || true
```
baijum: +1 to CodeRabbit here: the program's exit status is free to check and a nonzero exit after the expected line printed (teardown crash in the baseline archive) is exactly what assertion 3 exists to catch. The 2515 twin has the same `|| true`; worth fixing there too while you are at it.

baijum (commented):
Reviewed the build.zig change, the regression script, and the docs; ran `tests/scheme/compile/lib-cpu-baseline-2531.sh` on the PR branch on this M3 host (host=`apple_m3`, baseline=`apple_m1`): PASS in 2m07s, both differential assertions discriminating.

The archive pin itself is right: `.determined_by_arch_os → .baseline` on `lib_query` only, so an explicit `-Dtarget` (release workflow, every cross CI leg that runs `zig build lib -Dtarget=…`) resolves to the same baseline it always did, a bundling configure ends up with `lib_target == target` (one cache key), and `-Dcpu=native`/`-Dcpu=<model>` pass through untouched. The dev binary, tests and benches keep `target`, so the A/B protocol is unaffected. `zig build` alone never depended on `lib_install`, so plain builds pay nothing extra.

**One finding worth acting on before merge** (inline on `docs/dev/llvm-backend.md`): the PR body and the new docs say the emitted IR's own CPU tuning is "outside the build's control either way". It is not on the route `kaappi compile` prefers. `native_compiler.cc_search_order` tries `zig` first, and `zig cc` with no `-mcpu` compiles for the *host* CPU, not the triple's baseline. Measured here (Zig 0.16.0, cold cache):

| `zig cc -O2 -c` | enabled `-target-feature`s |
|---|---|
| default | 102 |
| `-mcpu=native` | 102, identical list to default |
| `-mcpu=baseline` | 92 |

The ten features the default adds over baseline on this host are `+amvs +bf16 +bti +ecv +fgt +fpac +fuse-adrp-add +hcx +i8mm +v8.5a +v8.6a`; on an AVX-512 x86_64 host the same mechanism hands clang `+avx512*`, and `-O2` (SLP/loop vectorizer, `ssub/smul.with.overflow` lowering) will happily use them in the program's own code. So after this PR a source-built `kaappi compile app.scm -o app` on such a host ships a baseline VM/GC linked to a host-tuned `app.o`, and can still SIGILL on the second machine, presenting exactly as the issue describes. `clang`/`gcc` are fine (their default is the generic model for the triple, which is why the `-Dtarget` legs and the FreeBSD system-cc route never saw this); only the `zig cc` route is affected.

The lever is in kaappi's hands: `tryLink` already knows `is_zig`, so appending `-mcpu=baseline` on that branch (and in the `cc_run` of `build.zig`'s `native` step, which also spells `zig cc -O2` with no `-mcpu`) closes it, with the same `-Dcpu=native`-style opt-out if you want one. Fine to land as a follow-up issue instead, but then the new README sentence ("so the result runs on other machines of the same architecture") and the llvm-backend.md note should say the archive is pinned and the linked object is not yet.

Two smaller things inline (test-runner.md rule wording; agree with CodeRabbit's exit-status nit on the script). CodeRabbit's other request, qualifying "host-tuned" with "by default", is fair but cosmetic.

baijum:
Pushed 67b81f1d addressing every review thread. Point-by-point:

**The main finding (link-route host tuning) — fixed in this PR, not a follow-up.** Reproduced the measurement first: `zig cc -###` resolves default and `-mcpu=native` to `apple-m3`, `-mcpu=baseline` to `apple-m1`, confirming plain `zig cc` is host tuning in disguise. `tryLink`'s `is_zig` branch now passes `-mcpu=baseline`, and build.zig's `cc_run` passes `-mcpu=<resolved lib-target model>` — so `-Dcpu=native zig build native` host-tunes the whole output consistently rather than leaving a baseline object inside a host-tuned build. On the no-new-flag question: the manual three-step flow (llvm-backend.md, example updated to `-mcpu=baseline`) is the escape hatch for hand-picked codegen; clang/gcc routes need nothing since their triple default is already generic. The argv buffers went 16 → 20 slots because the new flag filled the worst case (zig cc + OpenBSD `-z nobtcfi`) to exactly capacity.

**Exit-status masking (lib-cpu-baseline-2531.sh assertion 3, and the 2515 twin)** — fixed in both: nonzero exit now fails the test with the captured output; the output-content assertion is unchanged.

**New assertion 4** to pin the actual link-route behavior on every host (baseline hosts included, where the differential assertions are vacuous): a logging `zig` shim first on PATH captures the exact argv `kaappi compile` hands the compiler, and the captured `zig cc` line must contain `-mcpu=baseline`. Verified FAIL against pre-fix `native_compiler.zig` (captured line shows no `-mcpu`) and PASS with it.

**CodeRabbit's wording fixes** — CLAUDE.md and llvm-backend.md now say the dev binary is host-tuned *by default* with explicit `-Dcpu` respected everywhere, and the llvm-backend.md note no longer claims the emitted IR's tuning is outside the build's control: it states the zig-cc route is pinned and documents the compiler-dependent defaults honestly. README's bullet drops the archive-only hedge and promises the whole result is baseline, which is now true.

**test-runner.md rule wording** — rewritten along the suggested lines: no script may pass a host-naming `-Dcpu` to the shared builders or any build installing into `zig-out/`; a `--prefix`ed differential probe like the two CPU-baseline scripts is the sanctioned exception and costs one cold build each.

All verification re-run green: `zig build test`, `lib-cpu-baseline-2531.sh` (4/4), `bundle-cpu-baseline-2515.sh`, and `zig build native -Dnative-src=` end-to-end.

baijum:
Filed the `zig cc` host-tuning gap from the review as #2536 (the linked `.ll` object is the third shipped component, still host-tuned after this PR). If this PR stays archive-only, the README bullet and the llvm-backend.md note should reference #2536 instead of calling the object's tuning out of the build's control.

tests/scheme/compile/lib-cpu-baseline-2531.sh
```diff
@@ -0,0 +1,199 @@
...
+fi
+if ! grep -q -- '-mcpu=baseline' "$SHIM_LOG"; then
+    echo "FAIL: the zig-cc link route did not pin -mcpu=baseline —" >&2
+    echo "      the program code kaappi compile ships stays host-tuned (kaappi#2531)" >&2
+    echo "argv captured by the shim:" >&2
+    cat "$SHIM_LOG" >&2
+    exit 1
+fi
+
+OUTPUT="$(cd "$DIR" && ./prog 2>&1)"
```
baijum: Under `set -euo pipefail` a nonzero `./prog` exit does fail the script here, but silently: bash exits with prog's status before the `FAIL:` branch below runs, so the runner sees a red script with no message and no `compile.log`. The 2515 twin got the explicit form in this commit; same here:

```suggestion
if ! OUTPUT="$(cd "$DIR" && ./prog 2>&1)"; then
    echo "FAIL: program linked against the default (baseline) archive exited unsuccessfully" >&2
    echo "full output: $OUTPUT" >&2
    cat "$COMPILE_LOG" >&2
    exit 1
fi
```

baijum (commented):
Reviewed 67b81f1d against the first commit. It addresses every thread, and the link-route pin is the right shape. Re-verified on this M3 host on the new head:

- `lib-cpu-baseline-2531.sh`: PASS, 4/4 assertions, 2m00s (host=`apple_m3`, baseline=`apple_m1`).
- `zig build native -Dnative-src=` on a scratch program: links with `-mcpu=apple_m1`, runs, exits 0.
- `zig cc` accepts the Zig-table model names `build.zig` now passes (`apple_m1`, `apple_m3`, `x86_64`, `baseline` all compile; the feature counts match the earlier measurement: 92 for baseline/apple_m1, 102 for apple_m3).
- `kaappi compile` locates the compiler with `findInPath`, so the PATH shim in assertion 4 observes the real argv rather than a lookalike.

The argv-buffer growth is correct: the OpenBSD zig route was at exactly 16 slots including the null (15 args + terminator), so the new flag would have overflowed it.

Two small things, neither blocking:

1. Inline: assertion 3 now fails on a nonzero exit, but only through `set -e`, so a teardown crash exits the script with no `FAIL:` line and no compile log. The 2515 twin got the `if ! OUTPUT=…` form in this same commit; mirror it here.
2. This commit fixes #2536 outright, so the PR body should carry `Fixes #2536` alongside `Fixes #2531` (one keyword per issue, or GitHub closes only the first). I filed #2536 minutes after this commit was pushed; my earlier comment suggesting a follow-up is superseded, and I have cross-linked from the issue.

Aside, not a request: `zig build native` passes only the resolved model *name*, so an explicit `-Dcpu=x86_64_v3+avx512f` drops the `+avx512f` on the program object while the archive keeps it. Nobody does that today and the manual flow covers it.

LGTM — good to merge once CI is green (GitHub will not accept an Approve event from the author's own account, hence a comment review).

baijum:
Both small things addressed in d9a4741b:

1. **Assertion 3's silent exit** — now uses the explicit `if ! OUTPUT=…` form exactly as suggested, mirroring the 2515 twin: a nonzero program exit reaches the FAIL branch with the captured output and the compile log instead of dying silently under `set -e`. Re-ran the script: PASS 4/4; the `if !` control flow was also sanity-checked directly (nonzero exit reaches the FAIL branch, captured output intact, script returns nonzero as intended).
2. **PR body** — now carries `Fixes #2536` alongside `Fixes #2531` (one keyword per issue), and the part-2 section is retitled to name the issue it closes.

On the aside (resolved model *name* vs feature string in `zig build native`): agreed, not worth surface today — the manual flow is the escape hatch, and following the full `-Dcpu` feature string into the link flag would be a new spelling to keep in sync with zig's `-mcpu` grammar for no user we can name. If that ever comes up, it's a one-line change in `cc_run`.

PR #2538, opened 2026-09-06
Free completed-unjoined thread resources at process exit

### LINKED ISSUE #2537: srfi18-join-spawn-grandchild-2129.scm times out on the Debug CI leg: live grandchild at exit skips teardown, 29k-entry leak report blows the 240s budget
## Summary

`tests/scheme/srfi/srfi18-join-spawn-grandchild-2129.scm` was killed after 240s on the `test (ubuntu-latest, Debug)` leg of the CI run for the v0.26.2 release commit (0b8e6947), which changed only `build.zig.zon` and `CHANGELOG.md`. The same leg passed at the identical source tree 25 minutes earlier (82930426, run 34009086771). Every one of the file's 6 assertions had passed; the time went into the exit path.

Failing run: https://github.com/kaappi/kaappi/actions/runs/34010135155 (job 101424420232). It made `ci-gate` refuse to publish the release until the leg was re-run.

## What the log shows

```
  TIMEOUT  tests/scheme/srfi/srfi18-join-spawn-grandchild-2129.scm  (killed after 240s)
%%%% Starting test srfi18-join-spawn-grandchild-2129  (Writing full log to "srfi18-join-spawn-grandchild-2129.log")
# of expected passes      6
error(DebugAllocator): memory address 0x7fae00180001 leaked:
/home/runner/work/kaappi/kaappi/src/memory.zig:157:37: 0x12c844d in dupeSliceNoFill__anon_44338 (main.zig)
    const buf = try allocSliceNoFill(allocator, T, data.len);
                                    ^
/home/runner/work/kaappi/kaappi/src/gc_alloc.zig:163:38: 0x12c86d9 in allocString (main.zig)
...
```

followed by **29,366 leak entries** (~617k log lines), each with a symbolized stack trace. By allocation site:

| entries | site |
|--------:|------|
| 18000 | `dupeSliceNoFill` via `allocString` |
| 11366 | `allocPair` |

18000 = 6 busy grandchildren × 3000 `(string …)` allocations — exactly the heap contents of the "6 grandchildren keep allocating past the join" variant. So at process exit at least one busy grandchild was still alive, the root skipped its teardown (the #1814 rule: never deinit while children are alive), the retained heaps were reported by the Debug allocator, and symbolizing 29k stack traces through DWARF blew the 240s budget. Release legs use `c_allocator` and print nothing, so only the Debug leg can fail this way.

## Why it is a race

The middle threads return without joining the busy grandchildren, so the six 3000-iteration allocation loops run concurrently with the rest of the file. Debug builds are ~500x slower on allocation-heavy work, so whether the last grandchild finishes before the main thread reaches `(exit …)` depends on scheduling. Nothing in the file waits for them, and the test's point is the unjoined shape, so it cannot simply join them.

## Suggested fix

Either of:

- Give the file a bounded wait at the end (poll a shared flag each busy grandchild sets via a global, or a `thread-sleep!` long enough on Debug) so the process never exits with a live grandchild, keeping the unjoined shape for the assertions themselves.
- Or make the Debug leg's exit cheap when children are live: skip the leak report (or cap it) for heaps that are retained deliberately because a child is alive, since those are not leaks in the sense the report exists to catch.

The behaviour under test is correct; only the CI signal is wrong.

## Environment

- kaappi 0.26.2 (0b8e6947), CI `ubuntu-latest`, `-Doptimize=Debug`, `KAAPPI_TEST_TIMEOUT=240`

baijum:
Fixes #2537.

## Root cause

The issue's symptom is right but the mechanism is one step different. The Debug leak report comes from `da.deinit()`, which only runs on the **full-teardown** path — when `hasLiveChildThreads()` is false at the exit defer. So the failing run is the case where all six busy grandchildren had **finished (but were never joined)** before the main thread reached the exit path; the passing run 25 minutes earlier is the case where they were still alive and the `_exit` branch skipped teardown entirely.

Finished-but-unjoined threads keep their `child_registry` entry by design — the result envelope must survive a future `thread-join!` — so their child GC/VM are retained to process exit, and on a Debug build `da.deinit()` reports every object in those heaps (29,366 entries for this file, ~617k log lines once each is DWARF-symbolized). Whether teardown runs at all is the race; when it runs, the report alone blows the 240s budget.

I confirmed the mechanism on a pre-fix Debug binary: a single finished-but-unjoined grandchild with the same 3000-allocation loop produces 9,037 leak entries; the actual test file, made deterministic, produces 54,617 entries in ~2 minutes on this machine.

## The fix

**Engine (`primitives_srfi18.zig` + `main.zig`)** — the exit path's no-live-children branch now sweeps the registry (`freeUnjoinedExitedChildResources`): every entry still in the map belongs to a thread that has passed `threadEntryFn`'s `markExited` defer (the live-thread count is held from before the spawn until after it), so its resources are freed exactly as a join would have — unconsumed result/exception envelopes released, GC/VM deinited. Live children keep the `_exit` branch untouched, and the `thread_exited` check stays as a floor: if a future ordering change ever broke the invariant, the failure mode degrades back to a leak report, never to freeing a running thread's GC/VM.

**Test (`srfi18-join-spawn-grandchild-2129.scm`)** — the bounded drain the issue suggests, in the shape that actually helps: each busy grandchild marks its own slot (a root-heap box, `set-car!` with an immediate — the file's existing `gg-done` pattern) when its allocation loop finishes, and the main thread polls until all six are set before `test-end`, bounded at 60s so a genuinely stuck grandchild proceeds to the cheap `_exit` exit instead of hanging the leg. The unjoined shape stays; the process just no longer exits at a race-dependent moment relative to the grandchildren, which makes the completion path *and* the new sweep deterministic on every leg. (Draining alone, without the sweep, makes the Debug leg deterministically slow — verified — which is why the engine half is required.)

**Unit test (`tests_srfi18.zig`)** — a completed-unjoined thread's entry is freed by one sweep pass (second pass is a no-op), with the testing allocator's leak check as the second detector if any entry is ever skipped.

## Verification

| | pre-fix Debug | post-fix Debug | post-fix ReleaseSafe |
|---|---|---|---|
| test file, drained | 54,617 leak entries, 2:04 | 0 entries, **15s** | 0 entries, **0.9s** |
| assertions | 6/6 | 6/6 | 6/6 |

- `zig build test` — pass
- `bash tests/scheme/run-all.sh` — 756 scheme files + 1,395 R7RS tests, 0 fail
- `zig fmt --check` — clean

### DIFF
--- a/src/primitives_srfi18.zig
+++ b/src/primitives_srfi18.zig
@@ -125,9 +125,11 @@ const ChildThreadResources = struct {
 // Entries are freed when a thread is joined (freeChildResources called from
 // reapOsThread) or, when the join retires the entry because the thread still
 // has live descendants, by the last descendant's threadEntryFn defer once the
-// subtree drains (#2129). Threads that complete but are never joined leak
-// their child VM and GC — the result must survive until the parent copies it
-// out of its envelope, and automatic cleanup would race with that copy.
+// subtree drains (#2129). Threads that complete but are never joined keep
+// their child VM and GC until process exit -- the result must survive until
+// the parent copies it out of its envelope, and automatic cleanup would race
+// with that copy -- where the exit sweep frees them (kaappi#2537, called
+// from main.zig once no child thread is live).
 const ChildRegistry = struct {
     map: std.AutoHashMap(usize, ChildThreadResources),
     mutex: std.atomic.Mutex,
@@ -208,6 +210,22 @@ const ChildRegistry = struct {
             if (!entry.retired) return null;
         }
         if (self.map.fetchRemove(key)) |kv| return kv.value;
+        return null;
+    }
+
+    // kaappi#2537: pops one exited thread's entry, or null when none remains.
+    // The exit sweep (freeUnjoinedExitedChildResources) pops one at a time so
+    // freeChildResourcesEntry's real work -- two deinits plus frees -- runs
+    // outside the registry lock, same shape as the descendant-drain path.
+    fn removeAnyExited(self: *ChildRegistry) ?ChildThreadResources {
+        memory.spinLock(&self.mutex);
+        defer memory.spinUnlock(&self.mutex);
+        var it = self.map.iterator();
+        while (it.next()) |kv| {
+            if (kv.value_ptr.thread_exited) {
+                return self.map.fetchRemove(kv.key_ptr.*).?.value;
+            }
+        }
         return null;
     }
 };
@@ -1730,6 +1748,50 @@ fn freeChildResourcesEntry(res: ChildThreadResources, heir: ?*memory.GC) void {
     }
     res.child_gc.deinit();
     allocator.destroy(res.child_gc);
+}
+
+/// kaappi#2537: process-exit sweep of threads that completed but were never
+/// joined. The registry deliberately keeps such an entry alive -- its result
+/// envelope must survive until a future thread-join! copies it out -- but at
+/// exit there is no future join, and on a Debug build the retained child
+/// GC/VM surfaced as 29,366 DebugAllocator leak entries at da.deinit() for
+/// srfi18-join-spawn-grandchild-2129.scm, each symbolized through DWARF,
+/// which alone blew the Debug CI leg's whole 240s budget. Freeing here is
+/// what the exit path should have done all along: the same teardown a join
+/// performs, once it is certain no join will ever come.
+///
+/// Must only be called when no child thread is live -- main.zig's exit defer
+/// calls it in its hasLiveChildThreads() == false branch. That branch is what
+/// makes every remaining entry safe to free: a thread between its
+/// child_registry.put and threadEntryFn's markExited defer holds the live
+/// count throughout, so at count zero every entry in the map has exited. The
+/// thread_exited check stays as a floor anyway -- if a future ordering change
+/// ever breaks that invariant, skipping the entry degrades back to a leak
+/// report, never to freeing a running thread's GC/VM out from under it.
+///
+/// Returns how many entries were freed (unit-test seam).
+pub fn freeUnjoinedExitedChildResources() usize {
+    var freed: usize = 0;
+    while (child_registry.removeAnyExited()) |res| {
+        // A completed-unjoined thread's result/exception envelopes were never
+        // consumed by a join. They point into the child heap being freed, so
+        // there is nothing to copy out -- release them exactly as reapOsThread
+        // would have, then drop the entry.
+        switch (res.result) {
+            .envelope => |env| env.deinit(),
+            .none, .failed => {},
+        }
+        switch (res.exception) {
+            .envelope => |env| env.deinit(),
+            .none, .failed => {},
+        }
+        // Null heir, like the descendant-drain free: the process is
+        // quiescent, so there is no concurrent collection to protect and no
+        // parent that could still dereference this heap.
+        freeChildResourcesEntry(res, null);
+        freed += 1;
+    }
+    return freed;
 }
 
 fn threadJoinResult(target: *fiber_mod.Fiber) PrimitiveError!Value {
--- a/src/primitives_srfi18.zig
+++ b/src/primitives_srfi18.zig
@@ -217,6 +217,9 @@ const ChildRegistry = struct {
     // The exit sweep (freeUnjoinedExitedChildResources) pops one at a time so
     // freeChildResourcesEntry's real work -- two deinits plus frees -- runs
     // outside the registry lock, same shape as the descendant-drain path.
+    // Each pop rescans from the map's start, making a full sweep quadratic in
+    // the number of unjoined threads -- a non-issue at realistic counts, and
+    // the price of keeping the free outside the lock.
     fn removeAnyExited(self: *ChildRegistry) ?ChildThreadResources {
         memory.spinLock(&self.mutex);
         defer memory.spinUnlock(&self.mutex);
--- a/src/tests_srfi18.zig
+++ b/src/tests_srfi18.zig
@@ -1128,6 +1128,37 @@ test "kaappi#2473: thread-start! unwinds extra_roots when the spawn setup fails"
     try std.testing.expectEqual(before, gc.extra_roots.items.len);
 }
 
+// kaappi#2537: a thread that completes but is never joined keeps its
+// child_registry entry -- its result envelope must survive a future join --
+// so at process exit a Debug build's leak-tracking allocator reported every
+// object in its GC/VM as a leak: 29,366 entries for
+// srfi18-join-spawn-grandchild-2129.scm, each symbolized through DWARF,
+// which alone blew the Debug CI leg's whole 240s budget. The exit sweep must
+// free exactly those entries. The testing allocator is the second detector:
+// the child GC/VM are allocated from the root GC's allocator, so an entry the
+// sweep skips fails ctx.deinit's leak check even if a stale entry from
+// another test masked the count.
+test "kaappi#2537: exit sweep frees a completed-unjoined thread's registry entry" {
+    if (comptime platform.is_wasm) return error.SkipZigTest; // thread-start! is unregistered on wasm
+    var ctx: th.TestContext = undefined;
+    try ctx.init();
+    defer ctx.deinit();
+    _ = try ctx.vm.eval("(define sweep-t (make-thread (lambda () 'swept)))");
+    _ = try ctx.vm.eval("(thread-start! sweep-t)");
+    // The sweep may only run once no child is live (main.zig calls it in the
+    // hasLiveChildThreads() == false branch). The live count is held from
+    // before the spawn until after the child's markExited defer, so false
+    // here means the entry exists and its thread has fully exited.
+    var spins: u32 = 0;
+    while (srfi18.hasLiveChildThreads()) : (spins += 1) {
+        if (spins >= 10_000) return error.ChildThreadNeverExited;
+        platform.sleepNs(std.time.ns_per_ms);
+    }
+    try std.testing.expect(srfi18.freeUnjoinedExitedChildResources() >= 1);
+    // Everything exited is gone after one pass; nothing can be freed twice.
+    try std.testing.expectEqual(@as(usize, 0), srfi18.freeUnjoinedExitedChildResources());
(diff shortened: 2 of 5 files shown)


tests/scheme/srfi/srfi18-join-spawn-grandchild-2129.scm
```diff
@@ -130,6 +148,18 @@
...
+;; every grandchild's allocation loop has finished, so the process never
+;; exits while one is mid-loop. The bound exists only so a genuinely stuck
+;; grandchild fails by proceeding (the exit path handles live children
+;; cheaply) instead of hanging the leg; 60s is far beyond what the loops
+;; need even on a Debug build, and in practice the last flag is already set
+;; by the time we get here.
+(let drain ((polls 0))
+  (when (and (not (busy-done-all?)) (< polls 1200))
+    (thread-sleep! 0.05)
+    (drain (+ polls 1))))
```
baijum, line 166: I'd add the assertion CodeRabbit suggested here, and I think it is strictly better than the current shape rather than a trade-off:

```scheme
(test-assert "all six busy grandchildren finished their allocation loops"
  (busy-done-all?))
```

The stated reason for not asserting is that a stuck grandchild should proceed to the cheap `_exit` branch instead of hanging the leg. The assertion keeps that property: a failing `test-assert` bumps the fail count, the `(exit 1)` below fires, and a still-live grandchild still takes the `_exit` branch on the way out. What changes is that a grandchild that never finishes turns from a silent 60-second pass into a loud failure, which is the kind of regression this file exists to catch.

One accuracy nit on the comment block above (line 77, "deterministic on every run"): the flag is set *before* the grandchild's epilogue (result envelope build, descendant drain, `markExited`, the live-count decrement), so a small window remains where the main thread reaches the exit defer with `hasLiveChildThreads()` still true and takes the `_exit` branch. That is harmless now, since both branches are cheap, but "either exit branch is now cheap" describes the guarantee more precisely than "deterministic".

src/tests_srfi18.zig
```diff
@@ -1128,6 +1128,37 @@ test "kaappi#2473: thread-start! unwinds extra_roots when the spawn setup fails"
...
+    // The sweep may only run once no child is live (main.zig calls it in the
+    // hasLiveChildThreads() == false branch). The live count is held from
+    // before the spawn until after the child's markExited defer, so false
+    // here means the entry exists and its thread has fully exited.
+    var spins: u32 = 0;
+    while (srfi18.hasLiveChildThreads()) : (spins += 1) {
+        if (spins >= 10_000) return error.ChildThreadNeverExited;
+        platform.sleepNs(std.time.ns_per_ms);
+    }
+    try std.testing.expect(srfi18.freeUnjoinedExitedChildResources() >= 1);
```
baijum: Two related nits on the `>= 1` and the "stale entry from another test masked the count" clause in the comment above.

A stale exited-unjoined entry from an earlier test cannot reach this test without that earlier test having already failed: the child GC/VM are allocated from `std.testing.allocator`, whose instance is leak-checked and reset by the test runner after every test. And if one somehow did survive, this sweep would free it through the *reset* `DebugAllocator` instance, which reports an invalid free rather than cleaning up, so the sweep is not a safe fallback for that case either.

Given that, `expectEqual(@as(usize, 1), ...)` is both stricter and easier to reason about than `>= 1`: exactly one entry exists, exactly one is freed. If a future test starts leaving entries behind, an equality here surfaces it instead of absorbing it. I'd also trim the "masked the count" sentence to match.

src/primitives_srfi18.zig
```diff
@@ -1732,6 +1750,50 @@ fn freeChildResourcesEntry(res: ChildThreadResources, heir: ?*memory.GC) void {
...
+/// calls it in its hasLiveChildThreads() == false branch. That branch is what
+/// makes every remaining entry safe to free: a thread between its
+/// child_registry.put and threadEntryFn's markExited defer holds the live
+/// count throughout, so at count zero every entry in the map has exited. The
+/// thread_exited check stays as a floor anyway -- if a future ordering change
+/// ever breaks that invariant, skipping the entry degrades back to a leak
+/// report, never to freeing a running thread's GC/VM out from under it.
+///
+/// Returns how many entries were freed (unit-test seam).
+pub fn freeUnjoinedExitedChildResources() usize {
```
baijum, line 1776: Verified the "at count zero every entry has exited" argument this relies on, walking the branch: increment before `spawn` in `threadStartFn`, `put` after that on the child, `markExited` declared after the decrement defer in `threadEntryFn` so it fires first, and the descendant-drain defer plus `envelope.deinit()` declared later still so they finish before `markExited`. After the decrement the thread only touches the `c_allocator` exit flag. Free order among entries is also irrelevant: a child `GC.deinit` walks its own object lists only, and its `foreign_symbols` is empty because a child records internings in the root's list. Null heir is right here for the same reason it is in the descendant-drain path.

Optional: `removeAnyExited` rescans from the start of the map for every pop, so the sweep is quadratic in the number of unjoined threads. That is a non-issue at realistic counts and keeps the free outside the lock, so I would not change it; just noting it was considered.

baijum (commented):
Reviewed the sweep against the thread lifecycle on the branch; the safety argument holds and the fix is the right shape. Three small things inline (one test assertion, one test-comment/assertion tightening, one verification note), none blocking.

**Invariant check.** `live_child_threads` is incremented in `threadStartFn` *before* `std.Thread.spawn`, the child's `child_registry.put` runs after that on the child thread, and in `threadEntryFn` the `markExited` defer is declared after the decrement defer so it fires first. The descendant-drain defer and `envelope.deinit()` are declared later still, so they also complete before `markExited`. At count zero the only thing an entry's thread can still be doing is `releaseOsThreadExit`, which touches only the `c_allocator` exit flag. Free order across entries doesn't matter either: a child `GC.deinit` walks only its own object lists, and its `foreign_symbols` is empty (a child's internings are recorded in the root's list, which the root's own deinit frees after the sweep).

One note for the record: `(exit n)` goes through `std.process.exit` and skips `mainImpl`'s defers entirely, so the sweep only runs on the fall-off-the-end path. That is the path the failing run took, so it is the right place, but it means an explicit `(exit 0)` at the end of a script still never reaches `da.deinit()`. Fine as is, just worth knowing when reading the "what the exit path should have done all along" framing.

The `test (ubuntu-latest, Debug)` leg is still pending as I write this. It is the one leg the PR exists to fix, so I'd wait for it before merging.

tests/scheme/srfi/srfi18-join-spawn-grandchild-2129.scm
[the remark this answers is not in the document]
baijum: Done in 8165d385, using the file's box pattern for the new assertion: `all six busy grandchildren finished their allocation loops`. An expiry now fails loudly via the fail count → `(exit 1)`, and a grandchild still live at that point still exits through the cheap `_exit` branch, so the cheap-exit property the original comment leaned on is preserved.

baijum: Agreed on all points — adopted in 8165d385: your assertion wording, and the comment now says the guarantee is "either exit branch is cheap", with the flag-before-epilogue window spelled out.

src/tests_srfi18.zig
baijum: Switched to `expectEqual(@as(usize, 1), ...)` in 8165d385 and trimmed the masked-count clause — you're right that the per-test leak-checked allocator closes that door from both sides, and equality is the better tripwire for a future leaking test.

src/primitives_srfi18.zig
baijum: Thanks for walking the invariant — that's exactly the argument the doc comment leans on. Left the rescan as is per your suggestion, and recorded the consideration in removeAnyExited's comment (quadratic in unjoined-thread count, the price of keeping the free outside the lock) so the next reader doesn't re-derive it. Also took the `(exit n)` note from your review body: correct that the sweep only covers the fall-off-the-end path, which is the path the failing run took; no code change, since an explicit `(exit n)` bypasses `da.deinit()` too and so has no report to blow the budget.

baijum (commented):
Reviewed 8165d385. All three follow-ups land as asked, and the commit message says why for each.

- **Drain assertion.** `test-assert` after the bounded wait, with the surrounding comment now stating the actual guarantee (either exit branch is cheap; the `_exit` window before the grandchild's epilogue still exists). A stuck grandchild is now a loud failure that still leaves through `_exit`.
- **Exact count in the sweep unit test.** I checked the "every other thread test joins" premise rather than take it on faith. The other `tests_srfi18.zig` tests that matched `thread-start!` only match it in their wasm-skip comment; they use fibers via `spawn`/`fiber-join`, not OS threads. Across the other test files that do spawn OS threads, every GC is built on `std.testing.allocator`, so any test leaving an exited-unjoined entry would already fail its own per-test leak check. The `== 1` is safe in declaration order.
- **`removeAnyExited` comment.** Records the quadratic rescan and why it is kept. Fine.

CI for this commit is still running as I post this. The change is comment-and-test only on top of a tree whose Debug leg already went green (that run's log had zero leak-report lines, so the sweep did the work rather than the race landing on `_exit`). No further changes requested from my side; merge once the legs finish.

baijum:
CI update on the two failures from the review commit's run:

- **riscv64 rest chunk** — passed on re-run. The step runs none of this PR's code (`tests_srfi18.zig` is in the concurrency chunk, which passed on the same run, 232/235); it was QEMU load variance against the 24m step cap (measured healthy: 6.9m).
- **Debug leg** — the #2487 oracle abort did **not** recur on re-run (evidence from the first occurrence is on #2532), but `bundle-cpu-baseline-2515.sh` was killed at the 600s shell cap **again**, making it twice in a row. The identical script passed inside the same budget on this PR's first commit the same day, and the diff between heads is comments + test assertions — runner-load variance around a legitimate build. The test builds three bundles (the #2515 fixture plus two deliberately-distinct-cache-key variants), so its all-cold Debug total sits at ~600s.

Followed the #2232 precedent one budget higher in cdfed09a: `KAAPPI_SHELL_TEST_TIMEOUT` becomes Debug-scoped 900s (a hang still dies 3x inside the 40m job cap), ReleaseSafe legs stay at 600. Fresh CI is running on that head.

Issue #2448, opened 2026-09-01
SRFI 231: unsafe arrays silently alias out-of-domain indices; bulk ops pay checks they don't need


baijum:
**Kaappi version**

Reproduced on both, from separate builds:

- released **v0.25.0** (tag `767d54ec`)
- **HEAD** `8b0f3504` (v0.25.0-33-g8b0f3504)

Output is byte-identical between the two. The two SRFI 231 commits in that
range (`a6cf52db` f16-storage-class, `1503d1d4` c64/c128 bodies) touch only
`storage-classes.sld`; `arrays.sld` and `views.sld`, which hold the
`safe?` branch below, are unchanged since the tag.

**Platform**

macOS aarch64

**Build mode**

ReleaseSafe (default)

**Description**

Reported upstream by Brad Lucier (`gambiteer`), the SRFI 231 author, in
[r/KaappiScheme: "SRFI 231: Faster *and* safer"](https://www.reddit.com/r/KaappiScheme/comments/1w3r8qh/srfi_231_faster_and_safer/).

His argument: the `safe`/`unsafe` array distinction was a mistake. The spec
only says that `safe: #t` *adds* argument checking to the getter and setter —
it never says that with `safe: #f` an out-of-bounds multi-index or an invalid
setter value may take the system down. His sample implementation has dropped
the distinction entirely. Instead it keeps `unsafe-getter`/`unsafe-setter` as
*internal* array fields, always wraps the *user-visible* getter and setter in
validity checks, and uses the raw pair only inside bulk operations, where
indices are known valid by construction. The result is both faster and safer.

Our implementation is structured the opposite way on both counts, and the
safety consequence here is worse than the one Brad anticipated.

**1. Out-of-domain access silently returns/clobbers the wrong element**

`lib/srfi/231/arrays.sld` hands the user the *raw* accessor whenever `safe?`
is false:

```scheme
;; lib/srfi/231/arrays.sld:185-186
(getter (if safe? (%safe-getter interval raw-getter) raw-getter))
(setter (if safe? (%safe-setter interval (storage-class-checker storage-class) raw-setter) raw-setter))
```

The raw getter applies the lexicographic indexer without first checking
`interval-contains-multi-index?`. An out-of-domain multi-index therefore
still computes a flat body offset, and when that offset happens to land
inside the body it reads or writes a *different, valid* element with no
error at all.

This is not an opt-in path: `specialized-array-default-safe?` is `#f`, so
[...]

PR #2540, opened 2026-09-06
Make SRFI 231's accumulating non-! procedures call/cc safe

### LINKED ISSUE #2539: SRFI 231: array-copy and every accumulating non-! procedure are not call/cc-safe under repeated re-entry
**Kaappi version**

`Kaappi Scheme v0.26.2` (main at 24ef1309), macOS aarch64, ReleaseSafe.

**Reported by**

Brad Lucier (gambiteer, the SRFI's author) on r/KaappiScheme: "SRFI 231: array-copy and friends are not call/cc-safe". Filed here so the fix has a tracked home.

**Description**

`array-copy` (and `array-stack`/`array-append`/`array-block`/`array-decurry`, which delegate to it) collects the source's values into a **shared, mutable scratch vector** before allocating the destination (`%array-copy-impl`, `lib/srfi/231/views.sld`, since #2454). The position is threaded functionally, so a *single* re-entry of a getter's continuation looks right — the official suite's entries 737-741 pass. But the scratch itself is one object shared by every continuation captured during the collection. A second re-entry resumes over a buffer that an earlier re-entry has already overwritten at positions *below* its own capture point, so it materializes the wrong prefix.

The reporter's point generalizes: **any strategy that keeps a partially filled structure modified by a `*-set!` procedure is not call/cc-safe**. SRFI 231 defines the term precisely — a call/cc-safe procedure is one "written in a way that does not modify the state of any data captured by a continuation" — and says "it is intended that all procedures in this library whose names do not end with an exclamation point (!) be implemented in a call/cc-safe way". The same defect, in the older `set!`-accumulator shape, is in every other accumulating non-`!` procedure of this package:

- `interval-fold-left`, `interval-fold-right` (`intervals.sld`)
- `array-fold-left`, `array-fold-right`, `array-reduce`, `array-every`, `array->list` (and `array->vector`, `array->list*`, `array->vector*` over it) (`combinators.sld`)

The reference implementation threads every accumulator functionally (`%%interval-fold-left` → `%%array->reversed-list` → a body filled from the immutable reversed list, so each re-entry gets a fresh destination), which is why Gambit returns the expected answer below.

**Minimal reproduction**

```scheme
(import (srfi 231))

(let* ((cont1 #f)
       (cont2 #f)
       (i 5)
       (first-run? #t)
       (f (lambda (x)
             (lambda (c)
               (if first-run?
                   (case x
                     ((1) (set! cont1 c))
                     ((3) (set! cont2 c))
                     (else #f)))
               x))))
       (arg (make-array (make-interval '#(4)) f))
       (result-list '()))
  (let ((temp (array-copy arg)))
    (set! first-run? #f)
    (set! result-list (cons temp result-list)))
  (case i
    ((5) (set! i (- i 1)) (cont1 10))
    ((4) (set! i (- i 1)) (cont1 20))
    ((3) (set! i (- i 1)) (cont2 10))
    ((2) (set! i (- i 1)) (cont2 20))
    (else #t))
  (display (map array->list* result-list))
  (newline))
```

**Output**

```
$ kaappi cont-test.scm
((0 20 2 20) (0 20 2 10) (0 20 2 3) (0 10 2 3) (0 1 2 3))
```

**Expected** (Gambit, running the reference implementation)

```
((0 1 2 20) (0 1 2 10) (0 20 2 3) (0 10 2 3) (0 1 2 3))
```

The 20 that the second call to `cont1` put into element 1 of *its* copy leaks into both copies made by `cont2`, whose continuation was captured on the first run when element 1 was 1.

The same shape reproduces against `interval-fold-left` with `cons` as the operator, and `array->list`.

**Fix**

Thread every accumulator functionally, as the reference does: one lexicographic walk that passes the accumulator through the axis loops and the recursion's return value, never a `set!` cell; `array-copy` collects into a reversed list and fills a freshly allocated body from it afterwards, so a re-entry re-runs the collection over the immutable prefix its own frames hold. Regression tests must exercise **two** continuations, each invoked **twice** — one re-entry alone cannot distinguish a shared buffer from a functional accumulator, which is exactly how the official suite's 737-741 passed.

baijum:
Closes #2539 — Brad Lucier's r/KaappiScheme report "SRFI 231: array-copy and friends are not call/cc-safe".

## What was wrong

`array-copy` collected a non-specialized source's values into a **shared, mutable scratch vector** before allocating the destination (#2454), threading only the position functionally. One re-entry of a getter's continuation looks right that way, and that is all the official suite's continuation cases (entries 737-741) exercise. But the scratch is one object shared by every continuation captured during the collection: a second re-entry, or a second continuation captured on the first run, resumes over positions an earlier re-entry already overwrote and materializes the wrong prefix. `array-stack`/`append`/`block`/`decurry` delegate to `array-copy`, so one loop accounted for five procedures.

| | result of the reporter's case |
|---|---|
| Gambit (reference) | `((0 1 2 20) (0 1 2 10) (0 20 2 3) (0 10 2 3) (0 1 2 3))` |
| Kaappi before | `((0 20 2 20) (0 20 2 10) (0 20 2 3) (0 10 2 3) (0 1 2 3))` |
| Kaappi after | `((0 1 2 20) (0 1 2 10) (0 20 2 3) (0 10 2 3) (0 1 2 3))` |

The spec is precise about the rule: a call/cc-safe procedure is one "written in a way that does not modify the state of any data captured by a continuation", and every procedure without a trailing `!` is intended to be one. The older `set!`-cell accumulators in `interval-fold-left`/`-right`, `array-fold-left`/`-right`, `array-reduce`, `array-every` and `array->list` are the same defect. (The two fold-left variants escaped notice only because Kaappi's left-to-right argument evaluation reads the accumulator into a register before the getter runs, so the continuation happened to carry a snapshot — the same code returns garbage under chibi's right-to-left order.)

## The fix

The reference implementation's shape, validated against the spec text and `generic-arrays.scm`:

- **`%interval-fold`** (`intervals.sld`, exported for sibling files like `arrays.sld`'s `%`-helpers): one lexicographic walk that threads the accumulator through loop variables and return values, never a cell. `interval-for-each`, both interval folds, and every accumulating combinator are built on it.
- **`array-copy`** of a non-specialized source collects into a reversed list, allocates the destination only afterwards, and fills the body by linear position through the storage class's own setter (a fresh destination's body *is* the lexicographic order), which also drops the per-element multi-index regeneration the copy-out used to do. A specialized source takes the direct fill, as in the reference (`%!array-copy`), since no user code runs in its getter.
- The costs the scratch design was chosen to avoid (#2464) return for the non-specialized path only: N live pairs during the collection.

## Tests

Regression tests in the four SRFI 231 test files drive **two continuations, each invoked twice** — a single re-entry cannot distinguish a shared buffer from a functional accumulator, which is how the previous design passed its tests. The driver is one included fixture, `tests/scheme/srfi/fixtures/srfi231-reentry.scm`. Sixteen new assertions; **nine** fail against main's `lib/srfi/231/*.sld` (verified by shadowing the checkout via an isolated `KAAPPI_HOME`): both fold-rights, `array-reduce`, `array->list`/`->vector`, `array-append`, `array-stack`, and both `array-copy` re-entry cases. The rest pin behavior: the two fold-lefts (safe on main only by evaluation-order luck, explained in a comment), `array-every` (returns the last predicate result, so only the cont2 re-entries can move it), the zero-dimensional and empty `array-copy` shapes, and two `interval-for-each` multiple-values callbacks.

- the 7 `srfi231*.scm` unit files: all pass
- `srfi231-official.scm`: 10936 passed, 0 unexpected failures, 1 known divergence (unchanged)
- `docs/dev/srfi-implementation-notes.md` gains the call/cc-safety rule; markdownlint clean

## Measurements (recorded in `views.sld`)

| shape | source | time | peak RSS |
|---|---|---|---|
| #2454 scratch vector | non-specialized, 1M elements × 20 copies | 114.9s | — |
| this PR's list collector | same | 73.5s | — |
| direct fill (kept) | specialized u8, 1M × 6 copies | 33.3s | 12.5 MB |
| list collector (not used here) | same | 18.7s | 176 MB |

The list collector is *faster* than the scratch it replaces — the old copy-out's per-element indexer call cost more than the pairs it avoided, so the #2464 objection no longer holds for that shape. For specialized sources the pairs would cost 14× the peak memory, so the direct fill stays, as in the reference's `%!array-copy`; that exemption (a custom storage-class getter or a `specialized-array-share` mapping is user code inside the direct fill) is now documented as a deliberate exception.

### DIFF
--- a/lib/srfi/231/combinators.sld
+++ b/lib/srfi/231/combinators.sld
@@ -110,14 +110,19 @@
         (interval-for-each (lambda multi-index (apply f (map (lambda (a) (apply (array-unsafe-getter a) multi-index)) all)))
                             (array-domain array))))
 
+    ;; Every accumulator below is threaded through %interval-fold rather
+    ;; than a set! cell: these are non-! procedures, which the spec
+    ;; intends to be call/cc safe, and a shared cell hands a re-entered
+    ;; getter/operator continuation the LAST run's accumulation instead
+    ;; of its own -- see %interval-fold's comment in intervals.sld
+    ;; (kaappi#2539).
     (define (array-fold-left operator identity array . arrays)
       (%check-procedure! operator "array-fold-left")
-      (let ((all (cons array arrays)) (acc identity))
+      (let ((all (cons array arrays)))
         (%check-same-domain! all "array-fold-left")
-        (interval-for-each (lambda multi-index
-                              (set! acc (apply operator acc (map (lambda (a) (apply (array-unsafe-getter a) multi-index)) all))))
-                            (array-domain array))
-        acc))
+        (%interval-fold (lambda (multi-index acc)
+                          (apply operator acc (map (lambda (a) (apply (array-unsafe-getter a) multi-index)) all)))
+                        identity (array-domain array))))
 
     ;; Per spec, ALL f-evaluations complete before ANY operator
     ;; application -- collect per-index element-lists via cons during one
@@ -126,12 +131,12 @@
     ;; interval-fold-right's own already-verified technique.
     (define (array-fold-right operator identity array . arrays)
       (%check-procedure! operator "array-fold-right")
-      (let ((all (cons array arrays)) (results '()))
+      (let ((all (cons array arrays)))
         (%check-same-domain! all "array-fold-right")
-        (interval-for-each (lambda multi-index
-                              (set! results (cons (map (lambda (a) (apply (array-unsafe-getter a) multi-index)) all) results)))
-                            (array-domain array))
-        (let loop ((xs results) (acc identity))
+        (let loop ((xs (%interval-fold (lambda (multi-index acc)
+                                         (cons (map (lambda (a) (apply (array-unsafe-getter a) multi-index)) all) acc))
+                                       '() (array-domain array)))
+                   (acc identity))
           (if (null? xs) acc (loop (cdr xs) (apply operator (append (car xs) (list acc))))))))
 
     ;; No identity/seed -- a private sentinel distinguishes "no elements
@@ -146,12 +151,11 @@
       (unless (array? array) (error "array-reduce: not an array" array))
       (unless (procedure? operator) (error "array-reduce: not a procedure" operator))
       (when (array-empty? array) (error "array-reduce: cannot reduce an empty array" array))
-      (let ((acc %array-reduce-sentinel) (getter (array-unsafe-getter array)))
-        (interval-for-each (lambda multi-index
-                              (let ((val (apply getter multi-index)))
-                                (set! acc (if (eq? acc %array-reduce-sentinel) val (operator acc val)))))
-                            (array-domain array))
-        acc))
+      (let ((getter (array-unsafe-getter array)))
+        (%interval-fold (lambda (multi-index acc)
+                          (let ((val (apply getter multi-index)))
+                            (if (eq? acc %array-reduce-sentinel) val (operator acc val))))
+                        %array-reduce-sentinel (array-domain array))))
 
     ;; Short-circuits via call/cc on the first non-#f result, returning
     ;; that ACTUAL result (not just #t) -- matching the spec's own
@@ -172,15 +176,14 @@
     ;; returns the LAST nonfalse result, not just #t.
     (define (array-every predicate array . arrays)
       (%check-procedure! predicate "array-every")
-      (let ((all (cons array arrays)) (last-result #t))
+      (let ((all (cons array arrays)))
         (%check-same-domain! all "array-every")
         (call/cc
          (lambda (return)
-           (interval-for-each (lambda multi-index
-                                 (let ((r (apply predicate (map (lambda (a) (apply (array-unsafe-getter a) multi-index)) all))))
-                                   (if r (set! last-result r) (return #f))))
-                               (array-domain array))
-           last-result))))
+           (%interval-fold (lambda (multi-index last-result)
+                             (let ((r (apply predicate (map (lambda (a) (apply (array-unsafe-getter a) multi-index)) all))))
+                               (if r r (return #f))))
+                           #t (array-domain array))))))
 
     ;; --- products ---
 
@@ -217,10 +220,11 @@
 
     (define (array->list array)
       (unless (array? array) (error "array->list: not an array" array))
-      (let ((acc '()))
-        (interval-for-each (lambda multi-index (set! acc (cons (apply (array-unsafe-getter array) multi-index) acc)))
-                            (array-domain array))
-        (reverse acc)))
+      ;; `reverse`, never reverse!: a re-entered getter continuation still
+      ;; holds the prefix of the reversed list its own run built.
+      (let ((getter (array-unsafe-getter array)))
+        (reverse (%interval-fold (lambda (multi-index acc) (cons (apply getter multi-index) acc))
+                                 '() (array-domain array)))))
 
     (define (array->vector array) (list->vector (array->list array)))
 
--- a/lib/srfi/231/combinators.sld
+++ b/lib/srfi/231/combinators.sld
@@ -228,6 +228,13 @@
 
     (define (array->vector array) (list->vector (array->list array)))
 
+    ;; list->array and vector->array below still thread their position
+    ;; through a set! cell. That is safe ONLY because no user code runs
+    ;; inside the walk: the checker runs before the loop (#2448) and the
+    ;; setter is a fresh built-in-or-custom storage class's, called with
+    ;; already-checked values. Moving any user-callable call (the checker,
+    ;; an f applied per element) into the loop would reopen the
+    ;; shared-cell hazard %interval-fold exists to close (kaappi#2539).
     (define (list->array interval lst . opts)
       (unless (interval? interval) (error "list->array: not an interval" interval))
       (unless (list? lst) (error "list->array: not a list" lst))
--- a/lib/srfi/231/intervals.sld
+++ b/lib/srfi/231/intervals.sld
@@ -193,8 +193,10 @@
                       acc)))))
         (go 0 '() seed)))
 
+    ;; f's result is discarded, never threaded: interval-for-each is for
+    ;; effect, and f may return zero or several values.
     (define (interval-for-each f interval)
-      (%interval-fold (lambda (indices acc) (apply f indices)) #f interval)
+      (%interval-fold (lambda (indices acc) (apply f indices) acc) #f interval)
       (if #f #f))
 
     (define (interval-fold-left f operator identity interval)
--- a/lib/srfi/231/intervals.sld
+++ b/lib/srfi/231/intervals.sld
@@ -39,7 +39,11 @@
           interval= interval-subset? interval-contains-multi-index?
           interval-for-each interval-fold-left interval-fold-right
           interval-dilate interval-translate interval-permute interval-scale
-          interval-intersect interval-cartesian-product interval-projections)
+          interval-intersect interval-cartesian-product interval-projections
+          ;; internal: the functional lexicographic walk sibling files build
+          ;; their own accumulating procedures on (see its comment); not
+          ;; re-exported by the (srfi 231) hub, same as arrays.sld's %-helpers
+          %interval-fold)
   (begin
 
     (define-record-type <interval>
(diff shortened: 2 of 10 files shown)


lib/srfi/231/combinators.sld
```diff
@@ -217,10 +220,11 @@
 
     (define (array->list array)
       (unless (array? array) (error "array->list: not an array" array))
-      (let ((acc '()))
-        (interval-for-each (lambda multi-index (set! acc (cons (apply (array-unsafe-getter array) multi-index) acc)))
-                            (array-domain array))
-        (reverse acc)))
+      ;; `reverse`, never reverse!: a re-entered getter continuation still
```
baijum, line 223: Same defect family as this PR, exempted only by luck of content: `list->array` and `vector->array` (just below, the `set! i` at line 291) still thread their position through set! cells. They're safe today solely because *no user code runs inside the walk* — the checker is deliberately run before the loop (#2448) — so nothing can capture a continuation mid-fill, which is the same reasoning this PR uses to let a specialized source take the direct fill. Worth one line saying that, so a future edit that moves any user-callable call (e.g. the checker, or an f applied per element) into the loop doesn't silently reopen the shared-cell hazard this PR is fixing everywhere else.

tests/scheme/srfi/srfi231-intervals.scm
```diff
@@ -192,6 +192,53 @@
...
+
+;;; --- call/cc safety under REPEATED re-entry (kaappi#2539, reported by
+;;; the SRFI's author) ---
+;;; Drives `collect` (a procedure of one argument: a getter over 0..3 that
+;;; captures its continuation at x=1 and x=3 on the first run) through a
+;;; two-continuation, two-re-entry schedule and returns every result,
+;;; newest first. Any collector keeping a shared mutable accumulator --
+;;; a set! cell or a scratch vector -- leaks the second cont1 re-entry's
+;;; 20 into both cont2 results; one re-entry alone cannot see that.
+(define (re-entry-results collect)
```
baijum: `re-entry-results` + `re-entry-expected` are duplicated verbatim in all four test files (~55 lines each). If the files must stay standalone that's defensible, but if there's any shared-include mechanism in the suite this is a lot of copy for a helper that future SRFI 231 work may well need again (any new non-! accumulator has to be tested against it). At minimum the other three copies should reference this comment block's rationale rather than repeating it — three of the four copies currently re-explain it identically.

baijum (commented):
Verified on the branch (built + isolated KAAPPI_HOME):

- all four updated srfi231 unit files pass; `srfi231-official.scm` reports 10936 passed / 0 unexpected failures / 1 known divergence, exactly as claimed
- pre-fix check confirmed by running the new tests with main's `lib/srfi/231/*.sld` shadowing via `KAAPPI_HOME`: 9 unexpected failures (both fold-rights, array-reduce, array->list/vector, append, stack, both array-copy re-entry tests) and the two fold-lefts pass — consistent with the register-snapshot explanation
- `specialized?` binding, `storage-class-setter`'s `(body index val)` signature, the linear lex-indexer position, and the non-re-export of `%interval-fold` from the `(srfi 231)` hub all check out

One measurement worth recording: the list collector is not just safe, it is *faster* than the #2454 scratch design — 73.5s vs 114.9s for 20× `array-copy` of a 1M-element non-specialized array. The copy-out's per-element multi-index regeneration (%make-lex-indexer per element) was costing more than the N live pairs it avoided, so the #2464 objection no longer holds for this shape. Suggest adding that to the views.sld comment (or the PR body), since #2464 is the reason the scratch existed and a future reader will wonder why the costlier-looking shape won.

Two nits below, one about the PR description's arithmetic: there are 13 new assertions (2 intervals + 5 combinators + 4 views + 2 assembly), and **9** of them fail pre-fix — the four passing pre-fix are the two fold-lefts plus the zero-dimensional and empty array-copy tests, which don't drive continuations. The story is unchanged but "ten of the twelve" is off on both numbers.

No blockers — approving with the nits.

lib/srfi/231/intervals.sld
```diff
@@ -142,36 +146,60 @@
...
+                (let loop ((i lo) (acc acc))
+                  (if (< i hi)
+                      (loop (+ i 1) (go (+ axis 1) (cons i rev-index) acc))
+                      acc)))))
+        (go 0 '() seed)))
 
     (define (interval-for-each f interval)
-      (%interval-for-each-indices (interval-lower-vec interval) (interval-upper-vec interval)
-                                   (lambda (indices) (apply f indices))))
+      (%interval-fold (lambda (indices acc) (apply f indices)) #f interval)
```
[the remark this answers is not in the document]
baijum: Fixed in 90131fc4: the callback's result is discarded (`(apply f indices) acc`). For the record it did not fail here — Kaappi tolerates a multiple-values object as a plain value, and nothing read the threaded slot — but for-each is for effect and the shape was wrong. Two assertions in srfi231-intervals.scm now pin zero-values and two-values callbacks.

tests/scheme/srfi/srfi231-combinators.scm
```diff
@@ -214,6 +214,61 @@
...
+            (re-entry-results (lambda (f) (array-fold-right cons '() (over-f f)))))
+(test-equal "array-reduce threads its accumulator functionally"
+            '(23 13 25 15 6)
+            (re-entry-results (lambda (f) (array-reduce + (over-f f)))))
+(test-equal "array->list threads its accumulator functionally"
+            re-entry-expected
+            (re-entry-results (lambda (f) (array->list (over-f f)))))
+(test-equal "array->vector threads its accumulator functionally"
+            (map list->vector re-entry-expected)
+            (re-entry-results (lambda (f) (array->vector (over-f f)))))
```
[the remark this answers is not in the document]
baijum: Added in 90131fc4, but not with the suggested expectation: `(lambda (x) x)` over elements 0..3 returns #f at element 0, so array-every exits before either continuation is captured and the driver's `(cont1 10)` then calls #f. array-every returns the LAST predicate result, so only the cont2 re-entries (which replace element 3) can move it; the assertion uses a predicate that tags values ≥ 10 and expects `((last 20) (last 10) 3 3 3)`. Like the fold-lefts it pins the functional shape rather than reproducing a failure — the old set!-cell version already gave the same answers, since every re-entry runs through the last element.

tests/scheme/srfi/srfi231-views.scm
```diff
@@ -356,6 +356,51 @@
...
+(define re-entry-expected '((0 1 2 20) (0 1 2 10) (0 20 2 3) (0 10 2 3) (0 1 2 3)))
+
+(test-equal "array-copy: each re-entry materializes its own array from its own prefix"
+            re-entry-expected
+            (re-entry-results
+             (lambda (f) (array->list* (array-copy (make-array (make-interval '#(4)) f))))))
+(test-equal "array-copy into a typed storage class is call/cc safe too"
+            re-entry-expected
+            (re-entry-results
+             (lambda (f) (array->list (array-copy (make-array (make-interval '#(4)) f) u8-storage-class)))))
```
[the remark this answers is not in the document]
baijum: Declining the implementation change; taking the documentation half. The direct fill for specialized sources is the reference implementation's own shape (`%!array-copy` in generic-arrays.scm takes `%%move-array-elements` for any `specialized-array?` source, custom storage class or shared indexer included), and this SRFI's documented rule here is to follow its code where prose and code diverge. I measured the alternative: routing a 1M-element u8 specialized source through the list collector is 1.8× faster (18.7s vs 33.3s for 6 copies) but peaks at 176 MB RSS against 12.5 MB — 14× the memory on the most common typed-array copy, for a hazard that needs a continuation-capturing `make-storage-class` getter or `specialized-array-share` mapping. 90131fc4 rewrites the dev-notes sentence so the exemption is stated as a deliberate exception naming exactly those two user-code routes, rather than as "no user code runs"; the views.sld comment already carried it.

lib/srfi/231/combinators.sld
baijum: Done in 90131fc4 — a comment above list->array says both walks are exempt only because no user code runs inside them (checker before the loop, #2448) and that moving any user-callable call into the loop reopens the shared-cell hazard.

tests/scheme/srfi/srfi231-intervals.scm
baijum: Done in 90131fc4. `include` resolves relative to the script and tests/scheme/CLAUDE.md already sanctions fixture subdirectories, so the driver is now the one file tests/scheme/srfi/fixtures/srfi231-reentry.scm, included by all four suites; it opens no suite so run-all's reachability check ignores it. The rationale lives once, in the fixture.

baijum:
Thanks for the verification and the measurement. Both nits taken in 90131fc4: the 73.5s vs 114.9s number is now in the views.sld comment next to the cost claim (with a second measurement explaining why the direct fill nevertheless stays for specialized sources: 1.8× faster through the list but 14× the peak RSS), and the PR description's arithmetic is corrected — with the review additions it is now sixteen assertions, nine of which fail against main.

PR #2541, opened 2026-09-07
Add differential testing of SRFI 231 against the reference implementation

baijum:
Follow-up to #2540. The official suite passed all 10,936 evaluations while `array-copy` broke the spec's definition of call/cc-safe (#2539): fixed cases find bugs at the rate of ideas. This adds `tools/srfi231_diff.py`, which generates random SRFI 231 programs from a seed, runs each under Kaappi and an oracle, and reports every program whose printed output differs.

**Oracle.** The reference implementation itself, as bundled by Gambit (`(srfi 231)` in 4.9.8), since this SRFI's documented rule is "trust the code" where prose and code disagree. Chibi's independent port is the tie-breaker (`--oracle chibi`).

**First mode: `reshape`.** A chain of view operations (extract, translate, permute, reverse, sample, curry-pick, tile-pick, copy, `specialized-array-reshape` with and without `copy-on-failure?`) over one specialized array whose elements are their own multi-index. After every step it prints domain, `specialized-array?`, `mutable-array?`, `array-packed?` and contents; at the end it writes through the final view and prints the base, so body sharing across the chain is compared. Every step is guarded, so whether a step errors is compared too. The generator is biased toward the affine boundary (merging axes whose strides stop chaining after a permute or sample), which is where the reshape code reasons about index arithmetic rather than copying the reference's structure.

**Results so far.**

| oracle | cases | mismatches |
|---|---|---|
| Gambit 4.9.8 | 1,500 (seeds 1000..2499) | 0 |
| chibi 0.12 | 500 (seeds 5000..5499) | 4, all one chibi bug, Gambit sides with Kaappi on each |

Chibi raises on a `copy-on-failure? #t` reshape that needs the copy, where the spec, Gambit and Kaappi return one (seeds 5227, 5248, 5281, 5412; every later error in those chains is the guarded step keeping the previous array). Recorded in the tool's docstring and `docs/dev/testing.md` as a known oracle divergence.

**Second mode: `storage`.** One storage class per case through everything that consults its checker, getter and setter (checker verdicts on boundary and wrong-typed values, `make-specialized-array` with an initial value, `array-set!` directly and through a reversed extract, `array-copy`/`list->array`/`vector->array` from a generic source, `array-assign!` into a view, `storage-class-length`). Values are canonicalized before printing (finite reals as exact rationals, complex as pairs of those, chars as code points) so f16/f32 rounding is compared bit-exactly and Gambit's `.5` never differs textually from Kaappi's `0.5`. `--classes u1,f16` narrows the draw.

| oracle | cases | mismatches |
|---|---|---|
| Gambit, all 16 classes | 400 | 48, every one on c64/c128 and every one **kaappi#2542** |
| Gambit, the other 14 classes (u1, f16, u64 … included) | 120 | 0 |
| chibi, all 16 classes | 100 | 22, all chibi's own (u16 accepts 65536, u64 accepts negatives, unvalidated initial values, c64 accepts reals); Gambit contradicts each, and Gambit's only disagreements with Kaappi on those programs are #2542 |

**#2542** is the mode's first real finding: the reference's c64/c128 checker rejects a real flonum because Gambit's `imag-part` returns an exact `0`, while Kaappi's line-for-line identical checker accepts it because its `imag-part` returns `0.0` (R7RS permits either; the SRFI's "trust the code" rule makes it a Kaappi divergence).

Two more oracle divergences had to be neutralized: Gambit's bundled reference flips `specialized-array-default-safe?` to `#t` against the spec and the SRFI repository's own copy of the same file (the storage prelude pins it to `#f`), and chibi defines a real `f8-storage-class` where the spec allows it (no f8 probe).

Each case is a pure function of (mode, seed); a mismatch is saved with both outputs and the run exits 1, to be reduced and pinned as an ordinary regression test. Not part of `run-all.sh` (needs an oracle, takes minutes). New modes are one generator function each; call/cc under every callback-taking procedure and the spec's prose examples are the planned next ones.

### DIFF
--- a/tools/srfi231_diff.py
+++ b/tools/srfi231_diff.py
@@ -1,0 +1,412 @@
+#!/usr/bin/env python3
+"""Differential testing of Kaappi's SRFI 231 against the reference implementation.
+
+SRFI 231's documented rule in this codebase is "when the spec's prose and its
+reference implementation disagree, trust the code" -- which makes the reference
+an oracle in the strict sense. Gambit bundles it as `(srfi 231)`, and chibi
+ships an independent port. This tool generates random SRFI 231 programs in the
+subset of R7RS all three run, executes each under Kaappi and an oracle, and
+reports any program whose printed output differs.
+
+Why this exists: the official conformance suite (10,936 evaluations) encodes
+the reference's answers on the inputs its author wrote down. It cannot see a
+property violation its cases never exercise -- kaappi#2539 passed the whole
+suite while breaking the spec's own definition of call/cc-safe. Generated
+inputs find bugs at the rate of inputs, not ideas.
+
+It is **not** part of `run-all.sh`: it needs an oracle installed (`brew install
+gambit-scheme` or `chibi-scheme`) and a useful run takes minutes. Cases it
+finds get reduced and pinned as ordinary regression tests under
+tests/scheme/srfi/, with the seed in the comment so they replay:
+
+    tools/srfi231_diff.py reshape --count 500              # gambit oracle
+    tools/srfi231_diff.py reshape --oracle chibi --count 200
+    tools/srfi231_diff.py reshape --seed 7 --count 1        # exactly one case
+    tools/srfi231_diff.py reshape --seed 7 --print          # show its program
+
+Modes
+  reshape  A chain of view operations over one specialized array -- extract,
+           translate, permute, reverse, sample, curry-pick, tile-pick, copy,
+           and specialized-array-reshape with and without copy-on-failure? --
+           printing the domain, array-packed?, mutability and contents after
+           every step, then writing through the final view and printing the
+           base array to compare body sharing. Every step is guarded, so
+           "errors here" versus "does not" is compared too. Aimed at the
+           NumPy-derived affine-reshape detection in lib/srfi/231/views.sld
+           and at array-packed? over composed views, the two places where the
+           implementation reasons about index arithmetic rather than copying
+           the reference's structure.
+
+Known oracle divergence: chibi 0.12's port raises on a `copy-on-failure? #t`
+reshape that needs the copy, where the spec (and Gambit, and Kaappi) return a
+copy -- expect that mismatch shape with `--oracle chibi` (seeds 5227 and 5248
+of the reshape mode show it) and confirm against Gambit before chasing it.
+
+Each case is a pure function of (mode, seed); `--seed N --count K` runs seeds
+N..N+K-1. Mismatches are saved as <save-dir>/<mode>-<seed>.scm with the two
+outputs beside them, and the run exits 1.
+"""
+
+import argparse
+import os
+import random
+import shutil
+import subprocess
+import sys
+import tempfile
+
+HERE = os.path.dirname(os.path.abspath(__file__))
+ROOT = os.path.dirname(HERE)
+
+PRELUDE = """\
+(import (scheme base) (scheme write) (srfi 231))
+(define (show . xs) (for-each (lambda (x) (write x) (display " ")) xs) (newline))
+(define (dom a)
+  (let ((d (array-domain a)))
+    (list (interval-lower-bounds->list d) (interval-upper-bounds->list d))))
+(define (describe tag a)
+  (show tag (dom a) (specialized-array? a) (mutable-array? a) (array-packed? a)
+        (array->list* a)))
+(define-syntax try
+  (syntax-rules () ((_ e) (guard (c (#t 'ERROR)) e))))
+;; write-through probe: set the r-th multi-index (lexicographic) of a to 'X
+;; and print the BASE array, so body sharing across the whole chain is
+;; compared; the index is chosen here, from a's actual domain, so it is
+;; valid whatever the chain did
+(define (probe a r)
+  (if (array-empty? a)
+      (show 'write-through 'empty)
+      (let* ((idxs (reverse (interval-fold-left list (lambda (acc x) (cons x acc)) '()
+                                               (array-domain a))))
+             (idx (list-ref idxs (modulo r (length idxs)))))
+        (show 'write-through idx
+              (try (begin (apply array-set! a 'X idx) (array->list* a0)))))))
+;; a step: bind the result if it succeeded, else keep the previous array so
+;; the chain (and the oracle comparison of every later step) continues
+(define-syntax step
+  (syntax-rules ()
+    ((_ name tag prev e)
+     (define name (let ((r (try e)))
+                    (if (eq? r 'ERROR) (begin (show tag 'ERROR) prev)
+                        (begin (describe tag r) r)))))))
+"""
+
+
+# --- interval helpers (Python side) ----------------------------------------
+
+def volume(lo, hi):
+    v = 1
+    for l, h in zip(lo, hi):
+        v *= h - l
+    return v
+
+
+def fmt_vec(xs):
+    return "'#(" + " ".join(str(x) for x in xs) + ")"
+
+
+def fmt_interval(lo, hi):
+    return f"(make-interval {fmt_vec(lo)} {fmt_vec(hi)})"
+
+
+def random_factorization(rng, v, max_axes=4):
+    """Split v >= 1 into 0..max_axes factors >= 1, width-1 axes included."""
+    if v == 1 and rng.random() < 0.2:
+        return []
+    n = rng.randint(1, max_axes)
+    dims = [1] * n
+    rest = v
+    # distribute prime factors randomly across the axes
+    p = 2
+    while rest > 1:
+        while rest % p == 0:
+            dims[rng.randrange(n)] *= p
+            rest //= p
+        p += 1
+    return dims
+
+
+def merge_or_split(rng, widths):
+    """Merge two adjacent axes into one, or split one axis into two factors.
+    Merging axes with non-chaining strides is the exact affine boundary."""
+    w = list(widths)
+    if len(w) >= 2 and (rng.random() < 0.6 or not any(x > 1 for x in w)):
+        i = rng.randrange(len(w) - 1)
+        return w[:i] + [w[i] * w[i + 1]] + w[i + 2:]
+    cands = [i for i, x in enumerate(w) if x > 1]
+    if not cands:
+        return w + [1]
+    i = rng.choice(cands)
+    divs = [f for f in range(1, w[i] + 1) if w[i] % f == 0]
+    f = rng.choice(divs)
+    return w[:i] + [f, w[i] // f] + w[i + 1:]
+
+
+def random_same_volume_domain(rng, lo, hi):
+    v = volume(lo, hi)
+    if v > 0 and rng.random() < 0.5:
+        widths = merge_or_split(rng, [h - l for l, h in zip(lo, hi)])
+        nlo = [rng.randint(-2, 2) for _ in widths]
+        return nlo, [l + w for l, w in zip(nlo, widths)]
+    if v == 0:
+        n = rng.randint(1, 4)
+        widths = [rng.randint(0, 3) for _ in range(n)]
+        widths[rng.randrange(n)] = 0
+    else:
+        widths = random_factorization(rng, v)
+    nlo = [rng.randint(-2, 2) for _ in widths]
+    return nlo, [l + w for l, w in zip(nlo, widths)]
+
+
+# --- the reshape-mode generator ---------------------------------------------
+
+class Chain:
+    """Tracks the current array's domain on the Python side so every generated
+    argument is valid by construction; the guards catch what is not."""
+
+    def __init__(self, rng):
+        self.rng = rng
+        self.lines = [PRELUDE]
+        # mostly 2-3 axes of width 2-4 -- the shapes where affine-reshape
+        # detection has real work to do -- with a tail of empty, width-1
+        # and zero-dimensional cases
+        d = rng.choice([0, 1, 2, 2, 2, 3, 3, 3, 4])
+        self.lo = [rng.randint(-2, 2) for _ in range(d)]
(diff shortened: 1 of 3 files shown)


tools/srfi231_diff.py
```diff
@@ -0,0 +1,584 @@
...
+value of `specialized-array-default-safe?` to #t (spec and the SRFI
+repository's copy: #f); the storage mode pins it to #f in its prelude.
+Kaappi's own open divergence kaappi#2542 (c64/c128 checkers accept real
+flonums, the reference rejects them) surfaces as `check` mismatches on those
+two classes; use `--classes` to exclude them until it is fixed, since a
+program's first difference hides everything after it. And chibi 0.12's port raises on a `copy-on-failure? #t`
+reshape that needs the copy, where the spec (and Gambit, and Kaappi) return a
+copy -- expect that mismatch shape with `--oracle chibi` (seeds 5227 and 5248
+of the reshape mode show it) and confirm against Gambit before chasing it.
+Its storage classes have checker bugs of their own (u16 accepts 65536, u64
```
baijum, line 66: This list is missing one chibi divergence I hit immediately: **chibi's u1 checker returns non-boolean truthy values**, so every u1-class storage case mismatches before anything array-related is compared.

Repro: `tools/srfi231_diff.py storage --oracle chibi --classes u1 --seed 700 --count 1`

```
kaappi:  check 1 #t
chibi  : check 1 (1)
```

Direct probe of `((storage-class-checker u1-storage-class) 0)`: Kaappi `#t`, Gambit `#t`, chibi `(0 1)` — chibi's u1 checker is evidently `memv`-shaped and returns the tail instead of a boolean, so the *verdict* agrees everywhere (0, 1 accepted; 2 and 1.0 rejected) and only the printed form differs. I confirmed it's systematic (seeds 600, 700–702 all show it as the first difference); Gambit sides with Kaappi on each.

Two ways to handle it:

1. Add it to this list and to `docs/dev/testing.md`'s divergence paragraph (which also omits chibi's unvalidated initial values, mentioned in the PR description but not in the docstring).
2. Nicer: canonicalize the printed verdict — `(show 'check (canon v) (and ((storage-class-checker sc) v) #t))` — which suppresses only the representation difference. Every *semantic* verdict difference (u16 accepting 65536, u64 accepting negatives, c64) still fires, so no signal is lost.

I'd do 2 plus a docstring line, but 1 alone is fine.

tools/srfi231_diff.py
```diff
@@ -0,0 +1,584 @@
...
+        return MODES[args.mode](seed, classes)
+
+    if args.print:
+        sys.stdout.write(gen(args.seed))
+        return 0
+
+    oracle_cmd = [args.gsi] if args.oracle == "gambit" else [args.chibi]
+    save = args.save or tempfile.mkdtemp(prefix="srfi231-diff-")
+    os.makedirs(save, exist_ok=True)
+    work = tempfile.mkdtemp(prefix="srfi231-diff-work-")
```
baijum, line 563: `work` is never removed, so every run leaves a `srfi231-diff-work-*` dir in $TMPDIR with the generated cases plus the warm KAAPPI_HOME cache — 1–6 MB each (24 of them from my session alone). The default save dir has the same issue one level down: it's created even on a fully clean run, and its path is only printed when there's a mismatch, so clean runs leave an empty anonymous dir behind.

`fmt_fuzz.py` already has the pattern to copy: `try/finally: shutil.rmtree(scratch(), ignore_errors=True)` with a `--keep` opt-out. The save dir could be created lazily on the first mismatch instead.

tools/srfi231_diff.py
```diff
@@ -0,0 +1,584 @@
...
+            sys.exit(f"unknown storage class(es): {', '.join(bad)}; known: {', '.join(CLASSES)}")
+
+    def gen(seed):
+        return MODES[args.mode](seed, classes)
+
+    if args.print:
+        sys.stdout.write(gen(args.seed))
+        return 0
+
+    oracle_cmd = [args.gsi] if args.oracle == "gambit" else [args.chibi]
```
baijum: If the oracle or kaappi binary doesn't exist — the default `/opt/homebrew/bin/gsi` on a Linux box, or an unbuilt worktree — `run_one` raises a raw `FileNotFoundError` traceback on the first case (only `TimeoutExpired` is caught).

`fmt_fuzz.py` has the precedent: `if not os.access(KAAPPI, os.X_OK): sys.exit("no kaappi binary at %s (set $KAAPPI or run zig build)")`. Checking both binaries up front would also fail before generating any programs rather than after the first one.

tools/srfi231_diff.py
```diff
@@ -0,0 +1,584 @@
...
+import random
+import shutil
+import subprocess
+import sys
+import tempfile
+
+HERE = os.path.dirname(os.path.abspath(__file__))
+ROOT = os.path.dirname(HERE)
+
+PRELUDE = """\
```
baijum, line 88: Asymmetry worth a line of justification: the storage prelude pins `specialized-array-default-safe?` to `#f` (because Gambit's bundled copy flips it to `#t`), but this reshape prelude doesn't. As far as I can tell that's deliberate and safe — arrays here are generic-storage-class so the checker accepts anything, every index is valid by construction, and `describe` doesn't print `array-safe?`, so nothing in this mode observes the flag (and the 1,500 clean Gambit cases back that up). But a reader comparing the two preludes will read the omission as an oversight. A one-line comment here or next to the pin in `STORAGE_PRELUDE` ("left unpinned in reshape mode: nothing there observes safety") would preempt the question.

tools/srfi231_diff.py
```diff
@@ -0,0 +1,584 @@
...
+        if self.rng.random() < 0.5:
+            expr = f"(specialized-array-reshape {self.cur} {fmt_interval(lo, hi)} #t)"
+            self.emit("reshape-copy", expr, lo, hi)
+            return True
+        # Without copy-on-failure? the reshape may legitimately error, and
+        # the step then keeps the previous array while the Python-side
+        # domain would advance -- so this is always the chain's last step.
+        expr = f"(specialized-array-reshape {self.cur} {fmt_interval(lo, hi)})"
+        self.emit("reshape", expr, lo, hi)
+        self.terminal = True
```
baijum, line 340: Nit: `terminal` is only ever assigned here and read via `getattr(c, "terminal", False)` in `gen_reshape`. Initializing `self.terminal = False` in `__init__` would make the state part of the constructor contract and turn a typo into an `AttributeError` instead of a silently-False.

baijum (commented):
Ran the tool end-to-end on this machine against both oracles and reviewed the generator in detail. It's solid — verified:

- **Both modes work and agree with the PR's numbers**: `reshape` vs Gambit clean on several seed ranges; `storage` vs Gambit clean on u1/f16/u64/f64; determinism holds (`--print` twice, byte-identical); exit codes right (1 on mismatch, 0 clean).
- **Every documented divergence reproduces exactly as described**: chibi's copy-on-failure? reshape raise (seed 5227, same output shape), and #2542 (5/5 c64-only seeds mismatch on `check`).
- **The generated Scheme is valid under all three implementations**, and the Python-side domain tracking is correct by construction — I hand-checked the index arithmetic for extract, translate, permute, reverse, sample (incl. the translate-to-zero step), curry-pick, tile-pick (both side and cut-vector forms, zero-width axes) and reshape, including empty and 0-dimensional edge cases. The 4-argument `interval-fold-left` call matches the SRFI 231 `(kernel combiner identity interval)` signature, and `array-packed?`, `mutable-array?`, `array-empty?`, `array->list*` are all present in Gambit's bundled `(srfi 231)` (probed directly — I initially suspected `array-packed?` was a Kaappi extension; it isn't).
- **Docs**: markdownlint (same rule set as the CI format job) reports 0 issues across the repo; the testing.md section sits at the right heading level and the implementation-notes edit reads fine in context.

One new finding from actually running it: **chibi's u1 checker returns non-boolean truthy values**, which isn't in the known-divergences list and fires on every u1-class storage case (inline comment with the repro). That's the only thing I'd fix before merge — it's exactly the kind of noise the list exists to prevent. The rest are small robustness items, both with precedent in `tools/fmt_fuzz.py`.

tools/srfi231_diff.py
```diff
@@ -0,0 +1,412 @@
...
+
+    mismatches = 0
+    for seed in range(args.seed, args.seed + args.count):
+        prog = gen(seed)
+        path = os.path.join(work, f"{args.mode}-{seed}.scm")
+        with open(path, "w") as f:
+            f.write(prog)
+        kc, ko, ke = run_one([args.kaappi], path, env, args.timeout)
+        oc, oo, oe = run_one(oracle_cmd, path, env, args.timeout)
+        same = normalize(ko) == normalize(oo) and (kc == 0) == (oc == 0)
```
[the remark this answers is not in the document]
baijum: Fixed in 4a10bc48. A timeout on either side is now always a mismatch: `run_one` returns a `TIMED_OUT` sentinel that never compares equal to an exit status, the saved output file says `timed out` in place of the exit line, and the report names which side hung. Verified with `--timeout 0.001`.

tools/srfi231_diff.py
```diff
@@ -0,0 +1,412 @@
...
+
+    gen = MODES[args.mode]
+    if args.print:
+        sys.stdout.write(gen(args.seed))
+        return 0
+
+    oracle_cmd = [args.gsi] if args.oracle == "gambit" else [args.chibi]
+    save = args.save or tempfile.mkdtemp(prefix="srfi231-diff-")
+    os.makedirs(save, exist_ok=True)
+    work = tempfile.mkdtemp(prefix="srfi231-diff-work-")
```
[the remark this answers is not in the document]
baijum: Fixed in 4a10bc48, following `tools/fmt_fuzz.py`: the work dir is removed in a `finally` (`--keep` opts out and prints the path), and the save dir is created on the first mismatch rather than up front, so a clean run leaves nothing in $TMPDIR. An explicit `--save` is preserved as before.

tools/srfi231_diff.py
baijum: Took option 2 plus the docstring line, in 4a10bc48: the check line is now `(and ((storage-class-checker sc) v) #t)`, so only the boolean verdict is compared. Your repro (`--oracle chibi --classes u1 --seed 700 --count 3`) drops from 3/3 to 1/3 mismatches, and the one left is chibi accepting 2^63-1 as a u1 initial value — its unvalidated-initial-value bug, which the docstring and the testing.md paragraph now list alongside the u16/u64/c64 checker bugs and the u1 non-boolean itself. Every semantic verdict difference still fires; I re-ran the c64 seeds to confirm #2542 is unchanged.

tools/srfi231_diff.py
baijum: Done in 4a10bc48, same shape as fmt_fuzz.py: `try/finally: shutil.rmtree(work)` with `--keep` as the opt-out (it prints the kept path), and the default save dir is created lazily on the first mismatch. Counted `srfi231-diff-*` dirs in $TMPDIR before and after a clean run: unchanged.

tools/srfi231_diff.py
baijum: Done in 4a10bc48: both binaries are checked with `os.access(..., X_OK) or shutil.which(...)` before any case is generated, and a missing one exits with a one-line message naming the binary and the fix (`run zig build, or pass --kaappi` / `brew install gambit-scheme / chibi-scheme, or pass --gsi/--chibi`). Verified with `--gsi /nonexistent/gsi`.

tools/srfi231_diff.py
baijum: Your reading is right — it is deliberate, and the reshape prelude now says so in 4a10bc48: every array there is generic (the checker accepts anything), every index is valid by construction, and nothing printed observes the flag, so Gambit's flipped default cannot show; the storage mode prints `array-safe?`, hence pins it.

tools/srfi231_diff.py
baijum: Done in 4a10bc48: `self.terminal = False` in `Chain.__init__`, with a comment saying what sets it, and `gen_reshape` reads `c.terminal` directly.

baijum (commented):
Verified every fix on this commit by running the tool — all five comments are addressed, and the timeout reclassification closes a hole my review missed (the old `-1` sentinel made a one-sided hang with matching partial output count as a *match*, the one case a differential tool must never hide).

- **u1 checker spelling** (`(and … #t)` wrap): `storage --oracle chibi --classes u1 --seed 700 --count 3` — seeds 701/702 now pass, and 700's first difference moved from the `check` line to line 10, `make 9223372036854775807 ERROR`: chibi accepting an out-of-range initial value, i.e. the semantic init-validation bug now surfaces instead of the spelling noise. Docstring list updated with both that and the u1 non-boolean return, as suggested.
- **Temp dirs**: a clean run leaves no work dir behind (count unchanged before/after); `--keep` keeps it and prints the path; the save dir is created only on the first mismatch.
- **Missing binaries**: `--gsi /nonexistent/gsi` and `--kaappi /nonexistent/kaappi` each produce a one-line hint and exit 1, before any program is generated.
- **Timeouts**: `--timeout 0.001` reports `timed out after 0.001s: kaappi, gambit` and the saved artifact's status line reads `timed out`; the double-timeout case the old code scored as a match is now a mismatch. Nice detail: `TimeoutExpired.stdout` is bytes even under `text=True` on POSIX, and `(e.stdout or b"").decode(errors="replace")` handles it — plus it preserves the partial output in the saved artifact, which the old empty-string sentinel threw away.
- **default-safe? note and `Chain.terminal`**: both in, and the PRELUDE comment states the reason precisely (nothing in reshape mode observes the flag; storage prints `array-safe?`, hence the pin).

Regression battery on this commit: reshape vs Gambit (seeds 100–107) and storage vs Gambit over u1/f16/u64/f64/generic — all clean; chibi seed 5227 still shows exactly the documented copy-on-failure divergence shape. Compiles, still `100755`. From my side this is ready to merge.

PR #2543, opened 2026-09-07
Reject bare real flonums in the c64/c128 storage-class checkers

### LINKED ISSUE #2542: SRFI 231: c64/c128 storage-class checkers accept real flonums that the reference rejects
**Kaappi version**

`Kaappi Scheme v0.26.2` (main at 5e70be41), macOS aarch64.

**Found by**

`tools/srfi231_diff.py storage` (PR #2541), differential testing against the reference implementation under Gambit 4.9.8. Seeds 3, 12, 27, 40 of the storage mode. (chibi 0.12's port returns an exact `0` from `imag-part` like Gambit yet still accepts `1.0`, so its checker is not the reference's; it is not an oracle for this.)

**Description**

The reference implementation's checker for `c64-storage-class` / `c128-storage-class` is

```scheme
(lambda (obj) (and (complex? obj) (inexact? (real-part obj)) (inexact? (imag-part obj))))
```

and Kaappi's `%inexact-complex-checker` in `lib/srfi/231/storage-classes.sld` is that line verbatim. The verdicts still differ on every **real flonum**: Gambit returns an exact `0` from `(imag-part 1.0)`, so the reference rejects `1.0`; Kaappi's `imag-part` returns `0.0` for a flonum, so the same checker accepts it.

Kaappi's `imag-part` is not wrong by R7RS — 6.2.6 only requires `(zero? (imag-part z))` for a real, not exactness (R6RS did require the exact zero, which is what Gambit and chibi do). But this SRFI's documented rule is "when prose and reference code disagree, trust the code", and the prose ("complex numbers with 32-bit floating-point numbers as real and imaginary parts") is the ambiguous side here. The observable divergence: a program that copies real flonum data into a `c64`/`c128` array succeeds on Kaappi and errors on the reference (and vice versa: nothing works on Kaappi that fails on the reference in the other direction, this is pure leniency).

**Minimal reproduction**

```scheme
(import (scheme base) (scheme write) (scheme complex) (srfi 231))
(define (show . xs) (for-each (lambda (x) (write x) (display " ")) xs) (newline))
(show 'checker-real-flonum ((storage-class-checker c64-storage-class) 1.0))
(show 'checker-complex ((storage-class-checker c64-storage-class) 1.0+2.0i))
(show 'imag-part-of-1.0 (imag-part 1.0) (exact? (imag-part 1.0)))
(show 'copy-reals-into-c64
      (guard (e (#t 'ERROR))
        (array->list (array-copy (make-array (make-interval '#(2)) (lambda (i) 1.5)) c64-storage-class))))
```

**Output**

| | Kaappi (and chibi) | Gambit (reference) |
|---|---|---|
| `checker-real-flonum` | `#t` | `#f` |
| `checker-complex` | `#t` | `#t` |
| `imag-part-of-1.0` | `0.0 #f` (chibi: `0 #t`) | `0 #t` |
| `copy-reals-into-c64` | `(1.5+0.0i 1.5+0.0i)` | `ERROR` |

**Fix**

The checker should not depend on `imag-part`'s exactness convention. `(and (complex? x) (not (real? x)) (inexact? (real-part x)) (inexact? (imag-part x)))` reproduces the reference's verdicts exactly under both conventions — `(real? 1.0+0.0i)` is `#f` in Kaappi as in Gambit, so an explicit inexact-zero imaginary part is still accepted, and a bare real is rejected as in the reference. Plus a test in `tests/scheme/srfi/srfi231-storage-classes.scm` for a real flonum, an exact complex, and `1.0+0.0i`. The official suite does not catch this because its c64/c128 fixtures are all proper complex values.

baijum:
Closes #2542

## What

`%inexact-complex-checker` in `lib/srfi/231/storage-classes.sld` gains an
explicit `(not (real? x))`:

```scheme
(define (%inexact-complex-checker x)
  (and (complex? x) (not (real? x))
       (inexact? (real-part x)) (inexact? (imag-part x))))
```

so `((storage-class-checker c64-storage-class) 1.0)` is now `#f`, matching
the reference implementation, while `1.0+0.0i` (explicit inexact zero
imaginary part) is still `#t`.

## Why

The reference's checker is textually `(and (complex? obj) (inexact?
(real-part obj)) (inexact? (imag-part obj)))`, and Kaappi's was that line
verbatim — yet the verdicts differed on every real flonum: Gambit's
`(imag-part 1.0)` is an exact `0`, so the reference rejects `1.0`, while
Kaappi's `imag-part` returns `0.0` (also R7RS-legal; 6.2.6 requires only
`zero?`, not exactness) and the same line accepted it. A program copying
real flonum data into a `c64`/`c128` array succeeded on Kaappi and errored
on the reference. The official suite never caught this because its
c64/c128 fixtures are all proper complex values.

## Verification

- The issue's reproduction script now matches the reference's column on
  every line (bare real rejected, exact complex rejected, `1.0+0.0i` and
  `1.0+2.0i` accepted, `array-copy` of real flonums errors).
- `tools/srfi231_diff.py storage --seed 1 --count 50` against Gambit
  4.9.8: 6 mismatches (seeds 9, 11, 12, 27, 40, 50, all this divergence)
  with the pre-fix binary → **0 mismatches** with this branch. The tool
  previously reported seeds 3/12/27/40 on a slightly earlier generator;
  same mechanism.
- `tests/scheme/srfi/srfi231-*.scm`: all 7 files pass (747 assertions),
  including the updated checker block — the bare-real case now asserts
  rejection, plus new `c128` bare-real and `1.0+0.0i` acceptance cases.
- Official suite `srfi231-official.scm`: 10,936 passed, 0 unexpected
  failures.

### DIFF
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -923,7 +923,14 @@ from a functional accumulator — the official suite's own continuation cases
 drive two continuations, each invoked twice — and `tools/srfi231_diff.py`
 (`docs/dev/testing.md`) now generates random view/reshape programs and diffs
 Kaappi against the reference under Gambit, for the properties fixed cases
-cannot see. `array-copy` of a non-specialized
+cannot see. The tester's first find (kaappi#2542) is the imag-part
+convention leaking through the reference's own code: its c64/c128 checker
+is textually `(and (complex? obj) (inexact? (real-part obj)) (inexact?
+(imag-part obj)))`, but Gambit's exact-0 `(imag-part 1.0)` makes it reject
+a bare real flonum that Kaappi's equally R7RS-legal inexact `0.0` accepted
+— so the shipped checker carries an explicit `(not (real? x))`, encoding
+the reference's verdict under either convention (an `1.0+0.0i` with a real
+inexact-zero imaginary part is still accepted). `array-copy` of a non-specialized
 source is therefore the reference's exact shape (reversed list, then a body
 filled by linear position). A specialized source takes the direct fill, as in
 the reference's `%!array-copy` — a deliberate, documented exception: a
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -935,8 +935,8 @@ cannot reach: a mixed-exactness complex (`1+2.0i`,
 `(make-rectangular 1 2.0)`) keeps an exact real part under Gambit, so the
 reference rejects it, while Kaappi's complex representation makes
 exactness contagious — both parts are inexact before any checker runs —
-and accepts it (chibi's `real-part` stays exact like Gambit's, so Kaappi
-is the only lenient side). R7RS-legal on both; if the tester's value pool
+and accepts it (chibi accepts it as well, through a checker that is not
+the reference's). R7RS-legal on both; if the tester's value pool
 ever grows a mixed-exactness literal, the Kaappi-accepts/Gambit-rejects
 verdict it will surface is this known residual, not a new bug.
 `array-copy` of a non-specialized
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -929,8 +929,17 @@ is textually `(and (complex? obj) (inexact? (real-part obj)) (inexact?
 (imag-part obj)))`, but Gambit's exact-0 `(imag-part 1.0)` makes it reject
 a bare real flonum that Kaappi's equally R7RS-legal inexact `0.0` accepted
 — so the shipped checker carries an explicit `(not (real? x))`, encoding
-the reference's verdict under either convention (an `1.0+0.0i` with a real
-inexact-zero imaginary part is still accepted). `array-copy` of a non-specialized
+the reference's verdict under either convention (a `1.0+0.0i` with a real
+inexact-zero imaginary part is still accepted). A residual the checker
+cannot reach: a mixed-exactness complex (`1+2.0i`,
+`(make-rectangular 1 2.0)`) keeps an exact real part under Gambit, so the
+reference rejects it, while Kaappi's complex representation makes
+exactness contagious — both parts are inexact before any checker runs —
+and accepts it (chibi's `real-part` stays exact like Gambit's, so Kaappi
+is the only lenient side). R7RS-legal on both; if the tester's value pool
+ever grows a mixed-exactness literal, the Kaappi-accepts/Gambit-rejects
+verdict it will surface is this known residual, not a new bug.
+`array-copy` of a non-specialized
 source is therefore the reference's exact shape (reversed list, then a body
 filled by linear position). A specialized source takes the direct fill, as in
 the reference's `%!array-copy` — a deliberate, documented exception: a
--- a/lib/srfi/231/storage-classes.sld
+++ b/lib/srfi/231/storage-classes.sld
@@ -129,13 +129,22 @@
 
     ;; fX/cX checkers match the reference exactly: f32/f64 accept only
     ;; inexact reals (flonum?, generic-arrays.scm's f32/f64 checker) and
-    ;; c64/c128 only complexes whose real and imaginary parts are both
-    ;; inexact. Accepting exact values silently coerces them (1/3 stored
-    ;; into c64 narrows to f32 precision), diverging from the reference's
-    ;; "value cannot be stored in body" error (#2355).
+    ;; c64/c128 only proper complexes whose real and imaginary parts are
+    ;; both inexact. Accepting exact values silently coerces them (1/3
+    ;; stored into c64 narrows to f32 precision), diverging from the
+    ;; reference's "value cannot be stored in body" error (#2355). The
+    ;; reference's checker is (and (complex? obj) (inexact? (real-part
+    ;; obj)) (inexact? (imag-part obj))), but under Gambit's exact-0
+    ;; (imag-part 1.0) that rejects a bare real flonum, while Kaappi's
+    ;; inexact 0.0 (also R7RS-legal, 6.2.6 mandates only zero?, not
+    ;; exactness) accepted it -- so (not (real? x)) encodes the
+    ;; reference's verdict independently of the host's imag-part
+    ;; convention; 1.0+0.0i with an explicit inexact zero imaginary part
+    ;; is still accepted (#2542).
     (define (%flonum-checker x) (and (real? x) (inexact? x)))
     (define (%inexact-complex-checker x)
-      (and (complex? x) (inexact? (real-part x)) (inexact? (imag-part x))))
+      (and (complex? x) (not (real? x))
+           (inexact? (real-part x)) (inexact? (imag-part x))))
 
     (define f32-storage-class
       (make-storage-class f32vector-ref f32vector-set! %flonum-checker
--- a/tests/scheme/srfi/srfi231-storage-classes.scm
+++ b/tests/scheme/srfi/srfi231-storage-classes.scm
@@ -1,11 +1,12 @@
-;; SRFI 231 (Intervals and Generalized Arrays) tests -- phase 1b: storage
-;; classes. Arrays themselves are a later phase of this multi-slice SRFI,
-;; tracked under issue #1694; (srfi 231) itself is not yet an importable
-;; bare library.
+;; SRFI 231 (Intervals and Generalized Arrays) tests -- storage classes.
+;; The intervals/arrays/views imports exist only for the end-to-end
+;; array-copy checker assertions below; the storage-class surface itself
+;; needs none of them.
 ;; Run directly: zig-out/bin/kaappi tests/scheme/srfi/srfi231-storage-classes.scm
 
 (import (scheme base) (scheme inexact) (scheme process-context) (srfi 64)
         (srfi 160 u16) (srfi 160 f32) (srfi 160 f64) (srfi 160 c64)
+        (srfi 231 intervals) (srfi 231 arrays) (srfi 231 views)
         (srfi 231 storage-classes))
 
 (test-begin "srfi-231-storage-classes")
@@ -223,6 +224,17 @@
 (test-equal #f ((storage-class-checker c64-storage-class) (make-rectangular 3 4)))
 (test-equal #f ((storage-class-checker c64-storage-class) 1/3))
 (test-equal #f ((storage-class-checker c128-storage-class) (make-rectangular 1/2 2)))
+;; the user-visible symptom of #2542 was array-copy of a real-flonum
+;; array SUCCEEDING -- the generic->typed copy branch is the only copy
+;; path that consults the checker at all (same-class copies skip it,
+;; #2448), so pin it end to end
+(test-assert "array-copy of real flonums into c64 is rejected (#2542)"
+             (guard (e (#t #t))
+               (array-copy (make-array (make-interval '#(2)) (lambda (i) 1.5)) c64-storage-class)
+               #f))
+(test-assert "array-copy of 1.0+0.0i into c64 is accepted (#2542)"
+             (guard (e (#t #f))
+               (array-ref (array-copy (make-array (make-interval '#(2)) (lambda (i) 1.5+0.0i)) c64-storage-class) 0)))
 
 ;;; --- c64/c128 use the reference's interleaved-float representation
 ;;; (#2382): the body is an f32/f64vector of twice the logical length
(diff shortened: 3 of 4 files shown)


docs/dev/srfi-implementation-notes.md
```diff
@@ -923,7 +923,14 @@ from a functional accumulator — the official suite's own continuation cases
...
 (`docs/dev/testing.md`) now generates random view/reshape programs and diffs
 Kaappi against the reference under Gambit, for the properties fixed cases
-cannot see. `array-copy` of a non-specialized
+cannot see. The tester's first find (kaappi#2542) is the imag-part
+convention leaking through the reference's own code: its c64/c128 checker
+is textually `(and (complex? obj) (inexact? (real-part obj)) (inexact?
+(imag-part obj)))`, but Gambit's exact-0 `(imag-part 1.0)` makes it reject
+a bare real flonum that Kaappi's equally R7RS-legal inexact `0.0` accepted
+— so the shipped checker carries an explicit `(not (real? x))`, encoding
+the reference's verdict under either convention (an `1.0+0.0i` with a real
```
baijum: Nit: "an `1.0+0.0i`" → "a `1.0+0.0i`". This is also the natural place for the mixed-exactness residual from the top-level comment: `1+2.0i` / `(make-rectangular 1 2.0)` are rejected by the reference (exact real part under Gambit) and accepted here because Kaappi coerces both parts inexact before the checker runs.

tests/scheme/srfi/srfi231-storage-classes.scm
```diff
@@ -208,16 +208,18 @@
...
 (test-equal #t ((storage-class-checker f64-storage-class) 3.5))
 (test-equal #t ((storage-class-checker f64-storage-class) -0.0))
 (test-equal #f ((storage-class-checker f64-storage-class) 3))
 (test-equal #f ((storage-class-checker f64-storage-class) 1/3))
 (test-equal #f ((storage-class-checker f32-storage-class) 42))
 (test-equal #t ((storage-class-checker c64-storage-class) (make-rectangular 3.5 -1.0)))
-(test-equal #t ((storage-class-checker c64-storage-class) 3.5))     ; inexact real: imag-part is 0.0
+(test-equal #t ((storage-class-checker c64-storage-class) 1.0+0.0i)) ; explicit inexact zero imag
+(test-equal #f ((storage-class-checker c64-storage-class) 3.5))     ; bare real: reference rejects (#2542)
+(test-equal #f ((storage-class-checker c128-storage-class) 1.0))    ; same, under the other convention too
```
baijum, line 223: Optional: the issue's user-visible symptom was `array-copy` of a real-flonum array *succeeding*, and these assertions test the checker in isolation. One end-to-end line would pin the path that actually reached users — it goes through `%copy-value-checker`'s generic→typed branch, which is the only copy path that consults the checker at all (same-class copies skip it, #2448):

```scheme
(test-assert "array-copy of real flonums into c64 is rejected (#2542)"
             (guard (e (#t #t))
               (array-copy (make-array (make-interval '#(2)) (lambda (i) 1.5)) c64-storage-class)
               #f))
```

lib/srfi/231/storage-classes.sld
```diff
@@ -129,13 +129,22 @@
...
+    ;; (imag-part 1.0) that rejects a bare real flonum, while Kaappi's
+    ;; inexact 0.0 (also R7RS-legal, 6.2.6 mandates only zero?, not
+    ;; exactness) accepted it -- so (not (real? x)) encodes the
+    ;; reference's verdict independently of the host's imag-part
+    ;; convention; 1.0+0.0i with an explicit inexact zero imaginary part
+    ;; is still accepted (#2542).
     (define (%flonum-checker x) (and (real? x) (inexact? x)))
     (define (%inexact-complex-checker x)
-      (and (complex? x) (inexact? (real-part x)) (inexact? (imag-part x))))
+      (and (complex? x) (not (real? x))
```
baijum, line 146: Confirmed equivalent to the reference's verdict under both `imag-part` conventions on every probe I tried (default, `make-rectangular` with `0.0` vs `0`, nan/inf in either part, set/copy paths) — 150 complex-class cases of the differential tester against Gambit, 0 mismatches. The one input class where the two still disagree is mixed exactness (`1+2.0i`), and that is decided by the reader/representation before this line runs, so nothing to change here.

baijum (commented):
Verified independently against the reference, not just the PR's own numbers.

- **Differential tester, complex classes only, PR library**: `tools/srfi231_diff.py storage --classes c64,c128 --seed 1 --count 150` against Gambit 4.9.8 → **0 mismatches**. The same range against main mismatches on every case (120/120), so the fix closes the whole family, not the six seeds the description cites.
- **Line-by-line probes vs Gambit** (PR library on the left, Gambit on the right): default value `(c 0 0)` accepted on both; `make-specialized-array` with no initial value works on both (the default still passes its own checker, which was the one way this change could have broken every c64 array); `(make-rectangular 3.5 0.0)` accepted on both, `(make-rectangular 3.5 0)` rejected on both; `+nan.0` rejected on both, `(make-rectangular +nan.0 1.0)` and `(make-rectangular 1.0 +inf.0)` accepted on both; `array-set!` of a real and `array-copy` of a real-flonum array error on both; a generic→c128 copy of proper complexes matches.
- The PR's `srfi231-storage-classes.scm` passes against the PR library (251 assertions); main's copy of that file fails exactly the flipped `3.5` assertion against it, so the regression test does fail without the fix.

Two things to take before merge, one residual to record:

1. **Stale text in `tools/srfi231_diff.py` (docstring, lines 57–61).** It says "Kaappi's own open divergence kaappi#2542 … use `--classes` to exclude them until it is fixed". This PR closes #2542, so that paragraph should go — or turn into a past-tense sentence saying the complex classes are back in the default draw — otherwise the next person running the tool excludes two classes for nothing. `docs/dev/testing.md:395` ("first real finding is kaappi#2542") reads fine as history.

2. Inline: a grammar nit in the notes, and an optional end-to-end assertion.

3. **Residual divergence, not fixable in the checker, worth one sentence in the notes.** Mixed-exactness complexes: Gambit keeps `1+2.0i` and `(make-rectangular 1 2.0)` with an *exact* real part, so the reference rejects them; Kaappi's complex representation makes exactness contagious (so does chibi's: `(exact? (real-part 1+2.0i))` is `#t` in both Gambit and chibi, `#f` in Kaappi), so the checker sees `1.0+2.0i` and accepts. R7RS-legal on both sides, and the value has already been coerced by the time any checker runs. The tester's value pool has no mixed-exactness literal yet; when it gets one this will surface, and a sentence in the notes (next to the `(not (real? x))` explanation) keeps it from being filed as a new bug.

Official suite against the PR library is running here as well; I'll add a line if it disagrees with the description's 10,936/0.

docs/dev/srfi-implementation-notes.md
baijum: Fixed — `a `1.0+0.0i``. The paragraph now also records the mixed-exactness residual: `1+2.0i` / `(make-rectangular 1 2.0)` keep an exact real part under Gambit and are rejected by the reference, while Kaappi's complex representation makes exactness contagious (both parts inexact before any checker runs), and chibi's `real-part` stays exact like Gambit's — so the notes name the Kaappi-accepts/Gambit-rejects verdict on a future mixed-exactness tester literal as the known residual, not a new bug.

tests/scheme/srfi/srfi231-storage-classes.scm
baijum: Added, both directions: the rejection line as given, plus an acceptance line for `1.0+0.0i` through the same path, so simplifying the checker away from `(not (real? x))` in either direction trips one of them. One discovery while doing it: `array-copy` is exported by `(srfi 231 views)`, not `(srfi 231 arrays)`, so the file imports intervals/arrays/views — the header's stale "arrays are a later phase / (srfi 231) not yet importable" claim went with it.

lib/srfi/231/storage-classes.sld
baijum: Agreed — left untouched.

baijum:
Addressed in ddd80d54:

1. **Diff-tool docstring** — the "use `--classes` to exclude them until it is fixed" paragraph is now past tense: #2542 used to surface as `check` mismatches on c64/c128, the workaround was excluding them, and both classes are back in the default draw; `--classes` stays as the way to focus a run (`testing.md:395` left as history, per the review).
2. **End-to-end assertions** — the generic→typed `array-copy` branch is now pinned both ways (real flonums rejected, `1.0+0.0i` accepted); replies inline.
3. **Mixed-exactness residual** — recorded next to the `(not (real? x))` explanation in the notes, so the first tester hit on a `1+2.0i`-shaped value reads as the known residual.

Re-verified after the changes: `srfi231-storage-classes.scm` 253/253 (was 251 — the two new assertions), all other SRFI 231 files unchanged and passing, and the edited tool still runs (`storage --seed 1 --count 2`: 0 mismatches).

docs/dev/srfi-implementation-notes.md
```diff
@@ -923,7 +923,23 @@ from a functional accumulator — the official suite's own continuation cases
...
+(imag-part obj)))`, but Gambit's exact-0 `(imag-part 1.0)` makes it reject
+a bare real flonum that Kaappi's equally R7RS-legal inexact `0.0` accepted
+— so the shipped checker carries an explicit `(not (real? x))`, encoding
+the reference's verdict under either convention (a `1.0+0.0i` with a real
+inexact-zero imaginary part is still accepted). A residual the checker
+cannot reach: a mixed-exactness complex (`1+2.0i`,
+`(make-rectangular 1 2.0)`) keeps an exact real part under Gambit, so the
+reference rejects it, while Kaappi's complex representation makes
+exactness contagious — both parts are inexact before any checker runs —
+and accepts it (chibi's `real-part` stays exact like Gambit's, so Kaappi
```
baijum: The premise is right but the conclusion isn't: chibi's `real-part` does stay exact, yet chibi's **checker** accepts `1+2.0i` and `(make-rectangular 1 2.0)` anyway — and a bare `1.0` too — because its port's checker is not the reference's (probed just now: chibi `((lit #t) (rect #t) (real #t))`, Gambit all `#f`, Kaappi on this branch `(#t #t #f)`). So Kaappi is not "the only lenient side"; it is the only side that is lenient *while running the reference's checker logic*. Suggest dropping the parenthetical, or: "(chibi accepts it as well, through a checker that is not the reference's)".

tools/srfi231_diff.py
```diff
@@ -56,10 +56,12 @@
...
-Kaappi's own open divergence kaappi#2542 (c64/c128 checkers accept real
-flonums, the reference rejects them) surfaces as `check` mismatches on those
-two classes; use `--classes` to exclude them until it is fixed, since a
-program's first difference hides everything after it. And chibi 0.12's port raises on a `copy-on-failure? #t`
+kaappi#2542 (the c64/c128 checkers accepted real flonums, which the
+reference rejects) used to surface as `check` mismatches on those two
+classes -- before kaappi#2543 fixed it, the workaround was excluding
+them with `--classes`, and both are back in the default draw. Since a
+program's first difference hides everything after it, `--classes` is
+still the way to focus a run on the classes you care about. And chibi 0.12's port raises on a `copy-on-failure? #t`
```
baijum: Nit: this line grew to 115 columns with the rewrite (it was already the longest in the docstring). Worth rewrapping the paragraph so "And chibi 0.12's port…" starts a fresh line — purely cosmetic, no need for another round-trip if you'd rather leave it.

baijum (commented):
ddd80d54 addresses all three comments; verified against the PR library and against main:

- The updated `srfi231-storage-classes.scm` passes on the PR library (253 assertions). On **main's** library it fails exactly three — the flipped `3.5` verdict, the `c128` bare real, and the new end-to-end `array-copy` rejection — so the regression tests do fail without the fix, and the `1.0+0.0i` acceptance passes on both, as it should.
- The tool docstring now reads as history, `--classes` is presented as a focusing tool, and the file still compiles and `--help`s. Notes lint clean.

One sentence in the notes is wrong and should be trimmed before merge (inline). Everything else is good to go.

docs/dev/srfi-implementation-notes.md
baijum: Good catch — corrected in 2df95915 with your phrasing: "(chibi accepts it as well, through a checker that is not the reference's)". The sentence after it ("the Kaappi-accepts/Gambit-rejects verdict it will surface") is unchanged since the tester's default oracle is Gambit, where that is exactly the shape.

tools/srfi231_diff.py
baijum: Rewrapped in 2df95915 — the paragraph now closes at "…focus a run on the classes you care about." and "And chibi 0.12's port…" starts a fresh line (55 columns); the longest line in my paragraph is back under 80. The chibi paragraph's own pre-existing long lines are left as they were, per "no need for another round-trip".

PR #2544, opened 2026-09-07
Add the call/cc mode to the SRFI 231 differential tester

baijum:
Third mode for `tools/srfi231_diff.py` (#2541, #2543): the spec's promise that every procedure without a trailing `!` is call/cc safe, "does not modify the state of any data captured by a continuation". #2539 showed why a single re-entry cannot check that promise — the official suite's own continuation cases passed over a shared scratch vector — so this mode generalizes the #2539 driver.

**What a case does.** One accumulating or callback-taking procedure per case: `array-copy` (generic and into a typed class), `array->list`/`->vector` and their `*` forms, `array-fold-left`/`-right`, `array-reduce`, `array-every`, `array-any`, `array-for-each`, `array-map` under a copy (one and two sources), `array-stack`/`append`/`block`/`decurry`, and the interval folds and for-each. Two continuations are captured on the first run at random positions of a random 1-D or 2-D domain, then re-invoked 3–5 times with random values in a random order, and the whole result history is compared. The capture sits either in the source array's getter or in the callback itself (fold kernel, predicate, operator), and one case in five is the escape variant: the hook throws to a continuation outside the procedure mid-walk, then the procedure runs again. The driver is one procedure body, so a re-entered continuation resumes inside that frame and never re-executes a top-level form.

**Negative control.** Against the library as it was before #2540 (scratch-vector `array-copy`, `set!`-cell folds) the mode mismatches on 25 of 60 cases, across exactly the families that fix changed: copy, the conversions, both folds, reduce, map under copy, and all four assembly procedures that delegate to `array-copy`.

| oracle | cases | mismatches |
|---|---|---|
| Gambit 4.9.8 (reference) | 400 (seeds 5000..5399) | 0 |
| chibi 0.12 | 100 (seeds 3000..3099) | 21, all chibi's own; Gambit sides with Kaappi on every one of the 21 programs |

**Two generator bugs found by the oracles and fixed in the second commit**, both worth knowing for future modes: `array-block` needs an outer array of the pieces' rank (both implementations rejected my 1-D outer for 2-D pieces; the only "difference" was Gambit printing the error on stdout), and an accumulator written as `(set! acc (cons (h x) acc))` reads `acc` before or after the capture depending on the implementation's argument evaluation order, which R7RS leaves unspecified — chibi's is right-to-left — so the hook's value is now bound first.

**A chibi finding.** Its `array-copy` (into a typed storage class, of an `array-map` result, and under stack/append/block) keeps a value written by an earlier re-entry across a second continuation: the #2539 shape, a shared scratch behind a call/cc-safe procedure. Its `array-fold-right` and `interval-fold-right` accumulate through a `set!` cell, so a re-entry conses onto the whole first run's list — e.g. `(0 1 2 30 1 2 3 4 5)` where Gambit and Kaappi give `(0 1 2 30 4 5)`. Both are the shapes #2540 removed from Kaappi. Recorded in the docstring's divergence list.

### DIFF
--- a/docs/dev/testing.md
+++ b/docs/dev/testing.md
@@ -391,7 +391,7 @@ under a copy, and stack/append/block/decurry) is driven by the kaappi#2539
 schedule generalized — two continuations captured on the first run at random
 positions, re-invoked with random values in a random order, the whole result
 history printed — with the capture placed either in the source's getter or
-in the callback itself, plus an escape-out variant. Against the pre-#2539
+in the callback itself, plus an escape-out variant. Against the pre-#2540
 library it mismatches on the accumulating families; against main it is
 clean. New modes are one generator function each in `MODES`.
 
--- a/docs/dev/testing.md
+++ b/docs/dev/testing.md
@@ -382,8 +382,18 @@ wrong-typed values, `make-specialized-array` with an initial value,
 into a view — with every value canonicalized before printing (finite reals
 as exact rationals, complex as pairs of those, chars as code points), so
 f16/f32 rounding is compared bit-exactly and Gambit's `.5` never differs
-textually from Kaappi's `0.5`; `--classes u1,f16` narrows the draw. New
-modes are one generator function each in `MODES`.
+textually from Kaappi's `0.5`; `--classes u1,f16` narrows the draw. The
+`callcc` mode checks the spec's promise that every procedure without a
+trailing `!` is call/cc safe: one accumulating or callback-taking procedure
+per case (`array-copy`, the list/vector conversions, the array and interval
+folds, `array-reduce`, `array-every`/`-any`, `array-for-each`, `array-map`
+under a copy, and stack/append/block/decurry) is driven by the kaappi#2539
+schedule generalized — two continuations captured on the first run at random
+positions, re-invoked with random values in a random order, the whole result
+history printed — with the capture placed either in the source's getter or
+in the callback itself, plus an escape-out variant. Against the pre-#2539
+library it mismatches on the accumulating families; against main it is
+clean. New modes are one generator function each in `MODES`.
 
 The known oracle divergences are recorded in the tool's docstring: Gambit's
 bundled reference flips `specialized-array-default-safe?` to `#t` (the spec
--- a/tools/srfi231_diff.py
+++ b/tools/srfi231_diff.py
@@ -554,12 +554,19 @@ CALLCC_PRELUDE = """\
     (let ((r (collect hook)))
       (set! first-run? #f)
       (set! results (cons r results)))
-    (if (< i (length schedule))
-        (let ((step (list-ref schedule i)))
-          (set! i (+ i 1))
-          (if (= (car step) 1)
-              (if cont1 (cont1 (cdr step)) 'never-captured)
-              (if cont2 (cont2 (cdr step)) 'never-captured))))
+    ;; a step whose continuation was never captured (the walk stopped
+    ;; before its position: array-any past its short-circuit, a reduce
+    ;; seeded from element 0) is skipped, never allowed to end the
+    ;; schedule and starve the steps after it
+    (let loop ()
+      (if (< i (length schedule))
+          (let ((step (list-ref schedule i)))
+            (set! i (+ i 1))
+            (let ((k (if (= (car step) 1) cont1 cont2)))
+              (if k (k (cdr step)) (loop))))))
+    ;; which positions were reached is itself a differential signal: a
+    ;; differing short-circuit point or fold seeding shows up here
+    (show 'captured (and cont1 #t) (and cont2 #t))
     (reverse results)))
 ;; escape variant: the hook throws to a continuation OUTSIDE the
 ;; procedure under test at position p1, then the same procedure is run
@@ -577,7 +584,7 @@ CALLCC_PRELUDE = """\
 # puts the capture in the callback, one that only uses {A} relies on the
 # getter capture. {IV} is the domain, {L} a lambda mapping a multi-index
 # to its linear position (so a capture position means the same thing in
-# every dimension), {N} the volume.
+# every dimension).
 CALLCC_FAMILIES = [
     # capture in the source array's getter
     ("copy", "(array->list* (array-copy {A}))"),
@@ -598,6 +605,11 @@ CALLCC_FAMILIES = [
     # so the hook's value is always bound first
     ("for-each", "(let ((acc '())) (array-for-each (lambda (x) (set! acc (cons x acc))) {A}) (reverse acc))"),
     ("map-copy", "(array->list* (array-copy (array-map (lambda (x) (* 2 x)) {A})))"),
+    # the same source twice: the hook fires twice per position and the later
+    # capture at p1 overwrites the earlier, which is benign only because both
+    # calls carry the same position and + commutes. An implementation that
+    # read a duplicated source once per position would show a re-entry as
+    # (+ v v) rather than (+ v p1) -- that is what a mismatch here means.
     ("map2-copy", "(array->list* (array-copy (array-map + {A} {A})))"),
     ("stack", "(array->list* (array-stack 0 (list {A} (make-array {IV} (lambda idx 99)))))"),
     ("append", "(array->list* (array-append 0 (list {A} (make-array {IV} (lambda idx 99)))))"),
@@ -640,7 +652,7 @@ def gen_callcc(seed, classes=None):
     # the plain source for callback-capturing families
     P = f"(make-array {iv} (lambda idx ({lin} idx)))"
     outer = f"(make-interval {fmt_vec([2] + [1] * (len(shape) - 1))})"
-    body = tmpl.format(A=A, P=P, H="h", IV=iv, L=lin, N=n, OUTER=outer)
+    body = tmpl.format(A=A, P=P, H="h", IV=iv, L=lin, OUTER=outer)
     L = [CALLCC_PRELUDE, f"(define collect (lambda (h) {body}))"]
     L.append(f"(show 'family '{name} 'shape '({' '.join(map(str, shape))}) 'p1 {p1} 'p2 {p2})")
     if rng.random() < 0.2:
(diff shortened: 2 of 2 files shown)


tools/srfi231_diff.py
```diff
@@ -502,7 +526,136 @@ def gen_storage(seed, classes=None):
...
+                              ((= x p2) (set! cont2 c))))
+                    x)))))
+    (let ((r (collect hook)))
+      (set! first-run? #f)
+      (set! results (cons r results)))
+    (if (< i (length schedule))
+        (let ((step (list-ref schedule i)))
+          (set! i (+ i 1))
+          (if (= (car step) 1)
+              (if cont1 (cont1 (cdr step)) 'never-captured)
```
baijum: When a step's target continuation was never captured, the driver returns `'never-captured` from the inner `if` and falls straight through to `(reverse results)` — the **rest of the schedule is silently dropped**, including later steps whose own target *was* captured.

This is reachable today: `kernel-reduce` with `p1` = 0 (the reference seeds `array-reduce` from element 0 via the sentinel, so the operator — and the hook — first runs at element 1), and the `any` families past their short-circuit point. Concretely, `callcc --seed 42` prints a three-step schedule but `results` has a single entry — the whole case degenerates to comparing one plain run. Deterministic across implementations, so no false positives, but a few percent of cases quietly exercise no re-entry at all, and the only way to tell is cross-referencing the schedule line against the results count.

Two possible small improvements (either is fine):
- print which continuations were captured — `(show 'captured (and cont1 #t) (and cont2 #t))` — so a degenerate case is visible in the output; this is also a differential signal in its own right, since a differing short-circuit point or fold seeding would show up directly;
- or skip to the next step instead of returning, so a `(2 v)` step can't starve later `(1 v)` re-entries.

baijum (commented):
Reviewed by reproducing the reported runs on this branch (after a `zig build`): `callcc --seed 5000 --count 400` against Gambit 4.9.8 — **0 mismatches**, and `--seed 3000 --count 100 --oracle chibi` — **21 mismatches, exactly the reported count**, every one in the families the docstring's divergence note names (copy, copy-typed, map/map2/kernel-map-copy, append, block, fold-right, interval-fold-right), with Gambit agreeing with Kaappi on all 21. The negative control also held here by accident: my first run used a pre-#2540 `zig-out` and Gambit disagreed with Kaappi on every case.

The comments below are all minor — nothing blocking.

tools/srfi231_diff.py
```diff
@@ -502,7 +526,136 @@ def gen_storage(seed, classes=None):
...
+           (collect (lambda (x) (if (= x p1) (k (list 'escaped x)) x)))))
+        (collect (lambda (x) x))))
+"""
+
+# Each family is (name, template). In the template, {A} is the source
+# array expression and {H} the hook; a template that uses {H} directly
+# puts the capture in the callback, one that only uses {A} relies on the
+# getter capture. {IV} is the domain, {L} a lambda mapping a multi-index
+# to its linear position (so a capture position means the same thing in
+# every dimension), {N} the volume.
```
baijum: `{N}` is documented here ("{N} the volume") and passed in the `tmpl.format(...)` call at line 643, but no template in `CALLCC_FAMILIES` actually uses it. Drop it from the comment — or from the format call too — before it sends the next reader hunting for a volume placeholder that doesn't exist.

tools/srfi231_diff.py
```diff
@@ -502,7 +526,136 @@ def gen_storage(seed, classes=None):
...
+    ("every", "(array-every list {A})"),
+    ("any", "(array-any (lambda (x) (and (>= x 2) (list x))) {A})"),
+    # the for-each families accumulate through a set! cell on the CALLER's
+    # side, which is legitimate -- but (set! acc (cons (h x) acc)) would read
+    # acc before or after the capture depending on the implementation's
+    # argument evaluation order (unspecified by R7RS; chibi is right-to-left),
+    # so the hook's value is always bound first
+    ("for-each", "(let ((acc '())) (array-for-each (lambda (x) (set! acc (cons x acc))) {A}) (reverse acc))"),
+    ("map-copy", "(array->list* (array-copy (array-map (lambda (x) (* 2 x)) {A})))"),
+    ("map2-copy", "(array->list* (array-copy (array-map + {A} {A})))"),
```
baijum, line 613: Passing `{A}` as both sources means the hook fires twice per position and the later capture at `p1` overwrites the earlier one. That's benign as generated — both Gambit's reference and Kaappi read each source once per element, the two calls carry the same position value, and `+` commutes, so which call site the surviving continuation belongs to doesn't change the result. But it's worth one sentence in the family comment: if an implementation ever read a duplicated source once per position instead of twice, a re-entry value would land as `(+ v v)` rather than `(+ v p1)`, and a future mismatch of exactly that shape will be much easier to diagnose if the double-call subtlety is written down here.

docs/dev/testing.md
```diff
@@ -382,8 +382,18 @@ wrong-typed values, `make-specialized-array` with an initial value,
...
+textually from Kaappi's `0.5`; `--classes u1,f16` narrows the draw. The
+`callcc` mode checks the spec's promise that every procedure without a
+trailing `!` is call/cc safe: one accumulating or callback-taking procedure
+per case (`array-copy`, the list/vector conversions, the array and interval
+folds, `array-reduce`, `array-every`/`-any`, `array-for-each`, `array-map`
+under a copy, and stack/append/block/decurry) is driven by the kaappi#2539
+schedule generalized — two continuations captured on the first run at random
+positions, re-invoked with random values in a random order, the whole result
+history printed — with the capture placed either in the source's getter or
+in the callback itself, plus an escape-out variant. Against the pre-#2539
```
baijum: "Against the pre-#2539 library" — #2539 is the *issue* (the bug report), not the change, so "pre-#2539" doesn't name the library state it means. The library changed in #2540, and both the tool's docstring ("Kaappi and Kaappi (since #2540)") and the PR description ("before #2540") use the fix's number. Suggest "pre-#2540" (or "the library with the #2539 bug").

tools/srfi231_diff.py
baijum: Took both, in 2d845139: a never-captured step is skipped and the schedule continues (a `let loop` around the dispatch, so a `(2 v)` step can no longer starve later `(1 v)` ones), and the driver prints `(show 'captured (and cont1 #t) (and cont2 #t))` before returning. Seed 42 now shows `captured #f #t` with its single result, so the degeneracy is visible in the output rather than by cross-referencing the schedule. Over 100 seeds: 72 capture both, 6 capture one, 2 capture neither. 40 cases against Gambit stay clean.

tools/srfi231_diff.py
baijum: Dropped in 2d845139 — from the comment and from the format call.

tools/srfi231_diff.py
baijum: Written down in 2d845139, on the `map2-copy` entry: benign only because both calls carry the same position and `+` commutes, and a mismatch there would mean an implementation read a duplicated source once per position, showing a re-entry as `(+ v v)` rather than `(+ v p1)`.

docs/dev/testing.md
baijum: Fixed in 2d845139: "pre-#2540".

tools/srfi231_diff.py
```diff
@@ -502,7 +526,148 @@ def gen_storage(seed, classes=None):
...
+    ;; a step whose continuation was never captured (the walk stopped
+    ;; before its position: array-any past its short-circuit, a reduce
+    ;; seeded from element 0) is skipped, never allowed to end the
+    ;; schedule and starve the steps after it
+    (let loop ()
+      (if (< i (length schedule))
+          (let ((step (list-ref schedule i)))
+            (set! i (+ i 1))
+            (let ((k (if (= (car step) 1) cont1 cont2)))
+              (if k (k (cdr step)) (loop))))))
```
baijum, line 566: Verified working: seed 92 (`kernel-any`, cont2 never captured, two `(2 …)` steps interleaved between three `(1 …)` steps) produces four results where the previous driver produced two — the skip lets the later re-entries run — and all three implementations agree byte-for-byte.

tools/srfi231_diff.py
```diff
@@ -26,8 +26,25 @@
...
+  callcc   Call/cc safety of every accumulating or callback-taking non-!
+           procedure -- the spec's promise that such procedures "do not
+           modify the state of any data captured by a continuation". One
+           procedure per case (array-copy, array->list/vector and their *
+           forms, the array and interval folds, array-reduce, array-every,
+           array-any, array-for-each, array-map under array-copy, and
+           array-stack/append/block/decurry) is driven by the kaappi#2539
+           schedule generalized: two continuations captured on the first
+           run at random positions, then re-invoked with random values in
+           a random order, the whole result history printed at the end.
```
baijum, line 41: Minor: the mode description still ends at "the whole result history printed at the end" — but every generated program now also prints which of the two continuations were captured, and steps whose continuation was never captured are skipped rather than ending the schedule. Both are part of what gets compared, and the `captured` line is the first place a differing short-circuit point or fold seeding would show, so a clause here (e.g. "…printed at the end, along with which of the two continuations were captured; steps whose continuation was never captured are skipped") would keep the tool's primary documentation matching its output. Purely optional.

baijum (commented):
Reviewed `2d845139` by re-running everything on this head:

- `callcc --seed 5000 --count 400` (Gambit): **0 mismatches**, exit 0 — the new `captured` line agrees everywhere it should.
- `--seed 3000 --count 100 --oracle chibi`: **21 mismatches**, same seeds/families as before the change (copy, copy-typed, map/map2/kernel-map-copy, append, block, fold-right, interval-fold-right), and Gambit still agrees with Kaappi byte-for-byte on all 21 — including the `captured` line, which is expected since those divergences are about accumulator state, not which positions were visited. The new signal correctly does *not* fire on them.
- The skip itself: seed 92 (`kernel-any`, p1=2 captured, p2=3 past the short-circuit, schedule `(1 28) (2 15) (1 28) (2 11) (1 37)`) now yields `captured #t #f` and **four** results — the two `(2 …)` steps are skipped and all three `(1 …)` re-entries run, where the previous driver stopped at step 2 with two results. Kaappi, Gambit, and chibi agree byte-for-byte on the case.
- Seed 42 (`kernel-reduce`, p1=0) is now self-describing: `captured #f #t` tells you why `results` has a single entry without cross-referencing the schedule.

The re-entry control flow also checks out: the named `loop` sits in the `let*` body, so each resumed continuation re-executes it with the already-incremented `i`; the skip branch is a tail call; and `(show 'captured …)` / `(reverse results)` run exactly once because every intermediate pass ends in a continuation invocation that abandons the rest of that pass's body.

One small doc nit inline; otherwise this resolves all four comments cleanly. From my side: ship it.

Issue #1996, opened 2026-08-01
[SRFI-181/192] port-position on a custom port ignores the port's own read-ahead — over-reports after a burst read! and after any peek


baijum:
**Category:** SRFI 181 custom ports × SRFI 192 positioning · **Severity:** wrong-result · **Found by:** Systematic audit v2, Phase 2.3 (#1890)

`portPositionFromCustomPort` (`src/primitives_io.zig:986`) calls `get-position` and returns its answer verbatim. But the port holds its own lookahead — the remainder of the last `read!` burst in `port.read_buf`, plus `peek_byte`/`peek_extra` — which the *source* has already handed over and the *program* has not yet consumed. `get-position` reports the source position; `port-position` must report the port position.

The fd-backed path in the same function gets this right, one branch below:

```zig
const ahead: i64 = @as(i64, @intCast(port.read_buf_len)) +
    @as(i64, if (port.peek_byte != null) 1 else 0) +
    @as(i64, port.peek_extra_len);
const behind: i64 = @intCast(port.write_buf_len - port.write_buf_start);
return types.makeFixnum(os_pos - ahead + behind);
```

The custom-port branch returns before reaching it.

## Symptom 1 — a `read!` that returns more than one element

```scheme
(import (scheme base) (scheme write) (srfi 181) (srfi 192))
(define src (bytevector 10 20 30 40 50 60))
(define pos 0)
(define p (make-custom-binary-input-port "b"
  (lambda (bv start count)                    ;; hands over 3 bytes per call
    (let loop ((i 0))
      (if (or (= i 3) (>= pos (bytevector-length src))) i
          (begin (bytevector-u8-set! bv (+ start i) (bytevector-u8-ref src pos))
                 (set! pos (+ pos 1)) (loop (+ i 1))))))
  (lambda () pos) #f #f))

(read-u8 p)        ;; ⟹ 10   — one byte consumed
(port-position p)  ;; expected: 1   actual: 3
(read-u8 p)        ;; ⟹ 20   — two bytes consumed
(port-position p)  ;; expected: 2   actual: 3
```

## Symptom 2 — after a peek, the spec mandates the cached position

SRFI 181, § *Port positioning and peeking*, verbatim:

[...]

Issue #1743, opened 2026-07-25
kaappi compile: third-party library imports produce a broken binary, with a successful exit code


baijum:
`kaappi compile` reports success for a program importing a third-party library, then the binary dies at runtime with "library not found".

## Environment

- Kaappi Scheme v0.21.0 (build `0687dc4`), target `aarch64-macos-none`
- `kaappi-json` installed with `thottam install kaappi-json`

## Reproduce

```scheme
;; withjson.scm
(import (scheme base) (scheme write) (kaappi json))
(write (json-read-string "{\"a\":1}"))
(newline)
```

Interpreted, it works:

```console
$ kaappi withjson.scm
(("a" . 1))
```

Compiled, the compile step reports success and exits `0`:

```console
$ kaappi compile withjson.scm -o withjson
Wrote /tmp/kaappi_native_20003.ll
Compiled withjson
$ echo $?
0
```

…and then the binary fails:

```console
$ ./withjson
eval error: library not found: (kaappi.json) (CompileError)
$ echo $?
1
```

## Not fixable from the outside

Neither of the obvious escape hatches helps:

```console
$ kaappi --lib-path ~/.kaappi/lib compile withjson.scm -o withjson2   # lib path at compile time
$ ./withjson2
eval error: library not found: (kaappi.json) (CompileError)

$ KAAPPI_HOME=$HOME/.kaappi ./withjson                                # lib path at run time
eval error: library not found: (kaappi.json) (CompileError)
```

Same result with `(kaappi sqlite)`, so it isn't specific to one library.

## What does work

Core Scheme, built-in SRFIs, file I/O, and environment access all compile and run correctly —
verified with a program using `(srfi 1)`, `call-with-output-file`, and
`get-environment-variable`. The limitation is specifically third-party/`.sld` libraries resolved
from the library path.

`zig build -Dbundle-src=app.scm` handles the same program correctly, so there is a working path
for shipping such programs.

## Expected

At minimum, **`kaappi compile` should fail at compile time** with a clear diagnostic rather than
emitting a binary that cannot run. Silently producing a broken executable and exiting `0` is the
worst outcome — it passes CI, ships, and fails on the user's machine.

Ideally the native backend would resolve and link library imports the way the interpreter does;
if that's out of scope for the LLVM backend, then the docs
([deployment.md](https://kaappi-lang.org/guide/deployment/#standalone-binaries), which presents
[...]

Issue #1941, opened 2026-07-31
[SRFI-192] String ports ignore peek read-ahead: port-position over-reports and set-port-position! lands wrong


baijum:
**Category:** SRFI 192 port positioning · **Severity:** wrong-result · **Found by:** systematic audit v2, Phase 2.2 (#1890)

The fd branch subtracts `peek_byte`/`peek_extra`/`read_buf` and clears them on seek. The string-port branch does neither. One root cause, two symptoms.

| | repro | actual | expected |
|---|---|---|---|
| position | `(let ((p (open-input-string "a\rb"))) (read-line p) (port-position p))` | **`3`** | `2` |
| seek | `(let ((p (open-input-string "a\rXbcd"))) (read-line p) (set-port-position! p 0) (read-char p))` | **`#\X`** | `#\a` |

`read-line`'s CR/CRLF handling leaves a pushed-back byte; the string branch counts it as consumed, and a subsequent seek does not clear it.

## Discriminating controls

- **The LF variant** (`"a\nb"`, no pushback) reports `2` — correct. So it is the read-ahead, not `read-line` generally.
- **The byte-identical CR cases on a *file* port** report `2` and `#\a` — correct. That isolates it to the string branch specifically.

## Oracle

`CONFORMANCE.md:109` documents that the fd correction exists because *"otherwise the reported position would drift from what a subsequent read or seek expects"* — while claiming all four SRFI 192 procedures work "for every port kind". The string branch is the counter-example to its own stated rationale.

**Verified:** fresh ReleaseSafe build, isolated `KAAPPI_HOME`; both symptoms and all three controls reproduced.

PR #2546, opened 2026-09-07
Implement SRFI 277: cyclic ports (open-cyclic-input-string / open-cyclic-input-bytevector)

### LINKED ISSUE #2545: [SRFI 277] Implement Cyclic ports (open-cyclic-input-string / open-cyclic-input-bytevector)
## Summary

Implement [SRFI 277: Cyclic ports](https://srfi.schemers.org/srfi-277/srfi-277.html) (Wolfgang Corcoran-Mathe, finalized 2026-09-06).

A cyclic port is an ordinary input port over a string or bytevector that delivers its elements in order, repeatedly, forever: for `x0 … xn` it produces `x0 … xn x0 … xn …` and **never returns an EOF object**. It is the companion SRFI to SRFI 271 (random port libraries), written to close 271's one gap: a reproducible seed that does not depend on the stability of a random-port state's format across runs and implementations. Two other uses the SRFI names: a portable `/dev/zero` (`(open-cyclic-input-bytevector #u8(0))`) and any endless source of a fixed byte/character pattern. Unlike SRFI 158's `circular-generator`, the result is a real port, so every standard port-reading procedure accepts it.

Kaappi has nothing for it yet: no `lib/srfi/277.sld`, no issue, not listed in `docs/dev/srfi-exclusions.md`. It has no `(srfi 277)` sub-libraries and no new syntax — a small, self-contained addition.

## Specification

| Procedure | Returns | Semantics |
|-----------|---------|-----------|
| `(open-cyclic-input-string string)` | textual input port | Delivers the characters of `string` cyclically. **An error is signaled if `string` is empty.** Behavior is undefined if `string` is modified after the call. |
| `(open-cyclic-input-bytevector bytevector)` | binary input port | Delivers the bytes of `bytevector` cyclically. **An error is signaled if `bytevector` is empty.** Behavior is undefined if `bytevector` is modified after the call. |

- Reading from a cyclic port *shall not* produce an end-of-file object.
- Cyclic ports *should* support port positioning per SRFI 192 / R6RS (`port-position`, `set-port-position!`, and the two `port-has-…?` predicates).
- "Undefined behavior" is used in the ISO C sense — an implementation may do anything, including a well-defined extension.

Spec examples:

```scheme
(let ((p (open-cyclic-input-bytevector #u8(1 2 3 4))))
  (read-bytevector 6 p))                  ; ⇒ #u8(1 2 3 4 1 2)

(let ((p (open-cyclic-input-string "frob")))
  (read-string 8 p))                      ; ⇒ "frobfrob"
```

## Why it matters here: SRFI 271 seeding

`(srfi 271 determinized)`'s `make-random-port` already accepts a binary input port and reads exactly 32 seed bytes from it (`seed-from-port` in `lib/srfi/271/determinized.sld`, raising `random-port-initialization-error` on a short read). Today a reproducible port therefore needs a caller-built 32-byte bytevector; the SRFI-intended idiom is

```scheme
(make-random-port (open-cyclic-input-bytevector #u8(1 2 3)))   ; any non-empty, non-all-zero seed
```

and two such ports built from equal cyclic seeds must yield identical streams (271's "identical sequences ⇒ identical initial states" requirement, which is what makes cyclic seeds meaningful). That integration deserves a test of its own.

## Dependencies — all present

| 192 | port positioning (`port-position`, `set-port-position!`, `port-has-*?`) | built-in, `primitives_io.zig` |
| 181 | custom ports (only if the *portable* route below is taken) | portable over `(srfi 181 primitives)` |
| 271 | the consumer this SRFI exists for | portable, `lib/srfi/271*.sld` |

### A. Portable port of the reference over `(srfi 181)` — possible, but carries a positioning trap

The SRFI repo's sample implementation is ~50 lines of R6RS: each constructor wraps `make-custom-binary-input-port` / `make-custom-textual-input-port` with a `read!` that copies `count` elements from the source at `(mod j len)`, a `get-position` returning the source index `pos ∈ [0, len)`, and a `set-position!` that stores it (the bytevector variant rejects `new ≥ len`). Porting that verbatim to `(srfi 181)` **passes the read tests but fails the SRFI's own positioning test on Kaappi**, for a Kaappi-specific reason:

- `readOneByteFromCustomPort` calls `read!` in bursts of `read_chunk_size = 4096` and parks the tail in `read_buf`.
- `portPositionFromCustomPort` then reports `get-position − read_buf_len − peek bytes` (#1996), which assumes `get-position` is a *monotonic source position*.
- The reference's `get-position` is the modular index, so after `(read-bytevector 13 p)` on a 3-byte source Kaappi computes `(4096 mod 3) − (4096 − 13) = −4082`, and `(set-port-position! p -4082)` is rejected as "non-negative exact integer".

If this route is taken, the port's `get-position` must return a monotonic count of elements handed out and `set-position!` must accept any non-negative integer (store `(mod new len)`). A second wrinkle: Kaappi's positions are **byte** offsets (the SRFI 192 note in `docs/dev/srfi-implementation-notes.md`), but a custom *textual* port's `read!` counts characters, so a non-ASCII `open-cyclic-input-string` would still report inconsistent positions through this route. And every 4096-byte burst is a 4096-iteration Scheme loop.

### B. Native, mirroring SRFI 271's shape — recommended

Kaappi's string/bytevector input ports are already exactly a cyclic port minus the wrap: an **owned copy** in `Port.string_data` plus a cursor `string_pos`, and every port primitive (`read-char`, `peek-char`, `read-line`, `read-string`, `read-u8`, `peek-u8`, `read-bytevector`, `read-bytevector!`, `char-ready?`, `u8-ready?`, `read`) funnels through the single byte source `readOneByte`. The string branch there has one EOF exit:

```zig
if (port.string_pos >= data.len) return null;
```

A `cyclic: bool` on `Port` (a plain bool — **no Value field, so `Port.forEachValue` and the GC arms are untouched**) turns that exit into a wrap. Keeping `string_pos` unbounded and indexing `data[string_pos % data.len]` gives SRFI 192 positioning for free: `port-position` stays the existing `string_pos − read-ahead` (monotonic, in bytes, exactly like `open-input-string`), and `set-port-position!`'s string branch only needs to skip its `pos > len` bound for a cyclic port. Everything above the byte source — UTF-8 decoding, `peek_byte`/`peek_extra` lookahead, `read-line`'s CR handling, `close-port` — works unchanged, because the wrap happens at the end of the whole buffer, which is always a codepoint boundary.

Plumbing, copied from 271: two `%`-prefixed constructors (`%open-cyclic-input-string`, `%open-cyclic-input-bytevector`) registered with `primitives.INTERNAL_PUBLIC` next to `open-input-string` / `open-input-bytevector`, and a portable `lib/srfi/277.sld` importing `(kaappi primitives)` that does the argument checks (type, and the spec's **empty-argument error**, raised as an ordinary `error` so `guard` + `error-object?` see it) and exports the two public names. The `.sld` has to exist regardless of route: the build-time `lib/srfi` scan (`build.zig`) is what registers `srfi-277` for `cond-expand`, `(features)`, and `kaappi features`.

Since the port reads its own copy, mutating the source after the call is well-defined here (the port keeps the snapshot) — an allowed implementation-defined extension, worth one sentence in the notes.

## Behaviors to pin in `tests/scheme/srfi/srfi277.scm` (SRFI 64, every assertion named)

- The two spec examples verbatim; reads whose length is not a multiple of the cycle, spanning several cycles; a 1-element cycle (`#u8(0)` as `/dev/zero`).
- Port predicates: `input-port?`, `binary-port?` vs `textual-port?`, `port?`, and `call-with-port` closing it.
- **Never EOF**: `eof-object?` is `#f` after a long read; `char-ready?` / `u8-ready?` are `#t`; `read-bytevector!` fills the whole target.
- Textual details: a multi-byte string (e.g. `"λμ"`) wraps cleanly under `read-char`; `peek-char` then `read-char` across the wrap point; `read-line` on `"ab\n"` returns `"ab"` repeatedly. (`read-line` on a cycle with no newline never returns — spec-conforming, document it, don't test it.)
- Positioning: the SRFI's own re-read test (`read 13`, `port-position`, `read 5`, `set-port-position!`, `read 5` equal), plus `port-has-port-position?` / `port-has-set-port-position!?` both `#t`.
- Errors: `""` and `#u8()` raise a catchable error; a non-string / non-bytevector argument raises; reading after `close-port` errors instead of cycling.
- 271 integration: two `(srfi 271 determinized)` ports seeded from equal short cyclic bytevectors produce identical streams, and a 3-byte cycle satisfies the 32-byte seed read.

Run it with an isolated home so the checkout's `lib/` wins over any `~/.kaappi/lib` (kaappi#2352); `tests/scheme/run-all.sh` already does this.

## Bookkeeping the PR carries

- `docs/dev/srfi-implementation-notes.md`: a `### SRFI 277 — cyclic ports` section (route chosen, the burst/positioning trap above, the snapshot-copy extension), and its header count.
- SRFI counts 180 → 181 and portable 164 → 165 in `CLAUDE.md` (two places), `README.md` (three), `docs/dev/srfi-implementation-notes.md`; a `| 277 | Cyclic ports |` row in `CONFORMANCE.md`. Re-derive with `kaappi features --json` after the build.
- Follow-up in `kaappi.github.io`: `docs/guide/srfi-support.md`.
- `srfi-status-check` needs nothing — the SRFI is already `final`.

Reference implementation and tests: https://github.com/scheme-requests-for-implementation/srfi-277 (MIT).

**Priority rationale:** `priority: low` per `docs/dev/github-issues.md` — a new feature, nothing is wrong today, and the 271 seeding use case already has a workaround (a caller-built 32-byte bytevector).

baijum:
Closes #2545.

## What

SRFI 277 — cyclic ports, finalized 2026-09-06: `(open-cyclic-input-string string)` and `(open-cyclic-input-bytevector bytevector)` return ordinary input ports whose stream repeats forever and never yields an EOF object.

```scheme
(let ((p (open-cyclic-input-bytevector #u8(1 2 3 4))))
  (read-bytevector 6 p))                  ; ⇒ #u8(1 2 3 4 1 2)

(make-random-port (open-cyclic-input-bytevector #u8(1 2 3)))  ; reproducible 271 seed
```

## How

The native route from the issue, mirroring SRFI 271's shape:

- **`Port.cyclic: bool`** (plain bool — no Value field, so the GC walkers are untouched; the `tests_gc_tracing` inventory is re-pinned). A new `allocCyclicInputPort` snapshots the source like `allocStringInputPort` always has.
- **`readOneByte`**'s string branch indexes `data[string_pos % data.len]` for a cyclic port and lets the cursor keep counting past the end — that unbounded count is what makes SRFI 192 positioning come free: `port-position` is unchanged (monotonic `string_pos − read-ahead`), and `set-port-position!` merely skips its `pos > len` bound. `char-ready?`/`u8-ready?` needed nothing (string ports are always ready).
- **`read`** routes cyclic string ports through the incremental path instead of the string fast path: the fast path slices `data[string_pos..]` (out of bounds past the wrap) and reports EOF for a datum unterminated at the snapshot's end, which a cyclic port must never do. The incremental path parses until a datum completes, byte-at-a-time through `readOneByte`.
- **Two `%`-prefixed constructors** registered `INTERNAL_PUBLIC` next to `open-input-string`/`open-input-bytevector`, and a portable **`lib/srfi/277.sld`** that does the type checks and the spec's empty-argument error (ordinary `error`, catchable with `guard` + `error-object?`) and exports the two public names. Its existence is also what registers `srfi-277` as a `cond-expand` feature.
- The raw `%` constructors tolerate an empty source (reads EOF forever) rather than modulo-by-zero, since `(kaappi primitives)` exposes them directly.

The portable route over `(srfi 181)` was rejected for the reason the issue documents: the 4096-byte `read!` burst parks its tail in `read_buf` and the custom-port position correction (#1996) assumes a monotonic `get-position`, so the reference's modular index yields negative positions after one burst; a textual custom port would additionally misreport non-ASCII cycles (byte vs character offsets).

Since the port reads its own snapshot copy, Kaappi defines the SRFI's "source modified after the call" case: mutations never affect the port. Documented in `docs/dev/srfi-implementation-notes.md`'s new SRFI 277 section, alongside the burst/positioning trap.

## Tests

- `tests/scheme/srfi/srfi277.scm` — 39 named SRFI 64 assertions: both spec examples verbatim, reads spanning several cycles, the `#u8(0)` /dev/zero, port predicates and `call-with-port`, never-EOF (100-byte read, `char-ready?`/`u8-ready?`, `read-bytevector!` filling fully), multi-byte wrap under `read-char`, `peek-char`/`read-char` across the wrap, `read-line` on a cycle with a newline, the SRFI's own re-read positioning test plus arbitrary-position seeks, empty/type/closed error set, the snapshot extension, and the 271 seeding integration (equal cyclic seeds ⇒ identical streams; a 3-byte cycle satisfies the 32-byte seed read).
- Zig unit tests in `tests_io.zig` (wrap, monotonic position, unbounded seek, empty-cycle degenerate path, `read` datum parsing), and the `(kaappi primitives)` export-contract test now covers both new `%` names.

Verified: `zig build test` and `-Dgc-stress=true`, `tests/scheme/run-all.sh` (2153 pass, 0 fail — srfi277 included), R7RS suite (1395 pass), `zig fmt`/markdownlint clean, `kaappi features --json` re-derived (12 built-in + 165 portable, `srfi-277` present). `kaappi compile` of a program importing `(srfi 277)` gets the standard #1743 refuse-loudly diagnostic exactly like every other disk-resolved portable SRFI; `-Dbundle-src` produces a working standalone binary.

## Bookkeeping

- Counts 180 → 181 SRFIs, 164 → 165 portable in `CLAUDE.md` (AGENTS.md), `README.md`, and the notes header; `| 277 | Cyclic ports |` row in `CONFORMANCE.md`.
- Follow-up for the docs site repo: `docs/guide/srfi-support.md` (not in this repo).
- `srfi-status-check` needs nothing — SRFI 277 is `final`.

### DIFF
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -13,8 +13,8 @@ Companions:
 
 ## What ships
 
-180 SRFIs supported. 12 built-in (Zig primitives): 1, 9, 13, 18, 39, 69, 133,
-170, 192, 254, 258, 260. 164 portable R7RS .sld files loaded on demand via
+181 SRFIs supported. 12 built-in (Zig primitives): 1, 9, 13, 18, 39, 69, 133,
+170, 192, 254, 258, 260. 165 portable R7RS .sld files loaded on demand via
 `(import (srfi N))`, plus SRFI 261 (Portable SRFI Library References) as an
 import-resolver convention with no library file, and SRFI 226, SRFI 160, and
 SRFI 211 (see below) as sub-libraries only with no bare `(srfi 226)`/`(srfi
@@ -29,7 +29,7 @@ features`' scan): 0, 2, 4, 5, 6, 7, 8, 11, 14, 16, 17, 19, 23, 25, 26, 27, 28,
 201, 202, 203, 207, 209, 210, 213, 214, 215, 216, 217, 219, 221, 222, 223,
 224, 225, 227, 228, 229, 231, 232, 233, 234, 235, 236, 237, 238, 239, 240,
 241, 242, 244, 247, 248, 250, 251, 252, 253, 255, 257, 259, 263, 264, 267,
-270, 271, 273, 274. Sub-libraries: (srfi 146 hash), (srfi 171 meta), (srfi 166 pretty),
+270, 271, 273, 274, 277. Sub-libraries: (srfi 146 hash), (srfi 171 meta), (srfi 166 pretty),
 (srfi 166 columnar), (srfi 166 unicode), (srfi 166 color), (srfi 211
 explicit-renaming), (srfi 211 define-macro), (srfi 211 syntax-parameter),
 (srfi 226 control prompts), (srfi 226 control continuations), (srfi 226
@@ -1714,3 +1714,81 @@ sub-library catalogue — it is plumbing) holding the shared `argcheck!` and
   `(srfi 274 internal)`; the reference recurses, also properly tail-recursive).
   Boundedness is the point: the check terminates on circular lists and is
   what lets every `end`-supplied conversion accept them.
+
+### SRFI 277 — cyclic ports
+
+A cyclic port is an ordinary input port over a string or bytevector whose
+stream repeats forever: for `x0 … xn` it delivers `x0 … xn x0 … xn …` and
+never an EOF object. It exists chiefly to give SRFI 271's
+`make-random-port` a reproducible seed that is not a caller-built 32-byte
+bytevector — `(make-random-port (open-cyclic-input-bytevector #u8(1 2 3)))`
+— plus the usual endless-pattern jobs (a portable `/dev/zero` is
+`(open-cyclic-input-bytevector #u8(0))`).
+
+**The native route, mirroring SRFI 271's shape.** Kaappi's string/bytevector
+input ports are already a cyclic port minus the wrap: an owned snapshot in
+`Port.string_data` plus a cursor `string_pos`, with every read primitive
+funneling through the single byte source `readOneByte`. So the whole engine
+change is a plain-bool `Port.cyclic` turning that path's one EOF exit into a
+wrap:
+
+- `readOneByte` indexes `data[string_pos % data.len]` and lets
+  `string_pos` keep counting past `data.len` (guarded on `len > 0` — the raw
+  `%` constructors are callable from `(kaappi primitives)` with an empty
+  source, and modulo-by-zero would panic under ReleaseSafe; such a port
+  reads EOF forever, degenerate but safe).
+- `port-position` needs no change: it already reports
+  `string_pos − read-ahead` (kaappi#1941), which is now an unbounded
+  monotonic byte count — exactly the SRFI 192 contract, same as
+  `open-input-string`.
+- `set-port-position!` skips its `pos > len` bound for a cyclic port and
+  stores any non-negative position; the wrap makes it meaningful.
+- `read` takes the incremental (custom-port/fd) path instead of the
+  string-port fast path — that path slices `data[string_pos..]`, which the
+  unbounded cursor would run out of bounds, and worse, it reports EOF for a
+  datum unterminated at the snapshot's end, which a cyclic port must never
+  do. The incremental path just keeps pulling bytes until a datum completes.
+- `char-ready?`/`u8-ready?` needed nothing: a string port is always ready,
+  cyclic ones included.
+
+Everything above the byte source works unchanged because the wrap lands at
+the end of the whole snapshot, which is always a codepoint boundary:
+UTF-8 decoding, `peek-char`'s lookahead (its pushed-back-byte
+reconstruction indexes modulo len for the same unbounded-cursor reason),
+`read-line`'s CR handling, `close-port`.
+
+**Why not the portable route over `(srfi 181)`.** The reference
+implementation is ~50 lines of `make-custom-*-input-port`, and a verbatim
+port passes the read tests but fails the SRFI's own positioning test on
+Kaappi, for a reason worth remembering before anyone retries that route:
+`readOneByteFromCustomPort` calls `read!` in bursts of
+`read_chunk_size = 4096` and parks the tail in `read_buf`, and
+`portPositionFromCustomPort` then reports `get-position − read-ahead`
+(kaappi#1996) — an adjustment that assumes `get-position` returns a
+*monotonic source position*. The reference's `get-position` returns the
+modular index into the cycle, so after `(read-bytevector 13 p)` on a 3-byte
+source the correction computes `4096 mod 3 − (4096 − 13) = −4082` and
+`set-port-position!` rejects it as negative. A custom port *could* paper
+over this by reporting a monotonic count, but two further costs remain:
+kaappi's positions are byte offsets while a custom *textual* port's `read!`
+counts characters, so non-ASCII cycles would still misreport, and every
+4096-byte burst is a 4096-iteration Scheme loop. Native was cheaper and
+exactly right.
+
+**Snapshot copy is the documented extension.** The SRFI leaves "source
+modified after the call" undefined behavior; since the port reads its own
+copy (`allocCyclicInputPort` dupes the data, as `allocStringInputPort`
+always has), Kaappi defines it: mutations of the source never affect the
+port. Pinned in `tests/scheme/srfi/srfi277.scm`.
+
+**Positions are byte offsets**, as everywhere else in Kaappi (see the SRFI
+192 section in CONFORMANCE.md): `port-position` on
+`(open-cyclic-input-string "λμ")` counts UTF-8 bytes, not characters.
+
+**Endless means endless.** `read-line` on a cycle with no newline never
+returns, and `read` on a cycle that never completes a datum accumulates
+until memory does — spec-conforming (the same is true of reading
+`/dev/zero` anywhere), so the test suite deliberately does not exercise
+either. Positions on a cyclic port grow without bound; `string_pos` is a
+`usize` and would take ~584 years of one-byte reads per nanosecond to
+overflow it.
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -1747,7 +1747,14 @@ wrap:
   string-port fast path — that path slices `data[string_pos..]`, which the
   unbounded cursor would run out of bounds, and worse, it reports EOF for a
   datum unterminated at the snapshot's end, which a cyclic port must never
-  do. The incremental path just keeps pulling bytes until a datum completes.
+  do. The incremental path just keeps pulling bytes until a datum completes
+  (a whole 4096-byte burst per parse attempt — the cycle can never block or
+  EOF, so reading ahead is free) and then rewinds `string_pos` over the
+  unconsumed tail instead of buffering it into `read_buf`: no string port
+  may carry read-ahead there, because `peek-char`'s string-port fast paths
+  rewind the cursor on exactly that assumption. Buffering the tail (the
+  burst refill's first cut) made `peek-char` serve one character and the
+  following `read-char` a different one after every `(read p)`.
 - `char-ready?`/`u8-ready?` needed nothing: a string port is always ready,
   cyclic ones included.
 
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -1777,9 +1777,9 @@ exactly right.
 
 **Snapshot copy is the documented extension.** The SRFI leaves "source
 modified after the call" undefined behavior; since the port reads its own
-copy (`allocCyclicInputPort` dupes the data, as `allocStringInputPort`
-always has), Kaappi defines it: mutations of the source never affect the
-port. Pinned in `tests/scheme/srfi/srfi277.scm`.
+copy (`allocStringInputPort` dupes the data, and the cyclic constructors
+just flip the flag on the result), Kaappi defines it: mutations of the
+source never affect the port. Pinned in `tests/scheme/srfi/srfi277.scm`.
 
 **Positions are byte offsets**, as everywhere else in Kaappi (see the SRFI
 192 section in CONFORMANCE.md): `port-position` on
@@ -1789,6 +1789,10 @@ port. Pinned in `tests/scheme/srfi/srfi277.scm`.
 returns, and `read` on a cycle that never completes a datum accumulates
 until memory does — spec-conforming (the same is true of reading
 `/dev/zero` anywhere), so the test suite deliberately does not exercise
-either. Positions on a cyclic port grow without bound; `string_pos` is a
-`usize` and would take ~584 years of one-byte reads per nanosecond to
-overflow it.
+either. Positions on a cyclic port grow without bound; on 64-bit targets
+`string_pos` would take ~584 years of one-byte reads per nanosecond to
+overflow, and on wasm32 `set-port-position!` bounds the value to `usize`
+explicitly. `read` pulls a whole 4096-byte burst per parse attempt (the
+cycle can never block or EOF, so reading ahead is free) — one byte per
+re-parse of the whole accumulation made an 85 KB datum take 50 s, versus
+1 ms on the plain string port.
--- a/src/gc_alloc.zig
+++ b/src/gc_alloc.zig
@@ -572,37 +572,6 @@ pub fn allocStringOutputPort(self: *GC) !Value {
     return types.makePointer(&port.header);
 }
 
-/// SRFI 277: a cyclic input port — an ordinary input string port whose
-/// stream wraps to `data`'s start instead of ending. The port reads its own
-/// snapshot copy, so mutating the source afterwards cannot affect it (the
-/// SRFI leaves the source-mutation case undefined; the snapshot is the
-/// well-defined extension). Callers flip `is_binary` themselves for the
-/// bytevector constructor, exactly as openInputBytevector does.
-pub fn allocCyclicInputPort(self: *GC, data: []const u8) !Value {
-    // Copy before collecting: `data` usually aliases a SchemeString or
-    // Bytevector.
-    const owned = try self.allocator.dupe(u8, data);
-    errdefer self.allocator.free(owned);
-    try self.maybeCollect();
-    const port = try self.allocator.create(Port);
-    port.* = .{
-        .header = .{ .tag = .port },
-        .fd = -1,
-        .is_input = true,
-        .is_output = false,
-        .is_open = true,
-        .name = "string",
-        .owns_name = false,
-        .peek_byte = null,
-        .is_string_port = true,
-        .string_data = owned,
-        .string_pos = 0,
-        .cyclic = true,
-    };
-    self.finishAlloc(&port.header, @sizeOf(Port) + data.len);
-    return types.makePointer(&port.header);
-}
-
 /// SRFI 181: a port backed by user-supplied Scheme procedures. Up to 6
 /// Values need protecting across the collection at once -- beyond
 /// rootArgs2's 2-Value cap -- so this follows allocMultipleValues's
--- a/src/gc_alloc.zig
+++ b/src/gc_alloc.zig
@@ -572,6 +572,37 @@ pub fn allocStringOutputPort(self: *GC) !Value {
     return types.makePointer(&port.header);
 }
 
+/// SRFI 277: a cyclic input port — an ordinary input string port whose
+/// stream wraps to `data`'s start instead of ending. The port reads its own
+/// snapshot copy, so mutating the source afterwards cannot affect it (the
+/// SRFI leaves the source-mutation case undefined; the snapshot is the
+/// well-defined extension). Callers flip `is_binary` themselves for the
+/// bytevector constructor, exactly as openInputBytevector does.
+pub fn allocCyclicInputPort(self: *GC, data: []const u8) !Value {
+    // Copy before collecting: `data` usually aliases a SchemeString or
+    // Bytevector.
+    const owned = try self.allocator.dupe(u8, data);
+    errdefer self.allocator.free(owned);
+    try self.maybeCollect();
+    const port = try self.allocator.create(Port);
+    port.* = .{
+        .header = .{ .tag = .port },
+        .fd = -1,
+        .is_input = true,
+        .is_output = false,
+        .is_open = true,
+        .name = "string",
+        .owns_name = false,
+        .peek_byte = null,
+        .is_string_port = true,
+        .string_data = owned,
+        .string_pos = 0,
+        .cyclic = true,
+    };
+    self.finishAlloc(&port.header, @sizeOf(Port) + data.len);
+    return types.makePointer(&port.header);
+}
+
 /// SRFI 181: a port backed by user-supplied Scheme procedures. Up to 6
 /// Values need protecting across the collection at once -- beyond
 /// rootArgs2's 2-Value cap -- so this follows allocMultipleValues's
--- a/src/primitives_io.zig
+++ b/src/primitives_io.zig
@@ -1271,6 +1271,14 @@ fn setPortPositionBang(args: []const Value) PrimitiveError!Value {
             // position is valid — the unbounded cursor simply counts into
             // the endless repetition, and reads wrap modulo len.
             if (!(port.cyclic and len > 0) and pos > len) return primitives.indexError("set-port-position!", pos, len);
+            // Positions are fixnums (±2^47), so on 64-bit targets the
+            // @intCast below is provably in range. wasm32's usize is u32,
+            // though, and its shipped build is ReleaseSmall — no trap — so a
+            // cyclic position in [2^32, 2^47] would silently truncate to a
+            // wildly different position rather than error.
+            if (port.cyclic and len > 0 and @as(u64, @intCast(pos)) > std.math.maxInt(usize)) {
+                return primitives.indexError("set-port-position!", pos, std.math.maxInt(usize));
+            }
             port.string_pos = @intCast(pos);
             // Discard stale read-ahead exactly as the fd branch below
             // does: a pushed-back peek byte describes the *old* position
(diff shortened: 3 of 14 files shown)


src/primitives_io.zig
```diff
@@ -2135,13 +2164,14 @@ fn readDatumFn(args: []const Value) PrimitiveError!Value {
...
-        // as EOF, so `read` never invoked the callback at all.
-        if (port.custom_backend != null or port.transcode != null) {
+        // SRFI 181 ports have no fd (the -1 sentinel), and SRFI 277 cyclic
+        // string ports must never see EOF: refill through readOneByte — the
+        // single byte source for every port kind — which invokes the read!
+        // callback / transcoder decode / cycle wrap and stashes the burst's
+        // tail into read_buf, drained right back out here. Before #1995 this
+        // fell through to portFdRead(-1, ...), whose EBADF read as EOF, so
+        // `read` never invoked the callback at all.
+        if (port.custom_backend != null or port.transcode != null or port.cyclic) {
```
baijum, line 2194: **`read` on a cyclic port is quadratic in the datum's length — 50 s for an 85 KB datum, versus 1 ms on the plain string port.**

This branch pulls **one byte** per loop iteration for a cyclic port (the string branch of `readOneByte` never fills `read_buf`, unlike the custom/transcoded branches whose 4096-byte burst gets drained into `buf` right below), and the loop re-parses the whole accumulation from scratch after every byte. Measured on this head, `(read p)` of one list datum followed by a space:

| datum | `open-input-string` | `open-cyclic-input-string` |
|---|---|---|
| 3.9 KB (1000 elems) | 0.05 ms | 136 ms |
| 19 KB (4000) | 0.2 ms | 2.75 s |
| 85 KB (16000) | 0.9 ms | 50.4 s |

4.5× the bytes → 18× the time, i.e. O(n²). Nothing in the suite sees it because the only `read` test uses a 4-byte datum. A user who does `(read (open-cyclic-input-string big-config-text))` gets what looks like a hang.

Cheap fix, in this branch only for `port.cyclic`: pull a burst rather than a byte before falling back to the parse — e.g. loop `readOneByte` up to `read_chunk_size` times (a cyclic port can never block or EOF, so reading ahead is free) and append them all to `buf`. The datum-complete path already hands the unconsumed tail back to `port.read_buf`, so `port-position` and `set-port-position!` stay correct; my probes of both after `read` (position after two reads with a buffered remainder, then seek 0 and re-read) pass on the current code and would be unaffected. That turns the cost into O(n²/4096), which is what fd and custom ports pay today.

Worth a regression test with a multi-kilobyte datum (a list of a few thousand fixnums) — not timed, just large enough that the quadratic form would take seconds under the Debug CI leg.

src/primitives_io.zig
```diff
@@ -1264,7 +1267,10 @@ fn setPortPositionBang(args: []const Value) PrimitiveError!Value {
     if (port.is_string_port) {
         if (port.is_input) {
             const len = if (port.string_data) |d| d.len else 0;
-            if (pos > len) return primitives.indexError("set-port-position!", pos, len);
+            // SRFI 277: a cyclic port has no end, so any non-negative
+            // position is valid — the unbounded cursor simply counts into
+            // the endless repetition, and reads wrap modulo len.
+            if (!(port.cyclic and len > 0) and pos > len) return primitives.indexError("set-port-position!", pos, len);
```
baijum, line 1273: Agree with CodeRabbit on **the cast, not the increment**, and narrower than it says. Positions arrive as fixnums (±2^47), so on every 64-bit target `@intCast(pos)` is provably in range and nothing changes. On **wasm32** `usize` is `u32` and the shipped build is `ReleaseSmall` (safety off, `build.zig:538`), so a cyclic-port position in `[2^32, 2^47]` silently truncates instead of trapping — `(set-port-position! p 4294967296)` lands at 0. One line closes it:

```zig
if (port.cyclic and len > 0 and pos > std.math.maxInt(usize)) return primitives.indexError("set-port-position!", pos, std.math.maxInt(usize));
```

(or `argError`, whichever reads better). The `string_pos += 1` wrap CodeRabbit also flags needs 4 GiB of reads inside a browser tab first; not worth code, but the notes sentence about 584 years should say "on 64-bit targets" (see my comment there).

docs/dev/srfi-implementation-notes.md
```diff
@@ -1714,3 +1714,81 @@ sub-library catalogue — it is plumbing) holding the shared `argcheck!` and
...
+**Positions are byte offsets**, as everywhere else in Kaappi (see the SRFI
+192 section in CONFORMANCE.md): `port-position` on
+`(open-cyclic-input-string "λμ")` counts UTF-8 bytes, not characters.
+
+**Endless means endless.** `read-line` on a cycle with no newline never
+returns, and `read` on a cycle that never completes a datum accumulates
+until memory does — spec-conforming (the same is true of reading
+`/dev/zero` anywhere), so the test suite deliberately does not exercise
+either. Positions on a cyclic port grow without bound; `string_pos` is a
+`usize` and would take ~584 years of one-byte reads per nanosecond to
```
baijum: Nit, tied to the `set-port-position!` comment: the 584-year figure is true for 64-bit `usize` only — on wasm32 the counter is 32-bit (4.3 s at that rate), and separately the *settable* range is capped by fixnums at 2^47 on every target. Suggest: "Positions on a cyclic port grow without bound; on 64-bit targets `string_pos` would take ~584 years of one-byte reads per nanosecond to overflow, and on wasm32 `set-port-position!` bounds the value to `usize` explicitly."

src/gc_alloc.zig
```diff
@@ -572,6 +572,37 @@ pub fn allocStringOutputPort(self: *GC) !Value {
     return types.makePointer(&port.header);
 }
 
+/// SRFI 277: a cyclic input port — an ordinary input string port whose
+/// stream wraps to `data`'s start instead of ending. The port reads its own
+/// snapshot copy, so mutating the source afterwards cannot affect it (the
+/// SRFI leaves the source-mutation case undefined; the snapshot is the
+/// well-defined extension). Callers flip `is_binary` themselves for the
+/// bytevector constructor, exactly as openInputBytevector does.
+pub fn allocCyclicInputPort(self: *GC, data: []const u8) !Value {
```
baijum: Nit (optional): this is `allocStringInputPort` verbatim plus `.cyclic = true`. The bytevector constructor two files over already flips `is_binary` on the returned port after allocation, so the same pattern — `gc.allocStringInputPort(...)` then `.as(types.Port).cyclic = true` — would drop these 31 lines and the `memory.zig` alias with no behaviour change. Keeping one allocator also means a future field added to string input ports cannot be initialised in one copy and forgotten in the other. Fine either way.

tests/scheme/srfi/srfi277.scm
```diff
@@ -0,0 +1,206 @@
...
+  (test-assert "cyclic bytevector port is not textual" (not (textual-port? p))))
+
+(test-assert "call-with-port closes the cyclic port afterwards"
+  (let ((p (open-cyclic-input-string "x")))
+    (call-with-port p (lambda (port) (read-char port)))
+    (not (input-port-open? p))))
+
+;;; --- reading never produces EOF ------------------------------------------
+
+(test-assert "a long read never yields the eof object"
```
baijum, line 71: Nit: the assertion only inspects the 100th value (`last` is overwritten each iteration, and `(= i 100)` checks it once). It passes for the right reason today, but the name promises more than it checks — `(let loop ((i 0)) (or (= i 100) (and (not (eof-object? (read-u8 p))) (loop (+ i 1)))))` pins every read.

baijum (commented):
I built this head, ran the PR's Zig tests by name filter (pass), the Scheme suite (39/39), `kaappi features --json` (277 present), `kaappi check` on the test file (clean), and a dozen extra wrap-point probes the suite does not cover: `read-line` over a cycle ending in a bare CR (the #1941 pushback lands exactly on the wrap), `peek-char` after that pushback with a multi-byte lead byte (the new modulo reconstruction), `peek-u8` across the wrap, `port-position` after peeks and after `read` leaves a buffered remainder, `set-port-position!` mid-codepoint (same lenient result as a plain string port), seek-then-`read`, thread deep-copy (refused, same as `open-input-bytevector`), and `(read p)` over `"42 "` / `"(a . b)"`. All correct. The design matches the issue and the docs section is the one I would have wanted to read.

**One thing I think should land before merge:** `read` on a cyclic port refills one byte per loop iteration and re-parses the whole accumulation each time, so it is quadratic — 50 s for an 85 KB datum (1 ms on `open-input-string`). Details and a small fix inline at `primitives_io.zig:2174`.

**Minor:** the wasm32 `@intCast` window in `set-port-position!` (inline at :1273), plus a nit each on the duplicated allocator, the notes' overflow sentence, and one test assertion.

**CodeRabbit triage** — I would decline four of its five:
- *CLAUDE.md 717 → 718*: the built-in count is refreshed at release time by `/github-release` Step 5, deliberately not per PR (and the diff adds two `%` names, so 718 is not even the right number).
- *cond-expand-gate `(kaappi primitives)`*: eight other shipped `.sld`s (27, 74, 271/*, 57, 131, 136, 150, 237) import it unconditionally; `lib/srfi` is Kaappi's own library tree, not a portable distribution.
- *`(srfi 64)` in tests*: 201 of the 240 files in `tests/scheme/srfi/` use it. The `tests/**` path instruction in `.coderabbit.yaml` is stale, not the PR — worth fixing there separately.
- *usize increment overflow*: see inline; the cast is the real half of that finding.

**riscv64 leg:** the `rest` unit-test chunk was cancelled at the 55-minute job limit after the other seven chunks passed; nothing in this diff is arch-specific and it is the known QEMU-slowness shape (#2488 family). A re-run should clear it.

baijum:
Addressed the review in efb83478:

**Merge-blocking — quadratic `read`** (primitives_io.zig): the cyclic refill now pulls a full `read_chunk_size` burst per parse attempt instead of one byte (custom/transcoded refills keep their existing shape — one `readOneByte` call is already one burst there). Measured: 19 KB datum 2666 ms → 1.2 ms; 85 KB datum 50.4 s → ~14 ms. Position-after-read and seek-reread probes still exact (the parse hands the unconsumed tail to `read_buf`). Added the suggested regression test — a 4000-element datum, untimed.

Honest note: my first version of that burst loop hung the unit suite — its `orelse break` ended the *burst*, not the parse loop, so any custom/transcoded port whose datum completes only at EOF spun forever (`tests_srfi181` caught it). The pushed fix propagates EOF via a flag into the post-loop final-verdict parse, and a new unit test pins the empty-cycle `read` through `readDatumFn`. Full unit suite re-run green after.

**wasm32 `@intCast` window** (primitives_io.zig): added `pos > maxInt(usize)` → `indexError` before the cast, exactly the narrow form from the review comment (fixnum range ⇒ no-op on 64-bit; prevents silent truncation on wasm32/ReleaseSmall). Verified `zig build wasm` compiles it.

**Nits, all applied:**
- Notes overflow sentence now says "on 64-bit targets" and points at the explicit wasm32 bound; recorded the burst-refill rationale there too.
- `allocCyclicInputPort` removed — both constructors call `allocStringInputPort` and flip `.cyclic` (plus `is_binary`) on the result, matching `openInputBytevector`'s pattern; the `memory.zig` alias went with it.
- The never-EOF scheme test now asserts every `read-u8` result, not just the 100th.

**Declined**, per review triage: CLAUDE.md 717→718, cond-expand gating of `(kaappi primitives)`, `(srfi 64)` harness swap.

Verified after: `zig build test` (exit 0), `-Dgc-stress=true` (exit 0), full `tests/scheme/run-all.sh` (2153 pass, 0 fail, srfi277 now at 40 assertions), R7RS 1395 pass, `zig fmt` + markdownlint clean, wasm build clean.

baijum (commented):
Re-reviewed efb83478 the same way: built it, ran the SRFI 277 unit tests by name filter (pass) and `srfi277.scm` (40/40), re-timed the 85 KB datum (50 s → 14 ms, as reported), and re-ran the earlier wrap-point probes plus a new set targeting what the burst changes. The allocator dedup, the wasm32 bound, the `hit_eof` handling (I agree the flag is needed so a custom port whose datum completes only at EOF still reaches the final-verdict parse), the every-read assertion, and the notes edits all look right, and the four CodeRabbit declines match my triage.

**One new finding, which I think should land before merge:** after `(read p)` on a cyclic port, `peek-char` returns one character and the following `read-char` returns a different one — `(#\b #\()` for `"(a)b"`, and worse for multi-byte and after a `read-line` CR pushback. Cause and a fix I verified locally are inline at `primitives_io.zig:2193`. In short: `read` now leaves ~4 KB in `read_buf` on a string port, which no string port ever had before, and `peek-char`'s string-port fast paths rewind the cursor on that assumption. Rewinding `string_pos` at the datum-complete site instead of buffering the tail keeps the invariant and fixes all three cases with no cost to the speed-up. The narrow form of this was already in 84c1adbe (only when a token needed a delimiter byte) and I missed it then — my probes checked position and seek after `read`, not `peek-char`.

CI on this head is green except the still-pending riscv64 and windows-arm legs.

src/primitives_io.zig
```diff
@@ -2135,28 +2172,53 @@ fn readDatumFn(args: []const Value) PrimitiveError!Value {
...
+            // readOneByte call is one read! callback / decode burst, and the
+            // drain below moves its whole read_buf tail into `buf`. A cyclic
+            // port's readOneByte hands out one byte, so pull a full burst per
+            // iteration: a cyclic port never blocks and never EOFs, and one
+            // byte at a time would re-parse the entire accumulation after
+            // every byte — quadratic in the datum's length (a 19 KB datum
+            // took 2.7 s, versus 0.2 ms on the plain string port). The parse
+            // at the top of the loop hands any unconsumed tail back to
+            // read_buf, so port-position and set-port-position! stay exact.
+            const burst: usize = if (port.cyclic) read_chunk_size else 1;
```
baijum, line 2205: **The burst fixes the quadratic cost (85 KB datum: 50 s → 14 ms on my machine, matching your numbers) but it widens a bug I missed on the first pass: after `(read p)` on a cyclic port, `peek-char` and `read-char` disagree.**

On this head, with an isolated `KAAPPI_HOME`:

```scheme
(let ((p (open-cyclic-input-string "(a)b")))
  (read p)
  (list (peek-char p) (read-char p)))        ; ⇒ (#\b #\()   expected (#\b #\b)

(let ((p (open-cyclic-input-string "(a)λ")))
  (read p)
  (list (peek-char p) (read-char p) (port-position p)))   ; ⇒ (#\λ #\( 4)   expected (#\λ #\λ 5)

(let ((p (open-cyclic-input-string "(a) \rλ")))
  (read p) (read-line p)                     ; read-line's CR pushback (#1941)
  (list (peek-char p) (read-char p)))        ; ⇒ (#\Î #\λ)   expected (#\λ #\λ)
```

The same three on `open-input-string` all come out right, and `set-port-position!` in between (which discards `read_buf`) also makes them right — which is the tell.

**Why.** `read` hands the unconsumed tail of the parse back to `port.read_buf` (`primitives_io.zig:2158`). Until this PR no string port ever carried a `read_buf` — `portPosition`'s own comment says so (`primitives_io.zig:1192`, "no string-port path fills them today") — and two `peek-char` paths lean on that invariant: the un-peeked path rewinds the cursor with `string_pos -= len` (`:1890`) instead of pushing the bytes back, so a character served from `read_buf` is consumed and the cursor moves *backwards* over bytes that were never read; and the pushed-back path reconstructs a multi-byte character from `data[string_pos − 1 + i]`, which is the wrong place when `peek_byte` came out of `read_buf`. The first commit had this too but narrowly (a tail existed only when a symbol or number needed one delimiter byte to complete); the burst makes every `read` leave ~4 KB behind, so any `read` followed by `peek-char` now hits it.

**Fix I verified locally** (probes above all correct, 85 KB read still 14 ms, `srfi277.scm` 40/40): keep the string-port invariant instead of teaching `peek-char` about `read_buf`. At the datum-complete site (`:2153`–`:2160`), a cyclic port rewinds rather than buffers:

```zig
const remaining = buf.items[reader.pos..];
if (port.cyclic and remaining.len > 0) {
    // Every byte in `buf` advanced string_pos (peek_byte was already past
    // it; string ports never fill read_buf/peek_extra), so the unconsumed
    // tail is simply the last remaining.len positions — rewind, and keep
    // string ports read_buf-free, which peek-char's rewind relies on.
    port.string_pos -= remaining.len;
} else if (remaining.len > 0) {
    ... existing read_buf save ...
}
```

`string_pos ≥ remaining.len` holds because everything in `buf` either advanced `string_pos` (the burst) or sat behind an already-advanced cursor (`peek_byte`); `set-port-position!` clears both before it can interleave. The post-EOF site (`:2268`) is reached by a cyclic port only on the empty cycle, where `buf` is empty, so it can stay as is. Then the "no string-port path fills them today" comment stays true, and the peek-char rewind at `:1890` stays correct by construction. Please pin the three cases above (ASCII, multi-byte, CR-pushback) in `srfi277.scm` — the position assertion in the second one is what catches the backwards cursor.

src/primitives_io.zig
```diff
@@ -1264,7 +1267,18 @@ fn setPortPositionBang(args: []const Value) PrimitiveError!Value {
...
+            // SRFI 277: a cyclic port has no end, so any non-negative
+            // position is valid — the unbounded cursor simply counts into
+            // the endless repetition, and reads wrap modulo len.
+            if (!(port.cyclic and len > 0) and pos > len) return primitives.indexError("set-port-position!", pos, len);
+            // Positions are fixnums (±2^47), so on 64-bit targets the
+            // @intCast below is provably in range. wasm32's usize is u32,
+            // though, and its shipped build is ReleaseSmall — no trap — so a
+            // cyclic position in [2^32, 2^47] would silently truncate to a
+            // wildly different position rather than error.
+            if (port.cyclic and len > 0 and @as(u64, @intCast(pos)) > std.math.maxInt(usize)) {
```
baijum, line 1279: LGTM. `@as(u64, @intCast(pos))` is safe because `pos < 0` was already rejected above, and on 64-bit the comparison folds to `false`. I also checked a 2^47−1 seek (`mod 3 = 1`) reads `#\b` as expected.

tests/scheme/srfi/srfi277.scm
```diff
@@ -0,0 +1,221 @@
...
+    (list (read-line p) (read-line p) (read-line p))))
+
+(test-equal "read parses successive datums from a cyclic port"
+  '((a) b)
+  (let ((p (open-cyclic-input-string "(a)b")))
+    (list (read p) (read p))))
+
+;; Large enough that the one-byte-per-reparse form of the refill loop (fixed
+;; to a 4096-byte burst) would take seconds under the Debug CI leg; not timed.
+(test-equal "read of a multi-kilobyte datum from a cyclic port"
```
baijum, line 150: Good regression shape — 4000 elements is 19 KB, which was 2.7 s in the quadratic form. Alongside it please add the `read`-then-`peek-char` cases from my comment at `primitives_io.zig:2193`; with the current head they fail, and they are the ones that guard the `read_buf` invariant rather than the speed.

baijum:
Addressed in b76d903a.

**The peek-char finding** — reproduced all three probes first (`(#\b #\()`, `(#\λ #\( 4)`, `(#\Î #\λ)`), then applied the inline fix: at the datum-complete site, a cyclic port now rewinds `string_pos` by the unconsumed tail's length instead of saving the tail to `read_buf`, keeping string ports read_buf-free so `peek-char`'s rewind paths stay correct by construction. After: `(#\b #\b)`, `(#\λ #\λ 5)`, `(#\λ #\λ)`; the 85 KB `read` is still ~14 ms; the rewind is safe against underflow because every buffered byte either advanced `string_pos` (the burst) or sat behind an already-advanced cursor (the drained `peek_byte`), so the tail can never exceed what the cursor advanced. All three probes are pinned in `srfi277.scm` (43 assertions now), including the position assertion that catches the backwards cursor. The post-EOF site is untouched, as suggested — a cyclic port reaches it only with an empty buffer.

The invariant rationale is recorded in the SRFI 277 notes section, since it's the kind of constraint that isn't visible from the code.

**CodeRabbit's two trivia, declined:** splitting `primitives_io.zig` — the file-size policy explicitly says flat primitives files are exempt from the 1500-line cap ("size from breadth is fine"), and `port_readiness.zig`'s header already documents `primitives_io` as such a file; and `TestContext` in the new unit test — `tests_io.zig` uses `makeTestVM` 36 times and `TestContext` zero times, so the test follows its file's convention.

Verified after: full unit suite exit 0, `-Dgc-stress=true` exit 0, `tests/scheme/run-all.sh` 2153 pass / 0 fail, `zig fmt` + markdownlint clean.

Issue #2548, opened 2026-09-07
Reader accepts only #u8( of SRFI 4's homogeneous-vector literal syntax


baijum:
**Kaappi version**

`Kaappi Scheme v0.26.2` (main), macOS aarch64.

**Found by**

`tools/srfi231_diff.py prose` (the spec-examples mode, PR in progress): SRFI 231's example that inspects a `u1` array's body writes its expected value as `#(16 #u16(3895))` in code position, and Kaappi rejected the whole program at read time.

**Description**

Kaappi advertises SRFI 4 (`kaappi features --json` lists 4 among the portable SRFIs; `lib/srfi/4.sld` is a re-export over `(srfi 160 <tag>)`). SRFI 4 specifies the external representation of every homogeneous vector type — `#s8(...)`, `#u8(...)`, `#s16(...)`, `#u16(...)`, `#s32(...)`, `#u32(...)`, `#s64(...)`, `#u64(...)`, `#f32(...)`, `#f64(...)` — as syntax `read` accepts. The reader accepts only `#u8(` (the R7RS bytevector form; `reader_tokens.zig:846`); every other prefix is a read error, so a conforming SRFI 4 program that uses a literal fails before it runs.

**Minimal reproduction**

```scheme
(import (scheme base) (scheme write) (srfi 4))
(write (u8vector-length #u8(1 2 3)))(newline)
(write (u16vector-length #u16(1 2 3)))(newline)
```

**Output**

Kaappi:

```
3
u16.scm:3:27: read error[KP1002]: unexpected character
```

Gambit 4.9.8 (reference for the SRFI 231 corpus; chibi reads it too):

```
3
3
```

**Notes**

- `write` of a u16vector prints `#<u16vector 3895>` here, so the written form is not readable either; SRFI 4 says the external representation is what `write` produces.
- `docs/dev/srfi-exclusions.md` discusses the *SRFI 58* family of `#NdTAG(...)` multi-dimensional literals as a deliberate exclusion; the flat SRFI 4 forms are a different, much smaller addition to the `#` dispatch (ten fixed prefixes, each followed by `(`) and are what the advertised SRFI requires.
- Priority medium: a spec violation, loud (read error), narrow (literals only; the procedures work).

PR #2549, opened 2026-09-07
Add the spec-examples mode to the SRFI 231 differential tester

baijum:
Fourth and last of the planned modes for `tools/srfi231_diff.py` (#2541, #2543, #2544): the spec's own code examples as a corpus. `srfi-231.html` carries 172 code blocks, most annotated with the value each expression should produce — a corpus the official suite does not contain.

**What a case is.** One code block per case (the seed is the 1-based example number; `--count` defaults to all). A small s-expression reader wraps each top-level expression — and each body expression of a top-level `let`, which is where the spec puts its `;; =>` annotations — in a printer, so every annotated value is compared, not only a block's last. Values go through a canonical renderer: arrays as domain plus contents (so a lazy array is forced through its own getter), intervals as bounds, opaque objects as a tag, floats as exact rationals, homogeneous vectors as tagged lists, the unspecified value as a symbol. `display`, `pretty-print` and `pp` are replaced by it in the prelude, since Gambit's line breaking and its `#u16(...)` printing would never match. Each program carries the definitions of every earlier block (the examples build on their own helpers: `array-unveil`, `array-squeeze`, `make-u16-array`, …) but not the prose's illustrative re-definitions of library procedures (`index-first`, `array-curry`, …), which would otherwise shadow the implementation under test for every later example. The spec is fetched, or given with `--spec FILE`, and its revision date is printed.

**What the corpus taught the harness** (second commit): the spec's blocks bind free pseudocode variables in `let` bindings, write `=> value` in code position, include fragments and result listings that start with a parenthesis, use `[optional]` signature lines, name identifiers a digit first (`1D-transform`, quoted with `|…|`), call Gambit built-ins without importing (`identity`, `fl+` and friends, shimmed like `fx+`), and sum a lazy array of a billion elements in one example. Thirteen blocks are skipped, each saying why: four result listings (119, 128, 144, 169), three unbalanced fragments (93, 94, 102), two `[optional]` signature lines (48, 166), the PGM reader (158), the `read-char` block (154), the `random` block (152), and the billion-element sum (116). Of the 159 that run, 26 only define helpers and print no value, and 53 are the spec's implementation fragments whose every value is an unbound-variable `ERROR` on both sides; the summary line reports all three counts, so 80 programs compare real values.

| oracle | examples | mismatches |
|---|---|---|
| Gambit 4.9.8 (reference), spec revision 2026-02-18 | 172 (159 run, 13 skipped) | **0** |
| chibi 0.12, re-run on the final head | 172 | 12, all chibi's own (Gambit agrees with Kaappi on every one of the 169): five where chibi dies mid-example (its `(srfi 4)` import warns `u8vector` is undefined; the million-element `u1` sieve never finishes), and seven where its `array-assign!` and relatives return a value where the reference returns an unspecified one |

**One Kaappi finding, filed as #2548.** The example that inspects a `u1` array's body writes its expected value as `#(16 #u16(3895))`, and Kaappi's reader rejected the whole program: it accepts only `#u8(` of SRFI 4's homogeneous-vector literal syntax, though SRFI 4 is on the advertised list. The mode now keeps such annotations as comments, so the corpus runs; the reader gap stands on its own.

### DIFF
--- a/tools/srfi231_diff.py
+++ b/tools/srfi231_diff.py
@@ -719,7 +719,14 @@ PROSE_PRELUDE = """\
         ((number? v) (list 'c (render (real-part v)) (render (imag-part v))))
         ((char? v) (list 'ch (char->integer v)))
         ;; every SRFI 4 body type, so a body an example displays never
-        ;; reaches write (whose float spelling differs)
+        ;; reaches write (whose float spelling differs). The three storage
+        ;; classes that are not SRFI 4 -- f16, c64, c128 -- have no
+        ;; predicate bound by (srfi 4) in either implementation, and a bare
+        ;; (c64vector? v) arm here would raise unbound-variable for EVERY
+        ;; value that reaches it, which try turns into ERROR: the mode would
+        ;; silently degrade to ERROR==ERROR agreement. Their bodies are
+        ;; f32/f64vectors in the reference and here, so they render through
+        ;; the arms below; an arm for a native type would need a guard.
         ((u8vector? v) (list 'u8 (u8vector->list v)))
         ((s8vector? v) (list 's8 (s8vector->list v)))
         ((u16vector? v) (list 'u16 (u16vector->list v)))
@@ -958,11 +965,21 @@ def load_prose(spec):
     # the SRFI's own export surface, from its "Procedure: (name ..." headings
     text = html_mod.unescape(re.sub(r"<[^>]+>", "", page))
     exports = set(re.findall(r"^\s*(?:Procedure|Variable|Parameter|Syntax):\s*\(?([^\s()]+)", text, re.M))
+    def is_code(text):
+        # a block may open with comment lines (the last one does, before
+        # defining the tables its expressions use); judge by the first
+        # line that is neither blank nor a comment
+        for line in text.splitlines():
+            stripped = line.strip()
+            if stripped and not stripped.startswith(";"):
+                return stripped.startswith("(")
+        return False
+
     examples = []
     for i, text in enumerate(pres):
-        if not text.lstrip().startswith("("):
+        if not is_code(text):
             continue
-        nxt = pres[i + 1] if i + 1 < len(pres) and not pres[i + 1].lstrip().startswith("(") else None
+        nxt = pres[i + 1] if i + 1 < len(pres) and not is_code(pres[i + 1]) else None
         examples.append((i, text.strip("\n"), nxt))
     _PROSE = (revision, examples, exports)
     return _PROSE
@@ -1129,6 +1146,8 @@ def main():
 
     if args.mode == "prose":
         revision, examples, _ = load_prose(args.spec)
+        if not 1 <= args.seed <= len(examples):
+            ap.error(f"prose: --seed must be in 1..{len(examples)} (the example number)")
         remaining = len(examples) - args.seed + 1
         args.count = remaining if args.count is None else max(0, min(args.count, remaining))
         print(f"prose: {len(examples)} code examples, {revision}")
@@ -1186,14 +1205,15 @@ def run(args, gen, oracle_cmd, work):
         # values are all ERROR on both sides (a prose fragment with free
         # variables, or a broken prelude) agrees trivially, and a summary
         # of "0 mismatches" must not hide that
-        lines = [ln for ln in normalize(ko).splitlines() if ln and ln != "end"]
-        if any(ln.startswith("skipped ") for ln in lines):
-            skipped += 1
-        elif not lines:
-            silent += 1
-        elif all(ln.endswith("ERROR") for ln in lines):
-            vacuous += 1
         if same:
+            # only an agreeing case can be a vacuous agreement
+            lines = [ln for ln in normalize(ko).splitlines() if ln and ln != "end"]
+            if any(ln.startswith("skipped ") for ln in lines):
+                skipped += 1
+            elif not lines:
+                silent += 1
+            elif all(ln.endswith("ERROR") for ln in lines):
+                vacuous += 1
             continue
         mismatches += 1
         if save is None:
--- a/tools/srfi231_diff.py
+++ b/tools/srfi231_diff.py
@@ -392,7 +392,7 @@ class Chain:
         return "\n".join(self.lines) + "\n"
 
 
-def gen_reshape(seed, classes=None):
+def gen_reshape(seed, classes=None, spec=None):
     rng = random.Random(seed)
     c = Chain(rng)
     ops = [c.op_extract, c.op_translate, c.op_permute, c.op_permute, c.op_reverse,
@@ -484,7 +484,7 @@ def value(rng, cls):
     return likely_valid(rng, cls) if rng.random() < 0.7 else any_value(rng)
 
 
-def gen_storage(seed, classes=None):
+def gen_storage(seed, classes=None, spec=None):
     rng = random.Random(seed)
     cls = rng.choice(classes or CLASSES)
     sc = f"{cls}-storage-class"
@@ -651,7 +651,7 @@ CALLCC_FAMILIES = [
 ]
 
 
-def gen_callcc(seed, classes=None):
+def gen_callcc(seed, classes=None, spec=None):
     rng = random.Random(seed)
     # a small domain: 1-D of 4..6, or 2-D 2x2 / 2x3 / 3x2, lower bounds 0
     shape = rng.choice([[4], [5], [6], [2, 2], [2, 3], [3, 2]])
@@ -718,8 +718,16 @@ PROSE_PRELUDE = """\
                          (else (exact v))))
         ((number? v) (list 'c (render (real-part v)) (render (imag-part v))))
         ((char? v) (list 'ch (char->integer v)))
-        ((u16vector? v) (list 'u16 (u16vector->list v)))
+        ;; every SRFI 4 body type, so a body an example displays never
+        ;; reaches write (whose float spelling differs)
         ((u8vector? v) (list 'u8 (u8vector->list v)))
+        ((s8vector? v) (list 's8 (s8vector->list v)))
+        ((u16vector? v) (list 'u16 (u16vector->list v)))
+        ((s16vector? v) (list 's16 (s16vector->list v)))
+        ((u32vector? v) (list 'u32 (u32vector->list v)))
+        ((s32vector? v) (list 's32 (s32vector->list v)))
+        ((u64vector? v) (list 'u64 (u64vector->list v)))
+        ((s64vector? v) (list 's64 (s64vector->list v)))
         ((f32vector? v) (render (list 'f32 (f32vector->list v))))
         ((f64vector? v) (render (list 'f64 (f64vector->list v))))
         (else v)))
@@ -794,7 +802,7 @@ def tokenize_scheme(src):
         elif c in ")]":
             yield ("close", i, i + 1)
             i += 1
-        elif c == "#" and i + 1 < n and (src[i + 1] == "(" or re.match(r"[usfc]\d+\(", src[i + 1:i + 5])):
+        elif c == "#" and i + 1 < n and (src[i + 1] == "(" or re.match(r"[usfc]\d+\(", src[i + 1:])):
             j = src.index("(", i)
             yield ("open", i, j + 1)
             i = j + 1
@@ -836,8 +844,9 @@ def parse_scheme(src):
         kind, a, b = toks[pos]
         if kind == "datum-comment":
             pos += 1
-            datum()
-            return datum() if pos < len(toks) else None
+            if pos < len(toks):
+                datum()
+            return None
         if kind == "prefix":
             pos += 1
             inner = datum()
@@ -1092,7 +1101,8 @@ def main():
     ap = argparse.ArgumentParser(description=__doc__.split("\n")[0])
     ap.add_argument("mode", choices=sorted(MODES))
     ap.add_argument("--seed", type=int, default=1)
-    ap.add_argument("--count", type=int, default=100)
+    ap.add_argument("--count", type=int, default=None,
+                    help="cases to run (default 100; prose: every remaining example)")
     ap.add_argument("--oracle", choices=["gambit", "chibi"], default="gambit")
     ap.add_argument("--kaappi", default=os.path.join(ROOT, "zig-out", "bin", "kaappi"))
     ap.add_argument("--gsi", default=shutil.which("gsi") or "/opt/homebrew/bin/gsi")
@@ -1119,15 +1129,14 @@ def main():
 
     if args.mode == "prose":
         revision, examples, _ = load_prose(args.spec)
-        if "--count" not in " ".join(sys.argv):
-            args.count = len(examples) - args.seed + 1
-        args.count = max(0, min(args.count, len(examples) - args.seed + 1))
+        remaining = len(examples) - args.seed + 1
+        args.count = remaining if args.count is None else max(0, min(args.count, remaining))
         print(f"prose: {len(examples)} code examples, {revision}")
+    elif args.count is None:
+        args.count = 100
 
     def gen(seed):
-        if args.mode == "prose":
-            return gen_prose(seed, classes, args.spec)
-        return MODES[args.mode](seed, classes)
+        return MODES[args.mode](seed, classes, args.spec)
 
     if args.print:
         sys.stdout.write(gen(args.seed))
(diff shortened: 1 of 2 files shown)


tools/srfi231_diff.py
```diff
@@ -667,7 +686,388 @@ def gen_callcc(seed, classes=None):
...
+    toks = list(tokenize_scheme(src))
+    pos = 0
+
+    def datum():
+        nonlocal pos
+        kind, a, b = toks[pos]
+        if kind == "datum-comment":
+            pos += 1
+            datum()
+            return datum() if pos < len(toks) else None
```
baijum: The datum-comment branch has two broken shapes (latent — the current spec carries no `#;`; I grepped the fetched HTML, zero occurrences):

- `(a b #;c)` — after skipping `c`, `pos` sits on the close token, and the trailing `datum()` eats it as an atom child; the list's `end` then falls back to the open token's end. `parse_scheme("(a b #;c)")` returns `('list', [a, b, <atom ')'>], 0, 1)`, so `node_text` gives just `"("`.
- `(x) #;` — `pos` runs past the end and the unconditional `datum()` on the previous line raises `IndexError`. The context loop in `gen_prose` catches that, but the current block's `parse_scheme(code)` call is outside the `try`, so it would crash the run instead of skipping the block.

Both callers already skip `None` returns, so the branch could simply be `pos += 1` / `if pos < len(toks): datum()` / `return None` — which also preserves the one shape that works today (`(a b #;c d)` → `[a, b, d]`, verified).

tools/srfi231_diff.py
```diff
@@ -667,7 +686,388 @@ def gen_callcc(seed, classes=None):
...
+        elif src.startswith("#;", i):
+            yield ("datum-comment", i, i + 2)
+            i += 2
+        elif c in "([":
+            yield ("open", i, i + 1)
+            i += 1
+        elif c in ")]":
+            yield ("close", i, i + 1)
+            i += 1
+        elif c == "#" and i + 1 < n and (src[i + 1] == "(" or re.match(r"[usfc]\d+\(", src[i + 1:i + 5])):
```
baijum: The 4-character window `[usfc]\d+(` fits the 2-digit storage classes (`#u16(`, `#f32(`) but not a 3-digit one: `#c128(1 2)` tokenizes as atom `#c128` followed by an unrelated list, so at top level it would be wrapped as two separate expressions and compare nothing. The current corpus only carries `#u16(` (2 occurrences, both inside `=>` annotations), so this is latent — but #2548 is exactly about which homogeneous-vector literals a reader accepts, so matching against the open remainder, e.g. `re.match(r"[usfc]\d+\(", src[i + 1:])` and taking `j` from `m.end()`, removes the window width from the list of things to remember.

tools/srfi231_diff.py
```diff
@@ -667,7 +686,388 @@ def gen_callcc(seed, classes=None):
...
+        ((and (real? v) (exact? v)) v)
+        ((real? v) (cond ((nan? v) '(nan))
+                         ((infinite? v) (if (> v 0) '(inf 1) '(inf -1)))
+                         (else (exact v))))
+        ((number? v) (list 'c (render (real-part v)) (render (imag-part v))))
+        ((char? v) (list 'ch (char->integer v)))
+        ((u16vector? v) (list 'u16 (u16vector->list v)))
+        ((u8vector? v) (list 'u8 (u8vector->list v)))
+        ((f32vector? v) (render (list 'f32 (f32vector->list v))))
+        ((f64vector? v) (render (list 'f64 (f64vector->list v))))
```
baijum: Only `u8/u16/f32/f64` bodies are canonicalized; an `s8/s16/s32/s64/u32/u64/c64/c128` body falls through to the final `else` and gets raw `write` — where Gambit's float spelling (`1.` vs `1.0`) would false-positive, the same reason the other classes go through `render`. No current example displays those bodies (the Gambit run is clean), so this is future-proofing for a revised spec: either enumerate the remaining classes or note here that these four are the only bodies the examples print.

tools/srfi231_diff.py
```diff
@@ -715,7 +1117,16 @@ def main():
         if bad:
             sys.exit(f"unknown storage class(es): {', '.join(bad)}; known: {', '.join(CLASSES)}")
 
+    if args.mode == "prose":
+        revision, examples, _ = load_prose(args.spec)
+        if "--count" not in " ".join(sys.argv):
```
baijum: Substring-sniffing `sys.argv` happens to work for both `--count 5` and `--count=5`, but argparse already tells you whether the option was supplied: give `--count` `default=None` and fill in the per-mode default (100 / all examples) after parsing. One less stringly check to reason about.

tools/srfi231_diff.py
```diff
@@ -715,7 +1117,16 @@ def main():
...
+            args.count = len(examples) - args.seed + 1
+        args.count = max(0, min(args.count, len(examples) - args.seed + 1))
+        print(f"prose: {len(examples)} code examples, {revision}")
```
baijum, line 1153: Worth having the run say how much of the corpus actually compared something. Running all 169 programs under Kaappi alone: **48 of the 156 non-skipped programs print only `ERROR` values** — they're the spec's implementation-fragment blocks (seed 4 is `(<= (vector-ref lower-bounds i) ...)`, seed 7 is `(make-interval lower-bounds upper-bounds)`; free variables nothing defines), so both sides agree trivially.

That's a stable corpus fact, but `run()` also counts identical crashes as agreement (equal nonzero exit codes + equal output passes), so a harness regression — a typo'd prelude helper, an import that stops resolving — would make *every* program agree at `ERROR` while the summary still reads `0 mismatches`. A one-line count of programs that printed no non-ERROR value ("48 programs compared nothing") would separate the corpus's fragments from that failure mode.

baijum (commented):
Ran the mode end to end before reading too closely: fetched `srfi-231.html` (revision 2026-02-18, same as the PR) and ran all 169 cases against Gambit 4.9.8 — **0 mismatches, exit 0**, no saved cases. The headline reproduces.

**One correction to the PR description:** the skip count is 13, not 11, so the table's "158 run" is 156. The itemized list also sums to 12 and omits the `read-char` block. Skipped seeds: 48, 166 (`[` signature lines), 93, 94, 102 (unbalanced), 116 (billion-element sum), 119, 128, 144, 169 (result listings), 152 (`random`), 154 (`read-char`), 158 (`read-pgm`). The commit bodies don't carry the wrong numbers, so this only needs a description edit.

Things I checked that came back clean:
- `array->list*` is a real SRFI 231 export (`lib/srfi/231/combinators.sld`), so the guarded `(try ...)` in `render` isn't masking a typo.
- `quote_digit_identifiers` against string literals and comments across the corpus: only comments are touched (`1s` in example 168); the digit-letter sequences inside example 62's display strings (`u16vector-ref`) are lookbehind-protected. A string like `"the 1D case"` in a future revision would get silently bar-quoted, worth remembering if the corpus is ever regenerated.
- The in-place let-body wrapping (reverse order so offsets hold), the named-let `body_from` split, and the `=>`-as-comment handling all generate correct programs on the seeds I inspected.

The inline comments are all latent robustness gaps I verified are *not* triggered by the current corpus, plus one reporting suggestion — nothing here blocks merge from my side. The design choices are good ones: filtering carried context by the spec's own export surface so later examples test the library instead of the prose's pseudocode, and skips that print why instead of vanishing, are what make the run auditable. And #2548 is a genuine reader gap the other suites had no way to notice.

Tiny extra nit not worth an inline thread: `main`'s `gen()` special-cases prose before the `MODES` dispatch, so the `MODES["prose"]` entry is only there for argparse's `choices` — a `functools.partial`-style adapter would keep the map as the single dispatch path.

tools/srfi231_diff.py
baijum: Taken as suggested, in 73d91921: the branch is now `pos += 1` / `if pos < len(toks): datum()` / `return None`. Checked all three shapes: `(a b #;c)` → one list, `(x) #;` → `[(x)]` with no exception, `(a b #;c d)` → `[a, b, d]` unchanged.

tools/srfi231_diff.py
baijum: Done in 73d91921: matched against the open remainder, `j` from the `(`. `#c128(1 2) x` now tokenizes as one list plus an atom.

tools/srfi231_diff.py
baijum: Enumerated all ten in 73d91921 (probed first that both Kaappi's and Gambit's `(srfi 4)` export the `s8/s16/s32/s64/u32/u64` predicates and `->list` conversions — they do), so no body reaches `write`.

tools/srfi231_diff.py
baijum: Done in 73d91921: `--count` defaults to `None` and is filled after parsing — 100, or every remaining example for prose. Also took your top-level nit: every generator now takes `(seed, classes, spec)`, so `MODES` is the single dispatch path and `main` has no prose special case.

tools/srfi231_diff.py
baijum: Added in 73d91921, and it counts a bit more than your 48 because it also separates a third bucket: the summary line now reads `13 skipped, 26 printed no value (definitions only), 52 compared nothing (every value ERROR on both sides)` for the current corpus — 78 of 169 programs compare real values. The 52 vs 48 gap is the criterion: mine also counts a guarded top-level `let` whose bindings failed (`form ERROR`) and the `=>`-only blocks. A broken prelude helper would push the third number toward 156, which is the failure mode you described.

tools/srfi231_diff.py
```diff
@@ -667,7 +686,397 @@ def gen_callcc(seed, classes=None):
...
+        ((pair? v) (cons (render (car v)) (render (cdr v))))
+        ((vector? v) (list 'vec (render (vector->list v))))
+        ((and (real? v) (exact? v)) v)
+        ((real? v) (cond ((nan? v) '(nan))
+                         ((infinite? v) (if (> v 0) '(inf 1) '(inf -1)))
+                         (else (exact v))))
+        ((number? v) (list 'c (render (real-part v)) (render (imag-part v))))
+        ((char? v) (list 'ch (char->integer v)))
+        ;; every SRFI 4 body type, so a body an example displays never
+        ;; reaches write (whose float spelling differs)
```
baijum: Verified the ten SRFI 4 classes are covered, and confirmed the three storage classes in the tool's own `CLASSES` that are *not* SRFI 4 — f16, c64, c128 — have predicates unbound under `(srfi 4)` in both Kaappi and Gambit (probe on both: `s8vector? bound, c64vector? unbound, c128vector? unbound, f16vector? unbound`), so the omission is forced rather than an oversight.

Worth a sentence saying that here, because the naive completion is actively harmful: a bare `((c64vector? v) …)` arm mid-`cond` raises unbound-variable for **every** value that reaches the arm — booleans, symbols, strings — and the enclosing `try` turns each into `ERROR`, silently degrading the mode to ERROR==ERROR agreement. If those bodies ever matter (a future example displaying a c64 body, which would otherwise hit raw `write` and Gambit's float spelling), the arm needs a `bound?`-style guard or a `cond-expand` import, not the plain predicate.

baijum (commented):
Re-verified the whole mode on the new head before commenting: full 169-case Gambit run — **0 mismatches, exit 0** — and the new summary line reproduces the commit message's numbers exactly on my machine: 13 skipped, 26 printed no value, 52 compared nothing, so 78 compare real values.

Every fix from the last round probed directly, all good:
- datum-comment: `(a b #;c d)` → `[a, b, d]`; `(a b #;c)` → one node with the correct full span (was garbled to `"("`); `(x) #;` → no crash.
- tokenizer: `#c128(1 2)` is a single open token now; `#u8x(` is still correctly *not* a vector literal (atom + paren).
- dispatch/count: `reshape`/`callcc --print` generate through `MODES` with the uniform signature; `--count=5` and `--count 5` both take the argparse path now.

The PR-body numbers match what I measure independently (13 skipped with seed numbers, 156 run, 26/52/78), and the skip classification living in `run()` behind the shared `end` marker rather than in the prose generator is the right factoring — it'll say something true for the random modes too.

Two small inline comments (one documentation, one cosmetic) and one flag I could not verify here: **the table's chibi row predates the six new render arms.** If chibi's `(srfi 4)` is missing any of `s8vector?`…`s64vector?`, a mid-`cond` unbound test would flip every value that reaches the arm to `ERROR` under chibi only, which would change the mismatch shape (or mask old ones). I couldn't re-run it — this machine's chibi 0.12 has no `srfi/4.sld` on its module path at all — so the row is worth a re-run against this commit wherever the original chibi measurement was done.

tools/srfi231_diff.py
```diff
@@ -715,8 +1127,16 @@ def main():
         if bad:
             sys.exit(f"unknown storage class(es): {', '.join(bad)}; known: {', '.join(CLASSES)}")
 
+    if args.mode == "prose":
+        revision, examples, _ = load_prose(args.spec)
+        remaining = len(examples) - args.seed + 1
+        args.count = remaining if args.count is None else max(0, min(args.count, remaining))
```
baijum, line 1152: With `--seed` past the end of the corpus and no `--count`, `remaining` is negative and flows through the `None` branch unclamped: `prose --seed 171` prints `prose: -1 cases from seed 171, oracle gambit, 0 mismatches` (the range is empty so nothing runs — cosmetic only). `args.count = max(0, remaining) if args.count is None else max(0, min(args.count, remaining))` would keep the summary honest.

tools/srfi231_diff.py
```diff
@@ -667,7 +686,397 @@ def gen_callcc(seed, classes=None):
...
+            for x in re.findall(r"<pre[^>]*>(.*?)</pre>", page, re.S)]
+    rev = re.findall(r"(\d{4}-\d{2}-\d{2}) \(([^)]*)\)", page)
+    revision = (f"spec revision {rev[-1][0]} ({re.sub(r'<[^>]+>', '', rev[-1][1])})"
+                if rev else "spec revision unknown")
+    # the SRFI's own export surface, from its "Procedure: (name ..." headings
+    text = html_mod.unescape(re.sub(r"<[^>]+>", "", page))
+    exports = set(re.findall(r"^\s*(?:Procedure|Variable|Parameter|Syntax):\s*\(?([^\s()]+)", text, re.M))
+    examples = []
+    for i, text in enumerate(pres):
+        if not text.lstrip().startswith("("):
```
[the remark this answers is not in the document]
baijum: Confirmed and fixed in 2b92b051. Three blocks open with comment lines before their first form (spec `<pre>` 174, 196 and 199 — the last defines `TABLE1`/`TABLE2`), and the first-character test dropped all three and paired one as the previous block's output. Classification now looks at the first line that is neither blank nor a comment, for code selection and output pairing alike. The corpus is 172 examples, still 0 mismatches against Gambit.

tools/srfi231_diff.py
```diff
@@ -715,8 +1127,16 @@ def main():
         if bad:
             sys.exit(f"unknown storage class(es): {', '.join(bad)}; known: {', '.join(CLASSES)}")
 
+    if args.mode == "prose":
+        revision, examples, _ = load_prose(args.spec)
+        remaining = len(examples) - args.seed + 1
```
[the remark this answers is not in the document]
baijum: Done in 2b92b051: for prose, a seed outside `1..N` is an `ap.error` naming the range; `remaining` is computed only after that.

tools/srfi231_diff.py
```diff
@@ -759,6 +1182,17 @@ def run(args, gen, oracle_cmd, work):
...
+        # equal output is not the same as a comparison: a program whose
+        # values are all ERROR on both sides (a prose fragment with free
+        # variables, or a broken prelude) agrees trivially, and a summary
+        # of "0 mismatches" must not hide that
+        lines = [ln for ln in normalize(ko).splitlines() if ln and ln != "end"]
+        if any(ln.startswith("skipped ") for ln in lines):
+            skipped += 1
+        elif not lines:
+            silent += 1
+        elif all(ln.endswith("ERROR") for ln in lines):
```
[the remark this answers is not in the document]
baijum: Done in 2b92b051: the three buckets (skipped, definition-only, all-ERROR) are counted inside the `same` branch only, so a mismatching case is never reported as a vacuous agreement.

tools/srfi231_diff.py
baijum: Written into `render` in 2b92b051, with the failure mode spelled out: a bare `(c64vector? v)` arm would raise for every value reaching it and `try` would turn the mode into ERROR==ERROR agreement. It also notes that c64/c128 bodies are f32/f64vectors in both implementations, so they already render through those arms; only a native type would need a guard.

tools/srfi231_diff.py
baijum: The seed validation from the CodeRabbit thread covers this: `prose --seed 171` (now past 172) is an argparse error rather than a `-1 cases` summary. In 2b92b051.

PR #2551, opened 2026-09-07
Report whether a failed REPL pty scenario's output arrives late

### LINKED ISSUE #2550: CI flake: repl-mouse-click-2264.sh 'ahead' scenario times out mid-echo on the ubuntu Debug job
**Where**

CI, `test (ubuntu-latest, Debug)`, Scheme suites step, on PR #2549 (a change to `tools/srfi231_diff.py` and `docs/dev/testing.md` only — no Zig, no Scheme, no CI config): https://github.com/kaappi/kaappi/actions/runs/34141905300/job/101805621497

**What failed**

`tests/scheme/smoke/repl-mouse-click-2264.sh`, the `ahead` scenario only (the four click cases passed):

```
FAIL b'ahead': expected b'3 then (9 8)' in
b'kaappi> (kaappi> (+kaappi> (+ kaappi> (+ 1kaappi> (+ 1 2kaappi> (+ 1 2)kaappi> (+ 1 2)\n3\nkaappi> kaappi> (kaappi> (lkaappi> (likaappi> (liskaappi> (list
```

The scenario sends `(+ 1 2)\r` and immediately `(list 9 8)\r` (typed while the first form evaluates), then waits up to 20 s each for `\n3\n` and `\n(9 8)\n`. The first result arrived; the echo of the second form stops at `(list` and nothing more was seen within the 20 s wait plus the idle. The `(` was not eaten by the DSR anchor query (that is what the scenario guards against, kaappi#2264's pushback), so this is a different shape: the remaining ` 9 8)\r` bytes were neither echoed nor evaluated within the window.

**Not reproducible locally** (macOS aarch64, same commit): 5 runs on the ReleaseSafe binary, 5 on a Debug build on an idle machine, 5 on the Debug build under 6 busy loops — 15 of 15 pass, all five checks each. The Debug job on the last 10 `main` runs passed too. One occurrence so far.

**Two readings, and what would separate them**

- A timing flake in the driver: the Debug build (~500x slower on some paths) plus a loaded runner pushed the echo past the 20 s window. If so the remaining bytes would have shown up later. The test could, on failure, keep pumping for an extra interval and print whether the tail eventually arrived — that one line would decide this on the next occurrence.
- Genuine loss or stall of type-ahead bytes after the DSR reply in the isocline reader (the pushback path of kaappi#2264 / PATCHES.md #5). That would be `priority: high` — lost input in the REPL is user-reachable data loss — but nothing observed so far shows it.

Filed at `low` as a CI flake with an escalation criterion: if a second occurrence shows the tail never arriving, or it reproduces interactively, raise to `high` and treat it as a reader bug.

**Version**

main at 0555079e (v0.26.2 line); the job's Debug build.

baijum:
## Summary

Closes #2550.

The `ahead` scenario of `tests/scheme/smoke/repl-mouse-click-2264.sh` failed once on the ubuntu Debug CI job (PR #2549) with the echo of the second form stalled after `(list` — and nothing more in 20 s of continuous pumping plus the idle drain. The issue identifies two candidate causes that the captured failure buffer cannot distinguish:

1. a **timing flake** — a loaded runner plus the ~500x-slower Debug build pushed the echo past the 20 s waits;
2. **genuine loss/stall of type-ahead bytes** in the isocline pushback path (kaappi#2264 patch 5) — which would be `priority: high` as user-reachable input loss.

This PR implements the discriminating observation the issue asks for: when any scenario fails, the driver now keeps the pty open and keeps pumping for **30 more seconds** (still answering `ESC[6n` anchor queries, stopping early once every missing marker has shown up) before reporting. The failure output then gains a one-line `TAIL-CHECK` verdict:

- `arrived during the 30s drain after the failure — late output on a loaded runner, not lost bytes` → it was a slow-output flake;
- `never arrived, even after 30s more pumping — candidate byte loss in the reader; this is the escalation criterion in issue #2550` → raise to `high` and treat as a reader bug.

All failure paths (click cases, off-mode, wrapped-row, `ahead`, `garble`) record through the same check, so the verdict covers every scenario, not only the one that flaked.

## Verification

- All 5 scenarios pass on a ReleaseSafe build and on a Debug build (the CI configuration that flaked).
- Both verdict branches exercised by running extracted driver copies with an unreachable needle (→ `never arrived`) and with a zero-timeout wait on a still-evaluating form (→ `arrived during the drain`).
- `zig build test` passes.

## Checklist

- [x] `zig build test` passes
- [x] `zig fmt --check src/` passes (no Zig changes)
- [x] Scheme test suites pass (`bash tests/scheme/run-all.sh` not needed — driver-only change; the affected script itself passes in full on both build modes)
- [ ] New tests added for new behavior (this *is* test-driver diagnostics; both branches verified manually as above)
- [ ] Documentation updated (if applicable — none needed)
- [x] Commit body explains *why* — release notes are written from it at release time; do not edit `CHANGELOG.md`

### DIFF
--- a/tests/scheme/smoke/repl-mouse-click-2264.sh
+++ b/tests/scheme/smoke/repl-mouse-click-2264.sh
@@ -195,6 +195,34 @@ def send(data):
     os.write(fd, data)
     time.sleep(0.05)
 
+# Issue #2550: when a scenario fails, a 20 s wait that a loaded, Debug-speed
+# runner outran and bytes that were lost look identical in the captured
+# output. Before reporting, keep the pty open and keep pumping for this many
+# more seconds (answering anchor queries as they come, stopping early once
+# everything missing has shown up). The failure line then says which it was:
+# a marker that arrived during the drain is a slow-output flake; one that
+# never arrived is candidate reader byte loss — that report is the issue's
+# escalation criterion, so do not remove it.
+DRAIN_S = 30
+
+def record(failed, expect, mark, needles):
+    produced = seen(mark)
+    missing = [n for n in needles if n not in produced]
+    late, never = [], []
+    if missing:
+        end = time.time() + DRAIN_S
+        while time.time() < end:
+            if all(n in seen(mark) for n in missing):
+                break
+            answer_dsr()
+            if eof:
+                break
+            pump(0.5)
+        final = seen(mark)
+        late = [n for n in missing if n in final]
+        never = [n for n in missing if n not in final]
+    failures.append((failed, expect, produced, late, never))
+
 if not wait_for(b'kaappi> ', 0, 25):
     # Two very different things look alike here, and conflating them is how a
     # test goes quietly green: a REPL that wrote *nothing at all* did not run,
@@ -225,7 +253,7 @@ if scenario == 'ahead':
     ok2 = wait_for(b'\n(9 8)\n', mark, 20)
     idle()
     if not (ok1 and ok2):
-        failures.append((b'ahead', b'3 then (9 8)', seen(mark)))
+        record(b'ahead', b'3 then (9 8)', mark, [b'\n3\n', b'\n(9 8)\n'])
 elif scenario == 'garble':
     # Hostile terminal report (maintainer review, memory safety): a CSI with
     # 40 digits/semicolons not terminated by R, arriving while the DSR
@@ -244,7 +272,7 @@ elif scenario == 'garble':
     ok2 = wait_for(b'\n42\n', mark, 20)
     idle()
     if not (ok1 and ok2):
-        failures.append((b'garble', b'3 then 42 (no crash)', seen(mark)))
+        record(b'garble', b'3 then 42 (no crash)', mark, [b'\n3\n', b'\n42\n'])
 else:
     for send_bytes, echo, clk, editkey, expect, reject in cases:
         mark = len(buf)
@@ -253,7 +281,7 @@ else:
         # and typing on while the REPL is still evaluating puts the bytes in the
         # tty's canonical buffer, where the click then arrives as a literal ESC.
         if not wait_for(echo, mark, 20):
-            failures.append((send_bytes, expect, seen(mark)))
+            record(send_bytes, expect, mark, [echo])
             idle()
             continue
         idle(0.3)
@@ -272,7 +300,7 @@ else:
         # substring check.
         produced = seen(mark)
         if not ok or (reject is not None and (b'\n' + reject + b'\n') in produced):
-            failures.append((send_bytes, expect, produced))
+            record(send_bytes, expect, mark, [b'\n' + expect + b'\n'])
 
 send(b',quit\r')
 while pump(1.0):
@@ -283,8 +311,18 @@ try:
 except OSError:
     pass
 
-for failed, expect, produced in failures:
+for failed, expect, produced, late, never in failures:
     sys.stdout.write('FAIL %r: expected %r in\n%r\n\n' % (failed, expect, produced))
+    if late:
+        sys.stdout.write(
+            'TAIL-CHECK: %r arrived during the %ds drain after the failure'
+            ' — late output on a loaded runner, not lost bytes\n\n'
+            % (late, DRAIN_S))
+    if never:
+        sys.stdout.write(
+            'TAIL-CHECK: %r never arrived, even after %ds more pumping'
+            ' — candidate byte loss in the reader; this is the escalation'
+            ' criterion in issue #2550\n\n' % (never, DRAIN_S))
 sys.exit(1 if failures else 0)
 PY
 
--- a/tests/scheme/smoke/repl-mouse-click-2264.sh
+++ b/tests/scheme/smoke/repl-mouse-click-2264.sh
@@ -241,14 +241,16 @@ def record(failed, expect, mark, needles):
                          ' drain -- late output on a loaded runner, not lost'
                          ' bytes' % (late,))
         if never and eof:
-            # The child is gone: the garble scenario's expected failure mode
-            # is exactly this, so name it instead of claiming a 30 s silence
-            # that never happened. eof, not a flag set inside the loop: the
-            # final pump(0.5) can be the one that observes the exit, after
-            # which the deadline check ends the loop without ever looking.
-            lines.append('TAIL-CHECK: %r never arrived and the REPL exited'
-                         ' during the drain -- not a 30 s silence, see the'
-                         ' buffer above' % (never,))
+            # The child is gone -- during the drain, or already when the
+            # failure was detected (a failed send sets eof too): the garble
+            # scenario's expected failure mode is exactly this, so name it
+            # instead of claiming a 30 s silence that never happened. eof,
+            # not a flag set inside the loop: the final pump(0.5) can be
+            # the one that observes the exit, after which the deadline
+            # check ends the loop without ever looking.
+            lines.append('TAIL-CHECK: %r never arrived; the REPL had exited'
+                         ' by the end of the drain -- not a 30 s silence,'
+                         ' see the buffer above' % (never,))
         if never and not eof:
             # Still missing and the child is alive. Passive pumping cannot
             # separate bytes the reader dropped from a reader wedged shut,
@@ -264,7 +266,12 @@ def record(failed, expect, mark, needles):
             idle(0.3)
             send(b'(+ 40 2)\r')
             pend = time.monotonic() + 5
-            while time.monotonic() < pend:
+            while time.monotonic() < pend and not eof:
+                # not eof, matching the drain loop: a dead master's select
+                # reports readable at once and read raises, so pump returns
+                # immediately -- without the check this loop would spin at
+                # full speed for the remaining seconds and then misreport
+                # an exited REPL as a wedged reader.
                 if b'\n42\n' in seen(p0):
                     break
                 pump(0.5)
@@ -276,6 +283,9 @@ def record(failed, expect, mark, needles):
                 lines.append('PROBE: ctrl-C then (+ 40 2) printed 42 -- the'
                              ' reader is alive; the missing bytes were'
                              ' dropped')
+            elif eof:
+                lines.append('PROBE: the REPL exited during the probe --'
+                             ' see the buffer above')
             else:
                 lines.append('PROBE: no answer to ctrl-C then (+ 40 2)'
                              ' within 5 s -- the reader is wedged, not'
--- a/tests/scheme/smoke/repl-mouse-click-2264.sh
+++ b/tests/scheme/smoke/repl-mouse-click-2264.sh
@@ -151,13 +151,17 @@ def pump(timeout):
 def answer_dsr():
     """Answer every ESC[6n anchor query the child sent, the way a terminal
     emulator would. The child blocks ~400ms waiting, so we answer on sight."""
-    global answered_dsr
+    global answered_dsr, eof
     while True:
         idx = buf.find(b'\x1b[6n', answered_dsr)
         if idx < 0:
             return
         answered_dsr = idx + 4
-        os.write(fd, DSR)
+        try:
+            os.write(fd, DSR)
+        except OSError:
+            eof = True        # the child is gone; same handling as a dead read
+            return
 
 def seen(mark=0):
     """Everything the child has written since *raw* byte `mark`, escapes
@@ -192,7 +196,16 @@ def idle(quiet=0.5, limit=5.0):
             return
 
 def send(data):
-    os.write(fd, data)
+    global eof
+    try:
+        os.write(fd, data)
+    except OSError:
+        # A child that died mid-scenario (the garble case's expected failure
+        # mode) leaves the master unwritable: writing raises EIO. Mark eof,
+        # exactly like a dead read, and let the caller's wait end in the
+        # eof checks — no call site needs a guard of its own.
+        eof = True
+        return
     time.sleep(0.05)
 
 # Issue #2550: when a scenario fails, a 20 s wait that a loaded, Debug-speed
@@ -212,34 +225,31 @@ def record(failed, expect, mark, needles):
     lines = []
     if missing:
         n0 = len(buf)
-        died = False
         end = time.monotonic() + DRAIN_S
         while time.monotonic() < end:
             if all(n in seen(mark) for n in missing):
                 break
             answer_dsr()
             if eof:
-                died = True
                 break
             pump(0.5)
         final = seen(mark)
         late = [n for n in missing if n in final]
         never = [n for n in missing if n not in final]
-        tail = ansi.sub(b'', buf[n0:])
-        if tail:
-            lines.append('DRAIN-TAIL: %r' % (tail,))
         if late:
             lines.append('TAIL-CHECK: %r arrived during the post-failure'
                          ' drain -- late output on a loaded runner, not lost'
                          ' bytes' % (late,))
-        if never and died:
+        if never and eof:
             # The child is gone: the garble scenario's expected failure mode
             # is exactly this, so name it instead of claiming a 30 s silence
-            # that never happened.
+            # that never happened. eof, not a flag set inside the loop: the
+            # final pump(0.5) can be the one that observes the exit, after
+            # which the deadline check ends the loop without ever looking.
             lines.append('TAIL-CHECK: %r never arrived and the REPL exited'
                          ' during the drain -- not a 30 s silence, see the'
                          ' buffer above' % (never,))
-        if never and not died:
+        if never and not eof:
             # Still missing and the child is alive. Passive pumping cannot
             # separate bytes the reader dropped from a reader wedged shut,
             # so probe it: ctrl-C clears whatever open form the editor
@@ -270,6 +280,14 @@ def record(failed, expect, mark, needles):
                 lines.append('PROBE: no answer to ctrl-C then (+ 40 2)'
                              ' within 5 s -- the reader is wedged, not'
                              ' merely lossy')
+        # Everything that arrived after the failure was detected: the drain
+        # and, when the probe ran, its exchange. That output is the evidence
+        # for the verdicts above -- a ctrl-C redraw showing what the buffer
+        # still held, the missing result turning up, or silence -- so print
+        # it rather than discard it.
+        tail = ansi.sub(b'', buf[n0:])
+        if tail:
+            lines.append('DRAIN-TAIL: %r' % (tail,))
     failures.append((failed, expect, produced, lines))
 
 if not wait_for(b'kaappi> ', 0, 25):
@@ -356,17 +374,14 @@ else:
             # #2550 drain would only misreport it as byte loss -- and this
             # is exactly the failure shape a click-mapping or repl.mouse
             # regression produces, the thing this script exists to catch.
-            failures.append((send_bytes, expect, produced, []))
+            failures.append((send_bytes, expect, produced, [
+                'VERDICT: the no-click result %r printed -- the form arrived'
+                ' and was evaluated, the click did not reposition (not a'
+                ' #2550 timing question)' % (reject,)]))
         elif not ok:
             record(send_bytes, expect, mark, [b'\n' + expect + b'\n'])
 
-try:
-    send(b',quit\r')
-except OSError:
-    # A child that died mid-scenario (the garble case's expected failure
-    # mode) leaves the master unwritable: writing raises EIO. The failure
-    # report below is the whole point, so absorb this and carry on.
-    pass
+send(b',quit\r')
 while pump(1.0):
     pass
 try:
--- a/tests/scheme/smoke/repl-mouse-click-2264.sh
+++ b/tests/scheme/smoke/repl-mouse-click-2264.sh
@@ -199,29 +199,78 @@ def send(data):
 # runner outran and bytes that were lost look identical in the captured
 # output. Before reporting, keep the pty open and keep pumping for this many
 # more seconds (answering anchor queries as they come, stopping early once
-# everything missing has shown up). The failure line then says which it was:
+# everything missing has shown up). The verdict lines then say which it was:
 # a marker that arrived during the drain is a slow-output flake; one that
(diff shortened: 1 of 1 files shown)


tests/scheme/smoke/repl-mouse-click-2264.sh
```diff
@@ -272,7 +300,7 @@ else:
         # substring check.
         produced = seen(mark)
         if not ok or (reject is not None and (b'\n' + reject + b'\n') in produced):
-            failures.append((send_bytes, expect, produced))
+            record(send_bytes, expect, mark, [b'\n' + expect + b'\n'])
```
baijum, line 392: **Misdiagnosis on the reject path.** When the click did not reposition, the REPL prints the `reject` line (`(1 2 3 9)`, `(1 2 3 49)`, `(1 2 3 94)`, …) — the form was received in full and evaluated. `expect` can then never appear, so `record` drains for 30 s and prints:

```
TAIL-CHECK: [b'\n(1 2 3 4)\n'] never arrived, even after 30s more pumping — candidate byte loss in the reader; this is the escalation criterion in issue #2550
```

(reproduced by running the driver in `off` mode with `repl.mouse: true` in the home). That is the exact opposite of what happened, and it is the failure shape a regression in PATCH 5's click arithmetic or the `repl.mouse` gate produces — the thing this script exists to catch. The "never arrived" verdict only means something when the REPL produced *no* result for the form; a result line, right or wrong, settles the #2550 question as "bytes arrived".

Suggest keeping the reject case out of the drain and saying what it is:

```suggestion
        produced = seen(mark)
        rejected = reject is not None and (b'\n' + reject + b'\n') in produced
        if rejected:
            # The REPL evaluated the form and printed the no-click result: the
            # bytes arrived, the click just did not reposition. Nothing is late
            # or lost here, so the #2550 drain would only misreport it.
            failures.append((send_bytes, expect, produced, [], []))
        elif not ok:
            record(send_bytes, expect, mark, [b'\n' + expect + b'\n'])
```

Same set of failures as before (`not ok or rejected`), only the reject case no longer waits 30 s or claims byte loss.

tests/scheme/smoke/repl-mouse-click-2264.sh
```diff
@@ -195,6 +195,34 @@ def send(data):
...
+    missing = [n for n in needles if n not in produced]
+    late, never = [], []
+    if missing:
+        end = time.time() + DRAIN_S
+        while time.time() < end:
+            if all(n in seen(mark) for n in missing):
+                break
+            answer_dsr()
+            if eof:
+                break
```
baijum, line 234: **Child exit during the drain is reported as "never arrived, even after 30s more pumping".** If the REPL dies here — the `garble` scenario is a memory-safety case, so a crash is its expected failure mode — the loop breaks on the first iteration, `never` is populated, and the report says the bytes were pumped for 30 s and are "candidate byte loss in the reader". Neither half is true: the drain lasted milliseconds and the process is gone.

Cheap to make honest: note the exit and print it instead of the byte-loss line, e.g.

```python
    died = False
    ...
            if eof:
                died = True
                break
    ...
    failures.append((failed, expect, produced, late, never, died))
```

and in the report, when `died`: `TAIL-CHECK: the REPL exited during the drain — see the buffer above; not a timing question`. Then the "never arrived" text is only printed for a REPL that was still alive and silent, which is what #2550's criterion describes.

tests/scheme/smoke/repl-mouse-click-2264.sh
```diff
@@ -283,8 +311,18 @@ try:
 except OSError:
     pass
 
-for failed, expect, produced in failures:
+for failed, expect, produced, late, never in failures:
     sys.stdout.write('FAIL %r: expected %r in\n%r\n\n' % (failed, expect, produced))
```
baijum, line 404: **What arrived during the drain is discarded.** `produced` is the pre-drain snapshot, so on a `never` verdict the reader sees the same buffer the issue already had and nothing of the 30 s that followed. But that tail is the evidence: for the #2550 `ahead` shape, a lost `9 8)\r` leaves the editor sitting on an open `(list` form (the completeness callback keeps prompting, so the continuation prompt redraws would show up), a stalled reader shows nothing at all, and an error message would name a third cause. All three print "never arrived" identically right now.

Suggest returning the drain tail from `record` (`final[len(produced):]`, or just `final`) and printing it when non-empty, e.g. `DRAIN-TAIL: %r`. It costs nothing and turns the one-line verdict into something the next occurrence can actually be triaged from.

tests/scheme/smoke/repl-mouse-click-2264.sh
```diff
@@ -272,7 +300,7 @@ else:
         # substring check.
         produced = seen(mark)
         if not ok or (reject is not None and (b'\n' + reject + b'\n') in produced):
-            failures.append((send_bytes, expect, produced))
+            record(send_bytes, expect, mark, [b'\n' + expect + b'\n'])
 
 send(b',quit\r')
```
baijum, line 394: **Optional — an active probe would separate "lost" from "stalled".** After a `never` outcome the REPL is in one of two states: alive but holding an open `(list` form (bytes lost after the DSR reply — note `,quit\r` here then lands *inside* that open form and the REPL never quits on its own; the `os.close(fd)` below is what ends it, as before), or wedged in the reader (a stall). Thirty seconds of passive pumping cannot tell them apart. Sending `\x03` and a probe like `(+ 40 2)\r` (the same recovery the `garble` scenario already uses) and waiting a few seconds for `\n42\n` would: a `42` means the reader is fine and the bytes were dropped, which is the `high` case; no `42` means a hang. Not blocking — the passive drain already answers the issue's first question — but it is the second question, and it is a four-line addition to `record`.

tests/scheme/smoke/repl-mouse-click-2264.sh
```diff
@@ -283,8 +311,18 @@ try:
 except OSError:
     pass
 
-for failed, expect, produced in failures:
+for failed, expect, produced, late, never in failures:
     sys.stdout.write('FAIL %r: expected %r in\n%r\n\n' % (failed, expect, produced))
+    if late:
+        sys.stdout.write(
+            'TAIL-CHECK: %r arrived during the %ds drain after the failure'
+            ' — late output on a loaded runner, not lost bytes\n\n'
```
baijum: nit: the two verdict strings carry a U+2014 em dash, the only non-ASCII output the driver writes. Python ≥ 3.7 coerces a C/POSIX locale to UTF-8 so CI is fine, but under any non-UTF-8 locale `sys.stdout.write` raises `UnicodeEncodeError` — on exactly the one path this PR exists to print, replacing the verdict with a traceback. Two `--` would make the report unconditional.

baijum (commented):
Reviewed on a ReleaseSafe build of ea4d3860: all five scenarios pass (15 s wall), and I ran an extracted copy of the driver through the new failure path. The drain is bounded and fits the budgets (worst case per `run_case` is ~137 s: startup 25 + 2 × (20 echo wait + 30 drain + 5 idle), against `SHELL_TIMEOUT` 300 s local / 600–900 s in CI), and the "late" branch is the discriminating observation #2550 asked for.

One real problem, though, and it sits on the path this test primarily guards: **a click-mapping regression is reported with the #2550 escalation text.** I simulated one by running the driver in `off` mode against a home with `repl.mouse: true` (so the click *does* reposition and `OFF_CASE`'s reject line prints). Output, after a 30 s drain:

```
FAIL b'(list 1 2\r3 4)': expected b'(1 2 3 4)' in
b'kaappi> (kaappi> ... 3 94)\n(1 2 3 94)\nkaappi> '

TAIL-CHECK: [b'\n(1 2 3 4)\n'] never arrived, even after 30s more pumping — candidate byte loss in the reader; this is the escalation criterion in issue #2550
```

The bytes demonstrably arrived and were evaluated; the verdict says the opposite and points the triager at isocline's pushback path. Details and a fix inline, plus two smaller diagnostic gaps (the `eof` case, and the drained bytes being discarded) and an optional stronger probe. Worth fixing the first before merge; the rest are your call.

baijum:
Addressed everything in 78b9f02c — all five inline comments plus both CodeRabbit findings. Each verdict branch was re-validated by running extracted driver copies; outputs below are from those runs.

**Reject-path misdiagnosis (the pre-merge one).** The reject case now records the failure directly instead of entering the drain: when the no-click reject line printed, the form demonstrably arrived and was evaluated, so no verdict is emitted at all. Reproduced with the same setup (driver `off` mode, `repl.mouse: true` in the home): the run now fails in ~23 s (the normal case-execution time, no 30 s drain appended) and prints only the FAIL line — no `TAIL-CHECK`, no byte-loss claim.

**Child exit during the drain.** The stop reason is tracked; the verdict for that case is now `TAIL-CHECK: ... never arrived and the REPL exited during the drain -- not a 30 s silence, see the buffer above`. Validated with a wrapper child that SIGKILLs the REPL 8 s into the run. This exposed a latent driver bug the verdict depends on: writing the trailing `,quit` to the dead pty raised EIO and crashed the driver *before any report printed* — pre-existing, but fatal to this exact path, so the exit sequence now absorbs the EIO.

**Drained bytes discarded.** `record` now captures what arrived after the failure and prints it as `DRAIN-TAIL: %r` when non-empty. Visible in the late-arrival validation: the tail shows the editor's per-keystroke redraws up to the late result.

**Active probe (lost vs wedged).** Implemented as suggested: after a full silent drain with the child alive, send ctrl-C (clears any open form, the garble scenario's recovery), then `(+ 40 2)`, wait 5 s. Validated both outcomes: an unreachable marker with a healthy REPL → `PROBE: ctrl-C then (+ 40 2) printed 42 -- the reader is alive; the missing bytes were dropped`; with the probe's check needle made bogus → `PROBE: no answer to ctrl-C then (+ 40 2) within 5 s -- the reader is wedged, not merely lossy`. Probe output lands before the next scenario takes its mark, so later assertions are unaffected.

**Monotonic clock.** The drain deadline and loop now use `time.monotonic()`; the probe window too.

**Em dashes.** Every string that reaches `sys.stdout.write` is now pure ASCII (`--`); verified by scanning the driver — the remaining U+2014s are in comments only, which are never printed.

On the early-EOF "regression case": there is no in-repo harness for the driver's internals (it is a heredoc-driven pty script), and no scenario can crash the current REPL on demand — so the case is the manual validation above, which I've recorded here rather than as a committed test. The full five-scenario suite is green on ReleaseSafe and Debug after the changes.

tests/scheme/smoke/repl-mouse-click-2264.sh
```diff
@@ -195,6 +195,83 @@ def send(data):
...
+                         ' drain -- late output on a loaded runner, not lost'
+                         ' bytes' % (late,))
+        if never and died:
+            # The child is gone: the garble scenario's expected failure mode
+            # is exactly this, so name it instead of claiming a 30 s silence
+            # that never happened.
+            lines.append('TAIL-CHECK: %r never arrived and the REPL exited'
+                         ' during the drain -- not a 30 s silence, see the'
+                         ' buffer above' % (never,))
+        if never and not died:
```
baijum: **`died` can be `False` while `eof` is `True`, and then the probe crashes the driver.** The loop only sets `died` when it observes `eof` *between* iterations. If the final `pump(0.5)` is the one that hits EOF and the deadline has passed by the time it returns, the `while` condition fails first and the flag is never set. `never and not died` then runs the probe, and `send(b'\x03')` writes to a pty whose slave is gone — that raises `EIO` (checked on this Mac: `os.write` on the master after the child exits gives `OSError: [Errno 5]`), the same failure the `,quit` guard below was added for, but here it escapes `record` as a traceback and nothing is reported. The window is the last 0.5 s of the 30 s drain, and the crash-during-drain case is the `garble` scenario's expected failure mode, so this is exactly the path the commit set out to make honest.

The flag is redundant with the global — test it after the loop instead:

```suggestion
        if never and not eof:
```

and the `died = False` / `died = True` lines go with it (`never and died` becomes `never and eof`). Independently, the probe's two `send`s and its `idle(0.3)` sit after up to 30 s during which the child can die between any two of them; the robust shape is for `send` (and `answer_dsr`'s write) to absorb `OSError`, set `eof`, and return, so no call site needs its own guard — that would subsume the `try` around `,quit` as well.

tests/scheme/smoke/repl-mouse-click-2264.sh
```diff
@@ -195,6 +195,83 @@ def send(data):
...
+                break
+            answer_dsr()
+            if eof:
+                died = True
+                break
+            pump(0.5)
+        final = seen(mark)
+        late = [n for n in missing if n in final]
+        never = [n for n in missing if n not in final]
+        tail = ansi.sub(b'', buf[n0:])
```
baijum: **The probe's exchange is not in the report.** `tail` is taken here, before the probe, and `seen(p0)` is only consulted for the `42` test. But the probe's output is the evidence for the `PROBE` verdict: when ctrl-C reaches an alive editor it redraws, and that redraw shows what the buffer held — `(list` alone (the tail was dropped in flight), `(list 9 8)` never submitted (the `\r` went missing, a different bug), or leftover digits (the pushback replayed something). All three currently print the same `PROBE: ... 42 -- the missing bytes were dropped`, and the bytes that would distinguish them are thrown away.

Cheapest fix is to compute the tail after the probe, i.e. move this line (and the `DRAIN-TAIL` append) to the end of `record`, so `buf[n0:]` covers both windows; or add a `PROBE-TAIL: %r` line from `seen(p0)` next to the `PROBE` verdict. In my bogus-needle run the report had no `DRAIN-TAIL` at all (nothing arrived during the drain) and no trace of the probe, so a real occurrence would give the triager the verdict and nothing to check it against.

tests/scheme/smoke/repl-mouse-click-2264.sh
```diff
@@ -271,10 +348,25 @@ else:
...
-            failures.append((send_bytes, expect, produced))
+        rejected = reject is not None and (b'\n' + reject + b'\n') in produced
+        if rejected:
+            # The REPL evaluated the form and printed the no-click result:
+            # the bytes all arrived and were evaluated, the click just did
+            # not reposition. Nothing can be late or lost here, so the
+            # #2550 drain would only misreport it as byte loss -- and this
+            # is exactly the failure shape a click-mapping or repl.mouse
+            # regression produces, the thing this script exists to catch.
+            failures.append((send_bytes, expect, produced, []))
```
baijum: nit: this path appends no verdict line, so the output is a bare `FAIL ... expected (1 2 3 4) in <buffer>` and the reader has to spot the `(1 2 3 94)` in the buffer and know what it means. The code comment above explains it well; a one-liner in the report would too, e.g. `VERDICT: the no-click result %r printed -- the form arrived and was evaluated, the click did not reposition (not a #2550 timing question)`. Then every failure shape names itself, which is the point of the whole change.

baijum (commented):
Re-reviewed 78b9f02c. All five earlier findings are addressed and I re-verified the two that mattered most on a ReleaseSafe build: the simulated click regression (`off` mode with `repl.mouse: true`) now fails fast in 23 s with no drain and no byte-loss claim, and a live reader with a bogus needle runs the full drain and reports `PROBE: ... printed 42`. The full five-scenario script is green (14 s). The `,quit` EIO catch is a good find, and `time.monotonic` is right.

Three things left, one of which is a crash on a path this commit specifically set out to cover:

- **`died` misses an `eof` that lands on the drain's last `pump`** — the probe then writes to a dead pty and the driver dies with `EIO` before printing anything (inline at line 242).
- **The probe's own output is discarded** — `tail` is cut before the probe runs, so the redraw that shows what the editor's buffer held when ctrl-C hit it never reaches the report (inline at line 228).
- nit: the reject path prints a bare `FAIL` with no verdict line, so the reader has to infer "the no-click result printed" from the buffer (inline at line 359).

Fix the first and this is good to merge from my side; the other two are cheap but optional.

baijum:
All three addressed in bfb980e0.

**The EIO window.** `died` is gone — the branches test the `eof` global after the loop, which has no window: whether the exit was observed by an intermediate `pump` or by the final one after which only the deadline check runs, `eof` is already set when the probe branch is reached, so the probe is skipped and the exited-during-drain verdict prints. Going further on your robust-shape suggestion: `send` and `answer_dsr` now absorb `OSError` into `eof` themselves, so *no* call site needs a guard and the `try` around `,quit` from the previous commit is removed as subsumed. Re-validated with the SIGKILLed-child run: the report prints, no traceback.

**The probe's evidence.** The tail is now computed at the end of `record` and covers the drain *and* the probe, so the `PROBE` verdict ships with the bytes behind it. The never+alive validation now reports:

```
PROBE: ctrl-C then (+ 40 2) printed 42 -- the reader is alive; the missing bytes were dropped
DRAIN-TAIL: b'kaappi> \nkaappi> kaappi> (kaappi> (+kaappi> ... (+ 40 2)\n42\n'
```

— the ctrl-C redraw, the probe form's echo, and the 42, in the report.

**Reject verdict.** The reject path now names itself: `VERDICT: the no-click result b'(1 2 3 94)' printed -- the form arrived and was evaluated, the click did not reposition (not a #2550 timing question)`.

Full five-scenario suite green on ReleaseSafe and Debug; all six verdict paths (reject, late, never+alive, never+wedged-check, child exit, and the no-drain happy path) re-validated from extracted driver copies.

tests/scheme/smoke/repl-mouse-click-2264.sh
```diff
@@ -192,9 +196,100 @@ def idle(quiet=0.5, limit=5.0):
...
+            # and the missing bytes were dropped in flight (the issue's
+            # high-priority case); silence means wedged. The probe's output
+            # lands in buf before the next scenario takes its mark, so it
+            # cannot pollute later assertions.
+            p0 = len(buf)
+            send(b'\x03')
+            idle(0.3)
+            send(b'(+ 40 2)\r')
+            pend = time.monotonic() + 5
+            while time.monotonic() < pend:
```
baijum: The drain loop above breaks on `eof`; this one does not. If the child dies during the probe — between the two `send`s, or while `(+ 40 2)` is being read — `send` now sets `eof` cleanly, but this loop keeps calling `pump(0.5)`, and on a dead master `select` returns readable at once and `read` raises, so `pump` returns immediately: the loop spins flat out for the remaining seconds (~580k iterations/s measured on a dead pty here) and then prints `PROBE: no answer ... the reader is wedged`, which is the wrong verdict for a process that exited. Mirror the drain:

```suggestion
            while time.monotonic() < pend and not eof:
```

and let the `else` branch say "the REPL exited during the probe" when `eof` is set, so the three probe outcomes (42, silence, exit) each name themselves like the drain's do. Not blocking.

Related wording nit, same theme: the `never and eof` verdict at line 249 says "exited during the drain", but `eof` can already be set on entry when the child died during the scenario itself (a failed `send` in `ahead`/`garble` now sets it). "the REPL had exited by the end of the drain" covers both.

baijum (commented):
Re-reviewed bfb980e0. All three points from the second round are addressed the way I would have done them: the `died` flag is gone and both branches test `eof` after the loop, so the last-`pump` window is closed; `send` and `answer_dsr` absorb the write error into `eof` in one place, which also let the `,quit` guard go; the tail is computed after the probe so the ctrl-C redraw ships with the verdict; and the reject path names itself. The five-scenario script is green here on ReleaseSafe (14 s).

One small leftover, inline at line 267: the probe's 5 s wait has no `eof` check, so a child that dies *during* the probe makes that loop spin at full speed (I measured ~580k select+read iterations per second on a dead pty master on macOS) and then reports "wedged" for a REPL that exited. Same class as the fix in this commit, three lines, and I'd merge either way.

Good to merge from my side (GitHub refuses a formal approval on your own PR, so this is a comment review).

baijum:
Both addressed in 13c00704.

**Probe loop / mid-probe exit.** The wait is now `while time.monotonic() < pend and not eof:`, mirroring the drain — no spin on a dead master — and the three probe outcomes each name themselves, checking the printed `42` first (output evidence wins over a subsequent exit):

```
PROBE: ctrl-C then (+ 40 2) printed 42 -- the reader is alive; the missing bytes were dropped
PROBE: the REPL exited during the probe -- see the buffer above
PROBE: no answer to ctrl-C then (+ 40 2) within 5 s -- the reader is wedged, not merely lossy
```

**Regression case** (exit after the drain, before the probe result, asserting the exit-specific diagnostic): driver copy with a slow counting loop as the probe form and a wrapper that SIGKILLs the child mid-evaluation. Actual output:

```
TAIL-CHECK: [b'\n(9 999)\n'] never arrived, even after 2s more pumping -- candidate byte loss in the reader; this is the escalation criterion in issue #2550
PROBE: the REPL exited during the probe -- see the buffer above
```

The run also ends at the eof (8.1 s wall) rather than spinning to the probe deadline — structural confirmation of the loop fix. As before, this lives here rather than as a committed test, since the driver has no in-repo harness and no scenario can crash the current REPL on demand.

**Wording.** The drain verdict is now `never arrived; the REPL had exited by the end of the drain` — true whether the exit happened mid-drain or before the failure was even detected. Re-validated with the kill-at-8 s wrapper, where the child is dead on record entry.

All verdict paths re-validated from the current driver (reject VERDICT, late, never+alive 42, wedged, eof-on-entry, mid-probe exit) and the five-scenario suite is green on ReleaseSafe and Debug.

baijum (commented):
Re-reviewed 13c00704. Both leftovers from the third round are done: the probe's wait now stops on `eof` like the drain's, so a mid-probe exit ends the loop instead of spinning on the dead master, and the three probe outcomes each name themselves (`42`, exited during the probe, no answer). The `elif eof` branch is unambiguous because that whole block is entered under `not eof`, so an `eof` seen there can only have arrived during the probe. The drain verdict's "had exited by the end of the drain" is true for both the died-in-scenario and died-in-drain cases, and the comment now says why.

I traced the failure-during-probe path once more: a failed `send(b'\x03')` sets `eof`; `idle(0.3)` returns on its first `pump` (False on a dead master, no spin); the second `send` is absorbed; the loop is skipped; the report gets `PROBE: the REPL exited during the probe`. The five-scenario script is green here on ReleaseSafe (13.5 s), and the driver parses cleanly.

Nothing further from me. Good to merge (comment review only, since GitHub refuses a formal approval on your own PR).

Issue #1940, opened 2026-07-31
Tokens straddling the 4096-byte read chunk split silently: a line comment's tail becomes program data


baijum:
**Category:** Reader / port I/O · **Severity:** wrong-result (silent) · **Found by:** systematic audit v2, Phase 1C (#1890)

#1893 covers string literals straddling the read-chunk boundary, which **raise**. This is the sibling class that does **not** raise: `read` returns a *prefix* of the real datum and resumes mid-token. No error, no diagnostic.

## Where the chunking actually lives

Not in `reader.zig` — the reader is always handed a complete slice. `read_chunk_size = 4096` (`primitives_io.zig:151`) drives `readDatumFn`'s incremental loop (~`:1660-1710`), which appends a chunk, re-parses from offset 0, and treats **any** error other than `ReadError.UnexpectedEof` as fatal.

So the invariant every scanner must satisfy is: *truncation must be reported as exactly `UnexpectedEof`, and no scanner may treat end-of-slice as a legitimate token terminator.* Every failure below breaks one of those two.

## The worst case: a comment becomes code

A line comment straddling the boundary has its tail read as program data:

```text
file port:    1023 datums; non-(a) datums seen: (zzzzzzzzzzzzzzzzzzzzzzzzzz SECRET)
string port:  1021 datums; non-(a) datums seen: ()
```

The buffer holding only a comment is discarded (`buf.clearRetainingCapacity()`) and reading resumes **inside** the comment. Commented-out text becomes executable data.

## The silent-split class

| token | behaviour at the boundary |
|---|---|
| bare symbol | splits into two symbols |
| number | `1` + `2345678901234567890…` instead of one integer |
| `#\space` | `#\s` + symbol `pace` |
| `#\x41` | `#\x` + number `41` |
| `#true` | `#t` + symbol `rue` |
| line comment | tail becomes program data |

Swept across **every interior split point**, 1714 configurations; the 4096 and 8192 rows are identical in every cell. Oracle throughout: the identical bytes through `open-input-string`, which structurally cannot refill.

## This corrects #1893's stated control

#1893 says a byte-identical file with a bare **symbol** in place of the string "reads all 1024 datums". That control is unsound — it passed only because the symbol was wrapped in a list (lists *do* refill correctly) and because it counted datums rather than comparing them.

Measured directly:

```text
wrapped in a list:  1022 datums; #1021 = (zzzz…40 z's)     correct
[...]

PR #2552, opened 2026-09-07
Implement SRFI 4's homogeneous-vector literal syntax (#2548)

### LINKED ISSUE #2548: Reader accepts only #u8( of SRFI 4's homogeneous-vector literal syntax

baijum:
Closes #2548.

## What

The reader now accepts SRFI 4's external representation for all ten homogeneous-vector kinds — `#s8(...)`, `#u8(...)`, `#s16(...)`, `#u16(...)`, `#s32(...)`, `#u32(...)`, `#s64(...)`, `#u64(...)`, `#f32(...)`, `#f64(...)` — plus SRFI 160's optional extension of the same shape to `#c64(...)`/`#c128(...)`. `write` produces the same readable form, so a written homogeneous vector round-trips through `read` bit-exactly, as the SRFI requires. Previously the reader accepted only `#u8(` (the R7RS bytevector form) and rejected every other prefix at read time, so a conforming program — the SRFI 231 spec corpus writes `#(16 #u16(3895))` in code position — failed before it ran.

```scheme
$ kaappi repro.scm   # (import (srfi 4)) (write (u16vector-length #u16(1 2 3))) ...
3                    # was: read error[KP1002]: unexpected character
(write #s16(1 -2 #xff))  ⇒ #s16(1 -2 255)
```

## How

- **Reader** (`reader_tokens.zig`, `reader_datum.zig`): the `#`-dispatch arm for `u`/`s`/`f`/`c` matches the fixed prefix table (a new token carries the `NumericElementKind`; `u8` keeps the bytevector form and `#u8"..."` SRFI 207's notation, and `#f`/`#false` share the arm). Elements take full number syntax — the spec's own example is `#u8(0 #e1e2 #xff)`, so the bytevector path accepts it too — and are read token-level, so a literal is ONE datum to the reader's depth limit (the printer-gaps depth-1023 controls pin that). Literals are immutable like `#u8(`/`#(...` literals, and `%numeric-vector-set!` now refuses immutable vectors to match `set-car!`/`bytevector-u8-set!`. A prefix cut at end-of-slice refills under incremental reads (#1940 semantics preserved and extended).
- **Element coercion** (`primitives_srfi160.zig`): split into a pure core (`encodeElementRaw`) the literal reader shares, with the primitive-shaped `encodeElement` wrapping it for `set!` — ONE range/exactness table, so a literal and a constructor can never disagree, and a read failure can never write the VM error-detail channel (that detail would otherwise leak into an unrelated runtime error's report — caught in testing).
- **Printer** (`printer.zig`): `write` emits `#s16(1 -2 255)` / `#c64(1.5+0.0i)` for every kind; f32/c64 elements format at f32 precision, so `#f32(0.1)`, not `#f32(0.10000000149011612)` — both round-trip bit-exactly, the f32-shortest form is what other Schemes write. This readable form is also what makes the **native tier** work: `emitQuotedEvalExpr` embeds heap constants by printing them and re-reading at run time.
- **Compiler** (`ir.zig`): a NumericVector literal in code position is a self-evaluating constant (was KP2001 invalid syntax).
- **`.sbc` codec** (`bytecode_file*.zig`): `TAG_NUMERICVECTOR` (immutability byte, kind byte, raw element bytes, backref-eligible) — a program with literals now caches and bundles instead of taking a permanent cache miss.
- **`kaappi fmt`** (its lexer's round-trip guard refused a file containing a new literal), the **REPL structural editor** (slurp/barf/rotate) and the **REPL highlighter** learned the `#TAG(` opens.
- Stale "no reader syntax to round-trip" notes updated in `printer.zig`, `primitives_srfi160.zig` and `docs/dev/srfi-implementation-notes.md`; the audit assertions that pinned the old opaque `#<u16vector ...>` print now pin the readable form.

## Tests

- Reader unit tests: kinds/bytes, cut-prefix refills, immutability, error kinds, write→read round-trip (`tests_reader_incremental.zig`).
- `.sbc` equivalence through the new tag, bare and nested (`tests_bytecode_cache.zig`); fmt round-trip and idempotence (`tests_fmt.zig`); slurp/barf and highlight spans (`repl_sexp.zig`, `repl_highlight.zig`).
- `tests/scheme/srfi/srfi4.scm`: +30 literal assertions (all ten kinds, the spec's `#e`/`#x` example, self-evaluation incl. the SRFI 231 spec-example shape, write/read round-trips, immutability, error cases).
- New `tests/scheme/compile/srfi4-literal-roundtrip-2548.sh`: the whole literal surface through `kaappi compile` (LLVM native) against the interpreter oracle.
- `tools/srfi231_diff.py prose` (the tool that found this): 172 examples, 0 mismatches.

`zig build test` 1959 pass / 0 fail; `tests/scheme/run-all.sh` 2154 pass / 0 fail (incl. R7RS 1395); targeted `-Dgc-stress=true` runs on the touched reader/cache filters green.

### DIFF
--- a/src/bytecode_file.zig
+++ b/src/bytecode_file.zig
@@ -60,7 +60,10 @@ pub const MAGIC = [4]u8{ 'K', 'P', 'B', 'C' };
 /// else stops the reader from parsing an older tail under the new layout. A
 /// bump that only appends header fields may leave `MIN_READ_VERSION` alone,
 /// provided the readers gate the new fields on `ver >=` a named threshold
-/// like `TARGET_FIELDS_SINCE`.
+/// like `TARGET_FIELDS_SINCE`. v14 also gained `TAG_NUMERICVECTOR`
+/// (kaappi#2548) without a bump: additive, and an older same-version reader
+/// that meets tag 19 returns CorruptedFile, which the cache treats as a
+/// plain miss — it misses rather than misreads.
 pub const VERSION: u16 = 14;
 /// Oldest header version this binary reads. v13 files lack the target and
 /// version strings but share v14's tail layout (entry kind, slots, sections),
@@ -146,16 +149,23 @@ pub const TAG_RATIONAL: u8 = 13;
 pub const TAG_COMPLEX: u8 = 14;
 pub const TAG_EOF: u8 = 15;
 pub const TAG_UNDEFINED: u8 = 16;
-/// Back-reference to an already-decoded pair/string/vector/bytevector, by
-/// pre-order first-encounter index (kaappi#2111). This is what lets datum-label
-/// sharing (`'(#1=(1 2) #1#)`) and cyclic literals round-trip as the *same*
-/// object instead of a copy per reference.
+/// Back-reference to an already-decoded pair/string/vector/bytevector/
+/// numeric-vector, by pre-order first-encounter index (kaappi#2111). This is
+/// what lets datum-label sharing (`'(#1=(1 2) #1#)`) and cyclic literals
+/// round-trip as the *same* object instead of a copy per reference.
 pub const TAG_BACKREF: u8 = 17;
 /// A closure (kaappi#1888): a nested function-table reference plus its captured
 /// upvalues. Reached from serialized procedural (`er-macro-transformer`)
 /// transformer `proc`s; an ordinary compiled body never needs it, because a HIT
 /// re-runs the body and recreates its closures against the live environment.
 pub const TAG_CLOSURE: u8 = 18;
+/// SRFI 4/160 homogeneous-vector literal (#s16( ... / #f64( ... / #c128( ...):
+/// an immutability byte, a kind byte (a NumericElementKind discriminant),
+/// a u32 byte length, then the raw element bytes in host-native order — the
+/// same element encoding the heap type and the #TAG( reader already share.
+/// Byte order is host-local by construction: a cache entry or bundle is
+/// consumed by a binary of the same architecture that wrote it.
+pub const TAG_NUMERICVECTOR: u8 = 19;
 
 pub const BytecodeError = error{
     InvalidMagic,
--- a/src/fmt.zig
+++ b/src/fmt.zig
@@ -98,7 +98,8 @@ pub const Node = struct {
     /// separated them. Drives trailing-vs-leading comment placement and
     /// blank-line grouping; ordinary code layout is otherwise reflowed.
     newlines_before: u32 = 0,
-    /// A `#(` / `#u8(` literal, whose first element is data, never an operator.
+    /// A `#(` / `#u8(` / `#s16(`-style homogeneous-vector literal, whose
+    /// first element is data, never an operator.
     is_data: bool = false,
     /// Memoised inline width (see `fmt_print`), so fit checks stay linear.
     inline_width: ?usize = null,
@@ -123,7 +124,7 @@ const max_nesting: u32 = 1024;
 const TokKind = enum {
     lparen,
     rparen,
-    list_open, // "#(" or "#u8("
+    list_open, // "#(" or "#u8("/"#s16("-style vector opens
     atom,
     prefix, // ' ` , ,@ #N=
     datum_comment, // #;
@@ -235,9 +236,22 @@ const Lexer = struct {
             self.pos += 2;
             return .list_open;
         }
-        if (rest.len >= 4 and std.mem.eql(u8, rest[0..4], "#u8(")) {
-            self.pos += 4;
-            return .list_open;
+        // SRFI 4/160 homogeneous-vector opens -- #u8( plus the eleven
+        // NumericElementKind prefixes (#s16(, #f64(, #c128(, ...). The tag
+        // and its '(' are ONE lexeme, a list_open, exactly like #(: the '('
+        // belongs to the literal, so splitting "#s16" off as an atom and
+        // "(1 2)" off as a following list would reflow the program into
+        // something the reader rejects (or reads differently), which the
+        // round-trip check would then refuse.
+        if (rest.len >= 2) {
+            var i: usize = 1;
+            while (i < rest.len and std.ascii.isAlphanumeric(rest[i])) i += 1;
+            if (i >= 2 and i < rest.len and rest[i] == '(' and
+                types.isHomogeneousVectorTag(rest[1..i]))
+            {
+                self.pos += i + 1;
+                return .list_open;
+            }
         }
         // SRFI 207 string-notated bytevector #u8"...": one verbatim lexeme,
         // like the ordinary string it contains-ish -- checked before the
--- a/src/ir.zig
+++ b/src/ir.zig
@@ -621,7 +621,8 @@ pub fn isSpecialForm(name: []const u8) bool {
 pub fn lowerWithMacros(ir: *IR, expr: Value, macros: ?*std.StringHashMap(Value)) CompileError!*Node {
     if (types.isFixnum(expr) or types.isFlonum(expr) or types.isBignum(expr) or
         types.isComplex(expr) or types.isRationalObj(expr) or types.isString(expr) or
-        types.isChar(expr) or types.isVector(expr) or types.isBytevector(expr))
+        types.isChar(expr) or types.isVector(expr) or types.isBytevector(expr) or
+        types.isNumericVector(expr))
     {
         return ir.makeConst(expr);
     }
--- a/src/primitives_srfi160.zig
+++ b/src/primitives_srfi160.zig
@@ -27,8 +27,9 @@
 //! default for both).
 //!
 //! Multi-byte elements are stored in host-native byte order (matching
-//! printer.zig's own numeric-vector display arm) -- there is no reader
-//! syntax to round-trip and no cross-process persistence, so native order
+//! printer.zig's own numeric-vector display arm) -- the #TAG( literal syntax
+//! and the .sbc constant codec both carry kind-tagged raw bytes, and both
+//! halves live in the same binary reading its own output, so native order
 //! avoids needless byte-swaps on big-endian hosts (s390x, ppc64le).
 
 const std = @import("std");
@@ -102,67 +103,6 @@ fn magnitudeAndSign(val: Value) ExactMag {
     return .not_exact;
 }
 
-fn expectSignedInRange(proc: []const u8, val: Value, comptime bits: u7) PrimitiveError!i64 {
-    const ms = switch (magnitudeAndSign(val)) {
-        .fits => |m| m,
-        .too_wide => return argError(proc, "integer does not fit a signed {d}-bit element", .{bits}),
-        .not_exact => return typeError(proc, "exact integer", val),
-    };
-    const shift: u6 = bits - 1;
-    const max_mag_pos: u64 = (@as(u64, 1) << shift) - 1;
-    const max_mag_neg: u64 = @as(u64, 1) << shift;
-    if (ms.positive) {
-        if (ms.mag > max_mag_pos) return argError(proc, "integer does not fit a signed {d}-bit element", .{bits});
-        return @intCast(ms.mag);
-    }
-    if (ms.mag > max_mag_neg) return argError(proc, "integer does not fit a signed {d}-bit element", .{bits});
-    if (ms.mag == max_mag_neg) {
-        // Only reachable when bits == 64 and mag == 2^63 (i64::MIN) -- for
-        // bits < 64, max_mag_neg fits comfortably in a positive i64 already,
-        // so the general path below would work too; this keeps it exact.
-        if (bits == 64) return std.math.minInt(i64);
-        return -@as(i64, @intCast(ms.mag));
-    }
-    return -@as(i64, @intCast(ms.mag));
-}
-
-fn expectUnsignedInRange(proc: []const u8, val: Value, comptime bits: u7) PrimitiveError!u64 {
-    const ms = switch (magnitudeAndSign(val)) {
-        .fits => |m| m,
-        // A negative one is rejected for being negative, not for being wide --
-        // the same reason, and the same message, a negative fixnum gets below.
-        .too_wide => |w| return if (w.positive)
-            argError(proc, "integer does not fit an unsigned {d}-bit element", .{bits})
-        else
-            argError(proc, "negative integer for an unsigned {d}-bit element", .{bits}),
-        .not_exact => return typeError(proc, "exact integer", val),
-    };
-    if (!ms.positive and ms.mag != 0) return argError(proc, "negative integer for an unsigned {d}-bit element", .{bits});
-    const max_mag: u64 = if (bits == 64) std.math.maxInt(u64) else blk: {
-        const shift: u6 = bits;
-        break :blk (@as(u64, 1) << shift) - 1;
-    };
-    if (ms.mag > max_mag) return argError(proc, "integer does not fit an unsigned {d}-bit element", .{bits});
-    return ms.mag;
-}
-
-fn expectReal(proc: []const u8, val: Value) PrimitiveError!f64 {
-    if (types.isFixnum(val) or types.isFlonum(val) or types.isBignum(val) or types.isRationalObj(val)) {
-        return types.toF64(val);
-    }
-    return typeError(proc, "real number", val);
-}
-
-const ComplexParts = struct { re: f64, im: f64 };
-
-fn expectComplexParts(proc: []const u8, val: Value) PrimitiveError!ComplexParts {
-    if (types.isComplex(val)) {
-        const c = types.toComplex(val);
-        return .{ .re = try expectReal(proc, c.real), .im = try expectReal(proc, c.imag) };
-    }
-    return .{ .re = try expectReal(proc, val), .im = 0.0 };
-}
-
 fn makeExactFromI64(gc: *memory.GC, n: i64) PrimitiveError!Value {
     const fixnum_min: i64 = -(@as(i64, 1) << 47);
     const fixnum_max: i64 = (@as(i64, 1) << 47) - 1;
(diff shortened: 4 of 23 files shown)


src/ir.zig
```diff
@@ -621,7 +621,8 @@ pub fn isSpecialForm(name: []const u8) bool {
 pub fn lowerWithMacros(ir: *IR, expr: Value, macros: ?*std.StringHashMap(Value)) CompileError!*Node {
     if (types.isFixnum(expr) or types.isFlonum(expr) or types.isBignum(expr) or
         types.isComplex(expr) or types.isRationalObj(expr) or types.isString(expr) or
-        types.isChar(expr) or types.isVector(expr) or types.isBytevector(expr))
+        types.isChar(expr) or types.isVector(expr) or types.isBytevector(expr) or
+        types.isNumericVector(expr))
```
baijum, line 625: **Bug: the sibling self-evaluating list in `compiler.zig` was not updated, so a macro expanding to a bare literal is still KP2001.**

`Compiler.compileExpr` (compiler.zig ~L818–L832) has its own `isVector` / `isBytevector` arms and falls through to `CompileError.InvalidSyntax` for a NumericVector. It is reached from `compiler_macro.zig:490` (`compileExpr(final_form, …)` after expansion), from `.passthrough` IR nodes, and from the usertext-marker safety net at compiler.zig:891. Repro on this branch:

```scheme
(define-syntax lit (syntax-rules () ((_) #s16(1 2))))
(write (lit))
;; => macro.scm:3:1: compile error[KP2001]: invalid syntax
```

The same template with `#u8(1 2)` works, and `(let-syntax ((m (syntax-rules () ((_) #u16(7))))) (m))` fails the same way. Wrapping the literal in a call (`(list x #f32(1.5))`) is fine because that goes through `lowerWithMacros` here.

I verified the fix locally — changing compiler.zig's `if (types.isBytevector(expr))` to `if (types.isBytevector(expr) or types.isNumericVector(expr))` makes both repros print the literal. Worth a regression assertion in `tests/scheme/srfi/srfi4.scm` (a `define-syntax` whose template *is* the literal) and one line in the compile script so the native tier's macro path is covered too.

src/fmt.zig
```diff
@@ -272,6 +287,18 @@ const Lexer = struct {
         return .atom;
     }
 
+    /// Is `name` one of the SRFI 4/160 homogeneous-vector tags (u8, s8,
+    /// u16, s16, u32, s32, u64, s64, f32, f64, c64, c128)? fmt's lexer is a
+    /// deliberate reimplementation of the token layer, so this mirrors
+    /// reader_tokens.homogeneousVectorPrefix, the authoritative table.
+    fn homogeneousVectorTag(name: []const u8) bool {
```
baijum: **Nit (drift risk): this string table now exists in four hand-kept copies** — here, `repl_highlight.isHomogeneousVectorTag`, `repl_sexp.homogeneousVectorTag`, and the authoritative `reader_tokens.homogeneousVectorPrefix`. Each says it "mirrors" the reader, which is exactly the kind of parallel list CLAUDE.md warns drifted every time for the native-backend gates.

All four are derivable: the eleven non-u8 names are `@tagName` of `types.NumericElementKind`, so `std.meta.stringToEnum(types.NumericElementKind, word)` plus the `u8` special case *is* the table. One `pub fn isHomogeneousVectorTag(name) bool` next to the enum in `types_numeric.zig` (or exported from `reader_tokens.zig`) would let fmt/highlight/sexp call it — fmt's lexer being a deliberate reimplementation of the *token layer* doesn't require it to re-spell the name set. Failing that, a `comptime` assertion that each copy equals the enum's field names would at least make drift a build error.

src/primitives_srfi160.zig
```diff
@@ -304,6 +361,10 @@ fn numericVectorRefFn(args: []const Value) PrimitiveError!Value {
 fn numericVectorSetFn(args: []const Value) PrimitiveError!Value {
     if (!types.isNumericVector(args[0])) return typeError("%numeric-vector-set!", "numeric-vector", args[0]);
     const nv = asNumericVector(args[0]);
+    // A #TAG( literal reads back immutable, exactly like #u8( and #(...)
+    // literals — set-car!/bytevector-u8-set! on a literal raise, so
+    // TAGvector-set! must too.
+    if (types.toObject(args[0]).flags.immutable) return argError("%numeric-vector-set!", "cannot mutate an immutable numeric vector", .{});
```
baijum: **Nit: the user sees the internal primitive name.** `(s8vector-set! #s8(1 2) 0 9)` reports

```
%numeric-vector-set!: cannot mutate an immutable numeric vector
```

whereas the siblings say `vector-fill!: cannot mutate an immutable vector` / `bytevector-copy!: cannot mutate an immutable bytevector`. Naming the `%`-primitive is a pre-existing property of every `%numeric-vector-*` error (index errors do it too), so not this PR's to solve — but since the message is new, it could at least name the kind: `cannot mutate an immutable {s}vector literal` with `@tagName(nv.kind)` costs nothing and tells the user which literal they hit. (Also reached via `s8vector-fill!` / `-copy!` / `-reverse!` / `-swap!` from `(srfi 160 s8)` — all correctly refuse, I checked.)

src/bytecode_file.zig
```diff
@@ -146,16 +146,23 @@ pub const TAG_RATIONAL: u8 = 13;
...
 /// transformer `proc`s; an ordinary compiled body never needs it, because a HIT
 /// re-runs the body and recreates its closures against the live environment.
 pub const TAG_CLOSURE: u8 = 18;
+/// SRFI 4/160 homogeneous-vector literal (#s16( ... / #f64( ... / #c128( ...):
+/// an immutability byte, a kind byte (a NumericElementKind discriminant),
+/// a u32 byte length, then the raw element bytes in host-native order — the
+/// same element encoding the heap type and the #TAG( reader already share.
+/// Byte order is host-local by construction: a cache entry or bundle is
+/// consumed by a binary of the same architecture that wrote it.
+pub const TAG_NUMERICVECTOR: u8 = 19;
```
baijum, line 168: **Nit (doc): record the additive tag in the version history above.** The comment block over `VERSION` lists what each format version changed (v12: TAG_COMPLEX payload, v13: entry kind, v14: target fields), and its bumping rule says a tail change must raise `MIN_READ_VERSION`. A new tag *is* a tail change, and this PR deliberately leaves v14 alone — which is fine (the entry is keyed by build id, and a same-version reader that meets tag 19 returns `CorruptedFile`, which `cache.zig:270` treats as a plain miss), but a future reader of that rule will wonder whether this was an oversight. One line — "v14 also gained `TAG_NUMERICVECTOR` (kaappi#2548) without a bump: additive, and an older same-version reader misses rather than misreads" — closes the question.

src/reader_datum.zig
```diff
@@ -355,17 +367,77 @@ fn readBytevector(self: *Reader) ReadError!Value {
...
+        // spec's own example -- #u8(0 #e1e2 #xff) -- writes elements in full
+        // integer syntax, so accept any exact integer 0..255. Elements are
+        // read token-level (nextToken + tokenToValue, no datum nesting), so
+        // they do not count toward the reader's depth limit -- a #u8( leaf
+        // at depth 1023 round-trips exactly as it did before (printer-gaps).
+        // Anything not in 0..255 is a read error, as the R7RS bytevector
+        // grammar demands.
+        const elem = try readElementValue(self);
+        var elem_root = elem;
+        self.gc.pushRoot(&elem_root);
```
baijum: Nit: this root is not load-bearing — nothing between `pushRoot` and `popRoot` allocates (`toFixnum` and two compares), and `elem` is only read as a fixnum. Harmless, but a reader may take it as a hint that something here can collect. The one in `readNumericVector` below is equally inert (`ArrayList.resize` on `gc.allocator` bypasses `maybeCollect`, and `encodeElementRaw` never allocates), though keeping that one as cheap insurance around the coercion call is defensible.

baijum (commented):
Reviewed the whole diff and exercised the branch locally (build, `zig build test -Dtest-filter=2548`, the new compile script, `kaappi fmt`/`check`, a 200k-sample f32 write→read fuzz, and an interpreter-vs-native comparison of literal immutability). The design is right — one coercion table, readable `write` form driving the native tier, token-level element reads keeping the depth seam — and everything I probed round-trips: `+nan.0`/`-0.0`/subnormals, `#u64` max / `#s64` min, exact and complex elements in float vectors, datum labels, comments inside literals, `#f`/`#false`/`#f32(` disambiguation, and native vs. interpreter agree on output and on immutability.

One real gap, verified with a repro and a one-line fix (inline on `ir.zig`): **a macro whose expansion is a bare `#TAG(` literal fails with KP2001** because `compiler.zig`'s `compileExpr` self-evaluating list has `isBytevector` but not `isNumericVector`. `#u8(1 2)` works from the same macro, so the asymmetry is easy to hit. The rest are non-blocking: the tag table now lives in four hand-kept copies, the immutability error names `%numeric-vector-set!`, and `kaappi check` calls a numeric-vector literal "a value literal" (`check_lint.zig` `describeLiteral` has no arm for it — `(vector-ref #s16(1) 0)` reports `a value literal was given` where `#u8(1)` gets `a bytevector literal`; cosmetic, not in the diff).

tests/scheme/srfi/srfi4.scm
```diff
@@ -134,6 +134,68 @@
...
+(test-equal '(-2147483648 2147483647) (s32vector->list #s32(-2147483648 2147483647)))
+(test-equal '(0 18446744073709551615) (u64vector->list #u64(0 18446744073709551615)))
+(test-equal '(-9223372036854775808 1) (s64vector->list #s64(-9223372036854775808 1)))
+(test-equal 1.5 (f32vector-ref #f32(1.5) 0))
+(test-equal -1.5 (f64vector-ref #f64(-1.5) 0))
+;;; The spec's own example: full integer syntax inside the literal.
+(test-equal '(0 100 255) (u8vector->list #u8(0 #e1e2 #xff)))
+;;; An empty literal.
+(test-equal 0 (u16vector-length #u16()))
+;;; Literals are self-evaluating in code position, including nested inside
```
baijum, line 157: Suggest adding the macro shape here, since it is the one code position that still fails on this branch (see the `ir.zig` comment):

```scheme
;;; A literal that IS a macro's expansion goes through compileExpr, not lowerWithMacros.
(define-syntax %lit (syntax-rules () ((_) #s16(1 2))))
(test-equal '(1 2) (s16vector->list (%lit)))
(test-equal 7 (u16vector-ref (let-syntax ((m (syntax-rules () ((_) #u16(7))))) (m)) 0))
```

Both fail with KP2001 today and pass with the one-line compiler.zig fix.

src/ir.zig
baijum: Fixed in bc256e94: the isBytevector arm in Compiler.compileExpr now also accepts NumericVector, with a comment naming the two code positions (macro expansion via compiler_macro.zig, passthrough IR). Regression assertions added both to tests/scheme/srfi/srfi4.scm (define-syntax and let-syntax expanding to a bare literal — both pass) and to the compile shell test's probe + golden output, so the native tier's macro path is covered too.

src/reader_datum.zig
```diff
@@ -344,6 +345,17 @@ fn patchPlaceholder(gc: *memory.GC, datum: Value, placeholder: Value) error{OutO
...
 
+/// One element of a `#u8(` / `#TAG(` homogeneous-vector literal: the next
+/// token, converted by the same datum constructor readDatum uses, but
+/// WITHOUT the per-datum depth increment — a literal's elements are lexical
+/// parts of the enclosing datum, not nested datums, so a #u8( leaf at depth
+/// 1023 reads exactly as it did before the SRFI 4 syntax existed (the
+/// printer-gaps depth-seam controls pin this).
+fn readElementValue(self: *Reader) ReadError!Value {
+    const tok = try self.nextToken();
+    return tokenToValue(self, tok);
```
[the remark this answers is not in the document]
baijum: Fixed in bc256e94: readElementValue now guards compound element tokens (hash_u8_lparen / hash_numvec_lparen) with the depth check and increment exactly once, while scalar elements stay depth-neutral (the printer-gaps depth-1023 controls still pass). Regression test added: MAX_NESTING_DEPTH+50 nested #s8( literals reads as NestingTooDeep.

src/repl_sexp.zig
```diff
@@ -116,19 +116,37 @@ fn skipTrivia(src: []const u8, from: usize, depth: u32) usize {
...
     if (i + 1 < src.len and src[i + 1] == '(') return 2;
-    if (i + 3 < src.len and (src[i + 1] == 'u' or src[i + 1] == 'U') and
-        src[i + 2] == '8' and src[i + 3] == '(') return 4;
+    // #T( where T is one of the SRFI 4/160 tags (u8, s16, f64, c128, ...).
+    // Tag table mirrors reader_tokens.homogeneousVectorPrefix.
+    if (i + 2 < src.len and std.ascii.isAlphanumeric(src[i + 1])) {
+        var j = i + 1;
+        while (j < src.len and std.ascii.isAlphanumeric(src[j])) j += 1;
+        if (j < src.len and src[j] == '(' and j > i + 1 and homogeneousVectorTag(src[i + 1 .. j])) {
+            return j + 1 - i;
```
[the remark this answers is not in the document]
baijum: Fixed in bc256e94: rotate now branches on src[form.open] == '#' — a #(... / #TAG(... form rotates ALL children (minimum two, since there is no head to keep), while plain lists keep the kids[0]-head behavior. Verified against the pinned #(a| b c) → #(b c a|) expectation.

tests/scheme/compile/srfi4-literal-roundtrip-2548.sh
```diff
@@ -0,0 +1,109 @@
...
+
+n=$((n + 1))
+native_status=0
+native_out="$("$bin" 2> /dev/null)" || native_status=$?
+if assert_tiers_agree "native binary vs interpreter" \
+    "$interp_out" "$interp_status" "$native_out" "$native_status"; then
+    ok "the native binary agrees with the interpreter on stdout and exit status"
+else
+    fail=1
+fi
```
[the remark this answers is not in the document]
baijum: Fixed in bc256e94: the script now asserts a golden 18-line expected output for the interpreter run (and exits on mismatch with a diff) before any tier comparison.

tests/scheme/compile/srfi4-literal-roundtrip-2548.sh
```diff
@@ -0,0 +1,109 @@
...
+KAAPPI="${1:-zig-out/bin/kaappi}"
+KAAPPI_ABS="$(cd "$(dirname "$KAAPPI")" && pwd)/$(basename "$KAAPPI")"
+REPO_DIR="$(cd "$(dirname "$0")/../../.." && pwd)"
+
+ensure_runtime_lib "$REPO_DIR"
+
+DIR="$(mktemp -d)"
+trap 'rm -rf "$DIR"' EXIT
+export KAAPPI_HOME="$(mktemp -d)"
+trap 'rm -rf "$KAAPPI_HOME"' EXIT
```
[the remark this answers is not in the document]
baijum: Fixed in bc256e94: one cleanup() removing both DIR and KAAPPI_HOME registered as a single EXIT trap, and the mktemp/KAAPPI_ABS assignments are now split declare/assign with explicit failure exits (SC2155).

src/fmt.zig
baijum: Done in bc256e94: the name set now lives in ONE place — pub fn types_numeric.isHomogeneousVectorTag (u8 special case + std.meta.stringToEnum over NumericElementKind, so a future element kind cannot drift out of the syntax) — re-exported from types.zig and used by reader_tokens.homogeneousVectorPrefix, fmt.zig, repl_sexp.zig and repl_highlight.zig. The three hand-kept copies are deleted.

src/primitives_srfi160.zig
baijum: Done in bc256e94: the message now names the kind — 'cannot mutate an immutable s8vector' (via @tagName(nv.kind)).

src/bytecode_file.zig
baijum: Done in bc256e94: the VERSION comment block now records that v14 gained TAG_NUMERICVECTOR without a bump, and why that is safe (additive; an older same-version reader returns CorruptedFile, which the cache treats as a plain miss).

src/reader_datum.zig
baijum: Done in bc256e94: the inert root in readBytevector is removed (with a comment explaining why nothing in the window can collect); the one in readNumericVector stays as cheap insurance around the coercion call, now with a comment saying exactly that.

tests/scheme/srfi/srfi4.scm
baijum: Added in bc256e94, alongside the compiler.zig fix — both assertions are in tests/scheme/srfi/srfi4.scm and the same shape is in the compile shell test's probe + golden output.

tests/scheme/audit/primitives_srfi160-audit.scm
```diff
@@ -258,8 +258,8 @@
                (eq? (real? e)
                     (real? (read (open-input-string (write-to-string e)))))))
 ;; Control 3: the vector printer is honest about the zero imaginary part.
-(test-equal "#<c128vector 1.5+0.0i>" (write-to-string (c128vector 1.5)))
-(test-equal "#<c64vector 1.5+0.0i>" (write-to-string (c64vector 1.5)))
+(test-equal "#c128(1.5+0.0i)" (write-to-string (c128vector 1.5)))
+(test-equal "#c64(1.5+0.0i)" (write-to-string (c64vector 1.5)))
```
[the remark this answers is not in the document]
baijum: Done in bc256e94: descriptive names added to the c128/c64 zero-imaginary assertions and the f64/c64 literal assertions in this file.

tests/scheme/srfi/srfi4.scm
```diff
@@ -134,6 +134,68 @@
   (s32vector-set! v 0 -1000000)
   (test-equal -1000000 (s32vector-ref v 0)))
 
+;;; --- the external representation: #TAG( ... ) literals (kaappi#2548) ---
+;;; SRFI 4's read/write syntax for every one of its ten kinds. Before the
+;;; fix the reader accepted only #u8( and rejected the rest at read time,
+;;; so a conforming program failed before it ran.
+(test-equal 3 (u16vector-length #u16(1 2 3)))
```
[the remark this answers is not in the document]
baijum: Done in bc256e94: all new test-equal calls in tests/scheme/srfi/srfi4.scm now use the named form.

src/fmt.zig
```diff
@@ -272,6 +287,18 @@ const Lexer = struct {
...
+        const tags = [_][]const u8{ "u8", "s8", "u16", "s16", "u32", "s32", "u64", "s64", "f32", "f64", "c64", "c128" };
+        for (tags) |t| {
+            if (std.mem.eql(u8, name, t)) return true;
+        }
+        return false;
+    }
```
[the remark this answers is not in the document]
baijum: Same resolution as the duplicate thread on fmt.zig: one shared predicate, types.isHomogeneousVectorTag in types_numeric.zig (derived from the enum), now used by all four consumers — reader, fmt, repl_sexp, repl_highlight.

src/reader_datum.zig
```diff
@@ -344,6 +345,29 @@ fn patchPlaceholder(gc: *memory.GC, datum: Value, placeholder: Value) error{OutO
     }
 }
 
+/// One element of a `#u8(` / `#TAG(` homogeneous-vector literal: the next
+/// token, converted by the same datum constructor readDatum uses, but
+/// WITHOUT the per-datum depth increment — a literal's elements are lexical
+/// parts of the enclosing datum, not nested datums, so a #u8( leaf at depth
+/// 1023 reads exactly as it did before the SRFI 4 syntax existed (the
+/// printer-gaps depth-seam controls pin this). The one exception is a
```
baijum: **Nit (comment accuracy):** this and the `(printer-gaps)` parenthetical in `readBytevector` below (L387) say the printer-gaps controls pin a `#u8(` leaf reading at depth 1023. They don't: `tests/scheme/compliance/printer-gaps.scm` has no bytevector or numeric-vector case at all — its only reader-depth control is `parens 1023 "x"` around a plain atom (L181). The seam behaves as described (I checked `#u8(1 2)` and `#s16(1 2)` under 1023 parens read, 1024 is NestingTooDeep), but the claim should point at a test that exists — see the `deep_leaf` suggestion in `tests_reader_incremental.zig`, which is the natural place to make it true.

src/tests_reader_incremental.zig
```diff
@@ -183,6 +191,95 @@ test "whole-input mode keeps the precise final verdicts" {
...
+    var gc = memory.GC.init(testing.allocator);
+    defer gc.deinit();
+
+    // Scalar elements are depth-neutral, so a literal leaf still reads at
+    // depth 1023 (the printer-gaps control); a nested literal recurses into
+    // another literal loop and must hit NestingTooDeep, not the native
+    // stack limit.
+    // A literal wrapped in plain nesting reads: scalar elements are
+    // depth-neutral.
+    const deep_leaf = "(((((((((" ++ "#s16(1 2)" ++ ")))))))))";
```
baijum: **Nit (test strength):** `deep_leaf` nests 9 levels, so it passes whether or not scalar elements pay the depth gate — it cannot distinguish the property the comment above it describes. Building it at the boundary makes the seam a real regression guard and gives the `printer-gaps` references in `reader_datum.zig` something to point at:

```zig
    var leaf: std.ArrayList(u8) = .empty;
    defer leaf.deinit(testing.allocator);
    const plain = Reader.MAX_NESTING_DEPTH - 1; // 1023: the last depth that reads
    for (0..plain) |_| leaf.append(testing.allocator, '(') catch return error.OutOfMemory;
    leaf.appendSlice(testing.allocator, "#s16(1 2)") catch return error.OutOfMemory;
    for (0..plain) |_| leaf.append(testing.allocator, ')') catch return error.OutOfMemory;
    _ = try parseOne(&gc, leaf.items, false);
```

(Also: the two comment sentences at L266–L270 say the same thing twice — "Scalar elements are depth-neutral, so a literal leaf still reads…" and "A literal wrapped in plain nesting reads: scalar elements are depth-neutral." — one can go.)

src/repl_sexp.zig
```diff
@@ -430,8 +438,13 @@ fn rotate(allocator: std.mem.Allocator, src: []const u8, form: Form, pos: usize)
     var kids: std.ArrayList(Span) = .empty;
     defer kids.deinit(allocator);
     try collectChildren(allocator, src, form, &kids);
-    if (kids.items.len < 3) return null;
-    const args = kids.items[1..];
+    // A `#(`/`#TAG(` literal has no call head -- every child is data and all
+    // of them rotate. A plain list keeps kids[0] as the head and cycles only
+    // its arguments.
+    const is_data = src[form.open] == '#';
```
baijum, line 444: **Nit (test coverage):** this branch changes `#(` rotate as well as the new `#TAG(` opens, but the only literal-rotate expectation (`#(a| b c)` → `#(b c a|)`, cursor on the *first* child) is one the old head-skipping code could also satisfy for some cursor placements. A cursor on a middle element is what separates "every child rotates" from "kids[0] is a head": with this code `#(a b| c)` → `#(b| c a)` and `#s16(1 2| 3)` → `#s16(2| 3 1)`. Two `expectEdit` lines with those shapes would pin the all-children semantics the comment describes.

baijum (commented):
Re-reviewed the amended commit bc256e94 against 766d071e. Every item from the first pass is addressed as described, and I re-verified each on the new head: the macro-expansion repro (`define-syntax` and `let-syntax` templates that *are* the literal) now prints the literal in both the interpreter and through `kaappi compile`; the tag table is one enum-derived predicate; the immutability message names the kind; the VERSION history records the additive tag; the inert root is gone.

The new depth gate for compound elements is a good catch — that recursion was unbounded on 766d071e. I probed it beyond the unit test: a 2000-deep `#s8(` chain, a 2000-deep `#u8(` chain, and 3000-level alternations of `#s8(` with `#(`, `'`, `(` and `#0=` all report `nesting too deep`; a `#s16(...)` or `#u8(...)` leaf under 1023 plain parens still reads and 1024 is refused. `tests/scheme/srfi/srfi4.scm` (115 passes), the compile script with its new golden output, and the touched unit-test filters are green here.

Two small remaining notes inline, both about the comments/tests that *describe* the depth seam rather than the code: the "printer-gaps pins this" claim points at a control that does not exist for bytevector leaves, and the new `deep_leaf` test nests 9 levels, so nothing actually pins the 1023 boundary the comments rely on.

src/reader_datum.zig
baijum: Fixed in 734eb5c6: both comments now state the actual seam — a literal under 1023 plain parens reads, one under 1024 is NestingTooDeep, same as a plain atom — and point at the depth-gate test in tests_reader_incremental.zig instead of printer-gaps.

src/tests_reader_incremental.zig
baijum: Adopted in 734eb5c6: the leaf is now built at the boundary (MAX_NESTING_DEPTH-1 = 1023 plain parens around #s16(1 2)), so it fails if elements ever start paying the per-datum increment; the duplicated comment sentence is gone (one comment remains, describing the seam the boundary build guards). Test passes.

src/repl_sexp.zig
baijum: Added in 734eb5c6: both middle-cursor expectations are in the test block — #(a b| c) → #(b| c a) and #s16(1 2| 3) → #s16(2| 3 1). I verified both against the actual apply() output before pinning them (they match your derivation exactly).

src/tests_reader_incremental.zig
```diff
@@ -183,6 +191,99 @@ test "whole-input mode keeps the precise final verdicts" {
...
+    // controls in tests/scheme/compliance/printer-gaps.scm sit at) reads
+    // fine — a real seam guard, since it fails if elements ever start paying
+    // the per-datum depth increment.
+    var leaf: std.ArrayList(u8) = .empty;
+    defer leaf.deinit(testing.allocator);
+    const plain = Reader.MAX_NESTING_DEPTH - 1; // 1023: the last depth that reads
+    for (0..plain) |_| leaf.append(testing.allocator, '(') catch return error.OutOfMemory;
+    leaf.appendSlice(testing.allocator, "#s16(1 2)") catch return error.OutOfMemory;
+    for (0..plain) |_| leaf.append(testing.allocator, ')') catch return error.OutOfMemory;
+    _ = try parseOne(&gc, leaf.items, false);
```
baijum: Optional: the comment in `reader_datum.zig` (L352) now says this test pins *both* halves of the seam — a literal leaf under 1023 reads **and** one under 1024 is NestingTooDeep — but only the positive half is asserted here; the negative below is the nested-chain shape, not a leaf. Reusing the buffer with one more paren on each side closes it in two lines:

```zig
    leaf.insert(testing.allocator, 0, '(') catch return error.OutOfMemory;
    leaf.append(testing.allocator, ')') catch return error.OutOfMemory;
    try testing.expectError(ReadError.NestingTooDeep, parseOne(&gc, leaf.items, false));
```

(I checked it holds on this head.) Not worth another push on its own.

baijum (commented):
Third pass, 734eb5c6 vs bc256e94: the delta is exactly the three nits from the last round, each done as suggested — the `deep_leaf` guard now sits at `MAX_NESTING_DEPTH - 1` and reads, the two reader comments now cite that test instead of a printer-gaps control that had no bytevector case, and `rotate` has the middle-cursor expectations for both `#(` and `#s16(` (`#(a b| c)` → `#(b| c a)`, `#s16(1 2| 3)` → `#s16(2| 3 1)`, matching what I'd computed from the code).

Re-verified on this head: the depth-gate, forms-too and 2548 unit-test filters pass; a `#s16(`/`#u8(` leaf under 1023 parens reads and 1024 is NestingTooDeep; 2000-deep `#s8(`/`#u8(` chains report NestingTooDeep. Nothing blocking remains from my side (GitHub will not let the author approve, so this is a comment review). One optional one-liner inline on the seam test.

src/tests_reader_incremental.zig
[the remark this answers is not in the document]
baijum: Applied in 684e056a: the boundary parse now asserts a non-null datum, so a regression that answers clean EOF for this nonempty input fails the test.

baijum: Adopted in 684e056a (since I was pushing anyway): the buffer gains one paren on each side and the same shape asserts NestingTooDeep, so the test now pins both halves of the seam the reader_datum comment describes.

tests/scheme/audit/primitives_srfi160-audit.scm
```diff
@@ -814,21 +815,20 @@
...
+;;; and what `read` accepts for every kind, kaappi#2548: the written form
+;;; round-trips through read bit-exactly, which the native tier's
+;;; print-and-re-read constant embedding relies on.
 ;;; ------------------------------------------------------------------
 
 (test-equal "#u8(1 2 3)" (write-to-string (u8vector 1 2 3)))
-(test-equal "#<s8vector 1 2 3>" (write-to-string (s8vector 1 2 3)))
-(test-equal "#<f32vector 1.5>" (write-to-string (f32vector 1.5)))
+(test-equal "#s8(1 2 3)" (write-to-string (s8vector 1 2 3)))
+(test-equal "#f32(1.5)" (write-to-string (f32vector 1.5)))
```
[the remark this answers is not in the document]
baijum: Done in 684e056a: both assertions now carry descriptive names, expected values and expressions unchanged.

tests/scheme/compile/srfi4-literal-roundtrip-2548.sh
```diff
@@ -0,0 +1,146 @@
...
+skip_on_windows "compile suite needs a native Zig toolchain on this machine (kaappi#1613)"
+
+KAAPPI="${1:-zig-out/bin/kaappi}"
+KAAPPI_ABS="$(cd "$(dirname "$KAAPPI")" && pwd)/$(basename "$KAAPPI")" || exit 1
+REPO_DIR="$(cd "$(dirname "$0")/../../.." && pwd)"
+
+ensure_runtime_lib "$REPO_DIR"
+
+DIR="$(mktemp -d)" || exit 1
+KAAPPI_HOME="$(mktemp -d)" || exit 1
```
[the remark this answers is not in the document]
baijum: Fixed in 684e056a (with the variable name corrected — the assignment target is KAAPPI_HOME): a failed second mktemp now removes DIR before exiting.

tests/scheme/audit/primitives_srfi160-audit.scm
```diff
@@ -814,21 +815,22 @@
...
-;;; an unreadable `#<...>` form. If the read syntax is ever added, these
-;;; are the assertions to revisit.
+;;; SRFI 4's external representation (#TAG( ... ) for the flat kinds,
+;;; SRFI 160's optional extension for c64/c128) is what `write` produces
+;;; and what `read` accepts for every kind, kaappi#2548: the written form
+;;; round-trips through read bit-exactly, which the native tier's
+;;; print-and-re-read constant embedding relies on.
 ;;; ------------------------------------------------------------------
 
 (test-equal "#u8(1 2 3)" (write-to-string (u8vector 1 2 3)))
```
baijum: Not from this PR, but since the two lines under it just got names: this `#u8(1 2 3)` assertion is the last unnamed one in the block, and an unnamed SRFI 64 failure prints a bare `#f`. Feel free to leave it — a one-word name would round the block off if you touch the file again.

baijum (commented):
Fourth pass, 684e056a vs 734eb5c6. The delta is three test-only changes and all are right: the seam test now asserts both halves (a literal leaf under 1023 parens reads, under 1024 is NestingTooDeep — and it checks the positive read is non-null rather than discarding it), the two remaining unnamed write-to-string assertions in the SRFI 160 audit have names, and the compile script frees `DIR` if the second `mktemp` fails.

Re-verified on this head: the depth-gate and 2548 unit filters pass, the compile script passes (3/3), and the SRFI 160 audit runs clean (1096 expected passes). Nothing further from me — the PR is in good shape to merge once CI on this head is green. (GitHub does not allow an approval from the author's account, hence a comment review.)

tests/scheme/audit/primitives_srfi160-audit.scm
baijum: Added in the head commit — the block's assertions all carry names now.

baijum (commented):
Fifth pass, 22213c01 vs 684e056a: a one-line delta — the last unnamed `write-to-string` assertion in the SRFI 160 audit block now has a name, which was the only remaining note from the previous round. No `src/` change, so the previously verified binary is the one this head builds. The audit file runs clean on this head (1096 expected passes) and it no longer has any unnamed literal-shaped `test-equal` in it.

Nothing left from my side; ready to merge once CI on this head is green.

Issue #2434, opened 2026-08-30
build_lock can hand a shared prefix to a second writer when a killed script orphans its zig build


baijum:
Raised during PR #2432 review and deliberately left out of that PR, which fixed the corruption path it caused but not the underlying race. Line numbers are against `4740b54`.

## What happens

`build_lock` (`tests/scheme/shell-common.sh`) steals a lock whose recorded holder is no longer alive:

```bash
    while ! mkdir "$lock" 2> /dev/null; do
        holder=$(cat "$lock/pid" 2> /dev/null || true)
        if [ -n "$holder" ] && ! kill -0 "$holder" 2> /dev/null; then
            rm -rf "$lock"
            continue
        fi
```

The steal equates *"the recorded pid is dead"* with *"no work is in flight"*. That does not hold when the holder is killed rather than exiting: `run-all.sh` kills a shell script that overruns `SHELL_TIMEOUT`, but the `zig build` it forked is not necessarily reaped with it. The next waiter then sees a dead pid, steals the lock, and starts its own build while the orphan is still writing.

For `fixture_interpreter` and `bundle_fixture_binary` the contended resource is a **shared install prefix** (`.zig-cache/kaappi-test-fixtures/`), so two concurrent installs can interleave into the same directory. The observable result is a half-written interpreter or bundled binary, surfacing as an unrelated-looking failure in whichever script loses — the class of thing that reads as a flake and gets re-run rather than diagnosed.

## What PR #2432 already fixed

Two contributing pieces are gone:

- `fixture_interpreter` no longer prints its path, so callers no longer wrap it in a command substitution. Bash keeps `$$` at the *outer* script's pid inside one, so the pid `build_lock` recorded used to name a process that was not the one holding the lock — the liveness check was reasoning about the wrong process entirely.
- `build_unlock` is now ownership-checked. A bare `rm -rf` let a late holder whose lock had already been stolen delete the **new** holder's lock and admit a third writer.

What remains is the steal itself.

## Suggested fix

The review's own second option: make the timeout kill the whole worker tree, so a dead pid really does mean no work in flight. `run_shell_suite`'s timeout path (`run-all.sh`) kills the script pid only; running each script in its own process group and signalling the group would cover the forked `zig build`.

[...]

PR #2557, opened 2026-09-08
Add a per-script timeout for shell test scripts (#2555)

### LINKED ISSUE #2555: bundle-cpu-baseline-2515.sh sits close enough to the 600s shell-test cap that a cleared Zig cache times it out
`tests/scheme/compile/bundle-cpu-baseline-2515.sh` timed out at the 600s
`KAAPPI_SHELL_TEST_TIMEOUT` on the **ubuntu-latest / ReleaseFast** leg during the
v0.27.0 tag push, cancelling the CI run and failing the release workflow's
`ci-gate`. Nothing is wrong with the test's *behaviour* — it exercises #2515
correctly and passes when given the time. The problem is that its cost now has
little headroom under the cap, so an unlucky runner tips it over.

## What happened

CI run [34207033778](https://github.com/kaappi/kaappi/actions/runs/34207033778),
job `test (ubuntu-latest, ReleaseFast)`:

```
TIMEOUT  tests/scheme/compile/bundle-cpu-baseline-2515.sh  (killed after 600s)
Scheme files: 758 pass, 0 fail, 1 timeout, 1 skipped
```

The same job's log shows the Zig cache was discarded partway through the run:

```
Cache directory reached 2945272600 bytes, exceeding limit of 2147483648 bytes; clearing cache
```

so the bundle builds started cold.

## Why this is a flake and not a regression

Worth spelling out, because the same shape could be a real failure:

- The test **passed on 4 of the 5 legs** of that same run — including the
  *slower* Debug leg. Only ReleaseFast failed.
- The commit under test (`21297ad1`, the v0.27.0 release commit) changed
  **one non-Markdown line**: the version string in `build.zig.zon`. Its parent
  `699ab8ec` had passed the identical CI, ReleaseFast included, 25 minutes
  earlier ([run 34198351907](https://github.com/kaappi/kaappi/actions/runs/34198351907)).
- On rerun the same leg reported `PASS`, 2154 pass / 0 fail.

## Why the margin is thin

The script performs three full bundle builds — the default `-Dbundle`, then
`build_variant baseline -Dcpu=baseline` and `build_variant native -Dcpu=native`
— plus a `zig build-exe` CPU probe. Each bundle build compiles the whole
interpreter.

Measured locally on Apple Silicon with a **warm** cache: **116s**. That is the
floor on fast hardware with nothing to rebuild. A shared GitHub runner starting
from a cleared cache does all of that work from scratch, and 600s stops being
comfortable.

## Options

1. Give this one script a longer per-test timeout (the runner already supports a
   per-script override) — smallest change, keeps the coverage intact.
2. Drop one of the three builds. Assertion 1 compares default against
   `-Dcpu=baseline`, and assertion 2 needs `-Dcpu=native` to show host tuning is
   still reachable; both look load-bearing for #2515, so this likely trades real
   coverage for time.
3. Reduce the per-build cost (e.g. a smaller bundled program) if the bundled
   source is what dominates rather than the interpreter compile.

(1) looks right to me, but I have not measured which of the three builds
dominates on a cold cache.

## Priority

`low` under the rubric's "the behaviour is right" test — the code does the
correct thing and the test asserts the correct property; only its CI reliability
is wrong. Flagging that it is not *costless*: this one cancelled a release run
and required reruns of both CI and `ci-gate`, so it is a low-priority issue with
a sharp edge rather than a purely cosmetic one.

Found while cutting v0.27.0.

baijum:
Closes #2555.

## Summary

`bundle-cpu-baseline-2515.sh` was killed at the flat 600s `KAAPPI_SHELL_TEST_TIMEOUT` on the v0.27.0 tag push (ubuntu-latest / ReleaseFast), cancelling the release `ci-gate` — GH Actions had discarded the Zig cache mid-run for size, so the script's three full ReleaseSafe builds (fixture interpreter, default bundle, `-Dcpu=native` twin) all ran cold. The test's behaviour and its assertions are correct and load-bearing for #2515; only its budget was wrong. This implements the issue's option 1: a per-script timeout override, the shell-suite twin of the `.scm` runner's `PER_FILE_TIMEOUTS`.

- `tests/scheme/run-all.sh`: new `PER_SCRIPT_TIMEOUTS` table + `shell_timeout_for()`, consulted by `run_shell_worker` before every launch. The one entry gives `bundle-cpu-baseline-2515.sh` **1200s** — twice its measured all-cold total (~600s of builds on an idle 4-core runner), so a genuine hang still dies at 20 minutes, inside every job cap that runs this suite. Tunable via `KAAPPI_BUNDLE_CPU_BASELINE_TIMEOUT`. A killed script's report now prints the budget that actually applied instead of hardcoding `SHELL_TIMEOUT`.
- `tests/scheme/test-runner/shell-timeout-override.sh`: regression test. It drives run-all.sh's **own** dispatch functions (sed-extracted, the technique `runner-agreement.sh` uses for its verdict regex) against sleeping fixtures with second-scale budgets, and pins three things, each of which fails without the fix: the routing (a script that outlives the global but fits its override must PASS — the v0.27.0 incident at 1/300th scale), the bounding (past its own budget the listed script still dies, and every other script still lives under the global), and the reporting (the kill time names the budget that applied). Needs no kaappi binary and no Zig, so it runs on every bash-capable leg.
- `.github/workflows/ci.yml`: the Debug-900 comment block was falsified by this very flake ("ReleaseSafe legs stay at 600, where this has not flaked") — rewritten to record the v0.27.0 incident and point at `PER_SCRIPT_TIMEOUTS`. The env values themselves are unchanged: the 900/600 split stays as generic Debug headroom for every other script, and the per-script entry supersedes it for this one.
- `docs/dev/test-runner.md`: documents the table, the sizing, and the env knob.

Options 2 (drop a build) and 3 (shrink the bundled program) were rejected per the issue: both assertions the script makes are load-bearing for #2515, and the interpreter compile — not the bundled source — dominates each build.

## Checklist

- [ ] `zig build test` passes — not run: no `.zig` file changed (runner/docs/CI only)
- [x] `zig fmt --check src/` passes — no `src/` change to format
- [x] Scheme test suites pass — full `run-all.sh` not run; the changed dispatch is exercised by the whole `tests/scheme/test-runner/` suite (9/9 green through the same worker code) plus the new test, under both bash 5 and macOS bash 3.2
- [x] New tests added for new behavior — and verified to fail without the fix (stash → guard fires; flat-timeout rewiring → behavioural assertion fires)
- [x] Documentation updated (test-runner.md, ci.yml comment)
- [x] Commit body explains *why* — no `CHANGELOG.md` edit

### DIFF
--- a/tests/scheme/run-all.sh
+++ b/tests/scheme/run-all.sh
@@ -92,6 +92,33 @@ timeout_for() {
     echo "$TIMEOUT"
 }
 
+# The shell-suite twin of the list above, for scripts that legitimately outlive
+# SHELL_TIMEOUT. bundle-cpu-baseline-2515.sh is the one entry so far: warm it
+# is ~116s of cache hits, but its premise is three full -Doptimize=ReleaseSafe
+# builds (fixture interpreter, default bundle, -Dcpu=native twin), and when the
+# Zig cache is cold — the Debug leg always, and any leg the moment GH Actions
+# discards the cache mid-run for size — that is ~600s of builds alone on an
+# idle 4-core runner, leaving nothing for load. The flat 600s cap killed it on
+# PR #2538 and again on the v0.27.0 tag push, the second time cancelling the
+# release ci-gate over a commit whose only code change was a version string
+# (kaappi#2555). A leg-wide bump is the wrong tool — every other script fits
+# the shared budget — so this one script gets its own, at twice the measured
+# all-cold total. A genuine hang still dies at 20 minutes, inside every job
+# cap that runs this suite.
+PER_SCRIPT_TIMEOUTS="bundle-cpu-baseline-2515.sh:${KAAPPI_BUNDLE_CPU_BASELINE_TIMEOUT:-1200}"
+
+shell_timeout_for() {
+    local base entry
+    base=$(basename "$1")
+    for entry in $PER_SCRIPT_TIMEOUTS; do
+        if [[ "$base" == "${entry%%:*}" ]]; then
+            echo "${entry##*:}"
+            return
+        fi
+    done
+    echo "$SHELL_TIMEOUT"
+}
+
 # How many .scm files to run at once. Each file is a fresh interpreter with no
 # shared state (see tests/scheme/CLAUDE.md), so they parallelise cleanly — the
 # suite's own audit found no cross-file collisions on fixed paths or ports.
@@ -334,7 +361,9 @@ run_suite() {
 # Same shape as run_file_worker: record the verdict, never touch the counters.
 run_shell_worker() {
     local script="$1" slot="$2"
-    local pid status
+    local pid status tmo
+    # The budget is per script (PER_SCRIPT_TIMEOUTS), not the flat SHELL_TIMEOUT.
+    tmo=$(shell_timeout_for "$script")
     # Launch under `set -m` so the script leads its own process group (pgid ==
     # pid): a shell script can fork a whole `zig build`, and on timeout we must
     # signal that build too. Killing the script pid alone leaves the build
@@ -349,7 +378,7 @@ run_shell_worker() {
     KAAPPI="$KAAPPI" bash "$script" "$KAAPPI" > "$slot.out" 2>&1 &
     pid=$!
     set +m
-    if wait_with_timeout "$pid" "$SHELL_TIMEOUT"; then
+    if wait_with_timeout "$pid" "$tmo"; then
         status=0
         wait "$pid" || status=$?
     else
@@ -358,7 +387,7 @@ run_shell_worker() {
         # with the script rather than racing the next writer's install.
         kill -- "-$pid" 2>/dev/null || kill "$pid" 2>/dev/null || true
         wait "$pid" 2>/dev/null || true
-        echo "TIMEOUT" > "$slot.rec"
+        echo "TIMEOUT ${tmo}s" > "$slot.rec"
         return 0
     fi
     case $status in
@@ -384,8 +413,10 @@ report_shell_result() {
             echo "  SKIP  $script"
             SKIPPED=$((SKIPPED + 1))
             ;;
-        TIMEOUT)
-            echo "  TIMEOUT  $script  (killed after ${SHELL_TIMEOUT}s)"
+        TIMEOUT*)
+            # ${kind#TIMEOUT } is the "<n>s" the worker recorded, which for a
+            # PER_SCRIPT_TIMEOUTS entry is not SHELL_TIMEOUT.
+            echo "  TIMEOUT  $script  (killed after ${kind#TIMEOUT })"
             [[ -f "$slot.out" ]] && cat "$slot.out"
             TIMEDOUT=$((TIMEDOUT + 1))
             ;;
--- a/tests/scheme/run-all.sh
+++ b/tests/scheme/run-all.sh
@@ -103,8 +103,10 @@ timeout_for() {
 # release ci-gate over a commit whose only code change was a version string
 # (kaappi#2555). A leg-wide bump is the wrong tool — every other script fits
 # the shared budget — so this one script gets its own, at twice the measured
-# all-cold total. A genuine hang still dies at 20 minutes, inside every job
-# cap that runs this suite.
+# all-cold total. A genuine hang still dies at the 1200s tick budget —
+# wait_with_timeout counts sleep 0.05 ticks, not wall clock, so that is
+# ~20 min of wall clock on Linux and stretches further where sleep spawns
+# cost more — inside every job cap that runs this suite.
 PER_SCRIPT_TIMEOUTS="bundle-cpu-baseline-2515.sh:${KAAPPI_BUNDLE_CPU_BASELINE_TIMEOUT:-1200}"
 
 shell_timeout_for() {
--- a/tests/scheme/test-runner/shell-timeout-override.sh
+++ b/tests/scheme/test-runner/shell-timeout-override.sh
@@ -1,0 +1,152 @@
+#!/bin/bash
+# The shell suites' per-script timeout override (kaappi#2555).
+#
+# bundle-cpu-baseline-2515.sh does three full -Doptimize=ReleaseSafe builds
+# when the Zig cache is cold, and the flat KAAPPI_SHELL_TEST_TIMEOUT cap
+# (600s on the Release legs) killed it on the v0.27.0 tag push — a commit
+# whose only code change was a version string — cancelling the release
+# ci-gate. The fix was not a smaller test but a per-script budget: the shell
+# workers consult PER_SCRIPT_TIMEOUTS exactly the way the .scm workers
+# consult PER_FILE_TIMEOUTS.
+#
+# This pins the mechanism by driving run-all.sh's own dispatch functions
+# against fixture scripts that only sleep, with both budgets dialed down to
+# seconds. The functions are extracted from run-all.sh with sed rather than
+# copied here, so the shipping code is what runs — the same technique
+# runner-agreement.sh uses to keep its verdict regex honest.
+#
+# No kaappi binary and no zig are needed, so this runs on every leg that can
+# run bash, including the ones with no toolchain.
+
+set -euo pipefail
+
+RUN_ALL="$(cd "$(dirname "$0")/.." && pwd)/run-all.sh"
+if [[ ! -f "$RUN_ALL" ]]; then
+    echo "FAIL: cannot find run-all.sh next to this suite" >&2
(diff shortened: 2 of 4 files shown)


baijum (commented):
Ran the new test locally and traced the runner change; the routing, the record, and the report all do what the description says, and the guard does fire against `main`'s `run-all.sh` (`FAIL: run-all.sh has no PER_SCRIPT_TIMEOUTS entry`). Two things need a change before merge, the rest are nits:

1. **The new test fails deterministically on macOS** (3 of 3 runs here, Apple Silicon, bash 3.2 and 5). `wait_with_timeout` counts `sleep 0.05` ticks rather than wall time, so on a platform with ~18ms per spawn the 6s budget becomes ~8.2s and the 8s fixture finishes first, recording `PASS`. Details and a verified fix inline (let the over-budget fixtures sleep 60). FreeBSD already passed because its spawn overhead is small; I'd expect the still-pending `test (macos-latest, ReleaseSafe)` leg to go red the same way.
2. **DCO is red**: the commit has no `Signed-off-by` trailer (`git commit --amend -s`).

+1 to the CodeRabbit wording fix on the ci.yml comment (`when the Zig cache is --` is missing its `cold`).

tests/scheme/test-runner/shell-timeout-override.sh
```diff
@@ -0,0 +1,152 @@
...
+mk_fixture bundle-cpu-baseline-2515 3
+if [[ "$(run_fixture listed "$DIR/bundle-cpu-baseline-2515.sh")" != "PASS" ]]; then
+    echo "FAIL: script killed inside its per-script budget (was the worker wired to shell_timeout_for?)" >&2
+    exit 1
+fi
+
+# The override is a budget, not a waiver: past it, the listed script dies too.
+# A second copy under slow/ so both basename matches can exist at once.
+mkdir -p "$DIR/slow"
+mk_fixture slow/bundle-cpu-baseline-2515 8
```
baijum: This is the assertion that fails on macOS: the verdict comes back `PASS` and the case takes the full ~8s.

`wait_with_timeout` counts ticks of `sleep 0.05`, not wall time, so its budget stretches by the per-spawn cost of `sleep`. Measured on this Mac:

| nominal budget | ticks | wall |
|---|---|---|
| 6s | 120 | 8.17s |
| 2s | 40 | 2.65s |

An 8s fixture under a "6s" budget finishes before the 120th tick, so the worker never reaches the kill. Same result on three consecutive runs.

The fix is to stop racing the drift rather than tune it: an over-budget fixture is killed at the budget whatever it would have slept, so let it sleep far past it. `sleep 60` here (and on line 125) turns a 33% margin into 10x. Verified with exactly those two edits: the whole file passes twice in a row here, 15.6s wall, and the process-group kill leaves no orphaned `sleep` behind.

tests/scheme/test-runner/shell-timeout-override.sh
```diff
@@ -0,0 +1,152 @@
...
+if [[ "$(run_fixture listed-slow "$DIR/slow/bundle-cpu-baseline-2515.sh")" != "TIMEOUT 6s" ]]; then
+    echo "FAIL: listed script was not killed (and recorded) at its own budget" >&2
+    exit 1
+fi
+
+# Every other script still lives under the global budget...
+mk_fixture unlisted-slow 4
```
baijum: Same shape: `4` under a 2s budget is 1.35s of margin on this machine (40 ticks = 2.65s) and it only shrinks under load. `60` here too.

tests/scheme/test-runner/shell-timeout-override.sh
```diff
@@ -0,0 +1,152 @@
...
+# (600s on the Release legs) killed it on the v0.27.0 tag push — a commit
+# whose only code change was a version string — cancelling the release
+# ci-gate. The fix was not a smaller test but a per-script budget: the shell
+# workers consult PER_SCRIPT_TIMEOUTS exactly the way the .scm workers
+# consult PER_FILE_TIMEOUTS.
+#
+# This pins the mechanism by driving run-all.sh's own dispatch functions
+# against fixture scripts that only sleep, with both budgets dialed down to
+# seconds. The functions are extracted from run-all.sh with sed rather than
+# copied here, so the shipping code is what runs — the same technique
```
baijum: Nit: `runner-agreement.sh` doesn't sed-extract anything. It carries a *copy* of the net regex and `grep -Fq`s that the copy still appears verbatim in `run-all.sh`. Extracting the real functions is the stronger technique; just don't attribute it to that file.

tests/scheme/test-runner/shell-timeout-override.sh
```diff
@@ -0,0 +1,152 @@
...
+# runner-agreement.sh uses to keep its verdict regex honest.
+#
+# No kaappi binary and no zig are needed, so this runs on every leg that can
```
baijum, line 19: Worth watching on the two Windows legs (both still pending). They run `tests/scheme/test-runner/*.sh` directly with a plain `bash "$s"`, never through `run-all.sh`, so this file is the first time `run_shell_worker`'s `set -m` + `kill -- "-$pid"` path executes under Git Bash. `smoke/build-lock-orphan-2434.sh` skips there precisely because process-group kill is "not reliable under Git Bash/MSYS". The `|| kill "$pid"` fallback should still yield the `TIMEOUT` verdict (an orphaned `sleep` is harmless), so it may well pass; if either leg goes red, the precedent is `. "$(dirname "$0")/../shell-common.sh"` + `skip_on_windows`.

tests/scheme/run-all.sh
```diff
@@ -92,6 +92,33 @@ timeout_for() {
...
+# is ~116s of cache hits, but its premise is three full -Doptimize=ReleaseSafe
+# builds (fixture interpreter, default bundle, -Dcpu=native twin), and when the
+# Zig cache is cold — the Debug leg always, and any leg the moment GH Actions
+# discards the cache mid-run for size — that is ~600s of builds alone on an
+# idle 4-core runner, leaving nothing for load. The flat 600s cap killed it on
+# PR #2538 and again on the v0.27.0 tag push, the second time cancelling the
+# release ci-gate over a commit whose only code change was a version string
+# (kaappi#2555). A leg-wide bump is the wrong tool — every other script fits
+# the shared budget — so this one script gets its own, at twice the measured
+# all-cold total. A genuine hang still dies at 20 minutes, inside every job
```
baijum: Nit on the figure only: because `wait_with_timeout` counts `sleep 0.05` ticks, the effective budget is nominal x (1 + spawn overhead / 0.05). That is a few percent on Linux but ~36% on macOS here (120 ticks = 8.17s), so on the macOS leg a hang dies at ~27 min, not 20. Still inside that leg's job cap, so nothing to change in the value; maybe just say "1200s of ticks (~20 min on Linux)" so the next reader doesn't take the 20 as exact.

tests/scheme/test-runner/shell-timeout-override.sh
baijum: Applied — `sleep 60` on both over-budget fixtures, with a comment recording the drift numbers so the next reader doesn't retune it back. Confirmed your diagnosis against the CI logs: all three red legs (macOS + both Windows) fail this same assertion with the fixture recording PASS, i.e. the drift race, not the Git Bash kill path. File passes twice here too, 15.5–16.0s wall.

tests/scheme/test-runner/shell-timeout-override.sh
baijum: Applied — 60 here as well, same reasoning.

tests/scheme/test-runner/shell-timeout-override.sh
baijum: Fixed — the header now says extraction-and-execution is one step further than runner-agreement.sh, which keeps a copy and greps that it still appears verbatim.

tests/scheme/test-runner/shell-timeout-override.sh
baijum: Your prediction held in an unexpected way: both Windows legs did go red, but from the tick drift, not the kill — the 8s fixture completed inside the stretched 120-tick window and recorded PASS (same failure as the macOS leg). With `sleep 60` the kill path now genuinely executes under Git Bash; if either leg still disagrees I'll take the `skip_on_windows` route per the #2434 precedent.

tests/scheme/run-all.sh
baijum: Reworded in both run-all.sh and docs/dev/test-runner.md: "the 1200s tick budget — ~20 min of wall clock on Linux, stretches further where sleep spawns cost more". No change to the value.

PR #2558, opened 2026-09-08
Call SRFI 231's bundled implementation what the SRFI calls it

baijum:
We described the implementation Gambit bundles as SRFI 231's **"reference implementation"**, and wrote up *"when prose and reference code disagree, trust the code"* as **this SRFI's documented rule** — in `docs/dev/srfi-implementation-notes.md`, `docs/dev/testing.md`, and `tools/srfi231_diff.py`'s docstring.

Neither is right, and I checked the document rather than taking it on faith:

| | |
|---|---|
| `"sample implementation"` in srfi-231.html | **15** |
| `"reference implementation"` | **0** |
| a "trust the code" rule | **not present** |

The heuristic was ours. Calling it *documented* asserted something about the spec that isn't in it. Brad Lucier (the SRFI's author) [corrected us on r/KaappiScheme](https://github.com/kaappi/kaappi/pull/2541): the bundled code is a *sample* implementation, and "generally speaking the document is the specification."

## What changed

- **The rule paragraph says what it actually is** — a working heuristic adopted after two real prose/code divergences (`check-nested-list`'s dimension-0 case; `array-inner-product`'s missing `array-curry` argument) — and records that the prose governs unless there's positive reason to think it's in error.
- **The differential tester's oracle is demoted to match.** A Kaappi/oracle divergence is a lead to check against the prose, not a verdict on its own. Its CLI is unchanged.
- **Terminology follows** across `lib/srfi/231/*.sld`, `docs/dev/testing.md` and the tool.

Most of the diff is reflow: "sample implementation" is 12 characters longer than "reference", which pushed 30 comment lines past these files' 79-column wrap. **No code changed** — `git diff` on the `.sld` files touches only `;;` lines.

## This puts #2543 in question, and the note now says so

That PR added `(not (real? x))` to the c64/c128 checkers so a bare real flonum is rejected — purely to reproduce the sample implementation's verdict under Gambit, where `(imag-part 1.0)` is an exact `0`. It was justified by the rule this PR retracts.

The normative prose says `cX-storage-class` procedures "manipulate complex numbers with, respectively, 32- and 64-bit floating-point numbers as real and imaginary parts". Under Kaappi's representation `1.0` **is** one — verified: `(complex? 1.0)` → `#t`, `(real-part 1.0)` → `1.0`, `(imag-part 1.0)` → `0.0`, both inexact. And the one normative constraint on a checker, that `(checker (getter v i))` be `#t`, holds either way, since the getter returns `1.0+0.0i`.

So rejecting a bare real is a **portability choice** — code written against Kaappi won't silently break on Gambit — not conformance. **Left in place and asked upstream**; this PR makes no behavior change. The note records the reasoning so the answer can be applied without re-deriving it.

## Verification

- All eight SRFI 231 suites pass; the official suite reports **10936 passed, 0 unexpected failures**
- `tools/srfi231_diff.py` parses and its `--help` is unchanged
- `markdownlint-cli2`: 0 issues

## Not in scope

Other SRFIs' notes also say "reference implementation". Many SRFIs do use that term for their own bundled code, so this is scoped to 231 — where the author told us the term is wrong — rather than blanket-renamed. The released v0.27.0 changelog entry and its GitHub release body carry the phrase too; leaving those, since editing a published release body is a separate call.

### DIFF
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -873,7 +873,7 @@ textual relationship, per its own spec) shipped across 6 phases (issues
 tracked under #1694;
 `lib/srfi/231/{misc,intervals,storage-classes,arrays,views,combinators,assembly}.sld`,
 merged into a public `lib/srfi/231.sld` re-export hub — 118 bindings total, an
-exact bijection confirmed against the reference implementation's own export
+exact bijection confirmed against the sample implementation's own export
 clause) — the largest single SRFI in this codebase by an order of magnitude.
 An `<interval>` is two parallel exact-integer vectors (lower/upper bounds,
 arbitrary — including negative — per axis), not 25/164's shape-of-pairs nor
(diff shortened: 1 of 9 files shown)


tools/srfi231_diff.py
```diff
@@ -1,15 +1,19 @@
...
-"""Differential testing of Kaappi's SRFI 231 against the reference implementation.
-
-SRFI 231's documented rule in this codebase is "when the spec's prose and its
-reference implementation disagree, trust the code" -- which makes the reference
-an oracle in the strict sense. Gambit bundles it as `(srfi 231)`, and chibi
-ships an independent port. This tool generates random SRFI 231 programs in the
+"""Differential testing of Kaappi's SRFI 231 against the sample implementation.
+
+"Sample implementation" is the SRFI's own term -- the document uses it
+throughout and never says "sample implementation" -- and its author has
```
baijum: This sentence contradicts itself as written: the document *uses* "sample implementation" throughout, so the phrase it never says has to be **"reference implementation"**. Presumably a casualty of the rename pass.

tools/srfi231_diff.py
```diff
@@ -1,15 +1,19 @@
...
+is a lead to check against the prose, and where the two genuinely disagree
+the prose governs unless there is positive reason to think it is an error.
+Gambit bundles the sample implementation as `(srfi 231)`, and chibi ships an
+independent port. This tool generates random SRFI 231 programs in the
 subset of R7RS all three run, executes each under Kaappi and an oracle, and
 reports any program whose printed output differs.
 
 Why this exists: the official conformance suite (10,936 evaluations) encodes
-the reference's answers on the inputs its author wrote down. It cannot see a
+the sample implementation's answers on the inputs its author wrote down. It cannot see a
```
baijum: This line (88 chars) and four others added by this PR — lines 93, 97, 419, 732, at 81–87 chars — now exceed the 79-column wrap the rest of the file holds to (and that the PR description says the diff reflows to). Worth wrapping these five too so the file stays uniform.

lib/srfi/231/assembly.sld
```diff
@@ -117,21 +117,21 @@
...
-    ;; Takes a "curried" array of arrays (all inner arrays sharing one
-    ;; common domain -- actively validated here, matching the reference
-    ;; implementation's stricter-than-the-bare-spec behavior) and returns
-    ;; a single specialized array with domain (interval-cartesian-product
-    ;; outer-domain inner-domain) -- outer axes first, inner axes second.
-    ;; An inverse to array-curry; a multi-dimensional version of
-    ;; array-stack.
+    ;; Takes a "curried" array of arrays (all inner arrays sharing one common
+    ;; domain -- actively validated here, matching the sample implementation
+    ;; implementation's stricter-than-the-bare-spec behavior) and returns a
```
baijum, line 122: `...matching the sample implementation` / `implementation's stricter-than-the-bare-spec behavior` — the rename produced "the sample implementation implementation's". Dropping the second word ("matching the sample implementation's stricter-than-the-bare-spec behavior") fixes it.

lib/srfi/231/arrays.sld
```diff
@@ -111,14 +111,12 @@
...
     ;; Returns the checker to apply per element, or #f when the check is
-    ;; provably vacuous:
-    ;;   - a generic destination manipulates every value; or
-    ;;   - source and destination share a storage class, so every value
-    ;;     read out of the source body already satisfies the destination's
-    ;;     checker by construction.
-    ;; (The reference goes further with a widening table -- u8 into u32 and
-    ;; friends. Not needed for the paths here; the two cases above already
-    ;; cover same-class copies and generic destinations.)
+    ;; provably vacuous: - a generic destination manipulates every value; or -
```
baijum: The reflow merged the two sub-bullets into one run-on line (`vacuous: - a generic destination ...; or - source and destination ...`). Keeping one bullet per line, as before, reads much better and costs a line or two at most.

lib/srfi/231/arrays.sld
```diff
@@ -148,19 +146,20 @@
...
           (error "make-array: setter must be a procedure or #f" setter))
-        ;; The reference wraps EVERY generalized array's getter/setter in
-        ;; index checks (%%make-safer-array, generic-arrays.scm): out-of-
-        ;; domain, wrong-arity, and empty-domain calls all error there, and
-        ;; the official suite tests it -- plain arrays are not exempt
-        ;; (#2362). Valid accesses are unaffected. Plain arrays have no
-        ;; storage-class checker, so the setter's value check is vacuous.
+        ;; The sample implementation wraps EVERY generalized array's
+        ;; getter/setter in index checks (%%make-safer-array,
+        ;; generic-arrays.scm): out-of- domain, wrong-arity, and empty-domain
```
baijum: `out-of- domain` — the rewrap split the hyphenated compound and left the hyphen dangling at the end of the line. Moving the whole compound down fixes it.

lib/srfi/231/views.sld
```diff
@@ -310,62 +310,58 @@
...
-    ;; cons per element -- but the copy-out no longer regenerates a
-    ;; multi-index per element: a fresh destination's body IS the
+    ;; array-copy must be call/cc safe -- the spec's term for "does not modify
+    ;; the state of any data captured by a continuation", the whole documented
+    ;; difference from array-copy!, whose getter continuations it is an error
+    ;; to invoke more than once. A getter that captures a continuation and
+    ;; re-invokes it after the copy returned must get a FRESH array, computed
+    ;; from the values ITS run had collected at the capture point, and the
+    ;; array the first call already returned must never mutate under its
+    ;; holder. Two shapes fail that, and both have shipped here: * a direct
```
baijum: Same as in `arrays.sld`: the two `*` bullets — the two shapes that actually shipped and failed (pre-#2454 direct fill; #2454..#2539 scratch vector) — are now inline in the paragraph. This comment is the historical record for why `array-copy` looks the way it does; I'd keep the bullet structure through the rewrap.

docs/dev/srfi-implementation-notes.md
```diff
@@ -961,12 +974,22 @@ full per-axis width-consistency validation (reusing
...
+sample implementation. The SRFI's own prose pseudocode disagreed with its
+sample implementation at least twice (`check-nested-list`'s dimension-0
 case returns `'()`, not the prose's nonsensical `#t`; `array-inner-product`'s
-prose omits a required `array-curry` argument the reference code supplies) —
-confirming "when this SRFI's prose and its reference implementation disagree,
-trust the code" as a load-bearing rule for this SRFI specifically.
+prose omits a required `array-curry` argument the sample code supplies), so
+where the two diverge we read the code to work out what was meant.
+
+**That is a working heuristic, not a rule the SRFI states.** An earlier
```
baijum, line 984: The rename this paragraph introduces doesn't propagate through the rest of the SRFI 231 section: the full phrase survives at lines 876 ("confirmed against the reference implementation's own export") and 910 ("verified by reading the reference implementation"), and bare `the reference('s)` on about a dozen more lines (889, 891, 895, 900, 928, 930, 955, 959, 961, 967, 971), plus `docs/dev/testing.md:379` ("mirroring the reference's structure"). Since the point here is that "reference" mischaracterizes this code, the shorthand quietly re-asserts it. Either rename those too, or establish the shorthand once in this paragraph (e.g. "the sample implementation, 'the sample' below") — as it stands the section reads like an incomplete sed.

baijum (commented):
Reviewed the substance as well as the reflow; the retraction itself checks out. What I verified independently:

- **Against the live `srfi-231.html`:** "sample implementation" throughout (14 occurrences by my count, HTML→text conversion dependent), "reference implementation" zero, and no prose-vs-code precedence rule stated anywhere. The `cX-storage-class` quote and the `(checker (getter v i))` constraint quoted in the new notes paragraph are accurate.
- **`git diff main...HEAD -- lib/srfi/231/` touches only `;;` lines** — the "no code changed" claim holds exactly.
- **markdownlint-cli2: 0 issues** on both changed docs; the tool compiles and its CLI surface (flags, subcommands) is unchanged. One nitpick on the PR description: `--help` is not literally unchanged — the description line changed along with the docstring headline, which is of course the intent.

Two text bugs to fix before merge (docstring self-contradiction, duplicated word in `assembly.sld`), a handful of reflow nits, and an incomplete rename inside the 231 notes section — details inline. The new "working heuristic, not a rule" paragraph and the demoted-oracle framing in `testing.md` are the right shape: the heuristic is now described as what it is, and the c64/c128 "portability choice, not conformance" writeup records the reasoning where the next person will look for it.

baijum:
All seven addressed in b0d5c6e → the new head. Thanks for reading the substance and not just the rename — four of these were damage the mechanical reflow did, which a diff-skim would have waved through.

**The docstring self-contradiction** was the rename pass rewriting the quoted phrase inside my own sentence: it ended up claiming the document "never says *sample implementation*" when the phrase it never says is *reference implementation*. Fixed.

**`assembly.sld`'s "implementation implementation's"** — the rename hit a line already wrapped mid-phrase. Same root cause as the incomplete rename below: wrapped phrases are invisible to a single-line grep.

**Both flattened bullet lists restored** (`arrays.sld`, `views.sld`). You're right that `views.sld`'s is the point of the comment, not decoration — it's the record of the two `array-copy` shapes that shipped and failed, and a run-on sentence buries exactly the distinction a reader goes there for.

**`out-of- domain`** fixed, and the surrounding block reflowed rather than left ragged.

**The five 81–88 char lines** in the tool wrapped to 79.

**The incomplete rename** — this was the one worth catching. Renamed throughout rather than defining a shorthand: the section argues that "reference" mischaracterizes this code, so leaving the shorthand in place asks the reader to hold an exception in mind while reading the argument against it. Several of the surviving sites were wrapped mid-phrase (`the` ending one line, `reference's` starting the next), which is why the first pass missed them.

On your nitpick: fair, `--help` is not literally unchanged — the description line moved with the docstring headline. I meant the flags and subcommands, and should have said so.

Re-verified: eight SRFI 231 suites pass, official suite **10936 passed / 0 unexpected failures**; `git diff main -- lib/srfi/231/` still touches only `;;` lines; tool parses, flags and subcommands unchanged; markdownlint 0 issues.

PR #2559, opened 2026-09-09
Accept bare real flonums in the c64/c128 checkers again

baijum:
Reverts the `(not (real? x))` clause [#2543](https://github.com/kaappi/kaappi/pull/2543) added, restoring the sample implementation's checker line verbatim.

## Why

#2543 rejected a bare real flonum so Kaappi's verdict would match Gambit's. Both run the same checker text; they disagreed only because Gambit's `(imag-part 1.0)` is an exact `0` and ours is `0.0`.

Asked about it, Brad Lucier — who suggested the exact-0 rule in the first place — confirmed **R6RS adopted it and "that's not what R7RS small does"**, adding that implementers can argue either way.

Kaappi is R7RS-small, which leaves that exactness open. So under our numeric tower `1.0` is exactly what the normative prose asks a `cX-storage-class` to manipulate — "complex numbers with, respectively, 32- and 64-bit floating-point numbers as real and imaginary parts":

```scheme
(complex? 1.0)    ⇒ #t
(real-part 1.0)   ⇒ 1.0   ; inexact
(imag-part 1.0)   ⇒ 0.0   ; inexact
```

The one normative constraint on a checker — that `(checker (getter v i))` be `#t` — holds either way, since the getter returns `1.0+0.0i`. So the restriction was never conformance, only agreement with one other implementation's tower, and it cost a legal program the ability to copy real data into a complex array.

The justification #2543 rested on was *"when prose and reference code disagree, trust the code"* — the invented rule [#2558](https://github.com/kaappi/kaappi/pull/2558) just retracted.

## Guarding against re-derivation

Gambit will now disagree with us on c64/c128 in the differential tester **permanently**. Two places say so, because both are where the next person would otherwise re-derive #2542 and re-apply #2543:

- `tools/srfi231_diff.py`'s known-divergence block, upgraded from "was a bug, now fixed" to "expected and permanent, do not re-file"
- an in-line note in the test file telling a reader not to "fix" these assertions back to `#f` to match Gambit

## Also

The two prose/code divergences #2558 cited are corrected. Brad confirmed `array-inner-product`'s missing `array-curry` argument **is** a document error and will fix it — so the document gets corrected rather than overridden — and `check-nested-list` is not exported by the library, so what it returns is nobody's business. Neither supports the heuristic as it was stated.

## Behavior change

`(array-copy <real-flonum-array> c64-storage-class)` succeeds again, as it did before v0.27.0. v0.27.0 is the only release that rejected it; the [release announcement](https://github.com/orgs/kaappi/discussions/2556) already flagged the restriction as under review.

## Tests

Checker and end-to-end `array-copy` assertions flip to the accepting behavior, with the round-trip value pinned (`1.5` → `1.5+0.0i`). Eight SRFI 231 suites pass, storage-classes at 253 expected passes, official suite **10936 passed / 0 unexpected failures**.

### DIFF
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -931,10 +931,10 @@ find (kaappi#2542) is the imag-part convention leaking through the sample
 implementation's own code: its c64/c128 checker is textually `(and (complex?
 obj) (inexact? (real-part obj)) (inexact? (imag-part obj)))`, but Gambit's
 exact-0 `(imag-part 1.0)` makes it reject a bare real flonum that Kaappi's
-equally R7RS-legal inexact `0.0` accepted — so the shipped checker carries an
-explicit `(not (real? x))`, encoding the sample implementation's verdict under
-either convention (a `1.0+0.0i` with a real inexact-zero imaginary part is
-still accepted).
+equally R7RS-legal inexact `0.0` accepted — so #2543 shipped the checker with
+an explicit `(not (real? x))` clause, encoding the sample implementation's
+verdict under either convention (a `1.0+0.0i` with a real inexact-zero
+imaginary part was still accepted).
 
 **That clause was reverted in #2559.** The two checkers are the sample
 implementation's line verbatim; the verdicts differ because Gambit's
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -941,7 +941,19 @@ implementation's line verbatim; the verdicts differ because Gambit's
 `(imag-part 1.0)` is an exact `0` and Kaappi's is `0.0`. Brad Lucier confirmed
 that exact-0 is **R6RS**'s rule — adopted there on his own suggestion — and
 "that's not what R7RS small does", adding that implementers can argue either
-way. Kaappi is R7RS-small, which leaves the exactness open, and the normative
+way. Verified against `docs/errata-corrected-r7rs.pdf` §6.2.6 rather than
+taken on report: the section's sole exactness permission is that `real-part`
+and `imag-part` "may return exact real numbers when applied to an inexact
+complex number **if** the corresponding argument passed to `make-rectangular`
+was exact", which a bare `1.0` never went through; what governs it instead is
+"`(real? z)` is true if and only if `(zero? (imag-part z))` is true", and
+`(zero? 0.0)` is `#t`, so an inexact `0.0` satisfies that rule exactly as an
+exact `0` would. Nothing in the document requires one over the other. (That
+iff rule does *not* survive contact with the inexact-zero-imaginary case —
+§6.2.6 also prints `(real? -2.5+0.0i) ⇒ #f`, which the rule would make `#t`;
+the rule and the example contradict each other and every implementation,
+Kaappi and Gambit included, follows the example.) Kaappi is R7RS-small, and
+the normative
 prose asks only for "complex numbers with, respectively, 32- and 64-bit
 floating-point numbers as real and imaginary parts", which `1.0` is under our
 tower: `(complex? 1.0)` is `#t` with `real-part` `1.0` and `imag-part` `0.0`,
--- a/docs/dev/srfi-implementation-notes.md
+++ b/docs/dev/srfi-implementation-notes.md
@@ -936,17 +936,20 @@ explicit `(not (real? x))`, encoding the sample implementation's verdict under
 either convention (a `1.0+0.0i` with a real inexact-zero imaginary part is
 still accepted).
 
-**That clause is under review (see the terminology note below), because the
-document does not require it.** The normative prose says `cX-storage-class`
-procedures "manipulate complex numbers with, respectively, 32- and 64-bit
-floating-point numbers as real and imaginary parts" — and under Kaappi's
-representation `1.0` is one: `(complex? 1.0)` is `#t` with `real-part` `1.0`
-and `imag-part` `0.0`, both inexact. The one normative constraint on a
-checker is that `(checker (getter v i))` be `#t`, which holds either way
-(the getter returns `1.0+0.0i`). So rejecting a bare real matches Gambit's
-verdict rather than the document's requirement, and it is a portability
-choice — code written against Kaappi will not silently break on Gambit —
-not conformance. Raised with the SRFI's author; revisit when he answers.
+**That clause was reverted in #2559.** The two checkers are the sample
+implementation's line verbatim; the verdicts differ because Gambit's
+`(imag-part 1.0)` is an exact `0` and Kaappi's is `0.0`. Brad Lucier confirmed
+that exact-0 is **R6RS**'s rule — adopted there on his own suggestion — and
+"that's not what R7RS small does", adding that implementers can argue either
+way. Kaappi is R7RS-small, which leaves the exactness open, and the normative
+prose asks only for "complex numbers with, respectively, 32- and 64-bit
+floating-point numbers as real and imaginary parts", which `1.0` is under our
+tower: `(complex? 1.0)` is `#t` with `real-part` `1.0` and `imag-part` `0.0`,
+both inexact. The one normative constraint on a checker, that
+`(checker (getter v i))` be `#t`, holds either way. So a bare real flonum is
+accepted again, and the differential tester's c64/c128 disagreement with
+Gambit is a permanent known divergence rather than a finding — recorded in the
+tool so it is not re-filed as #2542 was.
 
 A residual the checker cannot reach: a mixed-exactness complex (`1+2.0i`,
 `(make-rectangular 1 2.0)`) keeps an exact real part under Gambit, so the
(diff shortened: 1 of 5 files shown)


baijum (commented):
Reviewed from the branch tip (`408162c2`, identical to the PR head), re-deriving every load-bearing claim from primary sources rather than the description. The revert itself checks out. What I verified independently:

- **The restored line is the sample implementation's verbatim.** Read in Gambit's bundled copy (gambit-scheme's `lib/srfi/231/generalized-arrays.scm`, the complex checker around line 1373): exactly `(and (complex? obj) (inexact? (real-part obj)) (inexact? (imag-part obj)))`. Identical modulo the parameter name, so the verdict difference really is the tower, not the text.
- **Both R7RS quotes, against `docs/errata-corrected-r7rs.pdf` §6.2.6.** The make-rectangular permission ("may return exact real numbers when applied to an inexact complex number if the corresponding argument passed to make-rectangular was exact") and the iff rule ("(real? z) is true if and only if (zero? (imag-part z)) is true") are both there verbatim, as are the examples `(real? -2.5+0i) ⇒ #t` / `(real? -2.5+0.0i) ⇒ #f`. The rule/example tension the note leans on is real: Kaappi's `(real? 1.5+0.0i)` and Gambit's `(real? -2.5+0.0i)` are both `#f`, following the example against the rule.
- **The tower facts the argument rests on.** Kaappi: `(imag-part 1.0)` ⇒ `0.0`, `(complex? 1.0)` ⇒ `#t`. Gambit: `(imag-part 1.0)` ⇒ `0` (exact). chibi: `(imag-part 1.0)` ⇒ `0` (exact) and its checker still accepts `1.0`, which matters for one inline comment below.
- **End-to-end:** `array-copy` of a real-flonum array into c64 succeeds; `array-ref` gives `1.5+0.0i`; checker verdicts `3.5` ⇒ `#t`, `1.0+0.0i` ⇒ `#t`, `3+4i` ⇒ `#f`, `1/3` ⇒ `#f`. The flipped assertions are exactly right.
- **All eight SRFI 231 suites green** under an isolated `KAAPPI_HOME`: storage-classes 253 expected passes, official **10936 passed / 0 unexpected failures** (the single accounted divergence is test 147, mutable strings — unrelated to this change), arrays 105, intervals 91, views 118, combinators 94, assembly 71, hub 17.
- **The diff-tool paragraph, empirically:** `storage --classes c64 --count 3` mismatches 3/3, each first difference a `check` line on a bare real flonum (kaappi `#t`, gambit `#f`; the canonicalized `1` / `0` / `7737125245533627/…` readings are `1.0` / `0.0` / `0.1`). Exactly the documented shape.

Four inline comments, all non-blocking: one confirmation, one test-strength suggestion, one tool-workflow question with a documentation tail, and one tense nit.

lib/srfi/231/storage-classes.sld
```diff
@@ -127,22 +127,45 @@
...
+    ;; The SRFI's own prose asks only for "complex numbers with ...
+    ;; floating-point numbers as real and imaginary parts", which 1.0 is under
+    ;; our tower, and the author confirmed implementations may differ here. So
+    ;; the differential tester's c64/c128 disagreement with Gambit is a known
+    ;; oracle divergence, not a bug.
     (define (%flonum-checker x) (and (real? x) (inexact? x)))
     (define (%inexact-complex-checker x)
-      (and (complex? x) (not (real? x))
+      (and (complex? x)
            (inexact? (real-part x)) (inexact? (imag-part x))))
```
baijum, line 169: Verified against the primary source: Gambit's bundled sample implementation (gambit-scheme's `lib/srfi/231/generalized-arrays.scm`, ~line 1373) carries exactly `(and (complex? obj) (inexact? (real-part obj)) (inexact? (imag-part obj)))` as its complex checker — this is that line verbatim modulo the parameter name. And the one normative constraint holds under it end to end: `(checker (getter v i))` is `#t` on the copied array (`1.5` in, `1.5+0.0i` back out, confirmed by running it).

tests/scheme/srfi/srfi231-storage-classes.scm
```diff
@@ -219,19 +219,26 @@
...
-(test-assert "array-copy of real flonums into c64 is rejected (#2542)"
-             (guard (e (#t #t))
-               (array-copy (make-array (make-interval '#(2)) (lambda (i) 1.5)) c64-storage-class)
-               #f))
+;; End-to-end through the generic->typed copy branch, the only copy path
+;; that consults the checker at all (same-class copies skip it, #2448).
+;; Copying real flonums into a c64 array succeeds and round-trips, the
+;; behavior #2543 removed and #2559 restored.
+(test-equal "array-copy of real flonums into c64 succeeds (#2559)"
+            1.5+0.0i
```
baijum: Non-blocking suggestion: `1.5` is exactly representable in f32, so this pins acceptance and type but not the narrowing the copy path performs. A value like `0.1`, which round-trips to `0.10000000149011612+0.0i` (verified by running it), would pin both in one assertion — the copy succeeds *and* the value demonstrably went through the interleaved f32 body. If the intent is to keep this test purely about accept/type and leave rounding to the round-trip tests below, it's fine as is.

tools/srfi231_diff.py
```diff
@@ -95,13 +95,17 @@
...
 value of `specialized-array-default-safe?` to #t (spec and the SRFI
 repository's copy: #f); the storage mode pins it to #f in its prelude.
-kaappi#2542 (the c64/c128 checkers accepted real flonums, which the
-sample implementation rejects) used to surface as `check` mismatches on those
-two
-classes; before kaappi#2543 fixed it, the workaround was excluding them
-with `--classes`, and both are back in the default draw. Since a
-program's first difference hides everything after it, `--classes` is
-still the way to focus a run on the classes you care about.
+The c64/c128 checkers disagree with Gambit on a bare real flonum -- Kaappi
```
baijum, line 98: Confirmed empirically on this branch: `storage --classes c64 --count 3` → 3/3 mismatch, each first difference a `check` line on a bare real flonum (kaappi `#t`, gambit `#f`). Two non-blocking thoughts, plus a documentation tail:

- A default-draw run now reports mismatches and exits 1 essentially every time (c64 and c128 are 2 of the 16 classes in the draw), making red the steady state. The `specialized-array-default-safe?` divergence is suppressed by *pinning* it in the storage prelude; the same philosophy could apply here — drop c64/c128 from the default draw (opt back in with `--classes`), or post-filter exactly this shape (`check <canonical real>` → `#t` vs `#f` on c64/c128 only) so every other probe — exact reals, complexes, setters, round-trips — stays compared. Documentation-only is defensible; this is a question, not a request.
- The chibi paragraph below still lists "c64 accepts a real flonum" among chibi's checker bugs, "all contradicted by Gambit," which now reads against the sanctioned verdict three paragraphs up. The distinction is real and worth half a sentence: I checked — chibi's `(imag-part 1.0)` is an exact `0` (Gambit's tower) and its checker still returns `#t`, so chibi diverges from the sample implementation's *text*, whereas Kaappi runs that text under a different tower.
- `docs/dev/testing.md`'s "known oracle divergences" summary (~line 411) enumerates this docstring's divergences and now misses the new permanent one — a matching clause there would keep the two documents in sync.

docs/dev/srfi-implementation-notes.md
```diff
@@ -936,17 +936,32 @@ explicit `(not (real? x))`, encoding the sample implementation's verdict under
...
-procedures "manipulate complex numbers with, respectively, 32- and 64-bit
-floating-point numbers as real and imaginary parts" — and under Kaappi's
-representation `1.0` is one: `(complex? 1.0)` is `#t` with `real-part` `1.0`
-and `imag-part` `0.0`, both inexact. The one normative constraint on a
-checker is that `(checker (getter v i))` be `#t`, which holds either way
-(the getter returns `1.0+0.0i`). So rejecting a bare real matches Gambit's
-verdict rather than the document's requirement, and it is a portability
-choice — code written against Kaappi will not silently break on Gambit —
-not conformance. Raised with the SRFI's author; revisit when he answers.
+**That clause was reverted in #2559.** The two checkers are the sample
```
baijum, line 939: Nit: the paragraph just above still ends in the present tense — "so the shipped checker carries an explicit `(not (real? x))`" — which this bolded retraction then has to undo for the reader. Making that sentence past ("so #2543 shipped the checker with an explicit `(not (real? x))` clause …") lets the paragraph stand alone as history instead of asserting something false of the current code.

baijum:
All six addressed. Thanks for re-deriving the claims from primary sources rather than the description — the Gambit checker line and both R7RS quotes being independently confirmed is worth more than my assertion of them.

**Named assertions** and the **"proper complexes" header** — both fixed. The header also still said "the reference"; that site lives in `tests/`, which my terminology pass in #2558 never reached.

**The `1.5` → `0.1` suggestion: taken.** You're right that `1.5` is f32-exact and so pinned acceptance and type but not narrowing. It's now `0.1` → `0.10000000149011612+0.0i`, which proves the element really went through the interleaved f32 body, with the c128 case beside it as the control — same input, f64 body, `0.1` back unchanged. One assertion pinning all three beats two pinning two.

**Tense nit:** fixed. The paragraph now reads as history instead of asserting something false of the current code.

**The chibi paragraph:** you're right that it read against the sanctioned verdict, and your check is the distinction — chibi's `(imag-part 1.0)` is an exact `0`, so the sample implementation's checker text would *reject* the value there. chibi diverges from the text; we run the same text under a different tower. Said in the docstring now.

**The default-draw question — I went with narrowing the draw, after trying the post-filter first and abandoning it.** Worth recording why, since the filter is the more appealing option on paper:

The divergence **cascades**. Once the checker accepts, the setter runs and the `contents` line diverges too. And infinities canonicalize as `(inf 1)`, not a bare rational. So a filter tight enough to be safe (`check <bare-rational> #t` vs `#f`) suppressed only 1 of 3 real instances, and one loose enough to catch them all would have to model the entire downstream consequence of the acceptance — which is exactly the shape that quietly swallows a real c64 setter bug two years from now.

Narrowing the default draw filters nothing and cannot mask anything: `--classes c64` still compares every probe, noise included, which is what you want when you're actually investigating c64. Verified: default draw 8/8 clean, `--classes c64` still 2/2 reporting.

**`docs/dev/testing.md`** now carries the matching clause, so the two documents stay in sync.

Re-verified: eight SRFI 231 suites pass, storage-classes **254** expected passes, official **10936 passed / 0 unexpected failures**, markdownlint clean.

PR #2561, opened 2026-09-09
Localise riscv64 unit-chunk wedges with a script-side watchdog

### LINKED ISSUE #2560: CI: riscv64 'rest' chunk exceeds its 24m cap in ~half of recent runs, and the cap localises nothing
The `rest` chunk of `riscv64-test` has hit its 24-minute `timeout-minutes` cap in **2 of 4 runs** over the last two days, on trees whose only changes were comments, tests and docs. Each time the other seven chunks passed and no test reported a failure — the step was killed.

This is not #2488 recurring in the form that issue asks for ("reopen with the test name"), and that is precisely the problem: **the cap fired without localising anything.**

## Observations

Same chunk, same 46 filters, same 911 tests:

| Run | `rest` chunk | Outcome |
|-----|-------------|---------|
| [34217768666](https://github.com/kaappi/kaappi/actions/runs/34217768666) (#2558) | 10m44s | pass |
| [34222560966](https://github.com/kaappi/kaappi/actions/runs/34222560966) (#2558) | **>24m** | killed at cap |
| 34222560966 rerun | 9m46s | pass |
| [34299484435](https://github.com/kaappi/kaappi/actions/runs/34299484435) (#2559) | **>24m** | killed at cap |

Healthy is ~10 minutes. The failures are not a slow creep toward the cap — they are bimodal, roughly 10m or >24m.

## Why the cap localises nothing

`.github/workflows/ci.yml` documents the intended relationship:

> "For that to happen the step must outlive the per-test bound, so every step cap below is its healthy time plus the 8m bound plus margin — **the cap is now the backstop, not the localiser**."

For `rest` that is 10m healthy + 8m bound + 6m margin = 24m, which is correct arithmetic for **one** hung test. What we observe is the cap acting as the localiser and localising nothing: the step printed its banner and then nothing at all for 24 minutes.

`tools/run-unit-test-chunk.sh` prints exactly two lines — the banner (:193) and the post-run summary (:209) — so silence in between is by design and is **not** itself evidence of a hang.

## Two readings, and I cannot separate them from the logs

1. **No single test hung.** If `--test-timeout 8m` had fired it would have named the test and carried on (that is what #2491 bought us), and `zig build ... | tee` should stream that line. No such line appeared, which points at the whole chunk running ~2.4x slower rather than one test wedging. If so the cap's arithmetic is wrong in a different way: it budgets for one hung test but not for variance in the *healthy* time itself.
2. **Two or more tests hung.** 10m + 8m + 8m = 26m > 24m, which would also produce a silent kill with no name surviving.

I have not verified that a `--test-timeout` message actually streams through `tee` under QEMU rather than being buffered until exit; if it buffers, reading 1 is unsupported and every hang will always be silent.

## Suggested direction

Whichever reading holds, the current setup cannot tell us which. Options, roughly in order of how much they buy:

- **Emit per-filter progress from the chunk script.** 46 filters, one line each, and the last line before a cap kill names the filter in flight. This makes every future occurrence reopenable against #2488 with a name, and costs 46 lines of log.
- **Confirm whether `--test-timeout` output streams** under `| tee` in this environment. If it buffers, that alone defeats #2491's localisation on every capped step, not just this one.
- **Re-derive the cap from observed spread** rather than from healthy + one bound, given healthy itself varies.

Raising the cap alone would only move the silence later.

## Impact

Not a correctness problem — the tests pass, and a rerun clears it. But it fails roughly half of `riscv64-test` runs right now, so it blocks PRs and costs a rerun each time, and the documented "cap is the backstop, not the localiser" guarantee does not hold in practice.

Related: #2488 (closed not-planned; reopen criterion is a test name), #2490 (chunking), #2491 (`--test-timeout`).

baijum:
Fixes #2560.

## The finding first

The issue's "one to answer first" question is answered, and the answer changes the fix: **`zig build` buffers its entire console output until the process exits.** Verified on Zig 0.16.0 with a synthetic hanging test: piped through `tee`, redirected to a file, or run under a pty, a test that overruns `--test-timeout` is killed and named the moment the bound fires — yet zero bytes leave `zig build` until it exits, and SIGTERM does not flush the buffer either.

So a step killed at its `timeout-minutes` cap takes the buffer — names included — with it. That is exactly the silent `>24m` kill of #2560, and it means #2491's naming was defeated on *every* capped step, not just this chunk. It also means the capped runs tell us nothing about which reading of the issue held: the name may have existed inside the buffer and died there.

## What this does

Since zig's own output cannot be made to survive, `tools/run-unit-test-chunk.sh` now carries the localisation itself, in its own unbuffered `echo` lines (which survive any kill):

- **Heartbeat** every 60s while a run is in flight, noting that zig's silence in between is normal.
- **Wall budget** (`KAAPPI_CHUNK_BUDGET`, seconds, per chunk step in ci.yml), sized to outlive one per-test bound (healthy + 8m + ~2m) so a *single* hung test is still named by zig at its own exit — only a second hang, or a wedge the runner cannot see, falls through. On expiry the script kills the zig process tree (children first, TERM then grace then KILL — the build runner orphans its test binary otherwise) and bisects.
- **Bisection** over the chunk's filter list under a tight 90s per-test bound, one streamed line per level naming the subset in flight *before* it runs. Each level is capped at 3/5 of the remaining pot so the descent degrades geometrically instead of being consumed by the first wedged level. Three honest outcomes:
  - a process-level wedge wedges every level containing it → descent ends on **one filter, confirmed by a solo run** (validated end-to-end locally with a real hanging test: `LOCALISED AND CONFIRMED: filter 'tests_fuzz.test.' …` plus a ready-to-paste reproduce command);
  - per-test hangs/slow tests never wedge a level (the tight bound kills them mid-level) → every name the bound flushes is echoed as a **NOTE line**, and if no level wedged, the whole chunk is re-run **cache-hot** under the tight bound to name every per-test culprit in full context — this covers the issue's "whole chunk 2.4x slower" reading;
  - nothing reproduces → the log says *that*, instead of guessing.
- The step exits **124** (the `timeout(1)` convention) when the budget fired, so the run fails loudly rather than just bleeding out at a CI cap.

## CI arithmetic (the issue's third ask)

Each riscv64 chunk step now sets its `KAAPPI_CHUNK_BUDGET` from its slowest observed healthy time (the #2560 spread for `rest`: 9m46s–11m32s) + the 8m bound + ~2m margin, and its `timeout-minutes` is re-derived as budget + an 18m bisection pot + 2m slop, with the arithmetic spelled out in the workflow. The job cap moves 55 → 80 to let a wedge in *any* chunk localise fully, and `rest` now runs **first**: it is the chunk that has wedged (both #2488 recurrences and both #2560 kills), so a wedge fails fast with the most headroom instead of burning ~36m of healthy chunks first. Healthy runs are unchanged — no budget is exceeded, one compile per chunk, same output and exit statuses.

## Test

New `tests/scheme/tools/` shell suite (wired into `run-all.sh`) drives the machinery with budgets small enough that the compile itself "wedges": asserts the streamed budget-exceeded line, a surviving heartbeat, a bisection level naming its subset, the level's decision line, the dry-pot verdict, and exit 124 — on both cold-cache and warm-cache outcomes of the level. `skip_on_windows`/`skip_without_zig` per the shell-common conventions.

## Not touched

s390x/ppc64le still run the whole suite in one `--test-timeout 8m` step and share the buffering vulnerability; they have not wedged yet, and converting them is out of scope here.

### DIFF
--- a/tests/scheme/tools/unit-chunk-watchdog-2560.sh
+++ b/tests/scheme/tools/unit-chunk-watchdog-2560.sh
@@ -215,9 +215,12 @@ grep -q "WALL BUDGET EXCEEDED after .*s (compiling)" "$out" ||
 # and depth-2 lines must carry estimates computed from each subset's own
 # declared-test count (the test derives them independently, from the same
 # grep of the same files), so a regression back to a flat or per-filter
-# estimate changes the printed numbers and fails here. Depth 1 completing
-# (the wedge-once stamp was consumed in phase 1) puts depth 2 at filters
-# 24-34, the two windows differing in density.
+# estimate changes the printed numbers and fails here. The two probe
+# windows are derived from --list with the same mid arithmetic the script
+# uses, so a new test file landing in the derived rest chunk moves them
+# with it instead of failing the case: depth 1 is the first half, and
+# depth 1 completing (the wedge-once stamp was consumed in phase 1) puts
+# depth 2 just past it.
 est_of() { # est_of <start-1-based> <count>: 2s base + 1 tenth x the slice's tests
     local start="$1" n="$2" total=0 f c
     for f in $(bash tools/run-unit-test-chunk.sh --list rest \
@@ -228,8 +231,13 @@ est_of() { # est_of <start-1-based> <count>: 2s base + 1 tenth x the slice's tes
     done
     echo $(( 2 + total / 10 ))
 }
-e1=$(est_of 1 23)
-e2=$(est_of 24 11)
+total=$(bash tools/run-unit-test-chunk.sh --list rest | wc -l | tr -d ' ')
+mid1=$((total / 2))             # depth 1: filters 1..mid1
+mid2=$(((mid1 + total) / 2))    # depth 2 after depth 1 completes: mid1+1..mid2
+n1=$mid1
+n2=$((mid2 - mid1))
+e1=$(est_of 1 "$n1")
+e2=$(est_of $((mid1 + 1)) "$n2")
 [ "$e1" -ne "$e2" ] || fail "case 7 setup: the two probe windows estimate equal; pick denser windows"
 rc=0
 # shellcheck disable=SC2086  # base_env is a deliberate VAR=val word list for env
@@ -238,9 +246,9 @@ env $base_env KAAPPI_CHUNK_BUDGET=3 KAAPPI_BISECT_BUDGET=45 \
     SHIM_WEDGE_ONCE="$work/once7" \
     bash tools/run-unit-test-chunk.sh rest > "$out" 2>&1 || rc=$?
 [ "$rc" -eq 124 ] || fail "case 7 (per-test funding): wall-budget overrun must exit 124 (got $rc)"
-grep -q "bisect depth 1: 23 filter(s) .* budget .*s of a ${e1}s estimate" "$out" ||
+grep -q "bisect depth 1: ${n1} filter(s) .* budget .*s of a ${e1}s estimate" "$out" ||
     fail "case 7: depth 1's estimate is not ${e1}s from its subset's test count"
-grep -q "bisect depth 2: 11 filter(s) .* budget .*s of a ${e2}s estimate" "$out" ||
+grep -q "bisect depth 2: ${n2} filter(s) .* budget .*s of a ${e2}s estimate" "$out" ||
     fail "case 7: depth 2's estimate is not ${e2}s from its subset's test count"
 
 echo "PASS: watchdog, heartbeat, decisive descent, per-test NOTEs, the"
--- a/tests/scheme/tools/unit-chunk-watchdog-2560.sh
+++ b/tests/scheme/tools/unit-chunk-watchdog-2560.sh
@@ -1,0 +1,78 @@
+#!/usr/bin/env bash
+# The chunk script's wall-budget watchdog and streamed bisection (kaappi#2560).
+#
+# `zig build` buffers all of its console output until the process exits
+# (verified on Zig 0.16.0: through a pipe, a file, or a pty; SIGTERM does
+# not flush it either), so a CI step killed at its `timeout-minutes` cap
+# loses even the per-test timeout names that #2491's `--test-timeout` had
+# already recorded. Every capped riscv64 `rest` run therefore died
+# silent -- the cap localised nothing, which is what #2560 filed.
+#
+# tools/run-unit-test-chunk.sh now localises with its OWN unbuffered echo
+# lines, which survive any kill: a heartbeat while the chunk runs, a
+# "WALL BUDGET EXCEEDED" line when the script's wall budget fires, and one
+# line per bisection level naming the filter subset in flight, exiting 124
+# (the timeout(1) convention) so the run fails loudly instead of just
+# quietly bleeding out at a CI cap.
+#
+# Both cases below drive that machinery with budgets small enough that the
+# initial test-binary compile itself "wedges": no real test hangs, and each
+# case is bounded by its budgets (~40s and ~75s). Which side of a bisection
+# level's own budget the machine lands on (cold compile vs warm objects)
+# varies by host, so every assertion accepts exactly the two outcomes the
+# script is allowed to print for it.
+
+. "$(dirname "$0")/../shell-common.sh"
+
+skip_on_windows "the watchdog kills the zig process tree via pgrep, which MSYS lacks"
+skip_without_zig "the chunk script runs zig build"
+
+cd "$(dirname "$0")/../../.." || exit 1
+
+fail() {
+    echo "FAIL: $1" >&2
+    exit 1
+}
+
+# Structure: --list still derives the rest chunk's filters from the tree,
+# and an unknown chunk name is a usage error, not a silent success.
+n=$(bash tools/run-unit-test-chunk.sh --list rest | wc -l | tr -d ' ')
+[ "$n" -ge 20 ] || fail "--list rest produced only $n filters (expected the derived remainder, dozens)"
+if bash tools/run-unit-test-chunk.sh --list bogus-chunk > /dev/null 2>&1; then
+    fail "an unknown chunk name must exit 2"
+fi
+
+# Case 1: the wall budget fires and the bisection pot is too dry for even
+# one level. The kill must still be loud, streamed, and exit 124.
+out=$(mktemp)
+trap 'rm -f "$out" "$out2"' EXIT
+KAAPPI_CHUNK_BUDGET=12 KAAPPI_BISECT_BUDGET=20 KAAPPI_HEARTBEAT_SECS=3 \
+    bash tools/run-unit-test-chunk.sh rest > "$out" 2>&1
+status=$?
+
+[ "$status" -eq 124 ] || fail "case 1: wall-budget overrun must exit 124 (timeout(1) convention), got $status"
+grep -q "WALL BUDGET EXCEEDED" "$out" || fail "case 1: no streamed budget-exceeded line"
+grep -q "still running at" "$out" || fail "case 1: no heartbeat line survived the kill"
+grep -q "too shallow for depth 1" "$out" || fail "case 1: the dry pot was not reported as too shallow"
+grep -q "RESULT:" "$out" || fail "case 1: no summary verdict line"
+
+# Case 2: enough pot for one bisection level (a level gets 3/5 of the pot
+# and needs >= 90s, so 180s is the smallest pot that starts one). A cold
+# cache cannot compile the subset in ~108s (level watchdog fires, descent
+# enters that half); a warm one runs the subset's tests (level finishes,
+# half is discarded). Both are correct; both must name the subset BEFORE
+# running it, and the level's outcome must be reported as a decision.
+out2=$(mktemp)
+KAAPPI_CHUNK_BUDGET=12 KAAPPI_BISECT_BUDGET=180 KAAPPI_HEARTBEAT_SECS=3 \
+    bash tools/run-unit-test-chunk.sh rest > "$out2" 2>&1
+status=$?
+
+[ "$status" -eq 124 ] || fail "case 2: wall-budget overrun must exit 124, got $status"
+grep -Eq "bisect depth 1: [0-9]+ filter\\(s\\) [a-z_0-9.]+\.\.[a-z_0-9.]+, budget [0-9]+s" "$out2" ||
+    fail "case 2: no bisect level line naming its filter subset and its budget"
+grep -Eq "(did NOT finish in [0-9]+s -- the wedge reproduces inside this half|finished in [0-9]+s \\(zig exit [0-9]+\\) -- no process wedge here)" "$out2" ||
+    fail "case 2: the level's outcome was not reported as a bisection decision"
+grep -Eq "(bisect incomplete|RESULT:)" "$out2" ||
+    fail "case 2: neither an incomplete-descent suspects list nor a verdict line was printed"
+
+echo "PASS: watchdog fired loudly, bisection streamed its subsets, exit 124"
--- a/tests/scheme/tools/unit-chunk-watchdog-2560.sh
+++ b/tests/scheme/tools/unit-chunk-watchdog-2560.sh
@@ -170,8 +170,14 @@ grep -q "NOTE: 'tests_fake.test.slow one' exceeded the tight 90s bound" "$out" |
 # the descent.
 echo 2 > "$work/count"
 rc=0
+# The pot must clear the level floor (BASE/3 = 10s here) by more than a
+# second: pot_start and depth 1's now-stamp straddle a clock-tick boundary
+# whenever a second flips between them, and a pot equal to the floor then
+# reads as one second short and the level is refused as "too shallow"
+# instead of run clipped -- a low-single-digit-percent flake per run.
+# 12 leaves that tick of slack while still clipping 12 < 30.
 # shellcheck disable=SC2086  # base_env is a deliberate VAR=val word list for env
-env $base_env KAAPPI_CHUNK_BUDGET=3 KAAPPI_BISECT_BUDGET=10 \
+env $base_env KAAPPI_CHUNK_BUDGET=3 KAAPPI_BISECT_BUDGET=12 \
     KAAPPI_BISECT_LEVEL_BASE_SECS=30 \
     SHIM_HANG_FIRST="$work/count" \
     bash tools/run-unit-test-chunk.sh rest > "$out" 2>&1 || rc=$?
@@ -205,5 +211,38 @@ env $base_env KAAPPI_CHUNK_BUDGET=3 KAAPPI_BISECT_BUDGET=0 \
 grep -q "WALL BUDGET EXCEEDED after .*s (compiling)" "$out" ||
     fail "case 6: a zig-named grandchild must be reported as compiling"
 
+# Case 7: the per-TEST funding term, with the term nonzero. The depth-1
+# and depth-2 lines must carry estimates computed from each subset's own
+# declared-test count (the test derives them independently, from the same
+# grep of the same files), so a regression back to a flat or per-filter
+# estimate changes the printed numbers and fails here. Depth 1 completing
+# (the wedge-once stamp was consumed in phase 1) puts depth 2 at filters
+# 24-34, the two windows differing in density.
+est_of() { # est_of <start-1-based> <count>: 2s base + 1 tenth x the slice's tests
+    local start="$1" n="$2" total=0 f c
+    for f in $(bash tools/run-unit-test-chunk.sh --list rest \
+               | sed -n "${start},$((start + n - 1))p" \
+               | sed 's/^-Dtest-filter=//; s/\.test\.$//'); do
+        c=$(grep -c '^test "' "src/$f.zig" | tr -d ' ')
+        total=$((total + c))
+    done
+    echo $(( 2 + total / 10 ))
+}
+e1=$(est_of 1 23)
+e2=$(est_of 24 11)
+[ "$e1" -ne "$e2" ] || fail "case 7 setup: the two probe windows estimate equal; pick denser windows"
+rc=0
+# shellcheck disable=SC2086  # base_env is a deliberate VAR=val word list for env
+env $base_env KAAPPI_CHUNK_BUDGET=3 KAAPPI_BISECT_BUDGET=45 \
+    KAAPPI_BISECT_LEVEL_PER_TEST_TENTHS=1 \
+    SHIM_WEDGE_ONCE="$work/once7" \
+    bash tools/run-unit-test-chunk.sh rest > "$out" 2>&1 || rc=$?
+[ "$rc" -eq 124 ] || fail "case 7 (per-test funding): wall-budget overrun must exit 124 (got $rc)"
+grep -q "bisect depth 1: 23 filter(s) .* budget .*s of a ${e1}s estimate" "$out" ||
+    fail "case 7: depth 1's estimate is not ${e1}s from its subset's test count"
+grep -q "bisect depth 2: 11 filter(s) .* budget .*s of a ${e2}s estimate" "$out" ||
+    fail "case 7: depth 2's estimate is not ${e2}s from its subset's test count"
+
 echo "PASS: watchdog, heartbeat, decisive descent, per-test NOTEs, the"
-echo "PASS: clipped-level INCONCLUSIVE stop, and both kill-time phase reads"
+echo "PASS: clipped-level INCONCLUSIVE stop, both kill-time phase reads,"
+echo "PASS: and level estimates that follow the subset's own test count"
--- a/tests/scheme/tools/unit-chunk-watchdog-2560.sh
+++ b/tests/scheme/tools/unit-chunk-watchdog-2560.sh
@@ -15,17 +15,19 @@
 # (the timeout(1) convention) so the run fails loudly instead of just
 # quietly bleeding out at a CI cap.
 #
-# Both cases below drive that machinery with budgets small enough that the
-# initial test-binary compile itself "wedges": no real test hangs, and each
-# case is bounded by its budgets (~40s and ~75s). Which side of a bisection
-# level's own budget the machine lands on (cold compile vs warm objects)
-# varies by host, so every assertion accepts exactly the two outcomes the
-# script is allowed to print for it.
+# Every case below drives the script through a fake `zig` (KAAPPI_ZIG): a
+# shim that plays wedged or clean subsets in seconds, with real child
+# processes so the watchdog's process-tree snapshot has a tree to kill.
+# That makes each outcome deterministic -- the descent, the solo confirm,
+# the NOTE lines for tight-bound overruns, the INCONCLUSIVE stop for a
+# level the pot could only clip, and the unbudgeted heartbeat (which died
+# on an unbound variable under bash >= 4 before the fix) -- without ever
+# compiling the unit suite or contending for the Zig cache locks. The
+# real-zig path is what the riscv64 chunk steps run eight times per CI job.
 
 . "$(dirname "$0")/../shell-common.sh"
 
 skip_on_windows "the watchdog kills the zig process tree via pgrep, which MSYS lacks"
-skip_without_zig "the chunk script runs zig build"
 
 cd "$(dirname "$0")/../../.." || exit 1
 
@@ -33,6 +35,61 @@ fail() {
     echo "FAIL: $1" >&2
     exit 1
 }
+
+work=$(mktemp -d "${TMPDIR:-/tmp}/kaappi-chunk-watchdog-XXXXXX")
+trap 'rm -rf "$work"' EXIT
+out="$work/out"
+
+# The fake `zig build test`: never compiles, decides from its arguments.
+#   wedge (hang with two children) iff, in order:
+#     - SHIM_WEDGE_ONCE is set and its stamp file is missing (first call), or
+#     - SHIM_HANG_FIRST is set and fewer than that many calls have hung, or
+#     - SHIM_WEDGE_FILTER is set and some -Dtest-filter= arg contains it;
+#   otherwise sleep SHIM_DELAY, print SHIM_TIMEOUT_NAME as a zig per-test
+#   timeout line (exercising the script's sed parser, spaces included),
+#   print a summary the script's floor check accepts, exit SHIM_EXIT.
+shim="$work/zig"
+cat > "$shim" <<'SHIM'
+#!/usr/bin/env bash
+set -u
+if [ -n "${SHIM_WEDGE_ONCE:-}" ] && [ ! -f "${SHIM_WEDGE_ONCE}" ]; then
+    : > "${SHIM_WEDGE_ONCE}"
+    sleep 9999 & sleep 9999 & wait
+fi
+if [ -n "${SHIM_HANG_FIRST:-}" ]; then
+    count=0
+    [ -f "${SHIM_HANG_FIRST}" ] && count=$(cat "${SHIM_HANG_FIRST}")
+    if [ "$count" -gt 0 ]; then
+        echo $((count - 1)) > "${SHIM_HANG_FIRST}"
+        sleep 9999 & sleep 9999 & wait
+    fi
+fi
+if [ -n "${SHIM_WEDGE_FILTER:-}" ]; then
+    for a in "$@"; do
+        case "$a" in
+            -Dtest-filter=*"$SHIM_WEDGE_FILTER"*) sleep 9999 & sleep 9999 & wait ;;
+        esac
+    done
+fi
+[ -n "${SHIM_DELAY:-}" ] && sleep "$SHIM_DELAY"
+if [ -n "${SHIM_TIMEOUT_NAME:-}" ]; then
+    printf "error: '%s' timed out after 90s\n" "$SHIM_TIMEOUT_NAME"
+fi
+printf '%s\n' \
+    "test success" \
+    "+- run test unit-tests 60 pass (60 total) 1s MaxRSS:1M" \
+    "+- run test thottam-tests 1 pass (1 total) 1s MaxRSS:1M"
+exit "${SHIM_EXIT:-0}"
+SHIM
+chmod +x "$shim"
+
+# Common knobs: the pinned KAAPPI_TEST_TIMEOUT guards against a caller's
+# run-all.sh KAAPPI_TEST_TIMEOUT (seconds, for Scheme tests) leaking into
+# the --test-timeout grammar (a Zig duration). The level-funding constants
+# are scaled down so a level's "full estimate" is seconds, not minutes.
+base_env="KAAPPI_ZIG=$shim KAAPPI_TEST_TIMEOUT=8m KAAPPI_BISECT_TEST_TIMEOUT=90s"
+base_env="$base_env KAAPPI_BISECT_LEVEL_BASE_SECS=2 KAAPPI_BISECT_LEVEL_PER_FILTER_SECS=1"
+base_env="$base_env KAAPPI_HEARTBEAT_SECS=2"
 
 # Structure: --list still derives the rest chunk's filters from the tree,
 # and an unknown chunk name is a usage error, not a silent success.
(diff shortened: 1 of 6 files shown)


tools/run-unit-test-chunk.sh
```diff
@@ -189,10 +264,230 @@ log="$(mktemp)"
...
+# Runs `zig build test <args...> --summary all`, teeing its output to the
+# live log as it appears (in practice once, at exit -- see the header) and
+# to "$log", while printing a heartbeat every $heartbeat_secs and enforcing
+# the wall budget. The zig pid is tracked directly (output goes to a file
+# the script forwards, not through a pipeline, so `$!` is zig and killing it
+# orphans nothing but its own tree). Sets on return: zig_status,
+# watchdog_fired (1 = budget expired and the tree was killed), ran_secs.
+run_supervised() {
+    local budget="$1" label="$2"; shift 2
+    local start now next_beat sent size grace budget_note
```
baijum: **Bug: the unbudgeted run dies at its first heartbeat under bash ≥ 4.**

`local budget_note` declares the variable without assigning it, and the `if [ "$budget" -gt 0 ]` below only assigns it when a budget is set. With `set -u`, the heartbeat's `${budget_note}` at line 320 is then an unbound-variable error. macOS's `/bin/bash` 3.2 tolerates that (`local x` leaves `x` set-but-empty there), which is presumably why it did not show up locally; bash 5 on Linux and Homebrew does not.

Reproduced on this branch with `KAAPPI_HEARTBEAT_SECS=3 bash tools/run-unit-test-chunk.sh native` under bash 5.3:

```
== unit-test chunk 'native': 5 filter(s), per-test timeout 8m, no wall budget (KAAPPI_CHUNK_BUDGET unset), extra args:
tools/run-unit-test-chunk.sh: line 320: budget_note: unbound variable
script exit=1
```

and `pgrep -fl 'zig build test'` afterwards still showed the zig build running: the script exits 1 and orphans the whole tree it was supposed to be supervising. That is the header's primary documented invocation (`bash tools/run-unit-test-chunk.sh <chunk>`, no budget), and any chunk longer than the 60 s default heartbeat hits it. CI is unaffected only because every step sets `KAAPPI_CHUNK_BUDGET`.

Fix is `local budget_note=""` (the bisection levels and the full re-run always pass a budget > 0, so this is the only path). The new test cannot catch this today because both of its cases set a budget; see my note on the test file for a way to make an unbudgeted heartbeat cheap to assert.

baijum (commented):
Reviewed the mechanism, the script, the CI arithmetic, and the test.

**What checks out.** The buffering finding matches Zig's source: step errors accumulate in `result_error_msgs` and are rendered by `printErrorMessages` only when `zig build` reports, and each is written as `error:` at column 0 followed by the message, so the script's `^error: '\(…\)' timed out after` parser matches the real rendering. Every step cap in ci.yml is budget + 18 m + 2 m as stated, each budget is slowest-healthy + 8 m + ~2 m, and 8 m setup + ~38 m healthy chunks + the 32 m cap of the last chunk fits the 80 m job cap. Moving `rest` first is right. `kill_tree` walking children first is right too (TERM on `zig build` alone does orphan the runner).

**One confirmed bug that should block merge** (inline at line 300): with no `KAAPPI_CHUNK_BUDGET` set — the header's primary usage — the script dies at its first heartbeat on bash ≥ 4 with `budget_note: unbound variable`, exit 1, and leaves the zig tree running unsupervised. Reproduced on this branch; one-line fix.

**Two design points** (inline at lines 376 and 419) that I would like at least acknowledged in the log wording: a bisection level that got less than its own estimate is still treated as a wedge verdict, and on riscv64 with the 1080 s pot that is every level after the first, so the descent will typically report narrowed suspects that are not actually narrowed. Related, the script cannot distinguish "still compiling" from "tests wedged", and several verdict lines are worded as if it could.

**Test** (inline): a fake-`zig` shim would make the new suite deterministic, cheap, and able to assert the paths it currently has to accept either way — including the unbudgeted heartbeat that would have caught the bug above.

s390x/ppc64le being left alone is fine for this PR.

tools/run-unit-test-chunk.sh
```diff
@@ -189,10 +264,230 @@ log="$(mktemp)"
...
+    echo "== bisect: ${#filters[@]} filter(s), per-test timeout $bisect_timeout, total budget ${bisect_pot}s"
+    while [ $((hi - lo)) -gt 1 ]; do
+        depth=$((depth + 1))
+        now=$(date +%s)
+        remaining=$((bisect_pot - (now - pot_start)))
+        mid=$(((lo + hi) / 2))
+        n=$((mid - lo))
+        allow=$((BISECT_LEVEL_BASE_SECS + BISECT_LEVEL_PER_FILTER_SECS * n))
+        pot_share=$((remaining * BISECT_LEVEL_POT_SHARE_NUMERATOR / BISECT_LEVEL_POT_SHARE_DENOMINATOR))
+        [ "$allow" -gt "$pot_share" ] && allow=$pot_share
```
baijum: **An under-funded level is still treated as evidence, and on riscv64 that is every level after the first.**

`allow` is clipped to 3/5 of the remaining pot, but a timeout under a clipped allowance is read exactly like one under the full estimate ("the wedge reproduces inside this half", `hi=$mid`). The only guard is the 90 s absolute floor, which the comment above itself describes as "one cold compile, no tests", i.e. an allowance under which a clean level cannot finish either.

The arithmetic for the leg this is built for (`rest`, 46 filters, pot 1080 s, ~1.5 m compile + ~12 s/filter healthy under QEMU per the #2560 numbers):

| depth | filters | estimate | pot share → allow | clean level needs |
|---|---|---|---|---|
| 1 | 23 | 1105 s | 648 s | ~370 s ✓ |
| 2 | 12 | 720 s | ~250 s (after a wedged depth 1) | ~240 s, at the line |
| 3 | 6 | 510 s | ~100 s | ~165 s ✗ misread as wedged |
| 4 | — | — | < 90 s | stop |

So after one reliable level the descent follows the first half regardless of what is in it, and reports three "suspect" filters that may not contain the wedge at all. Chunk times also vary ~2× with runner load (#2491's first run), which pushes depth 2 over the line too. The `LOCALISED AND CONFIRMED` path the PR validated is reachable natively, where compile and run are fast; on riscv64 the realistic outcome is `bisect incomplete` with suspects that the log presents as narrowed but which are not.

Suggestions, either of which keeps the output honest:
- refuse to run a level whose `allow` is below its own estimate (`BASE + PER_FILTER * n`), or run it but print and record it as **inconclusive** so it never moves `hi`; the descent then ends with "incomplete, pot too shallow at depth N" instead of a wrong subset;
- or drop the 3/5 share and let the pot fund one or two *full* levels, since a clipped level is worth less than no level.

A different lever worth considering for the process-wedge case: the test binary phase 1 compiled is still in `.zig-cache/o/*/unit-tests`, and run directly (no `--listen`) Zig's test runner writes a `N/M <name>...` line to stderr before each test, so the last line names the wedged test in one run with no bisection at all (that is how the #2488 local repro recipe worked). Not asking for that here, just noting the pot buys a lot more that way.

tools/run-unit-test-chunk.sh
```diff
@@ -189,10 +264,230 @@ log="$(mktemp)"
...
+    now=$(date +%s)
+    remaining=$((bisect_pot - (now - pot_start)))
+
+    if [ "$wedged_seen" = 0 ]; then
+        # No level wedged, so the descent proves nothing about filters: the
+        # phase-1 overrun was per-test (hangs or slowness the runner can
+        # bound) or it does not reproduce at all. The tight-bound full
+        # re-run names the former; the log's honesty covers the latter.
+        if [ "$remaining" -ge "$BISECT_LEVEL_MIN_SECS" ]; then
+            echo "== bisect: no level wedged -- re-running the whole chunk under the tight $bisect_timeout bound (compile is cache-hot) to name per-test culprits in full context, budget ${remaining}s"
```
baijum: **"compile is cache-hot" holds only when the wall budget fired *after* the compile finished.**

True for every CI budget (≥ 12 m against a ~1.5 m compile), so this is not a CI problem. But the new test's case 2 kills phase 1 at 12 s, mid-compile, so on its warm-cache path this re-run compiles the 46-filter binary cold, blows the remaining pot, and the script then prints `RESULT: the wedge reproduced only in the full chunk, never in any subset -- order- or interaction-dependent` for what was a compile. Same shape as the level comment above: the script cannot tell "still compiling" from "tests wedged", and its verdict lines are worded as if it could. Either hedge the two RESULT/decision lines ("did not finish", not "wedged"), or check for a runner child (`pgrep -P` on the build runner) at kill time and say which phase was in flight — that is one line and makes every verdict here trustworthy.

tests/scheme/tools/unit-chunk-watchdog-2560.sh
```diff
@@ -0,0 +1,78 @@
...
+# silent -- the cap localised nothing, which is what #2560 filed.
+#
+# tools/run-unit-test-chunk.sh now localises with its OWN unbuffered echo
+# lines, which survive any kill: a heartbeat while the chunk runs, a
+# "WALL BUDGET EXCEEDED" line when the script's wall budget fires, and one
+# line per bisection level naming the filter subset in flight, exiting 124
+# (the timeout(1) convention) so the run fails loudly instead of just
+# quietly bleeding out at a CI cap.
+#
+# Both cases below drive that machinery with budgets small enough that the
```
baijum: **Consider driving the script with a fake `zig` instead of the real unit-suite compile.**

Three reasons, in order of weight:

1. **The assertions accept both outcomes of every level**, so they cannot tell a working descent from a broken one: neither path exercises `hi=$mid` → confirm (`LOCALISED AND CONFIRMED`), the NOTE lines, or the `error: '…' timed out after` sed parser. And a case that runs *without* a budget would have caught the `budget_note` bug at line 300 of the script, but no such case is affordable when each case is a real compile.
2. **Cost per run-all.sh invocation.** Two real compiles of the unit suite (46-filter set, then a 23-filter set) on every CI `test` leg and every local run, only to be killed. On the Debug leg that is a cold ReleaseSafe compile (the kaappi#2431 situation shell-common documents), contending for the Zig cache locks with the compile suite's own builds running in parallel. Case 2's warm-cache path is ~12 + ~60 + 120 s + grace ≈ 3.5 min against the default 300 s `SHELL_TIMEOUT`, which is tight under load.
3. **Determinism.** Which path case 2 takes depends on the host's cache state, so a failure is not reproducible from the log alone.

A `KAAPPI_ZIG` (or plain `ZIG`) override in the script, defaulting to `zig`, lets this test point it at a shim on `PATH` that sleeps for a scripted number of seconds per filter set and prints canned `error: 'x.test.y' timed out after 90s` lines. Every path — confirm, NOTE, incomplete, full re-run, unbudgeted heartbeat — then runs in seconds with fixed outcomes, and `skip_without_zig` is no longer needed. The `pgrep`-based `kill_tree` still gets exercised because the shim is a real process tree (`sleep` children).

tools/run-unit-test-chunk.sh
baijum: Fixed in 01c9f85b — `local budget_note=""`, with a comment noting why the bare `local` was unset under `set -u` on bash ≥ 4 and set-but-empty on 3.2. Your repro also became the suite's case 1 (unbudgeted run, heartbeat, clean finish, floor-checked summary): I verified it fails on the unfixed script and passes on the fixed one. The real-zig unbudgeted path was re-run end-to-end (heartbeats, accounting, exit 0).

tools/run-unit-test-chunk.sh
baijum: Went with your first option, minus the refusal: the 3/5 share is gone entirely — the pot funds FULL estimates (`allow = min(estimate, remaining)`), and the estimate itself was re-derived to a clean level's plausible worst case under QEMU (120s + 25s/filter ≈ 2× the observed ~12s/filter + compile), so your table now reads: depth 1 fully funded and decisive (695s of the 1080s pot), depth 2 funded when depth 1 completed, and so on. A timeout under a clipped allowance is reported `INCONCLUSIVE` and never moves `hi`; a completion under a clipped allowance still exonerates, since finishing is finishing whatever the allowance was. On the pot-drained path the RESULT says exactly that and points at `KAAPPI_BISECT_BUDGET`. Both constants are env-overridable, which is also how the shim test scales them down to seconds.

Thanks for the table — it was the decisive argument. The direct-binary lever (running the phase-1 `unit-tests` from `.zig-cache/o` to watch the per-test stderr lines) is noted in the issue for a follow-up; agreed it buys more than any pot arithmetic.

tools/run-unit-test-chunk.sh
baijum: Both. `phase_of` snapshots the tree's shape at kill time — a grandchild that is the zig compiler itself means compiling, a grandchild that is anything else (the cached test binary, or the QEMU wrapper on the emulated legs) means running tests — and every kill/descent/confirm/re-run line now quotes it: `WALL BUDGET EXCEEDED after 12s (compiling)`, `... at its full estimated allowance (running tests at kill time)`. I verified the discrimination empirically before trusting it (zig's own compile workers are the runner's children mid-compile; the test binary is its child mid-run). The wording is hedged (`unknown` when pgrep is absent or the tree is gone), the 'cache-hot' claim is now conditional on the chunk run having got past its compile, and a full re-run killed mid-compile says the pot was too small to conclude anything instead of claiming order-dependence.

tests/scheme/tools/unit-chunk-watchdog-2560.sh
baijum: Done — `KAAPPI_ZIG` (default `zig`) drives every invocation, and the suite is now entirely shim-based: a fake `zig build test` that wedges or completes by filter-subset / first-N-invocations, with real `sleep` children so `tree_pids` has a tree to snapshot and kill. Four deterministic cases now assert exactly what previously accepted either outcome: the decisive descent ending `LOCALISED AND CONFIRMED` on the shim's wedge filter, the per-test NOTE lines (space-containing name whole, deduped across level + re-run), the clipped-level INCONCLUSIVE stop, and the unbudgeted heartbeat + floor-checked summary that catches the `budget_note` bug you found (verified it fails on the unfixed script). No unit-suite compiles, no cache-lock contention, ~2 minutes total, and `skip_without_zig` is gone.

Building the shim also caught a real script bug your review hinted at indirectly: the watchdog fired at the first poll where elapsed ≥ budget without re-checking liveness, so a run that finished during the 5s poll interval was killed as overran whenever the budget was under the interval — invisible at minute-scale CI budgets, immediate at shim scale. Liveness is now re-checked after the sleep.

baijum:
Review feedback is addressed in 01c9f85b — inline replies are on each comment; the short version:

**Blocking bug (unbudgeted run dying at its first heartbeat on bash ≥ 4, orphaning the zig tree):** fixed with `local budget_note=""` and pinned by a new unbudgeted test case, verified to fail on the unfixed script. Real-zig unbudgeted path re-run clean end-to-end.

**Under-funded levels treated as evidence:** the 3/5 pot share is gone. The pot funds full level estimates (re-derived to a clean level's plausible QEMU worst case, 120s + 25s/filter); a timeout under a clipped allowance is `INCONCLUSIVE` and never narrows the descent, while a completion still exonerates. On riscv64 that means ~1–2 decisive levels from the 18m pot, honestly labelled, instead of suspect lists that only looked narrowed.

**"Compile vs wedged" indistinguishability:** `phase_of` reads the process tree at kill time (grandchild = zig compiler ⇒ compiling; grandchild = test binary / QEMU wrapper ⇒ running tests — both shapes verified empirically), and every verdict line quotes the phase instead of asserting a wedge. The "cache-hot" claim on the full re-run is now conditional, and a re-run killed mid-compile says the pot was too small rather than claiming order-dependence.

**KILL escalation:** the tree is snapshotted before TERM (`tree_pids`, deepest first) and the KILL walks the snapshot, so descendants that survive TERM can't escape by getting re-parented when `zig build` exits.

**Test:** now entirely shim-driven via `KAAPPI_ZIG` — deterministic descent-to-confirm, space-safe deduped NOTE lines, INCONCLUSIVE stop, and the unbudgeted heartbeat, with real child processes for the tree-kill path; ~2 minutes, no compiles. `KAAPPI_TEST_TIMEOUT` is pinned in every invocation (the run-all.sh name collision), and run-all.sh gives the suite a 480s per-script budget plus slow-group dispatch.

One bonus fix the shim surfaced: the watchdog used to fire at the first poll where elapsed ≥ budget without re-checking liveness, so a run that finished inside the 5s poll interval was killed as overran whenever the budget was under the interval — unreachable at minute-scale CI budgets, but wrong, and now guarded.

baijum:
Review feedback is addressed in 01c9f85b — inline replies are on each comment; the short version:

**Blocking bug (unbudgeted run dying at its first heartbeat on bash >= 4, orphaning the zig tree):** fixed with `local budget_note=""` and pinned by a new unbudgeted test case, verified to fail on the unfixed script. Real-zig unbudgeted path re-run clean end-to-end.

**Under-funded levels treated as evidence:** the 3/5 pot share is gone. The pot funds full level estimates (re-derived to a clean level's plausible QEMU worst case, 120s + 25s/filter); a timeout under a clipped allowance is `INCONCLUSIVE` and never narrows the descent, while a completion still exonerates. On riscv64 that means ~1-2 decisive levels from the 18m pot, honestly labelled, instead of suspect lists that only looked narrowed.

**"Compile vs wedged" indistinguishability:** `phase_of` reads the process tree at kill time (grandchild = zig compiler means compiling; grandchild = test binary / QEMU wrapper means running tests — both shapes verified empirically), and every verdict line quotes the phase instead of asserting a wedge. The "cache-hot" claim on the full re-run is now conditional, and a re-run killed mid-compile says the pot was too small rather than claiming order-dependence.

**KILL escalation:** the tree is snapshotted before TERM (`tree_pids`, deepest first) and the KILL walks the snapshot, so descendants that survive TERM cannot escape by getting re-parented when `zig build` exits.

**Test:** now entirely shim-driven via `KAAPPI_ZIG` — deterministic descent-to-confirm, space-safe deduped NOTE lines, INCONCLUSIVE stop, and the unbudgeted heartbeat, with real child processes for the tree-kill path; ~2 minutes, no compiles. `KAAPPI_TEST_TIMEOUT` is pinned in every invocation (the run-all.sh name collision), and run-all.sh gives the suite a 480s per-script budget plus slow-group dispatch.

One bonus fix the shim surfaced: the watchdog used to fire at the first poll where elapsed >= budget without re-checking liveness, so a run that finished inside the 5s poll interval was killed as overran whenever the budget was under the interval — unreachable at minute-scale CI budgets, but wrong, and now guarded.

baijum:
The run-all.sh leg failures were my miss, not a flake: `tests/scheme/test-runner/shell-timeout-override.sh` (the #2557 guard) pins `PER_SCRIPT_TIMEOUTS` as one exact fixed-string line, and the new per-script entry for the watchdog suite had been spelled as a two-line assignment. Fixed in e6214043 — the table is one line again, the guard's expected line is updated (with a note on the new tenant), and the routing assertions now check the second entry's own budget. The guard test passes locally; I also checked that no other test pins `is_slow_shell_test` or the dispatch functions. My own suite passed on the failed legs (the macOS log shows `Repo tools … PASS`); the failures were entirely the guard.

tools/run-unit-test-chunk.sh
```diff
@@ -119,11 +182,39 @@ TOOLING_FILES="cli cli_spec completions config check_lint tests_check doctor
...
+# not evidence (a clean level can simply be slower than a clipped
+# allowance): an under-funded timeout is reported INCONCLUSIVE and never
+# narrows the descent, while an under-funded COMPLETION still exonerates
+# (finishing is finishing whatever the allowance was). Below MIN (BASE/3:
+# a fraction of even the compile) a level cannot decide anything and is
+# not started. Natively everything finishes far under the estimate; the
+# two constants are env-overridable for fast hosts and for the shim-driven
+# regression test.
+BISECT_LEVEL_BASE_SECS="${KAAPPI_BISECT_LEVEL_BASE_SECS:-120}"
+BISECT_LEVEL_PER_FILTER_SECS="${KAAPPI_BISECT_LEVEL_PER_FILTER_SECS:-25}"
```
baijum: **The estimate is per *filter*, but the `rest` chunk's tests per filter are skewed 2.4–4.7×, which is more than the 2× margin.** Counting `^test "` per file for the 46 `rest` filters in the order the descent splits them:

| subset | tests | average for that size |
|---|---|---|
| first half (23 filters) | 424 | 454 |
| heaviest 6-filter window (`tests_ir`..`tests_macros_procedural`) | 281 | 118 |
| heaviest 3-filter window (`tests_gc_worklist`..`tests_libraries`) | 153 | 59 |
| heaviest single filter (`tests_ir`) | 94 | 20 |

At the observed ~0.6 s/test under QEMU (907 tests in ~9.5 m after the ~1.5 m compile), a *clean* level in that region needs about: 6 filters → 90 + 170 ≈ 260 s against a 270 s estimate; 3 filters → 90 + 92 ≈ 180 s against 195 s; `tests_ir` solo → 90 + 57 ≈ 147 s against the 145 s confirm allowance. Depth 1 is fine (the halves are balanced), but from depth 3 down the margin is gone, and the last case matters most: the solo confirm can time out at its "full estimated allowance" on a clean file and print `LOCALISED AND CONFIRMED` for it, which is the confidently-wrong verdict the INCONCLUSIVE machinery was added to avoid.

The `rest` derivation already greps every file, so the fix is cheap and keeps the same shape: carry a test count per filter, make the estimate `BASE + PER_TEST × tests-in-subset` (with `PER_TEST` calibrated the same way, ~2× the observed 0.6 s), and the confirm allowance follows automatically. Non-blocking, but it is the same "clean misread as wedged" class as before, now confined to the deep levels.

tools/run-unit-test-chunk.sh
```diff
@@ -189,10 +280,294 @@ log="$(mktemp)"
...
+# that grandchild is the emulator wrapper, also not `zig`). zig prints
+# nothing (see the header), so the tree's shape is the only signal, and
+# the verdicts quote it as a heuristic, not a fact.
+phase_of() {
+    local pid="$1" child gc base verdict
+    command -v pgrep > /dev/null 2>&1 || { echo "unknown"; return; }
+    verdict=""
+    for child in $(pgrep -P "$pid" 2>/dev/null); do
+        for gc in $(pgrep -P "$child" 2>/dev/null); do
+            base="$(ps -o comm= -p "$gc" 2>/dev/null | awk '{print $1}')"
```
baijum: Nit: on macOS `ps -o comm=` prints the executable's full path, and the `awk '{print $1}'` splits it on whitespace, so a checkout under a directory with a space in its name yields a first word that is not `zig` and classifies a compile as `running tests`. Linux's 15-character `comm` is never a path, so CI is unaffected. `${base##*/}` on the unsplit line (`read -r base < <(ps -o comm= -p "$gc")`) avoids it.

tests/scheme/tools/unit-chunk-watchdog-2560.sh
```diff
@@ -0,0 +1,164 @@
...
+#   otherwise sleep SHIM_DELAY, print SHIM_TIMEOUT_NAME as a zig per-test
+#   timeout line (exercising the script's sed parser, spaces included),
+#   print a summary the script's floor check accepts, exit SHIM_EXIT.
+shim="$work/zig"
+cat > "$shim" <<'SHIM'
+#!/usr/bin/env bash
+set -u
+if [ -n "${SHIM_WEDGE_ONCE:-}" ] && [ ! -f "${SHIM_WEDGE_ONCE}" ]; then
+    : > "${SHIM_WEDGE_ONCE}"
+    sleep 9999 & sleep 9999 & wait
```
baijum: Nit, non-blocking: the shim's hang has children but no grandchildren, so `phase_of` always answers `unknown` here and its `compiling` / `running tests` branches rest on the manual verification in the reply. Two cheap additions would pin them: hang via `bash -c 'sleep 9999 & wait'` (a `sleep` grandchild → `running tests`), and a second variant that execs a `sleep` copied or symlinked as `$work/bin/zig` (a `zig` grandchild → `compiling`), then assert the phase text on the `WALL BUDGET EXCEEDED` line.

tools/run-unit-test-chunk.sh
```diff
@@ -26,15 +27,77 @@
...
+# filter list with a TIGHT per-test bound (`KAAPPI_BISECT_TEST_TIMEOUT`,
+# default 90s -- bisection is localisation, not adjudication). A half that
+# cannot finish in its level budget holds a process-level wedge; the
+# descent follows those halves down to a single filter, confirmed by a
+# solo run. A half that finishes never exonerates slow TESTS -- the tight
+# bound kills those mid-level and zig flushes their names when the level
+# exits, which the script echoes as NOTE lines; if no level ever wedged,
+# the whole chunk is re-run under the tight bound (the same filter set as
+# the chunk run, so its compile is cached whenever that run got past
+# compiling) so every per-test culprit is named in full context. The last line before any kill therefore names the filter, the
```
baijum: Nit: this rewrap left one ~130-column line in an otherwise 78-column comment block.

baijum (commented):
Reviewed 01c9f85b against the four earlier threads.

**Verified.**
- The `budget_note` bug is fixed, and the new case 1 covers the unbudgeted heartbeat path.
- The suite passes on this machine under both bash 5.3 and bash 3.2, 2 min 18 s each, with no orphaned shim children afterwards, and `shellcheck -S warning` is clean on both files. The 480 s per-script timeout has ample headroom.
- Funding levels at their full estimate and reporting a clipped timeout as `INCONCLUSIVE` resolves the design point: on the 1080 s pot a first-half wedge in `rest` now yields one decisive level plus an honest stop, and a second-half wedge funds three.
- The snapshot-before-TERM change to `tree_pids` is correct and was a real gap in the first commit: once `zig build` exits, `pgrep -P` cannot find the re-parented runner, so the earlier KILL walk signalled nothing.
- The liveness re-check after the poll sleep is a genuine fix too, not just a shim artefact: a run that finishes inside the 5 s poll window was previously reported as overran.

**One residual finding, non-blocking** (inline at the per-filter constant): the estimate counts filters while the `rest` chunk's tests per filter are skewed 2.4–4.7×, which exceeds the 2× margin from depth 3 down, and at the solo confirm step `tests_ir` alone needs about as long as its allowance. So `LOCALISED AND CONFIRMED` can still fire on a clean file. Weighting the estimate by test count per file, which the `rest` derivation already greps, closes it.

Three nits inline. Happy for this to land with the nits deferred; the test-count weighting is worth a follow-up commit or issue either way.

tools/run-unit-test-chunk.sh
baijum: Done in 39fd87b5 — each filter carries its file's `grep -c '^test "'` count (the rest derivation already touches every file), the estimate is `BASE + PER_TEST × tests-in-subset` with PER_TEST at 12 tenths (~2× the observed 0.63s/test), and the confirm allowance follows from the suspect's own count, so your worst case inverts: the tests_ir solo confirm gets 233s against its ~147s clean need, and the first half estimates 628s against ~357s. Knob renamed to `KAAPPI_BISECT_LEVEL_PER_TEST_TENTHS` (tenths because bash arithmetic is integer and 1.2s is the calibration). Re-checked against your table's windows: the heaviest 6-filter window (281 tests) estimates 457s against a ~260s clean need.

tools/run-unit-test-chunk.sh
baijum: Fixed — `read -r base < <(ps -o comm= …)` then basename, with a comment on the macOS-full-path vs Linux-15-char-comm difference. CI legs are on Linux comm= so nothing observable changes there, but the space-in-path checkout is now classified correctly too.

tests/scheme/tools/unit-chunk-watchdog-2560.sh
baijum: Both branches are pinned now. The shim's wedge runs through a grandchild (`bash -c 'sleep 9999 & wait'`), and case 5 asserts `WALL BUDGET EXCEEDED … (running tests)`; case 6 swaps the grandchild for a copy of sleep named `zig` and asserts `(compiling)`. One platform wrinkle your suggestion surfaced: macOS AMFI SIGKILLs a copied system binary that kept no valid signature, so the test re-signs the copy ad hoc on Darwin (Linux and the BSDs need nothing). Case 6 failed exactly that way before the codesign step, which is the pin doing its job.

tools/run-unit-test-chunk.sh
baijum: Rewrapped.

baijum:
The riscv64 `rest` chunk wedged on run 34327964811 — the first live encounter with the #2560/#2488 phenomenon since this machinery landed — and for the first time the log says where. Rerun requested; the analysis below is the evidence either way.

What the log shows ([job 102396…](https://github.com/kaappi/kaappi/actions/runs/34327964811)):

- Phase 1 ran its full 22m wall budget without finishing, killed with phase `(running tests)`. Healthy is 9m46s–11m32s, so ≥10m was stuck after the healthy envelope — and since an 8m per-test bound never carried the run to a named exit, this is a wedge the runner cannot see (process-level), not an ordinary hung test.
- Bisection depth 1 (first 23 filters, full 628s estimate) also did not finish → the wedge reproduces inside that half.
- Depths 2–5 completed at healthy speed (110s, 130s, 80s, 85s — so the runner was not uniformly slow, which would have been the false-positive explanation) and exonerated everything up to index 20.
- Depth 6 got only 43s of a 136s estimate (pot dry) and was correctly reported **INCONCLUSIVE** — no verdict — leaving:

```
== bisect incomplete: 2 suspect filter(s) remain:
==   tests_ffi
==   tests_gc_worklist
```

Narrowed from 46 filters to 2, with the one level that could not decide refusing to. That is the reopen criterion #2488 asked for, on the first try. A larger `KAAPPI_BISECT_BUDGET` (or the direct-binary lever noted in the issue) would finish the descent to one file.

The step's exit 124 is by design: a chunk that cannot finish inside 22m fails the leg loudly. If the rerun passes, this was the intermittent wedge doing what it has always done — except this time it left a name.

baijum:
Rerun of the failed jobs: **all green**, including riscv64-test. So the wedge was the intermittent one doing what it has always done — except this time the log carries `tests_ffi` / `tests_gc_worklist` as the narrowed suspects for the next occurrence (or for #2488, whose reopen criterion this now satisfies). The rerun's own `rest` chunk completed inside its budget, which is the expected bimodality from the issue.

tests/scheme/tools/unit-chunk-watchdog-2560.sh
```diff
@@ -0,0 +1,209 @@
...
+
+# Case 4: a level the pot can only clip. The chunk run and depth 1 both
+# hang (the counter file seeds two hangs), and the pot funds 10s of a 25s
+# estimate: the timeout must be reported INCONCLUSIVE and must NOT narrow
+# the descent.
+echo 2 > "$work/count"
+rc=0
+# shellcheck disable=SC2086  # base_env is a deliberate VAR=val word list for env
+env $base_env KAAPPI_CHUNK_BUDGET=3 KAAPPI_BISECT_BUDGET=10 \
+    KAAPPI_BISECT_LEVEL_BASE_SECS=30 \
```
baijum, line 181: **Case 4 sits exactly on the level floor and can flake by one clock tick.**

With `BASE_SECS=30` the floor is `30/3 = 10 s`, and the pot is `10`. The script stamps `pot_start=$(date +%s)`, prints the bisect banner, then stamps `now=$(date +%s)` for depth 1. Whenever a second boundary falls between those two stamps, `remaining` is 9, the level is refused as "too shallow" instead of run clipped, and all three of this case's assertions fail (`INCONCLUSIVE` never printed, no `RESULT: inconclusive -- the pot could not fund decisive levels`).

Reproduced deterministically with an always-hanging shim by passing the pot the tick produces:

```
=== KAAPPI_BISECT_BUDGET=10 ===
== bisect depth 1: 23 filter(s) bignum..tests_gc_worklist, budget 10s of a 30s estimate
== bisect depth 1: ... did NOT finish in 10s, but that allowance was clipped below the 30s estimate -- INCONCLUSIVE ...
== RESULT: inconclusive -- the pot could not fund decisive levels ...
=== KAAPPI_BISECT_BUDGET=9 ===
== bisect: remaining pot (9s) is too shallow for depth 1 (a level needs >= 10s to tell wedged from clean)
== RESULT: nothing reproduced under bisection -- the phase-1 overrun was order- or load-dependent, or slower than every budget here
```

The window is the fork latency of one `echo` plus one `date`, so it is a low-single-digit-percent flake per run under load, across three CI legs. Any slack fixes it: `KAAPPI_BISECT_BUDGET=12` (allow 12 vs floor 10), or `BASE_SECS=24` (floor 8). Worth a one-line comment saying the pot must clear the floor by more than a second, since the same shape recurs if someone retunes these.

tools/run-unit-test-chunk.sh
```diff
@@ -174,7 +275,15 @@ case "$chunk" in
 esac
 
 filters=()
-for n in $names; do filters+=("-Dtest-filter=$n.test."); done
+filter_tests=()
+for n in $names; do
+    filters+=("-Dtest-filter=$n.test.")
+    # The bisection's level estimates are per TEST (see the constants
+    # above), so carry each file's declared-test count alongside its
+    # filter. grep -c prints 0 (and exits 1, which set -u does not mind)
```
baijum: Nit: the option that would mind a `grep -c` exit of 1 here is `set -e`, which this script does not set; `set -u` is about unset variables and is unrelated. (With `pipefail` the substitution's status is simply discarded by the array append, which is the actual reason this is safe.)

baijum (commented):
Reviewed e6214043 and 39fd87b5.

**Verified.**
- Per-test funding is implemented as suggested and the quoted numbers check out: first half 424 tests → 629 s estimate against ~357 s clean need; `tests_ir` solo → 233 s against ~147 s; the heaviest 6-filter window → 457 s against ~260 s. The solo confirm now derives its allowance from the suspect's own count, so the false `LOCALISED AND CONFIRMED` path is closed.
- `phase_of` reads the whole `comm` line; both branches are now pinned by cases 5 and 6, and the macOS ad-hoc re-sign is the right handling for the copied `sleep`.
- The suite passes here under bash 5.3 and bash 3.2, 2 min 28 s each, no orphaned shim children; `shell-timeout-override.sh` passes with the collapsed one-line table. `shellcheck -S warning` is clean on the two new/changed scripts (the SC2034 warnings in `shell-timeout-override.sh` predate this PR: those globals feed the sourced run-all functions).
- The 480 s per-script budget leaves ample headroom.

**One finding, small but real** (inline in the test): case 4's pot equals the level floor exactly, so a one-second clock tick between the pot's start stamp and depth 1's stamp flips it from `INCONCLUSIVE` to "too shallow" and fails the case. Reproduced deterministically with pot 9. Two seconds of slack fixes it.

One wording nit inline. With the slack added this is good to merge from my side.

tests/scheme/tools/unit-chunk-watchdog-2560.sh
baijum: Fixed in 6df157c5 — pot is now 12 against the floor of 10, with a comment on the why (the start-stamp/depth-1-stamp tick) so the more-than-a-second rule survives a retune. Thanks for the deterministic repro; it made the one-line diagnosis easy.

tools/run-unit-test-chunk.sh
baijum: Reworded — the comment now says the status is discarded by the command substitution feeding the array append, and that `set -e` (unset here) is what would mind.

tests/scheme/tools/unit-chunk-watchdog-2560.sh
```diff
@@ -0,0 +1,209 @@
...
+exit "${SHIM_EXIT:-0}"
+SHIM
+chmod +x "$shim"
+
+# Common knobs: the pinned KAAPPI_TEST_TIMEOUT guards against a caller's
+# run-all.sh KAAPPI_TEST_TIMEOUT (seconds, for Scheme tests) leaking into
+# the --test-timeout grammar (a Zig duration). The level-funding constants
+# are scaled down so a level's "full estimate" is seconds, not minutes.
+base_env="KAAPPI_ZIG=$shim KAAPPI_TEST_TIMEOUT=8m KAAPPI_BISECT_TEST_TIMEOUT=90s"
+base_env="$base_env KAAPPI_BISECT_LEVEL_BASE_SECS=2 KAAPPI_BISECT_LEVEL_PER_TEST_TENTHS=0"
```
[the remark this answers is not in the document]
baijum: Added case 7 in 6df157c5: a nonzero per-test term, two probe windows of different density, and the test derives both expected estimates independently from the same `grep -c` of the same files — so the assertions track the tree as it changes while still failing on any flat/per-filter regression. Mutation-checked: reverting the script's estimate to `BASE + TERM × n` fails the case with 'depth 1's estimate is not 44s from its subset's test count'.

tests/scheme/tools/unit-chunk-watchdog-2560.sh
```diff
@@ -0,0 +1,248 @@
...
+    for f in $(bash tools/run-unit-test-chunk.sh --list rest \
+               | sed -n "${start},$((start + n - 1))p" \
+               | sed 's/^-Dtest-filter=//; s/\.test\.$//'); do
+        c=$(grep -c '^test "' "src/$f.zig" | tr -d ' ')
+        total=$((total + c))
+    done
+    echo $(( 2 + total / 10 ))
+}
+e1=$(est_of 1 23)
+e2=$(est_of 24 11)
```
baijum: **Case 7 hardcodes the split arithmetic for exactly 46 `rest` filters, so the next test file added to `rest` fails it.**

`est_of 1 23`, `est_of 24 11`, and the two `N filter(s)` literals in the assertions are what the descent produces for `hi=46` only. The `rest` chunk is derived precisely so that a new `src/*.zig` with a `test "…"` lands in it automatically, and with 47 filters depth 2 becomes 12 filters at 24..35, so the depth-2 assertion fails with "estimate is not Ns" on an unrelated PR; with 45 it is depth 1 that moves. The case's stated goal is to track the tree, and its estimates already do; only the windows are pinned.

Deriving them the way the script does keeps the case tree-proof with a few lines:

```bash
total=$(bash tools/run-unit-test-chunk.sh --list rest | wc -l | tr -d ' ')
mid1=$((total / 2))                 # depth 1: filters 1..mid1
mid2=$(((mid1 + total) / 2))        # depth 2 after depth 1 completes: mid1+1..mid2
n1=$mid1; n2=$((mid2 - mid1))
e1=$(est_of 1 "$n1"); e2=$(est_of $((mid1 + 1)) "$n2")
```

and then `"bisect depth 1: $n1 filter(s) …"` / `"bisect depth 2: $n2 filter(s) …"` in the two greps. Non-blocking for this PR, but it will bite the first `rest` addition after merge, and the failure message will point at the funding term rather than at the file count.

baijum (commented):
Reviewed 6df157c5.

**Verified.**
- Case 4 now runs with a 12 s pot against the 10 s floor, and the comment records the tick rule.
- The comment on the `grep -c` status is right now.
- Case 7 does what the reply says: I mutated the estimate back to a per-filter term in place and the suite fails at exactly `case 7: depth 1's estimate is not 44s from its subset's test count`; restored, it passes.
- The suite passes here under bash 5.3 and bash 3.2, run in parallel, 3 min 10 s each, no orphaned shim children, shellcheck clean. Still well inside the 480 s per-script budget.

**One non-blocking finding** (inline): case 7 pins its two probe windows to the current count of 46 `rest` filters. Since `rest` is derived so that new test files join it automatically, the first such file after merge fails case 7 with a message that blames the funding term. Deriving the windows from `--list` the way the script derives `mid` makes the case tree-proof; snippet inline.

Fine to merge as is from my side, with that as a follow-up or folded in, whichever you prefer.

tests/scheme/tools/unit-chunk-watchdog-2560.sh
baijum: Folded in (your snippet as-is, with the parity note in the comment) — verified end-to-end both ways: the suite passes with a dummy 47th `rest` filter dropped into the tree (the odd-count split your walkthrough gives: depth 1 stays 23, depth 2 becomes 12 at 24..35) and on the normal 46. Thanks for catching that the failure message would have pointed at the funding term; that is exactly the kind of landmine that costs someone an afternoon.

baijum (commented):
Reviewed ee29c64d.

**Verified.** Case 7 now derives both probe windows from `--list` with the script's own midpoint arithmetic, and the assertions use the derived filter counts.

- The suite passes here under bash 5.3 and bash 3.2, run in parallel, 3 min 12 s each, no orphaned shim children, shellcheck clean.
- The tree-proof claim holds end to end: with a dummy `src/tests_zz_probe_2561.zig` declaring one test, `--list rest` reports 47 filters, depth 2 moves to 12 filters at 24..35, and the whole suite still passes; with the file removed it is back to 46 and passes again.

Nothing further from me. All five rounds are resolved; good to merge.

Issue #2562, opened 2026-09-09
REPL structural-editing keys are unreachable on a default macOS terminal: Option-Shift-S inserts Í, Escape-then-S wipes the form


baijum:
## Summary

The four structural-editing keys from #2216 (`alt-shift-S` slurp, `alt-shift-B` barf, `alt-shift-R` raise, `alt-y` rotate) are bound to `ESC <char>`, i.e. they need the terminal to send Option as Meta. **No macOS terminal does that by default** — Terminal.app, iTerm2, kitty, Alacritty and Ghostty all send the *composed* character instead (WezTerm is the one exception I know of). So on the primary dev platform, out of the box, the documented keys do not run the commands; they insert a stray character into the form, and the natural fallback (press Escape, then the letter) wipes the whole input.

`,help` and `F1` print the bindings as `alt-…`; there is no Alt key on a Mac keyboard, and nothing user-facing says the feature needs terminal configuration (the dev doc `docs/dev/repl.md` does — see the companion docs issue).

## Reproduction

Driven over a pty against `zig-out/bin/kaappi` at bd57760e (`Kaappi Scheme v0.26.1` build; none of the files involved changed since), `TERM=xterm`, US layout characters. Each case types `(list 1 2) 3`, moves the cursor to just after `2`, sends the bytes a Mac terminal actually emits, then presses Enter:

| Bytes sent | What a Mac user did | Result |
|---|---|---|
| `ESC S` (within 200 ms) | Option-Shift-S with "Option as Meta" **enabled** | `(list 1 2 3)` → `(1 2 3)` — slurp works |
| `C3 8D` (`Í`, U+00CD) | Option-Shift-S in **default** Terminal.app / iTerm2 | buffer becomes `(list 1 2Í) 3` → `read error[KP1004]: invalid number literal '2Í'` |
| `ESC`, 400 ms pause, `S` | Escape, then S (Emacs / readline Meta habit) | lone ESC runs `edit_delete_all` — the whole form is gone — then `S` is inserted → `error[KP3001]: undefined variable 'S'` |
| `C2 A5` (`¥`, U+00A5) | Option-y in default Terminal.app / iTerm2 | `(list 1 2 3¥)` → `read error[KP1002]: unexpected character` |

[...]

PR #2565, opened 2026-09-09
Normalise Alt+Shift+letter on the Windows console so slurp/barf/raise fire

### LINKED ISSUE #2564: Windows console: Alt+Shift+S/B/R decode with a SHIFT bit that WITH_ALT(...) never matches, so slurp/barf/raise cannot fire (rotate can)
## Summary

By code reading — **not yet reproduced on a real console** — the three *shifted* structural-editing bindings (`alt-shift-S` slurp, `alt-shift-B` barf, `alt-shift-R` raise) can never fire on the Windows console, while the unshifted `alt-y` rotate can. The dispatch arms compare against `WITH_ALT('S')` (`'S' | KEY_MOD_ALT`), but the Windows key path delivers `'S' | KEY_MOD_ALT | KEY_MOD_SHIFT`, and no `case` matches that, so the keystroke is silently dropped (`debug_msg("edit: ignore code")` in a non-debug build is nothing at all).

## Trace

`vendor/isocline/src/tty.c`, the `_WIN32` `tty_waitc_console` path:

1. line 886–887: Alt+Shift+S arrives as one `KEY_EVENT_RECORD` with `dwControlKeyState` carrying `LEFT_ALT_PRESSED | SHIFT_PRESSED` and `uChar.UnicodeChar = 'S'` (the shifted character, as the console reports it) → `mods = KEY_MOD_ALT | KEY_MOD_SHIFT`.
2. line 948: `chr != 0`, so `tty_cpush_csi_unicode(tty, mods, 'S')`.
3. line 456–465: the fast paths require `mods == 0`, `mods == KEY_MOD_CTRL`, or `mods == KEY_MOD_SHIFT` alone; `ALT|SHIFT` takes the `else` and pushes `ESC [ 83 ; 4 u` (`csi_mods` = 1 + 1 (shift) + 2 (alt)).
4. `vendor/isocline/src/tty_esc.c:333–338`: `num2 = 4` → `3` → sets `KEY_MOD_SHIFT | KEY_MOD_ALT`. The decoded code is `'S' | ALT | SHIFT`.
5. `vendor/isocline/src/tty.c:186` `modify_code` does not strip SHIFT from letters.
6. `vendor/isocline/src/editline.c:1148` `case WITH_ALT('S')` does not match; the `default` arm's `code_is_ascii_char` / `code_is_unicode` both reject a code with modifier bits → dropped.

Alt+y goes through the same steps with `mods = KEY_MOD_ALT` only → `ESC [ 121 ; 3 u` → `'y' | ALT` = `WITH_ALT('y')` → rotate works.

On a POSIX tty the terminal has already applied Shift before sending `ESC S`, so `tty_esc.c:427` yields `'S' | ALT` with no SHIFT bit — which is why the pty smoke test (`tests/scheme/smoke/repl-structural-editing-2216.sh`, skipped on Windows) never sees this.

## Suggested fix

Normalise the Windows path to what the ESC path produces: when `KEY_MOD_ALT` is set and the character is a printable that Shift already transformed, drop `KEY_MOD_SHIFT` — either in `tty_waitc_console` before `tty_cpush_csi_unicode`, or in `modify_code`, which already exists to make virtual keys portable across platforms. The unshifted upstream bindings (`alt-d`, `alt-m`, `alt-f`, `alt-b`) are unaffected, so this is specific to the KAAPPI PATCH 3 keys, which are the only shifted-letter Alt bindings in the editor.

## Verification needed

The win11 ARM64 VM can confirm this; it needs a real console key event with Alt+Shift, so it is a hands-on check rather than an SSH script. If the console actually reports `uChar = 's'` with `SHIFT_PRESSED` instead of `'S'`, the bindings fail differently (`'s' | ALT | SHIFT`) but still never match.

Related: #2562 (macOS Option-as-Meta) and #2563 (the `,help`/F1/site wording).

baijum:
## Summary

On the Windows console, the three *shifted* structural-editing bindings —
alt-shift-S (slurp), alt-shift-B (barf), alt-shift-R (raise) — could never
fire, while the unshifted alt-y (rotate) worked.

`tty_waitc_console` (the `_WIN32` key path) delivers Alt+Shift+S as one
`KEY_EVENT_RECORD` whose `dwControlKeyState` carries `LEFT_ALT_PRESSED |
SHIFT_PRESSED` with `uChar.UnicodeChar` already the shifted glyph (`'S'`). It
builds `mods = KEY_MOD_ALT | KEY_MOD_SHIFT`, so the keystroke decodes back to
`'S' | KEY_MOD_ALT | KEY_MOD_SHIFT`. The dispatch arms in `editline.c` compare
against `WITH_ALT('S')` (= `'S' | KEY_MOD_ALT`), which carries no SHIFT bit, so
no `case` matched and the keystroke was silently dropped. alt-y carries no SHIFT
bit and matched `WITH_ALT('y')`, which is why only the shifted keys broke.

On a POSIX tty the terminal applies Shift before sending `ESC S`, so its ALT
path yields `'S' | ALT` with no SHIFT bit — which is why this is Windows-only
and the pty smoke test (skipped on Windows) never saw it.

## Fix

Normalise the Windows path to what the POSIX ESC path produces: in the
regular-character arm of `tty_waitc_console`, when `KEY_MOD_ALT` is set and the
char is already a printable that Shift transformed (`chr >= 0x20 && chr !=
0x7f`), drop `KEY_MOD_SHIFT` before `tty_cpush_csi_unicode`. The shift
information is already encoded in the glyph, so this loses nothing.

The change lives entirely inside the `#if defined(_WIN32)` console branch — not
the shared `modify_code` — so it cannot affect the POSIX/xterm decode. It is
documented as a sub-note of KAAPPI PATCH 3 in `vendor/isocline/PATCHES.md`, and
carries a `KAAPPI PATCH 3` marker at the fix site.

## Verification and its limits

- Cross-compiles cleanly for **both** Windows targets (`zig build
  -Dtarget=x86_64-windows` and `-Dtarget=aarch64-windows`) — the only builds
  that actually compile this `_WIN32`-only code.
- Host `zig build` and `zig build test` pass: unit-tests 2008 pass / 21 skip
  (2029 total), thottam-tests 88 pass.

This is a **code-reading finding**, not yet reproduced on a real console. No
host-runnable automated regression guard is feasible: `tty_waitc_console` is
`_WIN32`-only (not compiled off-Windows), and exercising it needs a real
Alt+Shift key event that an SSH script cannot inject. Rather than add a hollow
test, the invariant is captured in a clear code comment at the fix site and in
PATCHES.md. **Final confirmation requires a real Alt+Shift keypress on a Windows
console** (win11 ARM64 VM).

Closes #2564

### DIFF
--- a/vendor/isocline/PATCHES.md
+++ b/vendor/isocline/PATCHES.md
@@ -124,8 +124,12 @@ printable that Shift transformed (`chr >= 0x20 && chr != 0x7f`), drop
 encoded in the glyph, so this loses nothing and makes both platforms agree. It
 lives in the `#if defined(_WIN32)` console branch only, not the shared
 `modify_code` in `tty.c`, so it cannot affect the POSIX/xterm decode. The
-unshifted upstream Alt bindings (`alt-d`/`alt-m`/`alt-f`/`alt-b`) never carry a
-SHIFT bit and are unaffected.
+change is strictly parity-restoring: it activates no new bindings. The unshifted
+upstream Alt bindings (`alt-d`/`alt-m`/`alt-f`/`alt-b`) never carry a SHIFT bit
+and are unaffected, and their *shifted* variants remain unbound on both
+platforms — the arms match the lowercase glyph (`WITH_ALT('d')`) while a shifted
+key delivers the uppercase one, so alt-shift-D → `'D' | KEY_MOD_ALT` matches
+nothing here just as `ESC D` matches nothing on a POSIX tty.
 
 Found by code reading; end-to-end confirmation requires a real Alt+Shift key
 event on a Windows console.
--- a/vendor/isocline/PATCHES.md
+++ b/vendor/isocline/PATCHES.md
@@ -50,7 +50,7 @@ asking for 1000 gets 1000.
 ## Patch 3 — structural s-expression editing
 
 **Files:** `include/isocline.h`, `src/env.h`, `src/isocline.c`,
-`src/editline.c`, `src/editline_help.c`
+`src/editline.c`, `src/editline_help.c`, `src/tty.c`
 
 Upstream has no structural editing — no way to move a paren rather than a
 character. The patch adds four keys and one callback:
@@ -92,6 +92,43 @@ The commands and keybindings come from [bestline](https://github.com/jart/bestli
 (BSD-2-Clause) by way of kaappi#2216. The code does not: bestline's `raise` is
 an empty stub, its `rotate` rotates the kill ring rather than a form, and its
 barf and slurp count parens with no awareness of strings or comments.
+
+### Windows: normalise Alt+Shift+letter (kaappi#2564)
+
+**File:** `src/tty.c` (`tty_waitc_console`, the `_WIN32` key path).
+
+Three of the four bindings above are *shifted* letters (alt-shift-S/B/R); only
+alt-y is unshifted. On the Windows console those three could never fire while
+alt-y worked, because the shifted and unshifted Alt paths deliver different
+codes:
+
+- **POSIX tty:** the terminal applies Shift before sending `ESC S`, so the ALT
+  path in `tty_esc.c` yields `'S' | KEY_MOD_ALT` — no SHIFT bit.
+- **Windows console:** `ReadConsoleInputW` delivers Alt+Shift+S as one
+  `KEY_EVENT_RECORD` whose `dwControlKeyState` carries both `LEFT_ALT_PRESSED`
+  and `SHIFT_PRESSED`, with `uChar.UnicodeChar` already the *shifted* glyph
+  (`'S'`). `tty_waitc_console` therefore builds `mods =
+  KEY_MOD_ALT | KEY_MOD_SHIFT` and pushes `ESC [ 83 ; 4 u`, which decodes back
+  to `'S' | KEY_MOD_ALT | KEY_MOD_SHIFT`.
+
+The dispatch arms in `editline.c` compare against `WITH_ALT('S')` (=
+`'S' | KEY_MOD_ALT`), which has no SHIFT bit, so no `case` matched and the
+keystroke was silently dropped (`debug_msg("edit: ignore code")`, nothing in a
+non-debug build). alt-y carries no SHIFT bit, so it matched `WITH_ALT('y')` and
+rotate worked — which is why the bug was slurp/barf/raise-only.
+
+The fix normalises the Windows path to what the POSIX ESC path produces: in the
+"regular character" arm, when `KEY_MOD_ALT` is set and the char is already a
+printable that Shift transformed (`chr >= 0x20 && chr != 0x7f`), drop
+`KEY_MOD_SHIFT` before `tty_cpush_csi_unicode`. The shift information is already
+encoded in the glyph, so this loses nothing and makes both platforms agree. It
+lives in the `#if defined(_WIN32)` console branch only, not the shared
+`modify_code` in `tty.c`, so it cannot affect the POSIX/xterm decode. The
+unshifted upstream Alt bindings (`alt-d`/`alt-m`/`alt-f`/`alt-b`) never carry a
+SHIFT bit and are unaffected.
+
+Found by code reading; end-to-end confirmation requires a real Alt+Shift key
+event on a Windows console.
 
 ## Patch 4 — don't discard buffered input when leaving/entering raw mode
 
--- a/vendor/isocline/src/tty.c
+++ b/vendor/isocline/src/tty.c
@@ -941,6 +941,20 @@ static void tty_waitc_console(tty_t* tty, long timeout_ms)
 		}
     // regular character
     else {
+      // KAAPPI PATCH 3: see vendor/isocline/PATCHES.md — normalise Alt+Shift+letter.
+      // The Windows console reports Alt+Shift+S as one event carrying both
+      // LEFT_ALT_PRESSED and SHIFT_PRESSED with uChar already the *shifted*
+      // glyph ('S'), so `mods` becomes KEY_MOD_ALT|KEY_MOD_SHIFT. A POSIX tty
+      // applies Shift before sending `ESC S`, so its ALT path yields 'S'|ALT
+      // with no SHIFT bit — and the editor's structural-edit arms compare
+      // against WITH_ALT('S') (= 'S'|KEY_MOD_ALT), which has no SHIFT bit.
+      // Without this the shifted bindings (alt-shift-S/B/R) never match any
+      // case and are silently dropped, while the unshifted alt-y works
+      // (kaappi#2564). When Alt is held and the char is already a printable
+      // that Shift transformed, drop KEY_MOD_SHIFT so the two paths agree.
+      if ((mods & KEY_MOD_ALT) != 0 && chr >= 0x20 && chr != 0x7f) {
+        mods &= ~(code_t)KEY_MOD_SHIFT;
+      }
       tty_cpush_csi_unicode(tty,mods,chr);
 			return;
     }


vendor/isocline/src/tty.c
```diff
@@ -941,6 +941,20 @@ static void tty_waitc_console(tty_t* tty, long timeout_ms)
...
+      // glyph ('S'), so `mods` becomes KEY_MOD_ALT|KEY_MOD_SHIFT. A POSIX tty
+      // applies Shift before sending `ESC S`, so its ALT path yields 'S'|ALT
+      // with no SHIFT bit — and the editor's structural-edit arms compare
+      // against WITH_ALT('S') (= 'S'|KEY_MOD_ALT), which has no SHIFT bit.
+      // Without this the shifted bindings (alt-shift-S/B/R) never match any
+      // case and are silently dropped, while the unshifted alt-y works
+      // (kaappi#2564). When Alt is held and the char is already a printable
+      // that Shift transformed, drop KEY_MOD_SHIFT so the two paths agree.
+      if ((mods & KEY_MOD_ALT) != 0 && chr >= 0x20 && chr != 0x7f) {
+        mods &= ~(code_t)KEY_MOD_SHIFT;
```
baijum, line 956: Verified the guard's boundaries are exactly right: Tab/Enter keep SHIFT (chr < 0x20), virtual keys keep SHIFT (`chr == 0` branch — the only real `WITH_SHIFT` bindings), and AltGr is already stripped above, so it can't misfire on European layouts.

A strengthening point worth having in the record: this activates **no new bindings**. The upstream alt arms are lowercase — `WITH_ALT('d')` (editline.c:1168), `WITH_ALT('b')`, `WITH_ALT('f')` — and a shifted key delivers the *uppercase* glyph, so e.g. alt-shift-D → `'D'|ALT` still matches nothing after the fix, exactly what a POSIX terminal does with `ESC D`. The change is strictly parity-restoring: same dropped keys as POSIX, plus S/B/R/y now fire.

One cosmetic asymmetry left: the surrogate-pair arm just above (tty.c:935-941) pushes with the raw mods, so an Alt(+Shift)-prefixed astral char keeps its SHIFT bit on Windows while POSIX wouldn't set one. No binding exists at astral codepoints, so both platforms end at "ignore code" — fine to leave as-is, just noting it's the one remaining path where they disagree on paper.

vendor/isocline/PATCHES.md
```diff
@@ -93,6 +93,43 @@ The commands and keybindings come from [bestline](https://github.com/jart/bestli
...
+rotate worked — which is why the bug was slurp/barf/raise-only.
+
+The fix normalises the Windows path to what the POSIX ESC path produces: in the
+"regular character" arm, when `KEY_MOD_ALT` is set and the char is already a
+printable that Shift transformed (`chr >= 0x20 && chr != 0x7f`), drop
+`KEY_MOD_SHIFT` before `tty_cpush_csi_unicode`. The shift information is already
+encoded in the glyph, so this loses nothing and makes both platforms agree. It
+lives in the `#if defined(_WIN32)` console branch only, not the shared
+`modify_code` in `tty.c`, so it cannot affect the POSIX/xterm decode. The
+unshifted upstream Alt bindings (`alt-d`/`alt-m`/`alt-f`/`alt-b`) never carry a
```
baijum: Nit (optional sharpening): "unaffected" is true of the *bindings*, but the blast radius could be made explicit — shifted variants of those upstream letters (alt-shift-D → `'D'|ALT`) still match nothing after the fix, identical to POSIX `ESC D`, because the arms are lowercase while a shifted key delivers the uppercase glyph. One clause here ("…and their shifted variants remain unbound on both platforms, since the arms match the lowercase glyph") would save whoever re-applies these patches after an upstream merge from re-deriving the case-sensitivity argument.

baijum (commented):
**Fix verified correct — traced end-to-end against the PR head (4b74f79b), not just the diff.**

The bug chain, re-derived independently:

- Alt+Shift+S arrives as one event with `LEFT_ALT_PRESSED|SHIFT_PRESSED` and `uChar` already `'S'` → `mods = ALT|SHIFT` → `csi_mods` = 1+1+2 = **4** → `tty_cpush_csi_unicode` (tty.c:456) takes no fast path and pushes `ESC [ 83 ; 4 u`.
- `tty_read_csi` (tty_esc.c:333-338) maps num2 = 4 → 3 → sets bit 0 (SHIFT) + bit 1 (ALT): the decoded code is `'S'|ALT|SHIFT`.
- `editline.c:1148` compares against `WITH_ALT('S')` — no SHIFT bit — so nothing matches and it falls to the default `debug_msg("edit: ignore code")` at editline.c:1210, which is silence in a non-debug build. Bug confirmed.
- After the fix: `ESC [ 83 ; 3 u` → num2 = 3 → 2 → ALT only → `'S'|ALT` → matches the arm. The invariant being restored holds on the POSIX side too: the `ESC <char>` path returns `key_unicode(peek)|KEY_MOD_ALT` (tty_esc.c:429) — never a SHIFT bit for a printable.

Guard boundaries check out — every path that genuinely needs SHIFT can't reach the guard:

- Tab/Enter with Alt+Shift: chr = 9/13, excluded by `chr >= 0x20` (and `modify_code`'s KEY_ENTER remaps test exact-mods, so leaving them alone is right).
- Shift+arrows/Home/End — the only real `WITH_SHIFT(...)` bindings in editline.c — come through the `chr == 0` virtual-key branch, untouched.
- AltGr (`LEFT_CTRL|RIGHT_ALT`) is stripped before mods are built (tty.c:890-891), so the guard can't misfire on AltGr layouts.
- `mods &= ~(code_t)KEY_MOD_SHIFT` matches the house pattern a few lines up (`(DWORD)~SHIFT_PRESSED`, tty.c:886).

Also independently confirmed: compiling `tty.c` alone with `zig cc -target x86_64-windows-gnu -D_WIN32` is clean (only the expected cross-TU `ic_private` linkage warnings), and CI `windows-cross` is green on both x86_64 and aarch64. PATCHES.md is outside the markdownlint globs (`!vendor/**`), so no format gate applies there.

**One strengthening observation:** the normalization activates *no* new bindings — see the inline comment on the guard. The change is strictly parity-restoring, which is a stronger safety property than "cannot affect the POSIX decode".

**Non-blocking follow-ups:**

1. **AGENTS.md patch-count drift (pre-existing, not this PR):** AGENTS.md still says the isocline copy carries "six changes", but PATCHES.md has documented seven since Patch 7 (sticky-Meta, #2562) landed. This PR extends Patch 3 rather than adding a patch, so it doesn't change the count — but the next reader's arithmetic is already off by one. `six` → `seven` whenever AGENTS.md is next touched.
2. **On "no feasible regression test":** agreed for this PR's scope, but for the record — `WriteConsoleInputW` *can* synthesize `KEY_EVENT_RECORD`s with arbitrary `dwControlKeyState` (the same mechanism `tty_async_stop` already uses to inject ^C, tty.c:986), and the `windows-*-test` jobs run on real Windows runners. A small sidecar that attaches to kaappi's console and writes an Alt+Shift+S record would exercise `tty_waitc_console` end-to-end. Disproportionate for one keybinding today; worth remembering the mechanism exists if a Windows console harness ever materializes.

windows-arm-test / windows-x64-test were still pending at review time — nothing in this change can affect them beyond compiling, which windows-cross already proved.

vendor/isocline/PATCHES.md
baijum: Verified — this rewording is accurate, including the part that's easy to get wrong: "matches nothing" means *ignored*, not *inserted*. The default arm's predicates (`code_is_ascii_char` / `code_is_unicode`, tty.c:77-97) compare the raw `code_t` with the modifier bits included, so `'D' | KEY_MOD_ALT` (0x20000044) fails both `c <= 0x7F` and `c <= KEY_UNICODE_MAX` and falls to the ignore branch — on either platform, same as POSIX `ESC D`. And `tty.c` is byte-identical to the commit this comment was left on, so nothing else moved. Point closed; the paragraph now carries the full blast-radius argument for whoever re-applies the patches later.

baijum (commented):
Reviewed **d2d9b840** against 4b74f79b (the commit the previous review verified): `tty.c` is byte-identical, so everything established there stands. The only change is the PATCHES.md paragraph, which now states the blast radius explicitly — and every claim in the new sentence checks out against the code.

The one non-obvious claim, verified: an unmatched ALT+printable is *ignored*, not inserted. The default arm's predicates compare the raw `code_t` with modifier bits included (`code_is_ascii_char`: `c <= 0x7F`; `code_is_unicode`: `c <= KEY_UNICODE_MAX` — tty.c:77-97), so `'D' | KEY_MOD_ALT` fails both and lands in the `ignore code` branch, identically on both platforms. "Strictly parity-restoring" is now not just asserted but derivable from the doc alone.

DCO and format are green on this commit; the test/windows-cross jobs are re-running against an unchanged `tty.c`, so their previous green runs are the evidence that matters.

This fully addresses the review points (the AGENTS.md `six` → `seven` count fix stays a follow-up for whenever that file is next touched, as noted). The only thing still outstanding is the PR's own stated caveat: final confirmation with a real Alt+Shift keypress on a Windows console.

PR #2566, opened 2026-09-09
Treat REPL ESC as a sticky Meta prefix so alt-<key> bindings are reachable

### LINKED ISSUE #2562: REPL structural-editing keys are unreachable on a default macOS terminal: Option-Shift-S inserts Í, Escape-then-S wipes the form

baijum:
## Problem

The four structural-editing keys from #2216 — alt-shift-S (slurp),
alt-shift-B (barf), alt-shift-R (raise), alt-y (rotate) — are bound as
`ESC <char>` (Meta). But no default macOS terminal sends Option/Alt as Meta:
Terminal.app, iTerm2, kitty, Alacritty and Ghostty all insert the *composed*
character instead. So out of the box on the primary dev platform these keys
never ran — they inserted a stray glyph (`Í`, `¥`, …), and the natural readline
habit (press Escape, then the letter) ran `edit_delete_all` and **wiped the
whole form**. That is data loss, and the four structural keys have no ctrl/arrow
alternative, so the feature was simply unreachable.

## Fix

This implements option 1 from the issue (the author's recommendation): treat
`ESC` as a **sticky Meta prefix**, exactly like readline / zsh / Emacs. `ESC`
followed by any key becomes that key with Alt, with **no timeout**. A genuine
terminal escape sequence (arrows, mouse) still arrives as one burst and decodes
immediately; only a lone ESC waits — for the key the user is about to press —
and composes it as alt-<key>.

Three sites in the vendored isocline copy:

- `src/tty.c` (`tty_read_timeout`): the interactive ESC decode now reads the
  byte after ESC with a blocking (`-1`) read instead of `esc_initial_timeout`.
- `src/tty.c` (`tty_new`): `esc_initial_timeout` is unified to one
  platform-independent value; the old `__APPLE__` 200ms bump lost its rationale
  (it only helped a terminal already configured to send Option as Meta). The
  field now only bounds the terminal query-response readers.
- `src/editline.c`: the lone `KEY_ESC` arm no longer runs `edit_delete_all` —
  that was the data-loss path.

## Behaviour change (deliberate)

**A lone Escape no longer clears the input.** `ctrl-u` (delete-to-start) and
`ctrl-c` (cancel) still do. This is the convention every other line editor on
macOS already follows, and it turns the old destructive "Escape then S" into
the slurp the user meant.

This is a change to the vendored isocline library, so it is a **new
`vendor/isocline/PATCHES.md` entry (Patch 7)**, with a `KAAPPI PATCH 7` marker
at each changed site.

## Testing

`tests/scheme/smoke/repl-structural-editing-2216.sh` gains a case that drives
the keys the new way under a pty: a lone ESC, a pause far longer than the removed
compose window, then `S` — which now composes to slurp. Verified it fails
without the fix (form wiped, `S` undefined) and passes with it. `zig build test`
passes; the full smoke test passes.

Closes #2562

### DIFF
--- a/docs/dev/repl.md
+++ b/docs/dev/repl.md
@@ -55,7 +55,7 @@ cursor:
 
 They are listed under F1 too, and only when the callback is set. Each is
 reached by pressing `ESC` then the character. Since patch 7 (kaappi#2562) `ESC`
-is a sticky Meta prefix with no timeout, so `ESC` then `S` composes to
+is a sticky Meta prefix with no timeout, so `ESC` then `shift-S` composes to
 `alt-shift-S` on *every* terminal, whether or not it sends Option as Meta — no
 "Use Option as Meta key" setting is needed. The deliberate cost is that a lone
 `ESC` no longer clears the input; `ctrl-u` and `ctrl-c` still do.
--- a/docs/dev/repl.md
+++ b/docs/dev/repl.md
@@ -54,9 +54,11 @@ cursor:
 | alt+y | rotate | `(if a\| b c)` → `(if b c a\|)` |
 
 They are listed under F1 too, and only when the callback is set. Each is
-`ESC` followed by the character, so a terminal that does not send Option as
-Meta (macOS Terminal.app, unless "Use Option as Meta key" is on) will not
-deliver them.
+reached by pressing `ESC` then the character. Since patch 7 (kaappi#2562) `ESC`
+is a sticky Meta prefix with no timeout, so `ESC` then `S` composes to
+`alt-shift-S` on *every* terminal, whether or not it sends Option as Meta — no
+"Use Option as Meta key" setting is needed. The deliberate cost is that a lone
+`ESC` no longer clears the input; `ctrl-u` and `ctrl-c` still do.
 
 `src/repl_sexp.zig` holds the transforms — pure functions over
 `(buffer, byte cursor)`, so they unit-test without a terminal. The keys and the
--- a/tests/scheme/smoke/repl-structural-editing-2216.sh
+++ b/tests/scheme/smoke/repl-structural-editing-2216.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 # kaappi#2216: structural s-expression editing at the REPL.
+# kaappi#2562: ESC is a sticky Meta prefix so those bindings are reachable
+#              without a Meta-configured terminal.
 #
 # The transforms themselves are unit-tested in `src/repl_sexp.zig`. What this
 # covers is the half that lives in C and cannot be reached from Zig: isocline's
@@ -11,6 +13,12 @@
 # key, submits, and asserts on what the *evaluator* printed. That keeps the
 # check independent of how the editor redraws the line: the only way to print
 # `(1 2)` from `(list 1 2 3)` is for barf to have moved the paren.
+#
+# The final case is the #2562 regression: it sends a lone ESC, waits far longer
+# than the old ESC-compose window, then sends `S`. With the sticky-Meta prefix
+# (KAAPPI PATCH 7) that composes to alt-S (slurp); before the fix the lone ESC
+# ran `edit_delete_all` and wiped the whole form, and `S` was left as an
+# undefined variable.
 
 set -eu
 
@@ -173,6 +181,35 @@ for typed, lefts, key, expect, reject in CASES:
     if not ok or (reject is not None and reject in produced):
         failures.append((key, expect, produced))
 
+# kaappi#2562: ESC as a sticky Meta prefix, driven the way a Mac user actually
+# reaches these keys — press Escape, then the letter. The pause below is far
+# longer than the ESC-compose window this fix removed (200 ms on macOS, 100 ms
+# elsewhere), so it exercises the *no-timeout* prefix specifically. Before the
+# fix the lone ESC deleted the whole form and `S` evaluated as an undefined
+# variable; with it, ESC+S composes to alt-S = slurp. Same form/cursor as the
+# slurp case above so the only variable is how the key is delivered.
+esc_label = b'\x1b <pause> S'
+mark = len(buf)
+send(b'(list 1 2) 3')
+if not wait_for(b'(list 1 2) 3', mark, 20):
+    failures.append((esc_label, b'(list 1 2) 3', seen(mark)))
+else:
+    for _ in range(3):
+        send(b'\x02')          # ctrl-b: cursor to just after the `2`
+    idle(0.3)
+    send(b'\x1b')              # lone ESC — the sticky Meta prefix, no timeout
+    time.sleep(0.4)            # longer than the removed ESC-compose window
+    send(b'S')                 # composes to alt-S = slurp
+    idle(0.3)
+    send(b'\r')
+    ok = wait_for(b'\n(1 2 3)\n', mark, 20)
+    idle()
+    produced = seen(mark)
+    # Without the fix the form is gone and `S` is undefined; guard both the
+    # positive result and that no wipe-then-error snuck through.
+    if not ok or b'undefined variable' in produced:
+        failures.append((esc_label, b'(1 2 3)', produced))
+
 send(b',quit\r')
 while pump(1.0):
     pass
@@ -193,7 +230,7 @@ status=$?
 set -e
 
 case $status in
-    0)  echo "PASS: barf, slurp, raise and rotate all edit the buffer"; exit 0 ;;
+    0)  echo "PASS: barf, slurp, raise, rotate edit the buffer; ESC is a sticky Meta prefix"; exit 0 ;;
     77) echo "SKIP: no usable pty for the REPL here"; exit 77 ;;
     *)  echo "FAIL: structural editing did not take effect"; exit 1 ;;
 esac
--- a/vendor/isocline/PATCHES.md
+++ b/vendor/isocline/PATCHES.md
@@ -4,7 +4,7 @@ Vendored from <https://github.com/daanx/isocline> at commit
 `8d6dc1ef95b1b46711e66eb23d39d4467a0fcdac` (2026-04-23, v1.1.0), MIT licensed —
 see `LICENSE`.
 
-**This is a patched copy.** Six changes diverge from upstream. Each is marked
+**This is a patched copy.** Seven changes diverge from upstream. Each is marked
 in the source with a `KAAPPI PATCH <n>` comment pointing here. When updating
 isocline, re-apply them; `grep -rn 'KAAPPI PATCH' vendor/isocline/` finds every
 site.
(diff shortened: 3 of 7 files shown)


vendor/isocline/PATCHES.md
```diff
@@ -271,6 +271,58 @@ never compiles isocline (`use_isocline = !is_wasm_target` in `build.zig`), so
...
+   `tty_read_dsr_response` from Patch 5), which use `2*esc_initial_timeout`.
+
+3. **`src/editline.c`** — the lone `KEY_ESC` arm no longer runs
+   `edit_delete_all`. With the sticky prefix a lone ESC from an interactive
+   keypress never reaches the edit loop (the decoder blocks for the next key),
+   so this arm now only fires if the input stream ends right after an ESC byte;
+   deleting the buffer there was the kaappi#2562 data-loss path. The
+   empty-input `break` is kept.
+
+**Cost, and it is deliberate:** a lone Escape no longer clears the input.
```
baijum, line 367: **[P2] The F1 help overlay still advertises the removed behaviour.** `src/editline_help.c:54` lists `esc — "delete the current input, or done with empty input"` and the quick-reference diagram at `editline_help.c:129` has `esc : delete input, done`. After this patch that key no longer deletes anything — pressing F1 right after this change ships tells the user about a destructive shortcut that no longer exists.

Since the help table is part of the vendored copy (Patch 3 already touched `editline_help.c`), please fold the two-line correction into Patch 7: drop or reword the `esc` row (e.g. `esc : Meta prefix for alt-<key>`), and update the diagram entry. The `^u` / `^c` rows already document the surviving ways to clear input, so nothing is lost.

baijum (commented):
Reviewed the C logic in context and verified the test both ways. The core change is correct; the asks below are documentation/help-text consistency (P2) plus two P3 notes — nothing wrong in the logic.

**What I verified:**

- **Blocking semantics of `-1`:** POSIX `tty_readc_noblock` maps `timeout_ms < 0` to `tty_readc_blocking` (plain `read`), and on Windows `tty_waitc_console(-1)` skips the peek and blocks in `ReadConsoleInputW` — so the sticky prefix behaves the same on both platforms, as PATCHES.md claims.
- **Decoder paths:** with `-1`, a lone ESC only surfaces as `KEY_ESC` on EOF; `ESC ESC` alt-cursor handling, CSI/SS3/OSC-after-ESC, and mouse bursts still decode immediately (the blocking read gets `[` at once). A mouse click while the prefix is pending composes cleanly.
- **Binding codes match:** the sticky path produces `key_unicode(peek) | KEY_MOD_ALT`, and the structural keys are registered as `WITH_ALT('S')`/`('B')`/`('R')`/`('y')` on the raw byte — exact match, no shift-modifier mismatch.
- **No new blocking kind:** the edit loop's main read (`tty_read`, `tty_read_timeout(…, -1, …)`) already blocks between keys, so waiting inside `tty_read_esc` for the next key is not a new kind of block.
- **Test both ways:** with the fix, `tests/scheme/smoke/repl-structural-editing-2216.sh` passes. With the two vendor files reverted to `main` and rebuilt, it fails exactly as described — the form is wiped and `S` evaluates to `undefined variable 'S'` (KP3001). The 0.4s pause is a clean way to pin the no-timeout property specifically.

**Summary of asks:**

| Pri | File | Issue |
|-----|------|-------|
| P2 | `vendor/isocline/src/editline_help.c:54,129` | F1 help still advertises "esc: delete the current input" |
| P2 | `docs/dev/repl.md:56-59` | Says the keys "will not deliver" without a Meta terminal — now the exact opposite |
| P3 | `AGENTS.md:456` | "six changes" enumeration is stale with Patch 7 |
| P3 | `vendor/isocline/PATCHES.md` | One line on `ic_set_tty_esc_delay` no longer affecting interactive decode |
| P3 | `vendor/isocline/PATCHES.md` | SIGWINCH-drops-pending-prefix wrinkle worth recording |

vendor/isocline/PATCHES.md
```diff
@@ -271,6 +271,58 @@ never compiles isocline (`use_isocline = !is_wasm_target` in `build.zig`), so
 no WASI guard is needed. Fixes kaappi#2423; found by the KEP-0022 CLOEXEC
 audit (kaappi#2414).
 
+## Patch 7 — ESC is a sticky Meta prefix (reach alt-<key> without a Meta terminal)
+
+**Files:** `src/tty.c`, `src/editline.c`
```
baijum: **[P2] `docs/dev/repl.md:56-59` now states the opposite of the new behaviour.** It reads:

> Each is `ESC` followed by the character, so a terminal that does not send Option as Meta (macOS Terminal.app, unless "Use Option as Meta key" is on) will not deliver them.

With the sticky prefix this is inverted: `ESC`-then-character is now the *canonical* way to reach them, works with no timeout on every terminal, and a Meta-configured terminal gets the same compose from the burst. That paragraph should be rewritten, and the deliberate behaviour change (a lone Escape no longer clears the input; `^u`/`^c` still do) is worth a sentence there too — it's the one user-visible regression-risk existing users will notice.

vendor/isocline/PATCHES.md
```diff
@@ -271,6 +271,58 @@ never compiles isocline (`use_isocline = !is_wasm_target` in `build.zig`), so
 no WASI guard is needed. Fixes kaappi#2423; found by the KEP-0022 CLOEXEC
 audit (kaappi#2414).
 
+## Patch 7 — ESC is a sticky Meta prefix (reach alt-<key> without a Meta terminal)
```
baijum, line 315: **[P3] `AGENTS.md:456` still enumerates "six changes"** in the Dependencies section, with the list of what each patch does. It's the orientation map that points people at PATCHES.md, so landing Patch 7 without updating it leaves it counting wrong. Bump to seven and add the sticky-Meta entry to the list ("`ESC` as a sticky Meta prefix so alt-<key> bindings work without a Meta-configured terminal (kaappi#2562)").

vendor/isocline/src/tty.c
```diff
@@ -488,11 +496,14 @@ ic_private tty_t* tty_new(alloc_t* mem, int fd_in)
   tty_t* tty = mem_zalloc_tp(mem, tty_t);
   tty->mem = mem;
   tty->fd_in = (fd_in < 0 ? STDIN_FILENO : fd_in);
-  #if defined(__APPLE__)
-  tty->esc_initial_timeout = 200;  // apple use ESC+<key> for alt-<key>
-  #else
+  // KAAPPI PATCH 7: one platform-independent value. The interactive ESC decode
```
baijum, line 499: **[P3] `ic_set_tty_esc_delay` is now quietly query-readers-only.** The public setter (isocline.h:446 → `tty_set_esc_delay` at tty.c:536) writes this field, and with the interactive decode hardcoded to `-1` its `initial_delay_ms` no longer affects ESC compose at all — only `tty_read_esc_response` / `tty_read_dsr_response`. Kaappi itself never calls it, so there's no functional impact here, but a one-line note in PATCHES.md item 2 ("the field — and `ic_set_tty_esc_delay`'s initial parameter — now only bounds the query-response readers") would save an embedder or a future upstream-merge from puzzling over why the knob does nothing.

vendor/isocline/src/tty.c
```diff
@@ -164,7 +164,15 @@ ic_private bool tty_read_timeout(tty_t* tty, long timeout_ms, code_t* code)
 
   if (c == KEY_ESC) {
     // escape sequence?
-    *code = tty_read_esc(tty, tty->esc_initial_timeout, tty->esc_timeout);
+    // KAAPPI PATCH 7: ESC is a sticky Meta prefix (see PATCHES.md) — pass -1 so
```
baijum, line 167: **[P3, informational] A signal while the prefix is pending drops the prefix.** `tty_readc_blocking` swallows `EINTR` (SIGWINCH on resize) and returns false, so `tty_read_esc` returns `KEY_ESC` and the edit loop handles the resize — but the user's pending sticky prefix is silently abandoned; their next key arrives un-prefixed. Upstream had the identical race inside its 100-200ms window, and the no-timeout window makes it marginally more likely; the consequence is trivial (press ESC again) and the old code's failure mode in this exact race was *wiping the input*, so this is strictly an improvement. No code change requested — just worth a line in PATCHES.md so the behaviour is on record if anyone ever chases "my ESC-S sometimes types a plain S after a resize".

tests/scheme/smoke/repl-structural-editing-2216.sh
```diff
@@ -173,6 +181,35 @@ for typed, lefts, key, expect, reject in CASES:
     if not ok or (reject is not None and reject in produced):
         failures.append((key, expect, produced))
 
+# kaappi#2562: ESC as a sticky Meta prefix, driven the way a Mac user actually
```
baijum, line 184: Verified this is a real regression test: reverting the two vendor files to `main` and rebuilding makes it fail exactly as the comment predicts — the form is wiped and `S` comes back as `undefined variable 'S'` — and it passes with the fix. Reusing the slurp case's form/cursor so the only variable is key delivery, and the `undefined variable` negative guard, are both nice touches. No changes needed.

vendor/isocline/src/editline_help.c
```diff
@@ -51,7 +51,8 @@ static const char* help[] = {
   "alt-d",      "delete to the end of the current word",
   "^u",         "delete to the start of the current line",
   "^k",         "delete to the end of the current line",
-  "esc",        "delete the current input, or done with empty input",
+  // KAAPPI PATCH 7: esc is a sticky Meta prefix now, not delete-input (see PATCHES.md)
+  "esc",        "Meta prefix for alt-<key>, or done with empty input",
```
baijum: **[P3, non-blocking] The "or done with empty input" tail no longer describes anything a keystroke can do.** With the sticky decoder a lone `KEY_ESC` never reaches the edit loop from a keypress — plain ESC just waits for the next key, and ESC-then-Enter composes alt-Enter, which `modify_code` turns into `KEY_LINEFEED` (inserts a newline rather than submitting). The empty-input `break` in the `KEY_ESC` arm now fires only on EOF-after-ESC or the SIGWINCH race now on record in PATCHES.md. As written, the row invites a user to press ESC on an empty line to finish — which now visibly does nothing. Consider trimming to just `"Meta prefix for alt-<key>"` (the EOF behaviour is an implementation detail, not a documented affordance).

docs/dev/repl.md
```diff
@@ -54,9 +54,11 @@ cursor:
 | alt+y | rotate | `(if a\| b c)` → `(if b c a\|)` |
 
 They are listed under F1 too, and only when the callback is set. Each is
-`ESC` followed by the character, so a terminal that does not send Option as
-Meta (macOS Terminal.app, unless "Use Option as Meta key" is on) will not
-deliver them.
+reached by pressing `ESC` then the character. Since patch 7 (kaappi#2562) `ESC`
+is a sticky Meta prefix with no timeout, so `ESC` then `S` composes to
```
baijum: **[P3, non-blocking] "`ESC` then `S` composes to `alt-shift-S`" holds only for capital S.** The bindings are `WITH_ALT('S')`/`('B')`/`('R')` on the *capital* byte (only rotate is lowercase `y`); `ESC` then lowercase `s` composes to alt-s, which is unbound, so nothing happens at all. Since this paragraph exists to un-confuse key delivery, consider "`ESC` then `shift-S`" so nobody presses lowercase s and concludes the feature is broken. The `alt+shift+S` table row above already implies the shift, so this is just closing the gap.

baijum (commented):
All five asks from my previous review are addressed, verified on this commit — built, ran `tests/scheme/smoke/repl-structural-editing-2216.sh` (PASS), and re-checked the patch markers (Patch 7 now marked at all three sites: `tty.c` ×2, `editline.c` ×1, `editline_help.c` ×2; all seven patches present, matching the "seven changes" text).

| Ask | Where | Checked |
|-----|-------|---------|
| F1 help overlay (P2) | `editline_help.c:55,131` reworded + marked | yes — both the table row and the quick-reference diagram |
| `docs/dev/repl.md` inverted paragraph (P2) | rewritten: ESC-then-char is now the canonical route, no timeout, plus the lone-ESC-doesn't-clear note | yes |
| `AGENTS.md` "six changes" (P3) | fixed in `CLAUDE.md` — correct, since `AGENTS.md` is a symlink to it | yes |
| `ic_set_tty_esc_delay` note (P3) | PATCHES.md item 2, accurate | yes |
| SIGWINCH race on record (P3) | PATCHES.md "One race, on record" paragraph, matches the analysis | yes |

Also good: the patch 7 **Files:** line picked up `src/editline_help.c`.

Two non-blocking wording nits inline — both about text that still promises something a keystroke can no longer do. Neither needs to hold the merge — from my side this is ready to merge; GitHub just won't let me record an approval on my own PR, so this is formally a comment.

PR #2576, opened 2026-09-11
Run every sandbox script on the POSIX CI legs

### LINKED ISSUE #2575: tests/scheme/sandbox/srfi181-sandbox.sh runs on no POSIX CI leg
## What

`tests/scheme/sandbox/srfi181-sandbox.sh` (added in #1732, 2026-07-23) is executed by **no POSIX CI step** and not by `tests/scheme/run-all.sh`. Only the two Windows legs run it, and only because their shell-suite loop iterates over every `tests/scheme/sandbox/*.sh`:

```yaml
for dir in smoke errors compile test-runner pipeline doctor fmt cache timings \
           completions lsp thottam differential sandbox robustness; do
  for s in tests/scheme/$dir/*.sh; do
```

The `test` job on Ubuntu x86, Ubuntu ARM, and macOS instead names the sandbox scripts one by one, and the list stops at three:

```yaml
- name: Sandbox escape tests
  run: bash tests/scheme/sandbox/sandbox-escape.sh
- name: Parallel pool sandbox degradation tests
  run: bash tests/scheme/sandbox/parallel-degrades.sh
- name: Sysinfo sandbox degradation tests
  run: bash tests/scheme/sandbox/sysinfo-degrades.sh
```

`run-all.sh` leaves `sandbox/` to CI on purpose (`tests/scheme/CLAUDE.md`), so there is no second net. The script has run on Linux/macOS exactly as often as someone invoked it by hand since it landed.

## Why it matters

The script is the only check that the embedded copy of `(srfi 181)` stays importable under `--sandbox` — `--sandbox` blocks every file-backed `.sld` load, so the library works there solely because `vm_library.zig`'s `embedded_libraries` table carries its source. Dropping it from that table (or breaking the embedded copy) would fail on Windows only, and the Windows legs are the ones most likely to be read as "some Windows thing" and retried.

The same shape bit `(kaappi parallel)` and `(kaappi sysinfo)`, which is why those two scripts have their own steps. The next sandbox script will be orphaned the same way unless the pattern changes.

## Reproduction

Passes locally against `main`, so the fix is wiring only:

```console
$ zig build && bash tests/scheme/sandbox/srfi181-sandbox.sh
=== Results ===
Passed: 8
Failed: 0
(srfi 181) is fully available under --sandbox.
```

Confirm the gap:

```console
$ grep -rn srfi181-sandbox .github/workflows/ tests/scheme/run-all.sh
$
```

## Fix

Either of:

1. Add a fourth named step beside the other three in the `test` job (smallest change, keeps the per-script step names in the log).
2. Replace the three named steps with the same `for s in tests/scheme/sandbox/*.sh` loop the Windows legs use, so a new sandbox script can never be orphaned again. Keep SKIP (77) handling as the Windows loop does.

Option 2 is the one that fixes the class; option 1 is the one-line patch.

## Priority

`priority: low` per `docs/dev/github-issues.md`: nothing a user runs behaves wrongly, and the script itself passes. What is wrong is the coverage claim — "CI runs the sandbox suite" — which #2574 corrected in `docs/dev/testing.md` while this stays open.

`kaappi --version`: built from `main` at 3acb96a8 (v0.27.1 line).

baijum:
## Summary

Fixes #2575 (option 2 from the issue — the one that fixes the class, not just the instance).

The `test` job named its sandbox scripts one by one and the list stopped at three, so `srfi181-sandbox.sh` (#1732) — the only check that the embedded copy of `(srfi 181)` stays importable under `--sandbox` — never ran on any POSIX leg; only the Windows legs ran it, because they glob the whole `tests/scheme/sandbox/` directory. `run-all.sh` leaves `sandbox/` to CI on purpose, so there was no second net.

This replaces the three named steps in the `test` job with the same `for s in tests/scheme/sandbox/*.sh` loop the Windows "Shell suites" step uses, keeping its exit-77 (SKIP) handling, so the next sandbox script joins the POSIX legs without a workflow edit instead of being orphaned the same way. `docs/dev/testing.md` gains a sentence documenting the loop.

One visible behavior change: script output now prints only on failure (a `PASS/SKIP/FAIL` line per script otherwise), matching the Windows loop; successful runs get more compact logs.

## Verification

- Ran the exact loop body under `bash -e` against a current build — all four scripts pass, including the previously orphaned `srfi181-sandbox.sh`:
  ```
  PASS  tests/scheme/sandbox/parallel-degrades.sh
  PASS  tests/scheme/sandbox/sandbox-escape.sh
  PASS  tests/scheme/sandbox/srfi181-sandbox.sh
  PASS  tests/scheme/sandbox/sysinfo-degrades.sh
  ```
- Exercised the SKIP (77) and FAIL paths with stub scripts: a 77 exit is reported as SKIP and does not abort the step under `bash -e`; a failing script prints its output and the step exits 1.
- `npx markdownlint-cli2`: 0 issues. `zig fmt --check src/`: clean.

## Checklist

- [ ] `zig build test` passes — not run; this PR touches no `.zig` source
- [ ] Scheme test suites pass (`bash tests/scheme/run-all.sh`) — not run; the sandbox scripts it does not run are exactly what this PR wires into CI, and all four pass locally per above
- [ ] New tests added for new behavior — no new behavior; the fix wires an existing script into CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -383,17 +383,34 @@ jobs:
           # other script.
           KAAPPI_SHELL_TEST_TIMEOUT: ${{ matrix.optimize == 'Debug' && '900' || '600' }}
 
-      - name: Sandbox escape tests
+      - name: Sandbox tests
+        # Every script in the directory, not a hand-kept list of step names:
+        # the named steps this replaced stopped at three and left
+        # srfi181-sandbox.sh (#1732) unrun on every POSIX leg (kaappi#2575) —
+        # only the Windows legs, which glob the directory, ever ran it. A glob
+        # cannot orphan the next sandbox script the way that list orphaned
+        # this one. Exit 77 = SKIP (tests/scheme/shell-common.sh), handled
+        # like the Windows "Shell suites" loop.
         if: env.DOCS_ONLY != 'true'
-        run: bash tests/scheme/sandbox/sandbox-escape.sh
-
-      - name: Parallel pool sandbox degradation tests
-        if: env.DOCS_ONLY != 'true'
-        run: bash tests/scheme/sandbox/parallel-degrades.sh
-
-      - name: Sysinfo sandbox degradation tests
-        if: env.DOCS_ONLY != 'true'
-        run: bash tests/scheme/sandbox/sysinfo-degrades.sh
+        run: |
+          fail=0
+          for s in tests/scheme/sandbox/*.sh; do
+            [ -e "$s" ] || continue
+            # `|| status=$?`: the runner's bash -e would otherwise abort the
+            # whole step on the first SKIP (77).
+            status=0
+            out=$(bash "$s" 2>&1) || status=$?
+            if [ "$status" -eq 0 ]; then
+              echo "  PASS  $s"
+            elif [ "$status" -eq 77 ]; then
+              echo "  SKIP  $s"
+            else
+              echo "  FAIL  $s (exit $status)"
+              printf '%s\n' "$out"
+              fail=1
+            fi
+          done
+          exit $fail
 
       - name: Robustness tests
         if: env.DOCS_ONLY != 'true'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -384,37 +384,26 @@ jobs:
           KAAPPI_SHELL_TEST_TIMEOUT: ${{ matrix.optimize == 'Debug' && '900' || '600' }}
 
       - name: Sandbox tests
-        # Every script in the directory, not a hand-kept list of step names:
-        # the named steps this replaced stopped at three and left
-        # srfi181-sandbox.sh (#1732) unrun on every POSIX leg (kaappi#2575) —
-        # only the Windows legs, which glob the directory, ever ran it. A glob
-        # cannot orphan the next sandbox script the way that list orphaned
-        # this one. Exit 77 = SKIP (tests/scheme/shell-common.sh), handled
-        # like the Windows "Shell suites" loop.
+        # Every script in the directory via the shared driver, not a
+        # hand-kept list of step names: the named steps this replaced stopped
+        # at three and left srfi181-sandbox.sh (#1732) unrun on every POSIX
+        # leg (kaappi#2575). tools/run-shell-dir.sh is the same driver the
+        # Windows "Shell suites" steps call, so the two platforms' handling
+        # of SKIP (77), timeouts, and transcripts cannot drift apart again.
         if: env.DOCS_ONLY != 'true'
         run: |
-          fail=0
-          for s in tests/scheme/sandbox/*.sh; do
-            [ -e "$s" ] || continue
-            # `|| status=$?`: the runner's bash -e would otherwise abort the
-            # whole step on the first SKIP (77).
-            status=0
-            out=$(bash "$s" 2>&1) || status=$?
-            if [ "$status" -eq 0 ]; then
-              echo "  PASS  $s"
-            elif [ "$status" -eq 77 ]; then
-              echo "  SKIP  $s"
-            else
-              echo "  FAIL  $s (exit $status)"
-              printf '%s\n' "$out"
-              fail=1
-            fi
-          done
-          exit $fail
+          export KAAPPI_HOME="$(mktemp -d)"
+          bash tools/run-shell-dir.sh tests/scheme/sandbox
 
       - name: Robustness tests
+        # Same driver as the sandbox step. robustness/ held the identical
+        # hazard — one hand-listed script path where the Windows legs glob
+        # the whole directory — so a second script there would have run on
+        # Windows and never on Linux or macOS (kaappi#2575 review).
         if: env.DOCS_ONLY != 'true'
-        run: bash tests/scheme/robustness/robustness.sh
+        run: |
+          export KAAPPI_HOME="$(mktemp -d)"
+          bash tools/run-shell-dir.sh tests/scheme/robustness
 
       - name: E2E tests (LLVM native backend)
         if: matrix.optimize == 'ReleaseSafe' && env.DOCS_ONLY != 'true'
@@ -1323,25 +1312,17 @@ jobs:
           fi
           export KAAPPI=bin/kaappi.exe
           export KAAPPI_HOME="$RUNNER_TEMP/kaappi-test-home"
+          # Headroom over the shared driver's 600s per-script default:
+          # run-differential.sh is the expensive one at ~9 min (above).
+          export KAAPPI_SHELL_TEST_TIMEOUT=1200
           fail=0
           for dir in smoke errors compile test-runner pipeline doctor fmt cache timings \
                      completions lsp thottam differential sandbox robustness; do
-            for s in tests/scheme/$dir/*.sh; do
-              [ -e "$s" ] || continue
-              # `|| status=$?`: the runner's `shell: bash` implies -e, which
-              # would otherwise abort the whole step on the first SKIP (77).
-              status=0
-              out=$(KAAPPI="$KAAPPI" bash "$s" "$KAAPPI" 2>&1) || status=$?
-              if [ "$status" -eq 0 ]; then
-                echo "  PASS  $s"
-              elif [ "$status" -eq 77 ]; then
-                echo "  SKIP  $s"
-              else
-                echo "  FAIL  $s (exit $status)"
-                printf '%s\n' "$out"
-                fail=1
-              fi
-            done
+            # tools/run-shell-dir.sh is the same driver the POSIX `test`
+            # legs call for sandbox/ and robustness/, so the two platforms'
+            # SKIP (77), timeout, and transcript handling cannot drift
+            # apart again (kaappi#2575).
+            bash tools/run-shell-dir.sh "tests/scheme/$dir" "$KAAPPI" || fail=1
           done
           exit $fail
 
@@ -1491,25 +1472,17 @@ jobs:
           fi
           export KAAPPI=bin/kaappi.exe
           export KAAPPI_HOME="$RUNNER_TEMP/kaappi-test-home"
+          # Headroom over the shared driver's 600s per-script default:
+          # run-differential.sh is the expensive one at ~9 min (above).
+          export KAAPPI_SHELL_TEST_TIMEOUT=1200
           fail=0
           for dir in smoke errors compile test-runner pipeline doctor fmt cache timings \
                      completions lsp thottam differential sandbox robustness; do
-            for s in tests/scheme/$dir/*.sh; do
-              [ -e "$s" ] || continue
-              # `|| status=$?`: the runner's `shell: bash` implies -e, which
-              # would otherwise abort the whole step on the first SKIP (77).
-              status=0
-              out=$(KAAPPI="$KAAPPI" bash "$s" "$KAAPPI" 2>&1) || status=$?
-              if [ "$status" -eq 0 ]; then
-                echo "  PASS  $s"
-              elif [ "$status" -eq 77 ]; then
-                echo "  SKIP  $s"
-              else
-                echo "  FAIL  $s (exit $status)"
-                printf '%s\n' "$out"
-                fail=1
-              fi
-            done
+            # tools/run-shell-dir.sh is the same driver the POSIX `test`
+            # legs call for sandbox/ and robustness/, so the two platforms'
+            # SKIP (77), timeout, and transcript handling cannot drift
+            # apart again (kaappi#2575).
+            bash tools/run-shell-dir.sh "tests/scheme/$dir" "$KAAPPI" || fail=1
           done
           exit $fail
 
--- a/docs/dev/testing.md
+++ b/docs/dev/testing.md
@@ -551,6 +551,10 @@ hash tables). Uses `assert_blocked` and `assert_works` helpers.
 bash tests/scheme/sandbox/sandbox-escape.sh
 ```
 
+CI's `test` job runs every `tests/scheme/sandbox/*.sh` script in one loop
+step (kaappi#2575) — the same glob the Windows legs use, so a new sandbox
+script joins the POSIX legs without a workflow edit.
+
 ### Error format (`tests/scheme/errors/error-format.sh`)
 
 One of the `errors/` suite's scripts, so `run-all.sh` already covers it.
--- a/docs/dev/testing.md
+++ b/docs/dev/testing.md
@@ -235,8 +235,8 @@ tests/scheme/
   differential/     Execution-tier differential (opt-off, warm cache, WASM)
                     + probes/ (its own small corpus)
 
-  # not run by run-all.sh
-  robustness/       Malformed and adversarial input handling (its own .sh)
+  # not run by run-all.sh — CI drives both through tools/run-shell-dir.sh
+  robustness/       Malformed and adversarial input handling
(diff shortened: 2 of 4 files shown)


baijum (commented):
Reviewed the loop against the Windows precedent, `run-all.sh`, and the four sandbox scripts. The fix does what #2575 asked; the inline comments are about the loop itself. One more that is outside the diff hunk so GitHub won't let me anchor it:

**`robustness/` is left with the identical hazard.** The step right below this one still runs `bash tests/scheme/robustness/robustness.sh` as a single hand-listed path, while both Windows loops glob `robustness`. A second script under `tests/scheme/robustness/` would run on Windows and never on Linux or macOS, which is the state `sandbox/` was in until #1732. The step comment's "a glob cannot orphan the next sandbox script" reasoning applies there too; worth doing in the same PR so the fix is structural rather than directory-specific.

The empty-glob comment is the one I'd treat as blocking, since it reopens the class the PR closes.

.github/workflows/ci.yml
```diff
@@ -383,17 +383,34 @@ jobs:
...
-        if: env.DOCS_ONLY != 'true'
-        run: bash tests/scheme/sandbox/parallel-degrades.sh
-
-      - name: Sysinfo sandbox degradation tests
-        if: env.DOCS_ONLY != 'true'
-        run: bash tests/scheme/sandbox/sysinfo-degrades.sh
+        run: |
+          fail=0
+          for s in tests/scheme/sandbox/*.sh; do
+            [ -e "$s" ] || continue
```
baijum: **Empty glob exits 0 having run nothing.** If the glob matches no file, `$s` is the literal pattern, `[ -e ]` is false, the loop body never runs, and the step exits 0 with no PASS/SKIP/FAIL line. Reproduced with an empty `tests/scheme/sandbox/` under `bash -e`: no output, exit 0.

That is the same silent-orphan class this PR fixes, just one level up (move or rename the scripts without a CI edit and every POSIX leg stays green). `gc-stress-scheme` in this file already guards its list the other way ("The script FAILS if a name here matches no file"). Suggest a counter:

```bash
ran=0
for s in tests/scheme/sandbox/*.sh; do
  [ -e "$s" ] || continue
  ran=$((ran + 1))
  ...
done
[ "$ran" -gt 0 ] || { echo "no sandbox scripts found"; exit 1; }
exit $fail
```

or `shopt -s failglob` at the top of the step.

.github/workflows/ci.yml
```diff
@@ -383,17 +383,34 @@ jobs:
...
+            # `|| status=$?`: the runner's bash -e would otherwise abort the
+            # whole step on the first SKIP (77).
+            status=0
+            out=$(bash "$s" 2>&1) || status=$?
```
baijum: **No per-script timeout, and a hang logs nothing.** The `run-all.sh` step directly above sets `KAAPPI_SHELL_TEST_TIMEOUT`, which `run_shell_worker` honours via `wait_with_timeout`; this loop runs each script bare. And because output is captured into `$out`, a hung script leaves the step log empty until the job-level `timeout-minutes` kills the whole job. Before, the named step streamed the last `PASS:` line, which named the assertion in flight.

This script is a realistic hang vector: `sandbox-escape.sh` tests `--timeout 200` on `(let loop () (loop))`, so a regression in timeout enforcement hangs exactly here. `timeout "${KAAPPI_SHELL_TEST_TIMEOUT:-600}" bash "$s"` with a `TIMEOUT` verdict on exit 124 would restore both a bound and a name.

.github/workflows/ci.yml
```diff
@@ -383,17 +383,34 @@ jobs:
...
+            if [ "$status" -eq 0 ]; then
+              echo "  PASS  $s"
```
baijum: **Passing scripts' transcripts are dropped.** On status 0 only `PASS <path>` is echoed; `$out` is printed on FAIL alone. Each script prints ~35 per-assertion `PASS:` lines and a `Passed: N / Failed: 0` block, all of which used to be in the step log and now are not. A reviewer checking that an assertion ran for the right reason (e.g. the positive-control cache write) has to rerun locally.

Cheap to keep, collapsed:

```bash
echo "::group::  PASS  $s"; printf '%s\n' "$out"; echo "::endgroup::"
```

.github/workflows/ci.yml
```diff
@@ -383,17 +383,34 @@ jobs:
           # other script.
           KAAPPI_SHELL_TEST_TIMEOUT: ${{ matrix.optimize == 'Debug' && '900' || '600' }}
 
-      - name: Sandbox escape tests
+      - name: Sandbox tests
```
baijum, line 386: **One step name loses failure localisation.** Three named steps collapse into "Sandbox tests", so the Checks UI and `gh run view` no longer say which script failed; you have to open the log for the `FAIL tests/scheme/sandbox/<x>.sh` line. An annotation on the FAIL branch gets the name back into the Checks tab:

```bash
echo "::error file=$s::sandbox script failed (exit $status)"
```

.github/workflows/ci.yml
```diff
@@ -383,17 +383,34 @@ jobs:
...
-      - name: Parallel pool sandbox degradation tests
-        if: env.DOCS_ONLY != 'true'
-        run: bash tests/scheme/sandbox/parallel-degrades.sh
-
-      - name: Sysinfo sandbox degradation tests
-        if: env.DOCS_ONLY != 'true'
-        run: bash tests/scheme/sandbox/sysinfo-degrades.sh
+        run: |
+          fail=0
+          for s in tests/scheme/sandbox/*.sh; do
```
baijum: **Third hand-copied loop, already diverging.** This is the same glob/SKIP-77/capture loop the two Windows jobs carry, and it already differs from them (no `KAAPPI=` / positional arg, no `KAAPPI_HOME` isolation). The drift that produced #2575 came from exactly this shape. `run_shell_suite` in `tests/scheme/run-all.sh` already implements SKIP, per-script timeouts, and parallel dispatch, and neither `run-all.sh` nor `tests/scheme/CLAUDE.md` records a reason `sandbox/` is kept out of it ("CI runs it separately" states a fact, not a rationale).

Two options that would also address the timeout and output comments above: a small `tools/run-shell-dir.sh <dir> [kaappi]` called from here and from both Windows loops, or wiring `sandbox/` into `run-all.sh` via `run_shell_suite` so it is reproducible locally in one command.

.github/workflows/ci.yml
baijum: Fixed with the counter shape, in the shared driver this PR now uses: `tools/run-shell-dir.sh` collects the glob first and exits 1 with `FAIL no \*.sh found in <dir>` when nothing matched — verified with an empty directory (message printed, exit 1). This also covers the Windows steps, whose `[ -e ] || continue` had the identical silent pass.

.github/workflows/ci.yml
baijum: Addressed by borrowing run-all.sh's machinery instead of GNU `timeout`, which neither the macOS legs nor Git Bash ship: the tick-counted `wait_with_timeout` plus the `set -m` process-group kill (kaappi#2434), so a hung script's child interpreter dies with it. Budget is `KAAPPI_SHELL_TEST_TIMEOUT` (600s default; the Windows steps export 1200s for run-differential.sh's ~9 min). A kill reports `TIMEOUT <script> (killed after Ns)` with the partial transcript and a `::error` annotation, so the log names the stuck script. Verified locally with a stub sleeping past a 2s budget — killed, verdict line, exit 1 — including that its backgrounded `sleep 30` child died with the group.

.github/workflows/ci.yml
baijum: Kept, collapsed as suggested: on PASS the full transcript now prints inside `::group::`/`::endgroup::` under CI (raw when run locally, where the markers would be noise) — the per-assertion lines and the Passed/Failed block are back in the step log, positive-control cache write included.

.github/workflows/ci.yml
baijum: Added on the FAIL branch — `::error file=<script>::<script> failed (exit N)` — and the same on TIMEOUT, so the Checks tab and `gh run view` name the script again without opening the log.

.github/workflows/ci.yml
baijum: Went with the `tools/run-shell-dir.sh` option rather than wiring `sandbox/` into run-all.sh: the Windows legs can't share run-all.sh, so that route would have fixed the POSIX copy and left the two Windows copies as the drifted ones — and they are the legs that needed the timeout and transcript fixes anyway. The tool is now the single implementation all three call sites use (the POSIX sandbox and robustness steps, and both Windows Shell-suites loops, which call it once per directory); the KAAPPI env + positional-arg shape and KAAPPI_HOME isolation are aligned, with the POSIX steps isolating too. The rationale you noted was missing — why these directories stay out of run-all.sh — is now recorded in `tests/scheme/CLAUDE.md` next to the table.

baijum:
All review comments are addressed in 7f636cc1.

**The `robustness/` point from the review summary:** done in the same PR — its step now runs `bash tools/run-shell-dir.sh tests/scheme/robustness`, so the one-hand-listed-path hazard is closed the same way `sandbox/` was, and a second script in either directory joins every leg without a workflow edit.

**Verification (all local, against a current build):**
- All four `sandbox/` scripts and `robustness.sh` pass through the driver, with an isolated `KAAPPI_HOME`, in both the POSIX shape (no argument, scripts' default binary) and the Windows shape (binary passed as argument and `KAAPPI` env).
- Empty directory → `FAIL no \*.sh found`, exit 1; exit 77 → `SKIP`; exit 3 → `FAIL` with transcript and annotation; a script sleeping past its budget → `TIMEOUT` verdict, killed **with its process group** (verified its backgrounded `sleep 30` child died too).
- `CI=1` emits the `::group::`/`::error` workflow commands; without it the output stays marker-free for local runs.
- Workflow YAML parses; `markdownlint-cli2` clean.

One deliberate divergence from `run_shell_suite` worth flagging: the driver invokes scripts as `bash <script>` rather than gating on the executable bit, because two of the four sandbox scripts (`parallel-degrades.sh`, `srfi181-sandbox.sh`) are not `+x`, and both the old named steps and the Windows loops ran them that way — an exec-bit gate would have converted both to NOTEXEC.

baijum (commented):
**LGTM.** The new commit resolves all six points from the first review, and it does so at the right depth: one driver for every leg instead of a fourth loop.

What I verified on `7f636cc1`:

- All seven jobs that execute the driver are green on this head: the five `test` legs (including Debug, where the new 600 s per-script budget was the one thing that could have bitten) and both Windows legs.
- The Windows x64 log shows all four sandbox scripts and `robustness.sh` running through the driver, 88 collapsed `::group::` transcripts rendered, and no job-control noise from `set -m` under Git Bash — which was the open question for that approach.
- Locally, under macOS `/bin/bash` 3.2 and bash 5.3: empty directory → exit 1; a `77` script → SKIP; a hanging script with a backgrounded child → TIMEOUT with the child killed; a failing script → transcript plus `::error` under `CI=true`. `shellcheck` clean. The four real sandbox scripts pass with an isolated `KAAPPI_HOME`.
- The kill sequence (TERM to the process group, then `wait`) matches `run-all.sh`'s `run_shell_worker` exactly, so it's no new hazard.

The four inline comments are non-blocking: a rationale in `tests/scheme/CLAUDE.md` that doesn't hold against its own table, a missing `tests/scheme/tools/` test for the driver, and two cosmetic nits. GitHub won't let me approve a PR you opened yourself, so this is the approval.

tests/scheme/CLAUDE.md
```diff
@@ -28,8 +28,16 @@
 | `tools/` | The repo's own tooling scripts (`tools/run-unit-test-chunk.sh` watchdog, kaappi#2560) | yes |
 | `differential/` | Execution-tier differential harnesses (`--no-ir-opt`, cold-vs-warm cache; WASM-vs-interpreter) + its `probes/` | yes |
 | `coverage/` | Coverage gap-fillers (`zig build coverage-scheme`) | no |
-| `robustness/` | Stress tests | no (CI runs it separately) |
-| `sandbox/` | Sandbox isolation tests | no (CI runs it separately) |
+| `robustness/` | Stress tests | no (CI: `tools/run-shell-dir.sh`) |
+| `sandbox/` | Sandbox isolation tests | no (CI: `tools/run-shell-dir.sh`) |
+
+Why `robustness/` and `sandbox/` are not in `run-all.sh`: their coverage must
```
baijum, line 34: Non-blocking. The reason given here doesn't actually distinguish these two directories: `errors/`, `compile/`, `test-runner/` and the rest are *also* traversed by the Windows "Shell suites" steps rather than through `run-all.sh`, and they are in `run-all.sh` too. So "the Windows legs cannot share run-all.sh" explains why the shared driver exists, not why `sandbox/` and `robustness/` stay out of `run-all.sh`.

Two honest ways to write it: state that the exclusion is historical (I could find no recorded reason), or add both directories to `run-all.sh` via `run_shell_suite` as well, so `bash tests/scheme/run-all.sh` covers them locally while CI keeps the driver. Either is fine for this PR; I'd just not leave a rationale that a reader can falsify against the table above it.

tools/run-shell-dir.sh
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env bash
```
baijum, line 1: Non-blocking, but worth a follow-up if not here. The driver now carries real logic (glob-empty → FAIL, 77 → SKIP, tick-counted timeout with a process-group kill, transcript grouping), and `tests/scheme/tools/` exists precisely for the repo's own tooling scripts (`unit-chunk-watchdog-2560.sh` is the precedent). A small test there with stub scripts would pin the four verdicts and the empty-directory failure so they can't regress silently.

For what it's worth, I ran exactly that by hand against this file under `/bin/bash` 3.2 and bash 5.3: empty dir exits 1 with the FAIL line; a `77` script prints SKIP; a hanging script with a `sleep 300 &` child is reported TIMEOUT and the child is gone afterwards; a failing script prints its transcript and the `::error` annotation under `CI=true`. `shellcheck -s bash` is clean. So the behaviour is right today; the test is to keep it that way.

tools/run-shell-dir.sh
```diff
@@ -0,0 +1,133 @@
...
+        echo "  TIMEOUT  $s  (killed after ${SHELL_TIMEOUT}s)"
+        cat "$out"
+        annotate "$s" "$s timed out after ${SHELL_TIMEOUT}s"
+        fail=1
+        continue
+    fi
+    case $status in
+        0)
+            echo "  PASS  $s"
+            pass_group_open "$s"
```
baijum, line 118: Nit. Under CI this prints the PASS verdict twice in a row: the plain `  PASS  $s` line, then the `::group::  PASS  $s` header the runner renders as a collapsible line with the same text. Skipping the plain echo when `CI` is set (or making `pass_group_open` the only PASS line and having the non-CI branch print it plainly) gives one line per script either way.

docs/dev/testing.md
```diff
@@ -235,8 +235,8 @@ tests/scheme/
   differential/     Execution-tier differential (opt-off, warm cache, WASM)
                     + probes/ (its own small corpus)
 
-  # not run by run-all.sh
-  robustness/       Malformed and adversarial input handling (its own .sh)
+  # not run by run-all.sh — CI drives both through tools/run-shell-dir.sh
```
baijum, line 238: Nit. "CI drives both" reads as referring to everything under this comment header, but `bench/` and `coverage/` two lines down are not driven by anything. Either move the `bench/ coverage/` line under its own header or say "CI drives `robustness/` and `sandbox/` through …".

PR #2597, opened 2026-09-14
Accept the descent window in watchdog test case 7; widen the riscv64 bisection pot

### LINKED ISSUE #2596: unit-chunk-watchdog-2560.sh case 7 flakes on the netbsd-test VM leg (depth-2 estimate)
## Summary

`tests/scheme/tools/unit-chunk-watchdog-2560.sh` case 7 failed once on the `netbsd-test` CI leg and passed on the next run of the same tree with no relevant change in between, so it is a timing flake on the slow VM, not a regression.

## Evidence

- Run 34850016237 (PR #2592, commit 0fcbeb8f), `netbsd-test`: `FAIL: case 7: depth 2's estimate is not 38s from its subset's test count`. Everything else in that leg passed (2146 pass, 1 fail).
- Run 34854442968 (same PR, commit 35e8f399 — the diff between the two touches nothing this test or `tools/run-unit-test-chunk.sh` use), `netbsd-test`: success.
- `main` at 1d7bd352 (2026-09-14, after #2561 landed the test on 2026-09-09): `netbsd-test` success. Most other recent `main` pushes were docs-only and skipped the leg, so the sample is small.

## Likely mechanism

Case 7 derives the expected depth-2 estimate on the assumption that depth 1 *completes* inside its time-bounded level (`KAAPPI_BISECT_LEVEL_PER_TEST_TENTHS=1`, the wedge-once stamp already consumed in phase 1), which "puts depth 2 just past it" (`mid1+1..mid2`). If depth 1 overruns its level on a loaded VM, the bisection narrows into the first half instead, depth 2 has a different filter count, and the printed estimate differs from `e2`. The `netbsd-test` leg runs the whole suite inside a KVM NetBSD guest on a shared runner, where `wait_with_timeout` ticks were already measured to stretch (kaappi#2555 notes ~36% on macOS; the BSD VMs are slower still).

## Possible fixes

- Have the test print the depth-1/depth-2 lines it actually saw (`grep 'bisect depth' "$out"`) on failure, so the next occurrence confirms or refutes the mechanism without a re-run.
- Make the case-7 assertion tolerant of a depth-1 overrun: accept either the "just past" window or the "first half again" window, deriving both estimates the same way, or raise the per-test tenths on the BSD legs so depth 1 reliably finishes.
- Or skip case 7 on the VM legs (`skip_on_*` in `shell-common.sh`), keeping the deterministic cases.

## Priority

`low`: the behaviour under test is correct and CI is loud about it; the cost is a false red on a ~50-minute leg, which a re-run clears.

baijum:
## Summary

Two related changes to the unit-chunk watchdog's instruments.

**Case 7 of `tests/scheme/tools/unit-chunk-watchdog-2560.sh` tolerates a depth-1 overrun and is diagnosable when it fails** (#2596). The case asserted that bisect depth 2 is the window just past depth 1, which only holds when depth 1 completes inside its allowance; on the `netbsd-test` KVM leg it once did not (run 34850016237), the chunk script read the overrun as a wedge and descended, and the case failed saying nothing about which path ran. Now:

- both depth-2 windows are derived with the script's own mid arithmetic (`lo=0, hi=count, mid=(lo+hi)/2`): the just-past window `mid1+1..mid2` when depth 1 completes, and `1..mid1/2` (the first half's first half) when it overruns and the script sets `hi=mid`;
- each window is matched by filter count, `first..last` names *and* estimate, and the descent window is accepted only behind the script's own `did NOT finish ... descending into this half` line for depth 1, so a flat or per-filter estimate still fails either way;
- every failure prints the `WALL BUDGET EXCEEDED` / `== bisect` lines actually seen (`fail_bisect`), so the next occurrence confirms or refutes the mechanism without a re-run;
- the case's pot grows 45s -> 90s: with 45s a real 44s depth-1 overrun left under 1s, so the script would have stopped "too shallow" and printed no depth-2 line at all. On the usual path every level completes in one poll tick, so the pot never binds and the printed allowances are unchanged;
- a new **case 8** forces the overrun (`SHIM_HANG_FIRST` seeding two hangs: the chunk run and depth 1) with the per-test term zeroed, so the descent branch of the shared matcher runs on every host at a one-poll-tick cost instead of the full 44s allowance. `run-all.sh`'s per-script budget for this test goes 480 -> 600 for the added case (the test measures 4m41s solo).

**`riscv64-test`'s `KAAPPI_BISECT_BUDGET` rises from 1080s (18m) to 1620s (27m)** (refs #2488). The first descent the pot funded (run 34863343107, rest chunk) went depth 1 (23 filters, wedged at its full 628s estimate) -> depth 2 (11, clean 130s) -> depth 3 (6, clean 175s) -> depth 4 (3, clean 106s), then had 37s left, under the 40s level floor, with three suspects unsplit (`tests_exceptions`, `tests_ffi`, `tests_gc_worklist`). Finishing that descent costs at most three more levels at their full estimates, `120 + 12*tests/10` each: 166s + 136s + 124s = 426s on top of the 1043s spent = 1469s, so 1620s funds it with ~150s to spare. Cap arithmetic redone per the header's own rule:

| | before | after |
|---|---|---|
| pot | 18m (1080s) | 27m (1620s) |
| per-step cap = wall budget + pot + 2m | rest 42, process 32, concurrency 37, io 32, fuzz 36, gc 36, native 32, tooling 32 | rest 51, process 41, concurrency 46, io 41, fuzz 45, gc 45, native 41, tooling 41 |
| job cap: setup+build 8m + chunks healthy 38m + last chunk's full cap | 8 + 38 + 32 = 78 inside 80 | 8 + 38 + 41 = 87 inside 90 |

s390x/ppc64le jobs untouched. YAML validated with PyYAML.

## Verification

- `bash tests/scheme/tools/unit-chunk-watchdog-2560.sh` under `/bin/bash` 3.2.57 and `/opt/homebrew/bin/bash` 5.3.15: both PASS, 4m41s each.
- A scratch variant of case 8 with the real per-test term (`KAAPPI_BISECT_LEVEL_PER_TEST_TENTHS=1`, 90s pot): depth 1 expired at its full 44s allowance, the script descended, and the matcher accepted the descent window (11 filters `bignum..repl_highlight`, 15s estimate) -- the exact path the netbsd flake takes.
- `fail_bisect` checked negatively (prints the seen lines, then `FAIL`; prints `(none)` on empty output).
- shellcheck is not installed on this host and CI does not run it; the new lines follow the file's existing `# shellcheck disable=SC2086` convention for the `env $base_env` word lists.

Closes #2596
Refs #2488

## Checklist

- [ ] `zig build test` passes -- n/a, no Zig source touched
- [ ] `zig fmt --check src/` passes -- n/a
- [ ] Scheme test suites pass (`bash tests/scheme/run-all.sh`) -- the one suite this touches was run under both bash versions (above); the full `run-all.sh` was not re-run, nothing outside `tests/scheme/tools/` and its `run-all.sh` budget entry changed
- [x] New tests added for new behavior (case 8, the forced depth-1 overrun)
- [x] Documentation updated (if applicable) -- the `riscv64-test` job's cap-arithmetic comment and the `run-all.sh` per-script budget comment
- [x] Commit body explains *why* -- release notes are written from it at


.github/workflows/ci.yml
```diff
@@ -644,27 +644,36 @@ jobs:
...
-    timeout-minutes: 80
+    timeout-minutes: 90
     env:
-      # Total bisection pot per chunk step, seconds (18m); see the comment
-      # above. The per-test bound used during bisection levels (90s) and
-      # the rest of the knobs are the chunk script's own defaults.
-      KAAPPI_BISECT_BUDGET: "1080"
+      # Total bisection pot per chunk step, seconds (27m); see the comment
+      # above. Sized from the first descent the pot itself cut short
+      # (kaappi#2488, run 34863343107, rest chunk): depth 1 ran to its full
```
baijum: This run ID does not contain the events cited after it. I pulled 34863343107's full log: its `riscv64-test` job ran the rest chunk **healthy** — heartbeats to 541s, then `Build Summary: 13/13 steps succeeded; 912/913 tests passed` and `== unit-test chunk 'rest': 912 tests (floor 7)` at 9m36s. No `WALL BUDGET EXCEEDED`, no `== bisect` lines anywhere in the run.

The descent is real, but it lives in the three cancelled 2026-09-09 runs **34363328716**, **34370359849** and **34379866388** (the #2561 landing-day pushes). What those logs actually say:

| comment says | 34363328716 | 34370359849 | 34379866388 |
|---|---|---|---|
| depth 1 wedged at its full 628s estimate | 628s, wedged ✓ | 628s, wedged ✓ | 628s, wedged ✓ |
| depths 2–4 completed in 130s, 175s, 106s | 125s, 150s, 95s | 125s, 170s, 105s | 115s, 135s, 91s |
| then had 37s left, under the 40s floor, three suspects unsplit | depth 5 ran **clipped** (78s of a 166s estimate) on `tests_exceptions` alone, timed out **INCONCLUSIVE**, "not descending", then `bisect incomplete: 3 suspect filter(s) remain` | same shape, 48s clip, 3 suspects | depth 5 **completed clean** at a 108s clip, then `remaining pot (13s) is too shallow` with **2** suspects (`tests_ffi`, `tests_gc_worklist`) |

(The three suspect names — tests_exceptions, tests_ffi, tests_gc_worklist — do match runs 1–2 exactly, so this was assembled from those logs, but the quoted times match no run and the "37s under the floor with three suspects" ending is a blend of two different stops that occurred in neither.)

Redone with the real numbers, the conclusion survives: the pot was essentially exhausted at the 3-suspect stop (~1080s spent, not 1043s), and finishing from there costs at most three more levels at full estimates (the 166s for `tests_exceptions` matches the logs; the other two derive the same way), so ~1080 + 426 ~= 1506 <= 1620 — about **114s** of spare, not ~150s.

Since this comment is the permanent provenance for the 1620s value and the repo's convention is that a cited run can be fetched and re-verified (I just did, and it contradicts every specific number here), please rewrite it against 34363328716 with what that log actually shows — and amend the commit message too, since release notes are generated from it and it carries the same numbers.

tests/scheme/run-all.sh
```diff
@@ -107,7 +111,7 @@ timeout_for() {
 # wait_with_timeout counts sleep 0.05 ticks, not wall clock, so that is
 # ~20 min of wall clock on Linux and stretches further where sleep spawns
 # cost more — inside every job cap that runs this suite.
-PER_SCRIPT_TIMEOUTS="bundle-cpu-baseline-2515.sh:${KAAPPI_BUNDLE_CPU_BASELINE_TIMEOUT:-1200} unit-chunk-watchdog-2560.sh:${KAAPPI_UNIT_CHUNK_WATCHDOG_TIMEOUT:-480}"
+PER_SCRIPT_TIMEOUTS="bundle-cpu-baseline-2515.sh:${KAAPPI_BUNDLE_CPU_BASELINE_TIMEOUT:-1200} unit-chunk-watchdog-2560.sh:${KAAPPI_UNIT_CHUNK_WATCHDOG_TIMEOUT:-600}"
```
baijum, line 114: This retune is what the **freebsd-test leg of this PR's own CI run failed on** (job 104089675204, "Run test suites (FreeBSD VM)"):

```
FAIL  tests/scheme/test-runner/shell-timeout-override.sh
FAIL: run-all.sh's PER_SCRIPT_TIMEOUTS entry is not exactly:
       PER_SCRIPT_TIMEOUTS="bundle-cpu-baseline-2515.sh:${KAAPPI_BUNDLE_CPU_BASELINE_TIMEOUT:-1200} unit-chunk-watchdog-2560.sh:${KAAPPI_UNIT_CHUNK_WATCHDOG_TIMEOUT:-480}"
```

`tests/scheme/test-runner/shell-timeout-override.sh` pins this line verbatim — `grep -Fxq` on the whole string, plus the same string in its failure message — and its header says the pin exists so "a malformed key or a retuned default cannot slip through a looser match". It caught exactly this retune. The `480` needs to become `600` in both places there (grep pattern and echoed expected line).

For the record, the suite itself passed on that leg under the new budget (`PASS tests/scheme/tools/unit-chunk-watchdog-2560.sh` at 18:13:35), so the 600 looks adequate on a slow VM leg — only the guard is stale. This is also the miss the checklist's third box papers over: "the full `run-all.sh` was not re-run" is exactly the gap this guard sits in; running just `tests/scheme/test-runner/shell-timeout-override.sh` (it needs no kaappi binary and no zig) would have caught it in seconds.

tests/scheme/tools/unit-chunk-watchdog-2560.sh
```diff
@@ -215,42 +216,112 @@ grep -q "WALL BUDGET EXCEEDED after .*s (compiling)" "$out" ||
...
+# still pinned by count and names. A 20s pot lets the descent print two
+# or three levels and then dry up, which keeps the case short.
+echo 2 > "$work/count8"
+rc=0
+# shellcheck disable=SC2086  # base_env is a deliberate VAR=val word list for env
+env $base_env KAAPPI_CHUNK_BUDGET=3 KAAPPI_BISECT_BUDGET=20 \
+    SHIM_HANG_FIRST="$work/count8" \
+    bash tools/run-unit-test-chunk.sh rest > "$out" 2>&1 || rc=$?
+[ "$rc" -eq 124 ] || fail_bisect "case 8 (forced depth-1 overrun): wall-budget overrun must exit 124 (got $rc)"
+grep -q "bisect depth 1: .* did NOT finish in 2s at its full estimated allowance .* -- descending into this half" "$out" ||
```
baijum: Nit, non-blocking: the literal `2s` here is base_env's `KAAPPI_BISECT_LEVEL_BASE_SECS=2` re-stated by hand — a later retune of base_env breaks only this grep while `check_level_estimates` keeps deriving its estimates correctly. `did NOT finish in [0-9]*s` would track base_env; the decisiveness pin is already carried by "at its full estimated allowance" later in the pattern. Fine to keep as a deliberate pin of "the full 2s allowance" too — just flagging the coupling.

baijum (commented):
Verified where I could, and the substance holds up: the suite passes locally on this machine (4m47s; case 8's forced descent takes `check_level_estimates`' descent branch and matches `bignum..tests_printer`, 11 filters, 15s estimate — derived independently the same way the test derives it), the window arithmetic matches the chunk script's own `lo=0, hi=count, mid=(lo+hi)/2` derivation on both paths, and the cap arithmetic checks out (1320+1620+120 = 3060s = 51m rest; 8+38+41 = 87 <= 90m job cap; YAML parses; CI indeed runs no shellcheck).

Two blockers:

1. **freebsd-test on this PR's own run is red**: `tests/scheme/test-runner/shell-timeout-override.sh` pins the `PER_SCRIPT_TIMEOUTS` line verbatim — including the `:-480` default this PR retunes to 600 — and was not updated. That guard exists precisely so "a retuned default cannot slip through a looser match"; it caught exactly this retune. (Details inline on run-all.sh.)

2. **The permanent `ci.yml` comment cites run evidence that is not in the cited run.** Run 34863343107's riscv64-test rest chunk ran healthy (9m36s, `912 tests (floor 7)`, no watchdog, no bisect lines anywhere in the log). The descent the comment describes did happen — in runs 34363328716 / 34370359849 / 34379866388 from 2026-09-09 — but the quoted level times match none of them and the quoted stop shape ("37s left, under the 40s level floor, with three suspects") occurred in none of them. The corrected numbers still support 1620s (~1080s spent + 426s worst-case finish ~= 1506s <= 1620, ~114s spare), so I'd keep the value and rewrite the provenance. (Details inline on ci.yml.)

Minor, non-blocking: case 8's grep pins the literal `2s` allowance from base_env's `KAAPPI_BISECT_LEVEL_BASE_SECS=2`.

Also worth noting: netbsd-test — the leg that motivated #2596 — was still pending on this run when I reviewed; the freebsd leg ran the suite to completion under the new 600s budget, which is a good sign for it.

tests/scheme/run-all.sh
baijum: Fixed in bc7ea790: both the `grep -Fxq` pattern and the echoed expected line in `shell-timeout-override.sh` now carry `:-600`. Ran the guard directly (`PASS: shell-suite per-script timeout override routes, bounds, and reports correctly`) and re-ran the watchdog suite under /bin/bash 3.2 and bash 5.3 (rc=0, 4m43s / 4m48s). You are right that running the guard alone would have caught this in seconds; noted for the next shell-suite retune.

.github/workflows/ci.yml
baijum: I re-fetched the logs before touching the comment, and the numbers are real -- they are from **attempt 1** of run 34863343107. That run has `run_attempt: 2`: the rest chunk wedged on attempt 1, the failed-jobs re-run (attempt 2) passed, and `gh run view --log` returns only the latest attempt, which is why it reads as a healthy run. `gh api repos/kaappi/kaappi/actions/runs/34863343107/attempts/1/logs` gives the zip; its `20_riscv64-test.txt` has, verbatim: depth 1 `did NOT finish in 628s at its full estimated allowance`, depth 2 `finished in 130s`, depth 3 `finished in 175s`, depth 4 `finished in 106s`, then `remaining pot (37s) is too shallow for depth 5 (a level needs >= 40s ...)` and `bisect incomplete: 3 suspect filter(s) remain`. So the comment matched its citation; the citation just did not say which attempt. bc7ea790 pins it to attempt 1, says why the latest attempt looks clean, and names the three 2026-09-09 runs you pulled as the sibling descents (same suspects, cut short at a clipped depth 5 instead), with the tighter ~110s spare that shape implies alongside the ~150s for the attempt-1 shape. The 1620s value is unchanged. On the commit message: I have not amended (this repo's PRs land as squashes, and the follow-up commit body carries the correction); the squash message at merge time is the place to fold the attempt number in if you want it in the release note.

tests/scheme/tools/unit-chunk-watchdog-2560.sh
baijum: Applied in bc7ea790: `did NOT finish in [0-9]*s at its full estimated allowance`, and the fail message no longer says "2s". The decisiveness pin stays with "at its full estimated allowance". Re-verified with the full suite under both bash versions.

tests/scheme/run-all.sh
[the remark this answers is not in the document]
baijum: Fixed in bc7ea790 (both the grep pattern and the echoed expected line). The guard test and the watchdog suite pass locally.

.github/workflows/ci.yml
```diff
@@ -644,27 +644,36 @@ jobs:
...
+    # (41m) inside 90m, so a wedge anywhere localises fully; healthy runs
     # are unchanged (~50m) and stop at the first failed chunk anyway.
     #
     # `rest` runs FIRST although it is the biggest chunk: it is the one
     # that has wedged (both #2488 recurrences localised there, and both
     # #2560 cap kills), so a wedge fails fast instead of burning ~36m of
     # healthy chunks first, and its bisection is the least likely to be
     # clipped by the job cap.
-    timeout-minutes: 80
+    timeout-minutes: 90
```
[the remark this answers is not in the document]
baijum: Not changing this here: `cache-key: ${{ matrix.optimize }}` belongs to the `test` job (line 314) and is not touched by this PR, which edits only the `riscv64-test` job's bisection pot and step caps. The riscv64 unit tests run under QEMU in the dedicated `riscv64-test` job by design (a matrix leg cannot run riscv64 natively on GitHub runners), so the matrix is not missing an entry. Worth its own issue if the cache-key sharing between the two Linux legs is a real cache-hit problem.

PR #2598, opened 2026-09-15
Build the riscv64 release binary against glibc so (kaappi ffi) can dlopen

### LINKED ISSUE #2595: Released riscv64 kaappi is musl-static, so (kaappi ffi) cannot dlopen — the #1783 fix never reached riscv64
## Summary

`release.yml` builds the riscv64 artifacts with plain `-Dtarget=riscv64-linux`, which is Zig's musl-static default. A static musl binary cannot `dlopen`, so the released riscv64 `kaappi` rejects every `(kaappi ffi)` library — and with it the whole C-extension ecosystem (net/http/sqlite/pg/redis/crypto/math). This is the exact defect #1783 fixed for x86_64 and aarch64 by switching those rows to `zig_target: <arch>-linux-gnu.2.28` and adding the `linux-ffi-smoke` release gate; riscv64 was left on the default because it was interpreter-tier at the time, and #2592 (closing #2591) has now made it native-tier, which raises the stakes: `kaappi compile` on a riscv64 box links `libkaappi_rt.a` against the host's glibc anyway, so a glibc `kaappi` would also make the shipped binary and the shipped archive the same ABI.

## Repro (released binary, riscv64 Linux)

```scheme
(import (kaappi ffi))
(ffi-open "libm.so.6")   ; => error: dynamic loading unavailable in a static binary
```

## Fix

- `release.yml`: give the riscv64 row a `zig_target: riscv64-linux-gnu.2.28` (or whatever glibc floor the riscv64 distros need — Ubuntu 22.04 riscv64 ships 2.35, Debian 12 ships 2.36; 2.28 matches the other two rows if Zig's riscv64 glibc stubs go that far back) and keep `target: riscv64-linux` as the artifact-naming key, as the other rows do.
- Extend the `linux-ffi-smoke` gate to riscv64 (under QEMU, the `riscv64-test` pattern) so the regression cannot come back silently.
- The trade-off note from #1783 applies: a glibc binary does not run on Alpine riscv64; those users build from source until a musl-dynamic artifact ships.

## Priority

`medium`: wrong against a documented guarantee (`docs/dev/ffi.md`, the ecosystem's C-FFI libraries) on a shipped platform, but loud and narrow.

baijum:
## Summary

Closes #2595

The released riscv64 `kaappi` was Zig's musl-static default, so every `ffi-open` on a riscv64 install failed with `Dynamic loading not supported` — the defect #1783 fixed for x86_64/aarch64, which skipped riscv64 because it was interpreter-tier at the time. #2592 made it native-tier, where the mismatch was also incoherent: `kaappi compile` on a riscv64 box already links `libkaappi_rt.a` against the host's glibc, so the shipped `kaappi` was the one musl thing in the chain.

**`release.yml`**

- The riscv64 build row gets `zig_target: riscv64-linux-gnu.2.28` — the same floor as the other two Linux rows. `target: riscv64-linux` and every artifact name are unchanged, so `install.sh` keeps matching. The row comment now describes this state instead of the musl gap; the #1783 trade-off note carries over (no Alpine riscv64).
- `linux-ffi-smoke` gains a riscv64 leg. No GitHub runner executes riscv64, so the leg runs the same two smoke commands under QEMU user-mode inside the `riscv64/ubuntu:24.04` image `ci.yml`'s `riscv64-native-test` already pins (same digest, same `docker/setup-qemu-action` SHA; glibc 2.39, has `libm.so.6`). A matrix key `emulate: riscv64` gates the two QEMU/pull steps (`if:`) and swaps the artifact for a `docker run --rm -i --platform linux/riscv64 …` prefix in the smoke step; the two native rows leave it unset and run exactly what they ran before. Both `grep -qx` calls run on the runner under the default `bash -eo pipefail`, so a wrong answer or a refused `dlopen` fails the job. Also `fail-fast: false`, so a flaky QEMU leg cannot cancel the native legs' evidence (the `release` job still needs all three).

**Docs** — `docs/dev/porting.md` Stage 6 (a Linux release row needs the `gnu.2.28` target *and* a smoke leg; the checklist is where the omission was made), `docs/dev/ffi.md` (why a static musl binary cannot `dlopen`, and what gates the release on it), and the `post-release.yml` comment that lists what exercises riscv64. The 2026-09-14 postmortem's line about "the musl-static archive `release.yml` ships" is left as-is: postmortems are point-in-time records, and it was true on that date. No `CHANGELOG.md` edit (release-time only); the docs-site repo is not touched.

## What was verified locally (macOS aarch64, Zig 0.16.0, podman)

`release.yml` cannot be exercised without a tag push, so the workflow itself is unrun. **The first real check is the next release's `linux-ffi-smoke` riscv64 leg.** What could be verified:

- The exact build-step commands with the new target: `zig build -Doptimize=ReleaseSafe -Dtarget=riscv64-linux-gnu.2.28 -Dstrip=true` and `zig build lib …` both succeed. `file` on the result:

  ```text
  kaappi: ELF 64-bit LSB executable, UCB RISC-V, RVC, double-float ABI, version 1 (SYSV),
          dynamically linked, interpreter /lib/ld-linux-riscv64-lp64d.so.1, for GNU/Linux 2.0.0, stripped
  thottam: (same)
  libkaappi_rt.a: current ar archive
  ```

- The smoke commands, in the job's exact shape (piped stdin → `podman run --rm -i --platform linux/riscv64 … ./kaappi-riscv64-linux /dev/stdin` → host-side `grep -qx` under `pipefail`), against that binary inside the pinned `riscv64/ubuntu:24.04@sha256:f28b7182…` image:

  ```text
  $ ./kaappi-riscv64-linux --version      → Kaappi Scheme v0.27.1
  (display (+ 1 2))                        → 3      grep -qx '3'   exit=0
  (ffi-open "libm.so.6") … (c-sqrt 4.0)    → 2.0    grep -qx '2.0' exit=0
  negative control: grep -qx '4' on the first → exit=1 (the pipeline fails, as the job needs)
  ```

- A control build with the old bare `-Dtarget=riscv64-linux` (`statically linked`) run through the same ffi leg fails it with the error users saw — so the new leg would have caught the shipped artifact:

  ```text
  /dev/stdin:2:14: error[KP3002]: ffi-open: cannot open 'libm.so.6': Dynamic loading not supported (also tried .dylib/.so/.so.6 suffixes and /root/.kaappi/lib/)
  … exit=1
  ```

- `python3 -c 'import yaml; yaml.safe_load(…)'` on `release.yml` and `post-release.yml`; the parsed matrix shows all three Linux build rows on `gnu.2.28` and the smoke matrix with the `emulate` row. `markdownlint-cli2` on the two edited docs: 0 issues.
- Not verified: the workflow on GitHub's `docker` (podman stood in locally; the `-i`/`/dev/stdin`/`-v $PWD:$PWD` shape is the one `ci.yml`'s riscv64 job already uses with docker), and the e2e cross suite was not re-run here — #2592 reported 38/38 with a `-Dtarget=riscv64-linux-gnu` archive already.

## Checklist

- [x] `zig build test` passes — no `src/` change; the unit suite is unaffected by this PR
- [x] `zig fmt --check src/` passes — no `.zig` files touched
- [x] Scheme test suites pass (`bash tests/scheme/run-all.sh`) — not run; no interpreter or library change
- [ ] New tests added for new behavior — the new behaviour *is* a test (the `linux-ffi-smoke` riscv64 leg); it runs only on a tag push


.github/workflows/release.yml
```diff
@@ -55,12 +55,18 @@ jobs:
           # riscv64 is native-tier since 2026-09 (the archive is what
           # `kaappi compile` links on a riscv64 box; ci.yml's
           # riscv64-native-test exercises exactly these artifacts under
-          # QEMU). Still Zig's musl-static default rather than the
-          # gnu.2.28 the two arches above use, so this binary cannot
-          # dlopen (kaappi ffi) — the #1783 trade-off, not yet resolved
-          # for riscv64.
+          # QEMU). Same glibc-2.28 floor as the two rows above, for the
```
baijum: This sentence predates the `zig_target` and is now stale. `riscv64-native-test` cross-builds with the bare target — `bash tests/e2e/run-e2e-cross.sh riscv64-linux` — and that script's header documents the choice ("Zig's musl-static default, so the linked binaries need no target sysroot"). After this PR the release artifacts are gnu.2.28, so the job exercises this row's compiler/VM code under QEMU but no longer *exactly these artifacts*. Suggest rewording to something like "ci.yml's riscv64-native-test exercises this row's code under QEMU (cross-built with the musl default for sysroot-free linking)". The ABI gap is partly covered the other way: the 2026-09-14 postmortem records 38/38 e2e with a `-Dtarget=riscv64-linux-gnu` archive.

.github/workflows/release.yml
```diff
@@ -276,29 +282,65 @@ jobs:
   # dynamic loading (e.g. Zig's default musl-static target) passes every
   # build step but cannot load a single C-extension library once
   # installed. Run the actual release artifact and prove dlopen works
-  # before anything is published.
+  # before anything is published. Every glibc Linux row in `build` must
+  # have a leg here — riscv64 had none from #1783 (2026-07) until #2595,
+  # so its binary shipped musl-static with nothing to notice.
   linux-ffi-smoke:
```
baijum, line 292: Consider a `timeout-minutes:` on this job. This repo's own history is that riscv64-under-QEMU legs wedge: kaappi#2488 found the unit-test binary hung under TCG (twice since September), which is why every riscv64 chunk in ci.yml carries its own bound. The smoke leg runs two tiny programs so even a generous 15–30m cap costs nothing, but without one a wedged `docker run` would hold the tag's publish on the 360m runner default before failing — six hours of dead release pipeline instead of a red herring to re-run.

docs/dev/ffi.md
```diff
@@ -136,6 +136,18 @@ without library validation.
 the wrapper but never `dlclose`s, because a function bound from it may
 still be live.
 
+All of this presupposes a binary that can `dlopen` at all. Zig's bare
+`<arch>-linux` target is musl-static, and static musl has no dynamic
+loader, so such a `kaappi` rejects every `ffi-open` — which is how every
+released Linux binary was FFI-dead until #1783 switched the x86_64 and
+aarch64 release rows to `<arch>-linux-gnu.2.28`, and how riscv64 stayed
+that way until #2595 gave its row the same `zig_target`. `release.yml`'s
```
baijum: As written, this paragraph reads as if every released Linux binary can now `dlopen` — but the s390x and powerpc64le rows still carry no `zig_target`, so their release binaries are still musl-static and still reject `ffi-open`. #1783's changelog line "Interpreter-tier arches are unaffected" was about `kaappi compile` ABI coherence, not FFI availability. Half a sentence naming the standing exception ("the interpreter-tier s390x/ppc64le rows still ship musl-static") would keep the FFI doc from over-promising — or, if FFI is meant to work on those installs too, that's an issue worth filing separately (s390x QEMU legs already exist in ci.yml, so a smoke leg is cheap if it's ever in scope). Tiny nit in the same paragraph: "until #2595 gave its row the same zig_target" attributes the fix to the issue; the PR number will read better once this merges.

docs/dev/porting.md
```diff
@@ -275,6 +275,14 @@ thottam → #1608 readiness), and it kept every intermediate PR shippable.
 - [ ] Add a `release.yml` matrix row: `target`, `artifact`, `exe_ext`,
       `lib_ext`, `strip`, plus any post-processing (macOS signs +
       notarizes; Windows ships unstripped until #1613's toolchain bump).
+      A Linux row also needs `zig_target: <arch>-linux-gnu.2.28` — Zig's
```
baijum: Same nuance from the checklist side: "A Linux row also needs `zig_target`" is categorical now, but the two existing interpreter-tier rows (s390x, powerpc64le) don't have one, so a porter following this list will find the matrix contradicting it. Either add "(the interpreter-tier s390x/ppc64le rows are the standing exception — they ship musl-static and reject `ffi-open`)" or make the rule match reality. Cross-referencing the ffi.md comment.

baijum (commented):
Reviewed the workflow change end to end and re-verified the load-bearing claims locally. The change is correct and well-gated — four non-blocking comments inline.

**Independently verified:**

- Both workflow files parse; the smoke matrix is well-formed, and `release` already `needs: linux-ffi-smoke`, so the new leg genuinely gates the tag publish (with `fail-fast: false` keeping a flaky QEMU leg from cancelling the native legs' evidence).
- The image digest (`f28b7182…`) and the `docker/setup-qemu-action` SHA match ci.yml's `riscv64-native-test` pins exactly, as claimed.
- Zig 0.16.0 accepts `riscv64-linux-gnu.2.28` — `zig cc -target riscv64-linux-gnu.2.28 -c` cross-compiles a riscv64 object here, so the glibc-stub floor parses. (glibc's riscv64 port landed in 2.27, so a 2.28 floor is coherent for the arch, and every riscv64 distro ships ≥2.35.)
- `markdownlint-cli2` on the two docs: 0 issues.
- Artifact naming is untouched — `target` stays the key and the build commands use `zig_target || target` — so `install.sh` matching is preserved.
- The `docker run -i` / `/dev/stdin` / `-v $PWD:$PWD -w $PWD` shape is the proven `KAAPPI_EMU` pattern from ci.yml, plus the podman negative controls in the description (wrong expectation fails; the old musl build fails the ffi leg with the user-visible error). Good evidence discipline.
- Bonus the description doesn't claim: the 2026-09-14 postmortem records the e2e at 38/38 with a `-Dtarget=riscv64-linux-gnu` archive too, so the gnu ABI of the shipped archive is not untested.

As noted, the workflow itself can't run before a tag push; the first real exercise is the next release. CI on this branch is still running as I write.

.github/workflows/release.yml
baijum: Reworded in 448346dc. The row comment now says riscv64-native-test exercises this row's code under QEMU, cross-built with the musl default for sysroot-free linking (pointing at run-e2e-cross.sh's header), and that the gnu-archive coverage is the 2026-09-14 postmortem's 38/38 e2e run with a `-Dtarget=riscv64-linux-gnu` archive. The "exactly these artifacts" claim is gone.

.github/workflows/release.yml
baijum: Added `timeout-minutes: 20` on linux-ffi-smoke in 448346dc, with a comment naming why: the legs run two tiny programs, but the riscv64 one runs under TCG and riscv64-under-TCG legs do wedge (#2488), and without a cap a stuck `docker run` would hold the publish on the 360-minute runner default. The commit body carries the same reasoning.

docs/dev/ffi.md
baijum: Verified against release.yml: the s390x-linux and powerpc64le-linux rows carry no `zig_target`, so they still build on the bare musl-static target. 448346dc names them in ffi.md as the standing exception ("still ship musl-static and still reject `ffi-open`"), narrows "a new Linux row needs a leg there" to glibc rows, attributes the fix to #2598 with #2595 as the issue, and adds the Alpine riscv64 trade-off sentence CodeRabbit asked for in the same paragraph. Not filing an s390x/ppc64le FFI issue from here — whether FFI is in scope on those installs is a maintainer call.

docs/dev/porting.md
baijum: 448346dc makes the checklist match the matrix: the rule is now scoped to "a Linux row whose users are meant to load C extensions", and it ends with the parenthetical you suggested — the interpreter-tier s390x and powerpc64le rows are the standing exception, ship musl-static and reject `ffi-open` — cross-referencing ffi.md's "Finding the library" section, which says the same thing.

docs/dev/ffi.md
```diff
@@ -136,6 +136,18 @@ without library validation.
...
+`linux-ffi-smoke` job runs each glibc Linux artifact (riscv64 under
+QEMU) through an `ffi-open` of `libm.so.6` before anything is published;
+a new Linux row needs a leg there ([porting.md](porting.md) Stage 6).
+A from-source `zig build` on a glibc host is unaffected: it targets the
+host's own libc.
+
 ## Callbacks
 
 `ffi-callback` hands C a function pointer for a Scheme procedure. There
```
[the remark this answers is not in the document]
baijum: The paragraph did not say it; 448346dc adds one sentence to ffi.md: the price of a glibc artifact is that it does not run on a musl distro, so Alpine riscv64 users build from source until a musl-dynamic artifact ships, as on the other two arches.

Issue #20, opened 2026-06-26
thottam: subprocess exit-status decode treats signal-killed git/make as success


baijum:
## Summary

`thottam` decodes child exit status with hand-rolled bit math (`(status >> 8) & 0xff`) that is only valid when the child exited *normally*. A subprocess killed by a **signal** (e.g. `git`/`make`/`cp` segfaults or is killed) has its signal number in the low 7 bits and `WIFEXITED` false; `>> 8` then yields `0`, which this code treats as **success**. A crashed build/clone step is silently reported as having succeeded.

## Location

`src/thottam.zig:246` (`runCapture`) and `:284` (`runPassthrough`).

```zig
var status: c_int = 0;
_ = std.c.waitpid(pid, &status, 0);
const exit_code: u8 = @intCast(@as(c_uint, @bitCast(status)) >> 8 & 0xff);
if (exit_code != 0) { ... return error.CommandFailed; }
```

## Why it's a bug

`(status >> 8) & 0xff` is the `WEXITSTATUS` value, but it is only meaningful when `WIFEXITED(status)` is true. For a signal-terminated child, `WIFEXITED` is false and the low byte encodes the terminating signal (with the high exit-status byte zero), so `exit_code` computes to `0` → treated as success. This path backs both the capture path (`runCapture` → `runGitCapture`/`getPkgSha`/dylib+sld discovery) and the passthrough path (`runPassthrough` ← `runGit` ← clone/build), so a `git clone` or `make` that crashes or is killed is reported as a successful step.

## Impact

A subprocess that dies by signal (crash, OOM-kill, `SIGINT`/`SIGTERM`) is treated as success, so `thottam` proceeds with a half-installed or empty package — corrupt install state with no error surfaced to the user.

## Suggested fix

Check the status bits properly: treat the command as failed unless `WIFEXITED(status)` is true *and* `WEXITSTATUS(status) == 0`; treat `WIFSIGNALED(status)` as failure. Equivalently, use `std.posix.waitpid`, which returns a struct with the decoded status, instead of `std.c.waitpid` + manual bit math.

## Confidence

Medium (clear correctness defect in status decoding; impact is a silent failure rather than a crash/corruption).

PR #2599, opened 2026-09-15
Run the riscv64 native e2e suite against a gc-stress archive daily

### LINKED ISSUE #2594: riscv64: run the native-backend e2e programs against a -Dgc-stress runtime archive
## Summary

#2592 (closing #2591) verifies the riscv64 native tier with the e2e parity suite under QEMU (`tests/e2e/run-e2e-cross.sh`, CI job `riscv64-native-test`), but only against a normally built `libkaappi_rt.a`. The GC-rooting contract between native code and the runtime (`kaappi_gc_push_root` slots, the `alloca`-based root stack, the quote/eval caches) has never been exercised on riscv64 with collection forced on every allocation — and rooting bugs are exactly the class that stays latent until allocation order or stack layout differs, which a new arch changes (kaappi#1401, #1854, #1855 were all found that way on the primary platform).

## Ask

Run the e2e programs on riscv64 with a `-Dgc-stress=true` runtime archive:

```bash
zig build lib -Dtarget=riscv64-linux -Dgc-stress=true --prefix zig-out-cross-riscv64-linux
KAAPPI_EMU="podman run --rm --platform linux/riscv64 -v $PWD:$PWD -v /private/tmp:/private/tmp -w $PWD kaappi-builder-riscv64" \
  bash tests/e2e/run-e2e-cross.sh riscv64-linux
```

(`run-e2e-cross.sh` currently rebuilds the archive itself without the option; either add a `KAAPPI_CROSS_BUILD_ARGS` passthrough or run its link/compare phases against a pre-built stress archive.) Under TCG the stress archive will be slow; the loop-heavy programs (`tak`, `native-fib`, `native-loop-*`) are the ones worth the wait.

If it passes, consider a scheduled (not per-PR) CI leg, as `gc-stress` and `gc-stress-scheme` already are for the host. If it fails, that is a rooting bug in the native tier or the runtime bridge, not a riscv64 bug, and should be reproduced on aarch64/x86_64 first (lessons-learned #20).

baijum:
## Summary

Runs the riscv64 native-backend e2e parity suite against a `libkaappi_rt.a` built with `-Dgc-stress=true`, and adds a **scheduled** (not per-PR) CI leg that keeps doing so.

Closes #2594

### Local run (podman, Apple Silicon, `kaappi-builder-riscv64`; tree at e4a287e5 + this change)

| Phase | Wall |
|-------|------|
| `zig build -Dtarget=riscv64-linux --prefix P` (kaappi, default GC) | 55 s |
| `zig build lib -Dtarget=riscv64-linux -Dgc-stress=true --prefix P` | 37 s |
| `KAAPPI_CROSS_PREFIX=P bash tests/e2e/run-e2e-cross.sh riscv64-linux /usr/local/bin/zig` | **94 s** |
| same, plain archive (baseline) | 82 s |

```
=== Cross E2E Summary (riscv64-linux): 39/39 passed ===
```

All 37 programs plus the argv passthrough and the on-target `kaappi compile` smoke (which links against the stress archive via `KAAPPI_LIB_DIR`). The loop-heavy subset the issue names (`tak`, `native-fib`, the two `*-tail` programs) passed first in 67 s; `native-loop-*.scm` does not exist under `tests/e2e/programs/`. The stress GC adds ~15% — the programs are small, so the ~115 container starts dominate. **No failure, so no issue filed.**

The stress archive is verified as such: the freed-owner UAF panic string (`memory.uaf_detection`, compiled in only under `-Dgc-stress` in ReleaseSafe) occurs 1× in the stress archive and 0× in a plain one; the workflow refuses an archive without it.

### Where the leg lives

`.github/workflows/gc-stress-riscv64.yml` — `schedule: '17 3 * * *'` (after fuzz's 02:47, before the 05:17 ecosystem nightly) plus `workflow_dispatch` with an optional `programs` subset input. `timeout-minutes: 30` (the plain `riscv64-native-test` job runs ~6.5 min end to end on a hosted runner; assumption: TCG under docker on `ubuntu-latest` is comparable to podman on Apple Silicon), with the run step bounded by `timeout 22m` so a collector livelock fails the step with its PASS/FAIL lines intact rather than cancelling the job. Pins, image digest and the riscv64 Zig tarball are copied verbatim from `riscv64-native-test`, whose comment now points at this sibling.

A separate workflow rather than a job in `fuzz.yml`: its trailing `report` job auto-files issues for its job set with fuzz verdicts, and a parity failure here is a rooting bug to reproduce on aarch64/x86_64 first (lessons-learned #20), not a fuzz-infrastructure finding.

**Manual dispatch:** `gh workflow run gc-stress-riscv64.yml --ref claude/riscv64-gc-stress-e2e-2594` returns HTTP 404 — GitHub only registers a `workflow_dispatch` workflow once it exists on the default branch. Trigger it once after merge: `gh workflow run gc-stress-riscv64.yml`.

### Also

- `tests/e2e/run-e2e-cross.sh` gains `KAAPPI_E2E_PROGRAMS` (space-separated basenames) to narrow the parity loop for bisecting or splitting a slow run; the argv and smoke phases still run, and a list matching nothing exits 2 (verified) instead of passing on those two alone.
- `docs/dev/llvm-backend.md` (riscv64 "Testing" paragraph): the leg, the `KAAPPI_CROSS_PREFIX` recipe, the measured numbers, and the host reproduction recipe for a failure. `docs/dev/testing.md` and `docs/dev/fuzzing.md` name the workflow where they list the scheduled/riscv64 legs.
- No `src/*.zig` or `CHANGELOG.md` changes.

## Checklist

- [x] `zig build test` passes (2097/2118, 21 skipped; no Zig source changed)
- [ ] Scheme test suites pass (`bash tests/scheme/run-all.sh`) — not run; nothing here touches Scheme behaviour (workflow, one shell script, docs)
- [x] New tests added for new behavior — the scheduled leg itself; the filter's empty-match guard verified by hand (exit 2)


.github/workflows/gc-stress-riscv64.yml
```diff
@@ -0,0 +1,124 @@
...
+      programs:
+        description: 'Space-separated program basenames (without .scm) from tests/e2e/programs to run instead of all of them, e.g. "tak native-fib". The argv and on-target smoke phases always run.'
+        required: false
+        default: ''
+
+permissions:
+  contents: read
+
+jobs:
+  riscv64-native-gc-stress:
```
baijum, line 51: The one thing this workflow doesn't arrange is who *hears* about a red run. fuzz.yml trails with a `report` job that auto-files issues (`if: failure() || cancelled()`, `issues: write`) — deliberately not reused here so a parity failure isn't filed with a fuzz verdict, which is right — but the replacement is currently nothing: a nightly that goes red is visible only to whoever browses the Actions tab. Consider a minimal report step in this workflow too, in the same shape but filing a plain "gc-stress riscv64 nightly failed" issue with the run link and the PASS/FAIL tail. If you do, fuzz.yml's `cancelled()` lesson applies verbatim here: the run step's `timeout` produces a *failure*, but a job-level `timeout-minutes` cancellation is CANCELLED — so `failure()` alone would skip the report for exactly the livelock case the cap exists to catch. (Needs `issues: write` added next to `contents: read`.)

.github/workflows/gc-stress-riscv64.yml
```diff
@@ -0,0 +1,124 @@
...
+    # can present as a collector livelock rather than a panic, and the
+    # `timeout` on the run step below fails the step -- with the per-program
+    # PASS/FAIL lines already in the log -- before this cap cancels the job.
+    timeout-minutes: 30
+    env:
+      # ziglang.org/download/index.json, 0.16.0 -> riscv64-linux.shasum.
+      RV_ZIG_SHA256: bc069b0f2f568f54bafbdfc1d65b12fd386ed6a652044a37aee6a4f72f14076e
+      # riscv64/ubuntu:24.04 as of 2026-09-07; bump the digest deliberately,
+      # together with ci.yml's riscv64-native-test.
+      RV_IMAGE: riscv64/ubuntu:24.04@sha256:f28b71827ff6cb4fe9b6b536b583bc0227072516de20c9225847be7722ab5715
```
baijum, line 65: This makes the RV_ZIG_SHA256 + RV_IMAGE + action pins a two-place commitment, with comments on both sides pointing at each other — reasonable for two. If a third consumer ever appears (s390x gc-stress? an aarch64/x86_64 native-stress leg?), factoring the shared skeleton — checkout, setup-zig, QEMU, the riscv64-Zig fetch — into a reusable workflow (`workflow_call`) would beat a third copy of the tarball-fetch dance. Just something to keep in the back pocket, not a request to do it now.

baijum (commented):
Reviewed end to end and re-verified the load-bearing claims locally — including rebuilding both archives to test the UAF-string gate. The change is sound and unusually well-evidenced; four non-blocking comments inline.

**Independently verified:**

- **The leg really exercises the stress archive in both phases.** With `KAAPPI_CROSS_PREFIX` set, `run-e2e-cross.sh` skips both of its own builds and pins `cross_link` (`-L"$LIBDIR" -lkaappi_rt`) *and* the on-target `kaappi compile` smoke (`KAAPPI_LIB_DIR="$LIBDIR"`) to the prefix's archive — nothing rebuilds a plain archive over it. The oracle interpreter stays default-GC, as the design intends.
- **The UAF-string guard works empirically.** I built a plain and a `-Dgc-stress=true` `libkaappi_rt.a` here: the workflow's exact `grep -a -c "marking freed object"` reports **0 in the plain archive, 1 in the stress one**. Source agrees — `memory.zig:74` gates `uaf_detection` on `.Debug or build_options.gc_stress`, so a comptime-false branch drops the string from a plain ReleaseSafe archive. A plain archive cannot satisfy the gate; the leg cannot silently degrade into a copy of `riscv64-native-test`.
- **The filter logic is correct at the edges.** Simulated `KAAPPI_E2E_PROGRAMS` matching: exact word matching (`native` does not match `native-fib` — the space-delimited `case` pattern is doing word matching, not substring), whitespace-padded lists work, empty/unset runs everything, zero matches exits 2.
- YAML parses (schedule 03:17 UTC + `workflow_dispatch`), `bash -n` clean, markdownlint 0 issues on all three docs.
- Cross-references all check out: fuzz.yml's cron is 02:47 and its `report` job is real (which is exactly why a separate workflow was the right call here); lessons-learned #20 says what the comments claim; `KAAPPI_CROSS_PREFIX` is the #2592 mechanism.
- Timing margins are sane: the plain sibling runs ~6.5 min end-to-end on a hosted runner, stress adds ~1 min of host build plus ~15% on the run phase, and 22m/30m leaves roughly 2× headroom. The first post-merge dispatch will calibrate the podman-vs-docker-TCG assumption; the `programs` input exists precisely for splitting if it comes up short.

The description's caveat that `workflow_dispatch` 404s until the workflow exists on the default branch is accurate GitHub behavior — worth a calendar note to trigger it once after merge so the first scheduled run isn't also the first-ever run.

docs/dev/llvm-backend.md
```diff
@@ -922,6 +922,46 @@ binary (empty under binfmt; a `podman run --platform` line locally — the
...
+on-target smoke.) Measured 2026-09-15: 39/39 in 94 s against the stress
+archive vs 82 s plain — the programs are small, so container starts
+dominate and the stress GC adds ~15%. `KAAPPI_E2E_PROGRAMS="tak native-fib"`
+narrows the parity loop to bisect a failure. **A failure here is a rooting
+bug in the emitted code or the runtime bridge, not a riscv64 bug**
+(lessons-learned #20): reproduce it on the host first by linking that one
+program against a host stress archive —
+
+```bash
+zig build lib -Dgc-stress=true --prefix "$H"
```
baijum: This block assumes `zig-out/bin/kaappi` already exists, but its first command builds only the lib — on a fresh tree the second line dies with "no such file". Consider starting the block with a plain `zig build` line so the recipe is copy-paste complete.

docs/dev/testing.md
```diff
@@ -831,8 +831,14 @@ matrix covers:
...
 
-`fuzz.yml` runs the fuzz targets on a schedule — see
-[fuzzing.md](fuzzing.md).
+Two more workflows run on a schedule, not per PR. `fuzz.yml` runs the fuzz
+targets — see [fuzzing.md](fuzzing.md). `gc-stress-riscv64.yml` (daily at
+03:17 UTC, also `workflow_dispatch`) runs the riscv64 native-backend e2e
+parity suite of `riscv64-native-test` against a `libkaappi_rt.a` built with
+`-Dgc-stress=true` (kaappi#2594) — the one leg that exercises *native* code
+against a collection-per-allocation runtime, on the arch whose stack layout
+the GC-rooting bridge was written last for. `docs/dev/llvm-backend.md`
```
baijum: Tiny prose nit: "on the arch whose stack layout the GC-rooting bridge was written last for" is a garden path — the first parse is "written last, for…". Something like "on the newest arch under the GC-rooting bridge" reads on the first pass.

.github/workflows/gc-stress-riscv64.yml
baijum: Agreed, and left as two copies for now (no change in 507a8650). If a third consumer appears I would cut the seam at "cross-build for a QEMU target and hand back a prefix": a `workflow_call` reusable workflow taking `target`, `build-args` (e.g. `-Dgc-stress=true` for the lib), and `image`, owning checkout, setup-zig, setup-qemu, the pinned riscv64-Zig fetch and the digest-pinned image pull, and exporting the prefix path plus `RV_ZIG` as outputs — so ci.yml's riscv64-native-test, this leg, and an s390x/aarch64 stress sibling would each be a `uses:` plus one run step. The pins and `RV_ZIG_SHA256`/`RV_IMAGE` would then live in exactly one file. Not worth it at two; the cross-pointing comments are the cheaper contract until then.

docs/dev/llvm-backend.md
baijum: Fixed in 507a8650: the block now starts with a plain `zig build` (the driver) before `zig build lib -Dgc-stress=true --prefix "$H"` (the archive it links against), so it is copy-paste complete on a fresh tree.

docs/dev/testing.md
baijum: Reworded in 507a8650 to "on the newest arch under the GC-rooting bridge", as suggested.

.github/workflows/gc-stress-riscv64.yml
baijum: Done in 507a8650: a `report` job in fuzz.yml's shape — a separate job (the `issues: write` + `actions: read` token never coexists with the job that runs the emulated binaries; the workflow level stays `contents: read`), `if: failure() || cancelled()`, and for a cancellation it files only when the job carries the `exceeded the maximum execution time` check-run annotation, so a manual/concurrency cancel files nothing while a `timeout-minutes` kill — the livelock case — does.

It files one plain issue, "gc-stress riscv64 nightly failed", with the run link, a verdict (cap exceeded / parity FAIL with its expected-actual pair / suite killed mid-run with the last PASS naming where it stalled / failed before the loop), and the PASS/FAIL tail read from the job log via the API — no artifact, no fuzz verdict wording; the body points at llvm-backend.md "Testing" for the reproduce-on-host-first rule. It dedupes by exact title among open issues and appends there, as fuzz.yml does.

Filed `priority: high` (plus `ci`, `gc`, `native-backend`): a parity FAIL is a silently wrong answer from a legal program, and at filing time it cannot be told apart from a hang or an infrastructure failure — the body asks for a relabel to `medium`/`low` once triage says which.

Dry-ran the filter/verdict chain against the local run log (green, truncated, and injected-FAIL shapes) and `bash -n`-checked every extracted `run:` block; the job itself cannot be exercised until the workflow exists on main.

PR #2600, opened 2026-09-18
Add a tailcc/musttail probe; record why riscv64 cannot enable it

### LINKED ISSUE #2593: riscv64: enable tailcc/musttail (fast_tailcalls_supported) once verified on the target
## Summary

#2592 (closing #2591) adds riscv64-linux to the native tier with `fast_tailcalls_supported` (`src/llvm_emit.zig`) left `false` there. That is the safe choice — no `@name.fast` entries are emitted, every named function keeps its single uniform entry, cross-function tail calls are a best-effort `tail call` hint, and constant-stack self-tail-calls compile as loops regardless — but it means riscv64 native binaries do not get the guaranteed constant-stack mutual tail calls (`tailcc` + `musttail`, `docs/dev/llvm-backend.md` "Guaranteed mutual tail calls") that aarch64 and x86_64 do, and every cross-function tail call pays the uniform array-ABI trampoline.

## What enabling it needs

- Confirm LLVM's RISC-V backend honours `musttail` for `tailcc` callees with up to `max_fast_arity` (8) i64 register arguments under the lp64d calling convention, including the `@r{i}.fast` forward-reference stubs. RISC-V `musttail` support is more recent than AArch64/x86-64's and its guaranteed-tail-call coverage is narrower; a `musttail` the backend cannot honour is a hard verifier/codegen error, so this is a loud check, not a silent one.
- Flip the comptime switch to include `.riscv64` and re-run `bash tests/e2e/run-e2e-cross.sh riscv64-linux <target-zig>` (CI: the `riscv64-native-test` job), with `native-mutual-tail.scm` and `native-variadic-tail.scm` being the programs that exercise the fast entries.
- A deep mutual-recursion program (deeper than the 8 MB main-thread stack could hold without constant-stack calls) on the target, since parity alone does not prove the stack stays flat.
- Update the "Per-target gate" paragraph in `docs/dev/llvm-backend.md` and the comment on `fast_tailcalls_supported`.

Note: the emitter is a comptime switch on the *host* arch, so the IR has to come from a kaappi cross-built for riscv64 — the cross script already arranges that.

baijum:
## Summary

Closes #2593 — by settling it rather than enabling it.

The issue scoped riscv64's `fast_tailcalls_supported` as "confirm LLVM honours `musttail` for `tailcc` callees under lp64d, then flip the switch". Compiling the emitter's exact fast-entry shape with the LLVM 21 inside Zig 0.16 answers one layer earlier: **LLVM's RISC-V backend does not accept `tailcc` as a function calling convention at all.** `RISCVTargetLowering::LowerFormalArguments` admits C, Fast, PreserveMost, GRAAL and the vector conventions and reports `Unsupported calling convention` for the first `define tailcc`, at -O0 and -O2, before any `musttail` is looked at (a *call* to an external `tailcc` callee lowers, which is why the uniform-only IR never trips it). LLVM `main` has no `CallingConv::Tail` arm for RISC-V either, so a toolchain bump will not change the verdict. The gate stays as it is; its comment now says why.

**`tools/probe-tailcc.sh [<zig-target>]`** turns that check into a one-line experiment: two 8-ary `tailcc` entries with `musttail` both ways, a 1→8 mixed-arity `musttail`, the uniform trampoline, compiled with `zig cc`'s own LLVM. Both ways a backend can fail are loud, so the verdict is decisive. The porting checklist now starts the `fast_tailcalls_supported` decision there, and a unit test pins the per-arch table so a flip has to touch the test that names the reason.

| Target | Verdict |
|---|---|
| aarch64 (linux, macos), x86_64 (linux, windows), s390x | SUPPORTED |
| riscv64 | UNSUPPORTED — `Unsupported calling convention` |
| ppc64le | UNSUPPORTED one layer later — `failed to perform tail call elimination on a call site marked musttail` (ten integer args, eight ELFv2 GPRs) |

**The gate switches off more than the `musttail`.** `preScanReserve` is gated on the same switch, and `isKnownOrReservedGlobal` is the only thing that lets a reference to a user-defined top-level function count as a global rather than a free variable — so on riscv64 any define that names another user function, defined before *or* after it, is compiled by the interpreter (confirmed on the cross-built kaappi: a two-function cycle emits zero native definitions, and even a backward reference falls back). That is why `native-mutual-tail.scm` passes there at 2,000,000 alternating calls: every function in it runs interpreted and the VM's own TCO keeps the stack flat, so the riscv64 e2e diff certifies the interpreter, not native codegen. `llvm-backend.md` now says so and records the measured constraints of the one route that could give riscv64 the guarantee: `fastcc` is accepted and passes twelve integer arguments in `a0–a7`/`t3–t6`, but `musttail` under any convention other than `tailcc`/`swifttailcc` demands matching prototypes (the verifier rejects the 1→8 call), so every fast entry would carry one padded `max_fast_arity`-wide signature. A probe of that shape ran 20,000,000 alternating calls in a 1 MB guest stack under QEMU where the plain-call control overflowed at 100,000.

Follow-ups worth their own issues: decoupling the pre-scan reservation from the fast entries so user-to-user calls compile natively on riscv64 (uniform ABI, forward calls through the existing indirect path); and the `fastcc` padded-prototype route for the guarantee itself.

| Run | Result |
|---|---|
| `tools/probe-tailcc.sh` on host, riscv64, aarch64 (linux/macos), x86_64 (linux/windows), ppc64le, s390x | table above |
| new gate test, filtered, host and `-Dtarget=riscv64-linux` under QEMU | 6/6 both |
| `run-e2e-cross.sh riscv64-linux` (qemu-user binfmt, prebuilt prefix) | 38/39 — the one failure is the on-target compile smoke, which cannot run under bare user-mode QEMU on an x86-64 host (the riscv64 Zig reads the host's `/usr/bin/env` and links for x86-64); CI's container covers it |
| `zig fmt --check`, markdownlint | clean |

## Checklist

- [x] `zig build test` passes
- [ ] Scheme test suites pass (`bash tests/scheme/run-all.sh`) — not run: nothing Scheme-visible changed (a comment, a unit test, docs and a probe script)


tools/probe-tailcc.sh
```diff
@@ -0,0 +1,130 @@
...
+# drops a probe.o into the caller's cwd.)
+CC1="$(zig cc "${TFLAG[@]}" -v -c "$WORK/probe.ll" -o "$WORK/probe-v.o" 2>&1 | tr -d '"' | tr ' ' '\n' || true)"
+CPU="$(printf '%s\n' "$CC1" | grep -A1 -m1 '^-target-cpu$' | sed -n 2p || true)"
+ABI="$(printf '%s\n' "$CC1" | grep -A1 -m1 '^-target-abi$' | sed -n 2p || true)"
+INFO="cpu ${CPU:-?}${ABI:+, abi $ABI}; $(zig cc --version 2>/dev/null | head -1)"
+
+for opt in -O0 -O2; do
+    if ! zig cc "${TFLAG[@]}" "$opt" -c "$WORK/probe.ll" -o "$WORK/probe.o" 2> "$WORK/err"; then
+        diag="$(grep -m1 -i 'error' "$WORK/err" | sed 's/^.*error: //' || true)"
+        echo "UNSUPPORTED: tailcc/musttail on $LABEL at $opt ($INFO): ${diag:-see stderr}"
```
baijum: P2 — this is the one spot where "both ways a backend can fail are loud" breaks down: an *invocation* failure is loud too, and it wears the same UNSUPPORTED verdict word.

Concrete case, verified: `bash tools/probe-tailcc.sh ppc64le-linux` — the spelling porting.md line 12 uses for the port, and CI's job name (`ppc64le-test`) — prints

```
UNSUPPORTED: tailcc/musttail on ppc64le-linux at -O0 (cpu ?; …): unknown architecture: 'ppc64le'
```

and exits 1, because zig's canonical spelling is `powerpc64le-linux` (with that spelling the probe does print the documented `failed to perform tail call elimination` verdict). The porting checklist now *starts* with this probe and asks the reader to decide `fast_tailcalls_supported` from its verdict — and ppc64le is precisely the arch whose porter runs it next, armed with a doc that calls the target `ppc64le-linux`. A spelling error reads as a backend rejection, the exact false-verdict failure the probe exists to prevent.

Cheapest fix is normalizing before use:

```bash
TARGET="${TARGET/ppc64le/powerpc64le}"   # zig spells it powerpc64le; our docs say ppc64le
```

Alternatively (or additionally), reserve `UNSUPPORTED` for genuine backend diagnostics — the two strings this PR teaches are `Unsupported calling convention` and `failed to perform tail call elimination` — and print something like `PROBE FAILED: <diag>` with a distinct exit code otherwise, so the header's "decisive" claim stays literally true.

baijum (commented):
Verified this end to end before commenting — the technical claims all check out; one substantive issue on the probe script (inline).

**Probe table reproduced on this machine** (zig 0.16.0 / clang 21.1.8):
- host (aarch64-macos): `SUPPORTED … cpu apple-m1, abi darwinpcs`
- `riscv64-linux`: `UNSUPPORTED … at -O0 … error in backend: Unsupported calling convention`
- `powerpc64le-linux`: `UNSUPPORTED … at -O0 … error in backend: failed to perform tail call elimination on a call site marked musttail`
- `s390x-linux`: `SUPPORTED … cpu z10`

Which is exactly the PR's table, with the two diagnostics verbatim.

**LLVM source checked** (release/21.x `RISCVISelLowering.cpp`): `LowerFormalArguments`' switch admits C, Fast, SPIR_KERNEL, PreserveMost, GRAAL, `RISCV_VectorCall` and the VLSCall vector conventions, defaults to `report_fatal_error("Unsupported calling convention")`, and has zero `CallingConv::Tail` arms in the file — the doc text is accurate down to the "plus SPIR_KERNEL in 21" parenthetical, and `isEligibleForTailCallOptimization` does reject stack-passed calls as documented.

**Code claims checked**: `preScanReserve` early-returns when the gate is off (llvm_emit_tailcall.zig:105); `isKnownOrReservedGlobal` reduces to `ir.isKnownGlobal`, which covers only built-ins and special forms — so a define naming another user function really does fail `hasFreeVars` and fall back, backward or forward. The probe's `.ll` matches the emitter's real fast-entry header (llvm_emit_lambda.zig:675: `define tailcc i64 @name.fast(ptr %vm, …)`, `%upvalues` last).

**Test**: wired into the suite via `_ = llvm_emit` in main.zig's test block; passes on the host, `zig build test -Dtest-filter=fast_tailcalls` is green, `zig fmt --check src/` clean. CI's `riscv64-test` job does run the unit chunks under QEMU, so the comment's "riscv64 row runs in CI's QEMU unit leg" is accurate. The pin works as described: flipping an arch's arm without touching the test fails that arch's unit leg (s390x/ppc64le included, since the `else` arm is what they exercise).

**Docs**: the `#guaranteed-mutual-tail-calls-tailcc--musttail` anchor matches the heading and the doc's existing usage of the same anchor (lines 212, 508); lessons-learned entry is correctly numbered after #20.

The one thing I'd fix before merging is the probe script treating a toolchain invocation failure as a backend verdict — details inline, with the concrete `ppc64le-linux` misspelling case (the spelling porting.md itself uses for the port).

docs/dev/porting.md
```diff
@@ -441,11 +441,17 @@ joined the native tier in 2026-09 — the pathfinder port described in
...
-      regardless.
+- [ ] Decide `fast_tailcalls_supported` (`llvm_emit.zig`) by running
+      `bash tools/probe-tailcc.sh <zig-target>` first: it compiles the
+      exact `tailcc`/`musttail` fast-entry shape with `zig cc`'s own LLVM
+      and fails loudly in either of the two ways a backend can — refusing
+      `tailcc` as a function convention at all (riscv64, kaappi#2593) or
+      refusing a `musttail` it cannot lower (ppc64le: ten integer
+      arguments, eight GPRs). `false` is always safe — constant-stack
+      self-tail-calls compile as loops regardless — but it costs more than
+      mutual tail calls: the pre-scan reservation is gated on the same
```
baijum: P3 nit: "it costs more than mutual tail calls" momentarily reads as a comparison against what mutual tail calls cost, rather than against merely losing them. Maybe "— but it costs more than those mutual tail calls: …" or "— but the cost goes beyond the lost mutual tail calls: …".

tools/probe-tailcc.sh
baijum: Fixed in 84d98614: `ppc64le-*` is normalized to `powerpc64le-*` before use, and UNSUPPORTED is now reserved for a backend diagnostic (the LLVM fatal `error in backend: …` both known rejections arrive as). Anything else — an unknown target, a driver error, the probe IR failing the verifier — prints `PROBE FAILED` and exits 2, the code already used when zig is missing. Verified: `ppc64le-linux` and `powerpc64le-linux` both give the musttail verdict; a nonsense target gives PROBE FAILED, exit 2.

docs/dev/porting.md
baijum: Fixed in 84d98614: now reads "but the cost goes beyond the lost mutual tail calls: …".

Issue #2604, opened 2026-09-19
x86_64-windows: kaappi compile fails with "Can't handle guaranteed tail call under win64 yet" for a mixed-arity musttail whose callee takes stack arguments


baijum:
## Summary

On a Windows x86_64 host the native backend emits `tailcc` fast entries (`llvm_emit.default_fast_abi`, the `.x86_64` row), and LLVM's X86 backend refuses a *guaranteed* tail call under the Win64 convention when the callee's argument list does not fit the caller's incoming stack-argument area:

```
error in backend: Can't handle guaranteed tail call under win64 yet
```

A same-arity `musttail` compiles (the caller's frame already holds the slots); a low-arity fast entry tail-calling a higher-arity one whose arguments spill past Win64's four integer argument registers does not. So a legal program fails `kaappi compile` on Windows x86_64 with an LLVM fatal error — loud, not silent, and the interpreter runs the program fine — while the same source compiles natively on every other native-tier host.

Measured with the LLVM 21.1 inside Zig 0.16.0, `zig cc -target x86_64-windows` (gnu and msvc alike), 2026-09-19:

| Shape (`tailcc` fast entries, `musttail`) | Verdict at -O0 |
|---|---|
| same arity, 1–8 params | compiles |
| 2→3, 3→2, 4→5, 5→4 | compiles |
| 1→5, 1→8 | `Can't handle guaranteed tail call under win64 yet` |

Reproduced from IR the emitter itself produces (the emitter is a comptime switch on the host arch, so an x86_64 Linux host emits the same IR a Windows x86_64 host does, up to the triple):

```scheme
(define (a n) (if (= n 0) 'done (begin (display n) (b n 1 2 3 4))))
(define (b p q r s t) (if (= p 0) 'done (begin <sixty display calls> (a (- p 1)))))
(display (a 3)) (newline)
```

```
kaappi --emit-llvm -o m.ll m.scm
zig cc -target x86_64-windows -O2 -c m.ll    # error in backend: Can't handle guaranteed tail call under win64 yet
zig cc -target x86_64-linux   -O2 -c m.ll    # compiles
zig cc -target aarch64-windows -O2 -c m.ll   # compiles
```

[...]

Issue #2601, opened 2026-09-18
riscv64 native tier: reserve top-level defines regardless of fast_tailcalls_supported so calls between user functions compile natively


baijum:
## Summary

On riscv64 `fast_tailcalls_supported` is false and cannot be flipped — LLVM's RISC-V backend rejects `tailcc` as a function calling convention (#2593, #2600, `tools/probe-tailcc.sh riscv64-linux`). What that gate switches off is larger than the `musttail`: `preScanReserve` (`src/llvm_emit_tailcall.zig`) returns early on the same switch, and `isKnownOrReservedGlobal` (`src/llvm_emit.zig`) is the only thing that lets a reference to a user-defined top-level function count as a global rather than a free variable. So on riscv64 any top-level define whose body names another user-defined function — defined before **or** after it — fails `hasFreeVars` and is compiled by the interpreter (`kaappi_eval_cached`). Only functions that call built-ins and themselves go native.

Measured on a cross-built riscv64 kaappi (`zig build -Dtarget=riscv64-linux`, `--emit-llvm`, 2026-09-18):

| Program | Native definitions | Define-time eval fallbacks |
|---|---|---|
| `(define (f n) … (g …))` `(define (g n) … (f …))` — a 2-cycle | 0 | 3 |
| `(define (g n) (* n 2))` `(define (f n) (g n))` — `f` refers *backward* to native `g` | 1 (`g` only) | 2 |
| `tests/e2e/programs/native-mutual-tail.scm` | 1 (`sq`) | every cycle member |

That last row is why the 2,000,000-deep mutual recursion passes on riscv64: every function in it runs interpreted and the VM's own tail-call optimisation keeps the stack flat. The riscv64 e2e diff certifies the interpreter, not native codegen (`docs/dev/llvm-backend.md`, "Per-target gate").

## Proposal

Decouple the reservation from the fast entries: run `preScanReserve` on every arch, reserving `@r{i}`/`@r{i}.fast` name pairs exactly as today, and keep defining `.fast` bodies only where the gate is on (`emitLambdaFunction` already decides `use_fast` per function). With the gate off:

- a reserved name is known to free-variable analysis, so the caller compiles natively;
- a *backward* reference resolves through `native_fns` to a direct uniform-ABI `call` (`emitDirectCall`), as on every arch;
[...]

Issue #1499, opened 2026-07-13
LLVM backend: guaranteed native mutual tail calls via tailcc + musttail (ABI change)


baijum:
Part of #1491. Structural / longer-horizon.

## Goal

Guaranteed constant-space **mutual** tail recursion in fully-native code, using
LLVM's `tailcc` calling convention + `musttail` marker — the tools the evaluation
identifies as correct for this. Recent LLVM work has extended `musttail` to
RISC-V (incl. indirect-argument cases), so this is becoming portable across
Kaappi's non-x86 targets, not just AArch64/x86-64.

## The blocker (why this is not a flag flip)

Today, native→native tail calls emit the plain `tail call` **hint** (best-effort
sibcall, not a guarantee), and general tail calls route through
`kaappi_call_scheme` into the VM trampoline. Self-tail-recursion is handled by
overwriting `%args` in place and branching (`emitSelfTailCall`).

`musttail` requires the callee's frame not to alias caller state, but arguments
are currently passed via a **caller-frame `alloca`** (`%args`). A real tail call
tears that frame down — the callee would read freed stack. So genuine mutual TCO
needs an **arg-passing ABI change**, e.g.:

- pass a bounded number of args in registers (LLVM function params), or
- add a "fast entry" variant of each native function (register args) alongside
  the uniform `(ptr %vm, ptr %args, i64 %nargs, ptr %upvalues)` entry used for
  indirect dispatch via `kaappi_call_scheme`.

## Scope

- Introduce `tailcc` on native functions and a fast direct-call entry.
- Emit `musttail call tailcc … @lambda_N(…)` + `ret` for direct tail calls to
  known native functions (covers mutual recursion among compiled functions).
- Keep `kaappi_call_scheme` for indirect/unknown callees (closures, native fns,
  continuations, parameters, FFI) — `musttail` cannot go through the C trampoline.

## Acceptance criteria

- Mutually-recursive native functions run in constant stack.
- No regression for indirect calls; `tests/e2e` + continuation suites green.
- Gated per-target on backend `musttail` support.

PR #2605, opened 2026-09-19
Enable guaranteed mutual tail calls on riscv64 via padded fastcc fast entries

### LINKED ISSUE #2602: riscv64: guaranteed constant-stack mutual tail calls via fastcc fast entries with one padded prototype
## Summary

`tailcc` is not available on riscv64: `RISCVTargetLowering::LowerFormalArguments` reports `Unsupported calling convention` for it, in LLVM 21 and on `main`, so #2593 could not be done as a flag flip (#2600; `tools/probe-tailcc.sh riscv64-linux`). The one convention that can carry a guaranteed `musttail` there is `fastcc`. Measured with the LLVM 21.1 inside Zig 0.16.0 (2026-09-18):

- `fastcc` definitions and `musttail call fastcc` compile. `CC_RISCV_FastCC` passes integer arguments in `a0`–`a7` and `t3`–`t6` (twelve GPRs), so `%vm` + `max_fast_arity` (8) + `%upvalues` = ten arguments stay in registers — which LLVM ≤ 21 requires for *any* RISC-V tail call: `isEligibleForTailCallOptimization` returns false on a non-zero outgoing stack size and `LowerCall` then reports `failed to perform tail call elimination on a call site marked musttail`. (LLVM `main` has since let `musttail` bypass that check; it still has no `tailcc`.)
- `musttail` under any convention other than `tailcc`/`swifttailcc` requires the caller's and callee's prototypes to match: the IR verifier rejects a 1→8 mixed-arity `musttail` with `cannot guarantee tail call due to mismatched parameter counts`. So every riscv64 fast entry must carry one padded, `max_fast_arity`-wide prototype, and a lower-arity caller pads its call with `undef`/0.
- A hand-written probe of exactly that shape — two `fastcc` ten-argument entries, `noinline`, an *escaping* local args array (so LLVM's own tail-call elimination cannot help), `musttail` both ways — ran 20,000,000 alternating calls in a 1 MB guest stack under `qemu-riscv64 -s 1048576`; the plain-`call` control overflowed at 100,000.

## Design sketch

Two per-arch comptime constants next to `fast_tailcalls_supported` in `src/llvm_emit.zig`: the fast-entry convention spelling (`tailcc` on aarch64/x86_64, `fastcc` on riscv64) and whether the fast signature is padded to `max_fast_arity`. Sites that read them:

- `emitLambdaFunction` (`src/llvm_emit_lambda.zig`): the `define` line, and a prologue that copies only the real parameters into `%args`;
- `emitFastTrampoline` and `emitFastCall`: pad the argument list;
- `emitForwardStub` (`src/llvm_emit_tailcall.zig`): the padded signature;
- the `tailcc` spellings asserted in `src/tests_native.zig`.

`mustTailSafe` and the reservation logic are unchanged. aarch64 and x86_64 keep `tailcc` with exact-arity signatures — no change to the established arches. `fast_tailcalls_supported` then becomes true on riscv64, which also re-enables the pre-scan there (#2601 is the cheaper first step and is subsumed if this lands).

Two things to check in the implementation rather than assume: LLVM's `DeadArgumentElimination` must not narrow an `internal` padded stub that a `musttail` targets (it skips musttail-involved functions, but a test should pin it), and `-O2` must not change a padded `fastcc` entry's convention or prototype (non-`internal` real bodies are safe; stubs are `internal`).

## Verification

- `tools/probe-tailcc.sh` stays the `tailcc` probe; give it a `--fastcc-padded` mode (or a sibling script) so the porting checklist can ask the same question of the next arch.
- `zig build test` on the host and `-Dtarget=riscv64-linux` under QEMU — the #1499 tests in `src/tests_native.zig` stop skipping on riscv64.
- `bash tests/e2e/run-e2e-cross.sh riscv64-linux <riscv64-zig>`: `native-mutual-tail.scm` and `native-variadic-tail.scm` are the programs that exercise fast entries, and they then exercise *native* code on riscv64 for the first time.
- Run the linked `native-mutual-tail` binary under `qemu-riscv64 -s 1048576` (1 MB guest stack): parity proves the output, the small stack proves it stayed flat.
- A `workflow_dispatch` of `.github/workflows/gc-stress-riscv64.yml`.

baijum:
## Summary

Closes #2602 — the route #2600 sketched after settling #2593 — and #2601, which it subsumes.

`tailcc` cannot be a function calling convention on riscv64, so the fast-entry ABI becomes a per-host table, `llvm_emit.FastAbi` / `default_fast_abi`: the convention on every fast definition and call, and whether every fast prototype is padded to `max_fast_arity`. aarch64 and x86_64 keep `tailcc` with exact-arity prototypes — **the IR they emit is byte-identical before and after**, checked on all 37 e2e programs against a pristine build of `main`. riscv64 gets `fastcc` with one eight-wide prototype per fast entry:

- `musttail` under any convention but `tailcc`/`swifttailcc` demands matching caller and callee prototypes (the verifier rejects a 1→8 call: `cannot guarantee tail call due to mismatched parameter counts`), so every riscv64 fast entry, trampoline call, direct call and forward-ref stub carries the same eight `i64`; a lower-arity call pads with `i64 0`, and the prologue and the stub copy only the real parameters.
- LLVM ≤ 21 refuses any RISC-V tail call with a stack-passed argument, so `%vm` + 8 + `%upvalues` must fit `CC_RISCV_FastCC`'s twelve integer registers; a comptime check pins `max_fast_arity + 2 ≤ 12`.

The emitter reads its `fast_abi` field, initialised from the table, so a unit test on any host can emit another arch's shape: `tests_native.zig` pins the padded riscv64 prototypes, the mixed-arity padded `musttail`, the padded stub, the untouched `tailcc` row, and — for #2601 — that a host with no row sends every define naming another user function to the interpreter, while the riscv64 row compiles the same cycle natively. On the cross-built riscv64 kaappi the #2601 programs now emit native fast entries with a `musttail` on every edge (the 2-cycle: 2 fast entries, 2 musttails, 0 `tailcc`; `native-mutual-tail.scm`: 9 fast entries, 7 musttail sites) where before every cycle member fell back.

**`tools/probe-tailcc.sh --fastcc-padded`** is the same five-second experiment for the padded row. Both modes on every target:

| Target | `tailcc`, exact arity | `fastcc`, padded |
|---|---|---|
| aarch64 (linux, macos, windows) | SUPPORTED | SUPPORTED |
| x86_64 (linux) | SUPPORTED | SUPPORTED |
| x86_64 (windows) | UNSUPPORTED — `Can't handle guaranteed tail call under win64 yet` | SUPPORTED |
| riscv64 | UNSUPPORTED — `Unsupported calling convention` | SUPPORTED |
| ppc64le | UNSUPPORTED — `failed to perform tail call elimination on a call site marked musttail` | UNSUPPORTED, the same |
| s390x | SUPPORTED | SUPPORTED |

The x86_64-windows row is new information: #2600's table listed it SUPPORTED, but the probe's 1→8 `one.fast` fails there, and so does a real program emitted by kaappi — a 1-ary caller tail-calling a 5-ary callee with non-inlinable bodies fails to compile for `x86_64-windows` at -O2 while compiling for Linux and aarch64-windows. Same-arity cycles (every cycle in `tests/e2e/programs`) compile, which is why `windows-x64-test` is green. Filed as #2604 rather than changed here: it is one row in the table now, but nothing on this machine can run a Windows binary.

**`run-e2e-cross.sh`** grows the check parity could not give: it asserts `native-mutual-tail.scm`'s IR carries `musttail` and re-runs the linked binary on a 1 MB guest stack (`QEMU_STACK_SIZE`, set through `env` inside `$EMU` so it reaches the emulator in the CI container too), where 2,000,000 alternating calls cannot fit as real frames. Negative control: the same riscv64 IR with the `musttail` markers stripped segfaults on that stack and on the default 8 MB one.

`docs/dev/llvm-backend.md` "Per-target gate" is rewritten around the table and the measured constraints; the porting checklist starts with the two-mode probe; the decision record gets a dated note.

| Run | Result |
|---|---|
| emitter tests cross-compiled for riscv64, run under QEMU (`-Dtest-filter="LLVM emit"` etc.) | 126/126 — the #1499 tests that used to skip on riscv64 now run |
| `KAAPPI_CROSS_PREFIX=… run-e2e-cross.sh riscv64-linux` (qemu-user binfmt) | 39/39, including the new constant-stack phase (7 musttail sites, 1 MB guest stack); no target zig, so the on-target smoke did not run — CI's container covers it |
| host `--emit-llvm` for the 37 e2e programs vs a pristine build of `main` | 37/37 byte-identical |
| `tools/probe-tailcc.sh` both modes × host, riscv64, aarch64 (linux/macos/windows), x86_64 (linux/windows), ppc64le, s390x | table above |
| `tests/scheme/compile` shell suite on the host | pass |
| `zig fmt --check src/`, markdownlint | clean |

## Checklist

- [x] `zig build test` passes
- [ ] Scheme test suites pass (`bash tests/scheme/run-all.sh`) — not run in full: nothing interpreter-visible changed; the native tier's suites (`tests/scheme/compile`, the e2e cross run, the emitter unit tests) are in the table above


src/llvm_emit.zig
```diff
@@ -13,38 +13,97 @@ const let_emit = @import("llvm_emit_let.zig");
...
-    // riscv64's verdict is "Unsupported calling convention" (LLVM 21).
-    const expected = switch (@import("builtin").cpu.arch) {
-        .aarch64, .x86_64 => true,
-        .riscv64 => false,
-        else => false,
+pub const tailcc_abi: FastAbi = .{ .supported = true, .cc = "tailcc", .padded = false };
+pub const padded_fastcc_abi: FastAbi = .{ .supported = true, .cc = "fastcc", .padded = true };
+pub const no_fast_abi: FastAbi = .{ .supported = false, .cc = "tailcc", .padded = false };
+
+pub const default_fast_abi: FastAbi = switch (@import("builtin").cpu.arch) {
```
baijum: **Question — should `(x86_64, windows)` take a different row in this table now?**

The probe table in llvm-backend.md establishes that this row (inherited from the old `fast_tailcalls_supported` switch) makes `kaappi compile` a *fatal* backend error on Windows x64 for any legal program whose mutual-tail cycle mixes arities (#2604) — `Can't handle guaranteed tail call under win64 yet` — while the probe's padded mode is SUPPORTED there. The bug predates this PR, but this table is what turns the fix into a one-row change; the switch would just need `builtin.os.tag` alongside `cpu.arch`.

Deferring is defensible — nothing here can run a Windows binary — but noting the interim option: `no_fast_abi` for `(x86_64, windows)` is strictly conservative (trades the fatal error for interpreted fallback of defines that name other user functions) and, unlike the padded row, needs no Windows machine to reason about. If #2604 isn't picked up soon, that degradation seems worth a one-line follow-up PR of its own.

src/tests_native.zig
```diff
@@ -1195,14 +1217,141 @@ test "LLVM emit: a forward-referenced non-native define gets a tailcc stub (#149
...
+    var res = try emitMultiResultWithFastAbi("(define (one n) (if (= n 0) 0 (two n 1)))" ++
+        "(define (two a b) (if (= a 0) b (one (- a 1))))", llvm_emit.padded_fastcc_abi);
+    defer res.deinit();
+    const ll = res.toSlice();
+    try expectNativeDef(ll, "one");
+    try expectNativeDef(ll, "two");
+    try std.testing.expectEqual(@as(usize, 2), std.mem.count(u8, ll, "musttail call fastcc i64 @r"));
+    // one -> two passes two real arguments and six fillers, two -> one one
+    // real argument and seven; the trampolines end in `ptr %upvalues`.
+    try std.testing.expectEqual(@as(usize, 2), std.mem.count(u8, ll, ", i64 0, i64 0, i64 0, i64 0, i64 0, i64 0, ptr null)"));
```
baijum: **nit: the `2` here is only correct via substring overlap.** The 6-filler needle also matches inside the 7-filler site (`", i64 0"`×7 + `", ptr null)"` ends with the 6-filler pattern), and the two trampolines are excluded only because they end in `ptr %upvalues)` rather than `ptr null)`. Correct today, but nothing in the comment says the second `musttail` site contributes to the first needle's count — a future reader "fixing" the number to 1 (or adding a new call site ending in `ptr null)`) gets a confusing failure either way. One extra sentence in the comment above, or anchoring the needles on the real-arg temps (`…, i64 %t…, i64 0, …, ptr null)`), would pin the intent.

tests/e2e/run-e2e-cross.sh
```diff
@@ -214,6 +218,55 @@ if [[ -n "${KAAPPI_E2E_PROGRAMS:-}" && $MATCHED -eq 0 ]]; then
...
+elif ! grep -q 'musttail call' "$mt_ll"; then
+    echo "FAIL: constant-stack mutual tail calls — no musttail in the emitted IR (did the cycle compile natively?)"
+    FAIL=$((FAIL + 1))
+elif ! mt_cc=$(cross_link "$mt_ll" "$mt_bin" 2>&1); then
+    echo "  cc: $mt_cc"
+    echo "FAIL: constant-stack mutual tail calls — cross-link failed"
+    FAIL=$((FAIL + 1))
+else
+    mt_status=0
+    mt_actual=$($EMU env QEMU_STACK_SIZE=1048576 "$mt_bin" 2>&1) || mt_status=$?
```
baijum: Traced the `env`-inside-`$EMU` trick against the actual CI wiring (`KAAPPI_EMU="docker run --platform linux/riscv64 …"` + docker/setup-qemu-action's F-flag binfmt) and the empty-`KAAPPI_EMU` binfmt form: `env` sets the variable and `execve`s the binary, binfmt re-enters QEMU with `QEMU_STACK_SIZE` now in the emulator's own environment, so the fresh emulator honours it. Correct for both documented forms.

One footnote for the record: a *direct* prefix like `KAAPPI_EMU="qemu-riscv64"` (not a form the header offers) would misfire on this line only — qemu would try to load the host's `/usr/bin/env` as a guest binary, where every other `$EMU "<path>"` invocation in the script works. No change needed; just don't grow that usage later without revisiting this line.

baijum (commented):
Reviewed and verified locally (aarch64-macos host, zig cc's LLVM 21.1). Everything this PR claims that is checkable on this machine checks out:

- **Byte-identity on the established arches**: built `main` and this branch, emitted `--emit-llvm` for all 37 `tests/e2e/programs` with an isolated `KAAPPI_HOME` each — **37/37 byte-identical**.
- **Unit tests**: the new `default_fast_abi` test, the `#2602` padded-shape tests, and the `#2601` coupling test all pass on the host, along with the full `LLVM emit` filter; `zig fmt --check src/` clean; markdownlint-cli2 clean on the three changed docs.
- **Probe table fully reproduced** — both modes × host (aarch64-macos), `riscv64-linux`, `x86_64-windows`, and `powerpc64le-linux`: identical verdicts and diagnostics, including the newly-discovered x86_64-windows row (`Can't handle guaranteed tail call under win64 yet` in tailcc mode, SUPPORTED padded) and ppc64le refusing both modes with the same message.
- **`tests/scheme/compile` suite**: 39/39 pass from the repo root with a fresh `KAAPPI_HOME`.
- **Dispatch completeness**: no emission site reads the comptime table anymore — the fast-entry definition, trampoline, `emitFastCall`, and the forward-ref stub all read `self.fast_abi`, which is exactly what makes `emitMultiResultWithFastAbi` faithful on a foreign host; `expectEveryFastSignatureWidth` scanning *every* `define`/`call` line is a strong guard against a future missed padding site.

The not-run-here items are the QEMU ones (the cross e2e including the new 1 MB-stack phase, and the riscv64-hosted unit legs) — covered by the PR's table and by CI's `riscv64-native-test` job.

Two inline notes (a question on the x86_64-windows row, a nit on a test's substring-overlap count), plus one doc item that isn't in this diff:

**`docs/dev/postmortems/2026-09-14-riscv64-native-segfault.md` now states the opposite of the new state** — its Status says `fast_tailcalls_supported` "stays `false`" on riscv64, and "What the port needed" says mutual recursion "has no constant-stack guarantee there". `ci.yml`'s `riscv64-native-test` job links that postmortem, so it's one hop from a reader who'd be misled. Suggest a dated note in the Status section, matching the style of the one this PR adds to the decisions record — e.g. *(2026-09-19: superseded by #2602 — riscv64 now has padded `fastcc` fast entries; see llvm-backend.md, "Per-target gate".)* Fine to skip if postmortems are treated as strictly frozen records.

src/llvm_emit.zig
```diff
@@ -13,38 +13,97 @@ const let_emit = @import("llvm_emit_let.zig");
...
+    .aarch64, .x86_64 => tailcc_abi,
+    .riscv64 => padded_fastcc_abi,
+    else => no_fast_abi,
```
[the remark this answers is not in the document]
baijum: Taken in 8c8f6495: `default_fast_abi` returns `no_fast_abi` whenever `native_backend_supported` is false, and the table test expects the same gate. The old boolean switch had the same latent failure on an aarch64/x86_64/riscv64 host whose OS has no `targetTriple` arm; no supported host changes row.

tests/e2e/run-e2e-cross.sh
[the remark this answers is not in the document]
baijum: Not taken as written: a host-side assignment never enters the container, so CI's docker form would have run the check on the default 8 MB stack and passed it vacuously. The `env`-inside form is what makes the variable reach the emulator there (the on-target smoke relies on the same shape for PATH).

The bare-emulator case is real, though. 8c8f6495 probes once whether `$EMU` can run `env` and sets `QEMU_STACK_SIZE` inside `$EMU` when it can, outside otherwise, so the empty (binfmt), container and bare `qemu-riscv64` forms all work. Verified 39/39 with the constant-stack phase green under both the empty form and `KAAPPI_EMU=qemu-riscv64`; the header now documents the bare form (every phase but the on-target smoke).

src/llvm_emit.zig
baijum: Yes, and the padded `fastcc` row is the fix rather than `no_fast_abi`: the interim option trades a loud failure on one program shape (a low-arity fast entry tail-calling a higher-arity one with stack arguments) for interpreting every define that names another user function, the #2601 regression, on every Windows x64 program.

The row is one line either way now that the table is per host (8c8f6495 gates it on `native_backend_supported`, so it is already OS-aware), and `windows-x64-test` running `run-e2e.ps1` on windows-latest is the verification a Windows codegen change needs and this machine cannot provide. Keeping it as #2604's own PR rather than a Windows codegen change riding on the riscv64 one.

src/tests_native.zig
baijum: Fixed in 8c8f6495: the test now counts `, i64 0` per `musttail call fastcc` line, exactly one line with six fillers and one with seven, any other width failing, so the two widths cannot match inside each other and the intent no longer depends on the trampolines ending in `ptr %upvalues`.

tests/e2e/run-e2e-cross.sh
baijum: Agreed on both documented forms, and the footnote's bare-prefix case is handled too as of 8c8f6495: the phase probes `$EMU env true` once and sets the variable outside `$EMU` when the prefix cannot run `env`. Verified with `KAAPPI_EMU=qemu-riscv64`: 39/39, constant-stack phase green.

PR #2606, opened 2026-09-21
Use padded fastcc fast entries on x86_64-windows, where Win64 refuses a growing musttail

### LINKED ISSUE #2604: x86_64-windows: kaappi compile fails with "Can't handle guaranteed tail call under win64 yet" for a mixed-arity musttail whose callee takes stack arguments

baijum:
## Summary

Closes #2604.

On an x86_64 Windows host the native backend emitted `tailcc` fast entries, and LLVM's X86 backend refuses a guaranteed tail call under the Win64 convention when the callee's arguments do not fit the caller's incoming stack-argument area: `Can't handle guaranteed tail call under win64 yet`, a fatal backend error from `kaappi compile`. Win64 passes four integer arguments in registers, so a 1-ary fast entry tail-calling a 5-ary or wider one failed, while every same-arity cycle (all of `tests/e2e/programs`) compiled, and at -O2 a small callee was inlined into its caller and the offending `musttail` vanished with it, which is how `windows-x64-test` stayed green.

- **`llvm_emit.fastAbiFor(arch, os)`** replaces the arch-only switch behind `default_fast_abi`; x86_64-windows takes the padded `fastcc` row #2605 added for riscv64. With one `max_fast_arity`-wide prototype every `musttail` is a sibling call into exactly the caller's own argument area, which is what Win64 permits; six of the ten integer arguments travel on the stack and that is fine, so riscv64's register bound does not apply. aarch64-windows keeps `tailcc` (AAPCS64 passes eight in registers). The unit test now pins every row of the table on every host, and `windows-x64-test`'s `unit-tests.exe` pins the Windows row on the real target.
- **`tests/e2e/programs/native-mixed-arity-tail.scm`**: a 1-ary ↔ 8-ary cycle at 1,000,000 alternations whose 8-ary body calls a non-inlined primitive, so the mixed-arity `musttail` survives -O2 (a `kaappi_call_scheme` call allocates its args array outside the entry block, and LLVM's inliner never inlines a function with a dynamic alloca). Under the previous `tailcc` emission its IR fails to cross-compile for x86_64-windows at -O0 *and* -O2; under the padded emission it compiles at both. `windows-x64-test` runs it through `run-e2e.ps1` on every PR. The wide function also checks that all seven derived arguments (five of them stack-passed on Win64) arrive intact.
- **`run-e2e-cross.sh`**'s 1 MB guest-stack phase covers both cycle programs, so the padded mixed-arity `musttail` is exercised on riscv64 too.
- Docs: `llvm-backend.md` "Per-target gate" (the x86_64-windows row is no longer a live gap), `porting.md` checklist, `windows.md`, the decision record, and the probe script's header.

## Test plan

| Run | Result |
|---|---|
| `bash tests/e2e/run-e2e.sh` | 39/39 (the BDD phase skipped: `kaappi-bdd` is not checked out next to this worktree; 40/40 with it, per review) |
| `tools/probe-tailcc.sh x86_64-windows` plain / `--fastcc-padded` | UNSUPPORTED (`Can't handle guaranteed tail call under win64 yet`) / SUPPORTED |
| new program's IR under the previous `tailcc` emission, `zig cc -target x86_64-windows -c` at -O0 and -O2 | fails both (the regression is visible at the level `kaappi compile` links at); compiles for x86_64-linux and aarch64-windows |
| all 38 e2e programs' IR emitted under the padded row (Linux row temporarily padded to emit the Windows shape), `zig cc -target x86_64-windows` at -O0 and -O2 | 38/38 compile |
| the same padded codegen linked and run natively on x86_64-linux | 38/38 parity with the interpreter; both cycle programs pass on a 1 MB stack (`ulimit -s 1024`) |
| `--emit-llvm` for the 38 e2e programs, final binary vs a pristine build of the previous commit, x86_64-linux | 38/38 byte-identical |
| `run-e2e-cross.sh riscv64-linux` (qemu-user binfmt) | 41/41, constant-stack phase passing for both programs |
| `tests/scheme/compile` shell suite | pass |
| `zig fmt --check`, `npx markdownlint-cli2` | clean |

Nothing on this machine can run a Windows binary; `windows-x64-test` is the end-to-end verification on the real target, and it now has a program that fails without this change.


baijum (commented):
Reviewed from a fresh worktree of this branch; no blocking findings. Everything checkable on this machine (macOS aarch64) checks out — including the central cross-compile claim, reproduced both ways:

- `tools/probe-tailcc.sh x86_64-windows` → UNSUPPORTED: `Can't handle guaranteed tail call under win64 yet` (plain) / SUPPORTED at -O0 and -O2 (`--fastcc-padded`). Probe rows for aarch64-macos, riscv64-linux (both modes), and aarch64-windows all behave as the new table says.
- This program's IR emitted under the **tailcc** row fails `zig cc -target x86_64-windows` at **both -O0 and -O2** — the -O2 half being the one that matters, since `kaappi compile` links at -O2 and the whole gap hid behind the inliner. Under a padded row it compiles at both. All 38 programs' padded IR cross-compiles for x86_64-windows at -O2 (38/38).
- `zig build test` (the new `fastAbiFor` test pins every row on every host), `run-e2e.sh` 40/40, `tests/scheme/compile` suite 39/39, `zig fmt --check` and markdownlint clean.
- Byte-identity on a tailcc host: `--emit-llvm` for all 38 programs + test-llvm-backend.scm is byte-identical between this branch and an origin/main build here (39/39 files) — matches the PR's x86_64-linux result.

Also confirmed structurally: `windows-x64-test` runs `unit-tests.exe` plus `run-e2e.ps1`, which globs `programs\*.scm`, so the new program runs end-to-end on the real target on every PR; `run-e2e.ps1`'s parity phase doubles as the constant-stack check there (1M alternations can't fit as frames on any default stack). The riscv64 half stays with `riscv64-native-test` — no qemu/container image on this box to run it locally.

One bookkeeping note: the PR test table says `run-e2e.sh` 39/39, but with the new program the suite is 40/40 here (1 BDD + 38 programs + argv) — presumably that row was recorded before the program landed.

Two doc nits and a file-size heads-up inline; none blocking.

src/llvm_emit.zig
```diff
@@ -35,6 +39,22 @@ const Value = types.Value;
...
+//   fast entry (%vm, one value, %upvalues) tail-calling a 5-ary or wider one
+//   fails, while a same-arity cycle — every cycle in tests/e2e/programs
+//   before #2604 — compiles, and at -O2 a small callee is inlined into its
+//   caller and the offending `musttail` disappears, which is how the gap
+//   hid. The padded row fixes it: with one `max_fast_arity`-wide prototype
+//   every `musttail` is a sibling call whose stack-argument area is exactly
+//   the caller's own, and the backend's objection is specifically to
+//   *growing* that area — six of the ten integer arguments being
+//   stack-passed is fine, so riscv64's register bound does not apply.
+//   aarch64-windows is unaffected: AAPCS64 passes eight integer arguments
```
baijum: Doc nit, non-blocking: the "AAPCS64 passes eight integer arguments in registers" rationale (here, again at :137, in llvm-backend.md, and in windows.md) isn't quite the mechanism. The fast-entry prototype carries *ten* integer arguments (`%vm` + 8 values + `%upvalues`), so two travel on the stack under AAPCS64 as well — the 1→8 mixed-arity shape does grow the callee's argument area on aarch64 too. What actually keeps aarch64-windows on `tailcc` is that the AArch64 backend has no Win64-style refusal — it lowers a musttail with stack-passed callee arguments as a sibling call into the caller's own area — which is exactly what the probe's SUPPORTED verdict certifies (re-ran it from here: SUPPORTED at -O0/-O2). Since the row rests on the probe anyway, it might be worth rewording these four spots to lean on that verdict rather than on the register count, which a future reader can check and find doesn't cover ten arguments.

src/llvm_emit.zig
```diff
@@ -65,16 +85,23 @@ pub const tailcc_abi: FastAbi = .{ .supported = true, .cc = "tailcc", .padded =
 pub const padded_fastcc_abi: FastAbi = .{ .supported = true, .cc = "fastcc", .padded = true };
 pub const no_fast_abi: FastAbi = .{ .supported = false, .cc = "tailcc", .padded = false };
 
-pub const default_fast_abi: FastAbi = if (!native_backend_supported)
+/// The fast-entry ABI row for a host `(arch, os)` — the table behind
+/// `default_fast_abi`, as a function so a unit test on any host can pin
+/// every row rather than only its own.
+pub fn fastAbiFor(arch: std.Target.Cpu.Arch, os: std.Target.Os.Tag) FastAbi {
```
baijum, line 92: Heads-up, non-blocking: this lands `llvm_emit.zig` at 1499 lines — one under the 1500 ceiling (1451 on main). Nothing to do in this PR, but the next addition here trips the file-size policy, and the `FastAbi` commentary block above it (~80 lines of per-target rationale and probe guidance) is a natural seam for an `llvm_emit_fast_abi.zig` split when that happens.

tests/e2e/programs/native-mixed-arity-tail.scm
```diff
@@ -0,0 +1,46 @@
...
+; `musttail call @wide.fast(%vm, n, …seven more…, %upvalues)` — ten
+; arguments, six of them on the stack, from a caller whose own prototype put
+; none there — was a fatal "Can't handle guaranteed tail call under win64
+; yet" from `kaappi compile`. Under the padded fastcc row every fast
+; prototype is max_fast_arity wide, so the call is a sibling call into the
+; caller's own argument area and compiles (docs/dev/llvm-backend.md,
+; "Per-target gate"). aarch64-windows and every non-Windows host compiled
+; this all along.
+;
+; wide does its work through a primitive the emitter does not inline (max)
```
baijum, line 16: This design is the part worth protecting in future edits: `max` makes `wide` non-inlinable at -O2 — the level `kaappi compile` actually links at — via the dynamic alloca in the calling block, and the sum check turns a shifted or dropped stack slot into an immediate `bad-arguments` parity mismatch instead of a silently wrong countdown. Verified locally: under the tailcc row this program's IR fails `zig cc -target x86_64-windows` at -O0 *and* -O2 (so the inliner-invisibility path is real), and under a padded row it compiles at both.

src/llvm_emit.zig
baijum: Agreed, the register count was not the mechanism: the prototype is ten integer arguments, so two are stack-passed under AAPCS64 too and the 1→8 shape grows the area there as well. Reworded all four spots (both in `llvm_emit.zig`, `llvm-backend.md`, `windows.md`) in 60cdaed2 to say that the AArch64 backend has no Win64-style refusal — a growing `musttail` lowers as a sibling call into the caller's area — and that the row rests on the probe's SUPPORTED verdict.

src/llvm_emit.zig
baijum: Noted. The rewording in 60cdaed2 is trimmed to keep the file at exactly 1500 lines, so the next addition here is the trigger for an `llvm_emit_fast_abi.zig` split along the `FastAbi` commentary seam, as you suggest.

tests/e2e/programs/native-mixed-arity-tail.scm
baijum: Thanks for reproducing the -O0/-O2 half; the header comment on the program spells out both mechanisms (the dynamic alloca that keeps `wide` out of the inliner, and the sum check that turns a shifted stack slot into a parity mismatch) so a future edit knows what not to simplify away.

